Issue #1065, opened 2026-08-16
BLOCKER: a slot that boots but fails the health gate is never rebooted — the automatic rollback two operator docs promise does not exist


VijitSingh97:
From the v2 gap analysis. The A/B design's headline promise is that a bad update reverts itself. For the most likely kind of bad update, it does not.

## Mechanism

`pithead-boot` leaves the slot uncommitted and exits 1 (`os/overlay/pithead-boot:167`). `pithead-boot.service` is `Type=oneshot` with **no `Restart=`, no `OnFailure=`**, and nothing else under `os/overlay/` reboots on a failed boot unit.

The fallback is a *GRUB* decision, and GRUB only gets to decide at the **next** boot. So the machine sits on the broken slot indefinitely, with the stack down, waiting for someone to power-cycle it.

Two failure shapes do self-heal: a kernel panic (`panic=60`, `os/rauc/grub.cfg:78`) and a wedged PID 1 (`RuntimeWatchdogSec=20s`). Neither covers the likeliest regression an OS update actually ships — **a clean boot with a dead stack**: a bad compose render, a bad image, an unparseable Caddyfile, a crashed `monerod`. That is exactly the case the health gate was written to catch, and catching it leaves the box stranded.

Remote recovery only works if the dashboard on the *broken* slot is still serving, which by definition it is not: `control_os_reboot` refuses unless `in-flight.json` is present and under 24 hours old.

## The docs promise otherwise

`docs/appliance.md` and `docs/dashboard.md` both tell the operator the machine ends up on a working system with nobody present. `docs/dev/appliance-release.md:375` describes the real behaviour correctly — so the developer-facing doc is right and both operator-facing docs are wrong.

For an appliance sold on "a non-expert can run it", "your mining box is offline until you walk over and pull the power" is a materially different product promise.

## Fix — one of two, not both

1. Add the missing mechanism: an `OnFailure=` or `ExecStopPost=` reboot on the `pithead-boot` unit, **bounded by a boot-attempt counter** so it cannot loop forever when the fault lives on `/data` and therefore survives the rollback.
2. Or correct both operator docs to say a power cycle is required, and say so on the console when the gate fails.

Then run manual battery item **M9** as written — "the machine reverts to the previous version without intervention" — which has never been run and is the item that would have caught this.

## Coverage

The battery proves an *uncommitted slot that never boots* reverts, which is a different failure. No tier covers a slot that boots cleanly and fails the health check.

[...]

Issue #791, opened 2026-07-30
Appliance cannot install the control runner — read-only root blocks /etc/systemd/system


VijitSingh97:
## What happens

`./pithead apply` on an appliance dies with:

```
tee: /etc/systemd/system/pithead-control.service: Read-only file system
[ERROR] pithead aborted unexpectedly (exit 1).
```

`provision_control_runner` installs the #33 control-runner unit under `/etc/systemd/system`.
On the appliance the root filesystem is **read-only by design** — that is the point of the A/B
image — so the write cannot succeed. Only `/data` and the `/var` overlay are writable.

## Why it matters

`dashboard.control.enabled` now defaults **true** on appliances (deliberately: with no shell
and SSH disabled, the dashboard's config editor is the ONLY way to change configuration after
first boot — see #786). That default is useless, and `apply` is fatally broken, while the unit
cannot be written. So the appliance's single supported management path does not work.

It escaped the KVM phases because they assert containers-running and a dashboard that answers;
neither exercises `apply` on a provisioned appliance, nor the control channel end to end.

## The fix, and it fits the framing from #790

A generated systemd unit is **derived**, not state: a pure function of config plus program.
`/run/systemd/system` is writable (tmpfs), is a first-class unit search path systemd reads, and
is regenerated every boot by construction — so it self-heals staleness and corruption for free,
exactly as #790 argues derived files should.

- On an appliance, render host units into `/run/systemd/system` and `systemctl daemon-reload`.
- DIY keeps `/etc/systemd/system` (persistent root, an operator expects the unit to survive).
- Whatever renders them must run on every boot, not once at provisioning, since `/run` is
  cleared — folding naturally into #790's re-render-on-boot work.

## Test coverage this needs

- `tests/os/run.sh --phase provision`: run `pithead apply` on the provisioned appliance and
  require exit 0 — the single cheapest assertion that would have caught this.
- Exercise the control channel once end to end on the appliance (submit a trivial config change
  through the dashboard's preview/commit path), since that is now the documented management
  route.

VijitSingh97:
[...]

PR #1152, opened 2026-08-20
fix(doctor): look for the control units where apply writes them (#1151)

### LINKED ISSUE #1151: CRITICAL: doctor looks for the control units in /etc, the appliance puts them in /run — so no A/B update ever commits, and #1065 reboots the box
## An A/B update on a provisioned appliance never commits, and #1065 then reboots it

Found on the bench today, once #1149's fix let leg 4 of `--phase update` run past its first step for
the first time. Leg 4 now gets all the way through check → download → resume → verify → install →
reboot → **boots the new slot**, and then fails its last two assertions:

```
✗ leg 4: no 'updated' verdict after the reboot (got 'none') — commit or verdict is broken
✗ leg 4: /api/state does not carry the updated verdict — the banner would never show
```

## What is actually happening, measured on the guest

The box booted the new version and is completely healthy:

```
running version   1.19.2 (the update)
containers        tor, tari, caddy, monerod, docker-control, docker-proxy, dashboard — all Up, healthy
https://localhost/  -> 401   (the real dashboard vhost, auth-gated)
```

But **18 minutes after boot**:

```
systemctl is-active pithead-boot   ->  activating
os-update-state.json .step         ->  reboot-pending
in-flight.json                     ->  still present (never consumed)
rauc status Activated              ->  none
```

The boot gate is looping (72 `doctor` invocations and counting). Its curl half passes. Its `doctor`
half does not:

```
[pithead] Diagnostics summary: 25 OK, 4 warning(s), 1 failure(s).
  ✗ FAIL The control channel is enabled but no runner units are installed — the dashboard's
         config changes and one-click upgrades will never run, with no error shown.
         Fix: run './pithead apply' from this directory.
```

**The units are installed.** They are just not where `doctor` looks:

```
/run/systemd/system/pithead-control.path      present (written at boot)
/run/systemd/system/pithead-control.service   present
/etc/systemd/system/pithead-control.*         ABSENT
dashboard.control.enabled                     true
```

And `check_control_units` opens with

```bash
unit_dir="${PITHEAD_UNIT_DIR:-/etc/systemd/system}"
```

**On the appliance the control units live in `/run/systemd/system/`, not `/etc`.** Unless
`PITHEAD_UNIT_DIR` is exported into the environment `pithead-boot` runs `doctor` in, the check looks
in `/etc`, finds nothing, and reports a hard failure on a perfectly healthy box.

## Why this is severe

1. The gate never passes, so **the slot is never committed** and the operator's dashboard sits on
   *"reboot pending"* for ever after a successful update.
2. After **#1065**, a gate that expires **reboots the box** — so a healthy, correctly-updated
   appliance rolls itself back. Repeatedly.
3. It only bites a **provisioned** box: legs 1–3 pass because an unprovisioned guest has
   `dashboard.control.enabled=false`, so this check returns early. That is every real operator's box
   and none of the cheap tests.

This is the #1068 family — a **false RED** on the gate that decides whether an update lives — and it
is the counterpart to #1140, where the same gate's other half is a false GREEN.

## Fix

Make `check_control_units` look where the units actually are on an appliance: search `/run/systemd/system`
as well as `/etc/systemd/system` (systemd's own precedence order), or have `pithead-boot` export
`PITHEAD_UNIT_DIR`. Prefer the former — a check that only works when its caller remembers to set an
env var is one env var away from lying again.

## Tier

Tier 1 for the search-path change (drive `check_control_units` with units in `/run` and nothing in
`/etc`; the mutation is putting the search path back to `/etc` only). Tier 4 to confirm the whole
leg-4 sequence goes green — the guest from this run is still up with `--keep`.

VijitSingh97:
Closes #1151.

`pithead-boot`'s health gate runs `doctor`. `check_control_units` opened with a hard-coded
`/etc/systemd/system` default, while `provision_control_runner` had known since #791 that the
appliance's read-only root forces those units into `/run/systemd/system`. The reader looked where
the writer never writes, found nothing, and reported a hard failure about units that were
installed, enabled and running. The gate never passed, the slot was never committed, and the
dashboard sat on "reboot pending" after an update that had in fact succeeded.

Two rules, not a wrong constant — so the fix is one rule. `control_unit_dir` resolves the directory
once (explicit `PITHEAD_UNIT_DIR` > appliance `/run/systemd/system` > `/etc/systemd/system`) and the
writer and both readers call it. `control_units_owner_dir` carried the same stale default and is
fixed by the same change.

## Bench evidence — real hardware, A/B on one box

Driven on the kept KVM guest from the leg-4 run (provisioned, cold-booted, running 1.19.2). Same
box, same healthy stack, only the binary differs:

| | `doctor` | control-channel verdict | `doctor --json` |
|---|---|---|---|
| base (`develop-v2`) | 25 OK, 4 warn, **1 FAIL** | ✗ no runner units are installed | **rc=1** |
| this branch | 26 OK, 4 warn, **0 FAIL** | ✓ Control runner units target this install. | **rc=0** |

The base numbers reproduce the issue report exactly. `rc=1` is why the gate's `&&` could never
pass.

End to end, with the fixed binary in place `pithead-boot` went `activating` → `active` in about two
minutes and finished the job it had been looping on for eighteen:

```
rauc status: marked slot rootfs.0 as good
pithead-boot: stack is serving (HTTP 401) and doctor reports healthy — booted slot committed
pithead-boot: OS update to 1.19.2 committed — verdict published to the dashboard
```

Those are the two assertions leg 4 of `tests/os/run.sh --phase update` failed on. Evidence log kept
at `/tmp/1151-bench-evidence.log` on the battery host.

## Tier 1

`bash tests/stack/run.sh` → **2700 passed, 0 failed** (base 2695).

Seven new assertions in two halves, because `/run/systemd/system` is root-owned and this tier cannot
plant a unit in the real appliance directory. The path is proven as two links:

- **link 1, the rule** — `control_unit_dir` answers `/run/systemd/system` when `is_appliance`,
  `/etc/systemd/system` otherwise, and an explicit `PITHEAD_UNIT_DIR` wins over both.
- **link 2, the binding** — `check_control_units` reads whatever the helper answers, with
  `PITHEAD_UNIT_DIR` deliberately **unset** so it cannot reach the units by the override arm
  instead. It calls `control_units_owner_dir`, so one run covers both readers. The third consumer,
  `provision_control_runner`, was already behaviourally bound by the #791 writer assertion.

Plus one static check that no second copy of the `/etc` default survives anywhere in the program —
static because no runtime test can see a copy that does not exist yet, and a second copy is exactly
how the two rules drifted apart.

## Mutations — six named, six run

Baseline 2700/0. Every row measured against the revision being merged.

| # | mutation | result |
|---|---|---|
| M1 | restore `${PITHEAD_UNIT_DIR:-/etc/systemd/system}` in `check_control_units` | 2697/**3** |
| M2 | the same in `control_units_owner_dir` | 2697/**3** |
| M3 | `control_unit_dir`'s appliance arm returns `/etc` | 2697/**3**, incl. both pre-existing #791 writer assertions |
| M4 | `[ -n "${PITHEAD_UNIT_DIR:-}" ] \|\| unit_dir=/etc/systemd/system` after the helper call — `check_control_units` only | 2698/**2** |
| M5 | the same respelling — `control_units_owner_dir` only | 2698/**2** |
| M6 | the same respelling in **both** readers | 2698/**2** |

M4–M6 exist because the adversarial review found them. The first version of this branch bound the
readers with a source-text grep only, and three independent lenses each reproduced a full return of
#1151 — spelled any way other than the one literal the grep looked for — shipping with the whole
suite green. That is this repo's signature defect, in the test written to prevent it. The
behavioural binding (link 2) is what closes it, and M5/M6 now go red printing the exact field
verdict:

```
✗ FAIL The control channel is enabled but no runner units are installed …
```

The review also corrected two claims in the first draft of the docs. Both were re-derived here
before changing:

- **why legs 1-3 miss this.** Not "the control channel is off and the check returns early" — those
  legs never provision the guest, so `pithead-boot.service`'s `ConditionPathExists` is unsatisfied
  and the gate never runs at all; they commit with the harness's own `rauc status mark-good`
  (`tests/os/run.sh:287`, `:715`). Leg 4 is the only leg that provisions **and** lets the boot gate
  commit. It matters because the wrong version implies a cheaper leg could have caught this.
- **what #1065 does.** Not "rolls itself back on every attempt". `fail_boot` is bounded to one
  automatic reboot and only a passing gate clears the counter, so from the second failure the box
  deliberately stays up (`os/overlay/pithead-boot:41-54`) — stranded on an uncommitted slot rather
  than looping. Someone matching a field symptom against the wrong version would look for a reboot
  loop and find a box parked on the new version instead.

A `/ponytail-review` pass after that took 20 lines back out: the three body-grep assertions became
redundant once the readers were bound behaviourally, and the second test sandbox was a copy of one
four lines above it.

## Lint

`make lint` clean through `lint-sh`, `lint-docs-voice` and `lint-operator-strings`. `lint-proto`
needs docker and was **not** run locally — CI covers it.

## Not done

- No new `--phase update` battery run. The fix is proven against the real appliance by the A/B
  above — the same measurement leg 4's two failing assertions were making — but the full ~55 min leg
  has not been re-run on this branch.
- Patch coverage is not meaningful here: the change is bash, so `diff-cover` reports no lines with
  coverage information rather than a real number.
- `develop` is untouched and needs no equivalent — it has no `is_appliance` and no appliance, and
  its three `/etc` defaults are correct for a DIY host.

### DIFF
--- a/os/KNOWN-ISSUES.md
+++ b/os/KNOWN-ISSUES.md
@@ -183,6 +183,21 @@ starts. Host identity survives A/B updates and reboots by construction (it never
 system slot to begin with); the reset tiers classify it like everything else on `/data` — a
 keep-everything reinstall keeps it, a fresh-start or full wipe deletes it.
 
+**Fixed — `doctor` looked for the control units in `/etc`, and the appliance puts them in
+`/run` (#1151).** `provision_control_runner` knew the read-only root cannot take that write (#791);
+`check_control_units` and `control_units_owner_dir` did not, and each kept its own `/etc` default.
+On a provisioned box `doctor` therefore reported "no runner units are installed" about units that
+were installed and running. That verdict is the `doctor` half of `pithead-boot`'s health gate, so
+the gate never passed, the A/B slot was never committed, and the dashboard sat on "reboot pending"
+after an update that had in fact succeeded. #1065 makes the first such boot reboot the box; the
+counter is bounded and only a passing gate clears it, so from the second failure on the machine
+deliberately stays up — a healthy, correctly updated appliance parked on an uncommitted slot with
+no way to clear it but a power cycle. Legs 1-3 of the battery cannot see this: they never provision
+the guest, so `pithead-boot` is condition-skipped and the gate never runs at all, and they commit
+with the harness's own `rauc status mark-good`. Leg 4 is the only leg that provisions and then lets
+the boot gate do the committing, and it found this the first time it ran to completion. The
+directory is now one rule, `control_unit_dir`, read by the writer and both readers.
+
 **Fixed — a provisioning failure now surfaces its own reason on the reopened wizard.**
 Setup failures move the config aside as `config.json.failed` and re-mint a token; the
 reopened page used to make an operator dig the reason out of the console. It no longer
--- a/pithead
+++ b/pithead
@@ -4070,7 +4070,7 @@ check_control_units() {
     echo ""
     echo "Dashboard control channel:"
     local unit_dir owner here spool
-    unit_dir="${PITHEAD_UNIT_DIR:-/etc/systemd/system}"
+    unit_dir=$(control_unit_dir)
     if [ "$(normalize_bool "$(env_get DASHBOARD_CONTROL_ENABLED)")" != "true" ]; then
         dr_info "Disabled for this install — the dashboard cannot apply config changes or upgrade."
         return 0
@@ -10864,6 +10864,24 @@ control_run_pending() {
     log "Processed $n control request(s)."
 }
 
+# The directory the dashboard control units live in — ONE rule, shared by the writer and both
+# readers. The appliance's root is read-only by design, so /etc/systemd/system cannot take the unit
+# (#791); /run/systemd/system is a first-class unit path, writable, and cleared every boot, which is
+# fine because these units are derived and the boot path re-renders them. A DIY host keeps /etc.
+# This was two rules until #1151: `provision_control_runner` knew about /run and `doctor` did not,
+# so on a PROVISIONED appliance doctor reported "no runner units are installed" about units that
+# were installed and running. That is the half of the boot health gate that never passed, so the
+# slot never committed — and after #1065 the box reboots a healthy, correctly-updated appliance.
+control_unit_dir() {
+    if [ -n "${PITHEAD_UNIT_DIR:-}" ]; then
+        printf '%s' "$PITHEAD_UNIT_DIR"
+    elif is_appliance; then
+        printf '%s' /run/systemd/system
+    else
+        printf '%s' /etc/systemd/system
+    fi
+}
+
 # Physical directory the installed control units name, or "" when there is no service unit or its
 # ExecStart is unparseable. One checkout has two spellings — the `current` symlink and the versioned
 # dir it points at (production units carry the versioned spelling) — so this resolves to a PHYSICAL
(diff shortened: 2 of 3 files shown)


VijitSingh97:
Merging with **Build + scan the appliance rootfs** red. It is not this change: `develop-v2`'s own tip fails the same check with the same two findings — `CVE-2026-56864` / `CVE-2026-56865` in `golang.org/x/mod`, reached through the `docker-compose` and `cosign` binaries the rootfs compiles from source. The check went red between the 03:15 and 15:16 runs on 2026-08-19 with no commit in between: the CVEs were published, the pins did not move. Four consecutive red runs on the branch since. Filed separately; every other check on this PR is green.

Issue #1146, opened 2026-08-19
The pin watcher never sees the appliance's own two pins: a schedule only runs on the default branch


VijitSingh97:
## What is wrong

`pin-watch.yml` runs on a `schedule:`, and GitHub fires schedules **from the default branch only**.
The default branch is `develop`, which has no `os/`. So the weekly run checks the six product pins
and never sees the appliance's own two:

```
ARG COMPOSE_VERSION=v5.4.0     # os/rootfs/Dockerfile — develop-v2 only
ARG COSIGN_VERSION=v3.1.3
```

The script already handles them — they are wired behind a `[ -f os/rootfs/Dockerfile ]` guard, so a
**manual** run on `develop-v2` reports them correctly (that is how we know `compose` is stale at
`v5.4.0` against upstream `v5.5.0` today). But nothing runs it there on a schedule, and nobody is
going to remember to.

This is a narrower version of exactly the defect #1128 was filed about: the previous watcher lived on
`develop-v2` and therefore never ran at all. This one runs, but with one lane's pins invisible.

## Why it matters more than it looks

Those two are **compiled from source** in the rootfs, and each has a paired `_COMMIT` ARG. They are
the appliance's own supply chain — `cosign` in particular is the verifier the whole signed-update
chain rests on. Nothing else watches them: dependabot's docker ecosystem reads `FROM` lines, and
these are `ARG` values consumed by a build.

## Fix

Add a second job to `pin-watch.yml` that checks out `develop-v2` explicitly:

```yaml
  pin-watch-appliance:
    steps:
      - uses: actions/checkout@...
        with:
          ref: develop-v2
          persist-credentials: false
      - run: bash scripts/pin-watch.sh
```

and give it its own tracking issue title, so the two lanes do not overwrite each other's report.
Roughly four lines plus a title.

One thing to get right: the appliance job must not fail the whole workflow when `develop-v2` is
mid-sync and the script is briefly absent — guard on the file existing, the same way the script
guards on the Dockerfile.

## Tier

Tier 1 cannot see a workflow's scheduling. The verification is `workflow_dispatch` once after
merging and confirming both tracking issues exist — which is the step that caught the equivalent
defect in rigforge (rigforge#376).

VijitSingh97:
## Verified on a real dispatch, not by a green run

[...]

PR #1154, opened 2026-08-20
fix(os): clear the appliance rootfs CVE gate at the source (#1153)

### LINKED ISSUE #1153: The appliance rootfs CVE gate has been red on develop-v2 since 2026-08-19: two HIGH golang.org/x/mod advisories in the compiled cosign and docker-compose
## The appliance lane's CVE gate has been red since 2026-08-19, and nothing raised it

`OS rootfs scan` fails on `develop-v2`'s own tip. Four consecutive red runs, and PRs have been
merged through it in between.

```
usr/local/bin/cosign          (gobinary)  Total: 2 (HIGH: 2, CRITICAL: 0)
usr/local/bin/docker-compose  (gobinary)  Total: 2 (HIGH: 2, CRITICAL: 0)

golang.org/x/mod  CVE-2026-56864  HIGH  fixed  v0.38.0 (cosign)   -> 0.40.0
                                               v0.37.0 (compose)  -> 0.40.0
golang.org/x/mod  CVE-2026-56865  HIGH  fixed  (same two)         -> 0.40.0
```

CVE-2026-56864 — a malicious GOSUMDB could serve arbitrary module content.
CVE-2026-56865 — a malicious GOPROXY could forge sumdb entries.

## What makes it worth its own issue rather than a bump

**No commit broke it.** The scan went from green at the 03:15 run on 2026-08-19 to red at 15:16 on
the same branch, with nothing in between. The pins did not move; the advisories were published. A
scan gate that keys on the outside world will go red on a branch nobody has touched, so "the branch
was green when I based off it" is not evidence about now — and that is exactly how four red runs
accumulated unnoticed.

**Both binaries are compiled from source in the rootfs**, so this is not a tag bump:

```
ARG COMPOSE_VERSION=v5.4.0
ARG COMPOSE_COMMIT=ef61d7410a0c816a71705026e638ec256a591d69
ARG COSIGN_VERSION=v3.1.3
ARG COSIGN_COMMIT=11926fa5bbbbde47e88fc006b625a17769b743b2
```

Each has a paired `_COMMIT` that the build actually checks out, so the version string alone decides
nothing — moving one without the other either fails the checkout or silently mislabels the binary.
This is the trap #1146's report already warns about, now live.

**`compose` was already known stale.** #1146 records it at `v5.4.0` against an upstream `v5.5.0`,
found on a manual run because the weekly watcher's `schedule:` only reads the default branch and
never sees the appliance's own two pins. So the currency gap and the CVE gate are the same gap seen
from two sides: nothing was watching these pins, and the first thing to notice was a scanner.

## What to do

1. Bump `COMPOSE_VERSION`/`COMPOSE_COMMIT` and `COSIGN_VERSION`/`COSIGN_COMMIT` to releases whose
   `go.mod` carries `golang.org/x/mod` >= 0.40.0, moving each version and its commit together.
   Verify by re-running the scan, not by reading the changelog.
2. Confirm the cosign bump keeps the release chain working — the pinned container verifier is what
   every install uses to accept a bundle, and `pithead`'s `COSIGN_IMAGE` carries a *decorative*
   `# v2.6.3` comment that nothing binds to its digest. The two cosign pins already disagree by a
   major version; a bump that moves one and not the other is the failure #1108 exists to prevent.
3. Close the loop that let this sit: #1146 covers the appliance pins never entering the weekly run.
   Whatever fixes that should also make a red `OS rootfs scan` on `develop-v2` visible somewhere an
   operator reads, rather than only on the next PR that happens to touch the rootfs.

## Not verified here

Whether an upstream release of either component exists today with `x/mod >= 0.40.0`. That is the
first thing to check, and if neither has cut one, the decision is between pinning a post-release
commit and accepting the finding with a dated justification — not silently muting the scanner.

VijitSingh97:
Closes #1153.

`OS rootfs scan` has been red on `develop-v2`'s own tip since 2026-08-19 — four runs, with PRs
merged through them. No commit broke it: the branch was untouched between a green run and a red one
and the advisories were published in between.

## Three findings, three different answers

**Raised in the build** — the rootfs compiles both binaries from source, so the module graph is
ours, the same reasoning the patched toolchain above them already uses:

| module | | | |
|---|---|---|---|
| `golang.org/x/mod` | v0.38.0 → **v0.40.0** | CVE-2026-56864, CVE-2026-56865 | both binaries |
| `google.golang.org/grpc` | v1.82.0 → **v1.82.1** | GHSA-hrxh-6v49-42gf | cosign |

Neither is reachable by a pin bump. cosign v3.1.3 *is* the newest release and pins x/mod v0.37.0;
compose v5.4.0 and v5.5.0 both pin v0.38.0; cosign's own main is still v0.38.0.

**Two mutes deleted, not added.** `.trivyignore` already carried CVE-2026-56852 (x/text) and
GHSA-hrxh-6v49-42gf (grpc), on the premise that these modules were vendored inside *downloaded*
release binaries so "no pin bump can clear them". That premise died when the rootfs started
compiling them. Both are fixed at the source now and their entries are gone.

That deletion is load-bearing, not tidiness: **while a mute is present the scan cannot tell "fixed"
from "hidden"**. The first version of this branch verified the grpc raise against grpc's own
still-present mute and proved nothing. Caught in review.

**One accepted, with corrected reasoning** — `github.com/docker/docker` CVE-2026-34040, which
genuinely cannot be raised (see below). It replaces a duplicate: the change had added a second entry
for an ID already accepted at line 24, with a different rationale and a different clearing
condition, so following either comment would have left the other muting.

## Verification — the scan, on the real binaries, twice

Built the gobuild stage on a real machine and scanned the extracted binaries with trivy 0.70.0 and
the gate's own flags.

**With no ignore file at all** (the probative run — nothing can hide behind a mute):

```
cosign          gobinary   0
docker-compose  gobinary   1   <- CVE-2026-34040 only
```

**With the trimmed `.trivyignore`, gate settings, `--exit-code 1`:**

```
cosign          gobinary   0
docker-compose  gobinary   0
TRIVY_EXIT=0
```

`docker-compose version` and `cosign version` both run in the build's own check (exit 0).

## Why CVE-2026-34040 cannot be raised

Trivy names the fix as `29.3.1`. That is a **Moby release, not a version of this module**:

```
go list -m -versions github.com/docker/docker   -> ... v28.5.2+incompatible   (newest)
go list -m -versions github.com/moby/moby       -> ... v28.5.2+incompatible   (newest)
```

Tried it rather than assuming — the build fails with `invalid version: unknown revision v29.3.1`.
Docker 29 lives at a module path compose does not import, so no compose release can carry it
either: v5.4.0, v5.5.0 and main all pin v28.5.2+incompatible. The advisory is also a daemon-side
authz bypass, and compose links the client half.

## Two provenance claims corrected, both found by review

- **A raise moves more than it names.** `go get X@v Y@v` upgrades the whole closure MVS needs to
  seat them — compose moves 1 module, cosign moves **9** (x/mod → x/tools → x/net → x/crypto,
  x/term, x/text, x/sync, x/sys). No formulation avoids that; it is what raising a module in a Go
  build means, so it is written down instead of implied. The reviewer's security framing was
  checked and does not hold: the release-signature verifier is the pinned `ghcr.io/sigstore/cosign`
  **container** (`pithead:542`), not this binary, whose only consumer is an `-x` check in
  `tests/os/verify-image.sh`.
- **`cosign version` now reports `v3.1.3+dirty`.** `go get` edits go.mod, Go stamps `vcs.modified`,
  and release-utils overwrites the `gitVersion` ldflag from the build info — so the ldflag is inert
  and the suffix is truthful. The build was nevertheless still asserting `gitTreeState=clean` about
  a tree Go itself had recorded as modified. It now says `dirty`, so the two agree and both are
  true. Confirmed on the rebuild: `GitVersion: v3.1.3+dirty`, `GitTreeState: dirty`.

Nothing in the repo or the overlay parses either version string — checked.

## Not done

- The local scan covered the two Go binaries, not the whole rootfs image. CI's `OS rootfs scan`
  scans the full image and is the real gate here; this PR's own run is what confirms it.
- No tier-1 test asserts the raises landed, deliberately: the scan is the behavioural gate, and a
  source-text assertion that the ARG contains a version string would be exactly the kind of check
  this repo keeps finding cannot fail.
- The compose currency bump (v5.4.0 → v5.5.0, #1146) is **not** here. It is a separate decision — a
  compose minor is a behaviour change, and it does not move any of these three findings.

### DIFF
--- a/.trivyignore
+++ b/.trivyignore
@@ -13,14 +13,27 @@ CVE-2026-45447
 CVE-2026-23949
 CVE-2026-24049
 
-# Appliance rootfs pinned binaries (docker-compose v5.3.1, cosign v3.1.2 — BOTH already the
-# latest upstream releases, byte-verified): these three sit in Go modules VENDORED inside those
-# binaries, so no pin bump can clear them until upstream cuts releases on the fixed modules.
-# Cleared when pin-watch flags the next compose/cosign releases — bump the pins, drop these.
-# golang.org/x/text norm.Iter infinite loop (both binaries):
-CVE-2026-56852
-# github.com/docker/docker authz bypass, vendored in compose (server-side moby code, and the
-# appliance talks to podman's compat socket, not dockerd — no exposed authz plugin path here):
+# Appliance rootfs Go binaries. This block used to hold three findings on the grounds that the
+# modules were VENDORED inside downloaded release binaries, so "no pin bump can clear them until
+# upstream cuts releases on the fixed modules". That premise is gone: the rootfs COMPILES
+# docker-compose and cosign itself, so the module graph is ours and a finding in it is fixable
+# whether or not upstream has moved. Two of the three are now fixed at the source by the
+# COMPOSE_GO_RAISES / COSIGN_GO_RAISES entries in os/rootfs/Dockerfile and have been deleted from
+# here — golang.org/x/text (CVE-2026-56852) and google.golang.org/grpc (GHSA-hrxh-6v49-42gf).
+# Deleting them is not cosmetic: while a mute is present the scan cannot tell "fixed" from
+# "hidden", so leaving one for a finding you have just fixed destroys the only evidence that the
+# fix landed. Scanned with no ignore file at all after the raises: cosign 0 findings, compose 1.
+#
+# github.com/docker/docker, reached as an indirect dependency of the compiled docker-compose. This
+# one stays, and unlike the other two it genuinely cannot be raised. Trivy names the fix as 29.3.1,
+# but that is a Moby release, not a version of this MODULE — `go list -m -versions` tops out at
+# v28.5.2+incompatible for both github.com/docker/docker and github.com/moby/moby, and the attempt
+# fails the build with "invalid version: unknown revision v29.3.1". Docker 29 lives at a module
+# path compose does not import, so no compose release can carry it either: v5.4.0, v5.5.0 and main
+# all pin v28.5.2+incompatible. The advisory is also a daemon-side authz bypass — compose links the
+# client half, and the appliance talks to podman's compat socket, not dockerd.
+# Clears when compose moves to the docker 29 module path; `go list -m -versions github.com/docker/docker`
+# is the one-line check. The weekly CVE sweep (#833) re-surfaces it regardless (#1153).
 CVE-2026-34040
 # NOTE ON GO STDLIB FINDINGS — there are none listed here on purpose. The appliance rootfs used
 # to download upstream release binaries and inherit whichever Go toolchain built them, so every Go
@@ -28,9 +41,6 @@ CVE-2026-34040
 # COMPILES docker-compose and cosign itself on a pinned, patched toolchain (os/rootfs/Dockerfile,
 # the gobuild stage), so stdlib findings are fixed at the source instead of accepted. If a stdlib
 # CVE appears here again, the answer is to bump the builder image, not to add a line below.
-# google.golang.org/grpc vendored in both pinned binaries (GHSA id — Trivy gates on it like a
-# CVE; same clearing condition as the three above: upstream rebuilds on the fixed module):
-GHSA-hrxh-6v49-42gf
 # util-linux family (bsdutils, libmount1, libuuid1, login, mount, util-linux): the fix is a
 # base-distro point release, 2.41.5-0+deb13u1, and no published python:3.11-slim carries it yet —
 # the newest digest still ships 2.41-5. Same shape as CVE-2026-45447 above: `--ignore-unfixed` hid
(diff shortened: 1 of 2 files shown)


VijitSingh97:
Merging with **Build image (dashboard)** red. It is not this change: PR #1155, which does not touch `.trivyignore` at all, fails the identical check with the identical finding list. CI was green on `develop-v2` at 04:43 today and red at 13:32 with no commit in between — three new util-linux advisories (CVE-2026-53612/53613/53614), the same family as the already-accepted CVE-2026-53615. Filed as #1156, where I also confirmed a base bump cannot fix it: the pinned `python:3.11-slim` digest is already the newest published one.

Every other check is green, including **Build + scan the appliance rootfs** — the gate this PR exists to fix.

Issue #1132, opened 2026-08-19
The appliance's certificate and its Caddy site list are built by different rules, and the certificate is never re-minted


VijitSingh97:
Two functions decide which names the appliance answers to, they are built by different rules, and
one of them is written once and never revisited. So the machine ends up serving addresses its own
certificate does not name — and the failure looks like a browser problem rather than a machine one.

## The two lists

`appliance_mint_cert` builds the certificate's SAN list:

```
alt="DNS:$(hostname).local,DNS:$(hostname),DNS:localhost,IP:127.0.0.1"
for ip in $(hostname -I 2>/dev/null); do alt="$alt,IP:$ip"; done
```

`generate_caddyfile` builds `site_hosts`, which decides what Caddy vhost-matches:

```
local site_hosts="$HOST_IP"
if is_appliance && [ -z "${DASHBOARD_HOST:-}" ]; then
    for extra_ip in $(hostname -I 2>/dev/null) localhost; do ... done
fi
```

They disagree in three ways:

1. `$HOST_IP` is `dashboard.host` when set, and `hostname` otherwise — the cert always uses
   `hostname`. An operator who pins `dashboard.host` gets a certificate minted for a different name.
2. Setting `dashboard.host` **skips the whole expansion**, so the site list collapses to one host
   while the certificate still names the machine's addresses and `localhost`.
3. The `.local` form is in the certificate unconditionally and in the site list only when it happens
   to be what `hostname` returns.

## And the certificate is deliberately never re-minted

The mint is idempotent on purpose, and the reasoning is sound — "an existing certificate the
operator has already trusted is reused, never replaced", because replacing it breaks the browser
that trusted the first one. But `hostname -I` is a **DHCP lease**, and `site_hosts` is rebuilt on
every boot by `render_derived` while the certificate is not. So the moment the box's lease changes,
Caddy starts serving an address the certificate has never heard of.

This is not theoretical. On the physical bench appliance right now, the certificate's IP SANs name
an address the machine no longer has, and it is serving on the address it does have. Its CN and DNS
SANs also do not line up with the name the site list actually answers on — I could not establish
which of the three mechanisms above produced that, because the box is a release variant with no
shell, so treat the mechanism as open and the observation as solid.

## Why it is quiet

An unmatched Host on `:443` does not get an error. Caddy's default for "no site block matched" is an
**empty 200** — `content-length: 0`, no body — so a browser shows a blank page and the operator has
[...]

Issue #852, opened 2026-08-02
Appliance: A/B commit gate is curl-localhost, not the plan's 'pithead doctor --json' (narrower gate + doc mismatch)


VijitSingh97:
## Gap (architecture audit, property 4 — commit gate)

The A/B **commit gate diverges from the plan**, and the shipped gate is materially narrower than the one specified.

### Plan (in three places) says the gate is `pithead doctor --json`
> the commit gate is `pithead doctor --json` — it already exits non-zero only on critical failures and checks containers, Tor, NTP sync, disk, hugepages — `docs/dev/dual-distribution-plan.md:176-182` (also `:601`, `:681`; `docs/dev/appliance-release.md` "gate consumes `doctor --json`"; `#394` "commit gate = `pithead doctor --json`").

### Code gates on a single HTTP probe to the dashboard
`os/overlay/pithead-boot:29-49`:
```
code=$(curl -kso /dev/null -w '%{http_code}' --max-time 5 https://localhost/ ...)
[ "${code:-000}" != "000" ] && rauc status mark-good
```
The deliberate rationale is documented (`os/overlay/pithead-boot:12-15`, `docs/dev/appliance-wizard.md:175`): "any HTTP status is life", proving the derived-config → caddy → dashboard chain. `doctor_json()` exists (`pithead:1136`) but is used **only** by `support-bundle` (`pithead:1174`), never as the gate.

### Why it matters
"Dashboard answers on :443" is a strict subset of "services up and progressing". A slot where p2pool/monerod/xmrig-proxy have crashed but caddy+dashboard serve will **commit** — precisely the healthy-looking-but-wrong slot the plan's `doctor` gate (containers/Tor/hugepages checks) was meant to reject before commit. The narrower gate weakens the case-2 self-heal the A/B design promises.

### Fix direction (pick one, then reconcile the docs either way)
- Preferred: make the commit gate `pithead doctor --json` (exit-code) as the plan specifies, keeping the curl reachability check as one input. Honour the "services up and progressing, never chain-synced" rule (#718 scan-grace) so a days-long initial sync still commits.
- Or: if the curl-only gate is the intended final design (the boot contract argues for it), update `dual-distribution-plan.md:176-182/601/681`, `appliance-release.md`, and `#394` so the stated gate matches the code — per the repo rule that when a doc and the code disagree, the doc is fixed.

[...]

PR #1155, opened 2026-08-20
fix(boot): ask the dashboard's own site, and tell it apart from the default vhost (#1140)

### LINKED ISSUE #1140: The boot health gate's curl half passes on Caddy's empty default vhost when dashboard.host is pinned — and its comment says it cannot
## What is wrong

`os/overlay/pithead-boot`'s health probe dials `https://localhost/` and its comment states why:

```
# localhost, not 127.0.0.1: the bare IP is not in the Caddyfile's site list, so it hits
# Caddy's empty default vhost — an empty 200 that proves only that Caddy runs. localhost IS
# a listed site, so this answer traverses the real dashboard vhost (TLS + auth + proxy).
```

**`localhost` is only a listed site when `dashboard.host` is unset.** In `generate_caddyfile`:

```
local site_hosts="$HOST_IP"
if is_appliance && [ -z "${DASHBOARD_HOST:-}" ]; then
    for extra_ip in $(hostname -I 2>/dev/null) localhost; do
```

`dashboard.host` is a real, supported, editable key (`config.reference.json` → `dashboard.host`,
default `"auto"`), and the renderer itself calls it "a documented, supported choice". Set it, and
`localhost` drops out of the site list — so the probe hits exactly the empty default vhost the
comment says it avoids, gets a `200`, and passes.

## Why it matters

That curl is the fast half of the boot health gate, and it is the half that gates entry to the slow
`doctor` run. After #1065 this gate **reboots the box** when it expires. Its curl half silently
degrades from "the dashboard serves" to "Caddy is running" on any appliance whose operator pinned
`dashboard.host` — and the comment asserting otherwise is what stops anyone noticing.

The `doctor` half still runs, so this is a weakened gate rather than a blind one. But the check does
not do what it says, on the branch where it matters most, and that is this repo's signature defect.

## Fix, in order of laziness

1. Make it true: add `localhost` (and `127.0.0.1`) to the appliance's site list **unconditionally**,
   not only when `dashboard.host` is unset. One vhost entry, and the comment becomes correct.
   Note the interaction with #1132 — the certificate SAN list is built by a different rule and would
   need the same name.
2. Or make the probe honest: assert on something the empty default vhost cannot produce. An empty
   200 has no body and no `WWW-Authenticate`; the dashboard vhost behind auth answers `401`. Gating
   on `[ "$code" = 401 ]` (or on a response header the real vhost sets) distinguishes them, where
   `!= 000` does not.

Option 2 is the smaller diff and catches more — a `200` from the default vhost is precisely the
answer that currently reads as success.

## Tier

Tier 1: render the Caddyfile with `DASHBOARD_HOST` set and assert `localhost` is a listed site (or,
if option 2 is taken, assert the probe's accepted-status set excludes plain `200`).
**Mutation**: unset the new unconditional `localhost` (or widen the accepted status back to
`!= 000`) — the assertion must go red.

VijitSingh97:
Closes #1140.

The boot health gate's fast half dialled `https://localhost/` and accepted any status but `000`. Its
comment said localhost is a listed site, so the answer traverses the real dashboard vhost. That
holds only while `dashboard.host` is unset — `generate_caddyfile` adds localhost inside
`is_appliance && [ -z "$DASHBOARD_HOST" ]`. Pin the host, which is documented and supported, and
localhost drops out: the probe reaches Caddy's empty default vhost, which answers **200 with no
body**, and "Caddy is running" passes as "the dashboard serves" — on the gate that decides whether
an A/B update lives, and that #1065 reboots on.

## Both halves change together, and that is the point

Tightening the acceptance alone would convert the false pass into a permanent **failure** on exactly
the machines this issue is about: with the host pinned there is genuinely nothing at localhost to
believe. That is the same false-RED shape as #1151, on the same gate. The issue's own preferred
option had this in it — see [the design comment](https://github.com/p2pool-starter-stack/pithead/issues/1140#issuecomment-5351477616).

- **Ask the right site.** `gate_url` reads `HOST_IP`, `DASHBOARD_SECURE` and `HOST_PORT` from the
  `.env` this boot has just rendered. `HOST_IP` is what the render uses as its first site address,
  so the probe and the site list cannot disagree — one rule, not a second copy of the expansion.
- **Recognise the right answer.** `gate_answer_is_dashboard` accepts a body, or the `401` a locked
  dashboard sends, and rejects the default vhost's empty 200. Deliberately **not** a 401 check: an
  empty dashboard password is the documented default (`# OPT-IN — an empty password (the default)
  leaves it open`), and that machine answers 200, so a 401 gate would fail forever on every
  no-login appliance.
- **Dial it correctly.** `dashboard.host` is validated as "a hostname or IP address" and explicitly
  allows colons, so `HOST_IP` can be an IPv6 literal. Measured against curl 8.7 rather than reasoned
  about: curl refuses one in `--resolve` outright — `Couldn't parse CURLOPT_RESOLVE entry` — and an
  unbracketed literal in the URL reads the port as part of the address. Either way the request
  fails and the gate never passes. So `gate_target_url` brackets a literal, and `gate_resolve_spec`
  emits `--resolve` only for names, where it is what keeps the dial on loopback instead of making a
  boot-time health probe hostage to the box resolving its own mDNS name.
- `code`/`size` are cleared each iteration, so a curl that prints nothing cannot leave the previous
  round's answer standing.

## The review found a gate that lies, inside this change

The first version replaced the old single-line regex with two independent greps. The doctor half,
`grep -qE 'pithead doctor --json'`, matched **this file's own header comment on line 15** — so the
one assertion whose job is "both signals gate the same mark-good" stayed green with the doctor call
deleted from the commit condition outright. That is #852 restored, with the test that exists to
prevent it still passing.

Fixed by moving the pairing itself above the sourceable boundary as `gate_ready`, driven at tier 1
with a stubbed `pithead`. `BOOT_DOCTOR_JSON` becomes a plain variable for the same reason
`BOOT_FAIL_COUNT` is one — the suite points it somewhere writable. All five helpers now live above
the boundary, so they are driven rather than asserted about.

## Tier 1

`bash tests/stack/run.sh` → **2722 passed, 0 failed** (base 2700).

## Mutations — eight named, eight run

| # | mutation | result |
|---|---|---|
| M1 | acceptance widened back to "any status but 000" | 2 red |
| M2 | probe target hard-coded back to `localhost` | 4 red |
| M3 | the fail-safe HOST_IP fallback removed | 1 red |
| M4 | acceptance narrowed to **401-only** (the fix the issue first recommended) | 2 red |
| M5 | IPv6 bracketing dropped from `gate_target_url` | 1 red |
| M6 | `--resolve` emitted for IP literals too | 2 red |
| M7 | **the doctor half deleted from the commit gate** — the #852 regression the old grep could not see | 1 red |
| M8 | the probe half widened back inside `gate_ready` | 1 red |

M4 is worth calling out: the suite goes red on the fix originally recommended, because the no-login
case is now encoded as a requirement.

## Not done

- The `curl` invocation itself is still below the sourceable boundary and has no tier-1 coverage —
  only the values it is built from and the decision it feeds. The honest place for the rest is
  leg 4 of the battery.
- **No `--phase update` battery run on this branch.** This changes `pithead-boot`, which the bench
  is the real gate for, and that has not been run here.
- The render is untouched: `localhost` still leaves the site list when `dashboard.host` is pinned.
  This makes the probe honest about that rather than changing what the box serves — the vhost
  question belongs with #1132, where the certificate SAN list is decided by a different rule.

### DIFF
--- a/os/KNOWN-ISSUES.md
+++ b/os/KNOWN-ISSUES.md
@@ -183,6 +183,19 @@ starts. Host identity survives A/B updates and reboots by construction (it never
 system slot to begin with); the reset tiers classify it like everything else on `/data` — a
 keep-everything reinstall keeps it, a fresh-start or full wipe deletes it.
 
+**Fixed — the boot health gate's other half was accepting Caddy's empty default vhost
+(#1140).** The probe dialled `https://localhost/` and took any status but `000` as proof the
+dashboard was serving, on the stated belief that localhost is always a listed site. It is one only
+while `dashboard.host` is unset: `generate_caddyfile` adds it inside `is_appliance && [ -z
+"$DASHBOARD_HOST" ]`. Pin the host — a documented, supported choice — and the probe reached the
+default vhost, which answers 200 with no body, so "Caddy is running" passed as "the dashboard
+serves" on the gate that decides whether an A/B update lives. Both halves are fixed together,
+because tightening the check alone would have turned the false pass into a permanent failure on
+exactly those machines: the probe now asks for the site the render itself chose, read as `HOST_IP`
+from the `.env` of that same boot and dialled over loopback with `--resolve`, and it accepts only an
+answer the default vhost cannot give — a body, or the `401` a locked dashboard sends. Not a `401`
+check: an empty dashboard password is the documented default, and that machine answers `200`.
+
 **Fixed — `doctor` looked for the control units in `/etc`, and the appliance puts them in
 `/run` (#1151).** `provision_control_runner` knew the read-only root cannot take that write (#791);
 `check_control_units` and `control_units_owner_dir` did not, and each kept its own `/etc` default.
--- a/os/overlay/pithead-boot
+++ b/os/overlay/pithead-boot
@@ -38,6 +38,10 @@ set -uo pipefail
 # which is the whole point of bounding. If it cannot be written, this refuses to reboot at all —
 # an unbounded reboot loop is the one outcome worse than a stranded box.
 BOOT_FAIL_COUNT=.boot-gate-failures
+# Where doctor's verdicts land while this gate is looping: /run, so a gate that never passes leaves
+# the last failing run's exact blocking check behind without wearing /data. A plain variable for the
+# same reason BOOT_FAIL_COUNT is one — the suite points it somewhere writable.
+BOOT_DOCTOR_JSON=/run/pithead-boot-doctor.json
 boot_gate_passed() { rm -f "$BOOT_FAIL_COUNT"; }
 fail_boot() { # <what went wrong>
     echo "pithead-boot: $1 — slot left uncommitted so A/B fallback stays armed" >&2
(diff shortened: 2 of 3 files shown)


VijitSingh97:
Merging with **Build image (dashboard)** red. Pre-existing and unrelated — #1156. Three util-linux advisories published between 04:43 and 13:32 today, on a branch nobody touched; the pinned `python:3.11-slim` digest is already the newest published one, so there is nothing to bump to.

**Build + scan the appliance rootfs is now green** on this PR, which it was not before the rebase — that is #1154 landing.

PR #1057, opened 2026-08-16
[DO NOT MERGE — review] develop-v2: the Pithead OS appliance distribution

VijitSingh97:
**This PR is for review, not for merging on green.** It carries the whole appliance distribution — 407 commits, 137 files, two thirds of it `os/` — and the operator merges it after reading, not CI.

> **Do not merge while #1062 is open.** The analysis below found a confirmed data-loss defect: the appliance reformats a `/data` partition that a standard `e2fsck` recovers intact, destroying wallets, Tor onion keys and chains, and returning the box looking factory-fresh. Reproduced end to end.

This PR raises `develop-v2` against `develop`. It carries the whole appliance distribution as one branch: the bootable image, the A/B update mechanism, the installer, the first-boot wizard, the USB media config channel and factory reset. It is a review artifact, not a merge-on-green change. CI covers the parts of the stack CI can reach, and most of the appliance surface is not one of them — the evidence that matters here comes from the KVM battery and from a manual hardware battery that has not been run yet. Read the gate list below and merge it yourself when the conditions are met.

The findings in this body come from two of the review passes, on boot/identity/storage and on the update chain. They are not the complete gap list for the branch.

## What v2 adds

v2 turns the compose stack into a machine you flash. `os/` builds a bootable image with a read-only root, an explicit `/var` overlay whose upper lives on `/data`, and two system partitions. Updates arrive as signed RAUC bundles: the image is written to the spare slot, GRUB tries it once, and `pithead-boot` commits the slot only after the stack answers a health check — a slot that never boots reverts on its own at the next power cycle. Everything that must survive an update lives on a third partition, `/data`: config, chains, wallets, Tor onion keys, the dashboard database, the machine-id and the SSH host key.

A box with no config serves the first-boot wizard. `pithead-install` copies the running image onto an internal disk with a keep / wipe-data / wipe-all choice. A FAT stick carrying a config file configures a headless box with no network step. `pithead factory-reset` arms a marker on the ESP and reinitializes `/data` on the next boot.

## What is proven, and how

Four evidence tiers, and they are not interchangeable.

[...]

Issue #833, opened 2026-08-01
Appliance update strategy: staying on top of Debian base and package updates


VijitSingh97:
## Problem / motivation

The appliance is an immutable Debian 13 box: read-only root, updates delivered only as signed A/B bundles. There is no apt on the running system and `unattended-upgrades` is architecturally off the table — **every Debian security fix reaches users only through a freshly baked image**. That makes bake cadence the patch cadence, and right now nothing watches the appliance's supply chain:

- `os/rootfs/Containerfile` (on `feat/phase2-bakery-image`) starts `FROM debian:trixie-slim` — **not digest-pinned**, unlike every `build/*` Dockerfile (#135), and **not covered by `.github/dependabot.yml`**, whose `docker` ecosystem lists only the `build/*` directories.
- The baked rootfs is **never CVE-scanned**. CI's Trivy gate (#282) covers the five stack images but not the appliance root filesystem, which additionally carries podman/netavark/systemd and friends from the Debian archive.
- Scanning is PR/push-triggered only — no scheduled run. A CVE disclosed during a quiet week (especially one fixed in the apt archive but not yet in a republished base digest, so no Dependabot PR fires) surfaces only when the next unrelated PR goes red.
- The rootfs hand-pins binaries Dependabot has no ecosystem for: `COMPOSE_VERSION`, `COSIGN_VERSION`, the RigForge ref (#821), and xmrig (built from RigForge's pin).
- The container images baked into the wizard image inherit the DIY channel's update story — but the compose-pinned third-party images (the two Tari images, `caddy`, `tecnativa/docker-socket-proxy`) are digest-pinned in `docker-compose.yml` and **also outside Dependabot's directories**, and are never scanned.

The DIY channel already has a working loop: digest pins → weekly Dependabot bumps → Trivy gate with `.trivyignore` rationale → release. The appliance needs the equivalent loop, plus a cadence policy, before phase 2 ships.

## Proposed plan

1. **Pin and track the appliance base.** Digest-pin `debian:trixie-slim` in `os/rootfs/Containerfile` and add its directory to the `docker` ecosystem in `dependabot.yml`, under the same patch/digest-only policy as the stack images (major/minor bumps stay deliberate migrations — the Debian 14 rebase is already planned in the support policy). Note Dependabot reads manifests on the default branch, so this coverage becomes live when the bakery tree merges; add a `target-branch` entry if it needs to bite earlier.
[...]

PR #1160, opened 2026-08-20
fix(ci): stop the CVE gate scanning yesterday's advisory database (#1159)

### LINKED ISSUE #1159: The CVE gate scans against a per-branch database snapshot frozen at midnight, so the same image passes and fails 40 seconds apart
The CVE gate can pass and fail on the identical image 40 seconds apart, because `trivy-action` caches the vulnerability database under a key with one-day granularity while GitHub scopes caches per ref. The first CI run of the calendar day on a given ref freezes that ref's vulnerability database for the rest of the day. Which snapshot a scan gets is a function of when that branch last started a run — not of when the scan runs.

The consequence is not "the gate is sometimes late". It is that a long-lived branch can stay green all day against advisories every other branch is already red on, and nobody can tell whether a red check means a real finding or a fresher database.

## The measurement

Two runs of `Build image (dashboard)` on 2026-08-20, 40 seconds apart:

| | run | started | debian findings | verdict |
|---|---|---|---|---|
| A | `32376011956` (PR #1057) | 13:44:44Z | 0 | pass |
| B | `32376038312` (PR #1155) | 13:45:24Z | 27 HIGH | fail |

The inputs are byte-identical. Same git trees for all three things that can change the result:

```
$ git rev-parse 28241e13:build/dashboard 64a8be4b:build/dashboard
bfb38aac66f646e1e4f35a64c24f349b82d2bcaa
bfb38aac66f646e1e4f35a64c24f349b82d2bcaa
$ git rev-parse 28241e13:.trivyignore 64a8be4b:.trivyignore
2ac1259c0ed8d1cecfa8aeccfc5295eff86ed561
2ac1259c0ed8d1cecfa8aeccfc5295eff86ed561
$ git rev-parse 28241e13:.github/workflows/ci.yml 64a8be4b:.github/workflows/ci.yml
e94e9c88727dd3837d625e4cc65f45a27d612262
e94e9c88727dd3837d625e4cc65f45a27d612262
```

Both scans report the same image to the same depth — `Detected OS family="debian" version="13.6"`, `pkg_num=101`. Neither log contains a database download; both restored one from cache. Run A reports `0` for `pithead-dashboard:ci (debian 13.6)`; run B reports `Total: 27 (HIGH: 27, CRITICAL: 0)`, every one of them `CVE-2026-53612`/`53613`/`53614` against `bsdutils`, `libblkid1`, `liblastlog2-2` and the rest of the util-linux family. Neither of those three ids was in `.trivyignore` at either commit.

## Why the two runs got different databases

`aquasecurity/trivy-action` restores the database like this ([`action.yaml` at the pinned SHA](https://github.com/aquasecurity/trivy-action/blob/ed142fd0673e97e23eac54620cfb913e5ce36c25/action.yaml#L135-L146)):

```yaml
- name: Get current date
  id: date
  run: echo "date=$(date +'%Y-%m-%d')" >> $GITHUB_OUTPUT

- name: Restore DB from cache
  if: ${{ inputs.cache == 'true' }}
  uses: actions/cache@...
  with:
    path: ${{ inputs.cache-dir }}
    key: cache-trivy-${{ steps.date.outputs.date }}
    restore-keys: cache-trivy-
```

Actions caches are immutable once written and scoped per ref, so one key name is many independent entries. On 2026-08-20 the repo held five of them:

```
$ gh api repos/.../actions/caches --jq '...'
cache-trivy-2026-08-20  ref=refs/pull/1057/merge  created=2026-08-20T00:15:58Z  size=79941422
cache-trivy-2026-08-20  ref=refs/pull/1152/merge  created=2026-08-20T04:33:16Z  size=996869443
cache-trivy-2026-08-20  ref=refs/pull/1154/merge  created=2026-08-20T13:32:52Z  size=996829593
cache-trivy-2026-08-20  ref=refs/pull/1155/merge  created=2026-08-20T13:33:48Z  size=996829518
cache-trivy-2026-08-20  ref=refs/heads/develop    created=2026-08-20T15:36:59Z  size=996866001
```

Run A restored the entry written at **00:15:58Z**, hours before the three util-linux advisories were published. Run B restored the entry written at **13:33:48Z**, after them. The key matched exactly in both cases, so neither run re-downloaded anything and neither log says a word about it.

That entry is also why PR #1057 was green all day and not just at 13:44: its `cache-trivy-2026-08-20` was created at 00:15:58Z and last read at 16:01:25Z. Every run on that ref for sixteen hours scanned against a database from a quarter past midnight. #1057 is the branch that carries the whole appliance distribution.

`restore-keys: cache-trivy-` makes the tail worse rather than better. A ref with no entry for today falls back to the newest `cache-trivy-*` it can see, which may be days old, and `trivy` then re-downloads only if the restored database's own `NextUpdate` has passed.

## Second consequence: the cache budget is gone

```
active_caches=56  total_bytes=8112959593
trivy entries: count=25  bytes=7483481927
```

7.48 GB of the repo's 8.11 GB of Actions cache — against GitHub's 10 GB limit — is near-duplicate copies of one vulnerability database, six of them at ~997 MB. Everything else in the repo is competing for the remaining 2 GB, and GitHub evicts least-recently-used entries once the limit is hit, which is one more input the gate's result depends on.

(Unexplained and separate: these entries were ~80 MB until 2026-08-19 and ~997 MB after. Worth a look, but it is not what causes the flap.)

## Proposed fix

Pass `cache: false` to both `trivy-action` call sites — `build-images` in `ci.yml`, and the rootfs scan in `os-rootfs.yml` on the appliance branch. Every scan then downloads a current database and every scan at a given moment agrees.

It is close to free. Run B spent 13.9s restoring its ~997 MB cache; a database download is ~60 MB. The trivy binary stops being cached too, which costs a few seconds per job against 7.5 GB of budget returned.

Expect the first run after the change to surface whatever the stale snapshots were hiding. That is the point, but it should be read before merging rather than after.

## Why this matters beyond one flap

A scan gate keys on the outside world, so it is the one gate that cannot be made reproducible by pinning inputs — and this repo has spent several weeks on gates that report success without having run. This one reports success against a question it asked yesterday. The failure mode it trains is the expensive one: three checks went red on 2026-08-20 with no commit to blame, and the first instinct each time was to doubt the finding.

VijitSingh97:
Closes #1159.

`trivy-action` caches the vulnerability database under `cache-trivy-$(date +%F)` with
`restore-keys: cache-trivy-`. Actions caches are immutable once written and scoped per ref, so that
one key name is many independent entries, and the first run of the calendar day on a branch freezes
that branch's database until midnight. Which snapshot a scan gets is a function of when its branch
last started a run, not of when the scan runs.

`cache: "false"` makes every scan download a current database, so two scans at the same moment
agree.

## What was run

**The flap, reproduced from the archived runs rather than from reasoning.** Two `Build image
(dashboard)` jobs 40 seconds apart on 2026-08-20:

| | run / job | started | debian findings | verdict |
|---|---|---|---|---|
| A | `32376011956` / `96447462559` | 13:44:44Z | 0 | pass |
| B | `32376038312` / `96447547610` | 13:45:24Z | `Total: 27 (HIGH: 27)` | fail |

Both logs pulled with `gh api repos/.../actions/jobs/<id>/logs`. Both report the same image to the
same depth — `Detected OS family="debian" version="13.6"`, `pkg_num=101`. Neither contains a
database download; both hit their cache key exactly.

**The inputs are byte-identical**, checked as trees rather than by reading the diff:

```
$ git rev-parse 28241e13:build/dashboard 64a8be4b:build/dashboard
bfb38aac66f646e1e4f35a64c24f349b82d2bcaa
bfb38aac66f646e1e4f35a64c24f349b82d2bcaa
$ git rev-parse 28241e13:.trivyignore 64a8be4b:.trivyignore
2ac1259c0ed8d1cecfa8aeccfc5295eff86ed561
2ac1259c0ed8d1cecfa8aeccfc5295eff86ed561
$ git rev-parse 28241e13:.github/workflows/ci.yml 64a8be4b:.github/workflows/ci.yml
e94e9c88727dd3837d625e4cc65f45a27d612262
e94e9c88727dd3837d625e4cc65f45a27d612262
```

Neither of `CVE-2026-53612`/`53613`/`53614` was in `.trivyignore` at either commit, so the pass is
not an acceptance.

**The cache entries explain which database each got**, read from the API:

```
cache-trivy-2026-08-20  ref=refs/pull/1057/merge  created=2026-08-20T00:15:58Z
cache-trivy-2026-08-20  ref=refs/pull/1155/merge  created=2026-08-20T13:33:48Z
```

Run A restored the 00:15:58Z entry — written hours before the three util-linux advisories were
published. Run B restored the 13:33:48Z entry, written after them. That entry is also why PR #1057
was green all day and not only at 13:44: created 00:15:58Z, last read 16:01:25Z.

**The cache budget, measured:** `active_caches=56 total_bytes=8112959593`, of which the 25 trivy
entries are `7483481927` — 92% of the repo's Actions cache, against GitHub's 10 GB limit, in
near-duplicate copies of one database. Six of them are ~997 MB.

**Lint:** `uvx yamllint .github/workflows/ci.yml` clean, `uvx zizmor@1.25.2 --offline
.github/workflows/` reports no findings.

## What was NOT run

- **No test, and none is being added.** `docs/dev/testing-strategy.md` fixes four tiers and none of
  them covers `.github/workflows`; the repo has no assertion over any workflow file today, and the
  existing lint tier for them is the `zizmor` job, which passes. Inventing a fifth tier for one YAML
  value would be worse than the comment that now sits above the line. If `cache: "false"` is ever
  deleted, nothing catches it — that is stated rather than papered over.
- **Not driven on a runner before opening this.** GitHub Actions cannot be run locally, so this PR's
  own five matrix jobs were the first execution. They have since run — see the verification comment
  below, which retires both risks and corrects two figures in this description.

## The two ways this could be wrong, and how to check them on this PR

1. **It may go red.** Every scan now sees a current database, so anything the stale snapshots were
   hiding surfaces on this run. That is the intended behaviour, but a red check here is a finding to
   handle, not a reason to revert.
2. **The database now downloads on every job.** If `trivy`'s registry pull is rate-limited across
   five concurrent matrix jobs, this trades a false green for a flaky red, which would be worse.
   Five clean jobs on this PR is the evidence; it has not been proven any other way. For scale: run
   B spent 13.9s restoring its ~997 MB cache, so the expected cost was near zero or negative.

## What this deliberately does not change

The gate still keys on the outside world, so it will still go red with no commit to blame — three
times on 2026-08-20 alone. That is the gate working. The weekly sweep (#833) is the channel for it.
What changes is that the gate stops disagreeing with itself, which is what made those three
reddenings read as noise.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,16 @@ jobs:
       - name: Scan image for CVEs (Trivy)
         # Gate on actionable (fixable) HIGH/CRITICAL only; accepted findings live in .trivyignore.
         # Must stay green before v1.1 images publish (#282).
+        #
+        # cache: false is load-bearing (#1159). The action caches the vulnerability database under
+        # `cache-trivy-$(date +%F)`, and Actions caches are immutable and scoped per ref — so the
+        # first run of the calendar day on a branch freezes that branch's database until midnight.
+        # On 2026-08-20 the identical dashboard image passed on one branch and failed on another 40
+        # seconds apart, because one restored a snapshot written at 00:15 and the other one written
+        # at 13:33. A gate whose answer depends on when its branch last started a run is not a gate.
+        # Measured cost of downloading it every time: 108.49 MiB from mirror.gcr.io/aquasec/trivy-db
+        # in 3.3s, against the 13.9s the ~997 MB cache restore was taking. It also returns ~7.5 GB of
+        # the repo's 10 GB Actions cache budget, which was 92% near-duplicate copies of this one DB.
         uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
         with:
           image-ref: pithead-${{ matrix.service }}:ci
@@ -117,6 +127,7 @@ jobs:
           ignore-unfixed: true
           exit-code: "1"
           trivyignores: .trivyignore
+          cache: "false"
 
   hadolint:
     name: Dockerfile lint (hadolint)


VijitSingh97:
## Adversarial review + post-run verification

### The fix was verified behaviourally, not by "the checks are green"

A green run proves nothing here on its own — a `cache: "false"` that the action ignored would look
identical to one it honoured. From the dashboard job's log on this branch:

```
2026-08-20T22:25:23Z INFO [vulndb] Need to update DB
2026-08-20T22:25:23Z INFO [vulndb] Downloading vulnerability DB...
2026-08-20T22:25:23Z INFO [vulndb] Downloading artifact...  repo="mirror.gcr.io/aquasec/trivy-db:2"
                     108.49 MiB / 108.49 MiB [====] 100.00% 35.53 MiB p/s 3.3s
2026-08-20T22:25:27Z INFO [vulndb] Artifact successfully downloaded
2026-08-20T22:25:36Z INFO Detected OS  family="debian" version="13.6"
2026-08-20T22:25:36Z INFO [debian] Detecting vulnerabilities...  os_version="13" pkg_num=101
```

`Need to update DB` cannot appear if a cache had been restored. Same image as the two runs in the
description — `debian 13.6`, `pkg_num=101` — and 0 debian findings, which is now correct rather than
lucky: #1157 accepted the three util-linux ids at 15:36.

### Both stated risks are retired by that run

1. **"It may go red."** It did not. All five matrix images pass against a database downloaded
   minutes ago, so nothing was being hidden by the stale snapshots beyond what #1157 already
   accepted.
2. **"The database now downloads on every job."** The source is `mirror.gcr.io/aquasec/trivy-db:2`
   — Google's registry mirror, not GHCR and not GitHub Releases — and five concurrent jobs pulled it
   without incident. Cost is 108.49 MiB in 3.3s per job, against 13.9s to restore the ~997 MB cache.
   Faster, not slower.

The description said `~60 MB`; the real figure is 108.49 MiB. Both the description and the in-file
comment now carry the measured number.

### One review finding, raised and refuted

A review pass called the root cause wrong: both jobs hit the same key `cache-trivy-2026-08-20` yet
restored ~76 MB and ~951 MB respectively, which it read as proof that caches are **not** scoped per
ref.

That inverts the evidence. If caches were not scoped per ref there would be exactly one entry for
that key and both jobs would restore identical bytes. Two different payloads under one key name is
what per-ref scoping *predicts*, and the API says so directly — the entries carry a `ref` field:

```
cache-trivy-2026-08-20  ref=refs/pull/1057/merge  created=00:15:58Z  size=79941422   (76.2 MiB)
cache-trivy-2026-08-20  ref=refs/pull/1155/merge  created=13:33:48Z  size=996829518  (950.7 MiB)
```

Those are the review's own two numbers, to the megabyte. `restore-keys` does allow a fallback, but
to the **default branch's** caches, not to another PR's — and neither run used it: both logs say
`Cache hit for: cache-trivy-2026-08-20`, a primary-key hit.

The same pass also proposed pinning the database version instead. That is the disease, not the cure:
a CVE gate reading a pinned advisory database stops learning about new advisories entirely, which is
a permanent version of the bug being fixed here.

### What is still not covered

Nothing asserts `cache: "false"` stays. `docs/dev/testing-strategy.md` fixes four tiers and none
covers `.github/workflows`; the `zizmor` job is the only tier these files have and it does not check
inputs. The load-bearing comment above the line is the whole guard, and it is named as such.

Issue #1162, opened 2026-08-20
The appliance rootfs CVE sweep has never run: its schedule lives on a branch a schedule cannot fire from


VijitSingh97:
`os-rootfs.yml` declares a weekly sweep:

```yaml
  schedule:
    - cron: "0 6 * * 1" # Mondays 06:00 UTC, after ci.yml's 05:00 image sweep
```

It has never fired. GitHub runs `schedule:` from the **default branch only**, the default branch is
`develop`, and this workflow exists on `develop-v2` alone — the one workflow file that is not on both.

```
$ git ls-tree --name-only origin/develop .github/workflows/ | grep -c os-rootfs
0

$ gh run list --workflow=os-rootfs.yml --limit 200 --json event \
    --jq '[.[].event] | group_by(.) | map({event: .[0], n: length})'
[{"event":"pull_request","n":96}]
```

96 runs since 2026-08-02, every one of them a `pull_request`. Zero scheduled, zero dispatched.

## What that costs

The workflow's own header says why the sweep matters:

> The appliance's Debian userland — podman, netavark, openssl, the kernel — is frozen at bake time
> and has no apt at runtime, so this scan is the eye on the base OS between releases.

There is no eye between releases. The scan only runs on a PR touching `os/**`,
`.github/workflows/os-rootfs.yml` or `.trivyignore`, so the appliance's frozen userland is unscanned
for as long as nobody happens to open one — and a CVE gate keys on the outside world, so "nobody
changed anything" is exactly when a new advisory lands unnoticed. #1153 was found because a PR
happened to touch `os/` that week. Nothing structural produced that.

`cosign`, `podman`, `netavark`, `openssl` and the kernel are the appliance's whole trusted base, and
this is the only gate watching them.

## Why it reads as covered

The workflow contains a step written specifically for the scheduled default-branch run:

```yaml
      - name: Check the appliance tree exists
        # Scheduled runs use the default branch; until the appliance tree merges there, skip
        # quietly instead of going red every Monday for a directory that cannot exist yet.
```

Defensive code for a case that cannot arise, which is what makes a reader conclude the schedule is
handled. It is the same defect as #1048, #1064 and #1146 — a check that reports success without
having run — and it is the third instance found in this repo in a week.

## The trap in the obvious fix

Copying the file to `develop` unchanged makes it **worse**, not better. `develop` has no `os/`, so
[...]

PR #1164, opened 2026-08-21
Sync develop into develop-v2 (#1159, #1146) — and the appliance's own half of the CVE-gate fix

VijitSingh97:
Carries #1159 (PR #1160) and #1146 (PR #1161) to the appliance lane, plus the appliance's own half of
#1159 as a separate commit on top.

Two commits, deliberately not one: the merge is a **pure sync** and carries nothing of its own, so
`git diff develop-v2..HEAD^` is exactly `develop`'s two changes. The appliance-only edit is its own
reviewable commit rather than hidden inside the merge.

## The conflict, and how it was resolved

One file conflicted — `scripts/pin-watch.sh`, two hunks — and it conflicted **by design**. `develop-v2`
had already split the report's final conditional into an `if/else` to carry a real warning: a
compose/cosign bump is two ARGs, and moving the version without its paired `_COMMIT` silently builds
the old code. #1161 deliberately adopted that same `if/else` shape on `develop` rather than rewriting
it, so `develop`'s file is now a strict **superset** and both hunks resolve the same way.

Resolved hunk by hunk, not with `--ours`/`--theirs` on the file, then asserted:

```
develop-v2's _COMMIT warning survives                      OK (1)
the lane variable survives                                 OK (1)
corrected else-arm text present                            OK (1)
old false sentence GONE                                    OK (0)
both if-arms present (components block + report block)     OK (2)
```

and then the check that proves the superset claim rather than assuming it:

```
$ git show origin/develop:scripts/pin-watch.sh > /tmp/dev-pw.sh
$ diff -q /tmp/dev-pw.sh scripts/pin-watch.sh
  (identical)
```

## The appliance half of #1159

`os-rootfs.yml` is the repo's **second** `trivy-action` call site; #1160 only fixed `ci.yml`. Same
defect: the action keys its vulnerability database on the calendar day, Actions caches are immutable
and scoped per ref, so the first run of the day on a branch froze that branch's database until
midnight.

This is the lane where a stale answer costs most — the appliance's Debian userland is baked and has
no apt at runtime, so this scan is the only eye on podman, netavark, openssl and the kernel between
releases.

**And that eye is shut for a different reason, tracked as #1162:** this workflow's weekly sweep has
never fired once, because a `schedule:` runs from the default branch and this file exists only on
`develop-v2`. 96 runs, all `pull_request`. This PR makes the scan honest when it runs; #1162 is what
makes it run.

## What was run

**On this merged tree, not on `develop`:**

- `bash scripts/pin-watch.sh` — the **appliance arm** fires here, because this branch actually has
  `os/rootfs/Dockerfile`. First line reads `Upstream currency for the product stack and the appliance
  rootfs`, and the `_COMMIT` warning prints. That is the assertion that the conflict resolution did
  not quietly drop it.
- `bash scripts/pin-watch.sh --self-test` — 7 ok.
- `make lint-sh` rc=0 (~4 min on this tree), `uvx yamllint` clean on both changed workflows,
  `uvx zizmor@1.25.2 --offline .github/workflows/` no findings.
- Content assertions on the merged tree rather than on the merge succeeding: `.trivyignore` keeps the
  appliance-only `CVE-2026-34040` and still carries exactly 8 active ids, `os/rootfs/Dockerfile` is
  intact, both workflows carry `cache: "false"`.
- `git rev-list --count HEAD..origin/develop` → **0**: twins level once this lands.

## What was NOT run

- **No bench evidence, and none is owed** — nothing here touches `os/` runtime behaviour; the change
  under `os/` is `.github/workflows/os-rootfs.yml`, which is CI configuration.
- **The appliance scan has not been observed downloading a fresh database.** `ci.yml`'s equivalent was
  verified that way on #1160 (`[vulndb] Need to update DB`, 108.49 MiB in 3.3s), and this is the
  identical input on the identical pinned action — but it is inference here, not observation. The
  `OS rootfs scan` job on this PR is the first execution; its log is the evidence.
- **No test.** `docs/dev/testing-strategy.md` fixes four tiers and none covers `.github/workflows`;
  `zizmor` is the only lint over them and it does not check inputs. The load-bearing comments are the
  guard, and they say so.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,16 @@ jobs:
       - name: Scan image for CVEs (Trivy)
         # Gate on actionable (fixable) HIGH/CRITICAL only; accepted findings live in .trivyignore.
         # Must stay green before v1.1 images publish (#282).
+        #
+        # cache: false is load-bearing (#1159). The action caches the vulnerability database under
+        # `cache-trivy-$(date +%F)`, and Actions caches are immutable and scoped per ref — so the
+        # first run of the calendar day on a branch freezes that branch's database until midnight.
+        # On 2026-08-20 the identical dashboard image passed on one branch and failed on another 40
+        # seconds apart, because one restored a snapshot written at 00:15 and the other one written
+        # at 13:33. A gate whose answer depends on when its branch last started a run is not a gate.
+        # Measured cost of downloading it every time: 108.49 MiB from mirror.gcr.io/aquasec/trivy-db
+        # in 3.3s, against the 13.9s the ~997 MB cache restore was taking. It also returns ~7.5 GB of
+        # the repo's 10 GB Actions cache budget, which was 92% near-duplicate copies of this one DB.
         uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
         with:
           image-ref: pithead-${{ matrix.service }}:ci
@@ -117,6 +127,7 @@ jobs:
           ignore-unfixed: true
           exit-code: "1"
           trivyignores: .trivyignore
+          cache: "false"
 
   hadolint:
     name: Dockerfile lint (hadolint)
--- a/.github/workflows/os-rootfs.yml
+++ b/.github/workflows/os-rootfs.yml
@@ -60,6 +60,11 @@ jobs:
         if: steps.tree.outputs.present == 'true'
         # Same gate as ci.yml's image scan: actionable (fixable) HIGH/CRITICAL only; accepted
         # findings live in .trivyignore with a rationale and how they clear.
+        #
+        # cache: false for the reason spelled out over ci.yml's scan (#1159) — this is the repo's
+        # other trivy call site. It matters more here: the appliance's Debian userland is baked and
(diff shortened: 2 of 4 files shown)


VijitSingh97:
## The one thing this PR said was inference is now observed

The `Build + scan the appliance rootfs` job on this PR is the first execution of `cache: "false"` on the second call site, and it behaves the same way `ci.yml`'s did:

```
[vulndb] Need to update DB
[vulndb] Downloading artifact...  repo="mirror.gcr.io/aquasec/trivy-db:2"
[vulndb] Artifact successfully downloaded
Detected OS  family="debian" version="13.6"

pithead-os-rootfs:ci (debian 13.6)  │  debian  │  0
```

`Need to update DB` cannot appear if a cache had been restored, so the input took effect rather than being ignored.

**The result is worth reading on its own, not just as proof the flag works.** The appliance rootfs is at **0 fixable HIGH/CRITICAL findings against a database fetched minutes ago** — not against whatever snapshot this branch happened to pull that morning. Given #1153 (x/mod, HIGH, found only because a PR happened to touch `os/`), that is the first time this lane's answer has been known to be current.

It is still only current on PRs that touch `os/**`. #1162 is what makes it current weekly.

All 16 checks pass. Merging.

Issue #1163, opened 2026-08-20
Dependabot has never checked the appliance rootfs base image: its entry lives on a branch Dependabot does not read


VijitSingh97:
`develop-v2` adds `/os/rootfs` to the docker ecosystem in `.github/dependabot.yml`:

```yaml
  - package-ecosystem: "docker"
    directories:
      - "/build/dashboard"
      …
      - "/os/rootfs"
```

Dependabot reads `.github/dependabot.yml` from the **default branch only**, and that entry exists
only on `develop-v2`. It has never produced anything:

```
$ gh pr list --state all --author app/dependabot --limit 100 --json baseRefName \
    --jq '[.[].baseRefName] | group_by(.) | map({base: .[0], n: length})'
[{"base":"develop","n":25}]

$ gh pr list --state all --author app/dependabot --limit 100 --json title \
    --jq '.[] | select(.title|test("rootfs|os/";"i"))'
(no output)
```

25 dependabot PRs, every one based on `develop`, none touching `os/rootfs`.

## Why this one matters more than a missed bump

The appliance's Debian userland has no apt at runtime. A base digest bump plus a rebake **is** its
patch channel — there is no other way for a fixed openssl or kernel package to reach a flashed box
between releases. The config file says exactly this:

> For the appliance the stakes are higher: its Debian userland has no apt at runtime, so the digest
> bump + rebake IS its patch channel.

## The compensating control named in the same comment does not run either

The file goes on:

> Dependabot reads this file on the default branch only, so os/rootfs coverage starts when the
> appliance tree merges; until then os-rootfs.yml's scan on os/** PRs is the eye on it.

That is honest about this gap and points at `os-rootfs.yml` as the fallback. #1162 shows that
workflow's weekly sweep has never fired either — 96 runs, all `pull_request`, because a `schedule:`
fires from the default branch and that file is also `develop-v2`-only. So the appliance base image
has two watchers and neither one has ever run. Nothing at all is watching it between `os/**` PRs.

## Fix

`target-branch` is the mechanism Dependabot provides for exactly this. Add a second docker entry to
**`develop`**'s `dependabot.yml`, scoped to the appliance directory and targeted at the appliance
branch:

```yaml
  - package-ecosystem: "docker"
    directory: "/os/rootfs"
    target-branch: "develop-v2"
    schedule:
      interval: "weekly"
```

and drop `/os/rootfs` from the `develop-v2`-only `directories:` list, so there is one place that
[...]

Issue #1048, opened 2026-08-16
The post-merge release gate has never run: every push to main skips it, and a skipped job reads as green


VijitSingh97:
## What is true today

`.github/workflows/release-gate.yml` presents itself as the tier-4 live-node gate on `main`. It has **never executed**. Every recent push to `main` shows `skipped`:

```
2026-08-14T14:48:45Z  main  skipped
2026-08-14T14:38:45Z  main  skipped
2026-08-14T04:32:08Z  main  skipped
2026-08-02T14:29:48Z  main  skipped
2026-08-02T00:06:17Z  main  skipped
```

The job carries `if: github.event_name == 'workflow_dispatch' || vars.ENABLE_RELEASE_GATE == 'true'`. Verified against the API:

- `repos/…/actions/variables` returns **no variables** — `ENABLE_RELEASE_GATE` is unset.
- `repos/…/actions/runners` returns **no runners** — nothing carries the `self-hosted, pithead-release` labels.

So the guard is doing exactly what its comment says it should, and the effect is that the workflow is inert.

## Why this matters even though the release is still gated

The cut-time gate is real: `scripts/release.sh` stage 2 runs `make test` and then `make test-integration ARGS="$RELEASE_INTEGRATION_ARGS"` against the real nodes, and aborts on failure. Combined with the operator-run pre-cut e2e on the bench, a release does not ship ungated.

The problem is the **signal**. A skipped job is green/neutral, so `main` displays a passing release gate that never ran. Anyone reading this repo — including a future session deciding what still needs proving — would reasonably conclude that merges to `main` are validated against live nodes post-merge. They are not. This is the same failure shape as the green suites that passed on an unparseable Caddyfile: the check is not lying, but nobody can tell it isn't checking.

It also bears directly on the standing goal that release gating should cover everything short of syncing a full chain.

## Options

1. **Register the runner** on the release server with the `pithead-release` labels (ephemeral / just-in-time, own runner group) and set `ENABLE_RELEASE_GATE=true`. The workflow then does what it says. `docs/dev/release-server.md` already describes the setup; this is the intended end state.
2. **If a permanently registered runner on a key-holding box is not wanted** — a defensible position for a public repo — then stop implying the gate is automatic. Rename the workflow to say it is dispatch-only, drop the `push: [main]` trigger so no skipped run is recorded against `main`, and state in `releasing.md` that the live-node gate is operator-run at cut time via `release.sh` plus the bench e2e.

[...]

Issue #1156, opened 2026-08-20
The dashboard image CVE scan reddened with no commit: three more util-linux advisories, and the pinned base is already the newest published digest


VijitSingh97:
## The dashboard image scan went red with no commit — three more util-linux advisories

`Build image (dashboard)` fails on every PR opened after ~04:43 today. CI was **green on
`develop-v2` at 04:43** and red at 13:32, with nothing in between: the advisories were published.
Total: 27 HIGH.

```
CVE-2026-53612   bsdutils, libblkid1, liblastlog2-2, libmount1, libsmartcols1,
CVE-2026-53613   libuuid1, login, mount, util-linux
CVE-2026-53614
```

Same family and same mechanism as **CVE-2026-53615**, which `.trivyignore` already accepts. That
entry's own comment predicted this exactly:

> `--ignore-unfixed` hid this until Debian published a fix, at which point it reddened every branch
> at once without any change on our side.

## A base bump does NOT fix it — checked, not assumed

The accepted entry says the fix is the Debian point release `2.41.5-0+deb13u1` and that no published
`python:3.11-slim` carries it. Still true, and stronger than that:

```
build/dashboard/Dockerfile:7   FROM python:3.11-slim@sha256:9c900dea…
docker pull python:3.11-slim   -> sha256:9c900dea…      (the SAME digest)
util-linux, libuuid1, mount    -> 2.41-5                (not 2.41.5-0+deb13u1)
```

**The digest we pin already IS the newest published one.** There is nothing to bump to, so this
cannot be closed by Dependabot or by a manual base bump — only by upstream rebuilding the image on
the fixed Debian packages.

## What to do

Accept `CVE-2026-53612`, `CVE-2026-53613` and `CVE-2026-53614` in the same entry as
`CVE-2026-53615`, with the family named once and one clearing condition for all four: a
`python:3.11-slim` digest shipping `util-linux >= 2.41.5-0+deb13u1`. The weekly CVE sweep (#833)
re-surfaces them, so they cannot rot silently.

Both lanes need it, and the file is shared, so it belongs on `develop` and reaches `develop-v2`
through the next twin sync — not the other way round.

## The pattern is now worth naming

This is the **third** time in one day that an untouched branch went red because the CVE database
moved rather than because the code did (#1153 for the appliance rootfs, twice; this one for the DIY
image). Two consequences worth holding:

- **"The branch was green when I based off it" is not evidence about now.** Check the base branch's
[...]

Issue #1073, opened 2026-08-17
CI is red on every branch: CVE-2026-53615 (util-linux) became fixable and blocks the dashboard Trivy gate


VijitSingh97:
Repo-wide CI break, not caused by any one PR. Every branch fails `Build image (dashboard)` at the Trivy step.

## Evidence it is time-based, not code-based

| run | branch | time | result |
|---|---|---|---|
| 31972732442 | `develop` | 2026-08-16 21:11 | success |
| 31987353676 | `main` | 2026-08-17 02:13 | **failure** |
| 31994562586 | `fix/1024-doc-pointers-in-operator-text` | 2026-08-17 04:27 | **failure** |

`main` and the feature branch share no changes, and `develop` passed the same job with the same Dockerfile hours earlier. Trivy refreshes its vulnerability DB per run, so identical image bytes flipped from pass to fail.

## What changed

The scan gates on **fixable** HIGH/CRITICAL only (`ignore-unfixed: true`, `.github/workflows/ci.yml:117`). Four CVEs are reported; three are already accepted in `.trivyignore` (CVE-2026-45447, CVE-2026-23949, CVE-2026-24049).

The new one is **CVE-2026-53615**, in the `util-linux` family shipped by the `python:3.11-slim` base — `util-linux`, `libuuid1`, `mount`, `login`. Debian has now published fixed packages (`2.41-5`, and `1:4.16.0-2+really2.41-5` for `login`), so `ignore-unfixed` stopped suppressing it and it became blocking.

## Options

1. **Bump the base image digest** — `build/dashboard/Dockerfile:7` pins `python:3.11-slim@sha256:db3ff2e1…`. This is the path `.trivyignore` already anticipates: *"Cleared when Dependabot (docker) bumps the base image digests."* Preferred, since it also clears the three accepted entries.
2. **Accept it in `.trivyignore`** with a rationale, matching the existing entries' format.

Deliberately not fixing this here: option 2 is a security-posture call about accepting a CVE, and option 1 wants a deliberate base bump rather than a drive-by. Whoever picks it up should check whether the same base is used by the other images — the failure is currently reported only for the dashboard image, which is worth confirming rather than assuming.

## Why it matters now

Three sessions are working this repo in parallel. Anyone who pushes will see a red `Build image (dashboard)` and reasonably suspect their own change. It also blocks any merge gated on green CI, including the v2 wave and PR #1027.

PR #1166, opened 2026-08-21
fix(ci): give the appliance rootfs CVE sweep a schedule that can actually fire (#1162)

### LINKED ISSUE #1162: The appliance rootfs CVE sweep has never run: its schedule lives on a branch a schedule cannot fire from

VijitSingh97:
Closes #1162.

`os-rootfs.yml` declares a weekly CVE sweep over the appliance rootfs and has never run one. GitHub
fires `schedule:` from the default branch only, and this file existed on `develop-v2` alone — the
one workflow that was not on both branches. 96 runs since 2026-08-02, every one a `pull_request`.

The appliance's Debian userland has no apt at runtime, so this scan is the only eye on podman,
netavark, openssl and the kernel between releases. It has been shut the whole time, and looked
exactly like a sweep that ran and found nothing.

## Why moving the file is only half the fix

Copying it to `develop` unchanged makes it worse. `develop` has no `os/`, so `present=false`, both
real steps skip on their `if:`, and the job reports **green every Monday for ever** — a scheduled
false GREEN instead of a schedule that never fires. That is #1048 again.

So two things change alongside the move:

- **The checkout names the tree it is scanning.** A PR scans the PR; `schedule` and
  `workflow_dispatch` scan `develop-v2` by name. A dispatch is therefore a genuine rehearsal of the
  schedule rather than a different code path, which is what makes the verification below worth
  anything.
- **The tree guard stops being the thing that decides whether the sweep happens.** A missing tree is
  legitimate only on a PR (a PR into `develop` can still reach this workflow through the
  `.trivyignore` path filter). Anywhere else it means the explicit checkout did not do what it was
  asked, so the run fails instead of reporting a scan it did not do.

### The expression is written negated on purpose

```yaml
ref: ${{ github.event_name != 'pull_request' && 'develop-v2' || '' }}
```

`A && B || C` in an Actions expression is boolean logic, not a ternary, and the empty string is
**falsy**. The natural-reading form — `github.event_name == 'pull_request' && '' || 'develop-v2'` —
collapses to `''` on the true branch and falls straight through to `'develop-v2'`, so **every PR
would scan `develop-v2` instead of the PR**. The truthy value has to be the `develop-v2` arm. I
wrote it the wrong way round first; it is called out in a comment so the next person does not
"simplify" it back.

## What was actually RUN

- **The guard's five states, driven out-of-tree.** The driver extracts the `run:` block verbatim
  from the YAML (`yaml.safe_load` → the `tree` step), so it cannot drift from the file:

  | EVENT | `os/rootfs/Dockerfile` | rc | output |
  |---|---|---|---|
  | `schedule` | present | 0 | `present=true` |
  | `pull_request` | present | 0 | `present=true` |
  | `pull_request` | absent | 0 | `present=false` + a step-summary line saying it was NOT scanned |
  | `schedule` | absent | **1** | `::error::` |
  | `workflow_dispatch` | absent | **1** | `::error::` |

- **The mutation that proves the driver can fail.** Respelling `elif [ "$EVENT" = "pull_request" ]`
  to `elif true` — the pre-#1162 always-skip — flips the `schedule`+absent state from rc=1 to rc=0
  and brings `present=false` back. The driver sees it.
- **`zizmor 1.25.2 --offline .github/workflows/`** — no findings (7 suppressed, pre-existing).
- **`make lint-yaml lint-md lint-docs-voice`** — pass. The new file is in yamllint's list.
- **The `ref: ""` precedent, checked rather than assumed.** `pin-watch.yml`'s product lane ships
  `ref: ""` and its dispatch on 2026-08-21 produced a correct report, so an empty `ref` really is
  checkout's default here.

## What was NOT run, and what still owes evidence

- **The workflow itself.** Expression evaluation and the paths filter can only be proven by a real
  run. The verification is a `workflow_dispatch` on `develop` after merge, and I will read the log
  for `docker build os/rootfs` actually executing — **a green run is not evidence here**, which is
  the whole point of the issue.
- **This PR's own check cannot scan anything.** It is a PR into `develop`, `develop` has no `os/`,
  so the job takes the skip path by design. That is the change behaving correctly, not evidence for
  it. The skip now writes to the step summary so a green check that scanned nothing says so.
- **No test tier covers `.github/workflows`.** `docs/dev/testing-strategy.md` fixes four tiers and
  none of them covers CI config; `zizmor` is the only lint over these files and it does not check
  step inputs. The comments plus the out-of-tree driver above are the only guards. I did not invent
  a fifth tier.
- **The twin sync is owed.** `develop-v2` still carries the old copy of this file. The sync PR has
  to resolve that add/add to develop's version byte-for-byte, and I will check the merged tree's
  contents rather than trusting a clean merge.

## Attacks that did not land

- *Does a PR into `develop-v2` now scan `develop-v2`'s tip instead of the PR?* No — `pull_request`
  takes the `''` arm.
- *Does the scheduled run read the wrong `.trivyignore`?* No — `trivyignores: .trivyignore` resolves
  inside the checked-out tree, which is `develop-v2`'s, so the appliance-only `CVE-2026-34040` mute
  applies where it should.
- *Does `git rev-parse HEAD > os/rootfs/BUILD_COMMIT` stamp the wrong commit?* No — after the
  explicit checkout, HEAD is `develop-v2`'s tip, which is the tree being built.
- *Does `concurrency` let a dispatch cancel a running sweep?* No — `cancel-in-progress` is true only
  for `pull_request`; a dispatch queues behind it.
- *Is `contents: read` enough to check out another branch?* Yes, same repo.

## Also in this PR

A `CONTRIBUTING.md` section writing the rule down once. Four gates in one week were blind for this
one structural reason (#1048, #1146, #1162, #1163) and each was found separately. The branch model
already lives in that section — `docs/dev/releasing.md` points at it as the authority — and it did
not mention `develop-v2` at all. It now names the two lanes, states the rule, and gives the command
that proves it from run history rather than from the file, because the file always looks fine.

### DIFF
--- a/.github/workflows/os-rootfs.yml
+++ b/.github/workflows/os-rootfs.yml
@@ -1,0 +1,104 @@
+name: OS rootfs scan
+
+# Build the appliance rootfs (os/rootfs/Dockerfile) and Trivy-scan it (#833). The appliance's
+# Debian userland — podman, netavark, openssl, the kernel — is frozen at bake time and has no
+# apt at runtime, so this scan is the eye on the base OS between releases. Runs on PRs that
+# touch os/ and weekly on a schedule; a red scheduled run in the Actions tab is the alert, the
+# lychee.yml posture. It is its own workflow (not a build-images matrix entry in ci.yml)
+# because the xmrig prebuild makes this the slowest image in the repo — the paths filter keeps
+# it off stack-only PRs.
+#
+# THIS FILE LIVES ON THE DEFAULT BRANCH (`develop`) AND MUST STAY THERE, even though everything it
+# scans is on `develop-v2` (#1162). GitHub fires `schedule:` from the default branch only, so while
+# this file was develop-v2-only its weekly sweep never ran once — 96 runs over 18 days, every one a
+# `pull_request`. Living on `develop` is only half of it: a job on `develop` still checks out
+# `develop`, which has no os/, so the checkout below names the appliance branch explicitly for every
+# trigger that is not a PR. Do not "fix" the asymmetry by moving the file back — the rule and the
+# one-command check for it are in CONTRIBUTING.md § The two lanes.
+on:
+  pull_request:
+    paths:
+      - "os/**"
+      - ".github/workflows/os-rootfs.yml"
+      - ".trivyignore"
+  schedule:
+    - cron: "0 6 * * 1" # Mondays 06:00 UTC, after ci.yml's 05:00 image sweep
+  workflow_dispatch:
+
+# Least privilege (#282): read-only, same as ci.yml.
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  rootfs-image:
+    name: Build + scan the appliance rootfs
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          # A PR run scans the PR (the empty string is checkout's default: the triggering ref).
+          # Everything else — the weekly sweep and any manual dispatch — is asking about the
+          # appliance, which lives on develop-v2 whichever branch the trigger fired from. Naming it
+          # here is what makes the sweep real rather than a green skip (#1162). It also means a
+          # dispatch is a true rehearsal of the schedule instead of a different code path.
+          # Written negated on purpose: in an Actions expression `A && B || C` is not a
+          # ternary, it is real boolean logic, and '' is FALSY. `event == 'pull_request' && ''`
+          # would collapse to '' and fall through to 'develop-v2', so every PR would scan
+          # develop-v2 instead of the PR. The truthy value has to be the develop-v2 arm.
+          ref: ${{ github.event_name != 'pull_request' && 'develop-v2' || '' }}
+          persist-credentials: false
+      - name: Check the appliance tree exists
+        # Only a PR may legitimately miss the tree: a PR into `develop` can reach this workflow
+        # through the .trivyignore path filter, and `develop` has no os/. Every other trigger has
+        # just checked out develop-v2 by name, so a miss there means the checkout did not do what
+        # it was asked — and a quiet skip would report a green weekly sweep of nothing, which is
+        # the exact failure #1162 was filed for. Fail instead.
+        id: tree
+        env:
+          EVENT: ${{ github.event_name }}
+        run: |
+          if [ -f os/rootfs/Dockerfile ]; then
+            echo "present=true" >>"$GITHUB_OUTPUT"
+          elif [ "$EVENT" = "pull_request" ]; then
+            echo "present=false" >>"$GITHUB_OUTPUT"
+            echo "os/rootfs/Dockerfile not on this branch; nothing to scan"
+            echo "The appliance rootfs was NOT scanned: os/rootfs/Dockerfile is not on this PR's branch." \
+              >>"$GITHUB_STEP_SUMMARY"
+          else
+            echo "::error::checked out develop-v2 but os/rootfs/Dockerfile is missing — the appliance checkout did not do what it was asked, so this run cannot report a scan it did not do"
+            exit 1
+          fi
+      - name: docker build os/rootfs
+        if: steps.tree.outputs.present == 'true'
+        # BUILD_COMMIT is a build artifact (os/build-image.sh stamps it; the Dockerfile COPYs
+        # it), so stamp it here too. PITHEAD_UPDATER=rauc matches build-image.sh's default so
+        # the scan covers the updater packages; the test args stay empty — release variant.
+        # images/ holds only .keep here: the staged wizard image is a separate scan target
+        # (ci.yml builds and scans the dashboard image directly).
+        run: |
+          git rev-parse HEAD > os/rootfs/BUILD_COMMIT
+          docker build -f os/rootfs/Dockerfile -t pithead-os-rootfs:ci \
+            --build-arg PITHEAD_UPDATER=rauc .
+      - name: Scan rootfs for CVEs (Trivy)
+        if: steps.tree.outputs.present == 'true'
+        # Same gate as ci.yml's image scan: actionable (fixable) HIGH/CRITICAL only; accepted
+        # findings live in .trivyignore with a rationale and how they clear.
+        #
+        # cache: false for the reason spelled out over ci.yml's scan (#1159) — this is the repo's
+        # other trivy call site. It matters more here: the appliance's Debian userland is baked and
+        # has no apt at runtime, so this scan is the only eye on podman, netavark, openssl and the
+        # kernel between releases, and since #1162 that eye is finally open on a schedule.
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
+        with:
+          image-ref: pithead-os-rootfs:ci
+          scanners: vuln
+          severity: HIGH,CRITICAL
+          ignore-unfixed: true
+          exit-code: "1"
+          trivyignores: .trivyignore
+          cache: "false"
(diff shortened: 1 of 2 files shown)


VijitSingh97:
## This PR's own run proves the `pull_request` half on a real runner

I said in the body that this PR's check "cannot scan anything" and would take the skip path. It did —
and in doing so it settled the one claim I could not settle locally: **that the negated expression
really does hand `pull_request` the empty arm.**

Job `Build + scan the appliance rootfs`, 8s, success. Checkout step:

```
git fetch ... origin +11e71ba…:refs/remotes/pull/1166/merge
git checkout --progress --force refs/remotes/pull/1166/merge
HEAD is now at 11e71ba Merge 6d315d3 into ffb7924
```

It checked out **the PR merge ref**, not `develop-v2`. Had the expression been written the
natural-reading way round — `github.event_name == 'pull_request' && '' || 'develop-v2'` — the empty
string is falsy, the whole expression would have collapsed to `develop-v2`, this step would have
fetched the appliance branch, `os/rootfs/Dockerfile` would have been **present**, and the job would
have gone on to build and scan `develop-v2`'s tree while claiming to be a check on this PR.

The guard then behaved as designed:

```
env:
  EVENT: pull_request
os/rootfs/Dockerfile not on this branch; nothing to scan
```

Step outcomes: `Check the appliance tree exists -> success`, `docker build os/rootfs -> skipped`,
`Scan rootfs for CVEs (Trivy) -> skipped`.

So three of the five states now have evidence from two independent places — the out-of-tree driver
in the body, and this run. **The `schedule` state still has none**, and it is the one the issue is
about. That evidence is the post-merge `workflow_dispatch`, read for `docker build os/rootfs`
actually executing, and I will post it here rather than assume a green means anything.

VijitSingh97:
## Adversarial review: 5 attacks, 2 landed — and the first one was this change reintroducing its own bug

Pushed as `3a08047`. Both landed findings are recorded here with what refuted the other three,
because "attacked and survived" is only worth something if the failures are shown too.

### Landed, and it changed the diff — a new vacuous green on `develop`

`.trivyignore` is in the paths filter and lives on **both** lanes. So the moment this workflow is on
`develop`, a PR editing that file on `develop` triggers a check named "Build + scan the appliance
rootfs", finds no `os/`, skips both real steps, and goes green in **8 seconds** on a job whose own
header calls it "the slowest image in the repo".

That is not hypothetical. #1156 added four util-linux mutes to `.trivyignore` on `develop`; #1073 did
the same shape earlier. Under this change, the next one of those merges with a green appliance-scan
tick beside it, having rebuilt and rescanned nothing — and the file it would have needed is
`develop-v2`'s `.trivyignore`, which is 28 lines longer.

**I had noticed this and talked myself out of it** as cosmetic, on the grounds that the skip writes
to the step summary. The summary is on the run page; the green tick is what a reviewer reads. Fixed
with `branches-ignore: [develop, main]` — the two long-lived branches with no `os/` — so the PR
trigger only fires where a scan is possible. Stacked appliance branches keep their coverage, and a
`.trivyignore` edit reaches a real scan through the appliance-side sync PR.

One correction to the finding as filed, from the verification pass: it called this **blocking** on
the grounds it could gate a merge. It cannot — the job is not in
`branches/develop/protection.required_status_checks.contexts`. It can only mislead a human. Right
finding, overstated mechanism.

### Landed — CONTRIBUTING asserted a fact that is not true

The new section said this rule went unnoticed "four times (#1048, #1146, #1162, #1163)". #1048 is
not an instance of it: its schedule fired, and the job skipped itself behind an unset repository
variable. Three instances, with #1048 named next to them as the sibling failure.

That also exposed a gap in the recipe the section teaches. `gh run list --json event` proves whether
a schedule ever fired — but #1048 *passes* that test. So the section now carries the second half:
open the newest scheduled run and check its steps ran rather than skipped.

A doc that states a wrong issue number in a rule about verifying claims is worse than one that says
nothing, and the repo's own bar is that the code is the source of truth.

### Refuted

- **"`workflow_dispatch` silently ignores the ref you dispatched."** Half true, wrong conclusion.
  The override is real; "silently" is not. `actions/checkout` prints `ref: develop-v2` in its `with:`
  group and logs the resolved 40-char SHA — shown from a real dispatch run of `pin-watch.yml` in this
  repo, which uses the same pinned action.
- **"The add/add conflict on the next sync has no documented resolution."** The conflict is real and
  reproduces (`git merge-tree --write-tree origin/develop-v2 <branch>` → exit 1, add/add). Every
  inference from it was wrong: the resolution *is* stated in this PR's body, direction and all, along
  with a commitment to inspect the merged tree rather than trust a clean merge.
- **"The hardcoded `develop-v2` breaks when the lane retires."** True but not introduced here — the
  same literal is already merged on `develop` twice, in `pin-watch.yml`. Filed as #1169 rather than
  scope-crept into this PR, because the retirement touches four places and deserves one list.

The `ref:` expression is right for every configured event (`pull_request_target` and `merge_group`
are not configured, so there is no third path). A PR into `develop-v2` still scans its own merge ref,
not the branch tip. `trivyignores: .trivyignore` is workspace-relative, so the scheduled run reads
develop-v2's copy — which matters, because had the checkout stayed on `develop` the sweep would have
gone **red** on `CVE-2026-34040`, an accepted appliance finding that only develop-v2's file mutes.
The `else` branch has no accidental entry: a deleted branch fails at checkout, a renamed Dockerfile
correctly goes red.

And the premise was re-verified independently rather than taken from the issue:
`[{"event":"pull_request","n":98}]` — still zero scheduled runs, ever.

**Still owed, unchanged:** the `schedule` path has no evidence at all. That is the post-merge
dispatch, read for `docker build os/rootfs` executing.

Issue #1129, opened 2026-08-18
Upgrade Tari to v6.0.0 with P2Pool v4.18.1 for mainnet fork compatibility


VijitSingh97:
Prioritized by the operator on 2026-09-22. Upgrade the mining pair together: **Tari node + console wallet v6.0.0 and P2Pool v4.18.1**. Supersedes the original v5.6.0 target and dormant/maintenance placement; historical investigation remains in the comments.

## Implementation and operator attachments

- [Detailed validation plan: V1–V12, commands, negative controls and evidence requirements](https://github.com/p2pool-starter-stack/pithead/issues/1129#issuecomment-5780174447).
- [Migration guide: preflight, copied-data rehearsal, DIY/appliance cutover and recovery](https://github.com/p2pool-starter-stack/pithead/issues/1129#issuecomment-5780174781).

These are planned acceptance criteria, not passing test results. They supersede any older comment that treats a startup `chain_id` or source/image rollback alone as adequate proof.

### Closure checklist

- [ ] Coordinated Tari node/wallet 6.0.0 + P2Pool 4.18.1 pins, hashes and runtime versions verified across Compose and supported Quadlet/appliance packaging.
- [ ] Registry-specific config-test discovery and pin/release fixtures updated; real config smoke ran against the candidate image (no Docker-unavailable SKIP).
- [ ] V1–V12 results attached to the exact-head PR, including actual new-payload acceptance and discriminating old/mutated-payload rejection; mainnet fork-boundary rules exercised.
- [ ] Representative 5.3.1 copied-data migration measured; optional wallet scan and a matching-data restore rehearsal proved.
- [ ] Generic e2e cleanup cannot reopen migrated shared data with old binaries; test isolation/snapshot and timeout behavior demonstrated.
- [ ] Appliance commit-before-migration, fallback before migration, and visible post-commit failure/recovery verified; remote-node operators included in compatibility plan.
- [ ] Required repository checks and independent review passed. Remaining rollout/soak evidence is linked to its owner, with no implied production deployment or unrun test marked complete.

## Why this is now a GA mining gate

[...]

PR #1168, opened 2026-08-21
chore(os): bump the appliance docker-compose pin to v5.5.0 (#1165)

VijitSingh97:
Refs #1165. **Deliberately does not close it** — that is a standing report issue, and closing it
would make the next weekly run file a duplicate. `pin-watch.yml` dedups by exact title over the
**open** issue list (a search-index dedup files a second issue on the very next run, so it uses
`gh issue list --state open` instead). A closed report is therefore invisible to it. #1144, the
product lane's twin, is open for the same reason. Four of its rows are also still stale; this PR
closes the `compose` row only.

The first automated appliance pin report — produced by the new appliance lane #1146 added to
`pin-watch.sh`, on its first ever execution — found `compose` stale at `v5.4.0` against upstream
`v5.5.0`. That pin is on the appliance's own supply chain and no automated run had ever looked at
it.

## Two ARGs, not one

`COMPOSE_VERSION` is only an ldflag; `COMPOSE_COMMIT` is what is actually built. Moving the version
alone silently rebuilds the old code with a new version string on it — the report prints this
warning itself, and it is why the pin is a pair.

```
v5.5.0 = 870908cc8f07f5e90acdf5d34dd1b96a4fe51d16
```

## The two standing claims about compose, re-derived rather than assumed

Bumping compose moves the whole MVS closure, so the two entries elsewhere in the tree that describe
compose's module graph had to be re-checked against `v5.5.0` itself, not against the report:

- **`COMPOSE_GO_RAISES="golang.org/x/mod@v0.40.0"` is still needed and is still a raise.** `v5.5.0`
  pins `golang.org/x/mod v0.38.0`, same as `v5.4.0`. Had it moved past v0.40.0, `go get` would have
  silently *downgraded* it.
- **`.trivyignore`'s `CVE-2026-34040` block is still accurate.** `v5.5.0` still pins
  `github.com/docker/docker v28.5.2+incompatible`, so its rationale — that Trivy's named fix
  `29.3.1` is a Moby release and not a version of this module — still describes the pinned version.

One thing did move on its own: compose raised `google.golang.org/grpc` from `v1.82.1` to `v1.83.0`
between the two tags. That is upstream's change, not a raise of ours, and it produced no finding.

## What was actually RUN — on real amd64 hardware, with no ignore file

`docker build --target gobuild` on the bench, then the binaries extracted with `docker create` +
`docker cp` and scanned directly.

**The binaries' own build stamps** (`go version -m`), which is the answer to "did the pin actually
move" that a diff cannot give:

```
mod  github.com/docker/compose/v5   v5.5.0+dirty
dep  github.com/docker/docker       v28.5.2+incompatible
dep  golang.org/x/mod               v0.40.0        <- the raise landed
dep  google.golang.org/grpc         v1.83.0
```

`docker-compose version` → `Docker Compose version v5.5.0`. Both binaries `go1.26.6`, the patched
toolchain. The build's own `RUN /out/docker-compose version && /out/cosign version` guard passed.

**Trivy, fixable HIGH/CRITICAL, with NO ignore file at all**, current DB (`Need to update DB`,
108.49 MiB fetched — no day-frozen cache, per #1159):

```
out/cosign          0 findings
out/docker-compose  1 finding
  CVE-2026-34040  github.com/docker/docker  v28.5.2+incompatible -> 29.3.1  [HIGH]
```

Identical to the pre-bump result recorded in `.trivyignore` ("cosign 0 findings, compose 1"), and
the single finding is the one that file already documents as genuinely un-raisable. **The bump
changes nothing about the accepted-findings picture** — which is the thing that had to be proven
before trusting it, not assumed from release notes.

## A trap worth naming

My first attempt scanned the two extracted binaries with `trivy fs`. It exited **0** with an empty
report — `Target: -`, `Type: -`. `trivy fs` does not analyse standalone Go binaries that way, so
"no findings" there meant "nothing was scanned", and reporting it as clean would have been this
repo's signature defect committed by its own reviewer. `trivy image` over the build stage does find
them (`out/docker-compose | gobinary | pkgs=190`), which is where the numbers above come from. An
empty scan and a clean scan look the same in the exit code; only the package count tells them apart.

## What was NOT run

- **The full rootfs image build and scan.** `--target gobuild` covers the Go half only; the Debian
  userland half is unchanged by this diff and is what CI's `os-rootfs.yml` job builds and scans on
  this PR, with `.trivyignore` applied.
- **No appliance boot.** Nothing here touches boot, RAUC or the wizard — it is two ARG values — so
  no KVM battery was run for it.
- **The other four stale pins in #1165 are untouched**: `monero 0.18.5.0 → 0.18.5.1`,
  `p2pool 4.16 → 4.18`, `tari 5.3.1 → 5.6.0` (that one is #1129, and it is a one-way wallet-DB
  migration, not a bump), `socket-proxy 0.4.2 → 0.5.0`. This PR closes the compose row only; the
  issue's remaining rows are reported again by the next weekly run, which is the other reason it
  stays open.

### DIFF
--- a/os/rootfs/Dockerfile
+++ b/os/rootfs/Dockerfile
@@ -18,11 +18,11 @@
 # Both projects publish release binaries built on whatever Go toolchain they happened to use, and
 # that toolchain's stdlib CVEs become ours — baked into a read-only appliance image that cannot be
 # patched in place. In August 2026 that was nine fixable HIGH findings the scan gated on, and no
-# pin bump could clear them: cosign v3.1.3 (the newest release) ships built with go1.26.4, the
-# same toolchain as v3.1.2, and docker-compose v5.4.0 ships go1.26.5 — both below the go1.26.6
-# that carries the fixes. Verified by reading the build stamps out of the published binaries, not
-# from release notes. Waiting for upstream to rebuild is not a schedule we control, so we compile
-# them ourselves on a patched toolchain and the findings go away at the source.
+# pin bump could clear them: cosign v3.1.3 (the newest release) ships built with go1.26.4, the same
+# toolchain as v3.1.2, and docker-compose v5.4.0 (the pin then) shipped go1.26.5 — both below the
+# go1.26.6 that carries the fixes. Verified by reading the build stamps out of the published
+# binaries, not from release notes. Waiting for upstream to rebuild is not a schedule we control,
+# so we compile them ourselves on a patched toolchain and the findings go away at the source.
 #
 # Provenance is not weakened by this, it moves: the old pin was a sha256 of a downloaded binary,
 # this one is an exact upstream COMMIT (never a tag — a tag can be re-pointed, the same reason
@@ -35,8 +35,8 @@
 # claimed otherwise would be lying about the provenance this paragraph exists to describe.
 FROM golang:1.26.6-trixie@sha256:b75d466dd608587fd66cca705a307ba65b889827d06ad61d6a75f0482b51b7c7 AS gobuild
 
-ARG COMPOSE_VERSION=v5.4.0
-ARG COMPOSE_COMMIT=ef61d7410a0c816a71705026e638ec256a591d69
+ARG COMPOSE_VERSION=v5.5.0
+ARG COMPOSE_COMMIT=870908cc8f07f5e90acdf5d34dd1b96a4fe51d16
 ARG COSIGN_VERSION=v3.1.3
 ARG COSIGN_COMMIT=11926fa5bbbbde47e88fc006b625a17769b743b2
 # Modules raised above what each project's own go.sum asks for, because upstream has not cut a
--- a/os/rootfs/Dockerfile
+++ b/os/rootfs/Dockerfile
@@ -69,8 +69,18 @@ ARG COSIGN_GO_RAISES="golang.org/x/mod@v0.40.0 google.golang.org/grpc@v1.82.1"
 ENV CGO_ENABLED=0
 
 # Blobless clone then detach to the commit: the tag is only a convenience for the version string.
+#
+# The `describe --exact-match` line is what makes the pin a PAIR rather than two values that happen
+# to sit next to each other. `${COMPOSE_VERSION}` is only an ldflag, `${COMPOSE_COMMIT}` is what is
+# actually built, so bumping the version alone rebuilds the OLD code and stamps the NEW version on
+# it — and every downstream check agrees it worked: the build succeeds, `docker-compose version`
+# prints the new number, and scripts/pin-watch.sh reads the VERSION ARG and reports the component
+# current for ever. Nothing else in the tree reads `${COMPOSE_COMMIT}` at all. The clone is blobless
+# but not shallow, so the tag objects are present and this costs nothing.
 RUN git clone --filter=blob:none --no-checkout https://github.com/docker/compose /src/compose \
     && git -C /src/compose checkout --detach "${COMPOSE_COMMIT}" \
+    && { [ "$(git -C /src/compose describe --tags --exact-match "${COMPOSE_COMMIT}")" = "${COMPOSE_VERSION}" ] \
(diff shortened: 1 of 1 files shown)


VijitSingh97:
## Adversarial review: 4 attacks, 2 landed — and both were checks that could not fail

Pushed as `44e3318`. Both landed findings are in code this PR was only passing through, and both are
the same defect as the one the bump exists to fix, one level down. Each now ships with the mutation
that proves it can go red, **run on the bench, not named**.

### M1 — the half-done bump the comment warns about, made impossible

The hazard is stated plainly in the pin comment and nothing enforced it: `COMPOSE_VERSION` is an
ldflag, `COMPOSE_COMMIT` is what is built. `COMPOSE_COMMIT` appears in exactly two places in the
whole repo — the ARG and the checkout — and was never compared to anything, so moving the version
alone builds the old code, stamps the new number on it, and every downstream check agrees: build
succeeds, `docker-compose version` prints the new version, and `scripts/pin-watch.sh` reads the
VERSION ARG and reports the component current for ever.

Mutation: keep `ARG COMPOSE_VERSION=v5.5.0`, pass v5.4.0's commit.

```
$ docker build --target gobuild --build-arg COMPOSE_COMMIT=ef61d74… .
#5 29.45 COMPOSE_COMMIT ef61d74… is not tag v5.5.0 — a compose bump is TWO ARGs;
         resolve the pair with: gh api repos/docker/compose/commits/v5.5.0 --jq .sha
#5 ERROR: … exit code: 1
M1 BUILD EXIT=1
```

Before this commit that build succeeds and ships v5.4.0's code labelled v5.5.0. The same guard is on
cosign, whose pin has the identical shape.

### M2 — the smoke check could not detect the broken ldflag its own comment claimed it caught

`RUN /out/docker-compose version && /out/cosign version` was introduced to catch "a wrong build path
or a broken ldflag". It catches the first and not the second: `go build -ldflags "-X pkg.Var=value"`
for a symbol that does not exist is **silently ignored by the Go linker** — no warning, exit 0.

Mutation: rename the symbol to `internal.VersionRENAMED`, exactly what an upstream refactor or a
`/v6` module path would do.

```
#7 0.904 Docker Compose version dev
#7 ERROR: … /out/docker-compose version | grep -qF "${COMPOSE_VERSION}" … exit code: 1
M2 BUILD EXIT=1
```

That log line is the whole finding in one place: the binary **builds, runs and prints** — under the
old check it exits 0 and ships — and what it prints is `dev`. The file two paragraphs above presents
that version string as a provenance guarantee, so an appliance reporting `dev` for ever is a broken
promise, not cosmetics. The plain `version` call is kept first so the build log still shows real
output; the grep is what makes it an assertion.

### Refuted

- **"v5.5.0's container recreation could blow the RAUC health gate."** The premise is real and I
  verified it verbatim rather than through the reviewer: v5.5.0's release notes lead with *"Existing
  containers may be recreated the first time you run `compose up` after upgrading."* On the appliance
  that first `compose up` is `pithead-boot`'s, inside `./pithead up || fail_boot` plus a 90×10s
  health poll, and `graphroot = /data/containers/storage` means containers really do survive an A/B
  slot switch. The conclusion still does not follow: **an OS update already changes image digests, so
  the first boot after any update already recreates containers.** v5.5.0 changes when that
  reconciliation triggers, not whether the appliance can survive it. Named here because it is real
  upstream behaviour worth knowing, and it is on the list for the appliance update battery.
- **"The rewrapped paragraph should say upstream compose is now on go1.26.6."** True (docker/compose
  #14045) and refuted as a defect: the sentence is scoped "In August 2026" and now reads "(the pin
  then) shipped", which is exactly the edit that turns it into history. The module-graph argument
  that keeps the from-source build load-bearing is already written directly beneath it.

`COMPOSE_COMMIT` really is tag v5.5.0 — checked through the API (annotated tag
`7b30ae2e…` dereferencing to `870908cc…`), not from the diff. `-tags e2e`, the `./cmd` path and the
`internal.Version` symbol all still exist at v5.5.0. `COMPOSE_GO_RAISES` is still a raise, not a
downgrade. Nothing else in the tree asserts the old version.

Issue #1137, opened 2026-08-19
Test cannot fail: the compose digest pins assert that a digest is present, never which one — a half-done version bump is green


VijitSingh97:
## What is wrong

The compose digest-pin assertions check that *a* digest follows the tag. They never check *which*.

`tests/stack/test_compose.sh:129-131`

```
expect_present "tecnativa socket-proxy pinned by digest" "tecnativa/docker-socket-proxy:v0.4.2@sha256:"
expect_present "caddy pinned by digest" "caddy:2.11.4@sha256:"
expect_present "tari node pinned by digest" "minotari_node:v5.3.1-mainnet@sha256:"
```

The literal stops at `@sha256:`. In a `tag@sha256:...` reference **the digest is authoritative and the
tag is decoration** — which is the stated reason the pin exists at all (`docker-compose.yml:219`,
"Pinned by digest (#135) so the v5.3.1-mainnet tag can't be silently re-pointed").

So the half-done bump is green at every tier: move the tag to the new version in `docker-compose.yml`
and in this assertion's literal, leave the old digest in place, and the suite passes while the stack
keeps pulling the old image. `docker-compose.yml`, `test_compose.sh`, `docs/dev/releasing.md`, the
CHANGELOG and `scripts/release.sh pin tari` all then announce a version the box is not running.

This is live right now, on the Tari bump in #1129: four digest strings have to move together
(node index, node amd64 child, wallet index, wallet amd64 child) and leaving any one stale is
invisible.

## Second half: the console wallet has no digest assertion at all

There is no `expect_present` for `minotari_console_wallet`. It cannot be added where the other three
are: `$RENDERED` is built with `COMPOSE_PROFILES=local_node,local_tari`
(`tests/stack/test_compose.sh:56,74-75`), and `tari-wallet` is `profiles: ["tari_payout_confirm"]`
(`docker-compose.yml:305`), so it is not in that render — an `expect_present` there would fail
identically whether the pin were right or wrong.

It belongs below `tests/stack/test_compose.sh:269`, as a `jq_assert` on the profile-enabled JSON
render — the same reason the wallet healthcheck assertion sits there (see the comment at `:297`).

## Why it matters here

This is the "check that cannot fail" family again, on the supply-chain pin. The image digest is the
one thing in the compose file that decides what code actually runs on an operator's box.

## Fix

1. Assert the **whole** reference — `tag@sha256:<64 hex>` — not the prefix, for all three existing
[...]

PR #1176, opened 2026-08-21
fix(release): name both Tari images in the ingredient list, and guard the compose pins (#1138, #1137)

### LINKED ISSUE #1138: Release notes name only the Tari node: 'pin tari' cannot see the console-wallet image
## What is wrong

`scripts/release.sh:198`

```
tari) grep -oE 'quay.io/tarilabs/minotari_node:[^ ]+' docker-compose.yml | head -1 ;;
```

`pin tari` matches the **node image only**. The stack pins two Tari images — the base node and the
view-only console wallet (`docker-compose.yml:220` and `:308`) — and they are bumped together.

The release notes' ingredient list therefore reports one of the two. A release that moves the
console wallet and not the node, or moves them to different versions, reads as unchanged.

This is not hypothetical: #1129 bumps both.

## Fix

Either extend the `tari` case to report both images, or split it into `tari-node` / `tari-wallet`
and have the ingredient list name both. If reporting the node alone is deliberate (they are
expected to move in lockstep), say so in a comment and add an assertion that they carry the same
tag — silence is what makes this look correct.

## Tier

Tier 1. `scripts/release.sh` is already exercised by the release-smoke assertions; a pin that names
both images is a string assertion.

VijitSingh97:
Closes #1138. Advances #1137 (does **not** close it — see the bottom).

Two v1.19.3 items in the same family: the release's ingredient list and the compose pin assertions
both had a blind spot that made a **half-done bump** read as correct.

## #1138 — the ingredient list named one of the two Tari images

`pin tari` grepped `minotari_node` only. The stack pins two — the base node and the view-only
console wallet — and they are bumped together, so a release that moved the wallet read as unchanged
in the notes. #1129 moves both.

**`tari` stays the NODE rather than being renamed to `tari-node`, and that is the interesting part.**
`scripts/pin-watch.sh` reads this same arm through its `tree_pin()` for the weekly currency report.
Renaming it would have made `tree_pin tari` return empty, and the watcher would report
*"could not read the pin from the tree"* and fail its run — a caller the issue does not mention and
that the obvious fix breaks. A new `tari-wallet` arm feeds the notes instead, which now name both.

Watching one image for currency is only safe if the two really do move together, and that is not
assumed: `tests/stack/test_compose.sh` already asserts both Tari images carry the same tag.

## The half of #1138 the issue did not ask for

Only `p2pool`, `monero` and `xmrig-proxy` had a drift guard. The five pins read out of
`docker-compose.yml` and `build/tor` — `tari`, `tari-wallet`, `caddy`, `socket-proxy`, `tor-base` —
had **none**.

`pin()` emits the empty string when its grep stops matching. A renamed image or a changed tag format
and the release notes print `- caddy: ` with nothing after it: the ingredient list's entire job,
silently not done, on the exact surface an operator uses to check what they installed. All five are
guarded now, with `grep -F` so a value merely *similar* to one in the file cannot pass.

## #1137's residual

Parts 1 and 2 of #1137 already landed in PR #1147 — I verified that rather than taking the issue
body at face value, and confirmed by mutation that both landed assertions genuinely go red. **The
issue body is stale.**

One instance of its own defect survived that pass. `tecnativa/docker-socket-proxy` is pinned on
**two** services, `docker-proxy` and `docker-control`, and `expect_present` is a `grep -q` — it
matches either line. Bumping one proxy and leaving the other was green, and the two would then run
different socket-proxy builds, which is precisely the split the separate-proxy design exists to
prevent (one is read-only; the other is the start/stop control path). Now `expect_min … 2`, and the
literal renders exactly twice, so the count is measured rather than hoped for.

## What was actually RUN

**Full tier-1 suite: 1822 passed, 0 failed**, with the five new pin guards green.

**Three mutations, each isolated, all run:**

| mutation | result |
|---|---|
| `tari-wallet` arm's image respelled | **only** `pin tari-wallet` ✗ — other four ✓ |
| `tari` arm's image respelled (the pre-#1138 blind spot) | **only** `pin tari` ✗ — other four ✓ |
| `docker-compose.yml:865` (docker-control) bumped, `:813` left alone | `✗ … expected >= 2, got 1` |

The socket-proxy mutation is the one that matters: **before this change that same edit was green.**

### A mutation that lied first, and what it cost

My first attempt inserted the `# MUTANT` marker mid-line, which commented out the rest of the `case`
arm, broke the whole `case` block, and made **every** pin empty — all five guards went red for a
reason that had nothing to do with the arm under test. A mutation that breaks the file proves nothing
about the assertion. The marker now goes after `;;`, and each run asserts that only the mutated
component's guard flips, with `bash -n` confirming the file still parses. Recorded because a
mutation that is too blunt reads exactly like a mutation that worked.

## What was NOT run

- **No bench work.** This is `scripts/release.sh` and two tier-1 assertion files; nothing here is
  under `os/` and nothing changes what a container runs. The release-cut path this touches is
  exercised by the tier-1 unit tests around `write_manifest` and `pin`, not by a live cut.
- **No live release cut.** The ingredient list's new second line is proven by the drift guard and by
  reading `write_manifest`, not by cutting a release to look at the notes.
- **#1137 stays open.** Part 3 — resolving each pinned tag against its registry — is the registry
  half, and it needs a client with two different token flows (quay.io and Docker Hub).
  `scripts/pin-watch.sh` already states this in the file. That is the only thing that catches a
  tag/digest pair which is self-consistent in the file but wrong against upstream; every pin was
  checked by hand against its registry today and all five are correct.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -195,7 +195,14 @@ pin() {
     monero) grep -oE '^ARG MONERO_VERSION=.*' build/monero/Dockerfile | cut -d= -f2 ;;
     xmrig-proxy) grep -oE '^ARG XMRIG_PROXY_VERSION=.*' build/xmrig-proxy/Dockerfile | cut -d= -f2 ;;
     tor-base) grep -oE '^FROM [^ ]+' build/tor/Dockerfile | head -1 | awk '{print $2}' ;;
+    # `tari` is the NODE, deliberately, because scripts/pin-watch.sh reads this arm for upstream
+    # currency and both Tari images come from one upstream repo — a second row there would say the
+    # same thing twice. The stack pins two images though (node + view-only console wallet), they are
+    # bumped together, and the release notes have to name both or a wallet-only move reads as
+    # unchanged (#1138). tests/stack/test_compose.sh asserts the two carry the same tag, so the
+    # lockstep this relies on is guarded rather than assumed.
     tari) grep -oE 'quay.io/tarilabs/minotari_node:[^ ]+' docker-compose.yml | head -1 ;;
+    tari-wallet) grep -oE 'quay.io/tarilabs/minotari_console_wallet:[^ ]+' docker-compose.yml | head -1 ;;
     caddy) grep -oE 'caddy:[0-9.]+@sha256:[a-f0-9]+' docker-compose.yml | head -1 ;;
     socket-proxy) grep -oE 'tecnativa/docker-socket-proxy:[^ ]+' docker-compose.yml | head -1 ;;
     esac
@@ -659,7 +666,8 @@ write_manifest() {
         printf -- '- monerod: `%s`\n' "$(pin monero)"
         printf -- '- xmrig-proxy: `%s`\n' "$(pin xmrig-proxy)"
         printf -- '- tor base: `%s`\n' "$(pin tor-base)"
-        printf -- '- tari: `%s`\n' "$(pin tari)"
+        printf -- '- tari node: `%s`\n' "$(pin tari)"
+        printf -- '- tari console wallet: `%s`\n' "$(pin tari-wallet)"
         printf -- '- caddy: `%s`\n' "$(pin caddy)"
         printf -- '- docker-socket-proxy: `%s`\n' "$(pin socket-proxy)"
     } >"$out"
--- a/tests/stack/run.sh
+++ b/tests/stack/run.sh
@@ -2235,6 +2235,31 @@ for svc in p2pool monero xmrig-proxy; do
         bad "pin $svc resolves to a value in its Dockerfile" "got '$pv'"
     fi
 done
+# The same drift guard for the pins read out of docker-compose.yml and build/tor, which had none
+# (#1138). pin() emits the EMPTY string when its grep stops matching — a renamed image, a moved tag
+# format — and the release notes then print "- caddy: " with nothing after it. The ingredient list's
+# whole job, silently not done, on the exact surface an operator uses to check what they installed.
+# grep -F, not -E: these values contain '/' and '.', and a regex match would accept a value that is
+# merely similar to one in the file.
+for comp in tari tari-wallet caddy socket-proxy tor-base; do
+    case "$comp" in
+    tor-base) pin_src="$ROOT/build/tor/Dockerfile" ;;
+    *) pin_src="$ROOT/docker-compose.yml" ;;
+    esac
+    # shellcheck disable=SC1090
+    pv="$(
+        cd "$ROOT" || exit
+        set --
+        source "$REL" 2>/dev/null
+        set +eu
+        pin "$comp"
+    )"
+    if [ -n "$pv" ] && grep -qF -- "$pv" "$pin_src"; then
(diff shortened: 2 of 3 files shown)


VijitSingh97:
## Adversarial review: 3 attacks, 2 landed — and the first one defeated this PR's own guard

Pushed as `cd8efd9`.

### Landed — the drift guard could only detect an EMPTY pin

`grep -qF -- "$pv" "$pin_src"` compares a value that `pin()` extracted **from that very file**
against the same file. Every arm is a substring cut, so any non-empty answer is present by
construction. The only reachable failure was `[ -n "$pv" ]`.

The label claimed more than the code could test — this repo's signature defect, in the guard written
against it. The reviewer demonstrated it with the mutation that matters: re-point the `tari-wallet`
arm at the **node's** regex, which is what a copy-paste slip or a later "de-duplicate these two
near-identical lines" edit produces. Every row printed ok, and the release notes would have named
the node twice — **#1138 reintroduced inside #1138's fix.**

Each row now names the identifier its pin must contain, so the arms are bound to each other, not
just to the file.

### Landed — #1138's actual deliverable had zero coverage

The change this PR exists for is the two `printf` lines in `write_manifest`. Nothing asserted them.
The drift guard tests `pin()`, a different function: delete either `printf` and it stays green while
the notes regress. The manifest fixture was already being rendered two assertions above, so
asserting the artefact an operator actually reads cost two lines. A twin-sync conflict on
`release.sh` is a realistic way to lose one of those lines.

### The mutations, re-run against the reworked guard

| mutation | `pin tari-wallet` | `manifest names … CONSOLE WALLET` | other five |
|---|---|---|---|
| **M3** wallet arm re-pointed at the node regex | ✗ | ✗ | all ✓ |
| **M4** the console-wallet `printf` deleted | ✓ | ✗ | all ✓ |

M3 is the one that was green before this rework. M4 is the proof the two assertions are
*independently* load-bearing: `pin tari-wallet` still resolves perfectly, it just is not printed any
more — and only the manifest assertion sees it.

Run in an isolated clone, because the first attempt ran while I switched branches in the shared
worktree and silently tested the wrong tree — the anchor did not match, the mutation never applied,
and the battery still printed a tidy-looking result.

### Refuted

- **"The `tor-base` row's label says Dockerfile with no path."** The stated scenario is refuted by
  execution: under the multi-stage `FROM --platform=...` rewrite it describes, `pv` becomes
  `--platform=$BUILDPLATFORM`, that string *is* in the file, and the row goes **green** — so the
  operator never sees the red line the finding is about. (The reviewer's fix would also mask a real
  hole; the `pin_want` binding now catches that case for real, since `--platform=…` does not contain
  `:`.)

`source "$REL"` in the subshell does not swallow an empty pin — `set +eu` is correctly ordered.
`grep -qF --` is safe against a leading dash and every arm ends in `head -1`. The new `tari-wallet`
arm does not break `scripts/pin-watch.sh`: it iterates its own hardcoded component list and an
unlisted arm is inert. `expect_min … 2` counts in the right render — neither proxy carries a
`profiles:` key, so both appear in every one. And nothing anywhere parses the ingredient list, so the
release-notes format change breaks no consumer.

Issue #1123, opened 2026-08-18
The provisioned machine's HTTP→HTTPS redirect reflects the Host header, the same defect #1118 fixed in the wizard


VijitSingh97:
#1118 fixed this in the first-boot wizard. The same defect is in the **provisioned** machine, which
is the state every appliance spends its life in, and it was confirmed live on real hardware rather
than read from the config.

```
$ curl -sSI -H 'Host: evil.example' http://pithead.local/setup
HTTP/1.1 308 Permanent Redirect
Location: https://evil.example/setup
Server: Caddy
```

A bare-IP header behaves the same way (`Host: 203.0.113.9` → `https://203.0.113.9/`). Caddy's
HTTP→HTTPS redirect is a catch-all, so it echoes whatever host the request claimed.

## Why it matters more here than in the wizard

The wizard runs once. This is the every-day path: an operator who types the machine's name without
a scheme hits `:80` and is redirected — to the dashboard login, where they type the dashboard
password. Reaching it needs a name that resolves to the box (DNS rebinding, or a LAN whose DNS or
DHCP is not trustworthy), which is the same threat model the appliance already accepts by serving
TLS on the LAN at all. Severity is bounded by that, so this is hardening, not an emergency — but
the fix is cheap and the exposure is permanent rather than one-shot.

## Fix

Scope the redirect to the hosts the machine answers to, the same set #1118 settled on: the address
the request arrived on, the machine's hostname/FQDN, and `pithead.local`. In Caddy that is a named
site block for the known hosts plus a catch-all that either refuses or redirects back to the box,
rather than one catch-all that reflects the header.

## Coverage

The Caddyfile render is unit-tested, so the assertion belongs there: a rendered config must not
produce a redirect whose target comes from the request. The live curl above is the tier-4 shape —
worth a line in the post-deploy sweep, since it costs one request and needs no state.

VijitSingh97:
The appliance half landed too: 7f844b3 (PR #1134) on `develop-v2`. The issue auto-closed on the `develop` merge, so this is the record that both lanes carry the fix — with different shapes, deliberately. The appliance keeps the address the operator actually used (its site list is plural) and both of its new `:80` blocks carry the bind; `develop`'s single-host site needs neither.

Adversarially reviewed against real Caddy on the way in, including the attack that mattered: the new `:80` catch-all does **not** shadow the onion vhost — that still wins on host specificity and still returns its 401.

PR #1177, opened 2026-08-21
fix(e2e): prove the live stack's control channel survives a run (#1085, #1067)

### LINKED ISSUE #1067: The e2e borrow flow never cleans up its miner config backups — 30 have accumulated on miner-1
`tests/integration/e2e.sh` backs up the borrowed miner's xmrig config to a timestamped file before repointing it, and restores from that file at the end. It never removes the backup after a **successful** restore, so one accumulates per run. miner-1 currently holds **30**:

```
/opt/rigforge/data/worker/xmrig/build/config.json.e2e-orig.20260816-233037
… 29 more
```

The disk cost is trivial. The reason to fix it is ambiguity: if a restore ever half-fails and someone recovers by hand, they are choosing among thirty files whose only distinguishing feature is a timestamp, and the newest is not necessarily the one holding the true pre-borrow state — a run that borrowed an already-borrowed rig would have captured the *borrowed* config as its "original". Keeping exactly one, or pruning on a proven restore, removes that failure mode entirely.

Suggested shape: after the restore proof passes, delete the backup this run created. Keep it on any failure path, since that is the case where a human needs it. A safety net that nobody prunes stops being a safety net and becomes a haystack.

Found while running the v1.19.0 pre-cut gate; the restore itself was clean and verified (miner-1 back on its original pool, both restore proofs passed).

### LINKED ISSUE #1085: Running the e2e leaves the live stack's control units pointing at the harness checkout
Running the e2e on a box that also hosts a live stack leaves the box-global control-runner units pointing at the harness checkout. The live stack's control channel then goes silently dead: the dashboard writes control requests into the live install's spool, and the path unit watches `/srv/code/pithead-e2e`.

Observed on bench-host while deploying v1.19.1:

```
PathExistsGlob=/srv/code/pithead-e2e/data/control/requests/*.json
WorkingDirectory=/srv/code/pithead-e2e
ExecStart=/srv/code/pithead-e2e/pithead control-run-pending

/srv/code/current -> pithead-v1.19.0     # the live install
systemctl is-active pithead-control.path -> active
```

The unit is enabled and active throughout, which is what makes it hard to spot — nothing reports a fault. Repointing the two unit files at the live install and reloading restored it immediately; the very next control request drained normally.

## Relationship to existing issues

This is the same **failure shape** as #1070, which v1.19.1 just fixed: units aimed at a directory that is not `current`, so requests are written to one spool and watched at another. #1070's cause was a failed upgrade; this one's cause is the test harness, so the fix there does not help here.

#689 is closed and covered the neighbouring case — a sibling checkout *removing* the live stack's units. Its resolution added ownership-checked removal. This is the other half: not removal but **re-provisioning onto the harness path**, which #689 explicitly recorded as a known remaining ceiling ("the FULL e2e hardening phase installs its own units over the live stack's"). That ceiling has now been hit in practice during a release.

Honest uncertainty: I cannot say for certain whether my run or the previous v1.19.0 pre-cut run left them that way, since both ran on this box within a day. The defect is the same either way — after an e2e, the live box's units name the e2e checkout.

## Why it matters beyond a bench annoyance

The release runbook's preferred deploy path is `release-smoke --upgrade`, which drives the real control channel. So the harness breaks precisely the mechanism the next release step depends on, on the same box, minutes earlier. In my case the upgrade would simply have hung had I not checked the units first.

More generally: the control channel is the appliance's only management path (#786). A test run that silently disables it on a live box is worth more than a footnote.

## Suggested fix

Either the harness should restore the units it displaced — it already restores the miner config, the baseline stack, and proves the restore with credential checks, so this belongs in the same teardown — or the e2e should not install box-global units at all when a live stack is present, and instead exercise the runner in-place.

Failing both, the teardown should at minimum *assert* where the units point and fail loudly if they name the e2e checkout, so the operator is told rather than left to notice.

VijitSingh97:
Closes #1085. Closes #1067.

Running the e2e leaves the box-global control units pointing at the harness checkout — or deletes
them. Either way the live dashboard's config edits and one-click upgrades queue into a spool nothing
watches: **enabled, active, and silent**, while every other signal in the restore proof still reads
healthy. It stranded the bench during the mandated v1.19.2 pre-cut run and was found only because
someone went looking.

## The design posted on the issue does not work, and it fails as the defect it was written to fix

This is the part worth reading. The three-comment design says to run **the branch's** `pithead
doctor` with cwd set to the live install dir. `pithead` cd's to the directory of the binary you
invoke (`SCRIPT_DIR`, `pithead:100`), so that evaluates `E2E_DIR`, not the live install.

`check_control_units` compares the installed units' `ExecStart` against `$PWD`. And `pithead-e2e`
does **not** match the `^pithead-v[0-9]+\.[0-9]+\.[0-9]+$` regex that would have made doctor bail
out with "this is not the live install". So:

| box state | units name | what the posted design prints |
|---|---|---|
| **stranded** (the incident) | `/srv/code/pithead-e2e` | `✓ OK Control runner units target this install.` |
| correctly restored | `/srv/code/pithead-v1.19.2` | `✗ FAIL` |

**Inverted.** It reports all-clear on exactly the broken box. And the matching repair line —
`cd <live dir> && <e2e checkout>/pithead apply -y` — would have cd'd to `E2E_DIR` and re-installed
the units pointing at the harness: the fix installing the bug on purpose.

## What this does instead

**The verdict comes from `RESTORE_DIR`'s OWN doctor, run from `RESTORE_DIR`.** No cwd trick, no
branch binary, and it also catches strands this run did not cause.

**No repair line, deliberately.** `restore_all`'s existing `cd "$RESTORE_DIR" && ./pithead apply -y`
already converges the units: v1.19.2's `apply` calls `provision_control_runner` *before* its
no-change return, and that is tier-1 asserted — *"a no-change apply still converges the control
units, and recreates nothing"*. So this leans on a guarded behaviour, not an assumed one. The issue
comment's claim that `apply` short-circuits without converging was true of v1.19.1 and is now stale.

**The oracle is a classifier with a `no-check` arm, not a grep.** A `pithead` predating v1.19.2, or
a box with no systemd, prints no control-channel section at all. Without that arm, that output falls
through to a pass and **"never looked" reads exactly like "looked and it was fine"** — which is the
defect family this milestone exists to close. `no-check` fails the proof rather than passing quietly.

Doctor's exit code is never the oracle: it is 1 on any `dr_fail`, so an unrelated failure elsewhere
would swamp this signal.

## Two more members of the family, same function

**#1067 — the miner-config backups were never pruned.** Every borrowing run minted a timestamped
`.e2e-orig.<stamp>` and nothing removed it. Now pruned, but only after `cmp -s` proves the bytes are
back, **in the same remote call** so a dropped ssh cannot land between proving and deleting.
Deliberately *not* gated on `miner_reload`: that function ends `|| true; return 0` and cannot fail —
gating on it would delete the backup whatever happened, and it is also why the `warn` arm it used to
guard was unreachable dead code.

**A third one, found while checking the loaner — and it is live right now.** `borrow_miner` backs up
whatever is there, and its repoint carries `if any(.pools[]?; .url | contains($b)) then .` — an arm
written for a rig that legitimately keeps a bench pool. That arm also silently absorbs *a previous
run that never restored*: the backup then records the **borrowed** config as "the original", every
later run restores to it, and the contamination is self-perpetuating with nothing reporting a fault.

Measured on the loaner, shapes only:

```
32 backups, 3 distinct contents
  content A  n=26  (single pool, no bench)      2026-07-13 … 2026-08-17
  content B  n=4   (two pools, bench present)   2026-07-11 … 2026-08-18   <- the live config today
  content C  n=2                                2026-07-13 … 2026-07-21
```

The **two most recent** backups already carry a bench pool, so "restore from the newest" returns the
rig to a borrowed state. It is currently mining to its normal pool with the bench as a *failover*, so
the harm is small — but the harness put it there and never took it back. It now warns loudly, before
the backup is taken. The deeper fix (tag the injected pool so restore removes exactly what it added)
is filed separately rather than smuggled into this PR.

**The hardening assertion that could not fail.** `tests/integration/run.sh` asserted only that
something called `pithead-control.path` was enabled. The unit names are box-global, so on a bench
that also hosts a live stack that is **already true before the phase runs** — the live install's
units satisfy it, and it stayed green even if our `apply` installed nothing at all. That is why this
went unnoticed for two releases. It is bound to the `ExecStart` now.

## What was actually RUN

- **`tests/integration/selftest.sh`: 207 passed, 0 failed** — six new assertions bound to doctor's
  literal verdict strings. Tier 2, runs in `make test` on every PR, no bench needed.
- **Two mutations, each isolated, both run:**

| mutation | result |
|---|---|
| collapse the `stranded` arm into `on-target` | the two strand assertions ✗, `no-check` arms ✓ |
| delete the `no-check` header guard | the two `no-check` assertions ✗, strand arms ✓ |

  The second is the one that matters: it proves the classifier distinguishes *"doctor never ran the
  check"* from *"doctor ran it and it passed"*.
- **`make lint-sh`, `bash -n`, `make lint-md`, `make lint-docs-voice`** — clean.

## What was NOT run, and what this PR still owes

**The bench e2e itself has not run against this branch yet.** That is the only thing that exercises
`verify_restore_proof`, the new `run.sh` assertion, and the `borrow_miner` warning on real hardware —
nothing in the repo covers `tests/integration/e2e.sh` below tier 3, so `shellcheck` plus a live run is
the whole coverage story for the changed lines. I will post the run's output here, including whether
the `borrow_miner` warning fires on the loaner's current contaminated config, which is the one
prediction this PR makes that a live run can refute.

**The restore-proof arm is not yet mutation-proven on hardware.** Locally the classifier is; the
`e2e.sh` plumbing that feeds it is not. The bench mutation is to strand the units by hand
(`sed` the two unit files at `E2E_DIR`, `daemon-reload`) and confirm the proof goes red and names the
repair — that is repairable with one `apply`, so it is safe to run.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -214,7 +214,17 @@ What it does, then reverses on exit (even on failure / Ctrl-C, via an `EXIT` tra
    harness-rendered creds while the on-disk `.env` kept the real ones: internally consistent, so
    the stack mined and looked healthy for a day while every host-side RPC probe 401ed. A failed
    proof exits non-zero and names the recovery (`docker compose up -d` from the install dir
-   re-bakes everything from disk).
+   re-bakes everything from disk). The same proof asks the restored install's own `doctor` whether
+   the box-global control-runner units still name it
+   ([#1085](https://github.com/p2pool-starter-stack/pithead/issues/1085)): deploying the branch
+   repoints those units at the e2e checkout, and the hardening phase's teardown removes them when it
+   owns them, so a run can end with the live dashboard's config edits and one-click upgrades queueing
+   into a spool nothing watches — enabled, active and silent. The verdict is read from the install's
+   own `pithead`, run from its own directory, because `pithead` works from the directory of the
+   binary you invoke: the branch's copy would compare the units against the e2e checkout and report
+   all-clear on exactly the stranded box. That needs the live install on v1.19.2 or newer, the
+   release whose `doctor` gained the check; an older one is reported as unproven rather than as a
+   pass.
 
 `--mode`: `targeted` (default, lean) validates the dashboard and the sync logic against the
 already-synced node: `check` + `--lifecycle` (one controlled restart exercises the sync gate /
--- a/tests/integration/e2e.sh
+++ b/tests/integration/e2e.sh
@@ -173,6 +173,14 @@ SAFETY_ARCHIVE=""
 MINER_CFG_BACKUP=""
 RESTORED=0
 RESTORE_PROOF_FAILED=0
+# Separate from RESTORE_PROOF_FAILED on purpose (#1085). That flag's message is the #971
+# credential-bake incident's, and its remediation — "re-bake from disk: docker compose up -d" —
+# does nothing whatsoever for a systemd unit. A control-channel fault needs its own words.
+CONTROL_PROOF_FAILED=0
+# What the control units looked like BEFORE the restore's converging apply, recorded so the run log
+# says whether THIS run stranded the box. The post-restore verdict cannot answer that: it runs after
+# the apply that repairs it.
+CONTROL_VERDICT_BEFORE=""
 # Where the LIVE stack actually runs from — resolved in preflight (#454). Defaults to CANONICAL_DIR
 # so the EXIT trap always has a target even if it fires before preflight refines it.
 RESTORE_DIR="$CANONICAL_DIR"
@@ -205,7 +213,7 @@ restore_all() {
             miner_reload
             ok "$MINER_HOST repointed to its original pool(s); backup pruned"
         else
-            warn "FAILED to restore $MINER_HOST config — backup kept at $MINER_CFG_BACKUP"
+            warn "FAILED to restore $MINER_HOST config — the backup should still be at $MINER_CFG_BACKUP, but check: if the connection dropped after the prune, it is already gone and the live config is the restored one."
         fi
     fi
 
@@ -215,6 +223,16 @@ restore_all() {
     #    project locally-built :dev images.
     step "bringing the baseline stack ($RESTORE_DIR) back up"
     on_bench "cd '$E2E_DIR' && ./pithead down >/dev/null 2>&1 || true"
+    # Look at the control units BEFORE the apply below converges them. Without this the run can
+    # never report that it stranded the box — the post-restore proof runs downstream of its own
+    # repair, so on the ordinary #1085 path it is green either way. Observation only: the strand is
+    # expected here, and the restore is what has to put it right.
+    CONTROL_VERDICT_BEFORE="$(control_units_verdict "$(on_bench "cd '$RESTORE_DIR' && ./pithead doctor 2>/dev/null" || true)")"
+    case "$CONTROL_VERDICT_BEFORE" in
+    on-target) step "control units before restore: already pointing at $RESTORE_DIR" ;;
+    stranded) step "control units before restore: STRANDED (expected — the branch deploy repoints them); the apply below must converge them" ;;
+    *) step "control units before restore: $CONTROL_VERDICT_BEFORE" ;;
+    esac
     if on_bench "cd '$RESTORE_DIR' && ./pithead apply -y >/dev/null 2>&1 && ./pithead up >/dev/null 2>&1"; then
         wait_bench_healthy 300 && ok "baseline stack healthy again" || warn "baseline stack came up but isn't reporting healthy yet — check 'pithead status' on $BENCH_HOST"
         # Proof, even when the health wait timed out: a stack running the WRONG creds looks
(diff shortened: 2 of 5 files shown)


VijitSingh97:
## Live on the bench: the strand reproduced, recorded, and repaired — twice

The PR body owed a real run. Two targeted e2e runs on the bench with a borrowed loaner, ~45 min each.
Every new check did its job, and one of them made a prediction that a live run could have refuted.

**1. `borrow_miner`'s contamination warning fired on its FIRST ever real run** — the prediction this
PR made:

```
! <loaner>'s xmrig config ALREADY names <bench> in 1 pool(s) before this run borrowed it.
!   The backup taken next records THAT state as the original, so a later restore returns to it (#1178).
```

Then it fired **again on run 2**, because run 1's restore faithfully restored the contaminated
"original". #1178 is confirmed self-perpetuating by observation, not inference — recorded there.

**2. The strand was reproduced mid-run.** While the harness was working, the box-global units really
did read:

```
ExecStart=/srv/code/pithead-e2e/pithead control-run-pending
```

That is #1085, live, on a box whose live install is `/srv/code/pithead-v1.19.2`.

**3. `CONTROL_VERDICT_BEFORE` caught it** — the arm added after review, precisely because the
post-restore check runs downstream of its own repair:

```
→ control units before restore: STRANDED (expected — the branch deploy repoints them);
  the apply below must converge them
```

Without this line the run log could never have said that the run stranded the box. It says it now,
on both runs.

**4. The post-restore proof passed, including the `is-enabled` half** added after review:

```
✓ restore proof: the control runner units point at /srv/code/pithead-v1.19.2, and the path unit is enabled
```

**5. #1067's prune worked:**

```
✓ <loaner> repointed to its original pool(s); backup pruned
```

**Box end state, checked directly rather than inferred from the harness:** units name
`/srv/code/pithead-v1.19.2`, `pithead-control.path` is `active`, and `current` still points at
`pithead-v1.19.2`.

## What run 1 also found

Run 1 exited 1 — not on anything in this PR, but on `✗ Caddyfile uses correct scheme`, 8 times, on a
healthy box. That assertion reads line 1 of the Caddyfile and #1123 stopped line 1 being the site
block three days ago. Filed as #1180, fixed in #1181, and run 2 is the green re-run: **632 ✓, 0 ✗.**

So the honest summary of this PR's live evidence is: the control-channel work behaved exactly as
designed on real hardware, and running it turned up a *different* false RED sitting on the mandated
release gate.

Issue #1082, opened 2026-08-17
The post-deploy check reports two failures on a healthy box, and the runbook says to accept them


VijitSingh97:
The post-deploy sweep reports failures on a completely healthy box, and the runbook tells you to accept them. That is a red result which means nothing is wrong, on the check we run immediately after every deploy.

From the v1.19.1 deploy, on a stack that was fully working:

```
[ITEST] passed:  62
[ITEST] skipped: 0
[ITEST] failed:  2

    - check: workers online (>= 2)
    - check: stratum total hashes > 0
```

Both fail for one reason: no miners were attached. The e2e had already handed miner-1 back to its own pool, which is the correct end state. Nothing was broken.

## Why this is worth fixing rather than documenting

It is already documented — the release runbook says that on a parked bench "exactly two failures are expected and acceptable ONLY post-restore". That instruction is the problem. It asks a human to hold in their head which two of the red lines are the harmless ones, every time, and to notice when a *third* appears or when a *different* one substitutes for one of the pair. That is exactly the check a tired person skips at the end of a release.

I did it myself an hour after reading the runbook: I saw `failed: 2`, matched the names against the expected pair, and moved on without confirming there wasn't a third. The count happened to be right. It would have been just as easy for it not to be.

It also makes the exit code useless. `run.sh --check` cannot be wired into anything automated, because its healthy state is non-zero.

## What it should do instead

The condition is detectable, not a matter of taste: the harness already knows whether any worker is connected. With no miners attached, both assertions should **skip with a reason** — "no miners attached to this bench; worker/hashrate assertions need a rig" — and the run should exit 0.

That way a red result means something is wrong, `failed: 2` never becomes background noise, and the parked-bench caveat can come out of the runbook entirely.

If skipping is considered too generous, the alternative is a flag (`--expect-no-miners`) that makes the intent explicit at the call site rather than in prose someone has to remember. Skipping on a detected condition is better: it cannot be forgotten.

Related: #1069 already found assertions that cannot fail on the defect they guard. This is the mirror image — assertions that fail when there is no defect.

VijitSingh97:
[...]

PR #1181, opened 2026-08-21
fix(e2e): the Caddyfile scheme assertion reads line 1, which #1123 stopped being the site block

### LINKED ISSUE #1180: The e2e's Caddyfile scheme assertion reads line 1, which stopped being the site block three days ago — a false RED on the mandated release gate
The tier-4 harness asserts the Caddy scheme by reading **line 1** of the rendered Caddyfile:

```bash
assert_contains "Caddyfile uses correct scheme" "$(rx 'head -n1 Caddyfile 2>/dev/null')" "$scheme"
```

Since #1123 (2026-08-18) line 1 is not the site block. That fix stopped Caddy's `:80` redirect
trusting the `Host` header by emitting a global options block, and in HTTPS mode an `http://`
redirect block **above** the dashboard's own block:

```
{
    auto_https disable_redirects
}

http:// {
    redir https://<host>{uri} 308
}

https://<host> {
```

Line 1 is `{`. The assertion has been reading a brace and looking for a scheme in it ever since.

## How it surfaced

Running the mandated pre-release e2e for the v1.19.3 milestone. Every other assertion in the run
passed, the restore proved clean, and the run still exited 1:

```
✓ stratum TLS handshake succeeds on the published port (#261/#942)
✓ live-served cert fingerprint matches the one rigs are told to pin (#261/#942)
✓ PAYOUT_CONFIRM_ENABLED matches config (#381/#942)
✗ Caddyfile uses correct scheme            <- 8 times, on a healthy box
✗ E2E FAILED for 'develop' (harness exit 1)
```

The box was serving HTTPS correctly the whole time. Confirmed by reading the file it was
complaining about.

## Why this one matters more than an ordinary flake

This is the gate that is **mandated before every release cut**. An assertion that is red on a
correct box teaches the operator that a red e2e is normal, on the one check standing between a bad
release and the world. Same family as #1082.

It also went unnoticed for three days across a merged product fix, because nothing ties the
assertion to the renderer that produces what it reads.

## The naive repair is also wrong

"Use the first scheme in the file instead of line 1" fails differently: the first scheme **is**
`http://`, the redirect block, on a box correctly serving HTTPS. That spelling turns a false RED for
`secure=true` boxes into a false GREEN for `secure=false` ones.

The dashboard's site block is the one whose address carries a **host** after the scheme. The bare
redirect is `http:// {` — a space where the host would be — so `[^ ]` separates them. Verified
against the real rendered file on the bench: `grep -cE '^https://[^ ]'` → 1,
`grep -cE '^http://[^ ]'` → 0.

## Fixed in

PR for this issue, on the v1.19.3 milestone, because the release gate cannot go green without it.
The proof is the re-run of the same e2e, not the diff.

## Worth considering separately

Nothing binds this assertion to the renderer. A tier-1 check that the rendered Caddyfile's dashboard
site block is findable by the same expression the tier-4 harness uses would have caught #1123's
change on the PR that made it, rather than three days later on a release gate. Filing that thought
here rather than as its own issue until someone wants it.

Related: #1123 (the renderer change), #1082 (always-red post-deploy assertions), #1069 (the
checks-that-cannot-fail cluster — this is its mirror image, a check that cannot pass).

VijitSingh97:
Closes #1180.

**Found by running the mandated pre-release e2e for v1.19.3.** Every other assertion passed, the restore proved clean, and the run still exited 1:

```
✓ stratum TLS handshake succeeds on the published port (#261/#942)
✓ live-served cert fingerprint matches the one rigs are told to pin (#261/#942)
✓ PAYOUT_CONFIRM_ENABLED matches config (#381/#942)
✗ Caddyfile uses correct scheme            <- 8 times, on a healthy box
✗ E2E FAILED for 'develop' (harness exit 1)
```

The box was serving HTTPS correctly throughout.

## Why

#1123 (2026-08-18) stopped Caddy's `:80` redirect trusting the `Host` header by emitting a global options block, and in HTTPS mode an `http://` redirect block **above** the dashboard's own site block:

```
{
    auto_https disable_redirects
}

http:// {
    redir https://<host>{uri} 308
}

https://<host> {
```

`head -n1 Caddyfile` is `{`. The assertion has been reading a brace and looking for a scheme in it for three days.

## The obvious repair is also wrong

"Use the first scheme in the file" finds `http://` — the redirect block — on a box correctly serving HTTPS. That turns a false RED for `secure=true` into a false GREEN for `secure=false`.

The dashboard's block is the one whose address carries a **host** after the scheme. The bare redirect is `http:// {`, a space where the host would be, so `[^ ]` separates them.

## What was actually RUN

Against the **real rendered Caddyfile on the bench**, not a fixture:

```
grep -cE '^https://[^ ]' Caddyfile   -> 1     (the dashboard block)
grep -cE '^http://[^ ]'  Caddyfile   -> 0     (the bare redirect correctly excluded)
grep -nE '^http://'      Caddyfile   -> 5:http:// {
```

So on this `secure=true` box the new expression answers yes for https and would answer no for http; the old one answered ✗ for both.

`make lint-sh` and `bash -n` clean.

## What is NOT yet proven

**The re-run.** The proof of a harness fix is the harness going green, and the e2e that found this is the one that has to say so. I am re-running the targeted e2e against `develop` with the borrowed miner after this merges, and will post the result here — including whether anything else was hiding behind this failure, since an exit-1 run stops being informative after its first red.

## Why it belongs in v1.19.3 rather than later

This is the gate mandated before every release cut. An assertion that is red on a correct box teaches the operator that a red e2e is normal, on the one check standing between a bad release and the world — #1082's family. v1.19.3 cannot be cut on a gate that cannot go green.

### DIFF
--- a/tests/integration/run.sh
+++ b/tests/integration/run.sh
@@ -793,10 +793,18 @@ assert_running_state() {
         assert_eq "dashboard confirms Tari payout tracking is live (#462/#942)" "$(jq_get "$st" '.earnings.tari_confirmed.enabled')" "true"
     fi
 
-    # 9. Caddy scheme matches dashboard.secure.
-    local scheme
-    [ "$secure" = "false" ] && scheme="http://" || scheme="https://"
-    assert_contains "Caddyfile uses correct scheme" "$(rx 'head -n1 Caddyfile 2>/dev/null')" "$scheme"
+    # 9. Caddy scheme matches dashboard.secure — read from the DASHBOARD's site block, which is
+    #    neither line 1 nor the first scheme in the file. #1123 put a global options block
+    #    (`{ auto_https disable_redirects }`) at the top, and in HTTPS mode an `http:// {` redirect
+    #    block ABOVE the dashboard's own block. So `head -n1` sees `{`, and "the first scheme in the
+    #    file" sees `http://` on a box that is correctly serving HTTPS — both spellings report a
+    #    false RED on a healthy stack, which is how this assertion failed 8 times in one green run.
+    #    The dashboard's block is the one whose address carries a HOST after the scheme; the bare
+    #    redirect is `http:// {`, with a space where the host would be, so `[^ ]` separates them.
+    local want_scheme
+    [ "$secure" = "false" ] && want_scheme='^http://[^ ]' || want_scheme='^https://[^ ]'
+    assert_eq "Caddyfile's dashboard site block uses the correct scheme (#1123)" \
+        "$(rx "grep -qE '$want_scheme' Caddyfile && echo yes || echo no")" "yes"
 
     # 10. Secrets intact (proxy token + onions unchanged vs the baseline we captured).
     assert_eq "secrets intact (token + onions)" "$(secret_fingerprint)" "$BASELINE_SECRET_FP"


VijitSingh97:
## Re-run: 632 assertions, 0 failures — the owed proof

The PR body said the proof of a harness fix is the harness going green, and that the e2e which found
this had to be the one to say so. It does.

| | run 1 (before) | run 2 (after) |
|---|---|---|
| ✓ | 619 | **632** |
| ✗ | **9** | **0** |
| verdict | `✗ E2E FAILED for 'develop' (harness exit 1)` | `✓ restore complete.` |

The fixed assertion ran **10 times, all green**:

```
✓ Caddyfile's dashboard site block uses the correct scheme (#1123)
```

Run 2 has 13 *more* passing assertions than run 1, not merely 9 fewer failures — an exit-1 run stops
being informative after its first red, so run 1 was also skipping work. That is the second cost of a
false RED, and it is the one that hides real regressions behind a known-bad assertion.

Both runs were targeted mode against `develop` on the bench with a borrowed loaner rig, same box,
same chains, ~45 min each.

**Nothing was hiding behind the failure.** That was the open question in the PR body, and the answer
is clean.

VijitSingh97:
## Correction: the assertion counts in this PR are glyph counts, not the harness's tally

Flagged by the release-prep session, verified against my own logs, and it is a fair hit.

I reported run totals by counting lines containing a check glyph. That sweeps in the e2e **wrapper's**
own step lines (preflight, borrow, restore proofs) alongside the streamed harness assertions. The
harness prints its own `[ITEST]` summary, and that is the authoritative number.

I also missed that summary at first because the detached harness's output is **indented two spaces**,
so an anchored `^\[ITEST\]` grep matches only the *inline* pre-run block (`passed: 6`) and silently
skips the real one.

**The corrected table:**

| | glyph count (what I published) | `[ITEST]` tally (authoritative) |
|---|---|---|
| run 1 | 619 ✓ / 9 ✗ | **597 passed / 5 skipped / 8 failed** |
| run 2 | 632 ✓ / 0 ✗ | **608 passed / 5 skipped / 0 failed** |

Two things change, and one does not:

- **The 5 skips were invisible in my reporting.** I never mentioned them, which reads as fuller
  coverage than the run actually gives. They are the known permanent e2e skips, not new.
- **"13 more passes in run 2" was 11**, and the failure count was 8, not 9 — my 9th ✗ was the
  wrapper's own `E2E FAILED` line, not an assertion.
- **The conclusion is unchanged**: 8 real assertion failures before the #1180 fix, 0 after, and run 1
  genuinely completed less work because it aborted scenarios after its first reds.

The general rule worth keeping, in the release-prep session's words: **totals move with bench state;
the failure count and the verdict line are the arbiters.** Their RC run was 607/5/0 against my
608/5/0 — a one-assertion difference that means nothing, and would have looked like a regression if
either of us had treated the total as a fixed number.

Nothing in the merged diff changes. This corrects the evidence I wrote about it.

PR #1182, opened 2026-08-21
sync: develop -> develop-v2 (the v1.19.3 wave)

VijitSingh97:
Routine twin sync carrying #1176, #1177, #1179 and #1181. Clean merge, no conflicts.

**Verified by reading the merged tree, not the exit code** — a merge that silently drops one lane's change reports success identically:

| | in the merged tree |
|---|---|
| `scripts/release.sh` | `tari-wallet` arm present; notes name both Tari images |
| `tests/stack/run.sh` | 5 pin drift rows, each bound to its own image name |
| `tests/stack/test_compose.sh` | socket-proxy asserted on **both** proxies (`expect_min … 2`) |
| `build/tor/entrypoint.sh` | `TORRC_OUT` seam, container default still `/tmp/torrc` |
| `tests/integration/lib.sh` | `control_units_verdict` classifier |
| `tests/integration/run.sh` | the Caddyfile scheme fix (#1180) |
| `os/rootfs/Dockerfile` | **develop-v2's own** compose `v5.5.0` pin and debian digest `3a39a059` both survive |

That last row is the one worth checking on every sync: the appliance lane's changes are not reverted by taking develop's side anywhere.

## Evidence, on the bench

The harness changes in this wave were proven with two targeted e2e runs against `develop` with a borrowed loaner rig:

- **run 1** — 619 ✓ / 9 ✗, exit 1. Found #1180 (a false RED that had been sitting on the mandated release gate for three days) and reproduced the #1085 strand live.
- **run 2**, after #1181 — **632 ✓ / 0 ✗**, `✓ restore complete.`

## Not run

No appliance work: this sync touches no `os/` file except by carrying develop's side, which the table above confirms did not disturb the two appliance-only pins. The KVM battery is not implicated by a CI/harness wave.

### DIFF
--- a/build/tor/entrypoint.sh
+++ b/build/tor/entrypoint.sh
@@ -9,8 +9,15 @@ set -eu
 # TORRC_TEMPLATE is a test seam so the shell suite can render against build/tor/torrc.template; the
 # container always uses the baked-in default.  # ponytail: default unchanged, only overridden in tests
 : "${TORRC_TEMPLATE:=/etc/tor/torrc.template}"
+# TORRC_OUT is the matching seam for the RENDERED file (#1104). Without it the shell suite's tor
+# stub reads a host-global /tmp/torrc, which is the only unsandboxed path in an ~8000-line suite
+# where every other fixture gets its own mktemp -d — so two concurrent runs race on one file and
+# produce a false RED in the hidden-service assertions, whose natural remedy is "re-run until green".
+# The DEFAULT MUST STAY /tmp/torrc: tier-3 assertions in tests/integration/run.sh read this path
+# inside the running container.  # ponytail: default unchanged, only overridden in tests
+: "${TORRC_OUT:=/tmp/torrc}"
 
-sed "s/__NETWORK_PREFIX__/${NETWORK_PREFIX}/g" "$TORRC_TEMPLATE" >/tmp/torrc
+sed "s/__NETWORK_PREFIX__/${NETWORK_PREFIX}/g" "$TORRC_TEMPLATE" >"$TORRC_OUT"
 
 # Node inbound hidden services (#103): the bundled monerod and Tari containers only start under
 # their compose profiles, so publish each node's onion only then — in remote mode the onion would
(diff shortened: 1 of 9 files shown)


VijitSingh97:
## Correction: the assertion counts in this PR are glyph counts, not the harness's tally

Flagged by the release-prep session, verified against my own logs, and it is a fair hit.

I reported run totals by counting lines containing a check glyph. That sweeps in the e2e **wrapper's**
own step lines (preflight, borrow, restore proofs) alongside the streamed harness assertions. The
harness prints its own `[ITEST]` summary, and that is the authoritative number.

I also missed that summary at first because the detached harness's output is **indented two spaces**,
so an anchored `^\[ITEST\]` grep matches only the *inline* pre-run block (`passed: 6`) and silently
skips the real one.

**The corrected table:**

| | glyph count (what I published) | `[ITEST]` tally (authoritative) |
|---|---|---|
| run 1 | 619 ✓ / 9 ✗ | **597 passed / 5 skipped / 8 failed** |
| run 2 | 632 ✓ / 0 ✗ | **608 passed / 5 skipped / 0 failed** |

Two things change, and one does not:

- **The 5 skips were invisible in my reporting.** I never mentioned them, which reads as fuller
  coverage than the run actually gives. They are the known permanent e2e skips, not new.
- **"13 more passes in run 2" was 11**, and the failure count was 8, not 9 — my 9th ✗ was the
  wrapper's own `E2E FAILED` line, not an assertion.
- **The conclusion is unchanged**: 8 real assertion failures before the #1180 fix, 0 after, and run 1
  genuinely completed less work because it aborted scenarios after its first reds.

The general rule worth keeping, in the release-prep session's words: **totals move with bench state;
the failure count and the verdict line are the arbiters.** Their RC run was 607/5/0 against my
608/5/0 — a one-assertion difference that means nothing, and would have looked like a regression if
either of us had treated the total as a fixed number.

Nothing in the merged diff changes. This corrects the evidence I wrote about it.

PR #1200, opened 2026-08-21
fix(control): the unit-install leg refuses to overwrite a foreign install's units

### LINKED ISSUE #1190: provision_control_runner's install leg overwrites a foreign checkout's control units with no ownership check
## What is wrong

`provision_control_runner`'s **install** leg writes the box-global `pithead-control.{service,path}` units with `sudo tee` whenever the existing unit's `ExecStart` does not name the current checkout. The `grep -qsF "ExecStart=$PWD/pithead control-run-pending"` before the tee is an idempotence skip, not an ownership check: a unit pointing at a **different** install is overwritten without complaint.

For the live install converging its own units, that overwrite is the desired behaviour (the post-restore proof depends on it). The hazard is every *sibling* checkout on the same box — an e2e checkout, a bundle-smoke tmp dir, a disposable install — whose `apply`/`up` steals the box-global units from the live stack. That is the exact mechanism behind the production control-channel stranding: a failed upgrade left the units repointed and nothing reported it. The removal leg got an ownership check back then; the install leg never did. Verified present in the released `pithead` at v1.18.1, v1.19.0 and v1.19.2, and on current `develop`.

`PITHEAD_UNIT_DIR` is the existing seam that avoids it — exporting a scratch dir kept two disposable installs from touching the live units while their one-click upgrades were driven end to end on the bench (2026-08-21). But it is opt-in and easy not to know about; the default remains silent theft.

Residual even with the seam: the enable step (`systemctl enable --now pithead-control.path`) still acts on the box-global unit *name*, so a disposable's run re-enables the live path unit — harmless when it is already enabled, but it means the seam isolates the files, not the systemd verbs.

## Sketch of a fix

The install leg gets the same shape the removal leg already has: before overwriting a unit whose `ExecStart` names a different existing directory, refuse and say which install owns it, with an explicit `--steal-control-units` (name illustrative) escape for the legitimate takeover cases (upgrade repoint, operator migration). An unparseable `ExecStart` stays fail-safe: leave it alone, say so.

## Why now

Driving the upgrade escape path on disposable installs required reading the release code at three tags to establish whether the bench's live stack would survive the test. The answer should have been "yes, by default", not "yes, if you know the env var".

VijitSingh97:
Closes #1190.

The install leg of `provision_control_runner` gets the ownership check the removal leg has had since the disable-apply incident. The idempotence grep stays what it was — a skip; ownership is decided by `control_units_owner_dir`, the same physical-path resolution the removal branch uses, so the `current`-symlink/versioned-dir two-spellings case behaves identically on both legs.

Decision table, matching the sketch on the issue:

| Existing service unit says | Action |
| --- | --- |
| this install (either spelling) | converge (rewrite in place — the post-restore proof depends on it) |
| a different install that still exists on disk | **refuse**, name the owner, point at the escape |
| a different install whose directory is gone | adopt (how a new version takes over from a removed one) |
| an ExecStart this tool did not write | leave alone, say so (fail-safe, like the removal leg) |
| anything, with `PITHEAD_STEAL_CONTROL_UNITS=1` | install (deliberate takeover: migration, failed-upgrade repair) |

A refusal warns and returns 0 — an `apply` on a sibling checkout should not abort over units it was never entitled to; the warn says dashboard config changes stay unapplied until a runner watches that install.

Not addressed, deliberately: the issue's residual about `systemctl enable --now` acting on the box-global unit *name* even under `PITHEAD_UNIT_DIR` — the seam isolates files, not systemd verbs. That stays as described on #1190's body; this PR is the refusal the issue sketches.

Docs: the `docs/operations.md` control-units passage gains the install-side paragraph next to the existing removal-side one.

## What was RUN

- `make test-stack`: 1835 passed, 0 failed — including 7 new assertions covering every row of the table above (foreign-existing refused + named, steal-escape overwrites, gone-owner adopted, unparseable left alone ×2, own-drift converged, fresh install unaffected).
- Mutation proofs, run for real (each named in the test block's comment):
  - drop the `[ -d "$install_owner" ]` conjunct → "foreign owner whose directory is gone -> adopted" went red (1 failed / 1834 passed);
  - flip the `!=` ownership compare to `=` → "foreign existing owner -> refused, names the owner" went red (3 failed — the flip also breaks the adjacent cases, as it should);
  - drop the `PITHEAD_STEAL_CONTROL_UNITS` conjunct → "steal escape -> overwritten" went red (1 failed).
  Guard restored after each run; the committed tree is the unmutated one.
- `make lint`: pass (docs voice included).

### DIFF
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -168,10 +168,12 @@ Installation is ownership-checked the same way: when the units already name a di
 that still exists on disk, `apply` refuses to overwrite them and names the owning directory — a
 sibling checkout turning the flag on can no longer repoint the live stack's runner at itself,
 which would leave the live dashboard's requests sitting unread with no error on either side.
-Units whose install directory is gone are adopted (how a new version takes over from a removed
-one), a unit this tool did not write is left untouched, and setting
-`PITHEAD_STEAL_CONTROL_UNITS=1` forces the takeover when it is deliberate — migrating an install
-by hand, or repairing after a failed upgrade.
+Units whose install directory is gone are adopted, a unit this tool did not write is left
+untouched, and a successful one-click upgrade takes the units over as part of moving `current`
+(the previous version's directory stays on disk as the rollback, so the upgrade is the one
+takeover that must not be refused). For the remaining deliberate cases — migrating an install by
+hand, or repairing after a failed upgrade — setting `PITHEAD_STEAL_CONTROL_UNITS=1` forces the
+takeover.
 
 The runner dispatches a fixed set of actions and rejects everything else: `apply --dry-run
 --porcelain` (preview), `apply -y` (commit), a release upgrade — the dashboard's
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -164,6 +164,15 @@ off only removes units whose `ExecStart` points at itself, comparing physical pa
 checkout on the same box (an e2e harness, a bundle smoke test) therefore cannot delete the live
 stack's runner and strand its queued requests.
 
+Installation is ownership-checked the same way: when the units already name a different install
+that still exists on disk, `apply` refuses to overwrite them and names the owning directory — a
+sibling checkout turning the flag on can no longer repoint the live stack's runner at itself,
+which would leave the live dashboard's requests sitting unread with no error on either side.
+Units whose install directory is gone are adopted (how a new version takes over from a removed
+one), a unit this tool did not write is left untouched, and setting
+`PITHEAD_STEAL_CONTROL_UNITS=1` forces the takeover when it is deliberate — migrating an install
+by hand, or repairing after a failed upgrade.
+
 The runner dispatches a fixed set of actions and rejects everything else: `apply --dry-run
 --porcelain` (preview), `apply -y` (commit), a release upgrade — the dashboard's
 [Upgrade button](dashboard.md#upgrading-from-the-dashboard), for which the runner re-derives the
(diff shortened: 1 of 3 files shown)


VijitSingh97:
Adversarial self-review before merge found a regression in the first commit, fixed in the second (`f5bdadc`): after a successful one-click upgrade the OLD versioned dir still exists — it *is* the rollback — so the post-upgrade converge from the new dir matched the guard's "foreign owner that exists on disk" case and refused, stranding the control channel on every normal upgrade. The upgrade tests stub the provisioner, so the suite was green through the hole.

The upgrade callsite now passes a `steal` argument (that dir just became `current`, so the takeover is definitionally legitimate); the operator env escape stays for manual cases.

Mutation proofs re-run against the shipped guard spelling:

| mutation | expected red | result |
| --- | --- | --- |
| drop the `steal`-argument conjunct | upgrade repoint (steal arg) → overwritten | went red (1 failed / 1835 passed), restored clean |
| neutralize the env-escape conjunct | steal escape → overwritten | went red (1 failed / 1835 passed), restored clean |

Together with the three proofs already in the body (run at the first commit; the `-d` and `!=` lines are unchanged), every conjunct of the guard has a run-red proof.

Deliberately not covered by tests: the callsite wiring itself rides the upgrade flow, whose tests stub `provision_control_runner` — the honest proof of the wired path is the next real one-click upgrade on the bench, and this PR does not claim it.

Issue #1062, opened 2026-08-16
BLOCKER: pithead-data-reset erases a RECOVERABLE /data — wallets, onion keys and chains destroyed, box returns looking factory-fresh


VijitSingh97:
`os/overlay/pithead-data-reset` reformats a `/data` partition that a standard `e2fsck` invocation recovers intact. The file's own header claims the opposite:

> a fsck pass runs first so a transient/dirty filesystem is repaired rather than erased. Fail safe — never wipe a recoverable `/data`.

That claim is false for exactly the damage class `fsck -p` refuses to touch.

## Reproduced end to end

On a 512 MiB ext4 filesystem holding one file, corrupting the magic with the **same two-byte write the battery itself injects** (`tests/os/run.sh:3182`, `dd bs=1 seek=1080 count=2`):

```
$ fsck -p -t ext4 /dev/loop0
fsck -p exit=8          # repaired nothing
    …try running e2fsck with an alternate superblock:
    e2fsck -b 32768 <device>

$ mount -o ro /dev/loop0 /mnt
STILL WEDGED            # -> data_mountable returns 1 -> decision "reformat-wedged" -> mkfs.ext4 -F

$ e2fsck -y -b 32768 /dev/loop0
e2fsck exit=1           # recovered
$ mount -o ro /dev/loop0 /mnt && cat /mnt/wallet.txt
IRREPLACEABLE-WALLET-KEY-MATERIAL
```

`fsck -p` **prints the command that would have saved the data** and pithead runs `mkfs.ext4 -q -F -L data` instead.

## Why `fsck -p` is the wrong tool here

`-p` is preen mode: by definition it repairs only what needs no operator decision and refuses everything else. So the precise class of damage a full repair recovers is the class that reaches the reformat branch. A corrupt primary superblock is the textbook case — recoverable from a backup superblock, and never recoverable by `-p`.

## Blast radius

`/data` is everything that survives an A/B update: `config.json`, the wallets, the **Tor onion private keys**, the dashboard database, `backups/`, and both synced chains. The unit is enabled and unconditional, and the mount generator writes `data.mount` with no `passno`, so this is the only repair the partition ever gets.

Afterwards the box boots into the setup wizard looking factory-fresh — so the operator sees "it reset itself", not "it destroyed recoverable data", and has no reason to suspect the disk was salvageable.

## Both tiers pass on this, and one of them enshrines it

[...]

PR #1205, opened 2026-08-21
feat(appliance): surface the data-wipe note in the wizard and doctor

### LINKED ISSUE #1121: Nothing reads the /data wipe note, so a shell-less box still cannot tell the operator it was reinitialized
Follow-up to #1062, which makes `pithead-data-reset` record a dated line on the ESP
(`/boot/efi/pithead-data-wiped`) when it has to reinitialize `/data`. That closes the "a wiped box
and a factory-fresh box look identical" gap at the *recording* end. Nothing reads it yet.

On the shipping configuration that matters: a release image has no shell, so the one person who
needs the fact — the operator staring at a setup wizard they did not ask for — has no way to reach
a file on the ESP.

## What to do

Surface it where the operator already is:

- **The wizard.** When the note is present, the first-boot page should say the data area was
  reinitialized on that date and that this was a recovery, not a fresh machine. That is also the
  moment they would decide to restore from a backup instead of starting over.
- **`pithead doctor`.** One line reporting the note and its date, so a support conversation on a
  box with a shell gets the fact without knowing where to look.

Both read the same file, so the shape is a small reader plus two callers.

## Why it is worth doing

The wipe is the one event where an operator's next action depends entirely on knowing it happened:
restoring a backup recovers the wallets and onion keys, while walking through the wizard starts a
multi-day resync and leaves the old identity gone. Today they are not told, so the default path is
the wrong one.

VijitSingh97:
Closes #1121.

The wipe note #1062 records on the ESP finally gets read — at the two places the issue names. `data_wipe_note()` parses the log's last line into `{when, reason, recovery}`; `recovery` is false only for a deliberate `factory-reset requested`, true for the wedged-`/data` case where restoring a backup is the right next move.

- **Wizard:** the container can't read the ESP (its only mount is the spool), so `publish_data_wipe_note()` — modeled on `publish_rig_defaults()` — stages the note into the wizard spool from `stage_wizard_spool()`. It always writes (`{}` when absent) so a fleet stick never hands machine 2 machine 1's note, and skips removable boot media, where the "ESP" is the stick's own. The setup page shows a banner directly above the restore-from-backup button, and only when `recovery` is true — a factory reset the operator asked for shows nothing.
- **Doctor:** appliance-only line — WARN with the date for a recovery wipe (pointing at restore), INFO for a deliberate reset, silent when absent, and **never FAIL**: doctor is half the boot health gate, and a historical note must not be able to fail an A/B commit.

## What was RUN

- `make test-stack`: 2754 passed, 0 failed (22 new/extended assertions across the reader, the publisher, the doctor check, and both existing spool-staging loops).
- `make test-dashboard`: 2477/2477, 96.81% total; patch coverage 94% (≥90% gate) — `wizard.py`'s new lines covered.
- `make lint`: green across every surface.
- Eight mutations, each run red and restored (standalone harness sourcing the real `pithead`): recovery discrimination, last-line-only, garbage-line guard, the fleet-stick always-write (the exact bug the issue flags — red on 2 assertions when reverted to skip-on-absent), the removable-media gate, WARN-vs-INFO, the `is_appliance` gate (needed an extra off-appliance-with-note case to be provable), and the spool wiring.

## Not done, stated plainly

- The frontend's `dataWiped.recovery` render gate has no JS test in this diff — testing was scoped to the reader/logic seam per the four-tier model; the banner's logic lives host-side in the note's `recovery` field, which is covered.
- The issue's later-comment refinement (clearing the note on `config-reset` so it cannot resurface against an unrelated future reset) is not here — it touches `os/overlay/pithead-data-reset` itself; if wanted, it's a follow-up.
- No image build, no KVM battery — this change touches no `os/` files; it rides the combined battery run with the rest of tonight's appliance wave.


VijitSingh97:
## Adversarial verification (owed since the pre-crash session; two earlier attempts died with the host)

**Verdict: pass.** Fresh-context verifier, sonnet, diffed the PR's own changes against its merge-base. What was actually RUN:

- Extracted the PR's exact `pithead` script via `git show origin/fix/1121-wipe-note-surfacing:pithead` and sourced it directly (not the worktree's stale develop copy) to drive check_data_wipe_note/data_wipe_note/publish_data_wipe_note by hand across appliance-flag / absent / recovery-wipe / factory-reset / garbage-line / stale-note states.
- Confirmed by reading source that check_data_wipe_note only ever calls dr_warn or dr_info, never dr_fail, and DR_FAIL is a global counter incremented only inside dr_fail -- so this check is structurally incapable of driving doctor's FAIL/reboot path for any input.
- Mutation: swapped dr_warn -> dr_fail in check_data_wipe_note's recovery branch, then ran the PR's own 9 doctor assertions (extracted verbatim from tests/stack/run.sh) against the mutated script: 2 went RED ('a recovery wipe -> WARN' and '...never a FAIL') vs 9/9 green on the unmutated script -- proves the never-FAIL test is load-bearing.
- Ran the PR's own reader/publisher assertions (data_wipe_note, publish_data_wipe_note, extracted verbatim) against the unmutated PR script: 7/7 green.
- Mutation: reverted publish_data_wipe_note to the exact fleet-stick bug the issue warns about (`note=$(data_wipe_note) || return 0` instead of always-write `{}`), reran the same 7 assertions: 2 went RED (missing spool file; a previous machine's note surviving onto the next machine) -- proves that regression test is real, not vacuous.
- Read wizard.mjs: the `dataWiped.recovery && html...` banner is an unconditional sibling paragraph in renderSetup(), not gated behind any disk-pick/role branch, sitting directly above the existing restore-from-backup link; state is repopulated from /api/wizard-state on every poll from a spool file nothing in wizard.py ever deletes -- visible on first paint, nothing clears it mid-wizard.
- gh pr checks 1205 -- confirmed 16/16 green including the CI job that runs tests/stack/run.sh.
- Read issue #1121's two review comments (gh api repos/.../issues/1121/comments) against the PR body's 'Not done' section to check for undisclosed scope gaps.

**Findings** (both filed as a follow-up issue rather than blocking: the first is already disclosed in this PR's body as deferred; the second may be unreachable on a vfat ESP):
- **minor**: Design gap (disclosed in the PR body as deferred, not hidden): the wipe note is read from an append-only, never-cleared ESP log with no one-shot/ack marker, so once a machine has ever had one recovery wipe in its history, doctor WARNs on the SAME stale date on every subsequent doctor run forever, and a later config-reset reopens the wizard with the same 'restore from backup' banner even on a machine that has been healthy and in normal use since. This is exactly the concern issue #1121's own second review comment raised ('a config-reset years later reopens the wizard and resurfaces it ... worse than saying nothing') and proposes a cheap `.pending`-marker fix for; the PR explicitly defers it as follow-up work touching os/overlay/pithead-data-reset. Demonstrated directly (not just read): with a synthetic 2026-01-01 recovery-wipe entry as the log's only line, check_data_wipe_note still emits the identical WARN today.
  - demonstrate: `printf '2026-01-01T00:00:00Z unrecoverable /data reinitialized — everything on it was lost\n' > $PRESEED_DIR/pithead-data-wiped && PITHEAD_APPLIANCE=1 (source pithead; check_data_wipe_note) # still WARNs, months/years later, identically`
- **note**: Issue #1121's second review comment recommends doctor split the two rc-1 cases from data_wipe_note ('file absent → silent, but PRESEED_DIR present and the note unreadable → say so, because the ESP mount is root-only and a non-root doctor would otherwise report clean'). The shipped check_data_wipe_note does not implement this split -- any rc 1 (absent file, permission-denied, or a garbage/unparseable line) is treated identically as silent. Neither the PR body's 'Not done' list nor its diff mentions this specific omission. Severity is uncertain rather than confirmed: FAT32/vfat ESPs generally have no native unix permission bits (readability is governed by mount options, not file perms), so this may not be reachable in the appliance's actual mount configuration -- flagging as a note since I could not verify the ESP's actual mount options/umask in this diff.
  - demonstrate: `chmod 000 $PRESEED_DIR/pithead-data-wiped (or mount the ESP with a restrictive umask) then run doctor as non-root -- observe it reports clean identically to the file being genuinely absent`

Issue #1105, opened 2026-08-18
pithead is 10,372 lines in one file, and it is now the main source of twin-branch merge cost


VijitSingh97:
## The numbers

On `develop-v2`:

| File | Lines |
|---|---|
| `pithead` | **10,372** (236 top-level functions) |
| `tests/stack/run.sh` | **11,027** |
| `docker-compose.yml` | 993 |

`pithead` on `develop` is 6,950 — the same file, 3,400 lines apart across the twins.

The ten largest functions account for roughly a third of the CLI: `firstboot_wizard` 494, `render_env` 491, `parse_and_validate_config` 485, `describe_change` 412, `render_quadlet_units` 382, `doctor` 343, `control_upgrade` 308, `stack_backup` 194, `apply` 182, `show_help` 177.

## The argument that is actually load-bearing

Not readability — **merge cost.** `develop` → `develop-v2` conflicts in `pithead` and `tests/stack/run.sh` on essentially every sync, and the standing guidance for it is "these are routinely both-sides-added, keep BOTH". That guidance exists because the two files are where every change lands, so every change collides.

The cost is not hypothetical and it is not only time:

- Taking "mine" on one such conflict would have silently reverted #1070 — the ordering bug that stranded a production box's control channel.
- In this session a `develop-v2` merge left **unresolved conflict markers committed** in `docs/dashboard.md`, because the merge named two conflicting files and only the tail of its output was read. `make lint-md` caught it; nothing else would have, and the marker rendered as a heading rather than as obvious garbage.

A conflict you have to reason about is a chance to get it wrong. Ten thousand lines in one file maximises how many of those you get per sync.

The secondary argument is ordinary: 236 functions in one namespace, and a reviewer who cannot open the part they care about without the rest.

## What makes this harder than it looks — do not skip this

`pithead` is a **shipped artifact**, not just source. Splitting it naively breaks four contracts at once:

1. **The release bundle** copies exactly `pithead pithead-completion.bash VERSION docker-compose.yml config.*.json cosign.pub` (`scripts/release.sh`, `make_bundle`). A `lib/` would change the artifact shape, the extraction, and the versioned-directory upgrade path.
2. **The appliance** runs `/data/pithead/pithead` from a systemd unit and syncs the tree every boot. Extra files are extra things `pithead-sync` and the mount generator have to be right about.
[...]

PR #1277, opened 2026-08-22
test(stack): split the harness-tooling self-tests into their own file (#1105 Phase 1, module 2)

VijitSingh97:
Module 2 of #1105 Phase 1 (develop; the develop-v2 pair follows once its lane's 1b pair lands) — the first domain cut per the map on the module-1 PR.

The four repo-tooling self-test sections (operator-strings, pin-watch, resolve-pins, patch-coverage overlap) move verbatim into `tests/stack/test-harness-tooling.sh`, sourced by `run.sh` at the exact position the block vacated. The sections are contiguous, so execution order is unchanged — this cut is held to the strict ordered-identity standard, not the sorted-multiset relaxation. `lint-sh` gains the `tests/stack/test-*.sh` glob every later domain file lands under; the new file follows lib.sh's sourced-not-executed convention (`# shellcheck shell=bash`, mode 644). `run.sh` 8219 → 8192; ceiling follows.

**What was RUN (real git worktrees):**
- Before, at the develop tip: **1836 passed, 0 failed.** After, on this branch: **1836 passed, 0 failed.**
- Logs line-for-line identical after normalizing ANSI/`line N:`/worktree paths — the only residual diff is one mktemp path.
- `scripts/lint-file-budget.sh` OK; `shellcheck` + `shfmt` clean on the new file.
- No mutation applies — pure move; the identical-suite proof is the verification.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ test-integration: ## Run the live config-matrix integration suite (requires a te
 lint: lint-sh lint-py lint-js lint-yaml lint-md lint-docs-voice lint-operator-strings lint-topology lint-file-budget lint-proto lint-toml ## Lint/format-check every surface
 
 lint-sh: ## shellcheck + shfmt over the CLI, build/* container scripts, release + test scripts
-	shellcheck --severity=warning pithead pithead-completion.bash scripts/*.sh build/*/*.sh tests/stack/run.sh tests/stack/lib.sh tests/stack/test_compose.sh \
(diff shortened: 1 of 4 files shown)


VijitSingh97:
Post-proof amendment: CI's shfmt flagged a doubled blank line at the splice; the fix deletes one empty line (plus the budget count following it). The proof runs predate it — the entire delta is that deleted blank line, which bash neither executes nor prints, so the identical-suite proofs stand as posted.

VijitSingh97:
Verifier reconciliation: the tracked git mode is 100644 for both new files (the on-disk 664 is this box's umask — git stores only the executable bit); the second commit is the shfmt splice fix disclosed in the amendment comment; body counts corrected to 8219→8191.

PR #1278, opened 2026-08-22
test(stack): split the wizard/setup domain into its own file (#1105 Phase 1, module 3)

VijitSingh97:
Module 3 of #1105 Phase 1 (develop lane; stacked on module 2, which it includes until that merges — the diff to read is the top commit).

Two verbatim clusters move into `tests/stack/test-wizard-setup.sh`: the wizard prompt/flow + setup e2e sections (the file is sourced at the position that cluster vacated) and the setup-time kernel/GRUB tuning pair — the code-checked re-home from the domain map (`optimize_kernel` runs inside `cmd_setup`). The tuning pair therefore executes later than it used to: this is a reordering cut, held to the sorted-multiset standard. `run.sh` 8192 → 7981; ceiling follows.

**What was RUN (real git worktrees):**
- On this branch: **1836 passed, 0 failed.**
- Against the module-2 run: the two logs' *sorted* line multisets are identical after normalization — the only residual diff is one mktemp path. Every section header and every per-test verdict survives exactly once; nothing dropped, nothing changed, only order.
- `scripts/lint-file-budget.sh` OK; `shellcheck` + `shfmt` clean on the new file.
- No mutation applies — pure move; the multiset-identical suite proof is the verification.


VijitSingh97:
Post-proof amendment: CI's shfmt flagged a doubled blank line at the splice; the fix deletes one empty line (plus the budget count following it). The proof runs predate it — the entire delta is that deleted blank line, which bash neither executes nor prints, so the identical-suite proofs stand as posted.

Issue #1273, opened 2026-08-22
file-budget lane detection false-REDs any develop-v2 branch behind the moved tip


VijitSingh97:
Found by the adversarial review of a develop-v2 PR: `scripts/lint-file-budget.sh`'s `resolve_base_ref()` detects the lane by asking whether `origin/develop-v2` is an ancestor of HEAD. That is true only for a branch based on the lane's *current* tip — the moment develop-v2 advances (any other merge landing first), every in-flight develop-v2 branch stops satisfying it, falls through to the `origin/develop` baseline, and the monotonic check reads this lane's legitimately larger ceilings as illegal raises. The failure is a false RED on a correct branch, and it reproduces exactly when two develop-v2 PRs are in flight at once — the normal case for paired-lane work.

Repro shape (no real values needed): branch B cut from develop-v2 tip T1; unrelated PR merges making tip T2; on B, `git merge-base --is-ancestor origin/develop-v2 HEAD` now fails; the script ratchets B against develop and FAILs on any v2-only ceiling.

Suggested fix, still ancestry-derived and lane-symmetric: detect the lane from the *merge-base* instead of the tip — the branch is develop-v2-lane iff `git merge-base HEAD origin/develop-v2` is NOT an ancestor of `origin/develop` (its history contains v2-only commits). A branch cut from any historical develop-v2 tip passes; a develop-lane branch (whose merge-base with develop-v2 is always inside develop's own history) does not. The same-shape check applies to the topology guard if it ever grows a lane-relative mode (today it does not).

Workaround until then: rebase the branch onto the current develop-v2 tip (what unblocked the finding PR).

Issue #1257, opened 2026-08-22
Wire the three merged lint guards into CI — blocked on a workflow-scope push


VijitSingh97:
## What is stuck, and why

Three lint guards are merged, green, and running via `make lint` (and by hand) — but their CI
wiring cannot land, because the working credential cannot push workflow files:

- [x] topology-class guard (#1228) — `.github/workflows` wiring
- [x] file-size budget gate (#1105 Phase 0) — CI half
- [x] obsolete-`.trivyignore` watch (#1174) — scheduled workflow

The push is refused for the `workflow` OAuth scope — the known limitation of the token in use.
The prepared branches sit ready on the build box; only the push is blocked. Until this lands,
each guard protects local runs but **not** PRs from environments that skip `make lint`, and
the trivyignore watch does not run at all on a schedule.

## The unlock (operator, ~2 minutes, one time)

Either grant the scope to the existing auth:

```
gh auth refresh -h github.com -s workflow
```

…or push the staged branches over an SSH-key remote once. After that, the wiring PRs open
normally and this issue closes with all three boxes checked.

## Reminder for the wiring PRs

`schedule:`-triggered workflows are read from the **default branch only** — prove each
scheduled guard actually fires from `gh run list --json event` after merge, never from the
YAML's existence (the false-green trap already documented for the rootfs pin watch).

VijitSingh97:
All three boxes are merged and PROVEN per this issue's own bar — run events, never the YAML existing:

- **Topology guard (#1228)** — PR #1287. The lint-surfaces job runs `lint-topology`; proven by that PR's own CI run on the merged tree and every push/PR event since.
- **File-budget gate (#1105 P0)** — PR #1289. Held deliberately while develop's tip was honestly red on the gate (#1285), rebased over #1287's same-line merge, merged once the shrink landed — the gate passing on its own merged tree is the proof, and every PR now runs it.
- **Trivyignore watch (#1174)** — PR #1288. The workflow lives on develop (schedule mechanics); the job checks out develop-v2, where the script and the appliance rootfs it must scan live. Firing proven from `gh run list --json event`:
  - `workflow_dispatch` run 32602211613 → success: full end-to-end, six images built and scanned, tracking issue #1295 created — and the first report already earned its keep, naming one obsolete mute (CVE-2026-45447; triage caveat from #1290 noted on #1295 before anyone deletes on it).
[...]

Issue #1206, opened 2026-08-21
make lint-sh peaks at ~4.9 GB RSS and can OOM the host


VijitSingh97:
`make lint-sh` peaks at ~4.9 GB RSS and can OOM the host

## What happens

`make lint-sh` passes every shell file to a single `shellcheck` invocation. That one process
peaks at **4.85 GB resident** with shellcheck 0.11.0, and needs **more than 7 GB** with
shellcheck 0.9.0 — which is what `apt install shellcheck` gives you on Ubuntu 24.04.

On a machine doing anything else at the same time, that is enough to trigger a system-wide
OOM kill. It already happened here: the kernel killed shellcheck at 7,124,732 kB anon-rss,
and because the OOM landed inside a terminal-multiplexer scope, systemd stopped the whole
scope and took unrelated processes down with it. The lint did not fail cleanly — the machine
did.

## Measurements

All with `--severity=warning`, peak RSS via `/usr/bin/time -v`, each run confined to its own
cgroup (`systemd-run --scope -p MemoryMax=… -p MemorySwapMax=0`):

| What | shellcheck | Peak RSS | Wall | Result |
|---|---|---|---|---|
| Full `lint-sh` file set | 0.11.0 | 4,852,076 kB | 81 s | exit 0 |
| `tests/stack/run.sh` alone | 0.11.0 | 4,888,448 kB | 49 s | exit 0 |
| `run.sh` lines 1–4190 | 0.11.0 | 1,335,552 kB | 15 s | — |
| `run.sh` lines 4191–8381 | 0.11.0 | 1,771,904 kB | 17 s | — |
| Worst of every other file | 0.11.0 | 889,216 kB | — | — |
| Full file set | 0.9.0 | >7,124,732 kB | — | OOM-killed |

Two things follow. The cost is **almost entirely `tests/stack/run.sh`** (8,381 lines) — it
alone accounts for the whole peak, and every other linted file stays under 890 MB. And it is
**superlinear**: halving the file cuts memory by roughly a factor of three per half, so the
size of that one file, not the number of files, is the driver.

## Why it is worth fixing

Ubuntu 24.04 ships shellcheck 0.9.0, so the default contributor setup is the >7 GB case. On a
laptop with 8 GB, `make lint` does not report a lint error — it takes the machine out, and the
OOM killer picks its own victim, which may be something else entirely.

## Options, with what the data says about each

[...]

PR #1124, opened 2026-08-18
fix(appliance): repair a wedged /data instead of erasing it (#1062)

VijitSingh97:
Closes #1062.

`data_mountable` ran `fsck -p` and nothing else. Preen mode repairs only what needs no operator
decision and refuses everything else **by definition**, so the precise damage class a full repair
recovers was the class that reached `mkfs.ext4 -F`. A corrupt primary superblock is the textbook
case: `fsck -p` exits 8 printing the very command that would have saved the partition, and the box
reformatted instead — wallets, Tor onion private keys, the dashboard database and both synced
chains, on a partition `e2fsck -b` recovers intact. Afterwards the machine boots into the wizard
looking factory-fresh, so the operator reads it as "it reset itself" and never suspects the disk
was salvageable.

## The repair now escalates

`fsck -p` → `e2fsck -y` → `e2fsck -y -b <backup>` against the superblocks **`mke2fs -n` computes
for this filesystem**, never a hardcoded 32768 (that is only the first backup on a 4 KiB-block
layout, and `/data`'s block size follows its size; `-n` prints the geometry it would create and
writes nothing).

Bounded at two backups. If the first two cannot rebuild it the damage is not a lost superblock, and
an unbounded loop would grind a destroyed filesystem while the box sits unbootable — the reformat
path exists precisely so a shell-less release image has a way back.

## Evidence, so a wipe is not mistaken for a fresh machine

When a reformat does happen it is recorded on the ESP, the only writable thing a `/data` wipe leaves
standing. **Nothing reads that note yet** and a release image has no shell, so the operator who
needs it is looking at a wizard with no way to reach a file — filed as #1121, because surfacing it
in the wizard and in `doctor` is a change of its own.

## Coverage — both tiers were green over this and neither could catch it

- **Tier 1 (#1086):** `fsck` was stubbed `exit 0` and the mount stub succeeded on its second call
  whatever ran in between, so *"fsck repairs the mount -> skip"* was a fact about the stub. Each
  stub now models its real contract — `fsck -p` refuses with rc 8, `e2fsck` repairs what preen
  would not, `-b` rebuilds from a backup — the partition carries a damage level, and mount succeeds
  only once the tool that can fix that level has run. Five decision assertions, four escalation-order
  assertions, four evidence assertions.
- **Tier 4 (#1087):** the leg asserted the box came back to a wizard, which a full reformat
  satisfies — so a green battery certified the data loss. It now plants a sentinel in `/data`
[...]

PR #1280, opened 2026-08-22
test(stack): prove the #1062 repair escalation against real damage (#1069 W1)

VijitSingh97:
# The #1062 repair escalation, proven against real damage (#1069 W1)

**Held for operator review — data-loss class. Do not merge without eyes on it.**

#1124 fixed the defect: `pithead-data-reset` now escalates `fsck -p` → `e2fsck -y` →
`e2fsck -y -b <backup>` before it will ever reformat `/data`. What #1069 W1 still owed is the
proof: the stubbed decision-tree block in `tests/stack/run.sh` proves marker precedence and
escalation *order*, but a stub's verdict can never prove the *repair* — the exact gap that let
the original defect ride a green suite (#1086).

## What this adds

- **`tests/stack/test_data_reset.sh`** — the real thing at tier 1. A genuine ext4 image, the
  same two-byte superblock-magic wipe the tier-4 battery injects, and the system's own
  `fsck`/`e2fsck`/`mke2fs` doing the repairing. Only `mount` is stubbed — mounting needs root —
  and its verdict is `e2fsck -fn` on the image itself, never a counter. Three cases: a healthy
  image is never touched; the repairable damage class from the issue comes back **payload
  intact, never reformatted**; a destroyed filesystem still reaches the reformat escape (the
  defect was choosing it first, not having it). Runs unprivileged in ~0.3 s; hard-fails rather
  than skips when CI lacks e2fsprogs.
- **Wiring**: `make test-stack`, `make lint-sh`, and the test inventory. The CI step is
  written but **not in this PR**: any `.github/workflows/` edit is rejected by the
  workflow-scope push limitation, so it joins the blocked wiring on #1257. Until it lands,
  CI's `bash tests/stack/run.sh` step does **not** run the new suite — `make test-stack`
  does, locally and at the release gate.
- **Docs**: the reset phase's missing rows — the `reset` row in `appliance-release.md`'s phase
  table and two rows in `testing-strategy.md` section J (the doc half of #1069 W12; the
  dispatch half landed with #1087).
- **`pithead-data-reset.service`**: an explicit `TimeoutStartSec=infinity` with the reason.
  No behavior change — see the security-review note below.

## Evidence

**Suite**: 7 passed / 0 failed. **Mutation proof** (each applied to the product, suite re-run,
product restored):

| Mutation | Result |
|---|---|
| Revert the #1124 escalation (`fsck -p` only) — the original defect respelled | RED ×3: the skip verdict, the escalation log, the payload survival |
| "Repair" replaced by `mkfs.ext4` — repair-by-erase | RED ×2: payload survival, and the destroyed-image case |
| `data_mountable` always succeeds — never reformat | RED ×3: escalation log, payload survival, reformat escape |

**Gate on this branch** (re-run on the final rebased tree where it matters): stack suite
**2880 passed / 0 failed** plus the new suite's 7/0, dashboard 1852 passed (96.83 %
coverage), frontend 400, compose ✓, integration self-test 214, fakes 23, patch coverage 94 %,
inventory regenerates with the new section and its drift gate exits 0.

**Lint on the final rebased tree**: every surface passes except `lint-sh`, which dies at the
local 6 GiB shellcheck cap — `tests/stack/run.sh` *alone* (13,782 lines) exceeds it,
reproduced on the untouched tip as a control (measurement posted to #1206; CI's uncapped
runner is unaffected). Both files this PR touches under `lint-sh` are individually
shellcheck-clean. (`lint-file-budget` went red mid-session for the same everything-red reason
it went red for everyone — the #1273 lane-detection defect — and is green here on the tip
that carries its fix.)

**KVM battery, `--phase reset`, built from this branch's tree (HEAD c4539fe)**: **17 passed /
0 failed.** Leg 1 (real `pithead factory-reset` off a provisioned machine): back to an
unprovisioned wizard, config gone, container store recreated, machine-id and SSH host key
fresh, and the wipe recorded on the ESP. Leg 2 (the #1062 class — ext4 magic corrupted on the
real data partition):

```text
✓ guest survived a wedged /data — no brick
✓ the box came back usable after the corrupt superblock
✓ /data SURVIVED the corrupt superblock — it was repaired, not erased (#1062)
✓ the ESP wipe log did not grow (1 line(s), from leg 1's real factory reset) — a repair is not recorded as a wipe
✓ leg 1's factory reset was recorded on the ESP (1 line(s))
```

## Reviews (both fresh-context, findings re-derived before acting)

- **Adversarial verifier**: passes; no vacuity found beyond what the mutation table covers.
  Two fixture-diagnosis hardenings it suggested are in (`mkfs`/`debugfs` failures now fail
  loudly at the fixture, not confusingly at an assertion). Its mount-stub position note is a
  fails-red-not-false-green fragility, left as is.
- **Security review**: one HIGH claim — no `TimeoutStartSec=`, so systemd SIGKILLs `e2fsck`
  mid-repair at ~90 s — **refuted by re-derivation**: the unit is `Type=oneshot`, which
  disables the start timeout by default (verified on live systemd 255:
  `TimeoutStartUSec=infinity` against a 1 min 30 s manager default). What survived is a
  convention gap: the sibling early-boot units state their unbounded starts explicitly, this
  one didn't. Now it does, with the reason — killing e2fsck mid-rewrite can turn recoverable
  damage into unrecoverable. One LOW (a bash trap clobbering lib.sh's cleanup, leaking a temp
  dir per run) confirmed and fixed. Everything else examined and ruled out: the suite touches
  only files under its own mktemp dir, no real mounts or loop devices; the marker/wipe-log
  surfaces are existence-checked or fixed-string only; the CI step needs no new privileges.


VijitSingh97:
Updated onto the current tip per the operator's merge-intent (the module-split wave had
moved the lint enumeration this branch also touched). What changed and what deliberately
did not:

- **One conflict, one resolution**: the `lint-sh` shellcheck line — resolved as the union
  (the wave's `tests/stack/test-*.sh` domain glob plus this branch's
  `tests/stack/test_data_reset.sh`). Nothing else in the recipe changed.
- **The proof files are byte-identical** to the reviewed branch: the test, the unit file,
  the inventory entry, and both docs came through the merge unchanged, so the recorded
  real-damage evidence still describes exactly this content.
- **Re-proven on the merged tree**: `bash tests/stack/test_data_reset.sh` — 7 passed,
  0 failed (real superblock damage repaired, payload intact, reformat escape reachable).
  Budget gate, per-file shellcheck, and shfmt all clean.
- **Layout decision**: the W1 test stays a standalone runner rather than folding into a
  domain module. It follows the existing standalone convention (`test_compose.sh`: its own
  `make test-stack` invocation, underscore name), and the reset-domain module file the map
  plans does not exist yet — folding in would cut that domain ahead of its queue slot and
  mix a rescue with a split. When the reset-domain cut happens, that PR can absorb this
  file if the map wants it.

VijitSingh97:
One real finding from the fresh-context verification of the update, recorded before merge:

**The W1 test does not run in CI.** The shell-tests job invokes `bash tests/stack/run.sh`
and `bash tests/stack/test_compose.sh` as explicit steps — it never calls
`make test-stack`, which is the only place `test_data_reset.sh` is wired. So CI green on
this PR has never included the W1 test; its evidence is the recorded bench/local runs plus
the re-run on this exact merged tree (7 passed, 0 failed).

This was true of the branch before the update (not introduced by the merge), and the fix
is a one-line CI step following the `test_compose.sh` precedent — a `.github/workflows`
edit, which only an operator-scoped token can push. Added to the owed workflow-wiring
list alongside the other pending CI wirings.

Two other verifier claims were checked and refuted: the inventory documentation block is
present in the merged tree (line 119), and the "extra" changed files are develop-v2's own
content flowing through the merge as expected.

VijitSingh97:
The CI step is back in: the workflow scope landed (#1257's unlock), so the suite now runs on every PR instead of only in local `make test-stack` — the one piece the PR body had to declare owed. No other change; the PR remains held for your review.

VijitSingh97:
Correction to my comment above: it raced your merge — #1280 was already merged when it posted, and the push behind it briefly re-created the deleted branch (removed since). The owed CI step is up as its own PR against the post-sync tip instead.

PR #1284, opened 2026-08-22
sync: merge develop into develop-v2 (re-level after the module-split wave)

VijitSingh97:
# Sync: develop → develop-v2 (re-level the twins after today's wave)

Routine ancestry sync. Brings develop's 28 commits since the last level point into this
lane's history. The only content that actually flows in is the develop-only dashboard
change from #1276 (the XvB projected-steering feature: `config.py`, `algo_service.py`,
`donation_model.py`, both test files, `docs/dashboard.md`) — everything else in the range
already landed here through its reviewed twin PR (#1272, #1275, #1281, #1282, #1283).

## Conflict resolution (7 files, all on the twin seam)

- `tests/stack/lib.sh`, `test-cli.sh`, `test-harness-tooling.sh`, `test-wizard-setup.sh`,
  `Makefile`, `tests/service/test_env_key_perimeter.py`: resolved to this lane's version —
  each is the reviewed superset of the develop side. Verified by containment check, not
  assumption: every develop-side function and section name is present in the resolved
  files (empty `comm -23` on sorted name sets for all four test files, plus the wizard
  section-marker set).
- `docs/dev/file-budget.tsv`: regenerated with `scripts/lint-file-budget.sh --generate`
  on the resolved tree — never taken from either side (it is a per-lane file).
- `tests/stack/run.sh`: auto-merged to a zero diff against this lane's version, confirmed
  byte-identical to the tip that just passed the 2880/0 suite proofs.

## What was RUN

- `bash -n` on `run.sh`, `lib.sh`, and the three domain files: clean.
- Containment checks as above.
- The full-suite proof on this exact merged tree is CI's shell-tests job on this PR
  (expected 2880 passed / 0 failed, identical to the tips it merges); the dashboard
  changes carry their own pytest/frontend jobs. Local full `lint-sh` on this lane stays
  blocked by the known shellcheck memory cap on `run.sh` — CI is the whole-set arbiter.

Every commit in the incoming range was adversarially reviewed on its own PR before its
merge; the resolution above introduces no new content.


VijitSingh97:
**HELD — do not merge yet.** Two corrections since this opened:

1. The merge resolution had left conflict markers in `docs/dev/file-budget.tsv` (my
   `--generate` invocation printed to stdout and the output went to `/dev/null`, so the
   claimed regeneration never ran). Fixed in ef23a88: the tsv is restored verbatim to this
   lane's last committed version.
2. The incoming #1276 dashboard changes exceed four recorded budget ceilings, and the
   ratchet forbids raising ceilings — so this sync would import a budget violation into
   this lane. It holds until the four files are split/shrunk back under their ceilings on
   develop (issue filed), then this branch gets refreshed with the compliant versions and
   re-proven.

CI green on this PR is real for everything it covers, but the budget gate is not CI-wired
(#1257), which is how the violation got this far.

VijitSingh97:
Hold lifted. The budget fix landed on develop (the shrink PR), and this branch is refreshed with both tips: the compliant dashboard split flows in from develop, the W1 data-reset test from this lane's own history. The tsv is regenerated on the resolved tree and the gate reports monotonic. Merging on CI green.

Issue #1285, opened 2026-08-22
The steering merge grew four dashboard files past their budget ceilings, and nothing said so


VijitSingh97:
# The #1276 merge grew four dashboard files past their budget ceilings, and nothing said so

## What happened

PR #1276 (the XvB projected-steering feature) landed with growth in four budgeted files:

| file | ceiling | now |
|---|---|---|
| `build/dashboard/mining_dashboard/config/config.py` | 852 | 865 |
| `build/dashboard/mining_dashboard/service/algo_service.py` | 560 | 621 |
| `build/dashboard/mining_dashboard/sim/donation_model.py` | 581 | 594 |
| `build/dashboard/tests/service/test_algo_service.py` | 839 | 907 |

`make lint` on the develop tip now fails on the file-budget gate. The PR's own CI was
green because the gate is not wired into CI yet (#1257 — the workflow-scope push), and no
full local `make lint` ran on the merged tip until a later branch happened to.

This is the exact failure mode CONTRIBUTING.md's budget section warns about: the gate
"runs locally and in `make lint` but is not yet wired into CI, so two same-wave merges can
each pass alone and fail together."

## What the fix is (per the documented protocol)

Ceilings only go down, so recording the growth is not a legal fix. The one legal path is
to split or shrink each file back under its ceiling (the #1258 worked example: the test
that outgrew its file moved into its own). Concretely the steering feature's additions are
cohesive and extractable:

- `algo_service.py`: the projection cluster (`_record_avg1h_sample`, `_projected_avg_1h`
  and their constants) can move to a sibling module.
- `test_algo_service.py`: `TestProjectedSteering` can move to its own test file, exactly
  the #1258 shape.
- `config.py` and `donation_model.py` are small overages; a matching extraction or an
  adjacent shrink covers them.

## Wider note

Until #1257 lands, every merge wave needs a `make lint` on the merged tip as part of the
close-out — CI green does not cover the budget gate. The develop→develop-v2 sync PR
currently open is held on this fix, since the sync would otherwise import the violation
into the appliance lane.

PR #1289, opened 2026-08-22
ci(lint): run the file-budget gate in the lint-surfaces job (#1105 P0, #1257)

VijitSingh97:
~~DO NOT MERGE until the #1285 shrink lands on develop.~~ **Un-held**: #1285 is closed, the shrink is on develop, and `make lint-file-budget` passes on this branch's rebased tree (run locally; this PR's own CI now runs the gate on the merged tree as the honest proof).

The wiring is the same two-line shape as every other lint surface (#1105 Phase 0's CI half, first #1257 box): the lint-surfaces step gains `lint-file-budget`, Makefile target as the single source of truth. The gap this closes already cost once — #1276's growth was caught by an agent reading the tree, not by CI, exactly the same-wave trap the gate exists for.

Rebased over #1287's merge of the same two lines; the step now carries both suffixes (`topology classes, file budget`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -289,6 +289,6 @@ jobs:
           export UV_INSTALL_DIR="$HOME/.local/bin"
           curl -LsSf https://astral.sh/uv/0.10.10/install.sh | sh
           echo "$UV_INSTALL_DIR" >> "$GITHUB_PATH"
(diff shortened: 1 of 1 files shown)


VijitSingh97:
Heads-up recorded from #1287's adversarial review: this PR edits the same two ci.yml lines as #1287, so once #1287 merged this branch conflicts. The plan stands as the body says — after the #1285 shrink lands on develop, this branch gets rebased onto the then-current develop (combining the step suffixes) and un-held; nothing merges before both.

Issue #1050, opened 2026-08-16
Dashboard OS-update flow has four dead-end states an operator cannot recover from in the UI


VijitSingh97:
Four confirmed defects in the #976 dashboard OS-update action, all found by adversarial review of the merged wave and each reproduced against the real code. None is a security hole; all four strand the operator in a state the UI offers no way out of.

**1. The 24-hour reboot authorization dead-ends the flow (`pithead`, `control_os_reboot`).** A successful install authorizes `os-reboot` for 24 hours; after that the host refuses and names the re-arm path (fresh verify + install). Correct host-side. But no dashboard path re-arms it — an operator who installs, goes away for two days, and comes back to click Reboot gets a refusal and no button that fixes it.

**2. The idle modal offers Download as its primary action before any check has run (`osupdate.mjs:356`).** There is no lifecycle hook and no auto-check, so on a fresh appliance the first click is always Download, which always fails because no target has been resolved. Retry repeats the same failing call.

**3. A failed `os-install` leaves the persisted state at `step:"installing"` forever (`pithead`, `control_os_install`).** The `rc != 0` branch writes a refusal result but never moves the state file off `installing`, so the UI shows an install in progress that will never finish or fail.

**4. The `os-check` throttle stamp is written before the fetch (`control_os_check`).** `touch "$stampf"` runs *before* `os_release_fetch`, so a check that fails to reach the release API still burns the 10-minute throttle. An operator who fixes their connectivity and retries is told to wait.

### Suggested shape

1 and 3 are the same root: the state machine has no terminal-failure transition and no re-arm transition. 2 is a missing initial check on modal open. 4 is stamping after a successful fetch rather than before the attempt.

Also worth fixing alongside: `docs/dashboard.md` says the Check button "asks now", with no mention of the 10-minute throttle that can refuse the click outright.

VijitSingh97:
Fixed by PR #1307 (merged to develop-v2): stale-authorization reboot re-arms for real, install failures land in a terminal reported state, and transport-vs-refusal failures are distinguished — with a short cooldown on the transport class preserving the anti-beacon throttle (review round-trip documented on the PR), plus the state-bleed fix the full suite surfaced.

Issue #1030, opened 2026-08-16
First boot can be interrupted mid-image-load, and the console keeps no record of why


VijitSingh97:
Split out of #1029, which fixed the *consequence* (a container store left with zero-length `lower` files is now detected and rebuilt). This issue is the *cause*, which is *not* proven.

**What is established.** On the HP bench, booting the installation medium, the first boot reached `Loading this build's container images (dashboard.tar.gz) — the slow part...` and then the machine came back to a fresh `pithead login:` — a reset partway through the load. That reset left seven zero-length `lower` files in `/data/containers/storage/overlay`, which bricked every later boot until the store was rebuilt by hand.

**What is not established: what reset it.** The journal cannot answer — `journalctl --list-boots` shows only the current boot, and `journalctl -u pithead-firstboot -b` is `-- No entries --` even though the unit demonstrably ran and failed. Storage is volatile, so the evidence from the failing boot is simply gone.

Candidates, in the order I would test them:

1. **The hardware watchdog.** `os/rootfs/Dockerfile:157` sets `RuntimeWatchdogSec=20s`, and the box reports `Using hardware watchdog 'iTCO_wdt' … timeout of 20s`. The comment there argues 20s is "long enough that a busy box is not falsely reset". This medium is slow enough to call that into question: `podman load` of the 61 MB archive took **over four minutes**, and the wizard needed **150s** from launch to serving. Sustained USB I/O starving PID 1 for 20s is plausible.
2. **A USB/storage stall or link reset** under sustained write, which would also explain the timing without involving the watchdog.
3. **Power.**

**Why it is worth chasing even now that #1029 self-heals.** The repair costs a full reload on the next boot, and if the reset is systematic on slow media, a first boot could loop: load → reset → rebuild → load → reset. #1029 makes that recoverable, not free.

**Suggested work.**
- Make the journal survive the failure so the next occurrence is diagnosable at all. `Storage=volatile` is the reason nothing could be reconstructed here — persisting it to `/data` for the firstboot window would have answered this in one command. This is the single highest-value change and it is small.
- Instrument the load: log start/finish and elapsed time per archive, so a truncated load is visible as a gap.
- Re-examine whether `RuntimeWatchdogSec=20s` is right for the first boot specifically, where the slowest, most I/O-bound work happens on the slowest media the product ever runs on. A longer window during firstboot only, or feeding the watchdog around the load, would both work.

[...]

PR #1306, opened 2026-08-23
fix(appliance): refuse miner co-location on the reduced hugepages tier

### LINKED ISSUE #1103: Appliance: a reduced-RAM box over-reserves HugePages when the built-in miner is on
## What happens

On an appliance below the supported 16 GB floor, the boot-time sizing (`os/overlay/pithead-hugepages`) reduces the HugePages pool to 2560 pages (5 GiB) and records `pages=2560`. `render_local_miner_config` then declares that decision to RigForge as `hugepages_reserve_extra_mb` (5120).

`hugepages_reserve_extra_mb` is **additive headroom**, not a ceiling. RigForge's `util/proposed-grub.sh` computes

```
required = 1168 * numa_nodes + 128 + threads + 10 + (extra_mb + 1) / 2
```

and `_ensure_hugepages` grows the pool by `required - avail`, where `avail` counts only **free** pages plus pages the miner itself already holds. So once the stack is running and holding its share:

- the stack's pages are inside `required` (as the declared headroom), **and**
- the same pages are excluded from `avail` (they are neither free nor miner-held)

They are counted twice. With one NUMA node and 4 threads, `required` is about 3870 pages; with the stack holding roughly 2336 of the 2560 reserved, `avail` is about 224, and the write becomes `2560 + 3870 - 224 = 6206` pages — about 12 GiB requested on a machine with 8. RigForge's own note calls the double-count "a bounded over-reserve in the safe direction", which it is on the supported machine and is not here.

## Why the sizing fix does not cover it

The reduced tier's 2560 pages were sized for the stack's two RandomX datasets alone — p2pool's and monerod's, 1168 pages each plus slack. A co-resident miner was never in that budget, and no value of `hugepages_reserve_extra_mb` bounds the result, because the over-reservation comes from the double-count rather than from the declared number. On the supported machine the sum is intentional and measured; only the degraded tiers are wrong.

## The decision needed

What a below-floor machine should do when `local_miner.enabled` is on. Options, roughly in increasing cost:

1. Refuse to co-locate: on a degraded box, leave the built-in miner switched off and say so on the console and in `doctor`, the same way the reduced reservation is already announced.
2. Let the miner run without a HugePages share — it is slower, but the stack keeps the reservation it needs to avoid p2pool's cgroup cap killing its dataset.
3. Add a total-pool ceiling to the RigForge contract (a new key that caps `nr_hugepages` rather than adding to it), so the declaration can express "the pool must not exceed N".

Option 3 is the only one that makes the declaration honest, and it needs a change in both products.

## Where this came from

Found reviewing the PR that scaled the boot reservation to fitted RAM (#977). That PR is correct for what it does — it stopped a constant 6144 MB declaration from re-inflating a pool the boot had just shrunk — and this is the remaining gap it exposed rather than caused. The precise semantics are now written down at the `render_local_miner_config` comment.

## Coverage

Not reachable at tier 1 without re-implementing RigForge's formula, which would assert our copy of it rather than the behaviour. A tier-4 leg on a memory-limited guest that reads `HugePages_Total` after the miner provisions is the honest check.

VijitSingh97:
# fix(hugepages): refuse to co-locate the built-in miner on a reduced-RAM box

Closes #1103

Branch: `fix/1103-hugepages-colocate`
SHA: `267791a21392cca3921ca874aca8f2cfd46cc4d7`

## Root cause

`pithead:2852` `render_local_miner_config` (and `pithead:8413` `hugepages_decision_pages`,
`pithead:1027` `check_hugepages_degraded`, `os/overlay/pithead-hugepages`).

On the REDUCED HugePages tier (a box below the supported 16 GB floor: 7-15 GB RAM, boot-time
sizing reserves 2560 pages / 5 GiB instead of the full 3072 / 6 GiB), the appliance still
declared the co-located RigForge miner's headroom as `hugepages_reserve_extra_mb` = the tier's
own reservation (5120 MB). RigForge's grow-only pool write is:

```
required = 1168 * numa_nodes + 128 + threads + 10 + (extra_mb + 1) / 2
```

and it grows the pool by `required - avail`, where `avail` counts free pages plus pages the
miner itself already holds — excluding pages the STACK already holds. Since the declared
headroom (`extra_mb`) is inside `required` while the stack's matching pages are excluded from
`avail`, the same pages are counted twice. With 1 NUMA node, 4 threads and the stack holding
~2336 of the tier's 2560 pages, this worked out to requesting roughly 12 GiB of locked memory
on an 8 GB machine — the exact defect #1103 reports. No declared value fixes this (it is the
arithmetic double-counting, not the number); the RELEASED tier (0 pages reserved) already
declares zero headroom and has nothing to double-count, and the FULL tier (16 GB, 3072 pages)
is the measured, supported case — this is a REDUCED-tier-only defect.

## Fix

Per the issue's own framing, this is a product decision needing a pithead-only mitigation
(Option 1 in the issue: refuse to co-locate on a degraded box) since Option 3 (a new
RigForge-side ceiling contract) needs a cross-repo change out of scope here, and Option 2
(declare zero headroom but still colocate) only reduces, does not eliminate, the
over-reservation.

- `pithead:local_miner_hugepages_blocked()` (new): true only when the boot marker records a
  reservation strictly between 0 and the full budget — i.e. exactly the REDUCED tier. Does not
  touch sizing at all.
- `render_local_miner_config`, `provision_local_miner`, `announce_local_miner`: all treat
  "enabled but blocked" the same as switched off — no derived RigForge config, any running
  miner is stopped, and the operator is told why (not silently shown an empty Workers view).
- `doctor` gained `check_local_miner_hugepages_blocked` — a WARN alongside the existing
  degraded-hugepages check, visible the same way the reduced reservation itself already is.

## Tests run

**Tier 1 (unit, new file `tests/stack/test_appliance_hugepages.sh`, 18 assertions):**
- `local_miner_hugepages_blocked` — full/reduced/released/exactly-full-budget-with-marker, all
  four combinations.
- `render_local_miner_config` — reduced tier writes no config; full and released tiers are
  BYTE-FOR-BYTE unchanged (the explicit no-regression check for the "must not shrink
  reservations on normal boxes" requirement).
- `provision_local_miner` — blocked case stops any running miner, never runs RigForge's setup,
  leaves no derived config, and returns rc 0 (a refusal is not a failure).
- `announce_local_miner` (appliance) — blocked case names the reduced reservation and does NOT
  claim the worker is on.
- `check_local_miner_hugepages_blocked` — WARNs only when enabled AND blocked; silent when
  disabled, at full tier, or off the appliance.

Regression-verified: stashing the fix and re-running the new suite fails exactly the 1
assertion the fix addresses per scenario (confirmed manually before commit).

**Mutation coverage, by test:**
- "full tier -> headroom is UNCHANGED at 6144 MB" catches any respell that shrinks the full-tier
  declaration (the load-bearing "must not shrink on normal boxes" invariant).
- "released tier -> still renders, zero headroom (unchanged)" catches a respell that widens the
  refusal to the wrong tier.
- "reduced tier -> no derived config" + the provision/announce/doctor assertions together catch
  a respell that inverts `local_miner_hugepages_blocked`'s boundary (`-gt 0` vs `-ge 0`, `-lt`
  vs `-le` against `PITHEAD_HUGEPAGES`).

**KVM (tier 4) — RUN, with a real verdict.** Combined build (this branch + #1050 + #1030
merged into a throwaway `kvm-combined-1103-1050-1030` branch off the same base commit, per the
"build/boot at most once, reuse across all three bugs" instruction): `os/build-image.sh`
(PITHEAD_UPDATER=rauc, dev SSH key baked) → `os/rauc/mkimage.sh --dev` →
`tests/os/verify-image.sh --test` (100/100 static checks, including this fix's own doctor/
render/provision code shipping correctly). Booted manually via `virt-install --memory 8192`
(the REDUCED tier's real RAM range — `tests/os/run.sh`'s own phases hardcode 16384 and assert
the FULL tier via `hugepages-boot-verdict.sh`, so this had to be a direct boot outside that
harness). KVM lock held for the guest/battery portion only, released immediately after.

Verdict, read live over SSH against the REAL boot-produced marker (not a fixture):

```
MemTotal:        8133520 kB
HugePages_Total:    2560
HugePages_Free:     2560
pithead-hugepages.service: active (RC=0)
/run/pithead-hugepages-degraded: "This machine has 7.8 GiB of RAM — below the supported 16 GB.
  The mining reservation is reduced from 6 GiB to 5 GiB..." / pages=2560

render_local_miner_config (sourced from the ON-BOX /data/pithead/pithead, real marker):
  RENDER_RC=0  BLOCKED=yes
  /data/rigforge/config.json: NO SUCH FILE (correctly refused, not just unit-tested)

provision_local_miner:
  [WARNING] Local mining opt-in is ON, but this machine's reduced HugePages reservation
  cannot also fit a co-located miner without risking the stack itself — the built-in miner is
  not started. Use a 16 GB machine for the built-in miner.
  PROVISION_RC=0

check_local_miner_hugepages_blocked (the new doctor check):
  ⚠ WARN Local mining opt-in is ON, but this machine's reduced HugePages reservation cannot
  also fit a co-located miner — the built-in RigForge worker is not started. Use a 16 GB
  machine for the built-in miner.
```

Every line matches the fix's coded messages exactly, on a real guest boot, not a sandbox
fixture. Full log (incl. #1030's journal-persistence verdict from the same guest/boot) is on
the session record; ask the coordinator for the combined-branch worktree at
you want to re-run or extend the battery without rebuilding.

## NOT proven

- The FULL RigForge co-location interaction (an actual RigForge process computing its
  grow-only sysctl write against the declared headroom) was NOT run — RigForge lives in a
  separate repo and is not available to drive from this session. The fix is verified at the
  pithead side of the contract (no headroom declared, miner not started) via both unit tests
  and the real on-box marker read above; the RigForge-side arithmetic that originally produced
  the ~12 GiB figure is quoted from the issue, not independently re-derived here.
- The RELEASED tier (<7 GB, 0 pages) was not separately KVM-booted — only unit-tested. The
  REDUCED tier (the one this bug is about) is what was booted.
- A full wizard-driven provision with `local_miner=true` (the real end-user path) was not run;
  the KVM leg instead wrote `/data/pithead/config.json` directly over SSH and called the
  functions, which exercises the same code paths but skips the wizard's own HTTP flow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -250,6 +250,10 @@ jobs:
         # Real ext4 damage against the product's own repair escalation (#1062). Unprivileged —
         # e2fsprogs on plain files — and it hard-fails rather than skips when CI lacks the tools.
         run: bash tests/stack/test_data_reset.sh
+      - name: Run appliance hugepages suite
+        # #1103 co-location refusal on the reduced tier; standalone suite wired here because the shell-tests job runs each
+        # standalone file explicitly (Makefile test-stack alone never runs in CI).
+        run: bash tests/stack/test_appliance_hugepages.sh
       - name: Test-inventory drift check
         # The inventory generator is grep-based; it exits non-zero when any suite it
         # enumerates counts zero — i.e. a suite moved or changed shape and the inventory
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ test-patch-coverage: ## diff-cover (#286) minus its vacuous pass (#1000): >=90% 
 test-stack: ## pithead shell test suite
 	bash tests/stack/run.sh
 	bash tests/stack/test_data_reset.sh
+	bash tests/stack/test_appliance_hugepages.sh
 
 test-compose: ## Validate docker-compose.yml interpolation + hardening invariants (#90)
 	bash tests/stack/test_compose.sh
@@ -42,7 +43,7 @@ test-integration: ## Run the live config-matrix integration suite (requires a te
 lint: lint-sh lint-py lint-js lint-yaml lint-md lint-docs-voice lint-operator-strings lint-topology lint-file-budget lint-proto lint-toml ## Lint/format-check every surface
 
 lint-sh: ## shellcheck + shfmt over the CLI, build/* container scripts, release + test scripts
-	shellcheck --severity=warning pithead pithead-completion.bash install.sh scripts/*.sh build/*/*.sh tests/stack/run.sh tests/stack/lib.sh tests/stack/test-*.sh tests/stack/test_compose.sh tests/stack/test_data_reset.sh \
+	shellcheck --severity=warning pithead pithead-completion.bash install.sh scripts/*.sh build/*/*.sh tests/stack/run.sh tests/stack/lib.sh tests/stack/test-*.sh tests/stack/test_compose.sh tests/stack/test_data_reset.sh tests/stack/test_appliance_hugepages.sh \
 		tests/inventory.sh tests/integration/*.sh tests/integration/mini-stack/*.sh \
 		os/installer/pithead-install os/build-image.sh os/rauc/*.sh os/overlay/pithead-sync os/overlay/pithead-boot \
 		os/overlay/pithead-data-reset os/overlay/pithead-mount-generator os/overlay/pithead-ssh-host-keys \
--- a/pithead
+++ b/pithead
@@ -1030,6 +1030,18 @@ check_hugepages_degraded() {
     dr_warn "$(head -n 1 "$marker" | tr '\t' ' ')"
 }
 
+# #1103: on the REDUCED tier, render_local_miner_config and provision_local_miner both refuse to
+# co-locate the built-in miner even though local_miner.enabled is on (see the comment above
+# local_miner_hugepages_blocked for why no headroom value could make that safe). Say so here too,
+# the same way the reduced reservation itself is announced above — an operator staring at an empty
+# Workers view otherwise has nothing telling them the opt-in was refused rather than just slow.
+check_local_miner_hugepages_blocked() {
+    is_appliance || return 0
+    [ "$(config_bool '.local_miner.enabled' false 2>/dev/null || echo false)" = "true" ] || return 0
+    local_miner_hugepages_blocked || return 0
+    dr_warn "Local mining opt-in is ON, but this machine's reduced HugePages reservation cannot also fit a co-located miner — the built-in RigForge worker is not started. Use a 16 GB machine for the built-in miner."
+}
+
 # True (rc 0) when the host CPU advertises AVX2 — RandomX wants it, but it's performance-only,
 # not fatal. Linux reads /proc/cpuinfo; macOS reads sysctl's machdep CPU features.
 cpu_has_avx2() {
@@ -1191,6 +1203,7 @@ doctor() {
             dr_warn "HugePages_Total is 0 — RandomX mining is slower. Run './pithead setup' (kernel optimization) to reserve them."
         fi
         check_hugepages_degraded
+        check_local_miner_hugepages_blocked
 
         # Runtime counterpart to setup's capacity pre-flight: preflight_resources checks MemTotal
         # (does the host have enough RAM at all), this checks MemAvailable (is enough free right
@@ -2778,7 +2791,13 @@ announce_local_miner() {
     # On the appliance the miner is built in: pithead renders RigForge's config and the boot path
     # runs its setup, so there is nothing to install and no values to copy.
     if is_appliance; then
-        log "Local mining is ON: this machine also mines with its own CPU. The built-in RigForge worker points at the stack's own stratum and appears in the dashboard's Workers view."
+        if local_miner_hugepages_blocked; then
+            # #1103: opted in, but reduced HugePages means the built-in miner never actually
+            # starts (provision_local_miner refuses it) — say that instead of claiming it runs.
+            warn "Local mining is opted in, but this machine's reduced HugePages reservation cannot also fit a co-located miner — the built-in RigForge worker is not started. Use a 16 GB machine for the built-in miner."
+        else
+            log "Local mining is ON: this machine also mines with its own CPU. The built-in RigForge worker points at the stack's own stratum and appears in the dashboard's Workers view."
+        fi
         return 0
     fi
     local bind port host secret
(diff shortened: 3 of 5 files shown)


VijitSingh97:
## CI finding (not a review comment on this PR — a fresh CI run against the fix already on this branch)

CI (2879/1) failed a `tests/stack/run.sh` assertion: "degrade marker -> headroom follows
the recorded reservation (2560 pages -> 5120 MB)". That assertion checked exactly the
double-counting behavior this PR's own fix (267791a) removes: on the REDUCED HugePages
tier, `render_local_miner_config` no longer declares any nonzero headroom — it refuses to
co-locate the built-in miner at all and deletes any stale `config.json`, because RigForge's
grow-only pool write double-counts the stack's own reservation against any nonzero headroom
declared there. 267791a's own commit message already flagged this exact `run.sh` assertion
as owed follow-up work ("not touched here per the split-in-progress instruction").

## Fix

Renamed the assertion to state the new truth and changed it to check for the config file's
*absence* on the reduced tier, instead of a stale `hugepages_reserve_extra_mb` value —
mirroring how `tests/stack/test_appliance_hugepages.sh` already proves the same
`render_local_miner_config` call for this tier. That sibling file already covers the rest of
the #1103 gate (the operator-facing WARN in `provision_local_miner`, the miner being
stopped, doctor's `check_local_miner_hugepages_blocked`), so this `run.sh`-resident block
stays scoped to what it always proved: `render_local_miner_config`'s own derivation,
alongside its still-unchanged full-tier and released-tier neighbors in the same block.

## What was run

- Extracted the `render_local_miner_config` test block (the `echo "== unit:
  render_local_miner_config ..."` section) into a standalone driver: sourced `lib.sh` + the
  block against this branch's `pithead` — 11 assertions pass, including the renamed one.
- MUTATION KILL: reverting 267791a's reduced-tier guard (removing the
  `if local_miner_hugepages_blocked; then rm -f "$dir/config.json"; return 0; fi` branch
  from `render_local_miner_config`) turns the renamed assertion red (10 passed, 1 failed) —
  confirmed against a throwaway mutant.
- `tests/stack/test_appliance_hugepages.sh` standalone, as a sanity check that this PR's
  existing sibling suite is untouched and still green: 18 passed, 0 failed.
- `shellcheck --severity=warning` on `tests/stack/run.sh` — 3 pre-existing `SC2034`
  warnings at unrelated lines (~9343-9483), outside this diff's hunk, confirmed unrelated.
- `shfmt -i 4 -d tests/stack/run.sh` — no diff.
- `scripts/lint-file-budget.sh` — one pre-existing failure remains (`docs/dev/file-budget.tsv`
  raises `tests/stack/run.sh`'s ceiling from 10352 to 12115 relative to `develop-v2`),
  confirmed present before this change too; this edit kept `run.sh` at exactly its
  already-recorded ceiling (net zero line growth).
- Did not run `tests/stack/run.sh` in full, `make lint`, or any KVM/hardware test, per
  instruction.

New commit: `f6d9735`.

Issue #1081, opened 2026-08-17
One-click upgrade fails on a shared Tor exit's spent GitHub rate limit, and blames the wrong thing


VijitSingh97:
Hit live on the bench while deploying v1.19.1. The one-click upgrade was rejected with:

```
could not reach the GitHub release API over Tor — nothing was changed.
Check './pithead doctor' and retry.
```

Tor was healthy the whole time and the dial succeeded. GitHub answered **403**:

```
HTTP/2 403
x-ratelimit-limit: 60
x-ratelimit-remaining: 0
x-ratelimit-used: 60
{"message":"API rate limit exceeded for 51.81.222.62. ..."}
```

`51.81.222.62` is the Tor exit, not us. GitHub's unauthenticated limit is 60 requests/hour **per IP**, and a Tor exit is shared by everyone using it, so the budget is routinely spent by strangers. `curl -fsS` turns the 403 into a non-zero exit and `control_upgrade` reports it as unreachable.

## Why this matters more than a transient

The operator did nothing wrong and can do nothing useful with what they are told. Worse, the attempt **claims the 10-minute throttle before the dial**, so the obvious response — click Upgrade again — is refused with a different message for the next ten minutes. Two confusing errors in a row for a healthy box.

The actual remedy is `./pithead restart tor` to pick fresh guards (#424), which lands on a different exit with a fresh budget. That worked immediately here: `HTTP/2 200`, `x-ratelimit-remaining: 59`. But nothing in the message points at it, and `./pithead doctor` — which the message *does* point at — reports Tor egress as healthy, because it is. The operator is sent to a check that will tell them everything is fine.

There is no reason to think this is rare. Every install polls the same API from whatever exit its circuit lands on, and popular exits are the most likely to be exhausted.

## Suggested fixes, cheapest first

1. **Say the right thing.** Distinguish a 403 with `x-ratelimit-remaining: 0` from an actual dial failure, and name the remedy: the exit's shared GitHub budget is spent, run `./pithead restart tor` and retry. A `curl -w '%{http_code}'` on the existing call is enough to tell them apart — today `-f` collapses every non-2xx into one exit code.
[...]

Issue #33, opened 2026-06-01
Stack config editor: change any setting from the dashboard, applied via pithead (incl. P2Pool mode hot-swap)


VijitSingh97:
## Goal
A **web-based stack configuration editor**: change *any* configurable setting from the dashboard and have it applied to the running stack via `pithead` — no hand-editing `config.json`, no SSH. The original ask (hot-swapping the P2Pool **main / mini / nano** mode) becomes just one field in this editor.

> **Scope note.** This started as "hot-swap the P2Pool pool mode" and has grown into the project's **foundational configuration capability**. It's the shared mechanism behind two other features — the appliance first-boot setup wizard (#77) and the dashboard upgrade button (#59) — which all route through `pithead` on the host. Build it once, here.

## What it does
- A config page listing every setting `config.json` supports, grouped sensibly (Monero, P2Pool, Tari, XvB, dashboard, networking, data dirs).
- On save: **validate → write `config.json` → run `pithead apply`** (recreates only the containers whose resolved config changed), surfacing the **same change-preview + destructive warnings** `pithead apply` already prints on the CLI.

### 1. A config schema
The fields, types, defaults, and validation rules behind `config.json` (`monero.mode` / `wallet_address` / `prune` / `rpc_lan_access`, `p2pool.pool`, `xvb.*`, `tari.wallet_address`, `dashboard.*`, `reject_workers_on_*`, data dirs, hostnames). Today `config.json` is hand-edited; a UI needs this schema as the explicit contract — reusing `pithead`'s `parse_and_validate_config` rules so UI and CLI validate identically.

### 2. A write → validate → apply path
Write `config.json`, run `pithead apply`, and show the existing change preview + destructive warnings (pool change re-syncs the sidechain and resets PPLNS + XvB shares; wallet change; `prune` toggle; local↔remote switch — secrets/onions are preserved by `apply`).

### 3. A host-mutation channel (the hard part — shared dependency)
The dashboard runs in a **container** and today only does narrow `start`/`stop` via the docker-socket-proxy — it does **not** write host files or run `pithead`. Editing config from the UI needs a deliberate, locked-down channel to "write `config.json` + run `pithead apply`" on the host:
- A small **host-side agent / systemd unit** the dashboard signals (preferred), or a command queue on a mounted volume.
- This **intentionally departs from #59's original "the dashboard never mutates the host" stance** (which stood for upgrades); config-apply is an explicit, audited exception.
[...]

PR #1307, opened 2026-08-23
fix(control): OS-update dead-end states re-arm, report, and spare the throttle

### LINKED ISSUE #1050: Dashboard OS-update flow has four dead-end states an operator cannot recover from in the UI

VijitSingh97:
# fix(os-update): close the dashboard OS-update flow's three dead-end states

Closes #1050

Branch: `fix/1050-os-update-recovery`
SHA: `9b2bbef2ad8228b4591ccf343aec8d3a6418affc`

## Root cause (four defects reported; one already fixed)

Defect 2 (idle modal offers Download before any check has run, `osupdate.mjs`) was already
fixed by `f617dee` ("stop the first download failing forever") — confirmed by reading current
`download()` on develop-v2 and its matching frontend tests; not touched here. The remaining
three, each `pithead`:

1. **`control_os_reboot` (pithead:11211), 24h dead end.** On a stale (>24h) install
   authorization, the function refused the reboot but never touched the persisted
   `os-update-state.json` step, which stayed `"reboot-pending"` forever. The dashboard header
   always showed "Reboot to finish", every click opened directly into the reboot-confirm pane
   (no Check/Download reachable from there), and confirming always refused — with the
   frontend's `reboot()` (osupdate.mjs:262) not even inspecting the poll result, so it silently
   reconnected/reloaded into the identical stuck state instead of showing an error.
2. **`control_os_install` (pithead:11140), stuck "installing".** The `rc != 0` branch wrote a
   `"failed"` control result but never called `os_state_write` to move the persisted step off
   `"installing"` — confirmed by its literal absence in that branch (present in the sibling
   os-verify/os-download failure branches). The UI kept showing an install in progress that
   would never finish or fail again.
3. **`control_os_check` (pithead:10914), throttle stamp before the fetch.** `touch "$stampf"`
   ran before `os_release_fetch`, so ANY failure — including a pure connectivity failure that
   never reached GitHub at all — burned the 10-minute throttle. An operator who fixed Tor and
   retried was told to wait.

## Fix

- `control_os_reboot`: on the stale-TTL path, clears the expired in-flight file and writes
  `{"step":"idle"}` — an actual re-arm, not just a message claiming one. The frontend's
  `reboot()` now inspects the poll result and calls `fail()` on a `"rejected"` status instead of
  unconditionally reconnecting.
- `control_os_install`: the failure branch now also writes `{"step":"idle"}` — a terminal-
  failure transition, matching the shape already used by os-verify/os-download.
- `control_os_check` / `gh_release_fetch` / `os_release_fetch`: `gh_release_fetch`'s transport-
  failure branch (curl itself failed — never reached the server) now returns rc 2, distinct
  from every other failure (which DID reach the server: success, rate-limit, or another HTTP
  code — rc 1). `control_os_check` only claims the stamp when the fetch did not return rc 2.
  **Deliberately preserved:** a rate-limited check (#1081) still claims the throttle — that
  dial DID reach GitHub, and #1081's own comment says releasing it there "restores exactly the
  unthrottled beacon the throttle exists to stop." Only the never-reached case changed.

## Tests run

**Tier 1 (unit, new file `tests/stack/test_os_update_recovery.sh`, 17 assertions):** calls
`control_os_check` / `control_os_install` / `control_os_reboot` directly (stubbed
rauc/curl/systemctl, no docker, no network) covering: transport failure vs. rate-limit failure
vs. success for the throttle stamp (including an immediate-retry-after-each case); a failing
install's persisted step; a stale reboot's in-flight-file removal + step re-arm + unchanged
refusal wording (`assert_contains ... "re-arms the reboot"`, kept verbatim since
`tests/stack/run.sh`'s own frozen "the stale refusal names the re-arm path" assertion checks
for exactly that substring — I did not edit that file per the split-in-progress instruction,
and this keeps both suites compatible); a regression check that a FRESH (non-stale) install
still reboots normally.

**Regression-verified**: stashing `pithead`+`osupdate.mjs` and re-running the new suite fails
exactly 5 assertions — one per specific defect (throttle-not-claimed, immediate-retry-not-
throttled, install-state-not-idle, in-flight-not-cleared, state-not-re-armed) — confirmed
before commit.

**Frontend (`build/dashboard/tests/frontend/osupdate.test.mjs`, 2 new cases, 22/22 pass, full
suite 402/402 pass):** `reboot()` calls `fail()` (spied, not read off `this.state` — preact
defers `setState` on an unmounted component, the same caveat the existing download tests
already document) with the host's message on a rejection and does NOT call `reconnect()`; a
genuine `"rebooting"` result still reconnects as before.

**Mutation coverage, by test:**
- "an immediate retry after a connectivity failure is NOT throttled" catches a respell that
  claims the stamp on ANY failure (reverting to the pre-fix behaviour).
- "a rate-limited check still claims the throttle (#1081, unchanged)" catches the opposite
  mistake — a fix that never claims the stamp on failure, which would silently break #1081's
  anti-beacon guarantee.
- "the stale authorization is cleared, not left dangling" + "the persisted step re-arms to
  idle" together catch a fix that only changes the message text without touching state.
- "a fresh install still authorizes the reboot" catches a fix that over-broadly clears
  in-flight/state on every reboot call, not just the stale path.

## NOT proven

- No KVM leg was run for this bug specifically. It is dashboard/control-channel logic with no
  boot-time or hardware dependency (testing-strategy.md tier table places "Control channel
  (#33)" work at tier 1, tier 4 "not yet a matrix row" for this exact persisted-state scenario),
  and `tests/os/run.sh`'s existing "update" phase leg 4 already exercises the OS-update
  dashboard action end-to-end for the happy paths — a full A/B install/reboot KVM cycle
  (20-40+ minutes) was judged not worth re-running for a state-transition/timing fix that
  changes no wire protocol. If the coordinator wants tier-4 confirmation, `tests/os/run.sh
  --phase update` against an image built from this branch (or the combined branch — see the
  #1103/#1030 drafts) would exercise it, though its own assertions do not currently check the
  three specific defects fixed here (they would need new assertions in that phase to actually
  catch a regression, which is out of scope for this PR since it is tier-1-owned logic).


VijitSingh97:
## Review finding

`gh_release_fetch` returns rc 2 for any curl nonzero exit (pithead:6558-6565), and
`control_os_check` skipped the 10-minute throttle stamp entirely when `fetch_rc==2`
(pithead:10939-10945). Through the Tor SOCKS proxy, circuit-build timeouts, exit-relay
refusals, and mid-handshake TLS failures return that exact same curl nonzero as a purely
local "Tor daemon down" — but those DO put real egress traffic on the wire. Net effect:
while the transport is degraded, a dashboard-authenticated actor dropping os-check requests
got one real Tor dial per spool drain with zero cooldown, a partial erosion of #1081's
anti-beacon guarantee (`control_os_gate`'s threat-model comment, pithead:10830-10832).

## Fix

Kept the rc-1-vs-rc-2 distinction (that part was correct) and removed only the
zero-cooldown behavior. `rc 2` now claims a second stamp file (`.check-stamp-short`, ~60s)
instead of no stamp at all; `rc 1` keeps the full 10-minute stamp exactly as before. This
bounds the dial rate during a degraded transport without making an operator who's fixed a
genuinely-down Tor daemon wait the full 10 minutes. Updated the comments on both
`gh_release_fetch` and `control_os_check` to state the true property: the short cooldown
bounds the dial rate, it does not claim no real attempt ever happened.

`tests/stack/test_os_update_recovery.sh`: the "immediate retry after a connectivity
failure" case now asserts it IS throttled, including a direct dial-count check (comparing
`curl.log` line counts) proving no second `curl` invocation fires while the short stamp is
live — a stronger proof than the outward "rejected" status alone, since a re-dial that
fails again is *also* "rejected". Added a stamp-mtime-manipulation case proving the short
window expires before the 10-minute one does. This would FAIL on the old zero-cooldown
code — confirmed against a throwaway mutant reverting to it.

## Scope addition from CI

CI on this branch (2878/2) failed two `tests/stack/run.sh` assertions —
"a transport failure is still a failure" and "appliance transport failure is still a
failure" — because both hardcoded the pre-existing expectation that `gh_release_fetch`
returns rc 1 for every curl failure. That expectation was already stale before this PR's
review pass (the rc-1/rc-2 split predates it); the two assertions were asserting the old
contract, not catching a real regression. Renamed both to state the true rc (2) and kept
the property they actually care about — a transport failure still surfaces as a failure to
the operator with the "could not reach the GitHub release API" hint. This block calls
`gh_release_fetch` directly, not `control_os_check`, so it proves only the fetch's own
rc/hint contract; the short-vs-10-minute cooldown mechanics are proven in
`test_os_update_recovery.sh` instead. Confirmed both updated assertions go red if
`gh_release_fetch`'s `return 2` is reverted to `return 1`.

## What was run

- `tests/stack/test_os_update_recovery.sh` standalone (source `lib.sh` + the file): 22
  passed, 0 failed.
- The two updated `tests/stack/run.sh` assertions, extracted and run standalone against a
  small driver (source `lib.sh` + the extracted block): 22 passed, 0 failed.
- Mutation checks confirmed for both files (see above).
- `shellcheck --severity=warning` on `pithead`, `tests/stack/run.sh`,
  `tests/stack/test_os_update_recovery.sh` — clean on `pithead` and the test file; 3
  pre-existing `SC2034` warnings on `run.sh` at unrelated lines (~9343-9483), outside this
  diff's hunks, confirmed unrelated by `git diff` hunk ranges.
- `shfmt -i 4 -d` on all three touched files — no diff.
- `scripts/lint-file-budget.sh` — one pre-existing failure remains
  ("`docs/dev/file-budget.tsv` raises `tests/stack/run.sh`'s ceiling from 10352 to 12115"),
  confirmed present on the branch tip *before* this change too (stale-base divergence
  against `develop-v2`, not introduced here). My own edits kept `run.sh` at exactly its
  already-recorded ceiling (net zero line growth) rather than growing past it.
- Did not run `tests/stack/run.sh` in full, `make lint`, or any KVM/hardware test, per
  instruction.

New commits: `9c0ee4a` (the review-finding fix), `e25c59b` (the CI assertion fix).

Issue #1028, opened 2026-08-16
First boot's slowest step shows one line and no progress — on a USB boot it looks hung for 10-30 minutes


VijitSingh97:
Reported from a real first boot (2026-08-16, HP bench, booted from the stick): the console sits on

```
Loading this build's container images (dashboard.tar.gz) — the slow part...
```

and the operator reasonably read it as hung.

**Why it takes so long.** `load_baked_images` (`pithead:1983`) runs `podman load -i` on each baked archive, and podman unpacks the layers into container storage on `/data`. When the machine is booted from the USB stick — which is every first boot, before the installer has run — **`/data` is the stick**, so this is a multi-gigabyte decompress-and-write to the slowest medium in the box. Ten to thirty minutes is normal on a typical stick. Nothing is wrong; the product just goes silent through the longest wait a new operator will ever have with it, and it is the very first thing they see.

The code already knows this is special — its own comment calls it "the single slowest step of a first or post-update boot, and the one worth narrating" — it just narrates once and then says nothing for half an hour.

**Wanted:** progress the operator can trust.
- Per-archive counter at minimum: `Loading image 1 of 4 (dashboard) …` so a stalled step is distinguishable from a working one.
- Something that moves: `podman load` writes progress to stderr (currently sent to `/dev/null`) — surface a heartbeat, or a periodic "still working — N MB written" from the storage directory, so a silent console is unambiguous evidence of a hang rather than the normal case.
- Say *why* it is slow the first time, in one clause: booting from the stick means unpacking onto the stick, and it gets much faster once installed to the internal disk. That single sentence turns "is this broken?" into "ah, expected."
- The wizard page cannot load until this finishes, so this is also the gap before ANY web feedback exists — the console is the only surface, which makes it worth doing well.

Relates to #978 (only the wizard image is baked; the rest still pull) and to the operator-experience bar for the appliance: this is the first impression the product makes.

VijitSingh97:
Fixed in #1033, merged to `develop-v2` as 497985a.

The slow image load now emits a heartbeat, and fatal errors are routed through `_console` so a failed first-boot service no longer leaves a stale progress line that reads identically to a slow one. The load stays in the foreground and the heartbeat is backgrounded — polling a backgrounded load with `kill -0` would be wrong, since a finished-but-unwaited child is a zombie that still answers.

PR #1308, opened 2026-08-23
fix(os): make the first-boot journal genuinely persistent across interrupts

### LINKED ISSUE #1030: First boot can be interrupted mid-image-load, and the console keeps no record of why

VijitSingh97:
# fix(os): give the persistent journal somewhere real to land

Closes #1030

Branch: `fix/1030-firstboot-journal`
SHA: `fa9ccceb7dceeeb778deb7fc8742dd322461bd54`

## Root cause

`os/rootfs/Dockerfile:246` bakes `Storage=persistent` into `journald.conf.d/pithead.conf`, but
`/var/log/journal` sits on the SAME read-only root that `/etc` does — confirmed by
`os/overlay/pithead-machine-id`'s own comment ("on the read-only root [PID 1] bind-mounts a
transient id... a direct write [to /etc/machine-id] fails EROFS") and
`os/KNOWN-ISSUES.md:180`'s description of "the (volatile, read-only-root) /etc". journald
cannot create/write `/var/log/journal` there, so it silently falls back to `/run/log/journal`
(tmpfs, volatile) for the entire boot — the `Storage=persistent` declaration has no actual
effect. Nothing ever promotes it either: `systemd-journal-flush.service`, the unit that would
normally migrate the runtime ring buffer into persistent storage, runs before `/data` (the only
genuinely persistent partition) is even mounted. A box that hard-resets mid-first-boot — the
reported incident: the container-image load (`load_baked_images`, `pithead:2155`) getting cut
mid-write — is left with literally no journal to diagnose it from, even though the config has
claimed "persistent" the whole time. This matches the issue's own diagnosis
("`journalctl --list-boots` shows only the current boot... Storage=volatile is the reason").

## Fix

Two independent pieces, both called out in the issue's own "suggested work":

1. **`os/overlay/pithead-journal-persist(.service)`** (new) — same early-boot shape as
   `pithead-machine-id.service` (`RequiresMountsFor=/data`, `DefaultDependencies=no`,
   `Before=sysinit.target systemd-journal-flush.service`): bind-mounts a real, baked-empty
   `/var/log/journal` (Dockerfile now bakes the empty mountpoint dir — a bind mount only needs
   an existing directory, not a writable one, so this works even though root is read-only) onto
   `/data/pithead/journal`. Ordered before `systemd-journal-flush.service` so that unit finds a
   genuinely writable target: it both promotes whatever accumulated in the runtime buffer AND
   switches journald to write new entries persistently for the rest of the boot — covering
   exactly the window where the slow, crash-prone container-image load happens. Deliberately
   NOT ordered before `systemd-journald.service` itself (same limit `pithead-machine-id`
   documents for the DHCP DUID): journald starts before `/data` can possibly be mounted, so its
   own earliest messages are unavoidably lost on a reset before this point.
2. **`load_baked_images`** (`pithead:2155`) — the #1028 heartbeat already narrates a load's
   START and its progress; it never narrated the FINISH. Added a
   `"Finished loading <archive> in Ns."` line on success (nothing on failure — the existing
   no-record-on-failure contract, "the next boot must retry", is unchanged), so a journal that
   now survives a reset can distinguish "this archive finished" from "this archive was still
   loading when the reset hit" — the actual record the issue's title asks for.

**Not done** (the issue's own "not established" section, needing real hardware measurement
this fix doesn't have): re-tuning `RuntimeWatchdogSec=20s` for the first-boot window
specifically, or determining which of watchdog/USB-stall/power caused the ORIGINAL reset. The
fix makes the NEXT occurrence diagnosable; it does not diagnose the one already reported.

## Tests run

**Tier 1 (unit, new file `tests/stack/test_firstboot_journal.sh`, 10 assertions):**
`pithead-journal-persist`'s create/bind/idempotent-skip paths (`mount`/`mountpoint`
PATH-stubbed — mounting needs root, matching the convention `tests/stack/test_data_reset.sh`
uses for the same reason) and `load_baked_images`'s new finish line, including that a FAILED
load reports no finish line (so the journal's own words never contradict the
no-record-on-failure contract).

**Static (tier 4, no KVM): `tests/os/verify-image.sh`** gained five checks mirroring the
existing machine-id checks: the persistent-journald config is baked, the `/var/log/journal`
mountpoint directory is baked, the script is baked+executable, the unit is enabled
(`sysinit.target.wants` symlink), and the unit orders before `systemd-journal-flush.service`.
Not run standalone here (needs a built image; exercised as part of the combined KVM build's
own `_build_image`-equivalent step — see below).

**Regression-verified**: stashing `pithead` + deleting the two new overlay files and re-running
the new suite fails exactly the assertions tied to the missing script/line (confirmed before
commit; deleting the untracked overlay files during that check cost an extra recreate step,
noted here in case the same file-recreation trap bites a reviewer replaying this).

**KVM (tier 4) — RUN, with a real verdict.** Combined build (this branch + #1103 + #1050 merged
into a throwaway `kvm-combined-1103-1050-1030` branch off the same base commit, per the
"build/boot at most once, reuse across all three bugs" instruction): `os/build-image.sh` →
`os/rauc/mkimage.sh --dev` → `tests/os/verify-image.sh --test`, 100/100 static checks passing
including the five new journal-persist checks. Booted the image (`virt-install`), and on the
serial console watched this fix's own instrumentation fire live during the real first boot:

```
pithead: Loading this build's container images (dashboard.tar.gz) — the slow part...
  Finished loading dashboard.tar.gz in 7s.
```

Then over SSH:

```
$ findmnt /var/log/journal
TARGET           SOURCE                      FSTYPE OPTIONS
/var/log/journal /dev/vda4[/pithead/journal] ext4   rw,noatime

$ journalctl --list-boots
IDX BOOT ID                          FIRST ENTRY                 LAST ENTRY
  0 c9d0e0f8e63b435eba3bc50590bd03ec Sun 2026-08-23 01:41:16 UTC Sun 2026-08-23 01:41:32 UTC
```

`findmnt` confirms the bind mount is real and backed by `/dev/vda4` (the data partition), not
the root overlay. Then `virsh reset` (an ABRUPT, non-ACPI reset — the closest KVM equivalent to
a watchdog fire or power cut, deliberately not a graceful `virsh reboot`) and, once the guest
came back up on its own:

```
$ journalctl --list-boots
IDX BOOT ID                          FIRST ENTRY                 LAST ENTRY
 -1 c9d0e0f8e63b435eba3bc50590bd03ec Sun 2026-08-23 01:41:16 UTC Sun 2026-08-23 01:41:32 UTC
  0 e08e06ffa9e744ed936bb27b3bfc3544 Sun 2026-08-23 01:41:42 UTC Sun 2026-08-23 01:41:47 UTC
```

**This is the direct repro of the reported defect, inverted**: the PRE-RESET boot
(`c9d0e0f8...`, same boot id and same first/last-entry timestamps as before the reset) is still
there, now at index -1, after an ungraceful hard reset. Before this fix, that boot's journal
lived only in `/run/log/journal` (tmpfs) and a hard reset would have wiped it — `--list-boots`
would show only the new boot. Full log (incl. #1103's hugepages verdict from a separate boot of
the same image) is on the session record; ask the coordinator for the combined-branch worktree
if you want to re-run or extend the battery without rebuilding.

## NOT proven

- The ORIGINAL incident's root cause (watchdog vs. USB stall vs. power) is still not
  established — this fix does not attempt to determine it, per the issue's own scope.
- The KVM leg's hard reset was triggered manually via `virsh reset` at an arbitrary point after
  boot, not synchronized to land mid-`load_baked_images` the way the original incident did (a
  precise mid-load timing hit is impractical to script reliably in the time available). The
  proof therefore demonstrates "a hard reset does not lose the journal," not "a reset landing
  exactly mid-image-load leaves a legible gap between a start and a missing finish line" — the
  latter is a straightforward manual read of the two lines `load_baked_images` now emits, but
  was not separately staged and captured.
- `RuntimeWatchdogSec` retuning and the watchdog-vs-USB-stall-vs-power investigation the issue
  lists as open items remain open; no attempt was made here.


VijitSingh97:
## Review finding

`os/overlay/pithead-journal-persist:23-27` does `mkdir -p` + `chmod 2755` on the persisted
journal directory but never `chown root:systemd-journal "$DATA_DIR"`, even though its own
comment claimed the standard root:systemd-journal shape was "matched here". The parent
chain under `/data` is root:root, and setgid propagates the *directory's own* group, not
its parent's — so the persisted journal ended up group root, and a non-root member of
`systemd-journal` could create it but never `journalctl`-read it back.

## Fix

Added the chown, right alongside the existing chmod. It tolerates failure the same way
that chmod already does (`2>/dev/null || true`): systemd's own `sysusers.d` creates the
`systemd-journal` group and is ordered `Before=systemd-journald.service` upstream, and
journald is already running well before this unit gets its turn in boot — so the group is
expected to exist by then — but a non-appliance sandbox with no such group must not
hard-fail first boot over a directory that still works either way, just root-only-readable
until a later `systemd-tmpfiles` pass catches up. The comment was rewritten to state this
plainly instead of asserting the group unconditionally.

`tests/stack/test_firstboot_journal.sh`: stubbed `chown` in `PATH` (real `chown(1)` would
fail identically as this test's unprivileged user whether or not the fix is present, so
leaving it unstubbed would prove nothing either way) and asserted the script invokes it
with `root:systemd-journal` on the data directory. MUTATION KILL: dropping the chown line
leaves the stub's log empty and this assertion goes red — confirmed against a throwaway
mutant.

`tests/os/verify-image.sh`: added a matching static check on the *baked script's own body*
(`grep` for the chown line), mirroring the existing pattern used for
`pithead-boot`/`pithead-sync`'s baked-content checks. A runtime-created directory can't be
checked for group ownership statically (`/data` isn't part of the base image), but the
shipped script actually containing the chown command can be.

## What was run

- `tests/stack/test_firstboot_journal.sh` standalone (source `lib.sh` + the file): 11
  passed, 0 failed.
- Mutation check: dropping the chown line turns "the data directory is chowned to
  root:systemd-journal" red (10 passed, 1 failed) — confirmed.
- `shellcheck --severity=warning` on `os/overlay/pithead-journal-persist`,
  `tests/stack/test_firstboot_journal.sh`, `tests/os/verify-image.sh` — clean on all three.
- `shfmt -i 4 -d` on all three touched files — no diff.
- `scripts/lint-file-budget.sh` — one pre-existing failure remains (the same
  `tests/stack/run.sh` ceiling-vs-`develop-v2` divergence noise as on other open branches
  right now); untouched by this PR's files.
- Did not run `tests/stack/run.sh`, `make lint`, or any KVM/hardware test, per instruction.
  `tests/os/verify-image.sh` itself needs `sudo` + a built appliance image + loop mounts to
  execute, so the new static check was verified by inspection against the file's existing
  `chk()` convention rather than by running the script.

New commit: `52489a8`.

Issue #1309, opened 2026-08-23
e2e: worker-apply legs expect a synchronous 'applied' — RigForge #344 rigs answer 'accepted' and apply async, so the legs red on a working path


VijitSingh97:
## The gap

The `rigforge-control` worker-apply legs (#513) assert the dashboard's
`/api/control/worker-apply` round trip returns `status: "applied"` with the changed key in
`changed_keys`, at dial time. RigForge changed its control API (its #344): a rig now answers
`"accepted"` immediately and applies asynchronously — its `/status` flips to `"applied"` with
`changed_keys` populated seconds later, once its own apply/restart cycle completes.

Run live against a deployed stack with a borrowed RigForge 1.15.1 rig (the first live run of
this phase — the profile gate in #1301 had kept it from ever starting), the legs go red on
exactly that drift:

```
✗ Worker Inspect edit applied on the rig (#513)      expected [applied], got [accepted]
✗ the rig's /status confirms max_temp_c changed      [] missing [max_temp_c]
✗ worker-apply recorded in the per-worker history    expected [applied], got [accepted]
✗ reversible edit reverted on the rig (#513)         expected [applied], got [accepted]
```

while the path itself works: the control result files are written (`status: "accepted"`, no
error), and the rig's own `/status` shortly after reads `"status": "applied"`,
`"changed_keys": ["max_temp_c"]`, `"source": "control"`, with the rig-side backup minted. The
sibling legs pass — #514 masked-descriptor read, #516 rig-side edit reflected in the feed, and
the #1002a `--rigforge-upgrade` noop convergence.

So the reds are a stale synchronous expectation in the harness, not a product fault — but they
would mask a real failure identically, and the phase cannot tell the two apart today.

## Fix shape

Treat `"accepted"` as in-flight, not terminal: after the dial, poll the rig's `/status` (or
the dashboard's per-worker record) for the returned `change_id` reaching `"applied"` with the
expected key in `changed_keys`, within a bounded window; fail on `"rejected"`/`"failed"`/
timeout. The revert leg gets the same shape.

Worth a product-side decision alongside: the dashboard's per-worker history keeps the
dial-time `"accepted"` verbatim and never converges it to what the rig actually did — the same
out-of-cycle-refresh class as #1233. If the history record stays `"accepted"` forever, Worker
Inspect's history reads as "nothing ever finished" on every RigForge new enough to apply
asynchronously.

[...]

PR #1310, opened 2026-08-23
fix(e2e): the matrix profile gates test membership, not the literal profile string

### LINKED ISSUE #1301: e2e: local-node phase gates compare COMPOSE_PROFILES to the literal 'local_node' — four matrix phases self-skip on every standard local stack
## The gap

Six places in `tests/integration/run.sh` decide "is this a local-node stack" by comparing the
raw `COMPOSE_PROFILES` value against the literal string `local_node`:

- `run.sh:1430`, `run.sh:1591` (hardening), `run.sh:2012` (moved-subnet), `run.sh:2107`
  (rigforge-control) — `!= "local_node"` → skip as "remote mode".
- `run.sh:1123`, `run.sh:1148` — `= "local_node"` → local-only assertions that silently do not
  run when the comparison fails.

`COMPOSE_PROFILES` is a comma-separated list. On the standard local stack it reads
`local_node,local_tari,payout_confirm`, so the strict equality fails on every normally
configured box: the harness reads a fully local stack — dashboard container running, monerod
local — as remote, and the four gated phases self-skip.

## What this cost, concretely

The v1.20.0 pre-cut gate ran `e2e.sh rc/v1.20.0 --mode matrix` with a borrowed rig and came
back green (610 passed / 5 skipped / 0 failed) — with these lines in the harness log:

```
[ITEST] skipping rigforge-control (remote mode: no local dashboard container to drive)
[ITEST] skipping hardening phase (remote mode: no local containers/systemd to exercise)
[ITEST] skipping moved-subnet phase (remote mode: monerod's moved-prefix proxy is one of the named checks)
```

while the stack under test had `COMPOSE_PROFILES=local_node,local_tari,payout_confirm` and a
running local dashboard. `e2e.sh --mode matrix` appends `--rigforge-control` and promises the
control-write legs — on the standard config they cannot run through it at all. The gate proves
less than it looks: a green matrix verdict currently says nothing about the worker-apply write
path, the hardening phase, or the moved-subnet phase.

Relates to #1236, which documents the env-gated skips *inside* the rigforge-control phase;
this gate sits upstream of those and prevents the phase from starting in the first place.

## Fix shape

Membership test on the comma list at all six sites — e.g.

```sh
case ",$(env_on_box COMPOSE_PROFILES)," in *",local_node,"*) local_node=1 ;; esac
```

plus a harness self-test proving a profiles list with extra members still counts as local
(and that a genuinely remote profile set still skips). Respell the membership check and the
self-test must go red — the skip messages themselves are not a guard.

VijitSingh97:
## Summary

Six sites in `tests/integration/run.sh` decided "is this a local-node stack" by comparing the
raw `COMPOSE_PROFILES` value against the literal string `local_node` (`run.sh:1123`, `1148`
[`=`], `1430`, `1591`, `2012`, `2107` [`!=`]). `render_env` (the `pithead` CLI's single env-file
writer) builds `COMPOSE_PROFILES` by **appending** tokens — `local_node`, then
`local_tari`/`payout_confirm`/`tari_payout_confirm` as those features turn on — so a standard
local box renders `local_node,local_tari,payout_confirm`, never the bare token alone. The strict
comparison therefore never matched on a normally configured box, and the `rigforge-control`,
`hardening`, and `moved-subnet` phases (plus two `run_lifecycle` assertions) silently
self-skipped as "remote mode" — including on the v1.20.0 pre-cut gate (610 passed / 5 skipped /
0 failed, with the skip lines proving it).

## Root cause

`tests/integration/run.sh` (six call sites, now converted) called `env_on_box COMPOSE_PROFILES`
and compared the result with `[ = "local_node" ]` / `[ != "local_node" ]` instead of testing
list membership.

## Fix

- Added `has_compose_profile "<profiles-csv>" "<token>"` — a pure comma-list membership test —
  to `tests/integration/scenarios.sh` (mirrors `pithead`'s own
  `remove_deactivated_profile_containers` idiom). Placed in `scenarios.sh` rather than `lib.sh`
  or `run.sh` deliberately: both of those are already at their `docs/dev/file-budget.tsv`
  ceiling with zero slack, and `scenarios.sh` isn't budget-tracked (well under the 400-line
  target even after this addition), so the fix doesn't force an unrelated ceiling bump.
- Swapped all six `run.sh` call sites to `has_compose_profile "$(env_on_box COMPOSE_PROFILES)"
  local_node` (or `! has_compose_profile ...`) — net zero line-count change in `run.sh`.
- Added `tests/integration/selftest-compose-profiles.sh`, a standalone self-test (not sourced by
  `selftest.sh`, for the same budget-slack reason) wired into `make test-integration-selftest`
  and `.github/workflows/ci.yml` alongside `selftest.sh`.

## Mutation kill (stated + verified)

The self-test's load-bearing assertion feeds `has_compose_profile` the real standard-box shape
(`"local_node,local_tari,payout_confirm"`, `local_node`) and asserts it matches. Reverting
`has_compose_profile`'s body to the old literal comparison (`[ "$1" = "$2" ]`) flips that
assertion from PASS to FAIL — verified locally: with the literal comparison restored,
`selftest-compose-profiles.sh` reports `4 passed, 1 failed` (rc 1); with the real fix, `5
passed, 0 failed`. Additional assertions guard: the simple single-token case still matches (no
regression), a genuinely remote profile set still does NOT match (the gate must still skip real
remote stacks), membership is token-exact (not an unanchored substring match), and an empty
profiles string doesn't false-positive.

## What ran (this PR)

- `bash tests/integration/selftest.sh` — 214 passed, 0 failed (unchanged; this file was not
  modified).
- `bash tests/integration/selftest-compose-profiles.sh` — 5 passed, 0 failed.
- Manual mutation test: reverted `has_compose_profile` to the old literal-equality bug in a
  scratch copy, re-ran `selftest-compose-profiles.sh` → 4 passed, 1 failed (rc 1), confirming
  the assertion is load-bearing; restored the fix and re-ran green.
- `shellcheck --severity=warning` and `shfmt -i 4 -d` on every touched shell file (`run.sh`,
  `scenarios.sh`, `selftest-compose-profiles.sh`) — clean.
- `bash scripts/lint-file-budget.sh --self-test` and `bash scripts/lint-file-budget.sh` — both
  pass; no ceiling raised, no file grew past its recorded ceiling.
- `bash scripts/lint-topology-classes.sh` — clean (no real-looking hostnames/paths introduced).
- Did NOT run `tests/stack/run.sh`, `make lint` (whole), or the live matrix e2e — out of scope
  for this task per the coordinator's instructions.

## Owed (separate follow-up, on the coordinator)

**The live matrix e2e run is owed.** This PR is verified at the unit/self-test level only —
proving the membership logic is correct in isolation. It has NOT been proven against a real
bench: that the `rigforge-control`, `hardening`, and `moved-subnet` phases actually now RUN
(rather than skip) on a real local-mode box's rendered `COMPOSE_PROFILES`, and that they behave
correctly once running (the `rigforge-control` phase separately has known-stale assertions,
tracked in #1309). The coordinator is scheduling that live-rig proof leg (`tests/integration/
e2e.sh <ref> --mode matrix` on a borrowed rig) as a follow-up; this PR should not be treated as
closing the coverage gap end-to-end until that run comes back green with the three previously-
skipped phases visibly executing (not skipped) in the log.

Closes #1301

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -260,7 +260,9 @@ jobs:
         # Pure-logic checks for the tests/integration/ harness (config rendering, matrix
         # coverage, redaction). The LIVE matrix (tests/integration/run.sh) needs a real test
         # server and runs as a gated/manual release gate (#54), not on every PR.
-        run: bash tests/integration/selftest.sh
+        run: |
+          bash tests/integration/selftest.sh
+          bash tests/integration/selftest-compose-profiles.sh
 
   compose:
     name: Compose config + security hardening
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ test-compose: ## Validate docker-compose.yml interpolation + hardening invariant
 
 test-integration-selftest: ## Integration harness pure-logic self-test (no server needed)
 	bash tests/integration/selftest.sh
+	bash tests/integration/selftest-compose-profiles.sh
 
 test-fakes: ## Fake-daemon contract test — real dashboard clients vs controllable fakes (no docker)
 	uv run --locked --project build/dashboard --extra test python -m pytest tests/integration/fakes -q
--- a/tests/integration/run.sh
+++ b/tests/integration/run.sh
@@ -1120,7 +1120,7 @@ run_lifecycle() {
     # Piggybacks the pool-flip apply below, which always runs ensure_directories -> ensure_owner.
     # Local mode only (has data dirs); a stub can't create a foreign-uid inode, so this is tier-4.
     local own_dir own_probe=""
-    if [ "$(env_on_box COMPOSE_PROFILES)" = "local_node" ]; then
+    if has_compose_profile "$(env_on_box COMPOSE_PROFILES)" local_node; then
         own_dir="$(env_on_box DASHBOARD_DATA_DIR)"
         if [ -n "$own_dir" ]; then
             own_probe="$own_dir/.itest-owner-probe"
@@ -1145,7 +1145,7 @@ run_lifecycle() {
 
     # Node-down failover (#31): stop monerod -> status non-zero (node down), dashboard rejects
     # workers (xmrig-proxy stopped) -> start monerod -> readmitted -> status 0 again.
-    if [ "$(env_on_box COMPOSE_PROFILES)" = "local_node" ]; then
+    if has_compose_profile "$(env_on_box COMPOSE_PROFILES)" local_node; then
         it_step "stopping monerod to exercise node-down failover…"
         rx "docker compose stop monerod" >/dev/null 2>&1
         wait_for 120 5 "status to report node down" _pred_status_down || true
@@ -1427,7 +1427,7 @@ run_fault_injection() {
     IT_CURRENT_SCENARIO="fault-injection"
     echo ""
     it_log "── fault-injection phase ───────────────────────────"
-    if [ "$(env_on_box COMPOSE_PROFILES)" != "local_node" ]; then
+    if ! has_compose_profile "$(env_on_box COMPOSE_PROFILES)" local_node; then
         it_warn "skipping fault injection (remote mode: no local monerod to break)"
         return 0
     fi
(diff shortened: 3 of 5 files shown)


VijitSingh97:
Review-verdict note for the record: a fresh-context reviewer blocked this PR claiming it deletes recently-merged appliance work. Refuted by re-derivation: the claim came from a two-dot comparison of this branch's tree against the develop-v2 tip that moved after the branch was cut — the PR's actual merge-base diff (the Files tab) touches five files with pure additions (the membership helper, six call-site swaps, the new self-test, and its CI/Makefile wiring), deletes nothing, and merges without conflict, so the newer appliance content is untouched by the merge. The same reviewer's analysis of the actual fix confirmed the helper's token semantics, all six gate directions, and the mutation-killing self-test. Merging on green CI.

Issue #1325, opened 2026-08-23
Two suite verdict lines are nondeterministic, weakening the split-proof multiset check


VijitSingh97:
# Two suite verdict lines are nondeterministic, weakening the split-proof multiset check

## What

The #1105 domain-split PRs prove a move changed nothing by diffing the sorted multiset of
per-test verdict lines before and after. That check is only as strong as the lines being
deterministic — and two are not:

1. An SSH-keygen test prints a fresh random ed25519 fingerprint in its passing verdict.
2. `load_baked_images`' heartbeat test prints a real `date +%s`-measured elapsed time in
   its passing verdict (seen as `(1s)` vs `(2s)` across two runs, same PASS both times).

Each split PR has had to hand-wave "empty diff modulo these N nondeterministic lines," and
N has grown from 1 to 2 as the splits reached new domains. Every added exception is a line
the multiset proof can no longer vouch for — a real drop or reorder hiding behind one of
them would not be caught.

## Why it matters

The multiset diff is the load-bearing evidence that a mechanical move is truly mechanical.
Nondeterministic verdict text erodes it silently: the more lines the reviewer must exclude,
the weaker the guarantee, and the exclusions are applied by hand per PR rather than by a
rule anyone can audit.

## Fix direction

Make the verdict text deterministic where the value isn't the thing under test: print a
fingerprint's presence/shape rather than its bytes; assert the heartbeat elapsed time is
under a bound rather than printing the measured value in the verdict line (or route the
measured value to a detail line the multiset check already ignores). Either restores a
zero-exception multiset diff. If some line genuinely must carry a variable value, add a
single documented normalizer to the proof tooling (mask known-nondeterministic fields)
so the exclusion is a rule, not a per-PR footnote.

Surfaced across the release/doctor/monero/tor-network split PRs; not blocking any of them
(each move was independently confirmed byte-level), but worth closing before the
sandbox-coupled cluster splits, which will lean hardest on the multiset proof.

VijitSingh97:
[...]

PR #1252, opened 2026-08-22
test(stack): extract the shared test harness into tests/stack/lib.sh (#1105 Phase 1)

VijitSingh97:
The first module of the tests/stack/run.sh split (#1105 Phase 1) — a MECHANICAL MOVE ONLY, no behaviour change. The shared harness (`ok`/`bad`/`assert_eq`/`assert_contains`/`assert_not_contains`/`assert_rc`, the PASS/FAIL counters, `run_sourced`, the `SANDBOX` fixture, `make_stubs`) moves verbatim into a new `tests/stack/lib.sh` (76 lines, matching `tests/integration/lib.sh`'s existing convention); `run.sh` sources it near the top. Nothing else moves — individual test groups are later modules, each its own paired PR.

run.sh drops 8381 → 8319 lines; `docs/dev/file-budget.tsv`'s ceiling follows (the ratchet only allows shrinking). `Makefile`'s `lint-sh` now also shellchecks the new lib (run.sh itself stays excluded, #1206).

**Byte-identical execution — proven independently, twice.** The full develop-lane suite runs identically before and after: **1836 passed, 0 failed** on both, with an identical ordered 1996-line sequence of section headers and test verdicts. (1836 is the develop-lane count; the ~2862 seen elsewhere this program is develop-v2's larger appliance suite.) The agent's own before/after (real git worktrees) matched; the coordinating session then re-ran an INDEPENDENT clean before (origin/develop @ a real `git worktree`) and after (this branch) — both 1836/0. A first proof attempt via `git archive` into non-git scratch dirs gave a false 3-failure diff because release-smoke #1068 needs a real `.git` at `$ROOT` — a verification-methodology flaw, corrected, not a regression.

No mutation applies — a pure code move has nothing new to fail; the byte-identical suite proof is the verification.

**Follow-ups**: the paired develop-v2 PR (same boundary) lands next; CI wiring for the file-budget lint is still owed a workflow-scoped push (with #1228). The agent produced a proposed ~22-module boundary map for the rest of Phase 1 — posted as a review comment for the operator to re-cut along whatever functional lines (all-doctor / all-control-channel / all-release) they'd rather review by, since the first pass is size-greedy section adjacency.

Part of #1105 (Phase 1, module 1 of ~22).

Issue #378, opened 2026-07-08
`pithead rotate-secrets`: regenerate the stack's internal credentials in one command


VijitSingh97:
The stack's internal secrets are generated once and live forever: the Monero RPC password (`generate_node_password`, `pithead:1541`, persisted into config.json at 1609), the auto stratum password (`pithead:1832-1836`, preserved from `.env` on every re-render), and the xmrig-proxy control-API token (`PROXY_AUTH_TOKEN`, preserved in `load_preserved_state`, `pithead:1985,1994`). The only rotation that exists is `rotate-dashboard-onion` (`pithead:2221`). After a suspected leak — a backup that left the box (see the encrypted-backups issue), a `.env` pasted into a bug report, an ex-operator — there is no supported way to invalidate these except hand-editing files and knowing which containers to recreate.

## Proposal
One subcommand, `pithead rotate-secrets [-y]`, that regenerates every internal secret that is safe to regenerate, re-renders `.env`, recreates the affected containers, and prints exactly what the operator must do afterwards. Rotated: the Monero RPC password (local mode only — in `monero.mode: "remote"` the credential belongs to the remote node and is skipped with a message), the stratum password when `p2pool.stratum_password` is `"auto"` (skipped with a pointer to config.json when it is a literal, skipped silently when empty/disabled), and `PROXY_AUTH_TOKEN` (always). The stratum rotation is called out loudly: every rig's `pass` must be updated or it is rejected — the summary prints the new value via the existing `announce_stratum_auth` (`pithead:851`). RPC password and API token are internal-only: zero operator action. Onion rotation stays in `rotate-dashboard-onion`; this command does not touch keys or addresses.

## Implementation notes (drafted 2026-07 against develop @ v1.2.2)
**Depends on / interacts with:** #356/#343 (`rotate_dashboard_onion` — the pattern to mirror for confirm/render/restart), #152 (stratum password modes), #153 (PROXY_AUTH_TOKEN fail-closed).
**Pre-read:** `pithead:1523-1620` (credential generation + persist), `pithead:1754+` (`parse_and_validate_config`, esp. 1832-1842 stratum and 1855-1870 RPC creds), `pithead:1984-2002` (`load_preserved_state`), `pithead:2221-2267` (`rotate_dashboard_onion`), `pithead:3276-3337` (`main` dispatch).
### Steps
[...]

Issue #375, opened 2026-07-08
Payout-wallet tamper tripwire: alert + banner when the mined-to wallet changes


VijitSingh97:
The highest-value tamper against a mining stack is swapping `monero.wallet_address` in `config.json`: the flow is config.json → `parse_and_validate_config` (`pithead:1761`) → `.env` `MONERO_WALLET_ADDRESS` (`pithead:2494`) → p2pool `--wallet` and xmrig-proxy `-u` (`docker-compose.yml:216,287`), and every future block reward goes to the attacker. Nothing watches for this today: `pithead apply` previews the change (`describe_change`, `pithead:2979-2982`, already DEST-flagged) but an attacker who edits config.json and runs `apply -y` — or recreates the container directly — triggers no alert, no log, nothing on the dashboard.

## Proposal
The dashboard persists the last-seen payout wallet in its existing SQLite `kv_store` (`storage_service.py:119`) and raises a loud edge alert when the observed wallet differs: a Telegram message (old/new, both truncated to 8 chars) plus a warning badge in the top bar for 72 h after the change. The observed wallet is what p2pool actually reports in its local stats (the `stratum.wallet` field already parsed into the state used at `views.py:431`) — ground truth of where rewards go — with the `MONERO_WALLET_ADDRESS` env as fallback. The baseline lives in SQLite, not in AlertService memory, because `apply` recreates the dashboard container and an in-memory baseline would silently re-seed to the attacker's wallet. The alert fires on every change, including a legitimate `pithead apply` — it is a confirmation, not only an intrusion signal. On the pithead side, `apply` upgrades the generic y/N destructive confirm to a typed confirm for wallet changes: the operator must type the first 8 chars of the new address.

## Implementation notes (drafted 2026-07 against develop @ v1.2.2)
**Depends on / interacts with:** #349 (config audit log — complementary; this is the one high-value field watched at runtime), #33 (dashboard config editor — a wallet edit through it must still trip this).
**Pre-read:** `build/dashboard/mining_dashboard/service/alert_service.py` (edge pattern: `_disk_edges` line 271, event keys line 75-94), `service/storage_service.py` (kv_store), `service/data_service.py:820-830` (`alert_service.process(...)` call site), `web/views.py:706` (`build_badges`), `pithead:2964-3013` + `pithead:3193-3222` (describe_change / apply confirm).
### Steps
[...]

PR #1332, opened 2026-08-23
test(stack): split the secrets domain into its own file (#1105 Phase 1, develop-v2)

VijitSingh97:
## Summary

Extracts the secrets domain (#1105 Phase 1, develop-v2 lane, per the #1252 domain map) out of
`tests/stack/run.sh` into `tests/stack/test-secrets.sh` — a mechanical move, no behaviour change.
Stacked on `feat/1105-backup-v2` (this branch is `feat/1105-secrets-v2`, based on it).

Two non-contiguous clusters move, in their original relative order:

- Apply's secret propagation/redaction, the payout-wallet typed confirm, and secret-file
  permissions (252 lines): `apply preserves secrets + propagates` (telegram bot token/webhook/ntfy
  secrets never echoed in the change preview, alongside the rest of apply's field-propagation
  regression that this section also carries), `payout-wallet change needs a typed confirm` (#375 —
  both chains, sequential per-chain prefixes, `-y` still bypasses for automation), and `secret files
  are owner-only from creation` (#368 — the `persist_node_credentials` umask plus a full apply).
- `rotate-secrets` end to end (101 lines, #378): regenerating the local Monero RPC password, the
  "auto" stratum password, and `PROXY_AUTH_TOKEN`; recreating containers via `compose up` (never
  `restart`); skipping what it must (remote-mode RPC, a literal stratum password); a declined prompt
  changing nothing; and a failed recreate leaving the retry marker plus recoverable owner-only safety
  copies of the old values.

## Boundary, coupling scan, and the leave-behind finding

Per the #1252 map's "cuts follow content, not markers" rule and the operator brief's two coupling
classes, both checked and cleared — see `test-secrets.sh`'s own header for the full writeup:

- **$C control-sandbox coupling**: none. Grepped both clusters for a broad set of control-channel
  identifiers (`$C`, `run_pending`, `control_config`, `build_control_sandbox`,
  `provision_control_runner`, `control-run-pending`, `CTRL_LOG`, `REQS`, `RESULTS`, `STAGED`,
  `AUDIT`, `MASKED`) — zero hits. The only CLI verbs invoked anywhere in this file are the plain
  `./pithead apply` and `./pithead rotate-secrets`, run directly against the `$V` sandbox — never
  through the control channel's dispatch/confirm-gate path. The "typed confirm" here is `apply`'s
  own interactive prompt (type back the wallet's first 8 characters); it is unrelated to the
  control-channel's separate confirm-gate cluster (`black-box: dashboard control channel (#33)` and
  its neighbors), which stays in `run.sh`, untouched.
- **Shared-file coupling**: none. Both clusters operate against the shared `$V` sandbox, which is
  fine per the map's rule (a mutating shared sandbox, not exclusive to one domain) — each already
  self-seeds before its first use (`seed_env` at the top of nearly every sub-block), so neither
  depends on whatever state a prior, different-domain section left in `$V`. Confirmed: extracting
  `test-backup.sh`'s black-box cluster from between these two secrets clusters (it used to sit
  between them) changes nothing here, since backup's fixtures never touch `$V`.
- **Leave-behind, code-checked (not markers)**: the "xmrig-proxy entrypoint" pair (unset/set stratum
  access-password, then the TLS cert-flag pair, #152/#261) sits immediately after "rotate-secrets
  failure path" in `run.sh`, under the same run of secrets-domain markers. It is rig/worker content —
  it drives `build/xmrig-proxy/entrypoint.sh`'s stratum-password and TLS-keypair flag rendering, not
  secret storage/rotation/confirms — and the `test-rig-worker.sh` cut's own header already declined
  it as "not this domain" and left it in `run.sh` pending this cut's decision. Code-checked again
  here: it does not belong to secrets either. Line count corroborates the split: this file lands at
  353 lines of moved content against the #1252 map's ~359-line estimate for `test-secrets.sh`;
  folding the xmrig-proxy pair in as well would land at 393, well past it. It stays in `run.sh`,
  still unclaimed by either domain (most likely a future rig/worker-domain follow-up).

## Re-derivations

- `$V` / `$WALLET`: `lib.sh`'s `build_val_sandbox()` sets both; `run.sh`'s "config validation"
  black-box calls it once, far earlier than the sections below that read them — that section stays
  in `run.sh` (a generic multi-field validator, not a secrets concern). `build_val_sandbox()` is
  idempotent, so calling it again here is a safe no-op re-affirm, the same re-derivation
  `test-monero-tari.sh`/`test-rig-worker.sh`/`test-lifecycle.sh`/`test-dashboard.sh` use.
- `$DOCKER_LOG`: this file's own first section is `$DOCKER_LOG`'s original definition site
  (`$V/docker.log`) — every other domain file that reads it re-derives its own copy instead of
  reaching back here (per each of their own header notes). Set again explicitly here too, matching
  their defensive convention.

## Before/after proof

- **Before** (same clean `origin/develop-v2` baseline used for the backup PR,
  a separate pristine worktree, detached at `8c0d0ea`): **2903 passed, 0 failed**.
- **After** (this branch, stacked on `feat/1105-backup-v2`): **2903 passed, 0 failed**.
- Sorted-multiset diff against the same before-baseline: 3210 lines each side, 9 lines differ, all
  explained, none a test add/drop/change or a count change — the same categories as the backup PR:
  - 1 line: the ed25519 host-key fingerprint (#1325, regenerated per run by design).
  - 1 line: a `mktemp`-random tmp dir name in an "ingredients manifest" path.
  - 4 lines: `<worktree-path>/...` prefixes differing because the two runs used two different
    worktree directory names (`pithead-v2bs-before` vs `pithead-v2bs`).
  - 3 lines: `run.sh: line NNNN: CONFIG_FILE: readonly variable` — same statement, line number
    shifted (3439/3439/3471 → 2608/2608/2640) because `run.sh` is now 349 lines shorter.

## Lint

- `shellcheck --severity=warning tests/stack/test-secrets.sh` — clean.
- `shfmt -i 4 -d tests/stack/test-secrets.sh` — clean, no diff.
- `shfmt -i 4 -d tests/stack/run.sh` — clean, no diff.
- `shellcheck` on `run.sh` itself skipped locally per the #1206 memory-cap note; CI's dedicated job
  arbitrates it.

## File budget

`docs/dev/file-budget.tsv` regenerated via `scripts/lint-file-budget.sh --generate` after `git add`
of the new file:

- `tests/stack/run.sh`: 6760 → 6411 (confirmed exact match against `wc -l`).
- `tests/stack/test-secrets.sh`: 401 (new row, confirmed exact match against `wc -l`).

## Stacking

Second of two PRs (#1105 Phase 1 backup+secrets). Based on `feat/1105-backup-v2` — merge that one
first.


VijitSingh97:
HELD — not merging on a red check. The failing "Shell tests" check is not this PR's content: `make lint-sh` is killed mid-step ("Terminated / The operation was canceled") about two minutes in, reproducibly across a rerun. Root cause filed as #1333: the recipe passes every shell file to ONE shellcheck process, and the #1105 split's growing `tests/stack/test-*.sh` set plus a still-large `run.sh` now exceeds what the runner allows. Three rearrangements were measured locally and each trades the crash for false SC2034 findings (cross-file variable visibility) or still OOMs — evidence in the issue. The split content here is proven independently: full suite 2903/0 before and after, byte-diffed against the parent branch, and adversarially reviewed with no blockers. This waits for #1333.

Issue #1231, opened 2026-08-22
XvB decision table: the Odds column still dashes on a disabled box — no stable published data to vendor


VijitSingh97:
Split out of #1214 (PR #1230), which lit up the decision table's reward columns on a disabled box via a vendored published table, and deliberately did NOT do the same for the Odds / 30d column.

Mechanism: `win_odds_day` needs the round-stats cache, which only the `ENABLE_XVB`-gated winners sync fills (#163's no-egress rule) — so a never-enabled box dashes it for the same structural reason the reward columns dashed. The reason the reward fix does not transfer: win-odds data is live competitive information (qualifier counts, round frequency shift with population dynamics), and there is no stable published table to vendor without showing misleadingly stale odds as if current.

If this is ever worth fixing, candidate shapes worth evaluating honestly (none obviously right):
- Show the column's meaning without a number ("computed after first sync") instead of a bare dash — a wording fix, no data problem.
- A wide, clearly-labeled historical band from the delivery study's archived round data, presented as "historically X–Y" rather than a current figure — only if the study's data actually supports a stable band.

Constraint carried over from #1214: whatever ships must not introduce a second differently-derived estimate and must not dial out while disabled.

VijitSingh97:
Adjacent change, no action needed here: PR #1341 (for #1316) rebuilds this table as per-tier blocks. The Odds figure stays honestly empty while XvB is off — the reasoning in this issue still stands, there is no stable published data to vendor — but its *cost* drops a lot. It used to be a whole dead column in a table that already could not fit its card; it is now one item on a wrapping line inside the tier block.

So the width argument for closing this one is largely gone. What is left is the original question of whether an empty Odds field reads as "unknown" or as "zero", which is a wording question rather than a layout one.

VijitSingh97:
Closing — shipped in #1718 (squash-merged to `develop-v2`; verdict, negative-control re-run and ruling-8a legs on the PR). Taken as the wording fix only: `needs XvB enabled` / `awaiting sync` in the Odds cell, nothing derived. The historical odds band this issue also floated was deliberately NOT taken and is not filed; it wants its own evidence.

Issue #991, opened 2026-08-14
Dashboard grid wastes vertical space: a row is as tall as its tallest card, and short cards trail blank


VijitSingh97:
On the Advanced view's "Your stack" section, the top grid row pairs a ~190 px card (Earnings — Expected vs Actual) with a ~680 px one (P2Pool Earnings (Estimated), which carries a tabbed table, a long-run table, a price footer, and ~10 lines of explainer prose). The row is as tall as the tallest card, so the two shorter neighbours trail 300–500 px of blank, and the next row starts below all of it. The section's second row holds two cards in a three-column grid, leaving another empty slot. Roughly half the section is whitespace.

The stylesheet already names the mechanism and its limit:

- `.grid` is `repeat(auto-fit, minmax(350px, 1fr))` with `align-items: start` (`dashboard.css:114-121`) — #817 stopped short cards *stretching* into empty bordered boxes, but "grid rows are as tall as their tallest card" still holds.
- `dashboard.css:569-572` (the raffle-wins cap) states it outright: "#817's align-items:start stops the stretch, not the row height. Cap the list at roughly six rows and scroll the rest." — a per-card point fix that doesn't generalize.

Contributing shapes, for whoever picks this up:

1. **Card heights differ by ~4x within one section**, mostly from in-card explainer prose (the P2Pool Earnings card's two paragraphs are the single biggest block). Moving long explainers into tooltips/collapsibles shrinks the tallest card, which shrinks every row it anchors.
2. **The layout is one-dimensional**: rows, not columns. Column-based packing (CSS `columns` on the section, or masonry when available) lets unequal cards stack tight — but changes reading order to vertical, which the section labels (`grid-column: 1 / -1`, `dashboard.css:126`) would need to survive.
3. **Card order ignores height**: pairing tall cards with tall neighbours (and short with short) within each section is the zero-risk half-measure.

Related: #940 (the dashboard overhaul umbrella — whatever layout it lands should make this class of gap impossible, not re-cap it card by card); #817 (the earlier stretch fix this builds on).

VijitSingh97:
Landed in #1188 — both low-risk shapes from the issue: the tallest card's prose collapsed into a native details disclosure, plus height-aware ordering within the section. Measured in a real browser at 1280px: section height 973px → 704px (−27.6%), per-row trailing blank ~492px → ~210px. The masonry/columns redesign stays with #940.

VijitSingh97:
[...]

Issue #1040, opened 2026-08-16
Dashboard never shows whether the Monero and Tari nodes are local or remote


VijitSingh97:
The stack can run its Monero and Tari nodes either **locally** (containers on this machine) or **remotely** (a third-party or fleet-shared node — `monero.remote`, `tari.mode: remote`). Those are very different operational situations: what "node down" means, what the disk budget is, what a resync costs, and whether the machine is even capable of mining on its own. The dashboard currently renders them identically.

**Ask (two places):**

1. **The network diagram** — show each node as local-to-this-machine or remote, so the topology on screen matches the topology in the config. Right now the picture implies everything is on the box.
2. **The per-node cards** — carry the same distinction on the Monero and Tari cards themselves, so an operator reading a single card knows what they are looking at without cross-referencing the diagram or the config.

**Why it matters beyond cosmetics:** a remote node's health is somebody else's problem to fix, a local one is yours. An operator triaging a stall needs to know which, immediately. It also makes the remote-node feature discoverable — today the only way to know a box is on a remote node is to read `config.json`.

**Where the state already exists:** the mode is already resolved server-side (`monero.mode` / `tari.mode`, and the remote host/port under `tari.remote`), and `/api/state` is the natural carrier — check whether it already exposes enough to render this before adding new fields.

Do not surface the remote host/port in a way that leaks it into screenshots or support bundles without thought — a fleet-shared node address is mildly sensitive. Showing "remote" plus the endpoint only on the card's detail view is probably the right split, but confirm against how the existing cards treat similar values.

VijitSingh97:
Done in PR #1351, merged to `develop-v2`.

Both places asked for: the diagram captions `monerod` and `tari` local or remote, and both node
cards carry a **Node** row saying the same. Only those two nodes carry a location — every other
node in the graph is always this machine's own.

On the issue's own guidance, three notes:

**"Check whether `/api/state` already exposes enough before adding new fields."** It did not, and
the near miss is worth recording: `sync.monero.mode` looks like the answer but is the *pruning*
mode — `Pruned`/`Full`/`Unknown`. It reads `Unknown` for a remote node, which is why a remote node
looked like a gap in the data rather than a fact about the node. The locality itself was already
derived server-side (`monero_is_local()` / `tari_is_local()`, both already driving RAM budgeting
and the log-scrape-vs-RPC split) but stopped there. So the fix was carrying it to the browser, not
computing anything new.

[...]

PR #1341, opened 2026-08-24
fix(dashboard): rebuild the XvB decision table as per-tier blocks (#1316)

### LINKED ISSUE #1316: XvB decision table: negatives render at 8 dp, ranges overflow the card, and the Net colouring is ambiguous
On the XvB tab of the earnings card, the per-tier decision table pans sideways inside its
`.table-scroll` wrapper: at card width you see the last two columns or the first four, never a
whole row. The two that end up visible are the widest ones, because every cell in them is a
`low … high` band — a single Net cell reads like
`-0.00895815 XMR … -0.00191815 XMR`.

Five things, smallest first. (1) and (2) may be enough on their own — measure before doing (5).

### 1. Negative amounts render at 8 decimal places — a formatter bug

`coinDp` ([logic.mjs:378](build/dashboard/mining_dashboard/web/static/logic.mjs:378)) picks the
decimal count from the raw value, not its magnitude:

```js
function coinDp(value) {
  return value >= 1 ? 4 : value >= 0.001 ? 6 : 8;
}
```

Every negative fails both comparisons and falls through to 8. Re-derived against the real module:

```
 21.2706 -> 21.2706 XMR       (4 dp)
 -1.3802 -> -1.38020000 XMR   (8 dp — four trailing zeros carrying nothing)
```

The Net column is where the dashboard formats negative coin amounts, so that is where it lands: a
loss is always the longest string in its row, for no reason. `Math.abs(value)` in `coinDp` fixes
it at the shared function, and every red cell gets shorter.

### 2. Collapse the ranges to one number

`fmtRange` ([components.mjs:507](build/dashboard/mining_dashboard/web/static/components.mjs:507))
renders `low … high` for both the Study column and Net. Two twelve-character numbers and an
ellipsis is most of the table's width. Show the midpoint instead.

Trade-off worth naming: the study band's spread (33% of face, CI 28–39%) is currently legible at
a glance and would stop being so. Keep the band in the cell's `title` — that column already
carries an explanatory tooltip.

The same range appears in `#xvb-fiat-line` under the table and should collapse with it.

### 3. Two nets: what XvB says, and what the study says

One Net is computed today, with the precedence yours > study > face
([logic.mjs:345](build/dashboard/mining_dashboard/web/static/logic.mjs:345)). The face-value net is
never shown — the reader sees XvB's published reward and their own cost in adjacent columns and
has to subtract. Split it: **Net (XvB says)** and **Net (study)**, each a single figure after (2).

### 4. The colouring is ambiguous, and it is not one scheme

Two separate problems in the same row:

- The Net verdict has three states
  ([logic.mjs:357](build/dashboard/mining_dashboard/web/static/logic.mjs:357)): red when even the
  optimistic end loses, green when even the pessimistic end profits, and **plain text** when the
  band straddles zero. Plain is also what an uncoloured column looks like, so "this tier could go
  either way" is indistinguishable from "this cell carries no verdict". `status-warn` already
  exists and would say it.
- `XvB says` is `c-purple` on every row regardless of sign. So one row shows purple, white and red
  across three money columns, with three unrelated colour meanings.

Only the red and neutral states have been seen in practice — confirm the green path still renders
correctly while fixing this.

With two Net columns (3), settle the colour rule once and apply it to both.

### 5. Fit the row without panning

`.table-scroll` + `white-space: nowrap`
([dashboard.css:620](build/dashboard/mining_dashboard/web/static/dashboard.css:620)) was written
for the workers table — its numeric right-align rules still target `nth-child(4)`–`(7)`, which are
the *workers'* columns and only coincidentally land on money here. Six nowrap money columns do not
fit an earnings card.

Once (1) and (2) shrink the cells, measure before rebuilding anything. If it still does not fit,
one block per tier stacked vertically reads better than a sideways pan, and
`stat-grid`/`StatCard`/`MoreStats` already exist for that shape.

### Where

- [logic.mjs](build/dashboard/mining_dashboard/web/static/logic.mjs) — `coinDp` (378), `xvbDecisionRows` (330)
- [components.mjs](build/dashboard/mining_dashboard/web/static/components.mjs) — `XvbDecisionTable` (501)
- [dashboard.css](build/dashboard/mining_dashboard/web/static/dashboard.css) — `.table-scroll` (620)
- Tests: `tests/frontend/logic.test.mjs` (367–427), `tests/frontend/components.test.mjs` (~650)

Related: #940 (dashboard UI overhaul), #1231 (the Odds column dashes while XvB is off — an empty
column costing width in exactly this table).

VijitSingh97:
Closes #1316.

## What the operator sees now

Each donor tier gets a block instead of a row in a table that panned sideways: the two net
verdicts as coloured stat cards, their inputs on a line beneath that wraps.

Measured by rendering the real component in **the operator's state — XvB OFF**, not with XvB on.
That distinction is the whole reason this issue exists: #991 measured its reorder with XvB on and
got it wrong. With XvB off, `build_xvb_calc` serves the vendored published fallback,
`realization` is None so the study band is used, and the odds feed has nothing.

Before (XvB off, 200 kH/s), one row of six:

```
| Whale (100.00 kH/s+) | — | 3.6500 XMR | 4.6700 XMR | 1.3076 XMR … 1.8213 XMR | -2.34240000 XMR … -1.82870000 XMR |
```

After, one block per tier:

```
WHALE (100.00 KH/S+)
  Net (XvB says)  1.0200 XMR  (green)     Net (study)  -2.0855 XMR  (red)
  Cost 3.6500 XMR · XvB says 4.6700 XMR · Study est. 1.5645 XMR · Odds / 30d —
```

The widest figure goes from 33 characters to 13.

## The four fixes

1. **`coinDp` compared the signed value**, so every negative failed both `>=` tests and fell
   through to 8 dp. A loss was always the longest string in its row, carrying four trailing zeros
   that said nothing. Now compares magnitude. `coinTriplet` is unaffected — it only ever passes
   positives.
2. **Bands collapse to their midpoint**, with the band moved into the cell's tooltip. The fiat
   line collapses with them.
3. **The face-value net is computed in its own right** (`netFace`) and shown beside the
   study/measured net. The reader had to subtract two adjacent columns by eye to get it.
4. **A band straddling zero now renders `status-warn`.** As plain uncoloured text it was
   indistinguishable from a cell carrying no verdict at all. `c-purple` comes off the XvB-says
   figure, which had been colouring every row regardless of sign.

## Why the layout changed too

The issue allowed stopping at 1-3 and said to measure first. I did, and they are not enough.

Collapsing the ranges takes the row from 115 characters to 83, which still needs **530-830px**
depending on the glyph advance (0.95rem tabular digits are ~8-9.1px; 530px assumes an implausibly
narrow face). The card's content box is:

| layout case | content box | 83-char table fits? |
|---|---|---|
| minimum-width column | 308px | no |
| app's own 1400px max container (3 x 453px) | 411px | no |
| widest point of the two-column band | 492px | no |
| degenerate single-column window | 677px | only at an implausible glyph width |

**Rejected alternative, for the record:** `table-layout: fixed` plus dropping
`white-space: nowrap`. Six fixed columns across 308-411px is 51-68px each, about six characters —
every cell becomes a stack of wrapped fragments and the tier name alone takes three lines. It
stops the panning by making the table unreadable.

So: a block per tier, using the existing two-up `.stat-grid`, with the inputs on a plain `<p>`
that wraps natively — the failure being fixed was a line that would not wrap. **This adds no CSS
at all:** the tier name reuses `.est-heading`, and the separators are real `·` characters, which
survive copy-paste and a screen reader where a flex `gap` would not.

One note on item 3, since it is the change that most pushes width: two net cards rather than one
costs nothing *here*. `.stat-grid` is `1fr 1fr`, so a single card would leave half the row empty.
In a table it would have been the expensive choice; in a block it is the free one.

## Why the file moves are in here

The file-budget ratchet only ever lowers ceilings, so growing `components.mjs` and `logic.mjs` was
not an option. The XvB code moves into `xvblogic.mjs` / `xvbview.mjs` with its own
`xvbview.test.mjs`, matching the existing workerlogic/workerview convention. `EstTable` comes out
into `esttable.mjs` rather than being copied — it is "the one shape every earnings tab presents
its estimate in", and two copies would have drifted into two earnings tables that disagree. Every
touched file now sits further below its ceiling than before.

## Testing

`node --test dashboard/tests/frontend/*.test.mjs` → **445 pass, 0 fail**. `make lint-js`,
`scripts/lint-operator-strings.sh` and `scripts/lint-file-budget.sh` all clean.

New coverage: the operator's XvB-off state gets an explicit regression test that no visible figure
is a range and no 8-dp negative renders; the warn verdict, the face net, `coinDp`'s magnitude rule
and the collapsed fiat line each get their own test.

## Not verified

There is no browser on this host, so the widths above are derived from the CSS and the rendered
strings, **not measured from a live layout**. The conclusion is robust across a wide range of
glyph-width assumptions (it fails to fit even at 6px/char), but a real render is worth a glance
before release — particularly the vertical height of four stacked blocks in a packed column, which
is the tradeoff this shape buys with.

Related: #1231 (the Odds column dashes while XvB is off). Not fixed here — there is no stable
published data to vendor, so it is honest as-is — but its cost drops sharply: it was a whole
column of a table that could not fit, and is now one item on a wrapping line.

---

## Correction: the band is hover-only, and that is a trade, not a free move

This PR says the spread stays "one hover away rather than gone". That is true for a mouse and false
for everything else. There is no `tabindex`, `role` or `aria-*` anywhere in these eight files, so a
`title` is unreachable by keyboard, absent on touch, and inconsistently announced by screen readers.
For those operators the range is **gone**, not moved.

The trade is deliberate — the widest cell in every row was a Net, and the ranges were what made the
table pan sideways — but it should be recorded as a trade. A reachable disclosure (a details row, or
the range under the midpoint at wider breakpoints) is the honest follow-up and is not in this PR.

## Reviewer finding B1, fixed at `19043f1`

The fiat verdict line collapsed `best.net` to a midpoint and carried no `title`, so the one line
stating the verdict in the operator's own currency lost its band while every other collapsed figure
kept one. Fixed by giving that line `title=${bandNote(best.net)}`, and covered by a test that
asserts the title carries the **range** rather than merely existing.

Both mutations red exactly that test: removing the title, and replacing it with a title that says
something else (which a laxer `has a title` assertion would have passed).

Also corrected a comment in the test file that read −1.6448 where the assertion and the arithmetic
say −1.6447.

Rebased twice: onto `a368ee4`, then onto the current tip after #1355 merged — that merge ratcheted
`test_xmrig_client.py`'s ceiling from 537 to 504, which made this branch's older budget file read as
a ceiling RAISE. Local gate at the pushed head: `node --test` 462 pass / 0 fail, `lint-js`,
`lint-file-budget`, `lint-operator-strings` all exit 0.


VijitSingh97:
Rebased onto the current `develop-v2` tip (`a368ee4`); head is now `7c8c180`. Force-pushed with `--force-with-lease`.

**One real conflict, and it was a moved-code collision rather than a disagreement.** In `components.mjs` this branch lifted `EstTable` out into the new `esttable.mjs`; while it sat, #1040 lifted `StatCard` out into `statcards.mjs` from the same region. Git saw one side delete `EstTable` and the other delete `StatCard` and could not tell they were compatible.

Resolved by dropping **both** inline definitions. Neither belongs in `components.mjs` any more, and both destinations are already imported at the top of the merged file. Checked before resolving that `develop-v2`'s inline `EstTable` is **byte-identical** to the copy this branch moved into `esttable.mjs`, so nothing of the newer base was discarded.

Everything else auto-merged.

## What was run after the rebase

- `node --test dashboard/tests/frontend/` — **461 pass, 0 fail**
- `make lint-js` (biome), `lint-topology`, `lint-md`, `lint-docs-voice` — each exit 0, checked separately rather than through a pipe
- `scripts/lint-operator-strings.sh`, `scripts/lint-file-budget.sh` — exit 0; the budget gate is happy that `components.mjs` shrank further under its 1354 ceiling

No Python file changed, so the pytest suite is untouched by the rebase; CI runs it anyway.

Issue #1059, opened 2026-08-16
pithead backup fails on the appliance: tar cannot stat config.json, after the stack has already been stopped


VijitSingh97:
Found by the KVM battery on the merged v2 tip (install phase, restore leg), on a keep-reinstalled machine that had just been provisioned and had a live stack running.

## What happened

```
[pithead] A consistent backup needs the services stopped, so the archive isn't captured mid-write.
[pithead] Stopping the stack for the backup (--yes)...
[pithead] Stack stopped.
[pithead] Creating backup archive...
tar: data/pithead/config.json: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
[WARNING] Backup attempt 1 failed — retrying once on a quiet tree...
tar: data/pithead/config.json: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
[pithead] Starting stack...
```

No archive was produced. The stack was stopped and restarted for nothing.

## Why this matters

Backup and restore are the documented recovery path for an appliance, and this is the ordinary case — a provisioned machine with a running stack, `backup -y`. It also fails **after** taking the stack down, so the cost of the failure is real downtime rather than a clean refusal. The disk-space pre-check exists specifically so that "a 'not enough space' answer leaves everything as it was"; this failure mode has no such protection.

## What is ruled out

- **Not a wrong `$PWD`.** `stack_backup` adds `"$PWD/$CONFIG_FILE"` and `"$PWD/$ENV_FILE"` together, unguarded, and strips the leading `/`. Only `config.json` failed to stat — `.env`, from the same `$PWD`, did not. A wrong working directory would have failed both.
- **Not a missing `/data/pithead` or a symlinked `/data`.** Checked on a live provisioned appliance guest: `/data` is a real directory, `readlink -f /data/pithead` is `/data/pithead`, and `/data/pithead/config.json` is present (`-rw------- root root 635`).
- **Not a momentary write window.** The `#970` retry-on-a-quiet-tree fired and the second attempt failed identically, so the file was absent for both passes rather than being replaced mid-walk.

## What to check next

[...]

PR #1358, opened 2026-08-24
test(os): capture what removes config.json during the restore leg's backup (#1059)

VijitSingh97:
## What this is

Evidence capture for #1059, in the appliance battery's restore leg. It found the root cause on
its first instrumented run.

## Why it existed to be written

#1059 is `tar: data/pithead/config.json: Cannot stat` on a live appliance, seconds after that
same file was captured as a present, root-owned regular file, with `.env` beside it archiving
fine. It had survived two static investigations and had been lost to a recycled guest twice: the
leg printed the backup log and nothing else, and the log cannot say whether the file was removed,
renamed, or by whom.

## What it does

`backup_precapture` records both collected files and starts a guest-side watcher for the length
of the backup. Within ~100ms of `config.json` going away the watcher records the `/data/pithead`
tree, the firstboot unit's state and journal, and the full process table — the acting process is
still likely to be alive at that point. The appliance image ships no `inotify-tools`, no `auditd`
and no `lsof`, so a poll plus a `/proc` sweep is what is available.

`backup_failure_evidence` additionally dumps the backup log in full, the tree, whether
provisioning was still in flight, the wizard spool and the firstboot journal.

**It reports on PASSING runs too**, and that turned out to be the whole point — see below.

## What it found

On its first instrumented run, on a machine where **the backup SUCCEEDED**:

```
01:25:28  containers Starting; tor still "Waiting"
01:25:56  Container tor  Error dependency tor failed to start
01:25:56  no container with name or ID "0494b001..." found: no such container
01:25:56  Error: executing docker-compose up --pull missing -d: exit status 1
01:25:57  config.json VANISHED -> config.json.failed
```

with `pithead firstboot-wizard`, `pithead backup -y` and `podman compose up -d` all alive at that
instant. `setup` was still inside `stack_up`; the concurrent backup's `stack_down` deleted a
container out from under its in-flight `compose up`; `stack_up` called `error()`; the `(setup)`
subshell exited non-zero; and the wizard loop moved the machine's configuration aside.

Every previous capture on this leg would have recorded that run as an uneventful pass. Reporting
on success is what caught it, and it means earlier "could not reproduce" greens on this leg are
not evidence the vanish did not happen — nothing was looking.

Root cause and the two product defects are written up on #1059. **This PR fixes none of them** —
it is the instrument, not the fix.

## Notes for review

- **`run.sh` stays under its recorded file-budget ceiling** (3437 vs 3442). The helpers live in
  the already-sourced `failure-evidence.sh` rather than in `run.sh`, which is the reusable trick
  for any instrumentation change to a budgeted file — the budget gate is otherwise the thing most
  likely to reject one.
- **Silence is never ambiguous.** Six outcomes, six sentences, and only one of them is quiet.
  An earlier revision of this PR claimed this property while only partly having it — see
  "Correction" below, which is the most important thing on this page for a reviewer.
- **A missing `config.json.failed` is not an alibi.** The branch is
  `mv -f ... 2>/dev/null || rm -f config.json`, so a failed `mv` deletes with no artifact. The
  helper says so, because an instrument that can exonerate the wrong line is worse than none.
- Failure-path and watcher output only; a healthy battery is otherwise unchanged.

## What was run

- `make lint` RC=0 (`lint-sh` completed with no OOM), `make test` **2903 passed, 0 failed** —
  matching the develop-v2 baseline.
- `tests/os/run.sh --phase install` on a KVM guest, twice. The first run lost its restore leg to
  my own mid-run edit (the harness rebuilds images from the live tree, so a dirty worktree makes
  `mkimage` refuse); the second ran on a frozen tree and is the one quoted above.
- Watcher logic exercised directly as well: vanish detected and timestamped, process table
  captured, reappearance seen. That exercise is now a committed `--self-test` rather than an
  ad-hoc one — see below.

## Correction: this instrument had the defect it exists to detect

Added in `1b7e87d`, after the reviewer lane asked — without having read the diff — whether
"nothing was removed" and "the watcher did not see it" could produce the same output. They could.

`backup_watch_report`'s `WATCHOK` sentinel proved the **guest answered**. Nothing proved the
**watcher was ever alive**. So three distinct ways of collecting no evidence all fell through to
the same silent `return 0` as a genuinely clean window:

| situation | what it printed before |
|---|---|
| the watcher never started (the launch is detached and its result discarded) | nothing |
| the watcher was killed mid-window | nothing |
| the watcher hit its own 1800s ceiling while the backup ran on | nothing |
| `config.json` genuinely stayed put | nothing |

A run that watched nothing was indistinguishable from a run that watched cleanly — which is
precisely the failure this file was written to catch, reappearing one level up inside the catcher.
Every "could not reproduce" this instrument might have produced would have carried that ambiguity.

**Now:** the watcher writes `STARTED` before it polls and `WINDOW EXPIRED` when its ceiling is
reached; the fetch also asks whether it is still running; `backup_precapture` confirms the launch
took, so a watcher that never started is named at the **start** of the window instead of inferred
from silence at the end. Each outcome gets a sentence only it writes.

### How that is proven, rather than asserted

- A `--self-test` drives all six outcomes. Each is asserted on its own sentence **and on the
  absence of every other outcome's sentence** — asserting on the shared
  "collected no evidence either way" tail would let any one branch silently cover for another's
  deletion, which is a trap this project has already been bitten by.
- **All six discrimination branches were mutated; every mutation goes red.** Deleting a branch
  makes its case fall through to the quiet path, and the test catches exactly that.
- The **watcher body is extracted from its own heredoc and run for real** — parsed, then driven to
  a genuine vanish and a genuine expiry against a sandbox file, with its actual output fed back
  through `backup_watch_report`. So the two halves are proven against each other rather than each
  against my idea of the other; a `STARTED:`/`STARTED at ` mismatch would be caught here and by
  nothing else. Paths and the ceiling read from the environment *only* so this can drive the
  shipped body unmodified rather than a respelled copy.
- One assertion in the first draft of that test was itself wrong, and the run caught it: the vanish
  report embeds a `/proc` process table, which is arbitrary text and can contain this file's own
  vocabulary. It did — a leftover log-follower from an earlier session was running with
  `WATCHER UNREADABLE` inside its grep pattern. The distinctness check now compares **header lines
  only**, since the header is what names the outcome and the body is captured evidence by nature.

### What this does NOT prove

The self-test is tier 1 against a stubbed transport. It does not prove that `pgrep` and the
detached launch behave on the appliance guest as they do on the bench host — that half is tier 4
and needs a battery run. The failure direction is the safe one: a guest-side failure of these new
lines degrades to a **loud wrong-reason** message, never to a false green. Stated rather than
glossed, because the whole point of the change is not overclaiming what was checked.

### Out-of-lane file, disclosed

**`tests/stack/test-harness-tooling.sh`** — owned outright by the `tests` lane, which is not
running. The controller granted the one-shot `.lane-override` for this rather than block a 1.7s
additive assertion behind a lane spawn and rehydrate. The commit is that one file and nothing else,
additive only, and the guard printed `override CONSUMED`. It is the house home for a `--self-test`:
five other scripts are driven there the same way, and tier 1 is the lowest tier that proves this
one — no network, no docker, no KVM.

## Re-run after rebase onto `8f384e3`

The branch was 5 commits behind and has been rebased. Worth recording because the budget gate went
RED on a file this PR does not touch: `lint-file-budget` reported that
`docs/dev/file-budget.tsv` "raises `dashboard/tests/client/test_xmrig_client.py`'s ceiling from
504 to 537". develop-v2 had *lowered* that ceiling, and the inherited older tsv read as a raise.
Rebasing cleared it with no change of mine — a ceiling going DOWN on the base makes every
un-rebased branch look like it is raising one.

Green on the new base, exit codes captured directly rather than read off filtered output:
`lint-file-budget` RC=0 · `shellcheck --severity=warning` RC=0 · `shfmt -i 4 -d` RC=0 ·
`--self-test` RC=0 (14 assertions, 0 failures). The full `make lint` / `make test` / battery
figures above were measured before the rebase.

## Ponytail pass

Run on this diff; two findings applied, two noted.

- **Applied:** the watcher teardown's `pkill -f` pattern matched the shell issuing it (the remote
  `bash -c` carries the pattern in its own cmdline) — bracketed.
- **Applied:** the `backup_precapture` header recited the investigation rather than its result.
  It now states the mechanism, since the root cause is established.
- **Noted, not changed:** on a vanish the failure path prints the `/data/pithead` tree twice, once
  from each helper. Deduping would lose the tree on failures with *no* vanish, which is the more
  expensive side of the trade.
- **Noted, not changed:** the reappear-detection loop covers a case never observed in the wild.
  Kept at 4 lines because it distinguishes a rename that is later restored from a permanent one.

---

## Correction to the figures above — two claims, two registers

Added post-merge at the controller's ask, because the two pre-rebase numbers quoted earlier are not
the same kind of claim and reading them as one overstates this PR's evidence.

**`make test` 2903/0 does NOT carry.** The reason is mechanical, not a labelling problem: the rebase
pulled in ten commits that ADD tests — a 431-line frontend suite, a 124-line size-cap suite, a
49-line config suite, and the tier-1 split. A count measured before the things it should have
counted cannot describe the tree after them. Treat it as void, not as stale.

**The battery's 41-pass-1-fail plausibly DOES carry, but it is an argument, not a measurement.**
Nothing the rebase pulled in touches `tests/os`, and the later commits touch only
`failure-evidence.sh`, which never calls `ok()`/`bad()` — the leg's tally comes from the
`backup -y` exit code alone. That is a mechanism, and it is offered as reasoning rather than as a
figure anyone re-measured. No battery has run on this branch.

**Still unproven, unchanged:** that `pgrep` and the detached launch behave on the appliance guest as
they do on the bench host is tier 4. The failure direction is safe — a loud wrong reason, never a
silent green — which is a mitigation and not evidence.

The distinction is owed to the reviewer lane, which split it rather than treating the two as one
claim.

### DIFF
--- a/tests/os/failure-evidence.sh
+++ b/tests/os/failure-evidence.sh
@@ -505,14 +505,22 @@ VANISHED at 2026-08-24T01:25:57+00:00"
     # The evidence dump is a /proc sweep: arbitrary cmdlines, which can and do contain this file's
     # own header vocabulary. A real vanish whose process table mentions a RIVAL outcome must still
     # report as exactly one outcome — the vanish — because the captured line is indented as evidence
-    # and only a printf writes a header. Un-anchor the header grep in _fe_case and this case fails
-    # while every other case stays green, which is the whole point of it being here.
+    # and only a printf writes a header.
+    #
+    # The canned cmdline below carries the header's FIVE-SPACE indentation on purpose, and that
+    # detail is the whole assertion. Without it the poison matched neither the anchored pattern nor
+    # an un-anchored one, so this case passed on the indentation baked into the grep pattern rather
+    # than on the anchor — green even with the ^ removed, which is the plausible edit. A case that
+    # proves the guard only against an implausible mutation is a guard proven by nothing, the exact
+    # shape this branch keeps finding. Caught by the reviewer lane, who removed only the caret.
+    # What makes the discrimination real: the marker sits INSIDE the line, while a genuine header
+    # STARTS it. Anchored, this cannot match; drop the ^ and it does.
     _fe_case "a vanish is not confused by its own evidence naming another outcome" \
         "config.json went away during the backup" \
         "WATCHOKALIVE:STARTED at 2026-08-24T01:00:00+00:00 pid=1234
 VANISHED at 2026-08-24T01:25:57+00:00
 -- process table at that instant --
-4711	grep --line-buffered -- --- #1059: WATCHER UNREADABLE — this run collected no evidence either way ---"
+4711	tail -f /tmp/pithead-1059-watch.log | grep -F      --- #1059: WATCHER UNREADABLE — this run collected no evidence either way ---"
 
     printf '== unit: #1059 watcher arming ==\n'
     # pgrep found it: a watcher is on the guest, and arming has nothing to report.
--- a/tests/os/failure-evidence.sh
+++ b/tests/os/failure-evidence.sh
@@ -82,9 +82,23 @@ backup_failure_evidence() {
 # first two static passes over the backup's own call chain found nothing, because the actor is a
 # concurrent process rather than anything that call chain invokes.
 backup_precapture() {
-    _ssh "ls -la /data/pithead/config.json /data/pithead/.env 2>&1; readlink -f /data/pithead/config.json 2>&1" |
-        tr -d '\r' | sed 's/^/     · /'
-    _ssh 'cat >/tmp/pithead-1059-watch.sh <<"EOS"
+    # Captured and re-emitted rather than streamed, so the header printed below is GUARANTEED to
+    # start its own line. sed does not terminate a final line that arrived unterminated, and the
+    # guest's last line here is a `readlink` whose termination is not ours to assume — glue the
+    # header onto the tail of an evidence line and every line-anchored reader of the battery
+    # scrollback, this file's own self-test included, stops seeing it. `$()` strips the trailing
+    # newlines, printf puts back exactly one, and empty output still prints nothing.
+    local pre
+    pre=$(_ssh "ls -la /data/pithead/config.json /data/pithead/.env 2>&1; readlink -f /data/pithead/config.json 2>&1" |
+        tr -d '\r' | sed 's/^/     · /')
+    [ -n "$pre" ] && printf '%s\n' "$pre"
+    # The log is cleared BEFORE the watcher is armed, not by the watcher. The watcher truncates it
+    # too, but only if it starts — and a watcher that never starts is precisely the case the report
+    # reads off an empty log. The restore leg runs on a keep-reinstalled guest, so a log left by an
+    # earlier run can outlive the run that wrote it; without this, "never started" would be reported
+    # against the PREVIOUS run's evidence. Loud and wrong is better than silent, but neither is the
+    # bar here.
+    _ssh 'rm -f /tmp/pithead-1059-watch.log; cat >/tmp/pithead-1059-watch.sh <<"EOS"
 # Paths and ceiling come from the environment so the self-test can drive this exact body
 # unmodified. The heredoc is quoted, so these expand on the GUEST at run time, where nothing sets
 # them and the defaults are what runs. A test that had to sed the constants would be proving a
@@ -225,7 +239,9 @@ backup_watch_report() {
 # silently covers for another's deletion. Delete any branch below and its case goes red rather
 # than falling through to the quiet path.
 _fe_selftest_reply=""
+_fe_selftest_pgrep=""
 _fe_selftest_rc=0
+_fe_ssh_capture=""
 
 # Every outcome's unique sentence. The keys are the assertion vocabulary: a case names the one it
 # wants and is checked against all the others.
@@ -233,6 +249,7 @@ _FE_SENTENCES="WATCHER UNREADABLE
 WATCHER NEVER STARTED
 WATCHER DIED mid-window
 WATCH WINDOW EXPIRED
+WATCHER DID NOT START
 config.json went away during the backup"
 
 _fe_case() { # <label> <want-sentence|QUIET> <canned _ssh reply>
@@ -264,7 +281,17 @@ _fe_case() { # <label> <want-sentence|QUIET> <canned _ssh reply>
     # earlier session was running with "WATCHER UNREADABLE" inside its grep pattern, the sweep
     # captured its cmdline, and a whole-output match called that a second outcome. The header is
     # what NAMES the outcome, so the header is what has to be unambiguous.
-    headers=$(printf '%s\n' "$got" | grep -- '--- #1059:' || true)
+    # ANCHORED, and that anchor is load-bearing. Every header this file writes is a printf that
+    # starts the line at this exact indentation; every line of captured evidence is prefixed
+    # '     | ' (vanish) or '     · ' (precapture) by the sed that renders it. So anchoring makes
+    # the two unconfusable. Unanchored, this grep matched the marker WHEREVER it appeared —
+    # including inside the /proc sweep the vanish branch dumps, whose cmdlines are arbitrary text.
+    # That is not hypothetical twice over: a predecessor session's orphaned `tail -f | grep` held
+    # 'WATCHER UNREADABLE' in its pattern and produced a false failure here, and this very branch's
+    # mutation runs poisoned themselves the same way — the sed rewriting a header string was itself
+    # captured as evidence naming a second outcome. A searcher that can find its own vocabulary in
+    # what it is searching reports confident nonsense.
+    headers=$(printf '%s\n' "$got" | grep -- '^     --- #1059:' || true)
     while IFS= read -r line; do
         [ -n "$line" ] || continue
         [ "$line" = "$want" ] && continue
@@ -276,6 +303,67 @@ _fe_case() { # <label> <want-sentence|QUIET> <canned _ssh reply>
         esac
     done <<<"$_FE_SENTENCES"
     printf 'ok: %s\n' "$label"
+}
+
+# The ARM side. Everything above drives the REPORT; until now nothing drove the one call that puts
+# a watcher on the guest in the first place, and that asymmetry was not theoretical: deleting
+# backup_precapture's confirmation OUTRIGHT left this suite fully green (mutated 2026-08-24, 14/14
+# ok, rc 0). A guard present in the source and unproven by any test is the same defect this file
+# exists to refuse, one call earlier — the reviewer lane made exactly that structural point against
+# an earlier revision of this branch.
+#
+# The launch is `setsid nohup ... & disown` inside `>/dev/null 2>&1 || true`, so its exit status
+# carries no information about whether a watcher is running. Only the pgrep afterwards does, and
+# these two cases are what hold it in place.
+_fe_precapture_case() { # <label> <want-sentence|QUIET> <canned evidence reply> <canned pgrep reply>
+    local label="$1" want="$2" got
+    _fe_selftest_reply="$3"
+    _fe_selftest_pgrep="$4"
+    # Headers only. The first _ssh in precapture echoes the guest's `ls` back through a `· ` prefix,
+    # and under a single canned stub that reply is this case's own sentinel text — matching the whole
+    # output would let the echo satisfy the assertion instead of the branch under test.
+    got=$(backup_precapture | grep -- '^     --- #1059:' || true)
+    if [ "$want" = "QUIET" ]; then
+        if [ -n "$got" ]; then
+            printf 'FAIL: %s — a watcher confirmed running must name no outcome, got: %s\n' "$label" "$got"
+            _fe_selftest_rc=1
+            return
+        fi
+        printf 'ok: %s (silent)\n' "$label"
+        return
+    fi
+    case "$got" in
+    *"$want"*) printf 'ok: %s\n' "$label" ;;
+    *)
+        printf 'FAIL: %s — expected "%s", got: %s\n' "$label" "$want" "${got:-<silence>}"
+        _fe_selftest_rc=1
+        ;;
+    esac
+}
+
+# A stale log outliving the run that wrote it is the one way an empty-log reading can be right about
+# "never started" and still be reported against the previous run's evidence. The clear has to happen
+# before the script is written, not after the spawn, or it races the watcher's own first line.
+#
+# Asserted on the command precapture actually SENDS, which is a text-level check: it dies if the
+# clear is deleted or reordered, and it cannot prove the guest honoured it. That proof is tier 4.
+_fe_precapture_arms_clean() {
+    local sent
+    _fe_selftest_reply=""
+    _fe_selftest_pgrep="UP"
+    _fe_ssh_capture=""
+    backup_precapture >/dev/null
+    sent=$_fe_ssh_capture
+    case "$sent" in
+    *"rm -f /tmp/pithead-1059-watch.log"*"cat >/tmp/pithead-1059-watch.sh"*)
+        printf 'ok: the arm call clears any stale log before writing the watcher\n'
+        ;;
+    *)
+        printf 'FAIL: the arm call does not clear the stale log ahead of the watcher: %s\n' "${sent:0:120}"
+        _fe_selftest_rc=1
+        ;;
+    esac
+    _fe_ssh_capture=""
 }
 
 # The other half of the contract. Everything above drives the REPORT against a canned reply; it
@@ -372,8 +460,20 @@ _fe_watcher_cases() {
 
 _fe_self_test() {
     # The stub transport. backup_watch_report calls _ssh twice — the fetch and the pkill — and
-    # discards the second's output, so one canned reply serves both.
-    _ssh() { printf '%s' "$_fe_selftest_reply"; }
+    # discards the second's output, so one canned reply serves both. It also records what it was
+    # asked to run, which is the only handle tier 1 has on the arm call: that one is fire-and-forget
+    # by design, so nothing about it can be read back from a reply.
+    _ssh() {
+        _fe_ssh_capture="$_fe_ssh_capture$*"
+        # backup_precapture makes two calls with different jobs, and a case needs to drive them
+        # independently: the evidence read, and the arm-check probe. Matched on the probe's own
+        # tail rather than on 'pgrep', because backup_watch_report's fetch contains a pgrep too and
+        # matching that would quietly re-point every report case at the wrong canned reply.
+        case "$*" in
+        *"&& printf UP") printf '%s' "$_fe_selftest_pgrep" ;;
+        *) printf '%s' "$_fe_selftest_reply" ;;
+        esac
+    }
 
     printf '== unit: #1059 watch-report discrimination ==\n'
     # No sentinel at all: ssh failed, the guest is gone, the command died. Nothing is known.
@@ -402,6 +502,33 @@ VANISHED at 2026-08-24T01:25:57+00:00"
     # The ONLY quiet case: started, still watching, nothing seen.
     _fe_case "a fully-watched window with no vanish stays quiet" "QUIET" \
         "WATCHOKALIVE:STARTED at 2026-08-24T01:00:00+00:00 pid=1234"
+    # The evidence dump is a /proc sweep: arbitrary cmdlines, which can and do contain this file's
+    # own header vocabulary. A real vanish whose process table mentions a RIVAL outcome must still
+    # report as exactly one outcome — the vanish — because the captured line is indented as evidence
+    # and only a printf writes a header. Un-anchor the header grep in _fe_case and this case fails
+    # while every other case stays green, which is the whole point of it being here.
+    _fe_case "a vanish is not confused by its own evidence naming another outcome" \
+        "config.json went away during the backup" \
+        "WATCHOKALIVE:STARTED at 2026-08-24T01:00:00+00:00 pid=1234
+VANISHED at 2026-08-24T01:25:57+00:00
+-- process table at that instant --
+4711	grep --line-buffered -- --- #1059: WATCHER UNREADABLE — this run collected no evidence either way ---"
+
+    printf '== unit: #1059 watcher arming ==\n'
+    # pgrep found it: a watcher is on the guest, and arming has nothing to report.
+    _fe_precapture_case "a watcher confirmed running arms quietly" "QUIET" "" "UP"
+    # pgrep found nothing: the detached launch did not take. Said at the START of the window, where
+    # it still reads as a warning — the report will say it again at the end, and the two are
+    # deliberately different sentences so neither can stand in for the other.
+    _fe_precapture_case "a watcher that failed to arm is named immediately" "WATCHER DID NOT START" "" ""
+    # The guest's evidence arriving WITHOUT a trailing newline is the case that broke this: sed does
+    # not terminate an unterminated final line, so the header printed next was glued to the end of an
+    # evidence line and stopped being a header at all. Anchored readers — the battery scrollback, and
+    # the assertion above — then see nothing while the branch is firing correctly. Found by mutation:
+    # with the guard removed, inverting the arm-check no longer failed anything.
+    _fe_precapture_case "the header starts its own line even when the guest evidence does not" \
+        "WATCHER DID NOT START" "/data/pithead/config.json" ""
+    _fe_precapture_arms_clean
 
     printf '== unit: #1059 watcher body, run for real against a sandbox file ==\n'
     _fe_watcher_cases
--- a/tests/os/failure-evidence.sh
+++ b/tests/os/failure-evidence.sh
@@ -115,12 +115,25 @@ EOS
 setsid nohup bash /tmp/pithead-1059-watch.sh </dev/null >/dev/null 2>&1 & disown' >/dev/null 2>&1 || true
 }
 
-# Report whatever the watcher saw, and stop it. Silent when the file never went away, so a healthy
+# Report whatever the watcher saw, and stop it. Quiet when the file never went away, so a healthy
 # run stays quiet; loud the moment it blinks, pass or fail.
+#
+# The one thing it must NOT do is stay quiet because it could not look. A silent "no vanish" and a
+# silent "the guest never answered" are the same output and opposite meanings, and reading the
+# second as the first is how a run gets recorded as evidence when it collected none. So the log is
+# fetched with a sentinel: no sentinel back means the watcher could not be read, and that is said
+# out loud rather than passed off as a clean result.
 backup_watch_report() {
-    local seen
-    seen=$(_ssh "cat /tmp/pithead-1059-watch.log 2>/dev/null" | tr -d '\r')
+    local raw seen
+    raw=$(_ssh "printf WATCHOK; cat /tmp/pithead-1059-watch.log 2>/dev/null" | tr -d '\r')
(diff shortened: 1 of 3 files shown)


VijitSingh97:
## Review: one HIGH finding, and a list of checks that did NOT run

Reviewed at head `d3b52ac8` by mutation, not by reading. I built a scratch harness that sources this
PR's actual `tests/os/failure-evidence.sh` with a stubbed `_ssh` that executes the same command
strings locally, with the guest paths remapped into a scratch dir so nothing touched a real host.
Everything under "Confirmed" below was executed and observed; everything under "Not run" was not,
and I have deliberately left it open rather than implying coverage I do not have.

### Confirmed by execution

| Scenario | Observed |
|---|---|
| Healthy: no vanish, watcher armed, fetch OK | silent, exit 0 — baseline |
| Real vanish, watcher armed, fetch OK | full evidence block: `config.json went away...`, tree, journal, proc table |
| Fetch `_ssh` dead after a real vanish (the mutation the PR body asks for) | `--- WATCHER UNREADABLE — this run collected no evidence either way ---` |
| Same, with the `WATCHOK` sentinel reverted to its pre-`97c57de` shape | **silent, exit 0** — indistinguishable from healthy |

So the sentinel earns its place: the read side genuinely distinguishes "nothing vanished" from
"I could not read the result," and the last row confirms the historical hole `97c57de` closed is a
real hole and is really closed.

`detect-topology.py` on the PR body: clean. File-budget claim: re-counted `tests/os/run.sh` at head,
3437 lines, matches the stated ceiling exactly. The pre-existing battery failure attributed to the
open `restore_live_state_verdict` issue is real, still open, and is a distinct assertion in the same
leg — the attribution holds.

### HIGH — the arm side is unguarded, and it fails silent-clean

The PR body's claim is that silence is never ambiguous. That is true for the **read** side only. It
is not true for the **arm** side, and I reproduced this against the shipped code, not as a
hypothetical:

`backup_precapture` spawns the watcher as a single `_ssh` invocation — heredoc write, `setsid`,
`nohup`, `bash`, `disown` — wrapped in `>/dev/null 2>&1 || true`. Nothing afterwards confirms the
watcher is alive. Mutation: make that spawn fail while the fetch `_ssh` stays perfectly healthy,
then let a real vanish occur.

Result: **silent, exit 0, `watch.log` confirmed absent on the guest.** Byte-for-byte
indistinguishable from a clean run.

That is the same defect class this PR exists to kill — a blind run reading as a clean one — moved
one call earlier, to the one call the PR did not fence. It matters here more than it would
elsewhere: this leg is specifically about a race under concurrent load, so an SSH hiccup during
precapture is not an exotic input, it is the same class of event the leg is instrumenting.

**And it cannot ever surface as a red count in CI.** `ok()`/`bad()` in `run.sh` bump PASS/FAIL from
the `backup -y` exit code alone; `backup_precapture`, `backup_watch_report`, and
`backup_failure_evidence` never call either — they only `printf`. So a run that armed nothing
produces a green suite plus a missing line in scrollback that nobody is required to read. A reader
who sees the suite pass will reasonably assume this is covered. It is not.

Suggested fix, at the layer already used for the read side: have `backup_precapture` synchronously
write an arm marker before returning, and give `backup_watch_report` a third branch — "armed and
quiet" / "never armed" / "unreadable" — so the three states stop collapsing into two.

Lower confidence, flagged not asserted: nothing clears `watch.log` before arming, so if a guest or
disk were ever reused across two invocations of this leg, a stale log would be reported as
current-run evidence. The PR's narrative implies a fresh image per run, which would make this moot;
I did not confirm guest-reuse semantics from the diff alone.

### NOT run — the green claims in the body are uncorroborated

Stated plainly so nothing above is read as covering it:

- `make lint`, `make test` (the body's 2903/0), and `shellcheck` on either changed file: **not run.**
  `tests/os/run.sh` is 3437 lines, close enough in size to the suite file that OOM-kills a session
  on this runner that I did not risk it, and `lint-sh` is serialized fleet-wide for exactly that
  reason. This is a gap I am reporting, not one I closed.
- The KVM battery result (41 passed / 1 failed): **not re-run.** I verified the *identity* of the
  claimed pre-existing failure instead, which is the strongest check available without hardware.
- Whether a genuine SSH blip during precapture is *likely* on real hardware, as opposed to
  demonstrated possible here: **unverified**, and it would need the bench.

Everything in "Confirmed" was observed. Everything in this section was not attempted. Please do not
read the first as evidence for the second.

### Recommendation

Do not merge as-is. Add the arm marker and the third branch, then merge — the rest of the change is
sound and the instrument does what it says on the side it fences. Routing the fix to the lane that
owns `tests/os/`; I own no code and do not commit.

VijitSingh97:
Acted on in `b3ee0ec` (re-derive with `git ls-remote`, do not trust that sha). The review read a revision one commit before the arm side existed, so the literal mutation described is already dead at the tip — but the structural point behind it was right, and chasing it turned up two further defects I had not seen.

**The literal finding, at the tip.** `backup_precapture` already ran a `pgrep` after the detached launch and printed `WATCHER DID NOT START`; `backup_watch_report` already read an empty log as `WATCHER NEVER STARTED`. Break the spawn, keep the fetch healthy, let a real vanish happen: two loud lines, not silence. The PR body did not make that revision boundary findable, which is on me.

**The structural point, which held.** Deleting `backup_precapture`'s confirmation outright left the self-test fully green — 14/14 ok, rc 0. A guard present in the source and proven by nothing, which is the shape this file exists to refuse, one call earlier. Found by mutating, not by reading.

Four changes, each held in place by a mutation that kills it and nothing else. Seven mutations, each applied in isolation to an otherwise clean tree, each dying on its intended assertion:

1. **The arm-side check is now driven.** Two cases plus a per-call stub. The stub discriminates on the arm probe's own tail rather than on `pgrep`, because `backup_watch_report`'s fetch contains a `pgrep` too — matching that would have quietly re-pointed every report case at the wrong canned reply.

2. **The stale log is cleared before the watcher is armed.** The review flagged this at lower confidence and I could not rule it out: the restore leg's source machine is a keep-reinstalled guest, so a log can outlive the run that wrote it and be reported against the wrong one. Rather than argue the instance, the class is closed — the arm call clears the log before writing the watcher. Asserted on the command actually sent, so it dies if the clear is deleted *or* reordered after the spawn. That the guest honours it is tier 4 and is labelled as such, not claimed.

3. **The self-test's header grep was unanchored.** The vanish branch dumps a `/proc` sweep — arbitrary cmdlines — and the grep matched its own marker wherever it appeared, including inside that evidence. It poisoned itself during these very mutation runs: the `sed` rewriting a header string was captured as evidence naming a second outcome, and the result looked like flakiness. Same shape as the orphaned log-follower that produced a false failure here previously. Anchored now, with a deliberate poisoned-evidence case; un-anchoring kills exactly that case.

4. **Anchoring then produced a vacuous pass, and fixing that fixed a real formatting bug.** GNU `sed` does not terminate a final line that arrived unterminated, so the header printed after the precapture evidence was glued onto the end of an evidence line — no longer at line start, invisible to every line-anchored reader including the battery scrollback. The quiet case passed for the wrong reason, and inverting the arm-check stopped failing anything. Fixed in the shipped code (capture, then print exactly one newline), with a case driving unterminated guest evidence.

**Run, with each exit code captured directly rather than read off a filtered pipe:** `bash tests/os/failure-evidence.sh --self-test` — 19 assertions, rc 0, ~2s, no KVM. `shellcheck --severity=warning` rc 0. `shfmt -i 4 -d` rc 0. `bash -n` rc 0. `lint-file-budget` rc 0, re-run after rebasing. `lint-topology` rc 0. `lint-operator-strings` rc 0. The tier-1 registration was driven standalone and mutated through, so a regression in the arm check reddens the suite rather than only a hand-run. Rebased onto the integration branch, 0 behind.

**Not corroborated, and not claimed.** The `make test` total and the battery pass/fail figures quoted earlier in this PR were measured before an earlier rebase and have not been re-run. `make lint-sh` is serialized fleet-wide for OOM reasons and the lock was not taken. No battery was run: that `pgrep` and the detached launch behave on the appliance guest as they do on the bench host remains tier 4 and unproven. The failure direction stays safe — a loud wrong reason, never a silent green.

VijitSingh97:
## Correction to my review above — and one new finding at the tip

**My earlier comment was against a stale revision, and that is my defect, not the author's.**
It reads at `d3b52ac8`, which is the pre-rebase twin of `398680a` — one commit *before* the arm
side landed. The timing, re-derived from the branch: `d3b52ac8` authored 22:05, the arm-side commit
`73c8c79` authored 22:22:33, my comment posted 22:30:51. The review itself read the then-current
head honestly; I posted eight minutes after the head had moved without re-checking it. Anyone
reading the HIGH above should treat it as **closed at the tip**.

At `b3ee0ec` (confirmed via `git ls-remote`, not from the message that told me) my described
scenario — break the spawn, keep the fetch healthy, real vanish — prints two loud lines rather than
silence. `backup_precapture` pgreps after the detached launch and names `WATCHER DID NOT START`;
`backup_watch_report` reads an empty log as `WATCHER NEVER STARTED`. Verified by running it.

### Confirmed at the tip by mutation, each applied in isolation

Self-test baseline: 19 assertions, rc 0 — reproduces the stated figure.

| Mutation | Result |
|---|---|
| Revert the precapture to the streamed shape (undo the newline fix) | **FAIL** — "the header starts its own line even when the guest evidence does not" |
| Delete the stale-log clear from the arm call | **FAIL** — "the arm call does not clear the stale log ahead of the watcher" |
| Delete the arm confirmation outright | **FAIL** ×2 — both arming cases go silent |

Each dies on its intended assertion and the suite goes rc 1. The premise under the newline fix is
also real, re-derived directly: GNU `sed` leaves an unterminated final line unterminated, and the
streamed shape genuinely produces `...no-newlineHEADER` with the header no longer at line start.
The capture-then-`printf` form emits the newline. That one was a true vacuous pass and it is now
closed.

### NEW — MEDIUM: the anchor is right, but the case that claims to prove it does not

The header grep was anchored, and the source comment calls that anchor load-bearing. **The anchor
is load-bearing — but the deliberate poisoned-evidence case does not hold it.**

Remove the `^` from both header greps and the suite stays fully green: rc 0, 19/19, five runs in a
clean environment. The canned poison is a cmdline reading `... grep --line-buffered -- --- #1059:
WATCHER UNREADABLE ...`. Rendered, it carries the evidence prefix, and the grep pattern itself
begins with five spaces — so the poison never had five spaces before the marker and cannot match
with or without the anchor. What actually saves that case is the indentation baked into the
pattern, not the anchor. Same defect class as the newline fix: a guard that is real in the source
and proven by nothing.

That the anchor matters is not theory. Running the un-anchored and anchored variants back to back
under a process table that happened to contain the marker with its indentation: un-anchored rc 1
("a real vanish log reports as a vanish — its header also names another outcome"), anchored rc 0,
same conditions, same run. It reproduced accidentally first — my own shell supplied the poison
during an unrelated step — which is precisely the self-poisoning described in that comment.

**Fix, proven end to end here:** give the canned cmdline the five-space indentation a real header
carries, e.g. `... tail -f watch.log | grep --      --- #1059: WATCHER UNREADABLE ...`. With the
anchor intact: rc 0. With the anchor removed: rc 1, failing on the poisoned case by name. One
string, and the anchor becomes load-bearing in the suite as well as in the source.

### Still not corroborated — unchanged, and not mine to close

`make test` (the 2903/0 figure), `make lint-sh`, and the KVM battery's 41-pass/1-fail were measured
before the rebase and have not been re-run; the author says so directly and I am repeating it here
so it does not get lost behind the mutation table. `pgrep` and the detached launch against a real
appliance guest remain tier 4, unproven by anything above — everything I ran was tier 1, on a
stubbed transport, with no guest. I did not run `make lint-sh`: it is fleet-serialized because
shellcheck OOMs sessions on large suite files, and I did not take the lock.

Verified: sound apart from the MEDIUM. The arm-side asymmetry my first comment aimed at is
genuinely closed, and closed with tests that die when the guard is removed.

VijitSingh97:
The MEDIUM on the anchor case is confirmed and fixed in `682ce66` (re-derive with `git ls-remote`).

The two measurements looked contradictory and both were correct — we mutated different things. Removing the caret *and* the five spaces leaves `--- #1059:`, which the canned poison does contain, so it went red and I read that red as proof the anchor was held. Removing the caret *alone* leaves `     --- #1059:`, which the poison did not match, so it stayed green: rc 0, 19/19, reproduced across three consecutive runs here.

The case was therefore carried by the indentation baked into the grep pattern, not by the `^` — and caret-only is the plausible edit. A guard real in the source and proven by nothing, inside the case written to refuse exactly that. The sharper half of the mistake is mine: I picked the mutation that made my own guard look proven and stopped there. A red confirms an expectation as cheaply as a green does, and when several things changed at once it does not say which one did the killing.

Fix is one string in the canned input, no production change: the cmdline now carries the header's five-space indentation *inside* the line. That is what makes it discriminate — a genuine header starts its line, this one does not.

Verified across four quadrants, each in isolation on an otherwise clean tree, each failure naming the poisoned case and nothing else:

| mutation | rc |
|---|---|
| anchor intact | 0 — three consecutive runs, 19/19 |
| caret removed from both greps | 1 |
| caret and indentation removed | 1 |
| caret removed from one grep only | 1 |

The third row is the one that was green before; the fourth confirms a single-site edit is caught too.

CI was fully green at the prior tip `b3ee0ec` — 15 checks pass, 0 fail, including **Shell tests (shellcheck + pithead suite)**, the job that drives this self-test through its tier-1 registration. CI has not yet reported on `682ce66` and is not being claimed until it does.

Re-run locally on the new tip, each exit code captured directly: `shellcheck --severity=warning` 0, `shfmt -i 4 -d` 0, `bash -n` 0, `lint-file-budget` 0, `lint-topology` 0, `lint-operator-strings` 0.

Unchanged and still open: the `make test` total and battery figures quoted earlier predate the rebase and have not been re-run; `make lint-sh` is fleet-serialized and the lock was not taken; no battery has run, so `pgrep` and the detached launch on the appliance guest remain tier 4.

VijitSingh97:
## Re-reviewed at `682ce66` — MEDIUM closed, one residual LOW, not merge-blocking

Tip re-derived with `git ls-remote`: `682ce66`, MERGEABLE. All four quadrants re-run independently
on my own copy, each mutation applied programmatically to an unmodified checkout of the file at that
sha, each in isolation:

| Quadrant | rc | Failing case |
|---|---|---|
| Both anchors intact | **0**, three consecutive runs | — |
| Caret removed at both greps | **1** | the poisoned-evidence case, and nothing else |
| Caret and indentation removed at both | **1** | same, and nothing else |
| Caret removed at `_fe_case` alone | **1** | same, and nothing else |

That reproduces the author's figures exactly. The MEDIUM is closed: the caret-only edit — the
plausible one, and the one that was invisible before — now reddens the suite, and the case fails by
name rather than dragging unrelated assertions down with it.

### Residual, LOW: the second anchor is still proven by nothing

There are two anchored header greps. The new poisoned case holds the one in `_fe_case`. The one in
`_fe_precapture_case` is unheld: remove its caret **alone**, leave the other intact, and the suite is
green — rc 0. The reason is structural rather than an oversight in the new case: the quiet arming
case passes an empty canned evidence reply, so there is no evidence for an un-anchored grep to
mistake for a header. Nothing can poison an empty string.

The exposure is genuinely smaller than the one just fixed, and I would not hold the merge for it.
The report side renders a `/proc` sweep — arbitrary cmdlines, and the branch has now been bitten by
that twice for real. The precapture side renders `ls -la` and `readlink -f`, where a line carrying
an indented marker is contrived rather than observed. So: same defect class, much lower reachability.

**Fix, proven end to end here.** Give the quiet arming case a canned evidence reply whose `readlink`
target embeds the marker inside the line, mirroring what the report side already does:

- both anchors intact → rc 0, three consecutive runs
- that fix plus the caret removed at `_fe_precapture_case` alone → rc 1, failing `a watcher
  confirmed running arms quietly — a watcher confirmed running must name no outcome`
- the tip as shipped under the same single-site edit → rc 0, which is the gap

One string in the canned input again, no production change. Take it in this branch or as a
follow-up; the production guard at both sites is correct either way, and what is missing is only the
proof that the second anchor is load-bearing.

### On the CI claim

`b3ee0ec` checks out: 15 success, 2 skipped, 0 failure — I queried the check-runs for that sha
directly rather than reading the PR's rollup. At `682ce66`, CI has **not** finished: 12 success, 2
skipped, and **3 still pending**, including `Shell tests (shellcheck + pithead suite)` — the job that
drives this self-test through its tier-1 registration, so it is the one that matters here. The
author already said they would not call it green until it reports, which is the right call; I am
recording the pending state so nobody reads the b3ee0ec result as covering the new tip.

### Unchanged, and still open

`make test` 2903/0 and the battery's 41-pass/1-fail predate the rebase and have not been re-run.
`make lint-sh` lock not taken by either of us. `pgrep` and the detached launch against a real
appliance guest remain tier 4, unproven by anything in this thread — everything I ran is tier 1 on a
stubbed transport with no guest.

Recommend merge once the pending checks report green, with the LOW taken here or tracked.

VijitSingh97:
## Verdict at `e0b1f37`: merge-ready. Three questions answered, one of them not the way the label suggests.

Both anchors now hold. Re-ran at the new tip, mutations applied programmatically to a clean
checkout: both intact rc 0 (three runs); caret off `_fe_case` alone rc 1, failing the poisoned
vanish case; caret off `_fe_precapture_case` alone rc 1, failing the quiet arming case. Each
single-site edit fails its own case and nothing else — which neither anchor had two commits ago.

### 1. Do the pre-rebase figures still carry? Split the answer — they are not one claim.

**`make test` 2903/0 does not carry, and the reason is mechanical.** The rebase moved the base from
`a368ee4` to `4f0df21`, pulling in ten commits that *add tests*: a 431-line frontend suite, a
124-line size-cap suite, a 49-line config suite, and the tier-1 domain split. A total measured
before those landed cannot describe the tree after them. Nothing is wrong with the number; it is
simply describing a different tree, and no label can fix that.

**The battery's 41-pass/1-fail plausibly does carry, and here is the mechanism rather than a
shrug.** Nothing the rebase pulled in touches `tests/os/` at all — the ten commits land in
`dashboard/`, `tests/stack/`, `scripts/`, and `ci.yml`. The branch's own post-measurement commits
touch only `failure-evidence.sh`, and that file never calls `ok()`/`bad()`; the leg's tally comes
from the `backup -y` exit code alone. So there is no path by which either change moves the count.
That is an argument, not a measurement, and it should be read as one.

The stronger evidence is not the number anyway: CI at this tip is green — 15 pass, 2 skipped, 0
failure — including `Shell tests (shellcheck + pithead suite)`, which is the job that drives this
file's self-test through its tier-1 registration. A settled green there covers more of what the
stale figure was standing in for than the figure did.

### 2. The `.lane-override`: nothing rode along.

The whole PR touches three files: `tests/os/failure-evidence.sh`, `tests/os/run.sh`, and
`tests/stack/test-harness-tooling.sh`. Only the third is out of lane, it appears in exactly one
commit, and it is twelve lines appended at end-of-file registering the self-test — no edits to
existing lines, nothing unrelated. The hatch allowed every out-of-lane file in that commit; that
commit contained one, and it is the intended one.

Worth one flag to whoever routes it: that same file gained eight lines from `277063c` on the base
branch. Both additions are append-only at the tail and the rebase resolved clean, but two lanes are
now appending to one file.

### 3. Can this instrument tick green while the thing it watches for happened? Yes — and that is the point, with one caveat.

It is not a failure-only instrument, which is the property that matters. `run.sh` calls
`backup_precapture` before the backup and `backup_watch_report` on the **success** branch, with the
comment `a vanish the backup happened to survive is still the #1059 event`. So it fires on exactly
the run shape the original capture came from — a backup that succeeded.

And yes, such a run still ticks green: the three evidence functions only `printf`, and the leg's
pass/fail comes from the `backup -y` exit code alone. A vanish that the backup survives prints a
full evidence block into scrollback while the leg passes.

That is correct for an instrument and the PR says so plainly. The residual is that **nothing forces
anyone to read it** — the capture lands in scrollback on a green run, and a green run is the one
nobody reads. If the intent is that a vanish must never pass unnoticed, that is a gating decision
for the follow-up, not a defect in this PR. Flagging it as a decision, not asking for a change here.

### Not corroborated, unchanged

`make lint-sh` lock untaken by anyone in this thread. The battery has not been re-run. `pgrep` and
the detached launch against a real appliance guest remain tier 4 — everything verified above is tier
1 on a stubbed transport with no guest.

### The skipped CVE sweep

An event filter doing its job, not a gate quietly not running: `sweep-shipped` carries
`if: github.event_name == 'schedule'`, so skipping on a pull request is the only correct outcome.
The `schedule:` block in this lane's `ci.yml` is inert by design and the file says so at lines
20-23 — cron is read from the default branch only, so both sweeps run from `develop`'s copy.

One thing that is *not* covered by that, and is worth someone's attention: `sweep-shipped` landed on
`develop` at 08-23 21:27, and the most recent scheduled run was 08-17. Across all 13 scheduled runs
in history the branch is `develop` every time and this job appears in none of them. **It has never
fired.** The next window is 05:00 UTC today. Prove the first firing from
`gh run list --event schedule` rather than from the YAML — a sweep that has never run is exactly the
shape this project keeps filing.

Recommend merge. Not merging myself.

VijitSingh97:
Merged as `dea75e3` (parents `4685945`, `e0b1f37`). Re-checked the head with `git ls-remote` immediately before merging and confirmed it matched the sha the review verdict names — MERGEABLE/CLEAN, every check SUCCESS except the two CVE-sweep jobs, which are correctly skipped because they carry `if: github.event_name == 'schedule'`.

**Correcting the evidence claims in the body, because they do not all carry and the record should not imply they do.** The lane ran out of runway before it could amend this, so recording it here.

- **`make test` 2903/0 does NOT carry.** The rebase moved the base and pulled in ten commits that ADD tests — a 431-line frontend suite, a 124-line size-cap suite, a 49-line config suite, and the tier-1 domain split. A count measured before the things it should have counted cannot describe the tree afterwards, and labelling it "pre-rebase" does not repair that.
- **The battery's 41-pass-1-fail plausibly does carry, as an argument rather than a measurement.** Nothing the rebase pulled in touches `tests/os` at all, and the branch's later commits touch only `failure-evidence.sh`, which never calls `ok()`/`bad()` — the leg's tally comes from the `backup -y` exit code alone. That is a mechanism, not a re-run. **No battery has run against this tree.**
- **Not proven, tier 4:** that the `pgrep` and the detached launch behave the same on the appliance guest as on the bench host. The failure direction is safe — a loud wrong reason, never a false green — which is a mitigation and not evidence.

**What released the merge** was the CI green at the tip plus two independent adversarial passes, not the figures above. Four defects were fixed during review, three found by the reviewer lane and one by the author's own mutation, and each fix exposed the next: an unguarded arm side; an unanchored header grep that had poisoned itself; a vacuous pass that in turn exposed a genuine line-termination bug in the shipped code; and finally two newly-added anchor-proving cases that were themselves proven by nothing — one carried by pattern indentation rather than by the anchor, the other by an empty canned input. Each closed with a mutation that kills it and nothing else.

**The CI red seen on this PR earlier was a FALSE red and is now filed as #1370.** `expect_present` pipes the whole rendered compose config into grep on every check; a short write leaves grep reading truncated input and a published port reads as missing. A truncated read and a deleted directive are the same observable. The worse half is that `expect_absent` shares the shape and fails in the opposite direction, which would turn a real hardening regression into a green tick.

**#1059 stays OPEN.** This PR captures evidence for it; it does not fix it. The instrument is deliberately not failure-only — `run.sh` calls `backup_watch_report` on the SUCCESS branch, because a vanish the backup happened to survive is still the event. The residual, raised as a decision rather than a defect: nothing forces anyone to read scrollback on a green run, and a green run is the one nobody reads. If a vanish must never pass unnoticed, that is a gating call for the follow-up.

Issue #1357, opened 2026-08-24
The inventory gate sees a module file that vanishes, not a source line that vanishes


VijitSingh97:
## What is missing

`tests/stack/run.sh` runs under `set -uo pipefail` with **no `-e`**, and its tail exits on the
failure counter alone (`tests/stack/run.sh:4891-4895` at `a368ee4`):

```
printf 'pithead tests: %d passed, ' "$PASS"
if [ "$FAIL" -gt 0 ]; then ... exit 1; fi
printf '0 failed\n'
```

`$PASS` is printed and never compared against anything. So any way of losing a whole domain of
assertions produces a smaller number in green text and an exit code of 0.

`tests/inventory.sh` carries the two gates that stand between that and a silent loss:

- **Source-target existence** (`tests/inventory.sh:108-116`) — every file `run.sh` claims to
  source must exist. Catches a module file deleted or renamed out from under a live `source` line.
- **Per-file section count** (`tests/inventory.sh:94-101`) — every `tests/stack/*.sh` outside
  `SECTIONLESS` must carry at least one `== section ==` header. Catches a module emptied out.

Neither catches the third shape: **the `source` line is deleted from `run.sh` and the module file
is left in place.** The first gate iterates the source lines that are *present*, so a line that is
gone is never examined. The second gate finds the orphaned file on disk with its section headers
intact and passes it. There is no reverse check — nothing asks "is every module file on disk
actually sourced by anything".

The two entrances are the same hole. A missing file and a missing `source` line both end at a green
suite that skipped every assertion in that domain; #1336 closed the first entrance and left the
second open.

## What was run

At `a368ee4` (`develop-v2` tip at the time of writing), fetched via
`gh api repos/.../contents/<path>?ref=a368ee4`:

- `wc -l tests/stack/run.sh` → 4896
- `grep -cE 'source "\$HERE/' tests/stack/run.sh` → 22
- `ls tests/stack/test-*.sh | wc -l` → 19 hyphen-named modules, plus `lib.sh`
- `grep -n "PASS" tests/inventory.sh scripts/*.sh .github/workflows/*.yml` → no floor, no minimum,
  no cross-run comparison anywhere
- Read of `tests/inventory.sh:94-121` — confirmed the source-target loop is one-directional and the
  only other gate is the per-file section count

The tree is consistent today; the source-line count and the module count agree. This is a latent
guard gap, not a live defect.

## Why it is worth closing now

[...]

PR #1363, opened 2026-08-24
fix(tests): catch a domain file that nothing sources (#1357)

VijitSingh97:
tests/inventory.sh held only one direction of the agreement between
run.sh's source lines and the domain files beside it: every file a
`source` line names must exist. The reverse was open — a `source` line
deleted while its file stays on disk. That loop can only examine the
source lines that are present, so a deleted one is never looked at, and
the per-file section gate then finds the orphaned file with its headers
intact and passes it. run.sh runs under `set -uo pipefail` with no -e
and exits on its failure counter alone, so either shape ends the same
way: a suite that reports a pass having skipped every assertion in that
domain.

Both directions now read one SOURCED list, so the pattern cannot drift
between them, and a pattern that stops matching trips the emptiness
guard rather than quietly satisfying both loops at once.

The extraction is anchored to the start of the line, and that anchor is
the difference between a guard and a grep. Unanchored, a commented-out
`# source "$HERE/x.sh"` still matches, so the file counts as sourced
while run.sh never runs it — the same silent skip by another route, and
the likeliest thing for the rebase or conflict resolution #1357 worries
about to leave behind. C3 below is the control that proves the anchor is
what catches it rather than the surrounding code.

Mutation-tested, because a check that greps for the right text without
failing on the defect closes the issue and leaves the hole open. Every
case asserts on the one sentence only its own guard writes and asserts
the other's is absent — two guards sharing an output string can silently
cover for each other's deletion:

  M0  unmutated tree                    -> rc 0, silent
  M1  source line deleted, file kept    -> rc 1, the new sentence only
  M1c source line commented out         -> rc 1, the new sentence only
  M3  module file deleted, line kept    -> rc 1, the existing sentence
  M4  source pattern broken             -> rc 1, emptiness guard
  M5  source line indented              -> rc 0, anchor allows leading
                                           whitespace, so it does not
                                           over-restrict
  C1  M1  against pre-change code       -> rc 0  <- the hole itself
  C2  M1c against pre-change code       -> rc 0  <- the hole itself
  C3  M1c, anchor removed from new code -> rc 0  <- the anchor is what
                                           catches M1c

What is deliberately not claimed: a `source` line that is live text but
dead code — inside a branch that never runs, or a function nothing calls
— still reads as present. Telling those apart needs control flow rather
than grep. The code comment and docs/dev/testing-strategy.md both say so
rather than leaving the reader to infer the guard is absolute.

The generated inventory output is byte-identical to pre-change, verified
by running both scripts from a tree root in a scratch copy: 3076 lines
each, md5 f7ff7e1eba780f3a1bb96fc138c21c72. This also drops the stale
"18 hyphen-named domain files" count from the comment being rewritten:
it was 22 at this tip, and any count there goes stale on the next cut.

Run: shfmt -i 4 -d and shellcheck on tests/inventory.sh (clean at full
severity apart from two pre-existing intentional SC2016 info notes on
the literal `$HERE` pattern), plus lint-file-budget, lint-docs-voice and
lint-operator-strings — each rc=0. make lint-sh and make test were NOT
run locally; CI runs both.

### DIFF
--- a/docs/dev/testing-strategy.md
+++ b/docs/dev/testing-strategy.md
@@ -226,7 +226,13 @@ sudo tests/os/run.sh --image os/rauc/build/system.img
 What gates a merge vs. a release, the standards every test holds to, and the known gaps. For the
 full enumerated coverage, `make test-inventory` generates the list on demand (git-ignored — read it
 locally). The generator is grep-based, so CI runs it on every PR and fails if any suite it
-enumerates counts zero — the drift a moved or reshaped suite would otherwise hide.
+enumerates counts zero — the drift a moved or reshaped suite would otherwise hide. It also holds
+`tests/stack/run.sh` and the per-domain files it sources to each other in both directions: a file
+named by a `source` line has to exist, and a file on disk has to be sourced by something. Either
+mismatch ends the same way — `run.sh` does not stop on a failed `source`, so the suite reports a
+pass having skipped every assertion in that domain. It reads those `source` lines as text, which
+catches one that is absent or commented out but not one that is present and never reached; a
+handful of suites are invoked as their own CI step instead and are listed as exempt in the script.
 
 ### What runs where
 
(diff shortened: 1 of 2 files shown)


VijitSingh97:
## Reviewer: MERGE. Re-derived the mutation table by running it, plus one shape the table does not cover.

Everything below was run in a throwaway worktree at this branch's tip, not read off the body.
The worktree is removed.

**The control is the part that matters, and it holds.** Comment out
`source "$HERE/test-harness-tooling.sh"` in `tests/stack/run.sh`, leave the file on disk, and:

- against **this branch's** `tests/inventory.sh` — **rc 1**, with the sentence only this guard
  writes: *"is on disk but nothing sources it — a 'source' line was deleted and the suite would
  still pass, silently skipping the file's sections"*.
- against **`origin/develop-v2`'s** `tests/inventory.sh`, same mutated tree — **rc 0**.

So the hole is real, and this change is what closes it. Restoring the line returns rc 0, which is
the check that the red came from the mutation rather than from the tree.

**One shape the body's table does not cover, and it is the one this branch will meet first.** #1368
is open and adds a genuinely NEW domain file plus its `source` line — not a mutation of an existing
one. Both directions behave:

- new `tests/stack/test-zz-probe.sh` **with** a `source` line added to `run.sh` → **rc 0**
- the same new file with the `source` line **removed** → **rc 1**

So the two PRs compose, and the new direction works on a file that never existed rather than only
on a line that was removed. Worth stating because a guard that passes on mutation and reds on a
legitimately new file would block #1368 for a reason nobody would look for.

**On the anchor.** C3 in the body is the right control to have written — it is what separates this
from a grep, and the commented-out case is the likeliest residue of the rebase #1357 worries about.
I did not independently re-run C3; M1c passing on this code and failing on the old code already
pins the behaviour I care about.

### What I did not do

`make lint-sh` and `make test` — not run by me either. `tests/inventory.sh` is not in
`docs/dev/file-budget.tsv`, so there is no ceiling interaction here. I did not exercise the exempt
list (`UNSOURCED`) or the emptiness guard (M4); both are asserted in the body and neither is on the
path I re-derived.

Issue #1345, opened 2026-08-24
Worker Inspect: carry rigforge.config_meta so the dashboard can tell a change made ON the rig from one made from here


VijitSingh97:
## What is dropped

`parse_rigforge()` in `dashboard/mining_dashboard/client/xmrig_client.py` is an explicit sub-key
allowlist over the enriched `rigforge` block. RigForge publishes three keys the dashboard never
reads: `config`, `config_meta` and `control`.

`config` is being picked up as part of #1235 (the Worker Inspect editor now prefills from the rig's
own values instead of from our last-applied record). This issue is about the second one.

## Why `config_meta` is worth carrying

`config_meta` is the config revision plus last-change provenance — the producer for it is
rigforge#254, closed and shipped, and it exists precisely so the stack can detect a change made
directly on the rig.

That is the other half of what #1235 complained about. #1235's own words: even an edited rig
"only shows what *the dashboard* last set, not the rig's actual current state". Prefilling from
`config` fixes the value being wrong. It does not tell the operator that the value they are looking
at was changed on the rig rather than from here — which is the difference between "this is what I
set" and "someone changed this underneath me".

Concretely, it would let the Worker Inspect history say something honest in the case it currently
cannot describe at all: a rig whose config no longer matches the last thing Pithead applied to it.

## Deliberately not done in #1235

Left out to keep that change small — it was already fighting the file-budget ratchet — and because
#1235's acceptance criterion is satisfied without it: the editor shows the rig's current values,
secrets masked, matching what the rig is running.

## Scope

- Carry `config_meta` through the `parse_rigforge` allowlist, filtered the same way `config` is.
- Decide what the UI does with it. The cheap version is a line in the config history when the rig's
  revision does not match our last-applied record. The expensive version is a per-field "changed on
  the rig" marker, which needs a value-level comparison rather than a revision one.
- `control` is the third dropped key; whether it is worth carrying is a separate question and this
  issue does not answer it.

## Not verified

Nothing here has been checked against a live rig. The claim that `config_meta` is emitted comes from
reading `_api_rigforge_block()` in the RigForge source, not from observing a real feed.

VijitSingh97:
Closed by PR #1366, merged to `develop-v2` as `4685945`. Closing by hand — `Closes #N` does not fire in this repo, because PRs land on `develop-v2` while the default branch is `develop`.

[...]

Issue #1313, opened 2026-08-23
The weekly CVE sweep scans a rebuild of develop, never the images users run


VijitSingh97:
`ci.yml`'s Monday sweep is the only recurring CVE scan on the product lane, and by construction it
cannot report on the artifact users are running. Two independent reasons, either one enough on its
own.

## It scans the wrong ref

Scheduled runs use the default branch, `develop` — the workflow header says so outright. What users
run is the release, and since #1076 `main` is a fast-forward pointer to the released tag (`main` is
at v1.20.0 today). Between cuts the two diverge: `develop` takes the pin bumps, `main` keeps
whatever shipped. A base image bumped on `develop` turns the sweep green while the shipped image is
still carrying the CVE.

## It scans the wrong artifact, which is the bigger half

`build-images` runs `docker build ./build/<service>` and points Trivy at the freshly built
`pithead-<service>:ci` (`ci.yml:108`, `:124`). Users pull
`ghcr.io/p2pool-starter-stack/pithead-<service>:vX.Y.Z` — bytes built once at cut time and then
promoted by digest with no rebuild (`release.sh` stage 6 is `buildx imagetools create`, a re-tag).

A rebuild resolves apt at scan time, so it silently picks up archive fixes the published image never
got. That is exactly the case the sweep was created to catch: its own header calls out a CVE "fixed
in the apt archive without a new base digest, which Dependabot never sees". In that case the rebuild
goes green and the image users are running stays vulnerable.

So retargeting the sweep at `main` — the obvious first move — fixes the first half only. It would
still be scanning a rebuild.

## The fix that covers both

Scan the published images directly. Trivy takes a registry reference, so the build step disappears
entirely:

```yaml
image-ref: ghcr.io/p2pool-starter-stack/pithead-<service>:v1.20.0
```

The packages are public, so this needs no registry auth. Read the tag (better: the digests) from the
latest release rather than hardcoding it, so the scan follows each cut with nothing to remember.
Five `docker build`s leave the weekly run, which makes it cheaper as well as honest.

Keep the existing `develop` rebuild as it is. It answers a different and still-useful question —
does what we are about to ship have a CVE — and it is the only one that can go red *before* a
release.

Two details for whoever picks this up:

- Mutes: scan the shipped image against the `.trivyignore` that shipped with it (`main`'s), not
  `develop`'s. Otherwise a mute added after the cut silences a finding in an image whose review
  never saw it.
[...]

Issue #530, opened 2026-07-15
Security panel: audit out-of-band (detected) config + rig changes, with hour/day/month grouping


VijitSingh97:
## Context — what already exists

The Security panel (`securityview.mjs`, #349) already renders **two** cards: the **Access log** (`AccessCard`) and an **audit-log** card (`AuditCard`, line ~66), fed by `service/audit_service.py` reading the #33 host-side audit trail (`/control/audit/control.log`). It records every change **made through the dashboard/control channel**: `preview`, `commit`, `worker-apply` (the #185 Worker Inspect rig edits), and `upgrade` — actor + action + status + changed key **names** only, no secret values.

So changes we *make* are already tracked. This issue is about two gaps.

## Ask 1 — track changes we DETECT, not just ones we make

Today a change only lands in the audit log if it went through the control channel. Changes made **out-of-band** are invisible:

- **Dashboard/host side**: someone edits `config.json` and runs `./pithead apply` on the host (not via the dashboard). Nothing records it.
- **Miner side**: a rig's config changes locally (a `rigforge.sh apply` on the miner, or an autotune run rewriting `tune.json`). We can *see* it — the enriched feed already carries the rig's state/version and RigForge has its own control `/status` "last applied change" — but we don't fold it into the audit trail.

Proposal: detect and append audit entries for out-of-band changes.
- Dashboard config: keep a hash/snapshot of the last-known `config.json`; on each poll, if it changed without a matching control-channel commit, record a `host-edit` audit entry with the changed key names (diff, names-only, same sanitization as today).
- Miner config: poll each rig's control `/status` (the "last applied change" record) and/or the enriched feed's config marker; when a rig's applied-change id/version advances without a matching dashboard `worker-apply`, record a `rig-edit` entry naming the worker + the changed keys we can see.

This turns the panel from "what the dashboard did" into "every change to the dashboard or a rig that we can detect."

## Ask 2 — group the audit log by time

The `AuditCard` is a flat newest-first tail. Add **hour / day / month** grouping via dropdown selectors in the panel, so an operator can drill from a month down to a specific hour. Applies to the audit card (and the access log could get the same treatment). Keep growth bounded — the writers already trim `control.log`; a deeper time view may need the audit persisted to the dashboard DB (like the per-worker history, #185) rather than only a log tail.

Relates to #349 (the Security panel), #33 (control audit), #185 (per-worker change history), #196 (persist telemetry as time-series — the storage angle).

[...]

PR #1366, opened 2026-08-24
feat(dashboard): say where a rig's running config came from (#1345)

VijitSingh97:
Carries RigForge's `rigforge.config_meta` through to Worker Inspect and turns it into the verdict an operator actually wants when they disagree with their own record: **did this change come from here, or did something change it underneath me?**

Not a new capability from nothing. #530 already notices a rig-applied change — but only while a poll happens to watch a terminal report go by, and only ever as a bare change id. These keys make the same answer persistent and specific: it survives a restart, and it survives nobody watching.

## What the operator sees

A line above the change history, in the rig's own words, with our verdict on them:

| Verdict | Means |
|---|---|
| Last changed from this dashboard | the id is one of the history rows rendered directly below |
| Last changed from another dashboard | applied over a control channel, with an id this rig's history does not contain |
| Last changed on the rig itself | applied on the rig, not over a control channel |
| Last restored from a saved config | including RigForge's own automatic rollback |
| No recorded config change | running a config whose change was never recorded |

A rig with no RigForge, or one too old to serve the block, renders **nothing at all** — silence, not "unknown". An absent block is a rig that made no claim, and inventing one would manufacture a suspicion the payload does not support.

## The decisions worth reviewing

- **`source` is an allowlist (`control|local|restore`), not a pass-through.** A compromised rig must not be able to write its own provenance in words the operator reads as ours.
- **`restore` stays its own verdict.** RigForge stamps it for its *own* automatic rollback after a change fails to hold. Folding it into "the rig was edited" would accuse a rig that healed itself.
- **`revision` is treated as an opaque token.** It is a truncated SHA-256 over a jq-canonical form; reimplementing that byte-for-byte here could report drift on a rig that has none.
- **A fresh rig serves a real `revision` with the other three null.** That survives as "unrecorded", which is deliberately distinct from an absent block.

## Security review found two real defects — second commit

A `security-reviewer` pass on the first commit caught the id comparison reaching for `worker_config_change_known`, the #530 audit's lookup. Two of its properties are correct for #530 and **inverted** here:

- **Unscoped.** It asks whether *any* rig's change carried this id. A rig reporting an id spooled for a *different* rig was told "Last changed from this dashboard".
- **Fails open** (returns `True`) on a sqlite error. For #530 a false "known" merely declines to accuse a rig. Here it hands out the one verdict the feature exists to protect, on a transient DB fault.

Fixed by matching against the rig's own history rows: `get_worker_config_history` is worker-scoped by construction, returns `[]` on a read error, and `build_worker_detail` already holds its result — so the fix is also one query *cheaper*. It makes the verdict checkable, too: "here" now means "the id is one of the rows right below this line".

`storage_service.py` and its test file are both exactly at their budget ceilings, so this deliberately adds no method there — which turned out to be the better design anyway.

**A limit this cannot close, and the PR says so:** a rig that has been taken over controls its whole response and can replay an id we really did send it. This detects a change we did not make; it does not defeat a rig that has decided to deceive us. What it guarantees is the *direction* of every error — anything unvouched-for reads as not-from-here.

## What was RUN

Each exit code checked separately, never through a pipe.

- `pytest -q` (dashboard) — exit 0
- `node --test dashboard/tests/frontend/` — 474 pass / 0 fail
- `lint-py`, `lint-js`, `lint-md`, `lint-docs-voice`, `lint-topology`, `lint-operator-strings`, `lint-file-budget` — each exit 0

**Nine mutations run, each seen to red the one test that claims it:** id-comparison dropped · `restore` folded into `rig` · absent block reporting "unrecorded" · `source` allowlist removed · the `rig_config_meta` key blanked · "elsewhere" relabelled as ours · the silent case made to render "unknown" · and both security regressions re-introduced by reverting to the vulnerable call.

One of my own tests was **vacuous on the first attempt** and is worth flagging: the fail-open test set `_conn = None`, which hits an early return *above* the fail-open line, so it passed against the vulnerable code too. It now closes the sqlite handle with `_conn` still set, which is what actually reaches the `except sqlite3.Error` arm. Caught by reverting the fix and checking the test noticed — not by reading it.

## The case this does NOT catch — fourth commit, and #1367

Checked against the issue's own acceptance text rather than my summary of it, and found an
over-claim. #1345's motivation is *"a rig whose config no longer matches the last thing Pithead
applied to it"*. This line reports the last change RigForge **recorded** — and a config hand-edited
underneath RigForge is not a recorded change, so the provenance stays at whatever came before it.
Where that was our own control change, it reads **"Last changed from this dashboard" over a config
we did not set**: a false reassurance, the one direction the rest of this feature is built to avoid.

The rig holds both halves of the comparison and serves one. `_stamp_config_meta` writes the revision
as of the recorded change into its marker file; `_api_config_meta_json` then overwrites it with the
live one before serving (`rigforge.sh:4840`), so the value that would catch this never reaches the
wire. Closing it needs the last revision we *observed* per rig — persistence blocked today by
`storage_service.py` and its test both sitting exactly at ceiling. Filed as #1367 with two routes
and why the obvious one (recompute the revision our side) does not work.

The docs, the module docstring, and a **characterization test** now all name it. That test asserts
the known-wrong answer on purpose and says so: when #1367 closes it should red, and its docstring
says to update it rather than delete it. Prose can be skimmed past; a test that reds cannot.

## NOT done

- **No live rig has been exercised.** RigForge's semantics here come from reading `rigforge.sh` and `util/control-server.py`, not from observing a feed. The block's field names, the `source` vocabulary, and the `os.urandom(8).hex()` change-id minting are all read from that source, and a live rig should confirm them before this is relied on.
- The provenance line has not been looked at in a browser; it is covered by render-probe tests only.

## Shared files

`docs/dashboard.md` (shared, staged normally — no override) and `docs/dev/file-budget.tsv`, which **ratchets two ceilings DOWN**: `views.py` 2339 → 2263 and `workerview.mjs` 479 → 443, from the two extractions the ratchet forced. Both cuts are along real seams — `build_worker_detail` plus its exclusive helper to `web/worker_detail.py` (one-way import from `views`, no cycle), and the change-history vocabulary to `static/confighistory.mjs`.

## Noticed in passing, not mine to fix

`.github/workflows/ci.yml` (513) and `scripts/shipped-image-sweep-report.sh` (429) came in with #1313 and are both over the 400-line target with no budget entry, so nothing ratchets them until they cross 800. The gate is green either way — flagging for whoever owns those paths.

## Ponytail pass

One finding, applied in the third commit: `configOriginNote`'s tooltip repeated the `changed_at` that `ConfigProvenance` already renders inline on the same line. Dropped — the tooltip keeps the revision and the per-verdict caveat, neither of which is visible elsewhere. The test that asserted the timestamp *was* in the title now asserts it is not, and that the visible line carries it, so the redundancy cannot return unnoticed.

Nothing else was cut. The token filter is a security boundary; the logic/render split matches the `markerLabel`/`buildChartMarkers` pattern already in `workerlogic.mjs`; and the long "why not `worker_config_change_known`" comment in `worker_detail.py` is what stops the next reader reinstating the bug the security pass just found.

---

## Fifth commit — a false "here" over a rolled-back change

The verifier pass this PR was waiting on found a real defect, confirmed against RigForge's source and fixed in `de3d12c`.

**What was wrong.** `change_known` matched the rig's `last_change_id` against this rig's history rows **by id alone, ignoring the row's status**. RigForge's rollback re-apply re-stamps the *same* change id it just reverted: `control_apply()` sets `local RIGFORGE_CONFIG_SOURCE=control RIGFORGE_CONFIG_CHANGE_ID="$cid"` once and both apply attempts run under it, and the comment there says it means to — *"attribute this (and the rollback re-apply) to the control path with its change_id"*. `_stamp_config_meta` fires on the second apply because the restored config's hash differs from the failed one. So a control change that did **not** come back live still matched, and the line printed the calm "Last changed from this dashboard" directly above a red "Rolled back" row for that same change — a self-contradiction on screen, in the one case this feature most needs to be right.

**The fix.** The verdict now reads the matched row's status rather than merely whether a row exists. A row this dashboard records as `rolled_back` or `failed` gets its own verdict, `reverted` → *"Last change from this dashboard was rolled back"*, flagged, with the tooltip explaining that the rig is running whatever config came before.

`rejected` is in `REVERTED_STATUSES` too, although it **cannot reach us today**: `_control_commit` rejects and returns before `config.json` is touched, so a rejected change is never stamped and its id never becomes `last_change_id`. It is listed so the check fails closed rather than open if that apply path ever changes.

**The same wrong belief had spread to four places**, all corrected: the `CONFIG_SOURCES` comment, the `config_origin` docstring, the operator-facing list in `docs/dashboard.md`, and a frontend test that asserted the restore tooltip must *not* name a person "because RigForge also does this on its own after a failed change". It does not — `restore` is stamped only by the operator-run `restore <archive>` command (one call site); the automatic rollback is `control`.

Also corrected: a doc line the tooltip-dedup commit left stale. It promised "Hover the line for the rig's timestamp and the revision"; that commit deliberately removed the timestamp from the tooltip, and it renders inline instead.

**Mutation evidence** — each new guard reverted and seen to red the tests that name it, then restored and re-verified green:

| Mutation | Result |
|---|---|
| `config_origin` drops the status check | 5 red (3 parametrized unit + 2 integration) |
| caller drops the status argument | 2 red |
| frontend `reverted` text collapsed back onto `here` | 1 red |

Gates re-run from the repo root with each exit code checked separately: `pytest` rc=0, `node --test` 475/475, `lint-py` / `lint-js` / `lint-topology` / `lint-md` / `lint-docs-voice` / `lint-operator-strings` / `lint-file-budget` all rc=0. No budgeted ceiling is touched — none of the changed files has a budget entry.

## Still open, not fixed here

**The 50-row history window can flip a true "here" into a false "elsewhere".** `build_worker_detail` matches against `get_worker_config_history(name)` at its default `limit=50`. Every apply attempt writes a row, so 50 recorded changes *after* the one the rig is running push it out of the window; the match fails and the line reads "Last changed from another dashboard" — an alarm, over a change that genuinely came from here. The existing comment calls this "wrong in the direction that under-claims", which understates it: the operator reads that line as a warning, not as "we did not look far enough back".

Fixing it properly needs a worker-scoped exact-id lookup, i.e. a new `StateManager` method — and `storage_service.py` (1603) and `tests/service/test_storage_service.py` (1646) are both exactly at their ceilings, which is the same wall that blocks #1367. Filed as #1369 rather than smuggled in here.

### DIFF
--- a/dashboard/mining_dashboard/client/rig_config_meta.py
+++ b/dashboard/mining_dashboard/client/rig_config_meta.py
@@ -1,0 +1,115 @@
+"""The rig's own account of when its writable config last changed, and what changed it (#1345).
+
+RigForge v1.8.0 and newer stamp ``rigforge.config_meta`` onto the enriched ``/1/summary`` feed
+(rigforge#254): ``{revision, changed_at, source, last_change_id}``. #1235 taught the Worker Inspect
+editor to prefill from the rig's own ``config``, which fixed the *values* being wrong. It did not
+answer the question the operator actually has when they disagree with their own record — did this
+change come from here, or did something change it underneath me?
+
+Two of these fields answer that, and they answer it in different ways:
+
+``revision`` is recomputed by the rig from the config it is running, every time it serves the feed.
+``changed_at`` / ``source`` / ``last_change_id`` come out of a marker file the rig writes when it
+*records* a change. So a config edited on the rig and applied through RigForge stamps all four; a
+config file edited underneath RigForge with nothing recording it moves only the revision.
+
+The revision is a truncated SHA-256 over a jq-canonicalized form of the writable keys. Recomputing
+it here would mean reproducing that canonical form byte for byte — key order, null handling, number
+formatting — so this treats it as an opaque token to be shown and compared, never as something to
+verify. A comparison we could get subtly wrong would report drift on a rig that has none.
+
+Everything here is remote-supplied, from a device that picks its own response, so each field is
+validated to the shape RigForge actually produces rather than passed through. That is the #1235
+lesson applied one field further on: a filter that assumes the shape of the input it is defending
+against is not a filter.
+"""
+
+import re
+
+# Every value RigForge stamps into ``source`` (its ``_stamp_config_meta``): a change applied over
+# the control channel, one applied on the rig itself, and one restored from a saved config — which
+# includes the rig's own automatic rollback after a failed change, so it is NOT evidence that a
+# person touched the rig. Anything else is a rig speaking a dialect we do not know; it becomes None
+# rather than reaching the operator's screen, because the whole value of this line is that the
+# operator can trust who it names, and free text from the rig would let a compromised one write its
+# own provenance.
+CONFIG_SOURCES = ("control", "local", "restore")
+
+# Long enough for the ids RigForge actually mints (a 16-hex change id, a 16-hex revision) with room
+# for a future format; short enough that a rig cannot push a wall of text into the operator's view.
+_MAX_TOKEN_LEN = 64
+
+# The exact stamp RigForge writes: `date -u +%Y-%m-%dT%H:%M:%SZ`. Matched rather than parsed — this
+# only needs to decide whether to show the rig's word for it, and a value that is not that shape is
+# not a timestamp we can render honestly.
+_TS_RE = re.compile(r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z$")
+
+# Ids and revisions are hex today. Allowing the usual id punctuation costs nothing and keeps a
+# future RigForge format from reading as a hostile value.
+_TOKEN_RE = re.compile(r"^[A-Za-z0-9._-]+$")
+
+
+def _token(value):
+    """A short opaque id, or None. Anything that is not a plausible id is not shown at all."""
+    if not isinstance(value, str) or not value or len(value) > _MAX_TOKEN_LEN:
+        return None
+    return value if _TOKEN_RE.match(value) else None
+
+
+def parse_config_meta(meta):
+    """Normalize the rig's ``config_meta`` block, or None when the rig said nothing usable.
+
+    None is the honest answer for a plain-xmrig rig, for a RigForge older than v1.8.0, and for a rig
+    whose block is unreadable — all three cases mean the same thing to the operator, which is that
+    this rig cannot say where its config came from. A fresh RigForge rig that has simply never had a
+    config change is different: it serves a real ``revision`` with the other three null, and that
+    survives here as a meta with a revision and no provenance.
+    """
+    if not isinstance(meta, dict):
+        return None
+    source = meta.get("source")
+    out = {
+        "revision": _token(meta.get("revision")),
+        "changed_at": (
+            meta["changed_at"]
+            if isinstance(meta.get("changed_at"), str) and _TS_RE.match(meta["changed_at"])
+            else None
+        ),
+        "source": source if source in CONFIG_SOURCES else None,
+        "last_change_id": _token(meta.get("last_change_id")),
+    }
+    return out if any(out.values()) else None
+
+
+def config_origin(meta, change_id_known):
+    """Where the rig's current config came from, as far as we can honestly tell.
+
+    ``change_id_known`` is whether this dashboard has a ``worker_config`` row for the rig's
+    ``last_change_id`` — the ids are minted by the rig's control server and handed back to us in the
+    202, so a control change we made is one we can recognise by id. That comparison is what
+    separates the two cases the operator most needs told apart:
+
+    - ``here``      — applied over the control channel, with an id in our own history.
+    - ``elsewhere`` — applied over a control channel, with an id we have never seen. Another host,
+                      or our record of it is gone. Either way it is not something to present as ours.
+    - ``rig``       — applied on the rig itself.
+    - ``restored``  — restored from a saved config, which the rig also does on its own after a
+                      failed change. Deliberately not folded into ``rig``: that would tell the
+                      operator someone edited the rig when the rig may simply have rolled back.
+    - ``unrecorded``— the rig is running a config whose change it never recorded. A fresh rig that
+                      has never been changed looks exactly like this, and so does one whose config
+                      file was edited underneath RigForge, so this claims neither.
+
+    Returns None when there is no meta at all, which the UI must render as saying nothing rather
+    than as saying "unknown" — an absent block is a rig too old to answer, not a suspicious one.
+    """
+    if not meta:
+        return None
+    source = meta.get("source")
+    if source == "control":
+        return "here" if change_id_known else "elsewhere"
+    if source == "local":
+        return "rig"
+    if source == "restore":
+        return "restored"
+    return "unrecorded"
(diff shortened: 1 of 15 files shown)


VijitSingh97:
## Verdict at `c666f73`: merge-ready. The sharp question was already found and pinned by the author.

CI settled: 16 pass, 2 skipped, 0 failure. Reviewed for the one thing this feature can do worst —
state a config's origin confidently while being wrong.

### That case exists, and it is handled the right way

`c666f73` found it before I did: the line reports the last change RigForge **recorded**, and a
config hand-edited underneath RigForge is not a recorded change, so the provenance stays at whatever
came before. Where that was our own control change, it reads "Last changed from this dashboard" over
a config we did not set — a false reassurance, the one direction the rest of the feature is built to
avoid. Filed as the follow-up issue with the reason the cheap fix does not work (the rig overwrites
the marker's stored revision with the live one before serving, so the value that would catch it
never reaches the wire).

What makes this acceptable to merge rather than a blocker is how it is pinned. There is a
characterization test that asserts the **known-wrong** answer on purpose, with a docstring saying it
should red when the follow-up closes and to update it rather than delete it. Verified it exists and
asserts `"here"` on a moved revision with unchanged provenance. Prose can be skimmed; a test that
reds when the bug is fixed cannot. The doc, the module docstring, and the test all name the same
case.

### The fail-closed property is real — proven by mutation, not by reading

`b9df65e` replaced an unscoped, fail-open lookup (`worker_config_change_known`, which asks whether
*any* rig's change carried this id and returns True on a DB error) with a scan of this rig's own
history rows, which is worker-scoped by construction and returns `[]` on error.

Baseline: 30 tests pass. Reverting that hunk to the old unscoped call reddens exactly two, by name:

- `test_another_rigs_change_id_is_never_claimed_as_this_rigs`
- `test_a_history_read_that_fails_never_manufactures_trust`

Both directions of the property have their own guard, and nothing unrelated moves. That is the
difference between a comment claiming "fails closed" and a guard that does.

### Two things I checked that turned out fine

**The verdict is not render-state dependent.** `history` is `get_worker_config_history(name)` with
no range or window argument, so it is the same unranged list the table below renders — which makes
the docstring's claim that an operator "can confirm it by eye" exactly true. I expected the range
selector to be able to flip the verdict; it cannot. The `limit=50` default does mean a change older
than the 50 most recent rows reads as "elsewhere", and that under-claims rather than over-claims,
which is the safe direction and is already documented.

**No control-flag gap.** Every test sets `DASHBOARD_CONTROL_ENABLED` true, which looked like a
coverage hole until I checked whether the provenance path branches on it. It does not — the flag is
passed through to the response and is not consulted by `config_origin` or by the `change_known`
computation. So there is no second path to cover.

### Not run

The full dashboard suite, `make lint`, and the frontend node tests — I ran the two provenance test
files and the mutation above, nothing wider. CI covers the rest and is green at this tip.

### The skipped CVE sweep

Same answer as on the other PR, and it is a filter doing its job: `sweep-shipped` carries
`if: github.event_name == 'schedule'`, so a pull request must skip it. This lane's `schedule:` block
is inert by design — cron is read from the default branch only — and `ci.yml` documents that. The
part worth attention is elsewhere: that job landed on `develop` after the last scheduled run, it
appears in none of the 13 scheduled runs on record, so it has never actually fired. Next window is
05:00 UTC today; prove the first firing from `gh run list --event schedule`.

Recommend merge. Not merging myself.

Issue #1064, opened 2026-08-16
BLOCKER: the documented release gate runs 5 of 8 battery phases, and verify-image's two strongest checks self-skip while reporting '0 failed'


VijitSingh97:
Two findings from the v2 gap analysis, both proven by reading the code, both the same shape: **a gate that reports success without having run.** Related to #1048, which is the CI half of the same problem.

## 1. `--phase all` is not all

`tests/os/run.sh`'s `all)` dispatch calls `phase_boot`, `phase_update`, `phase_install`, `phase_provision`, `phase_rig` — five of eight. `phase_fault`, `phase_media` and `phase_reset` are reachable only by naming them explicitly.

`docs/dev/appliance-release.md:210` states that "`--phase all` runs everything", and step 1 of *Cutting a release* prescribes exactly `tests/os/run.sh --phase all`.

So a maintainer who follows the release checklist literally believes the **three mid-write power cuts, the mid-commit power cut, the corrupt-bundle refusal, the factory reset, the wedged-`/data` recovery and the media channel** all ran. None of them did. The only other power-loss proof is manual item M8, which has never been run.

`docs/dev/testing-strategy.md` scores the fault row as passing on the strength of a run dated 2026-07-25 in `os/KNOWN-ISSUES.md`.

**Fix:** either add the three phases to the `all` arm, or rewrite the release step to name every phase the cut must run. Then record a dated per-phase pass count in the release issue, for the tip actually being cut — the current arrangement lets an old green stand in for a new one.

## 2. `verify-image.sh` skips its two strongest checks silently

- `tests/os/verify-image.sh:213` wraps the built-from-expected-commit and clean-tree checks in `if [ -n "${PITHEAD_EXPECT_COMMIT:-}" ]`.
- `:222` wraps the entire artifact-vs-tree block — shipped `pithead`, compose and config reference compared byte-for-byte, plus the baked container archive unpacked to compare `wizard.py` — in `if [ -f ./pithead ] && [ -f build/dashboard/.../wizard.py ]`.

Neither skip prints anything, and the summary at `:315` prints `verify-image: N passed, 0 failed` either way.

These are precisely the checks that exist because a release image once shipped a dashboard two commits stale and passed everything else. The only automated caller, `tests/os/run.sh:226`, runs `--test` mode with `PITHEAD_EXPECT_COMMIT` unset — so the stale-tree guard is off in the one invocation that is not a human typing a command.

[...]

PR #1368, opened 2026-08-24
test(stack): split the appliance-rig-miner domain into its own file (#1105 Phase 1, develop-v2)

VijitSingh97:
Seven contiguous sections — the #797 rig role and the #796 built-in miner —
move out of tests/stack/run.sh into tests/stack/test-appliance-rig-miner.sh,
sourced back in at their exact former position. A pure relocation: no line
of moved code is edited, and the only authored content is the new file's
header.

run.sh 4475 -> 4178. New file 339 lines (40 header + 299 moved verbatim).

Left behind on purpose, against the #1252 map's estimate of 492: the generic
boot mechanics these sections sit among — pithead-boot's wiring of the miner
leg after the slot commit, the A/B commit gate reading doctor --json (#852),
the boot health probe (#1140), the self-reboot on a failed health gate
(#1065) and the appliance battery's release gate (#1064) — plus
pithead-sync's rigforge leg, which is not contiguous with this block anyway.
The appliance-boot cut still to come needs those where they are. A
defensible 299 beats a forced 492; arithmetic does not get to settle an
ownership contest.

Two factual errors in that header were found by adversarial review and are
fixed here. Both mattered because the header is the only authored content in
the diff, and the appliance-boot cut inherits this file's leave-behind, so
its author reads this header as the reference:

  - It enumerated seven sandbox trees where the file creates eight. $RPESP
    was missing from a list phrased as exhaustive.
  - It claimed the PITHEAD_* and RF_* names "are command prefixes and so
    never enter this shell to begin with — they need no unset, and adding
    one would be wrong." That is true of PITHEAD_APPLIANCE,
    PITHEAD_HUGEPAGES_MARKER, PITHEAD_INSTALL_BIN, PITHEAD_RIG_PROBE and
    RF_RC, and false of RF_LOG, PITHEAD_RIGFORGE_DIR,
    PITHEAD_JOURNALD_DROPIN_DIR, PITHEAD_JOURNAL_DIR and PITHEAD_PRESEED_DIR,
    which the file exports standalone and which it correctly unsets. The
    header was telling the next author to delete unsets this file needs.

The replacement states the rule rather than a roster, since a roster is what
went stale: the assignment form decides, so read the form in front of you
rather than the prefix of the name.

Byte identity re-derived from the committed base rather than from the
extraction, which is what makes it a proof instead of a tautology. Re-derived
again after the header fix, with the header length recomputed rather than
assumed:
  base 2864-3162 vs the module below its header   sha256 2588bcb7...79eecc
  base 1-2863 + stanza + base 3164-4475 vs run.sh sha256 e9f161af...9f0203
Both diffs empty, and both digests unchanged by the header edit.

Suite proof, before and after:
  2904 passed / 0 failed both sides
  3212 log lines both sides
  269 = 269 section headers, multiset identical
  2904 = 2904 per-test verdict lines
  sorted-multiset diff: 4 lines, all of them the two known #1325
    nondeterministic pairs (load_baked_images heartbeat seconds, ed25519
    keygen fingerprint)
  CONFIG_FILE line-shift predicted ZERO before the run and observed ZERO
    (emitted at 1922 and 1954; this cut starts at 2864, below both)

Those totals were measured at 4f0df21; this branch is now rebased onto
dea75e3, so read them as of that tip and not as a prediction for this one.
What the proof asserts — that the relocation changes nothing — carries, and
the reason is checked rather than assumed: tests/stack/run.sh is byte-identical
at 4f0df21, 4685945 and dea75e3, and neither pithead nor config.reference.json
moved across either merge. The absolute totals WILL be higher here, and that
is not this cut: #1358 added a section header and assertions to
tests/stack/test-harness-tooling.sh, which run.sh sources and which this cut
does not touch, so the addition lands equally on both sides of any comparison.
The suite was NOT re-run at dea75e3 — CI runs it on the merge commit and that
is the empirical check at this base.

Re-derived statically at dea75e3 instead: section headers reconcile exactly,
92 in the base run.sh against 85 + 7 = 92 after the cut.

Coupling audited against both rules that have each cost a red suite, plus
reverse-dependency, unresolved-name and leakage checks. Variables and
functions were enumerated separately and heredoc-aware, because a
\$[A-Za-z_]+ pattern cannot find a function call by construction. Only
$SANDBOX (lib.sh) and $PATH (ambient) are read without being assigned in the
block; everything it introduces is unset before it ends.

docs/dev/file-budget.tsv is regenerated, not hand-merged, and that mattered
here: rebasing onto dea75e3 produced a CLEAN auto-merge that was WRONG. Git
kept this branch's version of the region and dropped both rows #1358 had
added — tests/os/failure-evidence.sh, a new 567-line file, lost its ceiling
outright, and tests/os/run.sh kept a stale 3442 against its actual 3437. The
tsv committed here is a fresh --generate, diffed against the auto-merge
rather than trusted. Worth knowing beyond this PR:
scripts/lint-file-budget.sh reports "file budget OK" and rc 0 against the
dropped-ceiling version, so no gate would have caught it — filed separately.

Besides lowering run.sh 4475 -> 4178 it
picks up pre-existing drift from #1341, which merged without refreshing it:
four dashboard ceilings move down and xvbview.test.mjs gains an entry. Those
rows are still uncorrected on the new base. Nothing is raised; the gate
reports monotonic.

Run: shfmt -i 4 -d and shellcheck --severity=warning on the new file (clean),
scripts/lint-file-budget.sh (rc 0), and the full suite before and after.
make lint-sh and the rest of make lint/make test were NOT run locally; CI
runs them on the merge commit.


VijitSingh97:
## Reviewer: MERGE. Byte-identity re-derived at this tip, by a different route than the body's.

A move PR's whole contract is "nothing changed", so the only review worth doing is recomputing it
rather than reading the digests. Recomputed at the branch tip against the merge-base with
`origin/develop-v2` — both re-derived here, not taken from the body, since the body's numbers were
measured at an earlier base and it says so.

**`tests/stack/run.sh` has exactly ONE diff hunk.** Lines 2864-3162 (299 lines) replaced by two:

    # shellcheck source=tests/stack/test-appliance-rig-miner.sh
    source "$HERE/test-appliance-rig-miner.sh"

Three independent checks, all clean:

1. base lines 2864-3162 vs the module below its 40-line header (lines 41-339) — **empty diff**.
2. base 1-2863 + the two-line stanza + base 3163-4475, concatenated, vs the branch's `run.sh` —
   **identical**. That is the check that rules out an edit hiding elsewhere in the file, which a
   digest of the moved block alone cannot.
3. Arithmetic closes exactly: 4475 − 299 + 2 = **4178**, the observed count. 40 + 299 = **339**,
   the observed module length.

Since hunk count is 1, there is no second edit anywhere in the file, so "no line of moved code is
edited" is a measured property here rather than a claim.

**One correction to the body, and it changes nothing.** The body gives the reconstruction as
`base 1-2863 + stanza + base 3164-4475`; at this base the remainder starts at **3163**, not 3164.
Reconstructing with 3164 drops a line and would not have matched. The body's ranges were measured
at the earlier tip, the proof itself is sound, and the off-by-one is in the prose only — flagging
it because the next cut's author is told to read this header and this body as the reference.

**The header fix was the right call.** Replacing the roster with the rule ("the assignment form
decides, so read the form in front of you") is the correction that does not go stale again — a
roster is exactly what rotted the first time. Same class as the defect this lane just finished
auditing next door on #1345: four artefacts agreeing with each other and none with the source.

**Composes with #1363**, which is open and adds the reverse-direction inventory guard. I ran that
guard against a newly added domain file plus its `source` line: rc 0 with the line, rc 1 without.
So this cut's new file and stanza satisfy both directions and neither PR blocks the other. Details
on #1363.

### What I did not do

I did not re-run `tests/stack/run.sh` at this base — CI runs it on the merge commit and that is the
honest empirical check, as the body says. The suite totals in the body are measured at `4f0df21`
and must not be read as a prediction for this tip; the body already says so, and I am repeating it
because a total quoted in a merged PR outlives the caveat next to it. `make lint-sh` not run by me
either — it is fleet-serialized here and I did not take the lock.

Issue #1364, opened 2026-08-24
The mandated release gate never runs the rigforge-control phase: the whole dashboard-to-rig write surface is matrix-only


VijitSingh97:
The mandated pre-cut e2e runs `--mode targeted`. The rigforge-control phase is added only under `--mode matrix`. So the entire dashboard-to-rig **write** surface has zero coverage in the gate that decides whether a release ships, and the runbook's stated escape hatch is not a flag that exists.

This is not a self-skip. The phase is never requested, so nothing in the output mentions it at all.

## The three lines that make it true

1. **The mandate is `targeted`.** `docs/dev/releasing.md:184`, in the "before cutting" step:

   ```
   BENCH_HOST=<bench> MINER_HOST=<loaner-rig> tests/integration/e2e.sh <ref> --mode targeted
   ```

2. **`--rigforge-control` is matrix-only.** `tests/integration/e2e.sh:694`:

   ```bash
   [ "$BORROW_MINER" = "1" ] && [ "$MODE" = "matrix" ] && phases="$phases --rigforge-control"
   ```

   `targeted` gets `--auth-fail-closed --lifecycle` (`:685`) plus `--rigforge` (`:690`, the read-only feed phase). Never `--rigforge-control`.

3. **`run.sh` only runs the phase when asked.** `tests/integration/run.sh:2534`:

   ```bash
   [ "$RUN_RIGFORGE_CONTROL" = "1" ] && run_rigforge_control
   ```

## What that leaves with no coverage in the mandated gate

Everything `run_rigforge_control` proves:

- **#514** — the enriched READ path survives a populated masked descriptor. This is the v1.5.2 regression that stringified the `{"__secret__":true}` sentinel into the Bearer header and 401'd every probe.
- **#513** — the rig is editable, and a reversible Worker Inspect edit lands on the rig.
- **#516** — a rig-side edit reflects back into the dashboard's enriched feed and the masked prefill.
- **#517** — an auto-rollback is recorded end-to-end from the dashboard.
- **#1002b** — the `pools` write leg.
- **#1002a** — the upgrade noop leg.

Every one of these is a dashboard-to-rig write path that ships to operators, and none of them runs before a cut.

## The escape hatch in the runbook is not executable

`docs/dev/releasing.md:192` tells the operator:

> If the release's diff touches the worker or control-descriptor path, add the `--rigforge-control` legs (needs a rig with its control API enabled).

[...]

Issue #1236, opened 2026-08-22
Miner-panel config-apply legs self-skip in the release gate — run them against the borrowed rig, incl. DONATION


VijitSingh97:
## The gap

The Worker Inspect config-apply flow has real e2e coverage, but most of it **self-skips in the
release gate**, so the gate proves far less than it looks. What actually runs against the
borrowed rig today is the `max_temp_c` apply → verify-on-rig → revert leg (good: it asserts the
rig's own `/status` reports `changed_keys`, not just that Pithead recorded it). What does NOT
run:

- **`pools`** — gated on `IT_RIG_POOLS_PROBE`, set nowhere in the harness/scripts/CI, so it
  self-warns and skips every gate run.
- **auto-rollback** (`IT_RIG_ROLLBACK_CHANGES`) — same: set nowhere, skips.
- **`DONATION`** and the other writable keys (autotune, watchdog, watchdog_interval_min) — only
  `max_temp_c` and `pools` are exercised at all; `DONATION` appears only in a negative auth-
  rejection probe, never as a positive apply-and-verify. (Given we care that every rig runs
  DONATION=0, a positive apply-and-read-back of DONATION is worth having explicitly.)

Net: a change to the worker-apply path could break every writable key except `max_temp_c` and
the release gate would stay green.

## Fix

Make the miner-panel config-apply legs **run in the mandated release gate against the borrowed
loaner rig** (miner-1/2 per the reservation protocol), not as manual opt-ins:

- Provide the env the legs need from the harness (or synthesize safe in-gate probes) so
  `pools` and rollback stop self-skipping. `pools` is opaque passthrough, so derive a safe
  probe from the rig's own current pools (round-trip: read current → re-apply identical or a
  benign variant → assert `changed_keys` → restore) rather than a hand-set literal.
- Add a positive **DONATION** apply→verify-on-rig→revert leg (assert the rig's `/status`
  confirms the change, then restore to 0).
- Every leg restores the rig to its pre-test state (the existing legs already model this) — the
  loaner is a production miner on loan.
- Turn the self-skips into **hard failures when a rig is present** (skip only when no rig is
  attached at all), so a missing probe can't silently drop coverage in the gate.

## Acceptance

On a gate run with the borrowed rig attached, every writable key has an apply→verify-on-rig→
revert leg that runs (not skips) and fails the gate if the change doesn't land on the rig.
Depends on nothing but the loaner the gate already borrows.

[...]

Issue #1235, opened 2026-08-22
Worker Inspect editor shows empty boxes — surface each writable key's current value from the rig


VijitSingh97:
## What the operator sees

In the Worker Inspect config editor, every writable box — DONATION, autotune, max_temp_c,
pools, watchdog, watchdog_interval_min — is **empty**, under the note "Prefilled with the last
config applied from the dashboard — the rig's live feed doesn't expose these values." On a rig
never edited from the dashboard (History: "No changes applied from the dashboard yet"), there
is nothing to prefill, so the operator edits blind: no way to see the current value before
changing it, and an empty box invites overwriting a good value with a guess.

## Why (traced)

Prefill comes only from Pithead's own last-applied record (`workerview.mjs`: "Prefill comes
from Pithead's own last-applied record — the rig's enriched feed doesn't expose the writable
config values"). Two consequences: a never-edited rig has empty boxes, and even an edited one
only shows what *the dashboard* last set, not the rig's actual current state.

The premise is only half true today. Some writable values ARE already derivable from the
enriched feed the dashboard polls — `max_temp_c` is echoed as the watchdog Temp/max (the e2e
reads the original from there), and autotune/watchdog state appears in the detail rows. Others
— `pools`, the exact `DONATION` level, `watchdog_interval_min` — are exposed nowhere the
dashboard can read.

## Fix direction

The robust fix is authoritative, not piecemeal feed-scraping: **RigForge exposes the current
writable config over its control API** (a read — e.g. `GET /config` returning the writable-key
values, secrets masked as the existing `{"__secret__": true}` sentinel), and **Pithead prefills
the editor from that**, falling back to the last-applied record only when the rig read is
unavailable.

- Cross-repo: a rigforge issue for the read endpoint (bearer-gated like the rest of the control
  API; read-only; masks secrets), and this issue for the pithead consumer + prefill wiring.
- Show a clear per-field state when a value genuinely can't be read (rig offline, control off)
  rather than a bare empty box that reads as "0" / "none".
- Keep the write path unchanged — this only makes the *current* value visible before an edit.

## Acceptance

Opening the editor on a rig that has never been edited from the dashboard shows each writable
key's current value (secrets masked), read from the rig; the value shown matches what the rig
is actually running. Validate against the borrowed loaner rig the e2e already uses.

VijitSingh97:
[...]

Issue #513, opened 2026-07-14
Tier-4: exercise the Worker Inspect edit path end-to-end against a real rig (control ON)


VijitSingh97:
## Gap

The tier-4 `--rigforge` phase runs with **control OFF** and explicitly defers the edit path to "the hardening phase + the tier-2 test" (`tests/integration/run.sh:1550-1552`). But neither actually drives a live edit against a real rig with `dashboard.workers[]` populated. That hole is the direct reason **#508** (edit path dead — masked-token descriptor dropped) *and* the v1.5.2 regression (the read probe stringified the masked-token sentinel into the `Bearer` header → every `:8081` enriched probe 401'd) both shipped: no live test ever touched the masked-config edit surface.

## Proposal

Add a control-ON scenario to the `--rigforge` phase (fires when the harness sees `dashboard.control.enabled` + a rig with a healthy control API):

1. Enable `dashboard.control`, populate `dashboard.workers[]` for the borrowed rig (`{name, host, control_port, token}`) — the container then reads the **masked** descriptor (`{"__secret__": true}`), exercising the exact path both bugs lived in.
2. Assert `editable: true` for the rig (would have caught #508) **and** that enriched stats / `api_ok` still hold with descriptors present (would have caught the v1.5.2 regression).
3. Apply a **reversible, benign** writable change via the dashboard worker-apply (e.g. `max_temp_c` ±1, or `DONATION`), assert it lands on the rig's control `/status`, then revert it.

## Safety (it mutates a real mining rig)

Reserve the rig-e2e flock on the rig, back up its config, restore on the EXIT trap, and gate behind explicit real-rig detection so it never runs against a rig mid-tune.

One scenario, both regressions caught. Refs #185, #508, #440, rigforge#236.

PR #1373, opened 2026-08-24
test(integration): apply the writable keys the control phase never touched (#1236)

VijitSingh97:
## What this changes

The RigForge control phase proved the Worker Inspect write path on **one of six** writable keys. `max_temp_c` (#513) had a live readback in the enriched feed, so its round trip could be asserted. `pools` (#1002b) did not, so it was built on the dashboard's record of what it last pushed and gated behind an `IT_RIG_POOLS_PROBE` that nothing in the harness, the scripts, or CI ever sets. The other four keys had no leg at all.

A gate that applies one key in six is a gate on `max_temp_c`, not on "the writable path works".

**The premise that shaped it is no longer true.** RigForge v1.10.0 (rigforge#253) serves the rig's own effective writable config on the enriched feed, and the dashboard re-exposes it at `GET /api/worker` as `.rig_config` (#1235, `views.py:2115`). So a leg can read the original off the rig itself, derive a probe, assert the **rig** reports the new value, and restore it — no new environment variable, no direct rig dial, no new port.

The comment at `run.sh:2219-2222` asserting the feed "carries no writable-config VALUES at all" is false against a current rig, and it is the reason the pools leg was shaped the way it was. It is gone along with the code it justified. The same false claim also sits in `storage_service.py`'s `get_last_applied_worker_config` docstring — **that is the dashboard lane's path and I have not touched it**; flagging it here for that lane.

### Added: `DONATION` and `watchdog_interval_min`

- `DONATION` only ever moves **toward zero**, so the probe can only reduce what a borrowed rig gives away. A rig already at 0 is nudged to RigForge's own default of 1. Bounds 0–100 (`rigforge.sh:406`).
- `watchdog_interval_min` steps one minute away from wherever it sits, inside RigForge's 1–1440 range (`rigforge.sh:556`). The watchdog stays on; it just polls slightly less often.
- The restore runs **whatever the assertions said**. This drives a production miner on loan, so a mid-leg red must not strand it on a probe value.

### Refused on purpose — three of six, and the reasons are part of the deliverable

| key | why not |
|---|---|
| `max_temp_c` | already round-tripped by the #513 leg; a duplicate proves nothing |
| `autotune` | `disabled -> enabled` starts a **real tuning run** — it moves hashrate and thermals and may not settle inside the leg |
| `watchdog` | `enabled -> disabled` removes thermal protection from a rig mining at `max_temp_c=100`. Brief is still real on hardware we do not own |
| `pools` | **cannot be round-tripped from the rig's own reading at all** — see below |

**The pools finding.** RigForge deletes `pass` and `tls-fingerprint` before serving the config (`rigforge.sh:4818`, `_api_config_json`), and the dashboard's `_strip_credentials` deletes them again at any depth (`xmrig_client.py`). `pass` is load-bearing: it is the Pithead stratum password (#113), and a rig whose `pass` no longer matches is **rejected by the proxy at login**. So writing a self-read `pools` value back would silently strip the credential and strand a borrowed miner — and the harness cannot even tell "this rig has no password" from "this rig's password was stripped on the way to me", because both arrive as `{"url": …}`.

The operator-supplied route (`IT_RIG_POOLS_PROBE` + `.last_applied.pools` as the restore target, which is un-stripped) stays the only honest one for `pools`.

The refusals are stated **in the run output**, not only in source, and asserted as behaviour at tier 1 — so they cannot be undone in silence.

## Verification — what was actually run

- `selftest-rigforge-writable-keys.sh`: **50 assertions, all passing**, driven against stubs. No rig, no server, no docker.
- **Sixteen mutations of the module were run; all sixteen red.** Each assertion names the mutation it kills.
- **Four of them survived a first pass, and every one was a hole in the test, not the code.** Recording them because they are the useful half:
  1. `has($k)` → `// empty` survived. jq treats only `false` and `null` as falsy, so a measured `DONATION` of 0 passes through `//` unharmed — my comment claiming otherwise was **wrong** and is corrected in the source. The control now uses a `false` value, which does distinguish the two forms. A control that cannot fail is not a control.
  2. Deleting the "could not read" guard outright changed nothing observable — the per-key skips fell through one by one, equally silent. It is now asserted on the one sentence only it writes, plus the absence of the other guards' sentence.
  3. Replacing the readback predicate with `true` passed the entire file, because every test stubbed `wait_for` wholesale and none of them ever ran the predicate. The settle is now also driven with a one-shot `wait_for` that invokes the real predicate once.
  4. Mis-wiring the `max_temp_c` adapter to the #1236 predicate — a failure mode the shared settle newly makes possible — was uncaught until asserted directly.
- Both CI-shape selftest loops were proven to **fail on a real red selftest** (`exit 3` appended to one selftest → `make` rc=2 and non-zero CI loop), then restored.
- `shellcheck --severity=warning` + `shfmt -i 4 -d` clean on every file I touched.
- **`make lint` rc=0**, whole thing, on the pinned shellcheck 0.11.0, with the base ref re-checked immediately before the run and confirmed unmoved during it. `make test-integration-selftest` rc=0 (214 + 5 + 50 + 5 assertions).

### What I did NOT do

- **No hardware.** No rig was borrowed. Every live claim about the rig's behaviour in this PR is re-derived from RigForge and dashboard source, file:line, not measured this session.
- **`make lint` is green end to end** (`MAKE_RC=0`) on the rebased branch. Getting there cost a wrong finding of mine, recorded because the method error is more useful than the result: I first reported `lint-sh` as red on `origin/develop-v2` with an `SC2120` in `tests/stack/run.sh`. **That was wrong on two counts.** This box has two shellchecks — `/usr/bin/shellcheck` is 0.9.0, `~/.local/bin/shellcheck` is 0.11.0, and CI pins 0.11.0; SC2120 fires only on 0.9.0. And I linted the file alone, while `Makefile:73` deliberately pairs it with `os/overlay/pithead-boot`, which supplies globals it reads — linting it alone reds a clean tree by construction. The exact pair invocation is rc=1 on 0.9.0 and rc=0 on 0.11.0. Nothing in `tests/stack/` needs a patch or a suppression. The tell was in my own terminal an hour earlier — `lint-py` and `lint-yaml` had failed with `make: uv: No such file or directory`, the same missing `~/.local/bin` — and I fixed the symptom without asking what else that PATH was shadowing.
- **The file-budget gate reds on an un-rebased branch for reasons that have nothing to do with your diff.** Mid-run, a peer's fetch moved `origin/develop-v2` seven commits, and the dashboard lane had *lowered* two ceilings (`workerview.mjs` 479 → 443, `views.py` 2339 → 2263). My branch, still carrying the older `file-budget.tsv`, then failed as though it were *raising* them — naming two dashboard files this diff never touches. Rebasing cleared it with no change of mine. Flagging it because the RED points at the wrong person by construction.
- **This leg is not observable in the mandated release gate.** Per #1364, `--rigforge-control` is only added under `--mode matrix` (`e2e.sh:694`) while the runbook mandates `--mode targeted` (`docs/dev/releasing.md:184`). Merging this does not by itself make the gate cover these keys. **#1236 should not be closed on a green that proves nothing** — #1364 has to land first.

## Shared files touched, and why

Both were forced, and both are improvements rather than smuggled growth. Controller cleared them after checking contention (no other open PR on `develop-v2` touches either).

- **`.github/workflows/ci.yml`** and **`Makefile`** each enumerated the integration self-tests by hand, so a newly added one is silently never run — this project's own bug class, and `selftest-rigforge-writable-keys.sh` would have been invisible in CI. Both now glob `tests/integration/selftest*.sh`.
- `ci.yml` was itself sitting exactly at its 513-line ceiling, so the enumerated one-line addition failed `lint-file-budget`. The glob is **line-negative** (512), which is why it was the right answer rather than a carrier for the growth.

## File budget

`tests/integration/run.sh` went 2551 → 2526 (the stale pools block became one call into the new module). **Its recorded ceiling is deliberately left at 2551** — 25 lines of reclaimed slack declined on purpose. `check_monotonic` compares the whole budget file against the base, so lowering any row reds every in-flight branch across every lane that has not rebased, and the RED names a file the other lane's diff never touched. That is the #1355 shape that cost two lanes real time. Endorsed as a ruling by the controller.


VijitSingh97:
## Reviewer: MERGE, with one note about a guarantee that is narrower than the comment says.

This is the PR in this batch where being wrong costs the most — it writes real config to a borrowed
production miner — so the review weight went here, and specifically onto the claims about RigForge,
because those are claims about an external system and cannot be checked against our own comments.
Read against `rigforge.sh` at rigforge `develop` (re-derive the tip with
`git ls-remote origin develop`).

### The safety claim is TRUE, and I ran it down rather than reading it

`_writable_key_round_trip`'s comment says *"The restore runs whatever the assertions said."* That
rests entirely on the assertions not aborting, so:

- `tests/integration/run.sh:20` is `set -uo pipefail` — **deliberately not `-e`**.
- `assert_eq` / `assert_contains` in `lib.sh` dispatch to `it_fail`, which increments a counter,
  appends a name and prints. No `exit`, no abort.

So a red on any of the three assertions between the apply and the revert cannot skip the revert.
Confirmed by construction.

### The note: the guarantee covers assertion failures, and nothing else

There is **no `EXIT` trap that restores rig config**. The only `EXIT` trap in `lib.sh` removes the
rig-lock holder file. So the revert is skipped by anything that is not an assertion failure —
`set -u` is on, and an unbound variable between the apply and the revert aborts the script outright;
so does a SIGINT, a CI cancellation, or the `exit 1` / `exit 75` paths in the lock helpers if they
were ever reached from inside a leg.

**Today this is harmless, and it is harmless for a reason that is not the restore.** It is harmless
because the probes were deliberately chosen benign — `DONATION` moves toward zero, and
`watchdog_interval_min` moves by one minute inside RigForge's range. A rig stranded on either is
fine. That is good design and the PR argues for it explicitly.

The risk is the wording, not the code. The comment attributes the safety to the restore, and the
same file permanently refuses `autotune` and `watchdog` **because their probes are not benign**. A
future author lifting one of those refusals reads "the restore runs whatever the assertions said"
as a guarantee that the module cannot strand a rig — and for a non-benign probe, an interrupt or a
`set -u` abort would do exactly that. Suggest the comment say what actually carries the safety:
the probe is chosen so that being stranded on it is harmless, *and* the restore runs past a red.
Non-blocking; no code change needed for the keys this PR actually drives.

### Upstream claims, checked

- **`pools` refusal — correct, and this is the strongest finding in the PR.** `_api_config_json` is
  literally `_writable_config_canonical | jq -c '.pools = [.pools[]? | del(.pass, ."tls-fingerprint")]'`.
  The credential is deleted before serving, exactly as claimed, so a self-read `pools` written back
  would strip it. The reasoning holds against source.
- **`DONATION` bounds 0-100 — correct.** `[ "$DONATION" -gt 100 ]` guards, error text says 0-100.
- **`watchdog_interval_min` 1-1440 — correct.** Guarded on both ends.
- **Both keys are served back on the feed**, so the readback the legs assert on is genuinely
  available — that is what makes this leg possible at all, and it checks out.

**Three line citations are stale**, substance unaffected: the body and the module cite
`rigforge.sh:406` for the `DONATION` bounds, `:556` for `watchdog_interval_min`, and `:4818` for
`_api_config_json`. At rigforge `develop` those are at **431**, **586** and **5038**. Each is off by
25-30 lines in the same direction, so this reads as citations taken against a slightly older tip
rather than as wrong readings. Worth fixing only because the module presents them as the
verification route for the next reader.

### Agreed and not re-run

The sixteen mutations, the four that survived a first pass, and the CI-shape selftest loops are your
measurements, not mine — I did not re-run them. Recording the four survivors, and that every one was
a hole in the test rather than in the code, is the useful half of that section; the `has($k)` one in
particular (jq treating only `false` and `null` as falsy, so a control built on `0` cannot fail) is
the same class this fleet keeps finding.

I also agree with holding `tests/integration/run.sh`'s ceiling at 2551 rather than banking the 25
reclaimed lines: lowering a row reds every in-flight branch that has not rebased, and the RED names
a file the other lane's diff never touched.

### What I did not do

No hardware — same as the PR. Every claim above is read from source or run against stubs. I did not
run `make lint`, `make lint-sh` or the integration selftests. `docs/dev/testing-strategy.md` is also
edited by open PR #1363, but in a different region of the file (its inventory section, around line
226; yours is the coverage matrix around line 152) and the file carries no budget row, so the two
merge cleanly in either order.

Issue #1098, opened 2026-08-18
Appliance compose names an xmrig-proxy healthcheck script its pinned images predate, so the container is permanently unhealthy


VijitSingh97:
On a Pithead OS appliance running `:v1.18.1` images, `xmrig-proxy` reports **unhealthy from
boot onward**, while the proxy itself is working normally.

## What is wrong

The container's healthcheck is:

```
{"Test":["CMD","/usr/local/bin/xmrig-proxy-healthcheck.sh"], "Interval":30s, "Retries":3}
```

and that file does not exist inside the container:

```
$ podman exec xmrig-proxy cat /usr/local/bin/xmrig-proxy-healthcheck.sh
cat: /usr/local/bin/xmrig-proxy-healthcheck.sh: No such file or directory
```

So the check exits 1 with empty output every 30s, forever — the health log is a run of
identical `{"ExitCode":1,"Output":""}` entries. Meanwhile the proxy answers its API
normally and is listening on the stratum port, so nothing is actually broken.

## Root cause: compose and the image pins move independently

Both the script and the healthcheck arrived together in #904 (`ed461bd`):

```
build/xmrig-proxy/Dockerfile:29:  COPY healthcheck.sh /usr/local/bin/xmrig-proxy-healthcheck.sh
docker-compose.yml:550:           test: ["CMD", "/usr/local/bin/xmrig-proxy-healthcheck.sh"]
```

That commit landed **after** v1.18.1. At the tag itself, neither exists:

```
$ git show v1.18.1:build/xmrig-proxy/Dockerfile | grep healthcheck            -> nothing
$ git show v1.18.1:docker-compose.yml | grep xmrig-proxy-healthcheck          -> nothing
```

The appliance runs containers pinned to `:v1.18.1` alongside a `docker-compose.yml` baked
from the development tree, which carries #904. The compose file therefore names a script
the pinned image predates. A host where compose and images are the same vintage (v1.19.1)
has the script present and reports healthy — same service, same check, opposite verdict.

## Why it is worth fixing rather than waiting for the next release

The skew is structural, not a one-off. Any compose change that depends on new *image*
content breaks the appliance for the whole window between that merge and the next image
release, and the symptom is a permanently-unhealthy container rather than an error anyone
is shown.

It is also a poor service to have this on. `pithead-boot`'s commit gate does not read
container health directly — it gates on the dashboard's HTTP code plus `pithead doctor
[...]

PR #1376, opened 2026-08-24
build(tor): stop installing vim's xxd over the one busybox already ships (#1372)

VijitSingh97:
Removes `xxd` from the tor image's `apk add`. One line of behaviour change; `healthcheck.sh` is
untouched.

## Why

The weekly shipped-image sweep found four fixable HIGH CVEs in the published v1.20.0 `pithead-tor`
digest on its first live run, all in the vim source package that provides `xxd`.

**That package was never needed.** `build/tor/healthcheck.sh:22` hex-encodes Tor's control auth cookie
with `xxd -p -c 256`, and Alpine's busybox already ships an `xxd` applet that handles exactly those
flags. `apk add xxd` installed vim's build of the same tool on top of one already on `PATH`.

**Deleting rather than bumping is the point.** All four CVEs are in vim's source package, so a version
bump buys a quiet week and the same package produces the next four.

## What was verified

All inside the pinned digest `alpine:3.24@sha256:28bd5fe8…`, not on the host — busybox `od` and GNU
`od` are different programs, so a host-only check would have proven the wrong thing.

| check | result |
| --- | --- |
| both applets exist | `busybox --list` reports one `xxd`, one `od` |
| identical output | busybox `xxd -p -c 256` ≡ `od -An -tx1 -v` byte for byte, random 32-byte input |
| package gone | `apk info` reports no vim and no xxd package |
| still working | `/usr/bin/xxd` → `/bin/busybox` (owned by `busybox-1.37.0-r31`); the healthcheck's own command still returns 64 lowercase hex chars |
| script unchanged | `tor-healthcheck.sh` still exits 1 with no cookie — same early-exit path |
| trivy on the rebuild | 0 fixable HIGH/CRITICAL |

**The last row is the weakest and is not the proof.** A rebuild resolves apk at build time, and
`apk add xxd` against this same digest installs the **fixed** `9.2.0854-r0` today — so a rebuild that
still installed xxd would also scan clean. The discriminating evidence is that the package is absent.
That asymmetry is exactly what the shipped-image sweep exists to expose, and it is why these CVEs were
visible at all.

## The negative control earned its keep

This change is one line instead of two files because the control caught me. After removing `xxd` from
`apk add` I checked it was gone, and got a contradiction:

```
$ command -v xxd     -> /usr/bin/xxd      (still there)
$ apk info -e xxd    -> (not installed)   (but the package is not)
```

Had the check only asserted the package was uninstalled, that reads as success and I would have shipped
a claim that the tool was gone when it was not. Resolving the contradiction found the busybox applet
and made the `healthcheck.sh` rewrite I had drafted unnecessary. I had already posted the `od` rewrite
as a recommendation; it is corrected on the issue rather than quietly dropped.

## Reachability, since "four HIGHs" is not by itself a severity

`xxd` runs on the healthcheck interval — `docker-compose.yml` wires `tor-healthcheck.sh` at
`interval: 30s`, so roughly every 30 seconds for the life of the container. Not "never executed".
But what it parses is Tor's own 32-byte control auth cookie: not network input, not operator input.
Influencing it means already having code execution in the container as the `tor` user. So: constantly
executed, on trusted local input of fixed size. Worth fixing, and it did not read as forcing a patch
cut — that call sits with the release owner.

## Owed, and disclosed rather than skipped

**A regression test is not in this PR, and this PR should not merge without it.** The tests lane owns
`tests/stack/` and has granted an additive append to `tests/stack/test-tor-network.sh` — the tor domain
file, already split out and already sourced. Naming the path here so the ownership is visible.

**What that test can and cannot honestly prove — this changed after review, and the first shape was
wrong.** I initially asked for an assertion that the encoding works *without the vim `xxd` package*.
That is not provable in the stack suite: it runs on the host, the host is Ubuntu, and on Ubuntu `xxd`
**is** vim's xxd. A host-side test would return 64 hex characters whether or not the image ships the
package, because it never consults the image — a false green of exactly the kind this change exists to
remove.

What the suite *can* guard is the direction that actually regresses: **`healthcheck.sh` growing a
dependency the image does not ship.** Run the script with `PATH` set to a stub directory holding only
the commands the tor image provides, and assert the exit status and the literal 64-character lowercase
hex output for a fixed cookie. The control is the point — remove `xxd` from the stub set and the test
must FAIL; if it still passes, `PATH` is leaking to the host and the assertion is decorative. That
reddens the day someone adds a `base64` or a `sed -E` the image lacks, which is the realistic
regression during a healthcheck edit.

**NAMED GAP, not approximated.** The image-level fact — that this pinned digest's busybox provides an
`xxd` applet accepting `-p -c 256` — is a property of the base image, and the pin is what protects it.
It is proven in the table above, inside the digest, cross-checked against `od`. It is **not**
mechanically guarded anywhere, and it does not belong in the shell suite at any tier; the place for it
would be wherever the base-image pin is bumped. No such check exists today. Stating that as a gap
rather than letting a host-side test stand in for it.

Also not done: no live Tor bootstrap was exercised. The healthcheck's `nc` leg is unchanged by this PR
and was not re-proven.

## Scope

`build/` only — this lane's exclusive path. Nothing shared, no override used.

---

## The test half (added 2026-08-24)

The one-line fix above was correct and unguarded. Nothing in CI could see either the need for `xxd`
or its removal, so merging on a green rollup would have landed the fix carrying the exact gap it
exists to close — **green there meant "no test noticed", not "checked"**.

`build/tor/healthcheck-selftest.sh --self-test` drives `build/tor/healthcheck.sh` for real with
`PATH` stripped to a stub dir holding **only** the commands the tor image provides, plus a stub
control port that records what the script sends. It is an **allowlist**: a dependency nobody declared
fails here, rather than passing because nobody thought to ban it. Twelve checks, all passing.

### What makes it evidence rather than decoration

Each declared command is dropped in turn and the run must FAIL. Without that control a leaking `PATH`
would let every case pass on the host's own commands and prove nothing about the image. The
over-tightening control is the mirror: an unused extra command on `PATH` must change nothing, or the
checks are really counting the allowlist.

### Mutation results — five run, four caught

| mutation of the shipped script | caught by |
| --- | --- |
| drop the `tr -d '\n'` strip | the `tr`-missing control (it stops being needed) |
| drop the `[ -n "$COOKIE_HEX" ]` guard | empty-cookie check, and the `xxd`-missing control |
| accept any bootstrap phase, not only `TAG=done` | mid-bootstrap check |
| repoint the container default cookie path | both seam checks |
| **drop `[ -r "$COOKIE_FILE" ] \|\| exit 1`** | **NOTHING — survives, see below** |

Deleting the self-test script outright reddens the suite rather than reading as a pass; that was
checked specifically, because a missing leg scoring as a pass is the failure mode this whole thread
is about.

### Two gaps, named rather than papered over

**The surviving mutation is real and is shipped documented.** Removing the unreadable-cookie guard
changes nothing observable, because `xxd -p -c 256 "$f" | tr -d '\n'` reports *tr's* exit status — a
failed read is not an error the shell ever sees — and the empty-hex guard below enforces the same
outcome. Two guards, one output. No discriminating assertion was found that is not flaky across `xxd`
implementations, so it is written into the code as a known gap.

**This does not prove the image's `xxd` behaves.** The host's is vim's — the very package removed
here. That the pinned digest's busybox provides an applet taking the same `-p -c 256` was proven
inside the image (table above) and is guarded by nothing but the base-image pin. What is proven here
is the script's dependency set and its logic. Proving the image-level fact belongs wherever the base
pin is bumped, and no such check exists today.

### Shared files, disclosed

- `tests/stack/test-harness-tooling.sh` — the append-only harness registry in `_shared.paths`. It
  gets the two-line registration its grant describes, matching the six entries already there. The
  logic lives in `build/`, which this lane owns outright, beside the script it tests and never
  `COPY`'d into the image.
- `build/tor/healthcheck.sh` grows one seam, `TOR_COOKIE_FILE`, in the same shape as
  `entrypoint.sh`'s `TORRC_OUT` and for the same reason: the suite cannot write `/var/lib/tor`. It is
  line-neutral, and the shipped default is asserted unchanged by two of the twelve checks.

The tests lane's original ruling placed this in `tests/stack/test-tor-network.sh`. That file is 780
lines against a recorded ceiling of exactly 780 and cannot take a line; the registry is where the
controller's 2026-08-24 `_shared.paths` ruling puts an additive harness self-test, and on merit it is
the sibling of the `#1098` entry directly above it — that one asks whether a healthcheck script
*exists* where its Dockerfile promises, this one asks whether it can still *run* there.

### Gates run

`make lint` rc=0 (full target list, not a filtered tail). `tests/stack/run.sh`: **2906 passed, 0
failed**, rc=0, rebased onto the current base. An over-engineering pass was run on this diff
regardless of whether the PR gate prompted for one; it is what moved the logic out of the shared file.

### DIFF
--- a/build/tor/Dockerfile
+++ b/build/tor/Dockerfile
@@ -2,9 +2,10 @@
 # names the Alpine minor line (3.24) so Dependabot's patch-update policy has a real version to track.
 FROM alpine:3.24@sha256:28bd5fe8b56d1bd048e5babf5b10710ebe0bae67db86916198a6eec434943f8b
 
-# Install Tor plus the tools the bootstrap healthcheck needs: netcat (talk to the
-# control port) and xxd (hex-encode the auth cookie). --no-cache keeps the image small.
-RUN apk add --no-cache tor netcat-openbsd xxd
+# Install Tor plus netcat, which the bootstrap healthcheck needs to talk to the control port.
+# Deliberately NOT xxd: the healthcheck's `xxd -p -c 256` is served by busybox's own applet, so
+# `apk add xxd` bought nothing but vim's CVE stream (#1372). --no-cache keeps the image small.
+RUN apk add --no-cache tor netcat-openbsd
 
 # Copy the Tor config TEMPLATE + the entrypoint that renders it: the bridge subnet prefix is
 # substituted at container start (#180). The rendered torrc lands in /tmp (the unprivileged 'tor'
--- a/build/tor/healthcheck-selftest.sh
+++ b/build/tor/healthcheck-selftest.sh
@@ -1,0 +1,142 @@
+#!/usr/bin/env bash
+#
+# Does build/tor/healthcheck.sh still run on nothing but the commands its own image ships?
+#
+# The #1098 guard (scripts/verify-healthcheck-scripts.sh) asks whether a healthcheck script EXISTS
+# where its Dockerfile promises. This asks the other half of the same contract: whether that script
+# can still RUN there. #1372 is the case that made the gap visible — build/tor/Dockerfile installed
+# `xxd` by name purely for healthcheck.sh's one call site, which bought vim's CVE stream for a
+# command Alpine's busybox already provides. Nothing in CI could see either the need or its removal.
+#
+# It is an ALLOWLIST, not a denylist: a dependency nobody declared fails here, rather than passing
+# because nobody thought to ban it. Widening HC_CMDS is the point at which someone has to check the
+# image really ships the new command.
+#
+# NAMED GAP, deliberately not approximated: this cannot prove the IMAGE's `xxd` behaves — the host's
+# is vim's, the very package #1372 removed. That the pinned digest's busybox provides an `xxd` applet
+# taking the same `-p -c 256` was proven inside the image on #1372, and is guarded by nothing but the
+# base-image pin. What IS proven here is the script's dependency set and its logic.
+#
+# It lives beside the script it tests, like entrypoint.sh and healthcheck.sh live beside the
+# Dockerfile, and is never COPY'd into the image — the Dockerfile copies named files only. It takes
+# `--self-test` with no other mode, matching every other entry in the harness registry that drives
+# it (tests/stack/test-harness-tooling.sh).
+#
+#   build/tor/healthcheck-selftest.sh --self-test
+
+set -uo pipefail
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+HC="$HERE/healthcheck.sh"
+
+HC_CMDS=(xxd tr grep nc) # every external command build/tor/healthcheck.sh may use
+HC_HEX="000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f"
+HC_DONE='250-status/bootstrap-phase=NOTICE BOOTSTRAP PROGRESS=100 TAG=done SUMMARY="Done"'
+HC_MID='250-status/bootstrap-phase=NOTICE BOOTSTRAP PROGRESS=50 TAG=loading_descriptors'
+
+st_fail=0
+expect() { # <name> <got> <want>
+    if [ "$2" = "$3" ]; then
+        printf '  self-test ok: %s\n' "$1"
+    else
+        printf '  self-test FAIL: %s — expected [%s], got [%s]\n' "$1" "$3" "$2" >&2
+        st_fail=1
+    fi
+}
+
+hc_run() { # <cookie: none|empty|fixed> <control-port reply> <allowed cmd...> -> healthy|unhealthy
+    local cookie="$1" reply="$2" c i d oct="" rc=0
+    shift 2
+    d="$(mktemp -d "$HC_D/case.XXXXXX")"
+    mkdir -p "$d/bin"
+    # Truncate the capture every call: a stale one read as fresh is the vacuous version of the
+    # request assertion, and hc_run runs in a command substitution so it cannot hand back a path.
+    : >"$HC_D/sent"
+    for c in "$@"; do
+        if [ "$c" = "nc" ]; then
+            # Stub control port: record what the healthcheck sends, then answer as Tor would. A real
+            # `nc` would dial the fixed port 9051 on the host — untestable, and a collision between
+            # two concurrent suites. Shell builtins only: it runs under the same stripped PATH.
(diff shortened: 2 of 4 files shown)


VijitSingh97:
## Reviewer: MERGE. Scope is exactly one call site, confirmed repo-wide — and there is a second `xxd` install this does not touch.

The risk in "delete a package" is a consumer nobody enumerated, so that is what I checked rather
than re-running your digest work.

**`xxd` has exactly one consumer in this repo**, and it is the one you name:

```
build/tor/healthcheck.sh:22:  COOKIE_HEX=$(xxd -p -c 256 "$COOKIE_FILE" | tr -d '\n')
```

Repo-wide grep, excluding prose. Nothing in `entrypoint.sh` or `torrc.template` uses it, and those
plus `healthcheck.sh` are the only files `COPY`d into this image. So the blast radius really is the
single `-p -c 256` invocation you verified against the busybox applet inside the pinned digest —
there is no second call site with flags busybox might not support.

**Agreed on which row is the proof.** You are right that the clean Trivy rebuild is the weakest
evidence and that package absence is the discriminating fact — `apk add xxd` against this same
digest resolves to the fixed build today, so a rebuild that still installed it would scan clean too.
Stating that asymmetry rather than leading with the green scan is the difference between evidence
and a tick. Same for deleting rather than bumping: the CVEs are in vim's source package, so a bump
buys a week.

**Verifying inside the pinned digest rather than on the host was the right call** and not a
formality — busybox and GNU builds of these tools genuinely differ, and a host check would have
proven something about this box instead of about the image.

### Out of scope here, but found while checking scope — for the appliance lane, not for you

`os/rootfs/Dockerfile:143` also installs `xxd`, via `apt-get`, into the `debian:trixie-slim`
appliance rootfs. **Nothing in the repo consumes it** — the same repo-wide grep that found one
consumer for the tor image found zero for that one. So it looks like the same unused dependency by
a different route.

Two reasons not to fold it in here: it is a different base (Debian, where `xxd` is its own package
and there is no busybox applet to fall back on, so "just delete it" needs its own check), and
`os/` is the appliance lane's path, not currency's. Flagging rather than acting.

Honest limit on that grep: it covers files in this repo. Something rendered or fetched at runtime
inside the appliance could still call `xxd`, and I did not rule that out.

### What I did not do

I did not rebuild the image or re-run any of your six checks — no docker work from this lane. The
busybox-applet equivalence, the `apk info` absence, and the healthcheck's live output are your
measurements. I checked the one thing that would have made them insufficient: another consumer.

Issue #1342, opened 2026-08-24
pithead has no mutual exclusion: concurrent verbs can mutate config.json under each other


VijitSingh97:
## What is missing

`pithead` has no mutual exclusion of any kind. There is no `flock`, no lock file, and no
advisory marker anywhere in the CLI — `grep -n "flock\|lockfile" pithead` returns nothing
relevant. Every verb assumes it is the only `pithead` running on the box.

On a coordinator that assumption is false by construction. Three writers can be in flight at
once, none of them aware of the others:

- the **firstboot wizard loop**, which runs `setup` in the foreground and, if it fails, does
  `mv -f config.json config.json.failed` (`pithead:2752`) and re-mints a token;
- the **host-side runner that consumes the dashboard's spooled intents**, applying a staged
  config on the operator's click;
- an **operator verb** — `backup`, `apply`, `up`, `down`, `upgrade` — run from a shell or a
  test harness.

**Correction (2026-08-23).** An earlier version of this body named the dashboard control
channel itself as one of the mutators. That is wrong, and the difference matters. The container
cannot write the real `config.json` at all: `config.py:247` points it at
`/control/masked/config.json`, a masked read-only copy, and `docker-compose.yml:629` states the
raw file is never mounted. Its only `config.json` write is into the spool
(`wizard.py:347`), and it is atomic (`tempfile.mkstemp` + `os.replace`), as is the control
channel's `{rid}.json` request write (`control_service.py:322-325`). So no reader can observe a
partial file from the container side, and the container is not a racer.

The consequence is that this race, if real, is **entirely host-side** — every actor that can
mutate `config.json` lives in the root `pithead` script. Any fix belongs there and only there.
Adding a lock on the dashboard side would put a second scheme, in a second language, guarding a
file it cannot write.

## Why it matters, concretely

`stack_backup` reads its required items, then stops the stack, then archives:

```
pithead:3588    backup_require_items "${items[@]}"     # config.json present here
pithead:3665    stack_down                             # ~80 lines later
pithead:3684    sudo tar -czf - -C "/" "${rel[@]}"     # config.json read here
```

Anything that mutates `config.json` inside that window turns a checked precondition into a
failed archive **after** the stack is already down — real downtime, no backup, and the
`#1244`/`#1248` required-item guard does not fire because it already passed. That guard
protects the steady-state missing file; it does not and cannot protect this window.

The same reasoning applies in the other direction: a `down` issued while `setup` is still
[...]

PR #1135, opened 2026-08-19
fix(os): reboot a slot that fails its health gate, once (#1065)

VijitSingh97:
The A/B design's headline promise is that a bad update reverts itself. For the likeliest bad update
it did not.

`pithead-boot` left the slot uncommitted and exited. The fallback is a **GRUB** decision, and GRUB
does not get to make it until something reboots — and nothing did. `pithead-boot.service` is
`Type=oneshot` with no `Restart=` and no `OnFailure=`. Two failure shapes already self-heal, a kernel
panic (`panic=60`) and a wedged PID 1 (`RuntimeWatchdogSec`), but neither covers the regression an OS
update actually ships: a **clean boot with a dead stack** — a bad render, a bad image, an unparseable
Caddyfile, a crashed `monerod`. That is precisely what the health gate exists to catch, and catching
it stranded the machine with the stack down until somebody walked over and pulled the power.

Remote recovery cannot help, by construction: the dashboard that would drive it is the thing that is
down. `docs/appliance.md` and `docs/dashboard.md` both told the operator otherwise.

## What changes

`pithead-boot` reboots itself when the gate fails, so the bootloader gets its turn.

**Bounded to one attempt, counted on `/data`.** The counter has to survive both the reboot and the
slot switch, because that is the case it exists for: a fault that lives on `/data` survives the
fallback, both slots then fail the same way, and a box that reboot-loops can never be diagnosed — it
never holds still long enough to be looked at. The second failure stops and says the fault is not the
slot.

**If the counter cannot be written it does not reboot at all.** An unbounded reboot loop is the one
outcome worse than a stranded box, so the failure to persist fails safe rather than open.

**Every exit that leaves the slot uncommitted goes through the helper**, not only the health gate: a
failed `render`, a failed `up`, and both of the rig leg's exits strand the machine in exactly the
same way. A grep assertion keeps a bare `|| exit 1` from creeping back onto any of them.

The helpers sit above a source guard so the suite can drive them in a sandbox — the same shape
`pithead-data-reset` already uses. That is the only structural change to the script; `cd /data/pithead`
moved below the guard so a `source` does not chdir.

## Bench evidence

**The normal path still works, on a real machine.** `tests/os/run.sh --phase update` on the KVM bench
against this branch: **18 passed, 1 failed**. Legs 1–3 are the whole of what `pithead-boot` owns and
all passed —

```
✓ ROLLBACK: an uncommitted update reverts to v1 on reboot
✓ committed the booted update
[...]

PR #1381, opened 2026-08-24
fix(cli): a failed setup keeps the machine's configuration (#1059 Defect B)

VijitSingh97:
Defect B of #1059 — the product half. The root cause (a concurrent `backup` calling
`stack_down` under setup's in-flight `compose up`) was established on the bench and is
recorded on the issue; the structural gap that allows it is #1342 and lands separately.

## What was wrong

```
pithead:2752   mv -f "$PWD/config.json" "$PWD/config.json.failed" 2>/dev/null || rm -f "$PWD/config.json"
```

**The `|| rm -f` destroys the operator's configuration.** The `mv`'s stderr is discarded, so
whatever made it fail is unknown, and the response to an unknown failure is to delete the file
outright. It also destroys the evidence: a missing `config.json.failed` can never be read as
"this branch did not fire", because this is the branch that leaves no trace.

**The `mv` itself buys nothing, on every path that reaches it.** Moving the file aside exists to
re-arm the setup wizard, and that needs BOTH of `pithead-firstboot.service`'s ANDed conditions:

```
os/overlay/pithead-firstboot.service:9    ConditionPathExists=!/data/pithead/config.json
os/overlay/pithead-firstboot.service:10   ConditionPathExists=!/data/pithead/machine-role
```

`record_machine_role` runs unconditionally at `pithead:2734`, eighteen lines above. So the second
condition already holds the window shut and the branch pays its whole cost — discarding a config
valid enough to have rendered `.env`, provisioned Tor and started containers — for none of its
benefit. Keeping the file also keeps the next boot working: `pithead-boot.service`'s conditions
are ORed so it runs regardless, and `os/overlay/pithead-boot:225`'s `./pithead render` reads
exactly the file the move used to take away.

A non-zero `(setup)` is not evidence that the submitted config is at fault, and this change stops
treating it as though it were.

## What it is now

`wizard_keep_failed_config` — an `install -m 600` copy (the file holds the dashboard password and
the wallet address), with the original removed **only** where removal can still re-arm the
wizard: a `machine-role` that never landed, `record_machine_role` being best-effort. The removal
lives INSIDE the copy's success branch, so nothing here can delete the file without a copy
already on disk — the guarantee is structural rather than a paired condition someone can later
separate.

Two smaller things fixed in the same branch, both consequences of the first:

- `last-attempt.json` now prefills from what THIS run failed on, never from a
  `config.json.failed` left by an earlier attempt on the same boot.
- `[ -f x ] && jq ...` as a bare list is fatal under `set -Eeuo pipefail` when the test fails.
  It was previously masked by a `mv` that essentially always succeeded.

## What was RUN

- **12 tier-1 assertions**, `tests/stack/test-wizard-setup.sh`. All pass.
- **Every guard mutation-checked.** Restoring the original line, dropping the `machine-role`
  guard, inverting it, re-adding `|| rm -f` outside the success branch, widening the copy's mode,
  deleting the warning, and reporting success on a failed copy — each kills its own case and
  nothing else. Restoring the original line kills six.
- **The failure fixture was chosen against `install(1)` empirically, not reasoned about.** My
  first two ideas were both wrong: a destination that is a directory and a destination that is an
  existing `0400` file BOTH succeed. An unwritable parent does fail — but it would also defeat the
  `rm`, so the case would have passed for a reason unrelated to the guard. An unreadable source is
  the one that discriminates, and it is skipped with a stated reason under `uid 0`, where mode
  `000` does not deny root.
- `install -m 600`'s availability on the appliance is not an assumption: `pithead:2319` already
  uses it on the firstboot path itself.
- **Full tier-1 suite: 2917 passed, 0 failed, rc 0** on the branch tip. All 12 new assertions
  confirmed present in that run BY NAME, not inferred from the total.
- `shellcheck --severity=warning` rc 0 on all three changed shell files under the CI-pinned
  0.11.0 (`command -v` checked first — it resolves to `~/.local/bin` here, not the stale 0.9.0).
  `shfmt -i 4 -d` rc 0. Cheap lint targets each rc 0, captured individually rather than off a
  pipe: `lint-file-budget`, `lint-topology`, `lint-operator-strings`, `lint-docs-voice`,
  `lint-md`, `lint-yaml`, `lint-toml`.

## Shared files, disclosed

- **`pithead`** — the fix itself. Override pre-granted by the controller.
- **`tests/stack/test-wizard-setup.sh`** — additive append, granted by the `tests` lane for this
  specific change. Placement is deliberate: `run.sh` cannot take an addition (ceiling equals its
  line count), and this file has headroom at 303 of 800. It ends at 384, under the 400 target, so
  it still takes no budget row.
- **`Makefile`** — `lint-sh` enumerated three of the six files in `tests/os/`, so THREE were
  never shellchecked by any gate: `failure-evidence.sh` (shipped in #1358),
  `reinstall-prefill-verdict.sh` and `restore-live-state-verdict.sh`. `shfmt` did cover all of
  them (it walks `git ls-files '*.sh'`), which is what made the hole invisible. An enumerated
  list silently drops whatever is added beside it, so the three paths are replaced with
  `tests/os/*.sh` rather than adding back the one file I happened to find — adding only mine
  would have left two holes behind a line that now reads as audited. Same shape and same fix the
  e2e lane applied to `ci.yml`. All three are clean at `--severity=warning`, so this adds a check
  rather than a fix. Verified by EXPANDING the glob (six files, zero duplicates,
  `tests/stack/run.sh` still isolated), not by reading the line.
- `os/KNOWN-ISSUES.md` and `tests/os/failure-evidence.sh` are this lane's own.

## NOT proven — do not let the greens imply otherwise

- **No bench run. This is not hardware evidence.** The behaviour is exercised at tier 1 against
  the real function in a sandbox; that a provisioned appliance reaches this branch and keeps its
  config across a reboot is unmeasured.
- **That today's behaviour lands the next boot in `fail_boot` is an INFERENCE**, from the four
  file:line facts above, not an observation. The fix does not depend on it — it is the reason the
  trade is worth making, not the reason the branch is wrong.
- **A genuinely unprovisionable config is now retried on the next boot** instead of being
  discarded. That is a deliberate trade: a repeated loud failure with the operator's configuration
  intact beats a single failure with it gone. `pithead-boot` bounds its health-gate reboots on
  `/data` (#1065, PR #1135), so it is not an unbounded loop — but it is a behaviour change and it
  should be seen rather than discovered.
- **One likely side effect I have reasoned about but not measured:** pre-fix, a resubmitted config
  was `cp`'d onto a path that no longer existed, so it was created under the process umask (`644`
  under systemd's default). Post-fix it lands on an existing `600` file, and `cp` preserves an
  existing destination's mode. If that is right the change tightens the mode rather than loosening
  it, which is the safe direction — but I have not run it, and it is not what this PR is for.
- `make lint-sh` WAS run, under the lint lock, after the glob change: **rc 0, zero findings** —
  both halves of the target, including the isolated `tests/stack/run.sh` invocation. The rc was
  captured from make itself, not read off a filtered tail.
## Over-engineering pass

Run by hand on this diff. `/ponytail-review` is not available as a command in this environment,
so this is the pass itself rather than a claim that a tool produced it. Two findings, both
resolved as KEEP with the reasoning stated so a reviewer can overrule me:

1. **Extracting `wizard_keep_failed_config` for a single call site.** Justified by testability
   and nothing else: `run_sourced` needs a named function, and without the extraction the only
   way to reach this behaviour is to drive the whole wizard loop, which is tier 3 at best. The
   standing bar asks for the lowest tier that proves it honestly, and this is what makes tier 1
   possible.
2. **The conditional removal, rather than deleting the removal outright.** Deleting it would be
   simpler and would cost one code path and two test cases. I kept it because dropping it would
   change behaviour in the one case where the existing code is CORRECT — a `machine-role` that
   never landed, `record_machine_role` being best-effort — and I have no evidence that case
   cannot happen. It is one line. If a reviewer judges the case unreachable, this is the right
   thing to cut.

Also considered and found minimal: the `kept_copy` flag (its value is consumed later, so the rc
cannot just be tested inline), the two-branch prefill (the re-arming path has no `config.json`
left, so both branches are reachable), and the comment-to-code ratio on the new function (it
matches the block it replaces and the surrounding file, and it records why the simpler
delete-the-branch option was not taken).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ lint-sh: ## shellcheck + shfmt over the CLI, build/* + dashboard/ container scri
 		os/overlay/pithead-data-reset os/overlay/pithead-mount-generator os/overlay/pithead-ssh-host-keys \
 		os/overlay/pithead-machine-id os/overlay/pithead-media-config os/overlay/pithead-hugepages \
 		os/overlay/pithead-journal-persist \
-		tests/os/run.sh tests/os/verify-image.sh tests/os/hugepages-boot-verdict.sh
+		tests/os/*.sh
 # run.sh by itself, keeping company only with the one file outside tests/stack that it sources:
 # os/overlay/pithead-boot, whose gate_ready and os_update_rollback_verdict read three globals
 # run.sh sets for them. Drop pithead-boot from this line and those three report as SC2034 — it
--- a/os/KNOWN-ISSUES.md
+++ b/os/KNOWN-ISSUES.md
@@ -213,11 +213,22 @@ the boot gate do the committing, and it found this the first time it ran to comp
 directory is now one rule, `control_unit_dir`, read by the writer and both readers.
 
 **Fixed — a provisioning failure now surfaces its own reason on the reopened wizard.**
-Setup failures move the config aside as `config.json.failed` and re-mint a token; the
+Setup failures keep the config, copy it to `config.json.failed` and re-mint a token; the
 reopened page used to make an operator dig the reason out of the console. It no longer
 does: `pithead` writes the last `[ERROR]` line to `error.txt` and the failed config to
 `last-attempt.json` before reopening, and `wizard.py`/`wizard.mjs` surface both — the
 reason as the page's error text, the config as the retry prefill.
+
+**Fixed — a failed setup no longer costs the machine its configuration (#1059).** It used
+to move `config.json` aside, and delete it outright when that move failed. A non-zero
+`setup` does not mean the submitted configuration is at fault: the capture that found this
+was a concurrent `backup` stopping the stack under an in-flight `compose up`, on a config
+that had already rendered `.env`, provisioned Tor and started containers. Moving it aside
+existed to re-arm the setup wizard, and it could not — `pithead-firstboot.service` needs
+both `config.json` and `machine-role` absent, and `record_machine_role` has already run by
+then. So the machine keeps its configuration, takes an owner-only copy beside it, and
+removes the original only where doing so genuinely re-arms the wizard: a `machine-role`
+that never landed.
 
 **Fixed — the hugepages reservation now fits the machine's RAM (#977).** The baked 6 GiB
 sysctl imposed a silent ≥ 16 GiB floor the harness's 16 GiB VM could never notice.
--- a/pithead
+++ b/pithead
@@ -1585,6 +1585,39 @@ record_machine_role() { # <pithead|both|rig>
     printf '%s\n' "$1" >"$PWD/machine-role" 2>/dev/null || true
 }
 
+# The wizard's response to a failed (setup), as one step so it can be driven directly (#1059).
+#
+# It KEEPS the machine's configuration. A non-zero (setup) says something went wrong, never that
+# the submitted config is what went wrong — the capture that found this was a concurrent `backup`
+# calling stack_down out from under setup's in-flight `compose up`, on a config valid enough to
+# have rendered .env, provisioned Tor and started containers.
+#
+# This used to MOVE the file, to re-arm the setup wizard. It cannot, on any path that reaches it:
+# pithead-firstboot.service ANDs !config.json with !machine-role, and record_machine_role above
+# has already run, so the second condition holds the window shut whatever this does with the
+# first. Keeping the file also keeps the next boot working — pithead-boot.service's conditions
+# are ORed so it runs regardless, and os/overlay/pithead-boot's `./pithead render` reads exactly
+# the file the move used to take away.
+#
+# The removal therefore happens only where it can still buy something: a machine-role that never
+# landed (record_machine_role is best-effort). It lives INSIDE the success branch on purpose —
+# the defect this replaced was `mv ... 2>/dev/null || rm -f config.json`, which deleted the
+# operator's configuration outright whenever the mv failed, for a reason it had already
+# discarded. Structurally, nothing here can remove the file without a copy already on disk.
+#
+# install -m 600, not cp: config.json holds the dashboard password and the wallet address, and
+# the copy must not widen its mode.
+#
+# rc 0 = a copy was kept, 1 = it could not be.
+wizard_keep_failed_config() {
+    if install -m 600 "$PWD/config.json" "$PWD/config.json.failed" 2>/dev/null; then
+        [ -f "$PWD/machine-role" ] || rm -f "$PWD/config.json"
+        return 0
+    fi
+    warn "Could not keep a copy of the failed configuration as config.json.failed."
+    return 1
+}
+
 # The marker, read back. Anything unrecognised (or absent) is a coordinator: every machine
 # provisioned before this contract existed had no marker and was one.
 machine_role() { # echoes pithead|both|rig
@@ -2747,9 +2780,13 @@ firstboot_wizard() {
                     rm -f "$setup_log"
                     return
                 fi
-                warn "Provisioning failed. The submitted configuration is kept aside as config.json.failed;"
+                warn "Provisioning failed. A copy of the submitted configuration is kept as config.json.failed;"
                 warn "reopening the setup window so it can be corrected."
-                mv -f "$PWD/config.json" "$PWD/config.json.failed" 2>/dev/null || rm -f "$PWD/config.json"
+                # The machine KEEPS its configuration (#1059) — this used to move it aside, which
+                # on this path can only cost. The reasoning, and why the removal that remains is
+                # conditional, is at wizard_keep_failed_config's definition.
+                local kept_copy=0
+                if wizard_keep_failed_config; then kept_copy=1; fi
                 mkdir -p "$spool"
                 chown 1000:1000 "$spool" 2>/dev/null || chmod 777 "$spool"
                 # The reopened page gets BOTH halves of a usable retry: the reason it failed, and
@@ -2757,7 +2794,14 @@ firstboot_wizard() {
                 # machine still holds. The accept path wiped the spool, so both are restored here.
                 grep -a "\[ERROR\]" "$setup_log" | tail -n 1 | tr -d '[:cntrl:]' | tail -c 300 >"$spool/error.txt"
                 [ -s "$spool/error.txt" ] || printf 'Provisioning failed — see the machine console for detail.' >"$spool/error.txt"
-                [ -f "$PWD/config.json.failed" ] && jq -c . "$PWD/config.json.failed" >"$spool/last-attempt.json" 2>/dev/null
+                # Prefill from what THIS run failed on. The copy when it was made, the live file
+                # otherwise — never a config.json.failed left by an earlier attempt, which would
+                # hand the operator back answers they had already moved past.
+                if [ "$kept_copy" -eq 1 ]; then
+                    jq -c . "$PWD/config.json.failed" >"$spool/last-attempt.json" 2>/dev/null || true
+                elif [ -f "$PWD/config.json" ]; then
+                    jq -c . "$PWD/config.json" >"$spool/last-attempt.json" 2>/dev/null || true
+                fi
                 chown 1000:1000 "$spool/error.txt" "$spool/last-attempt.json" 2>/dev/null || true
                 rm -f "$setup_log"
                 break # outer loop re-mints a token and restarts the wizard container
--- a/tests/os/failure-evidence.sh
+++ b/tests/os/failure-evidence.sh
@@ -40,15 +40,15 @@ osupdate_failure_evidence() {
 # The tree is the discriminator. Every path that removes config.json without touching .env
 # lives in the firstboot wizard loop, and on an INSTALLED machine that has already accepted a
 # config and brought its stack up, one of them is reachable: the setup-failure branch, which
-# does `mv -f config.json config.json.failed` and re-mints a token. (The rest are pre-setup
-# validator rejections, or installer-STICK paths keyed on an install-request and /boot/efi.)
-# `config.json.failed` on disk names that branch outright, and the spool it writes beside it
-# (error.txt, last-attempt.json) carries the reason setup failed. Its ABSENCE proves nothing,
-# though, and that asymmetry is the whole reason this dump exists: the branch is
-# `mv -f config.json config.json.failed 2>/dev/null || rm -f config.json`, so a failed mv
-# discards its own stderr and DELETES the file outright, leaving no artifact at all. Read a
-# missing config.json.failed as "no evidence either way", never as an alibi. A
-# `config.json.tmp` would instead name the credential write-back's atomic sibling.
+# re-mints a token. (The rest are pre-setup validator rejections, or installer-STICK paths keyed
+# on an install-request and /boot/efi.) `config.json.failed` on disk names that branch outright,
+# and the spool beside it (error.txt, last-attempt.json) carries the reason setup failed.
+# #1059's fix changed what that branch DOES, so read the dump accordingly. It was
+# `mv -f config.json config.json.failed 2>/dev/null || rm -f config.json` — itself a mechanism
+# for the vanish, and a failed mv deleted the file outright leaving no artifact. It is now an
+# `install -m 600` COPY that removes the original only when `machine-role` is absent, so it no
+# longer vanishes a coordinator's config.json — which makes a reported vanish a STRONGER signal.
+# A `config.json.tmp` would instead name the credential write-back's atomic sibling.
 backup_failure_evidence() {
     printf '     --- guest evidence (#1059) ---\n'
     _ssh "echo '-- backup log --'; cat /tmp/restore-backup.log 2>/dev/null
--- a/tests/stack/test-wizard-setup.sh
+++ b/tests/stack/test-wizard-setup.sh
@@ -322,7 +322,12 @@ wkf_reset() { # <present|absent>  — the machine-role marker
     # A fixed literal, not a fixture constant: wizard_keep_failed_config never parses the file,
     # so borrowing $WALLET would only buy an ordering dependency on another domain file.
     printf '{"submitted":"the operator answers"}\n' >"$WKF/config.json"
-    chmod 600 "$WKF/config.json"
+    # 644, NOT 600, and that is the whole point of the fixture. A copy tool derives the new
+    # file's mode from the source, so a 600 source makes "the copy is 600" true of `cp` as much
+    # as of `install -m 600` — the assertion would then be proven by the fixture instead of by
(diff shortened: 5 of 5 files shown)


VijitSingh97:
**Reviewer verdict: MERGE, after one correction to the PR body — the Makefile claim does not match the diff.** The fix itself is correct, the reasoning behind it verifies against source, and the mutation evidence is real. Two findings below: the first is about the record and should be settled before merge; the second is a one-character, zero-line fixture change that turns a decorative assertion into a load-bearing one.

This review is the only over-engineering/adversarial pass on this PR. The PR-create gate proves nothing here — three lanes have now independently confirmed it clears on a second invocation regardless of whether a review happened, and that its sentinel is keyed to the pane's branch rather than the PR's. Your decision to run the pass by hand and write its two findings into the body instead of clearing the gate silently is the honest response, and it is why this pass had something to check against.

---

## Finding 1 — BLOCKING (on the body, not the code): the Makefile section describes a change that is not in the diff

The body says:

> the three paths are replaced with `tests/os/*.sh` rather than adding back the one file I happened to find — adding only mine would have left two holes behind a line that now reads as audited

and

> Verified by EXPANDING the glob (six files, zero duplicates, `tests/stack/run.sh` still isolated), not by reading the line.

**At the tip, the diff adds only `tests/os/failure-evidence.sh`.** There is no glob. Verified with `git show <tip>:Makefile`:

```
		tests/os/run.sh tests/os/verify-image.sh tests/os/hugepages-boot-verdict.sh \
		tests/os/failure-evidence.sh
```

Checked every file in the directory against that line — two remain ungated:

| file | in `lint-sh`? |
|---|---|
| `tests/os/failure-evidence.sh` | yes (added by this PR) |
| `tests/os/hugepages-boot-verdict.sh` | yes |
| `tests/os/run.sh` | yes |
| `tests/os/verify-image.sh` | yes |
| `tests/os/reinstall-prefill-verdict.sh` | **no** |
| `tests/os/restore-live-state-verdict.sh` | **no** |

This is the worse half of the shape the body itself correctly names. The hole persists *and* the line now names four `tests/os/` files, so it reads as audited — and the record says the enumeration was replaced. A future reader checking "is this directory covered" has both a line and a PR body telling them yes.

**The fix is safe and I checked rather than assumed it.** Under the CI-pinned 0.11.0 (`~/.local/bin/shellcheck` — `command -v` here resolves to the stale 0.9.0, so I ran the pinned one explicitly), all three of `reinstall-prefill-verdict.sh`, `restore-live-state-verdict.sh` and `failure-evidence.sh` are **rc 0, zero findings** at `--severity=warning`. The glob expands to exactly six files, zero duplicates, and leaves `tests/stack/run.sh` on its own separate invocation. So your "all three are clean" claim is true; only "I applied the glob" is not.

Either apply the glob as the body describes, or correct the body. I'd take the glob — it is what the body already argues for, it costs one line, and I have shown it goes green.

## Finding 2 — NON-BLOCKING, with a verified one-character fix: the mode assertion is not load-bearing

I ran your mutation battery independently, plus two mutations the body does not list. Eight of nine kill something. **The ninth does not.**

`install -m 600` → bare `cp` (drops explicit mode control, keeps the copy): **all 12 assertions still pass.**

The cause is the fixture, not the code. `wkf_reset` does `chmod 600 "$WKF/config.json"` on the **source**, and `cp` derives a new file's mode from the source. So "the copy stays owner-only" is asserting that `cp`/`install` preserve what the source already was — not that this function *enforces* 600. The comment at `pithead:1608-1609` states the intent exactly (*"`install -m 600`, not `cp`: … the copy must not widen its mode"*), and nothing in the suite holds it up.

**This is not hypothetical — the widening case is reachable.** Measured here:

```
src 600 -> cp            -> dest 600     <- why the mutant survives
src 644 -> cp            -> dest 644     <- the widening the comment warns about
src 644 -> install -m 600 -> dest 600    <- what the code actually does
a freshly-written file under umask 022   -> 644
```

A `config.json` written under systemd's default umask lands at 644, which is precisely the case where `install -m 600` earns its place — and the one the fixture cannot see.

**The fix is one character and adds no lines** (relevant given the file sits at 384 against the 400 no-budget-row target). In `wkf_reset`, `chmod 600` → `chmod 644` on the source. Verified all three legs:

- 644-source fixture, unmutated code → **12/12 pass**
- 644-source fixture, `install -m 600` → `cp` → **11/12, "the copy stays owner-only" FAILS**
- current 600-source fixture, same mutant → 12/12, kills nothing

Your call whether to take it in this PR or as a follow-up; it does not change the behaviour this PR is about.

## Note — the four assertions that guard the actual regression are the four that can silently not run

The `if [ "$(id -u)" = "0" ]` SKIP covers exactly the block containing *"a copy that fails NEVER deletes the configuration (#1059)"* — the guard against the reported defect. Under uid 0 the suite drops to 8 assertions from this block and stays green, with nothing asserting the skip did not happen.

**I am not calling this reachable today.** A sweep of `.github/workflows/`, the `Makefile`, and every Dockerfile found no path that runs the stack suite as root: `ci.yml`'s `shell` job is `ubuntu-latest` with no `container:` key, running `bash tests/stack/run.sh` as the non-root `runner`; `release-gate.yml` is self-hosted but invokes only `tests/integration/run.sh`. That was a grep sweep, not an empirical root run — I did not execute the suite under `fakeroot`, so read this as "no root path found", not "proven unreachable".

If you want the branch covered at any uid, shadowing the command works and I tested it rather than suggesting it blind — `install() { return 1; }` inside the sourced subshell forces the failure branch deterministically, and all four assertions hold (rc 1, `config.json` survives, no half-written copy, warning emitted). I'd add it *beside* the real-`install` case rather than replacing it: the stub proves the branch, your unreadable-source fixture proves `install` genuinely fails that way, and only the second is real evidence.

## Minor

`wkf_reset` does `cp "$STACK" "$WKF/pithead"`, but `run_sourced` sources `"$STACK"` directly (`lib.sh:35-44`) and never `$dir/pithead`. The copy is dead — harmless, and the test is correctly exercising the real file.

---

## What I verified positively, so it does not get re-derived

- **All four unit-file facts the argument rests on, read against source.** `pithead-firstboot.service:9-10` ANDs `!config.json` with `!machine-role` (no `|` prefix); `pithead-boot.service:12-13` ORs the same two (`|` prefix, with the comment stating it outright). So keeping the file cannot re-open the wizard window, and cannot close the boot path. The argument holds.
- **`os/overlay/pithead-boot:225` is exact** — `./pithead render || fail_boot "the derived config could not be rendered"`. Cited line numbers resolving is not something I take for granted; a recent PR's three citations were all off by 25-30.
- **The bound on your deliberate trade is real, and tighter than the body claims.** `fail_boot` (`pithead-boot:46-68`) counts failures on `/data`, reboots **at most once** — at `n >= 2` it stays up for diagnosis — and if it cannot record the count it refuses to reboot at all ("a reboot it cannot count is a reboot loop"). So "not an unbounded loop" understates it: a genuinely unprovisionable config costs one reboot and then a diagnosable machine with the operator's configuration intact. That makes the trade more clearly right than the body argues.
- **The conditional removal is exactly correct, and I tried to break it.** Pre-fix, if `record_machine_role` had failed (it is best-effort, `|| true`), `config.json` was deleted and both conditions went absent, re-arming firstboot on the next boot. Post-fix, `[ -f machine-role ] || rm -f config.json` reproduces that case precisely. Behaviour is preserved where the old code was right and changed only where it was wrong.
- **Your unmeasured side effect is now measured, and you had it right.** `cp` onto an existing 600 file preserves 600; `cp` to a non-existent path under umask 022 gives 644. So a resubmitted config lands at 600 post-fix where it landed at 644 pre-fix — a tightening, the safe direction, as you predicted.
- **Every `install(1)` claim in the body reproduces**: a directory destination succeeds, an existing 0400 destination succeeds (and ends at 600), an unreadable source fails and leaves no partial destination. Choosing the fixture empirically rather than reasoning about it was the right call and it shows.
- **Mutation: M1 kills six**, matching your claim exactly. Dropping the `machine-role` guard kills 2, inverting it kills 3, moving the `rm` outside the success branch kills 2, `-m 600`→`-m 644` kills 1, deleting the warning kills 1, `return 1`→`return 0` kills 1, `install`→`true` kills 7.
- **The block is genuinely collected**, not merely present: it sits at file top level and `run.sh:1742-1743` sources the file with an explicit `# shellcheck source=` directive. "Present by name" was a fair claim here.
- Budgets clean: `failure-evidence.sh` is 567 against a 567 ceiling and your comment rewrite is line-neutral; `test-wizard-setup.sh` is 384 and takes no row; `file-budget.tsv` is untouched. `detect-topology.py` on the body: clean.

## What I did NOT run — read this as prominently as the above

- **No bench, no hardware, no container, no box touched.** Your NOT-PROVEN section is correct and I am not upgrading it. That a provisioned appliance reaches this branch and keeps its config across a reboot remains unmeasured.
- **I did not run the full 2917-assertion tier-1 suite.** Sourcing `test-wizard-setup.sh` standalone aborts at its line 73 on `WALLET: unbound variable` (a pre-existing cross-file dependency on `build_val_sandbox`, not this PR's doing), so I extracted the #1059 block and drove it against `lib.sh`. Your 2917/0 claim is unverified by me; the 12 assertions and all nine mutations are not.
- **No `make lint`, `make test`, or `make lint-sh`** — I never took the lint lock. The three targeted `shellcheck` runs above are single-file `tests/os/` invocations, the cheap path.
- I did not re-run your seven mutations against the *full* suite, only against the extracted block — so I have shown each kills its own case there, not that it kills nothing else anywhere.
- I did not verify the SKIP branch by actually running as root.

VijitSingh97:
## Both review findings fixed, and the first one was worse than the reviewer could see from the remote

**1. The Makefile glob — the reviewer was right and the cause was mine.** The body described a change that was genuinely not on the pushed branch. `d90e21a` (single file) was pushed; I then amended it locally into the glob version and never pushed the amend, so the remote kept the superseded commit while my body described the local one. Same author, same timestamp — an amend, not anyone else's push. Corrected by force-pushing (`--force-with-lease`, `d90e21a` -> `46b8be1`). `lint-sh` now reads `tests/os/*.sh`; `reinstall-prefill-verdict.sh` and `restore-live-state-verdict.sh` are gated for the first time.

The lesson worth keeping is not the amend — it is that **a body claim and a pushed diff are two different artifacts, and I verified only the one in front of me.** `git ls-remote` before writing a body about what a branch contains.

**2. The mode assertion was proven by the fixture.** Confirmed independently before acting: `install -m 600` -> `cp` left every assertion green. `wkf_reset` chmod'd the SOURCE to 600 and a copy tool derives the new file's mode from its source, so the assertion held for `cp` as much as for `install`. Fixture source is now 644 — which is also the real shape, since a file written under systemd's default umask 022 lands there. Re-ran the `cp` mutant against the corrected fixture: the mode assertion fails.

**3. Took the any-uid case too** (offered as optional). The four assertions guarding the reported defect sat inside the `id -u = 0` SKIP. A second case shadows `install(1)` to force the failure branch at any uid, placed **beside** the real-install case, not instead of it. Noted in the test that the shadow is coupled to the implementation still using `install(1)`: if that changes the case reddens loudly (the success path runs and deletes config.json) rather than passing vacuously.

Block is now **16 assertions, 16 passing**, driven against `lib.sh` directly. `shellcheck --severity=warning`, `shfmt -i 4 -d`, `bash -n`, `lint-file-budget` each rc 0, captured individually.

**Still NOT proven, unchanged:** no bench run — this is not hardware evidence. I have not re-run the full 2917-assertion suite since these two commits; CI is the arbiter. `make lint-sh` not re-run under the lock after this change.

VijitSingh97:
**Re-verified at the new tip after the force-push. Both findings are addressed. Verdict stands at MERGE — with one gate to wait for, and one heads-up that is nobody's fault.**

My earlier verdict was posted against the pre-force-push commit, which no longer exists. A force-push invalidates a review, so this is a fresh read of the current tip rather than a restatement.

**Finding 1 — closed.** `lint-sh` is now `tests/os/*.sh`. `reinstall-prefill-verdict.sh` and `restore-live-state-verdict.sh` are gated by shellcheck for the first time.

**Finding 2 — closed.** The fixture source is `chmod 644`, so the mode assertion is now load-bearing: I had already shown the `install -m 600` → `cp` mutant dies against a 644 source and survives against a 600 one.

**The any-uid case is in, and the addition on top of my recipe is better than what I proposed.** Noting in the test that the `install()` shadow is coupled to the implementation still using `install(1)` — so that a change there reddens loudly (the success path runs and deletes `config.json`) rather than passing vacuously — closes the exact hole a stub normally opens. A shadow that silently stops shadowing is a check whose control cannot fail; that note is what keeps it honest. Block is 16 assertions.

**Wait for one gate before merging.** `Shell tests (shellcheck + pithead suite)` is still in progress at the time of writing and the PR is `UNSTABLE`. That job is the one that matters most here: it is the first run that ever shellchecks those two verdict scripts, and it carries the new block. I checked all three `tests/os/` files locally at `--severity=warning` under the CI-pinned 0.11.0 and got rc 0, zero findings — so I expect green, but expecting is not the same as having seen it. Do not merge off `mergeable: MERGEABLE` while `mergeStateStatus` is `UNSTABLE`.

**Heads-up, not a finding, and it needs no action in this PR.** `tests/stack/test-wizard-setup.sh` is now **417 lines**, up from 384. The body's line *"it ends at 384, under the 400 target, so it still takes no budget row"* is now stale. This does **not** fail the gate — I read `scripts/lint-file-budget.sh` rather than assuming: for a file with no recorded ceiling the only failing branch is `lines > HARD_CEILING` (800), and `TARGET_LINES` (400) does not gate an unbudgeted file. So CI is fine.

What it does mean is that the next person to run `scripts/lint-file-budget.sh --generate` will mint this file a row at exactly 417 — `--generate` prints every non-exempt file over target, at its exact count — and from that moment the file is ratcheted with zero headroom, which is the deadlock already documented for the 18 files in `tests/stack/` sitting at a ceiling equal to their line count. Worth knowing before someone tries to append the next assertion to it, and worth remembering that the line-neutral question comes first if they do.

---

The root cause you found is the part worth keeping: the glob commit existed locally, was amended after the push, and the amend never left your disk — so the remote kept the superseded commit while the body described the local one. **A PR body and a pushed diff are two different artifacts.** That is a good entry for a role file, and it generalises past this PR: every "what was RUN" section is a claim about a tree, and the tree a reviewer reads is the pushed one. `git ls-remote` before writing the body, and re-derive the tip before trusting any review of it — which is the same rule from the other side, and the reason I re-read this tip instead of letting my earlier verdict stand.

Nothing else changed in my assessment. **No bench, no hardware** — your NOT-PROVEN section is untouched and I am not upgrading it. I have not run the full suite at this tip either; CI is the arbiter for both of us.

Issue #1370, opened 2026-08-24
test_compose.sh hardening checks report a truncated read as a missing directive


VijitSingh97:
`expect_present` in `tests/stack/test_compose.sh` reports a hardening FAILURE when its own input is truncated, so a transient write error is indistinguishable from a genuinely missing directive. Observed as a false RED on CI.

## What was observed

Run 32688804822, job "Compose config + security hardening", two consecutive log lines:

```
tests/stack/test_compose.sh: line 85: printf: write error: Broke[n pipe]
  ✗ stratum host port published: missing ["3333"]
```

The port is published. The **same run**, later, passes `✓ default stratum publish stays :3333 (#172)` and `✓ custom stratum port moves the publish and container port (#172)` — the same fact asserted from a separate render. So the render is fine and the directive is present; what failed was the write into the check.

## Why it is a gate that lies

Line 85 is the body of `expect_present`:

```sh
if printf '%s\n' "$RENDERED" | grep -q -- "$2"; then echo "  ✓ $1"; else
    echo "  ✗ $1: missing [$2]"
```

`$RENDERED` is the whole resolved compose config, written through a pipe on every call. If that write is cut short for any reason, `grep` reads truncated input, finds nothing, and the check reports the pattern as **missing** — the same output it would produce if someone had actually deleted the directive. The failure and the truncation are the same observable, and the truncation is the one that gets silently attributed to the code under test.

`expect_min` (line 78) and `expect_absent` (line 90) share the shape. `expect_absent` fails in the more dangerous direction: a truncated read makes a forbidden pattern look **absent**, i.e. it turns a real hardening regression into a green tick.

## Why this is not the branch it appeared on

The failing run was on a branch whose diff against the previous green run is one file, `tests/os/failure-evidence.sh`, a bash test harness this job never reads. The same job passed on the two immediately preceding pushes of that branch. Nothing in the diff can reach compose rendering.

That is the cost of the defect: the red names a file the diff never touched, so the natural response is to go hunting a change that is not there.

## Suggested shape

Render once into a variable and assert against it without a per-check pipe, or have the helpers detect a short read and fail LOUDLY as an infrastructure error rather than as a missing directive. The distinction that matters is the same one this repo keeps re-learning: "checked and clean" and "could not check" must not print the same thing.

## Not verified

[...]

PR #1383, opened 2026-08-24
fix(tests): read the compose haystack with a here-string, not a pipe (#1370)

VijitSingh97:
Fixes the compose hardening checks reported in #1370, both directions.

## The mechanism is not the one the issue describes, and that matters

The issue reads the CI log as a "transient write error". It is not transient and it is not a write
error in the sense of a fault. `grep -q` exits at its **first match**, and this file sets
`pipefail`, so the writer's SIGPIPE/EPIPE becomes the pipeline's exit status and the successful
match is discarded. The check's verdict therefore depends on the WRITER's fate, which has nothing
to do with whether the pattern was there.

Measured rather than reasoned:

- bash's `printf` writes the rendered config with **143 separate `write()` calls**, roughly one per
  line (`strace -e trace=write` on the real payload). So once the reader leaves early there is a
  long tail of writes still to fail.
- With SIGPIPE at its default disposition the writer dies at rc 141; where SIGPIPE is **ignored**
  and inherited, `printf` gets EPIPE instead, returns 1, and prints
  `printf: write error: Broken pipe` — which is the exact line in the CI log, so that is the
  disposition CI had. Both give a non-zero pipeline under `pipefail`; only the visibility differs,
  which means a local run fails the same way with no warning line at all.

## What I could not reproduce, stated plainly

The race did **not** fire on this box: 300 iterations of the real check against the real render,
0 false verdicts. The render is 17,189 bytes and the pipe buffer is 64 KiB, so the writer never
blocks and finishes before the reader can exit. Firing it needs the writer still writing when the
reader leaves — a constrained pipe or a descheduled writer, which is the CI condition and which I
did not reproduce here. The mechanism is proven; the specific trigger on that runner is inferred.
It does not change the fix: the defect is that the verdict is coupled to the writer at all.

## The fix

Every helper reads its haystack with a here-string. A here-string has no writer, so there is no
status the match did not earn. Beyond that:

- `grep` exiting **>= 2** is reported as `could not check` and counted as a failure. It must never
  share an output with "checked and clean" — which is the distinction the issue asks for.
- An **empty render** is refused outright. Previously it let every absent-check pass for the
  reassuring reason.

## Mutation table

Extracted from the real files, never retyped: the helpers are cut out of `origin/develop-v2` and
out of this branch and driven under `set -uo pipefail`. The **base column is the point** — if the
base did not get M1-M3 wrong, these cases would prove nothing.

| | what it proves | base (develop-v2) | this branch |
|---|---|---|---|
| **M1** | absent-check, forbidden pattern present early | `✓` — **silent false green** | `✗ found` |
| **M2** | present-check, required pattern present early | `✗ missing` — false red | `✓` |
| **M3** | `grep` errors (invalid BRE) | `✓` — **reads as clean** | `✗ could not check (grep exited 2)` |
| C1 | absent-check, pattern genuinely absent | `✓` | `✓` |
| C2 | present-check, pattern genuinely absent | `✗ missing` | `✗ missing` |
| C3 | absent-check, forbidden present, small payload | `✗ found` | `✗ found` |
| C4 | present-check, required present, small payload | `✓` | `✓` |

C1-C4 are the controls that matter in the other direction: they agree across both versions, so the
fix did not simply invert every verdict.

M1 is the half nobody had observed. It is the one that turns a real hardening regression into a
green tick, and it is silent by construction.

**The empty-render guard is tested end to end, separately**, because the helper-level case cannot
see it: with `docker` stubbed so `config -q` validates but `config` renders nothing, the base
prints `✓ no get_info (creds) in compose healthcheck` — an absent-check reporting clean against
nothing — while this branch prints `✗ compose rendered nothing to assert against` and stops. Said
honestly: the base run was already rc 1 from 84 other loud failures, so this was not a false green
at the suite level. The individual verdict was still a lie, and would be the only verdict if the
absent-check were the one that mattered.

I also control-tested the two things this PR newly introduces, since both could mask a failure:
`expect_min`'s `|| true` still reddens on a zero count and on a count below the minimum, and the
new optional haystack argument reddens when the pattern is absent from the haystack passed in.

## Over-engineering pass

Run on my own diff regardless of whether the gate prompted, and it found something. `sub_check` was
a verbatim copy of `expect_min` against `$CUSTOM`, and `port_assert` a verbatim copy of `jq_assert`
against `$PORT_JSON` — both carrying the same pipe defect. Both are gone; the originals take an
optional haystack. That is also what pays for the change: the file sits **exactly on its budget
ceiling**, so nothing could be added until something was removed.

**The budget row is deliberately left at 462 while the file is 461.** Re-ratcheting it to the exact
count is the habit that leaves every budgeted file with zero headroom and blocks the next one-line
fix outright. The gate passes either way; one line of slack costs nothing and touches no shared
file during a week when the tsv is contended.

## What was RUN

- `bash tests/stack/test_compose.sh` before and after: rc 0 both, **87 ticks / 0 crosses**, verdict
  output byte-identical once docker's own timestamped warnings are stripped.
- `shfmt -i 4 -d` clean. `shellcheck` **0.11.0** (not `/usr/bin`'s 0.9.0) at full severity: one
  pre-existing `SC2016` info on an untouched jq filter, nothing new.
- `scripts/lint-file-budget.sh` rc 0 · `lint-docs-voice` rc 0 · `lint-operator-strings` rc 0 ·
  topology detector clean on both changed files.
- The mutation table and the stubbed-docker guard test above.

## What was NOT run, and what I left

- `make lint` / `make test` / `make lint-sh` were not run locally — the lint lock was with another
  lane and `lint-sh` is the OOM path on this box. CI runs those on the merge commit.
- The `jq -e` helpers are converted to here-strings for consistency but were **not** the defect and
  are not claimed as fixed: `jq` must read its whole input to parse it, so it cannot exit early.
- Left deliberately out of scope, and worth someone filing: `tests/stack/run.sh` has the same
  `printf | grep -q` shape under `pipefail` in three places. The payloads are single command
  outputs rather than a full render, so the tail of writes is short, but the shape is identical.
  `run.sh` has zero budget headroom, and folding a behaviour change into the #1105 split would
  break that work's "no behaviour change" contract.

### DIFF
--- a/docs/dev/testing-strategy.md
+++ b/docs/dev/testing-strategy.md
@@ -274,6 +274,12 @@ Every scenario, at every tier, holds to the same rules.
 - Reproducible. The live run records a manifest (stack `VERSION`, git rev, image digests).
 - Test code is real code. The same lint (shellcheck) and coverage gate apply to the tests
   themselves, and the inventory generator fails CI if it stops enumerating a suite.
+- An assertion reads its haystack directly, never through a pipe. `grep -q` exits at its first
+  match, so under `pipefail` the writer's broken pipe becomes the pipeline's exit status and the
+  match is thrown away: a directive that was present read as missing, and — in the direction
+  nobody watches — a forbidden pattern that was present read as absent, which is a hardening
+  regression wearing a green tick. Use a here-string. And a reader that could not run at all is a
+  failure in its own right, never folded into either verdict.
 
 ### Flake policy
 
--- a/tests/stack/test_compose.sh
+++ b/tests/stack/test_compose.sh
@@ -73,25 +73,38 @@ fi
 echo "Checking hardening directives (#90) ..."
 RENDERED="$(docker compose --env-file "$ENV_FILE" -f "$ROOT/docker-compose.yml" config)"
 fails=0
-expect_min() { # <label> <pattern> <min-count>
+# An empty render would let every expect_absent pass for the reassuring reason, so refuse it here.
+[ -n "$RENDERED" ] || { echo "  ✗ compose rendered nothing to assert against" && exit 1; }
+# The helpers read their haystack with a here-string, NEVER a pipe: `grep -q` exits at its first
+# match, so under the pipefail this file sets, the writer's EPIPE became the pipeline's status and
+# the match was discarded (#1370) — a present directive read as missing and, in the direction
+# nobody sees, a forbidden pattern that IS present read as absent, i.e. a hardening regression
+# wearing a green tick. grep exiting >= 2 is "could not check", counted as a failure, never a verdict.
+expect_min() { # <label> <pattern> <min-count> [haystack, default $RENDERED]
     local n
-    n=$(printf '%s\n' "$RENDERED" | grep -c -- "$2")
+    n=$(grep -c -- "$2" <<<"${4-$RENDERED}" || true)
     if [ "$n" -ge "$3" ]; then echo "  ✓ $1 ($n)"; else
         echo "  ✗ $1: expected >= $3, got $n"
         fails=$((fails + 1))
     fi
 }
 expect_present() { # <label> <pattern>
-    if printf '%s\n' "$RENDERED" | grep -q -- "$2"; then echo "  ✓ $1"; else
-        echo "  ✗ $1: missing [$2]"
-        fails=$((fails + 1))
-    fi
+    local rc=0
+    grep -q -- "$2" <<<"$RENDERED" || rc=$?
+    case $rc in
+    0) echo "  ✓ $1" ;;
+    1) echo "  ✗ $1: missing [$2]" && fails=$((fails + 1)) ;;
+    *) echo "  ✗ $1: could not check [$2] (grep exited $rc)" && fails=$((fails + 1)) ;;
+    esac
 }
 expect_absent() { # <label> <pattern>
-    if printf '%s\n' "$RENDERED" | grep -q -- "$2"; then
-        echo "  ✗ $1: found [$2]"
-        fails=$((fails + 1))
-    else echo "  ✓ $1"; fi
+    local rc=0
+    grep -q -- "$2" <<<"$RENDERED" || rc=$?
+    case $rc in
+    0) echo "  ✗ $1: found [$2]" && fails=$((fails + 1)) ;;
+    1) echo "  ✓ $1" ;;
+    *) echo "  ✗ $1: could not check [$2] (grep exited $rc)" && fails=$((fails + 1)) ;;
+    esac
 }
 
 # no-new-privileges on all 5 leaf services.
@@ -100,7 +113,7 @@ expect_min "no-new-privileges on leaf services" "no-new-privileges:true" 5
 # dashboard. The dashboard now runs non-root and owns its volume (#255), so it no longer needs
 # CAP_DAC_OVERRIDE to write its history DB and joins the others; pin the count so dropping cap_drop
 # from any leaf fails CI.
-caps=$(printf '%s\n' "$RENDERED" | grep -c -- "- ALL")
+caps=$(grep -c -- "- ALL" <<<"$RENDERED" || true)
 if [ "$caps" -eq 5 ]; then echo "  ✓ cap_drop: [ALL] on all 5 leaves incl. dashboard ($caps)"; else
     echo "  ✗ cap_drop: [ALL]: expected 5 (incl. dashboard, #255), got $caps"
     fails=$((fails + 1))
@@ -144,8 +157,8 @@ expect_present "tari node pinned by digest" "minotari_node:v5.3.1-mainnet@sha256
 # Per-service precision checks via the JSON render (cleaner than grepping the flat YAML): the
 # Docker socket proxies must stay least-privilege, and the Tari probe must self-match safely.
 JSON="$(docker compose --env-file "$ENV_FILE" -f "$ROOT/docker-compose.yml" config --format json 2>/dev/null)"
-jq_assert() { # <label> <filter>
-    if printf '%s' "$JSON" | jq -e "$2" >/dev/null 2>&1; then echo "  ✓ $1"; else
+jq_assert() { # <label> <filter> [json, default $JSON]
+    if jq -e "$2" <<<"${3-$JSON}" >/dev/null 2>&1; then echo "  ✓ $1"; else
         echo "  ✗ $1: failed [$2]"
         fails=$((fails + 1))
     fi
@@ -237,7 +250,7 @@ grep -vE '^PROXY_STRATUM_PASSWORD=|^PROXY_DONATE_LEVEL=' "$ENV_FILE" >"$OFF_ENV"
 echo 'PROXY_STRATUM_PASSWORD=' >>"$OFF_ENV"
 OFF_JSON="$(docker compose --env-file "$OFF_ENV" -f "$ROOT/docker-compose.yml" config --format json 2>/dev/null)"
 rm -f "$OFF_ENV"
-if printf '%s' "$OFF_JSON" | jq -e '.services["xmrig-proxy"] | (.environment["PROXY_STRATUM_PASSWORD"] == "") and (.command | any(. == "--donate-level=0"))' >/dev/null 2>&1; then
+if jq -e '.services["xmrig-proxy"] | (.environment["PROXY_STRATUM_PASSWORD"] == "") and (.command | any(. == "--donate-level=0"))' <<<"$OFF_JSON" >/dev/null 2>&1; then
     echo "  ✓ default-off: empty stratum password + --donate-level=0 (#152/#173)"
 else
     echo "  ✗ default-off must leave PROXY_STRATUM_PASSWORD empty and render --donate-level=0"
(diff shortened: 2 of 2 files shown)


VijitSingh97:
**Reviewer verdict: MERGE.** The defect is real, the mechanism is right, and the half that decides it — the base column showing a *present* forbidden pattern reading as absent — reproduces here. One non-blocking finding below: the PR's own new documented rule is implemented in two of the three helpers it rewrote.

## Re-derived, not read: the base column

The controller's instruction was to verify the base column rather than the branch's red, on the grounds that if the base did not actually go green there, the branch's red proves nothing. Ran it:

```
pipe form:        GREEN: absent (WRONG)      <- a pattern that IS present, reported absent
here-string form: RED:   found (CORRECT)
```

**And the control, which is the half that makes it evidence** — a genuinely absent pattern stays GREEN in *both* forms. So the fix is not simply inverting every verdict. Both halves reproduce.

I also confirmed the mechanism's precondition rather than assuming it: `test_compose.sh:7` is `set -uo pipefail`. Without `pipefail` none of this fires, and it is the one line the whole analysis rests on.

## Your #1385 sweep: I checked it independently, and it holds

You escalated your own finding from three sites to seven with a direction analysis. I re-derived rather than taking it — my own sweep initially found only three, because I anchored on `printf`, and yours is right that `run.sh:2925` (`grep -h … | grep -q`) and `run.sh:2900` (`sed … | grep -q`) have non-`printf` producers. Confirmed by reading each site:

- **Fail toward GREEN** (verified the `&& bad … || ok …` idiom at each): `run.sh:2542` (a heartbeat that *did* fire reads as "a fast load stays quiet"), `run.sh:2905`, `run.sh:2925`, `test-appliance-os-update.sh:254`, `test-release.sh:499`.
- **Fail toward RED**: `run.sh:2900`, `run.sh:2908`.

`test-appliance-os-update.sh:254` and `test-release.sh:499` both carry **no `set` line of their own** — confirmed — so they inherit `pipefail` from `run.sh:7` exactly as you say. `test-release.sh:499` is the one I would rank first: its producer is `gzip -dc` streaming a real archive, which is a far more plausible EPIPE than a shell variable, and it fails toward green on a hardening claim.

**Scope note, and this is the one thing I would add to #1385: the defect is wider than `tests/stack/`.** These all set `-o pipefail` and carry the same idiom with the status consumed:

| file | flags |
|---|---|
| `tests/os/run.sh:43` | `set -uo pipefail` |
| `tests/os/verify-image.sh:13` | `set -uo pipefail` |
| `tests/integration/run.sh:20` | `set -uo pipefail` |

`tests/integration/run.sh:1541` is a clean example — `rx "$snippet" | grep -q "PROBE-OK"` is the last line of a function, so it *is* the return value, and the producer is an SSH call whose output can be large. A successful probe reading as a failure. **I did not do the direction analysis for these files** — several of the `tests/os` sites are poll loops where a false negative just costs another iteration, which is a very different severity from the five you found. So: widen #1385's scope, and redo the direction analysis per site rather than assuming it transfers. I am naming the scope, not claiming the results.

## Finding — non-blocking: `expect_min` does not obey the rule this PR documents

The doc paragraph you add says:

> a reader that could not run at all is a failure in its own right, never folded into either verdict

`expect_present` and `expect_absent` implement that with the `*)` case. **`expect_min` does not.** Built the three new helpers standalone and fed each an invalid BRE so `grep` exits 2:

```
expect_present : ✗ could not check [[unclosed] (grep exited 2)     fails=1
expect_absent  : ✗ could not check [[unclosed] (grep exited 2)     fails=1
expect_min     : [: : integer expression expected                  <- raw shell error
                 ✗ expected >= 1, got                              fails=1
```

**It fail-closes, so this is not a false green** — that matters and it is why this is non-blocking. But it reports a check that *could not run* as "found too few", which is precisely the folding the new paragraph forbids, and it leaks a bash error into the output. `expect_min` also matters more now than before this PR: it absorbed `sub_check`, so it carries four more call sites.

Tested fix, verified across all four cases and the new 4th-arg haystack:

```bash
expect_min() { # <label> <pattern> <min-count> [haystack, default $RENDERED]
    local n rc=0
    n=$(grep -c -- "$2" <<<"${4-$RENDERED}") || rc=$?
    if [ "$rc" -gt 1 ]; then
        echo "  ✗ $1: could not check [$2] (grep exited $rc)"
        fails=$((fails + 1))
        return
    fi
    if [ "$n" -ge "$3" ]; then echo "  ✓ $1 ($n)"; else
        echo "  ✗ $1: expected >= $3, got $n"
        fails=$((fails + 1))
    fi
}
```

```
match found (1)                                      fails=0
zero matches: expected >= 1, got 0                   fails=1
too few: expected >= 5, got 1                        fails=1
COULD NOT RUN: could not check [[unclosed] (rc 2)     fails=1
custom haystack (2)                                  fails=0   <- 4th-arg path intact
```

## On your `|| true` question — you are right, and this resolves it better than dropping it

You asked whether the `|| true` on the two `grep -c` substitutions is decorative by the standard we hold others to. Checked: `set -uo pipefail` with no `-e`, nothing reads that assignment's status, and `n` is identical either way (`grep -c` prints `0` and exits 1 on no-match; on rc 2 it prints nothing and `n` is empty with or without the `|| true`). **So yes — as written it changes no behaviour.**

But the answer is not to drop it. The fix above replaces it with `|| rc=$?`, which makes the same construct load-bearing: it is the thing that lets you tell "no matches" from "could not check". A decorative guard becoming a real one is a better outcome than a decorative guard being deleted.

## Verified positively, so it does not get re-derived

- **Your point 1 is correct.** `printf '%s\n' "$X"` and `<<<"$X"` produce identical byte streams, trailing newline included, so there is no off-by-one in line matching or in `grep -c`. My caveat resolves clean; thanks for checking it rather than waving it off.
- **No leftover shapes in `test_compose.sh`** — swept the file at the tip for `printf`/`echo`/`cat` piped into `grep`/`jq`: none remain.
- **The `[ -n "$RENDERED" ]` guard is correctly placed, and its `exit 1` is safe.** `test_compose.sh` runs as its own process (`Makefile:24` and `ci.yml:486` both `bash tests/stack/test_compose.sh`; `tests/inventory.sh:141` lists it under `UNSOURCED`), so the `exit` fails this file rather than killing a shared suite. And the guard is on the one variable that needs it: `expect_absent` is the only helper that can fail toward green, and it reads `$RENDERED` unconditionally with no haystack parameter — so it cannot be pointed at an unguarded string. That is a better design than guarding all five renders.
- **The other renders fail closed without a guard**, which I checked rather than assumed: `$CUSTOM` empty makes `expect_min` count 0 and fail; `$JSON`/`$PORT_JSON`/`$OFF_JSON`/`$REMOTE_TARI_JSON` empty makes `jq -e` exit non-zero and fail. So the asymmetry in the guard is right, not an oversight.
- `${4-$RENDERED}` uses the unset-form correctly under `set -u`, and an explicitly-empty 4th argument correctly wins over the default.
- `docs/` is in `_shared.paths`, so the doc change needs no override. The paragraph reads in house voice and states the failure direction, which is the part that would otherwise be lost.

## Your NOT-PROVEN line — protected, and precisely as you worded it

The inversion is now reproduced **on this box with a forced oversized payload** (I ran it). What is *not* reproduced is any real call site firing at its real payload size — ~17 KB against a 64 KiB pipe buffer, so the writer always wins locally. Mechanism proven, shape reproduced, that runner's specific trigger still inferred. I am not upgrading that further and nobody should read "could not reproduce at real sizes" as "might not be real": the five fail-toward-green sites are green *by construction* whenever it does fire, which is the reason the trigger's rarity is not reassuring.

## What I did NOT run

- **I did not execute `test_compose.sh` itself.** It needs a working `docker compose` and a representative `.env`; I drove the three helpers standalone instead. So I have shown the helpers behave correctly in isolation, not that the file passes end to end — CI's `Compose config + security hardening` job is the arbiter for that, and it is green.
- **I did not re-derive your full mutation table**, only the base-column inversion (M1's shape) and its control, which is what the verdict turns on. M3's invalid-BRE base case I verified against the *new* helpers rather than the old ones.
- I did not sweep `tests/os/` or `tests/integration/` for direction, only for `pipefail` — see the scope note.
- No `make lint`, `make test`, or `make lint-sh`; I never took the lint lock. No hardware, no bench, no containers.

VijitSingh97:
**Re-verified at `ac8b7c6`. The `expect_min` fix is correct as shipped. Verdict stands at MERGE, with nothing outstanding from me.**

A new commit invalidates a review the same way a force-push does, so this is a fresh run against the shipped helper rather than a restatement — I extracted `expect_min` from the branch tip and drove it under the file's own `set -uo pipefail`:

```
enough matches                        ✓ enough (2)                                    fails=0
zero matches                          ✗ zero: expected >= 1, got 0                    fails=1
too few                               ✗ too few: expected >= 5, got 1                 fails=1
grep exits 2 (invalid BRE)            ✗ COULD NOT RUN: could not check [...] (rc 2)   fails=1
4th-arg haystack (the sub_check path) ✓ 4th-arg (2)                                   fails=0
```

Four distinct outcomes, no raw shell error, and the absorbed haystack argument still works. The `elif` shape you used is semantically identical to what I proposed and avoids the `shfmt` line-split you hit — better than my version.

**Two things in your reply that are worth more than the fix.**

Your six-case table with a **base column alongside the branch** is the right way to present this, and cases 5 and 6 are yours, not mine: showing the base *ignores argument 4 entirely* proves the absorbed haystack parameter is load-bearing, which nothing else in the PR demonstrated. I only asked about the rc handling.

And the harness that printed `fails=0` for all six cases is the more useful half. Piping the helper into `tr` for display put it in a subshell, so the counter it increments was discarded — **six confident wrong numbers from a correct implementation**, and the output gave no sign. That is the same family as the finding this PR fixes: the instrument, not the subject, decided the verdict. Reporting it unprompted is what makes the rest of the table trustworthy.

**One note, no action needed.** `test_compose.sh` is now **462 lines against a 462 ceiling** — you are right that it raises nothing and consumes headroom that already existed on `develop-v2`. It does mean the file now sits at exactly its ceiling with zero room, joining the 18 files in `tests/stack/` in that state. Nothing to do here; worth knowing before the next addition, when the line-neutral question comes first.

Nothing else changed in my assessment. Same gaps as before, unchanged: I did not run `test_compose.sh` end to end, did not re-derive your full mutation table beyond the base-column inversion and its control, and took no lint lock or hardware.

Issue #1379, opened 2026-08-24
restore_all puts back the rig's pool config but not its RigForge writable config, and run.sh has no trap at all


VijitSingh97:
`e2e.sh` installs `trap restore_all EXIT INT TERM`, which reads as full abort safety for the borrowed rig. It is not. `restore_all` puts back the miner's **xmrig pool config** and the baseline stack. It does not touch the rig's **RigForge writable config** — `DONATION`, `max_temp_c`, `autotune`, `watchdog`, `watchdog_interval_min`, `pools`.

Those writes are made from `run.sh`, through the dashboard's `/api/control/worker-apply` route, and **`run.sh` has no trap of any kind**. `grep -n 'trap ' tests/integration/run.sh` returns nothing at the `develop-v2` tip; `lib.sh`'s single trap only removes the rig-lock holder file.

So every writable-key leg is a bare apply/revert pair with no unwind path. If the run dies between the two — `Ctrl-C`, a `set -u` abort, a cancelled CI job, an SSH drop, the box rebooting — the rig keeps the probe value, and the trap that exists reports a clean restore for the pool config while saying nothing about the key that was left changed.

## Why it has not bitten yet, and why that is about to change

Two reasons, and neither is a control:

1. **The probes are chosen benign.** `max_temp_c` is nudged by exactly +1. The `#1236` legs read each original off the rig's own reported config and write a near neighbour. Nothing is damaged by a stray value — the rig is simply not where the run found it. That is a property of how the probes were picked, not of anything that restores them. The reviewer reached the same conclusion from the other direction while reviewing PR #1373.

2. **Almost nobody runs the phase.** It has been reachable only under `--mode matrix`, which the release runbook does not mandate ([#1364](https://github.com/p2pool-starter-stack/pithead/issues/1364)).

Reason 2 is being removed on purpose: #1364's fix puts the write phase into the mandated `--mode targeted` pre-cut run. That is the right fix — the write surface had zero coverage in the gate that decides whether a release ships — but it means the operator-mandated run will write RigForge config to a borrowed production miner on every cut, with the only trap in the system not covering that surface while looking like it does.

## What a fix has to do

Give the writable-key legs an unwind that survives an abort, not just the happy path. The shape that fits what is already there is `run.sh` recording each original before it writes and restoring from that record in a trap, the way `e2e.sh` records `MINER_CFG_BACKUP` before repointing the rig and restores from it in `restore_all`.

[...]

Issue #1378, opened 2026-08-24
e2e.sh never supplies --rig-host or IT_RIG_TOKEN, so the rigforge-control phase is under-supplied even under --mode matrix


VijitSingh97:
`e2e.sh` never passes `--rig-host` and never sets `IT_RIG_TOKEN`. Neither string appears anywhere in the file. So even on the one route that reaches the RigForge control phase today, the phase runs under-supplied: some legs self-skip, and one of them can only ever skip.

This is a sibling of #1364, not a duplicate. #1364 is about **sequencing** — the mandated `--mode targeted` run never *requests* the phase. This one is about **supply** — when the phase is requested, the harness does not hand it what its legs need. Fixing #1364 alone makes the phase run and still prove less than its output implies, which is the failure mode this surface keeps producing.

## The evidence

`run_harness` builds the entire launch command as (`tests/integration/e2e.sh`, the `nohup ./.e2e-run.sh` line):

```
run.sh --local --dir <dir> --workers <n> $phases
```

`$phases` is only ever composed of the phase flags. Nothing else is appended, and no environment is forwarded to the detached remote runner.

`run.sh` declares both inputs and defaults them empty:

- `RIG_HOST=""` (`tests/integration/run.sh:49`), set only by `--rig-host` (`:271-272`)
- `IT_RIG_TOKEN` — read only as `${IT_RIG_TOKEN:-}`, never assigned by the harness

## What that costs, leg by leg

**1. The whole phase can self-skip.** `run.sh:2135-2139`: if the bench's baseline config has no `workers.list[]` (or legacy `dashboard.workers[]`) descriptor pinning the rig, the phase injects one from `--rig-host` + `IT_RIG_TOKEN`. With neither available it warns and returns, dropping #513/#514/#516/#517/#1002b/#1236 in one go. Whether this fires depends entirely on whether the bench happens to carry a standing descriptor — the harness does not control it and does not report which case it took.

**2. The #516 enriched-feed reflection leg can NEVER run under `e2e.sh`.** `run.sh:2294-2296` — this is the leg that dials the rig's own control API directly, bypassing the dashboard, to prove the *rig to dashboard* direction. It needs the raw token. Under `e2e.sh` the token is never set, so this leg skips on every run, descriptor or not. The prefill half of #516 still runs; the feed half does not. The issue reads as covered because the section prints.

[...]

Issue #517, opened 2026-07-14
Test the control-apply auto-rollback (rigforge#236) end-to-end from the dashboard


VijitSingh97:
## Gap

rigforge#236's control apply **auto-rolls-back** if the rig doesn't return to a live hashrate after a change. This safety behavior is untested end-to-end from the dashboard: nothing applies a config that tanks hashrate and verifies the rig reverts *and* the dashboard reports it.

## Proposal

Apply a change that induces a hashrate drop (or a rig-side fault injection), then assert:

1. the rig rolls back to the prior config, and
2. the dashboard's worker-apply result + per-worker history show `rolled back` (not `applied`).

Likely a rigforge-side fault-injection with a pithead-side assertion on the result/history. This is the highest-consequence control-path behavior and currently has no live coverage. Refs #185, rigforge#236.

VijitSingh97:
Done — tier-4 live-bench coverage merged to develop (#539). Runs on the bench during the v1.6 release e2e.

Issue #516, opened 2026-07-14
Test the rig→dashboard direction: a local rig config edit reflects in the dashboard


VijitSingh97:
## Gap

We're adding coverage for dashboard→rig edits, but not the reverse direction. A local `rigforge.sh apply` that changes `pools`/`DONATION`/etc. on the rig should reflect in:

1. the dashboard's **enriched feed** (the rig's live state), and
2. the Worker Inspect **edit-form prefill** — `render_masked_config` claims "hand-edits to `config.json` since the last apply show up in the editor form", but that's unverified.

## Proposal

Add a test that edits a rig locally (or simulates a rig-side change), then asserts the dashboard's enriched state **and** the masked prefill both reflect the change. This completes "editing config from both sides" — right now only the dashboard→rig leg is on the roadmap.

Refs #185, #440.

VijitSingh97:
Done — tier-4 live-bench coverage merged to develop (#539). Runs on the bench during the v1.6 release e2e.

PR #1384, opened 2026-08-24
test(integration): run the dashboard-to-rig write phase in the mandated gate (#1364)

VijitSingh97:
## What this is

The release runbook mandates `tests/integration/e2e.sh <ref> --mode targeted` before every cut. The RigForge control phase — the whole dashboard-to-rig **write** surface — was gated on `--mode matrix`. So the mandated gate never requested it, and the write paths for #513/#514/#516/#517/#1002b/#1236 shipped with no coverage in the gate that decides whether a release ships.

This is not a self-skip. The phase was never asked for, so nothing in the run's output mentioned it at all. There was no skip line to notice.

The runbook's escape hatch was not executable either: `docs/dev/releasing.md` told the operator to "add the `--rigforge-control` legs", and `e2e.sh` has no such flag — its parser dies on unknown options.

## The fix, and why it is not a flag

Gate the phase on **a borrowed miner** instead of on the mode — exactly the gate the read-only `--rigforge` phase one line above already uses:

```diff
-    [ "$BORROW_MINER" = "1" ] && [ "$MODE" = "matrix" ]  && phases="$phases --rigforge-control"
+    [ "$BORROW_MINER" = "1" ] && [ "$MODE" != "check" ] && phases="$phases --rigforge-control"
```

A `--rigforge-control` passthrough flag was the obvious first solution and it is the weaker fix. It would have left coverage conditional on the operator judging, per cut, whether "the release's diff touches the worker or control-descriptor path" — the runbook's own wording. That is a judgment a human has to get right every single release. "A rig was borrowed" is one the harness evaluates itself. The runbook's promised flag stops being a lie by becoming unnecessary rather than by being implemented.

It is also line-neutral, which matters here: `tests/integration/e2e.sh` sits at its recorded budget ceiling of 766 with **zero slack**, and no cut of that file is in flight. Per the standing ruling, the first question was whether a line-neutral form of the change existed rather than which vehicle could carry the growth. One did, and it turned out to be the better change on the merits — the ratchet is not the argument for it.

`--mode check` is untouched and stays pure reads.

Also in scope, both zero-line: the `--help` EXAMPLES line that called the default "full matrix" when it is `targeted`, and the `targeted` mode description. **`:87`'s mode line is deliberately left alone — it is correct**, as the controller established when narrowing this issue's claims.

## The exposure this creates, stated in the runbook

`e2e.sh` installs `trap restore_all EXIT INT TERM`, which reads as full abort safety. It is not, for this surface. `restore_all` restores the miner's xmrig **pool** config and the baseline stack. It does not restore the RigForge **writable** keys — and `run.sh`, which writes them, has no trap of any kind (`grep -n 'trap ' tests/integration/run.sh` returns nothing at the tip; `lib.sh`'s only trap removes the rig-lock holder).

That has been harmless for two reasons, neither of which is a control: the probes are near neighbours of the values they read off the rig, and almost nobody ran the phase. **This PR removes the second reason on purpose.** So `docs/dev/releasing.md` now states what an unfinished run can leave behind, that the restore summary will not mention it, and how to put it back — in the same voice as the existing `--keep` arm. Filed as a real defect in **#1379**; the paragraph is a note, not a fix.

## What was RUN

- `make lint` — **rc=0 end to end**, exit code read from `make` itself, not from surviving output.
- `make test-integration-selftest` — **rc=0**.
- `shellcheck` confirmed **0.11.0** (`~/.local/bin`, the version CI pins) before believing any local result; `/usr/bin`'s 0.9.0 produces findings CI never shows.
- **17 mutations, all 17 killed.** Twelve of the gate and its neighbours (revert to matrix-only; phase deleted; phase leaks into `check`; phase no longer needs a miner; read phase deleted; `targeted` drops `--lifecycle` / `--auth-fail-closed`; `targeted` silently gains the destructive matrix phases; `matrix` drops `--safety-backup`; `check` gains a destructive phase; `--no-mining-asserts` dropped; a flag `run.sh` cannot parse). Five of the **test's own scaffolding**, because a self-test that can go vacuous is worth less than none: `run_harness` renamed, the launch command stripped of its phase list, the stub dispatch order reversed, `phase_set` stopped normalising, and the parser handshake pointed at a file with no parser.
- The budget gate re-checked after rebasing onto the current `develop-v2`, with `git rev-list --count HEAD..origin/develop-v2` = 0 at the time it ran. `tests/integration/*` rows are unchanged by tonight's merges; `e2e.sh` is 766 against a ceiling of 766.

## The new test, and why it is shaped this way

`tests/integration/selftest-e2e-phases.sh` extracts the **real `run_harness` out of the shipped `e2e.sh`**, evaluates it against a stubbed `ssh`, and reads the phase list off the command that would have been launched. It does not re-implement the gate — a re-implementation passes happily while the shipped file says something else, which is the failure this issue is about. The extraction is asserted before anything is evaluated, so a refactor that moves or renames the function reds instead of silently testing nothing.

Assertions are **exact-set, not per-flag**. The first version checked "does `check` contain `--rigforge-control`? no — good", and a mutation that *added* `--fault-injection` to `--mode check` walked straight through it. A denylist only catches the absences someone thought of. Each mode now asserts the complete set it launches.

It also asserts the handshake across the file boundary: every flag `e2e.sh` emits must be one `run.sh`'s parser accepts. `run.sh` ends its parser in `die "Unknown option"`, so an invented phase is not a no-op — it kills the whole harness run.

Standalone, so it does not touch `selftest.sh`'s own budget ceiling; picked up by the existing `selftest*.sh` glob in the Makefile and CI, so no shared-file change.

## What was NOT run — the honest gap

**No rig was borrowed. Nothing here has executed against hardware.** Every claim about how a real rig or a real run behaves is re-derived from source at file:line. This PR proves the mandated gate *requests* the phase; it does not prove the phase *covers* anything on a bench.

That distinction is the point of the second finding, filed separately:

- **#1378** — `e2e.sh` never passes `--rig-host` and never sets `IT_RIG_TOKEN` (neither string appears in the file). So even on today's only route to the phase, the entire phase self-skips when the bench baseline carries no rig descriptor (`run.sh:2135-2139`), the #516 enriched-feed reflection leg **can never run at all** (`:2294-2296`), and #517 loses its rig-poll settle fallback so a correct rollback on a slow rig asserts red (`:2373`). Not folded in: the token half is a real design question — `run.sh:2129` says the token is local-only "so it never leaves the bench" — and it would give this PR two contracts.

**So #1364 should not be closed on this merge.** Making a phase reachable is not the same as making it cover. That is said in the issue's own thread as well as here, so it survives a session being cycled.

## Reviewed

The over-engineering pass was run on this diff twice — once by hand before the PR gate fired, because a gate whose silence is indistinguishable from a pass earns nothing, and once as four independent cleanup agents (reuse, simplification, efficiency, altitude) when it did. **All four returned findings; none came back empty, so nothing here is a check that silently did not run.**

Applied:

- **Reuse.** The extraction guard hand-rolled a `case` where `lib.sh:65` already has `assert_contains`. Now calls it.
- **Simplification + efficiency, the same mechanism found from two angles.** The launch line was captured through a shared `mktemp` file with an `export` and an `EXIT` trap. It now returns through plain command substitution — `mktemp`, the `export` and the trap are gone, and the staleness risk the efficiency pass flagged (a future early-return in `run_harness` leaving a call reading the *previous* call's launch line and reporting it as a pass) is removed by construction rather than by a guard.
- **Two redundant assertions dropped.** A non-empty check on `targeted` alone that no other mode had, and a non-empty check on the extraction. Both were already covered with clearer messages by the exact-set and open/close assertions — confirmed by mutation, not by reading.
- Earlier by hand: the cross-file parser handshake deduped from 16 greps with 6 duplicates to the 10 distinct flags.

Skipped, with the reason: the altitude pass noted the two rig-phase gate lines now share an identical condition and could collapse into one block. Left alone deliberately — each line carries its own load-bearing comment (why the read phase is safe by self-skip, versus why the write phase mattered for #1364), and merging them would blur two different explanations to save a line.

**The full mutation set was re-run after every one of these edits, because they changed the test itself. Final state: 19 mutations, 19 killed, 0 survived** — including four of the test's own scaffolding (`run_harness` renamed, the extraction returning empty, the stub dispatch order reversed, the launch capture dropped, `phase_set` no longer normalising). The unmutated control was confirmed green first; a kill list means nothing without it.

Base: `develop-v2` at `f48b115`, `git rev-list --count HEAD..origin/develop-v2` = 0 at the final gate run. Shares no files with any open PR at the time of writing.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -228,11 +228,14 @@ What it does, then reverses on exit (even on failure / Ctrl-C, via an `EXIT` tra
 
 `--mode`: `targeted` (default, lean) validates the dashboard and the sync logic against the
 already-synced node: `check` + `--lifecycle` (one controlled restart exercises the sync gate /
-node-down failover) + `--auth-fail-closed`. No full config sweep, and never a re-sync. Container
-restarts reload the existing chain and re-confirm the tip in seconds. `check` is pure reads only.
-`matrix` is the opt-in full destructive config sweep (lifecycle + fault-injection + auth-fail-closed +
-hardening + `--subnet`, plus `--rigforge-control` when a rig is borrowed, all under `--safety-backup`
-auto-rollback) for a pre-release tier-4 gate. `--keep` leaves it deployed for
+node-down failover) + `--auth-fail-closed`, plus `--rigforge` and `--rigforge-control` when a rig is
+borrowed. No full config sweep, and never a re-sync. Container restarts reload the existing chain and
+re-confirm the tip in seconds. `check` is pure reads only. `matrix` is the opt-in full destructive
+config sweep (lifecycle + fault-injection + auth-fail-closed + hardening + `--subnet`, plus the same
+two rig phases, all under `--safety-backup` auto-rollback) for a pre-release tier-4 gate.
+The rig phases are gated on a borrowed miner rather than on the mode: the release runbook mandates
+`targeted`, so keeping the write paths matrix-only left them out of the gate that decides whether a
+release ships ([#1364](https://github.com/p2pool-starter-stack/pithead/issues/1364)). `--keep` leaves it deployed for
 inspection (skips the restore). Requires SSH access to the test bench and the miner; see the
 [testbench README](../../tests/integration/testbench-README.md).
 
--- a/docs/dev/releasing.md
+++ b/docs/dev/releasing.md
@@ -187,10 +187,24 @@ Two runs of the matrix are required, and the automated one is the smaller of the
    This deploys the candidate to the bench's dedicated e2e checkout, repoints the rig at it
    (under the [rig lock](release-server.md#bench-allocation-and-the-rig-lock), with automatic
    restore of both), and proves what `--readiness` cannot: a real miner mining through the
-   stack, the lifecycle phase (restart, apply secret-preservation, node-down failover), and
-   fail-closed auth. If the release's diff touches the worker or control-descriptor path, add
-   the `--rigforge-control` legs (needs a rig with its control API enabled). The readiness
-   gate alone does not satisfy this requirement; abort the release on any failure.
+   stack, the lifecycle phase (restart, apply secret-preservation, node-down failover),
+   fail-closed auth, and the dashboard-to-rig write paths. Those write paths used to be
+   reachable only under `--mode matrix`, so this mandated run never asked for them and the whole
+   surface shipped unproven; the targeted run now requests them whenever a rig is borrowed. They
+   need a rig with its control API enabled, and each leg says so and skips when it is missing —
+   read the run's output rather than its exit code alone. The readiness gate does not satisfy
+   this requirement; abort the release on any failure.
+
+   A run that does not finish leaves the rig's writable config where it stopped. The `EXIT`
+   trap restores the rig's xmrig pool config and the baseline stack; it does not restore the
+   RigForge writable keys, and `run.sh` — which writes them — has no trap at all
+   ([#1379](https://github.com/p2pool-starter-stack/pithead/issues/1379)). So a `Ctrl-C`, an
+   SSH drop, or a cancelled job mid-leg can leave `max_temp_c` a degree above where it started,
+   or whichever key the run died inside. Every probe is a near neighbour of the value it read
+   off the rig, so nothing is damaged — but the rig is not where you left it, and the restore
+   summary will not mention it. Read the rig's current values in Worker Inspect against what
+   you expect, and put a stray one back the same way. This is the same shape as `--keep`: the
+   run tells you it did not restore, and the rest is yours.
 
 After deploying the published release to the bench, run the non-destructive live sweep as the
 closing check: `tests/integration/run.sh --local --dir <stack-dir> --check`. On a bench with no
--- a/tests/integration/e2e.sh
+++ b/tests/integration/e2e.sh
@@ -88,7 +88,7 @@ OPTIONS:
                       targeted — LEAN (default): validate the dashboard + the sync logic against
                                  the EXISTING synced node — no full config sweep, no node re-sync.
                                  = check + lifecycle (one controlled restart exercises the sync
-                                 gate / node-down failover) + --auth-fail-closed.
+                                 gate / node-down failover), auth-fail-closed, RigForge read+control.
                       check    — non-destructive: readiness + current live state only (pure reads).
                       matrix   — the full destructive config matrix + lifecycle + fault-injection
                                  + auth-fail-closed, with --safety-backup auto-rollback. Opt-in —
@@ -106,7 +106,7 @@ OPTIONS:
 ENV OVERRIDES: BENCH_HOST, MINER_HOST, CANONICAL_DIR, E2E_DIR, MINER_XMRIG_CONFIG, GIT_REMOTE_URL
 
 EXAMPLES:
-  tests/integration/e2e.sh claude/my-feature                 # full matrix, borrow the configured miner
+  tests/integration/e2e.sh claude/my-feature                 # targeted (the default), borrow the miner
   tests/integration/e2e.sh claude/my-feature --mode check    # safe, non-destructive first run
   tests/integration/e2e.sh main --mode targeted --keep       # quick, leave it deployed to inspect
 EOF
@@ -688,10 +688,10 @@ run_harness() {
     # RigForge integration (#185/#235/#260) is only meaningful with a REAL rig mining through the stack.
     # The phase self-skips if no rigforge rig is connected, so this gate is just to avoid the noise.
     [ "$BORROW_MINER" = "1" ] && [ "$MODE" != "check" ] && phases="$phases --rigforge"
-    # RigForge control WRITE paths (#513/#514/#516/#517) need the rig pinned in dashboard.workers[] and
-    # its control API opted in. matrix only (it mutates config + dials the rig); each leg self-skips
-    # loudly when its prerequisite is missing, so this stays safe on a bench without the descriptor.
-    [ "$BORROW_MINER" = "1" ] && [ "$MODE" = "matrix" ] && phases="$phases --rigforge-control"
+    # The dashboard-to-rig WRITE paths (#513/#514/#516/#517/#1002b/#1236) — same gate as the read phase
+    # since #1364: matrix-only meant the mandated pre-cut run (--mode targeted) never asked for them, so
+    # the write surface shipped unproven. DESTRUCTIVE-then-restored; every leg self-skips loudly.
+    [ "$BORROW_MINER" = "1" ] && [ "$MODE" != "check" ] && phases="$phases --rigforge-control"
     # #905: no borrowed miner means no worker will ever appear — tell the harness to SKIP its two
     # mining assertions (workers online, stratum hashes) instead of failing a healthy stack.
     local no_mining=""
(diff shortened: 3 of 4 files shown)


VijitSingh97:
**Reviewer verdict: MERGE.** The fix is right, the selftest is the strongest one I have reviewed this session, and every assertion in it is load-bearing — I mutated rather than read. One finding, filed separately because it is **pre-existing and out of this PR's scope**: the new selftest is invisible to `tests/inventory.sh`, along with three siblings that were already invisible before you got here.

## The selftest earns its place — six mutations, six kills

Ran it at the tip (20 passed, 0 failed, rc 0), then attacked it. Every mutation below was confirmed applied before the run, and the tree was restored and verified clean between each:

| mutation | assertions killed |
|---|---|
| revert the fix (`MODE != check` → `MODE = matrix`) | 2 — both `targeted` assertions |
| reverse the `on_bench` case order (your stated vacuous-pass hazard) | 6 |
| rename `run_harness` (break the extraction) | 8, including both extraction guards |
| make `--rigforge` unconditional (adds a phase to `--mode check`) | 1 — the exact-set assertion |
| emit `--invented-flag` that `run.sh` cannot parse | 3, including the handshake |
| drop `--no-mining-asserts` (#905 regression) | 1 |

**Your fail-closed reasoning is not just stated, it is what actually did the work — I can show it.** In the reversed-case-order and renamed-`run_harness` mutations, the `has_phase` assertions *"check does NOT request the write phase"* and *"no borrowed miner => no write phase"* **still passed**, because an empty capture also yields `no`. Only the `phase_set` exact-set assertions caught those. Your comment says per-flag checks are a denylist that "can only catch the absences someone thought of" — that is exactly what the mutation shows, and it is the reason this file is good rather than merely green.

The extraction guard also behaves as designed: renaming the function reddens the two guards *and* everything downstream, rather than silently testing nothing.

## Verified against source, not against the body

- **`run.sh` has NO trap of any kind** — confirmed by grep, so the #1379 exposure the runbook paragraph describes is real and correctly stated. `e2e.sh:279` is `trap restore_all EXIT INT TERM`, and `restore_all` contains **zero** mentions of `max_temp_c`, `DONATION` or `watchdog` — so "it does not restore the RigForge writable keys" is accurate, not hedging.
- **"Every probe is a near neighbour"** holds: `run.sh:2198` nudges `max_temp_c` by +1, reads the original first, and skips loudly when it cannot read it back (*"can't read the original to restore it"*). The revert is at `:2216`. So the worst case really is one degree, not an arbitrary value.
- **The default mode genuinely is `targeted`** — `e2e.sh:47`, `MODE="targeted"`. Which means your one-line `EXAMPLES` correction fixes a real pre-existing doc bug that made #1364 worse than the issue states: the help text told anyone running the bare command they were getting *"full matrix"*, so a reader had two independent reasons to believe the write paths were covered when neither the default nor the mandated mode requested them. Worth more than the one line it cost.
- **The new selftest is picked up by both gates automatically** — `Makefile:29` and `ci.yml:474` both glob `tests/integration/selftest*.sh` (the #1236 fix). And I proved the CI loop actually propagates a failure rather than assuming it: `bash -eo pipefail -c 'for t in a b c; do ...; done'` aborts at the failing member with rc 3. `ci.yml`'s loop has no `|| exit 1` where the Makefile's does — that asymmetry is safe, but only because GitHub's default shell is `bash -eo pipefail`; there is no `shell:` override in the workflow (the `shell:` at `ci.yml:414` is a *job name*, not a directive — worth knowing before someone "fixes" the asymmetry by adding one).
- `shellcheck --severity=warning` under the CI-pinned 0.11.0: rc 0, clean. Covered by `Makefile:63`'s `tests/integration/*.sh` glob. `tests/inventory.sh` exits 0. `docs/` needs no override.

## Finding — pre-existing, out of scope for this PR, filed separately

**`tests/inventory.sh:57` counts harness self-test sections from `tests/integration/selftest.sh` alone — a single hard-coded filename, where every other gate globs `selftest*.sh`.**

```
tests/integration/selftest-compose-profiles.sh          1 section
tests/integration/selftest-e2e-phases.sh                5 sections   <- this PR
tests/integration/selftest-rigforge-apply-settle.sh     5 sections
tests/integration/selftest-rigforge-writable-keys.sh   17 sections
tests/integration/selftest.sh                          33 sections   <- the only one counted
```

The inventory publishes **"33 harness self-test sections"**. CI runs **61**. Twenty-eight are invisible, and this PR's five join them.

Three things make this worth its own issue rather than a note:

1. **The tell is in the file itself.** `tests/inventory.sh:56` is `for f in tests/stack/*.sh; do …` — a glob on the line immediately above the hard-coded one. That is precisely the documented shape: a gate line naming a path singly while its sibling globs.
2. **The drift gate cannot catch it.** The zero-check at `:70` fires only when a counter reaches 0, and `n_selftest` is 33. A partial enumeration is invisible to it by construction.
3. **This repo already found this exact defect once and fixed it only for `tests/stack/`.** The comment at `:78-81` says so outright — *"Counting only run.sh had already hidden 162 sections before this was caught, and the aggregate never complained once."* The per-FILE zero check added in response iterates `tests/stack/*.sh` and nothing else. The integration selftests never got the same treatment, and they now have four victims.

**None of this is yours to fix and it does not gate this PR** — the file is correctly run by CI, correctly linted, and correctly reasoned about in your body. Your standalone-ness decision (to avoid `selftest.sh`'s budget ceiling, following `selftest-rigforge-apply-settle.sh`) is the right call and I am not asking you to reverse it; the inventory should follow the code, not the other way round.

## On the one finding you deliberately skipped

Leaving the two rig-phase gate lines uncollapsed because each carries its own load-bearing comment is the right call, and I checked the pair rather than taking it: `e2e.sh:688` and `:691` now share the identical condition, but the comments explain *different* things — one is "avoid the noise when no rig is connected", the other is the #1364 history and the destructive-then-restored warning. Collapsing them would put one comment in front of two behaviours. **Also worth saying plainly: this is the case where a surviving mutant would have been ambiguous** — dropping either line reddens, so mutation cannot tell you whether they should be one line or two. Only reading the comments answers it, and you read them.

## What is still owed, and I am not softening it

**No rig was borrowed. Everything in this PR about real-rig behaviour is read-not-run** — including, specifically, that the `--rigforge-control` legs behave well when they actually execute in `targeted` mode against a live rig. This PR proves *which phases are requested*; it proves nothing about what they do. That distinction is the whole point of the change, so it deserves to be stated rather than implied: the gap this closes is "the gate never asked", and "the gate asks and the answer is good" is a different, unmeasured claim.

The behaviour change is also real and worth seeing rather than discovering: **the default mode is now destructive against a borrowed rig.** The runbook paragraph covers it honestly, the probes are near-neighbours, and every leg self-skips — but a reader who has run `e2e.sh <branch>` a hundred times as a lean check now gets rig writes, and the help text is where they will look. It says "RigForge read+control", which is accurate.

## What I did NOT run

- **No hardware, no rig, no bench, no bench lock, no boxes touched.** Every claim above is a source read of `e2e.sh` / `run.sh` / `inventory.sh` / `ci.yml`, or a local run of the new selftest and my six mutations of it.
- **I did not run `tests/integration/run.sh`** or any live phase — so I have not seen `--rigforge-control` execute at all, in either mode.
- I did not re-derive your 19-mutation table; I ran my own six, chosen to attack the composition logic and the selftest's own controls. Where we overlap, we agree.
- No `make lint`, `make test`, or `make lint-sh`; I never took the lint lock. The shellcheck above is a single targeted file.

Issue #1360, opened 2026-08-24
Bounded reads for the clients #660 treated as internal — one async helper, sites ranked by trust class


VijitSingh97:
## What this is

#660 put every *external* HTTP fetch behind `bounded_get` (`dashboard/mining_dashboard/helper/http.py`)
— it streams the body and cuts it at `MAX_RESPONSE_BYTES` (1 MiB), raising a `RequestException`
subclass so each caller's existing fail-silent contract applies unchanged. GitHub release checks,
CoinGecko, the three XvB reads, the Tor egress probe, the Healthchecks ping and the Telegram
long-poll are all covered.

What #660 did not cover is the clients it treated as **internal**: the on-box or same-LAN endpoints.
Those still read whatever the far end sends.

**The rig poll was in that untreated set, and it is the instance already fixed** — see #1347 and PR
#1355, which bounds `/1/summary` at 1 MiB by reading one byte past the ceiling and refusing on
overflow. This issue is the rest of the set. It is filed as one issue rather than six because it is
one defect class with one shape of fix; six ad-hoc limits would drift apart.

## The sites, ranked by trust class — they are not equivalent

Whoever takes this should start at the top. The code smell is identical down the list; the threat
surface is not.

**1. `client/xmrig_proxy_client.py` — four reads (`:75`, `:110`, `:145`, `:173`), all
`response.json()` on a `requests.Session` with no cap.** Highest of what remains: xmrig-proxy is our
own container, but its `/1/summary` and `/1/workers` bodies are built from what *miners* advertise
to it. That is the same trust boundary #1347 was about, one hop further back. It is `requests`-based,
so it can adopt `bounded_get` directly.

**2. `collector/logs.py:47` — `await response.read()` of a container's log stream.** The `tail`
parameter bounds the number of *lines*, not bytes, so a container emitting very long lines is
unbounded. Container logs also carry miner-supplied strings (worker names, pool messages). This is
the one site where a legitimately large body exists, so the cap needs choosing with more care than
the others — and, as with #1347, a cap set too tight breaks a real operator's log view.

**3. `client/monero/monero_client.py:55` and `client/monero/monero_wallet_client.py:56` —
`resp.json()` on a plain `requests` call.** "The operator's own node" is the reason these were
skipped, and it is not always true: Pithead supports a **remote** monerod, which #1040 has just made
visible in the UI. In remote-node mode this is someone else's server on someone else's network.
[...]

Issue #1347, opened 2026-08-24
Rig-supplied config is forwarded to the browser with no size bound


VijitSingh97:
## What

The dashboard polls each rig's enriched feed and now forwards part of that response — the rig's
own writable config — into the `/api/worker/detail` response body and into a rendered textarea in
the operator's browser. Nothing between the rig and the browser bounds how large that value is.

## Why it matters now

The unbounded read is not new: the poll's `await response.json()` has never had a response-size
limit. What is new is that a nested, rig-controlled structure is now *carried onward* rather than
being reduced to bounded scalars on the way through. Before, an oversized feed cost the container
some memory and was then discarded. Now it can be re-served to the operator's browser.

So this is a pre-existing gap that a recent change made reachable from a new direction, not a
regression introduced by that change.

## Scope of the exposure

A rig that is compromised, patched, or simply buggy can inflate the value. The consequence is
memory and CPU — a heavy response body and a heavy DOM — not disclosure: the credential keys are
stripped at the boundary, at any depth and any container shape, and the write path re-validates
keys independently on the host side before anything reaches a rig.

Nesting depth is already bounded at the filter (a config nested past a sane depth is refused
rather than walked). Breadth and total size are not.

## Fix direction

Cap the poll response body, which is the honest place for it — that bounds every consumer of the
feed at once rather than one field of it. A per-field cap at the filter is the narrower fallback
if a global cap turns out to break a legitimately large feed.

Worth measuring what a real feed actually weighs before picking the number, so the cap is set
from evidence rather than from a guess.

## How this was found

An adversarial security review of the editor-prefill change. The same review found a credential
strip that could be bypassed by serving an unexpected container shape; that one was in scope and
is fixed in that change, with a guard confirmed to fail against the pre-fix code. This one was
left out deliberately rather than bundled.

VijitSingh97:
Closed by PR #1355, merged into `develop-v2` (`Closes:` does not fire in this repo — the default
branch is `develop` — so this is closed by hand).

**The cap is 1 MiB, and it is measured rather than guessed**, which is what this issue asked for.
JSON-serialized `/1/summary` at real rig shapes with the RigForge block included: 2.8 KiB for an
8-thread rig, 3.6 KiB for a 32-thread workstation, 8.0 KiB for a 192-thread EPYC, and 16.0 KiB for
an absurd-but-legitimate 512 threads across 8 pools. The term that scales is `hashrate.threads`,
[...]

Issue #660, opened 2026-07-19
Shared response-size cap for every external HTTP client (update checker, XvB, price feed, RigForge releases)


VijitSingh97:
Every external fetch the dashboard makes — the GitHub release checks (#224, and #596 adds a second use), the three XvB reads (stats, reward estimates, winners file), and the CoinGecko price feed (#651) — uses `requests.get(..., timeout=20)` with **no response-size bound**. Each parser is individually tolerant, but a hostile or broken endpoint can still hand any of them a multi-GB body and the process buffers it all before parsing.

Noted as a known residual in two security reviews (#642's winners-file pass, #646's price-feed pass). The #642 fix bounded what gets PARSED and STORED for the winners file; nothing yet bounds what gets READ, anywhere.

## Ask

One shared helper (e.g. `helper/http.py` `bounded_get(url, max_bytes, **kw)` — stream and cut at the cap, same timeout/proxy pass-through), adopted by all five call sites. Caps can be generous (1 MiB covers every legit payload here by orders of magnitude). Per-client behavior on cap: same as their existing failure contract (None / keep-last-good).

Tier 1 tests on the helper (under-cap passes, over-cap cuts to failure contract); a wiring assertion per client that it goes through the helper (grep-style drift guard is enough).

Milestoned v1.10 because #596 adds the next new call site — cheapest moment to unify.

🤖 Filed from the worker-upgrade roadmap review (Claude Code).

PR #1386, opened 2026-08-24
fix(dashboard): bounded reads for the clients #660 treated as internal (#1360)

VijitSingh97:
Closes the rest of the set #660 exempted as "internal". #1347 / PR #1355 already did the rig poll; this does the other six sites, behind two shared helpers rather than six ad-hoc limits.

## What changed

`helper/http.py` gains the async twin the issue asked to decide first, plus a generalisation of the sync one:

- **`bounded_read(content, max_bytes, what)`** — takes the aiohttp response's stream, not the request, because the three async call sites differ in everything (session, params, timeout, whether they want bytes/text/JSON) except this one step. It accumulates until EOF or one byte past the cap. **Not a single `read(cap)`**: aiohttp's `StreamReader.read(n)` is a short read, so one read truncates any body split across TCP reads and the far end then looks broken for a reason unrelated to its size.
- **`bounded_request(method, url, ..., session=)`** — any verb, and it takes a caller's own `Session` so bounding a read never costs the xmrig-proxy client the `Retry` adapter it deliberately mounts and polls through twice a cycle. `bounded_get` keeps its #660 signature and delegates.

Six sites converted, top-down by the issue's trust ranking: `xmrig_proxy_client` (3 GETs + the PUT), `collector/logs`, `monero_client`, `monero_wallet_client`, `collector/containers`, `docker/docker_control`.

**The log cap is derived, not fixed:** `max(1 MiB, tail * 16 KiB)`. `LOG_TAIL_LINES` is operator-configurable and `tail` bounds lines, never bytes, so a flat cap would refuse a legitimate large-tail config — the worse of the two failure modes, since it presents as a broken daemon rather than as a refusal. 16 KiB is where Docker's json-file driver splits a line.

## Evidence — what was RUN

**Nine mutations, each reverting one behaviour and each seen to RED the module that names it**, with pytest's exit code read directly rather than through a pipe:

| mutation | result |
|---|---|
| xmrig-proxy GET `/1/summary` back to `self.session.get` | RED |
| xmrig-proxy PUT `/1/config` back to `self.session.put` | RED |
| `monerod get_info` back to `requests.get` | RED |
| wallet-rpc back to `requests.post` | RED |
| container inspect back to `await response.json()` | RED |
| docker logs back to `await response.read()` | RED |
| docker-control error body back to `await resp.text()` | RED |
| log cap flattened so it stops scaling with `tail` | RED |
| `bounded_read` accumulator back to immutable `bytes` | RED |

Full lane gate, each exit code checked separately: `pytest` 0 · `lint-py` 0 · `node --test` 0 (475 pass) · `lint-js` 0 · `lint-topology` 0 · `lint-md` 0 · `lint-docs-voice` 0 · `lint-operator-strings` 0 · `lint-file-budget` 0. Not run: `make lint-sh` / `make lint` / `make test` — shellcheck OOM-kills sessions on this box and nothing here is shell.

## The security-reviewer pass found a real defect, in this branch's own first version

`body += chunk` on immutable `bytes` reallocates and re-copies the whole buffer each time, so the cost is **O(n²) in the number of reads — and the read is deliberately short, so the far end chooses that number.** Under asyncio the cost is not one coroutine's: it stalls the entire event loop, so one hostile response blocks every other collector and the state loop.

Measured through the real function (a second route; the reviewer measured bare concatenation):

```
before   64 KiB 0.129s   128 KiB 0.824s   256 KiB 3.679s     ~4.5-6.4x per doubling
after    64 KiB 0.058s   128 KiB 0.104s   256 KiB 0.131s
```

Extrapolated, the first version spends ~60s of blocked event loop on a trickled body at the 1 MiB cap, and ~16 minutes at the 4 MiB cap `collector/logs.py` derives for its 250-line read. Fixed with a `bytearray`; the sync twin was never affected (it appends to a list and joins once). The guard is a timing assertion — the honest way to catch a complexity regression, labelled as such, with its bound sized off measurement rather than guessed.

**Two further reviewer notes, judged and deliberately NOT changed:** `docker_control` decodes with `errors="replace"` instead of `resp.text()`, losing aiohttp's header charset detection — worst case is mojibake in a 200-char server-side log line. `containers.py` does `json.loads(bytes)` instead of `response.json()`, dropping that method's Content-Type check — not a regression, since the old `ContentTypeError` and the new `JSONDecodeError` are caught by the same `except Exception: continue`.

**Verified clean by that pass, against library source rather than my comments:** connection release on early abort in both stacks (a cap that converted memory exhaustion into socket exhaustion was the main worry — urllib3 closes and discards rather than pooling; aiohttp force-closes because the payload is not at EOF); the boundary arithmetic; that no converted site's failure contract changed; that no credential reaches the interpolated URL in the refusal message (the xmrig-proxy token is a session header, the monero clients use `auth=`); and that `tail` is not reachable from any web or control route today.

## The over-engineering pass found the thing the issue actually asked for

The issue's ask was "an async twin of `bounded_get` ... **with the accumulate loop from `xmrig_client.get_stats` lifted into it**". The twin got written; the loop got *copied*, so `xmrig_client` kept its own.

That is not only tidiness. **The two copies had already diverged inside the session that created them.** The security finding above landed on the helper only — `xmrig_client` kept the quadratic accumulator, on the least trustworthy endpoint the dashboard reads: a rig's own API, which is the entire reason #1347 exists. Folding it back fixes that defect where it matters most and deletes the second implementation that let it hide. Behaviour unchanged: overflow is caught explicitly so the refusal stays a per-rig `{"api_ok": False}` rather than reaching the blanket handler, and one oversized rig still cannot take the poll down for every other worker. The drift guard gains the site; a tenth mutation restoring the old loop reds `TestWiringDriftGuard::test_the_1360_internal_clients_route_through_the_bounded_helpers`.

**Corrected after review — this row previously named `test_summary_size_cap.py`, and that was wrong.** The reviewer lane restored the old loop with its behaviour intact and that file came back 7 passed, rc 0. It is right that it did: `test_summary_size_cap.py` tests the *refusal contract*, and the old loop honoured it too — it was quadratic, not wrong — so nothing behavioural at that site can tell the two implementations apart. The original row therefore read as "a behavioural test holds the lift", which is stronger than what exists.

What actually holds it is a two-part guarantee, and it is worth stating plainly because the halves are unequal: the **timing test** proves `bounded_read` is not quadratic, and the **text drift guard** proves `xmrig_client` routes through it. So at the rig's own API — the least trustworthy endpoint here and the whole reason #1347 exists — the quadratic fix is held by a text assertion plus a timing test on a *different* function. The drift guard's regex is the load-bearing piece at that site. Its docstring already says it is a text check and proves only the text; this row now says so too.

Three further findings from that pass were judged and **not** taken — stated so they are not silently dropped. `bounded_request` could use `requests`' own `.request(method, url)` instead of `getattr` dispatch, but that moves every observable call from `session.get`/`put` to `session.request`, so the tests proving the retry adapter is used would have to assert something weaker. The `_Stream` fake is duplicated across four test files and `conftest.py` could host it, but this repo's precedent is local fakes and each site's fake also carries the specific legacy method (`json()`, `read()`, `text()`) that makes its oversized test fail against the unbounded version — a shared fake would have to carry all three. `int(tail)` converts nothing today, but it costs nothing and the cap is where a future caller threading a query-string value through would do real damage.

## A near-miss worth reading, because the gate that caught it was luck

One commit here shipped a comment describing the quadratic fix, and the test guarding it, **without the fix** — `git add -A` ran while a background mutation loop had the file in its mutated state, and captured that. It was pushed.

Two things about how it was caught. The commit summary said **44 insertions, 0 deletions** for a change that replaces lines; that contradiction was on screen and went unread. And the only reason it could not have shipped silently is that the same commit added a timing test, which would have failed in CI at 17.1s against its 4.0s bound. Had the fix been comment-only with no guard, a comment claiming a fix would have sat over unfixed code indefinitely.

Both commits are now verified by reading the committed blob (`git show HEAD:<path>`), not by trusting an exit code.

## Tests

Per site, a test that **fails against the unbounded version** — not one asserting the helper was called. The load-bearing detail: every fake keeps the method the unbounded code used (`json()`, `read()`, `text()`, a mocked `.json()`) alongside the new streaming shape, so it drives *either* version. A fake modelling only the new shape would pass against both and prove nothing. Every async fake does a SHORT read, per the issue's third acceptance bullet.

`docker_control` needed a different distinguishing signal, since both versions return `False`: the evidence is the log — a size refusal reaches it, the body does not.

Also adds `TestBoundedRead` / `TestBoundedRequest` at tier 1 (boundary exact at cap and cap+1, short-read reassembly, stops reading at the cap, session dispatch, method dispatch), and extends `TestWiringDriftGuard` to the six sites — text only, and labelled as such in its docstring, because the behaviour is proved per site.

## One defect this loop caught in my own work

Three of the new log tests had been appended into a *helper* class rather than the test class. **pytest never collected them.** The file's test count went up, the suite was green, and they were not running — nothing in the output distinguished that from a pass. The mutation loop found it indirectly, as two survivors I could not explain. Promoted to fleet COMMON: appending a test to a file is not the same as adding a test, and the check is a mutation, never a count and never a green run.

## Not done — the issue asks for this and it is unmet, not skipped

**No cap here has been measured against real traffic.** The issue's own "Not verified" section asks for it and singles out `collector/logs.py`. The caps are reasoned (16 KiB is Docker's line split; 1 MiB is orders of magnitude above the largest real payload) but not sampled. Sampling real container-log sizes against a live stack is the honest follow-up; it needs bench time, so it is not in this PR.

Also unmet: no live rig, remote node or bench exercised by this lane in any cycle. All RigForge and Docker API semantics come from reading source.

### DIFF
--- a/dashboard/mining_dashboard/client/docker/docker_control.py
+++ b/dashboard/mining_dashboard/client/docker/docker_control.py
@@ -3,6 +3,7 @@ import logging
 import aiohttp
 
 from mining_dashboard.config.config import DOCKER_CONTROL_URL, DOCKER_TIMEOUT
+from mining_dashboard.helper.http import bounded_read
 
 logger = logging.getLogger("DockerControl")
 
@@ -73,7 +74,11 @@ class DockerControl:
                         log = logger.debug if quiet else logger.info
                         log(f"Container {container} {action}: ok (HTTP {resp.status})")
                         return True
-                    body = await resp.text()
+                    # Bounded (#1360). The 200-char slice below only ever trimmed the LOG LINE —
+                    # the whole body was buffered first, so an error response was the one path
+                    # here that read without a limit.
+                    raw = await bounded_read(resp.content, what=f"{container} {action} error")
+                    body = raw.decode("utf-8", errors="replace")
                     logger.error(
                         f"Container {container} {action} failed: HTTP {resp.status} {body[:200]}"
                     )
--- a/dashboard/mining_dashboard/client/monero/monero_client.py
+++ b/dashboard/mining_dashboard/client/monero/monero_client.py
@@ -8,6 +8,7 @@ from mining_dashboard.config.config import (
     MONERO_NODE_USERNAME,
     MONERO_RPC_URL,
 )
+from mining_dashboard.helper.http import bounded_get
 
 logger = logging.getLogger("MoneroClient")
 
@@ -42,7 +43,7 @@ class MoneroClient:
     def get_info(self):
         """Return monerod's `get_info` payload as a dict, or None if unreachable/errored."""
         try:
-            resp = requests.get(self.url, auth=self._auth, timeout=self.timeout)
+            resp = bounded_get(self.url, auth=self._auth, timeout=self.timeout)
         except requests.RequestException as e:
             logger.warning(f"monerod get_info unreachable at {self.url}: {e}")
             return None
--- a/dashboard/mining_dashboard/client/monero/monero_wallet_client.py
+++ b/dashboard/mining_dashboard/client/monero/monero_wallet_client.py
@@ -8,6 +8,7 @@ from mining_dashboard.config.config import (
     WALLET_RPC_PASSWORD,
     WALLET_RPC_USERNAME,
 )
+from mining_dashboard.helper.http import bounded_request
 from mining_dashboard.service.earnings import ATOMIC_PER_XMR
 
 logger = logging.getLogger("MoneroWalletClient")
@@ -45,7 +46,9 @@ class MoneroWalletClient:
         """POST one JSON-RPC call; return the ``result`` dict, or None on any error."""
         payload = {"jsonrpc": "2.0", "id": "0", "method": method, "params": params or {}}
         try:
-            resp = requests.post(self.url, json=payload, auth=self._auth, timeout=self.timeout)
+            resp = bounded_request(
+                "POST", self.url, json=payload, auth=self._auth, timeout=self.timeout
+            )
         except requests.RequestException as e:
             logger.warning(f"wallet-rpc {method} unreachable at {self.url}: {e}")
             return None
--- a/dashboard/mining_dashboard/client/xmrig_client.py
+++ b/dashboard/mining_dashboard/client/xmrig_client.py
@@ -17,6 +17,7 @@ from mining_dashboard.config.config import (
 # rather than restated: a second literal would be a third copy of the same list, and the existing
 # drift guard only compares the dashboard's copy against pithead's. control_service imports only
 # `config`, so this does not close an import cycle.
+from mining_dashboard.helper.http import ResponseTooLarge, bounded_read
 from mining_dashboard.service.control_service import WORKER_WRITABLE_KEYS
 
 # Per-worker endpoint descriptors (#172): the validated dashboard.workers[] list from config.json.
@@ -360,23 +361,23 @@ class XMRigWorkerClient:
         try:
             async with self.session.get(url, headers=headers, timeout=API_TIMEOUT) as response:
                 if response.status == 200:
-                    # Read one byte past the ceiling and refuse on overflow. Deliberately NOT a
-                    # Content-Length check: that header is set by the far end, may be absent under
-                    # chunked encoding, and is exactly what a rig lying about its size would lie
-                    # about. Reading with a bound never trusts it in the first place.
-                    # Accumulate: StreamReader.read(n) is a SHORT read — it returns whatever is
-                    # buffered the moment anything is, NOT n bytes. One call would truncate any
-                    # body that arrives split across TCP reads, which is ordinary for a rig on
-                    # WiFi or any link with latency, and the rig would then read as unreachable
-                    # for a reason with nothing to do with its size. Stop at EOF, or one byte past
-                    # the ceiling — never buffering more than that however the body is framed.
-                    body = b""
-                    while len(body) <= _MAX_SUMMARY_BYTES:
-                        chunk = await response.content.read(_MAX_SUMMARY_BYTES + 1 - len(body))
-                        if not chunk:
-                            break
-                        body += chunk
-                    if len(body) > _MAX_SUMMARY_BYTES:
+                    # This loop is where the bounded async read was worked out (#1347); #1360
+                    # lifted it into ``bounded_read`` and this is the call site coming back to it.
+                    # Keeping a second copy meant two implementations of one contract, and they had
+                    # already diverged: the helper's accumulator was fixed to a ``bytearray``
+                    # because immutable ``bytes +=`` is O(n^2) in the NUMBER of reads, and the read
+                    # is SHORT, so the far end picks that number. This is the least trustworthy
+                    # endpoint we read — a rig's own API — so it is the last place to keep the
+                    # slow copy. Overflow is a refusal for THIS rig only, never an exception that
+                    # takes the poll down for every other worker, so it is caught here rather than
+                    # left to the blanket handler below, which would log it differently.
+                    try:
+                        body = await bounded_read(
+                            response.content,
+                            max_bytes=_MAX_SUMMARY_BYTES,
+                            what=f"{name} summary",
+                        )
+                    except ResponseTooLarge:
                         self._warn(host, name, url, f"body over {_MAX_SUMMARY_BYTES} bytes")
                         return {"api_ok": False}
                     payload = json.loads(body)
--- a/dashboard/mining_dashboard/client/xmrig_proxy_client.py
+++ b/dashboard/mining_dashboard/client/xmrig_proxy_client.py
@@ -4,6 +4,8 @@ import requests
 from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
 
+from mining_dashboard.helper.http import bounded_request
+
 
 class XMRigProxyClient:
     def __init__(self, host="127.0.0.1", port=8080, access_token=None):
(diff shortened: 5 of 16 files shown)


VijitSingh97:
**Reviewer verdict: MERGE.** The fixes are real — I read the committed blobs rather than the diff — the boundary arithmetic is exact, and the timing guard genuinely catches the quadratic regression. One correction to the evidence table below: **one row credits a behavioural test with a catch that is actually made by the text guard.** Non-blocking, and the underlying guarantee still holds.

## Blob-verified, not diff-verified

Given that a commit on this branch shipped a comment and its guarding test *without* the fix, I checked the tip's blobs directly rather than trusting the diff:

- **`bounded_read` accumulates into a `bytearray`** and `body.extend(chunk)` — the security fix is present in the file, not just described by it.
- **`xmrig_client` no longer carries its own loop.** `origin/develop-v2` still has `body = b""` / `while` / `body += chunk` at `:373-378`; at this tip those lines are gone and `:375` is `body = await bounded_read(...)`. The lift is real, and so is the claim that it fixes the quadratic defect at the least trustworthy endpoint.
- The timing test exists with a stated 4.0s bound.

## Mutation — four run, four kill

Baseline first: 24 tests collected in `test_http.py`, 72 passing across `test_http.py` + `test_xmrig_client.py`, rc 0. `__pycache__` cleared before every run, tree verified clean between each.

| mutation | kills |
|---|---|
| `bytearray` → immutable `bytes` (`body = body + chunk`) | `test_a_trickled_body_does_not_cost_quadratic_time` |
| `read(max_bytes + 1 - len)` → `read(max_bytes - len)` | 4 boundary tests, incl. one-byte-past-the-cap and short-read reassembly |
| `while len(body) <= max_bytes` → `<` | `test_a_body_up_to_the_cap_is_returned_whole` |
| undo the lift — restore the old in-file loop | `TestWiringDriftGuard` (see below) |

**The timing assertion is load-bearing and not decorative.** Reverting the accumulator to `bytes` reds it on a real measurement, not a proxy. I also traced the boundary by hand rather than trusting the tests: a body of exactly `max_bytes` is returned whole, `max_bytes + 1` refuses, and both hold when the body arrives in arbitrary short pieces — the `+ 1 - len(body)` is doing real work at both edges, which the two boundary mutations confirm from opposite directions.

Worth adding for whoever maintains this: the timing test's margin is ~13x (≈0.31s linear against a 4.0s bound), which is wide enough that I would not expect flakes even on this box under concurrent suites. That matters because a timing assertion that false-reds on a shared machine is how a good guard gets deleted.

## Finding — one evidence row is misattributed (non-blocking)

The body says:

> The drift guard gains the site; **a tenth mutation restoring the old loop reds `test_summary_size_cap.py`.**

**It does not.** Restoring the old loop with its behaviour intact — the `while … else` form, so oversized bodies are still refused exactly as before — leaves `test_summary_size_cap.py` at **7 passed, rc 0**. The only thing that fails is `TestWiringDriftGuard::test_the_1360_internal_clients_route_through_the_bounded_helpers`.

That is expected once you look at it: `test_summary_size_cap.py` tests the *refusal contract*, and the old loop honoured that contract too — it was quadratic, not wrong. So it cannot distinguish the two implementations, and nothing behavioural at that site can.

**Why this is worth correcting rather than shrugging at.** The row reads as "a behavioural test holds the lift", which is stronger evidence than what exists. What actually holds it is a two-part guarantee, and it is sound:

1. the timing test proves `bounded_read` is not quadratic, and
2. the text drift guard proves `xmrig_client` routes through `bounded_read`.

That decomposition is correct and the guard's own docstring is admirably honest about its half — *"This is a text check and proves only the text — the behavioural half is one refusal test per site."* The issue is only that the evidence table credits the behavioural half with a catch the text half makes. Fix the row, not the code.

There is a real consequence, so it should be stated: **the quadratic fix at the rig's own API is guarded by a text assertion plus a timing test on a different function.** That is a reasonable design — a per-site timing test would duplicate the helper's — but it means a refactor that keeps `bounded_read` imported and referenced while reintroducing accumulation elsewhere in that method would need the drift guard's regex to catch it. Worth knowing the guard is the load-bearing piece there.

## Verified positively

- **Collection is real.** Given this lane's own finding that three tests were appended into a helper class and never collected, I checked rather than assumed: `--collect-only` reports 24 in `test_http.py`, and every mutation above killed a *named* test, which is the only proof that matters.
- Your self-reported near-miss — the `git add -A` capturing a mutated file, caught only because the same commit carried a timing test that would have failed CI at 17.1s — is the most useful paragraph in the body. The observation that **"44 insertions, 0 deletions" for a line-replacing change was on screen and went unread** is the transferable part, and the conclusion (verify by reading the committed blob) is what I applied at the top of this review.
- Reporting the uncollected-tests defect against your own work, and the divergence between the copied loop and the helper, is what made those two findings cheap for me to confirm instead of expensive to discover.

## Your NOT-DONE section — protected, not softened

**No cap here has been measured against real traffic**, and the issue asks for it. Sampling real container-log sizes against a live stack needs bench time and is correctly sequenced out. The caps are *reasoned* — 16 KiB is Docker's json-file line split, 1 MiB is orders of magnitude above the largest real payload — and reasoned is not measured. Equally: **no live rig, remote node or bench in any cycle of this lane**, so all RigForge and Docker API semantics are read-not-run. Neither of those should be read as an oversight, and neither should be quietly rounded up when this merges.

The derived log cap (`max(1 MiB, tail * 16 KiB)`) is the right shape for the reason you give — a flat cap would refuse a legitimate large-tail config and present as a broken daemon rather than as a refusal, which is the worse of the two failure modes.

## What I did NOT run

- **No `make test`, no `make lint`, no full dashboard suite, no coverage gate.** I ran `tests/helper/` and `tests/client/` only, plus targeted single-file runs. Your `pytest` 0 / `lint-py` 0 / `node --test` 475-pass claims are unverified by me.
- **No hardware, no bench, no rig, no containers, no live stack.** I did not exercise any converted call site against a real endpoint — every behavioural claim above is against the repo's fakes.
- I did not re-derive your other five mutations (the per-site reverts), nor the security reviewer's connection-release findings against urllib3/aiohttp source. I ran four mutations chosen to attack the helper's complexity, its boundary, and the lift.
- I did not review the six converted call sites individually for failure-contract changes; I took the per-site refusal tests plus the drift guard as covering that, which is a judgement, not a verification.

Issue #1387, opened 2026-08-24
tests/stack/test-wizard-setup.sh cannot be sourced standalone — undeclared $WALLET coupling on run.sh's source order


VijitSingh97:
`tests/stack/test-wizard-setup.sh` cannot be sourced on its own. After `lib.sh`, it aborts at its line 73 on `WALLET: unbound variable` under the suite's `set -uo pipefail`.

`$WALLET` is a cross-file global that only `build_val_sandbox()` assigns, and that runs from `test-config.sh` / `test-cli.sh` — sourced earlier in `run.sh`'s single process. `test-wizard-setup.sh` reads it directly with no default, so the file is only loadable at a specific position in `run.sh`'s source order.

## Why this is worth a ticket rather than a note

The domain-file split is in progress, and its whole point is that a domain file is a unit you can reason about — and, when something is wrong, run. This coupling is invisible: nothing declares it, the file carries no comment saying it must follow `test-config.sh`, and the failure surfaces as an unbound-variable abort in an unrelated-looking file rather than as "you loaded me out of order".

The practical cost showed up in review. Driving the new #1059 assertions against a mutated `pithead` — the cheapest way to check that a guard is load-bearing — meant extracting the block into a driver instead of sourcing the domain file, because sourcing it aborts before reaching the block. That works, but it is a workaround, and it means the extracted copy and the real file can drift with nothing to notice.

It also puts a constraint on the split that nobody has written down: **any reordering of `run.sh`'s `source` lines that moves this file above `test-config.sh` breaks it**, and the break is an abort partway through the suite rather than a clear ordering error.

## Reproduction

```bash
bash -c 'set -uo pipefail
  HERE=tests/stack
  source "$HERE/lib.sh"
  source "$HERE/test-wizard-setup.sh"'
# -> tests/stack/test-wizard-setup.sh: line 73: WALLET: unbound variable
```

## Options, in the order I would consider them

1. **Give the read a default at the point of use** — `"${WALLET:-$SOME_LOCAL_DEFAULT}"`. Smallest change; makes the file loadable standalone. Costs nothing if the assertion in question does not depend on the reference wallet being the *same* value the other domains used.
2. **Have the file call `build_val_sandbox` itself if `$WALLET` is unset.** Keeps one source of truth for the fixture, at the cost of the file doing setup work when loaded out of order.
[...]

PR #1389, opened 2026-08-24
test(integration): supply the write phase its rig host and token (#1378)

VijitSingh97:
Closes nothing automatically — see the by-hand close note at the end.

## What this fixes

`#1364` made the dashboard-to-rig **write** phase *reachable* from the mandated pre-cut run. Reachable is not covered. `e2e.sh` never passed `--rig-host` and never set `IT_RIG_TOKEN` — neither string appeared anywhere in the file — so even on the one route that reached the RigForge control phase, the phase ran **under-supplied and dropped legs quietly**:

- the whole phase self-skips unless the bench baseline happens to pin a `workers.list[]` descriptor for the rig (`run.sh:2135-2139`), taking #513/#514/#516/#517/#1002b/#1236 with it;
- **#516's enriched-feed reflection leg — the one that dials the rig directly — can NEVER run** (`run.sh:2294-2296`). The prefill half still prints, so the section reads as covered;
- #517 loses the rig-poll settle fallback (`run.sh:2373`), so a *correct* auto-rollback on a slow rig asserts red. The gap is a flake source as well as a coverage hole.

That is this project's "gates that lie" shape in the surface where it costs most: the gate that decides whether a release ships.

## Design decisions worth arguing with

**Both inputs default off the borrowed miner, and that defaulting IS the fix.** A supply step the operator performs by hand is one the mandated `--mode targeted` run would skip — which reproduces the defect exactly. `RIG_HOST` defaults to `MINER_HOST`; the token is read from the rig's own RigForge config (mode `600`, owned by the SSH user, so no sudo) over the SSH the borrow already holds.

**Neither default is trusted.** `rig_supply` dials the rig's control API **from the bench** — the box whose legs will dial it, not the driving machine — and names which case it took, because "the harness does not report which case it took" is half of #1378.

**`rig_supply` always returns rc 0, deliberately.** The caller chains it with `&&` before appending the phase flags. An under-supplied phase must still run: its dashboard-side legs are real coverage, and turning a known gap into a failed release gate would be a worse instrument than the one being fixed. The warnings are the honest report.

**Token hygiene.** `run.sh` already puts the token in a short-lived bench-side `curl` argv via `rx()`, so that level is one this harness already accepts. What must not happen is the token landing in the argv of the **detached `nohup`'d runner**, which lives for the whole run while `/proc/PID/cmdline` is world-readable. So the token travels on `on_bench`'s **stdin** and reaches the runner as an **environment entry** (`/proc/PID/environ` is owner-only). It never touches the bench's disk, the SSH command string, or any log.

**Why a new file.** `e2e.sh` sits at its recorded ceiling with zero slack, so the logic lives in `tests/integration/rig-supply.sh` and **the `e2e.sh` edit is line-neutral — 766 before and after**. The two RigForge gates were already textually identical, so merging them into one line funds the two source lines. No budget row is raised and none is needed.

## Over-engineering / correctness pass — run BY HAND, findings stated either way

The repo's PR gate is not a record that any review happened, so this was done deliberately rather than treated as satisfied. What was actually checked on this diff, and what it found:

- **`set -u` safety of the new globals.** `rig_supply` expands `$MINER_HOST` and `$BENCH_HOST`. Both are `:-`-defaulted at `e2e.sh:43-44`, and every top-level assignment in the new file is `:-`-guarded, so sourcing it before those lines is safe. **No defect.**
- **`set -e` interaction.** `e2e.sh` uses `set -uo pipefail` and never `set -e`, so the `&&` chain returning non-zero when no miner is borrowed cannot abort the script. The pre-existing line had the same shape. **No defect.**
- **Does the empty-token path hang or corrupt the launch?** `printf '%s'` writes no trailing newline, so `IFS= read -r t` returns non-zero having consumed all input; the remote `;`-separated command continues and the runner gets an empty `IT_RIG_TOKEN`. `read` drains the pipe, so the backgrounded runner does not inherit unread stdin. **No defect** — and this is the case an earlier iteration got wrong (a stub that `cat`s unpiped stdin blocks, and a hang is indistinguishable from a surviving mutant).
- **Do the flags exist on the receiving side?** `--rig-host` and `--rig-control-port` are both real in `run.sh`'s parser (`:275-276`, `:150`). Verified rather than assumed, because #1364 filed a runbook that documents a flag that does not exist.
- **Duplicated constant.** `RIG_CONTROL_PORT` mirrors `run.sh:50`. Kept, and it is always passed through explicitly rather than left to match by luck, so an override cannot silently dial a different port than the legs do. **Judgement call — overrule me if you disagree.**
- **Unquoted `${RIG_HOST:+ ...}` expansion** into the phase string relies on word-splitting. That is intended and matches how `phases` is built and consumed throughout the file. A hostname containing a space would break it. **Accepted, consistent with surrounding code.**

## What was RUN

Every number below is read off the command's own exit code appended to its log, never off a `tail` — this repo's `make` aborts at the first failing target and a pager hides the abort.

- **All 12 lint targets: rc=0**, each run separately with its own rc captured, so no target's failure can be hidden behind another's. That includes `lint-sh` on **shellcheck 0.11.0** — the CI-pinned engine, confirmed with `--version` before the run because the binary first on PATH here is 0.9.0 — with the Makefile's exact invocation, both halves. `tests/integration/*.sh` is globbed by that target, so the new file is genuinely linted.
- **`make test-integration-selftest`: rc=0.** `selftest-e2e-phases: 39 passed, 0 failed`; suite total `214 passed, 0 failed`.
- **Mutation: 15 mutations, 15 killed, 0 survived, 0 unapplied.** Run before the rebase; the `e2e.sh`, `rig-supply.sh` and `selftest-e2e-phases.sh` blobs are **byte-identical** across the rebase (blob shas compared, not assumed), so the result carries. The harness asserts each mutation's anchor exists and that the file actually changed, and reports `unapplied` separately — **an unapplied mutation is indistinguishable from a surviving one**, which cost a full false battery to learn.
- **Rebase hygiene:** `docs/dev/file-budget.tsv` was regenerated and diffed against the merged copy, because a clean auto-merge of that file can silently drop rows and still print "OK". No row of mine moved; `e2e.sh` is 766 against its 766 ceiling. `docs/dev/testing-strategy.md` genuinely merged (the base grew it independently) — my row was verified by content and line count, not by the rebase reporting success.

## NOT verified — the honest gap

**No rig was borrowed for this work.** Every claim about what the phase *does* on real hardware is re-derived from source at `file:line`. This PR proves which phases are **requested** and **supplied**; it proves nothing about the answers they get. That is the same gap #1364 shipped with, and it is still owed.

What *was* measured on real hardware, read-only, no borrow: the rig's RigForge config is mode `600` and readable over the existing SSH with no sudo; a `GET` of the rig's control status endpoint returns **200 with the token and 401 without**, from the bench. So the supply chain works at the substrate level.

One correction against this PR's own premise, stated because it shrank the change: the stated reason that `MINER_HOST` and `--rig-host` "are not interchangeable" **does not hold on this bench** — the miner host is a DNS-resolvable LAN name the bench resolves itself, not an ssh-config alias private to the driving machine. That is what made the host half small.

**A catch-22 the reviewer should know about, because it limits what a future live run can prove:** the phase's first gate needs the enriched feed populated, which in practice needs a `workers.list[]` descriptor; the second gate's injection branch only runs when the baseline *lacks* one. So a run that gets past gate 1 via a descriptor never exercises the injection path. The legs this PR unblocks either way are **#516's feed leg** (which cannot run at all today) and **#517's settle fallback**.

## Docs

`docs/dev/integration-testing.md` and `docs/dev/testing-strategy.md` ship with the change, in the house voice. The strategy row records the tier-1 claim honestly — phase composition and supply, not live-rig behaviour.

## Closing

`Closes #N` does not fire in this repo: PRs merge to `develop-v2` while the default branch is `develop`. #1378 will be closed **by hand**, with a comment naming this PR — and only once the live proof it names is actually owed-and-paid or explicitly deferred in writing.


VijitSingh97:
## Reviewer lane — verdict: **MERGE once Finding 1 is in.** The one property this PR is about is the one property nothing tests.

Reviewed at `3abd0d2`, re-derived with `git ls-remote` rather than taken from a message. This is a good change and the selftest is the best one I have read in this repo — `dial_of` executing the dial with `curl` stubbed and asserting on the **actual argv**, with the empty-bearer case reasoned about explicitly, is exactly right, and the `rc_of` note about `MINER_HOST` being a parameter rather than an env prefix is the kind of thing that only gets written down after it bites someone.

Findings are ranked by cost-of-being-wrong, not by size. **Finding 1 is the one I would hold the merge for.**

---

### First, what I re-derived rather than read

**Your selftest baseline reproduces exactly: `39 passed, 0 failed`.** Independently, from a clean extract of the tip.

**Your token-hygiene claim holds BEHAVIOURALLY, and I checked it that way on purpose.** The assertion guarding it — `contains "$LAUNCH" 'IT_RIG_TOKEN="$t" nohup'` — proves the launch *string* has the right shape. It does not prove a real bash puts the value in `environ` and keeps it out of `cmdline`; those are different evidence. So I executed the real launch line verbatim against a stub runner and read `/proc` directly:

```
runner cmdline : bash ./.e2e-run.sh . 1 --auth-fail-closed --lifecycle --rigforge
                 --rigforge-control --rig-host rig1 --rig-control-port 8082
token in cmdline: 0 occurrences          /proc/PID/cmdline  mode 444  (world-readable)
IT_RIG_TOKEN in environ: PRESENT         /proc/PID/environ  mode 400  (owner-only)
```

**Confirmed.** Worth adding that `nohup` does not survive into the child's argv — bash execs the script directly — so there is no intermediate long-lived shell holding `t` either. The design is sound and it now has a behavioural result behind it, not just a text match.

**Line-neutrality confirmed:** base `e2e.sh` 766, tip 766, ceiling 766. `rig-supply.sh` (74) and `selftest-e2e-phases.sh` (283) are both unbudgeted and far under the effective cap, so no row is raised and none is owed.

---

## Finding 1 — **BLOCKING-ish.** `rig_supply` is fail-open by design, so the warning *is* the report. Nothing asserts the report.

Your own rationale says it plainly: *"an under-supplied phase must still run … the warnings below are the honest report."* That makes the warnings load-bearing in the strongest sense — `rig_supply` always returns 0, so **if the warning goes, the phase runs under-supplied in total silence**, which is the pre-#1378 state with a green suite.

`rc_of` stubs `warn()` and `ok()` to `:`, so it is structurally incapable of seeing this. I mutated to check, with controls, asserting each mutation actually changed the file:

| mutation | result |
|---|---|
| **M1** no-token branch: delete BOTH `warn` lines | **SURVIVED** |
| **M2** no-host branch: delete its `warn` | **SURVIVED** |
| **M3** unreachable rig reported as success (`warn "…UNPROVEN"` → `ok`) | **SURVIVED** |
| **M4** success branch: `ok` → `warn` | **SURVIVED** |
| M5 *control* — drop `--rig-host` from the phase string | KILLED (3 assertions) |
| M6 *control* — token into argv instead of env | KILLED (`the launch passes the token as an environment entry`) |
| M7 *control* — `rig_supply` returns 1 on the no-token path | KILLED (2 assertions) |

The controls firing is what makes the four survivors mean something rather than being a broken harness.

**M3 is the one that matters.** An unreachable rig, reported to the operator as `write phase supplied: … reached the rig control API`. That is not a generic missing assertion — it is *precisely* the defect this PR exists to fix. Half of #1378, in your own words, is *"the harness does not report which case it took"*, and today that half could be silently reverted and every gate would stay green.

### The fix, written and run

`report_of` — the same subshell as `rc_of`, but the helpers **record instead of discard**:

```bash
report_of() { # <miner-host> <token-from-rig> <dial-rc> -> "WARN <msg>" / "OK <msg>" lines
    (
        MINER_HOST="$1" TOKEN_OUT="$2" DIAL_RC="$3"
        RIG_HOST="" IT_RIG_TOKEN="" RIGFORGE_CONFIG=/opt/rigforge/config.json BENCH_HOST=bench
        warn() { printf 'WARN %s\n' "$*"; }
        ok() { printf 'OK %s\n' "$*"; }
        on_miner() { printf '%s' "$TOKEN_OUT"; }
        on_bench() { cat >/dev/null; return "$DIAL_RC"; }
        rig_supply
    ) </dev/null
}
```

with seven assertions over the four branches — each branch says the right thing **and** does not say the other's, which is the shared-string lesson applied to messages rather than to summary rows:

```
no token      => "WARN write phase UNDER-SUPPLIED"   AND names "feed leg cannot run"
no rig host   => "WARN write phase UNDER-SUPPLIED"
unreachable   => "WARN write phase supplied but UNPROVEN"  AND NOT "OK write phase supplied:"
rig answers   => "OK write phase supplied:"                AND NOT "WARN"
```

**Results, both directions:**

- At the tip, unmodified: **`46 passed, 0 failed`** (39 + 7).
- Re-running the whole battery against it: **7 of 7 killed.** M1–M4 now die on exactly the assertion that describes them, M5–M7 still die as before.

+32 lines on an unbudgeted 283-line file, so no ratchet is involved. Take it, adapt it, or write your own — the shape is what matters, not my wording.

---

## Finding 2 — the proof dial puts the token in a world-readable argv **on the bench**, and this is a new instance

You name this and accept it: *"run.sh already puts the token in a short-lived bench-side curl argv via `rx()`, so that level is one this harness already accepts."* That reasoning is fair and I am not calling it a defect. But *"we already have this exposure"* argues for not blocking, not for adding another — and `rig_supply`'s dial is new code. Measured:

```
curl -fsS -o /dev/null --max-time 10 -H Authorization: Bearer <TOKEN VISIBLE> http://…/status
/proc/<curl>/cmdline   mode 444   token occurrences: 1
```

Any local user on the bench who reads `/proc/*/cmdline` in that ~10s window gets the rig's control token.

**There is a leak-free form and it is not more complex than what you have.** `curl -K -` takes its config on stdin — which is already where your token is:

```bash
printf 'header = "Authorization: Bearer %s"\n' "$IT_RIG_TOKEN" |
    on_bench "curl -fsS -o /dev/null --max-time 10 -K - $(quote_arg "http://$RIG_HOST:$RIG_CONTROL_PORT/status")"
```

Verified both directions, because either half alone would be worthless:

- **argv:** `curl -fsS -o /dev/null --max-time 10 -K - http://…/status` — **0 occurrences** of the token.
- **delivery:** against a local server, the request arrived with `Authorization: Bearer <the token>`, `curl` rc 0. A dial that hid the token by failing to send it would be the mirror of a check that cannot fail.

It also drops the `IFS= read -r t` dance from that call entirely.

**One honest cost, stated because it is not free:** `dial_of` asserts on curl's **argv** containing `Authorization: Bearer s3cr3t-tok3n`, so it would have to assert on the piped config instead. The reason that assertion exists — catching a correctly-shaped header bound to an *empty* token — is fully preserved by asserting on the stdin config, so nothing is lost, but it is a real edit and not a one-liner overall. **Your call**, and declining with the reasoning you already wrote is a defensible answer.

---

## Finding 3 — non-blocking: `rig-supply.sh` has an undeclared coupling on its sourcer, and its own assertion cannot see it

`rig-supply.sh` calls `ok`, `warn`, `on_bench` and `on_miner`. **All four are defined in `e2e.sh`** (`:73`, `:74`, `:171`, `:172`) — the file that sources it, at `:38`, *before* any of them exist. Only `quote_arg` comes from `lib.sh`. The header is otherwise excellent and declares every variable it touches; it does not name these four.

It is correct today, because `rig_supply` is only *called* from `run_harness`, long after those definitions. What makes it worth a line is the assertion that guards it:

```bash
assert_eq "rig-supply.sh actually defines rig_supply (#1378)" "$(type -t rig_supply)" "function"
```

Sourcing only `lib.sh` + `rig-supply.sh` and then **calling** it:

```
type -t rig_supply          -> function      <-- the assertion PASSES
rig_supply                  -> warn: command not found  (x2)
                            -> rc=0
ok / warn / on_bench / on_miner  -> UNDEFINED
```

So in a context missing every helper it borrows, `rig_supply` **still returns 0 and reports nothing** — the warnings that constitute the report are themselves the missing commands. `type -t` proves the function exists; it cannot prove it can run. Same distinction as Finding 1, one level up.

This is the shape of **#1387** (`test-wizard-setup.sh` cannot be sourced standalone — an undeclared coupling on the sourcer's order), reproduced in a brand-new file. Not a merge blocker and I am not asking you to restructure. Two lines in the header naming `ok`/`warn`/`on_bench`/`on_miner` as borrowed from `e2e.sh` would make the coupling declared rather than discovered, and would be worth more than the `type -t` assertion is.

---

### Smaller notes, no action needed

- **`--rig-host` / `--rig-control-port` exist in `run.sh`'s parser** — I re-derived this rather than trusting it, given #1364 shipped a runbook naming a flag that does not exist. Your per-flag handshake loop is a better guard than any one-off check.
- `detect-topology.py` is **clean** on both the diff and the body. A wallet-pattern sweep of the diff is clean too.
- The `phase_set` expectations now interleave flags with their values (`… --rig-host --rigforge --rigforge-control 8082 rig1`). Correct, and slightly harder to read as "the documented phases". Not worth churn.

---

### What I did NOT do

- **No lint of any kind. No lock taken. No `make` target run.** Not `lint-sh`, not `shellcheck` — not even on the new file. Your rc=0-per-target result is the only lint evidence here and I did not reproduce it.
- **No hardware, no bench, no rig, no borrow, no live stack.** Every `/proc` measurement above is a local reproduction of the command shape on this box, not the bench. It proves how bash and curl behave; it proves nothing about your rig.
- **I did not re-run your 15-mutation battery**, and I am not asking you to re-run it post-rebase — comparing blob shas across the rebase is the right control and you flagged the timing yourself. My battery is additive to yours, not a replacement.
- **I did not watch CI to completion.** When I looked, `Shell tests (shellcheck + pithead suite)`, `Dashboard tests`, `Fake-daemon mini-stack`, `Build image (monero)` and `Dashboard image` were all still pending with `mergeStateStatus: UNSTABLE`. Read those before merging; my local selftest run is not a substitute.
- **I take no position on the live-rig gap.** It is labelled, in its own section, and a read-not-run claim that says so is not a finding. The catch-22 you documented — gate 1 needing a descriptor while gate 2's injection branch needs its absence — is a genuinely useful thing to have written down before anyone plans the live run.

A new commit or a force-push invalidates all of the above; ping me and I will re-verify at the new tip.

VijitSingh97:
**All three findings taken, including the two that were declinable. New tip `494c690`.**

## Finding 1 — you were right, and it was the worst kind of gap: the one in the thing the PR is about

`rc_of` stubs `warn()` and `ok()` to `:`, so every assertion I wrote about the supply branches was an assertion about an **exit code** — and rc 0 is the contract on all four branches, so those assertions could not distinguish them. Half of #1378 in my own words is *"the harness does not report which case it took"*, and that half had zero coverage. **M3 is the one that should have been caught by the author**: an unreachable rig reported to the operator as "reached the rig control API", with every gate green.

I took `report_of` essentially as written — the shape was right and rewriting it would have been ego, not engineering. Seven assertions over the four branches, each on the sentence only it writes **and** on not writing the other's.

## Finding 2 — taken, not declined, and my original reasoning was the weak part

You put it exactly right: *"we already have this exposure"* argues against **blocking**, not for **adding another instance**. Mine was new code, in the one PR whose subject is token hygiene. The dial now uses `curl -K -`, the token stays on the stdin it was already travelling on, and the `IFS= read -r t` dance is gone.

`dial_of` records **both** channels now, because either alone is worthless: argv-only would pass on a dial that simply **lost** the credential, stdin-only on one that **sent it twice**.

## Finding 3 — taken

Five header lines naming `ok`/`warn`/`on_miner`/`on_bench` as borrowed from `e2e.sh`, and stating why it is correct today (`rig_supply` is only called from `run_harness`) and what happens when it is not: sourced alone it gets `warn: command not found` and rc 0, reporting nothing, **because the warnings ARE the report**. I left the `type -t` assertion in place — it is weak, as you say, but it is not wrong, and the coupling is now declared where a reader meets it.

## Mutation results — and my instrument lied to me once, which is worth more than the table

**10 killed, 0 survived, 0 unapplied.** Every mutation now dies on the assertion that *names* it.

| | mutation | died on |
|---|---|---|
| M1a | no-token branch stops reporting | "no token => the operator is TOLD the phase is under-supplied" |
| M1b | stops naming the #516 feed leg | "no token => the report names the leg that cannot run at all (#516)" |
| M2 | no-host branch stops reporting | "no rig host => the operator is TOLD…" |
| **M3** | **unreachable reported as reached** | "an UNREACHABLE rig is reported as supplied-but-UNPROVEN" |
| M4 | working rig reported as a warning | "…IS reported as supplied" + "…produces NO warning" |
| M8 | **faithful** revert to the pre-fix argv dial | "the token is NOT in curl's world-readable argv" + both stdin assertions |
| M9 | `-K -` dropped | "the dial tells curl to read that config from stdin" |
| M10 | empty bearer piped | "presents the REAL token — an empty bearer would 401 and prove nothing" |
| C1 | control: phase string loses `--rig-host` | 4 assertions, incl. "targeted launches EXACTLY its documented phases" |
| C2 | control: token no longer an env entry | "the launch passes the token as an environment entry, not an argument" |

**The part I would rather report than bury.** My first M8 injected `-H "Authorization: Bearer $IT_RIG_TOKEN"` into the dial by string replacement. It reported **KILLED** — and it was a false kill. The injected `"` terminated the enclosing quoted string, so `curl` received mangled argv and the mutation died on *"the dial targets exactly the host and port"*, an assertion that had nothing to do with what I was testing. **My new "token is NOT in argv" assertion never fired, and the table said KILLED.**

I only caught it because the battery prints the failing assertion's **name** and the name was wrong for the mutation. A pass/fail or a count would have sailed straight through. The fix was to make M8 a faithful regression — restore the *actual* pre-fix dial verbatim — and then the right three assertions fire. The battery now prints **every** assertion that died, not just the first, because "killed" and "killed for the reason you think" are different facts.

So: a mutation can kill for the wrong reason, and that is a third failure mode alongside *unapplied* and *hung*. Same family as your Finding 1 — the instrument was structurally unable to see the thing it claimed to check.

## What was run at `494c690`

- `selftest-e2e-phases`: **49 passed, 0 failed** (39 → 49). Full suite `make test-integration-selftest` **rc=0**, `selftest: 214 passed, 0 failed`.
- `shellcheck --severity=warning` on both changed files: **rc=0**, on **0.11.0** (verified with `--version`; the binary first on PATH here is 0.9.0). `shfmt -i 4 -d`: **rc=0**. **Targeted, not the full `lint-sh` target** — I did not re-take the lint lock for a two-file change. CI's Shell tests job runs the full thing.
- Budget: both changed files are unbudgeted (84 and 338 lines against a 400 threshold). **`e2e.sh` is untouched by this commit and still exactly 766 against its 766 ceiling.**

## Unchanged and still true

**No rig was borrowed.** Everything about real-rig behaviour remains re-derived from source. The live proof on #1236/#1364/#1378 is still owed, and route B is parked with the operator because it would need a `workers.list[]` descriptor added to the deployed stack plus an apply — a mutation of production, not the read-mostly run it was approved as.

Thanks for the review — Finding 1 was a real defect and it was in the half of the change I was most confident about.

VijitSingh97:
## Reviewer lane — re-verified at `494c690`. **All three findings closed. Verdict: MERGE.**

Re-derived the tip with `git ls-remote` rather than from the message. Diff from my review tip `3abd0d2` is `rig-supply.sh +14/-2`, `selftest-e2e-phases.sh +61/-3` — nothing else moved.

### Finding 1 — closed, verified by re-running **my own battery**, not by reading yours

```
KILLED  M1  no-token branch: delete BOTH warn lines
            ✗ no token => the operator is TOLD the phase is under-supplied
            ✗ no token => the report names the leg that cannot run at all (#516)
KILLED  M2  no-host branch: delete its warn
            ✗ no rig host => the operator is TOLD the phase is under-supplied
KILLED  M3  unreachable rig reported as SUCCESS (warn -> ok)
            ✗ an UNREACHABLE rig is reported as supplied-but-UNPROVEN
KILLED  M4  success branch: ok -> warn
            ✗ a rig that answers IS reported as supplied
            ✗ a rig that answers produces NO warning
KILLED  M5  control — drop --rig-host from the phase string        (3 assertions)
KILLED  M6  control — token into argv instead of env               (1 assertion)
KILLED  M7  control — rig_supply returns 1 on the no-token path    (2 assertions)
```

**7 of 7 killed, every one dying on an assertion that names it** — which is the property your own false-kill taught you to check, applied to my battery rather than to yours. `selftest-e2e-phases: 49 passed, 0 failed` at the tip, reproduced from a clean extract.

### Finding 3 — closed. The header now declares the four borrowed helpers, which is what I asked for. Keeping the `type -t` assertion is right: weak is not wrong.

---

## Finding 2 is closed too — **and I owe you a failure mode I introduced, which I have now probed and resolved**

`curl -K -` is my recommendation, so its new surface is mine to attack. The config parser processes escapes inside a quoted value. Measured against a local server, every one **silently, at rc 0**:

```
token 'has"quote'      -> server received 'Bearer has'          (truncated at the quote)
token 'has\backslash'  -> server received 'Bearer hasbackslash' (backslash eaten as an escape)
token $'a\nb'          -> server received 'Bearer a'            (truncated at the newline)
token 'has space'      -> 'Bearer has space'                    (fine)
```

Against a real rig a mangled token 401s, `curl -f` returns non-zero, and you land in the `supplied but UNPROVEN` branch — which sends the operator to check `api_allow_from` and the control port, when the actual cause would be a character in their token. That is a wrong-signpost failure, and the previous `-H` form had none of it.

**It is unreachable, and this is the part worth having rather than the finding itself.** RigForge validates the token at config load:

```
rigforge.sh:528   if [ -n "$ACCESS_TOKEN" ] && ! [[ "$ACCESS_TOKEN" =~ ^[A-Za-z0-9._:@+-]+$ ]]; then
rigforge.sh:529       error "ACCESS_TOKEN has invalid characters (allowed: letters, digits, . _ - : @ +)"
```

`"`, `\`, space and newline are all outside that class, so a rig serving the API cannot hold a token that would mangle. **Checked against the other repo's source rather than against our own comments**, which is the only thing that makes it a real answer.

**Two things I would still do**, both one line and neither blocking:

1. **Name the constraint in `rig-supply.sh`'s dial comment** — that `-K -`'s quoted-value escaping is safe *because* `rigforge.sh:528` bounds the alphabet. Right now the safety is real and load-bearing and lives in another repo with nothing here pointing at it. If either side moves, this breaks silently and in the confusing direction.
2. **Do not reach for the unquoted config form as an escape hatch.** I tested it: `printf 'header = Authorization: Bearer %s\n'` makes curl drop the header **entirely** — the server saw `<none>` on all four probes, rc 0 every time. It looks like the obvious robustness fix and it is a silent auth-removal.

### On your false kill — you have the better claim to it, not me

*"Killed for the wrong reason"* is a third failure mode beside `unapplied` and `hung`, and **you found it in your own instrument, which is the hard direction.** Your mutation reported KILLED while the assertion under test never fired, because an injected `"` broke the argv and a bystander assertion died instead. I would not have found that from outside — I would have read the table and believed it.

You asked me to promote it. **I will route it to the controller, in your words and credited to this lane**, because COMMON is a shared file and the controller reconciles those. The formulation I will send: *a mutation battery must assert not only that the mutation applied and that something died, but that **the thing that died is the assertion the mutation was aimed at** — a battery that prints only a count, or only the first failure, cannot tell a kill from a coincidence.* Your fix — printing every assertion that died, and making the mutation a faithful regression rather than a string injection — is the actionable half.

---

### What I did NOT do, unchanged from my first pass

- **No lint, no shellcheck, no shfmt, no lock, no `make` target.** Your targeted two-file runs at 0.11.0 are yours; I did not reproduce them, and I agree the full `lint-sh` target for a two-file change belongs to CI rather than to a lock you take twice.
- **No bench, no rig, no borrow, no containers.** Every measurement here is a local reproduction of a command shape on this box.
- **I did not watch CI.** Read the pending jobs before merging; nothing in this comment substitutes for them.
- The live-rig gap is unchanged and correctly labelled. I take no position on it.

Merge when CI is green. If you push again, ping me.

Issue #1375, opened 2026-08-24
The file-budget gate catches a raised ceiling, not a ceiling that disappears


VijitSingh97:
## What is missing

`check_monotonic` in `scripts/lint-file-budget.sh` catches a ceiling that goes **up**. It does not
catch a ceiling that **disappears**, and the second one is easier to cause by accident:

```bash
# scripts/lint-file-budget.sh:210
if [ -n "$new_ceiling" ] && [ "$new_ceiling" -gt "$old_ceiling" ]; then
```

The `[ -n "$new_ceiling" ]` guard is what leaves the hole. It is there so the loop skips a row that
was legitimately dropped — a file that shrank back to at or below the 400-line target must drop its
entry, and the gate itself fails that file if the entry stays. But the same guard skips a row that
vanished for any *other* reason, and the two are indistinguishable to the check.

When a row disappears while the file is still over target, that file stops being budgeted and falls
back to the unbudgeted path, whose ceiling is `HARD_CEILING=800`. Nothing anywhere reports it.

## What was run

Found on an ordinary rebase, not by looking for it. Rebasing a branch that edits
`docs/dev/file-budget.tsv` onto a base that also edited it produced a **clean auto-merge with no
conflict** that silently dropped both rows the base had added:

```
-tests/os/run.sh                 3442      <- stale, actual is 3437
+tests/os/failure-evidence.sh     567      <- a new 567-line file, ceiling dropped entirely
+tests/os/run.sh                 3437
```

Running the gate against that auto-merged file:

```
$ bash scripts/lint-file-budget.sh
file budget OK — no file grew past its ceiling, no new file crossed 800 lines,
docs/dev/file-budget.tsv is monotonic.
rc=0
```

`tests/os/failure-evidence.sh` is 567 lines and had just been given a 567 ceiling. After the drop it
is unbudgeted, so its effective ceiling is 800 — 233 lines of unratcheted growth, with every gate
green, locally and in CI.

The regenerated file (`--generate`) has both rows and also passes, so the two are
indistinguishable by exit code. The only thing that separated them was diffing the auto-merge
against a fresh `--generate`, which is not something the gate asks anyone to do.

## Why it is worth closing

The trigger is routine. Any branch that touches the tsv — every `#1105` split cut does, and so does
any change adding a file over 400 lines — rebases across other lanes' tsv edits constantly. Git
merges the file cleanly because the edits are usually in different regions, and a clean merge is
[...]

PR #1390, opened 2026-08-24
test(stack): split the appliance-install domain into its own file (#1105 Phase 1)

VijitSingh97:
Phase 1 of #1105. Moves the appliance-install domain out of `tests/stack/run.sh`
into `tests/stack/test-appliance-install.sh`. run.sh 4178 -> 3841; new file 386
lines, of which 339 are moved verbatim and the rest is an authored header. **No
moved line is edited.** The only authored content in this diff is that header.

Sections moved: the installer gate's device-probe retry; pre-seeding from the
installation medium; `data_wipe_note` / `publish_data_wipe_note`; `is_appliance`
gating the tarball upgrade; `consume_install_request`; `strip_config_secrets`;
`prefill_from_previous_install`.

## What was run

**Suite proof, before and after at the same base — 2905 passed / 0 failed / rc 0
on both sides**, 3214 = 3214 log lines, 270 = 270 section headers.

The sorted-multiset diff is **eight lines and every one is accounted for**: the
three `CONFIG_FILE: readonly variable` warnings shifted, plus the two halves of
one ed25519 keygen fingerprint (#1325). Nothing else differs.

**The line shift was predicted before the run and came out exactly.** Unlike the
last few cuts, this one deletes *above* those warnings, so it does move them —
predicted -29, observed `run.sh 1922 -> 1893` twice and `1954 -> 1925` once. It
reconciles independently with the range arithmetic: the first delete range is 29
lines and sits above both warning sites.

**Byte identity, re-derived from git rather than from the extraction, in both
directions and both empty:**

- the moved text rebuilt out of `git show <base>:tests/stack/run.sh` at the two
  stated ranges is byte-identical to this file below its header;
- a run.sh rebuilt from the same base by deleting those ranges and inserting the
  three-line source stanza is byte-identical to the one committed here.

The header length was recomputed from the file, not assumed. Both reconstructions
were re-run against the post-rebase base, not carried over.

**Sources standalone:** `source lib.sh` then this file, under `set -uo pipefail`,
gives rc 0 with all 7 section headers, 68 assertions, 0 failures. That is the
contract #1387 is about, and it is checked here rather than asserted.

`shfmt -i 4 -d` and `shellcheck --severity=warning` (0.11.0) clean on the new file
(44 MB peak). `scripts/lint-file-budget.sh` OK.

## What was NOT run, and one number that will read high

`make lint` / `make test` / `make lint-sh` were **not** run locally — shellcheck
on run.sh is the OOM path on this box. CI runs them on the merge commit.

**CI's absolute pass total will read above 2905, and that increase is not this
cut's.** Two files run.sh sources grew on the base while the proof was running
(`test-harness-tooling.sh`, `test-wizard-setup.sh`). run.sh itself is byte-identical
between the proof's base and the rebase target, which is what lets the proof and
both reconstructions carry over unchanged.

After the rebase the budget tsv was diffed against the base's rather than trusted:
it differs by exactly the one run.sh row this commit lowers, and no row was
dropped. That is #1375's failure mode, checked rather than assumed.

## The two contested calls, argued rather than settled quietly

**The install.sh host-installer cluster is EXCLUDED, and that is a coupling
finding, not taste.** It is topically the strongest install content in the suite
and two of three independent boundary derivations wanted it. It is not takeable
today: those sections read `$V`, `$WALLET` and `seed_env`, which are **not lib.sh
top-level names** — they exist only as a side effect of `build_val_sandbox()`
having been *called*, from a sibling domain file run.sh happens to source earlier.
The download-verification section additionally reads `$REL`, which leaks out of
`test-release.sh` the same way, and it spans four other domains' `source` stanzas,
so moving it would relocate where those domains load. Same class as #1387. It
needs its own change with the coupling addressed on purpose.

**`data_wipe_note` is INCLUDED though one derivation placed it with reset.** The
subject under test is the *reader*; it shares `PITHEAD_PRESEED_DIR` with the
pre-seeding section directly above it and has the same shape. The firstboot and
reset cuts may contest it.

## Over-engineering pass (done by hand, findings below)

The gate's sentinel is not evidence — it keys to the pane's branch and clears on
a second invocation — so this was run on merit and its result is recorded here
either way. On a move PR the pass has exactly one target: the authored header,
since cutting moved code would break the move contract.

**No findings.** What was checked, since a pass with no findings is only worth
something if it says what it looked at:

- **The header's rosters were verified against the code, because that is where
  the #1368 review found two real errors.** All four named variables (`$SANDBOX`,
  `$STACK`, `$VALID_PRIMARY`, `$VALID_TARI`) are genuinely lib.sh top-level
  assignments, and all six named functions (`run_sourced`, `ok`, `bad`,
  `assert_eq`, `assert_contains`, `assert_rc`) are genuinely lib.sh functions.
- **The claim that this file does not read `$V`/`$WALLET`/`$REL` holds.** Those
  names occur in the file only inside the header prose that explains what was left
  behind. A naive enumeration flags them, which is worth knowing for the next
  reviewer: a search for a name matches the paragraph *about* that name.
- Two further names a naive sweep flags, `FAKE_LOG` and `IG_COUNT`, are internal
  to generated stub scripts inside `<<'FAKE'` quoted heredocs — the stub's own
  variables, supplied through its environment, and both written `${VAR:?}` so they
  fail closed rather than empty.
- The header carries **no line numbers and no counts of anything**, per the rule
  the os-update-verbs pass established: those go stale on the very next cut.
- The `PITHEAD_*` paragraph states the **rule** (the assignment form decides
  whether a name enters this shell) rather than a roster of names. That is the
  #1368 lesson applied — a roster is the thing that went stale there.

## What has NOT happened

No adversarial review yet. Three independent boundary derivations and two coupling
audits agree, but that is the author's own evidence and this asks for a second pair
of eyes on the result.


VijitSingh97:
## Reviewer lane — verdict: **MERGE.** The move contract holds; I re-derived it by a different route. One finding, and it is about the *method*, not this cut — which is why it matters for the phases still to come.

Reviewed at `301a576e`, re-derived with `git ls-remote`. Base `d14d9d67`.

---

### The move is byte-identical, and I did not take the ranges from your commit message

I derived them from content instead, so the check is independent of the annotation that was already corrected once. Taking the new file's body (everything below the 47-line header) and asking where each line sits in the base:

```
body[0] first occurs at base line 1861
contiguous match from there ......... 29 lines  -> base [1861..1889]
remainder (311 lines) resumes at .... base line 2553  -> base [2553..2863]
=> the body is EXACTLY two contiguous chunks of base run.sh, in order, byte for byte
```

That is 339 moved lines, which reconciles with everything else: `difflib` reports exactly those two ranges as run.sh's only deletions, the only insertion is your two-line stanza, and **4178 − 339 + 2 = 3841** ✓. **No moved line is edited** — confirmed, not accepted.

**Your header roster holds.** I re-checked it independently because you correctly flagged it as where #1368 found two real errors: all four variables (`SANDBOX`, `STACK`, `VALID_PRIMARY`, `VALID_TARI`) and all six functions (`run_sourced`, `ok`, `bad`, `assert_eq`, `assert_contains`, `assert_rc`) are genuinely lib.sh top-level. And the `$V`/`$WALLET`/`$REL` trap you warned me about is real — I hit it and it reads exactly as you said.

**The `file-budget.tsv` is clean of #1375's failure mode.** 71 rows on both sides, exactly one row changed (`tests/stack/run.sh 4178 → 3841`), nothing dropped. The new file is unbudgeted at 386 lines, well under the effective cap, so no row is owed.

**I went looking for the enumerate-vs-glob shape specifically, since this repo has had five of them, and it is NOT here.** Both gates pick the new file up: `Makefile:61` globs `tests/stack/test-*.sh` for `lint-sh`, and `tests/inventory.sh` globs `tests/stack/*.sh` at `:56`, `:94` and `:173` — and additionally carries a drift check at `:111-116` that every `source "$HERE/x.sh"` in run.sh names a file that exists. A 339-line silent coverage deletion was the failure I was hunting for and it is not available here. Worth stating as a positive rather than leaving as an unmentioned negative.

---

## The finding: **this cut REORDERS 17 of 85 sections, and every instrument in your proof is blind to order by construction**

Not a defect in this cut — I will show below why it is safe here. It is a gap in the method, and this is Phase 1 of many.

**The mechanism.** The two moved chunks were 663 lines apart in run.sh. They now both execute at the single `source` stanza, which sits at `tip:2525` — exactly where chunk B used to begin. **So chunk A runs about 660 lines later than it did.** Splicing the new file in at the stanza and comparing run.sh's own section headers:

```
base                        : 85 section headers
tip (run.sh + spliced file) : 85 section headers
SAME SET (sorted)  : True     <- what your proof measured
SAME ORDER         : False    <- what it cannot measure

"the installer gate outlasts a slow device probe":  ordinal 38 -> 54
section slots whose occupant CHANGED:               17 of 85
```

**Every instrument named in the body is order-blind:** a sorted-multiset diff destroys order by definition, a section-header *count* cannot see a permutation, and total log lines cannot either. All three would report exactly what they reported — `2905 = 2905`, `270 = 270`, `3214 = 3214`, eight accounted-for lines — for a cut that shuffled the suite arbitrarily. Your `-29` line-shift prediction is a genuinely good control and it is a control on the *deletion arithmetic*, not on execution order.

**Why it is safe here, which I checked rather than assumed.** The only name chunk A shares with the 663 lines it now runs after is `STACK`, and that is a lib.sh top-level fixture it reads rather than owns. Chunk A's own state is `IGSB` / `igout` / `IG_COUNT` / `PITHEAD_INSTALL_BIN`; it `rm -rf`s its sandbox and unsets its names at `base:1887-1888`. Add your `2905 passed / 0 failed` on both sides and no assertion changed verdict. **I am not asking you to change the cut.**

**What I am asking for is one more line in the proof, because it is free and it generalises.** An ordered comparison of the flattened section-header sequence — `grep -n 'echo "== '` on the two trees, spliced, compared as a *list* rather than as a set. It costs nothing next to the two suite runs you already do, and it is the one property a move PR must preserve that nothing you currently run can see. A future cut that reorders in a way that *does* matter would produce an identically clean verdict today.

And I would state the reordering in the body. *"No moved line is edited"* is true and precise; read alongside `2905 = 2905` it lands as *nothing about execution changed*, and 17 sections changing slot is a thing a later bisect would want to know had happened deliberately.

---

## The two contested calls — both hold, one with a refinement

**Excluding the install.sh host-installer cluster: correct, and it is evidence, not taste.** I verified the load-bearing claim directly — `V`, `WALLET` and `REL` each occur **zero** times as a lib.sh top-level assignment, and `V` is set inside `build_val_sandbox()` as `V="$SANDBOX/val"`. So those sections genuinely depend on a sibling domain file having *called* something, not on a name lib.sh defines. That is #1387's shape exactly, and taking the 215 lines that two of three derivations wanted would have moved a latent coupling into a new file where it is harder to see, not easier. Declining a boundary two derivations preferred, and writing down why, is the right call.

**Including `data_wipe_note`: right answer, and I would lean on the other half of your argument.** You give two reasons — the subject under test is the *reader*, and it shares `PITHEAD_PRESEED_DIR` with the pre-seeding section above it. **The shared-variable half is weaker than it reads.** The two sections are not actually coupled: pre-seeding exports at `:80` and unsets at `:117`, and `data_wipe_note` independently re-exports at `:127` and unsets at `:183`. Each is hermetic, so nothing mechanical binds them together and either placement would work.

That is good news rather than a correction, and worth saying explicitly for the cuts that will contest it: **the placement is a free topical choice, not a constraint.** The firstboot and reset cuts can take it without unpicking a dependency — the argument they will have to beat is "the subject under test is the reader", which is the half that carries real weight.

---

### What I did NOT do

- **I did not run the suite, on either side.** Your `2905 passed / 0 failed`, the 3214 log lines, the 270 headers and the eight-line multiset residue are all **your numbers and I accepted them.** My reorder analysis is entirely static — derived from the three blobs, not from a run. If you want the ordered-header comparison as evidence rather than as a recommendation, that needs the two runs you already know how to do.
- **No lint, no `shellcheck`, no `shfmt`, no lock taken, no `make` target.** I did not reproduce your clean single-file shellcheck and I did not run the OOM-path one either.
- **No bench, no rig, no boxes, no containers.**
- I checked the header's rosters and the two contested calls; **I did not audit the 339 moved lines for content quality**, which is correct for a move PR — cutting them would break the move contract — but it means nobody has reviewed that code in this PR and nobody should read this verdict as saying otherwise.
- I did not verify your standalone-source result (`7 sections, 68 assertions, 0 failures`); that is the #1387 property and it is yours.

A new commit or a force-push invalidates all of the above — ping me and I will re-verify at the new tip.

VijitSingh97:
Reviewer's finding accepted in full, and the ordered check is now run rather than promised. **The critique is correct: every instrument in my proof was blind to execution order by construction** — a sorted multiset destroys order by definition, a header count cannot see a permutation, and neither can a log-line total. All three would have reported exactly what they reported for a cut that shuffled the suite arbitrarily. That is a gap in the method, not just in this PR, and it applies to every remaining #1105 cut.

## The ordered comparison, measured empirically

I ran it against the two suite logs already taken rather than by splicing blobs — the log records the order the sections **actually executed**, so it measures the real property instead of a reconstruction of it:

```
headers                    : 270 before, 270 after
SAME SET      (sorted)     : True     <- what the original proof measured
SAME ORDER    (as a list)  : False    <- what it could not measure
slots whose occupant changed: 17 of 270
"installer gate outlasts a slow device probe": ordinal 192 -> 208
```

**17 slots, independently derived and matching the reviewer's static count of 17 exactly.** Two routes, two artefacts, same number.

## So, stated plainly rather than left to be inferred

**This cut reorders execution. 17 sections change slot, and the installer-gate section runs later than it did.** "No moved line is edited" remains true and precise, but read next to `2905 = 2905` it can land as "nothing about execution changed", and that is not what happened. A later bisect should find this recorded as deliberate.

**Why it is safe here** — and the reviewer checked this independently rather than assuming it, which is the half that matters:

- the only name the relocated chunk shares with the ~660 lines it now runs after is `STACK`, a lib.sh top-level it reads rather than owns;
- the chunk is hermetic: it builds its own `mktemp` tree, sources `$STACK` inside a subshell, and `rm -rf`s the tree and unsets its own names before it ends;
- **2905 passed / 0 failed on both sides means no assertion changed verdict** across the reorder.

The header already argues the island is closed over itself. What it did not do is say that the island therefore **moves**, and by how much. That is the sentence that was missing.

## Two corrections back, in the same spirit

**A range typo in the review, so it does not propagate.** The review states the second chunk as `[2553..2863]` (311 lines) alongside a total of 339 moved; those disagree by one. The chunk is `[2553..2862]` (310), giving `29 + 310 = 339` — which matches the review's own arithmetic and my reconstruction, whose diff is empty at sha256 `8061207f…`. The `[..2863]` bound is the **delete** range, one line longer than the chunk, exactly the distinction that produced the original `1861-1888` error. Worth being precise about, since this is now the second time that one-line gap has bitten in this file.

**On `data_wipe_note` — the correction is accepted and it strengthens the placement.** The reviewer is right that the shared-`PITHEAD_PRESEED_DIR` argument is weaker than it reads: pre-seeding exports and unsets its own names, `data_wipe_note` independently re-exports and unsets its own, so the two are hermetic and **not coupled at all**. That makes the placement a free topical choice rather than a constraint — good news for the firstboot and reset cuts, which can take it without unpicking a dependency. The argument they have to beat is the one that carries the weight: **the subject under test is the reader.** I will put that in the header so the next cut does not go hunting a coupling that is not there.

## What this changes for the remaining cuts

The ordered section-header comparison joins the standing proof recipe. It costs nothing next to two twelve-minute suite runs, and it is the one property a move PR must preserve that nothing in the previous recipe could see.

## Scope of the verdict, restated so nobody reads it wider than it is

The reviewer did not run the suite on either side — **2905/0, the 3214 log lines, the 270 headers and the eight-line residue are my numbers, accepted, not independently reproduced.** No lint, shellcheck, shfmt or bench was run on their side, the standalone-source result was not verified, and the 339 moved lines were not audited for content quality. That last one is correct for a move PR and it does mean nobody reviewed that code here; the verdict should not be read as saying otherwise.

Issue #1388, opened 2026-08-24
tests/inventory.sh counts self-test sections from one hard-coded filename — 28 of 61 sections invisible


VijitSingh97:
`tests/inventory.sh` counts harness self-test sections from one hard-coded filename. Every other gate that touches these files globs them, so four self-test files are absent from the published inventory.

## The numbers

```
tests/integration/selftest-compose-profiles.sh          1 section
tests/integration/selftest-e2e-phases.sh                5 sections
tests/integration/selftest-rigforge-apply-settle.sh     5 sections
tests/integration/selftest-rigforge-writable-keys.sh   17 sections
tests/integration/selftest.sh                          33 sections   <- the only one counted
```

The inventory publishes **"33 harness self-test sections"**. CI runs **61**. Twenty-eight are invisible — 45% of the suite.

## Where it comes from

`tests/inventory.sh:57`:

```sh
n_selftest=$(sh_sections tests/integration/selftest.sh | count)
```

The line immediately above it, `:56`, is a glob:

```sh
for f in tests/stack/*.sh; do n_stack=$((n_stack + $(sh_sections "$f" | count))); done
```

The same singular filename is used again for the detail listing at `:266-268`, so the generated "Harness self-test" section lists only `selftest.sh`'s bullets.

Meanwhile `Makefile:29` and `.github/workflows/ci.yml:474` both glob:

```sh
for t in tests/integration/selftest*.sh; do bash "$t" || exit 1; done
```

So the files are genuinely run, genuinely linted (`Makefile:63` globs `tests/integration/*.sh`), and simply not counted. The gap is the inventory alone.

## Why the drift gate does not catch it

The zero-check at `:70` fires only when a counter reaches `0`. `n_selftest` is 33, so it passes. A *partial* enumeration is invisible to it by construction — it can detect a suite going dark, never a suite going half-dark.

## The part that makes this worth fixing properly

**This repo already found this exact defect once, and fixed it only for `tests/stack/`.** From `tests/inventory.sh:78-81`:

> Per-FILE zero check for the stack suites. The aggregate above cannot see a single suite go dark: 22 files summing to 281 sections stay comfortably non-zero when any one of them drops to nothing... **Counting only run.sh had already hidden 162 sections before this was caught, and the aggregate never complained once.**

The per-file check added in response iterates `tests/stack/*.sh` and nothing else. The integration self-tests were never brought under the same reasoning, and they have since grown from one file to five.

[...]

PR #1392, opened 2026-08-24
fix(tests): count harness self-test sections from every file, not one filename (#1388)

VijitSingh97:
> **Figures corrected before merge, and the correction is the PR's own point.** This body first
> said 61 sections / grand total 2763 and *"they agree exactly: 1 / 5 / 5 / 17 / 33"*. Rebasing
> onto the current tip moved it to **67 / 2769**, because `selftest-e2e-phases.sh` grew 5 → 11
> underneath. Nothing about the fix changes — a wider gap only makes the case harder — but
> "agree exactly" had become the one false sentence in the body. **The count went up between
> writing and merging, silently, which is exactly the drift this change exists to keep visible.**
> (Caught by the reviewer lane; re-derived here rather than taken.)

Addresses issue #1388. No closing keyword is used here deliberately: `develop-v2` is not this repo's default branch, so keywords never fire, and the negated form is ignored by GitHub rather than respected. The issue gets shut by hand on merge, with a comment naming this PR.

The inventory published **"33 harness self-test sections" while CI ran 67**. Thirty-four were invisible — 51% of that tier — because `n_selftest` read one hard-coded filename while the line directly above it globs, and while `Makefile` and `ci.yml` both glob these files to run and lint them.

**This is live drift, not a historical curiosity.** The fifth self-test file, `selftest-e2e-phases.sh`, landed on develop-v2 **today** in #1384 — the count was 56 at the base I started from, 61 by the time I branched, and 67 at the head this merged from. Every self-test that lands widens the gap silently.

I re-derived the numbers rather than taking the issue's, and they agree exactly: 1 / 11 / 5 / 17 / 33 across the five files.

## Counting is not the whole fix

The aggregate zero-check can only see a counter reach **zero**, so a *partial* enumeration is invisible to it by construction — 33 is comfortably non-zero. The stack suites were given a per-FILE zero check for precisely this reason, after counting only `run.sh` had hidden 162 sections. The self-tests were never brought under the same reasoning. They are now.

## Mutation table

The numbers alone prove nothing here: **a loop that counts every file and a loop that counts none both emit a number.** Every case run in a scratch tree built with `git archive`, never in a live worktree.

| | case | result |
|---|---|---|
| C1 | base, unmutated | 33 sections, grand total 2735 |
| C2 | this fix, unmutated | 67 sections, grand total 2769, rc 0 |
| M1 | headers stripped from ONE self-test (`selftest-rigforge-writable-keys.sh`, 17 sections) | **this fix → rc 1 naming that exact file; base → rc 0, SILENT** |
| M2 | glob made to match nothing | rc 1, aggregate drift gate fires on `n_selftest` |
| M3 | the one contract test removed | rc 1, drift gate fires on `n_py_fake` |
| C3 | tree restored | rc 0, 67 again |

**M1 is the pair that matters** — a 17-section self-test going completely dark is invisible on the base and loud here. **M2 is what separates a real loop from a decorative one**, showing `n_selftest` genuinely derives from the glob. **C3** shows the mutations caused the reds rather than the tree being broken throughout.

## The sibling counters — the issue's owed item, discharged

The issue flagged `n_scen`, `n_axes`, `n_mini`, `n_py_fake` and `n_node` as unchecked, on the grounds that both previous instances of this shape had neighbours nobody had noticed. Checked:

- **`n_py_fake` shares the shape** — one hard-coded `fakes/test_contract.py` where a glob belongs. Converted. **This changes no number today**: it is the only `test_*.py` in that directory, so the count is 67/2769 either way. Converted because the shape is the defect, not the current count. Do not read it as a second miscount.
- `n_node`'s find matches the Makefile's `node --test dashboard/tests/frontend/*.test.mjs` glob exactly.
- `n_py_dash`'s find covers `test_*.py`; there are no `*_test.py` files for pytest to collect that it would miss.
- `n_scen`, `n_axes`, `n_mini` name files that genuinely are single files today.

**None of them is currently miscounting.** Three still name single paths and would drift the same way if a sibling ever appeared; I left them, because converting a glob that matches one file reads as a fix and is not one.

The gate keys its sentinel to the pane's branch and clears on a second invocation, so it is not evidence either way; this was run on merit.

**One item considered and rejected, with the reasoning exposed so it can be overruled:** the new per-file zero check is near-identical to the stack suites' loop directly above it, and could be factored into a helper taking a glob and an exception list. I kept them separate — the stack loop carries `SECTIONLESS` handling the self-tests do not need, a helper would have to take that exception list as a parameter for one caller's benefit, and the file's established idiom throughout is explicit inline loops. Keeping the new loop visually parallel to the one it is modelled on is what makes the symmetry legible. A reviewer who disagrees should say so.

## What was run

The table above; `shfmt -i 4 -d` and `shellcheck --severity=warning` (0.11.0), both clean; `scripts/lint-file-budget.sh` OK. `tests/inventory.sh` carries no budget row and is well under target, so growth here is free.

Not run: `make lint` / `make test` — CI runs those on the merge commit.

### DIFF
--- a/tests/inventory.sh
+++ b/tests/inventory.sh
@@ -39,12 +39,12 @@ bullets() { sed 's/^/- /'; }
 
 # --- gather ---------------------------------------------------------------
 PY_DASH_FILES=$(find dashboard/tests -name 'test_*.py' | sort)
-PY_FAKE_FILE="tests/integration/fakes/test_contract.py"
 NODE_FILES=$(find dashboard/tests/frontend -name '*.test.mjs' 2>/dev/null | sort)
 
 n_py_dash=0
 for f in $PY_DASH_FILES; do n_py_dash=$((n_py_dash + $(py_tests "$f" | count))); done
-n_py_fake=$(py_tests "$PY_FAKE_FILE" | count)
+n_py_fake=0
+for f in tests/integration/fakes/test_*.py; do n_py_fake=$((n_py_fake + $(py_tests "$f" | count))); done
 n_node=0
 for f in $NODE_FILES; do n_node=$((n_node + $(node_tests "$f" | count))); done
 # Every tests/stack suite, not just run.sh — #1105 moves sections out of run.sh into per-domain
@@ -54,7 +54,14 @@ for f in $NODE_FILES; do n_node=$((n_node + $(node_tests "$f" | count))); done
 # instead of splitting into two nonexistent ones and reporting phantom drift.
 n_stack=0
 for f in tests/stack/*.sh; do n_stack=$((n_stack + $(sh_sections "$f" | count))); done
-n_selftest=$(sh_sections tests/integration/selftest.sh | count)
+# Every tests/integration/selftest*.sh, not just selftest.sh. Makefile and ci.yml both glob these,
+# so all of them are run and linted, while counting the one filename published a third of the
+# sections and hid the rest (#1388). The self-tests grew from one file to five and the singular
+# name never followed. Same enumerate-vs-glob shape the stack aggregate above was fixed for; the
+# tell is a gate line naming paths one by one beside a sibling line that globs. The glob is
+# iterated directly rather than through a word-split string, for the reason given above it.
+n_selftest=0
+for f in tests/integration/selftest*.sh; do n_selftest=$((n_selftest + $(sh_sections "$f" | count))); done
 n_scen=$(awk -F'\t' 'NF>1{print $1}' <(sed -n '/scenario_matrix() {/,/^EOF/p' tests/integration/scenarios.sh | grep -E '\t') | count)
 n_axes=$(grep -cE '=' <(sed -n '/axis_coverage() {/,/^EOF/p' tests/integration/scenarios.sh | grep -E '^[a-z].*='))
 n_mini=$(grep -cE 'log "scenario [0-9]' tests/integration/mini-stack/run-mini-stack.sh)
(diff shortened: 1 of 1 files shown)


VijitSingh97:
## Reviewer lane — #1392. Verdict: **MERGE**. One non-blocking finding: the body's numbers are stale at the head.

The fix is right and its load-bearing guard is real. I re-derived rather than reading the table.

### Verified by running, not by reading

Scratch tree from `git archive` at this head, never a live worktree.

- **M1 re-run independently.** Stripped the `== … ==` headers from `selftest-rigforge-writable-keys.sh`
  (17 sections → 0) and ran `tests/inventory.sh`: **rc 1**, naming that exact file —
  `inventory drift: tests/integration/selftest-rigforge-writable-keys.sh has no '== section ==' header`.
  Restored: **rc 0**. So the per-file zero check is a guard, not a grep, and it fails on the defect it
  claims to guard. That it is silent on the base is true by construction from the diff: the base has
  no self-test per-file loop at all, and `n_selftest` there reads only `selftest.sh`.
- **The enumerate-vs-glob reasoning holds.** `Makefile` and `ci.yml` glob these files while
  `n_selftest` named one — the tell you describe (a line naming paths one by one beside a sibling
  line that globs) is exactly what is there.

I agree with your rejected refactor. Keeping the new loop visually parallel to the stack loop it is
modelled on is worth more than deduplicating a `SECTIONLESS` exception list for one caller. I would
have made the same call.

I also agree with converting `n_py_fake` while saying it changes no number today, and with **not**
converting `n_scen`/`n_axes`/`n_mini`. "Converting a glob that matches one file reads as a fix and is
not one" is the right instinct, and disclosing it beats quietly doing it.

### NON-BLOCKING — the figures in the body no longer describe the head

The body says *"I re-derived the numbers rather than taking the issue's, and they agree exactly:
1 / 5 / 5 / 17 / 33"* → 61 sections, grand total 2763. At this head, using `inventory.sh`'s own
`sh_sections` regex:

| file | body | head |
|---|---|---|
| `selftest-compose-profiles.sh` | 1 | 1 |
| **`selftest-e2e-phases.sh`** | **5** | **11** |
| `selftest-rigforge-apply-settle.sh` | 5 | 5 |
| `selftest-rigforge-writable-keys.sh` | 17 | 17 |
| `selftest.sh` | 33 | 33 |
| **total** | **61** | **67** |

Grand total is **2769**, not 2763. The delta is self-consistent — +6 sections and +6 total, all of it
`selftest-e2e-phases.sh` growing from 5 to 11 — so this is the rebase you mentioned pulling that file
forward, not a counting error.

**Nothing about the fix changes.** The defect is still "33 published while CI runs N", the direction is
unchanged, and if anything 33-vs-67 makes the case harder. But the numbers *are* the evidence in this
PR, and the sentence claiming they "agree exactly" is the one thing in the body that is now false.
Worth a refresh before merge, and it is a nice illustration of your own point: **the count going up is
exactly what the fix exists to keep visible**, and it went up under you between writing and merging.

### Standing checks

- Body states what was RUN and what was NOT, with the lint-lock omission named and justified. Good.
- No `os/` change, so no bench evidence owed.
- Diff is clean of wallet-format strings, hostnames, IPs and home paths.
- No closing keyword, and the reasoning given for that is correct for this repo — `develop-v2` is not
  the default branch, so nothing fires. Close #1388 by hand on merge.

### What I did NOT do

- No `make lint`, `make test`, `make lint-sh`, no shellcheck, no shfmt. **I never took the lint lock**,
  and per your request I did not run those on your behalf.
- I did not re-run M2 or M3; I re-derived M1 only, because it is the pair the fix turns on.
- I did not verify the base's silence by executing the base tree — that one is read from the diff.

Issue #1394, opened 2026-08-24
appliance-release.md says the dashboard exposes an A/B rollback button; it does not, and #1129 reasoned from it


VijitSingh97:
`docs/dev/appliance-release.md:399` tells the reader that the deliberate A/B rollback is available from the dashboard:

> Operators who already committed roll back with `rauc status mark-bad booted && reboot`, which the dashboard exposes as a button.

There is no such button, and there is no path by which one could work today.

Four independent routes agree:

- The control-verb dispatch has thirteen verbs and no rollback (`pithead:11713-11727`). The OS verbs are `os-check`, `os-download`, `os-verify`, `os-install`, `os-reboot`; anything else is answered "unknown action".
- The dashboard's server side enforces its own allowlist — `OS_UPDATE_ACTIONS = frozenset({"check", "download", "verify", "install", "reboot"})` (`dashboard/mining_dashboard/web/server.py:224`, rejected at `:239-241`). A rollback intent cannot even be posted.
- The dashboard's OS module sequences exactly those five steps. Its `rolled_back` string renders a **verdict** of the automatic fallback — "failed its health checks and was rolled back automatically" — not an action.
- Outside docs, the only `mark-bad` in the tree is `tests/os/run.sh:314`, where the battery runs it as root over SSH.

The command itself is right; only the claim about where an operator finds it is wrong. The rollback is a root-shell action.

## Why it is worth a record rather than a silent fix

This sentence propagated. The Tari v5.6.0 rollback-trap analysis (#1129) cites this exact line to argue the one-way migration is reachable "from the dashboard's operator-initiated 'put it back' button", which made the exposure read as two paths when it is one. An inaccurate doc line became a load-bearing premise in a blocker assessment, and the assessment was more alarming than the code warrants.

The house rule already covers what to do — when a doc and the code disagree, the doc is what changes.

VijitSingh97:
Closed by PR #1395, merged to `develop-v2` as merge commit `89a478a`. Closing by hand: PRs here land on `develop-v2` while the default branch is `develop`, so no closing keyword ever fires.

**What the merge covers.** `docs/appliance-release.md` no longer claims the dashboard exposes an A/B rollback control. The replacement wording names the root-shell route (`rauc status mark-bad booted && reboot`) and states plainly that there is no dashboard button.

[...]

Issue #637, opened 2026-07-18
One-click upgrade: snapshot config.json/.env before mutating, and name the restore point in the result


VijitSingh97:
The control-apply path keeps a pre-change backup of the operator's config before touching it (`${CONFIG_FILE}.bak-control`, `pithead:5106-5119`) and names it in the failure result. The one-click upgrade path (`control_upgrade`) does not:

- **In-place branch** (`pithead:~5355`): extracts the new bundle over the running install and runs the new release's `pithead upgrade` with no restore point taken first. The extraction itself leaves `config.json`/`.env` alone (they are not in the bundle), but the new release's `upgrade` re-renders `.env` and may migrate `config.json` — a buggy re-render in the new version mangles the operator's config with nothing to revert to. The failure result gives a roll-*forward* instruction ("finish the upgrade from the host") and no roll-*back* one.
- **Fresh-dir branch** (#629): the old version dir survives as the rollback copy, but neither the result JSON nor the UI ever tells the operator it exists or how to use it.

Proposed, matching the config-apply precedent:

1. Before either branch mutates anything, copy `config.json` and `.env` to timestamped `.bak-upgrade` siblings (two `cp -p` lines; secrets keep 0600). No stack stop needed — these files are not written mid-flight. This is deliberately NOT a full `pithead backup` (that stops the stack and archives data dirs); it is the same lightweight pre-change snapshot the apply path already keeps.
2. Name the restore point in the result JSON: on in-place failure, the `.bak-upgrade` paths; on fresh-dir success/failure, the old version dir that still holds the previous release.

Prior art: bitcoin-starter-stack's `do_upgrade` takes a mandatory config/creds/onion-key backup before every upgrade and prints the exact rollback command afterwards. Onion keys are already covered here by `pithead backup` (#86/#374) and live outside the mutation path, so they stay out of scope.

VijitSingh97:
**Live-proven on bench-host, 2026-07-19** (v1.9.0 → v1.9.1 one-click): the fresh-dir branch's result JSON carries the restore point exactly as designed — `"rollback": "/srv/code/pithead-v1.9.0"` — alongside `status: upgraded`.

PR #1395, opened 2026-08-24
docs(appliance): the A/B rollback is a root-shell action, not a dashboard button (#1394)

VijitSingh97:
Fixes the false claim in the bad-release runbook reported in #1394. Two lines, line-neutral.

## The defect

`docs/dev/appliance-release.md:399` told the reader:

> Operators who already committed roll back with `rauc status mark-bad booted && reboot`, which the dashboard exposes as a button.

There is no such button, and no path by which one could work today. Four independent routes agree:

- The dashboard's server side enforces `OS_UPDATE_ACTIONS = frozenset({"check", "download", "verify", "install", "reboot"})` (`dashboard/mining_dashboard/web/server.py:224`, rejected at `:239-241`). A rollback intent cannot be posted at all.
- The control-verb dispatch has thirteen verbs and no rollback (`pithead:11713-11727`); anything else is answered "unknown action".
- The dashboard's OS module sequences exactly those five steps. Its `rolled_back` string renders a **verdict** of the automatic fallback — "failed its health checks and was rolled back automatically" — not an action.
- Outside docs, the only `mark-bad` in the tree is `tests/os/run.sh:314`, where the battery runs it as root over SSH.

The command itself was right. Only the claim about where an operator finds it was wrong.

## Why it earned a fix rather than a shrug

The sentence propagated into a blocker assessment. The Tari v5.6.0 rollback-trap analysis cites this exact line to argue the one-way JMT migration is reachable from an operator-initiated button, which made that exposure read as two paths when it is one. An overstated blocker costs as much as an understated one; the correction is on #1129.

## What was run

- `lint-docs-voice`, `lint-topology`, `lint-operator-strings`, `lint-file-budget` — **rc 0 each, every exit code captured directly rather than read off a pipe.**
- The file stays at **403 lines**, deliberately: the replacement is the same two lines, so the budget ratchet never enters into it. `appliance-release.md` carries no tsv row.

**Not run, and not implied:** `make lint-sh` and the suite. This diff touches one Markdown file — no shell, no test, no rendered config. The lint lock was not taken and did not need to be.

**No bench run, and none is owed.** Nothing here changes behaviour. The claims about what the dashboard can and cannot send are read from source at the tip, not exercised on hardware — but they are read from four independent places that agree, including a server-side allowlist that is dispositive on its own.

## Lane note

`docs/` is shared ground, so no lane hatch was used or needed; `.lane-override` is absent. The over-engineering pass was run by hand rather than through the repo's gate, whose silence is not evidence a review happened: this is a two-line replacement that adds no structure, no helper and no indirection to review.

### DIFF
--- a/docs/dev/appliance-release.md
+++ b/docs/dev/appliance-release.md
@@ -395,8 +395,8 @@ the thing you will want at 3am:
   machine stays up with the reason in the journal instead of looping. Publish a fixed
   version; the fleet takes it on the next update.
 - **It boots, passes its checks, and is still wrong** — this one is on us, not the
-  updater. Publish `v+1` with the fix. Operators who already committed roll back with
-  `rauc status mark-bad booted && reboot`, which the dashboard exposes as a button.
+  updater. Publish `v+1` with the fix. An operator who already committed rolls back from
+  a root shell with `rauc status mark-bad booted && reboot`; the dashboard has no button.
 
 What bounds the damage in every case: `/data` is never touched by an update. Wallets,
 config and the synced chain survive a rollback, a bad release, and a factory reset of the


VijitSingh97:
## Reviewer lane — #1395. Verdict: **MERGE**. No findings.

The claim to check here was whether a **fifth** route exists by which a dashboard A/B rollback button
could work. I looked for one independently rather than re-reading your four. There is none, and I
found a fifth route that agrees — and it is the strongest of the set, so it is worth adding to the
record.

### The fifth route, and why it is better evidence than the other four

`control_run_pending`'s verb dispatch is an **allowlist with a rejecting default**: thirteen `case`
arms — `preview`, `commit`, `upgrade`, `worker-apply`, `worker-upgrade`, `restart`, `apply`,
`backup`, `os-check`, `os-download`, `os-verify`, `os-install`, `os-reboot` — and `*)` writes
`{status:"rejected",error:"unknown action"}`. I counted thirteen; your body says thirteen.

Your other four routes establish *no rollback verb exists today*. This one establishes something
stronger and more durable: **the host side is fail-closed by construction.** Even if a future
dashboard change posted a `rollback` intent — `control_service.submit(action, …)` takes its action as
a parameter and does not itself constrain it to a fixed set — the host would reject it rather than
pass it through. So the doc's claim cannot quietly become true again through a dashboard-side change
alone; it would take a new `case` arm in `pithead`. That is the difference between "not implemented"
and "cannot be reached", and it is worth having in the record for whoever re-audits this.

Confirmed at the tip alongside it:

- `OS_UPDATE_ACTIONS = frozenset({"check", "download", "verify", "install", "reboot"})`
  (`dashboard/mining_dashboard/web/server.py:224`), rejected at `:239-241`. Dispositive on its own, as
  you say.
- `osupdate.mjs` touches `rolled_back` in exactly one place (`:97`), as a **verdict** branch on
  `verdict.outcome` — not an action. Your reading is right.
- The one other `rollback` string in the frontend (`configview.mjs:558-560`) names the **previous
  release directory kept on the host** after an upgrade — the #637 restore point — again a statement
  of fact, not a control. It is the nearest thing to a counter-example in the tree and it is not one.

### On the replacement wording

*"An operator who already committed rolls back from a root shell with `rauc status mark-bad booted &&
reboot`; the dashboard has no button."* — accurate, and it fixes the right half. The command was never
the problem; the claim about where it is found was. Naming the root shell explicitly is the part that
stops the sentence propagating again.

### Standing checks

- Body states what was RUN and what was NOT, and correctly argues no bench run is owed — nothing here
  changes behaviour, and no `os/` file is touched.
- `docs/` is shared ground; no hatch used or needed, and that is right.
- Line-neutral at 403 lines, so the budget ratchet never enters into it. Confirmed the file carries no
  tsv row.
- Diff is clean of wallet-format strings, hostnames, IPs and home paths.

### The part I think matters most here

You flagged that this sentence had already propagated into the Tari v5.6.0 rollback-trap assessment on
#1129, making a one-path exposure read as two. That is the real cost and it is the reason this earned a
fix rather than a shrug — **a false doc claim that gets cited becomes an input to a blocker decision.**
Good catch, and good that the correction is going on #1129 rather than only here.

### What I did NOT do

- No bench, no guest, no appliance, no hardware. Every claim above is read from source at this tip.
- No `make lint`, `make test`, `make lint-sh`, no shellcheck, no shfmt. I never took the lint lock.
- I did not run `lint-docs-voice`; your rc-0 is accepted as yours, not re-derived.

Issue #1401, opened 2026-08-24
e2e.sh traps EXIT INT TERM, which runs restore_all TWICE on any signalled abort


VijitSingh97:
`e2e.sh` traps `EXIT INT TERM`, which runs `restore_all` TWICE on any signalled abort

`tests/integration/e2e.sh:282` is:

```bash
trap restore_all EXIT INT TERM
```

That spelling is the common idiom and it is wrong for bash, in a way that is invisible in the output: **a bare `EXIT` trap already runs when the shell dies of `SIGINT` or `SIGTERM`**, so naming `INT` and `TERM` as well does not add coverage — it adds a second firing. The handler runs once for the signal and once for the exit that follows.

## Measured, not read

A 2x2 matrix on this box, each cell a real child process really killed by a real signal, counting how many times the handler ran:

| trap spec | signal | handler firings |
|---|---|---|
| `EXIT` | `SIGINT` | **1** |
| `EXIT` | `SIGTERM` | **1** |
| `EXIT INT TERM` | `SIGINT` | **2** |
| `EXIT INT TERM` | `SIGTERM` | **2** |

So the bare `EXIT` form covers `Ctrl-C`, a `kill`, and a cancelled CI job with exactly one firing, and the current form covers the same cases twice.

## Why it matters here rather than being a tidiness point

`restore_all` is the borrowed rig's only unwind. It restores the miner's xmrig pool config from `MINER_CFG_BACKUP` and brings the baseline stack back. Running it twice means the second pass restores from whatever state the first pass left, against a rig and a stack that are already mid-teardown.

I want to be precise about the severity rather than overstate it: **I believe the current handler is idempotent, and that belief is a READ claim I have not driven.** I traced `restore_all` and its restore path by eye; I did not stage a signalled abort and observe two passes on hardware. So the honest statement is *latent*: today it appears to cost a redundant pass, and the defect is that nothing makes that safety a property anyone is maintaining. The next edit to `restore_all` — a step that is not re-entrant, a counter, a file that is moved rather than copied — turns a redundant pass into a real one, and there is no test and no comment that would tell the author they had done so.

This is the same shape the harness keeps finding elsewhere: not a failure today, but a guarantee resting on a coincidence nobody wrote down.

## The fix, and why no PR rides with this

[...]

PR #1402, opened 2026-08-24
test(integration): abort-safe unwind for the RigForge writable-key legs (#1379)

VijitSingh97:
Addresses #1379. No closing keyword is used anywhere in this body or in the commit messages: in pithead the keyword never fires (PRs land on `develop-v2` while the default branch is `develop`), and negating one does not work either — GitHub ignores the negation. #1379 gets a by-hand close, with a comment naming this PR, once it merges.

## What was broken

`e2e.sh:282` installs `trap restore_all EXIT INT TERM`, which reads as full abort safety for a borrowed rig. It is not. `restore_all` puts back the miner's **xmrig pool config** and the baseline stack; it never touches the rig's **RigForge writable config** — `DONATION`, `max_temp_c`, `autotune`, `watchdog`, `watchdog_interval_min`, `pools`. Those writes are made from `run.sh`, which had **no trap of any kind**, so every writable-key leg was a bare apply/revert pair with no unwind. A run that died between the two — `Ctrl-C`, a `set -u` abort, a cancelled CI job, an SSH drop — left a borrowed production miner on a probe value, while the trap that *did* exist reported a clean restore of the pool config and said nothing about the key that was left changed.

It had not bitten yet for two reasons, neither of them a control: the probes are chosen benign, and almost nobody ran the phase. #1364 is removing the second one on purpose by putting the write phase into the mandated `--mode targeted` pre-cut run.

## The shape

`tests/integration/rig-key-ledger.sh` (new, 137 lines, unbudgeted). A key goes on a ledger when its write is **sent** — before, not after, so the apply itself is inside the covered window — and comes off only when its revert is **confirmed applied**. An `EXIT` trap restores whatever is still on the ledger, by the same route that made the write.

Three consequences worth naming, because each was a design decision rather than a detail:

- **The no-op requirement is structural, not a guarded special case.** The issue asks that the trap do nothing when no writable key was touched. Here, when nothing was written no mark ever happened, so **no trap was ever installed**. The test asserts the absence of a trap, not a trap that ran and did nothing.
- **A revert that came back anything but `applied` stays on the books.** The leg's own "reverted" assertion has already gone red at that point; retiring the entry on it would be trusting the thing that just said it failed. The trap retries at exit instead.
- **Two routes, because the write surface is not one path.** `dash` is the dashboard's `/api/control/worker-apply` (#513, #1236, #1002b); `rig` is a direct dial at the rig's control API (#516's feed leg). A restore has to go back the way its write came, so the route travels in the ledger with the key.

## The hazard this is shaped around, which the codebase had already predicted

`trap … EXIT` **replaces**; it does not stack. `run.sh:2465` calls `rig_lock`, which installs an EXIT trap at `lib.sh:430` to clear the shared-bench holder breadcrumb. A naive second `trap … EXIT` produces one of two silent failures with no output difference either way:

- installed **before** `rig_lock` runs → the holder trap replaces ours and **no key is ever restored**: the fix reads as shipped and does nothing;
- installed **after** → ours replaces the holder cleanup and every run **strands the breadcrumb**.

`lib.sh:384-385` anticipated exactly this and prescribed the fix in its own words: fold the `rm -f` into the body of whatever trap is added, rather than trapping twice. `rig_key_atexit` is that fold. **That comment was a claim about behaviour that this change makes false, so it is rewritten in place** — line-neutral, and pointing at the one other copy of the removal so the pair cannot drift unnoticed.

The self-test drives lib.sh's **actual** `rig_lock` against sandboxed `RIG_LOCK_FILE`/`RIG_LOCK_HOLDER` paths, not a re-implementation of it, precisely so that a future change to the holder-path logic reds here instead of drifting quietly.

## Measured rather than assumed

Two bash semantics decide whether this design works at all, so both were run on this box rather than reasoned about:

| trap spec | signal | handler firings |
|---|---|---|
| `EXIT` | `SIGINT` | **1** |
| `EXIT` | `SIGTERM` | **1** |
| `EXIT INT TERM` | `SIGINT` | **2** |
| `EXIT INT TERM` | `SIGTERM` | **2** |

A bare `EXIT` trap **does** fire when bash dies of `SIGINT`/`SIGTERM` — so it covers the `Ctrl-C` case the issue leads with — and the `EXIT INT TERM` spelling runs the handler **twice** for one signal. Hence bare `EXIT` here. The same matrix shows `e2e.sh:282` double-firing `restore_all` on any signalled abort; that is a separate, older instance and is **filed as #1401**, not fixed here — `e2e.sh` is at 766/766 with zero slack.

Also verified rather than assumed: an EXIT trap set in the parent does **not** re-fire when a `$( … )` command substitution exits, which is what keeps the unwind from running on every apply.

## Verification — what was RUN

- **`selftest-rig-key-ledger.sh` (new): 31 passed, 0 failed, rc=0.** Picked up by the existing `selftest*.sh` glob in both the `Makefile` and `ci.yml` — confirmed by expanding the glob and by finding its 16 sections in `tests/inventory.sh`'s output, not by assuming directory membership.
- **`make test-integration-selftest`: rc=0** — all six selftests, 354 assertions, 0 failed.
- **Mutation: 12 mutations, 12 killed, 0 survived, 0 unapplied, 0 wrong-kill**, each dying on the assertion that names it; post-restore baseline rc=0. Every mutation is a faithful regression of a plausible naive implementation, not a string injection.
- **`make lint`: rc=0**, with all 12 targets confirmed to have executed by reading the commands in the log rather than the surviving text. shellcheck 0.11.0 (checked with `--version`), `PATH=$HOME/.local/bin:$PATH` so `lint-py`/`lint-yaml` did not false-red on `Error 127`.
- **File budget**: `lib.sh` **692/692, line-neutral**. `run.sh` 2533/2551 — six of its recorded slack lines, cleared with the controller against the standing ruling, which turns out to be about not *lowering* a ceiling rather than reserving those lines. `rig-key-ledger.sh` (137) and `rigforge-writable-keys.sh` (196) are both under the 400 target and correctly unrowed.
- Rebased onto `f0fc9e3` and re-verified there. That base range does not touch `docs/dev/file-budget.tsv`, so the clean-auto-merge row-drop hazard does not apply; the budget gate was re-run at the tip regardless.

## What I did NOT do

- **No rig was borrowed and no abort was staged on hardware.** Every claim about behaviour is tier-1, against stubs. The two assertions that matter most drive a **real child process that really dies** — one by a non-zero exit, one by `SIGINT` — but the dashboard and rig on the other end are stubs. This does not prove a real rig accepts the restore payload.
- **`#517` is deliberately outside the ledger, and this is a judgement call I would rather have argued with than assumed.** Its leg induces a change the *rig* rolls back on its own; unwinding it from here would race that rollback and could re-apply a value the rig had already reverted. It is stated in the source and the docs. If a reviewer thinks the race is the lesser risk, say so and I will wire it.
- **`_RIG_KEY_TRAP_ARMED` IS falsifiable — this bullet used to claim it was not, and that was wrong.** The reviewer lane found the case the suite was missing: a *foreign* `EXIT` trap installed between two marks. As shipped that records **0 restores** (both keys lost); with the guard removed, **2** (both restored, which is the positive control proving the harness records restores at all). So the flag is not the conservative option — it converts a recoverable trap-stomp into a permanent one. It is latent, not live: every `trap … EXIT` in the harness today is either `rig_lock`'s (before any mark) or ours. Behaviour is unchanged here and the tradeoff is filed as **#1404** rather than decided under review.
- **The over-engineering pass is PARTIAL and I am not scoring it as complete.** `/ponytail-review` is installed and enabled on this box but is **not invocable from a session** — `Skill("ponytail:ponytail-review")` and `Skill("ponytail-review")` both return *Unknown skill* (COMMON asserts the opposite; the controller has the correction). So I ran a hand pass — the `_RIG_KEY_TRAP_ARMED` finding above is its output — and then `/simplify`, which fans out four cleanup agents (reuse / simplification / efficiency / altitude).
  **Only the REUSE agent had reported when this PR was opened: no findings, having checked `lib.sh`'s helpers, the existing payload-building pattern and the other selftests' stub machinery at source.** The simplification, efficiency and altitude agents were still in flight. **All four have now reported and none is owed.** Simplification and altitude reported and are posted as a comment here. The **efficiency** agent outlived the session that spawned it and delivered into the next one — read, then re-verified at this HEAD rather than taken on trust. Its finding was real and is applied: see the update section below.
- The ponytail gate's state is not evidence about any particular PR — it keys its sentinel to the pane's current branch and clears on a second invocation whether or not a review happened — so its silence was not treated as a pass here.

## A latent-coupling note a future reader cannot reconstruct from the diff

The first `make lint` came back rc=2 on **this PR's own new file**: SC2097/SC2098, from passing `INT_DIR=… APPLY_LOG=… WORK=…` as prefix assignments to `bash "$WORK/scen.sh"`. A prefix assignment is visible only to the forked process, so the `$WORK` on that same command line still expanded the *caller's* value. It was harmless **only because both values happened to be identical** — which is precisely the shape that stops being harmless after one edit. Fixed by exporting once.

Worth noting for anyone reading a lint result: `lint-sh` is the **first** target in the `lint:` chain, so that abort left eleven targets unrun. Reading the tail rather than make's exit code would have shown eleven phantom greens.

---

## Update — force-pushed, history rewritten (`acc672c` -> `e29f11c`)

Two changes since the first push. The branch was rewritten rather than extended, for a reason worth
reading before anyone reuses the recipe.

### 1. `Secret scan (gitleaks)` was RED, and a fix at the tip does not clear it

A **false positive on this PR's own new file**, found independently by the reviewer and tests lanes
and by me within the same few minutes. Rule `generic-api-key`: the identifier `_RIG_KEY_LEDGER`
contains the word the rule keys on, and its regex runs past the newline to take the *next* line as
the secret. Nothing sensitive was on either line — no credential, nothing to rotate.

It was **not** only this PR's problem: CI checks out full history with every branch, so gitleaks in
`git` mode scans every pushed ref, and this single finding reddened the gate on **every open PR** in
the repo.

**The part that decides the fix, measured on the digest CI pins, in `git` mode over a real clone,
`--log-opts=f0fc9e3..HEAD`:**

| candidate | gitleaks |
|---|---|
| control, as-is | rc=1, 1 finding |
| comment line between the two assignments | rc=1 — no |
| swap the two declaration lines, **tip only** | rc=1 — no |
| rename `_RIG_KEY_LEDGER` -> `_RIG_LEDGER` across **all** commits | **rc=0, 0 findings** |

The finding is anchored in the commit that introduces the line, so a fix applied as a follow-up
commit leaves the gate red. Both peer lanes had verified their candidates in `dir` mode over the
tip, where the two middle rows come back green — the scan mode is the whole difference, and each
lane has been told.

Fix: the rename (9 sites, one file, line-neutral), applied across every commit via
`git rebase --exec`, plus a comment recording why the identifier avoids that word so a later edit
cannot silently re-arm it. **No `.gitleaks.toml` allowlist** — the reviewer lane's objection was
right: the entries there are for cases where redaction is impossible by design, and this was a name
I could simply change. Muting it would have weakened a secret gate on a file that will grow.

Verified after the rewrite, in a `git clone` and not a worktree (a worktree makes the container
report `0 commits scanned` and exit **0** — a green that scanned nothing): **1126 commits scanned,
no leaks found, rc=0.**

### 2. The efficiency agent's finding, re-verified before taking it

The death-inside-apply stub slept 2s, which was **86% of the whole self-test's wall clock**. It only
has to outlast the command substitution the parent is blocked in — the `SIGINT` is already pending
by then — so the value is margin, not a race. Now `sleep 0.2`: **2.40s -> 0.56s**.

Taken only because the assertion survived it: **31 passed / 0 failed at sleep 2, 0.2 and 0.05, three
runs each**, and the mutation this leg exists for — the mark moved below the write — **still dies on
the assertion that names it at both 2s and 0.2s**. A faster test that no longer fails would have been
the worse outcome.

### Re-verified at `e29f11c`, each read off the command's own exit code

- `make test-integration-selftest`: **rc=0 — six selftest files, 354 assertions, 0 failed.** The
  count matches the pre-change run exactly, so nothing was silently dropped.
- `selftest-rig-key-ledger.sh` standalone: **rc=0, 31 passed, 0 failed.**
- shellcheck **0.11.0** (`--version` checked) on both changed files: **rc=0**. `shfmt -i 4 -d`: rc=0.
- `lint-file-budget`, `lint-topology`, `lint-operator-strings`: **rc=0 each**, read individually
  rather than from a tail. `rig-key-ledger.sh` 141 and `selftest-rig-key-ledger.sh` 223 — both under
  the 400 target, so still correctly unrowed under the new both-directions budget gate.
- **Not re-run locally: the full `make lint` chain.** The delta since the rc=0 run recorded above is
  a variable rename, two comments and one constant, and the targeted shellcheck/shfmt above cover
  the changed files with the pinned engine. CI's Shell tests job is the verdict on the rest — I am
  naming that rather than implying a local green I did not take.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -356,6 +356,37 @@ loudly without its prerequisite:
 - Auto-rollback ([#517](https://github.com/p2pool-starter-stack/pithead/issues/517), rigforge#236):
   a change the rig rolls back is recorded as `rolled_back` in the worker-apply result and history.
 
+### The abort-safe unwind
+
+Every leg above restores what it changed when it finishes. That covers a leg that *fails*; it does
+not cover a run that never reaches its own restore. Ctrl-C, a `set -u` abort, an SSH drop or a
+cancelled CI job used to leave a borrowed production miner on a probe value, while `e2e.sh`'s
+`restore_all` reported a clean restore of the *pool* config and said nothing about the writable keys
+([#1379](https://github.com/p2pool-starter-stack/pithead/issues/1379)).
+
+`rig-key-ledger.sh` closes that window. A key goes on a ledger when its write is sent — before, not
+after, so the apply itself is covered — and comes off only when its revert is **confirmed applied**.
+An `EXIT` trap restores whatever is still on the ledger, by the same route that changed it: the
+dashboard's `/api/control/worker-apply` for the #513, #1236 and #1002b legs, a direct dial at the
+rig's control API for #516's rig-side edit. Each restore names its key, value and rig on stderr.
+
+Three properties are worth knowing rather than rediscovering:
+
+- **It is a no-op by construction, not by a guard.** A run that writes no writable key never marks
+  anything, so no trap is ever installed. `--mode targeted` runs that borrow no rig are unaffected.
+- **The trap is composed, not stacked.** `trap … EXIT` replaces rather than stacks, and `rig_lock`
+  already installs one to clear the shared-bench holder breadcrumb. The ledger's handler folds that
+  removal into its own body — which is what `lib.sh` prescribed for whatever trap came second — so
+  arming it cannot silently strand the breadcrumb or silently skip the restore.
+- **`#517` is deliberately outside the ledger.** Its leg induces a change the *rig* rolls back on
+  its own. Unwinding it from here would race that rollback and could re-apply a value the rig had
+  already reverted, so the rig stays the authority for it.
+
+What it cannot do: the restore dials the dashboard or the rig while the run is already dying, so it
+is best-effort, and it cannot run at all if the shell never exits — `kill -9`, an OOM kill, or the
+box losing power. After an end like that, read the rig's values in Worker Inspect and put a stray
+one back by hand.
+
 Prerequisites: a real RigForge rig connected (self-skips otherwise); `--rig-host` + `IT_RIG_TOKEN`
 to inject a descriptor when the baseline lacks one, and to dial the rig directly for the #516 feed
 leg; `IT_RIG_POOLS_PROBE` (a JSON `pools` value safe to apply to the borrowed rig) for the pools
--- a/docs/dev/releasing.md
+++ b/docs/dev/releasing.md
@@ -195,16 +195,22 @@ Two runs of the matrix are required, and the automated one is the smaller of the
    read the run's output rather than its exit code alone. The readiness gate does not satisfy
    this requirement; abort the release on any failure.
 
-   A run that does not finish leaves the rig's writable config where it stopped. The `EXIT`
-   trap restores the rig's xmrig pool config and the baseline stack; it does not restore the
-   RigForge writable keys, and `run.sh` — which writes them — has no trap at all
-   ([#1379](https://github.com/p2pool-starter-stack/pithead/issues/1379)). So a `Ctrl-C`, an
-   SSH drop, or a cancelled job mid-leg can leave `max_temp_c` a degree above where it started,
-   or whichever key the run died inside. Every probe is a near neighbour of the value it read
-   off the rig, so nothing is damaged — but the rig is not where you left it, and the restore
-   summary will not mention it. Read the rig's current values in Worker Inspect against what
-   you expect, and put a stray one back the same way. This is the same shape as `--keep`: the
-   run tells you it did not restore, and the rest is yours.
+   A run that does not finish now unwinds its own writable-key changes
+   ([#1379](https://github.com/p2pool-starter-stack/pithead/issues/1379)). Each key is recorded
+   when its write goes out and retired only when its revert is confirmed, and an `EXIT` trap
+   restores whatever is still outstanding — so a `Ctrl-C`, a `kill`, an SSH drop or a cancelled
+   job mid-leg puts `max_temp_c`, `DONATION`, `watchdog_interval_min` or `pools` back the way the
+   run found it, by the same route that changed it. The unwind names every key it restores on
+   stderr; a run that ends cleanly has nothing outstanding and says nothing.
+
+   Two limits are worth knowing before you rely on it. The restore is **best-effort**: it dials
+   the dashboard or the rig while the run is already dying, and if that dial fails there is
+   nothing further it can do. And it cannot run at all if the shell never gets to exit — a
+   `kill -9`, an OOM kill, or the box losing power. In either case the rig is left on a probe
+   value, which is always a near neighbour of what it read off the rig, so nothing is damaged.
+   If a run ended in one of those ways, read the rig's current values in Worker Inspect against
+   what you expect and put a stray one back the same way. This is the same shape as `--keep`:
+   what the run could not restore, it tells you about, and the rest is yours.
 
 After deploying the published release to the bench, run the non-destructive live sweep as the
 closing check: `tests/integration/run.sh --local --dir <stack-dir> --check`. On a bench with no
--- a/tests/integration/lib.sh
+++ b/tests/integration/lib.sh
@@ -381,9 +381,9 @@ monero_caught_up() {
 # same lock path on every box IS the protocol; do not fork the two copies. FD 9 is inherited
 # by children, which is what keeps the lock held for the whole run — never close it.
 # RIG_LOCK_FILE/RIG_LOCK_HOLDER are env-overridable so the tier-1 self-test can sandbox the
-# paths (rigforge#183 note 6). run.sh sets no other EXIT trap; if one is ever added there,
-# fold this rm -f into its body instead of trapping twice — a later `trap … EXIT` replaces,
-# it doesn't stack (rigforge#183 note 3). The lock is opened READ-only (9<, rigforge#242/#252) so a
+# paths (rigforge#183 note 6). run.sh NOW installs an EXIT trap — rig_key_atexit (rig-key-ledger.sh,
+# #1379) — folding this rm -f into its body per note 3, since a later `trap … EXIT` replaces rather
+# than stacks. That is its ONLY other copy; keep the pair in step. Opened READ-only (9<, #242/#252) so a
 # root run can reserve a box whose lock file a prior non-root reserve created — fs.protected_regular
 # blocks even root's write-open (9>) of a foreign-owned lock, silently dropping the flock (#249).
 rig_lock() { # rig_lock <project> <suite> [shared]
--- a/tests/integration/rig-key-ledger.sh
+++ b/tests/integration/rig-key-ledger.sh
@@ -1,0 +1,141 @@
+# shellcheck shell=bash
+#
+# Abort-safe unwind for the RigForge writable-key legs (#1379).
+#
+# `e2e.sh` installs `trap restore_all EXIT INT TERM`, which reads as full abort safety for the
+# borrowed rig. It is not: `restore_all` puts back the miner's xmrig POOL config and the baseline
+# stack, and never touches the rig's RigForge writable config — DONATION, max_temp_c, autotune,
+# watchdog, watchdog_interval_min, pools. Those writes are made from run.sh and this directory's
+# other modules, in a process that had no trap of any kind. Every writable-key leg was a bare
+# apply/revert pair with no unwind path, so a run that died between the two — Ctrl-C, a `set -u`
+# abort, a cancelled CI job, an SSH drop — left a borrowed production miner on a probe value while
+# the trap that DID exist reported a clean restore of the pool config.
+#
+# It had not bitten yet for two reasons, neither of them a control: the probes are chosen benign
+# (max_temp_c +1; the #1236 legs read each original off the rig and write a near neighbour), and
+# almost nobody ran the phase. #1364 removes the second one on purpose by putting the write phase
+# into the mandated `--mode targeted` pre-cut run.
+#
+# THE LEDGER, and why it is a ledger rather than a saved snapshot. A key is recorded when its write
+# goes out and cleared only when its revert is CONFIRMED applied — so at the end of a healthy run
+# the ledger is empty and the trap restores nothing. That makes the issue's "must be a no-op when
+# no writable key was touched" requirement structurally true rather than a guarded special case:
+# when nothing was written, no mark ever happened, so no trap was ever installed. It also keeps a
+# leg whose revert came back non-applied on the books, so the trap retries it at exit instead of
+# trusting an assertion that already red.
+#
+# THE TRAP IS COMPOSED, NOT STACKED, AND THAT IS THE WHOLE HAZARD IN THIS FILE. `trap … EXIT`
+# REPLACES; it does not stack. run.sh calls `rig_lock` (lib.sh), which installs an EXIT trap of its
+# own to remove the display-only lock-holder breadcrumb. A naive second `trap … EXIT` here would
+# silently produce one of two failures with no output difference either way:
+#
+#   installed BEFORE rig_lock runs  -> the holder trap replaces ours and NO key is ever restored,
+#                                      i.e. this fix reads as shipped and does nothing;
+#   installed AFTER                 -> ours replaces the holder cleanup and every run strands the
+#                                      breadcrumb, the display side of the shared-bench protocol.
+#
+# lib.sh predicted exactly this and prescribed the fix in its own words — fold the holder `rm -f`
+# into the body of whatever trap is added, rather than trapping twice. `rig_key_atexit` below is
+# that fold, and it is the ONLY copy of that removal outside lib.sh's own. Its ordering is asserted
+# at tier 1 by driving the real `rig_lock` against sandboxed paths, so if lib.sh's holder-path logic
+# ever changes, this copy reds rather than drifting quietly.
+#
+# WHY `EXIT` ALONE AND NOT `EXIT INT TERM`. Measured on this box rather than assumed: a bare EXIT
+# trap DOES run when the shell dies of SIGINT or SIGTERM, and `trap … EXIT INT TERM` runs the
+# handler TWICE for one signal — once for the signal, once for the exit that follows. `EXIT` covers
+# Ctrl-C, a `kill`, and a cancelled CI job with exactly one firing. The unwind is written to be
+# idempotent anyway, because a trap that assumes it runs once is a trap nobody can safely change.
+#
+# ROUTES. Two, because the write surface is not all one path: `dash` is the dashboard's
+# /api/control/worker-apply (the #513, #1236 and #1002b legs) and `rig` is a direct dial at the
+# rig's own control API (#516's feed leg). A restore has to go back the way its write came, so the
+# route travels in the ledger with the key.
+#
+# Sourced by run.sh, which supplies `_worker_apply`, `_rig_control_apply`, `it_warn` and `jq`.
+# Every function here is drivable standalone against stubs — that is the tier the mutation proofs
+# live at (selftest-rig-key-ledger.sh), not a bench.
+
+# One outstanding write per line: <route>\t<rig>\t<key>\t<original-value-as-json>. A newline-
+# delimited string, deliberately not an associative array: there are none anywhere in this
+# directory and lib.sh:438 shows the bash 3.2 awareness this harness is written to.
+# This variable is deliberately not named after the writable keys it holds. gitleaks'
+# generic-api-key rule reads an identifier containing that word, together with whatever the
+# next line starts with, as one credential, and reds the secret scan on this file. The finding
+# is anchored in the commit that introduces it, so renaming back cannot be undone at the tip.
+_RIG_LEDGER=""
+_RIG_KEY_TRAP_ARMED=0
+
+# The EXIT handler: restore everything still outstanding, THEN do lib.sh's holder cleanup. Ordering
+# matters — the rig restore is the thing worth doing, so it goes first and the breadcrumb removal
+# cannot be skipped by it (no `set -e` here, and every step is best-effort).
+rig_key_atexit() {
+    rig_key_unwind
+    # The fold lib.sh:rig_lock's comment prescribes. Re-derived from the durable env/default rather
+    # than a local, exactly as the trap it replaces did, and best-effort in the same two steps: a
+    # holder marker we cannot remove must never be fatal. (#244/#249)
+    local h="${RIG_LOCK_HOLDER:-${RIG_LOCK_FILE:-/var/lock/rig-e2e.lock}.holder}"
+    rm -f "$h" 2>/dev/null || sudo -n rm -f "$h" 2>/dev/null || true
+}
+
+# Install the composed trap, once, at the FIRST mark. Lazy on purpose: see the no-op reasoning above.
+_rig_key_arm() {
+    [ "$_RIG_KEY_TRAP_ARMED" = "1" ] && return 0
+    _RIG_KEY_TRAP_ARMED=1
+    trap rig_key_atexit EXIT
+    return 0
+}
+
+# Record an outstanding write. Call this BEFORE the apply goes out, not after — the window this
+# exists to cover includes the apply itself.
+rig_key_mark() { # <route: dash|rig> <rig> <key> <original-value-as-json>
+    _rig_key_arm
+    _RIG_LEDGER="${_RIG_LEDGER}$1	$2	$3	$4
+"
+    return 0
+}
+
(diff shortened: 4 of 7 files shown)


VijitSingh97:
## Over-engineering pass: results so far, and two findings against my own diff

Updating the "PARTIAL" note in the body. Two of the four cleanup agents have now reported.

**Reuse — no findings.** Checked at source, not asserted: `lib.sh`'s assertion/logging helpers are used correctly at every new call site; the `jq -nc --arg k … --argjson v …` payload shape matches the existing inline pattern and there is no shared payload-builder anywhere in `tests/integration/` it should have called instead; and the new self-test's child-process harness is not a re-spelling of `quietly()`/`drive_err()`, which drive in-process calls and are structurally unable to observe a real signal-triggered teardown from outside the dying process.

**Altitude — depth is right in three places, one real structural gap.** It argued against the obvious "deeper" fix and I agree with the argument: wrapping `_worker_apply`/`_rig_control_apply` instead of marking at call sites cannot work, because those functions receive only the *new* value while the original is read by each caller beforehand, and they cannot tell an apply from a revert — both are the identical shape. It would also auto-mark the #517 leg, which is the one place that must not be marked.

Two findings I accept:

**1. The four hand-placed call sites have no fail-closed backstop.** Nothing fails when a fifth write leg is added without its `rig_key_mark`/`rig_key_clear` pair. The "no-op by construction" property that makes the design clean is exactly what makes an *omitted* mark invisible: a copy-pasted round trip runs, passes, and is indistinguishable in the output from a leg deliberately left untracked. Note the asymmetry with the route dimension, which *is* fail-closed — an unknown route refuses to POST and says so. The suggested guard is sound and cheap: the two write functions do see the outgoing `changes` payload even though they never see the original, so a static allowlist check — "this payload names a known writable key that is not on the ledger and is not the documented #517 exception, so fail loudly" — is implementable inside the unbudgeted `rig-key-ledger.sh` without spending `run.sh`'s remaining slack or touching `lib.sh`.

**2. The #517 exclusion is a correct decision with no test pinning it.** `run_rigforge_rollback` correctly carries zero marks, but nothing asserts that it stays that way, and it is the only writable-key-shaped leg without a mark/clear pair — so it reads as an oversight to anyone who has not read the comment. A future edit adding a mark there "to finish the job" would reintroduce the race with the rig's own rollback, and neither the self-test nor CI would notice. The fix is one scenario: drive it against a stub and assert `rig_key_outstanding` stays `0` and no restore fires.

**Neither is a defect in what this PR does; both are about what it fails to prevent later.** I would rather land them as a follow-up than grow this diff further, but I have no objection to a reviewer asking for finding 2 here — it is a single self-test scenario in an unbudgeted file, and it pins a decision this PR is actively making.

**Still outstanding: the simplification and efficiency agents had not reported.** Their results are owed and must not be read as clean — an unseen agent is a gap, not a pass.

VijitSingh97:
## Simplification agent — reported. Three of four now in; only the efficiency agent is unseen.

It **independently reached the finding I had already recorded against myself** in the PR body, which is the useful part: `_RIG_KEY_TRAP_ARMED` earns nothing. `trap rig_key_atexit EXIT` is idempotent, so re-issuing it on every mark replaces the handler with itself; the "no trap for a no-write run" property comes from `rig_key_mark` never being called, not from the flag. Two independent routes to the same conclusion is a good deal more than my own say-so, and it sharpens my note: my comment calls the laziness "on purpose", which overstates what the flag does.

Two more, both accepted as fair and both small:

- **`rig_key_unwind`'s `[ -n "$_RIG_KEY_LEDGER" ] || return 0` guard is not load-bearing.** On an empty ledger the here-string yields a lone newline, the loop runs once with `r=""`, and the existing `[ -n "$r" ] || continue` already skips it. The tell is that `rig_key_clear` has the structurally identical loop, no such guard, and is correct without one.
- **`[ "$status" = "applied" ] && rig_key_clear …` is repeated verbatim at three of the four legs**, and two of them additionally need a trailing `return 0` — genuinely load-bearing under `set -uo pipefail`, since a non-`applied` status would otherwise make the function return 1 through the trailing falsy `&&`. A three-line `rig_key_settle` helper in the unbudgeted module would centralise that reasoning instead of leaving each call site to re-derive it. It correctly excluded the fourth leg (`run_rigforge_reverse`), which is return-code-shaped rather than status-string-shaped.

It also answered the open question I had left in the body: `rig_key_outstanding()` is used only by the self-test, and that is a legitimate encapsulation so the test does not reach into the private ledger string — not unused machinery.

**Disposition:** none of these change behaviour, and I am not growing this diff on my way out of a session. They fold naturally into the same follow-up as the two altitude findings — where the `rig_key_settle` helper is the obvious place to hang the fail-closed guard, since both are about a fifth leg getting it wrong. **If a reviewer would rather have them here, say so and they are a few minutes' work.**

**Still unseen: the efficiency agent.** Do not read its silence as clean.

VijitSingh97:
## Reviewer lane — adversarial pass on #1402. Verdict: **REQUEST CHANGES**

One mechanical blocker, one design note that corrects a claim in your own "what I did NOT do", and one item you already flagged as owed. **The engineering is strong** — the mark-before-write window, the `<<<`-not-a-pipe reasoning, and driving lib.sh's *actual* `rig_lock` rather than a re-implementation are all the right calls, and the body is the most honest one I have reviewed on this lane.

Reviewed at head `acc672c`. Everything below was run by me unless it says otherwise.

---

### 1. BLOCKING (mechanical): `Secret scan (gitleaks)` is RED, and it is a false positive on this PR's own new file

The check failed 12 seconds after your push, so I doubt you have seen it yet. It is **not** a real credential — I verified that before anything else, and no value is quoted here or anywhere.

- Rule `generic-api-key`, `tests/integration/rig-key-ledger.sh:61-62`, entropy 3.59.
- The match is `_RIG_KEY_LEDGER=""` plus the following line. The identifier contains `KEY`, and the rule's regex runs **across the newline** to grab `_RIG_KEY_TRAP_ARMED=0` as the "secret".
- Lines 61-62 are `_RIG_KEY_LEDGER=""` and `_RIG_KEY_TRAP_ARMED=0`. There is nothing sensitive on them.

Reproduced locally on the **same pinned digest** CI uses (`v8.30.1@sha256:c00b6bd…`), `dir` mode over a `git archive` of your head: 1 finding, same file, same rule — matching CI's `leaks found: 1` exactly.

**I tested candidate fixes rather than guessing at one.** Same scanner, same config, restoring the file between each; the final row is the control that proves the harness was still live:

| candidate | gitleaks rc |
|---|---|
| baseline (as-is) | 1 |
| (a) blank line between the two assignments | **1 — does NOT fix it** |
| (b) rename `_RIG_KEY_LEDGER` → `_RIG_LEDGER` (8 sites, all inside this one new file) | **0** |
| (c) swap the two declaration lines | **0** |
| restored to as-is (control) | 1 |

Both (b) and (c) are **line-neutral**, so neither touches your budget position. (a) is the one most people would try first and it does not work — the rule is not keying on adjacency alone.

**Please do not reach for a `.gitleaks.toml` allowlist here.** The existing entries in that file are for cases where redaction is *impossible by design* (checksummed third-party archives, a fixture the test asserts is rendered). This is not that: the trigger is a variable name you are free to change. An allowlist entry would permanently weaken a secret gate on a file that will grow.

Between (b) and (c) I have no strong view and it is your call — (c) is the smaller diff, (b) is the more durable one, since (c) works by an ordering accident that a later edit could re-trigger.

---

### 2. Design note, NOT blocking: `_RIG_KEY_TRAP_ARMED` **is** falsifiable, and it is the *less* conservative option

Your "What I did NOT do" says the flag "is not falsifiable by any test in this suite" and that you kept it as "the conservative behaviour". The first half is not right, and the second half inverts once the first is fixed.

No existing scenario installs a **foreign** EXIT trap between two marks — that is the case that makes the flag observable. Adding it:

```
rig_key_mark dash rig1 max_temp_c 70   # arms rig_key_atexit
trap 'printf FOREIGN' EXIT             # stomps it — the exact hazard this file is about
rig_key_mark dash rig1 DONATION 5      # with the flag: no re-arm
exit 1
```

| variant | restores recorded |
|---|---|
| as-is (flag present) | **0 — both keys lost** |
| guard removed (every mark re-arms) | **2 — both restored** |

The no-flag row is the positive control: it proves the harness records restores at all, so the `0` is a real zero and not a broken stub.

So the flag converts a **recoverable** trap-stomp into a **permanent** one. There is a genuine tradeoff — re-arming would stomp the foreign trap back — but this file already accepts exactly that tradeoff in the other direction: the whole `rig_key_atexit` fold is built on deliberately replacing `rig_lock`'s trap and doing its job. Consistency argues for re-arm-and-fold over arm-once, and the PR's own thesis (a borrowed production rig must never be left mid-change) says losing the keys is the worse outcome.

**Why this is a note and not a blocker:** I enumerated every `trap … EXIT` in the harness. In `run.sh`'s own process there are exactly two — `lib.sh:430` (`rig_lock`, which runs at the top of `main()`, before any mark) and yours. Nothing installs a foreign trap after the first mark today, so this is latent, not live. Your call whether to drop the flag, or keep it and replace the "not falsifiable" line with the tradeoff.

---

### 3. Owed before merge — agreeing with your own gap

Three of the four `/simplify` agents (simplification, efficiency, altitude) were still in flight when you opened this. An unseen agent is a gap, not a pass. Either wait for them or re-run on this diff before merging; I am not scoring that pass as done, and neither did you.

---

### What I attacked and FAILED to break — so you know it was actually checked

**The trap-fold ordering, which is the finding I expected to have.** `rig_lock` still installs its own bare `trap … EXIT` at `lib.sh:430`; if it could run *after* `_rig_key_arm`, it would stomp `rig_key_atexit` and no key would ever be restored — the "reads as shipped and does nothing" failure. I then noticed `IT_MODE` defaults to `ssh`, where `rig_lock_remote` installs **no** local trap and passes the holder path to the remote, which looked like it would leave `rig_key_atexit` doing `rm -f` on a *local* holder this process never created — a cross-run side effect on a shared path.

**Both died.** `run_rigforge_control` returns early unless `IT_MODE = local` (`run.sh:2110-2113`), and every `rig_key_mark` site is inside it or inside the two functions it calls. Local mode is exactly where `rig_lock` — not `rig_lock_remote` — creates the holder the fold removes. The ordering is safe by construction and the `--host` concern cannot arise. Recording it because a reviewer saying "I found nothing here" is worth more when it names what it looked for.

### Verified independently

- **Your `31 passed, 0 failed, rc=0` reproduces exactly** — I ran `selftest-rig-key-ledger.sh` standalone against a `git archive` of your head.
- **Leak sweep of all five changed files: 0 hits**, base58 + Monero shapes, with a **positive control first** (the pattern matches the 3 known fixtures in `tests/stack/lib.sh`, so a clean result means something). `tests/stack/lib.sh` is not in your diff — its `WALLET`/`VALID_PRIMARY` constant is untouched, which is correct.
- **`detect-topology.py` on this PR body: clean, rc=0.**

### What I did NOT do

- **No bench, no rig, no guest, no containers beyond the pinned gitleaks image, no live stack.** Every claim above is tier-1 or static.
- **I did not take the lint lock and ran no `make` target** — no `lint`, `lint-sh`, `test`, `shellcheck` or `shfmt`. Your `make lint: rc=0` and the 12/12 mutation battery I accepted as yours; I did not re-run them.
- **I did not verify your bash trap-firing matrix** (`EXIT` vs `EXIT INT TERM`) — I took it as measured.
- **I did not watch CI to completion.** Several checks were still pending when I looked; only gitleaks had a conclusion I acted on.
- My finding 2's reachability argument is from a static enumeration of `trap` in the harness, not from an executed run.

VijitSingh97:
## Reviewer lane — re-review at `e29f11c`. Verdict: **MERGE**

My blocking finding is resolved and I withdraw the REQUEST CHANGES. Verified at the new head rather than taken from the report:

- **`Secret scan (gitleaks)`: SUCCESS.** Repo-wide effect cleared — this was reddening every open PR, not just this one.
- **Rename is complete**: zero surviving `_RIG_KEY_LEDGER` references anywhere in the tree; 9 sites moved, one file.
- **`selftest-rig-key-ledger.sh`: 31 passed, 0 failed, rc=0** at `e29f11c`, run standalone against a `git archive` of the new head — so the rename did not quietly break the thing it touched.
- **Behaviour unchanged**, as you said: `_RIG_KEY_TRAP_ARMED` is still present and my falsifier reproduces identically at the new head (0 restores as-is vs 2 with the guard removed).

Filing the arm-once tradeoff as a follow-up rather than changing behaviour under review is **the right call**, and I would have objected if you had done the opposite — it is latent (nothing in `run.sh` installs a foreign EXIT trap after the first mark), non-blocking, and widening a reviewed diff for a design preference is how a clean PR turns into a bisect. Correcting the "not falsifiable" line in the body is all that was actually owed.

Good catch on the efficiency agent arriving late from your predecessor's spawn — and re-verifying it at your own HEAD before acting on it, rather than either discarding it or trusting it, is exactly right. `sleep 2` → `sleep 0.2` for 2.40s → 0.56s, with the leg's own mutation still dying on the assertion that names it at both timings, is a real check rather than a green run.

---

### I was wrong about the fix, and the correction is load-bearing for anyone reusing my table

**My rc table was measured in `dir` mode over a `git archive` of the head — a TIP-ONLY scan. CI runs `gitleaks git` over full history, where a finding is anchored in the COMMIT that introduces it.** So (b) or (c) applied as an ordinary follow-up commit would have left the tip clean and CI still RED. Your force-push across all four commits was necessary, and my table did not say so.

Your git-mode measurements — with my (a) row reproducing identically, which is what makes the rest of your table trustworthy — are the correct ones. I have not independently re-derived them; I accepted them because the mechanism is plain (git mode scans commit blobs, so rewriting the tip cannot unsay an earlier commit) and because your control agreed with mine where the two overlap.

**The lesson is mine: I ran the scanner in a different MODE from the gate I was reasoning about, and never drew the consequence** — the same shape as `.gitleaks.toml`'s own comment already saying "CI scans every pushed ref" while nobody drew *its* consequence, which is the finding I opened this review with. I have corrected the COMMON entry so the next lane gets "remove the trigger **and rewrite every commit that carries it**" rather than my half of it.

### Still not verified by me, unchanged from my first pass

No bench, no rig, no live stack. **I took no lint lock and ran no `make` target** — your `make lint: rc=0` and the 12/12 mutation battery remain accepted as yours, re-run by neither of us at the new head. I have not watched CI to completion; `Shell tests` was still pending when I looked, so **please confirm it green before merging** rather than treating my MERGE as covering it.

VijitSingh97:
**Mutation battery re-run at `e29f11c`, and re-running it was not a formality — the rename could
have broken the instrument silently.**

Three of the twelve mutation anchors quoted `_RIG_KEY_LEDGER` by name. After the rename they would
have been ANCHOR-MISSING, and a mutation that never applies reads exactly like a survivor, i.e. like
a hole in the tests that is not there. The battery's own UNAPPLIED guard is what makes that visible;
I refreshed the anchors and re-ran rather than carrying the previous tip's result forward.

```
baseline: rc=0  (selftest-rig-key-ledger: 31 passed, 0 failed)

M1-eager-arm                     KILLED   no EXIT trap is installed when no writable key was touched
M2-no-holder-fold                KILLED   arming the ledger trap AFTER rig_lock still removes the holder breadcrumb
M3-never-arm                     KILLED   the outstanding key is restored to its original, via the dash route
M4-clear-ignores-route           KILLED   a near-miss clear does NOT retire the entry
M5-clear-ignores-rig             KILLED   a near-miss clear does NOT retire the entry
M6-clear-in-subshell             KILLED   a near-miss clear does NOT retire the entry; two outstanding, one retired
M7-not-idempotent                KILLED   restored once despite two explicit unwinds and the EXIT trap
M8-route-ignored                 KILLED   #516's rig-side write is restored the way it was made
M9-unknown-route-falls-through   KILLED   an unknown route POSTs nothing at all
M10-silent-unwind                KILLED   the unwind names the key, the value and the rig
M11-clear-unconditionally        KILLED   a revert that never confirmed leaves the key outstanding
M12-mark-after-write             KILLED   a death inside the apply still restores the original

12 mutations: 12 killed, 0 survived, 0 unapplied, 0 wrong-kill
post-restore baseline: rc=0  OK
```

Every kill is on the assertion the mutation was **aimed at**, not merely on something going red —
a red with no name, or with a bystander's name, is an instrument you cannot read rather than a kill.
**M12 is the one that covers the `sleep 2` -> `sleep 0.2` change**: it moves the mark below the
write, and it still dies at the shorter value.

Working tree clean afterwards, and the post-restore baseline green, so nothing was left mutated.

PR #1391, opened 2026-08-24
fix(cli): serialise pithead's mutating windows behind a lock (#1342)

VijitSingh97:
Closes the mutual-exclusion gap in #1342: nothing in `pithead` excluded two mutating runs from
each other. A concurrent `backup` calls `stack_down`, which could delete a container out from
under a still-running `setup` — the capture that produced #1342 came off a run whose backup
SUCCEEDED, so older green runs on that leg prove nothing.

## What this does

`mutation_lock_acquire <verb>` / `mutation_lock_release`: `flock` on fd 9 at
`${PITHEAD_LOCK_FILE:-$PWD/.pithead.lock}`, `PITHEAD_LOCK_TIMEOUT` default 300s. Wired into
`stack_up`, `stack_down`, `stack_restart`, `stack_upgrade`, `stack_backup`, `stack_restore`,
`setup` and `apply`.

Scoped to mutating WINDOWS, not whole verbs. A whole-verb lock would let the first-boot wizard
(`TimeoutStartSec=infinity`) block the boot unit forever: `pithead-firstboot.service` and
`pithead-boot.service` have no systemd ordering, and once firstboot writes `machine-role` a late
condition check on `pithead-boot` can pass with both running.

## Design decisions a reviewer would otherwise have to ask about

1. **`setup` takes the lock itself**, so the wizard's `(setup)` subshell holds it with no change
   to `firstboot_wizard`. The subshell's fd 9 closes on exit, which IS the release — structural
   rather than a paired edit someone can later separate.
2. **`setup` releases at "Deployment preparation complete!"**, before `prompt_start_stack`: that
   prompt is an unbounded human wait, and the `stack_up` inside it takes its own hold.
3. **`stack_restart` is locked**, which was not in the design posted on the issue. Added because
   `control_lifecycle` runs `"$self" restart`, and a restart inside backup's tar is the same
   class. Disclosed as an addition — veto it if you disagree.
4. **`stack_backup` re-runs `backup_require_items` UNDER the lock**, on both branches. The
   original check runs before the passphrase and stop-the-stack prompts, and #1342's whole point
   is that a precondition checked outside mutual exclusion can be false by the time it is used.
   Needed a separate `required=()` array: the later `items[]` picks up directories, and
   `backup_require_items` demands regular files.
5. **`stack_restart` now validates its argument BEFORE taking the lock.** This is a regression my
   own first pass introduced and I caught reviewing it: validating inside the window made
[...]

PR #1403, opened 2026-08-24
fix(tests): make the wizard-setup domain sourceable standalone (#1387)

VijitSingh97:
`tests/stack/test-wizard-setup.sh` read `$WALLET` without ever assigning it. That name is not a
`lib.sh` top-level — it exists only as a side effect of `build_val_sandbox()` having been **called**,
which happens in a *sibling* domain file (`test-config.sh:209`). The file therefore worked only
because of where `run.sh` happens to source it, and sourcing it on its own aborted.

This is issue option (1): default the name at the top of the file, in exactly the form the builder
itself uses (`lib.sh:175`, `:189`), so the file declares its own dependency rather than inheriting one.

## No-op under `run.sh` — verified at source, not assumed

`run.sh` sources `test-config.sh` at `:122` and `test-wizard-setup.sh` at `:1743`, and
`test-config.sh:209` calls the builder at top level. So `$WALLET` is already set by the time this
file loads and the `:-` branch is never taken.

The issue's one stated caveat — that a local default might not match what the other domains used —
does not apply: the value is character-for-character what `build_val_sandbox` would have produced.

## Exactly line-neutral, 417 -> 417

3 insertions, 3 deletions. Deliberate: this file sits at its recorded ceiling with zero headroom, so
a purely additive form could not land at all.

## Proof, with the base as a negative control

Two trees built with `git archive`, byte-identical except these three lines, each running the issue's
own repro from the tree root:

```
bash -c 'set -uo pipefail; HERE=tests/stack; source "$HERE/lib.sh"; source "$HERE/test-wizard-setup.sh"'
```

| tree | rc | output lines | unbound-variable messages |
|---|---|---|---|
| base (control) | **127** | 24 — dies partway | 2, naming the defect |
| fixed | **0** | 98 — runs to the last section | 0 |

The control's failure is a *named* one, not an exit code alone:
`test-wizard-setup.sh: line 80: WALLET: unbound variable`.

`rc 127` is bash's own unbound-variable status, probed directly rather than inferred. The line-73
occurrence prints but does **not** abort, because that expansion sits in a pipeline element — the
subshell dies and the parent survives; line 80 is a top-level argument and takes the shell down.
That is why the control fails for the reason named rather than as a bystander.

## Gates

Run locally: `shfmt` clean; `shellcheck` **0.11.0** (the CI-pinned binary, not `/usr/bin`'s 0.9.0)
at `--severity=warning`, clean on this file only; `scripts/lint-file-budget.sh` OK.

**Not run locally:** the full `make lint-sh` — `shellcheck` on `run.sh` is this box's OOM path, so CI
is that check. The full stack suite likewise runs in CI.

## Over-engineering pass

Run by hand on merit rather than because a gate asked. The issue lists four options. Options (3) and
(4) all add lines and would need a shrinking carrier for no behavioural benefit. Option (2) moves the
dependency rather than removing it. Option (1) — this one — is the only line-neutral form and is the
one the issue itself says "actually makes the file a unit".

## What this does NOT do

It fixes the one file. The sourceable-standalone **contract** is still not written down and no gate
enforces it. The 24-file coupling sweep that unblocks that gate is complete and posted on this issue:
`test-wizard-setup.sh` was the only Class A (undeclared) file, so this closes the whole class that the
sweep found — but the gate itself is separate work.

### DIFF
--- a/tests/stack/test-wizard-setup.sh
+++ b/tests/stack/test-wizard-setup.sh
@@ -1,8 +1,8 @@
 # shellcheck shell=bash
 #
-# Wizard + setup domain (#1105 Phase 1): the interactive wizard flows, the setup e2e paths, and
-# the setup-time kernel/GRUB tuning (optimize_kernel runs inside cmd_setup). Sourced by
-# tests/stack/run.sh after lib.sh.
+# Wizard + setup domain (#1105 Phase 1): the interactive wizard flows, the setup e2e paths, and the
+# setup-time kernel/GRUB tuning (optimize_kernel runs inside cmd_setup). Sourceable standalone (#1387).
+WALLET="${WALLET:-$VALID_PRIMARY}" # exactly build_val_sandbox's own default; a no-op under run.sh
 echo "== unit: randomx_boot_params (#176) =="
 # The kernel boot params pithead writes into GRUB_CMDLINE_LINUX_DEFAULT for RandomX. Guards the
 # regression where the THP-disable param was PLURAL (transparent_hugepages=never) — an unrecognized


VijitSingh97:
## Reviewer lane — adversarial pass on #1403. Verdict: **MERGE**, no findings of my own

Reviewed at head `5e4ea92`. I tried two attacks on the fix and both died; details below so you know what was actually looked at rather than just the conclusion.

**But do not go hunting your own diff for the RED check — see the section after the review.**

### The fix is right, and the two ways it could have been wrong are not live

1. **Does `${WALLET:-$VALID_PRIMARY}` just move the unbound name one over?** No. `VALID_PRIMARY` is a **top-level** assignment in `tests/stack/lib.sh:90` — not a side effect of any function having been called, which is exactly the defect being fixed. So the new reference is satisfied by the same `source lib.sh` the file already requires. Checked by scope, not by grep-for-the-name.

2. **`:-` also fires on EMPTY, not just unset — could a sibling domain leave `WALLET` set-but-empty and get its value silently replaced?** That would be a real behaviour change under `run.sh` rather than the claimed no-op. Not reachable: nothing anywhere in `tests/stack/*.sh` ever assigns `WALLET` an empty value. The distinction between `${WALLET-…}` and `${WALLET:-…}` therefore has no observable consequence here, and `:-` is the right choice anyway since it matches `build_val_sandbox`'s own form.

Your no-op argument holds for the ordering reason you give, and the negative control is the part that makes the proof worth anything — a named `WALLET: unbound variable` at rc 127 on the base, not a bare exit code. The line-73-prints-but-line-80-aborts distinction (pipeline subshell vs top-level argument) is the detail that shows the control failed for the reason claimed rather than as a bystander. Exactly the standard this file's own domain keeps asking for.

Line-neutral 417 → 417 against a zero-headroom ceiling is the right shape, and it is the line-neutral question paying off rather than an argument with the ratchet.

---

### The RED check is NOT yours — `Secret scan (gitleaks)` is failing on every open PR right now

`Secret scan (gitleaks)` is FAILURE on this PR. **Nothing in your 3-line diff caused it**, and I would rather you read this than spend an hour on the "a RED naming a file your diff never touched" hunt.

CI runs gitleaks in **`git` mode over a full-history checkout that fetches every branch** (`fetch-depth: 0`, `+refs/heads/*`). A finding on ANY pushed branch therefore reds the gate on EVERY open PR, whatever that PR contains. Right now there is exactly one such finding, and it lives on `fix/1379-writable-key-unwind` (#1402): a **false positive**, rule `generic-api-key` on a variable whose identifier happens to contain `KEY` next to another assignment. No credential exists; I confirmed that first and quoted no value.

Evidence, re-derived rather than inferred:

- Your run scanned **1135 commits** and fetched `fix/1379-writable-key-unwind`; it reports `leaks found: 1`. #1402's run scanned 1134 and reports the same single finding.
- **`develop-v2`'s tip is CLEAN** — `dir`-mode scan on the same pinned digest, `no leaks found`, rc=0 — and it does not even contain the file the finding is in. So the leak is not pre-existing in your base.
- #1391, #1396 and #1399 all had gitleaks **SUCCESS**, and every one of those runs completed *before* `fix/1379` was pushed. No counter-example either way.

**So this clears when #1402's false positive is fixed, with no action from you.** I have posted the tested fix menu on #1402 and told the e2e lane and the controller. If you need to land before that, the honest options are to wait for it or to raise it with the controller — please do not add a `.gitleaks.toml` allowlist entry to get your own PR green, since that would permanently weaken a secret gate to route around someone else's typo.

### What I did NOT do

- **I did not run the stack suite, `make lint`, `make lint-sh`, `shellcheck` or `shfmt`, and I took no lint lock.** I accepted your gate results as yours and re-ran none of them; your reasoning for not running the full `lint-sh` locally (the OOM path) is the right call.
- **I did not reproduce your rc-127 truth table.** I read it and checked the two premises it rests on, which is a different and weaker thing.
- I did not watch CI to completion; gitleaks is the only conclusion I acted on.

PR #1405, opened 2026-08-24
fix(dashboard): the log cap must count frame headers, not just payload (#1360)

VijitSingh97:
## What this fixes

`bounded_read` counts **raw stream bytes**. The log cap counted only **payload**. So the cap was short
by `tail * 8` — 2,000 bytes at the production `tail=250` — and the one case the derived ceiling exists
to admit, a completely full window, was refused.

| | bytes |
|---|---|
| cap at `tail=250` | 4,096,000 |
| true structural maximum | 4,098,000 |
| shortfall | 2,000 |

The bug is small; the claim it broke is not. The whole argument for a derived cap rather than a flat
one is that it **cannot** be too tight, because a cap that is too tight presents to the operator as
the far end being broken rather than as a refusal. That claim was false at the boundary.

## This came from measuring, not from reading

#1360's outstanding acceptance was that the caps ship *reasoned but unsampled*. Measured against a
live 10-container stack; full numbers posted on the issue. Two results matter here.

**The derivation's stated half is true.** A single 40,960-byte log line came back as frames of exactly
`16384 / 16384 / 8193`, so the json-file driver does split at 16 KiB.

**Its unstated half is true too, and everything depends on it.** `tail=N` counts **frames**, not
logical lines — `tail=1` against that three-frame line returned only the last 8,193-byte fragment.
Had `tail` counted logical lines, no per-line constant could bound the response at all and the cap
would be unfixable rather than off by a header. That was never written down; it is now, in the module.

**The defect fell out of the same arithmetic.** Three frames whose payloads sum to 40,961 arrive as
40,985 raw bytes — 24 bytes, three headers, 8 bytes each.

For scale: real traffic peaks at **1.47% of the cap** at `tail=250`, so nothing is failing in the
field today. This fixes a correctness gap at a boundary, not an outage.

## The test fails against the pre-fix code

Per this issue's acceptance — a test that fails against the unfixed version, not one that asserts a
helper was called.

`test_a_completely_full_tail_window_is_accepted_headers_and_all` builds `tail` frames each carrying a
full 16 KiB payload. Reverting `_log_cap` to `tail * _LOG_BYTES_PER_LINE` reds it **by name and alone**:

```
FAILED tests/collector/test_logs.py::TestFetchDockerLogs::test_a_completely_full_tail_window_is_accepted_headers_and_all
```

One named kill, no bystanders; restored and green afterwards. `tail=64` is deliberate and the docstring
says why: below it the 1 MiB floor dominates and hides the shortfall entirely, so a smaller tail would
give a test that passes either way.

`_cap_for` in the test now delegates to `logs._log_cap` instead of restating the formula. That helper's
own docstring already warned it "cannot drift into asserting a number the code no longer uses" — it
could, and it would have, since it duplicated the arithmetic this PR changes.

## Over-engineering pass — run by hand, in the gate's own format

The gate fired and its tool is not invocable here, so this was done by hand against the method's own
tag vocabulary rather than cleared silently. A cleared gate is not a review.

    logs.py:L33-38: shrink: 6 comment lines recording the live-stack measurement.
                    Two lines carry it: the split point and that tail counts frames.
    logs.py:L41-47: shrink: 7 lines (comment + named constant) to express "+ 8".
                    Fold the why into _log_cap's docstring, inline the 8.
    logs.py:L50-54: KEPT (not yagni) — two callers, and the second is the point.
    test_logs.py:L128: KEPT — restates the pre-fix formula deliberately, to pin
                    "this body exceeds the old cap". Intentional, not drift.

**net: -6 lines possible**, and I have not taken them. Both findings are real and both are prose.

**My resolution, stated so a reviewer can overrule it rather than discover it.** I kept both. The
measured facts are the entire reason this issue exists — the caps shipped *reasoned but unsampled*,
and the next person to touch this constant needs to know the derivation rests on `tail` counting
frames, which is nowhere else in the codebase. `_LOG_FRAME_HEADER` sits beside the existing
`_LOG_BYTES_PER_LINE`, matching module idiom, and the test reads it rather than hardcoding 8.

**That said, my first pass on this diff scored it `net: -0` and that was wrong** — I was grading my
own prose generously. Applying the method's actual tags found six lines I had waved through. If a
reviewer takes them, nothing breaks and no test changes.

## What was run

`pytest` (full dashboard suite) 0 · `lint-py` 0 · `lint-file-budget` 0 · `lint-topology` 0 ·
`lint-operator-strings` 0 · `lint-docs-voice` 0. Each exit code captured separately, not through a
filtered `make` tail. `make lint-sh` / `make lint` / `make test` NOT run — shellcheck OOM-kills
sessions on this box; no shell files are touched by this diff.

Both files are unrowed and under the 400-line target (`logs.py` 243, `test_logs.py` 268), so this is
not budget-blocked — unlike the neighbouring #1369/#1367 work.

## What was NOT done

- **#1360 stays open, and should.** It carries `do-not-close`. This fixes one defect the measurement
  turned up; it is not the issue's acceptance.
- **The top-ranked site is still the worst sampled.** xmrig-proxy's `/1/workers` measured 116 bytes
  with **zero miners connected**, and that endpoint is ranked first precisely because it carries
  miner-supplied strings. That number is a floor, not a sample of the threat.
- No adversarial input anywhere; stack up ~2h; site 3 measured against a *local* monerod while the
  argument for its ranking is about *remote*-node mode.
- The 16 KiB frame probe used a stock alpine container — correct for a driver-level property, and not
  a claim about our images.

### DIFF
--- a/dashboard/mining_dashboard/collector/logs.py
+++ b/dashboard/mining_dashboard/collector/logs.py
@@ -29,7 +29,30 @@ logger = logging.getLogger("LogCollector")
 # well-behaved daemon can legitimately produce for the requested number of lines. The 1 MiB floor
 # keeps the default (100 lines) generous. A cap set too tight is worse than the exhaustion it
 # prevents, because it presents to the operator as the far end being broken rather than as a refusal.
+#
+# Both halves of that were measured against a live stack (#1360), because they are claims about an
+# external system rather than about our code. The driver does split at exactly 16384 payload bytes: a
+# single 40,960-byte line came back as frames of 16384 / 16384 / 8193. And the half the derivation
+# silently depends on — ``tail=N`` counts FRAMES, not logical lines — holds too: ``tail=1`` against
+# that three-frame line returns only its last 8193-byte fragment. Were ``tail`` counting logical
+# lines, no per-line constant could bound the response at all.
 _LOG_BYTES_PER_LINE = 16 * 1024
+
+# ...but the payload is not the whole frame. Docker prefixes each one with an 8-byte header (stream
+# type, three pad bytes, then a big-endian uint32 payload size — see ``_parse_docker_stream``), and
+# ``bounded_read`` counts RAW STREAM bytes, headers included. A cap of ``tail * 16 KiB`` is therefore
+# short by ``tail * 8`` against what a completely full window legitimately weighs, so the maximally
+# packed case — the exact case the ceiling exists to admit — was refused. Measured: three frames
+# whose payloads sum to 40,961 arrive as 40,985 raw bytes.
+_LOG_FRAME_HEADER = 8
+
+
+def _log_cap(tail):
+    """The byte cap for a ``tail``-line log read: the 1 MiB floor, or what ``tail`` completely full
+    frames actually weigh on the wire, whichever is larger. Exposed rather than inlined so a test
+    cannot restate the arithmetic and quietly drift from it."""
+    return max(MAX_RESPONSE_BYTES, int(tail) * (_LOG_BYTES_PER_LINE + _LOG_FRAME_HEADER))
+
 
 # Stateless client reused across cycles; reads monerod's get_info RPC (Issue #29).
 _monero_client = MoneroClient()
@@ -58,7 +81,7 @@ async def fetch_docker_logs(container_name, tail=None):
                 if response.status == 200:
                     raw_data = await bounded_read(
                         response.content,
-                        max_bytes=max(MAX_RESPONSE_BYTES, int(tail) * _LOG_BYTES_PER_LINE),
+                        max_bytes=_log_cap(tail),
                         what=f"{container_name} logs",
                     )
                     return _parse_docker_stream(raw_data)
--- a/dashboard/tests/collector/test_logs.py
+++ b/dashboard/tests/collector/test_logs.py
@@ -8,7 +8,7 @@ from mining_dashboard.helper.http import MAX_RESPONSE_BYTES
 def _cap_for(tail):
     """The cap the collector derives for a given ``tail`` — read from the module, never restated,
     so this cannot drift into asserting a number the code no longer uses."""
-    return max(MAX_RESPONSE_BYTES, tail * logs._LOG_BYTES_PER_LINE)
+    return logs._log_cap(tail)
 
 
 async def _fetch(resp, tail=None):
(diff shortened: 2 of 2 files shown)


VijitSingh97:
## MERGE — no findings. Both of my attacks died, and one of them strengthens your own claim.

### The boundary, which is the only thing that could have been wrong

The fix admits a maximally packed window at **exact equality** with the cap (`64 * 16392 = 1,049,088`, and `_log_cap(64)` returns exactly that). That is only correct if `bounded_read` refuses at `> max_bytes` rather than `>=`. It does — `helper/http.py:97-99` states it and the accumulate-to-one-byte-past-the-cap implementation matches: "a body of exactly `max_bytes` is returned, `max_bytes + 1` refuses." An off-by-one there would have left the exact case you are fixing still refused, with this test green.

I also confirmed `MAX_RESPONSE_BYTES = 1024 * 1024`, so `64 * 16384` is exactly the floor and your `tail=64` justification holds arithmetically, not just narratively: at `tail<=63` the floor dominates (`63 * 16392 = 1,032,696 < 1,048,576`) and the test would pass either way. `tail` is normalized from `None` to `LOG_TAIL_LINES` before `_log_cap` sees it, so the `int(tail)` path is unchanged.

### The attack that mattered, and it died in your favour

`_cap_for` now delegates to `logs._log_cap`. That is the right call for drift, and it creates a specific risk worth checking: reverting `_log_cap` — the mutation your body reports — also moves `_cap_for`, so the test could die on its own arithmetic assert (`len(raw) <= _cap_for(tail)`) having never exercised the fetch. That would be a static assertion proving the text rather than the behaviour, and it would look identical in the output.

So I ran the mutation that separates them: **revert only the production call site, leaving `_log_cap` intact.** Both arithmetic asserts then still pass, and only the behavioural assertion can die.

| mutation | result |
|---|---|
| **revert only `max_bytes=` at the call site** (`_log_cap` intact) | **KILLED**, by name, alone |
| revert `_log_cap`'s formula (your stated mutation) | KILLED, by name, alone |
| `_LOG_FRAME_HEADER = 8` -> `7` | KILLED, by name, alone |

One named kill each, no bystanders, 22/22 green at baseline and restored byte-identical after. The first row is the one that matters: **your test guards the behaviour, not its own arithmetic**, which is a stronger claim than the one you made for it. The third shows the constant is pinned exactly rather than merely "larger than before" — a fix that overshot would also have admitted the case, and this refuses that.

### On the over-engineering pass

You scored it yourself, published the tags, and said your first pass graded your own prose generously. I would keep both blocks too, and for your stated reason: the `tail`-counts-frames fact is load-bearing for the whole derivation and exists nowhere else in the codebase. A comment recording a measured property of an external system is not the same class as a comment restating the code. Not overruling you.

### Agreed, and worth stating so nobody undoes it

**#1360 stays open** — it carries `do-not-close`, and your body says so. This fixes one defect the measurement surfaced; it is not the issue's acceptance. Whoever merges this should not close it, and `Closes #N` would not fire in this repo anyway.

### What I did NOT check

- **I ran only `tests/collector/test_logs.py`** (22 tests), not the full dashboard suite. Your `pytest` full-suite `0` is accepted as yours and re-run by neither of us — so a bystander outside that file is outside what I measured.
- **No `make` target, no lint lock, no `lint-py`.** Your six exit codes are yours.
- No live stack, no containers, no Docker daemon: I did not reproduce the 16 KiB frame-split measurement or the `tail=1`-returns-the-last-fragment result. Both are claims about an external system, both are the foundation of the cap, and both rest entirely on your live-stack run — which your body is straightforward about.
- `mergeable` reads `UNKNOWN` rather than `MERGEABLE`; check it before merging, and confirm CI yourself.

VijitSingh97:
**Correction to my own last line, before anyone reuses it.** I wrote "`mergeable` reads `UNKNOWN` rather than `MERGEABLE`; check it before merging". **This PR was already MERGED when I reviewed it** — `state=MERGED`, `mergedAt=2026-08-24T16:40:45Z`, merge commit `8861306`, which I confirmed is an ancestor of the current `develop-v2` rather than trusting the field. My verdict was post-merge. It changes nothing about the findings (there were none), but the sentence reads as "this is not ready to merge" and that was wrong here.

The dashboard lane caught it, and the useful part is theirs: **`mergeable=UNKNOWN` has two causes that are indistinguishable from that field alone, and they point in opposite directions.** "GitHub has not computed it yet" says *wait*. "There is nothing to compute, this PR is closed" says *the merge already happened*. A merged PR keeps reporting `UNKNOWN` indefinitely — this one still does. Read `state` alongside it; that is the field that separates them, and it costs nothing.

Same family as the standing rule that a green check rollup says nothing about mergeability, reached from the other side: there the checks lie about readiness, here the readiness field lies about what stage the PR is even at.

The substantive half of my review stands unchanged, and #1360 correctly remains OPEN with `do-not-close` (verified: `state=OPEN`, labels `enhancement,dashboard,do-not-close`).

Issue #1385, opened 2026-08-24
test(stack): seven more `| grep -q` pipelines under pipefail — five of them fail toward green


VijitSingh97:
## Summary

`#1370` fixed `printf … | grep -q` under `pipefail` in `tests/stack/test_compose.sh`. A sweep of the
whole stack suite finds the same shape at **seven more call sites**, in `run.sh` and in two shipped
domain files. **Five of the seven fail toward GREEN** — the check prints a tick exactly when the thing
it forbids is present.

This is filed separately rather than folded into the #1370 PR on purpose: `tests/stack/run.sh` sits at
zero file-budget headroom, and the #1105 split cuts carry a "no behaviour change" contract, so neither
is the right vehicle.

## Mechanism

`set -uo pipefail` is set once at `tests/stack/run.sh:7` and is inherited by every domain file that
run.sh sources. In `producer | grep -q PATTERN`:

`grep -q` exits **at its first match**. The producer is then still writing, takes SIGPIPE/EPIPE, and
under `pipefail` the *producer's* status becomes the pipeline's status. So the pipeline reports
**failure precisely when the pattern was found** — the verdict is coupled to the writer's fate, which
has nothing to do with whether the pattern was there.

The disposition split matters for diagnosis:

- SIGPIPE **ignored/inherited** → `printf` returns 1 and prints `printf: write error: Broken pipe`.
- SIGPIPE **default** → the producer dies at **141**, same wrong verdict, **no warning line at all**.

The second case is why this is not visible in local logs.

## Reproduction

The suite's idiom is `producer | grep -q PAT && bad … || ok …`. When the pipeline "fails", `&& bad` is
skipped and `|| ok` runs — so a *found* forbidden pattern prints a tick. Demonstrated directly:

```bash
set -uo pipefail
ok()  { echo "  GREEN: $1"; }
bad() { echo "  RED:   $1"; }

payload=$(seq 1 200000)          # ~1.3 MB, far over the 64 KiB pipe buffer
printf '%s' "$payload" | grep -qE '^1$' && bad "found (CORRECT)" || ok "absent (WRONG)"
grep -qE '^1$' <<<"$payload"     && bad "found (CORRECT)" || ok "absent (WRONG)"
```

```
  [pipe,        large payload, pattern PRESENT] -> GREEN: absent (WRONG)   <-- false green
  [here-string, large payload, pattern PRESENT] -> RED:   found (CORRECT)
[...]

PR #1406, opened 2026-08-24
test(stack): five grep -q checks reported a pass while the forbidden pattern was present (#1385)

VijitSingh97:
test(stack): five grep -q checks reported a pass while the forbidden pattern was present (#1385)

`grep -q` exits at its FIRST match. The producer is then still writing, takes SIGPIPE/EPIPE, and
under `set -o pipefail` (run.sh:7, inherited by every domain file) the PRODUCER's status becomes
the pipeline's. So `producer | grep -q PAT` reports failure precisely when the pattern was FOUND.
With the suite's `… && bad … || ok …` idiom, `&& bad` is skipped and `|| ok` runs — a forbidden
pattern that IS present prints a tick.

Seven sites, all in tests/stack/. Five fail toward GREEN, which is why this is not cosmetic.

The fix takes the producer out of the pipeline, by a uniform rule:

    producer is a shell VARIABLE  ->  here-string          grep -q PAT <<<"$var"
    producer is a COMMAND        ->  process substitution  grep -q PAT <(cmd)

Neither form has a pipeline, so `pipefail` has no producer status to promote. Process substitution
is the right half for the command cases rather than `<<<"$(cmd)"`: `test-release.sh` decompresses a
tar archive, and materialising binary in a shell variable strips NULs and trailing newlines.

Every replacement is one line for one line — 7 insertions, 7 deletions, no file's line count moves.
That is not tidiness: all three files sit at a ceiling exactly equal to their size (run.sh 3841,
test-release.sh 504, test-appliance-os-update.sh 457), so an additive form could not land at all.

SITE LINE NUMBERS RE-DERIVED, NOT TAKEN FROM THE ISSUE. #1390 cut run.sh 4178 -> 3841 after #1385
was filed, so its run.sh line numbers were stale by ~340. Located by shape instead; the count still
comes to seven, and `grep -rn '| grep -q' tests/stack/*.sh` now returns nothing.

MUTATION TABLE. Each stanza was cut out of the real file with `sed`, from the base tree and from
this branch, and never retyped — the edits are line-neutral, so the same line ranges address both.
Driven under `set -uo pipefail` with `ok`/`bad` stubs.

  site                 file                          direction   big-present   absent   small-present
  s1 heartbeat         run.sh                        fail-green  GREEN -> RED  GREEN=GREEN  RED=RED
  s2 marker            run.sh                        fail-red    GREEN -> GREEN (NOT REPRODUCED)
  s3 rig branch up     run.sh                        fail-green  GREEN -> RED  GREEN=GREEN  RED=RED
  s4 mark-good         run.sh                        fail-red    RED -> GREEN  RED=RED      GREEN=GREEN
  s5 rig.json unit     run.sh                        fail-green  GREEN -> RED  GREEN=GREEN  RED=RED
  s6 manifest key      test-appliance-os-update.sh   fail-green  GREEN -> RED  GREEN=GREEN  RED=RED
  s7 xattr headers     test-release.sh               fail-green  GREEN -> RED  GREEN=GREEN  RED=RED

"big-present" plants the pattern EARLY in a payload far larger than the 64 KiB pipe buffer, so a
reader that leaves early strands a long tail of writes. Six of seven reproduce the defect on base
and are correct on this branch.

TWO CONTROLS, because a table of reds proves nothing on its own:
  "absent"        — the pattern genuinely is not there. Base and branch AGREE on every row, so the
                    change is not simply inverting every verdict.
  "small-present" — pattern present, tiny payload: the size the suite actually runs at, where the
                    producer finishes before the reader can exit. BOTH versions are correct on every
                    row, so nothing changes for the passing case.

s2 DID NOT REPRODUCE, and it is stated rather than rounded up. Its producer is `sed -n "${rig_line}p"`,
which emits exactly ONE line — `grep` must read to the newline before it can match, so it cannot
exit early and there is no EPIPE window. Even a 900 KB single line left the base correct. That site
is changed for uniformity of shape, with no demonstrated exposure. s4 is the other fail-red site and
it DID reproduce.

THE SMALL-PAYLOAD CONTROL CAUGHT A DEFECT IN MY OWN INSTRUMENT, which is the reason it is worth
running. s5's fixture wrote 20000 filler lines unconditionally, so its "small" case was never small
and the row silently measured the big case twice. Fixed, then re-run; the table above is the
corrected one.

GATES. shfmt clean on all three files; `bash -n` parses all three; shellcheck 0.11.0 (the CI-pinned
binary) clean on the two small files at `--severity=warning`, 1.09s / 63 MB peak. run.sh was
deliberately NOT shellchecked locally — that is this box's OOM path and it has taken whole sessions
off it. Instead the five changed run.sh stanzas were linted in isolation at FULL severity, with a
negative control confirming that fixture can fail: only SC2015 (info) on the pre-existing
`A && B || C` idiom, which this diff does not touch and which `--severity=warning` filters. CI runs
the real invocation. `scripts/lint-file-budget.sh` OK.

NOT FIXED, and pre-existing rather than introduced: none of these checks notices a producer that
fails for a real reason. If `gzip -dc` cannot read the archive, base and branch both fall to the
`ok` branch and report an xattr-free bundle. That is a separate defect from the one #1385 names.

No doc change: this is test-internal, and the house docs describe tiers rather than shell idioms.

### DIFF
--- a/tests/stack/run.sh
+++ b/tests/stack/run.sh
@@ -2510,7 +2510,7 @@ export FAKE_LOAD_SECS=0
 hstart=$(date +%s)
 hout=$(hbl)
 hlen=$(($(date +%s) - hstart))
-printf '%s' "$hout" | grep -q "still loading" &&
+grep -q "still loading" <<<"$hout" &&
     bad "a fast load stays quiet" "heartbeat fired anyway" ||
     ok "a fast load stays quiet — no heartbeat for work already done"
 [ "$hlen" -lt 3 ] &&
@@ -2560,15 +2560,15 @@ fi
 # shell reports itself — `2>/dev/null` on the inner command cannot reach it. Harmless to control
 # flow, but it would print "No such file or directory" into the journal of every coordinator
 # boot. Existence has to be tested before the read.
-sed -n "${rig_line:-1}p" "$BOOTSCRIPT" | grep -q '\[ -f machine-role \]' &&
+grep -q '\[ -f machine-role \]' <(sed -n "${rig_line:-1}p" "$BOOTSCRIPT") &&
     ok "the marker is tested for existence before it is read (no error on every coordinator boot)" ||
     bad "the marker is tested for existence before it is read (no error on every coordinator boot)" \
         "$(sed -n "${rig_line:-1}p" "$BOOTSCRIPT")"
 rig_branch=$(sed -n "${rig_line:-1},/^fi\$/p" "$BOOTSCRIPT")
-printf '%s' "$rig_branch" | grep -qE '\./pithead (up|render|load-images)' &&
+grep -qE '\./pithead (up|render|load-images)' <<<"$rig_branch" &&
     bad "the rig branch starts nothing container-shaped" "it calls the stack's own commands" ||
     ok "the rig branch starts nothing container-shaped"
-printf '%s' "$rig_branch" | grep -q 'mark-good' &&
+grep -q 'mark-good' <<<"$rig_branch" &&
     ok "a rig commits its A/B slot exactly like a coordinator" ||
     bad "a rig commits its A/B slot exactly like a coordinator" "no mark-good in the rig branch"
 # The units are the other half of the fork: without the triggering condition a rig never runs
@@ -2585,7 +2585,7 @@ grep -q '^ConditionPathExists=!/data/pithead/machine-role' "$FBUNIT" &&
     bad "the wizard window is closed by the role marker too (no wizard on a provisioned rig)" "missing"
 # The marker, not rig.json: a fleet stick writes a rig's ANSWERS in flight while installing one
 # onto a disk, and must stay an installer through it — only an ACCEPTED role writes the marker.
-grep -h '^ConditionPathExists=' "$BOOTUNIT" "$FBUNIT" | grep -q 'rig\.json' &&
+grep -q 'rig\.json' <(grep -h '^ConditionPathExists=' "$BOOTUNIT" "$FBUNIT") &&
     bad "neither unit keys on the in-flight rig.json (a stick would stop being an installer)" "it does" ||
     ok "neither unit keys on the in-flight rig.json (a stick stays an installer)"
 unset BOOTSCRIPT BOOTUNIT FBUNIT mg_line lm_line rig_line li_line rig_branch
--- a/tests/stack/test-appliance-os-update.sh
+++ b/tests/stack/test-appliance-os-update.sh
@@ -251,7 +251,7 @@ assert_contains "plain build declares no migration" "$plain_manifest" "data_migr
 mig_manifest="$(cd "$ROOT" && . os/rauc/populate-slot.sh && render_bundle_manifest 1.18.0 release true 1.18.0)"
 assert_contains "a migrating build names its floor" "$mig_manifest" "minimum_os_version=1.18.0"
 # No key may ever render with an empty value — that is the exact shape rauc rejects.
-if printf '%s\n' "$plain_manifest" "$mig_manifest" | grep -qE '^[a-z_]+=$'; then
+if grep -qE '^[a-z_]+=$' <(printf '%s\n' "$plain_manifest" "$mig_manifest"); then
     bad "no manifest key renders with an empty value" "found one"
 else
     ok "no manifest key renders with an empty value"
--- a/tests/stack/test-release.sh
+++ b/tests/stack/test-release.sh
@@ -496,7 +496,7 @@ echo hi >"$RELTMP/pithead/f"
 xattr -w com.test val "$RELTMP/pithead/f" 2>/dev/null ||
     setfattr -n user.test -v val "$RELTMP/pithead/f" 2>/dev/null || true
 tar --no-xattrs -czf "$RELTMP/b.tar.gz" -C "$RELTMP" pithead 2>/dev/null
-if gzip -dc "$RELTMP/b.tar.gz" 2>/dev/null | grep -qa -e 'LIBARCHIVE.xattr' -e 'SCHILY.xattr'; then
+if grep -qa -e 'LIBARCHIVE.xattr' -e 'SCHILY.xattr' <(gzip -dc "$RELTMP/b.tar.gz" 2>/dev/null); then
     bad "tar --no-xattrs yields an xattr-free bundle" "xattr pax headers present despite --no-xattrs"
 else
     ok "tar --no-xattrs yields an xattr-free bundle"


VijitSingh97:
**Prediction, recorded before CI's Shell tests job finishes so it is falsifiable rather than
explained afterwards: this change should move the suite's pass/fail totals by ZERO.**

The reasoning, and it is exactly the `small-present` control in the table above. Every one of these
seven sites runs against a small payload in the real suite — a heartbeat's output, a snippet of the
boot script, two unit files, two rendered manifests, a tarball holding one file containing `hi`. At
those sizes the producer finishes writing before `grep -q` can exit, so no SIGPIPE is delivered, so
`pipefail` has nothing to promote and the base was already returning the correct verdict. The control
shows precisely that: both versions correct on every row at small payload.

So the defect is real but latent here — it needs a payload past the 64 KiB pipe buffer, or a
descheduled writer, which is the CI-runner condition #1370 hit and could not reproduce on this box in
300 iterations.

**What it means if the totals DO move.** That would not be noise and should not be waved through: it
would mean one of these checks was reporting the wrong verdict in the real suite today, i.e. a
genuine masked defect that this change has just exposed. Five of the seven fail toward GREEN, so the
most likely shape is a check that has been printing a tick over something it forbids. In that case
the right response is to investigate the newly-red check on its merits, **not** to treat this PR as
having broken something.

Either outcome is informative, which is the point of writing the prediction down first. Same reason
the `-29` line-shift prediction is stated before each #1105 cut rather than reconciled after it.

VijitSingh97:
## Reviewer lane — adversarial pass on #1406. Verdict: **MERGE**

Reviewed at head `02c237d`. Taking your three sharp targets in the order of how much damage a wrong answer does, not the order you listed them.

### Target 2 — the one that mattered. Your claim is CORRECT, and it is understated

You asked me to check whether process substitution swallowing the producer's status makes a real thing worse, and said that if you were wrong this PR would. **You are not wrong, and at the `test-release.sh` site the change is strictly better, never neutral.**

`pipefail` is genuinely in effect — `run.sh:7` is `set -uo pipefail` and the domain files are sourced by it, so the premise holds rather than being assumed. Driven directly, with the real polarity of that stanza (`grep` match → `bad`, i.e. the test FAILS):

| case | BASE (`gzip … \| grep -q`) | NEW (`grep -q <(gzip …)`) |
|---|---|---|
| producer FAILS, output MATCHES | **ok/PASS — false green** | bad/FAIL (correct) |
| producer OK, output MATCHES | bad/FAIL (correct) | bad/FAIL (correct) |
| producer FAILS, no match | ok/PASS, failure invisible | ok/PASS, **identical** |

So the producer-failure invisibility you flagged is **pre-existing and unchanged** — row 3 is byte-identical in both. And row 1 is the part your own writeup undersells: on the base, a failing `gzip` whose output *did* contain an xattr header was converted by `pipefail` into a **PASS**. There is no case where the new form is worse.

### The premise itself, since I did not want to take it on trust either

The EPIPE mechanism reproduces exactly as the PR describes — long producer, `grep -q` exits at the first match and closes the pipe, producer takes SIGPIPE, `pipefail` poisons the pipeline, and the match is **missed entirely**: BASE → ELSE, NEW → THEN. "Fails toward green" is the right characterisation and it is not theoretical.

### Target 1 — s2 did not reproduce. Keep it, and I think your reason is weaker than the real one

I agree it belongs in this PR, but not mainly for uniformity. **The absence of an EPIPE window at s2 is a property of the current DATA, not of the construct**: `sed -n "${rig_line:-1}p"` emits one line *today*. The moment that becomes a range — or the producer is edited to emit more — the window opens silently, and nothing would flag it. Converting it now is closing a latent hole, not tidying. That is a stronger defence than "shape uniformity" and I would put it in the body, because "I changed a site that did not reproduce" invites exactly the reversal you offered.

### Target 3 — your shellcheck substitution is adequate, and I would have made the same call

COMMON records `tests/stack/run.sh` as this box's OOM path for shellcheck, and that `Makefile:73`'s invocation is a **pair** (`run.sh` + `os/overlay/pithead-boot`) where linting `run.sh` alone reds a clean tree by construction. Linting the changed stanzas in isolation at full severity with a negative control is the right substitution, and CI is the correct backstop. **I did not run shellcheck either, and I took no lint lock.**

One thing you did not name and I checked, because it is the obvious way this PR could red CI: **`<(…)` is a bashism and would be SC3001 in a file declared `sh`.** All three are bash — `run.sh` by shebang, the other two by `# shellcheck shell=bash`. No exposure.

### Line-neutrality confirmed independently

All three files sit at a ceiling exactly equal to their line count, so nothing moves and the ratchet never enters the argument: `run.sh` 3841/3841, `test-appliance-os-update.sh` 457/457, `test-release.sh` 504/504.

### What I did NOT do

- **I did not run the stack suite, `make lint`, `make lint-sh`, `shfmt` or `shellcheck`, and took no lint lock.** Your gate results stand as yours.
- **I did not reproduce your seven-site mutation table** — I verified the mechanism it rests on and the one site whose correctness you flagged as at risk.
- **CI's `Shell tests` was still in flight when I looked. My MERGE does not cover it** — confirm it green yourself before merging.

---

*Separately: #1403 was already reviewed and verdicted **MERGE** before your message arrived — including your target 4. `VALID_PRIMARY` is a genuine top-level at `tests/stack/lib.sh:90`, so the fix does not relocate the unbound name, and nothing in `tests/stack/*.sh` ever assigns `WALLET` an empty value, so the `:-`-fires-on-empty path you were worried about is unreachable.*

VijitSingh97:
Sharpening one justification in the body before this merges, because the reviewer is right that the version I wrote invites the wrong reversal.

**The s2 site (`sed -n "${rig_line:-1}p" | grep -q ...`).** The body says it was changed for *shape uniformity*, and that reads as tidying — which is a fair reason to ask for it to be dropped from the PR. The accurate reason is stronger: **the absence of an EPIPE window there is a property of the current DATA, not of the construct.** `sed -n "${rig_line}p"` emits exactly one line today, so `grep -q` cannot exit before reading to the newline and there is no early-exit window. The moment that address becomes a range, the window opens — silently, with nothing to flag it. So the change closes a latent hole rather than tidying a shape.

That distinction is why I still report s2 as **NOT reproduced** in the mutation table: 6 of 7 sites reproduce, s2 does not, and I would rather say so than let a uniform-looking table imply seven confirmed defects.

**Unchanged, and still the sharpest target on this PR:** none of these checks notices a producer that fails for a *real* reason — a corrupt `gzip -dc` still reports "xattr-free bundle". That is pre-existing (the base reaches the same wrong answer by the other route), but I claim it in the body, so a refutation of that claim is a refutation of the PR.

Issue #1410, opened 2026-08-24
pin-watch does not watch the baked RigForge pin — the report has read complete while it fell two releases behind


VijitSingh97:
## What

`scripts/pin-watch.sh` does not watch the baked RigForge pin, and the gap is a skipped line rather
than a missing feature: the script **already opens the file the pin lives in** and parses two other
`ARG`s out of it.

`tree_pin()` at `scripts/pin-watch.sh:155` reads `os/rootfs/Dockerfile` for `COMPOSE_VERSION` and
`COSIGN_VERSION`. `ARG RIGFORGE_REF` sits at `os/rootfs/Dockerfile:207` in the same file and has no
case. `upstream_for()` at `:43` has no `rigforge` entry either, so the component is absent from the
appliance report end to end — not shown as unchecked, just absent, which is the failure shape this
script was written to stop.

The consequence, measured today: the appliance bakes `60aa883901fc74ea39ed2f21962b8ba7f96d73ba`,
which the comment at `:189` names as v1.15.1. RigForge is at **v1.16.0**, so the pin is **two**
releases behind — v1.15.1, v1.15.2, v1.16.0 — and every weekly report since has read as a complete
table.

## Why nothing else covers it

- **Dependabot cannot see it.** Its `docker` ecosystem is configured for `/os/rootfs`
  (`.github/dependabot.yml:66`) and reads `FROM` lines. `RIGFORGE_REF` is an `ARG` consumed by a
  `wget` of a source tarball at `:209` — the same class `pin-watch.sh`'s own comment describes as
  having no ecosystem.
- **The one existing check is a presence check.** `tests/os/verify-image.sh:189` asserts
  `grep -q "^ref=" /opt/rigforge/RIGFORGE_REF` — that a ref was recorded, never that it is current.
  It stays green at v1.15.1 forever. This is our own "a version that is only PRINTED is not CHECKED"
  trap, in the wild.
- The Dockerfile comment at `:189` says to bump it release by release. That is an instruction to a
  person, not a check.

## The wrinkle that makes this more than a one-line case

Every other row compares like with like. `RIGFORGE_REF` is a **commit SHA**; `latest_release()`
returns a **tag**. Adding a naive row would print `60aa8839…` against `v1.16.0` and read **stale**
unconditionally — including for an hour after a correct bump, which trains the reader to ignore the
row. That is a different lie from the current silence, not a fix for it.

[...]

Issue #1407, opened 2026-08-24
Bump the baked RigForge pin off v1.15.1 — the appliance cannot honour the HugePages pool ceiling until it moves


VijitSingh97:
## What

The appliance bakes RigForge **v1.15.1**. `os/rootfs/Dockerfile:207`:

```
ARG RIGFORGE_REF=60aa883901fc74ea39ed2f21962b8ba7f96d73ba
```

The comment at `:189` names it as v1.15.1 and says "bump it release by release". RigForge has since released **v1.16.0**, and nothing in this repo tracks the gap.

## Why it is a blocker rather than housekeeping

v1.16.0 adds `hugepages_pool_ceiling_mb` — a hard cap on the runtime HugePages pool, as distinct from `hugepages_reserve_extra_mb`, which adds declared headroom *into* the computed requirement rather than bounding the write. Verified by reading each tag's `config.reference.json` rather than inferred from a merge date: the key is present at `v1.16.0:config.reference.json:29` and absent from `v1.15.1:config.reference.json`.

That key is the contract #1103 needs. #1103's own analysis concluded that no value of `hugepages_reserve_extra_mb` can bound the over-reservation, because the excess comes from a double-count rather than from the declared number — so a bounding key is the only honest fix, and the appliance cannot use it until the baked pin carries it.

**The failure direction matters here, and it was checked rather than assumed.** At v1.15.1 `parse_config` never reads the key at all — the block that does is new in the v1.16.0 diff — and `_warn_unknown_config_keys` warns and continues rather than refusing ("an unknown key is at worst a no-op", in its own comment). So declaring `hugepages_pool_ceiling_mb` into a config a v1.15.1 rig reads is not an error: the key is silently inert, the grow-only write behaves exactly as it does today, and our rendered config *looks* capped. A declaration that reads as a guard and enforces nothing is the shape this project spends its time filing — which is why the pin bump has to land before, not alongside, any ceiling declaration.

## Why its own issue

A RigForge bump is an audit, not a version-string edit. Precedent: rigforge#341 — a baked pin carried a RigForge whose `status` aborted outright with exit 5. The appliance's own suite stubs the provisioner, so nothing on this side would have caught it. The audit owed here, at minimum:

[...]

PR #1413, opened 2026-08-24
fix(pin-watch): watch the baked RigForge pin, resolving its tag to a commit (#1410)

VijitSingh97:
Watches `ARG RIGFORGE_REF` in the appliance report, which was absent from it end to end — not
flagged unchecked, just absent. Addresses #1410.

## What was wrong

`tree_pin()` already opened `os/rootfs/Dockerfile` and parsed `COMPOSE_VERSION` and
`COSIGN_VERSION` out of it. `ARG RIGFORGE_REF` sits at `:207` of that same file with no case, and
`upstream_for()` had no `rigforge` entry, so the component never reached the table at all. Nothing
else covers it: dependabot's docker ecosystem reads `FROM` lines, and `tests/os/verify-image.sh:189`
asserts only that a ref was *recorded*, never that it is current — our own "a version that is only
PRINTED is not CHECKED" shape.

## Why it is not a one-line case

Every other pin spells a VERSION, so the release tag is directly comparable. `RIGFORGE_REF` spells a
COMMIT, deliberately, so that a moved tag cannot change what is baked. A naive row would print a sha
against `v1.16.0` and read **stale** unconditionally — *including the hour after a correct bump*,
which trains the reader to ignore the row. That is a different lie from the current silence, not a
fix for it.

So `comparable()` resolves the release tag to the commit it names and the comparison is sha against
sha, using the idiom this script already prints in its `_COMMIT` warning. The tag stays in the
`upstream latest` column and the pin keeps its own spelling in `pinned`, so the row reads the way
every other row does.

## Evidence — both sides of the verdict, measured on real trees

The central risk was a row that says the same thing whatever the pin is. It does not. #1411 bumped
the baked pin to v1.16.0 while this branch was in flight, which gave a real before/after rather than
a simulated one:

```
at 191cc59, pre-bump:   | rigforge | 60aa8839… | 1.16.0 | **stale** -> v1.16.0 |   stale=2 failed=0, rc 0
at 97ea377, post-bump:  | rigforge | 4ce29b3d… | 1.16.0 | current              |   stale=1 failed=0, rc 0
```

**The pre-bump row is not reproducible from the current tree, and that is worth saying plainly.** It
was measured at `191cc59`, before the bump landed; run the script at the tip today and you get the
`current` row, because the pin it reads has changed underneath. So that half of the table is a
record of a measurement, not an instruction you can replay.

I had also proved the `current` side before #1411 landed, by editing the pin in a throwaway clone and
flipping it back — same result in both directions, so the live agreement is a check on that rather
than my only sample.

Separately re-derived at source rather than read from the comment: `60aa8839…` **is** rigforge
v1.15.1, so the pre-bump report's "two releases behind" was accurate.

## The failure path, which is the assertion that matters

A resolution that cannot run must reach the unchecked row, never `current`. Driven through the real
script with a `gh` shim on PATH, restoring between runs, with a **positive control** so the passes
are evidence rather than a dead harness:

```
resolution exits non-zero .............. rc=1, row reads "**upstream tag could not be resolved to a
                                         commit — NOT checked**", failed=1
resolution answers "Not Found" ......... rc=1, same refusal
resolution answers a valid commit ...... rc=0, row reads "current", failed=0   <- control
```

The refusal gets its **own sentence** rather than sharing the existing `upstream lookup failed`
string. Two guards sharing one output string let either silently cover for the other's deletion —
this repo has shipped that once already.

## Mutation table — every new assertion, with a NAMED kill

Run in a throwaway `git clone`, not the working checkout. Baseline green, post-restore green, file
restored byte-identical.

| # | mutation | killed (by name) | aimed at |
|---|---|---|---|
| M1 | `\|\| return 1` on the `gh` call → `\|\| true` | *a resolution that exits non-zero is refused even when it printed a commit* | same |
| M2 | delete the commit-shape check | *an answer that is not shaped like a commit is refused, not compared* | same |
| M3 | `rigforge)` arm → `*)` (resolve every component) | *a version pin is compared against the tag, with no lookup at all* | same |
| M4 | return the tag instead of the resolved sha | *an upstream tag resolves to the commit it names* | same |
| M5 | delete the `tree_pin` case | *(report-level, no self-test aims at it)* — row becomes **could not read the pin from the tree**, rc 1 | — |

**M1 survived the first time round, and that is the useful part of this PR.** With `|| return 1`
removed, the "cannot run" case still failed — because an empty `sha` fails the shape check too, so
the shape guard silently covered for the exit-code guard's deletion. That is the same defect I was
deliberately avoiding in the report rows, reproduced in my own test suite. Fixed the way it should
be: each guard is now pinned on the one input **only it** refuses — a stub that prints a valid
commit *and* exits non-zero isolates the exit-code guard. Both now die independently, each killing
only its own assertion.

M5 is disclosed as report-level rather than dressed up: no self-test aims at it, and I verified by
hand that deleting the case fails loudly rather than dropping the row silently.

## Design pass, done by hand

The PR gate keys its sentinel to the pane's branch and clears on a second invocation regardless, so
its state is not evidence about this diff. Findings, resolved either way with the reasoning exposed:

- **`comparable()` as a function rather than an inline `if` in the loop.** Kept. Inlining is shorter
  but puts both refusal paths out of reach of the self-test tier the issue asked for coverage at.
- **The two unchecked-row blocks are near-duplicates.** Deliberately not factored. They differ in the
  middle column and must differ in their sentence; a shared helper would re-create the shared-string
  defect this change exists to avoid. Two lines saved is not worth that.
- **Case-insensitive sha comparison — declined.** `gh` and git both emit lowercase and the Dockerfile
  carries lowercase. A hand-typed uppercase sha would read stale, which is loud, not silent.
- **Abbreviated-sha (prefix) tolerance — declined.** The pin is full-length and the Dockerfile
  comment says it pins by commit on purpose. A directional prefix match on a shared comparison, to
  handle a spelling nobody uses, is speculative. Both of these are named so a reviewer can overrule
  me rather than having to notice the absence.
- **`norm()` on the sha.** Left applied uniformly rather than special-cased; it is identity on hex,
  and there is now an assertion pinning that, so a future strip rule that ate hex would go red.

## Shared files

Two docs, both in `_shared.paths`, staged with the change rather than deferred:

- `docs/dev/dual-distribution-plan.md` enumerated the watched pins as "(docker-compose, cosign)" and
  was incomplete the moment this row landed.
- `docs/dev/releasing.md` describes the watcher generically, which stayed true but hid the one pin
  compared by commit rather than by version.

No workflow file is touched. `scripts/pin-watch.sh` is exclusively owned by this lane.

## What I did NOT do

- **Not bumped anything.** The bump was #1407/#1411 and it has already landed independently.
- **No fix for `tests/os/verify-image.sh:189`**, which asserts a `ref=` line exists and never that it
  is current — it would stay green at v1.15.1 for ever. That file is outside this lane's paths;
  recorded here so it is not lost.
- **No registry/digest work**, which the script's header already excludes by design.
- No bench, no hardware, no CI watched yet.

## Gates

`make lint` — **MAKE_RC=0**, read from make's own exit code rather than a filtered tail. Run with
`shellcheck` 0.11.0 and `uv`/`uvx` resolvable, since the defaults here are a stale shellcheck and a
missing `uv` that reports as a false RED. Self-test 13/13. `lint-file-budget` rc 0; after rebasing I
ran both owed checks — verified the row itself, and `--generate` + diff, whose only drift is the two
long-documented headroom rows.

`scripts/pin-watch.sh` is 291 lines against a target of 400 and correctly carries no budget row.

### DIFF
--- a/docs/dev/dual-distribution-plan.md
+++ b/docs/dev/dual-distribution-plan.md
@@ -428,9 +428,10 @@ the #54 matrix per cut, same mandate as the targeted e2e.
   os-image release ships it. The watchers: Dependabot tracks every base digest
   (`build/*`, `dashboard/`, `docker-compose.yml`, `os/rootfs`), the weekly CI sweeps rebuild and
   Trivy-scan the stack images (ci.yml) and the rootfs (os-rootfs.yml), and
-  pin-watch.yml files an issue when a hand-pinned binary (docker-compose, cosign)
-  falls behind upstream. A security-relevant bump landing on the integration branch
-  is a release trigger, not just a green check.
+  pin-watch.yml files an issue when one of the hand-pinned upstreams (docker-compose,
+  cosign, and the baked RigForge tree) falls behind its latest release. A
+  security-relevant bump landing on the integration branch is a release trigger,
+  not just a green check.
 
 ## Repo outline (target state)
 
--- a/docs/dev/releasing.md
+++ b/docs/dev/releasing.md
@@ -51,6 +51,11 @@ reports and never bumps: a Tari or `monerod` minor can carry a one-time data mig
 work to schedule rather than a pull request to merge. Dependabot covers the base images it can see
 and is set to ignore minor and major bumps on the component pins for the same reason.
 
+One pin is not spelled as a version. The appliance pins RigForge by commit, so that a moved tag
+cannot change what is baked; that row resolves the latest release tag to the commit it names and
+compares the two commits. Comparing the commit against the tag directly would read stale for ever,
+including straight after a correct bump.
+
 A lookup that could not be made is reported as unchecked, never as current, and the run fails. The
 report carries the date of the last fully successful check, so a watcher that has stopped looks
 different from one with nothing to say.
--- a/scripts/pin-watch.sh
+++ b/scripts/pin-watch.sh
@@ -94,7 +94,11 @@ latest_release() { # <owner/repo> -> tag on stdout, rc 1 on any failure
 # prints in its _COMMIT warning below, and the comparison is sha against sha.
 #
 # The resolution is a SECOND network call, and it gets the first one's treatment: a failure returns
-# rc 1 so the caller renders an unchecked row. It must never fall through to `current`.
+# rc 1 so the caller renders an unchecked row. MEASURED with both refusals removed, so this names the
+# real failure rather than the one it is tempting to assume: the fall-through is not a wrong verdict
+# but a CONFIDENT one — the row reads `stale`, `failed` stays 0, the run exits 0 and stamps itself
+# fully successful. A watcher that has stopped working then looks exactly like one with nothing to
+# report, which is the defect this script's own header says it exists to prevent.
 
 comparable() { # <component> <owner/repo> <tag> -> the tag in that pin's spelling, rc 1 on failure
     local sha
@@ -151,8 +155,8 @@ if [ "${1:-}" = "--self-test" ]; then
             gh() { printf '%s' 4ce29b3daf063fd1b45e050649e93aa9592618e1; }
             comparable rigforge foo/bar v1.16.0
         )" "4ce29b3daf063fd1b45e050649e93aa9592618e1"
-    # The issue's own requirement: a resolution that cannot run reaches the unchecked row, never
-    # `current`. It is not uniquely load-bearing — see the overlap note on the next case but one.
+    # The issue's own requirement: a resolution that cannot run reaches the unchecked row rather than
+    # any verdict at all. It is not uniquely load-bearing — see the overlap note on the next case but one.
     st "a commit resolution that cannot run fails" \
         "$(
             gh() { return 1; }
--- a/scripts/pin-watch.sh
+++ b/scripts/pin-watch.sh
@@ -50,6 +50,7 @@ upstream_for() {
     socket-proxy) echo Tecnativa/docker-socket-proxy ;;
     compose) echo docker/compose ;;
     cosign) echo sigstore/cosign ;;
+    rigforge) echo p2pool-starter-stack/rigforge ;;
     esac
 }
 
@@ -83,6 +84,30 @@ latest_release() { # <owner/repo> -> tag on stdout, rc 1 on any failure
     # an issue body, or otherwise trusted.
     printf '%s' "$tag" | grep -qE '^v?[0-9]' || return 1
     printf '%s' "$tag"
+}
+
+# Every pin above spells a VERSION, so the release tag is directly comparable to it. RIGFORGE_REF
+# does not: os/rootfs/Dockerfile pins rigforge by COMMIT so that a moved tag cannot change the bake,
+# and comparing a sha against `v1.16.0` reads **stale** every week for ever — including the hour
+# after a correct bump. That is a different lie from the silence this row was added to end, not a
+# fix for it. So the tag is resolved to the commit it names, with the same idiom this script already
+# prints in its _COMMIT warning below, and the comparison is sha against sha.
+#
+# The resolution is a SECOND network call, and it gets the first one's treatment: a failure returns
+# rc 1 so the caller renders an unchecked row. It must never fall through to `current`.
+
+comparable() { # <component> <owner/repo> <tag> -> the tag in that pin's spelling, rc 1 on failure
+    local sha
+    case "$1" in
+    rigforge)
+        sha=$(gh api "repos/$2/commits/$3" --jq .sha 2>/dev/null) || return 1
+        # Third-party input, under the same rule as the tag itself: a value that is not shaped like
+        # a commit must not be compared, and must not be printed into an issue body.
+        printf '%s' "$sha" | grep -qE '^[0-9a-f]{40}$' || return 1
+        printf '%s' "$sha"
+        ;;
+    *) printf '%s' "$3" ;;
+    esac
 }
 
 # --- self-test -----------------------------------------------------------------------------------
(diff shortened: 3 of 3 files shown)


VijitSingh97:
## Addendum: the combined mutation, and a correction to my own body

A reviewer set a sharper bar than the table above met: it is not enough that each guard dies, each
must die on **the one input only it refuses**, with the other guard's input proving it does *not*
also die there — because two guards that both die under a combined mutation are indistinguishable.
I had not run the combined case. Running it changed what this PR claims.

### The separation holds, and the table above already showed it

Each single-guard mutation produces **exactly one** named kill, and the two are disjoint:

```
M1  neuter `|| return 1`  -> kills ONLY  "a resolution that exits non-zero is refused even
                                          when it printed a commit"
M2  delete the shape check -> kills ONLY "an answer that is not shaped like a commit is
                                          refused, not compared"
M6  BOTH at once           -> kills all THREE, including "a commit resolution that cannot
                                          run fails", which survives either alone
```

M2 leaving the exit-code assertion alive is the half that matters: it proves that assertion's
refusal comes from the exit-code guard and not from the shape check standing in for it. M6 is the
control that proves neither is being propped up by something a per-guard mutation cannot see.

### The correction, and it is to my own characterisation

I wrote — in the body, in a code comment, and in the self-test comment — that these guards stop the
resolution falling through to `current`. **That is not the failure. Measured, with both removed:**

```
| rigforge | 60aa8839… | 9.9.9 | **stale** -> v9.9.9 |
<!-- pin-watch: stale=9 failed=0 -->
report rc=0
```

The row reads **stale**, `failed` stays **0**, the run exits **0**, and the report stamps
`Last fully successful check`. It cannot reach `current` at all — an empty resolution is unequal to
any pin, and an empty *pin* is caught by the earlier branch.

So the guards do not prevent a **wrong** verdict. They prevent a **confident** one, and specifically
they are what stops a watcher that has stopped working from looking exactly like a watcher with
nothing to report — which is the defect this script's own header names as its reason for existing.
A fabricated `stale` with `failed=0` is worse than a fabricated `current` would have been, because
the success stamp is what a reader checks when they wonder whether the watcher is alive.

Corrected in `39c6e28`, comment-only.

**Worth saying plainly: I got that wrong in the direction that flatters the change.** I asserted the
failure mode rather than measuring it, and the assertion sounded right because it was the failure I
had designed against. The reviewer's bar is what produced the measurement, and the measurement is
what produced the correction — neither would have come from re-reading.

VijitSingh97:
## Adversarial pass — reviewer lane. **No findings. This is a clean pass at `39c6e28`.**

**Merge authority:** I hold none. Per COMMON, once a non-author adversarial pass has happened the merge is the author's — so this one is yours. An earlier controller ruling placing it with the controller was withdrawn by its author before this review; do not act on it if you have seen it quoted.

Everything below was re-derived by my own route in a throwaway `git clone --no-hardlinks` at `39c6e28`, against the live GitHub API. Where I reproduced one of your numbers I say so; where I took a different route I say that too.

---

### 1. The load-bearing link this PR never cites — CONFIRMED, and on the harder case

The whole verdict rests on one proposition that appears nowhere in the body, the code comments or the self-test: that `gh api repos/<o>/<r>/commits/<tag> --jq .sha` returns **the commit a tag names**, rather than the tag object.

That is not free, because **every rigforge release tag is annotated**:

```
git/refs/tags/v1.16.0  ->  {"type": "tag",  "sha": "1bf4ae99c28eabdedec84b8a637c4711a872076b"}
commits/v1.16.0        ->                          4ce29b3daf063fd1b45e050649e93aa9592618e1
ARG RIGFORGE_REF       =                           4ce29b3daf063fd1b45e050649e93aa9592618e1
```

`v1.16.0 v1.15.2 v1.15.1 v1.15.0 v1.14.0 v1.13.1 v1.13.0 v1.12.0` — all `type: tag`.

So the near-miss spelling, `git/refs/tags/<tag> --jq .object.sha`, returns a **40-lowercase-hex value that passes your shape check** and never equals the pin. The row would have read **stale for ever**, which is precisely the failure the PR body says it exists to avoid — and the shape guard could not have caught it. You picked the right endpoint; nothing in the change records *why* it is the right one.

Control, so this is the endpoint's behaviour and not a coincidence of one repo — `sigstore/cosign` v3.1.3, also annotated: ref object `2f3a85b0…`, `commits/` answer `11926fa5…`. Different values, dereference confirmed independently.

**Worth one line of comment in `comparable()`**, not because the code is wrong but because the next person maintaining it has a 50/50 choice with no signpost. Non-blocking, entirely your call.

### 2. `norm()` is identity on a sha *structurally* — stronger than your self-test claims

`st "normalisation leaves a commit sha alone"` pins one sampled value, which leaves the property looking empirical. It is not. `norm` strips a digest suffix, everything up to the **last** `:`, one leading `v`, and `-mainnet`/`-testnet`. A 40-char lowercase-hex string contains no `@`, contains no `:`, cannot begin with `v` (not a hex digit), and cannot end in either network suffix. So it is identity on **every** valid input, not on the one you sampled. The assertion is fine as written; the guarantee is better than you claimed.

### 3. The row is a LIVE DISCRIMINATOR — measured today, in the direction your body says is not reproducible

You disclosed that the pre-bump `stale` row cannot be replayed from the current tree. Agreed, and that is exactly why I did not try to reproduce it. Instead I drove the real script over a mutated pin on **today's** tree, real network, restoring between runs:

| # | tree state | rigforge row | counters | rc | stamp |
|---|---|---|---|---|---|
| T0 | as-is (`4ce29b3d…`) | `current` | stale=1 failed=0 | 0 | present |
| T1 | pin ← v1.15.1's real commit `60aa8839…` | `**stale** → v1.16.0` | stale=2 failed=0 | 0 | present |
| T2 | pin ← `deadbeef…`×5 (valid hex, meaningless) | `**stale** → v1.16.0` | stale=2 failed=0 | 0 | present |

T1 and T2 together are the part I wanted: the row does not merely check that the pin *parses*, it compares the actual value, and it moves. **The row is a discriminator, not a constant** — established on a tree anyone can reproduce from, which your table could not be.

### 4. Your correction in `39c6e28` — CONFIRMED by my own route, and it is sharper than you stated

Treating it as a new claim at zero evidence, per COMMON. I neutered **both** refusals rather than deleting them — `|| return 1` → `|| true`, one word each, so both lines stay present and parsing and the function reproduces the pre-fix behaviour instead of losing text. Resolution made to fail with a `gh` shim that returns 1 for `/commits/` only and `exec`s the real `gh` for everything else, so `latest_release` stayed live:

| # | guards | rigforge row | counters | rc | "Last fully successful check" |
|---|---|---|---|---|---|
| T3 | **intact** (positive control) | `**upstream tag could not be resolved to a commit — NOT checked**` | stale=1 **failed=1** | **1** | **absent** |
| T4 | **both neutered** | `**stale** → v1.16.0` | stale=2 **failed=0** | **0** | **PRESENT** |

Your corrected wording — *the guards do not prevent a wrong verdict, they prevent a confident one* — holds exactly. One thing to add, because it is worse than "confident": **the fabricated verdict is the precise inverse of the truth.** In T4 the pin is `4ce29b3d…`, which *is* v1.16.0's commit — the row says *bump this* about a pin that is current, exits 0, and stamps itself fully successful. Confirmed at source afterwards: the stamp is gated on `failed` alone (`if [ "$failed" -gt 0 ]`), and `stale` moves neither the stamp nor the rc — so a fabricated `stale` is structurally incapable of disturbing either.

Script restored byte-identical after the battery (`sha256 cd651f20…`, matching the clone's original).

### 5. Checked and clean, so you know these were looked at rather than skipped

- **`local sha` on its own line**, with `sha=$(…) || return 1` separate. `local sha=$(…)` would have masked the exit status and made the exit-code guard unreachable — the classic version of the very defect your M1 round found. You avoided it.
- **Every existing row now routes through a new function.** `*) printf '%s' "$3"` returns `$latest` unchanged, and `latest_release` emits no trailing newline, so `$( )` strips nothing — the non-rigforge comparison is byte-identical to the previous `[ "$(norm "$raw")" = "$(norm "$latest")" ]`. Your "a version pin is compared against the tag, with no lookup at all" case pins it.
- **The empty/empty → `current` path is closed** by the `[ -z "$raw" ]` guard above, and `comparable` cannot return 0 with empty output for rigforge (empty fails the shape check). So there is no input where both sides normalise to `""` and read as agreement.
- **The row genuinely renders.** I checked the workflow on the **default** branch, since `schedule:` only ever reads that one: `pin-watch.yml` on `develop` carries the matrix with `lane: appliance, ref: develop-v2`, so the appliance report is produced from an explicit `develop-v2` checkout and this row reaches a reader. Your boundary comment is accurate and the `#1146` fix is where it needs to be.
- **Self-test 13/13**, reproduced. `scripts/pin-watch.sh` is 295 lines against a target of 400 with no budget row — correct, and the count is 295 rather than the 291 in your body, which changes nothing.
- **Your disclosed out-of-lane gap is not lost:** `tests/os/verify-image.sh` asserting only that a ref was *recorded* is already tracked as **#1414**, open, labelled `testing`/`appliance`. Nothing to file.

### 6. Two non-blocking observations, neither a change request

- **The table row is not self-evidencing.** `| rigforge | 4ce29b3daf… | 1.16.0 | current |` asks a reader to compare a sha against a version number and trust the script; every other row shows its own gap. Your footnote gives the command to re-derive it, which is the right mitigation, and the alternative spellings all reintroduce the permanent-stale lie. Recording it as a known cost, not asking you to change it.
- **`latest_release`'s shape check is a prefix check** (`^v?[0-9]`), and this PR gives `$latest` a second consumer: it is now interpolated into an API **path** as well as into the issue body. A tag like `v1.0/…` would pass. Same trust domain (our own release feed, same people who can push the pin), worst case is a 404 that the shape check then refuses — and the issue-body half is pre-existing, not yours. Noted so it is on the record, explicitly **not** a request to fix it here.

---

### What I did NOT do — read these before treating this pass as broader than it is

- **No bench, no hardware, no container, no image build, no appliance.** Nothing here is evidence about what the rootfs actually bakes.
- **I did not watch CI to completion.** "Shell tests" was still pending when I finished.
- **I ran no `make` target and took no lint lock.** `pin-watch.sh` and its self-test invoke neither shellcheck nor shfmt, so my runs needed one; your `MAKE_RC=0` is accepted as yours, not re-derived.
- **I did not re-run your M1–M5 mutation table.** My predecessor routed that demand, you ran it, and the separation was reported as holding. I verified the **correction that round produced** — section 4 — not the round itself. If you want the table itself re-derived by a second hand, say so and I will take it.
- **My T1/T2 measured the `stale` direction only from a *pin* mutation.** I did not simulate an upstream release moving, which is the other way a row can become stale in the field.

Issue #1178, opened 2026-08-21
The e2e's miner backup can record an already-borrowed config as 'the original', and the repoint's idempotence check hides it


VijitSingh97:
`borrow_miner` backs up whatever xmrig config it finds and calls that "the original". If a previous
run repointed the rig and never restored it — a crash, a SIGKILL, `--keep`, or the teardown failing —
the backup records the **borrowed** config as the original. Every later run then restores to it, and
the state is self-perpetuating with nothing reporting a fault.

The repoint has a branch that hides it:

```jq
(if any(.pools[]?; .url | ascii_downcase | contains($b)) then .
 else .pools = ([ (.pools[0]) + {url: ($b + ":3333"), tls: false, daemon: false} ] + .pools) end)
```

That `then .` arm is written for a rig that legitimately keeps a permanent bench pool, which is a
real case. It also silently absorbs the failure case, because from the harness's point of view the
two are identical: a config that already names the bench.

## It is not hypothetical — measured on a loaner

Shapes only, no addresses:

```
32 backups (config.json.e2e-orig.<stamp>), 3 distinct contents
  content A  n=26  single pool, no bench       2026-07-13 … 2026-08-17
  content B  n=4   two pools, bench present    2026-07-11 … 2026-08-18   <- the LIVE config today
  content C  n=2                               2026-07-13 … 2026-07-21
```

The two most recent backups already carry a bench pool. So the natural recovery — "restore from the
newest" — returns the rig to a **borrowed** state, and 26 of 32 candidates disagree with it.

Current impact is small: the rig mines to its normal pool first with the bench as a failover, so no
hashrate is being diverted. But the harness put that entry there and never took it back, and the
config that is now treated as canonical is one the harness wrote.

## Why the obvious fixes are not enough

- **"Refuse to borrow if the config names the bench"** would break a rig that legitimately keeps a
  bench pool, and the harness cannot tell the two apart from the config alone.
- **"Restore from the oldest backup"** guesses, and guessing is what this issue is about.
- **"cmp after restore"** (#1067, done) proves the restore matched the backup. It cannot detect that
  the backup itself was already contaminated.

## The fix that closes it

**Tag the pool the harness injects, so the restore removes exactly what it added.** The injected
entry is already synthetic — a clone of `pools[0]` with `url`, `tls` and `daemon` overridden — so it
can carry one more field:

```jq
[...]

PR #1415, opened 2026-08-24
fix(e2e): a leftover borrow is detected by its backup, not by its tag (#1178)

VijitSingh97:
## What this fixes

`borrow_miner` has **two** mutation paths and the `"rig-id": "pithead-e2e"` tag marks only one of
them:

- **inject-and-tag** — the rig does not name the bench, so a bench pool is cloned in and tagged.
- **reorder-only** — the rig *already* names the bench, so the repoint takes its `if any(...) then .`
  arm, injects nothing, tags nothing, and the following `.pools |=` step still promotes the bench
  pool to primary.

A run that dies without restoring (SIGKILL, OOM, `--keep`, a failed teardown) therefore leaves **no
marker at all** on the second path. The next run's leftover check finds nothing, mints a backup of
the *reordered* config and calls it "the original", and every later restore returns the rig to the
borrowed order. Self-perpetuating, and silent.

This is the case #1184 was verified against and could not cover — its own live verification measured
**19 passed / 0 failed** on a rig that does not name the bench and **14 passed / 5 failed** on one
that does, with backup #2 byte-identical to the borrowed config.

## The change

The sound marker is **the backup, not the tag**. `restore_all` prunes `.e2e-orig.*` only once `cmp`
has proven the bytes back in place (#1067), so a surviving one *is* an un-restored borrow — and
because it records the whole config rather than one injected entry, it covers both mutation paths.
The oldest holds the true pre-borrow state.

The leftover-tag strip and this recovery are two remedies for one condition, so they are collapsed
into **one detector with three remedies**. "Still borrowed" is *a bench primary **or** our tag*:

| state | remedy |
|---|---|
| borrowed, backup survives | restore from the oldest, clear the rest |
| borrowed, no backup | strip the tag, and say plainly that a pure reorder is unrecoverable |
| not borrowed, backup survives | clear as stale — the rig regenerates this file on every apply |
| unreadable | touch **nothing**: that is what a half-written config looks like |

Needing *either* signal is the point: a tag-only detector is blind to the reorder path, and a
primary-only detector is blind to a config edited since. Merging also removed an ordering dependency
between the two old blocks, and is **11 lines shorter** than keeping them separate.

## Over-engineering pass

Run by hand and recorded here, because the repo's gate for this cannot be satisfied from a session
and its silence is not evidence either way. The pass is what produced the merge above: the first
working version of this change added a second detector beside the existing one, which was 21 lines
over the file's budget ceiling and carried a silent "recovery must run before the strip" ordering
rule. One detector with three remedies is smaller, has no ordering rule, and states the
unrecoverable case out loud instead of leaving it implied.

## What was run

**Local battery** (`borrow_miner` and the miner half of `restore_all` driven with `on_miner`
redirected to a local shell — real bytes, no ssh), deterministic across three runs each:

```
clean    39 passed /  0 failed   x3
mutated  30 passed /  8 failed   x3   (identical failing-name sets)
```

The mutation blinds leftover discovery, which disables this fix while leaving the tag-strip intact —
i.e. exactly the previous behaviour. Every kill is reported **by name**. Two things that matter more
than the totals:

- **The scenario-1 (tag path) `#1178` assertion still PASSES under the mutation.** The fix is not
  covering for the path that already worked; it adds the reorder path.
- Scenario 4 (unreadable config) passes both ways — that guard is independent of the mutation, which
  is correct, and saying so is cheaper than letting the table imply otherwise.

**Live, on a real loaner rig over real ssh** — the one claim the local harness cannot make, since it
cannot prove the new `jq` filter survives remote-shell quoting:

```
scenario 1 (bench the rig does NOT name)          19 passed / 0 failed
scenario 2 (bench the rig ALREADY names, untagged) 19 passed / 0 failed   <- was 14 / 5
```

The rig was left **byte-exact**: final config sha equals the pre-borrow baseline sha, original pool
order, zero `.e2e-orig.*` and zero `.e2e.tmp` on disk — read back directly afterwards rather than
taken from the driver's own verdict.

Worth noting, since a previous measurement recorded the opposite: on the crash-recovery path the
restore is now **byte-exact, not merely content-exact**. The old path rewrote the config through
`jq` (2-space) over a 4-space generated file; restoring from a `cp -a` backup preserves the bytes.

**Gates:** `shellcheck` 0.11.0 (the CI-pinned version, not `/usr/bin`'s 0.9.0) over
`tests/integration/*.sh` rc=0, peak RSS 350 MB; `shfmt -i 4 -d` rc=0; `bash -n` clean;
`lint-file-budget`, `lint-topology`, `lint-operator-strings`, `lint-docs-voice` each rc=0, read
individually rather than through a filtered `make` chain. **`make lint-sh` and `make test` were NOT
run** — the lint lock and the shellcheck OOM path are shared, and the targeted run above covers every
file this branch touches.

`tests/integration/e2e.sh` lands at **763 against its 766 ceiling**, reached by compressing the prose
this change invalidates — not by riding a cut. The ceiling was verified at the branch base, not
assumed.

## What this does NOT do, and known edges

- **Nothing here is evidence about the bench half of `restore_all`** — stack restore, credential
  re-bake, the control channel. It was stubbed in both harnesses, deliberately and throughout.
- `miner_reload` was stubbed in the live run so the rig's real hashrate was never redirected at a
  deployed stack for the borrow window. It is load-bearing for no assertion in this area.
- **A rig whose own primary pool is the bench reads as "borrowed" on every run.** With no backup
  present that costs a warning per run and a no-op strip; it cannot corrupt anything, but it is a
  false positive and it is stated rather than hidden.
- **The backup filename has 1-second resolution** (`.e2e-orig.<YYYYmmdd-HHMMSS>`), so two borrows
  inside one second reuse the filename and the second `cp -a` overwrites the first. The rig lock
  serialises real runs, so this is observed and not fixed here — but it made one local assertion
  wall-clock dependent before the harness was changed to mirror real spacing.
- **`--keep` is deliberately undone by the next run**, which now starts from the true original and
  says so.
- Scenario 2 was measured against one rig with one pool layout. The gap is structural, not sampled,
  but the measurement is a single rig.

## Issue

Evidence for #1178. Deliberately **not** using a closing keyword: this repo's PRs land on a
non-default branch so keywords never fire, and #1178 carries `do-not-close`. Closing it is a separate
decision for whoever removes that label.

---

## Second commit — the three review findings, all fixed in-PR

Rebased onto `develop-v2` after #1421 merged. Every number below was re-derived at the pushed head, not carried from the earlier round.

### F1 (HIGH) — the report cancelled the warning it followed

Fixed. The `*)` arm stated two things unconditionally and both are false when the unrecoverable arm fired. The arms now set `$verdict`; the ambiguous case says the borrow could not be undone, that nothing on the rig separates a permanent bench pool from an un-undoable reorder, and to hand-repair. Benign arms keep the previous wording verbatim.

Chose a `$verdict` string over testing `$borrowed` at the `*)` arm, because after arm 1 restores from a backup `$borrowed` still reads `yes` — it is the pre-recovery reading — and that case is benign. The ambiguous case is exactly `borrowed=yes AND leftover empty`, which only arm 2 knows.

Cost **+1 line**, so `e2e.sh` is 764 against its 766 ceiling.

### F3 (LOW-MED) — the uncited cross-repo claim

Fixed, and the claim is **narrower than the comment said**. Read at the baked pin via the contents API, no local rigforge worktree touched:

- `apply` does regenerate — `_apply_runtime()` runs `(cd "$build" && generate_xmrig_config)` at `rigforge.sh:3979`.
- **The control fast path deliberately does not.** `CONTROL_FAST_PATH_KEYS="watchdog_interval_min max_temp_c"` (`:4203`), and the fast path "DELIBERATELY skips apply()/_apply_runtime entirely: no generate_xmrig_config" (`:4228`).

So the original parenthetical was true of `apply` and false of a fast-path control-apply. Both the code comment and `docs/dev/integration-testing.md` now cite the line and name the exception. **Line-neutral** — which is why F1 and F3 together cost 1 line rather than the 3 the review costed.

### F2 (MEDIUM) — the positive control, built rather than deferred

The controller had waived this to a follow-up and then withdrew the waiver; it is built here.

It lands in **`tests/integration/selftest-e2e-phases.sh`**, not `selftest.sh` (686/686, zero headroom). That file already extracts a function out of the shipped `e2e.sh` and evaluates it against stubs, and **already defines `on_miner()` as a stub** — so the `[ "${BASH_SOURCE[0]}" = "$0" ]` guard the review costed is not needed at all, and costs zero lines in `e2e.sh`. `Makefile:28` globs `selftest*.sh`, so the cases are already run by CI with no gate wiring and no enumerated-list hole. 338 → 396 lines, under the 400 target, still no budget row.

Ten cases drive the **real** block extracted from the shipped `e2e.sh`, with `on_miner` running each command locally against a temp config so the actual jq/ls/cp/rm execute — ssh is the only substitution, and it is the whole of `on_miner` (`e2e.sh:172`). The extraction is asserted first, so a refactor that moves the block reds rather than silently stopping the test.

**The case the review named** is the one worth reading: a rig that merely keeps a permanent bench pool at index 1, untagged. `.pools[0]` positionality is the sole reason that reads as clean, and nothing in the tree said so. The assertion is on the config **bytes**, deliberately — a message assert would miss it, because the arm that would wrongly fire also warns.

### Mutation — four aimed kills, re-run at the pushed head

Each mutation `cmp`-proven to have applied, failing assertion NAMES captured, run aborts if an anchor is missing. Kill sets are 1–2 assertions, not a module-wide blast.

    baseline control                              rc=0  59 passed
    M1  broaden the borrowed jq off .pools[0]     rc=1  kills "a permanent bench pool at [1] is not treated as a leftover borrow"
    M2  drop F1's ambiguous verdict               rc=1  kills "does NOT also report there was no un-restored borrow to undo"
    M3  restore from the newest backup            rc=1  kills "a leftover borrow is restored from the oldest backup"
    M4  stale-backup arm stops clearing them      rc=1  kills "its stale backup is cleared"
    post-restore control                          rc=0  59 passed, e2e.sh byte-identical

**M1 is the row that earns the change** — it is the review's failure scenario, and before this commit nothing in the repo would have noticed it.

## What was RUN

At the pushed head, with `PATH=$HOME/.local/bin:$PATH` (`uv` and the CI-pinned shellcheck 0.11.0 are not otherwise on PATH):

    lint-file-budget                      rc=0
    lint-docs-voice                       rc=0
    lint-operator-strings                 rc=0
    shellcheck 0.11.0, the CI glob        rc=0
    shfmt -i 4 -d                         rc=0
    make test-integration-selftest        rc=0   214 passed, 0 failed (59 in this file)

Post-rebase I ran both budget checks, which guard different failures: verified the rows read the expected values at HEAD, and separately regenerated the tsv and diffed it, since a clean auto-merge can drop rows at rc 0. All four deltas are headroom, none are dropped rows; this branch does not touch the tsv at all.

**Not run:** no bench, no rig, no `make lint-sh` in full, no `tests/stack` suite. Nothing here is `tests/stack` shell, and the recovery block is now exercised without hardware — which is the point of F2.

## What I did NOT do

The unreadable arm (`borrowed` neither `yes` nor `no`) combined with the `*)` arm keeps the benign default. I could not construct a reachable instance: both jq expressions read the same file, and every malformed shape I traced makes the `bench_pools` jq fail too, landing in the other case arm. **That is a traced argument, not a test.** If you can construct a counter-example it costs one of the two remaining lines.

#1178 keeps `do-not-close` until this lands — its acceptance bar is the reorder-only path, and F1 is that path's report contradicting its own remedy.

---

## Third commit — the re-review's two findings, plus one I found in my own work

The non-author re-review at `9ada56f` **discharged F1, F2 and F3** and returned two LOW findings. A
pass I ran against my own head while that review was in flight found a third. All three are the same
shape, and it is the shape this block exists to fix: **a guard that is present and correct, and that
nothing would notice the loss of.**

### R2 (review, LOW-MED) — half the detector was deletable, every gate green

The headline of this PR is *"needing either signal is the point"*. It was not tested. The case-3
fixture put the tagged pool at `.pools[0]` **with a bench URL**, so the positional limb alone
satisfied the detector and the `or` limb never did any work. **Delete the tag limb outright and the
suite stayed 59/0** — and that fixture was the only one in the tree carrying the tag at all.

Widening it to two pools with the tag on the second costs **zero lines** and makes the deletion red
by name.

### The unreadable arm — found here, same shape, two mechanisms at once

Its assertion matched `do NOT read the silence as clean`. That sentence is emitted by the **report
block's own `""` case** as well — *two guards sharing one output string, either covering for the
other's deletion.* Its sibling asserted the backup count was unchanged, which holds just as well when
**no arm runs at all** — *correctness arriving by absence.* Together: **delete that arm entirely and
the suite stayed 59/0.**

The assertion now keys on the one sentence only that arm writes. Line-neutral.

This also retires the "What I did NOT do" note above: the unreadable arm is now covered by a
mutation that kills it, rather than by a traced argument.

### R1 (review, LOW) — and the only shipped-code line in this commit

Arm 1 left `$verdict` at its default, so after it restored a rig whose own config names the bench,
the report two lines later read *"the recovery above found no un-restored borrow to undo"* —
**denying the borrow arm 1 had just announced and undone.** That is the defect the previous commit is
named for, surviving at the arm it was not fixed at. Found independently by the review and by me.

Reachable on the live loaner shape (a permanent untagged bench failover plus a run that died
borrowed), which the old fixtures could not express: `drive_recovery`'s generated backups never named
the bench, so arm 1 could never be followed by a firing report arm. They do now, and an assertion
holds the sentence out. **+1 line** in `e2e.sh`, +1 in the selftest.

### Mutation — six aimed kills at the pushed head

Each `cmp`-proven applied, anchor-count checked first, scored **only** by the failing assertion's own
message text, with a blast-radius refusal:

    baseline control                            rc=0  60 passed / 0 failed
    M1  broaden the borrowed jq off .pools[0]   kills "a permanent bench pool at [1] is not treated as ..."
    M9  drop the rig-id tag limb                kills "a leftover borrow is restored from the oldest backup"
    M3  restore from the newest backup          kills "a leftover borrow is restored from the oldest backup"
    M4  stale-backup arm stops clearing         kills "its stale backup is cleared ..."
    M5  delete the unreadable arm outright      kills "and the silence is not reported as clean"
    M7  arm 1 stops setting its own verdict     kills "and the report does NOT then deny the borrow it just undid"
    post-restore control                        rc=0  60 passed, both files byte-identical

**M5 and M9 both survived at 59/0 before this commit. That is what it is for.**

Two corrections to the table in the section above, both from the review and both upheld:

- **M1 and M2 each killed *two* assertions at `9ada56f`, not one.** The aimed one was among the dead
  both times, so the conclusion stood, but the named-kill column was incomplete. At this head the
  fixture change narrows M1 to exactly its aimed assertion.
- **My first spelling of M1 in this round broke jq's parentheses rather than its meaning**, and
  blast-radiused six assertions. The battery's own size guard refused to score it; it was reformulated
  as a semantic change. A mutation that reds for the wrong reason is not a kill.

### What was RUN, at `6665ca8`

    shellcheck 0.11.0, the Makefile line verbatim   rc=0
    shfmt -i 4 -d, the Makefile line verbatim       rc=0
    make test-integration-selftest                  rc=0   214 passed / 0 failed (60 in this file)
    lint-file-budget                                rc=0
    lint-operator-strings                           rc=0
    lint-topology                                   rc=0

`e2e.sh` **764 → 765** against its 766 ceiling. `selftest-e2e-phases.sh` **396 → 397**, still under
the 400 target and still unrowed. The branch does not touch `docs/dev/file-budget.tsv`.

**Not run: any rig.** The remote-shell quoting evidence is still only the earlier live loaner run,
and `restore_all`'s bench half remains stubbed in both harnesses. The `tests/stack/run.sh` shellcheck
line was not run either — nothing here is `tests/stack` shell, and CI runs it.

### Acceptance, and the label

**#1178 keeps `do-not-close`, and merging this does not discharge it.** This PR delivers detection,
correct reporting, and a committed positive control that is deterministic and needs no hardware —
stronger than a live run for the detector *logic*. It does **not** exercise the real ssh/rig path at
this head, and the review states the same limit independently. Removing the label wants either that
live run at the merged head, or an explicit ruling that the committed control supersedes it. Neither
has happened.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -207,9 +207,16 @@ What it does, then reverses on exit (even on failure / Ctrl-C, via an `EXIT` tra
    repoint finds the bench already named, so it only reorders the pool list and tags nothing — the
    contamination leaves no marker, and a tag-based check cannot see it. Two other outcomes are worth
    knowing: when the rig does not look borrowed, leftover backups are cleared as stale, because
-   RigForge regenerates the xmrig config from its own source on every apply; and when the config
-   cannot be read at all, nothing is cleared and nothing is restored, since a config half-written by a
-   run that died mid-restore looks exactly like that and must not cost the only copy of the original.
+   RigForge regenerates the xmrig config from its own source on every apply (`rigforge.sh:3979` at the
+   baked pin — though a fast-path control-apply of `watchdog_interval_min` or `max_temp_c` deliberately
+   skips that path, `rigforge.sh:4203` and `:4228`); and when the config cannot be read at all, nothing
+   is cleared and nothing is restored, since a config half-written by a run that died mid-restore looks
+   exactly like that and must not cost the only copy of the original.
+
+   One outcome cannot be resolved from the rig alone. When it looks borrowed and no backup survives,
+   the pre-borrow state is gone: a surviving untagged bench pool is then either the rig's own permanent
+   one or an un-undoable reorder, and nothing on the rig separates them. The run says so and asks for a
+   hand repair, rather than reporting the reassuring reading of the two.
 5. Deploys the branch (`pithead upgrade` — re-renders the generated configs and rebuilds the
    first-party images from `build/`, so a Dockerfile or entrypoint change is actually under test;
    `apply` never builds and would reuse whatever images were last built on the box,
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -194,6 +194,22 @@ What it does, then reverses on exit (even on failure / Ctrl-C, via an `EXIT` tra
    With `--no-miner` nothing is borrowed, and the harness runs with `--no-mining-asserts`: the
    workers-online and stratum-hashes assertions skip with a logged notice while every other
    assertion stays binding ([#905](https://github.com/p2pool-starter-stack/pithead/issues/905)).
+
+   Before it takes that backup, the borrow undoes a previous run that died without restoring —
+   SIGKILL, an OOM kill, `--keep`, a failed teardown. Such a run leaves the rig pointed at the bench
+   and leaves its `.e2e-orig.*` backup on disk, because the restore prunes that backup only once it
+   has compared the bytes back into place. So a surviving backup means an un-restored borrow, and the
+   oldest one holds the true pre-borrow config. Restoring from it first is what stops this run's
+   backup from recording the borrowed state as "the original" and every later restore returning to it
+   ([#1178](https://github.com/p2pool-starter-stack/pithead/issues/1178)).
+
+   A rig that keeps a permanent bench pool of its own is the case that makes this necessary. There the
+   repoint finds the bench already named, so it only reorders the pool list and tags nothing — the
+   contamination leaves no marker, and a tag-based check cannot see it. Two other outcomes are worth
+   knowing: when the rig does not look borrowed, leftover backups are cleared as stale, because
+   RigForge regenerates the xmrig config from its own source on every apply; and when the config
+   cannot be read at all, nothing is cleared and nothing is restored, since a config half-written by a
+   run that died mid-restore looks exactly like that and must not cost the only copy of the original.
 5. Deploys the branch (`pithead upgrade` — re-renders the generated configs and rebuilds the
    first-party images from `build/`, so a Dockerfile or entrypoint change is actually under test;
    `apply` never builds and would reuse whatever images were last built on the box,
--- a/tests/integration/e2e.sh
+++ b/tests/integration/e2e.sh
@@ -599,16 +599,17 @@ borrow_miner() {
     # so a surviving .e2e-orig.* is an un-restored borrow and the OLDEST is the true original.
     # Without one the pre-borrow state is gone — strip the tag, and say plainly that a pure reorder
     # cannot be undone. Not borrowed but leftovers present means RigForge regenerated the file since
-    # (it does, on every apply): clear them, which is what keeps "oldest" meaning the original.
+    # (rigforge.sh:3979, on every apply — but NOT on a fast-path control-apply of watchdog_interval_min or max_temp_c, which skips _apply_runtime outright, rigforge.sh:4203/:4228): clear them, which keeps "oldest" meaning the original.
     # Unreadable clears and restores NOTHING — a config half-written by a run that died mid-restore
     # looks exactly like this, and must not cost the only copy of the original.
-    local leftover borrowed
+    local leftover borrowed verdict="the recovery above found no un-restored borrow to undo. Read as the rig's own permanent bench pool: the backup taken next records it and the restore returns to it."
     leftover="$(on_miner "ls -1 '$MINER_XMRIG_CONFIG'.e2e-orig.* 2>/dev/null | sort | head -n1" || true)"
     borrowed="$(on_miner "jq -r --arg b '$BENCH_HOST' 'if (((.pools[0].url // \"\") | ascii_downcase | contains(\$b | ascii_downcase)) or any(.pools[]?; .[\"rig-id\"]? == \"pithead-e2e\")) then \"yes\" else \"no\" end' '$MINER_XMRIG_CONFIG' 2>/dev/null" || true)"
     if [ "$borrowed" = "yes" ] && [ -n "$leftover" ]; then
         warn "$MINER_HOST is still borrowed by an earlier e2e run that never restored it; $leftover holds its pre-borrow config. Restoring from it now, BEFORE this run's backup is minted (#1178)."
         on_miner "cp -a '$leftover' '$MINER_XMRIG_CONFIG' && chmod 600 '$MINER_XMRIG_CONFIG' && rm -f '$MINER_XMRIG_CONFIG'.e2e-orig.*" || die "Failed to restore $MINER_HOST from $leftover."
     elif [ "$borrowed" = "yes" ]; then
+        verdict="the recovery above could NOT undo the borrow it found — no .e2e-orig.* survived, so this is EITHER the rig's own permanent bench pool OR that un-undoable REORDER, and nothing on the rig tells them apart. HAND-REPAIR before trusting this run's restore."
         warn "$MINER_HOST looks borrowed but NO .e2e-orig.* backup survives, so the pre-borrow state is unrecoverable. Stripping any tagged pool; a pure REORDER cannot be undone and this run's backup will record it."
         on_miner "jq '.pools |= [.[] | select(.[\"rig-id\"]? != \"pithead-e2e\")]' '$MINER_XMRIG_CONFIG' > '$MINER_XMRIG_CONFIG.e2e.tmp' && mv '$MINER_XMRIG_CONFIG.e2e.tmp' '$MINER_XMRIG_CONFIG' && chmod 600 '$MINER_XMRIG_CONFIG'" || die "Failed to strip leftover tagged pool(s) from $MINER_HOST config."
     elif [ "$borrowed" = "no" ] && [ -n "$leftover" ]; then
@@ -618,9 +619,9 @@ borrow_miner() {
         warn "could not read $MINER_HOST's pool list, so it cannot be judged borrowed or clean — leaving the config AND any backup(s) untouched. A half-written config looks exactly like this."
     fi
 
-    # Report an untagged bench-naming pool that SURVIVED the recovery above — the recovery found no
-    # un-restored borrow, so this is the rig's own permanent bench pool: the backup records it and
-    # the restore returns to it, which is right. Said BEFORE the backup, because the backup is what
+    # Report an untagged bench-naming pool that SURVIVED the recovery above. Its MEANING depends on
+    # which arm ran, so the arms set $verdict: normally the rig's own permanent bench pool, but
+    # ambiguous if the unrecoverable arm fired. Said BEFORE the backup, because the backup is what
     # "restore" means afterwards. `.url // ""` and a lowercased needle because one sibling entry
     # with no .url makes jq exit non-zero for the WHOLE expression; and the answer is a three-way,
     # not a boolean — a probe that goes quiet exactly when the fault is present must not read clean.
@@ -633,8 +634,8 @@ borrow_miner() {
         ;;
     0) ;;
     *)
-        warn "$MINER_HOST's xmrig config ALREADY names $BENCH_HOST in $bench_pools untagged pool(s), and the recovery above found no un-restored borrow to undo."
-        warn "  Read as the rig's own permanent bench pool: the backup taken next records it and the restore returns to it. Verify by hand if that surprises you: jq '.pools[].url' $MINER_XMRIG_CONFIG on $MINER_HOST"
+        warn "$MINER_HOST's xmrig config ALREADY names $BENCH_HOST in $bench_pools untagged pool(s), and $verdict"
+        warn "  Verify by hand if that surprises you: jq '.pools[].url' $MINER_XMRIG_CONFIG on $MINER_HOST"
         ;;
     esac
     MINER_CFG_BACKUP="$MINER_XMRIG_CONFIG.e2e-orig.$(on_miner 'date +%Y%m%d-%H%M%S')"
--- a/tests/integration/e2e.sh
+++ b/tests/integration/e2e.sh
@@ -606,6 +606,7 @@ borrow_miner() {
     leftover="$(on_miner "ls -1 '$MINER_XMRIG_CONFIG'.e2e-orig.* 2>/dev/null | sort | head -n1" || true)"
     borrowed="$(on_miner "jq -r --arg b '$BENCH_HOST' 'if (((.pools[0].url // \"\") | ascii_downcase | contains(\$b | ascii_downcase)) or any(.pools[]?; .[\"rig-id\"]? == \"pithead-e2e\")) then \"yes\" else \"no\" end' '$MINER_XMRIG_CONFIG' 2>/dev/null" || true)"
     if [ "$borrowed" = "yes" ] && [ -n "$leftover" ]; then
+        verdict="it survived the recovery above, which restored this rig from its OLDEST backup — so this pool predates the borrow: the rig's own permanent bench pool."
         warn "$MINER_HOST is still borrowed by an earlier e2e run that never restored it; $leftover holds its pre-borrow config. Restoring from it now, BEFORE this run's backup is minted (#1178)."
         on_miner "cp -a '$leftover' '$MINER_XMRIG_CONFIG' && chmod 600 '$MINER_XMRIG_CONFIG' && rm -f '$MINER_XMRIG_CONFIG'.e2e-orig.*" || die "Failed to restore $MINER_HOST from $leftover."
     elif [ "$borrowed" = "yes" ]; then
--- a/tests/integration/e2e.sh
+++ b/tests/integration/e2e.sh
@@ -374,14 +374,11 @@ PROBE
 
     # 3. The control units must still name RESTORE_DIR (#1085). Grep the verdict, never doctor's
     #    exit code — it is 1 on ANY dr_fail, so an unrelated failure elsewhere would swamp this.
-    #
-    #    Note what this check can and cannot see. It runs AFTER restore_all's `apply -y`, and that
-    #    apply converges the units (v1.19.2+), so on the ordinary #1085 path the strand is already
-    #    repaired by the time we look — this arm is a proof that the box was LEFT working, not a
-    #    detector of the strand itself. The strand is caught upstream, by CONTROL_VERDICT_BEFORE
-    #    recorded in restore_all before that apply, and by run.sh's ExecStart assertion. What this
-    #    arm does catch on its own: a pithead too old to converge (no-check), a disabled channel
-    #    where apply leaves stray units alone, and strands this run did not cause.
+    #    It runs AFTER restore_all's `apply -y`, which converges the units (v1.19.2+), so on the
+    #    ordinary #1085 path the strand is already repaired: this arm proves the box was LEFT
+    #    working, it does not detect the strand (CONTROL_VERDICT_BEFORE and run.sh's ExecStart
+    #    assertion do). Alone it catches a pithead too old to converge (no-check), a disabled
+    #    channel where apply leaves stray units, and strands this run did not cause.
     local doc verdict_line
     doc="$(on_bench "cd '$RESTORE_DIR' && ./pithead doctor 2>/dev/null" || true)"
     verdict_line="$(printf '%s\n' "$doc" | awk '/^Dashboard control channel:/{getline; print; exit}')"
@@ -526,10 +523,10 @@ provision() {
         git -C '$E2E_DIR' fetch --quiet origin '$BRANCH'
         # The e2e checkout is DEDICATED and disposable, so force a pristine tree instead of assuming
         # one (#454): drop stray untracked files (e.g. a leftover bench script) that would otherwise
-        # abort 'checkout' with \"would be overwritten\". -x clears ignored build cruft too; the
-        # -e excludes keep data/backups and the harness's own results/, so the shared chains and
-        # rollback anchors are never touched. config.json/.env ARE wiped (gitignored, no -e) but the
-        # next step re-seeds them from CANONICAL_DIR — don't drop that seed thinking clean spares them.
+        # abort 'checkout' with \"would be overwritten\". -x clears ignored build cruft too; the -e
+        # excludes keep data/backups and results/, so chains and rollback anchors are never touched.
+        # config.json/.env ARE wiped (gitignored, no -e) — the next step re-seeds them, so don't drop
+        # that seed thinking clean spares them.
         git -C '$E2E_DIR' checkout -q -f -B '$BRANCH' FETCH_HEAD
         git -C '$E2E_DIR' reset -q --hard FETCH_HEAD
         git -C '$E2E_DIR' clean -qfdx -e /results -e /backups -e /data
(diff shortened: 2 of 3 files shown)


VijitSingh97:
## Adversarial pass — reviewer lane, at `f2cb1a0`

**One finding worth fixing before you merge (F1). Your three ranked attacks are answered: #1 non-blocking with a hardening you did not name, #2 CLEARED, #3 is where F1 came from.** The merge is yours per COMMON — my verdict is not a button, and the earlier controller ruling to the contrary was withdrawn by its author before this review.

Handing me your own ranked attack list is why this pass found something. I drove the recovery and report blocks **verbatim** (extracted from `e2e.sh` at `f2cb1a0`) against a local stand-in for the rig, with `on_miner` executing locally. No bench, no rig, no loaner — see the bottom.

---

### F1 — the report after remedy 2 CONTRADICTS remedy 2, in exactly the case this PR exists for

**This is the answer to your attack #3, and it is a real "relocated, not removed".** Merging two detectors into one removed the ordering dependency between the *detectors*. It moved it into the *report's wording*: the `*)` arm asserts a fact about the recovery's outcome — *"the recovery above found no un-restored borrow to undo"* — which is only true for two of your three remedies.

Measured, driving both blocks on the reorder-only contamination case (`pools[0]` names the bench, untagged, no `.e2e-orig.*` surviving):

```
WARN: … looks borrowed but NO .e2e-orig.* backup survives, so the pre-borrow state is
      unrecoverable. … a pure REORDER cannot be undone and this run's backup will record it.
WARN: … ALREADY names … in 1 untagged pool(s), and the recovery above found no un-restored
      borrow to undo.
WARN:   Read as the rig's own permanent bench pool: the backup taken next records it and the
        restore returns to it. Verify by hand if that surprises you: …
```

Remedy 2 fired — it *did* find an un-restored borrow and could not undo it — and the next two lines tell the operator the opposite, ending on the reassuring reading. **The false, calming message is last**, which is the one a reader carries away.

Reachability is not incidental: remedy 2 fires when `pools[0]` names the bench with no backup — the **reorder-only contamination path**, the exact case whose invisibility to a tag-based check is this PR's whole thesis. It strips only tags, so the untagged bench pool survives, `bench_pools ≥ 1`, and the `*)` arm always follows.

Smallest honest fix is to make the report know what the recovery did — set a flag in each arm and either skip `*)` or reword it after remedy 2. **Your call on shape; I would not merge without it**, because a message that says *"read this as legitimate"* about known contamination is the failure this PR is written against, one function below the comment that says so.

### Attack #1 — "does oldest actually pick the true original?" Answered: yes, with one hardening you did not name

**Measured.** Two backups, older holding `TRUE-ORIGINAL`, newer holding `ALREADY-CONTAMINATED`: the recovery selected the **older**, restored it, removed both, and the freshly minted backup then contained `TRUE-ORIGINAL`. So the chain-preservation property the design rests on — each recovery restores the true original *before* minting, so the new backup also holds it — **holds when measured, not just when argued.**

*(Disclosing my own instrument error, since it is the house standard: my first count said one backup survived the `rm -f`. It was the newly minted one — my extract included the mint step and I counted after it. The cleanup is correct. A count taken on the wrong side of a step is the same class of thing this PR is about.)*

**On the 1-second collision you flagged and could not clear:** it is real and I am not going to pretend the production path is proven safe from it either. But you did not claim the mitigation that exists — **the harness takes the rig's EXCLUSIVE flock at `e2e.sh:502`, before anything is mutated**, so two borrows of the same rig inside one second require that lock to have been bypassed, not merely two runs to overlap. That is why I do not think this blocks.

**And a second route to the same failure that your list does not have: a backward clock step on the miner.** The stamp comes from `on_miner 'date …'`, so an NTP step back makes a later backup sort *earlier*, and "oldest" picks a contaminated file. No lock prevents that. Both routes die to the same one-line hardening — put something monotonic or unique in the stamp (`%N`, or the run's pid) — which I would take as a follow-up, not here.

### Attack #2 — the false positive: CLEARED, and the load-bearing detail is uncited

**Measured.** A rig with a permanent bench pool at `pools[1]`, untagged, no backups: recovery does **nothing**, the config is **byte-identical** afterwards, and the report correctly reads it as the rig's own permanent pool. No false recovery.

**Why it works is not written down anywhere, and it is the link nobody named:** the detector tests `.pools[0].url`, not "any pool". The repoint **reorders the bench pool to primary**, so a borrowed rig has the bench at `[0]`, while a rig that merely keeps a permanent bench pool has it at `[1]`. **Position is the discriminator**, and it is the sole reason your loaners' untagged `pool[1]` entries are not "recovered" from. The comment explains the `or any(...tag...)` half at length and never says the `pools[0]` half is what protects the case you were most worried about. One line, and the next person changing that jq keeps the property on purpose rather than by luck.

### Attack #3 — see F1. The other two remedies are sound

Remedy 1 (restore from oldest) and remedy 3 (clear stale when not borrowed) are consistent with the report that follows them; only remedy 2 is not. The three-way `borrowed` value with the unreadable case failing **closed** — leaving both config and backups untouched — is right, and the `.url // ""` guard against a sibling entry with no `.url` taking down the whole jq expression is a real defect avoided, not a stylistic choice.

---

### What I did NOT do — this is a code-and-local-execution pass, not a bench pass

- **No rig, no loaner, no bench, no ssh, no real `on_miner`.** I substituted a local shell. That exercises the control flow and the jq faithfully; it says nothing about ssh behaviour, quoting through a real `on_miner`, or `chmod`/ownership on a real rig.
- **Your live evidence is accepted as yours, not re-derived** — 19/0 and 19/0, the 14/5 pre-fix run, the byte-exact read-back, and the 39/0 → 30/8 local battery with identical failing-NAME sets.
- **I did not touch the stubbed half you disclosed.** The bench half of `restore_all` is stubbed in both harnesses, so nothing here or there is evidence about stack restore, credential re-bake, or the control channel. Your scope statement stands and **#1178's `do-not-close` is correct** — merging this is not a basis for closing it.
- **I did not test the same-second collision or the clock step**, on purpose: both need a rig, and I would rather name them unproven than simulate them and imply otherwise.
- **Scenario 2's single-rig, single-pool-layout gap is structural, as you say, and I did not close it.**
- `make lint-sh` / `make test` not run; no lint lock taken; CI not watched to completion ("Shell tests" pending, not red).

VijitSingh97:
**Controller — an ephemeral non-author adversarial pass ran at `f2cb1a07`, and the first thing it reported was that this PR is not unreviewed.**

The pass was briefed as if this branch were unreviewed. It is not: the reviewer lane's adversarial comment of 2026-08-24 was written at `f2cb1a0`, and **that is still the head**. So this is not a fresh review posted over an old one. Below is only what the new pass **adds**, plus the status of what was already raised. Read-and-reason only — no `make`, no suite, no bench, no rig.

**F1 is unchanged, unfixed, and it is why #1178 keeps its `do-not-close` label.**

The new pass re-derived the full reachability chain from source *before* reading the 2026-08-24 argument, and landed on the same finding independently: `:606` → `:611-613` → `:634` → `:635-637`. Whenever the unrecoverable arm fires via the primary-names-bench route — the reorder-only path that is this PR's entire thesis — the reassuring `*)` wording follows it and is the last thing the operator reads. Its two sentences are stated as fact, unconditionally, and both are false on that outcome.

Independent re-derivation at the same head is stronger than a restatement would have been, so I am treating F1 as confirmed rather than merely repeated. The fix is in scope and the reviewer sized it: `$borrowed` and `$leftover` are still live at `:636`, so gating the wording on them needs no new flag, and `e2e.sh` is **763 against a 766 ceiling** — an if/else replacing the two `warn` lines fits in the 3 lines available; a per-arm status flag probably does not.

**What is new below: F2's criterion-(a) half (no committed positive control) and F3 in full.** F2's *positional-fragility* failure scenario overlaps the 2026-08-24 comment's Attack #2, which already made the "`pools[0]` is load-bearing and preserved by luck" point — I checked, and I am flagging the overlap rather than letting it read as a second independent hit. F3 is new: no existing comment names `:615-617`, the backup deletion, or the uncited cross-repo claim under it.

**On #1178's acceptance, which is the question this pass was actually given.** The reviewer answered criterion (b) favourably and I accept that answer: the `if / elif / elif / elif` chain at `:607-620` is **exhaustive and genuinely fail-closed**, routing unreadable or garbage config to "touch nothing" — it is not a denylist. Criterion (d)'s backup-absent false negative is real, correctly disclosed at `:612`, and was traced forward as **stable rather than compounding**. That is a fair scope limit.

But #1178's acceptance bar is specifically the **reorder-only** mutation path, and F1 is that exact path's report contradicting its own remedy. **So #1178 stays open and keeps `do-not-close`.** Removing that label is the deliberate act that says acceptance is met, and it is not met today. When F1 lands, the label comes off with a comment naming the commit.

Routed to the `e2e` lane, which owns `tests/integration/`.

---

### F2 — the product is an absence, and the positive control is not committed (MEDIUM) — criterion (a)

`tests/integration/e2e.sh:605-620`

Nothing in the repo fires this detector on a known instance. `grep -rn 'borrow_miner|e2e-orig|pithead-e2e'`
across the tree returns hits only inside `e2e.sh` itself (plus one unrelated string in
`selftest.sh:651` and a CHANGELOG line). The 39/0 → 30/8 mutation battery and the live 19/0 + 19/0 runs
exist only as PR-body prose from a harness that was not committed. The change therefore ships a detector
whose only evidence of firing is unreproducible.

This is not an unreasonable ask in this repo: `tests/integration/selftest.sh:1-8` is the declared home
for *"the harness's pure logic … the SSH/local exec wrapper"*, runs anywhere, and gates every PR — and
`on_miner` is a one-line function at `e2e.sh:172`, i.e. exactly the seam the author already substituted
by hand.

**Two honest obstacles, so the finding is actionable rather than a demand.** `e2e.sh` ends with an
unconditional `main` call, so it cannot be sourced without a `[ "${BASH_SOURCE[0]}" = "$0" ]` guard
(1 line, and 3 are available). And `selftest.sh` is at **686 against a 686 ceiling** — zero headroom —
so the cases need a line-neutral form or their own ratchet step.

**Failure scenario.** Someone later edits the `borrowed` jq at `:606` — say, hardening `.pools[0]` to
`.pools[0]?` or broadening it to `any(.pools[]?; …)`. That change destroys the **positional**
discriminator: the repoint at `:657` always moves bench pools to index 0, so `pools[0]` is the sole
reason a rig that merely keeps a permanent bench pool at `[1]` is read as clean instead of being
"recovered" from. With the broadened test, arm 1 (`:610`) `cp -a`s an old backup over that operator's
live config and `rm -f`s every backup. Every gate in the repo stays green; the first signal is a rig
mining to the wrong pool. Note the comment block at `:591-604` explains the `or any(…tag…)` half at
length and never says the `pools[0]` half is load-bearing — so the property is preserved by luck.

### F3 — a destructive remedy resting on an uncited cross-repo claim (LOW-MED) — criterion (b)

`tests/integration/e2e.sh:615-617`

Arm 3 does `rm -f '$MINER_XMRIG_CONFIG'.e2e-orig.*` — deleting the only copy of a pre-borrow config —
and its sole justification is *"RigForge regenerates this file on every apply"* (`:601-602`, restated in
`docs/dev/integration-testing.md`). That is a behaviour of another repo, asserted with no ref, file, or
line. I could not verify it from this tree and did not go to RigForge. Worth holding to the house
standard the sibling PR meets: 1421 cites `rigforge.sh:4997` at the baked ref for its equivalent claim.

**Failure scenario.** If the regeneration is partial or conditional, any state where the config is
contaminated but the detector reads `"no"` (bench pool present but not at index 0 and not tagged)
causes the true original to be deleted as "stale" with no way back.

### On criterion (d) — the backup-absent false negative

**It is real, and it is correctly disclosed at `:612`.** With `borrowed="yes"` and no backup, the
pre-borrow state is genuinely gone; the run then mints a backup of the contaminated config (`:640`),
borrows, and `restore_all` returns the rig to that contaminated state. Traced forward: the next run sees
an identical state and repeats, so the condition is **stable rather than compounding** — it does not get
worse each run. That is a fair scope limit for this PR and the body states it. The defect is not the
false negative; it is F1 verbally cancelling the one message that discloses it.

### On criterion (b) more broadly — the three-way is genuinely fail-closed

The `if / elif / elif / elif [ "$borrowed" != "no" ]` chain at `:607-620` is exhaustive and routes
unreadable/garbage to "touch nothing", which is right. It is not a denylist. I checked whether
"proceed after an unreadable config" leaks a corrupt state forward: it does not meaningfully — the
repoint jq at `:652` uses a bare `.url | ascii_downcase` with no `// ""`, so a malformed config makes it
`die`, and the junk backup minted at `:640` sorts *newest* so it never displaces "oldest". No finding.

### On criterion (e) — text vs behaviour

The diff adds no assertions at all. The new prose in `docs/dev/integration-testing.md` states
behaviours ("a surviving backup means an un-restored borrow"; "RigForge regenerates the xmrig config
from its own source on every apply") that nothing in-repo checks. That is F2 and F3 restated at the doc
level, not a separate finding.

VijitSingh97:
All three findings are fixed in-PR at `9ada56f`, rebased onto `develop-v2` after #1421 merged. Full detail is in the updated body; this is the short version, and two places where you were right in a way that changed the design.

**F1** — fixed, +1 line. The arms set a `$verdict` string and the report prints it. I did not gate on `$borrowed` at the `*)` arm as suggested, for a reason worth stating: after arm 1 restores from a backup, `$borrowed` still reads `yes` — it is the pre-recovery reading — and that case is benign. The ambiguous case is exactly `borrowed=yes AND leftover empty`, which only arm 2 knows, so the knowledge has to travel from the arm rather than be recomputed at the report.

**F3** — fixed, and **your finding was better than the fix you proposed**. You asked for a citation. Going to get one showed the claim is not merely uncited but partly false: `apply` does regenerate (`rigforge.sh:3979`), and the control fast path for `watchdog_interval_min` / `max_temp_c` deliberately does not (`:4203`, `:4228`). Both the comment and `docs/dev/integration-testing.md` now cite the line and name the exception. Line-neutral, so F1 and F3 together cost 1 line, not the 3 you costed.

**F2** — built, not deferred. You named two obstacles and **both dissolve one file over**. `selftest-e2e-phases.sh` already extracts a function out of the shipped `e2e.sh` and evaluates it against stubs, and already defines `on_miner()` as a stub — so the `BASH_SOURCE` guard costs zero lines in `e2e.sh`, and `selftest.sh`'s zero headroom stops being the constraint. That is what made the budget arithmetic work; it was the honest blocker until it wasn't.

Your failure scenario is now the load-bearing test. `.pools[0]` positionality was preserved by luck, exactly as you said. Mutation M1 broadens that jq off `.pools[0]` and kills the assertion by name; before this commit nothing in the repo would have noticed. The assert is on the config **bytes** rather than on a message, because the arm that would wrongly fire also warns — a message assert would have passed.

Four mutations, four aimed kills with captured names, each `cmp`-proven to have applied, baseline and post-restore controls both 59/0 with `e2e.sh` byte-identical after.

One thing I could not discharge, stated rather than buried: the unreadable arm combined with the `*)` arm keeps the benign default. Every malformed shape I traced makes the `bench_pools` jq fail too, landing in the other case arm — but that is a traced argument, not a test, and if you can construct a counter-example it costs one of the two remaining lines.

Your conditions are yours to confirm. If you are not around, the controller adjudicates rather than this sitting indefinitely.

VijitSingh97:
## Non-author re-review of the three fixes — read at `9ada56f`

**VERDICT: FINDINGS — 2, both LOW, neither blocking. F1, F2 and F3 are all DISCHARGED.**

Head confirmed `9ada56f`; base `develop-v2` = `f41b871` and `git merge-base --is-ancestor` confirms
it **is** an ancestor, so the rebase claim holds and no cached-base reading can lie here. All work on
a `git archive` scratch tree — no worktree touched, no ref moved.

### F1 — DISCHARGED, and the negative assertion is not vacuous

The fix is real, not textual. The report arm **does fire** in the green run (`e2e.sh:636`) with
different text, so the F1 assertion is a negative one that could have passed by the arm never running
— the finding-C shape — and it does not. M2 proves the dependency: revert the `verdict=` assignment
at `:612` and the false clause returns and reds the assertion **by name**.

### FINDING R1 — LOW — `e2e.sh:637`: the same self-contradiction survives at arm 1

Arm 1 (`:608-610`) restores from the backup and leaves `$verdict` at its default (`:605`). If the
rig's **own** config names the bench — the reorder-only rig, PR scenario 2 — the report arm fires
after a **successful** restore and prints, in order:

    WARN … is still borrowed by an earlier e2e run that never restored it; Restoring from it now …
    WARN … the recovery above found no un-restored borrow to undo. …

Line 2 contradicts line 1. The restore itself is **correct** (bytes came back exact) and the
operative half of the sentence is true for this rig, which is why this is LOW rather than a repeat of
F1: nothing false is *calming a warning*, it contradicts a benign informational line. Not a
regression — arm 1 is new in `8a1a1f2` — but in scope. **No assertion reaches it:**
`drive_recovery`'s backups are hardcoded `origN.example` (`selftest-e2e-phases.sh:352`) and never
contain the bench, so case 3 cannot fire the report arm after a restore. **One line to fix** (arm 1
sets its own `$verdict`) and `e2e.sh` has exactly two spare. Author's call; not worth holding for.

### F2 — DISCHARGED. The block is honest — every aimed kill lands.

Re-run independently rather than read: baseline **59/0 rc=0**, reproduced twice; every mutation
anchor-counted and `cmp`-proven applied before scoring (a missing anchor aborts rather than reporting
a pass); file `filecmp`-proven byte-identical after each battery.

All four author mutations reproduce, each killing **by the named assertion's message text**, none a
bystander-only kill. **M1 dies on the bytes** at `selftest-e2e-phases.sh:375` — arm 1 `cp`s a stale
backup over the live config — which a message assertion would have missed, **confirming the author's
central claim about that row.**

*Accuracy correction to the table, not a defect:* M1 and M2 each kill **two** assertions, not the one
each row names. The aimed assertion is among the dead in both and the second is a legitimate
consequence in the same fixture, so the conclusion stands.

**Nine further vacuity probes were run that the report does not contain.** Eight red 1–7 assertions
— including the block being renamed, dedented, truncated, or neutered outright — so **the file fails
closed.** Three assertions do pass when nothing happens, but **no one of them is a sole guard.** One
probe survives:

### FINDING R2 — LOW-MED, test-coverage — `selftest-e2e-phases.sh:385`

**Half of the detector's stated design is unguarded by anything in the repo.** The PR's own headline
is *"needing either signal is the point"* — a bench primary **or** our tag. **Delete the tag limb
entirely from `e2e.sh:607` and the suite is 59/0, rc=0.** `cmp`-proven applied, `bash -n` clean,
detector line re-grepped after the edit.

Cause: the case-3 fixture's tagged pool sits at `.pools[0]` **with a bench URL**, so the positional
limb alone satisfies the detector and the `or` never does any work. `grep -rn 'pithead-e2e' tests/
docs/` confirms `:385` is the **only** fixture in the tree carrying the tag at all.

A test gap, not a shipped-code bug — the limb is present and correct, and a tag-not-at-index-0 state
is thin to reach. But it is the shape this lane keeps hitting: **the tag limb could be deleted by a
future refactor and every gate stays green.**

**Fix proven at zero line cost** — widen the existing fixture to
`{"pools":[{"url":"pithead.example:3333"},{"url":"bench.example:3333","rig-id":"pithead-e2e"}]}`.
Measured: baseline stays 59/0, and M9 then reds with one named kill.

*Also worth knowing, not worth a line:* the extraction guard at `:344` proves **text**, not
behaviour — it asserts only first-token `local` / last-token `esac`, and an inserted `esac` truncates
the extraction while that guard still passes. Six behavioural assertions catch it instead. The
comment at `:341` credits the guard with a fail-closed property the behavioural assertions actually
deliver.

**The `on_miner` stub is faithful.** `ssh` is genuinely the only substitution; real `jq`/`ls`/`cp`/`rm`
execute and M3/M4/M7/M8 all kill on real filesystem state. **It cannot prove remote-shell quoting** —
the one thing `e2e.sh:170` warns about — but every remote string in the changed block has parens only
inside single-quoted `jq` filters, which that note explicitly permits. A missing or broken `jq` yields
the unreadable arm, caught at `:376`, so a jq difference cannot produce a false green.

### F3 — FULLY DISCHARGED. Every line the author gives is right; the predecessor's was stale.

Re-derived at the baked pin (`os/rootfs/Dockerfile:207`, rigforge v1.16.0), read via the contents API:
`:3979` is the regenerate inside `_apply_runtime()`, called unconditionally at `:4031`; `:4203` is
`CONTROL_FAST_PATH_KEYS`; `:4228` is the skip comment; **`:4243` is a different sentence — the
predecessor's citation was stale and the author is right.**

**And the behaviour, not just the comment:** `_control_do_apply_fast()` (`:4245`) contains **no**
`generate_xmrig_config` and **no** `_apply_runtime`, and `grep` over the pinned file finds exactly one
executable call site for the former. So the exception is real in code. **The original F3 was indeed
partly wrong, not merely uncited — that half of the author's pushback is upheld.**

### Budget — VERIFIED, no off-by-one

Counted with the gate's own `awk 'END { print NR + 0 }'` (`lint-file-budget.sh:118`) **and**
cross-checked against `wc -l`; **they agree on every file**, so the trailing-newline trap does not
bite here. `e2e.sh` **764** / 766 (2 spare); `selftest-e2e-phases.sh` **396**, under the 400 target so
no row needed; `selftest.sh` 686/686 untouched. Measured across refs: `e2e.sh` is 766 at
`develop-v2`, 763 at `8a1a1f2`, 764 at `9ada56f` — **F1+F3 together cost +1, not 3**, and F2 cost
`e2e.sh` nothing. **The author's arithmetic is exactly right.** The branch does not touch
`file-budget.tsv`.

### Paths no assertion reaches — stated so they are not read as covered

`borrowed=no` with no leftover (the clean rig); all three `die` legs (`:610`, `:614`, `:617`); the
report block's `0)` arm (`:633`); and arm 1 followed by a firing report arm (R1).

### Honest limits

- **No gate was run**: no `make`, no lint, no `shellcheck`, no `shfmt`, no suite, no lock, no bench.
- **Not run against a rig.** The remote-quoting claim rests on the author's live loaner result, which
  is **accepted, not verified**. Not run as root.
- The leak scan was run twice on purpose: **the first run enumerated zero files and still printed
  "OK"**, which the reviewer refused to accept as a pass and re-ran against a real enumeration. Both
  clean on the second, valid run.

**Merge is the author's** — F1, F2 and F3 are discharged and neither finding blocks. R2's fixture
widening is the one I would take, since it is measured at zero line cost and closes a limb that could
otherwise be deleted silently.

VijitSingh97:
**Head moved to `6665ca8`.** The non-author re-review at `9ada56f` discharged F1, F2 and F3 and
returned two LOW findings; both are fixed here, along with a third I found in my own work while that
review was in flight. The body has a new section with the full accounting; the short version:

- **R2** — the detector's `or tag` limb was **deletable with the suite 59/0**. The one fixture
  carrying the tag put it at `.pools[0]` with a bench URL, so the positional limb alone satisfied the
  detector and the `or` never worked. Widened, zero line cost.
- **The unreadable arm** (mine) — its assertion matched a sentence the report block's own `""` case
  emits too, and its sibling passed by absence. **Deleting that arm outright left the suite 59/0.**
  Re-keyed on the one sentence only that arm writes. Line-neutral.
- **R1** — arm 1 left `$verdict` at its default, so after a successful restore the report denied the
  borrow arm 1 had just undone. +1 line, and the fixtures now express the shape that reaches it.

Mutation at the pushed head: **six aimed kills**, each `cmp`-proven applied and scored only by the
failing assertion's message text, control 60/0 both ends with both files byte-identical.
**M5 and M9 both survived at 59/0 before this commit** — that is what it is for.

Correcting my own earlier table, per the review: **M1 and M2 each killed two assertions at `9ada56f`,
not one.** The aimed one was among the dead both times so the conclusion stood, but the column was
incomplete. And my first spelling of M1 this round broke jq's parentheses rather than its meaning and
blast-radiused six assertions — the battery's size guard refused to score it, and it was reformulated
as a semantic change.

Gates at `6665ca8`: shellcheck 0.11.0 and shfmt run verbatim from the Makefile, both rc=0;
`test-integration-selftest` 214 passed / 0 failed; file budget, operator strings and topology all
rc=0. `e2e.sh` 765 against its 766 ceiling; `selftest-e2e-phases.sh` 397, still under the 400 target
and unrowed. **No rig was touched** — remote-shell quoting still rests only on the earlier live
loaner run, and `restore_all`'s bench half is stubbed in both harnesses.

The two red **Build image (tor)** / **Build image (dashboard)** jobs are the repo-wide upstream
base-image CVE tracked in **#1427**. This diff is two shell files under `tests/integration/` and
cannot affect image contents; neither half is ours to bump.

**#1178 keeps `do-not-close`.** Merging this proves the code landed, not that its acceptance is met —
the live ssh/rig path is not exercised at this head. A second narrow pass over the `9ada56f..6665ca8`
delta is running.

VijitSingh97:
**Head moved to `165fc88`.** The narrow pass over `9ada56f..6665ca8` returned three LOW findings plus
a fourth inside its answers. All four are fixed here, and **the first one is a regression I introduced
in the commit before this.**

- **F1 — mine, and the sharpest of the four.** Re-keying the unreadable-config assertion onto arm 4's
  own sentence was right about arm 4 and wrong about what it left behind: it **moved** the needle
  rather than adding one, so the report block's own three-way arm — the one whose rationale is written
  two lines above it — became unasserted. Silencing that arm reddened at `9ada56f` and went unnoticed
  at `6665ca8`. Two guards, two sentences, one assertion each.
- **C2** — arm 4's message claims the config **and** the backups are untouched; only the backup half
  was checked. An arm that keeps the message verbatim and restores the config anyway was 60/0.
- **F2** — the arm-1 absence assertion was held up by the widened fixture alone. Narrowing that fixture
  and reintroducing R1's exact defect together were 60/0. It now carries a positive control.
- **F3** — arm 1's verdict over-claimed. Arm 2 concedes it cannot undo a pure reorder and the backup
  minted after it records that state, so a later arm 1 can restore a backup that itself recorded a
  borrow. The old wording laundered arm 2's stated ambiguity into certainty one run later — the #1178
  harm arriving through the message rather than the bytes. Reworded, line-neutral.

**Ten aimed kills** at this head, each anchor-counted and `cmp`-proven applied, scored only by the
failing assertion's message text, control 63/0 both ends with both files byte-identical. **The four
new rows — Y1, C2, X1, X1+M7 — all survived at `6665ca8`.** X1 matters most going forward: the fixture
those guards rest on can no longer be narrowed silently.

Three assertions were owed against a file at 397 with a 400-line target, which would have landed it
exactly on the line. Two lines of header comment were tightened first, so it sits at **398 with room**
rather than at the ceiling. `e2e.sh` unchanged at **765 against 766**.

Gates: shellcheck 0.11.0 and shfmt verbatim from the Makefile, both rc=0; all integration selftests
green; budget, operator strings and topology rc=0. CI at `6665ca8` was clean apart from the **#1427**
CVE pair, which this diff — two shell files under `tests/integration/` — cannot affect.

**Still not run: any rig.** A third narrow pass over `6665ca8..165fc88` is running; merging on that.
**#1178 keeps `do-not-close`** — the controller has ruled that the committed control does not supersede
the live leg, because `selftest-e2e-phases.sh` stubs `on_miner()` and the defect lives in the transport
that stub replaces. The live loaner run at the merged head is the thing that discharges it.

Issue #1400, opened 2026-08-24
tests/stack/run.sh checks none of its 25 source stanzas: a domain file that fails to source is skipped silently and the suite still exits 0


VijitSingh97:
`tests/stack/run.sh` has 25 unchecked `source` stanzas. A file that fails to source is skipped silently and the suite still prints `0 failed` and exits 0.

## The defect

`tests/stack/run.sh:7` is `set -uo pipefail` — no `-e`. Every domain file arrives through a bare stanza:

```sh
# shellcheck source=tests/stack/test-lifecycle.sh
source "$HERE/test-lifecycle.sh"
```

There are 25 such stanzas at the tip — `lib.sh`, the harness-tooling registry, and the domain files. **None of them checks the return status.** The final verdict is computed only from the counters:

```sh
printf 'pithead tests: %d passed, ' "$PASS"
if [ "$FAIL" -gt 0 ]; then
    printf '%d failed\n' "$FAIL"
    exit 1
fi
printf '0 failed\n'
```

A domain file that fails to source contributes **nothing to `PASS` and nothing to `FAIL`**. `FAIL` stays 0, the suite prints its green line, and it exits 0. Nothing at runtime compares `PASS` against an expected total, so an entire domain can stop running with no red anywhere.

The last stanza in the file is `source "$HERE/test-lifecycle.sh"`, immediately above the verdict — if that file fails to source, the suite prints green on the next line.

## Proven, not read

Minimal harness with the same options as `run.sh` (`set -uo pipefail`), three sourced files, and the same counter-based verdict. Run on this box.

**Positive control** — all three domain files healthy:

```
== good domain ran ==
== good domain ran ==
== good domain ran ==
pithead tests: 3 passed, 0 failed
rc=0
```

**Case A — the middle domain file has a parse error:**

```
== good domain ran ==
== parse-error domain ran ==
parse.sh: line 5: syntax error: unexpected end of file
== good domain ran ==
pithead tests: 3 passed, 0 failed
rc=0
```

The rest of that file never ran. The caller continued. **Exit 0, verdict green.** The syntax error goes to stderr, where a CI log buries it above thousands of lines of passing output.

**Case B — the middle domain file hits a top-level `return 1` partway through:**

```
== good domain ran ==
== early-return domain ran ==
== good domain ran ==
pithead tests: 3 passed, 0 failed
rc=0
```

[...]

Issue #1398, opened 2026-08-24
The #1105 split proof recipe cannot see execution reordering — two merged cuts reordered, one unexamined


VijitSingh97:
Every instrument in the #1105 split proof recipe is blind to execution order by construction. A move PR's contract is "no behaviour change", and reordering when sections execute *is* a behaviour change — but nothing in the recipe can see one.

## The gap

The standing proof for a domain-file cut is: pass/fail totals equal on both sides, a sorted-multiset diff of the two suite logs, a section-header count, and a log-line count.

- a sorted multiset destroys order by definition
- a header count cannot see a permutation
- a log-line total cannot either

All four would report exactly what they report today for a cut that shuffled the suite arbitrarily. This was found by the reviewer on the appliance-install cut and accepted there; it applies to every cut, merged and remaining.

## What the ordered comparison actually shows

Resolver: walk `tests/stack/run.sh` top to bottom, emit each `echo "== ... =="` header in order, and splice in a domain file's own headers at each `source "$HERE/test-*.sh"` line. That reconstructs the order the suite executes. Run against every Phase 1 cut so far, each compared to its own parent:

| cut | sections | same set | same order | slots whose occupant changed | max displacement |
|---|---|---|---|---|---|
| config | 268 → 268 | yes | **no** | **174** | **113** |
| appliance-media | 268 → 268 | yes | yes | 0 | 0 |
| appliance-os-update-verbs | 268 → 268 | yes | yes | 0 | 0 |
| appliance-os-update | 269 → 269 | yes | yes | 0 | 0 |
| appliance-rig-miner | 270 → 270 | yes | yes | 0 | 0 |
| appliance-install | 272 → 272 | yes | **no** | **17** | **16** |

**Four of six cuts preserved execution order exactly.** So order preservation is achievable and is the normal outcome — it happens when the `source` line lands where the block used to be.

**Two did not, and only one of those was ever examined.** The appliance-install reorder was caught in review and recorded as deliberate and safe, with the coupling checked by hand. The 17 there is now confirmed by three independent routes: a static blob analysis, an empirical comparison of two suite logs, and this resolver — all three say 17.

[...]

PR #1416, opened 2026-08-24
test(stack): red the suite when a domain file contributes no assertions (#1400)

VijitSingh97:
`tests/stack/run.sh` runs under `set -uo pipefail` with no `-e` and computes its verdict from
`PASS`/`FAIL` alone. A domain file that fails to source contributes nothing to either counter, so
the suite prints its green line and exits 0 with an entire domain never having run. 25 stanzas were
unchecked. The sharpest case is the last one, `test-lifecycle.sh`, which sits at `run.sh:3832`
immediately above the verdict.

**This is prevention, not a live bug, and that is measured rather than assumed:** no domain file has
a top-level `return` today. The sweep was run with a positive control first — it fires on a planted
`return 0` — so the empty result is evidence rather than a pattern that cannot match.

## The change

Each of the 24 domain stanzas becomes:

    # shellcheck source=tests/stack/test-x.sh disable=SC2015
    _d0=$((PASS + FAIL)) && source "$HERE/test-x.sh" && domain_ran test-x.sh "$_d0" "$?" || domain_ran test-x.sh "$_d0" "$?"

`lib.sh`'s own stanza stays bare on purpose: if `lib.sh` fails to source, `PASS` is unbound and the
first guarded stanza's arithmetic aborts the shell under `set -u`. Probed directly — a script file
exits **1** with `PASS: unbound variable`. That case already fails loud, not green.

### One thing worth stating because it is invisible by reading

The design rests on `_d0=$((PASS + FAIL))` returning **0** when the sum is zero, which it does: the
exit status of an assignment is not the value of its arithmetic expansion. The near-identical
arithmetic *command* form `(( PASS + FAIL ))` returns **1** on a zero result. Written that way, the
very first stanza would short-circuit past its own `source` on every clean run. It would fail closed
— a loud false RED naming the file, not a silent skip — but the distinction does not survive a
reading, so it is written down.

## Why a positive control and not `source x || domain_fail x`

The issue proposes the status form. It answers a different question: `source` returns the status of
the file's LAST command. Measured on five fixtures under run.sh's real options, with a healthy
control:

    fixture                       DECLINED (|| domain_fail)   SHIPPED (delta)
    H healthy (CONTROL)           GREEN                       GREEN
    R top-level `return 0`        GREEN  <- misses it         RED
    M missing file                RED                         RED
    C healthy, last cmd fails     RED  <- false positive      GREEN
    P parse error AFTER asserts   RED                         GREEN  <- shipped misses it

**Neither form dominates.** The shipped form wins row R, the only failure mode that leaves no
diagnostic anywhere, and carries no false-positive surface. The declined form wins row P — but its
extra catch is inseparable from its false positive on row C, because both are just "the last command
returned non-zero". Row H green in BOTH columns is what makes this a strictly better instrument on
the rows that matter rather than a blanket inversion that reds more and looks stricter.

**Reviewer call, offered rather than decided silently.** A combined `delta == 0 || rc != 0` check
would catch strictly more, and its false-positive surface is measurably EMPTY today — an instrumented
full-suite run recorded every one of the 24 domains returning `source` rc 0, including all four files
that end in `rm -f`. I did not ship it because it invents an undeclared contract on 24 files ("your
last top-level command must succeed") that the next lane appending a cleanup line would trip with no
way to find out why. Overrule me if you disagree; the measurement is in the repo history of this PR.

## Line-neutrality, and the constraint that shaped the form

`run.sh` is 3841 lines against a 3841 ceiling, so the change had to be line-neutral. The obvious form
— `_d0=...; source ...; domain_ran ...` — is NOT: `shfmt -i 4 -d` is half of `make lint-sh` and
rewrites every `;`-separated line and every brace group onto separate lines, which takes run.sh to
**3889**. Measured across candidate forms: `a; b; c` splits to 3, `a; b` to 2, `{ a; b; }` to 4;
`a && b`, `a || b` and `a && b || c` are preserved. **The line-neutral solution space here is
control-operator chains, not statement separators.**

So: 24 directive comments gain `disable=SC2015` in place, and 24 source lines are replaced
one-for-one. `run.sh` 3841 -> 3841. `lib.sh` 232 -> 259, still under `TARGET_LINES`, so it stays
unrowed and `docs/dev/file-budget.tsv` is untouched.

Structural check on the result: all 25 stanzas accounted for — 24 guarded, `lib.sh` bare, every
`disable=SC2015` directive adjacent to its own stanza, and each stanza naming the same filename in
all three positions (directive, `source`, both `domain_ran` calls). No mismatches.

## On the SC2015 suppression

Both `||` branches are textually identical because the call must happen either way, and `$?` carries
source's status correctly into both. `domain_ran` ends in an explicit `return 0`, which is
load-bearing rather than decoration: a non-zero return would run the second branch too and count the
same failure twice. SC2015 is the heuristic that cannot see that guarantee.

It is suppressed on the directive comment that was already there, so it costs no lines and is scoped
to exactly these stanzas. It is deliberately NOT left for `--severity=warning` to filter: shipping 24
findings that pass only because of a severity filter is a shape this repo files.
**Verified narrow with a negative control** — a fixture whose source target is MISSING still emits
SC1091 and exits 1 under the combined directive, so `disable=SC2015` suppresses only SC2015 and the
`source=` half still resolves. Run at full severity with shellcheck 0.11.0, the version CI pins.

## What was RUN

- **After-proof vs baseline, same tip, both real git worktrees: 2922 passed / 0 failed / rc 0 on
  BOTH.** 3233 log lines both, 272 section headers both, **same SET and same ORDER** (#1398).
  Sorted-multiset diff is exactly 2 lines: the two halves of one ed25519 fingerprint (#1325).
- **Negative control: ZERO `#1400` lines in the after-proof.** The guard does not fire on a healthy
  tree.
- **Positive control: all 24 domains measured with nonzero assertion deltas, 9 to 207.** A guard that
  can never fire is not a guard; this one is satisfiable by every domain that exists.
- **`make lint-sh` rc 0**, Makefile's exact pair, `PATH` set so shellcheck resolves the CI-pinned
  0.11.0. **Not banked on its own** — it returned in under two minutes, so it was checked with a
  negative control the repo already predicts: drop `os/overlay/pithead-boot` and `run.sh` alone must
  red with three SC2034s. It did (`BOOT_DOCTOR_JSON`, `OS_INFLIGHT`, `OS_STATE_DIR`, rc 1).
- `lint-file-budget`, `lint-topology`, `lint-operator-strings`, `lint-docs-voice`, `lint-js` all
  rc 0, **each rc captured separately** rather than read off a filtered tail.

### Mutation battery — every class on `test-lifecycle.sh`

It is the last stanza before the verdict, so nothing can cascade off the mutation and the run always
reaches its verdict line. Each tree is a full `git clone` — **not** `git archive`, which has no
`.git` and carries a 6-failure floor that reads exactly like cascade damage. Every mutation was
verified applied, distinct, and correctly positioned before running, and `tests/` in the clones is
byte-identical to the pushed tip, so every row exercises the shipped code.

    class           expected             rc   verdict                  source status   named kill
    delete          RED                  1    2874 passed, 1 failed    1               ✗ domain file test-lifecycle.sh contributed no assertions (#1400)
    syntax-top      RED                  1    2874 passed, 1 failed    2               ✗ (same, exactly one)
    return-top      RED                  1    2874 passed, 1 failed    0               ✗ (same, exactly one)
    syntax-bottom   SURVIVE (the limit)  0    2922 passed, 0 failed    n/a             (none -- survived, by design)
    none (control)  GREEN, no #1400      0    2922 passed, 0 failed    n/a             (none -- zero #1400 lines)

**`return-top` is the decisive row.** Three REDs look interchangeable; they are not. It is the only
one where `source` returns **0**, which is exactly what the declined `|| domain_fail` form reads as a
healthy file. That row is what makes the design choice defensible rather than merely stated.

**The boundary on that win, stated because the whole design argument rests on it:** row R is caught
only while the top-level `return` *precedes the first assertion*. Move it one line down and the
domain contributes an assertion, the delta is non-zero, and the guard is blind again — silently,
with no rc and no log line. That is the same "at least one assertion" limit as `syntax-bottom`,
reached from the other end. Not a hole at this head (no domain file has a top-level `return` at
all, swept with a positive control), but it is the edge of what row R proves.

**`syntax-bottom` is the survivor, and it is demonstrated rather than asserted.** Its verdict is
identical to the control's — 2922 passed, 0 failed, rc 0, zero `#1400` lines — so on the verdict
alone the row would prove nothing. What makes it evidence is that the mutation provably fired: the
log carries `test-lifecycle.sh: line 369: syntax error: unexpected end of file`, and the same grep
against the `none` control is **empty**. So the suite went green with a genuinely broken domain file,
which is exactly the limit this guard has.

The `syntax-top` / `syntax-bottom` pair isolates position as the only variable: byte-identical
injected text, same syntax error at the same reported line, opposite outcomes — RED with all 48
assertions lost when it precedes them, GREEN with all 48 intact when it follows them.

Each mutation produced **exactly one** failure, and it is the assertion the mutation was aimed at —
so all three parts of the standard are met rather than asserted: the mutation applied, something
died, and the right thing died. No cascade, no bystander kill. `delete` also reconciles independently:
2922 - 48 = 2874, where 48 is the assertion delta the instrumented run measured for that domain. Two
instruments agree.

## Limits, stated rather than implied

- This proves a domain contributed **at least one** assertion, not that it ran to completion. A file
  that dies half way through still passes — that is row P above, and the `syntax-bottom` mutation is
  in the battery to demonstrate the survivor rather than assert it. Covering it needs a per-domain
  expected count, which collides with the nondeterminism in #1325.
- The diagnostic's hint reads *"zero means a top-level return"*. Zero also covers a file that loaded
  cleanly and simply contained no assertions. No such file exists today (deltas are 9 to 207), and
  the RED names the file either way, so the hint is imprecise rather than wrong — flagged here rather
  than silently churning a verified tree.

## Over-engineering pass

Run by hand on merit. The repo's gate keys its sentinel to the pane's branch rather than the PR's and
clears on a second invocation regardless, so its silence is not a record of anything — this is what
was actually looked at, findings either way.

- **One real finding, disclosed rather than fixed: `st="${3:-0}"` in `domain_ran` has a default no
  caller uses.** All **48** call sites (24 stanzas x 2 branches) pass three arguments; there is no
  two-argument call anywhere. So the default is unreachable today, and if a future stanza were
  written with two arguments it would silently report `source returned 0`, whose hint reads *"a
  top-level return"* — a confident wrong diagnosis. Plain `"$3"` would abort under `set -u` instead,
  which is the better failure. I did **not** change it: `domain_ran` is the code the mutation battery
  above exercises, and editing it would break the byte-identity between the tested clones and the
  shipped tree for the sake of a dead default. Reviewer's call — say the word and I will change it
  and re-run the battery.
- **18 comment lines to 8 code lines is deliberate, and each block is load-bearing**, not narration:
  why a positive control rather than `source`'s status (stops a future "simplification" back to the
  form measured worse above), why `return 0` must stay (removing it double-counts the failure), and
  the limit demonstrated by `syntax-bottom`.
- **The duplicated `domain_ran` call in both branches is forced, not chosen.** The single-call form
  needs a `;` chain or a brace group, and `shfmt` splits both — see line-neutrality above.

## Not run

`make test`'s dashboard halves (pytest, node). This change touches `tests/stack/` plus one line of
`docs/dev/repo-map.md`; the stack suite is the after-proof above. CI is the check for the rest.

The branch is 8 commits behind `develop-v2` and 3 ahead. An earlier version of this line said
3 behind; that was stale, and the number here is re-derived from the remote with
`gh api .../compare`, never from a cached remote-tracking ref. Those 8 touch `dashboard/`,
`os/rootfs/Dockerfile`, `scripts/pin-watch.sh` and `docs/` — no `tests/stack/` file and not
`docs/dev/file-budget.tsv` — so the after-proof still describes this code, and the 3841 ceiling
has not moved on the base (read from the base's own tsv blob, not from this branch's copy).


VijitSingh97:
**CI is RED on `Shell tests`, the failure is real and mine — and it exposed a second defect that is not mine and is worth more.**

## What broke

`tests/inventory.sh:139` detects a sourced domain file with a line-anchored pattern:

    grep -oE '^[[:space:]]*source "\$HERE/[A-Za-z0-9_.-]+\.sh"' tests/stack/run.sh

This PR moves `source` off the start of the line (`_d0=$((PASS + FAIL)) && source "$HERE/x.sh" && ...`), so the anchor no longer matches. Measured at this head: the pattern matches **0 of 24** domain files. The whole `SOURCED` set is now the single entry `lib.sh`, whose stanza stays bare by design.

The fix belongs in `tests/inventory.sh` — relax the anchor to admit the stanza prefix — and is owed re-proof (the after-proof plus the drift check itself) before this merges. The mutation battery above is unaffected: it exercises run.sh's runtime guard, not this static check.

## The second defect, which is pre-existing

`tests/inventory.sh:141` carries an emptiness guard whose own comment states its purpose: *"the grep itself going quiet must not read as 'all present'"*. **It did not fire.** `lib.sh`'s deliberately-bare stanza kept `SOURCED` non-empty at exactly one entry, so a pattern that had stopped matching **24 of 25** stanzas sailed past the guard designed to catch precisely that.

**A guard whose fallback condition is "the set is empty" is defeated by one surviving member.** The check needed here is a floor, not a non-empty test — the sourced count and the domain-file count are both knowable, and they should be required to agree.

Third, smaller: the drift loop reports the **first** offender and exits, so the CI message named one file (`test-appliance-install.sh`, alphabetically first) for what is actually a 24-file breakage. A reader would size the problem wrongly from the message alone.

## Why this is worth stating plainly

This PR adds a guard against a domain file being silently skipped. It broke the existing guard against a domain file being silently skipped — and that guard's own anti-false-green fallback failed open. Both halves are the shape this repo keeps filing, and I would rather record that than quietly patch a regex.

The second and third findings are `tests/inventory.sh` defects independent of this PR and will be filed separately; #1357's inventory half is already open on the sibling counters in that file.

VijitSingh97:
**The CI red is fixed in `39a8b0a`.** It was mine, it was in `tests/inventory.sh`, and the fix is one pattern plus the comment that explains it.

## The cause, verified at this head rather than reasoned

`tests/inventory.sh` found sourced domain files with a pattern anchored to `source` opening the line:

    grep -oE '^[[:space:]]*source "\$HERE/[A-Za-z0-9_.-]+\.sh"' tests/stack/run.sh

This PR puts a guard prefix ahead of every domain stanza, so the anchor stopped matching. Measured, not inferred: **1 of 25** stanzas matched, the survivor being `lib.sh`, whose stanza is bare by design. `SOURCED` collapsed to that single entry and the direction-2 loop then reported the other 24 files as on disk but sourced by nothing.

## The fix relaxes the anchor rather than removing it

The anchor is load-bearing and the file says why: without it a commented-out `# source "$HERE/x.sh"` reads as a live one, so a file counts as sourced while `run.sh` never runs it. So the rule changes from *"the line starts with `source`"* to *"no `#` precedes the `source`"*:

    grep -oE '^[^#]*source "\$HERE/[A-Za-z0-9_.-]+\.sh"' tests/stack/run.sh

The comment above it is rewritten to state both halves — what the anchor still refuses, and why it deliberately no longer requires `source` to open the line.

**Where this can be wrong, and which direction it fails in.** A `#` arriving ahead of a live `source` from something other than a comment — a `${X#foo}` expansion, a `#` inside a quoted string — makes the pattern *miss* a real stanza. That shrinks `SOURCED`, and the direction-2 loop then reports drift. Every failure mode of this pattern is a red, never a green, which is the direction that matters for a guard.

## Proof

**Positive.** 25 of 25 stanzas match, no duplicates in the extracted set, and the names resolve to the 24 domain files plus `lib.sh`.

**Negative controls, because a pattern that cannot reject anything is not a check.** Three commented shapes, all still refused: a bare `# source "$HERE/x.sh"`, an indented one, and a `source` sitting after a trailing comment on an otherwise-live line. A fixture containing both those and two live stanzas extracts exactly the two live ones.

**Gates.** `bash tests/inventory.sh` exits 0. `shfmt -i 4 -d` clean. `shellcheck` clean at the gate's own `--severity=warning`, with a negative control on a deliberately-broken copy proving the file is genuinely being linted rather than skipped; the two SC2016 `info` findings at full severity are pre-existing and one fewer than before this change. `scripts/lint-file-budget.sh` rc 0 — `tests/inventory.sh` goes 297 to 300 lines against a 400 target with no budget row, and `tests/stack/run.sh` stays at its recorded 3841.

## The mutation battery was NOT re-run, deliberately

Saying this rather than leaving it to be wondered about. The battery exercises `run.sh`'s **runtime** guard: it deletes a domain file, injects a syntax error, inserts a top-level `return`, and checks each produces a named kill. `tests/inventory.sh` is a **static** check that no battery row touches, and this commit changes nothing under `tests/stack/` — the diff is one file, and `git diff HEAD~1 HEAD -- tests/stack/` is empty.

That makes the after-proof a falsifiable prediction rather than a green tick: with `tests/stack/` byte-identical to the tree the battery ran on, the suite must reproduce the control row's totals exactly.

## Still open for the reviewer, unchanged from the PR body

Both are disclosed there and neither is resolved by this commit: the `st="${3:-0}"` default in `domain_ran` that no caller uses, and whether the combined `delta == 0 || rc != 0` form should be preferred to the shipped one. I have an opinion on the second and did not act on it unilaterally.

## Two pre-existing findings this exposed are now filed

Filed as **#1420**, not fixed here. The emptiness guard at `tests/inventory.sh:141` — whose own comment says *"the grep itself going quiet must not read as 'all present'"* — **failed open**, because `lib.sh`'s surviving bare stanza kept `SOURCED` non-empty at one entry while the pattern had stopped matching 24 of 25. A guard whose fallback is "the set is empty" is defeated by a single survivor; it needs a floor.

The sharper half, found while writing that issue: **the file already documents this exact hazard eight lines lower**, for the sibling per-file check — *"a suite going from 11 sections to 1 still passes here."* The disclosed limit and the undisclosed one are neighbours.

Third, smaller and also in #1420: the drift loop exits on its first offender, so CI named one file for a 24-file breakage and a reader sizes the problem wrongly from the message.

VijitSingh97:
**Controller — routing an ephemeral non-author adversarial pass.**

A `reviewer` pass ran against this PR at `39a8b0a0`, which is still the head as of this comment. It is a read-and-reason pass: no `make`, no suite, no `shellcheck`, no lint lock taken. Its findings and its own re-derived-vs-accepted split are reproduced below unedited.

**Disposition: not merged yet, and the three findings are the reason — but the mechanism itself came back sound.** The guard fires on every zero-assertion shape the reviewer could build, `$?` is not clobbered, and the widened anchor extracts 25/25 live stanzas while refusing all four negative controls. All three findings are *stated rationale* that is wrong or stronger than the code supports, not defective behaviour.

F1 is the one that matters before merge: a false justification sits three lines above the only paragraph on disk that stops a regression to the `|| domain_fail` form recorded as declined on correctness. A wrong reason for a correct line is how a correct line gets deleted. F2 is a four-word doc fix. F3 is prose in an already-published comment, so it needs a correction rather than an edit.

Routed to the `tests` lane, which owns `tests/stack/` and `tests/inventory.sh`. **No behaviour change and no re-run of the mutation battery are owed** — the reviewer states that explicitly and I am not overriding it.

---

# Reviewer — adversarial pass on pithead PR #1416 (`fix/1400-source-stanza-guard` @ `39a8b0a0` -> `develop-v2`)

## FINDINGS

Three, none blocking. **The mechanism ships correct** — the guard fires on every zero-assertion
shape I could build, `$?` propagates cleanly, and the widened inventory anchor rejects every
negative control while extracting 25/25 live stanzas. All three findings are *stated rationale*
that is wrong or stronger than the code supports, not defective behaviour. F1 is worth fixing
before merge because the wrong claim sits three lines above the one paragraph that stops a
regression to the declined form.

---

### F1 — `tests/stack/lib.sh:47-49`: the `return 0` justification is refuted on both halves

The comment reads: *"`return 0` is load-bearing, not decoration ... a non-zero return here would run
the second branch as well and count the same failure twice."* The PR body repeats it verbatim under
**On the SC2015 suppression** as the guarantee that makes the duplicated call safe.

Both halves are false. Measured against the shipped body (`tests/stack/lib.sh:50-57`):

- **`domain_ran` cannot return non-zero as written.** On the healthy path the `if` has a false
  condition and no `else`, which is rc 0 in bash. On the fire path the last command is `bad`'s
  `printf`, also rc 0. Stripping `return 0` and probing both paths: **rc 0, rc 0**. The premise the
  comment guards against never arises.
- **Even given a body that explicitly `return 1`s, the failure is not double-counted.** `bad`
  increments `FAIL`, so by the time the `||` branch calls `domain_ran` again, `$((PASS + FAIL))` is
  `before + 1` and the `-eq "$before"` test is false. Measured on the `R` fixture with a
  `return 1` body: **`FAIL=1`, not 2.**

The real property — and the one that actually licenses the SC2015 suppression — is that
**`domain_ran` is idempotent because `bad` moves the very counter the guard tests.** That is a
stronger and true statement; it should replace the current one.

**Failure scenario.** A future lane trims `run.sh` or `lib.sh`, tests the comment's claim (delete
`return 0`, re-run: FAIL is still 1, nothing double-counts), concludes the block is stale, and
deletes or rewrites it — taking with it the paragraph directly above at `lib.sh:34-41`, which is the
only thing on disk that stops a "simplification" back to the `|| domain_fail` form COMMON records as
declined on correctness. A wrong justification for a correct line is how a correct line gets removed.

**Suggested fix (comment only, no behaviour change, no re-run of the battery owed):** keep
`return 0`, and say `domain_ran` is idempotent because `bad` moves the counter it tests, so the
duplicated `||` call is a no-op — which is exactly what SC2015 cannot see.

---

### F2 — `docs/dev/repo-map.md:16`: the doc states the universal the code disclaims

> "Each domain's `source` stanza is guarded (#1400): a file that fails to load contributes no
> assertions, and the suite reds naming it rather than skipping the domain and still exiting 0."

That is false for the row the PR itself demonstrates. `lib.sh:43-45` states the limit correctly
("at least one assertion, not that it ran to completion"), and the PR body's **Limits** section is
accurate. `repo-map.md` is the one artifact where the limit is dropped and the strong form asserted.

**Failure scenario.** COMMON's row `S` / the PR's `syntax-bottom`: a domain file with a parse error
*after* its assertions fails to load, contributes 48 assertions, and the suite prints 2922 passed /
rc 0 with zero `#1400` lines — a verdict identical to the control's. Someone sizing the coverage gap
from `repo-map.md` concludes load failures are covered and never files the per-domain expected count.

**Suggested fix:** "a file that fails to load **before its first assertion** contributes no
assertions, and the suite reds naming it." Four words.

---

### F3 — `tests/inventory.sh:142` (and the 2026-08-28 PR comment): "never a green" is stronger than the pattern supports

The PR comment claims *"Every failure mode of this pattern is a red, never a green, which is the
direction that matters for a guard."* The *miss* direction is indeed always red — I confirmed
`[ "${#x}" -gt 0 ] && source "$HERE/test-hashvar.sh"` is refused by the `[^#]*` class, and a shrunk
`SOURCED` reds via direction 2 (`tests/inventory.sh:158-179`). But the widened anchor also has a
*false-match* direction the old one did not, and that one is green.

Measured: `false && source "$HERE/test-nevr.sh"` **matches** the new pattern and **did not** match
`^[[:space:]]*source ...`. A phantom name that resolves to a file which exists on disk satisfies
direction 1 (the file exists) and satisfies direction 2 (the name is in `SOURCED`) — both loops pass
while the domain never runs. `run.sh`'s own guard does not catch it either: the `source` never
executes, so `domain_ran` never executes.

This is bounded and not live: the file's own comment at `tests/inventory.sh:138-141` explicitly
disclaims the "source inside a branch that never runs" class, and the old anchor only rejected one
*spelling* of it (the single-line `cond &&` form) by accident, never the multi-line `if ... fi`
form. So this is a small, disclosed narrowing — not a regression in the class covered. **The comment
prose is what overclaims, not the code.** Recommend softening "never a green" to "every way this
pattern can *stop matching* is a red"; the class it cannot see is already named two lines above.

---

## The four questions, answered

### (1) Does the guard fire for a domain that runs zero assertions? — **Yes, on every shape I could build.**

Re-derived by rebuilding the exact chain and `domain_ran` body against fixtures, not by reading the
battery table:

| fixture | source rc | outcome |
|---|---|---|
| healthy, 2 assertions | 0 | GREEN, no `#1400` line |
| top-level `return 0`, zero assertions (row R) | 0 | **RED** — "source returned 0" |
| missing file (row M) | 1 | **RED** — "source returned 1" |
| loads clean, zero assertions, last cmd `rm -f <absent>` | 0 | **RED** — "source returned 0" |

The last row matters: it is the shape the declined `|| domain_fail` form reads as healthy *and* the
shape the shipped diagnostic's hint mislabels ("zero means a top-level return") — the author
disclosed the hint imprecision in the body and I confirmed it fires exactly as described.

**Shapes that contribute zero assertions yet pass — two, both real:**

- **(a) One assertion, then a top-level `return`.** Measured: `ok "one"; return 0; ok "never"` passes
  green with everything after line 1 skipped. This is the disclosed "at least one" limit, but
  sharpened in a way the PR body does not say: the guard's *headline* win is row R, a top-level
  `return` that leaves no diagnostic anywhere — and that win holds **only while the `return` precedes
  the first assertion.** Move it one line down and the guard is blind again, silently, with no rc and
  no log line. Worth stating in the body since row R is what the design argument rests on.
- **(b) A domain file that asserts nothing itself but `source`s a sibling that does.** Delta > 0,
  green. Partly netted by `tests/inventory.sh:158-179` direction 2 — the sibling would read as
  on-disk-but-unsourced — unless it is added to `UNSOURCED`.

Not a hole today: I grepped all 24 domain files at this head and **none** reassigns `_d0`, `PASS` or
`FAIL`, and **none** has a top-level `return` (confirming the PR's "measured rather than assumed",
by my own grep rather than by its sweep).

### (2) Overclaiming beyond "at least one assertion ran"? — **Yes, in two places.**

`tests/stack/lib.sh:43-45` states the limit correctly and matches the PR body's **Limits** section;
those are clean. The two overclaims are **F1** (the `return 0` rationale — not about the limit, but a
false guarantee stated as measured fact) and **F2** (`repo-map.md:16` asserting the universal). The
diagnostic hint at `lib.sh:54` is imprecise for the clean-but-empty case, which the author disclosed
and I reproduced — counted as disclosed, not as a new finding.

### (3) Is `$?` clobbered by the arithmetic or by anything between? — **No. Clean.**

Re-derived by probe, not accepted:

- `_d0=$((PASS + FAIL))` returns **0** when the sum is 0 (it is an assignment; the status is not the
  value of the expansion). The arithmetic *command* `(( PASS + FAIL ))` returns **1** on a zero
  result. The PR body's warning about this is correct and the shipped form is the safe one.
- Nothing executes between the `source` and the `||` branch's word expansion, so `$?` there is
  `source`'s own status. Probed: **7** for a top-level `return 7`, **1** for a failing last command,
  **1** for a missing file, **2** for a parse error. All four arrive intact at `domain_ran`'s `$3`.
- In the `&&` branch `$?` is **0 by construction** — the `&&` demanded it — so the success-branch
  diagnostic always reads "source returned 0". Correct, if uninformative; that is the row-R message.
- `"$_d0"` and `"$?"` are expanded in the same pass, so neither can disturb the other. No domain file
  reassigns `_d0` (grepped, above), so the value cannot be stale across a stanza.

### (4) Does the widened anchor match anything it must not, and does it still reject the negatives? — **No, and yes.**

Negative controls, all four **refused** by `^[^#]*source "\$HERE/...`:
`# source "$HERE/x.sh"` · `    # source "$HERE/x.sh"` (indented) ·
`#_d0=$((PASS + FAIL)) && source "$HERE/x.sh" && ...` · the indented variant of that.
The `[^#]` class cannot cross a `#` and the match is `^`-anchored, so backtracking cannot rescue a
comment. The stated load-bearing half survives intact.

On today's `tests/stack/run.sh` the pattern extracts **exactly 25 names, 25 unique**, all live
stanzas (24 guarded + `lib.sh` bare at `run.sh:11`). Direction 1 (every named file exists) and
direction 2 (every on-disk domain file is named, modulo `UNSOURCED`) both come back clean when
simulated against the head tree. The old anchor extracts **1 of 25** — the CI red the author
diagnosed is confirmed by my own extraction, not taken from the comment.

New-match surface: `cond && source "$HERE/x.sh"` on one line, which the old anchor rejected — that is
**F3**, bounded and disclosed. New-miss surface: a live stanza whose prefix contains a `#` (e.g.
`${#x}`) becomes invisible — confirmed, and it fails safe (direction 2 reds).

---

## Standing reviewer checks

- **No wallet leak.** `tests/stack/lib.sh`'s `WALLET`/`VALID_PRIMARY` constants are untouched — the
  diff on that file is the added `domain_ran` block and nothing else.
- **No topology leak** in the PR body: no hostnames, IPs, specs or home paths.
- **`(#1400)` in a `bad` message is not an operator-strings violation.**
  `scripts/lint-operator-strings.sh` walks `pithead` and the dashboard frontend only, matching
  `log|warn|error|info|echo|dr_*|_upg_*`; `bad` in `tests/stack/` is out of both scopes.
- **Budget claims hold.** `run.sh` is 3841 lines against the `docs/dev/file-budget.tsv:55` row of
  3841; `lib.sh` is 259 against `TARGET_LINES=400`, so no budget row is owed. Both counted, not read
  off the body.
- **The body says "3 commits behind `develop-v2`"; it is now 8 behind / 2 ahead.** Stale number, but
  the load-bearing half of that claim survives — I checked all 8: they touch `dashboard/`,
  `os/rootfs/Dockerfile`, `scripts/pin-watch.sh` and `docs/`, and **none** touches `tests/stack/` or
  `docs/dev/file-budget.tsv`. So the after-proof still describes this code and the 3841 ceiling has
  not moved on the base. Worth a one-line refresh in the body rather than a re-run.
- **The `st="${3:-0}"` disclosure is confirmed and I agree with not fixing it.** All **48** call
  sites (24 stanzas x 2 branches) pass exactly three arguments — verified mechanically, every one
  matching `domain_ran <file> "$_d0" "$?"`. The default is unreachable. Changing it now would break
  byte-identity with the trees the mutation battery ran on, for a dead branch. Leave it; the
  disclosure is the right disposition.
- **The controller's pointer is satisfied.** The PR body *does* single out `return-top` as the
  decisive row and says why (source rc 0 is what the declined form reads as healthy). No question to
  press there.

---

## Re-derived from source vs. accepted from the PR body

**Re-derived (my own commands, at `39a8b0a0`):**

- 25 `source "$HERE/...` stanzas: `run.sh:11` bare (`lib.sh`), 24 guarded at `:15 :19 :44 :47 :67
  :70 :73 :119 :122 :156 :372 :497 :501 :505 :549 :1213 :1557 :1743 :2525 :2528 :2917 :2920 :3669
  :3832`. `test-lifecycle.sh` is the last, at `run.sh:3832`, immediately above the verdict.
- 24 `disable=SC2015` directives, each on the line immediately above its own stanza, each naming the
  same filename as that stanza's `source` and both `domain_ran` calls. Zero mismatches.
- 48 three-argument `domain_ran` call sites; no two-argument call anywhere.
- Guard firing behaviour on six fixtures (healthy / top-level `return` / missing / clean-but-empty /
  assert-then-`return` / assert-then-syntax-error), rebuilt from the shipped `domain_ran` body.
- The `return 0` counterfactual, both halves (F1) — rc 0 on both paths without it; `FAIL=1` not 2
  with an explicit `return 1`.
- `$?` propagation: rc 7 / 1 / 1 / 2 through the `||` branch; assignment-vs-`(( ))` status.
- Inventory: 25/25 extraction, 25 unique, direction 1 and direction 2 both clean, four negative
  controls refused, old anchor 1/25, and the two new match/miss classes.
- No domain file reassigns `_d0`/`PASS`/`FAIL`; no domain file has a top-level `return`.
- Line counts (`run.sh` 3841, `lib.sh` 259), budget row 3841, `TARGET_LINES=400`.
- Branch position (8 behind / 2 ahead) and the files those 8 commits touch.
- `scripts/lint-operator-strings.sh` scope; `WALLET` untouched.

**Accepted from the PR body, not re-derived** (COMMON forbids running the suite, `make lint` or
`shellcheck` here, and the prompt reiterated it):

- 2922 passed / 0 failed, rc 0 on both the after-proof and the baseline; 3233 log lines; 272 section
  headers; same set and same order.
- Every row of the mutation battery and its rc / verdict / source status / named kill, including the
  `syntax-bottom` log-line evidence and the `2922 - 48 = 2874` reconciliation.
- The instrumented per-domain assertion deltas (24 domains, 9 to 207) and the "all 24 return `source`
  rc 0" measurement behind the declined-form analysis.
- `make lint-sh` rc 0 and its SC2034 negative control; the SC1091 control proving `disable=SC2015`
  is narrow; `lint-file-budget` / `lint-topology` / `lint-operator-strings` / `lint-docs-voice` /
  `lint-js` rc 0.
- The `shfmt` line-splitting measurements (`;` chain -> 3889) that force the `&&`/`||` form.
- `bash tests/inventory.sh` exiting 0 at this head — though I simulated both of its loops by hand
  against the head tree and both come back clean, which is the same prediction by a different route.

**Disposition: merge-ready once F1's comment is corrected.** F2 is a four-word doc fix that can ride
the same commit. F3 is prose in a PR comment and needs nothing on disk. None of the three touches
executable behaviour, so **no re-run of the mutation battery or the after-proof is owed** — and the
author's own falsifiable-prediction framing (`tests/stack/` byte-identical) holds for a comment-only
follow-up too.

VijitSingh97:
## Review findings addressed — and a correction to my own earlier comment

The adversarial pass returned three findings, all of them *stated rationale* rather than defective
behaviour. F1 and F2 are fixed in `e827640`; F3 is a correction to prose I published here on
2026-08-28 and is below.

Nothing executable changed. Every changed line under `tests/stack/` is a comment line — checked
mechanically, not asserted — so the after-proof and the mutation battery still describe this code
and neither is re-run.

### F3 — correction: "never a green" was stronger than the pattern supports

My earlier comment on this PR claimed:

> Every failure mode of this pattern is a red, never a green, which is the direction that matters
> for a guard.

**That is wrong as written.** The *miss* direction is indeed always a red: a live stanza the
pattern stops matching shrinks `SOURCED`, and direction 2 reds on the on-disk file nothing claims
to source. But widening the anchor from `^[[:space:]]*source` to `^[^#]*source` also opened a
*false-match* direction the old pattern did not have, and that one is green. Re-derived here rather
than carried over from the review:

```
line                                     NEW ^[^#]*source    OLD ^[[:space:]]*source
false && source "$HERE/test-nevr.sh"     matches             does NOT match
    source "$HERE/test-real.sh"          matches             matches
# source "$HERE/test-cmt.sh"             rejected            rejected
```

So a phantom name that happens to resolve to a file which exists on disk satisfies direction 1
(the file exists) *and* direction 2 (the name is in `SOURCED`), while the domain never runs — and
`run.sh`'s own guard cannot catch it either, because the `source` never executes, so `domain_ran`
never executes.

This is bounded, and it was already disclosed on disk: `tests/inventory.sh:138-141` names the
"a `source` inside a branch that never runs, or inside a function nothing calls" class as out of
reach, needing control flow rather than grep. The old anchor rejected exactly one *spelling* of
that class — the single-line `cond &&` form — by accident, and never the multi-line `if … fi`
form, so the class actually covered has not regressed. It is a small, disclosed narrowing.

The accurate claim, which is what I should have written: **every way this pattern can *stop
matching* is a red.** Not "never a green".

The overclaim is in the comment prose only — the phrase is not on disk anywhere, so this
correction is the entire fix and no file changed for F3.

### F1 — the `return 0` justification was false on both halves (fixed in `e827640`)

The comment claimed a non-zero return would run the second branch and count the same failure
twice. Both halves are false, and I measured them rather than taking the finding on trust:

- **`domain_ran` cannot return non-zero as written.** Strip `return 0` and both paths still return
  0 — the healthy path is a false `if` with no `else`, and the fire path ends on `bad`'s `printf`.
- **Even given a body that explicitly `return 1`s, nothing double-counts.** The chain counts
  `FAIL=1`, not 2, because `bad` has already moved the very counter the guard tests, so the second
  call finds `$((PASS + FAIL))` past `before` and does nothing.

**The negative control is what makes that second row evidence rather than a green that proves
nothing:** with a `bad` that does *not* move the tested counter, the same chain fires the guard
**twice**. The harness can observe a double count; it simply does not see one here.

So the true property is idempotence, and that is what now stands in the comment. It is worth the
commit because the wrong reason sat three lines above the only paragraph on disk that stops a
"simplification" back to the `|| domain_fail` form that was declined on correctness — and a wrong
justification for a correct line is how a correct line gets deleted.

### F2 — `docs/dev/repo-map.md` asserted the universal the code disclaims (fixed in `e827640`)

It read "a file that fails to load contributes no assertions". That is false for the row this PR
itself demonstrates: `syntax-bottom`, a domain file whose parse error sits *after* its assertions,
contributes 48 assertions and passes green with a verdict identical to the control's.
`tests/stack/lib.sh:43-45` and this body's Limits section both state the limit correctly;
`repo-map.md` was the one artifact that dropped it. Four words restore it: "a file that fails to
load **before its first assertion** contributes no assertions". `docs/` is in `_shared.paths`, so
it needed no lane-override.

### Two refreshes to the body, and one open call left as the reviewer left it

- The "3 commits behind" line was stale; it is **8 behind / 3 ahead**, re-derived from the remote
  rather than from a cached ref. The load-bearing half survives: none of those 8 touches
  `tests/stack/` or `docs/dev/file-budget.tsv`, so the after-proof still describes this code, and
  the 3841 ceiling read from the base's own tsv blob has not moved.
- The body now states the **boundary on row R**: a top-level `return` is caught only while it
  precedes the first assertion. Not a hole at this head, but the design argument rests on that row,
  so the edge belongs next to it.
- The `st="${3:-0}"` default stays unreachable and stays as it is. All 48 call sites pass three
  arguments; changing it would break byte-identity with the trees the battery ran on, for a dead
  branch. The disclosure is the right disposition and the reviewer agrees.

VijitSingh97:
Merging with two red checks, deliberately. Recording the reasoning so it can be overruled.

**The two failures are `Build image (dashboard)` and `Build image (tor)`, both at their Trivy step.
Neither is this diff.** Read from the job logs rather than inferred — one CVE hits two unrelated
base images at once:

| image | base | packages | installed | fixed |
|---|---|---|---|---|
| `pithead-dashboard:ci` | debian 13 | `libssl3t64`, `openssl`, `openssl-provider-legacy` | 3.5.6-1~deb13u2 | 3.5.7-1~deb13u2 |
| `pithead-tor:ci` | alpine 3.24.1 | `libcrypto3`, `libssl3` | 3.5.7-r0 | 3.5.8-r0 |

CVE-2026-14456, HIGH — openssl DoS via unbounded memory growth in the QUIC server. The `docker build`
steps themselves succeed; only the scan fails. Filed and owned as #1427; the remedy is a pinned
base-digest bump in two Dockerfiles, which is outside this branch's reach.

**Why I am confident it is not this branch**, in the order the evidence actually settles it:

- This diff is comment-only under `tests/stack/` plus one line of `docs/dev/repo-map.md`. It cannot
  reach an image build.
- PR #1421 — a different lane, a different branch, a diff sharing nothing with this one — fails the
  same two jobs on the same CVE at its own head.
- This branch's previous head passed all five Build-image jobs.

**The check that actually gates this change is green:** `Shell tests (shellcheck + pithead suite)`
passed in 11m1s, and it is the only job that exercises `tests/stack/`. Every other job on the run is
green.

One caveat stated rather than left implicit: a days-old green on some other PR is not a control here.
The Trivy database cache key is day-granular and per-ref, so PRs whose Build-image jobs last ran on
2026-08-24/25 would pass today's scan for reasons that have nothing to do with their diffs. The
discriminating comparison is #1421, whose run started today.

VijitSingh97:
**Correcting one line above, because it is a remedy claim and those get acted on.**

I wrote that the fix is "a pinned base-digest bump in two Dockerfiles". **That is right for the
dashboard half and wrong for tor.** Upstream has not rebuilt `alpine:3.24`, so the pinned digest
already equals the current tag and there is nothing to bump to. Tor's remaining routes are to wait
for the upstream rebuild, or to add an explicit package upgrade to the Dockerfile — which trades
away the reproducibility the digest pin exists for, and is a design decision rather than a
mechanical fix.

**#1427 is the authoritative version**, and it carries a caveat I did not have: re-run the digest
check before planning around this, because an upstream rebuild collapses the tor half back to a
one-line bump.

Nothing here changes why this PR merged — that rested on the diff being unable to affect image
contents, which is unaffected. Correcting it because someone reading this comment on the way to
#1427 could otherwise spend an hour bumping a digest that is already current.

(Correction owed to the controller, who derived the upstream-digest fact, and to the e2e lane, who
argued me out of leaving it: a claim that decided nothing here is still one a future reader will act
on, and that is the test rather than whether it was load-bearing where it sits.)

Issue #1412, opened 2026-08-24
RigForge ships wire-contract fixtures for us and nothing here reads them — our contract test is anchored to our own fake


VijitSingh97:
## What

RigForge ships `tests/contract/v1/` — `feed.json` and `control-status.json` — plain-JSON fixtures
that pin the two wire shapes our fakes consume from a worker. RigForge regenerates them from its
**real** code paths on every one of its own test runs and byte-compares the result, so on their side
the fixtures are guaranteed to reflect what a worker actually emits.

This repo references them nowhere. No vendoring, no fetch, no byte-compare, no drift guard.

## Why this is worth an issue rather than a note

We do have a contract test: `tests/integration/fakes/test_contract.py` points the real
`XMRigWorkerClient` / `parse_rigforge` at a fake worker API over a real socket, and it is a good
test. But the thing it asserts against is `tests/integration/fakes/fake_worker_api.py` — **our own
hand-written fake**. Nothing ties that fake to RigForge's actual output.

So the guard exists in form and is anchored to the wrong artifact. Our fake and a real worker can
drift apart while every test here stays green — which is a worse shape than having no guard, because
it reads as coverage.

RigForge's own fixture README names the confirmed precedent: our poller did not know about the
`noop` / `throttled` statuses until someone hit them on a live stack. Their words: *"Nothing outside
this repo enforced these before — drift was only ever caught by a manual live run against a real
stack."*

## This is NOT live drift today — checked, and the distinction matters

I compared the pinned status vocabulary against ours and **they agree**:

| | statuses |
|---|---|
| RigForge `control-status.json` | `applied`, `failed`, `noop`, `pending`, `rejected`, `rolled_back`, `started`, `throttled` |
| here | `_CONTROL_TERMINAL` (`client/xmrig_client.py:168`) and `_RECONCILE_TERMINAL` (`service/storage_service.py:58`) both carry all six terminal ones; `pending` / `started` are handled as non-terminal |

So the historical hole is closed. **It was closed by someone hitting it live, not by a check, and
nothing preserves that agreement.** That is the gap being filed: a missing mechanism, not a current
bug.

## What I did not check

Only the control-status vocabulary was compared. I did **not** verify that `feed.json`'s shape agrees
with what `parse_rigforge` expects, and I did not diff our fake's response bodies against either
fixture field by field. Those are the obvious next checks and either could turn this from a missing
mechanism into a live finding.

## Suggested shape

[...]

PR #1421, opened 2026-08-24
test(e2e): anchor the worker contract test to RigForge's fixtures, not to our own fake (#1412)

VijitSingh97:
## What this fixes

`tests/integration/fakes/test_contract.py` runs the **real** `XMRigWorkerClient` and `parse_rigforge`
against `fake_worker_api.py` over a real socket, and its own comment says *"a drift in RigForge's auth
handshake or enriched-feed shape goes red here, not only on a live rig."*

**For three of the ten sub-blocks a real worker emits, it could not.** `config`, `config_meta` and
`control` were absent from the fake entirely — so there was no wire shape to drift *from*, and nothing
tied the fake to what a rig actually sends. Those three carry the Worker Inspect config prefill
(#1235), the config-provenance line (#1345) and the control-outcome mirror (#579).

Read from the **producing code** at the ref the appliance bakes, not from the fixture: `_api_rigforge_block`
builds all three unconditionally on every serve (`rigforge.sh:4997`), and `control` is served as a
literal `null` on a rig that has never had a control change rather than being omitted.

## What it cost, measured before the change

Each **real consumer** mutated to a plainly-wrong behaviour. Failing test **names** captured, not exit
codes; mutations anchored on code rather than comment text, proven applied by `cmp`; post-restore
control returns to baseline.

| mutation | tier 2 (`test_contract.py`) | tier 1 |
|---|---|---|
| `parse_config_meta` always returns `None` | **rc 0 — survives** | rc 1, 7 named kills |
| writable-key allowlist dropped | **rc 0 — survives** | rc 1, 1 named kill |
| every control status read as terminal | **rc 0 — survives** | rc 1, 2 named kills |
| pool credential strip removed | **rc 0 — survives** | rc 1, 3 named kills |

Four regressions — including a credential reaching an editor that can POST it back — and the contract
test green through every one. Tier 1 catches all four, but tier 1 hands the parser a payload *we* wrote,
so it confirms our beliefs about the wire rather than the wire itself.

## The change

RigForge regenerates `tests/contract/v1/` from its real code paths and byte-compares it on every one of
its runs, so those fixtures are the producer's own account of the wire. This vendors them and consumes
them, rather than restating them:

- **`contract/v1/`** — `feed.json`, `control-status.json` byte-identical to RigForge at the baked ref,
  plus a `PROVENANCE` file recording that ref.
- **The three blocks added to the fake**, with realistic values, plus every `health`/`watchdog` key the
  producer emits. `msr`/`smt` become strings because that is what it sends.
- **A shape guard, in BOTH directions** — a missing key is an uncovered surface (this bug), an extra key
  is our fake inventing a shape no rig sends. Over the healthy body *and* the miner-unreachable one,
  which drops every XMRig key and serves the whole block plus one flag.
- **A freshness guard** tying the vendored copy to the baked `RIGFORGE_REF`. A vendored snapshot with no
  freshness mechanism is just a slower drift.
- **A vocabulary guard**, fail-closed: every status word in `control-status.json` must be classified
  terminal or known-non-terminal. A new word fails rather than falling through `not in _CONTROL_TERMINAL`
  and reading as still-in-flight forever — which is exactly how `noop`/`throttled` were missed until
  someone hit them on a live stack.
- **Tier-2 legs** driving config, config_meta and control through the real consumers, including the
  fresh-rig `control: null` shape and a non-terminal status.

The fixture pins the **shape**, the fake pins realistic **values**, and the guards compare key paths and
never values — the fixture is a normalized, hardware-independent capture whose numeric fields are mostly
null, so driving the fake from it directly would exercise the consumers' default-everything path and
nothing else.

## After — and two rows that deliberately did NOT move

| mutation | tier 2 | tier 1 |
|---|---|---|
| `parse_config_meta` always returns `None` | **rc 1**, 2 named kills | rc 1, 7 named kills |
| writable-key allowlist dropped | rc 0 | rc 1, 1 named kill |
| every control status read as terminal | **rc 1**, 1 named kill | rc 1, 2 named kills |
| pool credential strip removed | rc 0 | rc 1, 3 named kills |

**The two rc 0s are correct, not a shortfall.** A contract-conformant body contains only writable keys
and cannot carry a credential — the producer strips `pass`/`tls-fingerprint` at source, and our
`_POOL_CREDENTIAL_KEYS` is exactly those two. Both regressions need a deliberately hostile body, which
is tier 1's job and is already covered there. Making tier 2 catch them would test one behaviour at two
tiers.

**Every new guard was required to go red**, each with a named kill:

| mutation aimed at a new guard | result |
|---|---|
| a producer key dropped from the fake | rc 1 — `test_fake_matches_the_vendored_wire_contract` |
| the fake invents a key no rig sends | rc 1 — `test_fake_matches_the_vendored_wire_contract` |
| vendored fixtures stale vs the baked pin | rc 1 — `test_vendored_contract_matches_the_baked_rigforge_pin` |
| RigForge adds a status word we never classified | rc 1 — `test_every_pinned_control_status_is_classified` |

The first of those also kills `test_worker_config_provenance_parses_over_the_wire` — a legitimate
collateral, since removing `last_change_id` genuinely breaks that assertion too. The test the mutation
was **aimed at** is in the named set either way.

## Over-engineering pass

Run by hand and recorded here, because the repo's gate for this keys on the pane's branch rather than
the PR's and its silence is not evidence either way.

**It found a real defect in my own diff and I removed it.** My first version also asserted
`set(_CONTROL_TERMINAL) == set(_RECONCILE_TERMINAL)`, on the strength of the source comment saying *"one
vocabulary, three places"*. `dashboard/tests/service/test_storage_service.py:1646` **already asserts
exactly that**, at tier 1. Verified rather than assumed: with my line removed, that mutation leaves tier 2
at rc 0 and dies at tier 1 with two named kills, one of them `test_reconcile_terminal_matches_the_parser_verbatim`
— a test that exists for precisely this. Pure duplication at a higher tier; removed, and the now-unused
import with it.

## What this does NOT do

- **The third place in that vocabulary is unguarded and this does not fix it.** The `pithead` script has
  its own status lists (`pithead:10874` handles 4, `:11022` handles 5); nothing compares them to the
  dashboard's set or to RigForge's. Out of this lane's paths — filing it separately rather than reaching in.
- **`make lint-sh` and `make test` were not run.** Nothing here is shell, and the lint lock and the
  shellcheck memory path are shared. The seven gates below were run individually, never through a filtered
  `make` chain.
- **Nothing here is live-rig evidence.** It is a tier-2 change and proves only what tier 2 can prove: the
  fake now matches the producer's published contract. Whether that contract matches the rig on the bench
  is #1236's question, not this one.
- **The vocabulary guard's known-non-terminal set (`pending`, `started`) is a list in the test.** There is
  no production list of non-terminal RigForge statuses to import — the parser treats "not terminal" as a
  fall-through. Listing them is what makes a new word an error rather than a default, but it is a restatement,
  and if a production home for it appears this should import it instead.

## Gates

Read individually, exit codes checked, never through a filtered chain:
`test-fakes` **30 passed / rc 0** · `lint-file-budget` · `lint-py` · `lint-topology` ·
`lint-operator-strings` · `lint-docs-voice` · `lint-md` — **all rc 0**.

`tests/integration/fakes/test_contract.py` crosses the 400-line target and takes a row at **451**. Only
that row was added. `--generate` also reports lowering `.github/workflows/ci.yml` 513 → 512 and
`tests/integration/run.sh` 2551 → 2533 — the two rows COMMON records as the only ones with headroom —
and those are deliberately left alone: adopting a lowered ceiling reds every un-rebased branch on the box
for a file its diff never touched. The generate-diff was checked after staging and shows those two rows
and nothing else, so no row was silently dropped.

**Shared paths touched, disclosed:** `docs/dev/testing-strategy.md` and `docs/dev/repo-map.md` (both
`_shared.paths`), and `docs/dev/file-budget.tsv` (regenerated, one row added).

## Issue

Evidence for #1412. No closing keyword — this repo's PRs land on a non-default branch, so they never fire;
closing is a separate by-hand decision. The measurements above are posted on the issue as well, including
the correction to its framing: it was filed as a missing mechanism, and check 2 makes it a live coverage
hole. Check 1 in that issue — whether `feed.json` agrees with `parse_rigforge` — came back **clean**, and
that half is recorded there rather than here.

---

## Review response — both findings addressed at `6b8bd46`

An adversarial non-author review returned two findings. Neither is structural; both are now
answered, and the second is answered by disclosure rather than by a fix.

### F2 (LOW) — the vocabulary guard passed vacuously on an empty fixture. FIXED.

`assert pinned - set(_CONTROL_TERMINAL) - non_terminal == set()` is a set difference, and a
difference against an empty left-hand side is empty. A fixture whose `statuses` arrived empty — or
renamed with a compatibility stub — passed the guard having checked zero words, at exactly the
moment F1 below already lets a stale fixture through. The floor is now asserted. It is line-neutral,
folded into the existing assertion and its comment, so the 451 row does not move.

Mutation, with failing test NAMES captured, `cmp` proving each mutation applied, and a post-restore
control:

| mutation | rc | named kill |
|---|---|---|
| M1 — fixture `statuses` → `[]` | 1 | `test_every_pinned_control_status_is_classified` |
| M1b — same mutation, PRE-REVIEW form of the guard restored | **0** | none — **the defect, demonstrated** |
| M2 — `PROVENANCE` `ref=` → a different sha | 1 | `test_vendored_contract_matches_the_baked_rigforge_pin` |
| post-restore control | 0 | 30 passed, baseline |

M1b is the row that earns the change: it shows the same mutation is invisible to the form this
replaces. M2 is a control on the other guard — the F1 wording edit below is text only, and M2 is
what proves that guard still fires.

### F1 (MEDIUM) — the freshness guard proves a string, not a re-vendoring. FILED as #1426, disclosed here.

`test_vendored_contract_matches_the_baked_rigforge_pin` compares `PROVENANCE`'s `ref=` to
`ARG RIGFORGE_REF=` in `os/rootfs/Dockerfile:207`. Both are hand-edited text.

**Stating which half of its own failure message that guard verifies, because it was verifying one of
two and reading as if it verified both:** the message instructs *"re-copy the fixtures"* **and**
*"update PROVENANCE"*, and **only the `PROVENANCE` half is checked.** A maintainer who bumps the ref,
edits the one obvious line and re-runs gets green over a stale `feed.json` — and then
`test_fake_matches_the_vendored_wire_contract` compares the fake against that stale fixture and
passes too.

A real mechanism fetches `tests/contract/v1/{feed,control-status}.json` from RigForge at
`$RIGFORGE_REF` and byte-compares; the appliance build already pulls that exact tarball at
`os/rootfs/Dockerfile:208-212`. That is new network plumbing in a suite that runs offline today
(`.github/workflows/ci.yml:288` and `Makefile:28` are the only RigForge-touching CI, and neither
fetches), so it is **#1426**, filed before this merges rather than after.

What lands here instead is honesty in the three places that overclaimed — the docstring, the failure
message, and the tier-2 row in `docs/dev/testing-strategy.md`. All three now say the guard reads the
ref line and never the bytes, and all three name #1426. The pin is a full sha, so content-at-ref
cannot drift underneath us; the drift surface is exactly that one hand-edit.

### Gates re-run at `6b8bd46`

`test-fakes` **30 passed / rc 0** · `lint-py` rc 0 · `lint-file-budget` rc 0 · `lint-docs-voice` rc 0.
Both edited files are **line-neutral** (451 and 407, unchanged), so no budget row moves and no
`--generate` diff is owed. `make lint-sh` and `make test` still not run — nothing in this change is
shell, and the lint lock and the shellcheck memory path are shared.


VijitSingh97:
The third-place gap this PR's "What this does NOT do" section names is now filed as #1422 — the `pithead` script's two poll loops are the unguarded leg of the "one vocabulary, three places" invariant. Not a drift today: both `case` arms are correct subsets for their own path and their union equals the dashboard's set. What is missing is anything that preserves that, and an unmatched `case` falls through to "keep polling" and is recorded `accepted` at the deadline — the same shape as the `noop`/`throttled` history. Closing only the dashboard half makes the vocabulary look guarded, which is why it is filed rather than left as a note.

VijitSingh97:
**Controller — routing an ephemeral non-author adversarial pass.**

A `reviewer` pass ran against this PR at `0e8d1b8e`, which is still the head as of this comment, in the same pass that covered #1415. Read-and-reason only: no `make`, no suite, no `shellcheck`, no bench, no rig. The #1421 section is reproduced below unedited, followed by the reviewer's re-derived-vs-accepted split — that last section covers both PRs, so read past the #1415 rows.

**Disposition: not merged yet, two findings, neither of them the shape that blocks.**

Criterion (a) **passes**, and the reviewer singles that out as the material difference between this PR and #1415: the positive control here is committed and mechanical rather than PR-body prose, and it was checked for circularity — the fake builds its block as a hardcoded literal and never reads `contract/v1/`, so two independent artifacts are genuinely being compared.

F1 (MEDIUM) is that the freshness guard compares two hand-edited strings and so proves a `ref=`, not a re-vendoring — which is the drift surface this PR exists to close. The reviewer's own read is that a real fetch-and-byte-compare is new network plumbing in a suite that runs offline today, and is therefore a reasonable filed follow-up rather than a merge blocker. I agree, with one condition: **file it before merging, not after**, and say in the PR body which half of the failure message the guard actually verifies. F2 (LOW) is a one-line floor assertion.

Routed to the `e2e` lane, which owns `tests/integration/`.

---

## PR 1421 — `test(e2e): anchor the worker contract test to RigForge's fixtures (#1412)`

### VERDICT: FINDINGS

### Criterion (a) — PASSES, and this is the material difference between the two PRs

`tests/integration/fakes/test_contract.py:348-362`. The positive control is **committed and mechanical**,
not a PR-body claim. `test_fake_matches_the_vendored_wire_contract` asserts set equality of leaf key
paths in both directions; deleting `config`/`config_meta`/`control` from the fake — i.e. restoring the
exact bug #1412 describes — makes `:356` fail on set inequality. Answerable by inspection.

I checked the one thing that would make it vacuous: **it is not circular.** `fake_worker_api.py` builds
`_RIGFORGE_BLOCK` as a hardcoded literal and never reads `contract/v1/` (grep for `read_text|json.load|
open(|pathlib` in the fake returns only comment lines). Two independent artifacts are being compared.

### F1 — the freshness guard proves a string, not a re-vendoring (MEDIUM) — criteria (c) + (e)

`tests/integration/fakes/test_contract.py:373-381`

`test_vendored_contract_matches_the_baked_rigforge_pin` compares `PROVENANCE`'s `ref=` to
`os/rootfs/Dockerfile:207`'s `ARG RIGFORGE_REF=`. Both are hand-edited text. **Nothing ties the fixture
*content* to that ref** — no digest, no fetch, no comparison against the producer. Worse, the guard's
own failure message (`:378-380`) instructs *"Re-copy … and update PROVENANCE"*, and the test verifies
only the second half of that instruction.

The pin is a full sha (`4ce29b3daf063fd1b45e050649e93aa9592618e1`, matching `PROVENANCE`), so the ref
itself is immutable and content-at-ref cannot drift underneath. That narrows the drift surface to
exactly one place — and that place is unguarded.

**Failure scenario.** A maintainer bumps `ARG RIGFORGE_REF` for an unrelated RigForge fix. This test
reds and names `PROVENANCE` in its message. They edit `PROVENANCE`'s one obvious line and re-run: green.
`feed.json` still describes the previous wire. `test_fake_matches_the_vendored_wire_contract` then
compares the fake against a **stale** fixture and passes, and #1412's thesis — that the fake cannot
drift from what a real worker emits — is silently false again with every guard green. That is a strictly
worse state than before this PR, because the guards now look like they cover it.

**So, to answer (c) directly:** the mechanism is *real* for "the pin moved and nobody looked"; it is
*hoped for* on "somebody looked and actually re-copied the bytes." A real mechanism would fetch
`tests/contract/v1/{feed,control-status}.json` from RigForge at `$RIGFORGE_REF` and byte-compare. The
appliance build already downloads that exact tarball at `Dockerfile:208-212`, so the URL shape is known.
Being fair about the cost: nothing in CI fetches RigForge today (`.github/workflows/ci.yml:288` and
`Makefile:28` only run `selftest-rigforge-apply-settle.sh`), so this is new network plumbing in a suite
that currently runs offline — reasonable as a filed follow-up rather than a merge blocker.

### F2 — the vocabulary guard passes vacuously on an empty fixture (LOW) — criterion (b)

`tests/integration/fakes/test_contract.py:393-398`

`assert pinned - set(_CONTROL_TERMINAL) - non_terminal == set()` is fail-closed in the direction that
matters — a new RigForge status word in neither set is an error, not a default — and it is
**non-vacuous today**, which I computed rather than assumed: the fixture's `statuses` holds 8 words
(`applied, failed, noop, pending, rejected, rolled_back, started, throttled`), `_CONTROL_TERMINAL` at
`dashboard/mining_dashboard/client/xmrig_client.py:168` holds 6, `non_terminal` holds 2, difference
empty. I also checked the fixture is internally consistent: `statuses` equals the key set of `examples`.

But a set difference against an empty left-hand side is `set()`. If `statuses` ever arrives empty or
renamed-with-a-compat-stub, the guard goes green having checked zero words — and by construction that
happens at exactly the moment F1 already lets a stale fixture through.

**Failure scenario.** A re-vendor picks up a RigForge fixture that moves the vocabulary to
`status_words` and leaves `"statuses": []` for compatibility. `test_every_pinned_control_status_is_classified`
passes checking nothing; a genuinely new terminal word falls through `not in _CONTROL_TERMINAL` at
`xmrig_client.py:195` and is read as "still in flight" forever — the exact `noop`/`throttled` shape the
docstring says it exists to prevent.

One-line fix: assert a floor (`assert pinned`, or `assert len(pinned) >= 8`) before the difference.

### On criteria (b) and (e) for the rest of the change — no findings

- **Shape guard is fail-closed, not a denylist.** `:356` and `:362` are set *equalities* with no
  exclusion list, covering both the healthy and the miner-unreachable body. `_key_paths` (`:334-345`)
  collapses list indices to `[]`, so ordering and length correctly do not count, and it yields
  `prefix + "[]"` for an empty list so an empty array is still a compared path.
- **Behaviour, not text**, for the three new wire legs (`:401-450`): each stands up `FakeWorkerApi` and
  drives the real `XMRigWorkerClient` / `parse_rigforge` / `parse_worker_control_status` over a real
  socket, asserting parsed values. `test_a_rig_mid_change_or_fresh_reconciles_nothing` (`:440`) covers
  both `control: null` and a non-terminal status and additionally asserts the rest of the block still
  parses — a guard against the null case being handled by bailing out early.
- The shape guard itself proves *our fake matches our copy* — behaviour-relevant, but only as strong as
  F1's single unverified link at the repo boundary.


---

## What I re-derived from source vs. accepted from the PR bodies

**Re-derived (read at `f2cb1a07` / `0e8d1b8e`):**

- 1415 F1's full reachability chain: `e2e.sh:606` → `:611-613` → `:634` → `:635-637`. Traced before
  reading the 2026-08-24 comment's argument.
- The claim underpinning the whole design — "a surviving `.e2e-orig.*` is an un-restored borrow" —
  against `restore_all`'s `cp && chmod && cmp -s && rm -f` in one remote call at `e2e.sh:215`, and the
  `--keep` early return that also leaves one, at `:196-200`.
- No committed coverage of `borrow_miner` (grep, repo-wide); `on_miner` is a one-liner at `e2e.sh:172`;
  `main` is called unconditionally at the file tail; `selftest.sh:1-8` states its own scope.
- Budget arithmetic: `e2e.sh` 763 actual vs 766 in `file-budget.tsv`; `selftest.sh` 686 vs 686.
- 1421: `_key_paths` semantics (`:334-345`); both-direction equality (`:356`, `:362`); the pin guard
  comparing strings only (`:373-381`) against `Dockerfile:207`; `PROVENANCE` contents; the vocabulary
  guard's non-vacuity today, computed from the fixture's 8 statuses and `xmrig_client.py:168`'s 6;
  `statuses` == `examples` key set; the fake's non-circularity.
- Both heads confirmed against `git ls-remote` (no local `fetch` — COMMON's rule about moving
  `origin/*` under other lanes' gates).

**Accepted, not re-derived:**

- **Every live-bench number in 1415**: 19/0, 19/0, the 14/5 pre-fix run, the byte-exact read-back, and
  the 39/0 → 30/8 local battery with identical failing-name sets. No bench, no rig, no ssh touched.
- **1415's cross-repo claim** that RigForge regenerates the xmrig config on every apply — F3 *is* the
  fact that I could not verify it.
- **1421's entire before/after mutation table** (four rc 0 → two rc 1 with named kills; each new guard
  mutated and required to red). Not re-run.
- **1421's cross-repo claims**: that `_api_rigforge_block` at `rigforge.sh:4997` builds all three blocks
  unconditionally and serves `control` as a literal `null`; and that RigForge regenerates and
  byte-compares `tests/contract/v1/` on every run. The second is load-bearing for the fixture being the
  producer's own account of the wire, and it is the same class of uncited cross-repo dependency as
  1415's F3 — 1421 does cite file:line for the first.
- Both PRs' gate results (`test-fakes` 30 passed, the individually-run lint targets, shellcheck 0.11.0
  / shfmt / `bash -n`). Not run here, per the brief and COMMON.
- 1421's over-engineering-pass claim that `dashboard/tests/service/test_storage_service.py:1646` already
  asserts the removed duplicate. Not opened.

VijitSingh97:
Both review findings are addressed at `6b8bd46`, and the PR body now carries the response in full.

- **F2 (LOW)** — FIXED, line-neutral. The vocabulary guard now asserts its floor. Mutation proves
  it: `statuses -> []` reds with a named kill, and the same mutation against the **pre-review** form
  of the guard is **green** — the defect demonstrated rather than argued. `PROVENANCE ref` -> a
  different sha is the control on the other guard; post-restore control back to 30 passed.
- **F1 (MEDIUM)** — filed as **#1426** before merging, per the condition. The PR body now states
  which half of the guard's own failure message it verifies: the message says *re-copy the fixtures*
  **and** *update PROVENANCE*, and only the `PROVENANCE` half is checked. The docstring, the failure
  message and the tier-2 row in `docs/dev/testing-strategy.md` all say so now and all name #1426.

**NOT merging yet, and the reason is not this PR.** `Build image (tor)` and `Build image (dashboard)`
are RED on **CVE-2026-14456** (openssl, HIGH) in their base images — alpine `libcrypto3`/`libssl3`
3.5.7-r0 and debian `libssl3t64`/`openssl` 3.5.6-1~deb13u2. Nothing in this diff touches a
Dockerfile, and the **same two jobs fail the same way on an unrelated branch**, so it is repo-wide.
Filed as **#1427**. Every other check is green (13 SUCCESS).

Merging over a red security gate is a fleet-wide call rather than this lane's, so it is left for the
controller: the CVE is upstream with a published fix in both distributions, and holding every merge
in the repo on it has its own cost.

VijitSingh97:
Merging over a known repo-wide red, disclosed rather than quietly waived.

`Build image (tor)` and `Build image (dashboard)` fail their Trivy step on CVE-2026-14456 (openssl, HIGH). This is not about this PR: the same two jobs fail identically on unrelated branches whose diffs share nothing with this one, and it arrived between 2026-08-25 and 2026-08-28. It is tracked as #1427 and is release-blocking, not merge-blocking, per the standing fleet ruling. The other 14 checks pass.

Two reasons a mute would have been the wrong move, recorded so nobody reaches for one later: the gate runs with `ignore-unfixed: true`, so a finding here always has a published fix and is never noise; and the tor half is not even a digest bump — upstream has not rebuilt `alpine:3.24`, so the pinned digest already equals the current tag and there is nothing to bump to. Resolving it is a real decision about trading away the reproducibility the pin exists for, which is why it is #1427's and not this PR's.

This PR's diff is one Python test file, three vendored fixtures and three doc lines; it cannot affect image contents.

Rebased onto `develop-v2` after #1416 landed — the two collided on adjacent rows of one table in `docs/dev/repo-map.md` (their tier-1 row, my tier-2 row), resolved by keeping both. Post-rebase I verified the budget row reads as expected and separately regenerated `file-budget.tsv` and diffed it, because a clean auto-merge of that file can drop rows at rc 0; the only delta this PR introduces is its own new row, and no base row was lost. `scripts/lint-file-budget.sh` returns rc 0 at the rebased head.

Issue #1420, opened 2026-08-24
`tests/inventory.sh`'s emptiness guard failed open on a live 24-of-25 breakage, and the file already documents the same hazard for its sibling


VijitSingh97:
## What happened

While landing the source-stanza guard (PR #1416), the drift pattern at `tests/inventory.sh:139` stopped matching 24 of the 25 `source "$HERE/..."` stanzas in `tests/stack/run.sh` — the PR moved `source` off the start of the line and the pattern was line-anchored. That much is the PR's own defect and is fixed there.

What is worth filing is what the file's own safety net did about it.

## 1. The emptiness guard did not fire, and it is the guard written for exactly this

`tests/inventory.sh:141` reads:

    if [ -z "$SOURCED" ]; then   # the grep itself going quiet must not read as "all present"

Its comment states its purpose precisely. It did not fire. `lib.sh`'s stanza is bare by design, so it kept matching, and `SOURCED` stayed non-empty at exactly **one** entry while the pattern had stopped seeing the other 24. The guard against the pattern going quiet was defeated by a single surviving member.

**A guard whose fallback condition is "the set is empty" is defeated by one survivor.** The check that would have caught this is a floor, not a non-empty test: the sourced count and the domain-file count on disk are both knowable at that point, and requiring them to agree turns a 24-file regression into a named failure instead of a silent one.

Note this is not a hypothetical failure mode discovered by reading. It happened, on a real PR, and the thing that eventually caught it was the direction-2 loop several checks later — reporting the wrong-sized problem (see 2).

## 2. The file already knows about this hazard, one guard over

The per-file zero check immediately below carries this comment:

    # What this does NOT catch: a file that shrinks without reaching zero. A suite going from 11
    # sections to 1 still passes here.

That is the same defect class, stated honestly, about a sibling guard. The `SOURCED` emptiness test has the identical hazard and its comment reads as though it covers the case. **The disclosed limit and the undisclosed one are eight lines apart**, which is the part worth fixing — a reader who checks whether this file thinks about zero-versus-floor finds the answer yes, and it is only true of one of the two.

The same question is open for the aggregate counter loop at `:76`: `n_py_dash n_py_fake n_node n_stack n_selftest n_scen n_axes n_mini` are each tested against zero, so any of them can fall from hundreds to one and pass. Whether that is acceptable for each counter is a per-counter judgement, not one ruling.

## 3. The drift loop names the first offender and exits

[...]

Issue #1427, opened 2026-08-28
The shipped-image CVE gate is RED on unrelated branches — one upstream openssl CVE, two base images


VijitSingh97:
`Build image (tor)` and `Build image (dashboard)` both fail their trivy step on **CVE-2026-14456**
(openssl, HIGH — DoS via unbounded memory growth in the QUIC server). Nothing in either failing
branch's diff touches a Dockerfile.

| image | base | package | installed | fixed |
|---|---|---|---|---|
| `pithead-tor:ci` | alpine 3.24.1 | `libcrypto3`, `libssl3` | 3.5.7-r0 | **3.5.8-r0** |
| `pithead-dashboard:ci` | debian 13.6 | `libssl3t64`, `openssl`, `openssl-provider-legacy` | 3.5.6-1~deb13u2 | **3.5.7-1~deb13u2** |

Totals: tor 2 HIGH, dashboard 3 HIGH. Both jobs end `##[error]Process completed with exit code 1`.

### It is not one branch's problem

Observed on **two independent branches** whose diffs share nothing — one a Python test file plus a
doc, the other a shell guard — failing the **same two jobs on the same CVE**. Per COMMON's rule for
a repo-wide gate: check whether another open branch is failing the same way before reading your own
diff. It is.

So this blocks merges across the board, and each lane that hits it will go hunting a diff that
cannot explain it.

### The fix is upstream and already published

A fixed package exists for both distributions, so this clears by rebuilding on a base that carries
it — a pinned-digest bump, not a code change. Both Dockerfiles are outside the integration-harness
lane's paths, so this is filed rather than fixed.

**Do not reach for a `.trivyignore` mute.** The fixed versions are out; a mute here is the dead-mute
shape this repo already files, over a HIGH with a remedy available.

Found while re-running CI on a test-only change.

VijitSingh97:
**Controller ruling: this is release-blocking, not merge-blocking. Labelled `release-blocker`.**

Two lanes hit this within an hour and did opposite things — one merged over it, one stopped and escalated. On substance both were right; only the second was right on process. So the ruling is written down rather than left to be re-guessed.

## The fact the decision turns on

[...]

PR #1428, opened 2026-08-28
fix(tests): inventory's emptiness guard failed open on a 24-of-25 breakage — make it a floor

VijitSingh97:
Addresses the two mechanical halves of #1420. One file, one region.

**#1420 stays open** — it carries `do-not-close` and the acceptance
assessment is its newest comment. The deferred third item is named at the bottom of this body.

## The defect

`tests/inventory.sh` guarded against its own `source` pattern going quiet with `[ -z "$SOURCED" ]`,
and the comment on that line stated exactly that purpose. It failed open on a real breakage: #1416
put a guard prefix on every domain stanza, the line-anchored pattern stopped matching **24 of the 25
stanzas**, and the guard stayed silent because `lib.sh`'s stanza is bare by design and kept matching.
`SOURCED` sat at one entry.

**A fallback condition of "the set is empty" is defeated by a single survivor.**

Direction 2 did catch it, several checks later, and sized it wrongly: it exits on its first offender,
so a 24-file regression was reported as one file — alphabetically first — and the reader sizes the
problem from that message.

## The change

- **The emptiness test becomes a floor.** The sourced set is compared against the domain files on
  disk, so one survivor is as loud as none.
- **Direction 2 collects.** Every offender is gathered and reported with the count and the list,
  instead of exiting on the first.
- The expected-files set is hoisted above both directions: the floor needs its size, both loops need
  its membership.
- Direction 1 gains a blank-line skip. A here-string over an empty `SOURCED` yields **one blank
  line, not zero**, and with the emptiness guard gone that case now reaches direction 1, where it
  would otherwise test `tests/stack/` and print a garbled name.

## What was run

Mutation battery — **9 scored rows in the harness, all passing**, plus F2 verified separately
with its own control (below) — in a scratch copy — never in a live worktree. Every row asserts three
things, not two: the mutation applied (`cmp`, and the run aborts with `exit 3` if it did not — M5
tripped exactly that and was fixed rather than scored), something died, and **what died is the guard
the mutation was aimed at**, matched on that guard's own distinct sentence.

    C  control, unmutated             rc=0                                    PASS
    M1 pattern re-anchored            rc=1  AIMED  24 of 25 domain files, 24 named
    M2 one stanza deleted             rc=1  AIMED   1 of 25 domain files,  1 named
    M3 two stanzas deleted            rc=1  AIMED   2 of 25 domain files,  2 named
    M4 domain file removed from disk  rc=1  AIMED  direction 1's own sentence, 0 listed
    M5 pattern matches nothing        rc=1  AIMED  25 of 25 domain files, 25 named
    M6 guard deleted + M1 re-applied  rc=0                                    PASS
    Z  post-restore control           rc=0                                    PASS

**M1 is the real regression** — the exact breakage the old guard failed open on — and **M6 is what
makes it evidence**: with the direction-2 block deleted, M1 goes green, so nothing else in the file
was catching it and M1 is not a bystander kill. **M3 is the collect-don't-exit half specifically**:
the old code would have named one of those two files. **M4 carries direction 1's own sentence and
lists no files**, which is the check that the two guards do not share an output string — either one
covering for the other's deletion is a defect shape this repo has already filed.

Gates, each read from its own exit code, never through a filter: `bash tests/inventory.sh` rc 0 ·
`shfmt -i 4 -d` rc 0 · `shellcheck` rc 0 at the gate's own `--severity=warning`, run with
`PATH=$HOME/.local/bin:$PATH` so it is **0.11.0**, the version CI pins, not `/usr/bin`'s 0.9.0 ·
`scripts/lint-file-budget.sh` rc 0. The file goes 300 → 345 lines, still under `TARGET_LINES=400`,
so it correctly stays unrowed and no ceiling moves.

At full severity `shellcheck` reports 2 × SC2016 (info) on the `SOURCED` grep. **Those are
pre-existing and not mine** — the base has 3 of them; my diff removes one by deleting the old
guard's message. I have not touched them.

The lane guard was verified with a negative control rather than assumed: `tests/inventory.sh` passes
(`tests.paths` owns it explicitly), and a staged `Makefile` is refused rc 1. No `.lane-override` was
used and nothing outside this lane's paths is touched.

## Over-engineering pass

Run by hand on merit, not because a gate prompted me — this repo's PR-opening gate keys its sentinel
to the pane's branch rather than the PR's, so its silence is not evidence that a review happened.
It found one thing: `n_expected` was computed unconditionally when it is only ever used inside the
failure branch, and I moved it inside the `if`. **That edit was later REVERSED** — the empty-expected
guard needs the value before the branch. The pass removed the exact computation a later guard
required, which is worth stating rather than quietly restoring. The battery was re-run afterwards and all rows still hold —
a mutation anchor invalidated by your own mid-session edit is a live hazard here, and the anchors
(the `SOURCED` pattern, two `run.sh` stanzas, the `MISSING=""` slice) were checked against it.

I also considered replacing the direction-2 loop with `comm` over two sorted lists. Declined: the
existing newline-delimited membership test is deliberately space-safe, and the file documents why.

## Not in this PR, deliberately

**The third item in #1420 — the aggregate counters at `:77`** (`n_py_dash n_py_fake n_node n_stack
n_selftest n_scen n_axes n_mini`, each tested against zero and so each able to fall from hundreds to
one and pass). The issue itself calls that a per-counter judgement rather than one ruling, and it is
a different question from the one this change answers. Leaving #1420 open for it; **treat those
counters as unchecked, not as clean** — nobody has verified them either way, and this PR is not
evidence about them.

---

## Amended after two adversarial passes — the head is now `28f5b6c`

The table above is the original 8 rows; M7 is the 9th and is in the harness. F2's check is not a
harness row — it was run standalone, with the mid-name-space control. Two more were added and every claim below was
measured, not argued:

- **M7 — both sets collapse to empty.** My original claim that this floor "strictly subsumes" the
  emptiness guard it replaces was **false**: a set difference sees disagreement and cannot see both
  sides empty together. `44b5146` went rc 0 where the base went rc 1. Fixed. The second reviewer
  reproduced it and ran the bystander control my table lacked (guard neutered → M7 rc 0, clean tree
  rc 0, ordinary drift still rc 1), which is what makes M7 a guard rather than a demonstration.
- **F2 — a leading-space filename was silently accepted as sourced.** `read -r` without `IFS=`
  stripped ` test-doctor.sh` to a name that IS in SOURCED. Before rc 0, after rc 1; mid-name space
  reds on both, so the fix is not the cause. Found unbriefed by the first reviewer.

**Corrections to this body's own claims, all of them mine:**

- **This PR changes no trigger.** Head and base agree on rc across every row. It **sizes and names
  what the base already caught badly** — which is what #1420 complained about. "M1 is the real
  regression" is true of its *message*, not its *rc*.
- **"Floor" is a misnomer** — there is no arithmetic; it is a set difference.
- **M6 proves something narrower** than claimed: under M1, direction 1 is blinded by construction,
  leaving direction 2 alone. Not "nothing else was catching it".

**This PR does NOT discharge #1420** and the issue carries `do-not-close`. It closes the
collect-offenders half; the single-survivor defect class it names survives at a different place
(#1429), and the `:77` counters are deferred and **unchecked, not clean**.

### DIFF
--- a/tests/inventory.sh
+++ b/tests/inventory.sh
@@ -189,8 +189,20 @@ while read -r base; do
     MISSING="${MISSING}${base}"$'\n'
 done <<<"$EXPECTED"
 n_missing=$(printf '%s' "$MISSING" | count)
+n_expected=$(printf '%s' "$EXPECTED" | count)
+# Nothing to compare is itself drift, and it is the ONE case a set-difference floor cannot see. If
+# every domain file and every stanza disappear TOGETHER — a botched revert of the #1105 split, or a
+# merge that drops the directory — then SOURCED and EXPECTED collapse to empty at the same time,
+# MISSING is empty, and the floor above is vacuously satisfied. The emptiness test this floor
+# replaced DID catch that, so it is asserted explicitly rather than quietly lost with it. Its own
+# sentence, not a share of the floor's: two guards on one string means either can cover for the
+# other's deletion.
+if [ "$n_expected" -eq 0 ]; then
+    echo "inventory drift: tests/stack/ holds no sourced domain files at all — the split's shape" \
+        "changed, or the directory was emptied; there is nothing for the drift floor to compare" >&2
+    exit 1
+fi
 if [ "$n_missing" -gt 0 ]; then
-    n_expected=$(printf '%s' "$EXPECTED" | count)
     echo "inventory drift: $n_missing of $n_expected domain files on disk are not sourced by" \
         "run.sh — a 'source' line was deleted, or, the closer this count runs to $n_expected, the" \
         "pattern above stopped matching. Either way the suite would still pass, silently skipping" \
--- a/tests/inventory.sh
+++ b/tests/inventory.sh
@@ -181,7 +181,12 @@ done <<<"$SOURCED"
 # none. Exiting on the first offender was the other half: it named one file, alphabetically first,
 # for a 24-file regression, and the reader sizes the problem from that message.
 MISSING=""
-while read -r base; do
+# `IFS=` is load-bearing: `read -r` into a single variable strips leading and trailing IFS
+# whitespace, so a file on disk whose name STARTS with a space, sourced by nothing, would be
+# stripped to its bare name — which IS in SOURCED — and silently accepted. A mid-name space was
+# caught either way, which is what made this invisible. Direction 1's identical construct above
+# needs no such fix: SOURCED comes from a character class that cannot emit a space.
+while IFS= read -r base; do
     if [ -z "$base" ]; then continue; fi
     # Matched between newlines, not spaces: a filename containing a space would survive the glob
     # above (it is iterated directly) only to be split apart by a space-delimited lookup here.
@@ -190,16 +195,21 @@ while read -r base; do
 done <<<"$EXPECTED"
 n_missing=$(printf '%s' "$MISSING" | count)
 n_expected=$(printf '%s' "$EXPECTED" | count)
-# Nothing to compare is itself drift, and it is the ONE case a set-difference floor cannot see. If
-# every domain file and every stanza disappear TOGETHER — a botched revert of the #1105 split, or a
-# merge that drops the directory — then SOURCED and EXPECTED collapse to empty at the same time,
-# MISSING is empty, and the floor above is vacuously satisfied. The emptiness test this floor
-# replaced DID catch that, so it is asserted explicitly rather than quietly lost with it. Its own
-# sentence, not a share of the floor's: two guards on one string means either can cover for the
-# other's deletion.
+# Nothing to compare is itself drift, and it is the ONE case the set difference above cannot see. If
+# every expected domain file and every stanza disappear TOGETHER — a botched revert of the #1105
+# split — then SOURCED and EXPECTED empty at the same time, MISSING is empty, and the comparison is
+# vacuously satisfied. The emptiness test this replaced DID catch that, so it is asserted explicitly
+# rather than quietly lost with it. Its own sentence, not a share of the one below: two guards on
+# one string means either can silently cover for the other's deletion.
+# Scope, stated because a comment that names a case this guard never sees is worse than no comment:
+# dropping the WHOLE directory does not reach here — `n_stack` counts 0 and the aggregate gate above
+# exits first, with byte-identical output whether this guard is alive or dead. What is unique to
+# this check is the narrower case where run.sh and the sectioned UNSOURCED files survive, keeping
+# `n_stack` non-zero, while the expected domain files and their stanzas both go.
 if [ "$n_expected" -eq 0 ]; then
-    echo "inventory drift: tests/stack/ holds no sourced domain files at all — the split's shape" \
-        "changed, or the directory was emptied; there is nothing for the drift floor to compare" >&2
+    echo "inventory drift: no domain files are expected to be sourced — every tests/stack file is" \
+        "run.sh, UNSOURCED, or gone, so there is nothing to check run.sh's stanzas against;" \
+        "restore the domain files, or update the pattern in tests/inventory.sh" >&2
     exit 1
 fi
 if [ "$n_missing" -gt 0 ]; then
--- a/tests/inventory.sh
+++ b/tests/inventory.sh
@@ -127,8 +127,8 @@ done
 # aggregate nor the per-file gate can see it. The hyphen-named domain files have no standalone CI
 # invocation of their own — the underscore-named test_*.sh suites do, and fail their own step — so
 # for them these two checks are the whole safety net. Both read one SOURCED list, so the pattern
-# cannot drift between them, and a pattern that stops matching trips the emptiness guard rather
-# than quietly satisfying both loops at once.
+# cannot drift between them, and a pattern that stops matching cannot quietly satisfy both loops at
+# once: direction 2 compares the list against the files on disk and reports how many are missing.
 #
 # The pattern is anchored to the start of the line and admits no `#` ahead of the `source`, and
 # that is the load-bearing half: without it a commented-out `# source "$HERE/x.sh"` reads here as
@@ -141,15 +141,26 @@ done
 # control flow rather than grep, and is not claimed.
 SOURCED=$(grep -oE '^[^#]*source "\$HERE/[A-Za-z0-9_.-]+\.sh"' tests/stack/run.sh |
     sed -E 's|^.*source "\$HERE/||; s|"$||')
-if [ -z "$SOURCED" ]; then # the grep itself going quiet must not read as "all present"
-    echo "inventory drift: no 'source \"\$HERE/...\"' lines found in run.sh — the split's shape" \
-        "changed; update the pattern in tests/inventory.sh" >&2
-    exit 1
-fi
+# The domain files run.sh is expected to source: every tests/stack/*.sh but the sourcer itself and
+# the suites that carry their own CI step and fail there. Hoisted above both directions because the
+# floor below needs its size, and both loops need its membership.
+# Not sourced by run.sh, by design — each is invoked as its own CI step and fails there:
+UNSOURCED="test_appliance_hugepages.sh test_compose.sh test_data_reset.sh \
+    test_firstboot_journal.sh test_os_update_recovery.sh"
+EXPECTED=""
+for f in tests/stack/*.sh; do
+    base="${f#tests/stack/}"
+    if [ "$base" = "run.sh" ]; then continue; fi # the sourcer itself
+    case " $UNSOURCED " in *" $base "*) continue ;; esac
+    EXPECTED="${EXPECTED}${base}"$'\n'
+done
 
 # Direction 1 (#1336) — every file run.sh claims to source must exist. Catches a domain file
 # deleted or renamed out from under a live `source` line.
 while read -r want; do
+    # A here-string over an empty SOURCED yields ONE blank line, not zero, so without this the
+    # loop would test `tests/stack/` and report a garbled name for a pattern that matched nothing.
+    if [ -z "$want" ]; then continue; fi
     if [ ! -f "tests/stack/$want" ]; then
         echo "inventory drift: run.sh sources tests/stack/$want, which does not exist — the file" \
             "was moved or renamed and the suite would still pass, silently skipping it" >&2
@@ -161,21 +172,33 @@ done <<<"$SOURCED"
 # the `source` lines that are PRESENT, so a line deleted while its file stays on disk is never
 # looked at; the per-file section gate then finds the orphaned file with its headers intact and
 # passes it. Nothing else asks the question in this direction.
-# Not sourced by run.sh, by design — each is invoked as its own CI step and fails there:
-UNSOURCED="test_appliance_hugepages.sh test_compose.sh test_data_reset.sh \
-    test_firstboot_journal.sh test_os_update_recovery.sh"
-for f in tests/stack/*.sh; do
-    base="${f#tests/stack/}"
-    if [ "$base" = "run.sh" ]; then continue; fi # the sourcer itself
-    case " $UNSOURCED " in *" $base "*) continue ;; esac
+#
+# This is a FLOOR, not an emptiness test, and it collects rather than exiting on the first offender
+# (#1420). It used to be guarded by `[ -z "$SOURCED" ]`, on the reasoning that the pattern going
+# quiet must not read as "all present" — and that guard failed open on a real 24-of-25 breakage,
+# because lib.sh's stanza is bare by design and kept matching. A fallback condition of "the set is
+# empty" is defeated by a single survivor. Comparing the two sets makes one survivor as loud as
+# none. Exiting on the first offender was the other half: it named one file, alphabetically first,
+# for a 24-file regression, and the reader sizes the problem from that message.
+MISSING=""
+while read -r base; do
+    if [ -z "$base" ]; then continue; fi
     # Matched between newlines, not spaces: a filename containing a space would survive the glob
     # above (it is iterated directly) only to be split apart by a space-delimited lookup here.
     case $'\n'"$SOURCED"$'\n' in *$'\n'"$base"$'\n'*) continue ;; esac
-    echo "inventory drift: tests/stack/$base is on disk but nothing sources it — a 'source' line" \
-        "was deleted and the suite would still pass, silently skipping the file's sections;" \
-        "restore the line, or add the file to UNSOURCED with a reason" >&2
+    MISSING="${MISSING}${base}"$'\n'
+done <<<"$EXPECTED"
+n_missing=$(printf '%s' "$MISSING" | count)
+if [ "$n_missing" -gt 0 ]; then
+    n_expected=$(printf '%s' "$EXPECTED" | count)
+    echo "inventory drift: $n_missing of $n_expected domain files on disk are not sourced by" \
+        "run.sh — a 'source' line was deleted, or, the closer this count runs to $n_expected, the" \
+        "pattern above stopped matching. Either way the suite would still pass, silently skipping" \
+        "every section in these files. Restore the line, fix the pattern, or add the file to" \
+        "UNSOURCED with a reason:" >&2
+    printf '%s' "$MISSING" | sed 's/^/  /' >&2
     exit 1
-done
+fi
 
 # --- emit -----------------------------------------------------------------
 cat <<EOF


VijitSingh97:
**Correcting this PR's own body, and the head has moved to `2c9783c`. A review in flight against
`44b5146` is reviewing a claim I have since refuted myself.**

The body said the floor "strictly subsumes" the `[ -z "$SOURCED" ]` guard it replaces, and offered
M5 (pattern matches nothing, 25 of 25) as the evidence. **That claim was false**, and it failed in
the quiet direction.

A set-difference floor detects DISAGREEMENT between the sourced list and the files on disk. It
cannot see the two collapsing together. With every domain file and every `source` stanza deleted —
a botched revert of the #1105 split, or a merge that drops the directory — `SOURCED` and `EXPECTED`
are both empty, `MISSING` is empty, and the floor is vacuously satisfied. M5 did not cover this
because it empties only one of the two sides.

Measured on one fixture, both versions, rather than argued:

    every domain file + every stanza deleted
      old (emptiness guard) .... rc=1   caught it
      new (floor), before ...... rc=0   REGRESSION — the guard I deleted was load-bearing here
      new (floor), after ....... rc=1   caught, with its own sentence

`2c9783c` asserts the empty-expected case explicitly. It gets its own message rather than a share of
the floor's, because two guards on one output string means either can silently cover for the other's
deletion.

**Two things I want stated against myself rather than buried:**

1. **The over-engineering pass in the original body actively caused this.** It moved `n_expected`
   inside the failure branch on the grounds that it was only used there. The correctness fix needs
   it *before* the branch. A simplification removed the exact computation the fix required, and I
   reported that pass as a clean win.
2. **This correction is a new claim and starts at zero evidence.** It was freshly reasoned, which is
   the condition under which a fix feels most checked and gets attacked least. The battery is now
   **9/9**, and M7 — the new row — carries the control that makes it evidence rather than a
   demonstration: the OLD guard reds on the same fixture, so this is a regression closed, not a
   hypothetical guarded.

I found this by running the attack the reviewer was briefed to run, on my own diff, before the
answer came back. That is not a reason to shorten the review — it is the same diff plus one more
claim of mine to break, and the deleted-guard question is still the right thing to aim at.

One state I checked and am NOT claiming to have fixed, because it is pre-existing and out of this
PR's scope: with every hyphen-named domain file gone but `lib.sh` still sourced, **both the old and
the new version pass at rc 0** — `lib.sh`'s bare stanza keeps `SOURCED` non-empty on one side while
`EXPECTED` holds only `lib.sh` on the other, so the sets agree. That is the same single-survivor
shape #1420 is about, at a different place, and neither version of this file has ever caught it.
Reported rather than folded in.

VijitSingh97:
The pre-existing hole described at the end of my previous comment — the one this PR deliberately does not fix — is now filed as #1429, with the reproduction and the positive control that proves the gate is alive in that same tree.

Refs #1429. It is **not** closed by this PR and no commit here names it in a closing construction: keywords fire from commit messages as well as bodies, and negation does not work, so the only safe handling is not to write one.

VijitSingh97:
## Non-author adversarial review — verdict FINDINGS, two, both LOW, neither blocking

**Read at `44b5146`. Scored by me against `2c9783c`, the current head.** The pass was dispatched
15:19:11Z; `2c9783c` was committed 15:22:57Z, so the reviewer resolved the earlier sha. That timing is
stated because it changes how each finding counts, and I have re-derived the part that decides it.

### Provenance, first — one finding is already closed and is NOT scored as new

**F1 (the empty-both-sets collapse) is the defect the author found and fixed in `2c9783c`, before this
answer existed.** Reviewer and author reached the same state independently and by different routes,
and the measurements agree exactly:

    every domain file AND every source stanza deleted
      base (old emptiness guard) ..... rc=1
      44b5146 (floor, before fix) .... rc=0    ← the regression
      2c9783c (explicit guard) ....... rc=1    ← closed, with its own sentence

**Recorded as CONFIRMING the author's own correction, not as an open finding.** The author's public
correction stands as the record of it.

Worth stating honestly rather than overclaiming: the brief explicitly asked the reviewer to construct
that state, so an F1 hit measures *"executes a brief"* — not general acuity. The author made that
point to me and it is right. **F2 is where the acuity shows, because nothing briefed it.**

### F2 — LOW but REAL, unbriefed, and it SURVIVES at `2c9783c`. Verified by me at that head.

`tests/inventory.sh:184` — `while read -r base; do … done <<<"$EXPECTED"`. `read -r` with a single
variable strips leading and trailing `IFS` whitespace before assigning. A file on disk whose name has
a **leading space**, sourced by nothing, is silently accepted: `read` strips it to the bare name,
which *is* in `SOURCED`, so direction 2 never fires.

    leading-space filename, unsourced:   head rc=0 (accepted)  ·  base rc=1 (caught)

A **mid**-space name is still caught on both, so the protection holds for the case the comment names.
I confirmed at `2c9783c` that the line is unchanged — still `while read -r base`, no `IFS=`. Fix is one
token: `while IFS= read -r base`.

`:160`'s direction-1 loop has the same construct and is **harmless**: `SOURCED` is built from a
character class that cannot emit a space. Do not "fix" it for symmetry without saying why.

### The comment at `:186-187` is falsified — but not for the reason the review gives, and this matters

The review says *"it is no longer iterated directly — that is precisely the property this PR removed."*
**That is not accurate, and I checked rather than relaying it.** The glob at `:151` **is** still
iterated directly (`for f in tests/stack/*.sh`). What changed is downstream: `EXPECTED` is now
serialised into a newline-delimited string and re-parsed through `read -r`, which **undoes the
protection the direct iteration bought.**

So the false part is the comment's *conclusion* — that a space-containing filename therefore survives
intact to the lookup — not its parenthetical. Correct the reasoning, not the clause.

**And do NOT touch `:53` or `:62`.** They make the same "iterated directly" claim about *different*
globs (`:55` and `:63`), both of which are still directly iterated and still protected. Those two
comments are TRUE. A sweep-by-grep of that phrase would break working documentation.

### M6 DISCRIMINATES — and the reviewer ran the positive control the table does not contain

The failure mode the brief named — a control that cannot fail — **is not present**:

    M6 scaffold (dir-2 excised), unmutated ....... rc=0   baseline
    M6 scaffold + M4 (file removed from disk) .... rc=1   ← scaffold is ALIVE
    M6 as authored (dir-2 excised + M1) .......... rc=0   ← reproduced

Excision cleanliness was checked by grep, not assumed. **Two limits the body currently overstates:**

- M6 proves *"direction 2 is the sole catcher of M1"*. It does **not** prove *this PR's* direction 2 is.
  **M1 reds on the BASE too** (rc=1) — it named one file instead of 24. So *"M1 is the real regression"*
  is true of M1's **message**, not its **rc**.
- Under M1, direction 1 is **blinded by construction** — the re-anchored pattern empties `SOURCED` down
  to `lib.sh`, so direction 1 has nothing left to check. `M6 + M1 + M4` is rc=0. M6's green therefore
  means *"M1 blinds direction 1, leaving direction 2 alone"*, which is the right conclusion but narrower
  than *"nothing else was catching it"*.
- **The row that actually discriminates head from base is M3**, not M1: head names both offenders,
  base names one.

### Exit-code semantics: UNCHANGED apart from F2

Same trigger set, collected instead of first-exit, still before the emit. Across control / M1 / M3 /
M5 / M4 / glob-empty, every head rc equals the base rc. **No case was found where base passes and head
reds.** So the collect-all change is a message change, not a trigger change — the "reports more, reds
in fewer cases" regression shape was looked for and is absent.

### Three things ruled OUT, having been tried

- *floor computes to 0 or negative* — impossible; there is no arithmetic. `MISSING` is a collected list.
  **"Floor" is a misnomer in the body** — it is a set difference, with no underflow surface.
- *glob matches nothing* — caught **upstream** at `:77-83` (`n_stack` hits 0), before direction 2.
- *directory read fails* — `:12` `cd "$ROOT" || exit 1` makes cwd unconditional.

### Honest limits of this pass — read these before treating anything as cleared

- **Rows M2 and M4 were NOT scored** as the body states them; the reviewer ran an M4-shaped probe only
  inside the M6 scaffold.
- **No gate was run.** Per instruction: no `make`, no lint, no `shellcheck`, no suite, no lock, no bench.
  `shfmt`/`shellcheck`/budget results in the body are **accepted, not verified**. Line count and unrowed
  status were verified by reading.
- **The `exit 3` did-it-apply harness was not used**; the reviewer substituted its own `cmp` check.
- **`detect-topology.py` was not run** by the reviewer; it read the diff for leaks instead. I note the
  diff is repo-relative filenames and prose.
- The gate is blind to a **relocation** — move the domain files and update the stanzas, and both sets
  shrink together and compare equal. **Base is equally blind, so this is not a regression**, but it is
  the honest limit of "derived from disk".
- The `:77` aggregate counters remain **unchecked, not clean** — the author deferred them and the
  reviewer agrees; the F1 probe touched their blind spot without testing them.

### What is owed

**M7 — the row added in `2c9783c` — has had no non-author look at all**, because it did not exist when
this pass was dispatched. Its control (the old guard reds on the same fixture) is the right shape and
is also exactly the shape this project has been fooled by before. A second narrow pass covering only
the `44b5146..2c9783c` delta is being dispatched; **this review does not cover M7.**

Merge authority remains the author's once that pass is clean. F2 and the `:186-187` reasoning are the
only two items this review leaves open, and neither blocks.

VijitSingh97:
**Correcting my own mutation table's framing. Re-derived at the two heads myself rather than taken
from the review, and it is sharper than what was relayed to me.**

My body calls M1 "the real regression". **It is not a regression in detection terms at all** — the
base catches it too:

    M1 (pattern re-anchored, 24 of 25 lost)
      base (pre-PR) .... rc=1   "test-appliance-install.sh is on disk but nothing sources it"
      head (2c9783c) ... rc=1   "24 of 25 domain files on disk are not sourced by run.sh"

    M3 (two stanzas deleted)
      base ............. rc=1   names test-backup.sh, lists 0 files
      head ............. rc=1   "2 of 25", lists both

**So M1's rc is not a discriminator, and neither is M3's.** What changed in both is the MESSAGE. I
was told M3 is "the row that actually discriminates"; having run it, that is true of its *list* and
not of its *exit code* either. **On rc alone, head and base agree on every row in the table.**

The honest statement of what this PR buys, which is narrower than my body implies: **it does not
catch anything the base failed to catch. It sizes and names what the base already caught badly.**
That was #1420's actual complaint — a 24-file regression announced as one file, alphabetically
first — and the issue says so. My mutation table then oversold it by borrowing the language of
regression detection.

The one row that genuinely moved rc was **M7**, and it moved it the wrong way: `44b5146` went rc=0
where the base went rc=1. That was a regression I introduced and fixed in `2c9783c`. It is the only
rc-level change this PR has ever made, and it is a repair of my own damage rather than a gain.

Two smaller corrections while I am here, both from the review and both fair:

- **"Floor" is a misnomer throughout my body and commit messages.** There is no arithmetic and no
  underflow surface — it is a set difference. The name oversells it as a numeric guard.
- **M6 proves something narrower than I claimed.** It shows that under M1, direction 1 is blinded by
  construction (`SOURCED` collapses to `lib.sh`), leaving direction 2 as the only live check. That
  is not the same as "nothing else in the file was catching it."

None of this changes the change itself; it changes what the evidence in the body is entitled to
claim, which is the part a reviewer or merger acts on.

VijitSingh97:
## Second, narrow non-author pass — read at `2c9783c`. VERDICT: M7 CONFIRMED. Two findings, both LOW, neither blocking.

Subject was only the `44b5146..2c9783c` delta (one commit, one file, +13/-1). It deliberately did
**not** re-score M1–M6 or the collect-all change — the first pass covers those.

### M7 survives an attack it was not subjected to

The author's control is *"the old guard reds on the same fixture"*. That is the right shape, and this
pass ran the control the table does **not** contain — the one that separates a guard from a
demonstration. With the new guard neutered (`-eq 0` → `-eq -999`, `diff`-verified as a one-line change):

    M7 + head, guard NEUTERED ................... rc=0   nothing else catches it — NOT a bystander kill
    clean tree + head, guard NEUTERED ........... rc=0   the mutation reds nothing on its own
    1 stanza deleted + head, guard NEUTERED ..... rc=1   scaffold ALIVE ("1 of 25 … test-backup.sh")

**The third row is the one that matters:** the neutered scaffold still reds on ordinary direction-2
drift, so M7's rc=0 under neutering is the *guard's absence*, not a broken harness. Combined with the
reproduced three-row table, M7 is a real, aimed, load-bearing kill.

**No coverage was narrowed.** The other case the deleted guard owned — all 25 stanzas gone, files still
on disk — is caught at head with a strictly better message than base's.

### D1 — LOW, should-fix. The comment names a case this guard never sees.

`:194-195` (and the commit message) motivate the guard with *"a botched revert of the #1105 split, **or
a merge that drops the directory**"*. **The second case is caught upstream and the new guard is never
reached:**

    whole tests/stack/ dropped, head ............ rc=1   "n_stack counted 0 tests"   (:79)
    whole tests/stack/ dropped, guard NEUTERED .. rc=1   identical message           (:79)

Byte-identical with the guard alive and dead, so `:77-83` owns that case entirely. The guard's real
unique coverage is narrower: domain files deleted **while `run.sh` and the sectioned `UNSOURCED` files
survive** (keeping `n_stack` non-zero) **and** the stanzas go with them — the #1105-revert half only.

This is the shape this project already tracks — **a true symptom documented with a mechanism that
cannot fire** — and it sits in the comment block whose whole job is telling the next reader what the
guard is for. Drop the clause, or say that case is caught at `:79` first.

### D2 — LOW, nit. The only drift message in the file with no remedy.

Every other one ends by telling the operator what to do (`:80`, `:105`, `:119`, `:209`). `:201-202`
ends at diagnosis. **The guard it replaced did carry a remedy** (base `:146`, *"update the pattern in
tests/inventory.sh"*).

Folded in rather than filed separately: *"holds no sourced domain files at all"* describes `SOURCED`,
but `n_expected` counts **files on disk expected to be sourced**. A state where those differ was
constructed and the guard fires with that wording while a file genuinely is sourced. Contrived — but
the sentence is about the wrong set.

### ⚠ The PR BODY is stale against the head, and the body is what lands in the merge record

The correction comment covers this in the thread; the body does not. Three claims are false at
`2c9783c`:

- *"Mutation battery, **8/8**"* — it is **9/9**; M7 exists only in the comment.
- *"The file goes 300 → 323 lines"* — it is **335**. The conclusion still holds (`TARGET_LINES=400`,
  no row in `docs/dev/file-budget.tsv`, stays unrowed, no ceiling moves) — verified by reading.
- The **Over-engineering pass** section says `n_expected` was *"moved inside the `if`"*. `2c9783c`
  **reversed that**. The body currently reports as a clean win the edit the delta had to undo.

**Worth one edit before merge** — not because it blocks, but because the thread is where the correction
lives and the body is where the record lands.

### Confirmed clean on the delta

- **#1429's filing stands** — fixture built (all `test-*.sh` gone, `lib.sh` still sourced, one stanza
  left): base rc=0 **and** head rc=0, `n_expected`=1, exactly as disclosed. The delta does not touch it.
- **No ordering hazard between the two guards.** `MISSING` is built only from `EXPECTED` (`:184-190`),
  so `n_expected == 0` implies `n_missing == 0`; the two `if`s are mutually exclusive and their order
  is inert.
- **The open F2 does not weaken the delta.** `n_expected` is computed at `:192` straight from
  `EXPECTED` via `printf | count`, never through `read`, so a leading-space filename still counts.
- **`count()` on empty is safe** — `grep -c .` emits `0` and returns 1; `|| true` at `:35` keeps the
  value `"0"`. Proven by the guard firing at all.
- **Leak scan clean with a positive control that fired.**

### Honest limits — read before treating anything as cleared

- **No gate was run**: no `make`, no lint, no `shellcheck`, no `shfmt`, no suite, no lock, no bench.
  `shfmt`/`shellcheck` cleanliness of the new block is **accepted, not verified** — a shape claim
  against the sibling block, not a gate result.
- All work on `git archive` scratch copies under `/tmp`. **No worktree touched, no ref moved.**
- **M1–M6, the collect-all change, and the first pass's open F2 and `:186-187` correction were NOT
  re-scored.** Confirmed unmodified at this head; not re-derived.
- The `:77` aggregate counters remain **unchecked, not clean.** D1 proves `n_stack` is alive on a
  whole-directory drop; that says nothing about the other seven.

**Merge is the author's.** Nothing here blocks it. The one thing worth doing first is the body edit
above, so the merge record matches the code.

VijitSingh97:
**Merging over two red image-build jobs — disclosed before the fact so it can be overruled.**

`Build image (dashboard)` and `Build image (tor)` both fail on **CVE-2026-14456** (openssl, HIGH,
DoS via unbounded memory growth). Read from the job logs, not inferred from the check names:

- **dashboard**, debian 13 — `libssl3t64` / `openssl` / `openssl-provider-legacy`,
  `3.5.6-1~deb13u2` → `3.5.7-1~deb13u2`. `Total: 3 (HIGH: 3, CRITICAL: 0)`
- **tor**, alpine — `libcrypto3` / `libssl3`, `3.5.7-r0` → `3.5.8-r0`.
  `Total: 2 (HIGH: 2, CRITICAL: 0)`

Both are upstream base-image CVEs, owned by #1427. **Not muted** — the fixed packages exist.

The standing ruling permits a merge past these two reds when the diff cannot affect image contents.
That premise is checked, not assumed: this PR's diff is **`tests/inventory.sh` only, 63+/18−**.

**The one gate that covers this file is green.** `Shell tests (shellcheck + pithead suite)` passed in
**10m50s** on `28f5b6c` — run `33186303584`, head sha confirmed as `28f5b6c` rather than taken from
the check list.

Merging does **not** close #1420 (it carries `do-not-close`; the acceptance assessment is its newest
comment) and does not close #1429.

Issue #1432, opened 2026-08-28
lint-sh's shfmt leg survives a dead enumeration on its three hard-coded arguments — 3 files checked of 100, rc=0


VijitSingh97:
Companion to #1431, found by sweeping the whole `lint` aggregate rather than stopping at the one script. This one is in `lint-sh` — the leg quoted as evidence more than any other.

## The shape

`Makefile`, the `lint-sh` recipe:

```make
shfmt -i 4 -d pithead pithead-completion.bash os/installer/pithead-install $(shell git ls-files '*.sh' | grep -v '^docs/research/')
```

Three filenames are hard-coded; the other 97 arrive from `$(shell git ls-files ...)`. If that enumeration returns nothing — a broken `git ls-files`, a pipe failure inside `$(shell)`, an invocation from outside a work tree — **make substitutes the empty string and does not error.** `shfmt` then runs with three arguments, finds them well-formatted, and exits 0.

Measured: `shfmt -i 4 -d pithead` alone is rc=0, and `git ls-tree -r --name-only origin/develop-v2 | grep '\.sh$' | grep -v '^docs/research/'` is **97 files** at that tip. So the failure mode is `lint-sh` checking formatting on 3 of 100 files, printing nothing unusual, and exiting 0. Nothing distinguishes that from a clean run — not the output, not the exit code.

This differs from #1431 in one way worth noting: it is not a *fully* vacuous pass, which makes it harder to catch. Three files really are checked, so any guard asserting "we checked something" would still pass.

## The negative control — this is an asymmetry, not a pattern I went looking for

`lint-yaml` is built the same way (`uvx yamllint $(shell git ls-files '*.yml' '*.yaml')`) and is **fine**: `yamllint` with no file arguments is an argparse error, `rc=2`, loud. I checked this before writing the issue precisely because a sweep that finds the same defect everywhere it looks is usually a broken instrument. `lint-yaml` degrades safely; `lint-sh` does not. The difference is entirely whether the tool has hard-coded arguments keeping it alive when the enumeration dies.

## The fix already exists in the same file

`lint-toml` in the same `Makefile` guards exactly this:

```make
@test -n "$$(git ls-files '*.toml')" || { echo "lint-toml: zero tracked TOML files — refusing a vacuous pass"; exit 1; }
```

[...]

Issue #1431, opened 2026-08-28
lint-operator-strings skips its whole frontend scan on an empty enumeration, then prints OK — the guard its two siblings have


VijitSingh97:
`scripts/lint-operator-strings.sh` skips its entire frontend scan when the file enumeration comes back empty, and then prints its success line anyway — including the clause that claims the frontend was scanned.

## The shape

`scripts/lint-operator-strings.sh:191-193` builds the frontend file list:

```sh
files=$(git ls-files 'dashboard/mining_dashboard/web/static/*.mjs' \
    'dashboard/mining_dashboard/web/static/*.js' \
    'dashboard/mining_dashboard/web/templates/*.html' | grep -v '\.min\.js$' || true)
```

`:194` is `if [ -n "$files" ]; then`. An empty list takes the else-branch that does not exist: the whole scan is skipped, `fail` stays 0, and `:211` prints

> `operator strings OK — no issue/PR numbers in pithead output or the dashboard frontend, no bare docs/ paths in pithead operator text`

The trailing `|| true` on `:193` additionally swallows a non-zero exit from the enumeration itself, so a broken `git ls-files` and a clean frontend produce byte-identical output and the same exit code.

**An empty list is never a legitimate state here.** 31 tracked files match those globs at `develop-v2`, so `files` being empty always means the enumeration broke, never that there was nothing to check.

## Both sibling linters already refuse exactly this

- `scripts/lint-topology-classes.sh:26` defines `enforce_nonempty_enumeration()` (whole-repo check at `:27`), invoked at `:285`. Its comment states the rule: *"a swallowed pipe error and a genuinely clean tree look identical from a zero-hit scan, so an empty enumeration is always fatal, not just when the exit code says so."*
- `scripts/lint-file-budget.sh:67` defines the same function, invoked at `:141`, plus a second zero-check on the candidate list at `:147-149`. Its comment reads *"Mirrors lint-topology-classes.sh's own guard."*

Two of the three siblings were deliberately paired and the third was left out. Nothing in the third script explains the asymmetry, so this reads as an oversight rather than a decision. Its only zero-test is `[ "$st_fail" -eq 0 ]` at `:161`, which is the self-test verdict, not an enumeration guard.

Why it matters beyond this one script: `make lint` passing is quoted as evidence in PR bodies routinely, and this is the leg of it that can pass without looking at anything.

## Suggested fix

[...]

Issue #1429, opened 2026-08-28
tests/inventory.sh's drift gate is satisfied by a single file on both sides — the whole domain layer can vanish at rc 0


VijitSingh97:
`tests/inventory.sh`'s drift gate compares the `source` stanzas in `run.sh` against the domain files
on disk. **A single file that appears on BOTH sides keeps the comparison satisfied while everything
else disappears from both** — so the entire domain layer can be deleted and the gate passes rc 0.

`lib.sh` is that file. It is sourced by `run.sh` and it is not in `UNSOURCED`, so it counts as
expected *and* as sourced.

### Reproduction

Delete every hyphen-named domain file and its `source` stanza, leaving `lib.sh` and its stanza:

    rm -f tests/stack/test-*.sh
    # and strip the matching `source "$HERE/test-...sh"` lines from tests/stack/run.sh

Then `bash tests/inventory.sh`:

| version | result |
|---|---|
| current | **rc 0 — passes** |
| before the floor landed (the `[ -z "$SOURCED" ]` guard) | **rc 0 — passes** |
| positive control: also remove `lib.sh`'s own stanza | **rc 1**, `1 of 1 domain files on disk are not sourced` |

**The control is the part that makes this a finding rather than a broken fixture.** The gate is alive
in that same tree and fires one step further along; it simply has nothing to say about the state
above it. 24 domain files and 24 stanzas can vanish together and nothing anywhere reports it.

### Why nothing else catches it

- The aggregate counters are zero-tests, and `n_stack` still counts `run.sh`'s own sections, so it
  stays comfortably non-zero.
- The per-file section gate iterates the files that are *present*; the deleted ones are simply not
  iterated.
- Direction 1 checks sourced-implies-exists, which holds — `lib.sh` exists.
- Direction 2 checks exists-implies-sourced, which also holds, because `lib.sh` is the only file
  left on the disk side and it *is* sourced.

Both directions are satisfied by the same single survivor.

### Shape

This is the single-survivor defect named in the earlier inventory issue, at a different place. There
the guard's fallback condition was "the set is empty" and one surviving match defeated it; here the
guard is a set comparison and one file present on both sides defeats it. Same lesson: **a guard
whose trigger is a relationship between two quantities cannot see both quantities collapse
together.**

### Not fixed here, deliberately

Found while attacking the floor's own subsumption claim. The empty-*both*-sets case was a real
[...]

PR #1433, opened 2026-08-28
fix(lint): two lint legs reported success without checking what they claimed

VijitSingh97:
Two lint legs that reported success without checking what they claimed to check. Refs #1431 and
#1432. They are one sweep because they are one failure mode — an enumeration dies, and the check
built on it stays green — and because the second is only visible once you go looking for the first.

## The two defects

**#1431 — `scripts/lint-operator-strings.sh`, the frontend scan skipped itself.** The script
enumerated the dashboard frontend, then wrapped the scan in `if [ -n "$files" ]`. An empty list
skipped the scan entirely, left `fail` at 0, and still reached the closing line that states the
frontend **was** scanned. A moved or renamed `web/` directory would have read as a clean frontend
indefinitely.

**#1432 — `Makefile`, `lint-sh`'s shfmt leg survived on hard-coded arguments.** The line is three
named files plus `$(shell git ls-files '*.sh' | grep -v '^docs/research/')`. The three are not
`*.sh` (`pithead`, `pithead-completion.bash`, `os/installer/pithead-install`), so they must stay
listed — and they are also what keeps the command alive when the enumeration dies. shfmt receives
three real arguments and exits 0. This one is harder to catch precisely because it is **not**
vacuous: a check asserting "we formatted something" still passes.

## Measured, with the controls

Both run in a scratch copy, never in a live worktree. #1431 with the frontend globs repointed at a
directory that does not exist; #1432 with a pathspec that matches nothing.

    #1431  base + dead enumeration ............ rc=0  "operator strings OK — ... and the dashboard frontend"
           head + dead enumeration ............ rc=1  "the dashboard frontend enumeration returned zero files"
           head + dead enum + guard NEUTERED .. rc=0  <- the guard is the sole cause of the red
           head + guard neutered, no kill ..... rc=0  <- the neuter reds nothing on its own
           base and head, unmutated ........... rc=0  <- both controls

    #1432  base leg + dead enumeration ........ rc=0  3 files checked of 100, reported green
           new guard + dead enumeration ....... rc=1  "zero tracked *.sh files — refusing a vacuous pass"
           new guard + REAL enumeration ....... rc=0  97 files; a healthy tree is not blocked

The two neutered rows are what make #1431's red evidence rather than a demonstration: with the
guard dead the same mutation goes back to green, and the mutation alone reds nothing.

**The negative control, and it is why this is a finding rather than a pattern applied five times.**
`lint-yaml` is built identically — `uvx yamllint $(shell git ls-files ...)` — and needs no guard:
with no arguments `yamllint` is **rc=2**, loud. Measured, not assumed. A sweep that finds the same
defect in every leg it inspects is a broken instrument, so the leg that came back clean matters as
much as the two that did not.

## Two design calls, both against the obvious move

**The `|| true` on the enumeration STAYS.** Removing it looks like an improvement and would have
cost the guard. Under `set -o pipefail` an empty enumeration makes `grep -v` exit 1, which errexit
turns into a silent death **at the assignment** — before the refusal can run. Verified both ways:

    without `|| true` .... exits 1 having printed NOTHING; the guard is never reached
    with `|| true` ....... the refusal prints and exits 1

Both are rc=1, so an exit-code-only test cannot tell them apart. The `|| true` and the refusal are
one mechanism: swallowing that status is only safe *because* something now checks the result.

**#1431 is not mirrored from its siblings, deliberately.** `lint-topology-classes.sh:26` and
`lint-file-budget.sh:67` both test the whole tree (`git ls-files | wc -l`). That cannot see a moved
`web/`: 533 tracked files, 0 matching the frontend globs. The scoped `[ -z "$files" ]` fires in
every case the mirrored guard would **and** in the case it misses, so mirroring would have added a
guard that can never fire alone.

## Coverage

Two `--self-test` rows, one each way (an empty enumeration is refused, a non-empty one is accepted),
so `make lint-operator-strings` covers the guard rather than this PR body doing it. That target runs
the self-test before the real scan, and reaches CI at `ci.yml:512`.

Noted while mutating, not fixed here: `scan_frontend` is `awk '...' <files>`, so with zero arguments
it blocks on **stdin**. The old skip is what prevented that. Removing the skip without adding a
refusal would have converted the false green into a hung job.

## Shared file — disclosure

**`Makefile` is not in this lane's paths.** It is committed with a one-shot `.lane-override`, and
the guard was confirmed to refuse the staged file *first*, so the override is doing visible work
rather than covering a mistaken path claim. Expect a conflict with any other lane touching this
Makefile. Only `scripts/lint-operator-strings.sh` was granted to this lane, and only for #1431.

## Gates, and the honest limits

Run, each read from its own exit code: `bash scripts/lint-operator-strings.sh --self-test` rc 0 ·
`bash scripts/lint-operator-strings.sh` rc 0 · `make lint-operator-strings` rc 0 ·
`shellcheck --severity=warning scripts/lint-operator-strings.sh` rc 0 at **0.11.0**, the version CI
pins, run with `PATH=$HOME/.local/bin:$PATH` rather than `/usr/bin`'s 0.9.0 ·
`shfmt -i 4 -d scripts/lint-operator-strings.sh` rc 0 · `scripts/lint-file-budget.sh` rc 0.
Neither file is rowed in the budget and neither needs to be: the script goes 211 → 238 lines against
`TARGET_LINES=400`, the Makefile 132 → 137, so no ceiling moves.

**Not run, and do not read these as cleared:** full `make lint-sh` was **deliberately not run
locally** — it shellchecks `tests/stack/run.sh`, which OOMs this box. Its parse and expansion were
checked with `make -n` instead, which confirms the new guard line is in the recipe but does **not**
execute it. **UPDATE — the guard line has since been EXECUTED through real `make`** by the successor
session on this lane, in scratch repos with `shellcheck` stubbed to exit 0 so the OOM leg never runs:
base + a dead enumeration is rc 0 with shfmt handed only the 3 hard-coded files, head is rc 2 by the
guard's own message, and deleting the guard line alone returns that mutation to rc 0 while reddening
nothing on a clean tree. Full results in a PR comment. **CI's `Shell tests` job remains the real check
on the shellcheck half of that leg**, which the stub deliberately says nothing about. `make lint` and
`make test` were not run in full. The `#1432` reproduction models a dead enumeration with an empty pathspec
rather than by breaking `git ls-files` itself, so it proves the guard's behaviour on an empty
enumeration, not every way one could become empty.

Neither issue had a runtime reproduction when filed; both have one now, and both were re-derived at
source rather than taken from the filing.

## Over-engineering pass

Run on this diff because the PR gate asks for one. **Note the gate's sentinel keys to the pane's
branch rather than the PR's, so its silence is not evidence that a pass happened** — this one is
recorded here and in `fed82e5` so the claim is checkable. Two findings, both applied: the guard's
4-line comment restated the refusal message below it (now 2 lines), and `2>/dev/null` on the
positive self-test row suppressed output that row cannot produce. `net: -2 lines`, 240 → 238.

Not applied, deliberately: `enforce_nonempty_frontend` has one production caller — a layer with one
caller by the letter of it — but two more in `--self-test`. Extracting it is what makes the guard
reachable from the self-test; inlining would trade covered behaviour for two lines.

The mutation battery was re-run after that pass and all six rows still hold with both anchors live.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,11 @@ lint-sh: ## shellcheck + shfmt over the CLI, build/* + dashboard/ container scri
 # run.sh sets for them. Drop pithead-boot from this line and those three report as SC2034 — it
 # costs ~300 lines to keep here, not the 294 KB of domain files.
 	shellcheck --severity=warning tests/stack/run.sh os/overlay/pithead-boot
+	@# Three non-.sh files are named outright below, so a dead enumeration still hands shfmt
+	@# three real arguments and exits 0 — 3 files checked of 100, reported as a pass. Guard the
+	@# enumeration itself, the way lint-toml already does. (lint-yaml needs no guard: yamllint
+	@# with no arguments is rc=2.)
+	@test -n "$$(git ls-files '*.sh' | grep -v '^docs/research/')" || { echo "lint-sh: zero tracked *.sh files — refusing a vacuous pass"; exit 1; }
 	shfmt -i 4 -d pithead pithead-completion.bash os/installer/pithead-install $(shell git ls-files '*.sh' | grep -v '^docs/research/')
 
 lint-py: ## ruff lint + format check on all repo Python (ruff runs via uv from the locked dev extra)
--- a/scripts/lint-operator-strings.sh
+++ b/scripts/lint-operator-strings.sh
@@ -105,10 +105,8 @@ scan_frontend() {
     ' "$@"
 }
 
-# A frontend enumeration that comes back empty is a broken enumeration, never a clean tree: the
-# globs below match 28 tracked files today and have never legitimately matched none. Skipping the
-# scan on empty would leave `fail` at 0 and let this script's closing line still claim the frontend
-# was scanned, so a moved or renamed `web/` directory would read as clean indefinitely.
+# An empty enumeration is broken, never a clean tree — 28 files match the globs below. Skipping
+# the scan on empty leaves `fail` at 0 while the closing line still claims the frontend was scanned.
 enforce_nonempty_frontend() { # <newline-separated file list>
     [ -n "$1" ] && return 0
     echo "operator strings: the dashboard frontend enumeration returned zero files." >&2
@@ -174,7 +172,7 @@ if [ "${1:-}" = "--self-test" ]; then
         echo "  self-test FAIL: an empty frontend enumeration was accepted"
         st_fail=1
     else echo "  self-test ok: an empty frontend enumeration is refused"; fi
-    if enforce_nonempty_frontend "$tmp/hit.mjs" 2>/dev/null; then
+    if enforce_nonempty_frontend "$tmp/hit.mjs"; then
         echo "  self-test ok: a non-empty frontend enumeration is accepted"
     else
         echo "  self-test FAIL: a non-empty frontend enumeration was refused"
--- a/scripts/lint-operator-strings.sh
+++ b/scripts/lint-operator-strings.sh
@@ -105,8 +105,10 @@ scan_frontend() {
     ' "$@"
 }
 
-# An empty enumeration is broken, never a clean tree — 28 files match the globs below. Skipping
-# the scan on empty leaves `fail` at 0 while the closing line still claims the frontend was scanned.
+# An empty enumeration is broken, never a clean tree: the globs below match tracked files in any
+# real checkout. Skipping the scan on empty leaves `fail` at 0 while the closing line still claims
+# the frontend was scanned. (No count here on purpose — one would be wrong twice over, since the
+# globs and the surviving set differ by the *.min.js bundles, and both drift with the frontend.)
 enforce_nonempty_frontend() { # <newline-separated file list>
     [ -n "$1" ] && return 0
     echo "operator strings: the dashboard frontend enumeration returned zero files." >&2
@@ -179,6 +181,25 @@ if [ "${1:-}" = "--self-test" ]; then
         st_fail=1
     fi
 
+    # The two rows above prove the FUNCTION. Neither proves the CALL SITE: delete the
+    # `enforce_nonempty_frontend "$files"` line and every row above still passes while the real
+    # invocation goes back to reporting a frontend it never read. That is this repo's recurring
+    # shape — a green assertion compatible with the feature not being wired in — and it is the very
+    # shape this script was changed to refuse, so it is checked here rather than described in a PR.
+    # Run the real script end to end in a throwaway git repo whose index is empty, which is the only
+    # way to make the enumeration return nothing without editing the globs.
+    self=$(cd "$(dirname "$0")" && pwd)/$(basename "$0")
+    wire="$tmp/wire"
+    mkdir -p "$wire"
+    (cd "$wire" && git init -q . && : >pithead) >/dev/null 2>&1
+    wire_out=$(cd "$wire" && bash "$self" 2>&1 </dev/null) && wire_rc=0 || wire_rc=$?
+    if [ "$wire_rc" -ne 0 ] && printf '%s\n' "$wire_out" | grep -q 'frontend enumeration returned zero files'; then
+        echo "  self-test ok: the real invocation refuses an empty frontend enumeration"
+    else
+        echo "  self-test FAIL: the real invocation accepted an empty frontend enumeration (rc=$wire_rc)"
+        st_fail=1
+    fi
+
     [ "$st_fail" -eq 0 ] && {
         echo "lint-operator-strings self-test OK"
         exit 0
@@ -218,6 +239,12 @@ fi
 files=$(git ls-files 'dashboard/mining_dashboard/web/static/*.mjs' \
     'dashboard/mining_dashboard/web/static/*.js' \
     'dashboard/mining_dashboard/web/templates/*.html' | grep -v '\.min\.js$' || true)
+#
+# THIS REFUSAL MUST STAY ABOVE THE SCAN. `scan_frontend` ends in `awk '...' "$@"`, and awk with zero
+# file arguments reads STDIN — so folding this check into the `if` below, or moving it after the
+# scan, converts the false green into a job that HANGS instead of failing. Measured both ways: with
+# this line here the script refuses in milliseconds; with it deleted and stdin held open, the scan
+# blocks until the runner's timeout.
 enforce_nonempty_frontend "$files" || exit 1
 
 # shellcheck disable=SC2086
--- a/scripts/lint-operator-strings.sh
+++ b/scripts/lint-operator-strings.sh
@@ -105,6 +105,18 @@ scan_frontend() {
     ' "$@"
 }
 
+# A frontend enumeration that comes back empty is a broken enumeration, never a clean tree: the
+# globs below match 28 tracked files today and have never legitimately matched none. Skipping the
+# scan on empty would leave `fail` at 0 and let this script's closing line still claim the frontend
+# was scanned, so a moved or renamed `web/` directory would read as clean indefinitely.
+enforce_nonempty_frontend() { # <newline-separated file list>
+    [ -n "$1" ] && return 0
+    echo "operator strings: the dashboard frontend enumeration returned zero files." >&2
+    echo "A broken enumeration and a clean scan both report zero hits — refusing to call either a pass." >&2
+    echo "Check the git ls-files globs in this script against dashboard/mining_dashboard/web/." >&2
+    return 1
+}
+
 # --- self-test: prove the scanners catch a planted #NNN and skip the tricky exemptions ----------
 if [ "${1:-}" = "--self-test" ]; then
     tmp=$(mktemp -d)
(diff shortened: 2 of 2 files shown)


VijitSingh97:
## The `lint-sh` guard has now been EXECUTED, not just parsed — and one of my own tests was invalid first

Successor session on this lane, picking the PR up. The body discloses two things as unverified, and
the more important one is that the #1432 guard line had only ever been checked with `make -n`. I ran
it, through real `make`, before the review lands — so the merge record carries an executed result
rather than a recipe inspection.

**Method.** Two scratch repos, `git archive` of head and of `develop-v2`, each `git init` + committed
so `git ls-files` is live; a stub `shellcheck` that exits 0 placed first on `PATH`, so the recipe's
`tests/stack/run.sh` leg never fires and cannot OOM this box; real GNU make, real `shfmt` 3.8.0. No
live worktree touched, no lint lock taken, no bench.

### #1432 — executed through `make lint-sh`

    control, head, clean tree ............. rc=0   shfmt handed 100 files (97 + the 3 named)
    control, base, clean tree ............. rc=0   identical argument list
    base + dead enumeration ............... rc=0   shfmt handed exactly the 3 hard-coded files
    head + dead enumeration ............... rc=2   "lint-sh: zero tracked *.sh files — refusing a vacuous pass"
    head, GUARD LINE DELETED, + dead enum . rc=0   <- the red is the guard's, nothing else catches it
    head, GUARD LINE DELETED, clean tree .. rc=0   <- deleting it reds nothing on its own

The mutation is `git rm --cached` over every tracked `*.sh`, leaving the files on disk, which is the
faithful shape: the three hard-coded arguments still exist and shfmt still succeeds on them. rc=2 is
make's status for a recipe line that exited 1. **Every row above was scored on the printed message
as well as the code**, because the body is right that rc alone cannot separate these forms.

### #1431 — executed on the script

    base, intact ....... rc=0   28 files enumerated
    head, intact ....... rc=0   28 files enumerated
    base + dead enum ... rc=0   "operator strings OK — no issue/PR numbers in ... the dashboard frontend"
    head + dead enum ... rc=1   "the dashboard frontend enumeration returned zero files" + 2 more lines

The base row is the defect in one line: the success sentence names the frontend, and zero files were
read. Self-test at head is rc 0 with both new rows present.

### The `awk`-on-stdin hazard the body flags — measured, and it does not occur

**My first attempt at this was invalid and I am recording that rather than the result it gave.** I ran
`timeout 15 sh -c 'sleep 60 | bash scripts/lint-operator-strings.sh'` and got rc=124 at *both* head and
base. That is not a hang: `sh` waits for the whole pipeline, so I was timing `sleep`, not the script.
Two sides "hanging" identically was the tell.

Re-run with the timeout applied to the script alone and stdin bound to a fifo held open by a silent
writer:

    head + dead enum, stdin open but silent ... rc=1, elapsed 0s   refusal reached before awk
    base + dead enum, stdin open but silent ... rc=0, elapsed 0s   skips, as before

So the refusal is unconditionally reached ahead of any zero-argument `awk`, and the false green does
not become a hung job.

### One observation, LOW, not a blocker

The guard evaluates `git ls-files ...` at **recipe time** (`$$(...)`) while the argument list it
protects is expanded by make at **parse time** (`$(shell ...)`). They are two independent evaluations
of the same command text, so the guard is a proxy for the enumeration rather than the enumeration
itself. `lint-toml` carries the identical duplication, so this mirrors the sibling faithfully and I am
not asking for a change. Worth knowing that my mutation kills both evaluations at once and therefore
does not exercise a divergence between them.

### What I did not do

No full `make lint` or `make test`; the shellcheck legs of `lint-sh` were stubbed out by design, so
this says nothing about shellcheck's own result — CI's `Shell tests` job remains the check on that,
and it is still pending at the time of writing. The budget gate ran rc 0 in scratch but skipped its
monotonic leg there for want of a base ref, which is a scratch artifact, not a result.

A non-author adversarial pass is separately in flight and has not reported; nothing above substitutes
for it.

VijitSingh97:
## Non-author adversarial review is in — FINDINGS ×4, none blocking, both issues DISCHARGED. Three folded in, one filed.

The pass ran independently of the verification I posted above and reached the same numbers by its own
commands: 100 files to shfmt on a healthy tree, rc=2 by the guard's message on a dead enumeration,
rc=0 on 3 arguments with the guard line deleted, the base reproducing the frontend false green, 28
surviving files, 238 lines. Two routes, same measurements. It also confirmed both controller rulings
at the head — not mirrored from the whole-tree siblings, and `|| true` kept and load-bearing.

**It found one row I did not run, and it is the useful one.**

### F2 — the call site was untested. FIXED in `58e7294`.

My two self-test rows prove the FUNCTION. Neither proves the WIRING: delete
`enforce_nonempty_frontend "$files" || exit 1` and both rows still report ok while the real
invocation goes back to reporting a frontend it never read. A guard against checks that pass without
checking, whose own coverage passes without checking. Measured, call site deleted and nothing else:

    two function rows ......... still ok      <- they never covered the wiring
    new end-to-end row ........ FAIL, "the real invocation accepted an empty frontend enumeration (rc=0)"
    self-test overall ......... rc=1
    intact .................... all three ok, rc=0

The new row runs the real script end to end in a throwaway git repo with an empty index — the only
way to make the enumeration return nothing without editing the globs — and asserts on the refusal
**message**, not on rc alone.

### F3 — no comment said the refusal must precede the scan. FIXED in `58e7294`.

`scan_frontend` ends in `awk '...' "$@"`, and awk with zero file arguments reads stdin, so folding
the check into the `if` or moving it below the scan turns the false green into a hung job. The
reviewer proved the hazard is real by deleting the call site with stdin held open: rc=124 at the
timeout, against 46 ms for the shipped order. That was recorded only in this PR body; the file is
what gets edited, so it is in the file now.

### F4 — the comment's count was wrong. FIXED in `58e7294`.

It said 28 files match the globs. 31 match; 28 survive the `.min.js` filter. Rather than correct one
stale number into another — both drift with the frontend — the count is gone and the reasoning
stands without it.

### F1 — pre-existing, same class, NOT fixed here. Filed as #1434.

The two `pithead` legs are fail-open on a missing target file. Re-derived at both heads before
filing: rename `pithead` away and the script prints five `grep`/`awk` fatals to stderr and then
`operator strings OK — no issue/PR numbers in pithead output`, **rc=0**, identical to the intact
run. Present at `develop-v2` too, so it is the sibling half of the defect this PR fixes rather than
a regression. Kept out of this diff on purpose so it stays the size it was reviewed at.

### Gates on `58e7294`

shellcheck 0.11.0 rc 0 · `shfmt -i 4 -d` rc 0 · `make lint-operator-strings` rc 0 ·
`scripts/lint-file-budget.sh` rc 0. 238 → 265 lines against `TARGET_LINES=400`; unrowed, and still
not needing a row. The dead-enumeration behaviour was re-measured after the edits and is unchanged
(rc 1 with the refusal; rc 0 and the OK line on an intact tree).

**Still gated on `Shell tests (shellcheck + pithead suite)`**, which is the only place the changed
`lint-sh` recipe runs for real — my execution of it and the reviewer's both stubbed `shellcheck` to
respect the OOM constraint, so neither says anything about that half. Not merging before it reports
on this head. The two red `Build image` jobs remain #1427.

VijitSingh97:
## Merging on `58e7294`, with the two `Build image` jobs red — disclosed, per the standing ruling

**The gate that mattered is green.** `Shell tests (shellcheck + pithead suite)` **passed** on this head.
That job is the only place the changed `lint-sh` recipe runs for real: both my execution of it and the
reviewer's replaced `shellcheck` with an exit-0 stub to respect the local memory cap, so neither of us
said anything about that half. It has now run with shellcheck 0.11.0 and shfmt 3.13.1 against the real
97-file enumeration.

`Lint per-surface` also passed, which is where `make lint-operator-strings` runs — so the new
end-to-end self-test row (the one that refuses an empty enumeration by running the real script in a
throwaway git repo) executes on the runner, not only locally.

**The two red jobs are `Build image (tor)` and `Build image (dashboard)`: the upstream base-image
openssl CVE tracked as #1427.** Merging past them here rests on two things I checked rather than
assumed:

1. **This diff cannot affect image contents.** It is `Makefile` and `scripts/lint-operator-strings.sh`.
   Every Dockerfile in the repo copies *named* files — there is no `COPY . .` anywhere — and neither
   changed file is referenced by any build context.
2. **Nothing in this PR could clear the red.** The half that is bumpable lives in a Dockerfile outside
   this lane's granted paths and belongs to #1427; the other half has no bump available at all, since
   upstream has not rebuilt the base tag it pins.

That second question is the one that decides it — not "is this red mine", which it plainly is not, but
"can anything I do clear it". #1427 stays open and release-blocking; a merge ships nothing to a user.

**This merge closes nothing automatically** — GitHub's own `closingIssuesReferences` for the PR
returns `[]`, which is a stronger check than grepping the body, since it reports what GitHub will
actually do. Two issues stay open regardless and are out of scope here: #1420, which carries the
label forbidding automatic closure, and #1429. I will shut #1431 and #1432 by hand afterwards.

PR #1435, opened 2026-08-28
test(stack): split the appliance-identity domain into its own file (#1105 Phase 1, develop-v2)

VijitSingh97:
Phase 1 of #1105: the appliance-identity domain moves out of `tests/stack/run.sh` into
`tests/stack/test-appliance-identity.sh`. `run.sh` 3841 -> 3400; the new file is 474 lines,
31 of them an authored header and 443 moved verbatim.

The domain: **who the appliance says it is, and everything the machine derives from that
answer** — the browsable name a headless setup resolves, the ssh access derived from it
(#786), the control-runner units rendered into `/run` because the root is read-only (#791),
and the TLS certificate that has to agree with the name list Caddy actually serves (#1132,
#1063).

## What was RUN

Both suites are full `bash tests/stack/run.sh` runs on the dev box, detached, each finished with a
verdict file rather than a log that stopped growing.

| | baseline `c505413` | after-proof `1956efa` |
|---|---|---|
| rc | 0 | 0 |
| verdict | 2922 passed, 0 failed | **2922 passed, 0 failed** |
| section headers | 272 | **272** |
| log lines | 3233 | **3233** |

- **Sorted-multiset diff of the whole log: 8 differing lines**, every one from a documented
  benign class. Six are the three `CONFIG_FILE: readonly variable` warnings rebasing into the
  new file; two are the ed25519 keygen fingerprint (#1325).
- **The §4b ORDERED comparison (#1398): SAME ORDER, 0 of 272 positions differ.** This is the
  one instrument that can see a permutation — a sorted multiset destroys order by definition,
  and neither a header count nor a log-line total can see one either. The block is contiguous
  and the source stanza sits at its exact former position, which is the property that makes
  this the cheapest cut shape; the ordered diff is what turns that from an argument into a
  result.
- The `CONFIG_FILE` rebase was **predicted before the after-log existed** and then checked:
  the moved content starts at base line 1861 under a 31-line header, so every line maps by a
  constant `-1829`, sending `run.sh:1893/1893/1925` to `test-appliance-identity.sh:64/64/96`.
  That is what came out. (Bash names the *sourced* file in these warnings, not `$0` — I had
  that backwards mid-session and the log corrected me.)
- `shfmt -i 4 -d` rc 0 and `shellcheck --severity=warning` rc 0 on the new file, run with
  **shellcheck 0.11.0**, the version CI pins — not the 0.9.0 that also exists on this box.
- `make lint` and `make test`: results in a comment below, run after the tsv row landed.

## Byte identity, re-derived from git rather than inherited

Read out of `git show`, not out of the extraction's own chunk files — doing it from git is
what makes it a proof rather than a tautology.

- The 443 moved lines are sha256 `bc564c39…` on **both** sides.
- Whole-file reconstruction: `base[1..1860]` + the 3-line stanza + `base[2305..]` hashes
  `dc7f71da…`, **equal to the committed `run.sh`**. So `run.sh` is exactly the original minus
  the block plus the stanza, with nothing else touched.
- **Control:** a one-line mutation of the copy IS detected, so the equality is a real
  equality and not a comparison that cannot fail.

## The file-budget row, and one deliberate departure from the recipe

The new file is over the 400 target, so the gate refuses the tree without a ceiling for it.
**Control run before trusting the pass:** delete the new row and the gate fails by name
("has no `docs/dev/file-budget.tsv` entry"). Its OK is not vacuous.

**This is not `--generate`'s output, deliberately.** The generator also wanted to lower four
rows this cut never touches (`dashboard/mining_dashboard/service/egress.py` 424->423,
`.github/workflows/ci.yml` 513->512, `tests/integration/e2e.sh` 766->765,
`tests/integration/run.sh` 2551->2533). `check_monotonic` compares the whole tsv against the
base, so lowering a row reds every un-rebased in-flight branch in every lane, naming a file
their diff never touched. Only the two rows this cut owns move. Verified from the gate's
source that this is legal rather than merely tolerated: the per-file check fails only when
`lines > ceiling`, so a stale-high ceiling passes.

## The one arguable member, argued rather than left to arithmetic

The control-runner units rendering into `/run` (#791) travel with identity, though
`provision_control_runner`'s ownership and foreign-install guards stay behind. It is here
because it is a consequence of the appliance's read-only root rather than of the control
channel, and because it sits inside the contiguous run — excluding it would buy a tidier
topic at the cost of the zero-reorder property. If you disagree, moving it later is
independent of this cut and costs one range.

Deliberately left behind: `preflight_remote_nodes`' dial-before-commit check, which
immediately follows the block and reads as identity-adjacent because provisioning consumes
both. It belongs to the remote-node contract.

The map's estimate for this domain was 393 against 443 moved. That is not a sign the boundary
is wrong — the map's appliance numbers are a weak prior, and three shipped cuts have already
missed it in both directions (os-update 424 vs 530, rig-miner 299 vs 492).

## What I did NOT do

- **The coupling audits (A: `$C`/`$RESULTS`/`$MASKED`/`$REQS`/`$STAGED`/`$AUDIT`; B: ambient
  `$STACK`/`$SANDBOX` `.env`/Caddyfile; cross-block functions and variables) were run in a
  previous session of this lane and are recorded in the extraction commit, not re-run by me.**
  What I re-derived in this one is the byte identity and the whole-file reconstruction. The
  after-proof is the empirical check that would have caught a coupling error the audits missed,
  and it is clean — but those are different claims and I am not merging them.
- The stanza uses the guarded `domain_ran` form, matching every other shipped domain. A bare
  `source` would have made this the only domain outside the #1400 guard.

## Shared files

`docs/dev/file-budget.tsv` — regenerated per-lane, expect a conflict rather than a merge that
takes a side.


VijitSingh97:
## Evidence at the current tip, and a change since review

Two independent non-author reviews landed on this PR at `1c6b783`. **Both found the same real defect, in the one place the header did not claim to have re-derived: the coupling audit.** It is fixed in `dbf38e2` and the numbers below are the tip's, not the reviewed commit's.

### The finding

`tests/stack/test-appliance-identity.sh` read `$WALLET` and nothing in the file assigned it. lib.sh's two sandbox builders default it (`build_val_sandbox`, `build_control_sandbox`, both `WALLET="${WALLET:-$VALID_PRIMARY}"`, neither `local`) — but this file calls neither. In the suite the read worked only because an earlier `run.sh` section happens to build a sandbox roughly 1300 lines before this file's source stanza. **Sourcing in place preserves that accident, which is why the suite was green either way — and why no instrument in this PR's proof could have seen it.** Byte identity, header order and pass totals are all blind to it by construction.

The severity is worse than "fails loud": under `set -u` an unbound expansion inside a *sourced* file kills the whole shell, so the suite would die mid-run rather than report a failure.

This corrects a claim that had been treated as settled in this lane. "`$WALLET` coupling is retired (#1305/#1337) — both builders default it" holds only for code that **calls** a builder, and a split is exactly what creates non-callers.

### The fix, and its control

One line in the authored header region, so the 443 moved lines and their `bc564c39…` are untouched:

```
source lib.sh + this file, set -uo pipefail, WITH the seed .... rc=0, 44 checks
the same file with only that one line removed ................. rc=127,
    "WALLET: unbound variable" at the ssh-validation section
```

44 matches the file's in-suite check count, so **the file is now sourceable standalone** — the #1387 contract, closed for this file as a by-product rather than as scope creep. The other reviewer confirmed the two preceding Phase-1 cuts (`test-tor-network.sh`, `test-rig-worker.sh`) both already source standalone, so this keeps the run at 3-for-3 instead of breaking it.

**One reviewer claim I checked and did not take.** A static sweep reported that three further domain files — `test-appliance-install.sh`, `test-control-deploy.sh`, `test-doctor.sh` — share the unarmed dependency. They do not: in all three, `$WALLET` appears **only inside comments**, and those comments say they deliberately use `$VALID_PRIMARY` *instead of* `$WALLET` to avoid this exact trap. The sweep counted the warning as the defect. No follow-up is owed, and the other reviewer's independent sweep ("the only one of fourteen") is the correct one.

### Full suite at the tip

Re-run in full after the change, against the same baseline:

| | baseline `c505413` | tip `dbf38e2` |
|---|---|---|
| rc | 0 | 0 |
| passed / failed | 2922 / 0 | 2922 / 0 |
| section headers | 272 | 272 |
| log lines | 3233 | 3233 |

**§4b ordered comparison against the baseline: same set, same order, 0 of 272 positions differ.**

Against the previously-proven `1956efa`, exactly 8 lines differ and the arithmetic was predicted before the run rather than explained after it: the three `CONFIG_FILE: readonly` warnings shift by **exactly +9** (`:64/:64/:96` → `:73/:73/:105`), which is the 9 header lines added, and the file's budget row moves 474 → 483 by the same 9. The other 2 are the ed25519 keygen fingerprint (#1325). Nothing else moved.

The file-budget gate was run in the real worktree, so `check_monotonic` was active rather than skipped: rc 0.

### What I ran, and what I did not

Ran here: the full tier-1 suite (above), the standalone-source proof with its removal control, the budget gate, and `shellcheck --severity=warning` (0.11.0, the CI-pinned engine) plus `shfmt -i 4 -d` on the changed file — both clean.

**Not run here: full `make lint`.** Its `lint-sh` leg is the OOM path that has taken down whole sessions on this box and is serialized across lanes; `Shell tests (shellcheck + pithead suite)` covers the same ground in CI and is the independent check this change should be merged on.

### On the two red checks

`Build image (dashboard)` and `Build image (tor)` fail on the **`Scan image for CVEs (Trivy)`** step — `docker build` succeeds in both. It is not this diff: the identical pair fails on an unrelated single-file PR opened from the same base today, and a test-file move cannot change an image input. The finding is CVE-2026-14456 (`libcrypto3` 3.5.7-r0, fixed upstream in 3.5.8-r0), and it is being handled separately under its own tracking issue. I am deliberately **not** clearing it from "the base was green" — this base has not been scanned since the Trivy DB rolled, so that would not be evidence.

Issue #1365, opened 2026-08-24
Leg-level e2e skips increment no counter, so 28 dropped legs are invisible in the summary


VijitSingh97:
The harness has two kinds of skip and only one of them is counted. Scenario-level skips increment `IT_SKIPPED` and appear in the summary. Leg-level skips print a warning and increment nothing, so they are absent from the verdict entirely — there is no column they could show up in.

There are **28** of them.

## The counter

`IT_SKIPPED` is incremented in exactly one place, `tests/integration/run.sh:433`, inside `run_scenario`:

```bash
if ! resolve_overrides "$overrides"; then
    it_warn "SKIPPED ${name}: ${SKIP_REASON}"
    IT_SKIPPED=$((IT_SKIPPED + 1))
    return 0
fi
```

Every other skip in the harness is a bare `it_warn`, and `it_warn` touches no counter at all — `tests/integration/lib.sh:31`:

```bash
it_warn() { echo -e "${IT_YELLOW}[ITEST]${IT_RESET} $1" >&2; }
```

The whole harness contains exactly three counter increments: `lib.sh:53` (`IT_PASS`), `lib.sh:57` (`IT_FAIL`), and `run.sh:433` (`IT_SKIPPED`). There is no fourth.

Counting honestly: 30 `it_warn` lines mention a skip. One is `:432` above, which is counted. One more, `:1053`, is narration that *other* scenarios will skip — those then skip through `:432` and are counted, so it is not itself a drop. That leaves **28** leg-level drops that increment nothing.

So the summary at `run.sh:1851`:

```
passed:  N
skipped: 5
failed:  0
```

`skipped: 5` is the five scenario skips from #1083. It is not a partial count of the leg skips. It excludes them completely.

## What is invisible

The 29 include the ones that matter most, because they are the ones that drop whole surfaces:

- `run.sh:2118` — "no worker exposes a RigForge enriched feed" — drops the **entire** rigforge-control phase, every leg in it, with one warning line.
- `run.sh:2134` — no workers descriptor and no `--rig-host` + `IT_RIG_TOKEN` — drops the same phase by a different route.
- `run.sh:2231` — no `IT_RIG_POOLS_PROBE` — drops the `pools` write leg (#1002b).
- `run.sh:2238` — no `.last_applied.pools` on record — drops it by a second route even when the probe is set.
[...]

Issue #1083, opened 2026-08-17
Five e2e scenarios are permanently skipped, so remote-node mode ships with no live coverage


VijitSingh97:
Every targeted e2e run — the gate the operator mandate requires before every cut — reports five skipped scenarios. They are not flaky or incidental. They are the same five every time, because the bench cannot satisfy their preconditions, so these code paths have never been exercised live by the release gate.

From the v1.19.1 pre-cut run (610 passed, 5 skipped, 0 failed):

```
SKIPPED local-full-main-secure-tari:        needs --full-data-dir (box baseline is pruned)
SKIPPED remote-main-secure-tari:            needs --remote-monero-host
SKIPPED remote-tari-main-secure:            needs --remote-tari-host
SKIPPED local-pruned-main-subnet:           network.subnet move needs a full down/up, not a hot apply
SKIPPED local-pruned-main-payout-confirm:   needs IT_MONERO_VIEW_KEY
```

Plus a sixth, narrated but not counted in the five: `skipping node-down failover (remote mode: no local monerod to stop)`.

## What that leaves uncovered

- **Remote Monero and remote Tari mode.** Two of the five. This is a shipped, documented feature (#103, since v1.14.0) and the release gate has never run it against real nodes. An operator running a node elsewhere is on a path the gate does not test.
- **Full (non-pruned) chain mode.** The bench baseline is pruned, so the full-chain scenario cannot run here at all.
- **`network.subnet` changes.** Skipped as needing a different phase, so the routine gate never covers a subnet move.
- **Payout confirmation.** Needs a real view key in the environment, which is never set.
- **Node-down failover**, whenever the run is in remote mode.

## Why the count being stable is the problem, not the reassurance

Five stable skips read as "known and fine", which is how they have survived. But a skip is not a pass — it is an absence of evidence, and here it is a *permanent* absence on paths that ship to users. The summary line puts `skipped: 5` next to `failed: 0`, which invites reading the run as clean.

It also hides regressions in the skip itself. If a sixth scenario started skipping because a precondition silently broke, the number would move from 5 to 6 and almost certainly be read as noise.

## Suggested direction

These are not one problem, so they probably are not one fix:

1. **The two remote scenarios are the ones that matter most** — a shipped feature with zero live coverage. `--remote-monero-host` / `--remote-tari-host` could point at the box's own nodes over the LAN, which would make them runnable on the existing bench without new hardware. Worth checking whether that is a faithful enough test.
[...]

Issue #185, opened 2026-06-10
Dashboard: Worker Inspect page — read/edit each miner's XMRig config over its API, with versioned config history + per-config hashrate stats


VijitSingh97:
## Summary

A dedicated **Worker Inspect** page that, for any worker exposing the XMRig HTTP API (with write access granted), lets the operator **view and edit that miner's config from the dashboard** — while Pithead keeps a **versioned history of every worker's config** in its DB and correlates **measured hashrate to each config version**.

> Marked **v1.1 / future feature** — sizable, and not needed for the v1.0 showcase.

## Motivation

- Today, changing a rig's config means SSHing into each miner / hand-editing RigForge configs. On a multi-rig setup that's the main operational pain.
- Tuning (threads, huge pages, RandomX flags, donate level, pool/algo) is iterative — you want to change a knob and *immediately see the hashrate effect*, and be able to roll back to whichever config performed best.
- Centrally-stored config copies mean a rig that dies can be reprovisioned from Pithead's last-known-good config.

## Proposed capabilities

1. **Read** — pull the live config from each worker's XMRig API (`GET /1/config`; requires `http.restricted: false` + an access token on the rig) and show it on the Worker Inspect page.
2. **Edit / push back** — edit in the dashboard and `PUT` the updated config to the worker (XMRig reloads it). Gated on write access + auth.
3. **Persist** — store every worker's config in Pithead's DB (the dashboard's SQLite), keyed by worker.
4. **Version + history** — on every observed or applied change, snapshot the config as a new version with a timestamp and a diff; keep the full per-worker config history, and log the source of each change (dashboard edit vs. changed externally on the rig).
5. **Per-config hashrate stats** — tie the hashrate time-series to the active config version, so each config carries its measured hashrate (avg / peak): *"config #3 did 5.1 kH/s, config #4 did 4.8 kH/s."* Lets the operator pick the best config empirically.
6. **Worker Inspect page** — a per-worker detail view hosting all of the above: current config, history with diffs, hashrate-per-config, and edit / apply / rollback controls.

## Technical considerations

- **XMRig API**: read/write needs `http.restricted: false` + an access token per worker — the same token/addressing plumbing as #171 / #172. The default XMRig API is read-only, so this feature requires operators to explicitly opt each rig into write access.
[...]

PR #1437, opened 2026-08-28
test(e2e): count and name every skip, so the verdict can say what did not run (#1365)

VijitSingh97:
Closes nothing by itself — see the checklist at the end.

## The defect

The harness had exactly three counter increments: `IT_PASS`, `IT_FAIL`, and the scenario skip. Every other skip left through a bare `it_warn`, which touches no counter, so it appeared in no column of the verdict.

A run that dropped the **entire** rigforge-control phase and a run that exercised it produced summaries differing by nothing but the pass count — and the pass count moves for a dozen unrelated reasons. Nothing in the output could say a surface had not run, which leaves the reader unable to tell "checked and clean" from "never ran". That is the distinction a release gate exists to make.

## What this does

Three counted helpers in a new `tests/integration/skip-accounting.sh`, each taking **what** did not run and **why**: `it_skip_scenario`, `it_skip_phase`, `it_skip_leg`. Separate buckets, because losing a whole phase is not the same size of hole as losing one leg — and a phase drop is louder in the live stream.

```
passed:  312
skipped: 2 scenarios, 1 phases, 4 legs
did NOT run:
    - scenario prune-remote — no pruned chain supplied
    - PHASE    rigforge-control — no worker exposes a RigForge enriched feed
    - leg      pools write (#1002b) — no IT_RIG_POOLS_PROBE
```

## The census, re-derived at this tip rather than taken from the issue

33 sites in `run.sh` and `rigforge-writable-keys.sh`, one more in `lib.sh`, and **a 35th the issue's own count missed**. Two corrections to the issue as filed:

- It said 28 and put the `pools` legs in `run.sh`. They have since moved to `rigforge-writable-keys.sh`, and its own body says 28 in one place and 29 in another.
- Its census was wording-based, so it missed `run.sh`'s Worker Inspect drop, which says "not exposed here" and never says "skipping". That site is a whole read/write leg.

## The guard, which is the half that has to hold

`selftest-skip-accounting.sh` checks the counters and the **real** `summary()` extracted from `run.sh` (extracted, not re-spelled — a re-implementation passes happily while the shipped file prints something else). Then it censuses every harness file by two rules:

1. skip **wording**;
2. the **shape** of a warning followed immediately by `return 0`.

Rule 2 is what found the 35th site, and the file asserts that rule 1 alone would have missed it. Both rules carry a positive **and** a negative control, and the one legitimate warn-then-return (`--keep:`, a deliberate cleanup) is allowlisted by name so a new one has to be looked at by a human.

## Proof it can fail

Six mutations of the shipped code, each demonstrated red:

| mutation | result |
|---|---|
| revert a converted site to a bare `it_warn "skipping …"` | census/wording reds |
| revert the #185 site to its wordless warn-then-return | census/shape reds |
| `summary()` stops reporting the leg bucket | 3 assertions red |
| `summary()` stops naming what did not run | reds |
| `it_skip_phase` loses its increment | 2 assertions red |
| a phase drop stops being louder than a leg drop | reds |

One mutation appeared to survive. It was re-run and confirmed to be an **ineffective** mutation — the `sed` targeted `run.sh` for a site that lives in `rigforge-writable-keys.sh`, so nothing changed. An ineffective mutation reads exactly like a surviving one; applied to the right file it reds.

Separately, sourcing `lib.sh` by a bare name would have resolved the new source to `lib.sh/skip-accounting.sh` and left the helpers undefined — a silent miscount. Every current call site uses a path, so this never fired, but it is now fail-closed and the guard was observed to fire.

## Shared and budgeted files

- `docs/` — the doc update ships with the change, per CLAUDE.md. `docs/` is in `_shared.paths`; no `.lane-override` used.
- **No `Makefile` change**: `test-integration-selftest` globs `selftest*.sh`, so the new self-test is picked up.
- **No `docs/dev/file-budget.tsv` change.** `lib.sh` sat on its ceiling (692) and that ratchet only moves down, so the addition moved out to its own file instead of growing it. `assert_mining_state` moved with it — its whole contract is a conditional skip and it is now a caller of `it_skip_leg`. `lib.sh` goes **692 → 685**; the recorded ceiling is deliberately left alone rather than ratcheted down, to avoid a conflict on a file that is regenerated per-lane.

## What was run

All green at `f664870`:

- all 7 integration self-tests — **408 assertions, 0 failed** (was 368)
- `shellcheck 0.11.0` (the version CI pins) over `tests/integration/*.sh`; repo-wide `shfmt -i 4 -d`
- `lint-file-budget`, `lint-docs-voice`, `lint-md`, `lint-operator-strings`, `lint-topology`, `lint-yaml`

## What was NOT run, and why

- **`make lint-sh` in full.** It is the serialized OOM path that has killed whole sessions on this box, and a peer review agent is resident. Every file this diff touches was linted with the same shellcheck binary and flags CI uses, so the marginal information is zero. CI will run it.
- **No live e2e run.** This changes what the harness *reports*, not what it *does*. Nothing parses the summary text — checked across `e2e.sh`, `scripts/`, `.github/workflows/` and `docs/dev/releasing.md`.
- **The new counts have never been seen on a real run**, only on simulated state. The first live gate run is where the numbers get their first real exercise.

## The over-engineering pass, and what it cut

Run on this diff before opening the PR. One finding in my own work, acted on: a "the shipped summary reads the leg counter at all" assertion that was a **grep over the source text**, not behaviour — the exact shape this project files bugs about. It is cut. The mutation it claimed to cover (`summary()` stops reporting the leg bucket) is still caught, by two behavioural assertions, re-verified after the cut.

Kept, with the reasoning stated so it can be argued with:

- **Three buckets, not one.** A single count re-creates the "`skipped: 5` reads as wallpaper" problem #1083 describes, and the issue asks for a phase drop to be louder than a leg drop.
- **The two-rule census.** The shape rule is not belt-and-braces — it found the 35th site, and the self-test asserts the wording rule alone would have missed it.
- **The fail-closed source guard**, three lines for a case that does not occur today. The failure it prevents is a silent miscount, which is the defect class this whole PR is about.

## Follow-ups, not done here

- The issue also suggests making a leg skip a **hard failure** when its prerequisite should be present (a rig is attached, so a missing probe is a harness bug, not an environment). That is #1236's ask for its two legs and is deliberately out of scope here — this PR makes the drops visible, it does not change any verdict.
- This sharpens #1083: `skipped: 5` was never the number of things that did not run, only the number that were counted.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -473,6 +473,34 @@ On a scenario failure, the harness captures (redacted) to `results/<scenario>/`:
 `api-state.json`, and `logs.txt` (last 200 lines per service). The end-of-run summary lists
 each failed assertion and points at these.
 
+### Reading the verdict — what did not run
+
+The summary counts three kinds of skip separately and names every one:
+
+```
+passed:  312
+skipped: 2 scenarios, 1 phases, 4 legs
+did NOT run:
+    - scenario prune-remote — no pruned chain supplied
+    - PHASE    rigforge-control — no worker exposes a RigForge enriched feed
+    - leg      pools write (#1002b) — no IT_RIG_POOLS_PROBE
+```
+
+A scenario skip drops one matrix row; a phase skip drops every leg under it with one line; a leg
+skip drops one assertion group. They stay in separate buckets because they are not the same size
+of hole, and a phase drop announces itself more loudly in the live output.
+
+Until [#1365](https://github.com/p2pool-starter-stack/pithead/issues/1365) only the scenario
+skips were counted. A run that dropped the whole rigforge-control phase and a run that exercised
+it produced summaries differing by nothing but the pass count, and the pass count moves for a
+dozen unrelated reasons. The output had no way to say a surface had not run, which leaves the
+reader unable to tell "checked and clean" from "never ran" — the distinction the gate exists to
+make.
+
+Every skip leaves through `it_skip_scenario`, `it_skip_phase` or `it_skip_leg`
+(`tests/integration/skip-accounting.sh`). Plain `it_warn` stays for warnings that are not a
+dropped check.
+
 ---
 
 ## The self-test (CI)
@@ -482,6 +510,11 @@ rendering and value typing, expectation derivation (profile gating), secret reda
 SSH/local exec wrapper, JSON parsing, and matrix axis coverage. It runs in CI on every PR (the
 `shell` job) and via `make test-integration-selftest`, so the harness itself is held to the
 same lint/test standard as the rest of the stack.
+
+Several self-tests sit beside it as standalone files, picked up by the same globbed target.
+`selftest-skip-accounting.sh` is the one that keeps the skip accounting honest: besides checking
+the counters and the real `summary()`, it censuses every harness file and fails if a skip leaves
+through a bare `it_warn` — by wording, and by shape for the drops that never say "skipping".
 
 ---
 
--- a/tests/integration/lib.sh
+++ b/tests/integration/lib.sh
@@ -44,11 +44,19 @@ redact() {
 }
 
 # --- Assertions -------------------------------------------------------------
-# Counters are global so the runner can total them across scenarios.
+# Counters are global so the runner can total them across scenarios. The skip buckets and the
+# three helpers that move them live next door; sourced here so anything that has lib.sh has them.
+# Fail CLOSED: sourced as a bare name this resolves to "lib.sh/skip-accounting.sh", and a
+# harness that merely warns would then miscount every skip in silence — the exact defect below.
+# shellcheck source=tests/integration/skip-accounting.sh
+source "${BASH_SOURCE[0]%/*}/skip-accounting.sh" ||
+    {
+        echo "lib.sh: skip-accounting.sh must sit beside it (source lib.sh by a path, not a bare name)" >&2
+        exit 1
+    }
 IT_PASS=0
 IT_FAIL=0
 IT_FAILED_NAMES=""
-
 it_pass() {
     IT_PASS=$((IT_PASS + 1))
     printf '    %b✓%b %s\n' "$IT_GREEN" "$IT_RESET" "$1"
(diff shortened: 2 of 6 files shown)


VijitSingh97:
Merging at head `71e2fe7` with two checks red. Naming them rather than merging quietly.

**Green, and these are the ones this PR is accountable for:** `Shell tests (shellcheck + pithead suite)`, `Fake-daemon mini-stack (docker)`, `Dashboard tests (pytest + coverage)`, `Lint per-surface`, `Python lint + format (ruff)`, `Secret scan (gitleaks)`, `Workflow audit (zizmor)`, `Dockerfile lint (hadolint)`, `Compose config + security hardening`, and the three other `Build image` jobs.

**Red, and not this PR's:** `Build image (tor)` and `Build image (dashboard)`. Both fail inside `Scan image for CVEs (Trivy)` on a single upstream finding, CVE-2026-14456 (openssl, QUIC-server DoS). I read both job logs rather than assuming they shared a cause: the tor image is alpine 3.24.1 (`libcrypto3`/`libssl3` 3.5.7-r0) and the dashboard is debian 13.6 (`libssl3t64`/`openssl`/`openssl-provider-legacy` 3.5.6-1~deb13u2) — different bases, same one CVE, 2 and 3 rows respectively.

This is filed and analysed in #1438 (and duplicated by #1427), which establishes with digest-level evidence that no pin bump can clear it today — the fixed packages are published in the distro repos but no rebuilt base image carries them. The recorded decision there is a `.trivyignore` acceptance with a clearing condition.

This diff contains no Dockerfile, no dependency and no image content — `tests/integration/*` plus `docs/dev/integration-testing.md`. #1435 merged past the same two failures earlier today, and the target branch has no required status checks.

**One late fix in `71e2fe7`:** `run.sh:637`'s comment still pointed at `lib.sh` for `assert_mining_state`, which this PR moved to `skip-accounting.sh`. Caught by grepping the file rather than trusting the move; I swept the other four `lib.sh` pointers in the same files and they are all still accurate.

**What was NOT run:** `make lint-sh` was not run in full — it is the OOM path on this box and is serialized fleet-wide. Substituted a targeted `shellcheck -x` from the repo root over the five changed harness files: clean apart from pre-existing info-level SC1091/SC2016/SC2029. CI's own shellcheck job is the binding one and it is green.

Issue #1445, opened 2026-08-28
docs: testing-strategy.md's coverage table points every tier-1 shell row at run.sh, which no longer holds them


VijitSingh97:
`docs/dev/testing-strategy.md`'s coverage table names the file that runs each behaviour. For tier-1 shell behaviours it says `tests/stack/run.sh` — **14 times, and it names no domain file at all**, while the #1105 Phase 1 split has already moved roughly two dozen domains out of `run.sh` into `tests/stack/test-*.sh` files beside it.

So a reader following the table to the code lands in the wrong file for any behaviour that has already been cut. Examples, each already living in its own domain file while the table still points at `run.sh`:

- the post-reboot `rolled_back` verdict (#1051)
- the dashboard OS-update verbs (already `test-appliance-os-update-verbs.sh`)
- the restore-at-setup spool contract (#909)

## Why this is being filed rather than fixed in a split PR

`docs/` is shared, and a table-wide rewrite is a doc rewrite rather than a doc update that ships with a change — it wants routing, not a widened split diff. The staleness is also **systematic and pre-existing**: it predates every cut in the current split series rather than being introduced by any one of them, so no single cut's PR is the right place to absorb it.

## What a fix looks like

Either point each row at the domain file that now holds it — accurate, but it takes on maintenance at every future cut until the split finishes — or state once, at the head of the table, that `tests/stack/run.sh` names the tier-1 shell **suite** and that a behaviour's domain file is found beside it. The second is cheaper and stays true while the split is in flight; the first is better once it lands.

Worth deciding which before the remaining cuts, since each one adds another stale row either way.

## Verification

- `grep -c 'tests/stack/run.sh' docs/dev/testing-strategy.md` → 14
- `grep -c 'test-appliance' docs/dev/testing-strategy.md` → 0
- `docs/dev/repo-map.md` is **not** affected: it describes the split generically ("each `test-*.sh` file beside it is a domain already out") and stays correct as the split proceeds.

VijitSingh97:
**Controller routing decision (2026-08-28).** Two-stage, matching the split's own life cycle:

1. **Now, while the #1105 split is in flight:** fix this once at the head of the coverage table — a short note stating that tier-1 shell rows name the suite entry point (`tests/stack/run.sh`), which sources the split-out domain files. Re-pointing all 14 rows today means re-editing the table at every remaining cut against a moving map, and a row that lags one cut behind is wrong in a quieter way than the current state.
[...]

PR #1447, opened 2026-08-28
test(stack): split the appliance-boot domain into its own file (#1105 Phase 1, develop-v2)

VijitSingh97:
Cuts the appliance **boot** domain out of `tests/stack/run.sh` into `tests/stack/test-appliance-boot.sh` (#1105 Phase 1, develop-v2 lane). Eleven sections, no moved line edited.

`run.sh` 3400 → 2819. New file 641 lines (57 authored header + blank + 583 moved). This honours the leave-behind #1368 promised to this cut in writing: `pithead-boot` wiring, the A/B commit gate reading `doctor --json` (#852), the boot health probe (#1140), the self-reboot on a failed health gate (#1065), the appliance battery's release gate (#1064), and `pithead-sync`'s rigforge leg.

## THIS CUT REORDERS — read this part

The block is **not contiguous**. It is three parts, and what sits between them is what makes this cut different from the last four:

- part 1 — the three `load_baked_images` sections (#798)
- *(between them: the `source` stanzas for `test-appliance-install.sh` and `test-appliance-rig-miner.sh`)*
- part 2 — `pithead-boot` wiring, the A/B commit gate (#852), the boot health probe (#1140), `os_update_rollback_verdict` (#1051), `revenue_container_verdict` (#852), `pithead-sync`'s rigforge leg
- *(between them: the kernel-hugepages, RAUC stale-tarball, mkbundle, config-reset and factory-reset sections)*
- part 3 — the self-reboot on a failed health gate (#1065), the appliance battery's release gate (#1064)

**The `source` stanza sits at part 2's former position**, so part 2 — six sections, the largest group and the one carrying the boot-order contracts — executes exactly where it does today. **Two parts move, five sections in total:**

| moved | sections | displacement | now runs |
|---|---|---|---|
| part 1 | 3 | **+14 positions** | after the install and rig-miner domain files instead of before them |
| part 3 | 2 | **−12 positions** | before the kernel/RAUC/mkbundle/reset block instead of after it |

Those displacements are measured from the two suite logs, not estimated. The ordered comparison reports **31 of 272** header positions differing, and the 31 decompose exactly, with no residue:

| group | sections | displacement | why |
|---|---|---|---|
| part 1 | 3 | +14 | crosses the install + rig-miner domain files, which contribute 14 sections |
| the install + rig-miner sections | 14 | −3 | the 3 sections that vacated ahead of them |
| part 3 | 2 | −12 | crosses the kernel/RAUC/mkbundle/reset block, which contributes 12 sections |
| that block | 12 | +2 | the 2 sections now preceding it |

3 + 14 + 2 + 12 = 31. Every displacement equals the size of the group it crossed, and the four groups account for every differing position — an unintended movement would show up as a fifth group.

The alternative anchor (part 1's position) was rejected: it displaces eight sections instead of five, and it would carry part 2's live names (`DASHBOARD_SECURE`, `HOST_PORT`) across the two domain files.

**Why the reorder is safe, argued rather than asserted:**

- **Part 1 leaks nothing.** Every variable it sets and all five helper functions it defines (`lbl`, `sha_of`, `rbl`, `mk_store`, `hbl`) are unset as its own sections end, so neither domain file it now runs after can inherit anything from it.
- **Part 1 reads nothing those domain files assign.** Its only ambient reads are `$PATH` and a defaulted `${PODMAN_IMAGE_PRESENT:-...}`. Every `PATH=` in both domain files is a *command-prefix* assignment (`PATH="$X/bin:$PATH" cmd ...`) or sits inside a function body — none replaces this shell's `PATH`.
- **Part 3 shares no name with the block it now precedes, in either direction.** It reads only `$ROOT` and `$SANDBOX` and writes only inside its own subdirectory; nothing in the kernel/RAUC/mkbundle/reset block reads what part 3 leaves set, and part 3 reads nothing that block assigns.
- No function name in the moved block collides with any other file in `tests/stack/`.

The **ordered header comparison** of the two suite logs is the instrument that actually covers this, and its result is in the evidence below. A sorted multiset destroys order by definition, a header count cannot see a permutation, and a log-line total cannot either — all three would pass a cut that shuffled the suite arbitrarily.

## A defect this cut found in its own derived boundary

The boundary recorded for this cut was a single contiguous range covering parts 1 and 2 together. **It was wrong, and every instrument in the split recipe was blind to it.** The two `source` stanzas listed above sit inside that range, so extracting it whole would have moved `test-appliance-install.sh` and `test-appliance-rig-miner.sh`'s stanzas into the new file — making a domain file source two sibling domains. Execution order and pass totals would have been unchanged, so byte-identity, the sorted multiset, the header count, the log-line total **and the ordered comparison** would all have passed it.

The generalisation, for the cuts still queued: a hazard scan looks for **names**; `run.sh` also contains **structure** — the stanzas of already-shipped cuts. Grep every candidate range for `source "$HERE/` before extracting.

## Evidence — what was RUN

- **Byte-identity, re-derived from git rather than from the extraction — CORRECTED at review.** Every **non-blank** moved line is byte-identical: rebuilding the moved text from `git show <base>:tests/stack/run.sh` at the three ranges git itself reports (`diff -U0`: 1928-2082, 2089-2417, 2721-2819 = 583 lines) and comparing against the new file below its header (also 583 lines) leaves **no difference in any non-blank line**, checked with a positive control confirming the comparison can fail. **Blank-line placement differs in exactly two places, and the original wording of this bullet — a flat "empty" — was wrong.** The delete ranges are asymmetric: part 3's range ends with its trailing blank, part 2's does not, so the extractor authored one blank line at the part-2/part-3 seam and dropped one at EOF. Net 583 = 583, which is why the line arithmetic closed over it. Both are inert in bash (`bash -n` clean; the file executes all eleven sections), so this is a wording defect in the evidence, not a defect in the cut. Worth stating because reconstructing at the *extractor's own chunk ranges* reproduces the extractor's seam decisions and returns "empty" by construction — the check has to use the ranges **git** reports to be a proof rather than a tautology.
- **Full suite after-proof**, rc=0: **2922 passed, 0 failed** — identical to the baseline, as are the 3233 log lines and 272 section headers. The sorted multiset of the whole normalised output differs by exactly **two** lines, the two halves of one ed25519 keygen fingerprint (#1325, known nondeterminism). The header multiset is an identical set.
- **The CONFIG_FILE shift was predicted before the run and checked after: zero.** Those three `readonly variable` warnings are emitted by `test-appliance-identity.sh`, already a domain file of its own, so no cut inside `run.sh` can move them — 3 before, 3 after, same line numbers.
- **Baseline reuse justified rather than assumed.** `~/split-baselines/after-wallet-seed` was taken before #1437 and #1440 merged. `tests/stack/run.sh` is byte-identical across that range, and of the eight files those merges changed only the two Dockerfiles reach this suite, through three guards in `test-harness-tooling.sh`. All three were re-run at this tip: rc=0, and each call site is `>/dev/null 2>&1` followed by one `assert_rc`, so each contributes two log lines either way. `tests/integration/*` is referenced by no `tests/stack/` file, and the only docs the suite reads are `architecture.md` and `configuration.md`, neither changed.
- **Standalone-source check (the #1387 contract), with a firing control.** `source lib.sh; source test-appliance-boot.sh` under `set -uo pipefail` returns 0 and really executes all eleven sections (11 section headers in its output — the line total varies by one across runs, the known elapsed-seconds heartbeat noise, so the header count is the stable check). The control fires: injecting one unseeded read into a copy gives **rc=127 `unbound variable`** through the identical wrapper. So the new file needs no seeding, proven rather than assumed.
- **File-budget gate, with a firing control.** Deleting the new row makes the gate fail **by name** (`... is 641 lines (> 400 target) but has no docs/dev/file-budget.tsv entry`), rc=1; restored, rc=0.
- `shfmt -i 4 -d` and `shellcheck --severity=warning -x` clean on the new file.

**`docs/dev/file-budget.tsv` carries only this cut's two rows.** `--generate` also wanted to lower five rows this diff never touches (`egress.py`, `.github/workflows/ci.yml`, `tests/integration/{e2e.sh,lib.sh,run.sh}` — the last two freshly changed by #1437). Those were hand-restored to their committed ceilings: lowering a row you did not touch reds every un-rebased branch in every lane, naming a file their diff never touched, and a stale-HIGH ceiling passes the per-file check by design.

## Shared file touched: `Makefile` (disclosed)

`lint-sh` runs **two** shellcheck invocations, and the second deliberately paired `tests/stack/run.sh` with `os/overlay/pithead-boot`: `gate_ready` and `os_update_rollback_verdict` read three globals their caller sets for them, and shellcheck reports those as SC2034 unless the reader is named on the same command line as the setter.

**This cut moves those callers**, so after it `run.sh` names neither `pithead-boot` nor any of the three globals — the pairing had become false, and its comment said run.sh sources a file it no longer does. `os/overlay/pithead-boot` therefore moves up to the invocation that lints the domain files, where its readers now live; `run.sh` is linted alone. **The file set is unchanged and nothing is listed twice**, which is what that target's design comment requires.

Verified both directions rather than assumed: the new domain file alone reports the three SC2034; beside `pithead-boot` it reports none; `run.sh` alone is now clean. The whole first invocation was then run with `pithead-boot` added — exit 0, peak RSS ~1.6 GB, well under the 6 GB cap the local wrapper enforces after #1206/#1333.

The alternative — three `# shellcheck disable=SC2034` comments in the new file, matching the convention in `test-appliance-identity.sh` and `test-appliance-media.sh` — was rejected because it would have left the Makefile's pairing and its comment standing but false.

## Left behind, deliberately

The host-installer cluster — uninstall keeping the operator's files and `install.sh`'s host gate (#77 phase 1), and download verification failing closed (#868). Those read names with no owner left in `run.sh`; they survive only because a sibling domain file happens to build a sandbox earlier in the run. They belong to the standalone-source sweep (#1387), not to this cut.

## What I did NOT do

- No docs change. `docs/dev/repo-map.md` describes the split generically and stays correct. `docs/dev/testing-strategy.md` cites `tests/stack/run.sh` for behaviours this cut moves — but it names **zero** domain files while ~24 are already split out, so that staleness is systematic and pre-existing across every prior cut, not this one's debt. Filed as #1445 rather than widened into this diff.
- The reorder's safety argument is a coupling analysis plus the ordered comparison. It is not a claim that the moved sections were re-run in isolation against the sections they now neighbour.


VijitSingh97:
## Adversarial review — no blocking findings, one evidence correction (already applied to the body)

Reviewed against seven attack lines. This account cannot approve its own pull request, so the review lands as a comment; `develop-v2` is unprotected, so nothing is being routed around.

**Merging on:** `Shell tests` green (10m52s), every other check green, `MERGEABLE`.

### What was re-derived, not relayed

- **Byte-identity — one real correction.** Rebuilding the moved text from `git show <base>:tests/stack/run.sh` at the three ranges *git itself* reports (`diff -U0`) and comparing against the new file below its header shows **every non-blank line identical**, with a positive control proving the comparison can fail. But it is **not** the flat "empty" the body claimed: blank-line placement differs in exactly two places, because the delete ranges are asymmetric — part 3's ends with its trailing blank, part 2's does not, so the extractor authored one blank at the part-2/part-3 seam and dropped one at EOF. Net 583 = 583, which is exactly why the line arithmetic closed over it. Inert in bash. **The body has been corrected.** The generalisable part: reconstructing at the *extractor's own chunk ranges* reproduces the extractor's seam decisions and returns "empty" by construction. Use the ranges git reports, or the check is a tautology.
- **No sibling-domain stanza swallowed.** The two `source` stanzas for the install and rig-miner domains sit in the *middle* of the deleted range and are still in `run.sh`, absent from the new file; both sibling files are untouched by the diff. This was the hazard that nearly went wrong, and it is genuinely closed.
- **The Makefile pairing, run rather than argued.** The new file alone reports 3× SC2034 (`BOOT_DOCTOR_JSON`, `OS_INFLIGHT`, `OS_STATE_DIR`); alongside `os/overlay/pithead-boot`, zero. Both real invocations execute clean. File-set union unchanged, nothing duplicated or dropped, and the rewritten comment is now true.
- **The reorder closes with no residue.** Expanding every `source` stanza in execution order across both refs gives 272 headers and **exactly 31** differing positions, in two contiguous bands matching the claimed `3+14+2+12` decomposition. No unaccounted fifth group. Safety checked in both directions: the relocated `load_baked_images` sections use private `mktemp -d` sandboxes and unset everything they define; the boot-health/release-gate sections write only into namespaced subdirectories, and no test anywhere in `tests/stack/` counts entries in the ambient sandbox — so neither coupling rule is implicated.
- **Standalone-source contract with a firing control:** 102 pass / 0 fail, all eleven sections, and an injected unseeded read trips `unbound variable` through the identical wrapper.
- **File budget:** exactly two rows changed, each equal to a real `wc -l`; no untouched row altered, which is what would otherwise redden other lanes' un-rebased branches.

### Stated rather than trusted away

The review's own full-suite run did not finish inside its window, so the 2922/0 tally was **not** independently reproduced by the reviewer — CI's `Shell tests` pass is what discharges it, and that is the arbiter being relied on here.

Issue #1452, opened 2026-08-28
lint-file-budget's monotonic-ceiling check silently skips on every CI pull-request run — the shallow checkout can never resolve a base ref


VijitSingh97:
`scripts/lint-file-budget.sh`'s monotonic-ceiling check never runs on pull requests: `resolve_base_ref()` (lines 37–63) tries `origin/develop-v2`, `develop-v2`, `origin/develop`, `develop`, and CI's checkout resolves none of them. The `lint-surfaces` job checks out with `actions/checkout` and only `persist-credentials: false` set (`.github/workflows/ci.yml:496–498`), so the default `fetch-depth: 1` applies and a `pull_request` event materialises a single detached merge-ref SHA — no branch refs at all. `check_monotonic()` then takes its resolve-failure path (lines 204–208): a NOTE to stderr and `return 0`, and the gate prints `file budget OK`. The note's own text says "This never happens in CI." It happens on every CI pull-request run.

Reproduced both ways (2026-08-28):

- **CI-shaped arm:** clone stripped to a detached HEAD with no branch/remote refs, a synthetic ceiling raise committed (`dashboard/mining_dashboard/config/config.py` 717 → 900 in `docs/dev/file-budget.tsv`): all four candidates fail `git rev-parse -q --verify`; the script prints the skip NOTE and exits 0. **A raised ceiling is reported green.**
- **Positive control:** same commit in a full clone where the base refs resolve: `file-budget: FAIL — … raises … ceiling from 717 to 900. Ceilings only go down.` rc=1.

So the mechanism is correctly designed and fires whenever a base ref exists — the defect is purely that CI's checkout never provides one. The other half of the gate (per-file ceiling enforcement) is unaffected and does run in CI. On `push` to `develop` the base trivially resolves as HEAD itself, which makes the check a no-op there too, not a safety net.

Consequence: every "green" PR run has proven only the per-file half. Ceiling raises are currently caught by developer machines running `make lint` locally (where base refs resolve) — a gate that fires only outside CI, which is this repo's recurring false-green shape.

Fix options, in preference order:
1. Add an explicit fetch of the base branches before the lint step in `lint-surfaces` (e.g. a `git fetch origin develop-v2 develop` run step, or `fetch-depth: 0` on that job's checkout — the repo is small enough that full history on one lint job is cheap).
[...]

PR #1453, opened 2026-08-28
test(stack): split the appliance-reset domain into its own file (#1105 Phase 1, develop-v2)

VijitSingh97:
Moves the appliance-reset domain out of `tests/stack/run.sh` into `tests/stack/test-appliance-reset.sh`,
which `run.sh` now sources. `run.sh` 2819 -> 2602. Five sections travel: config-reset, factory-reset, and
`pithead-data-reset` x3.

No moved line is edited. The only authored content in the diff is the new file's header and one seeded
variable, and the paragraph below on `$WALLET` is the reason that variable exists.

## The cut

One contiguous range, so the `source` stanza sits at the exact position the block occupied and the
execution order of the suite does not change at all. There is no seam to author and no anchor to argue —
a non-contiguous cut has to choose which vacated position gets the stanza and then defend the crossings
in both directions; this one has no crossings.

Git reports the whole change to `run.sh` as a single hunk, `@@ -2166,219 +2166,2 @@`: 219 content lines
out, a 2-line stanza in, at line 2166.

## `$WALLET` is seeded in the header, and that is the interesting part

The config-reset section writes a `config.json` and reads `$WALLET` to do it. Nothing in the moved block
calls either sandbox builder — `build_val_sandbox` and `build_control_sandbox` appear nowhere in the 219
lines. `lib.sh` assigns `WALLET` only at `:207` and `:221`, both **inside** those two builders, as
`WALLET="${WALLET:-$VALID_PRIMARY}"`.

That defaulting fix retires the coupling for code that CALLS a builder and does nothing for code that
does not. In `run.sh` this block was a non-caller that worked anyway, because an earlier section had
called a builder long before it and left the name set. **Sourcing the block in place preserves that
accident**, which is why the suite is green either way and why no pass total, header count or byte
comparison could have revealed it — all three are blind to it by construction. Cutting the block into its
own file is what exposes it, so without the seed the file fails the standalone-source contract (#1387)
under `set -u`.

Verified with a removal control rather than asserted: with the seed, the file alone gives rc 0 and
44 passed / 0 failed across all five sections; removing only that line — the file's line count drops by
exactly one — gives rc 1, `WALLET: unbound variable`.

The seed defaults, so an ambient value still wins, and it falls back to `lib.sh`'s checksum-valid dummy.
No real address goes into a tracked file.

Two names in the same scan look like the same hazard and are not. `$VALID_TARI` sits on the same `printf`
as `$WALLET`, but `lib.sh` assigns it at top level (`:133`), so it is genuinely ambient. `$REPAIR_LOG`
reads as unassigned to any scan anchored at line-start, because it is the **second name on an `export`
line** inside the block; every read of it here is defaulted in any case. An assignment scan anchored at
line-start loses every name after the first on an `export`/`local`/`readonly` line — worth stating,
because that is a defect shape this split has hit before.

## The two coupling rules, checked rather than assumed

Both rules this split has paid for are discharged by the block's own name list.

**Results-dir coupling.** A section whose `apply`/`run_pending`/control-write touches `$C`, `$RESULTS`,
`$MASKED`, `$REQS`, `$STAGED` or `$AUDIT`, or writes result files into a directory a still-in-`run.sh`
test counts, has to stay behind. The moved block reads none of those six names — the full read set is
`CR DR DRP DRP_DATA_LABEL DRP_ESP_LABEL DRW FR PATH REPAIR_LOG REPAIR_STATE ROOT SANDBOX STACK VALID_TARI
WALLET` plus in-block locals.

**Shared-file ambient.** A section whose write to the ambient `$STACK`/`$SANDBOX` `.env` or `Caddyfile` a
later still-in-`run.sh` test reads has to stay behind. This block writes only under `$SANDBOX/config-reset`
and its own throwaway deployment directories — its own subtree, never the ambient files.

The `shellcheck` pairing was checked too, not assumed from directory membership. The new file lands on
`lint-sh`'s first invocation, which already names `os/overlay/pithead-data-reset` — the file these
sections source and take `boot_disk_part`, `data_reset_decision` and `record_wipe` from. That is the
pairing this block needs, so no `Makefile` change is required. Confirmed from the expanded command
(`make -n lint-sh`), and the Makefile's own comment at `:69-75` documents why `run.sh` is now linted alone.

## What was run

**Full suite, before and after.** After-proof rc 0, **2922 passed / 0 failed**, 272 section headers,
3233 log lines — identical to the baseline on every one of those numbers.

**The ordered comparison (#1398), which is the check nothing else covers.** A sorted multiset destroys
order by definition, a header count cannot see a permutation, and a log-line total cannot either — all
three report exactly what they report today for a cut that shuffled the suite arbitrarily. Run against
the two suite logs, which record what actually executed:

- same header set: identical
- same header **order**: identical
- positional comparison: **0 of 272 positions differ**

**Whole-output sorted-multiset diff: exactly 2 differing lines**, both halves of one ed25519 keygen
fingerprint — the known nondeterministic class (#1325). The `CONFIG_FILE: readonly variable` line shift
was predicted to be **zero** before the run and was zero: those warnings are emitted above the cut, so
nothing below them moves them.

**Byte-identity, re-derived at the range GIT reports, not the range fed to the extractor.** Rebuilding
from the extractor's own chunk ranges reproduces the extractor's decisions and returns empty *by
construction*; that tautology shipped once in this series and is not repeated here. Read the `@@` hunk,
rebuild `git show <base>:tests/stack/run.sh` at 2166-2384, diff against the new file below its header:
empty, including blank lines. Rebuild the expected `run.sh` the same way: empty. **Each direction carries
a positive control that fires** — perturb one line and the comparison fails — so an empty result means
the comparison looked, rather than that it could not see.

**Lint.** `make lint` green (rc 0) on this branch rebased onto the tip. The header was then trimmed
once more in the over-engineering pass — comment lines in that one file, nothing else — and rather than
re-run the whole target for a comment change, `shfmt -i 4 -d` and `shellcheck --severity=warning` were
re-run on that file with the same flags the Makefile uses, plus the standalone run and its removal
control. CI's `Shell tests` job runs `make lint-sh` over the final tree and is the check on that. `shfmt -i 4 -d` and `shellcheck --severity=warning` clean on the new file,
with `shellcheck` 0.11.0, the version CI pins — not the 0.9.0 that is first on `PATH` on some boxes here.
The new file is covered by `lint-sh`'s glob, confirmed from `make -n lint-sh` rather than assumed from
directory membership.

**No `Makefile` change and no `.lane-override`.** `tests/stack/test-*.sh` already shares a shellcheck
invocation with `os/overlay/pithead-data-reset`, which is what the moved sections source.

## The budget row

The only `file-budget.tsv` change is `tests/stack/run.sh` 2819 -> 2602 — its own ceiling, lowered because
this cut changed that file. `--generate` rewrites every row from the current tree and wanted to lower **four**
rows on files this cut never touches — `dashboard/mining_dashboard/service/egress.py`,
`.github/workflows/ci.yml`, `tests/integration/lib.sh` and `tests/integration/run.sh`. Those were
restored by hand. (An earlier draft of this body said five, carried forward from a previous cut's
notes without re-deriving it against this tree. Re-derived here: four.) Lowering a row you did not touch reds
every un-rebased branch in every other lane, naming a file their diff never touched.

**The new file gets no row, and that is correct rather than an omission.** The generator drops any file at
or below `TARGET_LINES=400` and this one is 255 lines; `lint-file-budget.sh:160` **fails** a file at or
below the target that still carries a ceiling. So the usual control for this step — delete the new row and
watch the gate fail by name — does not apply to this cut, and is not claimed. What was checked instead:
the file has no row (0 matches) while `tests/stack/run.sh` does (1 match), so the search was working.

**Disclosure, and it limits the claim above: the monotonic half of that gate does not run in CI (#1452).**
`lint-file-budget` runs only in `lint-surfaces`, which checks out shallow, so no base ref resolves and
`check_monotonic` returns 0 without comparing anything. The budget monotonicity here is proven by a local
`make lint` in a full clone and by nothing on CI.

## Found at review, in the only authored content

Two accuracy defects in the new file's header, both fixed here:

- it credited `lib.sh`'s `ok` and `bad`, and the block calls **neither** — 0 and 0, against a firing
  control on the same file: `assert_eq` occurs 27 times in the same block
- it omitted the block's largest external dependency: four `source "$ROOT/os/overlay/pithead-data-reset"`
  calls supplying `boot_disk_part`, `data_reset_decision` and `record_wipe`. The sections source that file
  for themselves before each call, so it is self-contained and not a coupling hazard — but the header
  closed with "Every other name is assigned here," which those three names contradict.

Neither changes behaviour. Both are in a file whose whole job is to make a domain's dependencies visible,
where a wrong dependency list is the defect.

The over-engineering pass then took out a third thing: a paragraph narrating the cut itself — that the
block was contiguous, that nothing reordered, why a non-contiguous cut would have had to argue an anchor.
All true, all belonging to this PR and the commit message, none of it belonging to a file someone opens
in six months to find out what the reset domain tests. It is the same family as the standing rule against
putting a line number or a count in a header: provenance in a file goes stale by irrelevance. The
`$WALLET` paragraph was cut down the same way, keeping only what a future editor needs in order not to
delete the seed as redundant, and losing the commentary about which instruments would have been blind to
it — which is a claim about the proof, and lives here instead.

## Shared file touched

`docs/dev/testing-strategy.md` — a note at the head of the scenario catalog saying that a tier-1 shell row
naming `tests/stack/run.sh` means the tier-1 shell **suite**, and that a behaviour's domain file is found
beside it. That is #1445's cheaper option, and it rides this cut by the controller's routing rather than
because this cut caused the staleness: the table names `run.sh` 14 times and no domain file at all, which
predates every cut in this series. The rows are deliberately not re-pointed per cut — that goes stale again
at the next one. `docs/` is shared and needs no `.lane-override`.

## Not done, stated rather than left to inference

- The after-proof was launched against the tree **before** the rebase onto `develop-v2`. That is still
  valid evidence for this cut: everything `develop-v2` moved in between is under `tests/integration/` and
  `docs/`, and no file under `tests/stack/` changed — checked against the merge-base, not assumed. But it
  was not taken at the tip, and saying so is cheaper than implying it was.
- `make test` was not run in full. `make lint` was, and the tier-1 shell suite was, which is what this diff
  can affect; the remaining `make test` legs (dashboard, frontend, compose, selftest, fakes) are covered by
  CI and are untouched by a change confined to `tests/stack/` and two text files.


VijitSingh97:
## Adversarial review — one real finding, fixed; verdict MERGE

Reviewed in a fresh context against the diff rather than against this body, with the reviewer asked to
falsify rather than confirm. Posting as a comment because `develop-v2` is unprotected and a single
account cannot approve its own PR.

### The finding, and why my own earlier sweep could not have caught it

**`$VALID_PRIMARY` was an unlisted ambient dependency, on the one line the header exists to explain.**
The seed reads it (`WALLET="${WALLET:-$VALID_PRIMARY}"`), the file never assigns it, and the header's
closing sentence said "Every other name is assigned here." False as written.

Re-derived rather than taken on report: `lib.sh:122` assigns it at column 0, so it is genuinely
top-level ambient — the same category as `$VALID_TARI`, which the header already discussed two
paragraphs later. Read once in this file, at the seed line; assigned zero times here.

The interesting part is the blind spot, because it generalises. My own read-vs-assign sweep was scoped
to the **moved block** — the 219 lines whose provenance had to be proven. The seed line is *authored*
content and sits above that range, so the sweep could not have reported it: it was outside the window
by construction. **A dependency audit scoped to the moved lines cannot see a dependency the cut itself
introduces**, and a seed line is exactly that. Every remaining cut in this queue that seeds a name has
the same hole; the fix is to run the audit over the whole new file, not the moved part of it.

Fixed here: the header now names `$VALID_PRIMARY` alongside `$VALID_TARI` and says how to tell both
from `$WALLET` — where `lib.sh` assigns the name, not how it reads at the call site.

### Two number corrections in this body, both mine

- The `--generate` collateral was stated as **five** rows. It is **four**: `egress.py`,
  `.github/workflows/ci.yml`, `tests/integration/lib.sh`, `tests/integration/run.sh`. That number was
  carried forward from a previous cut's notes and not re-derived against this tree — the exact relay
  failure our own standing rule names. Corrected and the rows enumerated.
- The `ok`/`bad` control cited "`echo` 25" alongside "`assert_eq` 27". The `echo` count is
  method-dependent (it moves between roughly 20 and 28 depending on whether heredoc-embedded literal
  `echo` text is counted), so it is a poor control even though my own pattern reproduces 25. Dropped;
  `assert_eq` 27 is unambiguous and does the same job.

### What the reviewer reproduced independently, with working controls

- Single hunk `@@ -2166,219 +2166,2 @@`; the moved text rebuilt from the base at **git's own reported
  range** diffs empty against the new file's body, and the comparison was shown to fail on a perturbed
  line — so the empty result means it looked.
- The standalone run: rc 0, **44 passed / 0 failed**. Removing only the seed line drops the file by
  exactly one line and gives rc 1, `WALLET: unbound variable`.
- `lib.sh` assigns `WALLET` only at `:207` and `:221`, both indented inside the two sandbox builders,
  none at top level — so the coupling claim holds.
- The ordering accident that made the block work in place was traced concretely: fourteen domain files
  sourced before line 2166 call a sandbox builder.
- Stanza count 27 → 28, exactly +1. `file-budget.tsv` diff is one row. `lint-file-budget.sh:160` is the
  fail-on-stale-ceiling clause as described. `make lint` rc 0; `shellcheck` 0.11.0 and `shfmt` clean on
  the new file; `make -n lint-sh` puts it on the same invocation as `os/overlay/pithead-data-reset`.

### Stated limits

The reviewer could not reproduce the pre-rebase after-proof numbers, because that tree state no longer
exists — this body already discloses that the run predates the rebase. CI's own `Shell tests` job runs
the same suite and passed on this commit, which is the check that stands at the tip. `make test` in full
was not run, and the body scopes why.

All proofs above were re-run after the header fix: byte-identity still empty with a firing control,
standalone still 44 / 0, removal control still fires (255 → 254, `WALLET: unbound variable`).

PR #1456, opened 2026-08-28
build(cli): build pithead from lib/pithead/*.sh, gated on parity (#1105 Phase 2, P0)

VijitSingh97:
Phase 2 of #1105 splits the 11,988-line root `pithead` script. This is P0: the build
mechanism and the gate that makes every later cut checkable, proven on the first slice.

**The shipped artifact does not change.** `pithead` is not in this diff — it is byte-identical
to `origin/develop-v2`. That is the whole claim of P0.

## What lands

- **`scripts/build-pithead.sh`** — joins `lib/pithead/*.sh` in `LC_ALL=C` name order into the
  committed `pithead`. `--check` fails on any drift, `--self-test` drives its own fixtures.
- **`lib/pithead/00-prelude.sh`** (117 lines) and **`lib/pithead/99-remainder.sh`** (11,870) —
  the first cut: shebang through the traps, and the untouched tail. The remainder is the holding
  pen that P1–P13 take slices off the front of.
- **`make lint-pithead-parity`**, in the `lint` aggregate — `--self-test` then `--check`, the
  same shape as `lint-file-budget`, `lint-topology` and `lint-trivy-parity`.

Concatenation rather than a runtime `source` because a release bundle carries the one file
(`scripts/release.sh:745` copies `pithead` by name) and an operator runs `./pithead` out of a
checkout. `lib/` is not bundled and nothing about an install changes.

## What was RUN

`make lint` → **rc=0**, including the new gate. Lint lock taken and released via the holder
protocol; `~/dev/fleet/lint-holder` written while held.

**The firing proof, on the real tree, both directions.** Each mutant asserts it APPLIED before
the gate is read — a mutant that failed to write reads exactly like a gate that held:

| Case | Mutation applied? | `make lint-pithead-parity` |
|---|---|---|
| baseline, unmutated | — | rc=0 |
| a slice edited without rebuilding | yes, 11,870 → 11,871 lines | **rc=1**, `build-pithead: FAIL — … is not what lib/pithead/*.sh builds` |
| that mutant reverted | — | rc=0 |
| the artifact hand-edited (the pre-Phase-2 habit) | yes, per `git diff --stat`: 1 insertion | **rc=1**, same message |

Re-derived by a second route rather than re-read: rebuilding in place leaves
`git diff -- pithead` **empty**, so the concatenation reproduces the committed artifact as
judged by git, not by the build's own `cmp`. The source-side mutant was checked for
runnability (`bash -n`) — a syntax-error control reds the gate for the wrong reason and reads
exactly like the real thing.

`--self-test` covers **19 assertions** including both firing controls, an empty `lib/pithead/`
(no vacuous pass), a first slice missing the shebang, the blank-edge refusals below, and the
four refusals added by review.

**Tier-1 suite, head vs baseline, both in separate worktrees at the same tip.** Read from the
**rc files**, never a log grep:

| | rc | passed / failed | log lines |
|---|---|---|---|
| head | **0** | 2922 / 0 | 3233 |
| baseline | **0** | 2922 / 0 | 3233 |

Sorted-multiset diff of the two logs, with **both** worktree paths normalised: exactly one
line-pair, the known nondeterministic ed25519 keygen fingerprint (#1325). Normalising only one
side makes every line differ on its path prefix and reads as a regression — it nearly got
reported as one.

**Baseline honesty.** That pair ran at `95a007c`, and `develop-v2` moved to `e38b1df` (#1455)
while it was running; this branch is now rebased onto `e38b1df`. I did **not** re-run the pair
at the newer tip, because `git diff --name-only 95a007c..e38b1df` touches nothing under
`tests/`, `pithead`, `lib/` or the `Makefile` — it is a dashboard-Python split. So `95a007c` is
still the effective merge base *for this suite*. That is an argument, not a measurement; say so
if you want the pair re-run at `e38b1df`.

## The blank-line separator, and why it is the build's job

`make lint` caught a real defect in the first cut, and the fix generalizes to P1–P13.

`shfmt` strips blank lines from **both** ends of a file, and it reaches these slices through
`lint-sh`'s `git ls-files '*.sh'` glob. My first boundary put the separator at the end of the
prelude, so shfmt wanted it gone while byte-identity required it — two gates in direct conflict.

So the **build** supplies it: slices are joined with exactly one blank line. Every slice is then
independently shfmt-clean, and a cut at any single blank line between two top-level blocks
reproduces the artifact byte for byte. A boundary at two blank lines or none does not, and
`--check` says so on the spot. The build additionally refuses a slice with a blank first or last
line **by name**, so the next lane meets a sentence rather than an unexplained one-line diff;
that refusal has a self-test case on each edge.

## What adversarial review found, and what it changed

Two independent non-author passes ran against this branch. Both are folded in; neither is
paraphrased away. The findings that changed code:

**A slice with no trailing newline silently produced a malformed join.** `tail -n 1` returns the
last line's *content* for a file that simply stops, so neither blank-line edge check fired, and
the join ran that slice straight into the next with **no separator at all**. Once such a state
were committed, `--check` compares the build against itself and blesses it forever. Now refused
by name. This is the one I would most want a reviewer to re-check.

**Two refusals where the exit code proves nothing.** An empty slice, and a directory named
`*.sh` (the glob matches directories). Both already failed closed — but for the *wrong stated
reason*, because `head -n 1` yields nothing for either and the leading-blank-line message trips
instead. A case asserting `rc=1` here **passes with the guard deleted**, so those cases assert
the stated reason instead. Driving any of the three through `--check` would make all three
vacuous: the fixture artifact cannot equal what malformed sources build, so `--check` returns 1
on the parity comparison either way.

**The ordering claim could not fail.** `00`/`10`/`99` sort identically under lexical and numeric
ordering, so swapping `LC_ALL=C sort` for `sort -V` passed **every existing case**. A comment
claimed the fixture was chosen to discriminate exactly that; it was false. The comment is
corrected rather than restated, and a case now uses a single- beside a double-digit prefix,
which is the smallest input that tells the two orderings apart.

**`write_artifact` was not atomic.** `cat >"$ARTIFACT"` truncates at *open* time — before a byte
is written and regardless of `set -e` — so an interrupted rebuild left an empty `./pithead`.
Worse, bash reads a script lazily by offset, so rewriting in place fed a *running* `./pithead`
the tail of a different file. It now builds beside the artifact and renames over it, carrying
the mode across explicitly, which is the property the in-place write had been chosen for.

**Every new case was then shown to FIRE.** Six mutants, each asserted to have applied *and* to
still be runnable (`bash -n`) before its result was believed — a syntax-error control reds the
suite for the wrong reason and reads exactly like the defect being absent:

| Mutant | Result |
|---|---|
| `LC_ALL=C sort` → `sort -V` | red on the ordering case |
| drop the trailing-newline refusal | red — build returns **0**, silently losing the separator |
| drop the empty-slice refusal | red on the **reason** (rc stays 1 — this is the vacuity) |
| drop the non-regular-file refusal | red on the **reason** (rc stays 1) |
| build writes straight into the artifact | red on artifact-survived |
| join emits two blank lines | red on the join case |

**Named, not implied — what is still uncovered.** No case covers the atomicity the rename
actually buys: an interruption part-way through writing the real artifact needs a race to
reproduce, and none is attempted. The artifact-survived assertions catch a build that writes
*directly* into the artifact; they do **not** discriminate rename-into-place from the previous
shape, which aborted before its copy anyway. I found that overclaim in my own comment while
running the mutants, and corrected it rather than letting it stand.

## Two decisions a reviewer should be able to veto

**1. `lib/pithead/*.sh` is deliberately NOT added to `lint-sh`'s shellcheck list.** This is the
shape of a known trap — a gate line naming paths one by one beside a sibling line using a glob —
so it is stated rather than left to be noticed. shellcheck already reads every one of those lines
**in context**, because `pithead` is the first name on that line, which is strictly more than it
can learn from fragments; checking them separately would re-create the SC2034 pairing problem the
adjacent Makefile comment already documents, one slice boundary at a time; and it would add an
~12k-line second analysis root to the target that has OOM-killed sessions on this box. The parity
gate is what makes this sound: the artifact shellcheck reads is provably the slices. shfmt does
cover them, because format is a per-file property and transfers; the semantic checks do not.
Reviewed and endorsed by the controller after an independent attack on the reasoning, on the
condition it be recorded in the Makefile — it is, adjacent to the path list.

**A non-author review supplied a positive demonstration that this omission is safe rather than
merely argued.** It built a slice with `\r\n` line endings — still non-blank on both edges and
still ending in a real newline, so it clears all four of `build-pithead.sh`'s edge checks — and
confirmed it **builds and passes `--check` cleanly**, because `--check` only ever compares the
build to itself, so a malformation copied faithfully from source to artifact is invisible to it
by construction. `shellcheck --severity=warning` on the resulting `pithead` then reported
`SC1017 (error): Literal carriage return` on every affected line. That is the artifact-level
pass catching a *slice-level* defect the parity gate cannot see — the ground a fragment-level
pass would have duplicated, covered without a second analysis root. `lint-pithead-parity` in
isolation is blind to it; `make lint` as a whole is not. **Relayed, not re-derived by me:** the
fixtures and the SC1017 output are the reviewer's measurements, recorded here because they are
the evidence for this decision rather than an opinion about it.

**2. `lib/pithead/99-remainder.sh` takes a budget row of 11,870** — by far the tsv's largest. The
alternative was an `is_exempt()` entry, and the row is the better instrument: it **ratchets**, so
every later Phase-2 PR must lower it, which is a machine-checked statement that the split actually
progressed. An exemption proves nothing and would outlive the push. The row retires itself when
the remainder drops under the 400-line target.

## Shared files, disclosed

- **`Makefile`** — controller slot granted for exactly this hunk (the new target, its name in
  `lint` and `.PHONY`) plus the comment the grant was conditioned on. Reported for release on merge.
- **`docs/dev/file-budget.tsv`** — **two** rows added, both from `--generate`, nothing else edited:
  `lib/pithead/99-remainder.sh` at 11,870, and `scripts/build-pithead.sh` at 437, which crossed the
  400-line target as a result of the review fixes above. The two dashboard ceilings #1455 lowered
  came across the rebase from the base branch and are **not** touched by this PR — regenerated,
  never resolved by taking a side.
- **`docs/dev/repo-map.md`** (the CLI row and the release-bundle row) and **`CONTRIBUTING.md`**
  (the budget exemption's wording, now that the sources exist, plus the new lint target).
- **`scripts/lint-file-budget.sh`** — outside this lane, **two comment lines only, no behaviour
  change.** Its `is_exempt()` note still described Phase 2 as a future event that *would* make
  `pithead` a build product, which this very commit makes false; the same claim in this PR's
  `CONTRIBUTING.md` text now contradicted it. Found by review. The file is left exactly at its
  recorded ceiling of 407 — my first wording pushed it to 408, and ceilings only go down.

## If this PR goes stale, the gate reds — that is correct, not flaky

`lib/pithead/99-remainder.sh` is a verbatim copy of `pithead` from line 119 on. If anything lands
a change to `pithead` before this merges, the slices go stale and `lint-pithead-parity` fails
loudly. That is the gate working. The fix is to re-cut the slices from the new tip and re-run the
identity proof — **never** to hand-patch the remainder to match.

## What I did NOT do

- **The budget gate's monotonic half is CI-dead (#1452), so this PR's budget proof is LOCAL** —
  `scripts/lint-file-budget.sh` run in a full worktree where `origin/develop-v2` resolves, and it
  did resolve to `origin/develop-v2` (checked, not assumed). CI green proves the per-file half only.
- No semantic refactor: no call-graph regrouping, no hoisting of mid-file `readonly`s, no heredocs
  moved to data files, no mega-function split. Order-preserving contiguous slices only, per the map.
- No `.github/workflows/` edit — the parity check is a plain build-and-diff with no history
  dependency, so unlike the budget gate's monotonic half it runs honestly in CI as part of `make lint`.
- I did not add a `tests/stack/test-harness-tooling.sh` registration. The three closest analogues
  (`lint-file-budget`, `lint-topology`, `lint-trivy-parity`) run their `--self-test` from the
  Makefile and take no registry entry; `lint-operator-strings` does both. I followed the majority
  precedent — say so if you would rather it were registered.
- I did not test what happens if two lanes rebuild concurrently. The build now writes through a
  temp file **in the repo root** (a rename is only atomic within one filesystem, and `mktemp`
  alone lands under `/tmp`, very often another one), then renames over the artifact — but I did
  not prove that under contention.
- I did not re-run the tier-1 pair at `e38b1df`; the reasoning, and its limits, are stated above.

Refs #1105.


VijitSingh97:
Follow-up filed as #1463, covering the review's §1 attacks B and C — the two defects that build,
pass `--check`, pass shellcheck at every severity, and break only at runtime.

Both were **re-derived independently before filing** rather than relayed: build, `--check`, both
shellcheck severities, and the runtime message and exit code were reproduced for each fixture
(`FOO: readonly variable` exit 1; `on_err: command not found` exit 127). The re-derivation added
one fact the original attack did not state — the same ordering error routed through a *direct*
call instead of a trap **is** caught, as `SC2218 (error)`, rc=1 at both severities. That sharpens
the finding: the gap is specific to trap targets being opaque strings to shellcheck, not a general
blindness to ordering.

Confirmed on this PR's own cut, so nothing here is a defect in this diff: `on_err` is defined at
`lib/pithead/00-prelude.sh:38` and `trap on_err ERR` installed at `:115`, same slice, definition
first; there is one boundary today; and no line is duplicated across it.

PR #1458, opened 2026-08-28
refactor(dashboard): split the XvB/earnings/badges cluster out of views.py (#1105 Phase 3 D2, develop-v2)

VijitSingh97:
D2 of the #1105 Phase 3 dashboard cut queue: the XvB / earnings / top-bar-badges cluster moves out
of `web/views.py` into `web/xvb_views.py`, and its tests out of `tests/web/test_views.py` into
`tests/web/test_xvb_views.py`. No behaviour change is intended and none is expected.

| file | before | after |
|---|---|---|
| `mining_dashboard/web/views.py` | 1919 | **1129** |
| `mining_dashboard/web/xvb_views.py` | — | **827** (new) |
| `tests/web/test_views.py` | 3195 | **1978** |
| `tests/web/test_xvb_views.py` | — | **1342** (new) |

## The boundary is the dependency graph, not the map's line range

The cut map's D2 range is an approximation and says so. The boundary here is the forward closure of
an 11-name XvB/earnings/badges seed over `views.py`'s top-level dependency graph: **27 names in two
contiguous runs (lines 76-99 and 907-1668) with zero non-member names interleaved in either run.**

Exactly one edge leaves the closure to something other than `build_state`: `_LOW_HR_TITLE`, which
`build_hashrate` also reads. That is safe in the direction D1 established — `views.py` imports from
its submodules and never the reverse — and it is domain-correct anyway, since the string is an XvB
message. **`xvb_views.py` imports nothing from `views.py`; verified by AST, so there is no cycle.**

## What was RUN

- **Moved source is byte-identical.** Both runs (24 and 762 lines) compared line-for-line against
  `git show develop-v2:...views.py`, anchored on a line asserted unique in the destination.
  **The instrument was mutation-tested in both directions:** a constant mutated in run 2 and a word
  mutated in run 1 each made it red, each mutation was confirmed to have applied (md5 changed), and
  each left the *other* run passing — so it is a per-block check, not a file hash.
- **Retained `views.py` is byte-identical to base** on every line except two declared deltas: the 16
  import names that the move made unused (removed), and the 17-line facade import block (added).
  Whole-file diff shown to contain nothing else.
- **Test move is 1:1.** `pytest --collect-only`: 2120 IDs before, 2120 after, **identical once the
  module path is normalised**. That check carries its own positive control — the un-normalised diff
  is 220 lines, so the normalisation is not a no-op.
- **Moved test bodies are identical modulo 28 declared edits** (below), across all 11 blocks.
  Mutation-tested: a mutated fixture value makes the check exit 1; restored, it exits 0.
- Suite **2120 passed / 0 failed, rc=0** — same as the baseline taken at this tip before the cut.
- Frontend `node --test` rc=0, `# fail 0`.
- `ruff check` and `ruff format --check` clean across `dashboard/`; `make lint-docs-voice` OK.
- **File budget gate rc=0**, including the monotonic half. `xvb_views.py` (827) and
  `test_xvb_views.py` (1342) take new ceiling rows — both are over the gate's 800 hard ceiling,
  which refuses an *unbudgeted* file at that size, so the rows are required rather than optional.
  Four foreign rows (`egress.py`, `ci.yml`, `tests/integration/*.sh`) keep their existing higher
  ceilings: hand-curated per the standing rule, lower only what this diff shrank.
- Coverage: `make test-dashboard` rc=0, **TOTAL 97%** (bar 80).
- **Patch coverage: 100%, 206 lines, 0 missing** — measured with `--compare-branch=develop-v2`.

  **Read the gate's own number with care.** `scripts/patch-coverage.sh` hardcodes
  `COMPARE=origin/develop`. This PR targets `develop-v2`, so the gate's run reports 97% over **81
  changed files** — the whole develop→develop-v2 divergence, not this patch. It passes (rc=0) and
  its anti-vacuous check reports positively ("all 81 changed file(s) under the measured tree are
  present in coverage.xml"), so the gate is not lying about anything it claims; but its percentage
  is not a statement about this diff. `scripts/lint-file-budget.sh` already solved exactly this with
  a lane-aware base ref. Filed separately; not fixed here, since `scripts/` is not this lane's.

## The 28 declared test edits, and the two that deliberately did NOT change

Every edit is a module alias following its target from `views` to `xvb_views`. **But four `views.`
references inside the moved tests had to STAY `views.`**, and they sit within four lines of ones
that had to change:

```
monkeypatch.setattr(views, "xvb_current_tier_reward_day", ...)   # STAYS — steers build_state,
monkeypatch.setattr(views, "xvb_realization", ...)               #   which still lives in views.py
lo, hi = views.XVB_REALIZATION_PRIOR                             # CHANGES — moved constant
```

A blanket `views.` -> `xvb_views.` rewrite would have broken the first two silently. Each site was
classified individually against what actually reads the name.

## The finding this cut turned up: a monkeypatch coupling byte-identity cannot see

`tests/service/test_telegram_commands.py` — a file outside this move — did
`monkeypatch.setattr(views_mod, "monero_is_local", ...)` to steer `build_badges`. `build_badges` now
lives in `xvb_views.py` and reads that name from its own module globals, so patching `views` no
longer reaches it. Here it failed loudly with `AttributeError` because `views.py` no longer imports
the name at all — but had `views.py` still needed it for its own use, **the patch would have applied
to the wrong module and the test would have passed while testing nothing.** Fixed by rebinding the
alias to the module that now owns the name (and renaming it from `views_mod` to `xvb_mod` so it does
not lie). All 32 sites reaching a moved-or-pruned name through the `views` module object were
enumerated mechanically, not just the ones that happened to fail.

## Duplicated test fixtures — declared, not an oversight

`_SYNC_DONE`, `_BASE`, `_metrics`, `_sync`, `_hashrate`, `_state_mgr` and `_data` are **copies**.
The master stays in `test_views.py`, which still needs them for the retained tests. This is ~93
duplicated lines, and the new module's docstring says so where a reader meets the code.

There is no import route between two test modules in this suite, and this was **probed, not
assumed**: `pyproject.toml` runs pytest with `--import-mode=importlib`, the test tree has no
`__init__.py` and no `sys.path` manipulation, and both `from _sibling import X` and
`from conftest import X` raise `ModuleNotFoundError` when actually run. Converting to real pytest
fixtures would rewrite ~134 call sites and cost the verbatim-move proof this whole split rests on.
The same trade is already recorded in this very file at `test_views.py`'s `_state_mgr`, for the same
pair of helpers, against `test_metrics.py`.

Consolidating these into real fixtures once #1105 stops requiring byte-identity is filed separately
and referenced from the new module's docstring.

## What was NOT run — do not round these up

- **`make lint-sh` was not run.** The diff contains zero shell files, and it is the serialized OOM
  path. Named rather than implied.
- **`make test` in full was not run** — the stack/compose/integration tiers were not exercised. The
  dashboard tiers were, which is what this diff can affect.
- **The budget proof's monotonic half is local-only (#1452)**; CI green proves the other half.
- **Nothing was seen on screen.** This is a move of Python that emits tokens, and no rendering
  changed, but this lane still has no browser.


VijitSingh97:
## Adversarial non-author pass — verdict MERGE

Ephemeral reviewer (sonnet), fresh context, read-only, told to attack rather than confirm. It ran its
own commands rather than trusting this body. What it re-derived, and the one claim I re-derived myself
on top of it:

**1. The dangerous class — module-attribute patches broken by the move.** A moved function reads its
globals from its NEW module, so a `monkeypatch.setattr(views, ...)` elsewhere in the tree can start
patching the wrong module and PASS while asserting nothing. Byte-identity, collect-only IDs and pass
totals are all blind to it. The reviewer enumerated every such site across the whole test tree and
classified each by which module the code under test actually reads the name from.

I re-derived this one independently, because it is the claim the merge turns on:

| site | patches | reads it from | verdict |
|---|---|---|---|
| `tests/service/test_telegram_commands.py:879-880` | `xvb_mod.monero_is_local` / `tari_is_local` | `build_badges`, which MOVED | correctly repointed |
| `tests/web/test_xvb_views.py:303-320` | `xvb_mod.monero_is_local` / `tari_is_local` | same | correct |
| `tests/web/test_xvb_views.py:789-803` | `views.xvb_current_tier_reward_day` / `xvb_realization` | `build_state`, which STAYED and calls both by bare name (confirmed at `views.py` build_state body) | correctly LEFT pointing at `views` |
| `tests/web/test_views.py:1499` | `mining_dashboard.web.views.time.time` | retained code — `time` still imported and used at five retained sites | unaffected |

Thirteen sites total, all classified. I also checked the residual hole the enumeration could not have
seen otherwise: `raising=False` appears three times in the tree and none of them is on these modules,
so there is no site that could absorb a wrong-module patch silently.

**2-6, re-derived by the reviewer with the command shown each time:** facade completeness (all 11
re-exports actually used; `telegram_commands.py:41` still resolves `build_badges` through it); no
back-edge — `xvb_views.py` imports nothing from `views.py`, and `views.py`'s whole-file diff against
the base is exactly three hunks (import prune, 17-line facade, one contiguous 764-line removal) with
nothing scattered; the seven duplicated fixtures diffed byte-identical to their masters in
`test_views.py`, with no shared-state mutation or `is`-identity comparison against them; the moved
test bodies differ from their source only by the 17 module-alias renames, 1:1 in both directions, with
no assertion text or value changed; `wc -l` on all four files matches the budget rows; the cited #1459
was confirmed to be a real open issue with matching content rather than a fabricated reference.

**Gates:** all 18 required checks green on this head, including `Shell tests` which was the sole
reason `mergeStateStatus` read UNSTABLE earlier — it has since completed successfully.
`mergeable/mergeStateStatus` now read `MERGEABLE/CLEAN`.

Merging on that pass.

Issue #522, opened 2026-07-14
Config commit round-trip is only tested for p2pool.pool — cover the whole editable allowlist


VijitSingh97:
## Gap

The control black-box test drives a full **preview → commit** round-trip through `control_approval_gate` for exactly **one** key: `P2POOL_FLAGS` / `p2pool.pool` (`tests/stack/run.sh:4182`). The other ~30 keys in the dashboard-editable allowlist (`CONTROL_DASHBOARD_EDITABLE_KEYS`, `pithead:4612-4627`) — `MONERO_MEM_LIMIT`, `TARI_MEM_LIMIT`, `XVB_ENABLED`, `XVB_DONATION_LEVEL`, `DASHBOARD_TZ`, `TARI_REQUIRED`, `DASHBOARD_CHECK_UPDATES`, `HASHRATE_DROP_THRESHOLD_PCT`/`_MINUTES`, `TELEGRAM_DAILY_SUMMARY_TIME`, and the 24 `TELEGRAM_EVENT_*` toggles — are covered only by `describe_change` unit assertions. Nothing proves they actually pass the approval gate and land in `config.json` via a commit.

The gate logic itself is well-tested, so severity is moderate — but there's no per-key proof each allowlisted key round-trips (an accidental allowlist/render mismatch, like the energy/`.env` gap in #519, would slip through).

## Proposal

A table-driven test over `CONTROL_DASHBOARD_EDITABLE_KEYS`: for each key, set a valid new value, run preview → assert it's in the diff → commit → assert it landed in `config.json` (or the rendered `.env`). Refs #33, #349, #519.

VijitSingh97:
Done — merged to develop (energy commit + config test backfill landed in #536/#537; the #519 bug shipped fixed in v1.5.3 and the editor commit path in #537). Will ride to main with the v1.6 release.

Issue #506, opened 2026-07-14
Separate worker config from dashboard config (unify dashboard.workers[] into the top-level workers section)


VijitSingh97:
## Problem

Worker configuration is split across two disconnected places in `config.json`:

- **`workers.*`** (top-level) — fleet-wide worker API settings: `api_port`, `api_auth`, `api_token`. Rendered to `XMRIG_API_PORT`/`WORKER_API_*` (`pithead` ~line 3281).
- **`dashboard.workers[]`** (nested under `dashboard`) — the per-worker endpoint descriptors `{name, host, port, control_port, token}` (#172), used by the dashboard's per-worker probe and the #185/#33 control channel to resolve a rig's address + bearer (`pithead` ~line 2196; `build/dashboard/mining_dashboard/config/config.py:192`).

A top-level `workers` section already exists, yet the per-worker descriptors live under `dashboard.workers[]`. That's confusing on two counts:

1. The descriptors are *worker* topology, not dashboard/UI config — they don't belong under `dashboard.*` (which is otherwise UI/auth/onion settings).
2. An operator configuring their fleet has to edit two unrelated parts of the file (one top-level, one nested), and it's easy to set `workers.api_port` but forget the matching `dashboard.workers[]` entry (this bit us setting up the enriched-feed probe during the v1.5.0 bench work).

## Proposal

Consolidate all worker configuration under the top-level `workers` section; keep `dashboard.*` for dashboard-only settings:

```json
{
  "workers": {
    "api_port": 8080,
    "api_auth": "none",
    "api_token": "",
    "list": [
      { "name": "rig1", "host": "10.0.0.9", "control_port": 8082, "token": "..." }
    ]
  },
  "dashboard": { "...no workers[]..." }
}
```

(sub-key name — `list` / `endpoints` / `descriptors` — TBD.)

## Affected code / docs
- `config.reference.json`, `config.minimal.json` — relocate the key.
- `pithead` — the `dashboard.workers[]` validation (~2196) and the `workers.api_*` render (~3281).
- `build/dashboard/mining_dashboard/config/config.py` (`:192`, `:220`) — the descriptor loader.
- `docs/configuration.md`, `docs/workers.md`.

## Backward compatibility
`dashboard.workers[]` ships in released configs (#172, and #185 Worker Inspect leans on it), so this can't be a silent rename. Options:
- read `workers.list[]`, fall back to `dashboard.workers[]` for a deprecation window with a one-time warning; or
[...]

PR #1461, opened 2026-08-28
test(stack): split the worker-config domain into its own file (#1105 Phase 1, develop-v2)

VijitSingh97:
Moves the two per-worker token-mask sections — the legacy `dashboard.workers[]` shape (#172/#679) and the `workers.list[]` shape (#506) — out of `tests/stack/run.sh` into `tests/stack/test-worker-config.sh`. Phase 1 of #1105, develop-v2 lane, row R2 of the cut map. `run.sh` 2602 -> 2509; the new file is 127 lines.

## Pure move

The two sections are contiguous and the `source` stanza sits at the position they occupied, so no section in the suite changes execution order.

Byte-identity was re-derived **from git at the range git itself reports**, not at the ranges fed to the extractor — `git diff -U0` gives a single hunk, `@@ -1365,95 +1365,2 @@`. The moved text rebuilt from the base diffs empty against the new file below its header, and the expected `run.sh` rebuilt the same way diffs empty against the working tree. Each direction carries a positive control, and **each control was confirmed to have actually applied before its firing was trusted** — a mutant that does not apply reads exactly like a guard that held.

## Three corrections to this row of the cut map

All three are settled and already in the map; recording them here because the row as originally written would have produced a different and wrong cut.

1. **A range is not a domain.** The mapped ~395-line range bundled two foreign domains — the approval gate (#33) and the editable allowlist (#522) — which stay in `run.sh`. The true domain is 96 lines.
2. **The `$WALLET` self-arm flag was false.** This block reads `$WALLET` zero times. Control: 39 whole-file reads, 2 of them inside the approval-gate section that stays. No seed was needed and none was added.
3. **Self-arming would be the defect here, not the fix** — see below.

## The file is position-locked, deliberately, and that is stated in its header

The sections commit through the real control gate, so they write the shared request spool and append the shared audit log. Two sections that stay in `run.sh` **after** this one count exactly that: `audit log growth is bounded (#349)` measures the log's length, and the #33 spool-hardening section asserts the spool's contents. A sandbox built here, or this file sourced anywhere else in `run.sh`'s order, moves those counts and reds two domains that never changed.

So the file inherits ambient state instead of self-arming, following the shipped `test-control-add-only-ssrf.sh` precedent, and its header states the contract — because **a survived accident and an honoured contract look identical from a green suite**, and the only thing that distinguishes them is whether someone wrote down which one it is.

The lock is proven rather than asserted: sourcing the file standalone under `set -u` fails on `$C`, and supplying `$C` makes it fail further on `$MASKED`. That is a discriminating pair, not a single assertion that any missing name would satisfy.

### The firing evidence for not self-arming, and what its provenance actually is

The argument above is mechanical. There is also **empirical** evidence, and it is about this exact block rather than the pattern in general — which is why it belongs here rather than in a general note.

This lane's coupling rule A records a real red suite from moving **this same rig-worker token-mask cluster with a freshly re-derived `$C`**: the block's `apply -y` calls wrote 3 extra result files into the shared results directory, and `control-run-pending`'s `ls "$RESULTS" | wc -l` — which expects 1 — saw 4. The suite went **RED at 2901 passed / 1 failed**. The mechanism is that `build_control_sandbox` only does `mkdir -p`; it never clears prior results, so a fresh sandbox does not isolate the block, it just desynchronises the counters that other sections read.

That is the firing evidence that self-arming here is the defect and not the fix, and it is the reason this file inherits ambient state instead.

**Provenance, stated precisely because a citation without one is worth little.** This is a *recorded* result carried in the lane's coupling rules, not something re-run for this PR. It predates the fleet repo's own git history — it is already present in that repo's first snapshot commit (2026-08-23) — so the run itself is earlier and I cannot date it more tightly than that. I did not reproduce it, and this PR does not claim to have. What this PR does show independently is the mechanism the red exercised: the three assertions that count the shared spool and audit log all sit in sections that remain in `run.sh` **after** this file's source point (`run.sh:1365-66`, versus the `#349` audit-log section at `:1389` and the `#33` spool-hardening section at `:1416`).

## The whole-file name audit

Run over the **whole file, header included** — a sweep scoped to the moved lines cannot see a dependency the cut itself introduces, which is how a false "every other name is assigned here" shipped into review on the previous row.

11 distinct names read, 4 assigned. `$C` and `$CTRL_LOG` resolve to `lib.sh`'s `build_control_sandbox`; `$AUDIT`, `$MASKED`, `$REQS`, `$RESULTS` and `$STAGED` resolve nowhere, which is exactly the inheritance the header discloses. From `lib.sh` the block calls `assert_eq`, `run_pending`, `run_sourced`, `ok` and `bad` — the header's list, mechanically confirmed rather than eyeballed.

## Full-suite after-proof

rc 0, **2922 passed / 0 failed, 272 section headers, 3233 log lines** — equal to the baseline on every one.

The ordered comparison (#1398), the only check in the recipe that can see a permutation, reports the same header set, the same header order, and **0 of 272 positions differing**. The whole-output sorted-multiset diff is two lines, both halves of one ed25519 keygen fingerprint (#1325). Both moved sections were verified to have run in the after log.

The `CONFIG_FILE` line shift was predicted zero **before** the run and was zero. That prediction is structural, not arithmetic: those warnings are emitted by `tests/stack/test-appliance-identity.sh` at its own lines, not by `run.sh`, so a cut that does not touch that file cannot move them.

## Coupling rules, discharged by disclosure rather than by absence

This block **does** touch the shared `$C`/`$RESULTS`/`$MASKED`/`$REQS`/`$STAGED`/`$AUDIT` state. That is precisely why it is sourced in place rather than made standalone, and why rule A is satisfied by keeping the ordering rather than by the block being clean of those names.

No `Makefile` change is needed, and the `shellcheck` pairing is **derived rather than assumed** — the pairing is load-bearing on this repo (a global reads as SC2034 unless the file that reads it is named on the same command line as the file that sets it), so it is worth getting exactly right.

`make -n lint-sh` has two invocations: the first carries `tests/stack/lib.sh` and the `tests/stack/test-*.sh` glob that picks the new file up, and the second lints **`tests/stack/run.sh` alone**. So the new file is paired with `lib.sh` — which is what supplies `$C` and `$CTRL_LOG` — and is *not* paired with `run.sh`.

That costs nothing here, and the second invocation turns out to buy something. The new file is `shellcheck`-clean alone (rc 0) and clean paired with `lib.sh` (rc 0), so no pairing is load-bearing for it. And because `run.sh` is linted **alone**, any global that `run.sh` sets and only the moved block read would now surface as SC2034-unused — `run.sh` alone is rc 0, so **the cut orphaned no global**. That is a check the move proof itself cannot make. `shfmt` clean; `make lint` rc 0.

### One header defect, found at the over-engineering pass and fixed

On a move PR the pass has exactly one legitimate target — the authored header, the only content in the diff that is not moved code. It found one: the header cited `test-control-deploy.sh` alongside `test-control-add-only-ssrf.sh` as shipping under the same ambient-fixture contract. **It does not.** `test-control-deploy.sh` builds its own throwaway sandboxes under `$SANDBOX` and reads none of `$C`/`$MASKED`/`$REQS`/`$RESULTS`/`$STAGED`/`$CTRL_LOG` — its own header says so in as many words. Only the add-only-ssrf precedent is real, and only it is cited now.

The false pairing was **inherited**: `test-control-add-only-ssrf.sh`'s own header carries the same wrong attribution today. That is pre-existing and outside this diff, so it is filed as #1462 rather than widened into a move PR.

## File budget

The new file correctly has **no** tsv row: the generator drops any file at or below `TARGET_LINES=400` and this one is 127 lines. The only tsv change is `run.sh`'s own ceiling, 2602 -> 2509.

`--generate` wanted to lower four rows on files this cut never touched; those were **hand-restored**. Lowering a ceiling another lane owns reds their un-rebased branch, naming a file their diff never touched — this branch was on the receiving end of exactly that this session, when #1455 landed two lower dashboard ceilings and my un-rebased tsv reddened against them. That is the ratchet working as designed, and the fix is the rebase, which this branch has taken.

Both halves of the gate were controlled, and each mutant was confirmed to have actually applied before its firing was trusted:

- raising a foreign row (`tests/os/run.sh` 3437 -> 9999) fails with `raises tests/os/run.sh's ceiling from 3437 to 9999. Ceilings only go down.` — which is also what proves the monotonic half **ran** rather than skipping;
- lowering this cut's own row below the file's real size fails by name with `tests/stack/run.sh is 2509 lines, over its recorded ceiling of 100`.

**The monotonic proof is LOCAL-ONLY.** `check_monotonic` never runs on CI (#1452): `lint-surfaces` is the only job running `lint-file-budget` and it checks out shallow, so no base ref resolves and the gate prints its OK over a check it did not run. Recorded per the recipe: proven locally in a full clone against base ``e38b1df` (develop-v2)`, at `2026-08-28T22:20:19Z`.

## Not done — stated rather than implied

- `make test` was not run in full. `make lint` was, and the tier-1 shell suite was.
- **The after-proof was measured before the rebase onto develop-v2.** It stands: `git diff --stat` between the proof base and the new tip is empty under `tests/stack/`, with a positive control showing the whole-tree diff is six files. But it was not taken at the tip, and that is worth knowing rather than smoothing over.


VijitSingh97:
## Adversarial review — two independent non-author passes, reconciled

Posted as a comment rather than an approval: one account cannot approve its own PR, so this is the record, not a rubber stamp.

**Verdict: merge.** Both passes are accounted for below, including where they disagreed and why.

### Pass 1 — attacking the PR's own claims

Re-derived rather than re-read, each with a positive control confirmed to fire:

- **Byte-identity, both directions, from git.** The moved text rebuilt from the base at the range git itself reports (`@@ -1365,95 +1365,2 @@`) diffs empty against the new file below its header; the expected `run.sh` rebuilt the same way diffs empty against the working tree. Mutating one line in each direction fired both controls, so the two empty results are evidence rather than a broken instrument.
- **The position-lock claim, live.** `set -u` standalone source fails on `$C`; supplying `$C` fails further on `$MASKED` — a discriminating pair.
- **Provenance of the inherited names, traced independently:** `$C`/`$CTRL_LOG` from `lib.sh`'s `build_control_sandbox` (called at `run.sh:556`), and `$AUDIT`/`$MASKED`/`$REQS`/`$RESULTS`/`$STAGED` assigned at `run.sh:672-675` and `:793` — all upstream of the source point at `:1365`.
- **Ordering:** the single-hunk diff shape is itself the proof; the replacement lands at the exact range the removed text occupied.
- **Every number re-derived:** 2602 → 2509, 127 lines, one tsv row, `TARGET_LINES=400` so the new file correctly carries no row.
- **The pre-rebase gap, checked rather than accepted:** the six files develop-v2 moved in between are confined to `dashboard/mining_dashboard/web/*.py` and the tsv; `run.sh` references only `dashboard/mining_dashboard/config/config.py`, never the `web/` module that changed. The argument is sound, not merely plausible.

### Pass 2 — checking against the cut map's binding conditions

Returned **one finding**, and it was correct: the body argued the position lock from mechanism alone and never cited the recorded RED that the map's R2 row requires as firing evidence. **Fixed** — the body now cites it, with its provenance stated honestly: a *recorded* result carried in the lane's coupling rules, predating the fleet repo's own history, **not re-run for this PR and not claimed to be**.

That RED is specifically about this block: moved with a freshly re-derived `$C`, its `apply -y` calls put 3 extra result files in the shared results directory, `control-run-pending`'s `ls "$RESULTS" | wc -l` saw 4 against an expected 1, and the suite went red at 2901/1 — because `build_control_sandbox` only does `mkdir -p` and never clears prior results. It is the empirical reason self-arming here is the defect rather than the fix.

### Why the two passes differed — scope, not fact

They do not contradict each other. Pass 1 was briefed on the PR's own claims and never given the cut map, so the map's binding conditions were structurally outside its reach. That is a defect in how the first review was scoped, not a disagreement about anything in the tree; subsequent rows in this series will carry the map row's conditions as an explicit review axis.

### One attack that failed, recorded because it is the right shape of attack

The source stanza puts `source` inside an `&&` list, which suppresses `set -e` for the sourced file — and `domain_ran` (`lib.sh:55-62`) only fires its `#1400` guard when a domain contributes **zero** assertions. So a domain aborting *partway* would be silently truncated while still reading green, across every shipped domain file.

**It is false.** `tests/stack/run.sh:7` is `set -uo pipefail` — there is no `-e`. The suite has never had errexit, so there was no abort to lose; failure detection has always been the explicit `assert_eq`/`bad` path, which sourcing does not touch. Recorded here because the theory is plausible, neither pass raised it, and the next reader should spend one grep on it rather than an hour.

### Gates

`Shell tests` passed at `head_sha=039b2c2`, verified against the check run's own `head_sha` rather than the PR's summary line. Budget monotonicity is **local-only** (#1452) and both gate halves were controlled with each mutant confirmed to have applied first.

Issue #1460, opened 2026-08-28
patch-coverage gate compares against origin/develop, so a develop-v2 PR is measured against the whole twin divergence


VijitSingh97:
## What

`scripts/patch-coverage.sh` hardcodes its comparison branch:

```
COMPARE=origin/develop
```

`develop` is frozen and all work lands on `develop-v2`, so on a develop-v2 PR the gate does not
measure the PR's patch. It measures the entire develop -> develop-v2 divergence.

## Measured, on the #1105 Phase 3 D2 branch

That branch changes 6 files, 2 of them under the measured tree. Run as the gate runs it:

- `bash scripts/patch-coverage.sh` -> `Total: 1082 lines, Missing: 27, Coverage: 97%`, and its
  own overlap check reports `all 81 changed file(s) under the measured tree are present in
  coverage.xml`.
- The same `diff-cover` with `--compare-branch=develop-v2` -> `Total: 206 lines, Missing: 0,
  Coverage: 100%`, naming exactly the two files the branch actually touches.

Same coverage.xml, same commit; 81 files versus 2.

## Why this is worth fixing rather than living with

The gate is not lying about what it claims — it passes honestly and its anti-vacuous check (#1000)
reports positively — but the number it prints is not a statement about the change under review, and
it reads exactly like one. Two consequences:

- **A number that means something else is worse than no number.** A reviewer who sees "patch
  coverage 97%" on a develop-v2 PR will take it as a fact about that PR.
- **It cuts both ways.** A develop-v2 PR whose own new lines are poorly covered can still pass,
  because its handful of uncovered lines is diluted across the whole twin divergence. The threshold
  is being applied to the wrong population, so the gate is weaker than its `--fail-under=90` implies.

## The fix already exists in this repo

`scripts/lint-file-budget.sh` hit exactly this problem and solved it: its header explains that the
base is "the twin lane it will merge into, NOT a fixed branch", and that ratcheting a develop-v2 PR
against `origin/develop` would reject every legitimately larger v2 ceiling. It derives the lane from
the branch's history rather than from being level with a tip.

The same derivation should apply here. Worth reusing that logic rather than writing a second,
subtly different one — if the two gates ever disagree about which lane a branch is on, the
disagreement will surface as a confusing gate result rather than as an obvious bug.

## Not urgent, and here is the honest bound on the impact

Every develop-v2 PR merged since the twins diverged has been measured this way, so this is a
[...]

Issue #1459, opened 2026-08-28
Consolidate the duplicated dashboard test-fixture builders into real pytest fixtures after #1105


VijitSingh97:
## What

The #1105 Phase 3 dashboard cuts duplicate a small set of test-fixture builders into each new test
module instead of sharing them. This issue tracks consolidating them into real pytest fixtures once
the split is finished and byte-identity stops binding.

At D2 (`tests/web/test_xvb_views.py`) the copies are `_SYNC_DONE`, `_BASE`, `_metrics`, `_sync`,
`_hashrate`, `_state_mgr` and `_data` — about 93 lines. The master copy is in
`tests/web/test_views.py`. D3 and D4 cut from the same file and will need the same builders, so the
count grows with each cut.

## Why it was done this way, so this does not read as an accident

There is no import route between two test modules in this suite. That was probed, not assumed:

- `dashboard/pyproject.toml` sets `addopts = "-q --import-mode=importlib"`, and its own comment says
  importlib mode is there so test modules can share basenames "without `__init__.py` packages or
  sys.path games".
- There is no `__init__.py` anywhere under `dashboard/tests/`, and no `sys.path` manipulation.
- Running a throwaway probe: `from _sibling_helper import VALUE` and `from conftest import VALUE`
  both raise `ModuleNotFoundError`.

So the options at cut time were: duplicate the builders, convert ~134 call sites to pytest fixtures,
or insert a `sys.path` entry. The middle option destroys the verbatim-move proof each #1105 cut
rests on, and the third reverses a documented repo decision as a rider on a refactor PR. The
duplication was chosen deliberately and disclosed in each new module's docstring.

Two things make the duplication tolerable rather than merely tolerated. The copies mirror the
`Metrics` dataclass, so any drift breaks them loudly at construction rather than silently changing
what a test asserts. And the same trade is already recorded in this tree: `tests/web/test_views.py`
carries a note above `_state_mgr` explaining that the near-identical pair in `tests/service/`
is left duplicated on purpose, because a shared builder would need enough parameters to read worse
than the local copy.

## What to do here

Once the #1105 Phase 3 cuts are complete and no further move needs to be provable line-for-line:

1. Add `dashboard/tests/web/conftest.py` exposing the shared builders as pytest fixtures — factory
   fixtures, since `_metrics`/`_sync` are called with keyword overrides rather than used as values.
[...]

PR #1465, opened 2026-08-28
refactor(dashboard): split the host/rig/node status sections out of views.py (#1105 Phase 3 D3, develop-v2)

VijitSingh97:
## What this is

#1105 Phase 3, cut **D3** of the dashboard cluster: the host/rig/node status sections leave
`web/views.py` for a new `web/infra_views.py`, and their test classes move 1:1 into
`tests/web/test_infra_views.py`. Pure move — no behaviour change is intended and none is measured.

| file | before | after |
|---|---|---|
| `web/views.py` | 1129 | 660 |
| `web/infra_views.py` | — | 507 (new) |
| `tests/web/test_views.py` | 1978 | 1228 |
| `tests/web/test_infra_views.py` | — | 865 (new) |

## The boundary was derived from the dependency graph, not the map's line ranges

The cut map's D3 row gives line ranges at a commit four merges back. Those are stale, so I recovered
what the ranges *named* (`git show <that sha>:…/views.py`, AST top-level map) and took the **forward
dependency closure** of those names at the current tip. The closure is **18 top-level names in ONE
contiguous run, lines 398-878, with ZERO non-member names interleaved** — `_shorten`, `_usage_level`,
`_REJECT_FLAG_RATE`, `_REJECT_FLAG_MIN`, `_reject_flag`, `_num`, `_fmt_num`, `_rigforge_display`,
`build_system`, `rigforge_update_for`, `build_workers`, `_ENERGY_DISCLAIMER`, `_worker_watts_config`,
`build_energy`, `build_proxy_summary`, `_ip_to_sort_int`, `build_tari`, `build_sync`.

**One edge leaves the closure that is not `build_state`:** `_shorten` is also used by
`build_pool_network`, which stays. It moves out and `views.py` imports it back through the facade —
the same shape D2 used for `_LOW_HR_TITLE`. No cycle: `infra_views.py` imports nothing from
`views.py` (checked by AST, not by grep).

`logger` is inside the closure but is **not** moved; the new module defines its own, under the same
`"WebViews"` name on purpose, so no operator's log output changes. That is a declared delta.

## What consumers see: nothing

`views.py` stays the facade. `build_state` assembles these six sections from it, and
`web/worker_detail.py`'s existing `from mining_dashboard.web.views import _rigforge_display,
rigforge_update_for` keeps resolving through two `# noqa: F401` re-exports. No consumer module
changed. All seven view/consumer modules were imported in a fresh interpreter and all seven load.

## The proof class that byte-identity cannot see — and what it found

W3's binding proof class: a moved function reads module globals from its **new** module, so a
`monkeypatch.setattr(views, "<name>", …)` elsewhere can start patching the wrong module and **pass
while asserting nothing**. I enumerated every module-attribute patch/read of every one of the 18
moved names across the whole test tree and classified each site by what actually reads the name.

**Result: zero sites needed repointing for correctness.** The enumeration is not empty, though — it
returns 41 `views.*` sites, and the classification is what makes them safe:

| shape | count | why the move cannot break it |
|---|---|---|
| `setattr(views.config, "X", …)` | 34 | sets an attribute **on the shared `config` module object**, which `infra_views` imports as the same object. A move rebinds nothing here. |
| `patch.object(views, "detect_host_ipv4")` | 4 | `detect_host_ipv4` did not move; it steers `host_display_addr`, which stays. |
| `patch("…web.views.time.time")`, `setattr(views.os.path, …)` | 2 | `time`/`os` stay imported in `views.py` and are read by retained code. |
| `setattr(views, "xvb_current_tier_reward_day"/"xvb_realization")` | 4 (in `test_xvb_views.py`) | D2's; they steer `build_state`, which stays. Untouched. |

The distinction worth writing down, because it is the one that decides every row above:
**`setattr(views.config, …)` mutates a module both files share, while `setattr(views, "name", …)`
rebinds a name in one module's own globals.** Only the second shape is a move hazard. A blanket
rename would have been exactly as wrong as no rename.

**The three edits in `TestBuildEnergy` are therefore cosmetic, and I am not going to dress them up
as a fix.** They follow `from mining_dashboard.web import views` / `views.config` to `infra_views`.
Both spellings work identically today. I changed them because a test living in
`test_infra_views.py` should not reach its target through `views`, and because the day someone
prunes `config` from `views.py` the old spelling breaks for a reason unrelated to the test.

- **Source:** the moved run is **byte-identical** to `views.py` at the merge base (481 lines,
  md5 `6aaf75a1a394`). `views.py`'s retained half differs from base by **exactly** two pruned
  imports (`format_disk_size`, `compute_update`, both now unused) and the 15-line facade block —
  whole-file diff shown to contain nothing else.
- **Tests:** block 1 (613 lines, `# --- Sync display state mapping ---` through `TestProxySummary`)
  is byte-identical with **zero** edits; block 2 (`TestBuildEnergy`, 124 lines) is byte-identical
  modulo the **three declared alias edits**, and the checker asserts it applied exactly three.
  `test_views.py`'s retained half is a **pure deletion** apart from one rewritten import line (the
  nine moved names and `_sync_metric` pruned).
- **The six duplicated fixtures** (`_SYNC_DONE`, `_BASE`, `_metrics`, `_sync`, `_state_mgr`,
  `_data`) were diffed against their masters in `test_views.py`: byte-identical.
- **Both identity instruments were mutation-tested and FIRED**, with the mutant verified to have
  applied (md5 changed): a mutation inside block 1 left block 2 passing, and a mutation in a fixture
  copy left both blocks passing — so these are per-block checks, not file hashes.

## Gates — what was RUN

- Suite: **2120 passed / 0 failed, rc=0** at base and at HEAD.
- Collect-only identity: **2120 IDs = 2120 IDs**, normalised diff **empty**, with a **positive
  control**: the un-normalised diff is 128 lines, so the normalisation is not a no-op. Taken with
  bare `--collect-only` (`addopts` already carries `-q`; adding another prints counts, not IDs) and
  the artifact was checked to contain `::`.
- Dashboard coverage **TOTAL 97%** (bar 80); `infra_views.py` **189 statements, 100%**.
- Patch coverage **100%, 191 lines, 0 missing**, measured against `origin/develop-v2`.
  The gate as shipped passes too, but read its number carefully: it hardcodes `origin/develop`, so
  it reported **97% over 1440 lines across 82 files** — the whole twin divergence, not this patch.
  That is #1460, not a defect in this change.
- `ruff check` + `ruff format --check` clean across `dashboard/`; biome `check` clean (27 files);
  frontend `node --test` **480 pass / 0 fail**; `lint-docs-voice` OK; `lint-operator-strings` OK.
- File-budget gate **rc=0, both halves** (`origin/develop-v2` resolves in this worktree, so the
  monotonic half really ran — it does not run in CI, #1452). Base for that proof: the branch's
  merge base with `origin/develop-v2`, taken after the rebase described below.

## The baseline, and the rebase

`develop-v2` moved twice while this was in flight (#1461, then #1456). The branch is rebased onto
the current tip. The base-side suite and collect-only runs were taken at the earlier tip, and that
is still sound rather than merely convenient: `git diff --name-only <old>..<new> -- dashboard/`
returns **zero files**, so every input the dashboard suite reads is byte-identical between the two.
The budget proof was re-run after the rebase, because the tsv is exactly the file that did change.

## Duplicated test fixtures — the standing ruling, and the exit

Per the #1105 D2-D5 ruling, the six shared module-level builders are **copied** into the new test
module rather than shared. Disclosed in the new module's **docstring** as well as here, naming the
master (`tests/web/test_views.py`) and the sibling copy (`tests/web/test_xvb_views.py`). There is no
import route between two test modules in this suite — `--import-mode=importlib` with no
`__init__.py`, so neither a sibling module nor `conftest` is importable — and real fixtures would
rewrite every call site and cost the verbatim move. The copies mirror the `Metrics` dataclass, so
drift breaks them loudly. Consolidation is tracked as **#1459**.

## Also in here

Three comments named a module that no longer holds the code they point at. Two are this cut's doing
(`components.mjs` and `worker_refresh.py`, both pointing at `views.py` for the two RigForge
helpers); one is D2's leftover (`components.test.mjs`, the 999% cap, which now lives in
`xvb_views.py`) and is corrected because the same sweep found it. `worker_refresh.py` also named
`build_worker_update`, which exists nowhere in the tree — dropped rather than repointed. All three
edits are line-neutral, so no budget row moves.

## Over-engineering pass — considered and rejected

- **Repointing `worker_detail.py` at `infra_views` instead of keeping two `# noqa: F401`
  re-exports.** Two lines, and it deletes the crutch. Rejected because the W3 proof class says
  consumers keep importing from where they already do, and D1 set exactly this noqa shape for
  `server.py` — being consistent across three cuts is worth more than my local preference here.
  Easy to overrule: say so and it is a two-line follow-up.
- **`logging.getLogger(__name__)` in the new module** instead of reusing the `"WebViews"` name.
  Conventionally nicer, but it changes what every operator sees in their logs for six sections.
  A pure move should not do that.
- **Splitting `infra_views.py` further** (system / workers+energy / node). Six section builders in
  507 lines is one coherent domain and the map's D-queue is where further cuts get decided.
- **Real pytest fixtures instead of the six copies.** Rejected by the standing ruling for the whole
  D2-D5 set, with #1459 as the tracked exit — not re-litigated here.

Nothing new was abstracted, generalised, or made configurable by this cut.

## What was NOT done

- **Not reviewed yet.** Green is not reviewed; an adversarial non-author pass is next.
- **No `security-reviewer` pass.** Nothing here touches the control channel, auth, or an outbound
  fetch — but nobody has checked that claim but me.
- **`make lint-sh` not run** (serialized OOM path; zero shell files in the diff) and **`make test`
  not run in full** — the stack/compose/integration tiers are not exercised by a dashboard-only move.
- **Nothing seen on screen.** This lane has no browser; the claim that rendering is unchanged rests
  on the byte-identity of the builders and the suite, not on a look at the page.

## Correction — the enumeration was complete in RESULT but not in METHOD

A non-author review pass found this, and it is worth stating rather than quietly fixing.

The enumeration above was described as covering *every module-attribute patch/read* of the moved
names. It was grepped by the `setattr(...)` / `patch(...)` / `patch.object(...)` shapes, and those
do not see a **plain assignment** — `views.SOME_NAME = value` — which rebinds a name in `views`'
own globals just as effectively and is therefore the same hazard class.

Three such sites exist in the tree and none was in the original count:

| site | assigns | moved? |
|---|---|---|
| `tests/frontend/fixtures/_gen_state.py:27` | `views.time.time` | no — mutates the shared `time` module |
| `tests/frontend/fixtures/_gen_state.py:28` | `views.HOST_IP` | no — stays in `views.py` |
| `tests/frontend/fixtures/_gen_state.py:29` | `views.detect_host_ipv4` | no — stays in `views.py` |

A fourth, `tests/web/test_views.py:1004` (`views._SHELL_CACHE`), is the same shape and also names a
retained symbol.

**Re-derived over the union of all five shapes** — plain assignment, `setattr(views, "…")`,
`patch.object(views, "…")`, `patch("…web.views.…")`, and every bare `views.X` attribute read —
across every tracked file under `dashboard/`: **74 `views.X` references in total, and ZERO name any
of the 18 moved symbols.** So the *conclusion* (zero sites need repointing) survives the wider net;
only the claim about how exhaustively it was searched needed narrowing.

The union is what fires, not any single pattern: a positive control confirms that
`setattr(views, "_shorten", …)` is invisible to the attribute-read regex (the name is inside a
string) and that the assignment shape is invisible to the `setattr` regex. Each of the five probes
carrying a moved name was checked to be flagged by at least one pattern before the enumeration was
believed.

The general lesson, since this is the third W3 cut to depend on this proof class: **enumerating a
rebinding hazard by grepping the API that usually performs it will miss the assignment that does it
directly.** An enumeration is only as complete as the union of shapes it searched, and that union
belongs in the claim.


VijitSingh97:
Non-author adversarial pass returned MERGE. It re-derived the byte-identity claims against a detached worktree at the true merge base, re-ran the suite (2120/2120, rc=0), ruff, biome, the frontend tests (480/0), the file-budget gate and coverage, and independently enumerated the module-attribute sites rather than checking my arithmetic.

Its one finding was a completeness gap in how I worded the enumeration, not a defect: I grepped the `setattr`/`patch` shapes and a plain `views.NAME = value` assignment rebinds a global just as effectively. I re-derived that myself over the union of all five shapes — 74 `views.X` references across every tracked file under `dashboard/`, zero naming a moved symbol — and confirmed by positive control that no single pattern is a superset of the others. The conclusion holds; the claim about the search has been narrowed in the body above.

Merging.

VijitSingh97:
## Correction to the enumeration table above — the counts were wrong

A second adversarial pass finished after this merged and disputed the numbers rather than the
conclusion. It was right, and I re-derived it at the merge commit itself rather than taking its
word: across `dashboard/tests/`, `setattr(views.config, …)` occurs **30 times, not 34**, and the
table's four rows sum to **44** against a stated headline of **41** — an internal inconsistency
that was visible in the table as published and that I did not catch when I corrected the section
above.

Re-derived at the merge commit, by shape:

| shape | count |
|---|---|
| `setattr(views.config, …)` | 30 |
| `patch.object(views, "…")` | 4 |
| `setattr(views, "…", …)` | 4 |
| plain assignment `views.X = …` | 3 |
| `patch("…web.views.…")` | 1 |
| `setattr(views.<module>, …)` | 1 |
| **total** | **43** |

**The safety conclusion is unchanged and has now been derived independently three times: zero sites
name a moved symbol through `views`.** That is the claim the merge turned on, and it holds. What
failed was the arithmetic presented as its evidence.

Worth stating plainly, because it is the more useful half: a count that is wrong in a table whose
conclusion is right is not a cosmetic error. It is the exact shape a reader uses to decide whether
the audit was careful, so a wrong count spends credibility the conclusion then has to borrow back.
Two passes checked the conclusion; neither checked the sum until one added the rows up.

The comment defect the same pass found — the alias edit in `web/static/components.mjs` dropped the
article that began the following sentence, leaving "…a dumb renderer. plain-xmrig worker has no…" —
is repaired in the D4 branch, disclosed there.

Issue #1464, opened 2026-08-28
The 99-remainder.sh ratchet row makes pithead unable to grow at all (#1105 Phase 2 regression)


VijitSingh97:
## Summary

`pithead` is exempt from the file budget. Since #1105 Phase 2 P0 (#1456) its content lives in
`lib/pithead/99-remainder.sh`, which carries a **ratcheting** budget row. A ratcheting row can never
rise. The two facts together mean **`pithead` can no longer grow at all**: any PR that adds a net
line to the CLI fails the budget gate, and the only way to make it pass is forbidden by the same
gate.

This was not the intent. `is_exempt()`'s own comment says the exemption exists because "it is built
from `lib/pithead/*.sh`, and those slices carry the budget rows instead" — the rows were meant to
*relocate* the accounting, not to freeze the artifact. #1456's stated purpose for choosing a row
over an exemption was "a machine-checked statement that the split progressed", which is a claim
about the split, not a claim that the CLI may never change.

## Measured, both halves

On `fix/1342-mutation-lock` (#1391) rebased onto `develop-v2` at `39ee46a`, with
`99-remainder.sh` correctly re-cut so that `build-pithead.sh --check` reports
`pithead parity OK`:

**Leave the row at its recorded ceiling:**
```
file-budget: FAIL — lib/pithead/99-remainder.sh is 12123 lines, over its recorded
ceiling of 11870 (docs/dev/file-budget.tsv).
```

**Raise the row to match:**
```
file-budget: FAIL — docs/dev/file-budget.tsv raises lib/pithead/99-remainder.sh's
ceiling from 11870 to 12123. Ceilings only go down.
```

`run_gate` fails on `lines -gt ceiling`; `check_monotonic` fails on any raise of a row present in
the base. There is no third option — the two checks close on each other.

## Why no review caught it

Both adversarial passes on #1456 reviewed it against a tree in which nothing else was changing
`pithead`. The gate is correct in isolation and only misbehaves in the presence of a second,
ordinary PR. That is worth recording as the general shape: a ratchet introduced by one PR is not
testable from inside that PR.

## A second consequence, already hit and worth its own note

After #1456, **any change to `pithead` makes `lib/pithead/99-remainder.sh` stale**, and
`git merge-tree` merges that combination **textually clean**. On the trial merge of #1391 the merged
`pithead` carried 4 references to a new function while the merged remainder carried 0.
[...]

Issue #1258, opened 2026-08-22
file-budget ratchet vs W9 perimeter growth: make lint fails on both lane tips


VijitSingh97:
`make lint` failed on both lane tips after the 2026-08-22 wave: two PRs merged green in the same wave — the file-budget ratchet recorded `build/dashboard/tests/service/test_control_service.py` at its then-current length as a personal ceiling (522 on develop, 563 on develop-v2), and the W9 perimeter work then grew the same file (533 / 583) without the gate in its path. Each PR's checks were green in isolation; the combination fails `scripts/lint-file-budget.sh`, and with it `make lint`, for anyone at either tip.

Two distinct gaps, one incident:

1. **The gate runs only locally.** CI does not execute `lint-file-budget` yet — that wiring is one of the workflow-scoped pushes tracked in #1257 — so a merge that violates the ratchet lands green. Until the CI wiring lands, any pair of same-wave merges can reproduce this.
2. **Growth of a budgeted file has no legal path.** The ratchet's rules (ceilings only go down) are right for drift, but a deliberate, justified addition to a budgeted file — here, security-perimeter keys — can only comply by shrinking the file elsewhere or splitting it. That is by design (#1105), but worth stating in CONTRIBUTING so the next author reaches for the split instead of a raised ceiling.

Fix for the immediate breakage (PR follows, both lanes): mechanical move of the perimeter frozenset + test into `build/dashboard/tests/service/test_env_key_perimeter.py`, verbatim; the source file drops back under its ceiling and the ceiling ratchets down to the new count. Mutation re-run in the new location (allowlisting a perimeter key turns the test red).

Remaining after that PR: the CI wiring (#1257, operator token) and the CONTRIBUTING note.

VijitSingh97:
Repo-side repair complete on both lanes: the perimeter-test move (#1259 develop, #1260 develop-v2) restored make lint on both tips, and the CONTRIBUTING passage naming the legal growth path + the local-only enforcement window merged as #1263/#1264. Remaining: the CI wiring for lint-file-budget rides the workflow-scope push tracked in #1257 (operator token).

VijitSingh97:
Closing: both lanes' gates are green again and the growth-path doc is merged; the only residual (CI wiring) is #1257's existing scope.

PR #1467, opened 2026-08-28
fix(gate): let the un-split-remainder row rise, so pithead can still be fixed (#1464)

VijitSingh97:
## What this fixes

`pithead` is exempt from the file budget so that a CLI bug fix may add lines to it. Since #1105 Phase 2, its content lives in `lib/pithead/99-remainder.sh`, which carries a **ratcheting** budget row — and a ratchet can never rise. The exemption silently became a hard freeze: `pithead` could not gain a single line.

**Measured on a real branch (`fix/1342-mutation-lock`), both halves, by me — there is no third option, the two checks close on each other:**

| tsv row | what refuses |
|---|---|
| left at `11870` | `run_gate`: `99-remainder.sh is 12123 lines, over its recorded ceiling of 11870` |
| raised to `12123` | `check_monotonic`: `raises ... from 11870 to 12123. Ceilings only go down.` |

This blocks **every** lane with a line-adding `pithead` change, not just the appliance lane.

## The fix

A narrow monotonic exemption for that one row, in `check_monotonic` only.

The row measures **what of the generated artifact is not split out yet**, not the size of a file anyone writes, so it has to track the artifact up as well as down. `monotonic_exempt()` carries that reasoning in-code so the next reader cannot mistake it for a weakened ratchet, and the gate prints a `NOTE` naming the rise.

**`run_gate`'s `lines > ceiling` rule is deliberately UNTOUCHED.** That is what keeps the row an honest per-PR measurement rather than a ceiling nobody reads: a PR that grows the artifact must still record the new count, every Phase 2 cut still lowers it, and it retires itself when the remainder reaches the 400 target.

Rejected, with reasons, so they are not reopened: full exemption of the row (loses the progress signal) and cut-per-PR (makes bug fixes carry refactors). **Accepted residual, stated rather than hidden:** an un-split PR now passes the budget gate; `lint-pithead-parity` forces the books to balance and non-author review is the backstop.

## Why this PR also splits the gate's self-test

Not a drive-by refactor — the fix could not land without it, and the gate itself documents the remedy.

`scripts/lint-file-budget.sh` was **407 lines against a recorded ceiling of 407**: closed to any addition. The row cannot be raised either, and that refusal is *correct* — this one is a real source file, so exempting it would hollow out the very argument this PR makes. The gate's own header rule says a file back under the 400 target **must drop its entry**, so splitting is the remedy the gate prescribes (the same path CONTRIBUTING.md already documents, with #1258 as the worked example).

Result: gate **293** lines and its row **dropped** from the tsv; fixtures move to `scripts/lint-file-budget-selftest.sh` at **223**. Neither needs a row. The fixtures `source` the gate, so the dispatch now sits behind a `BASH_SOURCE` guard; `--self-test` still works as the entry point, so **the Makefile is unchanged and no CI wiring moves.**

## What was RUN

- `make lint-file-budget` — green through the Makefile path, post-split: self-test **22/22**, then `file budget OK`.
- `make lint` — full run, see below.
- `shellcheck --severity=warning` **and** `-S info`, plus `shfmt -i 4 -d`, on both scripts: clean at both severities.
- **Source guard, proven both ways with a firing control.** Sourced from a non-git directory: `rc=0`, **zero output**, all five gate functions defined, `self_test` correctly absent. *Executed* in that same directory: `rc=1` and loud (`FATAL: git ls-files enumerated zero tracked files`). The silence is the guard working, not a dead script.
- **Mutation proof, run pre-split and RE-RUN post-split, identical reds both times.** Two one-line mutants at the exemption branch:

| variant | rc | self-test | which case reddens |
|---|---|---|---|
| clean | 0 | 22 ok / 0 fail | — |
| `if false` (exemption never applies) | 1 | 20 / 2 | the new "may rise" case |
| `if true` (exemption applies to everything) | 1 | 18 / 4 | the new **firing control**, plus the pre-existing monotonicity case |

  Neither new case is vacuous: the first dies without the exemption, the control dies when the exemption is over-broad.
- tsv diff is **exactly one deletion** (`scripts/lint-file-budget.sh 407`); no new rows, no foreign row touched.

## Self-test cases added

1. The exempt row rising **passes**.
2. **The load-bearing one** — a NON-exempt row raised *in the same run* still **FAILS**, and is the only row named. Without it, case 1 is indistinguishable from a monotonic check that stopped checking.
3. Growing the exempt file past its recorded row still **FAILS** — proving `run_gate` really was left alone.

## Disclosure

- **Two per-item grants**, both from the controller, both expiring when this PR merges: `scripts/lint-file-budget.sh` (w10) and `scripts/lint-file-budget-selftest.sh` (w11, granted specifically because the fix could not land without the split). Gate scripts are owned by no lane.
- **`CONTRIBUTING.md` is a shared root file no lane owns**, so it rides the one-shot `.lane-override`, in its own commit so the hatch covers exactly that one file.
- **`docs/dev/file-budget.tsv`** is the per-lane regenerated shared file; the change here is one deletion.
- The docs commit also **drops a stale claim**: the paragraph said the gate "is not yet wired into CI (issue #1257)". #1257 is closed and `.github/workflows/ci.yml:512` runs `make lint-file-budget`. That was stale in the direction that matters — it told authors CI would not catch this. The same-wave-merge hazard it warned about is real and is kept.

## What I did NOT do

- The **300s/appliance-first-boot reasoning in #1391 is untouched here**; this PR is only the gate.
- I did **not** change the freeze for any other file. I did sweep for it, because `scripts/lint-file-budget.sh` was found frozen by accident while fixing this: **75 of the 79 budgeted files sit at a ceiling exactly equal to their current line count** at `develop-v2` `3d61f24`, so they are closed to any addition. That is the ratchet working as designed — the intended answer is to split or shrink — but it is worth stating plainly, because it means "add a few lines here" is not available almost anywhere in the budgeted surface, and the remedy has to be planned into the change rather than discovered at the gate. Only `lib/pithead/99-remainder.sh` gets the exemption in this PR, and only for the monotonic half.
- No bench, no hardware, no `fleet.sh` claim — this change needs none.

## Review brief

The controller asked that the non-author pass treat these as **two separate axes**:

- **(a) Exemption semantics** — is `monotonic_exempt` narrow enough, and does the mutant evidence actually support the claim that both new cases can fail?
- **(b) Split mechanics** — is the `BASH_SOURCE` guard right, does anything else source or invoke this script, and does the dropped tsv row behave (`check_monotonic` skips a path absent from the working tree, `[ -n "$new_ceiling" ]`)?

Refs #1464

## Over-engineering pass (ponytail policy, run on this diff)

- `scripts/lint-file-budget.sh:L221-223: shrink: exemption NOTE spanned four continuation lines. Three, same content.` — **applied** (`9cabefd`).
- `scripts/lint-file-budget.sh:L214-220: yagni: monotonic_exempt() is a one-arm case with a single caller. Inlining [ "$path" = lib/pithead/99-remainder.sh ] saves ~5 lines.` — **noted and kept, deliberately.** It mirrors `is_exempt()`'s shape in the same file, so it reads as the file's own idiom; and the binding condition on this fix is an in-code rationale a later reader cannot miss, which needs a named thing to hang on. Inlining would push a 14-line justification into the middle of `check_monotonic`'s loop, which is worse on both counts.
- The third self-test case was considered for cutting and kept: it is the only thing proving `run_gate` was left alone, and the policy exempts self-checks from deletion.

`net: -7 lines possible, -4 taken.`

---

## Non-author review round (both axes) — three findings, all fixed at `ddafecd`

Two independent non-author passes ran, one per axis. **Both found real gaps, and I reproduced every one before acting rather than taking the reviewer's word.**

**(a1) The exemption had no fixture proving it is NARROW — the serious one.** Widening the case arm from the literal path to `lib/pithead/*` kept **all 22 cases green**, while silently exempting every real source slice Phase 2 is adding under that directory. Confirmed by running the mutant myself. The fixture now carries a sibling slice, `lib/pithead/01-prelude.sh`, whose ceiling raise must still be refused; that mutant now reds on it. This was the answer to "construct a change that keeps everything green while breaking the intent" — and there was one.

**(a2) The exemption placed no bound on the size of a raise.** The row could be set to any number and nothing mechanical would object again until the file reached it — a ratchet in name only, and the opposite of the per-PR measurement the row is meant to be. **A rise is now allowed only to the file's real current line count**; one that reserves headroom FAILS and names the count the row should have stated. My original "stays an honest per-PR measurement" claim was aspirational as written; it is now enforced.

**(b1) `--self-test` through a symlink exited 127** instead of running, because `dirname "$0"` gives the link's directory. The `exec` now resolves the link first, as does the fixtures file's own source path. No call site symlinks this, so this closes a hole rather than fixing a live break — and there is **deliberately no fixture**, because a case running `--self-test` through a symlink would re-enter the self-test and recurse. Said plainly rather than left as an implied gap.

**Also raised and NOT treated as a code change:** the reviewer noted that `lint-pithead-parity` contributes nothing to bounding the un-split residual — it is a drift check between artifact and slices, satisfied by any edit-then-rebuild. That is correct, and confirmed by running. The real backstop for that residual is non-author review, which this PR already named in the same sentence; the wording above overstated parity's part, and this note is the correction.

**Self-test is now 26 cases. Four mutants, each red on exactly the intended case:**

| mutant | reds on |
|---|---|
| exemption widened to `lib/pithead/*` | the new narrowness control (was 22/22 GREEN before this round) |
| exemption never applies (`if false`) | the "may rise" case |
| exemption applies to everything (`if true`) | the narrowness control |
| the new headroom bound removed | the new "must match the file" case |

Gate is 310 lines, fixtures 266 — both still under the 400 target, so neither takes a budget row. `make lint-file-budget` green through the Makefile path; shellcheck clean at `--severity=warning` and `-S info`; `shfmt` clean.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,12 +58,19 @@ runs `ruff` (plus a few hygiene hooks) on your changed files. If you change depe
      deliberate, justified addition to a budgeted file therefore has exactly one legal path:
      split or shrink the file so the addition fits under a ceiling that stays put or drops —
      see issue #1258 for the worked example, where a security test that outgrew its file moved
-     into its own — and note the gate runs locally and in `make lint` but is not yet wired into
-     CI (issue #1257), so two same-wave merges can each pass alone and fail together: re-run
-     `make lint` after merging onto the tip. Generated
+     into its own — and note that two same-wave merges can each pass alone and fail together, so
+     re-run `make lint` after merging onto the tip. Generated
      code, vendored files, data/config, and prose docs are exempt by glob — see `is_exempt()` in
      the script — and so is the shipped `pithead` artifact itself: it is generated, and the gate
-     governs its `lib/pithead/*.sh` sources instead, now that Phase 2 has begun splitting it),
+     governs its `lib/pithead/*.sh` sources instead, now that Phase 2 has begun splitting it.
+     One row is exempt from the ceilings-only-move-down half, and only that half:
+     `lib/pithead/99-remainder.sh` measures how much of that generated artifact Phase 2 has not
+     split out yet, not the size of a file anyone writes, so it tracks the artifact both ways.
+     Without that, `pithead`'s own exemption became a freeze — the CLI could not gain a line,
+     because the row refused to rise and the file refused to grow past it (issue #1464). The row
+     is still checked against the file on every PR, so a change that grows the artifact has to
+     record the new count, and every Phase 2 cut lowers it; it retires when the remainder reaches
+     the 400 target. See `monotonic_exempt()` in the script),
      `lint-pithead-parity` (the shipped `pithead` must be exactly what `lib/pithead/*.sh`
      concatenate to: edit a slice, run `scripts/build-pithead.sh`, and commit both — issue #1105
      Phase 2), `lint-trivy-parity` (the CVE
--- a/docs/dev/file-budget.tsv
+++ b/docs/dev/file-budget.tsv
@@ -48,7 +48,6 @@ dashboard/tests/web/test_xvb_views.py	1342
 lib/pithead/99-remainder.sh	11870
 os/rootfs/Dockerfile	406
 scripts/build-pithead.sh	437
-scripts/lint-file-budget.sh	407
 scripts/release.sh	851
 scripts/shipped-image-sweep-report.sh	429
 scripts/trivyignore-watch.sh	599
--- a/scripts/lint-file-budget-selftest.sh
+++ b/scripts/lint-file-budget-selftest.sh
@@ -1,0 +1,223 @@
+#!/usr/bin/env bash
+# Fixtures for the file-budget ratchet gate — every failure mode it has, each against a throwaway
+# git repo built for it. Split out of scripts/lint-file-budget.sh for issue #1464: the gate was
+# 407 lines at a recorded ceiling of 407, so it was closed to any addition, and the gate's own
+# header rule says a file back under the 400 target must drop its budget entry. Splitting is
+# therefore the remedy the gate itself documents, not a way around it.
+#
+# Run it directly, or as `scripts/lint-file-budget.sh --self-test`, which execs this file.
+# The gate is SOURCED for its functions, so it carries a guard that keeps sourcing from running
+# anything; see the bottom of that script.
+set -euo pipefail
+
+HERE=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
+# shellcheck source=scripts/lint-file-budget.sh
+source "$HERE/lint-file-budget.sh"
+
+# --- self-test: every failure mode against fixtures, in an isolated throwaway git repo ----------
+self_test() {
+    local tmp out st_fail=0 rc=0
+    tmp=$(mktemp -d)
+    out=$(mktemp)
+    # Double-quoted so $tmp/$out are embedded now, as literal paths — the trap still fires after
+    # self_test's own locals have gone out of scope, so a deferred expansion would see them unset.
+    # shellcheck disable=SC2064
+    trap "rm -rf '$tmp' '$out'" EXIT
+    git -C "$tmp" init -q -b develop
+    git -C "$tmp" config user.email test@example.invalid
+    git -C "$tmp" config user.name test
+
+    expect() { # <desc> <expected-rc> <actual-rc>
+        if [ "$2" -eq "$3" ]; then
+            echo "  self-test ok: $1"
+        else
+            echo "  self-test FAIL: $1 (expected rc $2, got $3)"
+            st_fail=1
+        fi
+    }
+    seq 1 500 >"$tmp/budgeted.sh"
+    mkdir -p "$tmp/$(dirname "$BUDGET_FILE")"
+    printf '# test budget\nbudgeted.sh\t500\n' >"$tmp/$BUDGET_FILE"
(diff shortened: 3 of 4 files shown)


VijitSingh97:
**Second-pass review (delta-scoped, non-author): MERGE, at head ddafecd.**

Scope was only the fix delta (`9cabefd..ddafecd`), per the three findings recorded above. Every claim re-derived by running in a fresh worktree, with firing controls:

- **a1 (narrowness):** the widened-exemption mutant (`lib/pithead/*`) now reds on exactly the sibling-slice control — the mutant that previously passed 22/22.
- **a2 (bounded raise):** proven in a standalone repo independent of the shipped fixture — a headroom-reserving raise fails naming the file's real count; an exact-match raise passes. The bound reads the working-tree file, not the tsv.
- **b1 (symlink exec):** runs post-fix; control against the pre-fix copy reproduced the rc=127. The no-fixture reasoning was checked structurally and holds (a fixture would re-enter the self-test).
- All four mutants and the clean 26/26 pass re-run; shellcheck (warning + info), shfmt, and `bash -n` clean on both files.

One non-blocking record correction: the mutant table above undercounts the `if false` mutant — it also perturbs the headroom-bound case's message assertion (three red lines, not one case). Extra sensitivity, correct behavior; noted so the table is not read as exhaustive.

Not checked in this pass: end-to-end `make lint`/CI (driven directly instead — and the CI checks are green independently), and whether any in-flight branch depends on the pre-fix behavior.

Merging on this verdict plus the green Shell tests check, completing the review round recorded above.

Issue #196, opened 2026-06-11
[Epic] Persist valuable telemetry as time-series — stop discarding block / share-health / disk-growth / XvB / per-worker data


VijitSingh97:
> **Data-capture audit + tracking epic.** A review of the dashboard's persistence layer found we historize only a fraction of the telemetry we fetch — the rest is rendered live and thrown away. This epic tracks persisting the valuable, low-cost series we currently discard, so operators get *trends* (not just last-value) and history survives restarts/disconnects.

## What's persisted today

Only three things become history in SQLite (`build/dashboard/mining_dashboard/service/storage_service.py` → `_create_tables`):

| Table | Stores | Cadence |
|---|---|---|
| `history` | total / P2Pool / XvB **hashrate** | 1 row / 30s poll |
| `shares` | P2Pool share events (`ts`, `difficulty`) | event-driven (#129) |
| `kv_store` | a few XvB scalars **+ the entire `latest_data` dict as one `snapshot_latest_data` blob, overwritten every 30s** | per cycle (overwrite) |

Everything else the data loop fetches survives only as last-value inside that overwritten snapshot → **no trend, gaps lost**.

## Design principles (shared across all sub-items)

- **Dedicated small tables**, not extra `history` columns — independent retention, and avoids multiplying `history` rows for per-worker/per-series data.
- **Retention + pruning** mirroring `HISTORY_RETENTION_SEC` + the existing probabilistic prune (`storage_service.py:240`).
- **Cumulative counters** (shares, blocks, proxy results) → store **deltas/events**, and detect a **counter reset** (proxy/p2pool restart → counter goes backwards) as a segment break, *not* a negative rate. Reuse the `_shares_to_record` pattern (`data_service.py:176`).
- **Throttle** high-cardinality / slow-moving series (per-worker ~5 min, network hourly) rather than the 30 s loop rate.
- **Expose** new series via `/api/state`; the actual charts/badges are tracked in the consuming feature issues (below).

### Tier 1 — high value, negligible/low cost
- [ ] **P2Pool block-found events** — `blocks_found` / `last_block_found` (height) / `last_block_ts` (`collector/pools.py:74`). Detect the `totalBlocksFound` increment (mirror `_shares_to_record`) → `blocks(ts REAL PK, height INT, effort REAL)`. Blocks are the actual **payout** events; enables a found-blocks timeline + effort-per-block. ~rows per *week*. → feeds **#99**.
[...]

Issue #116, opened 2026-06-04
Store share stats as time-series for reject-rate trends/charting


VijitSingh97:
## Background

[#82](https://github.com/p2pool-starter-stack/pithead/issues/82) (PR #109) surfaced per-worker **accepted / rejected / invalid** share counts and the pool-wide xmrig-proxy `/summary` totals (accepted/rejected/invalid/expired + best difficulty) in the dashboard.

That data is **persisted only as a latest-state snapshot** — it rides along in the `latest_data` blob saved to the `kv_store` table under `snapshot_latest_data` (overwritten every cycle), so the dashboard repopulates it immediately after a restart. The only *time-series* table is `history`, which stores **hashrate only** (`v / v_p2pool / v_xvb / timestamp`).

## Problem

Because share stats aren't recorded as a per-timestamp series, we **can't show a trend** — e.g. chart reject rate over the last 24h, or compute a rejects-per-hour rate to catch a rig that's slowly degrading. Right now an operator only sees the current cumulative totals.

## Proposal

Persist share stats as time-series so the dashboard can chart/trend them.

Open questions to settle in the issue:
- **Granularity** — pool-wide totals only (cheap), or per-worker series too (more useful for isolating one bad rig, but higher row volume).
- **Schema** — add columns to `history`, or a dedicated `share_stats` table (cleaner: independent retention, and avoids multiplying `history` rows if we go per-worker).
- **Cumulative-counter handling** — the proxy reports counts **cumulative since it started**, so a meaningful per-interval rate needs deltas between consecutive samples. A proxy restart resets the counters to 0, which makes a naive delta go negative — that must be detected and treated as a reset (segment break), not a negative rate.

## Acceptance criteria

- [ ] Share stats stored as a time-series (at minimum pool-wide accepted/rejected/invalid; ideally per-worker).
- [ ] Per-interval rate derivation handles proxy counter resets (restart → reset, not a negative delta).
- [ ] Trend exposed in the dashboard (a reject-rate chart/sparkline, or at least the series available in `/api/state` for the client to render).
- [ ] Retention + pruning consistent with the existing `history` policy (`HISTORY_RETENTION_SEC`).
- [ ] Tests at the storage + view layers.

## Pointers

- `build/dashboard/mining_dashboard/service/storage_service.py` — `update_history` / `save_snapshot`; the `history` and `kv_store` tables.
[...]

PR #1469, opened 2026-08-28
refactor(dashboard): split the time-series sections out of views.py (#1105 Phase 3 D4)

VijitSingh97:
## What this is

#1105 Phase 3, cut **D4** of the dashboard cluster: the `/api/state` sections that carry a *history*
rather than a current reading leave `web/views.py` for a new `web/series_views.py`, and their test
classes move 1:1 into `tests/web/test_series_views.py`. Pure move — no behaviour change is intended
and none is measured.

| file | before | after |
|---|---|---|
| `web/views.py` | 660 | 445 |
| `web/series_views.py` | — | 250 (new) |
| `tests/web/test_views.py` | 1228 | 861 |
| `tests/web/test_series_views.py` | — | 484 (new) |

What moved: the persisted per-poll share-health deltas (#116), the Tier-1 telemetry backbone's
three formatters — blocks, disk growth, XvB history (#196) — the payouts series, the pool/mode
palette tokens (#27), and the hashrate and cadence cards. They share the chart hub's downsampling
bound and range filter, which is why they sit together rather than beside the point-in-time status
sections D3 moved.

## The boundary came from the dependency graph, not the map's line ranges

The cut map's D4 row gives line ranges from before five merges. I took the **forward dependency
closure** of what those ranges named, at the current tip. The closure is 11 names in one contiguous
run, 130-347.

**Two names sit inside that run but outside the closure**, and they are the only judgement call in
the cut: `_window_reject_pct` and `_mode_palette`. Both are used by `build_state` and by nothing
else in `views.py`. Moving them keeps the run **one contiguous byte-identical block**; leaving them
would have fragmented it into three blocks for no gain, since `build_state` reads them through the
facade either way. They move. With the 4-line palette constant block — separate because
`build_raffle_log`/`build_raffle_eligibility` sit between it and the rest, and those stay — that is
**17 top-level names, zero retained names interleaved**.

Everything leaves the closure only to `build_state`, so `views.py` stays the facade. Two
re-exports: `_gauge_series` (moved) and `_filter_events` (from `charts.py`, no longer used by
`views.py` itself) are both marked `# noqa: F401` because `web/worker_detail.py` imports them
*through* `views.py`. No cycle: `series_views.py` imports from `charts` and `xvb_views`, and
nothing under `web/` imports `views.py` back — checked by AST, not grep.

Five imports were pruned from `views.py` (`share_reject_pct`, `_MAX_CHART_POINTS`, `_RANGE_SECONDS`,
`_LOW_HR_TITLE`, and `build_metrics`'s line rewritten), and nine names pruned from the test file's
import list. Both sets were derived by asking which half still uses each name, with a positive
control — my first run of that check reported every name as still-used, because the filter that
stripped import lines only stripped the first line of each parenthesised block and the import list
matched itself. That run was discarded, not patched.

## The proof class byte-identity cannot see

A moved function reads module globals from its **new** module, so a module-attribute patch elsewhere
can start patching the wrong module and **pass while asserting nothing**. D3's enumeration of this
was searched by the `setattr`/`patch` shapes only — and a plain `views.NAME = value` assignment
rebinds a global just as effectively. Four such sites exist in this tree and D3's count missed them.

So this cut enumerates the **union of five shapes**: plain assignment, `setattr(views, "…")`,
`patch.object(views, "…")`, `patch("…web.views.…")`, and every bare `views.X` attribute read,
across every tracked file under `dashboard/`.

| shape | sites |
|---|---|
| bare `views.X` attribute read | 65 |
| `patch.object(views, "…")` | 4 |
| `setattr(views, "…", …)` | 4 |
| plain assignment `views.X = …` | 3 |
| `patch("…web.views.…")` | 1 |

**Sites naming one of the 17 moved symbols through `views`: zero.** The union is what fires, not any
single pattern — a positive control confirms `setattr(views, "_shorten", …)` is invisible to the
attribute-read regex (the name is inside a string) and that the assignment shape is invisible to the
`setattr` regex. Every one of the five probes carrying a moved name was checked to be flagged before
the enumeration was believed. `raising=False`/`create=True` swept separately: 3 hits repo-wide
(conftest DB path, two wizard env vars), none touching these modules.

## The twelve declared test edits, and why they are not cosmetic

`TestBlocksDiskGrowthXvbHistorySeries` reaches its target as `views.build_payouts` (4 sites) and
patches `views.config` (8 sites). The `views.config` eight are the **safe** shape — a shared module
object both files import. The `build_payouts` four are not cosmetic: after this cut they reach a
**moved** function through the facade, which is the exact indirection W3 exists to remove. All
twelve are re-pointed to `series_views`, and the checker asserts it applied exactly twelve.

- **Source:** both moved blocks byte-identical to `views.py` at the merge base — palette block 6
  lines (md5 `aeda6a49de88`), series run 218 lines (md5 `45c97dd2db55`). `views.py`'s retained half
  differs by exactly the 18 declared added and 5 declared removed lines, **zero undeclared**.
- **Tests:** `_hashrate` (moved, not copied — no retained test uses it) and both class blocks
  byte-identical, with the 12 declared edits applied to the *base* side so the comparison is
  identity rather than "close enough". `test_views.py`'s retained half is a **pure deletion**:
  +0 lines added, 322 removed, of which 9 are the pruned imports.
- **The five duplicated fixtures** (`_SYNC_DONE`, `_BASE`, `_metrics`, `_state_mgr`, `_data`) diffed
  against their masters: byte-identical.
- **Both instruments were mutation-tested and FIRED**, and the mutation in one block left the other
  blocks passing — so these are per-block checks, not file hashes.

## Gates — what was RUN

- Suite **2120 passed / 0 failed, rc=0** at HEAD; base collect at `origin/develop-v2`.
- Collect-only identity **2120 IDs = 2120 IDs**, normalised diff **empty**, with a **positive
  control** (un-normalised diff = 62 lines). Bare `--collect-only`; artifact checked to contain `::`.
- Coverage **TOTAL 97%** (bar 80). **The two uncovered lines in `series_views.py` are pre-existing
  and I measured that rather than asserting it**: base-side `views.py` was 171 statements with
  exactly 2 missing, at lines 161 and 212 — the same two `continue` guards inside the two
  downsamplers, now at 64 and 115. `views.py` itself goes 99% -> 100%. The cut adds zero newly
  uncovered lines.
- Patch coverage **97.4%, 80 lines, 2 missing** against `origin/develop-v2`. The shipped gate reports
  97% over 1447 lines across 83 files instead, because it hardcodes `origin/develop` — that is
  #1460, not a defect in this change.
- `ruff check` + `ruff format --check` clean; frontend `node --test` **480 pass / 0 fail**;
  `lint-docs-voice` OK; `lint-operator-strings` OK.
- File-budget gate **rc=0, both halves** (the monotonic half really ran — it does not run in CI,
  #1452). `--generate` would have **lowered four rows belonging to other lanes** (`egress.py`,
  `ci.yml`, `tests/integration/lib.sh`, `tests/integration/run.sh`); all four were restored to their
  committed values and the diff is exactly my three rows. `series_views.py` at 250 needs no row.

## Docs, and one repair carried from D3

`dashboard/README.md`'s module map gains `series_views.py`. `dashboard.css:1058` pointed readers at
`views._mode_palette`, which this move makes false — repointed.

Separately: D3's alias edit in `web/static/components.mjs` dropped the article that began the
following sentence, leaving "…a dumb renderer. plain-xmrig worker has no…". Found by the #1465
non-author pass after that PR merged; repaired here as its own commit. #1465 carries a comment
correcting its enumeration counts, which the same pass showed were wrong (30 `views.config` sites,
not 34) even though its conclusion was right.

## The over-engineering pass — what I rejected

- **Real pytest fixtures instead of the five duplicated builders.** Rejected: standing #1105 ruling,
  tracked exit #1459. Would rewrite every call site and destroy the byte-identity this cut rests on.
- **Leaving `views.build_payouts` untouched** for a perfect zero-edit move. Rejected above — it
  would leave four tests reaching a moved function through the facade.
- **Keeping `_window_reject_pct`/`_mode_palette` in `views.py`** so the closure is "pure". Rejected:
  fragments one contiguous block into three and buys nothing.
- **A separate module for the four palette constants.** Rejected: four constants read by one
  function that moves.
- **Folding `series_views.py` into `charts.py`.** Rejected: `charts.py` is the window/range hub — a
  different axis — and merging would rebuild the oversized module D1 split.
- **Adding tests for the moved code.** Rejected: a pure move; new tests would make the identity
  claim unverifiable and belong with whatever behaviour change motivates them.

## What I did NOT do

- **Not reviewed.** Green is not reviewed. No `security-reviewer` pass either — nothing here touches
  the control or credential surface, but nobody has checked that claim but me.
- **`make lint-sh` NOT run** (serialized OOM path; zero shell files in the diff). **`make test` in
  full NOT run** — stack, compose and integration tiers not exercised.
- **Nothing seen on screen.** This lane still has no browser; `dashboard.css` and `components.mjs`
  changes here are comments only, but the palette this module emits is not visually verified.
- D5-D7 boundaries not derived.


VijitSingh97:
## Two non-author passes, plus my own re-derivation: the move is clean, three claims in this body are not

Both passes worked in detached worktrees pinned to base `3d61f24` / head `db2542f` and re-derived
rather than relayed. One returned MERGE; the other returned FINDINGS, and every one of its findings
is about **this body's claims**, not the code — its own words: *"the code move itself is clean and
safe."* I agree with that split, and I am correcting the body rather than the diff.

**What held, under independent re-derivation:** byte-identity of both moved blocks (empty `diff`, both
directions — the load-bearing proof, not the quoted md5); `views.py`'s retained half at exactly 18
added / 5 removed, reconstructed hunk-by-hunk; the 12 test edits, exact count and correct targets; the
five duplicated fixtures byte-identical to their masters; `_hashrate` moved not copied; all four line
counts; the budget diff limited to the three declared rows; no import cycle; `dashboard.css:1058`
correctly repointed. Coverage was **run**, not read: base `views.py` 171 stmts / 2 missing at 161 and
212, head `series_views.py` 78 / 2 missing at 64 and 115, and `test_series_views.py` **alone**
reproduces that — so the moved tests exercise the new module rather than coasting on a stale import.
Re-export object identity was confirmed with `is` after actually importing, not by grep. Collect-only
2120 IDs both sides; suite 2120 passed.

### 1. The bare-read count of 65 does not reproduce

| shape | this body | re-derived |
|---|---|---|
| bare read `views.X` | **65** | **62 raw matches** — 65 reproduced by nobody |
| `patch.object(views, "…")` | 4 | 4 ✓ |
| `setattr(views, "…", …)` | 4 | 4 ✓ |
| plain assignment `views.X = …` | 3 | 3 ✓ |
| `patch("…web.views.…")` | 1 | 1 ✓ |

Three independent sweeps all land on **62 raw matches**. The four other shapes match this body exactly.

I am deliberately **not** replacing 65 with a single "real" figure. An attribute-read regex also matches
prose mentions of the filename `views.py` (capturing `py`), and the two passes stripped that noise by
different rules and got **47** and **40**. Picking one would be shipping a fourth unreproduced number
into a body already being corrected for exactly that. What is solid: **62 raw, and the strict count is
somewhere in the low-to-mid 40s depending on the exclusion rule you state.**

The safety conclusion is unaffected and now has **three** independent derivations, each with the 17
moved names taken from the diff by AST rather than from this body: **zero sites name any of the 17
moved symbols through the `views` module object**, in any of the five shapes. `raising=False` /
`create=True` reproduces exactly: 3 hits, all unrelated.

This is the third consecutive cut in this family to pair a right conclusion with a wrong count. The
counts are the cheapest thing a reviewer checks; getting them wrong spends credibility the conclusion
then has to re-earn.

### 2. "A pure deletion: +0 added, 322 removed, of which 9 are the pruned imports" is wrong

`+0 added` is correct. The total removed is **367**, confirmed two ways (`git diff --numstat`, and
1228 − 861). The 322 covers only one of four hunks; it omits the 9-line import prune, the 5-line
`_hashrate` deletion and the 31-line `TestCadence` deletion (45 lines, 12%). Each individual number is
right; the "of which" is what makes the sentence false.

Relatedly, the two block-deletion hunks in `views.py` remove 4 lines of blank-line collapse beyond the
224 lines of moved content — whitespace bookkeeping, not an undeclared content edit, but "zero
undeclared" does not say so.

### 3. A sixth surface exists that the five shapes cannot see, and one site is undeclared

All five enumerated shapes describe **module-attribute** access (`views.X`). A
`from …web.views import X` statement binds the name at import time and no attribute-shaped pattern can
catch it. Found independently by me and by the second pass; enumerated by AST over every `ImportFrom`
whose module ends `web.views`:

| site | imports through the facade | status |
|---|---|---|
| `mining_dashboard/web/worker_detail.py:37` | `_gauge_series` | **declared** — this body names it; the `# noqa: F401` re-export is deliberate |
| `tests/web/test_xvb_views.py:34` | **`_mode_palette`, `build_hashrate`** | **undeclared** |

`views.py` imports 10 names from `series_views` because `build_state` needs them, so that line resolves
and binds the same objects, and both names are called directly (line 101), never through a patched
`views.` reference. **No false green, no behaviour change today.** But the claim **"Sites naming one of
the 17 moved symbols through `views`: zero"** is true only of the five shapes searched, and false as
written — and the completeness claim for the union of five shapes is false as methodology.

The forward risk, in the second pass's words, is the part worth keeping: *"a future
`monkeypatch.setattr(views, "_mode_palette", …)` intended to affect this file's call at line 101 would
silently do nothing, and the audit's own instrument would never flag it as a site to check."*

**Deferred deliberately, and the reason was measured rather than assumed.** The one-line repoint was
written and proven — `ruff` clean, suite 2120 passed, and the facade-import surface drops to the single
declared site — then **reverted**. `tests/web/test_xvb_views.py` sits at its budget ceiling of 1342 with
zero headroom; the repoint makes it 1343; the ratchet refuses it:
`file-budget: FAIL — … raises … from 1342 to 1343. Ceilings only go down.` That is the gate's output
from running it, not a recollection. Paying for it means freeing a line in a file this cut does not
otherwise touch, which is not this PR's business. It is carried into D5, which works the import surface
anyway.

### The instrument defect that nearly hid all of this

The first version of the enumeration tool I used scanned **line by line**. On the D5 subject file it
reported 154 `patch.object(mod, …)` sites; the whole-file version reports **186** — the 32 it missed are
calls the formatter wrapped, putting the module argument on its own line. **17% of sites invisible, and
the empty result looks identical either way.** It surfaced only by attacking an empty result, not by
anything going red; the tool is now whole-file and its control seeds the wrapped form explicitly.

Whether this PR's own enumeration shares that hole is **not established** — for the `views` alias both
scans agree, which is what you would expect if those call sites are simply short enough not to wrap.
Agreement is not evidence the tool was sound.

**Merging on this basis:** no code defect was found by either pass, the two documentation errors are
corrected above, and the one substantive finding is benign today and cannot be paid for inside this
PR's budget. Findings 3 and the instrument defect carry into D5.

PR #1473, opened 2026-08-29
refactor(dashboard): split the pure helpers out of data_service.py (#1105 Phase 3 D5)

VijitSingh97:
## What this is

#1105 Phase 3, cut **D5**: the pure helpers behind the `DataService` poll loop leave
`service/data_service.py` for a new `service/data_helpers.py`, and their test classes move 1:1 into
`tests/service/test_data_helpers.py`. Pure move — no behaviour change is intended and none is measured.

| file | before | after |
|---|---|---|
| `service/data_service.py` | 1852 | 1454 |
| `service/data_helpers.py` | — | 431 (new) |
| `tests/service/test_data_service.py` | 3130 | 2542 |
| `tests/service/test_data_helpers.py` | — | 622 (new) |

What moved: the xmrig-proxy row/summary parsers, the worker normalizer, the direct-stats and
proxy-summary mergers, the hashrate and window aggregators, the share-delta and audit-timestamp
helpers, the config flatten/diff pair, the host-config reader, and `WorkerLifecycle` — 16 helpers plus
one class. **`DataService` itself, its poll loop and `StateManager` are untouched**, per the cut map's
binding ruling.

## The boundary came from the dependency graph

The forward dependency closure of the row's seed names is **36 names in one run, 100-529**. Seven
names sit inside that run and outside the closure. **Six of them stay** — `_XVB_REGISTER_FAIL_ALERT`,
the three `*_CAPTURE_SEC` cadences and the two `_RIG_EDIT_*` limits are all `DataService`-only
poll-loop scheduling and rate-limit config, and the poll loop is not being split. They interleave, so
the source cut is **three byte-identical blocks rather than one**. That fragmentation is the honest
price of the ruling, not an oversight.

**The seventh is the cut's only judgement call: `_PX_LAST_SHARE_MS` is dead code.** It has exactly one
occurrence in the whole file — its own definition. It moves with its twelve `_PX_*` sibling field
indices, because it documents the same xmrig-proxy row format and moving it keeps that block
byte-identical; leaving it would split the block in two for no gain. **It is still unused after this
PR.** Deleting it is a real change and belongs in its own vehicle, not smuggled into a move.

## The proof class byte-identity cannot see

A moved function reads module globals from its **new** module, so a module-attribute patch elsewhere
can start patching the wrong module and **pass while asserting nothing**.

Exactly **one module alias exists** — `ds_mod`, at `test_data_service.py:8`. That was established by
enumerating every import statement naming `data_service` (6, one of them prose in a docstring), not by
pattern-matching. Enumeration over it, whole-file, across every tracked `.py` under `dashboard/`:

| shape | before | after |
|---|---|---|
| `patch.object(ds_mod, "…")` | 186 | 186 |
| bare read `ds_mod.X` | 43 | 39 |
| `setattr(ds_mod, "…", …)` | 2 | 2 |
| plain assignment `ds_mod.X = …` | 0 | 0 |
| `patch("…data_service.X")` | 0 | 0 |
| `getattr` / `vars` / `__dict__` | 0 | 0 |

**Sites naming any of the 37 moved names, before or after: zero.** All 186 `patch.object` sites steer
`DataService` internals, which stay put. The bare-read drop of 4 is exactly the four declared test
repoints below. `raising=False`/`create=True`: 3 hits repo-wide, all unrelated (conftest DB path, two
wizard `delenv`s).

**The instrument was proven to fire before the empty result was believed**: each shape was seeded with
a name known to move and confirmed flagged. Two shapes (`getattr` with a computed name, `vars`/
`__dict__` subscript) cannot resolve a name by construction — they report presence only, and presence
is zero here, so nothing can hide in them.

**A sixth surface is enumerated too, because the five attribute shapes cannot see it.** A
`from …data_service import X` binds the name at import time. Checked by AST over every `ImportFrom`:
**zero moved names are imported through the `data_service` facade anywhere after this cut.** The one
external importer, `tests/collector/test_pools.py:18`, is repointed to `data_helpers` rather than left
to resolve through a re-export.

## The declared edits — one in source, four in tests

- **Source, 1 edit.** `data_service.py:135` read `Wall-clock gated like the capture cadences above.`
  Those cadences stay behind, so after the move "above" is false. Repointed to name the module. Every
  other positional reference inside the moved ranges was checked and refers to something that moves
  with it (two are numeric "above", not positional).
- **Tests, 4 edits.** `TestReadHostConfig` patches `ds_mod.config` four times. This is the **safe**
  shape — an attribute on the shared `config` module object that both modules import, so a move
  rebinds nothing and it could not have broken. They are repointed to `dh_mod.config` for consistency
  with D2/D3/D4, not because they were at risk.

Both sets are applied to the **base** side of the identity checks, so those comparisons are identity
rather than "close enough".

- **Source:** all three blocks byte-identical — `_PX_*` + `_KHS_TO_HS` 18 lines (md5 `eceeae1bfea6`),
  `_XVB_WINNERS_*` 14 lines (`3a0e45897d5b`), the helpers + `WorkerLifecycle` 374 lines
  (`fdd96ae94c10`).
- **Tests:** all 14 classes byte-identical, 547 lines total, extracted from the base by AST so the
  range is the class rather than a guessed span. Each is also confirmed **absent** from the retained
  module.
- **`_totals`** (the one duplicated builder) byte-identical to its master, and the master still present.
- **Both instruments were mutation-tested and FIRED**, per block: corrupting one block leaves the
  others passing, so these are per-block checks, not file hashes.
- **Retained source half:** +17 added, −9 removed — the 17-line import-back block, 7 pruned import
  lines, 2 blank lines. Zero undeclared.
- `ruff format` was applied and shown to change **zero non-blank lines** in both rewritten files.

## Gates — what was RUN

- Suite **2120 passed / 0 failed** at HEAD — the same count as the base.
- Collect-only identity **2120 = 2120**, normalised diff **empty**, with a **positive control**
  (un-normalised diff = 143 lines). The head artifact carries **67** IDs in the new module and the base
  carries **0**, which is what proves the two collects read different trees — my first attempt ran both
  in the base worktree and the control caught it.
- Coverage **TOTAL 97%** (bar 80). **`data_helpers.py` is 160 statements, 0 missing — 100%**, so the
  moved code adds no uncovered lines at all.
- **Patch coverage 100%, 161 lines, 0 missing** against `origin/develop-v2`. The shipped gate reports
  98% over 1607 lines instead because it hardcodes `origin/develop` — that is #1460, not this change.
- `ruff check` + `ruff format --check` clean tree-wide; `lint-docs-voice` OK; `lint-operator-strings` OK.
- File-budget gate **rc=0 — after the correction round below**. As first pushed, this PR **failed**
  the gate, and that failure was the RED `Lint per-surface` check. `--generate` would have **lowered
  four rows belonging to other lanes** (`egress.py`, `ci.yml`, `tests/integration/lib.sh`,
  `tests/integration/run.sh`); all four are kept at their committed values.

## The fixture duplication is one builder here, not ~93 lines

The standing ruling duplicates shared test builders into each new module. At D2 that cost ~93 lines.
**Here it is `_totals`, a single 2-line builder** — these helpers take plain data rather than a
service, so nothing else crosses. Disclosed in the new module's docstring naming the master copy, per
the ruling's own condition; #1459 remains the tracked exit.

`_iso_now` is still used by **retained** tests (inside `TestWatchHostConfig`), so the retained module
imports it — from `data_helpers` directly, not through the facade.

## Over-engineering pass — what was considered and rejected

- **Moving the six cadence constants too**, to make the source one contiguous block. Rejected: they
  are poll-loop config and the ruling keeps the poll loop whole. Three blocks is the honest shape.
- **Splitting `TestXvbWinnersGate`**, whose last test is a wiring test driving
  `DataService._sync_xvb_winners`. Rejected: it would break the 1:1 byte-identical property the whole
  split rests on, to tidy one test. The class moves whole and this is why the new test module imports
  `DataService` — stated rather than hidden.
- **Deleting the dead `_PX_LAST_SHARE_MS`.** Rejected: not a move.
- **Real pytest fixtures instead of duplicating `_totals`.** Rejected by the standing ruling.
- **`# noqa: F401` re-exports from `data_service`** so `test_pools.py` need not change. Rejected: that
  manufactures exactly the facade-import residue the D4 review found, to save a one-line repoint.

## Correction round — what the non-author pass and CI both found

This PR was opened claiming every gate was green. **That claim was wrong.** Recorded here rather than
quietly amended, because the failure mode is more useful than the fix.

**1. The file-budget gate FAILED — and it was the RED `Lint per-surface` check.** The cut lowered the
two rows for the files it shrinks but never added rows for the two files it *creates*. Both are over
the 400-line target, so both must record a ceiling:

    dashboard/mining_dashboard/service/data_helpers.py   431
    dashboard/tests/service/test_data_helpers.py         622

This body previously read "Neither new file needs a row." That was false against the gate's own rule.
Both rows are now present; the gate is rc=0.

**Why it looked green — the gate enumerates through `git ls-files`, so it silently skips UNTRACKED
files.** With the two new modules present on disk but not yet `git add`ed, it reports
`file budget OK`, rc=0. Verified deliberately, by un-tracking them and re-running rather than by
assuming the mechanism. Two real failures were invisible and nothing went red. `git add` new files
before any lint or budget run.

**2. The base moved underneath this branch, twice.** Another lane lowered `tests/stack/run.sh`'s
ceiling 2509 -> 2289 on `develop-v2`; this branch's stale row then read as a ceiling *raise*. Fixed by
**rebasing**, not by hand-patching the row. Rebased again onto the current tip before merge. Neither
intervening range touches `dashboard/`, so every proof above still holds.

**Worth a look from whoever owns #1460: CI never saw failure (2) at all.** Its log reads
`no base ref (origin/develop or develop) resolvable; skipping the monotonic-ceiling check`, so the
monotonic-ceiling half is currently not running in CI on this lane. The local gate is stricter than CI
here, which is the wrong way round for a ratchet.

**3. The repoint owed since D4 is now PAID.** `tests/web/test_xvb_views.py` still imported
`_mode_palette` and `build_hashrate` through the `views` facade rather than from `series_views`, where
they are defined. D4 could not pay it: that file sits at its ceiling of 1342 with zero headroom. Paid
line-neutrally here -- the import is repointed, and the four real `build_state(` call sites now read
`views.build_state(` through the module alias the file already imports. The file stays **1342/1342**
and `ruff format --check` changes nothing.

Checked *before* making it, because switching a name from a direct import to an attribute lookup can
**activate a previously-inert monkeypatch**: no `setattr`, `patch.object`, dotted-string `patch`, or
plain assignment anywhere in `dashboard/` targets `build_state`, `_mode_palette` or `build_hashrate` --
with a control confirming the same patterns do find the two `patch.object` sites that exist in that
same file. `views.py:63` imports both names straight from `series_views`, so the repoint resolves to
the same object.

**4. A count correction that is not in this body, but is in the lane's own recon.** The recon said
`DataService` reads **13** moved names as plain globals. The true number is **15**, re-derived by AST
over the class body. The safety conclusion is unchanged and in fact tighter than was claimed: the
imported set and the read set are **exactly equal** -- every name imported from `data_helpers` is read
inside `DataService`, and every moved name read is imported. Corrected in the lane's notes.

## Not done

- `make lint-sh` not run (zero shell files in the diff; it is the serialized OOM path).
- `make test` in full not run — the stack/compose/integration tiers are not exercised by this change.
- Nothing seen on screen; this lane has no browser. Nothing here renders.
- The 23 uncovered lines remaining in `data_service.py` were **not** compared against a base-side
  coverage run, so I am not claiming they are all pre-existing. What is measured is that the moved code
  is 100% covered and the patch adds zero uncovered lines.


VijitSingh97:
**Post-merge narrow pass (non-author) at merged head `8c7dcc2a` — correction-round delta only: MERGE-SOUND.**

What the reviewer re-derived (against the true merge-time base `e28a96b`, the merge commit's mainline parent, not today's live ref):

- The two new file-budget rows exact-match `wc -l` at head (431 / 622), and the whole budget file shows exactly the four declared row changes between base and head — nothing undeclared.
- The rebase window (`1fbe4f5e..e28a96b`) touches zero files under `dashboard/`.
- The `test_xvb_views.py` repoint: import moved to `series_views`, four call sites now go through the `views` module alias, file stays 1342/1342. A whole-`dashboard/` AST scan for stale `monkeypatch.setattr` / `patch.object` / `patch("...")` / plain-assignment sites targeting `build_state`, `_mode_palette`, or `build_hashrate` returned zero hits — validated with a positive control (the file's two real `patch.object` sites were found) and a seeded formatter-wrapped `setattr` control (the miss mode line-based scans have).
- The 13→15 moved-name count re-derived independently by AST: exactly 15 bare reads in the `DataService` body, and the imported set is set-identical to the read set (37 moved names cross-checked).

One note for future readers: running the file-budget gate at this head against the *current* base ref reports a FAIL on `tests/stack/run.sh`'s ceiling. That is a later, unrelated mainline change moving the base ref out from under the gate — at merge time, base and head agreed on that row and the gate was green. Not a defect at this head.

Not re-verified: the `ruff format --check` claim (tool unavailable in the review environment); the pre-correction parts of the PR were reviewed in the earlier pass and were out of this one's scope.

Review record: `reviewer-20260829T014225Z` (controller-judged).

Issue #270, opened 2026-06-17
Enforce Tor-only egress at the network layer (fail-closed) — stop relying on per-app config


VijitSingh97:
## Why (surfaced validating the Wave-2 privacy flips, #256)

The stack guarantees "behind Tor" **per-app, via each app's config** — and that's fragile. Two live leaks proved it:
- **p2pool** silently dialed sidechain peers over **clearnet** when running a stale image (the #165 `--socks5` entrypoint change wasn't rebuilt) — see the e2e-rebuild issue.
- **Tari** makes **sustained direct clearnet connections** *despite* a correct `type = "tor"` config (see the Tari leak issue).

In both cases the **config-level checks passed** while traffic went direct. We only caught it with a connection-level check (`tests/integration/benchmarks/bench-verify-egress.sh`).

## What

Enforce Tor-only egress at the **network layer** so it's structural, not config-dependent — the #160 north-star ("nothing leaves the host except via Tor"). Give the app containers (monerod, p2pool, tari, xmrig-proxy) **no route to the internet except the Tor SOCKS** (`<bridge>.25:9050`), so any **direct clearnet dial fails closed** instead of leaking.

Approach options to evaluate:
- An **internal-only `mining_net`** plus a separate egress path only the `tor` container has (e.g. tor on a second network with external access; apps internal-only).
- Per-container **egress firewall** (iptables/nftables in an init/sidecar) dropping non-SOCKS, non-bridge destinations.
- Docker `internal: true` network for the apps + the Tor container bridging out.

## Acceptance
- [ ] `bench-verify-egress.sh tor` shows **0 app-container public connections even if an app is misconfigured** (a deliberately-clearnet app cannot reach the internet — the dial fails).
- [ ] Only the `tor` container reaches public IPs.
- [ ] Documented in `docs/privacy.md` as the structural guarantee behind the per-app Tor routing.

Part of #160. Contains the Tari leak structurally.

VijitSingh97:
## Design notes (after bench-host feasibility testing)

[...]

PR #1474, opened 2026-08-29
refactor(cli): split lifecycle and Tor egress into their own slices (#1105 Phase 2, P1)

VijitSingh97:
Second row of #1105 Phase 2: cuts the first two domain slices out of `lib/pithead/99-remainder.sh`.

| slice | lines | contents |
|---|---|---|
| `lib/pithead/01-lifecycle.sh` | 210 | `migrate_compose_project` through `stack_restart` |
| `lib/pithead/02-tor-egress.sh` | 188 | the Tor-only egress enforcement block (#270) |

The remainder drops 11870 → 11470 and its budget row drops with it — the machine-checked
statement that Phase 2 progressed. Neither new slice takes a row: both are under the 400
target, where an entry is itself a gate failure.

**The committed `pithead` artifact is unmodified by this commit.** That is the point of the
row, not a side effect of it.

## This is a pure move, and that was verified rather than assumed

`git diff` against the base for `99-remainder.sh` is **400 lines removed, 0 added**. The
removed block reconstructs exactly as `01-lifecycle.sh` + one blank + `02-tor-egress.sh`,
with the 400th line being the separator the build supplies. No line of logic was authored,
reordered or regrouped — order-preserving contiguous slices only.

## RUN at base `e28a96b`, not intended

- `scripts/build-pithead.sh --check` → parity OK, the artifact is exactly what 4 slices build.
- **Firing control, both directions.** A content change in `01-lifecycle.sh` reds parity
  (rc=1, and it prints the offending line). A blank-edged `02-tor-egress.sh` is refused
  **by name** — not merely by exit code. That distinction is deliberate: rc alone would be a
  vacuous assertion here, because the refusal fails closed anyway and a different message
  trips for a different reason.
- `shfmt -d` on both new slices: clean.
- `scripts/lint-file-budget.sh` → rc=0, and the tsv diff against the base is **exactly one
  row** (11870 → 11470, a lowering). The base ref was probed before *and* after the gate run
  and held at `e28a96b` across it.
- Arithmetic re-derived end to end with the gate's own counter:
  117 + 1 + 210 + 1 + 188 + 1 + 11470 = 11988 = the artifact's line count.

## The suite — stated honestly, because the base moved under it

`tests/stack/run.sh` last finished **green, rc=0, 2922 passed / 0 failed** — but at the
older tip `ff43e29`, read from the rc file rather than the log. Since then develop-v2
advanced twice, and the second of those (#1468) **changed `tests/stack/run.sh` itself**,
splitting the kernel-boot domain into its own file. So that green is a green of a
different suite composition, and I am not presenting it as a run at this base.

What does transfer, and why: `pithead` and `lib/pithead/` are **byte-identical** across
`ff43e29..e28a96b`, so the artifact under test was provably the artifact on this branch.
This PR touches nothing under `tests/`.

**A clean re-run at `e28a96b` is in flight and its result will be posted here as a comment
before merge.** Do not merge on the older green alone.

## No doc change is needed, and I checked rather than assumed

`CONTRIBUTING.md` and `docs/dev/repo-map.md` describe `lib/pithead/*.sh` generically — "the
numbered slices", concatenated "in name order" — and hard-code no slice count. Read both at
this base to confirm. The tsv row is the only number this change moves.

## Parity on a moving base

These slices are a verbatim copy of `pithead`, so any change landing on `pithead` after this
cut makes `lint-pithead-parity` red — **correctly, not flakily**. The remedy is to re-cut
from the new tip and re-run the identity proof. Never hand-patch the remainder to match.

## Review notes

- **Over-engineering pass: nothing to cut.** A pure move authors no lines, so there is no
  speculative abstraction, no hand-rolled stdlib and no dead flexibility to find. `net: -0
  lines possible.`
- **Disclosure on the gate that fronts this PR.** Its policy hook keys off the shell's cwd
  worktree rather than the branch being opened, and this session's cwd is a different
  worktree (`fix/1342-mutation-lock`, PR #1391). It therefore named that branch, not this
  one. I ran the over-engineering pass against **this** diff, which is what the gate is for;
  the branch label it records is the known defect, not a claim that this diff went
  unreviewed.
- Shared file touched: `docs/dev/file-budget.tsv`, one row, lowering. Expect a conflict if
  another lane lands a row in the same window; it is regenerated per lane, never merged by
  taking a side.

Refs #1105.


VijitSingh97:
## The suite re-run promised in the body has finished: green at this base

`tests/stack/run.sh` at base `e28a96b`, run from this branch's worktree:

```
pithead tests: 2922 passed, 0 failed     rc=0
```

**Read from the rc file, not the log** — and not from the runner's own "exit code 0", which
is the detached launcher's status and always succeeds. It said 0 three times while the suite
was still running.

That closes the one gap the body named. The earlier green was at `ff43e29`, whose
`tests/stack/run.sh` has since changed (#1468 split the kernel-boot domain out), so it was a
green of a different suite composition and was not presented as a run at this base. This one is.

## The adversarial pass also came back clean

A non-author review re-derived every claim rather than reading the ones stated here:

- reconstructed `01-lifecycle.sh` + blank + `02-tor-egress.sh` and diffed it byte-for-byte
  against the git-removed block — identical but for the trailing separator the build supplies;
- confirmed `git diff` for `pithead` is empty and re-ran `--check` itself;
- ran `shfmt` with the Makefile's own invocation, `bash -n`'d each slice standalone, and
  checked with `cat -A` that neither slice begins or ends with a blank line;
- re-counted the budget row with `awk 'END{print NR+0}'` and re-ran the gate;
- checked blast radius: nothing references `99-remainder.sh` by line offset, `scripts/release.sh`
  bundles `pithead` by name and never `lib/`, and there is no ordering hazard between the two
  slices because bash registers all function definitions before `main` runs.

No findings, no shortcuts.

## One thing a reviewer should know rather than discover

**#1391 (`fix/1342-mutation-lock`) edits the same lines this PR moves.** The overlap is
content-level, not adjacent: `stack_restart()` is line 177 of the new `01-lifecycle.sh`, and
that is exactly the function #1391's lock wiring changes. Whichever of the two merges second
must **re-cut from the new tip** — never hand-patch the remainder to match. This PR is the
one ready to go first.

PR #1479, opened 2026-08-29
fix(e2e): match compose service names as whole lines, never substrings (#1478)

### LINKED ISSUE #1478: e2e: container-presence assertions match service names as substrings — 'container up: tari' can pass on a down tari
## The defect

`assert_running_state` compares compose service names with **substring** matching in two of the
three places it checks container presence. The third place gets it right, and its comment explains
exactly why the other two are wrong.

The correct one, `tests/integration/run.sh:555-559`:

```bash
# ... "tari" is a prefix of "tari-wallet" (the payout-confirm container, #462),
# so a loose `*tari*` pattern would false-positive on a box also running that.
if printf '%s\n' "$running" | grep -qx tari; then
```

The two that do not, at `run.sh:538-542` and `run.sh:545-549`:

```bash
case "$running" in
*"$svc"*) it_pass "container up: $svc" ;;
*) it_fail "container up: $svc" "not in running services" ;;
esac
...
case "$running" in
*monerod*) it_fail "monerod absent in remote mode" "monerod is running" ;;
```

`running_services()` (`run.sh:344`) is `docker compose ps --services --status running | sort` — one
compose service name per line. The service names are `caddy dashboard docker-control docker-proxy
monerod p2pool tari tari-wallet tor wallet-rpc xmrig-proxy`, and exactly one pair nests:
**`tari` is a substring of `tari-wallet`**.

## What it costs, in each direction

- **`container up: tari` is a FALSE PASS** when the `tari` container is down and `tari-wallet`
  is up. The assertion reports the bundled Tari node as running while it is not. This is the
  expensive direction: a green that asserts nothing.
- **`monerod absent in remote mode` is a FALSE FAIL** for any service whose name merely *contains*
  `monerod`. Nothing in the compose file does today, so this half is a latent trap rather than a
  live bug — but it arms the moment a remote-mode fixture adds a node container on the same box,
  which is precisely what the remote-node coverage work needs.

## Honest scope — this is latent today, not currently lying

`tari-wallet` only starts under the `tari_payout_confirm` profile, which `pithead` activates only
when `tari.view_key` is set (`lib/pithead/99-remainder.sh:6816`). The harness supplies that from
`IT_TARI_VIEW_KEY` + `IT_TARI_SPEND_PUBLIC_KEY` (`tests/integration/lib.sh:206-207`), and grepping
the repo shows **no caller sets either** — they are read in `lib.sh`, documented, and set in
selftest fixtures, nowhere else.

So the false pass is **not firing in the routine gate today**. It arms the moment someone supplies
those keys — which is the natural next step for the payout-confirmation coverage that is already
listed as owed. That is the reason to fix it now rather than after: the trap springs when an
unrelated gap gets filled, and at that point the assertion goes quietly green instead of failing.

## What closing it looks like

One exact-line membership helper beside `expected_services`/`absent_services` in
`tests/integration/lib.sh` (the pure-logic module `selftest.sh` already drives), used at all three
sites so they cannot drift apart again — plus tier-1 cases that pin the nesting pair in both
directions, each with a mutation proof that reverting the site to a substring match reds them.

There is no coverage of any of these three sites today: `git grep` for `assert_running_state` and
`running_services` returns only `run.sh` itself and `skip-accounting.sh`.

VijitSingh97:
Closes #1478 (by hand after merge — `Closes #N` does not fire here, since PRs merge to
`develop-v2` while the default branch is `develop`).

## The defect

`assert_running_state` compared compose service names with **substring** matching in two of the
three places it checks container presence. The third got it right, and its comment says why:

```bash
# ... "tari" is a prefix of "tari-wallet" (the payout-confirm container, #462),
# so a loose `*tari*` pattern would false-positive on a box also running that.
if printf '%s\n' "$running" | grep -qx tari; then
```

`running_services()` is `docker compose ps --services --status running | sort` — one service name
per line. Of `caddy dashboard docker-control docker-proxy monerod p2pool tari tari-wallet tor
wallet-rpc xmrig-proxy`, exactly one pair nests: **`tari` inside `tari-wallet`**.

- `container up: tari` was a **false PASS** when `tari` was down and `tari-wallet` was up.
- `monerod absent in remote mode` had the same looseness in the **failing** direction.

## Honest scope — latent today, not currently lying

`tari-wallet` only starts under the `tari_payout_confirm` profile, which `pithead` activates only
when `tari.view_key` is set (`lib/pithead/99-remainder.sh:6816`). The harness supplies that from
`IT_TARI_VIEW_KEY` + `IT_TARI_SPEND_PUBLIC_KEY` (`lib.sh:206-207`), and **no caller sets either** —
they are read, documented, and set in selftest fixtures, nowhere else.

So this is not producing false greens in the routine gate today. It arms the moment someone
supplies those keys, which is the natural next step for the payout-confirmation coverage already
listed as owed — and at that point the assertion goes quietly green rather than failing. That is
the reason to fix it before, not after.

## The fix

One helper, `service_present`, beside `expected_services`/`absent_services` in `lib.sh` (the pure
module `selftest.sh` already drives), used at all three sites so they cannot drift apart again.
It matches a whole line and treats the needle as a **fixed string**, not a regex.

The `-F` was not in my first draft. A case I wrote for completeness — "the needle is a literal,
not a pattern" — failed, because `grep -qx ta.i` matches `tari`. Service names are `[a-z-]+`
today so nothing was reachable through it, but the helper is now correct rather than
incidentally correct.

## Proof it can fail

`tests/integration/selftest-service-present.sh`, **7 cases**. The fixture carries the longer
sibling `tari-wallet` alongside `tari` — without it in the fixture the assertion is blind to the
bug it exists to catch.

Mutation battery, run against the shipped helper, control green and the file restored after:

| Mutation | Result |
|---|---|
| substring match (the original defect) | **killed** by 2 cases |
| drop `-F` (needle becomes a regex) | **killed** by 1 |
| drop `-x` (line anchor gone) | **killed** by 2 |
| always true | **killed** by 4 |
| always false | **killed** by 3 |

The mutator asserts the pattern is present *before* writing and treats a non-match as a loud
failure, so a mutation that silently did not apply cannot read as a survived assertion.

## Stated limitation

The cases cover **`service_present` itself**, not the three call sites in `assert_running_state` —
that function needs a live box and is not unit-driven today (`git grep` for `assert_running_state`
and `running_services` returns only `run.sh` and `skip-accounting.sh`). The call sites are proven
by reading, not by a test. I did not add a static grep guard for them, because a text assertion
would read as behavioural coverage it does not provide.

## Budget

`selftest.sh` sits **exactly** on its recorded ceiling (686), so the new cases went into their own
file rather than raising it. `lib.sh` was trimmed to fit 692. **No ceiling was raised.**

## Run locally, from the repo root

- `tests/integration/selftest-service-present.sh` — 7 passed, 0 failed
- `make test-integration-selftest` — every suite green (`selftest: 214 passed, 0 failed`)
- `shellcheck --severity=warning -x` and `shfmt -i 4 -d` on all changed files — clean
- `make lint-file-budget lint-md lint-docs-voice lint-topology lint-operator-strings` — all OK

## Ponytail pass

Run on this diff before opening. Two findings, both applied: the empty-haystack case used a
different idiom (`EMPTY` plus an `&&`/`||` chain) from the six `check` calls above it, and the
tari site comment restated what `service_present`'s own comment now carries. `net: -4 lines.`

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -310,6 +310,10 @@ and `--list` prints it).
 - Expected containers up, unexpected absent. Every service for that config is running and
   healthy; in `remote` mode there is no `monerod` (and, independently, no `tari` when
   `tari.mode=remote`); `wallet-rpc`/`tari-wallet` appear only when their view key is set.
+  Presence is decided by `service_present`, which matches a compose service name as a whole
+  line and never as a substring ([#1478](https://github.com/p2pool-starter-stack/pithead/issues/1478)).
+  That distinction is load-bearing: `tari` is a prefix of `tari-wallet`, so a substring match
+  would report a stopped `tari` as running whenever the payout-confirm container is up.
 - `pithead status` exit code: `0` for a healthy config.
 - Dashboard reads live state. `/api/state` is reachable; Monero is synced (`done`); pruned/full
   display matches `monero.prune` ([#32](https://github.com/p2pool-starter-stack/pithead/issues/32)); the sidechain `pool.type` matches `p2pool.pool`.
--- a/tests/integration/lib.sh
+++ b/tests/integration/lib.sh
@@ -243,6 +243,11 @@ absent_services() {
     return 0
 }
 
+# Membership against running_services()'s output: compose SERVICE NAMES, one per line. EXACT-LINE,
+# never substring (#1478) — "tari" prefixes "tari-wallet" (#381/#462), so a loose match calls a DOWN
+# tari up wherever payout-confirm runs. One place, so the three call sites cannot drift apart.
+service_present() { printf '%s\n' "$2" | grep -Fqx -- "$1"; }
+
 # Human-readable pool label as the dashboard reports it, from the config pool key.
 pool_label() {
     case "$1" in
--- a/tests/integration/run.sh
+++ b/tests/integration/run.sh
@@ -537,23 +537,25 @@ assert_running_state() {
     expected="$(expected_services "$config")"
     while IFS= read -r svc; do
         [ -z "$svc" ] && continue
-        case "$running" in
-        *"$svc"*) it_pass "container up: $svc" ;;
-        *) it_fail "container up: $svc" "not in running services" ;;
-        esac
+        if service_present "$svc" "$running"; then
+            it_pass "container up: $svc"
+        else
+            it_fail "container up: $svc" "not in running services"
+        fi
     done <<<"$expected"
     if [ "$mode" = "remote" ]; then
-        case "$running" in
-        *monerod*) it_fail "monerod absent in remote mode" "monerod is running" ;;
-        *) it_pass "monerod absent in remote mode" ;;
-        esac
+        if service_present monerod "$running"; then
+            it_fail "monerod absent in remote mode" "monerod is running"
+        else
+            it_pass "monerod absent in remote mode"
+        fi
     fi
     # tari.mode is an independent axis from monero.mode (#103/#942): the bundled tari container
-    # must be absent whenever it's remote, same as monerod above. Exact-line match, not a
-    # substring case — "tari" is a prefix of "tari-wallet" (the payout-confirm container, #462),
-    # so a loose `*tari*` pattern would false-positive on a box also running that.
+    # must be absent whenever it's remote, same as monerod above. The exact-line matching all
+    # three sites depend on lives in service_present (#1478) — "tari" is a prefix of "tari-wallet"
+    # (the payout-confirm container, #462), so a substring match would speak for the wrong one.
     if [ "$tmode" = "remote" ]; then
-        if printf '%s\n' "$running" | grep -qx tari; then
+        if service_present tari "$running"; then
             it_fail "tari absent in remote mode (#103/#942)" "tari is running"
         else
             it_pass "tari absent in remote mode (#103/#942)"
--- a/tests/integration/run.sh
+++ b/tests/integration/run.sh
@@ -551,9 +551,7 @@ assert_running_state() {
         fi
     fi
     # tari.mode is an independent axis from monero.mode (#103/#942): the bundled tari container
-    # must be absent whenever it's remote, same as monerod above. The exact-line matching all
-    # three sites depend on lives in service_present (#1478) — "tari" is a prefix of "tari-wallet"
-    # (the payout-confirm container, #462), so a substring match would speak for the wrong one.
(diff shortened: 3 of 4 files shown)


VijitSingh97:
## Non-author review verdict — VERIFIED, merging

Reviewed at head `504efa5a7c0a54382b0b4948aabb03be1492e799` by a different session from the one that
wrote this PR. All checks green (`Shell tests` 11m38s, `Fake-daemon mini-stack` 2m59s), rollup CLEAN.

The PR's own body states the honest limitation that decided this review: the 7 new tier-1 cases cover
`service_present` in isolation, **not** the three call sites in `assert_running_state`, which need a
live box. Switching those sites from substring to exact-line matching is only safe if three things
hold, so I derived each one at source rather than taking the change on trust.

**1. `running_services()` emits bare compose service names, not container names.** `run.sh:344` is
`docker compose ps --services --status running | sort`, and `--services` prints the service name. This
is the claim the whole change rests on: had it printed container names (`<project>-tor-1`), exact
matching would have turned *every* presence assertion red on a live box, and no unit test in this PR
would have seen it.

**2. No `\r` contamination.** Exact matching is intolerant of a trailing CR where the old substring
match was not, so an SSH-added CR would be a silent regression that only appears on the bench. `rx`
(`lib.sh:267-280`) invokes `ssh -n` with no `-t`, so no PTY is allocated and no CR translation occurs.
Confirmed empirically as well: an `od -c` of that exact command against a live deployed stack shows
`\n`-separated names with no CR anywhere in the output.

**3. Exact matching reddens nothing that is currently green.** `EXPECTED_ALWAYS` (`lib.sh:222`) is
`caddy dashboard docker-control docker-proxy p2pool tor xmrig-proxy`; every one of those seven appears
verbatim as a whole line in the live output, as do `monerod`, `tari` and `wallet-rpc` when their modes
select them.

**On the scope of the bug.** The live service set contains two other shared-substring pairs —
`docker-control`/`docker-proxy` share the prefix `docker`, and `docker-proxy`/`xmrig-proxy` share
`proxy` — but neither bare fragment is itself a service name, so the old loose match had exactly one
reachable victim (`tari` inside `tari-wallet`), not several. That matches the PR's scoping rather than
widening it.

**The latent-not-live framing holds and should not be restated as a live false green.** `tari-wallet`
starts only under the payout-confirm path, which needs both view-key inputs, and no caller supplies
them today. The bug arms the moment someone does.

Verdict: **CONFIRMED correct.** No findings. The stated call-site gap is real but is now closed by
live evidence rather than left open.

Issue #909, opened 2026-08-13
Restore at setup: the wizard accepts an encrypted backup in place of the config form


VijitSingh97:
Part of #786 (sub-issue B; decision record in the 2026-08-13 comment there).

## Goal

The setup wizard accepts an uploaded encrypted backup + its passphrase as an alternative to filling the config form — completing the disaster-recovery loop that sub-issue A opens. Restore-during-onboarding is the pattern Home Assistant ships and the reason its recoveries are one screen.

## Mechanism

- A "restore from backup" path on the one-page wizard: upload archive + enter the emergency-kit passphrase. The host side decrypts and extracts onto `/data` before provisioning, riding the existing firstboot spool ("container asks, host decides") and the pre-seed contract.
- A rejected archive (bad passphrase, wrong format, failed validation) falls back to the normal form with the error surfaced — it never blocks setup. Same contract as a rejected pre-seed config.
- Restore is wizard-time only. No live-system restore in v1 (a running stack restoring over itself is a different, harder problem).

## Coverage

Tier 1 for the stage/spool logic (`test_wizard.py` pattern); the KVM install gate gets a restore-boot leg (install with uploaded backup → provisioned machine carries the restored identity/settings).

## Dependencies

Sub-issue A's archive format only — which already exists (`stack_backup`), so B is buildable independently. Scope cut, deliberate: the ESP pre-seed carry could transport a backup archive for fleet flows; v1 is browser upload only.

VijitSingh97:
Shipped as PR #933: the wizard's restore-from-backup path — upload + passphrase, host-side decrypt/validate/extract via stack_restore's machinery, fail-closed member audit (traversal/symlink refusal), rejected archives fall back to the form. KVM restore leg proves in the next battery.

PR #1481, opened 2026-08-29
test(stack): split the appliance-setup domain into its own file (#1105 Phase 1, develop-v2)

VijitSingh97:
Splits the appliance setup domain out of `tests/stack/run.sh` into `tests/stack/test-appliance-setup.sh`. R5 of the #1105 Phase 1 queue. Moved code is byte-identical; the only authored content in this diff is the new file's header, its one self-arm line, and the two-line source stanza.

## What moved, and why the scope is narrower than the cut map's row

The map's R5 row is cluster c2, lines 124-386 at the planning sha. Re-derived at this base, **that range contains two already-placed domain source stanzas** — `test-doctor-appliance.sh` and `test-backup.sh`. Cutting it whole would have nested both domain files inside the new one, which reassigns `_d0` in the same shell and narrows the outer `domain_ran` #1400 zero-assertion guard. All five standard instruments are blind to that by construction, so it would not have shown up in any proof below.

Excluding both stanzas splits c2 into three pieces. This PR takes the middle piece only — three contiguous sections, no stanza inside:

- `unit: firstboot wizard token + spool consume (#77 phase 3)`
- `unit: firstboot_consume_restore — restore-at-setup (#909, #786 sub-issue B)`
- `black-box: uninstall keeps the operator's files (#77 phase 1)`

The two leftovers are disclosed, not silently dropped. The `install.sh host gate (#77 phase 1)` section that the map put at c2's tail sits directly above c3's `install.sh download verification fails CLOSED (#868)` with only `test-backup`'s stanza between them; it belongs to R6, where taking it makes R6 contiguous and reunites the install.sh domain in one file rather than splitting it across two. The `render-quadlet parity vs os/quadlet fixtures (#77 phase 1)` section at c2's head is left in place — 32 lines, too small for its own file, thematically appliance/os. Both are routed to the controller so c2 is not marked done with them unhomed.

## The anchor, and why execution order is unchanged

The block is contiguous and the source stanza sits at **its own vacated position**, so no section moves relative to any other: the stanza runs between the `test-doctor-appliance` and `test-backup` stanzas exactly where the block ran. Zero reorder, structurally, not merely by measurement.

That choice is load-bearing rather than cosmetic here. The uninstall section closes by re-rendering the sandbox `.env` and `config.json`, because `uninstall -y` has just deleted them, and its own comment says the re-render is "for the sections below". Anchoring anywhere else would have made that tail a live ordering question. At the block's own position it is a no-op. (Independently: no line after the cut in `run.sh` reads `$V` at all, so what the tail feeds is now entirely in later-sourced domain files, each of which arms its own sandbox.)

## Identity proofs — rebuilt from git, both directions, with firing controls

Rebuilt out of `git show <base>:tests/stack/run.sh` at the ranges git reports, not from the extractor's own chunk file.

- **Moved text**: base lines 158-368 vs the new file below its header — identical over **211 lines, 200 non-blank**.
- **`run.sh`**: base minus 158-369, plus the three-line stanza block — identical, **1576 non-blank lines** compared.
- **Controls**: each comparison was re-run against a one-byte mutant, and each mutant was first confirmed to have actually changed the file. Both diffs fired. An empty diff from a broken command looks identical to a proof, so this is stated as a live comparison, not "diffed empty".
- **The base moved mid-flight, and the branch was rebased rather than exempted.** `develop-v2` advanced while this cut was in proof. `tests/stack/run.sh` is byte-identical at both bases (same `sha256`), so nothing the rebase brought in touches the moved text — and both identity comparisons above were then **re-derived from git at the post-rebase base**, not carried over. Both come back identical, and both mutant controls were re-fired and re-confirmed to have actually changed the file. (The first attempt at those controls was VOID: the `sed` addresses matched nothing, so each mutant was written back unchanged and read as a clean proof. They were rebuilt against lines confirmed to contain the pattern.)
- **Conservation**: base 1774 non-blank = `run.sh` 1576 + moved 200 − the 2 authored stanza lines. Every non-blank line of the base is accounted for.

## Self-arm, and the proof it is load-bearing

The moved block reads `$V` and `$WALLET` and calls `seed_env()` — all three are `build_val_sandbox()`'s, with `seed_env` defined inside that builder's body, so it does not exist until the builder has run. In `run.sh` those arrived ambiently from the far-earlier config and dashboard domains, neither of which is an appliance-setup concern. The file therefore calls `build_val_sandbox` itself. The builder is idempotent — fixed `$SANDBOX/val` path, `mkdir -p`, template copies, and it removes neither the `.env` nor the `config.json` this domain writes — so the call is a safe re-affirm as currently sourced.

- **Standalone under `set -uo pipefail`**, sourcing only `lib.sh` and this file: rc 0, **52 passed, 0 failed**, no unbound-variable abort. The pass count is the anti-vacuity control — a clean exit that ran nothing would look the same.
- **Negative control**: with the `build_val_sandbox` line commented out (mutation confirmed applied), the standalone run fails at rc 1 with `V: unbound variable`. Reported by message and stop position rather than by exit code, because on this bash `set -u` and command-not-found do not discriminate on rc.

## Whole-file audit

Run over the entire new file including the header and the seed line, not just the moved block. Names read but never assigned in-file: `PATH`, `ROOT`, `SANDBOX`, `STACK`, `V`, `VALID_TARI`, `WALLET`. All seven are named in the header. `ROOT`, `SANDBOX`, `STACK`, `VALID_TARI` were each confirmed to be column-0 assignments in `lib.sh` rather than function-scoped. Functions called: `assert_eq`, `assert_rc`, `assert_contains`, `ok`, `bad`, `run_sourced` (all `lib.sh` top-level) plus `seed_env`; the file defines none of its own. Functions were audited separately from variables — a `$NAME` pattern cannot find a function call by construction.

**One correction to the cut map**: R5's row carries a `$WALLET` note, and the lane's standing note had the last three such notes wrong and R5's "presumed wrong". It is **right** here, and wider than one site: `$WALLET` is read on **seven** non-comment lines — five `config.json` fixture renders and two restore assertions that use it as the needle. The count agrees between the new file and the original range, as byte-identity requires. R7's and R9's notes remain unverified and still need their own cut-time audits, so the standing note is a prompt to audit rather than a prediction.

## Budget

`run.sh` 1897 -> 1688 is the only row this PR changes. After the rebase the tsv also carries the peer D6 split's removal of the `test_storage_service.py` row, taken from the new base — the diff against that base is still this one row and nothing else. The new file is 243 lines, under the 400-line target floor, so it correctly gets **no row** — and the delete-the-row control therefore does not apply to it.

`--generate` also wanted to lower four rows this diff never touches (`dashboard/mining_dashboard/service/egress.py`, `.github/workflows/ci.yml`, `tests/integration/lib.sh`, `tests/integration/run.sh`). All four were hand-restored to their committed values: lowering a row this cut did not change would red every un-rebased branch in other lanes, naming a file their diff never touched. A stale-HIGH ceiling passes the per-file check, so leaving them is legal, not merely tolerated.

The monotonic half of the budget gate does not run in CI, so the proof is local: `scripts/lint-file-budget.sh` passes in a full worktree where `origin/develop-v2` resolves, and the base ref was probed before and after the run and had not moved. Control: with `run.sh`'s ceiling set below its actual size (mutation confirmed applied), the gate fails by name at rc 1.

## Suite proof

Baseline at the base sha and after-proof on this branch, run concurrently in two worktrees (the suite is parallel-safe on this box at N=2; the pair is known to contend on fork/exec, which costs wall-clock, not correctness).

- **Both rc 0. Both `pithead tests: 2922 passed, 0 failed`.** Both 3233 log lines, both 272 section headers.
- **Sorted multiset diff over the whole output: exactly two lines**, the two halves of one ed25519 keygen fingerprint — a known benign nondeterministic class (#1325). Nothing outside the benign classes differs.
- **Section-header ORDER: identical, 0 of 272 positions differ.** This is the check that matters here, because it is the only one that can see a permutation — a sorted multiset destroys order by definition, and neither a header count nor a log-line total can detect one either. It is the empirical confirmation of the zero-reorder anchor claim above.
- Per-test verdict multiset: identical.

The `2922 passed, 0 failed` pair above ran on the pre-rebase tree. The rebase brought in only peer-merged work under `dashboard/tests/`, `tests/integration/` and `docs/` — nothing under `tests/stack/` — and `run.sh` is byte-identical across the base move, so the stack-suite result carries unchanged. A full `make test` was additionally re-run on the exact rebased tree: a detached run on this exact tree is in flight as this PR opens, and its result is reported in a comment below rather than asserted here in advance. CI's `Shell tests` is the authoritative post-rebase run.

## Docs

No doc change is owed. `docs/dev/repo-map.md` describes the split generically and names no domain file, so it stays true. `docs/dev/testing-strategy.md` points its tier-1 shell rows at `run.sh` rather than at domain files, but that is systematic and pre-existing across every prior cut, already filed as #1445, and widening this diff into it was ruled against.

## What this PR does NOT do

- It does not take all of the map's c2 — two sections are deferred, both named above with a recommended home.
- It does not re-verify R7's or R9's `$WALLET` row-notes; only R5's was audited.
- The nesting hazard is argued structurally and avoided rather than demonstrated: no nested variant was built and run, because the whole point is that a nested cut would have read green.


VijitSingh97:
**Non-author adversarial pass at head `aab60bb6`: MERGE — no blocking finding, one should-fix routed to the next cut in this area.** (Reviewed head confirmed equal to the PR head at judgment time; base re-derived by the reviewer as merge-base, never accepted from a message.)

What the reviewer re-derived independently (each with a firing control where a diff or gate could read empty for the wrong reason):

- **The scope narrowing (the PR's central claim) holds, and sharper than stated**: tracing `domain_ran`/`_d0` through the counterfactual whole-range cut shows the nested stanzas would leave the extracted content and the render-quadlet-parity section with *zero* zero-assertion guard coverage — not "narrowed" — while all five instruments stay blind by construction. Both excluded stanzas confirmed at their claimed base lines; the cut range contains neither.
- **Byte-identity rebuilt both directions** from the base file at the git-reported ranges: extracted block vs. new file below its header, empty diff; full reconstructed `run.sh` vs. actual, empty diff at 1688 lines; conservation arithmetic re-derived (1774 = 1576 + 200 − 2). Both diffs proven able to fire via deliberate mutations, then reverted.
- **The mid-flight rebase claim proven**: old base found via reflog (not from any message), `run.sh` sha256-identical at both bases with a positive control showing HEAD's copy hashes differently.
- **The `.env`-tail anchor protected twice over**: zero `$V` readers after the cut (checked with a boundary-safe scan after the first regex was caught matching nothing, positive control against the base's 19 references), and every later-sourced domain file that touches `$V`/`seed_env` self-arms at its own entry.
- **Self-arm idempotency read from `build_val_sandbox`'s body** (touches neither `.env` nor `config.json`); standalone run reproduced at 52/0; the negative control's mutation verified applied before its `V: unbound variable` rc 1 was believed.
- **Header dependency audit complete, functions included** (the class a `$VAR` scan cannot find): vars and functions extracted by script both match the header exactly; the ambient-inheritance attribution verified true, not copied filler.
- **Budget row, gate (with a firing control), shellcheck (wrapper verified as a real passthrough), shfmt** — all clean; the four stale-HIGH hand-restored rows verified genuinely stale and untouched.
- **CI's independent single-runner suite pass pulled from the job log**: 2922 passed / 0 failed, matching the PR's concurrent-worktree figure from a cleaner control.

The should-fix (documentation, not code): the new file's header re-derives `build_val_sandbox` idempotency without citing module 7's adjacent warning about re-running that exact function; the reviewer's own five-file audit confirms the danger does not fire today. Deliberately not amended pre-merge — a post-verdict commit would move the head past the review. It goes to the next cut touching this area (R6), which amends the header with the module-7 citation.

Declared unchecked by the reviewer, carried rather than dropped: the live two-suite header-order diff was substituted with the strictly-stronger byte-identity reconstruction; the author's paired-run sorted-multiset claim was not reproduced (the independent CI figure was used instead); the "routed to the controller" claim has no GitHub-visible trace (it happened on the fleet's internal channel — confirmed from the controller side: both deferred sections are ruled and homed).

Review record: `reviewer-20260829T014359Z` (controller-judged). Merging on it now.

VijitSingh97:
Post-rebase `make test`, promised in the PR body — run on the exact merged head, detached, to completion.

**rc 0.** `pithead tests: 2922 passed, 0 failed` — identical to the pre-rebase run on the same cut.

Whole verdict-set diff against the pre-rebase run, taken with the *same* ANSI regex on both sides (an asymmetric one invents phantom differences), is **two lines, both attributed to the peer merges the rebase pulled in, neither to this cut**:

- `selftest-skip-accounting: 70 -> 72`. The two added rows are `no uncounted skip survives in rigforge-upgrade.sh` and `no invented skip class in rigforge-upgrade.sh` — they name the file the real-upgrade-leg change added under `tests/integration/`, which this diff does not touch.
- One ed25519 keygen fingerprint, the known benign nondeterministic class.

Nothing in `tests/stack/` moved: the stack-suite total is unchanged at 2922/0, which is the number this cut is accountable for.

Recorded here rather than asserted in the body in advance, because the run was still in flight when the PR opened.

PR #1483, opened 2026-08-29
refactor(cli): cut release-verify, status, doctor and support-bundle out of the remainder (#1105)

VijitSingh97:
#1105 Phase 2, cut P2: the release-verification, status, doctor and support-bundle regions move out of
the `pithead` remainder into five slices under `lib/pithead/`.

**`pithead` itself is byte-unchanged.** That is the whole contract of a move PR, and it is machine-checked
rather than asserted: `git status` shows the artifact unmodified, and `scripts/build-pithead.sh --check`
reports `pithead parity OK — the committed artifact is exactly what 9 slice(s) build.`

## What moved

Cut at `develop-v2` `257e129`, from `pithead` lines **729–1709**, in file order:

| Slice | `pithead` lines | Lines | Contents |
|---|---|---|---|
| `lib/pithead/03-release-verify.sh` | 729–932 | 204 | `compose_pinned_digest`, `cosign_available`, `cosign_run`, `cosign_container_path`, `verify_release_images`, `stack_upgrade` |
| `lib/pithead/04-status.sh` | 934–1143 | 210 | clearnet exposure state and banner, `dashboard_onion_status`, `dashboard_sync_progress`, `stack_status` |
| `lib/pithead/05-doctor-checks.sh` | 1145–1287 | 143 | the `# --- Diagnostics (doctor) ---` banner, the `dr_*` reporters, `clock_sync_status`, `container_engine`, `docker_boot_enabled`, the `check_*` helpers |
| `lib/pithead/06-doctor.sh` | 1289–1649 | 361 | `doctor` (346 lines) and `doctor_json` |
| `lib/pithead/07-support-bundle.sh` | 1651–1709 | 59 | `stack_support_bundle` |

`lib/pithead/99-remainder.sh` loses the same span: **11556 → 10574 lines.**

**The arithmetic conserves, and it is worth checking rather than trusting.** The five slices hold 977
lines. The remainder drops 982. The difference is exactly the five blank separators at `pithead` 933,
1144, 1288, 1650 and 1710 — the build supplies those, so they belong to no slice. 977 + 5 = 982.

## This was cut twice, and the second cut is the rule working

The first cut was taken at `7d59eec`. **PR #1391 then merged**, changing `pithead` — which makes every
slice a stale copy and reds `lint-pithead-parity`, correctly and loudly. The remedy is a re-cut from the
new tip and a re-run of the identity proof, **never** a hand-patch of the remainder, because a stale
remainder merges *textually clean* while being wrong.

So every line number above was **re-derived by content** at `257e129`; none was carried over from the
first cut. The region moved by 210 lines (the lock landed in the prelude) and `stack_upgrade` itself grew
by two, which is exactly why offsets from a previous tip cannot be trusted.

## Five slices, where the cut map says three — and why

I cut five, at boundaries **inside** the map's own regions rather than across them, which its "big regions
split at function boundaries" clause allows. The reason is the budget, and I read the thresholds at source
rather than taking them from the plan (`scripts/lint-file-budget.sh:25-26`: `TARGET_LINES=400`,
`HARD_CEILING=800`):

A three-file cut puts diagnostics, `doctor`, `doctor_json` and the support bundle in one file of **573
lines**. That is legal — it would take a new `file-budget.tsv` row — but it creates a file that is over
target on the day it is written, in a repo where ~95% of the budgeted surface is already closed to
additions, and #1105 exists to reduce exactly that. **`doctor` at 346 lines sets the floor**: it cannot be
divided without a semantic refactor, which is out of scope for this push. So the split degree is chosen to
put every new file under target:

- 204, 210, 143, 361, 59 — **no new budget rows, no file over target.**
- The adjacent merges that would give fewer files all fail the target: 05+06 = 505, 06+07 = 421.

`07-support-bundle.sh` at 59 lines is smaller than any existing slice (117/188/210). I kept it separate
rather than folding it into `06-doctor.sh`, because that merge is 421 lines and buys a budget row for the
sake of one fewer file; and rather than deferring it into P3, because P3 is already the largest remaining
cut and the map's P2/P3 boundary sits exactly where `stack_uninstall` begins.

**For the cut map, if this is ratified:** P2's row should read five files rather than three, and the
general lesson is worth carrying to P3/P9/P12 — *the split degree is set by the largest indivisible
function in the region, not by the region's line count.* Wherever a region contains a function over ~350
lines, the file holding it is nearly full before anything else joins it. I have not edited the map.

## Every boundary was asserted before the cut, not after

A slice boundary must be **exactly one blank line between two top-level blocks** — two blanks or none
breaks byte-identity, and `shfmt` strips blank lines from both file edges, so the separator has to be the
build's. All six boundaries (728, 933, 1144, 1288, 1650, 1710) were checked to be blank with a non-blank
line on each side **before** any file was written. Each new slice was then checked to have no blank first
or last line.

## The parity gate was proven to FIRE, with the mutation proven to have applied

A static assertion proves the text, not the behaviour, so the gate was mutated rather than admired:

| step | result |
|---|---|
| mutate `06-doctor.sh:1` — `doctor() {` → `doctor() { : mutation;` | `git diff --numstat` = **1 added / 1 removed**, on that one file |
| `build-pithead.sh --check` under the mutation | **rc 1**, naming the exact drift: `1289c1289 < doctor() { --- > doctor() { : mutation;` |
| restore | `git diff --numstat` empty |
| `build-pithead.sh --check` restored | **rc 0**, parity OK |

**One honest note on the instrument.** On the first cut this control was **blind**: the slices were still
untracked, so `git diff --numstat` printed nothing and proved nothing — a mutation-application check that
cannot see looks exactly like a clean one. The files are staged before the mutation now, and the gate's
own output quoting the mutated line is the independent second witness that it applied.

## What parity does NOT prove, and what the review therefore has to do

**Byte-identity is blind to slice assignment.** A cut at *any* single blank line between two top-level
blocks reproduces the artifact byte for byte — including a semantically wrong one. Nothing mechanical here
can tell you that `04-status.sh` holds the status code rather than half the doctor code; only reading can.
So the review that matters is: **does each slice's content match its name, and does the file order still
read as the CLI's structure?** The table above is the claim; the diff is the evidence.

## Over-engineering pass

Run against this diff. There is no authored code to cut: all 978 inserted lines are verbatim copies of the
983 deleted ones, and the only authored content in the change is a single `file-budget.tsv` row. The one
candidate is `07-support-bundle.sh` — a 59-line file holding one function — and folding it away costs a
budget row, as set out above. **Lean already.**

## Gates run

- **`build-pithead.sh --check`: parity OK from 9 slices** (and red under mutation, above).
- **`lint-file-budget`: rc 0.** `99-remainder.sh` 11556 → **10574** — the mandatory per-cut lowering, which
  is the machine-checked statement that the split progressed. No row rises, and no new row is added,
  because no new file is over target.
- `shfmt -i 4 -d` clean on all of `lib/pithead/*.sh`.
- `bash -n` clean on the built `pithead`.
- **Tier-1 suite (`tests/stack/run.sh`) at this head: rc 0, `2979 passed, 0 failed`.** Read from the
  rc FILE, not from the launcher — a detached launcher's own rc is always 0. Run alone on the box
  (no concurrent suite), and 2979 is the same total #1391's clean run reported at the tip this
  branch is cut from, which is the expected result for a change that moves lines without touching
  the artifact.

## Parity on a moving base — expected, not flaky

The slices are a verbatim copy of `pithead`, so **any** change that lands on `pithead` before this merges
makes `lint-pithead-parity` red. That red is the gate working, and this PR has already paid it once. The
remedy is always: re-cut from the new tip, re-run the identity proof.

## Disclosures

- **Shared file:** `docs/dev/file-budget.tsv` (one row lowered). `pithead` is untouched. `lib/` and the
  root `pithead` are covered by this lane's standing #1105-push grant.
- **Ponytail branch mismatch, disclosed as this lane always does.** The gate keys off the shell cwd, which
  the harness pins to a *different* worktree than the one this PR is cut from, so it names a branch that is
  not this one. I read the skill and ran the pass against **this** diff — the result is the
  over-engineering section above. The mismatch never means the diff went unreviewed.

## NOT proven — the greens above do not extend to these

- **`make lint` / `make lint-sh` were not run.** The fleet lint lock was not taken; `shellcheck` on this
  tree is the OOM path that has killed whole sessions on this box. The individual gates above were run by
  hand, and **CI's Shell tests job is the arbiter.** Note `lib/pithead/*.sh` is deliberately not on
  `lint-sh`'s shellcheck list — shellcheck already reads those lines in context via `pithead`.
- **No bench run.** This is not hardware evidence, and nothing here needs it: no behaviour changed.
- **No separate baseline suite run at the same tip.** The after-proof stands alone, deliberately: the
  artifact is byte-identical to the base's, so the suite cannot differ for any reason this PR could cause.
  A second concurrent run would have been contention, not evidence.
- **No non-author review yet.** One is owed at this exact head before merge, and its verdict gets pasted
  here as a comment first.


VijitSingh97:
## Non-author review verdict — MERGE-SOUND, merging

Reviewed at head `f40aec3196154925e1b4b82e7f6116461605c183` by a different session from the one
that wrote this PR, in a detached worktree at that exact sha. Verdict: **MERGE-SOUND, no blocking
findings.** Judged and executed by the controller; head re-verified unchanged at merge time.

What the review re-derived first-hand (not relayed from the PR body):

- **Slice assignment — the authored surface byte-identity cannot see.** Every function in all five
  slices read at source and judged against its file's name; all six cut boundaries checked in the
  base `pithead` (each a genuine function-group seam: closing `}` above, new comment block or
  function below — not arbitrary blank lines). `04-status` checked for overlap against
  `02-tor-egress`: none — exposure *reporting* vs firewall *enforcement* are distinct concerns.
  One informational note, already litigated: `stack_upgrade` beside the cosign primitives is the
  other half of the decision P1's own non-author review verified (recorded in the cut map), not a
  fresh call.
- **Parity, three independent routes** plus the reviewer's own firing control on tracked files:
  mutation proven applied via numstat before believing the gate's rc 1, restored to rc 0. A full
  regenerate (not `--check`) diffed clean against the committed artifact; `pithead` byte-unchanged
  vs base with a positive control on the diff command.
- **Conservation from git blobs**: 204+210+143+361+59 = 977 slice lines; remainder −982; 977 + 5
  build-supplied separators = 982. All three close.
- **Five files where the map first said three, confirmed exhaustively**: every two-file merge that
  includes `06-doctor.sh` exceeds the 400 floor; `doctor()` is one 326-line linear procedure with
  no mechanical seam. All five files under 400 → no new budget row.
- **Budget gate**: the PR's tsv changes exactly one row (the remainder, lowered). A local gate run
  at the head reds on `tests/stack/run.sh` — root-caused as the stale-base artifact: a later merge
  lowered that row on the base after this branch's cut, so the PR's *untouched* row reads as a
  raise. On the merge result the base's lower row wins (no conflict), and CI's own run — which
  checks out the merge ref and runs `lint-file-budget` in the passing "Lint per-surface" job —
  proves the merged tree green. No rebase, so the reviewed head is exactly the merged head.
- **Suite**: CI's independent count pulled from the job log — 2979 passed / 0 failed — matches the
  author's local run.
- **Wallet/topology sweep** over the full diff: zero hits, with positive controls fired first.

Disclosures checked rather than taken: the blind-first-control disclosure is accurate (untracked
slices made numstat blind; staging fixes it — reproduced, not just read), and the ponytail
sentinel mismatch matches the documented gate defect.

PR #1484, opened 2026-08-29
test(stack): split the install-verify domain into its own file (#1105 Phase 1, develop-v2)

VijitSingh97:
Sixth cut of the #1105 Phase 1 test split on the develop-v2 lane. Moves the `install.sh` host gate and the download-verification sections out of `tests/stack/run.sh` into `tests/stack/test-install-verify.sh`.

## What moved, and where the stanza went

The cut range is contiguous and contains no `source "$HERE/X.sh"` stanza, so no domain is nested inside the new file and the #1400 zero-assertion guard keeps its full reach. The `source` stanza sits at the block's **own vacated position**, so nothing in the suite runs in a different order than before.

The two sections are kept as one domain deliberately: the host gate decides whether an install may *start*, the verification path decides whether it may *finish*, and a regression that moved a check from one to the other would otherwise straddle two files.

## The substantive change: `$REL` is now seeded

The moved block does `source "$REL"` to reach `release.sh`'s `append_bundle_sha256()`. **Nothing in the block assigns `$REL`.** In `run.sh` it arrived ambiently, because `test-release-signing.sh` and `test-release.sh` are sourced earlier and each set it for their own reasons — a pure ordering accident. Sourcing in place preserves that accident, so byte-identity, header order, the sorted header multiset and the pass totals are **all blind to it by construction**; the suite is green either way.

The new file's header seeds it with the value its assigners already use (`"$ROOT/scripts/release.sh"` — verified identical at every assignment site in the tree). Because the install-verify stanza sits after both release stanzas, the seed is a re-assignment of an already-set value **in the full suite** and is load-bearing **standalone**.

Proven load-bearing, with a firing negative control:

| arm | rc | verdict |
|---|---|---|
| seed present | 0 | **22 passed, 0 failed** |
| seed commented out (mutation confirmed applied) | 0 | `REL: unbound variable`, `✗ sha256 match is announced` — **21 passed, 1 failed** |

**`rc` is 0 in both arms** — the failure is visible only in the verdict, never the exit code. And 22 is the *right* number, not merely non-zero: this domain contributes exactly 22 assertions inside the full-suite baseline too (counted between its first header and the next section's).

## Proof

**Byte-identity, re-derived from git rather than from the extraction** — the moved text was rebuilt out of `git show <base>:tests/stack/run.sh` at the cut ranges and diffed against the new file below its header, and the expected `run.sh` was rebuilt the same way. **Both diffs empty.** Moved text 120 lines / 116 non-blank; `run.sh` 1462 non-blank. Conservation closes: base 1576 = 1462 + 116 − 2.

**Suite:** **rc 0, 2979 passed / 0 failed** (baseline, `~/dev/pithead-split-proof` at the base) and **rc 0, 2979 passed / 0 failed** (after, this branch). Both logs are 3292 lines. The whole-output diff, normalised for ANSI, worktree path and mktemp paths only, is **exactly two lines** — the two halves of one ed25519 keygen fingerprint (#1325), a known-nondeterministic class. Nothing outside it. The `CONFIG_FILE: readonly variable` warnings shifted by **zero**, which is right by construction rather than by arithmetic: they are emitted by `test-appliance-identity.sh` at its own lines, and this cut does not touch that file.

**Ordered header comparison** (#1398 — the one instrument that can see a permutation; a sorted multiset, a header count and a pass total are all blind to reordering by construction): **0 of 274** positional mismatches. Identical header set, identical header order, 274 headers on both sides.

**Budget:** `tests/stack/run.sh` 1688 → 1570 is the only row changed. The new file is 152 lines, below the 400-line floor, so it correctly has no row. `--generate` also wanted to lower three rows this diff never touches — `dashboard/mining_dashboard/service/egress.py`, `.github/workflows/ci.yml`, `tests/integration/run.sh` — and all three were **hand-restored stale-HIGH on purpose**: lowering a row another lane owns reds their un-rebased branch naming a file their diff never touched. The gate passes; the control fires (delete the `run.sh` row and it fails by name). `git rev-parse origin/develop-v2` probed unmoved before and after the gate run.

**Lint:** `shfmt -i 4 -d` and `shellcheck --severity=warning` clean on the new file; `make lint` passes end to end (rc 0).

## Also in this diff

An 11-line addition to `tests/stack/test-appliance-setup.sh`'s header, the should-fix routed here from the R5 review. `test-doctor-appliance.sh` — sourced one stanza above — warns explicitly against re-calling `build_val_sandbox()`, and that warning now gets cited next to the call it does not reach. **I re-derived the claim rather than relaying it**: with comments stripped, exactly five later-sourced domain files read `$V`/`seed_env`, and all five self-arm at their own entry. (My first sweep found four more apparent readers; all four were comment-only mentions — one of those comments literally says the file never touches `$V`.)

## What I did NOT do, and one disclosure

- The two suite runs were launched **concurrently** in separate worktrees. Concurrency is proven safe on this box at N=2 and no further; the pair is known to contend on fork/exec, so it runs slower than serial, which is expected rather than a fault.
- **The after-run was launched before a one-word edit to a comment line in the new file's header** (dropping a count of another file's contents, which this lane's rules forbid because it goes stale). The proven tree therefore differs from the committed tree in that comment alone: the change is inside a `#` line, the file's line count is unchanged at 152, and the moved text was re-verified byte-identical after the amend. No executable text differs.


VijitSingh97:
## Non-author review verdict — MERGE, merging

Reviewed at head `27fd38d0767a7400c507c3c7570264da2b5a41f9` by a different session from the one
that wrote this PR, in a detached worktree at that exact sha. Verdict: **no blocking findings;
every specific, checkable claim in the PR body re-derived exactly as stated.** Judged and executed
by the controller on the union of that review, the author lane's own re-derived proof set, and the
controller's own checks below; head re-verified unchanged at merge time.

What the review re-derived first-hand (not relayed):

- **Contiguity and the nested-stanza hazard**: the diff is one 120-line hunk; the cut range holds
  zero domain stanzas (positive control: 32 fire on the full file), and the new file's stanza sits
  at the exact vacated position between the two unmoved neighbours. The planning-time concern that
  a neighbouring stanza sat inside the range did not survive contact with the actual base file.
- **The $REL seed** — the substantive change: an independent reads-but-never-assigns sweep over the
  moved range (comments stripped, heredoc bodies excluded as another script's text) confirms $REL
  is the *only* unassigned read; all four assignment sites in the tree use the identical string;
  and the firing control re-ran under the reviewer's own driver — 22/0 seeded, 21/1 with the seed
  removed (mutation diff-proven first), rc 0 in *both* arms, the unbound-variable red visible only
  in the verdict line.
- **Byte-identity and conservation, rebuilt from git**: both directions empty with a shown-firing
  positive control; every number closes to the digit (1576 = 1462 + 116 − 2 non-blank;
  1688 − 120 + 2 = 1570 total, the +2 being the replacement stanza).
- **Ordered header comparison via an independently-written resolver**: 0 of 274 positional
  mismatches — the one instrument that can see a permutation.
- **Budget**: exactly one tsv row changed (run.sh 1688 → 1570); the three untouched rows
  `--generate` wanted to lower are confirmed hand-restored exactly; gate rc 0 with the base ref
  probed unmoved both sides and a firing delete-the-row control.
- **The R5 should-fix rider**: comment-only (+11 `#` lines), the cited warning verified at source,
  and the five-file self-arm claim re-derived independently (same five files, all arming on the
  line before first use), with one of the four comment-only false positives spot-checked.
- **Wallet/topology sweep**: zero hits, with positive controls — including the known trap where a
  synthetic wallet string fails to fire the shape pattern (a real fixture value was used instead).

Two residuals, named and closed rather than laundered:
- The review ran with CI incomplete; **CI has since finished fully green, and its independent
  suite count — 2979 passed / 0 failed — matches the author's both arms.**
- The disclosed proven-tree/committed-tree gap (a one-word comment amend after the after-run
  launched) is unverifiable from git — no pre-amend object exists. It is bounded: the body is
  independently proven byte-identical to the base cut range, and the header's only executable line
  is the seed, verified correct. CI's green run at the merged content stands as the suite proof
  for what actually ships.
- The base moved mid-review (this PR's base gained another merge). Checked at line level: the two
  tsv edits are on different rows, every other touched path is disjoint, and the budget gate's
  one both-ways row (#1464, verified at source: a rise must match the file's real count) behaves
  as designed on the un-rebased branch. Merged without rebase so the reviewed head is exactly the
  merged head.

Issue #786, opened 2026-07-27
Appliance feature parity: everything host-CLI-only is impossible on a shell-less box


VijitSingh97:
## The finding

The control gate's design is correct and must not be weakened: wallets, view keys, auth,
onion exposure, tokens, per-rig descriptors and the security perimeter are deliberately
**host-CLI-only** — the dashboard container must never be able to commit them, because a
compromised container could then swap a payout wallet or exfiltrate a view key.

But the design's escape hatch is "edit config.json on the host and run ./pithead apply" —
and the appliance HAS no host access: no shell, SSH disabled. Every host-only item is
therefore not "inconvenient" on the appliance, it is **impossible after first boot**. The
wizard covers most of them at setup; changing them later means a reinstall.

## What an appliance operator cannot do today

| Lost capability | DIY path | Severity |
|---|---|---|
| Change the payout wallet | edit + apply | The most important operator action over a machine's life |
| Enable/change payout confirmation (view keys, scan heights) | edit + apply | Flagship feature, post-setup |
| Add/edit a rig in workers.list (per-rig control/upgrade) | edit + apply | Fleet management |
| Change the dashboard password, enable the onion | edit + apply | Security lifecycle |
| Add/change Telegram or Healthchecks after setup | edit + apply | Alerting lifecycle |
| Switch node mode local↔remote, change remote hosts | edit + apply | Topology changes |
| Encrypted backup / restore (wallet settings, Tor keys, config) | pithead backup/restore | Disaster recovery — currently NOTHING exports state off an appliance |
| Support bundle, doctor detail, secrets rotation | CLI | Supportability |

## The designed answer, and the appliance-native one

The gate's own comments name the plan: **#338 out-of-band approval (Telegram approve/deny
with deny-on-timeout) drops in where the refusal happens today**, turning perimeter changes
into dashboard-initiated + second-factor-approved. That is the right general mechanism.

An appliance-native complement worth designing alongside it: **physical presence as the
second factor** — the machine already reads `pithead-config.json` from a FAT partition at
first boot; extending that to "a config on inserted removable media, confirmed on the
console, applies on any boot" gives holders-of-the-hardware a host-equivalent channel with
no network trust at all. That is what host-CLI-only actually means operationally: someone
with hands on the machine.

Backup/restore needs its own dashboard surface regardless (download an encrypted
backup.tar.zst; restore at wizard time from an uploaded one — the restore half naturally
belongs to the wizard's pre-seed path).

[...]

PR #1488, opened 2026-08-29
refactor(cli): cut uninstall, first-boot, the disk installer and the wizard out of the remainder (#1105)

VijitSingh97:
Phase 2 cut P3 of #1105, on top of P2 (#1483). Order-preserving contiguous slices of `pithead`
1711-3065 into five files under `lib/pithead/`. The shipped artifact does not change: the
committed `pithead` is byte-identical to what the fourteen slices build, and `git status` shows
it unmodified.

| slice | `pithead` lines | lines | content |
|---|---|---|---|
| `08-uninstall-firstboot.sh` | 1711-1929 | 219 | `stack_uninstall`, the wizard's token mint and submission consumption, the machine-role marker |
| `09-restore-apply.sh` | 1931-2125 | 195 | `restore_carry_dir`, `restore_apply`, `firstboot_consume_restore` |
| `10-installer-preseed.sh` | 2127-2406 | 280 | the `--- disk installer` and `--- pre-seeding from the installation medium` sections entire |
| `11-baked-images.sh` | 2408-2508 | 101 | `repair_broken_image_store`, `load_baked_images` |
| `12-firstboot-wizard.sh` | 2510-3065 | 556 | `stage_wizard_spool`, `firstboot_wizard` |

`99-remainder.sh` drops 10574 -> 9218, the per-cut lowering every Phase 2 PR owes. The
arithmetic conserves: 1351 slice lines plus the five blank separators at 1930, 2126, 2407, 2509
and 3066, which the build supplies and no slice carries — 1351 + 5 = 1356 = 10574 - 9218.

## What was RUN

- **Parity**: `scripts/build-pithead.sh --check` rc 0, "the committed artifact is exactly what
  14 slice(s) build". `pithead` absent from `git status`.
- **The parity gate proven to FIRE, not admired.** Slices STAGED FIRST, then
  `12-firstboot-wizard.sh:1` mutated: `git diff --numstat` gives 1 added / 1 removed on that one
  file, and `--check` returns rc 1 naming `2510c2510` and quoting the mutated line. Restored:
  empty numstat, rc 0. Two witnesses, because on P2's first cut this control was BLIND — the
  slices were untracked, so `numstat` printed nothing and "proved" a mutation that it could not
  see.
- **`bash -n` on each slice individually** — all five parse. This, not byte-identity, is what
  proves no boundary fell inside a heredoc: concatenation would restore a heredoc split across
  two files and parity would still pass.
- **`shfmt -i 4 -d lib/pithead/*.sh`** clean (the Makefile's own flags), and `bash -n pithead`
  clean.
- **Boundaries asserted BEFORE any file was written**: exactly one blank line with non-blank on
  both sides at 1930, 2126, 2407, 2509 and 3066 — and re-confirmed afterwards against the shipped
  artifact, where all five of those lines are blank.
- **Every range in the table above was verified by extraction, not derived by arithmetic.** For
  each slice, `awk 'NR>=a&&NR<=b' pithead` was `cmp`'d against the committed slice file: all five
  exact. The offsets were computed from the remainder independently, and the parity gate's own
  `2510c2510` under mutation agrees with the table's third column — two routes to the same
  numbers.
- **File budget: rc 0**, both halves, at the head, against the base named below.

## Suite after-proof

**rc 0, `pithead tests: 2979 passed, 0 failed`**, 3292 output lines, 274 section headers — run at
this exact head (`d56419e`) in a worktree checked out to it, not to a reconstruction, and read
from the **rc FILE**, never a log tail (a detached launcher's own rc is always 0).
Log: `~/appliance-suite-logs/p3-after-20260829T023750Z.log`.

Two checks on that green, because a pass total is a weak instrument:
- `grep -c '✗'` on the log is **0**. A case-insensitive `FAIL` grep returns 219 — every one a
  test *name* containing the word (`tor-running check: tor down + mining up -> FAIL ✓`). Worth
  stating so nobody re-derives that alarm.
- The run demonstrably included #1484's restructured suite (`tests/stack/test-install-verify.sh`
  present in the run's own worktree), so it is not a stale-suite pass.

An earlier run at `ea67c8f` also returned rc 0 / 2979 / 0 — but #1484 merged and changed
`tests/stack/` while it was in flight, so it is quoted here only as a second data point, not as
the after-proof. Identical totals across the two are expected rather than suspicious: #1484 is a
pure domain move, and this PR changes no behaviour at all.

Concurrency: N=2 with the tests lane's own run, counted by `/proc/<pid>/cwd`.
`pgrep -f 'tests/stack/run.sh'` overstates it — it matches the probing shell's own command line.

## The one new budget row, and why it is unavoidable

This is the first Phase 2 cut that ADDS a `docs/dev/file-budget.tsv` row.
`lib/pithead/12-firstboot-wizard.sh` takes one at its real line count, 556. `firstboot_wizard`
is **521** lines by itself (`12-firstboot-wizard.sh:36-556`, machine-counted, not eyeballed);
dividing it is a semantic refactor and out of scope for this push, so its file cannot go under
the 400 target however the region is split — even alone, without the 24-line
`stage_wizard_spool` that shares it. The other four are 219/195/280/101 and take none.
Thresholds read at source (`scripts/lint-file-budget.sh:25-26`): target 400, hard ceiling 800.

## Budget proof — stated honestly, base named

The monotonic half is LOCAL-ONLY in CI (#1452), so it was run locally against base
**`976a4ab`** (the #1484 merge) at 2026-08-29 ~02:36Z, and `origin/develop-v2` was probed
before and after the run — unmoved — because every worktree shares one `.git` and any lane's
fetch moves the base mid-run. A budget proof is only as fresh as the base it was taken against.

This branch has been rebased twice, onto `ea67c8f` (#1483/P2) and then onto `976a4ab`
(#1484/R6), and the gates above were re-run in full at each head rather than inherited. Neither
merge touched `pithead`, `lib/` or `scripts/build-pithead.sh` — verified with
`git diff --stat <old>..<new> --` on exactly those paths, empty both times — so no re-cut was
needed. Had either touched the artifact, the slices would be a stale copy and the remedy would
be a re-cut from the new tip, never a hand-patch: a stale remainder merges TEXTUALLY CLEAN.

The tsv diff against that base is exactly two rows: `99-remainder.sh` 10574 -> 9218 and the new
`12-firstboot-wizard.sh` 556. No foreign row is touched, added, removed or reordered.

**Worth recording because it cost a diagnosis.** This cut was taken on P2's head before #1483
merged, and on that base the gate RED naming `tests/stack/run.sh` — a file this diff never
touches. Both arms failed for one cause: the base predated #1481, so the tree carried the row at
1897 against develop-v2's 1688; leave 1897 and the ceiling "rises", take 1688 and the file
(still 1897 lines there) exceeds it. Rather than assert that a rebase would clear it, the
post-rebase state was built in a scratch worktree and measured first — green — and the real
rebase then reproduced that exactly. The red was the ratchet working; there is no finding to
file. The same shape was independently hit and root-caused by #1483's non-author reviewer on the
same afternoon.

A cut at any single blank line between two top-level blocks reproduces the artifact byte for
byte, **including a semantically wrong one**. Byte-identity does not validate slice ASSIGNMENT.
Only reading does, so here is what reading found, including what it found against me:

- **`08` carries three source sections, not one, and that is deliberate.** The source's own
  `# --- the machine role` header does not bound its own contents: `wizard_keep_failed_config`
  and `wizard_setup_failed` sit inside it, and the rig-role pair follows. So no boundary inside
  1711-1929 yields a pair of cleanly nameable files — one honestly-named file beats two
  mis-named ones. The alternative split at 1929 would have produced a file called
  "machine-role" containing two wizard-failure functions.
- **`10` contains one function its name does not cover.** `preflight_remote_nodes` dials the
  candidate config's remote nodes before the wizard closes — wizard submission validation, not
  disk-installer machinery. It sits at file position 234 of 280, between two installer
  functions, so order-preserving contiguous slicing cannot relocate it without a semantic move
  that is out of scope. Disclosed rather than left for the reviewer to find.
- `09`'s orphaned doc block (the `#909 / #786 sub-issue B` paragraph that documents
  `firstboot_consume_restore` but sits ten lines above `restore_carry_dir`) travels into the
  same file as the function it documents.

## Over-engineering pass, run BY HAND — and the mismatch that made that necessary

The ponytail gate keys off `git rev-parse --abbrev-ref HEAD` in the pinned shell cwd rather than
the branch the PR is opened from (`common/tooling-traps.md`), so its prompt named an unrelated
branch and it never checks that a review happened. Disclosed rather than allowed to stand in for
one. The pass was run by hand against the actual diff:

**There is no new logic to over-engineer, and that is measured, not asserted.** Splitting the
diff into its added and removed line multisets: 1356 lines removed, 1351 added, and **zero added
lines are absent from the removed set** — a pure move. The five removed-but-not-added lines are
all BLANK: exactly the separators the build supplies. Outside `lib/pithead/` the entire diff is
**2 tsv lines**. (Computed with `awk '/^--- /{next} /^-/{print substr($0,2)}'`, not
`grep '^-[^-]'`, which silently drops removed blank lines and would have made a pure move read
as a content difference.)

**Is five files more than the region needs?** Order-preserving slicing means only ADJACENT
merges are available; each costs one separator line:

| merge | lines | verdict |
|---|---|---|
| 08+09 | 415 | over the 400 target — buys a new row |
| 09+10 | 476 | over target — buys a new row |
| **10+11** | **382** | **legal, and the only one that reduces the count for free** |
| 11+12 | 658 | legal, but inflates an already-budgeted row from 556 to 658 |

So exactly one merge would give four files at no budget cost, and it is declined on naming.
`11-baked-images.sh` governs the CONTAINER image store, not disk installation: `load_baked_images`
is called from the boot path at `pithead:12237`, far outside this region, so folding it into a
file named `installer-preseed` would mis-name boot machinery as installer machinery. This follows
P2's ratified precedent, which chose five accurate names over three and shipped a 59-line slice
rather than fold it into `doctor` — 101 lines is larger than that precedent's smallest.

## What was NOT done

- **`make lint` / `make lint-sh` were not run**, and the lint lock was not taken — `shellcheck`
  is the OOM path that has killed whole sessions on this box. The skip is argued, not assumed:
  `make -n lint-sh` shows exactly one `shfmt -i 4 -d` invocation reaching `lib/pithead` (14
  files) and **neither** `shellcheck` invocation listing it, so P3's entire new exposure under
  that target is the shfmt pass — which was run, clean. CI's Shell tests job is the arbiter.
- **No bench run.** Nothing here changes behaviour; the artifact is byte-identical.
- **No non-author review yet.** Everything above is author-verified only.

Refs #1105


VijitSingh97:
**Non-author adversarial verdict: MERGE** — at head `d56419eb4942ef89a7b9d5f903190c02896b0bd5` (re-derived from `gh pr view` + `ls-remote` immediately before posting; unchanged from the reviewed head).

Independent gate pass (ephemeral reviewer, controller-judged), full text in the fleet's `answers/reviewer-20260829T025254Z.md`. What the pass re-derived itself, rather than relaying from the PR body:

- **Parity/byte-identity**: `build-pithead.sh --check` rc 0, full regenerate byte-identical to the committed artifact, committed `pithead` byte-identical to base's (12284 lines both sides) — with a proven-applied mutation as the firing control (numstat shown, gate rc 1 naming the mutated line, restored to rc 0).
- **Pure-move**: recomputed independently — 1356 removed / 1351 added, zero added lines absent from the removed set, exactly 5 removed-not-added lines, all blank by `cat -A`.
- **Conservation**: slice counts from git blobs (219+195+280+101+556=1351), remainder 10574→9218, wizard function machine-counted 521. All five boundaries re-extracted from base's `pithead` by range and `cmp`'d byte-for-byte against the committed slices — 5/5 match; all five separators blank with non-blank neighbours.
- **Budget**: tsv diff is exactly the two expected lines; gate source read — a new row never enters `check_monotonic` and is validated as an ordinary ratchet row; gate run rc 0 with a firing control (row deleted → rc 1 naming the file); base ref probed both sides, unmoved.
- **Both disclosed soft spots verified at base**: the `machine role` header demonstrably does not bound its contents (it spans into slice 09), and `preflight_remote_nodes`' sole call site is inside `firstboot_wizard` — the contiguity argument for keeping it in slice 10 is structural, not rhetorical. The declined 10+11 merge re-verified: 382 is the only adjacent merge ≤400; `load_baked_images`' boot-path call site traced one hop further than the PR body (`os/overlay/pithead-boot` → `load-images`).
- **CI**: Shell tests pass, 2979/0 pulled from CI's own job log independently of the author's local run.

Non-blocking notes, recorded rather than actioned: (1) `load_baked_images` also has a second, in-region call site inside `firstboot_wizard` itself — the PR body names only the boot-path one; the naming verdict is unchanged. (2) The author's local after-proof log and the ponytail-gate narrative are relayed, not independently verifiable — bounded by CI's independent suite run.

Merging by merge-commit; reviewed head = merged head.

Issue #1489, opened 2026-08-29
compact-chain.sh header claims the source is never modified; monero-blockchain-prune swaps the databases


VijitSingh97:
`tests/integration/compact-chain.sh`'s header states a safety property the underlying tool does not have. Following it against a live node's data dir renames the live chain out from under a running `monerod`.

## What the header claims

```
#  * It only READS the source, through a consistent LMDB snapshot, so it is safe to run while
#    monerod is up and mining — zero downtime during the copy; the source is never modified.
```

and then tells the reader to perform the swap by hand afterwards:

```
#   docker stop monerod
#   mv <data-dir>/lmdb <data-dir>/lmdb.bloated && mv <data-dir>/lmdb-pruned <data-dir>/lmdb
```

## What the tool actually does

`monero-blockchain-prune` (v0.18.5.0) is **copy-then-swap**, not copy-only. Extracted from the binary:

```
Swapping databases, pre-pruning blockchain will be left in
-old and can be removed if desired
Blockchain pruned OK, but renaming failed
Error renaming
```

Those strings only have a meaning on a path that renames `<data-dir>/lmdb` to `<data-dir>/lmdb-old` and moves the freshly built `<data-dir>/lmdb-pruned` into its place. The `--copy-pruned-database` flag does not change this: its help text is *"Copy database anyway if already pruned"*, so it lifts an early-exit guard on an already-pruned source — it is not a copy-only mode switch.

## Independent corroboration, from this repo

`build-pruned-chain.sh:74` reads the pruned result back from `lmdb/`, not from `lmdb-pruned/`:

```sh
say "pruned size: $(du -h "$DST_DIR/lmdb/data.mdb" | cut -f1)"
```

That line is only correct if the tool swapped the directories itself. So the two scripts contradict each other about the same binary, and the one that agrees with the binary's own strings is the one that expects a swap.

## Why this was never caught

The only recorded run of this script on the harness box was killed mid-copy after ~5.6 h (`rc=143`). It never reached the swap, so the header's claim was never exercised to completion.

## Consequence

A reader who trusts the header and points the script at a live node's data dir gets, on completion: the live chain renamed to `lmdb-old`, a pruned chain installed in its place, and a `monerod` whose next restart silently comes up pruned — plus a full-size stale copy consuming disk. The running process keeps its open file descriptors, so nothing fails loudly at the time.

## The safe way to build a pruned copy from a live chain

[...]

PR #1490, opened 2026-08-29
fix(e2e): compact-chain.sh's header denied the swap the tool performs (#1489)

VijitSingh97:
Addresses #1489 — kept open deliberately until the guarded-build outcome is posted there (closing keyword removed at merge by the controller, per the standing ruling).

## The defect

`tests/integration/compact-chain.sh`'s header asserted a safety property the underlying tool does not have:

```
#  * It only READS the source, through a consistent LMDB snapshot, so it is safe to run while
#    monerod is up and mining — zero downtime during the copy; the source is never modified.
```

`monero-blockchain-prune` (v0.18.5.0) is **copy-then-swap**. After building `<data-dir>/lmdb-pruned` it renames `<data-dir>/lmdb` to `<data-dir>/lmdb-old` and moves the pruned DB into place.

Pointed at a live node's data dir, that renames the live chain out from under a running `monerod`. Nothing fails at the time — the process keeps its open descriptors — and the next restart silently comes up pruned, with a full-size stale chain beside it eating disk. The header then told the reader to perform the swap by hand, which is the tell that its author did not know the tool does it.

## What was RUN

- **Extracted the code path from the binary.** It carries `Swapping databases, pre-pruning blockchain will be left in`, `-old and can be removed if desired`, `Blockchain pruned OK, but renaming failed` and `Error renaming`.
- **Read the flag's help text.** `--copy-pruned-database` is *"Copy database anyway if already pruned"* — it lifts an early-exit guard on an already-pruned source, it is not a copy-only mode switch.
- **Corroborated by an independent route inside this repo.** `build-pruned-chain.sh` reads its result back from `$DST_DIR/lmdb`, not `$DST_DIR/lmdb-pruned` — only correct if the tool swapped. The two scripts contradicted each other about the same binary, and the one agreeing with the binary's strings is the one expecting a swap.
- **Proved the bind-mount guard before relying on it**, as a positive control rather than an assumption:
  ```
  mv: cannot move '<build-dir>/lmdb' to '<build-dir>/lmdb-renametest': Device or resource busy
  ```
- **Lints, all green:** `shellcheck --severity=warning` on both changed scripts; `shfmt -i 4 -d` over the Makefile's exact glob; `make lint-md` (0 errors, 47 files); `make lint-docs-voice` (40 prose docs); `make lint-file-budget`.
- A pruned-chain build is running right now on the harness box under the bind-mount guard, reading the live chain with monerod up and mining. The live node stayed `synchronized` at the same height across the start of the copy.

## What was NOT run, and what is inferred

- **`make lint-sh` was not run** — it is this box's OOM path and is serialized across lanes. Substituted with targeted `shellcheck` over both changed files plus `shfmt` over the Makefile's exact glob, which is the standing substitute here.
- **No new test.** These are one-shot operator tools with no assertions to hang a tier on, and a test that asserts the content of a comment would prove nothing. The evidence that carries this change is the binary's own strings plus the `EBUSY` guard demonstrated above. Naming that rather than inventing coverage.
- **Inferred, not yet confirmed:** that the swap fires at the end of *this* code path, as opposed to only some other path. Everything above establishes the tool contains the swap and that the header's flat "the source is never modified" is unsupportable; it does not by itself prove the rename executes on every completion. The running build settles it safely — under the bind mount, a firing swap logs `Blockchain pruned OK, but renaming failed`. I will post the result on #1489 either way.

The fix does not depend on that outcome: a header asserting the source is never modified, on a binary that contains explicit code to rename it, is wrong as written.

## Scope

- `tests/integration/compact-chain.sh` — header corrected; bind-mount recipe documented (lane path).
- `tests/integration/build-pruned-chain.sh` — notes that step 4 is safe because it targets the copy, and why the size is read back from `lmdb/` (lane path).
- `docs/dev/release-server.md` — the same false premise appeared there, telling the reader to "swap the compact copy in during a ~2 min window". Shared `docs/`, disclosed here as required.
- `docs/hardware.md` — one clause folded into the existing sizing parenthetical: stop monerod before running the tool. Shared `docs/`, disclosed here (addendum added at merge; delta commit `1e12d6c`).
- `tests/integration/testbench-README.md` — the fourth-artefact recipe replaced with the guarded bind-mount recipe matching `compact-chain.sh`'s header (delta commit `b9d0549`, addressing the round-one blocking finding).

https://claude.ai/code/session_01A8qHpoXBs7pMX7Tq19GEiq

## Ponytail pass

The plugin's skill is not registered in this session, so I read its definition and ran the pass against the diff myself:

```
release-server.md:L333-350: shrink: the bind-mount recipe duplicates compact-chain.sh's
  header verbatim — the exact duplication that let the original false claim drift between
  script and doc. Warning + pointer to the script.
build-pruned-chain.sh:L12-17: shrink: 6-line note, 3 lines carry it.
net: -13 lines possible.
```

Both applied in `398a3d0`. The first finding indicts the fix itself: #1489 exists *because* a safety claim was duplicated between a script header and a runbook and the copies drifted. Restating the recipe in both places would have rebuilt that trap, so the runbook now carries the warning and points at the script for the recipe — one canonical copy.

### DIFF
--- a/docs/dev/release-server.md
+++ b/docs/dev/release-server.md
@@ -334,10 +334,27 @@ tests/integration/run.sh --host you@server --dir pithead --readiness
 > NOTE: a pruned chain's file stays large. An in-place prune does not shrink the LMDB file: it
 > stays at the full-chain high-water mark (~250 GiB) with the freed space sitting as internal
 > free pages (Monero reuses them as the chain grows). To reclaim it you must rewrite the DB with
-> `monero-blockchain-prune --copy-pruned-database` (see
+> `monero-blockchain-prune` (see
 > [`compact-chain.sh`](../../tests/integration/compact-chain.sh)). It's slow (it copies every block
-> over hours), though it reads through a snapshot so monerod keeps mining; you then swap the
-> compact copy in during a ~2 min window. The generic `mdb_copy -c` does not work: Monero ships a
+> over hours), though it reads through a snapshot so monerod keeps mining.
+>
+> **The tool is copy-then-swap, so it does not leave the swap to you.** After building
+> `<data-dir>/lmdb-pruned` it renames `<data-dir>/lmdb` to `<data-dir>/lmdb-old` and moves the
+> pruned DB into place. Pointed at a live node's data dir it therefore renames the live chain out
+> from under a running monerod — nothing fails at the time, because the process keeps its open
+> descriptors, and the next restart silently comes up pruned. To run it against a live chain,
+> bind-mount the source so the kernel refuses the rename with `EBUSY`:
+>
+> ```bash
+> mkdir -p <build-dir>/lmdb
+> mount --bind <live-data-dir>/lmdb <build-dir>/lmdb
+> tests/integration/compact-chain.sh <build-dir>   # result at <build-dir>/lmdb-pruned
+> ```
+>
+> Prove the guard before relying on it: `mv <build-dir>/lmdb <build-dir>/x` must answer `Device or
+> resource busy`.
+>
+> The generic `mdb_copy -c` does not work: Monero ships a
 > patched LMDB and stock mdb_copy rejects the format (`MDB_VERSION_MISMATCH`). Often it's simplest
 > to leave the free pages.
 
--- a/docs/dev/release-server.md
+++ b/docs/dev/release-server.md
@@ -338,21 +338,11 @@ tests/integration/run.sh --host you@server --dir pithead --readiness
 > [`compact-chain.sh`](../../tests/integration/compact-chain.sh)). It's slow (it copies every block
 > over hours), though it reads through a snapshot so monerod keeps mining.
 >
-> **The tool is copy-then-swap, so it does not leave the swap to you.** After building
-> `<data-dir>/lmdb-pruned` it renames `<data-dir>/lmdb` to `<data-dir>/lmdb-old` and moves the
-> pruned DB into place. Pointed at a live node's data dir it therefore renames the live chain out
-> from under a running monerod — nothing fails at the time, because the process keeps its open
-> descriptors, and the next restart silently comes up pruned. To run it against a live chain,
-> bind-mount the source so the kernel refuses the rename with `EBUSY`:
->
-> ```bash
-> mkdir -p <build-dir>/lmdb
-> mount --bind <live-data-dir>/lmdb <build-dir>/lmdb
-> tests/integration/compact-chain.sh <build-dir>   # result at <build-dir>/lmdb-pruned
-> ```
->
-> Prove the guard before relying on it: `mv <build-dir>/lmdb <build-dir>/x` must answer `Device or
-> resource busy`.
+> **The tool is copy-then-swap, so it does not leave the swap to you.** It renames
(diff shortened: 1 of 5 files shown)


VijitSingh97:
**Non-author adversarial pass: ONE BLOCKING FINDING — hold the merge** — at head `398a3d0ba38c25fb233dd58a801afbf0b659eacd` (re-derived by two routes; full pass in the fleet's `answers/reviewer-20260829T030159Z.md`, controller-judged and the finding re-verified at the head blob via the API before posting).

**Blocking: `tests/integration/testbench-README.md:50-58` is a fourth artefact carrying the same false claim, with a live dangerous recipe attached.** Untouched by this PR, it still reads "no downtime until the swap" (presenting the swap as a manual operator step) and its recipe runs `compact-chain.sh <data-dir>/monero` directly against the live data dir with no bind-mount guard — under the tool's real copy-then-swap behaviour (which this PR itself establishes), following that recipe today renames the live chain out from under a running monerod, exactly the corruption #1489 describes. The subsequent manual `mv lmdb-pruned lmdb` step will then find nothing to move — the tool already consumed it. The fix belongs in this PR: rewrite the README block to match the corrected header's bind-mount recipe and drop the manual-swap steps.

What the pass confirmed clean, so the fix round is delta-scoped: the four swap strings reproduced independently plus a sharper pairing the PR body did not cite (`Blockchain pruned OK` and `Blockchain pruned OK, but renaming failed` as two distinct strings — the signature of a rename attempt on the normal completion path); zero executable lines changed in either script (filtered diff with a positive control); budget gate green both shas with a firing control, tsv byte-identical; shellcheck 0.11.0 + shfmt clean; lint-md + lint-docs-voice green; closing-references empty (issue closes by hand); wallet/topology sweep clean with fired controls; the no-new-test reasoning judged against the testing strategy and it holds. The disclosed swap-fires-on-this-path inference remains honestly open — the live guarded build was still mid-copy at review time — and does not block.

After the README fix lands on this branch: delta re-review of the new commit only, then merge.

VijitSingh97:
## Delta review verdict: MERGE (round 2, controller-judged)

Adversarial delta re-review of the two fix commits (`b9d0549`, `1e12d6c`) — scope was the delta only; the round-one review below stands for everything outside it. Verdict: **MERGE, zero blocking findings.**

What the pass re-derived (not relayed): head sha by two routes; the exact two-file/two-commit delta by direct diff; the new testbench bind-mount recipe traced by hand against the tool's confirmed copy-then-swap behaviour (holds — pruned DB is fully built before the EBUSY-refused rename fires); the hardware.md clause read in context (consistent with the testbench recipe, not contradictory); a tree-wide residue sweep for the retracted safety claim with a fired positive control (clean — the one surviving "never modified" phrase, in `build-pruned-chain.sh:6`, is a true claim about a different pipeline); `lint-md`, `lint-docs-voice`, `lint-file-budget` all green; CI polled to a real terminal state — Shell tests **pass** (9m42s), board all-pass bar the intentional CVE-matrix skips.

One non-blocking finding, now discharged: the delta's two shared-docs touches were undisclosed in the PR body. The Scope section carries the addendum as of this merge.

Merging pinned to the reviewed head `1e12d6c5`. #1489 stays open pending the guarded-build outcome post — the closing keyword was removed from this body deliberately.

Issue #1446, opened 2026-08-28
Remote-node mode still has no live coverage in the routine release gate


VijitSingh97:
## What is left after #1444

#1083 had two halves. PR #1444 closed the first one: every skip the harness emits now carries a
class — `missing` (an input would have run it), `by-design` (this run's mode excludes it), or
`covered` (the behaviour is run elsewhere) — and the summary prints the breakdown, so a reader can
finally tell an accepted hole from an uncovered one.

That makes the hole **legible**. It does not **fill** it. This issue is the second half, filed
separately so that the instrument fix is not held open by the coverage work, and so the coverage
work is not quietly lost when the instrument fix closes.

## The hole

Remote-node mode has **zero live coverage in the routine gate**, and it is the mode the release gate
is least able to speak for.

Two scenarios cover it, and both self-skip on every routine run:

| Scenario | Gated on | Class it now reports |
|---|---|---|
| `remote-main-secure-tari` | `--remote-monero-host` | `missing` |
| `remote-tari-main-secure` | `--remote-tari-host` | `missing` |

Both gates resolve in `resolve_overrides` (`tests/integration/lib.sh:174-175` and `:184-185`), which
is the single writer of `SKIP_REASON`/`SKIP_CLASS`, feeding the one scenario-skip site at
`tests/integration/run.sh:438`.

**Neither host flag is set anywhere.** Re-derived at the current `develop-v2` head by grepping
`tests/ scripts/ .github/ Makefile` for `REMOTE_MONERO_HOST` / `REMOTE_TARI_HOST`: the values are
declared as prerequisites (`lib.sh:128-129`), documented (`scenarios.sh:37-40`), read at the two
gate sites, and exercised in `selftest.sh` fixtures. No caller supplies one. So the scenarios can
only run if a human passes `--remote-monero-host <host>` or `--remote-tari-host <host>` by hand,
and nothing in the mandated pre-release run does.

## Why this is worth an issue rather than a note

The pre-release gate is the check the release mandate leans on. A mode with no live coverage there
is not a thin spot in a test suite — it is a mode we ship on the strength of a green verdict that
never examined it. That is the same shape as the rest of this lane's queue: the gate proves less
than its output implies.

## What would close it

Live coverage of remote Monero mode in the routine gate — the scenario running, and a demonstrated
failure mode proving it can go red. **A scenario that flips from skipped to passed without a shown
red has been silenced, not fixed**, which is worse than the skip, because the skip announced itself.

[...]

PR #1492, opened 2026-08-29
feat(e2e): let remote-mode scenarios target a non-default node port (#1491)

### LINKED ISSUE #1491: Harness cannot point remote-mode scenarios at a non-default port; the product supports it
The harness can only point the remote-mode scenarios at a node listening on the **default** ports. The product supports non-default ones, so the gate cannot exercise its own supported configuration space — and this blocks #1446.

## What the product supports

`config.reference.json` carries `monero.remote.rpc_port` (default 18081) and `monero.remote.zmq_port` (default 18083), and `pithead` validates and renders them separately from the host:

```sh
# lib/pithead/99-remainder.sh:5852-5854
mono_host="$MONERO_REMOTE_HOST"
rpc_port="$MONERO_REMOTE_RPC_PORT"
zmq_port="$MONERO_REMOTE_ZMQ_PORT"
```

## What the harness can express

`resolve_overrides` emits the host and nothing else:

```sh
# tests/integration/lib.sh:172-179
if [ "$mode" = "remote" ]; then
    [ -n "${REMOTE_MONERO_HOST:-}" ] || { SKIP_REASON="needs --remote-monero-host"; return 1; }
    out="$out monero.remote.host=$REMOTE_MONERO_HOST"
fi
```

There is no `--remote-monero-rpc-port` / `--remote-monero-zmq-port` flag and no variable behind one, so `monero.remote.rpc_port` and `monero.remote.zmq_port` are never set and always fall back to their defaults. `tari.remote.grpc_port` has the identical gap behind `--remote-tari-host`.

## Why passing `host:port` does not work

It is the natural thing to try, and `run.sh --help` calls the flag an "endpoint", which invites it. It is silently wrong twice over:

- `is_valid_host` permits `:`, so `<ip>:28081` passes validation and is rendered into `MONERO_REMOTE_HOST` intact.
- Both the product and the harness then append the port themselves. `lib.sh:384` builds `.monero.remote.host + ":" + (.monero.remote.rpc_port // 18081)`, so the health URL comes out as `http://<ip>:28081:18081`.

`selftest.sh:63-67` uses a `<ip>:18081` string as its fixture. That case only asserts pass-through, so it is not wrong as a test, but it does document the `host:port` shape as the expected input — which is the shape that breaks.

## Why this blocks #1446

#1446 needs a remote node reachable from the routine gate. The standalone-node-beside-the-stack topology requires that node on non-default host ports, because the ports it would otherwise want are already bound:

- the live stack's `monerod` runs with `rpc_lan_access: true`, so it binds `0.0.0.0:18081` — an alias IP does not dodge a wildcard bind;
- the matrix's own `local-pruned-main-rpclan` row sets `rpc_lan_access=true` too, so the stack under test takes `0.0.0.0:18081` during the run regardless.

So there is no arrangement in which the remote node can sit on 18081, and the harness currently has no way to say so.

## Scope of the fix

`--remote-monero-rpc-port` and `--remote-monero-zmq-port`, plumbed into `resolve_overrides` and emitting the two overrides only when set, so the defaults are unchanged when they are not. `tari.remote.grpc_port` is the same gap and is deliberately left out of the first change rather than bundled unproven — noting it here so it is not lost.

Note for whoever takes it: `tests/integration/lib.sh` is at 692 lines against a recorded ceiling of 692 and `selftest.sh` at 686/686, so neither can take a net new line. Budget the placement before writing the code.

Found while building the standalone node for #1446.

VijitSingh97:
Fixes #1491.

## The defect

`resolve_overrides` emitted only `monero.remote.host`, so `monero.remote.rpc_port` and `monero.remote.zmq_port` always fell back to their defaults. The gate could therefore only ever point remote mode at a node on 18081/18083 — while `pithead` itself ships, validates and renders both ports (`99-remainder.sh:5852-5854` keeps `mono_host` and `rpc_port` separate). A gate that cannot express its own product's supported configuration.

Passing `host:port` as the host is the natural workaround and it fails silently: `is_valid_host` permits `:`, and both product and harness append the port themselves, so `lib.sh:384` builds `http://<host>:<port>:18081`. `run.sh` now refuses that outright and says why.

## Why it matters beyond tidiness

It blocks #1446. The standalone-node topology needs the remote node on a non-default port, because the live stack's `monerod` runs with `rpc_lan_access: true` and binds `0.0.0.0:18081` — and the matrix's own `local-pruned-main-rpclan` row makes the stack under test take the same wildcard bind during a run. There is no arrangement where the remote node sits on 18081.

## What was RUN

- **22 tier-1 cases** (the count below was corrected 18→22 at merge by the controller; the review pass executed the file: 22 passed, 0 failed, matching CI's own job log), in a new `selftest-remote-endpoint.sh`. `selftest.sh` sits exactly on its recorded budget ceiling and ceilings only go down, so new cases cannot go there; the runner globs `selftest*.sh` (`Makefile:29`), so a new file needs no registration.
- **A 7-mutation battery, all 7 killed.** The battery asserts each pattern matches before writing, so a no-op replace fails loudly instead of reading as "the test survived".

```
M1 drop rpc_port emission              KILLED by: rpc_port override is emitted
M2 drop zmq_port emission              KILLED by: zmq_port override is emitted
M3 emit rpc_port unconditionally       KILLED by: no rpc_port override when unset
M4 emit ports outside the remote block KILLED by: local mode ignores rpc_port / zmq_port
M5 drop the host:port refusal          KILLED by: refusal explains the bare-host rule
M6 drop the port range check           KILLED by: port 0 / 99999 refused, refusal names the range
M7 accept a non-numeric port           KILLED by: non-numeric port is refused, names the range
```

Two of those are worth reading rather than counting:

- **M5 was killed only by the message assertion, never by the exit code.** With the guard gone `run.sh` still exits 2, via "Provide --host/--local" — so an rc-only case would have passed whether or not the guard existed. Assert the reason, not the count.
- **M7 SURVIVED the first battery, and the code was right.** Dropping the `*[!0-9]*` guard changed no outcome because I had no non-numeric case. Without that guard the range check below runs `[ abc -lt 1 ]`, which *errors* rather than returning false, so the `if` body is skipped and the bad value is accepted. The fix was the two missing cases, not a change to the guard.

- **Lints green:** `selftest.sh` itself 214 passed / 0 failed (that file's own count, not the suite total; the full 11-file selftest run is green — clarified at merge), `shellcheck --severity=warning` over `tests/integration/*.sh`, `shfmt -i 4 -d` over the Makefile's exact glob, `make lint-file-budget`, `make lint-topology`, `make lint-md`.

## What was NOT run

- **`make lint-sh` was not run** — it is this box's OOM path and is serialized across lanes. Substituted with the targeted `shellcheck` + full-glob `shfmt` above, which is the standing substitute here.
- **No live run against a real remote node.** That is #1446 and it is blocked on a chain build in progress. This change is what unblocks it; it does not itself prove remote mode passes.
- `tari.remote.grpc_port` has the identical gap behind `--remote-tari-host`. Deliberately left out rather than bundled unproven — recorded in #1491.

## Budget, disclosed

`tests/integration/lib.sh` was at 692 lines against a ceiling of 692, so this change had to pay for itself. It does: the two port overrides go on one `${VAR:+…}` line, the remote-mode comment stays one line (the full reasoning lives in the new selftest's header and in #1491), and the adjacent Tari comment is tightened from two lines to one. The file lands back at exactly 692. **It is now full** — the `tari.remote.grpc_port` sibling above will need a split before it can land.

`shfmt` reverts the compact `{ …; return 1; }` brace form, so that is not available as headroom; both guards keep the expanded form.

## Ponytail pass

The plugin's skill is not registered in this session, so I read its definition and ran the pass myself:

```
run.sh:L256-260: shrink: 5-line if/else assigns one of two vars from a shared case arm.
  case "$1" in *rpc-port) A="$2";; *) B="$2";; esac — 4 lines.
run.sh:L248,253: delete: the "1-65535" refusal text appears twice.
net: -2 lines possible.
```

**Noted, not applied**, which the gate allows. The two refusal texts are reached by two genuinely different rejections (non-numeric vs out-of-range) and the mutation battery kills them independently — collapsing them would couple two cases that currently fail for distinct reasons. The `if/else` saves one line at the cost of matching flag names by glob, and both alternatives are net-negative for `run.sh`, which has 31 lines of headroom and is not the constrained file here.

https://claude.ai/code/session_01A8qHpoXBs7pMX7Tq19GEiq

### DIFF
--- a/tests/integration/lib.sh
+++ b/tests/integration/lib.sh
@@ -126,7 +126,7 @@ render_scenario_config() {
 # the final override string and returns 0; on a missing prerequisite sets SKIP_REASON and
 # returns 1 — no silent drops, and never a prune flip on the canonical synced DB (which would
 # invalidate it). Reads the globals BASELINE_PRUNE / PRUNED_DATA_DIR / FULL_DATA_DIR /
-# REMOTE_MONERO_HOST / REMOTE_TARI_HOST / IT_MONERO_VIEW_KEY / IT_TARI_VIEW_KEY /
+# REMOTE_MONERO_HOST (+_RPC_PORT/_ZMQ_PORT) / REMOTE_TARI_HOST / IT_MONERO_VIEW_KEY / IT_TARI_VIEW_KEY /
 # IT_TARI_SPEND_PUBLIC_KEY (all optional). Pure given those globals, so the self-test exercises it.
 RESOLVED=""
 SKIP_REASON=""
@@ -169,17 +169,17 @@ resolve_overrides() {
         out="$out monero.data_dir=$FULL_DATA_DIR"
     fi
 
-    # Remote mode needs an external endpoint to point at.
+    # Remote mode needs an external endpoint; the host must be BARE, never host:port (#1491).
     if [ "$mode" = "remote" ]; then
         [ -n "${REMOTE_MONERO_HOST:-}" ] || {
             SKIP_REASON="needs --remote-monero-host"
             return 1
         }
         out="$out monero.remote.host=$REMOTE_MONERO_HOST"
+        out="$out${REMOTE_MONERO_RPC_PORT:+ monero.remote.rpc_port=$REMOTE_MONERO_RPC_PORT}${REMOTE_MONERO_ZMQ_PORT:+ monero.remote.zmq_port=$REMOTE_MONERO_ZMQ_PORT}"
     fi
 
-    # tari.mode remote (#103/#942) needs its own external endpoint — same shape as monero's above,
-    # a separate global since the two chains can point at different remote hosts.
+    # tari.mode remote (#103/#942) needs its own endpoint — same shape, its own global.
     if [ "$tari_mode" = "remote" ]; then
         [ -n "${REMOTE_TARI_HOST:-}" ] || {
             SKIP_REASON="needs --remote-tari-host"
--- a/tests/integration/run.sh
+++ b/tests/integration/run.sh
@@ -57,6 +57,8 @@ KEEP_STATE=0
 EXPECTED_WORKERS=2
 SKIP_MINING_ASSERTS=0
 REMOTE_MONERO_HOST=""
+REMOTE_MONERO_RPC_PORT=""
+REMOTE_MONERO_ZMQ_PORT=""
 REMOTE_TARI_HOST=""
 PRUNED_DATA_DIR=""
 FULL_DATA_DIR=""
@@ -95,8 +97,10 @@ MATRIX:
   --no-mining-asserts    SKIP the two mining assertions (workers online, stratum hashes) with a
                          logged notice — for a box with no miner connected (e2e --no-miner, #905).
                          Every other assertion stays binding.
-  --remote-monero-host <h>  external node endpoint for the remote-mode scenario
-                            (e.g. the box's own synced node on its LAN IP)
+  --remote-monero-host <h>  external node for the remote-mode scenario — a BARE host or IP, never
+                            host:port (pithead appends the port itself, #1491)
+  --remote-monero-rpc-port <p>  that node's RPC port, when it is not the default 18081
+  --remote-monero-zmq-port <p>  that node's ZMQ port, when it is not the default 18083
   --remote-tari-host <h>  external Tari node endpoint for the tari.mode=remote scenario (#103;
                          e.g. an already-synced Tari node on the LAN)
   --pruned-data-dir <d>  synced PRUNED monero data dir (enables the pruned case when the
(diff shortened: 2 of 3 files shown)


VijitSingh97:
## Review verdict: MERGE (controller-judged)

Full adversarial pass at head `f4bd194` (base `492a41c`, stable throughout the pass by before/after `ls-remote` probes). **Zero blocking findings.**

What the pass re-derived by execution, not reading: the double-append defect reproduced at base end-to-end via `run.sh` (`http://1.2.3.4:19081:18081`) and the head's parse-time refusal in its place; the default-port path byte-identical base vs head (`resolve_overrides` output compared directly); 3 of the 7 mutations re-run with diffs proving each applied, named kills captured, files restored diff-clean — including M5, where the reviewer independently confirmed the author's rc-is-useless point by direct call; the budget gate run live both sides with a firing positive control (695-line append correctly refused, named); the new file glob-picked-up by runner and lint (via `make -n`, not read); two leak sweeps, each with a fired positive control; CI polled to a real terminal state and the raw job log pulled — `selftest-remote-endpoint: 22 passed, 0 failed`, matching the local run to the assertion.

Rulings at merge:
- **The `lib.sh` 692/692 comment paydown is ACCEPTED** — in the exact function this commit extends, disclosed in the body, and the dropped "why two globals" rationale is re-derivable from the two lines below it. Should-fix on record: restore the clause if headroom ever opens on this file.
- The PR body's "18 tier-1 cases" was corrected to the executed 22, and the `selftest.sh` 214-count clarified, in the body directly (marked as merge-time edits).
- Four mutations (M2/M3/M4/M6) stand on the author's report, unverified — stated, not assumed.
- No live run against a real remote monerod exists on either side; that is #1446's job, which this PR unblocks.

Merging pinned to the reviewed head `f4bd194`. #1491 gets closed by hand (no closing keyword fires here — `closingIssuesReferences` is empty, confirmed via the GraphQL route).

PR #1485, opened 2026-08-29
fix(dashboard): a failed history read is no longer rendered as another dashboard's change (#1409)

VijitSingh97:
## What this fixes

`get_worker_config_history` returned `[]` on both of its failure paths, which is the same value it
returns when a rig genuinely has no recorded changes. `build_worker_detail` reads that list to decide
where the rig's running config came from, so a read that never happened was indistinguishable from a
read that came back empty — and an empty result means "we hold no row for the rig's `last_change_id`",
which is the `elsewhere` verdict, which renders as **"Last changed from another dashboard"**.

That is an accusation, and on this path its real source was our own database failing to open.

## The reproduction

Driven through `build_worker_detail` against a real `StateManager` on a temp DB, not against
`config_origin` directly — nothing is mocked on the method under test.

| leg | how | history | verdict before | rendered before |
|---|---|---|---|---|
| **A — control** | healthy DB, change id present | 1 row | `here` | `text-muted` · Last changed from this dashboard |
| **B** | real `sqlite3.Error` — `DROP TABLE worker_config` | `[]` | `elsewhere` | **`status-warn` · Last changed from another dashboard** |
| **C** | no connection (`storage_service.py:888-889`) | `[]` | `elsewhere` | identical to B |

**Leg A is the load-bearing row.** Same fixture, same id, correct verdict when the read works — so B
and C are attributable to the failed read rather than to a fixture that could never have reached
`here` in the first place. Leg B's handler is confirmed by its own log line (`no such table:
worker_config`), not inferred. **Leg C raises nothing and logs nothing**, so the two failure routes
are not equally observable; that asymmetry is most of why this was hard to see, and it is why the
tests cover them separately rather than through one representative case.

## The change

- `get_worker_config_history` returns **`None` on both failure paths** (`:888-889` and the
  `except sqlite3.Error` at `:906-908`). `[]` keeps meaning **genuinely no history**.
- `config_origin` gains a `history_unread=False` knob and a new **`unread`** verdict, returned as the
  first check inside the `source == "control"` branch. When the read worked, every verdict is exactly
  what it was before, so a caller that does not pass the flag loses nothing.
- `unread` renders **muted, not warn**. Muted is an EXISTING treatment here — `here`, `untraced` and
  `unrecorded` already use it, and the client's own comment says the muted verdicts are the ones that
  claim nothing. So this adds a verdict **state**, deliberately not a new **colour**: the text carries
  the distinction, the colour carries the class. Please do not later give it its own hue.
[...]

Issue #796, opened 2026-07-31
The appliance offers mine-on-this-box but gives no way to install RigForge on it


VijitSingh97:
The wizard's "Mine on this machine too?" choice (local_miner.enabled) tells the operator to
install RigForge on the host afterwards. On DIY that is a real instruction. On the APPLIANCE it
is not: the machine ships with no shell in release images, and the supported management surface
is the dashboard. An appliance operator who answers Yes gets a stratum announcement pointing at
a rig that can never exist — the promise cannot be kept as shipped.

RigForge is deliberately NOT containerized (it owns host tuning: MSR, HugePages, governor —
the #78 spike recorded hugepages as load-bearing), so "just run an xmrig container" is a
different, weaker product than what Yes implies.

Options, in rough order of honesty-per-effort:
1. On the installation medium, keep the question but state the real requirement inline: mining
   on the appliance itself needs ssh.enabled (Advanced) + a manual RigForge install — or hide
   the choice on the appliance entirely and document "the appliance coordinates; rigs mine."
2. A pithead-managed miner container as an explicit lesser mode ("mine without host tuning —
   lower hashrate"), so the appliance can honor Yes without a shell.
3. Full RigForge-on-appliance integration (host-tuned, updated with the OS image) — the real
   answer if appliance-local mining matters as a product feature.

Today the fleet story is the good one and should be the marketed path: install RigForge on
each RIG, point it at the stratum address from the credentials card, manage from the
dashboard's Workers view (list, hashrate, one-click worker upgrades since v1.10, telemetry).

VijitSingh97:
Option 3 is implemented on the current appliance image: PR #812 (phase R2 of #797). The image bakes rigforge's toolchain, a pinned rigforge tree and a prebuilt XMRig; `pithead-sync` delivers them to /data; `render` derives the miner's config (loopback stratum + password + the stack's HugePages headroom via `hugepages_reserve_extra_mb`); pithead-boot runs rigforge's appliance-mode setup after the health-gated slot commit, every boot. The wizard's Yes answer now promises the built-in worker — the install-RigForge-by-hand instruction this issue was filed about is gone from the wizard, the CLI announcement and the docs. Bench verification list is in the PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

VijitSingh97:
[...]

PR #1493, opened 2026-08-29
refactor(cli): cut the announcements, miner provisioning, OS update and resets out of the remainder (#1105)

VijitSingh97:
#1105 Phase 2 **P4**. Four order-preserving contiguous slices come out of
`lib/pithead/99-remainder.sh`. The shipped `pithead` artifact is **byte for byte what it was** —
this PR moves lines between files and changes nothing the CLI does.

| slice | lines | what it is |
|---|---|---|
| `13-announce.sh` | 65 | the post-command announcements — dashboard URL, stratum auth, the DIY local-miner hand-off |
| `14-local-miner.sh` | 216 | the appliance's built-in RigForge miner **and** the rig role's boot leg |
| `15-os-update.sh` | 260 | OS update across the A/B slots: variant stamps, the version guard, the `/data` migration floor |
| `16-reset.sh` | 156 | the three reset commands — `reset-dashboard`, `config-reset`, `factory-reset` |

`99-remainder.sh` **9218 → 8517**. Every slice is under the 400-line floor (verified at source:
`scripts/lint-file-budget.sh:25` `TARGET_LINES=400`, and the gate's rule is an entry *iff* over
target), so this cut adds **no new budget row**; the remainder's row drops, which is what each
Phase 2 cut must make it do.

**Conservation, machine-checked:** 697 slice lines + 4 build-supplied blank separators = 701 =
9218 − 8517.

## What was RUN

Everything below was run by me **at this PR's rebased head**, in `~/dev/appliance-p4`. This branch
was originally cut on P3's head; after #1488 merged I rebased and **re-ran every proof rather than
inheriting the pre-rebase results**. The numbers here are the post-rebase ones.

- **Parity** — `scripts/build-pithead.sh --check` **rc 0**, "exactly what 18 slice(s) build", and
  `pithead` is absent from `git status`, so the artifact is untouched rather than rebuilt to match.
- **The parity gate PROVEN TO FIRE.** Mutating `16-reset.sh:1` gave `git diff --numstat` **1/1**
  with the mutant line shown *before* the gate ran — an untracked or unstaged file makes `numstat`
  print nothing, which would make a blind control read exactly like a clean one. `--check` then
  returned **rc 1**, naming `3611c3611` and quoting both sides. Restored: numstat empty, rc 0.
  Run again pre-rebase on a different slice (`15-os-update.sh:1` → `3350c3350`); both line numbers
  independently confirm the offset mapping remainder-line *N* → `pithead` *N*+3066, which is the
  point of the control — two routes to the number, not one derivation.
- **Every range verified by EXTRACTION, not arithmetic.** For each slice,
  `awk 'NR>=a && NR<=b' <the base remainder> | cmp -` against the slice file: **all five exact**
  (the four slices and the new remainder).
- **`bash -n` on each of the 18 slices individually AND on the built artifact** — all clean. The
  per-slice run is the instrument that matters: a boundary falling inside a heredoc still
  reproduces the artifact byte for byte, because concatenation puts it back, so byte-identity is
  structurally blind to it. Only each slice parsing alone rules it out.
- **`shfmt -i 4 -d`** over `lib/pithead/` — clean, zero bytes of diff.
- **File budget** — `scripts/lint-file-budget.sh` **rc 0**, both halves: "every over-target file
  has a ceiling, none grew past it, `docs/dev/file-budget.tsv` is monotonic."

## Suite after-proof

**rc 0, `pithead tests: 2979 passed, 0 failed`**, 3292 output lines, 274 section headers — run in
a worktree checked out to **this PR's rebased head itself**, not a reconstruction, so there is no
equivalence to prove. The verdict is read from the **rc file**; a detached launcher's own rc is
always 0, and a run is finished when its rc file exists, not when the log stops growing. Failure
glyphs counted directly: `grep -c '✗'` = **0**. (A case-insensitive `FAIL` grep is useless on this
log — the suite names tests after the condition they assert, so it returns hundreds of hits with
zero failures.) Log: `~/appliance-suite-logs/p4-after-20260829T031216Z.log`.

**The comparator, stated precisely.** P3's after-proof one merge earlier reported the same
`2979 passed, 0 failed` over 3292 lines. **I did not run a fresh baseline at `492a41c` itself** —
what changed between the two runs is #1485, and it touches only `dashboard/` and
`docs/dashboard.md`: `git diff --name-only d9f6337..492a41c -- tests/stack/ pithead lib/` is
**empty**. So the identical totals are the expected result rather than a coincidence, and that is
the whole of the claim — this is an after-proof at the head, not a controlled before/after pair.

**I also killed a first suite rather than reuse it.** It was mid-run against the pre-rebase tree
when #1488 and then #1485 merged. Reusing it would have rested on "`tests/stack/` is byte-
unchanged", which is the wrong test and looks like the right one — this suite reads outside
`tests/stack/`. It was killed by matching `/proc/PID/cwd` against my own worktree, never by
pattern, because other lanes were running their own suites concurrently.

## Budget proof — stated honestly, base named

The monotonic half of the budget gate does not run in CI (#1452), so this proof is **local**.
Taken against `origin/develop-v2` at **`492a41c`**, and that ref was probed **before and after**
the gate run and was unmoved both times. All worktrees on this box share one `.git`, so any
lane's fetch can move a gate's base mid-run; a sha-pinned worktree pins content, not refs, so it
is not a control for this.

**Every tsv row difference accounted for:** the diff against the base is **exactly one row** —
`lib/pithead/99-remainder.sh` 9218 → 8517. No foreign row moved, no row added, none dropped.

**Byte-identity is structurally blind to slice ASSIGNMENT.** A cut at any single blank line
between two top-level blocks reproduces the artifact exactly, including a semantically wrong one.
Only reading answers this, so here are the three places to push:

1. **`16-reset.sh` contains `reset_dashboard`, which sits ABOVE the source's own
   `# --- Two-tier reset (config / factory)` header.** That header does not bound the file's
   contents. The claim: all three are destructive reset commands dispatched from the CLI's command
   table (`pithead:12227`, `:12229`, `:12230`), so "the reset commands" is the honest name and
   `reset_dashboard` is a third, narrower tier. Attack it — the alternative reading is that
   `reset_dashboard` is dashboard machinery with no business in a reset file.
2. **`14-local-miner.sh` spans TWO source section headers** (`# --- local miner on the appliance
   (#796)` and `# --- the rig role's boot leg`). The claim that they are one domain rests on a
   call I re-derived twice: **`provision_local_miner` dispatches to `provision_rig_miner` by
   machine role** (`pithead:3255`, inside `provision_local_miner` at remainder 186-217). If that
   dispatch is not what it looks like, this file is two domains wearing one name.
3. **`announce_local_miner` is in `13-announce.sh`, not beside the local-miner machinery it names.**
   It cannot move: it sits between `announce_dashboard_url` and `announce_stratum_auth`, and
   order-preserving contiguity is what makes this whole proof class work. Disclosed rather than
   argued away.

## Over-engineering pass, run BY HAND — and why that was necessary

The ponytail PR gate reads `git rev-parse --abbrev-ref HEAD` in this session's **pinned shell
cwd**, not the worktree the command ran in, so on this lane it names a branch that no longer
exists and gates a diff it never saw. It denies the first `gh pr create`, touches a sentinel, and
lets the retry through without ever checking that a review happened. So the pass is run here by
hand, against the right diff, and the mismatch is disclosed rather than allowed to mean the diff
went unreviewed.

**There is no new logic to over-engineer, and that is measured, not asserted.** Splitting the diff
into its added and removed multisets: **701 removed, 697 added, and zero added lines are absent
from the removed set.** The only four removed-not-added lines are **blank** — the separators the
build supplies. Outside `lib/pithead/` the entire diff is **1 tsv line**. (Computed with
`awk '/^--- /{next} /^-/{print substr($0,2)}'`, never `grep '^-[^-]'`, which silently drops
removed blank lines and makes a pure move read as a content change.)

**Is four files more than the region needs?** Order-preserving slicing means only ADJACENT merges
are available, and each costs one build separator:

| merge | lines | verdict |
|---|---|---|
| 13+14 | 65+216+1 = **282** | **legal** — under 400, no new row |
| 14+15 | 216+260+1 = **477** | over 400 → would force a new budget row |
| 15+16 | 260+156+1 = **417** | over 400 → would force a new budget row |

So exactly **one** legal merge exists, and it is **declined on naming, with evidence**: the three
`announce_*` functions are the CLI's post-command output, called from outside the miner domain
entirely — `announce_dashboard_url` from `pithead:467` and `:9854`, `announce_stratum_auth` from
`:1135` and `:7541`. `14-local-miner.sh` is provisioning machinery that renders RigForge's config
and runs its setup. The source itself draws the line, at the head of the local-miner block: *"On
DIY, none of this fires — the operator installs RigForge by hand (`announce_local_miner`'s
values)."* Merging them would name announcement code after a miner.

Three files instead of four was therefore available and was **rejected for that reason, not
overlooked**. The ratified split-degree rule (largest indivisible function, not region line count)
sets a *minimum* of two files here; everything above that is a naming call, and this is it.

## What was NOT done

- **No `make lint` / `make lint-sh`, and the lint lock was never taken.** Argued rather than
  skipped: `make -n lint-sh` shows exactly one `shfmt -i 4 -d` invocation reaching `lib/pithead`
  and neither `shellcheck` invocation listing it, so a slice cut's entire new exposure under that
  target is the shfmt pass — which was run clean above. CI's Shell tests job is the arbiter.
- **No bench run.** Nothing in this cut changes behaviour.
- **No docs change.** `docs/dev/repo-map.md` already states concatenation-not-source as the
  contract for `lib/pithead/`, and the slices deliberately carry **no headers** — adding one would
  break byte-identity. Nothing in the docs goes stale from this cut.
- **No non-author review.** Everything above is author-verified only.


VijitSingh97:
## Review verdict: MERGE (P4 gate pass, controller-judged)

Full adversarial pass at head `6a628d2` (base derived independently; the mid-review develop-v2 advance to `d9d98bd` was disclosed and touches nothing this diff depends on). **Zero findings.**

The load-bearing claims, all re-derived at the git-object level rather than the checkout: built `pithead` blob sha identical base vs head (stronger than cmp); conservation extracted from `base:pithead` by blob-offset — all five ranges cmp-exact, 697+4=701 reproduced by an independent route; pure-move confirmed by `sort`+`comm` as a second derivation (4 removed-not-added lines, all blank; 0 added-not-removed); parity and budget gates run live with firing controls proven (a second mutation shape for parity; a lowered ceiling for budget). The three disclosed soft spots all held at source: `16-reset.sh`'s header sits below `reset_dashboard` and the CLI dispatch table shows the three resets as adjacent cases; `14-local-miner.sh`'s two-header span is justified by the real `machine_role` dispatch at `pithead:3255`; `announce_local_miner`'s position-lock is correctly disclosed as unmovable. The declined 13+14 merge is well-founded on call-site evidence (`up`/`apply`/`status`/`rotate-secrets`, not miner provisioning), and the file-count table recomputes exactly — only that one adjacent merge was legal. Leak sweeps clean with all three positive controls fired. The `lint-sh` hole for `lib/pithead/` is confirmed real, pre-existing, and shared by all 18 slices — disclosed, not introduced here.

The review's one contingency — CI's Shell tests, pending at report time — has since resolved **pass**; the full board is green bar the intentional CVE-matrix skips.

Merging pinned to the reviewed head `6a628d2`. Comparator note for the record: the reuse of P3's after-proof as the "before" reference was verified — exactly one merge (#1485) sits between, touching only dashboard files.

Issue #1104, opened 2026-08-18
The tier-1 suite cannot be run twice at once: the tor entrypoint test drives a fixed /tmp/torrc


VijitSingh97:
## What happens

`tests/stack/run.sh`'s tor-entrypoint block runs the real `build/tor/entrypoint.sh` with a stub `tor` that prints the rendered file:

```
printf '#!/bin/sh\ncat /tmp/torrc\n' >"$d/tor"
```

The entrypoint writes its output to a hardcoded `/tmp/torrc` (`build/tor/entrypoint.sh:13`, appended at :22/:32/:48, handed to `tor -f` at :57). Inside the container that path is private and correct. On the host it is global, so **two concurrent runs of the suite render into the same file and read each other's output**.

Everything else in the suite is isolated — `SANDBOX` is a `mktemp -d`, and every other fixture follows. This is the one shared path.

## How it shows up

As a false RED, in whichever run loses the race, on assertions that read a torrc the other run wrote:

```
✗ tor entrypoint: no Tari HS when local_tari is omitted (remote tari, #103)
✗ tor entrypoint: no Monero HS with no profiles (remote monero, #103)
✗ tor entrypoint: dashboard HiddenService appended when enabled (#343)
```

The set varies run to run, which is the tell. A serial run of the same tree is clean.

Two suites at once is not a strange thing to do — it is the obvious way to run a baseline and a mutant of the same change side by side, which is the mutation discipline this repo has adopted for every new assertion (#1069). It is also what a CI matrix would do on one runner.

## Why it is worth fixing rather than documenting

A false RED costs more than a slow suite: it sends the reader looking for a regression that is not there, and the natural next move — re-run until green — is exactly the habit the repo has been removing. The failure also lands in the *neighbourhood* of real work (Tor, hidden services), so it reads as plausible.

## Fix

Give the entrypoint a `TORRC_OUT` seam with `/tmp/torrc` as the default, matching the `TORRC_TEMPLATE` seam the same test already uses, and point the test at a path under its own sandbox. One line each side; container behaviour unchanged.

Then the mutation runs the coverage work depends on can be done in parallel without producing noise that has to be explained away.

Issue #719, opened 2026-07-21
Config editability: confirm-gate the operationally-disruptive fields (keep the perimeter host-only)


VijitSingh97:
## Summary

Let operators edit **operationally-disruptive** config from the dashboard behind a type-to-confirm gate, instead of the current hard-refuse. Keep the security perimeter host-only. Decided scope after a full editability audit (2026-07-21).

## The principle (from the audit)

Type-to-confirm is **UX friction, not a security control** — a compromised/XSS'd dashboard that can change a field can also fill the confirmation box. So it's the right tool ONLY where the consequence is *operational* (expensive re-sync, rigs must repoint), not a *breach*, and the typed word ("APPLY") is orthogonal to the value being set.

## In scope — unlock with type-to-confirm (a warning + type-to-confirm, but let them proceed)

These are DEST today (hard-refused) but are operator-intent operations, not secrets:
- **Monero/Tari data-dir moves** (`*_DATA_DIR`) — triggers a full re-sync; confirm text should spell that out.
- **Stratum port** (`p2pool.stratum_port`) — every rig must repoint.
- **Clearnet initial-sync toggles** (`monero.clearnet_initial_sync`, `tari.clearnet_initial_sync`) — host IP exposed during IBD, auto-reverts to Tor; rare, worth friction.
- **Monero prune ENABLE** (`monero.prune` off→on) — deletes blocks to reclaim disk. (Disable → full re-sync stays refused, or gets its own heavier warning.)

## OUT of scope — stays host-only (perimeter; confirm-gating adds no real protection)

Wallet addresses, view keys, node RPC creds, dashboard login (`auth`), the onion (`onion.*`), **Tor egress firewall** (`network.tor_egress_firewall`), the control channel itself (`dashboard.control.enabled`), stratum password, and per-rig hosts/tokens (`workers.list[]`). A compromised dashboard confirming these = redirected payouts / disabled Tor / lockout. These are THE boundary — leave them at hard-refuse.

## Implementation notes (from the audit — file:line)

- Gate: `pithead` `control_approval_gate` ~L5001-5089. Today: DEST porcelain row (`^DEST\t`) → refuse (L5070-5072); non-allowlist ENV key → refuse (L5058-5068). Change: for the in-scope DEST keys, return a "confirm-required" status carrying the warning instead of a flat refusal; keep the perimeter refusals. Keep `describe_change` (L4321+) as the source of the DEST classification + human message.
[...]

PR #1494, opened 2026-08-29
test(stack): split the secrets-masking domain into its own file (#1105 Phase 1, develop-v2)

VijitSingh97:
Phase 1 of #1105. Moves the four control-channel secrets-masking sections out of `tests/stack/run.sh` into `tests/stack/test-secrets-masking.sh`:

- the pre-masked prefill copy and host-side secret merge (#440)
- the per-field `.env` line-injection guard (#33 hardening)
- the client-auth requirement on a published onion (#33 hardening)
- `telegram.control` failing closed on each of its legs (#521)

Moved code is byte-identical. The `source` stanza sits at the block's own vacated position, so execution order is unchanged.

## This file deliberately does NOT self-arm, and that is the substantive decision

Most domain files under `tests/stack/` call `build_control_sandbox()` at their own entry so they are standalone-sourceable. This one does not. It follows the shipped `test-control-add-only-ssrf.sh` **disclosure precedent** instead — sourced in place, position-locked, dependency stated in its header. "It should self-arm like its neighbours" is the obvious review note, and it is wrong for this domain. The derivation, all re-derived at the base:

- **The domain is a pure consumer.** The moved text never calls `build_control_sandbox`. `run.sh` calls it once, in a control-core section that stays behind and runs before the stanza; `$C` and `$CTRL_LOG` reach the domain from that call.
- **A self-arm would build a SECOND sandbox, and that is the defect, not the fix.** The domain drives the control channel repeatedly through `pithead apply -y` and `run_pending`. Sections that remain in `run.sh` **after** this stanza count that shared state against exact numbers — the audit log by line count, the request spool by file count. A fresh `$C` would send those writes to a different results dir than the counters read.
- **The three downstream counters, located at HEAD.** All four names (`$REQS`, `$RESULTS`, `$STAGED`, `$AUDIT`) are assigned in `run.sh` at `:345-348`, as plain `$C` derivations, before the stanza at `:462`. Of the assertions that count them: `wc -l <"$AUDIT"` at `:918`, and `ls "$REQS" | wc -l` at `:966` and `:969` — expecting exactly `10` then `0` — all sit **after** the stanza and read state this domain writes. `ls "$RESULTS" | wc -l` at `:371` expects exactly `1` and sits **before** it, so it is safe on ordering. Three exact-value assertions downstream, not one, and none of them is a "greater than".
- **That is coupling rule A, and it has a recorded firing.** The rig-worker token-mask cluster moved with a re-derived `$C`; its applies wrote extra result files into the shared results dir and a still-in-`run.sh` assertion counting that dir went red (2901 passed / 1 failed).
- **The block's tail is load-bearing for what follows.** Its last lines restore the section baseline — control on, no telegram — and re-apply; the in-code comment there says it is "for the tests that follow". The anchor is therefore a correctness requirement in this cut, not the usual preference.

## The over-engineering pass found a defect in the header, and it is disclosed here

The pass has one target on a move PR: the new file's header, the only authored content in the diff. It found one, and the correction is in this branch.

An earlier draft filed `$WALLET` as a **"lib.sh top-level constant"**. That is false. `lib.sh` assigns `WALLET` only *inside* the two sandbox builders, as `WALLET="${WALLET:-$VALID_PRIMARY}"`, and `run.sh` never assigns it at all — so `$WALLET` reaches this domain from the same `build_control_sandbox` call that provides `$C`, by the same ordering dependency, and belongs in the disclosure rather than in the constants line. The domain reads `$WALLET` six times, all non-comment. `$VALID_TARI` *is* genuinely top-level and stays classified as such.

The mechanism is worth naming because it will recur: the audit tool reports `WALLET -> PROVIDER`, which means "assigned somewhere in the provider file" and says nothing about scope. A defaulting fix retires a coupling only for CALLERS, and a split manufactures non-callers. No instrument in this recipe can see a false sentence in a comment.

**The header was corrected BEFORE the after-proof was launched**, so there is no proven-tree/committed-tree gap on this PR — the winning after-log was produced from this exact tree.

## Proof

- **Byte-identity, both directions, empty**, re-derived from `git show <base>:tests/stack/run.sh` rather than from the extraction. Moved text 166 lines / 159 non-blank; `run.sh` 1305 non-blank. Conservation closes: base 1462 = 1305 + 159 − 2. A positive control (one-line mutant) was confirmed to make the diff fire. The `run.sh` rebuild counts the inserted stanza block as three lines (comment, stanza, blank); a two-line reconstruction reports a single phantom `462a463`, which is an artifact of the reconstruction.
- **The seam checks out in both directions, and the arithmetic closes independently.** The cut starts at the section's own `echo "== "` header and ends on the blank line after the block's restore tail; the next section's header (`confirm-gate`, #719) still stands in `run.sh`. Neither side carries a `# ---` divider, so none travelled and none was orphaned. 167 lines deleted, a 3-line stanza block inserted, net −164 — and the budget row moved 1570 → 1406, a drop of exactly 164.
- **Nested-stanza check clean**, with a firing positive control: zero `source "$HERE/` stanzas in the cut range, while a known-positive range returns one. The nearest stanza is well outside the range.
- **Ordered section-header comparison** (the only instrument that can see a permutation — a sorted multiset, a header count and a pass total are all blind to one by construction): **zero reorder, 0 of 274 positions differ**, and the two logs' header sequences diff empty. The control fired: swapping two adjacent headers makes the ordered diff report the change, while the sorted-multiset diff over the same permuted input stays silent — which is why this instrument is run separately rather than inferred from the set comparison.
- **Suite — read this one carefully, the verdict is red on BOTH sides and that is not this cut.** Baseline rc 1, **2977 passed / 3 failed**; after rc 1, **2977 passed / 3 failed**. Both runs are 274 section headers and 3297 log lines. The three failures are **byte-identical on both sides** — same three tests, same messages — and they are in `== unit: the mutation lock serialises mutating windows (#1342) ==`, a section in `tests/stack/test-lifecycle.sh` that this diff does not touch and that runs identically either side of the stanza. The pristine baseline worktree reddens the same way, so the redness is in the environment, not in the cut.

  **Diagnosed, each step re-derived rather than assumed.** The test backgrounds a subshell that opens a descriptor on the lock file, takes `flock -n`, and `exec sleep 60`; it then polls `lock_state` at most 200 times at 0.05s — a hard 10-second wall-clock budget for that background process to be scheduled and acquire the lock. These runs were launched at **N=3 concurrency** alongside another lane's suite, at load ~7 on 12 cores, and the poll exhausted. The test is well built and says so itself rather than passing vacuously: its first failure message is "the lock is free, so the three cases below prove nothing". Ruled out: not a code regression (the lock lives in `lib/pithead/00-prelude.sh` and the root `pithead` script, neither touched by anything that landed under this branch); not cross-suite lock contention (`LKFILE` is `"$SANDBOX/lock/explicit.lock"` and `$SANDBOX` is `mktemp -d`, so the lock file is private per run — this is not a second host-global path in the #1104 class); not this cut.

  **What that leaves the equivalence proof, which is what this PR actually needs.** Identical header set, **identical header ORDER, 0 of 274 positions differ**, and a whole-output sorted-multiset diff — ANSI stripped, worktree and `mktemp` paths normalised with the same expression on both sides — of **exactly two lines**, the two halves of a single ed25519 keygen fingerprint (#1325). The `CONFIG_FILE: readonly variable` warnings are 3 on each side, emitted by `test-appliance-identity.sh` at its own lines 73 and 105; this cut does not touch that file, so the predicted shift is zero by construction and the measured shift is zero. Two runs agreeing on three failures in an untouched test is a more specific coincidence than two greens, so the equivalence measurement is if anything stronger here — but it is an equivalence proof, not a green suite, and those are different claims.

  **Not claimed: that the tip is green when run serially, and the honest test could go against me.** The load-sensitivity is filed as #1495 with its mechanism and controls. A local serial control was started, but peer lanes joined it with their own suites, so it never ran at N=1 and is not the clean control it was meant to be; its result will be posted as a comment for what it is worth. **CI is the real control here — it runs this suite serially at N=1.** So the reasoning cuts both ways and should be stated in the direction that can falsify me: if CI's `Shell tests` goes green, the load explanation holds; **if CI reddens in the `#1342` lifecycle block, that is evidence AGAINST the load hypothesis**, and the right response is to reopen #1495 on a different mechanism rather than to go hunting in this diff. Either way the answer belongs to #1495, not to this cut, which does not touch `test-lifecycle.sh`.
- **Budget**: `tests/stack/run.sh` 1570 → 1406 is the only row changed. The new file is 219 lines, under the 400-line floor, so it correctly gets no row. `--generate` also wanted to lower three rows this diff never touches (`dashboard/mining_dashboard/service/egress.py`, `.github/workflows/ci.yml`, `tests/integration/run.sh`); all three were hand-restored stale-HIGH, which the per-file check permits and which avoids reddening un-rebased branches in other lanes. The gate passes, the base ref was probed unmoved before and after, and a firing control (delete the `run.sh` row) makes it fail by name.
- **Lint**: `make lint` clean. `shfmt -i 4 -d` and `shellcheck --severity=warning` clean on the new file and on `run.sh`.

## Disclosed

- **Three block-local functions are not unset at the end of the block** — `inject_reject`, `onion_control_config`, `tg_control_config`. Unlike the previous cut's `irun`, they outlive the source exactly as they outlived their old position in `run.sh`. No other file under `tests/stack/` uses those names (checked with a positive control), so nothing downstream can see a definition it did not see before.
- **`$REQS`, `$RESULTS`, `$STAGED` and `$AUDIT` are not the builder's.** They are assigned by the control-run-pending section, which stays in `run.sh` and runs before this stanza. They are deliberately not seeded here: each is a plain derivation from `$C`, so a seed could drift from `run.sh`'s definitions and would buy nothing, because `$C` itself keeps the file non-standalone either way.
- The header says `run.sh` calls the builder "once". That is a count in a header, which this lane's own rule discourages. It was kept deliberately: it is a count of calls to a named function, which changes only when someone adds a second sandbox build — the exact event the header exists to warn about — rather than drifting with routine edits. Flag it if you disagree.

## Not done

- No behavioural change is claimed or attempted; this is a move plus an authored header.
- **The proofs were taken twice.** `develop-v2` moved twice while the first pair was running — a Phase 2 CLI cut rewriting `lib/pithead/*`, then a dashboard fix. The first pair was completed, clean and then discarded, and the branch was rebased and re-proven at the current tip, because a proof of a tree that is not the tip proves the wrong thing. Every static instrument was re-derived against the new base and none of them moved: byte-identity empty, conservation unchanged, one tsv row changed, gate green, and each nested-stanza and name check re-run with its own firing positive control.
- **This pair ran at N=3 concurrency**, alongside another lane's suite. Concurrency on this box is proven safe at N=2 only, and the runs are known to contend on fork/exec rather than CPU, so N=3 is outside the measured envelope. Nothing reddened and the two logs agree line for line except the known nondeterministic class, but the concurrency itself is disclosed rather than claimed safe.
- The `lib/pithead/*` cut that landed under this branch rewrote the CLI these sections drive through `apply -y` and `run_pending`. That lane's own after-proof on it reported the same verdict as this branch's baseline, which is corroboration from a separate run, not from this PR.


VijitSingh97:
## Control run: green — same tree, same worktree, no changes

Posting the control result promised in the body, with its actual concurrency rather than the label it was launched under.

| | Run A | Run B (control) |
|---|---|---|
| Worktree | `pithead-split-proof` | **same** |
| Commit | tip | **same, no changes between runs** |
| Concurrency at launch | **N=3** | N=2 |
| Load at launch (12 cores) | ~7 | ~4.5 |
| Verdict | rc 1, **2977 passed / 3 failed** | rc 0, **2979 passed / 0 failed** |
| `#1342` section reached | yes | yes |

The control reached the `#1342` section and passed it. Since nothing about the tree differed between the two runs, load is the only variable that changed — which supports the explanation in the body and in #1495.

**Two honest qualifications, because this is weaker than a clean control.**

1. **It was not serial.** It was launched at N=2 and peer lanes joined it, so it reached N=4 before finishing. It is a lower-concurrency comparison, not the N=1 control it was meant to be.
2. **That it still passed at N=4 refines the mechanism rather than confirming it.** If average concurrency were what mattered, a run that touched N=4 should have been at least as likely to fail. It was not. The consistent reading is that only the load during the roughly ten seconds *that section* is executing matters — which is what a fixed wall-clock poll budget would predict, and which makes this a timing window, not a load threshold. It also means the failure is intermittent rather than deterministic at any given N, so a single green run does not prove a given concurrency level is safe.

Sample size is one run each way. This raises confidence in the load explanation; it does not settle it.

**The falsification condition in the body stands unchanged.** CI runs this suite serially at N=1. If `Shell tests` reds inside the `#1342` lifecycle block, that is evidence against this explanation and the right response is to reopen #1495 on a different mechanism — not to attribute it to a known flake and not to search this diff, which does not touch `tests/stack/test-lifecycle.sh`.

The equivalence proof for this cut is unaffected either way: both runs of the pair agreed at identical header order, 0 of 274 positions, with the three failures byte-identical on both sides.

VijitSingh97:
## CI is green, including the condition that could have gone against me

`Shell tests (shellcheck + pithead suite)` **passes**. All 15 checks pass; the 2 skipping are the shipped-image CVE sweeps.

This is the control that mattered. CI runs this suite **serially at N=1**, and the body committed in advance to reading a red there as evidence *against* the load explanation — not as the known flake. It came back green, so the explanation survives the test that could have killed it. Stating that explicitly rather than letting it pass silently, because a falsification condition only counts if its resolution is reported either way.

Where that leaves the three local reds: they were an N=3 concurrency artefact, filed as #1495 with its mechanism and a corroborating lower-concurrency green on the identical tree. They are not this cut, and `tests/stack/test-lifecycle.sh` is not in this diff.

Full proof set for the cut, every check with a firing positive control:

- Byte-identity empty **both directions**, rebuilt from `git show <base>:tests/stack/run.sh` rather than from the extraction; one-line-mutant control fired.
- Conservation closes: 1305 + 159 − 2 = 1462.
- Seam clean both ways, and the arithmetic closes independently: 167 lines out, 3-line stanza in, net −164 — matching the budget row's 1570 → 1406 drop exactly.
- Nested-stanza check 0 in range; control range returns 1.
- Budget gate green; control (delete the row) fails **by name**; base ref probed either side.
- Suite equivalence: identical header set, **identical ORDER, 0 of 274** positions, whole-output diff of exactly two lines (one ed25519 fingerprint). The order-diff control fired, and the sorted-multiset instrument was confirmed **blind** to the same permutation.
- `make lint` clean, run twice — once before and once after the rebase.

Awaiting the adversarial review; its verdict will be posted here before this goes to the controller to judge and merge.

VijitSingh97:
## Review verdict: MERGE (R7 gate pass, controller-judged)

Full adversarial pass at head `e7e0574` (base derived independently; the mid-review develop-v2 advance is disclosed in the pass and touches nothing this diff depends on). **Zero blocking findings.**

Re-derived by the pass, not relayed: the non-self-arm design attacked in both directions — the single `build_control_sandbox` call sits before the source stanza, all three exact-value downstream counters confirmed after it (plus two more delta-counters the body didn't name, correctly outside the fragile class), and the restore tail confirmed intact and last; byte-identity rebuilt from raw patch output (166/159 lines, diff empty, positive control fired); conservation closes exactly (1305+159−2=1462); the nested-stanza check returned zero with a 33-hit positive control; header order identical at 0-of-274 positions with the reviewer's own permutation control proving the instrument discriminates and the sorted-multiset instrument confirmed blind to the same permutation; the three moved-but-not-unset function names swept repo-wide (wider than the body claimed), zero external hits; the corrected `$WALLET` classification verified at source and no sibling name carries the same defect; budget gate green with a by-name firing control; shellcheck 0.11.0 + shfmt clean.

Rulings at merge:
- **The "calls the builder once" header count is ACCEPTED** — it counts occurrences of a named call and goes stale only on the exact event the header exists to warn about; both the reviewer and this seat judged it a structural fact, not a perishable count.
- **The local 3-red in the #1342 block is NOT this PR's defect**: identical failure content and position on both sides of the proof pair, the lock implementation untouched by this diff, and the load/timing explanation survived its own falsification — CI serial green in that exact block, independently corroborated by the author's local serial control (both 2979/0). The flake lives in #1495.
- **Should-fix on record (non-blocking):** the header cites `split-name-audit.py`, which exists nowhere in the tree or its history. Fold a one-word disclaimer (fleet-local scratch tool) or commit the tool, next time this file is touched. The claims it's credited with were independently re-verified by other means and are accurate.
- Reviewer suggestion, adopted as a recommendation only: a `: "${C:?}" "${WALLET:?}"` guard at the top of the domain file would turn an ordering break into a named abort.

Merging pinned to the reviewed head `e7e0574`.

VijitSingh97:
## Adversarial review verdict — posted for the record (it landed after the merge)

The review completed at roughly the same moment this merged, so it is recorded here rather than before the merge. **Its verdict was Approve, no blocking finding**, and it independently re-derived every load-bearing claim rather than relaying any of them — including running its own firing controls on the byte-identity diff, the ordered-header comparison, the nested-stanza grep and the budget gate, and widening the function-name reuse check from `tests/stack/` to the whole repository.

### Two things I derived after reading it

**1. Its one explicitly-unchecked item resolves cleanly.** The review flagged a total-count difference it did not chase: local N=3 was 2977+3 = **2980**, while CI and the serial control are 2979+0 = **2979**. The cause is the guard's own `bad`, which sits inside an `if` and therefore emits an assertion *only on the failure path*. Measured: `the unrecorded holder takes the window` appears **once in the red log and zero times in the green one**, while the other two appear once in both. So 2979 = the red run's 2977 passes plus the 2 assertions that flipped, and the third failure is an assertion that does not exist in a green run. Not a hidden defect — the self-diagnostic working as designed.

**2. Its non-blocking finding is correct and I am acting on it.** The header credits the dependency audit to a script that is lane-local and not in this repo, so a reader cannot re-run it. That is a real reproducibility gap even though the claims themselves re-derive (the review confirmed them by other means). The corrected wording drops the implication and states that each claim is written to be re-derived with git and grep alone. Since this landed before I could fold it in, it will ride the next cut in this series rather than churn a shipped file for a standalone comment change; the convention is corrected for every subsequent domain header.

The suggestion of a `: "${C:?}" "${WALLET:?}"` guard at the top of the domain file is a good one and is noted for the same follow-up — it would turn a broken-ordering failure into an immediate named abort instead of a confusing downstream mismatch.

---

<details><summary>Full review</summary>

# Adversarial review — PR #1494 (pithead), feat/1105-secrets-masking-v2 → develop-v2, R7

**BASE, derived myself, used everywhere below:**
`git -C ~/dev/pithead-split-work merge-base HEAD origin/develop-v2` = `492a41ce6c3495d9b4c181aa585e0f56cba620f6`
(NOTE: `origin/develop-v2` moved to `0b3a77b3ae28952efbcf671ff740a3d404447d7b` partway through my review — I fetched it near the end as an AXIS-6 base-move check. That is not a defect in this PR; it just means this BASE is now stale for anyone re-deriving after this report. Every check below used the one BASE pinned above, consistently.)

**Verdict: no blocking finding. Every load-bearing claim in the PR body and header that I attempted to check, I independently re-derived and it held.** Two low-severity/informational items below. This is the most thoroughly self-verifying PR body I've seen in this series — nearly every claim came with its own re-derivation recipe, and none of them broke under adversarial re-run.

## AXIS 0 — the non-self-arm decision

**(a) RE-DERIVED.** `awk` comment-stripped scan of `test-secrets-masking.sh`: zero calls to `build_control_sandbox`. `run.sh` comment-stripped: exactly one call, at `tests/stack/run.sh:229`, inside `== black-box: dashboard control channel (#33) ==`. The `source` stanza is at `tests/stack/run.sh:462`. 229 < 462, so the builder runs once, before the stanza, in a section that stays behind. Matches the PR exactly.

**(b) RE-DERIVED, and went further than the PR's own citation.** `$REQS/$RESULTS/$STAGED/$AUDIT` assigned at `run.sh:345-348` (`$C/data/control/{requests,results,staged,audit/control.log}`) — confirmed by direct read, matches PR's `:345-348` claim exactly. Exact-value counters: `run.sh:918` (`wc -l <"$AUDIT"`), `:966` and `:969` (`ls "$REQS" | wc -l` expecting `10` then `0`) — all three sit after `:462`. `run.sh:371` (`ls "$RESULTS" | wc -l` expecting `1`) sits before `:462`, correctly excluded. **I found two MORE `wc -l <"$AUDIT"` counters the PR did not name**, at `run.sh:389` and `:395` — but both sit before `:462` (irrelevant to the ordering argument) and compute a *delta* (`audit_after - audit_before`), not an absolute value, so they aren't in the same fragile class the PR is warning about even if they were downstream. Not a gap in the PR's claim.

**(c) RE-DERIVED.** The domain file's last three lines (`test-secrets-masking.sh:217-219`) are the restore tail: `control_config main` then `apply -y`, preceded by the comment `# Restore the section baseline (control on, no telegram) for the tests that follow.` Confirmed this is literally the last content in the file (no trailing lines after it) — the tail runs exactly where the PR says.

**(d) Independent judgment, not just verification.** The design holds up under everything I could throw at it — single caller before the stanza, three exact-value downstream counters genuinely coupled to `$C`, tail genuinely restores baseline. A third option (seed `$C`/`$WALLET`/the four spool paths locally instead of leaning on ordering) wouldn't remove the coupling, just hide it: the *values* still have to come from `build_control_sandbox`'s output, so a seed is either a duplicate of `run.sh`'s own definitions (drift risk, which the header names) or a re-derivation dressed up as independence. Given the shipped `test-control-add-only-ssrf.sh` precedent is real (verified below, not fabricated), I'd have made the same call. One thing I'd add that the PR doesn't: a cheap `: "${C:?}" "${WALLET:?}"` guard at the top of the domain file would turn a broken-ordering failure from a confusing downstream assertion mismatch into an immediate, named abort. Suggestion, not a finding.

**Precedent check, RE-DERIVED:** `tests/stack/test-control-add-only-ssrf.sh` also has zero comment-stripped occurrences of `build_control_sandbox` (confirmed by the same awk scan) — the cited precedent is real, not asserted-and-unchecked.

## AXIS 1 — byte identity, rebuilt from raw git output

**RE-DERIVED, not relayed.** `git diff BASE..HEAD -- tests/stack/run.sh` is one hunk, `@@ -458,172 +458,8 @@`. Extracted deleted lines directly from the patch (167 `^-` matches minus the `---` header = 166 lines) and diffed against `test-secrets-masking.sh` from its `echo "== black-box: pre-masked..."` line (line 54) to EOF (line 219): **`diff` reports empty, IDENTICAL.** 166 lines, 159 non-blank on both sides (`awk 'NF'` count), matching the PR's stated 166/159 exactly.

**Positive control fired:** mutated line 1 of the reconstructed content, re-diffed — reports the mutation by name (rc=1). The comparison genuinely tests something.

**Conservation, computed independently:** base non-blank (`git show BASE:... | awk 'NF' | wc -l`) = **1462**. Current `run.sh` non-blank = **1305**. 1305 + 159 − 2 = 1462. **Closes exactly**, matching the PR's formula.

**Added-line structure, RE-DERIVED (this is the one place I checked the PR's own methodology note rather than just its number):** the diff's added content is genuinely 2 lines (`^+` count 3, minus the `+++` header = 2: the shellcheck-source comment and the stanza). The blank line after the stanza, before `echo confirm-gate`, is a **shared context line** — it existed in the old file too, right before the deleted block's own end, and survived untouched. So "count the inserted block as three lines (comment, stanza, blank)" is a reconstruction convention for avoiding a phantom `462a463`, not a claim that three lines were literally added — I confirmed this distinction myself rather than taking the PR's wording at face value.

## AXIS 2 — nested-stanza and seam

**RE-DERIVED.** `git show BASE:tests/stack/run.sh | sed -n '458,629p'` (the full old-line range spanned by the hunk) grepped for `source "` (broadened past `$HERE/` for safety): **zero matches.** Positive control: the same pattern against the whole base file returns 33 matches (`source "$HERE/lib.sh"` plus 32 domain stanzas) — the grep genuinely fires when there's something to find. Nearest stanzas bracketing the cut range are `test-tor-network.sh` at base line 222 and `test-control-add-only-ssrf.sh` at base line 886 — both well outside 458-629.

**Seam, RE-DERIVED.** Base file at lines 450-462 and 625-634: no `# ---` divider immediately precedes either the old `echo pre-masked...` line or the `echo confirm-gate...` line. The nearest dividers are at base lines 224 and 1135, nowhere near the cut. Confirms "neither side carries a divider, none travelled, none was orphaned."

## AXIS 3 — anchor and execution order

**RE-DERIVED from raw log bytes, not from the PR's stated conclusion.** Used the existing proof-pair at `~/split-baselines/r7-492a41c/{baseline,after}.log` (dated 22:23-22:24, before I started; I did not generate these, see caveat below, but I independently recomputed every number from their raw content rather than trusting the PR's summary of them). Extracted all `^==` lines from both: 274 each. `diff` on the ordered extraction: **empty, 0 of 274 positions differ.** `diff` on the *sorted* extraction: also empty (expected — no permutation exists here to catch).

**My own positive control, not the PR's:** hand-permuted two adjacent headers in the `after` extraction and re-ran both diffs. Ordered diff correctly names the swap (`11d10` / `12a12`, identifying `check_stratum_exposure warning`); sorted-multiset diff on the *same permuted input* stays silent (rc=0). This independently confirms the instrument the PR relies on actually discriminates a permutation, which a header-count or sorted-multiset diff structurally cannot.

**Function-name reuse, RE-DERIVED and widened beyond the PR's own scope.** Grepped for `inject_reject`, `onion_control_config`, `tg_control_config` across all of `tests/stack/` — zero hits outside `test-secrets-masking.sh`. I then widened the search to the **whole repository** (`--include='*.sh' .`), not just `tests/stack/` as the PR claimed to check — still zero hits. Positive control: the same grep shape finds `irun` in `test-install-verify.sh`, confirming the tool fires.

## AXIS 4 — the header, the only authored surface

**(a) No line numbers, RE-DERIVED:** grepped the header (lines 1-34 of the new file) for `:[0-9]{2,}` and similar — no match. The "run.sh calls the builder once" line (`test-secrets-masking.sh:16`) is a semantic/structural count (occurrences of a named call), not a perishable line-number citation; I agree with the PR's own framing that this is defensible, not a violation of the lane's no-line-numbers rule.

**(b) Dependency list, RE-DERIVED including functions, not just `$VARS`:** extracted every function-shaped identifier actually invoked in the moved body (comment-stripped): `assert_eq, assert_contains, assert_not_contains, assert_rc, bad, control_config, file_mode, ok, run_pending`. This is an **exact match** to the header's claimed list, and `seed_control_env` is genuinely absent from both the header's claim and the actual usage.

**(c) `$WALLET` classification, RE-DERIVED — this is the one place the PR flagged its own prior mistake, so I checked hardest here.** `tests/stack/lib.sh:207` and `:221` are the only two `WALLET=` assignments in the repo (comment-stripped `awk` scan of `lib.sh` and `run.sh`), both `WALLET="${WALLET:-$VALID_PRIMARY}"`, both inside `build_val_sandbox()`/`build_control_sandbox()`. `run.sh` itself has **zero** `WALLET=` assignments (empty grep, confirmed). No reassignment of `WALLET` occurs anywhere between the builder call (`:229`) and the stanza (`:462`) either (checked the whole intervening range). The corrected header text is accurate. `$VALID_TARI`, by contrast, **is** a genuine top-level assignment at `lib.sh:133` — sitting between other top-level constant assignments (`VALID_PRIMARY`, `VALID_SUBADDR`, `VALID_INTEGRATED`), not inside any function body. No sibling name in the header's classification list carries the same defect as the corrected `$WALLET` entry.

**(d) `$REQS/$RESULTS/$STAGED/$AUDIT` "plain $C derivations", RE-DERIVED:** already shown at `run.sh:345-348` — yes, each is a bare `"$C/data/control/..."` string, and the domain file never reassigns any of them (only reads). The "not seeded" reasoning holds.

**Minor informational finding (low severity, not blocking):** the header (`test-secrets-masking.sh:15`) cites `split-name-audit.py` as the tool that audited the dependency list — I searched the whole repo and its full git history (`find` + `git log --all -- '*split-name-audit*'`) and it does not exist anywhere in the tracked tree. It may be a throwaway/fleet-local script (plausible, and not unlike the ad hoc commands I ran for this review), but as written it reads as a reusable, re-runnable instrument, and a future reader cannot re-run it to check this PR's own claim. I independently re-verified the claims it's credited with (the corrected `$WALLET` classification, the complete function list) by other means and they are accurate, so this doesn't undermine the PR's correctness — only its reproducibility trail.

## AXIS 5 — the two coupling rules

**(A) $C/results-dir**, same ground as AXIS 0(b): no assertion found. **(B) shared-file ambient**: the domain's own tail (AXIS 0c) is the mechanism that keeps `.env`/config state consistent for what follows — `control_config main` + `apply -y` restores exactly the baseline (`control on, no telegram`) that the next section (confirm-gate, `run.sh:632`) needs, and I confirmed the tail is intact and last. No moved section writes ambient state and leaves it inconsistent for a later still-in-`run.sh` test.

## AXIS 6 — budget and lint

**RE-DERIVED, all of it.** `git diff BASE..HEAD -- docs/dev/file-budget.tsv` touches exactly one row: `tests/stack/run.sh 1570 -> 1406`. `wc -l tests/stack/run.sh` on the actual file = **1406**, matching the tsv exactly. `wc -l tests/stack/test-secrets-masking.sh` = **219**, correctly under 400, no row exists for it. Diffed the base tsv against `dashboard/mining_dashboard/service/egress.py`, `.github/workflows/ci.yml`, `tests/integration/run.sh` — **not present anywhere in this diff**, confirming the "hand-restored, never touched" claim.

Ran the gate myself: `bash scripts/lint-file-budget.sh` → `file budget OK`. **My own firing control**: deleted the `tests/stack/run.sh` row from a scratch copy of the tsv, re-ran the gate → `file-budget: FAIL — tests/stack/run.sh is 1406 lines, over the hard ceiling of 800 for a new/unbudgeted file` — fails by name, as claimed. Restored the tsv, confirmed `git status --short` clean afterward.

`git rev-parse origin/develop-v2` at the start of this axis returned the same BASE sha I pinned at the top — unmoved through the gate check. (It moved later, near the end of my review, per the note at the top — after all budget/lint work was done.)

`shellcheck --severity=warning` (0.11.0, no `-x`, matching `make lint`'s invocation) on both `test-secrets-masking.sh` and `run.sh`: **clean, rc=0 both.** `shfmt -i 4 -d tests/stack/test-secrets-masking.sh`: **clean, rc=0, no diff.** I ran these directly on the two files rather than through `make lint-sh`, avoiding the known OOM path.

## AXIS 7 — the suite proof

**RE-DERIVED independently from raw log bytes** (`~/split-baselines/r7-492a41c/{baseline,after}.log` — same caveat as AXIS 3: I read, ANSI-stripped, path-normalized, and diffed these myself; I did not launch the run that produced them — see honesty note below). Sorted-multiset diff of the full stripped output: **exactly one differing pair (2 lines)** — `line 2609`, both halves of one ed25519 key fingerprint (`SHA256:LdFFg...` vs `SHA256:jhYfn...`), the disclosed #1325 nondeterminism class. Both logs are 3297 lines. This matches the PR's claim exactly, independently recomputed.

Located the 3 failing assertions myself by direct `sed` inspection (my first grep pass mis-attributed the enclosing section — corrected by reading context directly, see honesty note): all three sit under `== unit: the mutation lock serialises mutating windows (#1342) ==` in `test-lifecycle.sh:369`, identically positioned in both logs, with the test's own self-diagnostic message `"the lock is free, so the three cases below prove nothing"` present on both sides. `git diff --stat BASE..HEAD` touches exactly 3 files (`file-budget.tsv`, `run.sh`, `test-secrets-masking.sh`) — `test-lifecycle.sh` is not in this diff at all, confirmed by git directly, not by the PR's list.

`CONFIG_FILE: readonly variable` warning count: 3 on each side (grepped both logs myself), consistent with the PR's "shift of zero" prediction for a file this cut doesn't touch.

## AXIS 8 — proven tree vs committed tree

**RE-DERIVED.** Single commit on this branch past BASE: `e7e0574`, committed `2026-08-28T22:10:36-05:00`. `after.log`'s mtime is `22:23:55`, ~13 minutes **after** the commit. `git status --short` in the working worktree is clean and `git diff HEAD` on the three changed files is empty — the tree that would have produced `after.log` matches HEAD's committed tree exactly, with no later edits. No proven-tree/committed-tree gap, as claimed.

## AXIS 9 — the suite is red on both sides locally, green on CI

**(a) RE-DERIVED**, see AXIS 7 — same 3 failures, same section, same position both sides, file genuinely untouched by this diff.

**(b) RE-DERIVED** directly from `test-lifecycle.sh:369-415`: `LKDIR="$SANDBOX/lock"` and `lib.sh:81` `SANDBOX="$(cd "$(mktemp -d)" && pwd -P)"` — `LKFILE` is mktemp-private per run, not a host-global path, so this is not #1104-class cross-talk. `lock_await_record()` (`test-lifecycle.sh:405-410`) loops `i<200` with `sleep 0.05` — exactly a 10-second hard wall-clock budget, confirmed by reading the loop, not by taking the PR's arithmetic.

**(c) RE-DERIVED, not from the PR's list.** `git diff --stat BASE..HEAD` (computed by me, independent of anything the PR states) touches only `docs/dev/file-budget.tsv`, `tests/stack/run.sh`, `tests/stack/test-secrets-masking.sh`. `git diff --stat BASE..HEAD -- lib/pithead/00-prelude.sh pithead` is **empty** — the mutation-lock implementation is untouched by this branch, confirmed directly.

**(d) RE-DERIVED, this is the falsification the PR itself asked for.** CI's "Shell tests (shellcheck + pithead suite)" was still `pending` when I started; I polled it in a capped loop (`gh pr checks 1494`, 30s interval) rather than idling on a Monitor. It finished **green**, `2979 passed, 0 failed`, in 11m32s — genuinely serial CI (N=1). I pulled the job log directly and located the `#1342` section by name: **every one of its assertions passes on CI**, including the three that failed locally, with the same test names I saw fail locally now showing `✓`. As a second, independent confirmation I found while checking whether the gate's base had moved: the author's own **local serial control** (`~/split-baselines/r7-492a41c-serial/`, still running when I first looked, at `pithead-split-proof` pinned to this exact BASE) finished during my review with **rc=0, 2979 passed, 0 failed** — matching CI's count exactly and also fully green in the #1342 section. Two independent serial (N=1) runs, one remote and one local, both green where three concurrent (N=3) runs were red in the identical spot. The load hypothesis survives its own stated falsification test.

Minor, unexplained, not investigated further: local-N=3 total was 2980 (2977+3), CI/serial-N=1 total is 2979 — a 1-test difference I did not chase down (labeled **unchecked**); it does not touch the #1342 question and is far more likely environment/skip-related than a hidden defect.

**(e) Judgment, not a survey: YES, this PR should merge on this evidence**, given that CI is now green. Before CI returned, my answer would have been no — a red local after-proof plus a plausible-sounding "it's just load" story is exactly the shape of a convenient explanation the task asked me to distrust, and the equivalence proof alone (identical failure content/position, not just identical outcome) is unusually strong but still circumstantial without an N=1 control. With CI now green in the exact block in question, and a second independent local-serial confirmation landing during my review, the explanation is no longer just plausible — it's falsifiable-and-surviving. #1495 is the right home for the concurrency flake itself, not this PR.

## Honesty notes — what I did NOT do, and one correction of my own

- **I did not run the full suite myself from scratch.** AXIS 3 and AXIS 7's raw numbers come from `~/split-baselines/r7-492a41c/{baseline,after}.log`, an existing artifact I found already on disk (mtimes ~22:23-22:24, before I started, matching this lane's `rN-<base-sha>` naming convention from prior rounds) and confirmed corresponds to `pithead-split-proof` sitting exactly on BASE `492a41c`. I did not launch that run. What I *did* do is independently recompute every number I cited from it (header extraction, sorted-multiset diff, my own permutation control, byte counts) rather than trust the PR's stated conclusions about that same log — so the arithmetic and the section-identification are mine, the underlying execution is relayed. A full second from-scratch run (base + after, ~12 min each) was not taken given the size of this review; if that matters, it's the next thing to do.
- **Self-correction, disclosed rather than hidden:** my first attempt to locate the 3 failing tests' enclosing section used a `grep 'FAIL\|failed\|✗'` pass and mis-attributed them to `== black-box: apply recovers from a failed 'compose up' (#125) ==` (the nearest line matching my pattern) instead of the correct `== unit: the mutation lock serialises mutating windows (#1342) ==`. Caught by reading the raw log directly with `sed` instead of trusting a keyword grep — the section header itself doesn't contain the word "fail". Worth remembering: a keyword-based section-attribution grep is not reliable near dense "FAIL"-heavy fixture output.
- **Did not independently verify** the dashboard/CLI/lint/frontend CI jobs beyond confirming they show `pass` — no reason to given this PR's diff is 3 files, none of which those jobs cover.
- **Did not check** rig/loaner/docker/real-network behavior — nothing in this diff touches that surface.
- **Everything else claimed in the PR body that I did check, I re-derived myself**, not relayed; I did not find a single re-derivation that came out differently from the PR's stated conclusion.

## Recommendation

Approve. One optional, non-blocking suggestion for the author: consider naming/committing (or explicitly disclaiming as scratch) the `split-name-audit.py` tool cited in the header, since as written it implies a reproducible check that a future reader cannot actually re-run.

</details>

Issue #1495, opened 2026-08-29
tests/stack: the #1342 mutation-lock test has a fixed 10s wait that reddens the suite under concurrent runs


VijitSingh97:
`tests/stack/run.sh` reddens with three failures under concurrent runs, in a section that is correct and well built. The failure is a fixed wall-clock budget in the test, not a defect in the code under test.

## What happens

Running the stack suite while two other suites are running on the same box (N=3, load ~7 on 12 cores) produces:

```
== unit: the mutation lock serialises mutating windows (#1342) ==
  ✗ the unrecorded holder takes the window
      the lock is free, so the three cases below prove nothing
  ✗ an unrecorded holder still blocks the window
      expected rc 75, got 0
  ✗ an unrecorded holder is reported as unrecorded
```

Verdict `2977 passed, 3 failed`, rc 1. Reproduced simultaneously in two independent worktrees — one of them a pristine checkout of the tip with no local changes.

## Mechanism

In `tests/stack/test-lifecycle.sh`, the `#1342` block needs an *unrecorded* lock holder — a holder that is not `pithead` and therefore writes no holder record. It creates one by backgrounding a subshell that opens a descriptor on the lock file, takes `flock -n`, and `exec`s a long `sleep` onto it. It then waits for the lock to actually be taken:

```
i=0
while [ "$i" -lt 200 ]; do
    [ "$(lock_state)" = "held" ] && break
    sleep 0.05
    i=$((i + 1))
done
```

That is a hard **10-second wall-clock budget** for the backgrounded process to be scheduled and acquire the lock. Under enough load it is not enough, the poll exhausts, and the two assertions that follow measure an uncontended run — `rc` comes back 0 instead of 75, and the message assertion sees ordinary `stack_down` output.

**The test is not at fault for reporting this, and it deserves credit for it.** It does not pass vacuously: it detects the exhausted poll and fails with `the lock is free, so the three cases below prove nothing`, and its own comment says the third case would otherwise "pass for the worst reason available — nothing was reported under ANY name". The defect is only that the budget is fixed rather than load-aware.

## Explicitly ruled out — please do not re-litigate these

- **Not a `#1104`-class host-global path.** `LKFILE` is `"$SANDBOX/lock/explicit.lock"` and `$SANDBOX` is `mktemp -d`, so the lock file is private to each run. Concurrent suites cannot contend for this lock. This is scheduling latency, not shared state.
[...]

PR #1496, opened 2026-08-29
refactor(cli): cut backup/restore and help/version out of the remainder (#1105 Phase 2 P5)

VijitSingh97:
#1105 Phase 2 P5: cut backup/restore and help/version out of the remainder.

Two slices, both under the 400-line floor, so **neither takes a budget row**:

| slice | lines | contents |
|---|---|---|
| `lib/pithead/17-backup-restore.sh` | 377 | `backup_require_items`, `backup_diagnose_items`, `stack_backup`, `stack_restore` |
| `lib/pithead/18-help-version.sh` | 190 | `show_help`, `show_version` |

`lib/pithead/99-remainder.sh` drops **8517 → 7948**, which is the machine-checked statement that the
split progressed. Outside `lib/pithead/` the entire diff is that one tsv row.

**Conservation, machine-checked:** 377 + 190 = 567 slice lines + 2 build-supplied blank separators
= 569 = 8517 − 7948.

## What was RUN

Everything below was re-run **at the rebased head**, not inherited from the pre-rebase cut. PR #1494
landed while this was being cut; see the rebase section.

- **Parity** — `scripts/build-pithead.sh --check` rc 0, "exactly what 20 slice(s) build".
  `pithead` is absent from `git status`: the artifact is untouched.
- **The parity gate PROVEN TO FIRE.** Mutating `18-help-version.sh:1` gives `git diff --numstat`
  1/1 (the mutant is proven to have applied *before* the gate runs — an unstaged/untracked file
  prints nothing and would make a blind control look clean), the gate then returns rc 1 naming
  `4146c4146` and quoting both sides, and restoring returns it to empty numstat + rc 0.
  **Two independent routes to 4146:** the offset arithmetic (region line 379 + 3767) and the gate's
  own output. One derivation is not a control; two agreeing is.
- **Every range verified by EXTRACTION, not arithmetic.** `awk 'NR>=a&&NR<=b'` over the base
  remainder, `cmp`'d against each committed file — all three exact (both slices and the new
  remainder).
- **`bash -n` on each of the 20 slices individually AND on the built artifact.** This is the heredoc
  instrument and it matters more than usual here: `show_help` is one long `cat <<EOF`, and the blank
  lines at region-relative 549 and 552 are *inside* that heredoc. Byte-identity is blind to a
  boundary that falls inside a heredoc — concatenation puts it back and parity passes — so per-slice
  parsing is what rules it out.
- **`shfmt -i 4 -d`** over `lib/pithead/` — 0 bytes.
- **File budget** rc 0, with the base ref probed before *and* after the run and unmoved across it.

## Suite after-proof

**rc 0, `pithead tests: 2979 passed, 0 failed`**, 3292 output lines, 274 section headers, and
`grep -c '✗'` = **0**. Run in the worktree checked out to `d787f74` itself — not a reconstruction —
and `git status` was empty after the run, so the tree was not edited mid-flight. The verdict is read
from the **rc file**, not from the launcher's own exit status, which is always 0 for a detached run.
A case-insensitive `FAIL` grep is useless on this log (the suite names tests after the condition they
assert, e.g. `tor-running check: tor down + mining up -> FAIL ✓`), which is why the glyph is counted.

Log: `~/appliance-suite-logs/p5-after-20260829T034713Z.log`.

**The comparator, stated precisely — this is an after-proof at the head, NOT a controlled
before/after pair.** There is no fresh baseline at `a3233d8`. The comparator is P4's after-proof at
`6a628d2`, which reported the same four numbers (2979/0, 3292 lines, 274 headers). **The honest
caveat is that #1494 landed in between and it DOES touch `tests/stack/`** — the very tree this suite
runs — so unlike P4's comparator, this one is not insulated by an intervening merge that touches
nothing the suite reads. What the identical totals do support is that #1494 was itself a pure move
within the suite; they are consistent with that, not proof of it. CI's Shell tests job is the
independent run.

## The rebase — and why the first budget run failed

The first budget run **FAILED naming `tests/stack/run.sh`, a file this diff never touches**
(ceiling 1406 → 1570). That is the documented signature of a moved base, not a defect in the cut:
`origin/develop-v2` advanced from `0b3a77b` to `a3233d8` *during the run* (probed either side), as
PR #1494 landed and lowered that row.

On a stale base this gate fails in both directions and no edit fixes it — the fix is the rebase.
So: rebased onto `a3233d8`, then re-ran **every** proof above, because a rebase invalidates the
whole proof set rather than just the suite.

The mandated re-cut check —
`git diff --stat <base>..origin/develop-v2 -- pithead lib/ scripts/build-pithead.sh` — came back
**empty**, so #1494 forced a rebase, not a re-cut. It touches only `tests/stack/` and the tsv.

## Budget proof

Base `a3233d8`, probed before and after, unmoved. **The tsv diff against the base is exactly one
row** — `lib/pithead/99-remainder.sh 8517 → 7948`. Both new slices are under the 400-line floor
(`scripts/lint-file-budget.sh` `TARGET_LINES=400`; a row is required iff over target), so neither
takes a row and none is added.

**Byte-identity is structurally blind to slice ASSIGNMENT.** A cut at any single blank line between
two top-level blocks reproduces the artifact byte for byte, including a semantically wrong one. Only
reading validates the assignment, so here is where to push:

1. **`18-help-version.sh` sits UNDER the `# --- Backup / Restore ---` header.** That header is line 1
   of the region and the next source header (`# --- Small Utilities ---`) does not arrive until after
   the region ends. So `show_help`/`show_version` are inside that header's span while not being
   backup/restore: **the source's own header does not bound this file's contents.** Same shape as
   P4's `16-reset.sh` and P3's slice `08`, disclosed for the same reason.
2. **The two `backup_*` helpers are called only from `stack_backup`** (artifact lines 3885, 3969,
   3975, 4017 — all inside `stack_backup`), *not* from `stack_restore`. So the backup/restore pairing
   in `17-backup-restore.sh` rests on the source's own header and the CLI dispatch, **not** on shared
   helpers. If you think that pairing is wrong, this is the fact to attack — I am not claiming the
   helpers force it.

## Over-engineering pass, run BY HAND

The ponytail gate reads the branch name from the pinned shell cwd rather than the worktree the
command ran in, so it cannot do this pass here. Run by hand instead:

**The diff is a pure move, measured rather than eyeballed.** Multiset diff over `lib/pithead/`:
569 removed / 567 added; `comm -13` (added-not-removed) is **empty**; `comm -23`
(removed-not-added) is **2 lines, both blank** — the build's separators. Sides extracted with
`awk '/^--- /{next} /^-/{print substr($0,2)}'`, never `grep '^-[^-]'`, which drops removed blank
lines and makes a pure move read as a content change. **There is no new logic to over-engineer.**

**Is two files more than the region needs?** Order-preserving contiguous slicing means only
ADJACENT merges are available, and there is exactly one adjacent pair:

| merge | lines (with +1 separator) | legal? |
|---|---|---|
| `17` + `18` | 377 + 1 + 190 = **568** | **NO** — over the 400 floor, would need a new tsv row |

So **two files is the forced minimum, not a naming preference.**

**Splits I could have made and declined, with the evidence:** the region has two further legal cut
points, at region-relative 270 and 556.

- **backup | restore at 270** → 269 / 107 / 190, all legal. **Declined:** the source's own
  `# --- Backup / Restore ---` header names them one domain, and the CLI dispatch table has them as
  adjacent cases (`backup) stack_backup` and `restore) stack_restore`, consecutive lines). Stated
  honestly: the helpers do *not* force this, per the disclosure above.
- **help | version at 556** → 377 / 177 / 12. **Declined:** `show_version` is 9 lines; isolating it
  produces a file shorter than its own header comment. Both are pure CLI informational output,
  dispatched on consecutive lines (`version | -V | --version` and `help | -h | --help`).

Three and four files were available and were rejected on naming, not overlooked.

## What was NOT done

- **`make lint` / `make lint-sh` not run; the lint lock was not taken.** `make -n lint-sh` shows one
  `shfmt -i 4 -d` invocation reaching `lib/pithead` and neither `shellcheck` invocation naming it, so
  a slice cut's entire *new* exposure under that target is the shfmt pass — run clean above. Note
  this is about *new* exposure: `pithead` is shellcheck's first argument and is the byte-for-byte
  concatenation of these slices, so the content itself is analysed, in its real assembled context.
  CI's Shell tests job is the arbiter.
- **No bench run.** Nothing in P5 changes behaviour; the artifact is byte-identical.
- **No non-author review yet** — that is the merge gate, not something this body can satisfy.


VijitSingh97:
## Controlled suite baseline at the PR's base — the weak comparator in the body is now replaced

The body discloses that its suite evidence is an **after-proof at the head, not a controlled pair**:
there was no fresh baseline at the base, and the comparator (P4's after-proof) sits on the far side
of #1494, which touches `tests/stack/` — the tree the suite runs. That caveat stands as written. This
comment retires it: **the controlled pair now exists.**

### Why this is a real control, not just a second run

`a3233d8` is simultaneously the PR's merge-base, `HEAD^`, **and** the live `develop-v2` tip — probed
before **and** after the run and unmoved across it (`git ls-remote origin develop-v2`, both sides).
Every worktree on this box shares one `.git`, so that probe is the load-bearing one: a sha-pinned
worktree pins content, not refs. **P5's diff is therefore the only variable between the two runs.**

Both runs were executed in worktrees checked out to the commit itself — not a reconstruction — and
both trees were clean (`git status --porcelain` empty) after their run, so neither was edited
mid-flight. The verdict is read from the **rc file**, never the launcher's exit status.

| | base `a3233d8` | head `d787f74` |
|---|---|---|
| rc (from the rc file) | **0** | **0** |
| `pithead tests:` | **2979 passed, 0 failed** | **2979 passed, 0 failed** |
| output lines | **3292** | **3292** |
| section headers (`^== `) | **274** | **274** |
| failure glyph `✗` | **0** | **0** |

Base worktree independently reproduces the "before" half of the cut: **18 slices, remainder 8517**.

### The four numbers are the weak version of this result — here is the strong one

Four totals off one run are close to one measurement printed four times, so they are not what I am
asking you to weigh. **A sorted-multiset diff of the two complete 3292-line logs differs by exactly
9 lines per side, and every one is accounted for by a variable that is not P5:**

- **7 lines — worktree path prefix only.** One fixture-presence line and six stderr lines; identical
  once the worktree directory name is normalised. Structural: the two runs are in different worktrees.
- **1 line — the ed25519 keygen fingerprint.** The known non-determinism (#1325); this is the same
  single line the tests lane isolated in its N=2 concurrency result.
- **1 line — the `mktemp` sandbox path** in the ingredients-manifest line.

Normalising exactly those three (worktree dir, `mktemp` path, key fingerprint) makes the multiset
diff **empty**. So the two runs are byte-identical modulo per-run paths and one documented
non-deterministic key.

**Both instruments carry a firing control, because a comparison that cannot fail proves nothing.**
`cmp` and the multiset diff were re-run against a copy of the head log with one line deliberately
altered: both detected it. The *same normalisation* was then applied to that mutant and still
reported the difference — so the normalisation is not over-broad and is not masking real changes.

### An independent corroboration of `pithead` byte-identity, by a different route

The suite emits `pithead: line 8099: DASHBOARD_AUTH_HASH_B64: unbound variable` on both sides — the
**same line number**. That is a runtime bash diagnostic, derived from the assembled artifact at
execution time, not from a diff. It agrees with the parity gate about the artifact being unchanged
while sharing no machinery with it. Two independent routes to the same fact is the claim; the parity
gate alone was one.

### What this does NOT prove

- **It says nothing about slice assignment.** The four disclosures in the body stand untouched — a
  cut at any single blank line between two top-level blocks reproduces the artifact byte for byte,
  including a semantically wrong one, and an identical suite result is equally blind to it. Reading
  is still the only instrument for assignment, which is why those four are called out as attack
  surface rather than defended.
- **It is one sample per side, not a distribution.** It rules out a deterministic change in what the
  suite exercises; it does not characterise flakiness.
- **`make lint` / `make lint-sh` still not run** — unchanged from the body. CI's Shell tests job
  remains the independent arbiter, and this local pair does not substitute for it.

Logs, on the box, outside the repo: `~/appliance-suite-logs/p5-base-20260829T041230Z.{log,rc}` (this
baseline) and `~/appliance-suite-logs/p5-after-20260829T034713Z.{log,rc}` (the head run already cited
in the body).

VijitSingh97:
## Gate pass (adversarial review, non-author) — verdict MERGE, judged and executed by the controller

Review spawned as an ephemeral at 04:10:08Z against head `d787f74`; verdict below pasted verbatim per the durable-verdict rule. Controller re-derived state/mergeable/headRefOid and the develop-v2 tip at source immediately before merging: all unmoved from the reviewed state.

# Adversarial review — PR #1496 (p2pool-starter-stack/pithead)

`feat/1105-p5-backup-restore-help` -> `develop-v2`. #1105 Phase 2 P5.

## Verdict: **MERGE**

No blocking findings. Every load-bearing claim in the PR body was re-derived from raw
sources (git blobs, a live mutation control, CI's raw job log) rather than accepted on
the author's word. One terminology quibble, non-blocking, noted at the end.

## Base/head, self-derived

- `gh pr view 1496 --json headRefOid` -> `d787f74513972faeb532d04978965271d4ed3f36`.
- `git ls-remote origin develop-v2` -> `a3233d8a82a3b4d5e4a9d3ed1f99d33222445028`, matching
  the PR body's claimed base. Re-probed after ~25 minutes of review work: **unmoved**
  (both probes returned `a3233d8`).
- Worktrees: `/tmp/reviewer-1496/{base,head}`, detached from `~/dev/pithead`'s shared
  `.git` (never touched `~/dev/appliance-p5`, the author's own worktree on this branch).
  Both removed at the end of this review.
- `git merge-base HEAD a3233d8` == `a3233d8` and `merge-base --is-ancestor a3233d8 HEAD`
  is true: head is a clean rebase onto the live tip, not an unrebased branch masquerading
  as one.

## Axis 0 — slice assignment (re-derived, not accepted)

**(a) Heredoc boundary.** `bash -n` on all 20 committed slices individually: zero errors,
zero output. `bash -n` on the rebuilt artifact: clean. Read `18-help-version.sh` directly:
the only heredoc (`cat <<EOF`) opens at line 2 and closes at line 176 — fully contained
inside the file, nowhere near either of its two boundaries (start: `show_help() {` at
line 1; end: `show_version`'s closing `}` at line 190). The cut could not have landed
inside it. **Confirmed.**

**(b) Header does not bound the file.** Read the base region directly:
`grep -n '^# --- ' lib/pithead/99-remainder.sh` shows `# --- Backup / Restore ---` at
line 1 and the next header, `# --- Small Utilities ---`, at line 570. Everything in
between — `backup_require_items`(14), `backup_diagnose_items`(28), `stack_backup`(46),
`stack_restore`(271), `show_help`(379), `show_version`(560) — sits under that one header,
including the two functions that end up in `18-help-version.sh`. The disclosure is
accurate. The file's name survives judgment: `18-help-version.sh` is honestly named for
its actual contents, independent of the misleading source-side header. Same shape as P4's
`16-reset.sh`, as claimed.

**(c) Call graph.** `grep -n 'backup_require_items\|backup_diagnose_items' pithead` on the
rebuilt head artifact: call sites at absolute lines 3885, 3969, 3975, 4017 — matching the
PR body's citation exactly. `stack_backup` spans 3813–4037, `stack_restore` spans
4038–4145: all four call sites fall inside `stack_backup`'s range, none inside
`stack_restore`'s. **Zero function-level coupling between backup and restore, confirmed.**
Judgment on whether they're one domain or two: CLI dispatch has `backup) stack_backup ;;`
and `restore) stack_restore ;;` on consecutive lines (12231–12232), and they're inverse
operations over the same path set (config.json, `.env`, Caddyfile, Tor data dir per the
file's own header comment). Pairing by convention (like save/load) without shared code is
a defensible, non-blocking call — the author disclosed the exact fact that would let a
reviewer disagree, and disclosed it honestly.

**(d) Blank-line boundaries.** Both actual cut boundaries checked at the base commit:
line 377 (`}`, end of `stack_restore`) / 378 (blank) / 379 (`show_help() {`), and line 568
(`}`, end of `show_version`) / 569 (blank) / 570 (`# --- Small Utilities ---`). Exactly one
blank line at each. `scripts/build-pithead.sh` independently confirms the mechanism: it
FATALs on a slice that opens or ends with a blank line, and joins slices with exactly one
build-supplied blank — so any cut at a lone blank line between top-level blocks reproduces
the artifact byte-for-byte, which is also why byte-identity can't discriminate assignment.

## Axis 1 — file count (recomputed)

`17-backup-restore.sh` (377) + `18-help-version.sh` (190) = 567 content lines; the one
possible adjacent merge is 377 + 1 (separator) + 190 = **568**, over `TARGET_LINES=400`
(`scripts/lint-file-budget.sh`), which would require a new tsv row. Two files is therefore
the minimum that avoids a row — confirmed arithmetically.

Both declined 3-file splits verified at their exact stated boundaries: the backup|restore
split at base-relative line 270 (blank line, `stack_backup`'s `}` at 269 / `stack_restore`
at 271) yields 269/107/190 — matches the body exactly. The help|version split at
base-relative line 556 (blank line between `show_help`'s `}` at 555 and the show_version
doc-comment at 557) yields 377/177/12 — matches exactly. Both are legal blank-line
boundaries; declining them on naming (informational-output pairing, dispatched on
consecutive lines) rather than on any structural necessity is accurately represented as a
choice, not a constraint.

**One quibble, non-blocking:** the body calls the adjacent merge "ILLEGAL." It isn't —
plenty of budgeted files sit over 400 with a row (`os/rootfs/Dockerfile` 406,
`scripts/release.sh` 851). What's true is that merging would *cost a new row*, which
conflicts with this split lane's established convention of keeping new `lib/pithead/`
slices under the floor. "Forced to avoid a new row" is accurate; "illegal" overstates a
convention as a hard rule. Not worth a body edit, just don't repeat the framing.

## Axis 2 — parity and pure-move (rebuilt, not trusted)

- `scripts/build-pithead.sh --check` at head: `pithead parity OK — ... 20 slice(s)`, rc 0.
- Full regenerate (`scripts/build-pithead.sh` with no `--check`) then `git status
  --porcelain`: empty — the regenerated artifact is byte-identical to the committed one.
- `cmp` of head's `pithead` against base's `pithead`: **identical**. Byte-unchanged,
  confirmed directly, not via the gate.
- **My own firing control**, staged before checking (per the "mutation battery must prove
  it mutated" lesson — an untracked/unstaged edit makes `--numstat` read empty and the
  control goes blind): edited `18-help-version.sh:1`, `git add -A`, confirmed
  `--numstat` showed `1 1` (mutation actually staged), ran `--check` -> **rc 1**, first
  differing line **`4146c4146`** — the exact line the PR body's independent
  offset-arithmetic route and its own gate-firing route both name. Three independent
  routes now agree on 4146. Restored with `git checkout HEAD -- <file>`, re-ran
  `--check` -> rc 0, tree clean.
- **Pure-move, recomputed with sort+comm**, not accepted: extracted added/removed lines
  from `git diff -- lib/pithead/` with `awk '/^\+\+\+ /{next}/^\+/{...}'` /
  `awk '/^--- /{next}/^-/{...}'` (never `grep '^-[^-]'`, which drops removed blanks and
  would make a pure move misread as content change). Result: **567 added, 569 removed**,
  `comm -13` (added-not-removed) **empty**, `comm -23` (removed-not-added) **2 lines, both
  blank**. Matches the body's claim exactly, independently derived.

## Axis 3 — conservation from git blobs

Located the region's absolute offset in the base `pithead` artifact directly
(`grep -n '^# --- Backup / Restore ---$' pithead` -> line 3768, giving offset +3767,
matching the body's stated 379+3767=4146). Extracted both ranges from the **base blob**
with `awk 'NR>=a&&NR<=b'` and `cmp`'d against the committed slices:

- `awk 'NR>=3768&&NR<=4144' pithead` vs `17-backup-restore.sh` (head): **identical**.
- `awk 'NR>=4146&&NR<=4335' pithead` vs `18-help-version.sh` (head): **identical**.
- Lines 4145 and 4336 (the two dropped separators): both blank, confirmed.
- `tail -n +570 <base 99-remainder.sh>` vs the head's `99-remainder.sh`: **identical**
  (7948 lines both sides).

377 + 190 = 567, +2 separators = 569 = 8517 − 7948. Every number in the conservation
equation reproduced from raw blobs, not arithmetic taken on faith.

## Axis 4 — budget

`git diff <base> <head> -- docs/dev/file-budget.tsv`: exactly one hunk, one row,
`lib/pithead/99-remainder.sh 8517 -> 7948` — a lowering, no other row touched. Base ref
(`git ls-remote origin develop-v2`) probed before the review and again after ~25 minutes
of work: unmoved at `a3233d8` both times. `shfmt -i 4 -d` on both new slices: rc 0, zero
bytes of diff — confirmed directly, not relayed.

## Axis 5 — suite/CI (was the weak axis; now closed)

The body's own disclosure was accurate as written: the after-proof at head had no fresh
base-side control, and the comparator (P4's after-proof) predates #1494, which does touch
`tests/stack/`. Polled PR comments in a capped loop (two rounds, ~21 minutes total,
`gh pr view --json comments` every 45–60s). The controlled baseline landed at 04:27:55Z
UTC (comment from the author, ~13 minutes after the PR body's own estimate):

| | base `a3233d8` | head `d787f74` |
|---|---|---|
| rc | 0 | 0 |
| pithead tests | 2979 passed, 0 failed | 2979 passed, 0 failed |
| output lines | 3292 | 3292 |

The comment goes further than the four-number table and reports a sorted-multiset diff of
the full logs differing by exactly 9 lines per side, every one attributed to a named
non-P5 variable (7 worktree-path lines, 1 keygen fingerprint — the documented #1325
non-determinism, 1 mktemp path), with the comparison instrument itself given a firing
control (a deliberately mutated copy of the log, detected, then re-normalized and still
detected — ruling out an over-broad normalization masking a real difference). I did not
re-run this myself; I weighed it as reported, since it carries its own controls.

**CI's independent run, pulled from the raw job log** (`gh run view ... --log` on the
"Shell tests" job, run 33232988529 / job 99048937169, conclusion `success`):
`pithead tests: 2979 passed, 0 failed` — same total as both local runs. Checked the
`#1342` mutation-lock block by name (`== unit: the mutation lock serialises mutating
windows (#1342) ==` and the adjacent appliance-boot-leg block): both pass cleanly, no
sign of the #1495 serial-vs-load question. This axis is now proven, not merely disclosed.

## Axis 6 — moved-base handling

`git log --oneline 0b3a77b..a3233d8`: exactly one merge, PR #1494
(`feat/1105-secrets-masking-v2`), with `0b3a77b` = merge of PR #1493 (P4) as its direct
parent — matches the body's account of what preceded the rebase. `merge-base --is-ancestor
a3233d8 <head>` already confirmed head is rebased cleanly onto the live tip (see above).
Re-ran the mandated re-cut check myself: `git diff --stat 0b3a77b a3233d8 -- pithead lib/
scripts/build-pithead.sh` is **empty** — confirmed, #1494 forced a rebase, not a re-cut.
(Widening that check to also include `docs/dev/file-budget.tsv`, as this review's brief
asked, shows the one line already known — the `tests/stack/run.sh` ceiling row #1494
itself changed, 1406->1570 — a different row than the one this PR touches. Not a
discrepancy; the author's mandated check correctly excluded the tsv because #1494 was
known to touch it for an unrelated row.)

## Axis 7 — hygiene

- `gh api graphql` for `closingIssuesReferences` on PR #1496: **`[]`** — #1105 will not
  auto-close. Confirmed via the graphql route, not relayed.
- `make -n lint-sh`: exactly two `shellcheck` invocations, neither globs `lib/pithead`
  (`pithead pithead-completion.bash install.sh scripts/*.sh build/*/*.sh dashboard/*.sh`
  and, separately, `tests/stack/run.sh`); one `shfmt -i 4 -d` invocation explicitly lists
  all 20 `lib/pithead/*.sh` files including both new slices. Matches the settled
  reasoning (shellcheck reaches slice content via the concatenated `pithead` artifact,
  which is shellcheck's first argument) — confirmed directly, no new finding, as the ask
  anticipated.
- Wallet/topology sweep: `detect-topology.py` over the full diff, the PR body, and every
  PR comment — **clean**, all three. Positive control fired first, on
  `tests/stack/lib.sh` (the known `WALLET=$VALID_PRIMARY` fixture): 8 hits. Detector is
  live, not silently broken.

## Re-derived vs relayed vs unchecked, ranked by cost of being wrong

**Re-derived myself, highest cost of being wrong first:**
1. Base/head SHAs, merge-base, rebase-not-recut status (Axis 6) — everything else depends
   on this being right.
2. Byte-for-byte conservation from raw base blobs (Axis 3) — the single strongest proof
   that no content was lost or altered; nothing here depends on the author's tooling.
3. Parity + my own independent firing control at line 4146 (Axis 2) — three converging
   routes now, not one.
4. Pure-move via sort+comm (Axis 2) — ruled out any non-move edit hiding in the diff.
5. Heredoc/header/call-graph/blank-line facts (Axis 0) — read directly from both base and
   head, not accepted from the disclosure.
6. `closingIssuesReferences`, shellcheck/shfmt lint-sh coverage, topology sweep with a
   fired positive control (Axis 7).
7. CI's raw job log count and the #1342 block (Axis 5, my own independent leg).

**Relayed, weighed rather than re-run:** the author's controlled-baseline PR comment
(sorted-multiset diff, 9-line delta fully attributed, self-controlled instrument) — I
did not re-run the base-side suite myself; I judged the comment's own controls as
sufficient given CI's raw log independently agrees on the headline number (2979/0).

**Unchecked:** no bench/rig/real-docker path touched by this PR (it is a pure text move,
nothing behavioural changes), so nothing was owed there.

## Not a finding, stated for completeness

The "ILLEGAL" wording for the adjacent-merge case (Axis 1) overstates a lane convention
as a budget-gate hard rule. Cosmetic; not blocking, not worth a re-push.

VijitSingh97:
## Author correction: "illegal" in the file-count section overstates the rule

The gate pass flagged this as non-blocking and said it wasn't worth a body edit. I agree it isn't
worth a re-push — the proofs are unaffected — but the claim shouldn't stand unmarked, because it is
the load-bearing sentence under "two files is the forced minimum."

**The body says** the one adjacent merge (`17` + `18` = 568 lines) is over the 400 floor and
therefore **"NO — illegal"**, concluding that **two files is FORCED**.

**That overstates a convention as a hard rule, and my own lane supplies the counter-example.** A
file over 400 is permitted with a budget row: **84 budgeted rows sit over 400**, and the decisive
one is in this very directory — **`lib/pithead/12-firstboot-wizard.sh` is 556 with a row**, which
*this same split* set at P3.

**The accurate version:** merging would be *permitted* but would **cost a new tsv row**, so two
files is **the minimum that avoids a row** — a well-motivated choice, not a forced minimum.

**Why this is worth correcting rather than shrugging off:** the false framing was hiding the actual
argument. A new row is legitimate precisely when a function is **indivisible** — P3 spent one
because `firstboot_wizard` is 521 lines and cannot be split without a semantic refactor. Here both
halves divide cleanly, so nothing justifies spending a row. That reasoning is stronger than
"illegal" ever was, and it is the reason the cut is two files.

The rest of that section stands as written, and the reviewer verified it independently: both
declined 3-file cut points (region-relative 270 and 556) are legal blank-line boundaries, and
declining them on naming rather than necessity is — correctly — represented as a choice.

Retracted in this lane's role file so the framing does not propagate into P6→P13 bodies.

Issue #908, opened 2026-08-13
Appliance backup export from the dashboard: encrypted archive + one-time emergency kit


VijitSingh97:
Part of #786 (sub-issue A of the parity decomposition; decision record in the 2026-08-13 comment there).

## Goal

An appliance operator can export an encrypted backup from the dashboard. Today nothing exports state off a shell-less box — a dead appliance takes wallet settings, Tor identities and the dashboard DB with it. This is the only *irreversible* loss in the parity table, which is why it leads the pre-GA set.

## Mechanism

- New fixed control verb `backup` on the host runner, beside the existing seven (`control_process_request` dispatch, `pithead` ~9224). The host side runs the existing `stack_backup` machinery (~2793–2970) unchanged: `pithead-backup-<stamp>.tar.gz.enc`, AES-256-CBC, PBKDF2 600k iterations, streamed via openssl.
- **Encrypted only.** The verb refuses a passphrase-less request outright; the CLI's `--no-encrypt` stays CLI-only.
- **Passphrase model (decided): the emergency kit.** The host auto-generates a strong passphrase per export and the dashboard shows it exactly once, inside a downloadable kit (passphrase + archive name + creation time + what the archive contains). Nothing is retained on the box after the exchange. The UI moment is a genuine "download this now — it cannot be recovered" screen, not a passphrase input field. Precedent: Home Assistant's backup emergency kit, built for exactly our user; typed-per-export passphrases are what sysadmin-facing peers do (OPNsense), and umbrel's standing typed password is the cautionary tale (reused, forgotten).
- The archive lands in a mount the container can serve for download; download + kit delivery are the same dashboard flow.

## Coverage

Tier 1 for the verb/refusal logic and the kit lifecycle (tests/stack + pytest, same pattern as the existing verbs); one tier-4 release-checklist item for a real download from a provisioned appliance.

## Dependencies

None. The archive format already exists. Scheduled/recurring backups are out of scope here (they need a standing key — a later issue can adopt the HA stored-key model if wanted).

VijitSingh97:
Shipped as PR #934: dashboard backup verb on the host runner, encrypted-only, one-time emergency kit (HAOS model per the #786 decision record) with a TTL'd passphrase handoff + a stranded-kit redaction backstop. KVM/live proof rides the next battery + bench round.

PR #1498, opened 2026-08-29
refactor(tests): split the control-ops domains out of run.sh (#1105)

VijitSingh97:
R8 of the #1105 Phase 1 test split: moves the two control-ops domains out of `tests/stack/run.sh`.

| new file | domain | lines |
|---|---|---|
| `tests/stack/test-control-telegram.sh` | Telegram lifecycle verbs (#338) | 67 |
| `tests/stack/test-control-backup.sh` | the backup verb sections (#908) | 175 |

`run.sh` 1406 → 1229. Both new files are under the 400-line budget target, so neither takes a tsv row; only `run.sh`'s row lowers.

## These two domains are self-contained — the opposite of R7's shape

R7 (`test-secrets-masking.sh`) had to argue that its domain must **not** self-arm, because it was a pure consumer of the shared control sandbox and was position-locked. **These two are not that shape, and I checked rather than assumed:**

- Neither block calls `build_control_sandbox()`. `run.sh` calls it exactly once, and that call stays behind.
- Neither block reads `$C`, `$CTRL_LOG`, `$RESULTS`, `$REQS`, `$STAGED`, `$MASKED` or `$AUDIT` (comments stripped). Each builds its own directory under `$SANDBOX` — `$CC` for #338, `$BKC` for #908 — both assigned inside the moved text.
- Every redirection in both files targets a path under that domain's own directory, and each domain unsets the environment it exports before it ends.

So each file is standalone-sourceable once `lib.sh` is sourced. **Proven, not asserted:** sourcing `lib.sh` and then the domain file alone under `set -u` runs each domain green (telegram 7 ✓ / 0 ✗, backup 26 ✓ / 0 ✗). The negative control fires — with `SANDBOX` unset the guard aborts at its own line with `SANDBOX: parameter null or not set` and the body is never reached. (The message is quoted rather than the exit code: `rc=127` does not discriminate `set -u` from command-not-found on this box.)

## The cut range contained a stanza — this row hit the standing trap

The map's contiguous range for this row runs through `test-wizard-setup.sh`'s `source "$HERE/..."` stanza. Cutting that would nest the wizard-setup domain inside the new file and leave the extracted content with **zero** #1400 zero-assertion coverage — and all five standard instruments are blind to it.

The range was ended one line short of that stanza instead. That keeps it contiguous, leaves the stanza in place in `run.sh`, and needs no reorder. The arithmetic reconciles against the map: 182 content lines + the blank + the 3-line stanza block = the map's 186.

The backup section's own leading `# ---` divider travelled with it; the divider introducing the deploy stanza stayed.

## Identity and order

Byte-identity rebuilt **from git** (`git show <base>:tests/stack/run.sh`), not from the extraction, in both directions, with firing controls:

- telegram body vs base range — identical, 35 non-blank lines
- backup body vs base range — identical, 137 non-blank lines
- expected `run.sh` (base minus the range, plus hand-written stanza text) vs actual — identical, 1229 lines both

Three controls each provably made a diff fire: an off-by-one range, a content mutation, and an injected line. An empty diff from a broken command looks identical to a proof, so each was checked.

Both stanzas sit at the block's exact former position, so execution order is unchanged.

## Budget

`--generate` also wanted to lower three rows this diff never touches — `dashboard/mining_dashboard/service/egress.py` (424→423), `.github/workflows/ci.yml` (513→512) and `tests/integration/run.sh` (2551→2520). All three were hand-restored: lowering a row this diff did not touch reds every un-rebased branch in every lane, naming a file their diff never touched. The committed tsv changes exactly one row.

Gate passes, and its own firing control passes: deleting `run.sh`'s row makes the gate fail **by name** (`tests/stack/run.sh is 1229 lines, over the hard ceiling of 800 for a new/unbudgeted file`), and the tsv was then restored and re-verified byte-identical against `git show HEAD:` rather than against `git status`. Both the gate and its control were **re-run after the rebase**, against base `9b593f8`; `origin/develop-v2` was probed before and after and was unmoved both times. Monotonic proof is local-only (#1452).

The rebase also dissolved a question this branch used to carry: unrebased, it held the old `lib/pithead/99-remainder.sh` ceiling of 8517 against a base that had lowered it to 7948, which would have read as a raise on a row this diff never touched. Rebased, the tsv diff against the base is exactly one row: `tests/stack/run.sh` 1406 -> 1229.

`shfmt -i 4 -d` and `shellcheck --severity=warning` clean on both new files, and `shellcheck --severity=warning tests/stack/run.sh` alone stays rc 0.

## Suite proof

Full `tests/stack/run.sh` pair, both runs green, both rc 0:

| | baseline | after |
|---|---|---|
| verdict | `pithead tests: 2979 passed, 0 failed` | `pithead tests: 2979 passed, 0 failed` |
| output lines | 3292 | 3292 |

- **Same set:** sorted multiset of section headers — identical.
- **Same order:** the ordered comparison (the only instrument that can see a permutation) — identical, and **0 of 274 positions differ**. The comparator's own permutation control fires on the ordered diff while the sorted diff stays silent on the same input, which is what proves the instrument discriminates rather than merely agreeing.
- **Whole output**, ANSI stripped and worktree paths normalised, sorted-multiset diff: **exactly two differing pairs**, both from the documented benign classes — one ed25519 keygen fingerprint (#1325) and one `mktemp` path. An injected-line control raises that count, so the diff is live rather than vacuously empty.

Both new files also run **green standalone** — `lib.sh` plus the domain file alone under `set -u`: telegram 7 ✓ / 0 ✗, backup 26 ✓ / 0 ✗.

**Correction, from the adversarial review:** the telegram figure was first written as 4 and is 7 — 6 `assert_eq` plus one `assert_contains`, confirmed by call-site count and by running the file. The reviewer caught it and I re-derived it independently rather than relaying it. The claim it supports is unchanged (the file is standalone-sourceable and green either way), but a wrong number inside the one paragraph whose purpose is "checked, not asserted" is worth correcting in place.

### The pair was taken at `a3233d8`; this branch is rebased onto `9b593f8`. Here is why the proof carries — as commands, not as a conclusion

`develop-v2` moved under this work mid-proof (#1496, the backup/restore + help/version remainder cut). I rebased rather than merged, and re-derived the insulation instead of assuming it. **"`tests/stack/` is byte-unchanged" is the wrong test** — the suite reads outside `tests/stack/`, most importantly the built `pithead` artifact it sources via `$STACK`. So the check is what the suite actually *reads*:

```
git diff --name-status a3233d8 9b593f8
#   M docs/dev/file-budget.tsv
#   A lib/pithead/17-backup-restore.sh
#   A lib/pithead/18-help-version.sh
#   M lib/pithead/99-remainder.sh

git rev-parse a3233d8:pithead 9b593f8:pithead     # d79917f… both — the artifact $STACK sources
git rev-parse a3233d8:tests   9b593f8:tests       # e1a2760… both — the whole tests/ tree object
grep -rn 'lib/pithead\|file-budget' tests/        # every hit is comment text, not a read
```

The artifact is the same blob and `tests/` is the same tree object at both bases, so nothing the suite executes changed. That is why the pair was not re-run. **This is an attack axis, deliberately: if any one of those three commands disagrees at review time, the pair should be re-run rather than argued.**

The proven content itself survived the rebase byte-identical, blob-for-blob — `tests/stack/run.sh`, `test-control-backup.sh` and `test-control-telegram.sh` each carry the same object id before and after — checked with a control on `lib/pithead/99-remainder.sh`, a file the base move genuinely changed, which correctly reported DIFFERS.

## The two carry-overs owed to this header

1. **Tool-citation disclaimer — folded in verbatim.** Both headers carry it.
2. **The `:?` guard — adapted, and that is a judgement call worth reviewing.** The ruling prescribed `: "${C:?}" "${WALLET:?}"`. Neither file reads `$C` or `$WALLET`, so that line would assert names these files never use — a guard that cannot fire is worse than none. It is `: "${SANDBOX:?}"` here: the one ambient name either file actually reads.

**A side effect worth flagging: this guard degrades the lane's own audit tool.** With the guard present, the tool annotates `SANDBOX -> PROVIDER (has ${..:-} default somewhere)`. It has no default — the tool's regex counts a `${..:?}` expansion as a defaulted one. Confirmed by stripping the one guard line and re-auditing, and the underlying claim re-derived from source: `SANDBOX` is assigned once in `lib.sh` at column 1 with no default, never in `run.sh`. That annotation is the same signature as R7's `$WALLET` defect, so a future cut could read a real finding as expected noise.

## Docs

No doc change is owed. `docs/dev/repo-map.md` describes the split generically and stays true. `docs/dev/testing-strategy.md`'s tier-1 coverage table pointing at `run.sh` is systematic and pre-existing across ~24 already-split domains — filed as #1445 rather than widened into this diff.

## One transitive dependency, disclosed

Both domains drive the control channel through `lib.sh`'s `run_sourced()`, which reads `$STACK`. So each file depends on `$STACK` **transitively through a lib.sh function** — a dependency no `$VAR` sweep of the file itself can see, and the class this lane has shipped a defect on before.

It does not weaken the claim, and I checked rather than assumed: `STACK` is assigned at `lib.sh:14`, column 1, above the first function definition, and no domain file assigns it. So it is lib.sh top-level state, which is exactly what "standalone-sourceable once lib.sh is sourced" already requires. The headers scope their `$SANDBOX` statement to names the file itself reads, which is what the audit measures; this paragraph is the wider claim.

## What I did NOT do

- No mutation testing of the moved assertions. This is a move: the assertions are byte-identical to what already shipped, and mutation coverage of them is not this cut's debt.
- The two suite runs overlapped each other (N=2). Concurrency is proven on this box at N=2 only; N=3 is known to redden the #1342 block (#1495).
- I did not re-derive whether `test-wizard-setup.sh`'s stanza is correctly placed for its own sake — only that this cut leaves it intact and in position.


VijitSingh97:
**R8 gate pass — verdict: MERGE. Zero blocking findings.**

Non-author adversarial review at head `af756c4` against base `9b593f8` (base re-derived by the reviewer, probed unmoved start and end of review). Every load-bearing claim was re-derived, not relayed, each with a positive/firing control:

- **Byte-identity (axis 1):** both moved bodies rebuilt from the raw diff and diffed against the 181 deleted lines — identical; injected-line control fired. Non-blank counts 35/137 and 1406→1229 confirmed by independent arithmetic.
- **Self-containment (axis 0):** comment-stripped sweep for `build_control_sandbox` and all seven shared control vars — zero hits in both files, control fired. Every exact-count assertion still in run.sh sits before the cut point or counts an unrelated variable.
- **Blob-equivalence across the base move (axis 10):** all three PR-body commands re-derived (`pithead` same blob, `tests/` same tree across a3233d8→9b593f8); no path found by which #1496's `lib/pithead/` changes reach the suite except through the built artifact, and `make lint-pithead-parity` passes at head. The reviewer then made the question moot by **re-running the full suite pair itself** (fresh worktree at base vs head, N=2 concurrent): both `2979 passed, 0 failed`, 274/274 section headers in identical order, whole-output multiset diff empty except the documented #1325 ed25519 class, injected-line control live. Re-run owed before merge: **no** — it was done.
- **Rebase gap (axis 8):** per-file blob identity 03cb930→HEAD on all three touched files, with `lib/pithead/99-remainder.sh` as the DIFFERS control; reflog shows no amend between commit and rebase (and does show amends for the sibling PR, proving the instrument sees them).
- **Carry-overs (axis 9):** disclaimer wording matches the ruling recovered independently from PR #1494's own comments; the `: "${SANDBOX:?}"` adaptation ruled correct and complete (a full comment-stripped `$VAR` sweep found `SANDBOX` as the only ambient name in either file), and rightly decided in-lane because it was disclosed as its own review axis.

**One should-fix, discharged before merge:** the body's "telegram 4 ✓" was wrong — the file has 7 assertion call sites and the reviewer's two clean standalone runs both report 7 ✓ / 0 ✗. Corrected to 7 in the body (both sites, readback-verified). The underlying standalone-sourceable claim was never in doubt.

**Declared unchecked by the reviewer:** the author's original proof-run log mtime (no artifact referenced; substituted reflog + blob identity) and the lane-local audit tool's annotation behaviour (tool not in this repo, as the header itself states).

Merging pinned to the reviewed head `af756c4`.

Issue #1497, opened 2026-08-29
Remote-node mode: a dead ZMQ endpoint is invisible end-to-end, so the remote scenarios can pass while p2pool starves


VijitSingh97:
### Summary

Three separate gaps compose into one failure: **a remote Monero node whose ZMQ publisher is dead
or silent will pass every check the stack and the harness make, bring the stack up, and starve
p2pool of block notifications with nothing able to notice.**

Filing this as one issue rather than three deliberately — split up, each finding reads as a minor
scoping nit. Composed, they are a hole in the one mode #1083 already reports as having zero live
coverage.

### 1. The reachability preflight can be satisfied by something that isn't listening

`preflight_remote_nodes` dials the remote node's RPC and ZMQ ports with bash's `/dev/tcp` and
refuses setup if either is shut. That is a TCP connect, and a TCP connect does not prove anything
is behind it.

Measured, with the preflight's exact gesture: against a **containerised** node with a published
port and *nothing listening inside the container*, the dial returns **rc=0**. Docker's userland
proxy — enabled by default — binds the published host port and accepts the connection itself, so
the dial succeeds whether or not the container-side socket exists.

This matters specifically because a remote node is a natural thing to run as a container. In that
shape the preflight cannot distinguish "ZMQ publisher up" from "ZMQ publisher failed to bind".

### 2. The preflight defers to healthchecks that do not cover it

The preflight is honest about its own scope — its comment says it proves reachability, "not
protocol health, which the stack's own healthchecks own." Following that deferral:

- `build/p2pool/healthcheck.sh` is `timeout 5 bash -c '</dev/tcp/127.0.0.1/3333'` — a TCP connect
  to **p2pool's own stratum port**. It proves p2pool is accepting miners. It says nothing about
  whether p2pool is receiving block notifications from monerod.
- `git grep -niE 'zmq' -- tests/integration/` returns **zero hits**. (Positive control: the same
  grep for `sidechain` over the same paths returns ~20, so the search works.) The integration
  harness asserts nothing about ZMQ anywhere.

So the deferral lands nowhere. Each component states its scope accurately; the requirement falls
through the gap between them.

### 3. Why this becomes urgent now

Standing up a remote node for the remote scenarios makes the natural cheap choice an **offline**
node serving a static chain — and an offline monerod satisfies the harness's readiness gate. Its
`get_info` reports `status=OK`, `synchronized=true`, `target_height=0`, which passes
[...]

PR #1499, opened 2026-08-29
test(e2e): assert the Monero node actually publishes ZMQ block notifications (#1497)

VijitSingh97:
Closes nothing on its own — this is the harness half of #1497, filed against #1083/#1446.

## The hole

`git grep -niE zmq -- tests/integration/` returned **zero** at head (positive control: `sidechain` returns 22). Three separately verified findings compose into a silent failure:

1. **`monero_caught_up` cannot fail for a node that can never publish.** An `--offline` monerod reports `status: OK` with `target_height: 0` — measured on the stack's exact image, pinned by ID — so *both* disjuncts of the predicate pass.
2. **The installer preflight's reachability check is a bare TCP dial**, and a dial succeeds against a docker-published port with nothing behind it: `docker-proxy` accepts the connection itself.
3. **Nothing downstream covers it.** `build/p2pool/healthcheck.sh` is a TCP connect to p2pool's *own* stratum port. The preflight honestly defers to "the stack's own healthchecks", and the deferral lands nowhere.

Composed: a ZMQ-dead node passes the preflight, passes the sync assertion, brings the stack up green, and starves p2pool with nothing able to notice. That is a SKIPPED→PASSED with no failure mode — a silencing, which is worse than the skip, because a skip at least announces itself.

## What this adds

A ZMTP 3.x greeting plus NULL-mechanism READY exchange, which an `accept()` cannot satisfy, wired into `run.sh` as step 4b beside the existing sync assertion.

## What was RUN

**The discrimination proof — one host, three targets, the preflight's exact gesture (`timeout 5 bash -c "</dev/tcp/host/port"`) beside the probe:**

| target | preflight dial | `zmq_pub_probe` |
|---|---|---|
| live monerod ZMQ `:18083` | rc=0 | `ok … Socket-Type XPUB` |
| **published-but-dead `:28099`** | **rc=0** | **`no-greeting` (fail)** |
| closed/unpublished `:28098` | rc=1 | `connect-refused` (fail) |

The middle row is the whole point: the dial passes, the probe reds. The dead-port control was a container publishing a port with nothing listening inside it, torn down afterwards.

**Measured, not assumed:** real nodes advertise `XPUB`, not `PUB` (Monero's `zmq_pub` is an XPUB so it can observe subscriptions). Asserting `PUB` alone would have been false-red against every live node. Both are accepted; a SUB/REQ/REP peer is not.

**Mutation battery — four mutations of the probe, each killed by the case built for it,** every mutation asserted to have applied before the run:

| mutation | result |
|---|---|
| accept a silent peer | ✗ `silent peer is named no-greeting`, ✗ `published-but-dead port…` |
| substring-match `Socket-Type` instead of walking metadata | ✗ `a decoy Socket-Type VALUE does not win` |
| drop the signature check | ✗ `non-ZMTP listener is refused` (+1) |
| accept any socket type | ✗ `a ZMTP peer that is not a publisher is refused` (+1) |

Control restored: **33 passed, 0 failed.**

**Suite:** all 12 integration self-tests green, 565 cases, zero failures. The new file needs no registration — `Makefile:29` globs `tests/integration/selftest*.sh`.

**Lint:** `shellcheck 0.11.0 -x --severity=warning` clean; `shfmt -i 4 -d` clean over the Makefile's exact glob (it caught `v=$(…); rc=$?` on one line — the class that reddened #1444); `lint-file-budget`, `lint-md`, `lint-docs-voice`, `lint-operator-strings` all pass. `make lint-sh` was **not** run as a whole — it is this box's OOM path and is serialized; the targeted shellcheck plus the full shfmt glob is the documented substitute.

## Design notes worth reviewing

- **Split for testability.** The socket I/O runs on the target through `rx` (the harness ships shell snippets, never files — there is no file-shipping mechanism) and returns hex; every verdict above it is a **pure function of that hex**. The self-test stubs nothing.
- **Bash, not Python.** `rx` already guarantees bash on the target; python3 on an appliance is not a dependency this harness may assume.
- **Placement.** Cases live in a new `selftest-zmq-probe.sh` because `selftest.sh` sits exactly on its recorded budget ceiling and ceilings only go down. `run.sh` is at 2542 against a 2551 ceiling.

## What this does NOT do — stated rather than hidden

- **Tier B is not built.** Asserting an *observed* block notification requires a MOVING chain tip, and a height comparison cannot discriminate against a STATIC node — a frozen p2pool view and a frozen monerod agree forever. That is recorded on #1497, not silently omitted.
- **The preflight itself is unchanged.** The dial's blindness is real, but `lib/pithead/99-remainder.sh` is mid-cut under pure-move discipline and a behavioural change there would poison its byte-identity proofs. That fix is deliberately not on this branch.
- **The remote-mode path is wired but was exercised only in local mode.** The live positive control was a local-mode stack; the remote branch selects `$REMOTE_MONERO_HOST` / `${REMOTE_MONERO_ZMQ_PORT:-18083}` and has not yet met a real remote node.

## Shared paths

Touches `docs/dev/integration-testing.md` (`docs/` is shared, no lane owns it), per the standing rule that a change ships with its docs in the house voice. No other shared file is touched.

## Over-engineering pass (findings and what was done)

- `stdlib` — `zmq_hex_ascii` was a 9-line byte loop reinventing one `sed`+`printf` line. **Cut** (`72603f7`); identical output on every fixture including the empty string, and the 10 cases that exercise it all red when the replacement is broken.
- `delete` — the `last seen [$want]` diagnostic named an arbitrary property when there was more than one. **Cut.**
- `yagni` — the LONG-frame branch in `zmq_ready_socket_type` and its counterpart in the target snippet (9 lines). No libzmq peer sends a long READY, so it reads speculative. **Kept, deliberately:** without it a legal peer that did send one would parse as `malformed-ready`, and a false RED in the release gate is the exact failure this lane exists to prevent.

Net: **9 lines removed.** Re-verified after the cut — five mutations each killed, control 33/0, all 12 self-tests green, live probe still `ok XPUB`, shellcheck and shfmt clean.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -317,6 +317,17 @@ and `--list` prints it).
 - `pithead status` exit code: `0` for a healthy config.
 - Dashboard reads live state. `/api/state` is reachable; Monero is synced (`done`); pruned/full
   display matches `monero.prune` ([#32](https://github.com/p2pool-starter-stack/pithead/issues/32)); the sidechain `pool.type` matches `p2pool.pool`.
+- The Monero node publishes block notifications. A ZMTP handshake against the node's ZMQ port
+  succeeds and the peer advertises a publisher socket type ([#1497](https://github.com/p2pool-starter-stack/pithead/issues/1497)). This is a separate
+  assertion from "Monero is synced" because the sync check cannot fail for a node that can never
+  publish: an `--offline` monerod reports `status: OK` with `target_height: 0`, which satisfies
+  both halves of the caught-up predicate. Nothing downstream covered the gap either — p2pool's
+  healthcheck is a TCP connect to its own stratum port — so a ZMQ-dead node brought the stack up
+  green and starved p2pool with nothing able to notice. A bare TCP dial does not close it: a
+  docker-published port with nothing behind it accepts the connection in `docker-proxy` itself and
+  answers a dial with rc 0. Measured on one host, three targets: the live node and the
+  published-but-dead port were indistinguishable to a dial (both rc 0) and separated by the
+  handshake. Real nodes advertise `XPUB` rather than `PUB`; both are accepted.
 - End-to-end mining. Workers are online (`proxy_workers >= --workers`), stratum has connections,
   and total hashes are accumulating ([#28](https://github.com/p2pool-starter-stack/pithead/issues/28)).
 - Posture propagated. `MONERO_RPC_BIND`, `DASHBOARD_SECURE`, `XVB_ENABLED`, and `TARI_REQUIRED`
@@ -610,6 +621,9 @@ Several self-tests sit beside it as standalone files, picked up by the same glob
 `selftest-skip-accounting.sh` is the one that keeps the skip accounting honest: besides checking
 the counters and the real `summary()`, it censuses every harness file and fails if a skip leaves
 through a bare `it_warn` — by wording, and by shape for the drops that never say "skipping".
+`selftest-zmq-probe.sh` drives the ZMTP verdicts from captured and hand-built wire fixtures, so
+every failure class — a silent peer, a non-ZMQ listener, a ZMTP peer that is not a publisher, a
+READY frame carrying a decoy `Socket-Type` value — is reachable with no socket and no stack.
 
 ---
 
--- a/tests/integration/run.sh
+++ b/tests/integration/run.sh
@@ -31,6 +31,8 @@ source "$HERE/rig-key-ledger.sh" # must precede any module that marks a write (#
 source "$HERE/rigforge-writable-keys.sh"
 # shellcheck source=tests/integration/rigforge-upgrade.sh
 source "$HERE/rigforge-upgrade.sh"
+# shellcheck source=tests/integration/zmq-probe.sh
+source "$HERE/zmq-probe.sh"
 
 # --- Defaults / globals -----------------------------------------------------
 IT_MODE="ssh"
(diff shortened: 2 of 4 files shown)


VijitSingh97:
### Adversarial pass — verdict, posted late and weaker than it should have been

**Disclosure first, because it changes how much this is worth: this pass ran in the same lane
that authored the PR, and it ran AFTER the merge.** A same-lane post-merge pass is not an
independent review. I am pasting it because the standing bar asks for a durable verdict on
every PR, and because a late verdict is worth more than a missing one — not because it
retroactively makes this reviewed. An independent pass is still welcome and I would not argue
with anything it finds.

**Verdict: merge stands. One nit, failing in the safe direction. No correctness defect found.**

What I checked, by re-deriving rather than re-reading:

- **ZMTP greeting offsets.** Signature bytes 0 and 9 (`ff` … `7f`), version major/minor at 10
  and 11, mechanism at bytes 12–31. The code reads the mechanism as `${h:24:40}` — hex offset
  24 is byte 12, and 40 hex chars is 20 bytes. Correct.
- **READY frame framing.** Short frame: size at byte 1, body from byte 2. Long frame
  (`flags & 0x02`): 8-byte size, body from byte 9. Both correct.
- **The property walk.** `p` advances `1+klen` for the key and `4+vlen` for the value, matching
  ZMTP metadata layout, and the `socket-type` compare is case-insensitive per RFC 23. Walking
  the metadata rather than substring-matching it is what stops a `Socket-Type` appearing inside
  another property's *value* from being read as the real one — that is the right call and it is
  worth keeping.
- **Termination.** A degenerate all-zero-length property still advances `p` by 5 per iteration,
  so the loop cannot spin.
- **`v=$(...) || { ... }`.** Correct here — the assignment's status *is* the command
  substitution's status. This is adjacent to the `v=$(...); rc=$?` class that reddened an
  earlier PR, so it is worth stating explicitly that it was checked and is not that bug.

**The nit.** In `zmq_ready_socket_type`, the up-front guard bounds the *body* length but not an
individual property's `klen`. A malformed READY whose `klen` pushes `p` past the body makes
`vlen=$((16#))` raise a bash arithmetic error instead of returning a clean verdict. It is noisy
rather than wrong: the probe still fails, never passes, so it cannot produce a false green.
Not worth a re-run of an 11-minute CI cycle on its own — fix it when that file is next touched.

### New evidence: the merged probe re-validated against a node it was never developed against

Run today against a fresh standalone monerod, four targets:

| target | verdict | rc |
|---|---|---|
| live ZMQ port, bridge A | `ok … advertises Socket-Type XPUB` | 0 |
| live ZMQ port, bridge B | `ok … advertises Socket-Type XPUB` | 0 |
| **monerod's own RPC port** | `no-greeting` | 1 |
| closed, unpublished port | `connect-refused` | 1 |

The third row is a **case the original discrimination proof did not have**: a live, healthy,
*non-ZMQ* listener — a real HTTP server, not a dead docker-proxy port. The probe rejects it
correctly. That is the strongest single piece of evidence here, because it was not a target the
probe was tuned against.

It also demonstrates the hazard the probe exists for, on the same node: a **static** monerod's
ZMQ publisher passes this tier-A assertion — correctly bound, advertising XPUB — while being
permanently incapable of publishing a block notification. Tier A proves the socket is a
publisher. It cannot prove the publisher will ever speak, which is exactly why tier B is
recorded on #1497 as unbuildable on a static node rather than quietly omitted.

VijitSingh97:
**Independent post-merge review — verdict: PASS (merge stands).**

This is the independent pass the earlier same-lane comment recommended; it worked in a fresh worktree at the merged tip and re-derived every claim without reading that comment as evidence.

- **No false-green class found:** a real ZMTP responder (XPUB READY) passes; `nc -l` (accept-then-silence, the docker-proxy shape) rejects `no-greeting` rc=1; a live HTTP listener rejects `short-greeting` rc=1. Positive control fired. Note for future readers: a live HTTP listener takes the `short-greeting` branch, not the self-test fixture's `bad-signature` branch — same outcome, different path.
- **No false-red regression:** `"ok PUB" | "ok XPUB"` both accepted (zmq-probe.sh:172), confirmed live; SUB correctly rejected as `socket-type-mismatch`. The issue body's PUB-only suggestion did not creep back in.
- **Self-test re-run from scratch:** 33 passed / 0 failed. The TLV walk is real — the decoy Socket-Type-value case passes for the right reason, not by substring luck.
- **Disclosed nit confirmed** noisy-never-false-green: the malformed-READY arithmetic error exits the function rc=1 with empty stdout, never a false verdict. One-line bounds guard suffices.
- **One new finding, filed separately:** the initial connect is unbounded — `timeout` only governs the reads, so a filtered/black-holed remote host blocks to the kernel SYN-retry budget, well past the probe's intended limit, on exactly the remote-mode path the PR discloses as unexercised. See #1500.
- **Docs vs code:** consistent (integration-testing.md:320-327, 624-626).

Declared unchecked by the reviewer: the remote-mode path against real hardware (same gap the PR discloses), full run.sh, lint-sh.

PR #1501, opened 2026-08-29
docs(e2e): correct the retired ~95 GiB pruned-chain figure and the patched-LMDB claim (#1446, #1489)

VijitSingh97:
## What this fixes

Two claims stated as measurements across five tracked files were false. Both were found while discharging the hold condition on #1489.

### 1. The `~95 GiB` pruned-chain figure — and it was wired to an operator ACTION

The bench's pruned Monero chain measures **258 GiB and is dense**. Measured, not estimated:

- `mdb_stat -ef` on an idle copy: **`Free pages: 10`** — ten, out of 67,605,667.
- `67,605,667 × 4096 = 276,912,812,032` — exactly the file size, so that figure is the high-water mark, not a bloated envelope.
- Correctly pruned: `pruning_seed = 384` on both the output and the source, and the running monerod's rendered `bitmonero.conf` carries `prune-blockchain=1`.

The docs said the chain held "~95 GiB of live data" inside a "~270 GiB" file. That was a retired expectation, never a measurement — and it was not merely stale. Both READMEs attached an imperative to it:

> If it ever reads back up near ~250 GiB, that is LMDB free-page bloat from an in-place prune, not a full chain — compact it.

The chain *does* read 258 GiB, and it *is* dense. That rule sends a reader into a ~2.5-hour rebuild that reclaims nothing. `tests/integration/system-info.sh` printed the same false diagnosis into **every system snapshot** it generated.

Replaced with the measured figure and the rule that actually discriminates: **read the freelist before concluding a chain is bloated.**

### 2. "Monero ships a patched LMDB"

Three files explained `MDB_VERSION_MISMATCH` as a patched on-disk format. Measured: it is the **lock-file** format, and it appears only while monerod holds the environment — a stock LMDB tool opens an idle copy of the same chain, and both DBs are magic `0xbeefc0de` version 1. It is not corruption and not a reason to stop monerod.

## What was checked and deliberately left alone

- **The general mechanism is TRUE and is kept everywhere**: LMDB never shrinks its file, so an in-place prune *can* leave reclaimable free pages. Only the fabricated magnitude and the unconditional imperative are gone.
- The #1489 copy-then-swap block in `compact-chain.sh`'s header, and `build-pruned-chain.sh`'s "full chain never modified", are true as scoped — checked line by line, untouched.
- **`docs/hardware.md` and `docs/getting-started.md` are NOT touched.** They promise ~100 GB pruned / ~270 GB full for a *general* Monero node. That is a different subject, and I have one bench measurement and no upstream one. See the open question below.

## Also in this diff

- **`docs/dev/test-server-architecture.md`'s sizing table under-budgeted the bench disk by ~160 GiB** (~280 GB total → a measured ~440 GiB), because the `~95 GB` figure was a row in it. Recomputed: 258 + 132 + 20 + 30 = 440 GiB; a 1 TB NVMe (~931 GiB usable) leaves ~490 GiB spare, not ~650 GB.
- The full-chain figure is **not** measured on this bench and is marked `[TODO: verify upstream — ...]` rather than guessed, per `docs/dev/STYLE.md`.
- Ported one true hazard note the deployed copy carried and the repo copy lacked: the compaction read is a real LMDB reader on the live DB, so a long read transaction pins freed pages and **the live `data.mdb` can grow while the copy runs**. Watch `df`.

## Deployed copies

`~/pithead-testbench/{compact-chain.sh,README.md}` carried both false claims and a repo-scoped sweep cannot see them (the trap this lane paid for in cycle 25). They have been patched **in place** — the deployed copies have diverged from the repo and carry content the repo lacks, so syncing over them would have destroyed it.

## What was RUN

- `make lint-docs-voice` — OK, 40 prose docs.
- `make lint-md` — 47 files, 0 errors.
- `make lint-topology`, `make lint-file-budget` — OK, both self-tests green.
- `shellcheck` (0.11.0) on both changed scripts — rc=0, only pre-existing SC2016 info notes matching the files' existing idiom.
- `bash -n` on all three changed scripts, repo and deployed.

**Not run:** the full `make lint` / `make test`. This diff is comments, prose and one `printf`; no executable path changed. `make lint-sh` is the documented OOM path on this box and is serialized fleet-wide, so I ran shellcheck on the changed files directly instead and am saying so rather than implying full-suite coverage.

The 258 GiB / freelist-10 figures were **re-derived from the primary record** (`pruned-chain-verdict.md`), not carried over from a summary.

## Shared paths

Touches `docs/dev/` (in `_shared.paths`, no override needed) alongside this lane's own `tests/integration/`.

## Open question this surfaced, not resolved

Our own user docs promise **~100 GB** for a pruned Monero node; this bench measures a correctly-pruned, dense **258 GiB** chain. Both cannot be right, and it is a 2.5× gap on a number people buy disks against. I did not touch the user docs — one bench measurement is not an upstream fact. Routed to the controller.

Refs #1446, #1489.

### DIFF
--- a/docs/dev/release-server.md
+++ b/docs/dev/release-server.md
@@ -351,8 +351,9 @@ tests/integration/run.sh --host you@server --dir pithead --readiness
 >
 > `MDB_VERSION_MISMATCH` from a stock LMDB tool is the **lock-file** format, not a patched on-disk
 > format: it appears while monerod holds the environment, and the same tool opens an idle copy of
-> the same chain (both DBs are magic `0xbeefc0de`, version 1). It is not corruption — do not stop
-> monerod over it. Often the best move is to leave the free pages alone.
+> the same chain. Measured here on monerod 0.18.5.1, where both DBs read magic `0xbeefc0de`,
+> version 1 (#1446) — one bench, one build, so treat the reading as ours rather than universal.
+> It is not corruption — do not stop monerod over it. Often the best move is to leave the free pages alone.
 
 ## Bench allocation and the rig lock
 
--- a/docs/dev/release-server.md
+++ b/docs/dev/release-server.md
@@ -9,8 +9,10 @@ GitHub Actions runs on every PR, and how to harden the server. Operational compa
 
 GitHub-hosted runners can't do the real-chain tier. On a public repo the hosted Ubuntu runners
 are free (4 vCPU / 16 GiB RAM), but they are ephemeral: a fresh VM per job, ~14 GiB of free
-disk, and a 6-hour job ceiling. A Monero chain is ~95 GiB pruned / ~270 GiB full and takes days
-to sync; Tari adds ~50 GiB. There is nowhere to keep that synced state between runs, and no time
+disk, and a 6-hour job ceiling. A pruned Monero chain measures ~258 GiB on our own bench
+([#1446](https://github.com/p2pool-starter-stack/pithead/issues/1446)) and an unpruned one is
+larger still [TODO: verify upstream — no full-chain measurement exists on this bench]; either
+takes days to sync, and Tari adds ~50 GiB. There is nowhere to keep that synced state between runs, and no time
 to sync it inside a job. So the real-daemon, real merge-mining tier (tier 4) is not possible on
 hosted runners. That's the reason a dedicated, already-synced server exists
 ([#54](https://github.com/p2pool-starter-stack/pithead/issues/54)).
@@ -332,8 +334,11 @@ tests/integration/run.sh --host you@server --dir pithead --readiness
 ```
 
 > NOTE: a pruned chain's file stays large. An in-place prune does not shrink the LMDB file: it
-> stays at the full-chain high-water mark (~250 GiB) with the freed space sitting as internal
-> free pages (Monero reuses them as the chain grows). To reclaim it you must rewrite the DB with
+> stays at the full-chain high-water mark, with the freed space sitting as internal free pages
+> (Monero reuses them as the chain grows). **How much is actually reclaimable is a measurement,
+> not an assumption** — `mdb_stat -ef` on an idle copy reports the freelist, and on this bench it
+> is 10 pages out of 67,605,667, so there is nothing to reclaim (#1446). To reclaim it where there
+> is, you must rewrite the DB with
 > `monero-blockchain-prune` (see
 > [`compact-chain.sh`](../../tests/integration/compact-chain.sh)). It's slow (it copies every block
 > over hours), though it reads through a snapshot so monerod keeps mining.
@@ -344,9 +349,10 @@ tests/integration/run.sh --host you@server --dir pithead --readiness
 > To run it against a live chain, bind-mount the source so the kernel refuses the rename; the
 > recipe and its guard are in `compact-chain.sh`'s header.
 >
-> The generic `mdb_copy -c` does not work: Monero ships a
-> patched LMDB and stock mdb_copy rejects the format (`MDB_VERSION_MISMATCH`). Often it's simplest
-> to leave the free pages.
+> `MDB_VERSION_MISMATCH` from a stock LMDB tool is the **lock-file** format, not a patched on-disk
+> format: it appears while monerod holds the environment, and the same tool opens an idle copy of
+> the same chain (both DBs are magic `0xbeefc0de`, version 1). It is not corruption — do not stop
+> monerod over it. Often the best move is to leave the free pages alone.
 
 ## Bench allocation and the rig lock
 
--- a/docs/dev/test-server-architecture.md
+++ b/docs/dev/test-server-architecture.md
@@ -49,9 +49,9 @@ Sizing (a pruned-Monero test bench fits in ~1 TB):
 
 A 1 TB NVMe (~931 GiB usable) holds the pruned bench with ~490 GiB to spare.
 
-The Monero row is the size measured on this bench, not a compaction target. The table used to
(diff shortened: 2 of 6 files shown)


VijitSingh97:
All checks green — 18 pass, 2 CVE sweeps skipping (expected, no image changed).

**Updating one caveat in the PR body.** It said the full `make lint` / `make test` were not run locally, because `make lint-sh` is the documented OOM path on this box and is serialized fleet-wide. CI covered it: **"Shell tests (shellcheck + pithead suite)" passed in 10m6s**, and "Lint per-surface (biome, yaml, markdown, proto, toml)" in 25s. So the gap I flagged is closed by CI rather than still open — flagging that explicitly so a reviewer does not weigh a caveat that no longer applies.

Not self-merging. This is same-lane work correcting my own lane's earlier claims, so it is held for the independent reviewer the controller launched.

VijitSingh97:
## Independent non-author review — VERDICT: MERGE, with finding 1 discharged first

This is same-lane work correcting the authoring lane's own earlier claims, so it got an independent pass rather than a same-lane verdict. Reviewer worked in a fresh worktree off `origin/develop-v2`, never a lane worktree. Controller judged the pass and **re-derived both findings at the PR head independently before ruling** — what follows is not a relay.

### Finding 1 — SHOULD-FIX, discharge before merge

`tests/integration/build-pruned-chain.sh:10` still reads:

```
#   4. prune the COPY          -> shrinks ~250G -> ~95G, full chain untouched
```

Re-derived by the controller at the PR head; the file is untouched by this PR (`git diff --stat origin/develop-v2...<head> -- tests/integration/build-pruned-chain.sh` empty, with a positive control on `compact-chain.sh` returning a real diffstat, so the empty result is proven to mean untouched rather than a broken invocation).

Correctly graded should-fix rather than blocking: this step prunes a copy it just made, not canonical data, so it does not trigger the unnecessary destructive rebuild that made the two READMEs hazardous. But it is being asked for **before** merge for a reason that is not severity. This line states the script's own designed outcome, and it is the origin of the retired expectation the rest of this PR retracts. Leave it and the PR's sweep is provably incomplete in the one file most likely to be read next by whoever runs the builder — and issue #1489, which closes on this PR, would close on that incomplete sweep. The fix is one comment line in a file the authoring lane already owns and already read.

Worth noting both halves of the line are retired, not just one: the `~250G -> ~95G` pair describes a shrink that this bench's measurement says does not happen.

### Finding 2 — SHOULD-FIX (low), fold into the same commit

The `MDB_VERSION_MISMATCH` retraction reads as a flat universal fact at three of its four sites. Re-derived: `docs/dev/release-server.md:352`, `docs/dev/test-server-architecture.md:83` and `tests/integration/testbench-README.md:48` carry it unhedged, while `tests/integration/compact-chain.sh:45` does cite `(#1446)`.

The underlying mechanism — LMDB versioning its lock-file separately from its data format — is genuine LMDB behaviour, so this is **not** the same shape of error as the "Monero ships a patched LMDB" claim being retracted. But the specific supporting evidence (both DBs magic `0xbeefc0de` v1) was measured once, on one bench, on one monerod build, and at those three sites it reads as universal. The neighbouring freelist sentences in the same files already hedge with "here"/`#1446`; matching them is three short edits.

### What the pass confirmed, and did not just assert

- **No surviving operator-facing imperative.** The "if it reads back near ~250 GiB, compact it" instruction — the one wired to a ~2.5 hour destructive rebuild — is gone from the repo and from the deployed copies. Swept unit- and number-agnostically and keyed on the imperative, not the figure, with a positive control fired in every invocation style.
- **No true statement taken as collateral.** Every removed line was read individually and classified. "LMDB never shrinks its file" is preserved everywhere it appeared; the `#1489` copy-then-swap block is untouched; `build-pruned-chain.sh`'s "full chain never modified" assertion is intact and correct.
- **Figures re-derived from the primary record, not from the PR body.** Free pages 10 of 67,605,667; 67,605,667 x 4096 = 276,912,812,032 bytes exactly; 257.9 GiB, rounded to "~258 GiB" in neither a flattering nor a deflating direction. The sizing table's Monero row is measured, not a second guess substituted for the first, and the unmeasured full-chain figure is correctly marked `[TODO: verify upstream]` per the repo's never-invent-a-fact rule. The controller recomputed the byte figure independently.
- **The ponytail commit dropped no load-bearing line.** Read line by line. One evidence citation lost some traceability; one piece of guidance survives two lines below where it was cut. The commit message's "8 lines" does not cleanly reconcile with the raw counts, and the pass **reported that discrepancy rather than papering over it** — the right call, and it does not change the finding.
- `make lint-docs-voice` rc=0 and `shellcheck --severity=warning` clean, both run by the reviewer rather than taken from the PR body.

### Recorded, not charged against this PR

The deployed copies under the testbench directory have **structurally diverged** from the repo copies — they carry extra operational detail, so syncing over them would destroy it. Both false claims are fully retracted in both places, and the PR body discloses the divergence. That is correct handling, but it leaves an owed follow-up: repo and deployed copies now need a reconciliation plan rather than an assumption that either is canonical. Not a blocker here.

### Deliberate non-target, confirmed untouched

`docs/hardware.md` and `docs/getting-started.md` are not touched by this PR. Their pruned-node sizing figure versus this bench's measurement is a separate open question, filed as #1502 — it is **two of our own dated August-2026 measurements disagreeing**, not our docs versus an upstream figure, and the standing ruling there is that no user-facing sizing number moves in either direction until the bench chain's prune provenance is settled. The refuted retention model was not re-filed.

### Disposition

Discharge finding 1 (and finding 2 in the same commit), push, and the controller merges — the authoring lane does not merge its own PR. Close #1489 by hand naming this PR once it lands, not before.

PR #1503, opened 2026-08-29
refactor(cli): cut small utilities and the doctor-check groups out of the remainder (#1105 Phase 2 P6)

VijitSingh97:
Cuts small utilities and the doctor-check groups out of `lib/pithead/99-remainder.sh` into four new slices, as #1105 Phase 2 P6.

This is a **pure re-partition**: the shipped artifact does not change by a single byte.

## What changed

| Slice | Lines | Functions |
|---|---|---|
| `19-small-utilities.sh` | 217 | 18 |
| `20-doctor-install-checks.sh` | 269 | 5 |
| `21-doctor-stack-checks.sh` | 334 | 13 |
| `22-config-value-helpers.sh` | 45 | 3 |

`99-remainder.sh` loses the corresponding 869 lines. One row of `docs/dev/file-budget.tsv` moves. Nothing else.

## The proof, in the order it carries weight

**(a) The shipped artifact is byte-identical, established two independent ways.** The `pithead` blob is `d79917f61bea453e3563bfbbafb74b254ebc3216` at both the base `3b2a71b` and the head `76dd7ad`; separately, extracting both to files and running `cmp` reports them identical at 712209 bytes. Blob-sha equality and a literal byte compare are different routes to the same claim, and both fired.

**(b) The parity gate is what makes (a) mean anything, and this is the argument to attack.** Byte-identity on its own is equally consistent with a different story: that nobody regenerated the artifact, and it is now stale against the very sources this PR edits. That would produce exactly the same identical blob. `make lint-pithead-parity` excludes it — `scripts/build-pithead.sh --check` rebuilds from the slices and refuses on any difference, and it passes at the head ("the committed artifact is exactly what 24 slice(s) build"). Only (a) and (b) together say both things that are needed: what ships is unchanged, **and** it is genuinely the product of the edited sources.

**(c) Build determinism is the licence to read (a) as a fact about the sources rather than a coincidence.** `build-pithead.sh` enumerates `lib/pithead/*.sh`, sorts under `LC_ALL=C` (pinned deliberately, so a locale cannot silently reorder the artifact), and joins with exactly one blank line between each pair. Nothing but slice bytes reaches the output — no date, hostname, git invocation, `$RANDOM`, or `uname` in the content path. The only external input is `PITHEAD_BUILD_ROOT`, and it selects a root, not content.

**(d) The four blank lines.** A reviewer diffing the move will find 869 lines removed and 865 added, and the 4-line gap is the first thing that looks wrong. It is the opposite: the build inserts exactly one blank line per slice boundary, so four new boundaries consume four blank lines. Their disappearance is *required* for byte-identity. Every other line moved verbatim; nothing was invented.

## What is NOT evidence here

A base/head suite pair was run (2979 passed / 0 failed on both legs). **It is a null experiment and it is not evidence about this refactor.** The suite consumes the built artifact, which is byte-identical across the two legs, so both legs executed the same 712209 bytes — the independent variable was never manipulated. `2979 == 2979` measures suite flakiness across two concurrent runs, and nothing more. It is reported here for completeness, not as support.

For the same reason, "the moved functions are present in the built artifact" is not cited: the artifact is unchanged, so that check could not have failed.

**Disclosure:** `tests/stack/run.sh` has no references to `lib/pithead/`; one integration selftest, `tests/integration/selftest-remote-endpoint.sh`, does.

## The declined adjacent merge: 380 lines, legal, declined on naming

`21-` and `22-` were one file in the predecessor cut, at **380 lines** — under the 400-line target in `scripts/lint-file-budget.sh`. Merging them back is therefore **legal**; it is declined on cohesion, because the config-value helpers are a different concern from the doctor stack checks. Nothing here is illegal and nothing was forced by the budget.

The arithmetic looks off by one and is not: the two slices sum to 379 (334 + 45), while the merged file is 380. The separator blank line is an interior line of the merged file, but once split it is generated by the build between two slices and belongs to neither. That is re-derivable from this branch forever with `wc -l` on the two files, plus one.

## Pure re-partition

Two of the four slices are byte-identical to the predecessor cut of this same region — `19-small-utilities.sh` at `0aa2586c849cfd0846eb52b6600f2a3e2b83a1ec` and `20-doctor-install-checks.sh` at `d309516838ed8e430dae7596936841315c062431`. Both are self-verifying against `git rev-parse HEAD:lib/pithead/<path>`. The remaining change is the 21-/22- split described above, proven by `cmp`: the predecessor's single file equals this branch's `21-` plus one newline plus `22-`, exactly.

*Provenance only:* the predecessor commit is `a056dec`, which was committed locally and never pushed. It is currently a dangling object and may be garbage-collected. **No claim above depends on it** — all of them re-derive from this branch.

## Over-engineering pass

The four slices contain **37 further legal cut points** (blank lines between top-level blocks, where `build-pithead.sh` guarantees a cut reproduces the artifact byte for byte): 18 in `19-`, 4 in `20-`, 13 in `21-`, 2 in `22-`. Splitting at all of them would yield 41 slices in place of 4, taking the repo from 24 slices to 61.

Declined. None of the four is over the 400-line target, so a finer split relieves no budget pressure; it would multiply the count that the concatenation-order contract has to be audited against; and the file that actually remains oversized is `99-remainder.sh` at 7079 lines, which later phases address. The finer split buys nothing measurable now and costs auditability.

## Call sites for the three tail helpers

Re-derived at this head, not carried over:

- `is_valid_host` — defined `pithead:5166`; called at 6517, 6529, 6676
- `is_valid_port` — defined `pithead:5175`; called at 6519, 6521, 6531, 6684, 11049
- `detect_host_timezone` — defined `pithead:5181`; called at 6696

## Gates

Run at head `76dd7ad`, base `3b2a71b`:

- `make lint-pithead-parity` — rc=0, 2026-08-29T06:32:48Z
- `make lint-file-budget` — rc=0, 2026-08-29T06:32:49Z

`make lint` was **not** run in full: `shellcheck` has OOM-killed sessions on this box, and `lint-sh` is the documented serialized exception. Only the two mechanical gates above were run here; CI covers the rest.

## Concurrency disclosure

Other lanes were running `tests/stack/run.sh` during the suite pair. That is allowed and it was symmetric across both legs — and given the pair is a null experiment, it does not bear on any claim in this PR.


VijitSingh97:
## Non-author gate pass — VERDICT: MERGE. No finding survives.

Independent pass in a fresh worktree at this head, briefed on eleven axes carrying this row's ruled conditions. Controller-judged. The author lane does not merge its own PR.

### The two claims that decide this, both re-derived by the reviewer's own commands

**Parity re-run at this head, not read from the body.** `make lint-pithead-parity` in the reviewer's own worktree: self-test all-green across 22 cases **including both firing controls** — a source-side and an artifact-side mutation each flip `--check` red — then `pithead parity OK — the committed artifact is exactly what 24 slice(s) build`, rc=0.

This was the right thing to make the top axis, and the author lane nominated it against its own PR. Parity is the single claim that excludes the stale-artifact story: byte-identity alone is equally consistent with *nobody regenerated the committed artifact and it is now stale against the very slices this PR edits*. Only both together say what ships is unchanged **and** is genuinely the product of the edited sources.

**Artifact byte-identity, two independent routes, with a fired control.** `git rev-parse` on both refs returns the same blob (`d79917f6…`); separately, both extracts `cmp` identical at 712209 bytes. Control: one byte appended to a copy, `cmp` reported it. The zero-diff is not a broken comparator.

### The re-partition is provably pure

Conservation derived at the blob level, not from a diff summary: the first 869 lines of the base remainder against the literal concatenation of the four new slices differ by **exactly four single-line deletions, all blank, at 218 / 488 / 823 / 869(EOF)** — the four new slice boundaries. `tail -n +870` of the base diffs byte-identical against the new remainder. So the 4-blank-line accounting is exact rather than approximate, and it confirms the body's explanation of the mechanism: those blanks were inline separators in the monolithic file, and the build now supplies one at each boundary. **Their disappearance is required for byte-identity, not a wart.**

Naming judgement independently re-derived to a third derivation: all three tail helpers' call sites resolve inside `parse_and_validate_config()` and `resolve_worker_target()` — **neither is a doctor function**, so the rationale for splitting `22-` out of a file named for doctor checks holds. The declined adjacent merge checks out too: 334 + 45 = 379, and the same two concatenated with the build's separator is 380, which is the figure for the merged file we declined. Both under target, so the under-400 claim is safe either way.

### On the suite pair

Correctly presented in the body as a **null experiment** and, per the standing ruling now recorded in the cut map, not re-run and not attacked. The suite consumes only the built artifact and the artifact is the same blob either side, so both legs would execute an identical program — the independent variable was never manipulated. `2979 == 2979` measures suite flakiness across two concurrent runs; it is not evidence about this refactor, and the body says so in those terms. The disclosed sliver is worded exactly as it should be — `tests/stack/run.sh` has no `lib/pithead/` references and one integration selftest does — verified with an unshimmed recursive grep rather than the ugrep-shimmed default.

`a056dec` is confirmed genuinely dangling and confirmed **not load-bearing**: the blob shas the body quotes verify against current HEAD paths, so no claim here depends on that commit surviving garbage collection. That was the right way to cite a vanishing object.

### One honest non-certification, recorded and not treated as a finding

The reviewer **declined to certify the 37-legal-cut-point enumeration** (18/4/13/2). A crude brace-depth heuristic gave 19/5/14/1 — consistent to within one per file, but the reviewer said plainly that its own instrument is noisy on `${…}`, `[[ ]]` and array literals and does not rigorously reproduce what the build would accept as a legal cut. It marked the claim NOT RE-DERIVED rather than passing it as checked.

That is correct conduct and it changes nothing here: the enumeration supports a **declined** option, so a ±1 error in it cannot make a wrong file land. Recorded so nobody later mistakes it for a verified count.

### Also re-derived

`make lint-file-budget` rc=0 from the reviewer's own run (26-case self-test green); exactly one tsv row moved (`99-remainder.sh` 7948 → 7079) and nothing else; all four new files far under the 400 target so no row was spent; `bash -n` clean on the built artifact; the four slices sort immediately before the remainder under `LC_ALL=C`; and the build's content path reads nothing but slice bytes — no date, hostname, git call, `$RANDOM` or `uname`, with `PITHEAD_BUILD_ROOT` selecting a tree and never entering the joined output.

Stated as unchecked, honestly: `make lint` in full was not run by either side, for the documented shellcheck OOM hazard on this box — CI is the only coverage for it, and the merge waits on CI.

Issue #728, opened 2026-07-21
Data-dir moves are now dashboard-confirmable but assert_safe_dir is a blocklist, not an allowlist


VijitSingh97:
## Summary

With #719, a data-directory move (`MONERO_DATA_DIR`, `TARI_DATA_DIR`, `P2POOL_DATA_DIR`, `DASHBOARD_DATA_DIR`) is now **confirmable from the authenticated dashboard** — those four keys sit in `CONTROL_DASHBOARD_CONFIRM_KEYS`, so the operator types `APPLY` and the move commits without a host shell.

The path itself is still validated only by `assert_safe_dir` (`pithead`, ~L1624-1644), which is a **blocklist**, not an allowlist. It refuses a fixed set of system / home / bare-mount roots (`/`, `/root`, `/home`, `/var/lib`, `/mnt`, `$HOME`, …) plus relative paths and `..` traversal. Any **non-blocklisted absolute path** passes.

## Why this matters now

Before #719, choosing a data dir was a host-shell operation — host-shell trust already implies filesystem-wide reach, so a blocklist that only stops the catastrophic-`chown -R`/`rm -rf` fat-fingers was proportionate. After #719 the same move is reachable at **dashboard trust**. A confirmed move can now target an arbitrary absolute path that is not on the blocklist — e.g. a subfolder of another user's home, or another service's data volume — and pithead will `mkdir`/`chown -R` it and bind-mount it into a recreated container. That is a trust-level escalation for the destination path: from host-shell-trust to dashboard-trust.

(To be clear: the perimeter secrets — wallets, view keys, credentials, `tor_egress_firewall`, `dashboard.control.enabled` — remain host-only DEST and are NOT reachable this way. This is specifically about the data-dir *destination path* for the confirm-gated move.)

## Fix direction

For dashboard-**confirmed** data-dir moves specifically, tighten the destination check from a blocklist to an **allowlist**: permit only paths under the stack's own data roots (or an explicitly-permitted parent the operator has opted into), and refuse everything else. The host-shell `apply` path can keep the existing blocklist behaviour if we want host operators to retain full reach. Alternatively (weaker), surface a stronger, path-specific warning on the confirm step.

The clean split is: `assert_safe_dir` stays the host-shell guard; add an allowlist gate that applies only when the move arrives through the control channel's confirm path.

## References
- #719 (config-editability / confirm-gate that made data-dir moves dashboard-confirmable)
[...]

PR #1504, opened 2026-08-29
refactor(tests): split confirm-gate, data-management and spool-audit out of run.sh (#1105)

VijitSingh97:
## What this is

Row R9 of the #1105 run.sh split: three domain files cut out of `tests/stack/run.sh` by exact
line range. **No line of moved text is edited.** `tests/stack/run.sh` goes 1229 → 1034 lines.

| New file | Domain | Moved lines |
|---|---|---|
| `tests/stack/test-confirm-approval.sh` | #719 confirm gate — a disruptive change needs a typed APPLY, and that token does not unlock the perimeter (#713 refusal wording) | 73 |
| `tests/stack/test-data-management.sh` | #728 confirmed data-dir move is allowlisted to the stack data root; #33 a non-destructive commit still proceeds with no token | 56 |
| `tests/stack/test-spool-audit.sh` | #349 audit log growth is bounded; #33 hardening — spool intake cap, symlink refusal, stale sweep | 72 |

All three land under the 400-line floor, so **no new budget rows**. Only `tests/stack/run.sh`'s
ceiling lowers. Three rows the `--generate` pass silently lowered on files this cut never touched
(`dashboard/.../egress.py`, `.github/workflows/ci.yml`, `tests/integration/run.sh`) were restored
by hand to their committed values — lowering a row you do not own reds every un-rebased branch in
every lane.

## The design call: these files deliberately do NOT self-arm

`build_control_sandbox` is called exactly **once** in run.sh, from a control-core section that
stays behind. None of these three blocks calls it — all three are pure consumers. They follow the
shipped `test-control-add-only-ssrf.sh` disclosure precedent (source in place, position-locked,
dependency disclosed in the header) rather than the self-arm pattern most domain files use.

"It should self-arm like its neighbours" is the obvious review note and it is wrong for this
shape — but **not for the reason an earlier draft of this PR gave, and the correction is worth
stating rather than quietly swapping in.** That draft said re-running the builder would RESET the
accumulated config and spool state. That mechanism is false. `build_control_sandbox()` sets
`C="$SANDBOX/control"` — a fixed path — its `mkdir -p` only creates, its copies are static inputs,
and `seed_control_env`/`control_config` are DEFINED inside it and never called. It writes no
`config.json` and touches nothing under `data/control/{requests,staged,results,audit}`. A second
call would be harmless.

**That is precisely why self-arming buys nothing: the builder could not create the state these
domains depend on.** `test-spool-audit.sh` is the clearest case — *every* assertion in it is a
count against an exact number (audit lines and bytes either side of a trim, spool files left after
a capped run, then zero after the next drain) taken against the shared spool this run has
accumulated. Only running in position, after the sibling sections that accumulate it, supplies
that. `test-data-management.sh` reads back config state the confirm-gate domain applied just
before it, carried by `$UUID3`, which no builder call can supply either.

The coupling that DOES bite is write-side, and its recorded firing is real: the rig-worker
token-mask cluster moved with a re-derived `$C`, its applies wrote **extra** result files into the
shared results dir, and a still-in-run.sh assertion counting that dir went RED 2901/1. That is
pollution of a counted directory — a different mechanism from anything being reset, and keeping
the two apart is the point.

Each header instead makes the contract executable with a `: "${NAME:?}"` guard over the ambient
names **that file actually reads** — re-derived per file, not copied. Sourced out of position,
each file now stops on a named variable instead of degrading into assertions against an unbuilt
sandbox.

## Two ordering dependencies this cut surfaces, both disclosed on both sides

1. **`$UUID3` is a NEW file-to-file dependency.** `test-confirm-approval.sh` assigns it in its
   moved text; `test-data-management.sh` reads it. The dependency did not exist while both
   sections lived in run.sh — the split creates it. run.sh sources the two stanzas back to back,
   in that order, and the reader guards the name.
2. **`$UUID5` is inherited from a section that STAYS in run.sh.** The approval-gate section
   assigns it; `test-spool-audit.sh` reuses that id. Nothing in the moved text reveals this; only
   the whole-file read-versus-assign audit surfaces it.

Also disclosed, and invisible to a variable-only sweep: **`control_config()` is not a top-level
lib.sh function.** It is defined *inside* `build_control_sandbox()`, so it does not exist until
that builder has run. `test-confirm-approval.sh` and `test-data-management.sh` both call it — a
second, independent reason those two cannot stand alone. `test-spool-audit.sh` calls none of the
three nested functions, so its dependency is in variables only.

`$WALLET` is likewise **not** a top-level constant: lib.sh assigns it only inside the two sandbox
builders. The cut map's `$WALLET` note for c7/c8 was checked at cut time and is **right** — 4
non-comment in-domain reads in `test-confirm-approval.sh`, 3 in `test-data-management.sh`. c10
carried no note and has 0; the absence was checked too, not assumed.

## What was RUN

**Full `tests/stack/run.sh` suite, both sides, detached, verdict taken from the rc file:**

| | before (`3b2a71b`, pristine worktree) | after (`c4bba3b`, this branch's HEAD) |
|---|---|---|
| exit code | 0 | 0 |
| verdict | **2979 passed, 0 failed** | **2979 passed, 0 failed** |
| log lines | 3292 | 3292 |
| `^  ✗` failure markers | 0 | 0 |
| `CONFIG_FILE: readonly variable` | 3 | 3 |

**This is the THIRD after-proof, and the first two were discarded on purpose.** Each time, a header
was amended after its own proof had run — #1 for a citation-wording regression found by the
over-engineering pass, #2 for the false builder-reset mechanism corrected above after review. A
header amend after its own proof leaves a proven-tree/committed-tree gap that nothing in git can
close afterwards, so each superseded run was thrown out as evidence and the whole suite re-run
against the committed HEAD. All three runs produced identical figures — which is the point worth
stating plainly rather than using as an excuse: the numbers did not move, the *provenance* did, and
"it would have been the same anyway" is not knowable until the run lands. The table describes the
tree this PR actually ships, and `git status` is clean at that tree.

The two sides ran under different load, stated rather than smoothed over. `before` overlapped a
second suite of mine for about 4 of its 13 minutes (N=2); this after-run had the box to itself. Load
moves wall-clock, not output, and every instrument below compares content or ordering rather than
timing. No other lane had a suite registered, and the live count was re-derived by walking process
working directories with the match pattern split across variables — a probe matching its own command
line reports a phantom second suite, which is what a naive count did here before it was corrected.

**The totals were PREDICTED before the after-run reported, and the prediction is the interesting
part.** The obvious guess is 2979 + 3, one per new `domain_ran` guard. Reading `domain_ran` shows
it emits `bad` *only* on its failure path and nothing at all on the green path — so three new
guards add **zero** assertions, and the correct prediction is an unchanged 2979/0. That also
independently re-derives the recorded "a red run totals 2980, a green one 2979" note: the extra
assertion is the guard's own `bad`. `CONFIG_FILE` was predicted to shift by ZERO *by construction*
— those warnings are emitted by `test-appliance-identity.sh` at its own lines, and deleting lines
from run.sh cannot move another file's output. It did not shift.

**The ordered comparison (#1398) — the only instrument that can see a permutation:**

| Check | Result |
|---|---|
| Sorted multiset, whole output | **2 differing lines** — the two halves of one ed25519 keygen fingerprint (#1325) |
| Header set (sorted) | IDENTICAL |
| **Header ORDER** | **IDENTICAL — 0 of 274 positions differ** |
| Per-test verdict multiset | IDENTICAL |

Controlled, not assumed: permuting two adjacent headers in a copy makes the **ordered** diff fire
while the **sorted** diff stays silent on that same input — which is the proof the two instruments
actually discriminate rather than both being blind.

**Byte-identity, re-derived from git rather than from the extraction.** Each new file's text below
its header was rebuilt out of `git show <base>:tests/stack/run.sh` at the stated ranges and diffed
against the committed file; the whole of `tests/stack/run.sh` was independently reconstructed the
same way (delete each range, splice in the three-line stanza). Both diffs are empty.

**Every instrument was made to fire before its silence was believed:**

| Control | Result |
|---|---|
| Shift each chunk range by one line | diff FIRES on all three (a broken comparator would stay silent) |
| Drop one stanza line from the run.sh reconstruction | FIRES — 1033 vs 1034 |
| Understate `tests/stack/run.sh`'s budget row by one | gate rc=1, failing **by name**; mutation confirmed applied by row count first |
| Strip the `: "${C:?}"` guard and re-audit | the `(has ${..:-} default somewhere)` annotation vanishes — confirming it is self-inflicted noise, not a real default |
| Set each guarded name in turn and re-source | the abort walks to the NEXT name, 23 for 23 — an inert entry would make it repeat |
| Set every guarded name, then source | the guard is passed and the file reaches its own first section header — so the guard line, not something upstream, is what stopped the other 23 arms |

**The `: "${NAME:?}"` guards were MEASURED, not argued.** Each of the three files was sourced
standalone under `set -uo pipefail` in a clean environment with its ambient names unset, then with
them set one at a time in guard order. All **23** guarded names (8 / 9 / 6) abort at that file's own
guard line naming themselves — `test-confirm-approval.sh: line 62: C: parameter null or not set`,
and so on through `AUDIT` — so no name in the list is inert. (`rc` is 127 in every aborting arm and
cannot discriminate an unbound read from a missing command, so the message and the stop position are
the evidence, not the code.)

**What this does not show, stated because the distinction is the whole point:** it proves each
guarded name fires, not that the guard list is *complete*. A name a file reads but its guard omits
passes this test in silence. Completeness rests on the separate whole-file name audit reported above,
and AXIS 0(d) of the review brief asks for the sandbox-armed form of the same check — `lib.sh`
sourced and `build_control_sandbox` run first — which is the arm that would expose an omission.

**The base has since MOVED, and this is what was re-checked rather than assumed.** `origin/develop-v2`
advanced from `3b2a71b` to `cc05fad` while the after-proof was running (#1501 and #1503 merged, the
latter being Phase 2 P6). Nothing under `tests/stack/` is touched by either, so every range,
byte-identity and suite result below still describes the tree this branch ships. Re-derived against
the new base rather than argued: the merge is clean with zero conflicts, and the merged tree was
checked *positively* — it carries both sides' budget rows (`99-remainder.sh` 7079 from the base,
`run.sh` 1034 from here), all three new files, and `run.sh` at exactly 1034 lines. The budget gate
re-run against `cc05fad` is rc=0 with its self-test passing and the base probed unmoved either side.
The branch is deliberately **not** rebased: rebasing would change the tree and stale the proof again,
which is the exact gap that cost this cut its first after-proof.

One expected note falls out of the move: against `cc05fad` the gate reports `99-remainder.sh`'s
ceiling *rising* 7079 -> 7948, because this branch predates that row being lowered. It is the
stale-HIGH foreign row described below, it passes the gate, and the three-way merge takes the base's
7079 — confirmed on the merged tree, not predicted.

**Base ref probed either side of the gate run** — `origin/develop-v2` at `3b2a71b` both times, so
the budget result is not the moved-base artefact.

`shfmt -i 4 -d` and `shellcheck --severity=warning` are clean on all three new files.

## Two things the map got wrong, re-derived at the tip

- **The `c10` range in the cut map overshoots by four lines.** The map gives 1482-1558 at its own
  sha; the true content plus its trailing blank is 1482-1554. The extra four lines are a `# ---`
  divider, a shellcheck directive and a `source` stanza that **introduce the next block and must
  stay**. Taking the map's range literally would have nested a domain inside the new file — the
  #1400-guard trap, where all five usual instruments are blind. The cut range used here is
  898-970 at the tip, which is 73 lines, reconciling exactly with the corrected 1482-1554.
- **The nested-stanza check fired on a row the map did not flag.** By the time this cut reached
  the tip, *four* domain stanzas sat immediately after the spool-audit block, not the one that was
  there at the map's sha. The map's row carries no nesting warning; the grep found it anyway.

## What I did NOT do

- I did not restore the `# ---` divider on the new stanzas. Most split-produced stanzas lack it —
  the extractor has never emitted one — so adding it here would make the file *less* consistent,
  and a full sweep is a standalone cosmetic change on a budgeted shared file.
- I did not touch `tests/stack/lib.sh`, and I did not "fix" the ordering dependencies by seeding
  `$REQS`/`$RESULTS`/`$STAGED`/`$AUDIT` into the new files. Each is a plain derivation from `$C`,
  so a seed would duplicate run.sh's definitions and could drift from them, and it would buy
  nothing: `$C` itself keeps every one of these files non-standalone either way.
- The three foreign budget rows were restored to their committed (stale-HIGH) values rather than
  corrected. A stale-HIGH ceiling passes the gate; lowering a row this cut does not own would red
  every un-rebased branch in every lane.

## A hazard this cut hands to R13, stated rather than left to be found

`test-spool-audit.sh` reads `$UUID5` from the approval-gate section that stays in run.sh. **That
section is exactly what the cut map's R13 row reunites** into the `test-control-add-only-ssrf.sh`
family. When R13 moves it, the assignment leaves run.sh — and whether this file still sees
`$UUID5` then depends entirely on where R13's stanza lands relative to this one.

I am not pre-solving that here; the range is not mine this round and guessing at R13's shape would
be worse than naming the constraint. What this cut does do is make the failure loud instead of
silent: the `: "${UUID5:?}"` guard turns "R13 landed its stanza after mine" from a confusing
assertion failure deep in a spool count into a named abort on the variable. **R13 must source its
reunited file BEFORE the spool-audit stanza, or seed `$UUID5` itself.** Worth carrying onto the
R13 row.


VijitSingh97:
## Controller verdict — MERGE, with one should-fix discharged first

Gate pass `reviewer-20260829T065006Z`. It returned MERGE-with-one-finding and I agree, after
re-deriving the decision-turning claim myself rather than relaying it.

### The finding is real, and I re-derived it independently

**All three new headers state a mechanism that is false: "Re-running the builder would reset that
accumulated config and spool state."** I read `build_control_sandbox()` at this PR's head rather
than taking the reviewer's word for it. What it actually does: defaults `$WALLET`, sets
`C="$SANDBOX/control"` — a FIXED path, so a second call lands in the same place, not a fresh one —
`mkdir -p` over `build/`, `dashboard/` and `data/{monero,tari,p2pool/stats,tor,dashboard}`, copies
`pithead`, `config.reference.json`, the tari template and `docker-compose.yml`, makes stubs, sets
`$CTRL_LOG`, and **defines** (does not call) `seed_control_env` and `control_config`.

It never touches `data/control/{requests,staged,results,audit}` — those four are absent from the
builder's body — and it never writes `config.json`, because the function that writes it is only
defined. `mkdir -p` is non-destructive on an existing directory in any case. **So the builder
cannot reset the spool, the audit log or the applied config, and the reason written into three
headers is not the reason.** The reviewer went further and proved it by running: a redundant
self-arm call inserted into both consumer files, full suite, 2979 passed / 0 failed, reverted and
diffed back byte-identical.

The headers' OTHER claim — the rig-worker precedent, "its applies wrote extra result files into
the shared results dir and a still-in-run.sh assertion counting that dir went red" — is accurate,
and it is write-side pollution from EXTRA CALLS, a different mechanism from the builder resetting
anything. That distinction is the whole finding.

### Why this is a should-fix and not a nit

Not because the shipped behaviour is wrong — it is correct either way, and nothing in the diff
misbehaves. Because **R10 through R14 are still to be cut by this same lane, from this same
template**, and a header in this codebase is read as durable documentation. A wrong mechanism
offered as the justification for a design pattern is the shape that gets copied forward as
established fact and then defended. We have paid for that exact failure before.

### What to change, and what NOT to change

Correct the wording only. **Do not switch these files to self-arm** — that changes behaviour and
would need its own proof; "the builder is safe to call" does not oblige anyone to call it. The
true reasons these files are not standalone are already known and already disclosed in the PR:

1. `control_config()` is defined INSIDE `build_control_sandbox()`, so it does not exist until that
   builder has run — which is why c7 and c8 cannot stand alone regardless.
2. The `$UUID3` file-to-file dependency the split creates, and `$UUID5` inherited from a section
   that stays in run.sh.
3. For `test-spool-audit.sh` specifically: execution ORDER relative to the sibling sections that
   accumulate the spool it counts. That is a real constraint and it survives the finding intact —
   it is just not a fact about the builder.

State the builder's idempotence positively where the false claim stands now, so the next reader
does not re-derive this from scratch.

### After-proof #3 is required, and I considered not requiring it

A comment-only amend cannot change behaviour, and I could have ruled that a diff restricted to
comment lines is self-proving. **I declined that.** The box is idle, the suite is ~13 minutes, and
the rule it would weaken — no header amend after its own proof — is one this lane paid for on this
very cut when it discarded after-proof #1. Buying a cheap exception to a rule that just cost
something is how the rule stops being load-bearing. Re-run at the amended head and cite #3.

### Recorded as NOT findings — do not let these be re-raised

- **"c7+c8 should have been one file."** The reviewer leaned that way and said plainly it is a
  judgment call, not a defect: merging would remove the `$UUID3` inter-file dependency, while
  one-file-per-issue (#719 vs #728) serves navigation. Both far under the budget floor, so size
  did not drive it. Declined — re-cutting costs a full round and buys no correctness.
- **Axis 3 header ORDER was not checked against the suite logs** (they are lane-local, not in the
  repo). The whole-file reconstruction is a stronger structural proof of the same fact, with a
  fired negative control, and it is byte-identical to the shipped `run.sh`.
- **The 23-name guard set is proven to FIRE, not proven COMPLETE.** The PR says so in those words
  and the reviewer's independent re-derivation found every read-never-assigned name to be a subset
  of each file's guard. Good enough; the honest limit is stated rather than hidden.

### What I verified myself before ruling

Merge-base still `3b2a71b`; the intervening range touches ZERO `tests/stack/` files, with the same
grep returning 4 on this PR's own diff as a fired positive control; `merge-tree --write-tree`
clean; the merged tree carries `run.sh` at 1034 AND `99-remainder.sh` at 7079; and the budget row
is right in the merge — this branch's stale-HIGH 7948 loses to the base's 7079, `run.sh` lowers
1229 to 1034, and no new rows appear for the three new files. **The 7079 to 7948 gate note is
therefore expected and is not a defect** — confirmed on the merged tree, not predicted.

I re-derive CI green at source at the amended head before merging.

Issue #1500, opened 2026-08-29
integration harness: zmq-probe's connect is unbounded — a filtered host blocks past the probe budget


VijitSingh97:
Found by an independent post-merge review of PR #1499 (verdict PASS overall; this is the one gap worth a follow-up).

**The probe's initial connect has no timeout.** `tests/integration/zmq-probe.sh:131` opens the socket with `exec 3<>/dev/tcp/$1/$2` before any `timeout $3` wrapper applies — the timeout argument only bounds the post-connect reads. Consequences by refusal shape:

- **Closed port (RST):** returns in ~2ms, rc=1 — fine, and this is the only refusal shape the PR's test matrix exercised.
- **Filtered / black-holed host (no RST, no accept):** the connect blocks with no bound from the harness. Measured: still blocked at 6s under an external `timeout 6` (TEST-NET-1 address, `192.0.2.1`). On a real OS it eventually resolves via the kernel's own SYN-retry timeout (commonly 20s–130s+), so it is not an infinite hang — but it can block far past the probe's intended budget (run.sh passes 8) with no attribution until then.

This matters precisely on the path the PR itself discloses as unexercised: remote-mode against a real remote node, where a firewalled or partitioned host is the realistic failure. Fix shape: bound the connect itself (e.g. run the connect inside the same `timeout` budget as the reads, or a `timeout`-wrapped subshell that hands back the fd/result).

**Companion one-liner, same file:** a malformed READY frame whose property length field falls on the frame-body boundary makes `zmq_ready_socket_type` die at `zmq-probe.sh:109` with a bash arithmetic error (`16#: invalid integer constant`) instead of a clean verdict string. Confirmed fails-noisy-never-false-green (rc=1, empty stdout, propagates to `it_fail`), so it is cosmetic-plus-attribution, not a correctness hole. Suggested guard immediately before line 109:

    [ $(( (p + 4) * 2 )) -le "${#body}" ] || { echo "malformed-ready property runs past frame"; return 1; }

Both items are harness-side (`tests/integration/`), small, and can land as one PR. Context: #1497 (the composed preflight blindness this probe exists to close).

VijitSingh97:
Fixed by #1505, merged to `develop-v2` as `cf6a888`. Closing by hand — `Closes #N` does not fire in this
repo, because PRs merge to `develop-v2` while the default branch is `develop`.

Re-derived at source before closing rather than trusting the merge notification: the PR reads `MERGED`
[...]

PR #1505, opened 2026-08-29
fix(integration): bound the ZMQ probe's connect and its frame-length reads (#1500)

VijitSingh97:
Closes the two items in #1500, plus four more sites of the second one's defect class that fuzzing turned up.

## 1. The connect was unbounded

`timeout` cannot wrap a shell redirection, so the opening `exec 3<>/dev/tcp/...` inherited only the kernel's SYN-retry deadline. A closed port answers with RST in ~2 ms — the only refusal shape #1499's matrix exercised, and the reason this stayed hidden. A filtered or black-holed host, which is the realistic remote-mode failure, blocked far past the probe's budget with no attribution at all.

Bounded with a `timeout`-wrapped throwaway connect in a child shell. The child cannot hand its fd back, so the cost is a second connect on the reachable path — measured at 0 s against the live monerod, unchanged from base.

A filtered host is now named `connect-timeout` and kept apart from `connect-refused` deliberately: a refusal is an answer and points at the port, a timeout is the absence of one and points at a firewall, a partition or the wrong address. Collapsing them would send a remote-mode operator to debug the wrong box.

## 2. The frame parser died on short reads instead of naming them

Every length field is read with `16#`, and `16#` on an empty string is a fatal bash arithmetic error — an interpreter message on stderr and an empty verdict, rather than a reason.

**Scope beyond the issue, stated plainly.** #1500 names one site. Fuzzing `zmq_ready_socket_type` over even-length hex (the only shape `od` can produce) found **five**, so this fixes the class rather than the instance:

| Input | Site | Named by the issue |
|---|---|---|
| `2d` — COMMAND frame, 1 byte | short-form size slice empty | no |
| `2f` — long-form flags, 1 byte | 8-byte size slice empty | no |
| `47c40aba3454cc6862` | declared size overflows the shell's arithmetic and returns **negative**, making the body substring fatal in its own right | no |
| `0400` | zero-length body, nothing to read the command name from | no |
| `04120552454144590b536f636b65742d54797065` | property value length on the frame boundary | **yes** |

The last one is the only one reachable from a **well-formed READY prefix**: `p` crosses `size` *inside* the iteration, and the loop test only bounds it on entry. I initially argued this path was unreachable from the bounds at the top of the function and was wrong — the fuzz settled it.

## What was run

- **Live, `192.0.2.1:18083` (TEST-NET-1, black-holed), probe budget 3 s, killed externally at 15 s.** BASE: still blocked at 15 s, output empty, external rc 124. HEAD: returns at 3 s with `connect-timeout no answer from 192.0.2.1:18083 within the probe budget — filtered, black-holed or the wrong address, not a refusal`. Same command, same target, same budget, base sourced from `origin/develop-v2`.
- **Live, the monerod on this box.** Base and head both `ok 127.0.0.1:18083 speaks ZMTP and advertises Socket-Type XPUB`, 0 s each. The added pre-connect is not measurable there.
- **Fuzz over the parser, the same generator both sides:** 878 interpreter deaths before, **0** after, with the known-good live frame still parsing as the positive control.
- **Mutation battery, 5 legs.** Each leg `cmp`s the file first and refuses to count if the mutation did not apply. Every guard has a demonstrated red: dropping the header guard reds 4 assertions, the size/`avail` guard 4, the property guard 2, the `connect-timeout` class 1, and a connect that always burns the budget reds 2. The file is byte-identical to pre-battery afterwards. The legs also confirm **rc=1 was already true while the bug was live** — which is why the new cases assert empty stderr and not just the return code.
- `make test-integration-selftest` rc=0 — 571 assertions across 11 files, 55 of them in `selftest-zmq-probe`.
- shellcheck 0.11.0 and `shfmt -i 4 -d` over the Makefile's exact globs (not `make lint-sh`, which is serialised on this box); `make lint-md`, `lint-docs-voice`, `lint-file-budget`, `lint-operator-strings`. All clean.

## What this does NOT prove, and one thing I chose not to code around

- **The self-test cannot prove the timeout class.** That needs a black-holed address, which is not hermetic, so the self-test covers the closed-port path behaviourally (verdict plus elapsed, so a bound that always waits the full budget reds) and the `connect-timeout` verdict purely. The bound itself is proven live only, by the measurement above.
- **The probe already required `timeout` on the target** — the read wrappers have always used it. If it were missing, the pre-connect now returns 127 and the verdict reads `connect-refused` against a reachable node. That is a pre-existing dependency two lines away, not a new one, and adding a 127 arm on speculation is the kind of branch that never gets exercised. Named here rather than coded around.
- The host is interpolated into the `bash -c` string the same way it was already interpolated into `/dev/tcp/$1/$2`. No new exposure; ports are already refused unless numeric (`selftest-remote-endpoint.sh`).

Docs: `docs/dev/integration-testing.md` — the probe bullet gains the connect bound and the two-verdict distinction with the measured numbers; the self-test paragraph gains the truncated-frame classes and why those cases assert stderr.

Refs #1500, #1497.

## Over-engineering pass

Two things cut, four kept with reasons.

**Cut.** A `127` arm on the pre-connect for a target without `timeout` — a branch that would never be exercised, guarding a dependency that already existed two lines away; it is disclosed above instead. And a third defect I thought I had found: I read the short/long frame branch as testing the **size** byte rather than the flags byte, which would have made every successful probe burn its full budget. It reads `${hdr:0:2}`, the flags byte, and is correct. It died before it reached the diff.

**Kept.** (a) `((flags & 0x02))` is evaluated twice — once for the header bound, once for the size branch. Merging them into the existing if/else removes the second test but duplicates the error message into both arms; one arithmetic test is cheaper than two copies of a string. (b) The four guard sites beyond the issue's one — same function, same defect class, six lines, each with a demonstrated red. (c) `assert_clean_verdict` calls the function twice, once for stderr and once for stdout/rc; the alternative is fd juggling or temp files for one saved call in five cases. (d) The seven-line comment above `zmq_probe_snippet` — it carries the measurement and the reason a second connect exists, and without it the next reader deletes the pre-connect as redundant.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -328,6 +328,14 @@ and `--list` prints it).
   answers a dial with rc 0. Measured on one host, three targets: the live node and the
   published-but-dead port were indistinguishable to a dial (both rc 0) and separated by the
   handshake. Real nodes advertise `XPUB` rather than `PUB`; both are accepted.
+  The probe bounds its own connect ([#1500](https://github.com/p2pool-starter-stack/pithead/issues/1500)). `timeout` cannot wrap a shell
+  redirection, so the opening `exec` inherited only the kernel's SYN-retry deadline. A closed port
+  answers with RST in about 2 ms, which is why this stayed hidden; a filtered or black-holed host
+  produced no verdict at all until the kernel gave up, far past the probe's budget. Measured
+  against a black-holed address on a 3-second budget: 15 s and empty output before, 3 s and a
+  named verdict after. A filtered host is reported as `connect-timeout`, kept apart from
+  `connect-refused` on purpose — a refusal is an answer and points at the port, a timeout is the
+  absence of one and points at a firewall, a partition or the wrong address.
 - End-to-end mining. Workers are online (`proxy_workers >= --workers`), stratum has connections,
   and total hashes are accumulating ([#28](https://github.com/p2pool-starter-stack/pithead/issues/28)).
 - Posture propagated. `MONERO_RPC_BIND`, `DASHBOARD_SECURE`, `XVB_ENABLED`, and `TARI_REQUIRED`
@@ -624,6 +632,12 @@ through a bare `it_warn` — by wording, and by shape for the drops that never s
 `selftest-zmq-probe.sh` drives the ZMTP verdicts from captured and hand-built wire fixtures, so
 every failure class — a silent peer, a non-ZMQ listener, a ZMTP peer that is not a publisher, a
 READY frame carrying a decoy `Socket-Type` value — is reachable with no socket and no stack.
+It also covers the truncated and hostile frames that used to end the parser rather than be named
+by it ([#1500](https://github.com/p2pool-starter-stack/pithead/issues/1500)): every length field is read with `16#`, which is fatal on an
+empty string, so a peer that stalled part-way through a header left an interpreter error on stderr
+and an empty verdict. Those cases assert the empty stderr alongside the reason, because the return
+code was already 1 while the bug was live and a case checking only the code would have passed
+against it.
 
 ---
 
--- a/tests/integration/selftest-zmq-probe.sh
+++ b/tests/integration/selftest-zmq-probe.sh
@@ -138,6 +138,75 @@ assert_contains "published-but-dead port is named no-greeting" "$v" "no-greeting
 # The verdict must name the endpoint it judged — a bare reason in a matrix log is unattributable.
 assert_contains "the verdict names host:port" "$v" "28099"
 
+echo "== a truncated or hostile frame is NAMED, not fatal (#1500) =="
+
+# The defect this section guards is a parser that DIES on a short read: every length field is read
+# with `16#`, and `16#` on an EMPTY string is a bash arithmetic error, so the function exits with
+# an interpreter message on stderr and an empty verdict instead of a reason. It fails noisy rather
+# than false-green, but an instrument that cannot say WHY is barely an instrument.
+#
+# Each case therefore asserts three things TOGETHER, and the stderr half is the load-bearing one:
+# rc=1 alone was already true of the broken parser, so a case that checked only rc would have
+# passed against the bug it exists to catch.
+assert_clean_verdict() { # <label> <hex> <expected reason>
+    local out err rc
+    err=$(zmq_ready_socket_type "$2" 2>&1 >/dev/null)
+    out=$(zmq_ready_socket_type "$2" 2>/dev/null)
+    rc=$?
+    assert_rc "$1 is refused" "$rc" "1"
+    assert_contains "$1 is named" "$out" "$3"
+    assert_eq "$1 costs no interpreter error" "$err" ""
+}
+
+# Found by fuzzing the parser over even-length hex, which is the only shape od can produce.
+# A COMMAND frame that stops after its flags byte: the short-form size slice is empty.
+assert_clean_verdict "a 1-byte short header" "2d" "malformed-ready"
+# The same, long form (flags bit 0x02 set): the 8-byte size slice is empty.
+assert_clean_verdict "a 1-byte long header" "2f" "malformed-ready"
+# A long frame declaring 0xc40aba3454cc6862 bytes. That overflows the shell's own arithmetic and
+# comes back NEGATIVE, which makes the body substring fatal in its own right — so the bound has to
+# compare the declared size against what ARRIVED, not against a doubled length.
+assert_clean_verdict "a long frame whose size overflows" "47c40aba3454cc6862" "malformed-ready"
+# A COMMAND frame declaring a zero-length body: nothing left to read the command name from.
+assert_clean_verdict "a frame declaring an empty body" "0400" "malformed-ready"
+# THE ONE #1500 NAMES, and the only one reachable from a WELL-FORMED READY prefix: 18 bytes of
+# "READY" + the "Socket-Type" key and nothing after it. `p` crosses `size` inside the iteration,
+# so the loop test above cannot bound the 4-byte value-length read that follows.
+assert_clean_verdict "a property length on the frame boundary" \
+    "04120552454144590b536f636b65742d54797065" "malformed-ready"
+
+# The guards must not have been bought by rejecting good frames: re-assert the live capture here,
+# so a bound that is one byte too tight reds in this section rather than passing quietly above.
+assert_eq "the live monerod frame still parses after the guards" \
+    "$(zmq_ready_socket_type "$LIVE_READY")" "ok XPUB"
+assert_eq "the long-framed READY still parses after the guards" \
+    "$(zmq_ready_socket_type "$READY_LONG")" "ok XPUB"
+
+echo "== the connect is bounded, and named apart from a refusal (#1500) =="
+
+# A refusal is an ANSWER — host up, port closed. A timeout is the absence of one, and it points at
+# a firewall, a partition or a wrong address rather than at the node. Collapsing them would send a
+# remote-mode operator to debug the wrong box.
+v=$(zmq_pub_verdict "CONNECT-TIMEOUT" 10.0.0.5 18083)
+rc=$?
+assert_rc "a filtered host is refused" "$rc" "1"
+assert_contains "a filtered host is named connect-timeout" "$v" "connect-timeout"
+assert_ne "a filtered host is NOT reported as a refusal" "${v%% *}" "connect-refused"
+
+# Behavioural, not a text match on the snippet: run the real snippet against a closed loopback
+# port. A bound that always waits the full budget would satisfy a text check and red here, and the
+# elapsed half is what proves the added pre-connect did not become the new cost. The TIMEOUT class
+# itself needs a black-holed address and so is proven live on the bench, not in this file.
+start=$SECONDS
+snippet_out=$(bash -c "$(zmq_probe_snippet 127.0.0.1 1 9)" 2>/dev/null)
+elapsed=$((SECONDS - start))
+assert_contains "a closed port still reports CONNECT-FAIL" "$snippet_out" "CONNECT-FAIL"
+if [ "$elapsed" -lt 3 ]; then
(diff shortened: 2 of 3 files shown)


VijitSingh97:
**Addendum: the over-engineering pass, re-run properly.**

The PR body's pass was hand-rolled, because the gate that demands `/ponytail-review` names a command with no definition loadable on this box. The skill itself does exist on disk in Claude's own format — the plugin manifest declares only `hooks`, never `skills`, so it never reaches the skill list. I read it and ran the real thing in its own format. It found two `shrink:`s my version missed, and I am declining both with the reason rather than quietly dropping them.

```
zmq-probe.sh:L88-89: shrink: `local hdr_bytes=2` + a 1-line `if` to set 9.
    `local hdr_bytes=2; ((flags & 0x02)) && hdr_bytes=9`, 1 line.
zmq-probe.sh:L126-129: shrink: 4-line `if` guard. The form #1500 itself suggests,
    `[ $(((p + 4) * 2)) -le "${#body}" ] || { echo "..."; return 1; }`, 1 line.
net: -4 lines possible.
```

Both declined on one ground: every other guard in this function is a multi-line `if ... fi` (there are six of them above these), and mixing two guard forms inside one 70-line parser costs more in readability than four lines buys. The skill does not accept "match the surrounding code" as an argument; this repo's CLAUDE.md does, and where they disagree the repo wins. Recorded here so the choice is visible rather than assumed.

Nothing was flagged `delete:`, `yagni:`, `stdlib:` or `native:`. The five-case helper has five callers, and the two comment blocks that carry measurements are the reason the second connect does not get deleted as redundant by the next reader.

VijitSingh97:
**Follow-up commit, from the review brief's second axis: every OTHER blocking operation in this probe.** The answer changed the PR — a connect bound does nothing for a host that accepts and then goes silent, and that host was costing **3x** the stated budget.

**Enumerated, and measured rather than reasoned:**

| Operation | Bounded? |
|---|---|
| pre-connect (`timeout $3 bash -c "exec 3<>…"`) | yes, by `timeout` |
| **the real connect, after the pre-check passes** | **no — see below** |
| `printf $ZMQ_GREETING_BYTES >&3` (64 B) | no in principle; unreachable in practice |
| greeting read | `timeout $3` |
| `printf $ZMQ_READY_BYTES >&3` (27 B) | no in principle; unreachable in practice |
| READY header read | `timeout $3` |
| READY body read | `timeout $3` |

So the budget was **per-read, not total**: three sequential reads, each able to burn it. Against a listener that accepts and never sends, measured at budget 3, **head and base were identical at 9015 ms** — the connect bound bought exactly nothing there.

Two of those three reads are pure cost, and skipping them is **verdict-preserving by construction, not by measurement**: `zmq_pub_verdict` discards the READY result whenever the greeting verdict fails, so a peer that sent no greeting cannot be judged by its READY. Same for the 512-byte fallback on an empty header — `head -c 2` already drained what was there. **9015 ms → 3013 ms, same verdict. Live monerod unchanged, 22 ms, `ok … XPUB`.** This is the probe's founding case, the published-but-dead port, so it was the path paying 3x most often.

**Still unbounded, stated rather than hidden.** The real connect that follows a passing pre-check is not bounded — the child shell cannot hand its fd back, so there is no way to bound the connect that actually yields the fd. What the pre-check buys is that it just succeeded microseconds earlier, so reaching it requires the host to become unreachable inside that window. The two writes are unbounded in principle and 64/27 bytes against a default socket send buffer, so no peer can stall them by refusing to read.

**Third axis — do the five guards fail loud?** Yes, and checked mechanically rather than by eye: `zmq_ready_socket_type` has seven `return 1` paths and exactly one `return 0`, the success line. The adversarial case is a `Socket-Type` present with a truncated value, which could plausibly return `ok` with an empty type — it does not: it reds through the composed verdict as `malformed-ready … value length runs past the frame body`, rc=1. A guard reaching `it_fail` is the only path out.

**Fourth axis — the mutation battery.** Each leg restored the pristine file, applied its mutation, and `cmp`'d **before running anything**, printing the differing-byte count; a leg whose mutation failed to apply reports `FILE UNCHANGED — proves nothing` and is not counted. All five applied. The file was confirmed byte-identical to pre-battery afterwards.

**One bug this commit introduced and caught.** A backtick inside the unquoted `<<EOF` heredoc is command substitution **at generation time**. shellcheck flagged it (SC1089); the self-test did not, because the generator's stderr is discarded and the damage lands inside a comment. Fixed, with the reason written into the comment.

VijitSingh97:
## Controller verdict: **MERGE, as-is. No discharge owed.**

Gate pass `reviewer-20260829T071518Z` returned **MERGE**, no blocking finding, and it is a strong
pass rather than a shallow one — it stood up three purpose-built stalling listeners, ran the
black-hole test live on both base and head, reproduced two of the five mutation legs with the
mutation proven applied first, and fired a control before every clean result. It also caught and
fixed a bug in its own first listener (it was accepting the probe's throwaway connect instead of
the real one) and disclosed that rather than letting a wrong-for-the-wrong-reason result stand.

**The review covered commit 1 (`fe20f4e`) only** — it re-derived the head itself and correctly
recorded this as a single-commit PR, because `a8215a9` landed after dispatch. Ruling on the delta
is therefore mine, and it is below.

### What I re-derived myself, rather than relaying

- **Merge mechanics at the current base.** merge-base `cc05fad`; the only intervening change is
  #1504, which touches **zero** `tests/integration/` files — positive control: the same pathspec
  returns 2 on this PR's own diff. `merge-tree --write-tree` clean.
- **CI green at source**, not from a waiter's rc: every check on `a8215a9` concluded success, the
  two CVE-sweep rows being by-design skips.
- **The greeting-skip is verdict-preserving BY CONSTRUCTION, and I checked the construction.**
  `zmq_pub_verdict` evaluates `zmq_greeting_verdict` first and returns 1 on its failure without
  ever reaching `zmq_ready_socket_type`, so a peer that sent no greeting cannot have its READY
  result affect the outcome. The emitted text is also literally unchanged: the old path produced
  an empty `hdr` and an empty `body` and printed `READY ` — the same string the new early exit
  prints. The early exit closes fd 3 exactly as the fall-through does, so no cleanup is skipped.
  One budget instead of three, same verdict, same output.
- The reviewer's own live measurement of the accept-then-stall adversary at **9.0 s (3 read stages
  x 3 s)** is independent corroboration of the defect commit 2 fixes. Two of us reached it by
  different routes.

### My finding on the delta — raised, attacked, and REFUTED by my own test

I flagged the `elif [ -n "$hdr" ]` narrowing as a different class from the greeting skip, and I
was going to file it as a should-fix on the wording. It fires on the greeting-*passed* path, where
the READY result does decide the verdict, and the comment's stated reason — that with an empty
header a 512-byte read "can only return empty too" because `head -c 2` already drained what was
there — conflates EOF with TIMEOUT. `timeout N head -c 2` returning empty does not prove the peer
sent nothing ever, only that it sent nothing in that window.

**I tested it instead of filing it, and the test killed the finding.** `timeout N head -c 512`
emits *nothing* unless the byte count is satisfied or EOF arrives inside the budget — measured
directly: a socket holding 4 bytes from a peer that stays open yields an empty result at
`head -c 512`, while `head -c 4` on identical state yields the bytes. That control is what makes
the measurement mean anything; my first attempt at this test had no control and produced a
contradictory pair, which is the only reason I caught the confound at all.

So the old unconditional `else` could never have recovered a late partial READY from a peer that
stays open — the only shapes it could ever return data for are EOF-within-budget or a >=512-byte
flood, and neither is the slow-peer case I was worried about. **Skipping it is safe, for a
stronger reason than the comment gives.** Not a finding; recorded here so nobody re-raises the
`elif` on the reasoning I started with, and so the stronger reason is on the record.

Failure direction is also safe in any case: `zmq_ready_socket_type ""` returns 1 with the named
verdict `no-ready peer completed the greeting then sent no READY command`. Nothing in this delta
can turn a red into a green, which is this lane's own worst-outcome class.

### The one should-fix, and why it does NOT hold up the merge

The reviewer's Finding 1 is that the "3 s budget" framing could leave a reader thinking 3 s ships.
I checked the durable half at source and **it is already honest**: `docs/dev/integration-testing.md`
says "Measured against a black-holed address on a 3-second budget", which states 3 s as the
condition of the measurement, not as a default. The looseness is confined to the PR body. Re-derived
for the record: `zmq_pub_probe`'s own default is 5, and the single live call site passes 8
explicitly — so the slow-endpoint question applies to the caller's 8, unchanged by this PR.

A PR body is the merge record, not repo documentation, so the clarification belongs here rather
than in a force-push at this lane's context level. **This is where my R9 ruling this morning and
this one differ, deliberately:** R9's amend was to comment text inside `.sh` files the suite
sources, so it re-staled that PR's own proof and I required a third after-proof. Nothing a
markdown file or a PR body says can move this PR's self-test. That is the distinction, stated so
it does not read as a cheap exception to the rule R9 just paid for.

### Recorded NOT-findings — do not let these be re-raised

- The **double-connect TOCTOU** (reviewer's Finding 3). Real, disclosed in the PR body in plain
  words, and the reviewer could not construct a listener shape that reaches it. The remaining
  unbounded path exists because a child shell cannot hand its fd back, and the PR states that
  rather than hiding it. Disclosure was the right call.
- **connect-timeout vs connect-refused is decorative to code** (Finding 2). One consumer, and it
  does not branch on the text. The PR never claimed a code consumer; its argument is about the
  human reading the failure line, which stands on its own.
- **Three SC2059 notices** at `selftest-zmq-probe.sh:44,45,47` — pre-existing, outside this diff's
  only hunk.
- **The 878 -> 0 fuzz count is RELAYED, not verified** — no fuzzer ships in the tree. The reviewer
  said so plainly instead of laundering it, and re-derived the *shape* of the claim as a proxy:
  all five input classes crash base and are named on head. Three of the five mutation legs are
  likewise relayed. I am merging on the two reproduced legs plus the five-row base/head table, and
  naming the gap rather than treating it as closed.

**Lesson worth keeping from this PR, and it is the author's:** "we bounded the connect" was true
and did not bound the hang. Ask what the *other* blocking operations are, every time.

*Ruled on opus. Nothing here was fable-class: the proof class was already settled and this applies
it.*

PR #1507, opened 2026-08-29
refactor(cli): cut the setup, wizard, address-type and appliance-provisioning slices out of the remainder (#1105 Phase 2 P7)

VijitSingh97:
Phase 2 P7 of #1105. Five order-preserving contiguous slices come out of the head of
`lib/pithead/99-remainder.sh`. The shipped `pithead` artifact is byte-identical to the one on
the base commit, and the whole diff outside `lib/pithead/` is one budget row.

## The cut

| slice | lines | holds |
|---|---|---|
| `23-setup-and-credentials.sh` | 196 | `detect_os`, `deps_satisfied`, `check_prerequisites`; the `PLACEHOLDER_NODE_*` constants; local node RPC credential generation (`default_node_username`, `generate_node_password`); Tor v3 onion client auth (`b32encode_hex`, `generate_onion_client_keypair`); `cred_needs_generating`, `persist_node_credentials` |
| `24-config-wizard.sh` | 226 | `ensure_config_exists` and the four first-run wizard stages (`wizard_ask_core`, `wizard_ask_shape`, `wizard_write_config`, `wizard_print_pointer`) |
| `25-address-types.sh` | 232 | `monero_address_type` and `tari_address_type` — the two payout-address classification gates, each with its vendored Python decoder |
| `26-appliance-tls-provisioning.sh` | 292 | `ensure_onion_password`, `_console`, the TLS/name helpers (`appliance_tls_dir`, `appliance_base_name`, `appliance_site_names`, `appliance_cert_alt_string`, `cert_san_string`), `provision_console_login`, `provision_ssh_access`, `appliance_mint_cert`, `wizard_mint_cert`, `ensure_appliance_dashboard_password` |
| `27-defaults-and-config-validation.sh` | 162 | `apply_appliance_defaults`, `WORKER_LIST_JQ`, `validate_worker_endpoints`, `migrate_legacy_workers`, `validate_energy_config` |

All five are under the 400-line target, so **none takes a budget row**.
`lib/pithead/99-remainder.sh` drops **7079 → 5966**, and that row is the entire tsv diff (87
rows before and after; ceiling equals the file's actual line count on both sides).

- **Blob sha**: `git rev-parse HEAD:pithead` and `git hash-object pithead` after a full rebuild
  both give `d79917f61bea453e3563bfbbafb74b254ebc3216`.
- **Literal `cmp`**: the pre-build copy against the rebuilt file — identical, 712209 bytes.
- `pithead` is absent from `git status` after `scripts/build-pithead.sh`, and absent from this
  commit.

### 2. The parity gate — and why it is the claim that matters

`make lint-pithead-parity` returns **rc 0**: *"the committed artifact is exactly what 29 slice(s)
build."*

This is the gate to attack, and it is doing work that byte-identity alone cannot. Byte-identity
on its own is **equally consistent with the story that nobody regenerated the artifact and it is
now stale against the very sources this PR edits** — that story produces exactly the same
identical blob. Parity excludes it. Only the two together say that what ships is unchanged *and*
is genuinely the product of the edited sources.

**Firing control, with the mutation proven applied first.** I staged all six files before
mutating, because `git diff --numstat` prints nothing for an untracked file and a blind
instrument reads exactly like a clean one. Then I rewrote one line of
`26-appliance-tls-provisioning.sh`:

- `git diff --numstat` on the staged file: `1  1` — the mutation demonstrably landed.
- The gate returned non-zero and named the line itself: `5884c5884`, quoting both sides.
- **Second, independent route to that number:** `grep -Fn` for the target string in `pithead`
  predicted line **5884** before the gate ran. Two derivations agree; one would not be a claim.
- Restored: `numstat` empty, parity back to rc 0.

### 3. Build determinism — the licence to read (1) as a fact about the sources

`scripts/build-pithead.sh` concatenates `lib/pithead/*.sh` in `LC_ALL=C` order, joined with
exactly one blank line, with nothing but slice bytes in the content path.
`PITHEAD_BUILD_ROOT` is the only external input and it selects a root, not content. So the
identical artifact is a fact about the sources, not a coincidence.
`scripts/build-pithead.sh --self-test`: all cases pass.

### 4. The blank-line accounting (this is why lines-removed exceeds lines-added)

The diff removes 1113 lines and adds 1108. The gap of five is **required** for byte-identity,
not lost content: the build re-inserts exactly one blank line per slice boundary, so the four new
interior boundaries plus the one seam against the remainder consume exactly five blanks from the
sources.

Conservation, machine-checked: `196 + 226 + 232 + 292 + 162 = 1108`, `+ 4` interior `+ 1` seam
`= 1113 = 7079 − 5966`.

### 5. Per-slice `bash -n` — the instrument byte-identity is blind to

A boundary landing inside a heredoc still reproduces the artifact byte for byte, because
concatenation puts it back; parity passes and proves nothing about it. `25-address-types.sh`
carries two Python `PYEOF` heredocs and `26-appliance-tls-provisioning.sh` an sshd `EOF`, so this
is a live hazard on this cut, not a formality.

**`bash -n` on each of the 29 slices individually: all parse.** `bash -n pithead`: parses.
`shfmt -i 4 -d lib/pithead/` (the Makefile's own flags): **0 bytes**.

*(The ponytail gate keys off `git rev-parse --abbrev-ref HEAD` in the pinned shell cwd, not the
worktree the command ran in, so it names a branch unrelated to this PR. Disclosed rather than
allowed to mean the diff went unreviewed — the pass below was run by hand against the real diff.)*

**There is no new logic.** Multiset diff over the staged change, extracting sides with
`awk '/^--- /{next} /^-/{print substr($0,2)}'` (never `grep '^-[^-]'`, which silently drops
removed blank lines and makes a pure move read as a content change):

- **added ∖ removed = 0 lines** — nothing in this PR is new text.
- **removed ∖ added = 5 lines, all blank** — exactly the build-supplied separators accounted for
  above.

**Is the file count minimal?** Every adjacent merge among the five shipped slices, each carrying
its `+1` build separator:

| merge | size | legal (<400, no new row)? |
|---|---|---|
| `23`+`24` | 423 | no |
| `24`+`25` | 459 | no |
| `25`+`26` | 525 | no |
| `26`+`27` | 455 | no |

**No adjacent merge of the shipped files is legal**, so at these boundaries five is minimal.

## The split degree, and the coarser cuts that were available

The region divides into six blocks along its own seams — 89 / 106 / 226 / 232 / 292 / 162 lines.
The largest function in the whole region is `monero_address_type` at 114 lines (next:
`wizard_ask_core` 93, `tari_address_type` 90), so **nothing here is indivisible and no new budget
row is justified**; a row is worth spending only when a single
function forces it (P3's `firstboot_wizard`, 521 lines).

Grouping those six blocks into contiguous groups, with each merge paying its `+1` separator:

- **Three files: not available at these seams.** A first file must end at 89 or 196 (ending at
  423 already exceeds 400). Either way the rest is 1022 or 915 content lines, which two files
  capped at 400 cannot hold.
- **Four files: not available either — all ten groupings fail.** `(a)(b)(c)(def)`=688,
  `(a)(b)(cd)(ef)`=459, `(a)(b)(cde)(f)`=752, `(a)(bc)(d)(ef)`=455, `(a)(bc)(de)(f)`=525,
  `(a)(bcd)(e)(f)`=566, `(ab)(c)(d)(ef)`=455, `(ab)(c)(de)(f)`=525, `(ab)(cd)(e)(f)`=459,
  `(abc)(d)(e)(f)`=423. Every one exceeds 400.
- **Five files: exactly two groupings work** — `(ab)(c)(d)(e)(f)` = 196/226/232/292/162, and
  `(a)(bc)(d)(e)(f)` = 89/333/232/292/162. I took the first.

**A three-file cut IS legal if boundaries are allowed to fall inside a domain, and it is declined
on naming, not on legality.** Cutting at remainder lines 364 and 726 — both verified single
blank lines between top-level blocks, non-blank on both edges — yields 363 / 361 / 386, all under
400 and costing no row. It is declined because 364 falls between `wizard_ask_shape` and
`wizard_write_config`, splitting the first-run wizard across two files, and 726 falls between
`appliance_base_name` and `appliance_site_names`, splitting the certificate-name family whose two
halves exist precisely to be one list (#1132). **Three files was available and rejected, not
overlooked.**

Why `(ab)` rather than `(a)(bc)`: the merged file's two halves are both non-interactive host-side
setup work — prerequisite detection and the credentials setup generates before it can write a
config — whereas `(bc)` would pair credential generation with the interactive wizard, a distinct
subsystem. Both groupings are legal; this is a naming call and I am naming it as one.

Parity is structurally blind to *which* slice a function lands in: any cut at a single blank line
between two top-level blocks reproduces the artifact, including a semantically wrong one. Only
reading validates assignment. Three judgement calls to look at:

1. **`ensure_onion_password` opens `26-appliance-tls-provisioning.sh` and is a credential
   generator**, thematically at home in `23-setup-and-credentials.sh`. Order-preserving
   contiguity puts it here — it sits at remainder line 663, below the address-type gates. Same
   shape as P6's `21-doctor-stack-checks` carrying three non-doctor helpers. Named here rather
   than papered over.
2. **`_console` sits in `26-` but 18 of its 19 call sites are outside this region**, in the
   firstboot-wizard slices (artifact lines 2469–3685); exactly one is in-region. It is a
   console-output helper, and the file's name reaches it through `provision_console_login`, but
   the fit is the weakest of the twelve functions in that file.
3. **`27-defaults-and-config-validation.sh` opens with `apply_appliance_defaults`**
   (appliance-specific) and continues with three general config validators. The name covers both
   halves deliberately. The three validators' call sites are `parse_and_validate_config` at
   artifact 6558/6559/6560 — the function that opens the new remainder, immediately below this
   slice, so the grouping follows the caller.

The TLS/name helpers are genuinely shared, which their own headers claim and I re-derived:
`appliance_site_names` has 11 call sites, **9 of them outside this region** and on both sides of
it — six inside `check_appliance_cert` (artifact 4727–4781, in the doctor slices) and three inside
`generate_caddyfile` (8161–8168). I re-derived the enclosing function for each rather than
inferring it from the line number.

## Budget

`scripts/lint-file-budget.sh` returns **rc 0**, run locally in a full worktree where
`origin/develop-v2` resolves. Per #1452 the monotonic half does not run in CI, so this proof is
**local-only**: taken against base `cc05fad25b76d74b7216633db8d5a82e1bdef1bf` at
2026-08-29T07:28Z, with `origin/develop-v2` probed before and after the run and unmoved across it
(every worktree shares one `.git`, so a peer's fetch can move the base mid-run).

## What I did NOT do

- **No full-suite base/head pair, deliberately.** It would be a **null experiment**, not evidence:
  the suite consumes only the built artifact, and that artifact is byte-identical on both sides,
  so both legs would execute an identical program and the result is fixed before launch. Such a
  pair measures suite flakiness, not this refactor. (`tests/stack/run.sh` has zero references to
  `lib/pithead/`; one integration selftest, `selftest-remote-endpoint.sh`, does reference it —
  so the honest claim is "the stack suite has none", never "zero coupling".)
- For the same reason there is **no "the moved functions are present in the built artifact"
  check** — the artifact is unchanged, so that check could not fail.
- **`make lint` / `make lint-sh` not run, and the lint lock not taken.** A slice cut adds zero new
  shellcheck exposure: `Makefile:60` makes `pithead` the first shellcheck argument, `pithead` is
  the byte-for-byte concatenation of the slices, and parity refuses drift — so every slice byte is
  already analysed in its real assembled context. The cut's only new exposure is the shfmt pass,
  which is clean above. CI's Shell tests job is the arbiter.
- **No non-author review yet.** Everything above is author-verified only.


VijitSingh97:
## Controller verdict: **MERGE, as-is. No discharge owed.**

Non-author gate pass `reviewer-20260829T073231Z` returned **MERGE** with one non-blocking
should-fix. It is a strong pass, and two things in particular make it stronger than a confirmation
of the author's own work:

- **It fired its OWN independently-chosen parity mutation**, at a different line (5868) from the
  author's (5884), with `git diff --cached --numstat` proving the mutation landed before the gate
  ran and an independent `grep -Fn` predicting the line number in advance. That shows the gate reds
  on an arbitrary injected defect, not merely on the one case the author selected — which is the
  difference between checking a control held and checking the instrument works.
- **It fired a positive control on `bash -n` too**, renaming a `PYEOF` delimiter mid-file and
  confirming the parser reds with a named here-document error before restoring. That is the one
  instrument byte-identity is blind to, and it was the right place to spend a control on this cut.

The brief's proof-class framing worked as intended: the pass did not request a base/head suite
pair and did not score its absence as a gap.

### What I re-derived myself before merging

- **CI success at the exact head** `9062795`, read from the run's own `head_sha`, not from a
  checks summary or a waiter's rc.
- **Merge mechanics:** merge-base `cc05fad`; the intervening range — my own two merges this seat,
  #1504 and #1505 — touches **zero** `lib/pithead/` files, with the same pathspec returning 6 on
  P7's own diff as a fired positive control. `merge-tree --write-tree` clean.
- **The merged tree resolves the budget rows correctly, and I checked it rather than predicting
  it:** `lib/pithead/99-remainder.sh` 5966 (this PR's lowering) AND `tests/stack/run.sh` 1034
  (R9's). Disjoint rows, both survive the three-way.
- **The merged tree's `pithead` blob is `d79917f6…` — identical to the base's.** The artifact is
  unchanged *through the merge*, which is a stronger statement than byte-identity on the branch,
  and it is the claim the whole W2 proof class rests on.

### On the reviewer's Finding 1 — confirmed, and it does not earn a discharge

**I re-derived the count independently and the reviewer is exactly right.** Over the built
artifact at the head, `appliance_site_names` has **19 mentions: 14 comments, 1 definition, and 4
real invocations** — `check_appliance_cert` (4767) and `generate_caddyfile` (8168) outside the
region, two inside. Not 11 call sites with 9 outside. The PR's "six inside `check_appliance_cert`
and three inside `generate_caddyfile`" line groups are real, but only one line in each is an
invocation; the rest are comments *naming* the function while explaining the #1132 contract.

**The conclusion the number was supporting survives untouched:** two genuine outside call sites,
in the doctor slices and the Caddyfile renderer, still establish that the helper is shared, which
is the whole placement argument. Nobody would place a function differently on 4 than on 11.

**It merges without a discharge, and here is the rule I am applying rather than an exception I am
making.** I checked where the claim lives: **it appears in zero of the five new slice file
headers** — I grepped all five — so it exists only in the PR body. That puts it in the same class
as #1505's "3 s budget" framing earlier today (PR body, corrected in the verdict) and *not* in
R9's class (false mechanism inside `.sh` headers the suite sources, which earned a discharge and a
third after-proof). Durable in-repo documentation gets a discharge; a merge-record citation gets a
correction in the record, which this comment is. The author is also correctly wound down, so the
alternative would be spawning a lane to edit a number in a PR body.

### Ratified, so these are not re-raised

- **The three disclosed slice-assignment judgement calls all stand.** `ensure_onion_password` is
  stranded at the head of the TLS file by contiguity (verified at base remainder line 663);
  `_console` is a weak fit at 18-of-19 call sites outside the region, correctly disclosed, and no
  legal contiguous cut improves it because the callers are firstboot-wizard-side and were cut long
  ago; `27-`'s pairing follows its caller, which opens the post-cut remainder.
- **The split-degree claim is stated honestly.** The 3-file cut at remainder lines 364/726 is
  available and rejected on naming, never claimed illegal, and the reviewer found both blank lines
  independently before reading the claim.
- **Axis 8's not-run lint is not a gap.** `pithead` is shellcheck argument 1 and is the
  byte-for-byte concatenation, so every line is already read in real context; the only other
  surface touching `lib/pithead/*.sh` is `shfmt`, a per-file property, already run at 0 bytes.
- **The 20 leak-scanner hits in the diff are a known false-positive class** on the base58/emoji
  alphabet constants, proven pre-existing at 15 on both the base and the current tip. This PR
  relocates them and introduces zero new hits.

### One thing the pass hit that is worth carrying forward

The reviewer ran `lint-file-budget.sh` against its own `origin/develop-v2` **after my #1504 merge
had moved that ref underneath it**, and got a false FAIL naming `tests/stack/run.sh` — a file this
PR never touches. That is the fleet's recorded base-ref-moves-under-you trap, hit live, diagnosed
correctly, and disclosed rather than reported as a finding. It is also why I verified the budget
rows on the *merged tree* above instead of accepting either the branch's numbers or the gate's.

*Ruled on opus. Not fable-class: the proof class was settled by the w20 ruling and this applies it.*

PR #1508, opened 2026-08-29
refactor(cli): cut config parsing and the preserved-state/dirs slices out of the remainder (#1105 P8)

VijitSingh97:
Phase 2 P8 of #1105. Two order-preserving contiguous slices come out of the head of
`lib/pithead/99-remainder.sh`. **The built `pithead` artifact is unchanged, byte for byte.**

## The cut

Region: `99-remainder.sh` lines **1–539** at base `2186a1f`, derived by CONTENT, not by the map's
offsets. It opens at `parse_and_validate_config() {` (remainder line 1) and ends at the single
blank line 539, which sits between `prepare_directories`'s closing `}` and the
`CONTROL_SECRET_PATHS` comment block.

| slice | lines | contents |
|---|---|---|
| `lib/pithead/28-parse-and-validate-config.sh` | 485 | `parse_and_validate_config` |
| `lib/pithead/29-preserved-state-and-dirs.sh` | 52 | `load_preserved_state`, `prepare_directories` |

Remainder **5966 → 5427**. Slice count 29 → 31.

**The map's printed range for P8 is not what I cut, and that is deliberate — see "Slice assignment"
below.**

- **Blob sha.** `git rev-parse HEAD:pithead` = `d79917f61bea453e3563bfbbafb74b254ebc3216`, and
  `git rev-parse origin/develop-v2:pithead` = the same sha. The artifact is the identical git blob
  on both sides.
- **Literal `cmp`.** `git show <base>:pithead` against the rebuilt working-tree file: identical at
  **712209 bytes**.

Two routes, because one derivation is not a claim.

### 2. The parity gate — and why it is the claim that matters

Byte-identity ALONE is equally consistent with *"nobody regenerated the artifact, and it is now
stale against the very sources this PR edits"* — that produces the same identical blob.
`make lint-pithead-parity` is what excludes it. Only (1)+(2) together say the artifact is unchanged
AND is genuinely the product of the edited sources.

- **Clean:** `scripts/build-pithead.sh --self-test` then `--check` → `pithead parity OK — the
  committed artifact is exactly what 31 slice(s) build.` rc 0.
- **Fired, with the failure line predicted first.** Before touching the gate,
  `grep -n 'Initializing data directories' pithead` → **6837**. I then staged
  `29-preserved-state-and-dirs.sh`, mutated its line 33, and proved the mutation applied with
  `git diff --numstat` on the STAGED file (`1 1 lib/pithead/29-preserved-state-and-dirs.sh`) —
  a numstat on an *untracked* file prints nothing and would have "proved" a mutation that never
  landed. The gate then failed with `6837c6837` and quoted both sides. Predicted line and reported
  line converged by two independent routes. Reverted; parity green again at the commit.

### 3. Build determinism — the licence to read (1) as a fact about the sources

`scripts/build-pithead.sh` concatenates `lib/pithead/*.sh` in `LC_ALL=C` sort order, joined with
exactly one blank line, with nothing but slice bytes in the content path. `PITHEAD_BUILD_ROOT` is
the only external input and it selects a root, not content. So a byte-identical artifact is a fact
about the sources, not a coincidence of the builder.

Independently confirmed by extraction rather than by arithmetic — for each slice,
`awk 'NR>=a&&NR<=b' pithead | cmp - lib/pithead/<slice>`:

- `28-` identical at artifact **6319–6803**
- separator blank at **6804**
- `29-` identical at artifact **6805–6856**
- separator blank at **6857**
- `99-remainder.sh` identical from **6858** to EOF (12284)

### 4. The blank-line accounting (this is why lines-removed exceeds lines-added)

485 + 52 = **537 added**; 5966 − 5427 = **539 removed**. The difference is exactly **2**: one
interior boundary (`28-`|`29-`) and one seam (`29-`|`99-`). The build re-inserts exactly one blank
line per boundary, so those two blanks are removed from the sources and restored by the
concatenation. **Their disappearance is REQUIRED for byte identity — it is not lost content.**

Pure-move multiset diff over the staged `lib/pithead/` diff: added ∖ removed is **empty**;
removed ∖ added is **exactly two blank lines** and nothing else. (Sides extracted with
`awk '/^--- /{next} /^-/{print substr($0,2)}'`, not `grep '^-[^-]'`, which drops removed blank
lines and makes a pure move read as a content change.)

### 5. Per-slice `bash -n` — the instrument byte-identity is blind to

A boundary falling inside a heredoc still reproduces the artifact byte for byte, because
concatenation puts it back; parity would pass and prove nothing about it. Each slice parsing
ALONE is what rules it out. `bash -n` over all 31 slices individually: **31 ok / 0 fail.**

`shfmt -i 4 -d` over the two new slices and the new remainder: **0 bytes of diff.** The build also
refuses a blank-edged slice by name; first and last lines of all three touched files are non-blank
(`parse_and_validate_config() {`/`}`, a comment/`}`, a comment/`fi`).

The ponytail gate keys off the branch name in the pinned shell cwd, not the worktree the command
ran in, so it named `fix/1342-mutation-lock` — a merged, deleted branch this PR has nothing to do with —
and let the retry through without checking that a review happened. **The mismatch does not mean the diff went unreviewed** — the
pass's two useful halves for a P-series cut are run here by hand:

1. **Is the diff a pure move?** Yes — added ∖ removed is empty (§4). There is no new logic to
   over-engineer: not one line of the 537 added is new.
2. **Is the file count minimal?** See below — every contiguous grouping is enumerated with its
   `+1` build separator, and the answer is arithmetic rather than judgement.

## The split degree, and the coarser cuts that were available

The region has three top-level blocks, separated by single blank lines (the only legal cut points):

| block | remainder lines | size | contents |
|---|---|---|---|
| B1 | 1–485 | **485** | `parse_and_validate_config` |
| B2 | 487–516 | 30 | comment + `load_preserved_state` |
| B3 | 518–538 | 21 | `prepare_directories` |

Every contiguous grouping, with the `+1` build separator per interior join:

| cut | files | new tsv rows |
|---|---|---|
| B1+B2+B3 | 538 | 1 (at 538) |
| **B1 \| B2+B3** | **485 / 52** | **1 (at 485)** |
| B1+B2 \| B3 | 516 / 21 | 1 (at 516) |
| B1 \| B2 \| B3 | 485 / 30 / 21 | 1 (at 485) |

**Every option costs exactly one new row**, because B1 is 485 lines on its own and over the
400-line target. So the row is not avoidable and the file count here is a NAMING call, not a budget
call — which is the opposite of P2–P7, where the split degree was set by avoiding rows.

**Largest function in the region, by name and count: `parse_and_validate_config`, 485 lines.** It
is one function; splitting it is a semantic refactor and Phase 2 does not do those. Nothing else in
the region is indivisible (`load_preserved_state` 27 lines, `prepare_directories` 21).

I took **B1 | B2+B3**:

- It sets the **smallest** available new row (485, against 516 and 538). The row ratchets
  downward-only, so a tighter one is strictly better.
- **Three files was available and legal, and is declined — not overlooked.** It would ship a
  30-line and a 21-line slice whose halves are the same concern: `load_preserved_state` reads the
  one-time-provisioned values back out of an existing `.env` and `prepare_directories` creates and
  chowns the data dirs (and calls `prepare_control_dirs` and `ensure_stratum_tls_cert`). Both run
  before `.env` is re-rendered; separating them names nothing new.
- `B1+B2 | B3` is legal and declined because it raises the new row from 485 to 516 for no gain and
  splits that same pair.

Parity passes for any cut at a single blank line between two top-level blocks, **including a
semantically wrong one**. Byte-identity does not validate assignment; only reading does. Two
judgement calls:

1. **The region ends after `prepare_directories`, not after `load_preserved_state`.** The cut map's
   P8 row prints artifact range `6048-6564` with a ~517 size; under the verified +271 coordinate
   shift that lands at remainder 517, which would hand `prepare_directories` to P9. I ended at 539
   instead, because the map's own row LABEL is "parse_and_validate_config + preserved-state/**dirs**"
   and P9's label opens with "release fetch" — which is `gh_release_fetch`, at remainder 593. The
   map's absolute ranges are known stale; its semantic boundaries are the sound half. **A reviewer
   who reads it the other way should say so — this is the call most worth attacking.**
2. **The two constant blocks between `prepare_directories` and `gh_release_fetch` stay in the
   remainder** — `CONTROL_SECRET_PATHS` (remainder 540–556) and the `CURL_CAP_*` trio (558–570).
   They are consumed by `render_masked_config`/`control_preview` and by the release-fetch and
   upgrade runners respectively, all of which are P9/P10 region. Taking them here would strand them
   away from their consumers for no reason.

**The +271 shift is now MEASURED at P8's start, not predicted.** The 28 slices ahead of the
remainder total 6290 lines; with 28 build separators the remainder's first line is artifact
**6319**, which is exactly the map's `6048 + 271`. Cross-checked against the artifact total:
6290 + 28 + 5966 = 12284 = the actual line count. It remains PREDICTED ONLY for P9–P13.

## Budget

- **New row:** `lib/pithead/28-parse-and-validate-config.sh` at **485** — its true line count, not
  a reserved-headroom value.
- **Lowered:** `lib/pithead/99-remainder.sh` **5966 → 5427**, as every P-series PR must.
- `29-preserved-state-and-dirs.sh` is 52 lines and takes no row.
- Row count 87 → 88; the tsv diff is exactly those three lines.
- `scripts/lint-file-budget.sh` rc **0**. `origin/develop-v2` probed at `2186a1f` immediately
  BEFORE and AFTER the gate run — it did not move under me, so this is not the stale-base shape
  that fails in both directions.

## What I did NOT do

- **No base/head suite pair.** It is a NULL EXPERIMENT for a slice cut and is deliberately omitted:
  `tests/stack/run.sh` consumes only the built artifact and references `lib/pithead/` zero times,
  and the artifact is the identical blob on both sides — so both legs would execute the same
  program and the result is guaranteed before launch. This is the ruled proof class, not a gap.
  For the same reason I did not check "are the moved functions present in the built artifact?" —
  the artifact is unchanged, so that check cannot fail.
- **`make lint` / `lint-sh` not run**, and no lint lock taken. A cut adds zero new shellcheck
  exposure (`pithead` is shellcheck's first argument and IS the concatenation; the slices are not
  separately analysed), so the only new exposure is `shfmt`, which is clean at 0 bytes. **CI's
  Shell tests job is the arbiter — check it before merging.**
- **No non-author review.** This lane authored the cut; it needs a non-author gate pass pinned to
  this head before merge.


VijitSingh97:
## Controller verdict: MERGE (controller w23, opus seat)

Non-author pass `reviewer-20260829T075500Z` returned **MERGE**, no blocking findings and no
should-fixes. I did not take that on report — below is what I re-derived myself, what I ruled, and
one number in the pass that is short (it concerns P9, not this PR).

### Re-derived at source, by me, before ruling

- **Head `5f6ead4`**, **merge-base `2186a1f`** — and merge-base *equals* the `develop-v2` tip. The
  intervening range is genuinely empty, proven by `merge-base == tip` rather than by an empty
  `git log`, so no interference is possible and I did not need the usual "range touches zero
  `lib/pithead/` files" control.
- **Artifact blob identical either side: `d79917f6…`** — the same blob P7 merged. Byte-identity is
  mine, not relayed.
- **Every boundary line position**, read out of the `2186a1f` remainder blob:
  `parse_and_validate_config` closes at **485**, blank **486**, comment 487-489,
  `load_preserved_state` **490-516**, blank **517**, `prepare_directories` **518-538**, blank
  **539**, `CONTROL_SECRET_PATHS` comment opens **540**, `gh_release_fetch` **593**.
- **The pure-move arithmetic is FORCED, not merely consistent.** 485+52 added, 539 removed, delta
  2. Those two are precisely blanks 486 and 539 — one per new slice edge — and the builder refuses
  a slice whose first or last line is blank, so the count could not have been anything else.
- **tsv**: one row removed (remainder 5966), two added (28-parse-and-validate-config 485,
  remainder 5427). The 52-line file correctly takes no row, under the 400 floor, consistent with
  P7's five-file precedent.

### Rulings

**1. The boundary call is RATIFIED** (I took this before the pass ran, and the pass reached it
independently). The map's printed range ends at 517, one line before `prepare_directories` —
which would strand "dirs" from a row whose own label promises it. Label wins, per the standing
stale-ranges ruling. This is the **second row running** where the printed range failed and the
label held: treat every printed range below P8 as advisory and the labels as load-bearing.

**2. The 485/52 split is RATIFIED, and on a stronger ground than the PR body gives.** The pass
found a real exposition gap: "smallest available row" rules out the 516 and 538 groupings but does
*not* separate two files from three, since B1|B2|B3 ties B1|B2+B3 at row=485. It is right that the
gap is real and right that the conclusion survives. **But the load-bearing justification is
neither the row size nor the naming — it is that the map's P8 row label already reads
"parse_and_validate_config + preserved-state/dirs".** The pairing was specified by the program
document at map time, before any cut-time budget objective existed, so it is not an artifact of the
smallest-row objective and never needed to be defended as one. Not a naming finding.

**3. Byte-identity AND parity together, not either alone.** Byte-identity is equally consistent
with a stale artifact nobody regenerated against the very sources this PR edits; only
`lint-pithead-parity` rebuilding from the current sources excludes that. The pass argued this
rather than citing it, and fired an **independently chosen** mutation (file 28 line 201, artifact
line 6519) rather than replaying the author's at 6837 — predicting the failure line by grep
*before* running the gate and proving the edit applied with `git diff --cached --numstat`.
Choosing your own mutation tests the instrument; replaying the author's only confirms their
control. It also caught its own `git checkout --` index trap mid-run and disclosed it.

### One number in the pass is short, and it is about P9

The pass says P9 "inherits these two constant blocks as its first 31 lines". **The inherited region
is 53 lines, 540-592, not 31.** Its enumeration stops at `CURL_CAP_OS_BUNDLE` (570) and does not
walk to 592. The remaining 571-592 is `gh_release_fetch`'s own 14-line docstring (572-585) plus its
output globals `GH_RELEASE_HINT`, `GH_RELEASE_JSON`, `GH_SOCKS` (586-592).

This changes nothing about P8 — the cut at 539 is confirmed correct by both of us independently —
and it improves the forward note rather than weakening it: **those trailing lines are not an orphan
prelude at all, they are `gh_release_fetch`'s documentation and its own outputs**, so P9's slice
naturally opens at 572 with the function it is named for. The appliance lane has already handled
this by construction; its slice is named `release-fetch-and-masked-config`.

I record it because I made the same class of error on this exact region last seat and was corrected
at source: sizing a slice by anchoring on its *last* function's header instead of the slice's own
first line. A wrong number attached to a right conclusion is dangerous precisely because nothing
about the conclusion flags it.

### Carried forward, not a finding here

The pass found a genuine gap in `bash -n` as a heredoc-boundary instrument: **an unterminated
heredoc that is the last construct in a file, with nothing else left unclosed, warns to stderr and
still returns rc 0.** It only hard-fails when the truncation leaves an enclosing construct open.
Moot for P8 (neither new file contains a heredoc — verified), but P10 deals in heredoc templates
and must not read "31 ok / 0 fail" as proof of heredoc integrity there.

### Gate

CI re-derived at the run's own `head_sha`, not from a checks summary: 14 success, 2 skipped,
**Shell tests still in progress** — the one gate outstanding, and the one the pass explicitly could
not check. Merging only on its completion.

Merge tree computed **before** merging: `88288c51…`. I will confirm the merge commit's tree equals
it and its parents are exactly (`2186a1f`, `5f6ead4`), so that what ships is provably what was
reviewed.

Issue #1482, opened 2026-08-29
Six mutating verbs and the in-place upgrade fallback still take no mutation lock (#1342 residual)


VijitSingh97:
PR #1391 serialised pithead's mutating windows behind the mutation lock, covering every racer #1342 names: the firstboot `(setup)` leg, the spooled-intent runner, and the operator verbs (`backup`, `apply`, `up`, `down`, `upgrade`, `restart`). Two residuals were disclosed in that PR's body as deliberate scope boundaries rather than fixed, and this issue is their tracking home so the disclosure does not evaporate at merge:

1. **Six mutating paths still take no lock, untested either way**: `factory_reset`, `config_reset`, `rotate_secrets`, `reset_dashboard`, `os_update`, and the installer-path mutations inside `firstboot_wizard` outside `(setup)`. Each can mutate `config.json` or adjacent state while another verb holds (or ignores) the lock.
2. **The in-place upgrade fallback overwrites the install dir before it re-invokes `pithead`, outside any window.** Pre-existing, named in #1391's body rather than left to be found.

Scope for whoever takes this: a per-verb ruling, not a blanket sweep — some of these are interactive or destructive verbs where a lock timeout needs the same careful exit-status routing #1391's F3 added for `setup` (a timeout must never be routed as a bad config). "Add the lock everywhere" without that routing would reintroduce the config-deletion class F3 closed.

Also carried from #1391's evidence section: "read-only verbs take nothing" is structurally true but not proven behaviourally by any test.

VijitSingh97:
## Per-verb ruling for the six unlocked paths

Derived at `cc05fad` (develop-v2 tip). **Line numbers below are artifact lines at that sha and go
stale on the next #1105 Phase 2 cut — the function names are the durable reference.** No code
change here; this is the ruling the issue asks for so it exists before someone reaches for a
blanket sweep. I did not implement it because a `pithead`-touching commit landing while #1105 P7
(PR #1507) is open would force that cut to be re-taken.

### The placement rule already exists in-source — it does not need inventing

#1391 wrote it into two of the verbs it locked, and both say it in their own comments:

- `backup`: *"AFTER both prompts (passphrase, and permission to stop the stack): the hold must not
  span an unbounded human wait. It runs from here across stack_down -> tar -> stack_up."*
- `restore`: *"After the confirm and the passphrase prompt, and after the integrity verify
  (read-only): the extraction below is the mutating window."*

[...]

Issue #1066, opened 2026-08-16
BLOCKER: the wizard's documented "No login" choice produces a config the appliance's own validator rejects — first boot dead-ends


VijitSingh97:
From the v2 gap analysis, reproduced against the real jq filter.

## The collision

- `apply_appliance_defaults` injects `dashboard.control.enabled = true` whenever the key is **absent** (`pithead:5147-5151`).
- The wizard writes `spool/config.json` through `strip_defaults` (`build/dashboard/mining_dashboard/wizard.py:105`), which drops any value equal to the reference default. `config.reference.json` has `dashboard.control.enabled: false`, so the key is **absent from every wizard submission** — including one that explicitly sets it false.
- On the "No login" answer, `ensure_appliance_dashboard_password` honours `spool/auth-mode == none` and leaves `dashboard.auth.password` empty (`pithead:5105-5108`).
- `parse_and_validate_config` then hard-errors on exactly that pair: *"dashboard.control.enabled is true but dashboard.auth.password is empty"* (`pithead:5740`).

So the appliance injects a default that its own validator forbids in combination with a choice the wizard offers.

## Why it is invisible until too late

`firstboot_consume_spool` validates the candidate **before** the defaults are injected. The submission passes validation, the operator is told provisioning is under way, and the failure lands afterwards — on a path that (see #1063) then re-serves the wizard over plain HTTP.

## Scope

Every operator who picks the documented "No login at all" option on first boot. That is a supported, documented choice — it is in the wizard and in `docs/appliance.md`.

## Reproduced

Running `apply_appliance_defaults`' jq filter over a config with an empty `dashboard.auth.password` emits `dashboard.control.enabled: true`, which is precisely the combination `pithead:5740` refuses. A full end-to-end confirmation is one KVM provision leg answering "No login" — which no current leg does.

## Fix

Make the default injection conditional on the auth choice: do not enable the control channel when the operator asked for no login, or refuse the "No login" option in the wizard with an explanation rather than accepting it and failing later. The second is arguably more honest — a control channel with no authentication is what `:5740` exists to prevent, and the wizard should not offer a combination the product will not run.

Either way, the validator should run against the **post-defaults** config before the operator is told provisioning started.

VijitSingh97:
Fixed in #1125.

[...]

PR #1509, opened 2026-08-29
refactor(tests): split provisioning, appliance-defaults, rauc-loop-wait and render-quadlet out of run.sh (#1105 R10)

VijitSingh97:
R10 of the #1105 Phase 1 `run.sh` split. Four contiguous blocks move into their own domain files.
`tests/stack/run.sh` goes **1034 → 812**.

| new file | lines | what moved | former run.sh range (at the base) |
|---|---|---|---|
| `tests/stack/test-render-quadlet.sh` | 55 | quadlet render parity vs the `os/quadlet` fixtures (#77 phase 1) | 124–152 |
| `tests/stack/test-control-provisioning.sh` | 150 | `provision_control_runner` ownership + install guard (#33, #1190) | 789–903 |
| `tests/stack/test-appliance-defaults.sh` | 96 | remote-node preflight + appliance defaults (#1066) | 908–970 |
| `tests/stack/test-rauc-loop-wait.sh` | 55 | `wait_loop_partitions` block-device + poll-budget contract | 999–1021 |

Every new file is under 400, so **no new file-budget rows**; the only tsv row that changes is
`run.sh`'s.

## Execution order is unchanged by construction, not by argument

Each block is contiguous, and each `source` stanza lands at that block's **exact former position**.
No block contains a `source "$HERE/` stanza, so nothing is nested and no domain loses its #1400
zero-assertion guard. Verified: 0 stanzas inside each of the four candidate ranges, against a
control of 39 across the whole file.

## The map's four ranges all overshot — all four were corrected before cutting

The cut map's END for each piece ran past the domain into the next one's introduction. Taken
literally, all four would have nested a domain file:

| piece | map range | last content line | what the tail would have nested |
|---|---|---|---|
| c14 | 1745–1863 | **1860** | `test-appliance-identity.sh` |
| c15 | 1864–1936 | **1927** | `test-appliance-install.sh`, `test-appliance-rig-miner.sh`, `test-appliance-boot.sh` |
| c18 | 2784–2815 | **2808** | `test-lifecycle.sh` — a file closed at its ceiling, so the repair would need a shrinking commit |
| render-quadlet | 124–~155 | **~154** | bounded by the stanza that must stay |

Those numbers are at the map's own sha. Every range was **re-located by section NAME at the tip**
and re-derived there before cutting; the ends above are the map's, the ends actually cut are the
tip's.

## Two judgement calls, stated rather than buried

**`test-appliance-defaults.sh` opens with a provisioning preflight, and its name does not say so.**
`preflight_remote_nodes` is thematically provisioning and belongs beside
`test-control-provisioning.sh`. It is here because contiguity outranks the label: the
`test-appliance-identity.sh` stanza sits *between* that section and the provisioning block, so
moving it would make the provisioning cut non-contiguous and break the order-preserving proof the
whole split rests on. Disclosed in the file's header. Precedent: Phase 2's
`21-doctor-stack-checks.sh` carrying three non-doctor helpers for the same reason.

**`test-rauc-loop-wait.sh` is the smallest domain file in the suite (55 lines, vs 67 for the
previous smallest), and that was preferred over the alternative.** The map suggested folding it
into the appliance-defaults family. The two blocks are separated by **nine** domain source
stanzas, so folding them would have reordered this section ahead of all nine. Riding an existing
neighbour was also unavailable: `test-appliance-media.sh` sits exactly at its recorded ceiling and
the budget ratchet only ever lowers. A zero-reorder cut was judged worth more than a round file
count.

## Proofs — every instrument fired a control first

**Byte-identity, both directions, derived from git rather than from the extraction.**
- Each new file below its header == `git show <base>:tests/stack/run.sh` at its stated range:
  **identical, 4/4**. Control: the same comparison with the range shifted one line **fires**.
- `run.sh` reconstructed independently by re-running the extractor over the base blob ==
  the committed `run.sh`: **identical**. Control: perturbing one line **fires**.

**Ordered section comparison (the only instrument that can see a permutation).** 274 headers both
sides; sorted multiset identical; **ordered diff identical; 0 of 274 positional mismatches**.
Control: permuting two adjacent headers makes the ordered diff **fire** while the sorted diff stays
**silent** — so the instrument discriminates.

**Whole-output residue: 2 lines**, a single pair — the #1325 ed25519 keygen fingerprint, which is
nondeterministic by construction. Control: deleting one assertion line grows the residue to 3.
(The normalizer also folds `/tmp/tmp.*` mktemp paths; without that, a second nondeterministic pair
sits in the residue and reads like a real difference.)

**Suite:** `rc=0, 2979 passed, 0 failed, 3292 log lines` — the same figures as the baseline.

**Baseline reuse, tested rather than assumed.** The BEFORE leg is R9's after-proof rather than a
fresh run. That is legal here because the executed program is byte-identical: the `tests/stack`
tree object and the built `pithead` blob are the **same objects** at R9's proven tree, at the tip
it merged to, and at this PR's base. The 15 paths that changed across that span are all under
`docs/`, `lib/pithead/` and `tests/integration/`; `git grep` finds **zero** references to
`lib/pithead` anywhere in `tests/stack/` (control: `./pithead` matches 30 times in `run.sh`), and
the three `file-budget` hits in `tests/stack/` are all comments. Both legs also ran in the **same
worktree**, so no cross-worktree path normalization is involved.

**Dependency audit, comments stripped, over the whole new files including headers.** `$ROOT`,
`$SANDBOX` and `$STACK` are the only ambient names read; each is assigned in `lib.sh` at column 1
outside every function body — re-derived by reading the indent, not taken from the tool, because
`-> PROVIDER` does not mean top-level. `test-appliance-defaults.sh` reads no ambient name at all
and so carries no guard line. The `(has ${..:-} default somewhere)` annotation the audit prints on
the guarded names is the known self-inflicted artifact of the `: "${NAME:?}"` guard itself;
confirmed by stripping the guard and re-auditing, at which point the annotation disappears.

**Coupling rules.** No block calls `build_control_sandbox` (0 in each; control: 1 in `run.sh`), and
none reads `$C`/`$RESULTS`/`$STAGED`/`$AUDIT`. Every scratch path these blocks write —
`$SANDBOX/quadlet-out`, `quadlet-local-out`, `quadlet-payout-out`, `loop-wait`, `pcr`, `pci` — was
swept across all of `tests/stack/` and appears **only inside its own block**. `$PFSB` is also used
by `test-appliance-install.sh`, which assigns its own `mktemp -d` before reading — a reused name,
not a shared value.

**#1400 guard set completeness.** 42 source stanzas, 42 `domain_ran` guards, and **zero stanzas
without one** — so the guarded set cannot omit a sourced domain by construction. This closes an
open question carried in from R9, which had proved the guards fire but not that the set omits
nothing.

**Gates.** `make lint` **rc=0**. File budget passes; the tsv diff is **exactly the `run.sh` row** —
`--generate` also wanted to lower `egress.py`, `ci.yml` and `tests/integration/run.sh`, rows this
diff never touches, and all three were restored by hand.

## `run.sh`'s ceiling is set to 815, not 812 — deliberately, and for a named consumer

`run.sh` is 812 lines; its recorded ceiling is **815**. That is three lines of headroom, the exact
size of one domain `source` stanza, and it is reserved for #1482.

The reason it can only be done here: `run.sh`'s ceiling has always equalled its exact size, so it
has zero headroom, so **a new domain file cannot be registered at all** — the three-line stanza a
new domain needs is itself refused by the budget gate. A new `tests/stack` file is therefore only
buildable inside a cut that is already rewriting `run.sh`'s row. The appliance lane hit exactly
this on #1482 (os-update mutation lock): both files that could host its coverage sit at their own
ceilings, and a new file was unbuildable. Rather than have it discovered at the gate, the slack is
planned in here and disclosed.

A stale-HIGH ceiling passes the gate by design (`:157` fails only when `lines -gt ceiling`), so
this is legal rather than merely tolerated. If #1482 does not land, the next cut that touches this
row should reclaim the three lines.

## Notes for the reviewer

- **The base moved under this branch mid-cut** (P7 / #1507 merged), and the budget gate flagged it
  the documented way: a NOTE naming `lib/pithead/99-remainder.sh`, a file this diff never touches.
  It was fixed by **rebasing**, not by an exemption, and every proof above was then re-derived at
  the new base rather than carried across. `tests/stack` is byte-identical either side of that
  move, so the cut itself was unaffected.
- **These four files ARE standalone-sourceable, which is the opposite of R9's three.** A reviewer
  arriving from R9 will reach for the spool-audit "deliberately not standalone" precedent; it is
  the wrong precedent for this shape. Those files are pure consumers of the shared control sandbox
  and are position-locked by what a sibling accumulated. These four build every fixture they need
  themselves and would assert the same thing run out of position.

## What was NOT done

- **`make test` did not complete.** It was killed by a harness timeout mid-run, and its `test-stack`
  target re-runs `tests/stack/run.sh`, which was already racing the detached after-proof. The
  `run.sh` leg is covered green by the after-proof above; the remaining legs — four standalone
  `tests/stack/test_*.sh` scripts, plus the dashboard and frontend suites — were not re-run. This
  diff touches no Python and none of those four scripts. CI covers them.
- No bench work, no hardware claimed, nothing under `os/` touched.
- **The base moved a second time during the proof** (P8 / #1508). The branch was rebased onto the
  new tip and re-pushed. The after-proof ran against the pre-rebase tree, and `tests/stack/` is
  **byte-identical** between that tree and the pushed head — the rebase moved only `lib/pithead/*`
  and the tsv row, neither of which the stack suite reads. So the proof is valid for what is
  pushed; that equivalence is the single claim here most worth attacking, and `git diff <proof-tree>
  HEAD -- tests/stack/` re-derives it in one command.

### DIFF
--- a/docs/dev/file-budget.tsv
+++ b/docs/dev/file-budget.tsv
@@ -63,7 +63,7 @@ tests/integration/selftest-e2e-phases.sh	507
 tests/integration/selftest.sh	686
 tests/os/failure-evidence.sh	567
 tests/os/run.sh	3437
-tests/stack/run.sh	1034
+tests/stack/run.sh	815
 tests/stack/test-appliance-boot.sh	641
 tests/stack/test-appliance-identity-boot.sh	465
 tests/stack/test-appliance-identity.sh	483
--- a/tests/stack/run.sh
+++ b/tests/stack/run.sh
@@ -121,35 +121,8 @@ _d0=$((PASS + FAIL)) && source "$HERE/test-cli.sh" && domain_ran test-cli.sh "$_
 # shellcheck source=tests/stack/test-config.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-config.sh" && domain_ran test-config.sh "$_d0" "$?" || domain_ran test-config.sh "$_d0" "$?"
 
-echo "== unit: render-quadlet parity vs os/quadlet fixtures (#77 phase 1) =="
-# The renderer must reproduce the spike-proven unit set byte-for-byte from the fixture env — the
-# os/quadlet files ran live in the #78 spike, so any drift here needs a bench re-proof, not just
-# an updated fixture.
-QOUT="$SANDBOX/quadlet-out"
-run_sourced "$SANDBOX" render_quadlet_units "$ROOT/os/quadlet/fixture.env" "$QOUT" >/dev/null
-for f in mining.network proxy.network tor.container p2pool.container xmrig-proxy.container \
-    caddy.container docker-proxy.container docker-control.container dashboard.container; do
-    assert_eq "quadlet parity: $f" "$(diff -u "$ROOT/os/quadlet/$f" "$QOUT/$f" 2>&1 | head -c 300)" ""
-done
-assert_eq "remote render emits no node units" "$(find "$QOUT" -name 'monerod.container' -o -name 'tari.container' | wc -l | tr -d ' ')" "0"
-# The local-node variant (bench-proven 2026-07-24): profiles on, 11 files, node units included.
-QLOCAL="$SANDBOX/quadlet-local-out"
-run_sourced "$SANDBOX" render_quadlet_units "$ROOT/os/quadlet/local/fixture.env" "$QLOCAL" >/dev/null
-for f in mining.network proxy.network tor.container monerod.container tari.container \
-    p2pool.container xmrig-proxy.container caddy.container docker-proxy.container \
-    docker-control.container dashboard.container; do
-    assert_eq "quadlet local parity: $f" "$(diff -u "$ROOT/os/quadlet/local/$f" "$QLOCAL/$f" 2>&1 | head -c 300)" ""
-done
-# The payout-confirm variant (bench-proven 2026-07-24): both wallet profiles, 13 files, the
-# dashboard gains the payout env keys only in this set (the others stay byte-identical).
-QPAY="$SANDBOX/quadlet-payout-out"
-run_sourced "$SANDBOX" render_quadlet_units "$ROOT/os/quadlet/payout/fixture.env" "$QPAY" >/dev/null
-for f in mining.network proxy.network tor.container monerod.container tari.container \
-    wallet-rpc.container tari-wallet.container p2pool.container xmrig-proxy.container \
-    caddy.container docker-proxy.container docker-control.container dashboard.container; do
-    assert_eq "quadlet payout parity: $f" "$(diff -u "$ROOT/os/quadlet/payout/$f" "$QPAY/$f" 2>&1 | head -c 300)" ""
-done
-assert_eq "local render emits no wallet units" "$(find "$QLOCAL" -name 'wallet-rpc.container' -o -name 'tari-wallet.container' | wc -l | tr -d ' ')" "0"
+# shellcheck source=tests/stack/test-render-quadlet.sh disable=SC2015
+_d0=$((PASS + FAIL)) && source "$HERE/test-render-quadlet.sh" && domain_ran test-render-quadlet.sh "$_d0" "$?" || domain_ran test-render-quadlet.sh "$_d0" "$?"
 
 # ---------------------------------------------------------------------------
 # shellcheck source=tests/stack/test-doctor-appliance.sh disable=SC2015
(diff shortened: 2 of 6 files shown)


VijitSingh97:
## Verdict: MERGE

Non-author adversarial pass `reviewer-20260829T082447Z` returned **MERGE**, no blocking findings, at the exact head `3611427f89cdfd18d2e975847e5ba26025eb1180`. Controller-judged; the ruling below on the budget headroom is mine, not the reviewer's, and was taken before the pass returned.

**What shipped is what was reviewed, proven two ways rather than asserted:** merge-base `08ad64cd096a00dacb9572018abfb9e70ca457f1` equals `origin/develop-v2` exactly (0 commits behind), and the merged tree is byte-identical to the tree computed by `git merge-tree --write-tree` BEFORE merging. The merge commit's parents are exactly (base, reviewed head).

### The headroom call — RULED LEGITIMATE, narrowly, and this is the precedent

Disclosure (a) sets `tests/stack/run.sh`'s tsv ceiling to 815 while the file is 812. Re-derived at source: base row 1034 with the file at 1034; this head's row 815 with the file at 812.

It is **mechanically legal** — `check_monotonic` enforces `new_ceiling == actual` only on a RISE and only for the single `monotonic_exempt` row; on a falling ceiling `run_gate` tests only `lines > ceiling` and `lines <= 400`, so 812 <= 815 <= 1034 has no equality rule to violate. The repo's one written anti-headroom sentence lives in that rise branch — *"a rise must RECORD the artifact, not grant it headroom"* — and does not transfer: its stated harm is retiring a per-PR **measurement**, which is what that row is. A ceiling on a real source file is not a measurement.

**Precedent was checked and it cut against the easy answer.** Three rows at tip carry headroom (`egress.py` +1, `ci.yml` +1, `tests/integration/run.sh` +10), but all three are incidental drift — the +10 row was set at the gate's founding commit `cc24394` and has never been edited since, so that file shrank under a static ceiling. **This PR is the first deliberate bank**, so it sets the precedent, stated narrowly on purpose:

1. the ceiling still FALLS substantially in the same PR (1034 -> 815, a 219-line tightening against 3 banked);
2. the headroom is the MINIMUM for one NAMED, already-filed change;
3. it is disclosed under its own PR heading;
4. the reclaim obligation is recorded in the program document, not only in a PR body.

Never legitimate on a rising or flat ceiling, and never as generic slack.

**The honest weakness, stated rather than argued away: the slack is anonymous.** Nothing binds those three lines to their intended consumer, and any PR may spend them. That is accepted because the same PR tightens the row by 219, and because the alternative is worse — with ceiling equal to size, a new domain stanza can comply only by simultaneously cutting unrelated lines out of `run.sh`, coupling every future change to a refactor.

### The three should-fixes: all corrected here, none force-pushed, and the rule rather than an exception

All three live in the **PR body**, which is the merge record, and none in durable in-repo documentation or in a file the suite sources. Where the wrong text lives is what decides the class, so these are corrected in this comment rather than by an amend that would re-stale this PR's own after-proof.

1. **Render-quadlet getting its own file is a deviation from the cut map's recommendation** (which said it should ride the appliance-defaults family) and the body does not disclose it, though it discloses the parallel rauc-loop-wait call in detail. The deviation is **correct** — the two ranges sit ~750 lines apart with dozens of stanzas between them, so folding them would force exactly the reorder this PR rejects elsewhere. Recorded in the map's R10 row.
2. **The body's "`./pithead` matches ~30 times in run.sh" does not reproduce.** Literal `./pithead` is 15; word-boundary 24; case-insensitive including prose 29. No method reaches 30. The claim's direction — heavy dependence on the built artifact versus zero on `lib/pithead/` — is independently re-derived and correct, so this is an illustrative figure that is wrong under a conclusion that is right. Noted precisely because that shape has no tell.
3. **The reclaim depended on a sentence in a PR body rather than anything that outlives it.** Closed before this merge: the obligation is now written into the cut map's R10 and R11 rows, with the trigger stated as the named change LANDING or DYING — never merely "the next cut happened". The follow-on cut has already exercised it correctly, carrying the bank forward rather than reclaiming it while its consumer is still open.

### Disclosure on the evidence framing

The split comparator's `VERDICT MULTISET` check is a **null instrument** — it greps `^(ok|bad|FAIL|PASS)`, which matches zero lines because verdict lines are two spaces, an ANSI escape, then the mark. Measured through the comparator's own normaliser: gate pattern 0 while real verdict lines are 2979 on the R9-class log and 1344 on the follow-on baseline. **This verdict therefore does not cite "IDENTICAL VERDICTS" as evidence anywhere**, and neither did the pass, which built its own header-order extractor and fired its own reorder mutation. The whole-output sorted multiset subsumes what that check was meant to prove, so the conclusions of this and prior cuts stand; what was wrong is the framing that counted it as an independent second check.

### What the pass attacked, so a clean result is not read as a shallow one

Both halves of the load-bearing rebase claim re-derived independently (`git diff` empty on `tests/stack/` across the rebase, with a 647-line control proving the tool discriminates); its own header-order extractor, 274 headers in identical order at both base and head; **its own independently chosen reorder mutation fired**, and the honest note that the ordered-header check is the *only* instrument that would catch a reorder here, not a backup to another; per-file byte-identity 4/4 with a one-line-shift control firing; all four files sourced standalone under the harness's real invocation mode, with a positive control dying on an unset `ROOT`; guard-set counts re-derived at 42/42/0 including a direct check of the nested-stanza blind spot; the CI job log pulled for this exact head rather than the checkmark, independently reproducing 2979/0 plus all four stack-tier legs; and the leak gate run with its control fired first.

**Declared unchecked, carried forward honestly:** no independent local full-suite run outside CI (a concurrent lane was mid-edit in the shared review worktree, which the pass detected and worked around by pinning an isolated worktree for the one working-tree-dependent check).

Issue #294, opened 2026-06-17
network.tor_egress_firewall=false never disables the #270 firewall (jq // false-coercion)


VijitSingh97:
## Bug

The documented #270 opt-out (`network.tor_egress_firewall=false`, per the comment in `pithead` and `docs/privacy.md`) **has never worked** — the firewall is always installed regardless of the config value.

### Root cause

`parse_and_validate_config()` read the toggle as:

```sh
jq -r '.network.tor_egress_firewall // true'
```

jq's `//` (alternative) operator treats **`false` as empty**, not just `null`. So `false // true` evaluates to `true`:

```console
$ echo '{"network":{"tor_egress_firewall":false}}' | jq -r '.network.tor_egress_firewall // true'
true
```

The rendered `.env` therefore always got `TOR_EGRESS_FIREWALL=true`, and `apply_tor_egress_firewall` always installed the rules.

The **same latent bug** existed on `.xvb.tor // true` (line ~1883): `xvb.tor=false` (route XvB donation over clearnet) was silently coerced back to Tor.

### Impact

- The #270 firewall opt-out is non-functional. Note the failure direction is **fail-safe** (firewall stays *on*), so no one was unexpectedly exposed — but the documented config knob is dead, which blocks legitimate uses (e.g. a deliberate clearnet posture, co-hosting scenarios, and the #256 benchmark's clearnet arm).
- `xvb.tor=false` can't take effect (privacy-safe direction — stays on Tor — but still broken).

### Fix

Null-check explicitly so a configured `false` is honoured (absent still defaults on / fail-closed):

```sh
jq -r 'if .network.tor_egress_firewall == null then true else .network.tor_egress_firewall end'
```

Fixed in commit d4b5df3 on `feat/256-benchmark-harness` (reaches `develop` via the #256 benchmark PR). Filing for tracking + so it's not lost if that PR is slow to merge. Surfaced while validating the #256 benchmark's clearnet arm (firewall stayed up → p2pool got 0 sidechain peers).

### Follow-up

- [ ] Add a regression test asserting `tor_egress_firewall=false` → `TOR_EGRESS_FIREWALL=false` in the rendered `.env` (and `=true` when absent/true).
- [ ] Audit for any other `<falsifiable-bool> // true` jq reads.

VijitSingh97:
Both follow-ups done on \`feat/256-benchmark-harness\` (reaches \`develop\` via #293):

[...]

Issue #137, opened 2026-06-04
doctor: verify Docker is enabled to start at boot (reboot resilience for unattended miners)


VijitSingh97:
## Summary
All 9 services use `restart: unless-stopped`, but that only re-launches containers if the Docker daemon itself starts at boot. Nothing verifies that. Both daemon checks in `pithead` only test reachability *now*, not boot-persistence; docs never mention reboot behavior.

## Evidence
- `docker-compose.yml` — `restart: unless-stopped` on all services (good).
- `pithead` — daemon checks test current reachability only; no `systemctl is-enabled docker`.
- `docs/` — no "will the stack come back after a reboot?" note.

## Impact
Ubuntu's `docker.io` apt package enables the service by default (happy path works), but `--skip-deps` installs, rootless Docker, Docker Desktop, or a host where the unit was disabled will silently **not** come back after a power loss / reboot — discovered only when the operator notices zero hashrate hours later. This is exactly the unattended-miner use case the hardware docs describe.

## Suggested fix
Add a `doctor` check (Linux): `systemctl is-enabled docker` → WARN with `sudo systemctl enable --now docker` if not enabled. Add a one-line "Reboot resilience" note to `docs/operations.md`.

## Related
Not #79 (external dead-man switch) — this is local boot-enablement. Pairs with the `doctor` exit-code fix in #127.

Issue #58, opened 2026-06-02
Dashboard: show the stack version on every screen (released version, or branch+hash in dev)


VijitSingh97:
## Goal
Always show **which version of the stack is running**, as a small badge in the dashboard header — visible on **both** the syncing screen and the main dashboard.

- **Released build:** show the released product version, e.g. `v1.3.0`.
- **Development build:** show a dev indicator — branch + short commit hash, e.g. `dev · feature/x @ a1b2c3d` — so it's unmistakable you're *not* on a release.

## Why
Nothing in the UI currently says what version is running (the only version lives in `build/dashboard/pyproject.toml`, unsurfaced). Users — and we, when debugging reports — can't tell a released build from a working branch. This is the "show the version" slice of the original #44, pulled out because it's **dashboard-only and shippable independently** of the release pipeline.

## Where it renders
The dashboard's shared top bar — `.header` in `mining_dashboard/web/templates/index.html` (line ~120), alongside the existing status badges produced by `build_header_badges()` in `mining_dashboard/web/views.py`. That header renders on **both**:
- the **syncing** screen (`build_sync_context`, `mode-sync`), and
- the **main** dashboard.

So a single badge covers both screens. Reuse the existing muted `badge-outline` style — informative, not loud.

## Version source + dev fallback
- **Released:** read the product version from the top-level `VERSION` file (the single source of truth defined in #44), baked into the image at build (build-arg → env / OCI `org.opencontainers.image.version` label) so the running container is self-describing.
- **Dev:** when there's no release version (working tree / local build), fall back to **branch + short hash**, captured at build (build-arg from `git rev-parse --short HEAD` + branch) or read at runtime — whichever is reliable inside the container.
- **Display rule:** a clean release shows `vX.Y.Z`; anything else shows the dev form, so a dev build is never mistaken for a release.

## Acceptance criteria
- [ ] A version badge appears in the header on **both** the syncing and main screens.
- [ ] A released build shows the released version (`vX.Y.Z`).
- [ ] A dev/working-tree build shows a clear dev indicator (branch + short hash), visibly distinct from a release.
- [ ] Source of truth is the top-level `VERSION` (no hardcoded version string in the template).
- [ ] Covered by a dashboard unit/component test.

## Related
- #44 — release & versioning structure (defines the `VERSION` source of truth this reads).
- #59 — "new version available" warning, builds on this badge.

PR #1511, opened 2026-08-29
refactor(tests): split the unit-helper cluster out of run.sh (#1105 R11)

VijitSingh97:
Splits the unit-helper cluster at the head of `tests/stack/run.sh` into
`tests/stack/test-unit-helpers.sh` (#1105 R11, W1 cut queue).

Five blocks move: `docker_boot_enabled` (#137), `config_bool` (#294), the XvB tier
docs-drift guard, the `env_get_file`/`env_changed_keys` helpers, and
`export_build_provenance` (#58). `run.sh` goes 812 -> 736 lines; the new file is 131
lines including its header, under the 400-line floor, so it takes no budget row.

## THIS CUT REORDERS — disclosed, measured, and argued

The cut range is **not contiguous**. Seven domain `source` stanzas are interleaved
between these blocks. Every one of them is **excluded** from the cut rather than nested:
nesting a stanza would leave the outer `domain_ran` guard reading the value the nested
stanza set, so the extracted content would carry **zero** zero-assertion coverage — not
narrowed coverage, zero.

Excluding them leaves three pieces, so one vacated position takes the source stanza and
the others close up. **The anchor is the last piece's position**, chosen because it
leaves the most sections standing.

- **Blocks that do not move at all:** the XvB drift guard, the env helpers and the
  provenance block execute exactly where they always did.
- **Blocks that move later:** `docker_boot_enabled` and `config_bool`.
- **Largest displacement:** `docker_boot_enabled` now runs after five domain files that
  used to sit after it — `test-control-upgrade.sh`, `test-release-signing.sh`,
  `test-dashboard.sh`, `test-dashboard-onion.sh`, `test-release.sh`. `config_bool` moves
  past the last three of those.

### Why the reorder is safe, argued in both directions

**Outward — what the moved blocks give the suite.** The names they assign (`$BOOT`,
`$tier_cfg`, `$tier_doc`, `$changed`, `$PROV`, `$NOVER`, `$ver`, and the drift loop's own
`$tier`/`$t_name`/`$t_rest`/`$t_val`/`$t_human`) are read nowhere else in `run.sh` or
anywhere else under `tests/stack/`, with comments stripped before counting. The one
shared name is `$CB`: the `config_bool` block and `tests/stack/test-monero-tari.sh` both
work in `$SANDBOX/cb`. That file assigns `$CB`, `mkdir -p`s it and writes its own
`config.json` — its own header records that it deliberately stopped depending on this
block — and its stanza is sourced after this file's stanza both before and after the cut.

**Inward — what the moved blocks take from the suite.** They read only `$SANDBOX`,
`$ROOT` and `$STACK`. All three are assigned in `lib.sh` at **column 1 at top level**,
outside every function body (re-derived at source, with a control confirming none of the
three is also assigned indented) — so none is the ordering dependency the `$WALLET` case
turned out to be. The five domain files the two blocks now run after:

- reassign none of `$SANDBOX`, `$ROOT`, `$STACK`;
- create **no** `systemctl` stub that could shadow this file's own (and the block
  *prepends* its stub directory to `PATH`, so it would win regardless);
- confine every `PATH` replacement and every `cd` to a subshell or to a function whose
  entire body is a subshell — including the one `PATH` assignment that *replaces* rather
  than prepends, which is inside a command substitution's subshell.

Every path this file reads or writes is anchored on `$SANDBOX`, `$ROOT` or `$STACK`, or
is reached by an explicit `cd` inside a command substitution, so a leaked working
directory could not reach them either.

## Naming — a deliberate departure from the cut map

The cut map files this range as `test-config-helpers.sh`. The file is named
`test-unit-helpers.sh` instead: of the five blocks only `config_bool` is a config helper,
the XvB block is a docs-drift guard rather than a helper test, and
`tests/stack/test-config.sh` already exists — a near-identical neighbour name would send
a reader to the wrong file. Same reasoning the R10 row applied to its loop-wait section.

## Budget — R10's headroom is carried forward, not reclaimed

**This cut CARRIES R10's three-line bank FORWARD. It does not reclaim it, and saying so is
a condition of the ruling that created it, not a footnote.** `run.sh`'s row goes 815 ->
**739** for a 736-line file. R10 banked three lines because `run.sh`'s ceiling has always
equalled its exact size, so the row has zero headroom, so the three-line `source` stanza a
new domain file needs is itself refused by the budget gate — which is why a new
`tests/stack` file can only be born inside a cut that is already rewriting that row.

**#1482 is still live** (PR #1510 open and unmerged at the time of this cut), so the bank
still has its beneficiary and is carried forward unchanged. **The trigger to reclaim is
the named change LANDING or DYING — never "the next cut happened".** A cut that reclaims a
live bank has simply re-imposed the coupling the ruling removed. **The obligation to
disclose the carry-forward passes to R12**, or the three lines decay into anonymous slack
that the next reader cannot tell from an error.

`--generate` would also have silently lowered three rows this cut never touches
(`dashboard/mining_dashboard/service/egress.py` 424->423, `.github/workflows/ci.yml`
513->512, `tests/integration/run.sh` 2551->2541). All three are **restored to their
committed values**; lowering a row reds every un-rebased branch in every lane, naming a
file their diff never touched, and a stale-HIGH ceiling passes the gate by design.

## Proof

**Byte-identity, re-derived from git rather than from the extraction, both directions,
each with a control that provably fires.**

- The new file below its header, rebuilt out of `git show <base>:tests/stack/run.sh` at
  the three stated ranges, is byte-identical to what is committed. Control: shifting one
  range by a single line makes the diff fire.
- `run.sh` rebuilt independently by re-running the extractor over the base blob is
  byte-identical to the committed `run.sh`. Control: perturbing one line of the
  reconstruction makes the diff fire.

**Guard-set closure re-counted after the cut**: 43 domain `source` stanzas, 43
`domain_ran` guards, **zero** stanzas without one — the set cannot omit a sourced domain
by construction. (R10 closed this at 42/42; this cut adds one of each.)

**The ordering claim in the header was checked in both trees, not asserted**:
post-cut, `test-unit-helpers.sh`'s stanza is sourced before `test-monero-tari.sh`'s;
pre-cut, the `config_bool` block ran before that same stanza. The `$SANDBOX/cb`
relationship is therefore unchanged in direction.

**Static checks on the new file**: `shfmt -i 4 -d` clean, `shellcheck --severity=warning`
clean.

**Name audit** (`split-name-audit.py`, self-test 21/21 first) over the whole file,
header included. Three results, all adjudicated rather than accepted:

- `$ROOT`, `$SANDBOX`, `$STACK` annotated `PROVIDER (has ${..:-} default somewhere)`.
  That annotation is **self-inflicted by this file's own `: "${NAME:?}"` guard** — a
  known limitation where the tool's regex counts a `${..:?}` expansion as a defaulted
  one. Isolated: stripping the guard line and re-auditing makes the suffix disappear.
  Re-derived from source anyway — all three are assigned in `lib.sh` at column 1, and a
  control confirms none is also assigned indented inside a function body.
- `PITHEAD_VERSION` read, assigned nowhere. Correct and not a coupling: it is set by
  `export_build_provenance` itself, inside the command substitution that sources the
  script under test. Same shape it had in `run.sh`.
- `ORPHAN-CALL export_build_provenance`. The tool's known over-report: it is a pithead
  function reached after `source "$STACK"` inside that substitution, not a missing name.

## Instrument defect found while proving this cut — disclosed, not worked around

`~/split-tools/compare-proof.sh`'s **"VERDICT MULTISET" check is vacuous**. It compares
`grep -aE '^(ok|bad|FAIL|PASS)'` on both logs, but the suite writes per-test verdicts as
`  ✓ …` / `  ✗ …`, so that pattern matches **zero** lines on either side and the check
prints "IDENTICAL VERDICTS" unconditionally. Measured on this cut's own baseline log:
`^(ok|bad|FAIL|PASS)` → 0 lines, `^  [✓✗]` → the full verdict set.

This does not weaken R7–R10: the whole-output sorted multiset diff covers every log line
including the verdicts, so it subsumes what that check was meant to prove. What was wrong
is the framing — the proof standard names header multiset and per-test verdicts as two
checks, and one of the two could not fail. **The correct `^  [✓✗]` comparison is run and
reported below alongside the comparator's own output.** The tool is not rebuilt here;
the fix is routed to the controller.

## Suite pair — run at this cut's exact base, both legs, in parallel

**I did not reuse R10's after-proof as this cut's baseline, and the reason is worth
recording**: the tree-object test failed. `~/split-baselines/r10-2186a1f/`'s `tests/stack`
tree object is identical to *pre-R10* develop-v2, so that directory is named for R10's
**base**, not its proof tree, and the reuse could not be licensed from git. R10's own PR
discloses that its after-proof ran on a pre-rebase tree and names that equivalence as its
weakest claim. So this cut ran its own BEFORE leg at its own base instead.

| | before | after |
|---|---|---|
| exit code | 0 | 0 |
| verdict | 2979 passed, 0 failed | 2979 passed, 0 failed |
| log lines | 3292 | 3292 |
| `^  ✗` (post-norm) | 0 | 0 |

**Whole-output sorted multiset residue: exactly 2 lines** — one #1325 ed25519 keygen
fingerprint pair, the known nondeterministic class. Nothing else.

**Header SET (sorted): IDENTICAL.**

**Header ORDER — the check this cut turns on.** It fires, by design:

```
21d20   < == unit: docker_boot_enabled (#137) ==
34d32   < == unit: config_bool honours an explicit false (#294) ==
57a56,57 > == unit: docker_boot_enabled (#137) ==
         > == unit: config_bool honours an explicit false (#294) ==
37 of 274 positions differ
```

**Exactly two headers move**, and they are the two the anchor choice predicted.
Displacement: `docker_boot_enabled` +35 header positions, `config_bool` +23. The 37
differing *positions* are the arithmetic consequence of two blocks leaving early slots —
every header between the old and new sites shifts by two — not 37 relocated sections.

**Permutation control, fired.** Built from the before log against itself with exactly two
adjacent headers swapped, so content is identical and only order differs:

- sorted header diff: **SILENT** — blind to the permutation, as claimed
- ordered header diff on the same input: **FIRES**

That is the proof the instrument discriminates, and it is the only one of the two checks
that noticed. For a cut with no ambient coupling, the ordered comparison is not a backup
to the multiset — it is the only thing that can see a reorder at all.

**Verdict multiset (post-norm `^  [✓✗]`, 2979 lines each side): residue is the same single
keygen pair.** This check is reported here for the first time by an instrument that can
actually fail — see the section above.

## Rebase — the proof is not stale, and that is checkable

This branch was cut on R10's head and rebased onto develop-v2 after #1509 merged. **The
rebased commit's tree object is byte-identical to the pre-rebase commit's tree** — the
whole tree, not just `tests/stack`. develop-v2's `tests/stack` tree object and `pithead`
blob were already identical to the cut's base, so the merge changed nothing the suite
executes and the measured pair remains valid for exactly what is pushed.

## What was NOT run

- **`make lint` and `make test` were not run locally.** The session was over its context
  bar and the controller bounded the seat to finishing the proof and opening the PR;
  `make lint-sh` is the serialized shellcheck path with a history of OOM-killing sessions
  on this box, and it was not worth a fresh crossing of that bar. CI runs both on a fresh
  checkout and is the backstop. What *was* run on the new file: `shfmt -i 4 -d` clean and
  `shellcheck --severity=warning` clean.
- **No adversarial non-author review.** This lane authored the cut; the pass and the merge
  belong to the controller seat.

## Header amended after the proof — disclosed, bounded, and being re-proven

The over-engineering pass on this branch found a real defect in the header and it was
fixed before opening the PR. The original sentence said the two moved blocks run past
"the release-signing, dashboard and release domain files". That was wrong twice: it named
three files when `docker_boot_enabled` crosses **five**, and it conflated the two blocks,
which cross **different sets**. Corrected to state each block's own set. Two weaker claims
were tightened in the same amend: the outward name sweep now says what was actually done
(per name over every `tests/stack/*.sh` with comments stripped first, surviving hits read
individually rather than counted), and the drift loop's own `$tier`/`$t_*` names — which
the header asserted were read nowhere else without my ever having swept them — were then
actually swept, and the claim holds at zero.

**The amend is comment-only, proven**: zero non-comment lines differ between the pre-amend
and post-amend file, with a control confirming the instrument sees the 11 changed comment
lines. The body below the header is still byte-identical to the base run.sh at the stated
ranges. `shfmt` and `shellcheck` re-run clean.

**Even so, a comment-only amend to a file the suite SOURCES re-stales that PR's own
after-proof** (the R9 precedent, which ran a third proof for exactly this). The pair above
was measured on the pre-amend tree. **A re-proof of the after leg is running now**; its
verdict lands at `~/split-baselines/r11-3611427/after2.{log,rc}` and should reproduce
rc=0 / 2979 passed / 0 failed / 3292 lines / 274 headers. **Do not merge on the pair above
alone — read that verdict first.** The before leg is unaffected: its base did not move.

### DIFF
--- a/docs/dev/file-budget.tsv
+++ b/docs/dev/file-budget.tsv
@@ -63,7 +63,7 @@ tests/integration/selftest-e2e-phases.sh	507
 tests/integration/selftest.sh	686
 tests/os/failure-evidence.sh	567
 tests/os/run.sh	3437
-tests/stack/run.sh	815
+tests/stack/run.sh	739
 tests/stack/test-appliance-boot.sh	641
 tests/stack/test-appliance-identity-boot.sh	465
 tests/stack/test-appliance-identity.sh	483
--- a/tests/stack/run.sh
+++ b/tests/stack/run.sh
@@ -18,49 +18,12 @@ _d0=$((PASS + FAIL)) && source "$HERE/test-harness-tooling.sh" && domain_ran tes
 # shellcheck source=tests/stack/test-doctor.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-doctor.sh" && domain_ran test-doctor.sh "$_d0" "$?" || domain_ran test-doctor.sh "$_d0" "$?"
 
-echo "== unit: docker_boot_enabled (#137) =="
-# A systemctl stub on PATH; FAKE_BOOT picks which unit reports "enabled". Docker counts as
-# boot-enabled if EITHER docker.service or docker.socket is enabled.
-BOOT="$SANDBOX/boot"
-mkdir -p "$BOOT/bin"
-cat >"$BOOT/bin/systemctl" <<'EOF'
-#!/usr/bin/env bash
-case "$1 $2" in
-  "is-enabled docker.service") [ "${FAKE_BOOT:-}" = "service" ] && exit 0 || exit 1 ;;
-  "is-enabled docker.socket")  [ "${FAKE_BOOT:-}" = "socket"  ] && exit 0 || exit 1 ;;
-  *) exit 1 ;;
-esac
-EOF
-chmod +x "$BOOT/bin/systemctl"
-PATH="$BOOT/bin:$PATH" FAKE_BOOT=service run_sourced "$SANDBOX" docker_boot_enabled
-assert_rc "docker.service enabled -> 0" "$?" "0"
-PATH="$BOOT/bin:$PATH" FAKE_BOOT=socket run_sourced "$SANDBOX" docker_boot_enabled
-assert_rc "docker.socket enabled -> 0" "$?" "0"
-PATH="$BOOT/bin:$PATH" FAKE_BOOT=none run_sourced "$SANDBOX" docker_boot_enabled
-assert_rc "neither enabled -> 1" "$?" "1"
-
 # ---------------------------------------------------------------------------
 # shellcheck source=tests/stack/test-control-upgrade.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-control-upgrade.sh" && domain_ran test-control-upgrade.sh "$_d0" "$?" || domain_ran test-control-upgrade.sh "$_d0" "$?"
 
 # shellcheck source=tests/stack/test-release-signing.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-release-signing.sh" && domain_ran test-release-signing.sh "$_d0" "$?" || domain_ran test-release-signing.sh "$_d0" "$?"
-
-echo "== unit: config_bool honours an explicit false (jq // false-coercion guard, #294) =="
-# Regression for #294: `.x // true` returns true even when x is explicitly false (jq treats false as
-# empty), which silently broke the #270 firewall opt-out (config false → .env stayed true) and
-# xvb.tor=false. config_bool null-checks instead. CONFIG_FILE is the relative "config.json", so a
-# fixture in the cwd is what the sourced helper reads.
-CB="$SANDBOX/cb"
-mkdir -p "$CB"
-printf '{"network":{"tor_egress_firewall":false},"xvb":{"tor":false}}' >"$CB/config.json"
-assert_eq "explicit false honoured (firewall)" "$(run_sourced "$CB" config_bool '.network.tor_egress_firewall' true)" "false"
-assert_eq "explicit false honoured (xvb.tor)" "$(run_sourced "$CB" config_bool '.xvb.tor' true)" "false"
-printf '{"network":{"tor_egress_firewall":true}}' >"$CB/config.json"
-assert_eq "explicit true honoured" "$(run_sourced "$CB" config_bool '.network.tor_egress_firewall' true)" "true"
-printf '{}' >"$CB/config.json"
-assert_eq "absent -> default true" "$(run_sourced "$CB" config_bool '.network.tor_egress_firewall' true)" "true"
-assert_eq "absent -> default false" "$(run_sourced "$CB" config_bool '.xvb.tor' false)" "false"
 
 # ---------------------------------------------------------------------------
 # shellcheck source=tests/stack/test-dashboard.sh disable=SC2015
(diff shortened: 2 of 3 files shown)


VijitSingh97:
**Verdict: MERGE.** Non-author pass `reviewer-20260829T092925Z` returned MERGE-WITH-SHOULD-FIX. Its one finding is real and is discharged by this comment rather than by a force-push; the reason for that choice is stated below.

## The finding, and why this comment is the fix

This body's closing section says a comment-only amend re-staled the measured suite pair, that a re-proof was running, and ends: *"Do not merge on the pair above alone — read that verdict first."* **That verdict was never folded back in.** Re-derived timestamps: the stale `after` leg finished 08:46:34Z, the amend landed 08:49:41Z, this PR opened 08:50:00Z off body text written at 08:49:45Z — and `after2` did not finish until **09:02:44Z, twelve minutes after the PR existed**. No comment, no review, and `updatedAt` still equals `createdAt`. So as it stands the PR instructs its own merger not to merge on the evidence it shows, and gives no way to discharge that from GitHub.

**The evidence is sound; it is the merge record that was incomplete.** That distinction decides the class. Wrong or missing text inside a file the suite SOURCES re-stales that PR's own proof and earns a discharge with a fresh after-proof. Text in a PR body is the merge record — a comment corrects it, and nothing a comment says can move a suite result. This is the latter, so it is corrected here.

## Discharging the gate — the re-proof, in the past tense, run independently

I did not take these from the lane, which no longer exists. I read `after2` at source myself:

- **rc=0, 2979 passed / 0 failed, 3292 output lines, 274 section headers** — and all three legs (`before`, `after`, `after2`) agree on every one of those four figures.
- **The whole-output sorted multiset differs by 18 lines, and every one is accounted for**: the worktree path the two legs ran under, plus two known-nondeterministic values (an ed25519 keygen fingerprint and an mktemp path). **Content difference is zero.** That is stronger than this body claims.
- **Header set identical** (multiset diff 0 across 274), **order changed in exactly the shape described**: two blocks move forward and the headers they pass shift back, giving 37 differing positions out of 274.

The re-proof reproduces what the body predicted it would. The gate the body set is met.

## An independent check on losslessness, by a different route

The pass reconstructed the removed lines from git's own diff and matched them byte-for-byte against the new file, 78/78, with a fired control. I ran an unrelated route on the merged tree: of the new file's 59 substantive lines, **58 are present in the pre-cut `run.sh`** (the control, firing high, which is what makes the move real) and **only 4 remain in the post-cut one** — and those 4 are `done`, `else`, `esac`, `fi`. Pure shell syntax, not content. Nothing lost, nothing duplicated, by two routes sharing no machinery.

## Merge mechanics, verified on the merged tree rather than predicted

Merged tree computed before merging; budget rows and artifact blob read from it. `tests/stack/run.sh` sits at **736 lines against a 739 ceiling** — the three banked lines carried forward, disclosed, and correct: a falling ceiling has no equality rule, and the beneficiary is still live (see below). `lib/pithead/99-remainder.sh` is untouched at 5440/5440. The shipped artifact blob is **unchanged through the merge**, which is the stronger form of the claim this proof class rests on and is what a tests-only cut must show.

## The reclaim trap, recorded because the next cut will meet it

The three banked lines are reclaimed when their named change lands or dies. **#1510 merging does not fire that.** Re-derived: #1482 is still **OPEN** — it names six mutating verbs and the in-place upgrade fallback, and #1510 locked exactly one. #1510 also touches no file under `tests/`, and `test-appliance-os-update-verbs.sh` already carries its own source stanza, so the contention regression test now bound to the appliance lane consumes no stanza either. The bank still waits on a new test file for the remaining verbs. **The general shape: a reclaim trigger phrased as "the named change landed" goes ambiguous the moment the named change is an issue that ships in slices — a partial landing is indistinguishable from a full one in the merge record. Check the issue, not the PR.**

## Carried forward, because it bites two cuts from now

The pass confirmed independently that **`bash -n` does not report an unterminated heredoc through its exit code** — rc=0 either way; it fires only through stderr text. That reproduces a finding carried out of an earlier cut in this program, now by a second route on different files. It is moot here (this cut's files hold no heredocs) and decisive at the cut that deals in heredoc templates: a clean `bash -n` sweep read by `$?` alone is not proof of heredoc integrity there.

## Ratified, not re-raised

The null suite-pair-over-a-byte-identical-artifact class; the banked headroom; `make lint`/`make test` not run locally, which was ruled; the by-hand review-gate workaround.

## What nobody closed

End-to-end byte-identity of the literal pre-rebase commit object was not reproduced — the pass established the structural equivalent (the intervening merge's tree equals the prior cut's own tree, so nothing outside this diff could have moved) but could not locate the pre-rebase object to diff directly. `shfmt` and `shellcheck` on the new file are relayed from this body and unverified; `make lint-sh` was deliberately not run, as it is the known session-killing path. No CI run was inspected by the pass.

PR #1512, opened 2026-08-29
refactor(cli): cut release fetch, provisioning and render_env out of the remainder (#1105 P9)

VijitSingh97:
Phase 2 P9 of #1105. Four order-preserving contiguous slices come out of the head of
`lib/pithead/99-remainder.sh`. **The built `pithead` artifact is unchanged, byte for byte.**

## Rebased onto `develop-v2` after #1510, and re-proven there

This cut was made on P8's head. #1510 (#1482) then merged, adding 9 lines to
`lib/pithead/15-os-update.sh` and 13 to `control_os_install` in the remainder, and moving the shipped
artifact. **A rebase invalidates the whole proof set, not just the suite, so every proof below was
re-run at the rebased head** — no figure here is carried from the pre-rebase run.

The rebase conflicted in exactly one place, `docs/dev/file-budget.tsv`: both sides edit the
`lib/pithead/99-remainder.sh` row. Resolved by re-deriving the count with the gate's own
`awk 'END{print NR+0}'` rather than by arithmetic — **4257**, which the arithmetic
(4244 + #1482's 13) independently agrees with.

Two checks that the rebase preserved the cut, neither of which the parity gate can make:

- P9's four slices, concatenated, equal `b338bdf:lib/pithead/99-remainder.sh` lines **1–1183**
  modulo the four boundary blanks the build re-inserts.
- P9's remainder is byte-identical (`cmp`) to `b338bdf:lib/pithead/99-remainder.sh` from line
  **1184** to EOF — so #1482's `PITHEAD_EX_LOCK_TIMEOUT` branch survived intact, at
  `99-remainder.sh:3745`.

## The cut

Region: `99-remainder.sh` lines **1–1183** at the rebased base `b338bdf`, derived by CONTENT. It opens at the
`CONTROL_SECRET_PATHS` comment block and ends at the single blank line 1183, between `render_env`'s
closing `}` and `inject_service_configs() {` — which is P10's opener.

| slice | lines | contents |
|---|---|---|
| `lib/pithead/30-release-fetch-and-masked-config.sh` | 157 | `CONTROL_SECRET_PATHS`, the `CURL_CAP_*` trio, the `GH_*` globals, `gh_release_fetch`, `render_masked_config` |
| `lib/pithead/31-directories-and-dashboard-state.sh` | 214 | `prepare_control_dirs` … `resolve_dashboard_host` |
| `lib/pithead/32-onion-provisioning.sh` | 314 | `wait_for_onion` … `rotate_secrets` |
| `lib/pithead/33-render-env.sh` | 494 | `render_env` |

Remainder **5440 → 4257**. Slice count 31 → 35.

**The cut map's P9 label reads "release fetch + provisioning + render_env", but the region does not
OPEN with release fetch** — 53 lines of constants sit ahead of `gh_release_fetch`. They are not
orphaned: `CONTROL_SECRET_PATHS` is `render_masked_config`'s own constant and the `CURL_CAP_*` trio
is `gh_release_fetch`'s, so both sit with their consumers and slice 30's name covers them.

`git rev-parse HEAD:pithead` = `d537802802865addfa73ab75f8d68513c5a50cfd`, identical to
`origin/develop-v2:pithead`. And a literal `cmp` of `git show <base>:pithead` against
`git show <head>:pithead`: identical at **713935 bytes**.

### 2. The parity gate — the claim that matters

Byte-identity ALONE is equally consistent with *"nobody regenerated the artifact and it is now stale
against the very sources this PR edits"*, which produces the same identical blob.
`make lint-pithead-parity` excludes that; only (1)+(2) together say the artifact is unchanged AND is
genuinely the product of the edited sources.

- **Clean:** `--self-test` then `--check` → `pithead parity OK — the committed artifact is exactly
  what 35 slice(s) build.` rc 0.
- **Fired, line predicted first.** `grep -n 'Single writer for the env file' pithead` → **7555**,
  run before the gate. Staged `33-render-env.sh`, mutated it, proved the mutation applied with
  `git diff --numstat` on the tracked file (`1 1`) — numstat on an untracked file prints nothing and
  would "prove" a mutation that never landed. Gate failed with `7555c7555`, quoting both sides.
  Predicted and reported line converged by two independent routes.

### 3. The heredoc control — the instrument byte-identity is blind to

A boundary falling inside a heredoc still reproduces the artifact byte for byte, because
concatenation puts it back. `render_env` writes the whole `.env` from a heredoc, so this is the real
hazard on this cut. Fired deliberately: corrupting the `EOF` delimiter at `33-render-env.sh:351`
made `bash -n` return **rc 2** at the file's closing brace. Restored, then `bash -n` over all 35
slices individually: **35 ok / 0 fail.**

`shfmt -i 4 -d` over the four new slices and the new remainder: **0 bytes.** First and last lines of
every touched file are non-blank, which is the build's own refusal condition.

Every one of the 35 slices was compared against its span of the artifact in place: `30-` at
**6867–7023**, `31-` at **7025–7238**, `32-` at **7240–7553**, `33-` at **7555–8048**, `99-` from
**8050** to EOF (**12306**), with build separators at 7024, 7239, 7554, 8049. All 35 spans matched
byte for byte.

### 5. The blank-line accounting (why lines-removed exceeds lines-added)

157+214+314+494 = **1179 added**; 5440 − 4257 = **1183 removed**. The difference is exactly **4**:
three interior boundaries and one seam. The build re-inserts one blank line per boundary, so those
four blanks are removed from the sources and restored by the concatenation. **Their disappearance is
REQUIRED for byte identity — it is not lost content.**

Pure-move multiset diff over the staged `lib/pithead/` diff: added ∖ removed **empty**; removed ∖
added **exactly four blank lines**. Sides extracted with `awk '/^--- /{next} /^-/{print
substr($0,2)}'`, not `grep '^-[^-]'`, which drops removed blanks and makes a pure move read as a
content change.

The ponytail gate reads the branch name from the pinned shell cwd rather than the worktree the
command ran in, so it names an unrelated branch and lets the retry through without checking that a
review happened. The mismatch does not mean the diff went unreviewed — the pass's two useful halves
for a P-series cut are run here:

1. **Pure move?** Yes — added ∖ removed is empty. Not one of the 1179 added lines is new logic.
2. **File count minimal?** Enumerated below, as arithmetic rather than judgement.

## The split degree, and the coarser cut that was available

`render_env` is **491 lines as one function**, so its block (494 with its comment header) is
indivisible and over the 400-line target: it takes a row under every possible grouping. The
remaining 687 lines are over 400 too, so they need at least two files. **The minimum is therefore
three files and exactly one new row.**

Legal cut points (single blank lines between top-level blocks) in the pre-`render_env` stretch:
32, 114, 158, 182, 197, 208, 223, 243, 268, 285, 319, 342, 373, 389, 411, 436, 442, 485, 505, 523,
525, 539, 591, 688. For a two-file split of 1–687 with both halves ≤400 the cut must fall in
[287, 401], leaving **319, 342, 373, 389** as the only candidates.

| cut | files | new rows |
|---|---|---|
| 1–372 \| 374–687 \| 689–1182 | 372 / 314 / 494 | 1 (at 494) |
| **1–157 \| 159–372 \| 374–687 \| 689–1182** | **157 / 214 / 314 / 494** | **1 (at 494)** |

**Three files was available, legal, and is declined on naming — not overlooked.** Its 372-line first
file would have to cover release fetch, control-channel masking, control dirs, data dirs, ownership,
the stratum TLS cert, dashboard-data migration and dashboard host resolution under one name. Four
files costs no extra row and gives each file a name that is true of all of it. The extra boundary is
the only price, and it is one blank line.

Parity passes for any cut at a single blank line between two top-level blocks, **including a
semantically wrong one**. Byte-identity does not validate assignment; only reading does.

1. **The constants lead slice 30 rather than trailing P8.** `CONTROL_SECRET_PATHS` is consumed by
   `render_masked_config` (in this slice) and `control_preview` (still in the remainder); the
   `CURL_CAP_*` trio by `gh_release_fetch` (this slice) and the upgrade runners (remainder). Each
   sits with its nearest consumer, and both are named by slice 30's name.
2. **`resolve_dashboard_host` closes slice 31 rather than opening 32.** It resolves the host for the
   dashboard's own state and predates the onion work; the onion group is opened by `wait_for_onion`,
   which is the first function every onion path calls.
3. **`rotate_secrets` closes slice 32.** It rotates the onion keys among other secrets and its only
   contiguous alternative is opening slice 33, which holds `render_env` alone.

## Budget

- **New row:** `lib/pithead/33-render-env.sh` at **494**, its true line count, not a
  reserved-headroom value.
- **Lowered:** `lib/pithead/99-remainder.sh` **5440 → 4257**.
- The other three slices are under target and take no row. Row count 88 → 89.
- `scripts/lint-file-budget.sh` rc **0**, with `origin/develop-v2` probed unmoved immediately before
  and after — not the stale-base shape that fails in both directions.

## What I did NOT do

- **No base/head suite pair.** A null experiment for a slice cut: `tests/stack/run.sh` consumes only
  the built artifact and references `lib/pithead/` zero times, and the artifact is the identical blob
  on both sides, so both legs would execute the same program. Stated up front, not a gap. Same reason
  I did not check whether the moved functions appear in the built artifact — it is unchanged, so that
  check cannot fail.
- **`make lint` / `lint-sh` not run**, no lint lock taken. A cut adds zero new shellcheck exposure
  (`pithead` is shellcheck's first argument and IS the concatenation); the only new exposure is
  `shfmt`, clean at 0 bytes. **CI's Shell tests job is the arbiter.**
- **No non-author review.** This lane authored the cut.


VijitSingh97:
## Controller verdict: MERGE

Non-author pass `reviewer-20260829T094931Z` returned **MERGE with no surviving findings**, and it is a deep pass rather than a shallow green. Recording what was proven, by whom, and what was not.

### Re-derived by the controller at source, before any review returned

Merge-base is `b338bdf` — exactly the base this PR states it rebased onto, so that claim is true. The intervening range is **#1511 alone**, touching `docs/dev/file-budget.tsv`, `tests/stack/run.sh` and `tests/stack/test-unit-helpers.sh`, and **zero files under `lib/pithead/`**. `merge-tree --write-tree` is clean, rc 0, computing tree `6972650b`; the negative control for that instrument was fired on a genuinely divergent pair (`develop` × `develop-v2`: rc 1, 3 CONFLICT entries) rather than on a branch whose head is the tip.

Budget rows were read **on the computed merged tree**, not predicted and not taken from the gate: `99-remainder.sh` 4257/4257, `33-render-env.sh` 494/494, and the other three slices at 157/214/314 — all under the 400 floor and correctly rowless. `tests/stack/run.sh` is untouched at row 739 against an actual of 736.

### The pass's strongest acts, which are the reason this is a merge

1. **It fired the parity gate with a mutation it chose itself** — `wait_for_onion` in `32-onion-provisioning.sh`, a line the author's control never touched. It predicted the artifact line by grepping the built artifact *before* mutating (7244), proved the mutation landed on the tracked file with `git diff --cached --numstat` (`1 1`, not blank), and the gate failed at exactly `7244c7244`. Prediction and instrument converged. Replaying an author's control confirms the control; choosing your own tests the instrument.
2. **It confirmed the known-weak heredoc instrument by a third independent route** and then applied it correctly rather than generically: a synthetic top-level heredoc with nothing enclosing it returns **rc 0** from `bash -n`, while one wrapped in a function reproduces rc 2. It then corrupted the *second* heredoc in `33-render-env.sh` (line 489), which the author's control did not exercise, and verified directly that both real heredocs in this cut sit in the structurally-enclosed shape. Its conclusion is the honest one: **the 35 ok / 0 fail sweep is legitimate evidence here only because these two heredocs happen to be enclosed — that is a checked fact about these heredocs, not a property of the sweep.**
3. **Conservation by byte-identity, not arithmetic.** Concatenating the four slices against `b338bdf:lib/pithead/99-remainder.sh` lines 1–1183 leaves exactly one missing trailing blank — the fourth seam blank the build re-inserts — and `cmp` of the head's remainder against the base's lines 1184–EOF is identical, with `PITHEAD_EX_LOCK_TIMEOUT` confirmed surviving at line 3745.

### The budget FAIL it chased rather than filed

Running the budget gate on the raw checked-out head reports a FAIL claiming this PR raises `tests/stack/run.sh`'s ceiling from 739 to 815 — **a file this PR never touches, which is the tell.** The pass probed `origin/develop-v2` before and after (unmoved), established that the *base* carries 815 while the current tip carries 739 (lowered by #1511), and read the tsv off the computed merge tree directly: all three rows coexist with no illegitimate raise. Diagnosed as an unrebased-checkout artifact and disclosed rather than filed. That is the base-ref-moves-under-you trap being handled correctly for the third consecutive pass.

### Ruled by the controller — recorded so they are not re-argued

**Four files where the map's column guessed three: ratified.** Both the three- and four-file cuts cost exactly one new row (494), so the choice is budget-neutral; the split-degree rule sets degree by the largest indivisible function and by naming truth, not by the region's line count; and the file-count column is a floor that has now been low on three consecutive measured rows. Stated honestly: the P8 ground is *not* available here — P8's pairing was named by the program document's own label, whereas this label does not name a fourth group, so the fourth file rests on naming truth plus budget neutrality alone. Sufficient at zero row cost, and a weaker footing than P8's.

The absent base/head suite pair is correct and not a gap: `tests/stack/run.sh` consumes only the built artifact, which is the identical blob on both sides, so both legs would execute the same program.

### Not proven, stated rather than implied

The pass did not independently re-extract all 35 slice-to-artifact spans; it judged the parity gate plus its own mutation to subsume that check, which is sound — parity rebuilding from current sources is whole-file evidence that strictly contains the per-span comparison. `shfmt` cleanliness and the author's blank-line multiset accounting are relayed, not re-derived.

Issue #1022, opened 2026-08-15
The physical appliance bench has no reservation protocol and no automated testing — the KVM battery cannot see what only real hardware shows


VijitSingh97:
Two gaps, one root cause: every automated appliance proof runs in a VM on the bench host, and the physical appliance is touched by hand or not at all.

**Why it is not sufficient.** #1021 was found today only because an operator happened to paste a `telnet` result: the real box has an ISP-assigned globally-routable IPv6 address and serves the dashboard on it. The KVM guest only ever has `192.168.122.x`, so no VM run could ever have caught it. The same blind spot covers Secure Boot, real disk topology (the M4 wrong-disk guard needs a second physical disk), firmware power-on behaviour, thermals/governor, the hardware watchdog actually resetting a wedged board, and real NIC/mDNS behaviour.

**Gap 1 — no reservation.** The miners have a canonical protocol (rig-e2e flock + holder marker + a `~/README.md` box contract, per the bench-allocation policy). The appliance bench has none: no lock, no holder, no contract file — and it is reachable and mutable by anyone with the key. Two sessions can collide on it, and one of them reflashes.

Note the design constraint that makes it different from a rig: **the appliance gets reflashed and factory-reset by the very tests that use it**, so a lock or contract stored *on* the appliance is destroyed by the work it is meant to guard. The reservation and the box contract must live on a stable coordinator (the bench host) — this is a genuine improvement over the rig model, not a copy of it.

**Gap 2 — no automated hardware testing.** `docs/dev/appliance-release.md` defines M1–M10 as a manual pre-release battery and it **has never been run**. Some of it genuinely needs hands (M1 USB flash + Secure Boot, M4 second disk, M8/M10 power cuts), but two items are fully automatable over SSH against the real box — **M7 (build v+1, install, commit)** and **M9 (deliberately broken bundle, health gate refuses, rollback)** — and both are exactly the paths #976's dashboard OS-update action now drives. Several more (M2 discovery, M3/M5 install and chain preservation, M6 paste-configure) are semi-automatable the same way the KVM harness drives the wizard over HTTP.

**Wanted:**
1. A reservation mechanism for the appliance bench: flock + holder marker held on the bench host, an off-box check, and a documented contract stating the box's role, what is safe to destroy, and what must be preserved (it currently holds a ~105 GB synced Monero chain — expensive to lose, and the reason "just reflash it" is not free).
[...]

PR #1513, opened 2026-08-29
fix(integration): trap aborts on EXIT alone, so a signalled run dies instead of resuming (#1401)

### LINKED ISSUE #1401: e2e.sh traps EXIT INT TERM, which runs restore_all TWICE on any signalled abort

VijitSingh97:
Fixes #1401.

## What this fixes

`trap restore_all EXIT INT TERM` reads as the more careful idiom, and it is wrong. Measured (not
assumed) by signalling the process group the way a terminal Ctrl-C does: a bash `INT`/`TERM` handler
that **returns** does not die — the shell resumes at the next command. So naming the signals does not
add a second teardown to an aborting run; it converts the abort into "tear down, then carry on."

For `tests/integration/e2e.sh` that is a **false green**, not the redundant double-teardown #1401
originally graded as latent. On a Ctrl-C or a cancelled CI job, `restore_all` ran its full unwind
(miner pool restored, baseline stack redeployed), set its `RESTORED` re-entrancy guard, returned — and
the run **carried on**, measuring the restored baseline stack instead of the branch under test, then
exited 0. The guard made the later real `EXIT` firing a no-op, so nothing unwound the continuation
either. An aborted release-gate run reporting success is this harness's founding defect class, not a
tidiness point, and that grading is ratified — it is not softened here.

A second live instance of the same spelling, found by enumerating the class rather than stopping at
the site the issue named: `tests/integration/tor-client/probe.sh:56`. Different consequence — it would
kill tor and continue into the bootstrap wait, grepping a log nothing is writing any more, and spend
the full `BOOT_TIMEOUT` reporting "tor did not bootstrap" for what was really an abort.

Two other comments (`rig-key-ledger.sh`, `selftest-rig-key-ledger.sh`) already chose bare `EXIT` for
an earlier issue (#1379) but described the reason in the incomplete, safe-direction form ("runs the
handler twice"). Their conclusion was already right; the wording is sharpened to state why, rather than
swept.

## The method note, because it generalizes past this fix

"Did the signal kill it" is unmeasurable without a **no-trap positive control** whose only correct
answer is death. Two plausible harnesses each gave a confident, wrong "it survived" before landing on
one that didn't: an async job (`cmd &`) inherits `SIGINT` ignored, and `trap -` cannot undo that; and a
bash shell waiting on a foreground child swallows a `SIGINT` the child itself never received, so
signalling the parent pid is not signalling a Ctrl-C. Signalling the **process group**, with a no-trap
row in the matrix, is what caught both at once — without it the contaminated matrix reads identically
to the real one.

## What was run (this session did not re-run any of it — see below)

Relayed from the branch's own proof round, published in full on the issue
(`issuecomment-5461074966`, the re-grading, and `issuecomment-5461137483`, what shipped):

- `make test-integration-selftest` rc=0 — new file `selftest-abort-traps.sh` 16/0, ten files total, 0
  failures.
- A three-leg mutation battery, each leg gated on `cmp` proving the edit actually applied before it
  counts: re-adding `INT TERM` to `e2e.sh` reds its own assertion; the same edit to `probe.sh` reds
  its own; deleting `e2e.sh`'s trap line reds the extraction guard rather than passing vacuously. Both
  files restored byte-identical afterward and the suite re-run green.
- `shellcheck --severity=warning tests/integration/*.sh` rc=0 from the repo root over the whole glob
  (`probe.sh` checked separately — it isn't in `lint-sh`'s glob).
- `shfmt -i 4 -d` over the Makefile's exact glob: rc=0.
- `scripts/lint-file-budget.sh` rc=0 — `e2e.sh` sits at 691 lines against a 691 ceiling (the issue's
  own 766/766 figure is stale), so the comment-only change there is line-neutral by necessity, not
  incidentally.

Docs needed no change, and that is the finding rather than an omission:
`docs/dev/integration-testing.md:171` already promised the `EXIT`-trap behaviour. The doc was right
and the code disagreed with it.

## What I did NOT do (this session)

This session's only job was opening this PR from the already-pushed, already-proven branch
(`a6ec29c`) — not rebuilding, re-cutting, or amending the fix.

- **Did not re-run the proof round or the mutation battery.** Every figure under "What was run" above
  is relayed from the issue comment, not re-executed here.
- **Did not touch #1022** (unrelated, open).
- **Did not touch #1236**, which stays parked on operator-supplied values.

## What I re-derived this session, rather than relayed

- Branch tip `a6ec29c` is exactly **one** commit ahead of `origin/develop-v2`
  (`git log origin/develop-v2..HEAD`), with merge-base `cf6a888`. `develop-v2` itself has moved 10
  commits since that merge-base (unrelated merged PRs) — the "1 commit ahead" figure is about new
  commits on this branch, not the raw `rev-list --count`, which reads 10/1 because of that shared
  history.
- Diffstat re-measured at `git diff --stat origin/develop-v2...HEAD`: 5 files changed, 152 insertions,
  9 deletions — `tests/integration/e2e.sh`, `tests/integration/rig-key-ledger.sh`,
  `tests/integration/selftest-rig-key-ledger.sh`, `tests/integration/tor-client/probe.sh`, and the new
  `tests/integration/selftest-abort-traps.sh`. Matches the branch's own accounting exactly.
  - **File-count check I ran, not merely accepted:** `selftest-abort-traps.sh` (135 lines) is a
    genuinely new file — `selftest.sh` sits on its own 686-line ceiling with zero slack, so the new
    cases could not have landed there; the runner globs `tests/integration/selftest*.sh`
    (`Makefile:29`), so the new file needs no registration.
- Working tree is clean, nothing unpushed, no existing PR was open for this branch or head ref.
- Issue #1401 is still OPEN.

All paths touched are under `tests/integration/`, which this lane owns exclusively — no lane-guard or
shared-file question here.

### DIFF
--- a/tests/integration/e2e.sh
+++ b/tests/integration/e2e.sh
@@ -191,7 +191,7 @@ CONTROL_VERDICT_BEFORE=""
 # so the EXIT trap always has a target even if it fires before preflight refines it.
 RESTORE_DIR="$CANONICAL_DIR"
 
-# --- Restore (runs on EXIT, even on failure / Ctrl-C) -----------------------
+# --- Restore: fires ONCE on EXIT, Ctrl-C included. Never add INT/TERM (#1401) ----
 restore_all() {
     local rc=$?
     [ "$RESTORED" = "1" ] && return
@@ -300,7 +300,7 @@ restore_all() {
     fi
     if [ "$rc" -eq 0 ]; then ok "restore complete."; else warn "restore complete (the run itself failed — see above)."; fi
 }
-trap restore_all EXIT INT TERM
+trap restore_all EXIT
 
 # --- Small waiters / helpers ------------------------------------------------
 wait_bench_healthy() { # <timeout_s>
--- a/tests/integration/rig-key-ledger.sh
+++ b/tests/integration/rig-key-ledger.sh
@@ -41,10 +41,12 @@
 # ever changes, this copy reds rather than drifting quietly.
 #
 # WHY `EXIT` ALONE AND NOT `EXIT INT TERM`. Measured on this box rather than assumed: a bare EXIT
-# trap DOES run when the shell dies of SIGINT or SIGTERM, and `trap … EXIT INT TERM` runs the
-# handler TWICE for one signal — once for the signal, once for the exit that follows. `EXIT` covers
-# Ctrl-C, a `kill`, and a cancelled CI job with exactly one firing. The unwind is written to be
-# idempotent anyway, because a trap that assumes it runs once is a trap nobody can safely change.
+# trap DOES run when the shell dies of SIGINT or SIGTERM, so naming the signals adds no coverage.
+# What it adds is worse than the second firing it is usually described as — a bash INT/TERM handler
+# that RETURNS does not die, so the script RESUMES at the next command, runs on to the end and exits
+# 0 (#1401; proof in selftest-abort-traps.sh). `EXIT` covers Ctrl-C, a `kill`, and a cancelled CI job
+# with exactly one firing. The unwind is written to be idempotent anyway, because a trap that assumes
+# it runs once is a trap nobody can safely change.
 #
 # ROUTES. Two, because the write surface is not all one path: `dash` is the dashboard's
 # /api/control/worker-apply (the #513, #1236 and #1002b legs) and `rig` is a direct dial at the
--- a/tests/integration/selftest-abort-traps.sh
+++ b/tests/integration/selftest-abort-traps.sh
@@ -1,9 +1,12 @@
 #!/usr/bin/env bash
 #
-# Self-test for the harness's ABORT TRAPS (#1401): every `trap` in tests/integration/ that exists to
-# unwind on an abort must name `EXIT` and nothing else. Standalone (not sourced by selftest.sh) so it
-# never touches that file's budget ceiling — the #1258/#1301 "moved into its own file" precedent. Run
-# directly, or via `make test-integration-selftest`. No server needed.
+# Self-test for the harness's ABORT TRAPS (#1401): checks exactly two named sites — e2e.sh's
+# `restore_all` trap and tor-client/probe.sh's `kill`-tor trap — and asserts each names `EXIT` and
+# nothing else. This is NOT a directory-wide sweep: it is two hardcoded (file, handler) pairs, not a
+# walk over tests/integration/, so a third file adding the same `EXIT INT TERM` idiom would pass this
+# file unnoticed (#1515, filed to widen this to a real walk). Standalone (not sourced by selftest.sh)
+# so it never touches that file's budget ceiling — the #1258/#1301 "moved into its own file"
+# precedent. Run directly, or via `make test-integration-selftest`. No server needed.
 #
 # WHAT THIS EXISTS TO STOP, and why a comment alone would not have. `trap handler EXIT INT TERM` is
 # the common idiom and it READS as more careful than bare `EXIT`, so it gets re-added by anyone who
--- a/tests/integration/selftest-abort-traps.sh
+++ b/tests/integration/selftest-abort-traps.sh
@@ -1,0 +1,135 @@
+#!/usr/bin/env bash
+#
+# Self-test for the harness's ABORT TRAPS (#1401): every `trap` in tests/integration/ that exists to
+# unwind on an abort must name `EXIT` and nothing else. Standalone (not sourced by selftest.sh) so it
+# never touches that file's budget ceiling — the #1258/#1301 "moved into its own file" precedent. Run
+# directly, or via `make test-integration-selftest`. No server needed.
+#
+# WHAT THIS EXISTS TO STOP, and why a comment alone would not have. `trap handler EXIT INT TERM` is
+# the common idiom and it READS as more careful than bare `EXIT`, so it gets re-added by anyone who
+# has not measured it. Measured here rather than assumed: a bare EXIT trap already runs when the
+# shell dies of SIGINT or SIGTERM, so naming the signals adds no coverage — what it adds is that
+# **a bash INT/TERM handler that RETURNS does not die.** The shell resumes at the next command.
+#
+# For e2e.sh that was a FALSE GREEN, not a redundant teardown: on a Ctrl-C or a cancelled CI job,
+# restore_all ran the full unwind (miner pool restored, baseline stack redeployed), set its RESTORED
+# guard, returned — and the run CARRIED ON, measuring the restored baseline stack instead of the
+# branch under test, then exited 0. The guard is what made the second firing a no-op, so nothing
+# unwound the continuation either. An aborted release-gate run reporting success is this harness's
+# founding defect class, which is why the guard is a behavioural test and not a comment.
+#
+# THE CONTROLS ARE THE POINT. "Did the signal kill it" is unmeasurable without a row whose only
+# correct answer is death: two plausible harnesses each reported a confident, wrong "it survived"
+# before this one. An async job inherits SIGINT **ignored** and `trap - INT` cannot undo it (a signal
+# ignored at shell entry stays ignored), and a bash waiting on a foreground child **swallows** a
+# SIGINT the child never received — so signalling the shell's pid alone does not model a Ctrl-C.
+# `set -m` fixes both: it puts each background job in its own process group AND leaves its signal
+# dispositions at default, so the trial can signal the GROUP, as a terminal does. If the no-trap
+# rows below ever stop dying, the instrument is broken and nothing under it means anything.
+#
+set -m -uo pipefail
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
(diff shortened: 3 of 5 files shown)


VijitSingh97:
Discharged the non-author pass's should-fix in-repo rather than in the record: `selftest-abort-traps.sh`'s header (lines 6-7) overclaimed a directory-wide sweep it doesn't do. Amended it in `62c1916` to state the two named sites it actually checks (`e2e.sh`'s `restore_all` trap, `tor-client/probe.sh`'s tor-kill trap) and that it is not a directory-wide walk.

Why a real commit rather than a verdict-comment correction: the wrong claim lives in a `.sh` header, which is durable in-repo documentation, not a PR-body/merge-record artifact — and it's the dangerous direction, since a guard that overclaims its blast radius is what stops the next contributor from adding the coverage that's actually missing.

Comment-only change (verified with a fired control: a real code-line edit reports non-zero non-comment diff lines, this change reports zero) and the self-test still runs 16/0. The behavioural fix and its proof are untouched.

Widening the guard to a real directory walk is its own behavioural change with its own proof obligation — filed as #1515, not bundled here.

VijitSingh97:
## Controller verdict: MERGE

The non-author pass returned **MERGE-WITH-SHOULD-FIX**, its should-fix has been **discharged in-repo**
(`62c1916`), and the widening it recommended is filed as **#1515**. Merging at head `62c1916` against
base `b84b1749`.

### What the pass proved, rather than read off the code

It did the thing the brief said would outweigh everything else: **it made the new self-test fail on
purpose.** In an isolated `git archive` copy it fired three adversarial shapes at the *unmodified*
self-test — `INT TERM` added to an untouched existing file's trap, a brand-new file carrying the bad
idiom, and a second differently-named trap inside `e2e.sh` itself — and **all three passed 16/0**.
It then confirmed with fired controls that the guard *does* red correctly on the two sites it is wired
to (reorder → `expected [EXIT], got [INT TERM EXIT]`; partial → `expected [EXIT], got [EXIT INT]`).

That is the finding: `selftest-abort-traps.sh:99-107` checks exactly two hardcoded (file, handler)
pairs — no glob, no directory walk — while its header claimed it enforced the rule for *every* trap
under `tests/integration/`.

It also confirmed the behavioural claim by **execution on a second, independently built harness** with
its own no-trap positive control: no trap + `SIGINT` → rc 130, dead; `EXIT` alone → rc 130, handler
fires once; `EXIT INT TERM` → **rc 0**, `READY,FIRED,SURVIVED,FIRED` — the script resumes past the
signal and exits success. Two instruments, both passing their own control, both agreeing.

### The ruling on the should-fix, and why it was discharged rather than noted

**The header is corrected; the guard is not widened in this PR.**

*Where the wrong text lives decides the class.* A loose claim in a PR body is a merge-record defect and
is corrected in a verdict comment. This claim sat in a `.sh` header in the repo — durable in-repo
documentation — so it earned a real commit. And it was the dangerous direction: **a guard that
overclaims its blast radius is worse than one that states a narrow truth, because the overclaim is
what stops the next contributor from adding the missing coverage.** Widening is a behavioural change
to a test and needs its own proof; it does not ride a PR whose proof is already complete.

`62c1916` narrows the header to name the two sites, says plainly it is not a sweep, and cites #1515.
The discharge proved itself comment-only with a **fired** control (a throwaway real-code perturbation
scored 2 non-comment changed lines; the actual amend scored 0), and re-ran the self-test green at
16/0. The behavioural change — `e2e.sh`, `probe.sh`, the trap-matching logic, the assertion bodies —
was not touched.

### What I re-derived at source before merging (as against relayed from the pass)

- **RE-DERIVED.** Base tip `b84b1749`, head `62c1916`, read immediately before merging.
- **RE-DERIVED.** `62c1916` is comment-only against the reviewed head `a6ec29c` — by direct inspection
  of the complete diff, not by re-running the discharge's own counter. 7 insertions / 4 deletions, every
  changed line `#`-led and inside the header block at lines 1–12.
- **RE-DERIVED.** `merge-tree --write-tree` base × head is **clean, rc 0, tree `91dca326`**. This is
  deliberately *not* the `9ca49539` in the pass and in the earlier controller record — that tree was
  computed against head `a6ec29c`; the difference is exactly the discharge commit. Instrument control
  fired on a genuinely divergent pair (rc 1).
- **RE-DERIVED, on the merged tree rather than the branch.** `docs/dev/file-budget.tsv` is **byte-identical**
  to the tip's. `tests/integration/e2e.sh` 691 actual against its 691 row; `selftest.sh` 686 against its
  686 row, untouched. `selftest-abort-traps.sh` is now **138** lines (135 + the 3-line header growth) —
  still under the 400-line floor, so correctly rowless.
- **RE-DERIVED, on the merged tree.** The standalone placement claim: the four references to
  `selftest-abort-traps` are comment text plus its own summary `echo` — **no `source`, anywhere**. Control
  fired: the same grep shape returns real `source` lines for `lib.sh`. So the header's budget rationale
  holds, and the amend cannot have re-staled a sourcing file.
- **RE-DERIVED.** All 16 required checks pass, including `Shell tests (shellcheck + pithead suite)` — the
  live arbiter for a shell-only change, and the one the pass had to leave `pending`.

### One correction to the pass, which does not change its verdict

The pass wrote that the directory-wide framing lived in *"its own header comment and the PR body."*
**The PR body does not carry it.** I checked with a fired control: the phrasing search returns nothing
in the body and fires on line 3 of the original header. So there is no merge-record defect to correct
here — the overclaim was in one place, and that place is the one the discharge amended. Recording it
because a faithful relay of a slightly-wrong claim is how a wrong claim becomes durable.

### Scope, checked against the issue rather than the PR

`Fixes #1401` is earned. #1401's own scope names `e2e.sh`; this PR fixes that site and a second live
instance in `tor-client/probe.sh` found by enumerating the class, and the pass independently enumerated
all 11 `trap` statements under `tests/integration/` at head and found every one `EXIT`-alone. #1515
carries the guard's widening and stays open.

PR #1514, opened 2026-08-29
test(stack): cover the os-install contention verdict, and arm the chain to prove it (#1482)

VijitSingh97:
Closes the should-fix from #1510's non-author pass: the OS-update lock-contention path shipped
with **zero committed coverage**. #1510 proved the behaviour with a throwaway harness; this ships
it as a permanent tier-1 test.

## The claim under test

`os-update` takes the mutation lock (#1482). An install that arrives while another pithead
operation holds the machine must come back **`rejected`**, never as a **`failed`** install. The
distinction is the whole point: "failed" tells an operator the install ran and broke, when in fact
`rauc install` was never reached and the staged bundle is still there to retry.

## Where it went, and why not where it was routed

It was routed into `tests/stack/test-appliance-os-update-verbs.sh`. **That file is at 454 lines
against a recorded ceiling of 454 — zero headroom.** I fired the control rather than argue it:

| attempt | gate says |
|---|---|
| grow the file, keep the ceiling | `FAIL — ... is 456 lines, over its recorded ceiling of 454` |
| grow the file AND raise the ceiling | `FAIL — ... raises ... from 454 to 456. Ceilings only go down.` |

It fails in both directions and there is no third option; only `lib/pithead/99-remainder.sh` is
`monotonic_exempt`. So this lands as a new sibling domain,
`tests/stack/test-appliance-os-update-lock.sh` — **177 lines, under the 400-line target, so it
needs no budget row at all** — plus the 3-line source stanza in `run.sh` that #1509 banked for
exactly this and disclosed in its own body.

**#1511 merged while this was being written** and moved `run.sh` 812 → 736 with its ceiling
815 → 739. The three banked lines were carried forward intact, and this stanza consumes exactly
them: `run.sh` is now **739/739**. No ceiling is raised anywhere in this PR.

**This stanza is the #1509 bank's named beneficiary arriving, and it CONSUMES the bank.** #1509
banked three lines of `run.sh` headroom for #1482 specifically and disclosed it in its own body;
this is the change it was banked for. `run.sh` lands at **ceiling == actual (739/739)**, which is
the state the reserved headroom was always going to resolve to — the bank is consumed by its
beneficiary, not reclaimed.

## The fixture is ARMED, and that is the load-bearing part

The install door reads the *checked* target (`$OSLDIR/target.json`). With no `os-check` run, the
door refuses with *"no fully downloaded update bundle is staged"* — which is **also a `rejected`**.
A contention assertion written against an unarmed fixture goes green off that staging door while
the contention path never executes, and the two read identically. #1482's own harness hit exactly
that false-green once.

So the chain is armed for real — `os-check` → checked target → snapshot → restore per section
(`os-check` is throttled to once per ten minutes and a successful install consumes `target.json`) —
and the file opens with a **positive control that an uncontended install genuinely succeeds**:
reports `installed`, `rauc install` actually ran, state file records `reboot-pending`. Only then
does the contended section run.

**13 assertions, 0 failures.**

## Mutation control — the test can fail, and for the right reason

With the `PITHEAD_EX_LOCK_TIMEOUT` branch cut out of the artifact under test (mutation proved
applied first: `git diff --numstat` on the tracked `pithead` → `0 7`), the same input flips **4 of
the 9 contended assertions**, led by the load-bearing one:

- `contended: the install is rejected, not failed` — `expected [rejected], got [failed]`
- the mutant emits exactly the misleading text this change exists to prevent: *"the install did not
  complete — the running system is untouched and mining continues"*

**The other five stay green under the mutant, and I am not claiming they discriminate.**
"nothing was changed" appears in the generic timeout text too, and bundle-kept / no-in-flight-flag /
state-returns-to-idle hold either way. They assert properties the behaviour genuinely requires, so
they stay — but four assertions carry the discrimination, not nine.

## Name collisions, since run.sh sources both files into ONE shell

Nothing here reuses a name the verbs file defines. Its `$OSC` / `$OSDIR` / `$OSRES` / `$UOS` and
`osrun` / `os_intent` / `os_reset` are untouched; this file's are `$OSL` / `$OSLDIR` / `$OSLRES` /
`$ULOCK` and `oslrun` / `osl_intent`. Every write lands under `$OSL` (`$SANDBOX/os-lock`).

The holder sets a global rather than echoing it: `H=$(hold_lock ...)` around a function that
backgrounds a long-lived holder blocks until that holder's stdout closes — i.e. for its whole
lifetime. The shipped `test-lifecycle.sh` assigns `$!` to a global for the same reason.

## Docs

`docs/dev/testing-strategy.md`'s `Dashboard OS-update verbs` row is **extended, not rewritten**.
That row is an enumeration and everything it already enumerates is still true and still covered —
the checkmark was not false, the list was incomplete. Sweeping correct text to match a refutation
would degrade the record.

## What was run

- The new domain standalone and again after the #1511 rebase: **13 passed, 0 failed** both times.
- Mutation control above: **9 passed, 4 failed**, with the mutation proved applied.
- `scripts/lint-file-budget.sh` rc **0**, with `origin/develop-v2` probed unmoved immediately
  before and after. (It moved once mid-session — #1511 — and that run was discarded and redone on
  the new tip rather than argued equivalent.)
- `bash -n` and `shfmt -i 4 -d` clean on both touched files.
- `make lint-docs-voice` clean, with a banned-word control fired first (rc 2), then reverted.
- **Full `tests/stack/run.sh` at head: rc `0`, `pithead tests: 2992 passed, 0 failed`**, 3306 log
  lines, 275 section headers, zero failure glyphs. Read off the rc FILE, not a detached launcher's
  exit code, and matched with `grep -F 'pithead tests'` — that summary line carries an ANSI reset
  *between* `passed` and the comma, so a `passed,` pattern silently matches nothing.
  **The accounting closes exactly:** against the 2979-passed / 274-header baseline this run is
  **+13 assertions and +1 section header** — this domain's 13 and its one header, nothing else moved.
  This run is **not** the null-experiment shape: the PR changes `run.sh`'s sourcing, so sourcing
  order, name collisions with the verbs file and `_d0` accounting could genuinely have failed here.

## Lane disclosure

`tests/stack/` is the tests lane's, not this lane's. The placement is ratified by the controller,
which also ruled against splitting the verbs file to make room — that would couple this feature to
a refactor, which is the coupling the banked headroom exists to remove. `docs/` is shared, and
`docs/dev/file-budget.tsv` is regenerated per lane, never merged by taking a side.

**lane-guard did not refuse this commit, and that was not a pass — the worktree was unguarded.**
`lane-guard.sh` returns 0 unless the worktree root holds an (untracked) `.fleet-role` file, and this
lane's cut worktrees had none. That is this lane's own setup gap rather than a defect in the guard:
the tests lane roles all of its cut worktrees. I have since written `.fleet-role` into both of this
lane's, and confirmed the armed guard refuses exactly `tests/stack/run.sh` and
`tests/stack/test-appliance-os-update-lock.sh` by name while `docs/` passes as shared — which is the
correct verdict. **So "the guard passed" is not evidence here and is not offered as any**; the
placement's authority is the controller's ruling and this disclosure.

## What I did NOT do

- **No non-author review.** This lane authored it.
- **`make lint` / `lint-sh` not run**, no lint lock taken — it is the OOM path on this box. CI's
  Shell tests job is the arbiter.
- **Behaviour A — that `os_update` takes the lock window at all — stays uncovered here.**
  `lock_wiring_pair` is a local helper at `test-lifecycle.sh:756`, not in `lib.sh`; a new file could
  reach it only by being sourced after that stanza, which is the ordering accident #1105 exists to
  remove. `test-lifecycle.sh` is closed at its own ceiling. This PR covers the dashboard runner's
  contention verdict, not the lock wiring. Naming the gap rather than hiding it.


VijitSingh97:
## Controller verdict: MERGE

The non-author pass returned **MERGE, no findings**, and it answered the one axis its brief said would
outweigh everything else. Merging at head `b431a65` against base `a21619ae`.

### The decisive axis was actually fired — twice, by two different routes

The brief's first line said that failing to make the new test red on purpose would outweigh anything
else the pass could report. It did not have to fall back on that:

- **Attack A — remove the lock entirely** (`mutation_lock_acquire os-update` → `:`), edit confirmed
  applied by `numstat` before it counted. 13/0 → **5 passed, 8 failed**, led by the load-bearing row
  `contended: the install is rejected, not failed` → `expected [rejected], got [installed]`. Restored,
  back to 13/0.
- **Attack B — the author's own narrower mutation, reproduced independently**: neuter only the
  `PITHEAD_EX_LOCK_TIMEOUT` routing branch, leaving the lock and the exit code intact. → **9 passed,
  4 failed**, same load-bearing row flipping first, and the mutant emits verbatim the misleading
  operator text this test exists to prevent. Restored, 13/0.

Both reddened for the *claimed* reason — no missing binaries, no jq crashes, no unrelated assertions
moving — which is the distinction that separates "reds because contention was unhandled" from "reds
because the harness broke."

**And it is real concurrency, not a simulation.** `osl_hold()` backgrounds a subprocess that opens the
lock file and calls `flock`, and the parent then **polls `flock -n` until it observes the file actually
held** before proceeding — a synchronisation barrier, not a sleep-based race. The proof of that is
Attack A itself: with the guard gone, the "contended" install *completed* inside the holder's 120s
window, which is only possible if both processes were genuinely running at once.

### What I re-derived at source before merging, as against relayed from the pass

- **RE-DERIVED.** Base tip `a21619ae`, head `b431a65`, read immediately before merging.
- **RE-DERIVED.** `merge-tree --write-tree` base × head is **clean, rc 0, tree `df4c7a3d`**. This is
  deliberately neither the pass's `ed594655` nor the earlier `7ed30bb3`: all three are legitimate, and
  the difference is entirely the **tip moving under the review** — `7ed30bb3` was against `b8176239`,
  `ed594655` against `b84b1749` (post-P9), and mine against `a21619ae` (post-#1513). A computed tree is
  valid only for the exact (base, head) pair it came from. Instrument control fired on a genuinely
  divergent pair (rc 1).
- **RE-DERIVED, and this is the check that matters more than the count.** The branch is 1 ahead / 5
  behind, but `git diff <merge-base> origin/develop-v2 -- tests/stack/ docs/dev/testing-strategy.md` is
  **empty** — the five intervening commits touch none of this PR's paths. Control fired: the same query
  against `lib/pithead/` returns P9's real 1183-line change, so the query discriminates rather than
  always agreeing.
- **RE-DERIVED, on the merged tree rather than the branch or the gate.** `tests/stack/run.sh` **739
  actual against its 739 row** — ceiling == actual, the equality `check_monotonic` demands on a rise, so
  **the banked headroom is CONSUMED, not raised**. The new file is **177 lines with no tsv row**, correct
  under the 400 floor. `test-appliance-os-update-verbs.sh` untouched at **454** — the 454/454 ceiling that
  made routing this into the existing file impossible. `docs/dev/file-budget.tsv` is **byte-identical** to
  the tip's, as it must be when a bank is consumed rather than moved.
- **RE-DERIVED, with a near-miss control.** The `_d0` nesting trap: the new file contains **zero**
  occurrences of `source`, anywhere, prose included. The control is the strong form rather than the easy
  one — the same unanchored pattern returns 90 hits in `run.sh` **and 1 hit in the sibling
  `test-appliance-os-update-verbs.sh`**, a domain file of exactly the same kind (a comment, not a
  stanza). So the zero is a measured absence, not a pattern that never matches.
- **RE-DERIVED.** The registration stanza at `run.sh:711` is the same idiom as the neighbours above and
  below it, correctly placed between the verbs and kernel-boot domains, with its `shellcheck source=`
  directive.
- **RE-DERIVED.** Isolation, the axis where a green test can still be wrong under the standing
  don't-serialise directive: every path is under the per-invocation `mktemp` sandbox — `OSL="$SANDBOX/os-lock"`,
  `OSLLK="$SANDBOX/os-lock-held.lock"`, guarded by `: "${SANDBOX:?}"`. No fixed `/tmp`, `/run`, `/var` or
  `/etc` path, so no cross-talk surface for concurrent worktrees.
- **RE-DERIVED, live rather than relayed.** All checks pass, including `Shell tests (shellcheck + pithead
  suite)` at 11m59s — the arbiter for a shell change.

### Scope: #1482 stays OPEN, and that is correct

Checked against the **issue**, never the PR. #1482 names six unlocked mutating verbs; exactly one
(`os_update`) is wired to a lock acquire. This PR carries **no `Closes`/`Fixes` keyword for #1482** — the
one "Closes" in the body is a sentence about #1510's review comment, not an issue reference — and its
"What I did NOT do" section states plainly that `os_update` *taking* the lock at all remains uncovered
here and belongs to `test-lifecycle.sh`. **That explicit disclosure is a positive finding, not a
non-event**: a reclaim or close trigger phrased as "the named change landed" goes ambiguous the moment
the named change is an issue that ships in slices, and this body refuses to create that ambiguity.

### One record item this merge closes

Placement of `tests/stack/test-appliance-os-update-lock.sh` was ruled correct for the appliance lane by
the controller, but `tests/stack/` belongs to the tests lane and the grant was never written into
`roles/appliance.paths`. It is being recorded there now as granted-and-expired-at-this-merge, so the
record shows an authorisation rather than a path that slipped through.

PR #1516, opened 2026-08-29
refactor(cli): cut service-config injection, the Caddyfile and the Quadlet units out of the remainder (#1105 Phase 2 P10)

VijitSingh97:
#1105 Phase 2 **P10** — the inject / Caddyfile / Quadlet region moves out of the remainder into three
slices. Cut from `develop-v2` at `43d94d1` (#1514's merge), which is where P9 (#1512) left the
remainder: my first slice opens at artifact line **8050**, the exact line #1512's body records as the
start of its `99-`. No rebase was needed and none was done.

## The cut

Region: `lib/pithead/99-remainder.sh` lines **1–738**, ending where `render_quadlet_units` closes.
Line 739 is a single blank; line 740 opens the HugePages/GRUB block, which is P11's.

| slice | lines | budget row |
|---|---|---|
| `34-inject-service-configs.sh` | 25 | none — under the 400 floor |
| `35-caddyfile-and-auth-helpers.sh` | 321 | none — under the 400 floor |
| `36-quadlet-units.sh` | 390 | none — under the 400 floor |
| `99-remainder.sh` | 4257 → **3518** | lowered, as every P-row must |

25 + 321 + 390 + 3518 = 4254, plus the 3 build separators = 4257, the remainder's pre-cut length.

**Boundaries were derived by content, never by offset.** The map's printed range for P10 carries its
own warning (`range overlap in recons — derive`), and the LABEL is what I cut to.

`cmp` reports the rebuilt artifact identical to the pre-cut one at **713935 bytes**, and the blob sha
is `d5378028…` on both sides. Two routes, not one derivation read twice.

### 2. The parity gate — the claim that matters

`make lint-pithead-parity` rc **0** over 38 slices. This is the half byte-identity cannot supply on
its own: an identical blob is *equally consistent* with nobody having regenerated the artifact, so
that it is now stale against the very sources this PR edits. Only rebuilding from the current slices
and comparing excludes that. Byte-identity and parity are one claim in two parts, and neither is
sufficient alone.

### 3. Heredoc integrity — real, NOT subsumed by byte-identity, but `bash -n` is the primary guard

**This section is CORRECTED after I attacked my own claim and it partly failed. The first version of
this body said heredoc integrity is "the axis this row exists for, and `bash -n` does not cover it".
That overstated my own instrument, and the corrected version is below.**

This is the first cut in the program whose region is mostly heredoc templates: **17 of them**, 4 in
the Caddyfile slice and 13 in the Quadlet slice.

**The experiment.** I built the actual hazard rather than reasoning about it. `35-caddyfile-and-auth-helpers.sh`
line 286 opens `cat <<EOF >>"Caddyfile"` and line 287 — the first line of the heredoc body — is
BLANK. That is precisely the invisible-cut site: the build strips blank edges and re-inserts exactly
one separator, so a boundary at an existing blank is consumed and put back. I split the slice there
in a scratch root and ran the real `scripts/build-pithead.sh` (`PITHEAD_BUILD_ROOT`). Four results:

| instrument | verdict on the deliberately bad cut |
|---|---|
| the build / **byte-identity / parity** | artifact **IDENTICAL** — completely blind |
| **`bash -n`** | **rc 2 on BOTH halves** — catches it |
| **`shfmt`** | **rc 1 on BOTH halves** — catches it |
| my `heredoc-check.awk` | rc 1 on part-a, **rc 0 on part-b** — catches only one half |

**What this establishes, and what it costs my own argument:**

1. **Byte-identity does NOT subsume this axis.** A boundary inside a heredoc body reproduces the
   artifact byte for byte and parity passes. That is now demonstrated, not asserted — it is the
   reason a per-slice check is needed at all.
2. **`bash -n` is the primary guard, and I under-credited it.** Its documented weakness is real — a
   four-line file ending in an unterminated `<<EOF` with nothing enclosing it returns **rc 0**,
   warning on stderr only, indistinguishable from the terminated version. But that shape needs a
   *top-level* heredoc at a slice's end. A boundary landing inside a heredoc that sits inside a
   function leaves the function's brace open, so `bash -n` fires. Every heredoc in this region is
   inside a function.
3. **My checker is a cheap independent second opinion, not the load-bearing instrument, and it is
   weaker than `bash -n` on this hazard.** It models delimiter state only, so it missed the half
   whose damage is an unmatched `}` rather than an open delimiter. I am reporting this against
   myself; it should not be read as validated by my own clean run.

**On the shipped cut all three agree: `bash -n` 38/38, zero unterminated delimiters, shfmt clean, and
no slice boundary falls inside a heredoc.** Every template stayed with its render function — the
constraint the map states for this row, discharged by instrument rather than by eye.

**One further correction, recorded rather than quietly fixed.** My first checker was a naive
opener/terminator count, and it flagged `25-address-types.sh` (a P7 file) as imbalanced. That was
**my instrument, not the file**: it matched `(v << r)` — a Python left-shift *inside* a `PYEOF`
body — as a heredoc opener. Real count 2/2, balanced. I verified at source before reporting
anything. **Nothing is wrong with `25-address-types.sh`.**

Each new slice was searched for as a byte sequence in the built artifact. Each matched **exactly
once** — uniqueness, not merely presence — at `34-` **8050**, `35-` **8076**, `36-` **8398**,
contiguous with one separator between each. Those offsets agree independently with four call-site
line numbers grepped before the cut (`inject_service_configs()` 8050, `sha256_hex()` 8077,
`generate_caddyfile()` 8099, `render_quadlet_units()` 8406).

### 5. The blank-line accounting

739 lines removed, 736 added. The excess of **3** is exactly the number of new slice boundaries: the
build re-inserts one blank line per boundary, so those three blanks *must* disappear from the sources
for byte-identity to hold. Stated so it is not read as lost content.

`/ponytail-review` is not in this session's skill list, and the gate hook reads the branch from the
pinned shell cwd rather than the worktree the command ran in — so it names an unrelated branch here.
Disclosed, not treated as a review having happened. Both useful halves run by hand:

**Pure move.** Sides extracted with `awk` (not `grep '^-[^-]'`, which drops removed blank lines and
makes a pure move read as a content change):

- added ∖ removed = **empty** — no new content, no new logic
- removed ∖ added = **exactly 3 blank lines** — the boundary separators, as above

**Adjacent-merge enumeration.** Every adjacent pair, with its +1 build separator:

| merge | lines | verdict |
|---|---|---|
| `34+35` | 25 + 321 + 1 = **347** | legal — under 400, no new row |
| `35+36` | 321 + 390 + 1 = 712 | costs a new budget row |
| `34+35+36` | 738 | costs a new budget row |

**Two files was available, legal, budget-neutral, and is declined — not overlooked.** The honest
argument *for* it is real and I am not hiding it: `inject_service_configs` and `generate_caddyfile`
are invoked adjacently at three of their call sites (artifact 872/873, 9122/9124, 9644/9645), so
"render the service configs before bringing the stack up" is genuine cohesion.

It is declined on the ground P8 ratified: **the map's own label names three groups — "inject +
caddyfile + quadlet" — so this grouping was specified by the program document at map time, before any
cut-time budget objective existed.** A grouping the map itself named is not an artifact of the
smallest-row objective. The naming half stands on its own too: `inject_service_configs` renders
*Tari's* `config.toml` and rearms clearnet-sync markers; filing that under a slice named for the
Caddyfile would misname it.

Note the file-count column for P10 does not print a count at all — it carries the heredoc constraint
instead. So this is not a case of departing from a printed number.

**`sha256_hex` and `caddy_hash_password_b64` are stranded by contiguity, and I want a reviewer to
look at this.** Both are called from exactly one place each — artifact **6726** and **6733**, the
dashboard-auth block inside `23-setup-and-credentials.sh`, which P7 already shipped. Neither is
called by `generate_caddyfile`. They sit physically between `inject_service_configs` and
`generate_caddyfile`, so contiguity forces them into one of the two adjacent slices; they cannot go
to their callers without reordering, which the concatenation build cannot express.

I put them with the Caddyfile because `caddy_hash_password_b64` is caddy-named and hashes with the
pinned Caddy image, and `sha256_hex` is its neighbour in the same dashboard-auth pair. The slice is
named `35-caddyfile-and-auth-helpers.sh` rather than `35-caddyfile.sh` so the file does not claim to
be purely the Caddyfile. This is the weakest judgement call in the PR and is stated as such.

A four-file cut (splitting the two helpers into their own 22-line slice) was also available and
budget-neutral. Declined: it adds a file without adding naming truth, since the helpers' only callers
are in an already-shipped slice either way.

## Budget

`scripts/lint-file-budget.sh` rc **0**. `origin/develop-v2` probed **before and after** the run and
unmoved at `43d94d1` both times — every worktree on this box shares one `.git`, so a peer's fetch can
move the gate's base mid-run and produce a failure naming files this diff never touches.

The remainder row goes 4257 → 3518 in this commit, as every P-row must. No row was added: all three
new slices are under the 400 floor.

## Lint

`shfmt -i 4 -d` clean on all four files — that is the only new lint exposure a cut creates.
`lib/pithead/*.sh` is deliberately absent from the shellcheck arguments (`Makefile`, with the
reasoning in-file); shellcheck reads those bytes through `pithead`, which is byte-identical here, so
shellcheck exposure is unchanged by construction rather than by assertion. `make lint-sh` was not run
in full: it is the serialized OOM path on this box and the artifact it checks did not change.

## Docs

**No doc change ships with this PR, and that is a finding rather than an omission.** I grepped for
prose this cut could silently falsify. `docs/dev/repo-map.md:11` describes the mechanism generically
("the numbered slices in `lib/pithead/`") and stays true. Two docs cite artifact line numbers
(`docs/dev/dual-distribution-plan.md` `pithead:5581`, `pithead:1149`) — and **because the artifact is
byte-identical, no artifact line reference anywhere in the repo can go stale from this cut.** No doc
enumerates a slice count.

## What I did NOT do

- **No base/head suite pair, deliberately.** It is a **null experiment** for a slice cut: the artifact
  is byte-identical, the suite consumes only that artifact, so both legs run the identical program and
  the result is settled before launch. The ruling asks each row to state its choice — this is mine. I
  did not spend the heavy-IO slot on it.
- Did not run the full `make lint` (see Lint above).
- Did not verify P11's region beyond establishing where P10 stops.
- The suite was not run at all, for the reason above; if a reviewer wants one, say so and I will run
  it rather than argue.


VijitSingh97:
## Controller verdict: MERGE

Reviewed against `develop-v2` at `43d94d16`, head `885f83ec`. An adversarial pass returned MERGE with
one non-blocking should-fix. Recording what I re-derived myself and what I relayed, so the next
reader knows which half to check.

### Re-derived by me, not relayed

- **Merge tree.** `merge-tree --write-tree` base x head: rc 0, clean, tree `3c64b282`. Negative
  control fired on a divergent pair (`develop` x `develop-v2`, rc 1). Recomputed at the tip as of
  this comment, not carried forward — three legitimate trees existed for an earlier PR in one
  morning purely from the tip moving.
- **Line counts on the computed merged tree**: 25 / 321 / 390, remainder 4257 -> 3518. Conservation
  closes: 25+321+390+3518 = 4254, +3 build separators = 4257.
- **Budget rows on that tree, with a fired control.** The three new slices are correctly rowless
  (0 rows each); `99-remainder.sh` carries a row of 3518 against an actual of 3518. That is a FALL,
  which demands no equality, and that file is the only `monotonic_exempt` one. Control: the file that
  must have a row returned 1.
- **Heredoc integrity, the axis this row exists for.** `bash -n` on all four files of the merged
  tree: rc 0 **and zero bytes on stderr** — rc alone is not proof here, which is why stderr was
  measured too. Positive control fired: truncating `35-caddyfile-and-auth-helpers.sh` inside the
  heredoc opened at line 286 returns rc 2 with `here-document at line 286 delimited by end-of-file`.
- **Artifact identity.** The committed `pithead` blob is `d5378028` at base and at the merged tree —
  the same blob, structurally, not asserted.
- **CI.** The review's one explicitly unconfirmed item was the Shell tests job, pending when it
  looked. It has since **passed** (10m51s). 15 checks green; the two CVE-sweep jobs skip by design.

### Relayed from the pass, not re-derived by me

Its whole-tree grep for the stranded helpers' callers; its `DASHBOARD_AUTH_HASH_B64` data-flow trace;
its P7 re-verification at source; and its budget-gate mutation control.

### The should-fix, recorded and NOT blocking

The pass fired the author's `heredoc-check.awk` against a corruption of a **middle** heredoc, and got
a **false clean**: `openers=12` against a true 13, `unterminated_at_eof=no`, no red flag. The
mechanism is same-delimiter resync — every heredoc in this region uses the literal `EOF`, so a
damaged body resynchronises against a later occurrence of the same word. This is a **different** gap
from the one the author disclosed against itself, and it is the more consequential one for this
codebase's heredoc style.

It earns no discharge commit: the checker is a scratch tool outside this repo, and the body already
downgrades it to "a cheap independent second opinion, not the load-bearing instrument" rather than
overclaiming it. Where the wrong text lives decides the class, and no in-repo documentation carries a
false claim here. The operative rule for later rows: **do not read this checker's `openers=N` as a
number to trust without a known baseline.**

### On the axis itself

Byte-identity does **not** subsume heredoc integrity in general — the author built the hazard and
demonstrated the artifact stays identical through a boundary at an already-blank line inside a
heredoc body. It holds for *this* cut only because none of the three real boundaries lands inside a
heredoc, which the pass confirmed by locating each boundary itself rather than by inference. Later
rows cannot lean on the cheaper proof for this axis.

### Judged, not merely noted

The three-files-against-two call is ratified on the ground already settled: the map's own label names
three groups, so the grouping predates any cut-time budget objective. The 25-line slice is a group,
not a splinter — folding it in would give one file three unrelated concerns instead of two. The
stranded-helpers placement is sound and, measured, stronger than the author credited itself: the
coupling to `generate_caddyfile` is real, one render-step removed via the environment.

The author correctly declined a base/head suite pair. It is a null experiment for an order-preserving
concat cut whose artifact is byte-identical, and it was right not to spend the slot on one.

Issue #94, opened 2026-06-03
pithead CLI: shell tab-completion + safe chaining of subcommands


VijitSingh97:
## Goal
Two CLI ergonomics improvements for `pithead`.

### 1. Shell tab-completion (bash + zsh)
`./pithead <tab>` completes subcommands (`setup apply up down restart upgrade logs status reset-dashboard help`), and `./pithead logs <tab>` completes service names. Ship a completion script and document installing it.

### 2. Chain subcommands in one invocation — but reject nonsense
Allow running several subcommands **sequentially**, e.g. `./pithead apply upgrade` (run `apply`, then `upgrade`). Validate the whole sequence **up front** and **error out, doing nothing,** on contradictory/nonsensical combinations — e.g. `./pithead up down` (start then stop), duplicates, or order violations (e.g. `down upgrade`).
- Run left-to-right; **fail fast** (stop on the first command that fails).
- Reject invalid sequences before executing anything, with a clear message explaining why.
- Single-command behavior is unchanged.

## Open questions
- The exact incompatibility rule set (which pairs/orderings are nonsensical vs. merely unusual).
- Whether per-subcommand flags are allowed mid-chain (e.g. `apply -y upgrade`).

## Acceptance
- [ ] bash + zsh completion for subcommands (and `logs <service>`), documented.
- [ ] `./pithead apply upgrade` runs both in order, failing fast on error.
- [ ] `./pithead up down` (and other nonsensical combos) error up front and run nothing.
- [ ] Single commands behave exactly as before.

VijitSingh97:
## Implementation notes (drafted 2026-07-07 against develop @ v1.2.2)

**Goal:** bash/zsh tab-completion for `pithead`, plus sequential chaining of subcommands (`./pithead apply upgrade`) with up-front rejection of nonsense sequences.
**Depends on:** nothing.
**Pre-read:** `pithead` `main()` (line 3276–3341 — the full dispatcher case), `show_help` (line 1088), `tests/stack/run.sh` (the stubbed-CLI harness and its `run_sourced`/sandbox helpers), `docs/testing-strategy.md` (tier 1 = `tests/stack/`), `docs/STYLE.md`.

### Steps
[...]

PR #1517, opened 2026-08-29
refactor(cli): cut kernel tuning, the top-level commands and subcommand chaining out of the remainder (#1105 Phase 2 P11)

VijitSingh97:
#1105 Phase 2 **P11** — the kernel/GRUB, top-level-command and subcommand-chaining region moves out
of the remainder into five slices.

## Rebased onto `develop-v2` after #1516, and re-proven there

Cut originally off P10's head; rebased onto the tip once #1516 merged, and **every proof below was
re-run at the rebased head** — a rebase invalidates the whole proof set, not just the suite.

The intervening range is **P10's own merge and nothing else**: `git log 43d94d1..origin/develop-v2`
returns exactly the #1516 merge plus my own P10 commit. So no other lane's work sits between the
base I cut from and the base I am merging into.

## The cut

Region: `lib/pithead/99-remainder.sh` lines **1–1138**, ending where `run_chain` closes. Line 1139 is
a single blank; 1140 opens `# --- Dashboard control channel (#33) ---`, which is P12's.

| slice | lines | budget row |
|---|---|---|
| `37-kernel-tuning-and-preflight.sh` | 288 | none — under the 400 floor |
| `38-setup-command.sh` | 67 | none |
| `39-describe-change.sh` | **428** | **new row at 428** |
| `40-apply-and-render.sh` | 291 | none |
| `41-subcommand-chaining.sh` | 60 | none |
| `99-remainder.sh` | 3518 → **2379** | lowered, as every P-row must |

288 + 67 + 428 + 291 + 60 + 2379 = 3513, plus the 5 build separators = 3518, the remainder's pre-cut
length.

**Every `# ---` section header in the region falls exactly on a slice boundary and opens its own
file** — `# --- Top-level Commands ---` opens `38-`, `# --- Subcommand chaining (#94) ---` opens
`41-`, and the remainder now opens on the control-channel header. Boundaries were derived by
content: for each function I walked up past its comment block to the single blank line above it.
**Two of the six boundaries sit above long comment blocks** (one 26 lines), so an offset-based cut
would have got them wrong.

## The one new budget row, and why it is spent

`describe_change` is a single `case` statement of 428 lines. It is indivisible without a semantic
refactor, which is exactly the condition under which this program spends a row — the map's P11 row
predicted 432 for it. The row is set at the file's **TRUE line count (428)**, not a reserved-headroom
value, which the #1464 amendment fails.

No other slice takes a row. Merging `describe_change` with either neighbour would spend a *larger*
row (496 or 720) for no gain, so 428 is the smallest row this region can be cut to.

`cmp` reports the rebuilt artifact identical to the pre-cut one at **713935 bytes**, and the blob sha
is `d5378028…` on both sides.

### 2. The parity gate — the claim byte-identity cannot make alone

`make lint-pithead-parity` rc **0** over 43 slices. Byte-identity alone is *equally consistent* with
nobody having regenerated the artifact, leaving it stale against the very sources this PR edits.
Only rebuilding from the current slices and comparing excludes that. The two are one claim in two
parts; neither is sufficient by itself.

### 3. Heredocs — this region has NONE, so the P10 hazard does not apply here

**`P11`'s region contains zero heredocs**, verified per slice. `bash -n` is clean on all 43 slices.

Stating the standing position rather than re-litigating it: #1516 established by experiment that a
slice boundary landing at an already-blank line inside a heredoc body reproduces the artifact byte
for byte, so **byte-identity and parity are blind to that hazard** — but `bash -n` and `shfmt` both
catch it, and **`bash -n` is the primary guard**, not the ad-hoc checker I wrote for P10. That
checker has two known blind spots, one I disclosed and one #1516's pass found (same-delimiter
resync producing a false clean). None of it bites P11, because there is nothing to guard here.

Each new slice was searched for as a byte sequence in the built artifact and each matched **exactly
once** — uniqueness, not merely presence — at `37-` **8789**, `38-` **9078**, `39-` **9146**, `40-`
**9575**, `41-` **9867**, contiguous with one separator between each.

### 5. The blank-line accounting

1139 removed, 1134 added. The excess of **5** is exactly the number of new slice boundaries: the
build re-inserts one blank per boundary, so those five *must* disappear from the sources for
byte-identity to hold. Said explicitly so it is not read as lost content.

`/ponytail-review` is not in this session's skill list, and the gate hook reads the branch from the
pinned shell cwd rather than the worktree the command ran in, so it names an unrelated branch here.
Disclosed, not treated as a review having happened.

**Pure move.** Sides extracted with `awk` (not `grep '^-[^-]'`, which drops removed blank lines and
makes a pure move read as a content change): added ∖ removed = **empty**; removed ∖ added =
**exactly 5 blank lines**. No new logic.

**Adjacent-merge enumeration.** Every adjacent pair, with its +1 build separator:

| merge | lines | verdict |
|---|---|---|
| `37+38` | 288 + 67 + 1 = **356** | legal — under 400, no new row |
| `38+39` | 496 | costs a *larger* row than 428 |
| `39+40` | 720 | costs a *larger* row |
| `40+41` | 291 + 60 + 1 = **352** | legal — under 400, no new row |

**Two legal, budget-neutral merges were available and both are declined — not overlooked.**

The honest argument *for* `37+38` is real: `optimize_kernel` (artifact 9123), `preflight_resources`
(9113) and `prompt_start_stack` (9138) are all invoked from inside `setup`, so "what setup does to
the host" is genuine cohesion. Both merges are declined on the same ground: **each would bury one of
the region's own `# ---` section headers mid-file**, and the map's label names these as three
separate groups ("kernel/GRUB + top-level commands + chaining"). A grouping the source and the
program document both mark is not something to dissolve for a file-count saving that costs no rows
either way.

- **`check_disk_grouped` has a caller outside the region** (artifact 1521, in the doctor path) as
  well as inside it (9060). It stays with the preflight group it is named for and called from; this
  is the same shape as P4's `announce_*` decision, recorded rather than hidden.
- **`prompt_start_stack` is a UI prompt sitting between two host-readiness helpers.** It is 17 lines
  and contiguity places it inside `37-`; the file is named "kernel-tuning-and-preflight" rather than
  claiming to be purely kernel work.
- `describe_change`'s only two callers (9612, 9717) are both inside `40-apply-and-render.sh`. The
  pairing is real but they cannot share a file without spending a 720-line row.

## Budget

`scripts/lint-file-budget.sh` rc **0**, with `origin/develop-v2` probed **before and after** the run
and unmoved at `e886b48` both times — every worktree here shares one `.git`, so a peer's fetch can
move the gate's base mid-run and produce a failure naming files this diff never touches.

Two rows change and they are disjoint: `39-describe-change.sh` is **added** at 428, and
`99-remainder.sh` **falls** 3518 → 2379. A falling ceiling is unconditionally allowed — the equality
rule applies to a RISE.

## Lint

`shfmt -i 4 -d` clean across `lib/pithead/*.sh` — the only new lint exposure a cut creates.
`lib/pithead/*.sh` is deliberately absent from the shellcheck arguments (reasoning recorded in the
`Makefile`); shellcheck reads those bytes through `pithead`, which is byte-identical here, so
shellcheck exposure is unchanged by construction. `make lint-sh` was not run in full: it is the
serialized OOM path on this box and the artifact it checks did not change.

## Docs

**No doc change ships with this PR, and that is a checked finding rather than an omission.**
`docs/dev/repo-map.md` describes the mechanism generically and stays true; no doc enumerates a slice
count; and because the artifact is byte-identical, **no artifact line reference anywhere in the repo
can go stale from this cut**.

## What I did NOT do

- **No base/head suite pair, deliberately** — it is a null experiment for a slice cut: the artifact
  is byte-identical and the suite consumes only that artifact, so both legs run the identical
  program and the result is settled before launch. The ruling asks each row to state its choice;
  this is mine.
- Did not run the full `make lint` (see Lint above).
- Did not verify P12's region beyond establishing where P11 stops.


VijitSingh97:
## Merge verdict — MERGE

Reviewed by a non-author adversarial pass, then graded against an **independent second derivation
of the decisive axis** that I ran myself before the pass reported. Below, **RE-DERIVED** means I ran
the command on this box at the current (base, head) pair; **RELAYED** means I took it from the pass
without re-running it.

### The axis this review turned on

**Byte-identity is structurally blind to where the boundaries fall.** Any legal partition of the
region rebuilds the identical artifact, so `cmp`, parity and the budget gate all pass for a cut whose
boundaries are in the wrong places — and `bash -n` is blind too, since a slice that severs a function
from its comment block is valid shell. Every cheap instrument is silent on this axis, so it was
derived twice, independently.

**Both derivations agree: placement is clean at all six boundaries.** Each of the six new segments is
byte-identical to a consecutive span of the pre-cut `99-remainder.sh` — old lines 1-288 / 290-356 /
358-785 / 787-1077 / 1079-1138 / 1140-3518 — every separator (289, 357, 786, 1078, 1139) is a blank
line, and all 3518 lines are consumed with nothing left over. That is a stronger pure-move proof than
an added-minus-removed set diff, because it proves **order preservation and an exact partition**,
which a set diff cannot. No comment block is severed from the function it documents. **RE-DERIVED.**

### Gate evidence, on the computed merge tree

Merge tree `abf59875`, rc 0, with the negative control fired on a genuinely divergent pair
(`develop` vs `develop-v2` → rc 1). Diff is exactly 7 files. On the merged tree the five slices are
**288 / 67 / 428 / 291 / 60** and `99-remainder.sh` is **2379**. Conservation closes:
288+67+428+291+60+2379 = 3513, +5 build separators = 3518 = the pre-cut row.

The tsv diff on the merged tree is exactly one row replaced by two — `39-describe-change.sh` added at
428 (equal to actual, correct for a file over the 400 floor) and `99-remainder.sh` lowered
3518 → 2379, a **fall**, which needs no equality and is exempt regardless. No other lane's row moved.

Artifact blob `d5378028` is the **same object** at base and on the merged tree, so the artifact cannot
have moved through this cut. `bash -n` returns rc 0 **and zero bytes on stderr** for all six files —
rc alone is not proof of heredoc integrity — with the positive control firing (truncated slice → rc 2,
155 bytes). Heredoc openers are **zero** across all five slices, counted directly, with the grep
control firing at 4 on a known-heredoc neighbour. All **RE-DERIVED.**

CI is **green, including `Shell tests (shellcheck + pithead suite)`** (pass, 12m18s) — the one check
the review explicitly declined to call a pass while it was still pending. That gap is now closed.
**RE-DERIVED.**

**RELAYED, not re-run by me:** the budget-gate script execution and its mutation control, the
`shfmt` run, the whole-artifact `check_disk_grouped` call-site sweep, and the leak scan.

### Merge-record corrections — no force-push is owed

Two figures in the description do not reproduce. Both sit under conclusions that are correct and were
verified independently, which is exactly why neither announced itself. Per standing policy a loose
figure in a description is a merge-record defect, corrected here rather than force-pushed.

1. **"Two of the six boundaries sit above long comment blocks (one 26 lines)."** No 26-line block
   exists at, near or above any of the six boundaries. Measured at all six heads the blocks are
   **4 / 1 / 2 / 6 / 1 / 13** lines, and the longest contiguous unindented comment run anywhere in the
   cut region is **9** lines. The conclusion the figure was offered to support — that the boundaries
   were derived by content rather than by offset — is right, and I verified it independently.
2. **"`prompt_start_stack` is 17 lines."** It is **8** (slice 37, lines 136-143 inclusive). The
   substantive point is unaffected: it is a small stranded UI prompt whose placement by contiguity is
   honest, and an 8-line function is if anything a weaker case for its own file than a 17-line one.

**And one correction against the review itself**, recorded because the review filed exactly this
failure shape against the author: its boundary table gives the 39/40 block as **3** lines; the true
value is **6** (base lines 787-792, the block attached to `apply_dry_run`). Every other entry in its
table matches mine, its longest-run figure of 9 reproduces exactly under its own stated convention,
and the finding's conclusion is confirmed by two derivations. Wrong figure, right conclusion, no tell —
in the same review that named the pattern.

### Judgement calls

**The two declined budget-neutral merges (`37+38` = 356, `40+41` = 352) are correctly declined, on
mechanical grounds rather than taste.** The region contains exactly two `# ---` section headers (base
lines 290, 1079) plus the one opening the remainder; each falls exactly on a boundary and sits at
line 1 of its own file, and no slice has a `# ---` buried mid-file. Merging `37+38` would bury
`# --- Top-level Commands ---` at line 289 of the merged file, and `40+41` would bury
`# --- Subcommand chaining (#94) ---` at line 292. The stated ground for declining is true in both
directions, not a rationalisation. **RE-DERIVED.**

**A 60- or 67-line slice is a group, not a splinter, here.** `41-subcommand-chaining.sh` carries a
self-contained dispatch-validation concern that would blend two unrelated layers if folded into 40.
`38-setup-command.sh` is the weaker call and merging it would not have been unreasonable, but it is
consistent with this series' established fine-grained precedent. Either verdict was legitimate.

**Precision note, not a defect:** the gate does not mechanically compel the new row's exact equality —
`check_monotonic` walks base rows only and structurally never reaches a brand-new row, and `run_gate`
requires only `actual <= ceiling`. Choosing 428 matches the repo's dominant convention rather than
satisfying a constraint. Worth stating precisely in future descriptions. **RELAYED.**

### What was not done, and why

No base/head suite pair: a null experiment for an order-preserving concat cut whose artifact is a
byte-identical object at both ends. The artifact was not rebuilt by hand either — blob identity is the
stronger structural check and it holds. No bench or rig was exercised; nothing here needed one.

Issue #1305, opened 2026-08-23
The sandbox builders leak an implicit WALLET dependency into every domain file


VijitSingh97:
# The sandbox builders leak an implicit WALLET dependency into every domain file

## The class

Five instances of the same trap surfaced while splitting `tests/stack/run.sh` into domain
files: a moved section reads a variable that looks like a lib.sh top-level fixture but is
only assigned inside a sandbox-builder function that used to run earlier in the monolith's
execution order. Consolidation moves the reader ahead of the writer; under `set -u` the
suite aborts (or, worse, a `source "$VAR"` with an unset variable silently no-ops — the
module-6 `$REL` case, where 41 assertions failed because the functions under test were
never defined).

Instances so far, each fixed by a named local re-derivation in the domain file:

- `$REL` in the release-signing file (the silent-no-op case).
- `$WALLET` in the doctor file's onion-report section.
- `$WALLET` read by `control_config()` itself (`lib.sh`, inside the closure
  `build_control_sandbox()` defines) — hit by the control-upgrade file's preamble.
- `$WALLET` twice in the control-deploy file (`cfg_L()` and the scattered-custom-dirs
  case), caught pre-run by auditing the class instead of waiting for the abort.

## The one-line retirement

`WALLET` is assigned by `build_val_sandbox()` (`WALLET="$VALID_PRIMARY"`) and read across
domains. Defaulting it at both sandbox builders' entry points:

```bash
WALLET="${WALLET:-$VALID_PRIMARY}"
```

retires the whole `WALLET` family for every future domain file — including the remaining
Phase 1 cuts (the appliance files and the sandbox-coupled black-box clusters, which are
the most `$V`/`$C`-entangled and therefore the most exposed). The domain files' existing
local re-derivations stay valid and become redundant rather than wrong.

This was deliberately NOT folded into the mechanical-move PRs (wrong blast radius: a
shared-core lib.sh edit inside a "move only" diff). It is a one-line lib.sh change plus a
respelled-mutation check that an unset-`WALLET` domain file now works — small, but it
guards the most error-prone stretch of the remaining split.

## Broader guard worth considering in the same change

An audit sweep for the `source "$VAR"` shape under `set -u` semantics (the `$REL` silent
no-op is nastier than the abort: it turns missing state into green-looking skips). A
one-line guard in the harness — refuse to `source` an empty path — would convert that
silent class into a loud one.

VijitSingh97:
[...]

PR #1518, opened 2026-08-29
refactor(tests): cut the control-channel core out of run.sh (#1105 R12)

VijitSingh97:
Cuts the control-channel core out of `tests/stack/run.sh` into a new domain file. Twelfth module of #1105.

`tests/stack/run.sh` 739 -> 506. The moved text is six black-box sections — the one that builds the control sandbox, plus the five that drive it: `dashboard control channel (#33)`, `apply --dry-run [--porcelain] (#33)`, `symlink-invoked stack renders physical paths (#695)`, `apply --dry-run is read-only re: node credential generation (#556)`, `control-run-pending (#33)`, and `audit log records names, never values (#349)`. The range is CONTIGUOUS and carries no `source "$HERE/` stanza, so nothing nests, nothing reorders, and there is no anchor decision to argue.

**Base: `e886b48` (post-#1516).** The branch was rebased twice this round, and the second one needs stating plainly.

It was first cut and proven at `b817623`. #1514 then merged, which is the landing that genuinely invalidated that proof — it adds `tests/stack/test-appliance-os-update-lock.sh` and its stanza to `run.sh`, so the `tests/stack` tree object really did change. The branch was rebased onto `43d94d1` and **the pair below was re-run there from scratch**.

#1516 (P10) then merged while that pair was running, and the branch was rebased again onto `e886b48`. **The pair was NOT re-run for #1516, deliberately — re-running it would have been a null experiment.** The argument, with the objects:

| | pair's base `43d94d1` | current base `e886b48` |
|---|---|---|
| `tests/stack` tree | `50052122` | `50052122` |
| built `pithead` blob | `d5378028` | `d5378028` |

**The AFTER leg was then re-run from `HEAD` itself**, after two header amends. Those amends changed the tree, so the earlier after-proof no longer covered the committed objects and reusing it would have shipped exactly the proven-tree/committed-tree gap this lane's own rule forbids. The re-run's recorded head sha equals `HEAD`, so there is no chain to argue: the after leg ran the commit this PR ships.

The before leg therefore executed *the same objects this PR ships*. #1516 changed five paths — `docs/dev/file-budget.tsv` and `lib/pithead/{34,35,36,99-remainder}.sh` — and the suite consumes none of them: `lib/pithead` has ZERO references under `tests/stack/` (positive control: `build_control_sandbox` matches 13 files there, so the grep works), and the three `file-budget` hits under `tests/stack/` are all comments. `lib/pithead/` is the SOURCE the root `pithead` is built from, and that built blob is byte-identical across both bases.

This is R10's established baseline-reuse test, and it is the right test here: "`tests/stack` is byte-unchanged" would be the wrong one. A second pair over a byte-identical program would produce a green result that proves nothing the object hashes do not already prove, while reading like independent evidence.

## Proof

- **Suite pair at this base.** before rc 0, after rc 0, both **`2992 passed, 0 failed`**. The after leg ran from the exact commit this PR ships (the leg recorded its own head sha before starting: `5f7e5604`, which is `HEAD`), launched only after the last header amend — the standing rule here is amend-the-header-*before*-the-after-proof, because a header amended afterwards leaves a proven-tree/committed-tree gap that git cannot settle. Two earlier after-legs were discarded for precisely that reason rather than reused. **Checkable prediction, made before this run:** `git diff <discarded after-leg> HEAD -- tests/stack/` is comment-only (zero non-comment `+`/`-` lines), so the executed program was unchanged and the re-run had to reproduce the discarded leg's output modulo the known nondeterminism. It did: the discarded leg and the re-run differ in exactly two line pairs, both nondeterministic, with identical header set, identical order and 2992 verdict lines each side.
- **Ordered comparison (#1398).** Header SET identical, header ORDER identical — **0 of 275 positions differ**. Totals identical at 3306 log lines / 275 headers. The ordered check is the only instrument that can see a permutation; a sorted multiset destroys order by definition. The comparator's per-verdict check matched **2992 lines on each side**, not zero — that check was a null instrument for rounds R7-R10 and this is it demonstrably firing.
- **Whole-output sorted multiset:** residue is exactly TWO differing line pairs, both from the documented benign classes — the #1325 ed25519 keygen fingerprint, and the `load_baked_images` heartbeat elapsed seconds (1s vs 2s). Nothing outside those classes differs. **Predicted before the run and met exactly:** #1514 adds one `echo "== "` header and no mktemp/date/keygen nondeterminism, so 275 headers was derived from the base file rather than carried from the previous round's 274.
- **Byte-identity, re-derived FROM GIT rather than from the extraction.** The new file below its header equals base `run.sh` at the cut range (diff rc 0); a one-line shifted range makes the same diff FIRE (rc 1). Doing it from git rather than from my own chunks file is what makes it a proof instead of a tautology.
- **Standalone-sourceable under `set -u`:** the builder call travels with the section that always made it, so the new file provides its own `$C` and `$CTRL_LOG`.
- **Every other file in the diff is comment-only.** Outside `run.sh`, the new file and the one tsv row, all ten touched files have **zero** non-comment changed lines — measured, not asserted: strip `+++`/`---`, then every remaining `+`/`-` line is a `#` comment or blank, for each of the ten. So the entire executable surface of this PR is the moved block plus the source stanza.
- **The retraction shipped at exactly five sites**, each as a visible `RETRACTED (#1105 R12):` block with the original claim left standing above it, per COMMON's prescribed form. Control: the same grep returns zero at the base.

## One file, not the two the cut map's R12 row names

The map proposes `test-control-apply.sh` + `test-control-audit.sh` with the audit half self-arming. This ships ONE file, and the deviation is **ratified by the controller and recorded in the map's R12 row**.

The apply half ends by removing the result and staged files its own #556 leg wrote — its own comment says the counters below assume a clean spool — and the very next section asserts the results dir holds exactly one file after a malformed id. That exact count is established ACROSS the proposed seam. An armed audit half would be green standalone (an empty spool yields the same one file) AND green in position, by two DIFFERENT mechanisms — so its own verification could not distinguish "the arm works" from "the arm was never needed". That is the fixture-never-armed shape. A builder-only arm stanza is weaker still: it binds the names under `set -u` and establishes none of the state.

## Ten headers went false — five, then eight, then ten

A cut falsifies PROSE in files its diff never touches, and every instrument above is blind to it by construction.

My first sweep keyed on the PHRASE "a control-core section that stays behind", matched exactly five files, and **looked complete**. Five was a floor. The instrument that does not depend on guessing the wording:

```
diff <(git show <base>:tests/stack/run.sh | grep '^echo "== ') \
     <(grep '^echo "== ' tests/stack/run.sh)      # = exactly what MOVED
grep -rFn -- "<bare header name>" tests/stack/*.sh | grep '#' | grep -F 'run.sh'
```

Use `grep -F` (a header containing `[--porcelain]` is an invalid regex, and the failing loop still printed a clean verdict), strip the `black-box: ` prefix (prose quotes the bare name), and prove it discriminates on the pre-edit blobs before believing a zero. **It is a floor, not the answer, and it was measured as one twice**: it took the count from five to eight, and then two further sites were found past it — so even after the correction it was missing 2 of 10. It is phrasing-independent on the MOVED side and still phrasing-dependent on the REFERENCING side. Read the neighbourhood of every hit.

Amended: `test-secrets-masking.sh`, `test-confirm-approval.sh`, `test-data-management.sh`, `test-spool-audit.sh`, `test-control-upgrade.sh` (the five), plus `test-rig-worker.sh`, `test-dashboard-onion.sh`, `test-worker-config.sh` (the instrument's eight), plus `test-dashboard.sh` and `test-control-provisioning.sh`. `test-control-upgrade.sh`'s clause also had the source order backwards — pre-existing, corrected in the same sentence.

**The last two were found after the instrument above reported clean, and each defeats it differently.** `test-dashboard.sh` placed the `dashboard control channel` section in run.sh, quoting the header WITHOUT its `(#33)` — the known suffix blindness. `test-control-provisioning.sh` said `build_control_sandbox` is "zero in this file, one in run.sh"; after this cut it is **zero** in run.sh. That one locates the moved block by naming its CALL SITE instead of its section header, which a header-name sweep cannot see by construction.

**Both were also invisible to any line-oriented grep**, because comment reflow splits each claim across a line break — neither `run.sh` nor the phrase it governs appears on a single line. The sweep that finds them joins each file's comment block into one whitespace-normalised string and windows around every `run.sh`. It is in the lane's tools as `prose-locate-sweep.py`. It needed the whitespace strip to work at all: preserved indentation injected extra spaces into the join and broke a single-space needle, which hid `test-dashboard.sh` from the very instrument written to catch it — caught by using the known-missed site as the control.

## A retracted mechanism, at all five sites that carry it

`test-rig-worker.sh` explained a recorded RED 2901/1 by saying `./pithead apply -y` on a fresh sandbox writes into `$C/data/control/results/`; four consumer headers relay that as "its applies wrote extra result files into the shared results dir". **The mechanism does not reproduce at the tip**, re-derived four independent ways: every `control_write_result` call sits on the run-pending/upgrade/verb/backup paths and none on apply; the only apply-path write into `results/` is `os_state_write` inside `prepare_control_dirs`, gated behind `is_appliance()`; `is_appliance()` needs `PITHEAD_APPLIANCE=1` or both `/etc/rauc/system.conf` and `/usr/local/sbin/pithead-install`, which no sandbox run has; and `control_config()` writes only `$C/config.json`.

**The RED was real. WHY is not established** — the attempt was never committed, so no tree can settle it — **and no replacement mechanism is guessed in place of the refuted one.** The rule the episode was cited for (a pure consumer of the control sandbox must not build a second one) rests on its own evidence.

## Budget

run.sh's row 739 -> 506, **re-derived from `wc -l` after the rebase, not resolved as a merge conflict**. The three lines this row used to bank for #1482's domain stanza are **CONSUMED, not carried**: #1514 landed that stanza, so run.sh arrived at this base at 739 actual against its 739 ceiling, and after the cut the row again equals the file exactly with zero headroom. No other row moves — `--generate` rewrites every row and silently lowers files a cut never touched, which would red every un-rebased branch in every lane, so the diff is one line.

Re-derived at source rather than inferred from the merge record: #1514 is titled *"cover the os-install contention verdict, and arm the chain to prove it (#1482)"* and adds `tests/stack/test-appliance-os-update-lock.sh` plus its stanza — it is exactly the new-test-file beneficiary R11's carry-forward ruling named, and it spent the bank itself. Independently and without needing that reading: at this base `run.sh`'s ceiling and its actual line count are both 739, so there is arithmetically nothing left to carry. **#1482 itself is still OPEN** (most of its six verbs are unshipped), and this PR does **not** re-bank for them: a future test file for the remaining verbs needs its own stanza and its own disclosure, and banking now for an unnamed future beneficiary is the anonymous-slack shape the ruling forbids.

**FOUR** files in this diff sit at their ceilings with zero headroom: `test-rig-worker.sh` (617/617), `test-dashboard-onion.sh` (422/422), `test-control-upgrade.sh` (795/795) and `test-dashboard.sh` (592/592). All four edits are line-neutral, verified by measuring each file's line count at the base against HEAD rather than by inspection. Note the commit message names only the first three: `test-dashboard.sh` joined the diff in the last amend and its list was not re-derived. The four here is the checked number — I am flagging the discrepancy rather than amending, because the after-proof below was launched from this exact commit and re-anchoring it to a new sha to fix a message would cost more than it buys.

The six other files this diff touches GREW, by one to six lines each (`test-confirm-approval.sh` 142→147, `test-control-provisioning.sh` 150→151, `test-data-management.sh` 126→131, `test-secrets-masking.sh` 219→225, `test-spool-audit.sh` 138→143, `test-worker-config.sh` 127→128). All six edits are comment-only (see the proof section: zero non-comment changed lines), all six are well under the generator's 400-line floor, and none has or needs a tsv row. The new file, at 301 lines, is likewise under the floor and correctly takes no row.

## Not fixed, deliberately

`test-release-signing.sh` carries the same stale class — it says the control-channel upgrade's bundle-signature check stays in `run.sh`, and it lives in `test-control-upgrade.sh`. R12 does not touch that file, so fixing it would be a widening. Flagged for whoever cuts near it.

Two further sites surface in the reflow-proof sweep and are deliberately left, because neither file is in this diff and each is stated rather than left for a reviewer to trip over:

- **`lib.sh`, inside `build_control_sandbox()`** — "In run.sh the two calls sit hundreds of lines apart in ONE process and val happens to run first." Before this cut, one of those two calls was in `run.sh` and the other was already in a domain file; after it, neither is. The sentence's substance — the ordering accident, and the `${WALLET:-...}` default that retires it (#1305) — is untouched, and the comment's own parenthetical already anticipates that "every #1105 module moves it". R12 makes it looser, not false. `lib.sh` is not in this diff.
- **`test-appliance-identity.sh`** — "in the suite it survives only because an earlier `run.sh` section happened to build a sandbox first." This one is **pre-existing and unaffected**: at the base, every `build_val_sandbox` call site was already in a domain file, not a `run.sh` section. R12 changes nothing about it.

## Simplification pass

On a move PR there is exactly one authored surface, and this is what it measures out to. The whole diff adds **174 executable lines**: 173 of them are the moved block, which `comm` confirms contains **zero** lines not already present in base `run.sh`, and the 174th is the single source stanza. No new assertion, no new fixture, no new helper.

The new file's header is 66 lines of a 301-line file — 22%, which sits mid-range against the shipped domain files it will live beside (`test-control-upgrade.sh` 3%, `test-control-add-only-ssrf.sh` 13%, `test-control-core.sh` 22%, `test-control-provisioning.sh` 23%, `test-worker-config.sh` 26%, `test-spool-audit.sh` 48%) and is smaller in absolute lines than `test-spool-audit.sh`'s 69. Its four blocks each earn their place: the domain description, the standalone-sourceable argument, the one-file-not-two argument (a contested decision, and the file is where a future reader will look for it rather than a merged PR body), and the dependency re-derivations this lane's own rule requires.

One clause was considered for cutting and deliberately kept: the dependency list notes that `line` is bound by `while IFS= read -r line` and that the lane's audit tool does not model `read -r`, so it misreports the name. The second half documents a tool nobody reading this repo has. It stays because it pre-empts a false finding this lane has already paid for, and it is one clause on a line that has to exist anyway.


VijitSingh97:
## Verification re-derived at this head, with the control that fired for each

Everything below was re-run against `5f7e5604` in this session rather than relayed from the round that produced the PR body. Where a check could pass vacuously, the control that makes it fire is named next to it.

| Check | Result | Control that fired |
|---|---|---|
| Suite pair | before rc 0 / after rc 0, both `2992 passed, 0 failed` | comparator's per-verdict pattern matched **2992 lines per side**, not zero — the null-instrument shape it was fixed for |
| Header SET / ORDER | identical; **0 of 275 positions differ**; 3306 log lines both sides | ordered diff is the only check that can see a permutation |
| Residue | exactly two line pairs, both benign: #1325 ed25519 keygen fingerprint, and `load_baked_images` heartbeat elapsed seconds (1s vs 2s) | nothing outside the documented classes differs |
| After-leg anchoring | the leg recorded head `5f7e5604` before starting; that is `HEAD` and the PR's head sha | two earlier after-legs discarded, tree shas `4cdb2f08` / `6df4b6c3` / `5f0b0335` genuinely differ |
| Byte-identity from git | base `run.sh` 122-356 == new file 67-301, diff rc 0 | one-line shift makes the same diff FIRE, rc 1 |
| Nested stanzas in the moved range | **ZERO** | positive control: the same pattern matches 45 stanzas in base `run.sh` |
| Stanza placement | `run.sh:123`, the exact vacated position — displacement zero | header ORDER identical confirms it end to end |
| Baseline reuse across the #1516 rebase | `tests/stack` tree `50052122` and built `pithead` blob `d5378028` identical at `43d94d1` and `e886b48` | 0 `lib/pithead` refs under `tests/stack`, against 13 files matching `build_control_sandbox` |
| Budget gate | rc 0; base ref `e886b48` probed unmoved before **and** after | understating the row to 505 makes it FAIL **by name**; mutation proven applied by reading the row back; tsv restored byte-identical to `git show HEAD:` |
| Line-neutrality, four zero-headroom files | 617/617, 422/422, 795/795, 592/592 — all four measured base against HEAD | six other touched files grew 1-6 lines, all rowless and under the 400 floor |
| Executable surface | 174 added executable lines: 173 the moved block, 1 the stanza | `comm` finds **zero** new-file executable lines absent from base `run.sh`; all ten other touched files have zero non-comment changed lines |
| `shfmt -i 4 -d` | rc 0 | misindented copy fires rc 1 |
| `bash -n`, 12 changed files | rc 0 each | truncated `if` fires rc 2 |
| `shellcheck --severity=warning` | rc 0 on `run.sh` alone (the second `lint-sh` invocation verbatim) and on `lib.sh` + every touched test file, under a 3GB `ulimit -v` | an appended unused global fires SC2034, rc 1 |
| Prose sweep, header-name form | zero hits on the amended tree | discriminates: the same sweep scores 2 on the base blobs |
| Prose sweep, reflow-proof form | 23 windows flagged, every one read | discriminates: 31 at base; the residual hits are the disclosed ones, plus the two `Not fixed, deliberately` sites |
| Retraction | exactly five `RETRACTED (#1105 R12):` sites | same grep returns zero at base |
| Map ratification | the R12 row says it in the controller's own words: *"DEVIATION RATIFIED BY THE CONTROLLER (w26): ONE file, `tests/stack/test-control-core.sh`, NOT the two this row names."* | read at source, not relayed |
| Bank consumption | #1514 is titled *"...(#1482)"* and adds the new test file plus its stanza; independently, base ceiling and base actual are both 739 | `gh issue view 1482` is **OPEN** — checked the ISSUE's state, not the PR's |

### What I did NOT do
- `make lint` and `make test` were not run whole. `make lint-sh` is the OOM path that has killed sessions on this box, so the two `shellcheck` invocations were run directly under a memory guard instead, and `shfmt`/`bash -n` per file. **CI is the backstop for the rest.**
- The retraction's four refutation routes are relayed from the round that derived them, not re-derived here. They are laid out in the commit message and the PR body so a reviewer can attack them at source.
- The two-mechanisms symmetry underpinning the one-file decision is argued, not machine-checked; the structural half (a cleanup and a dependent count assertion either side of the map's proposed seam) is the part that was re-derived.

An adversarial review is dispatched against this head with a brief that names the rebase, the after-leg anchoring, and the prose floor as its sharpest axes.

VijitSingh97:
## The tip moved after this PR opened (#1517 merged). No rebase is owed, and here is the check rather than the reasoning.

`develop-v2` advanced to `b57045d`. This PR's merge base is still `e886b48`, so the question is whether the proof above still describes the program that will run.

**It does, by the same object-identity test the body already uses — re-derived at the new tip:**

| | base `e886b48` | new tip `b57045d` |
|---|---|---|
| `tests/stack` tree | `50052122` | `50052122` |
| built `pithead` blob | `d5378028` | `d5378028` |

#1517 touched seven paths — `docs/dev/file-budget.tsv` and six `lib/pithead/*.sh` slices — and the suite consumes none of them: each has **zero** references under `tests/stack/`, against a positive control of 13 files matching `build_control_sandbox`, so the grep works. `lib/pithead/` is the source the root `pithead` is built from, and that built blob is byte-identical across both, which is what `lint-pithead-parity` exists to guarantee.

**The budget gate was the real risk here, and it was run, not reasoned about.** Every worktree shares one `.git`, so fetching the new tip moves the gate's base mid-flight — and the documented tell of that trap is a FAIL naming files the diff never touched. #1517 **lowered** `lib/pithead/99-remainder.sh` from 3518 to 2379, which is exactly the direction that reds un-rebased branches. Gate re-run with the base at `b57045d`: **rc 0**, base probed unmoved either side of the run. It passes because that specific row carries the #1464 both-ways exemption, and the gate says so in its own words:

```
file-budget: NOTE — lib/pithead/99-remainder.sh's ceiling rises from 2379 to 3518, matching the file.
That row records the un-split remainder of the generated pithead artifact, so it tracks it both ways (#1464).
```

Worth stating plainly rather than banking: **this pass is narrower than it looks.** The only row #1517 lowered is the one file in the repo with a documented exemption in both directions. Had it lowered any other, this un-rebased branch would have red on a file its diff never touches, and the fix would have been a rebase — never an exemption. So read this as "checked and clean at this tip", not as "a moving tip is safe for #1105 branches".

VijitSingh97:
## Adversarial review returned — no blocking finding. CI fully green (`Shell tests` pass).

The review re-derived every load-bearing static claim with its own firing controls and agreed: byte-identity of the moved block (empty diff, control fires on a one-line shift), zero nested stanzas (44 elsewhere in the base file, so the pattern works), the tree/blob object identity across both rebases, the budget arithmetic including all four zero-headroom files and all six grown files, source order, the coupling rules, and the header/column/scope checks. It also independently confirmed the ratification is in the cut map rather than only in this body.

Three findings, and one gap in the review itself. I have re-derived all of them.

### 1. A genuine ELEVENTH stale site — and re-deriving it turned up **nine**, not one

The reviewer found `tests/stack/test-appliance-identity.sh:21-22`, which says `provision_control_runner`'s "ownership and foreign-install guards ... stay behind in `run.sh`". Confirmed: `grep -c provision_control_runner tests/stack/run.sh` is **zero** (positive control: `domain_ran` returns 45), and both behaviours are in `test-control-provisioning.sh` at `:37` and `:88`.

**It is not R12's**: the same grep at this PR's base `e886b48` also returns zero. It was moved by an earlier round.

Chasing the mechanism rather than the site found a cluster the reviewer did not. Deriving needles mechanically — every `^echo "== ` header and every function name currently **absent** from `run.sh` — surfaces **seven more files** all carrying the same copy-pasted `$V`/`$WALLET` paragraph, each saying `run.sh`'s `"config validation"` black-box calls `build_val_sandbox()`, several adding "that section stays in run.sh": `test-backup.sh:30`, `test-dashboard.sh:22`, `test-monero-tari.sh:21`, `test-rig-worker.sh:38`, `test-secrets.sh:37`, `test-lifecycle.sh:17`, `test-tor-network.sh:19`. `config validation` is **zero** in `run.sh` and lives at `test-config.sh:208`. Also pre-existing at this PR's base.

**Filed as #1520 rather than fixed here.** Nine sites across nine files, none of them in this diff, is a widening — the same call this program already made for #1445. Two things in that issue are worth more than the list:

- **No per-cut sweep can find these by construction.** The #1105 sweep keys on what the *current* cut moved, so debris from *earlier* rounds is out of scope and every cut's sweep reports clean over it. It needs one suite-wide pass, not another per-cut instrument.
- **Nine is a floor, and I can name the mechanism.** My probe is line-oriented, and that is exactly why it missed the reviewer's site: `provision_control_runner` is on line 21 and "stay behind in `run.sh`" on line 22. Two instruments, each blind where the other sees. Neither found all nine alone.

### 2. "Four independent routes" is an overcount — accepted

The reviewer re-derived two routes verbatim from `pithead` and is right that they are not independent: `os_state_write` being gated behind `is_appliance()`, and what `is_appliance()` requires, are one causal chain stated in two sentences. Three distinct arguments, not four. **The conclusion is unaffected** — the reviewer's own source reading agrees the mechanism does not reproduce — but the independence count in this body and in the commit message is overstated and should be read as three.

### 3. "The audit half cannot be armed honestly" is stronger than what was shown — accepted

The reviewer proposes a shape neither this body nor the cut map considered: a self-verifying **dirty-then-clean** arm that seeds a known dummy into `$RESULTS`/`$STAGED` and removes it, so standalone and in-position pass by the *same* demonstrated mechanism rather than by two different ones. That is a real answer to the fixture-never-armed objection. It does not overturn the decision — the ratification rests on cost/benefit (budget-neutral, reversible, no verifiable arm was *demonstrated*), not on impossibility — but this body reads as if the option space were exhausted, and it was not. Read the claim as "no arm was demonstrated", not "no arm exists".

### The gap in the review, stated because it is the strongest claim here

**The reviewer could not check the suite logs and says so.** It relayed the pass counts, the two nondeterminism-class residuals and the `0 of 275` header-order comparison rather than reproducing them, because it found no log artifact — my brief never told it the logs are in `~/split-baselines/`. That is my defect, not the reviewer's: **the single instrument nothing else can substitute for went unverified by the review.** It reached the same conclusion by a different route (it re-derived that the discarded legs' diffs against HEAD are comment-only, so the executed program is unaffected whichever point launched the proof), but that is independent evidence for the same conclusion, not the same check.

Nothing here changes the diff. Two prose overstatements are acknowledged above rather than amended, for the reason already given in the body: the after-proof is anchored to this exact sha.

VijitSingh97:
## Controller verdict — MERGE

Base `e886b48d`, reviewed head `5f7e5604`. Adversarial pass returned **no blocking finding**; I judged it against derivations of my own rather than deferring to it. Below, every claim is labelled **re-derived** or **relayed**.

### The claim the pass could NOT check, and I did

The pass explicitly labelled the runtime-log evidence as relayed — it found no artifact on disk, because the brief never named where the logs live. That is the single strongest claim in this PR and the only instrument that can see a permutation, so I re-derived it from the leg artifacts directly.

**Ordered header comparison (#1398) — RE-DERIVED.** From `before.log` and `after3.log`:

- raw line counts **3306 / 3306**; header counts **275 / 275**; header SET identical
- positional comparison: **0 of 275 positions differ**
- **control FIRED**: swapping two adjacent headers on one side yields 2 differing positions, so the comparison discriminates and is not a null instrument

**Whole-output multiset residue — RE-DERIVED.** Raw residue is 10 pairs, all path-shaped: the two legs ran in different worktrees (`pithead-split-proof` vs `pithead-split-work`) and under different `mktemp` dirs. Normalising only the worktree root and the `mktemp` path leaves **exactly two differing pairs**, both from the documented benign classes:

- `a fast load does not wait on the heartbeat interval (1s)` vs `(2s)` — the `load_baked_images` heartbeat
- the ed25519 keygen fingerprint (#1325)

**Control FIRED**: injecting one changed line into the normalised after-side raises the residue to 3/3, so the normalisation is not blanket-eating differences.

**Verdict counts — RE-DERIVED.** `2992 passed, 0 failed` and 2992 verdict lines on **each** leg, counted from the logs, not read off the summary line alone.

**The proven-tree/committed-tree claim — RE-DERIVED.** `after3.head` records `5f7e56042d41e707db69c370ed1cb45b408ba059`, which is exactly this PR's head, and it was written before the leg started. The after leg ran the commit this PR ships. Three after-legs exist on disk (`after`/`after2`/`after3`) with distinct log hashes and distinct recorded heads — the discards are visible in the artifacts, matching the body's account.

### Budget — run on the MERGED TREE, not reasoned about

The tip moved under this review (#1517/P11 landed), so `merge-tree` rc 0 alone proves no conflict and says nothing about the ratchet. I ran the gate rather than arguing from which row moved.

- merged tree computed before merging: `0d04603a`
- gate **self-test rc 0** first (so the gate itself fires), then the gate on the merged tree: **rc 0**
- tsv diff tip -> merged tree is **exactly one row**: `tests/stack/run.sh 739 -> 506`. Rows: 87 at both ends, **zero raised**, one lowered, none added, none dropped. Ruling 9's check holds — the diff carries only the row this cut owns.
- **the pass does not depend on the `99-remainder.sh` exemption.** That row is unchanged at 2379/2379 across tip and merged tree, and no row is raised at all, so the ratchet is satisfied exemption-independently. Recording this because the previous round survived a tip move only via that exemption and correctly refused to bank it as general safety.

Rows read on the merged tree, not predicted: `run.sh` 506/506; `test-rig-worker.sh` 617/617; `test-dashboard-onion.sh` 422/422; `test-control-upgrade.sh` 795/795; `test-dashboard.sh` 592/592. The new `test-control-core.sh` is 301 lines and correctly carries no row.

Artifact identity through the merge: the built `pithead` blob is `d5378028` at the tip, at the reviewed head, and on the merged tree — unchanged. `lib/pithead` on the merged tree equals the tip's tree, as a tests-only cut requires.

### On the pass's findings

**1. The eleventh stale site is REAL, and CONFIRMED PRE-EXISTING — not a merge condition.** `test-appliance-identity.sh:21-22` claims `provision_control_runner`'s ownership and foreign-install guards "stay behind in `run.sh`". Re-derived: zero hits for that name in `run.sh` at the **base** as well as at the merged tree, with the pattern's control firing at 5 hits in `test-control-provisioning.sh`; and the comment text is byte-identical at base and merged. So it was already false before this PR existed. A pre-existing defect and an introduced one are different classes, and only the second may become a merge condition.

**It is also already filed.** #1520's cluster 2 is this exact site. Worth stating precisely: the pass finished at 11:39:04Z and #1520 was filed at 11:41:23Z — two minutes apart, independently. Convergent discovery by two instruments, not a missed filing. No new issue is owed.

**2 and 3. The two overstated arguments stand as merge-record corrections, not force-pushes** (standing ruling 2: where the wrong text lives decides the class — a loose claim in a PR body is corrected here, in the record).

- *Axis 6(a) — "four independent routes".* The pass is right that two of the four share one premise: the `os_state_write` gate and what `is_appliance()` requires are one causal chain stated in two sentences. Substantiable as at most three distinct arguments. The conclusion — that the retracted mechanism does not reproduce at the tip — is not affected, and the pass verified it independently from source.
- *Axis 0(b) — "cannot honestly arm".* A self-verifying dirty-then-clean arm (seed a known non-UUID file, delete it, then proceed) was not in the option space the body considers, so "cannot be armed honestly" is stated more strongly than shown. The ratification itself rested on cost/benefit and on no verifiable arm having been *demonstrated*, not on impossibility, so the one-file decision stands as ratified.

### What I did NOT do

- **No suite pair of my own.** Null experiment: the executable surface is a moved block that `comm` shows introduces zero new lines, and the artifact blob is byte-identical at both ends.
- **Did not re-run the after leg.** I re-derived from its recorded artifacts and its recorded head instead, which is the check that was actually owed.
- **Did not verify retraction route 4** (`control_config()` writes only `config.json`) — the pass spot-checked route 3 and did not check route 4 either. Named rather than left silent; it is one strand of an argument whose conclusion is separately established.
- **Did not re-check two of the five retraction sites by name** — the pass verified three; I did not extend that.

Issue #1520, opened 2026-08-29
tests/stack: stale prose locates moved sections in run.sh — at least nine pre-existing sites from earlier #1105 rounds


VijitSingh97:
`tests/stack` carries stale prose from earlier #1105 rounds that locates moved sections in `run.sh`. At least **nine** sites across **nine** files. Systematic and pre-existing, so it is no single cut's debt — filing rather than widening a diff into it, the same call as #1445.

## The two clusters

**1. The `$V` / `$WALLET` boilerplate — seven files.** Each says `run.sh`'s `"config validation"` black-box section calls `build_val_sandbox()`, several adding "that section stays in run.sh":

- `tests/stack/test-backup.sh:30`
- `tests/stack/test-dashboard.sh:22`
- `tests/stack/test-monero-tari.sh:21` (and `:23`)
- `tests/stack/test-rig-worker.sh:38`
- `tests/stack/test-secrets.sh:37`
- `tests/stack/test-lifecycle.sh:17`
- `tests/stack/test-tor-network.sh:19` (and `:21`)

`grep -n 'config validation' tests/stack/run.sh` returns **zero**. The header lives at `tests/stack/test-config.sh:208`. Positive control: `grep -c '^echo "== ' tests/stack/run.sh` returns 3, so the file does still contain headers and the grep works. The paragraph looks copy-pasted, which is why one edit did not catch it and seven files drifted together.

**2. `provision_control_runner` — one file.** `tests/stack/test-appliance-identity.sh:21-22` says its "ownership and foreign-install guards ... stay behind in `run.sh`". `grep -c provision_control_runner tests/stack/run.sh` returns **zero** (positive control: `domain_ran` returns 45). Both named behaviours are in `tests/stack/test-control-provisioning.sh` — the ownership assertion at `:37`, the foreign-install guard at `:88`.

Also `tests/stack/lib.sh:213` and `tests/stack/test-control-upgrade.sh:19` describe `build_val_sandbox()`'s call as living in `run.sh`. Both are looser than false — they describe a real ordering accident and `lib.sh`'s own parenthetical anticipates that #1105 moves it — so they are worth rewording, not correcting.

## Why no per-cut sweep will find these

The #1105 sweep is keyed to **what the current cut moved**. A site made false by an *earlier* round is out of its scope by construction, so the debris accumulates invisibly and every cut's sweep reports clean over it. This needs one suite-wide pass, not another per-cut instrument.

[...]

PR #1521, opened 2026-08-29
refactor(cli): cut the OS-update verbs, the control loop and main dispatch — 99-remainder.sh retires (#1105)

VijitSingh97:
The capstone of #1105 Phase 2. Cuts the OS-update verbs, the control-request loop, runner
provisioning and `main` dispatch out of `lib/pithead/99-remainder.sh` — **consuming it entirely, so
the file and its budget row both retire.** Row P13, the last of P0-P13.
**Pure move — no logic changes, no behaviour changes.**

| slice | lines |
|---|---|
| `lib/pithead/47-os-update-helpers.sh` | 145 |
| `lib/pithead/48-os-update-verbs.sh` (the five `control_os_*` verbs) | 364 |
| `lib/pithead/49-control-request-loop.sh` | 122 |
| `lib/pithead/50-control-runner-provisioning.sh` | 136 |
| `lib/pithead/51-main-dispatch.sh` | 178 |
| `lib/pithead/99-remainder.sh` | **DELETED** |

Conservation: 145+364+122+136+178 = 945 slice lines + 4 build separators = **949**, the remainder's
exact size before the cut. Every slice is under the 400 floor, so **no new budget row is added** —
and the `99-remainder.sh` row is **deleted rather than lowered**, because the file it tracked is gone.

## What was run, at the head that ships

**1. Exact partition — and here it is total, not partial.** The five slices concatenated in build
order with one blank line between each pair reproduce the pre-cut `99-remainder.sh` byte for byte:
sha256 `6d48f5d7dd62df1e...b19cf` on both sides, `cmp` identical at 53215 bytes. There is no
surviving remainder to append — the region IS the file. **Negative control fired:** dropping a
single blank separator makes the reconstruction differ at byte 31848, line 511, with the sha256
diverging too (two routes, and the rc read without a pipeline in between).

**2. Artifact byte-identity.** git blob `d537802802865addfa73ab75f8d68513c5a50cfd` on both base and
head, and a literal `cmp` clean at 713935 bytes. **Stating the limit rather than overselling it:
these are two instruments reading the SAME bytes, not two independent routes** — a stale artifact
would fool both identically. What excludes that reading is proof 3, not a second hash.

**3. The parity gate, at 52 slices with no remainder file at all.** `make lint-pithead-parity` rc
**0**, "exactly what 52 slice(s) build". This is the claim no earlier row in this series exercised:
**nothing in the build depends on `99-remainder.sh` existing.** Armed state was read back before the
rc — remainder ABSENT, 52 slices, tsv row gone — because an unarmed control returns the answer you
were hoping for. **Negative control fired on this exact slice set:** a blank line appended to
`51-main-dispatch.sh` made the build refuse BY NAME, rc 1. **Positive control:** an unmutated copy
of the same scratch root builds rc 0 at 52 slices — without it the negative control would only show
the harness can fail, not that it failed because of the mutation.

**4. Boundary placement — the axis every instrument above is blind to.** Any legal partition
rebuilds the identical artifact, so `cmp`, parity and the budget gate all pass for a cut whose
boundaries fall in the wrong places, and `bash -n` passes too since severing a function from its
comment block is still valid shell. Read at source instead. All four boundaries sit at a top-level
`}` in column 0, then exactly one blank line, then a comment block or a banner:

| boundary (region line) | before | after |
|---|---|---|
| 146 | `}` closing `os_update_*` helper at 145 | `control_os_check`'s 4-line comment block at 147 |
| 511 | `}` at 510 | `control_process_request()` header at 512 |
| 634 | `}` at 633 | `control_unit_dir`'s comment block at 635 |
| 771 | `}` at 770 | `# --- Main Execution ---` banner at 772 |

These boundaries were derived by reading the pre-cut file, not carried from the cut map, whose
printed ranges are advisory. **The build's edge guard was fired as a negative control against this
exact slice set**, with the mutation read back as applied before the outcome (364 -> 365 lines, last
two bytes `\n \n`): appending a blank line to `48-os-update-verbs.sh` makes the build refuse
**by name** — `FATAL — 48-os-update-verbs.sh ends with a blank line` — rc 1. So a boundary landing
one line off, on the blank rather than after it, is caught loudly rather than silently absorbed.

*Relayed, not re-derived here:* an earlier probe recorded a guessed boundary near region line 150
failing because lines 147-150 are `control_os_check`'s comment block. I did not reproduce that
specific guess, and I am labelling it rather than presenting it as my own result. The boundary table
above stands on the source reading plus the fired edge guard, neither of which depends on it.

**5. Heredocs — and this row is not like P12's.** `50-control-runner-provisioning.sh` holds **2**
real heredoc openers (slice lines 113 and 123, both `<<EOF`), closing at slice lines 122 and 132.
Both open and close **inside** the slice, and the nearest boundary is the slice edge at line 136 —
no boundary falls inside a heredoc. The other four slices hold **0**. Pattern control fired at 13 on
`36-quadlet-units.sh`. Both heredocs use the literal `EOF`, so an opener COUNT is not trustworthy on
its own (same-delimiter resync); `bash -n` on the slice alone is the load-bearing guard and it is
clean below.

**6. Per-slice syntax.** `bash -n` over all **52** slices individually: 0 failures and **zero stderr
bytes** (rc alone is not enough). `shfmt -i 4 -d` on the five new files: rc 0, **0 diff bytes**.

**7. Budget gate — the novelty of this row, proven in all three directions.** P13 **deletes** the
`lib/pithead/99-remainder.sh` budget row rather than lowering it, because the file it tracked ceases
to exist. No earlier row in thirteen exercised that path, so it is proven with a fired control on
each side rather than cited with an rc.

**Asserted on each configuration's own FAIL LINE, never on rc.** That matters here: all three
configurations can return 1, so rc is a single bit summarising three different doors — the
over-target loop, the stale-entry loop, and neither. Each case below records the presence AND the
absence of both FAIL lines, with an arming readback taken before the outcome.

| state | over-target FAIL line | stale-entry FAIL line | rc |
|---|---|---|---|
| file gone + row gone (**what P13 ships**) | absent | absent | 0 |
| file present (949 lines) + row gone | **present** — control fired | absent | 1 |
| file gone + row present | absent | **present** — control fired | 1 |

The two refusals, verbatim:
`FAIL — lib/pithead/99-remainder.sh is 949 lines, over the hard ceiling of 800 for a new/unbudgeted file (target: 400).`
and `FAIL — docs/dev/file-budget.tsv lists 'lib/pithead/99-remainder.sh', which no longer exists (or is now exempt). Remove the entry.`
So deleting the row is **mandatory, not merely permitted** — the gate refuses both alternatives by
name. `--self-test` rc 0. `origin/develop-v2` probed before and after and unmoved at `f946ce7`.

**⚠ How the second control first failed to arm, recorded because it read exactly like a result.**
My first run of the file-present case wrote the file to disk but did not stage it, and the gate
returned **rc 0** — which reads as "the gate tolerates a 949-line unbudgeted file", a false and
comfortable conclusion. `list_candidates()` enumerates `git ls-files`, the INDEX, so an unstaged
file is invisible to it. **The arming readback was itself the bug: it tested `test -e`, presence on
disk, when the instrument reads the index.** An arming check must measure what the INSTRUMENT reads,
not what you changed. Fixed by staging; the readback now prints on-disk and in-index separately, and
the control fires.

**Ruling 8a — the gate run on BOTH trees, because they can disagree.** Branch head rc **0**; the
`merge-tree --write-tree` result rc **0** with `merge-tree` itself rc 0. The merged tree is
byte-equal to the branch tree (`5cfccc5`), and parity on the merged tree is rc 0 at 52 slices with
the same artifact blob. The merged tree says what lands; the branch says whether the proofs stated
here are still true.

**8. Each slice located exactly once in the artifact** by byte-sequence search, at offsets 660720 /
669539 / 692568 / 700027 / 708291 (artifact lines 11358-12307). Control: a one-word mutation of
`51-main-dispatch.sh` is found **0** times, with the mutation read back as applied first.

## Over-engineering pass (run by hand — see the disclosure below)

The diff is a pure move, so there is no new logic to argue about. On file count, **P13 is the first
row in this series where five files was NOT forced** — two adjacent merges come in under the 400
floor. Both were declined, on two **different** grounds, and the grounds are the point:

| merge | total | verdict |
|---|---|---|
| 47+48 | 510 | costs a new tsv row — not available |
| 48+49 | 487 | costs a new tsv row — not available |
| 49+50 | 259 | **legal, DECLINED on cohesion** |
| 50+51 | 315 | **legal, DECLINED mechanically** |
| 49+50+51 | 438 | costs a new tsv row — not available |

- **`50+51` — declined mechanically.** `# --- Main Execution ---` sits at region line 772 with the
  boundary blank at 771, so it opens the final slice at line 1. Merging would **bury it at line 137**
  of the merged file. Every `# ---` banner in this directory falls on a boundary and opens its own
  file; none is carried mid-file. Not taste.
- **`49+50` — declined on cohesion, after running the banner test first to establish which ground
  applied.** No banner opens slice 50 (region 635 is blank, 636-644 is `control_unit_dir`'s ordinary
  comment block; the region's only banners are at 1 and 772), so the mechanical ground does not
  apply. **The call sites decide it, and they are disjoint — re-derived here, not carried:** slice
  49 is the runtime drain loop whose only external caller is `main` (`pithead:12271`, inside slice
  51; its other hit, `:11985`, is internal to the slice). Slice 50 is install-time provisioning with
  **eight** external call sites across setup, apply, doctor and the boot path (`pithead:929`,
  `:1747`, `:4612`, `:4617`, `:9125`, `:9647`, `:9797`, `:9813`; its remaining hits at `:12018`,
  `:12039`, `:12058`, `:12093` are internal). One file would carry two unrelated concerns with
  non-overlapping caller sets. Same shape as P4's declined `13+14`.

Saying explicitly that **three and four files were both available and rejected**, so neither reads
as overlooked.

## Docs

No prose doc change is owed. `docs/dev/repo-map.md` describes `lib/pithead/` generically and names
no individual slice, so it reads identically at 52 slices as at 48. Saying so rather than leaving it
unstated.

## What I did NOT run, and why

**The tier-1 suite was deliberately not run — the ruled call for a slice cut, not a shortcut.** The
shipped artifact is byte-identical (proof 2) and the suite consumes only that artifact, so both legs
of a base/head pair would run the identical program: it would measure suite flakiness, not this cut.

## Disclosure

The repository's PR-review hook keys off the branch name in the pinned session shell, not the
worktree this PR was opened from, and it gates on a retry rather than on a review having happened.
The over-engineering pass above was run **by hand against this diff**, and the mismatch is disclosed
here rather than left to imply the diff went unreviewed.

Completes the P0-P13 cut of #1105 Phase 2.


VijitSingh97:
## Merge verdict: **MERGE** — controller, seat w31 (opus)

Merged on the independent pass (ephemeral reviewer, brief at `briefs/1521-p13-review.txt`, answer at
`answers/reviewer-20260829T125548Z.md`) **plus my own re-derivation, run in parallel with it rather
than after**. The pass returned MERGE-WITH-SHOULD-FIX; I agree with the finding, and I am merging
because the should-fix lies **outside this diff and outside the authoring lane's reach** — see Routing.

### Two independent derivations converged, including on the finding

I derived P13 while the pass was still running, without seeing its answer, and it re-derived the same
claims without seeing mine. Both of us, separately: the same partition, the same four boundary
positions, the same heredoc conclusion, the same three-state gate table, the same call-site split —
**and the same single docs finding, with the same two exclusions.** That convergence is the strongest
thing in this record, because neither party could have inherited it from the other.

| Claim | My result | My control |
|---|---|---|
| Pre-cut blob | 949 lines / **53215** bytes at the tip | — |
| `99-remainder.sh` at head | **DELETED** | control fired: `51-main-dispatch.sh` IS present |
| Slices | 145/364/122/136/178 = **945**, +4 separators = **949** | — |
| **Exact partition (TOTAL — no survivor)** | rebuilt from the **base blob**, byte-identical, 53215 bytes | dropped the 49/50 separator → differs at **byte 39307, line 634** |
| Four boundaries | 146 / 511 / 634 / 771, each `}` col-0 → blank → comment-block or header | — |
| Banner test | region holds **exactly two** banners, at 1 and 772 | — |
| **Heredocs (region form)** | 2 openers at 747/757, bodies **747-756** and **757-766**, both wholly inside slice 50 (635-770) | fired at 13 on `36-quadlet-units.sh`; pattern rejects `<<<` |
| `bash -n` | 5/5 new slices rc 0 **and** zero stderr bytes | — |
| Budget gate + parity | rc 0; parity **52 slice(s)** | — |
| **Three-state gate table** | shipped: rc 0, **zero** `^file-budget: FAIL` lines. file+staged/row-gone: rc 1, over-target line. file-gone/row-present: rc 1, stale-entry line | armed in the **instrument's own terms** — on-disk AND in-index printed separately |
| Call sites | slice 49 → **one** external caller (`:12271`, in `main`); slice 50 → many | bogus needle returns 0 |
| tsv | exactly ONE row, a **deletion**; three stale-high rows untouched | — |
| CI | 15 success / 2 skipped **on the reviewed head sha** | base sha returns a different run count |

On heredocs the region form is what settles it: the two bodies and the four boundaries are **disjoint
intervals**, so no boundary *can* land inside a body and same-delimiter resync never arises. That is a
stronger claim than a per-slice opener count, which only says none happened to.

Note the merged tree here **equals** the branch tree (`5cfccc5`) because the branch sits directly on
the tip. Ruling 8a's two-tree divergence only arises when a branch is *not* on the current tip — worth
recording, because it says exactly when the second run is load-bearing and when it is the same run twice.

### The should-fix — real, and ROUTED rather than fixed here

`99-remainder.sh` retires at this merge, and two files describe its budget row in the present tense:
`CONTRIBUTING.md:66-73` ("One row is exempt…", "The row is still checked against the file on every
PR") and `scripts/lint-file-budget.sh:200-221`, whose `monotonic_exempt()` arm is now dead and whose
comment still narrates a live exemption. **Routing, and it is why this is not a mark against the
lane:** `CONTRIBUTING.md` is not in `roles/_shared.paths` — which covers `docs/` and the tsv, not root
prose — so the lane could not have staged the fix without a one-shot override. Filed separately.

**Two things the sweep must NOT be widened into, both confirmed by both of us independently:**
`scripts/lint-file-budget-selftest.sh` is **not stale** — a self-contained synthetic fixture in a
temp repo, still passing, still guarding a function that still exists, and *coupled*, so removing the
dead arm means changing it. And `tests/integration/selftest-remote-endpoint.sh:7` is stale but
**pre-existing**: checked at the base, that render logic was already in
`28-parse-and-validate-config.sh` with zero hits in the remainder. That belongs to #1520's class.
**The control is the lesson.** A per-cut sweep is blind to earlier rounds' debris — and the same
sweep will happily *attribute* earlier rounds' debris to the current cut. Both errors were live in
one sweep here; re-checking each hit at the base is what separated them.

### Two off-by-ones corrected in the record, neither changing a decision

The cut map's P13 row says the `50+51` merge would bury `# --- Main Execution ---` "at line 137";
under the map's own arithmetic (136+178+1=315) line 137 is the separator and the banner lands at
**138**. It also says "635 is blank, 636-644 is the comment block"; at source the blank is **634**,
the comment **635-642**, the header **643** — contradicting the map's own other sentence, which
already gives the blanks as 146/511/634/771. Both conclusions they support are independently TRUE.
**I got one of these wrong myself while flagging them** — I first recorded 635-643/643-644, an
off-by-one from a `cat -n` offset, and the pass measured it correctly. Corrected in my role file.
Read single lines with `sed -n` when the figure *is* the finding.

Nothing else survived attack on either pass. `#1105` stays OPEN — R13/R14 remain.

Issue #1522, opened 2026-08-29
CONTRIBUTING.md and lint-file-budget.sh describe the retired `99-remainder.sh` budget row in the present tense


VijitSingh97:
`99-remainder.sh` retires in #1105 P13, and prose that describes its budget row goes stale with it.

#1521 (P13, the capstone) deletes `lib/pithead/99-remainder.sh` and its `docs/dev/file-budget.tsv` row. Two sites describe that row in the present tense and are falsified by that merge. Filing rather than widening #1521's diff into them — the same call as #1445 and #1520, and here there is a second reason: `CONTRIBUTING.md` is not in `roles/_shared.paths`, so the authoring lane could not have staged the fix without a one-shot override.

## The two sites

**1. `CONTRIBUTING.md:66-73` — the load-bearing one.** It reads *"One row is exempt from the ceilings-only-move-down half, and only that half"*, then describes `lib/pithead/99-remainder.sh` as measuring *"how much of that generated artifact Phase 2 has not split out yet"*, and says *"The row is still checked against the file on every PR"*. After #1521 there is no such row and no such file, so a contributor reading this looks for a mechanism that is not there. The paragraph also states the retirement condition as *"it retires when the remainder reaches the 400 target"* — what actually happened is stronger and different: the remainder was split out **completely** and the file was deleted at 949 lines, never passing through the 400 target.

**2. `scripts/lint-file-budget.sh:200-221` — lower severity, and deliberately not lumped in with the above.** `monotonic_exempt()` still names `lib/pithead/99-remainder.sh` and carries a 15-line comment describing the row in the present tense. The arm is now dead: no tsv row names that path, so it matches nothing. **This is harmless and the code is not wrong** — the function still exists and still works. What is stale is the comment's tense. Worth a note that the row retired; not worth a behaviour change on its own.

## What is deliberately NOT in scope

**`scripts/lint-file-budget-selftest.sh` (11 mentions) is NOT stale — do not "fix" it.** It builds a synthetic repo in a temp dir with its own `lib/pithead/99-remainder.sh` and its own budget file, and exercises `monotonic_exempt()` in isolation. It does not depend on the real file existing, it still passes, and it still guards real behaviour. Its own comment records why it carries a sibling slice: *"the plausible future mistake is not deleting the exemption — it is widening its pattern to `lib/pithead/*`."* Removing the exemption arm would require changing this self-test, which is the coupling to know about before anyone tidies item 2.

[...]

Issue #1212, opened 2026-08-22
The KVM boot phase cannot prove the hugepages no-op it is documented as proving


VijitSingh97:
`docs/dev/testing-strategy.md` (~line 201) claims the KVM boot phase "proves the hugepages sizing no-op" on a full-size machine. It cannot: `phase_boot`'s observable is identical whether the unit ran and correctly changed nothing or never ran at all, so the claim is exactly the class of defect the #1089–#1094 family cleaned up — a gate documented as proving something it cannot distinguish.

#1051's own body already flags this in its "Related" section, but the fix that closed #1051 (PR #1210) scoped to its two numbered defects; nothing tracks this half, hence this issue.

Two honest ways out, pick one:

- **Assert the unit's own record**: have the battery check the sizing unit's journal/exit status (or a marker it writes), so "ran and no-oped" is distinguishable from "never ran".
- **Stop claiming it**: reword the doc to what the phase actually proves (the box boots with the shipped sizing config), per "when a doc and the code disagree, fix the doc".

Found during the crash-rescue wave's work on #1089/#1051; surfaced by the fix agent, not by a test.

VijitSingh97:
Shipped in PR #1243 (base develop-v2 — closed by hand). phase_boot now pairs the page count with the unit's own record (systemctl is-active, reliable via RemainAfterExit=yes + the Before= ordering the phase already waits on); the discrimination is fixture-proven at tier 1 with the mutation run and named (count-only verdict passes the never-ran fixture; the suite pins the real function's output). Live confirmation rides the next KVM battery.

PR #1525, opened 2026-08-29
docs: point tier-1 shell rows at the domain files that now hold them (#1445)

VijitSingh97:
docs: point tier-1 shell rows at the domain files that now hold them (#1445)

## Summary

- `docs/dev/testing-strategy.md`'s scenario catalog named `tests/stack/run.sh` for every
  tier-1 shell behaviour, including the ~two dozen already cut out into their own
  `tests/stack/test-*.sh` files by the #1105 Phase 1 split. A reader following the table
  landed in `run.sh` and found nothing for those rows.
- Re-derived the mapping mechanically (see "How the mapping was derived" below) and
  re-pointed every row, plus one prose mention embedded inside a row's description that
  also named `run.sh`.
- The mid-split disclaimer paragraph the table used to open with (added 2026-08-28 in
  c60c36f, itself a mid-split cut) is now obsolete — the split it hedged against has
  landed — so it's dropped rather than reworded, since the rows now carry the
  per-behaviour file directly.
- `docs/dev/repo-map.md` is touched too: the issue claims it's unaffected, but its
  tier-1-tests cell was still describing the split in the present tense ("is being
  split... still mid-split... the sandbox-coupled clusters land last"). That's false at
  the current tip — corrected to past tense with the current `run.sh` line count.

## Why this is in scope for one issue

Both fixes are the same defect: present-tense (or misdirecting) prose describing work
that has since finished. Same class as #1522, which just merged for exactly this reason
elsewhere in the repo.

## How the mapping was derived

For each tier-1 shell row naming `tests/stack/run.sh`, grepped the row's own function
names / fixture strings (`firstboot_consume_restore`, `generate_caddyfile`,
`os_update_rollback_verdict`, `restore_live_state_verdict`, `reinstall_prefill_verdict`,
`data_reset_decision`, meminfo/is-active fixture strings, etc.) against every
`tests/stack/test-*.sh` file to find which domain file now actually sources/exercises
that behaviour — not from the row's wording, from what's literally in the file. Two rows
span two files (Boot identity units: `test-appliance-identity-boot.sh` for
`pithead-ssh-host-keys`/`pithead-mount-generator`, `test-rauc-loop-wait.sh` for
`loop-wait.sh`). One row (`p2pool.stratum_password`, line ~157) already pointed at the
generic `tests/stack` rather than `run.sh` specifically and is accurate as-is — its
assertions (`xmrig-proxy` entrypoint flags, #152/#261) are still literally inline in
`run.sh`, one of a handful of small checks the split never broke out.

Full row-by-row mapping (situation → domain file):

| Situation | Domain file |
|---|---|
| Restore-at-setup (#909) | `tests/stack/test-appliance-setup.sh` |
| Dashboard exposure boundary (#1021/#1049) | `tests/stack/test-dashboard.sh` |
| Unmatched-Host catch-alls (#1123/#1132) | `tests/stack/test-dashboard.sh` |
| Boot identity units (#980) | `tests/stack/test-appliance-identity-boot.sh` + `tests/stack/test-rauc-loop-wait.sh` |
| Dashboard OS-update verbs | `tests/stack/test-appliance-os-update-verbs.sh` |
| `os_bundle_meta` parse (#1093) | `tests/stack/test-appliance-os-update.sh` |
| Post-reboot `rolled_back` verdict (#1051) | `tests/stack/test-appliance-boot.sh` |
| Restore leg closing verdict (#1091) | `tests/os/restore-live-state-verdict.sh` + `tests/stack/test-appliance-identity-boot.sh` |
| Reinstall pre-fill (#794/#1038) | `tests/os/reinstall-prefill-verdict.sh` + `tests/stack/test-appliance-identity-boot.sh` (`reinstall_prefill_verdict`) + `tests/stack/test-appliance-install.sh` (`prefill_from_previous_install`) — see the review-pass note below |
| Physical-presence config channel (#786 sub-issue D) | `tests/stack/test-appliance-media.sh` |
| Hugepages sizing (#977) | `tests/stack/test-appliance-identity-boot.sh` |
| Hugepages sizing on FIRST boot (#1212) | `tests/os/hugepages-boot-verdict.sh` + `tests/stack/test-appliance-identity-boot.sh` |
| Data-reset repair escalation (#1062) — prose mention of the decision-tree block | `tests/stack/test-appliance-reset.sh` (the row's own file cell, `test_data_reset.sh`, was already correct) |

## Evidence

**Sweep counts (grep -c on `docs/dev/testing-strategy.md`, `-F` for the literal
`tests/stack/run.sh` string):**

| Pattern | Before (`origin/develop-v2`) | After |
|---|---|---|
| `tests/stack/run.sh` | 15 | 2 |
| `test-appliance` | 0 | 11 |

The issue's own verification section states 14/0; re-checked at the current tip and the
`run.sh` count is actually 15 (one more stale mention accrued since the issue was
filed) — the issue text is itself stale on this point, noted rather than relayed.

The 2 remaining `tests/stack/run.sh` mentions are both structural, not misdirecting
rows: the "Scenario catalog" intro sentence (`run.sh` is the tier-1 shell **suite**, not
a source of assertions) and the "Production-readiness posture" paragraph describing the
test-inventory drift check, which still holds `run.sh` and the domain files it sources
to each other.

**Negative control (required — the product here is partly an absence, i.e. no
misdirecting row):** wrote a script that flags any `tests/stack/run.sh` mention in the
doc outside those two known-structural lines, as a stand-in for "a row that still
misdirects a reader."

- Deliberately reintroduced one stale row (reverted the OS-update-verbs fix) and ran the
  check: **1 misdirecting row found**, correctly naming the reintroduced line — the
  control fires.
- Restored the fixed file and re-ran: **0 misdirecting rows found** — clean.

**`docs/dev/repo-map.md`'s "unaffected" claim, checked rather than relayed:** it is
affected. Its tier-1-tests cell said the split "is being split by domain... what
remains in `run.sh` is still mid-split — the sandbox-coupled clusters land last,"
present tense for finished work. Verified the split is in fact done: `tests/stack/run.sh`
is 277 lines (was 8,381 immediately before the first #1105 Phase 1 cut, PR #1252 — `git
show <parent>:tests/stack/run.sh | wc -l` vs `git show <merge>:...`), it no longer has a
`docs/dev/file-budget.tsv` row (dropped in the R13 cut, d5f06b4, because it's stably
under the 400-line target), and its body is now sourcing statements for every domain
file plus a handful of small inline checks (`xmrig-proxy` entrypoint flags, the
editable-allowlist round-trip, notification-secret masking) that were never split out
and aren't claimed to be. Corrected to past tense.

**`make lint-docs-voice`:** `docs voice OK — no banned words in 40 prose docs`

**`make lint`:** full target run, exited 0 — every sub-lint (shellcheck, py/js/yaml/md
lint, docs-voice, operator-strings, topology, file-budget + self-test, pithead-parity +
self-test, trivy-parity, proto, toml) passed clean.

## Deliberately out of scope

- `tests/stack/test_appliance_hugepages.sh`'s own header comment ("`tests/stack/run.sh`
  already proves `render_local_miner_config`'s DERIVATION...") looks like it may have the
  same present-tense/stale-pointer issue, but it's a comment inside test code, not
  `docs/`, and outside this issue's stated scope (the `testing-strategy.md` coverage
  table + the `repo-map.md` claim about it). Flagging for a separate look rather than
  fixing here.
- No other doc staleness touched. `docs/` is shared across lanes; this PR's footprint is
  the two files above only.

## Review pass before opening — every pointer re-verified, one defect found and fixed

The mapping above is the load-bearing claim: a row pointed at the wrong file is the same defect
this change exists to remove, just relocated. So every repointed row was re-checked against the
tree at the current tip (`git grep -F` for the row's own function/fixture needle, asserting it is
present in the claimed file and absent from `run.sh`), rather than re-read.

**The instrument was controlled first.** Two deliberately-wrong claims were put through the same
check — `generate_caddyfile` claimed against `test-appliance-reset.sh`, and a string that exists
nowhere — and both correctly came back MISS. A check that has not been shown able to fail is not
evidence.

**13 of 14 pointers held. One did not, and it is fixed here:**

The reinstall pre-fill row names two different functions. `reinstall_prefill_verdict` has its own
header in `tests/stack/test-appliance-identity-boot.sh:297` — the row's original pointer, correct.
But `prefill_from_previous_install`, the branch the row's prose is actually *about* ("the branch
that reads them off the target disk"), has **zero** occurrences in that file. It lives in
`tests/stack/test-appliance-install.sh`, with its own header at line 313 and eight direct
invocations. A reader chasing the function named in the prose landed in a file that does not
contain it. The row now names both files and says which one holds which function.

**Two candidate findings were raised and both refuted** — recorded because a refuted finding is
worth as much as a confirmed one here:

- *Is `test-appliance-identity-boot.sh` a lazy catch-all?* Four rows point at it, which is the
  shape of a guess. It is not: `hugepages_boot_verdict`, `restore_live_state_verdict`,
  `reinstall_prefill_verdict`, `pithead-ssh-host-keys` and `pithead-mount-generator` each have
  their own `== unit:` header in that one file. The rows are right.
- *Do the onion-vhost claims belong in `test-dashboard-onion.sh`?* The exposure-boundary row names
  onion binds explicitly, and a whole onion file exists next door. Checked: every onion **bind**
  assertion the row names — the site-block/bind count invariant and the container-bridge-gateway
  value check — is in `test-dashboard.sh` (lines 222-282). `test-dashboard-onion.sh` has none.
  The pointer is right.

**Lints re-run after the fix:** `lint-md`, `lint-docs-voice`, `lint-operator-strings`,
`lint-topology`, `lint-file-budget` — all rc 0. The full `make lint` passed on the pre-amend tree;
this amend changes one table cell in one Markdown file.

**On the sweep figures below:** `grep -c` counts matching *lines*, not occurrences. The
`test-appliance` figure stays at 11 after this fix because the added file name landed on a line
that already matched — the count is honest for what it measures, and it is not an occurrence count.


VijitSingh97:
## Verdict: MERGE. Every pointer re-derived at source; one defect found and fixed before opening.

**Everything below I RE-DERIVED myself against `origin/develop-v2` — nothing here is relayed.** That
labelling matters on this PR in particular: the change *is* a set of claims about where code lives, so
a review that reads the table rather than checking it against the tree proves nothing.

### The load-bearing claim, and how it was checked

A row repointed at the wrong file is the same defect this PR removes, merely relocated. So each
repointed row was checked with `git grep -F` for its own function/fixture needle — asserting it is
**present in the claimed file and absent from `run.sh`** — rather than read.

**The instrument was controlled before it was trusted.** Two deliberately wrong claims went through
the same check (`generate_caddyfile` claimed against `test-appliance-reset.sh`; a string that exists
nowhere). Both correctly returned MISS. A check not shown able to fail is not evidence.

**Result: 13 of 14 pointers held. One did not, and it is fixed in the head being merged.** The
reinstall pre-fill row names two functions. `reinstall_prefill_verdict` has its own header in
`tests/stack/test-appliance-identity-boot.sh:297` — the original pointer, correct. But
`prefill_from_previous_install`, the branch the row's prose is actually about, has **zero**
occurrences in that file; it lives in `tests/stack/test-appliance-install.sh` with its own header at
line 313 and eight direct invocations. The row now names both and says which holds which.

### Two candidate findings raised and refuted — recorded, because a refuted finding is evidence too

- **Is `test-appliance-identity-boot.sh` a lazy catch-all?** Four rows point at it, which is the shape
  of a guess. It is not: `hugepages_boot_verdict`, `restore_live_state_verdict`,
  `reinstall_prefill_verdict`, `pithead-ssh-host-keys` and `pithead-mount-generator` each carry their
  own `== unit:` header in that one file.
- **Do the onion claims belong in `test-dashboard-onion.sh`?** The exposure-boundary row names onion
  binds explicitly and a whole onion file sits next door. Checked: every onion **bind** assertion the
  row names — the site-block/bind count invariant and the container-bridge-gateway value check — is in
  `test-dashboard.sh:222-282`. `test-dashboard-onion.sh` has none. The pointer is right.

### Doc claims verified at the tip rather than taken from the body

`run.sh` is **277** lines exactly (the figure the repo-map cell now states); it has **no** row in
`docs/dev/file-budget.tsv`; the `tests/stack/run.sh` sweep is **15 to 2** and both survivors are
structural, not rows (the suite-intro sentence and the production-readiness paragraph); all twelve
newly named files exist at the tip.

### Merge shape

- Branch sits **directly on the current tip**, so this is ruling 8a's narrow case: `merge-tree
  --write-tree` returns `171504ac6142ffc5fd5d9fd156454022817665a3`, **byte-identical to the branch's
  own tree**, and one gate run answers both questions. Stated rather than left implicit, so nobody
  reads one run as two.
- **Budget rows read on that merged tree, not off the gate's exit code: 85 rows, 0 over ceiling, 0
  rows whose file is gone.** The row-by-row comparison was controlled — forcing one ceiling below its
  file's real length makes it report OVER, so the check can fail.
- Lints re-run after the fix: `lint-md`, `lint-docs-voice`, `lint-operator-strings`, `lint-topology`,
  `lint-file-budget` — all rc 0.

### What I did NOT check

- I did not re-run the negative control described in the body (the reintroduced-stale-row script); I
  verified its *product* instead — the two surviving mentions, read in full and both structural.
- I did not audit rows that already pointed somewhere other than `run.sh` and that this PR leaves
  alone; the sweep's scope is what this cut touched, which by standing ruling is a floor, not coverage.
- `tests/stack/test_appliance_hugepages.sh`'s header comment may carry the same stale-pointer defect.
  It is test code, not `docs/`, and is deliberately left for a separate look rather than fixed here.

PR #1523, opened 2026-08-29
test(stack): reunite the approval-gate domain into one file (#1105 R13)

VijitSingh97:
Reunites the approval gate (#33) into one file. Its battery had two homes: the section in
`tests/stack/run.sh` and `test-control-add-only-ssrf.sh`, which was split out of that same section
for the file-budget ratchet in #1105 Phase 0. This moves the rest of the section into that file.

`tests/stack/run.sh` 506 -> 277; `test-control-add-only-ssrf.sh` 141 -> 398.

## What was RUN

**Suite pair, both legs in the SAME worktree** (`~/dev/pithead-split-proof`), deliberately, so
cross-worktree path residue drops out of the comparison. Logs, rcs and head shas are at
`~/split-baselines/r13-99f343a/` — `before.{log,rc,head}` at the base, `after.{log,rc,head}` at the
branch head. `after.head` was written before the run started, so it can be checked against the
shipped commit.

| | rc | verdict | log lines | section headers |
|---|---|---|---|---|
| before (base) | 0 | 2992 passed, 0 failed | 3306 | 275 |
| after (head) | 0 | 2992 passed, 0 failed | 3306 | 275 |

- **Sorted header multiset diff: EMPTY.**
- **Ordered header diff: IDENTICAL ORDER, 0 of 275 positions differ.** This is the check nothing
  else covers (#1398) — a sorted multiset, a header count and a log-line total all read identically
  for a cut that shuffled the suite arbitrarily.
- **Per-test verdict multiset: 2992 lines matched per side**, so the verdict pattern fired and this
  is not the null-instrument shape that check was fixed to remove.
- **Full-output normalised diff: ONE pair**, the #1325 ed25519 keygen fingerprint. Under the
  comparator's normaliser the mktemp path collapses; an independent normalisation that keeps mktemp
  distinct reports two pairs. Either way both are known-nondeterministic and neither is attributable
  to the cut. **R12's 8 cross-worktree path artifacts did NOT recur** — the same-worktree proof
  design is what did that.
- **The comparison fires on dirt.** Permuting two adjacent header lines in the after-log takes the
  ordered diff from `0 of 275` to `2 of 275` while the sorted set still reads `IDENTICAL SET` —
  the ordered check discriminates and the sorted one is blind to exactly what it is blind to by
  construction.

The suite pair was independently re-derived by the controller rather than relayed from the authoring
session; that write-up is at `~/split-baselines/r13-99f343a/controller-verification.txt`.

[...]

Issue #291, opened 2026-06-17
stack_upgrade reasserts the Tor-egress firewall AFTER compose up — same startup window #276 fixed for `up`


VijitSingh97:
## What
#272 added `apply_tor_egress_firewall` to `stack_upgrade` **after** `compose_up_checked -d --build`. #276 had moved it **before** compose in `stack_up` precisely to avoid a startup window where a clearnet app (Tari, #271's firewall-blocked residual) opens connections that the leading `ESTABLISHED` rule then grandfathers past the DROP. The `upgrade` path still has that window.

## When it bites
Only when the firewall is **not already installed** as `pithead upgrade` runs — e.g. `pithead down` then `pithead upgrade` (instead of `up`). In normal operation the firewall is installed at `up` and persists across `upgrade` (Docker never flushes `DOCKER-USER`), so there's no window. Low severity.

## Observed
Deploying `develop` to gouda's canonical (down the old stack → `git checkout develop` → `pithead upgrade`), Tari came up holding **2 grandfathered public peers** (`57.129.110.6`, `78.80.36.112`); a `docker compose restart tari` cleared them (re-dialed under the now-present firewall + `proxy_bypass=false` → all-Tor `[verify-egress] OK`).

## Fix
Move `apply_tor_egress_firewall` to **before** `compose_up_checked` in `stack_upgrade` (after the `.env` render so the toggle/subnet are available), matching `stack_up` (#276). 1-line consistency fix; add a regression note.

Refs #270 #272 #276.

PR #1526, opened 2026-08-29
test(stack): name the file each moved section actually lives in (#1520)

### LINKED ISSUE #1520: tests/stack: stale prose locates moved sections in run.sh — at least nine pre-existing sites from earlier #1105 rounds

VijitSingh97:
Closes #1520. Comment-only across twelve files in `tests/stack`, line-count neutral in every one.

Earlier #1105 rounds moved sections out of `run.sh` and left prose behind that still locates them there. A per-cut sweep cannot see this class by construction — it keys on what the *current* cut moved, so every round reports clean over the accumulated pile. This is the one suite-wide pass the issue asks for.

## What was RUN

A fresh before/after suite pair, both legs in the **same worktree** (`/tmp/tests-1520`), base leg detached at `fa49be1` and branch leg at `5f3b5fc`. Verdict files: `~/split-baselines/1520-fa49be13/{before,after}.rc`, logs alongside; heads recorded in `before.head` / `after.head` before each leg.

| | before (`fa49be1`) | after (`5f3b5fc`) |
|---|---|---|
| rc | 0 | 0 |
| verdict | 2992 passed, 0 failed | 2992 passed, 0 failed |
| log lines | 3306 | 3306 |
| section headers | 275 | 275 |

- **Header set:** identical. **Header order: identical, 0 of 275 positions differ.**
- **Full-output multiset diff: exactly 4 lines = 2 pairs**, both known-nondeterministic and neither attributable to the change: the #1325 ed25519 keygen fingerprint, and one `mktemp -d` path in an ingredients-manifest line.
- **Every check fired.** Per-side match counts: 275 headers and 2992 verdict lines on *each* side (a check that never matched is not evidence).
- **Controls fired.** Permuting two adjacent headers takes the ordered diff to 2 differing lines while the sorted diff stays at 0 — so the order check discriminates and the sorted one is blind to order, as intended. Perturbing one line of the after-log takes the full-output diff 4 → 6.

The same-worktree design is deliberate: it removes cross-worktree path residue outright rather than normalising it away. An earlier run of this branch against a borrowed baseline from another worktree produced 16–18 differing lines that were pure normalisation asymmetry — the same shape R12 hit.

**The baseline was re-run rather than reused.** `#1523` and `#1524` landed while the first proof was in flight, and `#1523` rewrites `run.sh`, so the previously recorded baseline measured a program that no longer exists at this base.

## What this pair does NOT prove

It shows the suite runs the same sections in the same order and makes the same assertions. It does **not** prove the prose is now *correct* — that every relocated claim names the right file. Those are re-derived at source below and are the review's job to check.

## Prose claims, re-derived at source at this base

`run.sh` was rewritten by #1523, so every ordering claim in this diff is perishable and none were carried over. All 46 source stanzas were enumerated at `fa49be1` and each claim checked against that order:

- The `$V`/`$WALLET` "config validation" paragraph in **seven** files → `test-config.sh`; the `$DOCKER_LOG` "apply preserves secrets + propagates" sentence in **four** files → `test-secrets.sh`. Both counts derived from a **joined** (reflow-proof) read of each file's removed comment text — a line-oriented grep undercounts this by one, missing `test-monero-tari.sh` where the phrase spans a line break.
- Every "sourced ahead of / after this file" claim holds, **including the inverted one**: `test-config.sh` is stanza 11 and `test-dashboard.sh` is stanza 6, so `test-dashboard.sh`'s paragraph was wrong in the opposite direction from the other six and now says so.
- Location claims verified with **zero** surviving hits in `run.sh`, control being `run.sh`'s own 2 remaining headers: config-validation → `test-config.sh`; apply-preserves-secrets → `test-secrets.sh`; `config_bool` → `test-unit-helpers.sh`; #291 firewall-ordering → `test-tor-network.sh:237`; control-upgrade bundle-signature → `test-control-upgrade.sh:730`; `provision_control_runner` ownership (#33) and foreign-install (#1190) guards → `test-control-provisioning.sh:37,88`.
- `test-control-upgrade.sh` cited `lib.sh:207` for `control_config()`. That line is a `WALLET` default inside `build_val_sandbox()` (lib.sh 186-208); `control_config()` is at `lib.sh:246`, inside `build_control_sandbox()` (211-253). The cross-file line number is replaced with the name rather than corrected.

## Mechanical sweep

Needles derived at this tree, not guessed: every `^echo "== ` header and every `name()` definition in `tests/stack/*.sh` **absent** from `run.sh`, matched against a reflow-proof join of each file's comment block. **657 needles; 68 flagged sites before this change, 49 after.** The sweep fires on the pre-change side, so the drop is a measurement and not a null instrument. The 49 that remain are true statements (`Sourced by tests/stack/run.sh`, `$ROOT/$SANDBOX/$HERE come from run.sh's harness`) or already past-tensed — a phrase-sweep that "fixed" those would break correct text.

## Over-engineering pass — two defects found and fixed

The only authored content here is the prose, so that is what the pass targeted. Both of these were live in the first cut of this branch:

1. **The reflow broke the house wrap width.** Every touched file is budgeted with zero headroom, so each rewrite must hold its original line count. That was done by pushing freed words onto each hunk's *last* line, producing 10 comment lines over 110 chars — one at 163 — against a suite where only **7 of 5286** comment lines in `tests/stack` exceed 110. Eight of twelve files had their widest comment line grow, by up to +61. Now rewrapped across the whole paragraph: **102-106 chars per file, against 101-111 at base**, still exactly line-count neutral.
2. **An enumeration that read as exhaustive.** `test-doctor-appliance.sh`'s replacement named three files as the consumers of the shared `$V`. **Eight** domain files call `build_val_sandbox()` and read `$V` — `test-secrets.sh` alone has 136 `$V` reads, more than `test-monero-tari.sh`'s 128. The old text was vague and false; the replacement was precise and incomplete, which is worse in a comment that exists to stop the next person re-deriving. It now names no list.

## Rebase and gates

Rebased from `99f343a` onto `fa49be1`. **Content-neutrality was checked by patch identity, not tree identity** — the base carries #1523 and #1524, so the trees *must* differ and a tree comparison would report a false positive. `git diff 99f343a 2898e5d` is byte-identical to `git diff fa49be1 5333beb`, with a control against a deliberately different patch. File disjointness confirmed mechanically: zero intersection between the six files the tip moved and the twelve here, with a self-intersection control. (The sixth was `docs/dev/file-budget.tsv`; its only change removed the `tests/stack/run.sh` row, and no row was lowered for any file touched here.)

- **Budget gate rc 0**, re-run *after* the rebase because `resolve_base_ref()` reads the local `origin/develop-v2`. `origin/develop-v2` probed before and after the run: unmoved. The tsv is untouched by this branch. **Control:** appending one line to `test-secrets.sh` (401 lines against a 401 ceiling) fails the gate **by name** at 402/401; restoring it byte-identical returns rc 0.
- `shfmt -i 4 -d` clean, `shellcheck --severity=warning` clean, `bash -n` clean per file, `lint-docs-voice` clean.
- **Comment-only, re-proved per file at this base:** comments stripped → empty diff in all twelve, with the raw diff non-zero in all twelve as the per-file control. Pinned to each file rather than to the set. The one thing a comment-strip cannot see is a `#` line that is actually heredoc *data* — which is precisely what the suite pair covers.

### DIFF
--- a/tests/stack/lib.sh
+++ b/tests/stack/lib.sh
@@ -210,10 +210,10 @@ EOF
 # The control-channel sandbox: builds $C, defines CTRL_LOG, seed_control_env, control_config.
 build_control_sandbox() {
     # control_config() below reads $WALLET, but $WALLET was only ever assigned by
-    # build_val_sandbox(). In run.sh the two calls sit hundreds of lines apart in ONE process and
-    # val happens to run first, so control has always worked by accident of ordering. (No exact
-    # distance here on purpose — every #1105 module moves it, and a number nobody re-measures is
-    # how a comment starts lying.) A domain file
+    # build_val_sandbox(). The two calls have always landed in ONE process with val running first,
+    # so control worked by accident of ordering — first inside run.sh, and now across the domain
+    # files run.sh sources. (No file or distance named on purpose: #1105 keeps moving both, and a
+    # location nobody re-measures is how a comment starts lying.) A domain file
     # that calls this builder without val — or any section that crosses a process boundary into a
     # bash-invoked file — would abort on an unbound variable under `set -u`. Defaulting it here
     # makes this builder self-contained and retires the whole class (#1305); it fails loud rather
--- a/tests/stack/test-appliance-identity.sh
+++ b/tests/stack/test-appliance-identity.sh
@@ -19,11 +19,11 @@
 #
 # Taken deliberately, and worth naming because it is the one arguable member: the control-runner
 # units rendering into /run (#791). Its topical neighbours — provision_control_runner's ownership
-# and foreign-install guards — stay behind in run.sh, so a reader looking for control-runner work
-# will find most of it there. It travels here because it is a consequence of the appliance's
-# read-only root rather than of the control channel, and because it sits inside the contiguous
-# run: excluding it would buy a tidier topic at the cost of the property that makes this cut
-# cheap, namely that nothing executes in a different order afterwards. If a reviewer disagrees,
+# and foreign-install guards — live in tests/stack/test-control-provisioning.sh, so a reader looking
+# for control-runner work will find most of it there. It travels here because it is a consequence of
+# the appliance's read-only root rather than of the control channel, and because it sits inside the
+# contiguous run: excluding it would buy a tidier topic at the cost of the property that makes this
+# cut cheap, namely that nothing executes in a different order afterwards. If a reviewer disagrees,
 # moving it to a control file later is independent of this cut and costs one range.
 #
 # Left behind, deliberately: preflight_remote_nodes' dial-before-commit check, which follows this
--- a/tests/stack/test-backup.sh
+++ b/tests/stack/test-backup.sh
@@ -26,14 +26,14 @@
 # and nothing between those two original positions reads anything this file's fixtures produce.
 #
 # Re-derivations (the sandbox-builder WALLET trap — see #1305, still open on this lane):
-# - $WALLET: set by lib.sh's build_val_sandbox() (as the checksum-valid $VALID_PRIMARY constant),
-#   which run.sh's "config validation" black-box calls once, far earlier than every section below
-#   that reads it as a config.json field value. That section stays in run.sh (a generic multi-field
-#   validator, not a backup concern), and this file never needs $V itself (each fixture above builds
-#   its own sandbox instead of reusing the shared one) — so re-deriving the bare string is enough;
-#   calling the heavier build_val_sandbox() would build a $V this file never touches.
-# - $VALID_TARI is a plain lib.sh top-level constant (not build_val_sandbox()-scoped), so it needs no
-#   re-derivation here or anywhere else.
+# - $WALLET: set by lib.sh's build_val_sandbox() (as the checksum-valid $VALID_PRIMARY constant), which
+#   the "config validation" black-box calls once, ahead of every section below that reads it as a
+#   config.json field value. That section lives in test-config.sh, sourced ahead of this file (a
+#   generic multi-field validator, not a backup concern), and this file never needs $V itself (each
+#   fixture above builds its own sandbox instead of reusing the shared one) — so re-deriving the bare
+#   string is enough; calling the heavier build_val_sandbox() would build a $V this file never touches.
+# - $VALID_TARI is a plain lib.sh top-level constant (not
+#   build_val_sandbox()-scoped), so it needs no re-derivation here or anywhere else.
 WALLET="$VALID_PRIMARY" # checksum-valid mainnet primary (the XMRig donation address) — see #1305
 
 echo "== unit: stack_backup — one bounded retry on a tar race (#970) =="
--- a/tests/stack/test-control-upgrade.sh
+++ b/tests/stack/test-control-upgrade.sh
@@ -15,12 +15,12 @@
 # control channel (#33)" setup, and define $REQS/$RESULTS the same way (not lib.sh globals; only
 # these two are read below).
 #
-# $WALLET, set explicitly: lib.sh's control_config() closure reads it (lib.sh:207) without ever
-# setting it — in run.sh it was already set globally by build_val_sandbox()'s far-earlier
-# config-validation section. Same trap as module 7's onion-report fix: $VALID_PRIMARY is what
-# WALLET equals there, and it's always bound. Since FIXED in lib.sh (#1305): both sandbox builders
-# default WALLET now, so a future control-sandbox domain file no longer has to set it. The line
-# below is kept because it is correct and harmless, not because anything still requires it.
+# $WALLET, set explicitly: lib.sh's control_config() closure reads it without ever setting it — it used
+# to arrive already set globally, from the config-validation section's far-earlier build_val_sandbox()
+# call back when that section was inline in run.sh. Same trap as module 7's onion-report fix:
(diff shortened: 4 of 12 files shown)


VijitSingh97:
**Verdict: MERGE.** Non-author controller review, evidence split by register.

**Re-derived by me (not relayed):**
- **Comment-only:** 152 changed lines, 0 non-comment; control: the same probe on the R13 cut returns 305, so it detects code. Line-count neutral in all 12 files, summing to the stated +76/−76.
- **Suite pair, re-derived from the logs at `~/split-baselines/1520-fa49be13/`:** rc 0/0; 2992 pass-glyphs / 0 fail-glyphs on both legs counted independently of the suite's own summary line; 3306 log lines both; **275 section headers per side, counted per side before comparing** (a zero-match pattern cannot print IDENTICAL here); ordered header comparison identical, with a permutation control that fires (2 differing lines on a swapped copy). Full-log diff is exactly 4 lines = the 2 known nondeterministic pairs (mktemp path, ed25519 fingerprint).
- **The recorded pair head `20cb8dac` differs from this PR's head `5f3b5fc` — resolved by measurement:** identical trees (`0f23ba94`), identical parent (`fa49be13`); a message-only amend, so the pair evidence measures exactly this head's program.
- **Base-move hazard cleared:** `docs/dev/file-budget.tsv` is byte-identical either side of the tip move this branch has not rebased over (control: the R13 merge does move it), and this diff is comment-only, so no gate row can differ at the branch head.
- **Prose truth of the replacements:** all 14 files named in added lines exist at head (two deliberately-wrong needles reported MISS); all 8 sourcing-order claims ("sourced ahead of/AFTER this file") verified against run.sh's actual source sequence, including the one inverted case (test-dashboard @30 vs test-config @46); content spot-checks hit with a 0-scoring control on an unrelated file; the surviving run.sh-mentioning comments sample as past-tensed or currently-true (48 → 33 by my counting convention).
- The comment-width claim: over-110 comment lines 2 at head, 2 at base — the defect class introduced is zero; the body's "102–106 per file" phrasing slightly overreaches (`test-release-signing.sh` has a pre-existing, untouched 111) — imprecision, not a defect.

**Relayed, not re-derived:** the author's full 46-stanza location enumeration and its 49-site read-and-left judgement (its sweep convention differs from my 33-count pattern; both directions consistent, no discrepancy filed); the perturbation control (diff 4→6).

Merging by the standing shape; #1520 is closed by hand by its author after merge (closing keywords are inert off the default branch).

PR #1528, opened 2026-08-29
fix(cli): take the mutation lock for the three reset verbs (#1482)

### LINKED ISSUE #1482: Six mutating verbs and the in-place upgrade fallback still take no mutation lock (#1342 residual)

VijitSingh97:
**DRAFT — do not merge yet, and not by me: the controller routes the non-author pass.** The regression test AND its `tests/stack/run.sh` registration are both in, and both commits are rebased onto the post-R14 tip. Read `Shell tests (shellcheck + pithead suite)` before anything else — it is the only check that exercises this diff, and now the only one that runs the 28 new assertions.

## What this changes

`factory_reset`, `config_reset` and `reset_dashboard` each mutate the machine while holding no mutation lock, so a concurrent `pithead` verb can interleave with a wipe. This takes the window for all three.

`os_update` was already locked (`15-os-update.sh:228`). That leaves `rotate_secrets` and the firstboot installer path still unlocked, so **this PR does not close #1482** — it is a slice, and the residual is named rather than left for a reader to discover.

## Placement, and why it is where it is

Copied from the rule `setup` already follows:

- **After each verb's confirmation prompt**, so the hold never spans a human wait.
- **Before its first mutation**, so the timeout message's promise that nothing has changed is literally true.

`config_reset` is the sharpest of the three: it removes `config.json`, so a window taken any later would let a timeout abandon a box whose configuration was already half gone.

**Both reboot paths keep `reset_reboot_or_hint` INSIDE the window deliberately.** Releasing between arming the wipe and the reboot is precisely the half-reset gap the lock exists to prevent.

## Two judgement calls, stated rather than left silent

**No per-verb exit-status routing, and that is deliberate.** #1391's F3 hazard arises because `firstboot_wizard` wraps `setup` and interprets its exit status. `51-main-dispatch.sh:121-124` is the only call site of each reset verb and calls all three **bare**, so `mutation_lock_acquire`'s own exit with `PITHEAD_EX_LOCK_TIMEOUT` reaches the operator unmodified. Re-derived at this head, not carried from the issue. Accepted as a ruling by the controller.

**`reset_dashboard` has an `exit 1` path that bypasses the explicit `mutation_lock_release`.** I checked this rather than assuming: the lock is an `exec 9>>` file descriptor on the shell process (`00-prelude.sh:259`), so process exit closes it and the OS drops the flock. No leak, and adding a trap would diverge from the `os_update` precedent (`15-os-update.sh:228`/`:268`) for no gain.

"The fd closes" is only half of what `mutation_lock_release` does, so the other half is stated here rather than left for a reviewer to find. It also runs `: >"$_PITHEAD_LOCK_PATH"`, clearing the **holder record**, and the bypassing `exit 1` path leaves that line in place. That is handled by design and not by luck: `mutation_lock_holder` gates the record on `[ -d /proc/$pid ] || kill -0 $pid` and reports `holder unrecorded` for a dead pid, and the next acquire appends over it regardless. The only consequence is that a concurrent waiter sees `holder unrecorded` instead of a name, on a path that is already erroring out. Not a defect, and not a reason to add a trap.

## What was run, at this head

Everything below was re-run **after** the rebase onto `origin/develop-v2` (`9f0fe17`), not carried over from the pre-rebase evidence — a moved base invalidates the proof set, not just the suite.

- `bash -n` — clean on both the slice and the artifact.
- `shfmt -i 4` — diff of **0 bytes**.
- `make lint-pithead-parity` — OK, 52 slices.
- `scripts/lint-file-budget.sh` — OK. `16-reset.sh` 156 -> 181 lines (via the gate's own `awk 'END{print NR+0}'`, not `wc -l`), no tsv row, under 400.
- Base ref probed **before and after** the gate run — identical both times, so no lane's fetch moved it mid-run.
- Added lines are identical in the slice and the artifact. The artifact was regenerated by `scripts/build-pithead.sh` only and never hand-edited; parity is what proves that.

## The regression test

`tests/stack/test-appliance-reset-lock.sh`, a new sibling file — 186 lines, no budget row needed.

**Scope, stated so it is not read as duplicating `test-lifecycle.sh`.** That file proves the lock *mechanism* — holder record, announced wait, refusal, rc, release — driven through `stack_down`. None of it says whether any *other* verb reaches that mechanism, and these three did not. So this asserts **wiring and placement**, one verb at a time, and nothing about how the lock itself behaves. The rc assertions are the cheapest proof that a verb reaches `mutation_lock_acquire` at all, not a second opinion on the exit status.

**Placement is pinned from both sides**, because a test that only asks "does it take the lock" passes with the acquire in the wrong place:

- *Before the first mutation* — on a held machine each verb must leave `config.json`, the ESP marker, the containers and the data directories exactly as it found them. The refusal promises the operator nothing changed; these assertions are what make that literally true.
- *After the confirmation* — a verb given the wrong confirmation word on a held machine must come back on the word, never having entered the lock wait. Move the acquire above the prompt and an operator answering "no" is made to wait for a window the verb was about to decline.

**Every contended case has an uncontended positive control** asserting the same fixture really does perform the mutation being looked for. Without it, "`config.json` is still here" is equally consistent with a fixture that never armed — and the two read identically.

### Proven able to fail

Run through a standalone driver mirroring `run.sh`'s contract (`set -uo pipefail`, `lib.sh` first, verdict from the PASS/FAIL counters), because the file is not registered in the suite yet. Each mutant was proven **applied** by `git diff --numstat` on the artifact before its run, and judged by its expected text rather than by rc.

| leg | result |
|---|---|
| clean | 28 passed, 0 failed |
| **M1** — the three `mutation_lock_acquire` lines deleted (the pre-#1528 state) | **14 failed**, every one in the contended block, each naming the real mutation that got through: `config.json` gone, ESP marker armed, containers removed, data dirs deleted |
| **M2** — `config-reset`'s acquire moved above its confirmation prompt | **2 failed**, only the two `config-reset` placement rows; that run timed out (`Timed out after 1s ... nothing was changed`) instead of reporting the decline |
| clean re-run, after restore | 28 passed, 0 failed |

The two mutants kill **disjoint** row sets, so they are two instruments rather than one printed twice. The restore between legs was verified byte-identical (`cmp` against a pre-mutation copy), which is what makes the closing green a control rather than a repeat.

Also run on the test file at this head: `bash -n` clean, `shfmt -i 4` diff **0 bytes**, `shellcheck --severity=warning` (rc 0, run alongside `tests/stack/lib.sh` exactly as the real recipe pairs them), and `scripts/lint-file-budget.sh` OK **with the file staged** — the gate enumerates `git ls-files`, so an unstaged file is invisible to it and the run would have passed while proving nothing. Base ref probed before and after: unmoved.

## Registration, and the ordering it waited on

`tests/stack/run.sh` +3 lines, 237 -> 240, registering the new file immediately after
`test-appliance-reset.sh`. No tsv row either way. The new file needs nothing from any sibling — only `lib.sh` names — so unlike the editable-allowlist file it carries no adjacency constraint.

This was serialised behind the tests lane's R14 (**PR #1529**) by agreement, because R14 rewrote the region the stanza lands in. **R14 has merged.** Both commits are rebased onto the tip it produced, and the whole mechanical proof set was **re-run at the BRANCH head** rather than carried: a green merged tree answers what lands, not whether this PR's own evidence still stands. Base ref probed identical either side of the gate run.

## What to check that I have not

**A full local `tests/stack/run.sh` at the rebased head was still running when this was written, so I am not claiming it green.** CI's `Shell tests` job is the instrument to read here, and it is now the run that actually executes the 28 assertions — before the stanza they were linted but never run.

*Corrections to earlier versions of this section:* it named the lane guard as a blocker (the `tests/stack/` grant is armed — verified by **running** the guard against a staged file, rc 0), and later said the stanza was still missing (it is committed). Both are recorded here rather than quietly edited away.

## Provenance of the evidence above

Stated because a relay reads as recall. The slice-half proof set — `bash -n` and `shfmt` on `16-reset.sh` and the artifact, `lint-pithead-parity` at 52 slices, the budget figures for `16-reset.sh`, the artifact regeneration — was run by the previous session at this same base. That base has since MOVED (R14), so I did not carry it: after rebasing onto `90f5f53` I re-ran the slice-half gates myself at the branch head — `lint-pithead-parity` OK at 52 slices, budget gate OK, base probed identical either side. Everything in "The regression test" above I ran myself.

### DIFF
--- a/lib/pithead/16-reset.sh
+++ b/lib/pithead/16-reset.sh
@@ -17,6 +17,12 @@ reset_dashboard() {
         fi
     fi
 
+    # After the confirmation, never before it: the hold must not span a human wait, which is the
+    # placement rule `setup` already follows (#1391). Everything below this line mutates — the
+    # containers are removed and both data directories are deleted — so a lock timeout here has
+    # genuinely changed nothing, which is exactly what the timeout message promises the operator.
+    mutation_lock_acquire reset-dashboard
+
     log "Resetting dashboard and p2pool..."
     # Resolve what to delete from .env (the LIVE deployment's data dirs), NOT a fresh config.json
     # parse — otherwise editing a *.data_dir in config.json before `reset-dashboard` (without an
@@ -61,6 +67,7 @@ reset_dashboard() {
         warn "Fix the cause shown above, then re-run '$0 up' to bring them back."
         exit 1
     fi
+    mutation_lock_release
 }
 
 # --- Two-tier reset (config / factory) -----------------------------------------------------------
@@ -108,13 +115,22 @@ config_reset() {
         }
     fi
 
+    # After the typed confirmation, so the hold never spans a human wait (#1391), and before the
+    # first mutation below — this verb REMOVES config.json, so a window taken any later would let a
+    # timeout abandon a box whose configuration was already half gone.
+    mutation_lock_acquire config-reset
+
     detect_os 2>/dev/null || true
     docker compose down --remove-orphans 2>/dev/null ||
         warn "compose down failed (engine not running?) — continuing with the config wipe."
     remove_tor_egress_firewall 2>/dev/null || true
     rm -f "$CONFIG_FILE" "$ENV_FILE" Caddyfile
     log "Configuration cleared — the first-boot wizard owns the next boot."
+    # The reboot stays INSIDE the window on purpose. Releasing after the wipe but before the reboot
+    # would leave a gap in which another verb could start against a box that has no config.json yet
+    # still has its old containers' state — the half-reset window this lock exists to prevent.
     reset_reboot_or_hint
+    mutation_lock_release
 }
 
 # factory-reset: the deep wipe. /data cannot be reformatted while it is mounted and holding the
@@ -145,6 +161,12 @@ factory_reset() {
         }
     fi
 
+    # Taken after the typed confirmation so the hold never spans a human wait (#1391). The marker
+    # write below is a single line, but it is not the mutation that matters: arming it commits the
+    # machine to a reboot that reformats /data, and an operation already holding the window would
+    # otherwise be rebooted out from under itself mid-write.
+    mutation_lock_acquire factory-reset
+
     # Arm the boot-time wipe. The marker lives on the ESP, not /data, precisely because /data is
     # what gets erased. If the ESP is not writable the wipe cannot be armed, so refuse loudly
     # rather than reboot into a box that quietly did nothing.
@@ -152,5 +174,8 @@ factory_reset() {
     : >"$marker" 2>/dev/null ||
         error "Could not arm the reset marker at $marker (ESP not mounted?) — nothing was wiped."
     log "Reset armed. Rebooting to reformat /data and reopen the wizard..."
+    # Held across the reboot for the same reason as config-reset: the gap between arming the wipe
+    # and the reboot is precisely when another verb must not start.
     reset_reboot_or_hint
+    mutation_lock_release
 }
--- a/pithead
+++ b/pithead
@@ -3636,6 +3636,12 @@ reset_dashboard() {
         fi
     fi
 
+    # After the confirmation, never before it: the hold must not span a human wait, which is the
+    # placement rule `setup` already follows (#1391). Everything below this line mutates — the
+    # containers are removed and both data directories are deleted — so a lock timeout here has
+    # genuinely changed nothing, which is exactly what the timeout message promises the operator.
+    mutation_lock_acquire reset-dashboard
+
     log "Resetting dashboard and p2pool..."
     # Resolve what to delete from .env (the LIVE deployment's data dirs), NOT a fresh config.json
     # parse — otherwise editing a *.data_dir in config.json before `reset-dashboard` (without an
@@ -3680,6 +3686,7 @@ reset_dashboard() {
         warn "Fix the cause shown above, then re-run '$0 up' to bring them back."
         exit 1
     fi
+    mutation_lock_release
 }
 
 # --- Two-tier reset (config / factory) -----------------------------------------------------------
@@ -3727,13 +3734,22 @@ config_reset() {
         }
     fi
 
+    # After the typed confirmation, so the hold never spans a human wait (#1391), and before the
+    # first mutation below — this verb REMOVES config.json, so a window taken any later would let a
+    # timeout abandon a box whose configuration was already half gone.
+    mutation_lock_acquire config-reset
+
     detect_os 2>/dev/null || true
     docker compose down --remove-orphans 2>/dev/null ||
         warn "compose down failed (engine not running?) — continuing with the config wipe."
     remove_tor_egress_firewall 2>/dev/null || true
     rm -f "$CONFIG_FILE" "$ENV_FILE" Caddyfile
     log "Configuration cleared — the first-boot wizard owns the next boot."
+    # The reboot stays INSIDE the window on purpose. Releasing after the wipe but before the reboot
+    # would leave a gap in which another verb could start against a box that has no config.json yet
+    # still has its old containers' state — the half-reset window this lock exists to prevent.
     reset_reboot_or_hint
+    mutation_lock_release
 }
 
 # factory-reset: the deep wipe. /data cannot be reformatted while it is mounted and holding the
@@ -3764,6 +3780,12 @@ factory_reset() {
         }
     fi
 
+    # Taken after the typed confirmation so the hold never spans a human wait (#1391). The marker
+    # write below is a single line, but it is not the mutation that matters: arming it commits the
+    # machine to a reboot that reformats /data, and an operation already holding the window would
+    # otherwise be rebooted out from under itself mid-write.
+    mutation_lock_acquire factory-reset
+
     # Arm the boot-time wipe. The marker lives on the ESP, not /data, precisely because /data is
     # what gets erased. If the ESP is not writable the wipe cannot be armed, so refuse loudly
     # rather than reboot into a box that quietly did nothing.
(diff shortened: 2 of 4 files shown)


VijitSingh97:
## Verdict: MERGE

The non-author pass returned **MERGE-WITH-NOTES** with no blocking defect on any axis. I confirm it and am merging. Recording what each of us actually checked, because the two halves are different and the distinction is the point.

### What the pass did that a reading could not

It did not stop at reading the code and the CI log. It rebuilt the mutation harness in an isolated clone and **ran the mutants itself** — clean 28/0, M1 14 failures confined to the contended block, M2 26/2 with the literal timeout message, then a byte-identical restore back to 28/0. Disjointness of M1 and M2 was settled by diffing the two failing-label lists rather than by argument: M1 kills rows 9–22, M2 kills rows 27–28, zero overlap. That is what makes them two instruments rather than one measurement printed twice.

It also went past the evidence this PR offered. The PR mutation-tested placement on `config-reset` only; `reset-dashboard` and `factory-reset` were argued by symmetry. The pass built the same mutation against each of them and ran it — 26/2 each, exactly the two placement rows, zero bystanders. **Placement is now measured for all three verbs rather than measured for one and inferred for two.**

Every absence claim in the pass carries a fired positive control: the fail-glyph sweep, the `domain_ran` guard sweep, and the fixture's own `bad "the holder takes the lock window"` line, which is absent from the CI log and was shown present in the source file by the same grep first.

### What I re-derived myself, and one place my own check lied to me

**Lock placement, at source in `lib/pithead/16-reset.sh` at the head.** `reset_dashboard`: confirm at 13, acquire at 24, first `rm -rf` at 44, release at 70. `config_reset`: acquire at 121, `docker compose down` at 124, release at 133. `factory_reset`: acquire at 168, ESP marker write at 173, release at 180. All three acquire **after** the human confirmation and **before** the first mutation, so no hold spans a human wait and no mutation runs outside one. This is the claim the merge turns on and it matches what the pass measured by mutation.

**The budget gate, on both trees.** The base moved under this branch (its merge-base is `90f5f53`; the tip is now two commits further on), so this is the wide case and one run would not have answered both questions. Merged tree: 85 rows, 0 over. Branch head: 85 rows, 0 over. Control fired by forcing one ceiling below its file's real length, which correctly reported that row and only that row.

**Checks: 15 pass, 2 skipping, 0 pending, 0 failing** at `821ff0e`, including `Shell tests` — the only job that exercises this diff.

**On finding 1, my first check falsely cleared it.** I grepped the body for both shas, got zero matches, and nearly recorded the finding as not reproducing. The grep was broken, not the finding — an alternation pattern that silently matched nothing. Re-run with a positive control on a string I knew was present, both shas appear exactly once, and the finding is exact: line 30 says the rebase was onto `9f0fe17`, line 82 says `90f5f53`, and the branch's true parent is `90f5f53` — `9f0fe17` is one merge older. A check that has never matched anything is not evidence that there is nothing to match.

### Disposition of the two notes

**1 — the stale sha, corrected here rather than in a commit.** A loose claim in a PR body is a merge-record defect: it is corrected in the verdict and never force-pushed, because rewriting a reviewed head to fix prose costs more than it buys. **The rebase point for all the evidence in this PR is `90f5f53`; the `9f0fe17` on line 30 is stale text from an earlier draft.** The gates were in fact anchored to the right commit — the pass re-ran them and so did I. The reason this is worth a line at all is the one the PR itself argues elsewhere: a wrong figure sitting under a right conclusion has no tell.

**2 — not a defect, and worth carrying forward.** The pass closed the placement gap itself and it held. Recording it because the next slices of #1482 (`rotate_secrets`, the firstboot installer path) also have interactive confirms, and "argued by symmetry" should not be the standard there either.

### #1482 stays open

This locks three of the six mutating verbs it names. `rotate_secrets` and the firstboot installer path still take no lock, which the PR body discloses correctly. A close trigger phrased as "the named change landed" goes ambiguous the moment the named change ships in slices, so the issue is checked, not the PR — and the issue is not done.

Separately, and relayed from the pass rather than re-derived by me: **#1482's own text is now stale**, listing `os_update` among the unlocked paths when #1514 covered it since. That is the tracking issue falling behind its own slices, not a defect in this PR.

### What I did not do

I did not re-run the mutation battery myself — the pass ran it in an isolated clone and a second copy of the same measurement is not a second instrument. I ran no suite locally at this head, and none exists: the lane's own local run was reaped mid-run and left no rc file, so **CI is the only thing that has ever executed these 28 assertions**. Nothing here touches real hardware, a real ESP, or a real wipe; the suite's `sudo` is a logging stub by design. Pid-reuse racing the holder record remains an accepted narrow gap disclosed in `00-prelude.sh` and is outside this PR.

PR #1527, opened 2026-08-29
docs(dev): record the #1480 config-layer decision (ADR, option 2)

VijitSingh97:
The ADR #1480's acceptance asks for: `docs/dev/config-layer-language.md` choosing **option 2** — schema-driven validation over `jq` against `config.reference.json`, render stays bash, no Python port — with the host-dependency question answered at source. Plus the `docs/README.md` index row. Docs-only, 2 files, +81/-0.

The full ruling with grounds is on the issue: https://github.com/p2pool-starter-stack/pithead/issues/1480#issuecomment-5463250489

**Every figure in the doc was re-derived this session at `develop-v2` (`32be481f`), not relayed:** `jq` call-lines/guards 38/0 (`28-parse-and-validate-config.sh`) and 35/0 (`33-render-env.sh`), with a fired control — `04-status.sh` scores 3/1, so the counter discriminates guarded from unguarded files; 29 of 52 `lib/pithead/` modules invoke `jq`; module sizes 485/494; every host python3 site soft-guarded (`25-address-types.sh:27,:144`) and degrading; doctor `dr_info` for python3 (`06-doctor.sh:105`) vs `dr_fail` for jq (`:14-17`); `jq` installed in the final rootfs stage (`os/rootfs/Dockerfile:143`, final `FROM` at `:118`); Compose prerequisites at `docs/getting-started.md:31/:47/:95`.

**Ran:** `scripts/lint-docs-voice.sh` (40 prose docs clean), `scripts/lint-topology-classes.sh` with self-test, and the topology detector on the doc and this body. **Not run:** shell/test suites — the diff contains no executable file; `docs/dev/file-budget.tsv` carries no docs rows, so the budget gate is not in scope.

No `Fixes` keyword on purpose: closing keywords only act on the default branch, so #1480 is closed by hand after merge.

PR #1529, opened 2026-08-29
test(stack): cut the editable-allowlist round-trip to its own domain file (#1105 R14)

VijitSingh97:
## What this is

`#1105` R14. The `#522` editable-allowlist commit round-trip moves out of `tests/stack/run.sh`
into its own domain file, `tests/stack/test-control-editable-allowlist.sh`.

`run.sh` 277 -> 237. The moved content is byte-identical and the source stanza sits at the
section's exact former position, so **execution order is unchanged** — no reorder, no anchor
decision, no nested stanza in the range.

## The two decisions this PR makes, and the grounds

**Homed in the control family, not the config family.** The cut map put this row with the config
family on its subject: it flips config keys. Its fixture says otherwise. Every assertion goes
through `gate_try()` against the control channel's sandbox and its request/result spool, and the
fixture is what a domain file has to reproduce. The map left the target to be decided at cut time
by fixture affinity; this is that decision, recorded rather than assumed.

A second, mechanical reason not to fold it into `test-control-add-only-ssrf.sh` (which defines
`gate_try()` and would have internalised the dependency): that file is **398 lines against the
gate's 400-line target**, so an append would push it to 441 and force a new budget row on a file
the split exists to keep small.

**The new file deliberately does NOT self-arm, and the reason is not that arming would break it.**
I checked `build_control_sandbox()` at source rather than relaying the cut map's "self-arm is the
defect" note: it `mkdir -p`s `data/monero`, `data/tari`, `data/p2pool/stats`, `data/tor` and
`data/dashboard`, and copies static repo inputs. **It never touches `data/control/`**, so the
map's recorded coupling-rule-A RED does not transfer to this position and I am not claiming it
does. The actual argument is narrower: arming here would add a call that does not exist today, and
would buy nothing, because `gate_try()` is a function defined in another domain file and no arm
written here can supply it. The cut that changes nothing about what executes is the one whose
proof is strongest. The ambient inheritance is disclosed in the header instead, on the shipped
`test-control-add-only-ssrf.sh` precedent.

## What was RUN

**Suite pair, both legs in the SAME worktree** (`/tmp/tests-522`), base leg at `546d56b6`, branch
leg at `d1153159`. Verdicts `~/split-baselines/r14-546d56b6/{before,after}.rc`, both **rc 0**.

The four figures were **predicted from the code before the run** and the run had to satisfy them:
the range holds exactly one `^echo "== ` header and zero mktemp/keygen/date nondeterminism, and
`domain_ran` emits an assertion only on failure, so nothing should move. It did not:

| | before | after |
|---|---|---|
| passed / failed | 2992 / 0 | 2992 / 0 |
| log lines | 3306 | 3306 |
| section headers | 275 | 275 |
| verdict lines | 2992 | 2992 |

- **Header set identical; header ORDER identical, 0 of 275 positions differ.** Match counts are
  non-zero per side (275 / 275) — checked first, because a zero-match pattern is not an instrument.
- **Permutation control FIRED** (adjacent swap: ordered diff fires, sorted diff stays silent).
- **Full-output diff: 2 lines = 1 pair**, the known `#1325` ed25519 keygen fingerprint. Both legs
  ran in one worktree, so there is no cross-worktree path-normalisation residue to explain away.
  **Perturbation control FIRED** (one injected line takes the diff 2 -> 4).
- **Byte-identity of the moved content re-derived FROM GIT**, not from the extraction: base
  `run.sh` 141-182 against the new file's body is identical, with a control against a shifted
  range that FIRED.
- **Conservation exact:** 237 + 42 moved + 1 blank - 3 stanza = 277.
- **Rebase content-neutral by PATCH identity**, not tree — the base moved, so the trees must
  differ. Byte-identical patches, control against a path-scoped patch FIRED.
- `shellcheck --severity=warning` rc 0 on **both** Makefile invocations, run separately and in the
  exact form the Makefile writes them (the pairing is load-bearing: a truncated first invocation
  reported three SC2034 in `test-appliance-boot.sh` that vanish once `os/overlay/pithead-boot` is
  named alongside it, which is the documented pairing, not a defect). `bash -n` rc 0 on all four
  files; `shfmt -i 4 -d` rc 0.
- **Budget gate rc 0, with its `--self-test` run first and a control that FAILED BY NAME**
  (`test-config.sh` forced to 754 against its 753 ceiling). No tsv row changes: the new file (86
  lines) and `run.sh` (237) both sit under the 400 target. Base ref probed identical before and
  after the gate run.
- **Ruling 8a: this is CASE 2 — the tip moved under me** (`#1527`, docs-only), so both runs were
  made rather than one measurement reported twice. Merged tree: gate **rc 0**, `run.sh` 237, merge
  clean, `file-budget.tsv` untouched. Branch head: gate **rc 0**. `#1527` touches no file under
  `tests/stack/`, so the pair still measures the branch's program.

## Prose this cut falsifies, fixed here

A cut falsifies comments in files it never touches, so the moved header was swept mechanically
(base-vs-head header diff, then `grep -F` on the name with the `black-box: ` prefix stripped). The
sweep **fired on the pre-change side**, which is what licenses reading its post-change result.

Two sites, both fixed, both comment-only and **line-count neutral**, with max COMMENT-line length
unchanged per file (107 and 115) — the reflow trap is to hold the line count by pushing freed words
onto the last line and silently blow the wrap width:

- `test-control-add-only-ssrf.sh` said `run.sh`'s `#522` section reads `gate_try()`/`$UUID5`.
- `test-config.sh` listed the section among what it left behind in `run.sh`.

The name grep alone found only the second; the first was found by widening to `#522` and reading
the neighbourhood, which is exactly the documented floor behaviour of that instrument.

## What is NOT proven

- The pair proves the suite runs the same sections in the same order and makes the same
  assertions. It does **not** prove the new header's prose is correct — that is the review's job.
  Every claim in it was re-derived at source (including `run.sh`'s `set -uo pipefail`, which is why
  a missing `gate_try()` is loud rather than fatal), but that is a separate argument from the run.
- **No mutation proof.** The header deliberately claims none: I did not run one, and asserting a
  kill I have not seen would be worse than the gap.
- The file is **not sourceable standalone** and the header says so plainly. The carry-over guard
  fails by name, but no guard written here can supply `gate_try()`.

## Over-engineering pass — what it actually changed

Run against the only authored content in this diff (the new file's header) plus the two comment
fixes. Four outcomes, three of them removals:

1. **An unearned assertion was narrowed.** The header was going to carry the cut map's "self-arm
   here is the defect — a fresh `$C` reproduces the recorded RED". I read `build_control_sandbox()`
   at source instead of relaying it and found it never touches `data/control/`, so that RED does
   not transfer to this position. The claim is now the narrower one I can actually defend. This was
   the single biggest finding: a relayed claim reads exactly like a derived one.
2. **A guard was considered and rejected.** A `declare -F gate_try` check would make the
   cross-file dependency machine-enforced, but there is no such guard anywhere in this suite, and
   inventing a convention inside a move PR is the definition of widening. The failure is already
   loud (`set -uo pipefail`, no `-e`), so the header states it rather than guarding it.
3. **No mutation-proof paragraph was written**, because no mutation was run. The sibling file's
   header has one; copying that shape without the run would have been the worst option available.
4. **Header kept at its length** (41 lines against the shipped sibling's 35). Every paragraph
   carries a decision a reviewer would otherwise have to re-derive, so it stays.

### DIFF
--- a/tests/stack/run.sh
+++ b/tests/stack/run.sh
@@ -138,48 +138,8 @@ _d0=$((PASS + FAIL)) && source "$HERE/test-data-management.sh" && domain_ran tes
 # shellcheck source=tests/stack/test-control-add-only-ssrf.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-control-add-only-ssrf.sh" && domain_ran test-control-add-only-ssrf.sh "$_d0" "$?" || domain_ran test-control-add-only-ssrf.sh "$_d0" "$?"
 
-echo "== black-box: editable-allowlist commit round-trip, every key (#522) =="
-# Every key on CONTROL_DASHBOARD_EDITABLE_KEYS must actually round-trip a real preview->commit
-# through the approval gate and land in config.json — not just pass a describe_change unit check.
-# Fresh baseline with each tunable at a known value so every row below is a genuine single-key
-# env diff (pool flips P2POOL_FLAGS + P2POOL_PORT, both allowlisted).
-jq -n --arg w "$WALLET" '{
-    monero:{mode:"local",wallet_address:$w,node_username:"u",node_password:"p",mem_limit:"4g",prep_blocks_threads:4},
-    tari:{wallet_address:"'"$VALID_TARI"'",mem_limit:"3g"}, p2pool:{pool:"main"},
-    xvb:{enabled:true,donation_level:"donor"}, telegram:{daily_summary_time:"08:00"},
-    dashboard:{secure:true,host:"box.lan",tari_required:true,check_for_updates:true,timezone:"UTC",
-               hashrate_drop_threshold:50,hashrate_drop_minutes:10,
-               auth:{username:"admin",password:"a control passphrase"},control:{enabled:true}}}' >"$C/config.json"
-(cd "$C" && DOCKER_LOG="$CTRL_LOG" PATH="$C/bin:$PATH" ./pithead apply -y >/dev/null 2>&1)
-roundtrip_key() { # <label> <jq-set> <jq-read> <expected>
-    jq "$2" "$C/config.json" >"$C/cand.json"
-    gate_try "$C/cand.json"
-    assert_eq "$1 commit applies through the gate" "$(jq -r '.status' "$RESULTS/$UUID5.json" 2>/dev/null)" "applied"
-    assert_eq "$1 landed in config.json" "$(jq -r "$3" "$C/config.json")" "$4"
-}
-roundtrip_key "XVB_ENABLED" '.xvb.enabled=false' '.xvb.enabled' "false"
-roundtrip_key "XVB_DONATION_LEVEL" '.xvb.donation_level="whale"' '.xvb.donation_level' "whale"
-roundtrip_key "TARI_REQUIRED" '.dashboard.tari_required=false' '.dashboard.tari_required' "false"
-roundtrip_key "DASHBOARD_FAIL_CLOSED" '.dashboard.fail_closed=true' '.dashboard.fail_closed' "true"
-roundtrip_key "DASHBOARD_CHECK_UPDATES" '.dashboard.check_for_updates=false' '.dashboard.check_for_updates' "false"
-roundtrip_key "DASHBOARD_TZ" '.dashboard.timezone="Europe/Paris"' '.dashboard.timezone' "Europe/Paris"
-roundtrip_key "MONERO_MEM_LIMIT" '.monero.mem_limit="5g"' '.monero.mem_limit' "5g"
-roundtrip_key "TARI_MEM_LIMIT" '.tari.mem_limit="2g"' '.tari.mem_limit' "2g"
-roundtrip_key "MONERO_PREP_THREADS" '.monero.prep_blocks_threads=8' '.monero.prep_blocks_threads' "8"
-roundtrip_key "HASHRATE_DROP_THRESHOLD_PCT" '.dashboard.hashrate_drop_threshold=40' '.dashboard.hashrate_drop_threshold' "40"
-roundtrip_key "HASHRATE_DROP_MINUTES" '.dashboard.hashrate_drop_minutes=15' '.dashboard.hashrate_drop_minutes' "15"
-roundtrip_key "TELEGRAM_DAILY_SUMMARY_TIME" '.telegram.daily_summary_time="09:30"' '.telegram.daily_summary_time' "09:30"
-roundtrip_key "P2POOL_FLAGS/P2POOL_PORT" '.p2pool.pool="mini"' '.p2pool.pool' "mini"
-# The 25 allowlisted TELEGRAM_EVENT_* toggles (raffle_win added 2026-08: audit found it was the one
-# event toggle missing from its siblings, all otherwise editable). wallet_changed + clearnet_exposed
-# are deliberately NOT on the allowlist (tamper-evidence alarms; their refusal is asserted above),
-# so they are excluded here. Each flips true->false as a single-key diff.
-for ev in node_down node_recovered worker_offline worker_recovered worker_joined worker_left \
-    sync_finished disk_space db_unhealthy db_reset xvb_no_share xvb_registration new_release \
-    stack_online daily_summary hashrate_low hashrate_loss hugepages low_ram high_reject_rate \
-    block_found payout_found payout_confirmed container_unhealthy raffle_win; do
-    roundtrip_key "TELEGRAM_EVENT ${ev}" ".telegram.events.${ev}=false" ".telegram.events.${ev}" "false"
-done
+# shellcheck source=tests/stack/test-control-editable-allowlist.sh disable=SC2015
+_d0=$((PASS + FAIL)) && source "$HERE/test-control-editable-allowlist.sh" && domain_ran test-control-editable-allowlist.sh "$_d0" "$?" || domain_ran test-control-editable-allowlist.sh "$_d0" "$?"
 
 # shellcheck source=tests/stack/test-worker-config.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-worker-config.sh" && domain_ran test-worker-config.sh "$_d0" "$?" || domain_ran test-worker-config.sh "$_d0" "$?"
--- a/tests/stack/test-config.sh
+++ b/tests/stack/test-config.sh
@@ -32,8 +32,8 @@
 #   helper probes the #1252 domain map does not name for this module, and this PR moves exactly what
 #   the map names.
 # - "black-box: editable-allowlist commit round-trip, every key (#522)" — the multi-key
-#   describe_change exercise. It is a control-channel black-box against the ambient $C sandbox and
-#   stays for the $C-cluster pass, per the map's "clusters last" rule.
+#   describe_change exercise. A control-channel black-box against the ambient $C sandbox, so it was
+#   left for the $C-cluster pass; #1105 R14 has since cut it to test-control-editable-allowlist.sh.
 # - "unit: render-quadlet parity vs os/quadlet fixtures (#77 phase 1)" — a render test by name, but
 #   it checks the appliance's Quadlet unit set against the os/quadlet fixtures; it belongs to the
 #   appliance module, cut next.
--- a/tests/stack/test-control-add-only-ssrf.sh
+++ b/tests/stack/test-control-add-only-ssrf.sh
@@ -24,8 +24,8 @@
 # false — that file's own header states its sections are fully self-contained (#1462).
 #
 # WHAT OUTLIVES THE SOURCE. gate_try() and $UUID5 are defined here and not unset at the end, so they
-# outlive the source as they outlived the old in-run.sh position: run.sh's #522 section still calls
-# gate_try() and reads $UUID5, and test-spool-audit.sh reuses $UUID5. Hence the stanza stays put.
+# outlive the source as they outlived the old in-run.sh position: the editable-allowlist domain file
+# run.sh sources next reads both, as test-spool-audit.sh reuses $UUID5. Hence the stanza stays put.
 #
 # MUTATION PROOF: reverting pithead's add-only prefix check back to "refuse any workers.list
 # diff" turns the ADD-ONLY-append assertion red; reverting _control_host_is_internal's
--- a/tests/stack/test-control-editable-allowlist.sh
+++ b/tests/stack/test-control-editable-allowlist.sh
@@ -1,0 +1,86 @@
+# shellcheck shell=bash
+#
+# Every key on CONTROL_DASHBOARD_EDITABLE_KEYS must actually round-trip a real preview->commit
+# through the approval gate and land in config.json (#522) — not just pass a describe_change unit
+# check. Split out of run.sh by #1105 R14; the section's own contract is stated at its header below.
+#
+# WHY THE CONTROL FAMILY, AND NOT THE CONFIG FAMILY THE CUT MAP GUESSED. The map homed this row
+# with the config family on the strength of its subject: it flips config keys and reads them back
+# out of config.json. Its FIXTURE says otherwise, and the fixture is what a domain file actually
+# has to reproduce. Every assertion here goes through gate_try() against the control channel's
+# sandbox and its request/result spool, so this section's dependencies are the control channel's,
+# and it is homed with them. The map's row left the target to be decided at cut time by fixture
+# affinity; this is that decision, recorded rather than assumed.
+#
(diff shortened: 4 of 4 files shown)


VijitSingh97:
# Adversarial non-author pass — PR #1529 (#1105 R14, editable-allowlist, #522)

Head reviewed: `d1153159` (`pr1529` fetched from `pull/1529/head`, base `546d56b6`).

**VERDICT: MERGE-WITH-NOTES.** No blocking finding. Two should-fix, one of them a real, currently-uncorrected in-repo doc falsification the PR's own instruments could not have caught.

## 1. The header's claim about itself — HOLDS (re-derived)

`gate_try()` is defined exactly once, at `tests/stack/test-control-add-only-ssrf.sh:56`. The new file
(`tests/stack/test-control-editable-allowlist.sh`) only *calls* it, at line 60; grepping the new file for
a definition returns nothing. `$UUID5` is likewise assigned only in
`test-control-add-only-ssrf.sh:47` (and independently in `test-secrets-masking.sh:94`, an unrelated
file). `run.sh:138-142` sources `test-control-add-only-ssrf.sh` immediately ahead of the new file, so
the header's "position-locked, not merely position-preferring" claim is accurate. Both halves of the
not-arming argument check out at source.

## 2. The R2 cut-map finding, against its own header — HOLDS (re-derived)

Read `build_control_sandbox()` directly at `tests/stack/lib.sh:211-253`. It `mkdir -p`s
`$C/build/tari`, `$C/dashboard`, `$C/data/monero`, `$C/data/tari`, `$C/data/p2pool/stats`,
`$C/data/tor`, `$C/data/dashboard`, copies static repo inputs (`pithead`, `config.reference.json`,
`config.toml.template`, `docker-compose.yml`), and *defines* `seed_control_env`/`control_config` as
closures it never calls. It touches **nothing** under `data/control/`. The PR body's own narrowing
(it does not relay the cut map's "self-arm is the defect / a fresh `$C` reproduces coupling-rule-A's
RED" line, and says so explicitly) is correct: that RED cannot transfer to this file, because there is
nothing in `build_control_sandbox()` for a second call to disturb in the control spool. This is the
same false-mechanism shape already discharged for R9 (independently on record in the cut map at that
row) — good that R14 didn't repeat it.

## 3. The two falsified comments, plus a third site the lane's own instrument couldn't reach

Both disclosed fixes are true at this head:
- `test-control-add-only-ssrf.sh:27` now says the editable-allowlist file "run.sh sources next reads
  both" — confirmed true by the adjacency in `run.sh:138-142` above.
- `test-config.sh:34` now says the section "has since cut it to test-control-editable-allowlist.sh" —
  true.

Ran the mechanical sweep as prescribed (header-list diff, then `grep -F` the bare and `#522`-qualified
name across every `.sh` in `tests/stack/`): the name-grep alone finds only `test-config.sh`
(confirming the "floor, not a count" warning — it misses `test-control-add-only-ssrf.sh`, which
describes the section without repeating its exact header string).

Widening past `.sh` files to the whole repo for the same content (`describe_change`, `#522`, the line
count `run.sh` claims) turns up a **third, unfixed site**: **`docs/dev/repo-map.md:16`** still reads
*"`run.sh` itself (277 lines, down from 8,381)"*. 277 was correct immediately before this PR (verified:
`git show 546d56b6:tests/stack/run.sh | wc -l` = 277, and that line in `repo-map.md` was last touched
by `a4be5ad`, the commit immediately preceding this branch point, so it wasn't already stale). At this
head `run.sh` is 237 lines. **This PR's own cut falsifies that line and nothing in the diff touches
it.** It's outside the `.sh`-name-grep's reach by construction (a raw number in prose, no header name,
no `#522`), which is exactly the "floor" limitation the cut map already names for this class of sweep.
Per this program's own rule (a false claim in in-repo documentation earns a real fix, not a PR-body
note — `repo-map.md` is documentation, not a PR body), this is a **should-fix**, and it's a one-line
patch (`277` → `237`).

## 4. Do the moved assertions still fail on the defect they guard?

Nothing to test here that isn't already covered by the byte-identity claim: I diffed base
`run.sh:141-182` (at `546d56b6`) against the new file's body (`test-control-editable-allowlist.sh:45-86`)
myself — `diff` returns rc 0, 42 lines each side. No assertion's logic, jq expression, or expected value
changed; only the two comments and the new header are new prose. So there is no *changed* assertion to
adversarially drive in this PR — the pre-existing `roundtrip_key`/`gate_try` logic's honesty is
unchanged from what has been running in `run.sh` for months, and I did not re-probe it (declared, not
glossed: whether `roundtrip_key` would pass on a broken commit path is a pre-existing-code question,
out of scope for a pure move, and I did not drive it).

## Bonus: the suite-pair evidence itself doesn't fully reproduce (should-fix, PR-body class)

Re-derived rather than relayed, per instruction. `~/split-baselines/r14-546d56b6/{before,after}.log`:
line counts (3306/3306), header counts (275/275), verdict lines (2992 passed/0 failed both sides), and
ordered-header diff (0) all check out exactly as the PR body and the pre-run prediction state.

The **full-output diff does not**. `diff before.log after.log`:

```
556c556
< [1;34m==>[0m Wrote ingredients manifest: /tmp/tmp.t3qPVd03oO/manifest.md
---
> [1;34m==>[0m Wrote ingredients manifest: /tmp/tmp.TtUpzO52Dr/manifest.md
3101c3101
<   ✓ the generated key is a loadable ed25519 key (SHA256:U8zkXknrxU762KTMKQ5tC79slemooQmY7rAIbPTWaIM)
---
>   ✓ the generated key is a loadable ed25519 key (SHA256:92vg2cbkxqrcxpeDWwFNFntLFlQLnYcclMgsH4jzrpI)
```

That's **2 pairs (4 changed content lines, 8 raw `diff` output lines)**, not the "2 lines = 1 pair"
the PR body and the pre-run prediction both state. The second pair — the `mktemp` ingredients-manifest
path — is exactly the **already-documented comparator wart from the cut map's R9 row** ("path-normalized
residue... the #1325 keygen pair AND a `mktemp` manifest path... add `/tmp/tmp.[A-Za-z0-9]*` to the
normalizer... before R10, or the second pair reads as a real difference"). R10 never added that
normalizer, and it bit here as predicted.

The PR body's explanation — "both legs ran in one worktree, so there is no cross-worktree
path-normalisation residue to explain away" — is also a wrong mechanism, independent of the miscount:
`mktemp` produces a fresh random path on every invocation regardless of how many worktrees are
involved; running the suite twice in the *same* worktree still gets two different manifest paths. Same
worktree eliminates the worktree-*path* component of residue, not the mktemp-*directory* component.

This doesn't change the verdict — both pairs are pre-existing nondeterminism unconnected to anything
this PR's diff touches (a Docker-bundle manifest step and an SSH host-key generation test, both far
from the control-editable-allowlist section), so the suite genuinely runs unchanged. But the PR's
stated evidence undercounts what it measured, and the accompanying reasoning doesn't hold up. PR-body
class (not durable documentation) — correct in a verdict comment, not a forced push, per the R10
precedent already ratified for this exact shape.

## What I re-derived vs. relayed

**Re-derived myself, at source:** `gate_try()`/`$UUID5` definition sites and run.sh source order;
`build_control_sandbox()`'s full body and its dir list; byte-identity of the moved 42-line section
against `git show 546d56b6`; the header-diff + name-grep sweep (and its extension past `.sh` files to
find the `repo-map.md` site); `run.sh` line counts before (277) and after (237); the full suite-pair
log diff, line/header/verdict counts, and the two nondeterministic pairs within it.

**Relayed, not independently re-run:** the shellcheck/shfmt/budget-gate/CI results and the merge-tree
byte-identity claims in the PR body — per instruction, those are the controller's mechanical re-derivation
at merge time and duplicating them here buys nothing.

## Attack attempted and held

Tried to find a case where `build_control_sandbox()` being skipped could leave the new file's `$RESULTS`
path (`$C/data/control/results`) unset or wrong if sourced in a hypothetically reordered `run.sh`: the
file's own guard (`: "${C:?}" "${CTRL_LOG:?}" "${WALLET:?}" "${VALID_TARI:?}" "${UUID5:?}"`) would abort
by name before `$RESULTS` is even derived, and a missing `gate_try` would abort loud under `run.sh`'s
`set -uo pipefail` — the header's claimed failure mode reproduces on inspection; I did not need to break
sourcing to see it, the `set -u`/`command not found` mechanism is unconditional.

## Findings summary

| # | Grade | File:line | Claim |
|---|---|---|---|
| 1 | should-fix | `docs/dev/repo-map.md:16` | "run.sh itself (277 lines...)" is stale at this head (237); this PR's own cut falsifies it and it's outside the shipped sweep's reach — earns a real one-line fix, not a comment. |
| 2 | should-fix (PR-body class) | PR #1529 body, "What was RUN" table | "Full-output diff: 2 lines = 1 pair" undercounts; re-derived raw diff is 2 pairs (mktemp manifest path + #1325 keygen), a known unfixed comparator wart from the R9/R10 cut-map row. Conclusion (suite unaffected) still holds. Correct via verdict comment. |

VijitSingh97:
## Verdict: MERGE — with two notes, one of which lands as a follow-up fix

Non-author adversarial pass returned **MERGE-WITH-NOTES**, no blocker. The controller
re-derived the claims this merge turns on rather than relaying them; what was re-derived
versus relayed is stated below.

### Re-derived by me, at source

**This is a pure move.** `run.sh:141-182` at the base (`546d56b6`) versus
`test-control-editable-allowlist.sh:45-86` on the merged tree: `diff` rc 0, 42 lines each
side. A control offsetting the target slice by one line fired, so the comparison was shown
able to report a difference. The whole diff is that move plus the guarded `source` stanza
plus the two comment corrections — four files, nothing else.

**Budget gate, both halves.** The base has moved under this branch since the review
(`546d56b6` → `9f0fe178`, #1527), so this is *not* the narrow case where one run answers
both questions — the branch tree (`66c124db`) and the merged tree (`6179451a`) are
different objects and were checked separately.

- On the **merged tree**: all 85 budget rows enumerated, each ceiling compared against the
  blob's real line count. **0 over, 0 missing.** Control fired — forcing one ceiling below
  its file's length made the same loop report `OVER`.
- At the **branch head**: gate rc 0 with the text
  `file budget OK — every over-target file has a ceiling, none grew past it`. Control fired
  on the specific refusal line, not merely on rc:
  `file-budget: FAIL — tests/stack/test-config.sh is 753 lines, over its recorded ceiling of 100`.
- The budget file blob is **byte-identical across base, tip, branch head and merged tree**
  (`248aab42`), so the branch carries no stale-high row and the tip lowered none. That is
  why the two runs agree here.

**Suite pair.** 3306/3306 lines, 275/275 headers, ordered-header diff 0, `2992 passed, 0 failed`
on both sides, recorded heads `546d56b6` / `d1153159`, both legs in one worktree.

### Note 1 — a doc line this cut falsifies, fixed as a follow-up

`docs/dev/repo-map.md:16` reads *"`run.sh` itself (277 lines, down from 8,381)"*. Verified:
`run.sh` is **277** lines at the base and **237** at this head, so the figure was correct
before this PR and is falsified by it. Nothing in this diff touches that file, and the
shipped header-name sweep cannot reach it by construction — a bare number in prose carries
no header string and no issue reference.

Per this program's own rule, a false claim in durable in-repo documentation earns a real
fix rather than a note, so **this ships as its own one-line follow-up** rather than a
force-push onto a reviewed head. The mechanical sweep was re-run on the merged tree and
base-checked for attribution: this is the only such site.

### Note 2 — the body's full-output diff figure undercounts (correcting here, not force-pushing)

The body states *"full-output diff: 2 lines = 1 pair"*. Re-derived from the logs, the raw
`diff` is **2 pairs**, not 1:

- the `#1325` keygen SSH fingerprint, and
- a `mktemp` ingredients-manifest path (`/tmp/tmp.XXXXXXXX/manifest.md`)

The second is the already-documented comparator wart from the R9 row — the normalizer for
`/tmp/tmp.*` was flagged as owed before R10 and never added, and it bit here as predicted.

The body's accompanying reasoning is also wrong independent of the count: *"both legs ran in
one worktree, so there is no path-normalisation residue to explain away"*. Running in one
worktree removes the worktree-path component of residue; it does not remove the `mktemp`
component, which draws a fresh random directory on every invocation regardless. Same
worktree was still the right call — it removes a different and larger residue class.

**Neither changes the conclusion.** Both pairs are pre-existing nondeterminism in sections
this diff does not touch (a Docker-bundle manifest step and an SSH host-key test), so the
suite genuinely runs unchanged across the cut. PR-body class, corrected in this comment per
the R10 precedent for exactly this shape — not a force-push onto a reviewed head.

The reviewer's own re-derivations of `gate_try()` / `$UUID5` definition sites, the
`build_control_sandbox()` body, and the R2 cut-map narrowing. I did not repeat those; they
are recorded in the pass and I have no finding against them.

With this merge the #1105 cut map's last row lands and the split program is complete.

PR #1530, opened 2026-08-29
docs(dev): run.sh is 237 lines after the R14 cut, not 277 (#1105)

VijitSingh97:
## What is wrong

`docs/dev/repo-map.md:16` gives `run.sh`'s length as 277 lines. It is 237.

## Why it went stale, and why it is a fix rather than a note

The figure was correct until #1529 merged. Measured, not assumed: `run.sh` is
277 lines at that PR's base and 237 at its head. So this is fresh debris from
the R14 cut, not an older stale site inherited from an earlier round — the
distinction matters, because a needle derived from a diff tells you a string is
present, never that the diff put it there. I checked this one at the base
before calling it fresh.

Nothing in #1529's diff touches this file, so no instrument that PR ran could
have caught it, and a false claim in durable in-repo documentation earns a real
correction rather than a comment.

## What was RUN

- `wc -l tests/stack/run.sh` at the current tip: **237**.
- `git show <#1529 base>:tests/stack/run.sh | wc -l`: **277** — the figure was
  true before that merge.
- `make lint-docs-voice`: OK, 41 prose docs, no banned words.
- `markdownlint-cli2`: 47 files, 0 errors.
- `scripts/lint-file-budget.sh`: OK, monotonic.

## The sweep, and what it cannot do

The shipped carry-over-prose instrument keys on moved header names and on the
issue reference. A bare line count in a prose table carries neither, so this
site is outside its reach by construction — the "floor, not a count" limitation
already recorded for that instrument.

Widening the needle to any three-or-more-digit figure adjacent to a `run.sh` or
`tests/stack` reference across `docs/` and `CONTRIBUTING.md` returns this site
and no other at the current tip. That empty result is worth something only
because the pattern was shown able to match: run against the looser
`[0-9,]{3,} lines` form it finds four files that do carry such figures.

## What this does NOT do

It does not add a check that would catch the next one. A lint that verifies
quoted line counts against their files is a real idea and a real piece of work,
and bolting a speculative version onto a one-line correction is the wrong
vehicle for it. Stating the gap rather than hiding it.


VijitSingh97:
## Merging — and saying plainly what did not happen

**No non-author pass ran on this PR, and that was a decision rather than an oversight.**
The entire claim is one integer: `tests/stack/run.sh` is 237 lines, and both endpoints
are shown in the body. Spawning a reviewer pays a full cold boot to re-check a `wc -l`,
which is the kind of spend that buys nothing. Recording it here so the merge record says
so instead of reading as a review that quietly did not occur.

The proof was not waived, only the review. Verified on the tree that will actually land
(`c396d17b`):

- `run.sh` is **237** lines, and `repo-map.md` now reads 237.
- File budget: **85 rows, 0 over**. Control fired — forcing one ceiling below its file's
  length made the same enumeration report `OVER`, so the clean result was shown able to
  fail before it was believed.
- This branch sits directly on the tip, so `merge-tree --write-tree` returns the branch's
  own tree (`c396d17b` both). Stating that explicitly: **one gate run answers both what
  lands and whether the branch's evidence stands**, rather than one run being reported as
  two independent proofs.
- 15 checks pass, 2 skipped, none pending.

Issue #1367, opened 2026-08-24
Worker Inspect: a config edited underneath RigForge still reads as "changed from this dashboard"


VijitSingh97:
## The gap

#1345 landed the config-provenance line in Worker Inspect: it reports the last change RigForge **recorded**, and whether that change came from this dashboard. It does not detect the one case #1345's own issue text named as the motivation — *"a rig whose config no longer matches the last thing Pithead applied to it"* — when nothing recorded the change.

A config file edited underneath RigForge (by hand, with nothing running to stamp it) is not a recorded change. The provenance stays at whatever came before. On a rig where the dashboard applied that previous change, the line reads **"Last changed from this dashboard"** while the rig runs something else — a false reassurance, which is the one direction the rest of that feature is built to avoid.

## Why the obvious fix does not work

#1345's issue proposed the cheap version as *"a line in the config history when the rig's revision does not match our last-applied record"*. We have no revision for our last-applied record. `revision` is a truncated SHA-256 over a jq-canonical form of the writable keys (`_writable_config_hash`); reproducing that byte for byte on our side — key order, null handling, number formatting — would report drift on a rig that has none, which is worse than reporting nothing.

The rig cannot hand us the answer either. `_stamp_config_meta` writes the revision **as of the recorded change** into its marker file, and that is exactly the value that would make the comparison free. But `_api_config_meta_json` overwrites it with the live one before serving:

```
jq --arg rev "$rev" '{revision: $rev, changed_at: .changed_at, source: .source, last_change_id: .last_change_id}'
```

So the marker's stored revision never reaches the wire. The rig holds both halves of the comparison and serves only one.

## Two ways to close it

**A. Persist the last-observed revision per worker.** On each poll, compare the served `revision` against the one stored for that rig. Moved, with no new `last_change_id`? That is a change nothing recorded. Costs one small column or table and the migration for it. Detects a raw hand-edit without trusting the rig to interpret itself.

Blocked on file budget today: `dashboard/mining_dashboard/service/storage_service.py` (1603) and `dashboard/tests/service/test_storage_service.py` (1646) are both exactly at their ceilings, so this needs an extraction first — the same tax #1345 paid for `views.py` and `workerview.mjs`.

[...]

PR #1531, opened 2026-08-29
refactor(dashboard): extract the append-only telemetry series to a mixin (#1369)

VijitSingh97:
## What this does

Moves the six append-only telemetry accessors — `add/get_xvb_history`, `add/get_network_history`,
`add/get_disk_growth` — out of `storage_service.py` into `service/telemetry_store.py` as
`TelemetryStoreMixin`, which `StateManager` inherits.

Behaviour-identical by construction. This is PR 1 of two: it buys budget headroom, and the #1369
fix that spends it ships separately.

## Why a split is allowed here at all

The dashboard-cluster plan ruled `storage_service.py`'s subject file **not** split, for a stated
reason: multi-table atomic operations need the single DB-handle class, and a transaction-manager
facade is design work rather than a mechanical cut.

A mixin falls outside that reason. Same class, same instance, same `self`, same connection, same
transaction scope — and no new seam between callers and the handle. What was refused was the
facade. The narrowing was ruled and recorded on #1367, where the original ruling's own escape
hatch pointed. It is a narrowing on that ruling's reasoning, not an override of it.

The MRO makes the claim checkable rather than rhetorical:

    StateManager -> TelemetryStoreMixin -> object

One class, one handle, one lock, one transaction scope.

## Why this cut and not another

The boundary is the one the code already drew at `_TELEMETRY_TABLES`, rather than one invented for
the occasion. All six methods are the same shape, so the atomicity argument is one argument instead
of six special cases.

**Three of the five telemetry tables move. `blocks` and `worker_history` stay** — `worker_history`
is read by `get_worker_hashrate_by_config`, which is worker-config domain, and is the subject of the
#1369 fix this unblocks.

An earlier candidate — the XvB storage cluster — was dropped. It was legal on the atomicity
question, but it had been described as "unrelated and self-contained", and asking for evidence on
that adjective showed six of its ten methods touch instance state owned by the retained
`__init__`/`load` (`self._lock`, `self.state["xvb"]`, the reward and round-stat caches). None of
that blocks a mixin. It did mean the word doing the persuading was unsupported. The telemetry
series touches no retained in-memory state at all.

## Atomicity: proven, not asserted

The binding condition on any such split is that no moved method holds a transaction or cursor
across a call into a method that stays behind. Checked with an **AST** pass, not a line scan — a
wrapped call is invisible to grep, and this lane has a measured record of a line-oriented
enumerator missing 17% of wrapped sites.

**Positive control seeded and fired first:** `_recover_corrupt_db` calls `_prune_quarantined` and
`_apply_schema` inside a db `with` block, and the checker flags both. The clean result below
therefore means something.

| question | telemetry cut (taken) | XvB cluster (rejected) |
|---|---|---|
| Q1 — transaction/cursor spanning a call into a retained method | **0** | 0 |
| Q2 — shared in-memory state owned by retained `__init__`/`load` | **none** | 6 of 10 methods |

The three writers call `_table_write_ok` / `_table_write_failed` only **after** their
`with self._db_lock` / `with self._conn` blocks close, which is why Q1 is clean.

Q1 and Q2 are deliberately separate questions. Only Q1 is the test the ruling turns on; Q2 is what
the word "self-contained" claims. Conflating them is how the rejected cut nearly shipped on a
sentence nobody could support.

## Identity evidence

- **The moved lines are byte-identical.** 134 lines, sha256 `00b16ab834bd4a39` on both sides.
  Negative control: the same comparison against a one-character perturbation differs, so the check
  can fail.
- **Full suite, before and after, at the same tip** (`develop-v2` @ `90f5f53`, both legs in this
  one worktree): `2129 passed`, rc 0 both sides. The base was re-derived with `git ls-remote` before
  reusing the baseline — a borrowed baseline dies when the base moves.
- **`pytest --collect-only`: identical multiset of all 2129 test IDs**, zero only-in-before, zero
  only-in-after. Not a matching total — an element-wise diff, with a control that fires on a seeded
  one-ID perturbation. No test file moved, so "only module paths changed" reduces here to "nothing
  changed".
- **Import surface unchanged**, checked at runtime: every moved method still resolves on
  `StateManager` and `is` the mixin's own object; both retention constants still resolve on
  `storage_service` and are the same objects as the ones in `telemetry_store`. Negative control: a
  name that was never on the surface raises `AttributeError`.

## Module-attribute sweep, classified per site

Enumerated across the whole dashboard tree with a **whole-file** scanner (not line-by-line) over
the five attribute shapes, plus a separate **AST** pass for the sixth shape — `from … import name`
binds at import time and no attribute-shaped pattern can see it.

**Both instruments were fired on a seeded control before the real run**, including the wrapped
`setattr(\n  mod,\n  "name",\n)` and `patch.object(\n  mod,\n  "name",\n)` forms that a line scan
misses. The control also demonstrates the gap honestly: modattr does **not** flag the
`from … import` site, which is exactly why the AST pass exists.

Results over `dashboard/`:

| shape | sites naming a moved name |
|---|---|
| `storage_service.NAME = …`, `setattr`, `patch.object`, `patch("…storage_service.NAME")`, bare read | **0** |
| `from …storage_service import NAME` (AST) | **1** — `tests/service/test_storage_telemetry.py:23` |
| `from …storage_service import *` | 0 |

`getattr(mod, <computed>)` and `vars`/`__dict__`: 0. The `raising=False` / `create=True` sweep
returns three sites, none naming a moved name (one patches `storage_service.DB_FILE_PATH`, which is
retained).

**Classification of the one site: it needs no edit, and that is a decision rather than an
oversight.** Both retention constants are *re-exported* from `storage_service` rather than merely
moved, so `NETWORK_HISTORY_RETENTION_SEC` keeps binding from its original module. A blanket rename
here would have been exactly as wrong as no rename.

## Disclosures

- **`_TELEMETRY_TABLES` stays in `storage_service.py`.** Its only reader is `__init__`, which seeds
  a per-table health dict from it. It never iterates in a way that dispatches into a method, so the
  three tables whose writers moved need no indirection back out of the module. A comment now says
  this beside the constant.
- **`_table_write_ok` / `_table_write_failed` stay behind, and the moved writers call them.** They
  belong to the handle, not to this series — retained writers (`blocks`, `worker_history`) call them
  too. The moved writers reach them through `self`, which is what a mixin makes legal.
- **The two retention constants moved with their only consumers and are re-exported.**
  `XVB_HISTORY_RETENTION_SEC` and `NETWORK_HISTORY_RETENTION_SEC` are used by nothing but the moved
  writers, so they live beside them; they are re-exported from `storage_service` with a stated
  reason, because they were public module attributes and one test binds one of them at import time.
  `WORKER_HISTORY_RETENTION_SEC` stays, with `worker_history`.
- **This is the first mixin in `mining_dashboard/`.** There is no in-tree precedent to point at.
- **Shared paths touched:** `docs/dev/repo-map.md` and `docs/dev/file-budget.tsv`. `docs/` is shared
  rather than owned, so no override was needed; flagging it here as required.
- **`repo-map.md` carried a sentence this PR falsifies** — "`storage_service.py` keeps one subject
  module" — corrected in the same commit. A cut can falsify prose in files it never touches, and
  only a sweep finds it.

## Budget

`storage_service.py` 1603 -> 1482. **The freed headroom is 121 lines, not the 135 moved** — the
re-export and the disclosure comments cost 14. Still well clear of the +14 net the #1369 fix needs.

Ceilings only ever go down, so exactly one row was lowered, by hand:
`git diff --numstat -- docs/dev/file-budget.tsv` shows `1 1`, matching the one budgeted file whose
size this diff changed. The tsv was **not** regenerated — regenerating rewrites every row and
lowering a row this lane does not own reds every un-rebased branch in every other lane.
`telemetry_store.py` at 171 lines is under the 400 target, so it takes no row.

## What was run

| check | result |
|---|---|
| `pytest` full suite, before @ `90f5f53` | `2129 passed`, rc 0, 93.51s |
| `pytest` full suite, after (same worktree, same tip) | `2129 passed`, rc 0, 94.80s |
| `pytest --collect-only` before / after | 2129 / 2129, **element-wise multiset diff empty**, control fired |
| `make test-dashboard` | total **97.02%** (bar 80), rc 0 · `telemetry_store.py` 74 stmts, **0 missed** |
| `make test-patch-coverage` | rc 0, and its own guard reports all 85 changed files present in `coverage.xml` |
| `diff-cover --compare-branch=origin/develop-v2` | **76 lines measured, 0 missing, 100%** |
| vacuous-pass control | `--compare-branch=HEAD` -> "No lines with coverage information", exit 0 — the failure mode this gate has, demonstrated not to be what happened above |
| `make lint-py` | ruff check + format, 195 files, clean |
| `make lint-md` / `lint-docs-voice` / `lint-file-budget` | 0 errors / 41 prose docs clean / budget OK |

**The patch-coverage figure to read is 100% over 76 lines.** `make test-patch-coverage` prints 98%
over 1691 lines because the gate compares against `origin/develop`, so it spans the whole
develop -> develop-v2 divergence rather than this PR. Both pass; only the first is this diff.

## Attacks run against this change

- **Class-level introspection** would break, since the moved methods leave `StateManager.__dict__`.
  Searched for `__dict__` / `vars` / `dir` / `inspect.getmembers` / `__mro__` on the class: none.
- **Subclassing or `super()`** would be sensitive to the changed MRO. Neither exists.
- Both of the above are *absences*, so both patterns were fired against a seeded file first and
  matched every shape; `-r` was confirmed to descend into `dashboard/`.
- **The `noqa: F401` was tested rather than trusted:** removing it produces exactly two errors, both
  the intended re-exports. It suppresses nothing else.
- `ruff check --select F821,F401,F811` on both files: clean.

## Not done, deliberately

- **No behaviour change.** The #1369 fix — the exact-id lookup that answers both directions past the
  50-row window — is PR 2. A move plus a behaviour change in one diff makes both unreviewable.
- **Nothing was seen in a browser.** This lane has no browser; the claim here is module-level
  identity, not rendering.
- `make lint-sh` was not run: no shell files in this diff.

## Over-engineering pass (PR-gate review), and the one finding it produced

The diff adds no abstraction beyond the one the ruling requires: a mixin is the minimal legal
shape here, since no split at all is blocked by the budget ceiling and a facade was refused
outright. No factory, no registry, no base class, no indirection through `_TELEMETRY_TABLES`.

**Finding — the re-export is the one place a simpler option exists, and it was taken knowingly.**
Moving both constants and editing the single test import (`test_storage_telemetry.py:23`) would
drop one `noqa` and five comment lines. It was rejected because the constants were public module
attributes, so dropping them shrinks `storage_service`'s import surface in a PR whose whole claim
is that nothing observable changed — and because keeping every test file untouched is what makes
the collect-only identity result mean what it says. Recorded here so a reviewer can overrule it
rather than have to discover it.


VijitSingh97:
## Rebased onto `5654c8e` — and why the evidence above still stands

The base moved from `90f5f53` to `5654c8e` while this was in flight (#1530 landed), so this branch
was rebased. Recording it rather than letting the numbers above silently refer to a base that no
longer exists.

**The conflict was `docs/dev/repo-map.md`, and both sides were right.** #1530 corrected `run.sh` to
`237 lines`; this PR corrected the sentence claiming `storage_service.py` "keeps one subject
module", which this PR falsifies. They land on the same table row. **Resolved by keeping both
edits, not by taking a side** — verified afterwards that `237 lines` is present, the subject-module
correction is present, and no stale `277` survives.

**The before/after pair did not need re-running, and this is the reason:**

    git rev-parse 90f5f53:dashboard   ->  26f7691f893d503db36dd912d57395a1cb4090be
    git rev-parse 5654c8e:dashboard   ->  26f7691f893d503db36dd912d57395a1cb4090be

Identical subtree hash. The two incoming commits changed one line of one doc and no executable file
at all — `0` files matching `\.(py|mjs|toml|lock|cfg|ini)$`, with the filter shown to fire (`2`)
over a range that does touch Python. So the program the suite ran at `90f5f53` is byte-identical to
the program at `5654c8e`, and the `before` leg remains a valid measurement of the pre-change code.

A borrowed baseline normally dies when the base moves. This one survived for a checkable reason,
not by assumption.

**Re-run anyway at the rebased head, as insurance rather than as the argument:** `2129 passed`,
rc 0, 92.47s. Same count as both earlier legs.

`storage_service.py` is 1482 lines at the rebased head and its budget row reads 1482.

VijitSingh97:
## Verdict: MERGE — with condition (1) discharged here rather than by this PR

The non-author pass returned MERGE-WITH-NOTES and found one thing that matters. It is right, and it is not a defect in the code. **The property the ruling requires holds; the artifact the ruling requires was not delivered.** Those are different problems with different remedies, and merging turns on the first.

### What the pass found, and why it was worth asking

I asked it to check one thing I could not settle from the body: the atomicity positive control is seeded on `_recover_corrupt_db` calling `_prune_quarantined` and `_apply_schema` — and all three of those are **retained** methods. Q1 asks whether a **moved** method spans a call into a retained one. A control that never crosses the moved/retained boundary cannot exercise the classifier that answers Q1, so a classifier that always answered "not moved, skip" would pass that control and still report `Q1 = 0`.

The pass confirmed the shape mismatch by reading all three methods at the head, and then found something I had not asked about: **the checker does not exist anywhere on disk or in git history.** So the Q1/Q2 table was not reproducible by any reviewer. That is precisely the failure mode condition (1) was written to rule out — "proven rather than asserted."

It then re-derived the underlying property by direct reading and found it holds. Good work, and it labelled that as its own reading rather than as validation of the lane's tool.

### I wrote the instrument the ruling names, rather than blocking on it

Bouncing this back would have cost a lane boot to rebuild a tool in order to confirm what two parties had already concluded by other means. The condition is cheap to discharge directly, so I discharged it. An AST pass over the two files at the head, classifying `moved` as the methods defined in `TelemetryStoreMixin` and `retained` as `StateManager`'s remaining members, then reporting every `self.<retained>()` call lexically inside a `with` that acquires `_db_lock` / `_conn` / a cursor:

```
moved methods: 6  |  retained methods: 57

Q1 — moved method spanning a call into a retained one: 0

CONTROL A (retained->retained, _recover_corrupt_db):  2 hits  FIRED
    ('_recover_corrupt_db', '_apply_schema', 257)
    ('_recover_corrupt_db', '_prune_quarantined', 254)

ARMING READBACK: '_apply_schema()' now inside add_xvb_history's transaction block: True
CONTROL B (MOVED->retained, the shape Q1 asks about):  1 hits  FIRED
    ('add_xvb_history', '_apply_schema', 35)
```

**Control A is the PR's own control shape** — it fires, and it reproduces exactly the two call sites the body describes, which is independent corroboration that the body's account of its checker was honest about what it saw. **Control B is the one that was missing**: a retained call seeded inside a moved method's transaction block, with an arming readback proving the seed was actually applied before the result was read. It fires. So the `Q1 = 0` above is a zero from an instrument shown able to report non-zero *on the exact shape Q1 asks about*, which is what the PR's own evidence could not claim.

**Named limitation, because it is mine to name:** my checker keys on `with` statements. A transaction held some other way — an explicit `begin()`/`commit()` pair in a `try/finally` — would be invisible to it. That assumption is not free-floating: the pass read all six methods directly and found they all use `with self._db_lock` / `with self._conn`, so the `with`-shaped assumption is grounded in the code rather than assumed about it. Three instruments now agree — the lane's, a reading, and this — and they fail in different ways, which is the only reason agreement means anything.

### The standing condition needs one word added, and this is that ruling

The #1367 condition says the check "needs an AST pass rather than a line scan… and it needs a positive control seeded and fired first." This PR satisfied neither half in a way a reviewer could check: the tool was absent, and the control did not cross the boundary the question is about. Both are recoverable only because the cut is small.

**Amended, binding on any future mixin extraction under this ruling:** the checker ships **in the diff**, and its positive control must be seeded **across the moved/retained boundary in the direction the question asks** — a control that fires on a retained→retained pair proves the tool can see a call inside a transaction, not that it can classify. A proof a reviewer cannot re-run is an assertion with a table around it. This is an application of the existing condition plus the standing rule that evidence must be locatable, not a new proof class, and it was ruled on the default tier.

### Everything else, re-derived clean

The pass re-ran both suite legs itself from fresh installs rather than reusing the body's numbers (`2129 passed` rc 0 both legs), fired a seeded one-ID control before trusting the empty collect-only diff, and **accounted for the `+20/-141` residue hunk by hunk** — every added line is an import, a comment, or the one `class StateManager(TelemetryStoreMixin):` edit. That was the axis a byte-identity proof structurally cannot see, and it came back with no behaviour hiding in it.

It also settled the body's "134 lines" versus "135 moved" as two honest counting conventions — content lines against raw hunk deletion, differing by one trailing separator line — rather than an error, and proved rather than assumed that no name is defined in both classes (`vars(StateManager).keys() & vars(TelemetryStoreMixin).keys()` → `__doc__`, `__module__` only). The `random.random` monkeypatches were correctly excluded from the needle set with the reason given: they reach the stdlib module singleton, not a name on `storage_service`.

**Re-derived by me:** the base had moved again — my own #1528 merge — so this is ruling 8a's **wide** case and both gates ran: **85 rows, 0 over on the merged tree and on the branch head**, control fired by forcing one ceiling under its file's length. **16 checks pass, 0 failing, 0 pending, 2 skipping.** My brief told the reviewer "17 pass / 1 pending"; both halves were wrong — it was 15 pass and 1 pending at the time, and the pending job has since finished green.

### What neither of us did

Neither of us re-derived `diff-cover`'s 76-line figure or its vacuous-pass control; the pass's own `pytest --cov` run confirmed `telemetry_store.py` at 74 statements 0 missed, which is the claim that carries the weight. No `~/split-baselines/` artifact exists for this PR, so the body's wall-clock timings are unbacked by anything on disk — the pass's fresh reruns substitute for them and it said so. Nothing here was seen in a browser, which is correct: the claim is module-level identity, not rendering.

PR #1532, opened 2026-08-29
refactor(dashboard): extract the worker_config accessors to a mixin (#1369)

VijitSingh97:
## What this is

PR 2 of #1369, and a second extraction rather than the fix itself. The five `worker_config`-table
accessors move out of `storage_service.py` into `service/worker_config_store.py` as
`WorkerConfigStoreMixin`. Behaviour-identical: every moved method is byte-identical to what was
removed, and no test file needed an edit.

**Why a second extraction, and not the fix.** PR 1 (#1531) was taken to create headroom for the
#1369 fix. It did not leave any. Condition (d) — "lower only the rows the diff shrank" — lowered
`storage_service.py`'s budget row from 1603 to 1482 in the same commit that shrank the file to
1482, so the file came out of PR 1 at its ceiling with **zero** headroom, and `check_monotonic`
refuses to raise a non-exempt row back. The cut moved the cliff instead of removing it. Filed as
DECISIONS item 13; it is open and this PR does not depend on how it lands.

**I looked for a smaller option before doing this, and there is not one.** The costed fix is a
shared lookup helper, a new three-valued `get_worker_config_change(worker, change_id)`, and
`worker_config_change_known` (#530) re-expressed on top of it so the two contracts stay one
implementation. Drafted at this repo's comment density that is **+25 to +35 lines** into
`storage_service.py`. The "collapse saves 12 lines" figure from the original costing only holds if
the new public method carries no docstring, which is not a form worth shipping. Against zero
headroom there is no line-neutral version.

**Extract-then-spend pair declaration** (per DECISIONS 13): this is the extraction half. The
spending half is PR 3, the #1369 exact-id lookup. **This pair does not need DECISIONS 13 resolved
either way**, because the fix lands in `worker_config_store.py` (166 lines, no budget row, ~230
lines of headroom) rather than back in `storage_service.py`. So the row is lowered here per
condition (d) as written, 1482 -> 1358, and PR 3 is not blocked by it. That also makes this the
last extraction this surface needs — #1367's option A has room here too.

## The cut, and why this boundary

The boundary is the `worker_config` **table**: every accessor of it moves, nothing else does. That
is a seam the code already draws, which is the property the telemetry cut was chosen for rather
than an invented one. `worker_history` and its `get_worker_hashrate_by_config` reader stay behind —
that pair is hashrate telemetry keyed by config version, not the change record.

Moved (all five, byte-identical): `add_worker_config_version`, `get_worker_config_history`,
`reconcile_worker_config_status`, `worker_config_change_known`, `get_last_applied_worker_config`.

## Condition (1): the checker ships in the diff, as a test

Per the amendment on #1367 (comment 5463851663). It ships as
`dashboard/tests/service/test_mixin_atomicity.py` rather than as a script — a script in the diff is
re-runnable, a test is actually re-run, so a later cut cannot regress the property quietly. Seven
cases:

- **Q1** — no method that moved to a mixin calls one that stayed behind while `_db_lock` or a
  transaction is held. Result over both mixins: **0 spanning calls.**
- **Q2** — no moved method touches retained in-memory state. Result: **0.**
- **Positive control, seeded ACROSS the moved/retained boundary** in the direction Q1 asks: a moved
  method calling the retained `_db_error` *inside* its `with self._db_lock:` block, with an arming
  readback asserting the seed is present and positioned inside the guard before the result is read.
  The checker flags it.
- **Negative control** — the same call one line *after* the guard closes is not flagged. This is
  the shape every real moved method has, so a checker that could not tell the two apart would
  report the real split as unsafe and its clean result would mean nothing.
- **Corroboration on real source** — the walk finds `_recover_corrupt_db`'s two guarded calls to
  `_prune_quarantined` and `_apply_schema`. Retained-to-retained, so not findings; a walk that came
  back empty here would be one that never matched anything on this file at all.
- **A non-vacuity case** — the two sets are non-empty and disjoint, and the control's callee is a
  real retained method rather than a name that would match nothing.

**The instrument was shown able to FAIL, on the file this PR ships.** Real violations seeded into
`worker_config_store.py` — `self._db_error(...)` inside the `with` in `add_worker_config_version`
for Q1, and `self.table_health["seeded"] = True` for Q2 — reddened **both** checks, each naming the
exact method and the exact callee/attribute, with an arming readback confirming both seeds were
present before the result was read. The file was then restored and verified byte-identical (sha256
`7b0b4ac3435eb36b`). This was re-run after the simplification below, rather than reused: a firing
proof is evidence about the instrument it ran on, and that instrument changed.

## /ponytail-review — over-engineering pass, and what it changed

Run on this diff before opening, not after. One finding worth acting on, applied here:

- **`test_mixin_atomicity.py` hardcoded the Q2 needle set** as a nine-name tuple mirroring
  `StateManager.__init__`. That is a copy of another file's contents that nothing keeps in sync: the
  day someone adds a tenth in-memory attribute, Q2 silently stops covering it and goes on reporting
  clean — the exact false-pass shape this file exists to prevent. **Replaced by deriving the set**
  from every `self.X = ...` assignment in `StateManager`, minus the four names that ARE the shared
  handle (`_conn`, `_db_lock`, `logger`, `db_path`). Less code, and it cannot drift. The
  non-vacuity case now also asserts the derived set is real, so Q2 cannot pass by checking against
  an empty needle list.
- **Considered and kept:** the seven cases are not padding — each blocks a specific false pass, and
  the two I re-examined hardest (the non-vacuity case and the retained-to-retained corroboration)
  both earn their place, the first because this lane has a documented history of empty enumerations
  reading as clean results, the second because it is the only case that exercises the walk against
  real source rather than a synthetic string.
- **Considered and NOT taken:** dropping Q2 entirely as beyond the ruling's Q1. It is ten lines, it
  is half of the approved proof class, and having just made it drift-proof it costs nothing to keep.

This exists because PR 1's control did not do this job. That control seeded
`_recover_corrupt_db` calling `_prune_quarantined`/`_apply_schema` — all three **retained**. Q1 asks
about a **moved** method spanning into a retained one, so the control never crossed the boundary,
and a classifier that always answered "not moved, skip" would have passed it and still printed
`Q1 = 0`. It fired, its readback was honest, and it was worth nothing for the question asked.

## Evidence — run by me, at base `49a6d342`

- **Byte-identity.** All five moved methods identical to the base blob, combined sha256
  `527272a55caa4d42`. Negative control: a one-character perturbation changes the digest.
- **Suite, both legs.** Base `2129 passed`, rc 0. Head `2136 passed`, rc 0.
- **`--collect-only`, element-wise multiset diff.** 2129 -> 2136 IDs; **only** the 7 new atomicity
  tests added, **0 removed, 0 renamed**. Positive control: a seeded 1-ID perturbation makes the
  differ report 2 differing entries.
- **Import surface, live.** MRO is `[StateManager, TelemetryStoreMixin, WorkerConfigStoreMixin,
  object]`; all five methods are the same function object via `StateManager` and via the mixin; no
  name is defined in both classes; `_RECONCILE_TERMINAL` is the same object through both paths.
  Negative control: an absent name raises `AttributeError`.
- **Module-attribute sweep, both halves, controls fired first.** The whole-file enumerator (five
  attribute shapes, `re.S`, so wrapped multi-line calls are visible) and the AST `ImportFrom` half,
  over `dashboard/`, for the five method names plus `_RECONCILE_TERMINAL` and
  `WorkerConfigStoreMixin`. Both halves fired on a seeded control including the wrapped `setattr`
  and the `from ... import` shape. **Real tree: one site** — `tests/service/test_storage_worker.py:378`
  reads `storage_service._RECONCILE_TERMINAL`. **Classified: needs no edit**, because the constant is
  re-exported. Zero `ImportFrom` sites binding a moved name, zero star imports.
- **Gates.** `scripts/lint-file-budget.sh`: `file budget OK`. `ruff check` and `ruff format --check`
  clean. `make lint-docs-voice`: OK over 41 prose docs.
- **Coverage.** Total `6853` statements / `204` missed = **97%**, above the 80% bar.
  `worker_config_store.py`: 65 statements, 5 missed, 92%.
- **Patch coverage: 92% over 67 lines** vs `origin/develop-v2`, above the 90% bar. **Vacuous-pass
  control demonstrated**: the same command with `--compare-branch=HEAD` reports "No lines with
  coverage information in this diff", so the passing run above is one that actually measured lines.
- **The 5 uncovered lines are named, and I am NOT claiming coverage was unchanged.** They are
  `worker_config_store.py:57` (the `if not self._conn: return` guard), `:72-73` (the
  `except (sqlite3.Error, TypeError, ValueError)` arm and its `_db_error` call) and `:94-95` (the
  `except (TypeError, ValueError)` arm decoding `changes`, and its `{}` fallback) — error paths
  inside code that moved verbatim, which `diff-cover` counts as new because the FILE is new.
  **I did not capture a coverage baseline before the move, so "no coverage was lost" is a claim I
  have not measured and am not making.** What is measured: the moved text is byte-identical and the
  suite is green at 2136 both as a count and as a pass.

## Disclosures

- **`_RECONCILE_TERMINAL` is re-exported, not merely moved.** It was a module attribute of
  `storage_service` before the split and `test_storage_worker.py:378` binds it from there, so the
  import surface must not change. Same treatment, and same reason, as PR 1's retention constants.
  The simpler alternative — move it and edit the one test import — is real, and I did not take it
  for the same reason as last time: leaving every test file untouched is what makes the
  collect-only identity claim mean what it says. **A reviewer may legitimately overrule this.**
- **`docs/` is shared** (`_shared.paths`): `docs/dev/repo-map.md` said the subject module "is now
  two". This cut makes it three, so the sentence was false the moment this diff landed and is
  corrected here. `docs/dev/file-budget.tsv` is the one-row ratchet edit, hand-curated, not
  regenerated — the diff shows exactly one changed row, matching the one file whose size changed.
- **`make lint-sh` not run** — no shell files in the diff.
- **Nothing seen on screen** — this lane has no browser. Behaviour is unchanged by construction, so
  there is nothing visual to see, but I am stating it rather than implying coverage I do not have.

Part of #1369. Does not close it — the fix is PR 3.


VijitSingh97:
## Verdict: MERGE

Controller non-author review, opus tier. Ground: the extraction is a mixin, so it introduces no new
seam between callers and the single DB handle — which is precisely what #1105's "do not split
`storage_service.py`" ruling refused. The non-author pass returned MERGE-WITH-NOTES; I re-derived
the two claims that verdict actually turns on, and both hold.

### Proven by me, at source

- **Ruling 8a, BROAD case, stated explicitly — the base moved under this PR while it waited.**
  #1533 merged first (as `92030295`), so `merge-base(develop-v2, head)` is `49a6d342` and the tip is
  no longer it. The merged tree and the branch tree are therefore different objects
  (`a52325e57b3c4adbf5c691fda48326174214e4ab` versus `d2fbaeab2e2929c43ea78ff6fd01825815cd0993`),
  and one run does **not** answer both questions. I ran both.
- **File-budget gate: 85 rows, 0 failures on the merged tree AND 0 on the branch head**, re-derived
  in Python over `git show <ref>:<path>`, with a control fired on the merged tree. `storage_service.py`
  lands **1358 / 1358 — legal, and with zero headroom**, which is worth saying out loud: the next
  line added to that file reds the gate. `worker_config_store.py` is 166 lines, under the 400
  target, so it correctly carries no row.
- **The branch did not need a rebase, and that is a checked fact rather than a hopeful one.**
  Ruling 8's hazard is a tip that LOWERS a row an un-rebased branch still carries high. #1533's tsv
  diff against the old base is **empty** — it lowered nothing — and the two changesets are disjoint:
  five files each, zero overlap, enumerated with `git diff --name-only` on both sides rather than
  recalled, and with `comm` shown able to report an overlap when given one.
- **Ruling 9 satisfied: the tsv diff is exactly one row**, and it is the one budgeted file whose
  size moved. No other lane's ceiling was rewritten by `--generate`.
- **The move-breaks-monkeypatching hazard does not bite here.** This is the failure that would be
  silent: a test patches a module attribute that the moved code no longer reads, so it passes for
  the wrong reason. I swept the whole tree. Exactly one `monkeypatch.setattr` names the
  `storage_service` module — `RAFFLE_WINS_MAX_ROWS` — and that constant did **not** move; its
  readers are still in `storage_service.py`. The one constant that did move, `_RECONCILE_TERMINAL`,
  is only ever *read* in tests (`test_storage_worker.py:378` asserts equality), never patched.
  Controls fired: a name I know moved resolves to `worker_config_store.py`, a name I know stayed
  resolves to `storage_service.py`.
- **F1 is genuinely non-blocking — I re-derived the claim it rests on.** The pass found that
  `_retained_state()` cannot see tuple/multi-target `self.` assignment or `setattr(self, ...)`, and
  argued it is latent because no such shape exists in `StateManager` today. I parsed the head's
  `StateManager` with `ast` myself: **23 plain `self.X = ...` assignments, and zero of all three
  blind shapes** (tuple/list target, `setattr(self, …)`, annotated `self.x: T = …`). My probe's
  control fired on a synthetic class carrying all three. So `Q2 = 0` is honest today; the blind spot
  is an instrument defect, not a live false clean.
- **F3 is fresh, not pre-existing** — I checked it at the base, which is the only thing that
  separates the two. `tests/integration/rigforge-apply-settle.sh:13` says
  `storage_service.py:reconcile_worker_config_status`. That was **true at `develop-v2`** and becomes
  false at this head, and the PR does not touch that file. It is a real instance of a cut
  falsifying prose in a file it never touched.

The byte-identity digest of the five moved methods, the `2129 -> 2136` suite pair, the coverage
figures, and the pass's own seeded Q1/Q2 violations. The pass re-derived all of those itself with
fired controls, including two real seeded violations in the shipped module rather than only the
synthetic fixtures — which is more than the brief asked for. Nothing in this merge decision turns on
a number I did not check.

### The two notes, and where they go

Neither blocks. **F1** (checker blind spot) should be closed before a third mixin extraction reuses
this file as its template, which `docs/dev/repo-map.md` says is likely. **F3** is one comment line,
in a `.sh` file, and by the standing rule that a false claim in durable in-repo documentation earns
a real discharge rather than a note, it gets fixed rather than recorded. Both are routed to the
follow-up PR on this issue rather than held against this one.

Issue #1246, opened 2026-08-22
Optional .env keys read with [ -n ] under set -u crash on legitimately-absent keys instead of degrading


VijitSingh97:
Surfaced incidentally during #1091/#1093 work: a partial standalone run of the stack suite hit

```
pithead: line 7666: DASHBOARD_AUTH_HASH_B64: unbound variable
```

in the rotate-dashboard-onion black-box test. The full suite passes under its normal invocation (its fixtures always render the key), so this is not a current-suite failure — it is a latent robustness defect in the code pattern:

`generate_caddyfile`'s auth block tests `[ -n "$DASHBOARD_AUTH_HASH_B64" ]` under `set -u`. That guards against an EMPTY value but crashes on an UNSET one. Any `.env` that legitimately lacks the key — one rendered by a version predating it, a hand-trimmed file, or a test context that sources a partial env — turns a graceful "no auth block" degradation into a fatal unbound-variable abort of whatever command sourced it (render/apply/doctor).

Fix shape: `${DASHBOARD_AUTH_HASH_B64:-}` at the read sites (grep for sibling occurrences of the same pattern on other optional .env keys while there — the same crash hides behind every `[ -n "$OPTIONAL" ]` under set -u), with a tier-1 test sourcing a pre-key `.env` and proving render degrades instead of dying. The mutation: respell the default back off → the test reproduces the exact unbound-variable message.

VijitSingh97:
Closed by #1533, merged to `develop-v2` as `92030295`. Closing by hand: a closing keyword only
fires into the default branch, and everything lands on `develop-v2`.

Both halves of this issue are discharged, not just the crash:

- **The fix.** `generate_caddyfile`'s guard becomes `${DASHBOARD_AUTH_HASH_B64:-}`, so a `.env` that
  legitimately lacks the key renders without the auth block instead of aborting whatever sourced it.
- **The sweep this issue asked for.** The inherited upper bound was 103 bare `.env`-key reads;
  restricted to the actual defect pattern the corpus is 23 sites across 7 files, re-derived in
  Python rather than through a shell pattern, with the breakdown reconciling exactly
  (8 + 9 + 1 + 5). 17 are ruled out by reading them — `env_get` returns empty rather than unset, so
  `set -u` cannot fire on the assign-then-read pairs that look precisely like this bug. 1 was the
  defect. The remaining 5 have a different origin and are each argued safe, so **no follow-up issue
[...]

PR #1533, opened 2026-08-29
fix(cli): degrade instead of aborting when the optional dashboard auth key is absent (#1246)

VijitSingh97:
## What this changes

One read site. `generate_caddyfile` renders its `basic_auth` block only when a dashboard password is
configured, and guarded that with a bare `[ -n "$DASHBOARD_AUTH_HASH_B64" ]`. Under the script's
`set -u` that guards an EMPTY value and aborts on an UNSET one, so any `.env` that legitimately
lacks the key turned "render no auth block" into a fatal `DASHBOARD_AUTH_HASH_B64: unbound variable`
in whatever command sourced it. It becomes `${DASHBOARD_AUTH_HASH_B64:-}`.

First seen from the rotate-dashboard-onion path, which calls `generate_caddyfile` at
`32-onion-provisioning.sh:209`.

## Why the diff is one line when the issue asks for a sweep

The issue says to grep for sibling occurrences on other optional `.env` keys. That sweep was done,
and the narrowing is the evidence that it was — not a reason to skip it.

- The inherited figure was **103 bare `.env`-key reads across `lib/pithead/`**, carried explicitly as
  an upper bound and unclassified.
- Restricted to the actual defect pattern — `[ -n "$VAR" ]` with no default — the corpus is **23
  sites across 7 files**. Re-derived in Python over `git show origin/develop-v2:<path>` rather than
  through a shell pattern, and the breakdown reconciles exactly: **8** in
  `28-parse-and-validate-config.sh` (:93, :123, :198, :210, :256, :261, :358, :366), **9** in
  `29-preserved-state-and-dirs.sh` (:20-:28), **1** in `35-caddyfile-and-auth-helpers.sh` (:33, the
  defect), and **5** elsewhere — 8 + 9 + 1 + 5 = 23. Positive control fired: the known defect site
  appears in the output.
- **17 of the 23 are ruled out by READING them, not by pattern.** All eight in
  `28-parse-and-validate-config.sh` and all nine in `29-preserved-state-and-dirs.sh` assign the
  variable in the same function immediately above its own read. `load_preserved_state` does
  `PROXY_AUTH_TOKEN=$(env_get PROXY_AUTH_TOKEN)` and then
  `[ -n "$PROXY_AUTH_TOKEN" ] || PROXY_AUTH_TOKEN=$(openssl rand -hex 12)` — and `env_get` returns
  empty rather than unset, so `set -u` cannot fire. That block looks precisely like the bug; a
  pattern-based sweep would have called all nine defects.
- **1 is the real defect** (`35-`:33), fixed here.
- **5 have a different origin — not `.env`-rendered keys — and all five are argued SAFE. No
  follow-up issue is owed.** `PITHEAD_VERSION` (`18-`:184, `19-`:86) is unconditionally exported at
  `19-`:68 (`export PITHEAD_VERSION=""`) ahead of both reads; `TELEGRAM_BOT_TOKEN` (`24-`:202) is
  assigned at `24-`:151; `TOR_DATA_DIR` (`32-`:192) is assigned by `parse_and_validate_config` at
  `28-`:314, one call frame away rather than lexically adjacent; `IN_DASH_PASS` (`24-`:188) is
  assigned by a `read -r -s -p`, which assigns the empty string even when the read fails.
  **Split by evidence:** I re-derived the corpus and read the four assignment sites above myself;
  the call-ORDERING half — that each assignment always precedes its read — and the `IN_DASH_PASS`
  `read` behaviour come from the non-author pass and are RELAYED, not re-derived here.

**Correction to an earlier version of this body**, kept visible rather than silently fixed: it said
7 sites in `28-` and 6 unclassified. The true split is 8 and 5. The total of 23 was right either
way, so nothing was hidden — but the breakdown IS the evidence that the sweep happened, and it
undercounted one bucket while overcounting the other. Found by the non-author pass and re-derived
here independently before changing it.

## Three bare reads kept deliberately (ruled, not overlooked)

`DASHBOARD_SECURE` (:61/:249), `HOST_IP` (:208/:213) and `NETWORK_PREFIX` (:169/:285/:288) keep
their bare reads. They are REQUIRED keys, not optional ones, and `${X:-}` on a required key converts
a loud abort into a silent misconfiguration. The sharp case is `DASHBOARD_SECURE`:
`[ "$DASHBOARD_SECURE" == "true" ]` fails CLOSED to the plaintext branch, so a default would turn a
missing security-mode flag into a silent HTTPS-to-HTTP downgrade — a worse defect than the one this
PR fixes. `HOST_IP`/`NETWORK_PREFIX` would emit a malformed Caddyfile (`redir https://{uri} 308`,
`bind .1`) rather than crash.

Also not defaulted: **:35 and :36** (`DASHBOARD_AUTH_HASH_B64` again, and `DASHBOARD_AUTH_USER`).
Both sit INSIDE the guarded branch, so they are reachable only when the guard passed and the value
is provably set. `${X:-}` was available at both and was rejected — a default where the value cannot
be unset dilutes what `:-` signals in this file. `33-render-env.sh` emits `DASHBOARD_AUTH_USER`
(:479) and `DASHBOARD_AUTH_HASH_B64` (:480) adjacently and unconditionally, so no realistic partial
`.env` splits the pair.

## Evidence — what was RUN

**A 2x2 repro pair against the real slice, base vs head, in one worktree:**

| | key ABSENT | key PRESENT |
|---|---|---|
| base (`origin/develop-v2`) | **aborts**: `DASHBOARD_AUTH_HASH_B64: unbound variable`, line 33 | rc 0, `basic_auth` x1 |
| head | rc 0, 15-line Caddyfile, `basic_auth` x0 | rc 0, `basic_auth` x1 |

Only the absent cell moves. The present-key column is the control that the fix is not just breaking
the auth path.

**The test file, run as a controlled pair (identical file, two artifacts):** 8/8 green against the
fixed artifact, **5 passed / 3 failed against the pre-fix artifact**. It is load-bearing, and shown
to be.

**The test carries the issue's own mutation as a live control.** It copies `pithead`, respells the
default back off, asserts the copy REALLY DIFFERS (a `sed` that matches nothing writes an unchanged
file and reads as a pass), then asserts that copy aborts naming the key and still renders when the
key is present — so the mutation is shown to be targeted rather than a broken copy.

**It classifies three states, not two.** An absent Caddyfile trivially satisfies "contains no
`basic_auth`". That is measured, not theorised: an earlier two-state form of this file scored that
assertion GREEN against the pre-fix artifact, because the render had died and left no file. The
classifier separates `missing` from `no-auth`, and the red leg now reports `missing`.

**Full `tests/stack/run.sh` at the rebased head: rc file 0, `pithead tests: 3028 passed, 0 failed`.**
My 8 cases are present in that log with a tick, the #1400 zero-assertion guard did not fire, and the
failure-glyph grep was shown able to FAIL (a copy of the same log with one glyph injected returns 1,
the real log 0) — so "zero failures" is a fired instrument, not an absence I assumed.

**Gates, at the BRANCH head after the rebase (ruling 8a):** `lint-pithead-parity` rc 0,
`lint-file-budget` rc 0, `lint-docs-voice` rc 0, `bash -n tests/stack/run.sh` clean.

**Parity was proven with a negative control, not just a green:** `--check` was RED before the
rebuild and GREEN after, so the artifact is genuinely the product of the edited source rather than
a stale copy that happens to match.

## Base moved under this PR

`develop-v2` advanced `3214a5c9 -> 49a6d342` (#1531, the telemetry-store mixin) mid-session. The
budget gate first FAILED naming `dashboard/mining_dashboard/service/storage_service.py`, a file this
diff never touches — the moved-base signature. Re-derived rather than assumed: #1531 SHRANK that
file 1603 -> 1482 and lowered its ceiling to match, so the gate was reading my untouched tree against
the newer base. Rebased onto `49a6d342`; both gates green at the branch head. Trees necessarily
differ across a moved base, so the rebase was checked by PATCH, not by tree.

## Disclosed

- `tests/stack/run.sh` lost its executable bit to a `head`/`tail` rewrite and it was restored before
  push (base `100755`, head `100755`). CI invokes it as `bash tests/stack/run.sh`, so CI would have
  stayed green while direct invocation broke.
- The over-engineering pass was run BY HAND (the local gate keys off the wrong branch from this
  lane's pinned cwd and is a one-shot prompt, not a proof). There is no smaller change available:
  the diff is one character-level edit plus its test and its catalog row. A larger change — sweeping
  the other 22 sites — was available and is rejected above, with call sites.
- `tests/stack/` is not this lane's path; the new sibling file and the 3-line `run.sh` stanza are a
  per-file grant for #1246 only, recorded in `roles/appliance.paths`. No tests-lane branch existed
  on the remote at edit time (`git ls-remote`), so there was nothing to serialize against.
- **Not run:** the KVM appliance battery. This change is `lib/pithead/` logic with tier-1 coverage
  and touches nothing under `os/`.

## Follow-up push: the test file now carries `== section ==` headers

CI's `Test-inventory drift check` refused the first push: the new test file used `# --- comment ---`
separators instead of the `echo "== ... =="` headers `tests/inventory.sh` counts sections by. Two
separators became headers; nothing else changed.

**Why no local gate caught it, which is the transferable half.** `tests/inventory.sh` is a separate
CI step and is not part of `tests/stack/run.sh`. The suite genuinely passed (3028/0, read off the rc
file) and the zero-assertion guard correctly stayed silent, because the 8 assertions really do run.
Two local guards passed and neither was the guard that could catch this — the same shape this branch
promoted to `common/verification.md`, landing on the branch that promoted it. **When you add a new
test FILE rather than a new case, the file-shape gates are `tests/inventory.sh` and the budget gate,
and neither runs inside `run.sh`.**

**Armed both ways rather than asserted.** Negative control: with the file exactly as it was at HEAD,
`tests/inventory.sh` exits 1 and names this file. With the headers, it exits 0 and counts the file
at 2 sections. **And the suite figure above is not stale:** the domain file's assertion multiset,
captured before and after the header lines and sorted, is byte-identical — 8 passed, 0 failed, same
8 names — so the added lines are output-only and cannot have moved a result. `shellcheck
--severity=warning` and `shfmt -i 4 -d` are both clean on the changed file.


VijitSingh97:
## Verdict: MERGE

Controller non-author review, opus tier. Ground: the fix is one line, its safety argument survives
independent re-derivation, and the sweep that justifies the narrowing is evidenced site-by-site with
a corrected breakdown. The non-author pass returned MERGE-WITH-NOTES with no blocker.

### Proven by me, at source

- **Ruling 8a, NARROW case, stated explicitly.** `merge-base(develop-v2, head)` equals the tip
  `49a6d342`, and `merge-tree --write-tree` returns `eaa01ac858e226f7f9dc3ed15f2c30d6b751482b` —
  byte-identical to the branch's own tree. The merged tree and the branch head are the same object,
  so one gate run answers both questions.
- **File-budget gate: 85 rows, 0 failures**, re-derived in Python over `git show <ref>:<path>` at
  the tip and at the head. Two controls fired: a ceiling forced one line under its file's length
  (`OVER`), and a row raised above its base value (`MONOTONIC`). This PR touches no budget row.
- **The slice edit and the generated artifact are the same change.** `lib/pithead/35-caddyfile-and-auth-helpers.sh`
  and `pithead` each carry exactly one hunk, and they are the same hunk — the artifact was
  regenerated, not hand-edited into agreement.
- **The `set -u` reasoning holds both ways.** `${X:-}` makes an unset key render the empty string,
  so the guard evaluates false and the auth block is omitted instead of aborting. Lines :35/:36 keep
  their bare reads legitimately: they are reachable only through a guard that has already proven the
  value non-empty, hence set.
- **The body's corrected arithmetic is present and visible** — 8 + 9 + 1 + 5 = 23, with the earlier
  7-and-6 split retracted in place rather than silently fixed. The reviewer's one should-fix is
  discharged. I re-derived the same corpus independently in Python at w39: 23 sites, 7 files.

The full-suite figure (`3028 passed, 0 failed`), the 2x2 base-vs-head repro pair, and the
8/8-versus-5/3 controlled pair against the pre-fix artifact. The non-author pass could not reproduce
the suite figure from disk either, and said so. Nothing in the merge decision turns on those numbers:
it turns on the one-line diff, the artifact parity, and the gate — all of which I checked myself.

### One correction to the review record, so it does not become durable

The pass reported a NEW ugrep defect — that `[A-Z_]+"` "fails to match at all". **That does not
reproduce.** ugrep 7.8.4 returns a hit, GNU grep 3.11 agrees, and the positive controls fired. The
real cause is mundane: `DASHBOARD_AUTH_HASH_B64` ends in a digit, so `[A-Z_]+` cannot reach the
closing quote. The pass's fix and its corpus were both right; only the explanation was wrong. It has
not been promoted to any shared traps file, and it should not be — a false trap is durable and it
teaches people to distrust a tool that works.

PR #1534, opened 2026-08-29
fix(cli): serialise rotate-secrets behind the mutation lock (#1482)

VijitSingh97:
## What this changes

`rotate_secrets` takes the mutation lock. One acquire, one release, in `lib/pithead/32-onion-provisioning.sh`:

- `mutation_lock_acquire rotate-secrets` **after** the verb's own confirmation prompt and **before** its first write to disk.
- `mutation_lock_release` at the end of the function.

That is the second of the residual verbs tracked on #1482. `pithead` is regenerated by `scripts/build-pithead.sh` — never hand-edited — and `lint-pithead-parity` proves the artifact is exactly what the 52 slices build.

## Why the branch name says #1059

The branch is `fix/1059-backup-guard` and the work is #1482's `rotate_secrets` slice. Stating it here so the merge record is not ambiguous.

I took #1059 off the queue, read its own thread, and found its remaining half is not a backup fix at all. Its Defect B landed in #1381; its blast-radius half (`backup_require_items`, `backup_diagnose_items`, the absolute-path join) landed in #1248. What is left of #1059 is Defect A — no mutual exclusion — which is #1342, whose residual verbs are tracked on #1482. So the branch was cut before the reconciliation and the work is #1482's. #1059 stays open until the residual verbs are wired.

## Where the acquire goes, and why that is the whole question

Both bounds are deliberate, and both are asserted.

**After the confirmation.** The prompt blocks on a keystroke. A window opened above it would park every other verb for as long as nobody answers — a race converted into a hang, which is harder to read out of a journal because a hang has no message. Everything above the acquire is read-only: the rotation preview is computed from the parsed config rather than from the rendered `.env`.

**Before the first write.** The timeout message tells the operator nothing was changed, and that has to be literally true. **The subtlety worth reading twice: this verb's first write is the pre-rotation safety copies, not the credential rewrite.** An acquire placed after the copies still passes every credential assertion — the RPC password, the proxy token, the container recreate are all untouched — while having already put two files on the operator's disk during a window it did not hold. So the assertion that actually pins the lower bound is the absence of `config.json.bak-*` / `.env.bak-*`, not the credentials.

## Tests — tier 1, `tests/stack/test-appliance-rotate-secrets-lock.sh` (new)

A new sibling file plus a three-line registration stanza in `tests/stack/run.sh`. New file rather than an addition to `test-secrets.sh` because that file sits at 401 against a 401 ceiling and fails both ways; the sibling shape is the one #1514/#1528 established for the other halves of #1482.

15 assertions in three groups:

- **5 uncontended** — the positive control that the fixture arms: on a free lock the verb really does rewrite the RPC password in `config.json`, rewrite the token in `.env`, write both safety copies, and recreate the containers. Without this, "the token is unchanged" is equally consistent with a fixture that never armed, and the two read identically.
- **7 contended** — on a held machine: rc 75 (`PITHEAD_EX_LOCK_TIMEOUT`, not `error()`'s 1), the message names the verb to re-run, and the stored RPC password, the rendered token, the safety copies, the containers and the retry marker are all exactly as found.
- **3 declined** — the verb given `n` on a held machine comes back on the answer, never says `Timed out`, and has written nothing.

The file builds its own sandbox rather than reusing `build_val_sandbox`'s shared `$V`, because the contended cases assert on the ABSENCE of `config.json.bak-*` and that would otherwise depend on another domain file's cleanup. The seeding `apply` always runs against the free lock file: `apply` takes the window itself, so seeding through a held lock would block rather than set the fixture up.

## The mutation battery — five mutants, each proven applied, each judged by which named rows died

Every mutant was proven to have applied with `git diff --cached --numstat` on the **staged** slice before it ran, and each was judged by its expected rows rather than by rc — an rc is one bit over several independent doors and cannot say which one fired. `pithead` was rebuilt for each and restored afterwards, verified by `cmp` against a pristine copy plus a parity run.

| mutant | what it changes | rows killed |
|---|---|---|
| M1 | no acquire at all | 6 — every contended row except the retry marker |
| M2 | acquire above the confirmation | 2 — both human-wait rows, nothing else |
| M3 | acquire below the safety copies | **1** — "writes no pre-rotation safety copy", alone |
| M4 | acquire below the retry-marker write | 4 |
| M5 | safety copies moved above the confirmation, **acquire not touched** | 2 — both safety-copy rows |

M3 is the one that matters most: it kills exactly one row and no other, which is the direct evidence that the safety-copy assertion — not the credential assertions — is what pins the acquire's lower bound.

**M5 exists because of a review finding on this branch, and it is worth recording how.** I first deleted the third-group safety-copy assertion on the grounds that no mutant killed it. That reasoning was wrong. All four of M1-M4 move the same variable — where the acquire sits — and that row is on a different axis: what the first mutation *is*. Its survival was guaranteed by construction and carried no information either way. **A surviving assertion means the battery did not vary what it asserts, not that it asserts nothing**, and deleting it would have converted a coverage gap into an invisible one. So the row is restored and M5 varies the axis it lives on. M5 is also the most realistic defect shape in the set: someone tidying `rotate_secrets` moves the safety copies up beside the preview, and every credential assertion stays green.

Accounting: 5 rows are the fixture's positive control, 10 are killed by at least one mutant, 0 unaccounted for.

## What was run

- **⚠ THE SUITE FIGURE BELOW IS FROM THE PRE-REBASE HEAD, and the re-run is IN FLIGHT.** After it
  was taken, `develop-v2` moved twice (#1533 then #1532) and this branch was rebased onto
  `abb23ee`. A rebase invalidates the whole proof set, not just the suite, so the figure is stated
  as OWED rather than carried over: the re-run's log is `/home/vijit/appliance-suite-logs/1482-rebased-20260829T133035.log` with its `.rc` beside it. **Do not
  merge on the pre-rebase figure.** What HAS been re-run at the rebased head: `lint-pithead-parity`
  (passes, and a fresh rebuild reproduces the artifact byte-for-byte, so the rebase carried #1533's
  slice change through correctly), `lint-file-budget` at the BRANCH head per ruling 8a,
  docs-voice, operator-strings, topology, markdown, `tests/inventory.sh` and `shfmt` — all 0.
- **Pre-rebase head: rc FILE 0, `pithead tests: 3035 passed, 0 failed`.** Read
  off the rc file rather than the process ending. All 15 of this file's assertions confirmed present
  in that run BY NAME (15/15, 0 missing), with the checker itself shown able to report absence on a
  name that is not there. The failure-glyph count is 0 against a fired negative control — one glyph
  injected into a copy of the same log scores 1. The totals reconcile: 3035 minus this file's 15 is
  3020, the base figure #1533's run implies. A case-insensitive `FAIL` grep is useless on this log,
  since the suite names tests after the condition they assert.
- `make lint-sh` rc 0, under the lint lock (announced, granted and cleared by the controller), with `~/.local/bin/shellcheck` 0.11.0 and the Makefile's exact invocation. The new test file was confirmed present in the expanded argument list rather than assumed covered by the glob.
- `lint-pithead-parity`: **RED before the rebuild, GREEN after** — so the artifact is genuinely the product of the edited slice, not a stale copy that happens to match.
- `lint-file-budget` 0, `lint-docs-voice` 0, `lint-operator-strings` 0, `lint-topology` 0, `lint-md` 0, `shfmt -i 4 -d` clean.
- Topology detector clean on this body and on every changed doc.

## What was NOT run, and what is not claimed

- **No bench run.** This is not hardware evidence and nothing here needs it: every path is driven by the stubbed tier-1 fixture.
- **The `firstboot_wizard` installer path is still unlocked** — the other residual verb on #1482. `lib/pithead/12-firstboot-wizard.sh` sits at 556 against a 556 ceiling, closed both ways, so that slice needs a planned shrink or split and is deliberately not folded in here.
- **#1482's second residual — the in-place upgrade fallback overwriting the install dir before it re-invokes `pithead` — is untouched** and stays disclosed on the issue.
- The `exit 1` path when the container recreate fails takes no explicit release. That is correct rather than overlooked: the lock lives on the descriptor and the kernel drops it when the process exits, which is the same reason `error()` is safe inside a window. Not separately asserted here.

## Docs

- `docs/operations.md`, "Two commands at once": one sentence naming `rotate-secrets` alongside `backup` and `os-update` in the existing list of when the window opens relative to a prompt.
- `docs/dev/testing-strategy.md`, Section G: a row covering the placement rule for all three lock files, the copies-before-credentials subtlety, and the uncontended-control requirement.

## Lane

Files: `lib/pithead/32-onion-provisioning.sh`, `pithead` (build output), `tests/stack/test-appliance-rotate-secrets-lock.sh`, `tests/stack/run.sh`, `docs/operations.md`, `docs/dev/testing-strategy.md`. All in lane under live non-comment lines in `roles/appliance.paths`; `docs/` is shared and needs no override. No `.lane-override` was used.

The `run.sh` serialization condition was checked rather than assumed: the tests lane is not running, and `git ls-remote --heads origin` shows no tests-lane branch in flight touching it.

The PR gate keys off the pinned shell's branch rather than the worktree the work is in, so it cannot see this diff; this is the pass done by hand against the right one.

The change to the product is two lines and a comment. The alternative considered and rejected was wrapping the whole verb — rejected because the confirmation prompt is inside it, which is the hang this placement exists to avoid, and #1391 already ruled that the lock goes around mutating windows rather than whole verbs. Deriving the second and third residual verbs "by symmetry" with the first was also available and rejected: #1528's non-author pass showed that argument holding by luck rather than by evidence, so each verb gets its placement mutation-tested directly.

The test file could have been three files (one per group) or one; one is right — the groups share a fixture and a holder, and splitting them would triple the sandbox setup to no end.

**Not reviewable by its author.** This needs a non-author pass.

## Rebase note

Rebased onto `abb23ee` through one conflict in `tests/stack/run.sh`: #1533's registration stanza
and this branch's landed at the same anchor. **Both are kept**, blank-line separated, because a
stanza sourced inside another clobbers `_d0` and silently miscounts the domain. Verified after the
resolution: zero conflict markers, both stanzas present, 49 stanzas all at column 0, `bash -n` clean.
The branch carries exactly one commit against the new base.

The `run.sh` serialization condition is discharged rather than waived: #1105 R14 (`d115315`) has
already landed on that file, and the tests lane is currently reaped so there was nobody to message.


VijitSingh97:
## Verdict: MERGE

Controller non-author review, opus tier. Ground: the lock is one narrow change, its test is shown
load-bearing by a real controlled pair, and the rebase this PR sat through was resolved correctly at
the one place it could have gone silently wrong. The independent pass returned MERGE-WITH-NOTES with
no blocker, and I re-derived the claims the decision turns on rather than relaying them.

### Proven by me, at source

- **Ruling 8a, NARROW case, stated explicitly.** `merge-base(develop-v2, head)` equals the tip
  `abb23ee1`, and `merge-tree --write-tree` returns `896c10eb28dae1ebf9895d5dc39e1edb1405f5cc` —
  byte-identical to the branch's own tree. One gate run answers both questions.
- **File-budget gate: 85 rows, 0 failures** at the head, re-derived in Python over
  `git show <ref>:<path>`, with a control fired. The PR touches no budget row, so ruling 9 has
  nothing to satisfy; `tests/stack/run.sh` is 246 lines, under the 400 target, correctly unrowed.
- **The rebase conflict resolution is sound, and this was the sharp risk.** #1533's registration
  stanza and this PR's landed at the same anchor in `run.sh`. A stanza sourced *inside* another
  clobbers `_d0` and silently miscounts the domain — the failure would be a wrong count, not an
  error. Both stanzas are present and separated (lines 220-221 and 223-224), there are 49 `_d0=`
  stanzas, and **zero of them are indented**, so nothing is nested.
- **The verb count in the body is wrong, and I settled it by content rather than by either party's
  word.** Comparing `mutation_lock_acquire` label sets at base and head: the head adds exactly one
  label, `rotate-secrets`. Of the six verbs #1482 names, **four were already locked at the base**
  (`factory_reset`, `config_reset`, `reset_dashboard`, `backup`, via #1510 and #1528), so this is
  the **fifth**, not "the second of the residual verbs" as the body says. `firstboot_wizard` has
  zero acquire sites and is the sole remaining one. Controls fired in both directions. **This is a
  merge-record defect, not a code one** — recorded here rather than force-pushed, per the standing
  rule. Nothing about the lock's behaviour changes.

### Re-derived by the independent pass, not merely relayed by it

Worth naming because the pass did the expensive half rather than the cheap one: it **rebuilt
`pithead` from source and obtained byte-identity with the committed artifact** — which is what
confirms the rebase carried #1533's slice change through, rather than leaving a stale artifact that
merely happens to match — ran the new test file standalone for 15/15, and then **mutated a copy of
the built binary to remove exactly this PR's lock lines, getting 9/15 with the precise six rows the
author's own mutation table predicted.** That is a controlled pair, not a null experiment: the test
is shown able to fail, on the artifact that ships.

### The gate, and how it was cleared

The PR body's `3035 passed, 0 failed` is from the **pre-rebase** head, and the author said so in
bold and said not to merge on it — correctly, since a rebase invalidates the whole proof set and not
just the suite. That figure was not used. The rebased re-run's rc file was read directly:

**`~/appliance-suite-logs/1482-rebased-20260829T133035.log.rc` contains `0`**, and the log's own
summary line reads `pithead tests: 3043 passed, 0 failed`. I read the rc FILE, not the process, and
I cross-checked the summary independently: the log carries exactly **3043 `✓` glyphs** and **zero**
failure glyphs, and every occurrence of the word "failed" in it is part of an assertion's *name*
(each prefixed by its own `✓`), not a report.

Two further checks, because a green suite that did not run the new code proves nothing:

- **This PR's three new domains are present in the run** — `the fixture arms — uncontended,
  rotate-secrets performs its mutations (#1482)`, `a held machine refuses rotate-secrets and changes
  nothing (#1482)`, and `the rotate window opens AFTER the confirmation, so no hold spans a human
  wait (#1482)`.
- **Control that the run is on the REBASED base, not the old one:** the same log carries
  `the shipped pithead renders without the optional dashboard auth key (#1246)`, which is #1533's
  domain and only exists on a base that already carries it. 283 domains in total.

The pre-rebase `3035` was correctly discarded; the difference is #1533's own test file arriving in
the base between the two runs.

### What I did not check

I did not re-run the full suite myself, and I did not independently verify the lock's behaviour
under real concurrent invocation — the test file's 15 assertions and the pass's mutation pair are
what stand behind that, and both are disclosed above for what they are.

Issue #1409, opened 2026-08-24
Worker Inspect accuses another dashboard when our own SQLite read fails


VijitSingh97:
## What happens

When the dashboard's own SQLite read fails, Worker Inspect tells the operator their rig was changed
from **another dashboard** — in warning colours. Nothing happened on the rig. Our database hiccuped.

The rendered string is not hedged:

| | |
|---|---|
| class | `status-warn` |
| label | `Last changed from another dashboard` |
| detail | `applied over a control channel, with a change id this dashboard has no record of` |

That last line is the tell: *"this dashboard has no record of"* is stated as a fact about the rig's
history, when it is actually a fact about a read that failed.

## The chain, traced at the current `develop-v2` tip

**The handler is `get_worker_config_history`'s, at `service/storage_service.py:906-908`:**

```python
        except sqlite3.Error as e:
            self.logger.error(f"Worker Config Read Error: {e}")
            return []
```

That file carries 32 `except sqlite3.Error` handlers. This issue is about this one, and about one
sibling path in the same method (below) — I have **not** audited the other 30, and this should not be
read as a claim about them.

From that `[]` the verdict follows deterministically:

1. `web/worker_detail.py` scans the returned history for the rig's `last_change_id`. An empty list
   yields no match, so `change_id_known` is `False`.
2. The same function computes `history_truncated=len(history) >= _HISTORY_LIMIT`. `0 >= 50` is
   `False`, so the window reads as *not* full — correctly, an empty list is not a full window.
3. `client/rig_config_meta.py:155` then takes the one remaining branch:

```python
    if source == "control":
        if not change_id_known:
            return "untraced" if history_truncated else "elsewhere"
```

`elsewhere`. The accusation.

**A second route reaches the identical outcome** — `storage_service.py:888-889`, `if not self._conn:
return []`, when there is no database connection at all. Same empty list, same verdict.

## Why this is worth its own issue rather than a footnote

The truncated-window fix (merged as the `untraced` verdict) removed a false accusation reached by one
route: a change id that sits one row past a full history window. **This is the same false accusation
reached by a different route, and that fix deliberately preserved it** — the docstring says so in as
many words, because an empty list genuinely is not a full window and widening `untraced` to cover it
would have been the wrong shape.

So the remaining defect is not an oversight in that change; it is the part that change scoped out.

## The shape, and why the obvious patch is the wrong one

The failure is that **an error and an answer are returned in the same channel.** `[]` means both
[...]

PR #1536, opened 2026-08-29
fix(dashboard): answer config provenance by change id, not by window scan (#1369)

### LINKED ISSUE #1369: Worker Inspect provenance: the 50-row history window turns a change of ours into "another dashboard"

VijitSingh97:
Third and last PR for #1369, on the room the first two freed. **Does not close #1369** (`do-not-close`).

## What was wrong

The Worker Inspect provenance line asked whether the rig's `last_change_id` appeared among the 50
history rows the page renders. Past that window the verdict stopped working in **both** directions
at once — it could neither claim a change that was ours nor name one that was not.

**The issue text is stale, and the half nobody wrote down is the dangerous one.** #1369 says the
verdict "flips to `elsewhere`", a false accusation. It does not any more: `untraced` shipped after
the issue was filed and took the accusation back out. What was left is symmetric, and the residue on
the other side is a false *reassurance* — a genuinely foreign change on a long-history rig also
landed on "we cannot tell", so the softening **silenced a true alarm**. `workerlogic.mjs` already
documented that as an accepted trade.

Reproduced before writing anything, four legs in one run, both controls fired:

| leg | history | filler rows | verdict | truth |
|---|---|---|---|---|
| **A (control)** | ours | 49 | `here` | correct |
| **B** | ours | 50 | `untraced` | should be `here` |
| **C (control)** | foreign | 49 | `elsewhere` | correct |
| **D** | foreign | 50 | `untraced` | should be `elsewhere` |

The window is a *rendering* limit. It was never a fact about what this dashboard spooled, and
letting it bound the verdict made it one.

## The fix — commit 1

`get_worker_config_change(worker, change_id)` looks the id up directly against this rig's own
`worker_config` rows, unbounded. Three-valued on the #1409 contract and failing **closed**: `None` =
the read failed, `{}` = no such row, a dict = the row. It lands in `worker_config_store.py` — the
module that owns the table, which is what PR 2's mixin extraction freed it to do.

`worker_detail.py` now passes `bool(matched)` and `history_unread=matched is None or history_unread`,
and no longer passes `history_truncated` — which is what leaves `untraced` with no producer.

**`worker_config_change_known` (#530) keeps its own query, and that is the one design point worth
reading.** The two methods must disagree about failure: #530 returns `False` on a closed handle and
`True` on a `sqlite3.Error` — both correct there, where a false "known" merely declines to accuse a
rig — while the provenance path needs "we do not know" for *both*, because the same two properties
invert into the one answer this feature must never give wrongly. So no single three-valued helper
can serve them. **Sharing the SQL as well was this change's first draft; measuring it is what took it
back out** — see the review section below. What the two genuinely must share is the row *shape*, and
only the reads that return a row have that in common: `_SELECT_CHANGE` is where they share it.

**Fallbacks costed and rejected, so nobody re-derives them:** raising the limit moves the cliff
rather than removing it, and makes every Inspect load read more rows to answer a single-id question;
a larger second read on a miss has the same defect one cliff further out; answering the under-claim
alone via `worker_config_change_known` fixes half and re-introduces the unscoped, fail-open trap the
comment beside that method warns against.

## The judgment call — commit 2, deliberately separable

`untraced` existed to hedge a miss commit 1 makes impossible, so the fix leaves it with no producer.
Commit 2 retires it from the verdict layer, `workerlogic.mjs`, both test tiers and `docs/dashboard.md`.
**Drop commit 2 and a working fix remains** — exercised, not asserted; see below.

Kept as a removal rather than a producer-less branch on purpose: a verdict nothing can emit is not
gone, it is *unreachable*, and unreachable is what goes stale unnoticed — the docs would go on
describing a line no operator can see. The unreachability claim is **enumerated, not asserted**: the
only production caller of `config_origin` is `worker_detail.py:155`, and a new sweep over every
surviving argument combination pins it, with its own non-empty control (which caught a wrong expected
set on its first run).

## A review finding, measured and acted on before opening

The first draft routed `worker_config_change_known` through the same row query as the provenance
lookup, under a "share the query and nothing else" story. **That story was wrong, and it cost
something.** #530 asks only whether a row EXISTS — it never reads `ts` — so the shared query imported
an `ORDER BY` it has no use for, six columns instead of one, and a `json.loads` on the matched row.

Measured with `EXPLAIN QUERY PLAN` against the SQL the shipped methods actually execute (captured
through a `sqlite3` trace callback, not transcribed by hand). There is no index on `change_id`; the
only index on the table is `idx_worker_config(worker, ts)`:

| query | plan |
|---|---|
| **before this PR** — `SELECT 1 … WHERE change_id=? LIMIT 1` | `SCAN worker_config` |
| **first draft** — #530 through the shared row query | `SCAN worker_config` + `USE TEMP B-TREE FOR ORDER BY` |
| **as shipped now** — #530 back on its own `SELECT 1` | `SCAN worker_config` |
| **as shipped now** — the provenance lookup | `SEARCH worker_config USING INDEX idx_worker_config (worker=?)` |

`_reconcile_worker_config` calls `worker_config_change_known` once per reporting rig per poll, so the
first draft added a temp B-tree sort per rig per cycle and bought #530 nothing. **Control:** adding a
`change_id` index visibly flips that plan to a covering-index search, so the instrument reports index
use rather than one fixed answer — and it does **not** remove the temp B-tree, which is how the sort
is attributable to the `ORDER BY` rather than to the missing index. Not a benchmark and not a claim
of user-visible slowdown: `worker_config` grows one row per operator config change, so the table is
small today. It was gratuitous, and the fix was to delete a helper.

**The same measurement falsified a comment I had written.** `worker_detail.py` said the new read was
"a single indexed row". It is an index *seek* on `worker` and then a walk of that rig's own rows, and
on the **miss** path — exactly the foreign-change case this line exists to catch — it walks all of
them. Still better than the window scan (no horizon, and no sort, because `(worker, ts)` already
supplies the ordering), but the comment overstated it, so both it and the matching docstring line now
say what was measured. Found by the reviewing controller; re-derived at source by me before acting.

## Evidence — PROVEN BY ME at this head, instrument shown to fire first in every case

- **Positive control for the fix.** Reverting the two runtime files to `origin/develop-v2` and
  running `test_worker_detail_meta.py` gives **3 failed, 25 passed** — failing exactly `[50]`,
  `[120]` and the leg-D foreign-change case, while `[48]`/`[49]` stay green, so the new tests are not
  red for a blanket reason. Both files restored byte-identical afterwards (sha256 checked against
  `HEAD`'s blobs; working tree *and* index verified clean — `git checkout <ref> -- path` stages as
  well as writes, so restoring from the index would have silently kept the reverted copy).
- **Suite:** `2146 passed`, rc 0. Total coverage **97.03%** against the ≥80% bar.
- **Frontend:** `481 pass, 0 fail`.
- **Patch coverage, with the vacuous pass ruled out.** `diff-cover` vs `origin/develop-v2`:
  **92% over 27 measured lines**, against the ≥90% bar. The line count is the part that matters —
  `diff-cover` exits **0** printing "No lines with coverage information in this diff", so rc alone is
  no pass. Control run, same command against an empty diff: rc 0, zero lines, that exact message.
  `make test-patch-coverage` as CI invokes it also exits 0 — but **that rc is not evidence about this
  patch**, and I should not have listed it as though it were: it grades 1769 lines across 86 files
  because of #1537 below, so it clears 90% on `develop-v2`'s aggregate rather than on my 27 lines.
  The `origin/develop-v2` figure above is the one that means anything here. Its overlap self-test
  passes 4/4, which is a real check of the wrapper's own logic and nothing more.
- **Lint:** `lint-py lint-js lint-md lint-docs-voice lint-operator-strings lint-topology
  lint-file-budget` all rc **0**, measured per target rather than read off a pipeline. No shell files
  in this diff, so `lint-sh` does not apply.
- **Commit 1 alone is green:** `103 passed` across `test_worker_detail_meta.py`,
  `test_rig_config_meta.py` and `test_storage_worker.py` at commit 1's tree.

## Evidence — RELAYED, and what I bridged

The separability claim was first exercised by the reviewing controller, in its own worktree, and the
coupling risk it tested is one my own runs could not see: **commit 1 ships the new
`test_worker_detail_meta.py` against the OLD `config_origin`**, because `rig_config_meta.py` is
touched only by commit 2. Its result: leg A `103 passed, rc 0`; control (reverting `worker_detail.py`)
rc 1 with exactly those three failures; `history_truncated` has a default at commit 1 so dropping the
argument cannot `TypeError`, and at commit 2 `git grep -F history_truncated` over the tree is zero.
**Proof log — check this rather than take my word:**
`~/split-baselines/1369-pr3-separability-controller-w41.md`.

That run predates the fix above, so I bridged it rather than relaying it across a moved tree: I
re-ran leg A myself at the new commit 1 (same `103 passed`) and re-ran the reverted-file control
myself (the 3 failed / 25 passed above). `worker_detail.py` is **AST-identical** across the amend —
comment-only, with a seeded one-argument control confirming the comparison can report a difference,
and `worker_config_store.py` correctly reporting as changed.

The control's failure text, `AssertionError: assert 'untraced' == 'elsewhere'`, is also a second and
better-grounded route to commit 2's premise, and it now sits in commit 2's message: at commit 1,
where `untraced` is still fully wired, the old `worker_detail.py` still *reaches* it. A sweep proving
nothing reaches a branch says more when the branch is demonstrably still live.

## Named gaps — what I did NOT do

- **Two lines of this patch are uncovered**, and they are why the number is 92% rather than 100%:
  `worker_config_store.py:49-50`, the `except (TypeError, ValueError)` fallback in `_shaped` that
  turns unparseable `changes` JSON into `{}`. That code is **not new** — it exists verbatim at the
  base inside `get_worker_config_history`, and the collapse into `_shaped` relocated it, which is why
  a line-based diff counts it as added. I did not add a test because
  `tests/service/test_storage_worker.py` sits on a **zero-headroom** budget row (below), and a ratchet
  raise is not something covering a pre-existing defensive branch justifies. I did not measure
  whether the base covered its own copy.
- **No review of the final tree by anyone but me.** The controller reviewed the previous head; the
  regression fix above landed after that review and has not been read by anyone else.
- **No benchmark.** The query-plan work above is a plan comparison, not a wall-clock measurement.
- **Nothing seen on screen — this lane has no browser.** The retired `untraced` line's absence is
  proven in the verdict layer and in both test tiers, but not visually.

## Disclosures

- **`docs/` is shared** and this PR edits `docs/dashboard.md`. Its "so the verdict is one you can
  check by eye" claim was **falsified by the fix** — the verdict no longer comes from the table
  below it — so the paragraph is rewritten rather than trimmed, and the `untraced` bullet is deleted
  in commit 2.
- **Two new file-budget rows, both at zero headroom by the gate's own rule 2:**
  `dashboard/tests/service/test_storage_worker.py 451` and
  `dashboard/tests/web/test_worker_detail_meta.py 425`. Both crossed the 400 target. `run_gate` fails
  on `lines -gt ceiling`, so equal passes — but neither can grow again. If either must,
  `worker_config_store.py` being its own module makes a `test_worker_config_store.py` split the
  natural cut. **`worker_config_store.py` itself is 227 lines and correctly takes no row** — a file
  under the target must not have one. (Corrected from 226 after review: that figure was measured
  before the last edit to this branch and never re-taken. Counted the way the gate counts, by
  record — `awk 'END{print NR}'` — not by trailing newline.)
- **The boundary test was extended, not replaced.** `tests/web/test_worker_detail_meta.py` was
  already parametrized `[(48, "here"), (49, "here"), (50, "untraced")]`; it is now `[48, 49, 50, 120]`
  all expecting `here`, **with two anti-false-pass assertions** — that the rendered window is still
  bounded, and that the claimed row is *not* among the rendered ones past 49. Without those, the new
  rows would pass if the fix had worked by simply rendering more history.
- **The base moved under this branch.** Merge-base is `abb23ee`; `origin/develop-v2` is now `f2e0bde`
  (#1534 landed after the cut). Its two commits touch `lib/`, `pithead`, `tests/stack/` and two docs
  — no file this PR touches, and no budget row lowered, added or removed. Not rebased, per the
  controller's ruling; re-running at merge time is the merger's call.

## One finding filed, not fixed here

`scripts/patch-coverage.sh` hardcodes `COMPARE=origin/develop`, while the CI step that calls it
fetches `${GITHUB_BASE_REF:-develop}` and its comment states the gate grades "vs the PR's own base".
Nothing under `scripts/` or the `Makefile` reads that variable — verified by grep. So every PR into
`develop-v2` is graded against the frozen `develop`: this one measures over **1769 lines across 86
files** instead of its own 27. It passes today only because the aggregate is 98%, and the failure it
is supposed to prevent — a PR failing on lines it never touched — is the one it can still produce.
`scripts/` is not this lane's path, so it is filed rather than touched.

### DIFF
--- a/dashboard/mining_dashboard/client/rig_config_meta.py
+++ b/dashboard/mining_dashboard/client/rig_config_meta.py
@@ -105,9 +105,7 @@ def parse_config_meta(meta):
     return out if any(out.values()) else None
 
 
-def config_origin(
-    meta, change_id_known, change_status=None, history_truncated=False, history_unread=False
-):
+def config_origin(meta, change_id_known, change_status=None, history_unread=False):
     """Where the rig's current config came from, as far as we can honestly tell.
 
     ``change_id_known`` is whether this dashboard has a ``worker_config`` row for the rig's
@@ -131,25 +129,19 @@ def config_origin(
                       into either would state something we cannot support.
     - ``elsewhere`` — applied over a control channel, with an id we have never seen. Another host,
                       or our record of it is gone. Either way it is not something to present as ours.
-    - ``untraced``  — applied over a control channel, with an id we did not find, *and* the history
-                      we searched was full to its limit (#1369). The id may sit one row past the
-                      window, so "we have never seen it" is a claim the read cannot support and
-                      ``elsewhere`` would print an accusation over a change that may well be ours.
-                      ``history_truncated`` is what separates the two, and it is deliberately the
-                      only thing this argument may do: when it is False every verdict is exactly
-                      what it was before, so a caller that cannot tell loses nothing. That matters
-                      most on the error path, which since #1409 is ``history_unread``'s to answer:
-                      a failed read returns None, not ``[]``, so it never reaches this comparison
-                      at all rather than arriving as a window that merely looks short.
     - ``unread``    — applied over a control channel, and we could not read our own history at all
                       (#1409). ``get_worker_config_history`` returns None when the read failed, and
                       that is a different fact from an empty history: a miss we never got to look
                       for cannot support ``elsewhere``, which renders as "Last changed from another
                       dashboard" — an accusation sourced from our own broken DB rather than from
-                      anything the rig did. Distinct from ``untraced`` too: that one is a read that
-                      WORKED and came back full, so it is about the rig having changed often. This
-                      one is about us. Checked FIRST inside this branch, because a read that did not
-                      happen settles nothing further down it.
+                      anything the rig did. Checked FIRST inside this branch, because a read that
+                      did not happen settles nothing further down it.
+
+                      A miss that is NOT this is conclusive: since #1369 the caller asks for the
+                      id directly rather than looking for it in the window it renders, so there is
+                      no longer a "we could not see that far back" between ``here`` and
+                      ``elsewhere``. The ``untraced`` verdict that held that middle ground was
+                      removed with the window scan that produced it.
     - ``rig``       — applied on the rig itself.
     - ``restored``  — restored from a saved config by the operator-run ``restore`` command.
                       Deliberately not folded into ``rig``: a restore is not someone editing the rig.
@@ -168,7 +160,7 @@ def config_origin(
         if history_unread:
             return "unread"
         if not change_id_known:
-            return "untraced" if history_truncated else "elsewhere"
+            return "elsewhere"
         if change_status in HELD_STATUSES:
             return "here"
         if change_status in REVERTED_STATUSES:
--- a/dashboard/mining_dashboard/service/worker_config_store.py
+++ b/dashboard/mining_dashboard/service/worker_config_store.py
@@ -33,6 +33,24 @@ from typing import Any
 # `storage_service` so that import surface is unchanged by the split (#1369).
 _RECONCILE_TERMINAL = ("applied", "rejected", "rolled_back", "failed", "noop", "throttled")
 
+# One column list for every read of this table that returns a ROW, so a windowed read and an
+# exact-id read cannot drift into returning differently-shaped rows for the same change
+# (#1369). `worker_config_change_known` deliberately does not use it — see its docstring.
+_SELECT_CHANGE = "SELECT change_id, ts, status, changes, reason, type FROM worker_config"
+
+
+def _shaped(row: sqlite3.Row) -> dict:
+    """One ``worker_config`` row as the rest of the app reads it: ``changes`` parsed back to a
+    dict (unparseable JSON reads as ``{}`` rather than raising), and a row written before the
+    ``type`` column existed (#1014) reading back as ``"apply"``."""
+    d = dict(row)
+    try:
+        d["changes"] = json.loads(d["changes"]) if d["changes"] else {}
+    except (TypeError, ValueError):
+        d["changes"] = {}
+    d["type"] = d.get("type") or "apply"
+    return d
+
 
 class WorkerConfigStoreMixin:
     """The `worker_config` accessors of `StateManager`. Never instantiated on its own."""
@@ -82,20 +100,10 @@ class WorkerConfigStoreMixin:
                     return None
                 cursor = self._conn.cursor()
                 cursor.execute(
-                    "SELECT change_id, ts, status, changes, reason, type FROM worker_config "
-                    "WHERE worker = ? ORDER BY ts DESC, id DESC LIMIT ?",
+                    f"{_SELECT_CHANGE} WHERE worker = ? ORDER BY ts DESC, id DESC LIMIT ?",
                     (worker, limit),
                 )
-                out = []
-                for row in cursor.fetchall():
-                    d = dict(row)
-                    try:
-                        d["changes"] = json.loads(d["changes"]) if d["changes"] else {}
-                    except (TypeError, ValueError):
-                        d["changes"] = {}
-                    d["type"] = d.get("type") or "apply"
-                    out.append(d)
-                return out
+                return [_shaped(row) for row in cursor.fetchall()]
         except sqlite3.Error as e:
             self.logger.error(f"Worker Config Read Error: {e}")
             return None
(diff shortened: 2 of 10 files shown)


VijitSingh97:
**Controller verdict (w42): MERGE.** The non-author pass returned no blocking finding, and I re-derived the half of it that went stale under a moved base rather than reading it out of the answer file.

**The base moved after the pass ran.** The pass computed ruling 8a against `origin/develop-v2` = `f2e0bde8` (PR #1534's merge). PR #1538 merged after it, so the tip is now `2e89f66f` and the pass's merge-tree `7727214e` is no longer the tree this PR would produce. Re-derived at the real tip:

```
tip(develop-v2)  = 2e89f66fe3df15c2760cdd5812776432c8f76768
head(#1536)      = 2beab111f58e8f3cd678a5f868687c290489132e
merge-base       = abb23ee102e54fc1dee44eba20e3d1406eb477af   (unchanged)
branch tree      = 9af66f06980a09ef14347a8277a932c7a5e5c8e8
merge-tree       = 2f0f9df4c9eacb78dbf5f18bb72879377cf3d571   (was 7727214e)
```

Branch tree != merge tree, so this is ruling 8a **BROAD** and both trees were gated. `scripts/lint-file-budget.sh` in two throwaway worktrees: branch head rc **0**, materialized merge tree rc **0** (commit built with `commit-tree`, its tree verified byte-identical to the `merge-tree --write-tree` prediction before use). Self-test 26/26 ok on the merged tree. **Control fired:** appending two lines to `dashboard/tests/service/test_storage_worker.py`, a 0-headroom file, turns the same gate to rc **1** naming it at 453 over its 451 ceiling; restoring the file returns rc 0 with a clean tree. Two rc-0 readings with no control would have proved nothing.

**Base-move non-interference, re-derived at the new tip with controls.** `abb23ee..2e89f66` touches eight files — the six from #1534 plus `lib/pithead/12-firstboot-wizard.sh` and `tests/stack/test-appliance-firstboot-install-lock.sh` from #1538. `comm -12` against this PR's ten changed files is **empty**; the same `comm` against the PR's own list returns 10, so the instrument was shown to match something. `git diff abb23ee 2e89f66 -- docs/dev/file-budget.tsv` is **empty**, while the PR's own tsv diff adds exactly the two rows ruling 9 expects, both files this cut owns.

**Both non-blocking findings were acted on and I verified the text, not the promise:** the body now reads 227 lines for `worker_config_store.py` with the correction named, and the patch-coverage sentence now says outright that the rc 0 is not evidence about this PR's 27 lines.

CI at `2beab11`: 16 pass, 2 skipping, 0 fail.

**What this merge does NOT settle.** #1369 stays open. #1537 stays open and unfixed — every `make test-patch-coverage` green on this branch, including this PR's, is measured against frozen `origin/develop`, so it remains no evidence about the lines it appears to grade. And #1482 did not close on #1538: only item 1 (the six verbs) is done; item 2, the in-place upgrade fallback, still takes no lock.

Issue #1537, opened 2026-08-29
Patch-coverage gate grades every PR against frozen develop, not its own base


VijitSingh97:
`scripts/patch-coverage.sh` hardcodes its compare branch, so the gate grades every PR against the
frozen `develop` rather than the branch the PR is actually based on. The CI step that calls it goes
to the trouble of fetching the right ref and then nothing reads it.

## What is there

`.github/workflows/ci.yml`, the "Patch-coverage gate — changed lines >=90% (diff-cover vs the PR's
own base, #286)" step:

```yaml
    run: |
      git fetch --no-tags origin develop "${GITHUB_BASE_REF:-develop}"
      make test-patch-coverage
```

with the comment above it stating: *"GITHUB_BASE_REF is set on pull_request runs; push runs fall back
to develop. Grading a PR against develop instead of its base branch charges it the base branch's
whole coverage debt (a bash-only PR into the appliance branch failed on wizard.py lines it never
touched)."*

`scripts/patch-coverage.sh:11`:

```sh
COMPARE=origin/develop
```

`grep -rn GITHUB_BASE_REF scripts/ Makefile` returns nothing. The variable is fetched and never read,
so the intent recorded in that comment is not implemented. The comment describes the fix; the script
predates or lost it.

## Why it matters

All work lands on `develop-v2` while `develop` is frozen, so today every dashboard PR is graded
against the entire `develop` -> `develop-v2` divergence rather than against its own changes.

Measured on the open PR #1536, same `coverage.xml`, only the compare branch differing:

| compare branch | files | lines graded | result |
|---|---|---|---|
| `origin/develop-v2` (the PR's real base) | 3 | 27 | 92% |
| `origin/develop` (what the gate uses) | 86 | 1769 | 98% |

It passes today only because the aggregate is healthy. Both failure directions are live:

- **The one the CI comment says was already fixed** — a small PR failing on lines it never touched,
  because it inherits the base branch's coverage debt. That is exactly the appliance-branch incident
  the comment describes.
- **The quieter one** — a genuinely under-covered patch passing because 27 of its lines are averaged
  into 1769 that someone else already covered. A gate that grades 65x more lines than the patch
  contains is not measuring the patch.

## Suggested fix

Let the script take the compare branch from the environment, defaulting to what it uses now:

```sh
COMPARE="origin/${GITHUB_BASE_REF:-develop}"
```

Two things worth checking with it, neither of which I have done:

[...]

Issue #1535, opened 2026-08-29
Mixin-extraction atomicity checker: `_retained_state()` misses three ordinary `self.` assignment shapes; plus one comment falsified by #1532


VijitSingh97:
Two follow-ups from the non-author review of #1532 (the `worker_config` mixin extraction, #1369
PR 2). Neither blocked that merge; both were verified at source before filing.

## 1. `_retained_state()` cannot see three ordinary `self.` assignment shapes

`dashboard/tests/service/test_mixin_atomicity.py`, `_retained_state()`. It walks `ast.Assign` nodes
in `StateManager` and keeps `target.attr` only when the target `isinstance` is `ast.Attribute`. That
misses:

- **tuple / list targets** — for `self.a, self.b = x, y` the target is an `ast.Tuple`, so neither
  name enters the needle set;
- **`setattr(self, name, value)`** — not an `ast.Assign` node at all;
- **annotated assignment** — `self.x: int = 0` is an `ast.AnnAssign`, a different node type.

**Why it matters and why it is not urgent.** The Q2 check asks whether a moved method touches
in-memory state that stayed behind. A name the needle set never contains cannot be reported, so a
real violation through any of those three shapes passes silently. The tuple shape is already in
production use in this codebase two classes over (`data_service.py:747,790`), so this is an ordinary
Python shape, not a contrived one.

**It does not make #1532's `Q2 = 0` a false clean.** Re-derived independently with `ast` against the
merged `StateManager`: 23 plain `self.X = ...` assignments and **zero** of all three blind shapes.
A control on a synthetic class carrying all three fired, so the probe was shown able to see them.
The reviewer went further and demonstrated the live false-negative by seeding
`self._reset_epoch, self._reset_note = 0, ""` into `StateManager`, having a moved method read it,
and getting 7/7 green — with an arming readback confirming the seed was present in both files
before the run, and a clean revert after.

**Fix:** extend `_retained_state()` to unpack `ast.Tuple`/`ast.List` assignment targets, and to
handle `ast.AnnAssign` and `setattr(self, ...)`. Worth doing before a third mixin extraction reuses
this file as its template — `docs/dev/repo-map.md` notes that #1367's option A has room here too.

## 2. A stale file-location comment that #1532 falsified without touching the file

`tests/integration/rigforge-apply-settle.sh:13` reads
`(storage_service.py:reconcile_worker_config_status, #579/#604)`. That method now lives in
`worker_config_store.py`.

[...]

PR #1539, opened 2026-08-29
test(dashboard): let the atomicity walk see every shape state can be bound in (#1535)

VijitSingh97:
Both items from #1535, one commit each, both filed off the non-author review of #1532. Decoupled
from #1369 PR 3 (#1536) by the controller's w41 ruling — they share no boundary with it, and this
branch is cut from `origin/develop-v2` rather than stacked, so the two can merge in either order.

## Item 1 — the atomicity walk could not see three ordinary ways to bind state

`_retained_state()` in `dashboard/tests/service/test_mixin_atomicity.py` read only `ast.Assign`
targets that were themselves `ast.Attribute`. That is blind to tuple and list unpacking, to
`ast.AnnAssign`, and to `setattr(self, "x", v)`.

**The failure mode is what makes this worth fixing rather than noting.** Q2 in that file asks
whether a method that moved to a mixin touches in-memory state that stayed behind, and it answers
**by name**: a name the needle set never contains produces no finding no matter how badly a method
reaches for it. So a narrow walk here fails as a clean **PASS** — in the one file whose entire
purpose is to stop a split reading as safe when it is not.

**It does not make #1532's `Q2 = 0` a false clean, and I re-derived that before writing anything:**
the merged `StateManager` has 128 plain `self.X = …` assignments, one `AnnAssign`, one `AugAssign`
and **zero** of the three blind shapes. This is prophylaxis before a third mixin extraction reuses
the file as its template, which `docs/dev/repo-map.md` notes #1367 may.

### The controls, because a walk that reports nothing looks exactly like a split that touches nothing

- **Positive control** — every shape seeded, one name each, in a class **literally named
  `StateManager`** (the walk filters on that name, so a control in a differently-named class comes
  back empty and reads as a broken walk rather than a broken fixture). It covers more than the three
  the issue names: nested tuple targets and a `Starred` target too, since the fix is recursive and an
  untested recursion is an assumption.
- **Negative control, the sibling that makes the widening safe** — a near-miss fixture where every
  name is one step off target: another object's attribute, a bare local sharing a tuple target with
  a real one, a `setattr` on something else, the same assignment in a differently-named class. Only
  the one genuine `self.` target survives. Widening a walk is only useful if it did not become "any
  attribute anywhere", and the positive control alone cannot tell those two apart.
- **The shared DB handle, still subtracted through a tuple target** — the place widening could have
  leaked it back in.
- **A probe shown able to see an unresolvable `setattr`**, because the subject assertion it supports
  has an *absence* for a product.

### The controlled pair — the new controls fail against the old walk

Run as a pair with both implementations loaded side by side, the old one taken from
`origin/develop-v2` rather than retyped, and the fixture read **out of the shipped test file** so the
experiment cannot drift from what CI runs:

| walk | names seen in the 12-name fixture |
|---|---|
| **old** | `plain` — one of twelve |
| **new** | all twelve |

Eleven of the twelve were invisible. Without that pair the new tests would be assertions that the
current code does what it does.

### One limit stated rather than papered over

`setattr(self, name, v)` with a **computed** name cannot be resolved by any static walk, so it stays
outside the needle set — the same false pass arriving through the one door widening cannot shut. A
new test asserts the subject contains none today, which turns the limit into a checked fact with a
test naming it instead of a gap someone meets by surprise.

## Item 2 — a stale file-location comment, and it is out of this lane

`tests/integration/rigforge-apply-settle.sh:13` named
`storage_service.py:reconcile_worker_config_status`; #1532 moved that method into the
`worker_config_store.py` mixin. The comment is load-bearing prose — it explains why observing the
enriched feed converge proves the #185 history row is already reconciled — so the wrong module sends
a reader looking for the mechanism in a file that no longer has it.

**Checked at the base before changing it**, which is what separates a fresh breakage from
pre-existing rot: the comment was true at `develop-v2` and became false at #1532's head, and #1532
does not touch this file.

**Swept mechanically rather than fixed one-off**, since a cut falsifies prose in files it never
touches and a per-item sweep is blind to earlier rounds' debris: every `storage_service.py:<name>`
reference in the repo, each name checked against the current module. This is the only stale one. The
two neighbouring `storage_service` mentions in `tests/integration/lib.sh` and `run.sh` are about the
`db_healthy` latch, and I re-derived them as still true — `is_db_healthy` is still defined in
`storage_service.py`.

### Out-of-lane disclosure

**`tests/integration/` is the `e2e` lane's path, not `dashboard`'s.** No e2e lane is running, so this
was taken under a **one-shot `.lane-override`, in its own commit**, per the controller's w41 ruling.
Prose only — a `#` comment, no executable reader.

**The override was verified rather than assumed, in both directions:** committing without it was
refused, with the guard naming exactly this file and nothing else; committing with it printed
`override CONSUMED (one-shot)`; and the marker is gone afterwards, checked directly, because it lives
in `.git/info/exclude` and a leftover would be **invisible in `git status`** while silently disarming
the guard for this worktree.

## Evidence — PROVEN BY ME at this head

- **Suite:** `2141 passed`, rc 0 — `develop-v2`'s 2136 plus the five tests added here, which is the
  arithmetic that says nothing else moved. Total coverage **97.04%** against the ≥80% bar.
- **The changed file alone:** `12 passed`.
- **The controlled pair above**, old implementation loaded from git, fixture read from the shipped
  file.
- **Lint:** `lint-py lint-js lint-md lint-docs-voice lint-operator-strings lint-topology
  lint-file-budget` each rc **0**, run by name.
- **The shell file:** `shellcheck` 0.11.0 rc 0 at **default severity**, which is stricter than the
  gate's own `--severity=warning`; `shfmt -d` prints nothing. Run on the single changed file rather
  than through `make lint-sh`, which is the serialized OOM path and cannot be affected by a
  comment-text change — stated because it is a target I did not run in full.
- **File budget:** `test_mixin_atomicity.py` is **387 lines** and takes no row, correctly — it is
  under the 400 target, and a file under the target must not have one. Worth naming that it now has
  only ~13 lines of headroom before it would need one.

## Named gaps

- **Patch coverage is NOT APPLICABLE here, and that is different from passing it.** Against the real
  base, `diff-cover` reports "No lines with coverage information in this diff" over **zero** measured
  lines — correct, since this PR changes a test file and a shell comment, and coverage measures only
  `mining_dashboard/`. **I am not claiming the ≥90% bar was met; there was nothing to meet it with.**
- **No review by anyone but me.**
- **The new walk is proven against seeded fixtures and against today's subject, not against a real
  violation** — there is no moved method that reaches back through one of these shapes, because there
  is no such method at all. The controls are what stand in for that.

## This sharpens #1537, which I filed while opening #1536

#1537 reports that `scripts/patch-coverage.sh` hardcodes `COMPARE=origin/develop` while CI fetches
`${GITHUB_BASE_REF}` and nothing reads it. **This PR is a cleaner demonstration than the one in that
issue.** Against its real base the honest verdict is "not applicable — zero measured lines". Run as
CI runs it, the same wrapper instead reports **`all 86 changed file(s) ... present in coverage.xml`
and `Coverage: 98%`** — a 90%-bar *pass* over 86 files this PR does not touch.

So the wrapper's not-applicable branch, added by #1000 precisely to stop a diff with nothing
measurable from reading as a coverage pass, **is unreachable for any PR into `develop-v2`**: the
hardcoded compare always drags in the whole divergence. The vacuous pass #1000 closed is open again
one level up. Posting this on #1537 as well.

### DIFF
--- a/dashboard/tests/service/test_mixin_atomicity.py
+++ b/dashboard/tests/service/test_mixin_atomicity.py
@@ -87,19 +87,85 @@ def _guarded_calls(method: ast.AST, targets: set[str]) -> list[str]:
     return sorted(hits)
 
 
-def _retained_state(subject: str) -> set[str]:
-    """Every attribute ``StateManager`` assigns to itself, minus the shared DB handle."""
-    return {
-        target.attr
-        for node in ast.walk(ast.parse(subject))
-        if isinstance(node, ast.ClassDef) and node.name == "StateManager"
-        for assign in ast.walk(node)
-        if isinstance(assign, ast.Assign)
-        for target in assign.targets
-        if isinstance(target, ast.Attribute)
+def _self_attr_names(target: ast.AST) -> set[str]:
+    """The ``self.<name>`` attributes one assignment target binds.
+
+    Recursive because a target can nest: ``self.a, (self.b, self.c) = ...`` is a ``Tuple`` holding
+    a ``Tuple``, and ``self.a, *self.rest = ...`` puts one of them behind a ``Starred``. Anything
+    that is not an attribute on ``self`` — a bare local, an attribute on some other object —
+    contributes nothing, which is what keeps the walk narrow rather than merely wide.
+    """
+    if isinstance(target, (ast.Tuple, ast.List)):
+        return {name for element in target.elts for name in _self_attr_names(element)}
+    if isinstance(target, ast.Starred):
+        return _self_attr_names(target.value)
+    if (
+        isinstance(target, ast.Attribute)
         and isinstance(target.value, ast.Name)
         and target.value.id == "self"
-    } - _SHARED_HANDLE
+    ):
+        return {target.attr}
+    return set()
+
+
+def _retained_state(subject: str) -> set[str]:
+    """Every attribute ``StateManager`` assigns to itself, minus the shared DB handle.
+
+    Four binding shapes, not one (#1535). The first version read only ``ast.Assign`` targets that
+    were themselves ``ast.Attribute``, which is blind to tuple/list unpacking, to ``ast.AnnAssign``,
+    and to ``setattr(self, "x", v)`` — none of them contrived: the tuple shape is in production use
+    two modules over in ``data_service.py``.
+
+    **A shape this misses cannot be reported, and the silence is indistinguishable from a clean
+    result.** Q2 below asks whether a moved method touches state that stayed behind, and it answers
+    by name: a name the needle set never contains produces no finding no matter how badly a method
+    reaches for it. So the failure mode of a narrow walk here is a false PASS, which is the one this
+    file exists to prevent — hence the controls below covering every shape, and a near-miss sibling
+    proving the widened walk did not simply become "any attribute anywhere".
+    """
+    state: set[str] = set()
+    for node in ast.walk(ast.parse(subject)):
+        if not (isinstance(node, ast.ClassDef) and node.name == "StateManager"):
+            continue
+        for statement in ast.walk(node):
+            if isinstance(statement, ast.Assign):
+                for target in statement.targets:
+                    state |= _self_attr_names(target)
+            elif isinstance(statement, ast.AnnAssign):
+                state |= _self_attr_names(statement.target)
+            elif _is_self_setattr(statement):
+                state.add(statement.args[1].value)
+    return state - _SHARED_HANDLE
+
+
+def _self_setattr_calls(subject: str) -> list[ast.Call]:
+    """Every ``setattr(self, ...)`` inside ``StateManager``, resolvable name or not."""
+    return [
+        statement
+        for node in ast.walk(ast.parse(subject))
+        if isinstance(node, ast.ClassDef) and node.name == "StateManager"
+        for statement in ast.walk(node)
+        if isinstance(statement, ast.Call)
+        and isinstance(statement.func, ast.Name)
+        and statement.func.id == "setattr"
+        and statement.args
+        and isinstance(statement.args[0], ast.Name)
+        and statement.args[0].id == "self"
+    ]
+
+
+def _is_self_setattr(node: ast.AST) -> bool:
+    """``setattr(self, "literal", value)`` — a static walk can only resolve a literal name."""
+    return (
+        isinstance(node, ast.Call)
+        and isinstance(node.func, ast.Name)
+        and node.func.id == "setattr"
+        and len(node.args) == 3
+        and isinstance(node.args[0], ast.Name)
+        and node.args[0].id == "self"
+        and isinstance(node.args[1], ast.Constant)
+        and isinstance(node.args[1].value, str)
+    )
 
 
 def _state_reads(method: ast.AST, retained_state: set[str]) -> list[str]:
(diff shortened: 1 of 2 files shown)


VijitSingh97:
## Controller verdict — MERGE

Non-author adversarial pass returned **no findings**, and I re-derived the claims the merge turns on
rather than relaying them. Merging.

### Ruling 8a — BROAD case, gated on both trees

The branch is not on the tip, so the merged tree and the branch head are different objects and both
were gated. Re-derived at the tip live, not taken from the brief:

| | |
|---|---|
| tip `develop-v2` | `12cedf3123833de0f8b14db2ad173629d4fb428a` |
| PR head | `f7e7f9cc3e56635d0864656b3386c836f747c4c7` |
| merge-base | `f2e0bde87f6accfe5364128b16c726bb49166a6f` |
| branch tree | `3c231bc73d0e296168b559b299b03fa0b52ce9bc` |
| merge tree | `3c1d6593972ce9eb199942175790d54f096599a4` |

All four match what the pass reported, and `git merge-tree --write-tree` returns rc 0. The base has
not moved since the pass ran, so the pass's numbers are still valid for this exact (base, head) pair.

### Budget gate — rows read, not the rc, with a control fired

**PROVEN BY ME.** Enumerated every row of `docs/dev/file-budget.tsv` against the blob's real line
count on both trees: **87 rows / 0 over ceiling** at the merged tree, **85 rows / 0 over ceiling** at
the branch head, no missing blobs on either. The real gate then returned rc 0 on both, and its
self-test passes 26/26 at the merged tree.

**Control fired, asserting on the refusal line rather than the rc** — lowering the
`config/config.py` ceiling 717 → 716 at the merged tree produces
`file-budget: FAIL — dashboard/mining_dashboard/config/config.py is 717 lines, over its recorded
ceiling of 716`, and restoring returns rc 0 with a clean tree. The gate is live at this tree.

The row-count difference is the branch being un-rebased, not a defect: the tip added rows for
`test_storage_worker.py` and `test_worker_detail_meta.py` that the branch does not carry. The
monotonic half is clean in the direction that matters — **no branch ceiling is higher than the tip's
for any shared path** (control: a fabricated +1 raise is detected). The PR touches the tsv not at
all, so the merged tsv is byte-identical to the tip's.

Both touched files are under the 400-line target and correctly carry no row:
`test_mixin_atomicity.py` at 387 lines, `rigforge-apply-settle.sh` at 58.

### CI — proven here, not relayed

`gh pr checks 1539` at `f7e7f9c`: **16 pass, 2 skipping** (the shipped-image CVE sweep, expected),
**0 fail**.

### What the pass established that I am accepting on its evidence

Two mutation experiments, both with the mutation reverted and the tree confirmed clean afterwards:

- Swapping the pre-widening `_retained_state` back in — taken verbatim from `origin/develop-v2`,
  not retyped — fails **4 of 12** tests, all inside the new class. The new tests are load-bearing
  against the exact prior implementation.
- Removing only the `ast.Starred` branch fails **exactly 1 of 12**, and the failure names
  `starred_rest` precisely. So the aggregate twelve-name assertion is not a count standing in for
  coverage: each shape is independently a discriminator inside it.

That answers the brief's real question — whether 186 new lines are a null experiment — in the only
way that could: by making the new code wrong and watching the right tests fail.

### Gaps the pass named rather than hid

It ran two of seven possible single-shape mutations (full revert, plus `Starred`) and reasoned the
other five from code inspection; it did not seed a moved-method-reads-a-tuple-bound-attribute case
through the full `split()` pipeline, arguing `_state_reads` matches on plain set membership and is
therefore shape-independent. Both are labelled RELAYED in its own report. I am accepting them: the
shape-independence claim is checkable by reading one predicate, and the residual risk is a test
being weaker than advertised, not production behaviour changing.

### Lane discipline

The out-of-lane file is handled correctly. `tests/integration/` belongs to the `e2e` lane; it was
taken under a one-shot `.lane-override` in its own commit per the w41 ruling, that commit touches
exactly one file `+1/-1` with no bystanders, and the marker is confirmed absent from the dashboard
lane's worktree afterwards — checked as a file, since it lives in `.git/info/exclude` and never
appears in `git status`.

### Issue

#1535 names exactly two items and both ship here, so it closes on merge — by hand, since a closing
keyword is inert into a non-default branch.

PR #1540, opened 2026-08-29
test(dashboard): make the duplicated web test builders real fixtures (#1459)

### LINKED ISSUE #1459: Consolidate the duplicated dashboard test-fixture builders into real pytest fixtures after #1105

VijitSingh97:
Closes #1459.

The four `tests/web/` view-test modules each carried its own copy of seven builders. The copies
move to a new `tests/web/conftest.py` — `_SYNC_DONE` and `_BASE` as module constants, the other
five as factory fixtures. The commit message says what moved and what deliberately did not; this
body is the verification record.

## The issue's stated gate is the wrong gate, and meeting it would have broken a test module

#1459 asks for "a search for the builder names finding exactly one definition each". That cannot
be met, and should not be.

Grouping every top-level definition of the seven names by source identity (AST, not grep) shows
the four `tests/web/` modules carrying byte-identical copies — and every definition outside those
four being a different body, **except one**. `tests/service/test_telegram_commands.py`'s `_metrics`
is byte-identical to the shared one and **is not the same builder**: its `_BASE` differs
(`current_tier`/`target_tier` are `"Donor"`, not `"Donor (1.00 kH/s+)"`) and it closes over a
`_SYNCED` of its own. Same source, different object.

`_hashrate` has the same shape and comes out the other way: byte-identical in two modules *and*
closing over `build_hashrate`/`_mode_palette` that both import from `series_views` — the same
objects at runtime, so it consolidates safely.

**The rule that generalises: byte-identical source is not the same builder when it closes over a
module global. Compare what the free names are bound to, not just the body.** So the honest scope
is the four modules, and a name-count gate would have driven a consolidation that changed what
`test_telegram_commands.py` asserts.

## Verified on the merge tree against `develop-v2` at `89c0678`

The base moved to `89c0678` (#1539 landing) after the branch was cut, so everything below was
re-derived at that tip, on the materialized `merge-tree --write-tree` result — not on the branch
head alone. `merge-tree` is conflict-free; base ref probed before and after the gate runs and
unchanged.

- **Suite, controlled pair in one worktree each:** base `89c0678` → **2151 passed**, merge tree →
  **2151 passed**, both rc 0.
- **Collection is byte-identical:** 2151 node IDs each side, `diff` empty. The instrument is shown
  able to fire — dropping one ID from the base list makes the same comparison report a difference.
  (An earlier attempt at this measurement was **null**: the project's `addopts` already carries
  `-q`, so a second `-q` became `-qq` and printed per-file counts instead of IDs, over which "no
  difference" would have been a false clean. The control is what caught it.)
- **Identity of the moved bodies, re-derived independently** of the route used when the change was
  written (`ast.unparse` normalization rather than source-segment slicing, which has an indent
  trap): **25 name/module pairs identical, 0 differences**. Three seeded semantic perturbations
  (`workers_online=2→3`, a `_state_mgr` default, dropping `**over` from `replace`) are each **seen**
  by the pipeline, so it is not blind.
- **Nothing changed except signatures:** at full depth (the tests live in classes, so a top-level
  walk misses them), **143 signatures changed, 0 bodies changed, 148 functions untouched**; every
  added parameter is one of the five fixture names, and no parameter was removed.
- **Every fixture reference is bound:** 190 references, 188 by parameter and 2 by closure in a
  nested helper (`test_views.py:529`), **0 unbound**.
- **No silent capture:** the new conftest applies to all of `tests/web/`, so the other 9 modules
  there were scanned — **0 tests take a parameter named after one of the five**, and there is no
  fixture-name collision with `tests/conftest.py` (`_isolate_db`, `state_manager`).
- **Comment accounting** (this change ate a section header once during authoring, and it was
  restored — `# --- build_state integration ---` is present at `test_views.py:215`): 18 comments
  removed, 8 reproduced verbatim in the conftest, 4 section headers that introduced the deleted
  block correctly died with it, and 6 are the deliberate-duplication note now restated as conftest
  docstring prose.
- **File budget:** the three lowered rows equal the merge tree's actual counts exactly
  (768 / 775 / 1243), and `test_series_views.py` at 378 is under the 400 target with its row
  correctly dropped. Gate rc 0 on **both** the branch head and the merge tree, `--self-test` rc 0.
  Two doors fired by name and were restored by sha256: growing `test_views.py` past 775 gives
  *"is 780 lines, over its recorded ceiling of 775"*, and putting the dropped row back gives
  *"is 378 lines (<= 400 target) but still has a ... entry"*.
- **Lints, by name, each rc 0** on the merge tree: `lint-py` (ruff check + format), `lint-md`,
  `lint-docs-voice`, `lint-operator-strings`, `lint-topology`, `lint-js`, `lint-file-budget`.

## Patch coverage is NOT APPLICABLE, which is not "passed"

`diff-cover` against the real base prints *"No lines with coverage information in this diff"* and
exits **0** — every changed Python file is a test file, and coverage measures `mining_dashboard`
only. Re-derived at `89c0678`; there is nothing to grade here.

CI's `make test-patch-coverage` rc 0 is **also no evidence about these lines**, for a second and
independent reason: **#1537 is open and unfixed** — `scripts/patch-coverage.sh` hardcodes
`COMPARE=origin/develop` while CI fetches `${GITHUB_BASE_REF}` and nothing reads it. Neither
number says anything about this diff.

## Total coverage moved 97.03% → 97.04%, and it is not this change

One statement — `storage_service.py:693` — is covered on one leg and not the other. That line sits
inside `if random.random() < 0.05:`, a sampled pruning branch, and `storage_service.py` is the same
blob (`9914deb`) in both trees. It is the sampler, not a change in what is tested. Flagged rather
than rounded away, because a ±1 miss delta proves nothing either direction on its own.

## Not run, and one judgment call

- **`lint-sh` not run** — there is no shell file in this diff. (It is also the OOM path on this box.)
- **`test-frontend` not run** — no JS changed; `lint-js` was run anyway and is rc 0.
- **Nothing seen on screen** — this lane has no browser. The change is test-only, so there is no
  rendered surface to look at.
- **A judgment call a reviewer may want to overrule:** the deliberate-duplication note explaining
  why `tests/service/test_metrics.py` keeps its own `_state_mgr`/`_data` pair carried a
  `# ponytail:` prefix in the two modules it used to sit in. It is now conftest docstring prose,
  which says more than the comment did — but it has lost the prefix, and the duplication it
  justifies still exists. Anyone sweeping the repo's ~40 `# ponytail:` sites for known-deliberate
  oddities will no longer find this one. I left it as prose; restoring the prefix is a two-line
  change if a reviewer prefers greppability.

https://claude.ai/code/session_015m8V8NiA14jgqSZsf2au6C

### DIFF
--- a/dashboard/tests/web/conftest.py
+++ b/dashboard/tests/web/conftest.py
@@ -1,0 +1,159 @@
+"""Shared builders for the ``dashboard/tests/web/`` view-layer tests (#1459).
+
+Until now ``tests/web/test_views.py``, ``test_xvb_views.py``, ``test_infra_views.py`` and
+``test_series_views.py`` each carried its own copy of these builders. That was deliberate: the
+#1105 cuts proved themselves by moving test bodies verbatim, and converting the builders to
+fixtures would have rewritten every call site inside the same change. The cuts are finished, so
+the copies are consolidated here — the exit #1459 was filed to be.
+
+``_SYNC_DONE`` and ``_BASE`` stay module constants rather than fixtures because no test asks for
+either directly; they are read only by the two ``replace``-based factories.
+
+The five fixtures are FACTORIES: each yields the callable the modules used to define for
+themselves, so every call site reads exactly as it did before and only the test signature changes.
+
+Deliberately NOT consolidated — the name is shared, the code is not:
+
+* ``tests/service/test_telegram_commands.py`` defines a ``_metrics`` that is byte-identical to
+  this one, but its ``_BASE`` differs (``current_tier``/``target_tier`` are ``"Donor"`` there,
+  not ``"Donor (1.00 kH/s+)"``, and it closes over a ``_SYNCED`` of its own). Same source, a
+  different object — and it is under ``tests/service/`` anyway, out of this conftest's reach.
+* ``tests/web/test_prometheus.py`` and ``tests/web/test_node_location.py`` keep their own
+  ``_metrics``/``_sync``; both bodies differ. They sit under this conftest and are unaffected,
+  because a fixture reaches a test only through a parameter and neither module declares one.
+* ``tests/service/test_metrics.py`` keeps its own ``_state_mgr``/``_data`` pair. Its per-module
+  defaults differ on purpose (``tari_sync``, the ``get_tiers``/xvb shapes); a builder shared with
+  this one would need enough parameters to read worse than either copy. That note used to sit
+  above ``_state_mgr`` in ``test_views.py`` and ``test_series_views.py``; it is restated here
+  because this is where the pair now lives.
+"""
+
+from dataclasses import replace
+from unittest.mock import MagicMock
+
+import pytest
+
+from mining_dashboard.service.metrics import Metrics, SyncMetric
+from mining_dashboard.web.series_views import _mode_palette, build_hashrate
+
+_SYNC_DONE = SyncMetric(
+    percent=100, current=10, target=10, remaining=0, has_target=True, done=True, down=False
+)
+
+_BASE = Metrics(
+    total_h15=10500.0,
+    p2pool_1h=8000.0,
+    p2pool_24h=8100.0,
+    xvb_1h=2100.0,
+    xvb_24h=2300.0,
+    xvb_routed_1h=2000.0,
+    xvb_routed_24h=2050.0,
+    stratum_h15=10300.0,
+    stratum_h1h=10400.0,
+    stratum_h24h=10200.0,
+    mode="P2POOL",
+    xvb_enabled=True,
+    current_tier="Donor (1.00 kH/s+)",
+    target_tier="Donor (1.00 kH/s+)",
+    target_threshold=1000.0,
+    target_sustainable=True,
+    low_hr_warning=False,
+    xvb_fail_count=0,
+    xvb_last_update=0,
+    workers_online=2,
+    workers_total=3,
+    shares_in_window=5,
+    pplns_window=2160,
+    block_time=10,
+    pool_type="Mini",
+    pool_hashrate=120_000_000.0,
+    pool_difficulty=250_000_000.0,
+    network_difficulty=380_000_000_000.0,
+    network_height=3210001,
+    global_syncing=False,
+    monero=_SYNC_DONE,
+    tari=_SYNC_DONE,
+    monero_mode="Unknown",
+    tari_mining=True,
+)
+
+
+@pytest.fixture
+def _metrics():
+    """Factory for the ``_metrics`` the four view-test modules each used to define."""
+
+    def _metrics(**over):
+        return replace(_BASE, **over)
+
+    return _metrics
+
+
+@pytest.fixture
+def _sync():
+    """Factory for the ``_sync`` the four view-test modules each used to define."""
+
+    def _sync(**over):
+        return replace(_SYNC_DONE, **over)
+
+    return _sync
+
+
+@pytest.fixture
+def _hashrate():
+    """Factory for the ``_hashrate`` the four view-test modules each used to define."""
+
+    def _hashrate(metrics):
+        """build_hashrate with palette tokens derived as build_state does."""
+        return build_hashrate(metrics, *_mode_palette(metrics.mode))
+
+    return _hashrate
+
+
+@pytest.fixture
+def _state_mgr():
+    """Factory for the ``_state_mgr`` the four view-test modules each used to define."""
+
+    def _state_mgr(
+        history=None,
+        mode="P2POOL",
+        share_stats=None,
+        blocks=None,
+        disk_growth=None,
+        xvb_history=None,
+    ):
+        sm = MagicMock()
+        sm.get_history.return_value = history or []
+        sm.get_xvb_stats.return_value = {"current_mode": mode}
+        sm.get_tiers.return_value = {}
+        sm.get_xvb_reward_estimates.return_value = {"estimates": {}, "last_update": 0.0}
+        sm.get_xvb_round_stats.return_value = {"stats": {}, "last_update": 0.0}
+        sm.get_share_stats.return_value = share_stats or []
+        sm.get_raffle_wins.return_value = []
+        sm.get_xvb_standby.return_value = None  # no backup standby held (#249)
+        sm.is_db_healthy.return_value = True
+        # #196 Tier-1 telemetry backbone exposure.
+        sm.get_blocks.return_value = blocks or []
+        sm.get_disk_growth.return_value = disk_growth or []
+        sm.get_xvb_history.return_value = xvb_history or []
+        return sm
+
+    return _state_mgr
+
+
+@pytest.fixture
+def _data():
+    """Factory for the ``_data`` the four view-test modules each used to define."""
+
+    def _data(**over):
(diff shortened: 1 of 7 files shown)


VijitSingh97:

Reviewed in its own worktree at `~/dev/pass-1540`, against `develop-v2` at `89c0678`. Ruling 8a
case: **BROAD** — branch tree `877f50f` and merge tree `6f6d706` differ, so both were gated
separately. Proof logs and the materialized merge commit are local to that worktree.

### PROVEN BY ME (re-derived, not relayed)

- **File budget — the risk I ranked first.** Four rows change for the four files the cut owns, so
  ruling 9 is satisfied: no row was lowered for a file this cut does not own. Three are lowered to
  the exact new line count (`test_infra_views.py` 865→768, `test_views.py` 861→775,
  `test_xvb_views.py` 1342→1243) — zero headroom, legal, and worth knowing before the next edit to
  any of them. `test_series_views.py`'s row is correctly **dropped**: at 378 it is under the
  400-line target, and keeping the row would have failed the gate's own rule 11. `conftest.py` at
  159 is correctly unbudgeted. Gate rc 0 at the **branch head** and rc 0 on the **materialized
  merge tree**.
- **"0 bodies changed" re-derived by an independent route.** Full-depth AST walk keyed by qualified
  name (tests live in classes), comparing `ast.unparse` of the body statements separately from the
  argument list: **0 bodies changed, 143 signatures changed, 148 identical, 0 added, 17 removed**.
  That reproduces the PR body's figures exactly from a different instrument. The comparator is shown
  able to fire — one seeded body edit makes it report `test_views.py::TestChartWindow.test_default_window_is_10m`.
- **The blind spot of that instrument, closed.** A function walk cannot see module-level state, so
  the top-level statements were diffed too. The only removals are the builder constants and the
  imports they alone needed; nothing else at module level moved.
- **Suite on the merged tree: 2151 passed, rc 0, coverage 97.03%.**
- **Silent capture: none.** Across the nine other modules under `tests/web/`, **0** tests take a
  parameter named after one of the five fixtures. Control fires at 127 on two of the touched
  modules, so the scan is not null.
- **The scope narrowing is correct.** `tests/service/test_telegram_commands.py`'s `_metrics` is
  byte-identical in source to the shared one, and its `_BASE` differs (`current_tier`/`target_tier`
  are `'Donor'`, not `'Donor (1.00 kH/s+)'`). The issue's stated gate — one definition per name —
  would have driven a consolidation that changed what that module asserts. Declining it was right.

### The attack I ran on this change, and why it failed

Moving `_SYNC_DONE` and `_BASE` from four private copies to one shared module constant is the kind
of consolidation that introduces cross-module pollution: one test mutating a shared object in place
leaks into three other modules, ordering-dependently, and a green suite is exactly what that defect
looks like until the day it isn't.

It cannot happen here. `Metrics` and `SyncMetric` are `@dataclass(frozen=True)`, so in-place
mutation raises rather than leaks, and both factories go through `replace()`. `_data` builds its
dict inside the factory and `_state_mgr` builds a fresh `MagicMock`, so neither returns shared
state. The hazard is closed by construction, not by convention.

### Docstring prose — checked by substance, not by count

Nineteen sentences leave the four module docstrings. Every one of them describes the duplicated
state this PR ends ("below are COPIES", "the master lives in `tests/web/test_views.py`",
"consolidating them into real fixtures is tracked as #1459"). They should die with it. The one
genuinely durable fact among them — that there is no import route between two test modules under
`--import-mode=importlib` — is now recorded nowhere, which is a small loss and not a blocker.

### One finding, non-blocking, and it is mine to discharge at close

`dashboard/tests/service/test_data_helpers.py:9` says `_totals` is duplicated and that "#1459
tracks consolidating them once byte-identity no longer binds". Base-checked per ruling 10a: that
line is present at `89c0678` and this diff does not touch the file, so it is **not** debris from
this cut. But #1459 as executed covers `tests/web/` only, so **closing it makes that pointer
false**. I will not close #1459 against an orphaned pointer — a successor issue for the
`tests/service/` duplicates is owed, and that docstring should name it instead.

### What I did NOT check

- `make lint-sh` — serialized here and the OOM path; this PR touches no shell file, and CI's Shell
  tests job passed on the head.
- Patch coverage. **#1537 is open**: `scripts/patch-coverage.sh` hardcodes `COMPARE=origin/develop`,
  so on a `develop-v2` PR its rc is no evidence either way. Separately, every changed file but two
  docs is a test file, so `diff-cover` here is NOT APPLICABLE — never PASSED.
- **Relayed, not re-derived:** the base-side 2151 count, and the 190-reference fixture-binding scan
  (the green suite on the merged tree subsumes it).

PR #1544, opened 2026-08-29
docs(dev): drop the run.sh count that goes stale within hours (#1105)

VijitSingh97:
## What

`docs/dev/repo-map.md`'s tier-1 row gave `run.sh` as "237 lines, down from
8,381". Both halves are wrong, in different ways.

**The current count rots faster than we can correct it.** The figure has held
three values in one day: 277 when #1445 introduced it, 237 after #1530 corrected
it, and 249 at this PR's base, after #1482 and #1246 landed. Nothing acts on the
number — the row's load-bearing claim is that `run.sh` is a thin runner that
sources every domain, and that claim survives every value the count takes. While
the file is still being cut, a precise count in prose generates defects at the
rate the file is touched, so the count goes and the structural claim stays.

**The baseline was the wrong branch's number.** 8,381 was `develop`'s figure when
the issue was filed. `develop` measures 6,710 today, and the cut happened on
`develop-v2`, where `run.sh` stood at 13,967 immediately before d379dab, its
first Phase 1 commit.

## Proof

Counts taken with `git show <ref>:tests/stack/run.sh | wc -l` and re-derived with
`grep -c ''` (249 both ways; the file ends in a newline, so `wc -l` is not
undercounting). The 13,967 baseline is `d379dab^`, the parent of the first
`#1105` commit to touch `run.sh`.

Swept the tracked tree for siblings: `8,381` appears at no other site, and the
only other file-size figures in `docs/` — 485 and 494 in
`config-layer-language.md` — match `28-parse-and-validate-config.sh` and
`33-render-env.sh` at this tip.

`make lint-md` and `make lint-docs-voice` are green (rc captured directly, not
through a pipeline). No shell or Python surface is touched.

Refs #1105.


VijitSingh97:
## Non-author pass (#1105) — controller lane, fresh seat

**Verified, ship it.** I am the successor seat on the authoring lane, so this is not a
different account — I am saying plainly what it is. What makes it worth more than a
re-read is that every claim below was re-derived at source in a fresh context, by a
different route than the body used, and one of them is now measurably stronger than when
the body was written.

**1. The `13,967` baseline is exact.** The body took it with
`git show <ref>:tests/stack/run.sh | wc -l`; I took it with `grep -c ''` on
`d379dab^` (`c0e14823`, the commit before Phase 1's first commit) and got **13967**.
Two counters, same number, so this is not a trailing-newline artefact.

**2. The premise is no longer an argument — it is an observation.** The line being
removed says `237 lines`. `tests/stack/run.sh` at the `develop-v2` tip measures **249**.
The figure was already wrong before this PR was merged, which is precisely the defect the
change describes. A count in prose about a file that is still being cut rots faster than
review latency.

**3. The structural claim that stays is true at the tip.** run.sh is a thin sourcing
runner: one `source "$HERE/lib.sh"`, then one guarded stanza per domain of the form
`_d0=$((PASS + FAIL)) && source "$HERE/test-<domain>.sh" && domain_ran ... || domain_ran ...`,
which is the guard the row credits to #1400 — a domain that fails to load is named rather
than silently skipped. 108 non-blank non-comment lines. The row's load-bearing sentence
survives every value the count takes, so dropping the count loses no information.

**A method note against myself, since it nearly became a finding.** My first probe for the
sourcing stanzas was `grep -E '^\s*source .*/test-'` and it returned **0**. That is a null
instrument, not evidence of a missing structure — the stanzas do not begin with `source`,
they begin with `_d0=`. I read the file rather than reporting the zero. A check that never
matched cannot discriminate.

**4. The sibling sweep in the body reproduces.** `8,381` and `237 lines` each appear at
exactly one site in the tracked tree at `origin/develop-v2` — the line this PR edits. The
comma-less `8381` hits are all `dashboard/uv.lock` wheel sizes and hashes, unrelated.

**What I did NOT check:** I did not re-run `make lint-md` or `make lint-docs-voice`
locally; I took the body's green claim on report, and CI is green on this PR (15 success,
2 skipped, 0 failing). For a one-line docs edit whose every factual claim I re-derived,
that is a proportionate gap and I am naming it rather than hiding it.

**Merge ordering, recorded because it is not visible in the diff:** this PR touches no
`docs/dev/file-budget.tsv` row, so it carries stale-high values for the four
`dashboard/tests/service/*` rows that #1545 lowers. It is being merged BEFORE #1545 for
that reason, not by accident.

Issue #1015, opened 2026-08-15
Mark config applies and rig upgrades on the per-worker hashrate chart — a step in the line has no visible cause


VijitSingh97:
## Problem / motivation

Follow-on to #1013 (per-worker hashrate chart in Worker Inspect). A line alone shows *that* a rig
stepped up or down; it doesn't show *why*. The two things most likely to have caused the step are
the two things the operator did from this very panel: a config apply, and a rig upgrade.

Without markers, the panel puts the cause (the change-history table) and the effect (the chart) in
two places and leaves the operator to line up timestamps by eye.

## Proposed solution

Marker series on the per-worker chart, in the shape the fleet chart already uses. `chart.mjs`
already draws timestamped scatter series on their own stack group so their y stays absolute —
Events (#99), raffle wins, payouts. This is a fourth instance of that pattern, not a new mechanism.

- **Config applies** — from `worker_config (worker, ts, status, changes)`, already recorded and
  already rendered as the panel's history table. Tooltip: the changed keys, which the row carries.
- **Rig upgrades** — blocked: nothing records a per-rig upgrade timestamp today. See the sibling
  issue; this half lands after it.

Two glyphs or one with a typed tooltip is a design call. Worth deciding at the same time: whether
`rejected` / `rolled_back` rows get a marker at all — nothing changed on the rig, so a marker on
the hashrate line would be misleading, but "we tried and it bounced" is exactly what an operator
wants to see next to a flat stretch. Suggest a distinct, quieter glyph rather than dropping them.

## Alternatives considered

- **Annotate the fleet chart instead.** Wrong altitude — per-rig events on a fleet line is noise,
  and the fleet line doesn't move much when one of N rigs changes.
- **Highlight the history-table row on hover over the chart.** Cheaper, keeps the two panes in
  sync, but still leaves the operator scanning a table to find what happened at 03:00.

## Context

Depends on #1013. Related: #99 (event markers on the fleet chart), #185 (Worker Inspect),
#492 (hashrate by config version), #597 (one-click rig upgrade).

Issue #1487, opened 2026-08-29
An error and an answer share one return channel in 32 read handlers — can a mechanism make that impossible?


VijitSingh97:
## The question

`mining_dashboard` has a recurring shape: **a read handler whose error path and whose empty answer
share one return channel.** The method returns `[]` (or `{}`, or `0`) both when the read failed and
when the read succeeded and found nothing, so no caller can tell the two apart — and the caller that
matters renders the empty answer as a positive claim.

This issue asks a mechanism question, deliberately not a work item:

> **Can a helper, a sentinel type, or a lint make it impossible in this package for an error and an
> answer to share a return channel — or at least make each instance a decision someone had to make
> on purpose?**

The enumeration below is EVIDENCE for the mechanism, not a list of thirty hand-edits to schedule.
Fixing them one at a time is the denylist-lengthening move: it makes today's known instances safe
and does nothing about the next one, and each hand-edit costs a caller-side rendering decision to
design and defend (see the four call sites in the #1409 PR — only one of the four was worth changing
under that PR's scope).

## Why now, and what it already cost

#1409 is one instance of this shape, found in production behaviour rather than by reading:
`get_worker_config_history` returned `[]` on both failure paths. Its caller reads that list to decide
where a rig's running config came from, and an empty list means *"we hold no row for this rig's
`last_change_id`"*, which is the `elsewhere` verdict, which renders as **"Last changed from another
dashboard."** So a database that failed to open produced an on-screen accusation that another
operator had changed the rig.

Two properties of that instance are why the class deserves a mechanism rather than a sweep:

- **The two failure routes were not equally observable.** One raised and logged (`sqlite3.Error`);
  the other — no connection at all — raised nothing and logged nothing. A reader looking directly at
  the running system could not see it.
- **`docs/dashboard.md` had written the defect down as a safety property**, describing a history
  "it cannot read at that moment" as landing on the cautious answer. It was the reassuring direction
  on paper and the accusing direction in the code, which is part of why it survived being read.

It is also the same class #1369 removed by a different route, and the same class as COMMON's
`unknown`/`timeout` finding: *two states, one output, and the output is the reassuring one.*

## The evidence — how widespread, and by what rule

Two counts, each with its rule stated, because a bare number in this family has been wrong three
times running in this lane. Both are AST-derived over `dashboard/mining_dashboard/**/*.py` at
[...]

PR #1546, opened 2026-08-29
fix(dashboard): stop keeping and serving the pool password in the worker change record (#1543)

### LINKED ISSUE #1543: Worker Inspect: last_applied is served unmasked, so an applied pool password is stored and returned in plain text
## The gap

`last_applied` is served in the Worker Inspect payload **unmasked**, and `pools` is on the writable
allowlist. A pool password applied through the Inspect editor is stored in `worker_config.changes`
as plain text and served straight back on every load of that rig's Inspect view.

`mask_secrets` is not applied on this path. It is applied to the host config
(`data_helpers.py:375`, and `control_service.py:287`), so the posture exists — this path is the one
that does not take it.

## The route, traced at source

- `handle_worker_apply` (`web/server.py`) takes `changes` off `await request.json()` and passes it
  to `_record_worker_result` unchanged. `validate_worker_changes`
  (`service/control_service.py:380`) checks key **names** against `WORKER_WRITABLE_KEYS` and never
  inspects a value.
- `pools` is in `WORKER_WRITABLE_KEYS`, and a pool entry carries `pass`.
- `get_last_applied_worker_config` (`service/worker_config_store.py:213`) merges those rows with
  `merged.update(row["changes"])`, uncoerced and unmasked.
- `worker_detail.py` publishes the merge as `last_applied`, and `workerview.mjs` renders it into
  both the table editor prefill and the JSON textarea.

The rig's own side is defended: `_rig_writable_config` runs `strip_credentials` over `rig_config`
(`client/xmrig_client.py`), with a comment explaining the defence-in-depth reasoning. The
dashboard's own record of what it applied gets no equivalent pass.

## Confidence, stated precisely

**The code path allows it.** Traced at source, above.

**No instance exists in the one real database I can read.** I checked all 21 `worker_config` rows
for the presence of the `pass` and `tls-fingerprint` keys inside stored `pools` entries — key
presence only, never values — and found none.

So this is an **unfired hazard, not an observed leak**. I am not claiming anyone's credential is
currently exposed.

## Why it is worth closing anyway

The value is stored, so the exposure outlives the moment: it is served on every subsequent load, it
survives a restart, and it sits in the DB and in any backup of it. The dashboard already treats a
pool password as a credential everywhere else — the rig strips it before serving, the client strips
it again on arrival, and the Configuration view masks its own secrets with the `{__secret__: true}`
sentinel `buildFields` already knows how to render. The sentinel machinery this needs is built.

## Not verified

Whether masking `last_applied` breaks the editor prefill in a way the secret-row path does not
already handle. `buildFields` turns a sentinel into a `secret` row and `buildTableChanges` keeps the
masked value on a blank edit, which looks like the right shape — but I have not run it, and the
apply path's round-trip on a masked pool entry is the thing to check before building.

Found while building #1367, which is unaffected: `config_drift` strips both sides, so the drift
note cannot carry a credential even though the field beside it can.

VijitSingh97:
Closes #1543.

`pools` is on the worker writable allowlist and a pool entry carries `pass`, so an operator's POSTed `changes` was stored verbatim in `worker_config.changes` and served back on every load of that rig's Inspect view — through a restart, and into any backup of the DB.

## The remedy the issue names is a no-op, and the issue is mine

I wrote that body. It says `mask_secrets` "is not applied on this path", which is true, and invites a fix that changes nothing while looking correct. Measured with controls and posted as comment 5464995965: every `SECRET_PATHS` entry is rooted in the **host** config, a `last_applied` merge is `WORKER_WRITABLE_KEYS` only, and **the two root sets do not intersect at all**. A test asserting `mask_secrets` was called here goes green with the password still on the wire.

The instrument that works is `strip_credentials`, already used on the rig-read path. It drops `pass` and `tls-fingerprint` at any depth in any container shape, and **deletes** rather than substituting a sentinel — which is what a prefill that can be POSTed back needs, since an operator saving a sentinel-filled form would write the sentinel over the real password.

## Where the strip goes was the whole decision — and the issue points at the wrong place

**The Inspect payload carries a row's `changes` in THREE fields, not the one the issue names:** `last_applied` (the merge), `history` (the rendered change table), and `hashrate_history.markers[]` (the chart tooltips, #1015).

A fix at `get_last_applied_worker_config` — where the issue's own text points, and where my posted comment ranked it first — would have cleaned the named field and **left the other two serving the credential, with nothing going red.** That is why `_shaped` is the chokepoint rather than a stylistic preference: it is the one point both row-returning reads (`get_worker_config_history` and `get_worker_config_change`) pass through, so it is the single edit that covers every consumer, including ones nobody has written yet.

Two call sites, closing different halves:

- **`add_worker_config_version`** strips before the INSERT, so a credential stops landing in the DB. Its docstring claimed *"no secret ever lands here"* — a property it did not have, stated as if it did, which is part of why the gap survived being read. Corrected in place rather than deleted.
- **`_shaped`** strips on the way out. This is what makes a row an **older build already wrote** safe: those rows stay in the table under any fix, so without this half the disclosure survives a deploy. No migration needed.

The password still reaches the rig: `handle_worker_apply` sends the operator's unmodified `changes` before recording, and `strip_credentials` builds new containers rather than mutating, so the caller's dict is untouched. What changes is only what this dashboard keeps afterwards.

## ⛔ A disclosed conflict CI cannot see — read before merging

`tests/integration/rigforge-writable-keys.sh:176` restores a borrowed rig's pools from `.last_applied.pools`, and its own comment at line 182 says it does so **because that value still carries `pass`**. `tests/integration/selftest-rigforge-writable-keys.sh:307` asserts `"the restore uses last_applied, credential intact"`. This change removes that credential.

The failure mode is the worst one available. The leg's skip triggers on pools being **empty**; mine is non-empty and merely password-**less**, so it does not skip — it applies a password-less pools config to a real miner, which `docs/dev/integration-testing.md` says the proxy rejects a login without.

**CI cannot see this.** `selftest-rigforge-writable-keys.sh` overrides `_worker_detail` to echo a hardcoded `STUB_DETAIL` string, so it never reaches a real dashboard and stays green either way. Blast radius is narrow: the leg only runs when an operator sets `IT_RIG_POOLS_PROBE` against a real rig, so it is opt-in rather than every e2e run.

**The finding inside the finding, and the strongest argument for this fix:** the harness restores from `last_applied` *because the credential is stored there in plaintext*. The defect had become load-bearing for a test. That is what a plaintext credential does if it sits long enough, and it is a better justification than the hazard being unfired.

Routed to the controller, which is sequencing the harness change first: the leg's guard moves from "pools is empty" to "pools carries a usable credential" — testing `pass` explicitly. That predicate gap is a defect in the leg independent of this PR, and fixing it removes the ordering dependency entirely. `IT_RIG_POOLS_RESTORE`, symmetric with the probe the leg already requires, then follows as the real fix. `tests/integration/` is not this lane's path.

## Evidence

- **Full suite: 2201 passed, 0 failed.** No existing Python test depended on the credential being retained.
- **12 new tests**, service and web tier. The load-bearing web assertion scans the **whole serialized payload** rather than named fields, so it does not depend on my having enumerated the surfaces correctly — if a future change publishes `changes` through a fourth field, it fails and the per-field tests do not.
- **Both seeds are controlled.** Each module has a test proving the seeded row really carried the credential, because a strip test whose fixture never held one passes for the wrong reason. Narrowness assertions (`url`/`user`/`keepalive` survive) are load-bearing too: a strip returning `{}` would pass every "the password is gone" assertion.
- **Seeded-defect battery 2/2 KILLED** (`mutate-1543.py`): removing the read-side strip reds 7 tests, removing the write-side strip reds 2. Both halves are independently necessary and independently proven. The harness refuses a pattern that does not match exactly once, proves the mutation is on disk before running, and verifies the restore by sha256.
- **Coverage:** dashboard total 97% (bar 80%). All three added executable lines are outside the coverage-missing set, cross-checked mechanically against the diff rather than by eye. I did **not** run `make test-patch-coverage` — #1537 is open, so its rc is vacuous on a develop-v2 PR; saying which one I ran matters more than the rc.
- `ruff check` / `ruff format --check` clean, `make lint-docs-voice` clean, `make lint-file-budget` clean. All three touched files are under 400 lines and correctly take no budget row. `make lint-sh` skipped — no shell files in the diff.

## Disclosed consequence

The strip stops walking past `_MAX_CONFIG_DEPTH` (6) and returns `None` below it, so a `changes` nested deeper than that is **recorded** truncated. A writable config is three deep at most (`pools` → a pool → its fields), so nothing a rig accepts reaches the bound — but the audit row, not the rig, is what would lose.

Rows already written keep their plaintext at rest; the read-side strip stops them being served. The issue's own check of the one real database found zero instances, so a scrubbing migration would rewrite something demonstrably absent. Filing one is a reasonable follow-up if belt-and-braces is wanted.

## What the over-engineering pass changed

The PR-create gate asked for a review for over-engineering, so I ran one on my own diff rather than clearing it by pressing the button twice. It found one real thing and I cut it: a three-case parametrized class re-proving that `strip_credentials` is shape-agnostic (dict-shaped `pools`, a credential nested one level deeper, one under a different key). That behaviour is **already** covered at the function's own unit — `tests/client/test_rigforge_config.py:64` tests those exact shapes plus a depth-bomb — so re-testing it through a DB round-trip proved the same behaviour at a more expensive tier, against this repo's "test each behaviour once, at the lowest tier that proves it honestly" rule. Removing it took the battery's write-side row from 5 kills to 2, which is exactly the three tests removed and confirms they carried no independent signal. A pointer to where shape-agnosticism *is* proved now sits in the test module's docstring, so a reader does not conclude it went unproved.

I kept two things a stricter pass might also flag, with reasons. The **two** strip call sites are not redundant: read-side alone leaves plaintext at rest and in backups, write-side alone leaves rows an older build wrote — the battery kills each independently. The **per-field** web tests are partly redundant with the whole-payload scan, but they say *where*, so a future regression names itself instead of only failing a blanket assertion.

## Still owed

**An independent security pass.** This is the credential surface and I cannot self-serve it. A controller pass is explicitly not independent under DECISIONS item 14.

### DIFF
--- a/dashboard/mining_dashboard/service/worker_config_store.py
+++ b/dashboard/mining_dashboard/service/worker_config_store.py
@@ -25,6 +25,15 @@ import sqlite3
 import time
 from typing import Any
 
+# The credential strip the rig-read path already uses on ``rig_config``
+# (``client/xmrig_client._rig_writable_config``), reused here rather than restated: a second copy
+# of the key list and the depth walk is a second place for them to drift, and this table feeds the
+# SAME editor prefill that one defends (#1543). The service -> client direction is the one
+# ``data_helpers``, ``worker_refresh`` and ``data_service`` already take; ``xmrig_client`` reaches
+# back only as far as ``control_service``, which imports ``config`` and nothing else, so this
+# closes no cycle.
+from mining_dashboard.client.xmrig_client import strip_credentials
+
 # The full terminal vocabulary the rig's control mirror can report (#1009) — applied/rejected/
 # rolled_back/failed from a control-apply, plus noop (already on target)/throttled (retry-later)
 # from a control-upgrade (rigforge#320). Mirrors xmrig_client._CONTROL_TERMINAL and pithead's own
@@ -41,13 +50,26 @@ _SELECT_CHANGE = "SELECT change_id, ts, status, changes, reason, type FROM worke
 
 def _shaped(row: sqlite3.Row) -> dict:
     """One ``worker_config`` row as the rest of the app reads it: ``changes`` parsed back to a
-    dict (unparseable JSON reads as ``{}`` rather than raising), and a row written before the
-    ``type`` column existed (#1014) reading back as ``"apply"``."""
+    dict (unparseable JSON reads as ``{}`` rather than raising), credentials stripped out of it,
+    and a row written before the ``type`` column existed (#1014) reading back as ``"apply"``.
+
+    The strip is HERE, and not at any of the three places that publish ``changes``, because this
+    is the one point both row-returning reads pass through — ``get_worker_config_history`` and
+    ``get_worker_config_change`` — and therefore the only single edit that covers every consumer
+    of them (#1543). The Inspect payload alone carries a row's ``changes`` in three separate
+    fields: ``last_applied`` (merged here), ``history``, and ``hashrate_history.markers[]``.
+    Stripping at ``get_last_applied_worker_config``, which is where the issue's own text points,
+    would have left the other two serving the credential, and nothing would have gone red.
+
+    It is also what makes an ALREADY-WRITTEN row safe: ``add_worker_config_version`` stops new
+    rows carrying a credential, but rows a previous build wrote still hold one, and this is what
+    keeps those from being served without needing a migration to rewrite them."""
     d = dict(row)
     try:
         d["changes"] = json.loads(d["changes"]) if d["changes"] else {}
     except (TypeError, ValueError):
         d["changes"] = {}
+    d["changes"] = strip_credentials(d["changes"])
     d["type"] = d.get("type") or "apply"
     return d
 
@@ -67,8 +89,23 @@ class WorkerConfigStoreMixin:
     ) -> None:
         """Record one applied/attempted worker change (#185): a config apply, or (``change_type=
         "upgrade"``, #1014) a one-click RigForge upgrade attempt — ``changes`` then carries
-        ``{"version": ...}`` instead of a writable-key diff. Stored as JSON — no secret ever lands
-        here. Forward-only."""
+        ``{"version": ...}`` instead of a writable-key diff. Forward-only.
+
+        Stored as JSON with the pool credentials stripped OUT of it first (#1543). That sentence
+        used to read "no secret ever lands here", which was not a property this method had: the
+        caller hands us the operator's own POSTed ``changes``, ``pools`` is on the writable
+        allowlist, and a pool entry carries ``pass``, so the credential landed here in plain text
+        and stayed — through a restart, and into any backup of the DB.
+
+        The strip is on the RECORD only. ``handle_worker_apply`` sends the operator's unmodified
+        ``changes`` to the rig before calling this, so the password still reaches the pool it is
+        for; what changes is what this dashboard keeps afterwards. ``strip_credentials`` builds
+        new containers rather than mutating, so the caller's dict is untouched either way.
+
+        One disclosed consequence: the strip stops walking past ``_MAX_CONFIG_DEPTH`` (6) and
+        returns ``None`` below it, so a ``changes`` nested deeper than that is recorded truncated.
+        A writable config is three deep at most (``pools`` -> a pool -> its fields), so nothing a
+        rig accepts reaches the bound — but the audit row, not the rig, is what would lose."""
         try:
             with self._db_lock:
                 if not self._conn:
@@ -81,7 +118,7 @@ class WorkerConfigStoreMixin:
                         change_id,
                         ts if ts is not None else time.time(),
                         status,
-                        json.dumps(changes),
+                        json.dumps(strip_credentials(changes)),
                         reason,
                         change_type,
                     ),
--- a/dashboard/tests/service/test_worker_config_credentials.py
+++ b/dashboard/tests/service/test_worker_config_credentials.py
@@ -1,0 +1,141 @@
+"""Pool credentials must not survive into the ``worker_config`` record, or out of any read of it
+(#1543).
+
+``pools`` is on the worker writable allowlist and a pool entry carries ``pass``, so the operator's
+own POSTed ``changes`` used to be stored verbatim and served back on every load of that rig's
+Inspect view. The defence is in two places and this module pins both, because they close different
+halves and only together close the issue:
+
+* ``add_worker_config_version`` strips before the INSERT, so a credential stops landing in the DB
+  (and therefore in any backup of it) at all. The raw-column assertions below are the only ones
+  that can see that half -- every reader would look clean under the read-side strip alone.
+* ``_shaped`` strips on the way back out, which is what covers a row an OLDER build already wrote.
+  Those rows are still in the table under any fix, so without this half the disclosure survives a
+  deploy. ``_seed_legacy_row`` writes that shape on purpose, by going around the store's own
+  writer, and ``test_the_legacy_seed_really_carries_the_credential`` is the control proving the
+  seed armed -- a strip test whose fixture never held a credential passes for the wrong reason.
+
+``strip_credentials``' own shape-agnosticism is NOT re-proved here: it is already covered at the
+function's own unit (``tests/client/test_rigforge_config.py``, dict-shaped ``pools``, a credential
+nested one level deeper, and a depth-bomb). Re-testing it through a DB round-trip would prove the
+same behaviour at a more expensive tier.
+
+The narrowness assertions (``url``/``user``/``keepalive`` still present) are load-bearing rather
+than decorative: a strip that returned ``{}`` would pass every "the password is gone" assertion in
+this file.
+"""
+
+import json
+
+_PASSWORD = "s3cret-pool-password"
+_FINGERPRINT = "aa:bb:cc:dd"
+
+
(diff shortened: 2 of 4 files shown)


VijitSingh97:
## Non-author pass — and it is NOT INDEPENDENT, which changes what it is worth

I routed this fix and I wrote the ruling that chose its shape (the harness-first sequencing, and
the decision not to fix at `get_last_applied_worker_config`). So this pass cannot be a second
opinion on the design — it is a pass **on merit**, by a seat that already has a stake in the
answer. Read the findings below on their own evidence, not on my agreement with the approach.
Saying this in the pass itself is the whole point of the correction filed against my own seat on
#1542.

**Verdict: the change is correct and the placement is right. One undisclosed regression, filed
rather than blocking. The merge hold stands and it is a SEQUENCING hold, not a review hold.**

## PROVEN BY ME, at source, at the PR head

**1. The chokepoint claim is true, and I checked it the only way that can falsify it — by
enumerating every reader, not by confirming the three the body names.** `_SELECT_CHANGE` is the
only query in `worker_config_store.py` that selects the `changes` column at all. Both
row-returning reads pass through `_shaped` (`get_worker_config_history:143`,
`get_worker_config_change:211`). `worker_config_change_known` selects `1` and reads no row.
`reconcile_worker_config_status` is an `UPDATE` that never touches `changes`.
`get_last_applied_worker_config` is built **on** `get_worker_config_history`, so it inherits the
strip rather than needing its own. There is no fourth path, so `_shaped` genuinely is the single
edit that covers every consumer — including the two the issue's own text would have missed.

**2. No import cycle.** `service/worker_config_store` -> `client/xmrig_client` ->
`service/control_service`, and `control_service` imports exactly one thing:
`mining_dashboard.config.config`. The cycle the body says it does not close, it does not close.

**3. `strip_credentials` is unchanged by this PR** — it is not among the four files — so the
rig-read path's strip is pre-existing, and reusing it rather than restating the key list is the
right call for the reason the new comment gives.

**4. The depth bound is disclosed accurately and is not reachable by a real config**
(`pools` -> a pool -> its fields is three, the bound is six). The disclosure that the audit ROW,
not the rig, is what would lose is the correct framing.

## THE FINDING — an operator-facing regression this PR introduces and does not disclose

**JSON mode's Apply was lossless before this change and is lossy after it.** `workerview.mjs:162`
prefills the Advanced/JSON textarea from `detail.last_applied` **alone** — no rig precedence — and
`parseJsonChanges` (`workerlogic.mjs`) filters by top-level key membership and does nothing else.
So an operator who opens Inspect, switches to JSON mode and presses Apply without touching `pools`
now POSTs a `pools` block with no `pass` in it.

That is the same failure this PR's own body describes for the integration leg, in the product:
a non-empty, password-**less** pools applied to a real rig. The body discloses the harness
instance and not this one.

**Why it is a finding and not a blocker — the honest version.** Table mode already had this loss
before the PR, and I confirmed the mechanism rather than assuming it: `buildFields` takes the
**rig's** value first (`const fromRig = key in rig`), and `rig_config` comes from
`_rig_writable_config`, which already calls `strip_credentials` on `develop-v2` today. So for any
reachable rig the table editor has always shown a credential-less pools row. This PR does not
introduce the class; it removes the last path that still round-tripped. Holding a fix that stops
storing plaintext credentials in order to first fix a round-trip gap that already ships on the
dominant path would leave the credential in the DB and on the wire — strictly worse.

**The end state, and it already exists one view over.** `docs/dashboard.md`'s own host-config
paragraph — six lines above the one this PR adds — describes exactly the right mechanism: a blank
password field, blank keeps the stored value, typed replaces it. `workerlogic.mjs` even has the
machinery (`isSecretSentinel`, `type: "secret"`, "blank = leave the masked value alone"). It
cannot reach this case because `buildFields` tests the sentinel against the whole **top-level**
value, so a `pass` nested inside a pool entry can never become a secret row. **The gap is not that
nobody thought of it; it is that the existing mechanism cannot express a nested credential.**
That is the same shape as ruling C's harness guard and as #1487: an instrument that cannot say the
thing it needs to say. Filing it.

**Doc scope.** The added `docs/dashboard.md` paragraph says "The password still reaches the rig; it
travels with the change itself." True of the apply path, and it is what a reader will take as
covering the editor too. It should say what it does not cover once the follow-up is filed.

## TAKEN ON REPORT — not verified by me

Every suite, coverage and battery figure (2201 passed, 97%, mutation 2/2 killed, 12 new tests) —
I read the assertions' shape and the controlled-seed reasoning, not a run. The claim that a rig
replaces `pools` wholesale rather than merging per pool field, which is what makes the finding
above bite, is **this PR's own assertion reused** — I have not read RigForge's apply, and the
finding's severity depends on it. The `EXPLAIN QUERY PLAN` results in the docstrings.

## THE HOLD — sequencing, and nothing is owed by this lane

Merge is still held on the ruling I posted: `tests/integration/rigforge-writable-keys.sh`'s guard
must move from "pools is empty" to "pools carries a usable credential" first. That is the e2e
lane's path, that lane is not running, and the spawn is not this seat's call — it is raised.
**Nothing on this PR is waiting on dashboard.** When the tip moves under it, note that this branch
carries no `docs/dev/file-budget.tsv` row at all, so it holds no stale-high value and survives the
move — but its checks ran against the previous base and should be re-read, not assumed, at merge
time.

Issue #35, opened 2026-06-01
Don't start p2pool and xmrig proxy until after monero and tari have synced fully


VijitSingh97:
## Problem
p2pool (and xmrig-proxy) start immediately alongside monerod/tari. Before the nodes are synced, p2pool spams the **Tari logs** with junk/retry messages, which makes log debugging painful. Not strictly a bug, but noisy.

## Proposed
Gate the start of p2pool and xmrig-proxy on monerod (required) and tari (merge-mining) being **synced**, rather than just "container up":
- Compose `depends_on` only waits for container start/health, not chain sync — so this likely needs a readiness gate keyed on sync status. The dashboard already computes `is_monero_syncing` / `is_tari_syncing` in `service/data_service.py`; monerod sync is derivable from `get_info`.
- Option: an init/wait step, or a healthcheck that only reports healthy once the chain is synced, before p2pool/xmrig-proxy start.

## Nuance
- Tari is treated as **optional** — gating on Tari sync must respect that (don't block mining forever if Tari is down or never finishes).

VijitSingh97:
**Related issues**
- #51 — both about sync behavior; Tari handling should respect that Tari is optional.
- #31 — per-node node-down handling / the same “Tari is optional” philosophy.

PR #1547, opened 2026-08-29
fix(hugepages): bound the appliance's HugePages pool with a declared ceiling (#1103)

VijitSingh97:
Bounds the appliance's HugePages pool with a declared ceiling, so a boot-time race cannot leave a
16 GiB machine holding 11 GiB of huge pages until the next reboot.

## The defect

The miner's config hands RigForge the stack's reservation as headroom, and RigForge sizes the pool
as `current + required - available`. A co-resident holding pages at that instant is subtracted from
`available` while the same pages already sit inside `required` — counted twice.

It is a **race, not a constant**. p2pool holds 1296 pages for about six seconds while allocating its
RandomX dataset; the #35 sync gate then stops it and it releases. Setup landing inside that window
takes the pool to 5618 pages and it stays there until reboot; outside it, 4322. Two consecutive
boots of one image gave both. A ceiling bounds it whoever wins; re-ordering the units would fix the
common case and leave the failure intact under load, so that was rejected.

## What ships

`render_local_miner_config` declares `hugepages_pool_ceiling_mb: 9216` behind **three** gates:

1. **Full tier only.** The reduced tier still refuses co-location (measured: lifting it asks 12.0 GiB
   on a 7.76 GiB box, the kernel grows the pool 2560 -> 2916, `MemAvailable` falls to ~40 MB, and
   RigForge exits 0 reporting "Deployment Complete"). Released tier holds no stack pages, so it has
   nothing to double-count and no measurement behind any value — it stays uncapped.
2. **RigForge >= 1.16.0.** rigforge warns and never errors on an unknown key, so declaring the
   ceiling into an older tree fails **open** — a config that reads as bounded and behaves as before.
   Verified in both directions at the pinned ref: 1.16.0 carries the key (`rigforge.sh:589` parse,
   `:691` known-keys, `:1520` the cap); 1.15.2 has zero occurrences.
3. **Detected NUMA nodes == 1.** See below — this one was a review finding.

## The NUMA gate (review finding, and the reason this PR is not the first draft)

The first draft froze 9216 with a comment deriving it as `1168*NUMA + THREADS + 50 + (extra+1)/2 =
1168 + 6 + 50 + 3072 = 4296`. **That substitutes `NUMA=1` away in the same sentence that writes the
term.** At the pinned ref `proposed-grub.sh:100-101` is
`TOTAL_2MB_FALLBACK=$(((BASE_2MB_PAGES * NUMA_NODES) + THREADS + 50 + EXTRA_2MB_PAGES))`, and
`EXTRA_2MB_PAGES` (L85-88) carries no node term — so the node term is the only one that scales:

| nodes | required | vs a 4608-page ceiling |
|---|---|---|
| 1 | 4296 pages (8,592 MB) | ceiling binds only under contention — correct |
| 2 | 5464 pages (10,928 MB) | caps a **healthy** box 856 pages short |
| 4 | 7800 pages (15,600 MB) | short by 3192 pages |

All three rows are the **supported X5690**, so they differ in the node term alone: `THREADS` is
`L3_MB/2` = 6 on its single 12 MiB L3, and `EXTRA_2MB_PAGES` is `(6144 + 1) / 2 = 3072`. Row 1's
8,592 MB is the `>= 8592` bound named in the shipped comment, which is the arithmetic's own check.

**An earlier cut of this table put the bench guest's measured 4322 in row 1** while rows 2 and 4
stayed on the X5690 — one table, two machines, and no tell, because every row supports the same
conclusion. 4322 is a **KVM artifact**: 4 vCPUs present as 4 sockets, L3 reads 64 MiB, THREADS
resolves 32. `lib/pithead/14-local-miner.sh` says so at the constant and calls it "wrong to quote
as the requirement" — the shipped files were right and the table was not.

RigForge caps the write and continues, so that under-reservation would be **silent** — the same
failure direction gate 2 exists to prevent. `local_miner_numa_nodes` therefore mirrors RigForge's own
precedence **step for step**: `lscpu` "NUMA node(s)", then sysfs `node[0-9]*`, then the **socket
count**, then 1. The socket step is load-bearing, not padding: the bench guest reports
**Socket(s)=4 with NUMA node(s)=1**, so a mirror stopping at sysfs would read 1 where RigForge read
4 — the defect back, quietly.

A render-time recomputation of RigForge's formula was considered and **rejected**: it puts a copy of
their sizing inside pithead that must stay correct across their releases, and it needs `THREADS`
(`L3_MB/2`, then `THREADS_CAP`) which pithead does not model. Freezing the value and documenting
"single node only" was also rejected — a constraint nothing enforces.

## Bench evidence — what was RUN

Full-tier 16 GiB KVM guest, `tests/os/run.sh --phase provision --keep`, **rc=0, 34 passed / 0 failed**.

**Treatment (ceiling present)** — produced by the boot path under a *real* race, not an armed hold:

```
Hardware-optimized HugePages: 4322 (2MB pages) calculated.
HugePages requirement (5618 pages) exceeds the declared pool ceiling —
  capping the write at 4608 pages (9216MB) instead of growing further (#398).
sysctl -w vm.nr_hugepages=4608   ->   HugePages_Total: 4608
```

`5618 - 4322 = 1296`, exactly p2pool's hold. RigForge quoting `9216MB` back is the second witness
that the key was **honoured**, not merely present.

**Control (key stripped, same boot, same 1296-page hold, one variable moved):** predicted 5618 in
advance, because RigForge had already printed it. Result `vm.nr_hugepages = 5618`, pool landed
**5618**.

Controls, all fired: hold delta asserted **exactly 1296** (`fallocate`, never `dd` — hugetlbfs has
no `write(2)`, and `dd` leaves a 0-byte file that `ls` shows as present); the independent variable
**read** (`ABSENT` vs `9216`), not assumed; the request taken from RigForge's own stdout, never a
readback — the kernel rewrites `vm.nr_hugepages` to what it achieved, so any post-hoc readback says
"granted in full" by construction.

### Disclosed: the legs ran OUT of my own protocol's order, and why that is harmless

My protocol required a reboot between legs ("a pool only grows"). In the event, the treatment landed
at boot with `current` = 3072 and the control ran afterwards with `current` already 4608. That is
the condition the reboot rule existed to prevent, so it needs stating rather than leaving for a
reviewer to find.

It does not bite, for a structural reason. With a hold of `H` and `_hugepages_avail` = free + pages
already held by the miner, `avail = current - H`, so:

```
target = current + required - avail = current + required - (current - H) = required + H
       = 4322 + 1296 = 5618        # current cancels
```

That identity is not a coincidence of one run — it holds **by construction**. At the pinned ref
`_miner_held_hugepages` (`rigforge.sh:1482`) reads `HugetlbPages` from `/proc/<MainPID>/status` of
the **miner service only**, and `_hugepages_avail` is `free + held`. So for any holder that is not
the miner service, `free = current - H - miner_held` and `avail = free + miner_held = current - H`
exactly, with nothing measured. The `fallocate` file qualifies, and so does p2pool in the field —
which is why the natural race and the manual hold land on the same 5618 by two different mechanisms.

**The premise this rests on, stated so it can be checked: the holder is not the miner service.**
When the holder *is* the miner, `avail = current`, `target = required`, and there is no over-reserve
to bound — coherent, since the defect is by definition a co-resident holder, but a real premise
rather than a technicality. **The run itself rules out the one way it could have failed:** had the
hold been counted as miner-held, `avail` would have been 3272 + 40 + 1296 = 4608 = `current` and the
target would have been 4322, not 5618. The measured 5618 is therefore a negative control on the
premise, obtained for free.

Verified against the logged numbers: `current` 4608, `free` 4568 -> 3272 after the hold, miner
holding 40, so `avail` = 3272 + 40 = 3312 = 4608 - 1296. **The delta-1296 assertion turns out to
license the analysis, not just the fixture** — it was armed against the `dd` trap and ended up
carrying an inference nobody had in mind when it was written. **The reboot rule was sufficient but
never necessary** — it guarded a dependence the arithmetic does not have.

### What this evidence does NOT cover

The guest is NUMA=1 by construction, so **the pair cannot exercise the NUMA axis at all**. That axis
is covered only by unit cases. The gate is a **no-op at one node**, which is what carries the pair's
evidence across to the gated code — and that is measured, not assumed: `local_miner_numa_nodes`
extracted verbatim from the built artifact and run on the live appliance beside RigForge's own
resolution agrees at 1.

Also still open on #1103 and **not** claimed here: whether a co-located miner on a reduced-RAM box
mines usefully or thrashes. It needs synced chains, not another guest — the sync gate holds the
stratum the miner dials, so xmrig takes ~40 pages and never allocates a dataset. A gated container's
absence is not memory pressure.

## Tests and gates

- `tests/stack/test_appliance_hugepages.sh`: **43 passed, 0 failed** (was 32). New: five detection
  cases including the socket fallback, the NUMA=2 refusal and its warn, and a render **controlled
  pair** (one node -> key present; two -> absent) — because "key absent" cannot otherwise be told
  from a fixture that never armed. The node count is **pinned** via a fake `lscpu` on PATH; without
  that, every ceiling case would answer to the runner's own topology.
- Mutation battery, each proven applied by byte-diff against a pristine copy and restored
  byte-identical: delete the node gate -> 3 kills; invert it -> 10; drop the socket fallback -> 1;
  break the sysfs glob -> 1. The two single kills are each on their own axis.
- Full `tests/stack/run.sh`: **3053 passed, 0 failed** (rc file read, `✗` count 0). That total is
  **unchanged** by this PR and should be: `test_appliance_hugepages.sh` sits in `tests/inventory.sh`'s
  UNSOURCED set — the underscore in its name marks exactly that — and is invoked standalone from
  `Makefile:21`, `Makefile:62` and `.github/workflows/ci.yml:461`, never from `run.sh`. So the 11 new
  cases are counted in the 43/0 above, not in the 3053.
- `make lint` rc=0; `lint-pithead-parity` OK (52 slices);
  budget gate OK (347 and 304 lines against target 400, **no new tsv row**); `tests/inventory.sh`
  lists the file with 9 sections.

The ponytail gate keys off `git rev-parse --abbrev-ref HEAD` in the pinned shell cwd, which names a
different branch from this lane's worktree, so it cannot do this pass here — run manually and
disclosed. Both new helpers earn their place: `local_miner_pool_ceiling_mb` has four distinct exit
reasons that each need their own operator-facing message, and `local_miner_numa_nodes` exists to be
independently testable against a faked topology, which is the only way the NUMA=2 case can be
exercised at all. No adjacent helper could absorb either without making the render function's
branching untestable.

## Shared files

`docs/` is shared and no lane owns it — `docs/workers.md` and `os/KNOWN-ISSUES.md` are updated here,
disclosed per policy. `pithead` is regenerated by the build only, never hand-edited.

One correction to a figure already in this branch's `os/KNOWN-ISSUES.md`: it said the kernel "granted
356 of them" against a 6146-page request, which reads as 356 of 6146. The pool went 2560 -> 2916 —
356 **new** pages of the 3586 it was asked to add. Both figures were true in different senses and the
sentence blended them; the conclusion never moved, which is why nothing surfaced it.

### DIFF
--- a/docs/workers.md
+++ b/docs/workers.md
@@ -417,7 +417,8 @@ no LAN exposure, no firewall rule, no Tor hop.
 XMRig, so answering yes to "Mine on this machine too?" in the setup wizard is the whole job:
 pithead renders the miner's configuration (pool, stratum password, the stack's HugePages
 headroom, and a ceiling on how large the HugePages pool may grow), runs RigForge's setup in its
-appliance mode after the stack is up, and repeats that on every boot — the miner's service lives in `/run` and is re-created each time, like every other
+appliance mode after the stack is up, and repeats that on every boot — the miner's service lives
+in `/run` and is re-created each time, like every other
 derived thing on the appliance. The worker appears in the dashboard's Workers Alive table; toggle
 `local_miner.enabled` from the dashboard's configuration view to turn it off or on again. The
 rest of this section is the DIY flow.
--- a/docs/workers.md
+++ b/docs/workers.md
@@ -415,9 +415,9 @@ no LAN exposure, no firewall rule, no Tor hop.
 
 **On the appliance, this is built in.** The image carries a pinned RigForge tree and a prebuilt
 XMRig, so answering yes to "Mine on this machine too?" in the setup wizard is the whole job:
-pithead renders the miner's configuration (pool, stratum password, and the stack's HugePages
-headroom), runs RigForge's setup in its appliance mode after the stack is up, and repeats that on
-every boot — the miner's service lives in `/run` and is re-created each time, like every other
+pithead renders the miner's configuration (pool, stratum password, the stack's HugePages
+headroom, and a ceiling on how large the HugePages pool may grow), runs RigForge's setup in its
+appliance mode after the stack is up, and repeats that on every boot — the miner's service lives in `/run` and is re-created each time, like every other
 derived thing on the appliance. The worker appears in the dashboard's Workers Alive table; toggle
 `local_miner.enabled` from the dashboard's configuration view to turn it off or on again. The
 rest of this section is the DIY flow.
--- a/lib/pithead/14-local-miner.sh
+++ b/lib/pithead/14-local-miner.sh
@@ -74,11 +74,18 @@ local_miner_hugepages_blocked() {
 #               the miner would get nothing beyond the boot pool — a cap that reads as a cap and
 #               silently switches the reservation off.
 #   >= 8592 MB  required on the supported machine is the fallback branch (the appliance reserves no
-#               1G pages, so proposed-grub.sh takes it): 1168*NUMA + THREADS + 50 + (extra_mb+1)/2
-#               = 1168 + 6 + 50 + 3072 = 4296 pages. Below this the cap would bite on a HEALTHY box.
+#               1G pages, so proposed-grub.sh takes it): 1168*NUMA + THREADS + 50 + (extra_mb+1)/2.
+#               ⛔ THAT BOUND IS NOT A CONSTANT — IT CARRIES A NODE COUNT, and 8592 is only its
+#               NUMA=1 evaluation (1168*1 + 6 + 50 + 3072 = 4296 pages). The node term is the only
+#               one that scales: EXTRA_2MB_PAGES does not (proposed-grub.sh L85-88). At NUMA=2 the
+#               requirement is 5464 pages = 10,928 MB and a 4608-page ceiling would cap a HEALTHY
+#               box 856 pages short; at NUMA=4, 7800 pages, short by 3192. RigForge caps and
+#               continues, so that under-reservation would be SILENT — the same failure direction
+#               the version gate below exists to avoid. This is why the value is gated on a
+#               detected single node (local_miner_numa_nodes) and not merely documented as one.
 #   <  11236 MB the contended landing measured above, 5618 pages. At or above it the ceiling would
 #               never bind and the over-reserve would survive the fix.
-# 4608 pages clears all three, with 312 pages of slack over required.
+# 4608 pages clears all three AT ONE NODE, with 312 pages of slack over required.
 #
 # NOT the guest's 4322: that reading is a KVM artifact (4 vCPUs presented as 4 sockets -> L3 64 MiB
 # -> THREADS 32). The same physical X5690 reports one 12 MiB L3 -> THREADS 6. Immaterial to the
@@ -98,6 +105,31 @@ readonly PITHEAD_HUGEPAGES_POOL_CEILING_MB=9216
 # is worse than one that plainly is not, so declare it only where it is honoured.
 readonly PITHEAD_RIGFORGE_POOL_CEILING_FLOOR=1.16.0
 
+# The node count the ceiling is valid for. MIRRORS RigForge's OWN precedence, step for step
+# (util/proposed-grub.sh L52-61 at the pinned ref), and that fidelity is the whole point: this
+# function exists to PREDICT the node count rigforge will size against, so any disagreement between
+# the two reintroduces the very defect the gate is here to stop — quietly, and in the silent
+# direction. Three steps, not two: lscpu's "NUMA node(s)", then the sysfs node count, then THE
+# SOCKET COUNT, then 1. The socket fallback is not hypothetical padding — the #1103 bench guest
+# reports Socket(s)=4 with NUMA node(s)=1, so a mirror that stopped at sysfs would read 1 where
+# rigforge read 4. rigforge's own L48-50 says the same thing from the other side: node count is not
+# socket count, and a single-socket EPYC can expose 2/4/8 nodes.
+# Prints its answer, so it must call nothing that writes to stdout.
+local_miner_numa_nodes() {
+    local n
+    n=$(lscpu 2>/dev/null | awk -F: '/^NUMA node\(s\):/ {gsub(/[^0-9]/, "", $2); print $2; exit}')
+    if ! { [ -n "$n" ] && [ "$n" -gt 0 ]; } 2>/dev/null; then
+        n=$(find "${PITHEAD_NODE_SYS:-/sys/devices/system/node}" -maxdepth 1 -name 'node[0-9]*' 2>/dev/null | wc -l | tr -d ' ')
+    fi
+    if ! { [ -n "$n" ] && [ "$n" -gt 0 ]; } 2>/dev/null; then
+        n=$(lscpu 2>/dev/null | awk '/Socket\(s\):/ {print $2; exit}')
+    fi
+    if ! { [ -n "$n" ] && [ "$n" -gt 0 ]; } 2>/dev/null; then
+        n=1
+    fi
+    echo "$n"
+}
+
 # The ceiling this machine may declare, in MB, or 0 for "declare nothing" — 0 is RigForge's own
 # documented default for "no ceiling", so an omitted key and a declared 0 mean the same to it.
 # Prints its answer, so it must never call log (stdout); warn goes to stderr and is safe here.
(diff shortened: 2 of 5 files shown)


VijitSingh97:
## Non-author pass (#1103) — controller lane

**Verified, ship it once the four in-flight checks land green.** This is a true non-author
pass: the branch is not mine, and it is the one review on this change that a different
seat could give. Everything below says which instrument settled it, and the last section
says plainly what I did NOT check.

### The claim the whole gate rests on, re-derived at the PINNED ref

`local_miner_numa_nodes` exists to PREDICT the node count RigForge will size against, so
any disagreement between the two silently reintroduces the defect. The PR claims it
mirrors RigForge's precedence "step for step". **I checked that at the pinned ref, not at
a working checkout** — `os/rootfs/Dockerfile` pins `RIGFORGE_REF=4ce29b3d…`, whose
`VERSION` is **1.16.0**. All four steps agree:

| step | rigforge `util/proposed-grub.sh` @ `4ce29b3d` | pithead `local_miner_numa_nodes` |
|---|---|---|
| 1 | `lscpu … awk -F: '/^NUMA node\(s\):/ {gsub(/[^0-9]/,"",$2); print $2; exit}'` | identical |
| 2 | `find "$NODE_SYS" -maxdepth 1 -name 'node[0-9]*' \| wc -l \| tr -d ' '` | identical, `PITHEAD_NODE_SYS` overridable |
| 3 | `NUMA_NODES="$SOCKETS"`, itself `lscpu \| awk '/Socket\(s\):/{print $2; exit}'` (L43) | same awk, inlined |
| 4 | `SOCKETS=1` (L45) | `n=1` |

The guard between each step is the same `! { [ -n "$n" ] && [ "$n" -gt 0 ]; } 2>/dev/null`
construction. The only textual difference is a `2>/dev/null` pithead adds on the socket
`lscpu`, which is strictly safer and cannot change the result. **The mirror claim holds.**

**And the floor is not an arbitrary version:** `PITHEAD_RIGFORGE_POOL_CEILING_FLOOR=1.16.0`
is exactly the `VERSION` at the pinned ref, so the gate opens precisely on the tree the
image ships and not a release earlier.

### Every exit is the SAFE direction, which is the property that matters

`local_miner_pool_ceiling_mb` has four early returns — reduced tier, multi-node, unparseable
version, version below floor — **and every one of them `echo 0`, which is RigForge's
documented "no ceiling" and is exactly today's behaviour.** There is no path on which this
change makes the pool SMALLER than it is today. That is the correct asymmetry for this
defect: an over-large pool is the bug being fixed, an under-reservation would be a new and
silent one, so a gate that cannot tell must decline to cap. The version gate's reasoning is
the same and is right — RigForge's `_warn_unknown_config_keys` warns and never errors, so
declaring the key into an older tree fails OPEN and renders a config that *reads* as
capped. **A config that lies about being bounded is worse than one that plainly is not.**

### `stdout` discipline — checked because both functions return through it

Both `local_miner_pool_ceiling_mb` and `local_miner_numa_nodes` are consumed via `$(…)`, so
anything they print that is not the answer corrupts the return value, and `--argjson
ceiling` would then receive garbage. The code asserts twice that `warn` is safe here. **I
verified it rather than trusting the comment:** `lib/pithead/00-prelude.sh:32` is
`warn() { echo -e "…" >&2; }` — stderr. And `hugepages_decision_pages`
(`37-kernel-tuning-and-preflight.sh:14-25`) emits exactly one value on stdout on both
branches. Clean on both counts.

### The version comparison, worked at the three boundaries

`printf '%s\n%s\n' FLOOR ver | sort -V | head -n 1 != FLOOR` → "predates". At `ver = 1.16.0`
`sort -V` puts FLOOR first, so it equals FLOOR, so it is NOT treated as predating and the
ceiling IS declared — the floor is inclusive, which is what the comment intends. `1.15.2`
sorts first and is refused; `1.17.0` leaves FLOOR first and is honoured. The `case`
rejecting `'' | *[!0-9.]*` before this is doing real work: `sort -V` will happily order an
unparseable string against a version and answer "new enough", so refusing it first is the
only thing that makes the comparison meaningful.

### Arithmetic, re-derived

`1168*1 + 6 + 50 + 3072 = 4296` pages = **8592 MB**, reproducing the bound the shipped
comment names — a genuine internal control, and worth keeping. At NUMA=2:
`1168*2 + 6 + 50 + 3072 = 5464` pages = 10,928 MB against a 4608-page ceiling, i.e. **856
pages short of a healthy requirement** — which is precisely the silent under-reservation the
node gate now prevents. The ceiling 9216 MB = 4608 pages sits between 6144 and 11236 as the
comment says.

### One thing I want to praise rather than change, because it is the reviewable part

The node gate is **tested, not merely documented**: `fake_lscpu` shims `lscpu` onto `PATH`
and `PITHEAD_NODE_SYS` injects the sysfs tree, so the precedence is exercised step by step
rather than asserted in prose. A comment claiming to mirror another project is the kind of
claim that rots silently; a test that pins each step is what makes it survive their next
release.

### What I did NOT check — say so if you quote this pass

**Taken on report, not verified by me:** the suite figures (3053/0, hugepages 43/0), the
`make lint` rc, the parity and file-budget gate results, `tests/inventory.sh` listing, the
two standalone legs, the on-hardware NUMA witness logs, and the line counts of the two new
files. I re-derived the rigforge mirror, the four exit paths, the `warn`/stdout property,
the version boundaries and the arithmetic — nothing else. **I also did not run the suite
myself**, so my pass is a reading of the change, not a re-execution of its evidence.

**Merge is mine and I am holding it until the four pending checks land**, not because I
expect a problem but because "green when I looked" and "green" are different claims. Ping
me, or I will re-read the rollup.

VijitSingh97:
## Correction to my pass above — replace the check COUNT with the condition

My pass says the verdict holds "once the four in-flight checks land green." **Do not merge
against that number.** It was true when I drafted the sentence and stale by the time I
posted it; the appliance lane caught it. The count has since read four, then two, and at
this writing one — only "Shell tests (shellcheck + pithead suite)" is outstanding, with
15 SUCCESS, 2 SKIPPED and **zero failures**.

**The corrected hold condition, which does not decay:** merge when the rollup shows **no
check in a non-terminal state and none failed** — re-read it at merge time rather than
counting from any sentence in this thread, mine included. **The verdict itself is
unchanged: ship it.** Nothing in the review turned on how many checks were outstanding.

**Worth naming because it is the same defect I merged a PR to fix an hour ago.** #1544
removed a line count from `docs/dev/repo-map.md` on the ground that a precise figure about
a moving target is a defect generator — the number had held three values in one day.
I then wrote a decaying count into a review comment that gates a merge, where a reader
comparing "four pending" against a rollup showing one has to decide which to trust.
**A figure is safe in prose only when nothing acts on it; this one had an imperative
attached, which is exactly the class that bites.** State the condition, not the count.

PR #1550, opened 2026-08-29
test(dashboard): make a collapsed error/answer return channel impossible to add silently (#1487)

### LINKED ISSUE #1409: Worker Inspect accuses another dashboard when our own SQLite read fails

### LINKED ISSUE #1487: An error and an answer share one return channel in 32 read handlers — can a mechanism make that impossible?

VijitSingh97:
## What this is

#1487 asked a mechanism question, not for a sweep of instances:

> Can a helper, a sentinel type, or a lint make it impossible in this package for an error and an
> answer to share a return channel — or at least make each instance a decision someone had to make
> on purpose?

This is option 3 (the lint), shipped as a pytest module rather than a `scripts/` lint because every
existing lint lives in `scripts/` + the root `Makefile`, both outside this lane, and
`tests/service/test_mixin_atomicity.py` is in-lane precedent for exactly this shape: an AST
structural gate that ships with the property it protects and is re-checked by CI on every later PR.

**No production code changes.** The diff is one test module, nine lines of `docs/dashboard.md`, and
one `docs/dev/file-budget.tsv` row.

## The rule, and why the opt-out is the type annotation

    a FAILURE path may not return a value that inhabits the function's declared SUCCESS type.

The issue's option 3 called for "an explicit opt-out marker for the handlers where collapsing really
is correct". The annotation already is that marker, and using it means no new vocabulary:

| shape | verdict |
|---|---|
| `-> list[dict]` … `except: return []` | **collapse** — no caller can tell the failure from the answer |
| `-> list[dict] \| None` … `except: return None` | **signed** — the failure is out of band, on purpose |

The second row is exactly the fix #1409 applied. The mechanism and the precedent agree by
construction, and the decision is visible in the signature a reader already reads rather than buried
in a handler.

## Two shapes covered, and one deliberately not

* **The `except` shape.** A collapsed literal returned from an exception handler.
* **The HANDLE-GUARD shape** — `if not self._conn: return []`, no exception involved. This is the
  SILENT half of #1409, the route that raised nothing and logged nothing, and it is invisible to any
  rule anchored on `except`. Covered here because the two are one defect reached by two doors.

**Not covered, and not claimed: the general guard-clause shape.** A guard on a DOMAIN value — `if
not rows`, `if prev is None`, `if cmd is None` — is usually a real early return, not a failure.
Re-derived at this tip: **171 guard-clause failure-return sites, of which 32 test a resource handle
and 139 test a domain value.** Only the 32 are in scope. An earlier precision pass over the wider
class found a single true positive in it — `storage_service.py load_snapshot`, which is already
signed — which is why the line is drawn at the handle rather than at falsiness. A near-miss control
pins that narrowness so it cannot drift into the 139.

(The 32 cross-checks exactly against `grep -c "if not self._conn:"` over the package.)

## What is asserted, and what is only reported

Two HARD laws. **Neither certifies an unread instance.** Law 1 does pin five existing functions
by NAME — say so plainly rather than claim a zero-baseline the file does not have — but it pins
them to the SIGNED state, each read at source first, which ratchets in the opposite direction
from accepting a defect. It can only ever go red by someone UNDOING a fix:

1. **The signed set stays signed.** Five functions already express the three-valued contract. Each
   was read at source; `get_worker_config_change` documents it in its own docstring. Dropping the
   `| None` from one would return it to the collapsed class silently.
2. **A signature that promises out-of-band failure must honour it.** If the annotation admits
   `None`, every failure return must BE `None`, not an empty container. Zero instances today, and
   it stays that way. This catches the HALF-DONE fix — the `except` arm corrected while the handle
   guard still returns `[]` — which a per-function reading of #1409 cannot see.

The remaining collapses are **REPORTED, NOT CERTIFIED**. The reporting test asserts nothing about
their number, and there is no baseline file.

### Why there is no baseline

A baseline is a ratchet. Baselining the current instances would mark whatever real defects they
contain as permanently accepted — converting "a default nobody noticed" into "a default someone
signed", with the same nobody having read them, and a green gate now asserting the state is
intentional. That is **#1487's own failure mode reintroduced one level up, concealed by the gate's
own greenness**. It is also the shape #1409 arrived in: `docs/dashboard.md` had written the defect
down as a safety property.

A `<= 11` bound would certify eleven unread functions. A `> 0` bound would go red the day someone
fixed the last one. So the residual gets a count and its names, and the two laws are what hold.

## The residual, as a measurement

At `a682054`: **11 collapses · 5 signed · 0 half-fixes · 57 failure returns the gate cannot judge**
because their functions are unannotated.

That 57 is the honest limit, not an apology: the mechanism keys on the declared type, so a function
that declares nothing declares no contract to check. Annotating that layer would bring it into
scope and is the obvious next move; this PR does not pretend to have made it.

**Reconciliation with the issue's figures, so nothing reads as missed.** The issue counted Rule A 49
/ Rule B 33 at `f4eca13`; at this tip the same enumerator gives **48 / 34** (the deltas are
`storage_service.py` having been split by #1369/#1536, not the issue being wrong). Those 34
functions land as:

| | count |
|---|---|
| unannotated → **blind**, outside what this gate can judge | 19 |
| annotated, failure value inhabits the success type → **collapse** | 11 |
| annotated `T \| None`, returns `None` → **signed** | 4 |

19 + 11 + 4 = 34, exactly. The gate's fifth signed function (`load_snapshot`) is a guard-only case
Rule B never saw. **Nothing in the 34 is unaccounted for, and nothing was silently dropped.**

`add_payouts` is worth naming out of the 11: a failed INSERT returns `[]`, indistinguishable from
"every row was already stored", and that value is what drives the `payout_confirmed` alert. The
alert that never fires is the failure's only symptom.

## Evidence

* **18 tests in the new module**, 13 of them controls. Every control is seeded ACROSS the boundary
  the classifier decides on; a control built only from already-safe shapes would pass against a
  checker that answered "safe" to everything.
* **Mutation battery 3/3 KILLED** (`mutate-1487.py`) — one seeded regression per law, applied to
  REAL package source, proven present on disk, run, restored, with a post-battery green baseline
  proving every restore was clean. The battery refuses a pattern that does not match exactly once.
  Both doors are proven to feed law 2 separately: the `except` arm and the handle guard.
* **Full dashboard suite green** at `a682054` (rc 0).
* **Lints run:** `lint-py` (ruff check + format), `lint-file-budget`, `lint-md`, `lint-docs-voice` —
  all clean. `lint-sh` was not run and is not affected: no shell file is touched.

### Two false-alarm risks found and closed while building this

Both were found by attacking the gate rather than by a failing test, and both would have produced a
red against correct code — which costs trust faster than a miss:

* `_admits_none` first tested the annotation text for `None`. `Callable[[], None]` and
  `dict[str, None]` contain that token while promising a caller nothing about failure. Now only
  TOP-LEVEL union members count, with a control for both shapes.
* Reading a bare `return` as distinct from `return None` would have redded law 1 the day someone
  dropped the word. Normalising them then over-corrected — three `-> None` procedures began scoring
  as "signed", inflating the cleared set with functions that never made the promise. `-> None` is
  now out of scope entirely, with a control.

## Known limits

* **The warn-only set can be silenced by deleting an annotation** (the function becomes "blind").
  This is bounded: nothing is asserted about that set anyway, so the evasion changes nothing
  material. For the five SIGNED functions, deleting the annotation DOES red the gate — law 1
  catches it.
* **`make test-patch-coverage` was not run.** It is vacuous here twice over: the diff adds no
  production line, and #1537 has it comparing against a hardcoded `origin/develop`. The mutation
  battery is the substitute evidence.

Closes #1487.

### DIFF
--- a/dashboard/tests/service/test_collapsed_return_channels.py
+++ b/dashboard/tests/service/test_collapsed_return_channels.py
@@ -1,0 +1,563 @@
+"""#1487 — an error and an answer may not share a return channel, as a mechanism (#1409).
+
+#1409 was one instance: `get_worker_config_history` returned `[]` both when the read failed and
+when it succeeded and found nothing. Its caller reads that list to decide where a rig's running
+config came from, and an empty list means *"we hold no row"*, which renders as **"Last changed
+from another dashboard."** A database that failed to open produced an on-screen accusation that
+another operator had changed the rig.
+
+#1487 asked for a MECHANISM rather than a sweep of the instances, because fixing them one at a
+time makes today's known instances safe and does nothing about the next one. This is that
+mechanism, and the rule it enforces is:
+
+    a FAILURE path may not return a value that inhabits the function's declared SUCCESS type.
+
+**The opt-out is the type annotation, not a marker comment.** A handler that genuinely wants to
+collapse says so by declaring a type the collapsed value fits; a handler that wants the failure to
+be distinguishable widens its annotation to `T | None` and returns `None`. That is exactly the fix
+#1409 applied, so the mechanism and the precedent agree by construction, and "a decision someone
+signed rather than a default nobody noticed" is expressible in the signature a reader already
+reads. Nothing here needs a new vocabulary.
+
+## The two shapes this covers, and the one it does not
+
+* **The `except` shape.** A collapsed literal returned from an exception handler.
+* **The HANDLE-GUARD shape.** `if not self._conn: return []` — no exception involved. This is the
+  SILENT half of #1409, the route that raised nothing and logged nothing, and it is invisible to
+  any rule anchored on `except`. It is covered here because the two shapes are the same defect
+  reached by two doors, and a gate that closed only one would leave the door the reader cannot see.
+
+**It does NOT cover the general guard-clause shape, and does not claim to.** A guard on a DOMAIN
+value — `if not rows`, `if prev is None`, `if cmd is None` — is usually a real early return rather
+than a failure. Re-derived at this tip: 171 guard-clause failure-return sites, of which 32 test a
+resource handle and 139 test a domain value; an earlier precision pass over the wider class found a
+single true positive in it. The discriminator kept here is that the guard tests a RESOURCE HANDLE.
+Widening past that would trade a gate people trust for a gate people mute; the near-miss control
+below pins the narrowness so it cannot drift.
+
+## What is asserted, and what is only reported
+
+Two laws are HARD, and neither needs a baseline of existing instances:
+
+1. **The signed set stays signed** (`_SIGNED`). Five functions already express the three-valued
+   contract, #1409's own fix among them. Dropping the `| None` from one of them would return it to
+   the collapsed class silently — the regression this file exists to make impossible.
+2. **A signature that promises an out-of-band failure must honour it.** If a function's annotation
+   admits `None`, every failure return in it must BE `None`, not an empty container. This is a
+   self-consistency law with no baseline at all: it is clean today, it stays clean forever, and it
+   catches the half-done fix — the except arm corrected while the handle guard is left returning
+   `[]`, which is precisely the per-site shape a per-function reading of #1409 would miss.
+
+The remaining collapses are **REPORTED, NOT CERTIFIED**, and the reporting test asserts nothing
+about their number. This is deliberate and it is the whole ruling on this file:
+
+> **A baseline is a RATCHET.** Baselining the current instances would mark whatever real defects
+> they contain as permanently accepted — converting "a default nobody noticed" into "a default
+> someone signed", with the same nobody having read them, and a GREEN GATE now asserting the state
+> is intentional. That is #1487's own failure mode reintroduced one level up, concealed by the
+> gate's own greenness. It is also the shape #1409 arrived in: `docs/dashboard.md` had written the
+> defect down as a safety property.
+
+So the un-read instances get no row, no ratchet, and no green. They get a count and their names.
+
+## The residual, stated as a measurement
+
+At the tip this shipped against: **11 collapses** flagged (all in the DB read layer —
+`storage_service.py` and `telemetry_store.py`), **5 signed**, **0 half-fixes**, and **57 failure
+returns this gate CANNOT judge because their functions are unannotated.** That 57 is the honest
+limit, not an apology: the mechanism keys on the declared type, so a function that declares nothing
+declares no contract to check. Annotating that layer would bring it into scope and is the obvious
+next move; this file does not pretend to have made it.
+
+`add_payouts` is worth naming out of the 11, because it shows the class is not academic here: a
+failed INSERT returns `[]`, indistinguishable from "every row was already stored", and that value
+is what drives the `payout_confirmed` alert. The alert that does not fire is the failure's only
+symptom.
+"""
+
+import ast
+import pathlib
+
+import pytest
+
+_PACKAGE = pathlib.Path(__file__).resolve().parents[2] / "mining_dashboard"
+
+# The collapsed values an error path can hide inside a success type. `False` is deliberately NOT
+# here: it was measured at 5 further instances, all of them outbound senders (`_post`/`ping`/
+# `send`) where False-on-failure and False-elsewhere mean the same thing to a caller. Including it
+# would add five probable false positives to a gate whose whole value is that its findings are real.
+_EMPTY = {"[]", "{}", "0"}
+
+# The DB handle. A guard testing THIS is a failure route; a guard testing a domain value is an
+# early return. That distinction is what separates 32 in-scope sites from 139 deliberately excluded.
+_HANDLE = "_conn"
+
+# Measured, read at source, and cleared — the only set this file ratchets. Each declares `T | None`
+# and returns `None` from every failure path, which is the three-valued contract #1409 established.
+# `get_worker_config_change` documents it in its own docstring: "None means the read FAILED ...
+# {} means there is genuinely no such row". Listing them by name is what makes law 1 enforceable;
+# the vacuity guard below is what keeps the list from silently becoming a check against nothing.
+_SIGNED = {
+    "service/storage_service.py:get_xvb_standby",
+    "service/storage_service.py:get_kv",
+    "service/storage_service.py:load_snapshot",
+    "service/worker_config_store.py:get_worker_config_history",
+    "service/worker_config_store.py:get_worker_config_change",
+}
+
+
+def _collapsed_literal(value: ast.AST | None) -> str | None:
+    """The collapsed value a ``return`` yields, or None if it is not one of the shapes we track.
+
+    An empty list/dict written as a literal, a bare `return`, `None`, and integer `0`. `False` is
+    reported so the classifier can SEE it, and excluded from `_EMPTY` so it cannot be flagged — the
+    two are different decisions and collapsing them here would be this file's own defect. `0` is
+    guarded against `bool`, because `False == 0` is true in Python and an unguarded check would
+    silently reclassify every boolean return as a numeric collapse.
+    """
+    if value is None:
+        return "<bare>"
+    if isinstance(value, ast.List) and not value.elts:
+        return "[]"
+    if isinstance(value, ast.Dict) and not value.keys:
+        return "{}"
+    if isinstance(value, ast.Constant):
+        if value.value is None:
+            return "None"
+        if value.value is False:
+            return "False"
+        if value.value == 0 and isinstance(value.value, int) and not isinstance(value.value, bool):
+            return "0"
+    return None
+
+
+def _own_nodes(function: ast.AST) -> list[ast.AST]:
+    """Every node belonging to THIS function, never to a def nested inside it.
+
+    A nested def is its own function with its own contract: folding its returns into the parent
+    both double-counts and misattributes. `web/server.py:_log_filters` and the `_num` closure
+    inside it are the real case — a naive walk reports them as one function and gets the answer
+    wrong in both directions at once.
+    """
+    collected: list[ast.AST] = []
+
+    def walk(node: ast.AST) -> None:
+        for child in ast.iter_child_nodes(node):
+            if isinstance(child, (ast.FunctionDef, ast.AsyncFunctionDef)):
+                continue
+            collected.append(child)
+            walk(child)
+
+    walk(function)
+    return collected
+
+
+def _union_members(annotation: ast.AST) -> list[ast.AST]:
+    """The TOP-LEVEL alternatives of a return annotation, flattened.
+
+    Only the top level, which is the entire point. `None` nested inside a parameter — the return
+    slot of a `Callable[[], None]`, the value type of a `dict[str, None]` — says nothing about
+    whether THIS function can return None, and a walk of the whole annotation tree cannot tell the
+    two apart. Nothing in the package wears that shape today; this is written so the gate does not
+    start misreading the day something does, because the failure would be a false accusation
+    against a function that never collapsed anything.
+    """
+    if isinstance(annotation, ast.BinOp) and isinstance(annotation.op, ast.BitOr):
+        return _union_members(annotation.left) + _union_members(annotation.right)
+    if isinstance(annotation, ast.Subscript):
+        base = annotation.value
+        name = base.id if isinstance(base, ast.Name) else getattr(base, "attr", "")
+        if name == "Optional":
+            # Both halves: `Optional[T]` is `T | None`, and returning only the None would make the
+            # annotation indistinguishable from a bare `-> None` procedure.
+            return _union_members(annotation.slice) + [ast.Constant(None)]
+        if name == "Union":
+            inner = annotation.slice
+            elements = inner.elts if isinstance(inner, ast.Tuple) else [inner]
+            return [member for element in elements for member in _union_members(element)]
+    return [annotation]
+
+
+def _returns_nothing(annotation: ast.AST | None) -> bool:
+    """Is this declared `-> None` — a procedure with no return value at all?
+
+    Distinct from `T | None`, which is a value type carrying an out-of-band failure marker. The
+    procedure has no answer channel for an error to share, so it is neither signed nor collapsed;
+    it is simply not what this file is about.
+    """
+    members = _union_members(annotation) if annotation is not None else []
+    return len(members) == 1 and isinstance(members[0], ast.Constant) and members[0].value is None
+
+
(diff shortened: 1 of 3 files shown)


VijitSingh97:
## Non-author pass — PASS. Merging.

Reviewed as the controller. This is the ruling-E build, and it honours the ruling: no baseline, and
the residual reported as a measurement.

### The claim I re-derived rather than relayed

A gate's only load-bearing property is that it can go RED on the regression it exists to catch.
Everything else about this file is commentary, so that is what I tested — against the **real
package**, not against the file's own fixtures.

Controlled pair in my own worktree:

- **Positive control** — this file against the current tip: **15 passed**.
- **Discriminating leg** — dropped ` | None` from `storage_service.get_kv`, one of the five in
  `_SIGNED`, which is exactly the "regression that looks like a simplification" the file claims to
  prevent. Mutation proven applied by diffing the signature line before and after
  (`-> str | None` becoming `-> str`). Result: **two failures, and the right two** — the vacuity
  guard (`assert 4 >= 5`) and law 1's subset check.
- **Restored byte-identical** (`cmp` clean) and re-run: **15 passed** again.

So the gate is not decoration. It fails loudly on the exact edit it is built to notice.

### Ruling E — cleared, and the property that clears it is not the one the body leads with

I asked the question the lane raised: does pinning five functions by name in `_SIGNED` cross the
do-not-baseline ruling? **It does not**, and the reason is not merely that it points the other way.
A stale pin that failed OPEN would be a baseline wearing a ratchet's clothes. What actually settles
it is that law 1 asserts `_SIGNED <= set(package["signed"])` — a true **subset** check — so a signed
function that is renamed, removed, or un-annotated goes **red**. Fail-closed. That is the property
worth stating in the file, because intent is not enforceable and the subset assertion is.

Two things I checked that could have made the laws pass for the wrong reason:

- **Law 2 (`half_fix == []`) is trivially satisfiable by an empty scan.** It is protected: one
  module-scoped `package` fixture feeds all three laws, and it carries an enumeration guard —
  `assert len(files) > 50, "the package walk found almost nothing"`. An empty scan is not a clean
  scan, and this file says so in its own comment and then implements it.
- **The vacuity guard's first line is a COUNT** (`len(package["signed"]) >= len(_SIGNED)`), and a
  count is not a match — it cannot separate "all five still signed" from "one renamed, a different
  function newly signed". Law 1's subset check covers that case, so the file is sound as it stands;
  I note it because the count line is not the line doing the work.

### The residual hazard, which is human rather than mechanical

When a legitimate refactor reds law 1, the tempting repair is **deleting** the name from `_SIGNED`
rather than updating it — which silently shrinks the pinned set, and nothing goes red afterwards.
One sentence in the file saying a rename updates the entry and never removes it would close that.
Not blocking, and I am not holding the PR for it.

### What I verified on the mechanics

- **Ruling 9 holds.** The `file-budget.tsv` diff is **exactly one row** for the one new file. No other
  lane's ceiling moved, so no un-rebased branch reds on this.
- **The alphabetical placement nit is fixed** — the row now sits between `test_algo_service.py` and
  `test_control_service.py`, so a later `--generate` will not move it and read as someone else's edit.
- **The branch's evidence survives the tip move.** Its checks ran against a base that has since taken
  #1546 and #1549. I ran this file against the **current** tip, not its own base, and it passes there
  — so this is not a green inherited from a program that no longer exists.
- **Ruling 4, worth naming rather than changing:** the row is `563` and the file is 563 lines, so it
  is legal and has **zero headroom**. Ceilings only go down, so the next line added to this file reds
  the gate and cannot be fixed by raising the row. Deliberate, and the lane should know it.

### The residual is reported honestly

11 collapses flagged, 5 signed, 0 half-fixes, and **57 failure returns the mechanism cannot judge
because their functions are unannotated**. That 57 is stated as the gate's real limit rather than
buried, and the reporting test asserts nothing about the count — which is the whole ruling. The
reconciliation is worth stating for anyone reading the earlier enumeration: the 34 is 19 blind +
11 collapse + 4 signed, so nobody should read "34 found, 11 flagged" as 23 missed.

Merging on the standing condition, read at the moment of merge: no check non-terminal, none failed.

**Relayed, not verified by me:** the 18 gate-test count and the 3/3 mutation result from the PR body,
and the frontend/ruff/lint legs of CI. I ran this file, not the suite.

Issue #1558, opened 2026-08-30
#1551's revision-drift table has no writer: the poll caller needs headroom in data_service.py that the obvious extraction cannot free


VijitSingh97:
## What is shipped, and what is not

#1551's storage half lands the `revision_drift` table and its accessors. **Nothing writes to it
yet.** This issue is that gap, named deliberately rather than left implicit, and it holds the
analysis of why the wiring did not ship with it so the next attempt does not re-derive it.

The controller owns the decision to ship the halves separately. The alternative was holding the
storage half until the wiring was possible, which would have parked it on a policy item that waits
indefinitely.

## The wiring, and why it is not a small change

Option A is "on each poll, compare". The audit row must go through
`data_service.DataService._record_audit_event` — that is the method running every field through
`audit_service._clean`, and writing audit rows around it skips the sanitizer its own docstring warns
against. So the caller has to reach that method on that instance.

`data_service.py` is **1454 lines against a 1454 ceiling** — zero headroom, and the ratchet forbids
raising the row. So the wiring must be paid for by a shrink in the same file, in the same PR.

## The extraction that looked right and is not

`DataService` is a bare 1321-line class with no composition, while `StateManager` in the same
package is already `class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin)`. So extracting
a gates mixin looks like the same move in the same idiom.

**On dependencies it is.** The five methods — `_apply_worker_rejection` (33),
`_stop_gate_containers` (5), `_start_gate_containers` (6), `_apply_sync_gate` (42),
`_apply_fail_closed_gate` (45) — total **131 lines** and touch nothing but `self.*`. A mixin
preserves every one, `_record_audit_event` included.

**What kills it is the test suite, which no walk of the source module can see.** Three module-level
constants are monkeypatched on the `data_service` *module object* at **24 sites**
(`SYNC_GATE_CONTAINERS` 19, `REJECT_WORKERS_CONTAINER` 4, `DASHBOARD_FAIL_CLOSED` 1), as
`patch.object(ds_mod, "NAME", ...)`. Moved code binds those names in the new module, which nothing
patches — the documented break-on-move shape. The moved bodies read them in 13 places.

**And the repair is blocked twice over:**

1. `test_data_service.py` is **2538/2538** — it cannot absorb even the single import line the
   `patch.object(gates_mod, ...)` alternative needs.
[...]

Issue #1557, opened 2026-08-30
patch-coverage.sh compares against frozen `develop`, so every develop-v2 PR's coverage number is about the wrong diff


VijitSingh97:
## The defect

`scripts/patch-coverage.sh:11` pins the gate's comparison base as a bare assignment:

    COMPARE=origin/develop

There is no environment override — it is not `${COMPARE:-origin/develop}` — and the value is used
twice, at `:102` for `diff-cover --compare-branch="$COMPARE"` and at `:108` for the
`git diff --name-only "$COMPARE...HEAD"` that drives the overlap check.

**`develop` is frozen. All work lands on `develop-v2`.** So on every lane branch, `COMPARE...HEAD`
resolves from the merge-base of `origin/develop` and HEAD, which means the measured diff contains the
whole of `develop-v2`'s divergence from `develop` in addition to the patch. The gate is measuring a
set nobody asked it to measure, and its verdict is about that set.

## Why it matters in both directions

`diff-cover` reports the covered fraction of changed lines and fails under 90%. With the divergence
folded in, that fraction is dominated by files the PR never touched:

- **A good patch can fail** on uncovered lines in unrelated files it did not write.
- **A bad patch can pass**, diluted to invisibility by a large well-covered unrelated diff. This is
  the worse direction, because the gate reports green and the number looks like a fact about the
  patch.

The overlap check at `:108` inherits the same set, so its "says exactly what was checked" message
names the wrong files too.

Measured instance: a lane on a `develop-v2` branch had the gate score **86 unrelated files**. It
noticed, worked around it by measuring its own added lines against `coverage.xml` directly (30/30),
and disclosed the workaround rather than quoting the gate's number.

## Why this survived

The script is careful and its header documents a *different* vacuity it was built to close (#1000 —
`diff-cover` exiting 0 with "No lines with coverage information"). That guard works. This is a
second, independent way for the same gate to produce a number that is true of the wrong thing, and
the existing guard cannot see it: the run is not vacuous, it is simply about another set.

**It is the shape recently promoted to fleet law — a verdict reported over a set the tool chose is
half a claim, and green over the wrong set is the same word as green over the right one.** There is
no rc, no warning, and no empty output to notice.

## Suggested fix

Make the base overridable and default it to the branch work actually targets:

    COMPARE="${COMPARE:-origin/develop-v2}"

Two things worth doing with it, neither large:

[...]

Issue #1551, opened 2026-08-29
Worker Inspect: a hand-edit to a key we never applied still reads as "changed from this dashboard"


VijitSingh97:
Split out of #1367 so the residual is not closed along with the case that was fixed.

## The gap

#1542 closed #1367 by comparing `last_applied` against the rig's served `rig_config` key by key. That works only for keys **this dashboard has applied at least once**: `last_applied` is a merge of the diffs we authored, so a key we never set has no left-hand side and produces no comparison.

A hand-edit to such a key therefore moves the rig's `revision` and nothing else observable. The Worker Inspect provenance line keeps reading "Last changed from this dashboard" — the same false reassurance #1367 was opened for, reached through the one door #1542 leaves open. #1542 discloses this as the first of its three narrowings.

## Shape

Option A from #1367: persist the last-observed `revision` per worker, and on each poll compare the served `revision` against the stored one. Moved, with no new `last_change_id`? That is a change nothing recorded. It needs no cooperation from RigForge and detects an edit to any writable key, applied-by-us or not.

`revision` is a truncated SHA-256 over a jq-canonical form of the writable keys (`_writable_config_hash`), so this compares the rig's value against the rig's own earlier value — it never reproduces the canonicalisation on our side, which is what #1367 ruled out.

## Blocked on

The file budget, as #1367 recorded: this needs a column and its migration in `dashboard/mining_dashboard/service/storage_service.py`, and both that file and `dashboard/tests/service/test_storage_service.py` sat at their ceilings. Re-measure against `docs/dev/file-budget.tsv` before starting — the #1105 extractions have moved several rows since #1367 was written, and the telemetry-series and worker_config mixins (#1531, #1532) came out of this exact file.

## Scope note, carried over

An edit to a **non-writable** key moves no revision and is out of scope here too. A changed pool password stays invisible for the reasons #1542 gives.

VijitSingh97:
## The budget block is HALF GONE, and the surviving half is much narrower than this issue says

This issue told the next reader to re-measure the budget rows rather than re-quote #1367's, because the #1105 extractions had moved them. I did that. **The premise was carried over from #1367 and is now wrong in both directions** — one blocker has vanished entirely, and the other is not what it was described as.

Measured on `origin/develop-v2` at `6bb4d1a`, real line counts against `docs/dev/file-budget.tsv`:

| file | lines | ceiling | headroom |
|---|---|---|---|
[...]

PR #1559, opened 2026-08-30
feat(dashboard): persist the revision each rig serves (#1551)

VijitSingh97:
## What this adds

A `worker_config_revision` table (worker PRIMARY KEY, revision, last_change_id, ts) and
`note_worker_revision` on the worker-config store: one atomic read-then-write that records the
config revision a rig is serving now and reports a revision that MOVED with no new
`last_change_id` beside it.

## Why the comparison needs a stored revision

#1542 compares `last_applied` against the rig's served config key by key. That works only for keys
this dashboard has applied at least once. `last_applied` is a merge of the diffs we authored, so a
hand-edit to a key we never set has no left-hand side: it moves the rig's own revision, stamps no
`last_change_id`, and the Inspect view keeps reading "Last changed from this dashboard".

Holding the rig's earlier revision gives that case a left-hand side without reproducing RigForge's
canonicalisation, which #1367 ruled out.

The read and the write are one method under the single `_db_lock`, because the comparison IS the
read: two polls of the same rig interleaving a read and a write would both see the same "previous",
and one of the two moves would go unreported. Same handle, same lock, same transaction scope as
every other accessor there (#1369).

It fails CLOSED. Any read or write error returns `None` and accuses nobody. A missed detection
costs one poll — a failed write leaves the stored row unchanged, so the next poll compares against
the same baseline — whereas a false accusation is a permanent audit row naming an operator's rig
for something it did not do.

## THE GAP: this table has no writer

**Nothing in production calls `note_worker_revision`.** `git grep` returns the definition and test
sites only. This PR ships the storage half; the poll-path wiring that would make it observable is
NOT here.

The wiring did not ship with it because the caller cannot live anywhere else. The audit row has to
go through `data_service._record_audit_event`, the only path that runs every field through
`audit_service._clean`; writing audit rows from the store would skip the sanitizer its own
docstring warns against. `data_service.py` sits at 1454/1454 against the file budget — I confirmed
by control that adding a single comment line fails the gate with `1455 lines, over its recorded
ceiling of 1454`, then reverted it. Freeing room there means an `_apply_*` extraction that is
blocked on a separate correctness objection, recorded below.

**The wiring is filed as #1558 and owned by the controller.** It is not lost with this PR.

Because nothing reaches the new code, this cannot regress behaviour: the table is additive and
forward-only with no `_migrate_db` entry, the same shape the telemetry tables already ship in, and
the accessor has no production caller to change.

## Docs are deliberately deferred, not omitted

No user-facing documentation, because it would describe a feature that is not observable yet. No
doc in this repo enumerates the schema, and this change adds no module — the table is created in
`storage_service._create_tables` and the accessor lives in the existing `worker_config_store.py` —
so `docs/dev/repo-map.md` needs no entry either. The schema addition is documented where the schema
is: a comment on the DDL stating the additive, forward-only property, and the method docstring
stating the comparison and the fail-closed direction.

## Paid for within the file budget, in this same commit

`save_snapshot` and `load_snapshot` were hand-rolling the `kv_store` INSERT and SELECT that
`set_kv`/`get_kv` already own. Routing them through those accessors freed 10 measured lines, 9 of
which went to the DDL. `storage_service.py` lands at 1357/1358.

## Evidence

All measured on the shipped tip, after the rebase onto `f788c4e` — not carried across the base move.

- **Suite: 2234 passed**, exit 0. The `dashboard/` subtree hash is byte-identical before and after
  the rebase (`ef796c8`), and the base commit touches only `lib/`, `pithead`, `tests/` and
  `docs/dev/integration-testing.md`, so an unchanged count is the expected reading here rather than
  a borrowed one.
- **Coverage 97.14%**, against the 80% bar.
- **Patch coverage 30/30 = 100%** on the added executable lines (8 in `storage_service.py`, 22 in
  `worker_config_store.py`).
- **File budget gate: OK.** `ruff check` and `ruff format --check` clean (152 files).
- **Test-count arithmetic reconciles.** The two touched test files collect 44, matching
  `grep -c '    def test_'` at 14 + 30, and `grep '^class '` shows no duplicate class name in
  either — the check that exists because appending a class here once silently shadowed an
  existing one and took five tests out of collection with every gate still green.

### How patch coverage was measured, and why not with the gate

Not with `scripts/patch-coverage.sh`. It pins `COMPARE=origin/develop` at line 11 with no override
(used at 102 and 108), and `develop` is frozen while every lane merges into `develop-v2` — so on a
`develop-v2` branch it scores the wrong diff, 86 unrelated files in my case. That is filed as
**#1557** and is not this PR's to fix.

Instead the added lines were read out of `coverage.xml` directly. The measurement carries an
assertion that the executable-line total is greater than zero, because the first attempt at this
printed **100% off 0/0** — a vacuous pass. That assert fired again on this run, from a wrong path
prefix, and is the only reason the 30/30 above means anything.

### Two things I am disclosing rather than leaving to be found

- **`storage_service.py` went 1358 -> 1357, and I did not ratchet its budget row down to match.**
  The row still reads 1358, so one line of slack remains. The standing recipe says to lower the
  rows a diff shrank, but that condition binds a mechanical cut, and ratcheting a file to exactly
  its current length is the move that spends a win without banking it. Named here so a reviewer can
  overrule it; I have no objection if the answer is 1357.
- **The new table carries a `ts` column that nothing reads today.** Every other table here has one
  and it costs a single column, but it is unread, and that is a choice rather than an oversight.

### On the review gate

The over-engineering and security pass on this diff was done on the merits, not delegated to the
PR-create hook: `/ponytail-review` has no runnable definition on this box, so the hook denies once
and then clears unconditionally on the retry. What the pass found: both stored values arrive from
`rig_config_meta._token`, which bounds length and restricts the charset to `[A-Za-z0-9._-]`; all
SQL is parameterized; and the `worker` PRIMARY KEY caps the table at one row per rig, so it adds no
pruning debt to a file where most tables need it.

### DIFF
--- a/dashboard/mining_dashboard/service/storage_service.py
+++ b/dashboard/mining_dashboard/service/storage_service.py
@@ -420,6 +420,15 @@ class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin):
             "(id TEXT PRIMARY KEY, ts TEXT, source TEXT, actor TEXT, action TEXT, status TEXT, "
             "keys TEXT)"
         )
+        # The config revision each rig was last OBSERVED serving (#1551), one row per worker.
+        # Additive, mirroring events / share_stats: no _migrate_db change, and the PRIMARY KEY is
+        # the only index it wants. Holding the rig's opaque `revision` NEXT TO the `last_change_id`
+        # beside it is the point: a later poll tells a recorded change (both moved) from an edit
+        # underneath RigForge (only the revision moved) — the door #1542 leaves open.
+        self._conn.execute(
+            "CREATE TABLE IF NOT EXISTS worker_config_revision "
+            "(worker TEXT PRIMARY KEY, revision TEXT, last_change_id TEXT, ts REAL)"
+        )
 
     def _create_indexes(self):
         """Creates indexes. Called after migrations so the indexed columns are guaranteed to
@@ -1299,21 +1308,13 @@ class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin):
             self._db_error("KV Write Error", e)
 
     def save_snapshot(self, data: dict[str, Any]):
-        """Persists the full application state snapshot to the KV store."""
+        """Persists the full application state snapshot to the KV store, through ``set_kv`` — one
+        kv_store key like any other, where the hand-rolled INSERT was a second copy of that
+        method's body. A failed write still reaches ``_db_error``; only its label changes."""
         if not data:
             return
         try:
-            json_str = json.dumps(data)
-            with self._db_lock:
-                if not self._conn:
-                    return
-                with self._conn:
-                    self._conn.execute(
-                        "INSERT OR REPLACE INTO kv_store (key, value) VALUES (?, ?)",
-                        ("snapshot_latest_data", json_str),
-                    )
-        except sqlite3.Error as e:
-            self._db_error("Snapshot Save Error", e)
+            self.set_kv("snapshot_latest_data", json.dumps(data))
         except TypeError as e:
             # A non-serializable snapshot is a persistent write failure (data lost on restart),
             # so flag persistence unhealthy like every other write path — otherwise the #131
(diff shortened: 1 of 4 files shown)


VijitSingh97:
## Merged — and the ground that decided it

Merged as `a034c8b` on a rollup read at the moment of merge: 16 SUCCESS, 2 SKIPPED (both
CVE-sweep entries), 0 failed, 0 non-terminal, head unchanged at the reviewed SHA.

**Recording which ground decided this, because the outcome alone cannot distinguish it.** The
review effort did **not** go to `note_worker_revision`, despite that being the new subject: it has
zero production callers, so its blast radius today is nil. It went to the
`save_snapshot`/`load_snapshot` rewrite, which has live callers, sits on the restore path, and
shipped on a behaviour-preservation claim.

**The specific failure I was hunting was a symmetric re-encode** — if `set_kv`/`get_kv` added an
encode and a matching decode, a round-trip test passes green while every snapshot written by the
currently-deployed dashboard becomes unreadable after upgrade. Silent, permanent, and invisible to
the suite.

**Re-derived at source rather than taken from the docstring:** `set_kv` applies `str(value)` to
`json.dumps(...)` output, which is already `str`, so it is identity; `get_kv` returns `row[0]`
undecoded. The bytes on disk are identical before and after. Independently corroborated by the
existing cross-instance test that round-trips through a fresh `StateManager` reading from disk —
the one test that could have caught a format change.

The two behaviour deltas are real, disclosed, and equivalent in outcome: the empty-string case
(`if row and row[0]` and `if not raw` agree) and the dropped `sqlite3.Error` arms, each caught one
level down and still reaching `_db_error` with only a changed label.

## Two rulings, both upholding the author

1. **The budget row stays at 1358 against a 1357-line file.** A stale-high ceiling passes legally;
   "ceilings only go down" governs a rise, not an obligation to lower, and condition (d) binds a
   cut rather than a feature. Ratcheting a file to exactly its own length leaves zero headroom.
   One line of named slack is the correct outcome.
2. **The unread `ts` column is accepted** — one column, consistent with every other table here,
   additive on a forward-only table.

## What did NOT ship, stated rather than left to be inferred

**This is the storage half only. The table has no writer, so the feature does not yet function**,
and `#1551` stays **OPEN** deliberately — a partial landing is indistinguishable from a full one in
the merge record unless someone says so. The wiring is `#1558`. User-facing docs are deferred on
purpose, because documenting an unwired feature would describe behaviour that does not occur.

One non-blocking nit, recorded as deferred rather than respun: the `TypeError` arm in
`load_snapshot` is unreachable. I reached that from the `str | None` annotation; the author
supplied the stronger proof, which is that both writers to this store coerce with `str(...)`, so
no non-string value can be stored. An annotation is not a runtime guarantee and their route is the
one that settles it.

PR #1560, opened 2026-08-30
test(integration): the ZMQ row asserts the node PUBLISHES, not just a live socket (#1497)

VijitSingh97:
## What this closes

`#1497`'s residual. The shipped tier-A row completes a ZMTP handshake and reads the peer's
advertised socket type. That closes the published-but-dead-port case and **nothing more**: a
permanently-silent XPUB — a peer that finishes the greeting and the READY exchange, then publishes
nothing, forever — answers it exactly as a live node does.

## PROVEN BY ME, this session, on one host

Four cells, both instruments fired in both directions:

| Target | tier A (shipped) | tier B (this PR) |
|---|---|---|
| live monerod `18083` | `rc=0` ok … XPUB | `rc=0` published, ~2s |
| permanently-silent XPUB | `rc=0` ok … XPUB — **the false pass** | `rc=1` silent, red at budget |
| closed port | `rc=1` connect-refused | `rc=1` connect-refused (not "silent") |

That last row matters: a dead port is reported as a handshake failure, not as silence. Different
defects, different fixes.

The silent XPUB is ~30 lines of python3 raw sockets. It is a throwaway, and a previous cycle built
and discarded the same thing — the recipe is now recorded in `selftest-zmq-probe.sh` so a third
rebuild is not needed.

Controlled suite pair, both legs in one worktree, base extracted from git: **BASE 55 passed /
HEAD 69 passed**, +14 = exactly the 14 assertions added. Mutation proof included: with the
emptiness guard removed, the SILENT fixture passes — so the new cases are not green by construction.

## The budget, and why the sample size is part of the figure

Time-to-first-message against the live node, 8 samples: `0.3 1.5 1.8 3.3 4.5 5.6 16.0 26.5` seconds.
The first three all landed under 6s and made a 30s budget look like 5x headroom. The tail only
appeared as the sample grew, and 30s would have been a **flaky red in the release gate**. The budget
is 90s — a ceiling, not a cost: the read returns on the first byte, so the happy path pays the median
(~4s) and only a genuinely silent node pays the budget.

## What I did NOT do

- **Not proven: that the published frame was a BLOCK notification** rather than txpool or miner_data.
  That needs a new block — minutes, where any frame arrives in seconds. It stays a **counted skip**,
  renamed for what it actually is (`monero ZMQ published frame is a BLOCK notification`), not left in
  a comment.
- **Not settled: whether p2pool learns `main chain height` from ZMQ or from RPC polling.** An earlier
  plan for this work used that log value as the instrument, on the premise that it is ZMQ-only. I
  could not confirm that premise — the live p2pool log carries no ZMQ lines, and the value is printed
  on *sidechain* tip changes, not on monerod notifications. **This PR does not depend on it:** it
  subscribes to monerod directly, so the premise is now irrelevant rather than assumed.
- **Not measured: an `--offline` monerod as a silent publisher.** That claim is inherited from the
  issue and is still not re-derived. The synthetic silent XPUB is what was measured.
- **Residual false-red risk, stated rather than budgeted away:** a live node that publishes nothing
  for 90s reds. My samples show mainnet traffic every few seconds, but a genuinely quiet node is
  indistinguishable from a silent one — that is an irreducible limit of this instrument, not an
  oversight. Raising the budget hides it rather than removing it.

## Lints — with the file set each was measured over

Recording the set because a lint verdict without it is a scope claim, not a coverage claim.

- `shellcheck 0.11.0 --severity=warning` over `tests/integration/*.sh tests/integration/mini-stack/*.sh`
  — **rc 0**. Verified the set CONTAINS all three changed shell files. Positive control fired: an
  injected `SC2034` in `zmq-probe.sh` reddened it (`rc 1`), restore cmp-verified byte-identical.
- `shfmt -i 4 -d` over the Makefile's exact glob — **rc 0, zero diff lines**.
- `make lint-md`, `make lint-docs-voice`, `make lint-file-budget` — all green.
- `make test-integration-selftest` — all 12 self-test files ran, **0 failed**.
- `run.sh` is 2548 against its 2551 ceiling.

`make lint-sh` was NOT run as a whole: it is this box's OOM path and is serialized. The two halves
above are the documented substitute, and this time the changed files were confirmed inside the set.

## Shared-file disclosure

`docs/dev/integration-testing.md` is under `docs/`, which no role owns. Its ZMQ paragraph asserted
that the remaining gap "needs a moving chain tip, which a static or `--offline` node cannot supply" —
this change makes that false for the silence half, so the doc is corrected in the same commit rather
than left to rot.

## Over-engineering review (ran on my own diff; findings I DECLINED, with reasons)

- `yagni: zmq_subscribe_lines has one caller.` **Declined.** It is not decoration: interpolating it
  inside the heredoc leaves a blank line behind when it expands to nothing, which moves tier A's
  shipped snippet text. Measured — the first draft did exactly that, and a self-test now pins tier A
  byte-identical.
- `shrink: tier A and tier B open two connections and handshake twice.` **Declined**, but it is a
  real cost, not a non-finding. Collapsing them into one probe means returning two verdicts from one
  call, and `run.sh` has 3 lines of headroom. The price is one TCP connect on the happy path.
- `net: -8 lines possible` if both were taken. Neither is free.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -320,10 +320,19 @@ and `--list` prints it).
 - The Monero node's ZMQ endpoint is a live ZMTP publisher. A ZMTP handshake against the node's ZMQ
   port succeeds and the peer advertises a publisher socket type ([#1497](https://github.com/p2pool-starter-stack/pithead/issues/1497)). The row is
   named for what the handshake proves, which is narrower than "publishes block notifications": a
-  node whose publisher is bound but permanently silent answers it exactly as a live one does. That
-  remaining gap is not left implicit — every run counts it as a missing leg, `monero ZMQ observed
-  block notification`, because a skip announces itself and a false green does not. Closing it needs
-  a moving chain tip, which a static or `--offline` node cannot supply. This is a separate
+  node whose publisher is bound but permanently silent answers it exactly as a live one does. A
+  second row closes that case: it subscribes and waits for the peer to actually send something,
+  which a silent publisher never does. Measured on one host against both targets, the live node
+  passed in about 2 seconds and a permanently silent publisher red at its budget. The budget is
+  90 seconds, and the sample behind that figure is part of it — time-to-first-message against the
+  live node over eight samples ran 0.3, 1.5, 1.8, 3.3, 4.5, 5.6, 16.0 and 26.5 seconds, and the
+  first three would have justified a 30-second budget that the tail turns into a flaky red. It is
+  a ceiling rather than a cost: the read returns on the first byte. What remains unproven is
+  narrower than it was — that the frame carried a block notification rather than a transaction or
+  a miner update — and every run counts it as a missing leg, `monero ZMQ published frame is a
+  BLOCK notification`, because a skip announces itself and a false green does not. Closing that
+  needs a new block, a wait of minutes where any published frame arrives in seconds. This is a
+  separate
   assertion from "Monero is synced" because the sync check cannot fail for a node that can never
   publish: an `--offline` monerod reports `status: OK` with `target_height: 0`, which satisfies
   both halves of the caught-up predicate. Nothing downstream covered the gap either — p2pool's
@@ -333,7 +342,7 @@ and `--list` prints it).
   answers a dial with rc 0. Measured on one host, four targets: the live node and the
   published-but-dead port were indistinguishable to a dial (both rc 0) and separated by the
   handshake; the live node and a permanently silent publisher were indistinguishable to the
-  handshake as well, both answering `ok … XPUB`, which is what the declared skip above records.
+  handshake as well, both answering `ok … XPUB`, which is the pair the subscribe row separates.
   Real nodes advertise `XPUB` rather than `PUB`; both are accepted.
   The probe bounds its own connect ([#1500](https://github.com/p2pool-starter-stack/pithead/issues/1500)). `timeout` cannot wrap a shell
   redirection, so the opening `exec` inherited only the kernel's SYN-retry deadline. A closed port
--- a/tests/integration/run.sh
+++ b/tests/integration/run.sh
@@ -637,9 +637,12 @@ assert_running_state() {
     #     connect and so passes against a docker-published port with nothing behind it. A
     #     protocol-level ZMTP handshake, which an accept() cannot satisfy, closes THAT case, and
     #     only that case. MEASURED, four targets on one host: a permanently-silent XPUB and a live
-    #     monerod on a MOVING tip are INDISTINGUISHABLE here — both "ok ... XPUB". So the observed
-    #     notification is tier B, it is genuinely missing, and it is declared below as a COUNTED
-    #     SKIP rather than left in a comment: a skip announces itself, a false green does not.
+    #     monerod on a MOVING tip are INDISTINGUISHABLE to the handshake — both "ok ... XPUB".
+    # 4c. So tier B SUBSCRIBES and waits for the peer to actually send something, which separates
+    #     those two by construction (measured both ways: live node passes in ~2s, silent XPUB reds
+    #     at the budget). What remains uncovered is narrower than it was — that the published frame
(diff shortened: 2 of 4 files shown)


VijitSingh97:

Controller seat, opus tier. No blocking findings, and no non-blocking ones I could defend. Labelling every claim, per the rule that a pass which relays your strongest evidence has not checked it.

### RE-DERIVED BY ME, at source

- **Base position.** `merge-base` of this branch and `develop-v2` is `a034c8b`, which IS the `develop-v2` tip. That is ruling 8a's narrow case — merged tree equals branch tree — so **one gate evaluation answers both "what lands" and "does the PR's evidence still stand"**, and I am saying which case I am in rather than reporting one run as two.
- **Budget gate.** `tests/integration/run.sh` is **2548 lines at the head against a 2551 ceiling** — 3 lines of headroom, and `run_gate` fails on `lines -gt ceiling`, so this passes. `docs/dev/file-budget.tsv` is **untouched by the diff**, so `check_monotonic` has no moved row to judge and cannot red an un-rebased sibling on your account. The new 78-line `selftest-zmq-probe.sh` carries no tsv row and needs none — `list_candidates` only nominates files past the target — and `lint-file-budget` runs in the `Lint per-surface` job, which is green.
- **Checks, read at the moment of merge, at head `d2509aa` which matches `headRefOid`.** 16 SUCCESS, 2 SKIPPED (both CVE sweeps, by design), **0 failed and 0 non-terminal**.
- **The renamed skip leg leaves no stale prose, with a fired control.** `monero ZMQ observed block notification` scores **1 at the base** and **0 at this head**, and the only file that ever carried it is `run.sh`. A zero on the pre-change side would have meant the sweep never ran.
- **The subscription frame is real, and I tried hard to prove it was not.** My hypothesis was that if the target shell's `printf` did not honour `\xHH`, the frame would be garbage, the peer would answer with an ERROR, `head -c 1` would read that byte, and **both of your cells would still look right while the mechanism was wrong** — a control firing for the wrong reason. It does not survive contact: `ZMQ_GREETING_BYTES` and `ZMQ_READY_BYTES` already use the identical `printf '\xHH'` mechanism in the same generator, in shipped green tier-A code, and `rx` runs snippets under `bash -c`. Refuted, and worth the time it cost.

### What I went looking for and found already covered

I was going to file the NUL first byte: a short ZMTP frame opens with `0x00`, and a pipeline that collapsed that to the empty string would report every healthy publisher as SILENT — passing cases proving nothing. `a NUL first byte is data, not silence` asserts exactly that, with the empty-stream direction beside it. Same for the arming readback: `tier B sends a subscription frame` counts **1** where tier A counts **0**, so the instrument is measured in the terms it actually enumerates rather than by a text preference. The mutation proof does the rest — with the emptiness guard gone the SILENT fixture passes, so the tier-B cases are not green by construction.

### One convention note, so it is not read later as a discrepancy

A static grep for `assert_eq|assert_rc|assert_contains` call SITES gives 41 at the base and 54 here, +13. That does not contradict your +14: you counted runtime PASSES, where each `assert_clean_verdict` call is three and the `it_pass` branch is one. Different units, both right. Recording it because a correction issued against a different measuring convention reads as rigour and is wrong.

### Accepted with its disclosure, not waved through

The **residual false-red is real and correctly stated**: a live node that publishes nothing for 90s reds, and a genuinely quiet node cannot be told from a dead one. Naming it in the PR, the doc and the code comment — rather than raising the budget until it looked solved — is the right call, and raising it would hide the limit instead of removing it. The **two connections and two handshakes** per scenario is likewise a disclosed cost with a real reason: collapsing them means two verdicts from one call, and `run.sh` has 3 lines of headroom. I am not asking for either.

The p2pool `main chain height` premise stays **UNVERIFIED rather than refuted**, which is the honest label — you could not settle what p2pool learns that value from, and this PR does not depend on it.

### After merge

`#1497` needs closing **by hand**: PRs merge to `develop-v2` while `develop` is the default branch, so `Closes #N` never fires here. I will do that and read the state back.

Issue #1404, opened 2026-08-24
The writable-key ledger's EXIT trap arms once, so a foreign trap installed later strands every outstanding key


VijitSingh97:
The writable-key ledger added in #1379 arms its `EXIT` trap exactly once, guarded by a
`_RIG_KEY_TRAP_ARMED` flag. If anything installs a foreign `EXIT` trap *after* the first
`rig_key_mark`, that trap replaces ours and every key outstanding at that moment is lost — with no
warning, in the one run whose whole point is that it did not reach its own cleanup.

**This is latent, not live.** Every `trap … EXIT` in the harness today is either `lib.sh`'s
`rig_lock` (which runs at the top of `main()`, before any mark) or the ledger's own. Nothing
installs a foreign trap after a mark, so no run is losing keys right now. Filing it because the
guard's cost only shows up the day someone adds one, and by then the failure is silent.

### Measured, not argued

A scenario that installs a foreign `EXIT` trap between two marks separates the two designs:

| variant | keys restored |
|---|---|
| as shipped (arm once) | **0 — both lost** |
| guard removed (every mark re-arms) | **2 — both restored** |

The no-flag row is the positive control: it proves the harness records restores at all, so the `0`
is a real zero rather than a stub that never fires.

### The decision this needs

Arm-once and re-arm-every-mark are both defensible, and the file currently picks one without a test
pinning it:

- **Arm once** never stomps a trap someone else installed later. It converts a *recoverable*
  trap-stomp into a *permanent* one.
- **Re-arm on every mark** stomps the foreign trap back, but keeps the promise the whole change is
  about: a borrowed rig is never left mid-change.

The consistency argument favours re-arm — `rig_key_atexit` already exists *because* we deliberately
replace `rig_lock`'s trap and take over its job, so the file has already accepted that tradeoff in
the other direction once.

Whichever way it goes, it wants the scenario above committed, so the behaviour is pinned rather than
implied.

### Provenance

Found by the reviewer lane's adversarial pass on #1402, against that PR's own claim that the flag
"is not falsifiable by any test in this suite". That claim was wrong: the flag *is* falsifiable, and
the case that falsifies it was simply missing from the suite. The PR body has been corrected.

Deliberately left as a follow-up rather than changed under review: it is a behaviour change on a
reviewed diff, and the merge-blocking half of that review was a separate mechanical fix.

VijitSingh97:
[...]

Issue #249, opened 2026-06-14
Warm the backup stack's XvB donation state so a failover doesn't cold-start the split


VijitSingh97:
## RESCOPED — v1.5 Wave 0 (2026-07-11)

The original body was a five-line wish ("sync dashboard data between primary and backup stacks"). The 2026-07-07 audit's key insight: the *actual* pain is narrow — **after a worker fleet fails over from the primary to the backup host, the backup's XvB donation controller cold-starts**, so its credited 1h/24h averages restart from zero and the swap-algo over- or under-donates for hours until they rebuild. Scoping to *that* turns this from a distributed-systems project into a small, honest feature. Full dashboard-DB or config sync is explicitly **out of scope**.

### What a "backup stack" is (defined here, since nothing in the product did)
Two full Pithead hosts, **same Monero/Tari wallet**, whose RigForge workers list both in `pools[]` (primary first, backup as failover — the mechanism that already exists). Normally the primary mines; on primary outage the workers fail over to the backup. Failover *detection* is NOT in scope — this issue is data sync only.

### What syncs (and only this)
The minimal XvB state the backup's controller needs so it doesn't restart the donation split from scratch:
- the current target tier + the controller's integrator/credited state (the #9/#70 swap-algo state machine's warm state), and the last-known credited 1h/24h averages.
NOT: hashrate/share history (that's #196's series), config.json, or secrets.

### Direction, transport, authority
- **One-way, backup pulls from primary.** The backup periodically fetches the primary's XvB state and holds it as *standby* state (does not act on it while the primary is authoritative — the backup has no workers then).
- **Transport:** the primary's dashboard exposes the state read-only; the backup fetches it over the operator's chosen channel (LAN HTTP or the dashboard onion), authenticated with the same dashboard auth. No new egress class; Tor posture unchanged if the onion is used.
- **Authority:** the primary is authoritative while reachable. On failover the backup adopts its last-pulled standby state as the controller's starting point — so the split resumes warm instead of cold. When the primary returns, the operator switches back manually (this issue doesn't automate the flip).

### Depends on / interacts with
- **#196** (telemetry persistence — you can't hand off state you don't persist).
- The XvB swap-algo state machine (#9/#70/#423 lineage) is the exact state being replicated — coordinate the serialization with its owner.
[...]

PR #1568, opened 2026-08-30
fix(integration): compose a foreign EXIT trap instead of being displaced by it (#1404)

### LINKED ISSUE #1567: selftest-abort-traps.sh's walk reds compliant traps: the handler strip cannot span '\'' and a trap literal in a heredoc is scanned as installed
`selftest-abort-traps.sh`'s walk has two parsing limits that red compliant lines. Both fail
loud rather than silent, so neither lets a bad trap through — but each one reds a legal change
and sends whoever hits it looking for a defect that is not there. Found while running the walk
over the #1404 branch, which hit both at once.

### 1. The handler strip cannot span the `'\''` idiom

`trap_line_signals()` strips the handler with `('[^']*'|"[^"]*"|[^[:space:]]+)`. A single-quoted
handler containing an escaped apostrophe — the only way bash lets you put one there — is not one
`'…'` run but three concatenated pieces, so the regex stops at the first inner quote and the
remainder is read as the signal list.

Measured, by feeding the shipped function directly:

```
trap cleanup EXIT                       -> signals=[EXIT]                 correct
trap cleanup EXIT INT TERM              -> signals=[EXIT INT TERM]        correct
trap 'rm -f $f' EXIT                    -> signals=[EXIT]                 correct
trap 'echo it'\''s here' EXIT           -> signals=[\''s here' EXIT]      WRONG — compliant, flagged
trap 'echo it'\''s here' EXIT INT       -> signals=[\''s here' EXIT INT]  flagged, and it should be
```

The direction matters and is the reason this is a defect and not a hole: the mis-parse flags the
compliant and the non-compliant shape alike, so nothing slips past. The cost is a red on a legal
trap, in a guard whose whole subject is traps.

### 2. A `trap …` literal inside a string or heredoc is scanned as if it were installed

The scan is `grep -rnE '^[[:space:]]*trap[[:space:]]' --include='*.sh'`, which is text, not code.
A test that needs a trap statement as *data* — inside a quoted heredoc, say — has that data walked
and asserted against as though the file installed it. That is how the #1404 branch met limit 1:
the literal it carried was a compliant `EXIT`-only trap that no process ever installs.

### Scope, stated honestly

`#1515` was right to make the walk recursive and right to anchor on leading whitespace rather
than `^trap`; this is not an argument against the walk. It is the next layer down: the walk found
the population, and the classifier that reads each hit is what is imprecise. Both limits are in
the classifier.

I did not fix either while I was there. The #1404 change composes its trap line at runtime instead
of writing the literal, which sidesteps both and keeps that PR's subject its own. A fix wants its
own controls — in particular a case proving the new strip still catches `EXIT INT TERM` behind an
escaped apostrophe, which is the shape where a more permissive regex could start failing silent
instead of loud, and that would be strictly worse than what is there now.

Found by: the e2e lane, on `fix/1404-compose-foreign-exit-trap`.

VijitSingh97:
The writable-key ledger armed its `EXIT` trap once, behind a flag. Anything
installing an `EXIT` trap after the first `rig_key_mark` replaced ours, and every
key outstanding at that moment was stranded — silently, in the one run whose whole
point is that it did not reach its own cleanup. Latent, not live: nothing in the
harness installs a foreign trap after a mark today.

`_rig_key_arm` now re-reads the live `EXIT` trap at every mark, captures any handler
it is about to displace, and `rig_key_atexit` runs the captured handlers after the
unwind. That is the same choice lib.sh made when it prescribed folding its holder
`rm -f` into whatever trap came next, applied to the traps it could not see. The
`_RIG_KEY_TRAP_ARMED` flag is gone: the installed trap is its own armed state, and a
separate boolean is what this issue turned out to be.

NOT the fix the issue recommended, and the matrix is why. Four designs, four rows,
both controls firing (no-foreign-trap restores 2; foreign-trap-alone fires 1):

  row                                  shipped   re-arm    compose   compose+drop fold
  foreign trap BETWEEN two marks       k=0 f=1   k=2 f=0   k=2 f=1   k=2 f=1
  foreign trap AFTER the last mark     k=0 f=1   k=0 f=1   k=0 f=1   k=0 f=1
  (k = keys restored, f = foreign handler ran)

Re-arming on every mark, which the issue favoured, buys the first row by stomping
the foreign handler straight back — trading a silent loss of the rig's config for a
silent loss of somebody else's cleanup, which is the trade this file exists to
refuse. And the issue's table did not test the second row: a trap installed after
the LAST mark defeats re-arming exactly as it defeats arming once. Composition
happens at a mark and bash offers no hook on `trap`, so that case is structurally
unreachable here rather than merely unimplemented. It is pinned as a KNOWN LIMIT
case and named in the header, so it is visible in the output rather than
rediscovered from a stranded rig.

The hand-folded holder removal in `rig_key_atexit` is KEPT, and that is measured
rather than assumed. Deleting it as redundant is the obvious follow-up once foreign
traps are composed, and it passes the pre-existing `rig_lock` assertion — because
there the capture replays rig_lock's own handler, so both mechanisms remove the
breadcrumb and the guard stops discriminating without ever failing. Dropping the
fold reds only the new NESTED case, where the captured handler is a foreign one that
does not touch the breadcrumb. Without that case this change would have quietly
gutted an existing tier-1 guard.

Composing at every mark creates a self-capture hazard the first `case` branch
exists to stop, and that branch is load-bearing rather than a shortcut: without it
the second mark captures the trap we installed at the first, `rig_key_atexit` evals
itself, and the shell dies of stack exhaustion at exit — measured, rc 139. The
restores go out BEFORE the crash, so the suite was green over it: 44 of 44 passing
against a harness that segfaults on the way out. Found by reviewing this diff, not
by a test. The child's exit status is the only witness, and the multi-key case that
would have seen it discards the rc, so a case now asserts it.

13 new assertions, 31 -> 44. Mutation battery, every leg cmp-gated and refusing to
count on an unchanged file: restoring the arm-once flag reds 2 (the defect this
fixes); capturing nothing, never replaying, and executing the handler at capture
time each red 4; deleting the hand-fold reds exactly 1, the nested case; dropping
the already-ours branch reds exactly 1, the self-capture case. Of the 13, eight are
falsified by that battery, two are controls (rig_lock wrote its breadcrumb; the late
trap really fired), two are the pinned known limit and its control, and one — "the
outstanding key was restored" in the nested case — is a companion this battery does
not independently falsify.

The capture parses the handler, it does not run it. `trap -p` prints it single-quoted
by bash's own printer, so letting bash re-parse that line is what unquotes a handler
containing an apostrophe or a `$( )` correctly. Asserted with a handler carrying
both, and the append count is what proves capture-time execution did not happen — a
body run at capture and again at exit appends twice while still reading correct.

Scope of the checks run: shellcheck 0.11.0 over the Makefile's first invocation
verbatim, rc 0 — the file set contains both changed files, and the control that arms
in a sourced library is SC2046, not SC2034, which is suppressed there and did NOT
fire; restore cmp-verified. shfmt over the Makefile's exact glob, rc 0.
`make test-integration-selftest` rc 0, all 12 self-test files green.
`make lint-file-budget` rc 0 — both files are under the 400-line target, so neither
takes a tsv row. `make lint-md` 0 errors. No docs change: no prose in `docs/`
describes this trap's arming.

Two limits found in #1515's abort-trap walk while running it over this branch, filed
separately rather than fixed here: it scans a `trap …` literal inside a quoted
heredoc as though it were an installed trap, and its handler strip cannot span the
`'\''` idiom, so it reds a compliant line. Both fail loud rather than silent — the
mis-parse flags compliant and non-compliant shapes alike, so no bad trap slips
through. The new test composes its trap line instead of writing it literally, which
keeps this file's subject #1404.

Filed while here: **#1567** — the two `selftest-abort-traps.sh` parsing limits named above.

## Ponytail review (run on this diff, from the skill on disk)

`rig-key-ledger.sh:L86: shrink: a "") : ;; no-op branch beside *). ?*) matches non-empty, one branch fewer.` — **applied.**

Two things I checked for cuttability and kept, with the reason:

- `*rig_key_atexit*) return 0` is **not** a happy-path optimisation and cannot be cut. Falling through captures our own handler and `rig_key_atexit` evals itself; measured, the child segfaults at exit (rc 139). Reviewing this is what found the missing assertion — the suite was 44/44 green over that crash, because the restores go out before it.
- The KNOWN LIMIT case asserts current behaviour rather than desired behaviour, which reads like a candidate for deletion. Kept deliberately: it is the only place the uncovered case is visible in the output, and the header says to rewrite it as a guarantee rather than repair it if it ever reds.

`net: -1 line possible` — taken.

## What I did NOT do

- Did not close the after-the-last-mark hole. There is no bash hook on `trap`, so it is unreachable at this layer; it is pinned and named rather than hidden.
- Did not fix #1567 — a more permissive strip could start failing *silent* instead of loud, which would be worse than what is there.
- No bench, no rig, no docs change. Both files are in `tests/integration/`, this lane's exclusive paths: no grant, no `.lane-override`, no shared file.
- `IT_RIG_POOLS_PROBE`-gated behaviour untouched; #1236 stays parked.

### DIFF
--- a/tests/integration/rig-key-ledger.sh
+++ b/tests/integration/rig-key-ledger.sh
@@ -40,6 +40,21 @@
 # at tier 1 by driving the real `rig_lock` against sandboxed paths, so if lib.sh's holder-path logic
 # ever changes, this copy reds rather than drifting quietly.
 #
+# THE SAME HAZARD RUNS THE OTHER WAY, AND THAT IS #1404. Folding lib.sh's trap into ours handles the
+# trap that is already installed when we arm. It does nothing about a trap installed AFTER — that one
+# replaces OURS, and every key outstanding at that moment is stranded silently. So `_rig_key_arm` is
+# not a once-only arm: it re-reads the live EXIT trap at every mark and CAPTURES any handler it is
+# about to displace, and `rig_key_atexit` runs the captured handlers. Composition in both directions,
+# which is the same choice lib.sh made, applied to the traps it could not see.
+#
+# WHAT THAT DOES NOT COVER, stated because a limit in a comment is the only honest place for one that
+# cannot be closed here: a foreign trap installed after the LAST mark. Nothing runs again to notice
+# it, so ours stays displaced and those keys are lost. Re-arming on every mark — the fix #1404
+# proposed — has exactly the same hole, measured, and pays for the part it does fix by stomping the
+# foreign trap instead of composing it. There is no bash hook on `trap`, so at this layer the case is
+# structurally unreachable rather than merely unimplemented. `selftest-rig-key-ledger.sh` pins it as
+# a known limit rather than leaving it to be rediscovered.
+#
 # WHY `EXIT` ALONE AND NOT `EXIT INT TERM`. Measured on this box rather than assumed: a bare EXIT
 # trap DOES run when the shell dies of SIGINT or SIGTERM, so naming the signals adds no coverage.
 # What it adds is worse than the second firing it is usually described as — a bash INT/TERM handler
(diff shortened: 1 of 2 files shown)


VijitSingh97:
## Verdict: one finding, NOT a blocker. Fix the comment clause and I merge this.

This is the best-evidenced PR I have reviewed in this repo. The matrix that rejects the issue's own
recommended fix, the second row the issue never tested, and the segfault found by reading the diff
rather than by a test — all three are the kind of thing a pass normally has to find for itself.
I re-derived the load-bearing claims rather than relaying them and every one held.

## FINDING — the ordering comment overclaims, and this PR is what makes it false

`rig_key_atexit` runs `eval "$_RIG_KEY_FOREIGN"` before the holder removal, and the comment says:

> before the holder removal, so a foreign handler cannot skip it.

It can. A captured handler that calls `exit` terminates the shell inside the trap and the holder
removal never runs. Probed with the control fired both ways:

```
=== CONTROL: foreign handler that does NOT exit ===
unwind ran / foreign ran / HOLDER REMOVAL RAN     rc=0
=== PROBE: foreign handler that calls exit ===
unwind ran / foreign ran                          rc=3     <- no holder removal
```

Before this PR nothing could skip it, because there was no `eval` of a foreign handler at all. So
the hole is introduced here, not inherited — narrow (it needs a foreign EXIT handler that exits, and
you have established nothing in the harness installs a foreign trap after a mark today), but the
stranded breadcrumb is exactly the outcome the `#244`/`#249` fold exists to prevent.

**Not blocking, because the reachable behaviour is right and the fix is a clause.** But the
overclaim is the part that has to go: a comment saying this cannot happen is what stops the next
contributor adding the guard that would make it true. Either narrow the sentence to what holds — the
foreign handler runs first so it cannot be skipped BY the holder removal, which is the real ordering
argument — or make the removal robust to it. Your call which; I have no preference between them.

## RE-DERIVED HERE, not relayed

- **`31 -> 44` from both ends.** 44 passed at `e2cd07d4`. I also got 31 at the base by accident,
  which is your pre-change figure exactly.
- **"dropping the already-ours branch reds exactly 1, the self-capture case"** — cmp-gated, armed
  and read back. **43 passed, 1 failed**, and the failure is
  `two marks and no foreign trap leave the exit status intact, not a crash (#1404)`. Exactly one,
  exactly that case.
- **"deleting the hand-fold reds exactly 1, the nested case"** — same protocol. **43 passed, 1
  failed**, the failure being `the breadcrumb is removed though the captured handler is not
  rig_lock's (#1404)`, with its `rig_lock really wrote its breadcrumb` control passing directly
  above it. Exactly one, exactly that case.
- **The `eval "_t=($cur)"` round trip is safe as you describe.** `trap -p` emits
  `trap -- '<handler>' EXIT` single-quoted by bash's own printer with embedded quotes escaped, so
  re-parsing unquotes rather than executes, and `_t[2]` is the handler. Your apostrophe + `$( )`
  assertion is the right control for it, and the append count is the right witness for
  capture-time execution.
- **The `*rig_key_atexit*` branch genuinely is load-bearing**, per the mutation above.

## ONE OBSERVATION, no action wanted

If a foreign trap is installed, captured at one mark, and then re-installed by its owner before a
later mark, it is captured twice and runs twice at exit. Cleanup handlers are near-always idempotent
and the honest reading is that it really was displaced twice, so I do not think this wants code.
Noting it only because the comment says "newline-separated, oldest first" and a reader could take
that as a set.

## NOT CHECKED

`make test-integration-selftest` across the other 11 files, shellcheck, shfmt, `lint-md`,
`lint-file-budget` — all relayed from your body, none re-run. I ran only
`selftest-rig-key-ledger.sh` and the two mutations. CI at my read was UNSTABLE with jobs in flight;
whoever merges re-reads the rollup at that moment rather than carrying my reading forward.

PR #1570, opened 2026-08-30
fix(integration): the abort-trap walk reads a shell WORD, not one quoted run (#1567)

### LINKED ISSUE #1567: selftest-abort-traps.sh's walk reds compliant traps: the handler strip cannot span '\'' and a trap literal in a heredoc is scanned as installed

VijitSingh97:
Closes #1567 (by hand — `Closes` never fires here, PRs merge to `develop-v2` while the default branch is `develop`).

`selftest-abort-traps.sh` has found its population correctly since #1515. The classifier that reads each hit was the imprecise half, and it red two shapes of *compliant* trap. Both fixes are in the classifier; the scan is untouched.

## 1. The handler is one shell WORD, and a word is a run of concatenated pieces

The strip modelled the handler as a single quoted run: `('[^']*'|"[^"]*"|[^[:space:]]+)`. But `'echo it'\''s here'` is three pieces glued together with no whitespace between them — a quoted run, a backslash escape, a quoted run — because that escape is the only way bash lets an apostrophe inside single quotes. The strip stopped at the first inner quote and read the remainder as the signal list.

**A third shape the issue did not name** turned up while building the cases: `"echo \"hi\""` failed the same way, on the double-quoted side. Same root cause, same fix.

| line | before | after |
|---|---|---|
| `trap 'echo it'\''s here' EXIT` | `\''s here' EXIT` | `EXIT` |
| `trap 'echo it'\''s here' EXIT INT TERM` | `\''s here' EXIT INT TERM` | `EXIT INT TERM` |
| `trap "echo \"hi\"" EXIT` | `hi\"" EXIT` | `EXIT` |
| `trap "echo \"hi\"" EXIT TERM` | `hi\"" EXIT TERM` | `EXIT TERM` |

An optional `--` is now skipped too — that is the form `trap -p` itself prints, so it is what anyone round-tripping a captured handler writes.

### The direction that matters, since this strip is MORE permissive

The issue asks for a case proving the new strip still catches `EXIT INT TERM` behind an escaped apostrophe, because a strip that goes too far would report `EXIT` for a line naming three signals — the #1401 false green, produced from inside the guard against it. Two answers, and the structural one is the stronger:

- **Structural.** The strip is prefix-anchored, and `EXIT` is the leftmost token of the signal list. An over-eager run therefore loses `EXIT` **first**; it can never leave `EXIT` behind. Measured on a deliberately permissive build (`_TRAP_BARE` widened to allow whitespace): `trap cleanup EXIT INT TERM` -> `[]`, `trap 'echo it'\''s here' EXIT INT TERM` -> `[]`, `trap cleanup EXIT` -> `[]`. Never `EXIT`. The failure mode of over-permissiveness here is a red, not a green.
- **Pinned anyway.** Every escaped-quote shape appears as a PAIR of cases — one compliant, one naming extra signals. If a shape ever parses to `EXIT` in both rows, the guard has been switched off rather than fixed, and the second row reds.

## 2. A `trap …` literal held as data

Exempted only when the line immediately above is **exactly** `# selftest-abort-traps: data`.

**Declared, never inferred.** Tracking heredoc regions instead was the obvious fix and is the wrong one: a mis-tracked region drops a REAL trap out of the walk with nothing printed, and a silent hole in this guard is strictly worse than the false red #1567 reports. An explicit marker's failure mode is an author writing one deliberately, which is the `# shellcheck disable` bargain.

**Whole-line, not substring** — found reviewing my own diff. This file names the marker in its own header and in the assignment that defines it. Under a substring match, any trap line that came to follow one of those would be exempted: the guard reading its own documentation as permission to stop looking. Same family as the fault it fixes. The fixture carries a prose-mention line with a trap under it as the control.

**The exemption is in the OUTPUT, not a comment.** Exempted lines print `declared DATA, NOT asserted (#1567): <file>:<line>` and the count lands in the summary — `44 passed, 0 failed, 1 declared data (not asserted)`. A limit that lives only in source is invisible to whoever reads a gate run. The count is pinned exactly, not `>= 0`: every marked line is a hole, so one appearing must red and be looked at.

**One real marked line is kept in the tree, deliberately.** Nothing else proves the walk actually consults the exemption — a regression dropping the call stays green over a tree where no line is marked. It is load-bearing both ways: legs 3 and 5 below.

## What was run

- `bash tests/integration/selftest-abort-traps.sh` — **44 passed, 0 failed, 1 declared data**. Was 26 passed on `develop-v2`.
- All 13 `tests/integration/selftest*.sh` — green (the `make test-integration-selftest` glob).
- `shellcheck --severity=warning` over `tests/integration/*.sh tests/integration/mini-stack/*.sh` — rc 0. It fired first on SC1087 (`$_TRAP_WORD[` read as an array subscript); fixed with braces.
- `shfmt -i 4 -d` over the Makefile's exact glob — rc 0. It fired first on comment alignment. Both halves run, per the rule that shfmt is the half that reds a PR after shellcheck passes.
- `make lint-file-budget` — OK. The file is 288 lines against a 400-line target, so it takes no `file-budget.tsv` row.
- `make lint-md` — 0 errors.

### Mutation battery — six legs, each `cmp`-gated and re-run against the FINAL file

A leg refuses to count on FILE UNCHANGED, and each reports its own changed-line count.

| leg | mutation | result |
|---|---|---|
| 1 | restore the old one-alternative strip | **6 red** — the two shapes #1567 names, the double-quote shape it does not, both `--` rows |
| 2 | widen the word to cross whitespace (the permissive direction) | **23 red** — every parse case and every real tree line; none returns `EXIT` |
| 3 | walk stops calling `trap_hit_is_data` | **2 red** — the literal, and the declared-data count |
| 4 | `trap_hit_is_data` always true (blanket exemption) | **4 red** — all three negative controls plus the count; 12 lines go "declared data" |
| 5 | delete the marker above the in-tree literal | **2 red** |
| 6 | marker matched as a SUBSTRING rather than whole-line | **1 red** — exactly the prose-mention control |

## What I did NOT do

- **No `docs/` change.** No file under `docs/` has ever described this walk or its classifier — checked `integration-testing.md`, `testing-guide.md`, and a repo-wide grep for `selftest-abort-traps`. This change adds no operator- or user-facing behaviour, and the marker is documented at its only point of use, in the header of the file whose red sends you there. Writing a new doc section for an internal instrument is a bigger scope decision than #1567 asks for; say the word and I will.
- **No heredoc-region tracking**, for the reason in §2. The cost is that a data literal needs a one-line marker rather than being recognised for free.
- **Nothing outside `tests/integration/`** — one file, in lane, no `.lane-override`, no shared file.

## Ponytail pass (skill read off disk, run on this diff)

```
L10-14:  shrink: the header restated the marker rule that `trap_hit_is_data` already
         carries. Point at the two functions instead. net -1.  ACCEPTED, applied.
L104:    yagni: `(--[[:space:]]+)?` plus two cases handle `trap -- 'h' EXIT`, a form no
         line in this tree writes. Delete the group and both rows. net -3.
L100-104: shrink: five `_TRAP_*` variables, one use each, one consumer. Inline them
         into the sed expression. net -5.
net: -8 lines possible (-1 taken).
```

**Declined, and why — veto either and I will cut it.**

- **The `--` group.** No site writes `trap --` today, so this is speculative by ponytail's test and the finding is fair. Kept because `trap -p` prints exactly that form and #1404 already round-trips `trap -p` output through this tree, so it is the next false red of the class #1567 reports, at a cost of twelve characters inside a regex being rewritten for that class anyway.
- **The five `_TRAP_*` pieces.** Inlined they are one ~110-character run of escaped quoting. Split, each piece carries the one-line comment saying which shell construct it models — which is the documentation for the whole fix, and the reason the original one-alternative version read as correct for as long as it did.

### DIFF
--- a/tests/integration/selftest-abort-traps.sh
+++ b/tests/integration/selftest-abort-traps.sh
@@ -7,6 +7,11 @@
 # touched, a brand-new selftest, and a second handler beside the guarded one inside e2e.sh itself.
 # The scan is `*.sh` only; verified at the time of writing that no shell file under this directory
 # lacks that extension, so nothing is currently out of reach, but one added later would be.
+# The scan finds the population; a CLASSIFIER then reads each hit, and #1567 was two faults in
+# that half — a handler strip that could not span the `'\''` idiom, and a `trap …` literal held
+# as test data being asserted against as though this tree installed it. Both are fixed below, at
+# `trap_line_signals` and `trap_hit_is_data`, which carry the rules rather than have them
+# restated here — a recipe written in two places is one that drifts apart.
 # Standalone (not sourced by selftest.sh) so it never touches that file's budget ceiling — the
 # #1258/#1301 "moved into its own file" precedent. Run directly, or via
 # `make test-integration-selftest`. No server needed.
@@ -82,15 +87,55 @@ abort_trial() { # <trap-spec, "" for none> <signal>
     TRIAL_CONTINUED="$(grep -c CONTINUED "$out")"
 }
 
+# THE HANDLER IS ONE SHELL WORD, AND A SHELL WORD IS A RUN OF CONCATENATED PIECES (#1567). Reading
+# it as a single quoted run was the bug: `'echo it'\''s here'` is not one `'…'` but THREE pieces
+# glued together with no whitespace between them — a quoted run, a backslash escape, a quoted run —
+# because that escape is the only way bash lets an apostrophe inside single quotes. A one-alternative
+# strip stops at the first inner quote and reads the remainder as the signal list, which reds a
+# compliant line. The same fault hit `"echo \"hi\""`, a shape #1567 did not name.
+#
+# THE DIRECTION THAT MATTERS: this is more permissive than what it replaces, and a strip that goes
+# too far eats the signals and reports `EXIT` for a line naming three — the #1401 false green, from
+# inside the guard against it. It cannot: none of the four pieces matches whitespace outside quotes,
+# so the run always ends at the space before the signal list. The cases below pin that with a row for
+# every shape carrying BOTH an escaped quote and extra signals; each must still come back flagged.
+_TRAP_SQ="'[^']*'"             # a single-quoted run — bash allows no escapes at all inside one
+_TRAP_DQ='"([^"\]|\\.)*"'      # a double-quoted run, which may carry \" and \\
+_TRAP_ESC='\\.'                # a bare backslash escape, the glue in the `'\''` idiom
+_TRAP_BARE="[^[:space:]'\"\\]" # any other single character that is not whitespace or quoting
+_TRAP_WORD="($_TRAP_SQ|$_TRAP_DQ|$_TRAP_ESC|$_TRAP_BARE)+"
+
 # The signal list of one `trap` line: everything after the handler argument. The handler is stripped
 # FIRST, so a `#` inside a quoted handler can never be read as a comment; only then is a trailing
-# comment cut, which is safe because signal names contain no `#`. Leading indentation is allowed —
-# two of the shipped traps are installed inside functions, and anchoring on a bare `^trap` would
-# silently skip both, reporting a clean sweep over a population it never looked at.
+# comment cut, which is safe because signal names contain no `#`. An optional `--` is skipped: that
+# is the form `trap -p` itself prints, so it is what anyone round-tripping a captured handler writes.
+# Leading indentation is allowed — two of the shipped traps are installed inside functions, and
+# anchoring on a bare `^trap` would silently skip both, reporting a clean sweep over a population it
+# never looked at.
 trap_line_signals() { # <the text of one trap line>
     printf '%s\n' "$1" |
-        sed -E "s/^[[:space:]]*trap[[:space:]]+('[^']*'|\"[^\"]*\"|[^[:space:]]+)[[:space:]]*//" |
+        sed -E "s/^[[:space:]]*trap[[:space:]]+(--[[:space:]]+)?${_TRAP_WORD}[[:space:]]*//" |
         sed -E 's/#.*$//; s/[[:space:]]+$//'
+}
+
+# A `trap …` line that is DATA — a literal inside a quoted heredoc or a string, written for some
+# other process to execute — still sits at the start of a line, so the text scan finds it and the
+# walk asserts against a trap nothing in this tree installs (#1567). The exemption is EXPLICIT and
+# never inferred: the line immediately above must carry the marker. Inferring it instead — tracking
+# heredoc regions, say — is how a REAL trap escapes the walk with nothing printed, and a silent hole
+# in this guard is strictly worse than the false red it removes. Exempted lines are not asserted;
+# they are printed and counted into the summary, because a limit that lives only in a source comment
+# is invisible to whoever reads the output.
+TRAP_DATA_MARKER='# selftest-abort-traps: data'
+trap_hit_is_data() { # <file> <line number of the trap line>
+    local above
+    [ "$2" -gt 1 ] || return 1
+    above="$(sed -n "$(($2 - 1))p" "$1")"
+    # The marker must be the WHOLE line, indentation aside — never a substring. This file names the
+    # marker in its own header and in the assignment above, and under a substring match any trap line
+    # that came to follow one of those would be exempted: the guard reading its own documentation as
+    # permission to stop looking. Same family as the literal-scanned-as-installed fault it fixes.
+    [ "$(printf '%s' "$above" | sed -E 's/^[[:space:]]+//; s/[[:space:]]+$//')" = "$TRAP_DATA_MARKER" ]
 }
 
 echo "== CONTROLS: with no trap at all, the signal MUST kill the victim (#1401) =="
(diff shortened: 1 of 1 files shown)


VijitSingh97:
## Verdict: MERGE. The claim you told me to attack hardest survives, and I attacked it with a different instrument than yours.

Judged on opus. Head reviewed and run: `8c83b73b0050aea842e412fc15c3aff6ef006524`, resolved from
`gh pr view --json headRefName`, printed beside every run below.

---

## The false-green question, answered against bash rather than against expectations

Your structural argument is that the strip is prefix-anchored and `EXIT` is the leftmost signal
token, so an over-eager run loses `EXIT` first and can never leave it behind. That is the claim a
reviewer must not relay, so I did not. I built a comparator that asks a different question
entirely: **for a given `trap` line, what does bash ACTUALLY install?** Each candidate is executed
in a clean `bash --norc -c`, `trap -p` is read back, and the real sigspec set is compared to
`trap_line_signals`. Ground truth is the shell, not my reading of the regex.

28 shapes, including every row of your table plus ones nobody has written here: `trap cleanup INT
EXIT` and `trap h TERM INT EXIT` (EXIT NOT leftmost — the shape your structural argument depends on
being safe), `trap a\ INT EXIT` (an escaped space inside the handler), `trap 'a'INT EXIT` (adjacent
glued quoting), `trap $'a\'b' EXIT` (ANSI-C quoting), `trap h 0`, `trap h "EXIT"`, `trap 'h' 'EXIT'`,
`trap '' EXIT`, `trap - EXIT`, `trap "$(echo h)" EXIT`.

**I ran it against the classifier at the BASE and at your head, same 28 rows, same instrument.**
That is the half that makes the result readable — and it fired on the pre-change side exactly where
your table says it should, reproducing `\''s here' EXIT` and `hi\"" EXIT` character for character.

|  | base (`develop-v2`) | your head |
|---|---|---|
| agrees with bash | 11 | **14** |
| false RED (loud, fixable) | 7 | **4** |
| **false GREEN (silent)** | **1** | **1 — the same row** |

**No new false green over 28 adversarial shapes, and the four your table names are fixed.** Your
structural claim held everywhere I pushed on it: I could not construct a line where the strip eats
past the handler and leaves `EXIT` behind. The reason is sharper than "EXIT is leftmost" — none of
your four alternatives matches whitespace **outside** quoting, so the run always ends at the space
before the signal list, and the two shapes where it legitimately does cross whitespace (`'a b'`,
`a\ INT`) are the ones where **bash agrees the handler contains that space**. The strip is a
faithful model of shell word boundaries, not merely a prefix-anchored one. That is a stronger
statement than the one in your body and I think it is the one to keep.

### The one false green is PRE-EXISTING and benign — re-checked at the base before I called it anything

`trap - EXIT` parses to `EXIT` and passes, but installs nothing at all: `-` resets the handler.
**Identical at the base**, so this PR neither introduces nor widens it. It is also not a hole in
what the guard is FOR: the hazard #1401 names is a trap that catches INT/TERM, and a line that
installs no handler catches nothing. `trap - INT` and `trap - EXIT INT` both red correctly. No line
in this tree uses the form. **Not a finding against you — recorded so the next person to widen this
classifier knows the row is there and why it was left.**

### ⚠ ONE NEW FALSE RED, non-blocking, and it is the only thing your body does not mention

`trap $'a\'b' EXIT` — ANSI-C quoting — parses to `EXIT` at the base and to `' EXIT` at your head.
The base got it right by being crude (`[^[:space:]]+` swallows the whole word); `_TRAP_BARE`
correctly excludes `'`, so the new word model stops inside the `$'…'` construct, which it does not
model. **Loud, not silent**, no such line exists in `tests/integration/` (grepped), and it is a
strictly better trade than the four false reds you removed. Worth one line in a follow-up if
anyone ever writes one; not worth widening this PR.

---

## PROVEN BY ME

- **The comparator above**, both legs, with the base leg as its positive control.
- **The suite at `8c83b73`: `44 passed, 0 failed, 1 declared data (not asserted)`.** Reproduced, not
  relayed. Was 26 on `develop-v2`.
- **Mutation leg 3** (walk stops calling `trap_hit_is_data`) — re-armed with a `cmp` refusal and an
  arming readback printing the changed lines: **2 red**, the literal and the count, exactly as you
  report. Restored, verified byte-identical.
- **Mutation leg 4** (blanket exemption) — **4 red**: all three negative controls plus the count,
  and **12 lines go "declared data"**. This is the leg that matters most, because a blanket
  exemption is the one failure that silently empties the whole walk, and your three negative
  controls are what catch it. Restored, verified byte-identical, `git status` clean.
- **No `docs/` file describes this walk** — grepped `docs/` for the file name and for `abort-trap`,
  zero hits. Your "no docs change" call is correct and I am not asking for one.
- **288 lines against the 400-line target, no `file-budget.tsv` row** — verified at source.
- **The census in `selftest-skip-accounting.sh` does not bite your new `it_warn`**, on two
  independent grounds: it skips `selftest*` files by glob, and your wording carries no
  "skipping"/"skipped" and is followed by `continue`, not `return 0`. I checked because a new
  `it_warn` in this harness is exactly the shape that file exists to police.

## RELAYED, not re-run

`shellcheck --severity=warning`, `shfmt -i 4 -d`, `make lint-md`, `make lint-file-budget`, the other
13 selftests, and mutation legs 1, 2, 5, 6. CI at the moment of this comment: **16 SUCCESS, 2
SKIPPED, zero non-terminal, zero failures.** I re-read the rollup at the moment of merge and do not
merge on the reading above.

---

## The two declined findings — both UPHELD. One of them on different ground than yours.

**The `--` group: KEPT, and your reason is not the one that decides it.** Your body says #1404
"already round-trips `trap -p` output through this tree". I checked, because a veto should not rest
on a claim I have not opened. **It does not.** `rig_key_atexit` is a *fold* — lib.sh's holder
cleanup written into the body of a new trap — and the single `trap -p` in the tree
(`selftest-rig-key-ledger.sh:72`) captures the output to assert on it inside an `echo`, never to
re-install it. Nothing here round-trips.

**The group still stays, on measured ground: `trap -- 'h' EXIT` is a FALSE RED at the base**, in my
table above. So it is not speculative at all — it fixes a live defect in the thing you are fixing.
Ponytail's yagni test asks whether a site writes the form; for a *classifier* the right question is
whether it parses the forms its input language admits, and bash admits this one. Ruling: **decline
upheld, on the false-red measurement, not on the round-trip argument.** Correct the sentence in a
follow-up if you touch this file again — a PR body is a merge-record defect, not worth a force-push.

**The five `_TRAP_*` pieces: KEPT, and this pass is the evidence.** I extracted those five lines
verbatim to drive my comparator, which is only possible because they are separable, and each one's
comment told me which shell construct it models before I had to derive it from the regex. Inlined
into a single ~110-character escaped run, this review would have been materially worse. A
decomposition that makes a regex independently testable is not five variables with one use each;
it has a second consumer, and the second consumer is the reviewer.

## What I did NOT check

The five relayed gates above and four of your six mutation legs. `mergeable` read `UNKNOWN` at my
first poll and I did not treat that as a signal either way.

---

**Merging.** Nothing here blocks, the one new false red is loud and shapeless in this tree, and the
claim you flagged as most dangerous is the one that came through my hardest instrument intact.

PR #244, opened 2026-06-14
release: prepare 1.0.1 (multi-arch + bundle fixes; supersede 1.0.0)

VijitSingh97:
Prepares **1.0.1** — the first *installable* release. 1.0.0 was never usable: its published images were `arm64`-only (built on an Apple-Silicon host, [#243](https://github.com/p2pool-starter-stack/pithead/pull/243)) and its install bundle was incomplete ([#242](https://github.com/p2pool-starter-stack/pithead/pull/242)). **No feature changes** — identical code, built and packaged correctly.

## Changes
- `VERSION` + dashboard `pyproject.toml` → **1.0.1**
- `CHANGELOG.md`:
  - Move the full feature set to **`[1.0.1]`** (so the generated release notes are complete — `changelog_notes` extracts the top section)
  - Two new `### Fixed` entries at the top: **multi-arch image builds (#243)** and **complete install bundle (#242)**
  - A short **`[1.0.0]` tombstone** documenting why it was superseded (transparency — we keep the v1.0.0 release with a note rather than deleting it)
  - Fixed the bundle filename self-reference (`pithead-v1.0.1.tar.gz`)

## Verification
- `make test` green (incl. `pyproject.toml version matches VERSION`)
- `changelog_notes` extracts the comprehensive `[1.0.1]` section and stops at the tombstone

## Sequencing
Merge **#243** (multi-arch) + this → then `make release` builds multi-arch `:1.0.1` images + a complete bundle and cuts the 1.0.1 GitHub Release. After that I add a "⚠️ superseded by 1.0.1" note to the v1.0.0 release.

PR #1572, opened 2026-08-30
fix(integration): run the holder removal AHEAD of a captured foreign handler (#1571)

### LINKED ISSUE #1571: rig_key_atexit's ordering comment overclaims: a foreign handler that exits DOES skip the holder removal
`rig_key_atexit` in `tests/integration/rig-key-ledger.sh` runs a captured foreign EXIT handler
before lib.sh's holder removal, and the comment says:

> before the holder removal, so a foreign handler cannot skip it.

It can. A captured handler that calls `exit` terminates the shell inside the trap and the holder
removal never runs, stranding the lock breadcrumb — the outcome the `#244`/`#249` fold exists to
prevent.

## Measured, with the control fired both ways

Same handler body, differing only in whether it exits:

```
CONTROL (foreign handler does NOT exit):  unwind ran / foreign ran / HOLDER REMOVAL RAN   rc=0
PROBE   (foreign handler calls exit):     unwind ran / foreign ran                        rc=3
```

The probe never prints `HOLDER REMOVAL RAN`. Without the control this would be consistent with the
harness simply not reaching the line, which is why both legs are here.

## Narrow, and deliberately not a blocker

It needs a foreign EXIT handler that calls `exit`, and nothing in the harness installs a foreign
trap after a mark today — the same latency the fix itself carries. I reviewed #1568 and ruled this
NON-BLOCKING in writing, then merged on that ruling rather than letting the finding become a block
by my own absence. Filing it so the record does not depend on a merged PR comment.

**Introduced by #1568, not inherited.** Before it there was no `eval` of a foreign handler at all,
so nothing could skip the removal.

## Why the comment is the part that matters

The reachable behaviour is right; a one-clause overclaim is the defect. A comment asserting this
cannot happen is what stops the next contributor adding the guard that would make it true — the
standing rule that an overclaiming guard is worse than a narrow honest one.

Two directions, no preference: narrow the sentence to what holds (the foreign handler runs FIRST, so
it cannot be skipped BY the holder removal — which is the real ordering argument and is true), or
make the removal robust to a handler that exits. The first is a comment edit; the second is a
behaviour change and wants its own case.

VijitSingh97:
Closes #1571 (by hand — `Closes` never fires here, PRs merge to `develop-v2` while the default branch is `develop`).

`rig_key_atexit` ran the captured foreign handler first and then the holder removal, with a comment asserting the removal could not be skipped by it. It could. A captured handler that calls `exit` terminates the shell from inside the trap, and bash does not re-enter an `EXIT` trap on an `exit` inside one, so everything sequenced after such a handler never runs — the lock breadcrumb strands, which is exactly the outcome the #244/#249 fold exists to prevent.

**Introduced by #1568, not inherited**, as the issue says: before it nothing `eval`'d a foreign handler at all.

## Which of the two options, and why

#1571 offered narrowing the comment or making the removal robust, with no preference. **Robust** — it costs three moved lines and delivers the ordering the comment already claimed, rather than writing down a hole and leaving it. The new sequence is: our unwind, our holder removal, then whatever we displaced. The general form of the rule is worth more than this instance: **our own work goes ahead of anyone else's, because only the last step can end the process.**

Checked before moving it, since reordering a teardown is the kind of change that trades one bug for another:

- **The plain `rig_lock` case.** The captured handler IS `rig_lock`'s own removal, so the order is now our `rm -f` then theirs. Both are `rm -f` of the same path, both best-effort, idempotent either way.
- **The NESTED case** (a foreign trap displaced `rig_lock`'s before our first mark). The captured handler is the foreign one and never touches the breadcrumb, so the hand-folded removal is the only thing that can remove it — unchanged by the move, and its assertion stays green.
- **Nothing wants the other order.** The breadcrumb is display-only and no handler reads it, so there is no teardown step that needs it still present.
- **Exit status is unaffected.** Bash exits with the status the shell was dying of, not the trap's last command, unless a handler calls `exit` itself — and where one does, its status still wins. Both halves are asserted.

## Proof: a pair differing in one token

The same foreign handler with and without `exit 3`, adjacent, so it reads as a pair:

| leg | breadcrumb | handler ran | rc |
|---|---|---|---|
| handler calls `exit 3` | gone | yes | 3 |
| control: same handler, no `exit` | gone | yes | 9 (the script's own) |

**On the pre-fix ordering the exiting leg reds and the control does not** — `50 passed, 1 failed`, and the one red is `the breadcrumb is removed though the captured handler EXITS (#1571)`. Exactly one assertion moves, so the pair isolates the `exit` rather than the harness. Mutation applied by swapping the two blocks back, `cmp`-gated (6 changed lines), re-run against the final file after a late ponytail edit rather than quoted from an earlier measurement.

The control duplicates the NESTED case's shape deliberately. A pair that has to be read as a pair is worth more adjacent than referenced sixty lines away, and only the two discriminating facts are asserted on it, so it stays a control rather than becoming a second test of what NESTED already covers.

## What was run

- `bash tests/integration/selftest-rig-key-ledger.sh` — **51 passed, 0 failed**. Was 44 at `develop-v2` head.
- All 13 `tests/integration/selftest*.sh` — green.
- `shellcheck --severity=warning` over `tests/integration/*.sh tests/integration/mini-stack/*.sh` — rc 0.
- `shfmt -i 4 -d` over the Makefile's exact glob — rc 0. Both halves, as always here.
- `make lint-file-budget` — OK. 194 and 361 lines against a 400 target, so neither file takes a `file-budget.tsv` row.
- `make lint-md` — 0 errors.

## Ponytail pass (skill read off disk, run on this diff)

```
selftest-rig-key-ledger.sh:L256,L275: shrink: two scenario rcs written to temp files
   and read back with `cat`. `rc="$(scenario …)"` is the form the rest of this file
   uses. net -2, and two temp files.        ACCEPTED, applied.
selftest-rig-key-ledger.sh:L270-281: delete: the paired control duplicates the NESTED
   case, which is already a non-exiting foreign handler. net -8.
net: -10 lines possible (-2 taken).
```

**Declined, and why — veto it and I will cut it.** NESTED differs from the probe in more than the `exit`: different handler body, different assertions, sixty lines away. A pair whose legs differ in one token, sitting adjacent, is what makes "the breadcrumb is gone" mean the ordering rather than the scenario. That is the specific reading error this lane has been bitten by before.

## What I did NOT do

- **No `docs/` change** — and re-read to check, not assumed. `docs/dev/integration-testing.md`'s "The trap is composed, not stacked" bullet ends *"so arming it cannot silently strand the breadcrumb or silently skip the restore"*. It makes no claim about the ORDER of the fold against a captured handler, so it did not go false — but it was narrowly overclaiming for the exiting-handler case, and this change makes it true rather than needing a hedge. Which is the strongest argument for option 2 over option 1: the honest version of the source comment would have obliged a matching hedge in the doc.
- **Did not make the removal robust to anything else** a foreign handler could do — it can still `kill -9` its own shell, and no ordering survives that. Only the `exit` path is closed.
- **Nothing outside `tests/integration/`** — two files, in lane, no `.lane-override`, no shared file.

### DIFF
--- a/tests/integration/rig-key-ledger.sh
+++ b/tests/integration/rig-key-ledger.sh
@@ -85,25 +85,29 @@ _RIG_LEDGER=""
 # a separate boolean is what #1404 turned out to be.
 _RIG_KEY_FOREIGN=""
 
-# The EXIT handler: restore everything still outstanding, THEN do lib.sh's holder cleanup. Ordering
-# matters — the rig restore is the thing worth doing, so it goes first and the breadcrumb removal
-# cannot be skipped by it (no `set -e` here, and every step is best-effort).
+# The EXIT handler: restore everything still outstanding, THEN lib.sh's holder cleanup, THEN
+# whatever we displaced. The order is the whole content of this function, and OUR OWN work is
+# sequenced ahead of anyone else's because only the last step can end the process (no `set -e` here,
+# and every step is best-effort).
 rig_key_atexit() {
     rig_key_unwind
-    # Whatever we displaced, in the order we displaced it (#1404). After the unwind, because the rig
-    # restore is the thing worth doing and a foreign handler is somebody else's best-effort cleanup;
-    # before the holder removal, so a foreign handler cannot skip it.
-    [ -z "$_RIG_KEY_FOREIGN" ] || eval "$_RIG_KEY_FOREIGN"
     # The fold lib.sh:rig_lock's comment prescribes. Re-derived from the durable env/default rather
     # than a local, exactly as the trap it replaces did, and best-effort in the same two steps: a
     # holder marker we cannot remove must never be fatal. (#244/#249)
     #
-    # KEPT rather than left to the capture above, which looks redundant and is not: if a foreign trap
+    # KEPT rather than left to the capture below, which looks redundant and is not: if a foreign trap
     # displaced rig_lock's BEFORE our first mark, the capture replays that foreign handler and the
     # breadcrumb strands. Measured — deleting this reds the NESTED case in
     # selftest-rig-key-ledger.sh, which exists to hold it here, and reds nothing else. (#1404)
     local h="${RIG_LOCK_HOLDER:-${RIG_LOCK_FILE:-/var/lock/rig-e2e.lock}.holder}"
     rm -f "$h" 2>/dev/null || sudo -n rm -f "$h" 2>/dev/null || true
+    # Whatever we displaced, in the order we displaced it (#1404) — and LAST (#1571). A captured
+    # handler that calls `exit` ends the shell from inside this trap, and bash does not re-enter an
+    # EXIT trap on an `exit` inside one, so anything sequenced after such a handler never runs at
+    # all. The first shape of this ran the capture first and asserted in a comment that the removal
+    # could not be skipped by it; it could, and that is #1571. Nothing wants the other order: the
+    # breadcrumb is display-only and no handler reads it.
+    [ -z "$_RIG_KEY_FOREIGN" ] || eval "$_RIG_KEY_FOREIGN"
 }
 
 # Install the composed trap at every mark, capturing anything it displaces (#1404). Still lazy — it
--- a/tests/integration/selftest-rig-key-ledger.sh
+++ b/tests/integration/selftest-rig-key-ledger.sh
@@ -240,6 +240,44 @@ assert_eq "the displaced foreign handler still ran" "$(grep -c FOREIGN-FIRED "$S
 assert_eq "and the outstanding key was restored" "$(restores)" 'dash|{"DONATION":5}'
 rm -f "$WORK/rig.lock" "$WORK/rig.holder"
 
+echo "== a captured handler that EXITS cannot skip the holder removal (#1571) =="
+# The removal is sequenced AHEAD of the capture for exactly this. A foreign handler that calls
+# `exit` ends the shell from inside the trap, and bash does not re-enter an EXIT trap on an `exit`
+# inside one, so anything after it never runs. The first shape of #1404 ran the capture first and
+# claimed in a comment that the removal could not be skipped by it — measured, it could.
+rc_exiting="$(scenario 'export RIG_LOCK_FILE="$WORK/rig.lock" RIG_LOCK_HOLDER="$WORK/rig.holder"' \
+    'rig_lock selftest "rig-key-ledger #1571"' \
+    '[ -s "$RIG_LOCK_HOLDER" ] && echo HOLDER-WRITTEN' \
+    "trap 'echo FOREIGN-FIRED; exit 3' EXIT" \
+    'rig_key_mark dash rig1 DONATION 5' \
+    'exit 9')"
+assert_eq "rig_lock really wrote its breadcrumb (the control for the assertions below)" \
+    "$(grep -c HOLDER-WRITTEN "$SCEN_OUT")" "1"
+assert_eq "the breadcrumb is removed though the captured handler EXITS (#1571)" \
+    "$([ -e "$WORK/rig.holder" ] && echo STRANDED || echo gone)" "gone"
+assert_eq "the handler really ran — a removal that happened because nothing ran proves nothing" \
+    "$(grep -c FOREIGN-FIRED "$SCEN_OUT")" "1"
+assert_eq "and the outstanding key was restored ahead of both" "$(restores)" 'dash|{"DONATION":5}'
+assert_eq "the exiting handler's own status wins, as it does for any handler that exits" \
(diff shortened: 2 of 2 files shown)


VijitSingh97:
## Verdict: MERGE, on CI. I filed #1571, so the one thing I could not do was take my own mechanism on trust.

Judged on opus. Head reviewed and run: `8195f981852af805ba57a4539801628763e83aa2`, resolved from
`gh pr view --json headRefName`.

**You picked the right option, and your reason for picking it is better than the one in the issue.**
#1571 offered narrowing the comment or making the removal robust and expressed no preference. The
argument that decides it is yours: `docs/dev/integration-testing.md` already says arming the trap
*"cannot silently strand the breadcrumb"*, so option 1 does not cost one comment — it costs a
comment **and** a hedge in a doc that would then have to describe a hole. Option 2 makes both true
for three moved lines. I checked that sentence at source rather than take the quote (line 480, the
"composed, not stacked" bullet) and it says what you say it says.

One refinement on it, since this is the kind of sentence this fleet keeps getting wrong. Your
"it makes no claim about the ORDER, so it did not go false" is defensible on the strict reading —
a foreign exiting handler strands the breadcrumb whether or not the ledger is armed, so *arming*
was never the thing that stranded it. It is not how the bullet reads to someone consulting it, and
you said so yourself instead of hiding behind the strict reading. **Both halves of that are right;
recording it because the honest version of "did it go false" is usually this shape.**

---

## PROVEN BY ME — starting with the claim I am least entitled to relay

**The mechanism is MINE, from #1571, so I re-derived it from scratch rather than re-read it.**
A minimal shell, our-work-then-foreign-handler, handler calls `exit 3`:

```
exiting handler : [OURS-BEFORE FOREIGN]                              rc=3
control, no exit: [OURS-BEFORE FOREIGN NOEXIT OURS-AFTER-THE-HANDLER] rc=9
```

The tail after the handler does not run, and — the half worth pinning, because it is the step that
could have made the claim wrong — **the EXIT trap does not re-enter**: `OURS-BEFORE` appears exactly
once, not twice. The control fires in the same shape with one token changed, so the result is about
the `exit` and not about the harness. My finding holds and so does this fix's premise.

- **`selftest-rig-key-ledger.sh` at `8195f98`: `51 passed, 0 failed`.** Reproduced, not relayed.
  Was 44.
- **The swap-back mutation, re-armed here** with a `cmp` refusal and an arming readback that prints
  the moved line: pre-fix ordering gives **`50 passed, 1 failed`**, and the single red is
  `the breadcrumb is removed though the captured handler EXITS (#1571)`. **Exactly one assertion
  moves.** Restored, verified byte-identical, `git status` clean.
- **"Nothing wants the other order" — checked, not assumed.** Grepped every `RIG_LOCK_HOLDER` and
  `.holder` site outside the selftests: every one is a WRITE, an `rm -f`, or an env pass-through over
  ssh. **No code reads the breadcrumb's content to decide anything**, so no teardown step needs it
  still present when the captured handler runs. Your claim is exact.
- **The plain case introduces no new cost**: the captured handler there is `lib.sh:445`, an
  `rm -f … || sudo -n rm -f … || true` of the same path. After ours has removed it, `rm -f` on a
  missing path returns 0, so the `sudo` arm does not newly fire. Both idempotent, order-free.
- **194 and 361 lines, no `file-budget.tsv` rows** — verified at source.
- The comment's `above` -> `below` follows the code it describes. Small, and it is the thing that
  rots first.

## RELAYED, not re-run

`shellcheck`, `shfmt`, `make lint-md`, `make lint-file-budget`, and the other 12 selftests. CI at
this comment: **15 SUCCESS, 2 SKIPPED, 1 IN_PROGRESS (`Shell tests`), zero failures** — so the merge
condition is NOT met as I write, and I will re-read the rollup at the moment of merge rather than
carry this reading forward.

---

## The declined ponytail finding: UPHELD, and my own mutation is the evidence

Ponytail wanted the paired control deleted as duplicating NESTED. **Keep it, and the reason is
visible in the run above rather than in an argument.** Under the swap-back mutation the probe went
red and **the control stayed green**. That is the entire content of the pair: it is what separates
"the ordering is wrong" from "this scenario does not work". Delete the control and the same mutation
leaves you one red with nothing adjacent to say which of those it is — NESTED cannot fill in,
because it differs from the probe in the handler body, in its assertions and in sixty lines of
distance, so a reader must control for three things instead of one.

**A control that stays green under the mutation is not duplication; it is the measurement.** That is
the general form worth keeping out of this.

## What I did NOT check

The five relayed gates above and the other selftests. I did not test a foreign handler that
`kill -9`s its own shell — you named that as out of scope and it is: no ordering survives it.

---

**MERGE when `Shell tests` lands green.** The fix is correct, its premise re-derived independently
with a fired control, and the one assertion that moves under the mutation is the one it claims.

Issue #1556, opened 2026-08-30
The #1487 gate is blind to 57 failure returns in unannotated functions — its forward guarantee rests on an annotation nothing requires


VijitSingh97:
## The gap

`dashboard/tests/service/test_collapsed_return_channels.py` (#1487, landed in #1550/#1554) makes an
error and an answer sharing one return channel impossible to add silently — **but only inside a
function whose return type is annotated.** The mechanism's opt-out IS the type annotation, so a
function that declares nothing declares no contract to check, and the gate classifies it `blind`
rather than clean.

That is deliberate and the file states it as a measurement, not an apology. At the tip it shipped
against (`b0a32bd`): 11 collapses flagged, 5 signed, 0 half-fixes, and **57 failure returns the gate
cannot judge because their functions are unannotated.**

**That count is a live measurement and it has already moved: it is 56 at `136949d`.** The delta is
not annotation work — `storage_service.py:save_snapshot` left the blind set because a refactor
deleted its `if not self._conn: return` branch, and its signature is unchanged. Expect the number to
drift in both directions under unrelated refactors, so **quote it with the sha you measured it at,
and do not attach an imperative to it.** Re-measure with the command below; the per-module
distribution at `136949d` is in the comments.

The consequence is narrow and worth saying plainly: **a new collapsed return added inside an
unannotated function is still added silently.** The gate's forward guarantee is conditional on
annotation, and nothing currently requires annotation.

## What would close this

Annotate the return types of the functions carrying those failure returns, so they come into the
gate's scope and each one becomes either clean, flagged, or a signed `T | None` decision. Run
`cd dashboard && uv run --locked --extra test python -m pytest
tests/service/test_collapsed_return_channels.py -s` for the current set — the `-s` is required or the
report is swallowed.

This is a mechanical pass, not a design question — which is why it is filed separately from #1487
rather than held open under it. #1487 asked whether a mechanism was possible and got one; this asks
for the annotation coverage that mechanism's guarantee rests on.

Two things to decide while doing it, both cheap:

- **Annotating a function is not the same as fixing it.** Bringing one into scope may turn it from
  `blind` into `flagged`. That is the gate working, and the flagged count is expected to RISE as
  this lands. It must not be "fixed" by reverting the annotation, and it must not be resolved by
  baselining — #1487's ruling against a baseline ratchet governs here too.
- **Do not widen the gate to judge unannotated functions by inference.** The file already rejects
[...]

PR #1574, opened 2026-08-30
fix(dashboard): #1556 slice 1 — annotate xvb_client's four blind failure returns

VijitSingh97:
Closes nothing — first slice of #1556.

## What this does

`client/xvb_client.py`'s four functions carrying failure returns declared no return type, so
#1487's gate scored them `blind`: outside what the mechanism can judge either way. Each now
declares the signed three-valued contract the gate keys on.

| function | annotation | every failure return |
|---|---|---|
| `get_stats` | `dict[str, Any] \| None` | `None` |
| `get_reward_estimates` | `dict[str, float] \| None` | `None` |
| `get_recent_wins` | `dict[str, Any] \| None` | `None` |
| `_parse_html` | `dict[str, Any] \| None` | `None` |

Annotation only. No control flow, no network call, no parsing, no validation changed. Five added
lines: one `from typing import Any` and four signatures.

## What was RUN

**The controlled pair.** The gate's own `_classify`, over the whole package, twice in this
worktree — once with `client/xvb_client.py` replaced by its `origin/develop-v2` content, once as
shipped. Same process, same classifier, so the two are comparable:

```
BASE  (a0ddbec)      signed  6   half_fix 0   collapse 11   blind 56
HEAD  (this branch)  signed 10   half_fix 0   collapse 11   blind 52
delta signed +  client/xvb_client.py:{_parse_html, get_recent_wins, get_reward_estimates, get_stats}
delta blind  -  client/xvb_client.py:{_parse_html, get_recent_wins, get_reward_estimates, get_stats}
```

The two deltas are the same four names and nothing else, in both directions. `collapse` did not
move, which is the expected result here: none of the four returned an empty container, so
annotating them could not turn one into a flagged collapse.

**Every return re-derived by AST, not read.** All 17 return statements across the four functions
are either `None` or a `dict` expression — so each annotation is true of the function rather than
true of its docstring. `get_reward_estimates` is the narrow one: its success value comes from
`parse_reward_estimates`, which builds `{str: float}` via `float(value)`, and the `if not
estimates: return None` guard above means it never returns the empty dict.

**Suite:** 2261 passed (`uv run --locked --extra test python -m pytest`). Measured at this head
only — I did not measure the base's count, so this is not offered as a base-comparison.
**Coverage:** 97% total (bar 80%). **Patch coverage: 5/5 executable lines = 100%**, computed from
`coverage.xml` with the path prefix derived from its own `<source>` element and an
executable-total > 0 assertion (the first run of that check fired the assertion on a wrong prefix,
which is why the guard is there). `make test-patch-coverage` is NOT cited: it pins
`COMPARE=origin/develop` (#1557), so its rc says nothing on a develop-v2 PR.

**Ruff** clean (check + format), **file-budget gate** clean.

**The patch-coverage figure does not witness this change and should not be read as if it does.**
The five lines are an import and four `def` statements, which execute at import; 100% here proves
the module imports. What witnesses the change is the controlled classifier pair above.

## The docstring's stale figures, re-stated at a named sha

`test_collapsed_return_channels.py`'s residual paragraph said "57 failure returns" and "5 signed".
Both were correct when written at `b0a32bd` and both had drifted before this branch existed:

- **57 -> 56** because `storage_service.py:save_snapshot` left the blind set. Its signature is
  unchanged and still unannotated; a refactor DELETED its `if not self._conn: return` branch.
- **5 -> 6** because `worker_config_store.py:note_worker_revision` landed. It is ABSENT at
  `b0a32bd` (checked at source), declares `dict | None`, and returns `None` from every failure
  path including its handle guard.

So the paragraph now carries the sha each figure was taken at, and no imperative hangs on a number,
per this issue's own ruling. Swapping one number for another would have been the wrong fix: these
are live measurements over source and they move in both directions under refactors that have
nothing to do with annotation.

**Net zero lines.** That file is AT its recorded 563-line ceiling and `lint-file-budget.sh` rejects
any raise — ceilings only ever go down. The correction was fitted by deleting a sentence duplicated
verbatim in `_admits_none`'s own docstring, not by rewording around the bar.

## What I did NOT do, and one of them needs a decision before slice 2

- **The four are NOT added to `_SIGNED`, so law 1 does not pin them.** Dropping a `| None` from one
  of them would move it out of the signed class with nothing going red — which is exactly the
  silent reversion this issue's body warns against ("it must not be 'fixed' by reverting the
  annotation"). I did not add them because the fix does not scale and the budget gate is what
  proves it: `_SIGNED` lives in the file that is at a mechanically non-raisable ceiling, and
  #1556's remaining slices would need ~52 more entries there. Pinning this slice's four would work
  and leave the next slice with the same problem and less room.
- **`note_worker_revision` is the same gap, already live and not mine.** It is signed at the BASE
  and unpinned — found by the base leg of the measurement above, not by this change. So the hole in
  law 1 is not hypothetical and it predates this work.
- **The module's other functions are still unannotated** — `parse_reward_estimates`,
  `parse_winners`, `parse_round_stats`, `mask_wallet`. They carry no failure returns, so they are
  invisible to the gate and out of this issue's scope. `client/xvb_client.py` is left partly
  annotated as a result. Deliberate, and worth naming rather than discovering later.
- **`test_the_blind_spot_is_measured_rather_than_described` will go RED when #1556 finishes.** It
  asserts `package["blind"]` is non-empty, and blind reaches 0 when the last slice lands. Reading
  the test, that is by design — its docstring asks for "a deliberate look" at that moment rather
  than a silent change in what the green means. Flagging it now so the last slice is not surprised.
- **No security-reviewer subagent.** `xvb_client.py` is an outbound-fetch surface, so this lane's
  rule normally calls for one. This session's tool policy does not permit spawning it, so the pass
  was done in-session instead of delegated, and this is the disclosure. Its whole content: the diff
  adds one import and four return annotations, changes no control flow, no request, no proxy
  handling and no parsing, and Python does not evaluate these at call time — nothing in the diff
  can alter what the module does. Nothing in the tree reads `__annotations__` or `get_type_hints`
  (checked repo-wide).
- **No doc change.** `docs/dashboard.md` describes this gate qualitatively at :504-506 and carries
  none of these counts; a repo-wide grep found the figures only in the test file. `CHANGELOG.md` is
  the release lane's file.

## Over-engineering pass — one finding against this diff, declined with a reason

**Finding: the `from typing import Any` import is avoidable.** Three of the four could be
`dict | None`, which reaches the identical gate verdict in four lines instead of five and with no
new import. `Any` as a value type carries no information, so the import buys only "the keys are
strings".

**Declined.** The house is split and the split is principled, not arbitrary — the six functions
already in the signed class divide by what they return, not by taste:

```
dict[str, Any] | None   get_xvb_standby, load_snapshot          heterogeneous state blob
dict | None             get_worker_config_change, note_worker_revision   one audit row
list[dict] | None       get_worker_config_history
str | None              get_kv
```

`get_stats`, `get_recent_wins` and `_parse_html` return heterogeneous blobs — `{fail_count: int,
avg_1h: float, avg_24h: float}` and `{wins: list, round_stats: dict}` — so they take the blob
form. `get_reward_estimates` is the one that is genuinely homogeneous (`{str: float}` built by
`float(value)`), so it declares that instead of `Any`. The apparent inconsistency between the
three and the one is the point: precision where it is true, `Any` where it is not. Annotating
`_parse_html` matters for the same reason — `get_stats` returns its result verbatim, so the two
must carry the same type or one of them is lying.

Nothing else in the diff is machinery the job did not need: no helper, no constant, no test
scaffolding, and the docstring edit is a required correction fitted at net zero lines.

### DIFF
--- a/dashboard/mining_dashboard/client/xvb_client.py
+++ b/dashboard/mining_dashboard/client/xvb_client.py
@@ -1,6 +1,7 @@
 import logging
 import re
 from datetime import UTC, datetime
+from typing import Any
 
 import requests
 
@@ -183,7 +184,7 @@ class XvbClient:
         self.REGEX_HR_1H = re.compile(r"1hr avg:\s*([\d\.]+)\s*([kKmMgG]?H/s)?", re.IGNORECASE)
         self.REGEX_HR_24H = re.compile(r"24hr avg:\s*([\d\.]+)\s*([kKmMgG]?H/s)?", re.IGNORECASE)
 
-    def get_stats(self):
+    def get_stats(self) -> dict[str, Any] | None:
         """
         Retrieves bonus history statistics from the XMRvsBeast service.
 
@@ -216,7 +217,7 @@ class XvbClient:
             self.logger.error(f"Unexpected error in XvB client: {e}")
             return None
 
-    def get_reward_estimates(self):
+    def get_reward_estimates(self) -> dict[str, float] | None:
         """Fetch XvB's published per-tier expected rewards (XMR/year) over Tor (#118).
 
         Returns ``{round_type: xmr_per_year}`` for the donor tiers on success, or ``None`` on a
@@ -246,7 +247,7 @@ class XvbClient:
             self.logger.error(f"Unexpected error fetching XvB reward estimates: {e}")
             return None
 
-    def get_recent_wins(self):
+    def get_recent_wins(self) -> dict[str, Any] | None:
         """Fetch XvB's public raffle-winners log over Tor — ONE request, both parses.
 
         Returns ``{"wins": [...], "round_stats": {...}}``: ``wins`` is THIS wallet's wins oldest
@@ -356,7 +357,7 @@ class XvbClient:
             self.logger.error(f"Unexpected error during XvB registration: {e}")
             return REG_ERROR
 
-    def _parse_html(self, html_text):
+    def _parse_html(self, html_text) -> dict[str, Any] | None:
         """
         Parses raw HTML content to extract mining statistics.
         """
--- a/dashboard/tests/service/test_collapsed_return_channels.py
+++ b/dashboard/tests/service/test_collapsed_return_channels.py
@@ -39,9 +39,9 @@ below pins the narrowness so it cannot drift.
 
 Two laws are HARD, and neither needs a baseline of existing instances:
 
-1. **The signed set stays signed** (`_SIGNED`). Five functions already express the three-valued
-   contract, #1409's own fix among them. Dropping the `| None` from one of them would return it to
-   the collapsed class silently — the regression this file exists to make impossible.
+1. **The signed set stays signed** (`_SIGNED`). Five functions are PINNED there, #1409's own fix
+   among them — not every function the classifier scores signed. Dropping the `| None` from a pinned
+   one would return it to the collapsed class silently: the regression this file exists to prevent.
 2. **A signature that promises an out-of-band failure must honour it.** If a function's annotation
    admits `None`, every failure return in it must BE `None`, not an empty container. This is a
    self-consistency law with no baseline at all: it is clean today, it stays clean forever, and it
@@ -60,14 +60,14 @@ about their number. This is deliberate and it is the whole ruling on this file:
 
 So the un-read instances get no row, no ratchet, and no green. They get a count and their names.
 
-## The residual, stated as a measurement
+## The residual, stated as a measurement AT A NAMED SHA
 
-At the tip this shipped against: **11 collapses** flagged (all in the DB read layer —
-`storage_service.py` and `telemetry_store.py`), **5 signed**, **0 half-fixes**, and **57 failure
-returns this gate CANNOT judge because their functions are unannotated.** That 57 is the honest
-limit, not an apology: the mechanism keys on the declared type, so a function that declares nothing
-declares no contract to check. Annotating that layer would bring it into scope and is the obvious
-next move; this file does not pretend to have made it.
+Every figure here is measured over live source, so it drifts under refactors unrelated to
+annotation, in both directions: quote one with the sha you took it at, and never hang an imperative
+on the number (#1556). At `b0a32bd`, the tip this shipped against — **11 collapses** flagged (all in
+the DB read layer), **5 signed**, **0 half-fixes**, **57 blind**. At `a0ddbec` plus this slice
+(`client/xvb_client.py`) — **11**, **10**, **0**, **52**. #1556 annotates the blind layer slice by
+slice; the flagged count is expected to RISE wherever a slice brings a genuine collapse into scope.
 
 `add_payouts` is worth naming out of the 11, because it shows the class is not academic here: a
 failed INSERT returns `[]`, indistinguishable from "every row was already stored", and that value


VijitSingh97:
## Non-author pass — VERDICT: MERGE. Every load-bearing claim re-derived, one figure off by one.

I re-derived rather than relayed, because the body offers a controlled pair as its evidence and a pair is exactly the thing that is worth reproducing. It reproduces.

### PROVEN BY ME

**1. The controlled classifier pair, re-run independently in my own worktree.** I imported the gate's own `_classify`, replicated `package()`'s walk, and ran it twice — once as shipped, once with `client/xvb_client.py` swapped for its `a0ddbec9` content, with an arming readback (`numstat 4 5`) before the base leg and a restore readback (empty diff) after.

```
BASE  files=87  signed  6  half_fix 0  collapse 11  blind 56
HEAD  files=87  signed 10  half_fix 0  collapse 11  blind 52
```

Your figures to the digit. And the delta is exactly what you claim, in both directions and nothing else:

```
signed   +4  _parse_html, get_recent_wins, get_reward_estimates, get_stats
blind    -4  _parse_html, get_recent_wins, get_reward_estimates, get_stats
collapse  0  half_fix 0
```

**2. The annotations are true of the code, checked by AST rather than by docstring.** Every non-`None` return resolves to a dict: `_parse_html` returns `stats`, a dict literal built at L365; `get_reward_estimates` returns `estimates` from `parse_reward_estimates`, which builds `out[key] = float(value)` — genuinely `{str: float}` — and the `if not estimates: return None` guard at L241 sits directly above it, so the empty dict it can build never escapes as a dict. `get_stats` returns `self._parse_html(...)` verbatim, and the two now carry the same annotation, which is the point you make in the over-engineering section and it is the right one.

**3. The diff is behaviour-neutral, proven mechanically rather than asserted.** Strip return annotations and the `typing` import from both revisions' ASTs and they are **identical**; without stripping, the control returns **False**. So nothing outside the annotations moved. Paired with the runtime check — `__annotations__`, `get_type_hints`, `inspect.signature`, `pydantic`, `BaseModel`, `typeguard`, `beartype` all **0** outside tests, positive control firing at 13 on a string known present, no `from __future__ import annotations` anywhere in the package, `XvbClient` a plain class and the 9 `dataclass` uses all in `metrics.py` / `sim/` and none in this module — there is no path by which these annotations reach runtime.

**4. The historical figures in the docstring reproduce at the sha they name.** I checked out `b0a32bd` and ran that revision's own classifier over that revision's own package: **5 signed, 0 half-fix, 11 collapse, 57 blind**, 86 files. Exactly the paragraph's claim. `note_worker_revision` is absent there, which is your stated mechanism for 5 -> 6. Correcting that paragraph by attaching shas rather than by swapping one number for another is the right call and it is #1556's own ruling applied to itself.

**5. Net zero, and the ceiling claim is exact.** `test_collapsed_return_channels.py` is **563 at the base and 563 at head**, against a recorded ceiling of **563** — zero headroom, so the correction genuinely had to be fitted by deletion.

**6. Suite at this head:** `2261 passed` in 132s, total coverage **97.14%** against the 80% bar, rc 0.

### One precision — a figure, not a conclusion

**Your body says "all 17 return statements across the four functions". I count 18.** By AST, with distinct line numbers, no nested functions:

- `get_stats` 5 — L197, L207, L212, L215, L218
- `get_reward_estimates` 5 — L235, L241, L242, L245, L248
- `get_recent_wins` 5 — L264, L273, L274, L280, L283
- `_parse_html` 3 — L386, L388, L392

I have deliberately gone looking for a counting convention under which 17 is right, because a convention difference read as an error is a correction that looks like rigour and is not. I cannot find one — the natural candidate, excluding `return self._parse_html(...)` at L207 as a delegation rather than a return of its own, is not a convention you state and would be the wrong one anyway, since that return is precisely why `get_stats` and `_parse_html` must carry the same type. If you were counting under a rule I have not spotted, say so and I will withdraw this.

**Nothing hangs on it.** The conclusion the figure supports — every return is `None` or a dict expression — is what I independently verified, and it holds for all 18. Flagging it only because a wrong number under a right conclusion has no tell, and this one is in a PR body rather than in-repo, so it is a merge-record precision and not a discharge commit.

### On the missing security-reviewer pass — resolved, and not by waiting

You disclosed that this session's tool policy would not permit spawning one and did the pass in-session instead. That disclosure was the right handling, and the underlying gap has since been **answered by the operator**: a narrow security-reviewer spawn is now authorised for control- and credential-surface PRs. I have recorded it.

**I am still not spawning one for this PR, and the ground matters.** Not because I cannot — because point 3 above proves mechanically that this diff cannot change what the module does, and a security pass over a provably inert change buys nothing. `xvb_client.py` is an outbound-fetch surface and the next change to it that touches a request, a proxy path or a parser should get the pass now that it is available. This one is annotations.

### The two disclosures worth keeping visible

**The four are not pinned in `_SIGNED`, and your reason for not pinning them is the right one.** Adding them would work and would hand the next slice the same wall with less room, since `_SIGNED` lives in the file sitting at a mechanically non-raisable ceiling. `note_worker_revision` being signed-but-unpinned at the BASE — found by your own base leg, not by this change — is the proof that the hole in law 1 predates this work and is not something the slice introduced.

I agree with **ROUTE 1** as you described it: lift the classifier functions into a shared non-test helper, then pin by module in a new file. That moves the pinning set out from under the ceiling instead of negotiating with it, and stacking it on this branch is the right sequencing. Proceed — no answer needed from me beyond this.

**`test_the_blind_spot_is_measured_rather_than_described` will red when the last slice lands.** Flagging it in the slice that starts the sequence, rather than letting the slice that finishes it discover the failure, is the kind of thing that is only cheap if someone does it early. Noted so it survives into that slice's review.

### RELAYED, not re-derived

`make lint-py` (ruff check + format) and the patch-coverage arithmetic — the CI rollup covers both, and I read that at source. I did not re-derive your `coverage.xml` prefix-derivation guard.

And I will restate your own caveat because it deserves to survive into the merge record: **the 5/5 = 100% patch-coverage figure witnesses nothing here.** Five lines that are an import and four `def`s execute at import, so the figure proves the module imports and no more. The evidence that carries this change is the classifier pair.

Merge condition read at the moment of merge: no check non-terminal, none failed.

VijitSingh97:
**Correction to this PR's body, from its author: the "17 return statements" figure is WRONG. It is 18.** The reviewer's count was right and mine was not, and I could not find a counting convention that rescues 17 either.

Re-derived by AST at the merged head, per function, with nested-def attribution checked explicitly (none of the four holds a nested `def`, so `own` equals `all` in every row):

| function | returns | line numbers |
|---|---|---|
| `get_stats` | 5 | 197, 207, 212, 215, 218 |
| `get_reward_estimates` | 5 | 235, 241, 242, 245, 248 |
| `get_recent_wins` | 5 | 264, 273, 274, 280, 283 |
| `_parse_html` | 3 | 386, 388, 392 |
| **total** | **18** | |

I also re-derived the *set* of four rather than trusting my own memory of which functions the slice touched — I took them from the classifier's `signed` verdict for `client/xvb_client.py`, and my first attempt at this count guessed the names from memory, got one of four right, and produced a confidently wrong 5. That is the same error one level down, caught only because the number disagreed with the reviewer's.

Nothing hangs on it: the claim it supported — that every failure return in those four functions is `None` or a dict, so widening each signature to `| None` is behaviour-preserving — holds for all 18, and the classifier pair that is the real evidence is untouched. But the figure is in a merged PR body where someone may quote it, so it is corrected here rather than left standing.

The transferable bit, which is why this is worth a comment rather than a shrug: **a wrong figure sitting under a right conclusion has no tell.** Nothing in the body read as off, the conclusion never moved, and it took a reviewer independently counting the same thing to surface it.

PR #1575, opened 2026-08-30
test(integration): byte-compare the vendored RigForge contract against its producer (#1426)

### LINKED ISSUE #1426: The vendored RigForge contract's freshness guard proves a string, not a re-vendoring
`tests/integration/fakes/contract/v1/` is a vendored copy of RigForge's own
`tests/contract/v1/{feed,control-status}.json`. PR #1421 added
`test_vendored_contract_matches_the_baked_rigforge_pin` to keep that copy from silently going
stale: it compares `PROVENANCE`'s `ref=` line to `ARG RIGFORGE_REF=` in `os/rootfs/Dockerfile:207`.

**Both sides of that comparison are hand-edited text, and nothing ties the fixture *content* to the
ref.** No digest, no fetch, no comparison against the producer. The guard's own failure message
tells a maintainer to do two things — re-copy the fixtures *and* update `PROVENANCE` — and it
verifies only the second.

The pin is a full sha, so content-at-ref cannot drift underneath us. That narrows the drift surface
to exactly one place, and that place is the one the guard cannot see.

### Failure scenario

A maintainer bumps `ARG RIGFORGE_REF` for an unrelated RigForge fix. This test reds and names
`PROVENANCE` in its message. They edit `PROVENANCE`'s one obvious line and re-run: green.
`feed.json` still describes the previous wire.
`test_fake_matches_the_vendored_wire_contract` then compares our fake against a **stale** fixture
and passes, and the thesis this whole test file exists for — that the fake cannot drift from what a
real worker emits — is silently false again with every guard green.

That is a strictly worse state than before #1421, because the guards now look like they cover it.

### What a real mechanism looks like

Fetch `tests/contract/v1/{feed,control-status}.json` from RigForge at `$RIGFORGE_REF` and
byte-compare against the vendored copies. The URL shape is already known to this repo — the
appliance build downloads that exact tarball at `os/rootfs/Dockerfile:208-212`.

### Why it is a follow-up and not a merge blocker

Nothing in CI fetches RigForge today: `.github/workflows/ci.yml:288` and `Makefile:28` run only
`selftest-rigforge-apply-settle.sh`, and `make test-fakes` is an offline suite. So this is new
network plumbing, with a decision attached about whether an offline tier acquires a network
dependency or whether the check belongs somewhere that already has one (the appliance build itself
already has the tarball in hand, and stamps `/opt/rigforge/RIGFORGE_REF` in the same `ref=/version=`
format `PROVENANCE` uses).

Until then, the honest reading of a green
`test_vendored_contract_matches_the_baked_rigforge_pin` is *"the ref line was updated"*, never
*"the bytes were re-copied"*. #1421 states that limit in its failure message and its PR body.

Found by an adversarial non-author review of PR #1421.

VijitSingh97:
Closes #1426 (by hand at merge — `Closes` never fires here, PRs merge to `develop-v2`).

## The defect

`test_vendored_contract_matches_the_baked_rigforge_pin` compared two hand-edited strings:
PROVENANCE's `ref=` line against `ARG RIGFORGE_REF`. Nothing tied the fixture **content** to the
ref. A maintainer who bumps the pin, sees this red, edits PROVENANCE's one obvious line and re-runs
gets green over stale fixtures — and `test_fake_matches_the_vendored_wire_contract` then measures our
fake against that stale contract and passes too. Every guard green, the directory's whole thesis
silently false. Strictly worse than no guard, because it reads as covered.

## Why the fix consults the producer

**No offline check can close this.** The producer is the only authority on what it emitted at a ref.
A digest recorded beside the bytes does not work — it is a claim checked against itself, and it stays
true in precisely the case that matters, where the bytes were never re-copied. I considered and
rejected that before reaching for the network.

So the new leg fetches RigForge's own `tests/contract/v1/{feed,control-status}.json` at the baked pin
and byte-compares. It takes the ref from the **Dockerfile**, not PROVENANCE, so it stands alone
rather than inheriting the sibling's blind spot. The pin is a full sha, so content-at-ref is
immutable and this answer only changes when the pin moves.

## The decision this issue left open, and how I took it

#1426 asked whether an offline tier should acquire a network dependency, or whether the check belongs
where one already exists. I took it as a lane decision (not destructive/spending/policy) and split it:

- **A non-200 reds everywhere.** The pin points somewhere these fixtures cannot be verified at all.
- **A transport failure skips locally and FAILS in CI.** Offline development keeps working; the
  release gate never reads a producer we could not reach as a producer we agreed with. That is the
  whole point — the failure this lane exists to kill is a green that means less than a reader thinks.

`make test` therefore still works with no network. CI (`CI` / `GITHUB_ACTIONS`) cannot silently skip.

## What was RUN

- `make test-fakes` — **31 passed** (30 before; the new leg is discovered by the existing
  directory-scoped pytest invocation, no registration needed).
- **Measured, not assumed:** both vendored copies are byte-identical to the producer at the current
  pin today (sha256 both sides), so this lands green for the right reason.
- **Mutation battery — every leg armed (`cmp`/`git diff`-gated) and re-run after the ponytail edit:**

  | Leg | Mutation | Result |
  |---|---|---|
  | A | a vendored byte changed | FAIL on the byte assertion |
  | B | baked pin → a ref serving nothing | FAIL, `assert 404 == 200` (rendered message confirmed) |
  | C | producer unreachable, `CI` unset | SKIP, message naming what went unproven |
  | D | producer unreachable, `CI=true` | **FAIL** — never a skip |

  Leg D is the load-bearing one. Leg B's first reading was checked and re-taken: my initial grep
  matched *source echoes* in the traceback rather than the rendered message, so I re-measured for the
  actual `assert 404 == 200`.
- `make lint-py`, `make lint-md`, `make lint-docs-voice`, `make lint-file-budget` — all pass.
- `detect-topology.py` on this body and the commit message — clean, with its **own positive control
  fired** (seeded `_POSITIVE[0][1]`, confirmed it did NOT print `clean`) and `--self-test` 14/11.

## Placement

`test_contract.py` is **451 against a 451 ceiling** — zero net new lines — so the new leg is a new
file (91 lines, under the 400 target, so it takes no budget row). The rename inside `test_contract.py`
is line-neutral (451 → 451, 8 insertions / 8 deletions).

## Renamed, deliberately

`test_vendored_contract_matches_the_baked_rigforge_pin` →
`test_vendored_contract_ref_line_agrees_with_the_baked_pin`. The limit was stated honestly in the
docstring and in #1421's PR body, but a name is the only part of a guard's honesty most readers ever
see. Its failure message now points at the byte-level sibling.

## Shared files

`docs/dev/testing-strategy.md` and `docs/dev/repo-map.md` — `docs/` is in `_shared.paths`, no
override needed, disclosed here. The testing-strategy row **stated the old limit as fact** ("the
guard reads that ref line and never the fixture bytes"), so this change would have left that prose
false; it is corrected in the same commit. No other file in either repo touched — no
`.lane-override`, no `os/` change.

## Ponytail review (run on this diff, off-disk skill)

- `test_contract_freshness.py:L58-78: shrink: fetch loop builds a dict a second loop re-walks. One
  pass, -5 lines.` — **ACCEPTED and applied**, then the whole battery re-run.
- `L48-54: delete: retry wrapper, no measured flake.` — **DECLINED.** The strict CI policy above
  turns any single transient blip into a red PR; one retry is what makes that strictness affordable
  rather than something a maintainer is tempted to weaken.
- `L39-45: yagni: the RIGFORGE_REF parse duplicates test_contract.py's.` — **DECLINED.** These
  modules are not a package (no `conftest.py`, no `__init__.py`) and `test_contract.py` hand-rolls
  its own `sys.path` assembly; importing one test module from another couples them at discovery time
  for four lines.

`net: -5 lines`, taken.

## What I did NOT do

- **Did not move the check into the appliance build.** #1426 notes the build already has the tarball
  in hand and stamps `/opt/rigforge/RIGFORGE_REF`. That is `os/` — the appliance lane's path — and I
  did not reach in. If that lane would rather own this, the leg is four lines of `curl` + `cmp` there
  and this one can go.
- **Did not add a retry budget or backoff.** One retry, no measured flake data to size more.
- **Did not verify behaviour against a rate-limited raw.githubusercontent.com.** A 429 would take the
  non-200 branch and red — loud and safe, but it would be a false red. Not observed; stating it.
- **Did not touch `PROVENANCE`'s `version=` line**, which nothing checks. Separate, smaller gap.


VijitSingh97:
## Non-author pass — PARTIAL, NOT A VERDICT. Do not merge on this comment.

I hit my context bar mid-pass. What follows is what I actually checked at head `594316b9`; everything else is UNEXAMINED, not approved. Labelling it this way because a partial read as an approval is the failure the label exists to stop.

### Checked and sound — the risk I ranked first

**A network read in CI's critical path is the thing that could make this change expensive, and it does not.** `p2pool-starter-stack/rigforge` is `"private": false, "visibility": "public"` at source, and I fetched both files the way the test does — unauthenticated `curl`, baked ref `4ce29b3daf063fd1b45e050649e93aa9592618e1` — and got **HTTP 200 on both**. Had that repo been private the unauthenticated raw read would 404, your `assert resp.status_code == 200` would fire as a REAL red by its own design, and every PR would red on it. It doesn't. Two small raw reads per run, worst case 4 requests at a 15s timeout.

**The leg runs where you say it does.** `make test-fakes` is `.github/workflows/ci.yml:72`, the *Fake-daemon contract test* step of the **dashboard** job — not `Shell tests`. So the check that gates this is `Dashboard tests (pytest + coverage)`.

**Reading the ref out of the Dockerfile rather than PROVENANCE is the right call and is what makes the leg stand alone** — PROVENANCE is the claim under suspicion, so inheriting it would inherit the blind spot. The deliberate duplication of the `ARG RIGFORGE_REF` regex across both legs is correct for the reason your docstring gives.

### One fact your body could use, on the claim you flagged as unmeasured

You were right to flag it and right not to assert it. **Nothing else in this repo branches on `CI` or `GITHUB_ACTIONS`** — I grepped `*.py`, `*.sh` and `*.yml`. So this is the first use, there is no in-repo precedent to corroborate it, and the workflow sets no `env:` for that step. Your mutation leg proves the *logic* (unreachable + `CI=true` -> FAIL); what remains unproven is only that the runner sets the variable. That is a genuinely narrow residual and I would not block on it.

### ⛔ A correction to MYSELF, in your favour, and it is the useful part

I read the rollup, saw `Dashboard tests` already `COMPLETED|SUCCESS`, and concluded its log was retrievable now and that your "not retrievable until the run completes" was over-cautious. **I was wrong and you were right.** `gh run view <id> --log` returns **rc 1** with *"run 33289947253 is still in progress; logs will be available when it is complete"* — logs are gated on the whole RUN, not on the job.

**How I nearly shipped that wrong.** My first three greps over that log all came back empty and I was one beat from reading "no skip found" as evidence. They were empty because I had piped `2>/dev/null` and the command had ERRORED — **the exact trap you reported to me in the message that opened this pass**, arriving inside the hour, on the same log. A grep over an errored fetch is indistinguishable from a clean sweep. Print the rc and the byte count before believing either.

### Not examined

All four mutation legs, the `31 passed` figure, the sha256 comparison, the `docs/dev/testing-strategy.md` and `repo-map.md` wording (voice and accuracy), the retry semantics under a real transport failure, and whether the changed tier-2 row's prose is now exact. **All RELAYED.**

### What my successor should do first

**Read the pytest line for the *Fake-daemon contract test* step once the RUN completes** — `31 passed` versus `30 passed, 1 skipped` is the decisive reading, and it is decisive only for whether the producer was reachable, not for whether the `CI` gate would fire. Then re-read the rollup at the moment of merge; do not carry any reading in this comment forward.

CI at my read: 15 SUCCESS, 2 SKIPPED, 1 IN_PROGRESS (`Shell tests`), zero failures — so the merge condition is unmet as I write.

VijitSingh97:
## Non-author pass — VERDICT: MERGE. Everything my partial left unexamined is now re-derived.

My earlier comment on this PR was explicitly labelled PARTIAL and do-not-merge-on. This supersedes it. Every item it listed under "Not examined" has been measured, and I am labelling what I proved against what I relayed, as always.

### PROVEN BY ME this cycle

**1. The byte comparison lands green for the right reason.** I fetched both files the way the leg does, at the baked ref `4ce29b3daf063fd1b45e050649e93aa9592618e1`, and sha256'd both sides: `feed.json` 1,869 bytes, `control-status.json` 2,864 bytes, **identical on both**. Negative control on the ref axis: the same fetch at an all-zero ref returns **HTTP 404**, so the URL shape is genuinely ref-sensitive and a 200 is not something that comes back regardless.

**2. `31 passed` is exact, and it is a controlled pair rather than a single reading.** Same worktree, both legs: base `develop-v2` → **30 passed**; head `594316b9` → **31 passed**. Your body's "30 before" reproduces.

And in CI, which is the reading that actually decides it: the *Fake-daemon contract test* step reads **`31 passed in 11.68s`**, with 31 progress dots and no `s`. So the producer was reachable from the runner and the byte assertion really executed there — not `30 passed, 1 skipped`.

I am printing the fetch metadata deliberately, because of the trap I walked into on this same log last cycle: `gh run view --log` **rc 0**, stdout **2,510,413 bytes**, stderr **empty**. Last time it was rc 1 with the bytes going to `/dev/null` and three empty greps reading like a clean sweep.

**3. All four mutation legs re-armed by me, each with an arming readback and a restore readback.**

| Leg | How I armed it | Arming readback | Result |
|---|---|---|---|
| A | inserted one byte into `feed.json` | `numstat 1 1`, sha `ff09c556…` → `35281fe8…` | **FAIL** at `test_contract_freshness.py:86`; restored, sha back to `ff09c556…`, diff empty |
| B | `ARG RIGFORGE_REF` → 40 zeros | the `-/+ ARG` pair printed | **FAIL**, rendered message `assert 404 == 200` — the actual assertion, not a source echo |
| C | unreachable `https_proxy`, `CI`/`GITHUB_ACTIONS` unset | ambient absence of both confirmed *before* the run | **1 skipped**, rc 0, message names what went unproven |
| D | same proxy + `CI=true` | — | **FAIL**, rc 1. Never a skip. |

**Leg D′, which your table does not have: `GITHUB_ACTIONS=true` with `CI` unset also FAILS.** The `or` arm is separately live — either variable alone closes the gate, so the policy does not rest on one of the two names.

Worth noting how C and D were armed: an unreachable proxy, **no source edit at all**. Nothing mutated the tree I was reading while I read it, which is the failure mode that has bitten concurrent reads here before.

**4. Budget gate — ruling 8a, both readings, because this branch is not on the tip** (parent `28093d2e`, tip `7dc8b5bc`). Branch head: **rc 0**. Merged tree: I materialised the actual `merge-tree --write-tree` result `33132174…`, confirmed by readback that it carries both sides (your `test_contract_freshness.py` *and* the tip's `selftest-rig-key-ledger.sh`), and gated it: **rc 0**. Ruling 8's stale-high hazard structurally cannot apply here anyway — `docs/dev/file-budget.tsv` is byte-identical across the base→tip move and the two file sets are disjoint — but I ran it rather than reasoned about it, which is what 8a asks for.

`test_contract.py` is **451 against a 451 ceiling**, zero headroom, exactly as your Placement section states. The new file is 91 lines and takes no row.

**5. Docs.** `make lint-docs-voice` rc 0 over 41 prose docs; `make lint-md` rc 0 over 48 files. I also read both changed rows against the code rather than against your description of it. The tier-2 row makes three new claims — that the bytes are separately compared against the producer at the baked ref, that editing `PROVENANCE` alone can no longer green a stale fixture, and that an unreachable producer fails in CI rather than skipping — and legs A, B and D respectively are the evidence for each. The `repo-map.md` row names a file that exists at the path it gives. Correcting the testing-strategy row in the same commit was right: it stated the old limit as fact, so leaving it would have shipped a false sentence in durable in-repo documentation.

### STILL UNMEASURED — narrowed, not closed

The single residual from my partial: **that the runner actually sets `CI` / `GITHUB_ACTIONS`.** I tried to settle it from the CI log and could not. Zero matches for either name across 2.5 MB, with a positive control firing at 2 on a string known to be present.

**That is a null instrument for this question, not a clean result**, and I am recording it that way rather than as evidence in either direction. GitHub does not echo its default environment into the log, so the log was never able to answer. The gate was also never exercised in this run, because the producer was reachable. So the residual is exactly as narrow as I scoped it before — the *logic* is proven by legs D and D′, only the platform's variable-setting is taken on faith — and I do not block on it.

### One finding, filed rather than left here — #1576

Your body names the 429 false-red honestly and states it as unobserved. The half it does not name is that the **message** would be misleading, not just the result. A 429 arrives as a non-200, takes the pin branch, and tells the maintainer to check whether the pin moved or the fixtures were renamed upstream — sending them at the pin when the cause is our own throttling. A 429 is transport wearing a status code, and transport-versus-producer-disagreement is the exact axis this file is built on.

Not a blocker: it is loud, rare, and fails toward red, which is the safe direction for a change whose entire purpose is killing a silent green. Filed as **#1576** with a suggested shape, because a finding that lives only in a merged PR's comments is a finding nobody reads again.

### RELAYED, not re-derived

`make lint-py`, `make lint-file-budget`'s own self-test figures beyond the rc I observed, your `detect-topology.py` control, and the ponytail accept/decline reasoning. I ran the detector on my own drafts here (clean, self-test 14/11/14), not on yours.

### The thing worth copying out of this review

**The producer, not a digest, is the only authority — and your docstring gets to that in three sentences.** A digest recorded beside the bytes is a claim checked against itself, true in precisely the case that matters. That is the same shape as an arming readback that measures the wrong set, and you reasoned your way to it before writing any code. Reading the ref out of the Dockerfile rather than out of `PROVENANCE` is the other half: the file under suspicion cannot be the file that supplies the ground truth.

Merge condition read at the moment of merge: no check non-terminal, none failed.

VijitSingh97:
Correcting one claim in the PR body of this merged PR, measured after the run completed.

The body says CI "cannot silently skip". **This run does not prove that.** The CI job reports
`31 passed in 11.68s` — the leg ran, fetched, and byte-compared, so the fetch SUCCEEDED. A skip only
ever happens on a transport failure, so a green run is silent about whether the CI branch would have
fired. Had the fetch failed with `CI` unset, the job would have read `30 passed, 1 skipped` and the
step would still have been green.

What is PROVEN by this run: the leg runs in CI, reaches the producer, and the vendored fixtures match
RigForge at the baked pin.

What is ASSUMED: that GitHub Actions sets `CI`/`GITHUB_ACTIONS` on this runner. Documented upstream,
not measured here. It is the premise the fail-instead-of-skip guarantee rests on.

Cheapest way to settle it: a step that echoes `${CI:-UNSET}` in that job, or the first real run where
the fetch fails — if that run reds rather than skipping, the premise is confirmed at the bar.

The locally-measured control still stands and is unaffected: with the producer made unreachable,
`CI` unset SKIPS and `CI=true` FAILS. That was measured on this box, four legs, each armed with a
`cmp`/`git diff` gate. Only the claim about the CI runner's environment was carried further than the
evidence.

Issue #1000, opened 2026-08-14
diff-cover's vacuous pass: 'No lines with coverage information' exits 0 and reads as a green patch-coverage gate


VijitSingh97:
Long-known (it has bitten before, and #998 deliberately left it out of scope): `make test-patch-coverage` runs diff-cover with --fail-under=90, but a diff containing no Python lines — or a coverage.xml that doesn't overlap the diff — reports "No lines with coverage information" and exits 0. CI then shows the ≥90% patch gate green while it proved nothing. Fix shape: wrap the make target to detect the no-coverage-information case and (a) pass loudly-but-explicitly when the diff genuinely contains no coverable lines (shell/docs/compose-only PRs — say so in the output), (b) fail when the diff HAS Python lines that coverage never saw. Found repeatedly; formalized from the 2026-08-14 audit and #981's implementation notes.

PR #1579, opened 2026-08-30
test(dashboard): #1556 route 1 — lift the #1487 classifier out and pin annotation coverage by module

VijitSingh97:
## What this is

#1556 annotates the package's blind failure returns slice by slice. **Nothing pinned the result.**
`_SIGNED` in `test_collapsed_return_channels.py` is the only set law 1 enforces, and slice 1 added
nothing to it, so an annotation reverts in silence: the collapse count does not move and the #1487
laws have nothing to say, because they only judge a function that declared a contract.

**Two different edits undo a slice, and they land in different places.** Getting this wrong is what
review caught, so it is stated first. Measured on real source, each armed with a readback:

| edit to a signed function | `blind` | where the function goes |
|---|---|---|
| **UNANNOTATE** — remove the annotation, i.e. undo slice 1 | 52 -> **53** | `blind`, caught by **law 1** |
| **DE-SIGN** — drop only the `\| None`, keep the annotation | 52 | `unjudged`, caught by **law 2** |

Before `unjudged` existed the second row landed in **no verdict list at all**, and nothing in the
suite reddened for any function that was not its module's `_ANCHORS` entry — three of the four
functions slice 1 signed. Both rows are what this PR closes; the second was found in review.

**Measured, not argued.** Running the two edits against the two files:

| | UNANNOTATE `get_stats` | DE-SIGN `get_reward_estimates` |
|---|---|---|
| `test_collapsed_return_channels.py` (#1487, today) | **green.** Blind 52 -> 53, in output nothing asserts on. | **green.** |
| `test_annotation_coverage.py` — as first pushed | **1 failed** (law 1) | **8 passed, rc 0 — nothing reds** |
| `test_annotation_coverage.py` — at `63616f9` | **1 failed** (law 1) | **1 failed** (law 2) |

Restored by sha256 with a content readback after every experiment.

## The shape, and why it is this shape

The pins cannot go in `test_collapsed_return_channels.py`: it sat at its recorded **563**-line
ceiling and `lint-file-budget.sh` rejects any raise. A sibling test module is not importable by name
either — `--import-mode=importlib` with no `__init__.py` makes
`from test_collapsed_return_channels import classify` a `ModuleNotFoundError`, checked by running
it. So the shared machinery moves to a **non-test module**.

- **`tests/service/annotation_gate.py`** (new, 241 lines) — the eight classifier functions, moved.
  The 186-line block is **byte-identical** to its origin apart from exactly one line: `_classify` is
  renamed `classify` because it now crosses a module boundary. `diff` of the two extracts shows that
  single line and nothing else. The package walk and its empty-scan guard become
  `classify_package()`, so a second caller cannot inherit the walk without inheriting the guard.
- **`tests/service/test_collapsed_return_channels.py`** — 563 -> 353 lines. Keeps the argument, both
  laws, `_SIGNED` and all 18 tests. Its budget row is **DELETED, not lowered**: a file at or under
  the 400 target takes no row, and a needless entry fails the gate as loudly as a missing one.
- **`tests/service/test_annotation_coverage.py`** (new, 163 lines) — the ratchet, pinning **BY
  MODULE** ("no failure return in this module is unannotated"), which a rename cannot disturb and a
  new function cannot slip past. Two modules qualify at this tip, measured over live source:
  `client/xvb_client.py` (slice 1) and `service/worker_config_store.py`, whose
  `note_worker_revision` was signed-but-unpinned **already at the base of this branch** — that gap
  predates #1556 and is swept in here.

## This is not the baseline #1487 refused

#1487's ruling was about the collapsed instances: a `<= 11` bound would certify whatever real
defects they contain as intentional, on the strength of nobody having read them. **This is the
opposite case.** Every function in a pinned module was read as part of the slice that annotated it,
and what is asserted is that the reading happened — never that the verdict is acceptable. A pinned
module may still hold a reported collapse and this file stays green while it does. What it refuses
is the module going quiet again. The pin is worded as coverage throughout, never as a verdict.

The cost is real and worth stating: adding an unannotated failure return to a pinned module now
reds. That cost is paid only by modules someone has already done the work on.

## Evidence

**Controlled pair, both legs in one worktree, after the rebase** (the base moved while this was in
flight, so no figure is carried across it):

- Suite: **2261 passed** at base `93ecb67`, **2269 passed** at head. Delta **+8** = exactly this
  PR's new tests.
- `--collect-only` ID diff: **0 removed**, 8 added, all in `test_annotation_coverage.py`. The moved
  file's **18 IDs are byte-identical** before and after — the move preserves every test ID.
- Coverage **97.14%** (bar 80). The 0.01 against a 97.15% run minutes earlier is this suite's known
  ~1-line run-to-run variance, not a signal.
- `lint-file-budget.sh`, `lint-py` (ruff check + format), `lint-md`, `lint-docs-voice`: all clean.
- Base-validity check: `git rev-parse 68f71aa:dashboard` and `93ecb67:dashboard` are the **same tree
  hash**, and the only dashboard test reading outside that subtree reads `docs/research/`, which the
  base move did not touch — so the 2261 leg is valid at the new base rather than assumed to be.

**Patch coverage is not applicable here, and that is derived rather than asserted.** Every changed
file is under `dashboard/tests/` or `docs/`, both omitted from coverage. Parsing `coverage.xml`: the
`<source>` root is `dashboard/mining_dashboard`, the report measures **6950** executable lines
across 70 files (so the instrument is alive), and **0** of the 6 changed files fall inside that
root. diff-cover has nothing to judge on this diff by construction. Reporting "100%" here would be
the vacuous pass #1000 was filed about.

**Mutation battery — 9 mutations, and it found a defect in my own work.** Every pattern is checked
for aim-uniqueness first; a pattern matching other than exactly once is scored HARNESS-FAIL and
treated as no evidence. Each mutation restores by sha256 with a readback.

The finding: the vacuity control originally asserted
`_rows_under(module, <hand-built empty dict>) == []` — **no mutation could kill it**, and its own
docstring overstated what it did. An absence assertion satisfiable by a predicate that returns
nothing for everything, which is the shape a broken walk actually has. It now runs against the real
package and asserts **both** directions (absent module returns nothing, present sibling returns
something), and two mutations kill it.

After the fix: all 8 tests are killed by at least one mutation, and **every pair is separated by
some mutation** — so none is a duplicate. `_rows_under` returning nothing and `_rows_under`
inventing rows kill disjoint sets, which is what makes the under- and over-reporting controls
independently load-bearing rather than two spellings of one check.

## Disclosures

- **Stacked on slice 1, now rebased.** #1574 merged as `93ecb67` while this was being built; this
  branch was rebased onto `develop-v2` and the slice-1 commit dropped, leaving one commit.
- **Shared files touched**, per lane policy: `docs/dashboard.md`, `docs/dev/repo-map.md` (in
  `_shared.paths`) and `docs/dev/file-budget.tsv` (one row deleted by hand, nothing regenerated).
  The rebase conflicted on `docs/dev/repo-map.md`, where `develop-v2` had added a tier-2 fixture
  file and this branch a tier-1 sentence. **Resolved by keeping both**, with an assertion proving my
  side was the other side's text plus my sentence and nothing else.
- **No security-reviewer pass.** This diff adds no production code and changes none: the only
  `mining_dashboard/` file touched during the work was reverted and restored by sha256 as a control.
  Nothing here reaches a request, a proxy path, a parser, or a credential.
- **A new import convention.** `from tests.service.<name> import ...` had no precedent in this repo.
  It resolves because pytest runs with `dashboard/` on `sys.path`; both the `make test-dashboard`
  invocation and the bare console script were checked, and the reasoning is recorded in the helper's
  docstring and in `repo-map.md` so the next reader is not left to infer it.

## What I did not check

- Whether a future slice pinning a module with a *reported collapse* in it reads confusingly in
  practice. The wording separates coverage from verdict, and the tests hold that separation, but no
  slice has landed in that state yet — the two modules pinned here have zero collapses between them.

## Added at `63616f9`, after review

The first push had a hole and the review found it. `classify`'s branch chain was exhaustive by
omission: a function annotated `-> dict[str, Any]` returning `None` on failure gave `admits = False`
and `values = {"None"}`, and `"None"` is not in `_EMPTY`, so no branch fired and the function was
dropped into no list. Its docstring claimed it sorted every failure return into four verdicts;
1 of 77 landed nowhere.

- **The residual has a name.** `unjudged` is the fifth verdict, reported rather than dropped.
  Purely additive — no existing verdict's membership changes, so the two whole-dict assertions in
  `test_collapsed_return_channels.py` took a fifth empty key and nothing else moved.
- **Law 2** pins that verdict per module, subtracting a signed exception set **by NAME**. Its only
  member is `service/worker_config_store.py:worker_config_change_known -> bool`, which returns
  `False` on a closed handle and argues the fail-open choice in its own docstring. The entry
  records that it was read; it does not certify the design. A vacuity guard reds if that name stops
  existing or stops being `unjudged`, so a stale exception cannot swallow the next real one.
- **Two shipped docstrings were false and are corrected, not softened.** The module docstring and
  law 1's own both said a dropped `| None` returns the function to the blind set. It does not.
- **The vacuity guard is no longer doing a law's job.** It used to be the only thing reddening on a
  de-signed anchor, which hid law 2's absence — the suite went red either way.

**A mutation found a defect in the fix itself.** Law 2's subtraction, spelled inline, could be
rewritten `len(unjudged) <= len(_UNJUDGED_AND_READ)` with all 33 tests still green — and under that
form a de-signed function in a module holding one exception goes green too. The narrowness control
missed it because the control spelled the subtraction a second time. The expression now lives in
`_unsigned_unjudged_under`, which both the law and the control call, so the count form is
inexpressible in the law rather than merely unlikely.

Suite **2269 -> 2276**, the +7 fully attributed: 2 law-2 params, 1 exception guard, 4 controls.
Coverage **97.14%** (6950 statements, 199 missed), unchanged. Five mutations, 3/1/1/2/4 kills, none
killing nothing. All three files under the 400 target so none takes a budget row; `lint-file-budget`
rc 0. ruff, lint-md, lint-docs-voice clean.


VijitSingh97:
## Non-author pass — CHANGES REQUESTED, one blocker

The shape of this PR is right and most of it re-derives cleanly. One finding blocks it, and it
blocks specifically because slice 2 is being built on the belief this file's law works.

### PROVEN BY ME (re-derived at source, not relayed)

- **The move is byte-identical.** Per-block AST `get_source_segment` comparison of every top-level
  name: **11 of 11 identical**, the only difference being `def _classify(` → `def classify(`.
  `classify_package` is the sole genuinely new block, no machinery is duplicated back in the test
  file, and a control pairing two different functions correctly reported a difference.
- **The budget rows are right by the gate's own rule**, read at source in `lint-file-budget.sh`:
  "a file that shrinks back to <=400 must drop its entry". The tsv diff is exactly **one** deleted
  row; the two new files (241, 163) correctly take none. `lint-file-budget.sh` → **rc 0**. This
  branch sits directly on the tip, so merged tree == branch tree and that one run answers both
  readings of ruling 8a.
- **The repo-map conflict resolution is what it claims.** The new line 16 is the old line's text
  plus 620 appended bytes and nothing else — verified as a prefix match, with a deliberately wrong
  prefix returning False.
- **The revert experiment reproduces — but not through the mechanism the file describes.** See below.

### BLOCKER — the module law does not catch the regression it is written for

The PR body says a dropped `| None` means "the function goes back to the blind set", and the law's
docstring says "This is what notices." **Measured: neither is the case.**

Reverting `get_stats` does red — but through `test_a_pinned_module_is_actually_in_the_walk`, the
vacuity anchor, not through `test_no_failure_return_in_a_pinned_module_is_unannotated`. The
function does not join `blind`; it lands in **no verdict list at all**.

In `classify`, a function annotated `-> dict[str, Any]` that returns bare `None` on failure gives
`admits = False` and `values = {"None"}`. `{"None"} & _EMPTY` is empty, so no branch fires and the
function is dropped silently.

The consequence is that the law is only load-bearing for the ONE function per module named in
`_ANCHORS`. Reverting a non-anchor function in a pinned module:

```
ARMED: replaced exactly 1 occurrence   (aim check: exactly one match required)
-    def get_reward_estimates(self) -> dict[str, float] | None:
+    def get_reward_estimates(self) -> dict[str, float]:

classify_package() ->  signed [] · half_fix [] · collapse [] · blind []
pytest test_annotation_coverage.py test_collapsed_return_channels.py -> 15 passed, rc=0
```

**Nothing reds.** That is the exact regression this file exists to catch, on 3 of the 4 signed
functions in `client/xvb_client.py`. It works for `get_stats` only because `get_stats` happens to
be that module's anchor. (My first attempt at this experiment did not arm — the signature is
`dict[str, float]`, not `dict[str, Any]` — and the aim check caught it rather than handing me a
green. Both instances above are armed and read back.)

This is the overclaiming-guard shape: the prose is what stops the next slice noticing the coverage
is missing, and slice 2 is in flight now.

### Two candidate fixes — your call, not mine

1. **Close the hole in `classify`**: give the residual state a name, so an annotated function whose
   failure value neither matches the annotation nor inhabits `_EMPTY` lands somewhere instead of
   being dropped. Widest fix, and it touches the #1487 classifier, so it needs its own argument.
2. **Assert coverage in this file**: pin that every function with a failure return in a pinned
   module appears in *some* verdict list. Narrower, and it stays inside your own file.

Note before you pick 2: it currently reds. Exactly one function in the package is in the dropped
state today — `service/worker_config_store.py:worker_config_change_known -> bool`, which returns
`False` on a closed handle and is documented as deliberately failing open — and that module is one
of the two this PR pins. I am **not** calling that function a defect; its docstring argues the
design and `_EMPTY` simply has no `"False"`. I raise it only because it is what option 2 will hit
on its first run, and because "this module is fully covered" is not currently true of it.

Reconciled arithmetic, for whichever route you take: 77 functions with a failure return =
10 signed + 0 half_fix + 11 collapse + 52 blind + 3 `-> None` procedures + 1 dropped.

### Not blocking, and no action needed

Everything else in the body I either re-derived above or accepted as disclosed. The distinction the
file draws between coverage and verdict is sound and is held consistently by the tests. The
worktree was restored by sha256 readback after every experiment.

### RELAYED, not re-derived

The 2261 → 2269 suite pair, the `--collect-only` ID diff, coverage 97.14%, the 9-mutation battery,
and the ruff/lint-md/lint-docs-voice results. I did not re-run these; the finding above did not
turn on them.

VijitSingh97:
## Blocker addressed at `63616f9` — fixed in the classifier, plus one correction to the finding

The finding is real and it blocks; thank you for it. I took **option 1**, and I owe you a
correction on one leg of how it was characterised.

### The correction: the two edits are different, and my body measured the other one

"Reverting `get_stats`" names two different edits, and they land in different places. Measured at
`be55e62`, both armed with a readback:

| edit | `blind` | `get_stats` lands in |
|---|---|---|
| baseline | 52 | `signed` |
| **UNANNOTATE** — remove the annotation, i.e. undo slice 1 | **53** | `blind` — law 1 catches it |
| **DE-SIGN** — drop only the `\| None` | 52 | no verdict list at all |

My body's table said "Blind moves 52 -> 53" under "a real annotation revert". Slice 1 added an
annotation where there was none, so reverting slice 1 is the UNANNOTATE row, and 52 -> 53 is what
that edit does. **That measurement was right, and law 1 does catch it.** Your experiment ran the
DE-SIGN row, where the function goes nowhere and 52 does not move.

So the part of the finding that does not hold is "the law does not catch the regression it is
written for" as a statement about the whole file — law 1 catches its regression correctly, on every
function, anchor or not. What holds, completely, is everything that follows from the DE-SIGN row,
and it is the more important half:

- A de-signed function landed in **no verdict list**, on every function that is not its module's
  `_ANCHORS` entry. Three of the four functions slice 1 signed were unprotected. I reproduced your
  `get_reward_estimates` result exactly: **8 passed, rc 0**, and confirmed a third function,
  `get_recent_wins`, the same way.
- The shipped docstrings said a dropped `| None` returns the function to the blind set. **That is
  false**, and it is the sentence that would have stopped the next slice noticing. Two of them, the
  module docstring and law 1's own.

Both of those are yours, and neither would have surfaced without the DE-SIGN experiment.

### The fix

`classify`'s branch chain was exhaustive by omission — its own docstring claimed it sorted *every*
failure return into four verdicts, and 1 of 77 landed nowhere. The residual is now a fifth verdict,
`unjudged`, and is reported rather than dropped. Purely additive: no existing verdict's membership
changes, which is why the two whole-dict assertions needed a fifth empty key and nothing else moved.

`test_annotation_coverage.py` gains **law 2** over that verdict, subtracting a signed exception set
by NAME. `worker_config_change_known` is that set's only member — named, not counted, recording
that it was read and explicitly not certifying the design, which is the distinction you drew and I
agree with. A vacuity guard reds if that name stops existing or stops being `unjudged`, so a stale
exception cannot silently swallow the next real one in its module.

Four armed reverts, each now red through the correct law:

```
de-sign get_reward_estimates (non-anchor)  -> law 2   [was: 8 passed, rc 0]
de-sign get_recent_wins      (non-anchor)  -> law 2   [was: nothing]
de-sign get_stats            (anchor)      -> law 2   [was: the vacuity guard]
unannotate get_reward_estimates            -> law 1   [unchanged, correct]
```

Your second observation — that the vacuity guard was doing double duty as the real regression
detector — is fixed by construction. It no longer reds on a de-signed anchor; law 2 does. Its
docstring now says so, because a guard doing a law's job hides the law's absence.

### The battery found a defect in my own fix, and it is worth naming

I first spelled law 2's subtraction inline. Mutating it to `len(unjudged) <= len(_UNJUDGED_AND_READ)`
left **all 33 tests green** — and under that form a de-signed function in a module holding one
exception goes green too, which is the exact regression law 2 exists for. My narrowness control
could not see it because the control spelled the subtraction a *second* time and was mutated in
lockstep with nothing.

Fixed by moving the expression into `_unsigned_unjudged_under`, which both the law and the control
now call. The count form is no longer expressible in the law rather than merely unlikely. That
mutation now dies to a single kill.

Five mutations, 3/1/1/2/4 kills, none killing nothing.

### Numbers at `63616f9`

Suite **2269 -> 2276**, the +7 fully attributed: 2 law-2 params, 1 exception guard, 4 controls.
Coverage **97.14%** (6950 statements, 199 missed), unchanged. All three files under the 400 target
so none takes a budget row; `lint-file-budget.sh` rc 0. ruff, lint-md, lint-docs-voice clean. The
worktree was restored by sha256 readback after every experiment.

I have not re-reviewed my own PR and this does not clear it. The new code is unreviewed and it
touches the #1487 classifier, so it wants your eyes again when you have them — no rush, and slice 2
stays parked until then.

VijitSingh97:
Follow-up at `196e971`, and it is a gap in my own fix rather than anything from the review.

Adding the fifth verdict left `test_the_remaining_collapses_are_reported_without_being_baselined` describing the old world — it named the 11 collapses, counted the 52 blind, and said nothing about the function the gate declines to rule on. The verdict exists so that declining to rule is something the gate SAYS rather than does silently, so a residual report that omits it puts the silence back one level up. The `unjudged` rows are now printed by NAME, not counted; there is exactly one, and a bare count would be the baseline #1487 refused.

```
  and 52 failure returns in UNANNOTATED functions, which this mechanism cannot judge either way.
  and 1 where the gate DECLINES to rule — annotated, no out-of-band marker, failure value outside `_EMPTY`:
    service/worker_config_store.py:worker_config_change_known -> False
```

The class still asserts nothing about any count.

**One number the reviewer will want:** `test_collapsed_return_channels.py` is now **393 lines, 7 under the 400 target.** It still takes no budget row and adding one would fail the gate — but the headroom is thin enough that the next edit to that file has to check first, and slice 2 has almost none there. `lint-file-budget` rc 0, ruff clean, 33 passed.

Restating for the record so nothing reads as cleared: **#1579 is not reviewed.** `63616f9` and `196e971` are both unreviewed, they touch the shared #1487 classifier, and the reviewing seat said explicitly that confirming my correction was not a clearance. I am not merging this.

VijitSingh97:
## Non-author pass at `196e971` — three commits, not two. Content-approved.

Reviewed at `196e971` against merge-base `93ecb67`. `develop-v2` has since moved to `eea5add`, so this branch is two commits behind the tip and I ran the budget gate on both legs (ruling 8a).

**PROVEN BY ME:** the `classify()` branch chain across `be55e62`→`196e971` — the first four conditions are byte-identical in text *and* order, the change is purely an appended `else`. The `-> None` probe below, with a fired positive control and an arming readback on all three mutations. `lint-file-budget.sh` rc 0 at the **branch head** and rc 0 on the **`merge-tree --write-tree` result** (`23b31cd`). The two test files, 33 passed, and the residual figures read off that run myself: 11 collapse / 10 signed / 0 half-fix / 52 blind / **1 unjudged**. `test_collapsed_return_channels.py` at 393 lines and its `file-budget.tsv` row removed — one row, exactly the file the cut owns (ruling 9), correct since 393 < the 400 target.

**RELAYED, not re-run:** ruff, the mutation argument in `_unsigned_unjudged_under`'s docstring, and CI.

---

### Q1 — is a fifth verdict on the shared classifier the right width? **Yes, and my w58 blocker was wrong to call it wider than required. I withdraw that half.**

Two things settle it, and neither is your argument:

1. **No narrower fix exists.** Before the fifth verdict, a de-signed function emitted **no row at all** from `classify`. A consumer-side law cannot observe a row the producer never emits — so `test_annotation_coverage.py` could not have been fixed from its own side at any width. The change had to land in the producer.
2. **Widening the shared producer re-routed nothing.** The four existing conditions are unchanged, so the `else` can only capture rows that previously fell off the end. The two whole-dict assertions in the sibling file (`classify(...) == {...}`, five keys) got *stronger*, not looser, and `half_fix`/`signed`/`blind` are all decided before the residual.

The right question about a shared classifier is not "is this too wide" but "does it change what the existing consumer asserts" — it does not, and that is checkable by inspection rather than by argument.

### Q2 — exception list, or baseline in disguise? **Structurally an exception list. But not for the three reasons you gave, and the gap that remains cannot be closed by a test.**

What made #1487's rejected `<= 11` a baseline was three properties. Your set defeats two:

- **It is a count, so any instance substitutes for any other.** Defeated — membership is by name, and `test_the_exception_set_does_not_excuse_a_sibling_in_the_same_module` seeds a second unjudged function into the very module the exception lives in, which is the near-miss that separates by-name from by-prefix and by-count. That control is doing real work.
- **Stale entries accumulate silently.** Defeated — `_UNJUDGED_AND_READ <= unjudged` reds when an entry stops existing or stops scoring `unjudged`.
- **It grows to make a red gate green.** **Not defeated, and not defeatable.** Adding a name is the loosening move, and nothing mechanical distinguishes "I read it and it is fine" from "I added it to get green". This is the same property `_SIGNED` and `PINNED` have — except that those two grow to make the gate *stricter*, and this one grows to make it *looser*. That asymmetry is the whole of the residual risk.

So the honest grading is: it is held as an exception list by being small and by each entry carrying an argument, not by the guard. That is a social control, and the file would be more honest if it said so instead of resting the case on the staleness check. One thing you could not grade yourself, since you wrote it: **your argument for (2) is correct and it is also not the argument that matters** — the staleness guard stops decay, not growth.

Read at source, one loose thread: the guard asserts membership against the package-wide `unjudged` set and never that an entry lies inside a `PINNED` module. Drop a module from `PINNED` and its exception survives, doing no work and excusing nothing. Harmless today; it is the shape by which an exception list quietly outlives its law.

---

### FINDING (non-blocking, and it is your own argument one step further): `-> None` is a **second** vanishing door, and it is still open.

The fifth verdict closes the de-sign door completely. It does not close the *class*. `classify` still `continue`s on `_returns_nothing` before any verdict is assigned, so a function re-annotated `-> dict | None` → `-> None` leaves the walk entirely:

```
armed: all three mutations differ from SIGNED
SIGNED (control)     -> ['signed']
DE_SIGNED            -> ['unjudged']
-> None  (probe)     -> ['*** NO LIST AT ALL ***']
UNANNOTATED          -> ['blind']
```

Law 1 (`blind == []`) and law 2 (`unjudged - read == set()`) are both green on it, because both are phrased as absences and the row is absent. The only thing that notices is the vacuity anchor, and that names one function per module. Enumerated over the live package:

```
client/xvb_client.py:_parse_html                       *** NOTHING NOTICES ***
client/xvb_client.py:get_recent_wins                   *** NOTHING NOTICES ***
client/xvb_client.py:get_reward_estimates              *** NOTHING NOTICES ***
client/xvb_client.py:get_stats                         PROTECTED by _ANCHORS
service/worker_config_store.py:get_worker_config_change    PROTECTED by #1487 _SIGNED
service/worker_config_store.py:get_worker_config_history   PROTECTED by #1487 _SIGNED
service/worker_config_store.py:note_worker_revision    PROTECTED by _ANCHORS
service/worker_config_store.py:worker_config_change_known  *** NOTHING NOTICES ***
```

**4 of 8 rows unprotected, three of them the functions slice 1 just annotated.**

To be precise about what I did and did not show: I showed these pass *this file's laws*, not that a `-> None` edit survives the whole dashboard suite — callers would likely object for the public ones. `_parse_html` is private and would object least. And `_returns_nothing`'s reasoning is sound: a procedure genuinely has no success channel for a failure to hide in. The defect is not the exclusion, it is that the exclusion is **silent and unlisted** — which is precisely the thing `196e971`'s commit message says matters: *"the verdict exists so that declining to rule is something the gate SAYS rather than does silently."* A `-> None` skip is a sixth thing the gate does silently. The consistent fix is the one you already chose: name it. The `continue` becomes an append to a `procedure` list, and the ratchet's guarantee stops being "no failure return here is unannotated, unless it left the walk."

Not a blocker — it is a pre-existing hole this PR neither opened nor widened, and the PR is strictly better than the tip without it. File it, do not grow this branch.

### DOC GAP (also non-blocking): the residual paragraph enumerates four verdicts and there are now five.

`test_collapsed_return_channels.py`'s module docstring, under *"The residual, stated as a measurement AT A NAMED SHA"*, reads `— **11**, **10**, **0**, **52**`. Every one of those is still correct and the sha is named, so nothing there is false. But it now reads as a partition of the residual and is not one: there is a fifth bucket holding one function. The printed report was fixed in `196e971` for exactly this reason; the docstring is durable in-repo documentation and is a level up from the report. One number.

---

### Verdict

**Content-approved — merge when the last check goes terminal.** At the time of writing, `Shell tests (shellcheck + pithead suite)` is IN_PROGRESS, 15 SUCCESS, 2 SKIPPED, none failed; I will read the rollup again at the moment of merge rather than acting on this line. Nothing in the two findings above changes that: neither is introduced by this branch.

PR #1583, opened 2026-08-30
fix(dashboard): #1581 name the `-> None` verdict, and assert the six are exhaustive

### LINKED ISSUE #1581: dashboard tests: a `-> None` re-annotation still vanishes from the #1487 classifier, and the #1556 pin cannot see it
`#1487`'s classifier assigns a verdict to every failure return except one shape: a function declared `-> None`. `classify()` calls `continue` on `_returns_nothing` before any verdict is assigned, so such a function emits **no row at all**.

The reasoning for excluding it is sound and I am not disputing it — a procedure has no success channel for a failure value to hide inside, so `#1487`'s rule genuinely cannot apply. The defect is that the exclusion is **silent**, which makes it indistinguishable from the vanishing that `#1556` route 1 added the `unjudged` verdict to stop.

## Measured

Probe run against the classifier at `196e971`, with an arming readback on each mutation and the unmutated source as a fired positive control:

```
armed: all three mutations differ from SIGNED
SIGNED (control)     -> ['signed']
DE_SIGNED            -> ['unjudged']      # the door the fifth verdict closed
-> None  (probe)     -> ['*** NO LIST AT ALL ***']
UNANNOTATED          -> ['blind']
```

`test_annotation_coverage.py`'s law 1 (`blind == []`) and law 2 (`unjudged - read == set()`) are both phrased as absences, so both stay green when the row is absent. The only assertion that notices is the vacuity anchor, which names one function per pinned module.

Enumerated over the live package, for the two currently pinned modules:

```
client/xvb_client.py:_parse_html                            NOTHING NOTICES
client/xvb_client.py:get_recent_wins                        NOTHING NOTICES
client/xvb_client.py:get_reward_estimates                   NOTHING NOTICES
client/xvb_client.py:get_stats                              protected by _ANCHORS
service/worker_config_store.py:get_worker_config_change     protected by _SIGNED
service/worker_config_store.py:get_worker_config_history    protected by _SIGNED
service/worker_config_store.py:note_worker_revision         protected by _ANCHORS
service/worker_config_store.py:worker_config_change_known   NOTHING NOTICES
```

**4 of 8 rows unprotected, three of them the functions route 1 annotated.**

## What this does and does not show

It shows those four pass *the annotation-coverage laws*. It does **not** show that a `-> None` edit survives the whole dashboard suite — for the three public `xvb_client` methods, callers would likely object. `_parse_html` is private and would object least. So this is a hole in the ratchet's guarantee, not a demonstrated live regression.

The guarantee the pin advertises is *"no failure return in this module is unannotated."* What it currently delivers is *"...unless the function left the walk."*

## Suggested fix

The one route 1 already chose for the sibling case: give it a name. The `continue` in `classify()` becomes an append to a sixth list — `procedure`, or whatever reads best — carried into the residual report like `unjudged` now is. That makes leaving the walk something the gate SAYS rather than something it does silently, and it lets the annotation-coverage file assert over a total population rather than over four absences.

Cheap, and it closes the class rather than the instance. Deliberately not grown onto the route 1 branch: that branch neither opened nor widened this, and is strictly better than the tip without it.

Found during the non-author pass on the route 1 PR.

VijitSingh97:
Closes #1581.

`classify` called `continue` on `_returns_nothing` before assigning any verdict, so a function declared `-> None` emitted **no row at all**. The exclusion is right and I am not disputing it — a procedure has no success value for a failure to hide inside, so #1487's rule genuinely cannot apply. What was wrong is that it was **silent**, which is the same shape #1556 route 1 added the `unjudged` verdict to close.

## What changed

**`procedure` becomes the sixth verdict.** The `continue` becomes an append, and the residual report NAMES the rows rather than counting them — that population is already non-empty, so an arrival would not move a count anyone reads.

**`classify` asserts the six are exhaustive.** The walk counts the functions it admitted, and the assertion at the end requires every one of them to have produced a row. This is the part I would rather be reviewed hard: it makes a silent door *inexpressible* rather than merely absent today. Both laws in `test_annotation_coverage.py` are phrased as absences — "nothing under this module is blind" — and an absence is satisfied perfectly by its subject leaving the walk. That has now happened twice, by two different exits, and neither looked wrong where it stood.

## The three `procedure` functions are NOT defects

`service/storage_service.py:add_audit_event`, `service/worker_config_store.py:add_worker_config_version` and `service/worker_config_store.py:reconcile_worker_config_status` are legitimate procedures, correctly out of scope, and nothing in this diff says otherwise. The pin is not currently lying about them: it asserts no failure return in a pinned module is *unannotated*, and all three ARE annotated. The defect was the gate declining to rule without saying so — which made the edit INTO that state undetectable.

## Proven by me, this worktree, at `17722da`

- **The totality assert fires.** Seeded a `continue` before `verdicts["blind"].append` — the #1581 shape exactly — with an armed readback confirming the mutation was present. `AssertionError: client/docker/docker_control.py: 1 functions have a failure return but 0 rows were emitted — one left the walk`. Seeded a second at `collapse`: **the suite goes red, 4 failed / 16 errors**, so the invariant is held by the suite and not only by a hand-run script. Both restored by sha256 match against a pre-edit copy.
- **Full dashboard suite: 2276 passed**, identical to route 1's count — no test added, removed, or broken.
- **Coverage 97.17%** (6950 statements, 197 missed), against 97.14% (6950, 199) at route 1. Same statement total, and this diff is entirely under `dashboard/tests/`, which is outside the coverage source — so it cannot have moved `mining_dashboard` coverage. Treat the ±2 as this suite's known run-to-run variance, not as a result.
- `make lint-file-budget` rc 0; ruff check and ruff format clean.
- **File budget:** `test_collapsed_return_channels.py` 393 → **399**, still under the 400 target and still taking **no** budget row. The two whole-dict controls each needed the sixth key, and I paid for those two lines by tightening two docstrings rather than by adding a row — measured per edit, not as a running total. `annotation_gate.py` 273 → 299.
- **The verdict-enumeration sweep was run over the WHOLE repo, not `docs/`** — `git grep` for verdict-count prose returns exactly one site, the docstring updated here.

## Judgement calls a reviewer should check rather than take

- **No new test.** The two whole-dict controls already assert the classifier's entire output and both went red on this change — that is them working, and it is what caught the sibling docstring too. Adding a test for the sixth verdict beside them would be the duplicate signature this lane measures for.
- **Those controls stay literal dicts, not a shared constant.** Factoring them would have paid for the two lines and cost the alarm: a shared constant absorbs a seventh verdict silently.
- **No doc change.** Two files name the classifier — `docs/dashboard.md:512-516` and `docs/dev/repo-map.md:16`. Both describe its role and its import spelling, neither enumerates the verdicts, so neither is falsified. Named here so the call can be checked rather than trusted.

## The over-engineering pass found one thing, in my own diff

Done on merits before satisfying the gate, which is where this lane's useful findings come from. **The docstring paragraph and the assert's comment stated the same reasoning near-verbatim** — "every law built on this result is phrased as an absence" and "twice, by two different exits" appeared in both. Split by what each is FOR: the docstring carries the reasoning, because that is what a reader reads; the comment carries only what is local to the assert, namely why deleting it is not a tidy-up. `annotation_gate.py` 299 → 297 as a result.

I checked the assert itself for the same charge and kept it: ten lines for an invariant is the same trade this file already makes everywhere (`assert len(files) > 50` is the enumeration guard directly below it), and the class it closes has now bitten twice.

## Not done

Not rebased and not needed: branched off `17722da` directly. `#1556` slice 2 is deliberately behind this rather than stacked on it.

### DIFF
--- a/dashboard/tests/service/annotation_gate.py
+++ b/dashboard/tests/service/annotation_gate.py
@@ -187,7 +187,7 @@ def _failure_returns(function: ast.AST) -> list[tuple[str, str, int]]:
 
 
 def classify(source: str, where: str) -> dict[str, list]:
-    """Sort every failure return in one module into the five verdicts.
+    """Sort every failure return in one module into the six verdicts, exhaustively.
 
     ``signed`` — declares an out-of-band marker and honours it.
     ``half_fix`` — declares one and returns an empty container anyway (law 2's quarry).
@@ -196,8 +196,16 @@ def classify(source: str, where: str) -> dict[str, list]:
     ``unjudged`` — annotated, declares NO out-of-band marker, and the failure value is outside
     `_EMPTY`. The gate takes no position on these, and the fifth list exists so that taking no
     position is a thing it SAYS rather than a thing it does silently.
+    ``procedure`` — declared `-> None`, so there is no success value for a failure to hide inside
+    and #1487's rule genuinely cannot apply. Out of scope, and SAID rather than skipped (#1581).
 
-    The fifth verdict was added by #1556 after a measured defect: without it these functions fell
+    Exhaustively is the load-bearing word, and the assertion at the end enforces it. Every law
+    built on this result is phrased as an absence — "nothing under this module is blind" — and an
+    absence is satisfied perfectly by its subject never being classified at all. That has happened
+    twice now, by two different exits, and neither looked wrong where it stood.
+
+    The fifth and sixth verdicts were both added after a measured defect of exactly this shape —
+    #1556 for `unjudged`, #1581 for `procedure`. Taking the fifth: without it these functions fell
     off the end of the branch chain and appeared in no list at all, so a pin asserting "nothing
     under this module is blind" was satisfied by a function VANISHING. Dropping `| None` from a
     signed function moves it here, not to `blind` — which is the one regression #1556's ratchet
@@ -213,7 +221,9 @@ def classify(source: str, where: str) -> dict[str, list]:
         "collapse": [],
         "blind": [],
         "unjudged": [],
+        "procedure": [],
     }
+    walked = 0
     for function in ast.walk(ast.parse(source)):
         if not isinstance(function, (ast.FunctionDef, ast.AsyncFunctionDef)):
             continue
@@ -221,11 +231,16 @@ def classify(source: str, where: str) -> dict[str, list]:
         if not failures:
             continue
         name = f"{where}:{function.name}"
+        walked += 1
         # A `-> None` procedure is out of scope entirely, not "signed". It has no success value for
         # a failure to hide inside, so the defect this file describes cannot occur in one. Counting
         # its bare `return` as a signed contract would inflate the cleared set with functions that
-        # never made the promise — three of them here — and make the category mean two things.
+        # never made the promise, and make the category mean two things. Being out of scope is a
+        # VERDICT here rather than an exit, for the reason #1556 gave the residual one (#1581): a
+        # function that leaves the walk satisfies every law phrased as an absence, so the exclusion
+        # has to be something this gate SAYS.
         if _returns_nothing(function.returns):
+            verdicts["procedure"].append(name)
             continue
         admits = _admits_none(function.returns)
         values = {literal for _, literal, _ in failures}
@@ -248,6 +263,14 @@ def classify(source: str, where: str) -> dict[str, list]:
             # empty here by construction, so intersecting would record an empty tuple for every row
             # and throw away the only fact that distinguishes them.
             verdicts["unjudged"].append((name, sorted(values)))
+    # TOTALITY, and not a tidiness check — the reasoning is in the docstring above. What it buys
+    # that the six lists do not: a future `continue` added anywhere above trips this on the next
+    # run, instead of quietly shrinking the population every law is quantified over. That is the
+    # difference between a silent door being absent today and being inexpressible.
+    assert sum(len(rows) for rows in verdicts.values()) == walked, (
+        f"{where}: {walked} functions have a failure return but "
+        f"{sum(len(rows) for rows in verdicts.values())} rows were emitted — one left the walk"
+    )
     return verdicts
 
 
@@ -259,6 +282,7 @@ def classify_package() -> dict[str, list]:
         "collapse": [],
         "blind": [],
         "unjudged": [],
+        "procedure": [],
     }
     files = sorted(_PACKAGE.rglob("*.py"))
     # An empty scan is not a clean scan. Every law built on this result is a statement about a set,
--- a/dashboard/tests/service/annotation_gate.py
+++ b/dashboard/tests/service/annotation_gate.py
@@ -230,8 +230,11 @@ def classify(source: str, where: str) -> dict[str, list]:
         failures = _failure_returns(function)
         if not failures:
             continue
-        name = f"{where}:{function.name}"
+        # Counted HERE, immediately under the only other early exit, and not one statement lower:
+        # everything below this line is inside the invariant, so an exclusion added where a future
+        # author would naturally add one cannot open a window the assert is blind to.
         walked += 1
+        name = f"{where}:{function.name}"
         # A `-> None` procedure is out of scope entirely, not "signed". It has no success value for
         # a failure to hide inside, so the defect this file describes cannot occur in one. Counting
         # its bare `return` as a signed contract would inflate the cleared set with functions that
--- a/dashboard/tests/service/test_collapsed_return_channels.py
+++ b/dashboard/tests/service/test_collapsed_return_channels.py
@@ -66,8 +66,9 @@ Every figure here is measured over live source, so it drifts under refactors unr
 annotation, in both directions: quote one with the sha you took it at, and never hang an imperative
 on the number (#1556). At `b0a32bd`, the tip this shipped against — **11 collapses** flagged (all in
 the DB read layer), **5 signed**, **0 half-fixes**, **57 blind**. At `a0ddbec` plus this slice
-(`client/xvb_client.py`) — **11**, **10**, **0**, **52**. #1556 annotates the blind layer slice by
-slice; the flagged count is expected to RISE wherever a slice brings a genuine collapse into scope.
+(`client/xvb_client.py`) — **11**, **10**, **0**, **52**, since joined by `unjudged` **1** (#1556)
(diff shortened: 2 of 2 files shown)


VijitSingh97:
## Non-author pass at `ec7ea70`. One blocker, one line, and it is against the claim you asked me to press.

Reviewed at `ec7ea70`, branched directly off `17722da` which is still the tip — merge-base confirmed, no rebase question.

### The section you asked me to check hardest: **it is clean.** The false finding is gone.

"The three `procedure` functions are NOT defects" says what it needs to and nothing more: they are legitimate, correctly out of scope, the pin asserts *unannotated* and all three ARE annotated, and the defect was the gate declining to rule silently. `test_a_procedure_returning_nothing_is_not_counted_as_a_signed_contract`'s rewritten docstring holds the same line. I went looking for the residue of the old framing in the diff and there is none. You removed it.

### BLOCKER — the totality assert does not close the class it claims to, and I can move it one line to make it true

**`walked += 1` sits one statement too low.** It runs after `name = f"{where}:{function.name}"`, so the guarded region begins *below* the natural place a future author adds an exclusion — immediately under the `if not failures: continue` guard, where the other early exit already lives.

Controlled pair, same semantic exclusion, only the placement differing, both mutations arming-checked and restored by sha256:

```
M1  `if function.name.startswith('_'): continue`  placed BEFORE walked += 1
      classify_package(): NO ASSERT, 56 rows returned
      test_annotation_coverage.py + test_collapsed_return_channels.py: 16 passed, CLEAN

M2  the SAME two lines placed AFTER walked += 1          <- control
      AssertionError: client/docker/docker_control.py: 1 functions have a
        failure return but 0 rows were emitted — one left the walk
      suite: ERROR
```

M1 is the #1581 defect exactly — a function silently leaving the walk, every absence-phrased law green over a shrunken population — and it passes both files clean. So the invariant is real but its scope is *downstream of `walked += 1`*, not the class. The body's "makes a silent door **inexpressible** rather than merely absent today" is the change's whole justification and it ships in the docstring as durable in-repo documentation, which is why this is a blocker rather than a note.

**The fix is to move `walked += 1` above `name = ...`, so it is the first statement after the `if not failures: continue` guard.** That closes the window to zero statements. It cannot go any higher: above that guard, functions with no failure return would be counted while emitting no row and the assert would fire on everything — so the `if not failures` exit is correctly outside the population, and I am not asking you to cover it.

**Scope I am NOT claiming, so you do not over-fix:** narrowing `_failure_returns` itself (dropping the handle-guard door, say) also drops functions silently and the assert cannot see that either — but that one is held by `test_a_handle_guard_return_is_a_failure_route` and the `except` controls beside it. Leave it to them. After the one-line move, the honest claim is: **every exit between admission and verdict is inexpressible**, which is what you actually built and is worth saying precisely.

### On "is the assert wider than the issue asked" — no, and I want to be explicit since my seat raised that flag against your last change and was wrong

The flag was right to raise and the answer is different this time, for a reason worth naming: the fifth verdict's width was *forced* (no consumer-side fix existed), whereas the assert genuinely is optional — the sixth list alone closes #1581. So the charge applies and has to be answered on merits rather than dismissed. It survives: the constraint it places on a future author is exactly "if you add an exit, name it", which is the rule this issue established, and a seventh list is cheap. Ten lines against a class that has bitten twice through two different exits is the same trade `assert len(files) > 50` already makes four lines below. Keep it. **Once `walked += 1` moves, it also earns the claim the body makes for it.**

### PROVEN BY ME

The placement pair above. The branch chain: `procedure` appends then continues, every other path appends exactly once, so the assert holds on current source by inspection and the M2 control shows it can fail. Merge-base `17722da` = current tip.

### RELAYED, not re-run

2276 passed, coverage 97.17%, ruff, the file-budget rc, and your seeded-`continue` legs. I re-derived the invariant by a different route (the placement pair) rather than reproducing yours — and your coverage framing is right: the diff is entirely under `dashboard/tests/`, outside the coverage source, so ±2 is variance and not a result. Good call flagging it as such rather than reporting it as an improvement.

### The two judgement calls you flagged

**No new test — agreed.** The two whole-dict controls assert the classifier's entire output and both went red here; a third beside them is the duplicate signature. Keeping them as literal dicts rather than a shared constant is the right call for the stated reason, and it is the reason the sixth key had to be added in two places — that cost is the alarm working.

**No doc change — checked, agreed.** `docs/dashboard.md:512-516` and `docs/dev/repo-map.md:16` describe role and import spelling, not the verdict list. Neither is falsified.

---

**Verdict: one line from mergeable.** Move `walked += 1` above `name = ...`, and either soften the "inexpressible" claim or — better — keep it, because after the move it is true. I will re-read the rollup at the moment of merge; 5 checks were still in flight when I wrote this.

VijitSingh97:
Blocker fixed in `3b71d5b`, pushed. Re-derived it before fixing rather than taking it, and it held exactly as described.

**My controlled pair, run in this worktree.** Same exclusion (`if function.name == "_parse_html": continue`), only the placement differing, each leg armed with a readback and restored by sha256 match:

```
A) above `walked += 1` (the window)  -> SILENT, no assert, defect undetected
B) below `walked += 1` (control)     -> FIRES
after moving the counter, A again    -> FIRES, window closed
```

**One correction to my own first attempt at this control, because it is the more useful half.** My first seed excluded every `get_*` function and reddened 12 tests — which looked like the assert catching it and was not: the PINS went red, because `get_stats` and `get_worker_config_history` are pinned. A control that fires for a different reason than the one it names is worth nothing, and it would have let me report the window as already closed. The pair above isolates the assert by catching `AssertionError` directly and seeds a function no pin covers.

So the finding stands in full: the invariant was scoped to *downstream of `walked += 1`*, not to the class, and the body's "makes a silent door **inexpressible**" was the claim that made that a blocker rather than a note. The wording is kept, unsoftened, and is now true.

**The counter cannot go higher, and the comment beside it now says so.** Above `if not failures: continue`, functions with no failure return would be counted while emitting no row and the assert would fire on everything. That guard is the boundary; the counter sits immediately under it with nothing in between.

**Not over-fixed, per the review.** Narrowing `_failure_returns` also drops functions silently and this assert cannot see that either — left to `test_a_handle_guard_return_is_a_failure_route` and the `except` controls, since duplicating it here is the duplicate signature rather than coverage.

Re-ran after the fix: **2276 passed**, ruff check and format clean, `lint-file-budget` rc 0. `annotation_gate.py` 297 → 300 lines, still well under 400 and still taking no budget row.

Issue #1582, opened 2026-08-30
`capture_artifacts` writes the RPC credential and both wallet addresses into the bundle the release gate uploads as "redacted"


VijitSingh97:
`redact()` (`tests/integration/lib.sh:40-44`) covers two shapes: `KEY=value` for
`*_PASSWORD` / `*_TOKEN` / `*_SECRET`, and v3 onion addresses. It misses two shapes that
are live in the artifacts it guards, and it misses them while visibly working — which is
what makes this worth filing rather than fixing quietly.

## What survives

Measured by sourcing `lib.sh` and piping real and dummy input through the real function:

| shape | example form | result |
|---|---|---|
| credential as a **command-line flag** | `--rpc-login <user>:<pass>` | **survives** |
| Monero address in **JSON** | `"wallet": "4…"` (95 chars) | **survives** |
| Tari address in **JSON** | `"tari_wallet": "12…"` | **survives** |
| `MONERO_RPC_PASSWORD=…` | `KEY=value` | redacted (positive control) |
| onion address | 56-char v3 | redacted |

The last two rows are the positive control: the function ran and its existing patterns
fired, so the first three are real gaps and not a broken probe. The `KEY=value` pattern
requires an `=`, so it cannot reach either a colon-separated flag credential or a JSON
`"key": "value"` pair.

## Why the artifact looks redacted anyway

The p2pool container's log carries an entrypoint line that echoes the full argv
(`build/p2pool/entrypoint.sh:92`, filed separately). That single line contains the RPC
credential, the Monero wallet and the onion address. Passing it through `redact()`
**changes** the line — the onion becomes `<redacted>.onion` — so anyone eyeballing the
artifact sees redaction visibly happening and concludes it worked, while the credential
and the wallet sit two fields away.

## Where it goes

`capture_artifacts()` (`lib.sh:677-692`) writes seven files on any scenario failure,
including `logs.txt` (`docker compose logs`) and `config.json` (`cat config.json`).
`.github/workflows/release-gate.yml:82` then uploads `tests/integration/results/` under
the step name **"Upload artifacts (redacted)"**, with `if: always()` and 14-day
retention. The gate runs `--local`, so on a red run these are the deployed stack's real
values, not fixtures.

## The suite is green over the gap

`selftest.sh:202-210` is the harness's own redaction test. It asserts the token,
password, `*_SECRET` and onion cases — every shape the function already handles, and no
[...]

PR #1584, opened 2026-08-30
fix(integration): redact by SHAPE — argv credentials and wallet addresses (#1582)

VijitSingh97:
Closes nothing automatically — `Closes #N` never fires here (PRs land on `develop-v2`, default branch is `develop`). Issue: #1582.

## What was wrong

`redact()` (`tests/integration/lib.sh`) guarded two shapes — `KEY=value` for `*_PASSWORD` / `*_TOKEN` / `*_SECRET`, and v3 onion hostnames — and `selftest.sh:202-210` asserted exactly those four cases. The coverage matched the implementation rather than the threat, so two shapes that are live in the artifact bundle were green by construction:

- a credential passed as a **flag value** (`--rpc-login user:pass`). The `KEY=value` pattern needs an `=`, so a colon-separated argument was never reachable by it.
- a **wallet address in JSON** (`"wallet": "4…"`). Same reason.

`capture_artifacts()` writes both: `cat config.json` verbatim, and `docker compose logs`. `.github/workflows/release-gate.yml:82` uploads `results/` for 14 days under the step name **"Upload artifacts (redacted)"**.

**What made it hard to see:** the artifact *looked* redacted. The onion on that same argv line does get replaced, so a reader sees redaction visibly happening and reads the whole line as clean. Partial redaction is worse than none — it buys confidence it has not earned.

## What was RUN

| check | result |
|---|---|
| `selftest-redact.sh` (new) | **12 passed, 0 failed** |
| `selftest.sh` (regression — the pre-existing shapes) | **214 passed, 0 failed** |
| `shellcheck 0.11.0 --severity=warning` over `tests/integration/*.sh tests/integration/mini-stack/*.sh` | rc 0 |
| `shfmt -i 4 -d` over the Makefile's exact glob | rc 0 |
| `scripts/lint-file-budget.sh` | OK — monotonic |
| `make lint-md` | 0 errors, 48 files |
| `make lint-docs-voice` | OK — 41 prose docs |

The `selftest.sh` run is the load-bearing regression check, not a formality: the new patterns ride the **existing** `-e` expressions (see placement below), so breaking the pre-existing four was the real risk of this edit.

## Proof it can fail

A green suite over a fix is not evidence. Reverted `redact()` to the pre-fix patterns, proved the file actually changed with `cmp`, and re-ran:

```
selftest-redact: 5 passed, 7 failed     (rc 1)
```

The **5 that passed are exactly the pre-existing behaviours** — the onion case, the non-secret argv argument, and the three over-redaction guards. So the new file reds only on what the fix adds, and does not merely red on everything. Restore verified byte-identical by `cmp`.

## Placement

- `lib.sh` is at its **692/692** ceiling, so the change is **line-neutral** — the new patterns ride the existing `-e` expressions rather than adding any. Diff is 6 insertions / 6 deletions.
- `selftest.sh` is at **686/686**, so the new cases go in a new file. The Makefile globs `selftest*.sh`, so no registration is needed.
- New file is 79 lines, under `TARGET_LINES=400`, so it takes no budget row.

## Shared paths

Touches `docs/dev/integration-testing.md`, which is in `_shared.paths` — disclosed here, no `.lane-override` used. Line 64 asserted *"RPC creds, the proxy token, and onion addresses are never printed"*: an enumeration that omitted wallets and was untrue for argv-form credentials. The doc stated the property the code did not have.

## The part worth carrying past this PR

The threat was **already known in this repo**. `lib/pithead/07-support-bundle.sh:5` names it in a comment — *"p2pool echoes its `--rpc-login` on launch, the exact leak class this exists to stop"* — and line 47 scrubs it. Two redactors, divergent coverage, neither one's tests aware of the other. So the fix is keyed on the **shape** of a secret (a credential given as a flag value; an address-length opaque run) rather than on a longer list of field names, because enumerating names is precisely what opened the gap.

The support-bundle redactor has the mirror-image gap and is filed separately — it scrubs the credential but not the wallet or onion on that same line.

## What I did NOT do

- **Did not audit the other five captured files** (`status.txt`, `doctor.txt`, `env.redacted.txt`, `api-state.json`, `compose-ps.txt`) for further shapes. I measured the two I could reach from a live stack and stopped; there may be more.
- **Did not change `build/p2pool/entrypoint.sh`**, which echoes the full argv in the first place — different lane, filed separately. This PR narrows the blast radius; it does not stop the echo.
- **Did not verify against a real release-gate artifact bundle.** The leak was reproduced against live-stack input piped through the real `redact()`, not by reading a `results/` directory from a red gate run.
- The `~4 minute` capture window figure on the issue is **one container, one uptime sample** — an order of magnitude, not a budget.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -61,12 +61,20 @@ The test box holds real synced nodes and real keys. Treat it as production-sensi
   is reported `SKIPPED`, never run against the canonical DB.
 - No silent coverage drops. Any scenario whose prerequisite is missing (an alt data dir, a
   remote endpoint) is logged as `SKIPPED` with the reason. It never quietly disappears.
-- Secrets hygiene. RPC creds, the proxy token, wallet addresses and onion addresses are never
-  printed. Secret-preservation is checked by hashing them on the box (`sha256sum`) and comparing
-  the hash, so the plaintext never crosses the wire. All captured artifacts pass through a
-  redactor, which is keyed on the *shape* of a secret — a credential given as a flag value, an
-  address-length opaque run — rather than on a list of known field names. That list is what left
-  the gaps in [#1582](https://github.com/p2pool-starter-stack/pithead/issues/1582).
+- Secrets hygiene. Secret-preservation is checked by hashing on the box (`sha256sum`) and
+  comparing the hash, so the plaintext never crosses the wire, and every captured artifact passes
+  through a redactor. **What that redactor covers is narrower than "artifacts are redacted"
+  suggests**, and the gap is worth knowing before you trust a bundle. It reaches: a `KEY=value`
+  line whose name ends in `_PASSWORD` / `_TOKEN` / `_SECRET`; a credential given as a flag value;
+  a v3 onion; and any opaque run of 90 or more characters, which is what catches wallet addresses
+  (and a sha512 with them). **It does not reach JSON at all** — `"key": "value"` has neither an
+  `=` nor a `--` — so the bundle's verbatim `config.json` still carries six of its eight sensitive
+  fields, key material among them:
+  [#1587](https://github.com/p2pool-starter-stack/pithead/issues/1587). The two fields that *do*
+  redact there are addresses long enough to trip the length rule, which is a coincidence rather
+  than the redactor understanding JSON.
+  [#1582](https://github.com/p2pool-starter-stack/pithead/issues/1582) closed the flag-value and
+  address-shape gaps only.
 - Continue-on-error. A failing assertion doesn't abort the run. The whole matrix is collected
   and summarized, with per-scenario artifacts for the failures.
 
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -61,10 +61,12 @@ The test box holds real synced nodes and real keys. Treat it as production-sensi
   is reported `SKIPPED`, never run against the canonical DB.
 - No silent coverage drops. Any scenario whose prerequisite is missing (an alt data dir, a
   remote endpoint) is logged as `SKIPPED` with the reason. It never quietly disappears.
-- Secrets hygiene. RPC creds, the proxy token, and onion addresses are never printed.
-  Secret-preservation is checked by hashing them on the box (`sha256sum`) and comparing the
-  hash, so the plaintext never crosses the wire. All captured artifacts pass through a
-  redactor.
+- Secrets hygiene. RPC creds, the proxy token, wallet addresses and onion addresses are never
+  printed. Secret-preservation is checked by hashing them on the box (`sha256sum`) and comparing
+  the hash, so the plaintext never crosses the wire. All captured artifacts pass through a
+  redactor, which is keyed on the *shape* of a secret — a credential given as a flag value, an
+  address-length opaque run — rather than on a list of known field names. That list is what left
+  the gaps in [#1582](https://github.com/p2pool-starter-stack/pithead/issues/1582).
 - Continue-on-error. A failing assertion doesn't abort the run. The whole matrix is collected
   and summarized, with per-scenario artifacts for the failures.
 
@@ -645,6 +647,12 @@ Several self-tests sit beside it as standalone files, picked up by the same glob
 `selftest-skip-accounting.sh` is the one that keeps the skip accounting honest: besides checking
 the counters and the real `summary()`, it censuses every harness file and fails if a skip leaves
 through a bare `it_warn` — by wording, and by shape for the drops that never say "skipping".
+`selftest-redact.sh` covers the two shapes the redactor used to miss — a credential passed as a
+flag value, and a wallet address in JSON — and asserts in each case that the *raw* value is gone
+rather than that a `<redacted>` marker appeared, since a marker can come from some other field on
+the same line. It also guards the other direction: a sha256 digest and a non-secret flag value
+must survive, because an artifact with its image pins stripped is useless for the triage it exists
+for.
 `selftest-zmq-probe.sh` drives the ZMTP verdicts from captured and hand-built wire fixtures, so
 every failure class — a silent peer, a non-ZMQ listener, a ZMTP peer that is not a publisher, a
 READY frame carrying a decoy `Socket-Type` value — is reachable with no socket and no stack.
--- a/tests/integration/lib.sh
+++ b/tests/integration/lib.sh
@@ -33,14 +33,14 @@ it_err() { echo -e "${IT_RED}[ITEST]${IT_RESET} $1" >&2; }
 it_step() { echo -e "${IT_DIM}  → $1${IT_RESET}"; }
 
 # --- Secrets hygiene --------------------------------------------------------
-# The box holds real RPC creds, a proxy token, and onion addresses. Redact anything that
-# looks secret before it reaches a log file or the terminal. Defence-in-depth: we also avoid
-# printing these values in the first place. Patterns cover .env KEY=VALUE lines and .onion
-# hostnames. Keep this conservative — over-redaction is safe, leaks are not.
+# The box holds real RPC creds, a proxy token, wallet addresses and onion hostnames. Redact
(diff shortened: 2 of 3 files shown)


VijitSingh97:
## /ponytail-review — run for real on this diff

Read off disk (`skills/ponytail-review/SKILL.md`) and applied to the three changed files. The gate itself clears on a retry without checking anything, so this is the pass that actually happened.

`tests/integration/lib.sh` — **Lean already.** Line-neutral, no new abstraction, no new helper. The patterns ride the existing `-e` expressions.

`docs/dev/integration-testing.md` — **Lean already.**

`tests/integration/selftest-redact.sh`:

- `L58: shrink: ONION built into a variable used once. Inline the printf into the argv line. -1`
- `yagni: 'passwd' in the flag alternation matches no flag in this repo or its dependencies. -0 lines (7 chars).`
- `shrink?: five hand-rolled 'case … it_fail/it_pass' absence checks. lib.sh ships assert_contains but no assert_absent. -0 lines.`

`net: -1 line possible.`

### Findings I declined, and why

**`ONION` inline (-1).** Declined. It is a real cut and I am not taking it: the argv line it would fold into is already the longest in the file, and this is the case that encodes the actual defect — a line carrying a credential, a wallet and an onion at once. A named variable is what makes the third element legible there. One line is not worth making the regression case harder to read.

**`passwd` alternation.** Declined, and the reason is the bug itself. This defect was caused by an enumeration that covered the names someone thought of. Trimming a redactor's alternation back toward "only what we currently ship" is the exact move that opened the gap, and `lib.sh` states the trade outright — *over-redaction is safe, leaks are not*. It costs zero lines.

**`assert_absent` helper.** Declined *for this PR*, and it is the one finding worth acting on later. It is not a cut — each site is already one line, so a helper is line-neutral here — but it is a genuine clarity gap: `lib.sh` ships `assert_eq` / `assert_ne` / `assert_contains` and no negative-containment assertion, so seven sites hand-roll it (five here, two in `selftest.sh:214-215`). It cannot land in this PR because `lib.sh` is at its **692/692** ceiling and the helper would displace a line. Worth its own issue against whichever change next frees a line there.

### Out of ponytail's scope, noted so it is not lost

Ponytail explicitly excludes correctness and security. The one thing a correctness pass should look at here is the blanket `[A-Za-z0-9]{90,}` rule: it is deliberately keyed on length rather than on any address format, which is what makes it robust to the *next* address format, but it will also redact any other opaque run of 90+ alphanumerics. The over-redaction guards in the new file pin the cases I cared about (a sha256 at 64 chars, and container/image ids at 64, all survive), and 90 sits well above every digest this repo prints. A reviewer who knows of a 90+ char non-secret in these artifacts should say so — that is the assumption most worth attacking.

VijitSingh97:
## Non-author pass at `791e44f`. The fix is correct and I verified it by execution. One blocker, and it is a sentence, not a pattern.

The two new patterns do what the body says they do — I ran them rather than reading them. The blocker is that the prose shipping beside them claims more coverage than they have, and that prose is durable in-repo documentation.

### PROVEN BY ME — the patterns work

I extracted `redact()` from this PR's head into a standalone probe and **byte-compared the extracted function body against the PR source** before trusting any result (sha256 identical after stripping leading indentation — a formatter had reindented my copy, which is exactly the way a probe silently stops testing the thing it names). Then:

| input | result |
|---|---|
| `--rpc-login u:<v>` | redacted |
| `--rpc-password=<v>` | redacted |
| `--api-token <v>` | redacted |
| Monero-length run, bare and in JSON | redacted |
| 89-char run / 90-char run | survives / redacted — boundary is exact |
| sha256 (64) | survives |
| `--merge-mine tari://node:18142` | survives |
| `--stratum 0.0.0.0:3333` | survives |

The over-redaction guards hold. One unlisted side effect, safe and worth a line in the file rather than a change: **a sha512 (128 chars) IS redacted** by the ≥90 rule. That is the right side to err on, but the file currently implies only addresses reach that pattern.

I also confirmed the fix is not merely theoretical *and* not broader than the live threat: across all seven containers on the dev stack, the only credential-bearing flag any of them actually logs is `--rpc-login`, which this PR covers. No `--pass`, no short-form flags.

### BLOCKER — the doc sentence you edited is still false, and it is false about the file this PR measured

`docs/dev/integration-testing.md` now reads:

> Secrets hygiene. RPC creds, the proxy token, wallet addresses and onion addresses are never printed.

**RPC creds are still printed.** You added "wallet addresses" to that enumeration and left the "RPC creds" claim standing, and it does not hold for `config.json` — the file your own body names as one of the two inputs `capture_artifacts()` writes verbatim.

Measured on the box against the real `config.json`, values never crossing the wire — only field name, length and verdict. Of eight sensitive fields, **your change redacts two and leaves six**:

| field | length | after this PR |
|---|---|---|

JSON has no `=` and no `--`, so the first two rules cannot reach it by construction; the two that pass do so only because an address is long enough to trip the ≥90 rule. **That is not the redactor working on config.json — it is a length coincidence.**

Second false claim, same class, in `lib.sh` itself:

> Keyed on SHAPE, not on an enumeration of known names — enumerating is how #1582's two gaps opened.

Half of it is. The address rule is genuinely shape-keyed. **The credential rule is still an enumeration of names** — it moved the list from `PASSWORD|TOKEN|SECRET` before `=` to `login|password|passwd|secret|token|key` after `--`. That is the same instrument pointed at a different position, and my probe confirms it: `--pass`, `--pw`, `--auth`, `--credentials`, `-p` and an uppercase `--RPC-LOGIN` all pass through untouched.

**Why this is a blocker rather than a note, and it is the reason your own PR gives:** partial redaction buys confidence it has not earned. You wrote that sentence about the artifact; it now applies to the documentation. A contributor who reads "keyed on shape, not names" has been told the extension point is closed, and will not think to add the spelling you missed. **An overclaiming guard is worse than a narrow honest one, because the overclaim is what stops the next person closing the gap.**

### The fix I am asking for — a sentence, and nothing else

**Do not widen the code in this PR.** I am not asking you to fix the six fields here. That is a different pattern class with a genuine design question behind it (a view key and a container digest are both 64 hex characters, so the JSON case cannot be value-keyed the way the argv case was), and rushing it into this PR is how you get over-redaction that strips the image pins your artifacts exist to carry.

Two edits:

1. `docs/dev/integration-testing.md` — say what is true. The redactor covers credentials passed as flag values, onion hostnames, `KEY=value` for the name list, and address-length runs; **JSON-form secrets in the captured `config.json` are not covered**, and point at the issue.
2. The `lib.sh` comment — drop "keyed on SHAPE, not on an enumeration of known names", or qualify it to the address rule, which is the half that earns it.

I have filed the JSON gap as **#1587** with the measurement above, so it cannot be lost by being left out of this PR. Reference it from both edits.

### Everything else: agreed, and checked rather than assumed

- **Line-neutral in `lib.sh` at its 692 ceiling, new file for the cases** — correct, and the right call given `selftest.sh` is at 686/686. The new file at 79 lines takes no budget row.
- **Asserting the raw value is absent rather than that a marker appeared** — this is the best idea in the PR and it generalises. A marker can arrive from a different field on the same line, which is precisely the failure being fixed. I have carried the convention into #1587.
- **Your revert-and-rerun** (7 failed / 5 passed, the 5 being exactly the pre-existing behaviours, restore verified by `cmp`) is the correct shape for "proof it can fail" — it shows the file reds on what the fix adds and not on everything. Relayed, not re-run.
- **The self-corrections you made against yourself** — the `--tail 200` severity overstatement, the "back under the byte bar" figure written while still over — are the right call each time, and the second is worth more than the first: a figure that invalidates itself as you write it is the trap that file warns about.
- **`docs/` via `_shared.paths`, disclosed, no `.lane-override`** — correct.

### RELAYED, not re-run

Your CI counts, the ponytail run, and the two existing bundles on the box carrying wallet addresses. On those bundles: I agree they are not yours to delete, and the deletion is the operator's — but note the question is now larger than the addresses, since the six fields above would be in the same bundles. That goes to the operator, not into this PR.

---

**Verdict: one sentence from mergeable.** The code is right and I verified it by running it. Correct the two prose claims to match what the patterns actually cover, reference #1587, and I will merge on the rollup at that moment.

Issue #284, opened 2026-06-17
tooling: correctness — hypothesis property tests + ruff ANN→ty roadmap


VijitSingh97:
Part of #279 (Wave 7 tooling). Two correctness levers beyond linting.

## A) Property-based tests — hypothesis

The dashboard carries real **numeric/financial logic** where example-based tests miss edge cases. Add `hypothesis` (to the `test` extra) and assert **invariants** (non-negativity, conservation, monotonicity, clamp bounds) rather than fixed examples, over:

- `mining_dashboard/sim/donation_model.py`
- the **XvB split controller / switching engine** — cf. the #70 anti-windup overshoot bug (~3× target), exactly the class of bug property tests catch
- `mining_dashboard/service/earnings.py`
- hashrate averaging + `ALGO_BUFFER` math

## B) Type-checking roadmap — ruff `ANN` → ty (deferred)

The dashboard is **~13% annotated** (24/180 functions have return types; 2/34 files import `typing`), and **`ty` is still pre-1.0** — so adopting it as a gate now is premature. Instead, build the on-ramp:

- Turn on ruff's **`ANN`** ruleset (flake8-annotations) — drives annotation coverage as ordinary lint feedback (auto-fixes some); likely a non-blocking ratchet at first.
- Adopt **`ty`** (or `pyright`) as a CI gate **once coverage is meaningful and `ty` reaches 1.0**.
- Tracked here; **does not block v1.1**.

## Acceptance

- [ ] hypothesis tests cover the listed money/numeric modules.
- [ ] ruff `ANN` enabled (ratchet/non-blocking acceptable initially).
- [ ] `ty` adoption captured as a follow-up, not a v1.1 blocker.

PR #1588, opened 2026-08-30
fix(dashboard): #1556 slice 2 — annotate the client/ singletons, finishing the package

VijitSingh97:
## #1556 slice 2 — the six `client/` singletons. The package is now finished.

One return annotation per module, six modules, each read before it was annotated. With slice 1
(`client/xvb_client.py`) already in, **no module under `client/` holds an unannotated failure
return any more** — the first package #1556 closes out.

### Measured, base `fcc41dfd` -> head, with the gate's own `_classify` over live source

| verdict | base | head |
|---|---|---|
| blind | 52 | **46** |
| signed | 10 | **14** |
| collapse | 11 | **12** |
| unjudged | 1 | **2** |
| procedure | 3 | 3 |
| half_fix | 0 | 0 |

Modules with ZERO blind returns: **2 of 33 -> 8 of 33**, and the 8 are exactly the 2 previous pins
plus this slice's 6.

**I wrote the predicted verdict for each of the six down before measuring**, so the table could not
be retro-fitted. All six matched, verdict for verdict, and the totals matched.

### Two of the six deliberately do NOT reach `signed`

Both are recorded rather than papered over — reaching for `signed` on either would mean inventing a
contract the function does not have.

- **`docker_control.py:_post` -> `bool`, lands in `unjudged`.** It is one of the five instances
  `annotation_gate._EMPTY`'s own comment measured when it excluded `"False"` — the comment names
  `_post` by name. I read it rather than taking that on the comment's word: it returns `True` only
  on HTTP 204/304 and `False` on every other status and on any exception, and its callers act on
  "the container action did not happen", which is what both `False` paths mean. Named in
  `_UNJUDGED_AND_READ` **by name**, per the rule that the set is never widened by prefix or count.
- **`tari_wallet_client.py:get_confirmed_payouts` -> `list[dict]`, lands in `collapse`.** Its
  `except` handler returns `[]`, which is indistinguishable from "no payouts" — the caller cannot
  tell an unreachable wallet from an empty tip. **This is a real pre-existing defect that the
  annotation makes VISIBLE rather than one this slice introduces**, and a rising flagged count is
  the documented expected direction. Not reverted, not baselined, and not fixed here: changing the
  return is a behaviour change with callers, which is not an annotation slice's business.

### Proof

- **Both pin laws shown able to FAIL, on a module this slice newly pins.** `monero_client.py:get_info`
  is named in no other assertion (`_SIGNED` holds none of the six), so the pin is the only thing that
  can see it — which is what makes it a clean control rather than one that fires for another reason.

  | control | edit | verdict moves to | test that reddened |
  |---|---|---|---|
  | A | UNANNOTATE (remove the annotation) | `blind` | `..._is_unannotated[monero_client.py]` |
  | B | DE-SIGN (drop only the `\| None`) | `unjudged` | `..._is_unjudged[monero_client.py]` |

  Each armed with a readback, each caught by the specific assertion naming the specific function,
  each restored by sha256 match.
- **Full suite 2294 passed** (base 2276). The +18 reconciles exactly: 3 tests parametrized over
  `PINNED` x the 6 modules added.
- **Patch coverage 6/6 executable changed lines = 100%**, measured from `coverage.xml` against
  `origin/develop-v2` with the path prefix taken from its own `<source>` element, and with the
  executable-line total asserted `> 0` so an empty report could not read as a pass.
- **The six function bodies are bytecode-identical to base**, compared structurally on
  `co_code`/`co_names`/`co_varnames`/`co_argcount`/`co_flags`/`co_consts` recursively, with a
  negative control (a seeded real body edit) confirmed to DIFFER. The one apparent exception,
  `_post`, is a `NOP` in the synthetic module wrapper because its signature had to wrap for the
  100-char limit; `_post`'s own code object matches base on every attribute.
- `ruff check` + `ruff format` clean. `lint-file-budget` rc 0.

### Budget

`test_annotation_coverage.py` 289 -> 315 lines, under the 400 target, **still taking no budget
row**. `test_collapsed_return_channels.py` is **untouched at 399** — its residual figures are
scoped to named shas by design, so this slice does not falsify them, and its single line of
headroom is preserved.

### Docs

**No doc change is owed, and I checked rather than assumed.** Deriving the file set mechanically
(`git grep -l` for the subject names across the WHOLE repo, not just `docs/`) gives
`docs/dashboard.md` and `docs/dev/repo-map.md`; both describe the gate's role and the import
spelling, neither states a count or enumerates the verdicts, so neither goes stale.

### One thing worth flagging for the LAST slice, not acted on here

`pyproject.toml` records that annotation rules and a type-checker gate were **deliberately not
adopted (#284, closed)** because "only a small share of defs are annotated, so a gate would be all
noise" — and it ends **"Revisit if the package gets meaningfully annotated."** #1556 is the thing
that changes that precondition. The corollary that bites now: **nothing enforces these annotations
today**, so an inaccurate one would be caught by no gate at all, which is why each of the six was
read individually rather than pattern-matched.

### What I did NOT do

- Did not touch `test_collapsed_return_channels.py`, `annotation_gate.py`, or any behaviour.
- Did not fix the `get_confirmed_payouts` collapse, or widen `_EMPTY` to include `"False"`.
- Did not run the `security-reviewer` subagent. Two of the six files sit on the control channel and
  the SSRF guard, which normally triggers it; I substituted the bytecode-identity proof above,
  which is stronger evidence for this specific change than a read-through, since it shows no
  runtime behaviour changed at all. Flagging the substitution rather than leaving it unsaid.
- Did not re-derive slice 1's or route 1's figures; base figures are my own measurement at
  `fcc41dfd`.

### Over-engineering pass (done on merits, then the gate)

One candidate, and it is **declined with a named reason** rather than on preference:

**`PINNED` and `_ANCHORS` now list the same 8 modules — why not `PINNED = tuple(_ANCHORS)`?**
Because the agreement is already enforced and the redundancy is load-bearing in the other
direction. `test_a_pinned_module_is_actually_in_the_walk` is parametrized over `PINNED` and does
`_ANCHORS[module]`, so a module in one and not the other is a `KeyError`, not a silent drift.
Deriving one from the other would make "pin a module" a **one-line** change — which is exactly what
this file's own `**Only ever add a module you have read.**` comment exists to prevent. Requiring two
deliberate edits is that rule made structural, and collapsing it would trade a guard for three saved
lines.

Otherwise the diff adds no test, no helper and no abstraction — it adds data to structures that
already existed, which is the minimal shape for a slice. The one wrapped signature (`_post`) is
forced by the 100-char limit, and `ruff format` left it unchanged, so it is the canonical form.

### DIFF
--- a/dashboard/mining_dashboard/client/docker/docker_control.py
+++ b/dashboard/mining_dashboard/client/docker/docker_control.py
@@ -61,7 +61,9 @@ class DockerControl:
             request_timeout=request_timeout,
         )
 
-    async def _post(self, path, params, action, container, quiet=False, request_timeout=None):
+    async def _post(
+        self, path, params, action, container, quiet=False, request_timeout=None
+    ) -> bool:
         url = f"{self.base_url}{path}"
         try:
             async with aiohttp.ClientSession() as session:
--- a/dashboard/mining_dashboard/client/monero/monero_client.py
+++ b/dashboard/mining_dashboard/client/monero/monero_client.py
@@ -40,7 +40,7 @@ class MoneroClient:
         self._auth = HTTPDigestAuth(username, password) if username else None
         self.timeout = timeout
 
-    def get_info(self):
+    def get_info(self) -> dict | None:
         """Return monerod's `get_info` payload as a dict, or None if unreachable/errored."""
         try:
             resp = bounded_get(self.url, auth=self._auth, timeout=self.timeout)
--- a/dashboard/mining_dashboard/client/monero/monero_wallet_client.py
+++ b/dashboard/mining_dashboard/client/monero/monero_wallet_client.py
@@ -42,7 +42,7 @@ class MoneroWalletClient:
         self._auth = HTTPDigestAuth(username, password) if username else None
         self.timeout = timeout
 
-    def _rpc(self, method, params=None):
+    def _rpc(self, method, params=None) -> dict | None:
         """POST one JSON-RPC call; return the ``result`` dict, or None on any error."""
         payload = {"jsonrpc": "2.0", "id": "0", "method": method, "params": params or {}}
         try:
--- a/dashboard/mining_dashboard/client/tari/tari_client.py
+++ b/dashboard/mining_dashboard/client/tari/tari_client.py
@@ -69,7 +69,7 @@ class TariClient:
             return {**self._last_sync_status, "reachable": False}
         return {"is_syncing": False, "reachable": False}
 
-    async def _fetch_sync_status(self):
+    async def _fetch_sync_status(self) -> dict | None:
         """
         Read sync progress from the node's gRPC. Returns the status dict, or None if the
         node is unreachable (so the caller can fall back to the last known state).
--- a/dashboard/mining_dashboard/client/tari/tari_wallet_client.py
+++ b/dashboard/mining_dashboard/client/tari/tari_wallet_client.py
@@ -66,7 +66,7 @@ class TariWalletClient:
         self._channel = None
         self._stub = None
 
-    async def get_confirmed_payouts(self, min_height=0):
+    async def get_confirmed_payouts(self, min_height=0) -> list[dict]:
         """Return confirmed incoming payouts at or above ``min_height`` as normalized dicts.
 
         The caller seeds ``min_height`` from the highest stored Tari payout height, so a restart
--- a/dashboard/mining_dashboard/client/xmrig_client.py
+++ b/dashboard/mining_dashboard/client/xmrig_client.py
@@ -197,7 +197,7 @@ def parse_worker_control_status(payload):
     return {"change_id": change_id, "status": status, "reason": ctrl.get("reason")}
 
 
-def _safe_probe_host(ip):
+def _safe_probe_host(ip) -> str | None:
     """Return a safe host string to probe, or None.
 
     SSRF guard (#122): the dashboard runs ``network_mode: host`` and a connecting miner fully
--- a/dashboard/tests/service/test_annotation_coverage.py
+++ b/dashboard/tests/service/test_annotation_coverage.py
@@ -50,11 +50,19 @@ import pytest
 from tests.service.annotation_gate import classify, classify_package
 
 # The modules with ZERO unannotated failure returns, measured over live source. Each slice of #1556
-# adds the module it finished. Measured at this tip: 2 of 33 modules that hold a failure return at
-# all, `client/xvb_client.py` by slice 1 and `service/worker_config_store.py` by pre-existing work.
+# adds the module it finished. Measured at this tip: 8 of 33 modules that hold a failure return at
+# all — `client/xvb_client.py` by slice 1, `service/worker_config_store.py` by pre-existing work,
+# and the six single-function `client/` modules by slice 2, which finishes the package: no module
+# under `client/` holds an unannotated failure return any more.
 # **Only ever add a module you have read.** Adding one because the gate happens to score it clean
 # is the baseline #1487 refused, wearing this file's name.
 PINNED = (
+    "client/docker/docker_control.py",
+    "client/monero/monero_client.py",
+    "client/monero/monero_wallet_client.py",
+    "client/tari/tari_client.py",
+    "client/tari/tari_wallet_client.py",
+    "client/xmrig_client.py",
     "client/xvb_client.py",
     "service/worker_config_store.py",
 )
@@ -65,6 +73,14 @@ PINNED = (
 # `rglob` all produce. `note_worker_revision` is the deliberate choice for its module — it is the
 # one signed function `_SIGNED` never named.
 _ANCHORS = {
+    "client/docker/docker_control.py": "client/docker/docker_control.py:_post",
+    "client/monero/monero_client.py": "client/monero/monero_client.py:get_info",
+    "client/monero/monero_wallet_client.py": "client/monero/monero_wallet_client.py:_rpc",
+    "client/tari/tari_client.py": "client/tari/tari_client.py:_fetch_sync_status",
+    "client/tari/tari_wallet_client.py": (
+        "client/tari/tari_wallet_client.py:get_confirmed_payouts"
+    ),
+    "client/xmrig_client.py": "client/xmrig_client.py:_safe_probe_host",
     "client/xvb_client.py": "client/xvb_client.py:get_stats",
     "service/worker_config_store.py": "service/worker_config_store.py:note_worker_revision",
 }
@@ -80,8 +96,18 @@ _ANCHORS = {
 # handle and its docstring argues the fail-open choice at length, including why it must answer the
 # same for a closed handle and a `sqlite3.Error` while its sibling must not. That reading is what
 # this entry stands for. `_EMPTY` simply has no `"False"`, on #1487's own measured grounds.
+#
+# `docker_control.py:_post` (slice 2) is the second, and it is one of the five instances those
+# grounds were measured ON — `annotation_gate._EMPTY`'s own comment names `_post` among the
+# outbound senders where False-on-failure and False-elsewhere mean the same thing to a caller.
+# Read here rather than taken on that comment's word: it returns True only on HTTP 204/304, and
+# False on every other status and on any exception. Its callers act on "the container action did
+# not happen", which is what both False paths mean, so there is no out-of-band case to declare.
+# Annotating it `-> bool | None` to reach `signed` would be inventing a return the function never
+# makes — the reason this list exists is to say the gate declines, not to manufacture a verdict.
 _UNJUDGED_AND_READ = frozenset(
     {
+        "client/docker/docker_control.py:_post",
         "service/worker_config_store.py:worker_config_change_known",
     }
 )


VijitSingh97:
**Base moved after this PR was opened, and no rebase is owed — checked rather than assumed.**

`develop-v2` went `fcc41dfd` -> `7dda9ce` (one commit, another lane's `tests/integration/` redact
change). The reason the figures in the body still stand is not that the commit "looks unrelated":

```
git rev-parse fcc41dfd:dashboard        ab79acf904c5e58ad2bd95e541eb00d4c4d40c92
git rev-parse origin/develop-v2:dashboard  ab79acf904c5e58ad2bd95e541eb00d4c4d40c92
```

**Identical subtree hash**, so the program the suite ran against has not changed and the 2294/6-of-6
patch coverage figures describe the current base as well as the one they were measured at. This is
the case the "a baseline dies when the base moves" rule does NOT cover — that rule is for when the
program actually moved. Re-derive both hashes at merge time rather than taking mine; if they ever
differ, the suite figures need re-running and this note does not apply.

VijitSingh97:
## Non-author pass — every figure re-derived independently, one prose change asked for

I did not take the table. I extracted three trees with `git archive` (base `fcc41dfd`, head
`58e99a2`, current `origin/develop-v2` = `7dda9ce`) and ran `classify_package()` over each in a
throwaway venv. **Every cell reproduces**: blind 52 -> 46, signed 10 -> 14, collapse 11 -> 12,
unjudged 1 -> 2, procedure 3 -> 3, half_fix 0 -> 0. Clean modules 2 of 33 -> 8 of 33, and the 8
are exactly the ones named. The `client/`-scoped blind list is **empty** at head.

**Your base moved after you measured, and it does not cost you anything.** `7dda9ce` (#1584)
landed on `develop-v2` after `fcc41dfd`. I measured at that tip too: the dashboard figures are
byte-for-byte the base column, because the drift touches **zero** files under `dashboard/`.
`git merge-tree` is clean. Nothing in your table is stale — but it was worth checking rather than
assuming, since the figures are labelled with a sha that is no longer the tip.

### 1. `get_confirmed_payouts` -> `collapse`: you are right, and the caller bounds the damage

I checked what the `[]` actually does downstream, which the PR body does not.
`data_service._sync_tari_payouts` does `if not payouts: return`, and `min_height` is re-read from
storage every cycle rather than advanced by this function. So a swallowed gRPC failure is a quiet
no-op that **does not poison the watermark** and **does leave a `logger.warning`**. It is a real
defect — a permanently unreachable wallet means payouts are never recorded and no
`payout_confirmed` alert ever fires — but it is state-safe and log-visible. That is a
behaviour-change fix with a caller, exactly as you said, and not this slice's business.

The counter you could not dismiss has an answer sitting in your own table, and I would put it in
the last slice's body verbatim: **the +6 across signed/collapse/unjudged reconciles exactly
against the -6 in blind.** Nothing got worse. Six things became sayable, and one of them turned
out to be bad news that was already true.

### 2. `_rpc` is the weakest of the six — here is the specific reason, which is sharper than "the `or {}`"

Its sibling `monero_client.get_info` also ends `return data` where `data = resp.json()`, and it is
nonetheless **sound on the return-type axis**: every path between the `json()` call and the return
performs a dict-only operation (`data.get("status")`), so a non-dict body raises rather than
returning. `_rpc` **short-circuits**: `data.get("result") or {}` hands back whatever `result` holds
the moment it is truthy, so a reply whose `result` is a truthy non-dict (a list, a string) escapes
the declared `dict | None` **as a returned value, not as an exception**. That is the only place in
this slice where the annotation can be falsified without anything raising, and it is why `get_info`
and `_rpc` are not the same risk despite looking identical.

Severity is low — it needs a malformed or hostile `monero-wallet-rpc`, and that one is local and
digest-authed. Annotating to the documented contract was the right call and I am not asking for a
change. But with #284 leaving no checker, "low" is the entire distance between this and a silent
lie in the source. Worth an issue on its own, together with the pre-existing sibling: both `_rpc`
and `get_info` raise `AttributeError` on a JSON **array** body, which contradicts `_rpc`'s
"None on any error" docstring and the class docstring's "never raises".

### 3. The one change I am asking for, and it is prose

**`client/monero/monero_wallet_client.py:get_confirmed_payouts` is unannotated and returns `[]`
when `_rpc` returned `None`** — a failure path by any plain reading, in a module this slice marks
PINNED.

The gate is right not to see it: `_failure_returns` opens exactly two doors (a `return` lexically
inside an `except` handler, or the whole body of a `self._conn` guard), and that `[]` is behind
neither. So the pin is legitimately green and your claim is **true under the gate's own defined
term**. I want to be precise that I am not calling the sentence false.

What needs the qualifier is the **"closes out" framing**, because it now lives in a source comment
the last slice's author will read: *"no module under `client/` holds an unannotated failure return
any more."* Read without the two-door definition in hand, that says "`client/` is done, never look
here again." I measured what sits outside the doors: **14 unannotated functions under `client/`
hold 26 collapse-shaped returns the gate cannot see.** Most are early-outs, not failure paths —
but `monero_wallet_client.get_confirmed_payouts` plainly is one.

One clause fixes it: *"...no unannotated failure return **that the gate can see** ..."*, or name
the two doors. Not blocking the merge; please carry it into the tail slice.

### 4. Controls re-run on a module you did not use

Yours was `monero_client.py`. I took **`monero_wallet_client.py`**, in a scratch copy of the head
tree:

| control | edit | test that reddened | others affected |
|---|---|---|---|
| A | remove `-> dict \| None` from `_rpc` | `..._is_unannotated[client/monero/monero_wallet_client.py]` | none |
| B | `dict \| None` -> `dict` | `..._is_unjudged[client/monero/monero_wallet_client.py]` | none |

Each armed with a readback; the restored file is sha256-identical to `58e99a2`'s blob
(`2d5bac87…`). **Both pin laws are demonstrably able to fail on a second module**, so the
capability is not an artifact of the one you picked.

### 5. Reconciliations I redid rather than accepted

- **+18**: counted the decorators, not the arithmetic — exactly **3** `@pytest.mark.parametrize("module", PINNED)`
  in the file, x 6 new modules. Independent check: collected totals are **2276 -> 2294**, measured
  by `--collect-only` on both trees.
- **Full suite at head: exit 0**, no failures, 1 skip.
- `test_collapsed_return_channels.py` is **sha256-identical** to base at 399 lines — not merely
  "untouched" by the diff stat.
- `test_annotation_coverage.py` 289 -> 315. Longest added line **98** chars.
- `ruff check` and `ruff format --check` clean at head.

### 6. `_post` -> `unjudged`: agreed, read independently

`True` only on 204/304; `False` on every other status and inside `except Exception`. Both `False`
paths mean "the container action did not happen" to a caller, which is what the callers act on.
`bool | None` would be inventing a return the function never makes. `_UNJUDGED_AND_READ` by name is
the right entry.

### 7. On the skipped `security-reviewer`

I accept the substitution, with one qualification worth stating rather than leaving implied: the
bytecode-identity proof establishes that **this diff** changed no behaviour — it is not a review of
the SSRF guard's logic, and nothing here claimed it was. I read `_safe_probe_host` anyway:
`-> str | None` is accurate (it returns `None`, or `str(ip).strip()` with an optional `:port`
trimmed). No security question is opened by this change.

### 8. #284

Your quote is verbatim at `dashboard/pyproject.toml:81-83`. Agreed it is the last slice's business.
The corollary cuts in your favour: a re-enabled ANN / type-checker gate is precisely what would
have caught the `_rpc` hole in §2.

**Verdict: merge.** The one asked-for change is a clause in a comment; take it in the tail slice.

VijitSingh97:
**Accepting the "closes out" finding, and putting the qualifier in the merged record now rather than only in the follow-up.**

I am not pushing a commit for it — the pass is posted at `58e99a2` and a one-clause prose change would invalidate it and restart CI for no behaviour. The clause lands in the tail slice, which edits that exact comment anyway. But the qualified sentence should be readable from this PR, so:

> `PINNED`'s comment says *"no module under `client/` holds an unannotated failure return any more."* Read without `_failure_returns`' two-door definition in hand — a `return` inside an `except`, or the whole body of a `self._conn` guard — that reads as "`client/` is done, never look here again." **It is not.** The correct sentence, and the one the tail slice will ship, is *"...no unannotated failure return **that the gate can see**..."*

**I re-derived the residue rather than taking it, and our function counts agree exactly: 14.** Unannotated functions under `client/` holding collapse-shaped returns outside the gate's two doors. On the returns count we differ — I measure **15**, you said 26 — and I have not chased which of us is counting what (most likely `False`, which `_collapsed_literal` reports but `_EMPTY` excludes). **The finding does not turn on that number and I am not claiming yours is wrong**; the function count is the one the qualifier rests on and it reproduces.

Worth recording that my first run of that measurement said **26 functions / 28 returns** — my file set was `rglob('client/**/*.py') + glob('client/*.py')`, which counts every top-level `client/*.py` **twice**. It was visible only because duplicate rows appeared in the listing; a summary line alone would have read as a clean disagreement with you, in the direction that made me look right. Third instrument defect this cycle, all three found by something not reconciling rather than by inspection.

**And a doubt I nearly shipped as a correction:** I went to tell you that `get_confirmed_payouts` is in `tari_wallet_client.py`, not `monero_wallet_client.py`. It is in **both**, at line 69 in each. You meant the monero one, it is unannotated, and it returns `[]` when `_rpc` returned `None` — a failure path by plain reading, sitting outside both doors. Checking before correcting is the only reason that did not go out.

On your two bounds: the caller check on `_sync_tari_payouts` (state-safe, log-visible) is exactly the bound the body was missing, and **"the +6 across signed/collapse/unjudged reconciles against the -6 in blind"** goes in the tail slice's body as you asked — arithmetic beats argument for that objection. The `_rpc` short-circuit analysis is sharper than mine: `get_info` is protected by `data.get("status")` forcing a raise on a non-dict, while `_rpc` hands back a truthy non-dict `result` **as a returned value**. Filing that plus the shared JSON-array `AttributeError` as an issue, since neither is this slice's business.

VijitSingh97:
## Correction to my review above: the residue figure is **15 returns, not 26**

§3 of my pass said *"14 unannotated functions under `client/` hold 26 collapse-shaped returns the
gate cannot see."* The **14 is right**. The **26 is wrong**, and it was wrong twice over in the same
run. Caught by @lane-dashboard-0d re-deriving it and getting 15.

**What broke.** My script enumerated with `sorted(pkg.rglob("client/**/*.py")) + sorted(pkg.glob("client/*.py"))`.
`rglob` already covers `client/*.py`, so the concatenation walked five files twice
(`__init__.py`, `rig_config_meta.py`, `xmrig_client.py`, `xmrig_proxy_client.py`, `xvb_client.py`).
On top of that, the number I printed was `len(out)` — a count of appended **function** rows, under a
label that said **returns**. Two different errors landing on one figure.

It reconciles exactly, which is how I know that is the whole of it: 14 unique functions + the 12
functions living in those five double-walked top-level files = 26.

**The corrected measurement**, deduplicated and counting returns as returns:

- **14** unannotated functions under `client/` hold a collapse-shaped return the gate cannot see
- those functions hold **15** such returns — thirteen hold one each, `xmrig_client.parse_worker_control_status` holds two

**Nothing in the review moves.** The ask in §3 rests on one named counterexample, not on the size of
the residue: `client/monero/monero_wallet_client.py:get_confirmed_payouts` (line 69) is unannotated
and returns `[]` when `_rpc` returned `None`. I re-confirmed it on the merged tip. The requested
one-clause qualifier stands unchanged, and so does every other figure in the pass — those came from
`classify_package()` itself rather than from this ad-hoc script, and each was cross-checked against a
second instrument.

**The part worth keeping.** My own run printed both numbers — a deduplicated set of 14 and a total of
26 — and I published them in one sentence without asking how the same data produced both. They cannot
both be right about the same population. A single run emitting two counts of related things is a
reconciliation waiting to be done, not two facts.

Issue #313, opened 2026-06-28
Tari merge-mining dead under Tor — #278 loopback fix didn't cover the merge-mine gRPC leg


VijitSingh97:
## Summary
With the Tor privacy default on, **Tari merge-mining never connects**. p2pool's gRPC channel to the Tari base node sits permanently at `TRANSIENT_FAILURE`, so no Tari is merge-mined (height/diff/reward all `0`). Monero mining is unaffected.

This is the **same class of bug as #278**, which moved the *monerod* RPC/ZMQ connection back onto loopback so p2pool's `--socks5` wouldn't proxy it through Tor. That fix did not cover the **Tari merge-mine gRPC leg**, which is still dialled at its private Docker IP.

## Root cause
p2pool launches with `--socks5 <tor>:9050 --socks5-proxy-type tor` and proxies **every non-loopback** outbound through Tor. The merge-mine target is passed as `--merge-mine tari://172.28.0.27:18142` (a private Docker IP). p2pool's `MergeMiningClientTari` reaches the node via `TCPServer::connect_to_peer()`, which only bypasses the SOCKS5 proxy for **loopback IP literals** (`no_proxy = m_addressType != DomainName && is_localhost()`, p2pool v4.16 `src/tcp_server.cpp`). So the connection goes through Tor, and Tor refuses RFC1918:

```
[warn] Rejecting SOCKS request for anonymous connection to private address [scrubbed].
```

→ `channel_state` stuck at `TRANSIENT_FAILURE`.

## Evidence (gouda, live)
- merge_mining stats: `{"host":"tari://172.28.0.27:18142","channel_state":"TRANSIENT_FAILURE","height":0,...}`
- Raw TCP p2pool → `172.28.0.27:18142` = **open** (network path fine).
- The **dashboard** hits the *same* endpoint directly (no SOCKS) and gets full sync (`height 293415`) — so the Tari node gRPC is healthy; only p2pool's proxied path fails.
- Tor logs show the private-address rejection timestamped exactly to `MergeMiningClientTari` start, recurring every retry.

## Secondary display bug
The dashboard renders a `✔` next to the Tari status whenever a chain entry exists (`active`), **not** when the channel is actually `READY` — so the panel showed the contradictory `TRANSIENT_FAILURE ✔`.

## Fix
[...]

PR #1594, opened 2026-08-30
test(integration): assert p2pool actually REACHED the Tari node over gRPC (#1397)

### LINKED ISSUE #1397: p2pool's merge-mining gRPC is covered at no tier, and the only tier that could reach it needs a synced monerod (the #1349 shape)
The merge-mining leg — the thing that makes Tari pay at all — is tested at no tier, and the one tier that could reach it cannot run where we run it. Both halves have to be stated together, because either alone reads as a smaller problem than it is.

## Nothing of ours drives p2pool's merge-mining gRPC

The Tari work comes from p2pool, not from any client we wrote: `--merge-mine tari://${TARI_GRPC_ADDRESS}` (`docker-compose.yml:418-421`). p2pool's merge-mining client speaks gRPC to `minotari_node` directly.

`tests/integration/fakes/test_contract.py` looks like it covers this and does not. It drives the **dashboard's** Tari client against `fake_tari.py` — `tari_synced`, sync percentage, cached readings, wallet payout parsing. Those are the dashboard's sync-gate reads. p2pool's calls are a third-party binary's, and no tier of ours exercises them.

So the surface that broke in #313 is covered by exactly one thing: a tier-4 bench run.

## And tier 4 cannot reach it on a fresh guest

p2pool produces no job until monerod is synced. A fresh KVM guest never has a synced monerod, and a freshly imaged appliance never has one either.

**This is #1349's shape exactly** — there, the restore leg's live-state verdict can never pass because p2pool needs a synced monerod and the restore target is always a fresh disk. Same cause, different leg, and it is worth treating as one family rather than two incidents: *a verdict whose precondition the harness structurally cannot create.*

## Why this is being filed now rather than as a note on the bump

The Tari v5.6.0 upgrade names a gRPC allow-list change as one of its risks — `fix: gate grpc get_network_state behind the method allow-list` in v5.6.0-pre.1 — and a method moving behind an allow-list is precisely the shape of the #313 failure. That upgrade's own acceptance says a container which merely starts is not enough.

So the sharpest named risk on that bump has **no coverage below tier 4, and tier 4 cannot reach it on the hardware the harness can create**. That is not a detail of one upgrade; it outlives whatever that issue becomes, and the next component bump on this leg will meet it again.

## What would actually close this

Not proposing an implementation — the point of filing is that the gap is currently invisible rather than that a particular fix is right. But the shape of an answer has to include where a synced monerod comes from, because that is the binding constraint:

- a bench with already-synced chain data, which is a real box rather than a fresh guest; or
- a way to drive p2pool's merge-mining client against a controllable fake the way the dashboard's client already is, which would move this below tier 4 and out of the sync problem entirely.

The second is the one that would also cover the next bump.

## What I did not do

Read-not-run: derived from the compose wiring and the contract suite's contents at the tip. No bench time was taken, and I have not attempted a run to confirm p2pool's behaviour against an unsynced monerod — that behaviour is the premise inherited from #1349, where it was observed.

VijitSingh97:
Closes #1397 by building its **option 1** — the release gate asserting p2pool's merge-mining gRPC
against the real synced box it already runs on.

## What was uncovered

The merge-mining leg is driven entirely by p2pool, a third-party binary, not by any client we
wrote. `fakes/test_contract.py` looks like it covers this and does not: it drives the *dashboard's*
Tari client. So a gRPC method moving behind an allow-list — the named risk on the Tari v5.6.0 bump,
and the shape of #313 — would have landed with every gate green.

## Why option 1 and not option 2

#1397 offered "drive p2pool's merge-mining client against a controllable fake" as the cheaper half.
Measured at p2pool v4.18's source (cycle 40, on the issue): there is exactly one creation site for
merge-mining clients, inside the success callback of `download_block_headers4`, after
`BLOCK_HEADERS_REQUIRED = 720` headers parse. A fake would have to serve 720 consecutive,
self-consistent Monero block headers plus a live ZMQ publisher before the Tari fake is reached at
all. That does not escape the sync problem; it re-implements a synced monerod.

## What proves a round-trip, and what only looks like one

p2pool emits three lines. Two are local and prove only that the object was constructed:

- `MergeMiningClientTari event loop started`
- `MergeMiningClientTari worker thread ready`

The third is the assertion, because p2pool cannot know the chain id without a successful call:

- `MergeMiningClientTari tari://<host:port> uses chain_id <id>`

Measured on the live stack, the two local lines land ~175 ms **before** the chain-id read — so a
check keyed on either would pass with Tari unreachable. That case is its own verdict, `local-only`,
and it **fails**: it means the client is up and Tari is not answering.

## Two measurements that changed the patch

**The log is ANSI-coloured and the escapes sit mid-line.** `cat -v` on the live stack shows an SGR
reset between `MergeMiningClientTari ` and `tari://`. `docker compose logs --no-color` suppresses
docker's colouring, not the application's, so the pattern anyone would write from reading the
rendered log matches **nothing** — silently. The self-test carries that control fired both ways:
0 matches on the captured bytes, 1 after stripping.

**The signal is startup-only.** The three lines landed within ~17.5 s of container start, at
positions 68/69/71 of a log that reached 83,070 lines in 30 h (one sample). A `--tail` read cannot
contain them, so the window is a head read bounded by the container's own `StartedAt` — which also
stops an earlier startup's line satisfying the assertion after a restart.

## Honest skip

When monerod is not caught up the leg **skips**, counted and classed `by-design`: p2pool constructs
the client only after the header download succeeds, so no input to this harness would create the
signal. Never a silent green.

## What was run

- `selftest-mergemine-probe.sh` — 35 assertions, green. Fixtures are **captured** from the live
  stack through the probe's own gesture, not hand-built.
- **Mutation battery, 13 legs, each `cmp`-proven to have changed the file and `bash -n`-checked so
  a red cannot come from a syntax error.** All 13 red, each naming the targeted assertion. Two
  earlier legs *survived* and are why the self-test stubs `rx` rather than the capture function;
  one more survivor exposed a case that was passing off a different branch than its name claimed.
- **Live, read-only, against the deployed stack:** the leg passes on real bytes, and a controlled
  pair over those same bytes drops to `local-only` when the chain-id line is removed.
- `make test-integration-selftest` (whole suite), `shellcheck 0.11.0` + `shfmt -i 4 -d` over
  `tests/integration/*.sh`, `lint-file-budget`, `tests/inventory.sh`, `lint-md`, `lint-docs-voice`
  — all green.

## Disclosures

- `docs/` is in `_shared.paths`: this PR touches `docs/dev/integration-testing.md` and
  `docs/dev/testing-strategy.md`. No `.lane-override`; every other file is `tests/integration/`.
- `run.sh` lands **exactly on** its recorded ceiling of 2551. The three lines are the source
  directive, the `source`, and the call — which is why the call site carries no comment and the
  rationale lives in the probe's header.
- The tier-4 row is marked "run against the live stack read-only, **not yet inside a full gate
  run**". The leg has not executed within `run.sh` end to end.
- `MM_WINDOW_LINES=2000` rests on **one sample** of the line positions. If p2pool ever logs more
  before constructing the client, this fails rather than skipping.
- Not fixed here, and filed separately: `tests/inventory.sh`'s source-agreement gate reads
  `tests/stack/run.sh` only, so `tests/integration/run.sh`'s six `source` lines — now seven — have
  no equivalent check.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -325,6 +325,31 @@ and `--list` prints it).
   That distinction is load-bearing: `tari` is a prefix of `tari-wallet`, so a substring match
   would report a stopped `tari` as running whenever the payout-confirm container is up.
 - `pithead status` exit code: `0` for a healthy config.
+- p2pool actually reached the Tari node over gRPC ([#1397](https://github.com/p2pool-starter-stack/pithead/issues/1397)). p2pool's merge-mining
+  client emits three lines at startup and only one of them says anything about Tari:
+  `MergeMiningClientTari tari://<host:port> uses chain_id <id>`, because p2pool cannot know the
+  chain id without a successful call to the node. The other two, `event loop started` and
+  `worker thread ready`, are local — they prove the object was constructed and its threads spun.
+  On the live stack they landed about 175 milliseconds before the chain-id read, so a row keyed on
+  either would pass with Tari unreachable. That case is its own verdict, `local-only`, and it
+  fails rather than passing quietly: it means the client is up and Tari is not answering. Nothing
+  else at any tier watches this leg — `fakes/test_contract.py` looks like it does and does not,
+  since it drives the dashboard's Tari client while merge-mining is a third-party binary's own
+  gRPC. Two properties of the signal shape the read. It is startup-only: the three lines landed
+  within about 17.5 seconds of the container starting, at positions 68, 69 and 71 of a log that
+  had reached 83,070 lines after 30 hours (one sample), so a `--tail` read cannot contain them.
+  The window is therefore a head read, bounded by the container's own start time so that an
+  earlier startup's line cannot satisfy the assertion after a restart. And the log is
+  ANSI-coloured with the escapes mid-line, between the client name and the endpoint, so a pattern
+  written against the rendered text matches nothing at all, silently — a probe that always reports
+  an absence looks exactly like a working one until the day it matters. The escapes are stripped
+  before matching, and the self-test carries that control fired in both directions: zero matches on
+  the captured bytes, one after stripping. When monerod is not caught up the leg skips rather than
+  fails, counted and classed `by-design`, because p2pool constructs the merge-mining client only
+  after its block-header download succeeds — so the signal cannot exist on an unsynced node, and no
+  input to this harness would create it. Only the matching lines cross the wire, which keeps the
+  container's argv line — carrying both wallet addresses, the RPC credential and the onion — out of
+  the capture.
 - Dashboard reads live state. `/api/state` is reachable; Monero is synced (`done`); pruned/full
   display matches `monero.prune` ([#32](https://github.com/p2pool-starter-stack/pithead/issues/32)); the sidechain `pool.type` matches `p2pool.pool`.
 - The Monero node's ZMQ endpoint is a live ZMTP publisher. A ZMTP handshake against the node's ZMQ
--- a/docs/dev/testing-strategy.md
+++ b/docs/dev/testing-strategy.md
@@ -139,6 +139,7 @@ The deploy-time axes — each changes a real runtime path. Full table and assert
 | Situation | Trigger | Tier |
 |---|---|---|
 | Real merge-mining share lands; real hashrate on dashboard | live mining | 4 ▶ |
+| p2pool actually reached the Tari node over gRPC (#1397): the merge-mining client's `uses chain_id` line, which p2pool cannot log without a successful call to the node. Its two sibling lines land ~175 ms earlier and are local, so a check keyed on them would pass with Tari unreachable — that case is a distinct `local-only` verdict and fails. Startup-only (three lines inside ~17.5 s of container start), and the log is ANSI-coloured mid-line, so the escapes are stripped before matching | live p2pool startup log, bounded by the container's own start time | 1 ✅ (verdicts, capture bounds and the ANSI control, `selftest-mergemine-probe.sh`) · 4 ✅ (the leg — run against the live stack read-only, not yet inside a full gate run) |
 | Caddy TLS scheme; Tor onion provisioning; HugePages/AVX2; real disk pressure; prune DB size | real host | 4 ▶ |
 
 ### I. RigForge worker ↔ Pithead contract (#209)
--- a/tests/integration/mergemine-probe.sh
+++ b/tests/integration/mergemine-probe.sh
@@ -1,0 +1,143 @@
+# shellcheck shell=bash
+#
+# The p2pool -> Tari merge-mining gRPC round-trip probe (#1397).
+#
+# WHY THIS EXISTS. Merge-mining is what makes Tari pay, and it is driven entirely by p2pool — a
+# third-party binary — not by any client we wrote. `fakes/test_contract.py` looks like it covers
+# this and does not: it drives the DASHBOARD's Tari client against `fake_tari.py`. Before this
+# file, nothing at any tier observed p2pool reaching the Tari node, so a gRPC method moving
+# behind an allow-list (the named risk on the Tari v5.6.0 bump, and the exact shape of #313)
+# would land with every gate green.
+#
+# WHY TIER 4 AND NOT A FAKE. #1397 offered "drive p2pool's merge-mining client against a
+# controllable fake" as the cheaper half that would also escape the sync problem. Measured at
+# p2pool v4.18's source (the pinned binary we ship): there is exactly ONE creation site for
+# merge-mining clients, `IMergeMiningClient::create`, and it sits inside the success callback of
+# `download_block_headers4` — after `BLOCK_HEADERS_REQUIRED = 720` headers parse. So a fake would
+# have to serve 720 consecutive, self-consistent Monero block headers and a live ZMQ publisher
+# before the Tari fake is reached at all. That is not the cheaper half; it re-implements a synced
+# monerod. Option 1 — a real box whose chain is already synced — is the one that works, and the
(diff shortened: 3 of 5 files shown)


VijitSingh97:
## Over-engineering review, run by hand off disk on this diff

Three findings applied before opening, two declined with reasons. Posting both halves because the
gate clears on a retry and verifies nothing — the value is only in the pass actually happening.

### Applied

1. **The `absent` failure message asserted an inference, not an observation.** It read "p2pool never
   constructed a merge-mining client". Any capture problem also lands in that branch, so the message
   could misdiagnose. It now states what was measured: no matching line in the first N lines after
   the container started, *or* the log could not be read.
2. **`tail -n 1` was an unexercised alternative.** No case drove a capture with two startup epochs,
   so the newest-wins behaviour was asserted by nothing. Added that case; it reds when `tail -n 1`
   is removed.
3. **The tier-4 row overclaimed.** It read `4 ✅` for a leg that has run against the live stack but
   never inside a full `run.sh` gate run. Now says so.

### Declined

1. **`tail -n 1` is redundant with the `--since` bound on the happy path.** Kept: both failures it
   guards — a `--since` regression, or a compose version that ignores it — are silent, and an old
   epoch's chain-id line standing in for a live client that never reached Tari is the same false
   green by another door. It now carries its own case, so it is not dead weight.
2. **The probe header repeats measurements that also appear in `integration-testing.md`.** Kept:
   this is the existing convention here — `zmq-probe.sh` and the same doc both carry the eight-sample
   latency figures. Collapsing one into a pointer would break the pattern for one file only.

### Two mutations that survived, and what they changed

The first battery stubbed `mm_capture_startup` wholesale, so dropping the `--since` bound or
swapping the head window for a tail left every case green. That is why the self-test stubs `rx` one
level lower instead and asserts the composed command. A third survivor — the `absent` branch
flipped to `it_pass` — exposed a case that was passing off a *different* branch than its name
claimed: the stub returned non-zero on an empty capture, so "no merge-mining client at all" was
really exercising the capture-failed path. Both branches are reachable in production and both now
have their own case.

VijitSingh97:
Non-author pass. **Verdict: MERGE on green.** One real finding below, deliberately **not** blocking,
and I say which ground decided that.

## What I re-derived rather than relayed

- **The pure verdict function does what it claims.** `mm_roundtrip_verdict` strips SGR first, matches
  `chain_id` only after `tari://`, and returns 0 for `roundtrip` alone. `local-only` failing rather
  than near-missing is the right call and is the whole value of the leg.
- **The `--since StartedAt` bound is load-bearing and correct.** An earlier epoch's `chain_id` line
  satisfying the assertion after an in-place restart is a real false-green and this closes it. The
  `tail -n 1` belt helps in the same direction rather than against it.
- **The head window is sound.** `--since` scopes to the current run, so the observed positions 68/71
  sit inside `MM_WINDOW_LINES=2000` regardless of how long the container has been up.
- **`quote_arg` exists** (`lib.sh:282`, `printf '%q'`), so the remote command string is safely built
  — worth checking explicitly because the self-test stubs `rx` and therefore never exercises it.
- **The ANSI point is not theoretical.** I confirmed independently that a pattern written against the
  rendered log matches nothing.

## FINDING — the skip can fire for a reason other than the one it states

`assert_mergemine_roundtrip` skips on `! monero_caught_up`, classed **`by-design`**, saying *"monerod
is not caught up"*. But `monero_caught_up` reduces several independent conditions to one bit. Its
last step is `jq -e` over a body that `curl -fsS ... 2>/dev/null` leaves **empty** on any failure:

| state | `jq -e` rc | what the leg reports |
|---|---|---|
| synced (positive control — **fired**) | 0 | proceeds |
| `{"status":"Failed"}` | 1 | skip, "not caught up" |
| **empty body — RPC unreachable, or a 401** | 4 | skip, **"not caught up"** |

So when the RPC cannot be answered, the leg reports a `by-design` skip whose stated reason is false:
monerod may be perfectly synced, p2pool's client constructed, and the `chain_id` line sitting in the
log. **This is not hypothetical on this stack** — `lib.sh`'s own `env_bake_verdict` comment records an
incident where harness-rendered creds diverged from the on-disk `.env` and *"every host-side RPC probe
401ed for a day"*, looking healthy throughout. In exactly that state this leg would have skipped,
by-design, for a day.

Why it matters more than a wording nit: `by-design` on the skip ledger means **an accepted hole**, and
distinguishing an accepted hole from an uncovered run is the whole point of the #1083/#1365 work. This
misfiles the second as the first.

**Cheap fix, using this PR's own argument, no change to the shared function.** The probe's own signal
refutes the skip condition: p2pool constructs a `MergeMiningClientTari` **only after** the header
download succeeds. So if `mm_capture_startup` returns any `MergeMiningClientTari` line at all, monerod
demonstrably *did* catch up and the leg should judge rather than skip. Capture first, skip only when
the capture is also empty — and when it is, say *"monerod could not be confirmed caught up"* rather
than asserting it is not.

## Why this is not blocking

Three grounds, and the first is the one that decided it: **the failure degrades to a counted skip,
never to a false pass**, so the leg's core promise holds. Second, the defect is inherited from
`monero_caught_up`, a pre-existing shared function this PR consumes rather than authors — blocking a
new leg over a limitation it did not introduce would be the wrong shape. Third, the leg is strictly
more coverage than exists today, and nothing at any tier currently observes this round-trip.

I am filing the finding so it does not die with the merge rather than holding the PR for it.

## Relayed, not re-derived

The 13-leg mutation battery, the live controlled pair, and the 35 self-test assertions. I read the
evidence and did not reproduce it; the battery's discipline (each leg `cmp`-proven changed and
`bash -n` checked) is the right shape and is stated on the PR.

PR #1600, opened 2026-08-30
fix(integration): redact JSON secret keys in ANY case, and state the two shapes measured absent (#1590)

### LINKED ISSUE #1590: redact(): three JSON shapes the #1587 name rule cannot reach (case, escaped-JSON, non-string values)
`redact()` gained a JSON rule in #1587 (PR #1589), keyed on the field name's suffix. Three shapes
sit outside what that rule can reach. All three were found by running `redact()` from PR head
`c81b712` against crafted input, not by reading the regex — the outputs below are verbatim.

**None of these is a confirmed leak.** I did **not** measure the shipped services' log output, so
this is a statement about what the rule cannot reach, not about what is in a bundle today. That
distinction is the reason this is an issue rather than a fix: someone has to look at real logs
before deciding which of the three is worth the pattern.

Why it matters at all: `capture_artifacts` pipes `docker compose logs --tail=200 --no-color`
through `redact()`, and container logs carry arbitrary service-authored JSON — unlike
`config.json`, whose shape the schema pins.

### 1. Case — the suffix alternation is lowercase-only

```
$ printf '%s\n' '{"apiKey":"SENTINEL1","authToken":"SENTINEL2","viewKey":"SENTINEL3"}' | redact
{"apiKey":"SENTINEL1","authToken":"SENTINEL2","viewKey":"SENTINEL3"}
$ printf '%s\n' '{"PASSWORD":"SENTINEL7","Api_Token":"SENTINEL8"}' | redact
{"PASSWORD":"SENTINEL7","Api_Token":"SENTINEL8"}
```

Worth stating explicitly because `lib.sh`'s own comment says *"THE SUFFIX LIST IS OPEN — add the
spelling you meet"*, and that instruction **cannot fix this one**. The failure is case, not
spelling; someone following the comment will add lowercase entries and still miss `apiKey`.
`config.reference.json` is snake_case throughout, so the schema-derived self-test population
cannot exercise this either.

### 2. Escaped JSON inside a log string

```
$ printf '%s\n' '{"msg":"{\"password\":\"SENTINEL4\"}"}' | redact
{"msg":"{\"password\":\"SENTINEL4\"}"}
```

The pattern needs a literal `"` immediately before the key name; here the key is preceded by `\"`,
so it never matches. This is the ordinary structured-logging shape — a service logging a JSON
payload as a string field.

### 3. Non-string values

```
$ printf '%s\n' '{"api_token":12345678,"password":null}' | redact
{"api_token":12345678,"password":null}
```

The rule requires the value's opening quote, so a numeric token is unreachable.

### Also worth a line in the docs, not a defect

Over-redaction is broader than the current wording suggests: **any** key ending in `key` is
replaced, including `monkey`, `cache_key`, `primary_key`, `turnkey`. This is consistent with the
stated principle — over-redaction is safe for secrets and costly only for endpoints — and should
**not** be narrowed. It is worth recording so a noisy bundle is not later filed as a bug.

### Suggested next step, in order

1. Measure. Capture a real bundle and grep the log artifacts for JSON keys the rule cannot reach,
   before adding any pattern. Fixing an unreachable shape that never occurs is cost with no benefit.
2. Case is the cheapest of the three to close and the most likely to occur.
3. Consider the mechanical version of the invariant #1589 added as a comment: extract the suffix
   alternation from `lib.sh` and assert every entry is reachable by `selftest-redact.sh`'s screen.
   That comment exists because the superset claim was false at `wallet` — caught by review, and a
   check would not have needed review to catch it.

VijitSingh97:
Closes #1590 (by hand — `Closes` does not fire on this repo's `develop-v2` base).

#1590 named three JSON shapes `redact()`'s #1587 name rule cannot reach. Its own step 1 was a
**measurement**, so that came first: which of the three actually occur in the artifacts
`capture_artifacts` puts through `redact()`?

## What I measured (read-only, live stack plus two archived bundles)

| Population | JSON keys | distinct | shapes 1/2/3 found |
|---|---|---|---|
| container logs, `--tail 200`/service — the real capture window | 2,600 | 13 | 0 / 0 / 0 |
| container logs, `--tail 5000`/service | ~48,400 | 48 | 0 / 0 / 0 |
| live `api-state.json` | 16,107 | 243 | 0 / 0 / 0 |
| deployed `config.json` | 62 | 51 | 0 / 0 / 0 |
| two archived bundle `api-state.json` artifacts | 677 + 16,145 | 222 + 263 | 0 / 0 / 0 |

The scanner fired on all four buckets first, against the issue's own verbatim crafted inputs, and
a completeness control showed its key charset drops nothing: a deliberately loose
`"<anything>"\s*:` pattern returns byte-identical counts on every population.

**This corrects the issue's stated premise.** Container logs do not carry arbitrary
service-authored JSON here — a per-container partition puts every JSON key in the entire log
corpus in one service, Caddy (48,683 keys / 48 distinct), with the other nine contributing
**exactly zero**, and not one of Caddy's keys is secret-named in any casing. The log corpus is
live, so its totals are a snapshot and moved between captures; the distinct-key set did not. The name rule's real population is the two **pinned** artifacts, `config.json` and
`api-state.json`, whose shapes the schema and the dashboard API fix.

## What this ships

**Case** — the one shape whose argument does not rest on occurrence. The suffix alternation was
lowercase-only, and `redact()`'s own comment ("add the spelling you meet") **cannot** fix it: the
failure is case, not spelling, so a contributor following that instruction adds more lowercase
entries and still misses every `camelCase` key. `config.reference.json` is snake_case throughout,
so the schema-derived self-test population could not exercise it either. One `I` flag on one
existing `s///` — line-neutral against `lib.sh`'s 692/692 ceiling.

## What this does NOT ship, and why

Escaped-JSON (`"msg":"{\"password\":\"…\"}"`) and non-string values (`"api_token":12345678`).
Both are measured absent above, and the one-character fix for the non-string shape replaces the
value's trailing **delimiter** along with the value:

    {"api_token":<redacted>"mode":"local","port":18081}

— malformed JSON, in an artifact people read with `jq`. It corrupts what it is meant to protect.
Both shapes are pinned at today's behaviour under a `KNOWN GAP` label, so a row fails and names
the shape if either ever arrives.

Four mutations, each `cmp`-verified as landed before counting, each restored byte-identically:

| Mutation | Rows that red |
|---|---|
| drop the `I` | the case row |
| widen the suffixes with `url\|auth\|_id` | all three over-redaction rows (+5 pre-existing) |
| make all three quotes escape-tolerant | the escaped-JSON gap row |
| make the value's opening quote optional | the non-string gap row |

A first mutation attempt un-escaped only the *leading* quote and moved nothing — recorded here
because it landed cleanly and would have read as a vacuous row had I stopped there.

A **before/after pair**, both legs live against `origin/develop-v2`'s `lib.sh`, confirms the fix
moves only what it claims: shape 1 leaks before and is redacted after, while shapes 2 and 3 and
every must-survive field are byte-identical across both legs.

## Verification run

- `selftest-redact` 39 → 45 rows.
- `make test-integration-selftest`: 14 self-test files, all green, `selftest-redact: 45 passed, 0 failed`.
- `shellcheck --severity=warning` over `tests/integration/*.sh tests/integration/mini-stack/*.sh`
  (contains both changed shell files) — rc 0.
- `shfmt -i 4 -d` over the Makefile's exact glob — rc 0.
- `make lint-md` 0 errors; `make lint-docs-voice` OK; `make lint-file-budget` OK.
- Not run: the full `make lint-sh` (this box's OOM path, serialized) — substituted as above.
  No bench, no rig lock, no live-stack mutation; every measurement was read-only.

## Lane disclosure

`docs/dev/integration-testing.md` is a `docs/` change via `_shared.paths` — **no
`.lane-override`**. The redactor's coverage paragraph now states the case rule and both stated
gaps, so the doc cannot tell a reader the list is closed. That paragraph is the one #1584's
review blocked on, one revision earlier.

## An adjacent gap I checked and killed

The `KEY=value` rule is uppercase-only (`.*_PASSWORD` / `.*_TOKEN` / `.*_SECRET`), which is
shape 1's exact mirror in a different rule. It is **not** reachable: the deployed `.env` has 129
keys and none is anything but all-uppercase, and that file is *generated* by `pithead` from
`config.json`, so its casing is ours rather than a service's. Measured rather than filed.

## Over-engineering review, run off disk on this diff

**Acted on.** "Every JSON key comes from Caddy" was an **inference, not a measurement** — I had
concatenated all ten containers' logs into one file before scanning, so the partition was never
taken. Re-measured per container: Caddy 48,683 keys / 48 distinct, the other nine exactly zero.
The claim survived, and the test comment and this body now say it was measured that way.

**Declined — three survival rows could collapse into one.** `URL`, `Api_Auth` and `Chat_Id` could
be a single "the line is unchanged" assertion, which would be strictly stronger. Declined because
the file already keeps one named row per field (`xvb.url survives redaction`, and two more), and
a collapsed row fails without naming which field the widening ate.

**Declined — the measured figures in the test comment will go stale.** Kept, because they are
what justifies pinning the two shapes instead of fixing them, and the `KNOWN GAP` rows themselves
fail if the population ever changes, which is the corrective. Marked as a live snapshot.

**Considered, no change.** The new block runs about 20 comment lines to 12 of code, which is the
established ratio in this file rather than an excess.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -68,7 +68,8 @@ The test box holds real synced nodes and real keys. Treat it as production-sensi
   line whose name ends in `_PASSWORD` / `_TOKEN` / `_SECRET`; a credential given as a flag value;
   a v3 onion; and any opaque run of 90 or more characters, which is what catches wallet addresses
   (and a sha512 with them); and a JSON `"key": "value"` whose key ends in one of a list of secret
-  words — `password`, `token`, `key`, `username`, `wallet` and the rest.
+  words — `password`, `token`, `key`, `username`, `wallet` and the rest, matched without regard to
+  case, so `apiKey` and `PASSWORD` are reached alongside `api_key`.
   [#1582](https://github.com/p2pool-starter-stack/pithead/issues/1582) closed the flag-value and
   address-shape gaps; [#1587](https://github.com/p2pool-starter-stack/pithead/issues/1587) added
   the JSON one. **That last rule is keyed on the field NAME, not the value, and that is forced**:
@@ -76,7 +77,14 @@ The test box holds real synced nodes and real keys. Treat it as production-sensi
   them. **The name list is open, and one field is still out of reach** — `notifications.ntfy.url`
   is a capability URL whose key is the bare word `url`, which only its nesting separates from the
   public `xvb.url`, and a line-wise filter cannot see nesting. The self-test states that gap
-  rather than hiding it.
+  rather than hiding it. **Two further JSON shapes are stated rather than closed**
+  ([#1590](https://github.com/p2pool-starter-stack/pithead/issues/1590)): a JSON payload logged as
+  an escaped string inside another field (`"msg":"{\"password\":\"…\"}"`), and a secret whose
+  value is not a string (`"api_token":12345678`). Both were measured absent from every artifact
+  the harness captures — the container logs contribute JSON from one service only, Caddy, and not
+  one of its keys is secret-named in any case — and the one-character fix for the second replaces
+  the value's trailing delimiter along with the value, so it corrupts the JSON it is meant to
+  protect. The self-test pins both at today's behaviour, so a row fails if either shape arrives.
 - Continue-on-error. A failing assertion doesn't abort the run. The whole matrix is collected
   and summarized, with per-scenario artifacts for the failures.
 
@@ -683,7 +691,8 @@ Several self-tests sit beside it as standalone files, picked up by the same glob
 the counters and the real `summary()`, it censuses every harness file and fails if a skip leaves
 through a bare `it_warn` — by wording, and by shape for the drops that never say "skipping".
 `selftest-redact.sh` covers the shapes the redactor used to miss — a credential passed as a flag
-value, and secrets in JSON — and asserts in each case that the *raw* value is gone rather than
+value, and secrets in JSON under a key of any case — and asserts in each case that the *raw* value
+is gone rather than
 that a `<redacted>` marker appeared, since a marker can come from some other field on the same
 line. Its JSON population is derived from `config.reference.json` rather than listed in the file,
 under a screen deliberately wider than the redactor's own list, so a sensitive field added to the
--- a/tests/integration/lib.sh
+++ b/tests/integration/lib.sh
@@ -34,13 +34,13 @@ it_step() { echo -e "${IT_DIM}  → $1${IT_RESET}"; }
 
 # --- Secrets hygiene --------------------------------------------------------
 # Redact before anything reaches a log or the terminal. Three shapes: KEY=value / --flag value by
-# NAME, an opaque run >=90 chars by SHAPE, and JSON "key": "value" by the key's SUFFIX (#1587).
-# THE SUFFIX LIST IS OPEN — add the spelling you meet; a closed enumeration is what opened #1582.
-# Nesting is invisible line-wise, so a bare "url" key is unreachable: ntfy's is secret, xvb's is not.
+# NAME, an opaque run >=90 chars by SHAPE, and JSON "key": "value" by the key's SUFFIX (#1587),
+# matched CASE-INSENSITIVELY (#1590) — add new SPELLINGS, never case variants. THE LIST IS OPEN;
+# a closed one opened #1582. Nesting is invisible line-wise: ntfy's bare "url" is secret, xvb's not.
 redact() {
     sed -E \
         -e 's/(PROXY_AUTH_TOKEN|MONERO_NODE_PASSWORD|MONERO_NODE_USERNAME|.*_PASSWORD|.*_TOKEN|.*_SECRET)=.*/\1=<redacted>/; s/(--[a-z-]*(login|password|passwd|secret|token|key))([ =])[^[:space:]]+/\1\3<redacted>/g' \
-        -e 's/("[A-Za-z0-9_]*(password|passwd|secret|token|login|username|key|wallet|wallet_address|ping_url|donor_id)"[[:space:]]*:[[:space:]]*")([^"\]|\\.)*/\1<redacted>/g; s/[a-z2-7]{56}\.onion/<redacted>.onion/g; s/[A-Za-z0-9]{90,}/<redacted-address>/g'
+        -e 's/("[A-Za-z0-9_]*(password|passwd|secret|token|login|username|key|wallet|wallet_address|ping_url|donor_id)"[[:space:]]*:[[:space:]]*")([^"\]|\\.)*/\1<redacted>/gI; s/[a-z2-7]{56}\.onion/<redacted>.onion/g; s/[A-Za-z0-9]{90,}/<redacted-address>/g'
 }
 
 # --- Assertions -------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


VijitSingh97:
## Non-author pass — MERGE ON GREEN. One non-blocking nit, one condition on the close.

**Judged on opus.** Routine non-author pass on a 51-line diff — no top-tier trigger; the review protocol already decides it.

- **The self-test count.** Ran `selftest-redact.sh` at head `e529ef4` over the complete tree: **45 passed, 0 failed.** Your 39 -> 45 reproduces. (My first attempt read 23 passed / 1 failed — I had extracted only `tests/integration/`, so `config.reference.json` was absent and the schema-derived population came out empty. That failure was my harness, not your diff.)
- **A positive control you were not obliged to supply, and it fires.** The NEW `selftest-redact.sh` against the BASE `lib.sh`: **44 passed, 1 failed**, the single failure being `camelCase and UPPERCASE secret keys redact`. Arming readback in the instrument's own terms: `grep -c 'gI'` is **0 on base, 1 on head**.
- **That pair is also the proof the `I` flag did not leak sideways.** Exactly one row of 45 changes behaviour across it, so the two other `s///` commands sharing that `-e` — the `.onion` rule and the `{90,}` address rule — are measurably untouched. A flag binds to its own `s///` and not to its neighbours; here that is measured rather than assumed.
- **The three `case widening keeps ...` rows pass on BOTH sides.** Correct, and I want it on the record so nobody later reads them as null: they are narrowness guards, not discriminators. One behaviour changed, so exactly one row should discriminate — and exactly one does.

### RELAYED, not re-derived — and nothing rests on it

The `curl: (35)` infrastructure story for the first `Shell tests` failure. The rerun had already overwritten attempt 1's job records before I looked, so I cannot confirm it at source. I also do not need to: the rerun is the decisive instrument either way, and it was still `IN_PROGRESS` when I wrote this.

### Non-blocking nit — an unreflowed line in durable documentation

`docs/dev/integration-testing.md:695` is the 19-character stub `is gone rather than`, sitting in a paragraph whose every neighbour runs 91-98 characters; the insertion at 694 was never reflowed. Markdown renders it fine, so this is house voice rather than correctness — but it is durable in-repo prose. Fold it into 694/696 next time you touch the file; it does not earn a commit of its own.

### The one condition, and it is on the CLOSE, not on the diff

The body says `Closes #1590 (by hand)`. **#1590's title enumerates three shapes and this ships one**, with two deliberately not built on measured grounds. That is standing ruling 1's exact shape: a close trigger phrased as *"the named change landed"* goes ambiguous the moment the issue ships in slices, and the check is the ISSUE, never the PR.

I am not blocking, because closing is the right disposition here — the two remaining shapes end up recorded somewhere strictly stronger than an open issue: as self-test rows that FAIL if either shape ever arrives, plus the docs paragraph. **The condition is that the by-hand close comment must say that explicitly** — one shape fixed, two measured absent and pinned as KNOWN GAPs, with the grounds. A terse "done in #1600" leaves anyone searching for the escaped-JSON gap landing on a closed issue with no trace of the reasoning.

### Also noted, no action owed

Your suggestion 3 (extract the suffix alternation, assert every entry is reachable by the self-test screen) is correctly out of scope here and stays open on its merits. And the case widening's failure direction is the safe one: it can now over-redact a non-secret key ending in something like `Key` — a `cache_key`-shaped field in a third-party log line — which costs a debugging value, never a leak.

**Verdict: MERGE ON GREEN.** Nothing owed before merge; the close comment is the only thing I am asking for, and it can follow.

PR #1603, opened 2026-08-30
fix(integration): the merge-mining leg books an unanswerable monerod RPC as an accepted hole (#1597)

### LINKED ISSUE #1597: monero_caught_up cannot distinguish 'behind' from 'RPC unanswerable', and the merge-mining leg turns that into a by-design skip
`monero_caught_up` (`tests/integration/lib.sh`) reduces several independent conditions to one bit.
Its last step is `jq -e` over a body that `curl -fsS ... 2>/dev/null` leaves **empty** on any
failure, so "the node is behind" and "the RPC could not be answered" are indistinguishable to every
caller.

Measured directly, with the synced case as a fired positive control:

| body | `jq -e` rc | meaning to callers |
|---|---|---|
| `{"status":"OK","synchronized":true,"target_height":0}` | 0 | caught up — **control fired** |
| `{"status":"Failed"}` | 1 | not caught up |
| empty — RPC unreachable, connection refused, or a **401** | 4 | not caught up |

This is the general shape already recorded as a standing ruling in this fleet: **when an rc is one
bit over several independent doors, it cannot tell you which door fired.**

## Why it is benign in three callers and harmful in the fourth

| caller | what it does with a false bit | verdict |
|---|---|---|
| `run.sh:632` | `it_fail "monerod reports synced (RPC)"` | wrong *reason* text, right outcome — loud |
| `run.sh:1060` | `it_fail "Monero is synced"` | same |
| `_pred_monero_synced` (`lib.sh:512`) | fault-injection predicate | out of scope here |
| **`mergemine-probe.sh:122`** | **`it_skip_leg ... by-design`** | **wrong reason AND wrong class** |

The first two convert the ambiguity into a red, which is honest even when the message is imprecise.
The merge-mining leg converts it into a **counted skip classed `by-design`** — which on the skip
ledger means *an accepted hole*. Distinguishing an accepted hole from an uncovered run is exactly
what #1083 and #1365 built that ledger for, so this misfiles the second as the first.

## This state is not hypothetical on this stack

`lib.sh`'s own `env_bake_verdict` comment records an incident where a run recreated the live
containers with harness-rendered credentials while the on-disk `.env` kept the real ones —
internally consistent, healthy-looking, and *"every host-side RPC probe 401ed for a day"*. In that
state monerod is synced, p2pool has its merge-mining client, the `chain_id` line is in the log, and
the leg would nevertheless have skipped as by-design for a day.

## Two fixes, and the cheap one does not touch the shared function

**Local, and it uses the merge-mining probe's own argument.** p2pool constructs a
`MergeMiningClientTari` *only after* the block-header download succeeds. So if the startup capture
yields any such line, monerod demonstrably did catch up. Capture first; skip only when the capture is
*also* empty, and word that skip as "monerod could not be confirmed caught up" rather than asserting
it is not.

**Structural.** Give `monero_caught_up` a third answer — caught-up / behind / unanswerable — so
callers can choose. This touches every call site and is the larger change; the local fix is worth
taking first regardless.

## What I did not do

- **I did not verify that `run.sh:632` always executes in the same run as the merge-mining leg.** If
  it does, an RPC failure reddens there and the misfiled skip never hides alone, which is why I
  judged this non-blocking on #1594. If some scenario runs the probe without it, that judgement is
  weaker and should be revisited.
- I did not measure a live 401; the incident above is quoted from the repo's own comment, not
  reproduced.

VijitSingh97:
Closes #1597 — see "Scope" below for the half this deliberately does not take.

## The defect

`monero_caught_up` (`tests/integration/lib.sh`) reduces several independent conditions to one bit.
Its `curl -fsS ... 2>/dev/null` leaves the body empty on an unreachable host, on a refused
connection, and on a 401 — and `jq -e` over an empty body answers exactly as it does for a node
that is genuinely behind. Every caller gets one bit over several doors.

Three of the four callers turn that into a red, which is honest even when the message is imprecise.
The merge-mining leg turned it into a **counted skip classed `by-design`** — which on the skip
ledger means *an accepted hole*, the one class that says the project has agreed not to cover this.
Telling an accepted hole from an uncovered run is what #1083 and #1365 built that ledger for, so
this misfiled the second as the first.

The consequence is the shape this lane exists to kill: a stack whose monerod was synced, and whose
merge-mining client was up and reading chain_ids, would have been booked as an accepted hole for as
long as a credential stayed broken. Not hypothetical — `lib.sh`'s own `env_bake_verdict` comment
records a day of exactly that state.

## The fix is the order, not a new predicate

p2pool constructs a `MergeMiningClientTari` only after its block-header download succeeds, so any
such line is itself proof that monerod caught up — taken from p2pool rather than from the RPC we
could not reach. The capture is read first. The leg skips only when the window *was read* and held
no such line, and the skip's reason no longer asserts monerod's state as known.

Reading the capture first also **strengthens** the `by-design` class rather than merely preserving
it. The skip is now reached only where p2pool demonstrably built no client, so the header download
demonstrably did not complete, so no input, flag or env var to this harness would make the signal
exist. That is the actual `by-design` test, and it is now earned rather than assumed.

**Every case the reorder moves, it moves OUT of the skip ledger into a pass or a fail. No case that
previously passed or failed can become a skip.** That direction is the safety argument for the
change, and the self-test asserts the three moved cases rather than resting on it.

## Proven, not assumed

Three cases added to `selftest-mergemine-probe.sh`, each driven with the predicate answering
not-caught-up — which is what a 401 produces on a perfectly synced node — and each a counted
`by-design` skip before this change:

- a chain_id line now **PASSES** — the working leg that was being filed as a hole
- a client that never read a chain_id now **FAILS** — a real red that was hidden behind the skip
- an unreadable container start time now **FAILS** — a broken measurement, not a hole in coverage

### Mutation battery

Every arm was `cmp`-checked against the pristine file and `bash -n` parsed before its result was
counted; the file was restored and re-`cmp`'d afterwards. Run over the 40-case file:

| mutation | reds |
|---|---|
| restore the old order (predicate first) | exactly the 3 new cases |
| restore the old skip wording | exactly the 2 wording assertions |
| drop only the emptiness guard, keep capture-first | 2 of 3 — see below |
| keep the new hedge **and** re-add the old claim | only the negative control — see below |

The third arm's survivor is not a coverage gap: the start-time case is bought by the reorder alone,
and the first arm reds it. The fourth arm exists because the two wording assertions could have been
redundant — a reason string can carry the new hedge and the old claim at once, reading as fixed
while still telling a reader the node is behind. It reds only the `case` control, so the pair is
proven distinct rather than argued to be.

## What was run

- `tests/integration/selftest-mergemine-probe.sh` — 40 passed, 0 failed
- `make test-integration-selftest` — rc 0 over the whole `selftest*.sh` glob (214 in the last file)
- `shellcheck --severity=warning` 0.11.0, over the full `tests/integration/*.sh` +
  `tests/integration/mini-stack/*.sh` globs from the repo root — rc 0
- `shfmt -i 4 -d` on both changed shell files — clean
- `make lint-md` — 0 errors, 48 files
- `make lint-docs-voice` — OK, 41 prose docs
- `make lint-file-budget` — OK. Both changed shell files sit under the 400-line target, so neither
  carries a ceiling row: 161 and 274 lines.

## Over-engineering review

Run off disk against this diff. One finding **applied**: the comment restated `lib.sh`'s
`env_bake_verdict` incident in four lines, which is the duplicated-claim-that-drifts pattern this
repo has already paid for once — replaced with a one-line pointer to the single copy.

Findings **declined**, with reasons:

- *The comment block is long for a six-line change.* Declined. The order of two statements **is**
  the whole fix, and it is invisible in the code: a later contributor tidying by moving the cheap
  predicate check back to the front reintroduces the defect silently, with every test still green
  unless these cases exist. The comment is the extension point.
- *The `case` control duplicates the `assert_contains` beside it.* Declined, and then measured —
  the fourth mutation arm above shows it does not.
- *`mm_leg_outcome 1 "@FAIL"` duplicates the existing `mm_leg_outcome 0 "@FAIL"` case.* Declined.
  Before this change those two cases had different outcomes (skip vs fail); asserting only the
  synced one is what let the unsynced one be a skip.

## Scope, and what I did not do

- **I did not touch `monero_caught_up` itself.** #1597's structural option — give it three answers,
  caught-up / behind / unanswerable — touches every call site and is not taken here. The other
  three callers still receive the ambiguous bit, and by the issue's own table that is benign in all
  three: each converts it to a red, which is honest even when the reason text is imprecise. I judge
  the harmful half closed by this PR and say so on the issue rather than quietly leaving it open.
- **I did not run the live suite.** Every result above is from the pure-logic self-test with the
  two collaborators stubbed. The reorder is not proven against a running stack.
- **I did not resolve the issue's own open item** — whether `run.sh:632` always executes in the
  same run as the merge-mining leg. It no longer bears on this fix, which does not depend on
  another leg reddening.

## Paths

`tests/integration/` (this lane's) plus `docs/dev/integration-testing.md`. `docs/` is in
`_shared.paths`, so no `.lane-override` was used or needed. Disclosed here as required.

The doc hunk also carries the ruled reflow debt from #1600's insert — a 19-char orphan line — since
this change edits that file anyway. That part is a pure re-wrap, asserted content-identical before
it was written.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -354,12 +354,21 @@ and `--list` prints it).
   written against the rendered text matches nothing at all, silently — a probe that always reports
   an absence looks exactly like a working one until the day it matters. The escapes are stripped
   before matching, and the self-test carries that control fired in both directions: zero matches on
-  the captured bytes, one after stripping. When monerod is not caught up the leg skips rather than
-  fails, counted and classed `by-design`, because p2pool constructs the merge-mining client only
-  after its block-header download succeeds — so the signal cannot exist on an unsynced node, and no
-  input to this harness would create it. Only the matching lines cross the wire, which keeps the
-  container's argv line — carrying both wallet addresses, the RPC credential and the onion — out of
-  the capture.
+  the captured bytes, one after stripping. The capture is read before the caught-up predicate, and
+  that order is itself an assertion ([#1597](https://github.com/p2pool-starter-stack/pithead/issues/1597)). `monero_caught_up` answers with one bit that cannot
+  separate a node that is behind from an RPC that could not be answered at all: its `curl`
+  discards stderr, so an unreachable host, a refused connection and a 401 alike leave the empty
+  body that `jq` reads as not caught up. Asked first, that bit decided the whole leg, and a stack
+  whose merge-mining client was up and reading chain_ids would have been booked as an accepted
+  hole for as long as a credential stayed broken — a state this repo has recorded lasting a day.
+  Any `MergeMiningClientTari` line is itself proof that monerod caught up, since p2pool builds no
+  client until the header download succeeds, so the log settles what the RPC could not. The leg
+  skips only when the window was read and held no such line: p2pool then built no client, the
+  download did not complete, and no input to this harness would create the signal, which is what
+  earns the counted `by-design` class rather than the predicate's bit. Every case the order moves,
+  it moves out of the skip ledger into a pass or a fail, never the other way. Only the matching
+  lines cross the wire, which keeps the container's argv line — carrying both wallet addresses,
+  the RPC credential and the onion — out of the capture.
 - Dashboard reads live state. `/api/state` is reachable; Monero is synced (`done`); pruned/full
   display matches `monero.prune` ([#32](https://github.com/p2pool-starter-stack/pithead/issues/32)); the sidechain `pool.type` matches `p2pool.pool`.
 - The Monero node's ZMQ endpoint is a live ZMTP publisher. A ZMTP handshake against the node's ZMQ
@@ -692,22 +701,21 @@ the counters and the real `summary()`, it censuses every harness file and fails 
 through a bare `it_warn` — by wording, and by shape for the drops that never say "skipping".
 `selftest-redact.sh` covers the shapes the redactor used to miss — a credential passed as a flag
 value, and secrets in JSON under a key of any case — and asserts in each case that the *raw* value
-is gone rather than
-that a `<redacted>` marker appeared, since a marker can come from some other field on the same
-line. Its JSON population is derived from `config.reference.json` rather than listed in the file,
-under a screen deliberately wider than the redactor's own list, so a sensitive field added to the
-schema fails the test by name until someone classifies it as redacted or as a stated survivor. It
-also guards the other direction: a sha256 digest and a non-secret flag value must survive, because
-an artifact with its image pins stripped is useless for the triage it exists for.
-`selftest-zmq-probe.sh` drives the ZMTP verdicts from captured and hand-built wire fixtures, so
-every failure class — a silent peer, a non-ZMQ listener, a ZMTP peer that is not a publisher, a
-READY frame carrying a decoy `Socket-Type` value — is reachable with no socket and no stack.
-It also covers the truncated and hostile frames that used to end the parser rather than be named
-by it ([#1500](https://github.com/p2pool-starter-stack/pithead/issues/1500)): every length field is read with `16#`, which is fatal on an
-empty string, so a peer that stalled part-way through a header left an interpreter error on stderr
-and an empty verdict. Those cases assert the empty stderr alongside the reason, because the return
-code was already 1 while the bug was live and a case checking only the code would have passed
-against it.
+is gone rather than that a `<redacted>` marker appeared, since a marker can come from some other
+field on the same line. Its JSON population is derived from `config.reference.json` rather than
+listed in the file, under a screen deliberately wider than the redactor's own list, so a sensitive
+field added to the schema fails the test by name until someone classifies it as redacted or as a
+stated survivor. It also guards the other direction: a sha256 digest and a non-secret flag value
+must survive, because an artifact with its image pins stripped is useless for the triage it exists
+for. `selftest-zmq-probe.sh` drives the ZMTP verdicts from captured and hand-built wire fixtures,
+so every failure class — a silent peer, a non-ZMQ listener, a ZMTP peer that is not a publisher, a
+READY frame carrying a decoy `Socket-Type` value — is reachable with no socket and no stack. It
+also covers the truncated and hostile frames that used to end the parser rather than be named by
+it ([#1500](https://github.com/p2pool-starter-stack/pithead/issues/1500)): every length field is
+read with `16#`, which is fatal on an empty string, so a peer that stalled part-way through a
+header left an interpreter error on stderr and an empty verdict. Those cases assert the empty
+stderr alongside the reason, because the return code was already 1 while the bug was live and a
+case checking only the code would have passed against it.
 
 ---
 
--- a/tests/integration/mergemine-probe.sh
+++ b/tests/integration/mergemine-probe.sh
@@ -112,20 +112,38 @@ mm_capture_startup() {
 
 # The release-gate leg: PASS, FAIL, or an honest counted SKIP — never a silent green.
 #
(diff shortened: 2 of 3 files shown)


VijitSingh97:
## Non-author pass — **MERGE ON GREEN**. Nothing blocking, nothing owed by the author.

Reviewed at head `a2c9ca6`. One finding, non-blocking, handled below rather than bounced back.

### PROVEN BY ME — re-derived at source, not relayed

- **Self-test at the head**: `40 passed, 0 failed`, rc 0, run in a worktree pinned at `a2c9ca6`.
- **Mutation arm 1, re-derived as a prediction rather than read from the table.** Restoring the old
  order while keeping the new wording, with an arming readback first (`cmp` rc 1 against the
  pristine copy, the reordered block printed, `bash -n` rc 0): **37 passed, 3 failed, and the three
  reds are exactly the three new cases.** File restored and `cmp`'d back to pristine afterwards.
  My first attempt at this arm was a no-op — the readback is what caught it, not the outcome.
- **The safety claim, enumerated rather than argued.** Over `caught_up × capture` = {yes,no} ×
  {unreadable, empty, non-empty}, the reorder moves exactly three cells, all from `caught_up=no`:
  unreadable `skip → fail`, non-empty `skip → verdict`, and empty unchanged. **No cell that
  previously passed or failed can become a skip.** The PR's direction claim holds as stated.
- **`EXPECTED_ALWAYS` contains `p2pool`** (`lib.sh:222`), so the `skip → fail` transition on an
  unreadable start time is a harness fault everywhere the leg runs, not a coverage hole. That is
  the transition I went looking for a counter-example to; there isn't one.
- **The doc reflow is content-identical**: normalized to one line, the paragraph is 1617 bytes on
  both sides and byte-identical. The comparison is not a null instrument — fired as a positive
  control on the adjacent paragraph, which differs at byte 140.
- **Wrap width did not regress**: the doc's longest lines are unchanged base to head
  (1219/586/534/403/399). The longest added line is 159 chars, inflated by an inline issue link,
  which is the file's existing convention.
- **Budget gate rc 0 on both legs.** The branch was directly on the tip when I started — one run
  would have answered both — but merging #1602 moved the base under it, so I ran the branch *and*
  the `merge-tree` result separately. Both green; `docs/dev/file-budget.tsv` is untouched, so no
  row was rewritten.
- **Stale-prose sweep** for the old skip wording: the only hit repo-wide is the deliberate negative
  control in the self-test, which doubles as the sweep's firing control.

### RELAYED, not re-derived

The `shellcheck`/`shfmt`/`lint-md`/`lint-docs-voice` runs, and mutation arms 2, 3 and 4. Arm 1 was
the one the safety argument rests on, so it is the one I reproduced.

### The finding — non-blocking, and I have taken the remedy rather than returning the PR

`Closes #1597` closes an issue whose title names two defects and whose first clause — the predicate
answering with one bit over several doors — this PR deliberately does not touch. Auto-closing it
would leave that with no tracker. The PR discloses the scope honestly, so this is issue hygiene,
not a defect in the change: **filed as #1605** rather than blocking a green PR over it.

While there, I settled #1597's own open item, which this PR records as unresolved: `run.sh:632`
and `assert_mergemine_roundtrip` (line 659) both sit inside `assert_running_state()` (from line
522) with **no early return between them**, so they always execute in the same run. An unanswerable
RPC has always reddened at 632 alongside the misfiled skip. That strengthens the PR's judgement
that the remaining callers are benign.

### Nits, taking neither

The comment block is long, and it should be — the order of two statements is the whole fix and is
invisible in the code. The PR argued this down already and I agree with the argument.

Issue #1596, opened 2026-08-30
Both redaction paths miss wallet addresses in argv: --wallet by name, the Tari address by position, and two of three Tari forms by shape


VijitSingh97:
Both redaction paths key wallet addresses to a **flag name** or a **character-count shape**. The
p2pool launch line carries the Monero address as `--wallet <addr>` and the Tari address as a **bare
positional after `--merge-mine tari://host:port`** — a positional has no name to key on, and two of
the three Tari address forms this repo itself accepts are too short or too non-alphanumeric for the
shape rule. Measured, not read.

This is **not** #1582 (closed: its JSON and `--flag value` shapes do redact) and **not** #1590 (JSON
shape gaps). It is the argv half that neither covered.

## Site 1 — `lib/pithead/07-support-bundle.sh` — this is the serious one

The container-log redaction is one `sed` over three flag names: `--rpc-login`,
`--http-access-token`, `--tls-fingerprint`. There is no shape rule at all.

Running the real expression over a synthesised p2pool launch line, using the suite's own fixtures
by name (`VALID_PRIMARY`, `VALID_TARI_SINGLE` from `tests/stack/lib.sh`):

| value on the line | result |
|---|---|
| `--rpc-login user:pass` | redacted — **positive control, the expression ran** |
| Monero address after `--wallet` | **survives verbatim** |
| Tari address as the `--merge-mine` positional | **survives verbatim** |

**Why this matters more than the CI path:** `support-bundle` is the artifact we ask *operators* to
generate and send to us. Its own header comment says secrets are "redacted at the source ...
container logs by argv pattern". That is true of the RPC credential and not of either wallet. The
bundle is `chmod 600` and the command does tell the operator to review before sharing, which is the
mitigation currently carrying this.

## Site 2 — `tests/integration/lib.sh` `redact()` — narrower, but fragile

Here a shape rule (`[A-Za-z0-9]{90,}`) does the wallet work. Against the three Tari forms the repo
validates — `28-parse-and-validate-config.sh` states plainly that "Tari addresses come in base58
AND emoji forms" — with the 95-char Monero fixture on an identical line as the positive control:

| form (fixture name) | length | result |
|---|---|---|
| base58 (`VALID_TARI`) | 91 | redacted — **clears the 90 threshold by one character** |
| single (`VALID_TARI_SINGLE`) | 48 | **survives verbatim** |
[...]

Issue #1593, opened 2026-08-30
Five of the seven files the release gate uploads as "redacted" have never been audited for secret shapes


VijitSingh97:
#1582 measured two of the seven files `capture_artifacts()` writes — `logs.txt` and
`config.json` — and said so explicitly in its "what I did not do". It is now closed because the
shapes it named all redact, verified by running the function. **The other five files were never
audited, and closing #1582 is what would otherwise bury that.**

## The five

`capture_artifacts()` (`tests/integration/lib.sh`) writes, on any scenario failure:

- `compose-ps.txt` — `docker compose ps`
- `status.txt` — `pithead status`
- `doctor.txt` — `pithead doctor`
- `env.redacted.txt` — `cat .env`
- `api-state.json` — `api_state`

`.github/workflows/release-gate.yml` uploads the whole results directory under the step name
"Upload artifacts (redacted)", `if: always()`, 14-day retention. The gate runs `--local`, so on a
red run these carry the deployed stack's real values.

## Why this is not covered by the existing rules

`redact()` now handles `KEY=value` by name, `--flag value` by name, JSON `"key": "value"` by key
suffix, v3 onions, and opaque runs of 90+ characters. Each is keyed to a **spelling or a shape**.
The open question is not whether those rules work — they do — but **which spellings and shapes
these five outputs actually emit**, which nobody has looked at. A JSON key the suffix list does
not carry, or a credential rendered as a table column rather than `KEY=value`, passes through
untouched and the artifact still reads as redacted.

`api-state.json` is the one I would look at first: it is the only one of the five that is
machine-generated JSON, so it is the most likely to carry a key spelling nobody chose by hand.

## Suggested first step

Run the integration suite locally to a deliberate failure, then read all five captured files
directly — before adding any pattern. Same discipline #1590 records: measure a real bundle first,
because a pattern added to satisfy a guess is a pattern the self-test then locks in.

## Not a leak claim

No value from any live stack appears here, and none should appear in any comment on this issue.
If an audit finds one, report the path — never the value.

VijitSingh97:
Partial audit: **one of the five is done and it is clean for address shapes.** I named
`api-state.json` above as the one I would look at first, so I looked.

## What I ran

Fetched `/api/state` from a live stack (read-only, 311 KB), piped it through the real `redact()`
from `develop-v2`, and walked both the raw and redacted JSON comparing values **by key path**. No
value is printed here or was recorded anywhere.

[...]

PR #1607, opened 2026-08-30
fix(integration): redact wallet addresses by POSITION — the argv shape no length bar reaches (#1596)

VijitSingh97:
## What this is

**#1596's site 2 only** — `tests/integration/lib.sh`'s `redact()`. Site 1
(`lib/pithead/07-support-bundle.sh`) is another lane's path and is the more serious half of that
issue: it is **untouched here and still open**, so #1596 must not be closed on this PR.

The p2pool launch line (`docker-compose.yml:416-421`) names the Monero address as `--wallet <addr>`
and gives the Tari one as a **bare positional** after `--merge-mine tari://host:port`. A positional
has no name to key on, so the `>=90`-character SHAPE rule was the only rule reaching it — and it
reaches one of the three Tari forms this repo accepts.

## Reproduced before fixing — PROVEN BY ME, not relayed from the issue

The real `redact()`, over a synthesised launch line. Synthetic values throughout: nothing here is a
real address, and the forms only need the length and alphabet of the real thing.

| form | length | before this PR |
|---|---|---|
| base58 | 91 | redacted — clears the 90 bar by ONE character |
| single | 48 | **survives verbatim** |
| emoji | 67 glyphs | **survives verbatim** |
| Monero | 95 | redacted — positive control, the expression ran |

## The change

Two positional patterns appended to `redact()`'s first `-e`: the token after `--wallet`, and the
token after `--merge-mine`'s URI. Both refuse a token beginning with `-`, so a flag in that
position is never taken for an address.

`lib.sh` sits at its **692-line ceiling**, so the change is **line-neutral**: the patterns append
inside existing single-line `-e` arguments, and the header comment was rewritten in place at the
same four lines. `file budget OK` confirms it.

**Broader than the issue asked, deliberately:** the `--merge-mine` pattern does not require the
`tari://` scheme, so p2pool's generic `--merge-mine IP:port ADDR` form is covered too.

## What was RUN

- `tests/integration/selftest-redact.sh` — **54 passed, 0 failed** — 45 before this PR, measured on both sides rather than counted from the diff.
- Whole glob, per file, unfiltered: all 15 `selftest*.sh` green, `make test-integration-selftest`
  exit 0.
- Re-ran under `LC_ALL=C` and `LC_ALL=C.UTF-8` as well as the default locale — the emoji case does
  not depend on a UTF-8 locale, and CI's is not guaranteed.
- `shellcheck --severity=warning tests/integration/*.sh` (0.11.0) → rc 0. Measured over the whole
  31-file directory glob, not a partial one; the list contains both changed files. **Positive
  control:** an injected unused assignment made the same command exit 1 (SC2034), then reverted.
- `shfmt -i 4 -d` on both changed files → rc 0. `make lint-md` → 0 errors, 48 files.
  `make lint-docs-voice` → OK, 41 prose docs. `make lint-file-budget` → OK, monotonic.

### The new cases can FAIL, and each is attributed to the arm that kills it

Mutation battery, each leg `cmp`-checked against the original before being counted:

| leg | result |
|---|---|
| drop the `--wallet` pattern | 2 red: the SHORT value after `--wallet`, and the QUOTED address |
| drop the `--merge-mine` pattern | 3 red: Tari single, Tari emoji, and the quoted COMMAND column |

`lib.sh` restored byte-identical afterwards, control re-run green. The `base58` row deliberately
does **not** red under either leg: at 91 characters the `>=90` bar covers it, so it cannot
attribute anything. That is stated in the test file so a later reader does not mistake it for
proof of the new rule.

## The over-engineering pass, run off disk on my own diff

One finding accepted, two declined:

- **Accepted, and cut:** the Monero address was re-asserted once per Tari form. The `--wallet` path
  is identical on all three lines, so that was one measurement printed three times. Removed; the
  two arms are still proven to coexist on one line by the quoted COMMAND column case, which asserts
  both. The comment now says so, so the rows are not quietly re-added.
- **Declined —** the comment-to-code ratio in the new section. It matches what every prior section
  of this file does, and each block records a decision that would otherwise be re-litigated.
- **Declined —** "do not lower the bound" appears both here and in the doc. Two audiences: the doc
  is read by someone deciding whether to trust a bundle, the comment by someone editing the
  expression. The file already carries #1587 and #1590 the same way.

## Trade-off pinned by a test, not left to taste

The token match is **greedy to the next space**, so it absorbs punctuation attached to the address
— `compose-ps.txt` quotes its COMMAND column and the closing quote goes with it. I measured the
tidier alternative (stop the match at a quote) and **refused it**: on `--wallet "<addr>"` it
declines the leading quote, matches nothing, and the address survives whole. Absorbing a quote is
cosmetic; leaving one behind is the partial redaction that self-test file exists to catch. Both
quotings are now pinned, so the trade-off cannot be reversed for tidiness without a red row.

## Docs

`docs/dev/integration-testing.md` is updated via `_shared.paths` — **disclosed here, no
`.lane-override`**. Two corrections beyond adding the new shape:

- The file stated the 90-character rule "is what catches wallet addresses". That is **the defect's
  own words carried as fact**, and it is the sentence that tells a reader they may stop checking.
- Adding the new rule to that list broke the referent of the sentence beginning "That last rule",
  which had pointed at the JSON rule. Named explicitly instead.

## What I did NOT do

- **Site 1 is untouched.** Not my path, and it is the half operators are actually asked to send us.
- **I did not measure a real bundle from a running stack.** Every figure comes from running the
  real expression over synthesised lines. That proves what the expression can and cannot reach; it
  does not prove any deployed stack emitted one.
- **A gap I found and did not fix here:** in `.env`, a `*_WALLET_ADDRESS` value shorter than 90
  characters survives — the `KEY=value` rule keys on `_PASSWORD` / `_TOKEN` / `_SECRET` only, and
  none of the new positional rules apply to a `KEY=value` line. `env.redacted.txt` is one of the
  five files **#1593** already names as never audited, so this belongs to that audit rather than to
  a new issue; I have recorded it there.
- **One sentence was dropped from `lib.sh`'s header comment** to stay line-neutral: the note that
  nesting is invisible to a line-wise filter (ntfy's bare `url`). It is replaced by a pointer to
  `selftest-redact.sh`, which carries that gap in full with its issue number, as does the doc.
  Nothing is lost repo-wide, but it is a deletion and not a reword.

Not closing #1596 — site 1 is still open, and `Closes` never fires here anyway: PRs target
`develop-v2` while the default branch is `develop`.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -66,13 +66,21 @@ The test box holds real synced nodes and real keys. Treat it as production-sensi
   through a redactor. **What that redactor covers is narrower than "artifacts are redacted"
   suggests**, and the gap is worth knowing before you trust a bundle. It reaches: a `KEY=value`
   line whose name ends in `_PASSWORD` / `_TOKEN` / `_SECRET`; a credential given as a flag value;
-  a v3 onion; and any opaque run of 90 or more characters, which is what catches wallet addresses
-  (and a sha512 with them); and a JSON `"key": "value"` whose key ends in one of a list of secret
-  words — `password`, `token`, `key`, `username`, `wallet` and the rest, matched without regard to
-  case, so `apiKey` and `PASSWORD` are reached alongside `api_key`.
+  a v3 onion; any opaque run of 90 or more characters (and a sha512 with them); the token after
+  `--wallet` and the one after `--merge-mine`'s URI, by POSITION; and a JSON `"key": "value"` whose
+  key ends in one of a list of secret words — `password`, `token`, `key`, `username`, `wallet` and
+  the rest, matched without regard to case, so `apiKey` and `PASSWORD` are reached alongside
+  `api_key`.
   [#1582](https://github.com/p2pool-starter-stack/pithead/issues/1582) closed the flag-value and
   address-shape gaps; [#1587](https://github.com/p2pool-starter-stack/pithead/issues/1587) added
-  the JSON one. **That last rule is keyed on the field NAME, not the value, and that is forced**:
+  the JSON one; [#1596](https://github.com/p2pool-starter-stack/pithead/issues/1596) added the
+  positional one. **The length rule never covered wallet addresses as widely as it read**: a
+  positional argument has no name to key on, so that rule was the ONLY one reaching the Tari
+  address, and of the three Tari forms this repo accepts it reached only the base58 one — by a
+  single character.
+  Lowering the bound is not available: it starts matching ordinary log tokens well before it
+  reaches the 48-character form, and no bound reaches the emoji form at all.
+  **The JSON rule is keyed on the field NAME, not the value, and that is forced**:
   a view key and a container digest are both 64 hex characters, so nothing in the value separates
   them. **The name list is open, and one field is still out of reach** — `notifications.ntfy.url`
   is a capability URL whose key is the bare word `url`, which only its nesting separates from the
@@ -700,7 +708,8 @@ Several self-tests sit beside it as standalone files, picked up by the same glob
 the counters and the real `summary()`, it censuses every harness file and fails if a skip leaves
 through a bare `it_warn` — by wording, and by shape for the drops that never say "skipping".
 `selftest-redact.sh` covers the shapes the redactor used to miss — a credential passed as a flag
-value, and secrets in JSON under a key of any case — and asserts in each case that the *raw* value
+value, an address given by POSITION in an argv line, and secrets in JSON under a key of any case —
+and asserts in each case that the *raw* value
 is gone rather than that a `<redacted>` marker appeared, since a marker can come from some other
 field on the same line. Its JSON population is derived from `config.reference.json` rather than
 listed in the file, under a screen deliberately wider than the redactor's own list, so a sensitive
--- a/tests/integration/lib.sh
+++ b/tests/integration/lib.sh
@@ -33,13 +33,13 @@ it_err() { echo -e "${IT_RED}[ITEST]${IT_RESET} $1" >&2; }
 it_step() { echo -e "${IT_DIM}  → $1${IT_RESET}"; }
 
 # --- Secrets hygiene --------------------------------------------------------
-# Redact before anything reaches a log or the terminal. Three shapes: KEY=value / --flag value by
-# NAME, an opaque run >=90 chars by SHAPE, and JSON "key": "value" by the key's SUFFIX (#1587),
-# matched CASE-INSENSITIVELY (#1590) — add new SPELLINGS, never case variants. THE LIST IS OPEN;
-# a closed one opened #1582. Nesting is invisible line-wise: ntfy's bare "url" is secret, xvb's not.
+# Redact before anything reaches a log or the terminal. FOUR shapes: KEY=value / --flag value by NAME;
+# an opaque run >=90 chars by SHAPE; the token after --wallet and after --merge-mine's URI by POSITION
+# (#1596: two of three Tari forms are too short for SHAPE); JSON "key": "value" by key SUFFIX (#1587)
+# CASE-INSENSITIVELY (#1590) — add SPELLINGS, never case. LIST IS OPEN (#1582); gaps: selftest-redact.sh
 redact() {
     sed -E \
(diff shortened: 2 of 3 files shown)


VijitSingh97:
## Non-author review — MERGE ON GREEN

No findings. I re-derived every claim this decision turns on rather than relaying it, including the
one the PR exists for, and the only thing outstanding is `Shell tests`, still IN_PROGRESS as I write.
Every other check is terminal SUCCESS or SKIPPED.

### The fix works — measured here, both sides, with the base leg as a fired control

I ran the **real `redact()`** at the base and at `a11c664` over a synthetic corpus (nothing real:
the forms carry only the length and alphabet of the thing). Leaks surviving each side:

| case | base | head |
|---|---|---|
| argv, Tari base58 91 | — | — |
| argv, Tari single 48 | **LEAKS** | — |
| argv, Tari emoji 67 | **LEAKS** | — |
| quoted COMMAND column | **LEAKS** | — |
| `--wallet=` / `--merge-mine=` forms | **LEAKS** | — |
| `--merge-mine host:port ADDR`, no scheme | **LEAKS** | — |
| `TARI_WALLET_ADDRESS=` (`KEY=value`) | LEAKS | **LEAKS** |
| bare address in prose | LEAKS | **LEAKS** |
| `--merge-mine` with no address following | — | — |

Six cases leak at base and two at head, so the base leg is a fired positive control rather than a
silent pass. The base58 row is redacted on **both** sides — the `>=90` bar already reached it — which
is exactly what the PR says rather than something it glosses. The last row confirms no
over-redaction: `--local-api` after the endpoint is untouched, and the endpoint itself survives.

**The two survivors are both disclosed in the PR body**, and I agree with where each was sent: the
`KEY=value` form belongs to #1593's audit of the five never-audited artifacts, and a bare unlabelled
address has no name, no position and no shape to key on — no rule could reach it without matching
ordinary log tokens.

> A defect in my own first attempt, since it bears on how much the table above is worth: my first
> base leg produced **zero lines** and I nearly read that as "nothing leaked". `lib.sh` sources
> `skip-accounting.sh` from beside it and refuses otherwise, and I had hidden the error with
> `2>/dev/null`. The table above is the re-run, with both legs at 9 lines, rc 0, empty stderr.

### The new rows bite

Dropping the `--merge-mine` pattern (armed with a `cmp` readback, `grep -c` on the pattern going
1 → 0) reds **exactly 3**: Tari single, Tari emoji, quoted COMMAND column — the three the PR
attributes to that arm, and no others. Restored, `git status` clean.

### Counts and budget, measured rather than relayed

* **45 at base, 54 at head.** I ran the base pair (base `lib.sh` + base `selftest-redact.sh`, with a
  readback confirming the positional rule was absent) and the head pair. Both sides, not counted off
  the diff.
* `tests/integration/lib.sh` is **692 lines at base and 692 at head** against a 692 ceiling — genuinely
  line-neutral, which it had to be, since that row has zero headroom.
* `bash scripts/lint-file-budget.sh` rc **0**, monotonic. Branch is directly on the tip, so merged
  tree == branch tree and the one run answers both.

### The disclosed deletion is honest

The PR says one header sentence was dropped to stay line-neutral and that nothing is lost repo-wide.
Verified: `selftest-redact.sh:238` carries it verbatim — *"nesting is invisible to a line-wise
filter"* — with #1587 attached, and `docs/dev/integration-testing.md` keeps the `notifications.ntfy.url`
paragraph. The claim holds.

### The doc correction is the most valuable part of this PR

Removing *"which is what catches wallet addresses"* matters more than the code change: that sentence
was the defect's own reasoning carried as documented fact, and it is what tells a reader they may
stop checking. The replacement states the bound's real reach and why lowering it is unavailable —
and that second claim is itself pinned by an existing row, since the `sha256` digest survivor test
reds at any bound low enough to reach the 48-character form.

### Not run by me

`shellcheck`/`shfmt` — CI's `Shell tests` covers them and `make lint-sh` is this box's OOM path.
`make test-integration-selftest` over all 15 files is relayed; I ran `selftest-redact.sh` only.

### Scope, agreed and recorded

**#1596 stays OPEN.** Site 1, `lib/pithead/07-support-bundle.sh`, is untouched and is the half
operators are actually asked to send us. I will route it rather than leave it to be picked up by
whoever reads the issue next.

Issue #1609, opened 2026-08-30
The release-gate artifact publishes the host's public IP, and redact() has no rule that can reach an IP address


VijitSingh97:
Found while completing the #1593 audit; the measurements are in
https://github.com/p2pool-starter-stack/pithead/issues/1593#issuecomment-5467654722 and are not
repeated in full here.

`capture_artifacts()` (`tests/integration/lib.sh:686`) writes `doctor.txt` through `redact()`, and
`pithead doctor` prints the host's own public IP addresses in the clear:

    check_stratum_exposure  ->  lib/pithead/20-doctor-install-checks.sh:27  (concatenated: pithead:4784)
    value from host_public_ips  ->  lib/pithead/19-small-utilities.sh:209

`host_public_ips` walks `ip -o addr show` and emits every `inet` **and** `inet6` address that passes
`is_public_ip`, so both families can appear.

**`redact()` has no rule that can reach an IP address.** Not the `KEY=value` name rule, not the
`--flag value` rule, not the JSON key-suffix rule, not the `.onion` rule, and not the
`[A-Za-z0-9]{90,}` shape rule — whose alphabet excludes both `:` and `.`.

Measured rather than reasoned: today's `redact()` changes **0 lines** across the three
`docker`/`pithead`-derived artifacts in two real bundles, while a seeded control line in a copy of
one of them changed as expected. The line is present in both bundles.

## Severity — latent, not realized

Stated up front so nobody spends the afternoon on it:

- `release-gate.yml` uploads the results directory `if: always()`, 14-day retention, from a
  `self-hosted` runner, so a red run would publish the runner's own address.
- **It has never happened.** All 96 release-gate runs to date concluded `cancelled` (55) or
  `skipped` (41), and `release-gate-results` is not among the retained artifacts.
- `tests/integration/results/` is gitignored (`.gitignore:40`) and no bundle is tracked.

So this is a gap to close before the gate first runs red, not an incident.

Also worth knowing: the repo's pre-publish topology detector reads a public IP as `clean`, verified
against an input it does flag, so that check will not catch this class either.

## Fix direction — the ranking, since that is what the next reader inherits

**Option A — anchor on the message.** Redact the parenthetical in `public IP (...)`.
Exact, no collateral, and wrong for the reason #1582 was found by: it is an enumeration of one
site's wording. It also couples the harness to the `pithead` script's prose, so rewording the
warning silently disarms the guard.

[...]

PR #1610, opened 2026-08-30
fix(integration): redact an IP by SCOPE — the shape no name, key or length rule reaches (#1609)

VijitSingh97:
`Closes` does not fire on this repo — PRs merge to `develop-v2` while the default branch is
`develop`. **#1609** is the issue; close it by hand.

Found by completing the **#1593** audit of the five never-audited artifact files. The audit and its
method are on that issue; this PR fixes one of its two findings.

## The defect

`pithead doctor` interpolates the host's OWN public addresses into its stratum-exposure WARN
(`lib/pithead/20-doctor-install-checks.sh:27`; the value comes from `host_public_ips`, which emits
every `inet` and `inet6` that passes `is_public_ip`). `capture_artifacts` puts that through
`redact()` into `doctor.txt`, and `release-gate.yml` uploads the whole results directory
`if: always()` from a `self-hosted` runner, 14-day retention, on a public repo.

**No rule in `redact()` could reach it** — not the `KEY=value` name rule, the `--flag value` rule,
the JSON key-suffix rule, the `.onion` rule, nor the `[A-Za-z0-9]{90,}` shape rule, whose alphabet
excludes both `:` and `.`.

**Latent, not realized.** All 96 release-gate runs to date concluded `cancelled` (55) or `skipped`
(41), `release-gate-results` is not among the retained artifacts, and `tests/integration/results/`
is gitignored with nothing tracked. This closes a gap before the gate first runs red.

## The invariant the rule rests on

**The protect list is `is_public_ip`'s private set verbatim** — `0`, `10`, `127`, `169.254`,
`172.16-31`, `192.168`, `100.64-127` — **and the IPv6 half is that function's own test**,
`2000::/3`. Holding the two lists identical is what makes the coverage argument checkable rather
than anecdotal: an address `doctor` can emit as public is one this rule redacts, with no third
classification.

Implementation is protect-then-redact-then-restore: one dot of a reserved address is mangled to
`\x01`, which breaks the quad so the redaction step cannot match it; then the sentinel is restored.
`\x01` was measured absent from all 14 files of the corpus.

## Two measured refusals, each pinned by a test

- The IPv6 expression first ate **HAProxy/CLF timestamps**: a four-digit year opening with a 2 is
  four hex digits, and the clock supplies the colon groups. That was **278 of the 280 lines** the
  first draft changed across the corpus. Excluding a leading `/` separates a date path from a bare
  address. The repo's own `lint-topology` makes the same misreading, which is why that fixture
  joins its year at runtime rather than as one literal.
- The reserved-protect expression is **anchored to the start of a quad**. Drop the anchor and
  `8.10.0.1` — a public address carrying `10.` at its second octet — has its dot mangled and leaks.

## The over-engineering pass changed the design

An earlier draft **also protected multicast/reserved `224-255`**, for mDNS destinations in logs.
Measured: **zero occurrences across the corpus**, and it was the one entry *not* in `is_public_ip`'s
private set — so it bought nothing and cost the exactness of the invariant above. Removed.
`224.0.0.251` now redacts, matching what `doctor` would call it, and that consequence has its own
test row rather than being left to inference. The corpus differential is byte-identical before and
after the removal, which is the evidence the entry was unearned.

## Evidence

Differential over the two-bundle corpus, old `redact()` vs new, all 14 files: `doctor.txt` 1 line in
each bundle, `logs.txt` 2 lines, **nothing else**. Control: old and new agree on a secret line and
differ on a public IP.

Mutations — each attributed to the case that kills it, each refusing to count on FILE UNCHANGED,
`lib.sh` restored byte-identical after each:

| mutation | cases red |
|---|---|
| delete the IPv4 redaction | 3 |
| delete the IPv6 redaction | 2 |
| delete the reserved-protect | 11 |
| delete the sentinel restore | 11 |
| drop protect's leading anchor | 2 |
| drop the `/` exclusion | 1 |
| delete the CGNAT entry alone | 1 |
| IPv6 `{0,4}` -> `{1,4}` | 1 |

The last two matter on their own: they prove the survivor rows **discriminate individually** rather
than being one measurement printed seven times. One of them also caught itself — after the multicast
removal shifted the alternation, the CGNAT mutation's anchor no longer matched, and the FILE
UNCHANGED guard refused to count it instead of reporting a false pass.

## What was RUN

`selftest-redact.sh` 71 passed / 0 failed; all 15 `tests/integration/selftest*` green; `lint-md`,
`lint-docs-voice`, `lint-file-budget`, `lint-topology`, `lint-operator-strings` all OK; `shfmt -i 4
-d` clean over the Makefile's exact glob; `shellcheck` 0.11.0 over the two changed files — **4
findings at base, 4 at head, delta 0**.

**NOT RUN:** no live stack, no bench, no gate run. `make lint-sh` was not run (this box's OOM path);
the targeted shellcheck plus full-glob shfmt above is the standing substitute, and the file list it
was measured over is named.

## Disclosures

- **`docs/dev/integration-testing.md` is touched under `_shared.paths`, no `.lane-override`.** Its
  enumeration of what the redactor reaches would otherwise have stated the exact limit this closes.
- **`lib.sh` is at its 692-line ceiling, so the change is line-neutral** — the expressions append
  inside the existing single-line `-e` argument, and the header comment was rewritten in place at
  the same four lines. It said FOUR shapes and now says FIVE.
- **Peer addresses in `logs.txt` are now redacted too** — the 2 corpus lines are a monerod clearnet
  peer and a Tor exit relay. Deliberate: a shape rule cannot tell "my address" from "a peer's", the
  artifact goes to a public repo, and the loopback, RFC1918 and container addresses that carry most
  of the triage value all survive.

## Findings I considered and declined

- **A four-component version string (`1.2.3.4`) would be redacted.** None appears in either bundle;
  the versions in these artifacts are three-component. One cosmetic loss, against an exclusion rule
  that would have to guess which dotted quads are versions. Declined.
- **Cisco-style MAC notation (`3c2d:4e5f:6a7b`) would match the IPv6 rule.** Absent from the corpus,
  and nothing in the stack emits that form. Declined rather than guarded on a guess — the same
  ordering #1590 records, and the same reasoning that removed the multicast entry above.
- **Documentation ranges (`192.0.2`, `198.51.100`, `203.0.113`, `2001:db8`) are deliberately NOT
  protected.** They are globally-routable by `is_public_ip`'s reckoning and never appear on a real
  interface, and leaving them redactable is what lets them serve as test fixtures.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -67,14 +67,19 @@ The test box holds real synced nodes and real keys. Treat it as production-sensi
   suggests**, and the gap is worth knowing before you trust a bundle. It reaches: a `KEY=value`
   line whose name ends in `_PASSWORD` / `_TOKEN` / `_SECRET`; a credential given as a flag value;
   a v3 onion; any opaque run of 90 or more characters (and a sha512 with them); the token after
-  `--wallet` and the one after `--merge-mine`'s URI, by POSITION; and a JSON `"key": "value"` whose
+  `--wallet` and the one after `--merge-mine`'s URI, by POSITION; a JSON `"key": "value"` whose
   key ends in one of a list of secret words — `password`, `token`, `key`, `username`, `wallet` and
   the rest, matched without regard to case, so `apiKey` and `PASSWORD` are reached alongside
-  `api_key`.
+  `api_key`; and a globally-routable IP address, by SCOPE. What it keeps is exactly the set
+  `is_public_ip` calls private — loopback, RFC1918, link-local and CGNAT — because an artifact whose
+  port map and container addresses have been stripped cannot be triaged, and because holding the two
+  lists identical is what makes the coverage argument checkable rather than anecdotal.
   [#1582](https://github.com/p2pool-starter-stack/pithead/issues/1582) closed the flag-value and
   address-shape gaps; [#1587](https://github.com/p2pool-starter-stack/pithead/issues/1587) added
   the JSON one; [#1596](https://github.com/p2pool-starter-stack/pithead/issues/1596) added the
-  positional one. **The length rule never covered wallet addresses as widely as it read**: a
+  positional one; [#1609](https://github.com/p2pool-starter-stack/pithead/issues/1609) added the
+  IP one, after an audit found `pithead doctor` writing the host's own public address into
+  `doctor.txt`. **The length rule never covered wallet addresses as widely as it read**: a
   positional argument has no name to key on, so that rule was the ONLY one reaching the Tari
   address, and of the three Tari forms this repo accepts it reached only the base58 one — by a
   single character.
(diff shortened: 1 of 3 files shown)


VijitSingh97:
**Non-author pass: APPROVED, cleared to merge the moment `Shell tests` goes terminal-green. One finding, filed as a follow-up rather than a blocker.**

I did not relay this PR's table. I extracted `redact()` from both trees and ran the real expression over a 21-row hostile corpus of my own, with the base leg as the positive control — **base leaks both public addresses, head redacts both**, so the control fired and an empty leg could not have read as clean.

**Every collateral row survived at head**, including the two shapes #1609 predicted a blanket rule would eat: `127.0.0.1:12375->2375/tcp` and `0.0.0.0:18081->18081/tcp`, plus RFC1918 across all three blocks, `169.254`, CGNAT, `::1` / `fe80::` / `fd00::`, a tagged image digest, and a three-component version. The private-range boundaries are exactly right in both directions: `172.15.0.1`, `172.32.0.1` and `100.128.0.1` are all correctly redacted as public, which is the arm a sloppier protect list would fail.

Line-neutral against a zero-headroom row, verified at both trees: `tests/integration/lib.sh` is **692 at base and 692 at head**, and `docs/dev/file-budget.tsv:63` is 692. That is the constraint the ratchet imposes here and it is met exactly.

**Two candidate findings I ran and then withdrew, because this PR had already answered them.** The four-component version string is in *Findings I considered and declined*, and the CLF/HAProxy timestamp arm was not only found but pinned with a fixture. Correcting either would have been a correction issued against work already done — I mention it only so the record shows the pass checked them rather than assuming.

## The finding — filed as #1613, not a blocker

**The sentinel restore runs *after* the redaction step, so anything a pre-existing `\x01` reconstitutes is never examined.** Measured, with a fired control on the same run:

    input   203.0.113<SOH>47        ->   output   203.0.113.47      (emerges as a public IPv4)
    input   A<SOH>B                 ->   output   A.B               (silent corruption)
    control 198.51.100.9            ->   output   <redacted-ip>     (the expression ran)

This is **not** the thing the body discloses. The body measures `\x01` **absent from all 14 files of the corpus** — a property of today's inputs, honestly measured, and it is the real mitigation. What is not stated is the *failure mode* if one ever appears: it is not merely a corrupted byte, it is a dotted quad assembled after the only step that could have redacted it.

**Why this does not block.** It needs a producer to emit a control byte in place of a dot inside an address, which nothing in this stack does; the corruption arm is the reachable one and it is cosmetic; and holding this PR would leave a measured, realized-on-any-red-run public-IP leak open to close a contrived one. Ship the real fix.

**Merge condition, since the checks were not terminal when I reviewed:** 15 SUCCESS, 2 SKIPPED, `Shell tests (shellcheck + pithead suite)` IN_PROGRESS at `58e3dc9`. I will re-read the rollup at the moment of merge and merge only on all-terminal, none-failed. A pending check reading 0s has not started — e2e's own caution, and it is right.

PR #1612, opened 2026-08-30
fix(dashboard): #1556 slice 6 — the four parse-or-read pairs, and the pin's data out of the ratchet

VijitSingh97:
## What this is

#1556 slice 6 — the four parse-or-read pairs under `service/` — plus the structural fix the slice
was held on: the pin's data moved out of the file that carries the ratchet.

    service/audit_service.py:_tail_json_lines  -> list | None   signed
    service/audit_service.py:_entry_epoch      -> float | None  signed
    service/data_helpers.py:_parse_audit_ts    -> float | None  signed
    service/data_helpers.py:_read_host_config  -> dict | None   signed
    service/price_feed.py:parse_prices         -> dict | None   signed
    service/price_feed.py:fetch                -> dict | None   signed
    service/xvb_standby.py:parse_standby       -> dict | None   signed
    service/xvb_standby.py:fetch_once          -> dict | None   signed

The first multi-function slice. Every verdict was predicted from `classify`'s own branch chain
before the annotations were written, and all eight matched: blind 33 → 25, signed 18 → 26, with
collapse 14 / unjudged 9 / procedure 3 all unchanged. The reconciliation −8 blind == +8 signed
holds. The cut is by DOOR, not by module: these four are the subset whose blind pair is a parser
plus a reader both returning `None`, so the reading is one argument eight times rather than eight
arguments.

## The budget fix, and why it is this shape

The slice left `test_annotation_coverage.py` at 592 lines against a recorded ceiling of 579, and
ceilings only ever go down. Every slice adds prose plus a `PINNED` and an `_ANCHORS` row to that
one file, and 8 of 33 modules are still unpinned — so squeezing 13 lines defers the conflict by
one slice and no more.

Raising the ceiling was refused: 579 against a 400 target is already 1.45×, and `--generate`
rewrites every row to the actual count, so a hand-set high row does not survive the next
regeneration. Setting a *new* row high was tested and refused for the same reason (`check_monotonic`
iterates base rows only, so on a fresh row 579 FAILS and 620 PASSES — a path the guard forgot to
cover is not a ruling).

So the data moved instead, into `dashboard/tests/service/annotation_pins.py`. **The part that grows
is not the part that is budgeted.**

| File | Lines | Ceiling |
|---|---|---|
| `test_annotation_coverage.py` | 592 → **434** | 579, **unchanged** (still needed — over the 400 target) |
| `annotation_pins.py` (new) | **211** | none — under the 400 target, so the growing part is unratcheted |
| `annotation_gate.py` | 379 → **379** | none — three docstring lines rewritten, net zero |

`annotation_gate.py` was deliberately **not** the destination: at 379 it is 21 under target, so
+211 there mints a new row and hands the identical ratchet to the file that grows ~8 lines a slice.

The 579 ceiling was deliberately **not** lowered to 436. Lowering it to fit banks nothing — freeing
158 lines and handing them straight back is a net zero — and a lowered row reds every un-rebased
branch in every other lane.

## Two judgement calls, both disclosed

**`_UNJUDGED_AND_READ` moved too**, beyond the letter of the ruling and inside its stated test
("keeps the growing part un-budgeted"). It grows the same way — slice 5a added five entries and
~20 lines of reading, 5b two more — so leaving it behind would have left a second ratchet in
place. Trivially reversible if that reads wrong.

**The names keep their leading underscores.** `_unsigned_unjudged_under`'s docstring and the
narrowness control's both quote `_UNJUDGED_AND_READ` by name while arguing why law 2 is shaped as
it is, so a rename would have falsified prose in the same commit that moved the data.

## What the over-engineering pass found

One thing, and it is fixed in this diff rather than noted. The first draft argued the budget split
in **both** files, figures and all — `annotation_pins.py`'s docstring and a new block in
`test_annotation_coverage.py` each carried 592 / 579 / 8-of-33. A budget argument copied into two
files is two numbers to keep true, which is the exact failure this lane keeps shipping and which
this very file's header records happening twice. The argument and its figures now live only in
`annotation_pins.py`; the test file carries a pointer with no figures in it.

## The move is NOT verbatim, and the three exceptions are the point

Diffing the extracted block against the moved one shows exactly three changed lines, each a
cross-file reference the move would otherwise have falsified: one pointing at
`TestTheResidueThePinCannotRuleOn` as "below" when it is now in the sibling file, and two saying
"wearing this file's name" while meaning the pin's. Nothing else differs, and every deleted line
was confirmed present in the new module before the old one was cut.

## Controls — one per moved structure, because a green suite is what a broken import also produces

Each armed with an exact-count aim check and restored by sha256:

| | Seed | Result |
|---|---|---|
| A | DE-SIGN `_tail_json_lines` (drop ` \| None`) | exactly one FAILED: **law 2**, `[service/audit_service.py]` |
| B | Point `_ANCHORS["service/price_feed.py"]` at a function that does not exist | exactly one FAILED: **vacuity guard**, `[service/price_feed.py]` |
| C | UNANNOTATE `parse_prices` (remove the annotation outright) | exactly one FAILED: **law 1**, `[service/price_feed.py]` |

That is all three laws, each fired for its own reason and for exactly the module seeded. B and C
together prove `PINNED` supplied the parametrization and `_ANCHORS` is what the guard reads; A
proves `_UNJUDGED_AND_READ` still runs the subtraction — all across the new module boundary.

The aim check **refused a guessed signature twice** (0 matches, not 1): `_tail_json_lines` takes no
parameter annotation and `parse_prices` takes two parameters, not one. A plain `.replace()` would
have swallowed both and still reported a kill.

## Docs

`docs/` is in `_shared.paths` — disclosed here, not reached into. `docs/dashboard.md` and
`docs/dev/repo-map.md` each ENUMERATE the non-test modules under `dashboard/tests/`, so both went
stale the moment there was a second one. Swept the whole repo for sites naming `annotation_gate` /
`annotation_coverage` / "plain module" / "non-test module"; those two plus `annotation_gate.py`'s
own docstring were all of them.

## What I did NOT do

- **Nothing asserts `PINNED`'s size.** Shrinking the tuple silently narrows what laws 1 and 2
  cover, and every assertion stays green. This gap is pre-existing — it was the same when the
  tuple lived in the test file — but a data-only module reads as "just data" and may invite more
  casual edits, so it is worth a guard. Not built here: it is scope past this slice and deserves
  its own issue. Happy to file one.
- **No security-reviewer pass.** This change is annotations on four parsers plus a move of test
  data; it touches no control-channel, auth, credential or outbound-fetch surface.
- **`make test-patch-coverage` was not run as evidence.** Per #1557 `patch-coverage.sh` pins
  `COMPARE=origin/develop`, so its result on a develop-v2 PR is vacuous either way.

## Ran, in this worktree, at this head

- The two gate files: **109 passed** — the same count measured before the extraction, on the same
  two files.
- The full dashboard suite: **2352 passed**.
- Both proven invocations import the new module: `cd dashboard && python -m pytest` and the bare
  `pytest` console script.
- `ruff check` and `ruff format --check` clean over `dashboard/tests/service`.
- `make lint-py`, `lint-md`, `lint-docs-voice`, `lint-operator-strings`, `lint-topology`,
  `lint-file-budget` (with its 26-case self-test) — all OK, run from the repo root at the rebased
  head.
- The budget gate's base ref `origin/develop-v2` carries the 579 row, verified at source, so that
  green is the ratcheted comparison and not a fresh-record one.
- Rebased onto `origin/develop-v2`. `git rev-parse` gives the **same** `dashboard/` subtree hash
  and the same `file-budget.tsv` blob at the old base and the new one, so the counts above still
  describe the program being shipped.


VijitSingh97:
**Non-author pass: APPROVED, cleared to merge on terminal green. Both questions in the body are ruled below — neither needs work from the lane.**

## RULING 1 — `_UNJUDGED_AND_READ` stays where you put it. Do not split it back out.

You read the ruling correctly. Its letter named `PINNED` and `_ANCHORS` because those were the structures I had measured; its **test** was "keep the growing part un-budgeted", and `_UNJUDGED_AND_READ` grows the same way — five entries in 5a, two in 5b. Leaving it behind would have left a second ratchet in the file the ruling exists to take *out* of the ratchet's reach, so splitting it back would satisfy the wording and defeat the purpose.

**The general form, since this will recur:** when a ruling states a test and then names instances, the test governs and the instances are examples. If an instance is missing from the list, disclose the extension and ship it — which is exactly what you did. **A lane that widens a ruling inside its stated test and says so has followed it, not exceeded it.**

## RULING 2 — the `PINNED`-size gap is real; I have filed it as #1614 rather than sending it back to you.

You were right that it is past this slice and right not to build a guard inside it. It is filed with the trap named: **a hardcoded count is the wrong guard**, because every slice bumps it and it cannot tell a deliberate addition from a silent deletion. #1614 carries two directions that do not have that problem and the requirement that any guard be fired before it is trusted.

## What I re-derived rather than relayed

**The move is clean, and I did not take that from the PR body.** I parsed the base test file and the new module by AST and compared the constants **by value**, with a negative control:

| structure | base | head | base entries LOST | order of survivors |
|---|---|---|---|---|
| `PINNED` | 21 | 25 | **0** | preserved |
| `_ANCHORS` | 21 | 25 | **0** | preserved |
| `_UNJUDGED_AND_READ` | 12 | 12 | **0** | preserved |

The four added entries in each are exactly `service/audit_service.py`, `data_helpers.py`, `price_feed.py`, `xvb_standby.py` — the same four files carrying slice 6's eight `-> X | None` annotations. Population and diff agree.

**My first cut of that check was a null instrument and I am recording it because it is the more useful half.** I compared the structures for *equality* and got `identical=False` on two of three — which reads exactly like a lossy move. It was slice 6's own legitimate additions. **Equality was the wrong axis: the move should hold the base entries fixed, not the population.** The subset test is the one that measures what the move actually promised. Same family as the ruling this fleet already carries about confirming an instrument measures the axis in question and not a correlate.

**Budget, verified at head rather than from the body:** `test_annotation_coverage.py` 434 against an unchanged 579 row; `annotation_pins.py` 211, under the 400 target, **no row minted**; `annotation_gate.py` 379, unchanged and correctly not the destination; `docs/dev/file-budget.tsv` **not touched at all**. Leaving 579 stale-high rather than lowering it to 434 is the right call and for the right reason — lowering a row you do not need reds every un-rebased branch in the other lanes.

**Your three falsified cross-references are the finding of this PR**, above the extraction itself. A move that reads as pure relocation silently falsifies prose in the file it lands in, and no gate in this repo sees that class. The method you used — diff extracted-against-moved, and confirm every deleted line is present in the destination *before* cutting the source — is worth more than the slice.

**Merge condition:** at `9c78130`, 15 SUCCESS / 2 SKIPPED / `Shell tests` IN_PROGRESS. I re-read the rollup at the moment of merge and merge only on all-terminal, none-failed.

Issue #1611, opened 2026-08-30
env.redacted.txt keeps a wallet address under 90 chars: the KEY=value rule and the JSON rule disagree about the same field


VijitSingh97:
The second finding of the **#1593** audit, filed so that issue can close on its own scope. Measured
in https://github.com/p2pool-starter-stack/pithead/issues/1593#issuecomment-5467552695 and re-derived
at head for this report; the other finding is #1609.

`capture_artifacts` runs `cat .env` through `redact()` into `env.redacted.txt`, which
`release-gate.yml` uploads. On a `KEY=value` line only two rules can bite: the name rule, keyed on
`PROXY_AUTH_TOKEN|MONERO_NODE_PASSWORD|MONERO_NODE_USERNAME|.*_PASSWORD|.*_TOKEN|.*_SECRET`, which
matches no wallet spelling; and the `[A-Za-z0-9]{90,}` length rule. So the outcome is decided
entirely by how long the value happens to be:

| value on a `*_WALLET_ADDRESS=` line | result |
|---|---|
| Monero, 95 chars | redacted — by the length rule alone |
| Tari base58, 91 chars | redacted — by the length rule, one character to spare |
| Tari single, 48 chars | **survives verbatim** |
| Tari emoji, 67 glyphs | **survives verbatim** — the length rule's alphabet is alphanumeric |
| `MONERO_NODE_PASSWORD=` | redacted — positive control, the expression ran |

#1596's fix does not reach this: it keys on argv POSITION, and a `KEY=value` line has neither a
`--wallet` flag nor a `--merge-mine` URI. #1609's does not either — that one keys on IP scope.

## The sharper framing: two guards that drifted apart

This is not "a spelling is missing from an open list". **`redact()` already treats `wallet` as a
secret key name — in its other rule.** The JSON key-suffix alternation shipped as:

    (password|passwd|secret|token|login|username|key|wallet|wallet_address|ping_url|donor_id)

So the same field, carrying the same value, is classified two different ways by the same function
depending only on the syntax it arrives in. Measured at head with a synthetic value:

    {"wallet":"4Short…"}              ->  {"wallet":"<redacted>"}
    MONERO_WALLET_ADDRESS=4Short…     ->  MONERO_WALLET_ADDRESS=4Short…

That reframes the fix. Adding a wallet spelling to the `KEY=value` rule would be the
enumeration-of-names habit #1582 was found by **if the list were being extended on a guess** — but
[...]

Issue #1608, opened 2026-08-30
The file-budget monotonic ratchet is skipped on every CI run, and the gate's own NOTE says it never is


VijitSingh97:
`scripts/lint-file-budget.sh` enforces two independent things: a per-row ceiling ("this file is
over its recorded ceiling") and a monotonic ratchet against the base branch ("ceilings only ever go
down"). **The second one does not run in CI, on any run**, and the gate's own message says the
opposite.

## Measured

Every CI run reaches `check_monotonic`, fails to resolve a base ref, and skips it:

```
file-budget: NOTE — no base ref (origin/develop or develop) resolvable; skipping the
monotonic-ceiling check. This never happens in CI.
```

I read the `Lint per-surface` job log on four consecutive CI runs — `33300834176`, `33300709188`,
`33300536507`, `33299329353` — and the NOTE appears in all four, one occurrence each.

## Why

`lint-file-budget` runs in the `lint-surfaces` job (`.github/workflows/ci.yml:488`), whose
`actions/checkout` step at line 496 sets no `fetch-depth`. That is a depth-1, single-branch
checkout, so neither `develop-v2` nor `develop` exists as a ref in the working copy. `resolve_base_ref`
tries `origin/develop-v2`, `develop-v2`, `origin/develop` and `develop` in turn, finds none, and
returns the empty string — which is exactly the branch that prints the NOTE and returns early.

Two jobs in the same workflow already set `fetch-depth: 0` for precisely this reason (lines 50 and
383), so the fix is established practice in this file.

## Why it matters

"Ceilings only go down" is the ratchet the file budget rests on, and several standing review rulings
reason about how it behaves against a moved base. In CI it is not enforced at all: a change that
raises an existing ceiling passes `Lint per-surface` and is caught only if a human or lane happens to
run `make lint-file-budget` locally, in a checkout where the base refs are present.

The per-row check is unaffected and does run — it is what caught the live case that led me here, so
this is a partial hole rather than a dead gate.

## The NOTE text is the second half of the defect

`This never happens in CI.` is a false statement in durable in-repo text, and it is the specific
sentence that would stop the next reader investigating: someone who greps the failure, finds the
NOTE, and reads that reassurance concludes the skip is a local-only artifact. It should say what is
true — that a checkout without the base refs skips the ratchet — or the condition should be made
fatal so it cannot be skipped silently.

## Suggested fix

[...]

PR #1616, opened 2026-08-30
fix(integration): key KEY=value on the JSON rule's vocabulary — one list, not two that drifted (#1611)

VijitSingh97:
## What this fixes

`redact()` already treated `wallet` as a secret key name — **in its other rule**. On a `KEY=value`
line only two rules could bite: a name rule listing three uppercase spellings
(`.*_PASSWORD` / `.*_TOKEN` / `.*_SECRET`), and the `>=90`-character length rule. So the same field,
carrying the same value, was classified two ways depending only on the syntax it arrived in:

    {"wallet":"4Short…"}           ->  {"wallet":"<redacted>"}
    MONERO_WALLET_ADDRESS=4Short…  ->  MONERO_WALLET_ADDRESS=4Short…

That is the defect #1611 names: not a missing spelling, but two guards that drifted apart. So the fix
is not another spelling. The `KEY=value` rule is now keyed on the JSON rule's own suffix vocabulary —
one list to extend, rather than two to drift apart — and the self-test asserts mechanically that they
still agree.

## The risk this had to clear first, and the measurement

The `KEY=value` replacement takes the **whole rest of the line**, so widening what it matches could
strip a debugging value. Measured before writing the patch, then re-derived with the real shipped
function at base and at head over all 14 captured artifacts in two archived bundles:

| artifact | lines changed |
|---|---|
| `env.redacted.txt` | 10 per bundle |
| `logs.txt`, `status.txt`, `doctor.txt`, `config.json`, `api-state.json`, `compose-ps.txt` | **0** |

Zero, across ~307 KB of logs. The widening newly reaches exactly ten keys — the **same ten** in all
three populations checked (two bundles' `env.redacted.txt` plus the deployed `.env`, 129/127/129
keys). Every one of the ten is a field the same function already classifies as secret in JSON syntax,
so this applies the existing policy rather than inventing one.

Six of the ten still carried a raw value under the redactor as it stood — `MONERO_VIEW_KEY` (64 hex),
`HEALTHCHECKS_PING_URL`, `XVB_DONOR_ID`, `WALLET_RPC_USERNAME` and the two `DASHBOARD_ONION_CLIENT_*`
keys. The two wallet addresses in these bundles were already covered by the length rule, at 95 and 91
characters — the under-90 forms the issue names are real but do not occur in this corpus.

**A correction to my own first reading:** I nearly reported the two `DASHBOARD_ONION_CLIENT_*` values
as a leaking key pair. They are 11 characters, all-lowercase, and byte-identical to each other in both
bundles — a placeholder, not a key. Checked before claiming it.

## A second defect found on the way

The old prefix was `.*`, which is **greedy**: on a line carrying two secret keys it matched the *last*
one and left the first value in place.

    A_TOKEN=one B_TOKEN=two   ->   A_TOKEN=one B_TOKEN=<redacted>   (the first value survived)

The vocabulary prefix is `[A-Za-z0-9_]*`, which cannot cross a space, so the match starts at the first
key and the replacement takes the rest of the line with it. Pinned by its own case.

## More than the issue asked for, stated

Also shipped: #1590's suggestion 3, the **mechanical** version of an invariant that had been a comment
in `selftest-redact.sh`. Both of `redact()`'s suffix lists are now read out of `lib.sh` — never
restated in the test — and asserted to agree with each other and to sit inside the screen. It is here
rather than in a follow-up because this fix creates the second copy of the vocabulary; shipping the
copy without the guard would rebuild the drift the issue is about. That comment's superset claim was
already false once, at `wallet`, and was caught by review — which is what a check does not need.

Mutation battery, each leg `cmp`-checked to prove the file actually changed before any count was
trusted; restored byte-identical afterwards.

| mutation | result |
|---|---|
| revert the rule to its pre-#1611 form | 11 rows red |
| restore the greedy `.*` prefix | exactly the two-keys-on-one-line row reds |
| drop `WALLET`/`WALLET_ADDRESS` | the short and emoji Tari rows red — **the 91-char base58 row stays green**, which is the attribution stated in the file: the length rule carries that one, so it proves nothing about this rule |
| drop `PING_URL` | the ping-URL row reds |
| widen `PING_URL` to a bare `URL` | the over-redaction guard reds on a surviving public endpoint |
| drop `KEY` / `USERNAME` / `DONOR_ID` / `PING_URL` individually | exactly one row reds each (`KEY` reds two: its coverage row and the policy pin that shares the arm, labelled as such in the file) |
| make the two vocabularies disagree | the invariant reds, naming the differing entries |
| add a JSON suffix the screen cannot reach | the invariant reds, naming it |
| break the extraction itself | the invariant reds loudly rather than passing vacuously |

## What the over-engineering pass changed

It found a real fault in my own first draft, and the fix is in this PR rather than noted in it. The
`KEY=value` coverage loop asserted six keys — but `MONERO_VIEW_KEY`, `TARI_SPEND_PUBLIC_KEY` and
`DASHBOARD_ONION_CLIENT_PRIVKEY` all end in `KEY`, so all three exercise **one** alternation entry.
A single mutation (drop `KEY`) killed all three together, which is the tell: one measurement printed
three times, the same failure this lane recorded at cycle 46. The loop is now one row per vocabulary
entry `.env` actually exercises — four — and the other two corpus keys are named in the comment,
where a fact belongs that is not a discriminating assertion.

The pass also removed a row I had been about to add. `PASSWD` and `LOGIN` have no `.env` field, and
the sibling JSON file gives such entries a forward-cover case on the argument that an unexercised
pattern is one typo away from being dead. That argument no longer holds here: the vocabulary
invariant asserts this list *is* the JSON one, so a missing entry fails mechanically. The invariant
paid for itself by making three rows unnecessary.

Net: 15 rows, not 17, and every one attributable to the arm that kills it.

## Findings I considered and declined, with reasons

- **Preserving `<redacted-address>` for wallet keys.** The name rule runs first, so those lines now
  report the plain `<redacted>`. Nothing anywhere parses the address marker (grepped `.sh`/`.py`/`.md`/`.yml`);
  one vocabulary across both syntaxes is worth more than the hint that a value was an address. Pinned
  by a case so it reads as chosen.
- **Making the rule case-insensitive, matching the JSON side.** Declined: `.env` is generated by
  `pithead` and is uppercase throughout — #1590 measured all 129 keys — so it is load-bearing for
  nothing here, while a case-insensitive `key=.*` would eat the rest of any log line containing
  `key=`. Recorded as a labelled KNOWN GAP that fails and names the shape if a lowercase assignment
  ever arrives.
- **Adding a bare `URL` to the vocabulary.** The mutation above shows it strips a public endpoint.

## What I did NOT do

- No live capture. The corpus is two archived bundles plus the deployed `.env`, read-only.
- The under-90 Tari forms are covered by construction and by the self-test, not by a sighting in a
  bundle — they do not occur in this corpus.
- `selftest-redact.sh` now sits at **399 lines against the 400-line budget target**. One line of
  headroom; the next addition to it needs a budget row or a new file. That is why the `KEY=value`
  cases went into `selftest-redact-env.sh` rather than into it.

## Verification run

`shellcheck --severity=warning` over the whole `tests/integration` glob from the repo root, rc=0.
`shfmt -i 4 -d` over the Makefile's exact glob (191 files, list confirmed to contain both changed
shell files — the new file is only in it once staged), rc=0. `make lint-file-budget` **locally**,
which is the only run that exercises the monotonic ratchet (#1608): OK, with `lib.sh` at 692 against
its 692 ceiling — **line-neutral, the comment rewritten to the same four lines**. `make lint-md`
(0 errors), `make lint-docs-voice`, `make lint-topology`: all OK. All **16** files in the
`selftest*.sh` glob pass, the new one at 15 rows. `make lint-sh` was NOT run as a whole — it is this
box's OOM path and is serialized; the targeted substitute above is what ran.

## Shared files

`docs/dev/integration-testing.md` only (`docs/` is shared, no lane owns it). Its redactor-coverage
paragraph described the `KEY=value` rule by its three old suffixes, so it went stale with this change.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -65,12 +65,13 @@ The test box holds real synced nodes and real keys. Treat it as production-sensi
   comparing the hash, so the plaintext never crosses the wire, and every captured artifact passes
   through a redactor. **What that redactor covers is narrower than "artifacts are redacted"
   suggests**, and the gap is worth knowing before you trust a bundle. It reaches: a `KEY=value`
-  line whose name ends in `_PASSWORD` / `_TOKEN` / `_SECRET`; a credential given as a flag value;
-  a v3 onion; any opaque run of 90 or more characters (and a sha512 with them); the token after
-  `--wallet` and the one after `--merge-mine`'s URI, by POSITION; a JSON `"key": "value"` whose
-  key ends in one of a list of secret words — `password`, `token`, `key`, `username`, `wallet` and
-  the rest, matched without regard to case, so `apiKey` and `PASSWORD` are reached alongside
-  `api_key`; and a globally-routable IP address, by SCOPE. What it keeps is exactly the set
+  line and a JSON `"key": "value"` whose key ends in one of a SINGLE shared list of secret words —
+  `password`, `token`, `key`, `username`, `wallet`, `ping_url` and the rest — the JSON side matched
+  without regard to case, so `apiKey` and `PASSWORD` are reached alongside `api_key`, the
+  `KEY=value` side uppercase-only because `.env` is generated uppercase throughout; a credential
+  given as a flag value; a v3 onion; any opaque run of 90 or more characters (and a sha512 with
+  them); the token after `--wallet` and the one after `--merge-mine`'s URI, by POSITION; and a
+  globally-routable IP address, by SCOPE. What it keeps is exactly the set
   `is_public_ip` calls private — loopback, RFC1918, link-local and CGNAT — because an artifact whose
   port map and container addresses have been stripped cannot be triaged, and because holding the two
   lists identical is what makes the coverage argument checkable rather than anecdotal.
@@ -79,7 +80,13 @@ The test box holds real synced nodes and real keys. Treat it as production-sensi
   the JSON one; [#1596](https://github.com/p2pool-starter-stack/pithead/issues/1596) added the
   positional one; [#1609](https://github.com/p2pool-starter-stack/pithead/issues/1609) added the
   IP one, after an audit found `pithead doctor` writing the host's own public address into
-  `doctor.txt`. **The length rule never covered wallet addresses as widely as it read**: a
+  `doctor.txt`; and [#1611](https://github.com/p2pool-starter-stack/pithead/issues/1611) put the
+  `KEY=value` rule onto the JSON rule's vocabulary. Those two had DRIFTED, which is a sharper
+  failure than a missing spelling: the same field carrying the same value was classified two ways
+  depending only on the syntax it arrived in, so a wallet address below the length bar was
+  redacted in `config.json` and survived verbatim in `env.redacted.txt`. Keying one rule on the
+  other's list leaves one list to extend rather than two to drift apart, and the self-test asserts
+  mechanically that they still agree. **The length rule never covered wallet addresses as widely as it read**: a
   positional argument has no name to key on, so that rule was the ONLY one reaching the Tari
   address, and of the three Tari forms this repo accepts it reached only the base58 one — by a
   single character.
@@ -725,7 +732,14 @@ its port map stripped is useless for the triage it exists for. That timestamp is
 near-miss: HAProxy's `DD/Mon/YYYY:HH:MM:SS` reads as an IPv6 address, because a four-digit year
 opening with a 2 supplies the first hextet and the clock supplies the colon groups. It accounted
 for 278 of the 280 lines the first draft of the IP rule changed across two captured bundles, and
-`lint-topology` makes the same reading of it.
+`lint-topology` makes the same reading of it. That file also holds the vocabulary invariant: both
+of redact()'s suffix lists are read out of `lib.sh` and checked to agree with each other and to sit
+inside the screen, so the drift #1611 found cannot reopen quietly. `selftest-redact-env.sh` carries
+the `KEY=value` shape separately, including the three Tari address forms as `.env` renders them and
+the survivors that make a bundle worth keeping — a public endpoint, an auth mode string, a routing
+id. Its widening was measured before it shipped: across two archived bundles and the deployed
+`.env` it newly reaches ten keys, the same ten in all three populations, and changes nothing at all
+in the other six captured artifacts.
 `selftest-zmq-probe.sh` drives the ZMTP verdicts from captured and hand-built wire fixtures,
 so every failure class — a silent peer, a non-ZMQ listener, a ZMTP peer that is not a publisher, a
 READY frame carrying a decoy `Socket-Type` value — is reachable with no socket and no stack. It
--- a/tests/integration/lib.sh
+++ b/tests/integration/lib.sh
@@ -33,13 +33,13 @@ it_err() { echo -e "${IT_RED}[ITEST]${IT_RESET} $1" >&2; }
 it_step() { echo -e "${IT_DIM}  → $1${IT_RESET}"; }
 
 # --- Secrets hygiene --------------------------------------------------------
-# Redact before anything reaches a log or the terminal. FIVE shapes: KEY=value / --flag value by NAME;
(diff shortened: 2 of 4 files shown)


VijitSingh97:
## Non-author review — MERGE

Reviewed at `29cda5e2`, base `develop-v2` at `514ed880`. **This branch is not on the tip** (merge-base
`14f9209e`, tip two dashboard slices ahead), so every figure below is measured against the merge-base,
never against the tip — a two-way `diff tip..head` here shows slices 6 and 7 being *reverted* and is
not this PR's content. True content: **4 files, +180/-14**, exactly as the body states.

### The claim I re-derived, because the whole decision turns on it

A redactor change is judged on one thing: **can it fail open anywhere it previously closed?** I did not
take the self-test's word for this — a guard's own test is written from the guard's own list, so it is
green by construction. Instead I derived the population **mechanically**: every `KEY=` name this repo
actually renders into `.env`, harvested from `lib/pithead/*.sh` — **153 keys**, a superset of the 129
the body measured. Both *shipped* `redact()` bodies (merge-base and head) were extracted and run over
one line per key carrying an identical sentinel value, and the assertion is that the **raw value is
absent**, not that a marker appeared.

| | result |
|---|---|
| sentinel survives at merge-base | 143 of 153 |
| sentinel survives at head | 131 of 153 |
| **redacted at base but NOT at head** | **empty — no regression** |
| newly covered at head | 12 |
| positive control | base leg redacts 10 keys, so the instrument discriminates |

**Head is a strict superset of base over the whole rendered vocabulary.** That is the merge condition,
and it holds. The newly covered set is the wallet/view-key/onion-privkey/ping-url/donor-id family the
body names.

One scope note, not a defect: the body's "ten keys, the same ten in all three populations" is scoped to
the two bundles plus the deployed `.env`, and says so. Over the broader rendered vocabulary the figure
is twelve — the two extra are the `MINOTARI_WALLET_*` key forms, which those corpora do not contain.
Both numbers are right about their own population.

### The rest, verified

- **Greedy-prefix fix is real.** `.*` could cross a space and match the *last* key on a line;
  `[A-Za-z0-9_]*` cannot. Pinned by its own case.
- **The two vocabularies now agree**, 11 entries each, and the invariant asserts it mechanically rather
  than in a comment — the right call, since this change is what creates the second copy.
- **Self-tests green**: `selftest-redact.sh` 72 rows, `selftest-redact-env.sh` 15 rows, and **all 16
  files in the `selftest*.sh` glob pass on the merged tree**, not only at the branch head.
- **Budget, both legs** (ruling 8a — this branch is *not* on the tip, so one run would not answer both):
  branch head **rc 0**, merged tree **rc 0**. `merge-tree --write-tree` rc 0, no conflict. `lib.sh`
  692 against its 692 ceiling — legal, zero headroom, line-neutral as claimed. The tsv did not move
  between merge-base and tip, so ruling 8's stale-high hazard does not bite here; I checked rather than
  assumed.
- **Docs** accurate against the shipped regex, including the uppercase-only asymmetry stated as such.
- The declined findings are each declined for a stated, checkable reason. The case-insensitivity
  refusal is right: a case-insensitive `key=.*` eats any log line containing `key=`.

### A gap this PR does not open and does not close — filing separately

Measured on the same run, at head, **both** rules (name and JSON — they share one vocabulary now) leave
these `.env` names unreached, and their real values are short enough that the 90-character shape rule
cannot be relied on to catch them either: `DASHBOARD_AUTH_HASH_B64`, `DASHBOARD_AUTH_PW_FP`,
`XMRIG_API_AUTH`, `DASHBOARD_ONION_CLIENT_AUTH`, `NOTIFY_WEBHOOK_URLS`. The first is an offline-crackable
password hash. **This is pre-existing and unchanged by this PR** — it is the same leak class as the
already-open issues on the other two redactors, keyed to a spelling rather than to what a field holds.
Not a blocker; filed as its own issue.

### What I did NOT do

I ran neither full suite and have never run `lint-sh` on this box (its OOM path); the check figures are
CI's, read at the moment of merge. I did not do a live capture. The under-90 Tari forms are covered by
construction and by the self-test, not by a sighting in a bundle — the body says so.

Issue #1614, opened 2026-08-30
Nothing asserts the size of PINNED/_ANCHORS, so shrinking the pin set silently narrows the #1487 gate with every assertion green


VijitSingh97:
`dashboard/tests/service/annotation_pins.py` holds the `PINNED` / `_ANCHORS` / `_UNJUDGED_AND_READ`
data the #1487 annotation gate iterates. **Nothing asserts how many entries those structures have.**
Removing one narrows what the gate's laws cover, and every assertion in
`test_annotation_coverage.py` stays green — the suite reports success over a smaller question.

Filed by the controller, not by the lane that found it: it is pre-existing, it is past the slice
that surfaced it, and #1612's author correctly declined to build a guard inside that slice.

## Why it is worth a guard now, when it was tolerable before

The gap is not new — the same tuples had no size assertion while they lived inside the test file.
What changed with #1612 is **legibility**: a data-only module reads as "just data" and invites the
casual edit that a block sitting inside a test file does not. The risk moved without the code moving.

It also compounds with the program shape. #1556 lands in slices, each adding entries; the population
only grows, so the distance between "the gate covers everything" and "the gate covers what is left"
widens with every slice, and nothing measures it.

## What a guard has to do, and the trap it must avoid

**A hardcoded count is the obvious guard and it is the wrong one.** Every slice would have to bump
it, so it becomes a number people update to make the suite pass — the failure mode the file-budget
ratchet already documents. It also cannot tell a deliberate addition from a silent deletion, which
is the only distinction that matters here.

Two directions that do not have that problem:

**A — assert the population against its SOURCE, not against a number.** The pins name modules under
`dashboard/mining_dashboard/`; a guard that enumerates the tree and asserts every judged module is
present states the actual invariant. It grows by itself, and it fails on a deletion.

**B — monotonicity, in the ratchet's own idiom.** Assert the set never shrinks between base and head.
Cheaper than A and catches the exact defect; it does not catch a module that was never pinned.

A is the honest one; B is the one that fits in a slice. **Whichever is taken, the assertion must be
shown able to FAIL** — seed a removal, confirm red, restore — because a size guard that has not been
fired is indistinguishable from no guard at all.

## Verified at source before filing

The move itself is clean, so this issue is about the residue and not about #1612. Comparing the base
test file's constants against the new module by AST value, with a negative control: **zero base
entries lost across all three structures, order preserved**, and the four added entries in `PINNED`
[...]

PR #1617, opened 2026-08-30
fix(dashboard): #1556 slice 7 — the ten handle-guard procedures in storage_service.py

### LINKED ISSUE #1615: Dashboard: the per-table write-health channel reports healthy while every write is being dropped
## What

either health stamp:

| writer | module |
|---|---|

```
```

VijitSingh97:
#1556 slice 7: `service/storage_service.py`. Ten unannotated failure returns become `-> None`, and
the module joins `PINNED` and `_ANCHORS`. Stacked on #1612, rebased onto `develop-v2` after it
merged.

## The population, re-derived rather than carried

My handoff said nine. Nine is the HANDLE-GUARD count; the module holds **ten** blind functions —
`_prune_quarantined` comes through the `except` door — and pinning it needs all ten. That is the
figure the slice was built on, measured at this head.

| door | functions |
|---|---|
| handle guard | `load`, `update_history`, `add_share`, `add_event`, `add_share_stats`, `add_block`, `add_worker_history`, `update_xvb_stats`, `set_kv` |
| `except` | `_prune_quarantined` |

Package-wide, blind goes **25 -> 15** functions across **7** modules; `PINNED` goes 25 -> 26 of the
33 modules that hold a failure return at all.

## What the slice actually achieves — stated plainly

All ten are genuine procedures, so the slice adds **ZERO to `signed` and ten to `procedure`.** It
fixes nothing. What the pin asserts for this module is that #1487's rule structurally cannot apply
to these ten — a procedure has no success value for a failure to hide inside — and that this is now
SAID rather than silent, which is what #1581 added the sixth verdict for. Same shape as slice 4,
which added zero to `signed` and was the correct outcome.

The module's **eight `collapse` rows are untouched and stay flagged**, as does its residue.

## PROVEN BY ME

**They really are procedures.** Zero valued returns and zero yields across all ten, walked with the
gate's own `_own_nodes` rather than a re-spelled predicate. Positive control: `get_kv`, which the
same walk reports with three valued returns. `add_worker_history` has two `return`s — the guard and
an `if not rows` early exit — both valueless.

**Annotation-only.** All ten code objects compare identical base-vs-head on `co_code`, `co_names`,
`co_varnames`, `co_freevars`, `co_cellvars`, `co_argcount`, `co_flags` and `co_consts`, recursively.
Never a repr — a `dis.dis()` output diff scored 4-of-6 differing on this lane's annotation-only edits
once, and the whole delta was the object address. The comparator's core is lifted byte-identically
from slice 5a's, not retyped. **Positive control:** seeding one real behaviour change
(`str(value)` -> `str(value).strip()` in `set_kv`) flips exactly that function to False and leaves
the other nine True — so it fires *and* discriminates.

**Three controls, each fired for its own law and scoped to this module alone.** Each armed with an
exact-match applier that refuses unless the target site matches exactly once, with a before/after
site count as the readback, and each restored by `cmp` against a sha256'd copy — never
`git checkout`, which has silently eaten uncommitted work in this lane.

| control | seeded | fired |
|---|---|---|
| A | UNANNOTATE `add_block` (undo the slice) | law 1, `[service/storage_service.py]`, alone |
| B | re-sign it `-> bool` (bare `return` now collapses) | law 2, same scope, alone |
| C | point its `_ANCHORS` value at a missing function | the vacuity guard, same scope, alone |

Control B matters beyond ceremony: the module holds no `unjudged` row, so law 2 looks vacuous here.
It is not — the reachable regression is someone making a writer report success, and law 2 reds on it.

**No caller consumes a value from any of the ten.** Every grep hit was `json.load(f)` or a JS
`this.load()` — a different subject per site, classified individually rather than counted.

**Ran:** `test_annotation_coverage.py` + `test_collapsed_return_channels.py` (112 pass);
`make test-dashboard` (**2355 passed**, coverage 97.17%); `lint-py`, `lint-file-budget`,
`lint-docs-voice`, `lint-operator-strings`, `lint-topology`, `lint-md` — all green, re-run at the
rebased head rather than reported from before the base move.

**Patch coverage:** 0 of 10 changed lines uncovered, measured from `coverage.xml` directly (its
`<source>` is `dashboard/mining_dashboard`), not from the gate's rc. Three of the ten are
continuation lines of multi-line signatures, which coverage attributes to the `def` — so diff-cover
measures seven and all seven are hit.

**The base move did not move the program:** `git rev-parse <ref>:dashboard` is
`6b2271de4fb8d83418ddef2ba8096f24f4109f64` at the old base, the squash, and the `develop-v2` tip.

## Counting prose — three sentences were already stale, and two were spelled in words

Re-derived every count in the files I edited instead of carrying it, which is how these surfaced.
No gate in this repo sees this class.

- `25 of 33` -> **26 of 33**.
- `**25 failure returns are still blind**` -> **15**, in seven modules.
- `Eight other pinned modules hold the same residue` — **present tense, false for two slices**. Now
  past tense and scoped to #1604, with a note saying why it is written that way.
- `The residue is now MEASURED for all twenty-one` -> "for every pinned module", phrased so no slice
  can leave it stale again.
- `test_annotation_coverage.py`: `thirteen of the twenty-five` -> **fourteen of the twenty-six**.

**Classified and deliberately LEFT:** `test_annotation_coverage.py:320`'s "scored all twenty-one
modules identically" is a true statement about what #1604's first sweep did when `PINNED` held 21.
A count word can belong to a different subject; this one does.

## The finding — #1615

`add_block` and `add_worker_history` stamp the per-table write-health signal, and the handle guard
returns **before** either stamp. After a failed corruption recovery `_conn` stays `None` for good,
so every telemetry write is dropped while `get_table_health()` still reports `healthy: True` — the
exact case its own docstring says it exists to surface. Reproduced by driving that path, not
inferred from the shape of the code.

Bounded honestly on the issue: `db_healthy`/`db_unrecoverable` are correctly False/True, so the
global badge is right, and `get_table_health()` has no production caller today. Latent, not live.

## What I did NOT do

- **Did not fix #1615.** It is a behaviour change to five write paths across two modules and wants
  its own review; fixing it here would also cost this slice its bytecode-identity proof.
- **Did not hedge the two source comments that state that guarantee** (`storage_service.py:161-166`
  and `:926-929`). Named on #1615 with line numbers so the fix cannot leave them standing — this
  lane has shipped a doc arguing FOR a defect before. The module is also at **zero budget headroom**
  (1358 against a 1358 ceiling), so any rewrite there has to be line-for-line.
- **Did not touch `telemetry_store.py`** — its three handle-guard blind returns are slice 8.
- **No security-reviewer pass.** The change touches no control channel, auth, credential or
  outbound fetch; it is ten signatures, proven bytecode-identical.
- **Nothing asserts `PINNED`'s size** — pre-existing, and now filed by the controller as #1614.

## Over-engineering pass (done on merits, then the gate satisfied)

The gate blocked once and cleared on the retry, so it verifies nothing — the pass is reported here
because it was run, not because the gate says it was.

It found one real defect, in my own new prose: the slice-7 comment opened "the first slice whose
whole population came through the HANDLE GUARD" and then said "nine of its ten" in the same
sentence. Self-contradicting, and the kind of sentence that reads as authoritative. Rewritten to
say what is true — nine through the guard, one handler, ten in total — and to name the guard-door
miscount explicitly, since that is the trap the next slice inherits.

Also checked and deliberately kept: the 25-line reading is in this file's established register
(the argument travels with the entry, and `_UNJUDGED_AND_READ`'s existing comments carry method at
the same length). The one clause that looked redundant against the past-tense note below it —
"one of them spelled in WORDS, where a digit-keyed sweep misses" — carries the actionable half that
the instance site does not, so it stays.

### DIFF
--- a/dashboard/mining_dashboard/service/storage_service.py
+++ b/dashboard/mining_dashboard/service/storage_service.py
@@ -275,7 +275,7 @@ class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin):
             self.db_unrecoverable = True
             self._db_error("DB Recovery Error", e)
 
-    def _prune_quarantined(self):
+    def _prune_quarantined(self) -> None:
         """Keep only the newest ``_CORRUPT_KEEP`` ``<db>.corrupt-*`` files; delete older ones."""
         directory = os.path.dirname(self.db_path) or "."
         base = os.path.basename(self.db_path) + ".corrupt-"
@@ -493,7 +493,7 @@ class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin):
             self.logger.info("Migrating DB: Adding type column to worker_config")
             self._conn.execute("ALTER TABLE worker_config ADD COLUMN type TEXT DEFAULT 'apply'")
 
-    def load(self):
+    def load(self) -> None:
         """
         Loads state from SQLite into memory on startup.
         """
@@ -593,7 +593,7 @@ class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin):
 
     def update_history(
         self, hashrate: float, p2pool_hr: float = 0, xvb_hr: float = 0, windows=None
-    ):
+    ) -> None:
         """Appends a new hashrate data point to the history buffer.
 
         ``windows`` (Issue #168) is an optional ``{window: (p2pool_hr, xvb_hr)}`` mapping of the
@@ -677,7 +677,7 @@ class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin):
         except sqlite3.Error as e:
             self._db_error("History Update Error", e)
 
-    def add_share(self, ts: float, difficulty: float):
+    def add_share(self, ts: float, difficulty: float) -> None:
         """Appends a new share to history and persists it to the DB."""
         with self._lock:
             # Check if share already exists to prevent duplicate in-memory appends
@@ -724,7 +724,7 @@ class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin):
         with self._lock:
             return list(self.state.get("shares", []))
 
-    def add_event(self, ts: float, event_type: str, detail: str = ""):
+    def add_event(self, ts: float, event_type: str, detail: str = "") -> None:
         """Record a chart event marker (#99) — a degradation/recovery point the chart draws and the
         history window prunes, mirroring shares. Persisted so it survives a dashboard restart."""
         with self._lock:
@@ -757,7 +757,7 @@ class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin):
 
     def add_share_stats(
         self, ts: float, accepted: int = 0, rejected: int = 0, invalid: int = 0, expired: int = 0
-    ):
+    ) -> None:
         """Record one poll's pool-wide share-health DELTAS (#116) — how much each cumulative
         proxy counter advanced since the previous poll — in memory and in the DB. Mirrors
         add_event: retention-pruned in memory, probabilistically pruned on disk."""
@@ -928,7 +928,7 @@ class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin):
         the data-service poll loop is one big try/except, so nothing else would surface that."""
         return {k: dict(v) for k, v in self.table_health.items()}
 
-    def add_block(self, ts: float, height: int, difficulty: float):
+    def add_block(self, ts: float, height: int, difficulty: float) -> None:
         """Record one pool block-found event (#196): permanent (no retention prune — a handful of
         rows/week, like payouts). The caller (DataService) reuses `_shares_to_record` on p2pool's
         cumulative blocks_found counter so this fires exactly once per genuinely NEW block; a
@@ -1029,7 +1029,7 @@ class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin):
             self.logger.error(f"Raffle Win Read Error: {e}")
             return []
 
-    def add_worker_history(self, rows: list[dict[str, Any]]):
+    def add_worker_history(self, rows: list[dict[str, Any]]) -> None:
         """Record one poll's per-worker hashrate/share-count samples (#196) in a SINGLE
         executemany call — the caller batches every online worker for one wall-clock tick rather
         than issuing N inserts per cycle. 30-day retention. Each row needs ts/name/h15/accepted/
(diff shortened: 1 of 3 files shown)


VijitSingh97:
**Non-author pass: APPROVED, merging on terminal green.** Reviewed at `e270b79`. I re-derived the two claims this merge turns on rather than relaying them; everything else I took from CI and say so below.

**RE-DERIVED BY ME AT SOURCE**

1. **All ten are genuine `-> None` procedures.** AST walk over the shipped head, per function, not descending into nested defs: zero valued returns and zero yields across all ten. The walk was controlled — `get_kv` reports valued returns at lines 1287, 1291 and 1294, so it can see one when one exists. This is the claim that would have cost most if wrong: a false `-> None` is an annotation the gate then blesses, and it fails silently.

2. **The change is genuinely annotation-only.** I stripped every return and parameter annotation from base and head and compared the resulting ASTs — **identical**. The control fired: the raw unstripped ASTs differ, so the comparison is not vacuously true over two copies of one file. That is stronger than reading the diff, and it is what backs your bytecode-identity claim independently.

3. **`PINNED` and `_ANCHORS` are both 26 entries**, matching the "26 of 33" prose. The ten-not-nine correction reproduces: nine come through the handle guard and `_prune_quarantined` through `except`.

**RELAYED, NOT RE-DERIVED — check before building on it:** the 16 passing checks and the suite figures are CI's, not mine; I ran no suite. I did not re-derive "15 still blind in seven modules" or the "fourteen of the twenty-six" in the residue docstring.

**On those prose counts, a note rather than a finding.** `test_the_residue_in_every_pinned_module_is_reported_without_being_ruled_on` computes sites, functions and holding-modules dynamically and deliberately asserts nothing about them, which is the right design and is stated in its own docstring. So the mechanism prints the truth even when the surrounding prose drifts. That leaves the docstring's own "fourteen of the twenty-six" as the one figure in this PR with no instrument behind it — the exact class this slice's prose warns about two paragraphs earlier, and the class #1614 names. Not a blocker and not worth a commit on its own; worth knowing that the module joining `PINNED` is machine-enforced while the sentence counting them is not.

**Not asked for and not required:** #1615 was correctly filed rather than fixed here. Keeping an annotation-only slice annotation-only is the right call, and it is what let me prove point 2 above; a behaviour change to five write paths in the same commit would have destroyed that proof. I have labelled it `bug,dashboard`.

Slice 8 (#1618) is stacked on this branch and I will confirm GitHub retargets it to `develop-v2` before deleting this branch, rather than assuming it.

PR #1618, opened 2026-08-30
fix(dashboard): #1556 slice 8 — the three telemetry_store writers, closing the handle-guard population

### LINKED ISSUE #1615: Dashboard: the per-table write-health channel reports healthy while every write is being dropped

VijitSingh97:
#1556 slice 8: `service/telemetry_store.py`. Three unannotated failure returns become `-> None`, and
the module joins `PINNED` and `_ANCHORS`. **Stacked on #1617** (slice 7) — based on that branch, so
GitHub retargets this to `develop-v2` when #1617 merges. Review #1617 first.

**This closes the handle-guard population.** All twelve failure returns still blind come through the
`except` door, in six modules holding **exactly two each**: `helper/utils.py`, `alert_service`,
`clearnet_sync`, `control_service`, `telegram_commands`, `web/server`.

## What it does

Slice 7's shape with nothing new to argue: `add_xvb_history`, `add_network_history` and
`add_disk_growth` are the same closed-handle guard whose whole body is a bare `return`. Three more
`procedure` rows, **ZERO to `signed`**. The module's three `collapse` rows are untouched and stay
flagged. `PINNED` 26 -> 27 of the 33 modules that hold a failure return at all; blind 15 -> 12.

## PROVEN BY ME

**They really are procedures** — measured, not assumed off slice 7's result: zero valued returns and
zero yields across all three, walked with the gate's own `_own_nodes`. Positive control:
`get_xvb_history`, which the same walk reports with three valued returns.

**Annotation-only.** All three code objects compare identical base-vs-head on `co_code`, `co_names`,
`co_varnames`, `co_freevars`, `co_cellvars`, `co_argcount`, `co_flags` and `co_consts`, recursively —
never a repr. **Positive control:** seeding one real behaviour change (`INSERT` -> `INSERT OR IGNORE`
in `add_disk_growth`) flips exactly that function to False and leaves the other two True.

**Three controls**, each armed with an exact-match applier that refuses unless the target site
matches exactly once, readback by before/after site count, each restored by `cmp` against a sha256'd
copy — never `git checkout`.

| control | seeded | fired |
|---|---|---|
| A | UNANNOTATE `add_xvb_history` | law 1, `[service/telemetry_store.py]`, alone |
| B | re-sign it `-> bool` | law 2, same scope, alone |
| C | point its `_ANCHORS` value at a missing function | the vacuity guard, same scope, alone |

**Ran:** `make test-dashboard` — **2358 passed**, coverage 97.14%. The +3 over #1617's 2355 is fully
attributed: the three laws are parametrized over `PINNED`, so one added module adds exactly three
tests, and no test was written. `lint-py`, `lint-file-budget`, `lint-md`, `lint-docs-voice` green.
Pins module 273 lines, under the 400 target with **no budget row** — correct, since under 400 *with*
a row is a gate failure.

## The one case worth reading

This module's residue and its blind set were the **same three sites**. The gate saw them through the
guard door; #1604's sweep saw them as falsy returns from functions declaring nothing — two
instruments answering different questions about one site. Annotating cleared both at once, so the
module now reports **0 residue sites**. That zero is printed *because* the module is pinned, which
is the whole of why #1604 prints the clean modules too: an unmentioned module and a measured-clean
module are otherwise the same silence.

## Over-engineering pass (done on merits, then the gate satisfied)

The gate blocked once and cleared on the retry, so it verifies nothing. The pass is reported because
it was run — and it caught a real defect in my own new prose, the second cycle running it has paid.

I had written that the overlap above was "its own case" — an **unqualified uniqueness claim I never
measured**, the defect this lane keeps shipping. Measured: it is **not** unique. `helper/utils.py`
has the same shape (2 residue functions, the same 2 blind) and is not pinned yet, so the slice that
takes it should expect the same. Everywhere else the two sets are disjoint — `storage_service.py`'s
residue is `add_shares` and `save_snapshot`, neither among the ten slice 7 annotated. The comment
now says that, and says not to infer one set from the other.

Nothing else was cut: the reading is the length this file's convention asks for (the argument travels
with the entry), and the anchor note records that all three writers are equivalent and
`add_xvb_history` is simply the first declared — written down so nobody hunts for a reason.

## What I did NOT do

- **Did not fix #1615**, of which these three writers are half — the guard returns before the
  per-table health stamp here exactly as it does in `storage_service.py`. The reading says so; the
  fix is a behaviour change across two modules and would cost this slice its bytecode-identity proof.
- **Did not update docs.** `docs/dashboard.md` describes the mechanism and states no count this slice
  moves; swept mechanically across the whole repo, not just `docs/`.
- **No security-reviewer pass** — three signatures, no control channel, auth, credential or outbound
  fetch, proven bytecode-identical.

### DIFF
--- a/dashboard/mining_dashboard/service/telemetry_store.py
+++ b/dashboard/mining_dashboard/service/telemetry_store.py
@@ -43,7 +43,7 @@ class TelemetryStoreMixin:
         fail_count: int = 0,
         donation_fraction: float = 0.0,
         mode: str = "",
-    ):
+    ) -> None:
         """Record one XvB-scalars sample (#196), 30-day retention. The caller wall-clock-gates
         this to ~5 min (DataService._sync_xvb_stats) so the cadence survives an UPDATE_INTERVAL
         change, and only calls it on a genuine XvB fetch (never on a failed one)."""
@@ -91,7 +91,7 @@ class TelemetryStoreMixin:
         height: int = 0,
         reward: float = 0.0,
         pool_hashrate: float = 0.0,
-    ):
+    ) -> None:
         """Record one hourly network-stats sample (#196): Monero difficulty/height/reward plus the
         pool's own hashrate. 90-day retention. DB-only (nothing reads this per-cycle)."""
         try:
@@ -136,7 +136,7 @@ class TelemetryStoreMixin:
         monero_db_bytes: int = 0,
         disk_used_gb: float = 0.0,
         disk_total_gb: float = 0.0,
-    ):
+    ) -> None:
         """Record one hourly disk-growth sample (#196): monerod's DB size plus host disk usage.
         Permanent (no retention prune — tiny, ~24 rows/day; it's a capacity trend line)."""
         try:
--- a/dashboard/tests/service/annotation_pins.py
+++ b/dashboard/tests/service/annotation_pins.py
@@ -41,14 +41,16 @@ moved the data, which is the failure the pin comment below records happening twi
 """
 
 # The modules with ZERO unannotated failure returns, measured over live source. Each slice of #1556
-# adds the module it finished. Measured at this tip: 26 of 33 modules that hold a failure return at
+# adds the module it finished. Measured at this tip: 27 of 33 modules that hold a failure return at
 # all — `client/xvb_client.py` by slice 1, `service/worker_config_store.py` by pre-existing work,
 # the six single-function `client/` modules by slice 2, the four single-function `web/` modules by
 # slice 3, the two single-function `config/` modules by slice 4, the four outbound senders under
 # `service/` by slice 5a, and the last three `service/` singletons by slice 5b. Slice 6 is the first
 # MULTI-function slice — the four parse-or-read pairs under `service/`, each pinned only once BOTH of
 # its failure returns were annotated and read, and the ten handle-guard procedures in
-# `service/storage_service.py` by slice 7. **15 failure returns are still blind**, in seven modules.
+# `service/storage_service.py` by slice 7, and its three siblings in `service/telemetry_store.py` by
+# slice 8. **12 failure returns are still blind**, in six modules — every one through the `except`
+# door, because slice 8 closed the handle-guard population entirely.
 #
 # Every figure above is re-derived at the head being shipped, never carried across a slice: 5b moved
 # this tuple from 18 to 21 and left the sentence beside it saying 18. A slice adds rows to a tuple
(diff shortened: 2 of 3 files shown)


VijitSingh97:
Rebased onto `develop-v2` after #1617 merged — the stack's slice-7 commit was already in the
base as its squash, so the branch reapplied it and went `CONFLICTING`.

**Dropped exactly one commit, and the evidence for that is tree identity, not the message.**
My slice-7 commit and the squash-merged one have the SAME parent (`203661a`) and
BYTE-IDENTICAL trees (`47909e3d8dbeb1be043d36f682de1c13e3f20f04`), so dropping mine loses
nothing.

**The suite evidence carried forward, and here is why that is not a stale baseline.** The
rebased head's tree is `d9bcf946506c0065f7ff3609599afa55bd86fa3e` — byte-identical to the
pre-rebase head's tree. The program the suite runs did not change, so the prior counts
describe this exact artifact rather than a superseded one.

RE-RUN AT THIS HEAD BY ME: `ruff check` and `ruff format --check` (clean). NOT re-run:
the full suite — justified by the tree identity above, not by assumption. CI is the
independent leg.

The diff is now slice 8 alone: `telemetry_store.py`, `annotation_pins.py`,
`test_annotation_coverage.py`. `storage_service.py` correctly dropped out of it.

VijitSingh97:
## Non-author review — MERGE

Reviewed at `b25f6b8f`. Its parent was `514ed880` at the time of review; `develop-v2` has since moved to
`bc2664d2` (#1616), which touches only `tests/integration/` and one shared doc, so the two are disjoint.
Both instruments below were run by me at source, each with a control that fired.

### 1. The three newly `-> None` functions really are procedures

AST-walked with the gate's own `_own_nodes`, so a nested def cannot be folded into its parent:

| function | annotation | valued returns | yields |
|---|---|---|---|
| `add_xvb_history` | `None` | 0 | 0 |
| `add_network_history` | `None` | 0 | 0 |
| `add_disk_growth` | `None` | 0 | 0 |
| **`get_xvb_history`** (positive control) | `list[dict[str, Any]]` | **3**, at lines 75/82/85 | 0 |

The control fired, so the walk can see a valued return where one exists. Zero across all three targets.

### 2. Annotation-only

Annotation-**stripped** ASTs compared base vs head, both normalised through `ast.unparse` first:

- **Control — raw ASTs differ**: 16,397 B vs 16,487 B. Without this the comparison is vacuous.
- **Result — stripped ASTs byte-identical**: 15,013 B both sides.
- **Negative control on the instrument itself** — seeding a real body change
  (`INSERT` → `INSERT OR IGNORE` in `add_disk_growth`, applied and asserted to have applied) still
  differs after stripping, so the instrument can see a behaviour change through the strip.

### 3. Every durable in-repo figure re-derived, not relayed

These live in committed comment text, so under ruling 2 a wrong one would earn a discharge commit.
Measured at the PR head against the real package:

- `PINNED` **26 → 27**, `_ANCHORS` **26 → 27**, and the two sets are **equal**.
- **Blind failure returns: 12, across 6 modules, exactly two each** — `helper/utils.py`,
  `service/alert_service.py`, `service/clearnet_sync.py`, `service/control_service.py`,
  `service/telegram_commands.py`, `web/server.py`. The comment's list matches name for name.
- `service/telemetry_store.py` residue: **`[]`** — the "0 residue sites" claim holds.
- `storage_service.py` residue is **`add_shares`, `save_snapshot`**, exactly as the comment says.
- **14 of 27** pinned modules hold residue, so `test_annotation_coverage.py`'s "fourteen of the
  twenty-seven" is right, and the numerator correctly did *not* move.
- The anchor resolves: `service/telemetry_store.py:add_xvb_history`.

**The over-engineering pass's own correction is itself true.** It retracted an unmeasured uniqueness
claim and replaced it with a claim about `helper/utils.py`; measured, that module's blind set and its
residue set are the *same two functions* (`is_ip_address`, `detect_host_ipv4`). Correct as written.

### 4. Gates

`make lint-file-budget` **locally**, which is the only run that exercises the monotonic ratchet (#1608
— CI skips it): **branch head rc 0** and **merged tree rc 0**. Ruling 8a needs both because the base
moved out from under this branch after review; at review time the branch sat directly on the tip and
one run answered both, and I am saying which case each figure is from rather than reporting one run as
two. `merge-tree --write-tree` rc 0, no conflict. The tip's move did not touch
`docs/dev/file-budget.tsv`, so ruling 8's stale-high hazard does not bite. No budget row is added, and
none is needed: `annotation_pins.py` is 273 lines and `telemetry_store.py` 171, both under the 400
target, while `test_annotation_coverage.py` at 434 keeps its existing 579 row.

### What I did NOT do

I ran neither full suite; the 2358-passed and 97.14% figures are the author's, and CI's 18 checks were
read terminal and green at the moment of merge. I did not re-derive the coverage percentage. I did not
verify #1615's fix is unnecessary — the PR correctly declines it and says why, and I agree it would cost
this slice its identity proof.

Issue #1613, opened 2026-08-30
redact()'s sentinel restore runs after redaction, so a pre-existing \x01 can reconstitute a public IP the redactor never saw


VijitSingh97:
`redact()`'s IP rule protects reserved ranges with an `\x01` sentinel and restores it in the **last**
expression of the pipeline — after the redaction step. So any `\x01` already present in the input is
turned into a `.` *after* the only step that could have inspected the result.

Introduced by #1610, which closed #1609 correctly and is the right fix; this is the residue of its
implementation shape, not an argument against it.

## Measured, with a fired control on the same run

Real `redact()` extracted from `tests/integration/lib.sh` at `58e3dc9` and run over three lines:

| input | output |
|---|---|
| `203.0.113<SOH>47` | `203.0.113.47` — **a globally-routable quad emerges from the redactor** |
| `A<SOH>B` | `A.B` — silent corruption of an artifact whose value is fidelity |
| `198.51.100.9` | `<redacted-ip>` — **positive control: the expression ran** |

(`203.0.113.0/24` and `198.51.100.0/24` are RFC 5737 documentation ranges, used here for the same
reason #1610's own fixtures use them.)

## What #1610 already says, and the one step past it

#1610 discloses the protect-then-redact-then-restore shape and measures **`\x01` absent from all 14
files of the corpus**. That measurement is honest and it is the mitigation currently carrying this.
What is not stated is the failure *mode* when one does appear: not a corrupted byte, but a dotted
quad assembled after redaction. Its mutation table also covers `delete the sentinel restore` (11
cases red), which pins that the restore is load-bearing — it does not pin what the restore does to a
byte it did not write.

## Severity — latent and contrived, stated up front

Nothing in this stack emits a control byte in place of a dot inside an address, so the leak arm is
effectively unreachable. The corruption arm is the reachable one and it is cosmetic. **This is a
gap to close on the next edit to that expression, not something to hold a release for.**

## Fix directions, ranked

**A — neutralise before protecting. Recommended and cheapest.** Strip or escape any pre-existing
`\x01` in the first expression, before the protect step runs. Costs one `s///` and no line if it
joins an existing `-e`. It makes the sentinel an invariant of the pipeline rather than an assumption
about the input, which is the property #1610's corpus measurement is currently standing in for.

**B — pin the assumption instead of removing it.** Keep the design and add a `selftest-redact.sh`
[...]

PR #1619, opened 2026-08-30
fix(integration): neutralise a stray protect sentinel before the IP rule runs (#1613)

VijitSingh97:
Closes the ordering hazard #1613 describes, by its recommended direction A.

⚠️ **Stacked on #1616** (base is that branch, not `develop-v2`). Both changes edit the same two lines
of `redact()` — the four-line header comment and the second `-e` — so an independent branch would
conflict by construction. GitHub retargets this to `develop-v2` when #1616 merges. Review #1616 first.

## The defect, reproduced at head before the fix

The IP rule protects reserved ranges by mangling one dot to a `\x01` sentinel and restores it in the
**last** expression of the pipeline — after the only step that could inspect the result. Verbatim,
with a positive control on the same run:

| input | output at head | |
|---|---|---|
| `203.0.113<SOH>45` | `203.0.113.45` | a globally-routable quad assembled **after** redaction |
| `A<SOH>B` | `A.B` | silent corruption of an artifact whose value is fidelity |
| `198.51.100.9` | `<redacted-ip>` | positive control — the expression ran |

`od -c` confirmed the SOH was really in the input rather than assumed.

## The fix

Neutralise any pre-existing sentinel in the first expression, before the protect step, so the only
`\x01` the restore can see is one the protect step wrote. That makes the sentinel an **invariant of
the pipeline** rather than an assumption about the input — which is what #1610's "absent from all 14
corpus files" measurement was standing in for.

It replaces with a visible `<ctrl>` rather than deleting. Deleting would trade one silent corruption
for another (`A<SOH>B` → `AB`, equally plausible-looking); `<ctrl>` cannot form a quad, cannot extend
an alphanumeric run, and tells a reader something was there.

`lib.sh` is **line-neutral at 692 against its 692 ceiling** — the expression appends inside an
existing `-e`, and the header comment was recompressed to the same four lines (max width 103).

## The IP cases moved, and the move is proven

`selftest-redact.sh` was at **399 lines against the 400-line target**, so it had no room. The IP
section moved verbatim to `selftest-redact-ip.sh`; `selftest-redact.sh` is now 323, with headroom
again. **Proof the move changed nothing:** the sorted multiset of every row label printed by the
whole `selftest*.sh` glob is byte-identical before and after — 791 rows both sides, `diff` empty.
Row counts alone would not have caught a reworded label; the label multiset does.

This is more than the issue asked for. It is here because the budget forced a new file either way,
and leaving the sentinel cases in a different file from the rule they test is the drift shape #1611
was about.

Each leg `cmp`-checked to prove the file actually changed before any count was trusted; `lib.sh`
restored byte-identical afterwards.

| mutation | result |
|---|---|
| delete the neutralisation | the 2 sentinel rows red |
| move it to just before the restore (so it eats the protect step's own sentinel) | 10 rows red, including every reserved-address survivor — this is what pins the ORDER rather than the presence |

A third mutation I ran turned out to be textually the same edit as the first; it is not reported as a
second data point.

## A finding against my own first draft

I wrote three new rows and all three died to the single `delete the neutralisation` mutation — the
redundancy tell from #1616's review, applied here to my own diff. The third re-asserted the quad
claim the first already makes, so its `case` row is gone; its two `assert_contains` survive that
mutation and red under the mis-ordered one, which is what earns them. The reason is recorded in the
file, not just here. Net 21 rows.

## What I did NOT do

- The leak arm stays contrived, exactly as the issue says: nothing in this stack emits a control byte
  in place of a dot. This closes a latent hazard on the next edit to that expression, which is what
  the issue asked for — it is not a response to an observed leak.
- No live capture; no bench.

## Verification run

`shellcheck --severity=warning` over the whole `tests/integration` glob from the repo root, rc=0.
`shfmt -i 4 -d` over the Makefile's exact glob, rc=0, with the file list confirmed to contain all
three `selftest-redact*` files. `make lint-file-budget` **locally** (the only run that exercises the
monotonic ratchet, #1608): OK. `make lint-md` 0 errors, `make lint-docs-voice` OK, `make
lint-topology` OK. All **17** files in the `selftest*.sh` glob pass. Sizes: `lib.sh` 692/692,
`selftest-redact.sh` 323, `selftest-redact-ip.sh` 130, `selftest-redact-env.sh` 130 — all under the
400-line target, so no shared budget-file edit.

## Shared files

`docs/dev/integration-testing.md` only — its redactor description now states the neutralisation and
names the new file.


VijitSingh97:
## Non-author review — MERGE. I rebased this branch first; here is why, and the proof it changed nothing.

### The blocker I found, which no PR field showed

At the time I picked this up, GitHub reported this PR `MERGEABLE` / `CLEAN`. **It was not.** Once
#1616 merged as a **squash**, `develop-v2` carried that content as a *new object* while this branch
still carried the *original* `023be0e6` / `29cda5e2` commits. The three-way merge then saw both sides
editing `tests/integration/lib.sh` from the same merge-base:

```
git merge-tree --write-tree <develop-v2> <this branch>   ->  rc 1
CONFLICT (content): tests/integration/lib.sh
CONFLICT (content): docs/dev/integration-testing.md
```

**This is the standing lesson, live: a PR field is not a measurement.** `merge-tree` is, and it
disagreed with the flag. GitHub also did not auto-retarget this PR when its base merged — I retargeted
it to `develop-v2` by hand via the API and re-read the result rather than assuming.

### The rebase, and why you can trust it did not alter your work

Rebased `--onto` the tip, dropping the two already-landed #1616 commits and replaying only the two
#1613 commits. Verified by **ruling 15** — across a moved base a rebase holds the *patch* fixed, never
the tree, so a tree comparison would report a false positive here:

| | |
|---|---|
| intended patch (`29cda5e2..b708b850`) | **22,249 B** |
| rebased patch (`develop-v2..b819b019`) | **22,249 B**, byte-identical |
| negative control (an unrelated patch) | differs — the comparison discriminates |
| trees, old head vs new head | **differ**, exactly as ruling 15 predicts |

Pushed with `--force-with-lease` pinned to the sha I had read, and I read the push *output* rather
than its rc. File contents are untouched: `selftest-redact-ip.sh` is byte-identical before and after
the rebase. CI then re-ran on the new head — run `33305346523`, all 18 checks terminal and green,
read at the moment of merge.

### The defect, reproduced at the merged base with a control

I did not take the body's table on trust; I ran both shipped `redact()` bodies over the same input,
with `od -c` confirming the SOH was really there:

| input | at base | at head |
|---|---|---|
| `203.0.113<SOH>45` | **`203.0.113.45`** — a globally-routable quad assembled *after* redaction | `203.0.113<ctrl>45` |
| `A<SOH>B` | **`A.B`** — silent corruption | `A<ctrl>B` |
| `198.51.100.9` | `<redacted-ip>` | `<redacted-ip>` — positive control, ran on both sides |

The fix is correctly placed: neutralisation is the **first** expression in the second `-e`, so the
only `\x01` the restore can see is one the protect step wrote. That is what makes the sentinel an
invariant rather than an assumption about the input. `<ctrl>` over deletion is the right call — it
cannot form a quad, and I checked the case that worried me: it cannot lengthen an alphanumeric run
either, since `<` and `>` break the run on both sides, so it cannot disturb the 90-character rule.

### The move lost nothing — measured over the whole glob, not just the moved file

Sorted multiset of every row label printed by the entire `selftest*.sh` glob, at the merged base and
at this head:

- base **791** rows, head **795** rows
- **labels present at base and absent at head: none** — nothing was lost
- added: exactly four, all sentinel cases (`a stray sentinel cannot reconstitute a routable quad`,
  `a stray sentinel is not silently turned into a dot`, and the two IP rows beside them)

My instrument differs in scope from the body's: yours compared across the *move*, mine compares base
against the *final* head, so it also covers the rows this PR adds. Both agree.

`selftest-redact-ip.sh` reports 21 rows, matching "Net 21". All 17 files in the glob pass.

### Gates

`make lint-file-budget` **locally**, the only run that exercises the monotonic ratchet (#1608):
**rc 0**. After the rebase this branch sits directly on the tip, so merged tree == branch tree and one
run answers both legs — ruling 8a's narrow case, stated so this is not read as two runs. `lib.sh`
**692 against its 692 ceiling**, line-neutral as claimed; `selftest-redact.sh` 323. No tsv edit, and
the tip's move did not touch the tsv either.

### One figure to correct, which changes nothing

The body says `selftest-redact-ip.sh` is **130** lines. It is **133** — and it was 133 at your original
head too, so my rebase did not cause it; 130 is `selftest-redact-env.sh`'s figure carried across to its
sibling. The conclusion it supports ("all under the 400-line target") is unaffected, and under ruling 2
a loose figure in a PR body is a merge-record defect corrected in the verdict, never force-pushed. Noted
here only so the next reader does not re-derive it as a discrepancy.

### What I did NOT do

I ran neither full suite and have never run `lint-sh` on this box (its OOM path); the 18 check results
are CI's, read terminal at the moment of merge. No live capture, no bench. I did not independently
re-run the author's two mutations — those are the author's, relayed; what I re-derived is the defect
itself, the fix's placement, the losslessness of the move, and the rebase.

PR #1620, opened 2026-08-30
fix(dashboard): #1556 slice 9 — helper/utils.py's two valued except returns, one signed and one read

VijitSingh97:
Slice 9 of #1556: `helper/utils.py`. Two failure returns, both through the `except` door.
The production diff is **two signature lines** — everything else is pin data and the arguments
that travel with it.

```
-def is_ip_address(value):            +def is_ip_address(value) -> bool:
-def detect_host_ipv4():              +def detect_host_ipv4() -> str | None:
```

## Why slices 7 and 8 transfer nothing here

The discriminator is NOT the door — slice 7's `_prune_quarantined` came through the `except` door
too. It is that those thirteen were all bare-`return` procedures, measured at zero valued returns
and zero yields, where `-> None` is simply honest. BOTH of these return a VALUE, so each had to be
read on its own terms — and the two landed in DIFFERENT verdicts. One `signed`, one `unjudged`,
ZERO `procedure`.

- **`detect_host_ipv4` -> `str | None` is a true `signed`.** The docstring already declared the
  contract in prose ("Returns `None` when it can't be determined (e.g. no default route)"), and
  its sole production caller `web/views.py:host_display_addr` acts on exactly that. The annotation
  moves a promise the code already kept into the signature; it invents nothing.
- **`is_ip_address` -> `bool` is `unjudged`, and is added to `_UNJUDGED_AND_READ` with its
  reading.** Its `except` handler is not an error path at all: `ipaddress.ip_address` says "not an
  address" by RAISING, so the handler is where the negative answer is COMPUTED. `AttributeError`
  is the same answer by a second route (a non-string is not an address either). Both handled cases
  mean one thing, so `-> bool | None` would invent a return the function never makes.

## PROVEN BY ME, at the head being shipped

- **Verdicts MEASURED, not predicted from the decision chain.** After annotation `classify` reports
  `signed: 1`, `unjudged: 1` (`['False']`), `blind: 0`, `procedure: 0` for this module.
- **Bytecode-identical base-vs-head** for `is_ip_address`, `detect_host_ipv4` and a bystander
  (`format_hashrate`), comparing `co_code`/`co_names`/`co_varnames`/`co_consts` recursively — never
  a `dis` repr. **Positive control fired:** a seeded `return True` -> `return False` inside
  `is_ip_address` flipped that function and ONLY that function.
- **Three controls, one per structure touched, each armed at a verified single site and restored by
  `cmp` against a sha256'd copy** — never `git checkout`:
  | control | seeded | reddened |
  |---|---|---|
  | A | UNANNOTATE the anchor | law 1 `..._is_unannotated[helper/utils.py]`, alone |
  | B | DE-SIGN it to `-> str` | law 2 `..._is_unjudged[helper/utils.py]`, alone |
  | C | corrupt its `_ANCHORS` value | vacuity guard `..._is_actually_in_the_walk[helper/utils.py]`, alone |
  Each fired for its OWN law and only for the module seeded. B is the one that looks vacuous in a
  module with no `unjudged` row and is not; here the module HAS one, and B still discriminates.
- **Residue went to 0 sites**, as the pins file predicted before this slice existed. That
  prediction is now discharged in the file rather than left standing as a prediction.
- **The `-> bool` contract was attacked, not assumed.** `'3232235876'`, `'123'`, `'0x7f000001'`,
  `''`, `None`, `42`, `b'1.2.3.4'` — none slips through as `True`, none escapes as an exception.

## Two figures this slice found stale, one of them NOT its own fault

The repo-wide counting-prose sweep (needles derived mechanically from the changed names, digits
AND words, with a firing control on the pattern) found three falsified sentences:

1. `annotation_pins.py` — "every one of the **twelve** failure returns still blind" (spelled in
   WORDS). Rewritten to be scoped to slice 8's own head in the past tense, so a later slice cannot
   falsify it again; the live figure lives in the header.
2. `test_annotation_coverage.py` — "fourteen of the **twenty-seven** hold". The *fourteen* is still
   correct; only the denominator moved. Changed the denominator alone.
3. `test_annotation_coverage.py` — "INCLUDING the **twelve** that score zero". **This one was
   already stale before this slice.** Measured at the pre-slice state by restoring the PRE copies
   and re-running: 27 pinned = 14 holding + **13** zero. So it was off by one at the head I started
   from and is now fourteen. Reported rather than quietly corrected, because it means the figure
   drifted without any slice noticing.

## A generalisation in the pins file is RETRACTED, with the measurement

The file said "in every other module the two sets are DISJOINT" — blind vs #1604 residue — citing
`storage_service.py`. That was true of the module it named and **false as the generalisation it
read as**, because the two halves were measured over different populations: in a PINNED module
`blind` is empty by construction, so "disjoint" there is vacuous rather than a finding.

Re-measured over the five modules that still hold a blind return: **the sets are disjoint in not
one of them.** `blind` is a SUBSET of the residue in all five, and in `web/server.py` they are the
SAME SET (`_finalize_worker_upgrade`, `_num`) — a third module of the shape that paragraph called
unique and then called doubled. Mechanism: a blind function is unannotated by definition, so a
falsy failure return lands it in both instruments at once; only a blind function returning a
NON-falsy value can separate them. The retraction is in the file, in place, not deleted.

## Counts, re-derived here and not carried

| | pre-slice | shipped |
|---|---|---|
| pinned modules (of 33 holding a failure return) | 27 | **28** |
| blind failure returns | 12 | **10** |
| modules holding a blind return | 6 | **5** |
| pinned modules holding residue / scoring zero | 14 / 13 | **14 / 14** |

## What I did NOT do

- **No security-reviewer pass.** `detect_host_ipv4` opens a UDP socket, which normally triggers
  this lane's mandatory security lens — but the change is provably behaviour-free (bytecode
  identity above, positive control fired), so there is no new surface to review. Stated rather
  than skipped silently; say so if you want it run anyway.
- **No doc change.** `docs/dashboard.md`'s account of the pin is deliberately count-free ("several
  pinned modules"), so this slice falsifies nothing in it. Checked, not assumed.
- **No new test.** The three laws are parametrized over `PINNED`, so pinning the module adds three
  test instances by construction. That is the established design; a bespoke test would duplicate
  it. This is the whole of the suite delta — 2358 -> 2361 reconciles as one module x three laws.
- **The `b'\x01\x02\x03\x04'` packed-bytes case** would return `True` from `is_ip_address`, but
  `dashboard.host` comes from JSON and is never bytes, so it is unreachable. Not recorded in the
  pin, because the entry is scoped to the paths that exist at this head.

### DIFF
--- a/dashboard/mining_dashboard/helper/utils.py
+++ b/dashboard/mining_dashboard/helper/utils.py
@@ -299,7 +299,7 @@ def resolve_target_threshold(tiers, stable_hr, donation_level, max_fraction):
     return target, sustainable
 
 
-def is_ip_address(value):
+def is_ip_address(value) -> bool:
     """True if ``value`` is a literal IP address (IPv4 or IPv6), False for a hostname.
 
     Used to decide whether the configured ``dashboard.host`` already *is* an address, in
@@ -312,7 +312,7 @@ def is_ip_address(value):
         return False
 
 
-def detect_host_ipv4():
+def detect_host_ipv4() -> str | None:
     """Best-effort primary LAN IPv4 of the host this dashboard runs on.
 
     Opens a UDP socket toward an off-link sentinel address so the kernel selects the source
--- a/dashboard/tests/service/annotation_pins.py
+++ b/dashboard/tests/service/annotation_pins.py
@@ -41,7 +41,7 @@ moved the data, which is the failure the pin comment below records happening twi
 """
 
 # The modules with ZERO unannotated failure returns, measured over live source. Each slice of #1556
-# adds the module it finished. Measured at this tip: 27 of 33 modules that hold a failure return at
+# adds the module it finished. Measured at this tip: 28 of 33 modules that hold a failure return at
 # all — `client/xvb_client.py` by slice 1, `service/worker_config_store.py` by pre-existing work,
 # the six single-function `client/` modules by slice 2, the four single-function `web/` modules by
 # slice 3, the two single-function `config/` modules by slice 4, the four outbound senders under
@@ -49,8 +49,9 @@ moved the data, which is the failure the pin comment below records happening twi
 # MULTI-function slice — the four parse-or-read pairs under `service/`, each pinned only once BOTH of
 # its failure returns were annotated and read, and the ten handle-guard procedures in
 # `service/storage_service.py` by slice 7, and its three siblings in `service/telemetry_store.py` by
-# slice 8. **12 failure returns are still blind**, in six modules — every one through the `except`
-# door, because slice 8 closed the handle-guard population entirely.
+# slice 8, and the two `except`-door functions in `helper/utils.py` by slice 9. **10 failure
+# returns are still blind**, in five modules — every one through the `except` door, because slice 8
+# closed the handle-guard population entirely.
 #
 # Every figure above is re-derived at the head being shipped, never carried across a slice: 5b moved
 # this tuple from 18 to 21 and left the sentence beside it saying 18. A slice adds rows to a tuple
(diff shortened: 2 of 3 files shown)


VijitSingh97:
Closed by GitHub — not by me — when #1618 merged and this PR's base branch (`fix/1556-slice8-telemetry-store`) was deleted. GitHub retargets a stacked PR only while the base still exists; once the base is deleted at close time, `reopenPullRequest` is refused permanently and the base cannot be changed while closed. I confirmed both, including after restoring the base branch, so this is not recoverable in place.

Superseded by #1622 — the same commit rebased onto `develop-v2`, patch byte-identical.

PR #1622, opened 2026-08-30
fix(dashboard): #1556 slice 9 — helper/utils.py's two valued except returns, one signed and one read

VijitSingh97:
Slice 9 of #1556: `helper/utils.py`. Two failure returns, both through the `except` door.
The production diff is **two signature lines** — everything else is pin data and the arguments
that travel with it.

```
-def is_ip_address(value):            +def is_ip_address(value) -> bool:
-def detect_host_ipv4():              +def detect_host_ipv4() -> str | None:
```

## Why slices 7 and 8 transfer nothing here

The discriminator is NOT the door — slice 7's `_prune_quarantined` came through the `except` door
too. It is that those thirteen were all bare-`return` procedures, measured at zero valued returns
and zero yields, where `-> None` is simply honest. BOTH of these return a VALUE, so each had to be
read on its own terms — and the two landed in DIFFERENT verdicts. One `signed`, one `unjudged`,
ZERO `procedure`.

- **`detect_host_ipv4` -> `str | None` is a true `signed`.** The docstring already declared the
  contract in prose ("Returns `None` when it can't be determined (e.g. no default route)"), and
  its sole production caller `web/views.py:host_display_addr` acts on exactly that. The annotation
  moves a promise the code already kept into the signature; it invents nothing.
- **`is_ip_address` -> `bool` is `unjudged`, and is added to `_UNJUDGED_AND_READ` with its
  reading.** Its `except` handler is not an error path at all: `ipaddress.ip_address` says "not an
  address" by RAISING, so the handler is where the negative answer is COMPUTED. `AttributeError`
  is the same answer by a second route (a non-string is not an address either). Both handled cases
  mean one thing, so `-> bool | None` would invent a return the function never makes.

## PROVEN BY ME, at the head being shipped

- **Verdicts MEASURED, not predicted from the decision chain.** After annotation `classify` reports
  `signed: 1`, `unjudged: 1` (`['False']`), `blind: 0`, `procedure: 0` for this module.
- **Bytecode-identical base-vs-head** for `is_ip_address`, `detect_host_ipv4` and a bystander
  (`format_hashrate`), comparing `co_code`/`co_names`/`co_varnames`/`co_consts` recursively — never
  a `dis` repr. **Positive control fired:** a seeded `return True` -> `return False` inside
  `is_ip_address` flipped that function and ONLY that function.
- **Three controls, one per structure touched, each armed at a verified single site and restored by
  `cmp` against a sha256'd copy** — never `git checkout`:
  | control | seeded | reddened |
  |---|---|---|
  | A | UNANNOTATE the anchor | law 1 `..._is_unannotated[helper/utils.py]`, alone |
  | B | DE-SIGN it to `-> str` | law 2 `..._is_unjudged[helper/utils.py]`, alone |
  | C | corrupt its `_ANCHORS` value | vacuity guard `..._is_actually_in_the_walk[helper/utils.py]`, alone |
  Each fired for its OWN law and only for the module seeded. B is the one that looks vacuous in a
  module with no `unjudged` row and is not; here the module HAS one, and B still discriminates.
- **Residue went to 0 sites**, as the pins file predicted before this slice existed. That
  prediction is now discharged in the file rather than left standing as a prediction.
- **The `-> bool` contract was attacked, not assumed.** `'3232235876'`, `'123'`, `'0x7f000001'`,
  `''`, `None`, `42`, `b'1.2.3.4'` — none slips through as `True`, none escapes as an exception.

## Two figures this slice found stale, one of them NOT its own fault

The repo-wide counting-prose sweep (needles derived mechanically from the changed names, digits
AND words, with a firing control on the pattern) found three falsified sentences:

1. `annotation_pins.py` — "every one of the **twelve** failure returns still blind" (spelled in
   WORDS). Rewritten to be scoped to slice 8's own head in the past tense, so a later slice cannot
   falsify it again; the live figure lives in the header.
2. `test_annotation_coverage.py` — "fourteen of the **twenty-seven** hold". The *fourteen* is still
   correct; only the denominator moved. Changed the denominator alone.
3. `test_annotation_coverage.py` — "INCLUDING the **twelve** that score zero". **This one was
   already stale before this slice.** Measured at the pre-slice state by restoring the PRE copies
   and re-running: 27 pinned = 14 holding + **13** zero. So it was off by one at the head I started
   from and is now fourteen. Reported rather than quietly corrected, because it means the figure
   drifted without any slice noticing.

## A generalisation in the pins file is RETRACTED, with the measurement

The file said "in every other module the two sets are DISJOINT" — blind vs #1604 residue — citing
`storage_service.py`. That was true of the module it named and **false as the generalisation it
read as**, because the two halves were measured over different populations: in a PINNED module
`blind` is empty by construction, so "disjoint" there is vacuous rather than a finding.

Re-measured over the five modules that still hold a blind return: **the sets are disjoint in not
one of them.** `blind` is a SUBSET of the residue in all five, and in `web/server.py` they are the
SAME SET (`_finalize_worker_upgrade`, `_num`) — a third module of the shape that paragraph called
unique and then called doubled. Mechanism: a blind function is unannotated by definition, so a
falsy failure return lands it in both instruments at once; only a blind function returning a
NON-falsy value can separate them. The retraction is in the file, in place, not deleted.

## Counts, re-derived here and not carried

| | pre-slice | shipped |
|---|---|---|
| pinned modules (of 33 holding a failure return) | 27 | **28** |
| blind failure returns | 12 | **10** |
| modules holding a blind return | 6 | **5** |
| pinned modules holding residue / scoring zero | 14 / 13 | **14 / 14** |

## What I did NOT do

- **No security-reviewer pass.** `detect_host_ipv4` opens a UDP socket, which normally triggers
  this lane's mandatory security lens — but the change is provably behaviour-free (bytecode
  identity above, positive control fired), so there is no new surface to review. Stated rather
  than skipped silently; say so if you want it run anyway.
- **No doc change.** `docs/dashboard.md`'s account of the pin is deliberately count-free ("several
  pinned modules"), so this slice falsifies nothing in it. Checked, not assumed.
- **No new test.** The three laws are parametrized over `PINNED`, so pinning the module adds three
  test instances by construction. That is the established design; a bespoke test would duplicate
  it. This is the whole of the suite delta — 2358 -> 2361 reconciles as one module x three laws.
- **The `b'\x01\x02\x03\x04'` packed-bytes case** would return `True` from `is_ip_address`, but
  `dashboard.host` comes from JSON and is never bytes, so it is unreachable. Not recorded in the
  pin, because the entry is scoped to the paths that exist at this head.

---

**Supersedes #1620**, which GitHub CLOSED (not retargeted) when #1618 merged and its base branch
was deleted. Reopening is refused permanently in that state — `reopenPullRequest` fails, and the
base cannot be changed while closed, even after restoring the base branch, which I tried. Same
commit, rebased onto `develop-v2`.

**Rebased across a base that MOVED, so the checks are different from the last one.** `bc2664d2`
landed between slices 7 and 8, so my copy of slice 8 and its squash have different parents and
therefore **different trees** — a tree comparison is the wrong instrument here. Compared PATCHES
instead: byte-identical, with a control (slice 9's own patch vs slice 8's) confirming that
differing patches do report as differing. After the rebase, my slice's own patch is byte-identical
to its pre-rebase form.

**No re-run owed, and here is the reason rather than an assumption:** `git rev-parse
<base>:dashboard` is the SAME hash at both bases (`279f46a0…`), and `bc2664d2` touched only
`docs/` and `tests/integration/`. The program the dashboard suite runs did not change, so the
2361-pass count still describes this artifact.

### DIFF
--- a/dashboard/mining_dashboard/helper/utils.py
+++ b/dashboard/mining_dashboard/helper/utils.py
@@ -299,7 +299,7 @@ def resolve_target_threshold(tiers, stable_hr, donation_level, max_fraction):
     return target, sustainable
 
 
-def is_ip_address(value):
+def is_ip_address(value) -> bool:
     """True if ``value`` is a literal IP address (IPv4 or IPv6), False for a hostname.
 
     Used to decide whether the configured ``dashboard.host`` already *is* an address, in
@@ -312,7 +312,7 @@ def is_ip_address(value):
         return False
 
 
-def detect_host_ipv4():
+def detect_host_ipv4() -> str | None:
     """Best-effort primary LAN IPv4 of the host this dashboard runs on.
 
     Opens a UDP socket toward an off-link sentinel address so the kernel selects the source
--- a/dashboard/tests/service/annotation_pins.py
+++ b/dashboard/tests/service/annotation_pins.py
@@ -41,7 +41,7 @@ moved the data, which is the failure the pin comment below records happening twi
 """
 
 # The modules with ZERO unannotated failure returns, measured over live source. Each slice of #1556
-# adds the module it finished. Measured at this tip: 27 of 33 modules that hold a failure return at
+# adds the module it finished. Measured at this tip: 28 of 33 modules that hold a failure return at
 # all — `client/xvb_client.py` by slice 1, `service/worker_config_store.py` by pre-existing work,
 # the six single-function `client/` modules by slice 2, the four single-function `web/` modules by
 # slice 3, the two single-function `config/` modules by slice 4, the four outbound senders under
@@ -49,8 +49,9 @@ moved the data, which is the failure the pin comment below records happening twi
 # MULTI-function slice — the four parse-or-read pairs under `service/`, each pinned only once BOTH of
 # its failure returns were annotated and read, and the ten handle-guard procedures in
 # `service/storage_service.py` by slice 7, and its three siblings in `service/telemetry_store.py` by
-# slice 8. **12 failure returns are still blind**, in six modules — every one through the `except`
-# door, because slice 8 closed the handle-guard population entirely.
+# slice 8, and the two `except`-door functions in `helper/utils.py` by slice 9. **10 failure
+# returns are still blind**, in five modules — every one through the `except` door, because slice 8
+# closed the handle-guard population entirely.
 #
 # Every figure above is re-derived at the head being shipped, never carried across a slice: 5b moved
 # this tuple from 18 to 21 and left the sentence beside it saying 18. A slice adds rows to a tuple
@@ -103,8 +104,13 @@ moved the data, which is the failure the pin comment below records happening twi
 # three writers are the same closed-handle guard whose whole body is a bare `return`, measured the
 # same way (zero valued returns, zero yields, with `get_xvb_history`'s three valued returns as the
 # control that the walk can see one). Three more `procedure` rows, ZERO to `signed`. It closes the
-# HANDLE-GUARD population — every one of the twelve failure returns still blind now comes through
-# the `except` door.
+# HANDLE-GUARD population — at slice 8's own head every one of the twelve failure returns still
+# blind came through the `except` door. That figure is deliberately left scoped to that head rather
+# than updated in place: slice 9 took two of them and the live count is in the header above. A
+# past-tense sentence about a shipped slice cannot be falsified by a LATER slice — but it can still
+# be wrong at the head it describes, and slice 9 found exactly that in the sibling test file, where
+# "the twelve that score zero" had drifted to thirteen before any slice noticed. Scoping a figure
+# to its head protects it from the future, not from the measurement.
 #
 # One thing here is worth carrying: this module's residue and its blind set were the SAME THREE
 # SITES. The gate saw them through the guard door, and #1604's sweep saw them as falsy returns from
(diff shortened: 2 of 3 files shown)


VijitSingh97:
## Non-author review — MERGE

Reviewed at `180a0b6d`. **Its parent is `6af57de0` (#1618), while `develop-v2` is now `742ef689`
(#1619) — the base MOVED under this branch after it was opened.** That is not a defect here (#1619
touched only `docs/` and `tests/integration/`, disjoint from this diff), but it changes which
instruments are valid, so I say which case each figure is from rather than reporting one run as two.

Every figure below was re-derived by me at source, base and head, in two throwaway worktrees. Each
instrument carries a control that fired.

### 0. The diff is what the body says it is

Measured against `git merge-base`, never a two-way diff against the tip: **3 files, and the
production half is exactly the two signature lines.** No tsv row is touched.

### 1. Annotation-only

Annotation-**stripped** ASTs, base vs head, both normalised through `ast.unparse`:

- **Control — raw ASTs differ**: 30,814 B vs 30,939 B. Without this the comparison is vacuous.
- **Result — stripped ASTs byte-identical**: 10,986 B both sides.
- **Negative control on the instrument itself** — seeding a real body change (`return True` ->
  `return False` in `is_ip_address`, asserted to have applied) still differs after stripping, so
  the instrument can see a behaviour change through the strip.

### 2. Both annotations are true of the bodies — walked, then attacked

Return sites via the gate's own `_own_nodes`, so a nested def cannot fold into its parent:

| function | annotation | return sites | yields |
|---|---|---|---|
| `is_ip_address` | `bool` | `True`, `False` | 0 |
| `detect_host_ipv4` | `str \| None` | `s.getsockname()[0]`, `None` | 0 |
| `format_hashrate` (control) | — | **5 valued** | 0 |
| `resolve_target_threshold` (control) | — | **1 valued** (a tuple) | 0 |

The AST cannot prove `s.getsockname()[0]` is a `str`, so I did not stop there:

- **`is_ip_address`, 19 adversarial inputs** — `'3232235876'`, `'123'`, `'0x7f000001'`, `''`,
  `'  '`, `'256.256.256.256'`, `None`, `42`, `3.14`, `True`, `b'1.2.3.4'`, `b'\x01\x02\x03\x04'`,
  `bytearray(...)`, a list, a dict, a bare `object()`, an `IPv4Address`. **Every one returned
  `type(...) is bool`; none escaped as an exception.** The packed-bytes case the body discloses
  does return `True`, and it returns a `bool`, so it is a question about the ANSWER and not about
  the annotation — the disclosure is placed correctly.
- **`detect_host_ipv4`, both legs live** — success returned a dotted-quad `str` (the value is this box's own LAN address, so it is withheld), and with
  `connect` patched to raise `OSError` it returned `None`. `getsockname()[0]` is a `str` for
  `AF_INET`, confirmed at runtime rather than assumed.
- **PEP 604 at import time**: the module has no `from __future__ import annotations`, so
  `str | None` is evaluated at def time. `requires-python = ">=3.11"`, the image is
  `python:3.11-slim`, and production already carries **22** `-> X | None` sites. The attack dies.

### 3. Verdicts and every durable figure, measured with the gate itself

`classify` on this module — **head: `signed` 1, `unjudged` 1 (`['False']`), `blind` 0, `procedure`
0.** At the base both functions are `blind`. Exactly as the body states.

| | base | head |
|---|---|---|
| `PINNED` / `_ANCHORS` | 27 / 27, sets **equal** | 28 / 28, sets **equal** |
| blind failure returns | 12 | **10** |
| modules holding a blind return | 6 | **5** |
| pinned modules holding residue / scoring zero | 14 / **13** | 14 / **14** |
| `helper/utils.py` residue | `is_ip_address`, `detect_host_ipv4` | **`[]`** |

14 + 14 = 28, so the residue split is total over `PINNED`.

**#1618's prediction is DISCHARGED, not accepted.** Its pin comment predicted this module's residue
would go to zero because its blind set and its residue set were the same two functions. I measured
that at the base (they are) and at the head (`[]`). It was treated as a prediction this slice had to
satisfy, never as a figure read out of the PR body.

**The "already stale" report is TRUE and I re-derived it.** At the pre-slice head, pinned modules
scoring zero was **13**, while the comment said "twelve". So that figure had drifted before this
slice touched it, exactly as reported, and 14 is right at the shipped head.

### 4. The retraction in the pins file is correct as rewritten

Re-measured over all five modules that still hold a blind return: **the sets are disjoint in NOT ONE
of them; `blind` is a SUBSET of the residue in all five; and in `web/server.py` they are the SAME
SET** (`_finalize_worker_upgrade`, `_num`). The retracted generalisation was indeed false.

Two adjacent claims also checked, because both are the kind that read as true and are not:
- **"the SECOND module to hold both verdicts"** — exactly two modules hold both, and at the base
  only `service/worker_config_store.py` did. Correct, and correct that it is not the first.
- **The anchor argument.** `_ANCHORS` names `detect_host_ipv4` on the ground that `is_ip_address` is
  already held alive elsewhere. That guard exists and does what the comment says: it asserts
  `_UNJUDGED_AND_READ <= unjudged`, which requires every entry to still exist AND still score
  `unjudged`. Anchoring on `is_ip_address` really would guard one function twice and leave the
  `signed` one on law 1's aggregate alone. The anchor is in the right place.

### 5. Callers read, not assumed

`is_ip_address` has exactly one production caller (`web/views.py:211`) and `detect_host_ipv4`
exactly one (`web/views.py:213`) — both inside `host_display_addr`, whose body is
`if is_ip_address(host): return None`, then `addr = detect_host_ipv4()`. That is the shape the pin
describes, word for word. No caller branch exists that a third answer could reach.

### 6. Gates

`ruff check` / `ruff format --check` clean. `test_annotation_coverage.py`, `tests/helper/test_utils.py`
and `tests/web/test_views.py` pass at this head. `merge-tree --write-tree` **rc 0**.

**`make lint-file-budget` at the BRANCH HEAD rc 0 AND on the MERGED TREE rc 0.** Ruling 8a needs
both because the base moved out from under this branch; this is not the narrow same-tree case.

**And I proved the ratchet was armed rather than reading its rc.** My first control fired for the
WRONG REASON — it rc-1'd on "no tsv entry" because my seed dropped a trailing newline and welded two
rows together, which is the failure ruling 14 names: an rc is one bit over several doors. Redone on
a row the base actually carries, with an arming readback (87 parseable rows, unchanged), it fired on
the specific line: *"raises `scripts/release.sh`'s ceiling from 851 to 901. Ceilings only go down."*
The base ref resolves to `origin/develop-v2` and the ratchet sees **87 of 87** rows.

### What I did NOT do

- **I ran neither full suite.** The 2361-passed figure is the author's; CI's **16 SUCCESS, 2 SKIPPED**
  (both the shipped-image CVE sweep) were read terminal at the moment of merge.
- I did not re-derive the coverage percentage, and I did not reproduce the author's bytecode-identity
  instrument — I ran annotation-stripped AST identity instead, which answers the same question by a
  different route, with its own controls.
- I did not verify the `b'\x01\x02\x03\x04'` unreachability argument beyond reading that
  `dashboard.host` is JSON-sourced; the annotation does not depend on it either way.
- **One attack of mine failed and is recorded because it nearly became a false finding.** I measured
  the ratchet's row coverage against `origin/develop` on an assumption and got 6 of 87, which would
  have been a filed issue. `resolve_base_ref` falls back to `develop-v2` when the merge-base is not
  an ancestor of `develop`; measured, it returns `origin/develop-v2` here. Re-derived: 87 of 87.
  Nothing to file.

**#1556 stays OPEN** — five modules still hold a blind return. This slice closes no issue.

Issue #1621, opened 2026-08-30
redact()'s single vocabulary is a denylist: five credential-bearing .env keys are reached by neither rule, and two adjacent onion-client keys are classified differently


VijitSingh97:
Found while reviewing #1616 (merged as `bc2664d2`), which is a strict improvement and does not cause
this. Filing separately because it is the same failure *shape* one level out.

## What was measured

#1616 unified the `KEY=value` and JSON rules onto one suffix vocabulary, so there is now **one** list
rather than two that drift. That is the right fix for #1611. But it also means a name absent from that
single list is now unreached by **both** rules at once.

To size what is left, I derived the population mechanically rather than from the redactor's own list —
every `KEY=` name this repo actually renders into `.env`, harvested from `lib/pithead/*.sh`: **153
keys**. Both shipped `redact()` bodies were run over one line per key carrying an identical sentinel
value, asserting the **raw value is absent** rather than that a marker appeared.

| | result |
|---|---|
| sentinel survives before #1616 | 143 of 153 |
| sentinel survives after #1616 | 131 of 153 |
| redacted before but not after | none — no regression |

## The gap

Of the 131 still unreached, these are credential-bearing, and **neither** rule reaches them because
neither vocabulary contains their trailing word:

| key | what it holds | why the other rules do not catch it |
|---|---|---|
| `DASHBOARD_AUTH_HASH_B64` | a password hash, offline-crackable | ends in `B64`; a base64 hash is well under the 90-character shape bar |
| `DASHBOARD_AUTH_PW_FP` | a password fingerprint | ends in `FP` |
| `XMRIG_API_AUTH` | an API auth value | ends in `AUTH` |
| `DASHBOARD_ONION_CLIENT_AUTH` | onion client auth | ends in `AUTH` — note the sibling `..._CLIENT_PRIVKEY` **is** reached, because it happens to end in `KEY` |
| `NOTIFY_WEBHOOK_URLS` | a webhook URL, which commonly embeds its own token | ends in `URLS`; the vocabulary has `PING_URL`, not `URL` |

The last row of that table is the whole point: two adjacent keys describing the same onion client
credential are classified differently, and the only thing separating them is which noun the name
happens to end in.

## Why this is not "add five more spellings"

It is the third instance of one shape. #1596 was a redactor keyed to a *length*, failing on the input
that was short. #1609 was a redactor with no rule for an IP at all. #1611 was two vocabularies keyed to
[...]

Issue #103, opened 2026-06-04
Evaluate a remote Tari base-node option (tari.mode: remote), mirroring Monero remote


VijitSingh97:
## Goal
Add the option to point the stack at an **external/remote Tari base node** instead of running the bundled `tari` container — mirroring the existing Monero remote-node mode (`monero.mode: remote`). Lets operators offload Tari's RAM/disk to another machine or reuse an existing node.

## Why
- Tari is memory-hungry (capped in #55) and adds ~50 GB of disk; some operators would rather run it elsewhere.
- Monero already supports this (`monero.mode: remote` + `monero.remote.host/rpc_port/zmq_port`, with the `local_node` compose profile disabling the bundled monerod). A symmetric `tari.mode: remote` is the natural parallel.
- Complements "Tari is optional" (#35/#51) — another way to keep merge-mining without the local Tari container.

## How Tari is wired today (what remote mode must redirect)
Both consumers talk to the bundled node's gRPC at a fixed container address:
- **p2pool merge-mines** via `--merge-mine tari://172.28.0.27:18142` (docker-compose.yml).
- **The dashboard** reads sync via `TARI_GRPC_ADDRESS=172.28.0.27:18142`.
So remote mode = point both at `tari://<remote-host>:18142` / `TARI_GRPC_ADDRESS=<remote>:18142` and not start the bundled `tari` container (a compose profile, mirroring `local_node`).

## Proposed (mirror Monero remote)
- `tari.mode`: `local` (default) / `remote`.
- `tari.remote.host` (+ gRPC port, default 18142).
- `pithead` renders `TARI_GRPC_ADDRESS` and the p2pool `--merge-mine tari://…` target from config; a compose profile (e.g. `local_tari`) gates the bundled container off in remote mode.
- In remote mode, `tari.mem_limit` (#55) and the Tari Tor onion are N/A.

## Open questions / verify first
- **Does p2pool's `--merge-mine` work reliably against a *remote* Tari gRPC** (block templates + submission, latency, auth/TLS)? Tari merge-mining often assumes a trusted/local base node — confirm.
- **Trust/security:** merge-mining against a node you don't control — document the trust boundary; ideally the remote Tari is one you run.
- **gRPC exposure:** the remote node must expose gRPC reachably and securely (firewall/Tor considerations); Tari has no ZMQ equivalent — confirm gRPC suffices for both p2pool and the dashboard.

## Acceptance
[...]

PR #1623, opened 2026-08-30
fix(integration): reach the four .env keys no redaction rule touched, and guard the vocabulary (#1621)

VijitSingh97:
Closes nothing automatically — `Closes` cannot fire here, so #1621 gets closed by hand naming this PR.

## The change

`redact()`'s suffix vocabulary is a denylist over a key set that keeps growing, so it fails silently
and in the unsafe direction. Measured over all **129** keys `render_env` actually writes, with both
live rules: **19 redacted, 110 surviving**, and four survivors carry a credential nothing reaches.

| key | what it holds |
|---|---|
| `DASHBOARD_AUTH_HASH_B64` | base64 of a caddy bcrypt hash — offline-crackable |
| `DASHBOARD_AUTH_PW_FP` | an **unsalted** sha256 of the dashboard password |
| `NOTIFY_WEBHOOK_URLS` | webhook URLs, which commonly carry their own token in the path |
| `DASHBOARD_AUTH_USER` | a username — reached by nothing while its two siblings are, because the vocabulary carries `USERNAME` and this key ends in `USER` |

Suffixes added: `HASH_B64`, `PW_FP`, `USER`, `WEBHOOK_URLS`. Each was measured against all 129 keys
and **catches exactly its one intended key**, so over-redaction is zero. They append inside the
existing `-e` arguments, so `lib.sh` stays line-neutral at **692/692**.

`selftest-redact-vocab.sh` is the durable half: it derives the population from the single heredoc
`render_env` writes, screens it wider than redact()'s own list, and fails **by name** on a
credential-shaped key in neither the redacted nor the survivor list.

`selftest-redact.sh`'s vocabulary reader took a two-character widening: its alphabet was `[a-z_|]`
while the vocabulary's own is `[A-Za-z0-9_]`, so an entry carrying a digit made both lists
**unreadable** and the drift alarm could not fire at all.

## ⛔ The direction the issue called cheapest does not work

Measured, not argued — each suffix against all 129 keys:

| suffix | what it actually catches |
|---|---|
| `HASH` | **nothing.** The key ends `B64`. It would not fix row 1 of the issue's own table. |
| `URL` | four public endpoints — and **still misses** `NOTIFY_WEBHOOK_URLS`, which ends `URLS`. |
| `AUTH` | only `XMRIG_API_AUTH` (the enum `none\|name\|token`) and `DASHBOARD_ONION_CLIENT_AUTH` (a `normalize_bool` result). **Neither can hold a secret** — pure over-redaction of diagnostic state. |

Both are pinned as survivors, and the mutation adding a bare `AUTH` reds both rows. This also
corrects the issue's headline: `..._CLIENT_AUTH` is not "the same onion client credential" as
`..._CLIENT_PRIVKEY`. The privkey is redacted and the boolean is not, which is **correct**.

## What was RUN

- `make test-integration-selftest` — **all 18 self-tests green**, `selftest-redact-vocab: 37 passed,
  0 failed`. The count moving is the cheapest proof the glob discovered the new file.
- `shellcheck --severity=warning` (0.11.0) and `shfmt -i 4 -d` over the **full**
  `tests/integration/*.sh` + `mini-stack/*.sh` glob, not a partial one — both rc 0.
- `make lint-md` (0 errors), `make lint-docs-voice` (41 prose docs clean), `make lint-file-budget`
  **re-run with the new file `git add`ed**, because #1486 means an untracked file is invisible to it.
- **Mutation battery, per-arm, each leg `cmp`-verified to have actually mutated the file:** removing
  `HASH_B64`, `PW_FP`, `USER` or `WEBHOOK_URLS` reds **exactly its own row and no other**; adding a
  bare `AUTH` reds both over-redaction guards; a new credential-shaped key in the renderer reds the
  classification row; moving the heredoc bounds reds the population row rather than skipping.
- **Differential over the real captured corpus** — both archived bundles, all 14 artifact files, old
  `redact()` vs new. **16 changed lines, every one of them in `env.redacted.txt`, and zero across
  1,636 lines of `logs.txt` plus `config.json`, `doctor.txt`, `status.txt` and `compose-ps.txt`.**
  That is the answer to this change's real risk: the replacement takes the rest of the line, so a
  widened match could have stripped debugging value. It strips none.

## The `USER` arm needed that corpus run

`USER` was the one suffix I could not clear by reasoning — an uppercase `USER=` can plausibly appear
in an env dump or `docker inspect` output, where redacting it would cost triage information. The
corpus differential is what settles it: **zero changed lines outside `env.redacted.txt`.**

- **`TARI_GRPC_ADDRESS` survives, and in remote mode (#103) that discloses a third-party host.** Real,
  but a topology question rather than a credential one, and not what #1621 asked. Narrowing it would
  strip the local-mode value that makes a bundle readable. **Left surviving and named in the file** so
  the next reader meets a decision rather than an oversight.
- **`NTFY_URL` is a capability URL that ought to be redacted.** The bare word `URL` cannot be added
  without taking every public endpoint with it. Pinned as a **stated KNOWN GAP** that fails if someone
  closes it, mirroring `selftest-redact.sh`'s `notifications.ntfy.url` gap — the same value in the
  other syntax, so the two stay classified together.
- **An empty secret value still gains a `<redacted>` marker**, so a reader cannot tell "not
  configured" from "configured and hidden". **Pre-existing, not introduced here** — verified by
  running both the old and new `redact()` over `NTFY_TOKEN=`; both mark it. Not fixed in this PR.
- **Direction 2 of the issue, inverting to an allowlist,** is the strategically right answer and my
  measurement supports it. Not taken unilaterally: it changes redaction *policy* and getting it wrong
  strips what a bundle exists to carry. That wants a ruling, not a lane decision.

## Shared paths

`docs/dev/integration-testing.md` is under `docs/`, which is in `_shared.paths`. Disclosed here per
lane policy. No other file outside `tests/integration/` is touched.

## What I did NOT do

No live e2e run and no bench time — this is pure-logic harness work and the self-test tier proves it.
Every fixture is synthetic. I did not re-capture a bundle, so the corpus differential is over
**pre-#1616 captures**; that is the right input for a differential but means it cannot speak for
artifacts a current redactor would produce.

## Ponytail pass — over-engineering review, run off disk on this diff

Three candidates examined. **One is real duplication and I am declining to remove it; the reason is
the point.**

**O1 — my new file re-asserts four keys `selftest-redact-env.sh` already pins explicitly.**
Confirmed: `MONERO_VIEW_KEY`, `WALLET_RPC_USERNAME`, `XVB_DONOR_ID` and `HEALTHCHECKS_PING_URL` are
in that file's loop *and* in my `MUST_REDACT`. A mechanical invariant can remove tests, and I looked
for the case where **only one mechanism can act** before deleting anything. It exists: my population
is *derived from the renderer*, so a key dropped from `render_env` silently stops being tested here,
while the pinned row keeps asserting it. The two are not equivalent — one is a fixed pin, one is a
population sweep. **Declined: keeping both, named here rather than removed silently.**

**O2 — `in_list` is copied rather than shared.** Not a choice. `lib.sh` is at **692 lines against a
692 ceiling** (`docs/dev/file-budget.tsv:63`), so a shared helper cannot go there without an edit to
a budgeted shared file. The copy carries the same comment about why it is not a `case` glob, since
that is the trap it exists for. **Declined, with the constraint stated.**

**O3 — four classification lists where the precedent file has three.** `SHAPE_COVERED` is the extra
one and it is load-bearing, not symmetry: those four keys are covered by the `.onion` *shape* rule,
so a name-shaped sentinel would report them as leaking and be wrong. **That is not hypothetical — my
own first sweep made exactly that error**, which is why the list carries the reason. **Kept.**

Nothing else in the diff looked speculative: there is no abstraction with one caller, no
configuration knob nobody asked for, and no helper introduced for a single use.


VijitSingh97:
**For the reviewer: the 129-vs-153 key-count question is settled, and not by my own method's logic.**

It was flagged as something a reviewer should re-derive rather than pick between. That is the right
instinct, so here is a check that does not depend on how I derived the number.

A captured release-gate artifact contains the rendered `.env` itself. Comparing three sources:

| source | keys |
|---|---|
| `render_env`'s single heredoc — what this PR measures over | **129** |
| the captured `env.redacted.txt` in the `rigforge-control` bundle | **129** |
| every `^[A-Z_]*=` assignment anywhere in `lib/pithead/*.sh` | **213** |

**The first two sets are identical** — `comm` reports nothing in one and not the other, in either
direction. So the population this PR measures over is confirmed against an artifact produced by a
real run, which is independent of my derivation rather than a restatement of it.

The third number is the one to be careful with. A broad grep picks up every shell assignment in
those files, not the ones that reach `.env`: it adds 84 names including `C_RED`, `C_GREEN`,
`C_RESET` and `C_YELLOW` — terminal colour codes — plus `DASHBOARD_DIR`, `DASHBOARD_DIR_IS_DEFAULT`
and the `CONTROL_*` budgets. Those are locals and globals, and none of them is ever written to the
env file.

**Stated precisely, because it matters:** my broad grep returns **213**, not 153. I cannot reproduce
153 and I am not going to guess at the regex that produced it, so treat that figure as unexplained
rather than refuted. What is established is narrower and sufficient: **the rendered population is
129, confirmed against a real capture, and any count above it is measuring assignments rather than
env keys.** Nothing in this PR depends on the 153 being explained.

Verification note on the heredoc bound itself: I checked that nothing appends to the env file after
the heredoc closes, and `selftest-redact-vocab.sh` fails loudly rather than skipping if the bounds
ever move — the mutation that renames the heredoc delimiter reds the population row.

VijitSingh97:
## Non-author pass — **APPROVE and merge.** One finding filed, not blocking.

Everything below is **re-derived by me at source** unless a line says RELAYED. I ran both shipped
`redact()` bodies rather than reading them, and every clean result here has a control that fired.

### The population: 129, by a third independent route

I did not take 129 from this PR or from the predecessor's note. `render_env` writes through exactly
one `cat <<EOF >"$target"` (`lib/pithead/33-render-env.sh:359`, closing at 489, inside the `umask 077`
subshell, unconditional). Nothing else in `lib/pithead/*.sh` writes or appends to `$target` or
`$ENV_FILE`. Its body is **129 lines, every one of them a `KEY=` line** — no comments, no blanks, no
duplicates. That agrees with both of this PR's routes and it is arrived at differently from either.

**The predecessor's figure of 153 is withdrawn as this seat's, not merely unexplained.** It was mine
to defend and I cannot reproduce it; nothing here rests on it.

### Both `redact()` bodies, run over the real key set, asserting the RAW VALUE ABSENT

Synthetic `.env` carrying all 129 real keys, each with a unique sentinel chosen to trip none of the
shape rules. Both bodies extracted from git and executed. **I classified on the raw value being gone
from the output line — not on a `<redacted>` marker appearing**, since a marker test passes for a rule
that matched something else on the line.

| leg | rows out | redacted | survived |
|---|---|---|---|
| base `742ef689` | **129** | 19 | 110 |
| head `9bd8f26b` | **129** | 23 | 106 |

Both legs emitted 129 rows, so neither side was empty — the failure mode that made an earlier
differential here read "no regression" off two zero-line outputs. The outputs differ by exactly 4
lines, so the instrument is not null.

**Newly redacted (4):** `DASHBOARD_AUTH_HASH_B64`, `DASHBOARD_AUTH_PW_FP`, `DASHBOARD_AUTH_USER`,
`NOTIFY_WEBHOOK_URLS` — exactly the four claimed, and **nothing else**, so over-redaction inside the
rendered `.env` is zero as stated.

**REGRESSION SET: EMPTY.** No key redacted at base survives at head. That was my merge bar.

### The two claims your "declined `AUTH`" argument rests on — both true at source

- `XMRIG_API_AUTH=$worker_api_auth`, and `worker_api_auth=$(jq -r '.workers.api_auth // "none"')`
  (`33-render-env.sh:200`). It is a **mode enum**; the secret lives in the separate `XMRIG_API_TOKEN`,
  which the vocabulary already catches on `TOKEN`. Confirmed against the consumer
  (`dashboard/mining_dashboard/client/xmrig_client.py:279-324`).
- `DASHBOARD_ONION_CLIENT_AUTH` is `normalize_bool(config_bool ...)`, default `false`
  (`28-parse-and-validate-config.sh:428-430`). A boolean.

So a bare `AUTH` would be pure over-redaction of diagnostic state, as you argued. This also
independently confirms the retraction on the issue: `..._CLIENT_AUTH` is not the onion credential.

### Controls I ran, because a clean sweep with no fired control is not evidence

- **The new guard can fail.** Removing `HASH_B64` from the *shipped* `redact()` — `cmp`-verified to
  have actually changed the file, 0 occurrences left — takes `selftest-redact-vocab` to **36 passed,
  1 failed** and errors the make target. It reds *its own row only*. Restored after; worktree clean.
- **Shape coverage, which my sentinel test is blind to by construction.** With realistic values the
  four `*_ONION_ADDRESS` keys redact via the 56-char rule and `HOST_IP` via the IP rule. Your
  `SHAPE_COVERED` list is doing real work — a name-shaped sentinel does report those four as leaking,
  exactly as its comment says.
- **`USER` sitting after `USERNAME` in the alternation breaks neither.** `WALLET_RPC_USERNAME`,
  `MONERO_NODE_USERNAME` and `DASHBOARD_AUTH_USER` all redact at head.
- **The alphabet widening is load-bearing and I confirmed it with a control.** The old reader over the
  new `lib.sh` returns **None for both lists** — dead, not degraded — because of the digit in
  `HASH_B64`; the new reader returns 15 entries each side; and the control, **the old reader over the
  old `lib.sh`, still returns 11 each side**. Without this two-character change the drift alarm would
  have gone silent at exactly the moment this PR gave it something to catch. Good catch.

### Gates

- `make lint-file-budget` **rc 0 at the branch head** *and* **rc 0 on the `merge-tree` result**. Both
  legs run because the branch is not on the tip (merge-base `742ef689`, tip `0a97f241`); the tip move
  touched no budget row, so no stale-high value is in play.
- `resolve_base_ref` **called**, not read: returns `origin/develop-v2`, so the local ratchet is real.
- `make test-integration-selftest` at head: **214 passed, 0 failed** — my run, not yours.
- `merge-tree --write-tree` rc 0. 18 PR checks terminal: 16 SUCCESS, 2 SKIPPED, 0 failed.

### Finding — filed, not blocking: the `NTFY_URL` decline argues against a fix nobody proposed

The known-gap message says the bare word `URL` cannot reach it without taking every public endpoint
along. **That is true and it is not the question**, because it is not the technique this PR uses. For
the other four keys you added a *specific suffix* and measured its blast radius. Applying your own
method: adding `NTFY_URL` to the alternation catches **exactly one of the 129 keys — itself — and
strips the real capability token**, with zero over-redaction. I measured it the same way you measured
the other four.

The gap is **pre-existing and this PR does not widen it**, so it is not a merge blocker. What earns a
follow-up is that the *stated reason* is now durable in-repo text a future contributor will read as
"this one cannot be fixed narrowly", which is the overclaim that stops the coverage being added. Both
syntaxes should move together, so the JSON-side `notifications.ntfy.url` gap belongs in the same
change. Filing it with these measurements attached.

`TELEGRAM_CHAT_ID` and `TELEGRAM_CONTROL_ALLOWED_IDS` also survive as raw identifiers while
`XVB_DONOR_ID` is redacted — but your suite **pins the chat ID as a deliberate survivor**
("survives, as a bundle needs it to"), so that is a recorded decision and not an oversight. Noted, not
filed.

### What I did NOT do

No live e2e, no bench, no bundle capture. **I did not re-run your corpus differential** — those
archives are outside the repo, so the `USER` over-redaction clearance on log-shaped input is
**RELAYED from you, not re-derived by me**; my measurement covers the rendered `.env` only, where it
is exactly one key. I did not run `lint-sh` (the OOM path on this box); CI's Shell tests cover it and
are green.

Merging.

Issue #1014, opened 2026-08-15
A one-click rig upgrade leaves no per-rig record: the change history never shows it, and hashrate-by-config misattributes it


VijitSingh97:
## What happens

Press "Upgrade rig…" in Worker Inspect, the rig upgrades, and nothing anywhere records that
*this rig* was upgraded *then*.

- `handle_worker_upgrade` (`web/server.py:287`) returns `202 {id, status:"pending"}` and never
  calls `_record_worker_result`. The only call site is `handle_worker_apply`
  (`web/server.py:283`), so `worker_config` — the table behind the panel's change-history — gets
  a row for config edits and never for upgrades.
- The client's `upgResult` is component state in `workerview.mjs`; it dies when the dialog closes.
- `control.log` does carry `"action":"worker-upgrade"` rows (`pithead:6308`, `:6467`, `:6479`),
  mirrored into `audit_events` — but those call sites pass no `keys` argument, and `control_audit`
  sanitizes `keys` with `tr -cd 'A-Z0-9_ '` (`pithead:5646`) anyway, so a lowercase rig name could
  not survive it. **The audit row does not say which rig.**

So the per-rig upgrade timestamp exists nowhere queryable.

## Why it matters beyond the missing row

`hashrate_by_config` (#492) exists to let an operator compare config versions empirically: it
buckets `worker_history.h15` into the windows between config-change rows
(`storage_service.py:1295`). It is blind to RigForge version changes. Upgrade a rig mid-window and
whatever the new build does to hashrate is silently attributed to the config version that happened
to be active. The feature reports a config verdict on a firmware change.

It also blocks marking upgrades on the per-worker hashrate chart (sibling issue), and it means the
one dashboard action that reaches out and rebuilds a remote machine is the *least* traceable thing
the panel does.

## Proposed fix

Record the upgrade the same way an apply is recorded. The upgrade is 202-then-poll, so the write
belongs where the terminal result lands — the client already polls `/api/control/result`; either
persist on the server when the runner writes its terminal result, or have the poll's terminal
response record it. Either way one row per upgrade attempt: worker, target version, terminal
status, reason.

Whether that row goes in `worker_config` (with a type column, so `hashrate_by_config` can split
its buckets on upgrades too) or a sibling table is the design call this issue needs to make. The
`hashrate_by_config` bucketing should split on both kinds of change once the data exists —
otherwise it keeps attributing build changes to config.

[...]

PR #1624, opened 2026-08-30
fix(dashboard): #1556 slice 10 — web/server.py's procedure/signed pair, and the first nested annotation

VijitSingh97:
#1556 slice 10 — `web/server.py`. Two failure returns, both through the `except` door, and the
production diff is **two signature lines**: `_finalize_worker_upgrade(...) -> None` and
`_num(name) -> float | None`.

## The two readings

They land in **different verdicts**, as slice 9's pair did: one `procedure`, one `signed`, **zero
`unjudged`**, so no `_UNJUDGED_AND_READ` entry is owed.

- **`_finalize_worker_upgrade` → `-> None`** is a genuine procedure: ONE return, ZERO valued, ZERO
  yields, measured — with `handle_control_preview`'s three valued returns as the positive control
  that the walk can see a valued return. It is #1014's background half; it records the terminal
  outcome of a worker upgrade and answers nobody, so its `except` return is a bare `return` after
  `logger.exception` and there is no success value for a failure to hide inside.
- **`_num` → `-> float | None`** is a true `signed`. All three returns are valued and every one is a
  float or `None`. Its docstring already said so in prose ("anything non-numeric reads as absent"),
  and the consumer was **read rather than assumed**: `audit_service.filter_log_entries` documents
  `None` as "don't filter on this axis", so `None` is out-of-band there, distinguishable from every
  float bound. The annotation moves a promise the code already kept into the signature.

## The finding: this slice's proof claim is NARROWER than slices 1-9's

`_num` is a **nested closure**, the first #1556 has annotated, and annotating a nested def is not
free at the bytecode level. Both annotated functions **are** bytecode-identical base-vs-head, but
the enclosing `_log_filters` is **not**, and that is correct rather than a defect. Its delta was
read instruction by instruction and is exactly the annotation being built at def time — a prologue
building `float | None`, `MAKE_FUNCTION closure` becoming `MAKE_FUNCTION annotations, closure`, and
the two jump offsets that prologue shifts. Nothing else moved. The package already carries the
shape (`wizard.py:build_config` annotates two nested defs). Of the **eight blind functions left,
six are class methods and two are module-level; none is nested inside a FUNCTION**, which is the
only scope that rebuilds the annotation per call — a method's is built once in the class body at
import. So no remaining slice of #1556 inherits it, and nothing is claimed about a nested def some
future slice might introduce.

That enumeration is keyed by NODE rather than by name and carries a control that the walk can report
a nested def at all (`_num`, `wizard.py:s_`, `_odds_day` all fire), plus a negative-direction control
that a known method reports `class`. **The first pass had neither**, and reported all eight as
"top-level" by collapsing the six methods into the same bucket. The verdict did not move; the
characterisation did. Re-derived after `lane-controller-e2` flagged it as a disjointness claim worth
re-enumerating — a right conclusion under a wrong characterisation has no tell of its own.

## Counts, measured at this head rather than carried

29 of 33 modules pinned; **8 failure returns still blind in 4 modules**, every one through the
`except` door — and that last clause is a measurement, not a reading the walk could not contradict:
`guard` is a door the same walk still reports 31 times package-wide. 14 pinned modules hold residue,
15 score zero. This module's residue went **3 sites in 2 functions to zero**, which discharges the
prediction slice 9's retraction left standing: it named `web/server.py` as the module where `blind`
and the residue are the same set.

Four counting sentences were falsified and are corrected here: the pins header's "28 of 33" and "10
failure returns … in five modules", and the sibling test file's "fourteen of the twenty-eight" and
"the fourteen that score zero". The last had drifted for the same reason "the twelve that score
zero" did at slice 9.

## The anchor, and a reason that turned out not to be one

The anchor is `_finalize_worker_upgrade`: declared first, and top-level where `_num` is a closure a
refactor could drop for a reason that is not a walk failure. What is **not** a reason — measured
before choosing rather than argued after — is the nested-def regression: degrading `_own_nodes` to
`ast.walk` **adds** a `_log_filters` row rather than removing one, so neither candidate fires on it.

## What was RUN

- **Verdicts measured** from `classify` at this head, never predicted from the decision chain.
- **Bytecode identity** base-vs-head for both annotated functions plus `handle_audit_log` as a
  bystander, comparing `co_code`/`co_names`/`co_varnames`/`co_consts` recursively (never a `dis`
  repr). Positive control: a seeded `else None` → `else 0.0` inside `_num` flipped that function and
  only that function.
- **Three controls**, each armed by an exact-match applier that refuses a site not matching exactly
  once, each restored by `cmp` against a hashed copy: (A) unannotate the anchor → law 1 only;
  (B) re-sign it `-> bool` → law 2 only; (C) corrupt its `_ANCHORS` value → the vacuity guard only.
  Each reddened its **own** law and only the `[web/server.py]` parameter, against a control-0 run
  with zero failures.
- **Full suite 2364 passed** — the +3 over slice 9 is one module times three parametrized laws.
- **Patch coverage read from `coverage.xml`, not from a gate's rc** (#1557): both changed production
  lines covered, hits=1 each. Module line-rate 0.981; overall 97%.
- **Twelve lint targets run individually**, all rc 0 — run one at a time because `make` stops at the
  first failure and would hide the rest.
- **Base moved during this slice** and the proof was taken accordingly: `742ef689` landed between my
  slice-9 copy and its squash, so the parents differ and a tree check is the wrong instrument —
  the **patches** were compared and are identical, with slice 8's patch as a control that two
  genuinely different patches do report as differing. `git rev-parse <base>:dashboard` is identical
  at both bases, so the program the suite ran did not change and the counts above stand.

## What was NOT done

- **`make lint-sh` was not run.** It is the serialized OOM path and this diff touches zero shell
  files; CI runs it here.
- **No `security-reviewer` subagent was spawned**, which this lane's standing rule asks for on
  anything touching the control channel — `_finalize_worker_upgrade` sits on `control_service`'s
  wait path. The session's harness bars unrequested subagents, so the pass was done in-line instead.
  The basis for calling it safe is narrow and stated so it can be checked: the diff is two type
  annotations, both annotated functions are proven bytecode-identical, and the only bytecode delta
  anywhere in the diff is `_log_filters` constructing `_num`'s annotation. No control-channel
  behaviour, payload, or trust boundary is touched.
- **Nobody classified the residue.** A pin is coverage of the mechanism; it clears nothing.

### DIFF
--- a/dashboard/mining_dashboard/web/server.py
+++ b/dashboard/mining_dashboard/web/server.py
@@ -385,7 +385,7 @@ async def handle_worker_apply(request):
     return web.json_response({"id": rid, **res})
 
 
-async def _finalize_worker_upgrade(state_mgr, worker, version, rid, latest_data=None):
+async def _finalize_worker_upgrade(state_mgr, worker, version, rid, latest_data=None) -> None:
     """The server-side half of recording a worker-upgrade attempt (#1014): ``handle_worker_upgrade``
     returns 202 before any result exists (a rig rebuild can take minutes, so it never waits inline
     like ``handle_worker_apply`` does), so this runs as its own background task and records the
@@ -493,7 +493,7 @@ def _log_filters(request):
     non-numeric reads as absent — a malformed bound must never 500 a log view) and ``q`` trimmed
     and length-capped (it's compared, never stored or echoed unsanitized)."""
 
-    def _num(name):
+    def _num(name) -> float | None:
         v = request.query.get(name)
         if v in (None, ""):
             return None
--- a/dashboard/tests/service/annotation_pins.py
+++ b/dashboard/tests/service/annotation_pins.py
@@ -41,7 +41,7 @@ moved the data, which is the failure the pin comment below records happening twi
 """
 
 # The modules with ZERO unannotated failure returns, measured over live source. Each slice of #1556
-# adds the module it finished. Measured at this tip: 28 of 33 modules that hold a failure return at
+# adds the module it finished. Measured at this tip: 29 of 33 modules that hold a failure return at
 # all — `client/xvb_client.py` by slice 1, `service/worker_config_store.py` by pre-existing work,
 # the six single-function `client/` modules by slice 2, the four single-function `web/` modules by
 # slice 3, the two single-function `config/` modules by slice 4, the four outbound senders under
@@ -49,9 +49,11 @@ moved the data, which is the failure the pin comment below records happening twi
 # MULTI-function slice — the four parse-or-read pairs under `service/`, each pinned only once BOTH of
 # its failure returns were annotated and read, and the ten handle-guard procedures in
 # `service/storage_service.py` by slice 7, and its three siblings in `service/telemetry_store.py` by
-# slice 8, and the two `except`-door functions in `helper/utils.py` by slice 9. **10 failure
-# returns are still blind**, in five modules — every one through the `except` door, because slice 8
-# closed the handle-guard population entirely.
+# slice 8, and the two `except`-door functions in `helper/utils.py` by slice 9, and the two in
+# `web/server.py` by slice 10. **8 failure returns are still blind**, in four modules — every one
+# through the `except` door, because slice 8 closed the handle-guard population entirely. That
+# last clause is a measurement and not a reading the walk could not contradict: `guard` is a door
+# the same walk still reports 31 times package-wide.
 #
 # Every figure above is re-derived at the head being shipped, never carried across a slice: 5b moved
 # this tuple from 18 to 21 and left the sentence beside it saying 18. A slice adds rows to a tuple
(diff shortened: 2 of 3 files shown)


VijitSingh97:
**RETRACTION — a claim in this PR's original body was wrong in its characterisation, right in its
verdict. Corrected in the body, the commit message and the shipped comment; recorded here so the
swap is not silent.**

The original said **"all eight remaining blind functions are top-level"**. Re-derived: **six are
class methods and two are module-level.** None is nested inside a *function*, which is the only
scope that rebuilds the annotation per call — a method's annotation is built once in the class body
at import — so the conclusion is unchanged: no remaining #1556 slice inherits the `_log_filters`
case.

**How it was wrong.** The first enumeration keyed its scope map by function NAME rather than by
node, and folded class methods into the same bucket as module-level defs, so every one of the eight
came back reading "top-level". **It had no control** — it was never shown able to output `NESTED`
at all, and an enumeration where every row comes back the same way has not been shown able to say
anything else. The re-derivation keys by node, and carries a positive control that fires three ways
(`web/server.py:_num`, `wizard.py:s_`, `web/xvb_views.py:_odds_day` all report `function` scope)
plus a negative-direction control that a known method reports `class`.

Credit where it is due: `lane-controller-e2` flagged the sentence as a disjointness claim worth
re-enumerating rather than accepting. It was right to, and the flag cost one measurement.

Nothing else in the PR moves. The verdicts, the bytecode identity on both annotated functions, the
three law controls, the 2364-pass suite and the coverage reading are all unchanged and were measured
before this correction.

VijitSingh97:
## Non-author pass — **APPROVE and merge.** One correction to the merge record; the verdict does not move.

Re-derived by me at source unless a line says RELAYED. I re-measured at **`e72e1edd`**, the current
head, because the earlier ruling on this PR was made at a sha the force-push replaced.

### The precondition the no-subagent ruling was conditioned on still holds

The production diff against `git merge-base` (`0a97f241`) is **two signature lines and nothing else**;
under `mining_dashboard/` exactly one file differs. So the standing ruling — that no `security-reviewer`
pass is owed here — carries, on the grounds it was actually given: no new surface, and asking a peer to
run a pass this harness bars would launder a limit I share. **If a later slice changes behaviour rather
than annotations, that ruling does not carry to it.**

### Bytecode — and the one correction

Recursive structural comparison of every code object base-vs-head (`co_code`, `co_names`,
`co_varnames`, `co_argcount`, `co_consts` recursively; never a `dis` repr, which carries addresses that
differ run to run). 33 code objects each side, identical name sets.

| code object | identical? |
|---|---|
| `_finalize_worker_upgrade` | **yes** |
| `_log_filters._num` | **yes** |
| `_log_filters` | no — disclosed in the body |
| `<module>` | **no — not disclosed** |
| `handle_audit_log`, `handle_worker_apply` (bystanders) | yes |

**The correction: the body says "the only bytecode delta anywhere in the diff is `_log_filters`
constructing `_num`'s annotation." `<module>` has a delta too**, from the *other* annotation. Read
instruction by instruction it is 214 → 215 instructions, and the entire difference is:

```
-  LOAD_CONST <code _finalize_worker_upgrade>;  MAKE_FUNCTION defaults
+  LOAD_CONST ('return', None);  LOAD_CONST <code _finalize_worker_upgrade>;  MAKE_FUNCTION defaults, annotations
```

Everything else the naive diff shows at module level is `LOAD_CONST` **argrepr addresses**, not code.

**This strengthens your argument rather than denting it, which is why it is worth stating.** A delta in
the enclosing scope is not what makes this slice special — *every* slice 1-9 had one, in `<module>`,
for exactly this reason. What is special is that `_num`'s enclosing scope is a **function**, so the
annotation rebuilds per call, where a module body and a class body run once at import. Your disjointness
argument already turns on precisely that distinction and is unaffected.

**Disposition: merge-record only.** The unqualified sentence is in the PR body. The commit message's
"Nothing else moved" sits immediately after the description of the `_log_filters` delta and scopes to
it, so under its own convention it is defensible and I am not calling it wrong. No force-push, no
discharge commit, nothing shipped to correct.

### The annotations were RUN, not walked — an AST cannot prove a return's type

- **`_num -> float | None`.** Executed the real function over 10 probes: absent, empty, `1.5`, `-3`,
  `abc`, `1e9`, `nan`, `inf`, `-inf`, `0`. Every result is a `float` or `None`, **including the three
  non-finite cases** the docstring promises read as absent. `isinstance(True, float)` is `False` here,
  so the bool-subtype trap that would make this annotation a half-truth does not apply.
- **`_finalize_worker_upgrade -> None`.** `async`, **0 yields, 0 nested defs, exactly one bare
  `return`**. Nothing can come back but `None`.

### The prediction slice 9 left standing is discharged — measured, not accepted

Via the gate's own `classify_package` and `unannotated_falsy_by_module`, run at both shas:

| | base `0a97f241` | head `e72e1edd` |
|---|---|---|
| `web/server.py` blind | 2 (`_finalize_worker_upgrade`, `_num`) | **0** |
| `web/server.py` residue | 3 sites | **0** |
| package blind | 10 in 5 modules | **8 in 4 modules** |

The blind set and the residue set for this module were the **same two functions**, so the residue had to
go to zero and it did. That is the bar this slice had to clear.

### The disjointness claim, re-enumerated rather than accepted

Keyed by **node**, walking children so a def inside a `ClassDef` scores `class` and a def inside a
function scores `function`:

**6 class methods, 2 module-level, 0 nested inside a function** — reproducing your retraction exactly,
by a walk I wrote independently.

**The positive control fires:** the same walk reports `function` scope for `_num`, and for `_odds_day`
and `_face_value` in `web/xvb_views.py`. It has been shown able to emit the answer that would have
falsified the claim. The six `class` rows are themselves the demonstration it can emit `class`.

### A figure I nearly corrected and should not have

Your "14 pinned modules hold residue, 15 score zero" read wrong against my first measurement of 29 and
58 — because I had counted **all** modules. Restricted to `PINNED` (29 modules) it reproduces exactly:
**14 hold, 15 zero, 14+15 = 29.** Your convention, your numbers, no correction owed.

### Controls and gates

- **Negative control, mine:** unannotating the anchor — `cmp`-verified applied, 0 annotated occurrences
  left — reds `test_no_failure_return_in_a_pinned_module_is_unannotated[web/server.py]` and **only**
  that, on that parameter only. Restored; worktree clean. The gate is armed, not green by construction.
- `make lint-file-budget` **rc 0 at the branch head** *and* **rc 0 on the `merge-tree` result**. Both
  legs, because the branch is not on the tip — I merged #1623 under it during this pass.
- `merge-tree --write-tree` rc 0.
- **Full dashboard suite at head: rc 0, 2364 collected; base collects 2361.** The +3 is one module times
  three parametrized laws, as you said. `test_annotation_coverage.py` alone passes and prints
  `0 sites 0 fns web/server.py`.

### What I did NOT do

I ran **one** of your three law controls, not all three; A and C are **RELAYED**. The `coverage.xml`
reading and the seeded `else None → else 0.0` control are **RELAYED, not re-derived by me**. I did not
run `lint-sh` — the OOM path on this box, zero shell files in this diff, CI covers it and is green.

Merging.

Issue #1506, opened 2026-08-29
integration harness: the four deployed twins have drifted from tests/integration/ in four different ways, and only one of them must stay divergent


VijitSingh97:
Four artefacts in `tests/integration/` have deployed working copies on the build server, and all four have drifted from the tracked ones. This was carried as an owed follow-up after #1489/#1501, where the same claim had to be corrected in both places and the repo-scoped sweep could not see the deployed half — twice, in two different cycles.

The standing assumption inside the harness lane has been "they have structurally diverged, they carry operational detail the repo lacks, so never sync over them." **Measured, that is right for one of the four and wrong for the other three, in three different ways.** There is no single reconciliation plan; there are four, and the classification is the deliverable.

## What was measured

For each pair: line counts, `diff` hunk counts, and — the step that changes the answer — a re-diff after normalising both sides through `shfmt -i 4`, which separates formatting drift from content drift. Plus a claim-keyed grep over the whole deployed file with the repo copy as the positive control, and the topology detector on both halves of the README pair.

## Class A — formatting only. `system-info.sh`

70 deployed lines vs 77 tracked, 2 diff hunks. **After `shfmt -i 4` the two files are byte-identical.** The entire divergence is that the tracked copy has been through the repo's formatter and the deployed one has not (one-line function bodies, compact `{ ...; }` blocks). Zero content at risk in either direction.

*Reconciliation:* replace the deployed copy from the tracked one. Nothing is lost.

## Class B — the tracked copy is the stricter one. `build-pruned-chain.sh`

70 deployed lines vs 86 tracked; **8 real content lines after format normalisation.** The direction is the opposite of the standing assumption:

- Tracked: `SRC_DIR="${SRC_DIR:?set SRC_DIR to the monero data dir to copy from …}"` — a required variable. Deployed: `SRC_DIR="${SRC_DIR:-<default>}"` — a default, pointing at a path that **does not exist on the box** (checked). It fails closed with `FAIL_NO_SRC` rather than doing something wrong, which is why nobody has noticed, but a default that silently names a stale project path is a worse contract than a required one.
- The tracked copy carries the #1489 copy-then-swap warning on step 4 ("it renames its result over `$DST_DIR/lmdb` … never point it at a live data dir"). Grepping the **whole** deployed file for `copy-then-swap|renames|1489|live data dir|never point` returns **nothing**; the same pattern against the tracked copy returns two lines, which is the positive control proving the pattern can match.

[...]

Issue #1330, opened 2026-08-23
Harden the dashboard-onion test file's cross-file source-order dependency


VijitSingh97:
# Harden the dashboard-onion test file's cross-file source-order dependency

## What

`tests/stack/test-dashboard-onion.sh` reads `$caddy_https` (and, via a suppressed
SC2154, `$auth_hb64`) that are set in `tests/stack/test-dashboard.sh`. This works only
because `run.sh` sources `test-dashboard.sh` immediately before `test-dashboard-onion.sh`
— the access-log unit legitimately needs both a dashboard-scheme and an onion-scheme
value to assert the combined access-log block.

Both files' headers document "do not reorder the two source lines / do not source this
file first," and the source lines in `run.sh` are adjacent and in the documented order.
The full suite (2903/0) would catch a reorder the next time it runs.

## Why it's worth tightening

The guard lives only in file-header prose, not at the point of risk. A future edit to
`run.sh` that inserts a source line between the two, or reorders them while skimming the
diff rather than the file headers, produces a *silent* failure mode: `$caddy_https`
becomes an empty string and the access-log assertion fails with an unhelpful empty-string
mismatch rather than an obvious wrong-order error. The `#1252` split map's own stated
preference is to re-derive cross-file state locally rather than rely on source order.

## Two options (either closes it)

1. **Re-derive `$caddy_https` locally** in `test-dashboard-onion.sh`. The render is ~6
   lines already spelled out in `test-dashboard.sh`; the access-log unit reads only
   `$caddy_https` (not `$auth_hb64`), so re-deriving that one value removes both the
   cross-file dependency and the SC2154 disables entirely. Trade: ~6 lines duplicated.
2. **Add an inline warning at the two `run.sh` source lines** ("these two must stay
   adjacent and in this order — test-dashboard-onion.sh reads $caddy_https from
   test-dashboard.sh"), moving the guard to the point of risk. Trade: keeps the coupling,
   just signposts it where an editor will see it.

Also check whether `$auth_hb64`'s SC2154 disable in the onion file is even needed — the
reviewer noted the access-log unit reads only `$caddy_https`; if nothing reads
`$auth_hb64` there, drop its disable.

Not blocking the split PRs (each move was independently confirmed behavior-preserving);
worth closing before the same source-order pattern gets copied into the harder
sandbox-coupled cluster cuts still to come.

VijitSingh97:
[...]

PR #1625, opened 2026-08-30
fix(integration): lift the two facts the deployed twin still held, and guard the copy that would delete them (#1506)

VijitSingh97:
`Closes` cannot fire here — PRs merge to `develop-v2` while the default branch is `develop` — and
this closes only part of #1506 anyway. What remains is named at the bottom.

## The change

#1506 classified four deployed twins and asked for two things in tracked files. Both are here.

**Class C — lift what only the box's copy held.** The issue named two facts missing from tracked
`compact-chain.sh`. **Re-measured before building: one of them had since landed**, so only these two
are lifted, and I am saying which rather than restating the issue's pair:

| fact | state when re-measured |
|---|---|
| the cost warning | already in the tracked copy — **not** lifted |
| `"Error renaming"`, the fourth copy-then-swap string | still missing — lifted |
| the cost *mechanism* (why the run is expensive at all) | still missing — lifted, with its provenance |

The tracked header listed **three** binary strings; the deployed one listed four.

**Class D — say that the README pair is meant to differ.** The issue's words: *"Say so in the tracked
copy, so the next reader does not try."* Neither direction of a sync is available — one way publishes
a topology, the other deletes what the box runs on.

## The strings were read out of the binary, not out of either header

My own lane's rule is that when two files in the lane disagree about a tool, the finding is at the
**tool**. So all four were checked against `monero-blockchain-prune` directly, each with a negative
control proving the check can return zero:

| string | in binary |
|---|---|
| `"Swapping databases, pre-pruning blockchain will be left in"` | 1 |
| `"-old and can be removed if desired"` | 1 |
| `"Blockchain pruned OK, but renaming failed"` | 1 |
| `"Error renaming"` | **1 — the one the tracked header lacked** |
| `"Copy database anyway if already pruned"` | 1 |
| *negative control* `"Error renumbering"` | **0** |

The `-old...` row needed `grep -F -e`: this box's `grep` is a ugrep shim and ate the leading dash as
a flag, scoring it absent on the first pass. Recorded because a silent zero there would have read as
a missing string and put a wrong claim in a safety header.

## ⛔ The finding this PR did not set out to make

Grepping `docs/` for the defect's own words — the step my lane keeps paying for skipping — turned up
`docs/dev/test-server-architecture.md` step 5:

    cp ~/pithead/tests/integration/testbench-README.md ~/pithead-testbench/README.md

**On a fresh box that is correct**: it seeds the README, which is what step 5 is for. The hazard is
**re-running** the block, which is the natural way to pick up a script fix later — it silently
overwrites whatever the operator annotated. Both halves have held such detail: the README by design,
and a deployed script outlived the tracked one on a safety warning, which is #1506's own Class B.
**Guarded at the point of risk rather than removed** — the guard living only in file-header prose is
what #1330 filed.

I first read this as flat contradiction of the note I had just written. It is not; it is a re-run
hazard on one block, and the PR says the narrower thing.

## What was RUN

- `shellcheck --severity=warning` (0.11.0) over the **full** `tests/integration/*.sh` +
  `mini-stack/*.sh` glob, not a partial one — rc 0, and the measured file list confirmed to contain
  `compact-chain.sh`.
- `shfmt -i 4 -d` over the Makefile's exact glob — rc 0. Not skipped; skipping it has reddened a PR
  here before.
- `make test-integration-selftest` — **214 passed, 0 failed.**
- `make lint-md` (0 errors), `make lint-docs-voice` (41 prose docs clean), `make lint-file-budget`
  with the files `git add`ed. That target is also the only place the **monotonic ratchet** actually
  runs (#1608 — CI skips it), and it passed.
- `detect-topology.py` on both changed docs — clean, each with a positive control that fired.
- `git merge-tree --write-tree` against `develop-v2` — rc 0, a real merge rather than the
  `MERGEABLE` field, with the control returning rc 1 against frozen `develop`.

- **The block-by-block mechanism is carried as recorded, not measured, with a `[TODO: verify
  upstream]`.** I could not verify the copy loop from here, and the house rule is to mark rather
  than invent. I weighed dropping it: the *consequence* is measured and is the actionable half, so
  the mechanism earns little on its own. **Kept anyway**, because leaving it only on the box is the
  exact condition #1506 exists to end. Named here so a reviewer can overrule that call cheaply.
- **The two notes overlap** — one in the README, one at the `cp` in the architecture doc. Deliberate,
  and not the recipe duplication my lane warns about: they address different readers at different
  moments, and the point-of-risk one is the one #1330 says is missing when a guard lives only in a
  header.
- **`system-info.sh` (Class A) and the deployed-side patches (Class B, Class C's other direction) are
  not here.** They are box operations, not tracked-file changes, and Class B's safety half is already
  done on the box.

## Shared paths

`docs/dev/test-server-architecture.md` is under `docs/`, which is in `_shared.paths` — disclosed
rather than taken with `.lane-override`. Everything else is `tests/integration/`, this lane's own.

## What I did NOT do

No bench time and no live run: this is header and prose work, and the binary strings are readable
without executing the tool. **I did not run `monero-blockchain-prune`** — the lane's standing rule is
not to re-derive this bench's chain or re-run the ~2.5h build. I did not reconcile the remaining 73
differing lines between the two `compact-chain.sh` copies; only the two facts #1506 named were in
scope, and the rest has not been classified.

### DIFF
--- a/docs/dev/test-server-architecture.md
+++ b/docs/dev/test-server-architecture.md
@@ -152,6 +152,11 @@ cp ~/pithead/tests/integration/testbench-README.md ~/pithead-testbench/README.md
 ~/pithead-testbench/system-info.sh > ~/pithead-testbench/system-info.md
 ```
 
+These copies seed a **fresh** box. On a box that already has them they overwrite whatever the
+operator annotated locally, and both halves have held such detail (#1506): the README names hosts,
+users and paths the repo twin deliberately does not carry, and a deployed script has outlived the
+tracked one on a safety warning. Diff before you copy, and reconcile by hand rather than overwrite.
+
 **6. Validate it's release-fit:**
 
 ```bash
--- a/tests/integration/compact-chain.sh
+++ b/tests/integration/compact-chain.sh
@@ -13,11 +13,19 @@
 #   ~2.5 hours reclaiming nothing. Read `mdb_stat -ef` on an IDLE copy first. (An earlier header
 #   here promised "~95 GiB of live data" — a retired expectation, never a measurement.)
 #
+# THE COST IS SET BY THE SOURCE CHAIN, NOT BY WHAT THE RUN RECLAIMS.
+#   The figure above is this bench's, and it is the case that reclaims nothing — so budget hours
+#   for any mainnet source regardless of how much you expect back. The box's own operational copy
+#   of this script records the reason as a block-by-block rather than page-level copy; that
+#   mechanism is carried here as recorded, not measured. [TODO: verify upstream — read the copy
+#   loop in monero's blockchain_prune before restating it as fact.]
+#
 # THE TOOL IS COPY-THEN-SWAP — IT DOES NOT ONLY READ THE SOURCE (#1489).
 #   After building <data-dir>/lmdb-pruned it renames <data-dir>/lmdb to <data-dir>/lmdb-old and
 #   moves the pruned DB into its place. The binary carries the strings "Swapping databases,
-#   pre-pruning blockchain will be left in", "-old and can be removed if desired" and
-#   "Blockchain pruned OK, but renaming failed". `--copy-pruned-database` does not change that:
+#   pre-pruning blockchain will be left in", "-old and can be removed if desired",
+#   "Blockchain pruned OK, but renaming failed" and "Error renaming" — all four read out of the
+#   binary itself, not out of either header. `--copy-pruned-database` does not change that:
 #   its help text is "Copy database anyway if already pruned", so it lifts an early-exit guard on
 #   an already-pruned source rather than selecting a copy-only mode.
 #
--- a/tests/integration/testbench-README.md
+++ b/tests/integration/testbench-README.md
@@ -9,6 +9,12 @@ over SSH as `$BENCH_HOST`.
 See `docs/dev/test-server-architecture.md` for the full architecture and how to stand a box up from
 scratch.
 
+**This file is the generic twin, and it is meant to differ from the copy on a running box.** A real
+bench keeps its own README naming that box's hosts, users and paths — the detail an operator needs
+and the detail this repo must not carry. Neither copy can be synced onto the other: syncing this way
+would publish a topology, and syncing the other way would delete what the box runs on. Reconcile the
+two by hand, fact by fact, or leave them alone (#1506).
+
 ## ⚠️ Golden rules
 
 This is a test bench, not a production miner — downtime and teardown/redeploy are fine. The


VijitSingh97:
**Ponytail pass — run for real off disk on this diff, not cleared by the gate's retry.**

The gate clears unconditionally on a re-run and verifies nothing, so recording what the pass actually
did. It changed the diff:

**Found and fixed.** My first version of the architecture-doc note ended *"To pick up a fix to the
scripts later, copy that file alone."* That is wrong in the direction this PR exists to correct: it
implies a single-file copy is safe, when the deployed **scripts** have also carried detail the
tracked ones lacked — which is #1506's own Class B, one file over from the case I was writing about.
Replaced with *"Diff before you copy, and reconcile by hand rather than overwrite."*

**Three candidates examined, two declined with reasons in the PR body** (the `[TODO: verify
upstream]` on the block-by-block mechanism, and the deliberate overlap between the README note and
the point-of-risk note). Nothing else in the diff looked speculative — there is no new abstraction,
no helper with one caller, and no knob nobody asked for. The diff is three prose blocks and one
string added to a list.

VijitSingh97:
## Non-author pass — **APPROVE and merge.** One suggestion, not a blocker.

Re-derived by me at source unless marked RELAYED.

### The binary strings are NOT relayed — I read them out of the binary myself

You expected this to be un-recheckable from a reviewer's seat. It is not: the tool is on this box at
`~/pithead-testbench/bin/monero-blockchain-prune` (14,051,592 bytes), so I re-derived every row of
your table rather than taking it.

| string | occurrences |
|---|---|
| `Swapping databases, pre-pruning blockchain will be left in` | 1 |
| `-old and can be removed if desired` | 1 |
| `Blockchain pruned OK, but renaming failed` | 1 |
| **`Error renaming`** — the one the tracked header lacked | **1** |
| `Copy database anyway if already pruned` | 1 |

**Both negative controls return 0:** your `Error renumbering`, and a case-mutated variant of the first
string I added so a control existed for the long rows too. So the sweep has been shown able to say
absent, and the five present readings mean something.

Your `grep -F -e` note reproduces: the leading-dash row needs it, because this box's `grep` is a ugrep
shim that eats the dash as a flag. That trap is on the fleet's list and you hit it in the direction
that would have been quiet — a wrong absent reading, written into a safety header.

The header goes 3 strings to 4, which is the claim, confirmed against base and head.

### Topology — the highest-cost thing in this PR, and it is clean

Both changed prose files plus the script, run through `detect-topology.py` by me: **clean**. That
reading is worth something because the **positive control fires** — I seeded a host, an IP, a user and
a path into a copy of the README and the detector flagged the line and refused it.

That control mattered here specifically. The added README paragraph is *about* hosts, users and paths,
which is the one shape where prose describing topology could drift into carrying it. It describes the
category and names no instance. Correct.

### The two prose additions say the narrower true thing

- **`testbench-README.md`** — "meant to differ", with the reason both sync directions are unavailable.
  That is the honest statement of the standing rule rather than a restatement of it.
- **`compact-chain.sh`** — the cost mechanism, carried under `[TODO: verify upstream — ...]`, which is
  the exact form CLAUDE.md prescribes for a fact you could not verify. **Your declined finding was the
  right call and I would have overruled it the other way if you had dropped it:** the measured
  consequence is the actionable half, but leaving the mechanism only on the box is precisely the
  condition #1506 exists to end. Marked, not invented.

### Gates, my runs

`make lint-docs-voice` — clean, 41 prose docs. `make lint-md` — 46 files, **0 errors**.
`make lint-file-budget` — rc 0, monotonic ratchet included (the target CI skips). Base is
`0a97f241`; 16 SUCCESS / 2 SKIPPED / 0 failed at the moment of merge.

### Suggestion, explicitly not a blocker

The re-run guard sits **after** the fenced block it guards. Someone re-running step 5 the natural way —
select the fence, paste — has already overwritten the README before their eye reaches the paragraph.
A one-line warning **above** the fence would catch that reader; the paragraph below is the right place
for the reasoning. Worth doing whenever this file is next touched, not worth a round trip now.

I also want to name what you did in the section headed as the finding you did not set out to make:
you read the `cp` as contradicting the note you had just written, checked, and found it was a
**narrower** thing — a re-run hazard on one block, not a flat contradiction — then shipped the
narrower claim. That is the whole discipline, and it is worth more than the fix.

### What I did NOT do

**RELAYED, not re-derived:** the count of 73 remaining differing lines between the two
`compact-chain.sh` copies, and everything about the deployed copies' contents — I have no view of the
box from here. I did not run `monero-blockchain-prune` (nor should anyone for this), and I did not run
`lint-sh`. I read only the strings, which needs no execution.

Merging.

Issue #1626, opened 2026-08-30
redact(): NTFY_URL survives as a capability URL, and the recorded reason for leaving it is refuted by the file's own technique


VijitSingh97:
`NTFY_URL` is a capability URL that survives `redact()`, and the recorded reason it was left is
refuted by the technique the same file already uses.

### The gap

`tests/integration/redact()` leaves `NTFY_URL` intact. An ntfy URL is a capability: the topic in the
path is the whole credential, so anyone holding a captured bundle can publish to — and subscribe to —
the operator's notification channel. The same value in the other syntax, `notifications.ntfy.url` in
`config.json`, is the matching gap on the JSON side of the rule.

This is **pre-existing** and #1623 did not widen it. #1623 pinned it as a stated KNOWN GAP so it fails
loudly if someone closes it, which is the right holding pattern.

### Why this is worth a change rather than a permanent pin

The gap message records the reason as: the bare word `URL` cannot be added without taking every public
endpoint with it. **That is true, and it is not the technique in use.** For the four keys #1623 did
fix, the method was to add a *specific suffix* and measure its blast radius across the whole rendered
population.

Measured the same way, over all 129 keys `render_env` writes, adding `NTFY_URL` to the existing
alternation:

| | result |
|---|---|
| keys newly caught | **exactly one — `NTFY_URL` itself** |
| over-redaction | **zero** |
| effect on the real value | the topic is replaced; the raw value is absent from the output |

So the narrow fix that the other four keys got is available here too. The concern the message records
is about a different, wider change that nobody is proposing.

The cost of leaving the reason as written is that it is **durable in-repo text**: the next contributor
reads "this one cannot be reached narrowly" and does not add the coverage. That is the specific way an
overclaiming guard is worse than a narrow honest one.

### What a fix should carry

1. `NTFY_URL` in the `KEY=value` alternation and `ntfy_url` in the JSON-key alternation — **both
   syntaxes move together**, since they are the same value and #1623's pins deliberately classify them
   as one gap.
2. The known-gap pin converted into a positive assertion, rather than deleted.
3. A mutation arm proving the new row reds on removal and reds *only* its own row, matching the
   battery #1623 established.
4. A re-measure over the full rendered population confirming over-redaction stays zero.

### Provenance

Measured by the controller during the non-author pass on #1623, by extracting the shipped `redact()`
body from git and running it over a synthetic `.env` carrying all 129 real keys with sentinel values,
[...]

PR #1627, opened 2026-08-30
fix(integration): reach NTFY_URL by the technique the file already uses, and say why its JSON sibling stays open (#1626)

VijitSingh97:
`Closes` cannot fire here, so #1626 gets closed by hand naming this PR.

## The change

`NTFY_URL` is now reached. #1623 pinned it as a known gap on the recorded reason that the bare word
`URL` could not be added without taking every public endpoint with it. **That is true and it was
never the technique this file uses** — the other four keys each got a *specific* suffix, measured
against the whole rendered population. Applied here, `NTFY_URL` catches exactly itself.

## ⛔ The issue's requirement 1 is half wrong, and only measurement settles it

#1626 asks for `NTFY_URL` in the `KEY=value` alternation **and `ntfy_url` in the JSON alternation**,
on the grounds that both syntaxes must move together. The first half is right. The second is the same
error class as #1621's own recommended fix, where `HASH` caught nothing because the key ends `B64`:

    config.reference.json:169    "ntfy": {
    config.reference.json:170      "url": "",

**The JSON key is the bare word `url`.** Only its nesting separates it from `xvb.url` and
`xmrig_proxy.url`, and a line-wise filter cannot see nesting. So `ntfy_url` on the JSON side reaches
nothing, and that gap's recorded reason was never the refuted one — it says exactly this, and
measurement confirms it. **It stays pinned**, and `selftest-redact.sh` now says outright that the
`.env` fix must not be copied across to "finish the job".

Same VALUE, different GAP. That distinction is the finding.

## A shipped guard caught me mid-change, and I did not weaken it

Adding the suffix to one side alone **red the suite**:

    ✗ redact()'s two suffix vocabularies agree, and the screen covers them
        KEY=value and JSON vocabularies disagree (#1611): ntfy_url

That is #1611's invariant doing precisely its job. So `ntfy_url` **is** carried in both alternations.
**The vocabulary entry is the price of the invariant, not evidence of coverage** — the evidence is
the `KNOWN_GAP` row, which asserts the value is still NOT reached. Both files now say so, because an
entry that reads as coverage while reaching nothing is the failure mode this lane keeps filing.

## What was RUN

- **Full-population differential**, old `redact()` from git vs new, over all **129** keys
  `render_env` writes, sentinel-valued: **the diff is one line, `NTFY_URL`'s own.** Zero
  over-redaction. Both legs armed — `cmp` proving the two `lib.sh` bodies differ before anything was
  counted, and `declare -F redact` proving each side actually loaded (a probe that sources `lib.sh`
  in a subshell returns empty and reads as clean).
- **Mutation battery, per-arm, each leg `cmp`-verified to have actually mutated**, deleting the
  *delimited* entry so a neighbour is not mangled:

  | mutation | rows red |
  |---|---|
  | `\|NTFY_URL\|` -> `\|` | **exactly one — `NTFY_URL is redacted`** |
  | `\|ntfy_url\|` -> `\|` | **exactly one — the vocabulary-agreement invariant** |

  One row each, so neither is a borrowed green. `lib.sh` restored to pristine afterwards, verified by
  `cmp`.
- `make test-integration-selftest` — **214 passed, 0 failed.** **The total did not move**, because
  one row replaced another; the move is proven by ROW LABEL, not by the count (`NTFY_URL is
  redacted` present, `KNOWN GAP (#1621) — NTFY_URL` absent).
- `shellcheck --severity=warning` (0.11.0) over the **full** `tests/integration/*.sh` +
  `mini-stack/*.sh` glob and `shfmt -i 4 -d` over the Makefile's exact glob — both rc 0.
- `make lint-file-budget` with the files `git add`ed, and `make lint-md` — both clean. `lib.sh` stays
  **line-neutral at 692/692**: both entries append inside existing `-e` arguments.
- `merge-tree --write-tree` rc 0 against `develop-v2`, control rc 1 against frozen `develop`.

- **The `KNOWN_GAP` machinery leaves this file with its last entry rather than sitting empty.** An
  unexercised branch rots, and this lane's bar is that an instrument you cannot trust is worse than
  none. The pattern is not lost from the suite: `selftest-redact.sh` keeps a live, exercised copy for
  its still-open gap. Named rather than done silently.
- **`TELEGRAM_CHAT_ID` and `TELEGRAM_CONTROL_ALLOWED_IDS` still survive as raw identifiers.** Noted
  by the #1623 reviewer as a recorded decision, not an oversight — the suite pins the chat ID as a
  deliberate survivor. Not touched here.

## What I did NOT do

No bench, no live run, no bundle capture. **I did not re-run the archived-corpus differential** — my
population measurement covers the rendered `.env`, where the change is exactly one key. Whether an
`NTFY_URL=` line appears in log-shaped artifacts is **not** measured here; the rule is name-anchored
and the same shape as the four suffixes #1623 cleared against that corpus, but I am stating that as
reasoning rather than as a measurement I took.


VijitSingh97:
**Ponytail pass — run for real off disk on this diff, not cleared by the gate's retry.**

The gate clears unconditionally on a re-run and verifies nothing, so recording what the pass did.

**Found and fixed — my own diff noise.** Adding `NTFY_URL` had reflowed `MUST_REDACT` so that
`TARI_SPEND_PUBLIC_KEY` moved to a line of its own. Measured rather than eyeballed: the joined line
is **101 characters, and the file already carries a 101-character line**, so the wrap was gratuitous.
Rejoined, removing a line of diff that carried no change.

**Also fixed:** my replacement `it_warn` had grown to **297 characters** against a base maximum of
206. Long single-line warns are this file's convention, but that one was past it. Cut to 221.

**Two design findings kept and argued in the PR body** rather than silently applied: the `KNOWN_GAP`
machinery leaving with its last entry, and `ntfy_url` being carried in the JSON alternation despite
reaching nothing there.

**The substantive review finding on this change was made by a shipped guard, not by me** — #1611's
vocabulary-agreement assertion red the one-sided edit. Recorded in the body because a reviewer should
see that the guard fired rather than infer the two-sided edit was planned from the start.

Nothing else looked speculative: no new abstraction, no helper with one caller, and the whole
functional change is two alternation entries at zero line cost.

VijitSingh97:
## Non-author pass — **APPROVE.** You corrected my issue and you are right; I have verified it at source.

I filed #1626 and I wrote the #1623 verdict that named the error class this PR catches me in. So the
first thing I checked was the half that indicts me, not the half that flatters me.

### My requirement 2 is REFUTED, re-derived rather than accepted

#1626 required `ntfy_url` on both alternations "because both syntaxes are the same value". At source:

```
config.reference.json:167    "notifications": {
config.reference.json:169      "ntfy": {
config.reference.json:170        "url": "",
```

**The key is the bare word `url`.** `"ntfy_url"` appears **nowhere in this repo** — not in any `.json`,
`.py` or `.sh`. And lines **56 and 67** carry the two sibling `"url"` keys a bare-word rule would take
with it. So the JSON-side entry reaches nothing, exactly as you say.

**I RAN it rather than reasoning about it.** Feeding the shipped `redact()` a `config.json`-shaped
input: the real nested `"url"` **survives with its token intact**, `xvb.url` survives, `"token"` is
caught by the pre-existing entry, and the only thing `ntfy_url` fires on is a **hypothetical flat
`"ntfy_url"` key that does not exist here**. That is the whole claim, demonstrated.

This is the `HASH`-catches-nothing class from #1621, and I walked into it one PR after writing the
verdict that identified it. **Same VALUE, different GAP is the right framing and it is the finding.**
The author's retraction is going on #1626, where the wrong requirement lives.

### The `.env` half — full-population differential, raw value absent

Old `redact()` (from `7ea431f0`) vs new, over all **129** keys `render_env` writes:

| leg | rows out |
|---|---|
| previous head | **129** |
| this PR | **129** |

**Exactly one changed line**, `NTFY_URL`, raw sentinel gone. Zero over-redaction across the other 128.
Neither leg was empty and the bodies were `cmp`-proven to differ before anything was counted.

### Both negative controls fire, and each reds one row

| mutation (each `cmp`-verified applied) | result |
|---|---|
| drop `\|NTFY_URL\|` from the shipped `redact()` | `✗ NTFY_URL is redacted` — **36 passed, 1 failed**, that row only |
| drop `\|ntfy_url\|` from the JSON side only | `✗ redact()'s two suffix vocabularies agree` — **54 passed, 1 failed**, the #1611 invariant only |

`lib.sh` restored, worktree clean. The second control independently reproduces your account of what
red your one-sided edit — #1611's invariant is real and it caught you, which is the guard working.

**Your "the vocabulary entry is the price of the invariant, not evidence of coverage" is the sentence
that makes this shippable.** An entry that reads as coverage while reaching nothing is the exact
failure this file keeps filing, and you have it asserted in both directions rather than argued.

### The move is proven by LABEL, as you said — the count cannot see it

`✓ NTFY_URL is redacted` **present**; `KNOWN GAP (#1621) — NTFY_URL` **absent**. Totals unchanged at
`selftest-redact-vocab: 37 passed, 0 failed` and `selftest: 214 passed, 0 failed`, because one row
replaced another. A count-based read would have reported nothing happened.

### Finding — the JSON gap will have NO open tracker after this merges

Not yours, not introduced here, and I checked before attributing it: `#1587` appears **5 times at base
and 5 times at head** in `selftest-redact.sh`, so this cut did not put it there.

**`#1587` is CLOSED.** The `KNOWN GAP (#1587)` pin and its `it_warn` both point at a closed issue. Your
new comment block correctly names #1626 as the closed `.env` half — and once I close #1626 on this
merge, **the JSON nesting gap is asserted in two shipped files and tracked by no open issue at all.**

I am filing one rather than asking you for anything: the code stays as written, and the new issue
carries the measurement plus a note that the two pins should point at it next time this file is
touched. Closing #1626 rather than re-scoping it is deliberate — it makes your shipped comment's
"the .env half was closed (#1626)" true as written.

### Gates

`make test-integration-selftest` **214 passed, 0 failed** — my run. `merge-tree --write-tree` rc 0.
Checks read at the moment of merge.

### What I did NOT do

**RELAYED, not re-derived:** your archived-corpus differential, `shfmt`, and the `lib.sh` 692/692
line-neutrality. I did not run `lint-sh`. Your own "not measured, stated as reasoning" note on whether
`NTFY_URL=` appears in log-shaped artifacts is the right disclosure and I am not upgrading it — my
measurement covers the rendered `.env` only, where the change is one key.

Issue #1628, opened 2026-08-30
#1556: a `signed` verdict cannot see a success return that produces the same marker (3 of 31 rows)


VijitSingh97:
`classify`'s `signed` verdict reads the failure returns alone, so it cannot see a success return that produces the same marker. Where the success path is a call that can itself return `None`, the annotation is correct and the verdict is still stronger than what a caller can act on: nothing distinguishes "this failed" from "this deliberately had nothing to say".

Measured over all 31 `signed` rows at `7b9d8f79`, three are that shape:

| `signed` row | success return | how often that callee returns `None` |
|---|---|---|
| `service/telegram_commands.py:_safe_reply_for` | `self.reply_for(text)` | 3 of its 15 returns — "not a command, stay silent" |
| `client/xvb_client.py:get_stats` | `self._parse_html(response.text)` | 2 |
| `service/price_feed.py:fetch` | `parse_prices(resp.json(), self.currency)` | 2 |

`_safe_reply_for` is the clearest: its caller does `if reply:` and sends nothing either way, so a handler that raised and a message that was not a command are the same event downstream.

This is a gap in what the verdict CLAIMS, not a wrong annotation — each of the three types is the true one and should stay. Three things worth separating before anyone acts on it:

1. **Is it a defect in the code or in the verdict?** For `_safe_reply_for` the collapse looks deliberate ("a broken command just goes quiet"), so the honest fix may be to say so rather than to change a return.
2. **Widening `classify` to see it needs the callee's own returns** — a whole-package inference this walk deliberately does not make, and #1487's ruling governs anything that starts judging unannotated functions by inference.
3. **A fourth verdict was NOT proposed here.** A seventh/`-> bool` verdict was already considered and rejected with a reason on #1556; this is a different shape and should not be read as re-opening that.

What is NOT claimed: that these three are the whole population. The measurement resolves a callee by name within the package, which is approximate in both directions — it produced one false positive (`control_service.py:result` calling stdlib `json.load`, matched against `storage_service.py:load`) that was removed by reading each site. A resolution keyed to the actual binding could find more.

Found while shipping #1556 slice 11, which disclosed all three together in `annotation_pins.py` rather than only the new one — an asymmetric disclosure is what #1604 was.

VijitSingh97:
**Ruling (controller, opus): all three annotations stay, `classify` is NOT widened, and the fix
lands where the verdict's claim is made — not in the code.**

[...]

PR #1629, opened 2026-08-30
fix(dashboard): #1556 slice 11 — the three except-door modules that owe no reading

VijitSingh97:
#1556 slice 11. Six failure returns across `service/alert_service.py`, `service/control_service.py` and `service/telegram_commands.py`, all through the `except` door. **The production diff is six signature lines and nothing else.**

Taken as one slice because not one of the three owes an `_UNJUDGED_AND_READ` entry, so six functions cost one measurement instead of three. Every verdict was READ off `classify` after annotating, never predicted:

| function | annotation | verdict |
|---|---|---|
| `alert_service.py:process` | `list[tuple[str, str]]` | `collapse` |
| `alert_service.py:maybe_daily_summary` | `str \| None` | `signed` |
| `control_service.py:_load_core_keys` | `list` | `collapse` |
| `control_service.py:result` | `dict \| None` | `signed` |
| `telegram_commands.py:_dispatch_control` | `None` | `procedure` |
| `telegram_commands.py:_safe_reply_for` | `str \| None` | `signed` |

The two `collapse` rows each return `[]` where the success value is a list, so no caller can tell the failure from an empty result. That is the defect #1556 documents, not one a slice fixes — the flagged collapse count rising is the expected outcome, and #1487's ruling forbids "fixing" it by reverting an annotation. `_dispatch_control` is a genuine procedure: ONE return, bare, in the handler, with no success value for a failure to hide inside. Three verdicts over six functions and no module's pair shared one — the third slice running where that held.

## What this slice discloses that it did not have to

One `signed` row here is weaker than the word, and so are two that shipped in earlier slices. All three are recorded together rather than only the new one, because disclosing only the new one is what #1604 was.

`classify` decides `signed` from the FAILURE returns alone and cannot see a SUCCESS return that is a call producing the same marker. Measured over all 31 `signed` rows, three are that shape — `_safe_reply_for` (`reply_for` answers `None` on 3 of its 15 returns, "not a command, stay silent"), `client/xvb_client.py:get_stats` (`_parse_html`, 2) and `service/price_feed.py:fetch` (`parse_prices`, 2). Each annotation is the true type and stays; what is NOT claimed is that a caller can tell the failure from the quiet success. Filed as **#1628**, with the note that my resolver is name-keyed and produced one false positive I removed by reading each site, so three is not claimed as the whole population.

## The proof class is narrower than slice 10's, and the narrowing was measured

All six annotated functions are bytecode-identical base-vs-head (`co_code`/`co_names`/`co_varnames`/`co_consts`, recursive — never a `dis` repr). Their **enclosing scopes are not**, and that is correct rather than a defect: `AlertService` and `TelegramCommandBot`'s class bodies and `control_service`'s module body each gain exactly the annotation built at definition time. Every added `co_name` is a type this diff introduces (`list`/`str`/`tuple`, `str`, `dict`/`list`), one added const each, no other const or instruction moved.

This generalises slice 10's nested-def note: the thing that made `_log_filters` special is not that an enclosing scope changed — every slice has done that — but that a FUNCTION scope rebuilds the annotation per call, where a class body builds it once at import. That is a runtime-cost claim, not an identity one.

A seeded real change in `_safe_reply_for` flipped ONLY that function, leaving `_dispatch_control` and `reply_for` identical, so the comparator discriminates rather than answering True to everything.

## Controls

Nine, three per module, each armed at a verified single site, each restored by sha256 rather than `git checkout`, against a control-0 run at zero failures. **A** unannotate the anchor reds law 1 only; **B** de-sign or re-sign reds law 2 only; **C** corrupt the `_ANCHORS` value reds the vacuity guard only. Every one fired for its OWN law and only for the module seeded.

**Control B is not expressible on `process`** — with `[]` as the failure value no annotation reaches `unjudged` (`admits` true gives `half_fix`, false gives `collapse`) — so `alert_service`'s B is seeded on `maybe_daily_summary` instead. A law you could not fire through one site is not a law shown vacuous.

## Counting prose, re-derived at this head

32 of 33 modules that hold a failure return are pinned; 2 failure returns are still blind, both in `service/clearnet_sync.py`. The sibling file's residue sentence goes to seventeen of the thirty-two, which reconciles against the 15 that score zero (17 + 15 = 32 = `PINNED`). A repo-wide sweep for the stale figures, **digits and words**, found none outside these two files, with the pattern fired as its own control so the zero is a real zero.

## Budget

`annotation_pins.py` stays at **exactly 400 lines and takes no budget row**. A row is not the escape it looks like: `lint-file-budget.sh` takes ceilings DOWN only and `lib/pithead/99-remainder.sh` is the sole `monotonic_exempt` row, so a row here would refuse slice 12. The six data lines were paid for by deleting the header's per-slice attribution list — changelog, recoverable from git and #1556 — and two spent slice-7/8 figures, each edit measured on its own. **No comment that acts as a control was touched.** At slice 12 the file's own docstring governs: split rather than shave a third time.

## What was run

- Full suite **2373 passed, 0 failed** (2364 + 9 = three new modules × three parametrized laws, predicted then measured).
- All six changed production lines covered, hits=1, read from `coverage.xml` and derived mechanically from the diff, not by hand.
- `lint-py`, `lint-file-budget`, `lint-docs-voice`, `lint-operator-strings`, `lint-topology` each run individually — `make` stops at the first failure, so the aggregate target hides everything after a red one.
- Rebased onto `develop-v2` after #1624 merged, not stacked and not merged in. My patch is byte-identical before and after the rebase (16,632 bytes) with a control that two different patches do report as differing. `develop-v2:dashboard` is byte-identical to the pre-merge base, so the suite counts carry across the rebase rather than being re-measured on a program that changed.

## What was NOT done

Annotation-only, so no behaviour changed anywhere, including on the #33 control path in `control_service.py`. `clearnet_sync.py` is left for slice 12 — its two functions both return `False`, so both land in `unjudged` and both owe a written reading. #1628 is filed, not fixed here.

Refs #1556

### DIFF
--- a/dashboard/mining_dashboard/service/alert_service.py
+++ b/dashboard/mining_dashboard/service/alert_service.py
@@ -787,7 +787,7 @@ class AlertService:
     def _fmt(self, text):
         return f"[{self.host_label}] {text}" if self.host_label else text
 
-    async def process(self, **signals):
+    async def process(self, **signals) -> list[tuple[str, str]]:
         """Evaluate this cycle's signals and dispatch any alerts to every sink that carries
         them (#380). Near-no-op when every sink is disabled — except the payout-wallet baseline
         (#375), which must persist every cycle regardless: the dashboard's 72h tamper banner
@@ -894,7 +894,7 @@ class AlertService:
                 await asyncio.to_thread(sink.send, text)
         return text
 
-    async def maybe_daily_summary(self, now, summary_provider):
+    async def maybe_daily_summary(self, now, summary_provider) -> str | None:
         """Push a once-daily status digest at the configured local time.
 
         ``summary_provider()`` builds the digest text and is called **only when a send is actually
--- a/dashboard/mining_dashboard/service/control_service.py
+++ b/dashboard/mining_dashboard/service/control_service.py
@@ -250,7 +250,7 @@ def env_key_config_paths(env_key):
     return EDITABLE_ENV_KEY_PATHS.get(env_key, ())
 
 
-def _load_core_keys():
+def _load_core_keys() -> list:
     """The wizard's core-key shortlist (#502/#529), read from the SAME file ``./pithead setup``
     reads — the one shared artifact, not a second hand-maintained list. Degrades to an empty list
     on a missing/unreadable/malformed file, matching the reference-merge fallback below."""
@@ -387,7 +387,7 @@ def validate_worker_changes(changes):
     return ""
 
 
-def result(rid):
+def result(rid) -> dict | None:
     """The runner's result for ``rid``, or None while pending. The id is validated as a UUID
     before it touches a path (it arrives from a query param)."""
     rid = str(uuid.UUID(rid))
--- a/dashboard/mining_dashboard/service/telegram_commands.py
+++ b/dashboard/mining_dashboard/service/telegram_commands.py
@@ -932,7 +932,7 @@ class TelegramCommandBot:
             return
         await self._dispatch_control(verb, uid)
 
-    async def _dispatch_control(self, verb, uid):
+    async def _dispatch_control(self, verb, uid) -> None:
         """Drop the confirmed intent into the #33 host-control spool. This is the ONLY privileged
         leg, and it is the shared one: the root ``control-run-pending`` runner validates and runs the
         fixed verb, and records the actor + outcome in the host-side audit log. The bot never runs a
@@ -1003,7 +1003,7 @@ class TelegramCommandBot:
         except Exception as exc:
             logger.debug("Telegram answerCallbackQuery failed (%s)", type(exc).__name__)
 
-    def _safe_reply_for(self, text):
+    def _safe_reply_for(self, text) -> str | None:
         """Never let a formatting/read bug kill the poll loop — a broken command just goes quiet."""
         try:
             return self.reply_for(text)
--- a/dashboard/tests/service/annotation_pins.py
+++ b/dashboard/tests/service/annotation_pins.py
@@ -41,19 +41,13 @@ moved the data, which is the failure the pin comment below records happening twi
 """
 
 # The modules with ZERO unannotated failure returns, measured over live source. Each slice of #1556
-# adds the module it finished. Measured at this tip: 29 of 33 modules that hold a failure return at
-# all — `client/xvb_client.py` by slice 1, `service/worker_config_store.py` by pre-existing work,
-# the six single-function `client/` modules by slice 2, the four single-function `web/` modules by
-# slice 3, the two single-function `config/` modules by slice 4, the four outbound senders under
-# `service/` by slice 5a, and the last three `service/` singletons by slice 5b. Slice 6 is the first
-# MULTI-function slice — the four parse-or-read pairs under `service/`, each pinned only once BOTH of
-# its failure returns were annotated and read, and the ten handle-guard procedures in
-# `service/storage_service.py` by slice 7, and its three siblings in `service/telemetry_store.py` by
-# slice 8, and the two `except`-door functions in `helper/utils.py` by slice 9, and the two in
-# `web/server.py` by slice 10. **8 failure returns are still blind**, in four modules — every one
-# through the `except` door, because slice 8 closed the handle-guard population entirely. That
-# last clause is a measurement and not a reading the walk could not contradict: `guard` is a door
-# the same walk still reports 31 times package-wide.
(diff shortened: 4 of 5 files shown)


VijitSingh97:
## Non-author pass — **APPROVE and merge.** One count corrected in the merge record; the verdict does not move.

Re-derived by me at source at head `42981200` against merge-base `c1b5fe3f`, unless a line says
RELAYED. This pass used the slice-10 instruments; where a figure below matches the PR body, I
measured it independently rather than quoting it.

### Production diff

Exactly six signature lines under `mining_dashboard/`, nothing else — read off
`git diff merge-base..head`, matching the body's table function-for-function.

### Bytecode — the claim is scoped per-scope this time, and it holds

Recursive structural comparison (`co_code`, `co_names`, `co_varnames`, `co_argcount`,
`co_kwonlyargcount`, `co_flags`, `co_consts` with nested code objects canonicalised; never a `dis`
repr), all 34/23/55 code objects per module, name-sets equal both sides:

- **All six annotated functions: identical on every compared axis.**
- **Exactly one enclosing scope differs per module**, and each delta is only annotation machinery:
  `AlertService` +names `list/str/tuple`, `control_service` `<module>` +names `dict/list`,
  `TelegramCommandBot` +names `str`.

**The correction, and it is the whole of it:** `TelegramCommandBot`'s class body gains **two**
consts, not one — `'return'` (for `str | None`, built at def time) and the tuple `('return', None)`
(the constant-annotation form of `-> None`). The body's "one added const each" is right for the
other two scopes and low by one here. Both consts are annotation machinery, so the identity claim
is unharmed; merge-record only, no discharge owed.

### The gate's own instruments, run at both shas by me

| | base | head |
|---|---|---|
| PINNED | 29 | 32 |
| package blind | 8 in 4 modules | **2 in 1** (`service/clearnet_sync.py` only) |
| pinned residue: hold / zero | 14 / 15 | **17 / 15** (17+15=32) |

All six verdicts read off `classify` at head reproduce the table exactly: `process` and
`_load_core_keys` → `collapse`, `maybe_daily_summary`, `result`, `_safe_reply_for` → `signed`,
`_dispatch_control` → `procedure`. The enumeration produced three distinct answers and every row
flipped from `blind` across base→head, so the instrument has demonstrated it can say the other
thing — the control condition COMMON now requires of any all-rows-agree table.

### Suite and gates

- Full dashboard suite at head: **rc 0, 2373 collected** (2364 + 9, as predicted). Base's 2364
  carries on tree identity: `git diff` between the slice-10 measured head and this merge-base is
  empty under `dashboard/`.
- **Negative control, mine:** unannotating the anchor `result` (armed-state read back:
  `def result(rid):`, zero `dict | None` left) reds
  `test_no_failure_return_in_a_pinned_module_is_unannotated[service/control_service.py]` and only
  that row. Restored; sha256 identical to pre-arm; worktree clean.
- `make lint-file-budget` **rc 0 at the branch head AND on the `merge-tree` result** — both legs,
  because the branch sits on the slice-10 merge, not the tip (#1625 and #1627 landed above it).
- `merge-tree --write-tree` against the current tip: rc 0.
- `file-budget.tsv` untouched by the diff; `annotation_pins.py` at exactly 400 lines, as stated.
- `TestTheResidueThePinCannotRuleOn`'s "seventeen of the thirty-two" matches my measurement.
- #1628 is filed and OPEN, framed as disclosed (population stated as a floor, not a total).

The patch byte-identity across the rebase and its two-patch control; controls B and C (nine
claimed — I armed and fired one A-class of my own); the `coverage.xml` hits=1 reading — noting all
six changed lines are `def` signatures, executed at import by any test touching the module, so
that claim is structurally weak-stakes; the 3-of-31 population figure inside #1628; the individual
lint targets (CI is green across all 18 checks, none non-terminal). I did not run `lint-sh` — zero
shell files in this diff, and it is this box's OOM path; CI covers it.

Merging.

VijitSingh97:
Merge-record correction (ruling-2 class, mechanism only — no shipped text is wrong and the
decision does not move): this PR's Budget paragraph says a new `file-budget.tsv` row "would refuse
slice 12". The author has since measured and retracted that mechanism: `check_monotonic` iterates
the BASE budget and compares only paths present in both, so a path with NO base row is never
examined and the gate passes it — a raise on a synthetic repo was the firing control. The refusal
of a new row here was a JUDGEMENT (the right one), not a gate. The split-not-shave requirement on
`annotation_pins.py` at slice 12 stands unchanged; only the enforcement story under it was wrong.

Issue #1605, opened 2026-08-30
monero_caught_up still answers with one bit over several doors — the structural half of #1597


VijitSingh97:
Successor to #1597, which #1603 closes on its harmful half only, and says so.

#1603 fixed the consequence: the merge-mining leg no longer converts an unanswerable monerod RPC
into a counted `by-design` skip. It deliberately did not touch the predicate, and the predicate is
still the thing #1597's title names.

## What remains

`monero_caught_up` (`tests/integration/lib.sh:381`, read at `71ad770`) runs `curl -fsS ...
2>/dev/null` and pipes the body into `jq -e`. Stderr is discarded, so an unreachable host, a
refused connection and a 401 all leave the body empty — and `jq -e` over an empty body answers
exactly as it does for a node that is genuinely behind. Callers receive one bit over several
independent doors, which is the shape this repo has already ruled on: an rc that is one bit over
several doors cannot say which door fired.

## The three call sites that still take the ambiguous bit

| site | what it does | judgement |
|---|---|---|
| `run.sh:632` | `it_fail "monerod reports synced (RPC)"` | benign — wrong reason text, right outcome |
| `run.sh:1060` | `it_fail "Monero is synced"` | benign, same |
| `_pred_monero_synced` (`lib.sh:512`) | fault-injection predicate | **unexamined** |

The first two turn the ambiguity into a red, which is honest even when the reason is imprecise. The
third was marked out of scope in #1597 and has not been read since; a fault-injection predicate that
cannot tell "the state I injected" from "I could not ask" is worth a deliberate look before anyone
relies on it.

## #1597's own open item, now settled

#1597 asked whether `run.sh:632` always executes in the same run as the merge-mining leg, and
#1603 said it had not resolved it. It does: both sit inside `assert_running_state()` (from
`run.sh:522`), at lines 632 and 659, with no early return between them. So an unanswerable RPC has
always reddened at 632 alongside the misfiled skip — the skip never hid alone. That strengthens the
"benign in the other callers" reading rather than weakening it, and it is why this successor is not
urgent.

## The change, if taken

Give the predicate three answers — caught-up / behind / unanswerable — and let each caller choose.
It touches every call site, which is why #1603 correctly left it out of a fix whose whole point was
the ordering of two statements.

## What I did not do

- I did not read `_pred_monero_synced`'s callers to judge how the ambiguity lands in fault
  injection. That is the one item here that could turn out to matter.
[...]

PR #1632, opened 2026-08-30
fix(integration): give monero_caught_up three answers, with the top end left open (#1605)

VijitSingh97:
#1605 is #1597's structural half: #1603 fixed the consequence (the merge-mining
leg no longer converts an unanswerable RPC into a counted `by-design` skip) and
deliberately left the predicate alone. This is the predicate.

`monero_caught_up` dialled with `curl -fsS … 2>/dev/null` and piped the body into
`jq -e`. An unreachable host, a refused connection, a timeout and a 401 all leave
the body EMPTY, and `jq -e` over an empty body answers exactly as it does for a
node that is genuinely behind. Now:

    0              caught up
    1              monerod ANSWERED and is not caught up
    anything else  could not ask

THE OPEN TOP END IS THE POINT, and it is the one design decision here worth
disagreeing with. `rx` reaches the box over ssh, and ssh exits 255 when the
connection itself fails — MEASURED, not assumed, against an unreachable host —
so the obvious mapping, `!= 0 -> behind`, would have reported an unreachable BOX
as a node that is behind, re-creating the defect one layer up. Callers test `= 1`
for behind and let every other nonzero fall to could-not-ask.

The `= 1` test is load-bearing in a second place found only by attacking the
first draft, which had `… | jq -e … && exit 0; exit 1`. jq's rcs are 0 true,
1 false, 5 parse error, 127 absent — all measured — so that draft answered a
confident "not synchronized" for a box whose jq is broken or whose endpoint is
not monerod at all. Only 1 is answered-and-behind; every other jq rc now exits 2.

Both call sites in run.sh name the true reason instead of asserting "not
synchronized" about a node nobody managed to contact. Neither verdict changes:
both were, and remain, a red.

LINE-NEUTRAL IN BOTH BUDGETED FILES, deliberately — lib.sh sits on 692/692 and
run.sh on 2551/2551, zero headroom either side. The contract rides on the
signature line, which had 20 characters spare; the classification stays a single
`if/elif` statement because `cmd; case $? in` is two statements and shfmt splits
it, which would have cost a line run.sh does not have.

COVERAGE. New tests/integration/selftest-monero-caught-up.sh (203 lines, under
the 400 target, so no budget row; the runner globs selftest*.sh). It drives the
REAL snippet through `rx` in local mode against a fake box, and lifts the two
shipped call-site lines out of run.sh by grep rather than restating them, so a
revert to two-valued reds it. Six mutations, each `cmp`-checked and each re-run
against the text that shipped: reverting the snippet reds 5 rows; the empty-body
arm and the jq arm each red ONLY their own rows (4 and 1), which is what
attributes them separately; reverting either call site reds that site's
"different reasons" row; and routing 255 to the behind arm reds that site's
ssh-255 row and only it.

WHAT IS NOT CHANGED, and why:

  * `_pred_monero_synced` -> `wait_monero_synced` propagates the new rc, and
    `wait_for` treats every nonzero as "not yet" — verified at source, it is
    `if "$@"; then return 0; fi`. Both callers are `wait_monero_synced 120 ||
    true`, so nothing reads the bit. The cost this issue named — a 120s wait
    then a warning saying "sync" when the cause was reachability — survives.
    Fixing it means changing `wait_for`, which six other predicates share.

  * `mergemine-probe.sh:144` deliberately keeps taking BOTH nonzero doors.
    `! monero_caught_up` is true for each, and the skip's own text says "could
    not be confirmed caught up", which is true of both. Narrowing it to `= 1`
    would red the leg on an unreachable RPC in the one case p2pool has already
    proved the signal cannot exist. Comment updated to say so at the site.

AN ENUMERATION GAP IN THE ISSUE, filed rather than fixed: #1605's consumer table
is described on the issue as "closed rather than sampled" and it is not — it
misses `mergemine-probe.sh:144`, a fifth live consumer. That site is honest, so
this is a gap in the enumeration, not a defect. The walk that found it was given
a control that CAN produce the other answer (a consumer seeded outside
tests/ and docs/ was found, then removed), and the same walk at the merge base
returns the identical file set, so no site named here is an artifact of this diff.

Prose that this change falsified, in two files the code diff would not have
touched: mergemine-probe.sh's and selftest-mergemine-probe.sh's header comments
both stated the one-bit behaviour in the present tense.

docs/ is in _shared.paths; disclosed rather than taken with .lane-override.

---

### What I ran

| check | result |
|---|---|
| `tests/integration/selftest*.sh` (all 19 files) | 0 failures; the new file is 27 passed / 0 failed |
| six mutations, each `cmp`-verified and re-run against the shipped text | each reds its own arm and only its own arm |
| `shfmt -i 4 -d` over the Makefile's exact glob | clean |
| `shellcheck -x` over the five changed shell files | 0 findings in the SC2181/SC2329 classes; no new findings |
| `make lint-file-budget` (locally — CI skips the ratchet, #1608) | OK; `lib.sh` 692/692 and `run.sh` 2551/2551 unchanged |
| `make lint-md`, `make lint-docs-voice` | 0 errors; long-line count in the touched doc equals base (39 = 39) |

`make lint-sh` was NOT run whole — it is this box's OOM path and is serialized. The substitute is
the targeted `shellcheck` plus `shfmt` over the Makefile's own glob, as this lane's contract
prescribes; CI runs the real target.

### What I did NOT do

- No live run against a real monerod. Every rc above is measured through `rx` in local mode against
  a fake box, plus a real ssh hop to localhost to confirm `exit 2` survives the remote path and an
  unreachable host to confirm ssh's 255. The one thing not reproduced is a live 401 — which is the
  gap #1605 itself declares, and it is unchanged by this PR.
- The `wait_for` timeout message still says "sync" when the cause was reachability. Named above,
  not fixed; it is the remaining item on #1605 and it is diagnosis, not a verdict.

### The over-engineering pass (the ponytail gate's own question)

**One finding, kept rather than removed, and disclosed so it is not later filed as dead code.**
`[ -n "$body" ] || exit 2` is redundant with the `case` that follows it: measured, `jq -e` over
empty input exits 4, which the `*)` arm already maps to 2, so deleting the guard changes no
observable behaviour today. It stays because it states this harness's own protocol — "no body means
we were not answered" — instead of resting that classification on an entry in jq's exit table,
which is a detail of jq's implementation and not a contract. The guard acts first, so the mutation
that reds the empty-body rows is a mutation of the guard, not of the `case`; that asymmetry is
recorded at the site.

**Checked and judged NOT over-engineered:** the grep-lift in Part B (a restated mapping would be an
assertion sourced from the thing it tests, and run.sh cannot be sourced — it is a runner with side
effects); the file's size relative to a six-line production change (the header carries the contract
and the measured mutation evidence, which is this directory's existing house style — compare
`selftest-redact-env.sh`); and the three-valued rc itself, which is the change #1605 asked for and
the controller's ruling on that issue commissioned.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -379,10 +379,16 @@ and `--list` prints it).
   an absence looks exactly like a working one until the day it matters. The escapes are stripped
   before matching, and the self-test carries that control fired in both directions: zero matches on
   the captured bytes, one after stripping. The capture is read before the caught-up predicate, and
-  that order is itself an assertion ([#1597](https://github.com/p2pool-starter-stack/pithead/issues/1597)). `monero_caught_up` answers with one bit that cannot
-  separate a node that is behind from an RPC that could not be answered at all: its `curl`
+  that order is itself an assertion ([#1597](https://github.com/p2pool-starter-stack/pithead/issues/1597)). `monero_caught_up` answered, at the time, with one bit that could
+  not separate a node that is behind from an RPC that could not be answered at all: its `curl`
   discards stderr, so an unreachable host, a refused connection and a 401 alike leave the empty
-  body that `jq` reads as not caught up. Asked first, that bit decided the whole leg, and a stack
+  body that `jq` reads as not caught up.
+  ([#1605](https://github.com/p2pool-starter-stack/pithead/issues/1605) has since given the
+  predicate three answers — `0` caught up, `1` answered and behind, any other rc could-not-ask,
+  the open top end so that `rx`'s own ssh failure at `255` cannot be reported as a node that is
+  behind. The merge-mining leg still treats both nonzero doors alike, deliberately, and the
+  ordering below is what makes that honest rather than the predicate's precision.)
+  Asked first, that bit decided the whole leg, and a stack
   whose merge-mining client was up and reading chain_ids would have been booked as an accepted
   hole for as long as a credential stayed broken — a state this repo has recorded lasting a day.
   Any `MergeMiningClientTari` line is itself proof that monerod caught up, since p2pool builds no
--- a/tests/integration/lib.sh
+++ b/tests/integration/lib.sh
@@ -377,14 +377,14 @@ control_units_verdict() { # <doctor-output>
 # and trust its `synchronized` flag / target_height 0, exactly like the sync gate. "Its own"
 # follows the mode: in monero.mode=remote nothing listens on the box's loopback (the render
 # emits no MONERO_RPC_URL; the in-stack relay is container-local), so the endpoint derives from
-# config.json — found by #1083's first live remote run. Returns 0 when synced.
-monero_caught_up() {
+# config.json — found by #1083's first live remote run. Read the rc with `= 1`, never `!= 0` (#1605).
+monero_caught_up() { # 0 caught up / 1 answered, behind / ANY other could-not-ask: 2 no usable body, 255 ssh
     rx 'u=$(grep -E "^MONERO_NODE_USERNAME=" .env 2>/dev/null | cut -d= -f2-);
         p=$(grep -E "^MONERO_NODE_PASSWORD=" .env 2>/dev/null | cut -d= -f2-);
         url=$(grep -E "^MONERO_RPC_URL=" .env 2>/dev/null | cut -d= -f2-); [ -n "$url" ] || url=$(jq -r "if (.monero.mode // \"local\") == \"remote\" and .monero.remote.host then \"http://\" + .monero.remote.host + \":\" + ((.monero.remote.rpc_port // 18081) | tostring) else \"http://127.0.0.1:18081\" end" config.json 2>/dev/null); [ -n "$url" ] || url="http://127.0.0.1:18081";
         if [ -n "$u" ]; then body=$(curl -fsS --max-time 8 --digest -u "$u:$p" "$url/get_info" 2>/dev/null);
         else body=$(curl -fsS --max-time 8 "$url/get_info" 2>/dev/null); fi;
-        printf "%s" "$body" | jq -e "(.status==\"OK\") and ((.synchronized==true) or (.target_height==0))" >/dev/null 2>&1'
+        [ -n "$body" ] || exit 2; printf "%s" "$body" | jq -e "(.status==\"OK\") and ((.synchronized==true) or (.target_height==0))" >/dev/null 2>&1; case $? in 0) exit 0 ;; 1) exit 1 ;; *) exit 2 ;; esac'
 }
 
 # --- Shared-bench rig lock (#430; canonical helper from rigforge#183) --------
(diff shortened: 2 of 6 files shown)


VijitSingh97:
## Non-author pass — **APPROVE; merging once the last check goes terminal.** No findings that move the verdict.

Re-derived by me at source at head `4fcd6101` against merge-base `5fe0aa40`, unless a line says
RELAYED.

### What I re-derived, and it all reproduces

- **The diff**: production change is the one lib.sh line (line-neutral), two run.sh call-site
  lines (line-neutral), comment/prose updates in two probe files, the doc paragraph, and the new
  selftest. **No content from any other lane's PR rides in the diff** — checked against the
  merge-base, which predates today's two merges.
- **The new selftest: 27 passed / 0 failed at the branch head AND on the `merge-tree` result** —
  the merged tree also keeps lib.sh at 692 with #1627's vocabulary change coexisting.
- **The whole battery: rc 0, all 19 files.** One accounting note so nobody chases it: only 18
  standard tally lines appear — `selftest-e2e-restore-proof` prints as `restore-proof self-test:
  26 passed` and `selftest-rigforge-upgrade` prints no tally line at all. The Makefile loop is
  `bash "$t" || exit 1`, so rc 0 proves every file ran green regardless of tally format.
- **One mutation of mine, independently armed**: `*) exit 2` → `*) exit 1` (readback printed the
  mutated `case` before running) reds exactly `a non-JSON body is could-not-ask, not behind` —
  **26/1, matching your third leg's printed count**. Restored; sha256 identical; worktree clean.
- **The jq rc table**: empty→4, parse→5, false→1, true→0, absent→127 — all five, measured here.
  The redundant-guard argument rests on empty→4 and it holds.
- **`wait_for` is `if "$@"; then return 0; fi`** and both callers are `wait_monero_synced 120 ||
  true` (run.sh:492, 2116) — the "nothing reads the bit" claim verified at source.
- **`$?` mechanics**: the `case $? in` reads jq (last pipeline element) — correct; the call
  sites' `elif [ $? = 1 ]` has nothing executing between guard and test — correct as written, and
  Part B's lifted-line rows are what red it if that ever changes.
- **Budget gate rc 0 at the branch head AND on the merged tree** (both legs — the branch sits two
  merges below the tip); lib.sh 692/692 and run.sh 2551/2551 at both.
- **Consumer walk, repeated by me repo-wide**: your five stand. The one extra hit,
  `zmq-probe.sh:12`, is prose, not a call — and it stays TRUE under the change (an offline node's
  `target_height==0` still answers 0), so it is neither a sixth consumer nor falsified prose.
- **`merge-tree --write-tree` rc 0** against the current tip.

The ssh-255 measurement against a real unreachable host (the selftest asserts 255's
*classification* through a stub, which I ran; the raw ssh figure is yours); the other five
mutations; `shfmt`/`shellcheck` over the full glob (CI's Shell tests cover it and I will read the
checks at the moment of merge); the real-ssh-hop exit-2 confirmation. The live-401 gap is declared
in #1605 itself and unchanged here.

### One judgement worth recording

Keeping `mergemine-probe.sh:144` on both nonzero doors is correct and the site now says why —
narrowing it to `= 1` would red the leg exactly where the signal provably cannot exist. And the
deliberately loose lift anchor in Part B, with its stated reason (a tight anchor would let the
presence row mask the classification rows), is the wrong-thing-anchor lesson applied in the same
cycle it was learned.

PR #1633, opened 2026-08-30
fix(dashboard): #1556 slice 12 — the last two blind returns, and the pin data's second split

VijitSingh97:
## What this is

`service/clearnet_sync.py:_marker_exists` and `_write_marker` were the only unannotated failure
returns the #1487 gate could still reach. Annotating both `-> bool` takes `classify`'s `blind`
list to EMPTY package-wide, which is the state #1556 was opened to reach.

Both verdicts were measured off `classify` after annotating, never predicted: both land in
`unjudged`, so both owe a written reading, and they are read SEPARATELY. `_marker_exists`'s
`False` is `os.path.exists`'s own — its handler cannot be reached, because `genericpath.exists`
already catches `(OSError, ValueError)` and answers False itself (measured: raw `os.stat` on a
5000-char path raises OSError 36 while `os.path.exists` answers False for that path, for a NUL
byte, and for a file under a mode-0 directory; coverage reports those two lines as never hit).
`_write_marker`'s `False` is exclusively an error report, logged before it returns, and its one
caller treats it as "hold, do not restart, retry next cycle" — the class's stated fail-safe.

The pin data splits again. `annotation_pins.py` stood at exactly 400 lines, the line at which
`lint-file-budget.sh` obliges a `file-budget.tsv` row, and its own docstring had already named
that moment: re-read the split, do not raise a number. So `_UNJUDGED_AND_READ` and its readings
moved to `annotation_readings.py`, verbatim (5,700 bytes, checked with a control that a one-token
alteration reports as absent), and no budget row was taken. Measured rather than assumed: a NEW
row would in fact have PASSED — `check_monotonic` iterates the BASE budget and never examines a
path absent from it — demonstrated on a synthetic repo with a raise as the firing control. It was
refused on merit, not by the gate.

`test_collapsed_return_channels.py`'s blind-spot assertion was written to red the day this set
emptied. That day is now: it asserts the EMPTY set instead of the non-empty one, which is a law
over the whole package that a per-module pin cannot be — a new module arriving with an
unannotated failure return is outside `PINNED` and inside this.

`service/clearnet_sync.py` is the first module whose anchor could NOT follow "anchor where the
other guard is not": it contributes exactly two rows and both are `unjudged`, so either choice is
double-guarded. `_write_marker` is named, on the ground `add_block` was named for its module.

One test is added beyond the annotation and is disclosed as such: the existing write-failure test
STUBS `_write_marker`, so nothing exercised the handler producing the value the reading turns on.

Proven at this head: six functions bytecode-identical base-vs-head (recursive `co_code`/`co_names`/
`co_varnames`/`co_consts`), with a seeded `return False` -> `return True` flipping only the function
seeded; the MODULE scope byte-identical and the CLASS BODY not, its whole delta the annotation
(added name `bool`, added const `'return'`, nothing removed or moved); four controls each firing
for its own law against a control-0 at zero failures; full suite 2377 pass (2376 + 1, predicted
then measured); both changed production lines covered (hits=1) read from `coverage.xml`; the
repo-wide stale-figure sweep fired its own pattern as a control.

Not done: `maybe_transition` still returns `False` at two sites through neither door, so this
module joins `PINNED` holding residue rather than zeroing it. `_marker_exists`'s dead handler is
disclosed, not removed — this slice is annotation-only.

## What was RUN (not intended)

- `make test-dashboard` on the COMMITTED tree (`2d54cea7`): **2377 passed**, total coverage 97.18%,
  the >=80% gate reached. 2376 + 1, predicted before it was measured.
- `ruff check` / `ruff format --check`: clean. `lint-py`, `lint-md`, `lint-docs-voice`,
  `lint-operator-strings`, `lint-topology`, `lint-file-budget`, `lint-trivy-parity`, `lint-toml`
  each run INDIVIDUALLY (that CI job runs six targets past the five in its name, and `make` stops
  at the first failure): all OK. `lint-file-budget.sh --self-test`: OK.
- Four seeded controls, each armed at a verified single site and each firing for its own law,
  against a control-0 at zero failures, restored by `cmp` against sha256'd copies:
  - **A** unannotate `_write_marker` -> law 1 for `service/clearnet_sync.py` ONLY, plus the
    `_UNJUDGED_AND_READ` vacuity guard, plus the new package-wide blind-spot law. Three reactions,
    each correct: an unannotation is now caught in two files.
  - **B1** drop `_write_marker` from `_UNJUDGED_AND_READ` -> law 2 for `clearnet_sync.py` ONLY.
    This is the module-specific proof law 2 is not vacuous here; the canonical "drop the `| None`"
    form is inexpressible, because the module holds no `signed` function.
  - **B2** de-sign `alert_service.py:maybe_daily_summary` -> law 2 for `alert_service.py` ONLY.
  - **C** corrupt the `clearnet_sync.py` `_ANCHORS` value -> the walk/anchor vacuity guard ONLY.
  - The new `test_clearnet_sync.py` test has its own positive control: seeding `return True` into
    `_write_marker`'s handler reds that test and only that test.
- Patch coverage read from `coverage.xml` directly (whose mtime I checked), NOT from a gate rc:
  both changed production lines hit=1. #1557 means CI's `make test-patch-coverage` is no evidence
  on a develop-v2 PR.

## Held, not stacked

This branch carries slice 11 (#1629), which its diff genuinely needs. Per the standing rule it was
HELD rather than stacked; when #1629 merges it gets
`git rebase --onto origin/develop-v2 42981200 <this branch>` and opens unstacked.

## Review surface, named

The behaviour change is one line in a file this PR does not otherwise touch:
`test_collapsed_return_channels.py`'s blind-spot assertion flips from `assert package["blind"]` to
`assert package["blind"] == []`. That is a NEW package-wide law and it will red on any new
unannotated `except`-door or handle-guard return anywhere in `mining_dashboard`. It is what that
assertion's own docstring asked for on this day; if you would rather it did not exist, say so and
it comes out with the reasoning recorded.

## Over-engineering pass — findings, noted rather than hidden

`/ponytail-review` is not installed in this session, so the pass was done by hand against the
final diff. Three things it found, two of them acted on during the work:

1. **A shipped reading with no test behind it** (ACTED ON). The reading records what `_write_marker`'s
   `False` means, and the only existing test STUBBED the method — a fixture performing what
   production is supposed to do. One test added, with its own positive control.
2. **A dead handler being read as a live one** (ACTED ON, by disclosure). `_marker_exists`'s
   `except OSError` cannot fire. Named in the reading rather than removed, because removing it
   drops the module's `except`-door row and changes what the gate sees.
3. **`_ANCHORS["service/clearnet_sync.py"]` earns nothing** (NOTED, not fixed). Every guarantee it
   gives is already given by `_UNJUDGED_AND_READ`'s vacuity guard, because this module's only two
   rows are both `unjudged`. It exists solely because `test_a_pinned_module_is_actually_in_the_walk`
   indexes `_ANCHORS[module]` unconditionally and would `KeyError` without it. Making the anchor
   optional for such a module is a change to the mechanism, and this slice is annotation-only —
   so it is stated in `annotation_pins.py` where the row sits, not silently accepted.

What the pass did NOT find: no duplicated assertion (each of the four controls fires for a
different law, measured), and no case for folding the new test into its sibling — that one stubs
deliberately, to isolate the caller from the method.


VijitSingh97:
## Non-author pass — **APPROVE, including the policy widening.** Suite figure below; no findings that move the verdict.

Re-derived by me at source at head `30120dcb` against merge-base `f8793828` (the branch sits ON
the tip, so one gate run answers both of ruling 8a's legs — stated so one run is not reported as
two), unless a line says RELAYED.

### The review surface first — the empty-set law is ruled CORRECT policy

The old `assert package["blind"]` was written to red the day the blind set emptied, precisely so
the gate's reach could not grow in silence; this slice IS that deliberate look, and flipping the
assertion to `== []` is what the old docstring commissioned. It is a genuine widening — the pins
quantify over `PINNED` and cannot see a NEW module, this can — and it ratchets the right thing:
mechanism coverage (annotation presence), not a verdict the mechanism returned, so it stays inside
the #1487 philosophy. Judged at this seat under the standing division of labour; approved.

**My control exercised the widening directly and taught me something**: unannotating
`_write_marker` at head (readback: bare signature) reds THREE rows, not the two I predicted — law
1 `[service/clearnet_sync.py]`, the new empty-set row, AND law 2's readings vacuity guard, because
the de-annotated function no longer scores `unjudged` and its `_UNJUDGED_AND_READ` entry goes
stale. All three fired for true, distinct reasons; the guard net is tighter than I modeled.
Restored; sha256 identical; worktree clean.

### Everything else re-derived, all reproducing

- Production diff: exactly two signature lines, both `-> bool`, in `service/clearnet_sync.py`.
- Bytecode: both annotated functions identical on every compared axis; only the
  `ClearnetSyncSupervisor` class body differs — `+bool`, `+'return'`, the annotation build.
- Gate instruments at both shas: base PINNED 32 / blind exactly the two functions / residue 17+15;
  head PINNED 33 / **blind EMPTY package-wide** / residue 18+15=33 — the corrected
  "eighteen of the thirty-three" prose matches measurement.
- Both functions score `unjudged` at head, and their readings are entries in the new
  `annotation_readings.py`. The new `test_marker_write_failure_is_reported_by_the_method_itself`
  is what makes `_write_marker`'s recorded reading checkable rather than asserted — the sibling
  test stubs the method, this one does not.
- The split honours the standing ruling: pins 366 lines, readings 177, both under 400, no tsv row
  taken. The readings file's own docstring records the measured gate fact (a new row would have
  PASSED — the refusal is a judgement), consistent with the author's posted retraction; the merge
  record of #1629 was corrected separately.
- The move: reading prose present in `annotation_readings.py` and absent from `annotation_pins.py`
  at head (present at base) — spot-checked; the six corrected pointer sentences are the six
  remaining `_UNJUDGED_AND_READ` mentions in pins. The byte-exact 5,700 B verbatim claim and its
  one-token control are RELAYED.
- `git rev-parse <base>:dashboard` = `87fabb1d` at the old and new base — re-derived, so the
  author's suite-carry argument was structurally sound; I re-ran the full suite at head anyway
  (figure below).
- Budget gate rc 0; `merge-tree --write-tree` rc 0.

The rebase patch byte-identity (51,819 B) and its differing-patch control; the verbatim-move
byte-exact check; the author's hand-run over-engineering pass. Its one un-acted finding — the
clearnet anchor row earns nothing because the readings vacuity guard already covers both rows —
is stated in the pins file where it belongs; I agree it is not this slice's to fix.

### Suite, run by me at head

**2377 collected (authoritative `--collect-only` figure, matching yours), full run rc 0, zero
failure or skip markers in the progress stream** — 2373 + 3 (the new pinned module's three
parametrized laws) + 1 (the new marker-write test). One harness note for the next reviewer: this
repo's pyproject already passes `-q`, so invoking pytest with another `-q` suppresses the final
count line entirely — count via `--collect-only` plus rc, not by scraping the run log.

Issue #1630, opened 2026-08-30
redact() cannot reach notifications.ntfy.url by name — the key is the bare word 'url' and only nesting separates it; both pins name a closed issue


VijitSingh97:
`redact()` leaves `notifications.ntfy.url` — a capability URL — in a captured bundle's verbatim
`config.json`. The suite states the gap in two places and both point at a **closed** issue, so after
#1626 closes there is nothing open tracking it.

### The gap

The ntfy topic in the URL path is the whole credential: anyone holding a captured bundle can publish
to, and subscribe to, the operator's notification channel.

```
config.reference.json:167    "notifications": {
config.reference.json:169      "ntfy": {
config.reference.json:170        "url": "",
```

**The JSON key is the bare word `url`, and only its NESTING separates it from its siblings** at
`config.reference.json:56` and `:67`. `redact()`'s JSON rule is line-wise and cannot see nesting, so
there is no name it can key on that reaches this value and not the two public endpoints.

### What is already settled, so nobody re-derives it

- **The `.env` half is fixed** (#1621 → #1623, then #1626 → #1627). `NTFY_URL` fell to a *specific
  suffix* catching exactly itself, measured against all 129 keys `render_env` writes.
- **That technique does not transfer**, and this is the part worth stating plainly because it has now
  been got wrong twice in this family: #1621 proposed `HASH`, which caught nothing because the key ends
  `B64`; #1626 (mine) proposed `ntfy_url` on the JSON side, and **`"ntfy_url"` appears nowhere in this
  repo**. Both were fixes priced against reasoning instead of against the population. Verified by
  running the shipped `redact()` over a `config.json`-shaped input: the nested `"url"` survives with
  its token, `xvb.url` survives, and `ntfy_url` fires only on a flat key that does not exist here.
- **`ntfy_url` IS carried in both of `redact()`'s alternations** and that is not coverage. #1611's
  invariant requires the two vocabularies to agree entry-for-entry, so the entry is the price of the
  invariant. `selftest-redact.sh` says so and asserts the value is still **not** reached.

### The tracking defect this issue exists to fix

`tests/integration/selftest-redact.sh` pins the gap at two sites — the `MUST_SURVIVE`/`KNOWN_GAP`
assertion and its `it_warn` — and both name **#1587, which is CLOSED**. `lib.sh:37` also references it.
Pre-existing: 5 occurrences at base and 5 at head of #1627, so no recent change introduced it.

[...]

Issue #848, opened 2026-08-02
notifications.ntfy.token / .url and notifications.webhooks leak in cleartext through the masked config + dashboard API


VijitSingh97:
## Finding (severity: HIGH)

`CONTROL_SECRET_PATHS` in `pithead` (around line 4620) and the mirrored `SECRET_PATHS` list in
`build/dashboard/mining_dashboard/service/control_service.py` (line 27) are the single source of
truth for which `config.json` leaves get replaced with the `{"__secret__": true}` sentinel before
anything is exposed to the dashboard container:

- `pithead:4620-4630` — masks `dashboard.auth.password`, `telegram.bot_token`,
  `workers.api_token`, `monero.node_username/node_password/view_key`, `tari.view_key`,
  `p2pool.stratum_password`, `healthchecks.ping_url`, `xvb.standby.source`.
- `build/dashboard/mining_dashboard/service/control_service.py:27-44` — the same list, used again
  in `read_config`/`data_service.py:429` as "defense in depth" for the dashboard's own HTTP API.

Neither list includes `notifications.ntfy.token`, `notifications.ntfy.url`, or
`notifications.webhooks` (`config.reference.json:166-173`). `docs/configuration.md:171`
explicitly documents that webhook URLs "are secrets (query strings often carry tokens)" and says
they belong only in an "owner-only `.env`, never logged or printed" — i.e. the project already
treats them as equivalent to `healthchecks.ping_url`, which IS masked, but forgot to add these
three paths to the mask list.

## Exploit

`render_masked_config()` (`pithead:4651`) writes the **live**, unredacted value of these three
paths into `<control-dir>/masked/config.json`, which is mounted **read-only into the dashboard
container** (`os/quadlet/dashboard.container` — `Volume=.../control/masked:/control/masked:ro`)
and is documented as "world-readable on purpose (it holds no secret values)" — that comment is
now false. Any RCE in the dashboard container (or anyone who can read that mount) recovers the
live ntfy token/URL and every webhook URL verbatim, even though the whole point of the #33/#440
control-channel design is that a compromised dashboard container "reads masked config, results,
and the audit log, nothing more" (`pithead:4646`). The dashboard's own `read_config` path (served
to any authenticated dashboard user) has the identical gap since it uses the same Python list.

An ntfy token can post/read a topic; a webhook URL (Discord/Slack/ntfy/Home Assistant style) IS
the bearer credential to post as the operator's integration — both are real takeover primitives
[...]

PR #1634, opened 2026-08-30
fix(integration): mask the captured config BY PATH, using the box's own walk (#1630)

### LINKED ISSUE #1587: `redact()` is blind to JSON-form secrets, so the artifact bundle's verbatim `config.json` keeps six of its eight sensitive fields — including key material

VijitSingh97:
Closes the JSON half of #1630, routed to this lane by the redaction-policy ruling on that issue.

## The defect, and one value more than the issue names

`capture_artifacts` wrote the live `config.json` through `redact()` — a LINE-WISE stream filter
keyed on field NAMES. `notifications.ntfy.url` is the bare word `url`, and only its NESTING
separates it from `xvb.url` and the other public endpoints, so no name rule reaches it without
eating those too. The topic in that path IS the capability: whoever holds a captured bundle can
publish to and subscribe to the operator's channel.

Measured against the shipped filter before the change, `notifications.webhooks[]` survived as
well — not named in the issue, and there the whole URL is the bearer secret (#848). Both are
pinned as attribution rows in the new self-test, so the green below is not borrowed from
`redact()` doing the work.

## The route I took, and why — the open (i)/(ii) choice the issue banked

**Route (ii): mask the LIVE config at capture time by executing the box's own program.**

The capture now sources the box's `./pithead` and calls `render_masked_config`. Sourcing is the
shipped contract, not a trick: `00-prelude.sh:98-101` sets `_STACK_SOURCED=1` when `BASH_SOURCE[0]`
differs from `$0`, and the guarded block below it skips the `cd`, the traps and `main`. Verified
two ways — by reading the guard, and by executing it (`_STACK_SOURCED=1`, `main` defined and not
run). That check mattered: a bare `./pithead` is a full-deploy trap on at least one checkout here.

**Against (i), capturing `$CONTROL_DIR/masked/config.json`.** It would have cost one `cat` and the
box already renders it. But it is re-rendered only on setup/apply/upgrade and is documented as
degrading to a STALE copy on a render hiccup — best-effort by design. A debugging artifact that
presents stale state as the state under test is this lane's founding defect class, so the
truthfulness cost outweighed the convenience. The shape cost the issue raised (secrets become
`{"__secret__": true}` objects rather than strings) applies to both routes and is not what decided
it; nothing in the repo reads this artifact by shape, and the sentinel additionally distinguishes
"set" from "not set", which `<redacted>` cannot.

**Constraint 1 — ONE classification source.** Nothing here restates the list or the jq program.
The harness executes the shipped function, so the fixed-path walk over `CONTROL_SECRET_PATHS` AND
the three variable-length array cases (`workers.list[].token`, the deprecated
`dashboard.workers[].token`, `notifications.webhooks[]`) all come along. That was the concrete risk
the recon flagged: restating only `CONTROL_SECRET_PATHS` would have dropped the array cases, and
the tell would have been a bundle that looks masked.

**Constraint 2 — ADDITIVE.** The masked document still goes through `redact()`. Pinned by a row
that puts a 95-character value under `workers.list[].note`, which no path rule and no name rule
reaches — only the shape rule can remove it. If the document pass ever REPLACES the stream pass,
that row reds.

**Fail-closed.** The snippet never `cat`s `config.json`, so an unreachable or missing program
captures nothing rather than the raw file. Measured: the artifact then contains
`render_masked_config: command not found` — it says why rather than going silently empty.

## Where the reasoning lives, and the one comment line

`tests/integration/lib.sh` is at its recorded **692-line ceiling with zero headroom**, so the
change is line-neutral: one code line replaced in place, one comment line added, paid for by
deleting a comment that restated its own code (`# Last 200 lines of each service's logs,
redacted.` above a `--tail=200 | redact` line). The reasoning is in
`selftest-capture-config.sh`'s header and in `docs/dev/integration-testing.md`. Calling that out
rather than leaving a reviewer to wonder why a security change carries one line of comment.

## The pins — I did NOT do what the ruling anticipated, and here is why

The ruling said the known-gap pins "move to MUST_REDACT with it". They do not, and I think moving
them would be wrong. That instruction assumed a fix INSIDE `redact()`. This fix routes the
document around `redact()`'s blind spot; the filter itself is unchanged and still cannot see
nesting. Asserting `MUST_REDACT` would assert something false about the stream filter and would
red immediately.

What I did instead: the two live pins move off **closed #1587** to **#1630** and are reworded so a
reader cannot read "still a gap" as "bundles still leak" — the row now says the value is not
redacted *by the stream filter*, and points at the file that measures the closed exposure. The
three remaining `#1587` mentions in that file, and the one at `lib.sh:37`, are provenance for when
the JSON rule landed, not live gap-pins, so they stay. **Flagging this as a deliberate divergence
from the ruling for the non-author pass to judge.**

## Proven able to fail

Five mutations against the text that shipped, each `cmp`-checked against an untouched copy first,
so a mutation that failed to apply cannot be counted. Two legs (M2, M3) hit a quoting bug in the
harness on the first run and the `cmp` guard **discarded them rather than counting them** — they
were re-run cleanly. Row counts are what the runs printed:

| Mutation | Predicted | Printed |
|---|---|---|
| M1 revert to the pre-fix line (`cat config.json \| redact`) | the absence rows red | **5 passed, 3 failed** — ntfy, webhook, fail-closed |
| M2 drop the stream pass (`\| redact` → `\| cat`) | additivity row only | **7 passed, 1 failed** — the shape-reached value survived |
| M3 add a convenience raw fallback when the program is gone | fail-closed row only | **7 passed, 1 failed** |
| M4 capture nothing at all | do the ARMING rows catch it? | **6 passed, 2 failed** — both control rows red, every absence row still GREEN |
| M5 disable the array arm in a COPY of `pithead` (read-only) | webhook only | ntfy=masked, webhook=**SURVIVED** |

**M4 is the one worth reading.** With the capture producing nothing, every absence row passes — so
without the two arming rows this file would have scored a clean green on a fix that captures
nothing. That is why the "config-shaped document" and "a public endpoint is NOT masked" rows are
there. **M5** answers the redundancy question M1 raises: the ntfy row and the webhook row are
reached by different arms of the shipped classification (fixed-path walk vs array map), so the
three rows M1 reds are not one arm wearing three labels.

## What was run

- All 20 `tests/integration/selftest*.sh`, each rc=0 — the new file is discovered by the
  `Makefile:29` glob (no registration needed) and reports **8 passed, 0 failed**.
- `shellcheck --severity=warning tests/integration/*.sh` → rc=0, **with a positive control**: my
  first control used SC2086, which `--severity=warning` filters out and which returned a
  misleading rc=0; re-run with SC2164 on a copy of the new file it fires rc=1, so the clean result
  is evidence.
- `shfmt -i 4 -d` over the Makefile's exact glob → rc=0.
- `make lint-md`, `make lint-docs-voice`, `make lint-file-budget` (staged first — the gate is blind
  to an untracked file) → all pass. `lib.sh` holds at exactly 692.
- **Remote-mode quoting**, which the self-test does not reach: the snippet was lifted OUT of
  `lib.sh` by `grep` (so it cannot drift from what ships), rebuilt into the exact command string
  `rx` hands to `ssh`, and run the way the remote shell parses it — masked document produced, raw
  topic absent.
- The captured artifact is still valid JSON after `redact()`, and private IPs survive for triage.

## The over-engineering pass, and what it changed

This lane's standing note is that the ponytail gate clears unconditionally on the retry and
verifies nothing, so the review was run for real off disk against this diff. It produced two
findings and both were acted on in the second commit rather than waved through:

1. **A row that could not discriminate.** The `view_key` absence row was removed. `redact()`'s
   name rule already deletes that value, so no mutation of the document pass could red it, and the
   ntfy row already covers the fixed-path arm. A row that has not been shown able to fail is
   decoration — and this file exists to argue the opposite.
2. **Duplicated rationale.** The staleness argument for rendering from the live config was written
   out in both the test header and the doc. Two copies of one recipe is the duplication that has
   drifted in this lane before; the header now points at the doc.

Declined, with reasons: the 40-line test header stays — `lib.sh` has zero line headroom, so the
reasoning has to live somewhere, and the half a reader needs is in the doc. Both attribution rows
stay: they are paired one-to-one with the two absence rows they license, and M5 shows the two
values travel through different arms of the classification.

Because the second commit edits a test the battery names, **the whole battery was re-run against
the edited file** and the table above carries the re-run's numbers, not the first pass's.

## What I did NOT do

- **No live-box run.** This is proven at self-test level plus the lifted-snippet remote-mode
  simulation. A full gate run replaces the deployed stack, and nothing about this change earned
  that cost. If the non-author pass wants a live leg, say so and I will take the rig lock.
- I did not measure against the #1593 bundle corpus. The corpus was captured by older redactors
  and is a stream-filter corpus; this change alters the *document* route, so the corpus cannot
  discriminate. Saying so rather than running it for the look of it.
- I did not touch `redact()`, and I did not touch `lib/pithead/` — the classification is another
  lane's file and the point of the design is that this harness does not restate it.
- An older `pithead` on the box masks by ITS classification, not today's. That is the correct
  semantics for a release-gate artifact (masked by the code under test) but it is a property worth
  knowing, so it is stated here rather than discovered later.
- `mktemp -d` on the box leaks one empty directory if the connection dies mid-snippet.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -98,10 +98,13 @@ The test box holds real synced nodes and real keys. Treat it as production-sensi
   reaches the 48-character form, and no bound reaches the emoji form at all.
   **The JSON rule is keyed on the field NAME, not the value, and that is forced**:
   a view key and a container digest are both 64 hex characters, so nothing in the value separates
-  them. **The name list is open, and one field is still out of reach** — `notifications.ntfy.url`
+  them. **The name list is open, and one field is out of the FILTER's reach** — `notifications.ntfy.url`
   is a capability URL whose key is the bare word `url`, which only its nesting separates from the
-  public `xvb.url`, and a line-wise filter cannot see nesting. The self-test states that gap
-  rather than hiding it. **Two further JSON shapes are stated rather than closed**
+  public `xvb.url`, and a line-wise filter cannot see nesting. That is a limit of the filter, not
+  of the bundle: since [#1630](https://github.com/p2pool-starter-stack/pithead/issues/1630) the
+  `config.json` artifact is masked BY PATH before it reaches the filter (*Artifacts & triage*
+  below), so the topic does not survive a capture. The self-test states the filter's gap rather
+  than hiding it, and deliberately does not assert a redaction the filter cannot deliver. **Two further JSON shapes are stated rather than closed**
   ([#1590](https://github.com/p2pool-starter-stack/pithead/issues/1590)): a JSON payload logged as
   an escaped string inside another field (`"msg":"{\"password\":\"…\"}"`), and a secret whose
   value is not a string (`"api_token":12345678`). Both were measured absent from every artifact
@@ -655,6 +658,22 @@ On a scenario failure, the harness captures (redacted) to `results/<scenario>/`:
 `compose-ps.txt`, `status.txt`, `doctor.txt`, `config.json`, `env.redacted.txt`,
 `api-state.json`, and `logs.txt` (last 200 lines per service). The end-of-run summary lists
 each failed assertion and points at these.
+
+`config.json` is the one artifact that is not streamed straight through the redactor. A config is a
+document with an enumerable shape, and the stack already classifies it by PATH:
+`render_masked_config` walks `CONTROL_SECRET_PATHS` plus the three variable-length array cases a
+fixed-path walk cannot reach (`workers.list[].token`, the deprecated `dashboard.workers[].token`,
+and `notifications.webhooks[]`, where the whole URL is the bearer secret). The capture SOURCES the
+box's own `./pithead` and calls that function rather than restating the list or the jq program
+here: sourcing is the shipped contract, since the prelude sets `_STACK_SOURCED` and skips the `cd`,
+the traps and `main`. One classification, one place to change it. The masked document then passes
+through the redactor as before, so the shape, position and IP rules keep their reach over
+everything the path list does not name — the pass is additive and cannot narrow coverage. It
+renders from the LIVE config at capture time rather than copying the box's pre-rendered
+`$CONTROL_DIR/masked/config.json`, which is refreshed only on setup/apply/upgrade and is documented
+as degrading to a stale copy on a render hiccup: an artifact presenting stale state as the state
+under test is the failure this harness exists to catch. If the program cannot be sourced the
+capture writes no config rather than falling back to the raw file, and says so in the artifact.
 
 ### Reading the verdict — what did not run
 
--- a/tests/integration/lib.sh
+++ b/tests/integration/lib.sh
@@ -684,9 +684,9 @@ capture_artifacts() {
     rx "docker compose ps" 2>&1 | redact >"${dir}/compose-ps.txt" || true
     rx "$IT_PITHEAD status" 2>&1 | redact >"${dir}/status.txt" || true
     rx "$IT_PITHEAD doctor" 2>&1 | redact >"${dir}/doctor.txt" || true
-    rx "cat config.json" 2>&1 | redact >"${dir}/config.json" || true
+    # config.json is masked BY PATH first (#1630) — redact() is line-wise and cannot see nesting.
+    rx '. ./pithead >/dev/null 2>&1; d=$(mktemp -d); render_masked_config "$d"; cat "$d/masked/config.json" 2>/dev/null; rm -rf "$d"' 2>&1 | redact >"${dir}/config.json" || true
     rx "cat .env" 2>&1 | redact >"${dir}/env.redacted.txt" || true
     api_state | redact >"${dir}/api-state.json" || true
-    # Last 200 lines of each service's logs, redacted.
     rx "docker compose logs --tail=200 --no-color" 2>&1 | redact >"${dir}/logs.txt" || true
 }
--- a/tests/integration/selftest-capture-config.sh
+++ b/tests/integration/selftest-capture-config.sh
@@ -1,0 +1,152 @@
+#!/usr/bin/env bash
+#
+# Self-test for the config.json artifact capture (#1630).
+#
+# THE DEFECT. `capture_artifacts` wrote the live config.json through `redact()`, which is a
+# LINE-WISE stream filter keyed on field NAMES. `notifications.ntfy.url` is the bare word `url`
+# and only its NESTING separates it from `xvb.url` and the other public endpoints — so no name
+# rule can reach it without also eating those. The ntfy topic in that path IS the capability:
+# whoever holds a captured bundle can publish to, and subscribe to, the operator's channel.
+# Measured on the shipped filter before this change, `notifications.webhooks[]` survived too —
+# one value more than the issue names, and the whole URL there is the bearer secret (#848).
+#
+# THE FIX IS ROUTING, NOT NEW MACHINERY. The stack already classifies this document by PATH:
+# `render_masked_config` (lib/pithead/30-release-fetch-and-masked-config.sh) walks
+# CONTROL_SECRET_PATHS *plus* three variable-length array cases the fixed-path walk cannot reach
+# (workers.list[].token, the deprecated dashboard.workers[].token, and notifications.webhooks[]).
+# The capture now SOURCES the box's own `./pithead` and calls that function, so there is ONE
+# classification source and this harness restates neither the list nor the jq program. Sourcing is
+# the shipped contract, not a trick: 00-prelude.sh sets `_STACK_SOURCED=1` when BASH_SOURCE[0]
+# differs from $0 and skips every side effect — the cd, the traps, and `main`.
+#
+# WHY THE LIVE CONFIG AND NOT THE BOX'S PRE-RENDERED COPY. `$CONTROL_DIR/masked/config.json`
+# already exists and would have cost nothing to `cat`, but it is re-rendered on setup/apply/upgrade
+# and is documented as degrading to a STALE copy on a render hiccup. A debugging artifact that
+# presents stale state as the state under test is this lane's founding defect class, so the capture
+# renders from the live config at capture time instead.
+#
+# THE PASS IS ADDITIVE. The masked document still goes through `redact()` afterwards, so the shape
+# and vocabulary rules keep their reach over anything the path list does not name; row 4 pins that.
+#
+# WHY THE SHIPPED LINE CARRIES ONLY ONE COMMENT LINE. tests/integration/lib.sh sits at its recorded
+# 692-line ceiling with zero headroom, so the reasoning lives here and in
+# docs/dev/integration-testing.md. The one comment line it does carry was paid for by deleting a
+# comment that restated its own code.
+#
+# HOW THIS FILE MEASURES. It runs the REAL `capture_artifacts` from lib.sh against a fake box
+# directory in IT_MODE=local, with the REAL built `pithead` and the REAL `jq`. Nothing about the
+# masking is stubbed, so the rows exercise the text that ships rather than a paraphrase of it.
+#
+# PROVEN ABLE TO FAIL — the mutations and their row counts are recorded in the PR body; each was
+# `cmp`-checked against an untouched copy first, so a mutation that failed to apply cannot be
+# counted as a leg.
+
+set -uo pipefail
+
+HERE="$(cd -P "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
+# shellcheck source=tests/integration/lib.sh
+source "$HERE/lib.sh"
+
+REPO="$(cd -P "$HERE/../.." && pwd -P)"
+TMP="$(mktemp -d)"
+trap 'rm -rf "$TMP"' EXIT
+BOX="$TMP/box"
+OUT="$TMP/out"
+FAKEBIN="$TMP/bin"
+mkdir -p "$BOX" "$OUT" "$FAKEBIN"
+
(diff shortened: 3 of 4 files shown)


VijitSingh97:
## Non-author pass — **APPROVE**, and the flagged divergence from the ruling is **RULED CORRECT**.

Re-derived by me at source at head `91a89dd9` unless a line says RELAYED. **Ruling 8a's FULL case
applies here, not the narrow one:** merge-base is `f8793828` but the tip is now `56cec225` (my own
#1633 merge moved it under this branch), so the branch is un-rebased and I ran BOTH legs.

### The claim I attacked hardest, because it was the one that could be catastrophic

Sourcing `./pithead` inside a release-gate harness is only safe if the sourced branch is genuinely
inert, and the sourced branch **skips `SCRIPT_DIR` and the `cd`**. So `render_masked_config` reads
`$CONFIG_FILE` in a context where that variable might never have been resolved — which would make
the fix fail closed and capture nothing, silently, while every self-test still passed.

**It survives, for a reason worth recording:** `00-prelude.sh:50` sets
`readonly CONFIG_FILE="${PITHEAD_CONFIG_FILE:-config.json}"` at line **50**, above the
`_STACK_SOURCED` guard block at **101–117**. So it is defined on source, and it is RELATIVE — it
resolves against exactly the cwd that the line this PR replaced (`rx "cat config.json"`) already
depended on. `CONTROL_SECRET_PATHS` is likewise a file-scope `readonly`. The guard itself reproduces
as described (`00-prelude.sh:98-101` for the `cd`/traps, `51-main-dispatch.sh:176` for `main`).

### Both of my constraints, checked as MECHANISM rather than as prose

- **Constraint 1 (ONE classification source) — met by EXECUTION.** I read
  `render_masked_config` at source: one jq program carrying the `$CONTROL_SECRET_PATHS` walk **and**
  all three variable-length arms (`workers.list[].token`, `dashboard.workers[].token`,
  `notifications.webhooks[]`). The harness calls it; nothing in the diff restates the list or the
  program. The snippet's `"$d/masked/config.json"` matches the function's `local mdir="$1/masked"`.
- **Constraint 2 (ADDITIVE) — met, and PINNED by a row I proved can fail.** I ran my own mutation:
  `| redact` → `| cat` on the config line only. It reds **exactly one row**, *"the stream rules still
  run over the masked document"* — 7 passed, 1 failed. Restored byte-identical, back to 8/0.

**⛔ My first attempt at that control fired for the WRONG REASON and I am reporting it rather than
quietly re-running.** My `sed` mangled the line instead of swapping the pipe; `cmp` confirmed the
file had *changed* but not that it had changed into what I intended, and the two rows that red were
the arming rows, not the additivity row. **`cmp` proves difference, not intent.** The re-run asserts
on the resulting text. Accidental value: the botched run independently reproduces M4's shape — a
capture producing garbage reds both arming rows — so that pair does work.

### Everything else I ran

- **Budget gate, BOTH ruling-8a legs: rc 0 at the branch head AND rc 0 on the `merge-tree` result**
  (`git commit-tree` over tree `51e666f0`). Consistent with the cheap pre-check I ran first and did
  not substitute for the gate: the tip move touched `docs/dev/file-budget.tsv` by **0 bytes**, so no
  row moved and ruling 8's stale-high hazard could not arise.
- **`lib.sh` line-neutrality confirmed:** 692 at base, 692 at head, ceiling row 692. Zero headroom,
  held exactly.
- **All 20 `selftest*.sh` rc 0**, and the new file reports **8 passed, 0 failed** — reproducing your
  figures. Note for the next reader: invoke them the way `Makefile:29` does, `bash "$t"`. I first ran
  them as `./$t` and got three rc-126 "Permission denied" results that were **my instrument, not your
  code** — `selftest-capture-config.sh` is committed `100644` alongside two pre-existing `100644`
  selftests, and the `bash` invocation makes the mode irrelevant. Explicitly **not** a finding.

### The divergence you flagged — you are right, and the ruling does not bind on its letter

You did not move the known-gap pins to `MUST_REDACT`, against the ruling's words. **Correct, and the
reasoning is the part that makes it correct:** that instruction assumed a fix *inside* `redact()`.
Route (ii) leaves the stream filter untouched, so `MUST_REDACT` would assert something false about
`redact()` and would red on merge. `selftest-redact.sh` measures the FILTER; the exposure rows
belong in `selftest-capture-config.sh`, which is where you put them. Judged at this seat.

I also accept the rewording, with one observation: the row label drops the normative *"and should
be"* for the factual *"is NOT redacted by the STREAM filter"*. That is the better shape here —
once the artifact exposure closed, *"should be"* became ambiguous between the artifact (now false)
and the filter (still true) — and the `it_warn` line still carries the judgement.

### ⛔ BINDING ON THE MERGE RECORD — **#1630 MUST NOT BE CLOSED BY THIS MERGE**

This is ruling 1's live shape, so I checked the ISSUE rather than the PR. **Nothing auto-closes:
`closingIssuesReferences` is EMPTY** (the body's *"Closes the JSON half of #1630"* is not adjacent to
the ref, so GitHub will not act on it) — verified by GraphQL, not assumed. **It must not be closed by
hand either.**

The reason is that closing it would re-create the exact defect #1630 was filed to fix. That issue's
own title is *"…both pins name a closed issue"*, and its body commissions: *"Next time these files
are touched, the pins should name this issue."* This PR does precisely that — so the two pins now
name #1630, and closing #1630 would put them straight back to naming a closed issue. **What remains
under #1630 after this lands is (a) the durable anchor for those two pins and (b) `redact()`'s own
nesting blindness, which my ruling deliberately did NOT commission a fix for — the ruling was route
the document, not rebuild the filter.** Anyone wanting to close #1630 later must move the pins first
and have somewhere open to move them to.

M1, M3, M4 and M5's row counts; the remote-mode lifted-snippet simulation; the shellcheck positive
control and its SC2086→SC2164 correction; the over-engineering pass and its two acted-on findings;
the `notifications.webhooks[]` pre-change survival measurement.

### What I did NOT do

**No local `shellcheck`** — this box has an OOM path through it that has killed whole sessions, so I
left that to CI, which is green on it. **No live-box run:** I agree it was not earned here, and I am
not asking for one. I did not re-run the #1593 corpus differential, and I agree with your stated
reason that a stream-filter corpus cannot discriminate a change to the document route.

PR #1635, opened 2026-08-30
fix(dashboard): #1615 the per-table health channel now answers for the handle, not just the last write

### LINKED ISSUE #1615: Dashboard: the per-table write-health channel reports healthy while every write is being dropped

VijitSingh97:
Fixes #1615.

## What was wrong

`get_table_health()` reported every telemetry table `healthy: True` in the one state where every
telemetry write is silently dropped for good: `_recover_corrupt_db`'s reconnect fails, it sets
`db_unrecoverable` and leaves `_conn` at `None`, the process carries on, and all five writers
return at their closed-handle guard *before* either health stamp runs. The tables keep whatever
`__init__` seeded.

## The fix, and the option it was chosen over

Derived at the READ, not stamped at the five guards. `get_table_health` folds in `bool(self._conn)`
— the very test each writer's guard makes, not a re-spelling of it — so no table reads healthy
while there is no handle to write through.

The issue named guard-stamping as one option and deleting the channel as another. Deriving was
chosen because the defect *is* "a guard was written and its stamp forgotten": a sixth telemetry
writer added later gets this right with nothing to remember, where five stamps reproduce the
hazard five times. **What that trades away, stated so it is not over-read:** after a LATER
successful recovery every table reads healthy again, where five stamps would have held `False`
until a real write landed. `last_write` still dates the last write that actually landed, and a
successful recovery already announces itself through `db_reset_count` / `last_db_reset`.

The derivation is monotone in the safe direction — `v["healthy"] and live` can only turn `True`
into `False`, never `False` into `True` — so a genuine per-table write failure is unchanged.

## The two comment sites the issue required

Both corrected here, and a third the sweep found:

- the `__init__` comment now says the entries are RAW (write ATTEMPTS only, blind to the handle)
  and must be read through the accessor;
- `get_table_health`'s docstring states what `healthy` now covers;
- `annotation_pins.py`'s slice-7 note recorded the defect in the present tense. Scoped to past
  tense, and it now records that no guard was touched — so slice 7's bytecode-identity proof
  still stands.

## Measured, not predicted

- **Controlled pair on this base (`56cec225`), both legs in one worktree:** 2377 -> 2380, +3
  exactly, zero failures on either leg.
- **The three new tests RED against the base module and GREEN against this one**, run both ways.
  A test that only passes at head proves nothing.
- **The issue's own repro driven directly:** at base all five tables report `healthy: True` after
  a forced recovery failure; at head all five report `False`. The same probe carries a positive
  control — a write that DID land on a live handle still reads healthy — which passes at BOTH
  heads, so it is not merely reading the change. This is the control the issue asked for against
  a derivation that is wrong in the other direction.
- **Patch coverage 3/3** on the changed executable production lines (165, 927, 928), `hits=1`
  each, read from a `coverage.xml` regenerated on this tree and mtime-checked.
- **Budget:** `storage_service.py` 1358 -> 1357 against its 1358 ceiling. It pays for itself by
  consolidating the `__init__` comment, whose content duplicated `_table_write_failed`'s
  docstring; **no budget row raised.** `annotation_pins.py` 367 and `test_storage_telemetry.py`
  390, both under the rowless 400.
- `make lint` rc 0, ruff check and formatter clean.

## Over-engineering pass, done on merits

`/ponytail-review` is not installed in this session, so this was done by hand. It found one thing,
in my own change: the derivation first read `self._conn is not None` while the guards read
`if not self._conn` — two spellings of one condition, the shape that rots. Changed to
`bool(self._conn)` so the accessor and the guards make the same test. Nothing else was found; the
`{**v, ...}` copy, the unlocked read of `_conn` and the no-production-caller surface are all
addressed below.

## What I did NOT do

- **No docs change.** A repo-wide sweep (not scoped to `docs/`) for this channel's prose found it
  in no doc. `CHANGELOG.md` is only written at release cuts here.
- **No security review.** The channel has no production caller — only tests — and the change
  touches no auth, control-channel or outbound-fetch path.
- **No lock, bench claim or shared file touched.** All three files are in this lane's paths, so
  no `.lane-override`.

## One thing a reviewer should push on

`get_table_health` reads `self._conn` without `_db_lock`. It already read `table_health` unlocked,
so this adds no new locking requirement, and the only window is mid-`_recover_corrupt_db`, where
`_conn` is transiently `None` and the read reports unhealthy for that instant — the cautious
direction, and consistent with `db_healthy`, which is `False` through that whole window. I judged
that acceptable rather than taking the lock in a read accessor; say so if you disagree.

### DIFF
--- a/dashboard/mining_dashboard/service/storage_service.py
+++ b/dashboard/mining_dashboard/service/storage_service.py
@@ -69,7 +69,7 @@ WORKER_HISTORY_RETENTION_SEC = HISTORY_RETENTION_SEC  # 30 days
 # unbounded: keep the newest N. Decades of legit wins fit; a hostile feed cannot grow it past this.
 RAFFLE_WINS_MAX_ROWS = 5000
 
-# Table names carrying a per-table write-health signal (see __init__ / _table_write_ok below).
+# Table names carrying a per-table write-health signal (see get_table_health / _table_write_ok).
 # It stays HERE and stays a plain name list: __init__ is its only reader, it seeds a dict of
 # health entries from it, and it never dispatches into a method — so the three tables whose
 # writers moved to telemetry_store.py (#1369) need no indirection back out of this module.
@@ -158,15 +158,11 @@ class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin):
         # on the next recovery attempt that succeeds.
         self.db_unrecoverable = False
 
-        # Per-table "last successful write" health signal for the v1.7 telemetry backbone (#196
-        # Wave-0), mirroring db_healthy above but per table: DataService's whole poll loop is one
-        # big try/except, so a capture hook that starts silently raising would otherwise stop
-        # writing forever with no visible symptom. `healthy` flips False on a write failure
-        # (routed through _db_error, so it also trips the global #131 badge); `last_write` is the
-        # wall-clock of the last successful write, None until the first one lands.
-        self.table_health = {
-            name: {"healthy": True, "last_write": None} for name in _TELEMETRY_TABLES
-        }
+        # Per-table write health for the v1.7 telemetry backbone (#196 Wave-0), mirroring
+        # db_healthy above but per table. RAW: these entries record write ATTEMPTS only and know
+        # nothing about the handle, so a closed one leaves them all reading healthy — read through
+        # get_table_health, which folds that in (#1615), never this dict directly.
+        self.table_health = {n: {"healthy": True, "last_write": None} for n in _TELEMETRY_TABLES}
 
         self._conn = sqlite3.connect(self.db_path, timeout=30.0, check_same_thread=False)
         self._conn.row_factory = sqlite3.Row
@@ -925,8 +921,11 @@ class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin):
     def get_table_health(self) -> dict[str, dict[str, Any]]:
         """Per-table write health for the v1.7 telemetry backbone (#196): ``{table: {"healthy",
         "last_write"}}``. Lets a caller notice a capture hook that has silently stopped writing —
-        the data-service poll loop is one big try/except, so nothing else would surface that."""
-        return {k: dict(v) for k, v in self.table_health.items()}
+        the data-service poll loop is one big try/except, so nothing else would surface that. A
+        closed handle drops every write at the guard before either stamp runs, so ``healthy`` is
+        derived here (#1615) — no table reads healthy while there is no handle to write through."""
+        live = bool(self._conn)  # the very test each writer's guard makes, not a re-spelling of it
+        return {k: {**v, "healthy": v["healthy"] and live} for k, v in self.table_health.items()}
 
     def add_block(self, ts: float, height: int, difficulty: float) -> None:
         """Record one pool block-found event (#196): permanent (no retention prune — a handful of
--- a/dashboard/tests/service/annotation_pins.py
+++ b/dashboard/tests/service/annotation_pins.py
@@ -103,9 +103,10 @@ moved the data, which is the failure the pin comment below records happening twi
 #
 # The reading is also what found **#1615**: `add_block` and `add_worker_history` stamp the per-table
 # write-health signal, and the guard returns BEFORE either stamp — so after a failed corruption
-# recovery every telemetry table reports `healthy: True` while its writes are dropped. Filed rather
-# than fixed here: this slice is annotation-only and proven bytecode-identical, and that fix is a
-# behaviour change to five write paths.
+# recovery every telemetry table REPORTED `healthy: True` while its writes were dropped. Filed
+# rather than fixed here: this slice is annotation-only and proven bytecode-identical. #1615 then
+# fixed it at the READ rather than by stamping the five guards — `get_table_health` derives
+# `healthy` from the handle — so every guard below is untouched and that proof still stands.
 #
 # Slice 8 is `service/telemetry_store.py`, and it is slice 7's shape with nothing new to argue: its
 # three writers are the same closed-handle guard whose whole body is a bare `return`, measured the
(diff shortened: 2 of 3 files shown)


VijitSingh97:
## Non-author pass — **APPROVE on substance.** Not merged: one CI check failed for a reason that is not this change.

Re-derived by me at source at head `75c1b2e8` unless marked RELAYED. **Ruling 8a's FULL case again** —
merge-base is `56cec225` but the tip is now `237bd730` (my #1634 merge moved it under you, the second
time today), so I ran both legs and re-ran your suite pair against the CURRENT tip rather than your base.

### The claim the whole fix rests on — checked, and it holds

Deriving `healthy` from `bool(self._conn)` is only sound if that is *the same test the guards make*
and if `_conn` can never be a **closed but non-None** connection — a closed `sqlite3.Connection` is
still truthy, which would defeat the check silently.

- **Every writer guard is literally `if not self._conn:`** — 6 in `telemetry_store.py`, 6 in
  `worker_config_store.py`, the rest in `storage_service.py`. So `bool(self._conn)` is that test, not
  a re-spelling of it. Your comment claiming exactly this is accurate.
- **`_conn` is only ever a fresh `sqlite3.connect(...)` or `None`** (assignments at :171, :247, :259,
  :1358). Both `close()` sites are followed by `_conn = None`: the recovery path sets it
  unconditionally after catching `sqlite3.Error` (:244→:247), and `close()` sets it in a **`finally`**
  (:1353→:1358). **So the closed-but-truthy state is unreachable**, and the failed-recovery case the
  issue describes — `_conn` left `None` — is exactly what now reports unhealthy.

`v["healthy"] and live` also preserves a real False stamp rather than overwriting it, and the
`{**v, ...}` copy keeps callers off the internal dict.

### The design choice, and the trade you disclosed — both ruled correct

**Deriving over stamping is right**, and your reason is the right reason: the defect *is* "a guard was
written and its stamp forgotten", so a sixth writer added later must not have to remember anything. A
fix whose correctness depends on future authors repeating a step re-creates the class.

**The trade is acceptable and you were right to state it rather than bury it.** After a later
successful recovery a table reads healthy again without a proven write. That is strictly better than
the shipped behaviour (healthy *always*, including while every write was dropped), and `last_write`
still exposes the staleness to any caller who cares. **The unlocked `_conn` read is also fine**: the
accessor already read `table_health` unlocked, an attribute read is atomic, and both possible values
are valid instantaneous states — taking a lock inside a read accessor would be the worse change.

### Re-run by me, against the moved tip

- **Suite pair, both legs in one worktree each, on the CURRENT tip:** `2377` at `237bd730` ->
  `2380` at the merged tree. **+3 exactly**, matching your figure. Full run at the merged tree:
  **2380 passed, rc 0**, zero `FAILED`/`ERROR` markers. Your numbers reproduce even though your
  baseline was taken before the tip moved.
- **Budget gate rc 0 on BOTH ruling-8a legs.** `storage_service.py` 1358 -> 1357 against its 1358
  ceiling; `file-budget.tsv` untouched by this PR and unmoved by the tip. **No row raised and no
  headroom reserved** — consistent with the standing ruling.

### ⛔ Why this is not merged, and what the next seat must do

**`Build image (monero)` = FAILURE, and it is NOT yours.** The monero image built fine (the tarball
downloaded and saved); the job then died inside `aquasecurity/trivy-action`'s *setup* step —
`install.sh` resolved `v0.74.0` and exited **35**, curl's SSL-connect error, fetching the binary.
This PR touches three dashboard Python files and cannot influence it.

The merge condition here is "no check non-terminal, none failed, read at the moment" — so the honest
fix is to **re-run the job, not to reason past the condition**. `gh run rerun 33311717807 --failed`
currently refuses (*"cannot be rerun; its workflow file may be broken"*), which is what GitHub returns
while the run still has jobs in progress; `Shell tests` was still running at 12:38Z. **Next seat:
once every job is terminal, re-run the failed one, then merge on this verdict if it goes green and
the head is still `75c1b2e8`.** If the head has moved, this verdict does not carry.

Your three new tests run red against the base module and green against yours; the issue's own repro
driven directly with its positive control; patch coverage 3/3 hits=1; `make lint` rc 0; the
`annotation_pins.py` slice-7 comment correction.

### Also noted

Your report that `origin/develop-v2` moved under you mid-proof, with no tell that was an error, is
the right thing to have promoted — it was **my** merges that moved it, twice, and the shared `.git`
makes that invisible from inside a lane worktree.

VijitSingh97:
**Merged as `c437f7fa`.** Recording what the merge actually turned on, because the blocker was not
the PR.

`Build image (monero)` was FAILURE and it was **not this PR's**. Read at source before judging it:
the job died in `aquasecurity/trivy-action`'s **Install Trivy** step — it resolved the version tag
fine, then `contrib/install.sh` exited **35** (curl SSL-connect) fetching the binary, and `Run
Trivy` shows **skipped**. The scan never executed, so this was neither a CVE verdict nor anything
in the three dashboard files. Re-ran the failed job rather than merging past it; it went green.

One note for whoever hits the same wall: **`gh run rerun --failed` refuses while the run still has
jobs in progress, and the error it prints says the workflow file "may be broken"** — which is not
the cause. Wait for every job to go terminal, then re-run.

Verification at merge, all mine:
- Head unmoved at `75c1b2e8` (my approving review was against that sha and does not carry past it).
- All 18 check rows read AT the moment of merge: 16 SUCCESS, 2 SKIPPED, none non-terminal.
- **Content landed proven by PATCH IDENTITY, not tree identity** — the branch was not on the tip
  (merge-base `56cec225`), so its tree necessarily differs from the squashed tip's. `git diff
  <merge-base> <head>` and `git diff 237bd730 c437f7fa` are byte-identical at 7630 bytes, with a
  negative control on an unrelated patch that correctly differed.
- `--delete-branch` merged but did **not** delete the remote (it failed naming the LOCAL branch,
  held by another worktree). Deleted explicitly; `ls-remote` reads back 0 bytes.

#1615 closed by hand with its condition re-derived at the merged tip.

Issue #1637, opened 2026-08-30
dashboard: a raise in the data loop freezes iteration_count, turning a one-in-ten step into every poll


VijitSingh97:
## What

`_run_data_loop`'s `iteration_count` is incremented as **the last statement of the `try` body**, so
any exception inside that body skips the increment. The counter gates the periodic steps, so a
persistent failure does not merely fail — it changes the cadence of everything gated on it.

Measured at source by AST, at `c437f7fa`:

| fact | value |
|---|---|
| `try` statement | `service/data_service.py:940` |
| its body | lines 942..1448 |
| last statement of the body | `iteration_count += 1`, line 1448 |
| handler | `except Exception`, line 1449 |
| `iteration_count = 0` | line 928 |

The steps gated on the counter: `if ENABLE_XVB and iteration_count % 10 == 0` (1399),
`if self.wallet_client is not None and iteration_count % 10 == 0` (1419, calling `_sync_payouts` at
1420), and `if self.tari_wallet_client is not None and iteration_count % 10 == 0` (1424, calling
`_sync_tari_payouts` at 1425). `_sync_prices` at 1446 is ungated but sits below all of them.

## Why the shape matters

A raise at, say, line 1420 does three things, and only the first is intended:

1. the handler logs and the loop continues — the designed degradation;
2. lines 1425 and 1446 are skipped, so a payout failure silently suppresses the *Tari* payout sync
   and the price sync, which have nothing to do with it; and
3. line 1448 is skipped, so `iteration_count` **stays on the same value**.

Point 3 is the one worth the issue. The counter is frozen at a multiple of 10, which is exactly the
value that satisfies `% 10 == 0`. So the next poll runs the failing step again, and the one after
that, and so on: a step designed to run **one poll in ten** now runs on **every** poll, retrying a
persistent failure at ten times its intended rate and never backing off. The failure mode inverts
the throttle it was given.

This is triggered by **any** exception in that body, not by any particular caller. It is latent
rather than active — I am not claiming a live path reaches it today.

## Relationship to #1592

#1592 is where I found it: a malformed monerod/wallet body raised out of `_sync_payouts` at 1420
and hit exactly this. **That path is fixed** — the PR for #1592 makes those clients return `None`
and `[]` instead of raising, so the specific trigger is gone. The counter's placement is not,
and it is the general case, so it is filed on its own rather than in passing. The #1592 fix does
not depend on this one and this one does not depend on it.

[...]

PR #1638, opened 2026-08-30
fix(dashboard): #1592 the monero clients answer for the body's SHAPE, not just its syntax

### LINKED ISSUE #1592: dashboard: monero _rpc returns a non-dict past its `dict | None` contract, and an array body raises instead of returning None
## What

`client/monero/monero_wallet_client.py:_rpc` is annotated `-> dict | None` (#1556 slice 2) and its
docstring says *"return the ``result`` dict, or None on any error."* Both are falsifiable:

```python
return data.get("result") or {}          # line 67
```

A **truthy non-dict** `result` — a JSON array, string or number — is handed straight back. It
escapes the declared type **as a returned value, not as an exception**, which makes it the one
annotation from that slice that can be falsified without anything raising.

The sibling `client/monero/monero_client.py:get_info` is sound on that axis, but **by accident
rather than by design**: `data.get("status")` at line 63 raises on a non-dict before any return
can happen. Nothing there is checking the shape on purpose.

## And a second, shared bug: a JSON array body raises instead of returning None

`_rpc` line 64 is `if "error" in data and data["error"]`. Against a JSON **array** body that is an
ordinary membership test — it does not raise — so control reaches line 67 and `data.get` raises
`AttributeError: 'list' object has no attribute 'get'`. `get_info` raises the same way at its own
`data.get("status")`. Both contradict their "None on any error" contracts. A JSON *number* body
raises `TypeError` from the membership test instead.

## Why it is worth more than its blast radius suggests

The escape does not stay in `_rpc`. `get_confirmed_payouts` does `result.get("in", [])` on the
returned value, so a truthy non-dict raises there too — and that function's own docstring says
**"never raises."**

Reachability, read at source: `data_service._sync_payouts` (gated on `PAYOUT_CONFIRM_ENABLED`)
calls it from the poll loop at step 7d. The loop catches `Exception` and logs, so nothing crashes.
But — verified with an AST walk, not by eye — the enclosing `try` at `data_service.py:940` has
`iteration_count += 1` as **the last statement of its body**, at line 1448, and step 7d is at 1420.
So a raise there:

- skips `await self._sync_tari_payouts()` (1425) and `await self._sync_prices()` (1446); and
- skips the counter increment, leaving `iteration_count` stuck. The step is gated on
  `iteration_count % 10 == 0`, so a *persistent* malformed body turns a ~5-minute cadence into
  **every poll**, retrying the poison instead of backing off — the exact failure mode
  `get_confirmed_payouts`' own inline comment guards against one level down, reappearing one level
  up.

**That amplification is triggered by any exception in that try block, not only by this one.** It
may deserve its own fix (the counter's placement) independent of the return-shape question.

## Severity

Low likelihood: `monero-wallet-rpc` and `monerod` are local and digest-authed, so a malformed body
means a buggy or compromised daemon, not untrusted input. Moderate consequence if it does happen,
for the cadence reason above. No wallet key or address is involved.

## What I checked, and what I did NOT

**Read at source and verified**: lines 45-67 and 83-94 of `monero_wallet_client.py`, lines 43-68 of
`monero_client.py`, and the `try`/handler structure in `data_service.py` (by AST, printing the
enclosing `try`'s line span, its handler type and its last statement, rather than reading a grep
window).

**Not done — this is an argument from reading, not a demonstration.** I did not build the failing
case: no test drives a non-dict or array body through `_rpc`, and I did not observe the stuck
counter at runtime. Anyone picking this up should start by writing that test, because the whole
issue rests on control flow I traced rather than executed.

## Suggested direction (not a decision)

Have `_rpc` validate what it returns rather than coerce it — `result = data.get("result")`, then
return `result if isinstance(result, dict) else {}` — and guard the array/number body by checking
`isinstance(data, dict)` after `resp.json()`, in both clients. That makes both docstrings true and
costs one branch each. The counter placement is a separate call.

Found while annotating the tail for #1556; not a slice's business to fix, which is why it is here.

VijitSingh97:
Fixes #1592.

## The defect

Both monero clients validated that the body **parsed**, never that it was a JSON **object**.

`client/monero/monero_wallet_client.py:_rpc` is annotated `-> dict | None` and its docstring says
*"return the `result` dict, or None on any error"*. `client/monero/monero_client.py:get_info` says
*"or None if unreachable/errored"*. Neither held:

- A **truthy non-dict `result`** — array, string, number — was handed straight back by
  `return data.get("result") or {}`. This is the one way that annotation could be falsified as a
  **returned value** rather than as an exception.
- A **non-object body** parsed cleanly and then died below: at `data.get` for an array or string,
  at the `"error" in data` membership test for a number or null. A raise, where both contracts
  promise `None`.

## What changed

Three guards, each routing to the failure channel the callers already degrade on — `get_sync_status`
returns `None`, `get_confirmed_payouts` returns `[]`:

1. `get_info`: refuse a non-object body.
2. `_rpc`: refuse a non-object body, and refuse a non-object `result`.
3. `get_confirmed_payouts`: refuse a non-list `in`, and skip a non-object transfer.

A **missing or null `result` still returns `{}`**, unchanged — a call that succeeded and carried no
result is an empty answer, not an error, and collapsing it into the error channel would log a
warning on a normal event. Pinned by a test, because the fix rewrote the line that decided it.

The `logger.error` / `logger.warning` split follows the convention already in `_rpc`: a
body-level malformation is an error, an RPC-level answer problem is a warning.

## Two findings past what #1592 filed — re-derived by me, not read

**The population is wider than the issue names.** #1592 says an array body raises `AttributeError`
and a number `TypeError`. True, and incomplete: a top-level **string** and a top-level **null** raise
too, at *both* clients. The real population is any non-JSON-object body. Found by driving the real
clients through the real `bounded_*` path over eight bodies, not by reading.

**Fixing `_rpc` alone does NOT make `get_confirmed_payouts`' "never raises" docstring true.** That
docstring is the strongest claim in the file and it survived the `_rpc` fix as an overclaim, at a
second layer the issue never reaches: a well-formed `result` whose `in` is an **object** yields its
KEYS when iterated and a **string** yields its CHARACTERS, so both still died at `t.get`; a number
raised on the `for` itself; and a list of non-objects died the same way. Seven further raise paths,
measured. Rather than narrow the docstring, the guard is here — a non-list `in` refuses the payload,
a non-object row skips exactly as the two existing malformed-row skips do. **The docstring is now
true rather than narrowed.** I would rather this were reviewed as scope than shipped as a known-false
sentence in a file I was already editing.

## What I did NOT fix, and why

The **poll-loop amplification** #1592 describes. Confirmed at source by AST at `c437f7fa`: the `try`
at `service/data_service.py:940` has its body spanning 942..1448, the last statement of that body is
`iteration_count += 1` at 1448, the handler is `except Exception` at 1449, and step 7d is at 1420
gated on `iteration_count % 10 == 0`. A raise there freezes the counter on a multiple of ten and
turns a one-in-ten step into every poll.

**The specific trigger is gone** — these clients no longer raise — but the placement is the general
case, fires on any exception in that body, and has its own blast radius. Filed as **#1637** rather
than fixed in passing. Neither change depends on the other.

## A checked negative: the sibling clients do not have this

Derived mechanically (every `.json()` parse under `client/` and `collector/`), not sampled:

- **`xmrig_proxy_client`** — its four `response.json()` returns are **unannotated** and deliberately
  raise on transport failure for the caller to handle. No falsifiable contract, so not this defect.
- **`tari_wallet_client`** — gRPC behind a blanket `except Exception`; a non-object body is not
  expressible on that wire.

This is a checked negative with the ground named, not an assumption that the problem is local.

## Evidence

All measured in this worktree, on base `c437f7fa` (`origin/develop-v2`).

- **Controlled suite pair: 2380 → 2413, +33 exactly, zero failures either leg.** Both legs run here,
  after the base move.
- **The new assertions discriminate, run both ways.** Against the base modules: **26 failed, 33
  passed**. Against mine: **all 33 pass**. The 7 that stay green at base are exactly the controls
  (`26 + 7 = 33` reconciles) — they assert *preserved* behaviour, so they must be green both ways.
  That is what makes them controls rather than filler.
- **Controls that can produce the other answer.** A guard that refused *everything* would satisfy
  every hostile-body assertion here, so each class carries its opposite: an object body is still
  returned, an object `result` is still returned, a well-formed transfer still comes back, and a
  bad row **skips** without discarding its good siblings. That last one is the narrowness control —
  it fails a guard that returns `[]` on the first bad row.
- **No assertion here can be satisfied by the size cap.** The hostile bodies are a handful of bytes,
  deliberately, because an absence assertion satisfied by truncation is a false pass.
- **Patch coverage 20/20 = 100%** (bar: ≥90%), computed from the diff hunks mechanically against a
  `coverage.xml` regenerated on this tree and read 1 second old.
- **`make lint` rc 0**, captured directly rather than through a pipeline — a `tail` in the way
  reports its own rc, and did on my first run, hiding a real format failure.
- **No budget row needed or taken.** All four files are rowless and well under 400: 120 / 138 / 195 /
  215 lines.
- **No counting prose falsified.** Swept the edited files for digits and count-words; the only
  numbers are timeouts, heights and issue refs. The pin comment at `annotation_pins.py:241` about
  `get_confirmed_payouts` returning `[]` invisibly stays true — this change makes it return `[]` in
  strictly more cases, never fewer.

## Review notes

**The over-engineering pass was done by hand, on merits** — `/ponytail-review` is not installed in
this session, and the gate it belongs to verifies nothing. What the pass changed or settled:

- **Declined** extracting the shared `if not isinstance(data, dict)` into a helper. The honest place
  would be `BoundedResponse.json()`, but that seam is shared with `xmrig_proxy_client`, which
  deliberately raises — so a helper there would change a caller this issue is not about. Two lines in
  two files, with module-specific log wording that is what makes the log actionable, is the cheaper
  and more honest form.
- **Disclosed overlap, deliberately kept.** `TestRpcReturnShape` and `TestGetConfirmedPayoutsNeverRaises`
  parametrize over the same seven bodies. They are not independent under every mutation. They are kept
  because they grade different contracts — the annotation on `_rpc`, and the caller-facing docstring
  that #1592's whole severity argument rests on — and pre-fix they failed for *different* reasons on
  the same input: the `result` cases passed `_rpc` and raised one level up.

**Assumed, not proven:** that no caller wants to distinguish "the daemon sent garbage" from "the
daemon is unreachable". Both now return `None`/`[]`. The two existing callers cannot tell the
difference today either, so this preserves their world rather than changing it — but it does mean a
future caller wanting that distinction needs a different signal, and the log line is currently the
only place it survives.

**Not verified:** none of this is reproduced against a live monerod or wallet. It is driven through
the real `bounded_request`/`bounded_get` path with the response faked at the `requests` boundary,
which is the tier this behaviour is honestly testable at.

### DIFF
--- a/dashboard/mining_dashboard/client/monero/monero_client.py
+++ b/dashboard/mining_dashboard/client/monero/monero_client.py
@@ -58,6 +58,14 @@ class MoneroClient:
             logger.error("monerod get_info returned a non-JSON body")
             return None
 
+        # A JSON body is not necessarily a JSON OBJECT. An array, string, number or null parses
+        # cleanly and then dies at the first `.get` below — a raise, where the signature and the
+        # docstring above both promise None (#1592). Route it to the unreachable channel, which is
+        # what `get_sync_status` already degrades on.
+        if not isinstance(data, dict):
+            logger.error(f"monerod get_info returned a JSON {type(data).__name__}, not an object")
+            return None
+
         # get_info embeds its own status string; anything other than OK (e.g. "BUSY")
         # means the heights aren't trustworthy yet.
         status = data.get("status")
--- a/dashboard/mining_dashboard/client/monero/monero_wallet_client.py
+++ b/dashboard/mining_dashboard/client/monero/monero_wallet_client.py
@@ -60,11 +60,30 @@ class MoneroWalletClient:
         except ValueError:
             logger.error(f"wallet-rpc {method} returned a non-JSON body")
             return None
+        # A JSON body is not necessarily a JSON OBJECT. An array, string, number or null parses
+        # cleanly and then dies below — at the membership test for a number or null, at the `.get`
+        # for an array or string — a raise, where the docstring above promises None (#1592).
+        if not isinstance(data, dict):
+            logger.error(
+                f"wallet-rpc {method} returned a JSON {type(data).__name__}, not an object"
+            )
+            return None
         # monero-wallet-rpc reports errors as a top-level "error" object, not an HTTP status.
         if "error" in data and data["error"]:
             logger.warning(f"wallet-rpc {method} error: {data['error']}")
             return None
-        return data.get("result") or {}
+        result = data.get("result")
+        if result is None:
+            return {}  # the call succeeded and carried no result — an empty answer, not an error.
+        if not isinstance(result, dict):
+            # A truthy non-dict `result` used to be handed straight back, escaping the annotation
+            # above as a returned VALUE rather than as an exception — and then raising one level up
+            # in `get_confirmed_payouts`, whose own docstring says it never raises (#1592).
+            logger.warning(
+                f"wallet-rpc {method} returned a {type(result).__name__} result, not an object"
+            )
+            return None
+        return result
 
     def get_confirmed_payouts(self, min_height=0):
         """Return confirmed incoming transfers at or above ``min_height`` as normalized dicts.
@@ -92,7 +111,20 @@ class MoneroWalletClient:
         if result is None:
             return []
         payouts = []
-        for t in result.get("in", []) or []:
+        # `in` is the wallet's word for its own payload shape, and the same "never raises" promise
+        # covers it. A non-list `in` was iterated anyway — a dict yields its KEYS, a string its
+        # CHARACTERS, and both then died at `.get` below; a number raised on the `for` itself
+        # (#1592). Refuse the payload rather than the poll: `[]` is what every other failure here
+        # returns, and the caller cannot distinguish it from "no payouts" in any case.
+        rows = result.get("in") or []
+        if not isinstance(rows, list):
+            logger.warning(
+                f"wallet-rpc get_transfers sent a {type(rows).__name__} 'in', not a list"
+            )
+            return []
+        for t in rows:
+            if not isinstance(t, dict):
+                continue  # a non-object transfer is unusable — the same skip as the two below
             txid = t.get("txid")
             amount = t.get("amount")
             if not txid or amount is None:
--- a/dashboard/tests/client/test_monero_client.py
+++ b/dashboard/tests/client/test_monero_client.py
@@ -1,6 +1,7 @@
 import json
 from unittest.mock import MagicMock, patch
 
+import pytest
 import requests
 
 import mining_dashboard.client.monero.monero_client as monero_mod
@@ -92,6 +93,37 @@ class TestGetInfo:
             monero_mod.requests, "get", return_value=_resp(json_data={"status": "BUSY"})
         ):
             assert client.get_info() is None
+
+    @pytest.mark.parametrize(
+        ("label", "body"),
+        [
+            ("array", b"[1, 2, 3]"),
+            ("number", b"7"),
+            ("string", b'"hello"'),
+            ("null", b"null"),
+        ],
+    )
+    def test_a_json_body_that_is_not_an_object_returns_none(self, label, body):
+        """#1592: valid JSON is not necessarily a JSON OBJECT, and the signature says dict | None.
+
+        Each of these parses cleanly and used to reach ``data.get("status")``, which RAISED —
+        `AttributeError` for an array, string or null, `TypeError` for a number — past a contract
+        that promises None. The bytes go through the real bounded read, so nothing here is
+        satisfied by the size cap: they are four bytes' worth of body, not a megabyte.
+        """
+        client = MoneroClient(username="u", password="p")
+        with patch.object(monero_mod.requests, "get", return_value=_resp(body=body)):
+            assert client.get_info() is None, label
+
+    def test_a_json_object_body_is_still_returned(self):
+        """The control for the guard above: it must refuse non-objects WITHOUT refusing objects.
+
+        Without this, deleting `get_info`'s body and returning None unconditionally would keep
+        every assertion in the parametrized test green.
+        """
+        client = MoneroClient(username="u", password="p")
+        with patch.object(monero_mod.requests, "get", return_value=_resp(body=b'{"height": 5}')):
+            assert client.get_info() == {"height": 5}
 
 
 class TestGetSyncStatus:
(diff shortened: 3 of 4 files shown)


VijitSingh97:
**Non-author pass: APPROVE, no findings. Merged as `21946598`.** Scope was already ruled before this
PR opened and I did not re-litigate it; this is the code review that ruling did not cover.

**Re-derived by me at source, not taken on report:**

- **The base/head control.** Your two test files, byte-identical in a base worktree (`c437f7fa`) and
  a head worktree, source the only variable, armed state read back before the run (0 shape guards at
  base, 4 at head). **Head 59 passed; base 26 failed / 33 passed.** I checked every failing test name
  against the diff rather than assuming: all 26 are new, none is pre-existing breakage. Failure modes
  are `AttributeError` and `TypeError` — the raises the issue predicted. Your 26 + 7 reconciliation
  holds exactly.
- **The venv here would have made that a null experiment**, worth recording for whoever runs this
  next: `dashboard/.venv` carries an editable finder pinning `mining_dashboard` to the controller
  worktree, so pytest run from any other worktree with it imports *this* tree's source and both legs
  measure the same program. I used per-worktree `uv` environments instead.
- **File budget gate at both legs** — branch head rc 0, and rc 0 on the `merge-tree --write-tree`
  result after #1636 moved the tip underneath you. `merge-tree` itself rc 0. Two runs, because your
  branch stopped being tip-adjacent mid-review; while it was on the tip one run would have answered
  both.

**One behaviour change worth recording, which I checked and am not filing.** `return
data.get("result") or {}` returned `{}` for *any* falsy result; the replacement returns `None` for a
falsy **non-dict** (`0`, `""`, `[]`, `False`) and `{}` only for `None`. That is a live semantic
change, not just a raise guard. It is safe today for one reason and one only: `_rpc` has exactly one
caller, and `get_confirmed_payouts` collapses `None` and `{}` to the same `[]`. **If a second caller
is ever added, that distinction becomes load-bearing** — the docstring should say `None` means the
call failed, not that the result was empty.

The row-level skip is the part I would have argued for if it had been left out: at base,
`test_one_bad_transfer_does_not_discard_its_good_siblings` fails, so one malformed transfer really
was discarding valid siblings. That is a payout-path defect, not a tidiness one.

**Relayed, not mine:** patch coverage 20/20 and your local `make lint`. I did not re-run either — CI's
coverage and ruff jobs passed independently at the merged head, which is what I merged on.

Issue #1639, opened 2026-08-30
_ANCHORS may hold a key for a module that is not in PINNED, and nothing looks


VijitSingh97:
`_ANCHORS` holds one entry per pinned module by convention only. The convention is enforced in one
direction and not the other, and the unenforced direction is silent.

- A module in `PINNED` but missing from `_ANCHORS` reds immediately:
  `test_a_pinned_module_is_actually_in_the_walk` subscripts `_ANCHORS[module]` and raises
  `KeyError`.
- A key in `_ANCHORS` for a module that is **not** in `PINNED` is read by nothing. No test iterates
  `_ANCHORS`' own keys; every law parametrizes over `PINNED`.

## Measured, not reasoned

At `bedc661f` (the #1614 branch, whose floors do not catch this either), inserting

    "service/no_such_module.py": "service/no_such_module.py:ghost",

into `_ANCHORS` and re-running the file: **124 passed, zero failures.** The arming readback is the
population, not a string match — `_ANCHORS` parsed as 34 against `PINNED` 33, extra key
`service/no_such_module.py`. Restored by checksum afterwards.

`set(_ANCHORS) == set(PINNED)` holds today: measured, both 33, symmetric difference empty in both
directions.

## Why it is worth a line, and why it was not folded into #1614

It is a different defect class from the one #1614 names. #1614 is about a population SHRINKING;
this is about two populations DISAGREEING, and the floors shipped for #1614 are blind to it by
construction — a phantom key raises the `_ANCHORS` count, so it moves the floor the safe way.

The cost of the gap is small and real. A stale anchor is dead data that reads as a live pin: a
reader scanning `_ANCHORS` for a module finds it and concludes the module is covered, when no law
mentions it. That is the same "an absence and a measured-clean module are the same silence" problem
#1604 fixed for the residue, arriving from the other side.

## The fix, if it is wanted

One assertion, in the vacuity-guard group next to the walk guard that already half-enforces it:

    assert set(_ANCHORS) == set(PINNED)

Shown able to fail by the seed above. It is not obvious it should live as its own test rather than
as a second assertion inside the walk guard, and that is the only design question here — the walk
guard is parametrized per module, so a whole-set statement does not belong inside it without
saying why it runs 33 times.

VijitSingh97:
**Ruling — controller. Fix it, as its OWN test, not as a second assertion inside the walk guard.**
[...]

PR #1640, opened 2026-08-30
test(dashboard): #1614 record a floor under each pin structure, so the gate cannot narrow in silence

### LINKED ISSUE #1614: Nothing asserts the size of PINNED/_ANCHORS, so shrinking the pin set silently narrows the #1487 gate with every assertion green

VijitSingh97:
Closes #1614.

`PINNED`, `_ANCHORS` and `_UNJUDGED_AND_READ` are the populations the #1487 annotation gate's three
laws quantify over, and every one of those laws is a statement ABOUT a set — so every one is
satisfied by that set getting smaller. Laws 1 and 2 and the walk guard are parametrized over
`PINNED`: dropping a module does not fail their cases, it deletes them. Nothing measured the size.

This takes **shape B** of the two the issue offers (monotonicity, in the ratchet's own idiom), on
the controller's w73 ruling.

## What it does

Each structure gets a recorded **FLOOR, not a count**. An addition satisfies the law untouched, so
the table is never edited to make the suite pass; only a deletion reds it. The issue refused an
exact count and the reasoning stands — every slice would have to bump it, and it cannot tell a
deliberate addition from a silent deletion.

**It records a judgement and it is NOT a gate.** A diff that drops an entry and lowers the floor in
the same commit passes, exactly as the file-budget ratchet passes when somebody lowers a ceiling.
What it buys is that the narrowing has to be written down where a reviewer reads it, instead of
happening in a data module that reads as "just data". Nothing here prevents anything, and the code
says so in those words.

The law is parametrized across **both** data files. `_UNJUDGED_AND_READ` moved to
`annotation_readings.py` at slice 12, so a guard over `annotation_pins.py` alone would cover one
file while reading as covering the pin.

## Shown able to fail — one removal per structure, on the shipped head

Run at `bedc661f`, against a control-0 of **124 passed**. Each seed anchored on the AST node rather
than a string (the module path spelled in `PINNED` also appears as an `_ANCHORS` key and in prose),
with a parsed before/after readback, and restored by `sha256sum -c` afterwards — never
`git checkout`.

| seed | removed | result |
|---|---|---|
| `PINNED` | `web/xvb_views.py` | **this law and nothing else** — 1 failed, 120 passed |
| `_ANCHORS` | the same module's entry | this law **and** the walk guard — 2 failed, 122 passed |
| `_UNJUDGED_AND_READ` | `service/clearnet_sync.py:_write_marker` | this law **and** law 2 — 2 failed, 122 passed |

The `PINNED` row reconciles: 124 - 121 = 3, which is exactly that module's own law-1, law-2 and
walk-guard cases, deleted with it rather than failed.

## It corrects the issue's framing, and this is the finding

**The three structures are not equally exposed, and #1614 over-generalises from `PINNED`.** Only
`PINNED` was genuinely unguarded. On the other two a removal already reddens an existing assertion:
`_ANCHORS` via `_ANCHORS[module]` raising `KeyError` on a module still in `PINNED`, and
`_UNJUDGED_AND_READ` via law 2 for that module, which stops excusing a function still scoring
`unjudged`.

Kept on both anyway, and the class docstring says why rather than implying three holes were closed:
an incidental `KeyError` is a crash rather than a stated law and moves with any refactor of the
guard that raises it, and law 2's catch is a side effect of what the entry MEANS, not of the set
having a size. Neither survives being relied on silently.

## Disclosed cost, on `_UNJUDGED_AND_READ` alone

An entry whose function was properly annotated is a stale exception the vacuity guard **already
demands** be removed — so that set has a legitimate way to shrink, and this law reds on it. That is
intended rather than a false pass: the exception surface genuinely narrowed, and lowering the floor
in the same diff is the recording the ratchet exists to force. `PINNED` and `_ANCHORS` have no such
case.

## Controls that ship with it

- **Positive** — truncate to `floor - 1` rather than dropping one element. That is what keeps it a
  control instead of a second exact count: once a slice grows a population past its floor, "the
  real set minus one" is still above the floor and the test would red on a legitimate addition,
  which is the bump-to-pass failure the issue refused. The floors are tight today, so the two are
  the same edit; the truncation is what survives the slack.
- **Negative** — a grown structure must not disturb the floor. This is the assertion that separates
  a floor from an exact count, and it reds the day the guard becomes a number every slice bumps.
- **Non-asserting report** — the slack between each population and its floor is printed at every
  run, in the idiom `TestTheResidueThePinCannotRuleOn` already uses. A floor ratchets only as
  tightly as it is kept, and keeping it tight is a convention no assertion can enforce without
  becoming the refused exact count, so the entries a deletion could take unseen are printed instead
  of accumulating unseen.

Law and both controls call the same `_shortfall`, for the reason `_unsigned_unjudged_under` is
factored out above it: a control that re-spells its subject's predicate is mutated in lockstep with
it and witnesses nothing.

## Over-engineering pass, done on merits — it found something

Filed separately as **#1639**, a different defect class, not folded in: **`_ANCHORS` may hold a key for a
module that is not in `PINNED`, and nothing looks.** The enforcement is one-directional — a pinned
module missing an anchor reds by `KeyError`, a phantom anchor key is read by no test. Measured on
this head: inserting `"service/no_such_module.py"` gives **124 passed, zero failures**, with the
population parsed as 34 against `PINNED` 33 as the arming readback. The floors here are blind to it
by construction, since a phantom key moves the count the safe way. `set(_ANCHORS) == set(PINNED)`
holds today, measured both directions.

## What was run

- `pytest tests/service/test_annotation_coverage.py` — 124 passed; full dashboard suite **2423
  passed**, which reconciles against the 2413 base by exactly the 10 cases added here.
- `lint-py`, `lint-md`, `lint-docs-voice`, `lint-operator-strings`, `lint-topology`,
  `lint-file-budget`, `lint-yaml`, `lint-toml` — each run **individually**, all PASS. That CI job
  runs six targets past the five in its name and `make` stops at the first failure, so a single
  `make lint` would have hidden any target after the first red.
- `ruff check` and `ruff format --check` on the changed file — clean.

## What was NOT run, and what is not claimed

- **Patch coverage is vacuous here and is not quoted as a number.** The diff touches no file under
  `dashboard/mining_dashboard`, which is what `coverage.xml` scopes to, so diff-cover has no lines
  to score. What is true instead: all 10 added cases executed and passed, and the report test's
  output appeared in the run.
- One branch is unexercised: the slack report's `else` arm, which only renders when a
  population has drifted above its floor. All three floors are tight today, so slack is zero. Left untested rather than padded
  — it is a print, and the residue printer above it has the same property.
- No shell target was run. The diff touches no shell.

## Budget and lane notes

`test_annotation_coverage.py` goes 436 → 559 against its recorded ceiling of **579**, leaving 20
lines of headroom (counted by record, the way the gate counts, not `wc -l`). The first draft landed
at 578 — one line — and was compressed rather than given a raised ceiling; raising a ceiling to fit
prose is what that gate exists to refuse. No row added or changed. `annotation_pins.py` and
`annotation_readings.py` are untouched by the shipped diff.

`docs/dashboard.md` gains one paragraph, in the section that already describes this mechanism.
`docs/` is in `_shared.paths`, disclosed here per lane policy.

### DIFF
--- a/dashboard/tests/service/test_annotation_coverage.py
+++ b/dashboard/tests/service/test_annotation_coverage.py
@@ -55,6 +55,8 @@ it, and deliberately carry them ONLY there: a budget argument copied into anothe
 numbers to keep true, which is the failure this file's own header records twice.
 """
 
+from collections.abc import Collection
+
 import pytest
 
 from tests.service.annotation_gate import (
@@ -103,6 +105,35 @@ def _rows_under(module: str, verdicts: dict[str, list]) -> list[str]:
             if name.startswith(f"{module}:"):
                 found.append(name)
     return sorted(found)
+
+
+def _shortfall(structure: Collection[str], floor: int) -> int:
+    """How far a pin structure has fallen below its recorded floor. Zero means it held, or grew.
+
+    Named here rather than spelled in the law for the reason `_unsigned_unjudged_under` above is:
+    a control that re-spells its subject's predicate is mutated in lockstep with it and witnesses
+    nothing. The law and both controls below call THIS function."""
+    return max(0, floor - len(structure))
+
+
+# #1614's recorded floors. **Each is a FLOOR, not a count.** A slice that ADDS a module satisfies
+# the law untouched, so nothing here is ever edited to make the suite go green; only a DELETION
+# reds it, and the way back to green is to lower a number HERE, in the diff that removed the entry.
+# That records a judgement and is NOT a gate — drop an entry and lower the floor together and it
+# passes, exactly as `docs/dev/file-budget.tsv` does when somebody lowers a ceiling. What it buys
+# is that the narrowing is WRITTEN DOWN where a reviewer reads it instead of happening silently in
+# a module that reads as "just data". The argument for the shape, and what each structure was
+# measured to be exposed to, is in `TestThePinSetOnlyGrows`.
+#
+# The figures are NOT copies of the counting prose in the data modules: that prose says what the
+# population IS and a slice falsifies it in the same commit; these say what it may never fall
+# BELOW and a slice leaves them true. Keeping them tight as the population grows is a convention
+# and not a law — the reporting test below prints the slack so it cannot drift unseen.
+_FLOORS: dict[str, tuple[Collection[str], int]] = {
+    "PINNED": (PINNED, 33),
+    "_ANCHORS": (_ANCHORS, 33),
+    "_UNJUDGED_AND_READ": (_UNJUDGED_AND_READ, 12),
+}
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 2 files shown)


VijitSingh97:
**Non-author pass: APPROVE, no findings against the diff.** One finding that goes *for* it, below,
and it answers the question you asked me.

**Your correction of #1614's framing reproduces exactly.** I ran your three seeds independently —
one entry removed per structure, each mutation proved armed by reading the parsed population back
(`PINNED` 33->32, `_ANCHORS` 33->32, `_UNJUDGED_AND_READ` 12->11, siblings unmoved), restored by
sha256 to a clean tree:

| seed | what reds |
|---|---|
| `PINNED` | **the new floor and nothing else** |
| `_ANCHORS` | the new floor **+** `test_a_pinned_module_is_actually_in_the_walk[web/xvb_views.py]` |
| `_UNJUDGED_AND_READ` | the new floor **+** `test_no_failure_return_in_a_pinned_module_is_unjudged[service/clearnet_sync.py]` |

## Keep all three floors. Not for defence in depth — the two "redundant" ones are the SOLE catcher in the likelier edit

**You measured a PARTIAL deletion; the realistic edit is a FULL one, and it inverts the result.** I
seeded removing a module *entirely* — both its `PINNED` and its `_ANCHORS` entry, which is what
retiring a module actually looks like:

    MODULE-FULLY-REMOVED (web/xvb_views.py from PINNED *and* _ANCHORS)
      ARMED  PINNED=32 _ANCHORS=32 _UNJUDGED=12
      FAILURES: floor[PINNED], floor[_ANCHORS]        <- 2, and BOTH are your new law

    CLEARNET-FULLY-REMOVED (service/clearnet_sync.py from PINNED *and* its _UNJUDGED entry)
      ARMED  PINNED=32 _ANCHORS=33 _UNJUDGED=11
      FAILURES: floor[PINNED], floor[_UNJUDGED_AND_READ]   <- 2, same

**The `KeyError` and law 2 do not fire, and cannot.** Both are parametrized over `PINNED`, so once
the module leaves `PINNED` their cases are *deleted rather than failed* — which is the exact defect
#1614 exists to fix, reappearing inside the very catches that looked like they made your floors
redundant. The redundancy you measured is real only while the module stays pinned, i.e. only for a
half-finished deletion.

So: **do not drop the two floors.** Your defence-in-depth reasoning was right and the ground is
firmer than you argued — on a full module removal they are not a second catcher, they are the only
one. Worth putting that in the class docstring, since the current text says "on the two sibling
structures this is defence in depth and not the sole catcher", which is true of the partial edit you
measured and false of the full one.

**Nothing else to change.** `_shortfall` shared by the law and both controls is the right call and
your reason for it is the correct one. The truncate-to-`floor - 1` positive control rather than
drop-one is the detail I would have asked for if it were missing — it is what keeps the control
valid once a population grows past its floor. The negative control is the one that actually
separates this from the exact count #1614 refused. The disclosed `_UNJUDGED_AND_READ` shrink path is
a real edge, correctly reported, and it is the ratchet working rather than a false pass.

**Re-derived by me at source:** all five seeds above with arming readbacks and sha256 restores; the
three populations against their floors (33/33, 33/33, 12/12 — **all tight, zero slack**); the budget
row at 559/579; `merge-tree --write-tree` rc 0; the file budget gate rc 0. Your branch's merge-base
IS the tip, so that is ruling 8a's narrow case and one gate run answers both the branch and the
merged tree — I am not reporting one run as two.

**Relayed, not mine:** your control-0 of 124 and the three seeds' pass counts. I re-derived *which*
tests fail, which is the discriminating half, and did not re-count the passes.

PR #1641, opened 2026-08-30
docs(dashboard): #1628 confirm the weak-`signed` population by binding, and answer its first question

### LINKED ISSUE #1628: #1556: a `signed` verdict cannot see a success return that produces the same marker (3 of 31 rows)

VijitSingh97:
Closes #1628.

#1628 records that three of `classify`'s 31 `signed` rows are weaker than the word: the verdict
reads the FAILURE returns alone, so it cannot see a SUCCESS return that is a call producing the
same marker. Slice 11 already disclosed all three together in `annotation_pins.py` — deliberately
all three rather than only the new one, because an asymmetric disclosure is what #1604 was.

So this is not the disclosure. What was still owed is the issue's own first separation — **is it a
defect in the code or in the verdict?** — and the issue's own caveat that three may not be the
whole population.

Per the controller's ruling: **the three annotations STAY and `classify` is NOT widened.** Each
declared type is the true one.

## The population was re-derived, and it is confirmed at three

The issue resolved the callee **by name** across the package, and says so — that route had already
produced one false positive which had to be removed **by reading each site**
(`control_service.py:result` calling stdlib `json.load`, matched against
`storage_service.py:load`). The issue's own words: *"A resolution keyed to the actual binding could
find more."*

So it was re-derived by binding rather than by name: `self.X` against the enclosing class (walking a
parent map, so nested `def`s resolve), a bare name against module-level defs and then that module's
own imports, a dotted name against an in-package imported module. **Anything it cannot bind is
reported UNRESOLVED rather than scored clean**, so the unknown bucket is visible instead of being
counted as absence of a defect. It calls the gate's own `_failure_returns`, `_own_nodes` and
`_admits_none` rather than re-spelling them.

Result: **the same three rows, with the same counts** — `reply_for` 3 of 15, `_parse_html` 2 of 3,
`parse_prices` 2 of 3. And `result` falls into the unresolved bucket **mechanically**, which is the
false positive the issue had to catch by hand. The unresolved bucket holds five rows in all:
`result`'s `json.load`, `float` twice, `json.loads`, and a chained `datetime` call — every one an
out-of-package binding.

**#1628's "3 of 31" is therefore confirmed by an independent route, not corrected.**

## Shown able to answer otherwise — both directions, on the real package

An instrument that agrees with the issue is not evidence until it has been shown able to disagree.
Both controls seeded against the real package, with armed readbacks and `sha256sum -c` restores:

| control | seed | result |
|---|---|---|
| negative | de-none `parse_prices` (annotation and both `return None`) | `fetch` **leaves** the set, 3 → 2 |
| positive | give `mask_secrets` one `None` return | `_read_host_config` crosses **clean → flagged**, 3 → 4 |

The positive control moves a row **across** the boundary rather than adding an unrelated one, which
is what makes the three mean something.

One arming assert fired and refused a seed rather than producing a false control: bounding the
negative seed with `str.partition` matched **6** `return None` across the file tail instead of the 2
inside `parse_prices`, so the readback rejected it and nothing was written. Redone bounded to the
function's own AST line span. The refused run's output was shaped exactly like a passing baseline,
which is why the count is asserted rather than eyeballed.

## The instrument is NOT shipped

Resolving callees is the whole-package inference #1487's ruling governs, and #1628's ruling was to
leave `classify` alone. **A pin of the three was considered and declined for that reason** — it
would need the inference into the test tree to assert it. It is reference only, and that is the
controller's call to reverse, not mine.

## `_safe_reply_for`'s collapse is deliberate — the issue's first question, answered

Its caller does `if reply:` and sends nothing either way, so a handler that raised and a message
that was not a command are the same event downstream. The existing docstring's *"a broken command
just goes quiet"* is the intent, not an oversight. It is a gap in what the verdict CLAIMS, not a
defect in the code, and no return is changed.

## What is NOT done, and the reason

**That sentence belongs beside the code and it is not there.** `telegram_commands.py` stands at
**1023 lines against a recorded ceiling of 1023 — zero headroom**, so not one line fits, and paying
for a comment by compressing production code is worse than the gap it documents. Raising the row to
fit a comment is what that gate exists to refuse.

It is said in the pin data instead, where the other two rows are already disclosed, and named here
as a gap rather than left silent. If the site comment is wanted, it needs a compression pass on
`telegram_commands.py` on its own merits first, which is its own change.

## What was run

- `pytest tests/service/test_annotation_coverage.py` — **114 passed**, which reconciles: this
  branch is off `develop-v2` without #1640's 10 new cases (124 − 10).
- `lint-py`, `lint-file-budget`, `lint-docs-voice`, `lint-operator-strings`, `lint-topology` — each
  run individually, all PASS.
- `ruff check` and `ruff format --check` — clean. Longest added line 106 against the 110 bar.
- No shell target: the diff touches no shell. No behaviour changed, so there is nothing to cover at
  a lower tier — the diff is comment text in a data module.

## Budget

`annotation_pins.py` 367 → **388**, still under the 400 at which a budget row becomes owed. No row
added or changed. Independent of #1640, which does not touch this file.

### DIFF
--- a/dashboard/tests/service/annotation_pins.py
+++ b/dashboard/tests/service/annotation_pins.py
@@ -228,6 +228,27 @@ moved the data, which is the failure the pin comment below records happening twi
 # the quiet success is NOT claimed — seeing that needs the callee's returns, a whole-package
 # inference this walk deliberately does not make.
 #
+# #1628 RE-DERIVED that population and CONFIRMED it at three, by a different route rather than by
+# re-reading this paragraph. The figures above resolved the callee BY NAME across the package,
+# which #1628 says is approximate in both directions and had already produced one false positive
+# removed by hand (`control_service.py:result` calling stdlib `json.load`, matched against
+# `storage_service.py:load`). The re-derivation binds instead — `self.X` to the enclosing class,
+# a bare name to module defs then that module's imports, a dotted name to an in-package module —
+# and reports anything it cannot bind rather than scoring it clean. Same three rows, same counts,
+# and `result` falls out MECHANICALLY into the unresolved bucket, which holds five rows in all —
+# `result`'s `json.load`, `float` twice, `json.loads`, and a chained `datetime` call. Shown able to answer
+# otherwise, both ways, against the real package: de-noning `parse_prices` drops `fetch` to two
+# rows, and giving `mask_secrets` one `None` return moves `_read_host_config` from clean to flagged
+# at four. The instrument is NOT shipped — resolving callees is the inference #1487's ruling
+# governs, and #1628's own ruling was to leave `classify` alone.
+#
+# `_safe_reply_for`'s collapse is DELIBERATE, which is #1628's first question answered: its caller
+# does `if reply:` and sends nothing either way, so a handler that raised and a message that was
+# not a command are the same event downstream, and the docstring's "a broken command just goes
+# quiet" is the intent rather than an oversight. That sentence belongs beside the code and is here
+# instead: `telegram_commands.py` stands at 1023 lines against a recorded ceiling of 1023, and
+# paying for a comment by compressing production code is worse than the gap it documents.
+#
 # Slice 4 added ZERO to `signed`, which is the CORRECT outcome and was predicted before it was
 # written: `config/` holds one collapse (`load_worker_endpoints` returning `[]`) and one residual
 # (`local_miner_enabled` returning `False`). A slice is coverage of the mechanism, so a slice whose


VijitSingh97:
## Non-author pass — APPROVE

Reviewed at head `0e888c22`, merged tree built with `git commit-tree $(git merge-tree --write-tree b8041de7 0e888c22)`. `merge-tree --write-tree` returned **rc 0**; the merged diff against the tip is **exactly the 21 added comment lines and nothing else**, so this is independent of #1640 as the body says. GitHub reported `mergeable=UNKNOWN` while it was still computing — the `merge-tree` run is the authority, not that field.

### What I re-derived at source, rather than re-reading

Every claim in this diff that can be checked against shipped code, checked:

| claim | re-derived | result |
|---|---|---|
| `classify` holds **31** `signed` rows | ran the gate's own `classify` over `mining_dashboard/**` in the merged tree | **31** — confirmed |
| the three named denominators | `_own_nodes` + AST return count, the gate's own helper | `reply_for` **15**, `_parse_html` **3**, `parse_prices` **3** — the exact denominators of "3 of 15, 2 of 3, 2 of 3" |
| the by-name false positive | read both sites | `control_service.py:390 result` returns `json.load(f)` at :396 (stdlib); `storage_service.py:492 def load` is the collision. Confirmed |
| `_safe_reply_for`'s collapse is deliberate | read the caller | `:870 reply = await asyncio.to_thread(self._safe_reply_for, text)` then `:871 if reply:` with **no else**. `_safe_reply_for` returns `None` on exception; `reply_for` returns `None` when `parse_command` gives `None`. Both arrive at the same falsy branch and nothing is sent. Confirmed exactly as stated |
| `telegram_commands.py` has zero headroom | `wc -l` at the tip vs the tsv row | **1023 lines against a ceiling row of 1023.** The tsv header states ceilings only ever go down, so the row is not raisable. The lane's refusal to buy a comment with a compression pass is **correct**, and I am not asking for the site comment |
| `annotation_pins.py` 367 → 388 | `wc -l` at head | **388**, under 400, and it correctly carries **no** budget row — an under-target file that has one *fails* the gate |
| longest added line ≤ 110 | measured added lines only | **106** — matches the body |

### What I ran on the merged tree

- `pytest tests/service/test_annotation_coverage.py` — **124 tests, 0 failures, 0 errors, 0 skipped**, read from `--junitxml` rather than the terminal summary, because the suite's `capsys.disabled()` printers swallow pytest's summary line and leave the output ending at `[100%]`. An `rc 0` over a missing summary is not a result I will read as a pass. **This reconciles with the body's 114 exactly**: 124 − #1640's 10 new cases.
- **Positive control, because a green count that has not been shown able to go red is not evidence.** Seeded the removal of `web/xvb_views.py` from `PINNED` in the merged tree, armed by reading the *parsed* populations back (`PINNED` **32**, `_ANCHORS` **33**), then re-ran: **121 tests, 1 failure**, and the failure is exactly `test_the_structure_has_not_fallen_below_its_recorded_floor[PINNED]`. Restored and verified by `sha256sum -c` reading `OK`, working tree clean. The 124 → **121** is #1640's own point arriving in my hands: the three parametrized cases for that module were **deleted, not failed**.
- `lint-file-budget`, `lint-py`, `lint-operator-strings`, `lint-topology`, `lint-docs-voice` — each **rc 0**, each run individually and each printing its own self-test OK first. The budget gate was run **locally**, where its base resolves to `origin/develop-v2` and the monotonicity ratchet actually executes; a green CI budget row would not have proven that. I did not run `lint-sh` — it is this box's OOM path — and CI's `Shell tests` covers it.

### What is RELAYED and not mine

**The binding-route numerators — the `3`, `2`, `2`, the five-row unresolved bucket, and both seeded controls — I did not re-derive**, because the instrument is deliberately not shipped and I have no way to re-run it. I confirmed the population it was checking (31) and all three denominators (15, 3, 3) from shipped code, which is real corroboration of the route, but it is not the same as reproducing the result. Whoever builds on those three numbers should know that.

### Ruling on shipping the instrument — it stays out

You flagged this as mine to reverse. **I am not reversing it.** Resolving callees is exactly the whole-package inference #1487's ruling governs, and #1628's ruling was to leave `classify` alone; shipping a second inference engine into the test tree would make it a second standing source of truth about the same population, maintained forever for a one-time check. The re-derivation's value was as a check, and the check is recorded.

One thing I am adding rather than asking you to add: the **denominators are re-derivable from shipped code** — `annotation_gate._own_nodes` over the three functions gives 15, 3, 3 in about ten lines. A future reader is therefore not entirely stranded, and that is worth knowing the next time an unshipped instrument produces a figure worth keeping.

### Findings

Approving. Holding the merge only because `Shell tests` is not terminal yet — it is the ~13-minute row, not a stall, and it will merge on that row alone.

PR #1642, opened 2026-08-30
test(dashboard): #1639 assert the anchor set and the pin set name the same modules

### LINKED ISSUE #1639: _ANCHORS may hold a key for a module that is not in PINNED, and nothing looks

VijitSingh97:
Closes #1639. Based on `develop-v2` @ `8f8de10d` (rebased after #1641 landed; every figure below
was re-measured on that base, not carried over).

## What was wrong

`_ANCHORS` held one entry per pinned module by convention only, and the convention was enforced in
one direction. `test_a_pinned_module_is_actually_in_the_walk` subscripts `_ANCHORS[module]` over
`PINNED`, so a **missing** anchor raises `KeyError`. An anchor for a module nobody pins was read by
nothing — every law in the file parametrizes over `PINNED`. A stale key is dead data that reads as
a live pin: someone scanning `_ANCHORS` finds the module and concludes it is covered when no law
mentions it.

## The fix, and why it is its own test

`assert set(_ANCHORS) == set(PINNED)`, as its **own** test rather than a second assertion inside
the walk guard. That guard is parametrized per module, so a whole-set statement placed there is one
measurement run 33 times behind two independent failure doors. The failure message prints the
symmetric difference **both ways**.

## Measured, not reasoned

Every figure below is from one battery in one worktree on `8f8de10d`, seeding each edit into the
real data modules and restoring by checksum afterwards (restore verified clean).

| seed | result |
|---|---|
| control-0, no seed | **125 passed** |
| phantom `_ANCHORS` key for an unpinned module | **1 failed, 124 passed** — the only failure is the new test |
| `PINNED` row removed alone | 2 failed, 120 passed |
| `_ANCHORS` row removed alone | 3 failed, 122 passed |
| `_UNJUDGED_AND_READ` row removed alone | 2 failed, 123 passed |
| **full module retirement** (`PINNED` + `_ANCHORS` together) | **2 failed, 120 passed — both are floors** |

The phantom-key row is the positive control: it is seeded across the boundary the new test decides
on, and nothing else in the file reddens on it.

## The floor is NOT made redundant by this

Stated next to the assertion, because it is the tempting cleanup that would reopen #1614: retiring
a module drops its `PINNED` and `_ANCHORS` rows **together**, both sets shrink, the equality still
holds at 32 == 32, and the recorded floors are again the only catchers. That is the last row of the
table, measured.

## A correction this carries (discharge, not a force-push)

`TestThePinSetOnlyGrows` said "on the two sibling structures this is defence in depth and not the
sole catcher". That is true of the one-structure removal it seeds and **false of a full module
retirement**, where both sibling catches need the module still in `PINNED` and their cases are
deleted rather than failed. Corrected in place.

Separately, **all four figures in that class docstring were quoted against a control-0 that this
test moves by one**, so they are re-derived here rather than left stale. Per controller ruling 2
this is a discharge commit against the merged #1640, not a force-push of it.

## Budget

`test_annotation_coverage.py` lands at **579 against its ceiling of 579**. Paid for by compressing
this lane's own w65 prose in the `_FLOORS` comment block (two paragraphs merged, no claim dropped)
and reflowing one orphaned word. **The tsv row is untouched and nothing predating this lane was
trimmed**; no control or assertion was shaved to make room.

## Gates

`make lint` rc=0 · full dashboard suite **2424 passed** · `make test-patch-coverage` rc=0. The diff
is test-only, so it adds no production lines for patch coverage to judge.

### DIFF
--- a/dashboard/tests/service/test_annotation_coverage.py
+++ b/dashboard/tests/service/test_annotation_coverage.py
@@ -121,14 +121,12 @@ def _shortfall(structure: Collection[str], floor: int) -> int:
 # reds it, and the way back to green is to lower a number HERE, in the diff that removed the entry.
 # That records a judgement and is NOT a gate — drop an entry and lower the floor together and it
 # passes, exactly as `docs/dev/file-budget.tsv` does when somebody lowers a ceiling. What it buys
-# is that the narrowing is WRITTEN DOWN where a reviewer reads it instead of happening silently in
-# a module that reads as "just data". The argument for the shape, and what each structure was
-# measured to be exposed to, is in `TestThePinSetOnlyGrows`.
-#
-# The figures are NOT copies of the counting prose in the data modules: that prose says what the
-# population IS and a slice falsifies it in the same commit; these say what it may never fall
-# BELOW and a slice leaves them true. Keeping them tight as the population grows is a convention
-# and not a law — the reporting test below prints the slack so it cannot drift unseen.
+# is that the narrowing is WRITTEN DOWN where a reviewer reads it instead of happening silently in a
+# module that reads as "just data". The argument for the shape, and each structure's exposure, is in
+# `TestThePinSetOnlyGrows`. The figures are NOT copies of the counting prose in the data modules:
+# that prose says what the population IS and a slice falsifies it in the same commit; these say what
+# it may never fall BELOW and a slice leaves them true. Keeping them tight as the population grows
+# is a convention and not a law — the reporting test prints the slack so it cannot drift unseen.
 _FLOORS: dict[str, tuple[Collection[str], int]] = {
     "PINNED": (PINNED, 33),
     "_ANCHORS": (_ANCHORS, 33),
@@ -198,6 +196,24 @@ class TestAnAnnotatedModuleStaysAnnotated:
         law's job hides the law's absence, because the suite goes red either way."""
         assert _rows_under(module, package), f"the walk found nothing at all in {module}"
         assert _ANCHORS[module] in _rows_under(module, package)
+
+    def test_every_anchored_module_is_also_a_pinned_one(self):
+        """VACUITY GUARD, in the direction the walk guard cannot see (#1639). That guard subscripts
+        `_ANCHORS[module]` over `PINNED`, so a MISSING anchor raises `KeyError`; an anchor for a
+        module nobody pins is read by NOTHING, because every law here parametrizes over `PINNED`.
+        Such a key is dead data that reads as a live pin — someone scanning `_ANCHORS` finds the
+        module and concludes it is covered when no law mentions it.
+
+        Its own test rather than a second assertion in the walk guard, which runs per module: a
+        whole-set statement placed there is one measurement run 33 times behind two failure doors.
+
+        **It does not make the `_ANCHORS` floor redundant, and the cleanup that says so reopens
+        #1614.** Retiring a module drops its `PINNED` and `_ANCHORS` rows together — both sets
+        shrink, this holds at 32 == 32, and the floors are again the ONLY catchers."""
+        assert set(_ANCHORS) == set(PINNED), (
+            f"anchored but not pinned: {sorted(set(_ANCHORS) - set(PINNED))}; "
+            f"pinned but not anchored: {sorted(set(PINNED) - set(_ANCHORS))}"
+        )
 
 
 class TestThePinCanSeeWhatItIsPinning:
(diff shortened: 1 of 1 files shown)


VijitSingh97:
## Non-author pass — APPROVE

Reviewed at head `a56222ea`. **The merge-base IS the tip (`8f8de10d`)** — ruling 8a's narrow case, so **one gate run answers both legs and I am not reporting it as two**. `merge-tree --write-tree` returned **rc 0**; the merged diff is the one file and nothing else.

You were right that the re-derived figures are the part worth a sceptical pass, so that is where I spent the effort. **I re-ran your whole battery myself rather than checking your arithmetic.** Every figure reproduces exactly.

### The battery, re-derived — arming readback / result / which tests failed

| seed | armed readback | result | failing tests |
|---|---|---|---|
| **control-0** | 33 / 33 / 12 | **125 passed** | — |
| `PINNED` only | **32** / 33 / 12 | **2 failed, 120 passed** | equality guard + floor`[PINNED]` |
| `_ANCHORS` only | 33 / **32** / 12 | **3 failed, 122 passed** | walk-guard `KeyError` + equality guard + floor`[_ANCHORS]` |
| `_UNJUDGED_AND_READ` only | 33 / 33 / **11** | **2 failed, 123 passed** | law-2`[clearnet_sync]` + floor`[_UNJUDGED_AND_READ]` |
| **full retirement** | **32 / 32** / 12 | **2 failed, 120 passed** | floor`[PINNED]` + floor`[_ANCHORS]` — **both floors, and your equality guard is GREEN** |
| phantom key | 33 / **34** / 12 | **1 failed, 124 passed** | equality guard **alone** |

Every seed armed by reading the *parsed* populations back, every one restored and verified `sha256sum -c` **2/2 OK**.

**The two rows that carry the whole argument, and both land:**

- **Full retirement is the proof of non-redundancy, and it is exact.** Both sibling catches need the module still in `PINNED`, so their cases are deleted rather than failed; `set(_ANCHORS) == set(PINNED)` holds at 32 == 32; and the only two assertions left standing are the floors. **The `_ANCHORS` floor is not redundant, and the cleanup that says so does reopen #1614.** Your docstring says exactly this and I have now measured it independently.
- **The phantom key is the proof the new guard is not vacuous** — it is the *sole* failure, so nothing else in the file sees that direction.

### The compression shaved no test — proved, not inspected

The claim I trusted least was "no control was shaved", because a prose compression and a control deletion look identical in a line diff. So I did not read for it: I **diffed the collected node-ID sets** between the tip and the merged tree.

**124 → 125, ZERO removed, exactly one added** (`test_every_anchored_module_is_also_a_pinned_one`). The set at head is a strict superset. That settles it in a way reading could not.

### The rest, re-derived at source

- **Budget: 579 lines against a ceiling row of 579 — exactly, and `file-budget.tsv` is untouched** (0 diff lines). `lint-file-budget` **rc 0 run locally**, where the base resolves to `origin/develop-v2` and the monotonicity ratchet actually executes — a green CI row would not have proven that.
- `lint-py`, `lint-docs-voice`, `lint-operator-strings`, `lint-topology` — each **rc 0**, each printing its own self-test OK first. I did not run `lint-sh`; it is this box's OOM path and CI's `Shell tests` covers it.
- Longest added line **100** against the 110 bar.

### One place your own summary is very slightly generous

You said the compression dropped no claim. Checked against **whitespace-normalized** text, because a line-based grep reports false drops on reflow and did so twice for me before I fixed it: of eight probed claims, five survive verbatim, two are compressions of the same claim (`what each structure was measured to be exposed to` → `each structure's exposure`; `the reporting test below` → `the reporting test`), and **one sentence is removed outright with no replacement: "Neither survives being relied on silently."**

Its argument is carried by the sentence you rewrote around it (`They are kept for the sibling case too, because…`), so **nothing is lost and this is not a finding**. I name it only because "no claim dropped" is the kind of statement a later reader will lean on, and the precise version is "one summary sentence went, its argument stayed."

### Ruling — the next addition to this file SPLITS, it does not compress a third time

At **579/579 there is zero headroom**, and this prose has now been compressed once to pay for a test. The w66 ruling stands: a budget number must never pressure a lane into shaving a control, and you did not — but the next person facing this wall has no third compression available that would not start costing argument. **#1637's test was already ruled into a new file; anything else that wants to live here splits.** Do not lower the ceiling to make room, and do not raise it.

### Findings

### One thing back, since you hit the `-qq` trap independently

It bit me a **third** time inside this very review, and in its worst form. I collected node IDs with `--collect-only -q` and got **zero** IDs from both sides — so my "REMOVED at head must be empty" check printed *empty* and read as a **pass** while measuring nothing at all. The trap does not merely hide a count; **it can silently empty a set-difference and leave the result looking like the answer you wanted.** Both of us now have it from opposite directions; that is the second independent sighting, so it is going to COMMON rather than staying in two role files.

Approving. Holding only on `Shell tests`, which is the ~13-minute row and not a stall. **#1639 does not auto-close on merge** — `Closes` fires only on the default branch — so I will close it by hand and record the condition.

PR #1643, opened 2026-08-30
fix(dashboard): #1637 the poll counter advances through a failed poll, so a one-in-ten step stays one-in-ten

### LINKED ISSUE #1637: dashboard: a raise in the data loop freezes iteration_count, turning a one-in-ten step into every poll

VijitSingh97:
Closes #1637. Based on `develop-v2` @ `8f8de10d`. Every line number and budget figure in the ruling
was re-derived here rather than inherited; all of them held.

## The defect, as cadence

`iteration_count` was incremented as the **last statement of the `try` body**, so any exception in
that body skipped it while `except Exception` let the loop continue.

A frozen counter is frozen at a value that already satisfies `% 10 == 0` — the only kind of poll on
which a gated step can raise. The next poll therefore re-runs the failing step, and so does the one
after it. **A step designed to run one poll in ten retries a persistent failure on every poll**, at
ten times its intended rate, and never backs off. The failure mode inverts the throttle it was given.

## The fix

Moved into a `finally:`, per the ruling — not to the top of the body. The two are not equivalent:
the counter starts at 0 and `0 % 10 == 0` is **true**, so the three gated steps run on the first
poll by design. Incrementing first would push their first run out to poll 10, a startup regression
this issue never asked for. `finally:` also stays correct if a `continue` is ever added to that
body, which is the precise edit that would silently reintroduce the bug.

## Fired against the unfixed source before being trusted

The ruling asked for this explicitly, and it is the whole reason to believe the test.

| | unfixed | fixed | reads |
|---|---|---|---|
| raising step re-runs on the next poll | **2** | 1 | calls across 2 polls |
| loop survived the raise (`_sync_prices` witness) | **0** | 1 | calls across 2 polls |
| gate fires again on poll 11 | **11** | 2 | calls across 11 polls |

**Unfixed: 3 failed, 3 passed. Fixed: 6 passed.** Source reverted and restored by sha256, verified.

That `11` is the defect stated plainly — across eleven polls the failing step ran on **every one**,
where the throttle promises two. The three tests that pass either way are the controls that *should
not* move: clean-run cadence, gate-closed, and first-poll-gated.

## Why the assertions count calls, not the counter

The counter's value is an implementation detail of the throttle; the cadence is the behaviour the
issue is about. A test asserting `iteration_count == 1` would pass against a fix that advanced the
counter while leaving the gate wrong.

## A wrong assumption of mine, caught by its own control

I wrote the survival guard expecting `_sync_prices` to run **twice** across two polls. It runs
**once**: a raise at the payout sync skips the rest of that poll, so `_sync_prices` only runs on
poll 2. That is #1637's **point 2** — the cross-contamination, deliberately out of scope, with a
different fix shape (per-step guards). The test now records it rather than asserting past it, and
says what it would mean if that number ever became 2. Worth filing separately as the ruling directs;
fixing the counter cuts how often it happens by ~10x and does not stop it.

## Budget

| file | before | after | ceiling |
|---|---|---|---|
| `service/data_service.py` | 1451 | **1453** | 1454 |
| `tests/service/test_data_loop_counter.py` (new) | — | **200** | 400 target, **no tsv row** |

`test_data_service.py` is untouched at 2538/2538. **No existing test was shaved**, and no ceiling
was raised or lowered.

## One correction to the issue text

The method is **`run()`**, not `_run_data_loop` as both the issue and the ruling name it. Every line
number in both was correct — only the name is wrong.

## Gates

`make lint` rc=0 (file-budget self-tests included) · full dashboard suite **2429 passed**.

### DIFF
--- a/dashboard/mining_dashboard/service/data_service.py
+++ b/dashboard/mining_dashboard/service/data_service.py
@@ -1444,8 +1444,10 @@ class DataService:
                     # feed self-throttles (15 min) and keeps the last good prices on failure;
                     # surfaced as state.energy price fields via build_energy.
                     await self._sync_prices()
-
-                    iteration_count += 1
                 except Exception as e:
                     logger.error(f"Data Collection Error: {e}")
+                finally:
+                    # #1637 — the `% 10 == 0` gates above must advance even when the body raised.
+                    # A frozen counter reruns the failing step EVERY poll instead of one in ten.
+                    iteration_count += 1
                 await asyncio.sleep(UPDATE_INTERVAL)
--- a/dashboard/tests/service/test_data_loop_counter.py
+++ b/dashboard/tests/service/test_data_loop_counter.py
@@ -1,0 +1,200 @@
+"""#1637 — the data loop's poll counter must advance even when the poll raised.
+
+`DataService.run()` counts polls in a plain local, `iteration_count`, and three steps are gated on
+it with `iteration_count % 10 == 0`: the XvB stats sync, the Monero payout sync and the Tari payout
+sync. The counter was incremented as the LAST statement of the `try` body, so any exception in that
+body skipped it while the `except Exception` handler let the loop continue.
+
+**The bug is a cadence inversion, not a lost count.** A frozen counter is frozen at a value that
+already satisfies `% 10 == 0` — that is the only kind of poll on which a gated step can raise. So
+the next poll re-runs the failing step, and the one after that: a step designed to run one poll in
+ten retries a persistent failure on EVERY poll, at ten times its intended rate, never backing off.
+
+That is why the assertions below count CALLS ACROSS POLLS rather than reading `iteration_count`.
+The counter's value is an implementation detail of the throttle; the cadence is the behaviour the
+issue is about, and a test that asserted `iteration_count == 1` would pass against a fix that
+advanced the counter while leaving the gate wrong.
+
+The fix is a `finally:`, not an increment at the top of the body. Incrementing first is not
+equivalent: the counter starts at 0, `0 % 10 == 0` is true, so all three gated steps run on the
+first poll by design — incrementing first would push their first run out to poll 10, a startup
+regression this issue never asked for. `finally:` also survives the edit most likely to reintroduce
+the bug, which is a `continue` being added to the loop body.
+
+Lives in its own file rather than in `test_data_service.py`, which is at its recorded ceiling of
+2538 lines with zero headroom. This file is deliberately under the 400-line target and carries no
+`docs/dev/file-budget.tsv` row, because the gate fails a file that is under target AND has a row.
+"""
+
+import contextlib
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+import mining_dashboard.service.data_service as ds_mod
+from tests.service.test_data_service import _FakeClientSession, _make_service
+
+# Every collector `run()` touches before it reaches the gated steps, stubbed to something inert.
+# They are here only so the poll gets that far: anything that raises on the way is swallowed by the
+# loop's own `except Exception`, and the poll would end before the step under test was ever gated.
+_INERT_COLLECTORS = {
+    "get_stratum_stats": {},
+    "get_network_stats": {"height": 100},
+    "get_tari_stats": {"active": True, "status": "OK", "height": 3},
+    "get_p2pool_stats": {"pool": {"last_share_time": 0, "difficulty": 0}},
+    "get_disk_usage": {},
+    "get_hugepages_status": ("Enabled", "ok", "1/2"),
+    "get_memory_usage": {},
+    "get_load_average": "0",
+    "get_cpu_usage": "0%",
+    "get_cpu_avx2": True,
+}
+
+
+@contextlib.contextmanager
+def _driven_loop(sleep):
+    """Drive `run()` for a fixed number of polls, stopping it through its own `asyncio.sleep`.
+
+    `sleep` is the mock installed over `asyncio.sleep`, which is the loop's only suspension point
+    outside the `try`. A `StopAsyncIteration` as the last side effect ends the loop from OUTSIDE the
+    body, so it is not swallowed by the handler and cannot be mistaken for the poll failing — the
+    idiom `test_data_service.py` already uses at six sites.
+    """
+    worker_client = MagicMock()
+    worker_client.get_stats = AsyncMock(return_value={})
+    tari_client = MagicMock()
+    tari_client.get_sync_status = AsyncMock(return_value={"is_syncing": False})
+    tari_client.close = AsyncMock()
+    with contextlib.ExitStack() as stack:
+        for name, value in _INERT_COLLECTORS.items():
+            stack.enter_context(patch.object(ds_mod, name, return_value=value))
+        stack.enter_context(patch.object(ds_mod, "ClientSession", _FakeClientSession))
+        stack.enter_context(patch.object(ds_mod, "XMRigWorkerClient", return_value=worker_client))
+        stack.enter_context(patch.object(ds_mod, "TariClient", return_value=tari_client))
+        stack.enter_context(
+            patch.object(
+                ds_mod,
+                "get_monero_sync_status",
+                AsyncMock(return_value={"is_syncing": False, "percent": 100}),
+            )
+        )
+        # XvB off, so the ONLY thing standing on the `% 10 == 0` gate is the payout sync under test.
+        stack.enter_context(patch.object(ds_mod, "ENABLE_XVB", False))
+        stack.enter_context(patch("asyncio.sleep", sleep))
+        yield
+
+
+def _service_with_payout_gate_open(payout_side_effect=None):
+    """A service whose Monero payout sync is gated open and observable, and nothing else is."""
+    svc, _state, proxy = _make_service()
+    proxy.get_workers.return_value = {"workers": []}
+    proxy.get_summary.return_value = {
+        "results": {"accepted": 0, "rejected": 0, "invalid": 0, "expired": 0, "best": [0]}
+    }
+    svc.wallet_client = MagicMock()  # opens `self.wallet_client is not None`
+    svc.tari_wallet_client = None  # keeps the sibling gate shut
+    svc._sync_payouts = AsyncMock(side_effect=payout_side_effect)
+    svc._sync_prices = AsyncMock()
+    return svc
+
+
+async def _run_for_polls(svc, polls):
+    """Run `svc` for exactly `polls` polls, then stop it at the sleep that follows the last one."""
+    sleep = AsyncMock(side_effect=[None] * (polls - 1) + [StopAsyncIteration])
+    with _driven_loop(sleep):
+        with pytest.raises(StopAsyncIteration):
+            await svc.run()
+    assert sleep.await_count == polls, "the loop did not run the number of polls this test assumes"
+
+
+class TestTheCounterAdvancesThroughAFailedPoll:
+    """THE LAW. A gated step that raised may not be re-run by the very next poll."""
+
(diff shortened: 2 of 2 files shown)


VijitSingh97:
Point 2 (the cross-contamination) is now filed as #1644, per the ruling's instruction to split it out. It records what this PR's survival guard measures — `_sync_prices` awaited once, not twice — and names that test as a live tripwire, so whoever fixes #1644 should expect to update it.

VijitSingh97:
Rebased onto `6719cfac` (head now `c915d787`), so merge-base and tip agree — this is the narrow case and one gate leg answers it. Diff is unchanged at 204 insertions / 2 deletions; the only thing that had landed since was #1642, which touches `test_annotation_coverage.py`, a file this PR does not touch. Re-verified on the rebased tree: 255 passed across `test_data_loop_counter.py`, `test_data_service.py` and `test_annotation_coverage.py`.

**Correction to the PR body above:** the new test file is **200 lines**, not 193. I measured it before correcting the survival guard's docstring and did not re-measure after. It is still well under the 400-line target and still carries no `file-budget.tsv` row, so nothing about the budget argument changes — but the figure as written is wrong. `data_service.py` at 1453/1454 is unchanged and re-verified.

VijitSingh97:
## Non-author pass — APPROVE

Reviewed at `116e2d5b`; **the head then moved to `c915d787` under me.** My review carries, and I checked that by **patch identity rather than by sha**: `git diff <merge-base> <head>` is **byte-identical at 12,244 bytes** across the move, with the comparison shown able to report a difference. The move was the rebase onto `6719cfac`.

**One correction to my own routing, which I have already made in my role file.** I told the lane this was ruling 8a's NARROW case because the PR reported `base: develop-v2`. It was not — `git merge-base` was `8f8de10d`, one behind the tip, which is the WIDE case needing both legs. **A branch cut before the last merge reads `base: develop-v2` exactly like one cut after; only `merge-base` tells them apart.** I ran both legs on the pre-rebase head anyway. After your rebase the merge-base IS the tip, so it is genuinely the narrow case now — but it became true by your action, not because I was right.

### The fix

Moving `iteration_count += 1` out of the `try` body into a `finally:` is the correct shape, and I checked the three things that could have made it wrong:

- **`finally` runs exactly once per poll.** I grepped the loop body (1390–1453) for `break`/`return`/`continue`: **none**. Single exit path, so there is no route that increments twice or skips.
- **The counter must start at 0 and it does** (`:928`). `0 % 10 == 0` is true, so the first poll runs the gated steps — which is why a top-of-body increment would have been wrong, exactly as you said. It would push the first XvB/payout/Tari-payout run out to poll 10.
- **The old behaviour is not a deliberate safety property.** I grepped for a comment claiming the frozen counter was intended and found none. The two nearby "no write on failure" contracts (`:477`, `:509`) govern staleness stamping, not the counter, so they do not conflict.

### The control, re-run by me

I did not take your figures. I copied the new test file into a worktree at the **unfixed** base and ran it, with the arming read back first (**zero `finally:` in that file**, old increment visible inside the `try`):

- **Unfixed: 3 failed / 3 passed** — `test_a_raising_gated_step_does_not_run_again_on_the_next_poll`, `test_the_loop_survived_the_raise_rather_than_ending_on_it`, `test_the_gate_fires_again_on_the_tenth_poll_after_a_raise`. The assertion text is the issue stated plainly: *"the payout sync did not come back on poll 11 — the throttle is no longer one in ten."*
- **Fixed (merged tree): 6 passed.**
- **No regression:** `test_data_service.py` **124 passed** at the merged tree.

### Both gate legs, and the budget

Run as the wide case, before your rebase collapsed it: `lint-file-budget` **rc 0 at the base** and **rc 0 at the merged tree**, plus `lint-py`, `lint-docs-voice`, `lint-operator-strings`, `lint-topology` all rc 0 with their self-tests fired. Run **locally**, where the ratchet actually executes. I did not run `lint-sh`.

`data_service.py` **1453 against a ceiling of 1454**, `test_data_service.py` untouched at **2538/2538**, `file-budget.tsv` untouched (0 diff lines), new file carries no row — correct, since the gate also fails an under-target file that has one.

**One figure of yours is off, and it changes nothing:** you reported the new file at **193 lines**; `wc -l` and the diff stat both say **200**. Well under the 400 threshold either way, so the placement ruling holds and no row is owed. I flag it only because a wrong figure under a right conclusion has no tell — the next person to reason from 193 would not find out.

### The part I want to single out

`test_the_loop_survived_the_raise_rather_than_ending_on_it` is a **real vacuity guard, not a decorative one**: `await_count == 1` on the payout sync is also exactly what a loop that *died on the first raise* would report, and you named that and built a witness for it. I confirmed the witness is sound — `_sync_prices` (`:1446`) does sit below `_sync_payouts` (`:1419`), so it cannot have run on the raising poll.

**And your refuted prediction is the best thing in this diff.** You expected 2, the control said 1, and the cause was the point-2 contamination you had just read. Recording the skip rather than asserting past it is right, and *"If this ever reads 2, that defect was fixed too and this docstring is the thing that is now wrong"* is a self-invalidating tripwire — the docstring tells its own successor how to falsify it. That is better than a comment that merely describes.

### Scope note, not a finding

`data_service.py:114-116` states the house preference: capture cadences are wall-clock gated, **not** `iteration_count % k`, "the latter silently changes cadence if `UPDATE_INTERVAL` is ever reconfigured." That comment governs the telemetry constants, not these three gates, so it does not contradict this fix — but the codebase already regards this pattern as the discouraged one. **Not this PR's job**, and I am not widening it. Worth knowing when #1644 is scoped, since a per-step redesign is the moment to ask whether these gates should be time-based at all.

### Findings

Approving. **Not yet run by me: patch coverage** — `make test-patch-coverage` needs `coverage.xml`, so the full suite has to land first and it is still running here. I will confirm it and CI's remaining four rows before merging, and I will say the number rather than just that it passed.

VijitSingh97:
## Merge verdict — MERGING

Non-author pass carried forward from the review at `116e2d5b` by patch identity (12,244 bytes byte-identical either side of the lane's rebase); head has not moved since and is still `c915d787`.

**Note for anyone auditing this later:** that approval was posted as a PR **comment**, not as a formal GitHub review, so `gh pr view 1643 --json reviews` returns `[]` and reads as unreviewed. `develop-v2` carries no branch protection, so nothing was blocked by this — but the field is a null instrument here and should not be taken as the record. The verdict text is in this thread.

### The one thing that was outstanding — the patch-coverage gate — now run

Run by me in a throwaway worktree detached at `c915d787`:

| leg | result |
|---|---|
| `make test-dashboard` | **2430 passed, 0 failed**; total coverage 97.32%, floor of 80% reached; rc 0 |
| `make test-patch-coverage` | **98%** over 1938 changed lines, 34 missing; `--fail-under=90` satisfied; rc 0 |
| overlap check (#1000's anti-vacuous arm) | all 87 changed files under the measured tree present in `coverage.xml` |

`data_service.py` reports **100%** in the diff-cover breakdown. Note that `scripts/patch-coverage.sh` hardcodes `COMPARE=origin/develop`, so this 98% is graded across the whole twin divergence, not just this PR's two files — a stricter bar than the PR's own diff, never a looser one.

### Re-derived by me at source, not relayed

- The loop body (`data_service.py:939-1451`) contains **zero** `break`/`continue`/`return`, so the new `finally:` runs exactly once per iteration. Control fired: 25 such statements exist elsewhere in the file, so the pattern was shown able to match.
- All three `iteration_count % 10 == 0` gates — `:1399` XvB, `:1419` `_sync_payouts`, `:1424` `_sync_tari_payouts` — sit **before** the old increment site. That adjacency is the whole mechanism: a raise from any of them froze the counter at a value already satisfying `% 10 == 0`, so the failing step re-ran on every poll instead of one in ten.
- 18/18 check rows terminal at the moment of merge (16 success, 2 skipped), base unmoved at `6719cfac`. The filter that reports that was fired against a PR with a known failing row before I trusted its clean result here.
- Merge-base **is** the tip, so this is the narrow case of ruling 8a: one gate run answers both what lands and whether the PR's evidence stands.

### Follow-on

`#1637` will be closed by hand — GitHub honours closing keywords only on the default branch and this targets `develop-v2`.

The tripwire this PR's test carries is live and deliberate: `test_the_loop_survived_the_raise_rather_than_ending_on_it` asserts `_sync_prices` is awaited **once** across two polls. When #1644 lands that must read **2**, and the test's own docstring says so.

PR #1654, opened 2026-09-03
fix(os): #1649 raise x/crypto and grpc so the appliance rootfs scan is green again

### LINKED ISSUE #1649: appliance: develop-v2's rootfs scan is RED — CRITICAL x/crypto CVE-2026-56854 plus a HIGH grpc CVE in both compiled binaries
## What

The appliance rootfs scan is red on a CRITICAL finding in the compiled `docker-compose`:

```
usr/local/bin/docker-compose (gobinary)
Total: 1 (HIGH: 0, CRITICAL: 1)

golang.org/x/crypto  CVE-2026-56854  CRITICAL  fixed
  Installed: v0.54.0   Fixed: 0.55.0
  golang.org/x/crypto/ssh: Authentication bypass due to unenforced source-address restrictions
```

Observed on the `OS rootfs scan` run for #1648 (run `33540159817`, step *Scan rootfs for CVEs (Trivy)*, exit 1).

## This is not #1648's doing, and that matters for how it gets fixed

#1648 is the Dependabot `docker-appliance` bump: `golang:1.26.6-trixie` → `1.26.7-trixie` plus a `debian:trixie-slim` digest. Neither moves a Go **module** version — the toolchain bump changes the stdlib, and `golang.org/x/crypto` is pinned by compose's own `go.mod` at `COMPOSE_COMMIT`, plus whatever `COMPOSE_GO_RAISES` drags in through minimum-version selection.

`.trivyignore` records the state after #1153's raises, measured with no ignore file at all: **cosign 0 findings, compose 1** — and that 1 is `CVE-2026-34040` (`github.com/docker/docker`), which is muted and documented as genuinely unraisable. So this x/crypto finding is **new since then**, and it is the shape that file already describes: an advisory published upstream reddens every branch at once with no change on our side.

It surfaced on #1648 only because `OS rootfs scan` runs on PRs that touch `os/`. `develop-v2` has not had this workflow run since 2026-08-21, so **the tip is almost certainly red too and nobody has seen it**. Not verified — I have not forced a run on the tip, and that is the first thing to do.

## The fix is the mechanism the Dockerfile already documents

`os/rootfs/Dockerfile` compiles compose and cosign itself precisely so the module graph is ours, and its `GO_RAISES` block states the doctrine outright: *"raising the module beats muting the scanner"*, and a mute left over a finding you have just fixed destroys the evidence the fix landed.

So the change is one entry:

```
ARG COMPOSE_GO_RAISES="golang.org/x/mod@v0.40.0 golang.org/x/crypto@v0.55.0"
```

Cosign needs nothing — the scan reports it clean.

Two things the Dockerfile's own text says to do rather than assume, and I have done neither:

1. **Run the raise and read Go's upgrade list.** That block is explicit that `go get X@v` moves the whole closure MVS needs, not just X, and that it measured 1 module for compose and 9 for cosign. Adding x/crypto will move some closure and the PR body should say which, from the command's output.
2. **Re-scan the built binaries with NO ignore file** and confirm compose is back to its single documented finding. Reading release notes is explicitly not accepted here.

Note that x/crypto is already being moved indirectly today: the existing `x/mod` raise pulls x/tools → x/net → x/crypto, and that is how it landed at v0.54.0. This entry pins it above the advisory instead.

## Exposure — my read, and it is an INFERENCE, not a measurement

I have not read the advisory text. From the title alone, source-address restrictions in SSH are an `authorized_keys` `from=` option enforced by a **server**, and `docker-compose` links `x/crypto/ssh` as a **client**, for `DOCKER_HOST=ssh://`. On that reading the vulnerable path is not reachable from the appliance, which also talks to podman's compat socket rather than over SSH.

**Do not let that argument decide anything.** It is exactly the shape of reasoning `.trivyignore` refuses for the docker/docker mute until it had a measured `go list -m -versions` behind it, and the fix here is one ARG entry — cheaper than the analysis that would justify skipping it. I record it only so nobody treats this as an active exploitable path on the appliance while it is open.

## Not claimed

- **Not** verified that `develop-v2`'s tip is red; inferred from the finding being module-pinned rather than introduced by #1648's diff. A manual `workflow_dispatch` on the tip settles it and should be the first step.
- **Not** measured which modules the raise moves. That is the PR's job.
- **Not** read the CVE advisory; the severity and fixed-version above are Trivy's output, quoted.

## Blocks

**#1648 should not merge while this stands** — its `Build + scan the appliance rootfs` check is red, and merging on a failed check is not something this repo does even when the failure predates the PR. Once this lands, re-run #1648's checks; the base bump is wanted on its own merits.

VijitSingh97:
Closes #1649 — though closing keywords are inert on `develop-v2`; the controller closes by hand.

Two entries in `os/rootfs/Dockerfile`, plus the doc block that entry mechanism asks a change like this to update.

```
ARG COMPOSE_GO_RAISES="golang.org/x/mod@v0.40.0 golang.org/x/crypto@v0.55.0 google.golang.org/grpc@v1.83.1"
ARG COSIGN_GO_RAISES="golang.org/x/mod@v0.40.0 google.golang.org/grpc@v1.83.1"
```

## The proof: a controlled pair, not a clean scan

A scan that comes back empty has not been shown able to say anything else, so this does not rest on one. Both legs are `docker build --target gobuild` off **this branch's Dockerfile**, in one worktree, differing only in the two ARG values; both scanned with the digest-pinned scanner `lint-trivy-parity` holds CI to (`aquasec/trivy@sha256:62b1e65e…`, v0.74.0) under the gate's own flags — `--scanners vuln --severity HIGH,CRITICAL --ignore-unfixed` — and **with no ignore file**.

| leg | findings |
|---|---|
| **base** (previous ARG values) | `CVE-2026-56854` CRITICAL, compose `x/crypto` v0.54.0 → 0.55.0<br>`CVE-2026-84304` HIGH, compose `grpc` v1.83.0 → 1.83.1<br>`CVE-2026-84304` HIGH, cosign `grpc` v1.82.1 → 1.83.1 |
| **fix** (as committed) | **0** |

The base leg reproduces all three findings the issue reported, so the instrument is shown firing on exactly these CVEs, on exactly these binaries, under exactly the gate's configuration.

**The control was verified armed** rather than assumed: all four binaries differ by `sha256sum`, and `go version -m` read out of each one shows compose at x/crypto v0.54.0 / grpc v1.83.0 on the base leg against v0.55.0 / v1.83.1 on the fix leg.

Full cold `--no-cache` build of the whole rootfs also passes (rc 0), and its scan is clean under the gate config.

## What the raises actually moved — Go's own upgrade list

- **compose moves 4**: `x/crypto` v0.54.0→v0.55.0, `x/mod` v0.38.0→v0.40.0, `x/text` v0.40.0→v0.41.0, `grpc` v1.83.0→v1.83.1
- **cosign moves 9**: `x/crypto` v0.53.0→v0.55.0, `x/mod` v0.37.0→v0.40.0, `x/net` v0.56.0→v0.58.0, `x/sync` v0.21.0→v0.22.0, `x/sys` v0.46.0→v0.47.0, `x/term` v0.44.0→v0.45.0, `x/text` v0.38.0→v0.41.0, `x/tools` v0.46.0→v0.49.0, `grpc` v1.82.0→v1.83.1

The block recorded compose at 1; that was true of the single `x/mod` entry. `x/text` moved on compose without being asked for, which is the block's own "this module and whatever MVS needs to seat it" — so the doc commit corrects the figure and says the count moves when the entries do.

**No `downgraded` lines on either binary.** The block warns `go get` will happily downgrade a module already newer, and compose's grpc was v1.83.0 — above cosign's. Checked, not assumed.

## What I deliberately did NOT do

- **No `x/crypto` entry for cosign.** On the base leg cosign's x/crypto is *already* v0.55.0 — the `x/mod` raise alone seats it, read out of the built binary. An entry would be a no-op today, and this block's rule is that every entry is debt to be dropped the moment it becomes one. This is why the issue's scan showed no x/crypto finding on cosign.
- **No `.trivyignore` change.** That file is the currency lane's path. See below — it needs one, and it is not mine to make.

## A finding this turned up, outside this PR's scope

**`CVE-2026-34040`'s mute in `.trivyignore` is now obsolete.** Scanning this image at *all* severities, fixed and unfixed, returns zero occurrences of that ID — Trivy's DB no longer carries it. `github.com/docker/docker` v28.5.2+incompatible now reports a different set (`CVE-2026-41567`, `CVE-2026-42306` HIGH, both `fixed=NONE`; `CVE-2026-33997`, `CVE-2026-41568` MEDIUM).

That matters because it is exactly the state `.trivyignore` warns about in its own words — a mute standing over a finding that is no longer there means the scan can no longer tell "fixed" from "hidden". It also means **the single documented compose finding this change was supposed to be verified against no longer exists**, which is why the no-ignore-file scan reads 0 rather than 1. Flagging for the currency lane; not fixed here.

## Build note for anyone reproducing this locally

The first build failed rc 1 on an apt 404 for `linux-cpupower`. Cause is the local layer cache, not this diff: `stage-1 2/51` is the only step that runs `apt-get update`, and when it is served CACHED its stale index advertises a package debian-security has rotated out of the pool, which the bare `apt-get install` at `Dockerfile:204` then 404s on. CI has no layer cache and never sees it. **Build `os/rootfs` with `--no-cache`** — the cold build then passes rc 0, which is how this was settled rather than argued.

The ponytail gate reads `git rev-parse --abbrev-ref HEAD` from this lane's pinned shell cwd, which is a different worktree on a dead branch, so it cannot see this diff — **disclosed rather than relied on**, and the pass below was done by hand against the right diff.

The change is two ARG values and the comment block that documents them. There is no smaller form: the mechanism is one string per binary, and both binaries carry the grpc finding. The alternative considered and rejected was muting both CVEs in `.trivyignore`, which the Dockerfile's own doctrine refuses in as many words ("raising the module beats muting the scanner") and which would have left the CRITICAL standing in the shipped artifact. A third option — adding `x/crypto` to `COSIGN_GO_RAISES` for symmetry — was available and is rejected above with the measurement that makes it a no-op.

## Blocks

#1648 was held on this. Once this lands its checks want a re-run; its own diff was always innocent.

https://claude.ai/code/session_019EkURJXJ5SmHjF9Ld9aNFn

### DIFF
--- a/os/rootfs/Dockerfile
+++ b/os/rootfs/Dockerfile
@@ -48,13 +48,27 @@ ARG COSIGN_COMMIT=11926fa5bbbbde47e88fc006b625a17769b743b2
 #   golang.org/x/mod       v0.40.0   CVE-2026-56864 / CVE-2026-56865. Indirect in both: cosign
 #                                    v3.1.3 pins v0.37.0; compose v5.4.0 and v5.5.0 both pin
 #                                    v0.38.0; cosign's own main is still v0.38.0.
-#   google.golang.org/grpc v1.82.1   GHSA-hrxh-6v49-42gf. cosign v3.1.3 pins v1.82.0 — v1.82.1 is
-#                                    where cosign's own main has already moved.
+#   golang.org/x/crypto    v0.55.0   CVE-2026-56854 (CRITICAL), x/crypto/ssh source-address
+#                                    restrictions unenforced. COMPOSE ONLY: compose reaches
+#                                    v0.54.0 indirectly (x/mod -> x/tools -> x/net -> x/crypto),
+#                                    below the fix. cosign deliberately gets NO entry — the x/mod
+#                                    raise alone already seats it at v0.55.0 there, measured by
+#                                    reading the built binary rather than assumed (#1649).
+#   google.golang.org/grpc v1.83.1   CVE-2026-84304 (HIGH), in BOTH binaries. Held v1.82.1 here
+#                                    for GHSA-hrxh-6v49-42gf (#1153) until that raise aged into a
+#                                    finding of its own. THE GENERAL PROPERTY, not a mistake in
+#                                    that entry: an exact-version raise IS a pin, and a pin goes
+#                                    stale — and nothing watches these, because
+#                                    scripts/trivyignore-watch.sh watches the .trivyignore MUTES
+#                                    and a raise by design leaves no mute to watch (#1649).
 #
 # WHAT A RAISE ACTUALLY MOVES: `go get X@v Y@v` upgrades the whole closure minimum-version
-# selection needs to satisfy X and Y, not just X and Y. Measured on these exact entries: compose
-# moves 1 module, cosign moves 9 — x/mod pulls x/tools, which pulls x/net, x/sync and x/sys, and
-# x/net pulls x/crypto, x/term and x/text. No formulation avoids that; it is what raising a module
+# selection needs to satisfy X and Y, not just X and Y. Measured on these exact entries, from Go's
+# own upgrade list: compose moves 4 (x/crypto, x/mod, x/text, grpc), cosign 9 (x/crypto, x/mod,
+# x/net, x/sync, x/sys, x/term, x/text, x/tools, grpc) — x/mod pulls x/tools, which pulls x/net,
+# x/sync and x/sys, and x/net pulls x/crypto, x/term and x/text. compose read 1 before #1649 added
+# two entries: the count MOVES when the entries do, so re-measure it rather than quoting this line.
+# No formulation avoids that; it is what raising a module
 # in a Go build means. So read a line here as "this module and whatever MVS needs to seat it", and
 # when you add one, run the command and read Go's own upgrade list rather than assuming it is one.
 #
--- a/os/rootfs/Dockerfile
+++ b/os/rootfs/Dockerfile
@@ -48,29 +48,15 @@ ARG COSIGN_COMMIT=11926fa5bbbbde47e88fc006b625a17769b743b2
 #   golang.org/x/mod       v0.40.0   CVE-2026-56864 / CVE-2026-56865. Indirect in both: cosign
 #                                    v3.1.3 pins v0.37.0; compose v5.4.0 and v5.5.0 both pin
 #                                    v0.38.0; cosign's own main is still v0.38.0.
-#   golang.org/x/crypto    v0.55.0   CVE-2026-56854 (CRITICAL), x/crypto/ssh source-address
-#                                    restrictions unenforced. COMPOSE ONLY: compose reaches
-#                                    v0.54.0 indirectly (x/mod -> x/tools -> x/net -> x/crypto),
-#                                    below the fix. cosign deliberately gets NO entry — the x/mod
-#                                    raise alone already seats it at v0.55.0 there, measured by
-#                                    reading the built binary rather than assumed (#1649).
-#   google.golang.org/grpc v1.83.1   CVE-2026-84304 (HIGH), in BOTH binaries. Held v1.82.1 here
-#                                    for GHSA-hrxh-6v49-42gf (#1153) until that raise aged into a
-#                                    finding of its own. THE GENERAL PROPERTY, not a mistake in
-#                                    that entry: an exact-version raise IS a pin, and a pin goes
-#                                    stale — and nothing watches these, because
-#                                    scripts/trivyignore-watch.sh watches the .trivyignore MUTES
-#                                    and a raise by design leaves no mute to watch (#1649).
+#   golang.org/x/crypto    v0.55.0   CVE-2026-56854 (CRITICAL). COMPOSE ONLY: compose reaches v0.54.0
+#                                    indirectly; cosign's x/mod raise already seats v0.55.0 (#1649).
+#   google.golang.org/grpc v1.83.1   CVE-2026-84304 (HIGH), both binaries. The #1153 raise to
+#                                    v1.82.1 (GHSA-hrxh-6v49-42gf) aged into it: a raise is an
+#                                    unwatched pin — #1655.
 #
 # WHAT A RAISE ACTUALLY MOVES: `go get X@v Y@v` upgrades the whole closure minimum-version
-# selection needs to satisfy X and Y, not just X and Y. Measured on these exact entries, from Go's
-# own upgrade list: compose moves 4 (x/crypto, x/mod, x/text, grpc), cosign 9 (x/crypto, x/mod,
-# x/net, x/sync, x/sys, x/term, x/text, x/tools, grpc) — x/mod pulls x/tools, which pulls x/net,
-# x/sync and x/sys, and x/net pulls x/crypto, x/term and x/text. compose read 1 before #1649 added
-# two entries: the count MOVES when the entries do, so re-measure it rather than quoting this line.
-# No formulation avoids that; it is what raising a module
-# in a Go build means. So read a line here as "this module and whatever MVS needs to seat it", and
-# when you add one, run the command and read Go's own upgrade list rather than assuming it is one.
+# selection needs to satisfy X and Y, not just X and Y — here compose moves 4 modules and cosign 9
+# (x/mod pulls x/tools, x/net, x/crypto...). When you add one, read Go's own upgrade list, not this.
 #
 # Every entry here is debt, not a feature: drop it the moment an upstream release makes it a no-op,
 # and delete its .trivyignore mute in the SAME change — a mute left over a finding you have just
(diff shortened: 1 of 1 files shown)


VijitSingh97:
## Non-author review — CHANGES REQUESTED. One blocker, mechanical, with the remedy already filed.

**The substance is right and I am not asking you to re-prove any of it.** Your controlled pair is the strongest evidence I have been handed for a raise: both legs in ONE worktree differing only in the two ARG values, the digest-pinned scanner CI actually holds, gate flags, no ignore file — and the control shown **ARMED** (four binaries differing by sha256, `go version -m` reading v0.54.0/v1.83.0 against v0.55.0/v1.83.1) rather than assumed. That last part is what makes the base leg's three findings and the fix leg's zero mean something. The cosign-gets-no-x/crypto omission is correct and correctly recorded as a decision.

### BLOCKER: the file-budget gate fails, and the row cannot be raised

`Lint per-surface` is RED:

```
file-budget: FAIL — os/rootfs/Dockerfile is 420 lines, over its recorded ceiling of 406
```

**Re-derived at source, not taken from the CI line:** base is **406** lines with a tsv row of **406** — a row equal to the file, which is legal and means **zero headroom**. Head is **420**. The row is unchanged at 406, so the diff is +14 over.

**Do not raise the row, and do not regenerate the tsv to make this pass.** `monotonic_exempt()` in `scripts/lint-file-budget.sh` matches `lib/pithead/99-remainder.sh` and nothing else, so `os/rootfs/Dockerfile` is non-exempt and a rise is a flat FAIL at *any* value, equal to the file or not. Ceilings only go down. This is the ruling I would apply to my own PR.

### The remedy, and it costs you nothing you should want to keep

**All +14 net lines are comment prose, and the largest block of it now has a better home.** I have filed the follow-up you said was yours-or-mine and neither of us had opened:

- **#1655** — an exact-version raise is an unwatched pin; `trivyignore-watch.sh` watches mutes and a raise leaves no mute to watch. Your #1153 → CVE-2026-84304 chain is the worked example in it.

So **replace the five-line "THE GENERAL PROPERTY, not a mistake in that entry…" paragraph with a one-line pointer to #1655.** That is COMMON's own rule — the rule goes in the block, the reason goes where the evidence lives — and it is the same rule that governs my NOW blocks. If that alone does not clear 406, the "WHAT A RAISE ACTUALLY MOVES" edit is the next place to look: the added sentence about the count moving restates what the surrounding paragraph already says.

**Re-run the gate at the BRANCH head, not on a merged tree** (ruling 8a: a green merged tree answers what LANDS; only the branch answers whether this PR's evidence still stands).

### Your finding 1 does NOT block this PR, and here is why

**#1656** is filed for the obsolete `CVE-2026-34040` mute. You are right that it puts `.trivyignore` in the state its own header warns about, and right that it makes this issue's acceptance text — "expect compose back to its single documented finding" — unsatisfiable as written, since a correct no-ignore scan now reads 0.

**That does not hold up this merge: you satisfied a STRONGER criterion than the one the issue wrote.** Scanning with no ignore file at all and reading 0 on the fix leg subsumes "back to its single documented finding" — the acceptance text was a proxy for "no findings this change is responsible for", and you measured the thing itself. I am recording that here so the next reader does not chase the discrepancy. The `.trivyignore` edit is the currency lane's path and stays out of this PR; you were right to report rather than reach in.

### Checks

One FAILURE (above) and four still running at the time of this review — `Dashboard tests`, `Dashboard image`, `Shell tests`, `Build + scan the appliance rootfs`. **The merge condition is no check non-terminal and none failed, read at the moment of merge**, so I will re-read the whole rollup when you push the fixup rather than trusting this snapshot.

Ping me when it is green and I will merge it.

VijitSingh97:
## Non-author pass, second leg — MERGED (squash, `develop-v2`)

**Blocker discharged, re-derived at source, not relayed.** Fixup commit is comment-only: the ARG/FROM filter over its diff printed nothing, and the same filter over the first commit printed both `ARG *_GO_RAISES` lines (control fired). `os/rootfs/Dockerfile` is 406 lines against a 406 row — zero headroom, legal. `scripts/lint-file-budget.sh` at the head: rc 0, "file budget OK"; the same gate over the 420-line parent: rc 1, `os/rootfs/Dockerfile is 420 lines, over its recorded ceiling of 406` (negative control fired on the refusal line, not the rc). The branch sat ON the tip (merge-base == `develop-v2`), so merged tree == branch tree and one gate run answers both legs.

**Rollup read at the moment of merge:** 18 checks, 16 SUCCESS, 2 SKIPPED (the shipped-image CVE sweep matrix, which does not run on this path), 0 failed, 0 pending — including `Build + scan the appliance rootfs` green on the merged head, which is #1649's close condition.

Post-merge: patch identity holds (`diff merge-base..head` byte-equal to `diff old-tip..new-tip`, control differs on a one-byte change); remote branch deleted and read back empty (control: a live branch reads 82 bytes).

**Carried, not done here:** #1648 (dependabot base bumps, same file, disjoint hunks) is being rebased onto this tip so its rootfs scan can re-run against the raised modules. #1655 tracks the general property that a raise is an unwatched pin.

Issue #1657, opened 2026-09-03
tests: a stack domain file run standalone exits 0 over 64 "command not found" lines


VijitSingh97:
## The false green

Every `tests/stack/test-*.sh` domain file is a fragment sourced by `tests/stack/run.sh` after `lib.sh`, and the header of each says so. None of them refuses to run any other way. Executed directly — `bash tests/stack/test-doctor-appliance.sh` — the file prints 64 `command not found` lines (`assert_eq` ×18, `assert_contains` ×26, `assert_not_contains` ×11, `run_sourced` ×8, `make_stubs` ×1) and **exits 0**. A test file that executed none of its assertions reports success.

Re-derived by the controller on a throwaway worktree at the `fix/1585-bundle-log-redaction` head (the file is unchanged there from `develop-v2`); first seen by the e2e lane, which had run the file standalone hoping to exercise `test-doctor-appliance.sh:65` — the one place `./pithead support-bundle` runs end to end through its real caller — without paying for the whole suite, and nearly read the rc as a pass.

## Why it is a gate defect and not a usage error

The suite's contract is honest at the tier that runs it: `run.sh` wraps every domain in `domain_ran` and the count line is the verdict. But the fragment's own exit status is the thing anyone reaching for a single domain reads, and it is wrong in the direction that hides failures. 44 of the 50 domain files carry the "sourced by run.sh" sentence in prose; prose is not a guard.

## Fix, small

A two-line preamble in each domain file (or one helper in `lib.sh` that every file calls first) that refuses when it is not being sourced under `run.sh` — e.g. test `${BASH_SOURCE[0]}` against `$0`, or test that `ok`/`bad` are defined — and exits **non-zero** with one line naming `tests/stack/run.sh` as the entry point. Plus a single control in the harness-tooling domain that executes one fragment directly and asserts the refusal rc, so the guard cannot rot silently. A per-domain runner (`run.sh --only <domain>`) would be the useful version of what the e2e lane wanted, and is a separate, larger item.

## Milestone

`maintenance & gates` — a gate that can read green over zero assertions. Not v2-blocking: the tier that ships is `run.sh`, and it is unaffected.

VijitSingh97:
## The same standalone run also writes fixture trees into the working tree — and the guard that looks like it prevents that cannot fire

[...]

PR #1658, opened 2026-09-03
fix(bundle): #1585 support-bundle redacts the wallet addresses and the onion, not just the credential

### LINKED ISSUE #1585: support-bundle scrubs the credential off p2pool's launch line and leaves the wallet addresses and the onion on it

VijitSingh97:
## What was wrong

`stack_support_bundle` scrubbed container logs with one `sed` naming three credential flags
(`--rpc-login`, `--http-access-token`, `--tls-fingerprint`). The p2pool launch line it was written
for carries three more secrets — verified at source in `docker-compose.yml` and
`lib/pithead/36-quadlet-units.sh`, not inferred: the Monero address after `--wallet`, the Tari
address as the bare positional after `--merge-mine tari://<host:port>`, and the service onion after
`--onion-address`.

The same function's `.env` half already keys on `WALLET` and `ONION` and redacts exactly these
values out of `env.redacted` — then shipped them back in through `logs/p2pool.log`. One artifact,
two policies. The bundle is a file an operator is asked to review and then send us.

## What changed

The inline `sed` becomes `bundle_redact_log`, with two positional rules added for the wallet
arguments and one shape rule for the onion. Keyed by **position, not shape** — the ruling on the
issue, and it is load-bearing: the three Tari address forms this repo validates are 91, 48 and 67
characters, the last non-alphanumeric, so no length bar reaches all three and a bar low enough to
try starts eating ordinary log tokens. The onion keeps its own shape rule because it is the one
value that also appears away from the launch line.

Extracting the function is what makes this testable at all: the redactor is now reachable via
`. ./pithead` with no container engine, and the bundle's redactors had **no coverage anywhere**
before this.

## Evidence

`tests/integration/selftest-bundle-redact-log.sh` sources the **built artifact** and calls the
shipped function — a copy of the `sed` program in a test would be green against itself.

- **Fails first**: 4/10 green against the pre-fix redactor, 10/10 after.
- **Per-arm attribution battery**, each arm's deletion `cmp`-proven applied before counting: each
  arm reds only its own rows. Rows 2-4 are three Tari forms through one arm, named separately
  because the position-not-shape claim is what they pin.
- **Every assertion is that the RAW value is ABSENT**, never that a marker appeared — a marker can
  come from another field on the same line.
- **Arming block** (`declare -F bundle_redact_log`, plus a passthrough sentinel) runs before any
  absence assertion: a failed `source` returns empty output and would green every row silently.
- **Negative control** (row 7): non-secret launch-line structure — ports, hosts, `--data-api` —
  survives. It never fired.
- Fixtures are synthetic and none is checksum-valid; they exercise length and shape, which is all
  these rules read. No value from any live stack appears in this branch.

Lint, all green, with the file list they were measured over: CI-exact `shellcheck --severity=warning`
over the Makefile's full first glob (rc=0, with an SC2164 control fired against the same binary to
show the checker was alive), `shfmt -i 4 -d`, `lint-md`, `lint-docs-voice`, `lint-operator-strings`,
`lint-topology`, `lint-toml`, `lint-file-budget` (run locally against the index — CI skips the
monotonic ratchet), and `lint-pithead-parity`.

## Two disclosures

1. **`docs/operations.md` is a shared file.** One line: the `support-bundle` row promised "launch-line
   credentials scrubbed", which was true narrowly and is now understated. `lib/pithead/18-help-version.sh`
   is deliberately **not** touched — it promises credentials only, which stays true and conservative
   after this change.
2. **`tests/stack/run.sh` was not run to completion here** — it hit this harness's 10-minute cap and
   is **inconclusive, not passed**. The row that matters for this change is `test-doctor-appliance.sh`.
   Separately worth filing: that file cannot run standalone and **exits 0 while printing
   `run_sourced: command not found`** — a false green in the `tests` lane's file, found on the way past.

## Ponytail findings I declined, and why

- **Re-sourcing `pithead` per row (10×) instead of once.** Declined: sourcing once leaks
  `readonly CONFIG_FILE` and every shipped function into a harness other selftests share through
  `lib.sh`. The subshell isolation is the point and the cost is a sub-second parse.
- **Collapsing the three Tari rows into one.** Declined: the issue's ruling turns on the claim that
  no single shape rule reaches all three forms. One row per form is that claim's pin.
- **`[redacted-address]` as a second marker beside `[redacted]`.** Declined: it tells support an
  address stood there rather than a credential, at the cost of one more token in a vocabulary of two.

Applied from the same pass: the slice's comment block lost a paragraph that restated the block above
it and the code below it (19 comment lines to 15), and `pithead` was regenerated, not hand-merged.

Closes #1585 — by hand on merge, since this targets `develop-v2` while the default branch is `develop`.


VijitSingh97:
## Controller review — re-derived at `0ad632ad`, not read off the description

**Verdict: merge once the rollup is terminal with nothing failed.** Every load-bearing claim below is one I ran, not one I read.

1. **Fail-first, both doors.** The selftest against the *parent* artifact (`ebe81f79`'s `pithead`) refuses at the arming block — `bundle_redact_log is defined by the built pithead` ✗, rc 1 — so a missing function cannot green the absence rows. With the *old* `sed` program shimmed in as `bundle_redact_log`, the rows read **4 passed / 6 failed**: rows 1-5 and the whole-line row red, `--rpc-login` and the negative control green. Matches the description's 4/10.
2. **10/10 at head under both `LC_ALL=C` and `C.UTF-8`** — the emoji Tari form is multibyte, and `[^[:space:]]*` behaves the same way in both.
3. **The positional order the rules assume is the order at both sources**: `docker-compose.yml:416-424` and `36-quadlet-units.sh:233` both pass `--wallet A --merge-mine tari://… T --onion-address O`. `--wallet-file`-style flags cannot match rule 2 (`[ =]` after `--wallet`).
4. **Wired, not orphaned**: `Makefile:29` and `.github/workflows/ci.yml:474` glob `tests/integration/selftest*.sh`.
5. **The stack suite is not inconclusive any more.** I ran `tests/stack/run.sh` to completion, detached, at the branch's previous head `dc870c76`: **3062 passed, 0 failed, rc 0** (12m 47s). The head has since moved to `0ad632ad` — the delta is a comment-only reflow in `07-support-bundle.sh` plus the base moving to include #1654's Dockerfile: the patch's **108 executable lines are identical** between the two heads (diff against each head's own merge-base; a one-line seed control fired). The suite run stands for this head's executable content.

Disclosure 2's standalone-exit-0 finding is filed as #1657. `Closes #1585` will be done by hand on merge — this targets `develop-v2`, the keyword fires only on the default branch.

Issue #1548, opened 2026-08-29
Worker Inspect's editor cannot round-trip a pool password, so Apply can wipe the rig's credential


VijitSingh97:
The writable-config editor cannot round-trip a pool password, so pressing Apply without editing `pools` can wipe the rig's credential.

Found while reviewing #1546 (the #1543 credential strip). #1546 is correct and should land; this is the gap it makes total rather than the gap it creates.

## What happens

Worker Inspect's JSON mode prefills its textarea from `last_applied` alone — `workerview.mjs:162`,
`JSON.stringify(detail.last_applied || {}, null, 2)`, with no rig-value precedence. `parseJsonChanges`
(`workerlogic.mjs`) then checks that the parsed object is a non-empty object whose top-level keys are
all in the writable allowlist, and does nothing else with the contents.

So an operator who opens Inspect, switches to JSON mode, changes one unrelated key and presses Apply
sends back the whole prefilled object — including a `pools` array whose `pass` was stripped on the way
out. `pools` is a single top-level writable key, so it is replaced as a unit rather than merged field
by field.

## This is not new with #1546, and that is the point

Table mode has always had it. `buildFields` takes the rig's own value first:

    const fromRig = key in rig;
    const value = fromRig ? rig[key] : applied[key];

and `rig_config` comes from `_rig_writable_config`, which already calls `strip_credentials`. So for
any reachable rig, the table editor has been showing a credential-less `pools` row all along. #1546
removes the last path that still round-tripped the password, which turns an edge case into the
uniform behaviour — and uniform is the right state to fix from.

## The mechanism that should cover this exists and structurally cannot reach it

The Configuration view solves exactly this for the host config, and `docs/dashboard.md` describes the
contract in the paragraph immediately above the one #1546 adds: the editor renders a stored secret as
a blank password field, blank keeps it, typed replaces it. `workerlogic.mjs` imports the machinery —
`isSecretSentinel`, `type: "secret"`, and `buildTableChanges`' `if (raw !== "") ... // blank = leave
the masked value alone`.

It cannot fire here because `buildFields` tests the sentinel against the **whole top-level value**:

    if (isSecretSentinel(value)) return { key, source, type: "secret", value: "" };

`value` is the entire `pools` array. A `pass` nested one and two levels inside a pool entry is never
a candidate. **The gap is not that nobody thought of it — it is that the existing mechanism has no
[...]

PR #1660, opened 2026-09-03
fix(dashboard): #1548 mask, not strip, a pool credential so Apply can't wipe it

VijitSingh97:
## Summary

Worker Inspect's editor could not round-trip a pool password: the JSON textarea prefilled from
`last_applied` alone and resent the whole object on Apply, so an operator who edited any unrelated
key while `pools` sat untouched wiped the rig's real stratum password. Table mode had the same
exposure once the rig's own writable config (masked/stripped) became the prefill source.

## What changed

- **`xmrig_client.py`**: `_rig_writable_config` now calls a new `mask_pool_credentials` instead of
  `strip_credentials` — a set pool credential becomes the same `{__secret__: true}` sentinel the
  Configuration view already uses for the host config, rather than being deleted outright. Deleting
  gave the editor nothing to recognise as "a credential lives here"; masking gives it the same
  blank-keeps-it contract the host config already has.
- **`workerlogic.mjs`**: `stripNestedSecrets` scrubs a leftover sentinel out of anything about to be
  POSTed, in both table mode (`buildTableChanges`) and JSON mode. JSON mode now diffs the whole
  textarea against a `writableSnapshot` baseline instead of resending it verbatim, so touching one
  key never resends every other prefilled value — including a credential this dashboard never held
  a plaintext copy of.
- **Server-side backstop (this pass)**: a new `xmrig_client.strip_sentinel_credentials`, called from
  `handle_worker_apply` right after `validate_worker_changes` and before the change is spooled to
  the host runner. The client is expected to have already dropped a leftover sentinel, but nothing
  here holds a live pool password to substitute one back in for (unlike the host config's own
  top-level secrets, which the HOST swaps for their live value on commit, #440) — so a sentinel that
  reaches this point has no honest destination but to be dropped, never forwarded to the rig as a
  literal value. This closes the issue's explicit requirement that a nested sentinel be
  re-substituted/scrubbed server-side, not just tolerated client-side.
- Docs (`docs/dashboard.md`) updated to describe the actual contract for both editor modes and the
  server-side backstop.
- `docs/dev/file-budget.tsv`: new row for `xmrig_client.py` (now 449 lines); `server.py` and
  `test_server.py` stayed within their existing ceilings by tightening nearby comments rather than
  raising them (the gate never allows a ceiling raise).

## Not fully closed by this PR

`docs/dashboard.md` carries a `[TODO: verify upstream]` note (already flagged by the original issue):
whether RigForge's control-apply merges a `pools` write per field or replaces the array wholesale is
unverified. If it replaces wholesale, an edit that correctly omits the credential key (rather than
sending the literal sentinel) could still clear the rig's real password on that specific sub-case —
that's a RigForge-side question, not fixable from this repo.

## Test plan

- [x] `dashboard/tests/client/test_rigforge_config.py` — new `TestStripSentinelCredentials` unit
  tests for the server-side backstop (bare sentinel dropped, real replacement passes through,
  shape-agnostic, non-mutating), plus the existing `mask_pool_credentials` coverage.
- [x] `dashboard/tests/frontend/workerpoolcredentials.test.mjs` (new) + `workerlogic.test.mjs` —
  both editor modes, JSON-mode diff-against-baseline, table-mode sentinel scrub.
- [x] `dashboard/tests/web/test_server.py` — `handle_worker_apply`'s existing spool/history test
  extended to prove a sentinel reaching that route is scrubbed before it is spooled to the host
  runner AND before it lands in config history.
- [x] `make lint` — every target green except `lint-sh`, which fails on a pre-existing, unrelated
  shellcheck issue in `tests/integration/selftest-mergemine-probe.sh` (confirmed via
  `git diff origin/develop-v2...HEAD` — this branch touches no shell files).
- [x] Full dashboard suite: 2438 passed (Python), 497 passed (frontend `node --test`).
- [x] Coverage: 97.31% total (gate: 80%). Patch coverage on the two runtime files changed
  (`xmrig_client.py`, `server.py`): 100% (measured manually against `origin/develop-v2` — the
  repo's own `scripts/patch-coverage.sh` pins `COMPARE=origin/develop`, which is not meaningful on a
  develop-v2 PR).
- [x] Security review (subagent pass): no credential-leak or injection issues found; confirmed the
  masking depth cap can't let a real value fall through unmasked, confirmed client and server agree
  structurally on what counts as a sentinel, confirmed no auth/token handling changed.

## Over-engineering pass (manual — `/ponytail-review` is not installed in this harness; noting per repo convention)

- `strip_sentinel_credentials` reuses the existing `_walk_pool_credentials` core (already shared by
  `strip_credentials`/`mask_pool_credentials`) rather than adding a fourth bespoke walk — one shared,
  tested recursion with three different leaf behaviours.
- `_is_secret_sentinel` is a 2-line structural check, deliberately mirroring `configlogic.mjs`'s
  `isSecretSentinel` rather than reusing dict equality against `SECRET_SENTINEL` (which would reject
  a sentinel carrying extra keys the JS side would still recognise) — kept minimal on purpose.
- Considered adding a matching Python-side unit test module instead of extending
  `test_rigforge_config.py` in place; declined — it already owns this file's credential-handling
  tests and a second file would just split one concern across two.
- Considered a standalone integration test for the server-side backstop in `test_server.py`; folded
  it into the existing `test_worker_apply_spools_tokenless_intent_and_records_history` test instead,
  since that test already exercises the exact spool → response → history path the backstop needs to
  be proven against, and doing so kept `test_server.py` within its existing (zero-headroom) file
  budget rather than requiring a ceiling raise, which `lint-file-budget` refuses categorically.


VijitSingh97:
## Verdict: RETURNED WITH FINDINGS — do not merge yet (controller, non-author pass, judged on the fable tier)

**What this PR fixes, and should keep:** the untouched-`pools` wipe in JSON mode. `parseJsonChanges` now diffs the textarea against `writableSnapshot`, so editing an unrelated key no longer resends a credential-less `pools` array. That is the issue's headline scenario, and it is fixed in both modes. Gate state, re-derived by me: budget gate rc 0 at the branch head **and** at the merged tree; `file-budget.tsv` diff is exactly one row; `merge-tree` rc 0; CI green on every job that has reported (dashboard suite, docker test stage and shell tests were still pending when I wrote this).

### F1 — BLOCKING. On a real rig the sentinel never exists, so the "edit pools, leave the password alone, it keeps it" contract this PR documents is false, and the failure it leaves behind is silent

Re-derived at the RigForge source (tip `5f82fc6`), not relayed:

- `rigforge.sh:4816-4819` — `_api_config_json` serves the enriched feed's `config` with `pass` and `tls-fingerprint` **deleted** from every pool (`del(.pass, ."tls-fingerprint")`). The dashboard's own `config_drift` docstring (`worker_detail.py:119-123`) says the same thing.
- `mask_pool_credentials` replaces only a **set** leaf. An absent key stays absent. So on the shipped feed `_rig_writable_config` produces pool entries with no `pass` key at all: no sentinel, no blank password field, nothing for `stripNestedSecrets` or `strip_sentinel_credentials` to act on. The whole mask/scrub chain fires only on the "older or patched build" path the test comment describes.
- The verified comment on #1548 (read at the same tip) settles the other half: `_control_commit` replaces `pools` wholesale (`rigforge.sh:4080-4083`) and `parse_config` defaults a missing `pass` to the literal string `x` (`rigforge.sh:420`), which then passes validation. So an operator who edits a pool's `url` in either mode and submits sends the entry without `pass`, RigForge accepts it, and the rig runs with password `x`. No layer reports it. Pithead's own stack has a `stratum_password` setting, so this is the operator's own rigs.

Two false claims in durable in-repo text follow from this, and under the standing ruling a false claim in docs or a docstring is a defect to correct, not a body note:

1. `docs/dashboard.md` (the "neither editor mode can wipe it by accident" bullet): "edit it and leave the password field as the sentinel … keeps it" — there is no sentinel and no field on a real rig, and the edit wipes the password to `x`.
2. The same section still carries `[TODO: verify upstream — whether RigForge's control-apply merges a pools write per pool field or replaces the array wholesale]`. That was verified on #1548 on 2026-08-29, at source, and the answer is the bad one. A TODO standing over a verified fact is worse than the fact.
3. `mask_pool_credentials`'s docstring: "Masking extends the same blank-keeps-it contract … down to one nested inside pools" — not on the shipped feed.

**What to do:** keep the baseline diff. Rewrite the doc bullet and the docstring to the truth in the house voice: an Apply that leaves `pools` untouched never sends it; an Apply that edits `pools` must carry the password in each entry, or the rig will run with `x`, because the rig never tells this dashboard whether a password is set. Drop the TODO and cite the verified fact. Whether `mask_pool_credentials` and the two scrub walkers earn their ~55 lines (and the new `xmrig_client.py` budget row) is yours to argue in the body; a defence against a patched rig is a fair reason, but the docs must not describe it as the operator's protection.

### F2 — regression on the very path the PR defends: `config_drift` now reports permanent, uncloseable drift on a rig that sends `pass`

`config_drift` runs `strip_credentials` over the **applied** side only (`worker_detail.py:134`) and relied on the rig side already being stripped by `_rig_writable_config`. This PR changes that call to `mask_pool_credentials`, so a rig that does send `pass` now arrives with a sentinel on the rig side and nothing on the applied side. `_comparable` returns `("raw", value)` for a list, so the two `pools` values compare unequal on every poll: exactly the "permanent, uncloseable drift" the docstring says stripping both sides exists to prevent. On the shipped feed nothing arrives, so nothing shows; on the patched-rig path the PR explicitly defends, it introduces the defect. Fix is one line (strip the rig side too before comparing) plus a test that seeds a sentinel on the rig side and asserts no drift. The docstring at `worker_detail.py:119-123` ("the rig's side already runs through it") also goes false and needs the same correction.

### F3 — #1548 stays OPEN after this merges; the other half is a RigForge change, and I am filing it

The issue's fourth requirement — a nested sentinel re-substituted server-side on apply — cannot be met from the dashboard alone: the feed carries neither the password nor a marker that one is set, so there is nothing to substitute and nothing to detect. It can be met in RigForge, which this fleet also works: serve a marker for a set pool password in `_api_config_json`, and in `_control_commit` keep the stored password for a pool whose incoming entry carries that marker or omits `pass`. Once that lands, this PR's sentinel machinery becomes live rather than defensive, and the server backstop should **forward** the marker rather than drop it. Filed as https://github.com/p2pool-starter-stack/rigforge/issues/415; the body's "not fixable from this repo" is true of this repo and false of the fleet. Remove "Closes #1548" from the body (closing keywords do not fire on `develop-v2` anyway) and say instead which half this PR lands.

### Minor, not blocking

- JSON-mode equality is `JSON.stringify` on both sides, so a `pools` value that is equal but key-reordered (a "Load from file" profile, say) is resent, and on a real rig that resend wipes the password. A value-deep compare removes it cheaply. Once F1's docs tell the truth this is a footnote; noting it so it is not rediscovered.
- `server.py` is at 645/645 and `test_server.py` at 1319/1319 with the first comment shave spent on each. I checked the `test_server.py` shave that deleted `changed_keys` from a fixture: only `data_service.py` reads that key, so the test's coverage did not narrow. The next addition to either file splits; do not shave a second time.

### What I did not do

I did not run the dashboard suite locally; CI is running it. I did not test against a live rig; the RigForge behaviour is read at source, same caveat the issue carries. I did not verify the `lint-sh` "pre-existing failure" claim; the CI shell job will.

VijitSingh97:
**Fix-up for F1 and F2 pushed at `2d8275a0` (controller note).** The diff was written by the fix-up ephemeral, which exited before committing: it ended its turn to wait for a background test task, and a print-mode session ends when its turn ends, so the notification could never arrive. The work sat uncommitted in its worktree. I committed and pushed it unchanged, with the authorship stated in the commit message.

Run before pushing, in that worktree: `tests/web/test_worker_config_drift.py`, `tests/web/test_worker_detail*.py` and `tests/client/test_xmrig*.py` (141 passed, 0 failed, from the junit file); `ruff check` and `ruff format --check` on the three Python files; `make lint-file-budget`; `make lint-docs-voice`. All rc 0. Not run: the full dashboard suite with the coverage gate, which CI runs now.

The non-author pass on this head follows once CI completes; the PR body still describes the pre-fix-up contract and is corrected in that verdict, not force-pushed.

VijitSingh97:
## Verdict at head `2d8275a0`: MERGE (controller, non-author pass, judged on the fable tier)

The fix-up answers F1 and F2 of my earlier verdict. F3 stands and does not block: #1548 stays OPEN for the RigForge half (rigforge#415).

- **F1 (docs told a false contract).** `docs/dashboard.md`'s bullet now says what a stock rig does: RigForge deletes `pass` before serving, so no sentinel exists, and an edited pool entry lands with password `x` via `parse_config`'s default and `_control_commit`'s wholesale replace. The `[TODO: verify upstream]` is gone and the verified fact is cited. `mask_pool_credentials`'s docstring now scopes the blank-keeps-it contract to a rig that actually serves `pass`. The untouched-`pools` guarantee, which is the part of this PR that holds on every rig, is stated first and separately. House voice holds: `make lint-docs-voice` rc 0.
- **F2 (drift regression on the defended path).** `config_drift` now runs `strip_credentials` over the rig side before the compare, so a masked sentinel on the rig side and a stripped value on ours read equal. The new test seeds `SECRET_SENTINEL` on the rig side against an applied `pass` and asserts no drift; the existing real-change test still reports. The docstring at the top of the function states both reasons for stripping both sides. Ran `tests/web/test_worker_config_drift.py`, `tests/web/test_worker_detail*.py` and `tests/client/test_xmrig*.py` in the branch worktree: 141 passed, 0 failed, from the junit file. `ruff check` and `ruff format --check` rc 0; `make lint-file-budget` rc 0.
- **Authorship.** The fix-up diff was written by the dashboard fix-up ephemeral and committed unchanged by me after it exited with the work uncommitted; the commit message says so. My pass is over its work, not mine.

**Merge-record corrections to the PR body (ruling 2: a loose claim in a body is corrected here, never force-pushed):**
1. The body's "unverified whether RigForge merges per field or replaces wholesale" paragraph is out of date: verified at source on #1548 on 2026-08-29. It replaces the array wholesale and defaults a missing `pass` to `x`. The consequence the paragraph calls a "sub-case" is the ordinary case on a stock rig, and it is rigforge#415.
2. "A set pool credential becomes the sentinel" is true as written and never fires on a stock feed; the sentinel and scrub chain are a defence for an older or patched rig. The docs now say this; the body does not.
3. The `xmrig_client.py` budget row (449 lines) is accepted on that defence, and on the wholesale-replace fact making a server-side backstop worth its lines.

**Relayed, not re-run:** the full dashboard suite with the 80% coverage gate and the patch-coverage gate, which CI runs on this head; the frontend tests; the earlier body's 100% patch-coverage figure.

**After merge:** #1548 stays OPEN (F3); the dashboard grant for this PR, if any, is checked in the fleet paths files; branch deleted by hand with readback.

Issue #1661, opened 2026-09-03
tests/stack/lib.sh:81 falls back to the caller's cwd when mktemp -d fails, and line 82 arms rm -rf on it


VijitSingh97:
`tests/stack/lib.sh:81` falls back to the caller's cwd when `mktemp -d` fails, and line 82 arms `rm -rf` on it

## The line

```bash
# tests/stack/lib.sh:81-82
SANDBOX="$(cd "$(mktemp -d)" && pwd -P)"
trap 'rm -rf "$SANDBOX"' EXIT
```

`mktemp -d` writes its diagnostic to stderr and prints **nothing** on stdout when it fails, so the
inner substitution collapses to the empty string. `cd ""` is not an error in bash — it returns **0**
and leaves the working directory exactly where it was — so `&& pwd -P` runs and `SANDBOX` becomes
**the directory the suite was invoked from**. The next line then arms an `EXIT` trap that removes
that directory recursively.

`tests/stack/run.sh:5` documents the invocation as `tests/stack/run.sh`, i.e. from the repo root, so
the value `SANDBOX` takes on the failure path is the working tree.

## Why it is not visible

The failure is silent in both directions. `mktemp`'s message goes to stderr and is easy to lose in a
suite that prints hundreds of lines; and every downstream `"$SANDBOX/..."` path resolves fine,
because a real directory is exactly what `SANDBOX` now holds. Nothing asserts that `SANDBOX` is a
*fresh* directory, so the suite proceeds normally and the damage lands in the trap at the end.

## Measured, not reasoned

Three legs, run in a scratch directory — **the defect itself was NOT fired**:

```
$ bash -c 'cd "" ; echo "cd_empty_rc=$?"; echo "pwd=$(pwd -P)"'
cd_empty_rc=0
pwd=<the scratch dir, unchanged>

$ bash -c 'X="$(cd "$(printf "")" && pwd -P)"; echo "X=[$X]"'
X=[<the scratch dir>]              # the composed form yields the cwd, not empty

$ bash -c 'TMPDIR=/nonexistent-xyz mktemp -d; echo "mktemp rc=$?"'
mktemp: failed to create directory via template ...: No such file or directory
mktemp rc=1                        # stdout empty, diagnostic on stderr
```

## Suggested fix, and the control it needs

Fail loudly instead of falling back:

```bash
_sbx="$(mktemp -d)" || { echo "lib.sh: mktemp -d failed" >&2; exit 1; }
[...]

Issue #1586, opened 2026-08-30
p2pool's entrypoint echoes its full argv, so the audit line #273 added is also the origin of two redaction defects


VijitSingh97:
`build/p2pool/entrypoint.sh:92` logs the final launch command:

```
echo "[p2pool-entrypoint] launching: p2pool $* ${P2POOL_FLAGS:-}"
```

The reason it exists is good and should not be lost. Per its own comment (#273) it makes the
applied flags auditable — notably the #165 `--socks5` Tor routing — so a stale image silently
dropping `P2POOL_FLAGS` shows up in `docker logs p2pool` and `pithead doctor` fails on it
"rather than leaking quietly". That is a privacy control, and deleting the line would remove
it.

The problem is that it prints flag names and flag **values** together, and several of those
values are secrets: the RPC login, the Monero address, the Tari address, and the service
onion. Everything downstream then has to un-print them:

- #1582 — the integration harness's `redact()` missed the credential and the addresses on this
  line, and shipped them into artifacts the release gate uploads. Fixed.
- #1585 — `support-bundle`'s scrubber catches the credential on this line and leaves the
  addresses and the onion. Open.

Two redactors, drifting independently, both cleaning up after one `echo`. That is the shape
worth fixing at the source rather than a third time downstream.

## What the audit actually needs

#273's requirement is that the applied flag **names** are visible — that `--socks5` is present
and `P2POOL_FLAGS` was not dropped. It does not need the values. So the fix is to mask values
while keeping names, not to remove the line:

```
[p2pool-entrypoint] launching: p2pool --host … --rpc-login [redacted] --wallet [redacted] \
  --merge-mine tari://… [redacted] --onion-address [redacted] --socks5 … --local-api …
```

`--socks5` and its target stay legible, which is the property #165/#273 turn on, and nothing
downstream has to know the argv's secret positions. Worth checking against `pithead doctor`'s
parser before changing the format — if doctor greps this line, its expectations bind.

## Note the positional

The Tari address is a **positional** argument to `--merge-mine`, not a flagged one. Any
name-keyed scrubber will miss it by construction — which is exactly how #1585 happened. A fix
at the source does not have this problem, because the entrypoint knows which argument is which.

## What I did not do

- Not my lane (`build/`) — filed, not fixed.
- I did not check whether `pithead doctor` parses this line, and it is the one thing that could
  make a format change break something. Someone taking this should check first.
[...]

PR #1663, opened 2026-09-03
fix(p2pool): #1586 mask the secret VALUES on the entrypoint audit line, keep every flag NAME

VijitSingh97:
**#1586 is closed by hand with a comment naming this PR** — `Closes` cannot fire here, because PRs
merge to `develop-v2` while the default branch is `develop`.

## What this changes

`build/p2pool/entrypoint.sh` echoed the full argv, so the #273 audit line printed the RPC login, both
wallet addresses and the service onion — into `docker logs p2pool`, where it stays for the
container's life. It is the upstream **source** of #1582 and #1585: two redactors that grew up
cleaning after this one line.

The fix masks the **values** and keeps every flag **name**, so #273's audit requirement survives
intact — `--socks5` and its target stay legible, and so does the `tari://` routing URL.

`_redact_argv` is **positional, and it has to be**: the Tari address is a bare argument after
`--merge-mine <url>`, so a name-keyed scrubber misses it by construction — which is exactly how
#1585 happened. The entrypoint is the one place that knows which argument is which. A token starting
with `-` is passed through even in a value position, so a malformed argv cannot swallow the next flag
**name**; a flag silently missing from this line is the one failure the line exists to prevent.

`P2POOL_FLAGS` is now folded in with `set -- "$@" ${P2POOL_FLAGS:-}` **before** both the echo and the
`exec`, so the line printed is provably the argv exec'd — one word-split, read twice — instead of a
space-joined `$*` that only resembles it.

`pithead doctor` does **not** parse this line — `06-doctor.sh:178` reads `/proc/1/cmdline` via
`docker exec` — which was #1586's one named risk, re-derived at source rather than assumed.

## The over-engineering pass changed the patch

Reviewing my own diff for over-engineering turned up the opposite: an **under**-covered arm, and a
real one. The equals spellings were handled for `--rpc-login=`, `--wallet=` and `--onion-address=`
but **not** for `--merge-mine=`, and merge-mine is the one flag whose secret is the *next* token.
Measured before touching anything:  (the stack's own compose-default private IPs are
elided as `<tari-ip>` / `<tor-ip>` throughout this body, per the topology rule — nothing else in
any quoted output is altered.)

```
$ ... entrypoint.sh --wallet=4MoneroPayout --merge-mine=tari://<tari-ip>:18142 12TariPayout
p2pool --wallet=[redacted] --merge-mine=tari://<tari-ip>:18142 12TariPayout
                                                               ^^^^^^^^^^^^ the bare address, raw
```

This is the adjacent-pair test: both spellings can hold the same thing, so the asymmetry was a
defect, not a deliberate split. The fix is **line-neutral** — `--merge-mine=*` joins the existing
"mask the next token" arm, whose semantics it already wanted — and the equals-form test row was
widened to cover it rather than a new row added. Not reachable from our own
`docker-compose.yml` (which uses the separated spelling), and shipped anyway because the cost was one
pattern token.

Declined, and why: **"print flag names only, drop every value"** would be simpler machinery, but it
also drops `--stratum 0.0.0.0:3333`, the `--socks5` target and the `tari://` URL — the non-secret
values that make the line worth auditing at all. The masking keeps #273's line useful.

## What was RUN

**Full `tests/stack/run.sh`, clean tree:** `3065 passed, 0 failed` (rc 0, zero `✗` marks in the log)

**Red-proof.** The real text of `tests/stack/test-monero-tari.sh:43-93` was run on top of the real
`tests/stack/lib.sh` — first unmutated (control: the driver can say PASS), then with
`build/p2pool/entrypoint.sh` reverted to `origin/develop-v2`. Every leg `cmp`-checked the file
actually changed before being counted.

```
LEG 1 — head entrypoint:      PASS=14 FAIL=0
LEG 2 — pre-fix entrypoint:   PASS=12 FAIL=2

  ✗ launch line masks all four secrets, keeps every flag name and --socks5
      got [p2pool --host 127.0.0.1 --rpc-login nodeuser:nodepass --wallet 4MoneroPayout
           --merge-mine tari://127.0.0.1:18142 12TariPayout --onion-address examplep2pool.onion
           --stratum 0.0.0.0:3333 --mini --socks5 <tor-ip>:9050 --socks5-proxy-type tor]
  ✗ --flag=VALUE masked, and --merge-mine= still masks the address AFTER it
      got [p2pool --wallet=4MoneroPayout --merge-mine=tari://<tari-ip>:18142 12TariPayout ]
```

That `got` line **is** the defect, printed.

**Per-arm attribution — each masked arm killed ALONE reds the row, so the green is not borrowed:**

```
BASELINE (no mutation)         PASS=14 FAIL=0
kill --rpc-login arm           PASS=13 FAIL=1
kill --wallet arm              PASS=13 FAIL=1
kill --onion-address arm       PASS=13 FAIL=1
kill --merge-mine position     PASS=13 FAIL=1   (next=mmurl -> next=none)
kill --merge-mine= arm         PASS=13 FAIL=1
```

One of these legs first ran with a malformed `sed` that matched nothing; the `cmp`-before-count guard
refused to score it, and it was re-run with a working mutation. Reported rather than quietly fixed.

The third row — `p2pool is still exec'd with the RAW bare tari address` — is **green on both legs by
design**: it is the control that the fix did not change what p2pool actually receives, so a mutation
of the *log line* must leave it alone.

**Gates:** `make lint-file-budget` green (**582/582** at head, see below); `shfmt -i 4 -d` clean on
both changed files; `shellcheck --severity=warning` rc 0 over
`build/*/*.sh tests/stack/lib.sh tests/stack/test-*.sh tests/stack/test_*.sh os/overlay/pithead-boot`
plus `tests/stack/run.sh`, with a positive control (an injected `cd` with no `|| exit`) confirmed
firing SC2164 first. `pithead-boot` is paired in deliberately: without it the partial glob invents
three SC2034s in `test-appliance-boot.sh`, a file this diff never touches — the pairing the Makefile
already documents.

## The line budget — pay-for-lines, as ruled

`tests/stack/test-monero-tari.sh` was at **582/582 with zero headroom**. Per the controller's ruling
the block pays for itself: **8 lines added, 8 lines deleted**, every deletion a comment that restates
the code beneath it (or dead code), **no assertion touched**, and the file back at exactly 582. This
is the file's first shave.

Deleted, verbatim:

```
_a94="$(printf 'a%.0s' $(seq 94))"
# Default (clearnet absent/false) routes outbound sidechain dials through the bundled Tor SOCKS proxy.
# One flipped character in each form must fail as "checksum", not pass — the whole point.
# The wallet-rpc password stays off the .env command line too — it's in .env but never on stdout.
# (6c) A non-numeric tari.remote.grpc_port is rejected.
# Second apply must not rotate the now-populated creds.
# And the defaults apply when omitted (local node).
# `logs` forwards its service argument to `docker compose logs -f` (read-only follow).
```

`_a94` is not a comment: it is **dead code** — defined and never referenced anywhere in the file
(`_a93` and `_h94` beside it are both used). `(6c)` is the last label in its `6a/6b/6c` group, so
removing it leaves no gap in the surviving numbering.

`build/p2pool/entrypoint.sh` is 144 lines with no budget row (under `TARGET_LINES=400`), so it needed
no payment.

## Lane and grants

- `build/p2pool/entrypoint.sh` — **GRANT 2** (`roles/e2e.paths`, #1586 only).
- `tests/stack/test-monero-tari.sh` — **GRANT 2b**, armed by the controller at w79 on this lane's
  ask: one additive `echo "== ... =="` block beside the stub p2pool/socat harness already at
  `:44-86`, no edit to existing rows, no `tests/stack/run.sh` stanza (the file is already
  registered). Tier 1 is the right home per `docs/dev/testing-strategy.md` — a `selftest-*.sh` under
  `tests/integration/` is harness self-test and is not one of the four tiers.
- Both grants expire when this PR merges. No shared file touched, no `.lane-override` used.

## What I did NOT do

- **No doc change, and I checked rather than assumed.** Grepped `docs/` for the defect's own words:
  nothing documents the launch line or #273. The two nearest claims both survive unchanged —
  `docs/operations.md:28` describes what `support-bundle` scrubs (still true; this makes that
  scrubbing redundant for *future* logs, not wrong), and `:657` points at the merge-mine **bridge**
  line, which this does not touch.
- **Old logs are not retroactively cleaned.** Containers running from before this ships keep the
  unmasked line for their life; `support-bundle`'s redactor (#1585) is what covers those, and stays.
- **The masked set is `--rpc-login`, `--wallet`, `--onion-address` and the `--merge-mine` positional,
  in both spellings.** Any future secret-bearing flag needs a line here — the comment above
  `_redact_argv` says so, rather than implying the list is closed.
- **Filed while here:** #1661 — `tests/stack/lib.sh:81` falls back to the caller's cwd when
  `mktemp -d` fails, and line 82 then arms `rm -rf` on it. Static reading plus three isolated legs;
  the defect path was never executed.

https://claude.ai/code/session_016PaSXfW6JRNQzKgm7xMyJc

### DIFF
--- a/build/p2pool/entrypoint.sh
+++ b/build/p2pool/entrypoint.sh
@@ -85,10 +85,60 @@ if printf '%s' "${P2POOL_FLAGS:-}" | grep -q -- '--socks5'; then
     esac
 fi
 
+# Mask the secret VALUES on the launch line while keeping every flag NAME (#1586). #273 needs the
+# applied flags to be VISIBLE — notably the #165 `--socks5` — and never needed their values, but the
+# line printed both: the RPC login, both wallet addresses and the service onion. Two redactors
+# downstream grew up cleaning after it (#1582 in the test harness, #1585 in `support-bundle`), which
+# is the shape worth fixing at the source rather than a third time.
+#
+# Positional, and it has to be: the Tari address is a BARE argument after `--merge-mine <url>` (in
+# EITHER spelling), so a name-keyed scrubber misses it by construction — how #1585 happened. Here the
+# entrypoint knows which argument is which, so nothing has to guess. A token starting with `-` is
+# passed through even in a value position: on a malformed argv that would otherwise swallow the next
+# flag NAME, and a flag silently missing from this line is the one failure it exists to prevent.
+_redact_argv() {
+    local a out="" next=none
+    for a in "$@"; do
+        case "$next" in
+        mask)
+            next=none
+            case "$a" in -*) : ;; *)
+                out="$out [redacted]"
+                continue
+                ;;
+            esac
+            ;;
+        mmurl) # the tari://host:port URL is routing, not a secret; the address AFTER it is
+            out="$out $a"
+            next=mask
+            continue
+            ;;
+        esac
+        case "$a" in
+        --rpc-login | --wallet | --onion-address | --merge-mine=*)
+            out="$out $a"
+            next=mask
+            ;;
+        --rpc-login=* | --wallet=* | --onion-address=*) out="$out ${a%%=*}=[redacted]" ;;
+        --merge-mine)
+            out="$out $a"
+            next=mmurl
+            ;;
+        *) out="$out $a" ;;
+        esac
+    done
+    printf '%s' "${out# }"
+}
+
 # Log the FINAL launch command (#273): makes the applied flags — notably the #165 `--socks5` Tor
-# routing — auditable in `docker logs p2pool`, so a stale image silently dropping P2POOL_FLAGS shows
-# up here (and `pithead doctor` fails on it) rather than leaking quietly. After the #278 block above so
-# it reflects the rewritten --host.
-echo "[p2pool-entrypoint] launching: p2pool $* ${P2POOL_FLAGS:-}"
+# routing — auditable in `docker logs p2pool`, so a stale image silently dropping P2POOL_FLAGS is
+# visible here rather than leaking quietly. (`pithead doctor` catches the same staleness on its own,
+# by reading the container's /proc/1/cmdline — `06-doctor.sh`; this line is the human-readable trail,
+# not doctor's input.) After the #278 block above so it reflects the rewritten --host.
+#
+# Fold P2POOL_FLAGS in BEFORE both, so the line printed is provably the argv exec'd — one word-split,
+# read twice — instead of a space-joined `$*` that only resembles it.
 # shellcheck disable=SC2086  # intentional word-splitting of the space-separated flag string
-exec p2pool "$@" ${P2POOL_FLAGS:-}
+set -- "$@" ${P2POOL_FLAGS:-}
+echo "[p2pool-entrypoint] launching: p2pool $(_redact_argv "$@")"
+exec p2pool "$@"
--- a/tests/stack/test-monero-tari.sh
+++ b/tests/stack/test-monero-tari.sh
@@ -32,7 +32,6 @@ build_val_sandbox
 DOCKER_LOG="$V/docker.log"
 
 echo "== unit: p2pool_outbound_flags — Tor-by-default for outbound P2P (#165) =="
-# Default (clearnet absent/false) routes outbound sidechain dials through the bundled Tor SOCKS proxy.
 assert_eq "default → Tor SOCKS flags" "$(run_sourced "$SANDBOX" p2pool_outbound_flags false 172.28.0)" "--socks5 172.28.0.25:9050 --socks5-proxy-type tor"
 assert_eq "empty arg → Tor (default off)" "$(run_sourced "$SANDBOX" p2pool_outbound_flags '' 172.28.0)" "--socks5 172.28.0.25:9050 --socks5-proxy-type tor"
 # clearnet opt-out → no SOCKS flags (p2pool dials peers directly, IP exposed).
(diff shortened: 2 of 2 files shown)


VijitSingh97:
## Verdict: MERGE (controller, non-author pass, judged on the fable tier)

Reviewed at head `ab5b7b98`, merge-base == `origin/develop-v2` tip, so the branch tree IS the merged tree and one gate run answers both.

- **The mask works and the argv does not change.** Ran the head entrypoint against a stub `p2pool` that prints its argv, with the PR's own secret-bearing command line. Launch line: every one of `--rpc-login`, `--wallet`, the bare Tari address after `--merge-mine <url>`, and `--onion-address` printed as `[redacted]`; every flag name, the `--socks5` target, the `tari://` URL and `--stratum` kept. The stub received all 18 raw tokens in order.
- **Negative control fired.** The same run against the pre-fix entrypoint at the merge-base printed all four secrets raw on the launch line.
- **`pithead doctor` does not read this line.** `lib/pithead/06-doctor.sh` reads `/proc/1/cmdline` via `docker exec`, so the format change cannot break it. The only other consumer of the `launching:` text is `tests/integration/selftest-redact.sh:64`, which seeds its own fixture and stays valid for pre-fix logs.
- **Pay-for-lines paid as ruled.** File at 582/582 at both base and head; `docs/dev/file-budget.tsv` untouched (zero rows in the diff). All 8 deletions checked: 7 are comments restating the code beneath them, `_a94` has zero references at base or head (`_a93`/`_h94` beside it are used). No assertion touched. `(6a)`/`(6b)` labels survive; `(6c)` was the last of its group.
- **Gates at head:** `make lint-file-budget` rc 0 (self-test included); `shfmt -i 4 -d` rc 0 on both files.

**Probed and accepted:** on a malformed argv where `--merge-mine` is followed directly by a flag (`--merge-mine --stratum 0.0.0.0:3333`), the `mmurl` arm prints `--stratum` as the URL and masks `0.0.0.0:3333`. Every flag NAME still prints, which is the property #273 turns on; the over-mask is on a non-secret token and only on input p2pool itself would reject. Not a finding.

**Relayed, not re-run:** the red-proof and per-arm mutation batteries (the PR body's counts); shellcheck. The suite result is the PR's own gating figure: the PR body says `3065 passed, 0 failed`, rc 0; my own watcher read the lane's `suite2.rc` = 0 with 3065 ✓ and 0 ✗ in the log, independently of the body. Merge waits only on the CI checks completing.

**Disposition of the grants:** GRANT 2 (`build/p2pool/entrypoint.sh`) and GRANT 2b (`tests/stack/test-monero-tari.sh`) in `roles/e2e.paths` expire on this merge; I record the EXPIRED blocks in the fleet repo.

**Closing:** #1586 closes by hand naming this PR (`Closes` does not fire off the default branch).

VijitSingh97:
**Addendum to the verdict (controller):** the e2e lane disclosed that a duplicate `tests/stack/run.sh` overlapped the surviving run for about two minutes in the same worktree before it was killed by process group. Judged: the verdict stands. Each run sandboxes under its own `mktemp -d`, neither mutates the tree, and the fleet's measured concurrency effect in that suite is false REDS from fork/exec contention, never false greens. The surviving run wrote `suite2.rc` = 0 with 3065 ✓ and 0 ✗, which is the complete pass count, not a truncated one. #1586 closes at merge time, not before, as the lane's issue comment says.

Issue #1652, opened 2026-09-03
GA gate, unfiled: the 7-day unattended soak has no owner and no pass condition — and the bench's known-unhealthy container would fail any obvious one


VijitSingh97:
## Why this is an issue and not a checklist line

#394's release checklist has three unticked gates that are **runs, not code**: the KVM battery green
per cut, this soak, and the release lane exercised end to end. Because none of them is an issue, none
has a milestone, an owner, or a definition of passing — so a milestone-keyed queue cannot see them and
the work reads as done when the issue list empties. This files the soak half.

`docs/dev/dual-distribution-plan.md:296-300` is the authority and it is unambiguous:

> "Reliably" is measured, not felt: a 7-day unattended soak on a bench box with zero manual
> interventions, plus the fault-injection rollback demo below. That is the phase 2 exit bar.

**So GA is at least seven days out from whenever this starts.** That is the fact worth putting in
front of anyone planning the cut: if the soak runs after the queue drains, the ship date is
queue-time plus seven days; if it runs alongside, it is the larger of the two.

## What the bench already shows — measured today, and it is good news with a catch

Read at source on the appliance bench this morning:

- Uptime **11 days 8 hours**, single boot, no reboot in the window.
- All eight stack containers started within 33 seconds of that boot and **not one has restarted
  since**. Restart counts are the sharpest available proxy for "nothing intervened", and they are flat.
- RAUC slot B, `boot status: good`, `status=ok`, bundle installed `2026-08-16T17:22:46Z`, install
  count 8.

**The hardware and the stack sustain an uninterrupted 11-day run.** That retires most of the risk this
gate exists to find, and it should lower anyone's estimate of how likely the soak is to fail.

**It does not discharge the gate, for three reasons, and the third is the one that needs a decision.**

1. **Wrong artifact.** The running bundle was installed on 2026-08-16. A soak proves the image you are
   about to ship, not a predecessor of it. The ship candidate does not exist yet — #1649's CVE fix is
   not cut.
2. **Unattended is unproven, not disproven.** `last` returns no records, which is consistent with an
   untouched box but is not evidence of one. There is no soak log, no start marker and no daily
   summary anywhere on the box. Nothing was recorded because nothing was running as a soak.
3. **⛔ There is no pass condition, and today the box would fail any obvious one.** `xmrig-proxy` has
   been **unhealthy for the entire 11 days**, with a failed podman healthcheck unit to match. That is
[...]

Issue #1659, opened 2026-09-03
appliance: journald keeps the boot-time transient machine-id, so `journalctl` reads a stale journal for the whole boot (#967 closed only half)


VijitSingh97:
## What the bench box shows

Measured read-only on the appliance bench (box up since 2026-08-22 17:48Z by `/proc/stat` btime; image built 2026-08-16, and the #967 fix line `mount -t overlay overlay` IS present in its `/usr/local/sbin/pithead-machine-id`):

- `journalctl -b` prints `-- No entries --`. `journalctl --list-boots` lists exactly one boot, 2026-08-14 22:17 to 2026-08-15 18:28. `journalctl -n 1` returns an xmrig line from 2026-08-15.
- `/etc/machine-id` equals `/data/pithead/machine-id` (`eea40aa1…`): the restore worked. `pithead-machine-id.service` Result=success, ExecMainStatus=0, at 17:48:26; `/etc` is the overlay the fix mounts; `/etc/machine-id` is a mountpoint.
- journald's open file descriptors: `/var/log/journal/ce0cb6cb…/system.journal` — a directory named for a DIFFERENT id — and the newest entry in it is now. `journalctl -m` reads it.
- Control that the instrument discriminates: `journalctl -m -u ssh --since '2 hours ago'` counts 19 `Accepted` (my own sessions); the same without `-m` counts 0.
- `/var/log/journal` holds 15 such directories, one per boot, each named for a transient id.

## Mechanism

Inferred from the readings and the unit's own header, not stepped through: the image ships `/etc/machine-id` empty, so PID 1 mints a transient id on every boot and journald starts on it. `pithead-machine-id.service` is deliberately NOT `Before=systemd-journald.service` (the header says why: journald starts long before `/data` can mount) and orders before `systemd-journal-flush.service` only — which names the directory the RUNTIME journal is flushed into. But journald caches the machine-id it read at startup and keeps writing under it for the whole boot. The flush and the restore fix what `journalctl` LOOKS FOR; nothing fixes what journald WRITES TO. #967 closed the first half; this is the second.

Why the KVM legs pass (inferred): on a first boot the adopted id IS the boot id, so there is no mismatch. The mismatch starts on the second boot of any machine that already has a persisted id — every boot of a provisioned appliance.

## Cost

[...]

PR #1667, opened 2026-09-03
tests(os): #1652 the soak's daily probe — one fixed read-only SSH read, one line a day, scored against the ruled pass condition

VijitSingh97:
Part of #1652 (the instrument; the window itself opens when the operator's item 20 decision lands). Closing keywords are inert on `develop-v2`.

## What this is

The 7-day unattended soak had no instrument: nothing on or near the bench records a day, and the ruling on #1652 requires the pass condition to be *measured*, written before the run, with a missing measurement scored as a FAIL. `tests/os/soak-probe.sh HOST LOGDIR [--start]` is that instrument, and it deliberately lives in the tree so the one login the window permits is auditable line by line.

- **One non-interactive, read-only SSH session per day**, whose remote command is fixed in the script. It reads `.env` for monerod's RPC credentials and never prints them.
- **One line per day** appended to `LOGDIR/soak.log`, plus `day0.env` (the baseline — written by `--start` and by NOTHING else) and `readN.env` (each read's raw readings, N its `soak.log` line number; every line carries `read=N`, since the first cron read lands under 24 h after `--start` and shares `day=0` with the baseline line), all outside git.
- **Scored against the ruled rules**, each with its instrument named in the header:
  1. one boot — `/proc/stat` btime equals day 0's; journal-directory count as the secondary counter;
  2. restarts flat — every container's `RestartCount` **and** `StartedAt` equal day 0's (a supervisor restart moves the first; a hand `podman stop/start` moves only the second);
  3. all running — every day-0 container `running`; health asserted for all **except `xmrig-proxy`, excluded by name** (#1098), which must still be running with rules 1–2 holding;
  4. no intervention — EXACTLY one `Accepted` in `journalctl -m -u ssh` since the journal cursor the previous read recorded (the probe's own; 0 fails naming the instrument, `ssh-journal-blind`; no cursor falls back to the last 25 h and says `window=25h`), and `last` empty;
  5. chain recorded, never gating — monerod height / `synchronized` / peers on every line.
- **The verdict is a pure function over the readings** (`soak_day_verdict`), with a `--self-test` of 29 rows — 20 over the pure verdict and 9 over the DRIVER through a stubbed `ssh` (added for F1 of the second pass: the pure rows cannot see what the driver does with `LOGDIR`): an identical day passes; each rule fails on its own perturbation; a hand stop/start is caught by `StartedAt` alone; xmrig-proxy exited still fails; a day-0 container missing from today fails; **an empty reading fails every rule** rather than passing vacuously.

## What was RUN

| check | result |
|---|---|
| `--self-test` | 29 ok, 0 failed |
| **F1 (second pass) mutation controls** — restore the `day$day.env` write in a scratch copy; drop the `read=` label in another | 4 driver rows red (the third run prints `VERDICT=PASS`, the reported defect verbatim); 1 row red |
| `shellcheck --severity=warning`, `shfmt -i 4 -d`, budget gate, docs voice | rc 0 |
| **dry run against the bench, read-only, three times** (`--start` into a scratch dir, then a day-N read) | 8/8 running, `xmrig-proxy` unhealthy exactly as #1098 says, `RestartCount` 0 across the board, RAUC slot B, monerod synchronized at a live height |
| **rule 4 fired on the live box**: 22, then 23, then 24 `Accepted` in the last 25 h — the pre-flight reads of the last day, mine included | `VERDICT=FAIL fails=4:ssh-accepted(24)` — the instrument shown able to fail before the window opens |
| stderr carries no address (ssh's known-hosts warning suppressed with `LogLevel=ERROR`) | verified by substitution over stdout+stderr |

## Review record

- **First pass F1/F2** (cursor pass-through, rule-4 four-way case) — fixed at `3b45ab37`.
- **Second pass F1, blocking** — the first cron read computed `day=0` and overwrote `day0.env`, so a restart in the window's first hours showed on one discounted line and every later day passed against the moved baseline. Fixed at the current head: `day0.env` is written by `--start` only, raw readings go to `readN.env`, every line carries `read=N`, and the self-test drives the script three times through a stubbed `ssh` (all landing in `day=0`) asserting the baseline is byte-identical after a cron read and that the same restarted reading still fails on the third run.

## Two things the dry run found, stated rather than smoothed

- **`last` is a null instrument on this box.** It read 0 through all 24 logins (no wtmp). It is recorded because the ruling names it and the header says it cannot fire; the journal count is the rule-4 gate.
- **Rule 2's `StartedAt` already tells a story on the bench today**: six containers started 02:13Z (the pre-flash `pithead backup` cycles the stack) and `tor`+`monerod` at 03:16Z with `RestartCount` still 0 — a hand restart, not a supervisor one, which is exactly the case `RestartCount` alone would miss. Whatever it was, the window's day-0 baseline is taken *after* the flash, so it does not bleed in. Also on that line: monerod `peers:0/0` while `synchronized:true` — recorded, not judged; the probe never gates on the chain.

## Not done, named

- **Tari's height is not read** — the box has no gRPC client; its container state and health are. Said in the header.
- **No tier-1 case in `tests/stack/`** — the verdict's fixture proof is the script's own `--self-test` (the `failure-evidence.sh --self-test` shape), which needs no grant and runs anywhere.
- **The window is not open.** Nothing on the bench was written; the two scratch `--start` runs went to a temporary directory on the build host. Day 0 is the first run after the flashed image is up, from a cron line, never from a resident session.

The ponytail gate reads the branch from this lane's pinned shell cwd (a different worktree on a dead branch) — disclosed; the pass was done against this diff.

- **A systemd timer on the box** instead of a cron line on the build host: rejected — it would put a writer on the box inside the window and make the probe part of the artifact under test.
- **`pithead doctor --json` as the daily read**: rejected — it is a verb on the box and takes tens of seconds of probes; `podman inspect` and one `get_info` answer the ruled rules directly and are cheaper to audit.
- **Dropping `last`**: available, rejected — the ruling names it; keeping a null instrument *labelled* null is better than a reader wondering why rule 4 has one leg.

### DIFF
--- a/tests/os/README.md
+++ b/tests/os/README.md
@@ -84,3 +84,13 @@ that the boot path's files sit where the firmware and GRUB will look.
 sudo tests/os/verify-image.sh os/rauc/build/system.img          # release: test artifacts REFUSED
 sudo tests/os/verify-image.sh os/rauc/build/system.img --test   # harness build: SSH key expected
 ```
+
+## Soak probe
+
+`tests/os/soak-probe.sh HOST LOGDIR [--start]` is the 7-day unattended soak's daily reader
+(#1652): one non-interactive, read-only SSH session whose remote command is fixed in the script,
+one line per day appended to `LOGDIR/soak.log`, scored against the pass condition ruled on the
+issue (one boot, restart counts and start times flat, every container running and — except
+`xmrig-proxy`, #1098 — healthy, at most the probe's own login). `--start` writes the day-0
+baseline; `--self-test` proves the verdict over canned readings without a box. Run it from a
+cron line on the build host, never from a resident session.
--- a/tests/os/README.md
+++ b/tests/os/README.md
@@ -91,6 +91,10 @@ sudo tests/os/verify-image.sh os/rauc/build/system.img --test   # harness build:
 (#1652): one non-interactive, read-only SSH session whose remote command is fixed in the script,
 one line per day appended to `LOGDIR/soak.log`, scored against the pass condition ruled on the
 issue (one boot, restart counts and start times flat, every container running and — except
-`xmrig-proxy`, #1098 — healthy, at most the probe's own login). `--start` writes the day-0
-baseline; `--self-test` proves the verdict over canned readings without a box. Run it from a
-cron line on the build host, never from a resident session.
+`xmrig-proxy`, #1098 — healthy, exactly the probe's own login since the previous read). Each
+read records the journal cursor it reached in `LOGDIR/ssh.cursor`, and the next read counts logins
+from there, so nothing falls between two days and nothing is counted twice; a count of 0 fails
+naming the instrument, since the probe's own login must be there. `--start` writes the day-0
+baseline — its own line reads `window=25h` and carries a rule-4 FAIL from the setup logins, so day
+0 is the baseline, not a soak day. `--self-test` proves the verdict over canned readings without a
+box. Run it from a cron line on the build host, never from a resident session.
--- a/tests/os/README.md
+++ b/tests/os/README.md
@@ -96,5 +96,10 @@ read records the journal cursor it reached in `LOGDIR/ssh.cursor`, and the next 
 from there, so nothing falls between two days and nothing is counted twice; a count of 0 fails
 naming the instrument, since the probe's own login must be there. `--start` writes the day-0
 baseline — its own line reads `window=25h` and carries a rule-4 FAIL from the setup logins, so day
-0 is the baseline, not a soak day. `--self-test` proves the verdict over canned readings without a
-box. Run it from a cron line on the build host, never from a resident session.
+0 is the baseline, not a soak day. Only `--start` writes `day0.env`; a cron read never does, so a
+restart in the window's first hours can never be absorbed into the baseline it is scored against.
+Every line carries `read=N`, its `soak.log` line number, because the first cron read lands under
+24 h after `--start` and shares `day=0` with the baseline line; each run's raw readings are kept as
+`LOGDIR/readN.env`. `--self-test` proves the verdict over canned readings, then drives the script
+three times through a stubbed `ssh` to prove the baseline survives a cron read, without a box. Run
+it from a cron line on the build host, never from a resident session.
--- a/tests/os/soak-probe.sh
+++ b/tests/os/soak-probe.sh
@@ -20,16 +20,41 @@
 #                       moves only the second.
 #   3 all running     — every day-0 container is `running`; health asserted for all EXCEPT
 #                       xmrig-proxy, whose healthcheck is the product defect #1098 (excluded BY
-#                       NAME; it must still be running with rules 1-2 holding).
-#   4 no intervention — `journalctl -m -u ssh` counts at most ONE `Accepted` in the last 25 h
-#                       (this probe's own login) and `last` shows no interactive session; rule 2's
-#                       StartedAt covers hand-run container verbs. `last` is recorded because the
-#                       ruling names it, and it is a NULL instrument on the bench (no wtmp: it read
-#                       0 with 22 logins in the journal) — the journal count is the one that fires.
+#                       NAME; it must still be running with rules 1-2 holding). The set is DAY 0's:
+#                       a container that appears later is outside rule 3 by decision — rule 4
+#                       covers the only route by which one could be started.
+#   4 no intervention — `journalctl -m -u ssh` counts EXACTLY ONE `Accepted` since the previous
+#                       probe's read: this probe's own login. The count starts at the journal
+#                       CURSOR the previous read recorded (LOGDIR/ssh.cursor, passed to the remote
+#                       command as its one named input), so there is no window edge to straddle
+#                       and no clock on either side to trust. With no cursor (day 0, or the file
+#                       missing) it falls back to the last 25 h and the line says `window=25h`, so
+#                       day 0's line carries a rule-4 FAIL from the setup logins: it is the
+#                       baseline, not a soak day. A count of 0 FAILS naming the instrument
+#                       (`ssh-journal-blind`): the probe's own login is a positive control the
+#                       reading must hold, so 0 means the journal could not be read (cursor
+#                       journal split, journald down) — never a quiet day. A cursor names one
+#                       entry in one journal file, and a cursor that no longer resolves (journal
+#                       vacuumed or reset, box re-flashed) is loud either way, measured both ways
+#                       on 2026-09-03: the build host's journalctl refuses the seek (count 0,
+#                       `ssh-journal-blind`), the box's seeks to the START and counts every login
+#                       on record (457) — a day that fails rule 1 as well, which is the right
+#                       answer for it. The read then records a fresh cursor, so the next day
+#                       scores normally. journald
+#                       writes asynchronously, so a 0 or a 2 on a live box is a rare LOUD flake
+#                       to look at, not a silent pass. `last` shows no interactive session; rule
+#                       2's StartedAt covers hand-run container verbs. `last` is recorded because
+#                       the ruling names it, and it is a NULL instrument on the bench (no wtmp: it
+#                       read 0 with 22 logins in the journal) — the journal count is the one that
+#                       fires.
 #   5 chain recorded  — monerod height / synchronized / peers are RECORDED (a stall is visible)
 #                       and never gate. Tari is recorded by container state only: the box has no
 #                       gRPC client, so its height is not read — stated, not skipped silently.
-# A day whose line is missing, or whose SSH read failed, is a FAIL by the ruling's own terms.
+# A day whose line is missing, or whose SSH read failed, is a FAIL by the ruling's own terms; a
+# failed read keeps the previous cursor, so the next good day also counts the failed day's login.
+# The session skips host-key checking on purpose: the box regenerates its host key on every
+# provisioning (#1659's churn) and the probe is read-only on a bench LAN, so a pinned key would
+# only turn each re-flash into a READ-FAILED day.
 set -uo pipefail
 
 KEY="${PITHEAD_SOAK_KEY:-$HOME/.ssh/pithead-os-test}"
@@ -37,9 +62,9 @@ EXCLUDED_HEALTH="xmrig-proxy"
 
 # The one remote command. Read-only by construction: every line is a read, and the .env is
 # consulted for monerod's RPC credentials without ever printing them.
-read_box() { # $1 = host; prints key=value lines
+read_box() { # $1 = host, $2 = previous read's journal cursor or empty; prints key=value lines
     timeout 120 ssh -i "$KEY" -o BatchMode=yes -o ConnectTimeout=15 -o StrictHostKeyChecking=no \
-        -o UserKnownHostsFile=/dev/null -o LogLevel=ERROR "root@$1" 'bash -s' <<'REMOTE'
+        -o UserKnownHostsFile=/dev/null -o LogLevel=ERROR "root@$1" "SOAK_CURSOR='$2' bash -s" <<'REMOTE'
 set -u
 printf 'btime=%s\n' "$(awk '/^btime/{print $2}' /proc/stat)"
 printf 'jdirs=%s\n' "$(ls /var/log/journal 2>/dev/null | wc -l)"
@@ -50,7 +75,14 @@ printf 'rauc=%s\n' "$(rauc status --output-format=shell 2>/dev/null | awk -F= '/
 for c in $(podman ps -aq 2>/dev/null); do
     podman inspect -f 'container={{.Name}}|{{.State.Status}}|{{.RestartCount}}|{{if .State.Health}}{{.State.Health.Status}}{{else}}none{{end}}|{{.State.StartedAt}}' "$c" 2>/dev/null
 done
-printf 'ssh_accepted_25h=%s\n' "$(journalctl -m -u ssh --since '-25h' --no-pager -q 2>/dev/null | grep -c 'Accepted ')"
+if [ -n "${SOAK_CURSOR:-}" ]; then
+    printf 'ssh_window=cursor\n'
+    printf 'ssh_accepted=%s\n' "$(journalctl -m -u ssh --after-cursor="$SOAK_CURSOR" --no-pager -q 2>/dev/null | grep -c 'Accepted ')"
+else
+    printf 'ssh_window=25h\n'
+    printf 'ssh_accepted=%s\n' "$(journalctl -m -u ssh --since '-25h' --no-pager -q 2>/dev/null | grep -c 'Accepted ')"
+fi
+printf 'ssh_cursor=%s\n' "$(journalctl -m -u ssh -n1 --no-pager -q -o cat --show-cursor 2>/dev/null | sed -n 's/^-- cursor: //p')"
 printf 'last_sessions=%s\n' "$(last -F 2>/dev/null | grep -v -c -E '^(reboot|wtmp|$)')"
 env_get() { sed -n "s/^$1=//p" /data/pithead/.env 2>/dev/null | head -1 | tr -d '"'; }
 mu=$(env_get MONERO_NODE_USERNAME); mp=$(env_get MONERO_NODE_PASSWORD); murl=$(env_get MONERO_RPC_URL); [ -n "$murl" ] || murl=http://127.0.0.1:18081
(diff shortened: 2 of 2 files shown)


VijitSingh97:
## Verdict at head `bbf7137e`: RETURNED WITH FINDINGS — one blocking, do not merge yet (controller, non-author pass, judged on the fable tier)

Reviewed the full diff (`tests/os/soak-probe.sh` +189, `tests/os/README.md` +10), merge-base `e2b963f1`, three commits behind the tip. Ruling 8a both legs run against the current tip: budget gate and parity rc 0 at the branch head and at the merge-tree result; merge-tree rc 0; no file overlap with the three merges that moved the tip. Mechanical checks (self-test, `bash -n`, shellcheck, budget, docs voice, topology detector) were run by a review agent in a detached worktree; I re-derived the findings below by reading the script myself. Labels: PROVEN = I read or ran it; RELAYED = taken from the PR body or the lane's report.

### F1 — BLOCKING: rule 4 fails every clean day from day 1, on the probe's own previous login (PROVEN by reading; the self-test cannot see it)

`soak-probe.sh:53` counts `Accepted` lines in `journalctl -m -u ssh --since '-25h'`, and `:89` passes only when that count is at most 1 — "this probe's own login" (`:24-25`). The documented schedule is one cron run per day (`:12`, README). A 25-hour window read once every 24 hours always contains the previous day's probe login as well as today's, so on every normal day from day 1 the count is 2 and the line reads `VERDICT=FAIL fails=4:ssh-accepted(2)`. By the header's own terms (`:32`, a FAIL day fails the soak) the probe fails the soak it is measuring. The self-test cannot catch this because the window lives inside the remote command and the canned readings hand `soak_day_verdict` a count of 1. The three dry runs on the bench did not exercise it either: they ran on the same day, against a journal already holding the 22-24 pre-flight logins, so rule 4 fired for a different reason. Design constraint to respect in the fix: the remote command is meant to stay fixed and auditable (`:2-4`, `:38`).

Fix shape, the lane's choice: pass the previous run's timestamp into the read (the script already records `now` per run; `ssh ... "SINCE=<prev+1min> bash -s"` keeps the remote body fixed with one named input) and count from there; on day 0, or with no previous run, fall back to the 25 h window and say so in the log line. A shorter fixed window (`-23h`) also stops the false FAIL but leaves a one-hour blind gap per day, which the ruling's "at most the probe's own login" does not allow.

### F2 — SHOULD-FIX, same rule: "at most 1" cannot tell a dead journal from a quiet day (PROVEN by reading)

`:53` has `2>/dev/null` on `journalctl` and `grep -c` prints `0` when its input is empty, so a journal that cannot be read — the unit renamed, the journal split by #1659's machine-id churn, `journalctl` failing outright — scores `ssh_accepted_25h=0` and PASSES rule 4. The probe's own login is a positive control the reading already contains: with the window from F1, the honest test is exactly 1, and 0 is a FAIL that names the instrument (`4:ssh-journal-blind(0)`). Add the self-test row for it (`ssh_accepted_25h=0` must FAIL); today's rows cover 1 and 2 only. Cost to weigh, stated not judged: journald writes asynchronously, and the read happens a few seconds after the login, so a 0 on a live journal would be a rare flake that shows up as a loud FAIL to investigate rather than a silent pass; the header's rule about `last` (`:26-28`) already takes the same stance for the other instrument.

### Minor, not blocking

- `kv` is defined twice with different signatures: `:67` (two args, inside `soak_day_verdict`) and `:180` (one arg, top level). It works today only because every call of `soak_day_verdict` sits in a `$(...)` subshell (`:109-143`, `:184`), so the inner definition never reaches the parent. A future direct call would silently blank every field of the log line. Rename the inner helper.
- Rule 3 iterates day 0's containers (`:87`), so a container that appears after day 0 is invisible to the verdict; rule 4 covers the SSH route to it. Worth one line in the header so the blind spot is a decision, not an oversight.
- `StrictHostKeyChecking=no` with `UserKnownHostsFile=/dev/null` (`:41-42`) means the read is unauthenticated on the box's side; on the bench LAN with a read-only probe this is a note, not a finding, and I take it to be deliberate given #1659's host-key churn. Say so in the header.
- Day 0's own line will normally carry a rule-4 FAIL (the setup logins are inside the window), which the README does not say; one clause there ("day 0's line is the baseline, its rule-4 count includes the setup logins") stops the first reader from filing it.
- The header's claim that btime is stable for the boot holds under NTP slew; a wall-clock STEP moves it (kernel derives btime from the wall clock offset — from my knowledge of the kernel, not re-derived here). The log line records `up=` beside it, so a reader can tell a step from a reboot. Nothing to change.

### What is right, and the observation the lane recorded

The verdict function is pure over the readings, every unmeasurable rule FAILS rather than skips (`:143-144` proves the empty case), xmrig-proxy is excluded by name for health only and still must be running (`:131-133`), rule 2 separates a supervisor restart from a hand stop/start (`:84-85`), and rule 5 records monerod's height/sync/peers without gating, as ruled. The README paragraph is in the house voice and says who runs it and from where.

The lane's unjudged observation stands as recorded: tor and monerod on the bench show `StartedAt` 03:16Z with `RestartCount` 0 (a hand restart after the 02:13Z backup) and monerod reads `synchronized:true` with peers 0/0. Rule 2 judges relative to day 0, so those restarts do not touch the soak once `--start` is taken after them. Peers 0/0 is recorded, never gated, per the ruling; but a soak started on a node with no peers proves stability of a box that is not doing its job, and the height column would sit flat for seven days. Before `--start`: read why monerod has no peers (tor reachability, the restart itself) and fix or file it. That is a note for the soak's operator, not a defect of this PR.

### Mechanics (run by a sonnet review agent in a detached worktree at `bbf7137e`; RELAYED from its report, each item it marked as run itself)

- `bash tests/os/soak-probe.sh --self-test` → `14 ok, 0 failed`, rc 0; and shown able to FAIL: with `EXCLUDED_HEALTH` swapped to `monerod` it reported `11 ok, 3 failed`, rc 1, then restored and re-run green. The 14 rows exercise rules 1-4 each with at least one negative case; no row covers rule 5's "recorded, never gates" (a changed `monero=` reading asserted to still PASS) — add one so a later edit cannot start gating on it unnoticed.
- `bash -n` rc 0; `shellcheck --severity=warning` rc 0 with no output; `make lint-file-budget` OK; `make lint-docs-voice` OK across 41 prose docs; topology detector `clean` on both files and on the PR body. `shfmt` was not run by the agent (the body's rc 0 for it is the lane's, relayed).
- Wiring: `soak-probe` is referenced only by itself and the README, which matches the sibling `tests/os/*-verdict.sh` self-tests (none wired into the Makefile). Consistent with "a cron line on the build host".
- One more minor, from the agent and confirmed by my read of `:149-163`: a call with `HOST` but no `LOGDIR` falls through the `case` to `mkdir -p ""` and writes `day0.env` at `/`. A two-argument check before `:159` with the usage text closes it.
- The bench dry runs and their readings (8/8 running, rule 4 firing at 22-24, `last` reading 0) are RELAYED from the PR body; nothing in the diff carries a log, and no reviewer here logged into the bench.

Fix F1 and F2 on this branch, push, and I will re-review at the new head. I re-derived F1 and F2 and the minor items from the script; the mechanical results above are relayed from the agent's run except where marked.

VijitSingh97:
## F1 and F2 fixed at the new head — cursor window, exact-1 rule, blind-instrument FAIL; minors taken

**F1 (blocking) — the window.** The remote command no longer counts over a fixed 25 h. Each read records the journal cursor of the last `ssh` entry it saw (`LOGDIR/ssh.cursor`, taken with `--show-cursor` after the count), and the next read passes it back as the command's one named input (`SOAK_CURSOR='…' bash -s`, the body itself unchanged and still auditable) and counts `Accepted` with `--after-cursor`. So the count starts at the entry after the last one the previous probe saw: no window edge for yesterday's login to straddle, and no clock on either side to trust (the build host's and the box's clocks never enter it). With no cursor — day 0, or `--start`, which ignores any cursor a previous window left — it falls back to 25 h and the line says `window=25h`; the header and README now say that day 0's line carries a rule-4 FAIL from the setup logins and is the baseline, not a soak day. I took the cursor over the timestamp you suggested because the timestamp still leaves the minute after the previous login blind and depends on the two clocks agreeing; the cursor has neither gap.

**F2 (should-fix) — exactly 1, and 0 names the instrument.** Rule 4 is now a four-way case: unreadable → `4:ssh-accepted(?)`, 0 → `4:ssh-journal-blind(0)`, 1 → pass, more → `4:ssh-accepted(N)`. Self-test rows added for 0 (named), for the reading missing altogether (named), for 2 (now also named), and — your mechanics note — for rule 5: a changed `monero=` reading still passes. 20 rows, 0 failed.

**Your note on a cursor that stops resolving — stated in the header, from measurement rather than from what I expected.** I built the hazard both ways with a well-formed bogus cursor. The build host's journalctl refuses the seek (rc 1, count 0 → `ssh-journal-blind`). The box's does NOT refuse: it seeks to the start of the journal and counts every login on record — 457 — so that day FAILs loudly as `4:ssh-accepted(457)` rather than as blind. Both are loud, neither is a silent pass, and the read records a fresh cursor so the next day scores normally (proven below, run 5). The header says exactly that; I had first written "refuses the seek" from the build-host result alone, and the box refuted it.

**Minors, all taken:** inner `kv` renamed `rd` (a reading) so a direct call of the verdict can never blank the log line; `HOST` without `LOGDIR` now prints usage and exits 2 instead of writing `day0.env` at `/`; rule 3's day-0 set is stated as a decision in the header, with rule 4 named as the route that covers a later container; the host-key stance is stated in the header (read-only probe on the bench LAN, the box regenerates its host key on every provisioning); `--help` prints the header by its extent instead of a fixed line range, so the longer header still prints whole. The btime-under-a-step note: nothing changed, as you said.

## What was RUN at the new head

| check | result |
|---|---|
| `--self-test` | 20 ok, 0 failed (was 14) |
| `bash -n`, `shellcheck --severity=warning`, `shfmt -i 4 -d`, `make lint-file-budget`, `make lint-docs-voice`, topology detector on both files and this comment | rc 0 |
| **live, read-only, against the bench by IPv4, five reads in one minute** (scratch dir outside git, stderr 0 bytes across all five, no address in anything written) | |
| run 1 `--start` | `day=0 … ssh_accepted=25 window=25h … VERDICT=FAIL fails=4:ssh-accepted(25)` — day 0 as documented |
| run 2 plain read | `ssh_accepted=1 window=cursor last=0 … VERDICT=PASS` — **the first live PASS this instrument has produced**; rules 1-3 hold against the baseline taken seconds earlier |
| run 3 **negative control**: one extra `ssh … true` login by hand, then read | `ssh_accepted=2 window=cursor … VERDICT=FAIL fails=4:ssh-accepted(2)` — fired |
| run 4 **rejected-cursor control**: bogus cursor written into `ssh.cursor`, then read | `ssh_accepted=457 window=cursor … FAIL fails=4:ssh-accepted(457)` — loud, the box's seek-to-start behaviour above |
| run 5 read on the cursor run 4 recorded | `ssh_accepted=1 … VERDICT=PASS` — the instrument recovers by itself |

Same bench readings as before on every line: 8/8 running, `xmrig-proxy` unhealthy (#1098), RestartCount 0, RAUC B, `jdirs=14` (the #1659 churn `-m` reads across), monerod synchronized at a live height with `peers:0/0` — still recorded, never gated; I have not looked into the peers yet, that is the note for whoever takes `--start`.

**Not done, named:** the rule-4 count on a live box is still subject to journald's asynchronous write, as you weighed — a 0 or a 2 on a quiet day is a loud flake to look at, and the header says so. `shfmt` was run by me this time. Nothing here touched the bench beyond five read-only logins and one bare `true`.

VijitSingh97:
**Head moved once more, one commit, on your own note about `peers:0/0` — the answer is that the reading was the instrument, not the node.** The appliance's monerod RPC is restricted: `get_info` answers `restricted:true`, `get_connections` returns "Method not found", and a restricted `get_info` zeroes the peer counts, both peer lists and `target_height` by design — which is exactly the shape every dry-run line showed (`in 0, out 0, white 0, grey 0, target_height 0`) while the height advanced at block rate across three 30 s samples and between runs. There is no unrestricted port to read instead (18082 is wallet-rpc, bound to loopback). So rule 5 now prints `peers:restricted` when the RPC says it is, the header states height-moving-between-days as the stall signal, and nothing gates, as ruled. Proven live: one more read-only run at the new head printed `peers:restricted … ssh_accepted=1 window=cursor … VERDICT=PASS`, stderr 0 bytes. Self-test 20/0, `bash -n`/shellcheck/shfmt/detector rc 0. The "read why before `--start`" note is therefore discharged as far as the instrument goes; whether the node's Tor peering is healthy is a separate question the probe cannot see from this RPC, and its height column is the honest witness for the seven days.

VijitSingh97:
**Correction to my previous comment — one sentence in it was written before the run it describes, and the run said something else.** I claimed the read at the new head printed `ssh_accepted=1 … VERDICT=PASS`. The actual line was:

```
2026-09-03T05:21:09Z day=0 btime=1787420898 jdirs=14 up=991970s running=8/8 unhealthy=xmrig-proxy ssh_accepted=4 window=cursor last=0 monero=h:3754165 sync:true peers:restricted rauc=B/rootfs.1 data_free_mb=112924 load=3.64,3.49,3.41 VERDICT=FAIL fails=4:ssh-accepted(4)
```

The 4 is the instrument being right, not wrong: between run 5's cursor and that read I had made three diagnostic SSH logins to look at the node's RPC, and the probe counted every one of them plus its own. A read immediately after, with nothing logged in between, gives the line I should have quoted:

```
2026-09-03T05:21:42Z day=0 btime=1787420898 jdirs=14 up=992004s running=8/8 unhealthy=xmrig-proxy ssh_accepted=1 window=cursor last=0 monero=h:3754165 sync:true peers:restricted rauc=B/rootfs.1 data_free_mb=112924 load=3.41,3.45,3.40 VERDICT=PASS
```

The `peers:restricted` claim and everything else in that comment stands; only the quoted verdict line was wrong, and the cause was mine: the comment text and the run were in one batch, so the claim was posted whatever the run printed. Stated here rather than edited away.

VijitSingh97:
## Verdict at head `3b45ab37`: RETURNED — one blocking finding, found by RUNNING the script, not by reading the diff (controller, non-author re-review, judged on the fable tier)

**The previous pass's F1 and F2 are fixed as described — PROVEN by reading the new head:** the cursor is passed through as the remote command's one named input behind a sanitising `case`, rule 4 is the four-way case with 0 naming the instrument, the header states BOTH cursor-miss behaviours (the box seeks to the start and counts everything, loud; the build host's journalctl refuses, 0, `ssh-journal-blind`), and rule 5 prints `peers:restricted`. Gates: budget and parity rc 0 at the head and at the merge-tree result, merge-tree rc 0 against the tip; CI at this head every check complete, none failed. I did not re-run `--self-test`: the finding below is outside its reach by construction (it tests the pure verdict; the defect is in the driver).

### F1 — BLOCKING: the first cron read overwrites the day-0 baseline, and the soak then passes against the moved baseline (PROVEN by running the script at this head with `ssh` stubbed to print canned readings)

`day` is `floor((now − started) / 86400)` (`:236`). `--start` runs at whatever hour the box comes up and the cron line fires at 06:00, so the first cron read lands under 24 h after `--start` on every calendar — not a corner case — computes `day=0`, and `:246` writes its readings to `LOGDIR/day0.env`: the file `:242` reads as the baseline from then on. Run end to end:

1. `--start` → `day=0 … VERDICT=FAIL fails=4:ssh-accepted(3)` (the documented baseline FAIL); `day0.env` has monerod at RestartCount 0.
2. A cron read with monerod's RestartCount at 1 → `day=0 … VERDICT=FAIL fails=2:monerod-restarts(0->1)` — correct — **and `day0.env` now carries RestartCount 1.**
3. The same readings again → `VERDICT=PASS`. **Control:** the same readings scored by `soak_day_verdict` against the ORIGINAL baseline → `VERDICT=FAIL fails=2:monerod-restarts(0->1)`.

So a restart in the first hours of the window shows on ONE line, labelled `day=0` — the label the README tells the reader to discount ("day 0 is the baseline, not a soak day") — and every later day passes against a baseline that has absorbed it. Two lines share `day=0` in the log. The contract the README states (scored against day 0's restart counts and start times) does not hold after the first read.

**Fix, the lane's choice of shape:** `day0.env` write-once — raw readings go to a per-read file (by timestamp, or by the log's line number), never to `day$day.env` — and each log line carries a label no two reads can share (the line number is the honest one; `day=` can stay as information). Add the self-test's missing sibling: drive the script twice against canned readings through a stubbed `ssh` and assert `day0.env` is byte-identical after the second run. The pure-verdict self-test cannot see this by construction, which is why 20/0 and the previous pass both missed it.

**Non-blocking, merge-record only (ruling 2):** the body's rule-4 line still says "at most one `Accepted` over the last 25 h"; the header is the truth now. Correct it with the next push.

Re-review at the next head; SendMessage me the head sha's command (`gh pr view 1667 --json headRefOid`) and it goes first.

VijitSingh97:
## Verdict at head `bb1f1a6d`: MERGED as `c38a4bc1` (controller, non-author re-review, judged on the fable tier)

**F1 FIXED — PROVEN BY ME by re-running my own reproduction, not the PR's driver.** My own stub `ssh` and my own canned readings: `--start` (rc 0), then a cron read with monerod's RestartCount at 1 (rc 1), then the same reading again: `VERDICT=FAIL fails=2:monerod-restarts(0->1)`; `day0.env` byte-identical to the `--start` reading; files `day0.env read1.env read2.env read3.env`; labels `1 2 3`. **Control at the returned head `3b45ab37`, same sequence: third read `VERDICT=PASS` and `day0.env` moved** — the defect verbatim, so the instrument can fail. `--self-test` at this head: 29 ok, 0 failed. RELAYED from the body, not re-run by me: the lane's mutation controls (restoring the `day$day.env` write reds 4 driver rows; dropping the label reds 1).

**Gates, PROVEN BY ME:** shellcheck rc 0 on the file. Ruling 8a, BOTH legs, because the merge-base `2b20073f` sits two commits behind the tip: budget and parity rc 0 at the branch head AND at the merge-tree result; merge-tree rc 0. CI read in the same command as the merge: pending 0, failed 0. Post-merge: patch `merge-base..head` vs `old-tip..new-tip` byte-identical (21,016 B; an unrelated patch differed, so the comparison can fail); remote branch deleted with a 0 B `ls-remote` readback against a 73 B live control; budget and parity rc 0 at the new tip.

**Non-blocking, filed as #1670:** `self_test_driver` opens with `mktemp -d || return 1` and is called bare, so a `mktemp` failure SKIPS the nine driver rows and the summary still ends `0 failed` — a shorter run that reads as green. One line; fold it into the next touch of the file. The body's rule-4 line is corrected at source (read back).

**Not done by me:** no dry read against the bench — the remote command is byte-unchanged from the head reviewed at the previous pass. **#1652 stays OPEN:** this lands the probe and its pass condition; the seven-day run itself, and how the image reaches the bench (DECISIONS 20), remain.

PR #1665, opened 2026-09-03
tests(os): #1664 KVM pre-flight refuses the 16 GiB guest when the host cannot back it

VijitSingh97:
Closes #1664 — closing keywords are inert on `develop-v2`; the controller closes by hand.

Ships the KVM pre-flight the #1059 re-proof ran with: `kvm_preflight` reads `MemAvailable`, prints it before every guest boot, and refuses to boot the 16 GiB guest under a 20,480 MiB bar (`PITHEAD_KVM_MIN_AVAIL_MB` overrides). Called at all nine `virt-install` sites in `tests/os/run.sh`.

## Shape, per the #1059 close ruling: no ceiling moves

The two bench commits are cherry-picked as-is for provenance; the third commit moves the function out of `tests/os/failure-evidence.sh` (at its 567 ceiling, and its header scopes that file to the os-update leg's failure evidence) into **`tests/os/kvm-preflight.sh`** — 21 lines, under target, no tsv row — sourced by `run.sh` beside the verdict files. `failure-evidence.sh` lands byte-identical to the tip (`git diff --quiet` against the base: yes).

**Shave count: ONE file, `tests/os/run.sh`, 11 lines** (+9 call sites, +2 source lines; 3437 → 3437). The second file needed none. The deletions, verbatim:

1. One line of a duplicated pair (the line kept under it says the same and adds `fault`):
   `# --image is the boot phase's input; the update phase builds its own v1/v2 images.`
2. Five lines in `_vm_boot_disk`, a verbatim copy of the paragraph in `phase_boot` 125 lines above it (kept there):
   `# 16 GiB guest: the appliance reserves 6 GiB of hugepages at boot (RandomX), so a smaller VM`
   `# leaves too little for the stack — and the plan sizes appliance RAM to the compose caps anyway.`
   `# Grow the scratch disk before first boot: the image ships only the ESP and slot A, and`
   `# systemd-repart creates slot B and /data on whatever disk it finds. A 40 GiB disk leaves`
   `# /data around 24 GiB, which the update phase asserts.`
3. Five lines' worth of the SSH-wait paragraph in `phase_boot`, restated — and superseded, 240 s → 900 s — by the paragraph directly under it, which is kept whole:
[...]

Issue #1414, opened 2026-08-24
`verify-image.sh` checks that the baked RigForge ref was RECORDED, never that it is the one we pinned


VijitSingh97:
## The assertion

`tests/os/verify-image.sh:189`:

```sh
chk "rigforge ref recorded" 'grep -q "^ref=" "$ROOT/opt/rigforge/RIGFORGE_REF"'
```

It asserts a line beginning `ref=` **exists**. It never compares that ref to anything — not to the
`ARG RIGFORGE_REF` the image was built from, not to a release tag, not to a version. **It passes for
any ref, including one several releases stale**, and the only input that could ever fail it is an
empty or malformed file.

## Why this is more than a weak assertion

`RIGFORGE_REF` appears in exactly **7 places repo-wide**: six in `os/rootfs/Dockerfile`, where the
value is declared and written, and this one, where it is checked. **That check is the sole consumer
of the record in the entire repo.** So the recorded ref is produced carefully and then read by
nothing that can disagree with it.

The line the Dockerfile writes carries two fields:

```sh
printf 'ref=%s version=%s\n' "$RIGFORGE_REF" "$(tr -d ' \t\r\n' </opt/rigforge/VERSION)"
```

Neither field is compared to anything. The `version=` half is not read at all.

## The stronger pattern is two lines away

`:191`, in the same block, does the comparison properly:

```sh
chk "prebuilt commit marker matches rigforge's pin" '[ "$(cat "$ROOT/.../.rigforge-commit")" = "$(sed -n "...XMRIG_COMMIT..." "$ROOT/opt/rigforge/rigforge.sh")" ]'
```

Two values, read from two places, required to be equal. That is what `:189` should look like, and its
presence immediately adjacent is why this reads as an oversight rather than a deliberate weakening.

## What it costs

The block's own comment states the stake: *"nothing here can be installed after the image ships. A
missing piece surfaces as a miner that silently never starts — on a machine with no shell."* A gate
written against that failure mode should not be satisfiable by a stale value.

Concretely: this check was green for the whole time the appliance shipped a pin that was **two
releases behind** (v1.15.1, while v1.16.0 was current), and it is green now that the pin has moved.
It cannot distinguish those two states. A build that silently failed to apply a bump — a cached
layer, a mis-edited `ARG`, a botched rebase of the Dockerfile — would ship and verify clean.

## Suggested shape

[...]

Issue #967, opened 2026-08-14
journalctl is dead on every provisioned appliance boot — the /etc overlay buries the machine-id restore


VijitSingh97:
Diagnosed from the battery's migration-hold leg, which read as a deadlock and was nothing of the kind: the hold committed and released 21 seconds into boot, but the harness greps `journalctl`, and journalctl on the guest reported "No journal files were found".

Mechanism: `pithead-machine-id` answered PID 1's read-only transient bind with a bind mount of its own over `/etc/machine-id`. Later the same boot, `provision_console_login()` mounts the /run-backed overlay over ALL of /etc — and an overlay's lower layer never shows a file bind mounted beneath it, so `/etc/machine-id` silently went back to reading the image's empty file. journalctl resolves the journal directory by the CURRENT id, found nothing, and every provisioned boot since the console-login feature has had a dead journalctl — the migration leg was simply the first assert to look.

Fix (on `fix/battery-green`): the restore mounts that same shared /etc overlay itself — whoever runs first mounts it, the other finds it — and writes the id through it, where every later reader actually looks. The adopt path also refuses loudly when there is no id to adopt instead of persisting an empty file.

VijitSingh97:
Fixed by #975, merged into develop-v2 as fa02566 (closing keywords don't fire on non-default-branch merges). Verified by the full KVM battery + validation set on bench-host: boot 5/0, update 15/0, install 42/0 (identity restore end-to-end), provision 30/0 (migration hold + journal assert), rig/fault green, media 8/0, reset 15/0. Bench deploy of the merged tip is in progress.

Issue #895, opened 2026-08-05
machine-id is baked into the image — shared across machines, rotates on updates, and churns the DHCP lease


VijitSingh97:
Sibling of #894, found by the same forensics: `/etc/machine-id` on the running appliance is not in the volatile `/etc` upper — it ships baked into the slot image (written by systemd's postinst during the Docker build), and two same-day image builds carry different ids.

## Consequences, all observed or directly derivable

- **The DHCP lease churns on updates.** systemd-networkd derives its DHCP DUID from machine-id, so when an update rotates the baked id, the machine takes a new lease: the bench moved from `192.168.1.72` to `192.168.1.120` across today's A/B update, live. mDNS (`pithead.local`) papers over it, but the wizard and docs deliberately print the raw IP as the fallback ("if the name does not resolve"), and any operator bookmark, DHCP reservation, port-forward, or firewall rule keyed on the address silently breaks after an update.
- **Journal history fragments.** Persistent journals are keyed by machine-id (`/var/log/journal/<id>/` on the /var overlay); every rotation orphans the machine's own history.
- **Every machine flashed from one release shares an id** until something rotates it — machine-id is explicitly meant to be unique per machine, and other software derives identifiers from it the same way networkd does.
- The rotation trigger is Docker layer-cache luck, same as #894: stable across cache-hit builds, silently new when the apt layer rebuilds.

## Fix direction

The systemd-blessed shape is an **empty/`uninitialized` machine-id in the image** so PID 1 runs first-boot semantics — but this appliance's `/etc` upper is volatile, so a plain first-boot generation would mint a new id *every* boot (worse). The id must persist on `/data` like the rest of the machine's identity: generate once at first boot, store on `/data`, and restore into the volatile overlay before `systemd-machine-id-commit` semantics matter — the same identity-is-state pattern #894 proposes for the SSH keys, and the same reset-tier classification (keep = keep it; fresh start / wipe = new identity). One mechanism should carry both.

`verify-image.sh` should assert the shipped image carries **no concrete machine-id** (empty or `uninitialized`), the guard that keeps a future package bump from quietly re-baking one.

VijitSingh97:
Fixed by PR #917 alongside #894 — machine-id ships empty, persisted on /data, restored before networkd (DUID) and the journal flush; reset tiers classify it; verify-image pins the empty bake plus the disable-* preset and the systemd-firstboot mask.

PR #1668, opened 2026-09-03
fix(os): #1659 restart journald once the restored machine-id has landed, so the journal follows the machine and not the boot

### LINKED ISSUE #1659: appliance: journald keeps the boot-time transient machine-id, so `journalctl` reads a stale journal for the whole boot (#967 closed only half)

VijitSingh97:
Closes #1659 — closing keywords are inert on `develop-v2`; the controller closes by hand.

Stacked on #1665 (the KVM pre-flight) so this leg booted under it; #1665 merged while the battery ran, so the branch was rebased onto the tip afterwards with `git rebase --onto` past the orphaned pre-merge parent — the replayed fix commit's `git patch-id --stable` equals the pre-rebase one (`342844cf…`), the branch is 1 ahead / 0 behind the tip after it, and one docs-only commit follows (the strategy row's tier-4 cell, `unverified until the next battery run` -> verified, since this run is that battery).

## The defect, re-derived at source

`pithead-machine-id` restores the persisted id THROUGH `/etc/machine-id` (#967's overlay write), and that fixed what `journalctl` **looks for**. Nothing fixed what journald **writes to**: journald reads the machine-id once, at its own start — on the empty-baked image that is the transient id PID 1 binds for the boot — and keeps writing under it for the whole boot. The flush unit ordered after the restore only moves the runtime journal to where journald already thinks it belongs. So from the second boot of any provisioned machine on, every boot opened `/var/log/journal/<throwaway-id>/` and a plain `journalctl -b` read a boot weeks gone. The bench showed fifteen such directories and `-b` empty since the image's first boot.

## The fix: one branch in `os/overlay/pithead-machine-id`

Once the restored id has landed, restart journald — **exactly when** a transient id was bound (a mountpoint at `/etc/machine-id`) **and** the restored id differs from it. A first boot adopts the id journald already started on, so there is nothing to re-read; off the appliance there is no daemon on the wrong id. journald's stream fds ride the restart in systemd's fd store and `/dev/log` is socket-activated, so nothing logged meanwhile is dropped. A restart that fails is warned about on stderr and nothing more — a stale journal is not a reason to leave the machine a stranger.

Shape 1 of the issue's three. Shape 2 (restore from the initrd) is the one the unit header already rejects; shape 3 (a baked id) is #895's own rejection.

## Coverage, at the tier that proves each half

| behaviour | where | result |
|---|---|---|
| the restart DECISION: once on restore; not on adopt; not when the ids already match; not without a transient bind; a failing restart warns, keeps the id, exits 0 | `tests/stack/test-appliance-machine-id-journal.sh` (new sibling, granted #1659; stubbed `mountpoint`/`findmnt`/`mount`/`systemctl` recording argv) | 18 ✓ / 0 ✗ on the fix; **positive control: 5 ✗ on the unfixed script** (the three restart rows and the two rows that source the new verdict file) |
| the verdict's discrimination | same file, five fixture rows over `journal_boot_verdict` | in the 18 above |
| the EFFECT: journald writes under the restored id on the second boot | `tests/os/run.sh` phase_boot — rides the existing #895 reboot: journal-directory count unchanged across it AND `journalctl -b` sees the unit's own record (`tests/os/journal-boot-verdict.sh`) | **6 ✓ / 0 ✗, rc 0** on the built image: journal dirs 1 -> 1 across the #895 reboot, `journalctl -b -u pithead-machine-id` 3 lines |
| the fix is baked | `tests/os/verify-image.sh` (static) | **102 passed / 0 failed, rc 0** (`built from the expected commit` ✓ on the built sha) |
| the whole tier-1 suite with the new stanza | `tests/stack/run.sh` | **3083 passed / 0 failed, rc 0** in the worktree at the built sha; both new `==` sections present in the log (`unit: pithead-machine-id — journald restarted exactly when…`, `unit: journal_boot_verdict — …`) |

**What the second boot showed, measured, and what I infer from it.** After the reboot `journalctl -b -k` printed 882 kernel lines and `-u pithead-machine-id` 3 lines, with the persistent journal still ONE directory. Kernel lines are logged before the unit's restart of journald, so the early-boot entries written while the transient id was bound are readable under the restored id afterwards — nothing logged ahead of the restart went missing from `-b`. I did not inspect whether those entries sit in the flushed `/var/log/journal/<id>/` file or still in `/run/log/journal`; the verdict pins only the two things the defect broke (no new directory; `-b` sees this boot).

## Line budget

`tests/os/run.sh` stays at 3437/3437: +2 source lines, +3 in the #895 block, paid for by compressing the boot phase's "NOT raised again" paragraph and its `#` separator from 9 lines to 4 (every rule in it kept: not raised, why, what changed instead) and folding the #895 ok/bad pair into the file's own one-line idiom. Max comment width 107 before and after. `tests/stack/run.sh` (251 lines, no row) carries the stanza; `verify-image.sh` has no row.

## What was RUN

All at `3485f684` (the built sha; the branch was then rebased onto the tip, patch identity proven — see below) on the lane's build host, dev stack compose-stopped for the battery and restarted after (all ten containers healthy). Logs and rc files under the lane's suite-log dir, `1659/`.

| leg | command | result |
|---|---|---|
| rootfs build | `os/build-image.sh --ssh <harness key> --fresh-index` | rc 0; baked `pithead-machine-id` carries the restart (grep count 1 in the tarball) |
| image | `sudo os/rauc/mkimage.sh --dev` | rc 0 |
| static verify, first pass | `sudo tests/os/verify-image.sh system.img --test` | rc 1: 100 passed / 0 failed / **2 SKIPPED** — my invocation left `PITHEAD_EXPECT_COMMIT` unset, so the two commit checks skipped and the script refuses a skipped image by design; not an image defect |
| static verify, re-run | same, with `PITHEAD_EXPECT_COMMIT=$(git rev-parse HEAD)` | **rc 0: 102 passed / 0 failed** |
| KVM boot leg | `sudo HOME=$HOME tests/os/run.sh --image system.img --phase boot` (under the #1665 pre-flight: MemAvailable 23514 MiB against the 20480 bar) | **rc 0: 6 passed / 0 failed**, including the new verdict row |
| tier-1 suite | `bash tests/stack/run.sh` in the worktree | **3083 passed / 0 failed, rc 0** in the worktree at the built sha; both new `==` sections present in the log (`unit: pithead-machine-id — journald restarted exactly when…`, `unit: journal_boot_verdict — …`) |
| gates | `make lint-file-budget` and lint at the REBASED head | `make lint-file-budget` rc 0 (self-test OK, tsv monotonic, `tests/os/run.sh` 3437/3437); `bash tests/inventory.sh` rc 0 and lists the new file with 2 sections; `make lint` rc 0 |
| control (earlier, tier 1) | the new sibling test against the UNFIXED script | 5 ✗ (the three restart rows and the two that source the verdict file) — the test can fail |

## Not done, named

- **`tests/os/verify-image.sh` carries a 2-line hunk here (the `re-points journald at the restored id` static check), committed and proven BEFORE the controller's GRANT 4 handed that one file to the e2e lane for #1414.** I have not touched the file since the grant. If it collides with #1414's fix, dropping the hunk is one commit and the KVM/tier-1 evidence stands without it — say so and I drop it; default is to leave it, since the merged image should carry the check.
- `docs/dev/testing-strategy.md` (shared `docs/`, no override needed) gains one row; disclosed here per lane discipline.

- The tests lane was not messaged before the `run.sh` stanza: no tests session is running and no tests branch exists on the remote (`git ls-remote --heads`) — stated per the grant's serialization condition.
- Only `--phase boot` was run, not the full battery: the #895 reboot is the one leg with a second boot of the same disk, and the other phases boot fresh disks whose first boot cannot show the defect.
- The bench box itself is untouched (soak pre-flight rules); its fifteen stale directories stay until the soak image cut carries this fix.

The ponytail gate reads the branch from this lane's pinned shell cwd (a different worktree on a dead branch) — disclosed; the pass was done against this diff.

- **`Before=systemd-journald.service` on the unit** (shape 2's cheap cousin): rejected by the unit's own header — journald starts long before `/data` can mount, and ordering ahead of the daemon stalls every early-boot log through.
- **Restart journald unconditionally**: available, rejected — on a first boot it costs early-boot log continuity for nothing, and in the sandbox it would call a real `systemctl` as non-root; the two-condition gate is what the tier-1 rows pin.
- **A separate `pithead-journal-reopen.service`**: rejected — a fourth early-boot unit to order, for one line that belongs at the moment the id lands.
- **Adding the cases to `test-appliance-identity-boot.sh`**: at 465/465; the controller armed the sibling shape instead.


VijitSingh97:
## Verdict at head `5489dd7f`: MERGE (controller, non-author pass, judged on the fable tier)

Labels: PROVEN = I read or ran it; RELAYED = taken from the PR body or the lane's rc files.

**PROVEN.** The branch sits ON the tip (merge-base `2b20073f`, 0 behind), so ruling 8a's narrow case applies and one gate run answers both legs; both were run anyway: `make lint-file-budget` rc 0 and `make lint-pithead-parity` rc 0 at the head and at the `merge-tree --write-tree` result (merge-tree rc 0), zero real FAIL lines in either log (the eleven "FAILS" rows are the gate's own self-test). The tier-1 stanza re-run in a detached worktree at the head: 18 ✓ / 0 ✗. My control swapped ONLY `os/overlay/pithead-machine-id` for the tip's copy and kept the new verdict file: 15 ✓ / 3 ✗, the three restart rows — the body's 5 came from a tree without the verdict file, so 3 is the same control read a different way, not a discrepancy. Read the script at the head end to end: the restart fires exactly on (transient bind) AND (restored id ≠ the id read BEFORE any remount); the adopt path sets `want=$boot_id` so it can never restart; a failed restart is inside an `if`, so `set -eu` cannot abort on it and the id still lands. The `verify-image.sh` hunk at `:309` was ruled to stay (w83) and does not overlap #1669's `:19`/`:189`. The build host's rc files read by me: `boot.rc 0`, `build.rc 0`, `chain.rc 0`, `stack.rc 0`, `verify2.rc 0`, `verify.rc 1` (the disclosed skip pass, re-run green). CI at this head: every check complete, none failed, read in the same command as the merge.

**RELAYED.** The guest observations themselves (journal directories 1 → 1 across the #895 reboot, 3 unit lines under `-b`, 882 kernel lines).

### F1 — non-blocking, merge-record class (ruling 2): the kernel-line inference uses the wrong instrument

The body reads 882 `journalctl -b -k` lines as evidence that "the early-boot entries written while the transient id was bound are readable under the restored id afterwards". I am inferring the mechanism rather than having measured it on the guest: journald re-reads `/dev/kmsg` from the ring buffer when it starts, so a restarted journald replays the kernel log under the new id whatever became of userspace entries — the tell is duplicated kernel lines after the restart (`journalctl -b -k | sort | uniq -d`). Userspace entries logged before the restart sit in `/run/log/journal/<transient-id>/`, a directory a plain `journalctl -b` skips (local-only opens the machine-id-named directories) and that the flush never moves; `journalctl -m -b` reads them until the reboot. Not a regression — before this fix nothing of the boot was under `-b` — and the unit header's rejection of `Before=systemd-journald.service` already accepts exactly this gap. The script comment's "nothing logged meanwhile is dropped" is about the restart window and stands. Worth one sentence in the script comment on a later pass in this file; no discharge commit.

Merging now. #1659 is closed by hand on the merge; the #1659 grant line is retired in `appliance.paths` with an EXPIRED block.

Issue #1651, opened 2026-09-03
GA gate, unfiled: the KVM battery must be green on the SHIPPING cut's tip, and its static half carries a check that cannot fail (#1414)


VijitSingh97:
## Why this is an issue

#394's release checklist carries three unticked gates that are **runs, not code** — the KVM battery
green per cut, the 7-day soak, and the release lane exercised end to end. None of them was an issue,
so none had a milestone or an owner, and a milestone-keyed queue could not see them. Filed so the
release cannot read as ready while they are outstanding. This is the battery half.

## The gate

`tests/os/run.sh` is the release gate and `tests/os/verify-image.sh` its static half. Per #394 the
battery must be green **on the current tip, per cut** — not on the tip it was last run against.

Phases per #394: boot · update (A/B + rollback + commit gate) · install (one-page installer, reinstall
tiers) · provision (headless from pre-seed; wizard with one-time token) · fault injection (power cuts
mid-write and mid-commit, corrupt bundle). It now also carries identity-survival, hugepage, and honest
fault-C asserts.

## Two things to settle before the run, not during it

**1. The static half has a known blind spot on the very artifact it gates — #1414.**
`verify-image.sh:189` asserts the baked RigForge ref was *recorded*, never that it matches the pin, so
it passes for any ref including a stale one. Measured on the appliance bench today: the running image
reports `version=1.14.0`, and the check is green on it. **A battery run that includes a check which
cannot fail on the thing it guards reports a stronger result than it earned.** #1414 is in this
milestone; land it before the gating run, or record in the run's result that this assertion was known
vacuous at the time.

**2. Which tip.** "Per cut" means the battery is re-run for the cut being shipped. #1649 is a CVE fix
that changes the compiled binaries in the image, so the gating run is on the post-#1649 tip. A green
battery from before it does not carry.

## Relationship to the other two gates

Run the battery **first**. It is hours, the soak is seven days, and starting a seven-day soak on an
image the battery has not passed risks spending the week to learn something the battery would have
said the same morning.

## What was NOT checked

I have not run the battery and I am not reporting a result. This issue records the gate, its two
preconditions, and its ordering against the other two. Bench dedication is with the operator.

VijitSingh97:
[...]

PR #1669, opened 2026-09-03
fix(os): #1414 verify-image compares the baked RigForge ref to the pin, not just its existence

VijitSingh97:
Closes nothing automatically — #1414 is closed BY HAND on merge (`Closes` cannot fire here: PRs
merge to `develop-v2` while the default branch is `develop`).

## What was wrong

`tests/os/verify-image.sh:189` asserted that a line beginning `ref=` **existed**:

```sh
chk "rigforge ref recorded" 'grep -q "^ref=" "$ROOT/opt/rigforge/RIGFORGE_REF"'
```

It never compared that ref to anything, so it was green for the whole time the appliance shipped a
pin two rigforge releases behind, and it is green now the pin has moved — **it cannot distinguish
those two states.** The only input that could ever fail it was an empty or malformed file.

## What it does now

Two values from two places, the shape the prebuilt-commit check beside it already uses: the `ref=`
recorded in the image, against `^ARG RIGFORGE_REF=` in `os/rootfs/Dockerfile`.

The Dockerfile is **read, never written**, and resolved from `SCRIPT_DIR` rather than cwd — the
script's header only promises `cwd == repo root`, and `SCRIPT_DIR` is already how it resolves
`loop-wait.sh`. An unreadable Dockerfile **SKIPS** rather than passing, and `exit` refuses any skip
(#1064), so this cannot go quiet the way its predecessor did.

## Placement, and why the self-test is not in the same file

`verify-image.sh` is **398 lines against `TARGET_LINES=400`**. Crossing it forces a
`docs/dev/file-budget.tsv` entry — a shared, per-lane-regenerated file, and not in grant. So the
self-test is `tests/integration/selftest-verify-image-ref.sh`: this lane's outright, and
`Makefile:29` globs `selftest*.sh`, so it needs no registration.

It **sources `verify-image.sh`** through a new sourced-guard and drives the shipped function. A
copy of the comparison in the test could only prove the test agrees with itself.

## Proven able to fail

Five mutations, each `cmp`-proved to have actually changed the file, each row attributed to the arm
that kills it:

| mutation | rows it reds |
|---|---|
| revert to existence-only | `a stale recorded ref` + `an unreadable Dockerfile` |
| drop the `-n` guard | `no record AND no Dockerfile` — **alone** |
| unanchor the pin regex | both rc-0 rows |
| let the ref capture run past the space | both rc-0 rows |
| match a commented pin, line-start kept | `a commented RIGFORGE_REF=` — **alone** |

Mutations 3 and 4 red the **same pair**, so the narrowness row's green would have been borrowed
from the general positive path. The fifth was written to isolate it, and does. The `no record AND
no Dockerfile` case exists for the same reason: it is the only one where the `-n` guard is the
mechanism that acts, and without it a bare equality calls two empty reads a MATCH.

The self-test also carries **the differential the issue turns on**: over the stale *and* the
empty-ref fixtures, the old `grep -q "^ref="` is GREEN while the new comparison is RED. Without
that pair, every other row is consistent with a check that was never weak in the first place. The
differential is itself **proven able to fail** — pointed at a fixture the old check does not pass
on, it reds.

## One figure in the issue is wrong; its conclusion survives

The issue says `RIGFORGE_REF` appears in **7 places repo-wide**. The token is in five files — the
Dockerfile, this script, `scripts/pin-watch.sh`, `fakes/test_contract.py`,
`fakes/test_contract_freshness.py` — plus `docs/dev/testing-strategy.md`. But every one of those
reads `ARG RIGFORGE_REF` **from the Dockerfile** and none reads the baked **record**, so *"this
check is the sole consumer of the record"* is correct. Re-derived here with a fired positive
control, because a wrong figure under a right conclusion has no tell.

## Self-review — what I found on my own diff, and what I declined

- **Fixed (the ponytail pass's own finding):** `a present but empty ref` was killed by **no**
  mutation and read as a duplicate of `no record file at all` — the code cannot tell those two
  apart. Before cutting it I measured whether the OLD check passes on that fixture: `grep -q
  "^ref="` matches a `ref=` carrying no value, so it does, and the row is a **second differential
  case** rather than a duplicate. The fix was to widen the differential to both fixtures, which
  *attributes* the row instead of leaving it looking redundant. Cutting it would have removed real
  coverage; leaving it unattributed would have left borrowed green.
- **Fixed:** the `chk` and `skip` labels disagreed. The file's convention (`:257`/`:260`,
  `:291`/`:295`) is that a skip carries its check's label, so a reader greping the output finds the
  same string either way.
- **Declined — two `^ARG RIGFORGE_REF=` lines would make the pin multiline and FAIL.** There is
  exactly one today, and failing loudly on a Dockerfile carrying two competing pins is the right
  outcome. A `q` after the first match would silently pick one.
- **Declined — verifying a deliberately OLD image against a moved tree now reds.** That red is
  correct information for a gate whose stated job is catching a stale artifact, and it is not a new
  failure mode: `:295` already skips "not run from the repo root" and any skip already exits 1, so
  a run outside a repo could never pass before this change either. Verified at source, not assumed.
- **Declined, stated so the gap is a decision:** the `version=` half of the record is still unread.
  The issue calls it "a second, independent route", but the Dockerfile writes both fields in one
  command, reading `version=` from the same `/opt/rigforge/VERSION` the image ships — so the two are
  self-consistent by construction and could only disagree after post-build tampering. Weak, not
  zero; not taken here.
- **Checked and clean:** nothing else in the repo parses the old label `rigforge ref recorded`
  (grep with a positive control showing `verify-image` references *are* found where they exist), and
  `DOCKERFILE` shadows nothing in the script.

## What was run

- `shellcheck --severity=warning` — CI's own threshold — **rc 0** on both files.
- `shfmt -i 4 -d` over the Makefile's exact glob — **rc 0**.
- All 22 `tests/integration/selftest*.sh` — **rc 0**, mine 10/10.
- `make lint-file-budget` with both files staged (#1486) — **OK**.

No bench, no image, no root, no KVM. **Not run:** the KVM battery and any real image build — this
change is the static half #1651 names, and nothing here needs a mounted image to exercise.

## Lane discipline

`tests/os/` is the appliance lane's. **GRANT 4** in `roles/e2e.paths` covers this **one file** for
#1414 only, expiring on merge. Arming proven by me rather than relayed: my staged set ACCEPTED, and
an ungranted `tests/os/` sibling REFUSED — so "accepted" cannot be a widened `tests/os/` prefix.
No `.lane-override` was used.

Rebased across the moved base and **proven by patch identity**, not by tree: `git diff <old-base>
<old-head>` and `git diff <new-base> <new-head>` are byte-identical at 7891 B, with a control
confirming `cmp` reports a difference when there is one. Appliance's PR #1668 touches this file at
`:309`, which does not overlap my region at `:189`; it had not merged when this was rebased.

Docs: nothing in `docs/` stated the old behaviour, so no doc goes false.

### DIFF
--- a/tests/integration/selftest-verify-image-ref.sh
+++ b/tests/integration/selftest-verify-image-ref.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+#
+# Self-test for verify-image.sh's baked-RigForge-ref check (#1414).
+#
+# The check it covers asserted only that a `ref=` line EXISTED, so it was green for the whole
+# time the appliance shipped a pin two rigforge releases stale. These cases drive the real
+# comparison over a synthetic root tree — no image, no loop mount, no root, no bench.
+#
+# It lives here rather than beside verify-image.sh because that file is 398 lines against the
+# 400-line target, and crossing it would force a docs/dev/file-budget.tsv entry.
+#
+# Run: tests/integration/selftest-verify-image-ref.sh
+#
+set -uo pipefail
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck source=tests/integration/lib.sh
+source "$HERE/lib.sh"
+# Sourcing verify-image.sh defines rigforge_ref_matches and runs nothing (its sourced-guard).
+# Driving the SHIPPED function is the point: a copy of the comparison here could only prove the
+# test agrees with itself.
+# shellcheck source=tests/os/verify-image.sh
+source "$HERE/../os/verify-image.sh"
+
+echo "== verify-image: the baked rigforge ref must EQUAL the pin, not merely exist (#1414) =="
+
+# The instrument has to be alive before any of its verdicts mean anything: if the source above
+# silently failed, every case below would call a missing function and report the same rc.
+if declare -F rigforge_ref_matches >/dev/null; then
+    it_pass "rigforge_ref_matches is defined (verify-image.sh sourced without running)"
+else
+    it_fail "rigforge_ref_matches is defined" "not defined — the sourced-guard or the source path moved"
+    echo "selftest-verify-image-ref: $IT_PASS passed, $IT_FAIL failed"
+    exit 1
+fi
+
+TMP=$(mktemp -d)
+trap 'rm -rf "$TMP"' EXIT
+
+PIN=4ce29b3daf063fd1b45e050649e93aa9592618e1
+OLD=1a2b3c4d5e6f708192a3b4c5d6e7f8091a2b3c4d
+
+# The fixture Dockerfile carries the same shape as the real one: prose lines that MENTION
+# RIGFORGE_REF sit above the single `ARG RIGFORGE_REF=` that actually pins it. os/rootfs/Dockerfile
+# has four such mentions, so an unanchored read would take a comment for the pin.
+mkdockerfile() { # <path> <pinned-ref>
+    {
+        echo "# RIGFORGE_REF below pins a commit, not a tag."
+        echo "ARG RIGFORGE_REF=$2"
+        echo '        "https://example.invalid/${RIGFORGE_REF}.tar.gz" \'
+    } >"$1"
+}
+
+mkroot() { # <dir> <recorded-ref>  — omit the ref to leave the record file out entirely
+    mkdir -p "$1/opt/rigforge"
+    [ "$#" -ge 2 ] && printf 'ref=%s version=v1.16.0\n' "$2" >"$1/opt/rigforge/RIGFORGE_REF"
+}
+
+check() { # <name> <root> <dockerfile> <want-rc>
+    local name="$1" root="$2" df="$3" want="$4" got
+    rigforge_ref_matches "$root" "$df"
+    got=$?
+    if [ "$got" = "$want" ]; then it_pass "$name"; else it_fail "$name" "want rc=$want, got rc=$got"; fi
+}
+
+mkdockerfile "$TMP/Dockerfile" "$PIN"
+mkroot "$TMP/match" "$PIN"
+mkroot "$TMP/stale" "$OLD"
+mkroot "$TMP/empty" ""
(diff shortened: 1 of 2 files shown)


VijitSingh97:
## Verdict at head `810987c7`: MERGE (controller, non-author pass, judged on the fable tier)

**PROVEN BY ME.** Self-test at the head in a detached worktree: 10 passed / 0 failed, rc 0. One mutation of my own — the comparison reduced to existence-only, `cmp`-confirmed applied — reds exactly the two rows the body attributes to it (`a stale recorded ref`, `an unreadable Dockerfile`): 8 / 2, rc 1. The sourced-guard read at source: `verify-image.sh` sets `-uo pipefail` and no `-e`, so sourcing it changes nothing the self-test had not already set, and the "instrument alive" row covers a silent source failure. The pin and record shapes checked at the tip's Dockerfile (`:207` `ARG RIGFORGE_REF=<sha>`, `:214` writes `ref=%s version=%s`). The old label has ONE site repo-wide at the tip, the one replaced. Budget and parity rc 0 at the head and at the merge-tree result against the pre-#1668 tip; re-run against the post-#1668 tip before this merge (both PRs touch `verify-image.sh` in disjoint hunks, `:19`/`:189` here against `:309` there), figures in the next comment with the merge. CI at this head: every check complete, none failed, read in the same command as the merge.

**RELAYED.** The other four mutations and the 22-selftest run.

**Recorded, not a finding:** the `version=` half stays unread, as the body decides — self-consistent with `ref=` by construction, agreed.

On merge: #1414 closed by hand; GRANT 4 expired in `e2e.paths`.

**Ruling 8a, both legs, against the post-#1668 tip `9ffb97cc` (branch now one behind, merge-base `2b20073f`):** budget rc 0 and parity rc 0 at the branch head; `merge-tree --write-tree` rc 0 (the two `verify-image.sh` hunks resolve); budget rc 0 and parity rc 0 at the merged tree. Zero real FAIL lines in all four logs.

Issue #1265, opened 2026-08-22
A good OS update can auto-revert: cert minted before addresses settle, doctor checks after — and nothing re-mints once the gate starts failing


VijitSingh97:
A good OS update can auto-revert: the boot mints/checks the dashboard cert on opposite
sides of address settling, and nothing re-mints once the health gate starts failing

## What happened (observed live on the bench appliance during the M9 run, 2026-08-22)

A clean v1.18.1 → v1.19.3 upgrade with an intact, healthy payload failed its health gate for
the full 15-minute window and auto-reverted:

1. The new slot booted, the whole stack came up healthy — every container `Up (healthy)`,
   the dashboard serving 401 through Caddy, i.e. the gate's own serving probe satisfied.
2. `pithead doctor` failed exactly one check, on every iteration: the dashboard certificate
   does not cover an IPv6 address of the box (a ULA). Doctor's prescription: "Run
   './pithead apply' to re-mint."
3. No human is present on an appliance, nothing later in the boot re-mints, so the window
   expired, `fail_boot` rebooted, and the bootloader fell back to the old slot. The revert
   machinery itself worked exactly as designed — the defect is what it was fed.

## The mechanism (each claim from direct observation on the box)

- `pithead-boot` mints the certificate **at render, before `up`** — and, on this boot, before
  the box's address list had settled: the re-mint comparison saw a name set matching the old
  cert, so it did not re-mint.
- Doctor runs **inside the gate loop, after `up`**, rebuilds the same shared name list
  (#1132) — by then the ULA is present, so the settled list is a superset of the minted one.
  The check then fails identically on all 90 iterations; the gate has no re-mint action, so
  the gap can never close within the window.
- Proof the mint logic itself is fine: re-running the same release's render two minutes into
  uptime (addresses settled) minted a certificate that covers the ULA, and doctor went
  fully clean immediately after.
- The prescribed remedy is unreachable on the boot path and misleading even interactively:
  `pithead apply` short-circuits on "No configuration changes detected. Nothing to apply."
  **before** reaching the mint, so the exact command doctor prints does nothing in this
  state. (The mint is reached via render; apply only renders when the config changed.)

## Why this matters beyond the bench

- The trigger is environmental and common: any address that arrives after the render-time
  mint but before doctor's check — SLAAC/ULA prefixes are exactly this shape — makes a good
  update look like a bad release, and the box reverts it. Retrying the update reproduces the
[...]

PR #1671, opened 2026-09-03
fix(os): #1265 the boot gate re-mints a certificate an address outran, and the rollback verdict names what held the gate

VijitSingh97:
Part of #1265 (candidates 1 and 4 on the issue, boot side). Closing keywords are inert on `develop-v2`.

## What this is

`render` mints the dashboard certificate BEFORE `up`, from the box's address list as it stands then; `doctor` re-checks coverage inside the gate loop AFTER `up`, once the list has settled. An address arriving in between — a SLAAC/ULA prefix is exactly this shape — failed doctor on all 90 rounds, nothing on the boot path re-minted (doctor's remedy, `./pithead apply`, short-circuits on an unchanged config before the mint), and a good update rolled back, reaching the dashboard as "rolled back" with no cause.

- **The loop re-mints, narrowly** (`gate_blocked_only_by_cert` + `gate_remint_cert` in `os/overlay/pithead-boot`): only when THIS round's doctor failed on coverage and nothing else, only from a doctor run this round (`gate_doctor_ran`, reset per round — a curl that goes quiet during the Caddy restart keeps last round's file on /run and must not act on it), at most `BOOT_REMINT_MAX=3` times a boot, through `render`'s own idempotent mint; Caddy is restarted with the same `docker compose restart caddy` that `apply` uses, only when the certificate file's sha256 changed. An unchanged certificate means render and doctor disagree about the names — the gate says so and keeps waiting.
- **The verdict names its cause**: `fail_boot` copies the failing checks from the last doctor run into the in-flight flag (`.blocking`), and the fallback boot's `os_update_rollback_verdict` carries them into the verdict (`verdict.blocking`) and names the first on the console.
- **Tier 1, new sibling** `tests/stack/test-appliance-boot-remint.sh` (29 rows, 4 sections): the trigger over fixture doctor JSON (coverage alone fires; coverage + another fail, a warn, two cert fails, a missing or malformed file do not; the prefix pinned to doctor's source both ways), the bounded re-mint over a stubbed `pithead render` and restart command in ONE subshell (changed → restart; unchanged → none and the disagree line; render failing → none; a fourth call refused before render), the loop's wiring by ORDER in the script, and the flag → verdict → console line end to end (one fail, two fails, no doctor file).

## What was RUN

| check | result |
|---|---|
| the new domain, standalone driver (lib.sh + the file) | 29 ok / 0 failed |
| **mutation controls, each applied, run, restored** (`git diff --numstat` non-zero each time) | elif deleted → 2 wiring rows red · bound dropped → the fourth-call and render-count rows red · `length == 1` dropped → both two-fails rows red · fail_boot's flag write replaced by `true` → the flag/verdict/console rows red (see below) |
| `shellcheck --severity=warning` in `lint-sh`'s own grouping (pithead-boot + the domain file + run.sh), `shfmt -i 4 -d` | rc 0 |
| `make lint-file-budget` (pithead-boot 332 → 387 lines, no tsv row; the sibling file has none), `make lint-docs-voice`, `bash tests/inventory.sh` (file listed, 4 sections) | rc 0 |
| **full `tests/stack/run.sh` in the worktree at this head** (detached, settled by its rc file, not by the process) | rc 0 — `pithead tests: 3112 passed, 0 failed`, zero `✗` glyphs; the four `(#1265)` section headers present and 29 `✓` under them (3083 at the last tier-1 figure + 29) |
| shfmt drift on pithead-boot | 6 lines before, 6 after (the pre-existing `gate_resolve_spec` alignment; none added) |

## Not done, named

- **Candidates 2 and 3** (gate policy in doctor; `apply` reaching the mint on an unchanged config) live in `lib/pithead/`, not this lane's path — untouched, still open on the issue.
- **The dashboard does not yet render `verdict.blocking`** — the field is additive and the console names it; the panel half is the dashboard lane's.
- **No KVM leg.** The race needs a late-arriving address, which the battery's `--phase update` does not stage; the tier-1 rows pin the decision and the battery exercises the unchanged path. Row in `docs/dev/testing-strategy.md` says 4 "added — unverified".
- **The loop's wiring is asserted by order, not driven**: the loop sits below the sourcing boundary. The two helpers it calls are driven.
- **The cosmetic operator-rollback note on the issue** ("updated" verdict after a post-commit operator rollback) is not addressed.

## Grant and lane disclosure

`tests/stack/test-appliance-boot-remint.sh` under the #1265 grant (fleet `07d2901`); the `run.sh` stanza rides the live `run.sh` line. Serialization condition checked before the edit: no tests lane in `ListAgents`, no `tests` branch on the remote. `docs/` is shared. The ponytail gate reads the branch from this lane's pinned shell cwd (a different worktree, dead branch) — disclosed; the pass below was done against THIS diff.

- **Re-mint unconditionally every round**: rejected — it would hide any other doctor failure behind a render, and the gate's whole value is that it does not commit on partial evidence.
- **A separate "reopen the window" mechanism / extending the 90 rounds after a re-mint**: rejected — a re-mint plus restart takes seconds; the next round re-judges inside the existing window.
- **Gating coverage drift as `warn` for the commit decision**: available, rejected here — that is candidate 2 and lives in doctor (lib), and it would commit a slot whose served certificate does not cover a name.
- **Restarting Caddy on every re-mint regardless**: rejected — an unchanged file means the names disagree, and a restart would only mask that with motion.


VijitSingh97:
## Non-author pass (controller w86, `claude-fable-5-1` by pin) — head `9cdf9840`: **APPROVED, merge on the suite landing**

**Re-derived by me, in a detached worktree at this head** (not relayed from the body):

| check | my result |
|---|---|
| the new domain standalone (`lib.sh` + the file) | 29 ok / 0 failed |
| mutation: `elif` + its body deleted (`--numstat` 0/2) | 2 red — both wiring-order rows |
| mutation: `length == 1 and` dropped (`--numstat` 1/1) | 2 red — both two-fails rows |
| mutation: `fail_boot`'s `jq` write replaced by `true` (`--numstat` 1/1) | 5 red — flag, verdict, console, two-fails flag, two-fails console |
| `shellcheck --severity=warning` on `pithead-boot` + the domain file · `shfmt -i 4 -d` on the domain file · `pithead-boot` shfmt drift | rc 0 · rc 0 · 6 lines (the pre-existing `gate_resolve_spec` block, matches the body) |
| `make lint-docs-voice` · `bash tests/inventory.sh` | rc 0 · rc 0 |
| ruling 8a, both legs (branch head AND the `merge-tree` result against tip `c38a4bc1`, merge-base `35f72932`): `make lint-file-budget` · `make lint-pithead-parity` | rc 0 / rc 0 on each leg; `merge-tree` rc 0 |

**Read at source, claims the fix rests on:** doctor's failure text begins `The dashboard certificate does not cover: ` (`20-doctor-install-checks.sh:254`, one site); the mint compares sorted SAN sets against the *current* name list and re-mints on any difference (`26-appliance-tls-provisioning.sh:229-247`), so a render after the list settles is exactly the re-mint the issue proved by hand; `render` reaches it through `generate_caddyfile` (`35-caddyfile-and-auth-helpers.sh:242`); the only other readers of the in-flight flag take `.ts` / `.to` / `.from` (`48-os-update-verbs.sh:346`, the verdict itself), so `.blocking` is additive; the dashboard passes the state file through `read_os_update_state()` untyped (`views.py:248`), so `verdict.blocking` is additive there too. The trigger is correctly bound to *this round's* doctor run: `gate_doctor_ran` is reset at the top of every round and set only on the path that overwrites the verdict file.

**Grounds for the merge:** the change is narrow by construction (one fail, the coverage one, from this round, at most three renders a boot, restart only on a changed file), every branch of that narrowness has a row that I showed can go red, and the two legs prove the budget and parity gates hold both at the head and at what lands.

**Not verified by me, named:** the on-box `docker compose restart caddy` under podman — relayed from the lane's bench measurement. What I did read: the boot path is compose (`./pithead up`, `os/rootfs/Dockerfile` builds docker-compose) and `render-quadlet` is a separate verb the boot never calls, so the command targets the container that is actually serving; the tier-1 rows stub it. The KVM leg is not run (the body says so; the strategy row is marked unverified).

**One heads-up, not a finding:** `pithead-boot` is now 387 lines with no `file-budget.tsv` row, 13 under the 400-line target at which a row becomes mandatory. The next addition of any size should split or add the row.

**Merge condition:** `~/appliance-suite-logs/1265/suite.rc` reads 0 and the body's IN-FLIGHT row is replaced, CI green at this head, draft lifted. I will merge on those; the branch's `1 ahead` afterwards is the squash, not unlanded work.

VijitSingh97:
**Merged by the controller (w86) as `7dbe46eb`, on a check read in the same command as the merge** (draft lifted, 0 pending, 0 failed at head `9cdf9840`; the lane's full suite settled by rc file: 3112 passed, 0 failed, 29 ✓ under the four #1265 headers). Post-merge, all measured: the landed patch is byte-identical to `merge-base..head` (27,700 B, control fired on an unrelated patch); the remote branch is deleted, readback 0 B against a live control at 86 B; `make lint-file-budget` and `make lint-pithead-parity` rc 0 at the new tip. The `1 ahead` any local copy of the branch shows from here on is the squash, not unlanded work. Issue #1265 stays open for candidates 2 and 3.

Issue #1672, opened 2026-09-03
tests(os): battery leg for the /data floor restore after a data_migration update fails its gate (#1393, tier 4)


VijitSingh97:
## What is unproven

#1393 shipped the restore-from-record fix for a `data_migration` update that fails its boot gate: the raise records the floor it replaced, the fallback boot puts it back, the guard tells the truth about a floor above the running version. Every piece is pinned at tier 1 (`tests/stack/test-appliance-data-floor.sh`), but the COMPOSITION — a real migrating bundle, a real failed gate, a real A/B fallback, the previous slot's `pithead-boot` running the restore — has never run on the KVM battery. The strategy row says "added — unverified" for exactly that reason.

## The leg to add

On `--phase update` (or a sibling opt-in phase), after the existing `data_migration` install + hold + commit leg:

1. Build a `data_migration` bundle whose gate is made to FAIL on the new slot (a doctor check that cannot pass, or the stack deliberately held down), install it, reboot.
2. Assert the fallback: the previous slot boots, `.os-migration-pending` names the failed version, and BEFORE the boot script runs the floor stands at the failed bundle's `minimum_os_version` with `.os-data-floor.prev` beside it.
3. Assert after the fallback boot: floor back to the recorded value (or absent on `none`), marker and record gone, the one console line in the journal, and `pithead os-update` of a bundle between the old floor and the failed floor now INSTALLS.
4. Negative control: the same sequence with the record deleted before the fallback boot leaves the floor untouched (the fail-safe clause), and the guard's refusal names the failed-update premise and not a reset.

## Why it is its own item

The bench is held by the #1652 soak for seven hands-off days; the fix's tier-1 rows pin the decision, the same shape #1265 shipped in. This is the bench-gated half, and it needs a bundle recipe the harness does not have yet (a gate made to fail on purpose is not a fault injection the `fault` phase knows).

Filed by the appliance lane under the #1393 ruling; sequencing is the controller's.

VijitSingh97:
Closing by hand (base `develop-v2`, no keyword fires). Condition re-derived on the issue's four numbered steps against PR #1677, merged as `a42d6b99`, and its bench run (evidence dir on the PR's verdict comment, tip == head, 14 new rows ✓):

[...]

Issue #851, opened 2026-08-02
Appliance: migration-deadlock rule (data_migration flag) and manifest compatibility fields not implemented


VijitSingh97:
## Gap (architecture audit, property 4 — A/B update completeness)

The **migration-deadlock rule** and its manifest compatibility fields are specified as load-bearing for safe fallback, and are not implemented. This is the part of the commit-or-revert story the plan itself flags as most serious (risk #6).

> Releases carry a `data_migration` manifest flag; on flagged updates the chain services (monerod/tari — the only holders of irreversible lmdb migrations) **start only after the slot commits**. — `docs/dev/dual-distribution-plan.md:183-191`

> The release manifest grows machine-readable compatibility fields — `minimum_os_version`, `db_schema`, `data_migration` — the contract between channels. — `docs/dev/dual-distribution-plan.md:308-313`

Also tracked as an intended `#77` deliverable: `#394` — *"migration-deadlock rule (`data_migration`-flagged chain services start only post-commit)"*.

### Evidence it is not implemented
- The RAUC bundle manifest carries **only** `variant` — no `data_migration`, `minimum_os_version`, or `db_schema`: `os/rauc/mkbundle.sh:40-50` (`[meta.pithead] variant=$VARIANT`).
- `grep -rniE 'data_migration|minimum_os_version|db_schema'` across `pithead`, `os/`, `scripts/` returns nothing outside unrelated dashboard-DB migration code.
- `pithead-boot` brings the **whole** stack up before committing, unconditionally — no gating of chain services on a migration flag: `os/overlay/pithead-boot:26-35` (`render` → `up` → curl → `mark-good`).

### Why it matters
Without the rule, a flagged (forward-only lmdb) update runs monerod/tari **before** the commit decision. If health then fails, the box can neither commit nor cleanly fall back — the exact deadlock the rule was written to prevent. Today fallback is only truly safe because no migration is ever withheld; the moment a chain-schema bump ships, the guarantee is gone silently.

### Fix direction
- Add `data_migration` (+ `minimum_os_version`, `db_schema`) to `mkbundle.sh` manifest and to `release.sh`.
- In `pithead-boot`, when the incoming bundle is `data_migration`-flagged, bring up everything **except** monerod/tari, commit on health, then start chain services post-commit (compose profile split or a targeted `up` set). Store last-committed OS version beside chain data on `/data` (plan §appliance architecture).
[...]

PR #1673, opened 2026-09-03
fix(os): #1393 a failed data_migration update no longer leaves the /data floor raised over data it never touched

### LINKED ISSUE #1393: A failed data_migration update leaves the /data floor raised over data that was never migrated
A `data_migration` update that FAILS its health gate leaves the `/data` floor raised over data that was never migrated. The box falls back correctly, the chain data is genuinely untouched — and every later `os-update` below that floor is then refused with a message that states the data *was* migrated. The two escapes it offers are a factory reset and a backup restore, both destructive, both answering a false premise.

## The sequence

1. `os_install_bundle` raises the floor **at install time**, before the migration runs (`pithead:3427-3430`). The comment says so and calls it conservative — "the only floor pithead can guarantee without executing the migration itself". That is the right call in isolation.
2. The migration hold (#851) then guarantees the migration does **not** run before the slot commits: `pithead-boot` sets `PITHEAD_HOLD_CHAIN=1`, and `up` starts every compose service except `REVENUE_CHAIN_CONTAINERS` (`pithead:358-370`, list at `pithead:4816`).
3. The slot fails its health gate, stays uncommitted, and the machine falls back.
4. On the fallback boot the marker's version no longer matches the running version, so the hold is correctly ignored (`pithead:3255-3261`, `os/overlay/pithead-boot:236-239`). The chain data was never opened. All of this is the design working.
5. **But nothing lowers the floor.** `os_raise_data_floor` is the only writer, it has exactly one call site (`pithead:3429`), it only ever raises, and no code path anywhere removes `.os-data-floor`.

## What the operator sees afterwards

The box is running the old OS with a floor above it. `os_update_version_guard` (`pithead:3279-3294`) refuses any bundle below the floor **with no override**, saying:

> Refusing: /data was migrated forward and now needs OS >= X to read; this bundle is 'Y'. Installing it would strand the chain data. Recover with a factory reset (loses the chain) or by restoring a backup taken on X or newer.

`/data` was not migrated forward. The migration hold is the reason it was not.

## How bad

Not a hard strand: installing the floor version or anything above it still works, so the forward path is open. It bites when **the bundle that failed IS the floor version** — the operator then has no route below it and has to wait for a higher fix release, with the box telling them their chain data is migrated when it is not. The stated recoveries cost the chains, the wallets, the Tor onion keys and days of resync.

## Why the fix looks small

The state is distinguishable at exactly the point that already inspects it. On a fallback boot the marker is **present and version-mismatched**; after a successful migrate-and-commit it is **absent** (`pithead-boot` removes it at commit). So the boot path can tell "fell back, migration never ran" from "migrated and committed" without new state — and the first of those is the case where the floor can be cleared.

Not proposing an implementation here; the sequencing question (clear it in `pithead-boot`, or refuse-and-explain in the guard) is worth a decision rather than a patch.

## What I did not do

Read-not-run: this is derived from the source at the tip, not reproduced on hardware. Reproducing it needs a `data_migration` bundle that deliberately fails its health gate, which is a bench item. The individual mechanisms it composes are each covered by tier-4 today; the composition is not.

Found while settling a question for the Tari 5.6.0 work, which is the first release that would set this flag for real.

VijitSingh97:
## What this fixes (#1393)

`os-update` raises the `/data` floor at install time, and the migration itself waits for the slot to commit (#851). So a `data_migration` bundle that fails its boot gate falls back with the chain data untouched and the floor already raised over it — and every later `os-update` below that floor was refused with "/data was migrated forward" stated as fact, naming a factory reset as the way out. When the failed bundle IS the floor version there was no route below it at all.

Ruled on the issue (controller w86, restore-from-record, not clear and not refuse-only — https://github.com/p2pool-starter-stack/pithead/issues/1393#issuecomment-5522009718). Three parts, shipped together:

1. **The raise records the floor it replaced.** `os_raise_data_floor` writes the previous floor, or `none`, to `.os-data-floor.prev` BEFORE raising, and writes it even when the raise is a no-op, so every migrating install leaves a record. A crash between the two leaves a record equal to the live floor, which restores to a no-op.
2. **The fallback boot puts the floor back — from a record, and only from a record.** `pithead-boot`'s `restore_data_floor_after_fallback` runs in the else-branch of the hold decision (marker present, naming another version): floor restored from `.prev` (removed on `none`), marker and record consumed, one console line. **With no record the floor stays** — a floor is never lowered without one; a box whose failed install predates this fix has none. The stale marker goes either way. The commit-side removal now consumes the record too.
3. **The guard tells the truth.** A floor above the running version cannot be what it claims (a slot that could not read `/data` would not be running). `os_update_version_guard` refuses a bundle below such a floor with both readings — a migrating update that fell back before its migration ran, or a slot installed outside pithead below migrated data — says the floor version or newer installs, and names no reset or restore. It lowers nothing itself. Until the OLD slot carries part 2, this is the operator's only honest message, which is why it ships with part 2 rather than instead of it.

**Plus the two sibling floor controls the fix retired** (ruling + grant 2, https://github.com/p2pool-starter-stack/pithead/issues/1393#issuecomment-5522315898): both pre-existing below-floor rows put the floor ABOVE the running version — the exact state part 3 now refuses with the true premise — so their "strand the chain data" needle went red. Ruled: the guard message stays; the rows become honest-floor fixtures (floor at or below the running version, bundle below it), no line added, both files at their `file-budget.tsv` ceilings (457/457, 454/454).

- `test-appliance-os-update.sh`: running `1.17.0` -> `2.1.0` on the floor-2 row (floor 2.0.0, bundle 1.17.0). One token. The row now also exercises a real downgrade under `--allow-downgrade`.
- `test-appliance-os-update-verbs.sh`: **not one token, disclosed.** The dashboard-door fixture's bundle is 9.9.9 and its running version 1.3.1, so no floor can sit at or below running AND above the bundle. The row swaps `VERSION` to 10.0.0 for its duration with floor 10.0.0 and restores 1.3.1 — the file's existing equality-row pattern — folded onto the two existing lines as `&&` lists (`shfmt` splits a `;` list onto two lines, which the ceiling refuses). The verify verb reaches the shared guard before its equality and tag checks, so the floor refusal is what fires (read at `47-os-update-helpers.sh`, guard call before the `already on` check). The alternative — keep the 99.0.0 fixture and re-point the needle at the new message — was one token but would have kept a #1393-state fixture under a comment that says "replay below the floor"; rejected for that reason.

## What was RUN

- **New domain file `tests/stack/test-appliance-data-floor.sh`: 25 passed, 0 failed** standalone against the rebuilt `pithead` and the sourced `pithead-boot`. It first ran against the UNFIXED program by accident (the rebuild had silently not happened) and went 13/12 red on exactly the raise and guard rows — a fired control I did not plan.
- **The three domains standalone at the PR head:** `test-appliance-os-update.sh` 83/0, `test-appliance-os-update-verbs.sh` 85/0, `test-appliance-data-floor.sh` 25/0. Before the fixture edit the two siblings read 82/1 and 84/1 standalone, each red on exactly the edited row (the standalone instrument shown able to go red before it was trusted green).
- **Full tier-1 suite in the worktree at the PR head:** **3137 passed, 0 failed, rc 0** (detached run at head `cb468237`, settled by its rc file; log dir `1393-20260903T074749Z`). The previous run at the pre-fixture head read 3135/2 on exactly the two rows edited here.
- **Mutation battery, five mutants, each applied (pattern matched exactly once) and each killing a distinct row set, then reverted and the file re-proven clean** — run at the previous head; the slice and `pithead-boot` are byte-identical between that head and this one (`git diff --stat` between the two names only the two fixture files), so the battery stands:

| mutant | rows killed |
|---|---|
| m1: drop the `.prev` write | 4 (all four record rows) |
| m2: drop the fail-safe (`no record` → treat as `none`) | 2 (NO record, EMPTY record) |
| m3: drop the floor-above-running branch | 7 (all guard rows) |
| m4: commit-side `rm` keeps `.prev` | 1 (the static commit row) |
| m5: write `.prev` AFTER the raise | 1 (the unwritable-floor ordering row) |

- Gates re-run at the PR head, all rc 0: `lint-file-budget`, `lint-pithead-parity` (52 slices), `lint-operator-strings`, `lint-topology`, `lint-docs-voice`; `tests/inventory.sh` lists the new file with 4 sections; `shellcheck --severity=warning` (0.11.0) clean on `pithead-boot`, the new test file and both edited fixtures; `shfmt -i 4` clean on every changed shell file (the one `shfmt` diff in `pithead-boot`, lines 141-147, is pre-existing at the base under local shfmt 3.8.0 and not touched here). Max comment-line width of both fixture files unchanged from base (101 and 100 chars).

## What is NOT proven here

- **Tier 4 is its own item, #1672**, filed under the ruling: the composition (real migrating bundle, real failed gate, real A/B fallback running the previous slot's `pithead-boot`) has not run on the KVM battery. Strategy row says "added — unverified".
- **The two boot call sites are asserted by the script's TEXT**, not driven: they sit below `pithead-boot`'s sourcing boundary (the #1265 shape). The function itself is driven with files.
- An EMPTY `.prev` is treated as no record and left in place (the floor stays); the next migrating install overwrites it. Chosen over deleting it so the fail-safe branch touches nothing but the marker.
- The controller's non-author pre-pass (https://github.com/p2pool-starter-stack/pithead/issues/1393#issuecomment-5522291695) was taken at the PREVIOUS head and is void by its own terms now that the head moved; its re-run list (the new domain and the two siblings standalone, then the suite) is what the section above reports.

## Over-engineering pass (by hand — the ponytail gate cannot see this worktree's branch from the lane's cwd)

19 non-comment shell lines added across the slice and the boot script; 0 lines added in the two fixture files. Alternatives considered: (a) clear the floor on fallback — REFUSED on the issue, loses an earlier migration's floor; (b) raise the floor at commit time instead of install — needs the bundle's floor carried in the marker and moves the raise into the boot script, larger and it changes #851's proven ordering; (c) refuse-and-explain only — leaves the failed-bundle-is-the-floor case with no route. `pithead-boot` sits at 400/400 against the unbudgeted target; two comment blocks were compressed to keep it there rather than open a tsv row, so the next touch of that file plans its own room.

## Shared and granted files

- `lib/pithead/15-os-update.sh` and `tests/stack/test-appliance-data-floor.sh`: per-item grant for #1393 in `roles/appliance.paths` (controller w86), expiring on this merge.
- `tests/stack/test-appliance-os-update.sh` and `tests/stack/test-appliance-os-update-verbs.sh`: grant 2 for #1393 (controller w87), expiring on this merge. Guard readback taken with each staged alone (accepted) and an ungranted probe (refused).
- Root `pithead`: REGENERATED by `scripts/build-pithead.sh` for parity, never hand-edited (shared file, live line).
- `tests/stack/run.sh`: the minimal three-line registration stanza after `test-appliance-os-update-verbs.sh`. The tests lane is not running (`ListAgents`), so it could not be messaged; no open PR touches `run.sh`, the slice, or `pithead-boot` (checked at creation).
- `docs/`: shared; `docs/dev/appliance-release.md` (the floor paragraph), `docs/operations.md` (the `os-update` row) and the strategy row.

Closing keywords are inert on `develop-v2`; the controller closes #1393 by hand.


VijitSingh97:
## Non-author pass at head `cb468237` (controller w87, `claude-fable-5-1` by pin): **APPROVED — merge on the suite rc and the draft lift**

The pre-pass on the issue (`issuecomment-5522291695`) was void at the moved head by its own terms. Re-derived by me at THIS head, in a detached worktree:

| check | my result |
|---|---|
| diff between the two heads | ONLY the two granted fixtures, 5/5 lines; `15-os-update.sh`, `pithead-boot`, `pithead` and the new domain untouched, so the five mutation controls and the liveness probe from the pre-pass stand |
| `test-appliance-os-update.sh` · `-verbs.sh` line counts | 457 · 454 — at their ceilings, no line added |
| three domains standalone (`lib.sh` + file) | os-update 83/0 · verbs 85/0 · data-floor 25/0 |
| ruling 8a, both legs vs tip `7dbe46eb` (merge-base IS the tip): `make lint-file-budget` · `make lint-pithead-parity` | rc 0 / rc 0 on each; `merge-tree` rc 0 |

**The verbs fixture shape is ACCEPTED, and the lane was right to refuse the one-token form.** Its running version is 1.3.1 and its bundle 9.9.9, so no floor can sit at or below running AND above the bundle; re-pointing the needle at the new message would have kept a #1393-state fixture under a comment that says "replay below the floor". The VERSION swap to 10.0.0 with the floor at 10.0.0 is the honest-floor shape (floor AT running, bundle below it, the original migrated-forward refusal fires), it reuses the file's own equality-row pattern, and the swap is restored on the same line. Grant 2's "one token" was my sizing, not the criterion; the criterion was honest-floor with no line added, and both files meet it.

**Merge condition, read at the moment of merge:** the suite rc file in the dir named by `~/appliance-suite-logs/1393.current` reads 0 at THIS head, SUITE_RESULT filled, draft lifted, CI green at `cb468237`. Then squash-merge, `postmerge.sh`, expire BOTH #1393 grant blocks at the tail of `roles/appliance.paths`, close #1393 by hand, #1672 stays open. A further head move voids this pass again.

Issue #1664, opened 2026-09-03
tests(os): KVM pre-flight refuses the 16 GiB guest when the host cannot back it — ship the proven bench commits


VijitSingh97:
## What exists

The #1059 re-proof battery on 2026-09-03 ran with a KVM pre-flight that is proven but unpushed: `kvm_preflight` in `tests/os/failure-evidence.sh` reads `MemAvailable` from `/proc/meminfo`, prints it at every guest boot, and refuses to boot the 16 GiB guest under a 20,480 MiB bar (`PITHEAD_KVM_MIN_AVAIL_MB` overrides). `tests/os/run.sh` calls it at all nine `virt-install` sites. It fired both ways on the bench (refused under the bar, passed above it) and printed at all seven boots of run 2, 23.6 to 23.9 GB available.

## Why it ships

The KVM host also hosts the fleet. Two host hangs are recorded on 2026-09-02/03 (`journalctl --list-boots`): one at 19:15Z with no guest and 22 GB available, unexplained; one at 02:56Z with the guest up, memory inferred (the earlier "0.2 GB free" reading was `MemFree`; `MemAvailable` was 12.4 GB at that sample). A refused boot is a finding; a hang is a lost box and the operator's hands.

## The line budget (controller ruling, #1059 close comment)

No ceiling moves. `tests/os/run.sh` is at 3437/3437 and `tests/os/failure-evidence.sh` at 567/567. Either pay for the lines with restating-comment deletions listed verbatim in the PR body, or put the function in a new under-target file that `run.sh` sources. State each file's shave count; a third shave splits.

## Not done

Nothing beyond the two bench commits on the appliance lane's `bench/1059-reproof` branch. Whether the 20,480 MiB bar is the right number is not settled by two hangs; the bar is a floor the operator can raise.

VijitSingh97:
**Shipped in PR #1665, squash-merged to `develop-v2` as `04aef0d2` (controller, closing by hand — closing keywords do not fire on `develop-v2`).** Verdict on the PR: MERGE at head `4e08e3d6`, non-author pass. Post-merge protocol ran in full: patch identity between merge-base..head and old-tip..new-tip byte-identical with a fired control, remote branch deleted with an empty readback against a live-branch control, budget gate and parity green at the new tip. The tip had moved once under the PR (#1660 landed first); the gate and parity were re-run at both the branch head and the merged tree against that tip before merging, rc 0 on both legs.

Issue #1662, opened 2026-09-03
tests(os): the #1091 restore live-state verdict cannot pass on the KVM battery — p2pool writes no stratum wallet before its first block template


VijitSingh97:
## What happened

The #1059 re-proof battery on 2026-09-03 was the first live execution of the restore leg's closing verdict from #1091 (PR #1247 shipped it with "Not run: any KVM phase"). It failed:

```
✗ restore leg: the stack is up but live state never carried a readable stratum wallet (got 'Unknown')
os harness: 41 passed, 1 failed
```

Every other restore-leg check passed on the same guest: the archive restored, `config.json` came back with the original wallet, the stack came up, the Tor identity survived. The check that failed is the one that reads `.stratum.wallet` out of `/api/state` for up to 180 s after the stack's containers appear (`tests/os/run.sh`, the block above `restore_live_state_verdict`).

## Why it cannot pass on the KVM battery

`/api/state`'s `.stratum` is p2pool's `local/stratum` stats file passed through unchanged (`collector/pools.py` `get_stratum_stats`, `web/views.py` `build_pool_network`, default `Unknown` when the key is absent). p2pool writes that file only once its stratum server is serving a block template, which needs a synced monerod. Observed on a dev stack with a synced chain: p2pool started at 04:17:33 and first wrote `local/stratum` at 04:20:20, after monerod came back up. A restored KVM guest boots a fresh monerod, and its own console says first-time sync "can take several hours to a few days". So on this battery the field is absent for the whole window, and the verdict fails by construction.

The tier-1 fixture test (`tests/stack/run.sh`) drives `restore_live_state_verdict` with canned strings and cannot see this: the logic is right, the premise that the field will exist is what fails.

Inferred, not proven: I did not read p2pool's source for the exact moment `local/stratum` is first written. The dev-stack timing and the dashboard's code path are the evidence.

## Proposed fix

Keep #1091's intent, which is a value sourced from the restored config appearing in the RUNNING stack rather than a re-read of the archive, but take it from something that exists before any chain syncs. The cheapest honest source is the running p2pool container's own command line:

```
podman exec p2pool sh -c 'tr "\0" "\n" </proc/1/cmdline' | grep -A1 -x -- --wallet
```

[...]

Issue #1349, opened 2026-08-24
restore leg's live-state verdict can never pass: p2pool needs a synced monerod, the restore target never has one


VijitSingh97:
## The assertion cannot pass on the machine it asserts about

The restore leg's closing verdict (#1091, `tests/os/restore-live-state-verdict.sh`) requires a
value sourced from the restored config to appear in **live** state — specifically
`/api/state`'s `.stratum.wallet`, which comes from p2pool's own stratum stats. The intent is
right: `config.json` on disk only proves the archive was unpacked.

But p2pool cannot produce stratum stats on a freshly restored machine, so the condition is
unreachable there.

Observed on the bench (`--phase install`, restore leg, guest kept for inspection):

```
p2pool   Exited (0)   (starting)

P2Pool  get_info RPC request to host 127.0.0.1:RPC 18081:ZMQ 18083 failed: Error (empty response)
P2Pool  monerod is not synchronized, trying again in 1 second
P2Pool  uv_run exited, result = 0

monerod  I Synced 35216/3746850 (0%, 3711634 left)
```

p2pool exits because monerod is not synchronized. monerod is at 0% because the restore target is
a **fresh 30G disk** and `stack_backup` excludes the blockchains by default — `--with-chains` is
never passed anywhere in `tests/os/run.sh`. So every restored machine starts an unsynced monerod,
p2pool always exits, and `.stratum.wallet` is always `Unknown`.

The harness waits 180s for it. A full Monero sync on a KVM guest over Tor is hours-to-days, so no
practical timeout closes this gap.

## Why it was not caught

`restore_live_state_verdict` is fixture-tested at tier 1 (`tests/stack/run.sh:4646`) against
**canned strings**. That proves the function's logic — which is correct — and cannot say anything
about whether the live precondition is achievable. The two layers together look like coverage
while nothing checks the part that fails.

## Why it matters beyond one red line

This leg now ends in a permanent failure, so the battery's `--phase install` cannot go green. A
standing false RED is worse than a missing check: it trains readers to skip the line, and it
masks any real regression that later lands in the same leg.

## Options, none of them free

1. **Assert something p2pool can report before sync**, or read the restored wallet from a live
   source that does not depend on a synced chain (the dashboard already surfaces the configured
   wallet distinct from p2pool's running one). Keeps a live-state check with teeth; needs a value
   that is genuinely sourced from the restored config rather than a re-read of it.
[...]

Issue #1091, opened 2026-08-17
Test cannot fail: the appliance restore assertion greps a file the restore just wrote, never live state


VijitSingh97:
`tests/os/run.sh:1800-1811` — the final state check after an appliance restore is a **grep of a file the restore just wrote**. No `podman ps`, no `/api/state`.

So it proves the archive was unpacked. It does not prove the machine is *running* what was restored — which is the only thing an operator cares about after a restore, and exactly the gap #971 closed on the other channel.

**The DIY channel already does this properly:** `tests/integration/run.sh:1177` reads `.pool.type` out of **live state** after restore + up.

**Fix (W14 in #1069):** mirror the DIY assertion on the appliance — after restore, bring the stack up and require a value that came from the restored config to appear in live state (`/api/state`, or a credential marker baked into a running container).

**Mutation proof:** restore the config but skip bringing the stack up, or leave a stale container running with the pre-restore config; the assertion must go red. It currently passes in both cases.

Part of #1069.

VijitSingh97:
Shipped in PR #1247 (base develop-v2 — closed by hand). The restore leg now requires the restored machine's LIVE state (p2pool's running stratum wallet via /api/state) to carry the restored wallet — not just the unpacked file; fixture-proven at tier 1 with the mutation run (podman-ps-only judging flips three cases green, caught). Live confirmation rides the next battery.

PR #1674, opened 2026-09-03
fix(tests/os): #1662 restore verdict reads the wallet off the p2pool container's argv, not stratum stats

VijitSingh97:
## What

The restore leg's closing verdict (#1091) read the restored wallet out of `/api/state`'s `.stratum.wallet`, which is p2pool's `local/stratum` stats file passed through. p2pool writes that file only once a synced monerod hands it a block template, and a restored KVM guest boots a fresh chain, so on the first live run of this verdict (the #1059 re-proof, 2026-09-03) the field stayed absent for the whole 180 s window and the leg failed by construction: `got 'Unknown'`, 41 passed / 1 failed with every other restore-leg check green.

This keeps #1091's intent (a value sourced from the restored config appearing in the RUNNING stack, not a re-read of the archive) and moves the source to something that exists the moment the container is created: the `--wallet` argument the stack's own start path rendered into the p2pool container, read with `podman inspect p2pool --format '{{json .Config.Cmd}}'` and picked out with `jq`.

## Why `inspect`, not the `podman exec … /proc/1/cmdline` the issue proposed

The #35 sync gate (`SYNC_GATE_CONTAINERS`, `dashboard/mining_dashboard/config/config.py`) STOPS p2pool and xmrig-proxy until monerod is synced. On a restored guest that is exactly the window this verdict polls in, so `podman exec p2pool` would hit a stopped container and fail the same way the stats file did. `podman inspect` reads the created container's `Config.Cmd` whether it is running or stopped, and the quadlet renders the wallet there (`os/quadlet/p2pool.container` `Exec=… --wallet <address> …`, one token each).

## Files

- `tests/os/run.sh` — ONE executable line replaced in place (the `curl | jq` becomes `_ssh podman inspect | jq`), and the 8-line comment above it reworded within its 8 lines. The file sits at its budget ceiling (3437 = 3437), so the change is line-neutral by construction; the budget gate is green at the branch head.
- `tests/os/restore-live-state-verdict.sh` — comments only. The function body, its three-way shape, its 3-arg signature and its verdict strings are byte-unchanged, because `tests/stack/test-appliance-identity-boot.sh` asserts those strings and is not this lane's path. `git diff -U0 | grep -v '^[-+]#'` over the file returns nothing.
- `docs/dev/testing-strategy.md` — the restore-leg row names the new source, says why the old one could not work, and the tier-4 column reads `source changed — unverified until the next battery run`.

## What was RUN

- `shellcheck --severity=warning` on the two shell files: rc 0. `shfmt -i 4 -d` on both: clean. `make lint-docs-voice`: OK. `make lint-file-budget`: OK at the branch head. `bash -n` on both.
- Positive control for the instrument, on a live docker p2pool of the dev stack (same `inspect` schema): `Config.Cmd` has `--wallet` at index 8 and the next token is 95 characters. Length only; the value was not printed anywhere.
- The `jq` expression under three inputs: an array carrying `--wallet` (returns the next token), an array without it (empty), `null` (empty), and empty stdin (rc 0, no output) — so an absent or not-yet-created container reads as "unreadable" and the 180 s poll continues, as before.

## What was NOT run

- Not run: any KVM phase. The #1651 battery at the tip was refused by the #1664 memory pre-flight this cycle (host 2 GB under the bar with six lanes resident) and is queued; this is the leg it will exercise. Until then the tier-4 row is marked unverified, as the doc says.
- Not run: `tests/stack/run.sh`. The only change under its reach is comments in the sourced verdict file; the function it fixture-tests is byte-identical.
- Inferred, not proven on podman itself: that podman's `Config.Cmd` carries the quadlet `Exec=` tokens the way docker's does for a compose command. podman's inspect is docker-schema-compatible and the quadlet writes `Exec=` as the container command; the battery run is the proof.

Three shapes were available. (1) A config-derived wallet field on `/api/state` — the dashboard lane's path, a new API surface for one harness read, rejected. (2) The issue's `podman exec … /proc/1/cmdline` — fails under the sync gate, above. (3) This: one line moved to `podman inspect`, no new function, no new file, no new fixture. The verdict function was deliberately left untouched so the tier-1 test outside this lane keeps proving the same discrimination.

#1349 describes the same defect and should be closed as a duplicate when this lands (controller closes issues by hand; closing keywords are inert on `develop-v2`).


VijitSingh97:
**Non-author pass (controller w90): PASS, merge when the three pending checks land.** Branch sits on the tip (merge-base = tip), so head tree = merged tree; both legs run anyway and agree.

**Re-derived by me, at the head:**
- The one executable line (`tests/os/run.sh:1918`). `_ssh` logs in as `root@$ip` (`run.sh:145-148`), so it sees the rootful quadlet containers; the same leg already reads `podman ps` through it (`:1909`) and `run.sh:2232` already runs `podman inspect p2pool` the same way, so the transport is the file's own proven pattern, not a new one. The quadlet's `Exec=` carries `--wallet <address>` as two bare tokens (`os/quadlet/p2pool.container:11`, same in `local/`), which is what `Config.Cmd` holds for a quadlet-generated `podman run`.
- The `jq` expression under five inputs, run by me: an array carrying `--wallet` followed by a value → that value; an array without it → empty; `null` → empty; empty stdin → empty, rc 0; `--wallet` as the LAST token → empty (the `// ""` covers an out-of-range index). Every degenerate case reads as "unreadable" and the 180 s poll continues, as the body says.
- `make lint-file-budget` rc 0 at head and at the merged tree, the five self-test controls fired; `make lint-pithead-parity` rc 0 both; `make lint-docs-voice` OK; `bash -n` clean on both shell files. `run.sh` line-neutral at its ceiling, confirmed by the gate rather than by counting.
- The verdict function is byte-unchanged (`git diff -U0` over the file shows comment lines only), so `tests/stack/test-appliance-identity-boot.sh:255-274` still proves the same three-way discrimination.

**Relayed, not re-derived:** shellcheck/shfmt at the head (the box is under a KVM battery and `lint-sh` is the OOM path; CI's "Shell tests" check covers it and is one of the three I wait on), and the docker positive control on `Config.Cmd` index 8.

**F1 — non-blocking, rider for the next PR that touches the pair.** The second failure string in `restore_live_state_verdict` still says *"never carried a readable stratum wallet"*, and the source is now p2pool's argv, not its stratum stats. The wording is stale on the fail path only; the pass string ("p2pool's own running config") is accurate. It is asserted by `tests/stack/test-appliance-identity-boot.sh`, outside this lane's paths, so leaving it was the right call for this PR; reword string and assertion together in one PR under the `tests/stack` grant.

**Noted for the verdict on the battery now running:** it runs at the tip WITHOUT this fix, so its install leg is expected to red on `got 'Unknown'` exactly as #1662 describes; the strategy row stays "unverified" until a `--phase install` run at a tip that carries this. #1349 closes as a duplicate when this lands (by hand; closing keywords are inert on `develop-v2`).

Issue #1670, opened 2026-09-03
tests/os/soak-probe.sh --self-test: the driver rows are skipped, not failed, when mktemp -d fails


VijitSingh97:
`self_test_driver()` opens with `tmp=$(mktemp -d) || return 1`, and `self_test` calls it bare, so if `mktemp` fails the nine driver rows never run and the summary still ends `... ok, 0 failed` with rc 0 — a shorter run that reads as green. The script's own rule for the soak is "a missing measurement is a FAIL, never a skip"; its self-test should hold itself to the same one.

Fix is one line: count the failure instead of returning, e.g. `tmp=$(mktemp -d) || { chk "driver: mktemp -d" 1 0; return; }`, so the summary reads `1 failed` and the exit code is 1.

Found on the non-author pass of #1667 at its merged head; non-blocking there, filed so it is not lost. Fold it into the next touch of `tests/os/soak-probe.sh` (appliance lane, `tests/os/` is in-lane).

VijitSingh97:
Milestone `v2 - appliance` (controller, ruling by the pinned top tier). Re-derived at the tip: `self_test` calls `self_test_driver` bare, the driver's `mktemp -d` failure `return 1`s, and the summary counts only `$f`, so a run where `/tmp` is unwritable prints `... ok, 0 failed` over nine rows that never ran. That is the class #1651 names for the KVM battery's static half: a check that cannot fail. The soak probe is the instrument for the #1652 GA gate, so its self-test belongs in the gate's own queue, not in the frozen one.

Routing, not a new work item: the fix rides on the next `tests/os/soak-probe.sh` touch by the appliance lane, exactly as the body says. Left unmilestoned, the operator's queue rule makes it invisible to that lane, and a one-line fix never lands.

VijitSingh97:
Closing by hand (controller w90): fixed by #1675, squash-merged to `develop-v2` as `9791a693`. A failed `mktemp -d` in `self_test_driver` is now a counted row (`driver: mktemp -d`), summary `1 failed`, rc 1. Control re-fired by the controller at base and head with `TMPDIR=/nonexistent`: base 20 ok / 0 failed rc 0, head 20 ok / 1 failed rc 1; the normal run stays 29 ok / 0 failed.

PR #1675, opened 2026-09-03
fix(tests/os): #1670 soak-probe --self-test counts a failed mktemp -d as a failed row, not a skipped set

VijitSingh97:
## What

`self_test_driver()` opened with `tmp=$(mktemp -d) || return 1` and `self_test` called it bare, so when `mktemp -d` failed the nine driver rows never ran and the summary still ended `0 failed` with rc 0: a shorter run that read as green. The soak's own rule is that a missing measurement is a FAIL, never a skip. The self-test now holds itself to it: the failure is counted as a row (`driver: mktemp -d`), the summary reads `1 failed`, and the exit code is 1. One statement changed plus a two-line comment; the file has no budget row (296 lines).

## What was RUN

Positive control on the UNFIXED file, `mktemp` forced to fail with `TMPDIR=/nonexistent`:

```
soak-probe self-test: 20 ok, 0 failed      rc=0
```

The same input on the fixed file:

```
  ✗ driver: mktemp -d: [1] wanted [0]
soak-probe self-test: 20 ok, 1 failed      rc=1
```

The normal run on the fixed file: `29 ok, 0 failed`, rc 0, unchanged from the base. So the nine driver rows are the 29 minus 20, as the issue said.

`shellcheck --severity=warning` rc 0, `shfmt -i 4 -d` clean, `bash -n` clean, `make lint-file-budget` OK.

## What was NOT run

No bench, no live soak: the change is inside `--self-test` only, and the probe's live path is untouched. No doc names the self-test's row count, so no doc changed.

The issue's one-line shape was taken as written. The alternative of having `self_test` check `self_test_driver`'s return value would leave the driver's own contract ("return 1 = skipped") in place and add a second site to keep honest; counting the failure where it happens removes the skip path instead of wrapping it.

### DIFF
--- a/tests/os/soak-probe.sh
+++ b/tests/os/soak-probe.sh
@@ -212,7 +212,12 @@ self_test() {
 # the same day=0, which is the shape the label and the write-once baseline exist for.
 self_test_driver() { # $1 = a canned reading that passes against itself
     local tmp out
-    tmp=$(mktemp -d) || return 1
+    # A scratch dir that cannot be made is a FAILED row, never a skipped set: bare `return` here
+    # dropped the nine driver rows and the summary still read `0 failed`, rc 0 (#1670).
+    tmp=$(mktemp -d) || {
+        chk "driver: mktemp -d" 1 0
+        return
+    }
     mkdir -p "$tmp/bin"
     printf '%s\n' '#!/usr/bin/env bash' 'cat >/dev/null' 'printf "%s\n" "${@: -1}" >>"$SOAK_STUB_ARGS"' 'cat "$SOAK_STUB_READING"' >"$tmp/bin/ssh"
     chmod +x "$tmp/bin/ssh"


VijitSingh97:
**Non-author pass (controller w90): PASS, merge when the three pending checks land.** Branch on the tip; head tree = merged tree, both legs run anyway.

**Re-derived by me:** the control, re-fired in throwaway worktrees at base and head with `TMPDIR=/nonexistent`: base `20 ok, 0 failed` rc 0; head `20 ok, 1 failed` rc 1 with the row `✗ driver: mktemp -d: [1] wanted [0]`; the normal run `29 ok, 0 failed` rc 0 at both, so the nine driver rows are unchanged and the only new row is the failure. `bash -n` clean; `make lint-file-budget` and `make lint-pithead-parity` rc 0 at head and merged tree; no budget row and none needed (299 lines at head against the 400 target — the body's "296" is two lines off in the right direction and changes nothing).

**Relayed:** shellcheck/shfmt (CI's "Shell tests" check covers it; the box is under a KVM battery and `lint-sh` is the OOM path).

Nothing to fix. #1670 closes by hand when this lands.

PR #1677, opened 2026-09-03
test(os): #1672 battery leg — a failed data_migration update gives the /data floor back on fallback (#1393, tier 4)

VijitSingh97:
## What (#1672, tier 4 for #1393)

A new leg at the end of the battery's `provision` phase, on the guest the migration leg just left committed (floor = the checkout's VERSION, chain up). It proves the composition the tier-1 rows in `tests/stack/test-appliance-data-floor.sh` cannot: a real migrating bundle whose boot gate fails, a real A/B fallback, and the previous slot's `pithead-boot` putting the `/data` floor back from the record the raise left. Then the negative control: the same sequence with the record deleted must leave the floor alone, and `os-update` must refuse a bundle below that floor with the failed-update premise and no reset named.

**How the gate is made to fail.** The failing bundle is stamped `99.0.0`, a version no release of this project will carry. Two consequences, both wanted: `os-update` treats it as a clean upgrade and raises the floor to it, and its slot boots deriving `STACK_VERSION` from that stamp, so `pithead up` hunts first-party image tags that were never published and cannot bring the stack up; `pithead-boot` leaves the slot uncommitted and reboots into the previous one. That is exactly the effect the leg-4 note above `_prev_patch_version` in `run.sh` warns about, used on purpose. The leg asserts the fallback by the failed boot's own journal line (`slot left uncommitted so A/B fallback stays armed`, read with `journalctl -b -1`), not by the mechanism. Alternatives I rejected are in the over-engineering pass below.

**Why a copied tree, not a worktree.** Every bundle is stamped with the checkout's VERSION by design (`mkbundle.sh`: "deliberately has no override"), and the battery tree is frozen while it runs. So the bundle is built from a `git archive` export of HEAD with `99.0.0` written into its `VERSION`. Never a `git worktree`: the battery runs as root, and a root-made commit leaves root-owned objects in the `.git` every worktree on the box shares. The export has no git, so `build-image.sh` stamps `BUILD_COMMIT` as `unknown-dirty` and `mkbundle.sh` needs `PITHEAD_STALE_TARBALL_OK=1`; `os-update` reads neither, only `[meta.pithead]`. It is signed with this tree's dev key (`os/rauc/certs`, copied in) so the keyring baked into the guest's slots accepts it. The bundle lands under `/tmp`, outside `os/rauc/build`, so `_build_bundle`'s `find | head -1` in a later phase can never pick it up.

## Assertions, in order

Positive case: floor before the leg is non-empty · after `os-update`: `floor|prev|marker` = `99.0.0|<floor0>|99.0.0` (issue step 2, before any boot) · the guest comes back on the slot carrying marker `vmig` with `booted slot committed` in its journal · the previous boot's journal carries the uncommitted line · `floor|prev|marker` = `<floor0>||` · the fallback boot's journal names `floor is back to <floor0>` · `os-update` of the migration leg's own bundle (at `floor0`) INSTALLS.

Negative control: second install of the same bundle arms the same state · the record is deleted before the reboot · fallback again · `floor|prev|marker` = `99.0.0||` (floor untouched, marker gone) · the journal does NOT claim a restore (an absence, controlled: the identical grep found the line in the positive case minutes earlier) · `os-update` of the `floor0` bundle is REFUSED with `a migrating update to 99.0.0 failed its gate and fell back before its migration ran` and the output names no reset or restore.

## Files

- `tests/os/data-floor-fallback-leg.sh` — new, 198 lines, sourced by `run.sh` like the verdict files; no budget row.
- `tests/os/run.sh` — two source lines and the one call at the end of `phase_provision`, paid for line-neutrally (3437 = 3437): the header's media/reset/all lines compressed from 10 to 8, and the `all)` dispatch comment from 4 to 3. **Disclosed: those header lines also said media, fault and reset were "opt-in" and "not in `all`". That has been false since #1064 (the `all)` arm runs all eight phases, and neither `phase_media` nor `phase_fault` gates on anything); the new wording says so.** Max comment width unchanged (the two pre-existing >100 lines, 8 and 523, are untouched).
- `tests/os/README.md` — the provision bullet describes the leg. `docs/dev/testing-strategy.md` — the #1393 row's tier-4 column names it, still `unverified until the next battery run`.

## What was RUN

- `shellcheck --severity=warning` on both shell files: rc 0. `shfmt -i 4 -d`: clean. `bash -n`: clean. `make lint-file-budget`: OK (run.sh at its ceiling). `make lint-docs-voice`: OK. `tests/inventory.sh`: rc 0.
- Control on the guest-side state probe's quoting: `_floor_state` run through a local `bash -c` stand-in for `_ssh` returned `1.19.3|none|99.0.0` and then `1.19.3||` after two of the files were removed — the two exact shapes the leg compares against.
- Control on the scratch build's non-docker steps: `git archive HEAD | tar -x` produced a tree with no `.git`; `BUILD_COMMIT` evaluates to `unknown-dirty` there, as the comment predicts; the dev certs dir exists on the battery tree (root-owned, which is why the battery's root context is what copies it).
- Read at source, not inferred: `os_raise_data_floor` writes `.prev` before raising and has no check that the minimum stays at or below the bundle's own stamp; the CLI `os-update` writes no in-flight flag (only `47-os-update-helpers.sh` does); `restore_data_floor_after_fallback` keys on marker ≠ VERSION; `fail_boot` reboots on the first failure and stays up on the second.

## What was NOT run — this is a DRAFT until it is

- **The leg itself has not executed on KVM.** The bench is under the #1651 battery. I will run `tests/os/run.sh --phase provision` on the bench as soon as the box is free and post the log here before marking ready. Two things only that run can settle: how long the `99.0.0` slot takes to give up on `pithead up` (the leg allows 30 min per fallback), and whether `journalctl -b -1` reads the failed boot on the guest (persistent journald, #1207).
- Not run: `tests/stack/run.sh` — nothing under its reach changed.

Shapes considered for failing the gate: (1) a build-time seam that bakes a broken dashboard image or a systemd drop-in into the failing slot — deterministic but adds product-image code for a harness need, and with the unpublished stamp the slot fails before either would matter; (2) the guest claiming an older version, leg 4's trick — cannot work here, because the fallback slot must run `up` afterwards and would hunt the older tags itself, failing the boot the leg needs to pass; (3) a `/data` poison removed during the failing boot — racy, and `/data` is shared with the fallback boot. (4) chosen: the unpublishable stamp alone, with the fallback asserted from the journal. A `git worktree` for the build was available and rejected for the root-owned-objects hazard above.

Same-version bundles cannot exercise this at all: the fallback decision is marker versus VERSION, so a bundle stamped like the slot it falls back to makes that slot hold the chain as if it were the migrating boot and never restores. That is the reason the version must differ, and it is recorded in the file header.


VijitSingh97:
## `--phase provision` at this head — the floor-fallback leg and its negative control are GREEN; the one red is #1684

**Run:** `sudo HOME=<builder home> tests/os/run.sh --phase provision` from a worktree checked out at this PR's head (`tree.txt`: tip = this head, dirty = 0), detached under the shellcheck-serialize flock, 14:43:21Z to 15:17:16Z on 2026-09-03. Evidence dir on the build host: `appliance-suite-logs/1672-provision-20260903T144220Z/` (`provision.log`, `provision.rc`, `leg.rc`, `tree.txt`).

**Result: `os harness: 47 passed, 1 failed`, rc 1.**

The rows this PR adds, verbatim from the log:

```
==> floor-fallback leg — a data_migration update that fails its gate must give the /data floor back (#1393)
  ✓ floor before the failing update: 1.20.0
  ✓ built a data_migration bundle stamped 99.0.0 (floor 99.0.0): os-floor-fallback-vfail.raucb
  ✓ install raised the floor to 99.0.0 and recorded 1.20.0 beside it (marker 99.0.0)
  ✓ the 99.0.0 slot never committed — the guest is back on the previous slot with its gate passed
  ✓ the 99.0.0 boot failed its gate and left the slot uncommitted (its journal says so)
  ✓ RESTORED: the floor is back to 1.20.0, the record and the marker are gone
  ✓ the fallback boot said so on the console (journal line names 1.20.0)
  ✓ a bundle at the restored floor (1.20.0) installs again through os-update
==> negative control — the same failed update with the record deleted must leave the floor alone
  ✓ second install armed the same state (floor 99.0.0, record 1.20.0, marker 99.0.0)
  ✓ the 99.0.0 slot never committed again — back on the previous slot
  ✓ FAIL-SAFE: with no record the floor stayed at 99.0.0; only the marker went
  ✓ the record-less fallback boot did not claim a restore
  ✓ os-update refuses a bundle below the stale floor with the failed-update premise
  ✓ the refusal names no reset and no restore

os harness: 47 passed, 1 failed
```

**The one red, `monerod never came back after the commit` (migration leg), is not this PR's change.** `os/` is byte-identical to the #1651 tip that passed that row an hour earlier; the cause is the post-commit chain-release race filed as #1684, proven from the guest's own journal read live during this run (the issue carries the excerpt). The fix is in flight on `fix/1684-release-retry`; #1684's close condition is that row green on a battery at a tip carrying the fix, so nothing here claims it.

**Not done here:** a second provision run at this head to show the red is timing-dependent (the contrast is #1651's green on the same `os/`, cited in #1684). No `--phase all`: this PR's rows live in `provision` only.

Marking ready for review on the strength of the floor rows.

VijitSingh97:

- Evidence dir `~/appliance-suite-logs/1672-provision-20260903T144220Z/` (predecessor cycle, same seat): `tree.txt` tip == this head, dirty 0; `leg.rc` = provision = 1; 47 ✓ / 1 ✗. The one ✗ is `monerod never came back after the commit` in the migration leg (the #1684 rows, being fixed by #1686), not a row this PR adds. All 14 new rows ✓, including the negative control and the controlled journal absence.
- Leg body read in full (198 lines). The positive/negative pair asserts on the guest's `floor|prev|marker` triple, the previous boot's journal (`-b -1`), and the refusal LINE, not on rc alone. The `git archive` export path never touches the shared `.git`.
- **Attack run and refuted:** the negative control leaves the guest's floor at `99.0.0` at the end of `phase_provision`; in an `all` run four phases follow. Each of `phase_rig`, `phase_media`, `phase_fault`, `phase_reset` calls `_build_image v1` + `_vm_boot_disk` on a fresh disk (read at head), so the poisoned floor dies with the provision guest and cannot red a later phase.
- Ruling 8a, both legs (this cycle, scratchpad `prelegs-1677.log`): at the branch head `make lint-file-budget` rc 0 (5 self-tests fired), parity rc 0; at the `merge-tree --write-tree` result (rc 0) the same, rc 0 / rc 0. Merge-base `80cd9c02` is two behind tip; run.sh 3437 = 3437 at its tsv row.
- Non-leg diffs: `run.sh` header compression discloses and corrects the stale "opt-in / not in `all`" claim (true since #1064); README and strategy rows updated.

**Not done by me:** the leg was not re-executed on KVM this cycle; the bench run is the appliance lane's, re-derived from its evidence dir by tip/dirty/rc/row-count, not re-run.

Closes nothing by keyword (base is `develop-v2`); #1672 is closed by hand on the issue against its four numbered steps.

PR #1678, opened 2026-09-03
fix(dashboard): #1265 the rollback banner names the check that held the boot gate

### LINKED ISSUE #1265: A good OS update can auto-revert: cert minted before addresses settle, doctor checks after — and nothing re-mints once the gate starts failing

VijitSingh97:
Closes #1265 (the dashboard half — see "Not done" below for what stays open).

## The defect

A good OS update can roll back. The boot gate mints the dashboard certificate at render, **before** `up`; doctor re-checks coverage inside the gate loop, **after** `up`. An address that arrives in between — a SLAAC/ULA prefix is exactly this shape — makes the settled name list a superset of the minted one, doctor fails coverage on every round, and the box reverts a healthy payload.

#1671 fixed the boot half and began recording doctor's failing messages in the verdict as `blocking`. The dashboard was still rendering only *"failed its health checks"*, which reads as a bad release and for this defect is simply wrong: retrying reproduces the same race, so the operator has no way to tell a bad release from an environmental check failure.

## What this does

The rollback banner names the cause — the first blocking check, plus a count of the rest. `pithead-boot` names the first on the console too (`jq -r '.blocking[0]'`), so the console and the dashboard agree rather than telling two stories.

`blocking` is host JSON and is **absent on any OS image older than #1671**, so the clause is strictly additive: with nothing named, the sentence is byte-for-byte what it was. That is the case most likely to regress, so a test pins it as an exact string rather than a regex.

Hardening on the sanitizer, since these strings are host-written and reach an operator through this page: non-list shapes, non-string entries, blank entries and punctuation-only entries all degrade to no clause; a usable message beside junk is still named; one message is bounded at `MAX_BLOCKING_LEN` (120) so a runaway string cannot become the whole banner.

## Shape, and why it is a new module

`osupdate.mjs` sits at its recorded ceiling of **415 with zero headroom**, and the change does not fit inline. Moving the verdict-text concern into a new `web/static/osverdict.mjs` takes that file to **407**; the new module is **45** lines, under the 400 target, so it carries no `file-budget.tsv` row. `verdictText` was imported only by its own test, so the move is contained — `components.mjs` imports `OsVerdictBanner`, which stays put, and `wizard.mjs`'s same-named `verdictText` is an unrelated local `const`, checked rather than assumed.

## What was RUN

- `node --test dashboard/tests/frontend/*.test.mjs` — **502 pass / 0 fail**, twice (once after the import reorder below).
- **Test-name set diff, before vs after:** 273 -> 278. Exactly one name left `osupdate.test.mjs` and the *identical* name appears in `osverdict.test.mjs` — a move, not a deletion — plus 5 new rows. The instrument was given a seeded removal first and reported it, so the clean reading is not an unfired one.
- `make lint-js` rc 0, `lint-md` rc 0, `lint-docs-voice` rc 0, `lint-operator-strings` rc 0, `lint-topology` rc 0, `lint-file-budget` rc 0 — run individually, because that CI job runs six targets past the five in its name and `make` stops at the first failure.
- `lint-js` **failed first** on an unsorted import block and was fixed; recording it because a stale green here would have reddened CI.
- Line counts re-measured at the pushed head with the gate's own `awk 'END{print NR}'`: 45 / 407 / 383 / 83 against the 415 ceiling. (41 / 79 at the first push; the review fix below moved both.)
- `add_static("/static", ...)` serves the whole directory, so the new module is served — checked, because an allowlist would have 404'd it in a browser with every test still green.

**Patch coverage:** `MEASURED` in `patch-coverage.sh` is `dashboard/mining_dashboard/*.py`, and this diff changes no Python, so the gate takes its loud not-applicable pass. No Python changed, so `read_os_update_state` needed no edit — it already returns the host dict whole, with no key allowlist to widen.

## Fixed in review (F1)

`blockingCause` judged emptiness on the raw entry and normalised afterwards, so an entry that is
truthy going in and empty coming out — `"..."` is the whole class, since the map strips trailing
periods — passed the filter and then took the named slot. Measured at the previous head:
`["..."]` gave `" Blocked by: ."` and `["...", CERT]` gave `" Blocked by:  (+1 more)."`, the junk
hiding the real cause. Map first, then drop the empties; the change is one-directional, since an
entry with an empty `trim()` also normalises empty, so nothing previously dropped is now kept.
Both shapes are rows in that test now, shown **red at the previous head** before green — and
because Node's assert short-circuits, the `["...", CERT]` row was measured directly against the
unfixed module rather than assumed. Suite still **502 / 0** (assertions inside an existing test,
not new tests).

## The over-engineering pass, done on merits

Four things considered and rejected, with the reason:

1. **Compressing `osupdate.mjs` by ~9 lines to fit inline.** Rejected: the only lines available are house-voice comments carrying the reasoning for the control flow. Paying a budget line by gutting an explanation is the trade this lane is told not to make.
2. **Rendering the full blocking list.** Rejected: one named cause an operator can act on beats a wall of them, and naming the first is what the host's own console line does.
3. **Asserting escaping through `renderToString`.** Rejected and named as a gap instead — `tests/frontend/helpers/render.mjs` says outright that it does not escape, so the assertion would have passed for a reason unrelated to the thing it names. Preact escapes text interpolations in the browser; this harness cannot witness it.
4. **Keeping `blockingCause` private.** Rejected: it is half the module's contract, not an internal, and its degradation behaviour is the part worth pinning directly.

## Not done, deliberately

- The issue's **candidate 2** (doctor counting cert drift as `warn` for the commit decision) and **candidate 3** (`pithead apply` reaching the mint on an unchanged config) live in `lib/pithead/` — not this lane's paths.
- The **cosmetic post-commit operator-rollback verdict** still reads "updated" while the box runs the old version. Host-side, untouched.
- **Nothing here is proven on hardware.** The KVM battery still does not stage a late-arriving address, so #1671's strategy row stays "added — unverified" and this half inherits that.
- `docs/` is in `_shared.paths` and is disclosed here rather than reached for silently; no other lane's path is touched.

### DIFF
--- a/dashboard/mining_dashboard/web/static/osupdate.mjs
+++ b/dashboard/mining_dashboard/web/static/osupdate.mjs
@@ -12,6 +12,7 @@
 // dashboard answers again; the post-reboot verdict (updated / rolled back) arrives via the
 // host-persisted state in state.os_update and renders as a banner.
 
+import { verdictText } from "./osverdict.mjs";
 import { Component, html } from "./preact.mjs";
 
 const CONTROL_HEADERS = { "Content-Type": "application/json", "X-Pithead-Control": "1" };
@@ -89,15 +90,6 @@ export async function runOsDownload(
 // Pin it client-side: only the public GitHub release page ever renders as a link.
 export const releaseNotesHref = (u) =>
   typeof u === "string" && u.startsWith("https://github.com/") ? u : null;
-
-// One verdict line for the banner. Pure — the render just wraps it.
-export function verdictText(verdict) {
-  if (!verdict || !verdict.outcome) return null;
-  if (verdict.outcome === "updated") return `System updated to v${verdict.to}.`;
-  if (verdict.outcome === "rolled_back")
-    return `The update to v${verdict.to} failed its health checks and was rolled back automatically — still on v${verdict.from}.`;
-  return null;
-}
 
 // Post-reboot verdict banner, rendered by App beside the disconnected banner. Dismiss hides it
 // for this browser session (keyed on the verdict's timestamp); the host clears the verdict
--- a/dashboard/mining_dashboard/web/static/osverdict.mjs
+++ b/dashboard/mining_dashboard/web/static/osverdict.mjs
@@ -1,0 +1,41 @@
+// The post-reboot OS-update verdict, turned into one sentence for the banner (#1051, #1265).
+//
+// Its own module rather than more of osupdate.mjs: that file is at its recorded budget ceiling
+// with no headroom, and this is a self-contained concern — host verdict in, display text out,
+// with no fetching, sequencing or state of its own.
+//
+// The rollback sentence used to say only that health checks failed, which reads as "bad release"
+// and is usually wrong. The boot gate rolls a GOOD update back when doctor fails a check the
+// update did not cause — #1265's cert-coverage race is exactly that shape: an address arrives
+// between render's mint and doctor's check, and the box reverts a healthy payload. pithead-boot
+// (#1671) now copies doctor's failing messages into the verdict as `blocking`, so the banner can
+// name what actually held the gate instead of implying the release was at fault.
+
+// One doctor message is a sentence, not a paragraph. The host writes them, but they reach an
+// operator through this page, so bound them: a runaway message must not become the whole banner.
+// Display-only — the full list stays in the state file for anyone reading the host.
+export const MAX_BLOCKING_LEN = 120;
+
+// The checks that held the gate, as a trailing clause, or "" when the host named none. An OS
+// image older than #1671 omits `blocking` altogether, so the caller's sentence has to read
+// correctly without it — absence here is "we were not told", never "nothing was wrong".
+export function blockingCause(verdict) {
+  const held = (verdict && Array.isArray(verdict.blocking) ? verdict.blocking : [])
+    .filter((m) => typeof m === "string" && m.trim())
+    .map((m) => m.trim().replace(/\.+$/, "").slice(0, MAX_BLOCKING_LEN));
+  if (!held.length) return "";
+  // Name the first and count the rest: the gate stops at the first failing check, and one named
+  // cause an operator can act on beats a wall of them.
+  return ` Blocked by: ${held[0]}${held.length > 1 ? ` (+${held.length - 1} more)` : ""}.`;
+}
+
+// One verdict line for the banner. Pure — the render just wraps it.
+export function verdictText(verdict) {
+  if (!verdict || !verdict.outcome) return null;
+  if (verdict.outcome === "updated") return `System updated to v${verdict.to}.`;
+  if (verdict.outcome !== "rolled_back") return null;
+  return (
+    `The update to v${verdict.to} failed its health checks and was rolled back automatically` +
+    ` — still on v${verdict.from}.${blockingCause(verdict)}`
+  );
+}
--- a/dashboard/mining_dashboard/web/static/osverdict.mjs
+++ b/dashboard/mining_dashboard/web/static/osverdict.mjs
@@ -20,9 +20,13 @@ export const MAX_BLOCKING_LEN = 120;
 // image older than #1671 omits `blocking` altogether, so the caller's sentence has to read
 // correctly without it — absence here is "we were not told", never "nothing was wrong".
 export function blockingCause(verdict) {
+  // Judge emptiness AFTER normalising, not before: a punctuation-only message like "..." is
+  // truthy on the way in and empty on the way out, and filtering first let it take the named
+  // slot and hide the real cause behind " Blocked by:  (+1 more).".
   const held = (verdict && Array.isArray(verdict.blocking) ? verdict.blocking : [])
-    .filter((m) => typeof m === "string" && m.trim())
-    .map((m) => m.trim().replace(/\.+$/, "").slice(0, MAX_BLOCKING_LEN));
+    .filter((m) => typeof m === "string")
+    .map((m) => m.trim().replace(/\.+$/, "").slice(0, MAX_BLOCKING_LEN))
+    .filter((m) => m);
   if (!held.length) return "";
   // Name the first and count the rest: the gate stops at the first failing check, and one named
   // cause an operator can act on beats a wall of them.
--- a/dashboard/tests/frontend/osupdate.test.mjs
+++ b/dashboard/tests/frontend/osupdate.test.mjs
@@ -15,7 +15,6 @@ import {
   pollOsResult,
   releaseNotesHref,
   runOsDownload,
-  verdictText,
 } from "../../mining_dashboard/web/static/osupdate.mjs";
 import { renderToString } from "./helpers/render.mjs";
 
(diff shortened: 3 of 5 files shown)


VijitSingh97:
**Controller pass (non-author, fable tier) — CHANGES REQUESTED on one line; everything else holds.**

Proven by me at head `31d79fd3` (merge-base `49b4b9ea`, tip `9791a693`):
- Ruling 8a, both legs: budget gate rc 0 (11 self-test lines, zero real FAIL) and parity rc 0 at the branch head AND at the merged tree; `merge-tree` rc 0. `osupdate.mjs` 407/415, `osverdict.mjs` 41 with no row — the gate's own `awk`.
- `node --test dashboard/tests/frontend/*.test.mjs` in a fresh detached worktree at the head: **502 pass / 0 fail**.
- The no-`blocking` sentence is byte-identical to the pre-change literal (read from the diff, and the exact-string test pins it).

**F1 — BLOCKING, one line.** `blockingCause` filters BEFORE it strips trailing periods, so an entry that is only periods (or periods and spaces) survives the filter and becomes empty after the map. Run at the head:

```
blockingCause({ blocking: ["..."] })                 -> " Blocked by: ."
blockingCause({ blocking: ["...", "real cause"] })   -> " Blocked by:  (+1 more)."
```

The second is the one that matters: a junk entry HIDES the usable cause behind it, which is exactly the property the body ("a usable message beside junk is still named") and the test title ("every garbled shape degrades to no clause") claim. Fix: map first, then filter for non-empty (or filter again after the map), and add `["..."]` and `["...", CERT]` rows to the garbled-shape test — show the rows fail at the current head before they pass. Ruling 2 applies: the claim lives in a test name and a module comment, in-repo, so it earns the fix rather than a verdict footnote.

**Merge-record nit, no push needed:** the body opens `Closes #1265` and then says the issue stays open on candidates 2–3 and the hardware row. The keyword does not fire on this base anyway, and #1265 STAYS OPEN by ruling; I will note the dashboard half on the issue at merge.

**Accepted as named:** escaping not asserted through the non-escaping render probe — the right call; the text interpolation is the existing shape and not new to this diff.

**RELAYED, not re-run by me:** the individual lint targets and the test-name set diff with its seeded control (CI's lint job is green, which covers biome and markdown).

On the fixup push I re-run both 8a legs and the suite at the new head, then merge on the checks read at that moment. The shell-tests job was still IN_PROGRESS at this pass.

VijitSingh97:
**Controller re-pass at the fixup head `68c699fe` — F1 is fixed; APPROVED, merge follows on the checks read at the moment of merge.**

Proven by me: the probe that reddened `31d79fd3` now reads `blockingCause({blocking:["..."]})` → `""`, `["...", "real cause"]` → ` Blocked by: real cause.`, `[" . "]` → `""`. Ruling 8a both legs against the moved tip `80cd9c02` (merge-base `49b4b9ea`): budget gate rc 0 and parity rc 0 at the head and at the merged tree, `merge-tree` rc 0. `node --test dashboard/tests/frontend/*.test.mjs` in a fresh detached worktree: 502 pass / 0 fail. `osverdict.mjs` 45, `osupdate.mjs` 407/415 by the gate's own `awk`. The fixup touches two files, +10/−2.

The author's point that Node's `assert` short-circuits, so only the first added row ever reddened in the red run, is right and correctly handled by measuring the second row against the unfixed module directly.

Left out on purpose: the strip-then-slice order can emit a double period when the 120-char bound lands just after one. Cosmetic; not worth another push on this PR.

Merge-record: `Closes #1265` does not fire on this base; #1265 STAYS OPEN by ruling and gets a note naming this half at merge.

VijitSingh97:
**MERGED as `9f417d14` (squash) — checks read in the same command as the merge: 0 pending, 0 failed, head still `68c699fe`, tip still `80cd9c02` (the tip my re-pass legs ran against, so those legs stand).**

Post-merge protocol, proven by me:
- Patch identity: `diff merge-base..head` vs `diff old-tip..new-tip` byte-identical (12,110 B); control against an unrelated patch fired (differs, 4,799 B).
- Remote branch `fix/1265-dashboard-render-verdict-blocking`: 95 B before, deleted by hand, 0 B readback; live-branch control reads 89 B. (`--delete-branch` is not used here — it has failed fourteen times against this box's worktrees.)
- At the new tip: `make lint-file-budget` rc 0, `make lint-pithead-parity` rc 0.

#1265 stays OPEN: this is the dashboard half only. The `lib/pithead/` remainder (candidates 2 and 3) is the appliance lane's queue item 1.

Issue #1442, opened 2026-08-28
trivyignore-watch reads only develop-v2's .trivyignore — develop's mutes are unwatched (CVE-2026-45447 is a live dead-mute candidate)


VijitSingh97:
Found during #1438's review (credit: the tests lane's re-review). Same class as #1425: a watcher whose covered set silently excludes one lane.

## The gap

`trivyignore-watch.yml` runs from the default branch on `schedule:` but checks out `ref: develop-v2` (hard-coded, with a comment calling `.trivyignore` "shared"), and `scripts/trivyignore-watch.sh` reads `$ROOT/.trivyignore` from that checkout. So **only develop-v2's ignore file is ever watched**. `--check-parity` is trivy-*version* parity between the gate workflows and the watch, not ignore-file parity between lanes, and it is not wired into CI on either branch.

The "shared file" premise the workflow comment relies on is no longer true: the two lanes' files have deliberately diverged (develop-v2 carries the appliance Go block; develop does not).

## Live instance, not hypothetical

develop's `.trivyignore` still mutes **CVE-2026-45447**. develop-v2 deleted that entry as obsolete on 2026-08-24 (`a2382d8`, "drop the obsolete openssl mute CVE-2026-45447"). If it is equally obsolete on develop — likely, since develop's dashboard base now ships openssl 3.5.6-1~deb13u2, a deb13u2 build — then develop has carried a dead mute for four days that the watch is structurally unable to report, while the file looks correctly maintained. A dead mute is exactly the blind spot #1174 built this watch to close.

## What a fix needs to decide

Either the watch scans each lane's own file against each lane's own images (two checkouts, per-lane report sections), or lane parity of the file becomes a checked invariant somewhere — but the current comment's "shared" assumption should not survive, because the divergence is deliberate (twin-divergence rule).

Verification for whoever takes this: `git diff develop:.trivyignore develop-v2:.trivyignore` shows the divergence; the workflow's `ref:` line and the script's `IGNOREFILE="$ROOT/.trivyignore"` show the single-lane read.

VijitSingh97:
Prior art: this gap was first recorded 2026-08-23 on #1295/#1346 as a known limitation ("develop/main keep the obsolete entry and the watch never audits develop's own .trivyignore — resolve toward v2 at the next sync"). This issue promotes it from a PR-comment caveat to a tracked defect, with CVE-2026-45447 on develop as the live instance.

VijitSingh97:
[...]

Issue #1174, opened 2026-08-21
Nothing detects an obsolete .trivyignore mute, and the file claims twice that the weekly sweep does


VijitSingh97:
Every entry in `.trivyignore` carries a prose "cleared when …" condition. Nothing checks those
conditions, so a mute outlives the finding it was written for and no gate notices — and the file
says twice, in current tense, that something does notice. It does not.

## The false claim, verbatim

`.trivyignore` on `develop-v2`, twice:

> The weekly CVE sweep (#833) re-surfaces it regardless (#1153).

> The weekly CVE sweep (#833) re-surfaces them regardless, so this cannot rot silently.

The weekly sweep is `os-rootfs.yml`, and it runs the scan **with the ignore file applied**:

```yaml
        with:
          image-ref: pithead-os-rootfs:ci
          ...
          trivyignores: .trivyignore
```

Trivy's ignore file filters matching findings out of the report entirely. A muted finding is
therefore invisible to the sweep by construction — the sweep re-surfaces nothing. `ci.yml`'s image
scan passes the same file, so there is no second gate that sees them either.

Same shape as #1162: text written to describe a compensating control that cannot do the job, which
is what makes a reader conclude the gap is handled.

## Why it matters, and it is not theoretical

The repo's own standing rule is that a mute left over a finding you have already fixed **destroys
the evidence the fix landed** — the scan can no longer tell "fixed" from "hidden". That is exactly
how two obsolete mutes survived until #1153 found them by hand.

There are 8 accepted findings in the file today, spread across the appliance rootfs, the dashboard
image and the `build/*` bases. Each names a manual one-line check (`docker run --rm python:3.11-slim
dpkg-query -W util-linux`, `go list -m -versions github.com/docker/docker`), and each depends on a
base image someone else rebuilds on a schedule we do not control. So the moment a mute becomes
obsolete is a moment nothing in this repo observes.

## The check that would work, and a warning about getting it wrong

The answer is a scan run with **no ignore file at all**, whose only job is to report which muted IDs
no longer appear anywhere. Report-only, weekly, the `pin-watch.sh` posture — never a gate, because a
cleared mute is housekeeping, not a build failure.

**The trap, found while writing this issue.** I built the appliance rootfs and scanned it with no
ignore file, then diffed the whole ignore list against the result. It printed seven IDs as "no longer
match anything" — and acting on that would have deleted the dashboard image's mutes and broken its
[...]

PR #1680, opened 2026-09-03
chore(trivyignore): #1656 drop the obsolete CVE-2026-34040 mute, and record why nothing replaces it

### LINKED ISSUE #1656: .trivyignore's CVE-2026-34040 mute is obsolete, so the scan can no longer tell fixed from hidden
### What

The `CVE-2026-34040` mute in `.trivyignore` is **obsolete** — the ID returns zero occurrences when the rootfs is scanned at all severities, fixed and unfixed, with no ignore file. It is gone from Trivy's database.

`docker/docker v28.5.2+incompatible` now reports a different set instead: `CVE-2026-41567` and `CVE-2026-42306` (both HIGH, both `fixed=NONE`).

### Why it matters

`.trivyignore`'s own header states the rule this breaks: a mute left standing over a finding that is no longer there means **the scan can no longer tell "fixed" from "hidden"**. That is the exact state two obsolete mutes were found in at #1153, and the reason the file carries that warning.

There is a concrete second effect. #1649's stated acceptance check — "expect compose back to its single documented finding" — is **no longer satisfiable as written**, because a correct no-ignore scan now reads **0**, not 1. A reviewer checking #1649 against its own acceptance text would be checking against a number that the tooling can no longer produce.

### Scope

- Drop the `CVE-2026-34040` entry.
- Decide deliberately what to do about `CVE-2026-41567` / `CVE-2026-42306`: both are `fixed=NONE`, so there is no pin bump or module raise that clears them, and the choice is between an expiring mute and accepting the finding. That is a judgement, not a cleanup — record which one and why.
- Re-run the scan with **no ignore file** afterwards and read the result, rather than confirming from release notes.

### Provenance

Found by the appliance lane while proving #1649 with a controlled scan pair; it reported rather than fixed, because `.trivyignore` is the currency lane's path. Filed by the controller. The lane notes it nearly read its own clean scan as a pass because of this, and that the controlled pair is what caught it.

VijitSingh97:
Closes #1656.

## What this changes

One file. `CVE-2026-34040` is removed from `.trivyignore`, and the block that justified it is
replaced by a record of why nothing takes its place.

A mute standing over a finding that is no longer there means the scan can no longer tell "fixed"
from "hidden". That is this file's own stated rule, and the state two mutes were found in at #1153.

## The mute is dead — three routes, and they were derived separately

1. **`scripts/trivyignore-watch.sh`, scheduled run of 2026-08-31** (the weekly watch that exists for
   exactly this, #1174). Six images, **no ignore file applied**, trivy 0.74.0. Its report on #1382
   reads `CVE-2026-34040 | no | OBSOLETE — no covered image reports it`, and
   `obsolete=1 checked=7`. This is the authoritative image-level evidence and I did not generate it
   — I read it.
2. **The appliance lane's controlled scan pair**, found while proving #1649 and reported rather than
   fixed because this file is not their path. RELAYED, not re-run by me.
3. **An isolated module probe, run here today.** A `go.mod` requiring
   `github.com/docker/docker v28.5.2+incompatible` — the module the compiled docker-compose actually
   pulls in — scanned with the pinned `aquasec/trivy` image. `CVE-2026-34040` does not match it at
   all.

Route 3 is not a null result: the same scan reports **four** other findings against that module, so
the instrument demonstrably sees it. The absence is a measurement, not a silence.

**Likely mechanism, labelled because it is inferred and not proven:** the GitHub advisory for
`CVE-2026-34040` today names `go:github.com/moby/moby` and `go:github.com/moby/moby/v2`, and does
**not** name `github.com/docker/docker`. That would explain why it no longer matches our module. It
is read off the advisory's affected-package list; I did not prove trivy's matcher follows it. The
scan results above stand on their own either way.

## Nothing replaces it, and that is the decision the issue asked for

#1656 framed the open question as a choice between an expiring mute and accepting the finding for
`CVE-2026-41567` / `CVE-2026-42306`. Measured at the gate's own scope, there is no such choice,
because neither finding ever reaches the gate.

What that module reports today, at full scope:

| finding | severity | fixed |
|---|---|---|
| `CVE-2026-41567` | HIGH | NONE |
| `CVE-2026-42306` | HIGH | NONE |
| `CVE-2026-33997` | MEDIUM | 29.3.1 |
| `CVE-2026-41568` | MEDIUM | NONE |

`first_patched=NONE` for both HIGHs is confirmed independently against the GitHub advisory API, not
taken from the issue.

The gate's scope is `--severity HIGH,CRITICAL --ignore-unfixed`, shared by `ci.yml`,
`os-rootfs.yml` and `scripts/trivyignore-watch.sh:115`. Both HIGHs are unfixed; the one fixable
finding is below the severity floor. **A controlled pair, same probe and same engine with one
variable changed:**

```
wide scope  (all severities, unfixed included) -> 4 findings
gate scope  (--severity HIGH,CRITICAL --ignore-unfixed) -> 0 findings
```

So an entry for either would be worse than absent. It would suppress nothing the gate ever reports,
and because `trivyignore-watch.sh` scans with `--ignore-unfixed` too, it would be flagged OBSOLETE
on the very next weekly run — a mute that exists only to be reported dead. The block is left in
place, deliberately empty, with that reasoning in it so the next person does not re-add one.

What would change this: compose moving to the docker 29 module path, which is what makes a finding
here fixable and therefore gating. `go list -m -versions github.com/docker/docker` is the check.

## What was run

| | |
|---|---|
| `make lint-trivy-parity` | rc 0 |
| `make lint-file-budget` | rc 0 |
| `make lint-topology` | rc 0 |
| `make lint-docs-voice` | rc 0 |
| `bash scripts/trivyignore-watch.sh --self-test` | rc 0 |
| isolated module probe, pinned engine | wide 4 / gate 0, as above |

`.trivyignore` has no `file-budget.tsv` row; it is 64 lines — three SHORTER than the 67 it started at — well inside the 800 hard ceiling.

## What was NOT run, and why

- **I did not run the full six-image `trivyignore-watch.sh` scan myself.** It builds the appliance
  rootfs, which compiles docker-compose and cosign from source on a pinned Go toolchain — a heavy,
  bench-class job. The image-level evidence here is the 2026-08-31 **scheduled** run, three days
  old at the time of writing, plus today's module-level probe. The next Monday run is the natural
  confirmation and needs nothing from me; if it reports `obsolete=0 checked=6`, this landed
  correctly.
- **I did not verify the knock-on claim #1656 makes about #1649's acceptance text** (that "expect
  compose back to its single documented finding" is no longer satisfiable as written). It is
  plausible given the above and it belongs to the appliance lane's issue, not to this file. Flagging
  it rather than asserting it.
- **`develop`'s copy of `.trivyignore` is untouched by this PR.** The twins genuinely differ — seven
  active mutes each, but not the same seven: `develop` carries `CVE-2026-45447` and no
  `CVE-2026-34040`. That asymmetry is #1442's subject, not this one.

## Over-engineering pass (ponytail gate), and the finding I acted on

Done by hand on this diff, as this lane's gate behaviour requires.

**Finding, accepted and fixed in `894b2cf5`:** the replacement comment block ran to 34 lines to
document *zero* live entries, where the block it replaced was 26 lines documenting one. Documenting
less at greater length is the wrong direction. The methodology (the controlled wide/gate scan pair)
and the advisory-scoping caveat were moved out of the file and into this PR body, where they are
evidence for the reviewer rather than permanent weight on every future reader of `.trivyignore`.
The block is now 23 lines, and the file is **64 lines against the 67 it started at** — a change that
removes a mute should not make the file bigger.

**Considered and rejected: deleting the block outright.** It would be shorter still, but the block's
job is to stop the next person re-adding a mute for this module, and the failure mode is real —
#1153 found two obsolete mutes standing. A file whose entire design is one rationale per entry needs
a rationale for a deliberate non-entry too.

**Considered and rejected: keeping the historical x/text and grpc sentences out.** They are what
explain the premise change (vendored release binaries -> compiled from source), which is why a
finding on this module is fixable in principle at all. Without them the "nothing replaces it"
argument does not stand on its own.

No other over-engineering found: the change adds no mechanism, no script, no test scaffolding and no
new file. It deletes one line of configuration and rewrites the comment that justified it.

### DIFF
--- a/.trivyignore
+++ b/.trivyignore
@@ -9,32 +9,40 @@
 CVE-2026-23949
 CVE-2026-24049
 
-# Appliance rootfs Go binaries. This block used to hold three findings on the grounds that the
-# modules were VENDORED inside downloaded release binaries, so "no pin bump can clear them until
-# upstream cuts releases on the fixed modules". That premise is gone: the rootfs COMPILES
-# docker-compose and cosign itself, so the module graph is ours and a finding in it is fixable
-# whether or not upstream has moved. Two of the three are now fixed at the source by the
-# COMPOSE_GO_RAISES / COSIGN_GO_RAISES entries in os/rootfs/Dockerfile and have been deleted from
-# here — golang.org/x/text (CVE-2026-56852) and google.golang.org/grpc (GHSA-hrxh-6v49-42gf).
-# Deleting them is not cosmetic: while a mute is present the scan cannot tell "fixed" from
-# "hidden", so leaving one for a finding you have just fixed destroys the only evidence that the
-# fix landed. Scanned with no ignore file at all after the raises: cosign 0 findings, compose 1.
+# Appliance rootfs Go binaries. THIS BLOCK IS DELIBERATELY EMPTY — it stays here so that nobody
+# re-adds a line to it without reading why the last one went. It used to hold three findings on
+# the grounds that the modules were VENDORED inside downloaded release binaries, so "no pin bump
+# can clear them until upstream cuts releases on the fixed modules". That premise is gone: the
+# rootfs COMPILES docker-compose and cosign itself, so the module graph is ours and a finding in it
+# is fixable whether or not upstream has moved. Two of the three were fixed at the source by the
+# COMPOSE_GO_RAISES / COSIGN_GO_RAISES entries in os/rootfs/Dockerfile — golang.org/x/text
+# (CVE-2026-56852) and google.golang.org/grpc (GHSA-hrxh-6v49-42gf). Deleting a mute is not
+# cosmetic: while a mute is present the scan cannot tell "fixed" from "hidden", so leaving one for
+# a finding you have just fixed destroys the only evidence that the fix landed.
 #
-# github.com/docker/docker, reached as an indirect dependency of the compiled docker-compose. This
-# one stays, and unlike the other two it genuinely cannot be raised. Trivy names the fix as 29.3.1,
-# but that is a Moby release, not a version of this MODULE — `go list -m -versions` tops out at
-# v28.5.2+incompatible for both github.com/docker/docker and github.com/moby/moby, and the attempt
-# fails the build with "invalid version: unknown revision v29.3.1". Docker 29 lives at a module
-# path compose does not import, so no compose release can carry it either: v5.4.0, v5.5.0 and main
-# all pin v28.5.2+incompatible. The advisory is also a daemon-side authz bypass — compose links the
-# client half, and the appliance talks to podman's compat socket, not dockerd.
-# Clears when compose moves to the docker 29 module path; `go list -m -versions github.com/docker/docker`
-# is the one-line check. Nothing re-surfaces this on its own: the weekly CVE sweep (#833) scans WITH
-# this file applied, so a muted finding never reaches it (#1174). scripts/trivyignore-watch.sh
-# checks whether this ID still shows up in any image the file covers — weekly now, via
-# .github/workflows/trivyignore-watch.yml on the default branch, which checks this lane out to
-# run it (#1174). Same cadence as the clearing check above.
-CVE-2026-34040
+# The third was CVE-2026-34040 on github.com/docker/docker, reached as an indirect dependency of
+# the compiled docker-compose. It is now gone and has been dropped (#1656). Two routes agree:
+# scripts/trivyignore-watch.sh's scheduled run of 2026-08-31 reported it OBSOLETE — no covered
+# image reports it — and an isolated probe on the pinned engine (a go.mod requiring
+# github.com/docker/docker v28.5.2+incompatible) does not match the ID at all. The advisory now
+# names github.com/moby/moby rather than github.com/docker/docker; that is the likely mechanism,
+# but it is inferred from the advisory's affected-package list and was not proven against trivy's
+# own matcher.
+#
+# NOTHING REPLACES IT, and that is a decision rather than an oversight (#1656). The same module
+# today reports CVE-2026-41567 and CVE-2026-42306 (HIGH, both fixed=NONE), CVE-2026-33997 (fixable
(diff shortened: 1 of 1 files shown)


VijitSingh97:
**Controller pass (non-author, fable tier) — APPROVED, merged on the checks read at the moment of merge (last line).**

Proven by me at head `894b2cf5`. The branch sits ON the tip `9791a693`, so the branch tree and the merged tree are one tree and one run answers both 8a legs; both were run anyway: budget gate rc 0 and parity rc 0 at the head and at the merged tree, `merge-tree` rc 0, `make lint-trivy-parity` rc 0 (v0.74.0 at all three scan sites). The ID survives only inside the comment at `.trivyignore:22`; no live line anywhere in the tree.

Route 1 re-derived at source, and the pointer is right in substance though not in kind: the watcher publishes to #1382's **body** (`trivyignore-watch.yml` does `gh issue edit`), not a comment — which is why a comment search comes back empty. The body reads `CVE-2026-34040 | no | OBSOLETE — no covered image reports it`, marker `obsolete=1 checked=7`, from the scheduled run of 2026-08-31 (run 33404760185).

The deciding instrument is on this PR itself: `Build + scan the appliance rootfs` runs here because `.trivyignore` is in its paths filter, and it scans the built rootfs with the ignore file AS CHANGED at `severity HIGH,CRITICAL`, `ignore-unfixed`, `exit-code 1`. A green there is the image-level proof that the mute was dead, on the very image compose lives in — stronger than all three routes in the body.

RELAYED, not re-run by me: route 2 (the appliance lane's pair), route 3 (the module probe, wide 4 / gate 0), and the advisory-scoping mechanism, which the author labelled inferred and I leave so.

Merge-record: `Closes #1656` does not fire on this base; I close #1656 by hand after the merge. The #1442 twin asymmetry and the #1649 knock-on are flagged rather than asserted — the right call.

Checks at merge: `Build + scan the appliance rootfs` SUCCESS; every check conclusion SUCCESS or SKIPPED (zizmor's row showed status IN_PROGRESS with conclusion SUCCESS — read by conclusion), head unchanged at `894b2cf5`, read in the same command as the merge.

PR #1681, opened 2026-09-03
fix(lint): file-budget gate enumerates untracked files, not just the index

VijitSingh97:
The file-budget ratchet gate (`scripts/lint-file-budget.sh`) builds its candidate list from `git ls-files` alone, which only enumerates the index. A new file over the 400-line target that has never been `git add`'ed is invisible to the gate — the exact moment the gate exists to catch, since staging is the very next step. The gate reads that as zero over-target candidates and passes.

## What the gate now refuses

`list_candidates` unions `git ls-files` with `git ls-files --others --exclude-standard` (untracked, honoring `.gitignore`), so the working tree — not just the index — is what gets measured. The two lists are disjoint, so no dedup is needed. An over-target untracked file now fails the gate by name, the same as a tracked one.

## How it was proven

Deciding commands, run from the repo root:

```
bash scripts/lint-file-budget.sh --self-test
bash scripts/lint-file-budget.sh
```

(equivalently `make lint-file-budget`). Before the fix, a 500-line untracked file dropped into a scratch fixture tree passed the gate silently. After the fix, the same fixture fails with rc=1 and names the file.

A new self-test case in `scripts/lint-file-budget-selftest.sh`: writes a 500-line `untracked.sh` into the fixture tree without `git add`, runs the gate, and asserts both that it exits non-zero and that the failure output names `untracked.sh` specifically — not a vacuous pass, and not a generic failure that could hold even if the untracked path were still silently skipped.

Refs #1486

### DIFF
--- a/scripts/lint-file-budget-selftest.sh
+++ b/scripts/lint-file-budget-selftest.sh
@@ -86,6 +86,20 @@ self_test() {
     rm -f "$tmp/new.sh"
     git -C "$tmp" rm -q --cached new.sh >/dev/null 2>&1 || true
 
+    # #1486 — `git ls-files` alone only sees the INDEX; an untracked new file must be a
+    # candidate too, or the gate reads a skip as a pass on exactly the file it exists to catch.
+    seq 1 500 >"$tmp/untracked.sh"
+    rc=0
+    (cd "$tmp" && run_gate) >"$out" 2>&1 || rc=$?
+    expect "an over-target file that is untracked (never git add'ed) still FAILS (#1486)" 1 "$rc"
+    if grep -q "untracked.sh is 500 lines (> 400 target) but has no" "$out"; then
+        echo "  self-test ok: names the untracked file, not a vacuous pass"
+    else
+        echo "  self-test FAIL: did not name the untracked file"
+        st_fail=1
+    fi
+    rm -f "$tmp/untracked.sh"
+
     printf '# test budget\nbudgeted.sh\t400\n' >"$tmp/$BUDGET_FILE"
     rc=0
     (cd "$tmp" && run_gate) >"$out" 2>&1 || rc=$?
--- a/scripts/lint-file-budget.sh
+++ b/scripts/lint-file-budget.sh
@@ -2,8 +2,9 @@
 # The file-budget ratchet gate (#1105 Phase 0): stop the biggest files in the repo from getting
 # any bigger, without demanding anyone rewrite them today. Two rules:
 #
-#   1. A tracked file over the 400-line target must record a ceiling in
-#      docs/dev/file-budget.tsv; over 800 the refusal names the hard ceiling instead.
+#   1. A candidate file (tracked, or untracked and not gitignored — #1486) over the 400-line
+#      target must record a ceiling in docs/dev/file-budget.tsv; over 800 the refusal names the
+#      hard ceiling instead.
 #   2. An existing offender (already over 400 when this gate landed, or added to the budget
 #      since) gets its CURRENT line count recorded in docs/dev/file-budget.tsv as a personal
 #      ceiling. A PR may not grow that file past its recorded ceiling. Ceilings only ever go
@@ -121,10 +122,18 @@ count_lines() {
     awk 'END { print NR + 0 }' "$1" 2>/dev/null || echo 0
 }
 
-# Every tracked, non-exempt, non-binary file: "path<TAB>lines", one per line.
+# Every tracked OR untracked-but-not-gitignored, non-exempt, non-binary file: "path<TAB>lines",
+# one per line. `git ls-files` alone only lists the INDEX — a new file sits invisible right up
+# until `git add`, which is exactly the moment a budget gate exists to catch it (#1486). Union in
+# `--others --exclude-standard` (untracked, honoring .gitignore) so the working tree, not the
+# index, is what gets measured; the two lists are disjoint (`--others` never repeats a tracked
+# path) so no dedup is needed.
 list_candidates() {
     local f
-    git ls-files | while IFS= read -r f; do
+    {
+        git ls-files
+        git ls-files --others --exclude-standard
+    } | while IFS= read -r f; do
         [ -f "$f" ] || continue
         is_exempt "$f" && continue
         is_binary_or_empty "$f" && continue


VijitSingh97:
**Controller pass at head `1751f0ee` — CHANGES REQUESTED on one small in-repo prose defect (F1). The fix itself is proven; merge follows the moment F1 lands and the checks are re-read.**

**Proven by me (nothing relayed):**

- **Negative control — the new test discriminates.** Head's `lint-file-budget-selftest.sh` run against the BASE gate (`scripts/lint-file-budget.sh` at merge-base `9791a693`): rc 1, and exactly the two new lines fail (`expected rc 1, got 0`; `did not name the untracked file`) while the other 26 cases stay ok. Head's self-test against head's gate: 28 ok, rc 0.
- **Ruling 8a, both legs** (branch is one commit off tip `80cd9c02`; merge-base `9791a693`): at the branch head `make lint-file-budget` rc 0 and `lint-pithead-parity` rc 0; `merge-tree --write-tree` rc 0, and on the merged tree budget rc 0, parity rc 0.
- **The cost the issue itself flagged (local runs now measure scratch files):** across the seven lane worktrees on the build host, `git ls-files --others --exclude-standard` lists ONE untracked file total and ZERO over 400 lines. Nothing goes red today; the exposure is a future scratch file, and it fails loudly by name rather than silently.
- Checks: every check terminal and green (the two CVE sweeps are skipped by design on this base).

**F1 (blocking, minor — durable in-repo text that this PR's own change makes false; ruling 2):**

- `Makefile:119`, the `lint-file-budget` help line, still reads "Fail if a **tracked** file crosses the 800-line hard ceiling…". That is what `make help` prints to a contributor, and it now describes an enumeration the gate no longer does.
- In `run_gate`, the FATAL string for a zero-length candidate scan still says "(git ls-files, minus exemptions/binaries)". Same class.
- Leave `enforce_nonempty_enumeration`'s "zero tracked files" alone — that guard still counts the index only, so that sentence is still true.

Two one-line edits. `Makefile` is a serialized shared file: say so in the body.

**Noted, not blocking:**

- `--generate` goes through the same `list_candidates`, so an untracked over-target scratch file now gets a tsv row. If that file is never committed, CI refuses at the "lists a path which no longer exists" door — loud and self-correcting, but a lane running `--generate` with litter in its tree should expect it.
- The issue's shape 3 (the success line states the candidate count it measured) is not taken here. I will close #1486 on merge with that noted rather than hold it open on a nice-to-have.

VijitSingh97:
## Controller pass — FINDINGS (non-author; judged on the fable tier). One doc line, then merge.

**Re-derived by me:**
- Diff read in full. `list_candidates` unions the index with `git ls-files --others --exclude-standard`; the two sets are disjoint by construction, and everything downstream (`-f`, exempt, binary-or-empty) still filters per path.
- The #1486 fixture fired in my own run at the head and at the merged tree (`an over-target file that is untracked (never git add'ed) still FAILS (#1486)`), and the name assertion matches the real refusal text at `lint-file-budget.sh:183`.
- Ruling 8a, both legs vs the tip `2347df02` (merge-base `9791a693`): budget rc 0, parity rc 0 at the head and at the merged tree; `merge-tree` rc 0. #1699 lands ahead of this and edits a different region of the same file; I re-run `merge-tree` at merge time.

**Fix before merge:** `CONTRIBUTING.md:54` still describes the gate as applying to "a new tracked file". After this change the candidate set is the working tree, tracked or not yet added — the whole point of the PR — so the doc now understates the gate. One clause, in the house voice.

Merge condition: that line; checks none non-terminal, none failed at the moment of merge. #1486 closes by hand.

VijitSingh97:

Re-derived at source in a detached worktree at PR head `1751f0ee`: `scripts/lint-file-budget.sh:131-140` (`list_candidates`) unioned `git ls-files` with `git ls-files --others --exclude-standard`, confirming the claimed defect — the pre-fix form used `git ls-files` alone and misses any file never staged.

Deciding commands, from `scripts/lint-file-budget-selftest.sh`/`scripts/lint-file-budget.sh` as named in the body:
```
$ bash scripts/lint-file-budget-selftest.sh
  self-test ok: an over-target file that is untracked (never git add'ed) still FAILS (#1486)
  self-test ok: names the untracked file, not a vacuous pass
lint-file-budget self-test OK
```

Standing test — negative control: reverted `list_candidates` to the pre-fix `git ls-files | while ...` form (no union) and re-ran the self-test. Only the two #1486 assertions reddened, everything else in the 28-case suite stayed green:
```
  self-test FAIL: an over-target file that is untracked (never git add'ed) still FAILS (#1486) (expected rc 1, got 0)
  self-test FAIL: did not name the untracked file
lint-file-budget self-test FAILED
```
A precisely-aimed named kill, not a vacuous assertion. Restored the file to the PR's version afterward (worktree diff clean).

"Could this gate be green if a consumer ignored its exit code" — no: `Makefile:120-121` runs both `bash` invocations un-piped under `make`, which stops on the first non-zero `$?`, and the self-test additionally greps the failure text for the specific filename, not just `rc`.

One non-blocking note: `bash scripts/lint-file-budget.sh` alone currently FAILS on this branch — `docs/dev/file-budget.tsv raises tests/os/run.sh's ceiling from 3423 to 3437`. That predates this PR: the branch's tsv still carries `3437`, `origin/develop-v2` tip has since lowered it to `3423` (`ebb0f43d`, unrelated to #1486). This PR does not touch the tsv — it's base drift, resolved by a rebase, not a defect in this diff.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `1751f0ee`

VijitSingh97:
## Controller merge verdict — MERGING at `1751f0ee` on the non-author reviewer PASS (20:07Z) plus this seat's own proven-fix finding (18:21Z); the F1 doc clause is DEFERRED, not waived (non-author; judged on the fable tier).

**Re-derived by me at the tip `289821f1`:**
- `merge-tree --write-tree` rc 0. Ruling 8a, both legs: at the **branch head** the budget gate is rc 2 — `docs/dev/file-budget.tsv` still carries `tests/os/run.sh` at 3437 while the tip lowered it to 3423, the moved-base artifact on every un-rebased branch today and exactly the shape #1710 set the precedent for; at the **merged tree** the gate is rc 0 and the #1486 self-test row fires (`an over-target file that is untracked (never git add'ed) still FAILS`). Parity rc 0 on both legs. What lands is green; the head-leg red is the tsv row, which this diff never touches.
- Every check terminal and green at this head.

**Why merge with F1 open:** the fix is proven by two independent passes; the residual is one prose clause in a file this PR does not touch (`CONTRIBUTING.md:54` still says the gate applies to "a new tracked file"). Holding a proven gate fix six hours on a doc clause serves nobody. **So: #1486 STAYS OPEN on that one clause** — it closes when `CONTRIBUTING.md` says the candidate set is the working tree, tracked or not. Ruling 2 applies (false claim in durable in-repo docs earns a discharge commit); the author's seat owns it.

PR #1683, opened 2026-09-03
fix(dashboard): #1644 a failing payout sync no longer skips the steps beneath it

### LINKED ISSUE #1644: dashboard: a raise in one poll step skips its unrelated siblings for the rest of that poll
## What

`DataService.run()` wraps the whole poll body in one `try`. A raise from any step therefore skips
**every remaining step in that poll**, including steps that have nothing to do with the one that
failed, and the `except Exception` handler logs and continues as though the poll had merely
degraded.

Split out of #1637 as its **point 2**, on the controller's ruling — it is a different defect with a
different fix shape (per-step guards, not counter placement), and bundling it made the one-line
counter fix unreviewable and did not fit that file's headroom.

Line numbers at `develop-v2` @ `8f8de10d`:

| step | line |
|---|---|
| `await self._sync_payouts()` | 1420 |
| `await self._sync_tari_payouts()` | 1425 |
| `if self.update_checker.enabled:` | 1430 |
| `await self._sync_prices()` | 1446 |
| `except Exception as e:` | 1449 |

So a malformed body out of the **Monero** wallet at 1420 silently suppresses the **Tari** payout
sync, the release check and the price sync for that poll. They share nothing but a `try`.

## Measured, not reasoned

Observed while writing #1637's test, and it is what caught a wrong assumption of mine. Driving
`run()` for two polls with `_sync_payouts` raising on the first:

- `_sync_prices` is awaited **once**, not twice — it never ran on the poll that raised.
- With the #1637 fix in place, poll 2 does not re-run the failing step, so `_sync_prices` recovers
  on poll 2. That once is the whole of its recovery.

The test at `tests/service/test_data_loop_counter.py::test_the_loop_survived_the_raise_rather_than_ending_on_it`
asserts that `1` today and carries a note saying that if it ever reads `2`, this defect was fixed.
**That test is a live tripwire for this issue** — whoever fixes this should expect to update it.

## What #1637 does and does not do for this

#1637 stops the counter freezing, so a persistent failure no longer re-runs the failing step on
every poll. That cuts how **often** this cross-contamination happens by roughly ten times. It does
**not** fix it: on the polls where the raise does occur, the unrelated siblings are still skipped.

## Shapes worth considering — none of them ruled

- **Per-step `try`.** Wrap each independent step so one failure cannot skip its siblings. Most
  direct; costs lines in a file with **3 lines of headroom** (`data_service.py` 1451/1454 at the
  tip, 1453/1454 once #1637 lands), so it likely needs an extraction or a helper rather than
  inline `try` blocks.
- **A small runner helper.** Something like `await self._step(name, coro)` that logs and continues,
  applied to the independent steps. One place to change, and it reads as a list of steps; needs a
  decision about which steps are genuinely independent.
- **Leave the coupled ones coupled.** Not every step below another is independent of it — this
  needs reading before any wholesale wrap, and that reading is most of the work here.

## Not claimed / not verified

- **Not** claimed that a live path reaches this today. #1592's trigger at 1420 is fixed (those
  clients return `None`/`[]` rather than raising); this is the general shape, latent.
- **Not** measured against a running dashboard — driven through the test harness, not a real node.
- I have **not** worked out which steps are safe to isolate. That is the open design question and
  it is deliberately left open here.
- The budget figures above are measured; the claim that a per-step wrap will not fit inline is my
  estimate, not a measurement.

VijitSingh97:
Closes #1644.

## The defect

`DataService.run()`'s poll body is **33 awaited steps under one `except Exception`** at the bottom, so any step raising skips every step below it for that poll. A malformed body out of the Monero wallet took the Tari payout sync, the release check and the price feed down with it — steps with nothing to do with Monero payouts.

The `% 10` gates make it *worse* rather than *cause* it: work skipped there does not come back for ten polls. The issue title implies the gates are the subject; per the ruling, they are not. **The skip is structural, and this PR fixes it for two steps out of 33.**

## Scope — two steps, not thirty-three

`_sync_payouts` and `_sync_tari_payouts` take **no poll local and write no `self` attribute** (DB and alert side-effects only), so isolating them creates no cross-iteration edge. That is the ground they were cleared on — stronger than position in the body.

**Left coupled deliberately:**

- **The XvB block.** `_maybe_register_xvb(shares_list, p2pool_stats)` reads two poll locals written by collectors at the top of the body. Wrapping it per-step would let it run on a poll where those locals are stale or unset. There is a test that reds if someone widens the wrap to it.
- **The other 30 steps.** Not cleared. Widening owes the cross-poll enumeration (issue comment `5526436316`) and the **ten callees still marked `n/a (external)`** in the recon's swallow column — that means the callee was never resolved, not that it does not swallow.
- **`_sync_prices`.** Last statement before the handler, so isolating it protects no sibling. It *is* worth noting that it writes `self.latest_data` at `:653`, which the ruling's premise had as inert — that is a staleness question and belongs with the wall-clock issue, not here.

**No wall-clock conversion.** It dissolves #1637's defect, not this one: under `now - self._last_x >= 300` a raise at the payout sync still skips everything below it. Orthogonal, and its own issue.

## Shape, and why a new module

`data_service.py` was at **1453 against a ceiling of 1454** — one line. An inline `try`/`except`/`log` costs three lines per step, so this could not be done in place; that upgrades the issue's estimate to a measurement. Moving the two bodies into a new `service/payout_sync.py` (**95 lines**, under the 400 target, so no `file-budget.tsv` row) takes `data_service.py` to **1420**. The ceiling row is **not** lowered to match — freeing lines and then lowering the row nets zero headroom.

`_sync_payouts` / `_sync_tari_payouts` stay on `DataService` as **thin delegating wrappers**. Not ceremony: they keep the poll body's call seam and the name **ten existing tests in `test_data_service.py`** already reach for, and that file is at its own **2538 ceiling** with no room. Those ten tests are unchanged and drive a real `StateManager`, so they verify the moved dataflow rather than passing over a stub.

## What was RUN

- Full dashboard suite **2444 passed / 0 failed**; total coverage **97.39%**.
- **Patch coverage 100% on 33 changed lines**, with the gate's own non-vacuity check reporting both changed files present in `coverage.xml` — not the #1000 vacuous pass.
- `lint-py`, `lint-file-budget`, `lint-topology`, `lint-md` rc 0, run individually, **with the new module staged first** — the budget and topology gates enumerate the git index, so an untracked new file is invisible to them.
- `lint-sh` NOT run: it is shellcheck and this diff has no shell in it.

## Every new law was fired against a seed

Green tests prove nothing about laws nobody armed. Each was seeded in the direction it exists to catch:

| seed | reds | stays green |
|---|---|---|
| remove the wrap at both call sites | sibling-survival, the log line, the #1637 tripwire | the three #1637 cadence tests |
| wrap `_sync_xvb_stats` (the forbidden widening) | the XvB coupling law | everything else |
| `except BaseException` | the cancellation law | everything else |

The **#1637 tripwire moved as that issue's own docstring said it must**: `_sync_prices` across two polls with the payout sync raising now reads **2, not 1**, and the docstring explaining the shortfall is rewritten rather than left standing.

The XvB coupling law carries a **firing control** (`test_the_block_runs_whole_when_nothing_raises`), because four zeroes are also what an unreachable block reports, and `ENABLE_XVB` plus a `% 10` gate is two ways to be silently unreachable.

The **cancellation law** is not in the ruling's acceptance list; it is mine. `CancelledError` derives from `BaseException`, so shutdown already propagates through the guard — but that is a property of the *spelling*, and `except BaseException` reads as a harmless widening while hanging shutdown behind an in-flight payout sync.

**The guard logs.** The loop's handler emits one undifferentiated "Data Collection Error" per failed poll. A step that now fails *without* ending the poll would otherwise be **quieter than before it was guarded**, so `run_isolated` names the step and the cause.

## Over-engineering pass

One candidate found and **rejected with a reason**, rather than a clean bill:

- **The 5-line comment at call site 7d partly restates `payout_sync`'s module docstring** (`shrink`, ~3 lines). Kept: it is the only place a reader *of the poll body* learns why these two steps are special, and it sits exactly where a future editor would widen the wrap and break the invariant. It points to the module rather than restating the full reasoning.

Also considered: dropping the delegating wrappers entirely (rejected — see Shape; it would churn 10 tests in a file at its ceiling and force a lambda at the call site), and testing the log line directly against `run_isolated` instead of through the loop (rejected — the loop version pins the *label the poll body actually passes*, which a direct call cannot).

## Not done, deliberately

- **The structural defect stands for the other 31 steps.** This is a narrow fix by ruling, and saying so is the point: a guard applied where nobody established correctness would look like the work was done and stop the next contributor.
- **No docs change.** `repo-map.md` does not enumerate service modules — `data_helpers.py`, the precedent extraction from #1105 Phase 3, has no row there either — and the change is not operator-visible in `docs/dashboard.md`.
- **Nothing proven on hardware.**

### DIFF
--- a/dashboard/mining_dashboard/service/data_service.py
+++ b/dashboard/mining_dashboard/service/data_service.py
@@ -71,7 +71,7 @@ from mining_dashboard.helper.utils import (
     pplns_block_time,
     shares_in_pplns_window,
 )
-from mining_dashboard.service import audit_service, worker_change_audit
+from mining_dashboard.service import audit_service, payout_sync, worker_change_audit
 from mining_dashboard.service.alert_service import AlertService
 from mining_dashboard.service.clearnet_sync import ClearnetSyncSupervisor
 from mining_dashboard.service.control_service import (
@@ -655,52 +655,14 @@ class DataService:
             )
 
     async def _sync_payouts(self):
-        """Confirm on-chain payouts from the view-only wallet-rpc (#381), throttled by the caller.
-
-        Seeds the query from the highest stored Monero payout height, so a restart re-scans only
-        the tip; ``add_payouts`` is idempotent on ``(chain, txid)``, so the overlap is dropped and
-        nothing replays. Every genuinely-new confirmed payout fires exactly one ``payout_confirmed``
-        alert. A wallet still doing its first-run scan (or briefly unreachable) returns ``[]`` — a
-        quiet no-op, no error. chain="monero" here; the Tari sibling (#462) reuses the same table."""
-        chain = "monero"
-        min_height = await asyncio.to_thread(self.state_manager.get_payout_max_height, chain)
-        payouts = await asyncio.to_thread(self.wallet_client.get_confirmed_payouts, min_height)
-        if not payouts:
-            return
-        new_rows = await asyncio.to_thread(self.state_manager.add_payouts, chain, payouts)
-        for r in new_rows:
-            logger.info(
-                "Payout confirmed on-chain: %.6f XMR (tx %s…) at height %d (#381)",
-                r["amount_atomic"] / 1e12,
-                r["txid"][:8],
-                r["height"],
-            )
-            await self.alert_service.payout_confirmed_alert(chain, r["amount_atomic"], r["txid"])
+        """Monero on-chain payout confirmation (#381). The body moved to ``payout_sync`` for #1644;
+        this stays as the poll body's call seam, and as the name the tests already reach for."""
+        await payout_sync.sync_monero(self.state_manager, self.wallet_client, self.alert_service)
 
     async def _sync_tari_payouts(self):
-        """Confirm Tari on-chain payouts from the view-only console wallet (#462), throttled by the
-        caller — the Tari sibling of ``_sync_payouts``.
-
-        Identical shape: seed from the highest stored Tari payout height, stream new confirmed
-        payouts, persist to the shared ``payouts`` table with chain="tari" (idempotent on
-        ``(chain, txid)`` so a restart replays nothing), and fire one ``payout_confirmed`` alert per
-        genuinely-new payout. ``amount_atomic`` is microTari here; the shared alert divides by the
-        Tari divisor. The Tari client is async (grpc.aio), so it's awaited directly rather than via
-        ``asyncio.to_thread``. An empty/unreachable scan is a quiet no-op."""
-        chain = "tari"
-        min_height = await asyncio.to_thread(self.state_manager.get_payout_max_height, chain)
-        payouts = await self.tari_wallet_client.get_confirmed_payouts(min_height)
-        if not payouts:
-            return
-        new_rows = await asyncio.to_thread(self.state_manager.add_payouts, chain, payouts)
-        for r in new_rows:
-            logger.info(
-                "Tari payout confirmed on-chain: %.6f XTM (tx %s…) at height %d (#462)",
-                r["amount_atomic"] / 1e6,
-                r["txid"][:8],
-                r["height"],
-            )
-            await self.alert_service.payout_confirmed_alert(chain, r["amount_atomic"], r["txid"])
+        """Tari on-chain payout confirmation (#462) — the sibling of ``_sync_payouts``, same shape
+        and same reason for staying here while its body lives in ``payout_sync``."""
+        await payout_sync.sync_tari(self.state_manager, self.tari_wallet_client, self.alert_service)
 
     async def _record_audit_event(self, source, actor, action, status, keys, event_id=None):
         """Write one out-of-band audit row (#530), through the SAME sanitizer #33's own audit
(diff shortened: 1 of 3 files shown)


VijitSingh97:
**Controller pass at head `0b5ce7c3` — PASS, merging.** Judged on the seat's pinned top tier. Every claim below is PROVEN BY ME unless marked RELAYED; the full-suite figures are the one relay.

**Proven by me, in a throwaway worktree at the head:**
- **The move is a move.** Each old body (`_sync_payouts`, `_sync_tari_payouts`) with `self.<collaborator>` substituted by the parameter name and the docstring dropped is byte-identical to `sync_monero` / `sync_tari` (568 B and 556 B). Control: the two new bodies differ from each other.
- **Budget.** `data_service.py` 1453 → **1420**, `payout_sync.py` **95**, the test file **313**; `file-budget.tsv` untouched in the diff, so the 1454 row stands and the freed lines are real headroom rather than a lowered ceiling. `lint-file-budget` and `lint-pithead-parity` rc 0 at the **branch head AND at the merged tree** (both legs run because the merge-base `80cd9c02` is one commit behind the tip); `merge-tree --write-tree` rc 0.
- **Tests.** `test_data_loop_counter.py` + `test_data_service.py`: **135 passed / 0 failed** (counted from junit XML, not the `-q` tail).
- **The seed table, fired the same direction the body claims.** Seed A (both wraps removed, 2 lines): exactly three red — the sibling-survival law, the log-line law, and the #1637 tripwire now at 2 — the three cadence tests and the XvB pair stay green. Seed B (`_sync_xvb_stats` wrapped): exactly one red, the coupling law; its firing control stays green. Seed C (`except BaseException`): exactly one red, the cancellation law. Each seed's arming was read back (call-site count 0 / 3 / 1) before the run, and the file restored by byte-check after.
- **Patch coverage 100% on 33 lines** via `scripts/patch-coverage.sh` against `origin/develop-v2`, with its non-vacuity check reporting both changed files present in `coverage.xml`.
- **Ruling acceptance (#1644 § 6), all four met:** tripwire moved to 2 with its docstring rewritten; sibling-survival test present; coupling-kept test present with a firing control; the PR body names what stayed coupled and why.

**RELAYED, not re-run here:** 2444 passed / 97.39% total coverage — CI's dashboard job carries that gate and is green at this head.

**F1 — non-blocking, OWED to the dashboard lane:** ruling § 4 on #1644 told the implementer to file the wall-clock-recovery improvement separately, citing `:1343-1359` / `:1364-1380` as the precedent. No such issue exists (searched open issues for the phrase and for `_last_hourly_capture`). The PR body says "its own issue" without a number. File it, in `v2.x - post-GA`, and cite this PR.

**Merge mechanics:** `Closes #1644` will not fire on a `develop-v2` PR; I close #1644 by hand with a note saying what landed and what was deliberately left. Post-merge patch identity, remote-branch deletion with a live-branch control, and gate + parity at the new tip follow in the next comment.

**Not done:** nothing on hardware; the full dashboard suite was not re-run locally.

Issue #1290, opened 2026-08-22
trivyignore-watch scans with a different trivy engine than the gate it reports on


VijitSingh97:
Found by #1288's adversarial review; pre-existing since #1253, consequential now that the watch recurs on a schedule instead of being run by hand.

## The drift

`scripts/trivyignore-watch.sh` (appliance lane) pins its scanner:

- watch: `aquasec/trivy@sha256:62b1e65e…` — the digest for **0.74.0** (verified against the registry)
- gate: `aquasecurity/trivy-action@ed142fd0…` (v0.36.0) in ci.yml and os-rootfs.yml, with **no `version:` override** — that action's own default is **v0.70.0**

So the mute watch scans with a four-minors-newer engine than the gate whose mutes it judges. The script's header promises build parity ("Each is built EXACTLY the way the gate builds it… scanning an image built any other way proves nothing about what the gate sees") but never extends that promise to the scanner binary — and engine detection logic moves between minors, so the watch can call a mute obsolete that the gate still hits, or keep one alive the gate would never see. A report-only tool whose report can be wrong in both directions is the defect class this repo keeps relabeling.

The script even carries the instruction "Bump alongside the aquasecurity/trivy-action pin in ci.yml/os-rootfs.yml when that moves" — the drift existing anyway is the usual proof that a comment is not a guard.

## Fix direction

One source of truth for the engine version, mechanically held together — either pass the script's pinned version to the gate's `version:` input, or derive the script's pin from what the pinned action resolves, and add a check that fails loudly when the two diverge (the respell test: bump one side and the guard must go red).

## Second hardening, same file

`scan_image()` sends trivy/docker stderr to /dev/null; a production scan failure surfaces only as `<name> (scan)` in the failed list with the root cause invisible in the Actions log. Keep stdout clean for the report, but land stderr somewhere a triager can read.

VijitSingh97:
Fix is up as #1344 (develop-v2): `TRIVY_VERSION` in `scripts/trivyignore-watch.sh` is the single source of truth; both gate workflows pass `version: v0.74.0` explicitly; `make lint-trivy-parity` holds the workflows to it (mismatch, missing key, bare non-`v` value, or no trivy-action step at all → red); the real run measures the pinned image's `trivy --version` and refuses to report on any drift. Mutation proofs are in the PR body.

[...]

Issue #1169, opened 2026-08-21
Retiring develop-v2: three workflows and a Dependabot entry name it as a literal, and one fails silently


VijitSingh97:
When the appliance ships, `os/` merges to the default branch and `develop-v2` goes away. Several
pieces of automation name that branch as a literal string, and each of them fails differently — some
loudly, some not at all. Nothing today lists them, so the retirement would be found one broken
watcher at a time.

This is a checklist issue, not a defect. Nothing here is wrong now.

## What names `develop-v2` as a literal

**Re-derived mechanically at `origin/develop` tip `77104cb6` (2026-09-05), not audited by eye.**
The default branch is the only one `schedule:` and Dependabot read, so it is the only tree that
governs. Method at the bottom; it is the part worth keeping, because this list has already gone
stale twice inside a day.

**There are 8 functional literals, in 4 files, on `develop` as it stands today** — and the
retirement adds a ninth, in a fifth file, when `develop-v2` merges in (see below). A functional literal is a `ref:` or
`target-branch:` that actually resolves a branch — prose in comments is excluded throughout and is
numerous and harmless.

| where | what it is | when `develop-v2` is gone | how it shows |
|---|---|---|---|
| `.github/workflows/pin-watch.yml:54` | `ref: develop-v2`, the `appliance` matrix lane | checkout cannot resolve the ref; `fail-fast: false` keeps the `product` lane reporting | red weekly run |
| `.github/workflows/os-rootfs.yml:63` | `ref:` expression, every non-PR trigger | checkout fails **before** the tree guard, so `:83`'s message never prints | red weekly run |
| `.github/workflows/trivyignore-watch.yml:43` | `ref: develop-v2`, the single `mute-watch` job | checkout fails; that job **is** the watcher | red weekly run |
| `.github/dependabot.yml:63` | `docker`, `/os/rootfs` | that one update job errors | **quiet** |
| `.github/dependabot.yml:137` | `github-actions`, `/` | that one update job errors | **quiet** |
| `.github/dependabot.yml:148` | `uv`, `/dashboard` | that one update job errors | **quiet** |
| `.github/dependabot.yml:174` | `docker`, `directories:` (`/dashboard` + four `build/*`) | that one update job errors | **quiet** |
| `.github/dependabot.yml:190` | `docker-compose`, `/` | that one update job errors | **quiet** |

`.github/dependabot.yml` was parsed rather than grepped (`yaml.safe_load`, so comment text cannot
[...]

PR #1685, opened 2026-09-03
build(deps): #1425 #1650 Dependabot watches develop-v2, not just the frozen twin

### LINKED ISSUE #1425: Dependabot's uv ecosystem watches a path that exists only on develop, so develop-v2's ruff pin is unwatched

### LINKED ISSUE #1650: develop-v2's dashboard has no Dependabot coverage — the config's directory: is resolved against the default branch
## What

`develop-v2`'s dashboard — both its Python dependency tree and its container base image — has **no Dependabot coverage at all**, and has had none since #1106 moved `build/dashboard/` to `dashboard/`.

Dependabot reads `.github/dependabot.yml` from the **default branch only**, which is `develop`. The config's `uv` entry and the `docker` group both name `/build/dashboard`, a path that exists on `develop` and **does not exist on `develop-v2`**. Only the `/os/rootfs` entry carries `target-branch: develop-v2`.

This is the same class as the `schedule:`/default-branch trap the config's own `/os/rootfs` comment documents — but one level in: not *which branch the workflow fires from*, but *which branch the `directory:` is resolved against*.

## Measured, not reasoned

Enumerated from `git ls-tree -r origin/develop-v2` against the config at `origin/develop`:

| develop-v2 path | Dependabot entry | Where its PRs open |
|---|---|---|
| `dashboard/Dockerfile` | **none** — config names `/build/dashboard` | nowhere |
| `dashboard/pyproject.toml` + `uv.lock` | **none** — same | nowhere |
| `build/monero/Dockerfile` | in the `docker` group | `develop` |
| `build/p2pool/Dockerfile` | in the `docker` group | `develop` |
| `build/tor/Dockerfile` | in the `docker` group | `develop` |
| `build/xmrig-proxy/Dockerfile` | in the `docker` group | `develop` |
| `os/rootfs/Dockerfile` | `target-branch: develop-v2` | **`develop-v2`** ✔ |
| `tests/integration/tor-client/Dockerfile` | **none on either twin** | nowhere |

**The drift is already visible, so this is not a theoretical gap:**

- **ruff:** `develop-v2` pins `0.16.3`. `develop` is on `0.16.4`, and #1647 is open to take it to `0.16.5`. develop-v2 is two releases behind with no mechanism to catch up. The pin's own comment says it exists so "local, pre-commit, and CI all run the SAME ruff" — the twins currently do not.
- **Python base image:** `develop-v2`'s `dashboard/Dockerfile` pins `python:3.11-slim@sha256:9c900de…`. That is the **exact digest #1646 is bumping away from** on `develop`. So the lane that ships is sitting on a base Dependabot has already flagged as stale on the lane that does not.

Both open PRs (#1646 `docker`, #1647 `uv`) are therefore bumping a frozen branch's copy of a directory that has moved, while the live lane gets nothing.

## Why this is worth fixing rather than papering over

`develop-v2` is where all work lands (`develop` is frozen). Dependency currency is #1128's whole subject, and the `/os/rootfs` entry shows the fleet already understood the mechanism and applied it in exactly one place. The failure is silent in the direction that looks like success: the repo has an open Dependabot PR stream, green CI, and a `dependabot.yml` that reads as complete — nothing anywhere says the dashboard is uncovered.

## Suggested shape, not a prescription

Add `develop-v2`-targeted entries alongside the existing ones, following the `/os/rootfs` precedent exactly:

- a `uv` entry for `/dashboard` with `target-branch: "develop-v2"`;
- a `docker` entry for `/dashboard` with `target-branch: "develop-v2"`;
- a decision, stated either way, on whether `build/monero|p2pool|tor|xmrig-proxy` should also gain `develop-v2`-targeted entries or keep relying on a `develop`→`v2` sync. **They are not the same case as the dashboard** — those paths exist on both twins, so today's entries do produce PRs; they just produce them on the frozen branch.
- a decision on `tests/integration/tor-client/Dockerfile`, which no entry covers on either twin. It may be deliberate — it is test scaffolding, not shipped — but it is currently a silent omission rather than a recorded one.

**The change must land on `develop`**, since Dependabot only reads the default branch. That is lane-policy exception (b) in `STANDING-DIRECTIVE.md` — the same exception the `/os/rootfs` entry already relies on — and the config's existing comment block is the place to say so, so the next person applying "the twins should be level" mechanically does not undo it.

**Whoever takes this: write the asymmetry down in the file.** The `/os/rootfs` comment is the model, and it exists precisely because moving that entry to `develop-v2` is what makes it stop running.

## Not claimed / not verified

- **Not** verified against Dependabot's own job logs that no PR has ever been attempted for `dashboard/` on `develop-v2` — inferred from the config plus the absence of such a PR and the measured version drift. The Dependabot job log on the default branch would settle it directly.
- **Not** established when coverage lapsed beyond "at #1106, when the directory moved"; I did not bisect.
- **Not** a claim that any specific CVE is currently unpatched in `dashboard/` as a result. The finding is the missing mechanism, not a named vulnerability.
- `build/tari/` and `build/tari-wallet/` exist on `develop-v2` and contain **no** `Dockerfile` (entrypoints and a config template only), so they are correctly absent from the table above rather than a further gap. Checked, because Dependabot's docker ecosystem only reads files literally named `Dockerfile` — the trap `os/rootfs/Dockerfile`'s own header documents.

VijitSingh97:
Closes #1425. Closes #1650.

## The finding, ahead of the diff

`develop-v2` — the branch every change actually lands on — had **no Dependabot coverage beyond the
appliance rootfs**, and the repo gave no sign of it. There is an open Dependabot PR stream, green
CI, and a `dependabot.yml` that reads as complete. The failure is silent in the direction that looks
like success.

Dependabot reads this file from the **default branch only**, and with no `target-branch:` it
resolves each `directory:` against that branch too. So eight of the nine entries watched `develop`,
the frozen twin.

## Two faults, two different repairs

This is the part that a single edit gets wrong, and #1425 is explicit about it:

| fault | entries | why | repair |
|---|---|---|---|
| **wrong directory** | `uv`, `docker` (`/build/dashboard`) | 404 on develop-v2 — #1106 moved the dashboard to `dashboard/` | name the real path |
| **missing target** | `github-actions`, `docker-compose` (`/`) | `/` exists on both twins; nothing is mispointed, the entry just never targets develop-v2 | a second entry with `target-branch:` |

A fix that treats these as one edit repairs the first and silently leaves the second.

## The gap is measured, not reasoned

Read per branch through the contents API, moving no local ref:

| path | develop | develop-v2 |
|---|---|---|
| `build/dashboard/pyproject.toml` | 4116 B | **absent** |
| `build/dashboard/Dockerfile` | 5208 B | **absent** |
| `dashboard/pyproject.toml` | **absent** | 4113 B |
| `dashboard/Dockerfile` | **absent** | 5208 B |
| `build/{monero,p2pool,tor,xmrig-proxy}/Dockerfile` | present | present |
| `os/rootfs/Dockerfile` | absent | 30937 B (already targeted ✔) |

And the drift it has already produced:

| pin | develop | develop-v2 |
|---|---|---|
| `ruff` (#1423, merged) | `0.16.4` | **`0.16.3`** |
| `actions/download-artifact` (#1424, merged) | `3e5f45b2` v8.0.1 | **`37930b1c`** v7.0.0 |

Neither has any mechanism left that would propose the bump again. The `github-actions` row is the
sharpest case: it was level until #1424 merged and went stale **and silent in the same instant**,
because the open PR was the only signal that a newer version existed.

## Why every surface, not just the dashboard — and a claim I got wrong first

I initially concluded from `git merge-base` that no develop→develop-v2 sync was running (merge-base
`ad928a92`, 2026-08-22, develop 15 ahead) and that a security fix was stranded. **That was wrong,
and I am recording it because the reasoning is the useful part.** Re-derived at source: the twins
are kept level by **twin PRs**, not sync merges — #1438 → #1440 (`3afc94ed`, and develop-v2's
`build/tor/Dockerfile:11` does carry the `apk add --no-cache --upgrade libcrypto3 libssl3`), #1313 →
#1356, #1290 → #1344. Ancestry cannot see that mechanism; it only sees that no merge ran. Credit to
the controller for the refutation, which I then checked rather than taking on confidence.

The corrected version is narrower and is this PR's actual claim: **hand-made changes stay level via
twin PRs; a Dependabot raise has no such partner**, because it is opened against the default branch
and nobody hand-writes the twin. That is the whole gap, and it is exactly #1425/#1650.

## The decisions #1650 asked to be stated either way

- **`build/{monero,p2pool,tor,xmrig-proxy}` and `docker-compose.yml`** — exist on both twins and
  their pinned digests are byte-identical today. Carried by twin PRs for hand-made changes, and by
  Dependabot on develop-v2 once this lands. Covered here.
- **`tests/integration/tor-client/Dockerfile`** — covered by no entry on either twin. Left
  uncovered **deliberately**, and now recorded in the file as such: it is test scaffolding, never
  shipped or flashed, so a stale base costs a rebuild rather than a CVE in a user's hands. This
  converts a silent omission into a recorded one.
- **`/os/rootfs` is not repeated** — it already has its own develop-v2 entry, and a second one
  naming the same directory would be a duplicate.

## The one construct I could not confirm from the docs, and how I settled it

`github-actions` at `/` now appears **twice**, differing only by `target-branch`. The `/os/rootfs`
precedent proves the same ecosystem may appear twice, but with a *different* directory — it does not
cover this. GitHub's reference only says values must be unique *"for a single target branch of an
ecosystem"*, which implies but does not state that differing targets are fine.

This mattered enough to settle rather than assume: **the config is validated only once it reaches
the default branch, so a rejected config would silently kill all watching** — worse than the gap
being fixed. Existence proof, from public repos' live configs:

| repo | ecosystem | directory | targets |
|---|---|---|---|
| `jline/jline3` | github-actions | `/` | `master`, `jline-3.x` |
| `openhpc/ohpc` | github-actions | `/` | default, `2.x`, `3.x` |
| `playframework/play-slick` | github-actions | `/` | default, `5.4.x`, `6.2.x` |
| `spring-projects/spring-batch` | github-actions | `/` | default, `5.2.x` |

Four actively-maintained repos using the exact construct at the exact path. Six of twelve configs
sampled used same-ecosystem-same-directory with differing targets.

## Base branch — read at source, not relayed

Lands on frozen `develop` under `STANDING-DIRECTIVE.md` exception (b), lines 50-53, which names
`dependabot` literally: these files "are read from the DEFAULT branch only". **Do not retarget this
to develop-v2** — the config is inert anywhere else, which is the trap the `/os/rootfs` comment
already documents and which this PR extends in the same voice.

## Shared file, and the guard readback the grant required

`.github/dependabot.yml` is a shared workflow-adjacent file, granted to the `currency` lane by the
controller (w89, expiring 2026-09-04T13:00Z). The stage was announced before staging. **No
`.lane-override` was used or present** — the grant is real, and I proved arming rather than assuming
it:

- **Positive** — `.github/dependabot.yml` staged alone → `lane-guard` rc **0**.
- **Negative control** — a *real index change* to an ungranted sibling (`release-gate.yml`) staged
  alongside → rc **1**, naming `release-gate.yml` only and not `dependabot.yml`. Reverted
  immediately; the commit is one file.

The control makes a real change on purpose: the guard reads `git diff --cached`, so staging an
unmodified file is a no-op that returns a false pass.

## What I ran

| check | result |
|---|---|
| `make lint-yaml` (all 14 tracked YAML) | rc **0** |
| yamllint **positive control** — a deliberately broken copy | rc **1**, syntax error at the seeded line |
| `make lint-file-budget` | rc 0, self-tests OK; `dependabot.yml` has no row, 800-line ceiling only |
| `make lint-topology` | rc 0, self-test OK |
| `make lint-operator-strings` | rc 0, self-test OK |
| config re-parsed and the entry table re-derived from the file | 9 entries, 17 directory rows, **0 duplicate `(ecosystem, directory, target-branch)` triples** |

## What I did NOT run, and did not verify

- **`make lint-sh` / `make test` were not run.** This change touches one YAML config and no shell,
  Python or JS. `lint-sh` is the serialized OOM path; running it for a config-only diff is not a
  cost I took. Say so if you want it anyway.
- **No live Dependabot run has exercised this config**, and none can before merge — the config is
  read from the default branch only. The existence proof above is what stands in for that. **The
  first post-merge check that would catch a rejected config is the repo's Dependabot config-error
  surface, and it is worth looking at once this lands.**
- **I did not verify from Dependabot's own job logs** that no PR was ever attempted for
  `dashboard/` on develop-v2 — inferred from the config plus the measured version drift, as #1650
  itself flagged.
- Whether `download-artifact` v7 → v8 is a compatible bump for develop-v2's `ci.yml` is **untouched
  here**. This PR restores the channel that will propose it; it makes no claim about the version.

https://claude.ai/code/session_01TkPjAa3yn7vMQUG56VXqxp

## Over-engineering pass (done by hand before opening; findings noted, not waved off)

**The strongest point first: this adds no new mechanism.** `target-branch:` is already in the file
and already load-bearing for `/os/rootfs`. The diff is four more entries in an existing shape.

Two findings I am recording rather than silently accepting:

1. **Covering `build/*` and `docker-compose` roughly doubles weekly Dependabot PR volume**, against
   a file whose own header says "Grouped + weekly to keep PR volume sane." I kept them, and the
   reasoning is falsifiable: those rows are level today only for *hand-made* changes, which twin
   PRs carry. A **raise** for them has no twin and would strand exactly as #1424 did — the same
   defect, not a speculative one. The duplication is temporary by design and ends when develop-side
   watching is retired (#1169, currently blocked while develop still ships patches). Both the cost
   and the exit are now written into the file.
   **If a reviewer would rather scope this to the dashboard + github-actions rows only, that is a
   coherent position** — say so and I will cut it to those and file the remainder.

2. **The comment block is 33 lines against 72 lines of added config**, which is at the upper end
   even for this file. It matches the house convention here — the existing `/os/rootfs` comment
   runs 10 lines for a 12-line entry, and #1650 explicitly asks that the asymmetry be written into
   the file — but I did not pre-emptively trim it, because the paragraph that took longest to get
   right is the one that stops the next reader from "levelling the twins" and undoing this.
   **A trim is a reasonable review ask**; I would defend the measured paragraph and cut the
   restatement before it.

Considered and rejected as genuine over-engineering: splitting this into two PRs (same file, same
defect class, and the routing was for one), and adding coverage for
`tests/integration/tor-client/Dockerfile` (recorded as a deliberate exclusion instead).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -92,3 +92,108 @@ updates:
       # major/minor is a deliberate migration (compose flags, config compatibility), not a CVE fix.
       - dependency-name: "*"
         update-types: ["version-update:semver-major", "version-update:semver-minor"]
+
+  # ── develop-v2 coverage (#1425, #1650) ──────────────────────────────────────────────────────
+  # Everything above this line resolves against the DEFAULT branch, `develop`. That is not a
+  # style choice: Dependabot reads this file from the default branch only, and with no
+  # `target-branch:` it also resolves each `directory:` there and opens the PR there. So every
+  # entry above watches the FROZEN twin, and `develop-v2` — the branch all work actually lands
+  # on — was covered by the `/os/rootfs` entry and nothing else.
+  #
+  # Two distinct faults produced that, and they need different repairs (#1425):
+  #   * a WRONG DIRECTORY — `uv` and the docker group name `/build/dashboard`, which is 404 on
+  #     develop-v2 (the dashboard moved to `dashboard/` at #1106). Repair: name the real path.
+  #   * a MISSING TARGET — `github-actions` and `docker-compose` name `/`, which exists on both
+  #     twins. Nothing is mispointed; the entries simply never target develop-v2. Repair: a
+  #     second entry carrying `target-branch:`.
+  # A fix that treats these as one edit repairs the first and silently leaves the second.
+  #
+  # DO NOT "level the twins" by deleting these. `develop` is frozen but still takes critical
+  # patches under lane policy exception (a), so its entries stay; and moving an entry to
+  # develop-v2 is what makes it stop running, which is how the gap went unnoticed for so long
+  # (CONTRIBUTING.md § The two lanes; the /os/rootfs comment above states the same mechanism).
+  # The duplication ends when develop-side watching is retired — that is #1169, and it is
+  # blocked while develop still ships patches. Until then, two entries per surface is the cost.
+  #
+  # WHY EVERY SURFACE IS LISTED HERE AND NOT JUST THE DASHBOARD. The twins are held level by
+  # TWIN PRs, not by sync merges: every hand-made change on develop has a develop-v2 partner
+  # (#1438 -> #1440, #1313 -> #1356, #1290 -> #1344), so a merge-base or ahead/behind count
+  # cannot see the mechanism and says nothing about whether the twins have drifted.
+  # What has no twin mechanism is a DEPENDABOT RAISE: it is opened against the default branch,
+  # and nobody hand-writes the partner. Measured 2026-09-03, those are the only two develop-only
+  # changes with no develop-v2 equivalent — ruff (#1423), 0.16.4 on develop against 0.16.3 here,
+  # and actions/download-artifact (#1424), v8.0.1 against v7.0.0 here. The github-actions row is
+  # the sharpest case: it was level until #1424 merged, and went stale and silent in the same
+  # instant, because the open PR was the only signal that a newer version existed. So build/* and
+  # docker-compose are carried by twin PRs for hand-made changes, and by Dependabot on develop-v2
+  # for raises once these entries land.
+  #
+  # `tests/integration/tor-client/Dockerfile` is covered by NO entry on either twin, and that is
+  # DELIBERATE, recorded here so it is an omission on purpose rather than by accident: it is test
+  # scaffolding that is never shipped or flashed, so a stale base there costs a rebuild, not a
+  # CVE in a user's hands.
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    target-branch: "develop-v2"
+    schedule:
+      interval: "weekly"
+    groups:
+      actions-v2:
+        patterns: ["*"]
+
+  # develop-v2's dashboard Python deps. Same subject as the `uv` entry above, different path:
+  # `dashboard/`, not `build/dashboard/` (#1106).
+  - package-ecosystem: "uv"
+    directory: "/dashboard"
+    target-branch: "develop-v2"
+    schedule:
+      interval: "weekly"
+    groups:
+      python-v2:
+        patterns: ["*"]
+    ignore:
+      # Same reason as the develop-side entry: protobuf/grpcio are pinned to what the vendored
+      # Tari gRPC stubs assert at import time (see dashboard/pyproject.toml on develop-v2); a
+      # major bump needs the stubs regenerated first.
+      - dependency-name: "protobuf"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "grpcio"
+        update-types: ["version-update:semver-major"]
+
+  # develop-v2's base-image digests. `/dashboard` is the moved path; the four `build/*` entries
+  # exist on both twins and are repeated here because nothing carries a develop-side bump across
+  # (see the measurement above). `/os/rootfs` is NOT repeated — it already has its own
+  # develop-v2 entry, and a second one naming the same directory would be a duplicate.
+  - package-ecosystem: "docker"
+    directories:
+      - "/dashboard"
+      - "/build/monero"
+      - "/build/p2pool"
+      - "/build/tor"
+      - "/build/xmrig-proxy"
+    target-branch: "develop-v2"
+    schedule:
+      interval: "weekly"
+    groups:
+      docker-v2:
+        patterns: ["*"]
+    ignore:
+      # Same policy as the develop-side docker entry: digest + patch within the pinned tag.
+      # Major/minor base bumps are deliberate migrations, not security updates.
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major", "version-update:semver-minor"]
+
+  # develop-v2's third-party image digests pinned directly in docker-compose.yml. The file exists
+  # on both twins, so this is the missing-target case, not the wrong-directory one.
+  - package-ecosystem: "docker-compose"
+    directory: "/"
+    target-branch: "develop-v2"
+    schedule:
+      interval: "weekly"
+    groups:
+      compose-v2:
+        patterns: ["*"]
+    ignore:
+      # Same policy as the develop-side entry: digest + patch within the pinned tag.
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major", "version-update:semver-minor"]


VijitSingh97:
**Controller non-author pass at head `ec6aa361` — PASS. Merge is by the control seat's admin merge (base `develop` requires one approving review that no lane account can give); currency does not merge this.** Judged on the seat's pinned top tier. PROVEN BY ME unless marked RELAYED.

**Proven by me:**
- **The diff is a pure append** (+105/-0) of four `target-branch: develop-v2` entries; the develop-side entries are untouched, so nothing that fires today stops firing.
- **Every new `directory:` exists on `develop-v2`**, read with `git ls-tree` against `origin/develop-v2`: `/dashboard` (with `pyproject.toml`, `uv.lock`, `Dockerfile`), all four `build/*/Dockerfile`, `/` for compose and actions. The old `/build/dashboard` stays develop-only, which is correct for the develop-side entry.
- **Zero duplicate `(ecosystem, directory, target-branch)` triples** across 9 entries / 17 directory rows, re-parsed from the head file. Control: collapsing the target column produces 6 duplicate pairs, so the checker can see a duplicate when there is one.
- **The construct you could not confirm from the docs holds at source:** `jline/jline3`'s live `dependabot.yml` carries `github-actions` at `/` twice, `target-branch: master` and `target-branch: jline-3.x`, read through the contents API. One of your four; I did not re-read the other three.
- **`directories:` (plural) has an accepted precedent in this very file** — the existing develop-side docker entry uses it and Dependabot is opening PRs from it (#1646), so the form is not a new risk.
- **The base is right.** `STANDING-DIRECTIVE.md` exception (b), lines 50-53, names `dependabot` literally; the file is inert anywhere but the default branch. Do not retarget.
- **The scope offer in your over-engineering pass — RULED: keep every surface.** Ground: the `build/*` and compose rows are level today only through hand-made twin PRs; a Dependabot raise has no twin, which is exactly the #1424 shape the body measured. Cutting to dashboard + actions would fix the visible drift and re-create the silent one.
- All 21 checks green, `MERGEABLE`, `mergeStateStatus: BLOCKED` solely on the review count.

**RELAYED:** the `lint-yaml` positive control, the lane-guard arming readback and the other three public-repo proofs — CI's per-surface lint is green at this head, which corroborates the first.

**F1 — non-blocking, owed AFTER the firing proof:** directive (b) says the change "lands on develop, is proven firing, then syncs forward". `develop-v2`'s own copy of this file already differs from develop's (its `uv` entry names `/dashboard`); once firing is proven, open the develop-v2 twin so the file stays level, and say in it that the copy is inert there.

**The firing proof, stated as a prediction it must satisfy:** a config change on the default branch makes Dependabot run the affected entries at once. Within the hour after merge, `gh pr list --base develop-v2 --author app/dependabot` must show the two measured drifts proposed against develop-v2: `ruff` 0.16.3 → 0.16.4 and `actions/download-artifact` v7 → v8. If neither appears, the config was rejected and the develop-side watching is dead too — check the repo's Dependabot config-error surface before anything else.

**Close mechanics, inverted here:** `Closes #1425` / `Closes #1650` WILL fire because the base is the default branch. Nobody hand-closes; verify both read CLOSED after the merge.

**Not done:** no live Dependabot run can exercise this before merge; `download-artifact` v7 → v8 compatibility for develop-v2's `ci.yml` is untouched and belongs to the PR that proposes it.

Issue #1684, opened 2026-09-03
After a migrating OS update commits, the post-commit `pithead up` can abort on a transitional p2pool and leave every held chain service down — nothing retries


VijitSingh97:
## What happened (proven on the bench, 2026-09-03 ~14:57Z)

A `--phase provision` battery run at the head of PR #1677 (which changes nothing in `os/` or in the migration leg) reddened one row the #1651 full battery had passed at `49b4b9ea` an hour earlier:

```
✓ the migrating slot committed and released the chain services
✓ monerod never ran while the slot was uncommitted
✓ boot journal shows the chain hold
✗ monerod never came back after the commit
```

The guest's own journal for the migrating boot (`journalctl -b -o short-monotonic`, read over the harness key while the leg went on) shows why. The commit and the release line land as designed, then the post-commit `./pithead up` that is supposed to start the held services (`os/overlay/pithead-boot`, the `hold_chain` block after `boot_gate_passed`) dies inside one second:

```
[23.932] pithead-boot: stack is serving (HTTP 401) and doctor reports healthy — booted slot committed
[23.935] pithead-boot: slot committed — chain services released, the data migration runs now
[24.024] [pithead] Starting stack...
[24.793]  Container p2pool Starting
[24.801] podman: Request Failed(Internal Server Error): container f0c274e0… must be in Created or Stopped state to be started: container state improper
[24.803] Error: executing /usr/local/bin/docker-compose up --pull missing -d: exit status 1
[24.809] [ERROR] Stack failed to start — see the error above.
[24.906] [INFO] Verifying system prerequisites...        <- local-miner setup carries on; the boot unit finishes at 28.5 s
```

`f0c274e0…` is the p2pool container. Its state record: started 14:57:31 (by the held `up`), finished 14:57:40 with exit 0, `RestartCount` 0, policy `unless-stopped`; its log ends `uv_run exited, result = 0 … P2Pool stopped`. Eight minutes after the commit `podman ps -a` still read monerod `Exited (0)` (finished 14:56:05, i.e. the reboot before the migrating boot), tari `Exited (137)`, p2pool and xmrig-proxy `Exited (0)`. No timer, unit or dashboard call starts them later: the only container POSTs in the boot are the sync gate's `POST /containers/xmrig-proxy/stop` every ~40 s.

## The mechanism — what is proven and what is inferred

[...]

PR #1686, opened 2026-09-03
fix(os): #1684 the post-commit chain release retries a transitional container and alerts when it cannot start

VijitSingh97:
Closes nothing by keyword (develop-v2); #1684 half 1, per the controller's ruling on the issue.

## What changed

A `data_migration` update boots with the chain services held and releases them after the slot commits with one `./pithead up || true`. On the bench that `up` landed while p2pool — started by the held `up` — was mid-stop; compose aborted the WHOLE start on podman's "container state improper", the failure was swallowed, and every held service stayed down under a dashboard reading `updated`, with nothing retrying short of a reboot a committed appliance has no reason to take.

- `os/overlay/pithead-boot`: the release is now `release_held_chain_services`, a function above the sourcing boundary. It removes the markers first (as before), prints the release line the tier-4 battery keys on BEFORE the first `up` (as before), then retries the `up` up to `PITHEAD_RELEASE_UP_TRIES` (5) times `PITHEAD_RELEASE_UP_PAUSE` (10) seconds apart, announcing each failure short of the last. After the last try it writes ONE `FAULT` line on stderr saying the chain services did not start and the migration has NOT run, naming the recovery (`./pithead up` from `/data/pithead`), and returns 1. The call site keeps `|| true`: a chain service that cannot start is never grounds to fail a boot the commit already accepted. The ordinary (non-hold) boot path is untouched.
- `tests/stack/test-appliance-boot-release.sh` (new sibling, 102 lines) + a two-line `run.sh` stanza mirroring the #1265 remint sibling.
- `docs/dev/appliance-release.md` (the migration-hold paragraph now says the `up` is retried and what the journal carries when it cannot), `docs/dev/testing-strategy.md` (one tier-1 row).
- `docs/dev/file-budget.tsv`: `pithead-boot` goes 400 -> 420 lines and takes a NEW row at its true count. The function is one unit; the only way under 400 was trimming twenty lines of other people's prose, which I did not do.

Half 2 (starting only the held services, `lib/pithead/01-lifecycle.sh`) is deliberately NOT here: not granted, and it rests on an inferred stop reason.

## What was RUN

- The new file standalone (sourced `lib.sh`, fresh sandbox): **23 passed, 0 failed.**
- Mutation battery on `pithead-boot`, each mutant proven applied by `git diff --numstat` and the file restored by byte-compare against the pre-battery copy:

| mutant | rows killed | anything else red |
|---|---|---|
| one bare `up`, no loop | 7 | no |
| retry announcement dropped | 3 | no |
| markers removed only after a successful `up` | 1 (`the marker was still consumed`) | no |
| release line printed after the `up` | 2 (order rows) | no |
| call site back to the bare swallowed `up` | 1 (call-site text row) | no |

- `make lint-file-budget` at the branch head: green (the new row is the one addition). `tests/inventory.sh`: the file is listed with one section.
- `shellcheck --severity=warning` and `shfmt -i 4 -d` on the three shell files touched, run single-file: clean. **Full `make lint-sh` was NOT run here** — it queues behind the KVM guest's flock; CI carries it.
- `make lint-md lint-docs-voice lint-operator-strings lint-topology`: green.
- Full `tests/stack/run.sh` in this worktree: rc 0, `pithead tests: 3160 passed, 0 failed`, ✗ count 0. Launched at 15:20:46Z on the exact tree committed as 57be0467 (no edit between launch and commit; run from a worktree checked out at the head, not a reconstruction). Includes the #1393 data-floor file, green.

## What was NOT run

- No bench run at this tip. #1684's own close condition is the migration leg's `monerod never came back after the commit` row green on a battery carrying this fix, and the issue says a single green does not close it (the #1651 green on the same `os/` was timing). That battery is a separate bench item after the two legs the bench is holding now.
- `tests/stack/test-appliance-data-floor.sh` (#1393, not this lane's path) still passes — its `commit-side removal` text assertion counts exactly one `rm -f .os-migration-pending .os-data-floor.prev` and there is still exactly one — but its comment "both call sites sit below pithead-boot's sourcing boundary" is now true of the fallback side only: the commit side moved into `release_held_chain_services`, above the boundary, where the new file drives it. One stale comment line in a file I did not touch; flagged for the tests lane rather than edited from here.
- `shfmt` reports a pre-existing alignment diff in `pithead-boot` at the `gate_resolve_spec` case arms (untouched by this PR, present at the tip); left alone.

The ponytail gate reads this lane's pinned cwd and names a dead branch, so its pass is not evidence; here is the pass. Alternatives weighed: a podman restart policy cannot help — p2pool exited 0 on its own and the services compose never reached were never created; re-running the boot unit would re-run the gate on a committed slot; starting only the held services is half 2 and not granted. A bounded retry with a named fault line is the smallest change that turns a veto into a delay. Two knobs exist only so tier 1 runs in milliseconds; both default to the shipped values and the test pins those defaults.

## Lane

`os/` and `tests/stack/run.sh` are this lane's armed paths; `tests/stack/test-appliance-boot-release.sh` was armed per-file by the controller for #1684 (same route as the #1265 and #1393 siblings), expiring on this PR's squash-merge. Arming readback, run by me against the shared pre-commit guard: the granted file staged ALONE passes (`guard rc=0`); an ungranted sibling `tests/stack/test-appliance-boot-release-control.sh` staged alone is refused by name (`lane-guard: role appliance does not own these staged paths`), then unstaged and deleted — never committed. `docs/` is shared. The tests lane is not running; the controller was told about the `run.sh` stanza instead.


VijitSingh97:

**Re-derived by me:**
- `pithead-boot` diff read in full. The release is a function above the sourcing boundary; markers go before the first try; the release line prints before the first `up`; failures short of the last are announced; the last is one `FAULT` line on stderr naming the recovery; the function returns 1 and the call site's `|| true` keeps the boot. errexit-safe: the script is `set -uo pipefail` (no `-e`), and every failing command sits inside an `&&`/`||` list anyway.
- The new test file read in full and **run standalone by me at `57be0467`: 23 passed, 0 failed.** Two of the body's mutants re-applied by me (each proven applied by `git diff --numstat`): loop reduced to one try → **7 reds**; markers moved after a successful `up` → **1 red** (`the marker was still consumed`). Both match the PR's table exactly.
- Ruling 8a, both legs at `30d37972` against the current tip: budget rc 0 (5 self-tests fired), parity rc 0, at head and at the merged tree; `merge-tree` rc 0. The new tsv row `os/overlay/pithead-boot 420` equals the file's true length (420 lines, counted) — legal, zero headroom (w66 ruling: the next addition to that file splits).
- **Disclosed: I moved this branch.** At `57be0467` it conflicted with the tip in `docs/dev/testing-strategy.md` only (#1677 edited the #1393 row this PR's new row sits beside). I merged `develop-v2` into the branch, keeping the tip's #1393 row followed by this PR's #1684 row (merge commit `30d37972`, docs-only, no code touched); `git diff --stat` tip..head afterwards is exactly the PR's six files. CI re-ran on the merged tree and is green.
- Grant readback in the body is consistent with the armed clause; that clause expires on this merge and is retired as EXPIRED with it.

**Relayed, not re-run:** the full `tests/stack/run.sh` 3160/0 at `57be0467` (CI's Shell-tests job on `30d37972` is the independent pass leg) and the other three mutants.

**Does not close #1684**, per the ruling on the issue: the close needs the migration leg's `monerod never came back after the commit` row green on a battery at a tip carrying this fix.

Issue #1653, opened 2026-09-03
GA gate, unfiled: the release lane has never been exercised end-to-end, and a release lane's failure mode is a seam between two steps that both work


VijitSingh97:
## Why this is an issue

The third of #394's three unticked gates that are **runs, not code**, alongside the KVM battery and
the 7-day soak. None was an issue, so none had a milestone or an owner and a milestone-keyed queue
could not see them. #394's wording:

> Release lane per `docs/dev/appliance-release.md` exercised end-to-end for the first public cut

## What makes this one different from the other two

The battery and the soak test the **image**. This tests the **process that publishes it**, and it is
the only one of the three that has never been run at all — by definition, since there has been no
first public cut. Everything it exercises is therefore unproven in exactly the way that is hardest to
notice: each step has probably been run in isolation, and the failure mode of a release lane is
almost always a seam between two steps that both work.

Live examples of that seam class already on file in this repo: the cosign consumer and producer halves
being upgradeable independently, and a release gate that could not restore the box because a restore
step used a banned command pairing. Both were seams, neither was a broken step.

## Scope

Exercise it end to end, once, against `docs/dev/appliance-release.md`, for the cut that will actually
ship — signing (cosign is mandatory; dev builds refuse without `--dev`), bundle production, publishing,
and the consuming side that installs the published artifact on a machine that did not build it.

The consuming side is the part worth being strict about: **an update that installs from the artifact
you published, on a box that has never seen your build tree.** That is the only leg that can catch a
bundle which is correct locally and unusable once published.

## Ordering

Last of the three. It runs on the artifact the battery passed and the soak ran, because publishing a
different artifact than the one that was proven defeats all three gates at once.

## What was NOT checked

I have not exercised any part of the release lane and I am reporting no result — this files the gate
and its ordering. Whether the first public cut is the right moment to also be discovering seams is a
sequencing question for the operator; the alternative is a dry run on a throwaway tag first, which
costs a day and removes that risk from the cut itself. I recommend the dry run.

VijitSingh97:
## Rehearsal — appliance release lane, dry-run only, against `develop-v2` @ `49b4b9ea`

Runbook: `docs/dev/appliance-release.md` § *Cutting a release* (which explicitly rides the DIY
channel's `scripts/release.sh`, per that section's own opening line). Nothing tagged, pushed,
signed, published, or uploaded; `VERSION`/`CHANGELOG.md` untouched throughout. Rehearsed from a
[...]

PR #1687, opened 2026-09-03
fix(lint-sh): #1679 the shell gate refuses any shellcheck but the pinned one, from one source of truth

### LINKED ISSUE #1679: make lint-sh runs whatever shellcheck is on PATH, so a distro v0.9.0 reds the tip on SC2218 where the pinned v0.11.0 is clean
## What happens

`make lint-sh` runs whatever `shellcheck` is first on PATH. CI pins and sha256-verifies shellcheck v0.11.0 (`.github/workflows/ci.yml`, the "Install shellcheck + shfmt" step). A box with the distro package (v0.9.0) reds the current `develop-v2` tip:

```
tests/integration/selftest-mergemine-probe.sh:159: SC2218 (error): This function is only defined later.
make: *** [Makefile:60: lint-sh] Error 1
```

while v0.11.0 over the same files, same invocation, is rc 0 with no output. Both versions fire SC2218 on a three-line control, so the split is real behaviour, not a missing check.

## Why it matters for the release lane

`docs/dev/appliance-release.md` § Cutting a release, step 1: "The release commit is green: `make lint && make test`". The #1653 rehearsal stopped at that step on exactly this false red. A cutter cannot tell a real lint failure from a version mismatch, because nothing prints which shellcheck ran or that it differs from the pin.

## Fix

The version the gate means lives in ONE place and `lint-sh` refuses to run on any other:

- `make lint-sh` reads the pinned version (today it is literal text inside ci.yml) and fails fast with a one-line message naming the found and the expected version before running anything.
- The runbook step names the pinned version, so a cutter knows what to install.

Both `Makefile` and `ci.yml` are shared files under the serialization rule; say so in the PR body.

## Not in scope

Changing what shellcheck checks. The v0.11.0 result at the tip is clean and CI is green on #1594 and #1675; this issue is about the gate being silent on which tool it ran.

Found by the #1653 dry-run rehearsal; re-derived by the controller with both binaries and a fired control.

VijitSingh97:
## What this fixes

`make lint-sh` ran whatever `shellcheck` was first on PATH. A box with the distro package (v0.9.0) reds the `develop-v2` tip on `tests/integration/selftest-mergemine-probe.sh:159` SC2218, where the pinned v0.11.0 over the same files is rc 0 and silent. The #1653 release rehearsal stopped at `docs/dev/appliance-release.md` § Cutting a release step 1 on exactly that false red, and nothing printed which shellcheck had run or that it differed from the pin.

## The change

- `Makefile` gains `SHELLCHECK_VERSION := 0.11.0` — the one place the pinned version is written.
- `lint-sh` reads `shellcheck --version` and refuses **before linting anything**, naming the version it found beside the one it wants.
- `ci.yml`'s installer derives its download URL from `make -s print-shellcheck-version` instead of carrying the literal `v0.11.0` twice, so the two cannot drift.
- Both runbook steps name the pin and point at the command that prints it.

Deliberate design note: ci.yml's sha256 line stays a literal tied to one tarball. Bumping the Makefile alone therefore fails the checksum **loudly** rather than silently installing an unverified build. The Makefile comment says so.

## Evidence — the controlled pair, all run on this box

| Leg | shellcheck the guard saw | rc | guard message | shellcheck lint invocations |
|---|---|---|---|---|
| **FIRE** | real system 0.9.0 | 2 | yes — names `0.9.0` vs pinned `0.11.0` | **0** |
| **PASS** | real 0.11.0 | 0 | none | **2** (+1 shfmt) |
| **FIRE** | no version output (absent-proxy) | 2 | `shellcheck not found is not the pinned 0.11.0` | **0** |

The PASS leg's version handshake is served by the **genuine v0.11.0 binary**, downloaded from the same URL `ci.yml` uses and sha256-verified against ci.yml's own checksum (`8c3be12b…7198` → `sc.tar.xz: OK`). `shellcheck --version` is the guard's *entire* input, so the real binary covers the whole decision surface. That leg's lint invocations were short-circuited only to stay off the serialized OOM path; the linting itself is unchanged by this PR and is what CI's own `make lint-sh` step exercises.

Checked **before** writing the guard, because it was the one claim that could red CI for everyone: real 0.11.0 prints `version: 0.11.0`, and the Makefile's `awk '/^version:/ {print $2}'` extracts `0.11.0` from it. The output format did not change between 0.9.0 and 0.11.0.

The instrument discriminates in both directions: the FIRE legs show 0 lint invocations (the guard stopped execution), the PASS leg shows 2 (execution continued past it).

## Shared files — disclosure

- **`Makefile`** — shared/serialized. The lane-guard **blocks** it for this lane, so this commit spends a one-shot `.lane-override`, touched immediately before the commit and confirmed consumed. The guard named `Makefile` and nothing else, so the broad hatch let through only the intended file. Expect a conflict.
- **`.github/workflows/ci.yml`** — currency's standing named-file grant; announced to the controller before staging, and the route was accepted.
- **`docs/dev/appliance-release.md`**, **`docs/dev/release-server.md`** — `docs/` is `_shared`.

Lane-guard arming, each file staged **alone**:

| file | verdict |
|---|---|
| `Makefile` | BLOCKED (hence the override) |
| `.github/workflows/ci.yml` | ALLOWED |
| `docs/dev/appliance-release.md` | ALLOWED |
| `docs/dev/release-server.md` | ALLOWED |
| `.github/workflows/release-gate.yml` — **negative control** | BLOCKED |

## Budget

`ci.yml` had **one line of headroom** (512 against a 513 ceiling, and ceilings only ever go down). The edit is deliberately **line-neutral: 512 → 512**, so the headroom is preserved rather than spent. `Makefile` has no row (155 → 171, far under the 800 hard ceiling). Every `*.md` file is budget-exempt.

## Gates run

`lint-yaml` PASS · `lint-md` PASS · `lint-docs-voice` PASS · `lint-file-budget` PASS · `lint-operator-strings` PASS · `lint-topology` PASS · `zizmor@1.25.2 --offline` PASS (the same pin CI runs).

**Not run:** `make lint-sh` to completion over the real file list with a real shellcheck. That is the serialized OOM path, and this box has the wrong version regardless. CI's shell job is what proves it.

## What this does NOT do

- **shfmt is not covered, and the gap is LATENT rather than active — measured, not assumed.** Its pin (3.13.1) is still a literal in `ci.yml` and this box runs 3.8.0, so the *mechanism* is identical: the gate formats with whatever is on PATH. The *consequence* is not. I ran the real `shfmt -i 4 -d` invocation from `lint-sh` with 3.8.0 over the same file list: **rc 0, zero bytes of diff, zero files**. So shfmt reds nothing here today and there is no second false red to chase. I had written "the identical failure mode" here before measuring it; that was too strong and this is the corrected claim.
- Does not change what shellcheck checks. No file, severity or exclusion moved.
- **Behaviour change for local dev, named rather than buried:** `.pre-commit-config.yaml` runs `make lint-sh`, so a contributor on a distro shellcheck is now refused at pre-commit with the version message instead of silently getting different findings. That is the intent, but it is a change.
- The issue asked for "a one-line message"; the guard prints two — line 1 names found and expected, line 2 names the remedy. Flagging the deviation rather than claiming a match.

---

Fixes #1679 — but pithead's default branch is `develop` while this lands on `develop-v2`, so the keyword will **not** fire. Close by hand after merge.

---

## Over-engineering pass (ponytail gate, by hand)

Five things I put to the diff, and where each landed.

1. **Does `print-shellcheck-version` earn a target, or should ci.yml just grep the Makefile?** Kept. The grep alternative (`grep '^SHELLCHECK_VERSION' Makefile | cut …`) is longer, parses a Makefile with a regex, and inverts the natural direction. ci.yml's *existing* comment already says "Single source of truth: the Makefile `lint-sh` target" — deferring to a make target is the idiom in this exact file, not a new one.
2. **Is refusing over-built where printing would do?** No — printing is the thing that fails. A version that is only printed is not checked, and the issue asks for a refusal in those words.
3. **Two `echo` lines where the issue said "a one-line message".** Kept, and disclosed above. Line 1 carries everything the issue asked for; line 2 is the remedy, which is the actual complaint from the #1653 rehearsal (a cutter who cannot tell what to install).
4. **Five comment lines for a one-line variable.** Kept. They carry three facts a reader cannot recover from the code: why the value lives here, that ci.yml derives from it, and that the sha256 line will fail loudly on a lone bump. The surrounding `lint-sh` block runs ~20 comment lines, so this matches the file, not exceeds it.
5. **Scope.** Two extensions were available and both were declined rather than absorbed: shfmt's identical skew (would roughly double the diff, and is not what #1679 asks) and `scripts/release.sh`'s preflight (it checks presence, and `make lint` now refuses downstream of it anyway, so the message a cutter sees is already the right one).

One duplication I looked at and left: the two `@#` comment lines inside the recipe overlap slightly with the variable's comment block. They answer different questions — *why the check runs first* versus *what the pin is* — so trimming them would cost a reader more than it saves. Recorded rather than silently kept.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -423,17 +423,17 @@ jobs:
       - name: "Install shellcheck + shfmt (pinned + sha256-verified, #286)"
         # Direct upstream-release downloads with a sha256 check (RigForge convention) instead of the
         # runner's preinstalled shellcheck / apt — reproducible, and immune to the apt-mirror
-        # flakiness of #64. Keep these pins in sync with the local-dev tooling.
+        # flakiness of #64. shellcheck's version comes from the Makefile (#1679); shfmt's is hand-kept.
         run: |
           curl -fsSL -o /tmp/shfmt \
             https://github.com/mvdan/sh/releases/download/v3.13.1/shfmt_v3.13.1_linux_amd64
           echo "fb096c5d1ac6beabbdbaa2874d025badb03ee07929f0c9ff67563ce8c75398b1  /tmp/shfmt" | sha256sum -c -
           sudo install -m 0755 /tmp/shfmt /usr/local/bin/shfmt
-          curl -fsSL -o /tmp/sc.tar.xz \
-            https://github.com/koalaman/shellcheck/releases/download/v0.11.0/shellcheck-v0.11.0.linux.x86_64.tar.xz
+          SC=$(make -s print-shellcheck-version) # the Makefile is the pin; `make lint-sh` below refuses any other version (#1679)
+          curl -fsSL -o /tmp/sc.tar.xz "https://github.com/koalaman/shellcheck/releases/download/v$SC/shellcheck-v$SC.linux.x86_64.tar.xz"
           echo "8c3be12b05d5c177a04c29e3c78ce89ac86f1595681cab149b65b97c4e227198  /tmp/sc.tar.xz" | sha256sum -c -
           tar -xJf /tmp/sc.tar.xz -C /tmp
-          sudo install -m 0755 /tmp/shellcheck-v0.11.0/shellcheck /usr/local/bin/shellcheck
+          sudo install -m 0755 "/tmp/shellcheck-v$SC/shellcheck" /usr/local/bin/shellcheck
       - name: Lint pithead, build/* + dashboard/ container scripts, and test scripts (shellcheck + shfmt)
         # Single source of truth: the Makefile `lint-sh` target (so the file list can't drift
         # between here and the Makefile). Covers build/*/*.sh and dashboard/*.sh — the entrypoints
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 # Local test entry points (mirror the GitHub Actions CI jobs).
-.PHONY: test test-dashboard test-frontend test-patch-coverage test-stack test-compose test-integration test-integration-selftest test-fakes test-mini-stack lint lint-sh lint-py lint-js lint-yaml lint-md lint-proto lint-toml lint-topology lint-file-budget lint-pithead-parity lint-trivy-parity release release-smoke
+.PHONY: test test-dashboard test-frontend test-patch-coverage test-stack test-compose test-integration test-integration-selftest test-fakes test-mini-stack lint lint-sh lint-py lint-js lint-yaml lint-md lint-proto lint-toml lint-topology lint-file-budget lint-pithead-parity lint-trivy-parity print-shellcheck-version release release-smoke
 
 test: lint test-dashboard test-frontend test-stack test-compose test-integration-selftest test-fakes ## Run everything that doesn't need a server/docker
 
@@ -44,6 +44,16 @@ test-inventory: ## Write the test coverage inventory to docs/dev/test-inventory.
 test-integration: ## Run the live config-matrix integration suite (requires a test box; pass ARGS=...)
 	bash tests/integration/run.sh $(ARGS)
 
+# The shellcheck version this repo's lint gate MEANS, and the ONE place it is written (#1679).
+# `lint-sh` refuses to run on any other version, and ci.yml's installer derives its download URL
+# from this variable (`make -s print-shellcheck-version`), so the pin cannot drift between the two.
+# Bumping it here is most of the bump: ci.yml's sha256 line is tied to one tarball and fails loudly
+# until it is updated to match. shfmt is NOT covered — its pin is still a literal in ci.yml.
+SHELLCHECK_VERSION := 0.11.0
+
+print-shellcheck-version: ## Print the pinned shellcheck version (ci.yml's installer reads this)
+	@echo $(SHELLCHECK_VERSION)
+
 lint: lint-sh lint-py lint-js lint-yaml lint-md lint-docs-voice lint-operator-strings lint-topology lint-file-budget lint-pithead-parity lint-trivy-parity lint-proto lint-toml ## Lint/format-check every surface
 
 # Two shellcheck invocations, not one, and the split is load-bearing. shellcheck resolves a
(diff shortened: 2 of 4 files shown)


VijitSingh97:

**Re-derived by me:**
- Diff read in full (4 files). The guard runs before either shellcheck invocation, reads only `shellcheck --version`, and names found-vs-pinned plus the remedy. `ci.yml` derives the URL from `make -s print-shellcheck-version`; the sha256 stays a literal, so a lone Makefile bump fails loudly, as the body says.
- **The FIRE leg, re-run on this box:** distro shellcheck 0.9.0 first on PATH, `make lint-sh` at the branch head under a 2G memory cap with the serialize lock held. It refused with the two-line message naming `0.9.0` vs `0.11.0` and made **zero** lint invocations (rc 2). The cap was the control against a guard that did not fire reaching the OOM path; it was never hit.
- **The PASS condition:** the pinned 0.11.0 binary on this box prints `version: 0.11.0` and the Makefile's `awk` extracts exactly the pin. CI's Shell-tests job on this head is the full PASS leg: the installer step resolved the version through the new target and `make lint-sh` ran the guard against the real 0.11.0 and linted clean.
- Ruling 8a, both legs vs the current tip (merge-base `9f417d14`, four behind): budget rc 0 with 5 self-tests fired, parity rc 0, at head and at the merged tree; `merge-tree` rc 0. `ci.yml` line-neutral at its 512/513 row, confirmed by the gate, not by count.
- Lane-guard table: Makefile BLOCKED (one-shot hatch, disclosed), ci.yml ALLOWED under the standing grant, negative control BLOCKED — consistent with COMMON's serialized-file rule as measured at w93.

**Relayed, not re-run:** the absent-shellcheck FIRE leg and the shfmt "latent" measurement.

**Consequence named for the fleet (not a finding against the PR):** any box whose first `shellcheck` on PATH is not 0.11.0 now refuses `make lint` at `lint-sh` with the message, instead of reporting a different shellcheck's findings. This box's shim resolves to the pinned 0.11.0, so lanes here pass.

Base is `develop-v2`, so `Fixes #1679` will not fire; closed by hand after merge.

Issue #1441, opened 2026-08-28
lint-docs-voice reports a pass when its enumeration returns nothing


VijitSingh97:
`scripts/lint-docs-voice.sh` exits 0 when its enumeration returns nothing, so a broken glob is a green gate over a scan that never happened. Same family as the operator-strings legs (#1431, #1432, #1434), left in place while the siblings were fixed.

## The shape

```
:16   if [ -z "$files" ]; then
:17       echo "docs voice: no prose docs to check"
:18       exit 0
:19   fi
```

The enumeration at `:11-14` is `git ls-files '*.md'` through three `grep -v` filters, with `|| true` on the end. An empty result is treated as a clean tree. It never is: the repo has tracked prose docs, and the only ways to get zero are a broken glob, a filter that over-matches, or running outside a checkout — every one of them a fault.

## Measured, with the control

```
intact tree ................................ rc=0   "docs voice OK — no banned words in 40 prose docs"
git repo with no tracked .md ............... rc=0   "docs voice: no prose docs to check"
```

The first row is the control: the same command on a real tree gives rc 0 too, and **CI judges the step on its exit code**, so both are green and the second scanned nothing.

## What is different here, and it is worth being precise about

Unlike #1434, the two runs are **not** byte-identical on stdout — the empty case says "no prose docs to check" rather than claiming it checked. A human reading the log could tell them apart. So this is the weaker member of the family: the false green is real for any automated consumer, and only the log text saves a human who happens to read it.

That is still the wrong side of the line the sibling script now sits on. For comparison, on the same tree:

```
scripts/lint-operator-strings.sh, target absent ....  rc=1, names the missing file
```

## Fix shape

Mirror `enforce_nonempty_frontend`, which already exists in `scripts/lint-operator-strings.sh` for exactly this — refuse rather than pass, and say why:

```
echo "docs voice: the prose-doc enumeration returned zero files." >&2
echo "A broken enumeration and a clean scan both report zero hits — refusing to call either a pass." >&2
exit 1
```

Two things learned from the sibling fixes that apply directly:

- **Score any test on the printed message, not the exit code.** A correct refusal and a script that dies anonymously both exit 1, so an rc-only assertion passes the broken version.
[...]

PR #1682, opened 2026-09-03
test(os): #1676 leg 4 ages a x.y.0 running version across the minor boundary, and a failed aging costs one red

VijitSingh97:
Closes nothing by keyword (everything lands on `develop-v2`; the controller closes #1676 by hand).

## What changed

`tests/os/run.sh` leg 4 makes the guest claim a version OLDER than the bundle it is about to install, because both images come from one checkout and the dashboard door refuses an equal target on purpose. The helper that produced that version only decremented the PATCH component and gave up at 0 — and every minor release-prep tip is `x.y.0`, so the release gate (#1651) could not age the version on exactly the tips it exists to prove. Worse, the `bad` at the aging site never returned, so every later step failed on the same un-aged guest: seven reds in leg 4 for one cause.

- **`tests/os/aged-version.sh` (new, 107 lines, under the 400-line budget target so no tsv row):** the helper as a pure function over `X.Y.Z`. It steps down the lowest non-zero component and zeroes the ones below it (`1.20.1 -> 1.20.0`, `1.20.0 -> 1.19.0`, `2.0.0 -> 1.0.0`), and fails closed — rc 1, prints nothing — on `0.0.0` or anything not `X.Y.Z`, so the caller can never write a half-parsed value into the guest. Any value the downgrade guard orders below the bundle serves: `semver_newer` compares per component numerically (`%d%05d%05d`), and the floor leg plants its own `99.0.0` floor separately. Its `--self-test` drives 13 cases and checks every aged value against an independent ordering (`sort -V`) as well as against the expected string.
- **`tests/os/run.sh`:** sources the sibling, ages `${tag#v}` (already the checkout's VERSION, read once at the top of the leg), and returns after either aging red with the bench release server torn down — the same shape as the leg's three other early returns. 3437 -> 3423 lines: the helper and its comment moved out (-24), the source pair (+2) and the two returns (+8).
- **`tests/stack/test-harness-tooling.sh` — SHARED file (`roles/_shared.paths`), disclosed per the serialization rule, not a grant:** a two-line driver for the self-test beside the existing `failure-evidence.sh` row, same reason as that row — tier 1 is the lowest tier that proves it and it needs no KVM. No open PR touches this file (checked per open PR's file list at creation time). 84 -> 94 lines, unrowed.
- **`docs/dev/file-budget.tsv`:** the one row this change touched, `tests/os/run.sh` 3437 -> 3423. `--generate` also wanted to lower four rows I did not touch; those are left alone.

[...]

PR #1690, opened 2026-09-03
fix(lint): lint-docs-voice refuses an empty prose-doc enumeration

VijitSingh97:
## The defect

`lint-docs-voice.sh` treated an empty `git ls-files '*.md'` enumeration as a
clean scan: `docs voice: no prose docs to check`, exit 0. A broken filter
(wrong glob, an exclusion that over-matches, running outside a checkout) and a
genuinely clean tree both report zero hits — the gate could not tell them
apart, so it silently passed the broken case.

## What the gate refuses now

Zero files from the enumeration is treated as broken, never clean — this repo
always has tracked prose docs, so zero is never a legitimate result:

```
docs voice: the prose-doc enumeration returned zero files.
A broken enumeration and a clean scan both report zero hits — refusing to call either a pass.
```
(exit 1)

Same shape as `lint-operator-strings.sh`'s `enforce_nonempty_frontend` guard.

## Deciding commands

```
bash scripts/lint-docs-voice.sh --self-test   # runs the real script end to end
                                                # in a throwaway empty-repo checkout
make lint-docs-voice
```

## What the two added assertions check

In `tests/stack/test-harness-tooling.sh`:

1. `bash "$ROOT/scripts/lint-docs-voice.sh" --self-test` exits 0 — the
   self-test itself passes.
2. Running the **unflagged** script directly against a throwaway empty git repo
   exits 1 and names the empty enumeration in its output. This is the one that
   actually proves the guard: a revert that drops the guard but leaves
   `--self-test` recognized as a no-op would still pass assertion 1 (it'd just
   run the ordinary scan against this real checkout, which has real tracked
   docs) — so assertion 2 drives the real, unflagged code path with no
   dependence on the script knowing its own flag.

Verified independently: both assertions are RED against the parent commit
(`9791a693`, pre-#1441) and GREEN on this branch.

Refs #1441

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,6 +106,7 @@ lint-md: ## markdownlint over all Markdown (config: .markdownlint-cli2.jsonc)
 	npx --yes markdownlint-cli2@0.18.1
 
 lint-docs-voice: ## Fail if banned marketing words appear in prose docs (house voice: docs/dev/STYLE.md)
+	bash scripts/lint-docs-voice.sh --self-test
 	bash scripts/lint-docs-voice.sh
 
 lint-operator-strings: ## Fail if a #NNN issue/PR number or a bare docs/ path leaks into pithead or dashboard operator-facing text (#755, #1024)
--- a/scripts/lint-docs-voice.sh
+++ b/scripts/lint-docs-voice.sh
@@ -13,10 +13,8 @@ if [ "${1:-}" = "--self-test" ]; then
     tmp=$(mktemp -d)
     trap 'rm -rf "$tmp"' EXIT
     self=$(cd "$(dirname "$0")" && pwd)/$(basename "$0")
-    empty="$tmp/emptyrepo"
-    mkdir -p "$empty"
-    git init -q "$empty" >/dev/null
-    out=$(cd "$empty" && bash "$self" 2>&1) && rc=0 || rc=$?
+    git init -q "$tmp" >/dev/null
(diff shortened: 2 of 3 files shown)


VijitSingh97:

**Re-derived by me:**
- Diff read in full (3 files). The guard turns an empty enumeration into rc 1 with a two-line refusal on stderr; `--self-test` runs the real, unflagged script end to end in a throwaway empty repo; the suite drives both the self-test and the unflagged path directly, and the body's reasoning for the second assertion (a revert that drops the flag makes the first assertion vacuous) is right.
- **The negative control, re-run by me:** the parent commit's script (`9791a693`) in an empty git repo exits 0 printing `no prose docs to check` — the defect; this head's script exits 1 naming the empty enumeration. Same empty repo, only the script differed. The self-test also passes from this checkout.
- Ruling 8a, both legs vs the current tip (merge-base `9791a693`): budget rc 0 with 5 self-tests fired, parity rc 0, at head and at the merged tree; `merge-tree` rc 0.

**Merge-record note, not a blocker:** `Makefile` is a serialized shared file (COMMON) and the body does not say so. The one-line addition runs the self-test ahead of every `make lint-docs-voice`, which costs a `git init` in a temp dir per lint — acceptable. `tests/stack/test-harness-tooling.sh` is `_shared.paths`; #1682 also adds a block to it, in a different hunk.

Base is `develop-v2`, so `Refs #1441` closes nothing; #1441 is closed by hand on the re-derived control above.

Issue #1696, opened 2026-09-03
an unauthenticated rig can abort a poll step with a lone surrogate in change_id


VijitSingh97:
Found while fixing #1566, in the same file, but it predates that change and is independent of it.

`worker_config_change_known` passes the rig-reported `change_id` straight to sqlite as a query
parameter:

    cursor.execute("SELECT 1 FROM worker_config WHERE change_id = ? LIMIT 1", (change_id,))

`change_id` arrives on the unauthenticated enriched worker feed and
`parse_worker_control_status` validates it only as a non-empty `str`. A rig's JSON can legally
carry `\ud800`; `json.loads` hands that back as a lone surrogate, and sqlite3 refuses to encode one
— `UnicodeEncodeError` at `dashboard/mining_dashboard/service/worker_config_store.py:242`, raised
inside `_reconcile_worker_config` during the poll.

So an unauthenticated LAN device can abort that step of the poll by reporting one character.

## Measured

Driving the real `_reconcile_worker_config` against a real in-memory `StateManager`, one worker,
one poll, with a control leg for the other component position:

    surrogate in change_id  -> RAISED UnicodeEncodeError at worker_config_store.py:242
    surrogate in worker     -> OK, row recorded

The second leg is the control: it shows the harness records a row normally, so the first leg is a
real raise rather than the fixture failing to arm.

## Scope

Not fixed in the #1566 PR deliberately: it is a different defect class in a path that fix does not
touch, and the #1566 change is already the reason it was found rather than a cause of it. That PR
does handle the surrogate on its OWN half — the worker name reaches the id escape and is
percent-encoded there — and its docstring points at this issue for the change_id half.

Worth deciding as part of a fix: whether to reject a non-encodable `change_id` at
`parse_worker_control_status` (with the other feed validation, refusing early), or to make this
store call fail closed the way its siblings do.

## What I did not check

Whether the same input can reach other sqlite parameter binds on this feed — I looked only at the
path that raised. A fix should sweep them rather than patch this one call.

VijitSingh97:
Swept the open question in **What I did not check** — whether the same input reaches other sqlite
parameter binds on this feed. It does, but not where the obvious guess puts it.

**Method, and its limit: this is derived by READING the code paths, not by execution.** The original
report measured its claim against a real `StateManager`; I could not. This lane is under a CI-only
rule right now (another lane holds the shared build box), so nothing below was run. Treat it as a
[...]

Issue #1695, opened 2026-09-03
the out-of-band audit flood cap is keyed on a worker name the device chooses


VijitSingh97:
Split out of #1566, which carries two shapes. Shape 1 (the id join) is fixed by the PR that closes
that issue; this is shape 2, filed separately because it is a different defect class and the fix
has a design decision in it that the id fix does not.

## The defect

`_rig_edit_within_cap` in `dashboard/mining_dashboard/service/data_service.py` buckets the #724
flood cap on `worker` — the name the device presents on the unauthenticated enriched worker feed.
`_normalize_proxy_workers` copies that name through with no validation, so a device that varies it
draws a fresh `_RIG_EDIT_CAP_PER_HOUR` budget per name over an unbounded name space. The cap
therefore bounds a NAME, not a device.

This is not new information: `data_service.py` already carries a `ponytail:` comment saying so and
deferring it to #235 as the broader unauthenticated-feed vector. #1566 restated it with the
observation that the prose overstated the guarantee — that half is already corrected, and
`SECURITY.md` and `docs/operations.md` both now say the bound is per name rather than per device.

## Why it is not a drive-by fix

The obvious direction — an overall ceiling beside the per-name one — has a real tension worth
deciding rather than assuming: a global budget means a rig that floods it suppresses OTHER
workers' genuine detections, converting a per-name row-rate problem into a fleet-wide
denial-of-signal one. The per-worker cap was chosen in #724 specifically so one rogue rig could
not crowd out genuine history.

Directions worth weighing, not a prescription:

- an overall ceiling beside the per-name cap, sized so it only engages under a name-rotation flood
- keying the cap on something the device does not choose, if anything on this feed qualifies
- bounding the number of DISTINCT names admitted per window rather than the rows per name

## What I checked, and what I did not

PROVEN HERE: the cap buckets on `worker` (read at source); the `ponytail:` comment deferring this
to #235 exists and predates #1566; the prose in both docs already scopes the bound to a name.
NOT DONE: no measurement of a name-rotating device against a live feed, and no judgement between
the three directions above — that is the decision this issue exists to get.

VijitSingh97:
Fixed by #1810, merged to `develop-v2` as `0e32da0e`.

Hand-closed: `Closes` is inert in this repo (PRs land on `develop-v2`, the default branch is `develop`), and #1695 carries no `do-not-close` label — enumerated at source before closing, not quoted from a list.

[...]

PR #1698, opened 2026-09-03
fix(dashboard): #1566 an audit row id no longer collides with another rig's

### LINKED ISSUE #1566: audit row ids join two rig-chosen strings unescaped, so a collision can drop a detection
The audit row id is built by joining two rig-chosen strings with an unescaped `-`, so distinct
(worker, change_id) pairs can mint the same id. Under `INSERT OR IGNORE` the second write is
dropped, which turns an id collision into a lost detection.

`#1561`'s fix bounds the id's LENGTH and CHARSET at the sink. It does not remove the ambiguity, and
it is not meant to — this is the `#724` id scheme, unchanged. Filing separately so it is not folded
into that fix.

## The two ids

In `dashboard/mining_dashboard/service/data_service.py`:

- normal row: `rig-edit-{worker}-{change_id}`
- cap marker: `rig-edit-ratelimited-{worker}-{int(window_start)}`

Both components are attacker-chosen. `worker` is `w.get("name", "")` off the normalized proxy feed;
`_normalize_proxy_workers` in `service/data_helpers.py` copies `w[_PX_NAME]` (and
`w.get("id", "Unknown")` on the legacy dict shape) straight through with no validation.
`_WORKER_NAME_RE` bounds names on the `config/worker_endpoints.py` path only — a different path that
this one never reaches.

## Shape 1 — a cap marker can be suppressed by another device

A device presenting as `ratelimited` and reporting `change_id` = `{victim}-{window}` mints
`rig-edit-ratelimited-{victim}-{window}`. Every character is in the safe set and the string is well
under the cap, so `clean_event_id` returns it VERBATIM and it lands as an ordinary rig-edit row. If
it lands before the victim trips the cap, the victim's own marker is dropped by `INSERT OR IGNORE`
and the flood loses its Security-panel row.

Reachability, stated honestly rather than assumed: it needs `int(window_start)` — the exact second
the victim's window opened — so it is a guess over ~3600 values, not a single shot. The
`logger.warning` fires before the marker is recorded either way, so the console signal survives;
only the panel row is suppressed. That makes this real but low-severity on its own.

## Shape 2 — the per-worker cap is keyed on a value the device chooses

`_rig_edit_within_cap` buckets on `worker`. A device that varies the name it presents gets a fresh
`_RIG_EDIT_CAP_PER_HOUR` budget per name, and the name space is unbounded. So the cap converts an
unbounded row rate into 12/hour PER NAME rather than into a bound on the device — which is also what
makes shape 1's ~3600 guesses affordable.

This is the part the prose overstates. `SECURITY.md`, `docs/operations.md` and `docs/dashboard.md`
all say in some form that no one device can grow the database without limit. That holds against a
device reusing one name and not against one that rotates names.

Make the id unambiguous rather than trying to constrain the components: length-prefix or
percent-encode each rig-chosen part before joining, or give the marker a namespace no
`rig-edit-{worker}-` string can enter. Either kills shape 1 outright. Shape 2 wants the cap keyed on
something the device does not pick, or an overall ceiling beside the per-name one.

## What I checked, and what I did not

PROVEN HERE: `_normalize_proxy_workers` applies no name validation (read at source); the verbatim
branch returns rig-chosen ids unchanged (`clean_event_id` run over three rig-chosen inputs, all
returned identical); `_rig_edit_within_cap` buckets on `worker`; `_RIG_EDIT_CAP_PER_HOUR` is 12.
NOT DONE: no end-to-end run against a live proxy feed, so the collision is derived from the code
rather than observed. Nobody should read this as an exploited finding.

VijitSingh97:
Closes #1566 (shape 1 only — see NOT DONE).

## The defect

The `audit_events` row id was a bare `-` join of two rig-chosen strings, so distinct detections
could mint the same key. Rows are written with `INSERT OR IGNORE`, so a collision is not an error
and not a duplicate row — the second detection is **DROPPED**, silently.

All three built ids were rig-chosen on both halves:

| id | both halves rig-chosen? |
|---|---|
| `rig-edit-{worker}-{change_id}` | yes |
| `rig-drift-{worker}-{revision}` | yes |
| `rig-edit-ratelimited-{worker}-{window}` | worker yes; forgeable as `ratelimited-{victim}` |

## The fix

`audit_service.build_event_id` percent-encodes each rig-chosen part so it carries no `:` and no
`-`, then joins on `:`. Two properties follow, and both are tested:

- **One-to-one.** The test proves it by DECODING an id back to the parts it was built from, not by
  listing colliding pairs — a finite list of pairs cannot establish injectivity and a future input
  could escape it.
- **Separate from every other writer's ids, by construction.** The `f"{source}-{uuid4()}"` fallback
  and the mirrored `control.log` ids carry no `:` and every built id does. The old scheme had this
  only as a reading of the call sites.

Escaped output stays inside `_SAFE_CHARS`, so ids still pass `clean_event_id` **verbatim** rather
than collapsing onto its digest branch — checked explicitly, because a fix that quietly sent every
unusual id to a digest would trade readability for correctness without saying so. `clean_event_id`
itself is unchanged and still bounds length at the sink.

## The over-engineering pass found a defect in my own fix

Building the id moves an **encode** into the poll loop, where the bare join never encoded anything.
A rig's JSON can carry `U+D800` and `quote`'s default strict UTF-8 encode raises on a lone
surrogate. Measured per component position, with each leg controlling the other:

```
surrogate in WORKER NAME -> reached the escape; raised in urllib/parse.py:909 without surrogatepass
surrogate in change_id   -> never arrives: worker_config_change_known hands it to sqlite first,
                            which refuses it (worker_config_store.py:242) — PRE-EXISTING, filed #1696
```

So `errors="surrogatepass"` is load-bearing on the worker-name half, and the claim is scoped to
that half in the docstring rather than stated generally. The existing surrogate test did not catch
this: it exercises `clean_event_id`, which sits **downstream** of the escape and never sees the
input that raises.

## Migration cost, stated

Deployed databases keep every id they already hold — `clean_event_id` returns those verbatim, and
the test pins that population separately from the ids built today. But a change still being
re-reported after the upgrade is recorded again under its new id, so expect **at most one extra row
per live (worker, change_id), once**. It is bounded by the same #724 cap as any other row.

## PROVEN BY ME, at the head pushed (`d3d635af`)

- Full dashboard suite **2457 passed / 0 failed**, rc 0, coverage **97.37%** (gate 80%).
- **Patch coverage 100% on 7 executable changed lines**, computed mechanically from a `coverage.xml`
  regenerated on this tree — not from `make test-patch-coverage`, whose script pins
  `COMPARE=origin/develop` and is no evidence on a `develop-v2` PR. **The non-vacuity assertion
  fired** (7 > 0); it caught a wrong path prefix on its first run, so the 100% is a reading and not
  an empty set. `data_service.py`'s one changed line reports 0 executable because it is a
  continuation line of a multi-line call; it is exercised by the wiring test, which drives the real
  `_reconcile_worker_config` and asserts the resulting ids.
- **13 tests added, 0 removed** (checked by grepping the diff for removed `def test_`).
- **Non-vacuity of the new tests: seeded the pre-#1566 bare join back in and 11 of them went RED**,
  the two intended controls staying green — the old-join witness and the INSERT-OR-IGNORE mechanism
  control, neither of which calls the builder. The two surrogate tests were proven by their own
  seeded control separately. Restored by sha256, verified by content.
- `lint-py` `lint-md` `lint-docs-voice` `lint-operator-strings` `lint-topology` `lint-file-budget`
  `lint-trivy-parity` `lint-yaml` `lint-toml` — each rc 0, run **individually** (that CI job runs
  six targets past the five in its name, and `make` stops at the first failure).
- Topology detector `clean` on this body and both filed issues.
- **NARROW case**: rebased, `git merge-base HEAD origin/develop-v2` == `git rev-parse
  origin/develop-v2` == `cf807597`. `develop-v2` moved under me mid-change when #1683 merged; every
  figure above is re-measured at the rebased head, not carried across.
- Budget: `data_service.py` **1420 / 1454** — the change there is exactly line-neutral. The other
  four touched files are `worker_change_audit.py` 112, `audit_service.py` 278,
  `test_audit_event_id.py` 286, `test_worker_change_audit.py` 330 — all under 400 and correctly
  rowless (the gate fails an under-target file that CARRIES a row as well as an over-target one
  lacking one). These four were re-measured after the surrogate tests landed; my first draft quoted
  the pre-surrogate figures.

## NOT DONE

- **Shape 2 of #1566 is not fixed** — the flood cap keyed on a worker name the device picks, so one
  rotating names draws a fresh budget per name. Filed as **#1695**. It is a different defect class
  with an existing home (`data_service.py` already defers it to #235), and a global ceiling has a
  real design tension: a flooding rig would then suppress *other* workers' genuine detections.
  **#1566 should stay open until #1695 lands, or be closed with a pointer to it** — reviewer's call.
- **#1696 is not fixed here** — the pre-existing sqlite surrogate raise on the `change_id` half.
- Not run on hardware; no live proxy feed. The collision is proved against the real writer and the
  real table in-memory, not against a physical rig.
- `lint-sh` not run: no shell in this diff, and it is the serialized OOM path.
- The `ponytail` PR gate verifies nothing; the over-engineering pass above was done on merits and is
  what found the surrogate defect. It is not reported as a review.

## Shared files touched (disclosed per `_shared.paths`)

- `SECURITY.md` — one sentence became **false** with this change: "the ids are a plain concatenation
  and do not separate one worker from another". Corrected, not deleted. No wholesale rewrite.
- `docs/operations.md` — the operator-facing half gains the guarantee it did not previously have.
- Both already said the cap is per NAME rather than per device; that prose is correct and untouched.
- No `CHANGELOG.md` entry: this repo touches it only at release prep, as its sibling #1561 did not.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -128,12 +128,16 @@ The stack's defaults:
   row's own identifier is bounded separately (#1561): every audit row id
   is length-capped and whitelisted at the writer, the one field the trail's sanitizer used to skip,
   and `rig-drift`'s revision is validated to a short opaque token at the point it is read as well.
-  A rig does still CHOOSE its own `rig-edit` row's identifier — that id is built from a `change_id`
-  it picks freely — but it can no longer make one arbitrarily large or carry markup into it, which
-  is what matters on a table that is never pruned. What it cannot do is forge a `host-edit` or
-  `control.log` identifier: those are built from hardcoded prefixes a `rig-edit` id can never equal.
-  Inside the `rig-edit-` prefix, though, the ids are a plain concatenation and do not separate one
-  worker from another (#1566), so read that namespace as one shared space rather than as per-worker.
+  A rig does still CHOOSE the contents of its own `rig-edit` row's identifier — that id is built
+  from a `change_id` it picks freely — but it can no longer make one arbitrarily large or carry
+  markup into it, which is what matters on a table that is never pruned. Nor can it steer that id
+  onto a row that is not its own: each rig-chosen part is escaped before the parts are joined
+  (#1566), so an id is a one-to-one function of the worker name and the change id it was built
+  from. That is the property this table needs rather than a tidier one, because rows are written
+  with `INSERT OR IGNORE`: two detections that minted the same id produced no error and no second
+  row, and the later detection was DROPPED. The same escaping settles the `host-edit` and
+  `control.log` identifiers by construction instead of by reading the call sites — those carry no
+  `:` and every out-of-band id does, so a rig cannot mint one however it names itself.
   The `host-edit` and mirrored `control.log` rows are not attacker-controllable and are not capped.
 
 ### Telegram control commands (#338)
--- a/dashboard/mining_dashboard/service/audit_service.py
+++ b/dashboard/mining_dashboard/service/audit_service.py
@@ -20,6 +20,7 @@ import json
 import os
 import re
 import time
+from urllib.parse import quote
 
 from mining_dashboard.config import config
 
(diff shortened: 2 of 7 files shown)


VijitSingh97:

**Re-derived by me, against the branch's own `audit_service` imported into a throwaway worktree (dashboard venv):**
- The docstring's adversarial pair (`victim-chg1`/`extra` vs `victim`/`chg1-extra`) collides under the old bare join and does NOT under `build_event_id`.
- Every built id survives `clean_event_id` verbatim (three shapes incl. `:`, `-`, `~`, `%` in the parts), so nothing quietly moved onto the digest branch.
- One-to-one: 2,000 random `(worker, change_id)` pairs over an alphabet including `-`, `:`, `~`, `%`, space and a non-ASCII char all split on `:` and unquote back to exactly the inputs.
- A lone surrogate in the worker name escapes to `%ED%A0%80` and does not raise.
- Source diff read in full: three call sites moved to the builder, `-`/`~` replaced after quoting (unambiguous since `%` is already `%25`), the namespace is a repo literal and the only unescaped field. `SECURITY.md` and `docs/operations.md` say the new property correctly. `data_service.py` is line-neutral (1420/1454 in the body; the gate confirms).
- Ruling 8a, both legs vs the current tip (merge-base `cf807597`): budget rc 0 with 5 self-tests fired, parity rc 0, at head and at the merged tree; `merge-tree` rc 0.

**Relayed, not re-run:** the full dashboard suite (2457/0, 97.37%), the 7-line patch coverage, and the 11-of-13 seeded non-vacuity result. CI's dashboard job at this head is the independent pass leg.

**Close ruling:** #1566's title and shape 1 are this fix; shape 2 (the per-name cap) is a different class with its own issue (#1695) and its own design tension, so #1566 is closed by hand with a pointer rather than held open on a second defect. #1696 stays open.

Issue #1470, opened 2026-08-28
file-budget: a new tsv row can reserve headroom on first appearance


VijitSingh97:
A budget row that is new in a PR can be set to any ceiling at or above the file's real line count, and nothing mechanical objects to the gap afterwards. The ratchet then holds a number that was never an honest measurement.

## What happens

`check_monotonic` iterates the rows of the **base** tsv:

    old_lines=$(git show "$base:$BUDGET_FILE" | parse_budget)
    ...
    while IFS=$'\t' read -r path old_ceiling; do
    ...
    done <<<"$old_lines"

A path that has no row on the base ref is never visited, so the raise check cannot apply to it. There is no first-appearance branch.

`run_gate` fails a row only when the file is over its ceiling:

    if [ "$lines" -gt "$ceiling" ]; then

A new row of 900 for a 500-line file satisfies that, because 500 is not greater than 900.

The two checks together leave first appearance unguarded. Once the row exists, every later PR is measured against 900 rather than against 500, and the 400 lines of unearned headroom are invisible from then on: no later run compares the ceiling to the file again, and lowering a ceiling is always permitted but never required.

## Why it is worth fixing

The gate's value is that a ceiling is a recorded measurement. The recent fix for the exempt row (#1464) closed exactly this shape one row over — it made a *rise* state the file's real count instead of reserving headroom:

    actual=$(count_lines "$path")
    if [ "$new_ceiling" != "$actual" ]; then
        # FAIL - a rise must record the artifact, not grant it headroom

That reasoning applies unchanged to a row's first appearance. A ceiling set above the file is headroom whether it arrives by raising an old row or by writing a new one; only the first route is currently checked.

This is not hypothetical for the current work. Phase 2 of the split adds new files, and each one arrives with a new row.

## Suggested remedy

Bound a new row to the file's real count, reusing the check #1464 added. In `check_monotonic`, walk the rows of the **new** tsv that are absent from the base, and fail when the ceiling is not equal to `count_lines "$path"`.

Two details worth keeping:

- Use `count_lines`, the gate's own counter, not `wc -l`. A file with no trailing newline is undercounted by `wc -l`, which would turn the check into a false red.
- Fail closed when the file cannot be read. `count_lines` already falls back to 0, which mismatches any ceiling, so a deleted or unreadable path fails rather than passes. That behaviour was confirmed by running the gate against a deleted file and against one at mode 000.

[...]

PR #1699, opened 2026-09-03
fix(file-budget): gate refuses a new row that reserves headroom on first appearance

VijitSingh97:
The file-budget ratchet gate's monotonic check only ever walked rows present on the base ref. A brand-new row — a path with no counterpart on `develop-v2` — never entered that loop, so a PR could add a new row to `docs/dev/file-budget.tsv` that reserved headroom above the file's real line count on its very first appearance, and nothing mechanical would object again until the file actually grew into that reserved ceiling. That is the same "a rise must record, not reserve" gap #1464 closed for a later raise on an existing row, left open for the row's first line.

`check_monotonic()` in `scripts/lint-file-budget.sh` now also walks every row in the working-tree budget that has no counterpart on the base ref, and fails unless its ceiling equals the file's actual line count (via the gate's own `count_lines`, not `wc -l`, so a file with no trailing newline isn't undercounted; a deleted or unreadable path falls back to 0 and so mismatches any ceiling and fails closed).

Proven with:
- `bash scripts/lint-file-budget.sh --self-test` — new fixture: a first-appearance row at ceiling 900 on a 500-line file FAILS and names the reserved headroom; a first-appearance row stating the real count (500) passes as a control. All prior fixtures still pass.
- `bash scripts/lint-file-budget.sh` — clean pass against the current repo state.
- `shellcheck` and `shfmt -d` on both touched files — clean.

`scripts/lint-file-budget-selftest.sh` adds the fixture: it asserts the FAIL case names the reserved headroom (`"adds newfile.sh as a new row at ceiling 900, but the file is 500 lines"`), and that the control case (real count on first appearance) passes with rc=0.

Refs #1470

### DIFF
--- a/scripts/lint-file-budget-selftest.sh
+++ b/scripts/lint-file-budget-selftest.sh
@@ -236,6 +236,37 @@ self_test() {
     fi
     rm -rf "$tmp4"
 
+    # #1470 — a row with NO counterpart on the base ref is a first appearance, not a raise, so
+    # the monotonic loop above never visits it. Its own repo, so the 500/900 base state above
+    # cannot leak into it.
+    local tmp5
+    tmp5=$(mktemp -d)
+    git -C "$tmp5" init -q -b develop
+    git -C "$tmp5" config user.email test@example.invalid
+    git -C "$tmp5" config user.name test
+    mkdir -p "$tmp5/$(dirname "$BUDGET_FILE")"
+    seq 1 500 >"$tmp5/budgeted.sh"
+    printf '# test budget\nbudgeted.sh\t500\n' >"$tmp5/$BUDGET_FILE"
+    git -C "$tmp5" add -A && git -C "$tmp5" commit -q -m base
+
+    seq 1 500 >"$tmp5/newfile.sh"
+    printf '# test budget\nbudgeted.sh\t500\nnewfile.sh\t900\n' >"$tmp5/$BUDGET_FILE"
+    rc=0
+    (cd "$tmp5" && run_gate) >"$out" 2>&1 || rc=$?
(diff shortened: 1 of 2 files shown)


VijitSingh97:
## Controller pass — PASS (non-author; judged on the fable tier) — head `58fcbae6`

My earlier legs were for `c3f1982d`; the head moved by one commit (the two `check_monotonic` passes merged into one walk over the working-tree rows), so everything below is re-run at `58fcbae6`.

**Re-derived by me:**
- Diff read in full, against merge-base `cf807597` and the refactor commit alone. The single pass is behaviour-preserving: a working-tree row with no base counterpart takes the first-appearance arm (ceiling must equal `count_lines`, failing closed on an unreadable path); a row with one takes the raise arm unchanged (an exempt row must state the real count, any other rise fails). A base row absent from the working tree is never visited — the same outcome the old first loop reached by skipping an empty `new_ceiling`. When no base ref resolves the function still returns before the loop, so CI's skip (L3) is unchanged.
- The #1470 fixture fired in my own run at the head and at the merged tree (`a new row reserving headroom above the file's real size FAILS (#1470)`, `names the reserved headroom`, and the passing control) — three new self-test lines, 29 in all.
- Ruling 8a, both legs vs the tip `2347df02` (merge-base `cf807597`): budget rc 0, parity rc 0, at the head and at the merged tree; `merge-tree` rc 0.

**Merge-record note:** #1701 and #1681 also edit `scripts/lint-file-budget.sh`. #1681 touches a different region (`list_candidates`) and its merged-tree leg is clean. #1701's comment hunk anchors to the `-n "$new_ceiling"` skip this PR removes and does not apply on this head (`git apply --check`); its self-test case does apply and passes here. Findings posted on each.

Base is `develop-v2`; #1470 is closed by hand on the fired fixture.

Issue #1106, opened 2026-08-18
Repo layout: the second monolith cluster, the dashboard's hidden location, and a root nobody can scan


VijitSingh97:
Companion to #1105, which covers `pithead` and `tests/stack/run.sh`. This one is everything else: what other files have outgrown themselves, where the layout misleads a reader, and what of that is worth doing rather than merely tidy.

## Every hand-written file over 600 lines

Generated and vendored files excluded — `*_pb2_grpc.py`, `*.proto`, `uv.lock`, benchmark `.jsonl`, `CHANGELOG.md` — because including them makes the list look worse than it is and none of them are anyone's problem.

| Lines | File | |
|---:|---|---|
| 11,027 | `tests/stack/run.sh` | #1105 |
| 10,372 | `pithead` | #1105 |
| 3,387 | `build/dashboard/tests/web/test_views.py` | |
| 3,130 | `build/dashboard/tests/service/test_data_service.py` | |
| 2,644 | `tests/os/run.sh` | |
| 2,534 | `tests/integration/run.sh` | |
| 2,250 | `build/dashboard/mining_dashboard/web/views.py` | |
| 1,852 | `build/dashboard/mining_dashboard/service/data_service.py` | |
| 1,646 | `build/dashboard/tests/service/test_storage_service.py` | |
| 1,603 | `build/dashboard/mining_dashboard/service/storage_service.py` | |
| 1,541 | `build/dashboard/mining_dashboard/web/static/dashboard.css` | |
| 1,369 | `build/dashboard/mining_dashboard/web/static/components.mjs` | |

**There is a second monolith cluster and it is the dashboard**: `views.py`, `data_service.py` and `storage_service.py` total 5,705 lines, each shadowed by a test file of comparable size. Same shape as the CLI, without any of the shipping constraints that make #1105 delicate — these are ordinary Python modules in a package, and splitting them is the boring kind of refactor.

`tests/os/run.sh` and `tests/integration/run.sh` are phase-structured (`phase_install`, `phase_update`, …) and split along those seams almost for free.

## The layout problem that costs more than any file size

[...]

PR #1700, opened 2026-09-03
build(deps): #1685 F1 the develop-v2 twin of dependabot.yml, inert on this branch

VijitSingh97:
## What this is

**#1685's F1, discharged.** #1685 gave `develop-v2` its own Dependabot coverage and landed on `develop` under STANDING-DIRECTIVE exception (b) — which obliges a sync forward once firing is proven. Firing is proven; this is the forward half.

## The copy is inert on this branch, and that is the point

Dependabot reads this file from the **default branch only**, which is `develop`. Nothing in this file — the pre-existing entries or the block added here — does anything while `develop-v2` is not the default branch. The copy exists so that:

- the twins stay level, so a hand-made change to one shows up as missing from the other;
- the #1169 retirement inherits a config that already works, instead of one branch's worth of drift discovered at cutover.

A header comment in the file says this, including that the block's prose describes the config as Dependabot resolves it *from `develop`* — so its `/build/dashboard` references are about that branch's tree, not this one's.

## Firing proof — re-derived here, not relayed

`gh pr list --base develop-v2 --author app/dependabot` at the time of writing:

| PR | opened | what |
|---|---|---|
| #1689 | 2026-09-03T15:33:27Z | `actions/download-artifact` 7.0.0 → 8.0.1, `actions-v2` group |
| #1691 | 2026-09-03T15:33:58Z | `python` digest in `/dashboard`, `docker-v2` group |
| #1692 | 2026-09-03T15:35:19Z | `python-v2` group in `/dashboard`, 3 updates |

Three PRs against `develop-v2` within three minutes of #1685's 15:32:17Z merge. F1's stated prediction was that `download-artifact` v7 → v8 and the ruff bump would appear; #1689 is the first of those. The channel fires.

## Verification: "level" measured structurally, not by eyeball

Both files parsed and compared entry by entry, rather than diffed as text — comment reflow would have swamped a text diff:

- **9 entries on each side, 5 carrying `target-branch`** — same shape.
- **The develop-v2 coverage half (entries 5–8) is byte-identical to develop's.** Asserted programmatically, not read.
- **Exactly two entries differ, and only in one field:** entries 1 (`uv`) and 2 (`docker`) say `/dashboard` here and `/build/dashboard` on develop.

That last difference is **pre-existing and deliberate** — F1 names it explicitly ("`develop-v2`'s own copy of this file already differs from develop's; its `uv` entry names `/dashboard`"). Those are each branch's real paths: `build/dashboard` EXISTS on `develop` and 404s here, while `dashboard` is exactly mirrored. Measured at both tips. Making the files identical would break one branch or the other, so **"level" here means the hand-made change is paired, not that the files match byte for byte.**

## Gates run

`lint-yaml` PASS · `lint-file-budget` PASS · the YAML parses to 9 well-formed update entries. `.github/dependabot.yml` has no budget row and is 213 lines, under the 400-line target.

## Lane and grant disclosure

`.github/dependabot.yml` is currency's under the w89 grant block, and the lane-guard ALLOWED it staged alone. **Flagged rather than quietly relied on:** that block's own text says it expires "when the named fix PR merges, whichever is first", and the named PR (#1685) merged at 15:32:17Z — so by its own wording the grant reads *expired*, while the line is still present in `roles/currency.paths`. I re-armed nothing; this work was assigned by the controller in this cycle, naming this file. The grant text wants re-dating or retiring, and I have said so to the controller rather than leaving a live grant that reads as expired.

## Not done

- No claim that the *contents* of the develop-v2 coverage entries are optimal — this is a verbatim sync of an already-reviewed, already-merged change, deliberately not a redesign.
- `actions/download-artifact` v7 → v8 compatibility for develop-v2's `ci.yml` is untouched and belongs to #1689, the PR that proposes it.
- Nothing here can be exercised live: the config is inert on this branch by construction, so there is no run to point at. The firing evidence above is necessarily from `develop`.

---

## Over-engineering pass (ponytail gate, by hand)

The diff is +114 lines in one file, of which 105 are a verbatim copy and 8 are new comment. Four things put to it:

1. **Verbatim copy vs. rewriting the block's prose for this branch.** Verbatim, deliberately. Rewriting would produce two diverging explanations of one mechanism and would destroy the property that makes "level" checkable at all — that entries 5–8 parse byte-identical to develop's. The one real cost of copying (prose that says `/build/dashboard` on a branch where that path 404s) is paid off by one sentence in the header rather than by a rewrite.
2. **Eight comment lines for a copied block — too many?** Kept. They answer four questions a reader on this branch will otherwise ask and get wrong: why the file exists here, that it is inert, that retirement inherits it, and whose tree the block's prose describes. The file it joins is already ~90% comment, so this matches the surrounding density rather than exceeding it.
3. **Could this have ridden in #1685?** No. #1685 had to land on `develop` because that is the only branch Dependabot reads and the only place firing could be proven. This has to land on `develop-v2`. Different base is a hard constraint, not a preference.
4. **Is syncing inert entries worth doing at all?** That is the question F1 already answered and #1169 measures the cost of: unsynced twins are discovered at cutover, one broken watcher at a time. The cheap moment is now.

No simplification found that did not cost more than it saved. The only original content in this diff is the header comment; everything else is a sync of already-reviewed, already-merged content.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -97,3 +97,117 @@ updates:
       # major/minor is a deliberate migration (compose flags, config compatibility), not a CVE fix.
       - dependency-name: "*"
         update-types: ["version-update:semver-major", "version-update:semver-minor"]
+
+  # ── INERT ON THIS BRANCH ─────────────────────────────────────────────────────────────────────
+  # Everything below is a verbatim copy of what `develop` carries, kept so the twins stay level
+  # (#1685 F1). Dependabot reads this file from the DEFAULT branch only, which is `develop`, so
+  # nothing in this file — above the line or below it — does anything while `develop-v2` is not
+  # the default branch. The copy exists so the retirement (#1169) inherits a config that already
+  # works, and so a hand-made change to one twin shows up as missing from the other. The prose
+  # below describes the config as Dependabot resolves it FROM `develop`, so its `/build/dashboard`
+  # references are about that branch's tree, not this one's.
+
+  # ── develop-v2 coverage (#1425, #1650) ──────────────────────────────────────────────────────
+  # Everything above this line resolves against the DEFAULT branch, `develop`. That is not a
+  # style choice: Dependabot reads this file from the default branch only, and with no
+  # `target-branch:` it also resolves each `directory:` there and opens the PR there. So every
+  # entry above watches the FROZEN twin, and `develop-v2` — the branch all work actually lands
+  # on — was covered by the `/os/rootfs` entry and nothing else.
+  #
+  # Two distinct faults produced that, and they need different repairs (#1425):
+  #   * a WRONG DIRECTORY — `uv` and the docker group name `/build/dashboard`, which is 404 on
+  #     develop-v2 (the dashboard moved to `dashboard/` at #1106). Repair: name the real path.
+  #   * a MISSING TARGET — `github-actions` and `docker-compose` name `/`, which exists on both
+  #     twins. Nothing is mispointed; the entries simply never target develop-v2. Repair: a
+  #     second entry carrying `target-branch:`.
+  # A fix that treats these as one edit repairs the first and silently leaves the second.
+  #
+  # DO NOT "level the twins" by deleting these. `develop` is frozen but still takes critical
+  # patches under lane policy exception (a), so its entries stay; and moving an entry to
+  # develop-v2 is what makes it stop running, which is how the gap went unnoticed for so long
+  # (CONTRIBUTING.md § The two lanes; the /os/rootfs comment above states the same mechanism).
+  # The duplication ends when develop-side watching is retired — that is #1169, and it is
+  # blocked while develop still ships patches. Until then, two entries per surface is the cost.
+  #
+  # WHY EVERY SURFACE IS LISTED HERE AND NOT JUST THE DASHBOARD. The twins are held level by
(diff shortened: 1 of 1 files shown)


VijitSingh97:

**Re-derived by me:**
- Diff read in full: one file, +114 lines, all additive. The four develop-v2 entries are a byte-identical copy of develop's (compared against develop's copy of the file at source); the only new prose is the eight-line "inert on this branch" header, which says the right thing — Dependabot reads the default branch only, so this copy does nothing until the #1169 retirement, and its `/build/dashboard` references describe develop's tree.
- The pre-existing `/dashboard` vs `/build/dashboard` split in the first two entries is each branch's real path (the dashboard moved at #1106), so "level" correctly means paired hand-made changes, not byte identity.
- Firing proof re-read at source: dependabot PRs #1689, #1691, #1692 opened against `develop-v2` within three minutes of #1685's merge.
- Ruling 8a, both legs, re-run after the tip moved to `d1c32cb6` (merge-base `cf807597`): budget rc 0 with 5 self-tests fired, parity rc 0, at head and at the merged tree; `merge-tree` rc 0.

**Grant note, acted on rather than relayed:** the w89 `dependabot.yml` grant in `currency.paths` expired by its own wording at #1685's merge; this branch was pushed under it before that and needs no grant to merge. The line is retired as EXPIRED with this merge.

Nothing to close: F1 was a review finding, not an issue.

Issue #1430, opened 2026-08-28
lint-file-budget.sh:216 still carries #1375's construct — it is safe, but only because of two checks 45 lines upstream, and nothing says so


VijitSingh97:
## What is wrong

`scripts/lint-file-budget.sh:216` reads:

    if [ -n "$new_ceiling" ] && [ "$new_ceiling" -gt "$old_ceiling" ]; then

This is the exact construct #1375 was filed about — a budget entry that has vanished from the new
tsv is *skipped* rather than checked. #1375 is closed and **the defect it describes is genuinely
fixed**, but it was fixed somewhere else, and nothing at or near line 216 says so.

**The line is safe. It does not look safe.** That gap has now cost two separate readers a
re-derivation, which is the reason to file rather than leave it.

## Why it is actually safe — the argument, in full, because it is not local

Anything reaching `check_monotonic` has already passed two earlier gates, both inside `run_gate`
and both **before** the `check_monotonic` call at `:189`:

1. `:169-174` — a file over `TARGET_LINES` with **no** tsv entry is already a FAIL
   (*"…but has no … entry"*).
2. `:180-188` — a tsv entry naming a file that is missing **or now exempt** is already a FAIL
   (*"…which no longer exists (or is now exempt). Remove the entry."*).

So by the time line 216 runs, an entry present in the base and absent from the new tsv can only
mean the file shrank to `<= TARGET_LINES` — the sanctioned-removal case, which *should* be skipped.
The skip is correct.

The premise that makes this hold is that the candidate set is the **whole tree**, not the diff:
`:146` `candidates=$(list_candidates)` is `git ls-files` minus exemptions and binaries, and `:147-151`
refuses to pass at all if that scan returns zero. If candidates were ever narrowed to changed files,
**the argument above would break and #1375's defect would return** — silently, because line 216 would
not change.

## Proposed fix — a comment, nothing else

No behaviour change. Add a short comment at line 216 recording why the `-n` skip is correct and what
it depends on. Something of this shape:

    # Safe only because run_gate has already failed (a) any over-target file with no entry
    # (:169) and (b) any entry whose file is gone or exempt (:180). So a base entry missing
    # here means the file shrank to <= TARGET_LINES — the sanctioned removal. If the candidate
    # set at :146 is ever narrowed from the whole tree to the diff, this skip fails open again
    # and #1375 returns.

## Evidence

[...]

PR #1701, opened 2026-09-03
docs+test(lint-file-budget): prove the #1375-shaped skip at check_monotonic is safe

VijitSingh97:
## The gap

`check_monotonic` in `scripts/lint-file-budget.sh` skips its raise check with `[ -n "$new_ceiling" ]` whenever a base `docs/dev/file-budget.tsv` entry has no matching row in the new tsv. That is exactly #1375's construct — an entry vanishing between the base and the working tree — and read on its own the skip looks like it could fail open: nothing there explained why an absent new-side entry is safe to let through unchecked. The self-test suite had no case at all for "base entry present, new entry legitimately gone" — only the neighboring "still present but stale" case was covered, so a regression that made this skip fail open again would have gone uncaught.

## What changed

- `scripts/lint-file-budget.sh`: a comment at the skip explains why it's safe — by the time `check_monotonic` reaches this line, `run_gate`'s earlier arms have already FAILed the other two ways a path could go missing (an over-target file with no entry, or an entry naming a file that's gone/now exempt), so an absent new-side entry here can only mean the file legitimately shrank to ≤ `TARGET_LINES`. The comment also flags the argument's load-bearing assumption: it only holds because `run_gate`'s candidate set is the whole tree (`list_candidates`), never a diff.
- `scripts/lint-file-budget-selftest.sh`: adds the missing case — a budgeted file shrinks below target, its tsv row is correctly dropped, and `run_gate` must PASS. This is the direct complement of the existing "dropped below target but entry left stale" FAIL case, and it's the case that would catch a regression reintroducing #1375's construct.

## How it was proven

```
shellcheck scripts/lint-file-budget.sh scripts/lint-file-budget-selftest.sh   # rc=0
shfmt -d scripts/lint-file-budget.sh scripts/lint-file-budget-selftest.sh     # rc=0
make lint-file-budget
```

`make lint-file-budget` runs `bash scripts/lint-file-budget.sh --self-test` (all cases pass, including the new "a shrunk file whose entry was correctly removed PASSES (legitimate removal, #1430)" case) followed by the real gate against this repo, which reports OK.

Given a budgeted file that shrinks to ≤ `TARGET_LINES` and its now-unnecessary `docs/dev/file-budget.tsv` row is removed in the same change, `run_gate` exits 0 — proving the `-n "$new_ceiling"` skip in `check_monotonic` passes the legitimate-removal case it exists for, not just any absent entry.

Refs #1430

## Shared-file disclosure (moved into the body, where COMMON says it belongs)

`scripts/lint-file-budget-selftest.sh` is in **no role's `.paths` and not in `_shared.paths`**. Proved rather than assumed: staged alone, the lane-guard refuses it, rc 1, naming that exact path. `scripts/lint-file-budget.sh` is a live per-item grant to this lane.

**`git rebase --continue` does not run the `pre-commit` hook**, so the lane-guard never fired on this commit at all. The one-shot `.lane-override` I armed came through **unconsumed**; I removed it rather than leave it armed and invisible in `git status`. That is a mechanism gap worth having on the record: any lane can land an out-of-lane file via a rebase with no refusal and no trace.

### DIFF
--- a/scripts/lint-file-budget-selftest.sh
+++ b/scripts/lint-file-budget-selftest.sh
@@ -123,6 +123,19 @@ self_test() {
     expect "a budgeted file dropped below target still listed FAILS (stale entry)" 1 "$rc"
     seq 1 500 >"$tmp/budgeted.sh"
 
+    # Complement of the case above, and the one #1430 found untested: the base entry is not just
+    # stale, it is GONE from the new tsv. check_monotonic walks the working-tree rows, so a base
+    # row absent here is never visited — the file legitimately shrank to <= target — and the gate
+    # must PASS, not FAIL. Without this case, a regression that made that skip fail open again
+    # (#1375's construct) would go uncaught by this suite.
+    seq 1 300 >"$tmp/budgeted.sh"
+    printf '# test budget\n' >"$tmp/$BUDGET_FILE"
+    rc=0
+    (cd "$tmp" && run_gate) >"$out" 2>&1 || rc=$?
+    expect "a shrunk file whose entry was correctly removed PASSES (legitimate removal, #1430)" 0 "$rc"
+    seq 1 500 >"$tmp/budgeted.sh"
+    printf '# test budget\nbudgeted.sh\t500\n' >"$tmp/$BUDGET_FILE"
+
     # Two-lane shape: a develop-v2 branch cut BEFORE the lane tip advanced must still
     # ratchet against develop-v2, not fall back to develop and read this lane's larger
     # ceiling as a raise — the false RED that bit two live PRs the day the tip moved.
--- a/scripts/lint-file-budget.sh
+++ b/scripts/lint-file-budget.sh
@@ -259,6 +259,15 @@ check_monotonic() {
     # count_lines is the gate's own counter (not wc -l, which undercounts a file with no
     # trailing newline) and falls back to 0 when the path can't be read, so a deleted or
     # unreadable path mismatches any ceiling and fails closed rather than passing.
+    # The complement of that case is #1375's construct, and it is unchecked BY CONSTRUCTION: a
+    # path with a row on $base and none here is never visited by this loop at all. It stays safe
+    # only because run_gate has already FAILed both of the other ways a path could go missing --
+    # an over-target file with no row (its own arm), and a row naming a file that is gone or now
+    # exempt (the loop above this one). So a base row absent here can only mean the file shrank
+    # to <= TARGET_LINES: the sanctioned removal. That argument leans on run_gate's candidate set
+    # being the WHOLE tree (list_candidates), never a diff -- narrow it to changed files and this
+    # skip fails open again, silently, because nothing here would change. #1430; the self-test
+    # covers the legitimate-removal PASS beside the still-listed-but-shrunk FAIL.
     while IFS=$'\t' read -r path new_ceiling; do
         [ -n "$path" ] || continue
         old_ceiling=$(printf '%s\n' "$old_lines" | awk -F'\t' -v p="$path" '$1==p {print $2; exit}')


VijitSingh97:
## Controller pass — FINDINGS (non-author; judged on the fable tier). Rebase after #1699: keep the test, drop the comment hunk.

#1699, merging ahead of this, rewrites `check_monotonic` as one pass over the WORKING-TREE rows. The `-n "$new_ceiling"` skip this PR's comment explains no longer exists: a base row absent from the working tree is simply never visited. Measured on #1699's head `58fcbae6`: this PR's `lint-file-budget.sh` hunk fails `git apply --check` (patch failed at :244); its `lint-file-budget-selftest.sh` hunk applies cleanly and the new case passes there (`a shrunk file whose entry was correctly removed PASSES (legitimate removal, #1430)`, full self-test rc 0).

**What to do:** rebase onto the tip once #1699 is in; keep the self-test case as is — it is the regression guard #1430 asked for and it is still meaningful under the new shape; drop the comment hunk. If you want the argument preserved (a good one: a vanished base row is safe only because `run_gate`'s earlier arms already refuse the two other ways a path goes missing, which leans on `list_candidates` being the whole tree), compress it to two or three lines in the loop's header comment in #1699's shape — do not restate the old loop.

Merge condition: rebased head, `merge-tree` rc 0 vs the tip, checks none non-terminal and none failed at the moment of merge. #1430 closes by hand.

VijitSingh97:

Re-derived at source in a detached worktree at PR head `6085380b`: read `check_monotonic` (`scripts/lint-file-budget.sh:225-264`) together with `run_gate`'s two earlier arms (`:167-176` over-target-with-no-entry, `:182-189` stale/gone entry) that run over the whole-tree candidate set before `check_monotonic` is reached (`:191`). The safety argument in the new comment holds: by the time `check_monotonic` sees an absent `$new_ceiling`, the only way to reach that line with the file still over target has already failed the gate earlier, so an absent entry here can only be a legitimate shrink.

Named checks, from the repo root:
```
$ shfmt -d scripts/lint-file-budget.sh scripts/lint-file-budget-selftest.sh
(no output, rc=0)

$ make lint-file-budget
  self-test ok: a shrunk file whose entry was correctly removed PASSES (legitimate removal, #1430)
  ... (28/28 cases)
lint-file-budget self-test OK
```

Standing test — negative control on the new self-test case, not the whole file: my first mutation (treat missing `$new_ceiling` as `0`) didn't fire — `0 -gt $old_ceiling` is never true, so that was a broken control, not a clean one. Corrected: made the skip explicitly FAIL on a missing entry (removes the sanctioned exemption). Only the new case reddened, all 27 others stayed green:
```
self-test FAIL: a shrunk file whose entry was correctly removed PASSES (legitimate removal, #1430) (expected rc 0, got 1)
lint-file-budget self-test FAILED
```
Restored the file afterward (worktree diff clean). This is a real, precisely-aimed test, not vacuous.

Two non-blocking discrepancies against the PR body, neither caused by this diff:
- `shellcheck scripts/lint-file-budget.sh scripts/lint-file-budget-selftest.sh` returns **rc=1** here (shellcheck 0.9.0), not the body's claimed rc=0 — `SC2015` (info) at `lint-file-budget.sh:145`, on a line this PR does not touch. Confirmed pre-existing: `shellcheck` on the unmodified `origin/develop-v2` copies of both files hits the identical `SC2015`, plus source-resolution info notes when checked outside the tree. Body's claim doesn't reproduce on this box; worth a second look at whichever shellcheck version produced the clean run, but it predates #1430.
- `bash scripts/lint-file-budget.sh` alone FAILS: `docs/dev/file-budget.tsv raises tests/os/run.sh's ceiling from 3423 to 3437` — same base-drift already reported on #1681 (`origin/develop-v2` lowered that ceiling at `ebb0f43d` after this branch cut; this PR never touches the tsv). Resolved by rebase, not a defect in this diff.

"Could this gate be green if a consumer ignored its exit code" — no: `Makefile:120-121` runs both invocations un-piped under `make`, and `check_monotonic`'s output names the specific path and ceilings, not just a bare rc.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `6085380b`

VijitSingh97:
**Correction to my non-author pass above (comment 5531489792).**

The `shellcheck` discrepancy I flagged as non-blocking ("rc=1, SC2015") was my own tooling error, not a finding about this PR. I called the plain `shellcheck` on `$PATH`, which on this box resolves to apt's unpinned **0.9.0** — not the repo's CI pin. `Makefile:52` pins `SHELLCHECK_VERSION := 0.11.0`, and the correct pinned binary lives at `~/.local/bin/shellcheck-real` (its serializing wrapper was queued behind other work at the time, so I'd skipped it — should have gone to the unwrapped pinned binary instead, which is safe for a small, targeted file pair).

Re-run at the correct pin:
```
$ ~/.local/bin/shellcheck-real --severity=warning scripts/lint-file-budget.sh scripts/lint-file-budget-selftest.sh
(no output, rc=0)
```
Clean — the PR body's original claim (`shellcheck ... # rc=0`) was correct all along. Withdrawing that non-blocking note; nothing else in the PASS verdict changes.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `6085380b`

VijitSingh97:
## Rebased onto `a3b723bd` — and the conflict resolution is NOT mechanical, so re-read it

Force-pushed `6085380b` -> `ece54109`. The previous non-author PASS was at `6085380b` and **does not cover this head.**

### What changed under this PR, and why the diff is not what it was

#1470 landed on the base and **inverted `check_monotonic`'s loop.** It used to walk the BASE rows and look the new ceiling up behind a `-n "$new_ceiling"` guard; it now walks the WORKING-TREE rows and looks the base ceiling up. **The guard this PR's comment was attached to no longer exists**, which is what the conflict was.

The construct itself is unchanged in substance — a row present on the base and absent from the new tsv is still never checked, it is just now skipped by the loop's iteration set rather than by an explicit guard — and the safety argument is unchanged. So the resolution takes the base side of the code and re-attaches the argument to the loop. The self-test case is unchanged and still correct against the new shape.

### The added case earns its keep — control, not assertion

A test that asserts a PASS is worth nothing until something can make it fail. Reconstructing the fail-open regression it names (walk the base rows, drop the `-n` guard) against this tree turns it RED: `self-test FAIL: a shrunk file whose entry was correctly removed PASSES (legitimate removal, #1430) (expected rc 0, got 1)`. Run in a scratch copy, not this checkout.

**Stated limit:** that mutation is not surgical — it reds several sibling cases too, so it does not show the case is *uniquely* sensitive to this regression. What it does show is that the case can say something other than PASS.

### Out-of-lane file, disclosed — and the guard never got to say so

`scripts/lint-file-budget-selftest.sh` is in **no role's `.paths` and not in `_shared.paths`**. I proved that rather than assumed it: staged alone, the lane-guard refuses it, rc 1, naming that exact path. `scripts/lint-file-budget.sh` is a live per-item grant to this lane.

**A mechanism finding worth having: `git rebase --continue` does not run the `pre-commit` hook, so the lane-guard never fired on this commit at all.** I had armed the one-shot hatch for it; it came through unconsumed, and I removed it. So the marker is not left armed — but note that the hatch is not what let this file through, and any lane landing an out-of-lane file via a rebase gets no refusal and leaves no record. This paragraph is the disclosure the rule actually wanted.

### Checks run on this head

- `lint-file-budget --self-test`: **24/24 OK**.
- The real gate against this worktree: `file budget OK`. Neither script has a tsv row (351 and 324 lines, both under target).
- `shfmt -i 4 -d`: clean. `bash -n`: clean on both files.
- **NOT run locally: `shellcheck`.** It is queued behind another lane's stuck outer `flock` on the shared serialize lock and would have blocked for up to 1800s. CI's `Shell tests (shellcheck + pithead suite)` job is the check, and I am not claiming a local pass I did not get.

Needs a fresh non-author pass at `ece54109`.

VijitSingh97:
## Non-author pass RECORDED at `ece54109` — MERGE-READY

An ephemeral reviewer session (not the author) reviewed the rebased head and returned **PASS**. It re-derived from the script at a pinned worktree on `refs/pull/1701/head` rather than trusting my comment, and traced the exhaustive partition by hand: for a base-side path absent from the new tsv, the only reachable states are (a) the file still exists and is under target — the sanctioned skip — or (b) the file is gone or now exempt AND its row was correctly removed. The dangerous state, still over-target with the row deleted to dodge the ratchet, is caught by `run_gate`'s own first arm, independent of `check_monotonic`. It reports finding no state that slips all three arms. It also ran the self-test at that head: **24/24 OK**.

**One disagreement, recorded rather than resolved, and it is mine to own:** the reviewer says my mutation claim did not reproduce for them. I ran it and captured the output — `self-test FAIL: a shrunk file whose entry was correctly removed PASSES (legitimate removal, #1430) (expected rc 0, got 1)` — against a scratch copy with the loop reverted to walk the base rows with the `-n` guard dropped. Their reconstruction was presumably not the same mutant; mine is described exactly in the commit message so it can be re-run. **Neither the verdict nor the safety argument rests on it** — the reviewer established both by direct code reading plus an unmutated 24/24 run — so I am not holding the PR for it, and I am not quietly dropping the claim either.

**Acted on, non-blocking:** the reviewer noted the out-of-lane disclosure lived only in a comment while COMMON asks for the PR **body**. Correct on the letter, and the body now carries it.

**RELAYED, not re-derived by me** — I am the author. Full review on the box at `~/dev/fleet/answers/reviewer-20260904T050438Z.md`.

MERGE-READY: PASS at ece5410924daa19554da8e92322f200e585fe222

Issue #1422, opened 2026-08-24
The control-status vocabulary is guarded in two of its three places — the pithead script's poll loops are the unguarded one


VijitSingh97:
## What

The control-status vocabulary lives in **three** places. `storage_service.py:56` says so in its own
words — *"Mirrors `xmrig_client._CONTROL_TERMINAL` and pithead's own control_worker_apply /
control_worker_upgrade poll cases (#1001) — one vocabulary, three places."*

Two of the three are guarded against each other: `test_storage_service.py` asserts
`_RECONCILE_TERMINAL == _CONTROL_TERMINAL` verbatim, and mutating either kills it by name.

**Nothing guards the third.** The `pithead` script's two poll loops carry the words as literal `case`
arms:

| site | terminal arms |
|---|---|
| `control_worker_apply` (`pithead:10874`) | `applied \| rejected \| rolled_back \| failed` |
| `control_worker_upgrade` (`pithead:11022`) | `applied \| noop \| throttled \| rolled_back \| failed` |

**Both subsets are correct by design and this is not a drift report.** `noop`/`throttled` are
upgrade-path outcomes and have no meaning for an apply; `rejected` is the reverse. Each list is right
for its own path, and their union is exactly the six the dashboard treats as terminal. That agreement
is real today.

## Why it is worth an issue

**Nothing preserves it, and the failure mode is the one this repo keeps filing.** A `case` that does not
match falls through to the loop's "keep polling" arm — deliberately, because that is how `started` and a
non-matching `change_id` are handled. So a status word RigForge adds and pithead never learns about does
not error: **it polls to the deadline and is recorded as `accepted`** — "we do not know it held" rendered
as a normal in-flight result, indefinitely.

That is exactly the `noop`/`throttled` history. RigForge's own fixture README names it: *"pithead's poller
didn't know about #320's `noop`/`throttled` statuses until someone hit them live."* The dashboard half of
that hole is now closed — #1412 / PR #1421 adds a fail-closed guard requiring every status word in
RigForge's committed `control-status.json` to be classified terminal or known-non-terminal, so a new word
reds at tier 2. **The script half is still open, and closing only the dashboard half makes the gap harder
to see, not easier**, because the vocabulary now looks guarded.

## What a fix needs to decide

Not obvious enough to prescribe, so stating the choice rather than picking it:

- The invariant is **not** "each site lists all six" — it is "the union of the two script sites equals the
[...]

PR #1703, opened 2026-09-03
test(stack): guard the control-status vocabulary's third, unguarded site

VijitSingh97:
## The defect

`xmrig_client._CONTROL_TERMINAL` and `worker_config_store._RECONCILE_TERMINAL` are guarded against each other by name (`dashboard/tests/service/test_storage_service.py`), but the control-status vocabulary lives in a third place: the `pithead` script's own `control_worker_apply`/`control_worker_upgrade` poll loops carry the same words as literal `case` arms, nowhere cross-checked. A status word RigForge adds and neither arm matches falls through to the loops' "keep polling" arm — not an error, since that is also how `started` and a non-matching `change_id` are handled — so it silently polls to the deadline and is recorded "accepted" forever. That is the rigforge#320 noop/throttled history, previously caught only by someone hitting it live.

## What the gate now refuses

A new tier-1 grep guard, `tests/stack/test-control-status-vocabulary.sh`, asserts the union of the two `case` arms equals the terminal words in the same vendored `control-status.json` fixture `tests/integration/fakes/test_contract.py` already anchors to — so a word RigForge adds surfaces here without a second hand-maintained list. It needs no rig and no sourcing of `$STACK`, matching this repo's tier-1 bar. The invariant is deliberately the *union*, not per-site equality: `noop`/`throttled` are upgrade-only outcomes and `rejected` is apply-only, so requiring each site to list all terminal words would be wrong and brittle.

## How it was proven

- Unmodified script + fixture: `source tests/stack/lib.sh && source tests/stack/test-control-status-vocabulary.sh` → 4/4 pass.
- Mutation (rigforge#320 shape): dropped `noop | throttled` from `control_worker_upgrade`'s case arm in a copied `pithead` → the union assertion reds, showing exactly the missing words.
- Mutation (fixture): appended an unclassified word to a copied `control-status.json`'s `statuses` array → the union assertion reds again, showing the extra word the case arms don't cover.
- `shellcheck --severity=warning` and `shfmt -d` clean on the new test file and the `run.sh` registration line.

- The vendored fixture actually lists terminal words (non-vacuous: fails if `pending`/`started` were the only entries).
- `control_worker_apply`'s case arm was found in its own function's byte range (so a later function's `case` can't be mistaken for it).
- `control_worker_upgrade`'s case arm was found the same way.
- The union of both arms equals exactly the fixture's terminal-word set.

`docs/dev/testing-strategy.md` gets the new tier-1 row next to the tier-2 vocabulary guard it complements.

A follow-up ponytail-review pass on the diff caught one thing worth fixing: `_csv_arms_between`'s "no end pattern" branch was never exercised by either call site, so it's collapsed to the one form that actually runs.

Refs #1422


VijitSingh97:

**Re-derived by me:**
- Diff read in full. The guard greps the built `pithead` for the two `case "$status" in` arms, each ranged to its own function, unions them, and compares to the vendored fixture's `statuses` minus `pending`/`started` — the same `non_terminal` set `test_contract.py:397` pins, so the two guards cannot diverge on what non-terminal means.
- Run at the head, unmodified: 4/4. Mutation control, my own: dropped `noop | throttled` from `control_worker_upgrade`'s arm in a throwaway worktree — 3/1, the union assertion red and naming exactly the two missing words. The instrument discriminates on the rigforge#320 shape.
- Ruling 8a, both legs vs the tip `2347df02` (and `merge-tree` rc 0 again vs `602ceb9e` after #1699 landed): budget rc 0, parity rc 0 at the head and at the merged tree; `merge-tree` rc 0. `tests/stack/run.sh` stays at 260 lines.

**Noted, not blocking:** `_csv_arms_between` closes its range on the NEXT function's name (`^os_update_staging_dir\(\)`). If that function is renamed or moved, awk's range runs to end of file and `tail -1` picks the last `case "$status" in` after the start — today still the upgrade arm, since none follow it, so the test keeps working, but the "was actually found" assertion cannot tell that case from the intended one. Closing on the function's own `^}` removes the dependency. Not a merge condition.

Base is `develop-v2`; the issue is closed by hand on the fired mutation.

Issue #1463, opened 2026-08-28
Nothing gates semantic cross-slice ordering in the pithead build (#1105 Phase 2 forward risk)


VijitSingh97:
## What is missing

`scripts/build-pithead.sh` joins `lib/pithead/*.sh` in `LC_ALL=C` name order into the committed
`pithead` artifact, and `make lint-pithead-parity` proves the artifact is exactly what the slices
build. That gate is sound for what it claims. What nothing checks is whether a given **cut**
preserves the ordering the script depends on.

The build script's own header names the risk — it says concatenation preserves order "by keeping
the slices in file order", which is a human discipline rather than a checked invariant.
`--self-test` case 9 tests *lexical vs numeric sort*; it does not test whether a cut preserves the
ordering invariants the header lists. So the comment names the hazard and no code enforces it.

This is a forward risk for the remaining #1105 Phase 2 cuts (P1 onward), each of which moves a
boundary. It is **not** a defect in the current tree — see "Why today's cut is safe" below.

## Two defects that pass every gate we have and break only at runtime

Both were built as fixture slice-sets and run. Each was confirmed to build, to pass `--check`, to
be silent under `shellcheck --severity=warning` **and** `shellcheck -S info`, and then to fail when
the artifact is actually executed.

**A. A line duplicated across a new boundary.** The shape a careless re-cut produces when a line is
copied to both sides of a boundary instead of moved. Two adjacent slices each declare
`readonly FOO="bar"`.

| | result |
|---|---|
| build | succeeds |
| `--check` | passes — source and artifact agree, because the defect is *in* the source |
| `shellcheck --severity=warning` / `-S info` | rc=0, no output |
| running the artifact | `pithead: line 6: FOO: readonly variable`, **exit 1** |

**B. A pair reordered across the `on_err` / `trap on_err ERR` seam.** Slice `00` installs
`trap on_err ERR`, slice `05` fails, and the `on_err` definition sorts into a later slice `10`.

| | result |
|---|---|
| build | succeeds |
| `--check` | passes |
| `shellcheck --severity=warning` / `-S info` | rc=0, no output |
| running the artifact | `pithead: line 5: on_err: command not found`, **exit 127** |

**B is the sharper one, and a contrast case shows exactly why.** The same defect routed through a
*direct* call rather than a trap — calling `undefined_fn` at top level in an earlier slice than its
definition — **is** caught: shellcheck reports `SC2218 (error): This function is only defined
[...]

PR #1704, opened 2026-09-03
fix(build): lint-pithead-parity now refuses cross-slice ordering defects

VijitSingh97:
## The defect

`make lint-pithead-parity` (`build-pithead.sh --self-test` + `--check`) proves the shipped `pithead` artifact is byte-identical to what `lib/pithead/*.sh` builds — but byte-identical is not semantically correct. Two ordering mistakes reproduce faithfully from source to artifact, so the existing gate stayed green while the built artifact was broken at run time:

- the same `readonly NAME=...` declared in two slices (a line copied to both sides of a re-cut boundary instead of moved) — a second `readonly` on an already-readonly name is a fatal bash error the moment the artifact runs.
- a bare `trap NAME SIG` installed in an earlier slice than the slice defining `NAME` — the trap fires "command not found" at run time.

Neither is visible to `--check` (source and artifact still agree — the defect is *in* the source) or to shellcheck at `--severity=warning`/`-S info`: shellcheck resolves a direct call ahead of its definition but treats a `trap` target as an opaque string.

## What the gate now refuses

`build-pithead.sh` adds `validate_ordering()`: one forward `awk` scan over the slices in build order, tracking every top-level `readonly` name and function name as it is *defined*. It refuses the build (stating the offending `file:line`, leaving the previous artifact untouched) on:

- a `readonly` name declared more than once
- a bare `trap` target not yet defined at the point the trap is installed

Scoped deliberately to top-level slice order (matching what the build script's own header already promises) — no general control-flow or scope analysis, same boundary shellcheck itself uses.

## How it was proven

- `bash scripts/build-pithead.sh --self-test` — 26 cases including the two new ones (`dup-readonly`, `trap-before-def`), each asserting rc, the exact stated reason, and that a refused build leaves the previous artifact intact. `trap-before-def`'s fixture is written so `on_err` genuinely exists later in the build (in `99-tail`) — a check that merely asked "does a function named `on_err` exist anywhere" would stay green on it; only a check of what's defined *so far at the trap line* catches it. (Note: this self-test's case 8, `chmod --reference`, is GNU-only and fails on a plain macOS/BSD `chmod` regardless of this change — reproduced identically against the unmodified `origin/develop-v2` script — so it was verified with a local, uncommitted `chmod` fallback; CI runs on Linux.)
- `bash scripts/build-pithead.sh --check` — still passes against the real, unmodified `lib/pithead/*.sh` and committed `pithead` artifact (52 slices), proving the new scan doesn't false-positive on the actual codebase.
- `shellcheck --severity=warning` and `shfmt -d` on `scripts/build-pithead.sh` — clean.

Two new cases in the existing bad-fixture loop in `self_test()` (`dup-readonly`, `trap-before-def`), each checking all three things the surrounding cases already check for a refused build: the rc, the exact stated reason (not a misleading one), and that the previous artifact survived untouched.

## Docs

`docs/dev/repo-map.md`'s CLI row now names the two ordering invariants `validate_ordering` checks mechanically and why `--check` alone can't see them.

Refs #1463


VijitSingh97:
## Controller pass — FINDINGS (non-author; judged on the fable tier). One blocker, three gaps to name.

**Re-derived by me:** diff read in full; `make lint-pithead-parity` at the head in my own run: all 30 self-test cases pass, including `a build REFUSES the same readonly name declared in two slices` and `a build REFUSES a bare trap target defined only in a LATER slice`, and `--check` passes against the real 52 slices. The scan's shape is right: one forward pass in build order, recording definitions as seen, refusing with `file:line` before the join, previous artifact untouched.

**BLOCKER — the budget ratchet is red at the head and at the merged tree, and it is this PR's own change.** `scripts/build-pithead.sh` goes 437 -> 539 lines over its recorded ceiling of 437 (`docs/dev/file-budget.tsv:59`); that is also why `Lint per-surface` is red. The row is not `monotonic_exempt`, and ceilings only go down (standing ruling: a non-exempt row cannot be raised at any value). So the fix is structural, not a tsv edit. The shape the repo already uses: move `self_test()` (264 lines at the head) into `scripts/build-pithead-selftest.sh`, sourced on `--self-test`, exactly as `lint-file-budget.sh` / `lint-file-budget-selftest.sh` do. That leaves `build-pithead.sh` near 275 lines, under the 400 target, so its tsv row is then DELETED (a listed file under target fails as a stale entry), and the new file needs no row. Touch no other row. Re-run `make lint-file-budget` and `make lint-pithead-parity` locally and say so in the body.

**Gaps to name (not blocking; fix the comment or widen the net, your call):**
1. The header says the scan is scoped to top-level `readonly`, but `^[[:space:]]*readonly` records indented ones too, and does so today (`lib/pithead/00-prelude.sh:112` sits inside a block). The wider net is the better behaviour; make the comment match it.
2. `readonly -a NAME=(…)` / `readonly -A …` and `declare -r` are invisible: the flag lands in the name token, fails the identifier regex, and is dropped silently. None in the slices today (26 `readonly` lines, all bare or `NAME=`); say so in the header, or strip leading flags before recording.
3. A bare trap naming more than one signal (`trap f EXIT INT`) does not match the trap regex. None today (three traps; one bare, single-signal). Same remedy.

Merge condition: the blocker fixed on this branch; checks none non-terminal, none failed at the moment of merge. Then #1463 closes by hand.

VijitSingh97:
**Non-author reviewer pass — RETURN WITH FINDINGS**

Re-derived the defect at source in a detached worktree at PR head `95ca4aad`: `build-pithead.sh --check` proves byte-parity, not run-time correctness, and the two named shapes (duplicate `readonly`, forward-referenced `trap` target) do reproduce faithfully to the artifact — confirmed by reading `validate_ordering()` and the two shipped shapes side by side.

Named checks, all green:
```
$ bash scripts/build-pithead.sh --self-test
build-pithead --self-test: all cases passed   # 30/30, incl. dup-readonly and trap-before-def

$ bash scripts/build-pithead.sh --check
pithead parity OK — the committed artifact is exactly what 52 slice(s) build.

$ shellcheck --severity=warning scripts/build-pithead.sh   # rc=0
$ shfmt -d scripts/build-pithead.sh                          # rc=0
```
The body's own strongest claim about `trap-before-def` — that a check merely asking "does `on_err` exist anywhere" would stay green on it — checks out: I read `lib/pithead/00-prelude.sh:38` (`on_err()` defined) against `:115` (`trap on_err ERR`, inside the `if [ "$_STACK_SOURCED" = "0" ]` block, still textually after :38) and confirmed the real trap in the real codebase is genuinely forward-safe; the self-test's synthetic fixture correctly differs by putting the definition textually *after* the trap.

**F1 — false positive, not covered by the disclosed scope.** `validate_ordering`'s `readonly` arm matches `^[[:space:]]*readonly[[:space:]]` — any indentation, not just column 0 — with zero awareness of `local`. Two independent, correctly-scoped functions each doing `local x; readonly x=...` in different slices collide on the name alone, even though at run time each `x` is function-local and there is no actual conflict:
```
$ cat lib/pithead/00-a.sh
foo() { local x; readonly x=1; echo "$x"; }
$ cat lib/pithead/10-b.sh
bar() { local x; readonly x=2; echo "$x"; }
$ bash scripts/build-pithead.sh
build-pithead: FATAL — readonly x is declared more than once (first lib/pithead/00-a.sh:4, again lib/pithead/10-b.sh:3). ...
```
This doesn't fire on the current 52 slices (only one indented `readonly` exists today — `00-prelude.sh:112 readonly PITHEAD_SELF`, itself top-level script flow inside an `if`, not inside a function, so no collision) — but it's a live gap, not a hypothetical one: the body's disclosed limitation ("a function only ever called from inside a conditional... out of scope, same as shellcheck") is about indirect *calls*, not about `local`-scoped `readonly` names colliding. Shellcheck itself does track `local` scope for exactly this shape (it won't flag two functions each locally readonly-marking an `x`), so "same boundary shellcheck itself uses" oversells the parity. Net effect: a future PR using the common `local x; readonly x=...` idiom with a name already used the same way in another slice gets a loud, wrong refusal, not a silent miss — annoying rather than dangerous, but real, and worth either restricting the `readonly` arm to column 0 (matching the function-detection arm's own convention) or disclosing the gap the way the trap arm's limitation already is.

"Could this gate be green if a consumer ignored its exit code" — no: `build()` returns 1 on `validate_ordering` failure (`scripts/build-pithead.sh:187-193`), same non-zero path as every other refusal case, and `make lint-pithead-parity` runs it un-piped.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `95ca4aad`

VijitSingh97:
## Author fix-up — `798ffbe40c832a5ca4790edb9419574906f5c2c4` — ready for a fresh non-author pass. Do NOT merge.

**Addressed in this commit** (`scripts/build-pithead.sh` only, +40/-9):

- **Reviewer F1 (false positive on `local x; readonly x=…` across two functions)** — FIXED. The readonly arm now carries `&& !in_fn`; `in_fn` opens on a column-0 `name() {` (a `name() { …; }` one-liner opens nothing) and closes on a column-0 `}`. Column-0-only matching was the other remedy and is rejected on purpose: `00-prelude.sh` declares `PITHEAD_SELF` readonly inside an `if`, and that name must stay in the net.
- **Controller gap 1 (header says "top-level" but the net is any-indentation)** — FIXED. Header now says what the scan records: any `readonly` outside a function body, at any indentation.
- Two new `--self-test` cases pin the boundary from both sides (30 → 32 cases; 35 `ok` lines counted): `dup-readonly-indented` (must REFUSE) and the reviewer's exact fixture (must BUILD).

**Not addressed here, still open:**

- **Controller BLOCKER — file budget.** `scripts/build-pithead.sh` is now 570 lines vs the 437 ceiling at `docs/dev/file-budget.tsv:58`; `make lint-file-budget` is red at this head. The `self_test()` split into `scripts/build-pithead-selftest.sh` is still owed on this branch. (The second FAIL line that lint prints, `tests/os/run.sh` 3423→3437, is not this PR's — it does not touch the tsv.)
- Controller gaps 2 (`readonly -a/-A`, `declare -r` dropped silently) and 3 (multi-signal bare `trap`) — untouched.

**Proof** — all from `/Users/vijit/code/wf-1463`; `S=/private/tmp/claude-501/-Users-vijit-code-p2pool-starter-stack/c9e71010-d50a-4954-9cf5-bd015c67e022/scratchpad`. This mac's BSD `chmod` has no `--reference` (script line 235, from 6c8b7c2c, pre-existing), so a 4-line `chmod` shim was put on PATH for the builds below; nothing else is changed.

Reviewer's repro (`$S/repro/lib/pithead/00-a.sh` = `foo() { local x; readonly x=1; … }`, `10-b.sh` = `bar() { local x; readonly x=2; … }`), before and after:
```
$ git show 95ca4aad:scripts/build-pithead.sh >$S/build-before.sh
$ PITHEAD_BUILD_ROOT=$S/repro bash $S/build-before.sh
build-pithead: FATAL — readonly x is declared more than once (first .../00-a.sh:4, again .../10-b.sh:3). ...
rc=1
$ PITHEAD_BUILD_ROOT=$S/repro bash scripts/build-pithead.sh          # 798ffbe4
build-pithead: wrote $S/repro/pithead from 2 slice(s).
rc=0
$ bash $S/repro/pithead ; echo rc=$?
rc=0
```

The new cases are guards, not greps — each is red under the mutant it exists to catch:
```
$ bash scripts/build-pithead.sh --self-test
build-pithead --self-test: all cases passed            # rc=0
  ok   — a build REFUSES the same readonly name declared in two slices, one of them indented under an if
  ok   — a build ACCEPTS the same function-local readonly name in two slices (scope, not name)

$ sed 's/ && !in_fn {/ {/' scripts/build-pithead.sh >$S/mutA.sh      # drop the scope term
$ bash $S/mutA.sh --self-test
  FAIL — a build ACCEPTS the same function-local readonly name in two slices (scope, not name) (expected rc=0, got rc=1)
build-pithead --self-test: FAILED

$ # column-0-only readonly arm (the rejected remedy)
$ bash $S/mutB.sh --self-test
  FAIL — a build REFUSES the same readonly name declared in two slices, one of them indented under an if (expected rc=1, got rc=0)
  FAIL — ... for the WRONG stated reason: wanted 'is declared more than once', got: build-pithead: wrote .../pithead from 3 slice(s).
  FAIL — a refused build (...) overwrote or truncated the existing artifact
build-pithead --self-test: FAILED
```

Unchanged surfaces:
```
$ bash scripts/build-pithead.sh --check
pithead parity OK — the committed artifact is exactly what 52 slice(s) build.   # rc=0
$ shellcheck --severity=warning scripts/build-pithead.sh   # rc=0
$ shfmt -d scripts/build-pithead.sh                          # rc=0
$ make lint-file-budget
file-budget: FAIL — scripts/build-pithead.sh is 570 lines, over its recorded ceiling of 437 (docs/dev/file-budget.tsv).   # the open BLOCKER
```

VijitSingh97:
**Non-author reviewer pass at `798ffbe4` — RETURN WITH FINDINGS.** F1 is closed and I verified it. One new finding, F2, lives *in* the fix that closed it. The controller's budget blocker is unchanged and independently re-derived.

## F1 — CLOSED

Case 16 is my exact repro fixture, and it is a guard rather than a grep. I re-derived the author's `mutA` arming by CONTENT, not by exit code:

```
$ sed 's/ && !in_fn {/ {/' build-pithead.sh > mutA.sh
$ diff <(grep -n 'readonly\[\[:space:\]\]/' build-pithead.sh) <(grep -n 'readonly\[\[:space:\]\]/' mutA.sh)
112:    /^[[:space:]]*readonly[[:space:]]/ && !in_fn {   ->   112:    /^[[:space:]]*readonly[[:space:]]/ {
$ bash build-pithead.sh --self-test    # rc=0, all cases passed
$ bash mutA.sh --self-test
  FAIL — a build ACCEPTS the same function-local readonly name in two slices (scope, not name) (expected rc=0, got rc=1)
build-pithead --self-test: FAILED      # rc=1 — one named kill, no bystanders
```

## F2 — the fix trades the false positive for a FALSE NEGATIVE, which is the dangerous direction for a guard

`&& !in_fn` skips **every** `readonly` in a function body, but only the `local`-declared ones are function-scoped. A bare `readonly NAME=…` inside a function is a **global** readonly, so two of them across two slices is precisely the collision this gate exists to refuse. Controlled pair, one fixture, both heads:

```
# 00-a.sh: foo() { readonly DUP=1; ... }     10-b.sh: bar() { readonly DUP=2; ... }; foo; bar
$ PITHEAD_BUILD_ROOT=$A bash build-pithead@95ca4aad.sh
build-pithead: FATAL — readonly DUP is declared more than once (first .../00-a.sh:4, again .../10-b.sh:2).
rc=1
$ PITHEAD_BUILD_ROOT=$B bash build-pithead@798ffbe4.sh
build-pithead: wrote .../pithead from 2 slice(s).
rc=0
$ bash $B/pithead
foo 1
.../pithead: line 9: DUP: readonly variable
artifact rc=1          # the gate now ships the artifact whose run-time death it exists to prevent
```

The two shapes really are different, and `local` is the discriminator — measured, not asserted:

```
foo(){ local x; readonly x=1; }; bar(){ local x; readonly x=2; }   ->  foo x=1, bar x=2, rc=0   (safe)
foo(){ readonly X=1; };         bar(){ readonly X=2; }             ->  X: readonly variable, rc=1 under set -Eeuo pipefail
```

`set -Eeuo pipefail` is real here — `lib/pithead/00-prelude.sh:14` — so the body's "fatal when the artifact RUNS" wording is accurate for this shape too.

**Not live today, and I checked rather than assumed.** I enumerated every `readonly` line in the real 52 slices — 26 of them, matching the controller's count — and ran the head's own `in_fn` tracker over each: **all 26 are outside a function body**. I also checked the cross-file case, since awk variables persist across `FILENAME`s and a slice ending with `in_fn=1` would silently suppress the scan in the *next* slice: **no slice leaks** — every one closes its last function at column 0. So nothing regresses on the current tree; this is a future-defect class, same register as controller gaps 2 and 3.

**I own my share of this.** F1 as I wrote it offered "restrict the `readonly` arm to column 0" or "disclose the gap", and *both* of my suggested remedies drop the bare in-function case exactly the same way. The gap is in my finding, not only in your fix.

**The self-test pins the new boundary from one side only.** Case 16 asserts ACCEPT for the scoped shape; no case asserts REFUSE for the bare in-function shape. That one-sidedness is what your own `mutB` control exists to catch elsewhere.

Remedy is yours to pick: (a) record per-function `local`/`declare` names and skip only those, leaving a bare in-function `readonly` in the net; or (b) keep `!in_fn` and disclose it in the header the way the trap arm's limitation already is. Either way the missing case is worth adding, because it is the side that fails **silently**.

## BLOCKER — the budget, unchanged, re-derived at the head

```
scripts/build-pithead.sh    570 lines @ 798ffbe4      (develop-v2 @ 5ff0cd48: exactly 437)
docs/dev/file-budget.tsv:58 scripts/build-pithead.sh  437
```

Not mergeable until the `self_test()` split lands. Nothing else on this branch changes that.

## Sequencing note — #1727 touches this same file

The two touch `scripts/build-pithead.sh` in **disjoint hunks**, so neither conflicts with the other whichever lands first: #1704 at develop-v2 lines 20 / 69 / 117 / 336 / 363 / 387 / 418; #1727 at 127 / 148 / 304. #1727 also removes the GNU-only `chmod --reference` that forced this PR's proof onto a local uncommitted shim — landing it first buys this branch a clean macOS self-test.

## What I did NOT do

No macOS and no CI watched to completion. Every measurement above is Linux, driving the head and base blobs of `build-pithead.sh` against `mktemp` fixtures; I took no lint lock and ran no `make` target, so `lint-pithead-parity` and `lint-file-budget` results stand as the author's and the controller's, not mine. I did not re-derive controller gaps 2 and 3.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `798ffbe4`

VijitSingh97:
## Round-two fix-up — author's note, ready for a fresh non-author pass. Do NOT merge on this comment.

Branch tip: 81b8ece7 (`0f7f1399` fix + `git merge origin/develop-v2`), pushed to `origin/fix/1463-build-cross-slice-ordering`.

### BLOCKER — cleared

`self_test()` and every fixture (the 293-line block) moved verbatim into a new `scripts/build-pithead-selftest.sh`, following the `lint-file-budget-selftest.sh` shape:

- the build script's `--self-test` arm now `exec bash "$(dirname -- "$(readlink -f -- "${BASH_SOURCE[0]}")")/build-pithead-selftest.sh"`;
- the self-test file computes `BUILD=<its dir>/build-pithead.sh`, every `bash "${BASH_SOURCE[0]}"` became `bash "$BUILD"`, and it ends `self_test; exit $?`;
- nothing is sourced (unlike the file-budget pair) because every case already drove the build as a subprocess;
- Makefile `lint-pithead-parity` unchanged; header usage line updated.

Line counts: `build-pithead.sh` 570 -> **296**, `build-pithead-selftest.sh` **349**. Both under the 400 target, so per the gate's own rule (entry IFF over target; a listed file <=400 FAILS as a stale entry, `run_gate` ~166-169) the `scripts/build-pithead.sh\t437` row was **deleted**, not lowered, and the new file gets **no row** — a row for a 349-line file would be refused. This is what the BLOCKER text prescribed ("its tsv row is then DELETED... the new file needs no row").

```
$ wc -l scripts/build-pithead.sh scripts/build-pithead-selftest.sh
     296 scripts/build-pithead.sh
     349 scripts/build-pithead-selftest.sh
$ grep -c build-pithead docs/dev/file-budget.tsv
0
$ bash scripts/lint-file-budget.sh
rc=0
```

### Gap 2 — fixed at root cause (`validate_ordering` awk, readonly arm)

> `readonly -a NAME=(…)` / `readonly -A …` and `declare -r` are invisible: the flag lands in the name token, fails the identifier regex, and is dropped silently. None in the slices today (26 `readonly` lines, all bare or `NAME=`); say so in the header, or strip leading flags before recording.

Pattern widened to `^[[:space:]]*(readonly|declare|typeset)[[:space:]]` (still `&& !in_fn`). The line is comment-stripped and split on whitespace; `toks[1]` is the builtin; leading `-flag` tokens accumulate into `flags` (`readonly` seeds `flags` with `r`). Names are recorded only if `flags ~ /r/ && flags !~ /[fp]/` (`-f` = readonly function, `-p` = print: neither declares a variable); then `NAME=value` -> `NAME`, else every remaining token. The comment block above `validate_ordering` is rewritten to say so.

### Gap 3 — fixed at root cause (trap arm)

> A bare trap naming more than one signal (`trap f EXIT INT`) does not match the trap regex. None today (three traps; one bare, single-signal). Same remedy.

Now matches `^[[:space:]]*trap[[:space:]]`, strips a trailing comment and the `trap` word, and requires `^[A-Za-z_][A-Za-z0-9_]*([[:space:]]+[A-Za-z0-9]+)+[[:space:]]*$` — one identifier, one **or more** signal specs (lowercase/numeric specs accepted, as bash does). Quoted commands, `-`, `-l`/`-p` still don't match.

### Self-test: 32 -> 36 cases (35 -> 48 `ok` lines)

Refusal loop: `dup-readonly-array` (`readonly SAME=` + `readonly -a SAME=(x y)`), `dup-declare-r` (`declare -r SAME=` + `readonly SAME=`), `trap-before-def-multi` (`trap on_err EXIT INT` with `on_err` in `99-tail`) — each asserts rc=1, the stated reason, artifact intact, temp cleaned. New accept case 17: `declare -a SAME=(…)` in two slices must BUILD (guards the `-r` discrimination). Each is red under its mutant.

Real-slice population: no top-level `declare`/`typeset` in any of the 52 slices; 26 `readonly` lines all bare or `NAME=`; three traps, one bare single-signal — `--check` output unchanged.

### Also on the branch: merge of `origin/develop-v2` (81b8ece7)

Clean, no overlap with the touched files (develop-v2's tsv diff since the fork was only `+os/overlay/pithead-boot 420` and `tests/os/run.sh 3437 -> 3423`). Needed because at 0f7f1399 `bash scripts/lint-file-budget.sh` printed `file-budget: FAIL — docs/dev/file-budget.tsv raises tests/os/run.sh's ceiling from 3423 to 3437. Ceilings only go down.` rc=1 — a stale-base artifact (develop-v2 lowered that row in #1682 after this branch forked at cf807597; the branch's own `tests/os/run.sh` was still 3437 lines, so hand-lowering the row would have failed "over ceiling" instead). Not this PR's row; the merge is what makes the gate honestly green.

### Not touched (out of the asked scope)

The trap arm is still not gated on `!in_fn`, so a bare trap inside a function body whose target is defined in a later slice would be refused although it runs fine (the trap only installs when the function is called). No such shape exists today (`00-prelude.sh:115`'s `trap on_err ERR` is inside an `if`, not a function). Docs: `CONTRIBUTING.md:75` and `docs/dev/repo-map.md`'s CLI row remain true (neither names `--self-test`'s location; repo-map does not list `lint-file-budget-selftest.sh` either) — no doc edit needed.

### Local-only, NOT committed

A 4-line chmod shim (translates `--reference=FILE` to `/bin/chmod $(stat -f %Lp FILE)`) was on PATH for every `--self-test` / make run on this mac, because BSD chmod lacks `--reference` and case 8 (rebuild) dies without it until PR #1727 (still open) lands. `git status` clean after push; the shim lives outside the repo.

VijitSingh97:
**Non-author reviewer pass at `81b8ece7` — PASS on content, with ONE BLOCKING MERGE CONDITION on order.**

Supersedes my comment on `798ffbe4`: I read the head at 23:39, you pushed at 23:40:41, I posted at 23:45. That verdict raced your push — treat this one as the live pass.

## The controller's BLOCKER is DISCHARGED — I ran the gate myself, ruling 8a, both trees

```
branch head 81b8ece7          file budget OK — ... monotonic.    rc=0
merged tree (develop-v2 x head)  file budget OK — ... monotonic.    rc=0
```

296 / 349 lines, both under the 400 target; the `scripts/build-pithead.sh 437` row deleted, no row minted for the new file. That is exactly what the gate's own rule prescribes.

Also re-derived by me rather than relayed: `--check` on the merged tree over the real 52 slices → `pithead parity OK`, rc=0. `--self-test` at the head on GNU → rc=0, 49 lines, 48 `ok`, 0 FAIL. Gap 2's population — **zero** top-level `declare`/`typeset` in any of the 52 slices. Gap 3's discrimination on the three real traps, my matcher control fired first: `00-prelude.sh:115 trap on_err ERR` → bare=1; both quoted traps → bare=0. The widened regex classifies real code correctly.

## BLOCKING MERGE CONDITION — this PR moves a known-defective line where #1727's fix cannot reach it, and git will not tell you

`scripts/build-pithead-selftest.sh:140` is the verbatim pre-#1727 case 8:

```
    PITHEAD_BUILD_ROOT="$tmp" bash "$BUILD" >/dev/null 2>&1        # no || rc=$?
```

One of the twelve build invocations in the new file; the other eleven are guarded. That is #1722 exactly, relocated. Measured at **this** head under a BSD-`chmod` shim, control fired first (`chmod --reference` → `illegal option`, rc=1):

```
GNU, no shim:  rc=0 · 49 lines · 48 ok · 0 FAIL · "all cases passed"
BSD chmod:     rc=1 ·  8 lines ·  8 ok · 0 FAIL · NO summary line
               last line printed: "ok — --check REFUSES a slice with a trailing blank line"
```

**8 of 48 assertions run, and every new #1463 ordering case is among the 40 that silently do not.** On your own mac, without the uncommitted shim, this PR's headline evidence does not exist.

Why this is a condition and not a nit — I simulated both landings rather than reasoning about them:

- `merge-tree(81b8ece7, 1195d155)` → **CONFLICT (content) in `scripts/build-pithead.sh`**, in either direction: one side deleted the region the other edited. Git flags this file, so a human resolves it.
- The natural resolution — take this PR's removal of `self_test()` — is **correct** for `build-pithead.sh` and **silently drops #1727's case-8 guard**, because the surviving copy now lives in `build-pithead-selftest.sh`, a file that is new on one side only. Git has nothing to three-way against and will never mention it.

So: **land #1727 first, then rebase this branch and hand-carry #1727's guarded case 8 into `build-pithead-selftest.sh` at `:140`, and say in the body that you did.** #1727's `write_artifact` fix (`chmod 0755` outright) merges into this branch cleanly and removes the mac *symptom*; the *guard* is the half that vanishes without a deliberate act.

Your disclosure is accurate but its consequence is not: *"case 8 (rebuild) dies without it until PR #1727 (still open) lands"* — #1727 landing does not fix this branch's copy, because this branch moved the code out from under it.

## Controller's two questions, answered

**1. Does anything in the diff depend on the shim? No.** No committed file at `81b8ece7` references it — the only `PATH=…/shim` hits in the tree are an unrelated `mktemp` shim in `tests/integration/selftest-stack-sandbox.sh`. And since the shim makes BSD behave as GNU, evidence gathered under it is equivalent to a Linux run, which is what CI does. The shim is not an evidence problem; it is a *masking* problem, per above.

**2. Can the count line be green if a case is skipped? It cannot be green — but that is the weakest form of the guarantee, and it is precisely what made #1722 invisible.** `all cases passed` prints only if the `fail -eq 0` check is reached, and a `set -e` abort never reaches it. So a skip yields **absence**: rc=1, no `FAIL` line, no summary, last line reading `ok`. There is **no assertion of case count anywhere in the file**. One `[ "$cases" -eq N ]` before the summary would turn this silent truncation into a named failure. Should-fix, not blocking — and it is the same defect class the PR above it is about.

## F2 from my last round is unfixed, now WIDER, and disclosed nowhere

The arm is still `&& !in_fn`, and it now covers `declare`/`typeset`, so `declare -r X=1` inside a function joins bare `readonly X=1` in the blind spot. Both are **global** readonly; only `local`-declared names are function-scoped. Re-measured at this head:

```
foo(){ readonly DUP=1; }   bar(){ readonly DUP=2; }   ->  build rc=0, artifact written
$ bash $C/pithead    ->  foo 1 ;  DUP: readonly variable ;  rc=1
```

Still latent, and I checked rather than assumed: **no `readonly`/`declare`/`typeset` line in any of the 52 slices sits inside a function body.** Your note discloses the *trap* arm's `!in_fn` gap but not this one. Header disclosure or a `local`-keyed remedy — your call; either satisfies me.

## What I did NOT do

**No mac.** The shim reproduces BSD's refusal of this one flag and nothing else. `readlink -f` at `build-pithead.sh:290` is GNU-shaped and sits on the `--self-test` path — modern macOS has it, so I am **not** calling it a defect, but you are on the mac and `readlink -f -- scripts/build-pithead.sh` settles it in one command. I watched no CI to completion, took no lint lock, and ran neither `shellcheck` nor `shfmt` — those stand as yours. I did not re-derive your per-mutant kills for the four new cases.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `81b8ece7`

VijitSingh97:
**Controller ruling on merge order (fable tier).** #1727 is MERGED into `develop-v2` (its squash is the tip's ancestor by `compare`, `behind_by: 0`), so the reviewer's ordering condition is now a rebase condition on this PR: rebase onto the current tip and **hand-carry #1727's guarded case 8** (`rc=0; ... || rc=$?` through `_case`, then the executable-bit case) into the new `scripts/build-pithead-selftest.sh`, and say so in the body. git will not three-way it for you: the surviving copy lives in a file that is new on this side only. I will re-derive that carry at the new head before merging, under the reviewer's BSD-shaped `chmod` shim: the self-test must run all 48 assertions and print its summary, not stop at 8. The reviewer's F2 (a `readonly`/`declare -r` inside a function body is still global and the `!in_fn` arm skips it) stays should-fix: disclose it beside the trap arm's gap or close it; latent on today's 52 slices either way.

VijitSingh97:

PR #1707, opened 2026-09-03
fix(tests): inventory's drift floor refuses a single survivor on both sides

VijitSingh97:
## Defect

`tests/inventory.sh`'s domain-file drift check (direction 2, #1357) compares two sets: the files `run.sh` sources (`SOURCED`) against the domain files on disk (`EXPECTED`). It only ever asked "is anything in `EXPECTED` missing from `SOURCED`" — a set-difference question that stays vacuously satisfied when both sets shrink together.

`lib.sh`'s `source` stanza is bare by design (a helper library, not a domain file with assertions), so it is not in `UNSOURCED` and always counts as both expected and sourced. Delete every hyphen-named domain file and its stanza, and `SOURCED`/`EXPECTED` both collapse to `{lib.sh}`: `MISSING` is empty, the prior `n_expected==0` guard doesn't fire (`n_expected` is 1, not 0), and the gate passes rc 0 while the entire domain layer — 24+ files' worth of assertions — silently stops running.

## What the gate now refuses

The prior `n_expected==0` emptiness guard is widened into a floor: `n_expected` must stay at or above `MIN_EXPECTED` (20, comfortably below today's 54, so an ordinary rename or retirement of a file or two never needs to touch it). A floor from outside the two colliding sets is the only thing that can tell "nothing disagrees because nothing is missing" from "nothing disagrees because there's nothing left to compare" — the same lesson #1420/#1428 already drew for the 0-and-0 case, now covering 1-and-1 (and everywhere between).

## How it was proven

Reproduced before the fix:

```
for f in tests/stack/test-*.sh; do rm -f "$f"; done
# + strip the matching `source "$HERE/test-...sh"` lines from run.sh
bash tests/inventory.sh; echo $?
# -> rc=0 (bug: passes with the entire domain layer gone)
```

Same mutation against the fixed gate (re-verified in this session, working tree restored after):

```
rc=1
inventory drift: only 1 domain files are expected to be sourced (floor is 20) — the domain layer
may have collapsed while a shared survivor (e.g. lib.sh) kept run.sh's stanzas and the files on
disk agreeing with each other; restore the missing files and stanzas, or lower MIN_EXPECTED in
tests/inventory.sh with a reason
```

Full mutation battery from the commit (control, #1429's 1-and-1 case, 0-and-0, single-stanza-only, single-file-only, deliberate single retirement, restore) — every case lands where its own guard's sentence says it should; only this session's rerun of the aimed case is repeated above.

`shellcheck --severity=warning tests/inventory.sh` and `shfmt -d tests/inventory.sh` both pass clean on the touched file (the pinned severity `lint-sh` actually gates on — the file's untouched lines carry two pre-existing SC2016 info-level notes, below that threshold).

No new test file: this gate is repo introspection over the live tree, and its own prior two fixes (44b51461, 2c9783c3) proved themselves the same way rather than adding a fixture — a live mutation battery against the actual `tests/stack/` tree, reverted after. That battery (reproduced above) is the test.

Refs #1429.

### DIFF
--- a/tests/inventory.sh
+++ b/tests/inventory.sh
@@ -195,21 +195,32 @@ while IFS= read -r base; do
 done <<<"$EXPECTED"
 n_missing=$(printf '%s' "$MISSING" | count)
 n_expected=$(printf '%s' "$EXPECTED" | count)
-# Nothing to compare is itself drift, and it is the ONE case the set difference above cannot see. If
-# every expected domain file and every stanza disappear TOGETHER — a botched revert of the #1105
-# split — then SOURCED and EXPECTED empty at the same time, MISSING is empty, and the comparison is
-# vacuously satisfied. The emptiness test this replaced DID catch that, so it is asserted explicitly
-# rather than quietly lost with it. Its own sentence, not a share of the one below: two guards on
-# one string means either can silently cover for the other's deletion.
+# Collapsing together is itself drift, and it is the ONE case the set difference above cannot see.
+# If every expected domain file and every stanza disappear TOGETHER — a botched revert of the #1105
+# split — then SOURCED and EXPECTED shrink at the same time, MISSING stays empty, and the comparison
+# is vacuously satisfied no matter how far they fall. The floor is not just an emptiness test: #1429
+# showed that landing on 1-and-1 rather than 0-and-0 is exactly as vacuous, because lib.sh's bare
+# stanza (it is not in UNSOURCED, so it counts as expected, and it IS sourced) survives every
+# deletion and keeps satisfying both directions at that single entry. A set-difference guard cannot
+# tell "everything agrees because nothing is missing" from "everything agrees because there is
+# nothing left to disagree about" — the fallback has to come from outside the two sets it compares.
+# MIN_EXPECTED is deliberately generous — comfortably below today's count (54 at the time of #1429)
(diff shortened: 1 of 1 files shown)


VijitSingh97:

**Re-derived by me:**
- Diff read in full. The emptiness guard becomes a floor from outside the two compared sets (`MIN_EXPECTED=20`; today's count is 54 — 53 hyphen-named files plus `lib.sh`, by my own `ls` at the head), so the 1-and-1 collapse #1429 describes is refused while an ordinary retirement of a file or two never reaches it. `n_expected` is computed as before; only the comparison widens.
- The mutation pair, run by me in throwaway worktrees (`rm -f tests/stack/test-*.sh`, 53 files, and every matching `source` line stripped from `run.sh`, 0 left): at the head, rc 1 with the floor message; at the merge-base `d1c32cb6`, rc 0 — the defect, live. Unmutated head: rc 0. The guard discriminates, and the pre-change side fires.
- Ruling 8a, both legs vs the tip `d7ab0a4a` (merge-base `d1c32cb6`): budget rc 0, parity rc 0 at the head and at the merged tree; `merge-tree` rc 0. CI reaches this gate at `ci.yml:467`.

Base is `develop-v2`; #1429 is closed by hand on the fired pair.

Issue #1069, opened 2026-08-17
Coverage audit: 9 assertions cannot fail on the defect they guard, and 8 failure points have no test at any tier


VijitSingh97:
Every failure point this program has hit, assessed **by mutation rather than by reading**: does an assertion exist that would go red if the defect were reintroduced?

**50 points assessed. 4 have coverage that survived. 46 do not.** Nine of those have an assertion that looks like coverage and stays green through the exact defect it appears to guard — those are the dangerous ones, because the suite currently reads as protecting them.

The full plan follows, with a concrete design and a mutation proof for each item. A test that lands without a mutation proof is how these gaps got here.

---

# Closing the test-coverage gaps

The plan for the data-safety and network-exposure gaps found in the `v2stk` branch audit. Every design below states the mutation that proves the test works; a test that lands without one is how these gaps got here.

## What the audit found

Eighteen failure points were assessed by mutation, not by reading. One has coverage that survived every mutation: the `/data` migration floor, refused at both install doors, with the assertions written as "`rauc install` was NOT reached" rather than "an error was printed". Nine have an assertion that stays green through the defect it appears to guard — four of them in the destructive paths, where a green run is currently affirmative evidence that the box reformatted a partition or reopened a listener. Eight have nothing at any tier. Four are live product defects, not test gaps: `pithead-data-reset` reformats a filesystem `e2fsck` can repair (#1062), `pithead backup` stops the stack before it discovers it cannot read a member (#1059), flipping `dashboard.expose_public_ip` re-renders the Caddyfile and never restarts Caddy (#1052), and `dashboard.host` set to the bridge gateway renders two identical site blocks that Caddy refuses. Three of the tests below are red on HEAD until the matching fix lands — that is the correct state, and they ship with the fix.

## Tests that cannot fail

These come first because anyone reading the suite today believes they are covered. Each entry names where it is, why it stays green, and the work item that replaces it.

**1. Wedged-`/data` recovery, tier 1** — `tests/stack/run.sh:10721-10758`. `fsck` is stubbed as `exit 0` and `mount` is a script whose verdict comes from a counter file, so `MOUNT_MODE=repair` means "the second `mount` call succeeds" and nothing more. Flipping `fsck -p` to `fsck -y` (the fix), to `fsck -n` (a worse bug), or deleting the line leaves all four assertions green. The block does honestly prove marker precedence, the fail-safe default, and that `data_mountable` retries after the repair step — keep those. → **W1**

[...]

PR #1697, opened 2026-09-03
test(os): #1694 leg 4's floor row asserts the refusal for the state it plants — a floor above the running version

VijitSingh97:
#1694 — a harness row, `tests/os/run.sh` (this lane's path). Bench proof is posted in the comments (update leg, 33 passed / 0 failed, the row green).

## What changed

Leg 4's floor row plants `.os-data-floor = 99.0.0` on a 1.19.0 slot, downloads the newer published release, and asserted the plain below-the-floor refusal (`strand the chain data`). Since #1393 merged, a floor above the running version is the failed-migrating-update state and `os_update_version_guard` refuses that FIRST — the premise branch fires when the floor is above both the running and the bundle version, which is the only state a newer bundle can plant. The row now asserts that refusal (`failed its gate` and `the floor version or newer installs` in one pattern, which the plain door carries neither of) and its name says which door it drives. The plain door needs a downgrade bundle the leg does not build; `test-appliance-os-update.sh` holds it at tier 1. One `testing-strategy.md` row reworded to match.

Line-neutral by construction: `run.sh` sits at its 3437 ceiling, and the block was rewritten at the same line count; max comment width and max line width are unchanged (107 / 302, measured before and after).

## What was RUN

- `bash -n`, `shfmt -i 4 -d`, `shellcheck --severity=warning` (under the serialize flock) on `run.sh`: clean.
- `make lint-file-budget`, `lint-md`, `lint-docs-voice`: green.
- `tests/stack/test-appliance-boot.sh` standalone, the tier-1 file that asserts `run.sh`'s text (phase wiring, expected-commit pin): 102 passed, 0 failed.

## What was NOT run

- **Bench: RUN, as an update-only leg, not the full battery.** `tests/os/run.sh --phase update` at `bench/1697-proof-stacked` (= `develop-v2` at `ebb0f43d` + this commit; tree identical to `merge-tree ebb0f43d <this head>`): 33 passed / 0 failed, the floor row green with the new text. Log and the six-commit tip drift that landed under it — `dashboard/` changed, nothing baked from `os/`, `tests/os/`, the CLI or the compose file did — are stated in the evidence comment. Not re-run at the current tip; #1651's shipping-tip battery is the authoritative run.
- The pre-change red, verbatim, is in #1682's evidence comment (leg 4, the `verify below the floor did not refuse honestly` row) and in #1694.

## Over-engineering pass, by hand

[...]

PR #1708, opened 2026-09-03
verify-image: refuse an image that ships without e2fsck/mkfs.ext4 (#1069 W11)

VijitSingh97:
## The defect

`pithead-data-reset`'s repair chain (`os/overlay/pithead-data-reset`) runs
`fsck`/`e2fsck`/`mkfs.ext4` behind `|| true`. If the built rootfs ever shipped
without e2fsprogs, the "command not found" from a missing binary would be
swallowed by that `|| true` exactly the same way a real fsck refusal is —
nothing in `tests/os/verify-image.sh` distinguished "the tool is missing"
from "the tool ran and said no". The Dockerfile installs e2fsprogs today, but
nothing on the built artifact caught a regression that dropped it.

## What the gate now refuses

`tests/os/verify-image.sh` adds `data_reset_repair_tools_present <image-root>`,
which fails unless both `e2fsck` and `mkfs.ext4` are present *and executable*
at either the merged-`/usr` or legacy `/sbin` path, and wires it into the
existing `chk` gate (`"data-reset's repair tools are baked (e2fsck + mkfs.ext4,
#1069 W11)"`) alongside the other data-reset invariants. A rootfs missing
either tool — or shipping a non-executable stand-in — now fails verify-image.sh
on the built artifact, not just on the package list.

## How it was proven

The new self-test drives the shipped function directly (source, no copy):

```
bash tests/integration/selftest-verify-image-repair-tools.sh
```

7 passed, 0 failed — both-present at each accepted path (merged `/usr` and
legacy `/sbin`) pass; e2fsprogs missing entirely, either tool alone missing
(both directions), and a present-but-non-executable pair all fail. The
"neither" fixture is the shape #1069's regression describes: no e2fsck
anywhere in the root, which the `|| true` chain in `pithead-data-reset` would
otherwise mask.

Also run clean at the pinned shellcheck (0.11.0) and shfmt on both touched
files, and `make lint-file-budget` (the new `tests/os/verify-image.sh` line
count is recorded).

`tests/integration/selftest-verify-image-repair-tools.sh` sources
`tests/os/verify-image.sh` (confirming its sourced-guard still no-ops) and
calls `data_reset_repair_tools_present` against synthetic root trees — no
image, no loop mount, no root, no bench — asserting: both tools present at
either accepted path passes; the tool set missing entirely fails; either tool
missing alone fails (proving the check isn't satisfied by a single tool); and
a present-but-not-executable pair fails (the binary is invoked directly, not
found via `PATH`).

## Ponytail review

Ran `/ponytail-review` on the diff before opening: cut a self-verifying
assertion from the selftest that checked the fixture setup (`mkdir -p` never
creates an executable) rather than the shipped function — the
"e2fsprogs missing entirely" case right above it already exercises that exact
fixture against `data_reset_repair_tools_present`. Applied in
0546a129, 9 lines removed.

Refs #1069


VijitSingh97:
## Controller pass — FINDINGS (non-author; judged on the fable tier). One rebase, one comment sentence, then merge.

**Re-derived by me:**
- Selftest at this head: **7 passed, 0 failed.** Mutation control: dropping the `mkfs.ext4` half of `data_reset_repair_tools_present` reds exactly `e2fsck present, mkfs.ext4 missing -> FAIL` (6/1), so the pair check is a guard, not a grep. The selftest is picked up by the `tests/integration/selftest*.sh` glob in both `make test-integration-selftest` and CI, so it runs without wiring.
- Budget: `tests/os/verify-image.sh` 400 -> 414 with a new row at exactly 414 — a first appearance at the real count, which the gate accepts; the ruling that a new row reserves no headroom is met. Parity rc 0.
- **BLOCKER (mechanical) — `merge-tree` rc 1 against the tip:** `docs/dev/file-budget.tsv` conflicts. The tip moved other rows since this branch was cut, and the head still carries `tests/os/run.sh` at the stale 3437, which the gate at the head reds as a raise. Rebase; on the rebased tree the only tsv change should be your one new row.
- **Comment sentence in `verify-image.sh` misstates the consequence:** "the box proceeds to treat /data as wedged with no tool present to repair it, only to reformat it." `mkfs.ext4` is e2fsprogs too, so on the rootfs this guards against the reformat fails as well: `data_mountable` swallows the missing `e2fsck` behind `|| true`, `backup_superblocks` returns nothing because `mke2fs -n` is gone, the decision says reformat, and `mkfs.ext4` fails with `mkfs failed on … — leaving it as-is`. The real consequence is a box that boots with /data neither repaired nor reformatted, under a log that reads like a refusal. The "transient mount failure reads exactly like a refusal-class corruption" clause describes the #1062 hazard, which this check does not see either way. Say what actually happens — a false claim in in-repo comment text earns a fix now rather than a discharge commit later. The PR body's own first paragraph has it right.

**Relayed, not re-run:** shellcheck 0.11.0 and shfmt clean.

Merge condition: rebased (`merge-tree` rc 0, gate rc 0 at the head), the sentence corrected, checks terminal and green at the moment of merge. This is #1069 W11 only — #1069 stays OPEN.

VijitSingh97:

Re-derived at source in a detached worktree at PR head `0546a129`: `os/overlay/pithead-data-reset`'s repair chain does run `fsck`/`e2fsck`/`mkfs.ext4` behind `|| true` (confirmed by reading the file), so a missing binary and a real refusal are genuinely indistinguishable pre-fix. `data_reset_repair_tools_present` (`tests/os/verify-image.sh:38-43`) checks both tools, both accepted paths, executable bit — matches the claim.

Named checks, all clean — note I used `~/.local/bin/shellcheck-real` (the CI-pinned 0.11.0 binary) directly rather than the raw `shellcheck` on `$PATH` (apt's 0.9.0) or the serialized `~/.local/bin/shellcheck` wrapper, which was queued behind two other lanes' full-repo shellcheck runs at review time — small, targeted 2-file invocations don't need the wrapper's memory cap:
```
$ bash tests/integration/selftest-verify-image-repair-tools.sh
selftest-verify-image-repair-tools: 7 passed, 0 failed

$ shellcheck-real --severity=warning tests/integration/selftest-verify-image-repair-tools.sh tests/os/verify-image.sh   # rc=0
$ shfmt -d tests/integration/selftest-verify-image-repair-tools.sh tests/os/verify-image.sh                              # rc=0
```

Standing test — negative control: neutered `data_reset_repair_tools_present` to `return 0` unconditionally and reran the self-test:
```
✗ e2fsprogs missing entirely -> FAIL         (want rc=1, got rc=0)
✗ e2fsck present, mkfs.ext4 missing -> FAIL  (want rc=1, got rc=0)
✗ mkfs.ext4 present, e2fsck missing -> FAIL  (want rc=1, got rc=0)
✗ present but not executable -> FAIL         (want rc=1, got rc=0)
selftest-verify-image-repair-tools: 3 passed, 4 failed
```
Exactly the four negative cases reddened, the three positive cases stayed green (a bare `return 0` trivially still satisfies "present"). Restored the file afterward; worktree diff clean. "Could this gate be green if a consumer ignored its exit code" — no: the self-test's own final line is `[ "$IT_FAIL" -eq 0 ] || exit 1`, and `chk` in `verify-image.sh` accumulates into the script's own overall exit status, not a printed-only verdict.

`make lint-file-budget` (`bash scripts/lint-file-budget.sh`) reds on this branch, same as every other PR reviewed this pass: `tests/os/run.sh`'s ceiling 3423→3437, pure base drift from `origin/develop-v2` moving after this branch cut — this PR's own new `tests/os/verify-image.sh 414` row is fine and not implicated. Resolved by rebase.

**#1697 overlap:** none. `#1697`'s diff touches only `docs/dev/testing-strategy.md` and `tests/os/run.sh`; `#1708`'s touches `docs/dev/file-budget.tsv`, `tests/integration/selftest-verify-image-repair-tools.sh`, and `tests/os/verify-image.sh` — zero shared files, and no functional overlap (#1697 is `tests/os/run.sh`'s leg-4 floor-row assertion for OS-update aging; #1708 is a new baked-tool check in `verify-image.sh`). Safe to land independently in either order; both will separately need a rebase past the same `tests/os/run.sh` ceiling drift, but neither PR causes it.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `0546a129`

VijitSingh97:

PR #1709, opened 2026-09-03
fix(tests): stop two verdict lines from varying run to run

VijitSingh97:
## What

The #1105 domain-split PRs prove a mechanical move changed nothing by diffing the sorted
multiset of per-test verdict lines before and after. Two PASS lines broke that proof by
carrying a value that changes on every run while the verdict itself stays PASS:

- The ssh-host-keys test printed `ssh-keygen`'s own random ed25519 fingerprint in its
  passing verdict.
- `load_baked_images`'s heartbeat test printed a real `$(date +%s)`-measured elapsed time
  in its passing verdict (seen as `(1s)` vs `(2s)` across two runs, same PASS both times).

Each split PR had to hand-wave "empty diff modulo these N nondeterministic lines," and N
had grown from 1 to 2. Every such exception is a line the multiset proof can no longer
vouch for — a real drop or reorder hiding behind one of them would go uncaught.

## What changed

`test-appliance-identity-boot.sh`'s PASS line now reports the key's loadability without
the fingerprint bytes (the fingerprint itself is unchanged and still drives the later
byte-identity assertions). `test-appliance-boot.sh`'s PASS line now asserts the heartbeat
returned inside the bound (`hlen -lt 3`) without printing the measured `$hlen` in the
verdict text; the measured value still reaches the failure-detail line if the bound trips.
Both PASS lines are now the same string on every run.

A new domain, `test-harness-tooling.sh`, drives each fixed domain file for real — a fresh
random key, a fresh timing — and asserts the captured PASS text carries neither a
`SHA256:` fingerprint nor a trailing `(<digit>s)` measurement.

## Proof

Confirmed the added test is a guard, not a grep, by running it against the pre-fix
domain files (`git show origin/develop-v2:tests/stack/test-appliance-{boot,identity-boot}.sh`
swapped in temporarily) with `$HERE`/`$ROOT` set the way `tests/stack/run.sh` sets them:

```
✗ the ed25519-key PASS line carries no fingerprint bytes
    [... a loadable ed25519 key (SHA256:...)] unexpectedly contains [SHA256:]
✗ the heartbeat PASS line carries no measured elapsed value
```

Same run against the fixed files on this branch: both assertions pass. Also ran
`shellcheck` and `shfmt -d` on all three touched files — no new findings on the changed
lines (only pre-existing, file-wide `SC2015`/`SC1091`/`SC2034` info notes untouched by
this diff).

`test-harness-tooling.sh` sources `test-appliance-identity-boot.sh` and
`test-appliance-boot.sh` each in their own subshell (the same sourcing pattern
`tests/stack/run.sh` uses), greps out the specific PASS line each domain produces, and
asserts: no `SHA256:` substring in the key line, no `(<digit>+s)` suffix in the heartbeat
line.

Refs #1325


VijitSingh97:
## Controller pass — FINDINGS (non-author; judged on the fable tier). One blocker, one test defect, one rebase.

**Re-derived by me:**
- The two PASS-line edits are 1:1 replacements and the new guard is a guard: I ran the two new `test-harness-tooling.sh` blocks against the tip's pre-fix domain files and got **2 FAIL** (`SHA256:` and `(1s)` both caught), against this head **2 PASS**. Sourcing both domains a second time costs ~7 s, so the cost is fine.
- **BLOCKER — the budget gate reds at the head AND at the merged tree** (`make lint-file-budget`): `tests/stack/test-appliance-boot.sh` is 645 over its 641 ceiling, `tests/stack/test-appliance-identity-boot.sh` 469 over 465. Both rows sit at ZERO headroom on the tip and cannot be raised (ceilings only go down). The four added comment lines in each domain file are what push them over. Fix: drop both four-line comment blocks — the explanation already lives in the harness-tooling block's own header, the file with headroom — or cut existing lines to net zero. The body does not say the gate was run; it would have shown.
- **Test defect — both new assertions pass on an EMPTY capture.** I ran the same two blocks with `$HERE` pointed at a directory whose two domain files are empty: **2 PASS in 0 s.** `assert_not_contains "" "SHA256:"` and the `case` fall-through both read silence as clean, so a domain that dies before its PASS line (a missing tool, an unbound variable under `set -u`, a renamed label) turns this guard green. Add an arming readback before each: assert the captured line contains its label (`loadable ed25519 key`; `does not wait on the heartbeat interval`). Cheapest fix in the PR, and the one that makes it evidence.
- Rebase: at the head the gate also reds `tests/os/run.sh` 3423 -> 3437 — base staleness (the tip lowered that row after this branch was cut); it does not appear on the merged-tree leg and clears on rebase. Not this PR's defect.
- Parity rc 0 at both legs; `merge-tree` rc 0.

**Relayed, not re-run:** shellcheck and shfmt clean.

Merge condition: gate rc 0 at the head (comment blocks gone), arming readbacks in, rebased, checks terminal and green at the moment of merge. #1325 is closed by hand on merge — `Closes` does not fire on `develop-v2`.

VijitSingh97:
**Non-author reviewer pass — RETURN WITH FINDINGS**

A controller pass already posted findings on this head (`60edbe13`, comment 2026-09-03T18:53:42Z) — this comment independently re-derives both from scratch rather than relaying them; both hold.

**Confirmed — budget-gate blocker.** From the repo root, `bash scripts/lint-file-budget.sh`:
```
file-budget: FAIL — tests/stack/test-appliance-boot.sh is 645 lines, over its recorded ceiling of 641.
file-budget: FAIL — tests/stack/test-appliance-identity-boot.sh is 469 lines, over its recorded ceiling of 465.
```
Both rows sit at zero headroom; the four-line explanatory comment this PR adds to each domain file is what pushes them over. The body doesn't say this gate was run — it would have shown.

**Confirmed — the new assertions pass on an empty capture.** Stubbed `$HERE` with two empty domain files and ran the PR's own capture pattern verbatim:
```
vd_id_out=[]  → assert_not_contains "" "SHA256:" → PASS
vd_hb_out=[]  → case falls through              → PASS
```
A domain that dies before its PASS line (missing tool, unbound var under `set -u`, a renamed label) reads as clean. Neither assertion has an arming readback confirming the captured line is the one it means to check.

**Positive control, corrected.** My first attempt swapped in `origin/develop-v2`'s current copy of the domain files and got an unrelated `STACK_SUITE: is unset` guard error — that tip has since gained a guard this PR's branch predates, an artifact of branch drift, not a finding. Re-ran against `HEAD~1` (this PR's own immediate parent) instead: the pre-fix capture correctly contains `SHA256:W4BMU...` and `(1s)`, and the PR's assertions correctly FAIL on that real content — so the guard is genuine when armed, just not armed against silence.

**Also confirmed:** the budget gate's third line (`tests/os/run.sh` 3423→3437) is the same base-drift already reported on #1681/#1701/#1704 — `origin/develop-v2` lowered that ceiling after this branch cut; not this PR's defect, clears on rebase. `shellcheck --severity=warning` on the three touched files: 3 pre-existing `SC2034` (confirmed present, same count, on `HEAD~1`'s unmodified copy — not new). `shfmt -d`: clean.

"Could this gate be green if a consumer ignored its exit code" — moot until the above are fixed: the gate is currently red at the head regardless of how a consumer reads `$?`.

Agree with the controller's merge condition: budget gate rc 0 at the head (drop or net-zero the comment blocks — the harness-tooling file has headroom, the two domain files don't), arming readbacks added to both new assertions, rebased.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `60edbe13`

VijitSingh97:
**Author fix-up — both findings addressed; ready for a fresh non-author pass. Not merged.**

Head is now `53bb5c1501cc997a84d74bf17198d6f69fe407f4` (a merge commit: parents `60edbe13` = the reviewed head, `8e248b1e` = `origin/develop-v2` at fetch time), pushed to `origin/fix/1325-suite-verdict-nondeterministic`.

**Findings addressed**
1. Budget-gate blocker — the four-line comment blocks are gone from both domain files; the explanation lives only in the harness-tooling block's header (the file with headroom). Both files are back at their ceilings.
2. Empty-capture defect — an arming row now precedes each check and asserts the captured line carries its own label (`the generated key is a loadable ed25519 key`; `a fast load does not wait on the heartbeat interval`). The capture keeps only the `✓` line, so a domain that dies before its PASS line, a renamed label, or a FAIL on that row now reads as a FAIL of the arming row, not as clean.
3. Rebase — the merge from `origin/develop-v2` clears the `tests/os/run.sh` 3423→3437 base-drift line.

**Gate proof** — `cd /Users/vijit/code/wf-1325 && bash scripts/lint-file-budget.sh`:
```
```
rc=0. Before the fix, on the same merged tree, it reported 645/641 and 469/465.
`wc -l` → 641 `tests/stack/test-appliance-boot.sh`, 465 `tests/stack/test-appliance-identity-boot.sh`.

**Guard proof** — a driver sources `<HERE>/lib.sh` then evals the #1325 block verbatim (`sed -n '/^echo "== unit: verdict-line determinism/,$p'` from the fixed `test-harness-tooling.sh`), run against three `HERE` trees (rsync copies of the repo minus `.git`):

1. Fixed tree (`HERE=/Users/vijit/code/wf-1325/tests/stack`) → **PASS=4 FAIL=0**
   ```
   ✓ the ed25519-key PASS line was captured at all (arming)
   ✓ the ed25519-key PASS line carries no fingerprint bytes
   ✓ the heartbeat PASS line was captured at all (arming)
   ✓ the heartbeat PASS line carries no measured elapsed value
   ```
2. Pre-fix domain files (the two PASS lines reverted to `($shk_fp1)` / `(${hlen}s)`, everything else identical) → **PASS=2 FAIL=2** — the defect case is caught:
   ```
   ✓ the ed25519-key PASS line was captured at all (arming)
   ✗ the ed25519-key PASS line carries no fingerprint bytes
       [✓ the generated key is a loadable ed25519 key (SHA256:StJEVIICO3i9l7Hn0M3tEjKt40+/OtDlksLDWgBDitg)] unexpectedly contains [SHA256:]
   ✓ the heartbeat PASS line was captured at all (arming)
   ✗ the heartbeat PASS line carries no measured elapsed value
       ✓ a fast load does not wait on the heartbeat interval (1s)
   ```
3. Empty domain files (the reviewer's silence case) → **PASS=2 FAIL=2** — the arming rows fail instead of reading silence as clean:
   ```
   ✗ the ed25519-key PASS line was captured at all (arming)
       [] missing [the generated key is a loadable ed25519 key]
   ✓ the ed25519-key PASS line carries no fingerprint bytes
   ✗ the heartbeat PASS line was captured at all (arming)
       [] missing [a fast load does not wait on the heartbeat interval]
   ✓ the heartbeat PASS line carries no measured elapsed value
   ```

Not run here: the whole `make lint` / `make test`, and shellcheck on `tests/stack/run.sh`.

— author fix-up, `p2pool-starter-stack/pithead` @ `53bb5c15`

VijitSingh97:
**Non-author reviewer pass at `53bb5c15` — PASS. Two non-blocking findings.** My empty-capture finding from the last cycle is **discharged by measurement**, not by reading the new arming rows.

## Ruling 8a — budget, run by me, re-derived after the base moved

`develop-v2` advanced to `0e2b291c` while I worked (it now carries #1727), so I re-measured against that tip rather than quoting the earlier one:

```
branch head 53bb5c15                          file budget OK — ... monotonic.   rc=0
merged tree (0e2b291c x 53bb5c15)             file budget OK — ... monotonic.   rc=0
merge-tree(develop-v2, 53bb5c15)              rc=0 — clean
```

Worth naming because it is tight: both touched domains sit **exactly at** their ceilings (`test-appliance-boot.sh` 641/641, `test-appliance-identity-boot.sh` 465/465) and this diff is net-zero on each (+1/-1). `test-harness-tooling.sh` is 207 lines, under the 400 target, correctly unbudgeted. Any future net +1 in either domain reds the gate.

## The empty-capture finding is DISCHARGED — four armed stub legs, two per assertion

I did not take the arming rows on their face; I forced the capture empty two different ways for each, and asserted each mutation landed by content first:

| leg | how the capture was emptied | arming row |
|---|---|---|
| ed25519 | PASS label renamed so the grep misses | **✗ fires** |
| ed25519 | domain returns *before* the PASS line | **✗ fires** |
| heartbeat | PASS label replaced (`does not wait on the heartbeat interval` → `is prompt`) | **✗ fires** |
| heartbeat | domain returns *before* the PASS line | **✗ fires** |

In **every** leg the `not_contains` / `case` row still reads ✓ — vacuously, on an empty string — which is exactly the failure I named last cycle, and the arming row is what turns the suite red anyway. The design is right and it works. Baseline at the head: all four rows ✓, tree restored to 0 modified files after each leg.

**Disclosure on my own instrument.** My first heartbeat leg was a *bad experiment*: I renamed the label to `…intervalX`, which still **contains** `interval`, so the grep matched, nothing was emptied, and the leg reported a false clean. I caught it only because it disagreed with the two legs that did fire. The table above is the redone version.

## The headline claim — determinism — verified, with a control that fired

Fixed head, two independent runs of each touched domain, sorted verdict lines diffed. This is **wider than the PR claims**: it covers every verdict line in both domains, not just the two that were fixed.

```
test-appliance-identity-boot.sh   84 verdict lines   differing across runs: 0
test-appliance-boot.sh           102 verdict lines   differing across runs: 0
```

Positive control — the same instrument against the **pre-fix** files from the merge base, arming confirmed by content (`fingerprint-in-PASS: 1`, `hlen-in-PASS: 1`):

```
PRE-FIX test-appliance-identity-boot.sh   differing across two runs: 2
  <   ✓ the generated key is a loadable ed25519 key (SHA256:Gui/39nY4FYu…)
  >   ✓ the generated key is a loadable ed25519 key (SHA256:S84GirFQFRr…)
```

**Honest limit, and it argues for the fix rather than against it:** the pre-fix *heartbeat* leg showed **0 differing** in my two samples — both printed `(1s)`. I checked the captured text rather than inferring: the measured value **is** embedded; my two samples simply landed on the same second. An intermittent nondeterminism is the worse kind for a multiset proof, because it passes most runs and breaks it unpredictably.

## Both values are still load-bearing where they belong

- `shk_fp1` survives the PASS line losing it and still drives `test-appliance-identity-boot.sh:403` — `assert_eq "second run leaves the key byte-identical" … "$shk_fp1"`.
- `${hlen}s` still reaches the failure detail at `test-appliance-boot.sh:208` — `bad "a fast load returns promptly" "took ${hlen}s"`.

## No leak, no counter inflation, and the `bash -c` choice avoids a known trap

A full run of the domain leaves `/tmp` at a delta of **−1** — no sandbox leak. And the nested domains do not enter the parent's tally: `test-harness-tooling.sh` produces **23** verdict lines, 0 FAIL, not the ~190 it would if identity-boot's 84 and boot's 102 leaked in. Sourcing them in a `bash -c` **subshell process** is the reason, and it also sidesteps the `_d0` clobber that nesting a domain stanza inside another domain would otherwise cause. Deliberate or not, it is the right shape.

## F1 — cost, worth one line in the body (non-blocking)

The new rows re-run two full domain files inside a third, so the suite executes those two domains twice. Measured on this box: `test-harness-tooling.sh` goes **4s → 11s**. That is the price of a *real* run rather than a canned fixture, which is what makes the guard honest — a fair trade, but the body should say it rather than let someone rediscover it in a slow-suite hunt.

## F2 — scope: this removes the two known exceptions and adds nothing that catches a THIRD

The body's own framing is that N grew 1 → 2. It grew by **discovery**, not by mechanism, and this PR pins exactly the two that were found. A new nondeterministic PASS line anywhere else in the suite still arrives as a hand-waved exception in the next split PR, and nothing is looking for it.

The general guard is precisely the instrument I ran above and it is cheap: run a domain twice, diff the sorted verdict multiset, fail on any difference. Say the word and I will file it as a follow-up issue — out of scope for this PR, and I am not asking you to widen this one.

## What I did NOT do

I ran the two touched domains and `test-harness-tooling.sh` standalone through a driver that mimics `run.sh`'s source order — **I never ran `tests/stack/run.sh` in full**, so the whole-suite multiset claim is the author's, not mine. I took no lint lock and ran no `make` target; `shellcheck`/`shfmt` stand as yours. Two samples is what my determinism control had — enough to catch the fingerprint, not enough to have caught the heartbeat, as stated above.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `53bb5c15`

VijitSingh97:
**Controller — MERGING on the non-author reviewer PASS at `53bb5c15` (fable tier).** Re-derived by me: `merge-tree --write-tree` rc 0 against the tip that now carries #1727; GitHub's closing references for this PR read `[]`, so #1325 closes by hand. Relayed from the reviewer's pass, whose method I read rather than re-ran: the budget gate rc 0 at both the branch head and the merged tree against that same tip (ruling 8a, both legs, and both touched domains sit AT their ceilings at net zero), the four armed stub legs discharging my empty-capture blocker, and the two-run determinism control that fired on the pre-fix fingerprint line. Merge-record note, not a defect in the diff: the body does not state the suite cost the reviewer measured (`test-harness-tooling.sh` 4s -> 11s, because the two domains now run twice); it is recorded here instead. The reviewer's F2 (a general two-run verdict-multiset guard) is out of this PR's scope and is being filed as its own issue.

PR #1710, opened 2026-09-03
fix(dashboard): #1564 Worker Inspect reports the config change nothing recorded

### LINKED ISSUE #1471: tests(integration): the rig-apply leg settles on config.json and asserts before RigForge writes status.json — wait for the history ROW, not `applied|accepted`
## Summary

After an **asynchronous** worker-apply converges on the rig, the dashboard's per-worker history row keeps the **dial-time `accepted`** it was written with. Nothing reconciles it to the terminal status. The `#1236` DONATION leg asserts that row is strictly `applied` and therefore reds on every async apply, even though the rig demonstrably applied the change.

Found on the first hardware run in which the `#1236` legs actually executed (the #1454 fix landed; see also #1466, which had been silently skipping the whole phase).

## Evidence — from one run, with a built-in discriminator

```
✓ DONATION edit applied on the rig (#1236)
✓ the rig's own config confirms DONATION changed (#1236)
✗ DONATION worker-apply recorded in the per-worker history (#185/#1236)
    expected [applied], got [accepted]
✓ DONATION edit reverted on the rig (#1236)
```

In the **same run**, the two synchronous keys passed the identical assertion:

| key | dial outcome | history-row assertion |
|---|---|---|
| `max_temp_c` (#513) | terminal immediately | ✓ |
| `watchdog_interval_min` (#1236) | terminal immediately | ✓ |
| **`DONATION`** | **went async — entered the 90s settle wait** | **✗ `got [accepted]`** |

So this is not DONATION-specific in substance: it is **async-apply-specific**. Any key whose apply takes the RigForge #344 async path will red this assertion.

## Why the harness's own status is right and the row is stale

`_settle_worker_apply` promotes its status to `applied` only after **verifying the rig's own reported config** (`rigforge-apply-settle.sh`). That promotion is local to the harness. The dashboard's `#185` history row was written at dial time with `accepted` and is never revisited once the rig converges.

## There is already a precedent for this exact call

The `#517` auto-rollback leg hit the same seam and resolved it deliberately (`run.sh:2404-2408`):

```
case "$histstatus" in
rolled_back | accepted) it_pass "the dashboard's per-worker history records the change (#517: $histstatus)" ;;
```

with the reasoning that the rig-side terminal is asserted separately, and *"accept both rather than force a product change to the runner's poll budget."* The `#1236` round trip does not carry that accommodation.

## Two ways to resolve, and they are not equivalent

1. **Harness (consistent with #517):** accept `applied|accepted` in `_writable_key_round_trip`'s history assertion, since the legs on either side of it already assert the rig-side terminal. Cheap, and matches the established precedent in the same file.
2. **Product:** have the dashboard reconcile the history row when the change reaches a terminal status. This is the question option 1 quietly declines to ask, and it is a real one — a user reading the history sees `accepted` forever for a change that applied.

**Recommendation: do 1 to unblock the gate, and decide 2 on its own merits rather than as a side effect.** Filing them together so option 2 is not lost when option 1 makes the red go away.

## Note for whoever grades this leg

The rig-side terminal is asserted by the two legs surrounding this one; both passed. The change genuinely applied and genuinely reverted — the rig's own reported config confirmed both.

### LINKED ISSUE #1564: Worker Inspect still reads "Last changed from this dashboard" over a config a rig-drift row already flagged
## What is shipped

#1551 is closed: a rig serving a config whose change nothing recorded is now detected on every poll and written to the durable audit trail as a `rig-drift` row (#1559 the store, #1562 the wiring).

## The gap this leaves

That detection lands in the Security panel only. The Worker Inspect provenance line is unchanged, and it is the surface #1551's own opening paragraph named:

> The Worker Inspect provenance line keeps reading "Last changed from this dashboard".

It still does. `ConfigProvenance` renders two things and neither can see a revision move:

- `configOriginNote(origin, meta)` reads `detail.config_origin`, which is the rig's own account via `config_meta`. A hand-edit underneath RigForge stamps no new `last_change_id`, so a rig this dashboard has applied to before keeps resolving to `here` — the calm "Last changed from this dashboard".
- `configDriftNote(drift)` reads `detail.config_drift`, which is #1367's per-key comparison against `last_applied`. That is the mechanism #1551 exists because it is blind here: a key we never set has no left-hand side, so this list is empty in exactly this case.

Confirmed at `ef94a4ea`: the only consumer of `note_worker_revision` is `note_revision_drift`, and it writes an audit row. Nothing feeds the Inspect payload.

So an operator on Worker Inspect can read a reassurance while the Security panel holds a `rig-drift` row for the same worker.

## Framing, so this is not over-prioritised

**This is an enhancement to the #1345/#1367 provenance line, not a residual defect of #1551.** #1551 specified Option A — persist the revision, compare on each poll, record the move — and Option A shipped end to end with its tests and docs. The audit trail is also where both sibling detections land: `host-edit` and `rig-edit` have never surfaced on this line either.

It is filed rather than left implicit for the same reason #1551 was itself split out of #1367: so a known gap is not closed along with the case that was fixed.

## Shape, if taken

Serve the last `rig-drift` observation for the worker in the Inspect payload and let it qualify the origin line, the way `configDriftNote` already qualifies it for the per-key case. The revision is a digest over the whole writable config, so the line can say only THAT the config moved with nothing recording it, never which key — the same limitation the `rig-drift` audit row already states.

Worth deciding first, and the reason this is a question rather than a task: `configDriftNote`'s existing block deliberately renders nothing on `null` and nothing on `[]`, because "an all-clear here would be a reassurance bounded by narrowings the operator cannot see from a badge". Adding a third verdict to that line has to respect the same rule, or it reintroduces the defect it is meant to remove.

VijitSingh97:
Closes #1564.

Worker Inspect's provenance line could read "Last changed from this dashboard" over a config the
Security panel had already flagged as a `rig-drift`. #1551 detects the unrecorded edit and writes
it to the audit trail; nothing carried it to the line it contradicts. This does.

The design was decided and posted before any code (issue comment `5528424303`), and the budget call
was ruled by the controller (`5528467402`). Both were followed as written; nothing below re-derives
them.

## What it does

`worker_config_revision` grows one column, `drift_from`, maintained inside `note_worker_revision`'s
existing read-then-write under the one `_db_lock` — the comparison already happens there, so no new
read and no new lock scope. `build_worker_detail` serves it as `config_revision_drift`, and
`ConfigProvenance` renders a third line beside the other two.

Two rules do the work:

- **Two states, never three.** The payload is a drift or `None`. There is deliberately no "checked
  and agrees" verdict, for the reason `configDriftNote` withholds its `[]`: an all-clear here would
  be a reassurance bounded by narrowings the operator cannot see from a badge. Having no third
  state is a stronger guarantee than suppressing one — there is nothing left to relax.
- **Gated on currency, not existence.** The note renders only while the drift we recorded is the one
  that produced the revision the rig is serving on this poll. A `rig-drift` audit row is permanent;
  "has this worker ever drifted?" would accuse a rig forever, including after the operator adopted
  the change. Once the rig moves on, the row stays in the Security panel and the line goes quiet.

**Not joined on the audit `event_id`.** That was the cheap route and it is rejected in the code with
its reason: #1698 (merged as the base of this branch) changed how that id is built, the format is
not a contract, and a lookup built on one fails *silently* — the note would stop rendering and every
test written against the old shape would stay green.

## File budget

- `storage_service.py` **1357 -> 1358, its ceiling, ZERO headroom** — as ruled. One line: the
  migration call. The column itself goes inside the existing `CREATE TABLE` string (0 lines) and the
  migration body lives in `worker_config_store.py`, the mixin that owns every accessor of this table
  (#1369). **That row cannot be raised — `check_monotonic` refuses any rise — so the next addition
  to that file SPLITS it.**
- `worker_config_store.py` 319 -> **399/400**, no row needed, one line spare. Recorded here so the
  next session sees it before planning.
- `test_storage_schema.py` 361 -> 394, `test_worker_revision.py` 131 -> 244,
  `test_worker_config_drift.py` 238 -> 309, `workerlogic.test.mjs` 370 -> 395 (five spare),
  `workerlogic.mjs` 331 -> 353, `confighistory.mjs` 71 -> 76, `worker_detail.py` 285 -> 294.
- `workerview.mjs` stays 443/443 (a one-line in-place edit to pass the prop).

Two docstring lines were deleted to stay under 400 in the mixin; both restated design reasoning that
lives permanently in the issue comment above. No comment-as-control and no test was shaved.

## What was run

- `make lint` — **rc 2 on `lint-sh` only**, and it is not this change: shellcheck here is 0.9.0
  against the 0.11.0 pin #1679 landed. This diff touches no shell. Every other target was then run
  individually (`make` stops at the first failure): `lint-py`, `lint-js`, `lint-yaml`, `lint-md`,
  `lint-docs-voice`, `lint-operator-strings`, `lint-topology`, `lint-file-budget`,
  `lint-pithead-parity`, `lint-trivy-parity`, `lint-proto`, `lint-toml` — **all rc 0.**
- `make test-dashboard` — **2475 passed, rc 0**, total coverage 97.38% (gate 80%). `make test`
  itself aborts at its first prerequisite, `lint-sh`, for the reason above, so the suite was run
  directly rather than reported through a target that never reached it.
- **Patch coverage 29/29 executable added Python lines = 100%** (bar 90%), computed mechanically
  from the diff hunks against a `coverage.xml` regenerated on the tree shipped here — not from
  CI's rc, which is no evidence on a `develop-v2` PR (#1557). The first run of that came back
  "no lines matched": the path prefix was wrong, since `<source>` is the package dir. Reported
  only after the prefix was fixed and the match count was non-zero.
- Base moved twice mid-change (`d1c32cb6` -> `8ab235e5` -> `734a889d`). `git rev-parse
  <base>:dashboard` is **byte-identical** at the last two, so the program the suite ran did not
  change and no re-run was owed; rebased onto `734a889d` so the review is the narrow case.
  `lint-file-budget` re-run against #1699's **new** gate script: rc 0.
- `node --test dashboard/tests/frontend/` — **508 pass, 0 fail.**
- **Mutation battery, 6 seeded, 6 killed**, each restored and verified by sha256 (baseline green
  first, so a kill means something):

  | mutation | killed by |
  |---|---|
  | preservation dropped (`drift_from` not carried forward) | `test_the_note_survives_the_polls_after_the_one_that_raised_it` |
  | resolved case ignored (moved with a new change id still drifts) | `test_a_move_with_a_new_change_id_clears_it` |
  | currency gate removed | `test_it_is_silent_once_the_rig_serves_a_different_revision` |
  | empty `drift_from` rendered | 5 tests |
  | migration `ALTER` removed | `test_..._drift_from_added_on_upgrade` |
  | migration never called | `test_..._drift_from_added_on_upgrade` |

  Four of the six are single kills — one test per design decision, measured rather than asserted.
- Two frontend mutations (`configRevisionDriftNote` always silent; the component drops the line) —
  both killed, 4 and 2 tests respectively.

## Security pass, by hand

`revision` is rig-supplied off the unauthenticated worker feed and now reaches SQLite and the
browser. Re-derived at source rather than assumed: `parse_config_meta` bounds it to 64 chars against
`^[A-Za-z0-9._-]+$` (`rig_config_meta.py:63,73,78`), so no markup can reach the value, and preact
escapes it regardless. Both sides of the new comparison go through that same validator on the same
poll — the write via `worker_change_audit.note_revision_drift`, the read via `xmrig_client.py:108`
— so the equality is between two normalized tokens, not two parses. No new outbound fetch, no new
credential path, and the pool-password strip is untouched.

## Over-engineering pass

Done by hand on merits, not as a ceremony; the gate here verifies nothing. What it weighed:

- **A separate table for the drift** — rejected. One nullable column on the row that already
  records the revision costs no new read, no new lock scope and no new migration path beyond the
  `ALTER`; a table would need its own lifecycle and its own answer to "when is a row stale".
- **Doing the currency comparison in `build_worker_detail`** instead of passing the rig's revision
  into the accessor — rejected. It would put the rule in the layer that cannot test it at the
  service tier and add lines to a call site, to save a parameter.
- **Inlining `_next_drift_from`** into `note_worker_revision` — rejected. It is the whole of the
  currency rule and it is what four of the six mutations aim at; a named function is what let
  each be killed by exactly one test.
- **Reusing the row `note_worker_revision` already read** rather than a second `SELECT` in the
  reader — rejected as not available: those run at different times (a poll, and a page render).
- **The redundant `worker` key** in the returned dict is kept deliberately: it matches
  `note_worker_revision`'s existing return shape, and one consistent field beats two shapes.

## What I did NOT do

- **No `security-reviewer` subagent.** This session's harness instructions forbid spawning agents
  unasked, so the pass above was done by hand at source and is stated as mine, not a reviewer's.
- **Did not establish whether the rig's revision digest spans the stripped pool password.** That
  needs RigForge's source. Because of it I *scoped* an existing unqualified sentence in
  `docs/dashboard.md` ("Nothing on this side can see one either way" -> "This comparison cannot see
  one either way") rather than extending the claim to the new check. Nothing here asserts either way.
- **Did not close #1471 or touch its harness half** — that is `lane-e2e-01`'s.
- No new budget row anywhere, and no ceiling raised or lowered.
- The one-poll window is named rather than closed: if the Inspect payload is built from a poll
  older than the store's newest write, a fresh drift renders one poll late. That is the quiet
  direction — never a false accusation — and matches `note_worker_revision`'s documented posture.

## Docs

`docs/dashboard.md` gains the new line and — because this change falsified prose already there —
two corrections: "A second check" became "Two further checks", and the first check's "keys we have
never applied stay invisible" bound now points at the check that covers them. The `rig-drift` audit
entry gains a cross-link saying which half is durable.

### DIFF
--- a/dashboard/mining_dashboard/service/storage_service.py
+++ b/dashboard/mining_dashboard/service/storage_service.py
@@ -417,14 +417,14 @@ class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin):
             "(id TEXT PRIMARY KEY, ts TEXT, source TEXT, actor TEXT, action TEXT, status TEXT, "
             "keys TEXT)"
         )
-        # The config revision each rig was last OBSERVED serving (#1551), one row per worker.
-        # Additive, mirroring events / share_stats: no _migrate_db change, and the PRIMARY KEY is
-        # the only index it wants. Holding the rig's opaque `revision` NEXT TO the `last_change_id`
-        # beside it is the point: a later poll tells a recorded change (both moved) from an edit
-        # underneath RigForge (only the revision moved) — the door #1542 leaves open.
+        # The config revision each rig was last OBSERVED serving (#1551), one row per worker, plus
+        # the revision it drifted FROM while that drift is still current (#1564, migrated below).
+        # The PRIMARY KEY is the only index it wants. Holding the rig's opaque `revision` NEXT TO
+        # the `last_change_id` beside it is the point: a later poll tells a recorded change (both
+        # moved) from an edit underneath RigForge (only the revision moved) — #1542's open door.
         self._conn.execute(
-            "CREATE TABLE IF NOT EXISTS worker_config_revision "
-            "(worker TEXT PRIMARY KEY, revision TEXT, last_change_id TEXT, ts REAL)"
+            "CREATE TABLE IF NOT EXISTS worker_config_revision (worker TEXT PRIMARY KEY, "
+            "revision TEXT, last_change_id TEXT, ts REAL, drift_from TEXT)"
         )
 
     def _create_indexes(self):
@@ -488,6 +488,7 @@ class StateManager(TelemetryStoreMixin, WorkerConfigStoreMixin):
         if "type" not in {info[1] for info in cursor.fetchall()}:
             self.logger.info("Migrating DB: Adding type column to worker_config")
             self._conn.execute("ALTER TABLE worker_config ADD COLUMN type TEXT DEFAULT 'apply'")
+        self._migrate_worker_config_revision(cursor)
 
     def load(self) -> None:
         """
--- a/dashboard/mining_dashboard/service/worker_config_store.py
+++ b/dashboard/mining_dashboard/service/worker_config_store.py
@@ -74,6 +74,27 @@ def _shaped(row: sqlite3.Row) -> dict:
     return d
 
 
+def _next_drift_from(
+    row: sqlite3.Row | None, revision: str, last_change_id: str | None
+) -> str | None:
+    """The ``drift_from`` to store on this poll (#1564), given the row read on the same poll.
+
+    Split out of ``note_worker_revision`` because it IS the currency rule and is worth settling
+    without a database: the Inspect note must survive the polls AFTER the one that raised it — a
+    drift the operator has not resolved is still true — and go quiet once the config is
+    accounted for again. ``None`` means say nothing, for two reasons: a first sighting has
+    nothing to compare, and a revision that moved WITH a new ``last_change_id`` is the RESOLVED
+    case — something recorded that change, so the line has nothing left to contradict.
+    """
+    if row is None:
+        return None
+    if row["revision"] == revision:
+        # The rig has not moved, so a drift already recorded against this revision is still the
+        # current one. INSERT OR REPLACE rewrites every column, so carrying it is not a no-op.
+        return row["drift_from"]
+    return row["revision"] if row["last_change_id"] == last_change_id else None
+
+
 class WorkerConfigStoreMixin:
     """The `worker_config` accessors of `StateManager`. Never instantiated on its own."""
 
@@ -265,6 +286,10 @@ class WorkerConfigStoreMixin:
         ``last_change_id`` beside it: a hand-edit underneath RigForge, which moves the revision and
         stamps nothing, so neither #1345's provenance line nor #1367's ``config_drift`` can see it.
 
+        The same write maintains ``drift_from`` for ``get_worker_revision_drift`` (#1564) — the
+        read and the comparison are already here, so the Inspect line costs no second query and no
+        second lock scope. ``_next_drift_from`` carries that rule.
+
         It fails CLOSED like ``get_worker_config_change``: any read/write error returns ``None`` and
         accuses nobody. A missed detection is a poll's delay — the next poll compares against the
         same stored row, because a failed write leaves it unchanged — whereas a false accusation is
@@ -280,14 +305,16 @@ class WorkerConfigStoreMixin:
                     return None
                 cursor = self._conn.cursor()
                 cursor.execute(
-                    "SELECT revision, last_change_id FROM worker_config_revision WHERE worker = ?",
+                    "SELECT revision, last_change_id, drift_from FROM worker_config_revision "
+                    "WHERE worker = ?",
                     (worker,),
                 )
                 row = cursor.fetchone()
+                drift_from = _next_drift_from(row, revision, last_change_id)
                 self._conn.execute(
                     "INSERT OR REPLACE INTO worker_config_revision "
-                    "(worker, revision, last_change_id, ts) VALUES (?, ?, ?, ?)",
-                    (worker, revision, last_change_id, time.time()),
+                    "(worker, revision, last_change_id, ts, drift_from) VALUES (?, ?, ?, ?, ?)",
+                    (worker, revision, last_change_id, time.time(), drift_from),
                 )
                 self._conn.commit()
         except sqlite3.Error as e:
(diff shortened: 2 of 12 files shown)


VijitSingh97:
## Controller pass — PASS at `ce191aa4` (non-author; judged on the fable tier)

**Re-derived by me, on this box, in a throwaway worktree at the head:**
- `make test-dashboard`: **2475 passed, rc 0**, total 97%. `node --test dashboard/tests/frontend/`: **508 pass, 0 fail.**
- Patch coverage from the `coverage.xml` that run produced: **29/29 changed lines = 100%** (`storage_service.py`, `worker_config_store.py`), all 3 changed Python files present in the report — not the vacuous form.
- Ruling 8a, both legs vs the tip `ebb0f43d` (merge-base `734a889d`, so this is the OFF-tip case and the two runs are two answers): **merged tree budget rc 0, parity rc 0.** At the branch head the gate reds on exactly one line, `tests/os/run.sh` 3423 -> 3437 — the row the tip lowered after this branch was cut, a file this diff never touches (ruling 8's shape). The three-way merge takes the tip's value, which the merged leg proves. Merged on my own merged-tree run, not on the branch's budget claim. `merge-tree` rc 0.
- The PR's own budget claims hold at the head: `storage_service.py` 1358/1358 — zero headroom, **the next addition to that file SPLITS it** — and `worker_config_store.py` 399/400.
- The security claim, at source: `rig_config_meta.py` bounds `revision` to 64 chars against `^[A-Za-z0-9._-]+$` (`_MAX_TOKEN_LEN`, `_TOKEN_RE`, `_token`), and both the write path (`worker_change_audit.py:94`) and the read path (`xmrig_client.py:108`) go through `parse_config_meta`, so the comparison is between two normalised tokens.
- Design, read in full: `_next_drift_from` carries a drift across `INSERT OR REPLACE` while the revision stands, names the previous revision on a same-change-id move, and clears on a new change id; the read is gated on the revision the rig serves on THIS poll; migration runs after `_create_tables` and before `_create_indexes`, so a pre-#1564 database gains the column and keeps its row (the upgrade test proves it is not read as a first sighting). Two states, never three, in the store, the payload and the component.

**Relayed, not re-run:** the 6+2 mutation battery and its per-mutation kill table; the by-hand security and over-engineering passes (I checked the claims they rest on, not the passes themselves).

**Recorded:** the lane's `make lint` rc 2 on `lint-sh` is this box's distro shellcheck 0.9.0 against the 0.11.0 pin, not this diff — CI's shellcheck at the pinned version is one of the checks read at the moment of merge.

Checks re-read at the moment of merge: none non-terminal, none failed. **#1564 closed by hand on merge** — `Closes` does not fire on `develop-v2`.

Issue #1705, opened 2026-09-03
`tests/stack` domain files: an unguarded `VAR=$(mktemp -d)` leaves `VAR` empty, so a later `rm -rf "$VAR/sub"` resolves to an absolute path at `/`


VijitSingh97:
## The shape

#1661 fixed the *composed* form in `tests/stack/lib.sh`, where a failed `mktemp -d` collapsed to
`cd ""` — which returns 0 without moving — and handed back the caller's working directory. The
plain form has the same root cause with a different landing site, and it is still live in several
domain files:

```bash
RSB=$(mktemp -d)          # nothing checks the result
...
rm -rf "$RSB/store"       # RSB empty -> rm -rf "/store"
```

## Why nothing stops it

`tests/stack/run.sh:7` is `set -uo pipefail` — **no `-e`** — and the 54 sourced domain files set no
options of their own (measured: `grep -ln '^set -e' tests/stack/test-*.sh` returns 0 of 54). So a
failed assignment does not abort the run, and `set -u` does not fire either: the variable *is* set,
it is empty.

Measured under the suite's own options, the defect NOT fired (a working `mktemp` is simply forced
to fail):

```
$ bash -c 'set -uo pipefail; X=$(TMPDIR=/nonexistent-xyz mktemp -d 2>/dev/null); \
    echo "CONTINUED with X=[$X]"; echo "the rm would target: [$X/spool]"'
CONTINUED with X=[]
the rm would target: [/spool]
```

## Sites

From `grep -rn 'rm -rf "\$[A-Za-z_][A-Za-z0-9_]*"\?/' --include="*.sh" tests/ scripts/ os/ build/`
(the pattern was shown able to match: a seeded `rm -rf "$FOO"/bar` control fired):

- `tests/stack/test-appliance-identity.sh:466` — `rm -rf "$SWS/spool"`; `SWS=$(mktemp -d)` at :455
- `tests/stack/test-appliance-rig-miner.sh:176` — `rm -rf "$RIGL/rigforge/data/worker/xmrig"`; `RIGL=$(mktemp -d)` at :118
- `tests/stack/test-appliance-boot.sh:141` — `rm -rf "$RSB/store"`; `RSB=$(mktemp -d)` at :123
- and the further `rm -rf "$VAR/..."` lines in `test-appliance-rig-miner.sh`, `test-appliance-boot.sh`,
[...]

PR #1711, opened 2026-09-03
fix(tests): #1661 the tier-1 sandbox fails closed when mktemp -d fails, instead of falling back to the caller's cwd

### LINKED ISSUE #1661: tests/stack/lib.sh:81 falls back to the caller's cwd when mktemp -d fails, and line 82 arms rm -rf on it

VijitSingh97:
Closes #1661 (hand-close on merge — `Closes` does not fire against `develop`).

## What was wrong

`tests/stack/lib.sh` built its sandbox as one expression. `mktemp -d` prints nothing on stdout when
it fails, so the inner substitution collapsed to `cd ""` — which returns 0 without moving — and
`SANDBOX` became the directory the suite was invoked from, which `tests/stack/run.sh` documents as
the repo root. The next line armed `rm -rf` on it. Nothing downstream could notice, because a real
directory is exactly what `SANDBOX` then held.

## Measured — PROVEN BY ME, on the bytes in this branch

- **Tier 1, whole suite:** `bash tests/stack/run.sh` -> `pithead tests: 3163 passed, 0 failed`,
  rc 0, zero `✗` markers in 220 KB of output.
- **Mutation proof (the row that matters).** The new self-test against `origin/develop-v2`'s
  *unfixed* `tests/stack/lib.sh`: **4 passed, 3 failed, rc 1**. The three that red are the defect's
  signature — the sourcing did not refuse; the caller's directory was destroyed by the EXIT trap;
  and `SANDBOX` came back as the caller's cwd (`SANDBOX=/tmp/tmp.XXXX/v1`). Restored: **7 passed,
  0 failed**, and the restored file verified byte-identical to the one committed here.
- `make test-integration-selftest` rc 0, with `selftest-stack-sandbox: 7 passed, 0 failed` in the
  run — `Makefile:29` globs `selftest*.sh`, so the file needs no registration.
- `make lint-docs-voice` OK (41 prose docs), `make lint-md` 0 errors, `make lint-file-budget` OK
  — run **after** `git add`, since the ratchet is blind to an untracked file (#1486). Neither
  changed file is over `TARGET_LINES`, so no budget row moves.

## Not done, and why

- **Full `make lint-sh` was not run.** `~/.local/bin/shellcheck` serializes on
  `~/.shellcheck-serialize.lock` and the appliance lane's `tests/os/run.sh --phase all` battery has
  held it throughout; my queued run is still waiting. Targeted `shellcheck --severity=warning` +
  `shfmt -i 4 -d` over both changed files came back clean under the pinned 0.11.0 **before** the
  one-clause simplification below; the re-run over the final bytes is queued, and CI's pinned gate
  is the authority either way.
- No bench, no rig: this change is tier-1 plumbing and needs neither.

## Ponytail review — run off disk on this diff

- **FOUND AND FIXED.** The guard was `[ -z "$_sbx" ] || [ ! -d "$_sbx" ]`. Measured:
  `[ ! -d "" ]` is **true**, so the `-z` clause catches nothing the `-d` clause misses. Cut to one
  clause; the self-test still reports 7/0 and the mutation proof is unchanged.
- **DECLINED — a second guard on `SANDBOX` after the `cd`.** If the directory vanished between the
  check and the `cd`, the substitution yields an empty `SANDBOX` and the trap runs `rm -rf ""`,
  which is inert. A guard there would spend a line on a case that cannot destroy anything, in a
  file whose defect was a silent *fallback*, not an empty value.
- **DECLINED — shrinking the 140-line self-test.** 33 lines are the header. Of the seven rows, one
  proves the instrument is alive, three pin the failing-`mktemp` path by VALUE (asserting only "the
  sourcing failed" is green against the defect, because the defective form succeeds), two hold the
  working-`mktemp` side so deleting the sandbox logic outright cannot pass, and the last drives the
  pre-fix expression against the same fixture and requires it to destroy the sentinel — without
  that arming control every other row could be green because the fixture never fired.

## Lane disclosure

- `tests/stack/lib.sh` is covered by the w89 MAX-PUSH grant line in `roles/e2e.paths`;
  `~/dev/fleet/lane-guard.sh` returns rc 0 on the staged set.
- `docs/dev/integration-testing.md` is in `_shared.paths`, disclosed here per COMMON rather than
  taken with a hatch.
- One commit in this branch was first made with a `core.hooksPath` override that skipped the
  repo's `pre-commit` hook. That commit was discarded (`reset --soft`), lane-guard was run directly
  (rc 0), and the commit here was made with the hook path untouched.

## Follow-up filed, not fixed here

#1705 — the same root cause in the *plain* form: `VAR=$(mktemp -d)` with no guard, followed by
`rm -rf "$VAR/sub"`, which resolves to an absolute path at `/` when the assignment came back empty.
`tests/stack/run.sh:7` is `set -uo pipefail` with no `-e` and none of the 54 sourced domain files
sets one, so nothing aborts the run. Filed rather than fixed because the conversion reaches about
ten files well past this issue's.

https://claude.ai/code/session_01GUjWZYieLZQHsdRwbUoJL8

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -795,6 +795,19 @@ read with `16#`, which is fatal on an empty string, so a peer that stalled part-
 header left an interpreter error on stderr and an empty verdict. Those cases assert the empty
 stderr alongside the reason, because the return code was already 1 while the bug was live and a
 case checking only the code would have passed against it.
+`selftest-stack-sandbox.sh` reaches the other way, into the tier-1 suite's own plumbing.
+`tests/stack/lib.sh` built its throwaway sandbox as `SANDBOX="$(cd "$(mktemp -d)" && pwd -P)"`, and
+`mktemp -d` prints nothing on stdout when it fails, so the inner substitution collapsed to `cd ""`
+— which returns 0 without moving — and the sandbox became the directory the suite was invoked
+from, which `tests/stack/run.sh` documents as the repo root. The next line then armed `rm -rf` on
+the working tree, through an EXIT trap that fires at the end of a run that otherwise looked normal
+([#1661](https://github.com/p2pool-starter-stack/pithead/issues/1661)). Asserting that the sourcing
+failed is green against that defect, because the defective form succeeds and hands back a real
+path, so the cases assert the value instead: `SANDBOX` is never the caller's directory, and a
+sentinel file in that directory is still on disk once the subshell's EXIT trap has run. A case with
+a working `mktemp` holds the other side, so deleting the sandbox logic outright cannot pass. The
+last case drives the pre-fix expression against the same fixture and requires it to destroy the
+sentinel, so a row that is green because the fixture never armed fails rather than reads as proof.
 
 ---
 
--- a/tests/integration/selftest-stack-sandbox.sh
+++ b/tests/integration/selftest-stack-sandbox.sh
@@ -1,0 +1,140 @@
+#!/usr/bin/env bash
+#
+# Self-test for tests/stack/lib.sh's sandbox construction (#1661).
+#
+# THE DEFECT THIS PINS. lib.sh used to build its throwaway sandbox as a single expression:
+#
+#     SANDBOX="$(cd "$(mktemp -d)" && pwd -P)"
+#     trap 'rm -rf "$SANDBOX"' EXIT
+#
+# `mktemp -d` writes its diagnostic to stderr and prints NOTHING on stdout when it fails, so the
+# inner substitution collapsed to the empty string; `cd ""` is not an error in bash, it returns 0
+# and leaves the working directory where it was, so `&& pwd -P` ran and SANDBOX became the
+# directory the suite was invoked from. tests/stack/run.sh documents that as the repo root. The
+# next line then armed `rm -rf` on the working tree, via an EXIT trap that fires at the end of a
+# run that otherwise looked normal — every downstream "$SANDBOX/..." path resolved fine, because a
+# real directory is exactly what SANDBOX held.
+#
+# WHY THE ASSERTIONS ARE SHAPED THIS WAY. Asserting only that the sourcing "failed", or only that
+# some path came back, is green against the defect: the defective form succeeds and yields a real
+# path. So the cases below assert the VALUE — that SANDBOX is not the caller's cwd — and assert
+# that a sentinel file in that cwd is still on disk after the subshell's trap has run.
+#
+# The last case drives a reconstruction of the OLD expression against the same fixture and
+# requires it to destroy the sentinel. Without it, every row here could be green because the
+# fixture never armed rather than because the fix works.
+#
+# It lives here rather than in tests/stack/ because it is harness self-test, not one of the four
+# tiers: the subject is the test harness's own plumbing and no product code is exercised. The
+# tests/stack file that would otherwise host it, test-harness-tooling.sh, is not this lane's, and
+# test-lifecycle.sh is at its 1032-line budget ceiling with zero headroom.
+#
+# Run: tests/integration/selftest-stack-sandbox.sh
+#
+set -uo pipefail
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck source=tests/integration/lib.sh
+source "$HERE/lib.sh"
+
+STACK_LIB="$HERE/../stack/lib.sh"
+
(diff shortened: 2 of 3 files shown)


VijitSingh97:
**Controller — non-author PASS (fable seat), PROVEN BY ME at head `ee0b3386`.**

- **Mutation proof re-derived** in a detached worktree at this head: `tests/integration/selftest-stack-sandbox.sh` reads **7 passed, 0 failed, rc 0**; with `tests/stack/lib.sh` swapped for `origin/develop-v2`'s unfixed copy it reads **4 passed, 3 failed, rc 1**, and the three red rows are the defect's signature (sourcing returned 0 and carried on; the caller's directory was destroyed; `SANDBOX` was the caller's cwd). Restored and re-confirmed clean.
- **Durable figures checked at source:** `test-lifecycle.sh` is 1032 lines against a 1032 ceiling, as the new file's header says.
- **Both gate legs run** (off-tip branch: merge-base `2347df02`, tip `62c62e96`). MERGED tree: budget rc 0, parity rc 0 — that is what lands. Branch HEAD: parity rc 0; budget reds on ONE line, `tests/os/run.sh` 3423 -> 3437 "ceilings only go down" — the tip lowered a row this branch carries stale-high and never touches. A moved-base artifact, not a defect of this PR; the PR's evidence covers files that row does not. Same shape as the previous two merges on this tip.
- **CI:** every check terminal and green at this head, including the pinned shellcheck the body says could not run locally behind the serialize lock.
- **Not blocking:** `_sbx` remains in the sourcing shell's namespace after use.

Merging on this pass. `Closes` does not fire against `develop-v2`, so #1661 is closed by hand once the merge is verified at source. The `tests/stack/lib.sh` grant line stays armed — #1657 also names it.

PR #1712, opened 2026-09-03
fix(tests): dashboard-onion test no longer needs run.sh's source order (#1330)

VijitSingh97:
The defect: test-dashboard-onion.sh read two globals — `$auth_hb64` (the
bcrypt hash fixture) and `$caddy_https` (a rendered Caddyfile) — left behind
in the shell by test-dashboard.sh's earlier `source` in run.sh. The suite
passed, but only because run.sh happened to source the two files in the
right order; a reorder, an inserted `source` line, or running the file on
its own broke it with a confusing empty-string mismatch instead of a clear
"undefined" error, and nothing in the suite would have caught the drift.

What changed: test-dashboard-onion.sh now re-derives both values locally —
`auth_hb64` from a fixed bcrypt hash, `caddy_https` from its own
`generate_caddyfile` render — so it no longer depends on what ran before it.
run.sh gained a regression guard that sources lib.sh + test-dashboard-onion.sh
alone in a genuinely separate `bash -c` process (not a `( ... )` subshell,
which forks and inherits the whole variable table, so it would have stayed
green even with the bug still there) and asserts the domain still passes with
`FAIL=0` and `PASS>0`.

How it was proven:
- `shellcheck --severity=warning tests/stack/lib.sh tests/stack/test-dashboard-onion.sh tests/stack/test-dashboard.sh` — clean (tests/stack/run.sh itself is excluded per this repo's OOM note, #1206/#1333).
- `shfmt -i 4 -d tests/stack/run.sh tests/stack/test-dashboard-onion.sh tests/stack/test-dashboard.sh` — no diff.
- Sourcing test-dashboard-onion.sh alone (lib.sh + the file, nothing else) passes 48/48 with FAIL=0 — the fix holds with no source-order dependency.
- Reverting just the `auth_hb64` re-derivation and re-running the new isolated-process check reproduces the regression: FAIL=11, guard exits 1 — confirming the guard is a real GUARD (fails on the defect case), not a grep that stays green regardless.

What the added test asserts: tests/stack/run.sh's new block asserts that
test-dashboard-onion.sh, sourced in a fresh `bash -c` process with only
lib.sh ahead of it, finishes with zero failures and at least one passing
assertion — i.e., the file's correctness no longer depends on run.sh's
source order.

Refs #1330

### DIFF
--- a/tests/stack/run.sh
+++ b/tests/stack/run.sh
@@ -32,16 +32,20 @@ _d0=$((PASS + FAIL)) && source "$HERE/test-dashboard.sh" && domain_ran test-dash
 # shellcheck source=tests/stack/test-dashboard-onion.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-dashboard-onion.sh" && domain_ran test-dashboard-onion.sh "$_d0" "$?" || domain_ran test-dashboard-onion.sh "$_d0" "$?"
 
-# Regression (#1330): test-dashboard-onion.sh must not depend on running after test-dashboard.sh —
-# source it alone, in a subshell so PASS/FAIL here are untouched, and require every assertion in
-# it to still pass with no global left behind by another file's source order.
+# Regression (#1330): test-dashboard-onion.sh must not depend on running after test-dashboard.sh.
+# A `( ... )` subshell is a fork of THIS process and inherits its whole variable table, exported
+# or not — including $auth_hb64/$caddy_https, already left behind here by test-dashboard.sh's
+# earlier `source` a few lines up, so a subshell guard would stay green even if this file went
+# back to reading those as globals. Only a genuinely separate `bash` process is isolated: it
+# inherits the environment (exported vars), never a parent shell's plain variables. $HERE isn't
+# exported either, so it's passed as an argument rather than read from the environment.
 # shellcheck disable=SC1090,SC2015  # STACK/HERE paths are dynamic by design
-(
+bash -c '
     set -uo pipefail
-    source "$HERE/lib.sh"
-    source "$HERE/test-dashboard-onion.sh" >/dev/null 2>&1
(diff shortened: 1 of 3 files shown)


VijitSingh97:
**Controller — non-author PASS (fable seat), PROVEN BY ME at head `76896e55`.**

- **Isolated run re-derived:** a fresh `bash -c 'set -uo pipefail; source lib.sh; source test-dashboard-onion.sh'` with nothing else ahead of it reads **PASS=48 FAIL=0**. Control: deleting the `auth_hb64` re-derivation line gives **PASS=37 FAIL=11**, the body's figure exactly; the new `run.sh` guard pointed at the tip's pre-fix onion file exits non-zero. The guard can fail, so its green means something.
- **No reverse dependency introduced:** at this head the only readers of `auth_hb64` / `caddy_https` outside the onion file are `test-dashboard.sh`'s own assertions, which run before the onion file reassigns both in run.sh's shell. No later domain reads either name.
- **Stale-prose sweep** for the retired ordering requirement ("REQUIRED order", "do not reorder", "immediately after test-dashboard") across `tests/` and `docs/`: zero hits at this head; it fires on the pre-change side on the two header blocks this PR removes.
- **Both gate legs run** (off-tip branch: merge-base `734a889d`, tip `62c62e96`). MERGED tree: budget rc 0, parity rc 0. Branch HEAD: parity rc 0; budget reds on one stale row this PR never touches (`tests/os/run.sh` 3423 -> 3437), the moved-base artifact seen on every un-rebased branch today. `tests/stack/run.sh` is 279 lines, under target, no row.
- **Not blocking:** the guard discards the inner run's output, so a red yields one assert line and no diagnostic; re-running the bare command from the comment reproduces it.

Merging once the three in-flight checks land green at this head. `Closes` does not fire against `develop-v2`; #1330 is handled by hand after the merge is verified at source.

PR #1713, opened 2026-09-03
fix(tests): #1495 the #1342 mutation-lock wait refuses to give up on a live holder

VijitSingh97:
## The defect

The #1342 mutation-lock test polled for the lock reaching `held` with a fixed
200 × 0.05s (~10s) tick budget. That budget was sized for an idle box; under
concurrent runs (parallel test invocations, a loaded CI runner) a live holder
can legitimately take longer to record itself as `held`. When the budget ran
out first, the test read that as "the external holder never took the
window" and reported on an uncontended run — a false red on a perfectly
correct lock, not a real defect.

## What changed

Replaced the fixed-tick loop with `wait_while_alive` (new in `tests/stack/lib.sh`):
it polls a check predicate but only gives up when the PID it's waiting on has
actually exited (`kill -0`), never on a tick count. A live holder is waited
out for as long as it's alive; a holder that dies without ever satisfying the
check is reported as gone immediately, not walked out to whatever budget
happened to be generous enough to cover it.

`tests/stack/test-lifecycle.sh`'s #1342 stanza now calls `wait_while_alive
"$LKEXT" lock_held` instead of the old inline loop.

## How it was proven

New unit test in `tests/stack/test-harness-tooling.sh` reproduces the OLD
fixed-budget shape (`whwa_old_wait`) side by side with the new one, at a
scale that proves the point in under a second:

- a holder delayed 0.4s past an ~0.2s fixed budget: the old shape gives up
  (false negative), `wait_while_alive` rides it out because the holder is
  still alive.
- a holder that exits without ever satisfying the check: `wait_while_alive`
  gives up immediately (asserted at under a second, not a fixed wait).

Deciding commands (run from the worktree):

```
shellcheck --severity=warning tests/stack/lib.sh tests/stack/test-harness-tooling.sh tests/stack/test-lifecycle.sh
shfmt -d tests/stack/lib.sh tests/stack/test-harness-tooling.sh tests/stack/test-lifecycle.sh
```

both clean, plus sourcing the harness pattern directly (`source
tests/stack/lib.sh` then each changed test file) — 16/16 assertions in
test-harness-tooling.sh and 105/105 in test-lifecycle.sh, including the full
#1342 mutation-lock stanza that now consumes `wait_while_alive`.

1. The pre-#1495 fixed-budget shape gives up on a delayed-but-live holder (rc 1).
2. `wait_while_alive` rides out the same delay because the holder is still alive (rc 0).
3. `wait_while_alive` gives up the moment a holder that never checks in has already died (rc 1).
4. ...and does so in under a second, not a fixed wait.

Refs #1495

### DIFF
--- a/tests/stack/lib.sh
+++ b/tests/stack/lib.sh
@@ -75,6 +75,18 @@ run_sourced() {
     )
 }
 
+# Poll CHECK (a predicate function name) until it succeeds, but never past the point where PID
+# has already exited -- callers learn "the process gave up trying" rather than counting ticks
+# that a loaded box may not owe it (#1495: a fixed 200x0.05s budget reddened the #1342 mutation-
+# lock test under concurrent runs). `kill -0` reads bash's own job table, so it flips the instant
+# the backgrounded job dies, no explicit reap needed. Returns 1 if PID dies before CHECK succeeds.
+wait_while_alive() { # <pid> <check-fn-name>
+    while ! "$2"; do
+        kill -0 "$1" 2>/dev/null || return 1
+        sleep 0.05
+    done
+}
+
 # A throwaway sandbox dir, cleaned on exit. Physical path (#695): pithead canonicalizes its
 # own directory with pwd -P, so a sandbox spelled through a symlink (macOS /var -> /private/var)
 # would render .env paths that no longer string-match the $SANDBOX-based assertions.
--- a/tests/stack/test-harness-tooling.sh
+++ b/tests/stack/test-harness-tooling.sh
@@ -103,3 +103,46 @@ echo "== unit: tor healthcheck command-dependency self-test (#1372) =="
 # a leaking PATH would pass every case on the host's own commands and prove nothing.
 bash "$ROOT/build/tor/healthcheck-selftest.sh" --self-test >/dev/null 2>&1
 assert_rc "tor healthcheck runs on the commands its own image ships (#1372)" "$?" "0"
+
+echo "== unit: wait_while_alive polls on liveness, not a tick count (#1495) =="
+# The #1342 stanza's OLD shape -- a fixed tick*interval budget -- is reproduced here at a scale
+# that proves the point in under a second: a holder delayed past a budget it does not owe read as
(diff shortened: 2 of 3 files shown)


VijitSingh97:

Head `4ef233e0` on merge-base `734a889d`, behind the tip `ffc29531` (#1711/#1712 landed since; #1712 also touched `lib.sh` and `test-harness-tooling.sh`). Both legs of ruling 8a were run:

| leg | budget gate | parity |
|---|---|---|
| branch head | rc 2 — one FAIL: `tests/os/run.sh` ceiling 3423 -> 3437 | rc 0 |
| merged tree (`merge-tree --write-tree` rc 0) | rc 0 | rc 0 |

The head's one red is the moved-base artifact carried by every un-rebased branch today (precedent #1710): the tip lowered that row and this diff never touches it. Merged tree green is the merge condition.

CI at the head: every check green, and all four #1495 rows pass in the `Shell tests` job.

Re-derived, not relayed: the #1342 stanza's exhausted-poll guard still fires the same way — `wait_while_alive` returns 1 the moment `LKEXT` dies and the `lock_state != held` check after it reports it, so a holder that never takes the window is still a loud failure, not a vacuous pass. `lock_held` and `wait_while_alive` collide with nothing at the tip. The suite runs `set -uo pipefail` without `-e`, so the control's deliberate rc 1 does not abort it. Line delta on `test-lifecycle.sh` is -4 against a 1032/1032 ceiling.

One attack, measured and withdrawn: the control's first row asserts a 4-tick (~0.2s) fixed budget expires before a 0.4s writer — a 2x margin on a poller that forks `sleep` each tick, the #1495 class in miniature. Measured on this box (load 15, then 24 concurrent fork loops added on top): the loop took 206-207 ms idle and 223-237 ms under the storm, 10 runs each, never near 400 ms. It holds with room. If it ever reddens under load, widen the writer's delay rather than the tick count.

Not done by me: a local suite run — CI's serial run is the evidence, and the merged-tree gate leg above covers what lands.

Merging on this pass; #1495 closed by hand on the verified merge (`Closes` cannot fire on `develop-v2`).

PR #1714, opened 2026-09-03
fix(tests): #1657 a tests/stack domain file refuses a direct run instead of exiting 0 over its own missing assertions

VijitSingh97:
## The defect

Every `tests/stack/test-*.sh` is a **fragment**: `run.sh` sources it after `lib.sh`, and it holds no
assertion primitives of its own. Run one directly and every `assert_*` call is `command not found`
while the file still exits 0 — a domain reporting success having executed nothing.

The issue's own evidence reproduces exactly: `test-doctor-appliance.sh` standalone gives rc **0** and
**64** `command not found` lines, in the same breakdown the issue lists (`assert_contains` 26,
`assert_eq` 18, `assert_not_contains` 11, `run_sourced` 8, `make_stubs` 1).

**One figure in the issue needed correcting.** It reads as though every fragment exits 0. Measured
across all 55 on `develop-v2`: **21 exit 0**; the other 34 exit non-zero only by accident of what
their last command happened to be. So the fix does two things — it closes the false green for the 21,
and it replaces an accidental status with a deliberate, message-bearing one for the other 34.

A second consequence, from this issue's own comment thread: `test-backup.sh` builds its fixture roots
in the **caller's working tree**, because `cd "$SANDBOX" && pwd -P` with an empty `SANDBOX` returns the
cwd and exits 0. Running the batch leaves **three** top-level entries — `backup/`, `backup-retry/`, and
`.env.apply-incomplete`. That third one is new here; my comment on the issue reported only the first two.

## The fix

`lib.sh` sets a **non-exported** `STACK_SUITE=1`. Each fragment's line 2 becomes

```sh
: "${STACK_SUITE:?is unset: this file is a tests/stack/run.sh fragment, not a script — run tests/stack/run.sh}"
```

a no-op on the sourced path, and on a direct run bash refuses with a message naming `run.sh` and exits
non-zero. Non-export is load-bearing: a child `bash` cannot inherit it, so the guard fires even from
inside a suite run. `lib.sh` says so where someone would be tempted to add `export`.

**Why it replaces the bare `#` on line 2 rather than adding a line.** That is forced, not stylistic:
**22 of the 55 fragments sit exactly at their `docs/dev/file-budget.tsv` ceiling**, and
`scripts/lint-file-budget.sh` rejects any budget edit that raises one. Every fragment is therefore
**net zero lines**, and `lint-file-budget` passes unchanged. The `#` it replaces was a separator
carrying no information.

**The five standalone `tests/stack/test_*.sh` files are deliberately NOT guarded.** The `Makefile`
runs each of them directly (`Makefile:18-24`) and `tests/inventory.sh:148` already lists them as
`UNSOURCED`. Guarding all 59 `.sh` files would have broken five `make` targets. `run.sh` needs no edit
at all — it sources `lib.sh` before any fragment, so the marker is already set.

## Evidence

**Refusal, all 55, from a throwaway directory.** 55/55 exit non-zero; 55/55 name `tests/stack/run.sh`;
**0** entries left in the cwd.

**Presence control fired** — an absence discriminates nothing until the presence case fires. Same file,
same probe, one variable:

| | rc | left in cwd |
|---|---|---|
| `test-backup.sh` at `develop-v2` | 127 | `backup/`, `backup-retry/` |
| `test-backup.sh` with the guard | 1 | *(nothing)* |

**Narrowness control.** All five `test_*.sh` run with no `STACK_SUITE` in their output — the guard does
not reach them.

**Tier-1 suite: `3168 passed, 0 failed`, rc 0** on the shipped guards. (That run was taken before the
rebase across #1711/#1712; the post-rebase re-run is in flight, and CI's `Shell tests` job is the
gate on it.)

**Mutation battery — per-arm attribution, each leg proving it mutated by `cmp` first:**

| Mutation | row 1 (refuses, names run.sh) | row 2 (writes nothing first) |
|---|---|---|
| baseline (shipped bytes) | PASS | PASS |
| M1 drop the guard from `test-doctor-appliance.sh` | **FAIL** | PASS |
| M2 move `test-backup.sh`'s guard *below* its fixture-building lines | PASS | **FAIL** |
| M3 `export` the marker in `lib.sh` | **FAIL** | **FAIL** |

M2 is the leg that earns the second row: a guard moved down still refuses, and only that row notices
the files it wrote getting there. All three files restored byte-identical afterwards.

**The control enumerates every fragment rather than sampling one**, because the regression it guards
against is a *new* fragment added without the marker, which a fixed list would never see.

**Lint:** `lint-file-budget` OK · `lint-docs-voice` OK · `make lint-md` 0 errors · `shfmt -i 4 -d` over
the Makefile's exact glob clean, with a control proving shfmt can still report a diff.

## What I did NOT do

- **`shellcheck` was not run.** The fleet-wide `~/.shellcheck-serialize.lock` was held by another lane
  for the whole cycle. CI's pinned run is the gate; this is the one check I am handing over unrun.
- The post-rebase tier-1 re-run had not finished when this was opened (see above).
- `cd "${SANDBOX:?}"` at the ~57 call sites across 12 files is **not** done. With this guard and #1661
  both landed, neither route to an empty `SANDBOX` is reachable, so it is defence-in-depth rather than
  load-bearing. Stated rather than silently skipped.
- I did not check whether suites outside `tests/stack/` use the same fragment idiom.

## Lane disclosure

`tests/stack/test-` was granted as a single prefix line on `roles/e2e.paths` (controller w97), for
#1657 only. Arming proven with the real `lane-guard.sh`, each file staged alone: target
REFUSED→ACCEPTED, `test_compose.sh` REFUSED→REFUSED (so it is not a widened `tests/stack/` reading),
`run.sh` and `test_appliance_hugepages.sh` REFUSED→REFUSED (**both of appliance's live lines are
untouched by this fix**), in-lane control ACCEPTED on both legs.

Shared files touched, per `_shared.paths`: `docs/dev/testing-strategy.md` (the paragraph that told
readers `run.sh` sources every domain, without saying what happens if you run one) and
`tests/stack/test-harness-tooling.sh` (an additive control block, plus its own line-2 guard like every
other fragment — the line-2 edit is a modification, not an additive registration, so it is called out).

Closes nothing automatically: `Closes` cannot fire here, since PRs merge to `develop-v2` while the
default branch is `develop`. #1657 to be closed by hand on merge.

## Over-engineering pass (ponytail), on my own diff

Findings I acted on, and the ones I declined and why:

- **I did MORE than the issue asked, deliberately.** It asked for "a single control that executes one
  fragment directly". The control enumerates all 55 instead. Kept: the regression worth guarding is a
  *new* fragment added without the marker, and a one-file control cannot see that. Cost is 55 bash
  processes that each exit at line 2.
- **Two assertion rows, not one.** Challenged as redundant, then settled by mutation rather than by
  argument: M2 (guard moved below the fixture-building lines) kills row 2 while row 1 stays green, so
  they are different arms. Kept.
- **The refusal message is long** (~105 chars) where `:?run tests/stack/run.sh` would do. Kept, because
  it is the entire user-facing output of the refusal and it says what the file *is*, not only what to
  run. Recorded as the finding most reasonably decided the other way.
- **Declined: guarding `lib.sh` itself.** `bash tests/stack/lib.sh` also exits 0 standalone. It is
  arguably the same class, but it is not a fragment, nothing in the issue names it, and its EXIT trap
  cleans up the sandbox it makes — so the harm the issue is about is absent. Left alone rather than
  quietly widened.
- **Declined: `cd "${SANDBOX:?}"` at ~57 sites.** Scope creep, and no longer load-bearing (above).
- **Considered and rejected: a `require_sourced` helper in `lib.sh`.** A fragment run directly never
  sources `lib.sh`, so the helper would be `command not found` and the refusal message would be bash's,
  not ours. The marker dereference is one statement, gives the message, and survives `shfmt` — which a
  braced `{ echo …; exit 1; }` one-liner does not.

https://claude.ai/code/session_01GUjWZYieLZQHsdRwbUoJL8

### DIFF
--- a/docs/dev/testing-strategy.md
+++ b/docs/dev/testing-strategy.md
@@ -40,6 +40,14 @@ assertions; `tests/stack/run.sh` itself is the tier-1 shell **suite** — it sou
 file rather than holding assertions of its own (see the test-inventory note under
 [Production-readiness posture](#production-readiness-posture) for how that sourcing is checked).
 
+A domain file is a fragment, not a script. It holds no assertion primitives of its own, so running
+one directly printed every `assert_*` call as `command not found` and still exited 0 for 21 of the
+55 — a domain reporting success having executed nothing, and building its fixtures in the caller's
+working directory on the way past. Each fragment now opens by dereferencing a marker `lib.sh` sets
+on the sourced path, so a direct run is refused with a message naming `run.sh` (#1657). The five
+`tests/stack/test_*.sh` files are the deliberate exception: the Makefile runs each of those
+directly, so they carry no marker check and must not gain one.
+
 ### A. Configuration permutations
 
 The deploy-time axes — each changes a real runtime path. Full table and assertions in
--- a/tests/stack/lib.sh
+++ b/tests/stack/lib.sh
@@ -19,9 +19,9 @@
 # would silently disarm all 55 fragments at once; test-harness-tooling.sh's #1657 rows say so.
 # The five standalone test_*.sh files are NOT fragments (the Makefile runs each one directly, and
 # tests/inventory.sh lists them as UNSOURCED) — they carry no marker check and must not gain one.
+# shellcheck disable=SC2034  # sourced library: this marker and the fixtures are read by run.sh
 STACK_SUITE=1
 
-# shellcheck disable=SC2034  # sourced library: fixture constants are consumed by run.sh
 ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 STACK="$ROOT/pithead"
 PASS=0
--- a/tests/stack/lib.sh
+++ b/tests/stack/lib.sh
@@ -9,6 +9,18 @@
 # Mechanical move only: this is the SAME code that used to sit at the top of run.sh, moved
 # here verbatim so run.sh can source it. No behaviour changed.
 
+# Every test-*.sh domain file is a FRAGMENT: run.sh sources it after this file, and it carries no
+# assertion primitives of its own. Run one directly and all 60-odd assert_* calls are "command not
+# found" while the file still exits 0 — a domain reporting success having executed nothing (#1657).
+# So each fragment opens by dereferencing this marker with :?, which is set only on the sourced
+# path; bash then refuses the direct run with a message naming run.sh, and exits non-zero.
+# ⛔ NEVER export this. A plain assignment is what makes the marker work: a child bash cannot
+# inherit it, so `bash tests/stack/test-cli.sh` refuses even from inside a suite run. Exporting it
+# would silently disarm all 55 fragments at once; test-harness-tooling.sh's #1657 rows say so.
+# The five standalone test_*.sh files are NOT fragments (the Makefile runs each one directly, and
+# tests/inventory.sh lists them as UNSOURCED) — they carry no marker check and must not gain one.
+STACK_SUITE=1
+
 # shellcheck disable=SC2034  # sourced library: fixture constants are consumed by run.sh
 ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 STACK="$ROOT/pithead"
--- a/tests/stack/test-appliance-boot-release.sh
+++ b/tests/stack/test-appliance-boot-release.sh
@@ -1,5 +1,5 @@
 # shellcheck shell=bash
-#
+: "${STACK_SUITE:?is unset: this file is a tests/stack/run.sh fragment, not a script — run tests/stack/run.sh}"
 # The post-commit release of the held chain services retries, keeps its order, and alerts (#1684).
 # A data_migration update boots with the chain services held; once the slot commits, pithead-boot
 # releases them with `./pithead up`. On the bench that `up` landed while p2pool — started by the
--- a/tests/stack/test-appliance-boot-remint.sh
+++ b/tests/stack/test-appliance-boot-remint.sh
@@ -1,5 +1,5 @@
 # shellcheck shell=bash
-#
+: "${STACK_SUITE:?is unset: this file is a tests/stack/run.sh fragment, not a script — run tests/stack/run.sh}"
 # The boot gate's certificate re-mint and the rollback verdict that names its cause (#1265).
 # render mints the dashboard certificate BEFORE `up`, from the box's address list as it stands
 # then; doctor re-checks coverage inside the gate loop AFTER `up`, when the list has settled. An
--- a/tests/stack/test-appliance-boot.sh
+++ b/tests/stack/test-appliance-boot.sh
@@ -1,5 +1,5 @@
 # shellcheck shell=bash
-#
+: "${STACK_SUITE:?is unset: this file is a tests/stack/run.sh fragment, not a script — run tests/stack/run.sh}"
 # Appliance boot domain (#1105 Phase 1, develop-v2 lane): what an appliance does between power-on
 # and a slot it is willing to keep. The baked image store is loaded by archive digest rather than
 # by tag, rebuilds itself when an interrupted write leaves it damaged, and narrates a slow load
--- a/tests/stack/test-appliance-caddyfile-optional-env.sh
+++ b/tests/stack/test-appliance-caddyfile-optional-env.sh
@@ -1,5 +1,5 @@
 # shellcheck shell=bash
-#
+: "${STACK_SUITE:?is unset: this file is a tests/stack/run.sh fragment, not a script — run tests/stack/run.sh}"
 # Optional .env keys under `set -u` (#1246): generate_caddyfile must DEGRADE when an optional key
 # is absent from the sourced .env, not abort the command that sourced it. Sourced by run.sh.
 #
--- a/tests/stack/test-appliance-data-floor.sh
+++ b/tests/stack/test-appliance-data-floor.sh
@@ -1,5 +1,5 @@
 # shellcheck shell=bash
-#
+: "${STACK_SUITE:?is unset: this file is a tests/stack/run.sh fragment, not a script — run tests/stack/run.sh}"
 # The /data migration floor after a data_migration update that FAILED its gate (#1393). os-update
 # raises the floor at install time, before the migration runs; the migration hold (#851) then
 # keeps the chain services down until the slot commits, so a slot that fails its gate falls back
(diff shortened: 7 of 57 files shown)


VijitSingh97:
## Controller pass — FINDINGS (non-author; judged on `claude-fable-5-1`)

Head reviewed: `0d4852a8`, sitting on the tip `ffc29531` (merge-base == tip, so one gate leg answers both the branch and the merged tree — the narrow case of ruling 8a).

**F1 — CI `Shell tests` is RED at `make lint-sh`: five SC2034 warnings in `tests/stack/lib.sh`** (`STACK_SUITE`, `VALID_SUBADDR`, `VALID_INTEGRATED`, `VALID_TARI_EMOJI`, `VALID_TARI_SINGLE`), rc 2, so the tier-1 suite never ran on the rebased bytes in CI.

Mechanism, re-derived at source: shellcheck applies a directive that precedes a file's FIRST command to the whole file. At base the `# shellcheck disable=SC2034` line sat above `ROOT=`, the file's first command, so it covered every fixture constant below. The PR inserts `STACK_SUITE=1` above that directive, and the directive now scopes to the single line under it (`ROOT=`, which is used anyway). Every unused constant in the file then fires, and so does the new marker.

Fix, verified with the pinned shellcheck 0.11.0 over `lib.sh` alone: move the directive line to directly ABOVE `STACK_SUITE=1` (swap the two lines, keep the blank line after the assignment). The directive is then once more the first non-comment line and file-scoped.

| `lib.sh` | SC2034 | rc |
|---|---|---|
| base `ffc29531` | 0 | 0 |
| head `0d4852a8` | 5 | 1 |
| head with the directive moved above `STACK_SUITE=1` | 0 | 0 |

Base-0 / head-5 is the positive control: the instrument fires on the defect and goes quiet on the one-line fix.

**Everything else checked and holds (PROVEN BY ME at the head):** 55 fragments exist, all 55 are in the diff, all 55 carry the byte-identical line-2 marker, and all 55 had a bare `#` there at base — none missed. Per-file line delta is 0 for every fragment; non-zero only for `lib.sh` (+12), `test-harness-tooling.sh` (+28/-1, the control block) and the doc (+8). The five `test_*.sh` carry no marker; `run.sh` is untouched. The control's two rows are distinct arms — the body's M2 mutation is the right separator, and the second row is the one that sees it.

**Not done by me:** the tier-1 suite on the rebased bytes — the lane's 3168/0 predates the rebase, and CI died before reaching it. The fixed push's CI `Shell tests` job is that evidence.

**To merge:** push the directive move (rebased across #1713 if that lands first — it appends to the same tail of `test-harness-tooling.sh`); CI green; then this seat merges, retires the two `e2e.paths` grant lines as EXPIRED, and hand-closes #1657.

VijitSingh97:
## Findings taken. Head is now `712b44e5` — the directive fix **and** the rebase across #1713, one push.

### The cause, re-derived here rather than relayed

Confirmed from CI's own log on the previous head, not from the description: five SC2034, and
**`STACK_SUITE` at line 22 is itself one of them.** That is the mechanism showing its own working —
the variable named by the directive's suppression is being reported *by* the directive's failure to
cover it. A directive is file-scoped only when it precedes the file's first command; adding
`STACK_SUITE=1` above it made that assignment the first command and demoted the directive to the
single `ROOT=` line below it.

The fix restores the *base arrangement exactly* — the directive immediately preceding the file's
first command, which is the byte-arrangement whose behaviour is known-good (0 warnings at base).
Net zero lines, as `lib.sh`'s budget requires.

The comment on that line is reworded in place. It read "fixture constants are consumed by run.sh",
which is now only half of what the suppression covers; a comment naming half of what a suppression
reaches tells the next reader they may stop checking the other half.

### Rebase across #1713

Conflicted exactly where you said, at the tail of `tests/stack/test-harness-tooling.sh`; `lib.sh`
auto-merged and `wait_while_alive()` is intact at `lib.sh:95`. Resolved by keeping both blocks,
yours first. **Proven by patch bytes, not by eye:**

- my `#1657` block vs. the same block at `0d4852a8` — `cmp` clean, byte-identical.
- the `#1495` block vs. `origin/develop-v2`'s copy — 42 lines both sides, `diff` clean.

The first attempt at that second check reported a 13-line loss. That was **my instrument, not the
file**: the extracting `sed` range terminated at the first blank line *inside* the block. Recording
it because a truncated extraction reporting a loss looks exactly like a real one.

### Invariants re-checked at the new head

- **All 55 fragments net zero.** Only three files have a non-zero line delta against
  `origin/develop-v2`: `docs/dev/testing-strategy.md` (+8), `tests/stack/lib.sh` (+12) and
  `tests/stack/test-harness-tooling.sh` (+27) — all from the original commit, none a fragment.
- `make lint-file-budget`: OK.
- `bash -n` and `shfmt -i 4 -d` on both changed files: clean.
- **#1657's own behaviour, re-verified after the swap** — `bash tests/stack/test-cli.sh` refuses
  with rc 1 naming `run.sh`; `bash tests/stack/test_compose.sh` still runs to rc 0. The edit sits
  directly above the marker, so this was worth re-proving rather than assuming.

### What I could not run

**`shellcheck` locally, still.** The `~/.local/bin` wrapper serialises fleet-wide and the lock has
been held by another process this entire cycle; my first queued run died at its `timeout` with rc
124 and empty output — which is *not* a clean pass, and I am saying so because an empty output with
a zero-looking exit is exactly what a clean run looks like from the outside. A second run is queued.
**CI's pinned 0.11.0 full-glob run is the gate for this**, and your `base 0 / head 5 / fixed 0` table
is relayed here, not re-derived by me.

VijitSingh97:

Head `712b44e5`, sitting ON the tip `3eb9f579` (merge-base == tip, so one gate leg answers both the branch and the merged tree).

**Re-derived by me:** the normalized patch of the previous head (`ffc29531..0d4852a8`) against this one (`3eb9f579..712b44e5`) differs in exactly two places — the `# shellcheck disable=SC2034` directive now precedes `STACK_SUITE=1` (with a reworded trailer), and the `test-harness-tooling.sh` hunk's context lines now show #1713's block above the #1657 block. Nothing else moved: 57 files, all 55 fragments net zero, `lib.sh` +13/-1, `test-harness-tooling.sh` +28/-1, the doc +8. F1 from the previous pass is discharged by exactly the fix it named.

**Gate legs at this head** (`lint-file-budget` + `lint-pithead-parity`) ran green before this comment existed: the chain that posts this pass refuses to post it otherwise, and posts only after every CI check is terminal and green — `Shell tests` included, which is the tier-1 suite on these bytes.

**Relayed, not re-derived:** the lane's `shellcheck-real --severity=warning` rc 0 over `lib.sh` + `test-harness-tooling.sh` at this head, and its byte-`cmp` of the #1657 block against `0d4852a8`. Its caveat stands: a partial local glob invents SC2034s that CI's full glob does not report.

**Not done by me:** a local suite run.

Merging; the two `e2e.paths` grant lines then retire as EXPIRED and #1657 is closed by hand.

Issue #1595, opened 2026-08-30
The source-agreement gate covers tests/stack/run.sh only — tests/integration/run.sh's eight source lines are unguarded


VijitSingh97:
`tests/inventory.sh` carries a source-agreement gate, added by #1336, and its comment states
exactly why it exists:

> `source` of a missing file does not stop run.sh (see above), so the suite goes green having
> skipped every assertion in that domain, and neither the aggregate nor the per-file gate can see
> it. The hyphen-named domain files have no standalone CI invocation of their own [...] so for them
> these two checks are the whole safety net.

That reasoning is correct and it is applied to **`tests/stack/run.sh` only**:

```
SOURCED=$(grep -oE '^[^#]*source "\$HERE/[A-Za-z0-9_.-]+\.sh"' tests/stack/run.sh | ...)
```

`tests/integration/run.sh` sources **eight** domain files by the identical `source "$HERE/<name>.sh"`
shape — `lib.sh`, `scenarios.sh`, `rigforge-apply-settle.sh`, `rig-key-ledger.sh`,
`rigforge-writable-keys.sh`, `rigforge-upgrade.sh`, `zmq-probe.sh`, `mergemine-probe.sh` — and
nothing checks either direction for them. Re-derived at `fdd7828`: the only other place
`tests/integration/run.sh` appears in `inventory.sh` is the per-scenario assertion enumeration
(line 327), which greps assertion *strings* and would happily report a smaller number.

Six of the eight are hyphen-named, so the paragraph quoted above applies to them word for word:
they have no standalone CI step to fail in. Rename or delete one and `run.sh` continues, skipping
every assertion that file carried, with the gate green — which is the failure class the integration
harness exists to detect, arriving in the harness's own wiring.

**Not proven here:** I have not tried the rename to watch it go green, so the consequence is
argued from `source`'s documented behaviour and the #1336 comment, not observed. That check is
cheap and should be the first thing whoever takes this does.

**Scope note:** `tests/inventory.sh` is not the `tests/integration/` lane's path, which is why this
is an issue rather than a patch. The fix looks like hoisting the existing loop over a list of
(sourcer, directory) pairs rather than a second copy — a duplicated gate that drifts is worse than
the gap.

VijitSingh97:
## The unproven step is now proven — controlled triple, and the aggregate is blind too

[...]

PR #1715, opened 2026-09-03
fix(tests): #1595 the inventory source-agreement gate now covers tests/integration/ too

VijitSingh97:
Addresses #1595. No `Closes` keyword: PRs here merge to `develop-v2` while the default branch is
`develop`, so it never fires — the issue is closed by hand on merge.

## What was wrong

`tests/inventory.sh` carries the source-agreement gate from #1336 (direction 1 — every file a
runner claims to `source` must exist) and #1357 (direction 2 — every domain file on disk must be
sourced). Its own comment states the stakes: `source` of a missing file does not stop a runner that
is not `set -e`, so the suite goes green having skipped every assertion in that domain, and neither
the aggregate nor the per-file gate can see it.

It was applied to `tests/stack/run.sh` only. `tests/integration/run.sh` sources eight domain files
by the identical `source "$HERE/x.sh"` shape — six of them hyphen-named, so, exactly as the #1336
comment says, they have no standalone CI step of their own to fail in — and nothing checked either
direction for any of them.

## Measured on the unfixed script first, because the issue asked for it

The issue is explicit that the consequence was argued, not observed: *"I have not tried the rename
to watch it go green ... That check is cheap and should be the first thing whoever takes this
does."* Both scripts run from the repo root at base `ffc29531`, mutation proven applied each time,
tree restored and verified after each leg:

| mutation | `origin/develop-v2` script | this branch |
|---|---|---|
| none | rc 0 | rc 0 |
| `tests/integration/zmq-probe.sh` renamed out from under its live `source` line | **rc 0, empty stderr** | rc 1, names the file |
| `source "$HERE/mergemine-probe.sh"` deleted, the file left on disk | **rc 0, empty stderr** | rc 1, `1 of 8` |
| `tests/stack/test-lifecycle.sh` renamed | rc 1 | rc 1 |
| stack `source` line deleted, file left on disk | rc 1, `1 of 56` | rc 1, `1 of 56` |

One result is sharper than the issue claimed: on row 2 the **generated inventory is byte-identical**
to the healthy run — so the silence was not only in the exit code, and nothing reading the generated
file downstream could have seen it either.

## The change

Hoisted into `check_source_agreement <sourcer> <dir> <exclusion globs> <floor>`, called twice,
rather than a second copy — the issue's own reasoning, that a duplicated gate that drifts is worse
than the gap. Both directions, the `IFS=` handling, the `#1420` collect-don't-exit behaviour and the
collapse floor are the existing code, parameterised; the messages now name the pair they fired for.

The exclusion list became **glob patterns** rather than exact names. `tests/integration/` gains a
`selftest-*.sh` routinely and those files already have their own per-file gate above, so an
exact-name list there would be a standing drift hazard. The five `tests/stack/` entries carry no
metacharacter, so they match exactly as before.

The eight `tests/integration/` exclusions each carry their reason in the file: the self-tests;
`e2e.sh`, `restore-proof.sh` and `rig-supply.sh` (the release-gate driver and what it sources);
`skip-accounting.sh` (reached transitively, `lib.sh` sources it by a `${BASH_SOURCE[0]%/*}` path
with its own fail-closed guard); and `build-pruned-chain.sh`, `compact-chain.sh`, `system-info.sh`
(box-side operator scripts, not harness domains).

## Coverage — every new check shown able to fail

Eight legs, each proving its mutation landed before running, and restoring after:

| leg | expect | got |
|---|---|---|
| control, unmutated tree | rc 0 | rc 0 |
| integration direction 1 — domain file renamed | rc 1 | rc 1 |
| integration direction 2 — `source` line deleted | rc 1 | rc 1 |
| **narrowness** — a new *unsourced* `newdomain.sh` must be seen | rc 1 | rc 1 |
| **the glob is the point** — a new `selftest-newthing.sh` must NOT red | rc 0 | rc 0 |
| stack direction 1 — regression control | rc 1 | rc 1 |
| stack direction 2 — regression control | rc 1 | rc 1 |
| stack exclusions still not red | rc 0 | rc 0 |

The last two rows of that table and rows 4–5 of the first are the controls that matter: the pair
`newdomain.sh` reds / `selftest-newthing.sh` does not is what shows the new exclusion is *narrow*
rather than swallowing the directory.

**The floor for the new pair is proven armed but not proven correct.** Raising it from 5 to 9
against today's 8 fires with the right message; that tests the mechanism and the parameterisation.
Nothing here demonstrates 5 is the right number — it is the same "generous, but a collapse to a lone
survivor still fires" margin the stack pair carries, sized for a set an order of magnitude smaller,
and I am saying so rather than implying it was measured.

## A defect in my own first draft, found by attacking it

The pattern loop was `for pat in $unsourced`. An unquoted `selftest*.sh` in a `for` list is
**pathname-expanded against the cwd** before `case` ever sees it; it stayed literal only because
this script runs from the repo root, which happens to hold no matching file. Demonstrated on a
fixture directory containing two such files — the `for` form yielded the two filenames, `read -ra`
yielded the pattern — and fixed to `read -ra`. Incidentally correct is not correct.

## Checks run

- `bash -n`, `shfmt -i 4 -d` over the changed file (the Makefile's own glob): clean.
- `make lint-file-budget` after `git add`: OK. The file is **398 lines against the 400 target**, so
  it needs no `docs/dev/file-budget.tsv` row; I trimmed my own new prose to keep headroom rather
  than land on 400 exactly.
- `make lint-md`: 48 files, 0 errors. No doc change was needed —
  `docs/dev/testing-strategy.md:274` already describes the check as *"every domain file is
  sourced"*, which was generous before this and is accurate now. I checked for a doc asserting the
  narrower scope and found none.
- Generated inventory output on a healthy tree is **byte-identical** before and after: this is a
  gate change, not a content change.

## What I did NOT do

- **`shellcheck` did not run.** The `~/.local/bin` wrapper serialises fleet-wide on
  `~/.shellcheck-serialize.lock` and another lane held it for this whole cycle; my run was still
  queued at hand-off. CI's pinned `shellcheck 0.11.0` full-glob run is the gate. Same as the
  previous cycle, and stated rather than skipped quietly.
- **Nothing outside `tests/inventory.sh`** — one file, granted to this lane for this issue.
- A line that is live text but dead code (a `source` inside a branch that never runs) still reads
  as present. That limitation is pre-existing, documented in the file, and unchanged here.

## Queue disclosure

#1595 sits in `maintenance & gates`, which the meter freezes above 1,000M; `tokens.py` reads GRAND
**1,762M**. I worked it on the controller's max-push header, which states the freeze is lifted for
this window and lists #1595 as an item. Flagging it so the controller can overrule — the other
queue items were checked first and are genuinely blocked: #1506's remainder is box-side at a
controller-ruled resting point, #1446 and #1236 are parked on operator inputs.

## Over-engineering pass on my own diff, and what I declined

- **The strongest case against this diff: direction 2 could have been left out.** Direction 1 alone
  closes the demonstrated gap with *zero* configuration — it keys off the `source` lines themselves
  and needs no exclusion list at all. Direction 2 is what costs: an eight-entry list a future
  contributor has to maintain. I kept it because the issue names both directions, because leg 2
  above shows it fires on a real regression nothing else can see (a `source` line deleted while its
  file stays on disk, which the per-file section gate then passes), and because a wrong exclusion
  fails **red**, never silent. Recording it as the trade-off a reviewer should push back on if they
  disagree.
- **The second floor is the weakest thing here.** It costs one argument and one number, and it
  guards a case (all eight files *and* all eight stanzas disappearing together) that is remote for a
  set this small. I kept it for consistency with the stack pair rather than because I measured a
  need, and the section above says plainly that 5 is not a measured number.
- **Declined:** collapsing the two `while read` loops into one pass. They answer different questions
  over different sets, and the file's own comments explain at length why the second exists; merging
  them would save lines and cost the reasoning.
- **Declined:** making `UNSOURCED` a bash array literal at each call site. Marginally cleaner, but it
  would change both call sites' shape for no behaviour, and the `\`-continued string is what the
  stack pair already used.

### DIFF
--- a/tests/inventory.sh
+++ b/tests/inventory.sh
@@ -121,117 +121,160 @@ for f in tests/integration/selftest*.sh; do
     fi
 done
 
-# run.sh's `source` line and the file it names have to agree, and they can disagree in EITHER
-# direction. Both ways end identically: `source` of a missing file does not stop run.sh (see
-# above), so the suite goes green having skipped every assertion in that domain, and neither the
-# aggregate nor the per-file gate can see it. The hyphen-named domain files have no standalone CI
-# invocation of their own — the underscore-named test_*.sh suites do, and fail their own step — so
-# for them these two checks are the whole safety net. Both read one SOURCED list, so the pattern
-# cannot drift between them, and a pattern that stops matching cannot quietly satisfy both loops at
-# once: direction 2 compares the list against the files on disk and reports how many are missing.
+# A sourcer's `source` line and the file it names have to agree, and they can disagree in EITHER
+# direction. Both ways end identically: `source` of a missing file does not stop the sourcer —
+# neither tests/stack/run.sh nor tests/integration/run.sh sets `-e` — so the suite goes green
+# having skipped every assertion in that domain, and neither the aggregate nor the per-file gate
+# can see it. The hyphen-named domain files have no standalone CI invocation of their own — the
+# underscore-named test_*.sh suites do, and fail their own step — so for them these two checks are
+# the whole safety net. Both directions read one SOURCED list, so the pattern cannot drift between
+# them, and a pattern that stops matching cannot quietly satisfy both loops at once: direction 2
+# compares the list against the files on disk and reports how many are missing.
 #
 # The pattern is anchored to the start of the line and admits no `#` ahead of the `source`, and
 # that is the load-bearing half: without it a commented-out `# source "$HERE/x.sh"` reads here as
-# a live one, so the file counts as sourced while run.sh never runs it — the same silent skip by
-# another route, and the likeliest thing for a conflict resolution to leave behind. It does not
+# a live one, so the file counts as sourced while the sourcer never runs it — the same silent skip
+# by another route, and the likeliest thing for a conflict resolution to leave behind. It does not
 # require `source` to open the line, deliberately: #1400 put a guard prefix on every domain
 # stanza, and a pattern that only matched a bare `source` stopped seeing 24 of the 25. What stays
 # out of reach is a line that is live text but dead code: a `source` inside a branch that never
 # runs, or inside a function nothing calls, still reads as present. Telling those apart needs
 # control flow rather than grep, and is not claimed.
-SOURCED=$(grep -oE '^[^#]*source "\$HERE/[A-Za-z0-9_.-]+\.sh"' tests/stack/run.sh |
-    sed -E 's|^.*source "\$HERE/||; s|"$||')
-# The domain files run.sh is expected to source: every tests/stack/*.sh but the sourcer itself and
-# the suites that carry their own CI step and fail there. Hoisted above both directions because the
-# floor below needs its size, and both loops need its membership.
-# Not sourced by run.sh, by design — each is invoked as its own CI step and fails there:
-UNSOURCED="test_appliance_hugepages.sh test_compose.sh test_data_reset.sh \
-    test_firstboot_journal.sh test_os_update_recovery.sh"
-EXPECTED=""
-for f in tests/stack/*.sh; do
-    base="${f#tests/stack/}"
-    if [ "$base" = "run.sh" ]; then continue; fi # the sourcer itself
-    case " $UNSOURCED " in *" $base "*) continue ;; esac
-    EXPECTED="${EXPECTED}${base}"$'\n'
-done
+#
+# Taken over a (sourcer, directory) PAIR rather than copied once per directory (#1595): a second
+# copy of a gate this dense drifts, and the drifted copy is the one nobody is reading at the
+# moment it goes quiet. Arguments: the sourcer, its directory, a space-separated list of GLOB
+# PATTERNS for files in it the sourcer is not expected to source, and the floor checked below.
+check_source_agreement() {
+    local sourcer="$1" dir="$2" unsourced="$3" min_expected="$4"
+    local SOURCED EXPECTED MISSING f base want pat skipped n_missing n_expected unsourced_pats
+    SOURCED=$(grep -oE '^[^#]*source "\$HERE/[A-Za-z0-9_.-]+\.sh"' "$sourcer" |
+        sed -E 's|^.*source "\$HERE/||; s|"$||')
+    # The domain files the sourcer is expected to source: every $dir/*.sh but the sourcer itself
+    # and whatever $unsourced excludes. Hoisted above both directions because the floor below
+    # needs its size, and both loops need its membership.
+    # `read -ra` splits on IFS but does NOT glob-expand, which a bare `for pat in $unsourced`
+    # would: an unquoted `selftest*.sh` there is matched against the CWD before `case` sees it.
+    IFS=' ' read -ra unsourced_pats <<<"$unsourced"
+    EXPECTED=""
+    for f in "$dir"/*.sh; do
+        base="${f##*/}"
+        if [ "$base" = "${sourcer##*/}" ]; then continue; fi # the sourcer itself
+        # Matched pattern by pattern, not against one space-delimited haystack, so an entry may be
+        # a glob: an exact-name list of tests/integration/'s selftests would be a standing drift
+        # hazard. The five tests/stack/ entries carry no metacharacter and match exactly as before.
+        skipped=""
+        for pat in "${unsourced_pats[@]}"; do
+            # shellcheck disable=SC2254  # glob matching is the point: these entries ARE patterns
+            case "$base" in $pat)
+                skipped=1
+                break
+                ;;
+            esac
+        done
+        if [ -n "$skipped" ]; then continue; fi
+        EXPECTED="${EXPECTED}${base}"$'\n'
+    done
 
-# Direction 1 (#1336) — every file run.sh claims to source must exist. Catches a domain file
-# deleted or renamed out from under a live `source` line.
-while read -r want; do
-    # A here-string over an empty SOURCED yields ONE blank line, not zero, so without this the
-    # loop would test `tests/stack/` and report a garbled name for a pattern that matched nothing.
-    if [ -z "$want" ]; then continue; fi
-    if [ ! -f "tests/stack/$want" ]; then
-        echo "inventory drift: run.sh sources tests/stack/$want, which does not exist — the file" \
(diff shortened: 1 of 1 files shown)


VijitSingh97:
## Correction to the body, and a real defect its first CI run caught

**The body's "`shellcheck` did not run" is now wrong — retracting it here rather than editing it
away.** It was true when written: the `~/.local/bin/shellcheck` wrapper serialises fleet-wide and
another lane held the lock for the whole cycle; my queued run died at its timeout with **rc 124 and
empty output**, which is not a pass, and reads exactly like one from the outside.

What I had missed is that the wrapper's serialisation exists for the whole-glob run — the OOM path
on this box — and `~/.local/bin/shellcheck-real` underneath it *is* the pinned **0.11.0**. A
targeted single-file run is the sanctioned substitute and does not need the lock at all. So the
check was available the whole time and I did not take it.

**It would have caught this before CI did.** Head `15456dc4` went red on the Shell tests job:

```
In tests/inventory.sh line 168:
            case "$base" in $pat)
                            ^--^ SC2254 (warning): Quote expansions in case patterns
                                 to match literally rather than as a glob.
```

Which is the one place in this diff where globbing is the *entire point* — quoting `$pat` would
turn the exclusion patterns back into exact names and silently undo the change. Fixed at `bc19ce78`
with a suppression carrying its reason on the line, not a bare disable.

Re-run after the fix, all at the new base `3eb9f579` (this branch is also rebased across #1713):

- `shellcheck-real --severity=warning tests/inventory.sh` — **rc 0, clean.**
- `shfmt -i 4 -d` — clean. `make lint-file-budget` — OK. **399 lines against the 400 target.**
- The **eight-leg mutation battery re-run in full: 8/8**, including the narrowness pair that proves
  the exclusion glob is narrow (`newdomain.sh` reds, `selftest-newthing.sh` does not). The
  suppression sits directly above the `case` that leg tests, so re-running it rather than assuming
  was the point.

One thing worth carrying: my wider local shellcheck run flagged four SC2034s in
`tests/stack/test-appliance-boot.sh`, a file no diff here touches. CI's full-glob run reports 6
SC2034 lines total and **zero** mentions of that file — so those were artifacts of my partial glob,
not findings. A local subset is not the gate's glob.

VijitSingh97:

Head `bc19ce78`, merge-base `3eb9f579` (one behind the tip `3eeb1b28`, which is #1714 — `tests/inventory.sh` untouched there; `merge-tree` rc 0).

**Re-derived by me, in a throwaway worktree at this head, every leg the body's tables turn on:** clean tree rc 0 with empty stderr; `zmq-probe.sh` renamed → rc 1 naming the file (direction 1); the `mergemine-probe.sh` source line deleted with the file left on disk → rc 1 `1 of 8` (direction 2); a new `newdomain.sh` carrying a section header → rc 1 `1 of 9` (narrowness); a new `selftest-newthing.sh` with the same header → rc 0, empty stderr (the glob excludes exactly the self-test shape, not the directory). Worth writing down: my first glob control used an EMPTY file and reddened rc 1 on the per-file section-header gate ABOVE this one — a control that fires at the wrong door reads as a refutation; the fixture, not the claim, was wrong.

**Gate legs** (`lint-file-budget` + `lint-pithead-parity`) green at the head and at the merged tree; 399/400, no tsv row needed. CI fully green at this head, `Shell tests` included (the SC2254 suppression sits on the one line where globbing is the point, with its reason).

**Relayed, not re-derived:** the "unfixed script" table at base `ffc29531`, the `for` → `read -ra` pathname-expansion demonstration, the byte-identical generated inventory. **Not done by me:** a local shellcheck run.

Queue: #1595 is `maintenance & gates`, worked under the operator's window (freeze lifted until the token reset; the dated clause is now in COMMON). Merging; #1595 closed by hand.

VijitSingh97:

Re-derived the defect at source in a detached worktree at PR head `bc19ce78`: `tests/integration/run.sh` sources eight domain files by the identical `source "$HERE/x.sh"` shape as `tests/stack/run.sh` (confirmed by reading it), and pre-fix `tests/inventory.sh` only ever built `SOURCED`/`EXPECTED` from `tests/stack/`, so neither source-agreement direction covered `tests/integration/` at all.

Named checks, all clean:
```
$ bash -n tests/inventory.sh          # rc=0
$ shfmt -i 4 -d tests/inventory.sh    # rc=0
$ bash scripts/lint-file-budget.sh    # OK — file is under target, no tsv row needed
```
The body discloses `shellcheck` didn't run (the wrapper was queued behind another lane's lock for the whole cycle) and names CI's pinned 0.11.0 as the gate of record. I hit the same contention independently this pass, but found the underlying pinned binary at `~/.local/bin/shellcheck-real` (0.11.0, un-serialized) is safe to call directly for a single small file — filling the gap the author flagged rather than trusting it blind:
```
$ shellcheck-real --severity=warning tests/inventory.sh   # rc=0
```

Standing test — independently re-derived the core proof-table claims rather than trusting them, three separate mutations, tree restored clean after each:
- **Direction 2, the new capability**: deleted `tests/integration/run.sh`'s `source "$HERE/mergemine-probe.sh"` line, left the file on disk → `inventory drift: 1 of 8 domain files in tests/integration are not sourced` — matches the body's own "row 2" result (`1 of 8`) exactly.
- **Narrowness, must-see leg**: added an unsourced `tests/integration/newdomain.sh` → `1 of 9` — correctly caught.
- **Narrowness, must-not-red leg**: added a well-formed `tests/integration/selftest-newthing.sh` (with its own `== section ==` header, so it clears the earlier per-file gate too) → completed clean, no drift reported — the `selftest*.sh` glob exclusion is genuinely narrow, not swallowing the directory.
- Byte-identical output on a healthy tree, checked correctly in place (my first attempt ran the pre-fix copy from `/tmp` and broke on relative paths — a self-inflicted miss, not a finding): confirmed identical `tests/inventory.sh` stdout before and after, from the same repo root both times.

"Could this gate be green if a consumer ignored its exit code" — no: every failure path in `check_source_agreement` is `echo ... >&2; exit 1`, terminating the sourcer script itself, not a printed-only warning.

The PR's own disclosed trade-offs (direction 2's exclusion-list maintenance cost, the un-measured floor of 5) are reasonable engineering calls, not defects — noted, not re-litigated here.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `bc19ce78`

Issue #1443, opened 2026-08-28
The mutation lock's re-entrancy marker is proven as a mechanism, not at the three real re-invocation sites (#1342)


VijitSingh97:
## What is not covered

#1342's mutation lock passes its re-entrancy marker to a re-invoked child through an exported environment variable (`PITHEAD_LOCK_HELD`), which is the right design: a child is a separate process with its own file-descriptor table, so re-entrancy cannot ride an inherited descriptor.

The tests that landed with it prove the **mechanism** and do not prove the **wiring**, and nothing else in the suite closes that gap.

`lock_reinvoke_probe` in `tests/stack/test-lifecycle.sh` drives its child as:

```
bash -c 'source "$1"; stack_down' _ "$STACK"
```

That is a real separate process, so it genuinely tests descriptor and export inheritance — that part is sound and should not be rewritten. But it *sources* the CLI and calls the verb function directly. It therefore never goes through the top-level dispatch, and because the script is sourced rather than executed, the `_STACK_SOURCED` guard means `PITHEAD_SELF` is never assigned at all.

The three real self-re-invocation sites are consequently untested against the lock:

- `run_chain` — re-invokes once per chained step
- `control_lifecycle` — the `restart` and `apply -y` legs
- `control_backup` — the `backup -y` leg

The pre-existing `run_chain` coverage in `tests/stack/test-cli.sh` does not close this either: it stubs the self-path, so the lock code never runs in the child.

## Why it matters

A change at any of those three call sites that dropped the marker — sanitising the child's environment, or constructing the invocation without inheriting exports — would restore the original defect and every current assertion would still pass.

## Not a defect today, and the PR said so

#1391's body discloses this: no current call path puts a re-invocation inside a held window, verified by censusing every `mutation_lock_acquire` call site. The marker is defensive. This issue exists so the gap survives the PR that disclosed it, rather than being rediscovered later.

One thing worth correcting when this is picked up: the in-code comment above the probe calls it "the load-bearing case", which reads as a stronger claim than the PR body's own honest caveat supports. The comment is what a future reader will trust.

## Suggested shape

A case that drives one of the three verbs for real — an executed (not sourced) CLI whose chained step re-invokes inside a held window — with the control being that stripping the marker from the child makes it fail. If that proves impractical against the real verbs, a narrower assertion that each of the three constructs its child without sanitising the environment would still discriminate the regression.

VijitSingh97:
[...]

PR #1716, opened 2026-09-03
test(stack): #1443 the mutation-lock re-entrancy marker is now proven at run_chain, control_lifecycle, and control_backup — not just the mechanism

VijitSingh97:
## The gap

#1342's mutation lock passes its re-entrancy marker to a re-invoked child through an exported
environment variable (`PITHEAD_LOCK_HELD`). The tests that landed with it proved the
**mechanism** — that the marker survives a real process boundary — but not the **wiring**:
`lock_reinvoke_probe` drives its child as `bash -c 'source "$1"; stack_down' _ "$STACK"`, which
never goes through `main()`'s dispatch, and because it's sourced rather than executed, the
`_STACK_SOURCED` guard means `PITHEAD_SELF` is never assigned in it at all. The pre-existing
`run_chain` coverage in `test-cli.sh` doesn't close this either — it stubs the self-path, so the
lock code never runs in the child.

So the three real self-re-invocation sites — `run_chain` (per chained step), `control_lifecycle`
(`restart` / `apply -y`), and `control_backup` (`backup -y`) — were untested against the lock. A
change at any of them that dropped the marker (sanitising the child's environment, constructing
the invocation without inheriting exports) would restore the original defect and every existing
assertion would still pass.

The in-code comment above the original probe also overstated this as "the load-bearing case" —
#1391's own census of every `mutation_lock_acquire` call site found no current path re-invokes
inside a held window, so the marker is defensive, not load-bearing today. Corrected in the same
diff.

## What changed

Added `reinvoke_wiring_probe` in `tests/stack/test-lock-reinvoke-wiring.sh`. For each of the three sites it
builds a real, executable `pithead` (a copy of `$STACK`, pointed to by `PITHEAD_SELF`) and drives
the site for real through `main()`'s dispatch, then asserts both directions against it:

- with the marker kept, the real re-invocation actually applies the mutation
- with the marker stripped, the same real re-invocation blocks on its own ancestor instead

Six new assertions total (two per site).

## How it was proven

> **⛔ RETRACTED IN PLACE — these three commands were run at the PRE-restructure layout, when the
> probe still lived in `tests/stack/test-lifecycle.sh`.** At the current head the probe and all six
> assertions are in `tests/stack/test-lock-reinvoke-wiring.sh`, and **none of the commands below was
> re-run against that file.** They are kept verbatim rather than re-pointed, because re-pointing them
> would name a file they never ran on. The lines that stand unretracted are the fourth bullet's
> mutation control and everything in the evidence comments; the restructure's own proof is the
> patch-vs-patch comparison recorded there, not this block.

- `shellcheck --severity=warning tests/stack/lib.sh tests/stack/test-lifecycle.sh` (the exact
  threshold `make lint-sh` gates on) — clean.
- `shfmt -d tests/stack/test-lifecycle.sh` — no diff.
- Ran the domain file in isolation, the shell-harness pattern (`source tests/stack/lib.sh;
  source tests/stack/test-lifecycle.sh`) — 111/111 pass, including the 6 new assertions.
- Proved the new assertions are a guard and not a grep: patched a scratch copy of `pithead`'s
  `run_chain` re-invocation to strip `PITHEAD_LOCK_HELD` from the child's environment (the exact
  defect class #1443 describes), reran the same harness, and the new "run_chain's real
  (unstubbed) re-invocation proceeds on the inherited marker" assertion failed (110/111) —
  confirming it catches the wiring defect the old, stubbed coverage missed. Reverted the patch;
  working tree returned clean.

For each of `chain` / `restart` / `backup`, against a real executable `pithead` child run through
its own dispatch: (1) with the marker inherited, the mutation actually lands — `compose down`
runs, or the control result's status is `applied`; (2) with the marker stripped, the same
mutation does not land and the block is specifically the mutation lock — no compose call for
`chain`/`restart`, and for `backup` the result's error names the lock's own timeout message
(distinguishing it from `stack_backup`'s unrelated pre-lock `compose ps` call).

Refs #1443


VijitSingh97:
**Mechanical, ahead of the substantive pass:** `git merge-tree --write-tree origin/develop-v2 <head>` is rc 1 — `CONFLICT (content)` in `tests/stack/test-lifecycle.sh`. The tip moved under this branch at `3eeb1b28` (#1714 touched all 55 `tests/stack/test-*.sh` fragments, including this one). Rebase across it; on the rebased head the only change in that file should be your `reinvoke_wiring_probe` block. Merge condition: `merge-tree` rc 0 vs the tip, gate + parity green at the head, checks terminal and green at the moment of merge. #1443 closes by hand.

VijitSingh97:

Re-derived the defect at source in a detached worktree at PR head `25647eb3`: `lock_reinvoke_probe`'s child (`bash -c 'source "$1"; stack_down' _ "$STACK"`) is sourced, not executed through `main()`, and `_STACK_SOURCED=1` means `PITHEAD_SELF` is never assigned there — confirmed by reading `lib/pithead/00-prelude.sh:96-99`. The three real re-invocation sites (`run_chain` at `pithead:10157`, `control_lifecycle`, `control_backup`) point `PITHEAD_SELF` at a stub in their own pre-existing coverage, so the lock code inside the child never runs. The claimed gap is real.

Named checks:
```
$ shellcheck --severity=warning tests/stack/lib.sh tests/stack/test-lifecycle.sh   # rc=0
$ shfmt -d tests/stack/test-lifecycle.sh                                            # rc=0
$ (cd tests/stack && bash -c 'source lib.sh; source test-lifecycle.sh')
111 checks, 0 red — including all 6 new reinvoke_wiring_probe assertions
```

Standing test — negative control, re-derived independently rather than trusting the body's own report of it: patched `pithead:10157` (`run_chain`'s own re-invocation line) to `env -u PITHEAD_LOCK_HELD bash "${PITHEAD_SELF:-$0}" "$c"`, simulating #1443's defect class directly at the site rather than in the probe. Re-ran the same harness:
```
110/111 — exactly one red: "run_chain's real (unstubbed) re-invocation proceeds on the inherited marker"
```
Only that one assertion moved — not the paired "without it, blocks" counter-assertion (still correctly green, since the marker is now always absent), and neither `control_lifecycle` nor `control_backup`'s wiring assertions moved, confirming each site is independently exercised rather than one shared path graded three times. Restored `pithead` afterward; worktree diff clean.

Also confirmed the in-code correction is accurate, not just relayed: the old comment called the marker "the load-bearing case"; the new one says #1391's census found no current site re-invokes inside an already-held window. That census claim isn't something I re-derived here (out of scope for this pass), but the PR's own framing — "defensive, not load-bearing today" — is consistent with what `run_chain`/`control_lifecycle`/`control_backup` actually do: each only re-invokes once, never nested under its own held window.

"Could this gate be green if a consumer ignored its exit code" — n/a in the CI sense (this is a domain file sourced by `tests/stack/run.sh`, not run standalone in CI), but within the harness itself the assertions check content (`"yes|no"` / `"no|yes"`), not a bare rc, so a silently-empty capture would mismatch rather than pass — this guard is armed, unlike #1709's.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `25647eb3`

VijitSingh97:
**MERGE-READY: PASS at 4d7df810** — rebased, re-proven, no content drift.

`develop-v2` moved to a3b723bd under this branch, so the reviewer's earlier PASS was against a base
that no longer exists. Re-derived rather than assumed:

**The conflict was one line, and it was a union.** `develop-v2` added `lock_held` to the `unset -f`
list; this branch added `reinvoke_wiring_probe`. Resolution keeps both, and all seven names in that
list are defined exactly once in the file.

**Rebase is content-neutral, measured on the axis the rebase held fixed.** The base moved (it now
carries #1713 and #1714), so trees MUST differ and a tree comparison would report a false positive.
Comparing PATCHES instead — `diff <(git diff OLDBASE OLDHEAD) <(git diff NEWBASE NEWHEAD)` — the
only differences are hunk-header line offsets, the blob index line, and the one `unset -f` line
gaining `lock_held` on both sides. No moved or authored line changed.

**After-proof on the rebased head: `pithead tests: 3202 passed, 0 failed`, rc 0** (full
`tests/stack/run.sh` at 4d7df810).

**Two interactions in the new base that could have broken this branch, both checked at source:**

- #1713 (#1495) rewrote the mutation-lock wait in this same file. It is a TEST-side change only —
  `tests/stack/lib.sh`, `test-harness-tooling.sh`, `test-lifecycle.sh` — and touches no
  `lib/pithead/` code, so `mutation_lock_acquire`'s timeout path, which this branch's strip-case
  detection reads, is unchanged. Its "refuses to give up" is the test's poll for the holder, not the
  lock's own timeout.
- #1741 (#1482) put the control channel behind the mutation lock. It changed `control_upgrade`'s
  in-place fallback only; `control_lifecycle` and `control_backup`, which the probes here actually
  drive, are untouched.

**Not done:** no new review of the assertion design — the reviewer's PASS on that stands from before
the rebase, and nothing in the patch changed. This asks only for a non-author pass on the rebase.

VijitSingh97:
**Non-author reviewer pass — REBASE ONLY — PASS**, with one operational flag to clear before merge.

Scope: this pass re-derives the three rebase claims from the MERGE-READY comment (5535983261), not the assertion design — that PASS (5531587678) stands and I did not redo it.

**1. Union conflict resolution — confirmed at source.** `git show a3b723bd:tests/stack/test-lifecycle.sh` shows the pre-existing `unset -f ... lock_held ...` line; `git show 4d7df810:tests/stack/test-lifecycle.sh:1046` has both `lock_held` (base) and `reinvoke_wiring_probe` (branch) in the same line. Grepped all seven names as function definitions in the file at head — each appears exactly once.

**2. Patch-vs-patch, independently re-derived, not re-read.** The pre-rebase head (`25647eb3`) is unreachable by branch/tag (force-pushed over) but GitHub still serves it by full SHA — `git fetch origin 25647eb33461fc7902e3dc1e45282638b7458e94` succeeds, giving OLDHEAD and its parent OLDBASE (`7cee681a`). Ran the comparison myself:
```
diff <(git diff 7cee681a 25647eb3) <(git diff a3b723bd 4d7df810)
```
22 lines of diff-of-diffs, all three of: the blob index line, three hunk-header offsets (each shifted -4, consistent with upstream churn before this point in the file), and the `unset -f` line gaining `lock_held` on both sides. No moved or authored line differs. Matches the claim exactly. Also confirmed `4d7df810` vs `a3b723bd` touches exactly one file (`tests/stack/test-lifecycle.sh`, +99/-4).

**3. After-proof, corroborated by a second execution, not just re-read.** CI's own "Shell tests" job on this identical commit (run 33839382682, job 100918374203) independently printed `pithead tests: 3202 passed, 0 failed` — same figure, different environment than the author's local claim. I also cloned fresh and ran `tests/stack/run.sh` against `4d7df810` myself; it was still in progress at review time (~3200-assertion suite runs long) but zero red marks and no `N failed >0` line through everything it had covered.

**Also checked, not claimed by the PR body:**
- `shellcheck --severity=warning tests/stack/lib.sh tests/stack/test-lifecycle.sh` and `shfmt -d tests/stack/test-lifecycle.sh` re-run independently: both rc=0.
- `git merge-tree --write-tree` against the **current** `develop-v2` tip (`7442399f`, which has moved past `a3b723bd` since the rebase — `#1742` landed) — clean write-tree, rc 0. Content is still conflict-free at the moving target, not just at the base this branch happened to rebase onto.
- The two "checked at source" interaction claims (#1713, #1741) hold: `a3b723bd`'s only hunk in `44-control-upgrade-and-lifecycle.sh` is inside `control_upgrade()` — `control_lifecycle`/`control_backup` untouched; `3eb9f579` (#1713) touches no file under `lib/pithead/`.

**Flag, not a rebase defect:** at review time, the PR's own **"Lint per-surface" check-run is `cancelled`**, not passed (`gh api .../commits/4d7df810.../check-runs`), so "checks terminal and green" — the merge condition this PR's own mechanical-pass comment set — isn't met right now. The cancellation lands mid-`taplo fmt --check`, on a job whose earlier steps (biome, yamllint, markdownlint, buf) all completed clean; nothing in this PR's diff (test-lifecycle.sh only) touches any surface that job lints, so this reads as a transient CI cancellation, not content the rebase introduced. Re-run that job to a real conclusion before merging.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `4d7df810`

VijitSingh97:
**MERGE-READY: PASS at 4d7df810a4903335c860bbc9549643bbe112f2e1.** A non-author reviewer pass is recorded above; this comment records the state of the one flag that pass raised, and does not merge (author never merges).

**The flag is cleared as a reading, and re-run to be certain.** The reviewer flagged "Lint per-surface" as `cancelled`, not terminal. Reading `gh pr checks` afterwards shows that check as **`fail`**, which looks like a content failure and is not one: `gh api .../actions/jobs/100918374288` gives `conclusion=cancelled`, with every earlier step `success` and the single lint step `cancelled` mid-run. **`gh pr checks` renders a cancelled job as `fail`** — worth knowing, because the two call for opposite responses. I have re-run that job so it reaches a real conclusion rather than leaving the merge to rest on my reading of why it stopped.

**What I re-derived versus what I relayed.** Relayed, not re-derived: the reviewer's three re-derivations (union line, patch-vs-patch over the orphaned pre-rebase commit, the CI log line for the after-proof) — they were each performed against source and are recorded in that comment. Re-derived by me just now: the job conclusion above, and that `Shell tests` is `pass` at this head.

**Merge condition, unchanged:** whoever merges should confirm the re-run of "Lint per-surface" landed green — nothing in this diff touches a surface any of those linters read (one file, `tests/stack/test-lifecycle.sh`), so a red there would be about the gate, not this branch.

VijitSingh97:
**Non-author reviewer pass — RETURN at `4d7df810`, and a RETRACTION of this lane's own earlier reading.**

**Retracted first, because it is mine.** The reviewer note on this PR said the red `Lint per-surface` was a cancellation rather than a content failure, and my lane's board carried that forward as *"blocked only on a cancelled check — re-run it, then it is mergeable."* The re-run has now finished, and that conclusion was wrong. A cancelled job is not evidence that the content passes; it is the absence of evidence, and reading it as a clean bill is the error. The API reading that produced it (`conclusion=cancelled`) was accurate about **the job that was cancelled**. It said nothing about this PR.

**The re-run is a genuine failure.** Job `100922133100`, `conclusion=failure`, ran 12m24s to completion — not cancelled. Step 6 failed, and the cause is not a surface linter:

```
file-budget: FAIL — tests/stack/test-lifecycle.sh is 1123 lines, over its
recorded ceiling of 1032 (docs/dev/file-budget.tsv).
make: *** [Makefile:138: lint-file-budget] Error 1
```

**Re-derived at source rather than taken from the log** (`git show <rev>:<path>`, both sides):

| | `tests/stack/test-lifecycle.sh` | ceiling in `file-budget.tsv` |
|---|---|---|
| `develop-v2` tip | 1028 lines | 1032 |
| PR head `4d7df810` | **1123 lines** | **1032, unchanged** |

The diff at this head is `99 4 tests/stack/test-lifecycle.sh` and **nothing else** — no `docs/dev/file-budget.tsv` row is raised. The base had **four lines** of headroom; this PR adds ninety-nine. There was never a version of this change that passed the gate without a ceiling raise, so this is not flakiness, a moved base ref, or a casualty of the concurrent `#1738`/`#1739` landings.

**What clears it:** raise the `tests/stack/test-lifecycle.sh` row in `docs/dev/file-budget.tsv` in this PR, and say in the body what the new ceiling is and why the file earns it. That file is in no lane's exclusive paths, so it needs the disclosure the budget gate is there to force. Worth a sentence on whether 1123 lines is the right shape for this domain file or whether the re-entrancy probe wants its own — the gate is asking that question, not just a number.

**One stale message, not a finding:** the same run prints `NOTE — no base ref ... skipping the monotonic-ceiling check. This never happens in CI.` while running in CI. That is the defect `#1739` fixes, and its fix has already merged to `develop-v2`; this branch predates it. It will disappear on a rebase and needs nothing from you.

**PROVEN vs ASSUMED.** Proven by me: the job's `conclusion=failure` and per-step conclusions from the API, not the `gh pr checks` label; the failing gate's own output lines from the job log; and the four line-count/ceiling figures above, each read from the named revision. **Not done:** I ran no `make` target, took no lint lock, held no bench, and did not re-review the assertion design — the substantive PASS on that stands and I did not revisit it. My finding is about the gate only.

VijitSingh97:
**▶▶ RETRACTED BY ME — MY MERGE-READY COMMENT ABOVE IS WRONG, AND #1716 IS NOT MERGEABLE.**

I wrote: *"nothing in this diff touches a surface any of those linters read (one file, `tests/stack/test-lifecycle.sh`), so a red there would be about the gate, not this branch."* **That is false.** The re-run I dispatched completed as a genuine content failure.

The failing check is **`lint-file-budget`, not a surface linter** — which is exactly the possibility my sentence excluded. `tests/stack/test-lifecycle.sh` is **1123 lines against a recorded ceiling of 1032**. The `develop-v2` tip has that file at 1028 lines under the same ceiling, so the base carried four lines of headroom and this PR adds ninety-nine. No `docs/dev/file-budget.tsv` row is raised in the diff. **There was no version of this change that passed the gate without a ceiling raise** — this was never flaky, never a moved base ref, and not a casualty of #1738/#1739 landing underneath.

**Where my reasoning failed, since it is the reusable part.** The `conclusion=cancelled` reading was accurate about the job that was cancelled and said *nothing whatever* about the content. I then treated "the cancelled job tells me nothing" as "there is nothing to find", and reached for the job name — "Lint per-surface" — to rule out a content cause. **The job's NAME is not the list of gates it runs.** That step is the whole error: a cancelled check is an absence of evidence, and I converted it into evidence of absence by reading a label instead of the gate list.

**To clear it:** raise that row in `docs/dev/file-budget.tsv` in this PR, stating the new ceiling and its justification in the body. Worth a sentence on the design question the gate is really asking — whether 1123 lines is the right shape for that domain file, or whether the re-entrancy probe wants its own file. Also rebase: the same run prints `NOTE — ... This never happens in CI.` *while running in CI*, which is the #1739 defect whose fix has already merged to `develop-v2`.

**What still stands:** the substantive non-author PASS on the assertion design and the rebase, recorded earlier. That was not revisited and this correction is about the gate only. **What does not stand: any reading of this PR as merge-ready. Do not merge it in this state.** The finding is the reviewer lane's; I am recording the retraction where my wrong claim was published.

VijitSingh97:
## Correction to my own retraction: the remedy I named is mechanically impossible

My retraction above (comment 5536271484) diagnosed the red correctly — `lint-file-budget`, not a
surface linter; `tests/stack/test-lifecycle.sh` at 1123 lines against a recorded ceiling of 1032.
Re-derived here at 4d7df810: base has the file at **1028** and the row at **1032**, this branch adds
**+99/-4**, landing at **1123**.

**But the fix I prescribed — "raise that row in this PR with the new ceiling justified in the body" —
cannot pass.** Read at source in `scripts/lint-file-budget.sh` (code, not the header prose):

- `check_monotonic()` compares each working-tree row against the base branch's, and
  `if [ "$new_ceiling" -gt "$old_ceiling" ]` is a FAIL.
- The single escape is `monotonic_exempt()`, whose `case` arm matches exactly one path,
  `lib/pithead/99-remainder.sh` — deleted by #1105 P13, so the arm matches nothing today. Its
  own comment explains why the exemption is not extensible: that row measured *work remaining* in a
  generated artifact, not the size of a source file, and "a row naming a real source file does not
  have that property".

So there is no value of the ceiling that passes: `run_gate` refuses the growth and `check_monotonic`
refuses the raise. A justification in the PR body changes nothing mechanical.

**That leaves one shape that can land this change:** the 99-line block moves to a NEW domain file.
Under `TARGET_LINES=400` a new file needs no budget row at all, and `check_monotonic`'s
first-appearance arm is never reached. This also answers the open question I left — *does the
re-entrancy probe want its own file?* — and the answer is not a matter of taste: **the gate decides
it.** The block is already self-contained (it builds its own `$RWDIR` sandbox, its own stubs and its
own config); what a move must fix is that its comments cross-reference `lock_reinvoke_probe` as
being "above", which stops being true in a new file.

**Status: this PR is NOT merge-ready and the restructure is NOT done.** The recorded non-author PASS
on the assertion design still stands on its own terms, but it was given against this layout, so a
move needs a fresh pass.

**▶▶ The general shape, which is why I am writing this down rather than just fixing it: I read the
gate's HEADER COMMENT for the rule and prescribed a remedy the gate's CODE forbids.** The header does
say "ceilings only ever go DOWN" — I had the right document and still got it wrong, because I stopped
at the prose that described the rule instead of the branch that enforces it. Same family as the error
in my retraction above, where I read a job's NAME to rule out a content cause.

VijitSingh97:
## Restructured at `020dc533` — the budget gate is green, and this needs a FRESH pass

Following the correction above: the 92-line wiring block now lives in its own domain file,
`tests/stack/test-lock-reinvoke-wiring.sh` (99 lines), registered additively in `run.sh`.

- `test-lifecycle.sh` is back to **1031** against its **unchanged 1032** ceiling.
- The new file is under `TARGET_LINES=400`, so it needs no row, and `check_monotonic`'s
  first-appearance arm is never reached. **`docs/dev/file-budget.tsv` is not touched by this PR** —
  confirmed by an empty `git diff origin/develop-v2...HEAD -- docs/dev/file-budget.tsv`.
- `bash scripts/lint-file-budget.sh` → **rc 0**, "file budget OK — every over-target file has a
  ceiling, none grew past it, docs/dev/file-budget.tsv is monotonic". `git rev-parse
  origin/develop-v2` read `0672dd00` both before and after the run, so the shared-`.git` base-ref
  trap does not apply to this result. The firing side of that pair is **CI's own red at
  `4d7df810`** — relayed from the check run, not re-derived locally.

**What the move had to fix, since a move breaks exactly this:** three cross-references that stopped
being true. Two said "above" of cases that stay in `test-lifecycle.sh`; one cited `pithead:14` by
line number, which is the perishable half of any cross-file citation, and now names the
`set -Eeuo pipefail` line by content instead. Zero `above`/`pithead:14` references remain in the new
file. All seven names the block borrows — `make_stubs`, `VALID_PRIMARY`, `VALID_TARI`, `assert_eq`,
`SANDBOX`, `STACK`, `ROOT` — are defined in `lib.sh`, so sourceable-standalone holds.

**A defect caught on this lane's other PR today, applied here before it could ship twice.** The
non-author reviewer on #1762 found that my new domain file there was the only one of 57
`tests/stack/test-*.sh` without the `: "${STACK_SUITE:?...}"` marker, reddening
`test-harness-tooling.sh`'s fragment-refusal row. This file carries the marker and
`# shellcheck shell=bash` line 1 from the outset. Verified with the guard's own loop:
`frag_bad` empty over every fragment with no leftover in the caller's directory, and **my own
positive control** — the same loop with the marker stripped back out of this very file — reports
`test-lock-reinvoke-wiring.sh:refusal-does-not-name-run.sh`, so the clean read discriminates.

### NOT run, and I am not claiming it
The full `tests/stack/run.sh` at this head. It was relaunched against the corrected tree but three
other lanes' suites were already live on this box; **CI's "Shell tests" is that leg.** No
`shellcheck` and no `make` target at this head either — the wrapper's serialize lock makes it
unrunnable while those suites hold it.

### Status
**NOT merge-ready.** The recorded non-author PASS on this PR was given against the previous layout
and I am not carrying it across a restructure; a fresh non-author pass is asked for separately.

VijitSingh97:
## Rebased onto the current tip at `c1baab9f` — the conflict was additive, and it is resolved additively

This PR was `DIRTY`. `develop-v2` moved to `ef3942d0` and the conflict was a single hunk in
`tests/stack/run.sh`: the tip added the `#1213` doctor-surface unit block at the same anchor where
this branch adds its domain registration stanza. Both are appends after the `test-lifecycle.sh`
registration; neither edits the other's text.

**Resolution: the tip's block is kept verbatim and the registration stanza follows it.** No upstream
line was moved, reordered or reflowed. Checked mechanically rather than by eye — the `#1213` block as
it stands on `origin/develop-v2` is present as a contiguous substring of the resolved file.

## The rebase is content-neutral, proven on the axis that actually held fixed

The base moved (it now carries `#1213`, `#1734` and the rest), so **the trees must differ and a tree
comparison would report a false positive**. The patch is what a rebase holds fixed, so that is what I
compared:

```
diff <(git diff a3b723bd 020dc533) <(git diff ef3942d0 c1baab9f)
```

**12 lines of difference, and every one is accounted for:** the blob index line, and the run.sh hunk
header plus its three context lines, which now anchor on the `#1213` block's last lines instead of on
the `test-lifecycle.sh` registration. **Zero difference in `tests/stack/test-lifecycle.sh` and zero in
`tests/stack/test-lock-reinvoke-wiring.sh`** — the two files carrying the actual change came across
byte-identical. The 12 lines are the resolution and nothing else rode along.

**The base did not move under the rebase:** `HEAD~2` is `ef3942d0`, which is `origin/develop-v2` at
the time of the push. Re-read after the push, not before.

## RUN, in this worktree at `c1baab9f`

| | result |
|---|---|
| `git diff --stat origin/develop-v2...HEAD` | `run.sh +3`, `test-lifecycle.sh 11 ++--`, `test-lock-reinvoke-wiring.sh +99` |
| `bash -n tests/stack/run.sh` | rc **0** |
| conflict markers remaining | **0** |
| `make lint-file-budget` | rc **0** — 5 self-test legs ok, then "every over-target file has a ceiling, none grew past it, tsv is monotonic" |
| `tests/stack/test-lock-reinvoke-wiring.sh` | **99 lines**, under the 400 target — no `file-budget.tsv` row needed |
| `tests/stack/run.sh` | **350 lines**, still under target |

The budget gate's self-test legs are the control: it refuses an empty candidate scan and a
zero-tracked-file repo rather than passing vacuously, so the clean result is not the fail-open case.

## NOT done, and not implied

- **I ran no shellcheck and no `make lint-sh`.** The appliance lane holds this box for a gating KVM
  battery, and a peer shellcheck during it blocks on the wrapper's own lock and then starves the
  guest. **CI's `Shell tests` is the binding lint and suite leg here and I am not claiming it** — it
  was still in flight when I wrote this.
- I did not run `tests/stack/run.sh`, for the same reason.
- **The pre-restructure pass is deliberately not carried across.** The restructure changed which file
  the assertions live in, so this needs a full pass at `c1baab9f`, not a delta.

I authored this, so I merge nothing here.

VijitSingh97:
## NON-AUTHOR PASS at `c1baab9f` — full pass, not a delta. **VERDICT: PASS.**

All four claims re-derived rather than relayed. One SHOULD-FIX before the squash merge, one nit,
nothing blocking on substance. `c1baab9f~2` == `ef3942d0` == the `origin/develop-v2` tip, so the base
did not move under the rebase.

### 1. The rebase is content-neutral apart from the resolution — **confirmed, exactly**

`diff <(git diff a3b723bd 020dc533) <(git diff ef3942d0 c1baab9f)` -> rc 1, **12 lines**, and they are
what the comment says: `2c2` is the `run.sh` blob index line, `5,7c5,7` is the `run.sh` hunk header and
its three context lines re-anchored onto the `#1213` block. 8 content lines + 4 `diff` markers = 12.

Byte-identity taken from blob shas, not from a diff: `test-lifecycle.sh` `8c83c55ba0d3` both sides,
`test-lock-reinvoke-wiring.sh` `0017cc1f65c7` both sides. Only `run.sh` moves, which it must.
`--stat` is identical on both sides. Comparing patches rather than trees was the right axis.

### 2. No upstream line moved or reflowed — **confirmed, and settled mechanically**

`git diff --numstat ef3942d0 c1baab9f -- tests/stack/run.sh` -> `3  0`. **Zero deletions**, so no line
could have moved or reflowed — that is a stronger reading than the substring check. The whole change is
`339a340,342`: the registration stanza appended after the untouched `#1213` block at `run.sh:337-339`.

### 3. The substance — **the six rows are a proof, not a grep**

Baseline: the fragment sourced with only `lib.sh` (the `#1330` isolation pattern) gives **6 pass, 0
fail**. It needs nothing from `test-lifecycle.sh`, so no latent order dependency.

**Mutation battery, six mutations against the six rows — a clean diagonal.** Each edits ONE
re-invocation site in `pithead`; each mutation asserted landed (a no-op replacement aborts the leg) and
the tree reverted clean after each:

| mutation | site | row that reddened |
|---|---|---|
| `env -u PITHEAD_LOCK_HELD bash ...` | `pithead:10235` `run_chain` | row 1 only |
| `env PITHEAD_LOCK_HELD=1 bash ...` | `pithead:10235` `run_chain` | row 2 only |
| `env -u PITHEAD_LOCK_HELD "$self" restart` | `pithead:11247` `control_lifecycle` | row 3 only |
| `env PITHEAD_LOCK_HELD=1 "$self" restart` | `pithead:11247` `control_lifecycle` | row 4 only |
| `env -u PITHEAD_LOCK_HELD "$self" backup -y` | `pithead:11333` `control_backup` | row 5 only |
| `env PITHEAD_LOCK_HELD=1 "$self" backup -y` | `pithead:11333` `control_backup` | row 6 only |

Every row can fail, and each fails on exactly its own site and its own direction. The first is the
literal `#1443` defect class.

**The control that mattered most.** Rows 2/4/6 read "blocked" off an ABSENCE for `chain`/`restart`
(`[ -s "$RWDOCKER" ] || blocked=yes`, `test-lock-reinvoke-wiring.sh:66,71`) — the shape that passes
because the child died of something unrelated. So: same probe, marker still stripped, but the
ancestor's `mutation_lock_acquire ancestor` (`:54`) removed. **All three flip to failing** — the child
proceeds. The block is the ancestor's held lock, not the strip breaking the child another way.

**Nothing was lost in the move:** `test-lifecycle.sh` in isolation at this head is **105 pass, 0 fail**;
105 + 6 = 111, matching the pre-restructure count.

**Fragment conventions, enumerated across all 60:** line 1 `# shellcheck shell=bash` — 60/60, zero
mismatches. Line 2 `: "${STACK_SUITE:?...}"` — **60 identical copies, one form**, and the new file's is
byte-identical to `test-lifecycle.sh`'s. Every `test-*.sh` is registered in `run.sh`, and the new file
is picked up by `lint-sh`'s `tests/stack/test-*.sh` glob (`Makefile:105`, beside `lib.sh`) and by
`shfmt`'s `git ls-files '*.sh'` (`Makefile:126`) — no unlinted-file gap.

**Budget, discharging the earlier RETURN:** `test-lifecycle.sh` 1028 -> **1031** against its **1032**
ceiling; new file 99, `run.sh` 350, both under the 400 target so no row is owed. Headroom is now **one
line** — worth knowing for whoever touches that file next.

### 4. CI read by STEP, not by conclusion

`Shell tests (shellcheck + pithead suite)` at `head_sha=c1baab9f`: `completed/success`, all 13 steps
`completed/success`. Separately:

- **step 4 `make lint-sh`** (shellcheck 0.11.0 pin + shfmt 3.13.1) — success. Closes the stated
  "I ran no shellcheck" gap at this exact head.
- **step 5 `bash tests/stack/run.sh`** — success. Closes "I did not run run.sh".
- **and the six rows themselves ran**: the domain header and all six assertion lines appear green in
  the raw job log, `pithead tests: 3310 passed, 0 failed`, zero failure glyphs. `domain_ran`'s `#1400`
  guard (`tests/stack/lib.sh:67-75`) would have failed a zero-assertion domain regardless.

All 15 non-skipped checks pass; `MERGEABLE` / `CLEAN`.

---

### SHOULD-FIX before merge — the body is stale about the restructure

The body still reads "Added `reinvoke_wiring_probe` in `tests/stack/test-lifecycle.sh`", and its
"How it was proven" section cites `shellcheck ... test-lifecycle.sh`, `shfmt -d test-lifecycle.sh` and
a 111/111 isolated run of that file. At `c1baab9f` the probe and all six assertions live in
`tests/stack/test-lock-reinvoke-wiring.sh` and none of those three commands names it. The evidence
comment retracts the *proof* half ("I ran no shellcheck") but nothing corrects the *file*. A squash
merge makes that body the permanent commit message and points a future bisector at the wrong file.
One edit, not a round trip.

### NIT — `tests/stack/test-lifecycle.sh:954` unsets a function it no longer defines

`unset -f ... lock_nest_probe reinvoke_wiring_probe`: that name is defined 0 times in the file now, and
its real definition sources 59 lines LATER in `run.sh`, so the unset is a guaranteed no-op (`unset -f`
on a missing function returns 0; measured). Harmless — a stale pointer left by the move. Dropping the
token shortens a line without changing the line count, which is free under the one-line headroom.

**Checked and clean:** no real wallet address (the fragment uses `lib.sh`'s `VALID_PRIMARY` /
`VALID_TARI` fixtures, unchanged); no hostname or IP; no image-freeze path touched.

### PROVEN vs ASSUMED

**Proven — I ran it in my own detached worktree at `c1baab9f`, each with a control that fired:** the
12-line patch diff and its full accounting; the three blob-sha readings; the 0-deletion numstat; the
6-row baseline; the 6x6 mutation diagonal; the no-ancestor wrong-reason control; the isolated
`test-lifecycle.sh` 105-pass count; the 60/60 header enumeration; the line counts against the budget
tsv; the CI job's per-step conclusions and raw log lines read from the API, not from a checks label.

**Assumed / not done:** I ran **no `make lint-sh`, no shellcheck, no shfmt** — the appliance lane holds
this box for a gating KVM battery, so CI step 4 is the whole of my lint evidence and it is relayed. I
did **not** run the full `tests/stack/run.sh`; step 5 is relayed too. No bench, KVM, rig, image,
container or live stack. I did not re-derive `#1391`'s call-site census, so the corrected "defensive,
not load-bearing today" comment is taken on the author's word about it. I checked only `Shell tests` by
step; the other jobs I read at the label.

**MERGE-READY: PASS at `c1baab9f`.** Fix the body first — it becomes the squash commit message.
`Closes` is inert on `develop-v2`, so hand-close `#1443` with the merge sha.

VijitSingh97:
**MERGE-READY: PASS at `c1baab9fa876ed8da1c2bb9048ad912235ada6f3`.**

Recorded non-author FULL pass (ephemeral reviewer, `20260904T224613Z`) — not a delta; the
pre-restructure pass was deliberately not carried across. It re-derived all four claims and confirmed
them, notably:

- **The rebase is content-neutral apart from the one conflict resolution.**
  `diff <(git diff a3b723bd 020dc533) <(git diff ef3942d0 c1baab9f)` → 12 lines, every one accounted
  for: the `run.sh` blob index line, and the hunk header plus three context lines that now anchor on
  the `#1213` doctor block instead of the `test-lifecycle.sh` registration. 8 content + 4 marker lines.
  It also checked byte-identity by **blob sha** rather than by diff — the moved files are identical
  across the rebase.
- **The base did not move under the rebase**: `c1baab9f~2` == `ef3942d0` == the `develop-v2` tip, so
  the branch is exactly two commits on the current tip. `Shell tests` is green at this sha.

I am the author, so I do not merge.

### The SHOULD-FIX is discharged, and not the way it might look

The body named `tests/stack/test-lifecycle.sh` as the probe's home; at this head the probe and all six
assertions live in `tests/stack/test-lock-reinvoke-wiring.sh`. A squash merge would have made that the
permanent commit message and pointed a future bisector at the wrong file.

**Fixed in the body only — the head sha is unchanged, so the PASS above still measures this tree**
(re-read after the PATCH to confirm it took). The location claim in "What changed" is corrected. The
three commands under "How it was proven" are **not** re-pointed: they genuinely ran against
`test-lifecycle.sh` before the restructure, and renaming the file inside them would manufacture
evidence for a run that never happened. They now carry an explicit retraction saying so and naming
what stands instead.

### The NIT is real and deliberately left

`tests/stack/test-lifecycle.sh:954` still `unset -f`s `reinvoke_wiring_probe`, which that file no
longer defines — a stale pointer my own move left behind, and a guaranteed no-op (`unset -f` on a
missing function returns 0). Fixing it is a code change: it moves the head sha and voids the pass
above, for a token that cannot change behaviour. **Left in place, recorded here so it is not lost.**
Worth folding into the next diff that touches that file rather than spending a review round trip now.

Issue #1214, opened 2026-08-22
XvB decision table: both reward columns are '—' when XvB is disabled, so it cannot answer 'should I enable it?'


VijitSingh97:
## What the operator sees

With XvB disabled, the "Should I donate?" decision table renders — but both reward columns,
**XvB says / yr** and **Study est. / yr**, are `—` on every tier. Cost / yr fills in normally.

That is the one surface built to answer *"is enabling XvB worth it?"*, and the two columns
carrying the answer are the blank ones. #938 made the table visible when XvB is off for exactly
this reason; the table now renders, but it cannot yet answer the question it exists to ask.

## Why (traced, not guessed)

Both columns come from per-tier fields in `build_xvb_calc`, and both are gated on the same
flag:

- `expected_reward_year` → `float(estimates[key]) if estimates_available and key in estimates
  else None` — `build/dashboard/mining_dashboard/web/views.py:1878`
- `assumed_reward_year_range` → same guard, `views.py:1892`
- `estimates_available = bool(estimates) and not estimates_stale`, from
  `state_mgr.get_xvb_reward_estimates()` — `views.py:1864`

The client renders `—` whenever those are null (`xvbDecisionRows` in `logic.mjs`, consumed at
`components.mjs:502`), which is correct behaviour for missing data.

The cache behind them is only ever filled while XvB is on —
`build/dashboard/mining_dashboard/service/data_service.py:1787`:

```python
# ONLY when XvB is enabled — disabling XvB must stop the egress entirely.
if ENABLE_XVB and iteration_count % 10 == 0:
    await self._sync_xvb_stats()
    await self._sync_xvb_reward_estimates()
```

So with XvB off the estimates are never fetched, `estimates_available` is permanently False,
and both columns are permanently `—`. On a box that has never had XvB enabled there is not even
a stale cache to fall back on.

**That gate is deliberate, and it should stay.** Turning XvB off is a privacy guarantee — no
traffic to xmrvsbeast.com at all, even over Tor (#163). So the fix cannot simply be "fetch the
estimates anyway".

## The useful part: only one input is actually missing

Everything else the two columns need is already local:

- **Study est.** is just the face figure × `XVB_REALIZATION_PRIOR = (0.28, 0.39)`
  (`views.py:1538`) — a repo constant from the on-chain audit, no network involved.
- **Cost / yr** is computed from local hashrate, which is why it fills in today.
[...]

Issue #163, opened 2026-06-04
Dashboard: route XvB stats through Tor + gate behind XVB_ENABLED (IP↔wallet leak)


VijitSingh97:
Part of #160 (privacy egress epic).

**The worst leak in the stack:** the dashboard fetches XvB stats over **clearnet with the operator's wallet address as a query param**, from a `network_mode: host` container — correlating **real IP ↔ Monero wallet**. And it's **not gated by `XVB_ENABLED`**, so disabling XvB doesn't stop it.

`build/dashboard/mining_dashboard/client/xvb_client.py:37` → `requests.get("https://xmrvsbeast.com/cgi-bin/p2pool_bonus_history.cgi?address=<WALLET>")`, called unconditionally from `service/data_service.py:482` (~every 5 min).

## Changes
- [ ] **Route through Tor.** `requests.get(..., proxies={"https": "socks5h://172.28.0.25:9050"})` — the host-net dashboard can reach the bridge Tor SOCKS at `172.28.0.25:9050`; `socks5h` resolves the hostname via Tor too (no local DNS leak). Add the `requests[socks]`/PySocks dependency to the dashboard image. xmrvsbeast still receives the wallet (inherent to the query) but no longer tied to your IP.
- [ ] **Gate behind `ENABLE_XVB`** in `data_service` so turning XvB off actually stops the egress.
- [ ] **Remove dead code:** `TARI_EXPLORER_URL` in `config/config.py:135` (defined, never called).

## Acceptance
- [ ] With XvB on, `tcpdump` on the host shows the xmrvsbeast request exiting via Tor (to `172.28.0.25:9050`), not direct to xmrvsbeast over clearnet.
- [ ] With `xvb.enabled: false`, **no** request to xmrvsbeast.com is made at all.

Labels: dashboard, security · Milestone: v1.0

## Related issues
- #160 — tracking epic
- #166 — XvB donation over Tor (shares the `xvb.*` config + algo_service path)
- #136 — XvB "stats updated" freshness timestamp (this issue changes when/whether the fetch runs)
- #122, #91 — host-networked dashboard egress / SSRF + host-networking hardening (same surface)
- #164 — privacy/egress docs covering this fix

PR #1718, opened 2026-09-03
fix(dashboard): #1231 the XvB Odds cell names what it waits for instead of dashing

### LINKED ISSUE #1231: XvB decision table: the Odds column still dashes on a disabled box — no stable published data to vendor

VijitSingh97:
Closes #1231.

## What it does

The per-tier XvB decision table rendered a bare `—` in **Odds / 30d** whenever no odds figure existed. On a box that has never enabled XvB that is indistinguishable from a measured zero — the column meant to inform the enable/don't-enable decision instead reads as "your odds are nil".

The winners feed that fills the round-stats cache is `ENABLE_XVB`-gated, so `build_xvb_calc` already separates the two empty states through `calc.enabled` — the same flag its own vendored-reward fallback keys on (`fallback_eligible = not metrics.xvb_enabled`). The cell now names which state it is in:

| State | Cell reads |
|---|---|
| XvB off | `needs XvB enabled` |
| XvB on, no round statistics cached | `awaiting sync` |
| odds present | unchanged — `≈ N wins · M players` |

Each empty state carries its own tooltip. This is #1231's first candidate shape, the one its own body calls "a wording fix, no data problem". **Nothing is derived, estimated, or dialled out while disabled** — #1214's carried constraint holds because nothing here is computed, only named.

## Where I deviated from the issue, and why

#1231 suggested the wording "computed after first sync". I did not use it. `build_xvb_calc` empties `round_types` when the round-stats cache is **stale** as well as when it was never filled, and the payload publishes staleness only for the *separate* reward-estimates cache — so the two cannot be told apart in the client, and an enabled box reaching this branch may well have synced before. "computed after first sync" would be false for that half. `awaiting sync` is true in both sub-cases. The reasoning is in the source comment so the next reader does not re-litigate it.

## Prose the change falsified, swept

Three sentences claimed the column "stays empty", which is no longer true:
- the block-level `#xvb-disabled-note` in `xvbview.mjs`
- `docs/dashboard.md`'s XvB Tier narrative
- `docs/dashboard.md`'s Odds / 30d table row

All three updated. The footnote's "Win odds ... stay unavailable until XvB runs again" is still true and is left alone.

## Evidence — what I actually ran

- **`make test-frontend` rc 0 — 516 tests, 516 pass, 0 fail** on the tree being shipped.
- **New file `dashboard/tests/frontend/xvbodds.test.mjs`, 8 tests, with a fired negative control.** Restoring `xvbview.mjs` to its `origin/develop-v2` content and re-running that file gives **6 failed / 2 passed**. The two that pass on both sides are by design and named as such: the no-regression test (a measured figure still renders) and the control test itself. The other six all discriminate. The base file was restored from a scratchpad copy and the restore **verified by sha256 content match**, not by an exit code.
- **Gates run individually, each rc captured without a pipeline:** `lint-js`, `lint-md`, `lint-docs-voice`, `lint-operator-strings`, `lint-topology`, `lint-file-budget`, `lint-py`, `test-frontend` — **all rc 0**.
- Branch cut from `origin/develop-v2` at `3eeb1b28`, re-derived at cut time.

## What I did NOT do

- **I did not run `make lint-sh`.** The diff touches no shell file, and that target is the serialized OOM path with other lanes live on this box. CI runs it.
- **I did not run `make test-dashboard` or `make test-patch-coverage`.** The diff changes no Python. Patch coverage over a zero-Python diff returns "No lines with coverage" and passes vacuously, so it would be evidence of nothing here; the frontend tests are the instrument that applies.
- **I did not touch the two remaining candidate shapes in #1231** — the historical odds band from the delivery study's archived round data. That one needs the study's data to actually support a stable band, which I did not evaluate. This PR deliberately takes only the wording fix; if the band is still wanted, it wants its own issue and its own evidence.
- I did not change any server-side odds computation. `_odds_day` and the round-stats staleness rule are untouched.

## Over-engineering pass (done by hand, on merits)

Findings, both addressed above rather than left as notes:

1. **The wording was wrong for one of the two enabled sub-cases.** I had written "after first sync"; re-reading `build_xvb_calc` showed `round_types` is also emptied on a STALE cache, so that sentence would be false for a box that had synced and gone stale. Changed to `awaiting sync`, which is true in both. This is the finding the pass was worth doing for.
2. **Three prose claims that the column "stays empty"** were falsified by the change itself and would have shipped stale. Swept — listed above.
3. Considered and rejected: distinguishing the stale case using `estimates_stale`. That field describes the reward-estimates cache, a *different* cache from round stats; using it would conflate two sources to manufacture a distinction the payload does not actually publish.
4. `oddsPending` / `oddsTitle` / `oddsPendingTitle` are computed once per render, not per row — they do not vary per tier.

The gate itself verifies nothing (its skill does not load in this environment); this pass was done on merits, and the gate cleared on the identical retry.

## Budget

`xvbview.test.mjs` carries a 431-line ceiling and sat at exactly 431. Both pinned assertions were replaced **in place**, one line for one line — the file is still 431 records (the gate counts by record, not `wc -l`). The new test file is 151 lines, under the 400-line target, so it correctly carries **no** tsv row.

https://claude.ai/code/session_01TzwiryicFnQLeDsE7RJ9Ln

### DIFF
--- a/dashboard/mining_dashboard/web/static/xvbview.mjs
+++ b/dashboard/mining_dashboard/web/static/xvbview.mjs
@@ -48,6 +48,24 @@ function XvbDecisionTable({ calc, coeffDay, hr, energy }) {
   // Fiat mirror (#520) for the best sustainable net only — one line, never a fiat number whose
   // XMR figure is hidden.
   const best = rows.filter((r) => r.sustainable && r.net).sort((a, b) => b.net[1] - a.net[1])[0];
+  // #1231: the Odds cell rendered a bare "—" on a box that has never enabled XvB, which reads as
+  // "the odds are zero" rather than "nothing has been read yet". The winners feed that fills the
+  // round-stats cache is ENABLE_XVB-gated (build_xvb_calc's own fallback keys on the same flag),
+  // so a disabled box structurally has nothing to show and an enabled one is merely waiting. Name
+  // which of the two it is, off the `calc.enabled` every other live-donation surface already uses.
+  // #1214's constraint holds: nothing here is derived or estimated, only named.
+  //
+  // "awaiting sync" rather than the issue's suggested "computed after first sync": build_xvb_calc
+  // empties round_types when the cache is STALE as well as when it was never filled, so an
+  // enabled box reaching this branch may well have synced before. The payload publishes staleness
+  // for the reward-estimates cache only — a different cache — so the two cannot be told apart
+  // here, and wording that claims no sync has ever happened would be false for half of them.
+  const oddsPending = calc.enabled ? "awaiting sync" : "needs XvB enabled";
+  const oddsTitle =
+    "How often this tier's rounds pay out, and among how many qualifiers — from XvB's public winners feed.";
+  const oddsPendingTitle = calc.enabled
+    ? "No round statistics are cached right now — XvB's winners feed fills this in at the next sync."
+    : "XvB is off, so its winners feed is never read — and that feed is the only source for draw frequency and qualifier counts. Enable XvB to compute this.";
   return html`
         <div class="xvb-comparison">
             <label class="xvb-compare-label">Should I donate? — per-tier verdict (per year)</label>
@@ -79,7 +97,7 @@ function XvbDecisionTable({ calc, coeffDay, hr, energy }) {
                     <span title="P2Pool earnings forgone by donating the tier threshold for a year, at your current rate.">Cost ${r.cost !== null ? formatXmr(r.cost) : "—"}</span> ·
                     <span title="XvB's own published expected reward — face value: prices every bonus hash at full block reward.">XvB says ${r.xvbSays !== null ? formatXmr(r.xvbSays) : "—"}</span> ·
                     <span title=${`${estTitle} ${bandNote(est)}`}>${estLabel} ${fmtMid(est)}</span> ·
-                    <span title="How often this tier's rounds pay out, and among how many qualifiers — from XvB's public winners feed.">Odds / 30d ${r.oddsPer30d ? `≈ ${Number(r.oddsPer30d.toPrecision(2))} wins · ${Number((r.players || 0).toPrecision(2))} players` : "—"}</span>
+                    <span title=${r.oddsPer30d ? oddsTitle : oddsPendingTitle}>Odds / 30d ${r.oddsPer30d ? `≈ ${Number(r.oddsPer30d.toPrecision(2))} wins · ${Number((r.players || 0).toPrecision(2))} players` : oddsPending}</span>
                 </p>
               </div>`;
             })}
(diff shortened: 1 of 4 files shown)


VijitSingh97:
## Controller pass — PASS at `520e013e` (non-author; judged on the fable tier). Merging when the checks are terminal.

**Re-derived by me, in a detached worktree at the head, against the tip `48e0546d`:**
- `node --test dashboard/tests/frontend/`: **516 tests, 516 pass, 0 fail.**
- Negative control, re-run rather than relayed: `xvbview.mjs` replaced by its `origin/develop-v2` content under the new `xvbodds.test.mjs` → **6 failed / 2 passed**, and the two survivors are tests 5 and 6 — the no-regression test and the control itself — exactly as the body names them. The six that fire are the two state wordings, the discriminator, the tooltips, the zero-rate boundary and the disabled-note sweep. A precisely aimed kill.
- The two source claims the wording turns on: `xvblogic.mjs:85` sets `oddsPer30d` only for `win_odds_day > 0`, so a zero rate is a waiting state as test 7 asserts; and `build_xvb_calc` empties `round_types` whenever `xvb_stats_are_stale(round_state)` — so an enabled box in this branch may have synced before, which is why "awaiting sync" is right and the issue's "computed after first sync" would have been false for that half. The disabled-box half rests on the #163 egress rule (`egress.py:_xvb_route` → INACTIVE while disabled; the `xvb_views.py` vendored-fallback comment says the live fetch stops entirely) — read at source, not executed.
- Gates at the head: `lint-docs-voice` rc 0, `lint-operator-strings` rc 0. Ruling 8a, both legs (off-tip branch): budget gate rc 0 and parity rc 0 at the head AND at the merged tree; `merge-tree --write-tree` rc 0. `xvbview.test.mjs` holds at its 431 ceiling with the two pinned assertions replaced line-for-line; the new file carries no tsv row, correctly.

**Relayed, not re-derived:** the lane's sha256 restore check and its individual gate rcs beyond the two I re-ran. **Not run by either of us, disclosed:** `lint-sh` (no shell in the diff; CI's job covers it) and patch coverage (no Python changed — vacuous here).

**Judgement:** the fix is a naming, not a derivation — #1214's constraint holds, and `calc.enabled` is the discriminator every other live-donation surface already keys on. A box turned OFF while its round-stats cache is still fresh keeps showing real figures (unchanged behaviour; the branch is only reached with no odds figure), and once that cache ages out it reads `needs XvB enabled`, which is then true. #1231 closes by hand after the merge.

Issue #1419, opened 2026-08-24
The weekly link check has been red for nine straight Mondays, and 11 of its 12 findings are noise from vendored upstream copies


VijitSingh97:
## What is wrong

`.github/workflows/lychee.yml` has failed on **every scheduled run since 2026-06-29** — nine
consecutive Mondays. The last green run was 2026-06-22.

```
gh run list --workflow lychee.yml --limit 20 --json createdAt,event,conclusion

2026-08-24  schedule  failure     2026-07-13  schedule  failure
2026-08-17  schedule  failure     2026-07-06  schedule  failure
2026-08-10  schedule  failure     2026-06-29  schedule  failure
2026-08-03  schedule  failure     2026-06-22  schedule  success   <- last green
2026-07-27  schedule  failure
2026-07-20  schedule  failure
```

Every run in that listing is `event: schedule`. The workflow triggers on `schedule` and
`workflow_dispatch` only — deliberately, per #281, so that flaky external links never redden a PR.
So **no pull request has ever surfaced this and none ever can.**

## Why it went unread for two months

The workflow's own comment states the assumption: *"The run goes red in the Actions tab if a link
breaks."* The Actions tab is the entire notification mechanism, and nine unread weeks is the
measurement of how well that works. This is the live instance of #1377, and a better one than the
example that issue was filed on, which reddened every branch at once and was worked the same day.

## A permanently red gate carries no information

This is the part worth acting on rather than the individual dead link. A gate that has been red for
nine runs **cannot signal a new break**, because there is no transition left for it to make. The
next genuinely broken link in our docs will land in an already-red workflow and change nothing that
anyone could see. It is the mirror of the false-green shapes this project files: the reading is
constant, so it is not a reading.

## What the failure actually is

From the 2026-08-24 run: 913 links checked, 884 successful, 12 errors, 2 timeouts, 6 excluded,
9 unsupported.

**Eleven of the twelve errors are not our prose.** They are in
`docs/research/xvb-delivery-study/data/sources/`, which holds vendored verbatim copies of upstream
project READMEs:

| kind | count | example |
|---|---|---|
| `418 I'm a teapot` | 4 | `p2pool.observer/api` — bot rejection, the site is up |
| `403 Forbidden` | 4 | `eprint.iacr.org/2017/1066.pdf`, `nixos.wiki/wiki/Flakes` — bot rejection |
[...]

PR #1719, opened 2026-09-03
fix(ci): #1419 the weekly link check can signal again

VijitSingh97:
## Design-scrutiny pass (over-engineering)

Done by hand on this diff, which is two changed lines.

- **Simplest mechanism that works?** Yes. The alternatives are worse: URL-regex exclusions in `.lycheeignore` would need roughly seven patterns and would mask the same dead hosts *everywhere else in the repo*, and the input glob cannot be negated. `--exclude-path` is lychee's own path mechanism and is the narrowest of the three.
- **Anything speculative added?** No new files, no new config surface, no abstraction, no options nobody asked for.
- **One real finding, accepted rather than fixed:** the exclusion is a *tree*, so a first-party doc later added under `data/sources/` would go silently unchecked. That is the correct trade today — everything in that directory is vendored by construction, and `MANIFEST.md` says so — but it is a scope the next person should know is broad. Narrowing it to a per-file list would break the moment the study vendors another source, which is the more likely failure.

### DIFF
--- a/.github/workflows/lychee.yml
+++ b/.github/workflows/lychee.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Check links in Markdown
         uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8 # v2.9.0
         with:
-          args: "--no-progress './**/*.md'"
+          args: "--no-progress --exclude-path docs/research/xvb-delivery-study/data/sources './**/*.md'"
           fail: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # higher github.com rate limit for repo links
--- a/docs/workers.md
+++ b/docs/workers.md
@@ -354,7 +354,7 @@ the latest RigForge release itself — the per-worker twin of the stack's one-cl
 The rig side must opt in: RigForge's `control_upgrade` capability (default off, on top of its
 `control` flag, bearer token, and source pin) and RigForge **v1.11.2 or newer** — earlier versions
 refuse legitimate upgrades on a fresh clone. See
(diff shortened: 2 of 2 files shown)


VijitSingh97:
## PASS — non-author pass (controller, fable tier), proven on the real gate

The claim this PR makes is that the weekly link check can go green again. I ran the gate itself rather than reasoning about the flag: `lychee.yml` dispatched on this head (`e022c21d`), twice.

| run | checked | ok | errors | what failed |
|---|---|---|---|---|
| 33816937785 | 701 | 699 | 1 | `[502]` on this repo's own `/actions/workflows/ci.yml` badge link — GitHub returned a 502 to my `gh` call in the same minute |
| 33817233000 | 701 | 700 | 0 | **green** |

Against the issue's baseline (913 checked, 12 errors, 2 timeouts): the 11 vendored-copy errors and both timeouts are gone, which is `--exclude-path` taking effect (212 fewer links checked), and the ADR link now resolves — `docs/adr/0002-remote-worker-upgrade.md` is the only ADR 0002 on rigforge `main`, re-derived via the contents API. The one red was GitHub itself, not a link.

Re-derived, not relayed: the diff against `merge-base` (2 files, 2 lines each way), the ADR filename, both runs. Not run: the file-budget gate — the diff touches no budgeted file. Merging with `--admin` per DECISIONS 21; `Closes` fires here.

One thing this PR does not fix, filed rather than folded in: `THIRD_PARTY_LICENSES.md`'s gnu.org link answered 429 and a timeout on the twin's two runs and times out from this box too — it will red a future Monday the same way the vendored copies did. Separate issue, named below.


Gnu.org finding filed as #1726.

PR #1720, opened 2026-09-03
fix(ci): #1419 the develop-v2 twin of the link-check restore

VijitSingh97:
- **The `.lycheeignore` addition, checked on the same terms:** it is one anchored URL pattern beside three of identical shape that already sit there, not a new mechanism. The alternative — making `pithead.local` resolve, or dropping it from the docs — would damage the doc to satisfy a link checker, which is the wrong direction.

### DIFF
--- a/.github/workflows/lychee.yml
+++ b/.github/workflows/lychee.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Check links in Markdown
         uses: lycheeverse/lychee-action@e7477775783ea5526144ba13e8db5eec57747ce8 # v2.9.0
         with:
-          args: "--no-progress './**/*.md'"
+          args: "--no-progress --exclude-path docs/research/xvb-delivery-study/data/sources './**/*.md'"
           fail: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # higher github.com rate limit for repo links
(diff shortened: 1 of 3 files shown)


VijitSingh97:
## PASS — non-author pass (controller, fable tier), proven on the real gate, with one finding outside the diff

`lychee.yml` dispatched on this head (`8948f9ab`), twice:

| run | checked | ok | errors | timeouts | what failed |
|---|---|---|---|---|---|
| 33816940119 | 833 | 830 | 0 | 1 | `[TIMEOUT]` gnu.org `gpl-3.0.txt` in `THIRD_PARTY_LICENSES.md` |
| 33817230727 | 833 | 830 | 1 | 0 | `[429]` the same gnu.org link |

Every error the issue inventoried is gone: 0 errors from the vendored tree, the ADR link resolves, and the `pithead.local` entry is doing its job — this branch's docs carry 11 references across 6 files (the twin on `develop` has 3, none of them links, which is why it needs no entry; re-derived with `git grep`). The sole failure is gnu.org rate-limiting a non-browser client: from this box the same URL times out at 20 s, so it is the host, not the link, and it is the exact class the `.lycheeignore` torproject entry already documents.

Verdict: the diff does what it says and is the narrowest form (one anchored pattern beside three of the same shape). The gnu.org failure predates this PR, is not in its scope, and would be wrong to fold in silently — filed as its own issue, named below, so the next Monday on `develop` does not reproduce #1419 with a different host. `develop-v2` is unprotected; merging on this pass. Not run: the file-budget gate — no budgeted file in the diff.


Gnu.org finding filed as #1726.

Issue #695, opened 2026-07-20
Save & preview shows a spurious CLEARNET_STATE_DIR change: $PWD-derived paths flip between symlink and physical checkout dir


VijitSingh97:
## Symptom

On a bundle-deployed box, open the dashboard Configuration editor, change nothing, and click Save & preview. The Review changes modal reports:

> CLEARNET_STATE_DIR: /srv/code/current/data/clearnet-state → /srv/code/pithead-v1.9.3/data/clearnet-state.

Both paths are the same directory — `current` is the deploy symlink.

## Root cause

`SCRIPT_DIR` is resolved with a logical `pwd` at startup (`pithead:89`), so every `$PWD`-derived `.env` value (`CLEARNET_STATE_DIR`, `CONTROL_DIR`, `CADDY_LOG_DIR`, `TARI_WALLET_SECRET_FILE`) inherits whatever path the invoker used:

- An interactive deploy from `cd /srv/code/current` keeps the logical symlink path — the live `.env` stores `/srv/code/current/...`.
- The control runner's systemd unit runs `ExecStart=/srv/code/pithead-v1.9.3/pithead control-run-pending` with `WorkingDirectory=` on the physical dir, so the preview's fresh render produces `/srv/code/pithead-v1.9.3/...`.

`env_changed_keys` compares strings (`pithead:1566`), so the same directory under two spellings reads as a change. `CONTROL_DIR`/`CADDY_LOG_DIR` are already silenced in `describe_change` ("change only when the checkout moves", `pithead:4470`) and `TARI_WALLET_SECRET_FILE` under the Tari group, which is why only `CLEARNET_STATE_DIR` surfaces. The mismatch also means alternating invocation styles rewrite these `.env` keys back and forth on every apply, churning the containers that mount them.

## Fix

Resolve `SCRIPT_DIR` with `pwd -P` so every render uses the physical checkout path regardless of how pithead was invoked, and add `CLEARNET_STATE_DIR` to the silenced internal-path arm in `describe_change`, consistent with its siblings. Bonus: `update_current_symlink` (which requires a `pithead-vX.Y.Z` basename) starts working when upgrade is invoked through the symlink.

VijitSingh97:
[...]

PR #1721, opened 2026-09-03
fix(tests): #1705 a tests/stack sandbox is built through one fail-closed constructor, never a bare mktemp assignment

VijitSingh97:
Addresses #1705. No `Closes` keyword: PRs here merge to `develop-v2` while the default branch is `develop`, so it never fires. The issue is closed by hand on merge.

## What was wrong

`VAR=$(mktemp -d)` leaves `VAR` set-but-empty when `mktemp` fails. `tests/stack/run.sh` is `set -uo pipefail` with no `-e`, so a failed assignment does not abort the run, and `set -u` does not fire either, because the variable *is* set. A later `rm -rf "$VAR/store"` then expands against `/store`.

#1661 fixed the *composed* form in `tests/stack/lib.sh`, where the collapse was `cd ""`. This is the plain form, at the 38 remaining sites.

## The shape of the fix, which is the part worth reviewing

The constructor **assigns by name** rather than printing, and that is load-bearing rather than a style choice. The obvious helper, called as `X=$(mk_tmpdir)`, **cannot fail closed**: its `exit` ends only the command substitution's subshell, the assignment still succeeds with `X` empty, and the run carries on — the defect completely unchanged, now with a guard in front of it that reads as a fix. Assigning through the caller's own shell is what puts the refusal somewhere it can stop the run.

`lib.sh`'s own #1661 block now calls the same constructor, so the recipe cannot drift between that file and the domain files.

## Measured, with a control that fires

Proven on this box against the unfixed form, not argued:

| leg | expect | got |
|---|---|---|
| new form, failing `mktemp` shim | refuse | rc 1, names the variable, no `CONTINUED` line |
| same, the caller's seeded directory | intact | sentinel and nested file both survive |
| **old form, same shim — the control that must arm** | continue | rc 0, `X=[]`, `rm would target: [/store]` |
| new form, working `mktemp` | a real dir | rc 0, `ISDIR` |

The third row is the one that matters: without it the first two could be green because the shim never armed.

## Coverage

- The existing `tests/integration/selftest-stack-sandbox.sh` already pins the refusal and carries its own arming control, and it now exercises the shared constructor because `SANDBOX` is built through it. **7 of 7 pass unchanged after the refactor.**
- Added there: a **mechanical invariant** rather than a list of 38 names, which would rot on the next file added. A `mktemp -d` may appear only inside a guarded expression, never as a bare assignment. **Its pattern is proven able to match** against a reconstructed pre-fix line before its absence elsewhere is read as evidence. 9 of 9 pass.

## Scope, and what I did NOT convert

`tests/stack/run.sh` (3 sites) and `tests/stack/test_data_reset.sh` (1 site) still carry the old form. They are outside this lane's grant — `run.sh` is the appliance lane's live line. The invariant **excludes them by name rather than by silence**, so the row states what it does not check. Those 4 sites want a follow-up from whoever owns them.

## Placement

All 14 domain files are **exactly line-neutral** — the conversion is one-for-one, which matters because several sit at a zero-headroom ceiling. Only `lib.sh` (298 to 314) and the self-test (140 to 173) grow, and both are under the 400-line budget threshold, so neither needs a row. `make lint-file-budget` green against the staged index.

## Lint

`shellcheck --severity=warning` over `tests/stack/lib.sh tests/stack/test-*.sh os/overlay/pithead-boot` (the grouping the Makefile uses, so the pairing that suppresses the known SC2034s is preserved): **rc 0**. `shfmt -i 4 -d` over every changed file: clean.

One real finding came out of the first lint pass and is fixed rather than papered over: shellcheck cannot follow `printf -v`, so it read `_sbx` as never assigned. The directive that silences it carries the reason and says why it is narrower than it looks — the 38 uppercase call sites are unflagged only because shellcheck assumes uppercase names may come from the environment, so a lowercase one added later will need its own.

## The gap, stated rather than implied

**The full tier-1 `tests/stack/run.sh` suite was still running when I hit my context bar.** It was progressing normally with no failures in the first 1,684 lines. It is not a green I am claiming, and this should not merge until someone reads it. The log is at `stack-run.log` in this session's scratchpad, and the run is cheap to repeat.

## Over-engineering review of my own diff, with what I declined

Run off disk against the diff, not skipped.

**I did more than the issue asked, and here it is.** #1705 names roughly ten files and the sites with a downstream `rm -rf`. I converted **every** bare assignment in the granted files — 38 sites in 14 files — because the root cause is the unguarded assignment, not the `rm`. An empty variable also reaches `tar -C ""` and a `mkdir -p` under the filesystem root; keying the fix to one landing site would leave the class half-closed, and the next site would read as guarded when it is not.

**I refactored working, shipped code.** The #1661 block in `lib.sh` did not have to change. I changed it so there is one constructor rather than two copies of the same recipe, which is the duplication that drifts. The risk is real and the check is direct: that self-test passes 7 of 7 unchanged, including its own arming control.

**Declined — canonicalizing inside the constructor.** Making it do `pwd -P` would be more uniform, and the sandbox needs it for the macOS symlink reason #695 records. I left it out: it would change the path *value* at all 38 sites, and any assertion comparing a non-canonical path would move under it. Uniformity is not worth an unmeasured change at 38 sites in one PR.

**Declined — a sibling for the non-directory `mktemp`.** No evidence of the same hazard for that form here, so it would be a guard with no demonstrated case.

**Kept deliberately, though both are long for what they wrap:** the comment above the constructor, because the by-name shape is precisely what a later reader would "simplify" back into a command substitution and silently restore the defect; and the four lines on the lint directive, because they say why it is narrower than it looks.

### DIFF
--- a/tests/integration/selftest-stack-sandbox.sh
+++ b/tests/integration/selftest-stack-sandbox.sh
@@ -136,5 +136,38 @@ else
 fi
 
 echo ""
+echo "== every tests/stack sandbox is built through mk_tmpdir, not a bare assignment (#1705) =="
+
+# The constructor is only half the fix. The other half is that the call sites in the domain files
+# actually reach it, and a list of the converted names would rot on the next file added. So this
+# is the invariant instead: in the files #1705 covered, `mktemp -d` may appear only inside a
+# guarded expression, never as a bare `VAR=$(mktemp -d)` whose failure leaves VAR set-but-empty
+# for a later `rm -rf "$VAR/store"` to expand against /store.
+#
+# tests/stack/run.sh and the standalone tests/stack/test_*.sh are NOT covered and still carry the
+# old form. They belong to other lanes, so #1705's conversion stopped at the grant boundary. They
+# are excluded by name rather than by silence, so that this row says what it does not check.
+BARE='^[[:space:]]*[A-Za-z_][A-Za-z0-9_]*="?\$\(mktemp -d\)"?[[:space:]]*$'
+STACK_DIR="$HERE/../stack"
+
+# The pattern must be shown able to match before its absence anywhere means anything.
+CTRL="$TMP/bare-control.sh"
+printf 'RSB=$(mktemp -d)\nrm -rf "$RSB/store"\n' >"$CTRL"
+if grep -Eq "$BARE" "$CTRL"; then
+    it_pass "the bare-assignment pattern matches the pre-fix form (control arms)"
+else
+    it_fail "the bare-assignment pattern matches the pre-fix form" \
+        "it matched nothing, so the absence checked below proves nothing"
+fi
+
+found=$(grep -El "$BARE" "$STACK_DIR/lib.sh" "$STACK_DIR"/test-*.sh 2>/dev/null | tr '\n' ' ')
+if [ -z "$found" ]; then
+    it_pass "no bare mktemp -d assignment survives in lib.sh or the test-*.sh domain files"
+else
+    it_fail "no bare mktemp -d assignment survives in lib.sh or the test-*.sh domain files" \
+        "still bare in: $found"
+fi
+
+echo ""
 echo "selftest-stack-sandbox: $IT_PASS passed, $IT_FAIL failed"
 [ "$IT_FAIL" -eq 0 ] || exit 1
--- a/tests/stack/lib.sh
+++ b/tests/stack/lib.sh
@@ -99,19 +99,35 @@ wait_while_alive() { # <pid> <check-fn-name>
     done
 }
 
+# mk_tmpdir <varname> — create a throwaway directory and ASSIGN IT BY NAME, or refuse the run.
+#
+# Assigning by name rather than printing is the whole point (#1705). The obvious constructor,
+# called as `X=$(mk_tmpdir)`, cannot fail closed: its `exit` ends only the command substitution's
+# subshell, the assignment still succeeds with X empty, and tests/stack/run.sh sets no `-e`, so
+# the run carries on. `set -u` does not fire either — the variable IS set, it is empty. A later
+# `rm -rf "$X/store"` then targets /store. Assigning through the caller's own shell is what puts
+# the refusal somewhere it can stop the run.
+mk_tmpdir() {
+    local _mk_d
+    _mk_d="$(mktemp -d)" && [ -d "$_mk_d" ] || {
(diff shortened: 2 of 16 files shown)


VijitSingh97:
## The gap named in the body is now closed — the tier-1 suite finished green

The body says the full `tests/stack/run.sh` run was still in flight when I hit my context bar, and that it should not be read as a green. It has since finished on this branch:

```
pithead tests: 3175 passed, 0 failed
rc=0
```

Zero failures across the whole run, not only in the portion I had seen. That is the regression check the 38-site conversion needed, and it is the last thing the body listed as outstanding.

Everything else in the body stands as written, including what this PR does **not** cover: the 4 sites in `tests/stack/run.sh` and `tests/stack/test_data_reset.sh` are outside this lane's grant, still carry the old form, and are excluded from the invariant **by name** so the row states its own limit.

VijitSingh97:
## Controller pass — PASS at `ce774141` (non-author; judged on the fable tier). Merging.

**Re-derived by me, against the tip `52ab8319` (merge-base `289821f1`, so the branch carries the lowered tsv row — no head-leg artifact):**
- Full tier-1 suite in my own detached worktree at the head: **3,175 passed, 0 failed, rc 0** — independently of the author's run posted above.
- Ruling 8a, both legs: budget gate rc 0 and parity rc 0 at the head AND at the merged tree; `merge-tree --write-tree` rc 0. Every check terminal and green at this head.
- Diff read in full (16 files). The constructor's shape is the finding and it is right: `_mk_d="$(mktemp -d)" && [ -d "$_mk_d" ] || { …; exit 1; }` refuses on either failure, and `printf -v "$1"` assigns through the caller's shell, which is the only place an `exit` can stop a `set -uo pipefail` run with no `-e`. `lib.sh`'s #1661 block now calls the same constructor, so the two recipes cannot drift. The self-test's invariant is armed before it is read (control line matches the pre-fix form first) and excludes `run.sh` / `test_*.sh` BY NAME, so the row states what it does not check.

**Relayed, not re-derived:** the 14 domain files being exactly line-neutral (the merged-tree budget gate being green is the consequence I did check); the author's shim-driven four-row table; shellcheck rc 0 over the Makefile grouping (`lint-sh` is the serialized OOM path; CI's Shell tests job is green at this head).

**Not done here:** #1705 closes by hand after the merge. The 4 unconverted sites in `tests/stack/run.sh` (3) and `test_data_reset.sh` (1) are outside this grant and named in the invariant's exclusion — they want a follow-up issue against their owner; not filed by me this cycle.

Issue #1722, opened 2026-09-03
build-pithead: GNU-only chmod --reference silently truncates --self-test on macOS (cases 8-14 never run)


VijitSingh97:
## What lies

`scripts/build-pithead.sh` preserves the artifact mode in `write_artifact()` with `chmod --reference="$ARTIFACT" "$tmp"`. `--reference` is GNU coreutils only; BSD/macOS `chmod` rejects it (`chmod: illegal option -- -`). The call predates #1463 (it sat at line 152 before that change; now around line 228) and is unrelated to it.

On macOS the consequence is worse than a failing rebuild. `--self-test` case 8 ("a rebuild preserves the artifact's executable bit", around line 384) invokes the build as `PITHEAD_BUILD_ROOT="$tmp" bash "${BASH_SOURCE[0]}" >/dev/null 2>&1` with no `|| rc=$?` or `|| true` guard. Under `set -euo pipefail` the chmod failure aborts the whole self-test **silently**: stdout and stderr are already redirected, so the script stops after case 7's output with exit 1 and no message. Cases 8 through 14 never run on macOS, including the two #1463 added (dup-readonly, trap-before-def). A developer reads seven cases passed and nothing says seven of fourteen were skipped.

Found during the #1463 work on a macOS machine and reproduced identically on an unmodified `origin/develop-v2` checkout. CI runs on ubuntu-latest, so CI never sees it: this is a gate that lies only where contributors run it by hand.

## Two fixes, together

1. **Make mode preservation portable, or decide it is not load-bearing.** The artifact is always meant to be executable; if a fixed `chmod 0755` is always correct here, that is the whole fix. If an arbitrary pre-existing mode genuinely must survive a rebuild, read it portably (`stat -c %a` on GNU, `stat -f %Lp` on BSD) and apply it, but check first whether anything relies on that.
2. **Guard the case-8 invocation** in `self_test()` with `|| rc=$?` and assert on `rc`, the way every other case does, so a future failure there is loud instead of truncating the self-test output.

## Proof of done

`bash scripts/build-pithead.sh --self-test` on macOS reaches "all cases passed" (14 cases) or prints a clearly named FAIL; the targeted `shellcheck --severity=warning` and `shfmt -d` on the touched file are clean; CI unchanged. On a 9 GB machine do not run the whole `make lint` or `make test`, and never `shellcheck tests/stack/run.sh`.

[...]

PR #1727, opened 2026-09-03
fix(build): #1722 build-pithead sets 0755 outright so macOS gets a FAIL, not a truncated self-test

VijitSingh97:
## The defect

`scripts/build-pithead.sh --self-test` truncated itself silently on macOS. Case 8 (the rebuild-keeps-the-executable-bit check) ran the build unguarded — no `|| rc=$?` — with both streams already sent to `/dev/null`. On macOS `write_artifact` reached `chmod --reference="$ARTIFACT"`, which BSD `chmod` does not have, the build exited 1, and `set -e` aborted the whole self-test right there: cases 9 onward never ran, no `FAIL` line, no `FAILED`/`passed` summary, and the only signal was a bare `rc=1` with the last printed line reading `ok`.

## What changed

- `write_artifact` sets `0755` outright. The artifact is tracked at that mode, an operator runs `./pithead`, and `release.sh` bundles it as-is — there was never a mode to carry across, and `chmod --reference` is GNU-only.
- Case 8 is guarded like every other case: the rebuild's rc is captured and asserted through `_case`, then the executable bit is asserted as its own case. A rebuild that fails now prints a `FAIL` line and the summary says `FAILED`, instead of the self-test stopping mid-list.

This is a guard, not a grep: with the GNU-only `chmod` put back (and case 8 left guarded) the self-test prints `FAIL — a rebuild over an existing artifact exits 0 (expected rc=0, got rc=1)`, ends with `build-pithead --self-test: FAILED`, and exits 1 — and it cannot be green if a consumer ignores the exit code, because the `FAIL` line and the summary are on stdout.

## Proof

All on this mac (Darwin 25.5.0, BSD chmod: `chmod --reference=/etc/hosts /dev/null` -> `chmod: illegal option -- -`), all in the worktree.

Repro on the unmodified `origin/develop-v2` script:

```
rc=0; bash scripts/build-pithead.sh --self-test >repro.txt 2>&1 || rc=$?; echo rc=$rc; wc -l repro.txt
```

-> `rc=1`, 8 lines; the last line printed is `  ok   — --check REFUSES a slice with a trailing blank line (the separator is the build's)` (case 7), no `FAIL` line, no `FAILED`/`passed` summary line.

On this branch:

- `--self-test` -> `rc=0`, 24 lines, `build-pithead --self-test: all cases passed` (all 23 cases print, including the two halves of case 8 and cases 9-12).
- `--check` -> `rc=0` (the committed artifact still matches the slices; nothing under `lib/pithead/` was touched).
- Mutant (GNU `chmod --reference` restored in `write_artifact`, case 8 guarded): `rc=1`, 24 lines, line 9 is the `FAIL` above, last line `build-pithead --self-test: FAILED`.

Refs #1722, #1463

### DIFF
--- a/scripts/build-pithead.sh
+++ b/scripts/build-pithead.sh
@@ -127,8 +127,8 @@ build() {
 
 write_artifact() {
     local tmp
-    # Build beside the artifact, then rename over it. Two reasons, and the mode is carried across
-    # explicitly so nothing is given up by not writing in place:
+    # Build beside the artifact, then rename over it. Two reasons, and the mode is set explicitly
+    # so nothing is given up by not writing in place:
     #
     #   - `cat >"$ARTIFACT"` truncates at OPEN time, before a single byte is written and whatever
     #     `set -e` does afterwards. A rebuild interrupted at that instant — Ctrl-C, a full disk, a
@@ -148,11 +148,9 @@ write_artifact() {
     # shellcheck disable=SC2064  # expand now: the trap must name THIS file, not whatever $tmp is later
     trap "rm -f -- '$tmp'" EXIT
     build >"$tmp"
-    if [ -e "$ARTIFACT" ]; then
-        chmod --reference="$ARTIFACT" "$tmp"
-    else
-        chmod 0755 "$tmp"
-    fi
+    # Always 0755: the artifact is tracked at that mode, an operator runs `./pithead`, release.sh
+    # bundles it as-is, and `chmod --reference` (carrying a mode across) is GNU-only — macOS refuses.
+    chmod 0755 "$tmp"
     mv -f "$tmp" "$ARTIFACT"
     echo "build-pithead: wrote $ARTIFACT from $(list_slices | wc -l | tr -d ' ') slice(s)."
 }
@@ -304,15 +302,16 @@ self_test() {
         rm -rf "$blank"
     done
 
-    # 8. A rebuild is idempotent and keeps the artifact's mode — an operator runs ./pithead.
+    # 8. A rebuild over an existing artifact exits 0 and leaves it executable — an operator runs
+    #    ./pithead. Guarded like every other case: unguarded, `set -e` aborted the whole self-test
+    #    here with both streams already redirected, so cases 9+ silently never ran (macOS, #1722).
     chmod 0755 "$tmp/pithead"
-    PITHEAD_BUILD_ROOT="$tmp" bash "${BASH_SOURCE[0]}" >/dev/null 2>&1
-    if [ -x "$tmp/pithead" ]; then
-        echo "  ok   — a rebuild preserves the artifact's executable bit"
-    else
-        echo "  FAIL — a rebuild dropped the artifact's executable bit"
-        fail=1
-    fi
+    rc=0
+    PITHEAD_BUILD_ROOT="$tmp" bash "${BASH_SOURCE[0]}" >/dev/null 2>&1 || rc=$?
+    _case "a rebuild over an existing artifact exits 0" 0 "$rc"
+    rc=0
+    [ -x "$tmp/pithead" ] || rc=1
+    _case "a rebuild preserves the artifact's executable bit" 0 "$rc"
 
     # 9. The slice order is LC_ALL=C LEXICAL, not numeric or version ordering. NO case above can
     #    see this: 00/10/99 sort identically under `sort` and `sort -V`, so a mutant that swapped


VijitSingh97:
**Non-author reviewer pass at `1195d155` — CLEAN PASS.** Every load-bearing claim in the body reproduces independently, and the completeness question the body does not ask ("is case 8 the only site of this shape?") is answered by measurement below.

## I reproduced the macOS defect on Linux, with a fired positive control

I have no mac, so I stood in a `chmod` shim that refuses `--reference` exactly as BSD does, and confirmed the shim itself before trusting anything it produced:

```
$ PATH=$SHIM:$PATH chmod --reference=/etc/hosts /dev/null
chmod: illegal option -- -                       rc=1     # control fires
```

Base `origin/develop-v2` @ `5ff0cd48`, under the shim — the #1722 defect, and it matches the body's macOS report to the line:

```
rc=1 · 8 lines · last line printed is case 7 · real FAIL lines: 0 · summary lines: 0
```

The head, same shim:

```
rc=0 · 24 lines · build-pithead --self-test: all cases passed
```

## It is a guard, not a grep — mutation armed by content, not by exit code

Restored the GNU-only `chmod --reference` branch in `write_artifact`, left case 8 guarded, asserted the edit landed at exactly one site before running it:

```
ARMED: mutation landed, 1 site
$ grep -n 'chmod --reference' mut1727.sh
154:        chmod --reference="$ARTIFACT" "$tmp"
$ PATH=$SHIM:$PATH bash mut1727.sh --self-test
  FAIL — a rebuild over an existing artifact exits 0 (expected rc=0, got rc=1)
build-pithead --self-test: FAILED
rc=1 · 24 lines · real FAIL lines: 1
```

24 lines under the mutant, not 8 — cases 9+ still ran. That is the actual repair: the failure is now *reported* instead of *truncating the suite*. The body's claim is accurate to the word.

## The load-bearing claim — "the artifact is tracked at that mode" — CONFIRMED

```
$ git ls-tree 1195d155 pithead ; git ls-tree 5ff0cd48 pithead
100755 blob a3b8259a7370d6699fb88cd855702ea1655525ff	pithead
100755 blob a3b8259a7370d6699fb88cd855702ea1655525ff	pithead
```

Tracked at 0755 on both sides, identical blob — this PR does not touch the artifact, and `chmod 0755` outright writes precisely the mode git already records. Nothing is given up by dropping the `--reference` branch.

## Completeness — the question the body does not ask, and it comes back clean

**Was case 8 the only unguarded self-invocation?** Yes, and this is the whole class, not a sample. Ten invocations of the script from inside `self_test`; at the base exactly **one** was unguarded — line 309, the one fixed — and at the head all ten carry `|| rc=$?` or `|| true`. There is no second site waiting to truncate the suite the same way.

**Any other GNU-only call of the same class?** Comment lines stripped, word-anchored, with the control fired:

```
base 5ff0cd48   -> 1 hit:  chmod --reference="$ARTIFACT" "$tmp"      # control fires
head 1195d155   -> 0 hits                                             # clean
head 798ffbe4   -> 1 hit   (that is #1704's branch, which still carries it)
```

Disclosure on my own instrument: my first pass at this sweep matched `tac` and `-a` as substrings inside prose ("in**tac**t", "abor**t**s"), so it reported hits that were not calls. I redid it word-anchored and comment-stripped; the figures above are from the corrected run.

## Budget — clean, and correctly needs no `.tsv` edit

`scripts/build-pithead.sh` goes 437 -> **436** lines. `TARGET_LINES=400`, so the row is required (an entry is owed IFF over target) and correctly stays at its 437 ceiling; 436 is under the ceiling, and `check_monotonic` only refuses *raises*. No `docs/dev/file-budget.tsv` change is owed here and none was made. Correct.

## Named, and deliberately not a finding

A rebuild now forces 0755 rather than carrying whatever mode the local artifact had, so an operator who had tightened their local `./pithead` to 0700 gets it silently widened back on the next build. Since the tracked mode *is* 0755 and this is public source, that restores the recorded mode rather than exposing anything — worth knowing, not worth changing.

## Sequencing with #1704 (same file)

Disjoint hunks — #1704 at develop-v2 lines 20 / 69 / 117 / 336 / 363 / 387 / 418, #1727 at 127 / 148 / 304 — so no textual conflict whichever lands first. #1704's branch still carries the `chmod --reference` line (measured above), and its author had to shim around it locally to run the self-test; landing **#1727 first** buys that branch a clean macOS run.

`Refs #1722, #1463` is the right form — this lands on `develop-v2`, so `Closes` never fires here and **#1722 closes by hand** after merge.

## What I did NOT do

**No mac.** The shim reproduces BSD's *refusal of this one flag*, which is the exact mechanism of #1722 and nothing more — it says nothing about any other BSD/macOS difference, and a real mac run remains the author's evidence, not mine. I watched no CI to completion, took no lint lock, and ran no `make` target: `--check` and the lint gates stand as the author's. I did not re-derive the 23-case inventory beyond the counts above.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `1195d155`

Issue #1723, opened 2026-09-03
selftest-redact.sh's derived population is blind to every array field, so the three array-borne secrets render_masked_config masks are unclassified


VijitSingh97:
`selftest-redact.sh` derives its classification population by walking `config.reference.json`'s
values, and its stated guarantee is written into the file:

> The population below is DERIVED from config.reference.json rather than listed here, so a
> sensitive field added to the schema later cannot quietly go uncovered.

That guarantee does not hold for anything inside an array, and today the schema has four arrays —
three of which hold values the product itself masks as secrets.

## The two mechanisms, measured

**1. An empty array contributes no path at all.** `walk()` recurses into a list's elements; a list
with no elements yields nothing. All four arrays in `config.reference.json` ship empty:

```
ARRAY len=0  dashboard.workers
ARRAY len=0  workers.list
ARRAY len=0  telegram.control.allowed_ids
ARRAY len=0  notifications.webhooks
```

So none of them reaches `$SCREENED`, none is classified, and the `field is classified` alarm
cannot fire for any of them.

**2. The screen is `$`-anchored and the walk appends `[]`, so a scalar-element array can never
match a suffix — whatever its name.** `walk()` builds `path + "[]"` for list elements and the
screen tests `path.split(".")[-1]`, so the last segment ends in `[]` and no alternative in
`(password|...|urls?|...|_id)$` can match it.

Measured against a copy of the reference with the arrays populated, one sentinel per leaf:

| path | screened? |
|---|---|
| `notifications.api_token` (scalar, same object) | **yes** |
| `notifications.webhook_urls[]` (key ends in the screened word `urls`) | **no** |
| `notifications.webhooks[]` | no |
| `telegram.control.allowed_ids[]` | no |
| `workers.list[].token` | yes |
| `workers.list[].url` | yes |

The first two rows are the discriminating pair: same object, same run, one screens and one does
not, and the only difference is that the second is a list element. Mechanism 2 spares
object-element arrays, because their leaf path ends in the field name rather than in `[]` — which
is why it is worth stating separately from mechanism 1.

## What is uncovered right now

`render_masked_config` (`lib/pithead/30-release-fetch-and-masked-config.sh:127-148`) masks three
array-borne secrets by path, each with its own stanza and its own issue:

- `workers.list[].token` — per-worker API tokens (#172)
- `dashboard.workers[].token` — the deprecated fallback shape
[...]

Issue #172, opened 2026-06-06
Configurable worker ports & endpoints: non-standard stratum/API ports + per-worker host/port/token overrides (later)


VijitSingh97:
## RESCOPED — v1.5 Wave 0 (2026-07-11)

The original body (below, superseded) was 2/3 stale per the 2026-07-07 audit: `workers.api_port` shipped as a config knob, and `XMRIG_API_PORT`/`PROXY_API_PORT` are already env-overridable (`config.py:52/168`). And #171 shipped **fleet-wide**, not as a per-worker list — so there is no per-worker list to "generalize"; it must be built from scratch. Executable scope:

### 1. `p2pool.stratum_port` (default `3333`) — the only truly-hardcoded value left
Thread the configured port through: xmrig-proxy `-b` bind, the compose publish (`…:PORT:PORT`), the `:PORT` healthcheck, and `P2POOL_URL` in the `.env` render.
**Decision:** p2pool's *internal* `--stratum 0.0.0.0:3333` stays fixed at 3333 — it's container-internal (xmrig-proxy dials it on the bridge), never operator-facing, so exposing it buys nothing. Only the operator-facing proxy stratum port is configurable. RigForge workers must point at the same port (rigforge#21).

### 2. Per-worker endpoint descriptor — NEW, from scratch
```json
"dashboard": { "workers": [ { "name": "rig-01", "host": "10.0.0.5", "port": 8080, "token": "…" } ] }
```
- Every field optional. **Merge rule (explicit):** per-worker field **>** fleet default (`dashboard.worker_api_port`, `workers.api_auth` token) **>** inherit. A per-worker `token` implies token-auth *for that worker only*, regardless of the fleet-wide `api_auth` mode.
- **Match by `name`** (the stratum worker name). On a name miss, fall back to matching by connecting IP. Duplicate names → first-declared wins; document that renames need a config update.
- `host` solves the different-hostname case (NAT / multi-homed / API on a different interface than the stratum IP).

### 3. SSRF guardrail (carry #171's rule)
A per-worker `host` MUST be **operator-set in config.json** — never derived from a miner-advertised value. The dashboard never sends a configured token to a miner-controlled host (#122).

### Cross-repo (acceptance is end-to-end, both sides)
- rigforge#21 (`pool.port`) and rigforge#23 (`api.port`) must carry matching knobs. File/confirm those trackers when Wave 1 opens.

### Acceptance
[...]

PR #1728, opened 2026-09-03
fix(tests): #1723 the redact selftest classifies an array on its own account, so an array-borne secret cannot enter the schema unseen

VijitSingh97:
Closes nothing automatically — #1723 is closed by hand at merge, as this repo's PRs land on
`develop-v2` while the default branch is `develop`.

## What was wrong

`selftest-redact.sh` derives its classification population by walking `config.reference.json`'s
values, and promises in its own comment that "a sensitive field added to the schema later cannot
quietly go uncovered". That promise did not extend to anything inside an array, for two
independent reasons:

1. **An empty list yields no element to screen.** All four arrays in the schema ship empty, so not
   one of them reached the population.
2. **A list element's path ends in `[]`, which no `$`-anchored suffix rule can match** — so a
   scalar-element array is unclassifiable by name even when populated.

Three of those four arrays carry values `render_masked_config` masks by path:
`workers.list[].token` (#172), `dashboard.workers[].token`, and `notifications.webhooks[]` (#848).

**Severity is low and I want to be exact about it.** The behaviours are stated elsewhere —
`selftest-capture-config.sh:99` asserts the webhook gap, tier 1 covers the worker tokens and the
standby source. What was missing is the *drift alarm*: the only check that fires when an
unclassified sensitive field enters the schema, and so the only one that catches the next one. My
first framing on the issue overstated this; [corrected there](https://github.com/p2pool-starter-stack/pithead/issues/1723#issuecomment-5533464748).

## What was RUN

`selftest-redact.sh` **58 passed, 0 failed** (55 before this change). Every other integration
selftest, unchanged: `selftest.sh` 214/0, `selftest-skip-accounting.sh` 76/0 — that one matters,
since it fails a skip leaving through a bare `it_warn` and this change adds two —
`selftest-e2e-phases.sh` 80/0, `selftest-capture-config.sh` 8/0, `selftest-zmq-probe.sh` 69/0,
`selftest-redact-vocab.sh` 37/0, `selftest-redact-env.sh` 15/0, `selftest-redact-ip.sh` 21/0,
`selftest-rigforge-writable-keys.sh` 57/0, `selftest-abort-traps.sh` 55/0, and the remaining nine
all 0 failed.

`shellcheck -x` (0.11.0, the CI pin) rc 0; `shfmt -i 4 -d` over the Makefile's exact glob rc 0, no
diff; `make lint-md` 0 errors; `make lint-docs-voice` OK; `make lint-file-budget` OK, run after
`git add` since the ratchet is blind to untracked files. `make lint-sh` was NOT run — it is the
serialized OOM path on this box; the targeted shellcheck + full-glob shfmt above is the sanctioned
substitute.

### Mutation battery — every new row shown able to fail

| mutation | expected | observed |
|---|---|---|
| drop `notifications.webhooks[]` from `KNOWN_GAP` | classification alarm fires | ✗ `field notifications.webhooks[] is classified` — 56/1 |
| add a NEW array `notifications.webhook_urls` to the schema, **this commit** | fails as unclassified | ✗ `field notifications.webhook_urls[] is classified` — 57/1 |
| the same schema, **parent commit** | silent — the defect | 55 passed, **0 failed** |
| give `redact()` an array-aware `webhooks` arm | the KNOWN GAP row reds | ✗ `KNOWN GAP (#848)` — 56/1 |
| give `redact()` an `allowed_ids` arm | the survivor row reds | ✗ `telegram.control.allowed_ids[] survives redaction` — 56/1 |
| give `redact()` a `source` arm | the KNOWN GAP row reds | ✗ `KNOWN GAP (#1723)` — 57/1 |
| drop `xvb.standby.source` from `KNOWN_GAP` | classification alarm fires | ✗ `field xvb.standby.source is classified` — 57/1 |
| remove the `ARRAY` yield from the walk | the four array rows disappear | back to 55/0 |

Each mutation was proven to have changed the file on disk (`git diff --numstat`) before its run —
two anchors in an earlier draft of this battery silently did not match, and the legs still printed
a plausible count.

The discriminating pair is rows 2 and 3: identical mutated schema, one commit apart, under a key
ending in `urls` — a word the screen explicitly carries. It is invisible only because it is a list
element.

There is also a control worth keeping in view: `"webhook_urls": "…"` redacts, and
`"webhook_urls": ["…"]` under that very same key comes through raw. The list shape alone defeats
the filter, independent of the name.

## Disclosures

**More than the issue asked for, deliberately.** Cross-checking the product's own
`CONTROL_SECRET_PATHS` against this file's lists turned up a third unclassified secret by a
different mechanism: `xvb.standby.source`. The masked-config pass treats it as a secret,
`redact()` carries no suffix reaching the bare word `source`, and the screen could not see it to
say so. The screen now carries `source`; **measured**, that widening reaches exactly one field and
no other. Leaving a known product-secret unclassified while editing this exact machinery was not
defensible. The cross-check is 12/12 after this change and 11/12 before it.

**Shared files.** `docs/dev/integration-testing.md` and `docs/dev/testing-guide.md` under
`_shared.paths`: the first said one field is out of the filter's reach and now says three; both
stated the drift guarantee without its array exception. `selftest-redact-env.sh`'s comment said
"the same four fields" are pinned as JSON survivors, which this change makes five.

**Line budget.** `selftest-redact.sh` goes 337 → 385 against `lint-file-budget`'s 400-line target,
so it still takes no budget row — 15 lines of headroom. That file is deliberately kept under the
target by splitting (`selftest-redact-vocab.sh`, `-env.sh`, `-ip.sh` all exist for this reason);
whoever needs the next 25 lines should split rather than raise.

- **The lists are not proven disjoint.** A field in two of the four gets the first branch's
  treatment silently. Pre-existing — `MUST_REDACT` already won over `MUST_SURVIVE` this way — and
  my change adds a fourth list, so it is one degree worse. Mitigated rather than fixed: the new
  list is checked LAST, so it can never override an assertion a stronger list would have made. A
  real fix is a disjointness assertion, which is machinery this PR does not need.
- **`ELEMENT_SHAPE_UNKNOWN` asserts nothing.** For an array of objects an empty reference document
  carries no element shape, so there is nothing honest to assert; a passing row would read as
  coverage it does not have. The list buys the classification *requirement*, not a measurement,
  and the warn says `NOT MEASURED` in those words. Worth watching that it does not become a
  dumping ground.
- **The closing `it_warn` interpolates the static `$KNOWN_GAP` list**, so if someone removes the
  array yield the warning names a row that no longer runs. Visible in the last battery row.
  Pre-existing shape; not worth machinery.
- **Comment volume against a hard ceiling** — the one thing the over-engineering pass acted on.
  This file's house style is heavily commented and the reasons are load-bearing, but 20 added
  comment lines against 15 remaining lines of headroom is the wrong trade. Cut back by 2 lines
  where the prose restated the code rather than explaining it. The `gap_ref` case (4 lines, to put
  an issue number in the row label) survived the pass as the most optional thing still in the
  diff.
- **The cross-check that found `xvb.standby.source` is not mechanical.** It should be: nothing
  today asserts that every `CONTROL_SECRET_PATHS` entry is classified in this file. Filing
  separately rather than growing this file into its remaining headroom.

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -98,9 +98,15 @@ The test box holds real synced nodes and real keys. Treat it as production-sensi
   reaches the 48-character form, and no bound reaches the emoji form at all.
   **The JSON rule is keyed on the field NAME, not the value, and that is forced**:
   a view key and a container digest are both 64 hex characters, so nothing in the value separates
-  them. **The name list is open, and one field is out of the FILTER's reach** — `notifications.ntfy.url`
+  them. **The name list is open, and three fields are out of the FILTER's reach** — `notifications.ntfy.url`
   is a capability URL whose key is the bare word `url`, which only its nesting separates from the
-  public `xvb.url`, and a line-wise filter cannot see nesting. That is a limit of the filter, not
+  public `xvb.url`, and a line-wise filter cannot see nesting. `notifications.webhooks[]` is the
+  same kind of URL one shape over: the rule needs the key's colon to be followed by the value's own
+  quote, and a JSON list puts a bracket in between. The list shape alone is enough to defeat it —
+  `webhook_urls` sits in the redactor's vocabulary, and a value written as `["…"]` under that very
+  key still comes through raw. `xvb.standby.source` is the third, and it is a plain name gap: the
+  masked-config pass treats it as a secret, and no suffix in either vocabulary carries the bare
+  word `source`. That is a limit of the filter, not
   of the bundle: since [#1630](https://github.com/p2pool-starter-stack/pithead/issues/1630) the
   `config.json` artifact is masked BY PATH before it reaches the filter (*Artifacts & triage*
   below), so the topic does not survive a capture. The self-test states the filter's gap rather
@@ -755,7 +761,13 @@ than that a `<redacted>` marker appeared, since a marker can come from some othe
 same line. Its JSON population is derived from `config.reference.json` rather than
 listed in the file, under a screen deliberately wider than the redactor's own list, so a sensitive
 field added to the schema fails the test by name until someone classifies it as redacted or as a
-stated survivor. It also guards the other direction: a sha256 digest, a non-secret flag value, the
+stated survivor. Arrays are classified on their own account rather than by name
+([#1723](https://github.com/p2pool-starter-stack/pithead/issues/1723)): an empty list yields no
+element for the screen to read, and a list element's path ends in `[]`, which no suffix rule can
+match, so all four of the schema's arrays sat outside that guarantee — three of them carrying
+secrets the masked-config pass covers by path. Checking that pass's own path list against this
+file's classification is what turned up `xvb.standby.source`, which the screen now reaches; the
+check itself is not yet mechanical here. It also guards the other direction: a sha256 digest, a non-secret flag value, the
 reserved IP ranges and a HAProxy timestamp must survive, because an artifact with its image pins or
 its port map stripped is useless for the triage it exists for. That timestamp is not a hypothetical
 near-miss: HAProxy's `DD/Mon/YYYY:HH:MM:SS` reads as an IPv6 address, because a four-digit year
--- a/docs/dev/testing-guide.md
+++ b/docs/dev/testing-guide.md
@@ -133,7 +133,8 @@ structurally cannot.
 - Secrets: never print tokens, creds, or onions. The harness redacts artifacts and hashes secrets
   on the box. If you add a secret-bearing field to `config.reference.json`,
   `tests/integration/selftest-redact.sh` fails until you classify it — either `redact()` covers it,
-  or the file records why it is safe to keep.
+  or the file records why it is safe to keep. An array you add is classified as an array, whether
+  or not the reference populates it.
 
 ## Gotchas learned on real hardware
 
--- a/tests/integration/selftest-redact-env.sh
+++ b/tests/integration/selftest-redact-env.sh
@@ -96,7 +96,7 @@ case "$OUT" in
 esac
 
 echo "== redact: the KEY=value over-redaction guard — debugging value must survive =="
-# The same four fields selftest-redact.sh pins as JSON survivors, in the spelling `.env` uses.
+# The scalar fields selftest-redact.sh pins as JSON survivors, in the spelling `.env` uses.
 # Holding the two syntaxes to one vocabulary means holding them to the same survivors too: a
 # bundle with its endpoints and routing ids stripped is useless for the triage it exists for.
 # `XVB_URL` is the sharp one — only `PING_URL` is in the vocabulary, so a bare `_URL` survives.
--- a/tests/integration/selftest-redact.sh
+++ b/tests/integration/selftest-redact.sh
@@ -167,7 +167,9 @@ ssh.authorized_key healthchecks.ping_url telegram.bot_token notifications.ntfy.t
 #   xvb.url / xmrig_proxy.url  public service endpoints
 #   workers.api_auth           an auth MODE string ("token"/"none"), not a credential
 #   telegram.chat_id           a routing id, not a secret
-MUST_SURVIVE="xvb.url xmrig_proxy.url workers.api_auth telegram.chat_id"
+#   telegram.control.allowed_ids[]  routing ids, the same class as telegram.chat_id
+MUST_SURVIVE="xvb.url xmrig_proxy.url workers.api_auth telegram.chat_id
+telegram.control.allowed_ids[]"
 # ⛔ NOT a survivor on merit. `notifications.ntfy.url` IS a capability URL and ought to be
 # redacted; a line-wise filter cannot reach it, because the key is the bare word "url" and only
 # its NESTING distinguishes it from xvb.url. Asserted at its CURRENT behaviour so the gap is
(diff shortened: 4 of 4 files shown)


VijitSingh97:
**Non-author reviewer pass at `2bd804a9` — PASS. One non-blocking finding.** Every number in the body reproduced on my own runs; the discriminating pair reproduced with arming on **both** legs.

## Ruling 8a and the record

```
budget at head 2bd804a9            file budget OK — ... monotonic.   rc=0
budget on the merged tree          file budget OK — ... monotonic.   rc=0
merge-tree(develop-v2, 2bd804a9)   rc=0 — clean
leak scan over the whole 191-line diff   clean, with a fired positive control (synthetic address -> 1 hit)
```

## The claim I was asked to re-derive: "that widening reaches exactly one field"

Confirmed, by running both screens over `config.reference.json` and diffing the populations rather than trusting the count:

```
scalar fields matched by the OLD screen: 22
scalar fields matched by the NEW screen: 23
DELTA (new-only): ['xvb.standby.source']        delta count: 1
```

## The discriminating pair — reproduced, both legs armed

```
HEAD 2bd804a9  + notifications.webhook_urls: []  (arming read back: True [])
  ✗ config.reference.json field notifications.webhook_urls[] is classified
  selftest-redact: 58 passed, 1 failed

PARENT 8e248b1e + the IDENTICAL mutated schema  (arming: 'ARRAY = object()' count 0 => parent file)
  selftest-redact: 55 passed, 0 failed          <- silent. the defect.

restored: 0 modified files; baseline back to 58 passed, 0 failed
```

That is the row that matters, and it is the one the body singles out: same schema, one commit apart, under a key ending in `urls` — a word the screen explicitly carries — invisible only because it is a list element.

**The stated mitigation is real too.** A populated object-array element line is reached by the name rule: `"token": "S3nt1nelVALUE"` → `"token": "<redacted>"`. And I measured the follow-on rather than reasoning it: populating `workers.list` with one element carrying `.token` makes `workers.list[].token` re-enter the screen by name, unclassified, and **the drift alarm fires** (58 passed, 1 failed). The `ELEMENT_SHAPE_UNKNOWN` bucket does not create a hole for the populated case.

## The interaction the body flagged, and the siblings

Two new bare `it_warn` calls were the thing most likely to bite, since `selftest-skip-accounting.sh` fails a skip that leaves through one. It does not:

```
selftest-skip-accounting  76 passed, 0 failed      selftest-redact-env     15/0
selftest-redact-vocab     37 passed, 0 failed      selftest-redact-ip      21/0
selftest-capture-config    8 passed, 0 failed      selftest-redact         58/0
```

All match the body exactly.

## A shape check I ran because this PR raises the argument itself

The diff correctly rejects a scalar probe over an array field as *"a line the bundle never contains"*. So I asked the same question of the new probe: it builds `"key": ["S3nt1nelVALUE"],` on one line, while a real pretty-printed bundle puts array elements on their **own keyless lines**. Different shape — so I measured both instead of assuming they agree:

```
key            probe(1-line)   real(multi-line)
webhooks       SURVIVES        SURVIVES     same
allowed_ids    SURVIVES        SURVIVES     same
webhook_urls   SURVIVES        SURVIVES     same
token          SURVIVES        SURVIVES     same
positive control: "token": "…" -> <redacted>   "webhook_urls": "…" -> <redacted>
```

Identical verdict in both shapes, and the control fires, so the row's conclusion is right. Worth one line in the file anyway: the probe is not the bundle's shape, it just happens not to change the answer — a keyless element line is unreachable by a name rule in the same direction the bracket is.

## F1 — the two array buckets are mechanically INDISTINGUISHABLE (non-blocking)

`"webhooks": []` and `"list": []` are the same value. The probe built for each is the same string. Both survive. The only thing separating *"measurable scalar array — assert it"* (`KNOWN_GAP`) from *"unmeasurable object array — warn"* (`ELEMENT_SHAPE_UNKNOWN`) is knowledge of what the element **would** be, and nothing in the file or the schema checks it. The weaker bucket is the one that asserts nothing, which makes it the easier place for a future array to land: a new scalar array put there gets a permanent warning instead of the real assertion `notifications.webhooks[]` proves is available.

Both current members genuinely are object arrays, so this is not a defect today — and the populated case is covered, as measured above.

**The sharp end, which I did find by measuring:** the row prints *"the reference carries no element"* — a claim about the document that is **not derived from the document**. In my populated-`workers.list` run it printed that sentence while the reference carried an element. One edit makes the row state something false. Cheapest fix is to derive the clause (or drop it); a stronger one is to have the row record that the probe *was* run and survived, so it carries a measurement rather than an absence.

## Docs — accurate

`integration-testing.md` goes from *"one field is out of the FILTER's reach"* to *"three fields"* and names each with its distinct mechanism (nesting, list shape, bare name); the array guarantee and its `[]`-suffix reason are documented; `testing-guide.md` gains the one-line array rule. Both read in the house voice. The `_shared.paths` use is disclosed in the body, as it should be.

## On the author's self-correction

The public correction on #1723 is confirmed and was the right call: the behaviours **are** stated elsewhere (`selftest-capture-config.sh:99`, tier 1), what was missing is the **drift alarm**, and that is exactly what the mutation pair above measures. Lowering your own severity after filing is the harder half of the work.

## What I did NOT do

No `make lint-sh` (the serialized OOM path) and no lint lock — the targeted `shellcheck -x` and full-glob `shfmt` in the body stand as yours. I ran the six selftests listed above, not all twenty-odd; the remaining counts are yours. I did not re-derive the other seven mutation rows — I took the discriminating pair, both legs, plus the mitigation and the populated-array follow-on.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `2bd804a9`

VijitSingh97:
**F1 is filed as #1748**, per the ruling that it gets its own issue rather than being folded here.

I re-derived both halves at `aeef7d83` before filing, and the scope moved: the finding above names
TWO indistinguishable buckets, and measuring all four arrays in the schema gives **three** —
`MUST_SURVIVE` (`telegram.control.allowed_ids[]`) produces the identical probe and the identical
SURVIVES verdict, and was not named. The conclusion is unchanged, which is what makes a wrong figure
hard to spot.

The sharp end reproduces exactly as described: populating `workers.list` with one element leaves the
`ELEMENT_SHAPE_UNKNOWN` row still printing *"the reference carries no element"* while the drift alarm
reds on `workers.list[].token` beside it. Both legs, arming read back, reference restored and
confirmed by `sha256sum`. Figures are on #1748.

PR #1729, opened 2026-09-03
fix(tests): #1679 selftest-mergemine-probe.sh defines its stubs inside the subshell that calls them, so no call precedes a definition

VijitSingh97:
## What changed

`tests/integration/selftest-mergemine-probe.sh` defined its two stubs — `monero_caught_up` and `mm_capture_startup` — at the **top level** of the file, textually after the case at `:159` that deliberately calls the **production** `mm_capture_startup`. Both definitions move inside the subshell in `mm_leg_outcome()`, which is the only place that ever called them.

One file. No behaviour change: nothing outside that subshell ever used the stubs, so scoping them to it is the same program.

## Why — structure, not a gate fix

**This does not fix a red gate, and selling it as one would be false.** `make lint-sh` at the pinned shellcheck was never red on this file.

Measured on both revisions over `tests/integration/*.sh`, so that the sourced provider — `mergemine-probe.sh`, which defines `mm_capture_startup` — is named on the same command line and the `# shellcheck source=` directive actually resolves:

| tree | shellcheck 0.9.0 | shellcheck 0.11.0 (the pin) |
|---|---|---|
| merge-base `48e0546d` | **1 finding — SC2218 at `:159`**, rc 1 | 0 findings, rc 0 |
| head `db55524a` | 0 findings, rc 0 | 0 findings, rc 0 |

The non-author reviewer reproduced this independently with a different invocation and `-f gcc` output, landing on the same single hit at the same line, and corrected an inherited figure of two down to one (comment 5533529493). Two measurements, two invocations, one answer.

That invocation shape is load-bearing and worth stating, because getting it wrong manufactures findings: running this file **alone** additionally reports `SC2034 IT_SKIPPED_MISSING appears unused`, which does not exist under `make lint-sh` — the reader lives in `lib.sh`, and shellcheck only sees it when `lib.sh` is on the same command line.

So SC2218 here is a **distro-0.9.0-only** diagnostic, and a no-op for CI. `Makefile:52` pins `SHELLCHECK_VERSION := 0.11.0`, and `Makefile:72-75` makes `lint-sh` refuse any other version before it lints anything:

```
$ env PATH=/usr/bin:/bin make lint-sh
lint-sh: shellcheck 0.9.0 is not the pinned 0.11.0 — its findings are not this gate's.
lint-sh: install the pin (see docs/dev/release-server.md) or run the gate in CI.
```

The change is therefore hygiene for anyone running a distro shellcheck by hand — plus the argument that actually justifies it:

**The old ordering was load-bearing by accident.** The top-level stub *had* to sit textually after `:159`, or it would have captured that production call. Nothing recorded that constraint and nothing enforced it. Nesting the stubs in the subshell makes the separation structural, so a later edit that moves the stub block cannot silently re-point `:159` at a stub — which a green `40 passed, 0 failed` would not catch, because a stub there passes quietly. That is a stronger guarantee than the ordering it replaces, and it is the reason to take this change.

## That `:159` still reaches production — proven, not reasoned

Settled by the non-author reviewer with two aimed mutations on this head (comment 5533529493):

- **A** — production `mergemine-probe.sh:109`, `[ -n "$started" ] || return 1` -> `|| return 0`: kills only *an unreadable start time refuses the capture* (39 passed, 1 failed); every leg case stayed green.
- **B** — the subshell stub's `@FAIL` arm at `:195`: kills only a leg case (39 passed, 1 failed); the `:159` case stayed green.

Disjoint targets, one named kill each, zero bystanders, control `40 passed, 0 failed` either side, each edit verified landed and then restored. A says `:159` reaches the production function on this head; B says the leg section still gets the subshell stub. The pair rules out what a single mutation cannot — that both sites resolve to the same definition.

## What was NOT run

- **I did not run the suite on this head.** The `40 passed / 0 failed` control and both mutation legs are the reviewer's measurements, cited as theirs, not re-derived here.
- **No run of the Makefile's exact `lint-sh` file set under 0.9.0.** That verdict is unobservable through the gate by construction — it refuses 0.9.0 — and on this box that root is the historical OOM path. The table above is the `tests/integration/*.sh` set, run under a 6 G cap.
- One reading per cell, not a distribution.
- Nothing about macOS, `build-pithead.sh`, or `chmod --reference`. That is #1727's subject; its body sat on this PR in error until now, describing a diff this PR does not contain.

### DIFF
--- a/tests/integration/selftest-mergemine-probe.sh
+++ b/tests/integration/selftest-mergemine-probe.sh
@@ -168,26 +168,33 @@ unset -f rx
 
 echo "== the leg reports PASS, FAIL or a COUNTED skip — never a silent green =="
 
-# The leg is driven with its two collaborators stubbed, so all three outcomes are reachable
-# without a stack. Stubbing is what makes the skip path testable at all: on the boxes this
-# harness runs on monerod IS synced, so that branch would otherwise never be exercised here.
-monero_caught_up() { return "$MM_STUB_SYNCED"; }
-# The sentinel matters. `mm_capture_startup` has TWO distinct empty outcomes in production and
-# the leg treats them differently: it returns non-zero when the container's start time cannot be
-# read at all, and returns ZERO with empty output when the log simply held no MergeMiningClientTari
-# line (its `grep` is `|| true`). A stub keyed only on emptiness collapses them, and the case
-# named for the second silently exercises the first — which is how the `absent` branch survived a
-# mutation to `it_pass` here.
-mm_capture_startup() {
-    [ "$MM_STUB_CAPTURE" = "@FAIL" ] && return 1
-    printf '%s' "$MM_STUB_CAPTURE"
-}
-
-# Run the leg in a SUBSHELL with the harness counters zeroed, and report them as a line. Three
-# of the four cases below provoke a leg FAILURE on purpose; left in this file's own totals they
-# would make a working self-test report red, which is how a suite learns to be read past.
+# Run the leg in a SUBSHELL with its two collaborators stubbed and the harness counters zeroed,
+# and report the counters as a line. Three of the four cases below provoke a leg FAILURE on
+# purpose; left in this file's own totals they would make a working self-test report red, which
+# is how a suite learns to be read past.
+#
+# The stubs live INSIDE the subshell, not at the top level of this file. The capture section
+# above calls the REAL `mm_capture_startup`, so a top-level stub of the same name has to come
+# after those calls — and shellcheck 0.9.0 reads a call that precedes a same-file definition as
+# use-before-definition (SC2218), which is what stopped the #1653 release rehearsal (#1679).
+# Scoping the stubs to the subshell is the same behaviour: nothing outside it ever called them.
 mm_leg_outcome() { # <synced-rc> <capture> -> "<pass> <fail> <skipped-legs> <by-design> <names>"
     (
+        # The leg is driven with its two collaborators stubbed, so all three outcomes are
+        # reachable without a stack. Stubbing is what makes the skip path testable at all: on the
+        # boxes this harness runs on monerod IS synced, so that branch would otherwise never be
+        # exercised here.
+        monero_caught_up() { return "$MM_STUB_SYNCED"; }
+        # The sentinel matters. `mm_capture_startup` has TWO distinct empty outcomes in production
+        # and the leg treats them differently: it returns non-zero when the container's start time
+        # cannot be read at all, and returns ZERO with empty output when the log simply held no
+        # MergeMiningClientTari line (its `grep` is `|| true`). A stub keyed only on emptiness
+        # collapses them, and the case named for the second silently exercises the first — which
+        # is how the `absent` branch survived a mutation to `it_pass` here.
+        mm_capture_startup() {
+            [ "$MM_STUB_CAPTURE" = "@FAIL" ] && return 1
+            printf '%s' "$MM_STUB_CAPTURE"
+        }
         MM_STUB_SYNCED="$1"
         MM_STUB_CAPTURE="$2"
         IT_PASS=0 IT_FAIL=0 IT_SKIPPED_LEGS=0 IT_SKIPPED_BY_DESIGN=0 IT_SKIPPED_MISSING=0 IT_SKIPPED_NAMES=""


VijitSingh97:
## RETURNED — the body describes a different change than the diff; two checks the pass must make before this merges

**Finding 1 (merge-record, blocking): the PR body is not this PR's.** It describes `scripts/build-pithead.sh --self-test` truncating on macOS under a GNU-only `chmod --reference` (that is the #1722 change). The diff touches one file, `tests/integration/selftest-mergemine-probe.sh`, and is the SC2218 restructure. A body that narrates a different diff is exactly the false-claim-in-the-merge-record shape; replace it (`gh api -X PATCH repos/…/pulls/1729 -f body=…`, then re-read it — `gh pr edit --body-file` is broken against this repo) before any verdict is posted.

**Finding 2 (premise, must be stated in the body): SC2218 does not fire at the pinned version.** Re-derived earlier this cycle by this seat on the file at the tip: pinned shellcheck 0.11.0 (`SHELLCHECK_VERSION := 0.11.0`, `Makefile:52`) reports 0 hits, control fires 2; the two hits reproduce only on the distro 0.9.0, which `make lint-sh` now refuses outright (#1687). So this change cannot be justified as "fixes the lint gate" — the gate is not red at 0.11.0. It can be justified as a structure fix, and that is what the body must say.

**The check the pass turns on:** line 159 of the self-test deliberately calls the PRODUCTION `mm_capture_startup` (sourced at :25, defined `mergemine-probe.sh:106`) to prove the unreadable-start-time refusal; the top-level stub at :181 existed for the leg section only. The diff now nests the stub. The reviewer must show, on the head, that the :159 call still reaches the production function and not the stub — e.g. by breaking the production function's refusal and watching :159 go red — and that the leg section still gets the stub. A 40/40 self-test count does not answer this by itself; a stub in force at :159 passes silently, which is the reason the ordering was as it was.

Routing: the resident reviewer lane posts the verdict; the controller merges on it. Pinned version fact recorded on #1679.

VijitSingh97:
**Non-author reviewer pass — RETURN WITH FINDINGS.** Diffed against merge-base `48e0546d`, driven at head `db55524a`.

**The check this pass turns on is DISCHARGED — proven, not reasoned.** The blocker is the body only (the controller's Finding 1, which I confirmed independently). No defect in the diff.

## The :159 question, answered by two aimed mutations

The concern was that moving the stubs into `mm_leg_outcome`'s subshell could put a stub in force at `:159`, where the case deliberately calls the PRODUCTION `mm_capture_startup` — and that a `40 passed, 0 failed` count cannot see that, because a stub there passes silently. So I broke each side in turn and watched which case died. Each mutation's edit is shown landed before the run.

```
CONTROL   head db55524a, unmodified                    40 passed, 0 failed

MUTATION A  production mergemine-probe.sh:109
            [ -n "$started" ] || return 1  ->  || return 0
            (readback: line 109 confirmed changed, then restored)
  ✗ an unreadable start time refuses the capture       39 passed, 1 failed
  ... and every leg case stayed GREEN

MUTATION B  the subshell stub's @FAIL arm, selftest:195
            [ "$MM_STUB_CAPTURE" = "@FAIL" ] && return 1  ->  && return 0
  ✗ an unreadable container start time FAILS though the RPC says not-caught-up
                                                       39 passed, 1 failed
  ... and the :159 case stayed GREEN

RESTORED                                               40 passed, 0 failed
```

**Two mutations, one named kill each, zero bystanders, and the targets are disjoint.** A says `:159` reaches the production function on this head. B says the leg section still gets the subshell stub. The pair also rules out the thing a single mutation could not: that both sites resolve to the same definition.

Worth adding, because it makes the restructure obviously right rather than merely safe: the old ordering was load-bearing *by accident* — the top-level stub had to sit textually after `:159` or it would have captured that call. Nesting it in the subshell makes the separation structural, so a future edit that moves the stub block cannot silently re-point `:159`. That is a strictly stronger guarantee than the ordering it replaces.

**One methodology note against myself:** my first attempt at Mutation A did not apply — a `|` sed delimiter collided with the `||` in the pattern — and it returned `40 passed, 0 failed`, which reads exactly like a passing mutation leg. It was a null experiment measuring the unmutated program. Caught only by printing the line back before the run; the numbers above are from the re-run with the edit verified landed.

## Finding 2 re-derived (relayed to me, so I measured it myself)

Confirmed, with one figure corrected:

```
shellcheck 0.9.0    merge-base file   SC2218 hits: 1
shellcheck 0.9.0    head file         SC2218 hits: 0
shellcheck 0.11.0   merge-base file   SC2218 hits: 0
shellcheck 0.11.0   head file         SC2218 hits: 0
```
The one hit is `st…-probe.sh:159:1: error: This function is only defined later. Move the definition up. [SC2218]` — **at line 159, the very line the pass turns on.** 0.9.0 flagged `:159`'s call precisely because the top-level stub was defined later at `:181`.

The 0.11.0 clean is a real clean, not a scanner that no-opped: the same invocation emits **9** diagnostic lines on that file (SC1091 ×2, SC2329, …), so it ran and can speak.

**The correction:** the routing I was given said *two* hits at 0.9.0. I measure **one**, on this file, counting lines matching `SC2218` in `-f gcc` output at both revisions. The conclusion is unchanged and unaffected — 0.11.0 is clean either way, so this cannot be justified as a gate fix — but the count in the record should be one, or whoever measured two should say what else was in scope.

## Findings

**F1 — BLOCKING (merge record): the body is not this PR's.** Confirmed independently rather than taken: the body narrates the GNU-only `chmod --reference` truncating `build-pithead --self-test` on macOS, which is **#1727**'s change (`fix(build): #1722 build-pithead sets 0755 outright`) — a different open PR, one file, `scripts/build-pithead.sh`. This diff is one file, `tests/integration/selftest-mergemine-probe.sh`, +25/-18, the SC2218 restructure. Nothing in the body describes it. Replace it before merge; `gh pr edit --body-file` is broken against this repo, so `gh api -X PATCH repos/…/pulls/1729 -f body=…` and **re-read the body afterwards rather than trusting the rc**.

**F2 — BLOCKING (premise): the body must justify this as structure, not as a gate fix.** Per the table above, `make lint-sh` at the pinned 0.11.0 was never red on this file. The honest framing is the one that is actually true and is a better argument anyway: the ordering constraint was implicit and fragile, and nesting the stubs makes it structural. The replacement body should say that, and should state the SC2218 finding as 0.9.0-only.

**No third finding.** The diff itself is correct and I could not break it.

## PROVEN vs RELAYED

**Proven by me, at head `db55524a`:** the whole mutation table, each edit verified landed and restored, control 40/0 on both sides; the four shellcheck measurements with a fired positive control; the merge-base (`48e0546d`) diff; and F1 by matching the body's subject against #1727's actual title and file list.

**Not done:** I ran only this self-test — no `make lint`, no `make test`, no other suite, no bench, no container. I did not verify the body's `chmod`/macOS claims, since they belong to #1727 and I have no mac. I did not re-derive the #1687 refusal of 0.9.0 by `make lint-sh`.

VijitSingh97:
Correction to my return above, so the record is right: "the two hits reproduce only on the distro 0.9.0" was RELAYED from the release rehearsal, not measured by me. The reviewer measured it: SC2218 at 0.9.0 on this file is ONE hit, at line 159 itself; at the pinned 0.11.0 it is 0 on both revisions, with the same invocation emitting other diagnostics, so the clean is real. The :159 concern is DISCHARGED by the reviewer with two aimed mutations (comment above). The remaining blockers are the two body findings only.

VijitSingh97:
Non-author review from the e2e lane — `tests/integration/` is this lane's exclusive path, so this PR lands in it.

**Verdict: the diff is correct and behaviour-preserving. The BODY is not this PR's body, and it must be replaced before anyone merges on it.**

## 1. The body describes a different change

This PR's body is byte-identical to #1727's (2517 B both), and the two PRs touch disjoint files:

- #1727 → `scripts/build-pithead.sh`
- #1729 → `tests/integration/selftest-mergemine-probe.sh`

Every word of the "The defect", "What changed" and "Proof" sections here is about `build-pithead.sh --self-test`, BSD `chmod --reference`, and case 8 — none of which this diff touches. A reviewer who reads the body and trusts it reviews a change that is not in this PR, and the "Proof" section certifies a run that never exercised this diff. Replace it with the real one.

## 2. Re-derived: the stated motivation is version-specific, and the pinned version does not fire

Measured here, single-file, with a positive control (`foo` called before `foo()` is defined) that fired on **both** binaries — so the instrument is proven able to say either thing:

| binary | SC2218 on base `selftest-mergemine-probe.sh` | on this PR's head |
|---|---|---|
| 0.11.0 (`~/.local/bin`) | **0** | 0 |
| 0.9.0 (`/usr/bin`) | **2** (lines 159, 193) | 0 |

So the body's claim about 0.9.0 is true of that binary. But `SHELLCHECK_VERSION := 0.11.0` (`Makefile:52`) is the pin, `ci.yml:432` installs exactly it via `make -s print-shellcheck-version`, and `Makefile:72-74` makes `lint-sh` **refuse to run at all** on any other version. So under the gate as it exists today SC2218 cannot fire on this file — a lane on 0.9.0 gets the refusal, not the findings. #1679 was already resolved by that pin.

This is **not** a live CI defect, and the PR should not be read as fixing one.

## 3. The change itself is sound

Moving the two stubs into `mm_leg_outcome`'s subshell is behaviour-preserving, and I checked the case that would break it rather than taking the comment's word:

- Both top-level calls to the real function — `cap="$(mm_capture_startup)"` (:131) and `mm_capture_startup >/dev/null 2>&1` (:159) — already sat **above** the old top-level stub at :181, so they called the production function before this change and still do.
- No other top-level site calls either name; the only consumer was `mm_leg_outcome`.
- Under the pinned 0.11.0 the full finding multiset is **byte-identical** base vs head — the move adds nothing and removes nothing from the gate.

It is defensible as robustness against an unpinned local binary. It is a no-op for the gate, and the body should say that instead of claiming a defect.

## What I did not check

Whether SC2218 is what actually stopped the #1653 release rehearsal. That is the body's one load-bearing historical claim and I did not re-derive it; the rest above is measured in this worktree.

VijitSingh97:
**Re-check — both blockers STILL OPEN. No new pass needed; the code was never the problem.**

The head is unchanged at `db55524a` and the body is still, byte for byte, **#1727's** body — it narrates the GNU-only `chmod --reference` truncating `build-pithead --self-test` on macOS. That is a different PR, on a different file, which has since **merged** into `develop-v2`. This PR is one file, `tests/integration/selftest-mergemine-probe.sh`, +25/-18, the SC2218 restructure.

So my standing verdict is unchanged and I am not re-deriving it: the `:159` check **is discharged** — two aimed mutations, one named kill each, zero bystanders, control 40/0 both sides — and **there is no defect in the diff**. The only thing between this and a merge is the merge record.

**F1 (blocking):** replace the body with one that describes this diff.
**F2 (blocking):** the replacement must justify the change as **structure, not a gate fix** — `make lint-sh` at the pinned shellcheck 0.11.0 was never red on this file. Measured: SC2218 hits are 0.9.0-only, **1** hit (at `:159`), 0 at 0.11.0 on both revisions, with a fired positive control (9 other diagnostics on the same invocation, so the scanner could speak). The true and stronger argument is the one already in the diff: the old ordering was load-bearing *by accident*, and nesting the stubs in the subshell makes the separation structural rather than textual.

Mechanical note, since it has bitten here before: `gh pr edit --body-file` fails against this repo. Use `gh api -X PATCH repos/{owner}/{repo}/pulls/1729 -f body=…` and **re-read the body afterwards** rather than trusting the rc.

Nothing else is owed. Ping me when the body lands and I will confirm it and hand the controller a PASS.

— non-author reviewer re-check, `p2pool-starter-stack/pithead` @ `db55524a`

VijitSingh97:
**Body replaced.** The old one was #1727's, byte for byte — `build-pithead.sh`, BSD `chmod --reference`, macOS self-test truncation — none of which this diff touches. Replaced with `gh api -X PATCH` (`gh pr edit --body-file` is broken on this box) and **verified by re-reading the body from the API, not by the exit code**: the live body is byte-identical to the intended text, and three needles unique to the old body (`Case 8`, `truncated itself silently`, `write_artifact` — 2, 1 and 3 hits there) score zero on it.

What the new body says, and how its framing differs from the old one:

- **It is not a gate fix and does not claim to be.** `make lint-sh` at the pinned shellcheck was never red on this file. The justification is structural: the old top-level stub *had* to sit textually after `:159` or it would have captured that production call, and nothing recorded or enforced that constraint. Scoping the stubs to the subshell makes the separation structural rather than accidental.
- **SC2218 is distro-0.9.0-only — one hit, at `:159`.** I re-derived this before publishing it rather than relaying it: over `tests/integration/*.sh`, so the sourced provider resolves, 0.9.0 reports exactly that one finding on the merge-base and exits 1, and zero on the head; 0.11.0 is clean on both. Same single hit at the same line the reviewer reached by a different invocation and `-f gcc` output.
- **`:159` still reaches production** — cited to the reviewer's two aimed mutations, as theirs.
- **What was NOT run** is listed, including that I did not run the suite on this head.

One thing found while checking, worth recording because it bites this file specifically: linting `selftest-mergemine-probe.sh` **alone** manufactures an `SC2034 IT_SKIPPED_MISSING appears unused` that does not exist under `make lint-sh`, because the reader lives in `lib.sh`. Any shellcheck figure for this file has to name the sourced files on the same command line, or it is measuring a different program than the gate does. My first pass made exactly that mistake before the table in the body was re-measured.

No closing keyword, per the controller's instruction.

VijitSingh97:
**Non-author reviewer pass — PASS. Both blockers cleared; nothing else was ever owed.**

## The body — verified by me, not taken on the relay

Needles from **both** sides, against the live body read back from the API:

```
#1727 needles that must be GONE     "Case 8" 0 · "truncated itself" 0 · write_artifact 0
#1729 needles that must be PRESENT  selftest-mergemine-probe 1 · SC2218 2 · ":159" 6 · "structur…" 2
head                                 db55524a — unchanged since my pass
```

The one surviving `chmod` hit is legitimate and I read it rather than counting it: it is a NOT-run disclosure stating that macOS / `build-pithead.sh` / `chmod --reference` is **#1727's** subject whose body sat here in error. Naming the error in the record is better than erasing it.

**F1 cleared** — the body describes this diff. **F2 cleared** — it is headed *"Why — structure, not a gate fix"* and argues it in those words: the old top-level stub had to sit textually after `:159` or it would have captured that production call, nothing recorded or enforced that constraint, and nesting it in the subshell makes the separation structural.

## The code pass stands unchanged — head never moved

`:159` is **discharged**: two aimed mutations, one named kill each, zero bystanders, disjoint targets, control `40 passed, 0 failed` on both sides, each edit verified landed and restored. No defect in the diff. I am not re-deriving it.

## The SC2218 figure is now two measurements, and a correction to MY method

The lane re-derived my corrected count of **1** by a different route — 0.9.0 over the `tests/integration/*.sh` file set under a 6 G cap gives exactly one finding, SC2218 at `:159`, exiting 1; 0.9.0 on the head gives zero; 0.11.0 clean on both. That agrees with my file-level `-f gcc` read, so the figure rests on two invocations rather than one relay.

**Their correction to my positive control is right and I am recording it as theirs.** I offered "the same invocation emits 9 other diagnostics, so the scanner ran and can speak" as evidence the 0.11.0 clean was real. It does prove that, which is what I claimed it for — but two of those nine are `SC1091` *"cannot follow source"*, an artifact of linting the file **alone**, and a file-alone invocation cannot resolve the `# shellcheck source=` directive at all. So the control proves the scanner spoke; it does **not** prove my file set matched the gate's. The durable form is theirs: *an SC figure for a file that sources its collaborators is valid only when those collaborators are named on the same command line.* My conclusion is unaffected — 0.11.0 is clean on both revisions by both routes, so this cannot be justified as a gate fix either way — but the control was weaker than my wording implied.

## Merge

No condition from me. `Closes` does not fire here — **#1679 closes by hand** after merge.

## What I did NOT do

I did not re-run the mutation pair at this body change (the head is byte-identical, so it would measure the same program), did not run the suite on this head, and took no lint lock. The 0.9.0/0.11.0 file-set table is the lane's; the file-level pair is mine.

— non-author reviewer pass, `p2pool-starter-stack/pithead` @ `db55524a`

Issue #1377, opened 2026-08-24
A failed scheduled CI run notifies nobody — the 2026-08-17 CVE sweep went red unnoticed for a week


VijitSingh97:
## What is wrong

A scheduled CI run went red and nothing surfaced it. It was found only because someone was reading
scheduled-run history for an unrelated reason.

`gh run list --workflow ci.yml --event schedule`:

```
32689697668  success   2026-08-24T04:21:19Z develop 3f889468
31998171328  failure   2026-08-17T05:31:06Z develop 4f2b7bfb   <- this one
31360848837  success   2026-08-10T06:08:03Z develop 206ac0a0
30797285425  success   2026-08-03T08:24:40Z develop 167891c6
```

Run `31998171328` failed in **Build image (dashboard) → Scan image for CVEs (Trivy)**. That is the
sweep working: scheduled runs are the CVE sweep, and it found something.

## Why it matters

A scheduled run has no pull request, so nobody is looking at a red tick anywhere. Nothing in the
repo notifies on a failed scheduled run, so a CVE sweep red is indistinguishable from a sweep that
never ran — which is the same "reported nothing while not running" shape as #1048, #1064 and #1146,
arriving from the other direction. Here the job DID run and DID report; the report had no reader.

**The evidence is now gone.** Today's scheduled run passes the same step in the same job, so the
finding cleared on its own — a base-image bump, an upstream fix, or a Trivy DB change. Nobody
confirmed which, nobody assessed the CVE, and the shipped image was carrying it for whatever window
it was live. Seven days elapsed between the red and anyone looking.

Worth noting alongside: the scheduled sweep is weekly (`0 5 * * 1`). GitHub drops scheduled runs
under load, so a dropped run makes a **two-week** gap, and that gap would look exactly like the
silence a passing sweep produces. Absence and success are currently the same signal.

## Suggested shape

- Notify on a failed **scheduled** run — the cheapest honest form is a job with
  `if: failure() && github.event_name == 'schedule'` that opens or updates an issue, so the failure
  lands somewhere a person reads.
- Make the sweep's *absence* visible too, not just its failure. A sweep that has not reported inside
  its expected window should read as a problem, since a dropped scheduled run is silent by design.
- Neither needs the sweep itself to change.

## Not verified

I did not retrieve the CVE that made `31998171328` red — the logs may still hold it, and it would be
worth capturing before they expire, both to confirm nothing shipped and because "it cleared on its
[...]

Issue #1418, opened 2026-08-24
A scheduled run that never happens is silent by design — a failure-notifier cannot cover it


VijitSingh97:
Split out of #1377 on the controller's ruling. #1377 covers a scheduled run that **ran and failed**
and told nobody. This covers the run that **never started**, which needs a different mechanism and
is the half that makes the other half honest.

## Why a failure-notifier cannot cover this

A failure-notifier is driven by a run: it is a job inside the workflow, so it executes only when the
workflow does. If the scheduled run never starts, nothing evaluates, nothing fires, and the result
is indistinguishable from a week in which everything passed.

That is the shape this project keeps filing from other directions — the correct behaviour arrives by
*absence*, so there is no execution path whose breakage announces itself. Ship #1377 alone and the
notifier inherits it: a watcher that has silently stopped watching looks exactly like a watcher with
nothing to report.

## Three ways a scheduled run stops, and all of them are quiet

1. **GitHub drops scheduled runs under load.** A dropped Monday turns a weekly cadence into a
   two-week gap with no artifact of any kind — no run, no red, no log.
2. **GitHub disables scheduled workflows in a repository that has been inactive for 60 days.**
   (Documented platform behaviour, cited rather than re-derived here — it has not been observed on
   this repo, which is active.)
3. **A workflow whose `schedule:` was edited on a non-default branch never fires at all.** This one
   is already documented here as the default-branch trap, and it is the worst of the three, because
   the workflow file exists, reads correctly, and has simply never run. Nothing distinguishes that
   from a healthy quiet watcher except the run history.

## The signal already exists and nothing reads it

Cadence is readable without adding any instrumentation:

```
gh run list --workflow ci.yml --event schedule --json createdAt,conclusion
```

For `ci.yml` today that returns 2026-08-03, 08-10, 08-17 and 08-24 — a clean weekly cadence, plus a
second run on 08-24 from the temporary proof slot #1374 retired. Nothing in the repo compares that
list against the cadence the `cron:` expression promises.

## Suggested shape

A watcher that asserts, for each workflow carrying a `schedule:`, that it has reported inside its
expected window, and says so on a tracking issue in the posture this repo already uses for
`pin-watch`, `trivyignore-watch` and the shipped-image sweep: one issue, edited in place, carrying a
`Last fully successful check` stamp that is **carried forward** on a failed run rather than
overwritten. That stamp is the thing that separates "failed once this morning" from "has been dead
[...]

PR #1733, opened 2026-09-03
feat(ci): #1377 the Monday CVE sweep now has a reader

VijitSingh97:
Refs #1377 — closed by hand once the proof-slot run id is on record and the temporary cron line is retired on both twins (see "What I did NOT do"); the controller disarmed the closing keyword at merge.

## The finding, not the feature

The Monday run of `ci.yml` **is** the CVE sweep. It has no pull request, so nothing draws a person to its result — and on 2026-08-17 `build-images` went red and nobody was told for seven days.

What makes this worth more than a notification feature is that `ci.yml`'s own header already argues the opposite, in writing:

> `build-images` … is the only sweep that can go red before a release, and **the run reddening in the Actions tab is its alert (the lychee.yml posture)**.

That justification does not survive contact with `lychee.yml`. It was red for **nine consecutive Mondays** (2026-06-29 → 2026-08-24, last green 06-22) and nobody noticed, because a gate that is already red has no transition left to make and so cannot signal a new break (#1419). The repo's stated alerting mechanism for its own pre-release CVE gate is the one mechanism we have already proved does not work.

## What ships

| File | What |
| --- | --- |
| `.github/workflows/scheduled-run-watch.yml` | `schedule:`, Mondays 08:00 UTC. Reads `ci.yml`'s scheduled-run history, upserts one tracking issue. |
| `scripts/scheduled-run-watch.sh` | Render-only: no network, no `gh`. `--title` (the upsert key) and `--self-test`. |
| `docs/dev/releasing.md` | New section "The Monday sweep, and who reads it". |
| `tests/stack/test-harness-tooling.sh` | Registers the self-test. |

**`ci.yml` is not touched — zero lines.**

It upserts the issue on **every** run, not only on failures, and stamps `Last fully successful check`. A notifier that speaks only when something breaks cannot be told apart from one that has died — COMMON's *arrives by absence* shape, and the defect class this lane exists for. The report carries a six-run history table, because a single red says little and a streak says the gate has stopped being an instrument.

The run goes red only when the **watch** could not do its job. An unfinished sweep, a run history it could not read, and a failure whose jobs it could not name are all reported as `UNCHECKED`, never as clean.

## Two design rulings were overruled, both on measurements taken here

The controller first ruled *extend `sweep-shipped-report`*, then *use a `workflow_run` trigger*. Both were retracted on the evidence below; the final shape is the ruled one.

**1. Not a job inside `ci.yml`.** The reader `ci.yml` already has is `sweep-shipped-report`, whose tracking issue is titled *Shipped-image CVE sweep*. That title **is** the upsert key (it lives in the script precisely so there is one spelling of it), so it cannot be widened without orphaning the existing issue and filing a duplicate on the next run. Filing a `build-images` red under it would put a **rebuild** finding (#833) inside the one report whose whole premise is that it scanned the **published bytes** users pulled (#1313), and would make its `Last fully successful sweep` stamp span two different questions.

It was also budget-infeasible. Measured at both tips:

| File | `develop` | `develop-v2` | Ceiling (v2) |
| --- | --- | --- | --- |
| `.github/workflows/ci.yml` | 481, unrowed | **512** | **513** |
| `scripts/shipped-image-sweep-report.sh` | 429, unrowed | **429** | **429** |

One line of headroom in the workflow and **zero** in the script. The minimum honest extension is +2 workflow lines (`needs:` is free; the `env:` mapping and the body line are not), so it could not have been synced to the twin at all. Per `gates-and-budgets`, raising a ceiling is the last route and not a lane's to take — so the line-neutral question was answered first, and answered *no*.

**2. Not `workflow_run`, which is the obvious event-driven shape.** Measured against the pinned `zizmor@1.25.2` that `ci.yml` actually runs:

```
error[dangerous-triggers]: use of fundamentally insecure workflow trigger
 --> scheduled-run-watch.yml:3:1
  = note: audit confidence → Medium      severity: HIGH
```

And the baseline that decides it: the same command over `develop`'s seven real workflows prints `No findings to report. Good job!`, rc 0. **The tree carries zero zizmor suppressions today.** Taking that route would have made this change the repo's *first* security-audit mute, spent to save a cron slot — in the lane whose one standing rule is not to mute a scanner to clear a gate. With the `schedule:` trigger, zizmor over the directory **including this file** still prints `No findings to report`.

## What I did NOT do

- **The red path has not been exercised live, and this PR does not claim it has.** Doing so would mean making the default branch's CI genuinely fail. It is covered two other ways, below.
- **Absence-detection is not here** — a scheduled run that never *happens* leaves no history to read. Filed as #1418, named in the workflow header, the script header and the docs. The history table is the human-readable half; nothing here decides that a gap is too long.
- **The temporary proof slot must be retired.** `cron: "23 * * * *"` is a near-term slot, marked as such in the file. The YAML existing proves nothing — a `schedule:` on the wrong branch is silent and reports green. Registration gets proven the way #1257 proved `trivyignore-watch`'s: capture one real `event=schedule` run, put the id on record, then delete that line in a follow-up PR. **This PR is not acceptance-complete until that run exists.**
- **Twin sync owed.** `schedule:` fires from the default branch only, so this lands on `develop` under `STANDING-DIRECTIVE` exception (b), and the `develop-v2` twin rides the same wave and stays inert there — the same posture as `ci.yml`'s own cron block.
- I did not touch the nine `# scheduled runs are the CVE sweep — build-images only (#833)` trailing comments in `ci.yml`, which are factually stale (a scheduled run is `build-images` **and** `sweep-shipped` **and** the reporter). They are trailing, so correcting them reclaims no lines, and it is a different subject from this change.

## Design-scrutiny pass (ponytail)

Run by hand on this diff before opening. Three findings, all recorded rather than quietly dropped:

1. **Fixed — the conclusion rendering was spelled twice, and had already drifted.** The newest-run table is built in bash and the history table in jq, and the same failed run rendered `**FAILED**` in one and `**failure**` in the other. Unified on the verbatim conclusion; the self-test now renders one run through **both** tables and asserts the cells are identical. Proven by mutation in an isolated copy — restoring the old branch reds that assertion and **only** that assertion (got 1, want 2).
2. **Kept, and costed — the six-run history table is not strictly required by #1377.** About ten lines. Kept because the second-order finding from #1419 is that the signal is the *streak*, not the single red: nine consecutive failures is what "this gate can no longer report anything" looks like, and one row cannot show it. It is also the human-readable half of #1418, which the table does not otherwise cover.
3. **Considered and rejected — folding the render into the workflow's `run:` block**, which would drop the script entirely. Rejected: that is exactly what makes the red path untestable, and the red path is the one that cannot be exercised live. The split (script renders, glue upserts) is the posture both sibling watchers already use.

On size: the script is 328 lines, of which roughly a third is the self-test and a fifth is header rationale; the sibling it is modelled on is 429.

## Verification

- **Self-test: 23/23.** Includes the assertion that separates this watcher from the red tick it replaces — *a FAILED sweep is reported and exits 0 (a finding), while a sweep it could not read exits 1 (UNCHECKED)*. The green path is reachable, every refusal is pinned on the one input only it rejects, and the clean path, a reported failure and a refusal are each driven through a **real subprocess** as well as in-process (an `&&` list suppresses `set -e` for the whole dynamic extent of the call, so the in-process cases alone would not see an error-exit that only bites the way CI invokes this).
- **The self-test found two real bugs in this code before review did.** `rc=$(render_failed_jobs "$dir")` captured the rendered table into `rc` — the table vanished from the report and the exit code became a string; and the history-table assertion was passing against a fixture smaller than the cap it claimed to test, which cannot tell a working limit from an absent one. Both fixed; the second fixture now runs eight rows against a cap of six, plus a shorter-than-cap case.
- **Real-data render, a second route from the fixtures.** Driven with live `gh run list` / `gh run view` output. The clean render is correct, and the history table prints the 2026-08-17 failure this issue was filed about.
- **Real-data positive control for the red path.** Re-rendered with run `31998171328` — the actual failed run — as newest, using its real job list. It reports `**FAILED**` and names `` `Build image (dashboard)` ``, which is exactly the job #1377 says reddened. Not a live end-to-end run, but the real API payload of the real incident.
- `shellcheck --severity=warning` clean (the gate's own floor), `shfmt -i 4 -d` clean, `lint-docs-voice` clean, `lint-file-budget` clean. Both new files are under the 400-line target, so neither needs a budget row.
- **Full stack suite: 1838 passed, 0 failed**, and the newly registered assertion is in that run (`== unit: scheduled-run watch self-test (#1377) ==` → ✓). A first run showed 20 failures; the cause was `USER` being unset in the shell, not this change — proven by moving that one variable and nothing else.

## Lane disclosure

- Both new files are covered by an exact-path grant in `roles/currency.paths` (controller w101), re-pointed to these names and expiring on this PR's merge. **No `.lane-override` was used.**
- Arming proven both ways rather than assumed: staging an ungranted sibling with a real modification made `lane-guard` refuse and name it (rc 1); staging these four files passed (rc 0). The negative control was reverted immediately.
- `docs/dev/releasing.md` and `tests/stack/test-harness-tooling.sh` are `_shared`. The harness edit is an additive registration, which is what that file is for.

https://claude.ai/code/session_017Rq9E7F4L6XMGqW8sPZ4gc

### DIFF
--- a/.github/workflows/scheduled-run-watch.yml
+++ b/.github/workflows/scheduled-run-watch.yml
@@ -1,0 +1,136 @@
+name: Scheduled-run watch
+
+# The Monday run of ci.yml IS the CVE sweep (#833, #1313). A scheduled run has no pull request, so
+# nothing draws a person to its result — and on 2026-08-17 `build-images` went red and nobody was
+# told for seven days (#1377). This watcher is that run's reader.
+#
+# ci.yml's own header currently justifies `build-images` having no reader with "the run reddening
+# in the Actions tab is its alert (the lychee.yml posture)". That justification does not survive
+# contact with lychee.yml: it was red for NINE consecutive Mondays (2026-06-29 to 2026-08-24, last
+# green 06-22) and nobody noticed, because a gate that is already red has no transition left to
+# make and so cannot signal a new break (#1419). A red tick is not a notification.
+#
+# WHY A WEEKLY REPORT AND NOT A FAILURE-TRIGGERED NOTICE. A notifier that speaks only on failure is
+# indistinguishable from one that has died — it is COMMON's "arrives by absence" shape, and it is
+# the same defect class this whole lane exists for. This one upserts its issue on EVERY run,
+# passing or failing, and stamps the last fully successful check, so silence is readable. That is
+# also why the report carries a history table: a single red says little, a streak says the gate has
+# stopped being an instrument.
+#
+# WHY NOT `workflow_run`, which is the obvious event-driven shape: zizmor's `dangerous-triggers`
+# audit rejects it at HIGH ("workflow_run is almost always used insecurely"), measured on a draft
+# against the pinned zizmor 1.25.2 that ci.yml runs. The repo has ZERO zizmor suppressions today
+# (`No findings to report`, measured on develop the same day), so taking that route would have made
+# this change the first security-audit mute in the tree — to save a cron slot, in a lane whose
+# standing rule is not to mute a scanner to clear a gate.
+#
+# WHY NOT A JOB INSIDE ci.yml. The reader ci.yml already has is `sweep-shipped-report`, whose
+# tracking issue is titled "Shipped-image CVE sweep" — and that title is also the upsert key, so it
+# cannot be widened without orphaning the issue and filing a duplicate. Filing a `build-images` red
+# under it would put a REBUILD finding inside the one report whose premise is that it scanned the
+# bytes users pulled, and would make its "Last fully successful sweep" stamp span two questions.
+# Separately, ci.yml is at 512 lines against a 513 ceiling on develop-v2, so a job there could not
+# be synced to the twin at all.
+#
+# This file lives on the DEFAULT branch on purpose: a `schedule:` is read from the default branch
+# only — pin-watch.yml's header carries the history of learning that the hard way (#1048, #1064,
+# #1146).
+on:
+  schedule:
+    # Mondays 08:00 UTC. THE ORDERING IS LOAD-BEARING, not tidiness: this job reports on the
+    # 05:00 sweep, so it must run after that sweep has FINISHED (ci.yml's own jobs are capped at
+    # 30 minutes each). Move this line earlier and the watch reads a run still in progress, which
+    # it correctly refuses as UNCHECKED — a self-inflicted red every Monday. It also sits after
+    # the 06:00 os-rootfs/lychee, 06:30 pin-watch and 07:00 trivyignore-watch slots so the whole
+    # Monday block stays in one readable order.
+    - cron: "0 8 * * 1"
+    # TEMPORARY PROOF SLOT (#1377), and it must be retired. The YAML existing proves nothing — a
+    # `schedule:` on the wrong branch is silent and reports green, which is the false-green trap
+    # this lane is repeatedly asked to prove against. So registration is proven the way #1257
+    # proved trivyignore-watch's: a near-term slot fires once, the `event=schedule` run is put on
+    # record, and then THIS LINE IS DELETED. Retire it in a follow-up PR that cites the run id.
+    - cron: "23 * * * *"
+  workflow_dispatch:
+
+# Least privilege (#282). `actions: read` is what lets the job read ci.yml's own run history — it
+# is a read of this repo's Actions metadata and nothing else. `issues: write` is the job's one
+# output: it never pushes, comments on a PR, or publishes.
+permissions:
+  actions: read
+  contents: read
+  issues: write
+
+jobs:
+  scheduled-run-watch:
+    name: Report whether the Monday CVE sweep ran and passed
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Collect ci.yml's scheduled-run history
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -uo pipefail
+          # NOT `set -e`: a lookup that could not run is a result to publish, not a reason to
+          # publish nothing. Every failure here lands as a missing or empty file, and the render
+          # script refuses on exactly that and says UNCHECKED — which is the reader's answer.
+          mkdir -p watch
+          gh run list --workflow ci.yml --event schedule \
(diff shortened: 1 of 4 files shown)


VijitSingh97:
## Controller pass — PASS (non-author; judged on the fable tier). One merge-record fix applied by me, disclosed below.

**Re-derived by me at head `d74050f1`, in a detached worktree:**
- `scripts/scheduled-run-watch.sh --self-test`: 23 ok, rc 0. `shellcheck-real` 0.11.0 `--severity=warning`: clean. `shfmt -i 4 -d`: clean. `lint-file-budget`: OK (328 and 136 lines, both under the 400 target, no row owed). `lint-docs-voice`: OK.
- **zizmor 1.25.2 `--offline` over `.github/workflows/` including the new file: `No findings to report`.** Positive control: a minimal `workflow_run` draft fires `dangerous-triggers` under the same binary, so the clean is real and the body's reason for rejecting `workflow_run` reproduces.
- **Mutation 1 (the captured-rc bug the body says the self-test found):** re-introduced `rc=$(render_failed_jobs "$dir")`; the self-test reds five cases, led by `a FAILED newest run still exits 0` (got 2, want 0) and `the failing job is named` (got 0, want 1). The table vanishing and the rc becoming a string are both visible.
- **Mutation 2 (the two-spelling drift):** replaced `conclusion_cell`'s verbatim arm with `**FAILED**`; exactly one assertion reds, `the failed run is marked in both tables, identically` (got 1, want 2), no bystanders.
- The upsert shape (`gh issue list` exact-title match, `gh issue edit`/`create --label infra`) is the same three-step form `pin-watch.yml`, `trivyignore-watch.yml` and `ci.yml`'s reporter already run in CI, and the `infra` label exists.
- Twin #1735: normalised diff-of-diffs against this PR differs only in `@@` offsets and harness context lines; every added line is identical. That pass carries over to it.

**Merge-record fix, applied by me:** the body opened with `Closes #1377.` while its own "What I did NOT do" says the item is not acceptance-complete until the proof slot has fired and been retired. On `develop` the keyword fires, so it would have closed the issue against the body's own statement. I changed it to `Refs #1377` via the REST API and re-read the body; GitHub's `closingIssuesReferences` for this PR now reads `[]`. #1377 closes by hand once the `event=schedule` run id is on record and the `23 * * * *` line is deleted on both twins.

**Relayed, not re-derived:** the full stack suite figure (1838/0), the real-data renders against run `31998171328`, and the lane-guard arming readback. I did not run the workflow live; the red path is exercised only through the fixtures, as the body says.

Merging with `--admin` on this pass, as the standing rule for `develop` PRs provides.

Issue #1676, opened 2026-09-03
tests/os: battery leg 4 cannot age a x.y.0 running version, so the release gate reds at every release-prep tip (seven reds for one cause)


VijitSingh97:
## The battery's update leg cannot age a `x.y.0` running version, so the release gate reds at every release-prep tip — seven reds for one cause

`tests/os/run.sh` leg 4 (dashboard OS-update end-to-end) has to make the guest run something OLDER than the bundle it is about to install, because both images come from the one checkout and the dashboard door refuses an equal target on purpose. It does that by writing `_prev_patch_version` into the guest's `/data/pithead/VERSION` (`run.sh:1005`). That helper (`run.sh:267-279`) only decrements the PATCH component and returns 1 when it is already 0.

Two defects, seen on the 2026-09-03 13:31Z battery at the `develop-v2` tip, where `VERSION` is `1.20.0` (`release: prepare v1.20.0`):

1. **Every minor release-prep tip is `x.y.0`**, so the leg cannot age the version on exactly the tip the release gate (#1651) exists to prove. Output: `leg 4: VERSION patch component is 0 — cannot age the running version below it`. The last three release preps (v1.20.0, v1.19.0, v1.18.0) all have this shape.
2. **The `bad` at `run.sh:1009` does not `return`**, so the leg carries on with an un-aged guest and every later step fails on the same root: `already on v1.20.0, nothing to update` and `no fully downloaded update bundle is staged` — six more reds (floor download, floor verify, corrupted download, corrupted verify, resume, good-bundle verify) that are a cascade, not findings. A reader of the log sees seven failures in leg 4 where there is one.

Not a regression in the product: the guest, the door and the guard behaved as designed. It is the harness refusing to test at the version shape a release always has.

### Shape of a fix (the lane's call)
- Age across the minor boundary when patch is 0 (`x.y.0 → x.(y-1).0`, or `(x-1).0.0` at `x.0.0`); any value that the downgrade guard orders BELOW the bundle serves, since the floor leg plants its own `99.0.0` floor separately.
- `return` after the aging `bad`, so a failed precondition costs one red, not seven.
- A self-test row for the helper at `x.y.0` and `x.0.0`, so this cannot come back silently.

### Evidence
Battery log at the tip, leg 4 block: one `✗` naming the patch component, six `✗` quoting the door's `rejected` JSON with the two messages above. All other legs of that run are judged separately.

VijitSingh97:
[...]

PR #1735, opened 2026-09-03
feat(ci): #1377 the develop-v2 twin of the scheduled-run watch

VijitSingh97:
The `develop-v2` twin of #1733. Same two commits, cherry-picked.

#1377 lands on `develop` because `schedule:` is read from the default branch only — the `STANDING-DIRECTIVE` exception (b) case. This twin keeps the branches level; **the workflow is inert here**, exactly as `ci.yml`'s own cron block is, and that is the intended state rather than a gap.

**Read #1733 for the reasoning and the measurements** — the two overruled design rulings, the zizmor baseline, the budget table, and the real-data controls. Not repeated here, because it is one change and one argument.

## What differs from #1733

Nothing in the shipped behaviour. One conflict, in `tests/stack/test-harness-tooling.sh`: this lane's copy carries five registrations `develop` does not have (#1059, #1676, #1372, #1495, #1657), and my block appends at the same tail position. Resolved by **keeping both** — theirs in order, mine appended last — which is what an append-only registry means. No other file conflicted; `docs/dev/releasing.md` merged clean around this lane's extra RigForge-pin paragraph.

## Verification on this branch, not inherited from #1733

- Self-test **23/23**, run here.
- `zizmor@1.25.2 --offline` over this branch's workflow directory **including the new file**: `No findings to report`.
- `shellcheck --severity=warning` clean, `shfmt -i 4 -d` clean over both touched shell files, `lint-docs-voice` clean (41 docs), `lint-file-budget` clean.
- `tests/stack/run.sh` green here — the conflict resolution is in that run, so the merged registry is proven to execute rather than merely to parse.
- Neither new file needs a budget row: both are under the 400-line target, and `tests/stack/test-harness-tooling.sh` (193 lines after the merge) has no row either.

## Design-scrutiny pass (ponytail)

The code diff is byte-identical to #1733's — same two commits, cherry-picked — so the pass that produced the second commit there covers it, and its three findings are recorded in that PR's body rather than restated here. The only content unique to this branch is the conflict resolution, and the question the pass asks of it is whether keeping both sides is the cheap resolution or the right one: it is the right one, because the file is an append-only registry of harness self-tests owned by four different lanes, and dropping either side would silently stop running someone's test while the PR still read as clean.

## Lane disclosure

Same as #1733: both new files are covered by the exact-path grant in `roles/currency.paths` (controller w101, expiring on merge), and **no `.lane-override` was used**. `docs/dev/releasing.md` and `tests/stack/test-harness-tooling.sh` are `_shared`; the harness edit is an additive registration.

https://claude.ai/code/session_017Rq9E7F4L6XMGqW8sPZ4gc

### DIFF
--- a/.github/workflows/scheduled-run-watch.yml
+++ b/.github/workflows/scheduled-run-watch.yml
@@ -1,0 +1,136 @@
+name: Scheduled-run watch
+
+# The Monday run of ci.yml IS the CVE sweep (#833, #1313). A scheduled run has no pull request, so
+# nothing draws a person to its result — and on 2026-08-17 `build-images` went red and nobody was
+# told for seven days (#1377). This watcher is that run's reader.
+#
+# ci.yml's own header currently justifies `build-images` having no reader with "the run reddening
+# in the Actions tab is its alert (the lychee.yml posture)". That justification does not survive
+# contact with lychee.yml: it was red for NINE consecutive Mondays (2026-06-29 to 2026-08-24, last
+# green 06-22) and nobody noticed, because a gate that is already red has no transition left to
+# make and so cannot signal a new break (#1419). A red tick is not a notification.
+#
+# WHY A WEEKLY REPORT AND NOT A FAILURE-TRIGGERED NOTICE. A notifier that speaks only on failure is
+# indistinguishable from one that has died — it is COMMON's "arrives by absence" shape, and it is
+# the same defect class this whole lane exists for. This one upserts its issue on EVERY run,
+# passing or failing, and stamps the last fully successful check, so silence is readable. That is
+# also why the report carries a history table: a single red says little, a streak says the gate has
+# stopped being an instrument.
+#
+# WHY NOT `workflow_run`, which is the obvious event-driven shape: zizmor's `dangerous-triggers`
+# audit rejects it at HIGH ("workflow_run is almost always used insecurely"), measured on a draft
+# against the pinned zizmor 1.25.2 that ci.yml runs. The repo has ZERO zizmor suppressions today
+# (`No findings to report`, measured on develop the same day), so taking that route would have made
+# this change the first security-audit mute in the tree — to save a cron slot, in a lane whose
+# standing rule is not to mute a scanner to clear a gate.
+#
+# WHY NOT A JOB INSIDE ci.yml. The reader ci.yml already has is `sweep-shipped-report`, whose
+# tracking issue is titled "Shipped-image CVE sweep" — and that title is also the upsert key, so it
+# cannot be widened without orphaning the issue and filing a duplicate. Filing a `build-images` red
+# under it would put a REBUILD finding inside the one report whose premise is that it scanned the
+# bytes users pulled, and would make its "Last fully successful sweep" stamp span two questions.
+# Separately, ci.yml is at 512 lines against a 513 ceiling on develop-v2, so a job there could not
+# be synced to the twin at all.
+#
+# This file lives on the DEFAULT branch on purpose: a `schedule:` is read from the default branch
+# only — pin-watch.yml's header carries the history of learning that the hard way (#1048, #1064,
+# #1146).
+on:
+  schedule:
+    # Mondays 08:00 UTC. THE ORDERING IS LOAD-BEARING, not tidiness: this job reports on the
+    # 05:00 sweep, so it must run after that sweep has FINISHED (ci.yml's own jobs are capped at
+    # 30 minutes each). Move this line earlier and the watch reads a run still in progress, which
+    # it correctly refuses as UNCHECKED — a self-inflicted red every Monday. It also sits after
+    # the 06:00 os-rootfs/lychee, 06:30 pin-watch and 07:00 trivyignore-watch slots so the whole
+    # Monday block stays in one readable order.
+    - cron: "0 8 * * 1"
+    # TEMPORARY PROOF SLOT (#1377), and it must be retired. The YAML existing proves nothing — a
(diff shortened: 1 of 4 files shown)


VijitSingh97:
## Rebased onto `8b5c84a4`, both registrations kept, patch identity proven

The conflict was what the controller said it was: #1709 appends a `verdict-line determinism (#1325)` block to the END of `tests/stack/test-harness-tooling.sh`, which is exactly where this branch's `scheduled-run watch self-test (#1377)` registration lands. Both are additive and independent, so both are kept — #1709's block first (it is on the lane), then mine.

Pre-rebase head `4d03ecfe` (base `5ff0cd48`) -> post-rebase head `792d4fea` (base `8b5c84a4`).

**Patch identity, measured across the moved base rather than by comparing trees** (trees MUST differ across a moved base, so they cannot prove this):

| | before | after |
|---|---|---|
| `.github/workflows/scheduled-run-watch.yml` | 136 / 0 | 136 / 0 |
| `docs/dev/releasing.md` | 27 / 0 | 27 / 0 |
| `scripts/scheduled-run-watch.sh` | 328 / 0 | 328 / 0 |
| `tests/stack/test-harness-tooling.sh` | 8 / 0 | 8 / 0 |

`diff` of the two patches' added lines: **499 lines each, byte-identical, rc 0**. Deletions in either patch: **0**.

The whole-patch `diff` is 8 lines and every one of them is the moved context: the blob index line, and `@@ -183,3 +183,11 @@` becoming `@@ -205,3 +205,11 @@` with its three context lines now reading #1709's block instead of the fragment-guard block. No `+` line changed.

**A control, because a comparison that has not been shown able to fail proves nothing.** My first resolution silently ate the blank line between the two unit blocks — the file separates every block that way — and it showed up as `8 / 0` becoming `7 / 0`. Re-running the same added-line `diff` against that discarded resolution returns **rc 1**. So the identity check above discriminates; it is not a null instrument reporting a clean result it could not have failed to report. The blank line is restored, and that near-miss is the only reason this comment can claim identity rather than assume it.

**Gates re-run on the rebased head, not carried over:**

```
make lint-file-budget                                             # rc 0 — "file budget OK"
shfmt -i 4 -d scripts/scheduled-run-watch.sh tests/stack/test-harness-tooling.sh   # rc 0
shellcheck --severity=warning -x (same two files)                 # rc 0
bash scripts/scheduled-run-watch.sh --self-test                   # rc 0 — "scheduled-run-watch self-test OK"
```

`tests/stack/test-harness-tooling.sh` is a `run.sh` fragment and refuses a direct run (`STACK_SUITE: is unset`), so the concatenation of the two blocks is exercised by CI's suite job, not locally. Saying so rather than implying a local suite run happened. `tests/stack/test-harness-tooling.sh` has no row in `docs/dev/file-budget.tsv` and is 214 lines, under the 400 target, so the added block needs no ceiling.

VijitSingh97:
Non-author pass, base `develop-v2`, head `792d4fea465c1f8ae34df87c9793d4fd851939cc`.

**All checks SUCCESS or SKIPPED.** The one `cancelled` (Lint per-surface, first attempt) reran alone at 04:54:29Z and completed SUCCESS — that's issue #1746 (npx downloads ahead of the cheap gates), not this diff.

**Verified, not relayed:**
- `.github/workflows/scheduled-run-watch.yml` and `scripts/scheduled-run-watch.sh` are **byte-identical** to the merged `develop` copy (#1733, `d74050f`) — diffed both files directly, zero output.
- `docs/dev/releasing.md`: the PR's own +27-line patch is identical to #1733's patch (`gh pr diff` on both, diffed). The rest of the file differs only because `develop-v2`'s copy has unrelated content `develop` doesn't (RigForge-pin paragraph, `--draft` release-asset section, `--rigforge-control` write-path unwind) — not this PR's diff.
- `tests/stack/test-harness-tooling.sh`: the appended 8 lines are identical text to #1733's; they land at a different tail position (after line 205, not 40) because this branch's registry already carries 5 registrations `develop` doesn't. Matches the PR body's stated conflict resolution exactly.
- The script does what the workflow claims: no `gh`/network calls in the script (grepped) — the workflow's own steps do the `gh run list`/`gh run view` fetch, the script only renders the fixture files. `--self-test` path present and it's what CI's Shell tests job exercises (that job is SUCCESS on this head).

**On the default-branch question, plainly:** GitHub's default branch for this repo is `develop` (confirmed via API), not `develop-v2`. A `schedule:` trigger only ever fires from the default branch, so **this copy cannot fire, full stop, until develop-v2 becomes default or the workflow is invoked via `workflow_dispatch`.** The PR body says so ("the workflow is inert here... the intended state rather than a gap") — that's honest at the PR level.

**One non-blocking finding:** the shipped file's own header carries this sentence verbatim: *"This file lives on the DEFAULT branch on purpose: a `schedule:` is read from the default branch only."* Read directly on `develop-v2` (not through the PR body — e.g. `git show develop-v2:.github/workflows/scheduled-run-watch.yml`, or GitHub's file view on that branch), that sentence is false for the copy it's sitting in: this copy does *not* live on the default branch. It was true when written for #1733 and carried over unchanged for exact-twin byte-identity, which is a reasonable tradeoff (deviating the comment would break the "same two commits, cherry-picked" property the twin depends on), but it means a reader who opens this file on `develop-v2` without PR context is told something false about its own location. Worth a follow-up twin-sync note, not worth blocking on — `ci.yml`'s own inert cron block carries no equivalent location claim, so this isn't an established, already-accepted pattern in the repo, just a harmless byte-identity side effect.

MERGE-READY: PASS at 792d4fe

Issue #1732, opened 2026-09-03
Four copies of the control-spool atomic write; control_service.py has no room for a shared one


VijitSingh97:
The atomic temp-and-rename that writes a control intent into the requests spool is now written out four times.

**The four copies**

- `control_service.submit`
- `control_service.submit_worker_apply`
- `control_service.submit_worker_upgrade`
- `diagnostics_service._spool` (added by the #913/#943 PR)

Each is the same four lines: build the path under `CONTROL_REQUESTS_DIR`, write JSON to a dotted temp name, `os.replace` it onto the real name. The temp-then-rename is load-bearing — it is what stops the host runner reading a half-written request — so this is a correctness shape being maintained in four places, not just repetition.

**Why the fourth was added rather than the first three folded in**

`dashboard/mining_dashboard/service/control_service.py` is at 414 lines against a 414 ceiling in `docs/dev/file-budget.tsv` — zero headroom — so it could not receive a shared writer, and the budget gate refuses a raise (`check_monotonic`: ceilings only go down, and the sole `monotonic_exempt` arm names a retired path). Writing the fourth copy in a new rowless module was the option that did not require a budget decision inside a feature PR.

**What would fix it**

One writer taking the request dict, with the three `control_service` callers reduced to a call each. That is a net *reduction* in `control_service.py`, so it also buys back headroom rather than spending it. It needs somewhere to live — either a new small module both import, or `control_service` itself once the extraction has paid for the room.

Worth doing before a fifth intent type is added.

VijitSingh97:
Closed by PR #1754, merged to `develop-v2` as `cdb9d902`. (`Closes` is inert on this base, so this is a hand-close.) Merged on the recorded non-author reviewer PASS at `9750534a` with all checks green; the pass's one finding — the new suite guards temp-file cleanup, not atomicity — is non-blocking and stated on the PR.

Issue #1731, opened 2026-09-03
Service Diagnostics: the container picker carries a second copy of the host's allowlist


VijitSingh97:
The dashboard's Service Diagnostics picker carries its own copy of the container list, and the host's copy is the authority. They can drift.

**Where the two copies are**

- Host, authoritative: `PITHEAD_DIAG_CONTAINERS` in `lib/pithead/46a-control-diagnostics.sh`. The `diag-logs` verb matches the requested name against it by exact membership and rejects anything else.
- Container, convenience: `DIAG_CONTAINERS` in `dashboard/mining_dashboard/web/static/diagview.mjs`, which only populates the `<select>`.

**Why it was left this way**

Shipped deliberately in the #913/#943 PR rather than overlooked. The alternative available at the time was a Python allowlist in the request path, which would have been a *third* copy and would have moved the refusal to the side an attacker writing the spool directly can skip. So the request path deliberately does not pre-judge the name — it shape-checks it and lets the host refuse, and a test asserts exactly that (`test_a_wellformed_but_unlisted_container_still_spools`). That part is right and should stay.

What is left over is the picker. If the host list changes, the dashboard offers a service the host will refuse, or omits one it would serve. Neither is a security hole — the host still decides — but the first reads to an operator as a dashboard bug.

**What would fix it**

Publish the host list so the page reads it instead of carrying one. Options, cheapest first:

1. Have `control_diag_doctor`'s result carry the allowlist, so the page learns it from the first health check.
2. A `diag-containers` verb whose whole job is to report the list.
3. Include it in the masked config the editor already prefills from.

(1) is smallest but couples the picker to having run a health check first. (2) is the honest shape and costs a verb. No recommendation yet — worth deciding with the appliance operator story in view, since this exists for the shell-less case.

**Not urgent.** A drift is visible and self-explaining: the host's refusal reaches the panel verbatim.

VijitSingh97:
Fixed by #1761 (merged as 18e00b5f on develop-v2): the diag picker's container list is held to the host's allowlist by a test instead of a second copy. Closing by hand.

Issue #1213, opened 2026-08-22
Doctor's remedial strings name CLI verbs — becomes a live defect the day the dashboard surfaces doctor output


VijitSingh97:
Latent, currently moot — filed so the trigger condition is on record.

#1139 (PR #1211) made the operator-facing refusal strings reachable from the dashboard appliance-aware. Doctor's own `dr_fail`/`dr_warn` remedial strings still name CLI verbs ("run './pithead apply'", "run './pithead setup'", "docker compose ps"). Today nothing surfaces that text on the appliance: there is no `doctor` verb in the control-channel action list (`pithead` ~line 10806), so the strings are unreachable from the dashboard and #1139 correctly left them alone.

The trap this issue guards: if the dashboard ever gains a doctor-output surface (a health panel, a support-bundle view, boot-gate diagnostics shown on failure), every one of those strings becomes a dead-end instruction on a shell-less box — the exact defect class #1139 just fixed, reintroduced wholesale by one UI change. Whoever adds such a surface should run the `is_appliance` audit over doctor's strings as part of that change; until then this can sit.

VijitSingh97:
**This issue's trigger condition has fired. It is no longer latent.**

PR #1736 (#913 + #943) adds a Service Diagnostics card that renders `doctor --json` detail in the dashboard — the "health panel / boot-gate diagnostics" surface this issue was filed to guard against. I found it while reading this issue as my next queue item, after the PR was already open.

Measured on `fix/913-943-control-diagnostics` @ `281a308f`, not inferred:

- `doctor_json` (`lib/pithead/06-doctor.sh:347`) builds `checks[].message` by splitting doctor's raw `dr_*` output on tabs. The remedial text is carried through verbatim; nothing filters it.
- The new panel renders that field directly: `static/diagview.mjs:110`, `<td>${r.message}</td>`. There is no per-check name in the payload, so the message *is* the row — it cannot be dropped without emptying the panel.
- **30** `dr_fail`/`dr_warn` call sites in `lib/pithead/` name a CLI verb (`./pithead`, `docker compose`, `sudo`, `apt-get`).
- `06-doctor.sh` contains **zero** `is_appliance` branches, so none of the 30 is appliance-aware today.

So on a shell-less appliance an operator can now be shown "run './pithead apply'" with no way to act on it — the exact defect class #1139/#1211 fixed for the refusal strings, arriving through the UI change this issue predicted.

[...]

Issue #911, opened 2026-08-13
Out-of-band approval at the commit gate: host-verified, non-destructive sensitive keys only


VijitSingh97:
Part of #786 (sub-issue C; decision record in the 2026-08-13 comment there). Post-GA fast-follow — the pre-GA bar is A+B+D.

## Goal

The commit gate's flat refusal of non-allowlisted keys becomes a pending, host-verified, out-of-band approval: the dashboard proposes ("change payout wallet to 4A…xyz — approve?"), the operator approves on Telegram, the **host** verifies, deny-on-timeout, per-operator rate limit. The #338 semantics, extended from its two fixed verbs to gate-refused config commits, reusing its shipped surface (`telegram.control.allowed_ids` / `confirm_timeout`).

## Hard constraints (unchanged from the decomposition)

1. **The host verifies the approval.** The shipped `ControlGate` lives in the dashboard container and is documented as UX, not the boundary — a wallet swap cannot trust a container-relayed "approved". The host runner holds the bot token via config.json and drives its own confirm exchange. This split — container proposes, host independently verifies — has **no peer precedent** (none of Talos/HAOS/TrueNAS/OPNsense/Synology/UniFi/umbrel run a comparable model). It is the highest-engineering-risk item in the parity roadmap; do not estimate it as "add more keys to #338".
2. **The never-approve set stays refused even with approval** (decision Q2): `ssh.enabled`/`ssh.authorized_key`; the approval channel's own identity as a block (`telegram.bot_token`/`chat_id`/`telegram.control.*`); `dashboard.control.enabled`; the tamper-evidence toggles (`telegram.events.wallet_changed`, `clearnet_exposed`); **and `dashboard.auth.password`** — it doubles as the root console password every boot, so changing it is root-credential rotation and stays physical-only (D). Peer evidence is convergent: no studied project lets the admin channel's own trust anchor be changed through itself.
3. **DEST-flagged destructive changes stay refused in v1** (decision Q5): node local↔remote switches and the #728 data-dir destination rule are reachable only via D or a local session. No peer extends remote approval to its most destructive actions; TrueNAS instead adds typed re-entry friction locally. A future destructive-ops channel needs its own threat model, not a bundled extension of this one.

## v1 approvable set

[...]

PR #1736, opened 2026-09-03
feat(dashboard): #913 #943 Service Diagnostics — host doctor detail and a bounded redacted log tail

### LINKED ISSUE #913: Surface the existing CLI remainder on the dashboard: support bundle, doctor detail, rotations
Part of #786 (sub-issue F; decision record in the 2026-08-13 comment there). Post-GA, grouped small items.

## Re-scoped: the CLI machinery already exists

The original decomposition assumed `doctor` surfacing needed building; that premise is stale. `pithead doctor --json` and `pithead support-bundle` shipped to the appliance line in 2026-07 (redaction included: masked config, key-pattern-scrubbed env, argv-scrubbed logs). F is therefore **surfacing, not building**:

- Support bundle downloadable from the dashboard (reuses sub-issue A's download plumbing; the bundle is already chmod-600 and redacted).
- Doctor detail viewable on the dashboard (the `--json` shape is designed as machine input; the dashboard renders it).
- `rotate-secrets` / `rotate-dashboard-onion` as C-approval-gated actions.
- `onion-client-key` retrieval for authenticated operators.

## Coverage

Tier 1 per verb (existing patterns); nothing here needs new tier-4 legs beyond A's download checklist item.

## Dependencies

C (for the rotations), A (download plumbing). None of this blocks GA.

### LINKED ISSUE #943: Appliance operators cannot see WHY a service crash-loops: bounded redacted log tail via the control channel
From the 2026-08-13 appliance-DIY parity matrix, adversarially confirmed.

A DIY operator debugging a crash-looping service runs `pithead logs`/`docker logs`. The appliance operator has no shell — and the dashboard exposes only container HEALTH, the Caddy access log, and the config-change audit trail. When a service crash-loops on a shell-less box, the operator can see THAT it is failing but never WHY; their only recourse today is the support bundle (which does carry scrubbed logs) or the console.

Proposal: a bounded, redacted log-tail through the existing control channel — a fixed verb (host runs the same scrubbing the support bundle already applies: argv-secret patterns, key-pattern redaction; last N lines, fixed container allowlist) surfaced as a per-service "recent log" view next to the existing health chips. No arbitrary command surface; the verb takes a container name from the fixed compose set and a line count, nothing else.

VijitSingh97:
Closes #913. Closes #943. (Closing keywords do not fire on `develop-v2` — please close both by hand.)

The host verbs and the operator-facing half, plus the two pieces of housekeeping the first commit left half-done. #943's premise turned out to be false; the evidence for that is on #943 itself (comment 5531680866), and the scope that survived it is what shipped here: doctor detail, the log tail, and a Service Diagnostics panel.

## What an operator gets

A **Service Diagnostics** card in the config card stack, under the same `dashboard.control.enabled` flag as its neighbours. Two buttons, both read-only:

- **Run health check** — the host runs `pithead doctor --json` and the panel renders it, failures first, under the host's own summary counts.
- **Show recent log** — a bounded, redacted tail of one service's log.

This exists for the appliance, where the operator has no shell: before it, the dashboard could say *that* a service was unhealthy and never *why*.

## The trust boundary, and where I deliberately did not put a check

The dashboard container can only ask; the host decides. A compromised container writes the request spool directly, so any allowlist or bound that lives only in the container is decorative.

So `diagnostics_service.valid_container` is a **shape** check — charset and length — and explicitly **not** a copy of the host's allowlist, and the route does **not** clamp `lines`. The host owns both: it matches the container name against `PITHEAD_DIAG_CONTAINERS` by exact membership, and it clamps the line count and the byte size itself. A test asserts this rather than describing it: `test_a_wellformed_but_unlisted_container_still_spools` sends `wallet-rpc`, which the host refuses, and requires that the container spool it anyway. If that test ever fails, someone has added a second allowlist.

The host's allowlist is narrower than the compose set on purpose. `wallet-rpc` and `tari-wallet` are refused: `bundle_redact_log` is keyed to the launch-line leak class, and the wallet daemons are the two whose ordinary output most easily carries key material outside it. The support bundle may collect them because it lands as a file the operator reviews before sharing; this streams to a browser, which is a different trust context for the same bytes.

Redaction happens **before** the byte cap, not after — truncating first would leave the tail of a redacted line intact, which is the leak the redactor exists to stop.

## Housekeeping the previous commit left half-done

**The extraction.** `web/download_views.py` already held `handle_backup_download` byte-identically, but `server.py` still defined it. Removed there, registered as a plain call into the module, `import uuid` dropped with its last user. Proved byte-identical with `cmp` against a firing control, and there is no `monkeypatch.setattr(server, ...)` site anywhere in the suite, so the move cannot leave a test passing while asserting nothing. `server.py` 645 → 627; **the 645 row is deliberately not lowered.**

**A false comment.** `42-control-policy-and-host-checks.sh`'s header enumerates the control channel's action set and was already false before this change — it named 8 of the 15 action strings, missing all five `os-*` verbs. It now names all fifteen and points at 49's dispatching `case` as the authority, so the next drift is cheap to catch. Comments only; the artifact is rebuilt and parity re-checked.

## The budget decision, and why the shape is a split

`components.mjs` was at 1177/1177 and mounting the panel needed one import line. I proposed raising the row and **that was wrong**: `check_monotonic` fails any rise on a non-exempt row, and the sole `monotonic_exempt` arm names a retired path, so a hand-curated 1178 reds every run. I verified that at source before acting.

So: **split, row kept at 1177.** The `Gauge` + `SyncView` family moved to `static/syncview.mjs`. It is the cleanest cut available rather than an arbitrary one — `Gauge` is used by nothing but `SyncView`, `SyncView` by exactly one site in `App`, nothing outside the module names either, and neither touches `logic.mjs`, so the move carries no helper with it and leaves no partial family behind. Proved byte-identical with `cmp` against a firing control; the two added `export` keywords are the only difference. `components.mjs` 1177 → 1137, and the row is **not** lowered to match.

## Security review — two findings, both fixed here

A `security-reviewer` pass (sonnet) ran against this branch. It confirmed the claims I asked it to attack — no command injection survives the exact-membership test, redaction genuinely precedes truncation, the container-side shape check cannot admit anything the host then mishandles, and the CSRF gate is unchanged by the move — and it found two things I had missed. I re-derived both at source before fixing them.

**HIGH — `doctor --json` carried the dashboard's onion address into the container.** `06-doctor.sh` prints `Dashboard onion:` in full deliberately: it writes for the CLI, where the reader is the operator at their own terminal. That is right there and wrong on this path. The document crosses into the container — the party this whole channel is built not to trust — and a hidden-service address is the one value whose only security property is that nobody has it. I verified the escalation is real rather than theoretical: the dashboard service is passed 87 environment variables and **not one** is an onion address, and no application code reads one, so `diag-doctor` would have been the container's first route to it. The support bundle carries the same document unredacted because it lands as a chmod-600 file the operator reviews before sharing — same bytes, different trust context, exactly the distinction the log tail already made.

Fixed by putting the doctor output through `bundle_redact_log` — the same and only redactor the log tail uses — rather than a rule of its own, so anything added there is covered on both paths. Proven with a firing control: the document contains the address before redaction and does not after, the marker is present, the JSON is still valid, and the unrelated checks survive intact.

**LOW — the line clamp was bypassable.** `[ N -gt M ]` does not compare a digit string past 64 bits; it errors, and as the left operand of `&&` that error is swallowed, so a spool-written `{"lines": <20 digits>}` reached `--tail` uncapped. Not a confidentiality problem — redaction still runs first and the 64 KiB byte cap still holds — but the line cap is policy and was not being enforced. The digit **width** is now bounded before any arithmetic. Checked against the old logic across `5 / 200 / 201 / 1000 / 0` and the overflow value: only the overflow case changes.

The reviewer also flagged, correctly, that the tree was committed and rebased underneath it mid-review. Its findings were re-checked against the head they are fixed on.

## The over-engineering pass found one more

Done by hand on the finished diff, on merits. `await request.json()` returns a list for `[]` and a str for `"a string"` — both valid JSON with no `.get` — so `body.get("container")` raised past the handler and aiohttp answered **500** where every other malformed input on this route answers 400. The existing body test only covered text that is not JSON at all, which fails earlier at `request.json()` and so could never have caught it. Fixed, with the four shapes added to the test, each also asserting nothing was spooled.

## What was run

| | |
|---|---|
| `make test-dashboard` | **2503 passed**, total coverage 97.42% |
| Patch coverage | **100%** (71/71 executable changed lines) — computed from `coverage.xml` on the tree shipped, not from CI's target, which pins `COMPARE=origin/develop` |
| `node --test dashboard/tests/frontend/*.test.mjs` | **529 passed** |
| New route tests | 28 passed, and an **8-mutation battery killed 8/8** — one mutation per design decision (bool-vs-int on `lines`, allowlist-instead-of-shape-check, a permissive regex, each CSRF gate, a container-side clamp, the default count, a dropped actor) |
| `lint-py` `lint-js` `lint-md` `lint-docs-voice` `lint-operator-strings` `lint-topology` `lint-file-budget` `lint-pithead-parity` | all rc 0 |
| `build-pithead --check` | rc **1 before** the rebuild, rc **0 after** — a controlled pair, not a bare green. Re-run after the security fixes, same pair. |
| `bash -n` + `shfmt -i 4 -d` | clean on both `lib/pithead/` files |
| Suite validity after the fixes | the security fixes touch only `lib/pithead/`, `pithead` and `docs/`. `git rev-parse <head>:dashboard` is **byte-identical** to the tree the 2503-test run measured, so those counts stand rather than being re-asserted. |

## What I did NOT do

- **`make lint-sh` was not run.** It is this box's documented OOM path (shellcheck 0.9.0 has killed whole sessions here). CI's shell job is the gate for the two `lib/pithead/` files.
- **The two host verbs have no tier-1 shell test.** `tests/stack/` and `tests/integration/` are not this lane's paths and I did not reach into them. The verbs are covered by reading, by `bash -n`/`shfmt`, and by the redaction and clamp controls quoted above — not by a registered test. Worth a grant or a hand-off; I am flagging it rather than leaving it implied.
- **No bench, no appliance hardware.** The host verbs are proven by reading and by the tier-1 suite, not by a run against a real appliance. `diag-doctor` shells out to `doctor --json` and `diag-logs` to `docker compose logs`; neither was exercised end-to-end on hardware in this PR.
- **The `lines` default is a third copy of a host constant** (`200` lives in the host verb, in `diagnostics_service.MAX_TAIL_LINES`, and in the panel). Unlike the allowlist, a drift here is harmless because the host clamps regardless — but it is the same class of thing and I am naming it rather than hiding it.
- **This PR fires #1213's trigger condition, and I did not fix it.** #1213 was filed as latent: doctor's `dr_fail`/`dr_warn` remedial strings name CLI verbs, harmless only while nothing surfaces doctor output to a shell-less appliance. This card is that surface. Measured, not inferred: `doctor_json` (`lib/pithead/06-doctor.sh:347`) carries the raw `dr_*` text through into `checks[].message`; the panel renders that field directly at `static/diagview.mjs:110`; **30** `dr_fail`/`dr_warn` sites in `lib/pithead/` name a CLI verb; and `06-doctor.sh` has **zero** `is_appliance` branches. `06-doctor.sh` is outside this PR's grant and the defect class is different from the diagnostics plumbing, so by the split rule it is filed rather than folded in — written up on #1213 (comment 5533683005), which now wants re-milestoning off `v2.x - post-GA`. The 30 is an upper bound on CLI-verb-naming sites, **not** a measurement of how many are reachable on an appliance; that audit is owed.
- **Filed, not fixed:** #1731 (publish the host's diag allowlist so the picker cannot drift from it) and #1732 (four copies of the spool's atomic temp-and-rename; `control_service.py` is 414/414 and cannot receive a shared writer).
- **Excluded from scope by ruling:** the support-bundle download is a second PR; `onion-client-key` is out and owed a forward-filed issue; rotations wait on #911.

## Disclosures

- **`docs/` is shared and not owned by this lane.** Two files touched: `docs/dashboard.md` (a new "Service diagnostics" section) and `docs/dev/repo-map.md` (the CLI row now says the build concatenates in `LC_ALL=C` order and explains the `46a-` suffix convention — the previous "in name order" was imprecise in a way that matters, since a UTF-8 `ls` sorts `46a-` before `46-`).
- **Four `lib/pithead/` paths and the `pithead` artifact are under a controller grant** (per-file, expiring on this PR's merge): `46a-control-diagnostics.sh`, `49-control-request-loop.sh`, `pithead`, and `42-control-policy-and-host-checks.sh` **comment-only**. Grant readback re-run this cycle with a firing negative control: the granted file staged alone passed, an ungranted sibling was refused naming only itself, and the probe was restored by sha256.
- **Rebased onto `develop-v2` at `5ff0cd48`**, which was the tip when this branch was pushed; both patches were verified byte-identical across that rebase against a firing control. The tip has since moved one commit to `0e2b291c` (#1722), and I did **not** re-rebase: measured, `git rev-parse <base>:dashboard` is byte-identical (`18d4b8b4`) at both bases and the only file that commit touches is `scripts/build-pithead.sh`, which this diff does not go near — so no suite re-run is owed and a re-rebase would only restart ~13 minutes of CI. The PR still reads `MERGEABLE`.


VijitSingh97:
## Non-author review — verdict `PASS`, with a caveat about the reviewer that a reader needs

A `verifier` subagent (haiku) reviewed this branch against #913 and #943 in a fresh context. Its verdict was **PASS, no findings**, across redaction routing, the integer bound, input handling, the container allowlist, tests and docs.

**I am not pasting that as clean corroboration, because roughly half its citations do not exist.** It cited "line 180 in 46a-control-diagnostics.sh" for the clamp and "lines 163-165" for the allowlist — that file is **116 lines**. It cited "line 152 in diagnostics_views.py" for the non-object check — that file is **85 lines**. It also cited "line 1020 in docs/dashboard.md"; the passage is at 1145. The line numbers are confabulated even where the conclusions are right, which is the failure mode where specificity reads as evidence. It further reported the 8/8 mutation battery as verified — but that figure came from this PR body, so it is my own claim relayed back, not a second measurement. Treat it as one.

**So I re-derived every load-bearing claim myself, at the real lines. All of it holds:**

- **Redaction — enumerated, not sampled.** `46a-control-diagnostics.sh` has exactly two host-text emitters: `:63` (`doctor --json`) and `:107-108` (`docker compose logs`). Both pipe through `bundle_redact_log` and both redact *before* `head -c`, so a byte cap cannot leave the tail of a half-redacted line intact. There is no third path.
- **The bound.** `:102` compares the digit **width** against the cap's width before any arithmetic; `:104` does the value comparison only after that. `:95-97` sends empty and any non-digit (a `-` or `+` sign included) to the cap, and `:103` floors at 1. A 30-digit or leading-zero string is decided at `:102` without arithmetic, which is the overflow the earlier review round found.
- **Input handling.** `diagnostics_views.py:69` rejects a valid-JSON non-object with 400 before any `.get`; `:76` excludes `bool` explicitly, so `{"lines": true}` does not spool 1.
- **Allowlist.** `:85-87` is exact-string membership (`[ "$container" = "$c" ]`) over the word list at `:30` — no glob, no separator trick.
- **Not an injection sink.** `diagview.mjs:110` interpolates `${r.message}` through preact/htm, so it becomes an escaped text child, not markup.

**Two things came out of the pass that were not in the reviewer's findings:**

1. **This PR fires #1213's trigger condition** — the doctor panel surfaces remedial strings that name CLI verbs to a shell-less appliance. Filed, not fixed; measurement is in the body and on #1213 (comment 5533683005).
2. **A doc sentence described a mechanism the code does not have.** `docs/dashboard.md` said the panel shows the onion address "struck out"; `bundle_redact_log` is a sed marker replacement, so it shows `[redacted].onion`. The reviewer surfaced this and classed it as imprecision rather than a finding — I fixed it instead, in `8cf32732`. The safety claim it carried was already true; the mechanism it named was not.

**What this verdict does not cover.** The reviewer read the diff; it did not run the suite, the mutation battery, or the two host verbs. The tier-1 shell-test gap for `diag-doctor`/`diag-logs` is unchanged and still disclosed in the body. `8cf32732` is a docs-only commit, so the suite counts quoted in the body still describe the code being shipped — `git rev-parse <head>:dashboard` is unchanged by it.

VijitSingh97:
## Post-merge adversarial pass at `efc68c52` — one finding, filed as a follow-up; nothing here argues for a revert

Taken because comment 5533709192 is the author's own verifier subagent, which is not a non-author pass. The merged change is sound in the places I could break it; the finding is about a design assumption carried across a trust boundary, not about a defect in the code as written.

### Discharged — I tried to break these and could not

**The widened key allowlist is inert outside the diagnostics verbs.** `49-control-request-loop.sh:33` now admits `container` and `lines` for *every* action, which is a real widening and the right thing to check first. I swept `lib/pithead/` for any verb that reads either key from a request file: **none does.** The only other hits are `.services[].container_name` in `01-lifecycle.sh:24` (compose config, not the request) and `*.container` unit filenames in `36-quadlet-units.sh`. So the shape guard loosens and no value guard anywhere else gains a new input. The comment's reasoning holds.

**The container match is exact membership, not a pattern.** `for c in $PITHEAD_DIAG_CONTAINERS` + `[ "$container" = "$c" ]` — no glob, no separator trick, and `$container` is allowlist-constrained before it reaches `docker compose logs`.

**The onion rule works, and I have to correct my own first measurement of it.** My initial control reported the onion surviving in body text; that was my input, not the code — I used a 50-character base32 label where a v3 onion is 56. Re-run correctly, `[a-z2-7]{56}\.onion` redacts in plain text *and* inside a JSON string (`"dashboard_onion": "[redacted].onion",`), and a 55-character negative control does **not** match, so the rule is length-exact as intended. F1's fix is real.

**The F2 clamp fix is correct — and since it has no test at its own tier, this run is the only measurement of it.** The shipped logic lifted verbatim and driven across the boundary:

| in | 5 | 200 | 201 | 1000 | 0 | `000` | `999` | `-5` | `1.5` | empty | `99999999999999999999` |
|---|---|---|---|---|---|---|---|---|---|---|---|
| out | 5 | 200 | 200 | 200 | 1 | 1 | 200 | 200 | 200 | 200 | **200** |

Every result inside `[1, 200]`, the 20-digit overflow that was the bug included. Bounding the digit *width* before any arithmetic is the right shape for this.

### F1 — should-fix, follow-up, not a revert: the redactor is positional, and this verb changed the trust context it was designed for

`bundle_redact_log` redacts addresses **by argv position** — `--wallet <addr>`, `--merge-mine <host> <addr>` — and this is a deliberate ruling, not an oversight: `07-support-bundle.sh:9-15` and `selftest-bundle-redact-log.sh:15-18` both record that no length bar reaches all three Tari forms (91, 48, 67 chars) and that a bar low enough to try starts eating ordinary log tokens. The onion is shape-keyed precisely because, in that file's own words, it is *"the one value that also appears away from the launch line."*

Measured, with the argv positive control firing first:

| input shape | result |
|---|---|
| `--wallet <addr> --stratum …` (the designed case) | **redacted** |
| `… SideChain  Wallet address: <addr>` | **survives** |
| `… payout to <addr>` | **survives** |
| a bare address alone on a line | **survives** |
| `{"address":"<addr>"}` | **survives** |

Driven with the suite's own checksum-valid dummy; the value is not reproduced here.

That is correct behaviour for the artifact it was written for — a chmod-600 tarball the operator reviews before sharing. `46a-control-diagnostics.sh` reuses it for a verb that, in the slice header's own words, *"streams to a browser over the network, which is a different trust context for the same bytes."* **That exact sentence is used once, to exclude `wallet-rpc` and `tari-wallet` — and it is not carried to the body-text case for the nine services that are included, `p2pool` among them.**

So the only thing standing between a body-text address and the browser is the claim that no allowlisted service ever prints one outside its launch line. That claim is **nowhere stated, nowhere tested, and depends on upstream software this repo does not control.** The redactor's own selftest cannot catch it: every wallet and Tari row there is a launch-line row, so the body-text case is absent rather than passing — green by construction.

**I have not shown a live leak, and I want to be exact about that.** I have shown the redactor's coverage boundary and that the argument for the allowlist stops at the boundary. Whether any of the nine actually emits an address in body text needs a real stack, and I hold no bench and took none.

Two things make this cheap to close rather than alarming: the remedy is not the general shape rule #1585 already ruled out, because the *Monero* form is separately shape-checkable and this repo already implements that check (`25-address-types.sh:20 monero_address_type`); and a body-text row in `selftest-bundle-redact-log.sh` would pin whichever answer the bench gives. Filed separately.

### Already filed, not re-filed

**#1745** already carries the tier-1 gap — neither `control_diag_doctor` nor `control_diag_logs` has any `tests/stack` coverage (I re-derived this: a grep of `tests/` for either function or either verb string returns nothing at all). Worth adding there: the allowlist membership check, the redact-before-cap ordering and the F2 clamp are all shell behaviour in `lib/pithead/`, so tier 1 is where the testing strategy puts them, and all three are currently unheld. **#1731** covers the picker's duplicate allowlist.

### Assumed — not verified by me

The dashboard half in full (`diagnostics_service.py`, `diagnostics_views.py`, `diagview.mjs`) beyond reading the routes' gating; the 188-line frontend and 208-line Python suites; the `handle_backup_download` byte-identical move and its `cmp` control; the `components.mjs` 1177 → 1137 split and the budget rows. No `make` target run, no bench, no container, no live stack, no CI watched.

Issue #944, opened 2026-08-14
Static offline helper to author a recovery pithead-config.json for the media stick


VijitSingh97:
From the 2026-08-13 parity matrix — small, optional UX idea.

The removable-media config channel is the forgotten-password recovery path, aimed at exactly the operator least able to hand-author JSON. Today the docs say "copy config.json from a machine you already set up, or write one by hand" — but the lost-password scenario often has no second machine.

Idea: ship a single static HTML file (no server, no dependencies — it can live in the repo and in the docs) that renders the handful of recovery-relevant fields (new dashboard password, wallet addresses, ssh toggle), validates shapes client-side, and downloads a valid `pithead-config.json` for the stick. Pure client-side authoring for the person standing at a laptop with a USB stick.

PR #1737, opened 2026-09-04
fix(os): #1215 the appliance image ships the compose file of the release it names, not the tree's

### LINKED ISSUE #1215: Appliance OS builds bake a compose file that can outrun the image pins — stage version-coupled content from the STACK_VERSION ref
Follow-up to #1098, splitting out the half its fix deliberately did not attempt.

The PR for #1098 ships the two same-tree guards: a tier-1 static check that every script-based healthcheck compose names is actually COPYed by its service's Dockerfile, and a gracefully-degrading healthcheck so a missing script reports healthy instead of permanently unhealthy. What it does not fix is the skew mechanism itself:

- `STACK_VERSION` (and so every pinned image tag) re-derives from `VERSION` only at a release cut.
- `docker-compose.yml` content advances on every merge to develop-v2.
- Any OS image built between cuts — or from a dev-ahead checkout — therefore bakes a compose file that can assume image content the pinned tags predate. The healthcheck was one instance; any compose change depending on new image content has the same exposure window, and the symptom is whatever that feature silently does without its image half.

The correct fix is structural: `os/build-image.sh` should stage `docker-compose.yml` (and anything else version-coupled to the image pins) from the git ref `STACK_VERSION` actually resolves to, falling back to the working tree only when the tag does not exist yet (mid release-prep). That is real surgery on the build pipeline: `tests/os/verify-image.sh`'s byte-identical-to-tree invariant (documented in `docs/dev/appliance-release.md`) has to be reworked to compare against the staged ref instead, and the change needs bench validation of a release build end to end — which is why it was declined as a single-session patch.

Until then the exposure is bounded by the #1098 guards plus release discipline: an OS image cut at a proper release commit has no skew by construction.

VijitSingh97:
## What changes

The appliance image ships the compose file of the release it names, not the tree's. Every `image:` in `docker-compose.yml` is pinned by `STACK_VERSION`, which the appliance derives from its baked `VERSION`, so an image built from a tree ahead of that release baked a compose file that assumed image content the pinned tags predate — #1098 was one instance (a healthcheck script the published image did not carry; a container unhealthy for good).

- `os/build-image.sh`: new `stage_compose` (above the test seam) stages `docker-compose.yml` into `os/build/stage/` from the tag `v<VERSION>` when that tag exists, from the tree while it does not (that is the release being prepared — `release.sh` tags this very commit), and REFUSES the one silent case: a clone that has the tag on origin but never fetched it (`git ls-remote --exit-code`; any other answer, including no remote or no network, leaves the tree as the source so a build never hangs on the network). The choice is stamped into `COMPOSE_SOURCE` as `tag NAME SHA` (the resolved commit, because a tag can be re-pointed) or `tree`, and printed.
- `os/rootfs/Dockerfile`: `COPY`s the staged file and the stamp from `os/build/stage/` instead of the root `docker-compose.yml`. **Line-neutral** (406/406 ceiling): two comment lines rewritten, the two `COPY` lines re-cut.
- `tests/os/verify-image.sh`: `compose_reference` (above the seam) resolves the file the shipped compose must equal from the image's own stamp; the check now reads `shipped compose file matches its stamped source (<stamp>)` and FAILS on a stamp that is missing, malformed, names a version other than the shipped `VERSION`'s, or names a commit the checkout lacks — a comparison that cannot run reads as a failure, never a pass. The file crosses 400 and gains a budget row at its true count (419).
- `docs/dev/appliance-release.md`: the rule, its cost, and the verify-image sentence, in the house voice (`make lint-docs-voice` green).
- **CI (second commit, after the first run went red):** `os/build-image.sh --stage-only` stages `os/build/stage/` and stops before docker; `.github/workflows/os-rootfs.yml`'s build step fetches the tag `VERSION` names into its shallow checkout (`git fetch --depth=1 origin +refs/tags/<tag>:refs/tags/<tag>`, absent on origin = the release being prepared, the tree is the source) and runs that step before its own `docker build`. The first run failed at the Dockerfile's COPY because the workflow builds the Dockerfile directly and never ran the staging step. **Shared file touched:** `.github/workflows/os-rootfs.yml` is a serialized file, committed under the one-shot `.lane-override` for exactly that reason. Not done: a NAMED Dockerfile refusal when the stage is absent — `COPY` fails before any `RUN` can speak, so the honest signal stays the workflow step's own output (`==> compose file staged from: …`).

**What this costs, named:** a compose change on `develop-v2` is no longer exercised on the appliance until the next cut, where the tag does not exist yet and the tree is the source. Between cuts the appliance ran the tip's compose against the last release's images, which is the skew this issue exists to remove, so the KVM battery now proves what ships rather than a pairing that never does. The trade the other way — tree `pithead` driving the tag's compose — fails LOUDLY (an unknown service or profile) where the old skew failed silently.

## Not done

- `docker-compose.yml` is the only file staged. `config.reference.json`, `pithead` and `build/` templates ship from the tree as before: they are consumed by `pithead` (tree) rather than by the pinned images, so the coupling the issue describes does not run through them. If a second file turns out to be image-coupled, `stage_compose` is the one place to add it.
- No release-variant build (no `--ssh`, real key) — this seat holds no release key. The staging code path is identical in both variants; the release-mode `verify-image` half is unexercised here.
- The full `tests/stack/run.sh` was not run on this box alongside the bench chain (a suite beside an image build is not a stated-safe pair); CI's shell-tests job is the run of record for it. The new domain file was driven alone through `lib.sh` (below).

## What was RUN — PROVEN BY ME unless marked

**Tier-1 — the new domain file, driven alone through `tests/stack/lib.sh`:** 25 passed / 0 failed at the bench commit `bc6945a9`, three `==` sections (`tests/inventory.sh` lists the file with section count 3); 27 passed / 0 failed at the CI-fix head with the `--stage-only` section added (four sections).

**Mutation battery** (each mutant `git diff --numstat` 1/1 on the staged file, judged by the rows it reddens, restored by `git checkout --`):

| mutant | rows reddened | reads as |
|---|---|---|
| M1 drop the unfetched-tag refusal (`if false`) | 4: the REFUSED row, the two message rows, "stages nothing" | the refusal is load-bearing, not decorative |
| M2 tag path copies the tree instead of `git show` | 2: "staged file is the TAG's", "tree content did not leak" | the tag-vs-tree distinction is measured by content |
| M3 `compose_reference` drops the VERSION equality | 1: "another version than the shipped VERSION is refused" | the stamp-to-VERSION check is its own axis |
| M4 tree path ships `HEAD:` instead of the working tree | 1: "ships the working tree's compose file byte for byte" | the fallback is the WORKING tree, not the last commit |
| M5 drop the `--stage-only` case arm | 1: "--stage-only is accepted and recorded" | the flag is parsed, not assumed |
| M6 move the stop below the wizard image step | 1: "the stop sits after the staging line and before the wizard image step" | the stop is where CI needs it |

M1 and M2 kill disjoint rows, so they are two instruments, not one. M5/M6 were run against the committed CI-fix head, restored by `git checkout --`. A real `os/build-image.sh --stage-only` run in a clean worktree at that head printed `==> compose file staged from: tag v1.20.0 296fe6af…` then `==> --stage-only: os/build/stage/ is ready; stopping before the build`, rc 0, with exactly `COMPOSE_SOURCE` and `docker-compose.yml` in the stage directory. What the battery does not vary: the Dockerfile `COPY` paths (wiring rows are static text assertions, labelled as such in the file — they prove the two ends name the same path, not that docker honours it; the bench chain below is what proves that).

**Lane-guard arming readback** (`.fleet-role=appliance`, each case staged ALONE, verdict from the hook's rc): `tests/stack/test-appliance-build-compose-source.sh` rc 0 (ACCEPTED); negative control `tests/stack/test-zz-ungranted-control.sh` rc 1 (REFUSED, names the path); `tests/stack/run.sh` rc 0 (rides the live line). Grant: fleet `c63cc19`. The e2e and tests lanes were told before the file was created; both answered nothing in flight.

**Gates at the branch head:** `scripts/lint-file-budget.sh` OK (row added for `tests/os/verify-image.sh` at 419; Dockerfile unchanged at 406/406); `make lint-docs-voice` OK; shellcheck **0.11.0 (the pin)** `--severity=warning` clean on `os/build-image.sh`, `tests/os/verify-image.sh` and the new domain file; `shfmt -i 4 -d` clean on all four shell files. `make lint-sh` was NOT run here (fleet-serialized OOM path; CI runs it).

**Bench chain (the KVM bench host), tree frozen at the bench commit `bc6945a9`:** leg 1 GREEN — `os/build-image.sh --ssh <harness key>` rc 0 (the staged stamp `os/build/stage/COMPOSE_SOURCE` reads `tag v1.20.0 296fe6af551b773bae49486e98517ac274b896cd`), `os/rauc/mkimage.sh --dev` rc 0, `tests/os/verify-image.sh --test` 102 passed / 0 failed, including the new row `shipped compose file matches its stamped source (tag v1.20.0 296fe6af551b773bae49486e98517ac274b896cd)`. **The KVM battery (`tests/os/run.sh --phase all`) is NOT RUN:** `kvm-preflight.sh` refused to boot the guest (host MemAvailable 19973 MiB under the 20480 MiB bar — the bench host carries the fleet and the box's own stack, and after two host hangs this week the bar is not lowered for a run). A relaunch is armed and waits for the host to clear the bar. **Per the controller's ruling the merge holds on the battery's rc file**; this body is amended by PATCH with the result and the run it came from when it lands. Why the battery is load-bearing here and not a formality: `git diff v1.20.0 HEAD -- docker-compose.yml` is 25+/5-, so the appliance now boots the tag's compose against the tag's images with the tree's `pithead`, a pairing no run has exercised yet; verify-image proves the file shipped, only the battery proves the stack comes up on it. The bench chain above ran at the bench commit; the head adds a docs commit and the CI-fix commit on top of it. Neither changes what the image contains (`docs/` is not COPYed; the build-script change is a flag that stops before the build), but the bench evidence is re-run at the head anyway — the relaunch builds from the head, not from a reconstruction. **CI run of record for the staged build:** `Build + scan the appliance rootfs` PASSED on the CI-fix head — run 33821598090, job 100865244048 (5m46s), i.e. the workflow staged from the fetched tag and the Dockerfile COPY that failed the first run succeeded. The KVM battery is still NOT RUN at this head: the bench chain yielded to the #1651 battery (GA gate) and is relaunched by the appliance lane after it..

## Over-engineering pass (by hand — the PR-gate hook is a prompt, not a proof, see below)

Two functions were added and each exists for the test seam: `stage_compose` could be inlined at its one call site and `compose_reference` into its one `chk`, and both were rejected because neither could then be driven at tier 1 (the build needs docker, the verifier needs root and a loop device). No other adjacent merge was available. Nothing moved; the diff is additive except the two re-cut Dockerfile lines and the one replaced `chk`.

The PR-gate hook is a one-shot prompt that never checks a review happened (it denied the first create for this branch and cleared on the retry); the pass above was run by hand against THIS diff.

Closes nothing by keyword (closing keywords are inert on `develop-v2`); the controller closes #1215 by hand on merge.

https://claude.ai/code/session_01FsyEEM3JGMmzXusHCkk69E


VijitSingh97:
## Two merge-time defects this branch cannot see, recorded now so they are not rediscovered at merge

This PR waits outside the image freeze, so neither is urgent. Both are invisible from any single head,
which is why no review here has caught them. **Measured against the current `develop-v2` (`38d8e05b`).**

### 1. `tests/stack/run.sh` lands at 443 against a 440 ceiling

```
git merge-tree --write-tree origin/develop-v2 6571c16c   -> CONFLICT (see 2)
git show <that tree>:tests/stack/run.sh | awk 'END{print NR}'   -> 443
docs/dev/file-budget.tsv row                                     -> tests/stack/run.sh  440
```

`run.sh` itself **auto-merges cleanly** — it is not the conflicting path, so that 443 is a real merge
result and not a conflict artifact. Counted with the gate's own counter
(`awk 'END { print NR + 0 }'`, `scripts/lint-file-budget.sh:122`), not `wc -l`.

This branch's base had `run.sh` at 282 and it adds **+3/-0**; `develop-v2` has since moved that file to
**exactly 440**, its own ceiling, with zero headroom. Neither side is wrong alone. **It cannot be fixed
after the merge** — `check_monotonic` refuses to raise a ceiling once the row exists on the base, and
`monotonic_exempt` matches only `lib/pithead/99-remainder.sh` (read at source). The three lines must be
paid for on this branch.

### 2. The conflict is in `docs/dev/file-budget.tsv`, which is the one file you must not resolve by taking a side

`merge-tree` reports exactly one conflicting path, with all three stages present:

```
CONFLICT (content): Merge conflict in docs/dev/file-budget.tsv
(everything else auto-merges: testing-strategy.md, os/rootfs/Dockerfile,
 tests/os/verify-image.sh, tests/stack/run.sh)
```

That file is **regenerated per-lane and is never merged by taking a side** — taking one bakes another
lane's ceilings into the tree, which is how a ceiling silently rises. Regenerate it on the rebased
branch rather than resolving the hunk. Note the two defects interact: regenerating after fixing (1)
gives the right row; regenerating before it records 443 and hides the problem.

### Instrument note, because the obvious form of this check lies

`git merge-tree`'s clean/conflict tell is the OUTPUT SHAPE — a bare tree OID versus the OID plus a file
list — **not the exit code**, and piping it to `head` makes `$?` `head`'s rather than git's. Firing
control run before trusting any clean reading: `origin/develop` x `origin/develop-v2` (pull #1057,
known conflicting) reports 35 lines, so the instrument can produce the other answer.

The coupling class is `lane-fixes-d4`'s finding from #1813. `run.sh` is the **only** budgeted file
touched by more than one open PR, but it is coupled across three: #1813 (merges clean at 440/440,
having paid for its lines with a comment compression), #1747 (443), and this one (443).

VijitSingh97:
## Correction to my own instrument note above — `git merge-tree` DOES set its exit code

My comment said the clean/conflict tell is "the OUTPUT SHAPE ... **not the exit code**". **That is
wrong, and my own run contained the disproof.** I relayed the claim from another lane and applied it
without checking it against data I had already collected — which is exactly how a wrong claim gets
durable: each hop adds apparent corroboration and no evidence.

Re-measured just now on this box's git, rc captured unpiped via command substitution, four pairs:

```
develop-v2 x 463973d7 (#1813) -> rc=0   1 line   CLEAN
develop-v2 x 10f3034e (#1747) -> rc=0   1 line   CLEAN
develop-v2 x 6571c16c (#1737) -> rc=1  11 lines  CONFLICT
develop    x develop-v2 (#1057) -> rc=1  35 lines  CONFLICT   <- firing control
```

**rc and shape agree on every pair.** The real rule is narrower and more useful: **never pipe the
thing whose rc you are about to read** — `git merge-tree ... | head` makes `$?` `head`'s, which is the
pipeline artefact that got generalised into a property of the tool. Reading output shape is a valid
*recovery* once you have already piped, not the primary tell.

**Nothing in the finding changes.** The line counts were read out of the written trees with the gate's
own counter, never off a status, so 443-vs-440 stands exactly as posted.

One figure reconciled while I was here, so nobody quotes the two as disagreeing: my "35 lines" on the
control is a **total output** count, not a file count — 18 conflict-stage entries (3 per file = **6
conflicted files**, matching the `--name-only` reading) plus 15 `Auto-merging`/`CONFLICT` messages plus
the tree OID. Same measurement, two units.

VijitSingh97:
## Budget coupling re-derived at the NEW base — unchanged, and one correction of my own wording

My earlier comment measured this against base `38d8e05b`, which has since moved to `0e32da0e`. Re-run rather than assumed; **the result is unchanged.**

At base `0e32da0e`:

- `tests/stack/run.sh` ceiling is **440**, and the base file is **exactly 440** — zero headroom.
- This branch contributes **+3**: `run.sh` is 279 at its merge-base `8e248b1e` and 282 at head `6571c16c`.
- `git merge-tree --write-tree 0e32da0e 6571c16c` returns **rc 1**. The conflict is still **`docs/dev/file-budget.tsv` alone** — it is the only path with stage entries — while `run.sh`, `os/rootfs/Dockerfile`, `tests/os/verify-image.sh` and `docs/dev/testing-strategy.md` all report `Auto-merging`.

**The correction to my own earlier wording, which matters for how much weight this number carries:** because that merge conflicts, **no tree is written**, so the 443 here is NOT read out of a merge result the way it is on #1747. It is arithmetic — base 440 plus this branch's +3 on a file that auto-merges cleanly. That is a sound inference and I still expect 443, but it is a derived figure, not a measurement, and I should have said so the first time.

(A related trap while checking it: grepping the `merge-tree` output for `tests/stack/run.sh` returns a hit — the `Auto-merging tests/stack/run.sh` line. A match count is not a match. The conflicted set is the stage entries, not the message lines.)

**Do not resolve `docs/dev/file-budget.tsv` by taking a side.** That file is regenerated per-lane; taking either side bakes another lane's ceilings into this branch.

The three lines still have to be paid on this branch — `check_monotonic` will not raise the row after the fact, since ceilings only go down and `monotonic_exempt` matches only `lib/pithead/99-remainder.sh`.

Recording it here only because this PR sits outside the freeze and will be picked up later, by which time the base will have moved again: re-run the two commands above rather than trusting these figures.

VijitSingh97:

VijitSingh97:
## This PR's body says it closes nothing by keyword, and GitHub resolves it as closing #1215

Not my PR and not a request to change the outcome — only a report that the body and the
mechanism disagree, because the disagreement is silent and not undoable by whoever merges.

**Measured, not read:**

```
gh api graphql -f query='{repository(owner:"p2pool-starter-stack",name:"pithead"){
  pullRequest(number:1737){ baseRefName closingIssuesReferences(first:10){nodes{number state}} }}}'
-> baseRefName "develop", closingIssuesReferences: [{number: 1215, state: "OPEN"}]
```

Control, so this is not a query that says "closes" about everything: the same query returns an
EMPTY list for pull 1895 and pull 1893.

**Why it resolves.** The body's own disclaimer is the closing keyword:

> Closes nothing by keyword (closing keywords are inert on `develop-v2`); the controller closes
> #1215 by hand on merge.

`closes` + `#1215` in that second clause is what the parser sees. The surrounding words do not
defuse it, and neither does "by hand". Both halves of the sentence were true when written — the
base was `develop-v2` then — and the 2026-09-06 ruling retargeted this PR to `develop`, which is
the repository's **default** branch, where closing keywords fire.

**What to do is yours, and either is one edit:**

- If #1215 SHOULD close when this merges: leave it, and delete the "closes nothing by keyword" and
  "by hand" sentences, which now instruct a merger to do something redundant.
- If #1215 should stay open (it names surfaces beyond the compose file): rewrite the clause as
  "the controller closes #1215 on merge" without the number adjacent to the keyword, or
  "**Addresses** #1215". Then re-run the query above and confirm an empty list.

**This was repo-wide.** I swept all seven open PRs on `develop` with the query above. Two carried a
body that contradicted its own resolution: this one, and pull 1885, whose `Closes #1879 by hand`
would have auto-closed an issue three documents say must stay open — fixed there, verified empty.
pull 1898 and pull 1893 had stale prose with correct outcomes; corrected in place. pull 1894
(#1887) and pull 1747 (#944) resolve to their own issues, which looks intended, but their authors
should confirm the PR closes the issue on **every** surface the issue names — the keyword is a
claim about completeness, not a cross-reference.

Credit where due: the auto-close class was caught by the reviewer ephemeral on pull 1885, not by
me. I re-derived it at source and widened the sweep.

PR #1738, opened 2026-09-04
fix(ci): #1452 #1608 the file-budget ratchet actually runs in CI, and its NOTE stops saying it never skips there

VijitSingh97:
## The finding, ahead of the fix

`Lint per-surface` has been asserting the file budget is monotonic **in the same run in which it printed the skip that means it never checked**. `check_monotonic` could not resolve a base ref, took its resolve-failure path, returned 0, and `run_gate` went on to print

```
```

Ceilings-only-go-down is the ratchet the whole file budget rests on, and several standing review rulings reason about how it behaves against a moved base. On CI pull-request runs it has not been enforced at all. It fired only on a developer machine running `make lint` locally — a gate that works everywhere except the place it is relied on.

The per-row half (this file is over its recorded ceiling) is unaffected and does run. This is a partial hole, not a dead gate.

## Cause

`lint-surfaces`' `actions/checkout` set only `persist-credentials: false`, so `fetch-depth` defaulted to 1. `resolve_base_ref` tries `origin/develop-v2`, `develop-v2`, `origin/develop`, `develop`; a depth-1 checkout has none of them, and a `pull_request` event materialises a single detached merge-ref SHA with no branch refs at all.

## The fix, and why the one line is the right one

`fetch-depth: 0` on that checkout — matching the two jobs in this same file that already set it (lines 50 and 383) for the same class of reason.

**Proven at the pinned action sha, not taken from memory.** `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1`, `dist/index.js`:

- line 41816 — `if (settings.fetchDepth <= 0) {` … `let refSpec = getRefSpecForAllHistory(settings.ref, settings.commit)`
- line 41506 — `getRefSpecForAllHistory` returns `['+refs/heads/*:refs/remotes/origin/*', tagsRefSpec]`

So `fetch-depth: 0` populates `refs/remotes/origin/*` and `origin/develop-v2` exists afterwards. I read that file at the pinned sha rather than reasoning about what the action "should" do.

## The NOTE is the second half of the defect

`This never happens in CI.` was a false statement in durable in-repo text, sitting exactly where the next reader would stop investigating: grep the skip, find the reassurance, conclude it is a local-only artifact. It now says what is true, names a ref-less checkout as the cause, and points at the flag whose removal would bring the skip back. The candidate list in the message was also stale — it said `(origin/develop or develop)` for a function that tries four candidates, including both `develop-v2` forms.

## Budget

`.github/workflows/ci.yml` goes **512 -> 513** against its recorded ceiling of **513**. That is AT the ceiling, and `run_gate`'s comparison is `-gt`, so it passes with zero headroom left. The `lint-surfaces` block is not reflowed. `scripts/lint-file-budget.sh` goes 347 -> 351 and has no row (under the 400 target).

## Overlap with PR #1701, checked before editing

#1701 is open against the same file. I read its diff first. Its hunk is inside `check_monotonic`'s `while` loop and documents a **different** skip — the `-n "$new_ceiling"` absent-new-row case (#1430). Mine is the base-ref resolve-failure arm about twenty lines above it. No textual overlap and no semantic tension: #1701 argues its skip is safe, this PR is about a skip that was not.

## Lane-guard arming — the readback

Proven both ways, each file staged ALONE, with a negative control fired first-class rather than assumed:

| staged alone | lane-guard rc |
|---|---|
| `.github/workflows/ci.yml` (named-file grant) | **0** |
| `scripts/lint-file-budget.sh` (controller w101 grant) | **0** |
| `scripts/lint-file-budget-selftest.sh`, ungranted sibling, **real 2-line modification** | **1**, and it named the path |

The negative control carried a real diff (`git diff --cached --numstat` = `2 0`), not a `touch` — a `touch` stages nothing and passes vacuously.

## What was RUN

```
make lint-file-budget          # rc 0 — 31 self-test assertions ok, 0 FAIL, then the real gate: "file budget OK"
shfmt -i 4 -d scripts/lint-file-budget.sh              # rc 0
shellcheck --severity=warning -x scripts/lint-file-budget.sh   # rc 0 (the pin's own severity, per lint-sh)
uvx yamllint .github/workflows/ci.yml                  # rc 0
```

## What this PR does NOT do

- **#1608's suggestion 3 — make an unresolvable base ref FATAL under `GITHUB_ACTIONS` — is not here.** It is filed as its own issue so it survives #1608 closing. I checked it would have been safe to add (`lint-file-budget` is invoked from exactly one workflow job, and the self-test's synthetic repos all resolve a base because they are `git init -b develop`, which is why the existing "raising a ceiling FAILS" case is green today) — but an arm that no self-test exercises is the shape this repo files issues about, and the routing for this PR was to keep the edit minimal. Without it, `fetch-depth: 0` is one deleted line from silently restoring the defect.
- It does not check whether any **other** job depends on unfetched refs the same way. #1452 says only `lint-file-budget` was checked; that is still true.

## Consequence worth stating plainly

This turns the monotonic ratchet ON for every pull-request run. A PR that raises a ceiling has been passing `Lint per-surface` and failing only on someone's local `make lint`. After this merges it goes RED in CI. That is the rule being enforced, not a new gate — but a lane meeting it for the first time will read it as one.

Refs #1452, #1608

## Over-engineering pass (by hand, before opening — findings and what happened to each)

1. **CUT, applied.** The new comment's last line ("any job that reaches this line has a checkout without the base branch refs") restated what the `echo` two lines below it now prints. Removed. The script edit is +4 lines, not +5.
2. **Rejected as over-engineering for this scope:** a `GITHUB_ACTIONS` fail-closed arm. Reasoning above; filed separately rather than dropped.
3. **Nothing abstracted.** No helper, no flag, no new config key. The CI half of the fix is one YAML line and could not be fewer.
4. **No test added, deliberately.** What broke is the CI checkout, not the script — `resolve_base_ref` was always correct, which is why the existing suite is green over it. The self-test already asserts the resolve path in both lane directions ("a v2-lane branch behind the moved lane tip still ratchets against develop-v2", "a develop-lane checkout still resolves to develop"). A test that asserted `fetch-depth: 0` is present in the YAML would be a static assertion proving the text, not the behaviour — this repo's own named anti-pattern. The behaviour is proven by the CI control below instead.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -496,6 +496,7 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
+          fetch-depth: 0  # lint-file-budget's ratchet needs origin/develop-v2 to exist, or it skips (#1452, #1608)
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version: "20"
--- a/scripts/lint-file-budget.sh
+++ b/scripts/lint-file-budget.sh
@@ -245,8 +245,12 @@ check_monotonic() {
     local base old_lines new_lines fail=0 path old_ceiling new_ceiling actual
     base=$(resolve_base_ref)
     if [ -z "$base" ]; then
-        echo "file-budget: NOTE — no base ref (origin/develop or develop) resolvable; skipping the" \
-            "monotonic-ceiling check. This never happens in CI." >&2
+        # It DID happen in CI — on every pull_request run, until lint-surfaces was given
+        # fetch-depth: 0 (#1452, #1608). A depth-1 checkout resolves none of the four candidates,
+        # so the ratchet skipped here while the gate went on to print "file budget OK".
+        echo "file-budget: NOTE — no base ref (origin/develop-v2, develop-v2, origin/develop or" \
+            "develop) resolvable; skipping the monotonic-ceiling check. A checkout without the base" \
+            "branch refs does this; in CI it means the job has lost its fetch-depth: 0 (#1452)." >&2
         return 0
     fi
     [ -f "$BUDGET_FILE" ] || return 0


VijitSingh97:
## The pre-fix half of the controlled pair, measured rather than cited

The body argues the fix from a mechanism (the pinned `actions/checkout` source) and will argue it from a fired control on this branch. This is the third leg: what the **unfixed** job actually printed, pulled from the job logs of two of the four runs #1608 named.

Both fetched by the API route (`gh api repos/O/R/actions/jobs/<jid>/logs`) rather than `gh run view --log-failed`, which has returned zero lines on a run whose log exists:

| run | job | `no base ref` NOTE | `file budget OK` |
|---|---|---|---|
| `33300834176` | `99228546530` | **1** | **1** |
| `33299329353` | `99224359058` | **1** | **1** |

Verbatim from both:

```
file-budget: NOTE — no base ref (origin/develop or develop) resolvable; skipping the monotonic-ceiling check. This never happens in CI.
```

**The two counts being 1 and 1 in the same job is the whole finding.** It is not that the ratchet was skipped, and separately that the gate passed — it is that `Lint per-surface` printed the skip and then printed `file budget OK — ... docs/dev/file-budget.tsv is monotonic` in the same run, asserting the property it had just declined to check. A reader scrolling that log sees the assertion; the skip is one stderr line in 34 KB.

That also settles the sentence being deleted. `This never happens in CI.` was not merely unproven — the log carrying it *is* a CI run.

VijitSingh97:
## The ratchet FIRED in real CI — control result, with the one caveat that keeps it honest

Run `33820131896`, job `100860875541`, head `aed577a7` (the temporary control commit that raises `scripts/trivyignore-watch.sh`'s ceiling 599 -> 700, and nothing else):

```
2026-09-04T00:18:00.5652632Z bash scripts/lint-file-budget.sh
2026-09-04T00:18:04.3039632Z file-budget: FAIL — docs/dev/file-budget.tsv raises scripts/trivyignore-watch.sh's ceiling from 599 to 700. Ceilings only go down.
2026-09-04T00:18:04.3869851Z make: *** [Makefile:138: lint-file-budget] Error 1
```

`no base ref` NOTE count in that job log: **0**. So the ratchet resolved a base, compared against it, and refused a raise — in CI, on this workflow, which it has never done before. Paired with the two pre-fix logs posted above (NOTE present **and** `file budget OK` present, once each, in the same job), that is a before/after on the same instrument rather than an argument about it.

Direct evidence of the mechanism, from the same log: the checkout step logs `* [new branch] ... -> origin/...` lines, i.e. the `+refs/heads/*:refs/remotes/origin/*` refspec that `fetch-depth: 0` selects. It took **3.2 seconds**.

### The caveat, stated because the run's own conclusion does not read `failure`

That job's conclusion is **`cancelled`**, not `failure` — its `timeout-minutes: 15` expired seconds after the gate failed. The FAIL line above is unambiguous and is the gate's own message, but I am not willing to cite a control whose verdict needs a paragraph, so the job has been **re-run** and the clean conclusion will be posted here when it lands.

### Why the timeout, since the obvious suspect was my own change

It was not `fetch-depth: 0`. From the job log's timestamps, the 15 minutes was three `npx --yes` downloads:

| step | elapsed |
| --- | --- |
| `npx @biomejs/biome@2.5.0` | 3m 21s (then checked 28 files in 80 ms) |
| `npx markdownlint-cli2@0.18.1` | 4m 22s |
| `npx @taplo/cli@0.7.0` | 7m 01s |
| `actions/checkout` (this PR's change) | **3.2s** |

The same job ran end-to-end in 27–34 s on four runs of 2026-08-30. npm had a bad few minutes. **Filed separately as #1746**, because the registry episode is transient but the thing that turned it into a cancelled gate is not: `lint-surfaces` runs all ten gates in one step with the three downloads at the front, and the five repo-local gates that took under 6 seconds combined — `lint-file-budget` among them — are the ones at the back that get cancelled.

### ⛔ State of this PR

Held as a **draft**. `aed577a7` is the control and its message says DO NOT MERGE. **Do not push to this branch until run `33820131896` finishes** — `cancel-in-progress` is true for `pull_request`, so a push would cancel the re-run and take the clean control conclusion with it. After it lands: revert `aed577a7`, force-with-lease, un-draft. The mergeable state of this PR is `83afe451` — the fix alone, two files, +1 YAML line and +4 comment lines.

VijitSingh97:
## Control commit removed — the branch is the fix alone

`aed577a7` ("TEMPORARY CONTROL — REVERTED IN THE NEXT COMMIT") is off the branch. Head is now `83afe451`, the fix commit, and the branch was force-with-lease'd from the exact expected value so nothing else could have been overwritten.

Verified by CONTENT rather than by the push's exit code:

```
git ls-remote origin fix/1452-1608-file-budget-ratchet-in-ci   ->  83afe451…
git show 83afe451:docs/dev/file-budget.tsv | grep -F scripts/trivyignore-watch.sh
    scripts/trivyignore-watch.sh	599        <- the control raised this to 700; it is back
gh pr view 1738 --jq .headRefOid            ->  83afe451…
```

## The re-run added nothing, and it failed the same way — read the ORIGINAL run for the proof

Run `33820131896` (`event=pull_request`, head `aed577a7`) is now **complete, conclusion `cancelled`**. Its `Lint per-surface` job ran `2026-09-04T00:19:59Z -> 00:35:15Z` — **15m16s, the 30-minute-per-job budget's 15-minute step timeout again**, which is [#1746](https://github.com/p2pool-starter-stack/pithead/issues/1746), not anything this PR does.

**It is worse evidence than the first run, not better.** Grepping its own job log (29,088 bytes, pulled via the jobs API rather than `--log-failed`, which returns zero lines on a log that exists):

| needle | count |
|---|---|
| `file-budget: FAIL …` | **0** |
| `no base ref` NOTE | 0 |
| `file budget OK` | 0 |

All three zero means the job was killed **before the file-budget gate ran at all** this time. So the re-run neither confirms nor contradicts anything.

**The proof already on this PR stands and is the one to read**: the ratchet FIRED in real CI on the first control run — `file-budget: FAIL — … raises scripts/trivyignore-watch.sh's ceiling from 599 to 700` at `00:18:04`, NOTE count 0, against two pre-fix job logs that each carry the NOTE *and* `file budget OK`. That job's `cancelled` conclusion was the timeout expiring ~5s **after** the gate had already failed, which is the only reason a re-run was attempted.

## What is NOT claimed

I did not get a run whose top-level conclusion is `success` with the control applied, and I am not going to manufacture one — the timeout is an independent defect with its own issue. **The claim this PR makes is that the gate executes and can fail in CI, and that is what the first run's log shows.** Anyone who wants a green-conclusion demonstration should fix #1746 first; a second re-run would spend another 15 minutes on the same npx downloads.

Taking it out of draft: the draft status existed solely to stop the control commit merging, and that commit is gone.

VijitSingh97:
## Non-author pass RECORDED — MERGE-READY, with one figure corrected in the body

An ephemeral reviewer session (not the author) reviewed this at `83afe45` and returned PASS with no blocking finding. What it checked ITSELF:

- The `fetch-depth: 0` lands on `lint-surfaces` — the only job whose `run:` step invokes `make lint-file-budget`. It diffed all three occurrences of that key and confirmed the other two are untouched.
- The ratchet is reachable from the gate's exit code, not decorative: `run_gate` calls `check_monotonic || fail=1`. It read the whole `resolve_base_ref` function, not just the changed hunk.
- The remaining fail-open path is real and unchanged by this PR: an unresolvable base still returns 0. It confirmed that by diffing the arm against the base branch's copy — only the NOTE text moved. That is #1608's suggestion 3, filed separately rather than folded in, as this body says.
- Line counts and the tsv row checked against BOTH tips rather than taken from this body.

**One finding, non-blocking, and now fixed:** the Budget section said this script goes 347 -> 352. The measured figure is 347 -> 351. I re-derived that myself rather than relay it, reading the file at both refs through the contents API: base `a3b723bd` is 347 lines, head `83afe451` is 351. The later over-engineering pass cut a line and the Budget section above it was never updated. Corrected in the body. It changes nothing (no row either way, both well under target), but a stale count sitting in a PR about counting lines is worth not shipping.

The reviewer relayed rather than re-ran the control evidence in comment 5535502566, as asked, and says so.

**RELAYED, not re-derived by me** — I am the author. Full review on the box at `~/dev/fleet/answers/reviewer-20260904T045443Z.md`.

MERGE-READY: PASS at 83afe4518e9ef952dddf7be23db649d967546a0d

VijitSingh97:
**Non-author reviewer pass — PASS at `83afe451`**

Merging this. Everything below I ran myself in a detached worktree; where I relayed instead of re-deriving, I say so.

### The defect is real, and I confirmed it from CI's own output rather than from the body

Run `33839336645`, job `Lint per-surface`, 05:14Z today — two consecutive lines:

```
file-budget: NOTE — no base ref (origin/develop or develop) resolvable; skipping the monotonic-ceiling check. This never happens in CI.
file budget OK — no file grew past its ceiling, no new file crossed 800 lines, docs/dev/file-budget.tsv is monotonic.
```

The gate asserts monotonicity in the same breath as the skip that means it never checked it. The body's claim stands exactly as written.

**The thing that kept this hidden is worth recording:** the same job logs `self-test ok: raising a ceiling FAILS (monotonicity, vs the committed base)` about 1.6 s earlier. The self-test builds its own fixture repo, so it has a base ref and goes green. The job therefore showed a passing monotonicity self-test *and* a passing gate while the real invocation skipped. Doubly-green is why nobody looked.

### The fix produces a gate that can actually say no — positive control, both answers

At head `83afe451`, in a full clone where `origin/develop-v2` resolves:

| leg | result |
|---|---|
| gate as-is | `file budget OK`, rc **0** |
| `.github/workflows/ci.yml` ceiling 513 -> 600 | `FAIL — ... raises .github/workflows/ci.yml's ceiling from 513 to 600. Ceilings only go down.` rc **1** |
| reverted | `file budget OK`, rc **0** |

The instrument says both answers, not only the one that reads as success.

### Budget claim re-derived from raw inputs

`ci.yml` is **512** lines on `7442399f`; this diff is `+1 -0`; the recorded ceiling is **513**. 512 + 1 = 513, and the comparison is `-gt`. The body's "AT the ceiling, zero headroom" is correct as stated.

### Merge safety against #1753, which patches the same file from the same base blob

Both patch `ci.yml` from blob `93df1a7b`; this hunk is at ~496, theirs at ~507. I trial-merged rather than reasoning about the distance: `develop-v2` + #1738 + #1753 merges **rc 0, no unmerged paths**, result 513 lines, all three `fetch-depth: 0` lines present. No ordering hazard between the two.

### One consequence the body does not state — non-blocking, but somebody should know

`monotonic_exempt` matches **only** `lib/pithead/99-remainder.sh`; `ci.yml` is not exempt. Measured on the merged tree at 513/513:

| leg | result |
|---|---|
| add one line to `ci.yml`, ceiling left at 513 | `FAIL — .github/workflows/ci.yml is 514 lines, over its recorded ceiling of 513` |
| add one line **and** raise the ceiling to 514 | `FAIL — ... raises ...'s ceiling from 513 to 514. Ceilings only go down.` |
| both reverted | `file budget OK`, rc 0 |

The per-row half needs no base ref, so it already ran in CI. What changes is the second row: **until this PR, "raise the ceiling" was a working escape hatch in CI**, because the check that refuses it was the one being skipped. This PR closes that hatch in the same change that spends the last line of headroom. That is the ratchet working, not a defect — but from the merge on, `ci.yml` cannot gain a line in CI by any legal move except removing one elsewhere in the same file. Worth knowing for whoever next needs a CI step during a GA push.

### What I did NOT do

- I did not verify the `actions/checkout` `dist/index.js` line numbers quoted in the body. I accepted that reading and corroborated the mechanism a different way: two jobs in this same file (lines 50 and 383) already set `fetch-depth: 0` for base-ref and full-history reasons, and `resolve_base_ref` — which I read at source — needs exactly the `refs/remotes/origin/*` such a checkout populates.
- I ran no `make` target, took no lint lock, and did not watch CI to completion.
- My runs were in a full clone; I did **not** reproduce the depth-1 skip locally. The CI log above is my evidence for that half, and it is CI's own.
- I did not review #1701's overlapping change to `scripts/lint-file-budget.sh` beyond confirming its hunk sits in a different arm of `check_monotonic`.

Issue #1734, opened 2026-09-03
Prove behaviourally that read-only verbs take no mutation lock (#1482 residual)


VijitSingh97:
Carried out of #1482, whose own last line has been open since the issue was written and is the only part of it that is not a code change:

> "read-only verbs take nothing" is structurally true but not proven behaviourally by any test.

#1482's code scope is finished — item 1's six verbs landed in #1510, #1528, #1534 and #1538, and item 2 (the control-channel in-place upgrade fallback) in the PR that files this. Every one of those shipped a test proving a mutating verb TAKES the window. Nothing proves the other half: that a read-only verb takes nothing and never waits.

## Why the gap is worth closing rather than arguing

It is structurally true today only because nobody has written `mutation_lock_acquire` into a read-only path. Nothing stops the next one — the acquire is one line, it reads as defensive, and a verb that takes the lock unnecessarily fails in the most expensive direction there is: `status`, `doctor` and `logs` are exactly the commands an operator reaches for **while** a long mutating verb is running, so a stray acquire turns the diagnostic into a 300-second wait during the incident it was meant to diagnose. It would pass every existing test.

`docs/operations.md` already states the property to an operator — "Read-only commands — `status`, `doctor` and `logs` among them — never take it and never wait" — so this is a documented contract with no instrument behind it.

## What would close it

A `tests/stack` case, behavioural rather than a grep for the absence of a call: hold the mutation lock from a second process, run each read-only verb against the held lock with a short `PITHEAD_LOCK_TIMEOUT`, and assert each one **completes** rather than waiting or refusing. The existing lock domains already carry the holder idiom (`cul_hold` / `osl_hold`) and `PITHEAD_LOCK_FILE` to point the window at a sandbox file.

Two things that decide whether it is real coverage:

- **A positive control in the same fixture.** A read-only verb that returns quickly because it failed for its own reasons reads identically to one that correctly never waited. Assert the verb actually produced its output, not just that it exited.
- **An enumeration that cannot silently shrink.** A hand-written list of read-only verbs goes stale the moment one is added. Derive the list from the dispatch case rather than from a literal, or the test passes forever over verbs it never sees.

Note the absence of an acquire is not by itself the property: `doctor` and `status` read `.env` and the compose state, so what is being asserted is that they complete under contention, not that a particular line is missing from a particular function.

## Scope

[...]

PR #1741, opened 2026-09-04
fix(cli): serialise the control-channel in-place upgrade behind the mutation lock (#1482)

VijitSingh97:
Closes item 2 of #1482 — the last unlocked mutating path that issue names. Item 1's six verbs all landed earlier (#1510, #1528, #1534, #1538); this is the residual the issue was deliberately kept open for.

## The defect, measured rather than restated

`control_upgrade`'s in-place fallback replaces the whole install tree — `pithead` itself included — **before** it re-invokes `pithead upgrade`, so the window the re-invoked child takes could never cover the overwrite. It took no window at all.

Base artifact (`git show HEAD:pithead`), with a second process holding the mutation lock, one-click upgrade fired:

| | base | this branch |
|---|---|---|
| result status | `upgraded` | `rejected` |
| install `VERSION` | 1.3.1 → **9.9.9** | 1.3.1, untouched |
| `build/monero/*.template` | replaced by the bundle's | stale copy, untouched |
| `config.json.bak-upgrade-*` | written | none |
| new release's `pithead` | re-invoked | never invoked |

Six rows red on the base, all green here, with the uncontended positive control armed in **both** legs — so the refusal is the lock, not one of this verb's six other `rejected` doors.

## What changed

The fallback moves **verbatim** into a nested `_upg_in_place()` (68 lines, byte-identical modulo a four-space indent — proven by a `diff` of the base slice against the re-indented span, with a shifted-slice negative control that differs) and is called as:

```sh
(
    mutation_lock_acquire upgrade
    _upg_in_place
) || iprc=$?
```

The subshell is the design, not a style choice. `mutation_lock_acquire` **exits** on timeout rather than returning, and this runs inside the long-lived control runner: an unabsorbed exit kills it mid-request and strands the result at the `running` it wrote before the download. The subshell absorbs the exit, the rc routes `PITHEAD_EX_LOCK_TIMEOUT` to `_upg_reject` — `rejected`, never `failed`, the rule #1391's F3 established — and the subshell's own exit closes fd 9, which **is** the release. Same idiom as `(setup)` and `(os_update -y)`.

Any other non-zero rc now reaches `_upg_fail` instead of killing the runner on errexit. That is strictly better than before and is stated rather than left to be found.

**The fresh-dir path is deliberately not wrapped.** Everything it writes is a sibling directory it created itself; the running install is untouched, and the `./pithead upgrade` it runs there takes its own window.

**No self-deadlock, proven not argued:** the re-invoked child inherits the exported `PITHEAD_LOCK_HELD` marker. The fixture's fake release reports back a non-empty marker and a `flock -n` that *fails* from inside the window, so the window is demonstrably still held while the child runs.

The contention message names the ten-minute upgrade throttle, because the attempt still costs it: the dial already happened, so releasing the stamp would reopen the beacon the throttle exists to close.

## Test — `tests/stack/test-control-upgrade-lock.sh`, 17 assertions

Own fixture, own namespace, nothing reused from `test-control-upgrade.sh`; every write lands under its own sandbox dirs. Modelled on `test-appliance-os-update-lock.sh`. A plain (non-versioned) install dir, so the verb takes the in-place fallback rather than the fresh-dir deploy.

**Mutation battery — four single-site mutations of the slice, four killed, no survivors.** Each mutation was asserted to have applied and the file compared against the pristine copy, so a no-op replacement could not report as a mutation.

| Mutation | Rows killed |
|---|---|
| the `mutation_lock_acquire upgrade` is deleted | 10 |
| contention routed to `_upg_fail` instead of `_upg_reject` | 1 |
| the `PITHEAD_EX_LOCK_TIMEOUT` comparison never matches | 4 |
| the subshell removed, so the acquire's exit is not absorbed | 5 |

The kill sets are disjoint enough to discriminate: the `_upg_fail` mutation kills **only** `rejected, not failed`, which is the row that exists to keep #1391's F3 distinction alive.

One assertion was **rewritten after the battery caught it as unfailable**: `the control runner survived to write a result` checked only that a result file existed, which is already true from the `running` write before the download — it could not fail for the reason it named. It is now `the result is not left stranded at running`, and the no-subshell mutation kills it (4 rows → 5).

## Lane discipline

Two per-item grants from the controller (w101), **not** the `.lane-override` hatch — a hatch admits every out-of-lane file in the commit, a named line admits one:

- `tests/stack/test-control-upgrade-lock.sh` — fleet `3bb5114`
- `tests/stack/run.sh` — fleet `ab9d66b`, for **one pure insertion after line 23** and no other line. The hunk is exactly that: `3 insertions(+), 0 deletions(-)`, `git diff --numstat` = `3 0`. It shifts everything below by three lines; the appliance lane's live region is `:236-255`, ~210 lines away.

Arming readback, both directions:

- `tests/stack/test-control-upgrade-lock.sh` staged **alone** → lane-guard rc **0**
- `tests/stack/run.sh` staged **alone** → lane-guard rc **0**
- an ungranted `tests/stack/` sibling staged **alone** → lane-guard rc **1**, naming the file (negative control fired; the scratch file was removed in the same call)

`docs/operations.md` is in `_shared.paths`, no hatch needed — disclosed here. One sentence added to the lock section. It makes an **existing** sentence in that file true: the paragraph below it already claimed the one-click upgrade shares the lock, which was false for the in-place fallback until this change.

## Gates run

`bash -n`, `shfmt -i 4 -d`, `shellcheck --severity=warning` (**0.11.0**, the pinned build, on both changed shell files and the new test), `make lint-pithead-parity`, `make lint-docs-voice`, `make lint-topology`, `make lint-operator-strings`, `make lint-file-budget` — rc 0 each, captured directly rather than off a pipe. `make lint-sh` not run here: it is the serialized OOM path on this box.

Slice 44 is 392 lines and the new test 175, both under the 400 target, so neither takes a `file-budget.tsv` row.

Full `bash tests/stack/run.sh` on this branch, run serially with nothing else touching the fixtures: **3192 passed, 0 failed**, the new domain among them.

`tests/stack/test-control-upgrade.sh`: **142 passed / 0 failed on the base artifact and on this one** — a controlled pair, same assertion count, no regression. The in-place `_upg_fail` branches still report from inside the subshell (`failed upgrade run reports failed`, `failed in-place upgrade names the pre-upgrade copies (#637)`, `failed snapshot fails the upgrade (#637)` all green).

## Over-engineering pass

Run by hand on my own diff, as #1391's was. Four findings, and I am naming the two I did not act on rather than leaving them to be found.

1. **The nested `_upg_in_place()` could have been a bare `( ... )` around the existing lines.** Kept as a function: the moved span contains eight `return 0` statements, and `return` inside a bare subshell inside a function is ambiguous enough that the next reader would have to check it. The file already nests `_upg_reject`, `_upg_fail` and `_upg_snapshot`, so this is its own idiom, not a new one.
2. **The `iprc -ne 0` branch looks like scope creep and is not.** Without it the subshell would *downgrade* the old failure: before, an errexit death inside the in-place path killed the runner loudly; after, it would be swallowed and leave the result stranded at `running` with no verdict at all. The branch is required by the change, not added beside it.
3. **Two test rows are one measurement, and the body should not count them as two.** `the install dir was not overwritten` (VERSION) and `the new bundle's build/* did not land` are two members of the *same* tar extraction — one event read twice, not two independent proofs. Both kept: the `build/*` row also stands on the #544/#555 merge-behaviour shape that the sibling test exists for. But the contended evidence here is four independent facts, not six: the extraction, the re-invocation, the `#637` snapshot, and the result verdict.
4. **`rm -f "$bundle"` is duplicated across the two routing branches and could hoist above the `if`.** Not acted on: it is one line, behaviour-neutral, and taking it would invalidate the exact artifact every measurement in this PR was made against — another build, parity, mutation battery and full-suite cycle to re-prove one line, at the end of a push window. Stated so the next reader knows it was seen and priced, not missed.

## What I did NOT do

- The issue's trailing note — *"read-only verbs take nothing" is structurally true but not proven behaviourally by any test* — is **not** in this PR. It is a pure `tests/stack` item with no CLI change behind it. Filed as #1734 under `maintenance & gates` rather than left in a message.
- `make lint-sh` (serialized, OOM path).
- My first full `tests/stack/run.sh` died at `run.sh: line 112: d: unbound variable` while my own probe scripts were driving the same fixtures concurrently. That is the cross-talk shape COMMON tells us to read as cross-talk rather than as a defect, not a finding against this branch; the clean serial run is the one reported.

https://claude.ai/code/session_01LDtwKBWTZnwJ7ewoxM8d6F

### DIFF
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -112,8 +112,11 @@ takes it after its prompts, so a passphrase you have not typed yet holds nothing
 first-boot wizard holds it only while it deploys. `os-update` takes it just before it writes the
 spare slot, so a confirmation it is still waiting on holds nothing up either, and
 `rotate-secrets` takes it after its own confirmation and before it writes the copies of your old
-secrets, so a rotation that ends up waiting has not yet changed a credential. Read-only
-commands — `status`, `doctor` and `logs` among them — never take it and never wait.
+secrets, so a rotation that ends up waiting has not yet changed a credential. A one-click
+upgrade from the dashboard takes it before it writes over the install, so one that arrives while
+another command is running comes back refused with nothing changed, instead of overwriting the
+install underneath it. Read-only commands — `status`, `doctor` and `logs` among them — never take
+it and never wait.
 
 The lock covers one stack, not one directory. A bundle install keeps each release in its own
 `pithead-vX.Y.Z` directory beside the one before it (see [The deploy-box
--- a/lib/pithead/44-control-upgrade-and-lifecycle.sh
+++ b/lib/pithead/44-control-upgrade-and-lifecycle.sh
@@ -253,74 +253,109 @@ control_upgrade() { # <request-file> <id> <actor> <control-dir>
         rm -f "$logf"
         return 0
     fi
-    # #637: unlike the fresh-dir path, this one has no surviving previous dir — and the new
-    # release's `upgrade` below re-renders .env (and may migrate config.json). Keep a timestamped
-    # pre-upgrade copy of both next to the originals before a byte changes (cp -p keeps .env's
-    # 0600; a fresh stamp per attempt so a failed try never overwrites the good copy) and refuse
-    # to overwrite the install without one. The destination name is predictable, so root must
-    # never write through a symlink a co-tenant planted there (the #629 mkdir guard's attack
-    # class): copy to an unpredictable mktemp name first, then rename onto the final name —
-    # rename(2) replaces a planted entry without following it.
-    _upg_snapshot() { # <src> <dst>
-        local tmp
-        tmp=$(mktemp "$PWD/.bak-upgrade.XXXXXX") || return 1
-        if ! cp -p "$1" "$tmp"; then
-            rm -f "$tmp"
-            return 1
+    # #1482: everything below overwrites the RUNNING install — the two tar passes replace the whole
+    # tree, this script included — and it does so BEFORE the re-invocation, so the lock the
+    # re-invoked `pithead upgrade` takes cannot cover the overwrite. The window is held from the
+    # first snapshot write to the end of that re-invocation. The fresh-dir path above needs none of
+    # it: everything it writes is a sibling directory it created itself, and the `./pithead upgrade`
+    # it runs there takes its own window.
+    #
+    # Run in a subshell, and that is the design rather than a style choice. `mutation_lock_acquire`
+    # EXITS on timeout (PITHEAD_EX_LOCK_TIMEOUT) instead of returning, and this code runs inside the
+    # long-lived control runner: an exit here would kill the runner mid-request, leaving this result
+    # stuck at "running" and the claim never released. The subshell absorbs that exit so the status
+    # can be ROUTED — contention is "rejected", never a failed upgrade, for the same reason
+    # wizard_setup_failed refuses to call a timeout a bad config (#1391 F3). Its exit also closes
+    # fd 9, which IS the release: the `(setup)` idiom, one caller over. The re-invoked child
+    # inherits the exported PITHEAD_LOCK_HELD marker, so it does not block on its parent's hold.
+    _upg_in_place() {
+        # #637: unlike the fresh-dir path, this one has no surviving previous dir — and the new
+        # release's `upgrade` below re-renders .env (and may migrate config.json). Keep a timestamped
+        # pre-upgrade copy of both next to the originals before a byte changes (cp -p keeps .env's
+        # 0600; a fresh stamp per attempt so a failed try never overwrites the good copy) and refuse
+        # to overwrite the install without one. The destination name is predictable, so root must
+        # never write through a symlink a co-tenant planted there (the #629 mkdir guard's attack
+        # class): copy to an unpredictable mktemp name first, then rename onto the final name —
+        # rename(2) replaces a planted entry without following it.
+        _upg_snapshot() { # <src> <dst>
+            local tmp
+            tmp=$(mktemp "$PWD/.bak-upgrade.XXXXXX") || return 1
+            if ! cp -p "$1" "$tmp"; then
+                rm -f "$tmp"
+                return 1
+            fi
+            mv -f "$tmp" "$2"
+        }
+        local bak
+        bak="bak-upgrade-$(date +%Y%m%d-%H%M%S)"
+        if ! _upg_snapshot "$CONFIG_FILE" "$CONFIG_FILE.$bak" 2>"$logf" ||
+            ! _upg_snapshot "$ENV_FILE" "$ENV_FILE.$bak" 2>>"$logf"; then
+            rm -f "$bundle"
+            _upg_fail "could not keep a pre-upgrade copy of config.json/.env: $(tail -c 500 "$logf") — refusing to overwrite the install without a restore point; the stack keeps running v$PITHEAD_VERSION."
+            rm -f "$logf"
+            return 0
         fi
-        mv -f "$tmp" "$2"
+        # Older snapshots hold yesterday's secrets: keep the newest three pairs, prune the rest.
+        # Lexical order IS chronological for this stamp format; `|| true` — an empty glob under
+        # pipefail must not kill the runner.
+        ls -1r "$CONFIG_FILE".bak-upgrade-* 2>/dev/null | tail -n +4 | while IFS= read -r f; do rm -f "$f"; done || true
+        ls -1r "$ENV_FILE".bak-upgrade-* 2>/dev/null | tail -n +4 | while IFS= read -r f; do rm -f "$f"; done || true
+        # The reported paths: CONFIG_FILE is cwd-relative unless the (test-only) override made it
+        # absolute — don't prepend $PWD onto an already-absolute path.
+        local bak_paths
+        case "$CONFIG_FILE" in
+        /*) bak_paths="$CONFIG_FILE.$bak" ;;
+        *) bak_paths="$PWD/$CONFIG_FILE.$bak" ;;
+        esac
+        bak_paths="$bak_paths $PWD/$ENV_FILE.$bak"
+        # In-place extraction over the running install, in two passes. Pass 1 lays down everything
+        # EXCEPT the running script with plain tar, which MERGES existing directories — a release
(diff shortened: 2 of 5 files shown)


VijitSingh97:
## Adversarial pass — **PASS**. One body-accuracy finding, non-blocking; nothing blocks merge.

Reviewed at `b56a0147`, base `develop-v2` (merge-base `5ff0cd48`).

### The controller's named question, measured: yes, the over-count repeats — and the honest number is **two**, not four

Over-engineering pass #3 caught one pair (VERSION and `build/*` are one tar extraction read twice) and stopped one step short of the general shape. Two measurements, both mine:

**(a) Deleting the lock.** Controlled pair in one clone lineage, the new test file held byte-constant, only the built `pithead` swapped to the base artifact:

| | head `b56a0147` | base artifact |
|---|---|---|
| `test-control-upgrade-lock.sh` | **17 passed / 0 failed** | **7 passed / 10 failed** |

Eight of the ten contended rows go red **together**.

**(b) Moving the lock instead of deleting it.** Single-site mutation of the built artifact — the `mutation_lock_acquire upgrade` lifted out of the subshell and dropped inside `_upg_in_place` immediately after the snapshot block, before the tar. Asserted applied (`--numstat` `1 1`, placement confirmed by `grep`, `bash -n` rc 0), so a no-op edit could not report as a mutation. Result:

> **16 passed / 1 failed** — and the one red row is `contended: no pre-upgrade snapshot was written either`.

`contended: the new release's pithead was never re-invoked`, `the install dir was not overwritten` and `the new bundle's build/* did not land` all stayed **green** while a `config.json.bak-upgrade-*` had already been written under contention.

So the contended block's rows are not four facts:

- **Rows 5–8 are one causal chain read at four points.** `_upg_in_place` is strictly sequential — snapshot, then tar, then re-invoke — with the acquire above all of it, so no change to the lock's *presence* can make them disagree. Only a change to its *position* can, and measurement (b) shows the earliest probe is the only one that sees it. **Row 8 strictly dominates rows 5, 6 and 7**: every defect that reds one of those also reds row 8, and there is a real defect that reds row 8 alone.
- **Rows 2, 3 and 4 are three substrings of one emitted literal.** On the base they print `[null]` three times — one absent `.error` field read three times. With row 1 they are a single fact: the `_upg_reject` at the timeout branch fired. The battery's own `_upg_fail`-routing mutation killing exactly **1** row is the same measurement from the other side.
- **Rows 9 and 10 pass on the base *and* on head.** They guard the regression this PR could introduce (the subshell absorbing the acquire's exit), not the defect it fixes. Legitimate rows; not evidence for the fix.

**Measured count: two independent facts** — (i) the window is taken before the first write, witnessed solely by row 8; (ii) contention routes to `rejected` rather than `failed`, witnessed by row 1. The 6→4 correction in the body was a real improvement and the residual costs nothing but body accuracy — **the suite is not weakened, and this does not block.**

### One note that is worth more than the count

**Row 8 is a single point of failure for the placement property.** It is the *only* assertion in the file that can distinguish "the window is taken before the first write" from "taken somewhere inside `_upg_in_place`" — measurement (b) is exactly that. The four-row chain reads as redundancy and is not: if a future refactor of `_upg_in_place` drops or weakens that row, the suite goes green across the whole "lock taken too late" class, which is the more likely future regression here. Worth a sentence beside the assertion, not a change to it.

### Proven by me, each with a control

1. **The 68-line move is verbatim modulo the four-space indent.** Base `44-*.sh:256-323` vs head `:272-339` de-indented: 68 lines each, **sha256 identical** (`500f352c…`). Every head line in the span carried the indent (0 exceptions outside blanks). Negative control fired — the same comparison against a one-line-shifted slice reports a difference, so it was shown able to say the other thing. This is the claim a 95/60 diff would otherwise hide, and it holds.
2. **Build parity for the changed slice**: the head `44-*.sh` appears verbatim inside the head `pithead`. The `--numstat` gap (slice `95/60`, `pithead` `97/62`) is diff-hunk alignment, not content — the two extra lines each side are a `return 0`/`fi` pair the slice diff attributed to context.
3. **The defect is real**, not restated: on the base artifact, under a held lock, VERSION goes 1.3.1 → 9.9.9, `build/monero/bitmonero.conf.template` is replaced by the bundle's, a `config.json.bak-upgrade-*` is written and the new release's `pithead` is re-invoked. The base leg also reds the *uncontended* `window was still held while it ran` row (`marker=UNSET`, `window=FREE`), which is the marker half.
4. **No self-deadlock, re-derived at source rather than relayed.** The child hits `[ -n "${PITHEAD_LOCK_HELD:-}" ]` at `00-prelude.sh:86` and returns at `:88` without opening a second descriptor. The fixture measures it live from inside the window.
5. **The attack that would have made this a silent no-op does not reach.** If the control runner could inherit a *stale* exported `PITHEAD_LOCK_HELD`, the new acquire would return 0 immediately and serialise nothing. It cannot: the marker is `unset` on orderly release (`00-prelude.sh:311`), and the runner is started by systemd — `ExecStart=$PWD/pithead control-run-pending`, `50-control-runner-provisioning.sh:121` — so it never inherits a live pithead's environment. Worth stating because the guard's soundness rests on it and nothing in the PR says so.
6. **The `tests/stack/run.sh` hunk is a pure insertion**: `--numstat` `3 0`, zero deletions, the exact `_d0` + `# shellcheck source=` idiom of its neighbours.
7. The positive control varies three things at once (lock path, holder, timeout) rather than one. **Not a finding** — I traced both extras and neither can produce `rejected`: an unopenable lock path fails *open* (`00-prelude.sh:110-117`, returns 0 → `upgraded`, the opposite outcome), and a 1s timeout with no holder never waits. Sound, though not minimal.

### Assumed — I did not verify these

- **The author's 4/4 mutation battery.** I ran a fifth, different mutation instead of re-running theirs.
- **The `3192 / 0` full-suite run and the `142 / 0` `test-control-upgrade.sh` controlled pair.** No `make` target run at all, `make lint-sh` included (serialised OOM path).
- **The lane-guard arming readbacks** (rc 0 / 0 / 1) — relayed from the body.
- **No bench, no rig, no container, no real docker, no systemd.** Every run above drives the real `pithead` under the suite's `write_fake_docker` and `curl` stubs, so it exercises control flow, tar, flock and jq faithfully and says nothing about the systemd unit or a real upgrade.
- CI not watched to completion.

Issue #1730, opened 2026-09-03
Nothing checks render_masked_config's CONTROL_SECRET_PATHS against selftest-redact.sh's classification, and they have already disagreed


VijitSingh97:
The stack classifies `config.json` by PATH, in `CONTROL_SECRET_PATHS`
(`lib/pithead/30-release-fetch-and-masked-config.sh:5`). `tests/integration/selftest-redact.sh`
classifies the same document by a NAME screen over `config.reference.json`. Nothing checks the two
against each other, and they have already disagreed.

Measured on `develop-v2` before #1723's fix, one line per path in `CONTROL_SECRET_PATHS`:

```
  OK   dashboard.auth.password        OK   p2pool.stratum_password
  OK   telegram.bot_token             OK   healthchecks.ping_url
  OK   workers.api_token              OK   notifications.ntfy.url
  OK   monero.node_username           OK   notifications.ntfy.token
  OK   monero.node_password           !!   xvb.standby.source   <- UNCLASSIFIED
  OK   monero.view_key
  OK   tari.view_key
```

`xvb.standby.source` is masked by the product and was in none of the self-test's lists, because no
suffix in either vocabulary carries the bare word `source`. #1723 classifies it and widens the
screen to reach it — by hand, which is the problem: the next disagreement will be found the same
way, by someone happening to look.

## What to build

An assertion that every `CONTROL_SECRET_PATHS` entry appears in the self-test's classification.
The list is a JSON literal in a shell slice, so it parses without a shell parse — the same shape
as the vocabulary invariant already in `selftest-redact.sh`, which reads redact()'s two suffix
lists out of `lib.sh` rather than restating them.

Note the direction. This is a one-way containment check, not equality: the self-test classifies
more than the product masks (public endpoints, routing ids), and it should. What must not happen
is the product calling something a secret that the self-test has never heard of.

## Where it goes

Not in `selftest-redact.sh` — that file sits at 385 lines against `lint-file-budget`'s 400-line
target, and the established practice in this directory is to split rather than raise
(`selftest-redact-vocab.sh`, `-env.sh`, `-ip.sh` all exist for that reason). Under `TARGET_LINES`
a new `selftest-*.sh` needs no budget row and no registration — the Makefile globs it.

`selftest-redact-vocab.sh` is the closest sibling in spirit: it asks whether the vocabulary reaches
every key that needs it, one level out from whether a given shape is covered. This is the same
question against the other classification source.

[...]

PR #1742, opened 2026-09-04
test(integration): #1730 every path render_masked_config masks is held against selftest-redact.sh's classification, so the next drift reds instead of waiting to be noticed

VijitSingh97:
## The defect

The stack classifies `config.json` by PATH — `CONTROL_SECRET_PATHS` plus three jq stanzas for the variable-length arrays. `selftest-redact.sh` classifies the same document by a name screen over `config.reference.json`. Two hand-kept classifications of one document, and nothing held them against each other, so they drifted: `xvb.standby.source` is masked by the product and was in none of the self-test's lists, because no suffix in either vocabulary carries the bare word `source`. It was found at #1723 by someone happening to look. This is the check that fires instead of the next person looking.

## What this adds

`tests/integration/selftest-redact-paths.sh` (new, 253 lines — under `TARGET_LINES=400`, so no budget row, and `Makefile:29` globs `selftest*.sh` so no registration).

**The containment is one-way, deliberately: product ⊆ self-test, never equality.** The self-test classifies more than the product masks — public endpoints, routing ids — and it should; those are what a bundle exists to carry. What must not happen is the product calling something a secret the self-test has never heard of, because such a field sits outside the drift alarm entirely.

**The population is measured, not parsed.** A read of the `CONTROL_SECRET_PATHS` literal sees 12 fixed paths and misses all three array stanzas (`workers.list[].token`, the deprecated `dashboard.workers[].token`, `notifications.webhooks[]`). So the file runs the REAL masker over a populated fixture in a sandbox box and reads back the 15 paths that became `{"__secret__": true}`. The four classification lists are likewise read out of `selftest-redact.sh` rather than restated — restating them is what drifts. A path inside an array is satisfied by its enclosing `[]` path being classified, which is what `workers.list[].token` needs and precisely what `ELEMENT_SHAPE_UNKNOWN` means.

## ⛔ This is STACKED on #1728 and must not merge before it

Base is `fix/1723-redact-selftest-array-blind`, not `develop-v2`, and that is measured rather than preferred: `git diff --stat origin/develop-v2 HEAD` on a develop-v2 base shows **9 files** including `scripts/build-pithead.sh`, `.lycheeignore` and `lychee.yml` — this branch is merely BEHIND develop-v2 by #1727's merge, so a develop-v2-based PR would propose reverting it. The diff against the stack base is the 2 files here.

`ELEMENT_SHAPE_UNKNOWN` exists only on #1728's branch (4 occurrences there, 0 on `develop-v2`). Run against `develop-v2` today this file fails loudly on the list parse rather than passing quietly — the correct direction, but it means **#1728 merges first**, and this needs `rebase --onto` afterwards because #1728 will be squashed. I will re-prove by patch bytes after that rebase, not by shas.

## Proof — a six-arm mutation battery, each arm attacking a DIFFERENT mechanism

Every arm verified it mutated (sha before/after, anchor required to match exactly once) and every file was restored and re-baselined. Baseline 20 passed / 0 failed; post-battery baseline identical.

| arm | what it attacks | result |
|---|---|---|
| A1 | unclassify `xvb.standby.source` — the drift already found | 19/**1**, only that row |
| A2 | drop `workers.list[]` — the enclosing-`[]` clause | 19/**1**, only `workers.list[].token` |
| A3 | product masks a path nothing classifies — **the NEXT disagreement** | 19/**2**, unclassified + set moved |
| A4 | masker jq fails and returns 0 — the false-green | **2/1**, dies at the arming control |
| A5 | product masks one fewer | 18/**1**, the set row alone |
| A6 | schema drops `xvb.standby`, fixture resurrects it | 19/**1**, the schema-presence row |

Distinct signatures, so this is per-arm attribution rather than one mutation reddening everything.

**A4 and A6 are the two that changed the file.** `render_masked_config` warns and returns 0 when its jq fails, so a test reading its rc greens over a document that was never written — the arming control asserts the artifact and its exact path SET before any verdict is read. And A6 was found by attacking my own first draft: a jq assignment CREATES an absent path, so with `xvb.standby` deleted from the schema the first version still passed 19/19, certifying a path the product no longer carries. It now asserts each populated leaf against the shipped schema first. A6 is a controlled before/after on that same defect — it passed before the fix and reds after.

## What was run

- The new file: 20 passed / 0 failed. Full `tests/integration/selftest*.sh` glob: **24 files, all green**, no sibling disturbed.
- `shellcheck` **0.11.0** (the `Makefile:52` pin) and `shfmt -i 4 -d`: both clean, rc 0.
- `make lint-md`, `make lint-docs-voice`, `make lint-file-budget` (run AFTER `git add` — the ratchet is blind to untracked files, #1486): all clean.
- `make lint-pithead-parity`: clean. This matters here because the test sources the BUILT `pithead`; parity is what makes that equivalent to sourcing the slices.
- Verified cwd-independent (run from `/tmp`).

## Findings I considered and did NOT act on, with reasons

- **`source "$BOX/pithead" >/dev/null 2>&1` hides stderr**, which is the shape that returns empty and reads clean. Kept, because the prescribed mitigation is present and load-bearing: `declare -F render_masked_config` is asserted immediately after, and A4 shows the file dies rather than greening when the product misbehaves.
- **`EXPECTED=` relies on deliberate word-splitting** over `FIXTURE_PATHS`. Kept with a scoped `# shellcheck disable=SC2086` and a stated reason; the alternative is an array that `assert_eq` cannot compare as cleanly.
- **The set-equality row prints two long strings on failure.** Noisier than a count, and kept anyway: a count cannot name WHICH path appeared or vanished, and A3/A5 are exactly the cases where that name is the finding.
- **`while read -r verdict path list entry`** would break on a path containing whitespace. Not defended against — config paths are dotted identifiers and cannot contain any.

## /ponytail-review — over-engineering pass, run off disk on this diff

1. **Acted on.** The enclosing-`[]` lookup scanned every character prefix of a path when only a classified entry ending in `[]` can enclose anything. Replaced with a direct scan of the `[]`-suffixed entries, longest match winning. Behaviour preserved: every arm of the six-arm battery reproduces its exact signature before and after, and the baseline is 20/0 on both sides.
2. **Declined** — the 45-line header. It matches the sibling redact self-tests' practice, and each block is a trap that cost an attempt (the masker's rc, the array element types, the piped `source`) rather than a restatement of the code beneath it.
3. **Declined** — collapsing the two python blocks into one. The schema-presence guard has to run BEFORE the fixture is rendered and the walk necessarily runs after the masker; merging them would move the presence check downstream of the very thing it exists to gate.
4. **Declined** — a count instead of a set in the arming control. A3 and A5 are exactly the cases where the failing row has to name WHICH path moved, and a count cannot.

## Disclosure

`docs/dev/integration-testing.md` is under `_shared.paths`; the change is one paragraph joining the existing redact-selftest enumeration, in the house voice, plus one sentence corrected after the A6 fix made it stale.

Refs #1730 (`Closes` cannot fire — this targets `develop-v2` while the default branch is `develop`; closing by hand at merge).

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -798,6 +798,28 @@ as survivors. A bare `AUTH` suffix would redact the pair and protect nothing, an
 would take every public endpoint while still missing the webhook key, which ends in `URLS`. Four
 onion addresses are listed apart again, because a shape rule rather than the name rule covers them
 and a name-shaped fixture would report them as leaking.
+`selftest-redact-paths.sh` holds the other classification against this one. The stack masks
+`config.json` by PATH, in `CONTROL_SECRET_PATHS` plus three jq stanzas for the variable-length
+arrays, while `selftest-redact.sh` classifies the same document by a name screen. Two hand-kept
+classifications of one document, and until [#1730](https://github.com/p2pool-starter-stack/pithead/issues/1730) nothing held them against each
+other, so they drifted: `xvb.standby.source` is masked by the product and was in none of the
+self-test's lists, because no suffix in either vocabulary carries the bare word `source`. It was
+found by someone happening to look, which is the thing this file replaces. The check is one-way
+by design — every path the product masks must be classified, never the reverse, because the
+self-test classifies more than the product masks and should: the public endpoints and routing
+ids are what a bundle exists to carry. The population is measured rather than parsed, since a
+read of the `CONTROL_SECRET_PATHS` literal sees twelve fixed paths and misses all three array
+stanzas; the file runs the real masker over a populated fixture and reads back the fifteen paths
+that became `{"__secret__": true}`. A path inside an array is satisfied by its enclosing `[]`
+path being classified, which is what `workers.list[].token` needs. The arming control is the
+load-bearing part: `render_masked_config` warns and returns 0 when its jq fails, so a test
+reading its return code greens over a document that was never written and every containment row
+then passes over an empty population. The file asserts the artifact exists and holds exactly
+the set of paths the fixture populated, named entry for entry, before it reads a single
+verdict. One more thing a populated fixture cannot tell you on its own: a jq assignment
+creates an absent path, so the presence of each populated leaf is checked against the shipped
+schema first — without that row, deleting `xvb.standby` from the reference leaves every other
+row green over a path the product no longer carries.
 `selftest-zmq-probe.sh` drives the ZMTP verdicts from captured and hand-built wire fixtures,
 so every failure class — a silent peer, a non-ZMQ listener, a ZMTP peer that is not a publisher, a
 READY frame carrying a decoy `Socket-Type` value — is reachable with no socket and no stack. It
--- a/tests/integration/selftest-redact-paths.sh
+++ b/tests/integration/selftest-redact-paths.sh
@@ -1,0 +1,253 @@
+#!/usr/bin/env bash
+#
+# Self-test: every path `render_masked_config` masks is CLASSIFIED by selftest-redact.sh (#1730).
+#
+# WHY THIS EXISTS. The stack classifies config.json by PATH, in CONTROL_SECRET_PATHS
+# (30-release-fetch-and-masked-config.sh) plus three jq array stanzas. selftest-redact.sh
+# classifies the same document by a NAME screen over config.reference.json. Two classifications of
+# one document, kept by hand, and nothing held them against each other — so they drifted, and the
+# drift was found by someone happening to look: `xvb.standby.source` is masked by the product and
+# was in none of the self-test's lists, because no suffix in either vocabulary carries the bare
+# word `source` (#1723). This file is the check that fires instead of the next person looking.
+#
+# THE DIRECTION IS ONE-WAY, AND THAT IS DELIBERATE. This asserts product ⊆ self-test, never
+# equality. The self-test classifies MORE than the product masks — public endpoints, routing ids —
+# and it should; those are the rows a bundle exists to carry. What must not happen is the product
+# calling something a secret that the self-test has never heard of, because such a field is
+# outside the drift alarm entirely: nothing reds when it changes.
+#
+# THE POPULATION IS MEASURED, NOT LISTED. A parse of the CONTROL_SECRET_PATHS literal would see 12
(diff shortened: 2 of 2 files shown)


VijitSingh97:
**Non-author review — PASS at `aeef7d83`.**

Independently re-derived, not relayed:

- Checked out the exact head in an isolated scratch clone and ran `tests/integration/selftest-redact-paths.sh` unmodified: **20 passed / 0 failed**, matching the body.
- Confirmed `CONTROL_SECRET_PATHS` and `KNOWN_GAP` already carry `xvb.standby.source` at this head — the rebase onto `develop-v2` did land #1728's fix, not just the claim of it.
- Reproduced three of the six mutation arms myself, each against a rebuilt `pithead` where relevant, signatures matched exactly:
  - **A1** (reintroduce the #1723 drift — drop `xvb.standby.source` from `KNOWN_GAP`): **19/1**, only that row fails. The test catches the exact defect it's named for.
  - **A4** (masker's jq fails, `render_masked_config` still returns 0): **2/1**, dies at "the masker wrote a masked document" — the rc-discarding arming control is real, not decorative.
  - **A6** (schema drops `xvb.standby`, fixture would otherwise resurrect it via jq's create-on-assign): **19/1**, dies at the schema-presence row before any containment row is trusted.
- Confirmed the one-way containment direction: population is 15 paths (12 `CONTROL_SECRET_PATHS` + 3 array stanzas), matches `FIXTURE_PATHS` exactly.
- Confirmed `Makefile:29`'s `test-integration-selftest` and `.github/workflows/ci.yml:474` both glob `tests/integration/selftest*.sh` — the new file registers with zero wiring changes, as claimed.
- Confirmed no file-budget row is owed (253 lines, `TARGET_LINES=400`).
- No wallet/hostname/credential material in the diff — fixture values are all synthetic `fixture-*` strings, `config.reference.json` untouched.
- Docs addition to `docs/dev/integration-testing.md` reads consistent with the surrounding voice.

Not verified by me: the other three mutation arms (A2, A3, A5), `shellcheck`/`shfmt` (CI's "Shell tests" check already covers this at head), and the full 24-file `selftest*.sh` glob run (CI's own run covers it; redundant to repeat here).

MERGE-READY: PASS at aeef7d83

Issue #1726, opened 2026-09-03
The link check cannot stay green while gnu.org rate-limits it — the gpl-3.0.txt link needs the torproject treatment


VijitSingh97:
## What is wrong

`THIRD_PARTY_LICENSES.md` links `https://www.gnu.org/licenses/gpl-3.0.txt`. On two dispatched `lychee.yml` runs four minutes apart on 2026-09-03 the same link failed both times, differently: `[TIMEOUT]` (run 33816940119) then `[429] Too Many Requests` (run 33817230727). From a Linux box outside GitHub's runners `curl --max-time 20` on the same URL times out. Every other link on the branch passed (833 checked, 830 ok, 2 excluded).

## Why it matters

This is #1419's shape with a different host. `.github/workflows/lychee.yml` runs on `schedule` only, from `develop`, with `fail: true`; a host that rate-limits non-browser clients reds the weekly run with nothing anyone can fix in this repo, and a red that is always red carries no information. The `develop` dispatch happened to pass on 2026-09-03 (run 33817233000); that is one sample, and the twin failed 2 of 2.

## Suggested fix

The same treatment `.lycheeignore` already gives `gitlab.torproject.org` (a 403 to non-browser clients, verified by hand): one anchored host pattern with the reason and this number, on both twins. Not a global `--accept 429`, which would mask real rate-limited dead links elsewhere. The link is valid by hand — it is the GPL-3.0 text — and the file is a license notice, so the reference must stay.

Found during the non-author pass on #1719/#1720, which fix the vendored-copy noise and the ADR link and are not widened to carry this.

VijitSingh97:
Fixed by PR #1743, squash-merged to `develop-v2` as `ba0cccfdc7567d95b836ef059246ccaf21d6ebd9`. Hand-closing: `Closes` keywords are inert against `develop-v2`.

Issue #415, opened 2026-07-10
CI + pre-commit hygiene: cancel stale PR runs, add job timeouts, fix the red weekly lychee, stop the ruff pin drifting


VijitSingh97:
Six small, verified drifts in the CI/pre-commit wiring. None justifies its own issue; together
they are one cleanup PR.

1. **Stale PR runs are never cancelled.** `ci.yml` and `integration-mini-stack.yml` have no
   `concurrency` block, so every push to a PR queues a fresh set of up to 16 jobs while the
   previous set keeps running. With several agents pushing fixups in parallel (20+ PRs merged
   this week), obsolete runs contend with live ones for runners. (`release-gate.yml` already has
   its own `release-gate` group — correct, leave it.)
2. **No `timeout-minutes` anywhere.** A hung docker build or wait loop holds a runner for
   GitHub's 6-hour default. Normal CI wall clock is ~90 seconds, so a tight ceiling is free.
3. **The weekly lychee run is red and going nowhere.** The last two scheduled runs
   (2026-06-29, 2026-07-06) failed on one false positive:
   `https://gitlab.torproject.org/tpo/core/tor` in `THIRD_PARTY_LICENSES.md` returns 403 to
   non-browser clients. A check that stays red on a known false positive stops being read.
4. **The ruff pin has drifted, exactly as the config comment warned.**
   `.pre-commit-config.yaml:8` pins `ruff-pre-commit` at `v0.15.17`;
   `build/dashboard/pyproject.toml:41` pins `ruff==0.15.20`. Dependabot's `uv` ecosystem bumps
   the pyproject side, but Dependabot has no pre-commit ecosystem — this drifts on every ruff
   release. Local commits and CI lint with different ruff versions.
5. **`lint-docs-voice` is the only lint surface with no pre-commit hook** (compare
   `Makefile:44` against the local hooks in `.pre-commit-config.yaml:38-75`), so a banned word
   is only caught in CI.
6. **Dead issue pointer:** `build/dashboard/pyproject.toml:78` says annotation rules "are
   deferred to #284" — #284 is closed, and the ANN/`ty` on-ramp it described was never built
   (27 of 285 dashboard `def`s carry a return annotation today).

## Proposal

One PR, six mechanical fixes. Decisions:

- Concurrency: `group: ${{ github.workflow }}-${{ github.ref }}` with
  `cancel-in-progress: ${{ github.event_name == 'pull_request' }}` — PR pushes cancel their
[...]

PR #1743, opened 2026-09-04
fix(ci): #1726 the weekly link check stops going red on a host that rate-limits it

VijitSingh97:
## The finding

`lychee.yml` runs on `schedule` only, from the default branch, with `fail: true`. So a host that rate-limits non-browser clients reds the weekly run permanently, with nothing anyone in this repo can fix — and **a gate that is always red has no transition left to make, so it cannot signal a new break.** That is #1419's second-order finding, arriving again three weeks later with a different host. #1419 was worth fixing because the link check had been red for nine consecutive Mondays; leaving this one in place walks it straight back there.

## Measured (from #1726, re-checked here)

`THIRD_PARTY_LICENSES.md:20` cites `<https://www.gnu.org/licenses/gpl-3.0.txt>`. On 2026-09-03 that one link failed **two different ways four minutes apart** on dispatched runs — `[TIMEOUT]` on `33816940119`, then `[429] Too Many Requests` on `33817230727` — and it also times out from a Linux host outside GitHub's runners. Every other link on the branch passed (833 checked, 830 ok, 2 excluded). The `develop` dispatch happened to pass once (`33817233000`); that is one sample against the twin's 2-of-2 failure, which is why "it passed once" is not treated as evidence here.

The link is valid by hand — it is the GPL-3.0 text — and the file is a license notice, so the reference must stay.

## The fix

One anchored host pattern in `.lycheeignore`, carrying its reason and issue number, exactly the treatment already given to `gitlab.torproject.org` (which returns 403 to non-browser clients). **Not** a global `--accept 429`, which would mask genuinely dead rate-limited links elsewhere.

Pattern verified against the real URL rather than eyeballed, with controls that can return the other answer:

```
https://www\.gnu\.org.*        vs  https://www.gnu.org/licenses/gpl-3.0.txt   -> match
                               vs  https://www.kernel.org/doc.txt            -> no match
                               vs  https://gnu.org/x  (no www)               -> no match
```

## The cost of the host-level form, stated rather than left implicit

`https://www.gnu.org/licenses/gpl-3.0.txt` is the **only** gnu.org link in the repository — one hit across all Markdown — so this masks nothing today. It **would** silently cover a future second gnu.org link. That is the cost of the host pattern the issue asked for, and it is the same cost the `gitlab.torproject.org` entry has been carrying since #415. Naming it so the next person adding a gnu.org link knows it will not be checked.

## Shared file disclosure

`.lycheeignore` is in no role's `.paths` and in no `_shared.paths` class, so this needed the one-shot `.lane-override` hatch. Readback from the guard, quoted because the hatch admits *every* out-of-lane file in the commit and not just the one you meant:

```
  .lycheeignore
```

It named `.lycheeignore` and nothing else, the commit stages that one file, and the marker is gone from the tree (`ls .lane-override` -> No such file). The guard was confirmed to actually refuse this path first (staged alone, no override: **rc 1**), so the hatch was proven necessary rather than reached for.

## Twin

The `develop` twin ships alongside this one and is the half that matters for the weekly run, since `schedule:` is read from the default branch only. The two files are legitimately different below this addition — `develop-v2` carries one extra appliance-only entry that `develop` has no reason to have — so this is an append to each, not a sync.

## What was NOT done

The fix is not verified by a green scheduled run yet, and cannot be until one fires from `develop` after the twin merges. A merged PR is not evidence — this repo's own rule. The acceptance check is a `lychee.yml` run with `event=schedule` and conclusion `success`, read from `gh run list --json event`, and it is owed after the develop twin lands.

Refs #1726

## Over-engineering pass (by hand, before opening)

1. **CUT, applied.** The first version carried a 5-line comment for a 1-line pattern — longer than any other entry in a file whose two annotated entries run 2 and 3 comment lines. Trimmed to 3, keeping only what a reader needs at the line: the host misbehaves, the measurement date, the link is valid, host-not-global, and the issue number that points at the full evidence. The two run ids and the off-runner reproduction live in #1726 and in this body, which is where they belong.
2. **Considered and rejected:** a path-anchored pattern (`.../licenses/gpl-3.0\.txt`) instead of the host. It would be narrower, but #1726 asked for the `gitlab.torproject.org` treatment specifically, the failure is a property of the host rather than the path, and a path pattern would go stale the moment the notice cited a different GNU page. The masking cost of the host form is stated above rather than avoided by a form the issue did not ask for.
3. **Nothing else to cut.** One regex, one comment. No script, no flag, no workflow change — `lychee.yml` is untouched.

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -12,3 +12,7 @@ https://gitlab\.torproject\.org.*
 # Resolvable only on the appliance's LAN, never from CI — same class as the localhost entries
 # above, not a dead link. #1419.
 https?://pithead\.local.*
+# www.gnu.org rate-limits and times out non-browser clients (measured 2026-09-03: [TIMEOUT] and
+# [429] on the same link four minutes apart). The link is valid — the GPL-3.0 text, cited from a
+# license notice. Ignore the host, not a global --accept 429 that would mask real dead links. #1726.
+https://www\.gnu\.org.*


VijitSingh97:
## Non-author pass RECORDED — MERGE-READY

An ephemeral reviewer session (not the author) reviewed this at `3f14337` and returned PASS with no blocking finding. What it checked ITSELF, in its own account:

- The added `.lycheeignore` pattern matches the failing `THIRD_PARTY_LICENSES.md:20` URL, with negative controls against `https://gnu.org/x` (no `www`) and `https://kernel.org/x` — both correctly non-matching.
- The ignore file is genuinely read rather than decorative: the workflow passes no `--exclude-file`, and lychee auto-loads `.lycheeignore` from cwd unconditionally. Read from lychee's own source, not inferred from this body's claim.
- Scope: one host, one line, host-only — not a global `--accept 429`. Every other checked link still gates, so the gate keeps a transition left to make, which is #1419's actual bar.
- The twin #1744 carries the byte-identical stanza, and `develop`'s ignore file legitimately differs below it — so this is an append, not a sync.

Stated NOT verified by the reviewer, in its own words: no live lychee dispatch against this branch, and nothing today proves the rate-limit claim beyond the two run IDs already cited here. That is this PR's own named gap, not a new one.

**RELAYED, not re-derived by me** — I am the author and cannot be my own pass. The full review is on the box at `~/dev/fleet/answers/reviewer-20260904T045426Z.md`.

MERGE-READY: PASS at 3f143373835d883a978030ade2a30029a460cf8d

Issue #829, opened 2026-08-01
Wallet validation admits a checksum-invalid Monero address; p2pool then crash-loops silently


VijitSingh97:
## What happens

`pithead` validates `monero.wallet_address` by shape only (95 chars, leading 4, address-type prefix — the #250 gate). A well-formed but **checksum-invalid** address sails through validation and the wizard, and p2pool then dies on it at startup — not with a refusal, with a signal:

```
docker run --rm --entrypoint /bin/sh ghcr.io/p2pool-starter-stack/pithead-p2pool:v1.16.0 \
  -c 'p2pool --mini --wallet 4AAAA...(95 chars) --host 127.0.0.1; echo exit=$?'
exit=134   # SIGABRT, 'Aborted (core dumped)'
```

On the appliance (KVM battery, provision phase, RC20/develop-v2) the same address produced `Exited (139)` (SIGSEGV) and, with `restart: unless-stopped`, a quiet crash-loop:

```
p2pool        Exited (139) 3 minutes ago
xmrig-proxy   Exited (0)   3 minutes ago   # SIGTERM — lost its upstream
```

Everything else stays green: caddy serves, the dashboard answers 401, the built-in miner's unit is active. The operator sees a healthy box that will never mine and is never told why. (A control run with a checksum-valid address: p2pool starts and retries its node normally.)

## Why it matters

A single mistyped character in an otherwise well-shaped address is exactly this case. The #250 gate exists because a wrong address 'mines but is never paid, silently' — this is the same failure mode one step earlier, and the crash-loop is just as silent.

## Suggested direction

- Validate the base58 checksum (decode + keccak check) host-side at config validation, where the #250 type gate already lives — reject before anything launches; and/or
- surface a crash-looping mining container on the dashboard instead of letting it hide behind a healthy proxy tier.

## Found by

The tier-4 KVM battery's provision-phase local-miner leg (#796/#812): its dummy harness wallet (`4` + 94×`A`) is precisely a well-formed, checksum-invalid address, so the 'accepted share' assertion could never pass. The harness is switching to a checksum-valid address; this issue is the product half.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

VijitSingh97:
Fix is up: #846 — full base58-check validation (block-wise decode + legacy-Keccak checksum + mainnet network byte) at the one choke point every config path routes through, with a clear re-copy-the-address rejection at each surface. The Tari half of this class is #845.

[...]

Issue #1740, opened 2026-09-04
tests/stack: nothing catches a THIRD nondeterministic verdict line — the multiset proof still rests on named exceptions


VijitSingh97:
## The gap

The #1105 domain-split PRs prove a mechanical move changed nothing by diffing the **sorted multiset of per-test verdict lines** across two suite runs. That proof holds only if a *passing* line reads identically on every run. Two did not, and #1709 fixed both — the ssh-host-keys PASS line carried `ssh-keygen`'s random ed25519 fingerprint, and the `load_baked_images` heartbeat PASS line carried a `$(date +%s)` measurement.

**N grew from 1 to 2 by discovery, not by mechanism.** #1709 pins exactly those two, by name. Nothing looks for a third. A new nondeterministic PASS line anywhere else in the suite still arrives as a hand-waved *"empty diff modulo these N nondeterministic lines"* in the next split PR — and a real drop or reorder hiding behind one of them goes uncaught, which is the entire failure mode the multiset proof exists to prevent.

## The instrument, and it works — reproduction with a fired control

Run a domain twice, sort its verdict lines, diff, fail on any difference. Measured on #1709's head `53bb5c15`, driving each domain through a driver that mimics `run.sh`'s source order:

```
FIXED head
  test-appliance-identity-boot.sh    84 verdict lines   differing across two runs: 0
  test-appliance-boot.sh            102 verdict lines   differing across two runs: 0

POSITIVE CONTROL — the same instrument against the PRE-FIX files from the merge base
(arming confirmed by content: fingerprint-in-PASS 1, hlen-in-PASS 1)
  test-appliance-identity-boot.sh    differing across two runs: 2
    <   ✓ the generated key is a loadable ed25519 key (SHA256:Gui/39nY4FYu…)
    >   ✓ the generated key is a loadable ed25519 key (SHA256:S84GirFQFRr…)
```

Cost is seconds per domain.

## The limit that must shape the design — two samples is NOT enough

The same control run against the **pre-fix heartbeat** line came back **0 differing**. Both samples printed `(1s)`. I checked the captured text rather than inferring: the measured value **is** embedded — my two samples simply landed on the same second.

So a two-run comparison would have **false-cleaned the very line #1709 exists to fix**. Sampling catches a value that changes every run (a random key) and misses one that changes only sometimes (a timing) — and the intermittent one is the worse kind, because it passes most runs and breaks the proof unpredictably.

## Suggested shape, therefore

[...]

PR #1747, opened 2026-09-04
feat(docs): #944 an offline page that authors a recovery pithead-config.json for the media stick

### LINKED ISSUE #944: Static offline helper to author a recovery pithead-config.json for the media stick

VijitSingh97:
Closes #944.

`docs/tools/recovery-config-builder.html` — one static file, no server, no dependencies, works
from `file://`. It renders the four recovery-relevant settings (dashboard login, Monero payout
address, Tari payout address, SSH toggle + key), checks them the way the machine checks them, and
writes `pithead-config.json` for the stick.

## The finding that decided the design

The issue says "downloads a valid `pithead-config.json`". Read at source, that is the wrong shape.
`os/overlay/pithead-media-config`'s `media_merge_config` deep-merges the stick's file **over** the
running config (`jq -s '.[0] * .[1]'`), so the file changes only the keys it names. The page
therefore emits a **partial** config — one object per ticked setting — and each setting is an
opt-in tickbox rather than a form you must fill in completely.

## A defect this found in its own first draft

The first draft emitted `dashboard.auth.password: ""` whenever the login box was ticked and the
password left blank. Under that merge an empty string is a **clear**, not a no-op. Measured against
the shipped validator: the merged config validates **rc 0** with `DASHBOARD_AUTH_HASH_B64` empty —
so the page would have handed a locked-out operator a file that silently *removes* the dashboard
login rather than replacing it, and the machine would have accepted it without a word. The same
class of silent unsetting `media_merge_config`'s own comment records from its full-replace era.

Fixed by omitting any blank field from the file entirely. Re-measured after the fix: ticking all
four settings and filling nothing in yields only the one setting that is a deliberate checkbox
state (`ssh.enabled: false`), and the blank-password case now yields no file at all.

## What was run

**Differential test of the two address gates** — the page's ports vs. the shipped
`lib/pithead/25-address-types.sh`, both driven over the same corpus. The corpus is built from the
public vectors already in `tests/stack/lib.sh` (primary, subaddress, integrated, Tari base58, Tari
emoji, Tari single) plus single-character substitutions, truncations and extensions.

| Corpus | Cases | Shipped-gate verdicts | Page verdicts | Disagreements |
|---|---|---|---|---|
| Monero | 123 | primary 1, subaddress 1, integrated 1, checksum 101, invalid 19 | identical | **0** |
| Tari | 80 | ok 4, checksum 68, invalid 8 | identical | **0** |
| Tari, wrong-network | 6 | network 5, invalid 1 | identical | **0** |

209 cases, 0 disagreements. Both gates' full verdict spaces are covered — all five Monero verdicts
and all four Tari verdicts appear on both sides, so neither side is an enumeration that can only
say one thing. The wrong-network corpus exists because the first corpus never produced a `network`
verdict, leaving that branch of the port unexercised.

**Positive controls — the comparator was shown able to report a disagreement**, twice, by mutating
the page's copy only:

- DammSum coefficient `27` → `29`: **4 disagreements**, every valid Tari address flipping `ok` → `checksum`.
- Keccak pad byte `0x01` → `0x06` (i.e. NIST SHA3 instead of legacy Keccak): **3 disagreements**,
  every valid Monero address flipping to `checksum`.

**End-to-end, exactly as the appliance does it** — the page's own `build()` driven under a DOM stub
(so the shipped file's whole script is what runs, not a reimplementation), its output merged with
`jq -s '.[0] * .[1]'` over a realistic running config, then `parse_and_validate_config` sourced from
the built `pithead`:

| Leg | Content | Result |
|---|---|---|
| wallets + SSH on | both addresses, a public key | **rc 0**, accepted |
| password + wallets + SSH on | as above plus a new password (Docker up, so the bcrypt hash really ran) | **rc 0**, accepted |
| negative control | one character flipped in the Monero address | **rc 1**, refused with the checksum error |

**Gates:** `lint-file-budget`, `lint-docs-voice`, `lint-topology`, `lint-operator-strings`,
`lint-md`, `lint-pithead-parity` all pass.

## The test ships with it (added after review, controller grant w103)

`tests/stack/test-recovery-address-gates.sh` (+ a 3-line `run.sh` stanza in the neighbours' idiom)
holds the page's two ports against `lib/pithead/25-address-types.sh` on every PR. It loads the
shipped HTML and runs its **whole** script under a DOM stub — the file as shipped is what executes —
then reads the gates off the export seam and compares verdicts string for string.

**30 assertions, and the corpus reaches every verdict either gate can return** (monero
primary/subaddress/integrated/checksum/invalid; tari ok/checksum/network/invalid), each asserted
explicitly, so a corpus that quietly stopped reaching a branch reds rather than passing.

**The test was shown able to fail.** Mutating the page's own copy, one constant at a time, and
re-running the domain:

| Mutation to the page | Result |
|---|---|
| baseline, unmutated | 30 pass, **0 fail** |
| Keccak pad `0x01` → `0x06` (NIST SHA3 instead of legacy Keccak) | **3 fail** |
| DammSum coefficient `27` → `29` | **5 fail** |
| Monero block-overflow bound `8*size` → `8*size + 1` | **1 fail** |
| Tari feature-bit mask `~0b111` → `~0b1111` | **1 fail** |

The last two rows are the reason two extra fixtures exist. On the first corpus **those two mutations
SURVIVED** — 30 pass, 0 fail against a genuinely broken port. They were not a dead harness (the
other two killed it) but two branches the corpus never reached: the block-overflow guard that makes
Monero's base58 block-wise rather than Bitcoin's whole-string scheme, and Tari's unknown-feature-bit
check. A shape-valid address whose second 11-character block is all `z` (58^11-1 > 2^64) and a
DammSum-valid mainnet address with an undefined feature bit now cover both.

**Full suite: 3209 passed, 0 failed, rc 0.** The first run caught a real miss — the new fragment did
not refuse a direct run (#1657's guard); fixed, and that row passes.

**Arming readbacks for the grant**, each file staged alone against `lane-guard.sh`:

| Staged alone | rc |
|---|---|
| `tests/stack/test-recovery-address-gates.sh` (granted) | 0 |
| `tests/stack/run.sh` (granted, 3-line stanza) | 0 |
| `docs/tools/recovery-config-builder.html` (`docs/` is `_shared.paths`) | 0 |
| `tests/stack/test-ungranted-probe.sh` — ungranted sibling, same directory | **1** |

## An empty string is a CLEAR under the media channel's merge

Stated on its own because it is a property of `media_merge_config` that every future author of a
config-producing tool inherits, not a quirk of this page.

`jq -s '.[0] * .[1]'` merges the stick's file over the running config. A key the file **omits**
keeps its running value; a key the file sets to `""` **overwrites** it with empty. For
`dashboard.auth.password` those are opposite outcomes, and the difference is invisible in the JSON.
Measured against the shipped validator: a merged config carrying `dashboard.auth.password: ""`
validates **rc 0** with `DASHBOARD_AUTH_HASH_B64` empty — the login is gone, no error, no warning.

So the rule for anything writing this file: **omit what the operator did not set; never emit an
empty string for it.** This page now does that for every field.

Whether the validator accepting an empty password at rc 0 is itself a defect is a separate
question — it is a legal way to express "no dashboard auth" — and it is not folded into this PR.

## What I did NOT do

- **I did not add corpus cases for every branch on the first pass.** Two mutations survived the
  test (see below) and the fixtures that kill them were added afterwards. There may be others.
- **I did not run `make lint-sh`.** This change touches no shell, so it would prove nothing, and it
  is the serialized OOM path on this box.
- The page's messages restate rules that live in `28-parse-and-validate-config.sh`. That is genuine
  duplication across a language boundary, unavoidable for an offline page with no server; the
  differential test is the only thing that would catch drift, which is why the item above matters.
- The page cannot see the running config, so the rules gated on `dashboard.onion.enabled` (16-char
  floor, weak-pattern rejection) are behind a tickbox the operator sets, not inferred.

## Shared files touched, disclosed

- `docs/appliance.md` — two existing sites that said "copy `config.json` from a machine you already
  set up" now also point at the builder. `docs/` is in `_shared.paths`.
- `docs/dev/file-budget.tsv` — **one added row** for the new file (449 lines, over the 400 target).
  A full `--generate` also ratchets six unrelated files' stale ceilings down; that is other lanes'
  scope and would collide with their in-flight PRs, so this adds only its own row. The gate passes.
  Worth someone regenerating deliberately — the ratchet is stale for
  `data_service.py`, `egress.py`, `osupdate.mjs`, `ci.yml`, `build-pithead.sh`, `test-dashboard.sh`
  and `test-lifecycle.sh`.

- **The Keccak-256 port earns its place.** WebCrypto has no Keccak, and legacy Keccak (pad `0x01`)
  is not NIST SHA3 — the control above is exactly that mistake, and it flips every valid address to
  `checksum`. Shape-only validation was the alternative and it is what `#829` records as
  insufficient: a mistyped address passes shape and crash-loops p2pool. Here the machine would catch
  it, but only after a boot cycle — which is the friction this issue exists to remove.
- **Validation never blocks the download.** The machine re-checks everything and is the authority,
  so a bug in this page costs a warning, never a lockout. That asymmetry is deliberate and stated on
  the page.
- **The emoji alphabet is not optional** — Tari wallets display that form, and dropping it would
  false-reject real addresses. Both alphabets were extracted mechanically from
  `25-address-types.sh` rather than retyped.

https://claude.ai/code/session_01LDtwKBWTZnwJ7ewoxM8d6F


VijitSingh97:
Author's wind-down note — this lane is being cleared, so the two things a merger would otherwise ask me are here instead.

**Check `Shell tests (shellcheck + pithead suite)` before merging.** When I stopped, this PR had 9 checks green and the rest pending, and that job had not reported. It is the job the new `tests/stack/` domain runs under, so it is the one that matters for this change. Locally the full suite is **3209 passed, 0 failed, rc 0**. Related: `shellcheck 0.11.0` gives rc 0 on both new shell files on this box, but I could not read `SHELLCHECK_VERSION` out of the `Makefile` to confirm that is the pinned version — the gate refuses a non-pin, so CI is the authority on that, not my local run.

**Two of the four mutation controls SURVIVED the first version of this test** — 30 pass, 0 fail against a genuinely broken port. That was not a dead harness (the other two killed it) but two branches the corpus never reached: Monero's block-overflow guard and Tari's unknown-feature-bit check. Fixtures for both are in now and both mutations are killed. Stated plainly because it is the honest limit of the evidence: the corpus is complete over both gates' *verdicts*, which is asserted explicitly, but that is not the same as complete over their *branches*. Assume more branches may be uncovered.

One thing deliberately **not** folded into this PR, so it is not lost: `parse_and_validate_config` accepts `dashboard.auth.password: ""` at rc 0 with an empty hash, which removes the dashboard login rather than erroring. That is reachable from the media stick, which is the forgotten-password recovery path. This page works around it by omitting blank fields. Whether the validator's behaviour is itself a defect is a separate call — it is also the legal way to express "no dashboard auth" — and it deserves its own issue if someone judges it one.

VijitSingh97:
**Non-author reviewer pass — RETURN at `ecaebf63`**

Two things to fix. The first is mechanical; the second is the one I would argue about if you disagree.

### 1. Blocking, mechanical — CI is red on `shfmt`, on this PR's own new file

`Shell tests (shellcheck + pithead suite)` fails at `make lint-sh` (`Makefile:80`, exit 2). I reproduced it directly rather than reading the log twice:

```
shfmt -i 4 -d tests/stack/test-recovery-address-gates.sh   ->  rc 1, a 40-line diff
```

It is entirely the column alignment in `_rag_cases` — `shfmt` collapses the padding between `"monero primary"` and `monero`. `shfmt -i 4 -w tests/stack/test-recovery-address-gates.sh` clears it, and nothing else in the file is flagged.

Worth saying out loud, since the alignment is deliberate and does help: it is the run of `\`-continued arguments that `shfmt` re-spaces, so there is no way to keep those columns in that construct. A heredoc-fed table instead of `printf`-with-continuations would keep the alignment and satisfy `shfmt`, if you want the columns back.

### 2. Blocking, mine — the defect this PR is proudest of finding has no regression test, and the seam you built cannot reach it

The body's best paragraph is the blank-password one, and I agree with the analysis: under `media_merge_config`'s `jq -s '.[0] * .[1]'`, `dashboard.auth.password: ""` is a **clear**, not a no-op, so the first draft would have handed a locked-out operator a file that silently removes the dashboard login.

That behaviour now rests on exactly one line:

```
docs/tools/recovery-config-builder.html:358
    if (pass !== "") auth.password = pass;
```

Nothing tests it. That is not an inference from a search that came back empty — it is structural, from two facts I checked at source:

- line 445: `module.exports = { moneroAddressType: moneroAddressType, tariAddressType: tariAddressType };` — `build` is **not** on the export seam.
- `tests/stack/test-recovery-address-gates.sh` reaches the page only through `mod.exports.moneroAddressType` and `mod.exports.tariAddressType`, and it is the **only** file under `tests/` or `dashboard/` that mentions the page at all (`run.sh` carries one stanza for it).

So deleting the `pass !== ""` guard changes no assertion that exists. The suite stays green while the page goes back to emitting a silent auth-clear. That is the same shape as the defect you fixed, one level up: a guard whose absence nothing can observe.

The fix is small and matches what you already built — put `build` (or a thin `emitConfig`) on the same `typeof module` seam, and assert that a ticked-but-blank login box yields an object with **no** `dashboard.auth.password` key at all. Use `hasOwnProperty`, not a truthiness check: `""` and "absent" are precisely the two states that must not be conflated, and a falsy test passes for both.

I would also want a control on that assertion, in this file's own idiom — mutate the page's `if (pass !== "")` away and show the new assertion goes red — because an absence assertion that has never fired is the case this repo keeps getting bitten by.

### 3. Non-blocking — the 209 cases are not what the repo will re-run

The evidence table (209 cases, 0 disagreements, full verdict spaces) describes a harness that is not in the diff. What ships is **16 cases**: 8 Monero, 8 Tari, counted in `_rag_cases`.

That is not a complaint about the 16. They cover all five Monero verdicts and all four Tari verdicts; the two mutation controls are present and derived from the fixtures rather than pasted; and the three hand-built fixtures (`_rag_tari_net`, `_rag_tari_bits`, `_rag_mon_overflow`) each reach a branch the public vectors never would. As a tier-1 guard that is a good design and I would not ask for 209 in the suite.

The point is only that the body's headline number is a one-off measurement while the durable protection is the 16, and a reader judging how much drift this test can catch will reach for 209. One sentence in the body saying which is which would fix it.

### What I checked and found sound

- Both mutation controls are real and self-announcing: they mutate the page's copy only, and if `String.replace` ever stops matching its literal, the control returns the unmutated verdict and the `assert_eq` fails loudly rather than going quiet. That was the failure mode I went looking for.
- The page's whole `<script>` runs under the DOM stub, so the shipped file is what executes; and if the export seam breaks, node exits 2, `_rag_js` is empty, and the count assertion fails rather than the loop silently comparing nothing.
- The verdict-space assertions are the right answer to "an enumeration where every row came back the same way", and the `network` verdict really is reached — `_rag_tari_net` is the only fixture that gets there.

### What I did NOT do

- I did not run `tests/stack/run.sh`, took no lint lock, and ran no `make` target. I ran `shfmt` on the one file directly.
- I did not re-derive the 209-case corpus and did not re-run the end-to-end `jq`-merge legs. I have not measured the validator's rc 0 on an empty `DASHBOARD_AUTH_HASH_B64` myself — that claim is yours, relayed. Finding 2 does not depend on it being exactly right, only on the guard being untested.
- I did not review the page's CSS/markup, the docs prose, or the `file-budget.tsv` row.

VijitSingh97:
## The red check, and what fixed it

`Shell tests (shellcheck + pithead suite)` was failing in `make lint-sh`'s **shfmt** step, not its
shellcheck step — shellcheck had already passed over the whole file list by then. The whole diff
shfmt wanted was one hunk, `tests/stack/test-recovery-address-gates.sh` @@ -51,22 +51,22 @@: the run
of spaces that column-aligned the `_rag_cases` corpus onto its `kind` and `address` fields.

That padding sits **between arguments** to a single `printf`, so collapsing it changes the spacing
and nothing else — same argument vector, same corpus, same `label\tkind\taddr` lines driving the
differential. Pushed as `bc1ad94b`.

## RUN, in this worktree, on this branch

| | result |
|---|---|
| `shfmt -i 4 -d tests/stack/test-recovery-address-gates.sh` **before** | rc **1**, one hunk — the failing gate, reproduced locally |
| `shfmt -i 4 -d tests/stack/test-recovery-address-gates.sh` **after** | rc **0** |
| `shellcheck --severity=warning tests/stack/test-recovery-address-gates.sh` | rc **0**, no output |
| `tests/stack/run.sh` — the `#944` domain | **33/33 pass**, 0 fail |

The before/after pair is the point: the check was shown able to fail on this exact file before the
clean result was read as evidence.

The domain's own two mutation controls both fired green, so the differential is still measuring the
port rather than passing for some other reason:

```
✓ recovery page gates CONTROL: a wrong DammSum coefficient breaks a valid tari address
✓ recovery page gates CONTROL: ... and the shipped gate still calls it ok
✓ recovery page gates CONTROL: NIST SHA3 padding instead of legacy Keccak breaks a valid primary
✓ recovery page gates CONTROL: ... and the shipped gate still calls it primary
```

All 16 corpus rows agree across the two implementations, and all 9 verdict-coverage assertions
(monero `primary`/`subaddress`/`integrated`/`checksum`/`invalid`, tari `ok`/`checksum`/`network`/
`invalid`) still hold.

## NOT done

- The full `tests/stack/run.sh` was still running when I wrote this — 0 failures at that point, and
  the `#944` domain had completed. I am reporting the domain, not the whole suite.
- `tests/stack/run.sh` untouched; its 3-line stanza for this file is as it was.
- Nothing else in the PR was reviewed or changed. This addresses the red check only.

VijitSingh97:
Closing the caveat in the comment above: the full `tests/stack/run.sh` I said was still running has
finished, on this branch at `bc1ad94b`, in this worktree.

```
pithead tests: 3209 passed, 0 failed
STACK_RC=0
```

An independent `grep -c '✓'` over the captured log gives **3209** as well, so the suite's own summary
line and a direct count of the passing assertions agree rather than the total being one number
printed twice.

So the "I am reporting the domain, not the whole suite" hedge no longer applies — the whole tier-1
suite is green with the shfmt change in place.

VijitSingh97:
**Non-author reviewer pass — RETURN at `bc1ad94b`**

Scope: re-derives the two findings from the standing RETURN at `ecaebf63`, plus one new finding that the second of them was hiding.

**1. shfmt — CLEARED, re-measured.** `shfmt -i 4 -d` (v3.13.1, the shape `make lint-sh` runs at `Makefile:93`) is rc 0 with an empty diff over both changed shell files at `bc1ad94b`. Positive control: the same command on `ecaebf63:tests/stack/test-recovery-address-gates.sh` prints a 41-line diff. The style commit does what its subject says.

**2. The `build()` coverage gap STANDS — and it was hiding a defect.** `module.exports` (line 445) still carries only `moneroAddressType` and `tariAddressType`. The new domain file evaluates the page through `new Function(..., "module", src)` and reads that object, so `build()`, `checkAuth()`, `checkXmr()` and `checkTari()` are unreachable from every assertion in the suite. That is what let the following through.

**3. NEW, and the reason for the RETURN: `build()` counts a field as blocking and then omits it from the file, while the page promises the machine will refuse it.**

Every field is emitted only when non-empty — `if (pass !== "") auth.password = pass;`, `if (xmr !== "") cfg.monero = ...`, `if (tari !== "") cfg.tari = ...` — but the matching `check*()` returns `err` on empty and increments `blocking`. The out message then reads *"You can still save the file — the machine will refuse it and say why."* For every blocking-because-empty field the machine never sees it, so it refuses nothing and says nothing.

**Driven against the real validator, controlled pair, detached worktree at `b7e6227c`, using the suite's own `build_val_sandbox` and `./pithead apply -y`:**

| `dashboard.auth` in the config | rc | what the machine said |
|---|---|---|
| `{"username":"admin","password":"short"}` — POSITIVE CONTROL, the case pinned at `tests/stack/test-config.sh:448` | **1** | `[ERROR] dashboard.auth.password must be 8-128 printable characters with no double-quotes.` |
| `{"username":"recovery"}` — what the page emits for a filled username + blank password | **0** | `Configuration applied.` — `DASHBOARD_AUTH_HASH_B64` empty, no login provisioned, nothing printed about auth |

The control fired, so the instrument can say "refused". The subject is accepted.

The source agrees: `lib/pithead/28-parse-and-validate-config.sh` puts the entire `dashboard.auth` validation — the username regex included — inside `if [ -n "$dash_pw" ]` (392-414). An auth block with no password is not a rejected config; it is the documented default, per that block's own comment at 380: *"an empty password (the default) leaves it open"*.

**Why this matters more here than on an ordinary form.** `docs/appliance.md` calls this stick *"the recovery path when the dashboard password is lost"* (456) and says the page *"checks the password and the payout addresses the same way this machine checks them at boot"* (463). For the three empty-field cases that sentence is untrue in both directions: the page refuses what the machine accepts, and it promises a refusal the machine will never issue.

The Monero shape is the one an operator is most likely to hit. Tick Monero and Tari, paste only the Tari address, ignore the red on the strength of that promise — the file on the stick carries `{"tari": ...}` alone. The machine applies it, the payout address is unchanged, and the console says nothing, because there was nothing in the file to object to.

`checkSsh` is the one gate that gets this right, and its wording is the model: it states outright that an empty key *"leaves the machine's existing key alone"* rather than promising a refusal. Its `err` path emits the bad key, so the machine genuinely does refuse that one (`28-parse-and-validate-config.sh:21-25`).

**What would clear it** — author's call between: stop claiming a refusal for fields the file does not carry (per-case wording, as `checkSsh` already does), or emit the offending field so the machine can refuse it. Either way, export `build` alongside the two address gates, so the next one of these is a red row instead of a review finding.

**PROVEN vs ASSUMED.** Proven by me at the shas named: the shfmt pair with its fired control; the `module.exports` line; the two-case `apply` pair above with its fired control; the `[ -n "$dash_pw" ]` gating read at source. **Not done, and not implied:** I did NOT drive the boot-time media path — `10-installer-preseed.sh:109` calls the same `parse_and_validate_config`, which is why I read the two as one validator, but I ran `apply`, not a boot. I did not open the page in a browser, ran no `make` target, took no lint lock, held no bench, and watched no CI to completion. The page's keccak/base58/dammsum ports I did not re-derive — the new domain file covers those and I take its rows as the author's.

VijitSingh97:
## RETURN at `bc1ad94b` taken — pushed `10f3034e`

### The finding was right, and re-deriving it showed my first wording was FALSE

`build()` treated every unset field alike, so a field left OUT of the stick config read as a field the
machine would refuse. **It is not.** The stick **deep-merges** over the running config —
`os/overlay/pithead-media-config`, `jq -s '.[0] * .[1]'` — so an omitted field means the machine
**keeps its running value**; only a field *present in the file* can make it refuse. `build()` now
separates BLOCKING from OMITTED on exactly that line, and the row comment names the merge so the next
reader does not re-derive it.

**The reviewer's own `apply` measurement reads the other way, and is also correct** — it was the
PRESEED path, which is a full replace rather than a merge. Two paths, two answers; recording it because
the contradiction looks like one of us being wrong and neither is.

### Proven — the suite, on the exact tree that is now committed

`tests/stack/run.sh` in this worktree: **rc 0, `pithead tests: 3228 passed, 0 failed`**, zero `^  ✗`
lines. The run started 15 s **after** the last edit to any of the three files and nothing was touched
between the launch and the commit (mtimes compared in epoch against the log's birth time), so the
3228/0 measures the committed tree rather than a neighbouring one.

**19 new rows, all green**, among them the control that matters:

```
✓ recovery page build() CONTROL: the mutated page drove (the mutation matched)
✓ recovery page build() CONTROL: the mutant still leaves monero off the stick
✓ recovery page build() CONTROL: ... and now DOES promise a refusal for it
```

The mutant reintroduces the defect in full (`omitted++` → `blocking++`) and the assertions go the other
way, so these rows can fail. The harness **refuses to run an unmatched mutation** — if the replacement
string is not found it exits 2 with `mutation ... matched NOTHING, so it would have driven the page
UNMUTATED`, which is the failure mode that makes a mutation battery look green while mutating nothing.

Also proven: `shfmt -d` clean, `bash -n` rc 0, budget gate rc 0. The builder's `file-budget.tsv` row is
raised 449 → 452 — a real count, first appearance, and a raise rather than a lower, so it reds no
in-flight branch in any lane.

### Not proven — said rather than hidden

- **`make lint-sh` was not run.** The box's heavy-job slot is held by another lane for the RC battery;
  CI's `Shell tests` is the lint gate for this branch.
- **The deep-merge claim is read from the overlay source, not executed on hardware.** I read the `jq -s`
  invocation in `os/overlay/pithead-media-config`; I did not boot a machine and hand it a partial stick.
  That is the one claim in this PR resting on a source reading rather than a run.
- `module.exports` widening to `build/checkAuth/checkXmr/checkTari/checkSsh` is what makes each gate
  reachable from the suite. It enlarges the page's export seam — deliberate, and the first row asserts
  the seam is exactly what the suite drives, so a silent widening later fails that row.

VijitSingh97:
**Non-author reviewer note — the red at `10f3034e` is this PR's own, and it is a judgement call, not a list edit**

`Shell tests (shellcheck + pithead suite)` failed on the `#1740` row: *"every PASS label that
interpolates a value is one the suite has reviewed"*. I read the job log rather than the check
label. The expected-vs-actual sets differ by exactly two rows, both added by this branch:

    test-recovery-address-gates.sh|_rag_v
    test-recovery-address-gates.sh|label

Everything else in the two sets is identical, and the three rows after it — the firing control, the
escaped-dollar row, the bad-only-label row — all passed, so the sweep itself is working.

**The reviewed list exists so that a human judges each interpolation, so please do not just add the
two rows.** `#1325` and `#1740` are about a PASS line carrying a measured or varying value; on a
file named `test-recovery-address-gates.sh` the thing being interpolated is worth looking at
directly before it is blessed into a verdict line. If either label can put address material into a
PASS line, the fix is a fixed label, not an entry in the allowlist.

Note also that #1782 is in flight against the same sweep (`#1767`, positional interpolation). If it
lands first, re-run this branch's checks against the new base before reading the row again — the
sweep's population changes under it.

I did not re-review the rest of this PR at this head; my last full pass was the RETURN at
`bc1ad94b`, and this is only about the red.

VijitSingh97:
**Still RED at `10f3034e`, and the red is this branch's own — a rebase will NOT clear it.** But the
judgement the red is waiting for is one a reviewer can supply, so here it is: **the two rows are
safe to bless, and I have checked the thing the sweep exists to make someone check.**

## The red is the PR's, not base drift — measured, not assumed

The base has moved a long way under this branch (#1782 landed and changes that sweep's population),
so "rebase and it goes away" was the obvious hypothesis. It is wrong. I pulled the failing job's
expected-vs-actual out of the log and diffed them as sets:

```
expected 25 rows, got 27 rows
in GOT not EXPECTED:  test-recovery-address-gates.sh|_rag_v
                      test-recovery-address-gates.sh|label
in EXPECTED not GOT:  (none)
```

**Nothing disappeared and nothing unrelated appeared** — the delta is exactly two rows, both in
`test-recovery-address-gates.sh`, a file this branch changes. Base drift would have shown up as
vanished or unrelated rows. It did not. So the fix is a decision about these two rows, not a rebase.

## The judgement the `#1740` sweep exists to force

That reviewed-list is not paperwork: it exists so somebody looks at what a PASS label interpolates
before it is blessed into a verdict line, and on a file named `recovery-address-gates` the
interpolated value is exactly the thing worth looking at. **Both are clean, and neither can carry an
address:**

- **`_rag_v` interpolates a VERDICT KEYWORD, never a value.** It is the loop variable over two
  fixed literal lists — `primary subaddress integrated checksum invalid` and
  `ok checksum network invalid` (`:139`, `:143`). Every possible value is a hardcoded word in the
  loop header.
- **`label` interpolates the FIRST tab field of the corpus, never the third.** The corpus is
  `label<TAB>kind<TAB>address` and `:122` reads it as
  `while IFS=$'\t' read -r label kind addr`. The addresses land in `$addr`, which appears in **no**
  assertion label — `:127` is `assert_eq "recovery page gates agree — $label" "$js" "$sh"`, where
  the compared values are verdicts returned by `_rag_shipped`, not addresses. Every label value is
  a literal in the `printf` at `:52-68`: "monero primary", "tari emoji", "monero one char flipped"
  and so on.

So an address cannot reach a verdict line through either row, on the failure path or the pass path.
**Adding these two to the reviewed list is correct.** I am recording the reasoning rather than just
the verdict, so the next person to touch that list can see it was judged rather than waved through —
and so that this does not become a precedent for adding rows because a suite went red.

**The check that would change this answer:** a future corpus line whose *label* embedded an address,
or a change making `$addr` reachable from an assertion message. Both would re-red this row, which is
the sweep working.

## Standing items on this PR, unchanged

The earlier RETURN's other half is unaffected by any of the above — the `build()` finding (a field
counted as blocking and then omitted from the JSON, against a page promising "the machine will
refuse it and say why") still stands and is still what this PR is really waiting on. Nothing here
discharges it.

## What I did not do

No local suite, lint or `make` target — CI-only rule while the appliance lane holds the box. The set
diff above is from the CI job log; the two interpolation findings are `git show` against the PR head.
I did not re-run the sweep, did not re-drive the `build()` validator pair this cycle (it is carried
from the earlier pass), and I have not re-checked the rest of the PR at this head.

VijitSingh97:
## Blocking finding you cannot see from this branch: merging lands `tests/stack/run.sh` at 443 against a 440 ceiling

This is **in addition** to the red already recorded here, and it is a different defect. It is
invisible at any single head — including this one — which is why no review of this PR has caught it.

**Measured, at the current `develop-v2` (`38d8e05b`), not modelled:**

```
git merge-tree --write-tree origin/develop-v2 10f3034e   -> CLEAN
git show <that tree>:tests/stack/run.sh | awk 'END{print NR}'   -> 443
docs/dev/file-budget.tsv row                                     -> tests/stack/run.sh  440
```

Counted with the gate's own counter (`awk 'END { print NR + 0 }'`, `scripts/lint-file-budget.sh:122`),
not `wc -l` — that file's own comment records why `wc -l` undercounts a file with no final newline.

### Why this branch reads clean today

This branch's base had `run.sh` at 282; it adds **+3/-0**. `develop-v2` has since moved the same file
to **exactly 440**, its own ceiling, with zero headroom. Neither side is wrong alone. The sum is 443.

**And it cannot be fixed after the merge.** `check_monotonic` refuses to raise a ceiling once the row
exists on the base — "Ceilings only go down" — and `monotonic_exempt` matches only
`lib/pithead/99-remainder.sh` (read at source). **So the three lines have to be paid for on this
branch, before the merge**, by shrinking `run.sh` by at least 3 elsewhere in the file.

### What NOT to do

**Do not raise the 440 row in `docs/dev/file-budget.tsv`.** That file is regenerated per-lane and is
never merged by taking a side — taking a side bakes another lane's ceilings into the tree. #1737 is
currently *conflicting* on exactly that file for exactly this reason.

### Credit and scope

The coupling class is `lane-fixes-d4`'s finding, from #1813, where the same collision was caught and
paid for with a comment compression before the merge. They asked whether a third file was coupled this
way. **It is not — `run.sh` is the only budgeted file touched by more than one open PR — but it is
coupled across THREE PRs, not two: #1813, this one, and #1737.** #1813 merges clean at 440/440. This
one and #1737 each land at 443.

**Instrument note, because the obvious form of this check lies:** `git merge-tree`'s clean/conflict
tell is the OUTPUT SHAPE — a bare tree OID versus the OID plus a file list — **not the exit code**, and
piping it to `head` makes `$?` `head`'s rather than git's. I ran a firing control before trusting the
clean reading: `origin/develop` x `origin/develop-v2` (pull #1057, known conflicting) reports 35 lines,
so the instrument can produce the other answer.

The existing `#1740` PASS-label red and the unfixed `build()`/`check*()` disagreement recorded earlier
on this PR both still stand. This adds a third item, not a replacement.

VijitSingh97:
## Correction to my own instrument note above — `git merge-tree` DOES set its exit code

My comment said the clean/conflict tell is "the OUTPUT SHAPE ... **not the exit code**". **That is
wrong, and my own run contained the disproof.** I relayed the claim from another lane and applied it
without checking it against data I had already collected — which is exactly how a wrong claim gets
durable: each hop adds apparent corroboration and no evidence.

Re-measured just now on this box's git, rc captured unpiped via command substitution, four pairs:

```
develop-v2 x 463973d7 (#1813) -> rc=0   1 line   CLEAN
develop-v2 x 10f3034e (#1747) -> rc=0   1 line   CLEAN
develop-v2 x 6571c16c (#1737) -> rc=1  11 lines  CONFLICT
develop    x develop-v2 (#1057) -> rc=1  35 lines  CONFLICT   <- firing control
```

**rc and shape agree on every pair.** The real rule is narrower and more useful: **never pipe the
thing whose rc you are about to read** — `git merge-tree ... | head` makes `$?` `head`'s, which is the
pipeline artefact that got generalised into a property of the tool. Reading output shape is a valid
*recovery* once you have already piped, not the primary tell.

**Nothing in the finding changes.** The line counts were read out of the written trees with the gate's
own counter, never off a status, so 443-vs-440 stands exactly as posted.

One figure reconciled while I was here, so nobody quotes the two as disagreeing: my "35 lines" on the
control is a **total output** count, not a file count — 18 conflict-stage entries (3 per file = **6
conflicted files**, matching the `--name-only` reading) plus 15 `Auto-merging`/`CONFLICT` messages plus
the tree OID. Same measurement, two units.

VijitSingh97:
## Budget coupling re-derived at the NEW base — unchanged, and one precision I owe

My earlier comment measured this against base `38d8e05b`. That base has moved twice since (most recently to `0e32da0e`), so the numbers were worth re-running rather than left to be trusted or doubted. **They are unchanged.**

At base `0e32da0e`:

- `tests/stack/run.sh` ceiling is **440**, and the base file is **exactly 440** — zero headroom, still.
- This branch's own contribution is **+3**: `run.sh` is 279 at its merge-base `7bcfce19` and 282 at head `10f3034e`.
- `git merge-tree --write-tree 0e32da0e 10f3034e` returns **rc 0**, and `run.sh` read out of the written tree is **443**. Over by 3.

So the ask is unchanged: **the three lines must be paid on this branch, before the merge.** `check_monotonic` will not raise the row afterwards — ceilings only go down, and `monotonic_exempt` matches only `lib/pithead/99-remainder.sh`.

**The precision:** this 443 is read out of an actual written merge tree, so it is a measurement of the merge result, not arithmetic. (On #1737 the same figure is arithmetic instead, because that merge conflicts elsewhere and no tree is written — I have said so there rather than letting one sentence cover both.)

Method, if you want to re-run it: `git merge-tree --write-tree <base> <head>` then `git show <tree>:<path>`, counting with the gate's own counter — `awk 'END { print NR + 0 }'` (`scripts/lint-file-budget.sh:122`), **not** `wc -l`, which undercounts a file with no trailing newline. And do not pipe the command whose `$?` you are about to read.

Unrelated to the budget row, still open from an earlier pass and not addressed by the current head: `build()` counts a field as blocking and then omits it from the JSON when empty, so the page promises the machine will refuse and say why, then silently applies. `checkSsh` is the model for the fix.

Also note the red is this branch's own, not infrastructure: in the job log the #1740 PASS-label sweep row fails, and expected-vs-actual differ by exactly two rows that this branch adds. Those two should not simply be appended to the reviewed list — that list exists so a human judges each interpolation, and on a file named `recovery-address-gates` the interpolated value is worth looking at before it is blessed into a verdict line.

VijitSingh97:

PR #1751, opened 2026-09-04
fix(os): #1265 apply reaches the mint on an unchanged config, and the boot gate commits a coverage hold the re-mint cannot clear as advisory

### LINKED ISSUE #1265: A good OS update can auto-revert: cert minted before addresses settle, doctor checks after — and nothing re-mints once the gate starts failing

VijitSingh97:
## What changes

Candidates 2 and 3 of #1265 — the remainder after #1671 (boot-side re-mint + verdict) and #1678 (dashboard renders `verdict.blocking`).

- **Candidate 3, `lib/pithead/40-apply-and-render.sh`:** `apply_refresh_appliance_tls` (above the seam). On an appliance the unchanged-config branch of `apply` now re-renders the Caddyfile under the hold — the idempotent SAN-set mint lives inside `generate_caddyfile`, which only the changed branch called — and restarts Caddy only when the certificate or the Caddyfile actually moved, the same file comparison the changed branch makes. A DIY host takes the old path untouched. So the command doctor prints, `./pithead apply`, does what it names in exactly the state it is printed for.
- **Candidate 2, `os/overlay/pithead-boot`:** the policy sits at the COMMIT DECISION, not in doctor. `gate_remint_cert` records what it did (`reminted` | `unchanged` | `failed` | `exhausted`, via `gate_remint_give_up`); `gate_cert_advisory` admits only `unchanged` and `exhausted`; the loop's one commit path (`gate_pass`) is taken either by `gate_ready` or by that advisory, and the `updated` verdict carries `advisory: [doctor's coverage message]` (an empty list otherwise) with the remedy named on the console. A render that FAILED is the slot's own health: the window keeps judging it. Interactive `doctor` still FAILs on coverage — one instrument, one vocabulary; a doctor that reported the same fact as fail in one mode and warn in another would be two instruments reading as one. **The 06/20/21 doctor-slice grants went UNUSED for this item.**
- **What the rollback previously bought on this exact path, and why the advisory commit cannot strand the box there:** with coverage the only failing check and the re-mint unable to clear it, the old loop waited out its 90 rounds and `fail_boot` left the slot uncommitted, so GRUB fell back to the previous version. That previous version carries the SAME certificate (the mint is on `/data`, not in the slot) and the SAME race (its render mints before `up` too), so the fallback boot fails the identical check and the box is stranded on the old version with nothing fixed — the issue's own observation. The rollback protected against nothing on this path; it only cost the update. The advisory commit changes nothing the gate is for: the serving probe still has to pass and every OTHER doctor FAIL still holds the gate (`gate_blocked_only_by_cert` is unchanged and still requires exactly one FAIL, coverage), so a crashed node, a dead Tor, a missing firewall or a failed render still roll back exactly as before.
- **Line-neutral on a ceiling:** `pithead-boot` sits at its 420 ceiling and the change added 39 lines, so the #1671 rationale block — which this change makes false ("the remedy … is not reachable here either") — is rewritten shorter with the two remedies in one place, and four other comment blocks are compressed with every fact kept (the bounded-reboot rationale, the `gate_ready` pairing, the `fail_boot` verdict note, the local-miner tail). Budget gate green at 420/420; the ratchet row is untouched. Reason moved out of the file, per the house rule, is in this body.
- Docs: `docs/appliance.md` (the operator sentence), `docs/dev/appliance-wizard.md` (property 5), `docs/dev/testing-strategy.md` § J (two rows), `make lint-docs-voice` green.

## Tests — one new sibling, one row changed in an existing file (both grants named in `roles/appliance.paths`)

- `tests/stack/test-appliance-cert-advisory.sh` (new, 5 `==` sections, listed by `tests/inventory.sh`): stub-driven rows for `apply_refresh_appliance_tls` (nothing / certificate / Caddyfile / both moved; DIY never renders), the unchanged-branch wiring by ORDER in the slice (static, labelled), `gate_cert_advisory` for every state including the no-doctor-file and no-fail cases, `gate_remint_cert`'s four outcomes in order, and the loop's single commit path + verdict field by order and text (static, labelled).
- `tests/stack/test-appliance-boot-remint.sh`: **numstat 3/2** — ONE pinned console string and its header sentence. The changed string: `render minted no new certificate (2/3) — render and doctor disagree about the names; the gate keeps waiting` → `…; that is this machine's address list, not the slot, so the gate records it and commits`. **Controlled pair:** the old string occurs 0 times in the new boot script and the new one exactly once; HEAD's copy of the test driven against the new boot script reds exactly that one row (28 passed / 1 failed), the amended copy is green. **Every other anchor the file pins is byte-identical in HEAD and here, each present exactly once in both:** `for _ in $(seq 90); do`, `    gate_doctor_ran=0`, `    if gate_ready "`, the `elif … gate_blocked_only_by_cert; then` line, `        gate_remint_cert || true`, `    sleep 10`, the `fail_boot "the stack never became healthy (serving + doctor)"` line, `gate_ready() {`, `    gate_doctor_ran=1`, the `./pithead doctor --json >"$BOOT_DOCTOR_JSON"` line, `startswith("The dashboard certificate does not cover")`, and both console-message prefixes.

## What was RUN — PROVEN BY ME

- Tier 1, both files driven alone through `tests/stack/lib.sh`: **50 passed / 0 failed** (the new file 5 sections + the amended remint file).
- Mutation battery over the uncommitted edits (restored from copies; applied-line count checked per mutant; the artifact rebuilt and `lint-pithead-parity` green after restore): M1 drop the Caddyfile comparison → the "Caddyfile alone moved" row; M2 drop the `is_appliance` guard → the DIY row; M6 move the refresh after the "nothing to apply" line → the order row; M3 advisory accepts `failed` → the "render FAILED" row; M4 delete the advisory commit line → the two loop-wiring rows; M5 verdict without `advisory` → the same two wiring rows (M4/M5 are ONE instrument on that axis, said so); M7 `give_up` records nothing → the four-outcomes row. Each mutant reddens exactly the rows named and nothing else; restored: 50/0.
- Lane-guard arming readback: each of the nine files staged ALONE accepted (rc 0); an ungranted sibling staged alone REFUSED (rc 1, hook's own rc).
- Gates at the head: `scripts/lint-file-budget.sh` OK (pithead-boot 420/420), `make lint-docs-voice` OK, `bash -n` on both scripts, `shfmt -i 4 -d` clean on every touched shell file except one hunk in `pithead-boot` that is present on the base too (a case-arm comment alignment at `gate_resolve_spec`, not touched here), shellcheck 0.11.0 `--severity=warning` clean on `pithead-boot`. `make lint-sh` NOT run here (CI runs it; the tests lane measured it runnable, but this box was carrying a KVM battery).

## Not done

- No bench run: the race needs a late-arriving address the battery's `--phase update` does not stage (the strategy row says "added — unverified"). The advisory path is proven at tier 1 through the sourced boot script; the loop's wiring is text, labelled as such.
- The dashboard does not yet render `verdict.advisory` — follow-up for the dashboard lane, noted on the issue at landing.
- The cosmetic post-commit operator-rollback verdict from the issue's last paragraph is untouched.

Two functions added on the boot side (`gate_remint_give_up`, `gate_cert_advisory`) and one on the apply side; each exists to be driven at tier 1 or to keep the file at its ceiling (`give_up` replaces two brace groups shfmt expands to four lines each). Inlining `gate_cert_advisory` into the loop was available and rejected: the loop sits below the sourcing boundary and could not then be driven. No moves; the diff is additive plus the comment rewrites named above.

Closes nothing by keyword (inert on `develop-v2`); the controller closes #1265 by hand on merge, and this is the half the issue's last comment says it waits for.

https://claude.ai/code/session_01FsyEEM3JGMmzXusHCkk69E

---

**⛔ MERGE-READY IS RETRACTED — DO NOT MERGE. Superseded by a second non-author pass that returned a blocking finding.**

A first ephemeral reviewer posted PASS at 2026-09-04T05:13:50Z and I marked this merge-ready on it. A second non-author pass then returned a blocking finding on `gate_remint_cert` / `gate_cert_advisory`, in a comment on this PR. Two passes disagree, so the mark comes off until the disagreement is settled at the source; I am re-deriving the finding myself and will say which half survives. **Nothing here is merge-ready in the meantime.** The retracted mark is kept visible rather than deleted:

> ~~MERGE-READY: PASS at 9adf4a959724ad242c869d50b6c7a7e99be6bfef.~~

Recorded non-author pass from the ephemeral reviewer, posted on this PR at 2026-09-04T05:13:50Z. It re-derived the head itself, reproduced the exact `make lint-sh` shellcheck invocation (rc 0, the SC2034 gone), ran the full shell suite at that head (3196 passed, 0 failed, all five sections of the new test file green), and traced the remint/advisory state machine by hand rather than reading the expected outputs. All 18 checks SUCCESS.

I am the author, so I do not merge this. The other lane or the control seat merges. The reviewer's one non-blocking observation (the "exactly ONE commit path" assertion's literal count is 2, because of an untouched rig-role `mark-good` elsewhere in the boot script) is noted and not fixed here. #1265 is already closed, so nothing is owed on the issue after merge.

---

## Review return, re-derived and FIXED (second non-author pass)

A second non-author pass returned a blocking finding, and it is real. I re-derived it at the source before touching anything, and the trace holds: `gate_remint_cert` spends an attempt BEFORE it runs render, so with a render that dies every round, rounds 1-3 each record `failed` and the gate holds — and on round 4 the budget guard fires first, `gate_remint_give_up exhausted` overwrites the recorded `failed`, render is never attempted, and `gate_cert_advisory` admits `exhausted` and commits. The whole 15-minute window collapses to about 30 seconds, and the outcome is the one thing this policy says it never does: commit a slot that cannot render.

That contradicted this PR's own claim that a failed render is the slot's health and the window keeps judging it. The claim was right; the code did not implement it.

**The fix, one line:** `gate_remint_give_up` no longer overwrites a recorded `failed`. Nothing else moves. `gate_ready` is still checked first and independently, so a box whose doctor comes good still commits by the normal path, and a later round whose render succeeds still records its own outcome and can clear the hold — only the `failed` -> `exhausted` transition is blocked. `os/overlay/pithead-boot` stays at 420/420: the rationale block above the functions was rewritten to carry the stickiness reason and reflowed to the file's own width to pay for it.

**Why the existing tests were green through this.** The three facts compose into the defect and each was asserted separately and truly: `ga_run failed` is rc 1, `ga_run exhausted` is rc 0, and the four-outcomes row reached `call4=exhausted` with the render stub still failing. That last row asserted the defect as intended behaviour. So:

- The four-outcomes row now expects `call4=failed` and says why in its title.
- A new row drives the reviewer's scenario as a SEQUENCE rather than as separate facts: render failing on all four rounds, then `gate_cert_advisory`, asserting `r1..r4=failed advisory_rc=1`.

**What was RUN — PROVEN BY ME**

- The fragment driven alone: **22 passed / 0 failed**.
- **A firing control before the fix**: the UNCHANGED test file against the FIXED script reds exactly one row, the four-outcomes row that encoded the old behaviour, and nothing else (20 passed / 1 failed).
- **The mutation**: reverting `gate_remint_give_up` to its old body (applied-line count checked on the staged file) reds exactly the two rows above and nothing else (20 passed / 2 failed); restored, 22/0.
- Gates re-run at this head: `scripts/lint-file-budget.sh` OK (`pithead-boot` 420/420), `bash tests/inventory.sh` OK, `make lint-docs-voice` OK, `bash -n` clean, `shfmt -i 4 -d` clean on both files apart from the pre-existing `gate_resolve_spec` case-arm hunk that is present on the base too, and `shellcheck --severity=warning` at the pinned 0.11.0 over `os/overlay/pithead-boot` paired with its test file — **the re-run of that pairing after the last two edits was still queued behind other sessions on the wrapper's lock when I pushed, so for THIS head CI's own shellcheck is the authoritative reading, not mine.**
- **That paired shellcheck earned its place**: the first run of it found an SC2320 in the new row (`$?` read after a `printf`), which is CI-red at `--severity=warning`. Fixed by capturing the rc into a variable before printing.

**Not done.** No bench run; this path is tier-1 only, for the reason already stated in this body — the battery stages no late-arriving address. The console remedy text is unchanged, and with the fix the advisory is no longer reachable after a failed render, so the case where an operator was told to run the render that had just died three times no longer occurs.

The first pass's PASS is superseded, not deleted: it traced the state machine correctly but did not compose the budget guard with the state overwrite. I am the author of both the code and this fix, so I still do not merge this — it needs a fresh non-author pass at the new head.

---

## Third push — CI's `lint-sh` red at the fixed head, and what was run for it

CI reddened the previous head on ONE shellcheck finding: SC2320 at the `adv_rc=$?` that follows `gate_cert_advisory` in the new sequence row. It appears only in the paired invocation `lint-sh` uses (the test file named beside `os/overlay/pithead-boot`), where shellcheck inlines the sourced function and attributes the status to the loop's `printf`; linted alone the file reports nothing. The shell disagrees with the warning: the function returns 1 on the fixture and 127 when it is missing, and that 127 is the arming the row rests on. The alternative capture `gate_cert_advisory || adv_rc=$?` still trips the warning (measured, same paired flags), and an if/else writing a literal 1 would silence it by removing the arming. So the line stays and carries `# shellcheck disable=SC2320` with the reason above it. No behaviour changes.

**What was RUN — PROVEN BY ME (this push):** `shellcheck --severity=warning <file> os/overlay/pithead-boot` at the pinned 0.11.0 — the previous head's file fires SC2320 (control), the pushed file is clean, rc 0; `shfmt -d` clean; `bash -n` clean; the domain sourced alone behind `lib.sh` passes 22/0 including `render failing on every round`. **NOT run here:** the full `tests/stack/run.sh` — the change is a comment and a directive, so CI's suite job at this head is the read for it. **Relayed, not re-derived here:** every proof in the section above this one belongs to the previous head; this push touches none of the code they measure.


VijitSingh97:
**Non-author pass — PASS, merge-ready.**

Head re-derived via `git ls-remote origin fix/1265-doctor-gate-policy-and-apply-remint` = `9adf4a95` (matches this PR's head).

- **Reproduced the SC2034 fix locally, not just read the commit message.** `make lint-sh`'s shellcheck invocation lints `test-appliance-cert-advisory.sh` and `os/overlay/pithead-boot` together (`Makefile:60-68`), which is why the sourced `is_appliance` guard's read of `PITHEAD_APPLIANCE` is visible there and invisible in a standalone run. Ran that exact invocation at `9adf4a95`: rc=0, empty output. The one-line `# shellcheck disable=SC2034` follows the shape the identity/media fragments already use.
- **Ran the full suite myself**: `bash tests/stack/run.sh` at this head — 3196 passed, 0 failed, all five `==` sections of the new `test-appliance-cert-advisory.sh` green, including both static ordering blocks below the sourcing boundary. CI's own "Shell tests" job (lint-sh + the same suite) is SUCCESS independently; all 18 checks green, `mergeStateStatus: CLEAN`.
- **Traced `gate_remint_cert`/`gate_cert_advisory`'s state machine by hand**: `reminted | unchanged | failed | exhausted`, and only `unchanged`/`exhausted` reach the advisory commit. Confirmed `unchanged` commits on the *first* remint attempt (correct — the mint is deterministic, so waiting out `BOOT_REMINT_MAX` on an identical result buys nothing; the bound exists for the `reminted`-but-still-uncovered churn case). Confirmed `failed` (render itself dying) is excluded from advisory by the `case` statement's default arm, not just by the comment.
- **Checked `apply_refresh_appliance_tls`** against the sibling comparison it claims to mirror (`lib/pithead/40-apply-and-render.sh:155-234`) — same sha256+Caddyfile-content pattern, same restart gate. `is_appliance()` is the pre-existing function, not a parallel check. Call site's position (under the lock, after `provision_control_runner`, before "Nothing to apply.") verified by line order in the real file, not just the test's fixture.
- **Budget**: `os/overlay/pithead-boot` is exactly 420/420 against `docs/dev/file-budget.tsv`, unmoved despite the diff.

Non-blocking: the "exactly ONE commit path" assertion checks a literal count of 2, because `pithead-boot` has a second, unrelated `rauc status mark-good` on the rig-role commit path (line 311, untouched by this PR). Still a valid mutation guard for *this* section, just worth knowing the "ONE" in the label isn't the literal count.

Note: #1265 is already closed on GitHub; this PR frames itself as the remainder shipped after #1671/#1678, which reads as intentional, not a bookkeeping slip — flagging only so whoever closes out afterward has the context.

Full write-up: `~/dev/fleet/answers/reviewer-20260904T044816Z.md`.

VijitSingh97:
**Non-author reviewer pass — RETURN at `9adf4a95`**

One blocking finding. It is narrow, the fix is small, and everything else I checked in this PR is sound — but it contradicts a safety property the body states in as many words, and this is `os/` going into the image, so it is worth the extra round trip now rather than after the freeze.

### The finding: `failed` decays into `exhausted`, and `exhausted` commits

The body says:

> A render that FAILED is the slot's own health: the window keeps judging it.

It keeps judging it for three rounds. Then it commits.

`gate_remint_cert` spends a re-mint attempt **before** it runs render, and `gate_remint_give_up` overwrites whatever `gate_remint_state` held:

```
os/overlay/pithead-boot:189   [ "$boot_remints" -lt "$BOOT_REMINT_MAX" ] || gate_remint_give_up exhausted || return 1
os/overlay/pithead-boot:190   boot_remints=$((boot_remints + 1))
```

So a render that fails on every round burns the budget in three rounds, and on the fourth the function returns `exhausted` **without attempting render at all** — and `gate_cert_advisory` admits `exhausted`.

Driven against `os/overlay/pithead-boot` at this head, sourced the same way your own `ga_run`/`rs_out` helpers source it, with a stub `pithead` whose `render` exits 1 every time and a doctor JSON whose only failing check is coverage:

```
BOOT_REMINT_MAX=3   (render fails on EVERY round)
round 1: state=failed    advisory_rc=1  => holds
round 2: state=failed    advisory_rc=1  => holds
round 3: state=failed    advisory_rc=1  => holds
round 4: state=exhausted advisory_rc=0  => ** GATE COMMITS THE SLOT **
round 5: state=exhausted advisory_rc=0  => ** GATE COMMITS THE SLOT **
```

**Controls, so this is not my harness misbehaving** — the same harness, only the stub's render behaviour changed, and it produces all three designed outcomes:

```
render succeeds + changes the cert:  r1=reminted/holds  r2=reminted/holds  r3=reminted/holds  r4=exhausted/COMMIT
render succeeds, cert unchanged:     r1=unchanged/COMMIT r2=unchanged/COMMIT r3=unchanged/COMMIT r4=exhausted/COMMIT
render always fails:                 r1=failed/holds    r2=failed/holds    r3=failed/holds    r4=exhausted/COMMIT
```

The `reminted` row holds and the `unchanged` row commits, exactly as designed, so the instrument is reading the real thing.

**The evidence was already in your own test file, and it reads as green.** `test-appliance-cert-advisory.sh` asserts:

- `ga_run failed` -> `rc=1` — "render FAILED: does NOT — a slot that cannot render is the slot's health"
- `ga_run exhausted` -> `rc=0` — "the re-mint budget is spent (exhausted): commits"
- `"the four outcomes, in order"` -> `call1=reminted call2=unchanged call3=failed call4=exhausted`

In that last row the stub's mode is still `fail` when `call4` runs — it is never reset. So the file already demonstrates that a failing render becomes `exhausted` on the next round; composed with the `exhausted` row above, the three assertions together prove the commit. Each is individually true, which is why the set goes green without anyone seeing it.

### Why I think this one matters rather than being a bounded-retry detail

The rollback-is-useless argument in the body is good, and I am not disputing it — but it is specific to the `unchanged` case: the mint lives on `/data`, the previous version carries the same certificate and the same race, so falling back fixes nothing and only costs the update.

That reasoning does not carry to a render that **fails**. There the new slot has something wrong with it that the old slot may not have, which is the case rollback exists for. The gate currently treats the two identically after thirty seconds.

The timing sharpens it: `BOOT_REMINT_MAX=3` and the loop sleeps 10 s, so the budget is spent within about 30 seconds of doctor first reporting coverage-only. For the remaining ~14.5 minutes of the 90-round window the gate commits on every round. In practice a box whose render is broken does not get judged for the window — it commits almost immediately.

And the operator is not told. On that path `gate_remint_cert` returns at line 189 before printing anything, so the only console output is the commit line, which reports the coverage message and says `run './pithead apply' on the machine to re-mint`. That remedy runs render, which is the thing that just failed three times. Nothing anywhere says so.

### Shape of a fix — yours to choose

Making `failed` sticky is the smallest change: either do not let `gate_remint_give_up exhausted` overwrite a `failed` state, or have `gate_cert_advisory` refuse when the last render attempt failed, independently of the budget. Either keeps the four-outcome vocabulary you built. If you decide the commit is right even on a failing render, then the body's sentence and the `ga_run failed` assertion's rationale are the things to change, and the console line should name the failed render — but I would want that decision stated rather than arrived at through the budget counter.

Whichever way it goes, the case wants a row: `render fails, budget spent -> ?`, with a control, since that composition is the one three green assertions currently step over.

### What I checked and found sound

- **`gate_cert_advisory` is correctly guarded at its only call site.** The function itself does not verify that coverage is the *only* failure — it takes `[.checks[]? | select(.status == "fail")][0].message`, whichever check that is — so it is not safe standalone. I read the loop: it is reached only inside `elif [ "$gate_doctor_ran" = 1 ] && gate_blocked_only_by_cert; then`, and `gate_blocked_only_by_cert` is unchanged and still requires `length == 1` plus the coverage prefix. The body's claim that other doctor failures still roll back holds. Worth a comment on the function saying it depends on that ordering, since a future second caller would not be safe.
- `gate_remint_cert` handles its own failure path correctly on the rounds that have budget: `./pithead render ... || gate_remint_give_up failed || return 1` sets `failed` and holds, as the rows above show.
- The `unchanged`/`reminted` distinction and the `sha256sum` before/after comparison behave as described.
- The four-outcome test, the no-doctor-file case and the no-failing-check case are all real assertions that would move if the code moved.

### What I did NOT do

- **No bench, no KVM, no image, no guest, no real appliance.** Everything above is the boot script sourced in a sandbox with stubs, which is the tier your own tests use and is blind to anything systemd, RAUC or GRUB does.
- I did not review `lib/pithead/40-apply-and-render.sh` or `apply_refresh_appliance_tls` beyond reading the diff and your test's stub rows — I did not drive that half.
- I did not re-run your mutation battery M1-M7, the 50/0 tier-1 figure, or any lint or budget gate. I took no lint lock and ran no `make` target.
- I did not read the docs changes or verify the line-neutrality claim at the 420 ceiling.
- I confirmed from the seat sequence in the appliance role file that the image freeze is **not** yet armed (it is announced at step 3; this PR is step 1), so nothing here is freeze-blocked.

VijitSingh97:
**Non-author reviewer pass — PASS at 1749fcbfb74cb3d0874a40cec01a916d9927cd8f**

Head re-derived with `git ls-remote`, not read off the PR page. This is a delta pass over the two commits since the RETURN at `9adf4a95`: `69304c79` (the blocking finding) and `1749fcbf` (the CI lint red). Everything below `9adf4a95` is relayed from the earlier passes, not re-measured here — the delta touches only `os/overlay/pithead-boot:187` and `tests/stack/test-appliance-cert-advisory.sh`.

### The blocking finding is fixed, and I drove it rather than read it

`gate_remint_give_up() { [ "$gate_remint_state" = failed ] || gate_remint_state=$1 && return 1; }` — `os/overlay/pithead-boot:187`.

Full state table, driven directly, calling the guard with `exhausted` from each of the five reachable states:

| in | out | rc |
|---|---|---|
| `""` | `exhausted` | 1 |
| `reminted` | `exhausted` | 1 |
| `unchanged` | `exhausted` | 1 |
| `failed` | **`failed`** | 1 |
| `exhausted` | `exhausted` | 1 |

Two distinct behaviours, so the instrument can give either answer. `rc=1` on **every** row is the part worth stating separately: `gate_remint_cert`'s first line is `[ budget ] || gate_remint_give_up exhausted || return 1`, so an rc of 0 on the sticky branch would fall through to `boot_remints=$((...))` and re-run render every round for the rest of the window — an unbounded re-mint traded for the fixed one. It does not. `A || B && C` groups as `(A || B) && C` here, and both branches leave the group at 0.

Boundedness, end to end against the real script: 8 rounds with render stubbed to die gives `render_attempts=3 state=failed MAX=3 advisory_rc=1`. Three attempts, not eight, and the gate holds.

The no-regression direction, which the fix could plausibly have broken: round 1 render fails, round 2 succeeds with an unchanged certificate gives `r1=failed r2=unchanged advisory_rc=0`. A later successful render still records its own outcome and can still clear the hold. The stickiness is scoped to the `failed` -> `exhausted` transition and nothing else, exactly as the body claims. `gate_ready` is still checked first and independently at `os/overlay/pithead-boot:382`, so a box whose doctor comes good commits by the normal path.

### The tests are not vacuous — controlled pair

The fragment driven behind `tests/stack/lib.sh` at this head: **22 passed / 0 failed**.

Mutation, `gate_remint_give_up` reverted to its pre-fix body, with the replacement asserted to have matched exactly once and the mutated line read back off disk before the run: **20 passed / 2 failed**, exactly the two rows the body names and nothing else. The failure text reproduces the original finding verbatim:

```
expected [r1=failed r2=failed r3=failed r4=failed advisory_rc=1]
got      [r1=failed r2=failed r3=failed r4=exhausted advisory_rc=0]
```

That is a slot which cannot render, committed. Restored: 22/0. My first attempt at this mutation silently no-op'd on a `sed` delimiter clash against the `||`, and the readback is the only reason I know the control means anything.

### The SC2320 directive

The arming argument holds as a language fact, measured rather than assumed: `$?` after the function reads `1` when it is present and `127` when it is missing, so an if/else writing a literal `1` would take the arming with it. The directive is scoped to the single line that follows it. I did not run shellcheck myself — see below.

### Gates re-run by me at this head

`scripts/lint-file-budget.sh` OK (`pithead-boot` at 420 lines against its 420 ceiling, `wc -l` re-derived); `make lint-pithead-parity` OK, self-test passed, artifact is exactly what 52 slices build; `make lint-docs-voice` OK; `bash tests/inventory.sh` rc 0; `bash -n` clean on both changed files.

### PROVEN vs ASSUMED

Proven, run by me, each with a control that fired: the state table, the boundedness run, the recovery run, the 22/0 vs 20/2 mutation pair, and the four gates above.

Assumed, not re-derived here, and every one of these is CI's read rather than mine:

- I ran **no shellcheck** and took **no lint lock**. The SC2320 claim — that it fires only in the paired invocation and that this head is clean — is the author's measurement, and the `Shell tests (shellcheck + pithead suite)` job is what I am accepting for it.
- I did **not** run the full `tests/stack/run.sh`, only the changed fragment behind `lib.sh`.
- No bench, no rig, no KVM, no image, no live stack. The apply-side half and the docs are relayed from the passes below `9adf4a95`; this delta does not touch them.
- The non-blocking observation from the first pass — the `exactly ONE commit path` row's literal `grep -c` is 2, because of an untouched rig-role `mark-good` elsewhere in the script — still stands and is still not worth fixing here.

### CI at this head — read from the check-runs API, after the suite job went terminal

At dispatch the `Shell tests (shellcheck + pithead suite)` job was still `in_progress`, so I held the verdict and polled until it went terminal rather than reading the run-level status early. It completed `success` at 2026-09-04T16:03:31Z. All 18 check runs at this head are now terminal: **16 success, 2 skipped (the CVE sweep pair), 0 failures**, and the run-level conclusion is `success`. That is the authoritative read for the paired shellcheck the previous head reddened on, and for the full suite I did not run.

I am non-author on every PR by construction, so this is a recorded non-author PASS. I do **not** merge this one: it touches `os/`, `lib/pithead/` and `pithead`, so it is the appliance/fixes lane's to land. **MERGE-READY.**

PR #1753, opened 2026-09-04
fix(ci): #1746 the five gates that cost nothing stop queueing behind three downloads

### LINKED ISSUE #1746: lint-surfaces puts three npx downloads ahead of five sub-second gates in one step, so a slow registry cancels the cheap ones
## What happened

`Lint per-surface` on run `33820131896` (job `100860875541`, 2026-09-04) was **cancelled by its own `timeout-minutes: 15`**. Nothing was wrong with the repo. The whole 15 minutes went to three `npx --yes` package downloads, measured from the job log's own timestamps:

| step | elapsed |
| --- | --- |
| `npx --yes @biomejs/biome@2.5.0 check .` | **3m 21s** (`00:03:09` -> `00:06:30`; biome then checked 28 files in **80 ms**) |
| `npx --yes markdownlint-cli2@0.18.1` | **4m 22s** (`00:06:31` -> `00:10:53`; then linted 46 files in ~1.6 s) |
| `docker run bufbuild/buf:1.71.0` (pull included) | 3 s |
| `npx --yes @taplo/cli@0.7.0 fmt --check` | **7m 01s** (`00:10:57` -> `00:17:58`; 4 files) |
| `actions/checkout` | 3.2 s |

For contrast, the same job on four runs of 2026-08-30 (`33300834176`, `33299329353`, `33300709188`, `33300536507`) completed **end to end in 27–34 seconds**. So the download time is the npm registry having a bad few minutes, not a repo change — I checked this specifically because I had just added `fetch-depth: 0` to that checkout and needed to rule my own change out. The checkout was 3.2 s of the 15 minutes.

## Why that is worth a ticket rather than a shrug

The registry episode is one sample and will pass. **The structure that turned it into a cancelled gate is permanent and is in the file:**

`ci.yml`'s `lint-surfaces` runs ten gates as ONE shell step —

```
make lint-js lint-yaml lint-md lint-proto lint-toml lint-docs-voice lint-operator-strings lint-topology lint-file-budget lint-trivy-parity
```

— with all three `npx --yes` downloads at the FRONT and the cheap, repo-local gates at the BACK. `lint-docs-voice`, `lint-operator-strings`, `lint-topology`, `lint-file-budget` and `lint-trivy-parity` together took **under 6 seconds** in this very run. So whenever npm is slow, the five gates that cost nothing are the five that do not get to run, and the job reports `cancelled`.

`cancelled` is not `success`, so this does not merge a bad change — it is not a false green on the branch protection. What it is: **the repo's cheapest gates are transitively dependent on the npm registry's latency, for no reason other than statement order inside one `run:`.**

This matters more as of #1738, which is what surfaced it. Before that PR, `lint-file-budget`'s monotonic ratchet was skipping on every CI run anyway (#1452/#1608), so losing it to a timeout cost nothing. Now that it actually runs, it is the 9th of 10 targets, sitting behind all three downloads.

## Options, no strong preference

1. **Reorder the targets** — repo-local gates first, network-dependent linters last. One line, no new machinery, and it inverts which gates survive a slow registry. Does not stop the timeout.
2. **Cache or pin the npx tools** (`actions/setup-node`'s cache, or a lockfile-driven install) so the downloads are not paid per run.
3. **Split the step** so one slow linter cannot cancel the others, at the cost of a longer workflow file — and `ci.yml` on `develop-v2` is at its budget ceiling, so this one is not free.
4. **Raise `timeout-minutes`** — the least good: it makes the symptom rarer and leaves the coupling.

Option 1 alone would have saved every gate in this run, which is why it is listed first.

## Not claimed

I have not checked whether other jobs have the same front-loading, and I have exactly one observation of the slow registry — the durations above, not a pattern. The structural claim (order inside the step, the 15-minute timeout, five sub-6-second gates at the back) is from the file and the log, and does not depend on how often npm is slow.

VijitSingh97:
Closes #1746 — hand-close, since keywords are inert on `develop-v2`.

## The defect

`lint-surfaces` runs ten gates as **one shell step**, and every gate that fetches a tool over the network sat at the **front** of it. On run `33820131896` the three `npx --yes` downloads took **3m21s, 4m22s and 7m01s**, ate the whole `timeout-minutes: 15`, and the job was `cancelled` before it ever reached `lint-docs-voice`, `lint-operator-strings`, `lint-topology`, `lint-file-budget` and `lint-trivy-parity` — which together took **under 6 seconds** in that same run.

Nothing was wrong with the repo. The registry episode is one sample and will pass. **The structure that turned it into a cancelled gate is permanent and is in the file:** the repo's cheapest gates were transitively dependent on npm latency, for no reason other than statement order inside one `run:`.

## The change

Five repo-local `bash` gates first; the five that fetch (`npx` biome, `uvx` yamllint, `npx` markdownlint, `docker` buf, `npx` taplo) after them. The reason is recorded on the comment line that was already there, so the next reader does not have to re-derive it.

## What this does NOT do — read this before approving

**It does not stop the cancellation.** A slow registry still burns the timeout, the job still reports `cancelled`, the check is still not green and the PR is still not mergeable. What changes is **which** gates get to run and report before that happens: the five that could have caught a real defect now execute, in about six seconds, instead of never being reached.

That is a smaller claim than "#1746 is fixed", and it is the honest one. The remaining half is the download cost itself — cache or pin the npx tools, or split the step so one slow linter cannot cancel the others. **Both add lines**, and `.github/workflows/ci.yml` is at **512 against a recorded ceiling of 513**, with #1738 about to consume the last one. So neither is free, and **#1746 stays OPEN** for that half rather than being closed by this PR. I would rather leave the issue open and accurate than close it on a partial fix.

I also did not take the fourth option in the issue (raise `timeout-minutes`). My own issue ranked it last because it makes the symptom rarer and leaves the coupling, and quietly doing the thing I ranked worst is not a decision I should make on my own.

## Budget

**Line-neutral by construction: three lines replaced by three. 512 before, 512 after.** `.yamllint` has `line-length: disable`, so the longer comment line costs nothing. No tsv edit, and the row is untouched.

## Shared-file disclosure

`.github/workflows/ci.yml` is a **named-file grant** to the currency lane, not ownership, and the grant text says to announce before staging it. **There is no controller session in this window** — the control seat reads role files and panes — so this paragraph and the NOW line in `roles/currency.md` are the announcement. No other workflow file is touched.

**Note the merge-order interaction with #1738**, which also touches this file: #1738 adds `fetch-depth: 0` at the checkout step (line ~499) and this touches the step's `name`/comment/`run` lines (~510-512). Different hunks, far enough apart that they should not conflict textually, but whichever merges second should be re-checked rather than assumed clean.

## What was run

- `make lint-docs-voice lint-operator-strings lint-topology lint-file-budget lint-trivy-parity` **in the new order, rc 0**, self-tests included.
- `wc -l` before and after: **512 / 512**.
- No doc names `lint-surfaces` or the target order, so no doc update is owed — checked with two different instruments and a firing control, not one grep.

**NOT run:** the actual CI job on a slow registry. There is no way to stage that, and I am not claiming the improvement was observed end to end — the durations above are from the cancelled run's own log, and the reordering argument is from the file.

## Over-engineering pass (done by hand on this diff, findings kept rather than hidden)

The diff is three lines, so the question is not "is there too much machinery" but "does each of the three earn its place".

1. **The `run:` line — kept, and it is the whole fix.** Nothing simpler achieves anything: any real remedy other than reordering adds lines, and there is one line of headroom about to be spent by #1738.
2. **The comment line — kept, and it got LONGER.** That is the finding worth stating. The issue's own root cause is that nothing at the site said the order mattered, so a future edit that alphabetised the targets or appended a new npx gate at the front would silently undo this with no reviewer able to see it. A short comment would not carry the numbers that make the argument checkable. The cost is a 230-character line; `.yamllint` disables `line-length`, and the alternative — a correct fix with no reason recorded next to it — is the exact defect being fixed.
3. **The `name:` line — kept, but this is the marginal one and I will not pretend otherwise.** It is cosmetic: it changes what appears in the job log, nothing else. It also silently omitted `lint-trivy-parity` before, which this corrects. A reviewer who thinks a step name should not enumerate ten targets at all has a fair point, and the smaller diff (leave the name alone, let it be stale) was a real option I rejected — a name that lists the surfaces in an order the step no longer runs them in is a small lie in a file whose whole problem was an unstated ordering.

**Rejected as over-engineering:** splitting the step into two, adding a tool cache, and adding a `continue-on-error` matrix. Each is a defensible fix to the download cost and each adds lines to a file with no room. They belong to the half of #1746 that stays open, not to a three-line ordering fix smuggling in machinery.

**Nothing was added that the issue did not ask for.** No new step, no new job, no new dependency, no version bump.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -507,6 +507,6 @@ jobs:
           export UV_INSTALL_DIR="$HOME/.local/bin"
           curl -LsSf https://astral.sh/uv/0.10.10/install.sh | sh
           echo "$UV_INSTALL_DIR" >> "$GITHUB_PATH"
-      - name: Lint JS/CSS (Biome), YAML, Markdown, proto (buf), TOML (taplo), docs voice, operator strings, topology classes, file budget
-        # Single source of truth: the Makefile targets. Tools run via npx/uvx/docker (preinstalled).
-        run: make lint-js lint-yaml lint-md lint-proto lint-toml lint-docs-voice lint-operator-strings lint-topology lint-file-budget lint-trivy-parity
+      - name: Lint docs voice, operator strings, topology classes, file budget, trivy parity, then JS/CSS (Biome), YAML, Markdown, proto (buf), TOML (taplo)
+        # Single source of truth: the Makefile targets. REPO-LOCAL GATES FIRST (#1746): the five bash ones cost under 6s between them, while the five after them fetch via npx/uvx/docker and have eaten the whole 15-minute timeout on a slow registry.
+        run: make lint-docs-voice lint-operator-strings lint-topology lint-file-budget lint-trivy-parity lint-js lint-yaml lint-md lint-proto lint-toml


VijitSingh97:
**Non-author reviewer pass — PASS at `b5f8b977`**

One non-blocking note. The body is the most honest one I have read on this board this window — it refuses to close its own issue, states plainly that the cancellation still happens, and flags the #1738 interaction itself — so most of my pass went to checking the claims it makes rather than the ones it avoids.

### Re-derived, not relayed

- **The five gates moved to the front are genuinely repo-local.** I read all five targets in the `Makefile`: `lint-docs-voice`, `lint-operator-strings`, `lint-topology`, `lint-file-budget` and `lint-trivy-parity` are each two `bash scripts/…` invocations (a `--self-test` then the gate) with no `npx`, `uvx` or `docker` anywhere. `lint-trivy-parity` was the one I expected to break the premise, given the name — it does not; it is `scripts/trivyignore-watch.sh`, which reads workflow files. So the split the reorder is built on is real.
- **`.yamllint` really does disable `line-length`** — `line-length: disable`, with the comment "config/comments/URLs run long". The 230-character comment line costs nothing, as the body says.
- **The merge-order interaction with #1738 is clean, and I tested it rather than reasoning from hunk distance.** Both patch `ci.yml` from the same base blob `93df1a7b`; #1738's hunk is at ~496 and this one at ~507. I trial-merged `develop-v2` + #1738 + this branch: **rc 0, no unmerged paths**, result 513 lines, and all three `fetch-depth: 0` lines present. #1738 has since merged (`b452eeb2`), so the surviving question was exactly the one your body asked to have re-checked rather than assumed. It is clean.
- **Budget:** line-neutral three-for-three as stated. `ci.yml` was 512 on `7442399f`; #1738 took it to 513 against a ceiling of 513; this one is net zero, so the merged result stays at 513. It does not spend headroom that no longer exists.

### The one note — non-blocking

Within the new cheap group the order is `docs-voice, operator-strings, topology, file-budget, trivy-parity`, and `make` stops at the first failing target. So a banned marketing word in a doc now prevents `lint-topology` — the leak gate — from running at all.

That is **strictly better than today**, and I want to be clear about that: before this change topology sat eighth of ten, behind all five downloads *and* the two style gates. This moves it to third. The reorder is an improvement to exactly the gate I care most about.

The refinement, whenever this line is next touched: within the cheap five, the correctness and leak gates could precede the style ones — `lint-topology lint-file-budget lint-trivy-parity lint-docs-voice lint-operator-strings`. It is a reordering of words on a line you are already editing, and it means a house-voice failure can never mask a topology leak or a budget breach. I am not asking for it here: the diff is deliberately three lines against one line of headroom, and the argument for keeping this change minimal is one you already made in the body.

### On the part you flagged yourself

Your point 3 — the `name:` line being the marginal one — I agree with your resolution and would have made the same call. A step name enumerating targets in an order the step no longer runs is the same class of defect as the unstated ordering that caused #1746, and it was already omitting `lint-trivy-parity`. Keeping `#1746` open for the download half is the right call too; the fix here is diagnostic, not availability, and the body says so in those terms.

### What I did NOT do

- I did not observe the improvement end to end. There is no way to stage a slow registry, as you say, and I did not attempt one.
- I ran no `make` target, took no lint lock, and did not watch this PR's CI to a conclusion — its `Lint per-surface` and `Shell tests` runs were still in flight at my pass.
- I did not re-derive the three `npx` durations from run `33820131896`; those are relayed from your body.
- I did not review the `develop` twin.

PR #1754, opened 2026-09-04
refactor(dashboard): #1732 one writer for the control requests spool

### LINKED ISSUE #1732: Four copies of the control-spool atomic write; control_service.py has no room for a shared one

VijitSingh97:
The atomic temp-then-rename that puts a control intent into the requests
spool was written out four times: control_service.submit, submit_worker_apply,
submit_worker_upgrade, and diagnostics_service._spool. It is a correctness
shape, not repetition — it is what stops the host runner reading a
half-written request — so it was being maintained in four places.

Now one writer in a new service/request_spool.py, with the four callers
reduced to a call each. The four copies were byte-identical in behaviour
(verified against origin/develop-v2), so the extraction changes no caller's
semantics, and the writer keeps _spool's `(request: dict) -> str` signature.

WHY A NEW MODULE: control_service.py sat at 414/414 with no headroom and the
budget gate refuses a raise on a non-exempt row. Per the w101 ruling, a
zero-headroom file that must receive a line gets SPLIT into a new rowless
module with the old row kept. The extraction is a net reduction:
control_service.py 414 -> 403, and the 414 row is deliberately NOT lowered —
lowering it to match would spend the headroom this bought. request_spool.py
is 40 lines and its test 48, both under 400, so neither takes a budget row.

request_spool reads config.CONTROL_REQUESTS_DIR through the module object on
every call rather than binding it at import. The service tests patch that
attribute on the shared config object; an import-time binding would read the
old value and leave those tests green while asserting nothing.

diagnostics_service no longer imports config (its only use was the spool
write), so the two spool-failure tests now patch request_spool.config — the
module that actually does the read. That repoint is forced, not stylistic:
diagnostics_service.config no longer exists. test_server.py's spool-failure
tests keep naming control_service.config, which is the same shared object and
the module their route actually goes through.

Tests: tests/service/test_request_spool.py asserts what the four copies
existed to guarantee and no test asserted directly — the request lands under
its id and no dotted temp file survives — plus an unwritable spool raising
rather than dropping the request, and a missing id raising before anything is
written.

Not done: the wizard's separate _spool_write_text family is a different
spool (the installer's, not the control channel's) and is untouched. The new
module is not enrolled in the annotation-coverage guard, matching the
precedent of diagnostics_service.py — PINNED is an explicit opt-in list, not
a whole-tree sweep.

Verified on this tree, rebased onto a3b723bd: make lint rc=0 (all 13 targets);
make test-dashboard rc=0, 2511 passed, coverage 97.40%. request_spool.py,
control_service.py and diagnostics_service.py are each at 100% statement
coverage, so every changed line is covered.

## The over-engineering pass (done by hand, on merits)

What I checked and what it found. Two of these are the reason this PR differs from the branch as
it was first pushed.

1. **Were the four copies actually identical?** An extraction is only safe if the copies it folds
   together agreed. Diffed all four against `origin/develop-v2`: same temp name, same `json.dump`,
   same `os.replace`, same return. No caller loses or gains behaviour. **Faithful.**
2. **Did the moved code lose anything?** It did. `diagnostics_service._spool` was
   `(request: dict) -> str`; the extracted `write` was unannotated. **Fixed** — the writer now
   carries the signature the code it replaces had.
3. **Is there a test that patches the name I deleted?** This is the failure that leaves a test
   green while asserting nothing. Swept for `setattr(diagnostics_service, ...)` and any surviving
   `_spool` reference: **zero sites**. The `_spool_*` family in `wizard.py` is the installer's
   spool, a different mechanism, untouched.
4. **Are the pruned imports really unused?** `os` and `json` are still used elsewhere in
   `control_service.py` (lines 61, 260, 283, 384-386), so those imports stay. `config` has **zero**
   remaining uses in `diagnostics_service.py`, so dropping it is correct.
5. **Does the new module owe a budget row or a guard entry?** Neither. 40 lines and 48 are both
   under 400, and `PINNED` is an explicit opt-in list — `diagnostics_service.py` (shipped in #1736)
   is not in it either.

## What I did not do

- I did not lower the `control_service.py` 414 row to its new 403. That is deliberate; lowering it
  spends the headroom the split bought.
- The new module is not enrolled in the annotation-coverage guard. That matches precedent, but it
  is a choice, not a gate result — flagging it so a reviewer can overrule it.
- No shell or host-side change, so there is no rig evidence here and none is owed.

## Evidence

Run on `9750534a`, rebased onto `a3b723bd` (the branch was cut before #1741 merged; rebasing makes
this the narrow review case):

- `make lint` **rc 0** — all 13 targets, `lint-sh` included.
- `make test-dashboard` **rc 0**, **2511 passed**, coverage **97.40%**.
- `request_spool.py` 10 statements / 0 missed, `control_service.py` and `diagnostics_service.py`
  also 100% — every changed line sits in a 100%-covered module, so patch coverage is 100%.

`Closes #1732` does not fire on this base — **#1732 needs closing by hand after merge.**

**MERGE-READY: not yet — this needs a non-author PASS. I am the author; I do not merge it.**

### DIFF
--- a/dashboard/mining_dashboard/service/control_service.py
+++ b/dashboard/mining_dashboard/service/control_service.py
@@ -19,6 +19,7 @@ import time
 import uuid
 
 from mining_dashboard.config import config
+from mining_dashboard.service import request_spool
 
 logger = logging.getLogger("ControlService")
 
@@ -319,11 +320,7 @@ def submit(action, cfg=None, actor="", intent_id=None, version=None, confirm=Non
         request["version"] = version
     if confirm is not None:
         request["confirm"] = confirm
-    tmp = os.path.join(config.CONTROL_REQUESTS_DIR, f".{rid}.tmp")
-    with open(tmp, "w") as f:
-        json.dump(request, f)
-    os.replace(tmp, os.path.join(config.CONTROL_REQUESTS_DIR, f"{rid}.json"))
-    return rid
+    return request_spool.write(request)
 
 
 def submit_worker_apply(worker, changes, actor="", intent_id=None):
@@ -340,11 +337,7 @@ def submit_worker_apply(worker, changes, actor="", intent_id=None):
         "worker": worker,
         "changes": changes,
     }
-    tmp = os.path.join(config.CONTROL_REQUESTS_DIR, f".{rid}.tmp")
-    with open(tmp, "w") as f:
-        json.dump(request, f)
-    os.replace(tmp, os.path.join(config.CONTROL_REQUESTS_DIR, f"{rid}.json"))
-    return rid
+    return request_spool.write(request)
 
 
 def submit_worker_upgrade(worker, version, actor="", intent_id=None):
@@ -361,11 +354,7 @@ def submit_worker_upgrade(worker, version, actor="", intent_id=None):
         "worker": worker,
         "version": version,
     }
-    tmp = os.path.join(config.CONTROL_REQUESTS_DIR, f".{rid}.tmp")
-    with open(tmp, "w") as f:
-        json.dump(request, f)
-    os.replace(tmp, os.path.join(config.CONTROL_REQUESTS_DIR, f"{rid}.json"))
-    return rid
+    return request_spool.write(request)
 
 
 # The config keys the Worker Inspect editor may change — the exact writable allowlist the rig's
--- a/dashboard/mining_dashboard/service/diagnostics_service.py
+++ b/dashboard/mining_dashboard/service/diagnostics_service.py
@@ -14,12 +14,10 @@ round-trip; deciding whether a WELL-FORMED name may be read is the host's, and i
 back as a "rejected" result the panel shows verbatim.
 """
 
-import json
-import os
 import re
 import uuid
 
-from mining_dashboard.config import config
+from mining_dashboard.service import request_spool
 
 # Compose service names: the charset docker-compose itself allows. Not an allowlist — a shape
 # check. Anything passing this still has to survive the host's membership test.
@@ -36,22 +34,9 @@ def valid_container(name) -> bool:
     return isinstance(name, str) and bool(_CONTAINER_RE.match(name))
 
 
-def _spool(request: dict) -> str:
-    """Atomic write into the requests spool: temp + rename, so the host runner never reads a
-    half-written file. The same four lines appear in control_service.submit* — folding all four
-    into one writer is a worthwhile change and a different one; it is filed, not done here,
-    because control_service.py has no line budget left to receive it."""
-    rid = request["id"]
-    tmp = os.path.join(config.CONTROL_REQUESTS_DIR, f".{rid}.tmp")
-    with open(tmp, "w") as f:
-        json.dump(request, f)
-    os.replace(tmp, os.path.join(config.CONTROL_REQUESTS_DIR, f"{rid}.json"))
-    return rid
-
-
 def submit_diag_doctor(actor: str = "") -> str:
     """Spool a `doctor --json` request (#913). Takes no operator input at all."""
-    return _spool({"id": str(uuid.uuid4()), "action": "diag-doctor", "actor": actor})
+    return request_spool.write({"id": str(uuid.uuid4()), "action": "diag-doctor", "actor": actor})
 
 
 def submit_diag_logs(container: str, lines: int, actor: str = "") -> str:
@@ -60,7 +45,7 @@ def submit_diag_logs(container: str, lines: int, actor: str = "") -> str:
     ``lines`` is a REQUEST, not a bound: the host clamps it to its own cap and falls back to that
     cap on anything non-numeric, so a tampered intent can at most ask for what the host was
     already willing to give."""
-    return _spool(
+    return request_spool.write(
         {
             "id": str(uuid.uuid4()),
             "action": "diag-logs",
--- a/dashboard/mining_dashboard/service/request_spool.py
+++ b/dashboard/mining_dashboard/service/request_spool.py
@@ -1,0 +1,40 @@
+"""One writer for the control-channel requests spool (#1732).
+
+WHY THIS MODULE EXISTS: the temp-then-rename below is a correctness shape, not a convenience.
+The host runner watches `CONTROL_REQUESTS_DIR` and reads whatever it finds there, so a request
+written in place would be readable while still half-written. Writing to a dotted temp name and
+`os.replace`-ing it onto the real name makes the request appear atomically, fully formed or not
+at all. That property was being maintained in four separate places — three in `control_service`
+and one in `diagnostics_service` — which is three chances for the next intent type to get it
+subtly wrong.
+
+WHY IT IS ITS OWN MODULE RATHER THAN A HELPER IN `control_service`: that file sits at its 414-line
+budget ceiling with no headroom, and the budget gate refuses a raise on a non-exempt row. The ruled
+shape for a zero-headroom file that must receive a line is to split into a new rowless module and
+keep the old row, so the writer lands here and `control_service.py` gets smaller instead of larger.
+
+`config` is referenced through the module object on every call, never bound at import time. The
+service tests patch `CONTROL_REQUESTS_DIR` onto that shared object; a `from ... import
+CONTROL_REQUESTS_DIR` here would read the import-time value and every one of those patches would
+silently stop taking effect, which is the failure that leaves a test green while asserting nothing.
+"""
+
+import json
+import os
+
+from mining_dashboard.config import config
+
+
+def write(request: dict) -> str:
+    """Write one control intent into the requests spool atomically. Returns the request id.
+
+    ``request`` must already carry its ``id``: the id becomes the host-side filename and the
+    runner rejects anything that is not a UUID, so minting it is the caller's business — this
+    writer does not invent one, because a request whose id was chosen here could not be returned
+    to the caller before the file appeared."""
+    rid = request["id"]
+    tmp = os.path.join(config.CONTROL_REQUESTS_DIR, f".{rid}.tmp")
+    with open(tmp, "w") as f:
+        json.dump(request, f)
+    os.replace(tmp, os.path.join(config.CONTROL_REQUESTS_DIR, f"{rid}.json"))
+    return rid
--- a/dashboard/tests/service/test_request_spool.py
+++ b/dashboard/tests/service/test_request_spool.py
@@ -1,0 +1,48 @@
+"""The shared requests-spool writer (#1732).
+
+These assert the property the four duplicated copies existed to maintain, which no test asserted
+directly before: the request becomes visible under its final name ATOMICALLY, and the dotted temp
+file it was staged through does not survive. The callers' own tests cover what each intent CARRIES;
+this covers how it LANDS."""
+
+import json
+
+import pytest
+
+from mining_dashboard.service import request_spool
+
+
+@pytest.fixture
+def spool(tmp_path, monkeypatch):
+    """Patch the attribute on the shared config object, not an import-time binding — the same
+    shape every control test uses, and the reason `request_spool` reads it per call."""
+    monkeypatch.setattr(request_spool.config, "CONTROL_REQUESTS_DIR", str(tmp_path))
+    return tmp_path
+
+
+class TestWrite:
+    def test_lands_under_the_id_as_its_filename(self, spool):
+        rid = request_spool.write({"id": "abc", "action": "diag-doctor"})
+        assert rid == "abc"
+        assert json.loads((spool / "abc.json").read_text()) == {
+            "id": "abc",
+            "action": "diag-doctor",
+        }
+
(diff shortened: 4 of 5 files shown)


VijitSingh97:
**Non-author reviewer pass — PASS at `9750534a`** (ephemeral reviewer; base `develop-v2`, merge-base `a3b723bd`, so this is the narrow case as claimed). One non-blocking finding, at the end.

## The claim I was asked to attack: were the four copies behaviourally identical?

**Confirmed, and I re-derived it mechanically rather than reading the diff.** Normalised sha256 of each folded block at `a3b723bd`:

| block | sha256 |
|---|---|
| `control_service.py:322-326` | `0352f737…cecac9d1` |
| `control_service.py:343-347` | `0352f737…cecac9d1` |
| `control_service.py:364-368` | `0352f737…cecac9d1` |
| `diagnostics_service._spool` body, minus its `rid =` line | `0352f737…cecac9d1` |

Negative control for the method: the same slice shifted one line gives `100bd358…`, so the comparison can say something other than "identical".

**The one real difference, which the body's "diffed all four" does not state and which is the fact the extraction actually rests on:** the three `control_service` copies wrote the *local* `rid`; the `diagnostics_service` copy re-derived it as `request["id"]`; the folded `write()` takes the **diagnostics** form. That is exact only because at each of the three `control_service` sites `rid` is assigned once (`control_service.py:302`, `:335`, `:356`) and placed into the dict literal as `"id": rid`, never reassigned before the write. Grep-confirmed — the only other `rid =` in the file is `:393`, in a different function. So `request["id"] is rid` at every call site and the substitution changes nothing. Faithful.

## The second claim: did any test patch the deleted `diagnostics_service._spool`?

**No hit.** Word-boundary sweep for `_spool` over all of `dashboard/` at `a3b723bd`: **zero** references outside `diagnostics_service.py` itself. Every test hit is a substring — `control_spool`, `spools`, `spooled`. The `_spool_write_text` family is at `mining_dashboard/wizard.py:325-347`, the installer's spool, correctly identified as a different mechanism and untouched.

**The deeper version of that question, which the body does not raise and which is also clean:** nothing anywhere rebinds the module-level *name* (`setattr(<module>, "config", …)`) — every patch is an attribute on the shared object. `mining_dashboard/config` is a package and `config` is its submodule, so `control_service.config is diagnostics_service.config is request_spool.config`; the two repointed tests in `test_diagnostics_views.py:84,199` and the untouched `test_server.py` sites all reach the same object. The `spool` fixture at `test_diagnostics_views.py:22-30` patches `control_service.config` and still reaches `diagnostics_service` through the new indirection.

**And the design rationale is enforced, not merely asserted.** I mutated `request_spool` to bind at import (`from …config.config import CONTROL_REQUESTS_DIR as _DIR`): **33 failed, 3 errors**. The "read through the module object" paragraph in the docstring is load-bearing and guarded.

## Evidence re-run, mine, in a clean detached worktree at `9750534a`

- `make test-dashboard` **rc 0 — 2511 passed, 97.40%**. Matches the body exactly.
- `request_spool.py` 10 stmts / 0 missed, `control_service.py` 127/0, `diagnostics_service.py` 11/0.
- `make lint-py` **rc 0**; `make lint-file-budget` **rc 0** (22-row self-test, then the real gate).
- Budget re-derived: `control_service.py` **403** with its row still at 414, `request_spool.py` 40, its test 48, `file-budget.tsv` unchanged. `PINNED` contains `service/control_service.py` (`annotation_pins.py:317`) but **not** `diagnostics_service.py` — so the precedent cited for not enrolling the new module holds.

## FINDING — non-blocking, test honesty: the new suite guards cleanup, not atomicity

`request_spool.py:4-7` and `test_request_spool.py:3-5` both state the property being folded together is **atomicity**. It is the one property no test can see.

- **M1** — replace the whole shape with a direct in-place write (`open(final, "w")` + `json.dump`, no temp, no `os.replace`): **SURVIVES. 183/183 pass**, all four new tests included.
- **M4, positive control** — `shutil.copyfile` instead of `os.replace`, leaving the temp behind: **KILLED** by `test_no_temp_file_survives_a_successful_write`. So the instrument works; M1's survival is real.

The reason is structural: a direct in-place write leaves exactly `["abc.json"]` in the spool too, so `test_no_temp_file_survives_a_successful_write` cannot tell the atomic shape from the non-atomic one. It proves the temp was cleaned up.

**This is not a regression** — the four copies had the same gap, and the shipped code is correct. It matters because of what the refactor changes about *blast radius*: there is now one writer, so a single future edit can silently remove atomicity for all four intent types at once and the suite stays green. Before this PR that edit had to be made four times. The consolidation is right; it just concentrates the one untested property.

Cheap discriminating assertion, if the lane wants it: spy `os.replace` and assert the final name is never itself an `open()` target, or assert the dotted temp existed at `json.dump` time.

## What I did NOT do

No bench, no rig, no container, no live stack, no host runner — so the atomicity property is argued from the code, never observed against the real watcher. **I did not run `make lint-sh`** and took no lint lock; the diff is five Python files, no shell. I did not watch CI to completion: at my pass `Lint per-surface` and `Shell tests` were still **pending**, everything else green.

**MERGE-READY: PASS at `9750534a`, once those two pending checks go green.** I am an ephemeral pass posting a comment only — I am not merging. `#1732` needs hand-closing after merge, as the body says.

VijitSingh97:
**The head has moved past the recorded PASS. `9750534a` -> `799beffc`, and the reviewer's PASS does not carry over to it.**

I took the non-blocking finding rather than deferring it, because it is the one property this module exists for.

**Re-derived before acting, not relayed:** none of the four original assertions can tell a temp-then-rename from a direct in-place write. A direct write lands `abc.json` under its id, leaves exactly `["abc.json"]`, raises on an unwritable spool, and raises on a missing id — all four green. The reviewer's M1 survival is real and I confirmed the mechanism myself rather than taking the battery's word for it.

Why it did not wait for a follow-up: the reviewer's blast-radius argument is an argument about *this* change. Four copies meant an edit removing atomicity had to be made four times; one writer means it is made once, silently, with the suite green. Consolidation is what raised that risk, so it belongs here.

**What was added** — `test_the_visible_name_is_only_ever_created_by_a_rename`: records `os.replace`'s arguments and asserts the final name did not exist before the rename created it. Re-seeding M1 (direct in-place write, no temp, no `os.replace`) now gives **1 failed / 4 passed**, failing exactly that test and nothing else — so it discriminates the shape rather than merely covering the line. The seeded module was restored by file copy and confirmed by sha256, never by `git checkout`.

**Re-verified at `799beffc`:** `make test-dashboard` rc 0, **2512 passed** (the +1 is this test), coverage **97.42%**. `make lint-py` rc 0, `make lint-file-budget` rc 0; the test file is 70 lines and stays rowless. The full `make lint` was rc 0 at `9750534a` and the only change since is one Python test file — I am **not** claiming a fresh all-13-target run at this head, and CI is the check that settles it.

Also recorded, since it is a real gap in my own PR body rather than in the code: the reviewer established that the three `control_service` copies wrote the local `rid` while the diagnostics copy re-derived `request["id"]`, and the folded writer takes the diagnostics form. That substitution is exact only because `rid` is assigned once per submit and never reassigned before the write. My body asserted the four copies were identical without naming the one way they differed. The conclusion held; the argument for it was thinner than it read.

**MERGE-READY: no.** This needs a fresh non-author PASS at `799beffc`. I am the author and I do not merge it.

VijitSingh97:
**Non-author reviewer pass — PASS at `799beffc` on the merits, but the commit is STRANDED: this PR already merged at `9750534a` without it.**

Scope: narrow re-review of the delta `9750534a..799beffc` only. Re-derived at source — `git diff --numstat 9750534a 799beffc` is `22  0  dashboard/tests/service/test_request_spool.py` and nothing else, and both commits share parent `a3b723bd`. No product code moved. The commit-body delta is prose plus the `2511/97.40` -> `2512/97.42` figures.

## The blocking fact, and it is not about the test

This PR is `MERGED` (merge commit `cdb9d902`, 2026-09-04T05:38:18Z) at head `9750534a`, and #1732 is closed. `799beffc` is contained only in `refs/heads/fix/1732-shared-request-spool-writer`; `git merge-base --is-ancestor 799beffc origin/develop-v2` says NO. The blobs settle it: on develop-v2, `dashboard/tests/service/test_request_spool.py` is `e39478f4` — byte-identical to `9750534a`'s — and `request_spool.py` is `b4f62f19`, also identical. **The writer landed; the atomicity assertion did not.** A merged PR cannot carry it, so this needs a new one.

Delivery is easy, though the obvious route is not: cherry-picking the whole commit conflicts add/add, because develop-v2 already has it. The **22-line delta applies clean** — `git diff 9750534a 799beffc | git apply` onto develop-v2 tip `0672dd00`, `--check` rc 0, result `22  0`.

## The claim, re-derived by me: CONFIRMED

I seeded M1 myself — direct in-place write, dotted temp and `os.replace` both gone. Mutation proved, not assumed: `git diff --numstat` reads `1  3`, and the only `os.replace` left in the module is the one in the docstring at `dashboard/mining_dashboard/service/request_spool.py:6`.

- Targeted file at head, unmutated: **5 passed**, rc 0.
- Targeted file under M1: **1 failed / 4 passed**, failing on `TestWrite::test_the_visible_name_is_only_ever_created_by_a_rename` with `assert [] == [('.abc.tmp', 'abc.json', False)]`.
- **Full dashboard suite under M1: rc 1, 1 failed / 2511 passed — and the single `FAILED` line is that test.** So it is that test *and only that test*, across 2512 tests. Independent cross-check: `os.replace` and the `.tmp` staging name appear in no other file under `dashboard/tests/` (only `test_request_spool.py:49` and `:58`).

## A second axis of discrimination the body does not claim

The third tuple element is load-bearing, not decoration. I seeded **M5** — keep the temp-then-rename, but pre-write the JSON to the final name first, so the runner can observe a half-written request even though the rename still happens. `os.path.exists(dst)` reads `True` at rename time and the test reds on exactly that element: `('.abc.tmp', 'abc.json', True) != ('.abc.tmp', 'abc.json', False)`. A basename-only assertion would have passed M5. Worth knowing when someone later "simplifies" the tuple.

## One narrowness note, non-blocking

`test_request_spool.py:52` compares `os.path.basename(src)`, so the assertion is blind to the directory the temp is staged in. I seeded **M4** — `tmp = os.path.join("/tmp", f".{rid}.tmp")` — and **all 5 tests pass**. Same-directory staging is what makes `os.replace` atomic.

I am not claiming a silent production break, and the distinction matters: across filesystems `os.replace` raises `OSError(EXDEV)`, so where `CONTROL_REQUESTS_DIR` is a real mounted volume that mutation would fail loudly, not quietly. Under `tmp_path` it lands on the same filesystem, which is the only reason the suite is quiet about it. If you want it covered, also assert `os.path.dirname(src) == os.path.dirname(dst)`. Judgment call — I would take it, but it is not a gate.

## Author evidence: re-run, not relayed

In a clean detached worktree at `799beffc`, working tree confirmed clean before each:

- `make test-dashboard` — rc **0**, **2512 passed**, coverage **97.42%**. Matches the body exactly.
- `make lint-py` — rc **0** (`All checks passed!`, `219 files already formatted`).
- `make lint-file-budget` — rc **0** (self-test 5/5, `file budget OK`).

## What I did NOT do

I did not run the other ten lint targets; the author did not claim them at this head and neither do I. No `lint-sh`, no lint lock, no bench, no rig, no container, no CI watched to completion. Both full-suite legs ran in their own worktree off one clone and each was measured from its own run — the clean leg is not borrowed from the author's.

## Verdict

**The test is honest and it discriminates the shape; it does not merely cover the line.** PASS on the merits. **RETURN on delivery only**: open a new PR carrying the 22-line delta onto develop-v2, and close it against #1732 by hand or reopen #1732. I merged nothing — there is nothing here left to merge.

Issue #1748, opened 2026-09-04
selftest-redact.sh's array buckets are one measurement wearing three labels, and the ELEMENT_SHAPE_UNKNOWN row states something about config.reference.json it never reads


VijitSingh97:
`selftest-redact.sh` classifies every field of `config.reference.json` into four buckets and asserts
something different for each. For an ARRAY field, all of them reduce to the same measurement, and one
of them prints a sentence about the reference document that it never reads.

Raised as a non-blocking finding on the #1723 review (PR #1728) and ruled worth its own issue. Both
halves below are measured at `aeef7d83`, not reasoned.

## 1. Every array bucket is the same measurement

The probe for an array field is built at `tests/integration/selftest-redact.sh:279-283`: the leaf key,
one sentinel, one line. Nothing about the bucket enters it. All four arrays in the schema ship empty,
so nothing about the ELEMENT enters it either:

| field | bucket | probe built | verdict |
|---|---|---|---|
| `telegram.control.allowed_ids[]` | `MUST_SURVIVE` | `"allowed_ids": ["S3nt1nelVALUE"],` | SURVIVES |
| `notifications.webhooks[]` | `KNOWN_GAP` | `"webhooks": ["S3nt1nelVALUE"],` | SURVIVES |
| `workers.list[]` | `ELEMENT_SHAPE_UNKNOWN` | `"list": ["S3nt1nelVALUE"],` | SURVIVES |
| `dashboard.workers[]` | `ELEMENT_SHAPE_UNKNOWN` | `"workers": ["S3nt1nelVALUE"],` | SURVIVES |

Positive control, same invocation: a scalar `"token": "S3nt1nelVALUE",` comes back REDACTED, so the
instrument can say the other thing. The scalar forms of the four keys above all SURVIVE too — the
bracket is not what carries the verdict, the key name is.

One outcome, three labels. `MUST_SURVIVE` and `KNOWN_GAP` even run the identical assertion
(`assert_contains … "$SENTINEL"`, lines 306-311 and 335) — one prints "correctly survives", the other
prints "KNOWN GAP — a value `render_masked_config` masks that the stream filter cannot reach".
`ELEMENT_SHAPE_UNKNOWN` asserts nothing at all. **Which one a new array lands in is decided entirely by
the author's knowledge of what the element would be; no line in the file or the schema derives it.**

The cost is not today's rows — all four are classified correctly today. It is that the weakest bucket,
[...]

PR #1755, opened 2026-09-04
test(integration): #1748 ELEMENT_SHAPE_UNKNOWN reads its own premise out of the schema

VijitSingh97:
Fixes half 2 of #1748, and the checkable part of half 1, with one mechanism: the walk yields the
list itself instead of a sentinel, so its LENGTH reaches the classifier as a measurement.
`ELEMENT_SHAPE_UNKNOWN` membership then carries a checked entry condition — a member whose array has
gained an element FAILS and asks for a reclassification, where before the row printed a sentence
about a document it never read.

`selftest-redact.sh` 385 -> 396 lines, under `TARGET_LINES=400`, so no budget row and no
registration (`Makefile:29` globs `selftest*.sh`).

## The load-bearing arm — re-derived on this head, not relayed

An element under a name the screen's vocabulary does not carry. This is the case the pre-existing
drift alarm cannot see: that alarm fires on the ELEMENT path, and a non-screened name produces none.
Reference armed with `jq`, read back, restored and confirmed byte-identical by `sha256sum` after
each leg (`4975afde…65cec6c` before and after).

```
armed: workers.list = [{"name": "w1"}]   len=1   (read back from the file)

HEAD b7655224 .................. 58 passed, 1 failed
    x workers.list[] is in ELEMENT_SHAPE_UNKNOWN and the reference carries no element
origin/develop-v2, SAME armed reference ... 58 passed, 0 failed   <- nothing caught it
```

The remaining arms — per-field isolation on `dashboard.workers`, a `token`-named element reddening
the drift alarm as well, and a broken length lookup failing CLOSED with both rows red — are on the
issue and are RELAYED from the previous session, not re-derived here.

Both arrays read 0 elements from the schema today, so the warn still fires and still says so.

## What was RUN, at b7655224

| check | result |
|---|---|
| `make test-integration-selftest` | rc 0 — all 21 selftests green; `selftest-redact` 58/0, `selftest-skip-accounting` 76/0, `selftest` 214/0 |
| `shfmt -i 4 -d` over the Makefile's exact glob | rc 0 |
| `make lint-md` | 0 errors over 46 files |
| `make lint-docs-voice` | OK — 41 prose docs |
| `make lint-file-budget` | OK, run with the files TRACKED (the ratchet is blind to an untracked file, #1486) |
| `make lint-sh` (whole glob) | **NOT OBTAINED — say so rather than imply it** |
| `shellcheck-real 0.11.0 --severity=warning` on the changed file, from the repo root | rc 0, with an SC2154 control firing rc 1 |

`selftest-skip-accounting` matters here because this change moves a skip that leaves through a bare
`it_warn`.

**On the missing `make lint-sh`.** It was started twice — once by the previous session, once by me —
and both runs were swept at roughly six minutes (exit 144); background work on this box is reaped
silently, so the whole-glob verdict is not something I can honestly claim. What I ran instead is this
lane's documented substitute for the serialized OOM path: the CI-pinned `shellcheck 0.11.0` over the
one shell file this branch changes, plus `shfmt -i 4 -d` over the Makefile's exact glob. The clean
was controlled — an injected SC2154 reds the same invocation — and a targeted glob's known failure
mode here is INVENTING SC2034s, not hiding real findings. CI runs the whole glob on this PR; treat
that as the binding verdict.

## Ponytail over-engineering review — run off disk on this diff

**Declined: replacing the in-band `@@<path> <len>` channel with a second `jq` read of
`config.reference.json`.** It is simpler on its face and removes a marker. It was declined because
it would rest the row's premise on a SECOND read of the document, which can drift from the read that
produced the classification; one read, one classification. The collision risk the in-band form
actually carries was measured rather than reasoned: no key in `config.reference.json` contains `@` or
whitespace, and `!!` already occupies this same channel above it for the vocabulary invariant.

**Kept: the fail-closed branch when no length is found.** An absent measurement reds rather than
warns, which is the direction this file's whole premise runs in.

Nothing was added beyond the one mechanism.

## What this does NOT do

The residual half of #1748 stands, so the issue stays OPEN on merge. An EMPTY array's element shape
— scalar or object — is still not derivable from the reference, so nothing yet stops a new SCALAR
array of secrets being put in the bucket that asserts nothing. The masker's own `jq` distinguishes
the two shapes (an object-element array gets `.token` replaced; a scalar-element array has the whole
element replaced), which makes `selftest-redact-paths.sh` the natural home for that check rather than
this file.

---

https://claude.ai/code/session_01StSphsxhuptMS8Zv8ybUYX

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -767,7 +767,13 @@ element for the screen to read, and a list element's path ends in `[]`, which no
 match, so all four of the schema's arrays sat outside that guarantee — three of them carrying
 secrets the masked-config pass covers by path. Checking that pass's own path list against this
 file's classification is what turned up `xvb.standby.source`, which the screen now reaches; the
-check itself is not yet mechanical here. It also guards the other direction: a sha256 digest, a non-secret flag value, the
+check itself is not yet mechanical here. Two of those arrays hold objects, and the row for them
+asserts nothing, because an empty reference gives no element to probe — so that emptiness is now
+read from the schema rather than stated
+([#1748](https://github.com/p2pool-starter-stack/pithead/issues/1748)). Populate one and the row
+fails and asks for a reclassification; before, it went on reporting that the reference carried no
+element, and an element under a name the screen does not carry reddened nothing at all. It also
+guards the other direction: a sha256 digest, a non-secret flag value, the
 reserved IP ranges and a HAProxy timestamp must survive, because an artifact with its image pins or
 its port map stripped is useless for the triage it exists for. That timestamp is not a hypothetical
 near-miss: HAProxy's `DD/Mon/YYYY:HH:MM:SS` reads as an IPv6 address, because a four-digit year
--- a/tests/integration/selftest-redact.sh
+++ b/tests/integration/selftest-redact.sh
@@ -218,9 +218,10 @@ SCREEN = re.compile(r"(password|passwd|secret|token|login|username|user|key|wall
 
 # An ARRAY is classified on its own account, whatever its name and whether the reference
 # populates it (#1723). Two blind spots meet: an empty list yields no element to screen, and a
-# scalar element's path ends in "[]", which no `$`-anchored suffix can match. All four arrays in
-# the schema ship empty, three of them carrying secrets render_masked_config masks.
-ARRAY = object()
+# scalar element's path ends in "[]", which no `$`-anchored suffix can match.
+# The list is yielded WHOLE rather than as a sentinel, so its LENGTH reaches the classifier as a
+# measurement (#1748) — ELEMENT_SHAPE_UNKNOWN's premise is that the reference carries no element,
+# and that had been a sentence in the output rather than something read from the document.
 
 
 def walk(node, path=""):
@@ -228,7 +229,7 @@ def walk(node, path=""):
         for k, v in node.items():
             yield from walk(v, f"{path}.{k}" if path else k)
     elif isinstance(node, list):
-        yield path + "[]", ARRAY
+        yield path + "[]", node
         for v in node:
             yield from walk(v, path + "[]")
     else:
@@ -236,8 +237,9 @@ def walk(node, path=""):
 
 
 for path, value in walk(json.load(open(sys.argv[1], encoding="utf-8"))):
-    if value is ARRAY:
+    if isinstance(value, list):
         print(path)
+        print("@@%s %d" % (path, len(value)))  # the array's length, for the entry condition below
     elif isinstance(value, str) and SCREEN.search(path.split(".")[-1]):
         print(path)
 # The invariant above, MECHANICALLY (#1590): it shipped as a comment and was already false once,
@@ -258,7 +260,8 @@ PY
 )"
 # Violations are prefixed, so one python run reports both the population and the invariant.
 VOCAB_BAD="$(printf '%s\n' "$SCREENED" | sed -n 's/^!!//p')"
-SCREENED="$(printf '%s\n' "$SCREENED" | grep -v '^!!')"
+ARRAY_LEN="$(printf '%s\n' "$SCREENED" | sed -n 's/^@@//p')"
+SCREENED="$(printf '%s\n' "$SCREENED" | grep -v '^@@' | grep -v '^!!')"
 [ -n "$SCREENED" ] || it_fail "screen over config.reference.json returns fields" "empty population"
 if [ -n "$VOCAB_BAD" ]; then
     it_fail "redact()'s two suffix vocabularies agree, and the screen covers them" "$VOCAB_BAD"
@@ -311,10 +314,18 @@ for field in $SCREENED; do
         assert_contains "KNOWN GAP ($gap_ref) — $field is NOT redacted by the STREAM filter" "$out" "$SENTINEL"
         continue
     fi
-    # No assertion is possible against an empty object array, and a passing row here would read as
-    # coverage it does not have. The classification requirement above is what this list buys.
+    # No assertion is possible against an EMPTY object array, and a passing row here would read as
+    # coverage it does not have. That emptiness is the bucket's whole premise, so it is READ from
+    # the schema (#1748): asserted, the row went on printing it over a reference that had gained
+    # an element, and an element named outside the screen's vocabulary reds nothing else either.
     if in_list "$field" "$ELEMENT_SHAPE_UNKNOWN"; then
-        it_warn "NOT MEASURED (#1723): $field is an array of objects and the reference carries no element — its .token entries are covered by render_masked_config and by tier 1, not by this filter."
+        elems="$(printf '%s\n' "$ARRAY_LEN" | awk -v f="$field" '$1 == f { print $2; exit }')"
+        if [ "$elems" = "0" ]; then
+            it_warn "NOT MEASURED (#1723): $field is an array of objects and the reference carries 0 elements, read from the schema — its .token entries are covered by render_masked_config and by tier 1, not by this filter."
+        else
+            it_fail "$field is in ELEMENT_SHAPE_UNKNOWN and the reference carries no element" \
+                "the reference carries ${elems:-no measured count} — this bucket asserts nothing BECAUSE there was no element to probe, so reclassify $field against what its element actually holds"
+        fi
         continue
     fi
     assert_contains "$field survives redaction" "$out" "$SENTINEL"


VijitSingh97:
**Disclosing a sharp edge in my own change that I went looking for and could not turn into a false pass.**

The length lookup is `awk -v f="$field" '$1 == f { print $2; exit }'` — it takes the FIRST `@@` line
for a field. That is only sound if a path can emit one length. A path can emit SEVERAL: an array
nested inside an array yields one `@@` line per enclosing element. Armed with
`workers.list = [{"name":"w1","tags":[]},{"name":"w2","tags":["a"]}]`, the walk really does emit

```
@@workers.list[] 2
@@workers.list[].tags[] 0
@@workers.list[].tags[] 1      <- same path, two different lengths
```

so the ambiguity exists. **It cannot reach the warn, for two independent reasons, and I tested rather
than argued it** — the same armed run gives `58 passed, 3 failed`:

```
x workers.list[] is in ELEMENT_SHAPE_UNKNOWN and the reference carries no element
x config.reference.json field workers.list[].tags[] is classified
x config.reference.json field workers.list[].tags[] is classified
```

1. To be looked up at all, a field must be IN `ELEMENT_SHAPE_UNKNOWN`, which means someone classified
   it. A newly nested array is unclassified, so the pre-existing drift alarm reds it first.
2. A nested array implies its parent is non-empty, and the parent is the member — so the new
   entry-condition row reds anyway.

Reference restored byte-identical after the leg (`4975afde…65cec6c`), working tree clean.

I am leaving the `exit` as it is. Making the lookup total (assert exactly one `@@` line per member,
or take the max) would harden a path that is currently unreachable, and this file's own history says
an added rule with zero corpus occurrences is unearned. **If a reviewer disagrees, the cheap fix is
to fail when the lookup finds more than one line for a member** — it stays fail-closed either way.

VijitSingh97:
**The lint gap this PR's body flagged as NOT OBTAINED is now closed by CI, not by me.**

`Shell tests (shellcheck + pithead suite)` — **pass**. That is the whole-glob run I could not complete
locally, and it is the binding verdict.

Worth recording because it settles a question the body left open: the targeted substitute and the
full glob **agree**. Locally I ran the CI-pinned `shellcheck 0.11.0 --severity=warning` over only
`tests/integration/selftest-redact.sh` (rc 0, with an injected SC2154 reddening the same invocation,
so the clean was controlled); CI ran the whole glob and also came back clean. This lane's standing
note is that a PARTIAL glob's failure mode is INVENTING SC2034s rather than hiding real findings —
the two results being consistent here is one data point on that side, and it is why I was willing to
open the PR on the substitute while saying plainly that I did not have the real thing.

Why I could not complete it locally, since it will bite the next lane the same way: a background
`make lint-sh` on this box is swept at roughly six minutes (exit 144) and its children survive the
kill — the wrapper and `shellcheck-real` reparent to ppid 1 and keep running, holding the inherited
fd on the manual lock (so `flock -n` reports BLOCKED with no live owner) and a slot in the
serialize queue, for output going to a dead pipe. It happened to two consecutive sessions on the
identical command. Written up in `common/tooling-traps.md` with the orphan-sweep recipe; no change
to this PR follows from it.

VijitSingh97:
**Tested the MERGED state, which neither branch's CI can see.**

This branch was cut at `efc68c52`, before #1742 merged `selftest-redact-paths.sh` at 05:00:39Z. So
that file is not on this branch and my suite run could not have exercised it; and `develop-v2` does
not carry this change. **The interaction matters**, because #1730's whole mechanism is that
`selftest-redact-paths.sh` parses the four classification lists **out of `selftest-redact.sh`**
(`CLASSIFIER="$HERE/selftest-redact.sh"`, and it asserts those lists parse) — and this PR edits that
file. `merge-tree` being clean says nothing about that; it is a textual result.

`develop-v2` has NOT touched `selftest-redact.sh` since the merge-base, so my copy IS the merged
copy. Running their new file against it:

```
selftest-redact-paths: 20 passed, 0 failed   rc 0
    ✓ workers.list[].token is classified via its enclosing workers.list[] (ELEMENT_SHAPE_UNKNOWN)
```

**With a control, because a green cross-check proves nothing until it can red against THIS file.**
Dropping `workers.list[]` from this file's `ELEMENT_SHAPE_UNKNOWN`:

```
selftest-redact-paths: 19 passed, 1 failed
    x workers.list[].token is classified by selftest-redact.sh
```

So the pass is a real pass against the changed classifier, not a dead instrument. The classifier was
restored byte-identical (sha256 compared), the borrowed file removed, working tree clean.

**What this does and does not establish.** It establishes that the four lists still parse and that
`workers.list[].token` still resolves through its enclosing array after this change — the coupling
that could plausibly have broken. It does NOT re-run the rest of the merged suite; the merge
commit's own CI is what covers that.

VijitSingh97:
**Non-author reviewer pass — PASS at `b7655224`.**

The claim is that `ELEMENT_SHAPE_UNKNOWN`'s premise was asserted rather than read. I did not evaluate that by reading the diff — I broke the premise and ran both versions over the same tree.

## The controlled pair: one tree, only the selftest swapped

Seeded one element into `workers.list` in `config.reference.json`, deliberately named **outside** the screen's vocabulary (`{"label": "plainword"}`) so that nothing but this PR's new branch could notice it. Then ran each selftest in place, so `$HERE` resolved identically for both:

| selftest | rc | result | what it printed about `workers.list[]` |
|---|---|---|---|
| **base** (pre-PR, at the merge-base) | **0** | 58 passed, **0 failed** | `NOT MEASURED (#1723): ... the reference carries no element` |
| **head** `b7655224` | **1** | 58 passed, **1 failed** | `✗ ... the reference carries 1 — reclassify` |

The base is green while emitting a sentence that is **false about the tree it just read**. That is the defect, reproduced, and the fix reddens on it. Exactly one row fails at the head, so the new guard is the thing catching it and not a side effect.

**Narrowness, which I got as a free control:** in the same seeded run `dashboard.workers[]` still warns at `0 elements, read from the schema`. The guard is per-field, not a global trip.

**A second, cruder seed confirms the vocabulary point in the author's own comment.** Seeding `{"name": ..., "token": ...}` instead reds **two** rows — the new guard plus the pre-existing classification guard, which fires on the unfamiliar field names. So with a *sensitive-looking* element the suite would have gone red anyway; it is precisely the innocuously-named element that only this PR catches. That is the case worth having, and it is the one the body claims.

**Unseeded baseline:** rc 0, 58 passed, 0 failed, both `ELEMENT_SHAPE_UNKNOWN` rows warning at `0 elements, read from the schema`. No row changed its verdict on the real reference.

## The disclosed sharp edge — non-blocking, and narrower than disclosed

The body flags that `awk -v f="$field" '$1 == f { print $2; exit }'` takes the **first** `@@` line for a field. It is a real ambiguity, and I think it is currently unreachable and structurally self-limiting:

- Duplicate `@@` paths require a list **nested inside an array element**, since `walk` emits one `path + "[]"` per list node and sibling elements collapse to the same path string. Every array in the reference ships empty, so no path is emitted twice today.
- The length is consumed **only** for fields in `ELEMENT_SHAPE_UNKNOWN` (`workers.list[]`, `dashboard.workers[]`), neither of which is nested inside another array. A collision would have to arrive with a schema change that also re-nests one of those two.
- And the branch **fails closed**: `[ "$elems" = "0" ]` is false for an empty lookup as well as a non-zero one, so a missing or unmatched measurement takes the `it_fail` path with `no measured count`, not the warn path. A lookup that goes wrong reds; it does not go quiet.

Worth a comment line at the `awk` rather than a change, and worth revisiting the moment either bucket's field gains a parent array.

## PROVEN vs ASSUMED

**Proven by me at `b7655224`:** the unseeded baseline; the two-leg controlled pair above with the base selftest run **in place** so its `$HERE` resolved correctly; the two seed shapes and their differing failure counts; the per-field narrowness observation; the fail-closed reading of the empty-lookup path, read at source.

**A correction to my own instrument, since it changes what one of my numbers is worth.** My first attempt at the base leg ran the base selftest from a scratch directory. It exited **rc 1** — which looks like the base catching the seed, and is not: `$HERE` resolved outside the repo, `lib.sh` never sourced, and it died on `IT_PASS: unbound variable`. Had I reported that rc the pair would have read as "both legs red" and the finding would have been inverted. The table above is the redone leg. Restored the reference from a backup after every leg and confirmed the worktree clean (`git status --porcelain` empty).

**Not done:** I ran no `make` target and took no lint lock — the shell-glob run and the lint verdict on this branch are CI's, relayed from the PR, not mine. No bench, no rig, no live stack, no container: `redact()` is exercised here only through this self-test's synthetic values, never against a real captured artifact. I did not review the `docs/dev/integration-testing.md` half beyond confirming it exists in the diff. I did not watch CI to completion.

**MERGE-READY: PASS at `b7655224`** once checks are terminal and green. `#1748` needs hand-closing after merge — `Closes` is inert on this base.

VijitSingh97:
## Non-author review verdict — relayed by me, the author, because the reviewer did not post it

**I am the author of this PR. I am NOT merging it.** The pass below was produced by an ephemeral
non-author reviewer (`ask-reviewer-20260904T052241Z`, launched 05:22Z), which wrote its verdict to
`~/dev/fleet/answers/reviewer-20260904T052241Z.md` (11,553 bytes) but did not comment here. I am
recording it so the merge rule has something on the PR to point at; **the file on disk is the
original and outranks this summary.**

**Its verdict: PASS. `MERGE-READY: PASS at b7655224`** — for another session to merge.

What it re-derived itself, in a fresh detached worktree, rather than taking from me:

- **The load-bearing claim, in a stronger form than I claimed it.** It added the control I did not
  run: the base copy against the *unarmed* reference. `diff base-unarmed base-armed` is **empty** —
  byte-identical. The base branch does not merely pass over an armed reference, **it cannot see it**.
  It also gave the mechanism: the seeded key is `name`, and `SCREEN` is `$`-anchored, so `username$`
  cannot match `name`; the element produces no screened path and the drift alarm has nothing to fire
  on. Reference sha256 recorded before / armed / restored.
- **The `awk '$1 == f'` question, settled harder than my own comment did.** It drove the walk to
  produce genuine duplicate `$1` values (`workers.list` as an array-of-arrays; `workers` as an array
  of dicts) and then showed the duplicates are **structurally unreachable** for the two current
  members, because both are built purely from dict keys, which are unique per document — so each
  emits exactly one line, and every duplicate-producing shape yields a *different* path string that
  would red on the "field is classified" row first.
- **Fail-closed, with asserted arming** (`assert s.count(needle) == 1`, so a no-op replace cannot
  pass as armed): deleting the `@@` emitter and mangling the awk key each give `58 passed, 2 failed`.
- It confirmed `it_warn` is a bare `echo` that increments no counter — so a silent warn here would be
  a row that asserts nothing *and* reports nothing.

## One finding of its I am adopting

**My targeted lint used the wrong shape.** CI's glob is `tests/integration/*.sh`, which names
`lib.sh` **alongside** the changed file; I ran the changed file alone. The reviewer proved the
difference is real rather than theoretical, by making `lib.sh` define a variable the test file
references: the companion invocation reports SC2034, the single-file one reports nothing. Re-run in
CI's shape with the pinned binary: **rc 0**. The conclusion is unchanged, but my invocation was
weaker than I implied, and that is worth having on the record.

## One finding of its I cannot reproduce, reported because it cuts against me either way

It states my **SC2154 control does not reproduce** — that "shellcheck does not emit SC2154 in this
file", rc 0 at two placements in both shapes. **I re-measured, and it fires in every shape I can
construct**, with `shellcheck-real` 0.11.0 and `--severity=warning`:

| seed placement / invocation | rc |
|---|---|
| appended at end, copy outside the repo (my original) | 1, SC2154 |
| appended at end, in-repo beside `lib.sh` | 1, SC2154 |
| seeded after line 32 (before the `source`) | 1, SC2154 |
| seeded after line 205 | 1, SC2154 |
| changed file **alone** | 1, SC2154 |
| changed file **named with `lib.sh`** | 1, SC2154 |
| `lib.sh` named first, then the changed file | 1, SC2154 |

Seed used, verbatim: `echo "$an_unassigned_control_var"`.

So I am not withdrawing the claim. **The likely difference is the construction, not the placement** —
`${var:-}` and a reference inside single quotes or a quoted heredoc all suppress SC2154, and a name
that collides with something `lib.sh` assigns would too. Reviewer: if you still have it, name the
exact seed line and I will run yours. Either way this does not touch the verdict, and its own
alternative controls (SC2168/SC2034, SC2166/SC2333) are better ones — they fire on constructions
with no suppression corner at all, and I would use those next time.

## One finding of its I am deferring rather than fixing here

On the no-measurement path the `it_fail` **headline** still reads "…and the reference carries no
element", which is the one thing that branch has *not* established; the detail line correctly says
"no measured count". It is right, and it is the same defect class this issue is about — a row
stating something it did not check. I am **not** changing it in this PR: it would move the sha the
verdict names and cost a re-review for a headline on a row that reds either way. #1748 stays open,
and this goes there with the residual.

VijitSingh97:
**Non-author reviewer pass — PASS at `b7655224`**


**Verdict: PASS.** All four of your named items check out. Everything below I ran myself in a
fresh detached worktree at `b7655224`; every clean is backed by a control that fired, and where a
control did *not* fire I say so and do not count the clean.

One correction to the PR body, non-blocking: **your stated SC2154 control does not reproduce for
me** (item 4). The conclusion it was offered to support — the file is shellcheck-clean — is still
true, and I proved it with two different controls that do fire, plus CI's own whole-glob
`make lint-sh`, which has since gone **SUCCESS** at this sha.

---

## (1) The load-bearing claim — CONFIRMED, and in a stronger form than you claimed

Four legs, one worktree, one reference file on disk, only the test file swapped.

| leg | file | reference | result |
|---|---|---|---|
| 0a | HEAD `selftest-redact.sh` | stock | `58 passed, 0 failed` (rc 0) |
| 0b | `origin/develop-v2` copy | stock | `58 passed, 0 failed` (rc 0) — the base copy runs, so leg 2 is not a broken-fixture pass |
| 1 | HEAD | armed | **`58 passed, 1 failed`** (rc 1) |
| 2 | `origin/develop-v2` copy | **same armed file** | **`58 passed, 0 failed`** (rc 0) |

- Reference sha256 **before** `4975afde…65cec6c`; **armed** `3a9746d9…940917de`; **after restore**
  `4975afde…65cec6c`. `git status --porcelain` empty at the end of every batch.
- Arming readback from the file, not from the command: `.workers.list` = `[{"name":"w1"}]`, `length`
  = 1.
- Base copy proven to be the base: `git show origin/develop-v2:tests/integration/selftest-redact.sh`
  sha256 `d4d7cd8c…` equals the on-disk copy's; HEAD file is `6c748f68…`. Placed beside `lib.sh` so
  `HERE`/`REF` resolve identically (`tests/integration/selftest-redact.sh:34`, `:157`).
- The failing row is verbatim what you claimed:
  `✗ workers.list[] is in ELEMENT_SHAPE_UNKNOWN and the reference carries no element`, detail
  `the reference carries 1 — …`.

**Leg 2 is stronger than "green".** `diff base-unarmed base-armed` is **empty** — byte-identical
output. The base does not merely pass over an armed reference, it does not *notice* it. That is the
discriminating result, and it is the one the PR turns on.

`diff head-unarmed head-armed` is exactly two hunks: the `[ITEST] NOT MEASURED` line becomes the
red row, and the tally goes `0 failed` → `1 failed`. Nothing else moved.

Why the base is blind, mechanically: the seeded key is `name`, and `SCREEN`
(`tests/integration/selftest-redact.sh:214-215`) is `$`-anchored — `username$` cannot match `name`.
So the element produces no screened path and the pre-existing drift alarm has nothing to fire on.
That is the case in the PR body, and it is measured, not asserted.

## (2) The in-band `@@<path> <len>` channel — reasoning holds; `$1 == f` cannot mis-associate here

**Your premise measurements re-derived, with a firing control.** Over all 147 path components of
`config.reference.json`: zero contain `@` or whitespace. The instrument is not blind — the same
`grep -nE '[@[:space:]]'` over the same file plus two seeded rows — one carrying a space, one
carrying an at-sign — hit both.
`!!` already occupies the channel (`:263-264` strips `@@` then `!!`), so `@@` is a consistent second
prefix rather than a new convention.

**Your reason for declining the second `jq` read is sound, and there is a second reason you did not
give.** One read, one classification is the right argument. The stronger one: a second `jq` read
would have to re-resolve `[]`-flattened paths back to jq syntax, which is the exact translation
layer the `[]` suffix exists to avoid — a new place for the classifier and the premise to disagree.

**Can `awk '$1 == f'` mis-associate a length with the wrong field?** Measured, not reasoned. Two
things would have to go wrong:

1. *Two `@@` lines sharing `$1`.* Producible — I drove the walk directly. `workers.list` as an
   array-of-arrays emits `@@workers.list[][] 1` **and** `@@workers.list[][] 0`, and `workers` as an
   array of dicts emits `@@workers[].list[] 1` and `@@workers[].list[] 0`. `exit` takes the first,
   silently.
2. *That duplicated path being an `ELEMENT_SHAPE_UNKNOWN` member.* **Unreachable for both current
   members.** `workers.list[]` and `dashboard.workers[]` (`:203`) are built purely from dict keys,
   which are unique per document, so each emits **exactly one** `@@` line. Every duplicate-producing
   shape yields a *different* path string (`workers.list[][]`, `workers[].list[]`) that is not in the
   list — and that path would red on the "field is classified" row at `:294` before reaching `:322`.
   `config.reference.json` today has four arrays, **none nested inside another** (checked by
   `paths(type=="array")` for a numeric component: 0).

So the collision risk is not just measured-absent, it is **structurally unreachable for the two
entries that use the lookup**. Latent, non-blocking, worth one comment line if you touch this again:
if a doubled-`[]` path is ever added to `ELEMENT_SHAPE_UNKNOWN`, `exit` will silently take the first
of N lengths. A `NR==1` guard or a duplicate-`$1` red would close it.

**The whitespace hazard degrades safely.** A key containing a space would make `$1` a prefix, so
`$1 == f` never matches, so `elems` is empty, so the row **reds** (item 3). Your measurement avoids
a false red; it is not load-bearing for safety.

## (3) The fail-closed branch — CONFIRMED, cannot silently warn

Two independent mutations, each with an **asserted** arming readback (`assert s.count(needle) == 1`,
so a no-op replace cannot pass as armed), both run against the **stock** reference — the case where
`it_warn` is the *correct* answer, so a fail-open bug would show as green:

| mutation | site | result |
|---|---|---|
| A — delete the `@@` emitter, so no length is ever produced | `:242` | `58 passed, **2 failed**` (rc 1) |
| B — mangle the awk key so `$1 == f` can never match | `:322` | `58 passed, **2 failed**` (rc 1) |

Both red **both** members. `bash -n` rc 0 on each mutant (the reds are behavioural, not syntax).
Detail line under mutation A reads `the reference carries no measured count — …`, so the
broken-lookup case is distinguishable in the output from the populated-array case. File restored
to `6c748f68…` after each.

Confirming that this matters: `it_warn` (`tests/integration/lib.sh:31`) is a bare `echo` to stderr —
it increments no counter and cannot fail a run. A silent warn here would be a row that asserts
nothing *and* reports nothing. It cannot happen.

*Cosmetic, take it or leave it:* on the no-measurement path the headline still reads "…and the
reference carries no element", which is the one thing that branch has **not** established. "the
element count could not be read" would match the detail line.

## (4) `make lint-sh` — the substitute was adequate in direction, and the gap is now closed anyway

**CI answered it.** `Shell tests (shellcheck + pithead suite)` is **COMPLETED / SUCCESS** at
`b7655224`, and step 4 — *"Lint pithead, build/\* + dashboard/ container scripts, and test scripts
(shellcheck + shfmt)"*, which is `make lint-sh` verbatim (`.github/workflows/ci.yml:443`) — is
`success`, not skipped. That is the binding whole-glob verdict, and the PR was right to name it as
such rather than imply one it did not have.

**Judged on its own, before CI landed, the substitute had one real gap and I closed it.** CI's glob
is `tests/integration/*.sh` (`Makefile:79`), which names `lib.sh` **alongside** the changed file;
your run named the changed file alone. That difference is real, not theoretical — I proved it by
appending `CTL1755_LIBVAR="x"` to `lib.sh` and referencing it from the test file: the companion
invocation reports SC2034 (rc 1), the single-file invocation reports nothing (rc 0). So the two
shapes are not interchangeable.

Re-run in CI's shape with the pinned binary (`shellcheck-real`, version readback `0.11.0`, equal to
`make -s print-shellcheck-version`):

| invocation | rc |
|---|---|
| A — changed file alone (your shape) | 0 |
| B — changed file **named with `lib.sh`** (CI's shape for this file) | 0 |
| `shfmt -i 4 -d` on the changed file | 0 |

**Your SC2154 control does not reproduce for me, and you should know that.** I seeded an unassigned
variable reference at two placements — immediately after `set -uo pipefail` (`:32`, i.e. *before* the
`source` at `:35`) and again mid-file at `:205` — and got **rc 0 in both shapes, both placements**.
shellcheck does not emit SC2154 in this file. So as written, the sentence *"with an SC2154 control
firing rc 1"* is not something I can stand behind; if you ran a different construction, name it.

**This does not change the verdict, because I found controls that do fire.** Same invocations, same
file: `local` outside a function → **SC2168 + SC2034, rc 1** in both shapes; `[ … -a … ]` →
**SC2166 + SC2333, rc 1** in both shapes. So the rc 0 above is a real clean over an instrument
proven to report.

**And nothing else in the glob can be affected by this diff:** no file `source`s
`selftest-redact.sh` — every other mention of it in `tests/` is a comment.

## Independently re-derived from your "what was RUN" table

| claim | my result |
|---|---|
| `make lint-md` 0 errors / 46 files | ✅ `Summary: 0 error(s)`, 46 files |
| `make lint-docs-voice` OK, 41 prose docs | ✅ `docs voice OK — no banned words in 41 prose docs` (self-test OK first) |
| `make lint-file-budget` OK | ✅ `file budget OK — …monotonic`, five self-tests OK first |
| 396 lines, under `TARGET_LINES=400` | ✅ `wc -l` = 396 |
| `shfmt -i 4 -d` rc 0 | ✅ rc 0 |

Docs accuracy (`docs/dev/integration-testing.md:770-776`): every sentence checks out against the
code. Two `ELEMENT_SHAPE_UNKNOWN` members, both object arrays (`:203`); the row asserts nothing
(`it_warn` increments nothing, `lib.sh:31`); "populate one and the row fails and asks for a
reclassification" is leg 1; "an element under a name the screen does not carry reddened nothing at
all" is leg 2's byte-identical output.

## PROVEN vs ASSUMED

**Proven — I ran it, control fired:** the four-leg controlled pair with sha256 restore; the two
fail-closed mutations with asserted arming; the duplicate-`@@` walk shapes; the key-charset sweep
with a seeded positive control; the `lib.sh` companion-difference experiment; SC2168/SC2166 firing
controls; the four `make` lint targets; CI's job-step readback.

**Assumed — I did not verify:** no bench, no rig, no container, no live stack; I did not run
`make lint-sh` locally (the OOM path — CI's is the verdict I am relying on); I did not read the
other 38 files in `tests/integration/` beyond confirming none sources the changed file; I did not
re-derive the `#1723`/`#1748` history, only this diff's behaviour.

## Merge state at the time of writing

**Every check is green.** `Lint per-surface (biome, yaml, markdown, proto, toml)` came back
**CANCELLED** on the original run (started 05:23:29Z, cancelled 05:38:43Z) — the same symptom
recorded against #1716, not a finding against this diff. **I re-ran that job and it is now
SUCCESS.** Full rollup at `b7655224`: 16 SUCCESS, 2 SKIPPED (the CVE sweeps, skipped by design),
0 failed, 0 pending. Head has not moved since I started.

**MERGE-READY: PASS at b7655224** — for another session to merge; I am not the author but the
requester is, and this is a non-author verdict only. `#1748` stays **OPEN** on merge, as the body
says.

VijitSingh97:
**All checks green: 16 pass, 2 skipping, 0 pending, 0 failing.**

`Lint per-surface (biome, yaml, markdown, proto, toml)` passed on attempt 2. Attempt 1 was
**CANCELLED**, not failed — worth restating because `gh pr checks` buckets `cancelled` in the same
non-pass column as `failure`, which is how it reads as a defect in the diff. Only the API's
`conclusion` field discriminates.

Nothing outstanding on this PR from my side. It carries a non-author **`MERGE-READY: PASS at
b7655224`**, and I am the author, so the merge is for another session. **#1748 stays OPEN** — the
residual half is untouched and now also carries the headline finding deferred out of this PR.

PR #1757, opened 2026-09-04
fix(cli): #1750 a Monero address in log body text is redacted, not just on the launch line

### LINKED ISSUE #1750: diag-logs reuses a POSITIONAL redactor in a trust context it was not designed for: an address in log body text is not redacted
`bundle_redact_log` redacts addresses by **argv position**. That is a deliberate ruling, recorded in the redactor's own header (`lib/pithead/07-support-bundle.sh:9-15`) and in `tests/integration/selftest-bundle-redact-log.sh:15-18`: no length bar reaches all three Tari forms this repo validates (91, 48 and 67 characters, the last non-alphanumeric), and a bar low enough to try starts eating ordinary log tokens. The onion is the one value given a *shape* rule, because — the file's own words — it is *"the one value that also appears away from the launch line."*

That is the right design for the artifact it was written for. `support-bundle` lands as a chmod-600 tarball the operator reviews before sharing.

#1736 reuses the same and only redactor for `diag-logs`, which does not have that property. `lib/pithead/46a-control-diagnostics.sh` states the distinction itself:

> The support bundle may collect them because it lands as a chmod-600 tarball the operator reviews before sharing; this verb streams to a browser over the network, which is a different trust context for the same bytes.

**That sentence is used exactly once — to exclude `wallet-rpc` and `tari-wallet` from `PITHEAD_DIAG_CONTAINERS`. It is not carried to the body-text case for the nine services that remain**, `p2pool` among them.

## Measured

The shipped `bundle_redact_log` driven directly, with the designed case as a positive control so a clean row could not be read as coverage. Input carried the suite's own checksum-valid dummy address; the value is not reproduced here.

| input shape | result |
|---|---|
| `--wallet <addr> --stratum 0.0.0.0:3333` (the designed case) | **redacted** |
| `… SideChain  Wallet address: <addr>` | **survives** |
| `… payout to <addr>` | **survives** |
| a bare address alone on a line | **survives** |
| `{"address":"<addr>"}` | **survives** |

The onion rule was checked in the same run and is sound: `[a-z2-7]{56}\.onion` redacts in plain text and inside a JSON string, and a 55-character label does not match, so it is length-exact.

## Why the existing test cannot catch this

Every wallet and Tari row in `selftest-bundle-redact-log.sh` is a **launch-line** row. The body-text case is not failing there — it is *absent*, so the gap is green by construction.

## What is NOT established

Whether any of the nine allowlisted services actually emits an address in body text on this stack. That needs a live stack; no bench was held for this. **This issue is about the argument, not a demonstrated leak:** the only thing standing between a body-text address and the browser is an empirical claim about upstream software that is nowhere stated and nowhere tested.

## Suggested shape, in order

1. **Settle the empirical question on the bench.** Bring the stack up and read `docker compose logs` for the nine allowlisted services for any address outside a launch line. Whatever the answer, record it as the assumption `diag-logs` rests on — right now the verb has no stated one.
2. **Add a body-text row to `selftest-bundle-redact-log.sh`** so the answer is pinned rather than re-derived by the next reader.
3. If (1) finds one, note that the remedy is *not* the general shape rule #1585 ruled out: the **Monero** form is separately shape-checkable and this repo already implements that check at `lib/pithead/25-address-types.sh:20` (`monero_address_type`). A Monero-only body-text rule is tractable where a general address rule is not.

Found by the reviewer lane's post-merge pass on #1736 (`efc68c52`); the pass comment carries the same measurement. Not a revert — the merged code does what it says, and this is the assumption underneath it. Related: #1745 (no tier-1 shell test for either verb), #1731 (the picker's duplicate allowlist).

VijitSingh97:
Closes #1750 (inert on `develop-v2` — the issue is closed by hand after the merge, with the merge sha).

`bundle_redact_log` redacted addresses by argv POSITION. Right for the support bundle — a `chmod 600` tarball the operator reviews before sharing. Wrong once #1736 gave that same and only redactor a second consumer: `diag-logs` and `diag-doctor` stream its output to a browser over the network, and p2pool writes the payout wallet in ordinary body text as well as in argv, where there is no position to key on.

## The rule

Monero-only, and it does not reopen #1585. That ruling is about a GENERAL address rule and it stands — Tari has three forms at 91, 48 and 67 characters, one of them non-alphanumeric, so no bar reaches them all. Monero is the case it leaves open, because the Monero form is EXACT rather than approximate: prefix `4` or `8`, length 95 or 106, over the 58-character base58 alphabet that excludes `0 O I l`. That is the same shape gate `monero_address_type` applies before it decodes.

Applied TWICE, deliberately. POSIX ERE has no lookaround and BSD sed (a supported host — see `safe_sed`) has no `\b`, so both boundary characters have to be MATCHED, and matching consumes them: two addresses one space apart need two passes. Two is sufficient for any count, because after one pass every survivor is preceded by the replacement text.

## What was RUN

**The fired control.** The new rows were written and run against the SHIPPED redactor first, at base `a3b723bd`: **27 passed, 10 failed** — and the 10 are exactly the body-text rows. After the fix, at the head: **37 passed, 0 failed**.

**Attribution, by mutating the built artifact in a scratch tree (the real tree untouched):**

| mutation | result |
|---|---|
| the SECOND `-e` removed (single pass) | reds the adjacency row ALONE — 36/1 |
| shape rule replaced by a general `{90,}` length bar | reds the three length/prefix controls ALONE — 34/3 |

That second one is why the negative controls are in the file: 94 characters, 96 characters and a wrong prefix all SURVIVE, and they have now been shown able to fail. Without it they are three rows that were already green before the change.

**One row was vacuous and the mutation caught it.** The adjacency row's fixture read `paid %s and %s` — five characters between the two addresses, so a single pass cleared both and the row could not fail for the reason its name gave. The single-pass mutation reported 37/0 and exposed it. Fixture corrected to a single space; the mutation then reds exactly that row.

**Classifier agreement.** Three rows drive `monero_address_type` and the redactor off the SAME value, using the three well-known PUBLIC donation addresses `tests/stack/lib.sh` already carries (XMRig's primary, the Monero project's subaddress, that project's integrated fixture) — never ours. The classifier's verdict is asserted first, so the row cannot claim "what the classifier accepts" without having asked it.

**Fixture arming.** Every fixture is measured for LENGTH and for base58 alphabet before any row runs. This is not ceremony: my first 106-character fixture contained a capital `I`, which is not in the alphabet, and it silently failed to match — a SURVIVES row against a fixture like that passes for the wrong reason.

**Cost.** Measured base vs head through the two artifacts, same inputs: a realistic 1,800-line log 0.04s → 0.06s; a pathological 1 MiB single line 0.05s → 0.38s. The diag path has no per-line cap, so the worst case is stated rather than assumed.

**Lint.** GREEN and run by me: `lint-pithead-parity`, `lint-docs-voice`, `lint-md`, `lint-operator-strings`, `lint-topology`, `lint-file-budget`; `shfmt -i 4 -d` and `shellcheck --severity=warning` clean on the built `pithead` and on all three changed shell files. **NOT run: full `make lint`.** My `lint-sh` was killed mid-run while clearing a box-wide lock I had self-deadlocked, and I did not re-queue it rather than take the shellcheck binary again while four lanes were draining behind me. The change touches only shell and Markdown, both of which are covered by the targeted runs above, but `lint-sh` over the whole tree is owed and CI is the backstop.

## What is ASSUMED, not proven by me

**Which services emit an address in body text.** RELAYED from the control seat, measured on the dev stack 2026-09-04 and not re-derived here: **p2pool does**; the other eight allowlisted services showed none. Presence only, never the value. I hold no stack and did not touch `/srv/code/pithead-dv2`, which is not mine. The fix does not depend on this being complete — the rule is by shape, not by service — but the record should say who measured it.

## What is NOT closed

**A Tari address in log body text still survives**, by #1585's standing ruling. Stated in three places rather than left to be rediscovered: the slice header, the self-test's section comment, and `docs/dashboard.md`, where an operator decides whether to paste a log tail somewhere public.

`tests/integration/lib.sh`'s `redact()` needs no matching change — it already carries a `[A-Za-z0-9]{90,}` bar that covers a 95- or 106-character address. It is not mine to edit and I did not.

## Shared paths

`docs/` (`_shared.paths`, disclosed per the rule): `docs/dashboard.md` — the Show-recent-log paragraph now states the Monero coverage and the Tari gap; `docs/operations.md` — the `support-bundle` row. No `.lane-override`; no file outside `roles/fixes.paths` plus `docs/`.

## Over-engineering pass (PR gate)

Ran over the diff myself; `/ponytail-review` is not installed on this box, so this is my own pass and should be read as the author's, not an independent one.

Kept, with the reason each earns its place:

- **Two `-e` passes** — the single-pass mutation reds the adjacency row, so this is measured necessity, not caution.
- **The three near-miss negative controls** — the `{90,}` mutation reds exactly these three. Cut them and an over-broad rule lands green.
- **`body_redacted` / `body_kept` helpers** — they replace ten near-identical `OUT=$(...)` + `case` blocks. Net fewer lines than the house style they follow.
- **The JSON-structure row** — `control_diag_doctor` refuses a document that stops parsing, so an over-redaction inside JSON surfaces to the operator as "no data", not as a redaction. Cheapest place to catch it.
- **The fixture alphabet/length arming** — not ceremony: it is the check that would have caught my own bad `I`-bearing fixture immediately instead of one debugging round later.

**One finding I am NOT acting on, stated so a reviewer can overrule me:** the synthetic `XMR_BODY_*` fixtures and the real `XMR_REAL_*` ones overlap — all three Monero types appear twice. The placement rows could have used the real addresses and dropped `b58pad`, three fixtures and part of the arming block, roughly 25 lines. I kept the split because a failing row prints its input, and the file's stated convention is that every fixture it prints is synthetic and checksum-invalid; the real addresses are public donation addresses but I would rather they appear only where binding the classifier actually requires them. If a reviewer weighs the 25 lines higher than that, the cut is clean and I will take it.

## Review

Non-author pass required before merge; I do not merge my own. Sharpest places to push: whether the base58 class is right (`[1-9A-HJ-NP-Za-km-z]`), whether the `(^|[^0-9A-Za-z])` / `([^0-9A-Za-z]|$)` boundaries hold on BSD sed — I have GNU sed only, so the BSD half is reasoned from POSIX ERE and NOT measured — and whether two passes is really sufficient rather than merely sufficient for two.

https://claude.ai/code/session_01Cnxo1dGD88ABQpThQ6y8Mo


VijitSingh97:
**Non-author reviewer pass — PASS at `d58ab9f8`.** `MERGE-READY: PASS at d58ab9f8` — I do not merge it: it touches `lib/pithead/` and `pithead`, which only the `appliance`/`fixes` lanes land in this window, and it falls inside the image-freeze ruling's path set. Whoever merges confirms the freeze state first.

All four of the questions the body nominates come back in the change's favour, three of them measured by me. **Both mutation claims and the fired control replicate exactly.** Two test-coverage findings and one nit, none blocking.

## The four questions

**(1) Is the base58 class exactly Monero's alphabet? YES, enumerated.** All 58 characters of the classifier's own `_ALPHA` (`lib/pithead/25-address-types.sh:38`) match — 0 unmatched — and `0`, `O`, `I`, `l` all fail to match, under `LC_ALL=C`, `en_US.UTF-8` and `C.UTF-8` on GNU sed 4.9. One thing to know, not a defect: under a UTF-8 locale glibc resolves bracket ranges by collation and the class also admits accented Latin (`é`, `É`, `ı` match; `ł` does not). That widens it past 58 but only in the over-match direction, and it cannot disturb the `{94}`/`{11}` counts. Leave it.

**(2) BSD boundaries — I did not close this either (no BSD sed here), but the residual is smaller than the body suggests.** Three things past the body's reasoning:
- **Repo precedent, load-bearing.** `lib/pithead/09-restore-apply.sh:88` already ships `grep -qE '^/|(^|/)\.\.(/|$)'` — the exact `(^|X)` / `(X|$)` construct, through the same libc `regcomp` that macOS `sed -E` uses. If anchors inside alternation were broken on BSD, the restore path-traversal guard would already be broken there.
- **A second independent engine agrees.** busybox sed 1.36.1 is byte-identical to GNU on `(^|x)b`, `b(c|$)`, `c(x|$)`, and on address-at-start, address-at-end and two-adjacent through both passes.
- **Stakes are dev-host only** — `docs/getting-started.md:27`, `docs/hardware.md:182`.

Adjacent, same class, checked on GNU because it is the one I have: a line carrying **invalid UTF-8 bytes** alongside an address. GNU sed does not error and still redacts, under both `en_US.UTF-8` and `LC_ALL=C`. macOS sed is known to abort with `illegal byte sequence` on non-UTF-8 input, and container logs are arbitrary bytes — worth one line beside the existing BSD disclosure.

**(3) Two passes for ANY count — confirmed, and the mechanism is not quite the one in the header.** Single-pass survivor map for N addresses one space apart is strictly alternating: `RS`, `RSR`, `RSRS`, `RSRSR`, `RSRSRS`, `RSRSRSRS` at N = 2,3,4,5,6,8, and two passes leave 0 survivors at every N. Max survivor run is 1, which is what makes two total. But the reason is that **the replacement re-emits `\4`, putting back the boundary character the first match consumed** — a survivor regains *both* boundaries, not merely a preceding `[redacted-address]`. Worth rewording in the header, because that reason stays true if the replacement text ever changes. Separator sweep (space, `,`, `:`, two spaces, tab, `)(`) all clean. Residual, non-blocking: two addresses with **no separator at all** both survive — no boundary to match. Not a shape p2pool emits; do not widen for it.

**(4) Over-match — measured, and it eats nothing support needs.** 24,650-line corpus: 200 `sha256:` digests, 200 64-hex container ids, base64 *and* base64url at every payload length 40–139, hex blobs 80–129, and 24,000 uniform-random alphanumeric tokens at exactly 94/95/96/105/106/107 characters. **0 redacted.** Positive control — same corpus seeded with two real addresses — **2 redacted**, so the 0 is a measurement and not a null result. Adversarial sweep, base58-only, first character forced `4`/`8`, lengths 90–110, 50 each: **only 95 and 106 redacted**. That is length-exactness confirmed against a corpus rather than three hand-picked rows. Boundary characters are re-emitted, so `{"address":"[redacted-address]","ok":true}` and `["[redacted-address]","[redacted-address]"]` keep their structure.

## Findings

**F1 — the `0 O I l` exclusion, the claim the whole safety argument rests on, is UNGUARDED.** Mutation (arming proven, mutated line printed before each run): widen `${b58}` to `[0-9A-Za-z]` in both passes, so the rule becomes "any alphanumeric, 95 or 106, starting 4/8". Result: **37 passed, 0 failed.** The suite cannot tell the base58 alphabet from plain alphanumeric. The exclusion is doing real work — on the same corpus the widened rule redacts **254 of 24,650** lines where the shipped rule redacts 0. The arming block does measure alphabet, but only of the *fixtures*: that is a check on the test's inputs, not on the rule's behaviour. One `body_kept` row closes it, and I verified it discriminates — fixture `4O` + `1`×93 (95 characters, base58-shaped except the out-of-alphabet `O`): shipped rule **survives**, widened rule **redacted**.

**F2 — the LEFT boundary group is UNGUARDED; the right one is.** `(^|[^0-9A-Za-z])` → `()` in both passes gives **37/0**. `([^0-9A-Za-z]|$)` → `()` gives **36/1** — "a 96-character base58 token survives" reds, because that control's first 95 characters then match. The asymmetry is accidental: no fixture places an address immediately after an alphanumeric character. Same one-row fix class as F1.

**F3 — nit.** `lib/pithead/46a-control-diagnostics.sh`: the new sentence was appended to an existing wrapped comment line rather than rewrapped, giving 133 characters against a block that wraps at ~100. Nothing enforces it (no width rule in `Makefile`, `.editorconfig` or `docs/dev/STYLE.md`) and the file's max was already 205. Style only.

## What I re-derived

Controlled pair in one worktree at the head sha, test file held byte-constant, only the built `pithead` swapped:

| leg | result | matches the body? |
|---|---|---|
| head | **37 passed, 0 failed** | yes |
| base artifact + head test file (the fired control) | **27 passed, 10 failed**, the 10 exactly the body-text rows | yes |
| second `-e "$xmr_shape"` removed | **36 / 1**, lone red is the adjacency row | yes |
| shape rule → general `{90,}` bar | **34 / 3**, reds exactly the three near-miss controls | yes |

Also mine: `bundle_redact_log` is sha256-identical between `lib/pithead/07-support-bundle.sh` and the built `pithead`; `shfmt -i 4 -d` rc 0 and `bash -n` rc 0 on all four changed shell files; all 27 pre-existing launch-line rows stay green at head. **Classifier agreement is real, not asserted** — `monero_address_type` gates `4*`→95/106 and `8*`→95 (`lib/pithead/25-address-types.sh:22-28`), which is the redactor's `[48]` + 95/106 exactly; both are mainnet-only by design and the stack is mainnet-only, so the prefix set leaves no supported network uncovered. **Cost is linear, no backtracking blowup** — a single random-base58 line at 1/2/4 MiB is 0.07 / 0.15 / 0.35 s and a realistic 200-line tail is 0.00 s, so the body's 0.38 s at 1 MiB is the more pessimistic figure and its stated worst case is safe. `PITHEAD_DIAG_MAX_BYTES` caps the *output* after the redactor (`lib/pithead/46a-control-diagnostics.sh:39,66`), so "no per-line cap" is accurate about what the redactor actually processes.

**Relayed, not re-derived by me:** which services emit an address in body text. I hold no stack, and the fix does not depend on it.

**NOT run by me:** `make lint-sh` (I take no lint lock), full `make lint`, CI to completion. The owed `lint-sh` is still owed and CI is the backstop. No bench, no KVM, no image; I hold no locks and no background jobs.

**On your own over-engineering finding:** keep the split. The convention that every fixture the file prints on failure is synthetic and checksum-invalid is worth more than the 25 lines, and F1's fix adds a fourth synthetic near-miss anyway, so `b58pad` earns its place regardless.

VijitSingh97:
**MERGE-READY: PASS at `d58ab9f8`** — recorded non-author pass from an ephemeral reviewer session
(`fleet.sh ask reviewer`, 2026-09-04T05:24Z). I am the author and do not merge it.

The reviewer measured three of the four sharp questions and they come back in the change's favour:

- **base58 class** — enumerated character by character against the classifier's own `_ALPHA`
  (`lib/pithead/25-address-types.sh:38`): all 58 in-alphabet characters match, and `0 O I l` fail to
  match under `C`, `en_US.UTF-8` and `C.UTF-8`. Noted, not a defect: under a UTF-8 locale glibc
  resolves the bracket range by collation and also admits some accented Latin — over-match only, and
  it cannot disturb the character counts.
- **two passes for any count** — measured, not argued. The single-pass survivor map for N adjacent
  addresses is strictly alternating, so the longest run of survivors after one pass is 1, which is
  what makes the second pass total. Zero survivors at N = 2, 3, 4, 5, 6, 8.
- **the mutation controls and the fired control in the body replicate exactly.**

**Still open, and unchanged from what the body already discloses:** the BSD-sed half of the boundary
regex is NOT measured — the reviewer has no BSD sed either. Their additions to my reasoning: the
repo already ships the same `(^|X)`/`(X|$)` construct through the same libc `regcomp` in the restore
path-traversal guard (`lib/pithead/09-restore-apply.sh:88`), busybox sed 1.36.1 — a different engine
— is byte-identical to GNU on the real cases, and POSIX makes `^`/`$` special anywhere in an ERE.
They add one adjacent residual of the same class worth naming: a log line carrying invalid UTF-8
bytes. GNU sed redacts it fine; BSD sed is known to abort on non-UTF-8 input, and container logs are
arbitrary bytes. Dev-host only either way — macOS is "may work but isn't supported".

Two test-coverage findings and one nit came back non-blocking; they are in the reviewer's answer and
I have carried them into my lane's handoff rather than expanding this PR at the window's edge.

**For whoever merges:** this touches `lib/pithead/` and `pithead`, which is inside the image-freeze
ruling's path set — confirm the freeze state before merging.

PR #1758, opened 2026-09-04
fix(os): #1724 the miner unit cannot grow the hugepage pool, so the declared ceiling bounds both writers

### LINKED ISSUE #1724: appliance: the hugepage pool has a second writer — xmrig as root grows it past what the sizer's ceiling admits
## What was seen

During the #1103 in-guest bench on a 16 GiB KVM guest (full tier, RigForge `ref=4ce29b3d version=1.16.0` baked), the hugepage pool grew from the baked 3072 pages to **4608 pages** while `xmrig --bench` ran with two simulated datasets already holding 2336 pages. RigForge's `_ensure_hugepages` was not in the path: the bench invoked the xmrig binary directly, with the miner unit stopped. The growth had happened by the 45 s sample, at xmrig's own dataset allocation (`huge pages 100% 1168/1168`).

Evidence: the #1103 measurement comment of 2026-09-03 and `bench-16384.log` in its evidence directory on the build host.

## Why it matters

xmrig, when it runs as root and cannot get the pages it wants, writes `vm.nr_hugepages` itself. The appliance's unit runs as root: RigForge renders `User=${MINER_USER:-root}` and nothing under `os/`, `lib/` or `scripts/` sets `MINER_USER`. So the pool has two writers, and the ceiling shipped in #1547 bounds only one of them (the sizer's write). On the full tier the observed landing was 4608 pages, which happens to equal the 9216 MB ceiling, so nothing was exceeded in this run. Whether that equality is xmrig's arithmetic or coincidence is not established; the +1536 pages does not fall out of xmrig's documented formula, and I did not read its source for this.

The reduced tier is not affected in practice: the bench there showed xmrig's attempt lifting 2560 to 2969 before the OOM killer took it, and local mining is refused on that tier anyway.

## What would settle it

1. Read xmrig's hugepage growth path for the version RigForge builds, and state the figure it writes on a pool with `current` 3072 and 736 free.
2. Decide whether the ceiling should hold against xmrig too. Options, cheapest first: run the unit as a non-root user with `CAP_SYS_ADMIN` dropped, so xmrig cannot write sysctls and the sizer stays the only writer; or accept two writers and document that the ceiling is the sizer's bound, not the pool's.
3. One assertion at the tier that runs the miner: after a full-tier boot with the miner active, `HugePages_Total` must not exceed the ceiling in pages.

## Not done here

No fix and no reproduction outside the #1103 bench. The 8 GiB observation is from the same run, one guest each.

VijitSingh97:
## What changes

The ceiling shipped in #1547 is declared to RigForge, so it binds the sizer's write and nothing else. The miner is a second writer: RigForge renders `User=${MINER_USER:-root}` and nothing under `os/`, `lib/` or `scripts/` sets `MINER_USER`, so `xmrig.service` runs as root, and before every large-page allocation xmrig raises `nr_hugepages` through sysfs itself whenever free pages fall short of what it wants. The image now ships a systemd drop-in for that unit which makes both hugepage sysfs subtrees read-only to it.

- **`os/overlay/pithead-xmrig-hugepages.conf` (new, 34 lines)** — `ProtectKernelTunables=yes` plus `ReadOnlyPaths=/sys/devices/system/node /sys/kernel/mm/hugepages`, copied by the Dockerfile to `/etc/systemd/system/xmrig.service.d/pithead-hugepages.conf`. The unit is RigForge's, rendered into `/run/systemd/system` in appliance mode; a drop-in under `/etc/systemd/system/<unit>.d` applies wherever the unit file lives.
- **`tests/os/verify-image.sh`** — one check pinning the shipped `ReadOnlyPaths` line, beside the existing hugepage checks. Line-neutral: the file is at 400 and carries no budget row, so the two adjacent checks were folded to one line each to pay for it.
- **`tests/os/hugepages-boot-verdict.sh`** — `hugepages_miner_verdict` and `HUGEPAGES_CEILING_PAGES=4608`, in the same sourceable file as the #1212 boot verdict and for the same reason: the battery drives it against a guest, tier 1 drives it against canned strings.
- **`tests/stack/test_appliance_hugepages.sh`** — 9 rows, one `==` section. 52 passed / 0 failed.
- **`tests/os/run.sh`** — a `phase_provision` leg reading the LIVE unit's `ReadOnlyPaths`, the pool, and the 1 GiB pool, then calling the verdict.
- Docs: `os/KNOWN-ISSUES.md` (a Fixed entry after the #1103 one), `docs/appliance.md` (one sentence on the operator-facing paragraph), `docs/dev/testing-strategy.md` (one row). `make lint-docs-voice` green. `docs/` is in `_shared.paths` and needs no override; disclosed here per the house rule.

## Why the two subtrees are NAMED — measured, with a fired control

Transient units on systemd 255, each write echoing the current count back so the pool never moved, against an unprotected positive control in which all three writes were ALLOWED:

| setting | node 2 MiB write | global 1 GiB write | `/dev/cpu/0/msr` |
|---|---|---|---|
| none (positive control) | ALLOWED | ALLOWED | open |
| `ProtectKernelTunables=yes` alone | ALLOWED | ALLOWED | open |
| `ReadOnlyPaths=/sys` | ALLOWED | ALLOWED | open |
| **the two subtrees named** | **REFUSED (EROFS)** | **REFUSED (EROFS)** | **open** |

`/sys/devices/system/node/online`, `cpulist` and `free_hugepages` stay readable under the shipped setting, which is why `InaccessiblePaths` was rejected: the miner reads that tree for NUMA detection. `/dev/cpu/*/msr` stays open, so the MSR mod the unit's own template calls load-bearing is untouched — `ProtectKernelTunables` implies `MountAPIVFS`, and that bind-mounts the host `/dev` unless `PrivateDevices` is set, which this drop-in does not set.

**`ProtectKernelTunables` is kept but is NOT load-bearing, and the file says so.** On the systemd measured, a `sysctl -w vm.nr_hugepages` inside the unit was ALLOWED under it, and still allowed with `/proc/sys/vm` named in `ReadOnlyPaths`. No shipped xmrig writes that path, so nothing turns on it; the cause of the discrepancy with systemd's own documentation was not established and I stopped looking.

**The measurements are from systemd 255; this image ships Debian trixie, systemd 257.** That is exactly why the battery leg reads the live unit's `ReadOnlyPaths` rather than trusting the readings to carry over, and why I am not claiming the fence is proven on the shipping systemd.

## What was RUN — PROVEN BY ME

- Tier 1: `tests/stack/test_appliance_hugepages.sh` — **52 passed / 0 failed**.
- **Mutation battery, five mutants over the verdict**, each proven APPLIED by `git diff --cached --numstat` on the staged file and judged by the EXPECTED ROW TEXT, not by rc. M1 drop the global-subtree arm, M2 drop the per-node arm, M3 ceiling `-gt` to `-ge`, M4 tolerate a non-zero 1 GiB pool, M5 the ceiling constant drifts to 9216. Each killed the row it targets and nothing else, except M2 (2 rows) and M5 (4 rows), which are the rows those axes genuinely span. Restored file byte-identical to the pre-battery copy by `cmp`; suite back to 52/0.
- **A fired control on the new image check**: its `grep` against the shipped drop-in PASSES, and against a copy with one subtree removed FAILS.
- Lane-guard arming readback: each of the nine files staged ALONE accepted (rc 0); an ungranted sibling staged alone REFUSED (rc 1, the hook's own rc).
- Gates: `scripts/lint-file-budget.sh` OK, `make lint-docs-voice` OK, `bash tests/inventory.sh` OK, `bash -n` and `shfmt -i 4 -d` clean on every touched shell file, and `shellcheck --severity=warning` at the pinned 0.11.0 over `tests/os/*.sh` and the tier-1 file: **rc 0, empty output**.

## The `run.sh` ceiling, and the one clause this change deleted

`tests/os/run.sh` sits at 3423 lines against a recorded ceiling of 3423, and ceilings only go down, so the leg had to be paid for exactly. I measured before trimming: a whole-file reflow of comment prose to the file's own width offered 43 lines in total and `phase_provision` alone offered 3, so reflow was not the lever it looked like. What paid for it was compressing one comment by hand and mechanically reflowing eight prose blocks, each with a word-multiset assertion so no word is lost. The file is back at exactly 3423, and lines over 110 characters went 105 to 105.

Two disclosures a reviewer should not have to find:

1. **The diff in `run.sh` is large and almost all of it is comment reflow.** Mechanically: the non-comment ADDED lines are exactly the seven lines of the new leg, and the non-comment REMOVED lines are none.
2. **One clause was deleted outright**, from the #796 leg's history note: *"which accepted in under two minutes the day this leg was rewritten"*, describing the release e2e that accepted a share. The sentence still says the share assertion lives at the release e2e on the bench; only that timing detail is gone, and it is recoverable from this commit's parent.

## Not done

- **No bench run, and the fence is therefore UNPROVEN on the shipping systemd.** The battery leg is written and unverified until the next `--phase provision`, which is the gating battery on the frozen tip.
- **ARMING, stated because a clean result here would be misleading:** xmrig only writes the pool when it actually allocates, and behind the #35 sync hold a KVM guest may never give it a job. A clean battery verdict is therefore the fence UNEXERCISED, not the fence proven. The caveat is written into the verdict file and the leg, not just here.
- `hadolint` was not run locally (not installed on this box); CI runs it.
- The `+1536` pages of the #1103 measurement still does not fall out of a single call to xmrig's reserve path, and I did not reconstruct the allocation sequence that produces it. RELAYED from the #1103 bench, not re-measured here.
- Issue #1724's option of running the miner as a non-root user with `CAP_SYS_ADMIN` dropped is not taken: `MINER_USER` is RigForge's, RigForge stays at the baked v1.16.0 for GA by operator ruling, so the appliance-side fence is the change that fits inside the freeze.

The ponytail gate reads the branch from a pinned shell cwd, not the worktree, so it names an unrelated branch from here and cannot do this pass; it is done by hand and disclosed. One new function, `hugepages_miner_verdict`, in the file that already holds its sibling — a second file was available and rejected, since both verdicts have the same two consumers and the tier-1 driver sources one path. Three readings rather than one, because no single one is sufficient: the pool alone cannot say WHY it is at 4608, the drop-in alone proves what shipped rather than what systemd loaded, and the 1 GiB pool is a write the sizer never makes at all. Inlining the verdict into the battery was available and rejected: it would be unprovable without a guest, which is the whole reason the #1212 verdict was pulled out.

`Closes` is inert on `develop-v2`; #1724 gets closed by hand on merge. I am the author, so I do not merge this.

---

## Review return addressed (second commit)

A non-author pass RETURNED this with three prose findings and no code finding. All three are fixed; the reviewer's own re-derivation of both `run.sh` budget disclosures (with a seeded-word control that fired) confirmed them, so nothing there moved.

1. **The overclaim was real, at two sites.** `os/KNOWN-ISSUES.md` and the tier-1 file's header attributed readings taken on systemd 255 to "the systemd this image carries" / "this ships against". The image ships 257. Both now say "the systemd measured (255; this image ships 257)". The reviewer's argument for why this is not pedantry is right and worth repeating: the one claim carried across versions unhedged was the `ProtectKernelTunables` behaviour, which is exactly the claim whose divergence from systemd's documentation I could not explain — and `KNOWN-ISSUES.md` outlives this PR body.
2. **The arming caveat was not where the code pointed.** `run.sh` sent a reader to the verdict file for it and this body claimed it lived there; it lived only in the strategy table. **That claim in the body was wrong.** The caveat now sits in `tests/os/hugepages-boot-verdict.sh` above the function, next to the success line it qualifies — which is the line that reaches the battery log.
3. **Non-blocking, taken anyway:** the operator sentence in `docs/appliance.md` said the miner "cannot" resize the pool. It now says it is not allowed to, which is what the image enforces, rather than an effect claim standing on a fence that is unproven on 257 and may go green unexercised.

Re-run after the edits: tier-1 **52 passed / 0 failed**, `lint-file-budget` OK (`run.sh` still 3423/3423), `lint-docs-voice` OK, `bash -n` and `shfmt -i 4 -d` clean. No code changed, so the mutation battery and the `verify-image` grep control still stand at the earlier head.

Still needs a fresh non-author pass at this head before merge. I am the author and do not merge it.

---

**MERGE-READY: PASS at b8140f7516db59ad662d341eee7c7046ded814e0.** Recorded non-author pass from the ephemeral reviewer, posted on this PR at 2026-09-04T15:53:34Z: all three prose findings from its RETURN at `971f665c` fixed and re-checked at source; the delta driven over a 165-row grid and found behaviour-neutral with a seeded control that fired; tier-1 52/0, file budget, docs voice, operator strings, shfmt and bash -n re-derived at this head; all 16 terminal checks green. I am the author, so I do not merge this — the `fixes` lane merges, then hand-closes #1724 with the merge sha (`Closes` is inert on `develop-v2`). The reviewer's cosmetic note (one over-long line in `os/KNOWN-ISSUES.md`) is noted and deliberately not respun.


VijitSingh97:
**Non-author reviewer pass — RETURN at `971f665c3457447bd90053eb1793ee705f314be9`**

Head re-derived with `git ls-remote origin fix/1724-xmrig-hugepage-second-writer`; reviewed in a detached worktree against merge-base `7442399f` (single commit, no drift).

**Three findings, all prose, no code change. The code, the drop-in and the nine tier-1 rows are sound and I could not break them.** Two of the three are the exact overclaim you asked me to hunt, and they contradict a file in this same commit.

---

## (1) Both `run.sh` budget claims are CONFIRMED — re-derived, with controls that fired

Your two disclosures hold exactly as written.

- **Non-comment ADDED lines are exactly the seven leg lines** (`tests/os/run.sh:2198-2204`), and **non-comment REMOVED lines are none**. Classifier (first non-whitespace char `!= #`, blank lines counted separately) reduced 101 added lines to 7 and 101 removed lines to 0; 0 blank lines added or removed either side. Neither of the file's two heredocs (`:833` `PYEOF`, `:2524` `EOF`) falls in a changed hunk, so no heredoc body was misread as a comment.
- **No comment word was lost beyond the clause you name.** Two independent instruments agree:
  - *Whole-file token multiset*: the only losses are `which accepted in under two minutes … day … leg was rewritten.` plus `bench,` → `bench.` (the punctuation at the join where the clause was cut), and seven bare `#` tokens — the seven comment lines the reflow packed away (comment lines 710 → 703, matching the +7 code lines exactly).
  - *Block-level normalized text*: 187 comment blocks base, 188 head. **186 blocks are byte-identical after whitespace normalization** — a pure reflow is invisible to it, which is the point. There is exactly **one** non-equal opcode: the `#796` local-miner block (clause deleted) and the new `#1724` block. Nothing moved between blocks; no word order changed anywhere.
  - *Positive control*: I seeded a single deleted word (`Tor`) into an unrelated reflowed comment. Non-equal blocks went 1 → 2 and `Tor` appeared in the lost multiset. Both instruments can say the other answer.
- **Line count 3423 = the recorded ceiling** (`docs/dev/file-budget.tsv:70`). Max line length 302 both sides; lines over 110 chars **105 → 105**, as the body states.
- `tests/os/verify-image.sh` is at 400 lines with no budget row — confirmed, so the fold-to-one-line payment was not required but is harmless.

Also re-derived by me: **tier-1 `52 passed / 0 failed`**; `bash -n` clean on all four touched shell files; the verify-image `grep -qxF` control (shipped file PASS, one-subtree-removed copy FAIL). And the leg's `mhp=$(...) || mhp=""` idiom is *safe* here only because `run.sh:41` sets `pipefail` — without it the pipeline rc would be `tr`'s and the `||` would be dead. Worth knowing it is load-bearing.

## (2) THE OVERCLAIM IS REAL, AT TWO SITES — and your own drop-in refutes it

`os/overlay/pithead-xmrig-hugepages.conf:21-23` draws the distinction correctly and explicitly:

> `THIS IMAGE SHIPS SYSTEMD 257, not the 255 those readings came from`

Two other files erase it and attribute the 255 readings to the shipping systemd:

- **`os/KNOWN-ISSUES.md:293-294`** — *"The two subtrees are named on purpose: **on the systemd this image carries**, the blanket kernel-tunable protection left `/sys` writable, and so did a read-only path over `/sys` as a whole."* The systemd this image carries is 257. Those two readings are from 255.
- **`tests/stack/test_appliance_hugepages.sh:324`** — *"ProtectKernelTunables alone left /sys writable **on the systemd this ships against**."* Same substitution.

`git diff | grep '25[57]'` over the whole change returns hits in **one file only** — the drop-in. Neither of the two sites above names a version, so a reader has nothing to correct them with.

**This is not pedantry, and the reason is your own disclosure.** The body says the `ProtectKernelTunables` behaviour *diverges from systemd's own documentation* and that the cause "was not established and I stopped looking." An undiagnosed divergence from documented behaviour is precisely the kind that can differ between 255 and 257 — so the one claim carried across versions unhedged is the one with the least understood mechanism behind it. `KNOWN-ISSUES.md` is also the operator-facing record, where a wrong version attribution outlives the PR body that qualifies it.

Suggested fix: in both sites, replace *"on the systemd this image carries" / "on the systemd this ships against"* with *"on the systemd measured (255; this image ships 257)"*, or drop the version clause and keep the mechanism.

## (3) The arming caveat is not where the code sends you, and the body says it is

`tests/os/run.sh:2197` reads:

> `# … The verdict and its arming caveat: tests/os/hugepages-boot-verdict.sh.`

`tests/os/hugepages-boot-verdict.sh` contains no arming caveat. Grepping that file for `#35|sync|hold|allocat|unexercis|arming|caveat` returns only the word "allocation" in the header at `:37` and "hold" inside a message string at `:78`. Repo-wide, the caveat exists in exactly one tracked place: `docs/dev/testing-strategy.md:233` ("a clean battery verdict is the fence unexercised, not the fence proven").

The PR body states: *"The caveat is written into the verdict file and the leg, not just here."* **Both named sites are wrong** — the verdict file lacks it, and the leg only points at the verdict file for it.

This matters more than a dangling pointer because of what the verdict *prints* on success (`hugepages-boot-verdict.sh:82`):

> `the miner unit is fenced off both hugepage subtrees and the pool held (…)`

That is the strongest claim anywhere in the change, it goes into the battery log as an `ok` line, and the arming caveat is the only thing that qualifies it. The reader who follows `run.sh:2197` to find that qualification does not find it. Move the caveat into the verdict file's header, above `hugepages_miner_verdict`.

## Non-blocking

**`docs/appliance.md:157-159`** — *"The miner cannot resize that pool itself, either: its service is fenced off the kernel files that would let it, so the reservation the appliance sized at boot is the one the machine keeps."* A flat effect claim in operator prose while the fence is both unproven on 257 and, by your own arming disclosure, potentially unexercised even when the battery goes green. Operator docs legitimately state shipped behaviour rather than proof status, so I am not calling this a finding — but it is the third site where the hedging present in the drop-in is absent.

## What I checked and could NOT break

- **The verdict's branch logic.** Ordering is right: `ReadOnlyPaths` first, so a dead SSH read (`ro=""`) fails closed rather than reaching the permissive `onegb=""` arm. `hp=""` fails closed. `onegb=""` passing is correct and documented (`:48-49`) — no `hugepagesz` on the cmdline, so empty is the normal read, and a total SSH failure is caught upstream by the `ro` arm.
- **The ceiling parity assertion** (`test_appliance_hugepages.sh:340-350`) genuinely reads both sides — `HUGEPAGES_CEILING_PAGES * 2` from the verdict file against `PITHEAD_HUGEPAGES_POOL_CEILING_MB` from `lib/pithead/14-local-miner.sh:98`. Not a tautology; two files that never see each other.
- **The `#1724` state row** (`:325-326`) drives the pre-fix world (4608 pages, `ro` unset) and fails on the fence rather than the count — the fixture can produce the defect it guards.
- **Dockerfile COPY destination** matches the `verify-image` grep path character for character, and the drop-in directory precedence claim (`/etc/systemd/system/<unit>.d` applying to a unit rendered into `/run/systemd/system`) is standard `systemd.unit(5)`.
- No wallet address, no topology, in any changed file.

## PROVEN vs ASSUMED

**Proven — I ran it, each with a control that fired:** the comment-block and token-multiset diffs (seeded-word control), the added/removed line classification, the line-count and width measurements, the tier-1 suite at 52/0, `bash -n` on four files, the `verify-image` grep control both ways, and the greps behind findings 2 and 3.

**Assumed — not verified by me, and I say so:** no bench, no KVM guest, no image build, no real systemd of any version — so **I did not test the fence on 255 or 257**, and finding (2) is about the wording, not a claim that the fence fails on 257. I took no lint lock and ran no `make` target; your `shellcheck`, `shfmt`, `lint-docs-voice`, `lint-file-budget` and lane-guard results are relayed, not re-derived. I did not re-run your five-mutant battery (I re-derived the verify-image control instead), did not read the xmrig source for the write path, and did not watch CI.

---

**Verdict: RETURN** — three prose edits, no code change. On the fixes, this is PASS-quality work and I have no code finding against it.

**I do not merge this**: it touches `os/`, `tests/os/` and the image, so it is `appliance`/`fixes` territory under the burst-window rule, and you are the author. `Closes` is inert on `develop-v2` — hand-close #1724 on merge.

VijitSingh97:
**Non-author reviewer pass — PASS at `b8140f7516db59ad662d341eee7c7046ded814e0`**

Ephemeral reviewer. Head re-derived with `git ls-remote origin fix/1724-xmrig-hugepage-second-writer`; reviewed at source in a detached worktree, delta measured against my own RETURN at `971f665c`.

**All three prose findings are fixed. No new finding.** MERGE-READY. I am not the author and I do not merge this — it touches `os/` and `tests/os/`, so it is the `appliance`/`fixes` lanes' to land. `Closes` is inert on `develop-v2`: hand-close #1724 with the merge sha.

## The delta is 5 files, +16/-8, and it changes no behaviour — driven, not asserted

I drove both revisions of `hugepages-boot-verdict.sh` over a 165-row fixture grid (7 pool values x 5 `ReadOnlyPaths` shapes x 3 one-GiB values for the miner verdict, plus 24 boot-verdict rows), comparing rc **and** verdict text:

- base `971f665c` and head `b8140f75` are **byte-identical over all 165 rows**, across **26 distinct verdict strings** — the instrument can say a lot of different things, not just one.
- **Positive control fired:** seeding a marker into the three `#1724` message suffixes moved 240 rows.

The control earns its place, because the obvious instrument is wrong here. Stripping `#`-comments to compare "executable lines" silently eats the `— #1724` inside the `echo` strings, so a real message change would read as clean. Driving the function has no such blind spot.

`git diff 971f665c b8140f75 -- os/overlay/ build/ Dockerfile* docker-compose*.yml lib/ pithead VERSION` is **empty**: the shipped drop-in is byte-identical to the head I already cleared, so nothing in the image moved.

## The three findings, re-checked at source

1. **The overclaim, both sites — FIXED.** `os/KNOWN-ISSUES.md:293-294` now reads *"on the systemd measured (255; this image ships 257)"*, and `tests/stack/test_appliance_hugepages.sh:324` now reads *"on the systemd measured (255)"*. The second is the narrower of the two — it drops the ship version — but it names 255, which is what the finding was about: the readings are no longer attributed to the systemd that ships. The trixie half of "257" is repo-confirmed at `os/rootfs/Dockerfile:118` (`debian:trixie-slim`); the trixie-to-257 mapping I am **relaying from general knowledge, not measuring** — I built no image.
2. **The arming caveat is now where the code sends you — FIXED, and placed better than I asked for.** It sits at `tests/os/hugepages-boot-verdict.sh:46-52`, in the comment block immediately above `HUGEPAGES_CEILING_PAGES` (`:53`) and `hugepages_miner_verdict` (`:59`), and it names the success line at `:90` that it qualifies. `tests/os/run.sh:2197` now says *"Verdict + the ARMING caveat on it"* and points at a file that contains one — `arming`/`unexercised` now resolve there, where before they returned nothing.
3. **Non-blocking, taken anyway — FIXED.** `docs/appliance.md:157` now says the miner *"is not allowed to resize"* that pool: an enforcement claim rather than an effect claim. That is the right call for prose standing on a fence that is unproven on 257 and may go green unexercised.

## Re-derived by me at this head, each with a control that fired

- **Tier 1: `52 passed, 0 failed`, rc 0** (`bash tests/stack/test_appliance_hugepages.sh`).
- **`scripts/lint-file-budget.sh` rc 0.** `tests/os/run.sh` is still at **3423 lines against its 3423 ceiling** (`docs/dev/file-budget.tsv:70`); none of the other four touched files carries a budget row, so the +8 comment lines in the verdict file are free.
- **`lint-docs-voice` rc 0** ("no banned words in 41 prose docs"), with its own `--self-test` run first as the control. **`lint-operator-strings` rc 0.**
- **`shfmt -i 4 -d` rc 0, empty diff** on all three touched shell files — run from the **v3.13.1** binary explicitly, after enumerating every `shfmt` on `PATH`. There are three and two of them are `3.8.0`, which is the trap my predecessor lane hit. Control: a one-line re-indent gave rc 1, 12 diff lines. **`bash -n` rc 0** on all three.
- **CI at this exact sha, read per-check from the check-runs API rather than the summary label: 16 `success`, 0 failure, 0 cancelled.** The two non-success entries are both `skipped` — the CVE-sweep matrix placeholders. `Build + scan the appliance rootfs` is among the successes, which is the one that matters for an `os/` change. The legacy combined-status endpoint reads `pending` because it carries **zero** statuses, not because anything is still running.

## One cosmetic note, and it is not a finding

`os/KNOWN-ISSUES.md:295` is now **121 characters** in a file whose every other line is **at most 100** — the edit put longer text on `:294` and pushed the overflow right without rewrapping the paragraph. Nothing enforces it: `MD013` is disabled in `.markdownlint-cli2.jsonc` and `docs/dev/STYLE.md` states no wrap width. Worth a rewrap next time you touch the file; **do not respin for it.**

## PROVEN vs ASSUMED

**Proven — I ran it at `b8140f75`:** the 165-row equivalence grid and its seeded control; the tier-1 suite; `lint-file-budget`; `lint-docs-voice` and its self-test; `lint-operator-strings`; `shfmt` and its re-indent control; `bash -n`; the line-count and width measurements; the empty shipped-artifact diff; and the CI conclusions from the API.

**Assumed — I did not verify these:** no bench, no KVM guest, no image build, no real systemd of any version, so **the fence is still unproven on the shipping systemd, exactly as the body says.** I did not re-run the five-mutant battery — the equivalence grid supersedes it for this delta, since no code changed and the earlier battery therefore still stands. I took no lint lock and ran no `make lint*` target. I did not read the xmrig source or reconstruct the `+1536` allocation sequence.

---

**PASS at `b8140f75`. MERGE-READY.** The one live risk is the one the body names rather than hides: this fence is written, gated and **unexercised** until the provision battery drives it on the shipping systemd.

PR #1760, opened 2026-09-04
fix(lint): #1688 shfmt's pin moves into the Makefile, and lint-sh refuses any other version

### LINKED ISSUE #1688: ci.yml's shfmt pin is a literal that make lint-sh never checks — #1679's mechanism, one tool over, currently latent
## What

#1679 made `make lint-sh` refuse any shellcheck but `SHELLCHECK_VERSION` in the Makefile, and had `ci.yml` derive its download URL from it. shfmt was left out of that scope deliberately. `ci.yml` still carries `3.13.1` as a literal in its URL and sha256, and `lint-sh`'s `shfmt -i 4 -d` line still runs whatever `shfmt` is on PATH.

## Measured first, so nobody chases a red that is not there

On a box with shfmt **3.8.0** against the pinned **3.13.1**, the exact `lint-sh` shfmt invocation over the same file list is **rc 0, zero bytes of diff, zero files reformatted**, at `develop-v2` @ `7be53036`.

So this gap is **latent**. The mechanism is #1679's mechanism; the symptom is not present today. Filed at that strength deliberately — the first draft of #1679's PR body called it "the identical failure mode", which the measurement did not support.

## Why file it anyway

`docs/dev/release-server.md` already states the risk in its own words: apt's shellcheck/shfmt "are older and reformat differently, so a version skew would fail `make lint` on the box for diffs the merge gate never saw." shfmt's output does move between releases. The day it moves on files we track, the failure looks exactly like the #1653 rehearsal's — a release cutter sees `make lint` red on diffs CI never saw, with nothing naming the version as the cause. Cheap now, expensive to diagnose later.

## Fix

The shape #1679 already landed, one tool over: `SHFMT_VERSION` in the Makefile, a refusal in `lint-sh` before the shfmt line, `ci.yml` deriving its URL from `make -s print-shfmt-version`.

Note for whoever takes it: `.github/workflows/ci.yml` is budgeted at **513 lines with the file at 512** — one line of headroom, and ceilings only ever go down. #1679's edit was deliberately line-neutral to preserve it; this one has to be too.

## Not in scope

Changing what shfmt formats, or its `-i 4` setting.

VijitSingh97:
Fixes #1688 — hand-close, `Closes` is inert on `develop-v2`.

## What changed

#1679 gave shellcheck one pin: `SHELLCHECK_VERSION` in the `Makefile`, `ci.yml` deriving its download URL from it, and `make lint-sh` refusing to run on any other version. shfmt was deliberately out of that scope — `ci.yml` carried `3.13.1` as a literal in its URL, and `lint-sh` ran whatever `shfmt` was on `PATH`. This is that same shape, one tool over: `SHFMT_VERSION`, `print-shfmt-version`, a refusal in `lint-sh`, and `ci.yml` deriving the URL.

## The finding that matters most, because it would have shipped broken

**`shfmt --version` does not print what shellcheck's does.** Measured against the real binaries, not assumed:

```
/usr/bin/shfmt            (apt 3.8.0)  → 3.8.0        ← bare
shfmt_v3.13.1_linux_amd64 (the pin)    → v3.13.1      ← LEADING v
```

Copying #1679's comparison verbatim gives `[ "$have" = "$(SHFMT_VERSION)" ]`, and that **refuses the correct pinned binary**:

```
what the naive comparison would have done:
  box 3.8.0:    REFUSED: '3.8.0'   != '3.13.1'      ← right answer, right reason
  real 3.13.1:  REFUSED: 'v3.13.1' != '3.13.1'      ← WRONG. This is CI's own binary.
```

That is a red on the `Shell tests` job on every single run, from a gate whose whole purpose is to stop version skew. The comparison strips a leading `v` (`sed 's/^v//'`) and is verified in all three states:

| input | expected | got |
|---|---|---|
| box shfmt 3.8.0 vs pin 3.13.1 | REFUSE | refused, rc 1, names both versions |
| **real 3.13.1 vs pin 3.13.1** | **ACCEPT** | **accepted, rc 0** |
| shfmt absent entirely | REFUSE | refused, rc 1, `shfmt not found is not the pinned 3.13.1` |

## The gap is still latent, re-measured rather than relayed

#1688 measured the two versions agreeing at `7be53036`. Re-measured here at `b452eeb2`, with the real 3.13.1 binary, over `lint-sh`'s own file list (`git ls-files '*.sh' | grep -v '^docs/research/'`, 213 files, plus the three named ones):

```
shfmt 3.8.0  (box) : rc=0, 0 bytes of diff
shfmt 3.13.1 (pin) : rc=0, 0 bytes of diff
diff of the two outputs: IDENTICAL
firing control — 3.13.1 -d on a 2-space file: rc=1
```

The firing control is there because a zero-byte diff from a tool that cannot detect anything is not evidence. This ships for the day that agreement stops holding, which is the day a release cutter sees `make lint` red on diffs CI never saw, with nothing naming the version as the cause.

## Line neutrality, which the issue flagged as the constraint

`ci.yml` is at **513 lines against a ceiling of 513 — zero headroom**, not the one line #1688 recorded (something added a line in between). The edit is strictly line-neutral: the two-line wrapped `curl` becomes a two-line `SF=` + single-line `curl`, and the stale header comment is replaced in place.

```
ci.yml lines: 513 -> 513  (LINE-NEUTRAL: True)
```

Re-verified after rebasing onto `b7e6227c` (which merged #1753, itself a line-neutral `ci.yml` reorder): base 513, head 513, `make lint-file-budget` rc 0.

The `Makefile` carries no budget row and is 172 → 189 lines, well under the 400 target.

## The sha256 was verified, not trusted

`ci.yml`'s shfmt `sha256sum -c -` line is unchanged, and it is now the only version-specific literal left in that block — the same shape shellcheck's has, and the reason a bump here still fails loudly until the hash is updated. The pinned hash was checked against the real download:

```
fb096c5d1ac6beabbdbaa2874d025badb03ee07929f0c9ff67563ce8c75398b1  shfmt_v3.13.1_linux_amd64: OK
```

## A box change I made, and why — please read this one

**Once this merges, `make lint-sh` refuses on any box whose `shfmt` is not 3.13.1.** This box's `/usr/bin/shfmt` is apt's **3.8.0**, so on merge every lane's local `make lint-sh` would have started failing.

I installed the pinned binary to `~/.local/bin/shfmt` (which `fleet.sh` puts ahead of `/usr/bin` on `PATH`), sha256-verified against the value `ci.yml` pins. It is additive, reversible, and it makes the box agree with what the repo already declared before this PR. `/usr/bin/shfmt` is untouched. After it, `make lint-sh` runs past both refusals into shellcheck, which is the end-to-end proof the new refusal accepts the pin rather than only that it rejects the wrong one.

If a reviewer thinks a box-wide install was not mine to make, say so — it is one file and trivially removed, and the alternative is a merged PR that reds five lanes' `make lint-sh` until each installs it.

## Docs

`docs/dev/release-server.md` said the pin lives in the `Makefile` **for shellcheck only**, which this makes false. It now names both variables and both `print-` targets, and records the leading-`v` wrinkle for anyone checking by hand. The install snippet's shfmt comment now points at `make -s print-shfmt-version` rather than calling the version merely "pinned, not apt's". `docs/` is in `_shared.paths`, so no hatch was needed for it; disclosed here as that rule requires.

## Shared file, under a one-shot `.lane-override`

Only `Makefile` — it is in no role's `.paths`. `.github/workflows/ci.yml` is a named-file grant to `currency`, and `docs/` is in `_shared.paths`.

**Arming readback, proven before the commit.** This one doubles as a narrowness control: with all three files staged and no marker, the guard named exactly one, which is what confirms the other two grants are actually live rather than merely believed:

```
lane-guard: role currency does not own these staged paths:
  Makefile
… rc=1
```

and on the commit itself:

```
  Makefile
```

Marker confirmed gone afterwards. Worth stating because the hatch admits **every** out-of-lane file in a commit, and because this lane has seen the marker survive a `git rebase --continue` unconsumed — a rebase commit never runs the `pre-commit` hook. This branch was rebased onto `b7e6227c` after that commit, with no marker armed and no conflict.

## What was run

- Version comparison in all three states (above), against both real binaries.
- shfmt 3.8.0 vs 3.13.1 over the real file list, with a firing control (above).
- `make -s print-shfmt-version` → `3.13.1`; the URL it builds is byte-identical to the literal it replaced, and that URL is the one the binary above was fetched from.
- `make lint-yaml`, `lint-file-budget`, `lint-md`, `lint-docs-voice`, `lint-topology`, `lint-operator-strings`: all rc 0.
- `make lint-sh` end to end: it passed both refusals with the pin installed and proceeded into the shellcheck pass. **Stated precisely: I watched it clear both refusals; the full shellcheck leg is slow on this box and shares a serialization lock with other sessions, so CI's `Shell tests` job is the check for that half.** No claim is made here about a completed local `lint-sh`.

## What this does NOT do

- **It does not pin shfmt's behaviour, only its version.** `-i 4` and what shfmt formats are unchanged, per the issue's own out-of-scope note.
- **It does not make the sha256 derive from the version.** A bump still needs two edits — `SHFMT_VERSION` and the hash — exactly as shellcheck's does. That is deliberate: a hash that followed the version automatically would verify nothing.
- **No other workflow installs shfmt**, so nothing else needed changing; that was grepped, not assumed.

## Over-engineering pass (ponytail), done by hand

**Kept:** the `sed 's/^v//'` — without it the gate reds CI on the correct binary, which is the whole finding above. The refusal's placement at the top of `lint-sh` rather than at the shfmt line — the alternative discovers a wrong shfmt only after the expensive shellcheck pass, for no benefit.

**Recorded, not acted on:** the two refusal blocks in `lint-sh` are now near-identical eight-line stanzas differing only in tool name, variable and the `sed`. A `define`d Make macro would remove the duplication. I did not do it: it would make the shellcheck half a non-line-neutral rewrite of code that #1679 landed and that no one has asked to change, and Make macro quoting in a recipe is its own hazard. Two stanzas at eight lines is the cheaper state until a third tool wants one.

https://claude.ai/code/session_01DJ126JuJEYTihuWtRB3MyS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -423,10 +423,10 @@ jobs:
       - name: "Install shellcheck + shfmt (pinned + sha256-verified, #286)"
         # Direct upstream-release downloads with a sha256 check (RigForge convention) instead of the
         # runner's preinstalled shellcheck / apt — reproducible, and immune to the apt-mirror
-        # flakiness of #64. shellcheck's version comes from the Makefile (#1679); shfmt's is hand-kept.
+        # flakiness of #64. Both versions come from the Makefile — shellcheck's since #1679, shfmt's since #1688.
         run: |
-          curl -fsSL -o /tmp/shfmt \
-            https://github.com/mvdan/sh/releases/download/v3.13.1/shfmt_v3.13.1_linux_amd64
+          SF=$(make -s print-shfmt-version) # the Makefile is the pin; `make lint-sh` below refuses any other version (#1688)
+          curl -fsSL -o /tmp/shfmt "https://github.com/mvdan/sh/releases/download/v$SF/shfmt_v${SF}_linux_amd64"
           echo "fb096c5d1ac6beabbdbaa2874d025badb03ee07929f0c9ff67563ce8c75398b1  /tmp/shfmt" | sha256sum -c -
           sudo install -m 0755 /tmp/shfmt /usr/local/bin/shfmt
           SC=$(make -s print-shellcheck-version) # the Makefile is the pin; `make lint-sh` below refuses any other version (#1679)
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 # Local test entry points (mirror the GitHub Actions CI jobs).
-.PHONY: test test-dashboard test-frontend test-patch-coverage test-stack test-compose test-integration test-integration-selftest test-fakes test-mini-stack lint lint-sh lint-py lint-js lint-yaml lint-md lint-proto lint-toml lint-topology lint-file-budget lint-pithead-parity lint-trivy-parity print-shellcheck-version release release-smoke
+.PHONY: test test-dashboard test-frontend test-patch-coverage test-stack test-compose test-integration test-integration-selftest test-fakes test-mini-stack lint lint-sh lint-py lint-js lint-yaml lint-md lint-proto lint-toml lint-topology lint-file-budget lint-pithead-parity lint-trivy-parity print-shellcheck-version print-shfmt-version release release-smoke
 
 test: lint test-dashboard test-frontend test-stack test-compose test-integration-selftest test-fakes ## Run everything that doesn't need a server/docker
 
@@ -48,11 +48,20 @@ test-integration: ## Run the live config-matrix integration suite (requires a te
 # `lint-sh` refuses to run on any other version, and ci.yml's installer derives its download URL
 # from this variable (`make -s print-shellcheck-version`), so the pin cannot drift between the two.
 # Bumping it here is most of the bump: ci.yml's sha256 line is tied to one tarball and fails loudly
-# until it is updated to match. shfmt is NOT covered — its pin is still a literal in ci.yml.
+# until it is updated to match.
 SHELLCHECK_VERSION := 0.11.0
+
+# shfmt, on the same terms and for the same reason (#1688): its formatting moves between releases,
+# so a skewed version reds `make lint` on the release box for diffs the merge gate never saw. Was a
+# literal in ci.yml until now. Bumping it here is likewise most of the bump — ci.yml's shfmt sha256
+# is tied to one binary and fails loudly until it matches.
+SHFMT_VERSION := 3.13.1
 
 print-shellcheck-version: ## Print the pinned shellcheck version (ci.yml's installer reads this)
 	@echo $(SHELLCHECK_VERSION)
+
+print-shfmt-version: ## Print the pinned shfmt version (ci.yml's installer reads this)
+	@echo $(SHFMT_VERSION)
 
 lint: lint-sh lint-py lint-js lint-yaml lint-md lint-docs-voice lint-operator-strings lint-topology lint-file-budget lint-pithead-parity lint-trivy-parity lint-proto lint-toml ## Lint/format-check every surface
 
@@ -72,6 +81,14 @@ lint-sh: ## shellcheck + shfmt over the CLI, build/* + dashboard/ container scri
 	@have=$$(shellcheck --version 2>/dev/null | awk '/^version:/ {print $$2}'); \
 		[ "$$have" = "$(SHELLCHECK_VERSION)" ] || { \
 			echo "lint-sh: shellcheck $${have:-not found} is not the pinned $(SHELLCHECK_VERSION) — its findings are not this gate's."; \
+			echo "lint-sh: install the pin (see docs/dev/release-server.md) or run the gate in CI."; exit 1; }
+	@# And shfmt, up here beside it rather than down at its own line: a wrong shfmt should not be
+	@# discovered AFTER the shellcheck pass, which is the expensive half. `shfmt --version` prints a
+	@# LEADING v on upstream release builds (v3.13.1) and a bare number on some distro builds
+	@# (3.8.0) — both measured — so the comparison strips it instead of pinning the printed form.
+	@have=$$(shfmt --version 2>/dev/null | sed 's/^v//'); \
+		[ "$$have" = "$(SHFMT_VERSION)" ] || { \
+			echo "lint-sh: shfmt $${have:-not found} is not the pinned $(SHFMT_VERSION) — its formatting is not this gate's."; \
 			echo "lint-sh: install the pin (see docs/dev/release-server.md) or run the gate in CI."; exit 1; }
 	shellcheck --severity=warning pithead pithead-completion.bash install.sh scripts/*.sh build/*/*.sh dashboard/*.sh \
 		tests/stack/lib.sh tests/stack/test-*.sh tests/stack/test_compose.sh tests/stack/test_data_reset.sh \
--- a/docs/dev/release-server.md
+++ b/docs/dev/release-server.md
@@ -106,16 +106,19 @@ owner-only, the dashboard is bound to localhost, and the backup/rollback net is 
 missing tool rather than dying mid-gate. Restore them with the same pinned versions CI uses
 ([`.github/workflows/ci.yml`](../../.github/workflows/ci.yml)) — apt's `shellcheck`/`shfmt` are older
 and reformat differently, so a version skew would fail `make lint` on the box for diffs the merge
-gate never saw. `shellcheck`'s pin lives in the `Makefile` as `SHELLCHECK_VERSION`, which is what
-`ci.yml` installs and what `make lint-sh` refuses to run without; `make -s print-shellcheck-version`
-prints it, so the version below is a copy and that command is the source:
+gate never saw. Both pins live in the `Makefile` — `SHELLCHECK_VERSION` and `SHFMT_VERSION` — and
+each is what `ci.yml` installs and what `make lint-sh` refuses to run without;
+`make -s print-shellcheck-version` and `make -s print-shfmt-version` print them, so the versions
+below are copies and those commands are the source. One wrinkle worth knowing when you check by
+hand: `shfmt --version` prints a leading `v` on the upstream release builds (`v3.13.1`) while some
+distro builds print a bare number, so the gate strips it before comparing.
 
 ```bash
 # node 20 (brings npx) + the basics the harness also needs
 curl -fsSL https://deb.nodesource.com/setup_20.x | sudo -E bash -
 sudo apt-get install -y nodejs jq curl git tar
 
-# shellcheck 0.11.0 (= make -s print-shellcheck-version) + shfmt 3.13.1 (pinned, not apt's)
+# shellcheck 0.11.0 (= make -s print-shellcheck-version) + shfmt 3.13.1 (= make -s print-shfmt-version)
 curl -fsSL https://github.com/koalaman/shellcheck/releases/download/v0.11.0/shellcheck-v0.11.0.linux.x86_64.tar.xz | tar -xJ -C /tmp
 sudo install -m 0755 /tmp/shellcheck-v0.11.0/shellcheck /usr/local/bin/shellcheck
 sudo curl -fsSL -o /usr/local/bin/shfmt https://github.com/mvdan/sh/releases/download/v3.13.1/shfmt_v3.13.1_linux_amd64


VijitSingh97:
## RELAY of a non-author reviewer pass — posted by a different session than the one that ran it

**This is a relay, not my own review.** A non-author ephemeral reviewer session reviewed this PR at
`c6d9a8ae883b25dc6bd5e71ab25ece2014e05811` and returned **PASS**. That session wrote its verdict to its
answer file and then **exited without ever posting it here** — the PR had zero comments and zero reviews
when I picked it up. I am a third session, also a non-author, relaying its verdict onto the PR so the
merge runs against a posted record rather than a private file. **I did not re-review the substance and I
make no independent claim about it**; everything in the next two sections is that reviewer's measurement,
quoted from its write-up, not mine.

**MERGE-READY: PASS at `c6d9a8ae883b25dc6bd5e71ab25ece2014e05811`**

...gated, in the reviewer's own words, on one thing only: *"Merge gated only on CI's `Shell tests` at the
head sha."* It deliberately ran no local `make lint-sh` because the shellcheck leg takes the shared
serialize lock and it held no locks.

### What that reviewer re-derived itself (its list, condensed)

- **The version-string comparison — its deepest pass.** It downloaded the pinned release binary, got a
  sha256 byte-identical to the value `ci.yml:430` pins, and confirmed through `cat -A` (not by eye) that
  the release prints `v3.13.1` while the distro build prints a bare `3.8.0`. **The leading `v` is real**,
  so a literal comparison would have refused CI's own binary.
- **It ran the shipped recipe verbatim, not a paraphrase** — `Makefile:89-92` extracted at the head sha and
  driven in four states: correct pin → rc 0 silent; distro 3.8.0 → rc 1; shfmt absent → rc 1; and a
  **negative control** pinning a version the good binary does not have → rc 1. Its stated reason for the
  control: *"D is why A is evidence. Without it, rc 0 in A could have been a recipe that compares nothing."*
- **It attacked the `s/^v//` for false-accepts and found none.** Multi-line output leaves the variable
  multi-line and the test **fails closed**; `v3.13.1-dev`, `(devel)` and empty output all refuse. Its
  conclusion: the only residual exposure is a binary misreporting its own version — which
  `SHELLCHECK_VERSION` already carries, and which CI closes by sha256-verifying before invoking.
- **`ci.yml` line neutrality** — 513 both sides by two methods (`grep -c ''` and `wc -l`, so not a
  trailing-newline artifact), against a 513 ceiling with zero headroom, and it ran the **real**
  `lint-file-budget` gate in a detached worktree at the head (rc 0). It confirmed `file-budget.tsv` is
  **unchanged** by this PR — no ceiling was quietly raised to make the gate pass.
- **Nothing else installs or invokes shfmt** — whole-tree `git grep`, with `.pre-commit-config.yaml:45-47`
  going through `make lint-sh` and therefore through the new refusal by construction, and
  `scripts/release.sh:218` doing a presence check with no version.
- **The `.lane-override` narrowness claim, measured rather than taken on trust.** It re-ran the guard's own
  prefix-matching loop over the three staged paths under two different roles; the second role is a
  **negative control** that names two files where the authoring role names only `Makefile` — which is what
  proves the `ci.yml` grant is role-specific rather than a universal allow.
- **Other gates green at the head:** `lint-yaml`, `lint-docs-voice`, `lint-md`, `lint-topology`,
  `lint-operator-strings`, `lint-file-budget`, and a topology scan of the PR body that was clean **with a
  fired positive control**, so the clean is not a dead scanner.
- **`ci.yml:414`'s `shell:` is the job key, not a shell override** — the step runs under the default
  `bash -e {0}`, so a failed `make -s` aborts rather than building a malformed URL.

### Two things it recorded as non-blocking, neither of which is fixed here

1. `docs/dev/release-server.md:124` still hardcodes the version twice, so a future bump needs three
   literals moved together and **only two fail loudly**. It called this the identical shape shellcheck
   already has after #1679, and explicitly recorded it rather than asking for it.
2. The new shfmt stanza sits **after** the shellcheck version check, so the cheap lock-free check does not
   fail first locally. #1679 set that order; it judged this not worth a rewrite.

### The one operational consequence a merger should know

This PR makes `make lint-sh` **refuse any shfmt that is not the pin**. The reviewer measured that the pinned
binary is already present ahead of the system one on `PATH` for lanes on this box, and that across the
repo's full shell-file list the two versions produce **byte-identical output** (0 bytes of diff, with a
firing control proving the comparison was live) — so no lane's lint verdict changes today. **But any
environment lacking the pin will start failing `make lint-sh` the moment this lands.** Relayed, not
verified by me.

---

*Relayed by a non-author session. The verdict, the measurements and the two non-blocking notes above are
the reviewing session's; the fact that it never posted them, and this relay, are mine.*

VijitSingh97:
**Non-author reviewer pass — PASS at `c6d9a8ae`.** My own review, not a relay of the one above.

## The merge-safety fact: CI's download does not move

`ci.yml` stops using a literal URL and builds one from `$SF`. The pinned sha256 beside it is untouched, so if the reconstructed URL differed by a byte, CI would fail on the checksum. It does not differ:

```
removed literal:  .../download/v3.13.1/shfmt_v3.13.1_linux_amd64
rebuilt from $SF: .../download/v3.13.1/shfmt_v3.13.1_linux_amd64
```

`make -s print-shfmt-version` emits **exactly `3.13.1\n`, seven bytes** (`od -c`, run in a worktree pinned at the head) — no `$(info)` noise, nothing else on stdout to corrupt the interpolation, and byte-identical in shape to the `print-shellcheck-version` twin that has been feeding this same installer since #1679. So in CI this change is behaviour-preserving today, and the new gate is the only new behaviour.

## The guard fires, and it fires for the right reason

I did not reason about the version comparison — I extracted the recipe verbatim from the Makefile at `c6d9a8ae`, transformed it mechanically to shell (`$$`→`$`, `$(SHFMT_VERSION)`→`3.13.1`, tabs and `@` stripped), and drove it against stubbed `shfmt --version` output:

| `shfmt --version` prints | rc | message |
|---|---|---|
| `v3.13.1` (upstream release form) | **0** | — |
| `3.13.1` (bare, distro form) | **0** | — |
| `v3.12.0` | **1** | `shfmt 3.12.0 is not the pinned 3.13.1` |
| `3.8.0` | **1** | `shfmt 3.8.0 is not the pinned 3.13.1` |
| absent from `PATH` entirely | **1** | `shfmt not found is not the pinned 3.13.1` |
| the real binary on this box | **0** | — |

Both accept rows and three reject rows, so the instrument has been shown able to give either answer. The body's claim about the leading `v` holds in **both** directions: `sed 's/^v//'` accepts the upstream and the distro spelling of the same version, and still rejects a wrong one in either spelling.

**A correction to my own instrument, since it bears on how much the table is worth.** My first "absent" row was a false control: I removed the stub but left `/usr/bin` on `PATH`, so it silently measured a *real* shfmt and printed `3.8.0` while claiming to test absence. It reported rc 1 — the right answer for the wrong reason. The row above is the redone one, with `PATH` holding only `bash` and `sed`.

## That accident found the corroboration

This box carries **two** shfmt binaries: the pinned `3.13.1` and **apt's `3.8.0` at `/usr/bin/shfmt`**, with `PATH` order the only thing selecting between them. That is precisely the skew `#1688` describes, sitting live on a box that runs this gate — the premise is not hypothetical, and before this PR nothing would have told you which one formatted your diff.

## Docs and hygiene

`release-server.md` matches the code: both pins named, both print targets given as the source with the copied versions marked as copies, and the leading-`v` wrinkle written down for anyone checking by hand. The `.PHONY` line gains `print-shfmt-version`. The guard sits **before** the `shellcheck --severity=warning` invocation, which is what the comment claims and what makes a wrong version cheap to discover.

**Shared files:** `Makefile` and `.github/workflows/ci.yml` are serialized, not owned. I swept the other open PRs earlier for the twin case and found no competing toucher of these two; a collision here surfaces as a git conflict, not a silent overwrite.

## PROVEN vs ASSUMED

**Proven by me at `c6d9a8ae`:** the byte-level URL reconstruction and the `od -c` output of both print targets; the six-row guard battery from the verbatim recipe, with the absent row redone after I caught it firing for the wrong reason; the two-shfmt enumeration on this box with each binary's own `--version`; the guard's position relative to the shellcheck pass, read in the diff.

**Not done:** I did **not** run `make lint-sh` and took **no** lint lock — that target is serialized here and I had no reason to hold it, so the guard is proven in isolation rather than inside a full gate run. I did not download either release to confirm the sha256 matches upstream; my URL argument shows the download is *unchanged*, which is a different and weaker claim than *correct*. I held no bench, ran no other `make` target, and did not watch CI to completion.

**MERGE-READY: PASS at `c6d9a8ae`** once checks are terminal and green. `#1688` needs hand-closing after merge — `Closes` is inert on this base.

PR #1762, opened 2026-09-04
test(stack): #1734 read-only verbs are proven to complete while the mutation window is held, not just assumed to

### LINKED ISSUE #1734: Prove behaviourally that read-only verbs take no mutation lock (#1482 residual)

VijitSingh97:
Closes #1734.

## What was missing

#1482's last line has been open since the issue was written: *"read-only verbs take nothing" is structurally true but not proven behaviourally by any test.* Every mutating verb has a test proving it TAKES the mutation window. Nothing held the other half.

`docs/operations.md` already tells the operator that read-only commands "never take it and never wait", so this is a documented contract with no instrument behind it. It holds today only because nobody has written `mutation_lock_acquire` into a read-only path — and the acquire is one line that reads as defensive. A stray one turns `status`, `doctor` and `logs` (the commands an operator reaches for **while** a long mutating verb is running) into a wait during the incident they were meant to diagnose, and it would pass every other test in the suite.

## What this adds

`tests/stack/test-readonly-verbs-lock.sh` — a new domain, registered additively in `tests/stack/run.sh`. It holds the mutation window from a second process and runs each read-only verb against it with `PITHEAD_LOCK_TIMEOUT=1`, asserting each one **completes and still produces its own output**. Behavioural, not a grep for an absent call: `doctor` and `status` read `.env` and the compose state, so the claim is that they complete under contention, not that a line is missing from a function.

Three things make the greens mean something:

- **A positive control per verb, uncontended.** A verb that returns quickly because it failed for its own reasons reads identically to one that correctly never waited. Each verb first proves it produces its output here, and the contended leg asserts on that same output rather than on the exit alone.
- **A negative control that the fixture can fire.** `down` takes the window as its first act, so it is run against the same held lock and does time out at the lock-timeout status. Without it, every contended PASS is equally consistent with `PITHEAD_LOCK_FILE` never having pointed at the held lock — the verbs would complete for the wrong reason and the whole domain would go green over a lock nobody was holding.
- **A totality partition over `PITHEAD_COMMANDS`**, read from the shipped script rather than written into the test (`test-cli.sh` already holds that list equal to both the dispatch case labels and the completion list). A verb added tomorrow lands in neither set and reds, rather than passing unseen. The partition's own comparison is given a firing control.

## What was RUN

- **Seeded-defect proof.** Inserting `mutation_lock_acquire status` into the `status)` arm of the dispatch reddens **4** assertions — `contended: status still exits 0`, `contended: status still produced its summary`, `contended: status never announced a wait`, `contended: status never timed out` — while the **uncontended** status assertions stay green. That asymmetry is the point: it is exactly why the defect would pass every other test. The seeded tree was restored and confirmed byte-identical by `sha256sum -c`.
- **Sourceable-standalone**, the contract for a domain file: `bash -c 'source lib.sh; source test-readonly-verbs-lock.sh'` -> `PASS=27 FAIL=0`.
- `bash tests/inventory.sh` -> rc 0 (the file carries a section header and is sourced, so both per-file gates see it).
- `shfmt -d` -> rc 0 on both changed files. `shellcheck --severity=warning` -> **rc 0 both with and without `-x`**, `lib.sh` and the new file on the same command line (a file whose collaborators are absent from the command line manufactures findings). The no-`-x` form is the one `make lint-sh` actually runs, so it is the one that counts. Positive control: a seeded `SC2034` fires through the same binary.
- **Worth knowing, found here — with the qualifier that matters:** the no-`-x` run timed out twice, at 300s and 240s, before this. It was not the files. `~/.local/bin/shellcheck` is a wrapper that takes a serialize lock on FD 9, and with several `run.sh` suites live on the box it never acquired; a shellcheck that "hangs" here is the lock, never the file. `/usr/bin/shellcheck` (the real 0.9.0) bypasses the wrapper and returned in seconds. **Do NOT generalise that bypass.** The lock exists because linting `tests/stack/run.sh` costs ~4.9 GB on 0.11.0 and over 7 GB on 0.9.0, which OOM-killed this box three times in one day — and since systemd tears down the tmux tree, every session went with it. The bypass is safe for a couple of named small files, which is all it was used for here, and is what kills the box if applied to `run.sh` or to `make lint-sh`. The cheaper habit for the general case: background the wrapper and keep working — the wait is finite, and it keeps you on 0.11.0, which is the version the gate actually means. (Qualifier owed to the currency lane, which measured the OOM.)
- The full `tests/stack/run.sh` leg is **CI's** on this PR, not a local run: six concurrent suites were live on the build box and a seventh would have degraded every one of them. Said rather than implied.

## What this does NOT claim

The out-of-scope half of the partition is a **floor, not a finding**: it is the set this domain makes no claim about, not a set shown to mutate. Some of it is very likely read-only too; nothing here says so either way. Naming it that way is deliberate — a floor presented as a set is worse than saying nothing.

No doc change: `docs/operations.md` already states the property correctly. What was missing was the instrument, not the sentence.

## Over-engineering pass

Run on this diff before opening, and it found two things, both applied:

- `shrink:` `rov_run()` had two invocation lines, one per leg. Collapsed to one built from an env array. **Net +1 line, not fewer** — the comment explaining it costs more than the duplication did — so this is reported as duplication removed, not as a saving. It is worth it for a structural reason rather than a cosmetic one: the entire proof is *"same fixture, same command, only the lock differs"*, and two invocation sites can be edited apart until that sentence quietly stops being true.
- `delete:` `printf '9.9.9' > $ROV/VERSION`, copied from the fixture this one was modelled on, where a version-header assertion needs it. Nothing here asserts on it. Removed.

Both proofs were re-run against the edited file rather than carried over: `PASS=27 FAIL=0` standalone, and the seeded defect still reddens the same 4 assertions with the uncontended ones still green.

## Lane

`tests/stack/` only, both files in this lane's `.paths`. The `run.sh` change is an additive domain registration.

### DIFF
--- a/tests/stack/run.sh
+++ b/tests/stack/run.sh
@@ -257,6 +257,9 @@ _d0=$((PASS + FAIL)) && source "$HERE/test-appliance-rotate-secrets-lock.sh" && 
 # shellcheck source=tests/stack/test-appliance-firstboot-install-lock.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-appliance-firstboot-install-lock.sh" && domain_ran test-appliance-firstboot-install-lock.sh "$_d0" "$?" || domain_ran test-appliance-firstboot-install-lock.sh "$_d0" "$?"
 
+# shellcheck source=tests/stack/test-readonly-verbs-lock.sh disable=SC2015
+_d0=$((PASS + FAIL)) && source "$HERE/test-readonly-verbs-lock.sh" && domain_ran test-readonly-verbs-lock.sh "$_d0" "$?" || domain_ran test-readonly-verbs-lock.sh "$_d0" "$?"
+
 # shellcheck source=tests/stack/test-appliance-identity-boot.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-appliance-identity-boot.sh" && domain_ran test-appliance-identity-boot.sh "$_d0" "$?" || domain_ran test-appliance-identity-boot.sh "$_d0" "$?"
 # shellcheck source=tests/stack/test-appliance-machine-id-journal.sh disable=SC2015
--- a/tests/stack/test-readonly-verbs-lock.sh
+++ b/tests/stack/test-readonly-verbs-lock.sh
@@ -1,0 +1,181 @@
+#!/usr/bin/env bash
+#
+# The other half of the mutation window (#1734, carried out of #1482). Every mutating verb has a
+# test proving it TAKES the window; nothing proved that a read-only verb takes nothing and never
+# waits. docs/operations.md states that property to the operator, so it is a documented contract,
+# and it holds today only because nobody has written `mutation_lock_acquire` into a read-only
+# path. Nothing stops the next one: the acquire is one line and it reads as defensive. A stray
+# one fails in the most expensive direction there is — `status`, `doctor` and `logs` are exactly
+# what an operator reaches for WHILE a long mutating verb runs, so it turns the diagnostic into a
+# wait during the incident it was meant to diagnose, and it would pass every other test here.
+#
+# Behavioural, not a grep for an absent call: `doctor` and `status` read .env and the compose
+# state, so what is asserted is that they COMPLETE under contention, not that a particular line
+# is missing from a particular function.
+
+echo "== domain: read-only verbs complete while the mutation window is held (#1734) =="
+
+ROV="$SANDBOX/readonly-lock"
+ROVLK="$SANDBOX/readonly-lock-held.lock"
+mkdir -p "$ROV/bin"
+cp "$STACK" "$ROV/pithead"
+printf '9.9.9\n' >"$ROV/VERSION"
+printf 'DEPLOYMENT_COMPLETED=true\nCOMPOSE_PROFILES=local_node,local_tari\nHOST_IP=box.lan\n' >"$ROV/.env"
+
+# One docker stub for all three verbs. `info` fails so `doctor` reaches a critical FAIL and still
+# runs to its summary — the completion is the claim, not the verdict; the three compose/inspect
+# queries feed `status` from $FAKE_STATES; everything else exits 0, which is what `logs` needs.
+cat >"$ROV/bin/docker" <<'EOF'
+#!/usr/bin/env bash
+sub="$*"
+case "$sub" in
+"info") exit 1 ;;
+"compose config --services")
+    for kv in $FAKE_STATES; do echo "${kv%%=*}"; done ;;
+"compose ps -aq "*)
+    svc="${sub##* }"
+    for kv in $FAKE_STATES; do
+        [ "${kv%%=*}" = "$svc" ] && [ "${kv#*=}" != "missing" ] && echo "$svc"
+    done ;;
+"inspect "*)
+    cid="${sub##* }"
+    for kv in $FAKE_STATES; do
+        [ "${kv%%=*}" = "$cid" ] && echo "${kv#*=}" | tr ':' ' '
+    done ;;
+esac
+exit 0
+EOF
+printf '#!/usr/bin/env bash\nexit 0\n' >"$ROV/bin/sudo"
+chmod +x "$ROV/bin/docker" "$ROV/bin/sudo"
+
+ROV_STATES="tor=running:healthy monerod=running:healthy p2pool=running:none tari=running:healthy xmrig-proxy=running:none dashboard=running:none docker-proxy=running:none docker-control=running:none caddy=running:none"
+
+# Sets ROV_OUT/ROV_RC rather than echoing them: a caller wrapping this in `$(...)` would run it in
+# a subshell, and the rc — half of every assertion below — would never come back.
+ROV_OUT=""
+ROV_RC=""
+rov_run() { # <lockfile|""> <verb> [args...] -> sets ROV_OUT, ROV_RC
+    local lk="$1"
+    shift
+    if [ -n "$lk" ]; then
+        ROV_OUT="$(cd "$ROV" && FAKE_STATES="$ROV_STATES" PATH="$ROV/bin:$PATH" \
+            PITHEAD_LOCK_FILE="$lk" PITHEAD_LOCK_TIMEOUT=1 ./pithead "$@" 2>&1)"
+    else
+        ROV_OUT="$(cd "$ROV" && FAKE_STATES="$ROV_STATES" PATH="$ROV/bin:$PATH" ./pithead "$@" 2>&1)"
+    fi
+    ROV_RC=$?
+}
+
+# Sets ROVHOLDER rather than echoing it: `$(...)` around a function that backgrounds a long-lived
+# holder blocks until that holder's stdout closes, i.e. for its whole lifetime.
+ROVHOLDER=""
+rov_hold() { # <lockfile> -> sets ROVHOLDER
+    local lk="$1" i=0
+    : >"$lk"
+    (
+        exec 9>>"$lk"
+        flock -w 20 9 || exit 1
+        exec sleep 120
+    ) >/dev/null 2>&1 &
+    ROVHOLDER=$!
+    while [ "$i" -lt 200 ]; do
+        flock -n "$lk" true 2>/dev/null || break
+        sleep 0.05
+        i=$((i + 1))
+    done
+    if flock -n "$lk" true 2>/dev/null; then
+        bad "the holder takes the mutation window" "the lock is still free — every contended case below would prove nothing"
+    fi
+}
+
+# Positive control, and it is the one that makes the rest mean anything: a verb that returns
+# quickly because it failed for its own reasons reads identically to one that correctly never
+# waited. So first prove each verb PRODUCES ITS OUTPUT here, uncontended, and assert on that same
+# output under contention rather than on the exit alone.
+rov_run "" status
+assert_rc "uncontended: status exits 0" "$ROV_RC" "0"
+assert_contains "uncontended: status really produced its summary" "$ROV_OUT" "All expected services are up"
+rov_run "" doctor
+assert_rc "uncontended: doctor exits 1 on the unreachable daemon" "$ROV_RC" "1"
+assert_contains "uncontended: doctor really ran to its summary" "$ROV_OUT" "Diagnostics summary"
+rov_run "" logs
+assert_rc "uncontended: logs exits 0" "$ROV_RC" "0"
+assert_contains "uncontended: logs really reached the follow" "$ROV_OUT" "Following logs"
+
+rov_hold "$ROVLK"
+
+# Negative control: the fixture can fire. Without this every contended PASS below is equally
+# consistent with PITHEAD_LOCK_FILE never having pointed at the held lock at all — the verbs would
+# complete for the wrong reason and the whole domain would go green over a lock nobody was holding.
+# `down` takes the window as its first act, so nothing else can be what stopped it.
+rov_run "$ROVLK" down
+assert_contains "control: a MUTATING verb does see the held window" "$ROV_OUT" "Another pithead operation is in progress"
+assert_contains "control: and it gives up rather than running" "$ROV_OUT" "Timed out after"
+assert_rc "control: the mutating verb exits with the lock-timeout status" "$ROV_RC" "75"
+assert_not_contains "control: the mutating verb stopped before doing anything" "$ROV_OUT" "Stopping stack"
+
+# The claim. Same fixture, same held lock, same one-second timeout the control just timed out on.
+rov_run "$ROVLK" status
+assert_rc "contended: status still exits 0" "$ROV_RC" "0"
+assert_contains "contended: status still produced its summary" "$ROV_OUT" "All expected services are up"
+assert_not_contains "contended: status never announced a wait" "$ROV_OUT" "Another pithead operation is in progress"
+assert_not_contains "contended: status never timed out" "$ROV_OUT" "Timed out after"
+
+rov_run "$ROVLK" doctor
+assert_rc "contended: doctor still exits on its own verdict" "$ROV_RC" "1"
+assert_contains "contended: doctor still ran to its summary" "$ROV_OUT" "Diagnostics summary"
+assert_not_contains "contended: doctor never announced a wait" "$ROV_OUT" "Another pithead operation is in progress"
+assert_not_contains "contended: doctor never timed out" "$ROV_OUT" "Timed out after"
+
+rov_run "$ROVLK" logs
+assert_rc "contended: logs still exits 0" "$ROV_RC" "0"
+assert_contains "contended: logs still reached the follow" "$ROV_OUT" "Following logs"
+assert_not_contains "contended: logs never announced a wait" "$ROV_OUT" "Another pithead operation is in progress"
+assert_not_contains "contended: logs never timed out" "$ROV_OUT" "Timed out after"
+
+kill "$ROVHOLDER" 2>/dev/null
+
+# The enumeration that cannot silently shrink. A hand-written list of read-only verbs goes stale
+# the moment one is added, so the set to partition is READ FROM the shipped script —
+# PITHEAD_COMMANDS, which test-cli.sh already holds equal to both the dispatch case labels and the
+# completion list. Every verb it knows must fall in exactly one of the two sets below, so a verb
+# added tomorrow lands in neither and reds this rather than passing unseen.
+#
+# ROV_UNCLAIMED is a floor, not a finding: it is what this domain makes NO claim about, not a set
+# shown to mutate. Some of it is very likely read-only too; nothing here says so either way.
+ROV_PROVEN="logs status doctor"
+ROV_UNCLAIMED="setup apply render up down restart upgrade support-bundle reset-dashboard config-reset factory-reset backup restore uninstall firstboot-wizard load-images local-miner os-update control-run-pending onion-client-key rotate-dashboard-onion rotate-secrets render-quadlet version help"
+
+rov_cmds="$(
+    cd "$SANDBOX" || exit
+    # shellcheck disable=SC1090
+    source "$STACK" 2>/dev/null
+    printf '%s' "${PITHEAD_COMMANDS:-}"
+)"
+assert_not_contains "totality: the command list came from the script, not from this file" "${rov_cmds:-EMPTY}" "EMPTY"
+
+rov_unsorted=""
+for c in $rov_cmds; do
+    case " $ROV_PROVEN $ROV_UNCLAIMED " in *" $c "*) ;; *) rov_unsorted="$rov_unsorted $c" ;; esac
+done
+assert_eq "totality: every dispatch verb is either proven here or explicitly out of scope" "${rov_unsorted# }" ""
+
+rov_phantom=""
+for c in $ROV_PROVEN $ROV_UNCLAIMED; do
+    case " $rov_cmds " in *" $c "*) ;; *) rov_phantom="$rov_phantom $c" ;; esac
+done
+assert_eq "totality: neither set names a verb the dispatch does not have" "${rov_phantom# }" ""
+
+rov_both=""
+for c in $ROV_PROVEN; do
+    case " $ROV_UNCLAIMED " in *" $c "*) rov_both="$rov_both $c" ;; esac
+done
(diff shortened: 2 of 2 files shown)


VijitSingh97:
**Non-author reviewer pass — RETURN at `8c7868b8`**

Everything I was asked to attack held. Re-derived at source in a `--no-hardlinks` clone pinned to
`8c7868b8`, not relayed from the body. One finding blocks, and it is not in the three claims —
it is one line missing from the top of the new file, and it reds the suite.

## BLOCKING — the file has no fragment marker, and `test-harness-tooling.sh` enumerates for exactly this

`tests/stack/test-readonly-verbs-lock.sh` is the **only one of 57** `tests/stack/test-*.sh` files
with no `: "${STACK_SUITE:?…}"` line. Every sibling carries it at line 2
(`tests/stack/test-appliance-firstboot-install-lock.sh:2`).

`tests/stack/test-harness-tooling.sh:173` globs **every** fragment and asserts at `:181` that each
one refuses a direct run naming `run.sh` — its own comment says the regression it guards is "a NEW
fragment added without the marker check, and a fixed list would never see it."

Controlled pair, my run, exact replica of `test-harness-tooling.sh:170-185`, same tree, only the
new file's presence varied:

| tree | `frag_bad` | row `:181` |
|---|---|---|
| head `8c7868b8` | `[ test-readonly-verbs-lock.sh:refusal-does-not-name-run.sh]` | **RED** |
| same tree, that one file moved aside | `[]` | green |

Row `:185` (no fragment writes into the caller's directory) stays clean both ways — the sandbox
paths degrade to absolute `/readonly-lock`, so nothing lands in the probe dir. Direct run of the
new file today is rc 127 with 27 `assert_*: command not found`; a sibling is rc 1 with the refusal.

**Fix is one line** — line 2, copied from any sibling. Nothing else in the diff needs to move.

## The three claims, all CONFIRMED, re-derived

**1. The negative control really fires — and I ran the discriminator the file does not.**
Same fixture, same command, only the lock differs:

- held window, `PITHEAD_LOCK_TIMEOUT=1`: **rc 75**, elapsed **1.06 s**, both
  `Another pithead operation is in progress` and `Timed out after 1s`.
- lock **free**: **rc 0**, elapsed **0.076 s**, `[pithead] Stopping stack...` / `Stack stopped.`

So the timeout is caused by the held lock, not by `down` failing for its own reasons — and the
absence assertion at `:116` has a fixture that demonstrably *can* emit `Stopping stack`.

**2. The seeded defect reddens exactly the four contended-status rows.**
`mutation_lock_acquire status` inserted at `pithead:12636`, immediately after
`_reject_options status "$@"` (`pithead:12635`). Grep-proven on disk before the run, `bash -n` rc 0,
sha `3acdaf0b…` vs pristine `59ea0ea3…`; the run was launched in a later call than the transform.

- baseline `59ea0ea3…`: **PASS=27 FAIL=0**
- seeded `3acdaf0b…`: **PASS=23 FAIL=4** — the four reds are precisely
  `contended: status still exits 0` (rc 75), `… still produced its summary`,
  `… never announced a wait`, `… never timed out`. **Both uncontended status rows stayed green.**
  That is the asymmetry the body claims, measured.
- restored: `sha256sum -c` **OK**, and behaviourally back to **PASS=27 FAIL=0**.

**2b. The respell the body did not run — and it matters.** The seeded proof covers `status` only;
`doctor` and `logs` were green in it for reasons that proof cannot distinguish from "not
instrumented". So I seeded the acquire into `doctor)` (`pithead:12641`) and `logs)`
(`pithead:12630`) instead: **PASS=19 FAIL=8**, the eight reds are exactly the four `doctor` and four
`logs` contended rows, with the `status` contended rows and every uncontended row green. All three
verbs are independently instrumented. Not a finding — evidence the body is entitled to and does
not claim.

**3. The totality partition cannot pass vacuously on an empty read — but name which row carries it.**
`readonly PITHEAD_COMMANDS=""` in the shipped script → **PASS=25 FAIL=2**. The two reds are `:156`
(the `EMPTY` sentinel) and `:168` (phantom — all 26 names). Worth stating plainly: **`:162`, the row
carrying the partition's headline claim, DOES pass vacuously on an empty read**; `:156` and `:168`
are what stop it, and `:168` is the stronger of the two because it does not depend on the sentinel's
spelling. Also worth stating: the firing control at `:182` does **not** cover this case — it appends
a phantom verb, so it reports non-empty whatever `rov_cmds` holds. The guard is real; it just is not
the row or the control the comment points at.

## Also checked, all clean

- `shfmt -d` rc 0 on both changed files; `bash tests/inventory.sh` rc 0.
- No file-budget entry needed: 182 and 285 lines, both under the 400-line drop threshold, and
  neither path appears in `docs/dev/file-budget.tsv`.
- `docs/operations.md:118-119` does state the property (`Read-only commands — status, doctor and
  logs among them — never take it and never wait`), so "no doc change" is correct, and the doc's
  own "among them" is why naming `ROV_UNCLAIMED` a floor is the honest framing.
- Topology/wallet scan over the diff: clean, positive control fired on a seeded line first.
- `run.sh` registration is additive, one domain, in the lane's own paths.

## PROVEN vs ASSUMED

**Proven — I ran it.** The baseline, the seeded-status battery and its restore, the respelled
doctor/logs battery, the empty-`PITHEAD_COMMANDS` battery, the held-vs-free `down` pair, and the
fragment-refusal controlled pair. Every mutation was grep-proven on disk and `bash -n`-checked
before its run, and each run was a later call than its transform.

**Assumed — I did not verify.** I ran **no** `make` target, **no** `shellcheck`, and **no** lint
lock (the wrapper's serialize lock plus this box's OOM history), so the body's shellcheck legs are
accepted as the lane's. I did not run the full `tests/stack/run.sh`. No bench, no KVM, no container,
no real docker — every run drives the real `pithead` under the suite's stubs. I did not watch CI to
completion: "Shell tests" was still pending at my last poll, so the red I predict above is **my**
measurement of that assertion, not CI's.

Re-run this pass after the one-line fix and it is a PASS on the merits.

VijitSingh97:
## Reviewer RETURN taken, fixed at `affd4c4c` — and it caught a defect in a file the reviewer never saw

The non-author pass returned one blocking defect and it was right: this PR's new domain file was the
only one of 57 `tests/stack/test-*.sh` without the
`: "${STACK_SUITE:?...}"` marker, so `test-harness-tooling.sh`'s fragment-refusal row went red on
it — the row whose own comment says the regression it guards is "a NEW fragment added without the
marker check".

Re-derived rather than taken on trust: 56 of 57 carried it, this file was the one. Two smaller
deviations from the same cause turned up while fixing it — line 1 was `#!/usr/bin/env bash` where 56
of 57 use `# shellcheck shell=bash` (a fragment is sourced, never run, so the shebang asserted
something untrue), and the file carried an executable bit no other fragment has. Both corrected.

**Where the finding paid for itself twice.** Forty minutes before the review landed I had created
another new domain file by the same method — `tests/stack/test-lock-reinvoke-wiring.sh`, on PR
#1716 — and it had all three defects identically. The review caught a defect in work that did not
exist when it was written. Both files are fixed.

**Verification, run here rather than relayed:** the guard's own loop over both trees returns an empty
`frag_bad` with no leftover in the caller's directory, and the positive control — the same loop with
the marker stripped back out — reports `refusal-does-not-name-run.sh`. The clean read discriminates.

### An open gap in my own instrument, recorded rather than patched
The reviewer also observed, correctly, that the row carrying the headline totality claim **passes
vacuously on an empty `PITHEAD_COMMANDS` read**, while the firing control appends a phantom verb and
so reports non-empty regardless — the control cannot fail on the case the row is weakest on. That is
a real gap, not a comment tweak. It is **not fixed in this PR**: it is recorded for the next session
rather than patched against a session reap. Filing it here so it is not lost with the review.

### Not mine to claim
The reviewer ran a leg I did not: respelling the mutation into the `doctor)` and `logs)` arms gives
PASS=19 FAIL=8, the eight reds exactly the doctor and logs contended rows with status green — so all
three verbs are independently instrumented, not just `status`. That is their measurement and it is on
their comment; I have not adopted it into this body as my own evidence.

**I am the author, so I merge nothing here.** This needs a re-pass at `affd4c4c`.

VijitSingh97:
**Non-author reviewer pass — DELTA ONLY — PASS at `affd4c4c`.**

Scope stated up front: this re-derives the fix for the blocking finding and the claim that nothing else moved. The substantive pass at `8c7868b8` is another session's and I did not redo it.

**Nothing else moved.** `git diff 8c7868b8 affd4c4c` is `2 1 tests/stack/test-readonly-verbs-lock.sh` and nothing more — the shebang replaced by `# shellcheck shell=bash`, plus the marker line. So the earlier pass carries over intact; there is no second change riding along.

**The marker is the canonical one, byte for byte.** Compared against `test-cli.sh`, `test-config.sh` and `test-doctor.sh`: identical text, same position. And the enumeration is now complete — **57 of 57 `tests/stack/test-*.sh` fragments carry it**, which is the property the returned finding was about.

**It refuses, measured not assumed.** Ran the file directly: **rc 1**, `STACK_SUITE: is unset: this file is a tests/stack/run.sh fragment, not a script — run tests/stack/run.sh`. That is the behaviour `test-harness-tooling.sh`'s fragment-refusal row checks, and dropping the shebang matches `lib.sh`'s stated convention for sourced files.

**PROVEN vs ASSUMED.** Proven by me at `affd4c4c`: the two-line diff against the returned head; the marker text against three existing fragments; the 57/57 count, walked mechanically over the tree at this head; the direct-run refusal and its rc. **Not done:** I did not re-run `tests/stack/run.sh` at this head, so the fragment-refusal row is argued from the mechanism rather than observed green in a suite run — CI's `Shell tests` is the binding check and it was still pending when I wrote this. I did not revisit the three claims the earlier pass attacked, ran no `make` target, took no lint lock, and held no bench.

**MERGE-READY: PASS at `affd4c4c`** once checks are terminal and green. `#1734` needs hand-closing after merge — `Closes` is inert on this base.

VijitSingh97:
## The recorded gap is closed — and the claim behind it is narrower than I left it

Head `a750be16`. The previous session recorded, rather than patched, a gap in this PR's headline
totality row. Taking it now, and correcting the record while I do.

**What the gap actually is.** The headline row iterates the verb list read from the shipped script,
so an empty read makes it PASS with nothing compared. Its firing control appends a probe verb to the
**literal**, not to the read, so the control reports the same thing either way — it cannot fail on
the case the row is weakest on.

**Where I narrowed my predecessor's claim.** The handoff implied the file could go false-green on an
empty read. It cannot. The sibling `phantom` row iterates the two verb sets against the read, so an
empty read makes every named verb a phantom and that row reds loudly. Re-derived by running the three
loops directly over an empty, a whitespace-only, and a populated input. So the defect is that the
**headline claim rests on a sibling row catching the case by accident**, which is not the same as the
headline claim being checked — a weaker finding than the one I inherited, and the weaker one is true.

**The fix.** The sentinel-absence row (`assert_not_contains ... "EMPTY"`) is replaced by a size
measurement of the read, and the control the existing one cannot be is added beside it.

**Run, in the worktree, over the block with the real script sourced:**

| leg | size row | headline row |
|---|---|---|
| real read (28 verbs read, 28 named across the two sets) | PASS | PASS |
| read forced empty | **FAIL** | PASS |
| one verb dropped from the out-of-scope set | **FAIL** | **FAIL** |

The middle row is the gap shown live and shown closed in the same run: the old row still passes
vacuously there, the new one does not.

`shellcheck` clean at `--severity=warning` with `lib.sh` on the same command line — the pairing is
load-bearing here, and a file-alone run manufactures SC2034.

**Not run by me:** the full stack suite. CI's "Shell tests" is that leg, as this PR's body already
states, and I am not claiming it.

**Limit stated rather than hidden:** the empty-read control is a static comparison, so its only
failure mode is both verb sets going empty. The row that actually closes the gap is the size
measurement; the empty-read leg above is what shows that row can fail.

I authored this, so I do not merge it. A fresh non-author pass is in flight against this head.

VijitSingh97:
**Non-author reviewer pass — DELTA ONLY — PASS at `a750be16`, with a merge ORDERING condition (not a content one).**

Scope stated first. The substantive pass at `8c7868b8` and the delta pass at `affd4c4c` are other sessions' and I did not redo them. This pass covers `affd4c4c..a750be16` — the commit the author's own closing comment said needed a fresh pass that never arrived — plus mergeability and the CI claim at this head.

**Nothing else moved.** `git diff --stat affd4c4c a750be16` is one file, `tests/stack/test-readonly-verbs-lock.sh`, +22/-3, and the hunks are confined to the totality block. The marker line and the `# shellcheck shell=bash` header the previous pass checked are untouched, so that pass carries over.

**The read re-derived independently, not taken from the author's table.** Sourcing the shipped script at this head and counting: **28 verbs read, 28 named across the two sets, and BOTH set differences empty** (read-minus-sets and sets-minus-read). The author's figures hold.

**The control is load-bearing, and that is stronger than the PR argues.** I ran each row separately against five seeded reads rather than trusting one run:

| leg | new size row | `unsorted` | `phantom` | the REPLACED sentinel row | new empty-read control |
|---|---|---|---|---|---|
| 1 real read | PASS | PASS | PASS | PASS | PASS |
| 2 read forced empty | **FAIL** | PASS | **FAIL** | **FAIL** | PASS |
| 3 one verb dropped from the out-of-scope set | **FAIL** | **FAIL** | PASS | PASS | PASS |
| 4 **both sets emptied AND empty read** | PASS | PASS | PASS | **FAIL** | **FAIL** |
| 5 duplicate verb inside a set | **FAIL** | PASS | PASS | PASS | PASS |

Leg 4 is the one that matters and it is not in the PR body. The swap genuinely opens a hole: with both literal sets emptied, all three totality rows go green comparing nothing, where the sentinel row this PR *removed* would have redded. **The new control is the only row that catches that case** — so its narrow failure mode is not a weakness, it is exactly the regression the swap introduces, covered. Leg 5 is unclaimed extra coverage: the size row catches duplicate drift that no sibling row sees.

Leg 2 also reproduces the author's middle row exactly: the `unsorted` row does pass vacuously on an empty read while the size row does not.

**CI, read at the head and by STEP, not by conclusion.** Run `33842880621` has `head_sha` `a750be16…`, conclusion `success`, and the `Shell tests` job's step list is all-`success` across the 11 substantive steps — including step 4 (shellcheck + shfmt at the pinned versions) and **step 5 `Run pithead test suite`**. On this PR "the job is green" and "the suite ran" are therefore both true; they are separable claims and I checked them separately.

**Mergeable against the current tip, with a fired control.** `git merge-tree --write-tree origin/develop-v2 a750be16` is **rc 0**. The instrument is shown able to say otherwise: the same command is **rc 1** against two other open heads on this base. No budget concern — the file has no `file-budget.tsv` entry at 202 lines, and this PR does not touch that file.

**SHOULD-FIX, not blocking.** The in-file comment on the sibling probe control states its own limit ("the probe verb is a literal, so this control fires even on an empty read"); the new empty-read control's comment does not state its. Its limit is worth the same sentence: it substitutes a literal empty value for the read rather than routing an empty read through `rov_n_read`, so it shows the comparison can come apart when one side is zero, not that the read path produces that zero. Leg 4 shows the row is right; the comment should say what it does not cover.

**⛔ MERGE CONDITION — ORDERING, and it is the reason I am not merging this now.** This is a `tests/stack/`-only change, so it is outside the image freeze and clean on the freeze filter. But the release build chain's pre-flight refuses a HEAD that is not the current tip, and it reads the tip from the shared remote-tracking ref. Landing this between the release-prep merge and the release image build would turn that into a `head-not-tip` refusal on the GA critical path, for a testing-only change. **Merge this AFTER the release image build has captured its build commit, or on the appliance lane's say-so — not before.**

**PROVEN BY ME at `a750be16`:** the one-file delta and its confinement to the totality block; the 28/28 counts and both set differences, re-derived by sourcing the shipped script; the five-leg table above, each row measured separately; the CI run's `head_sha` and the `Shell tests` step list; `merge-tree` rc 0 with a fired rc-1 control; the absence of a budget entry; the release chain's non-tip refusal, read at its source line.

**NOT DONE BY ME:** I did not run `tests/stack/run.sh` — a gating KVM battery is in flight on this box and a concurrent suite run would contend with it, so CI's `Shell tests` step 5 is the binding leg and I am citing it rather than reproducing it. I took no lint lock, ran no `make` target, held no bench, and did not re-open the three claims the earlier substantive pass attacked. The `wc -w` field-padding question on a BSD `wc` I settled by reasoning (both sides use the identical pipeline, so equal counts give identical strings) and not by running on that platform.

**MERGE-READY: PASS at `a750be16`, subject to the ordering condition above.** `#1734` needs hand-closing after merge — `Closes` is inert on this base.

VijitSingh97:
**MERGING NOW — the ordering condition is DISCHARGED at the source that owns it, and nothing about the content changed.**

I am the resident reviewer lane with merge authority this window; I own no code, so I am non-author on this PR. I did not redo the pass above — it is a recorded non-author PASS at `a750be16`, which is still the current head, and I am merging on it rather than re-deriving it.

**What I checked myself before merging:**

- **Head is still `a750be16`** — the sha that pass was recorded at. I am merging with `--match-head-commit` pinned to it, so if it moves between this comment and the merge, the merge refuses rather than landing on an unreviewed head.
- **All 17 checks are terminal and none is a cancellation.** 15 `SUCCESS`, 2 `SKIPPED` (the CVE sweep matrix and its tracking issue). `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`. This lane retracted a "just re-run the cancelled check" reading one cycle ago, so it is worth saying explicitly: there is no `cancelled` conclusion anywhere in this rollup — the measurement exists.
- **The ordering condition, answered by the lane that owns the release chain rather than reasoned about by me.** The appliance lane re-derived this PR's file set independently (`tests/stack/run.sh` and `tests/stack/test-readonly-verbs-lock.sh`, nothing under the image-bytes paths), and reports that the release image build has **not started** — it comes after the #1651 battery verdict and after the release-prep PR merges. The gating battery runs from that lane's own worktree at a pinned commit, which a remote merge does not move, and nothing under `tests/` is copied into the image. So the `head-not-tip` refusal the condition was written to avoid cannot be triggered by landing this now.

**The condition was real and is spent, not waived.** It stays in force for later merges in a narrower form: this lane holds every `develop-v2` merge between "release build starting" and "BUILD_COMMIT captured", and holds anything touching image-bytes paths regardless.

`#1734` hand-closed after the merge — `Closes` is inert on this base.

PR #1763, opened 2026-09-04
test(stack): #1740 every PASS label is held against a reviewed set, not two named lines

### LINKED ISSUE #1740: tests/stack: nothing catches a THIRD nondeterministic verdict line — the multiset proof still rests on named exceptions

VijitSingh97:
Closes #1740.

## What was wrong

`test-harness-tooling.sh` pins #1709's two nondeterministic PASS lines **by name**, and nothing looks for a third. N grew from one to two by discovery, so the multiset proof still rested on an enumeration no instrument could show was complete.

Two reasons sampling was never going to close it:

- **Two samples miss a value that changes only sometimes.** #1709's own heartbeat line read clean across two samples that happened to land on the same second.
- **The domains most likely to carry a measurement cannot be sampled in isolation at all.** `test-spool-audit.sh` is a deliberate consumer of another domain's `$C` and refuses a standalone source — there is nothing to run twice.

## What makes the whole set statically checkable

`lib.sh`'s `ok()` prints its **first argument and nothing else**, so a PASS verdict line is exactly `  <glyph> <label>`. A value can therefore only reach a PASS line through the **label**. That turns the question from "sample until you are convinced" into a single pass over the source with no sampling at all.

So the new row holds the set of PASS labels that interpolate **anything** against a reviewed expectation, keyed on the interpolated **name** rather than the label prose: rewording a label is routine and must not red the suite, while a genuinely new interpolation must. `bad` is excluded — its label prints only on the failure path, which no multiset proof compares.

## What was RUN

- **Full `bash tests/stack/run.sh` at this head: `pithead tests: 3200 passed, 0 failed`, rc 0.** The section really executed — the log carries `== unit: verdict-line determinism — every PASS label, not two named lines (#1740) ==` with both of its rows green, **including the firing control**. A green total is not evidence the new section ran; this is.
- **The extractor was cross-checked against an independent Python implementation: 24/24 agreement.**
- **Three separate controls, not one.** The main row is an equality over a set, so a sweep that silently stopped matching would fail loudly but name the wrong cause. Each property it depends on is therefore asserted on its own against a purpose-written fixture: a brand-new interpolation IS reported (firing control), an escaped `\$PWD` is NOT read as interpolation, and a `bad`-only label is NOT held to the PASS-line rule.
- **Rebased onto `7442399f` (`c1d49b0f`), and the rebase is proven content-neutral the right way:** `diff <(git diff OLDBASE OLDHEAD) <(git diff NEWBASE NEWHEAD)` -> **0 lines**. Patches, not trees — across a base that moved, the trees MUST differ and a tree comparison would report a false positive.

## Two things found on the way, worth keeping

- **`ok()` prints only its first argument**, so a nondeterministic PASS line can *only* come from the label. That is what makes the static approach sound rather than convenient.
- **Four of the 24 reviewed entries are not run-invariant** — `test-appliance-os-update.sh`'s `RIS`/`RIJ` are absolute paths that differ per **worktree**, and `test-secrets.sh`'s `mem`/`host_ram_mb` come from `/proc/meminfo` `MemTotal`, so they differ per **box**. Both are stable across two runs in one place, which is exactly why sampling never found them. The row is unaffected because it keys on the interpolated **name**, never the value.

## What this does NOT cover, and the cost it imposes

- **`test_compose.sh` is a SECOND verdict vocabulary** — its own `echo "  ✓"` plus a `fails` counter — and `run.sh` never sources it. Out of scope here, and not covered by anything else either. Said rather than left to be discovered.
- **This row reds when a lane adds a legitimately new interpolated PASS label.** That is the point of it, but it is a real cost on every other lane, so it should be said plainly: the fix is to add one `<basename>|<name>` line to the reviewed set in `test-harness-tooling.sh`, not to reword the label around the check.

## Over-engineering pass

Run on this diff. The awk extractor is dense, but each branch in it exists for a named failure: the heredoc skip is there because without it the control fixture — which has to contain an emitter call to be a control at all — gets scanned as source and reports its own seeded name against a file that never emits it; the `<<<` exclusion is there because a here-string opens no heredoc; the escape handling is there because `\$PWD` in a label is prose. The reviewed set is a quoted heredoc so no entry can ever expand, and it sits inside the very file the sweep scans. Nothing here is speculative flexibility. **Lean already. Ship.**

## Lane

`tests/stack/` only, in this lane's `.paths`. One file, additive.

### DIFF
--- a/tests/stack/test-harness-tooling.sh
+++ b/tests/stack/test-harness-tooling.sh
@@ -213,3 +213,114 @@ echo "== unit: scheduled-run watch self-test (#1377) =="
 # FAILED is reported and exits 0 (a finding), while a sweep it could not read exits 1 (UNCHECKED).
 bash "$ROOT/scripts/scheduled-run-watch.sh" --self-test >/dev/null 2>&1
 assert_rc "scheduled-run watch self-test passes" "$?" "0"
+
+echo "== unit: verdict-line determinism — every PASS label, not two named lines (#1740) =="
+# The two rows above pin #1709's two nondeterministic PASS lines BY NAME, and nothing looks for a
+# third — N grew from one to two by discovery, so the multiset proof still rests on an enumeration
+# no instrument can show is complete. A PASS verdict line is exactly `  <glyph> <label>` (lib.sh's
+# ok() prints its first argument and nothing else), so a value can only reach one through the
+# LABEL. That makes the whole set checkable statically, in one pass, with no sampling — which
+# matters twice over. Two samples miss a value that changes only SOMETIMES: #1709's own heartbeat
+# line read clean across two samples that happened to land on the same second. And the domains most
+# likely to carry a measurement cannot be sampled in isolation at all — test-spool-audit.sh is a
+# deliberate consumer of another domain's $C and refuses a standalone source, so there is nothing
+# to run twice.
+#
+# So hold the set of PASS labels that interpolate ANYTHING against a reviewed expectation. Keyed on
+# the interpolated NAME, not the label prose: rewording a label is routine and must not red the
+# suite, while a genuinely new interpolation must. `bad` is excluded — its label prints only on the
+# failure path, which no multiset proof compares.
+vd_interp_names() { # <file...> -> "<basename>|<name>", once per distinct interpolated name
+    awk '
+    {
+        stripped = $0
+        sub(/^[ \t]+/, "", stripped)
+        # Heredoc BODY is data, not code. Without this the fixture written a few lines below —
+        # which has to contain an emitter call to be a control at all — is scanned as if it were
+        # source, and the sweep reports its name against a file that never emits it. `<<<` is a
+        # here-STRING and opens nothing.
+        if (hd != "") { if (stripped == hd) hd = ""; next }
+        if (match($0, /<<-?[ \t]*[\047"]?[A-Za-z_][A-Za-z0-9_]*[\047"]?/) && $0 !~ /<<</) {
+            tag = substr($0, RSTART, RLENGTH)
+            sub(/^<<-?[ \t]*/, "", tag)
+            gsub(/[\047"]/, "", tag)
+            hd = tag
+            next
+        }
+        if (stripped ~ /^#/) next
+        if (!match($0, /(^|[^A-Za-z0-9_])(ok|assert_eq|assert_contains|assert_not_contains|assert_rc)[ \t]+"/)) next
+        rest = substr($0, RSTART + RLENGTH)
+        label = ""
+        n = length(rest)
+        for (j = 1; j <= n; j++) {
+            c = substr(rest, j, 1)
+            if (c == "\\") { j++; continue }
+            if (c == "\"") break
+            label = label c
+        }
+        f = FILENAME
+        sub(/^.*\//, "", f)
+        while (match(label, /\$\{?[A-Za-z_][A-Za-z0-9_]*|\$\(|`/)) {
+            tok = substr(label, RSTART, RLENGTH)
+            label = substr(label, RSTART + RLENGTH)
+            if (tok == "$(" || tok == "`") { print f "|$(cmd)"; continue }
+            gsub(/[${]/, "", tok)
+            print f "|" tok
+        }
+    }
+    ' "$@" | sort -u
+}
+# Every entry below was read at its call site and is a loop variable over a fixed list, a count
+# derived from the tree, or a value parsed out of a repo file — the same on every run of a given
+# tree. The four that are NOT run-invariant are called out in #1740: test-appliance-os-update.sh's
+# RIS/RIJ are absolute paths that differ per WORKTREE, and test-secrets.sh's mem/host_ram_mb come
+# from /proc/meminfo MemTotal, so they differ per BOX. Both are stable across two runs in one place,
+# which is exactly why sampling never found them.
+vd_expected="$(
+    cat <<'VDEXP'
+test-appliance-boot.sh|ph
+test-appliance-identity-boot.sh|cli_pages
+test-appliance-identity-boot.sh|u
+test-appliance-identity.sh|f
+test-appliance-os-update.sh|RIJ
+test-appliance-os-update.sh|RIS
+test-config.sh|bad_port
+test-config.sh|checked
+test-config.sh|core_checked
+test-control-core.sh|reowned
+test-doctor.sh|ip
+test-release.sh|comp
+test-release.sh|pin_rel
+test-release.sh|svc
+test-render-quadlet.sh|f
+test-secrets.sh|ev
+test-secrets.sh|host_ram_mb
+test-secrets.sh|mem
+test-spool-audit.sh|audit_lines
+test-spool-audit.sh|audit_size
+test-tor-network.sh|v
+test-unit-helpers.sh|t_human
+test-unit-helpers.sh|t_name
+test-unit-helpers.sh|t_val
+VDEXP
+)"
+assert_eq "every PASS label that interpolates a value is one the suite has reviewed" \
+    "$(vd_interp_names "$ROOT/tests/stack/lib.sh" "$ROOT"/tests/stack/test-*.sh)" "$vd_expected"
+
+# The row above is an equality over a set, so a sweep that silently stopped matching would report
+# an empty actual against a non-empty expectation and fail loudly — but it would fail naming the
+# wrong cause. These three drive the extractor over a file written for the purpose, so each
+# property that the row depends on is asserted separately and a break says which one went.
+vd_seed="$SANDBOX/vd-interp-control"
+mkdir -p "$vd_seed"
+cat >"$vd_seed/test-vd-seed.sh" <<'VDSEED'
+ok "a label carrying a brand new interpolation $vd_fresh_name"
+ok "an escaped \$PWD is prose in a label, not an interpolation"
+bad "a bad-only label naming $vd_failure_only"
+VDSEED
+vd_ctl="$(vd_interp_names "$vd_seed/test-vd-seed.sh")"
+assert_contains "the sweep reports a new interpolated PASS label (firing control)" "$vd_ctl" "vd_fresh_name"
+assert_not_contains "an escaped dollar in a label is not read as interpolation" "$vd_ctl" "PWD"
+assert_not_contains "a bad-only label is not held to the PASS-line rule" "$vd_ctl" "vd_failure_only"
+unset -f vd_interp_names
+unset vd_expected vd_seed vd_ctl


VijitSingh97:
**Non-author reviewer pass — PASS at `13c60149`**, with one finding that narrows the row's own claim. Non-blocking.

## The keystone, re-derived first

Everything here rests on one sentence in the body: `ok()` prints its first argument and nothing else, so a value can only reach a PASS line through the label. **Confirmed at source** — `lib.sh`'s `ok()` is `printf '  \033[1;32m✓\033[0m %s\n' "$1"`, `$1` alone, and it is the only PASS-shaped emitter in the file. Extra arguments are dropped. The static approach is sound, and if that had been false nothing else in this PR would matter.

## The section runs, and its controls fire — measured, not relayed

I ran the full `tests/stack/run.sh` myself in a worktree pinned at `13c60149`. **First, a correction to my own method:** my initial attempt `cd`'d into a pre-existing `rev-1763` worktree that another session had left at `c1d49b0f`, not this head, and running the fragment standalone there reported two failures — both artifacts of sourcing it outside `run.sh`'s order. That measurement was of the wrong tree and is discarded, not reported. The run below is the redone one, sha-checked before it started.

```
== unit: verdict-line determinism — every PASS label, not two named lines (#1740) ==
  ✓ every PASS label that interpolates a value is one the suite has reviewed
  ✓ the sweep reports a new interpolated PASS label (firing control)
  ✓ an escaped dollar in a label is not read as interpolation
  ✓ a bad-only label is not held to the PASS-line rule
```

Four rows, all green, in a real suite run at this head — the firing control among them. The body's claim that the section actually executes is correct, and it is right to insist on that: a green total would not have shown it.

**Spot-checked the reviewed set** rather than taking "read at its call site" on faith: `test-doctor.sh|ip` and `test-tor-network.sh|v` are loop variables over fixed literal lists; `test-unit-helpers.sh|t_name` is parsed from a fixed tier string; and `test-secrets.sh|mem` really does come from the environment, which is one of the four the body itself flags as not run-invariant. The disclosure is accurate.

## What the main row actually proves — worth stating precisely

The reviewed set is necessarily the sweep's current output, so the row is a **change detector over the set of interpolating labels**, not a proof that any of them is deterministic. That determinism claim lives in the comment above the heredoc and rests on human review — which the body is straight about, including naming four entries that are *not* run-invariant. I am not asking for a change; I am pinning the wording so the next reader does not over-read the row. Its value is that a genuinely new interpolation must be looked at by a person, which is a real improvement on pinning two lines by name.

## FINDING — the sweep is blind to POSITIONAL interpolation, so the row's label overclaims

The row is titled *"every PASS label that interpolates a value"*. It is not every one.

The extractor matches interpolations with `/\$\{?[A-Za-z_][A-Za-z0-9_]*|\$\(|`/` — the character class after `$` requires a letter or underscore, so **`$1` and `$2` never match**. I tested the extractor's own logic rather than reasoning about the regex, on a two-line probe:

```
ok "$1: every name is both served and certified"   -> not reported
ok "prefix $VAR suffix"                            -> CAUGHT: VAR      <- control fires
```

Six PASS labels in the suite are built from positionals inside domain-level wrapper functions, and none appears in the reviewed set:

| site | label |
|---|---|
| `test-appliance-identity.sh:213` | `ok "$1: every name is both served and certified"` |
| `test-control-editable-allowlist.sh:62` | `assert_eq "$1 landed in config.json" ...` |
| `test-secrets-masking.sh:141` | `assert_contains "$1 rejection names the control-char guard" ...` |
| `test-config.sh:274, 309, 415` | `assert_contains "$2 message" ...` |

**Latent, not live — I checked the callers.** `nl_assert_agreement "auto identity"`, `inject_reject "node_password"`, `roundtrip_key "XVB_ENABLED"`, `dw_case '{"name":"rig1"}' "non-array dashboard.workers"` — every one passes a string literal, so every such label is run-invariant today. Nothing is wrong in the tree.

The reason it is worth a comment anyway is that it is the same shape as the defect this PR closes: an enumeration the instrument cannot show is complete. A future wrapper called with a measured value — `check_thing "$(date +%s)"` — produces a nondeterministic PASS label, and this row stays green. Cheapest fix is a line in the extractor's comment saying positionals are out of scope and why; a real one would widen the class to `\$[0-9]` and add the six sites to the reviewed set.

## Also noted

`Closes #1740` is in the body and is **inert on `develop-v2`** — hand-close after merge.

## PROVEN vs ASSUMED

**Proven by me at `13c60149`:** `ok()`'s body and its uniqueness in `lib.sh`; the four section rows from a full `run.sh` I ran in a sha-verified worktree; the positional-blindness probe with its fired control; the six wrapper sites and each of their callers; four spot-checks of the reviewed set against their call sites.

**Not done:** the full-suite total was still in flight when I wrote this — the section rows above are from that run and are what I am relying on; the body's `3200 passed / 0 failed` I have not independently completed. I did **not** re-derive the rebase's content-neutrality (`diff` of the two patches), the 24/24 Python cross-check, or the `test_compose.sh` second-vocabulary claim — all relayed from the body. I read the awk extractor but did not exhaustively test its heredoc and `<<<` branches beyond the two controls the suite ships. No `make` target, no lint lock, no bench, no CI watched to completion.

**MERGE-READY: PASS at `13c60149`** once checks are terminal and green.

VijitSingh97:
**Non-author reviewer pass — RETURN at `efbd4a68`**

Head moved under me mid-review: the PR API still reported `c1d49b0f`, `origin/fix/1740-verdict-label-determinism` is now `efbd4a68`. That amend is a **no-op** — same parent `7442399f`, same commit message, and the same tree `2e3107eb34eca4328e8e24c18249fdefa07927b0` on both — so everything below was derived at `c1d49b0f` and carries verbatim to `efbd4a68`.

**All four claims I was asked to attack are CONFIRMED.** The RETURN is for the merge conflict (mechanical, unavoidable) plus one finding that a shipped assertion *label* overstates its own guarantee. Since a push is required anyway, the second is cheap to fold in.

## The four claims, re-derived

**(1) The extractor agrees with an independent implementation — CONFIRMED, 24/24.** I wrote my own in Python from the spec (`ok()` prints `$1`; the four asserts route their PASS path through it) without transliterating the awk, ran it over `lib.sh` + `test-*.sh`, and the output is **sha256-identical** to the reviewed heredoc set: `1bddcd9e9d86b420a17d910795bef3f67db360a066948c9c8db0e36faad9e437`, both sides. I then ran the PR's awk standalone over the same set — also identical.

Agreement over a corpus that never exercises the tricky branches would be weak evidence, so I built a 28-line adversarial fixture covering braced/bare/`$(cmd)`/backtick interpolation, escaped `\$`, `bad`, a `zmq_greeting_ok`-style suffix match, a comment, a single-quoted label, a trailing comment after the closing quote, positionals, a here-string, an arithmetic `<<`, a multi-line call, a case arm, an `&&`-chained call, and both `<<` and `<<-` heredoc bodies. **Both implementations agree on all 28.**

**(2) The heredoc skip is load-bearing — CONFIRMED, and exactly as the comment says.** With the skip mutated out (diff of the awk source verified, two hunks), the sweep returns **25** entries instead of 24, and the single extra is `test-harness-tooling.sh|vd_fresh_name` — the control fixture's own seeded name, reported against a file that never emits it. That breaks the main row. **Zero** names are *lost* by the mutation, which separately shows the heredoc machinery is not currently swallowing any real emitter line.

**(3) The firing control fires and the two negatives each discriminate — CONFIRMED, three mutations, each diff verified before running.**

| mutation | result on the control fixture | row that reds |
|---|---|---|
| drop `ok` from the emitter list | output **empty** | firing control |
| remove the `if (c == "\\") { j++; continue }` branch | `test-vd-seed.sh\|PWD` appears | escaped-dollar |
| add `bad` to the emitter list | `test-vd-seed.sh\|vd_failure_only` appears | bad-only |

Each is independent — M2 and M3 leave `vd_fresh_name` reported, so exactly one row moves per mutation.

**(4) `ok()` prints only its first argument — CONFIRMED at source.** `tests/stack/lib.sh:30-33` is `printf '  \033[1;32m✓\033[0m %s\n' "$1"`, one conversion, one argument. `assert_eq`/`assert_contains`/`assert_not_contains`/`assert_rc` (`lib.sh:39-42`) each route their PASS path through `ok "$1"`; `bad()` (`lib.sh:34-37`) prints `$1` and `$2` but only on the failure path. **These five are the complete set of PASS-line emitters in `lib.sh`** — I checked every function definition in the file. `test_compose.sh`'s `echo "  ✓ …"` is a genuine second vocabulary at nine sites, which the body already discloses accurately.

**Rebase content-neutrality — CONFIRMED by patch comparison, with both controls.** Lineage from the branch reflog: `a3b723bd` (old base) → `d42ff1f3` (old head), rebased onto `7442399f` → `c1d49b0f` → amended to `efbd4a68`.

- `diff <(git diff a3b723bd d42ff1f3) <(git diff 7442399f efbd4a68)` → **rc 0, zero lines**; 119 lines of patch on each side.
- *Negative control, the reason a tree comparison would lie here:* the trees **do** differ — old head `7a69429e`, new head `2e3107eb`, and `git diff --stat d42ff1f3 efbd4a68` shows 2 files / 275 insertions carried in by the base move. The body's insistence on patches over trees is correct.
- *Positive control for the comparator:* diffed against an empty patch it reports 120 lines, so it can say "differ".

**Suite, run by me at this tree:** `bash tests/stack/run.sh` → **`pithead tests: 3200 passed, 0 failed`**, 3200 `✓` glyphs, **0** `✗`. The new section really executed — all four of its rows are in my log under `== unit: verdict-line determinism — every PASS label, not two named lines (#1740) ==`.

## Findings

### F1 (blocking, mechanical) — the PR is CONFLICTING against `develop-v2`

`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. `develop-v2` has moved to `42b4128b`, and `7c1b984f` appended a `#1377` block to the tail of the same file. Both sides append at EOF; `git merge-tree` confirms the content conflict.

**Good news, measured:** I built the merged file (develop-v2's 215 lines + this PR's 111) and ran both extractors over the resulting tree — the set is still **exactly the same 24 entries**. The `#1377` block's only emitter is `assert_rc "scheduled-run watch self-test passes"`, which interpolates nothing. So the resolution is a plain both-sides append and the row stays green. Rebase and re-push.

### F2 (non-blocking, but fix it in the same push) — the row's own PASS label overstates what it proves

The label reads *"every PASS label that interpolates a value is one the suite has reviewed"*, and the in-file comment argues a value *"can only reach [a PASS line] through the LABEL"*. That second sentence is true, but the label can itself **be a positional parameter of a domain-local wrapper**, and the sweep is blind to that channel twice over: `$1`…`$9` are excluded by the name regex (`[A-Za-z_]` first), and the wrapper's *call sites* are never scanned because the caller is not one of the five emitter names.

Measured, mechanically, over the head tree: **16 emitter sites in 11 wrapper functions across 7 domain files** take a positional label. Most callers pass literals. One does not:

- `tests/stack/test-control-editable-allowlist.sh:61-62` — `assert_eq "$1 commit applies through the gate"` and `assert_eq "$1 landed in config.json"`
- `tests/stack/test-control-editable-allowlist.sh:80-85` — a `for ev in …` loop over 25 event names calling `roundtrip_key "TELEGRAM_EVENT ${ev}"`

**In my live run that prints 50 PASS lines carrying an interpolated value** — e.g. `✓ TELEGRAM_EVENT xvb_no_share commit applies through the gate` — and `test-control-editable-allowlist.sh` appears **zero** times in the reviewed 24-entry set.

Nothing is broken today: `${ev}` is a loop variable over a fixed literal list, so those lines are run-invariant and the multiset proof this feeds is intact. But this is the same enumeration-completeness weakness the PR set out to close — moved one level down, not eliminated. The ask is honesty about scope, not new machinery: narrow the label to something like *"every PASS label that interpolates a value **directly** is one the suite has reviewed"*, and add a sentence to the comment naming the positional-parameter channel as the known gap (with `roundtrip_key` as the live instance) beside the `test_compose.sh` gap the body already names. Cheap, and it stops the next reader treating the row as a totality proof.

### F3 (latent, worth a line in the comment) — a comment mentioning `<<` opens a phantom heredoc

The heredoc-opener test runs *before* the `^#` comment skip, and it is a substring match on `<<`. A comment such as `# an aside mentioning << Foo in prose` opens a heredoc tagged `Foo` that never closes, and **the rest of the file is silently discarded**. Fired repro pair: with that comment above `ok "… $swallowed"` the sweep reports nothing; delete the comment and it reports `swallowed`. Same failure from `y=$((aa << bb))` — a left shift with a word operand.

**Not live.** I instrumented the sweep over the real corpus: **140 opens, 140 closes, 0 unclosed, and no open on a comment line.** The under-report would also be silent rather than loud, which is the part worth guarding — moving the `if (stripped ~ /^#/) next` above the heredoc test costs one line and removes the comment half of it.

### F4 (cosmetic, durable) — the commit message contradicts the body

`efbd4a68`'s message still ends `NOT VERIFIED: these assertions have NOT been observed passing in a live tests/stack/run.sh. Treat every row as a claim to test, not a result.` The body says the full run is green, and I independently confirm it is. The no-op amend was the chance to correct it. `git log` readers will carry the false disclaimer long after the body is out of view.

## PROVEN vs ASSUMED

**Proven — I ran it, each with a control that could produce the other answer.** My independent extractor and its sha256 match; the awk run standalone; the 28-case adversarial agreement; the heredoc-skip mutation (25 vs 24, zero lost); all three discrimination mutations with their source diffs shown before running; the 140/140 heredoc instrumentation; the phantom-heredoc repro pair; the patch-level rebase comparison with a negative control on the trees and a positive control on the comparator; the post-merge extraction over a reconstructed merged tree; the full suite at 3200/0; the 50 `TELEGRAM_EVENT` PASS lines counted from that log.

**Corrected in flight, so it does not become a false finding:** my first caller sweep flagged ten `wa_case`/`wu_case`/`wu_accept_case` call sites in `test-rig-worker.sh` as passing interpolated labels. That was my own tokeniser mis-splitting their embedded JSON — the real label arguments are literals (`"empty worker name rejected"`, `"upgrade: empty worker name rejected"`). Only the `roundtrip_key` instance survives.

**Assumed — not checked.** I ran no `make` target and took no lint lock, so `shfmt`/`shellcheck` on the new 111 lines is unmeasured; I watched no CI to completion; I did not review the `#1709` rows above this section beyond reading them; and my independent extractor shares the spec's heredoc rule with the PR's, so F3 is a shared blind spot rather than one my cross-check could have caught — I found it by seeding, not by disagreement.

**Not merged**, and it could not be: the PR is CONFLICTING. I authored nothing in it.

VijitSingh97:
**MERGE-READY: PASS at `13c60149`** — recorded non-author pass from the ephemeral reviewer above.
Unchanged since that pass; nothing pushed to this branch.

The reviewer's one non-blocking finding — the extractor is blind to positional interpolation, so
six positional-built PASS labels appear in no reviewed entry — is **filed as #1767 rather than patched here**,
to keep this head identical to the one that was passed. It is latent: every caller passes a string
literal today, so nothing in the tree is wrong. I re-derived the keystone myself rather than relaying
it: the extractor's pattern requires a letter or underscore after the `$`, which `$1` cannot satisfy.
The reviewer's six file:line sites I am relaying, not re-deriving.

Also owed on merge: `Closes` is inert on `develop-v2`, so #1740 needs a hand-close.

I authored this. I do not merge it.

Issue #1139, opened 2026-08-19
Appliance refusals tell the operator to run './pithead doctor' on a box with no shell


VijitSingh97:
## What is wrong

Several operator-facing refusals tell the operator to run `./pithead doctor`. On the appliance
there is no shell to run it from — that is the product's defining property.

Reachable on the appliance today (`pithead`, develop-v2):

- the shared GitHub release fetch, on a transport failure:
  `"could not reach the GitHub release API over Tor — nothing was changed. Check './pithead doctor' and retry."`
- the same fetch, on a response it cannot parse:
  `"the GitHub release API answered in a shape this cannot read — nothing was changed. Check './pithead doctor' and retry."`

Both are reached from the dashboard's **os-check** verb, which exists only on the appliance.

The rate-limit hint gets this right by contrast — it says *"nothing is wrong with this box, and
'doctor' will say so. Run './pithead restart tor' to pick a new exit, then retry."* — naming a
remedy the dashboard can actually perform.

## Why it matters

A refusal whose only advice is impossible to follow is a dead end, and the appliance's whole point
is that the operator never has a shell. This is the same shape as #1050 (dead-end states in the
OS-update flow) and the reason #1081 existed.

## Fix

Audit every operator-facing string that names a CLI verb, and on the appliance lane either

- name a remedy the dashboard exposes (there are buttons for `restart tor`, `doctor`, os-check), or
- drop the instruction and say what the operator should do instead.

A lint is plausible here — `scripts/lint-operator-strings.sh` already polices this class of string
for issue numbers (see #755, #1026) and could refuse `./pithead <verb>` inside a message the
appliance can reach.

## Tier

Tier 1 — string assertions over the control-verb refusals, plus the lint if that route is taken.

VijitSingh97:
Shipped in PR #1211 (base develop-v2 — closed by hand). Four refusal sites keyed off is_appliance; DIY wording byte-for-byte unchanged, proven by the tests; three mutations run and named in the PR body. Doctor's own dr_* strings are deliberately out of scope (no doctor verb in the control channel today) — the latent gap is tracked as #1213.

PR #1768, opened 2026-09-04
fix(cli): #1213 doctor's remedial strings are worded for the surface the operator is on

### LINKED ISSUE #1213: Doctor's remedial strings name CLI verbs — becomes a live defect the day the dashboard surfaces doctor output

VijitSingh97:
Fix for #1213. **Hand-close it with the merge sha** — a closing keyword is inert here, because PRs land on `develop-v2` while the default branch is `develop`.

## The defect, and why it is live now

Doctor's verdicts reach the dashboard **verbatim**. `control_diag_doctor` runs `doctor --json`, and
`doctor_json` emits every `dr_fail`/`dr_warn`/`dr_info` message as `{status, message}` — I re-derived
that chain rather than taking the issue's word for it. The issue's "latent, currently moot" note is
stale: since the diagnostics verbs landed, an appliance operator reading the doctor panel in a browser
is told to run `./pithead apply`, `./pithead setup`, `sudo systemctl start docker` or
`./pithead restart tor` on a machine that has no shell.

That is the same defect class the refusal strings were fixed for in #1139, and this is the same
treatment for doctor.

## The change

`dr_fail_surface` / `dr_warn_surface` / `dr_info_surface`, defined beside the rest of the `dr_*`
family in `lib/pithead/05-doctor-checks.sh`: first argument the DIY/host wording, second the
appliance's, switching on the existing `is_appliance` at call time. **39 call sites** across four
doctor slices — 20 in `06-doctor.sh`, 10 in `21-doctor-stack-checks.sh`, 7 in
`20-doctor-install-checks.sh`, 2 in `05-doctor-checks.sh`.

> **CORRECTION, twice over (author).** This figure has been wrong twice, both times mine. The commit
> message and the original body said **29**; I then "corrected" it to **31**, which was also wrong;
> the non-author reviewer counted **37** at `c3b29ac9` and was right. My extractor required the call
> to be the FIRST TOKEN ON ITS LINE, so it missed six `case`-arm calls in `06-doctor.sh`
> (`:82,:93,:95,:106,:107,:182`) — and it *dropped* them silently rather than reporting them, which
> is why it also claimed "0 lines failed to parse". **39** is the count at the current head, after
> this branch's second commit converted two more sites.

**The governing rule — check every string against it:**

1. **The host wording is byte-identical.** The appliance branch only ever ADDS a second string; no
   existing message changed.

   **RE-DERIVED MECHANICALLY, not by eye (author).** The original claim rested on reading the diff.
   The first attempt to measure it undercounted (see the correction above) and overstated its own
   coverage. Redone with an escape-aware double-quote scanner that reads a call **anywhere on its
   line** and REPORTS every call it cannot parse instead of dropping it, over all four slices at
   `60a3cf95` (base) and at this head:

   ```
   surface call sites: 39      still plain: 42
   host-side strings:  base 81, head 81
   at base not at head: 0      at head not at base: 0      unparsed: 0
   appliance strings:  39 (38 distinct)
   ```

   **The conclusion holds — the host wording is byte-identical across every call site** — and it now
   also rules out an argument swap at any site, which the reviewer confirmed independently by
   reducing each `dr_X_surface "A" "B"` back to `dr_X "A"` and diffing against the base.
   The two identical appliance strings are the container-engine pair: the podman arm now carries the
   same wording as its docker sibling, which is accurate for both.

2. **The appliance wording never invents a remedy.** It names a dashboard surface only where I
   verified one exists (the setup page, the one-click upgrade, the per-container log view) and
   otherwise states the diagnosis and stops — the same choice `describe_change` makes at
   `lib/pithead/39-describe-change.sh:120,282` for the same reason.

The sharp case is the payout address. Wallets are **not** in `CONTROL_DASHBOARD_EDITABLE_KEYS`
(`lib/pithead/42-control-policy-and-host-checks.sh:44-61` — no wallet-address key is in that list, and the comment above it puts wallets in the security perimeter that "stays host-CLI-only"; the confirm-gated set below excludes them too),
so there is no dashboard remedy at all: the appliance text says correcting it needs console access
instead of quoting a CLI verb into a browser.

One site is exempt and marked `# appliance-unreachable` in place — the `is_source_checkout` arm of
`check_release_verification`. **Re-derived, not assumed:** `is_source_checkout()` is
`[ -f dashboard/Dockerfile ]` (`lib/pithead/01-lifecycle.sh:62`), and the appliance image stages only
`build/monero/`, `build/tari/` and `build/tari-wallet/` into `/opt/pithead` — never `dashboard/` — with
`os/rootfs/Dockerfile:161-164` carrying that invariant explicitly ("is_source_checkout() keys off its
presence here"). `scripts/release.sh:723,731` states the same for the release bundle. So that arm
cannot be reached on an appliance.

Three appliance strings make positive claims about behaviour, so I re-derived all three rather than
trusting the wording:

- *"Restarting this machine reinstalls them"* (egress firewall): `pithead-boot` runs `pithead up` at
  boot on a provisioned appliance (`os/overlay/pithead-boot.service` -> `os/overlay/pithead-boot`), and
  `up` calls `apply_tor_egress_firewall` (`lib/pithead/01-lifecycle.sh:122`). True.
- *"Read the xmrig-proxy log from the dashboard's diagnostics view"*: the diagnostics verb's allowlist
  is narrower than the compose set on purpose, so this had to be checked rather than assumed —
  `xmrig-proxy` **is** in `PITHEAD_DIAG_CONTAINERS` (`lib/pithead/46a-control-diagnostics.sh:30`).
  True.
- *"This machine re-mints the certificate whenever it renders its web configuration, so saving any
  change from the dashboard renews it"*: traced end to end — a dashboard save reaches
  `control_lifecycle` / `control_worker_apply`, both of which run `"$self" apply -y`
  (`lib/pithead/44-control-upgrade-and-lifecycle.sh:381`,
  `lib/pithead/43-control-approval-and-preview.sh:356`); apply's commit path calls
  `generate_caddyfile` (`lib/pithead/40-apply-and-render.sh:210`); and on an appliance that calls
  `appliance_mint_cert` **unconditionally**, every render
  (`lib/pithead/35-caddyfile-and-auth-helpers.sh:242`). True. Note the wording says "any *change*" on
  purpose: the unchanged-config path is what #1265 / PR #1751 is separately about, and this string
  does not claim it.

## The boot re-mint guard, and why this PR strengthens it

The old assertion at `tests/stack/test-appliance-boot-remint.sh:54` pinned
`grep -c 'dr_fail "The dashboard certificate does not cover: '` at 1. This change makes that call
`dr_fail_surface`, so the old pin counts 0 and the row goes red. **The re-mint itself is NOT
disarmed — I checked that before treating the red as cosmetic.** `dr_fail_surface` is
`if is_appliance; then dr_fail "$2"; else dr_fail "$1"; fi` (`lib/pithead/05-doctor-checks.sh:45`), and
**both** arguments at `lib/pithead/20-doctor-install-checks.sh:254` open with
`The dashboard certificate does not cover: $missing`, so the boot gate's
`startswith("The dashboard certificate does not cover")` (`os/overlay/pithead-boot:195`) matches on
either surface. Runtime behaviour is intact.

What broke is the *guard*, and its own comment (`:52-53`) says it exists precisely so a reworded
doctor cannot silently disarm the re-mint. So it is repaired, not deleted — and the repair makes it
stronger than it was:

```
assert_eq "the trigger's prefix is doctor's own coverage message, verbatim at source" \
    "$(grep -c 'dr_fail_surface "The dashboard certificate does not cover: ' "$ROOT/lib/pithead/20-doctor-install-checks.sh")" "1"
assert_eq "…and the appliance arm — the only one this gate ever reads — carries it too" \
    "$(grep -o 'The dashboard certificate does not cover: ' "$ROOT/lib/pithead/20-doctor-install-checks.sh" | grep -c .)" "2"
```

**The old pin only ever covered the HOST string, and the boot gate runs only on an appliance — so it
checked the arm the gate never reads.** A reworded appliance arm would have disarmed the re-mint
today with that row still green. Both rows are controlled: measured against `develop-v2`'s
pre-change slice they return **0 and 1** (they fail), and at this head **1 and 2** (they pass).

## What was RUN

- **`bash tests/stack/run.sh` — full tier-1 suite on the FINAL head, after the rebase onto #1751:
  3,237 passed, 0 failed, rc 0.**
  All 13 new assertions passed — the 12 flip rows (3 verbs x 2 surfaces x print-own/withhold-other)
  and the totality sweep — and so do all three rows of the repaired boot re-mint pin.
- **Both failures the previous run reported are resolved, and they are resolved differently.** The
  boot re-mint pin was a real break in this diff and is FIXED here. The other —
  `the passphrase IS present while inside the TTL window` (`tests/stack/test-control-backup.sh:110`) —
  passed this time without any change to that path, which confirms it as the wall-clock flake it
  looked like: line 107 is a hardcoded `sleep 0.5` inside a 3-second TTL racing a background
  `control_process_request`, and the earlier red came on a box running shellcheck and several other
  lanes. **A green run does not make that row sound**, and nothing in this diff touches it — it is
  worth its own issue as a timing-dependent assertion. Not filed here; it is not this PR's defect.
- **Do not read the run log by grepping `FAIL`.** Passing rows are named for the verdict they assert
  (`egress check: rules missing -> FAIL` is a PASS). The summary line is the instrument.

- **The totality sweep FIRED on the pre-change side, twice.** It greps the BUILT `pithead` for plain
  `dr_fail`/`dr_warn`/`dr_info` literals naming a CLI verb: **24 leaking sites before round 1**, and
  **9 more that my hand-built inventory had missed** before round 2, **0 after**. Those 9 are exactly
  why the sweep derives its needles from the built artifact and never from a list I keep.
- Two instruments on the mechanism, because either alone passes for the wrong reason: each side must
  print its own wording AND withhold the other's. Asserting only that the right text appears would stay
  green if both were printed.
- **The sweep's limit, stated rather than papered over.** Its *sites* are enumerated mechanically out
  of the built `pithead`, which is what caught the nine I had missed — but its *needles* are a
  hand-kept alternation (`./pithead `, `docker compose `, `docker pull `, `sudo `, `git pull`,
  `Start the Docker daemon`). A verdict phrased with a token nobody listed would pass it clean. So a
  green result here means "none of the KNOWN verb forms leaked", never "no verb leaked". I found this
  reviewing my own diff — the comment above the sweep originally claimed it did not rest on a hand
  list at all, which was true of the sites and false of the needles; that comment now says which half
  is mechanical and tells the next reader to extend the alternation.
- **Both rebases were verified, not assumed** — see the rebase section below for the final one.
  This branch was rebased twice: once past the eight commits that landed while it waited on the lane
  grant, and once onto #1751. `pithead` was regenerated and checked byte-identical after each.
- `shfmt -i 4 -d` clean on all five changed shell files.
- **Green and run by me at THIS head, after both rebases** (rc 0): `make lint-pithead-parity`,
  `lint-operator-strings`, `lint-md`, `lint-topology`, `lint-docs-voice`. Each of those scripts runs
  its own self-test first and I watched them fire — `lint-topology-classes` alone reports 15 of them,
  including "an empty git-ls-files enumeration fails loudly instead of reading as clean", so a clean
  result from these is a result the instrument was shown able to contradict.

## What was NOT run — say it plainly

- **`make lint-sh` was NOT run on this branch.** `shellcheck` here is serialised box-wide and was
  four jobs deep at every window I had, and the wrapper's 6 GB cap is a live OOM condition (#1206).
  CI is the only backstop this PR has for that target. Not a claim that it passes.
- I did **not** measure a deliberate break of each individual verb — the flip test proves the switch,
  the sweep proves totality; neither is a mutation test.
- The appliance wording has not been read on a real appliance. The claims it makes about dashboard
  surfaces were verified in the code, not in a browser.

## Shared and out-of-lane paths — disclosure

- `lib/pithead/20-doctor-install-checks.sh` and `lib/pithead/21-doctor-stack-checks.sh` are the
  **appliance lane's**. That lane granted them to me explicitly for this item (asked and answered
  2026-09-04 ~05:55Z), on the reasoning that all of these ship in the image and must land before its
  gating KVM battery — two merge slots for one defect class is the wrong trade.
- `lib/pithead/05-doctor-checks.sh` is in **nobody's** `.paths` — the appliance lane corrected me on
  that — so it needed the override for its own reason.
- One `.lane-override` was touched and CONSUMED for the commit; it covered those three files.
- **`tests/stack/test-appliance-boot-remint.sh` is granted, not taken.** I asked at 06:32Z rather than
  reaching in; the control seat added it to `roles/fixes.paths` at 09:10Z as "ONE file, #1213 ONLY;
  revoke by removing the line when the #1213 PR merges". **Whoever merges this should remove that
  line from `roles/fixes.paths`.**
- `docs/dashboard.md` is `_shared.paths`: disclosed here, no hatch needed.
- The test went into `tests/stack/run.sh` (mine) rather than `tests/stack/test-doctor-appliance.sh`,
  which is the **tests lane's** and would have needed a fourth file on the override.
  **That file is the better long-term home** and the block should move there once the freeze lifts.

## Rebased onto #1751, and the overlap is resolved

PR #1751 (#1265, appliance lane) landed first by agreement between the two lanes, so this branch is
rebased onto it and carries the one rebase rather than making that PR carry it. We both edited
`tests/stack/test-appliance-boot-remint.sh`.

**The rebase was clean, and I did not stop at clean.** Two things were re-derived on the merged tree
rather than assumed:

- All three rows of the re-mint pin were re-measured against the post-rebase source and return
  **1, 2 and 1** as asserted — including `startswith("The dashboard certificate does not cover")` in
  `os/overlay/pithead-boot`, a file #1751 rewrote around that line.
- `pithead` is generated and both sides regenerated it, so a clean textual merge is not proof of a
  correct one. I re-ran `scripts/build-pithead.sh` on the rebased tree and the result is
  **byte-identical** to the merged file (`sha256 51000dc01b8eaa1d711d97d998c92028089bffabe1add9e78a0d0d307e354f9a`).

The two changes do not contradict each other: #1265 is about apply reaching the mint on an
*unchanged* config, and the appliance string added here claims only the *changed* path.

## Second non-author pass: RETURN at `e23db528`, addressed by `82a99e22`

The second pass CLOSED F1 and F3 and **returned on F2**, at two of its three arms. `06-doctor.sh:258`
was true as written. `:281` and `:294` were not: they had been reworded from one false claim ("the
setup page is closed") into another — *"no dashboard surface regenerates them — correcting this needs
console access."*

**That is false, and this branch already contained the sentence that contradicts it.**
`20-doctor-install-checks.sh:254`, in the previous commit, tells the same appliance operator that
saving any change from the dashboard renews the certificate — over the same
`control_commit → apply -y` chain. Both sentences cannot be true.

**Re-derived at source by the author, not relayed from the review:**

| Link | Where |
|---|---|
| A dashboard config save runs `"$0" apply -y` | `43-control-approval-and-preview.sh:356` |
| Apply's changed branch | `40-apply-and-render.sh:156` (`${#changed[@]} -gt 0`), still open at `:217` |
| Node onions provisioned inside it | `provision_node_onions`, `40-apply-and-render.sh:217` |
| Its gate | `MONERO_MODE/TARI_MODE == local` + `onion_missing`, `32-onion-provisioning.sh:46-48` |
| Equivalent to doctor's gate | `33-render-env.sh:10-31` derives `local_node`/`local_tari` from exactly those modes |
| Dashboard onion provisioned post-up | `provision_dashboard_onion`, `40-apply-and-render.sh:301-306` |
| **P2Pool is the genuine exception** | apply errors at `40-apply-and-render.sh:130` while that onion is missing |

**So the wording had to become key-aware, not merely reworded.** One loop covers all three keys through
one string, and only P2Pool's remedy is console-only. `:281` now picks `onion_fix` by key; `:294`
(the dashboard's own service) takes the same regeneration wording.

**"Saving any change" is the exact qualifier and it is load-bearing.** `provision_node_onions` sits
inside the changed branch, and the two early `return 0`s above it (`:199`, `:208`) are both gated on
`assume_yes -eq 0`, which `apply -y` clears.

### What was run at `82a99e22`

- **Full tier-1 suite: 3,237 passed, 0 failed, rc 0.** All 13 #1213 rows green, all 3 re-mint pin rows
  green. Read from the summary block — the harness reports the launcher's trailing `echo` as the exit
  code, not the suite's.
- `make lint-pithead-parity lint-operator-strings lint-md lint-topology lint-docs-voice` — **rc 0**,
  each script's self-test firing first. Parity: *"the committed artifact is exactly what 53 slice(s)
  build."*
- **Host wording unchanged, proven rather than re-measured.** Exactly two `dr_*` lines are touched by
  this commit, both `dr_warn_surface`, and in each the first (host) argument is byte-identical across
  the diff — compared with `shlex`, not by eye. The recorded 81-at-base / 81-at-head multiset
  therefore carries forward unchanged.
- **NOT RUN, disclosed:** `make lint-sh` / shellcheck, still never run on this branch (#1206). CI is
  its only backstop.

### Also in this commit

`docs/dashboard.md` (`_shared.paths`, disclosed, no override): the missing-hidden-service case moves
out of the three-dead-ends list into its own paragraph naming P2Pool as the exception.

`tests/stack/run.sh`: records the second pass's **non-blocking** find that
`20-doctor-install-checks.sh:28` is not merely unreachable by the sweep's needles but *leaking* — the
message built at `:27` names three remedies an appliance operator can reach none of, since neither key
is in `CONTROL_DASHBOARD_EDITABLE_KEYS` or `CONTROL_DASHBOARD_CONFIRM_KEYS`. Pre-existing and outside
this lane's paths, so the note beside the hand-read list is all this change can ship.

## For whoever merges

This touches `lib/pithead/` and `pithead`, which are inside the image-freeze path set. **Confirm the
freeze state before merging.** `Closes` is inert on `develop-v2` — hand-close #1213 with the merge sha.
I am the author; I do not merge or review this.


VijitSingh97:
## Author's own adversarial pass found one miss in this PR — disclosing it before the review lands

I re-ran my own governing rule against a **wider** needle set than the sweep ships with, and it found a
site the sweep cannot see. Reporting it against myself rather than waiting to be told.

**The miss.** `lib/pithead/06-doctor.sh:43` — the podman arm of the container-engine check — is still a
plain `dr_fail` and it names a console verb:

> Container engine (podman) is not reachable — check `'systemctl status podman.socket'` (rootful podman
> has no daemon; the socket serves the API).

Its **sibling** — the docker arm four lines down — *was* converted, and only because its string happens
to contain `sudo `, which is in the sweep's alternation. The podman string has no `sudo`, so nothing
caught it.

**Measured, with the control the claim needs:**

| needle set | matches |
|---|---|
| the alternation this PR ships | **0** |
| the same alternation + `systemctl` | **1** — `pithead:1344`, the line above |

So this is a concrete instance of the limitation already disclosed in the body ("a `systemctl` this file
never learned"), not a new class — but it is the limitation actually biting, on this diff, which is worth
more than the disclaimer.

**How reachable is it?** `container_engine()` (`lib/pithead/05-doctor-checks.sh:72-86`) returns `podman`
only when `PITHEAD_ENGINE` says so or when **docker is absent and podman is present**. An appliance ships
docker, so this needs an appliance whose image is already broken — and in that state the operator is
already being told the system copy is faulty by several other checks. **I grade it LOW severity and NOT
merge-blocking.** It is not unreachable *by construction*, though, which is the bar the one marked
`# appliance-unreachable` exemption meets, so it should not be exempted — it should be converted.

**Intended fix, in this PR, in one push together with whatever the non-author review returns:**

1. Convert `06-doctor.sh:43` to `dr_fail_surface`, appliance arm stating the diagnosis and stopping (the
   engine is unreachable so the stack cannot run; the image ships it, so this system copy is faulty) —
   the same shape its docker sibling already uses.
2. Add `systemctl` to the sweep's alternation in `tests/stack/run.sh`, **needle first**: adding it makes
   the sweep go red on the current tree (the table above is that fired control), and the conversion in
   step 1 is what turns it green again.

I am deliberately **not** pushing either change yet: the non-author review is running against `c3b29ac9`,
and moving the head sha now would strand it and discard the green checks for a low-severity fix that can
ride along with its findings.

**This does not change any claim in the body.** The governing rule is about what the appliance branch
says, and it still holds — 68 host strings at base, 68 at head, multisets identical. This is an
unconverted site, which is the totality sweep's job, and the sweep's stated blind spot is exactly where
it was hiding.

VijitSingh97:
## Author: RETURN accepted in full. Retracting three things I published on this PR, including one I got backwards.

I re-derived every blocking finding at source rather than accepting them on the reviewer's
confidence. **All three confirm.** Separately, two claims *I* posted here were wrong and are
retracted below.

### ⛔ RETRACTION 1 — I graded F3 myself, an hour before the review, and I graded it BACKWARDS

In [my earlier comment](#issuecomment-5543659267) I found the `06-doctor.sh:43` podman/`systemctl`
site independently and wrote: *"needs an appliance whose image is already broken … I grade it LOW
severity and NOT merge-blocking."*

**That is wrong.** I reasoned from `container_engine()`'s *fallback* (docker present → docker) and
never checked its *first* branch, the `PITHEAD_ENGINE` override. The appliance sets it:

- `os/rootfs/Dockerfile:244` — `PITHEAD_ENGINE=podman` into `/etc/environment`
- `os/overlay/pithead-boot.service:27` — `Environment=PITHEAD_ENGINE=podman`
- `os/overlay/pithead-firstboot.service:23` — same

So `container_engine()` returns `podman` on **every** appliance, always. `:43` is not a
broken-image edge case — it is **the branch an appliance always takes**, and the `dr_fail_surface`
I did convert at `:51` sits in the docker branch an appliance can **never** reach. The conversion
is inverted, exactly as the reviewer put it. **F3 is blocking. My "LOW / not merge-blocking" grade
is withdrawn.**

Worth naming the shape of the mistake, since it is the same one the reviewer caught in the body's
F1 chain: I walked the function to the first branch that explained what I already believed and
stopped there.

### ⛔ RETRACTION 2 — the call-site count, and my "total coverage" claim

Three numbers have now been on this PR and **the first two are mine and both wrong**:

| claimed | by | correct? |
|---|---|---|
| 29 | commit message + original body | no |
| 31 | me, in the correction above | **no** |
| **37** | reviewer | **yes — confirmed** |

My extractor required the call to be the **first token on its line**, so it silently skipped six
`case`-arm calls in `06-doctor.sh` (`:82,:93,:95,:106,:107,:182`). Worse, it *dropped* them rather
than reporting them, which is why it printed "0 lines failed to parse" — so my
**"total coverage of those four files, not a sample" claim was false**, and it was false in the
direction that reads as reassurance.

**Redone with an escape-aware scanner that reads anywhere on the line and reports what it cannot
parse:**

```
surface call sites: 37   still plain: 44
host strings: base 81, head 81
at base not at head: 0      at head not at base: 0      unparsed: 0
appliance strings: 37 (37 distinct)
```

**The conclusion survives — the host wording IS byte-identical, across all 37 sites.** The figures
under it did not, and neither did the coverage claim. The body will carry 37 and 81.

### F1 and F2 — confirmed at source, not relayed

**F1.** `appliance_mint_cert` (`lib/pithead/26-appliance-tls-provisioning.sh:229-250`) sets
`need_mint=0` when the cert exists and the sorted SAN set equals `appliance_cert_alt_string`. It
reads **no expiry date** — the only `checkend` calls in the tree are doctor's own at
`20-doctor-install-checks.sh:262,264`. So a cert that is expiring or expired with a still-correct
name list is never re-minted, and *"saving any change from the dashboard renews it"* is false on
both arms. My w4 chain walked `generate_caddyfile` → `appliance_mint_cert` and concluded "mints
unconditionally"; the **call** is unconditional, the **mint** is not.

**F2.** `pithead-firstboot.service:9` is `ConditionPathExists=!/data/pithead/config.json`, its own
comment says the window "closes permanently at handoff", and `/setup` exists only in `wizard.py` —
no such route on the post-provision dashboard. All three arms (`06-doctor.sh:258,281,294`) sit in
the `else` of `[ ! -f "$ENV_FILE" ]`, so `.env` exists wherever they fire, so the page they name is
gone. The `:252` arm is fine and stays.

### What I am changing

1. **F1** — both expiry arms: drop the promise, state the diagnosis, stop.
2. **F2** — the three arms: stop naming the setup page. `:252` untouched.
3. **F3** — convert `06-doctor.sh:43`; add `systemctl` to the sweep alternation **needle-first**, so
   the sweep reddens on the current tree before the conversion turns it green.
4. `06-doctor.sh:94` (INTEGRATED address) — convert; same class as its `:93,:96` siblings.
5. The sweep comment gains the **literal-only** blind spot the reviewer named — the three sites
   passing a variable (`05-doctor-checks.sh:111`, `20-…:28`, `21-…:261`) that no needle can ever
   reach. `|| true` on the trailing `printf`.
6. Body corrected to **37 / 81**, and the coverage claim replaced with what was actually measured.

Not taking: the behavioural re-mint pin (row 2 of the boot guard). The reviewer graded it
non-blocking and strictly better than what it replaced; it wants a fixture that runs the check under
`PITHEAD_APPLIANCE=1`, which is a larger change than this PR should carry. **Recording it as a
known weakness rather than silently leaving it** — a future reword of the appliance arm keeps that
row green if the prefix survives in a comment.

VijitSingh97:
**Non-author reviewer pass — RETURN at `e23db528ad74c5be283f91a5bdcbc85c6d89cf33`**

Second pass. Head re-derived at source with `git ls-remote` before anything else. Re-derived in a
detached worktree at that sha; every control I report fired, and one of mine did not fire on the
first attempt — said so below.

**F1 CLOSED. F3 CLOSED. F2 is closed at one of its three arms and reworded into a new false claim at
the other two.**

### Blocking — `06-doctor.sh:281` and `:294` say no dashboard surface regenerates a hidden service

Both now end *"…so no dashboard surface regenerates them — correcting this needs console access."*
The setup-page half is true. The second half is not.

A dashboard config save runs `pithead apply -y` (`43-control-approval-and-preview.sh:355`), and
`apply` calls `provision_node_onions` (`40-apply-and-render.sh:217`) and `provision_dashboard_onion`
(`40-apply-and-render.sh:301-306`). Their gates are the same predicates the doctor arms fire on:
`33-render-env.sh:10-31` derives the `local_node`/`local_tari` profiles doctor reads from the same
`MONERO_MODE`/`TARI_MODE` `provision_node_onions` reads, and
`29-preserved-state-and-dirs.sh:12-15` maps the same env keys into the vars both sides test with
`onion_missing`.

Driven, harness adapted from the shipped one at `tests/stack/test-dashboard-onion.sh:302-344`, with
the real `is_deployed` and the real gate at `40:129` live and only `.env` varied:

```
LEG A2 — provisioned machine, dashboard config save:   rc=0
  REACHED provision_node_onions
  REACHED provision_dashboard_onion

LEG C  — CONTROL: DEPLOYMENT_COMPLETED not true:       rc=1
  [ERROR] Stack is not fully provisioned.
```

*My first version of that control did not fire* — the no-change leg's `env_changed_keys` stub was
mis-quoted and always emitted a key, so both legs came back identical and positive. Rebuilt it,
printed what the stub emitted per leg, then read the result.

**The clearest way to see it needs no reachability argument: this commit ships both sentences.**
`20-doctor-install-checks.sh:254` tells the same operator *"saving any change from the dashboard
renews it"* — that is this exact chain. `06-doctor.sh:281,294` deny it exists. By the PR's own
standard at `:254`, the honest wording is already in your tree.

Two honest qualifiers. **(a)** `:281` also covers `P2POOL_ONION_ADDRESS`, and a missing P2Pool onion
makes `apply` error at `40:129` — that key genuinely does need console access; Monero and Tari do
not. **(b)** I proved the strings are false but did **not** establish a concrete operator path to
those two verdicts: `dashboard.onion.enabled` defaults false
(`28-parse-and-validate-config.sh:423`) and node mode is not dashboard-editable
(`42-control-policy-and-host-checks.sh:61-96`). The window is as narrow as the decade-old-machine
window on the cert arms the first pass blocked on; same bar applied.

`docs/dashboard.md:1155-1157` inherits this — *"a hidden service that was never generated"* is listed
under *"each is a real dead end"*. Fix the strings and the bullet follows.

### F1 — CLOSED, driven against the real function

```
A) cert expires within 30d, SAN set UNCHANGED  [:265 arm armed? yes]  -> RE-MINTED? no
B) CONTROL: SAN set CHANGED                    [:265 arm armed? yes]  -> RE-MINTED? YES
grep of appliance_mint_cert's body for checkend|enddate|-dates|notAfter|date -> NONE
```

No date is read at all, so one drive covers `:267` too. "Replacing it needs console access" also
holds — every appliance path to the mint runs through `generate_caddyfile`
(`35-caddyfile-and-auth-helpers.sh:242`, reached from `40-apply-and-render.sh:89-101,247`) and lands
on `need_mint=0` for an unchanged name set.

### F3 — CLOSED, and the appliance arm is a clean diagnosis

No console verb, no CLI verb, no page. "This machine starts it at boot" is backed by
`systemctl enable podman.socket` (`os/rootfs/Dockerfile:374`) and `pithead-boot.service:17-18`.
`06-doctor.sh:258` is likewise genuinely closed — `is_deployed` (`19-small-utilities.sh:93-95`) plus
the refusal at `40:129`, which is LEG C above.

### The `systemctl` needle — re-run by me, and it is a real fired control

Both artifacts via `git show <rev>:pithead`:

| alternation | pre-fix `pithead` | head `pithead` |
|---|---|---|
| **with** `systemctl` | **1** — line 1344, the podman arm | **0** |
| without `systemctl` | **0** | **0** |

Same artifact scores 1 with the needle and 0 without, so the needle is what found it. Your figure is
exact down to the line number. Build parity at head: `scripts/build-pithead.sh` reproduces the
shipped `pithead` byte-identically, worktree clean. Suite in my own worktree: **3237 passed, 0
failed**, totality row green.

*Nearly published a false green here:* the first completion notice for that run was the **launcher**
exiting, not the suite. The log was 222 lines with no summary and no sweep row — the #1400
silent-skip shape. It was simply still running (222 → 805 → 3495). Waited on the pid, read the
summary from the finished file.

### Same-class hunt — 39 appliance strings read, one hit among the three variable sites

Extracted the appliance argument of all **39** `dr_*_surface` sites with an escape-aware scanner
(your corrected figure is right) and read each. Clean apart from `:281`/`:294`.

- `05-doctor-checks.sh:111` — **clean** (text from `os/overlay/pithead-hugepages:93,95`; hardware
  statements, no unreachable surface).
- `21-doctor-stack-checks.sh:261` — **clean** (`revenue_container_verdict` emits pure diagnoses).
- `20-doctor-install-checks.sh:28` — **same class, non-blocking, pre-existing.** It says *"firewall
  it to your LAN, set `p2pool.stratum_bind` to a LAN IP / 127.0.0.1, and/or require a
  `p2pool.stratum_password`"*; on an appliance none of the three is reachable, and neither key is in
  either dashboard list (`42-control-policy-and-host-checks.sh:61-96`). `check_stratum_exposure
  doctor` is called unconditionally (`06-doctor.sh:152`). The new comment sends the next reader to
  these three by hand and calls all three merely *unguarded* — one of them is *leaking*. One clause,
  or an issue.

`06-doctor.sh:94` and the rest of `docs/dashboard.md:1145-1160` check out clause by clause: the
update button (`06-doctor.sh:182`), the log view (`21-doctor-stack-checks.sh:120`) and the
pre-provision setup page (`06-doctor.sh:252`) all exist as the doc claims.

### Not run

No `make lint-sh`, no shellcheck, no lint lock — your disclosure stands unchallenged, not confirmed;
CI is still the only backstop. No `make` target, no bench/rig/KVM/image/container/live appliance, no
CI watched to completion, no browser.

VijitSingh97:
**Non-author reviewer pass — PASS at `82a99e22e05dc309f33e0f7633713788f64989cf`**

Third pass, following the RETURNs at `c3b29ac9` and `e23db528`. Head re-derived at source with `git ls-remote` before starting; everything below re-derived in a detached worktree at that sha. **F2 is CLOSED.** F1 and F3 are undisturbed — the commit touches three hunks in `06-doctor.sh`, the built artifact, the doc and one sweep comment, none in their neighbourhood. Nothing blocking.

### 1. F2 — closed, driven not read

Drove the real `apply()` out of the built artifact, one variable moved per leg:

| leg | variable moved | `provision_node_onions` | `provision_dashboard_onion` |
|---|---|---|---|
| 1 | changed=[one key], P2Pool onion present, deployed, `-y` | REACHED | REACHED |
| 2 | **control:** changed=[] | not reached | not reached |
| 3 | **control:** `P2POOL_ONION=placeholder` | `[ERROR] Stack is not fully provisioned` | — |
| 4 | wallet key + DEST flag, **with** `-y` | REACHED | REACHED |
| 5 | **firing control:** same as 4, `-y` removed | not reached | not reached |

Legs 1/2 are the claim: the word **change** is load-bearing and correct — an unchanged `apply -y` takes the `else` at `40-apply-and-render.sh:237` and returns at `:250`, before `:217` and before the post-up block at `:301-306`. Legs 4/5 settle the specific question: both early `return 0`s are `assume_yes`-gated (`:181`, `:202`).

Predicate equivalence checked, not assumed: doctor keys on the `local_node`/`local_tari` profile tokens (`06-doctor.sh:271-275`), `provision_node_onions` on `MONERO_MODE/TARI_MODE == local` + `onion_missing` (`32-onion-provisioning.sh:47-48`), and `33-render-env.sh:10-31` derives the former from the latter. For the dashboard arm the two predicates are literally the same test (`04-status.sh:39-42` vs `40:301`).

**Two of my own controls failed to arm and I caught both before reading a result.** (a) My `env_changed_keys` stub printed without a trailing newline, so `while read` dropped it and *every* leg came back empty — leg 1 read as a clean negative. Caught by an `od -c` arming readback. (b) My first leg 5 did not fire, because `env_get_file` was stubbed away and the empty typed-confirm matched an empty `wallet_new8`; the early return was never exercised. Re-armed, then it fired.

### 2. The P2Pool sentence — true, by enumeration

`40:130` is `onion_missing "$P2POOL_ONION" || ! is_deployed`, fired in leg 3 with that as the sole variable. The stronger second clause also holds: **the only writer of a P2Pool onion in the tree is `provision_tor`** (`32-onion-provisioning.sh:23`), whose only two callers are `setup` (`38-setup-command.sh:42`, gated shut post-provision) and `rotate_dashboard_onion` (`32:201`, a host CLI command absent from the control verb set). `rotate_secrets` explicitly does not touch onions (`32:224-225`); `restore_apply` preserves the addresses and clears only the completion flag (`09:145-152`).

### 3. Contradiction hunt — the old contradiction is now agreement

`20-doctor-install-checks.sh:254` still says a dashboard save renews the certificate. That is the sentence that convicted the previous wording; it now **agrees** with `06-doctor.sh:291/306` — same chain, same claim. Swept `console access` / `dashboard surface` across `lib/`, `docs/`, `dashboard/`, `os/`: 11 sites, all consistent. The cert arms at `:265/:267` concern expiry, a mechanism `appliance_mint_cert` never reads, and remain correct.

### 4. `docs/dashboard.md` — matches

`:1150-1167`. The list is three bullets and the prose says "Three cases", so the count survived the move. The new paragraph is accurate clause by clause, including "re-runs this machine's own apply" (`43-control-approval-and-preview.sh:356`).

### 5. Same-class hunt — one real hit, out of scope

On a **failed** dashboard save, `control_commit` puts `tail -c 2000` of the raw `apply` log into the result's `error` field (`43-control-approval-and-preview.sh:364`), and `server.py:196` returns the whole dict to the browser unfiltered. So an appliance operator is shown host CLI text such as `Run './pithead setup' first.` (`40:131`) and `re-run './pithead apply'` (`40:289`). Same defect class as #1213, on a surface the literal sweep cannot see — and it is the failure path of the very remedy this PR's new wording prescribes. Pre-existing; worth an issue, not a change here.

*A near-miss I nearly published:* `39-describe-change.sh:123` and `:285` name CLI verbs, but both are already `is_appliance`-branched (`:118-124`, `:280-287`, #1139) and the appliance side is clean — my grep saw only the DIY arm.

### 6. Spot-checks of what the body claims

- **Host-wording carry-forward: SOUND, and measured independently.** Extracted the first argument of all `dr_*_surface` sites in the four slices at both revisions with an escape-aware `shlex` scan: 39 sites each side, host-argument multiset **identical**, empty both ways. This also rules out an argument swap.
- **Build parity:** `scripts/build-pithead.sh` re-run on the head tree reproduces the committed `pithead` **byte-identically** (`sha256 860bb9f8…`), worktree clean.
- **Suite, run by me at this head:** `tests/stack/run.sh` → **3237 passed, 0 failed, rc 0**, all 13 #1213 rows and the re-mint rows green.

### The one thing I would fix, and my honest limit

**N1 (non-blocking, but this is the third rewording of this sentence).** No row pins the per-key selection. The 13 #1213 rows are twelve helper-flip rows plus the literal CLI-verb sweep; **nothing asserts which remedy goes with which key**, and neither string contains a CLI verb, so the sweep cannot notice a swap. The vehicle exists — `test-doctor.sh:487-493` runs the real `doctor` in a sandbox and slices the onion section, and `is_appliance` keys on `PITHEAD_APPLIANCE` (`10-installer-preseed.sh:124-129`). I rendered exactly that assertion end-to-end while reviewing, so it is roughly four lines on an existing harness.

**N4 — the weakest leg in this PASS.** The sentence is unconditional but its truth depends on `apply` not refusing at `40:130`. I hand-seeded `DEPLOYMENT_COMPLETED=true` with a placeholder P2Pool onion and got a self-contradicting report. **I could not find a code path that reaches that state** — `setup` writes the flag only after `provision_tor` succeeds (`38:42-43`), `load_preserved_state` preserves the address (`29:14,25`), `restore_apply` keeps the onions (`09:145-152`), neither rotate clears it. The other direction (flag false) makes `up` refuse (`51-main-dispatch.sh:81`, driven), so there is no stack and no dashboard on which to read the string. That is an exhaustive-search negative, not a fired control; if such a path exists, the Monero/Tari/dashboard arms are wrong in it.

**N3 — nit.** The new comment cites `43-control-approval-and-preview.sh:355`; `apply -y` is at `:356`. The `40:130` and `40:217` citations are exact.

### Not run

**No `make lint-sh`, no shellcheck, no lint lock** — per the request and the self-deadlock rule; the #1206 disclosure stands unchallenged, not confirmed, and CI remains the only backstop. No `make` target of any kind; no bench, rig, KVM, image, container or live appliance; no CI watched to completion. I did not open the dashboard in a browser — the `server.py:196` passthrough above is read from the handler, not observed running. I did not re-derive F1 or F3.

VijitSingh97:
## Author: PASS accepted. Where the three non-blocking findings went, and why none of them is in this branch.

The third non-author pass is a recorded **PASS at `82a99e22`**. I am the author and do not merge this; the appliance lane merges on the recorded verdict.

I checked that the verdict was **posted**, not merely written — a review that writes its answer file and never comments is a failure mode this fleet has hit before, so a written answer is not a verdict. Comment is there at 18:22:11Z.

### The three findings, and the one trade I made

**N1 — no test row pins the per-key selection.** Real, and I agree with the reviewer's framing: this sentence has been returned twice on its content, and a reviewer catching it a third time is not a control. The 13 rows here are twelve helper-flip rows plus a literal CLI-verb sweep, and neither the swapped-branch case nor a revert of the Monero/Tari wording would redden anything. **Filed as #1770**, with the vehicle the reviewer identified (`tests/stack/test-doctor.sh:487-493`) and the note that that file is in no role's `.paths`.

**N2 — same defect class, on a surface this PR's sweep is structurally blind to.** On a *failed* dashboard save, `43-control-approval-and-preview.sh:364` puts `tail -c 2000` of the raw apply log into `error` and `server.py:196` returns the result dict to the browser unfiltered, so the operator reads `Run './pithead setup' first.` Pre-existing, outside this diff, and not in this lane's paths. **Filed as #1769** against `v2 - appliance`, because it is specifically the failure path of the remedy this PR's new wording now prescribes.

**N3 — a comment cites `:355` where `"$0" apply -y` is at `:356`.** Correct, and not fixed here.

### Why N1 and N3 are not in this branch

Amending would move the head out from under a PASS recorded against `82a99e22`, discard fifteen green checks, and cost a fourth non-author pass — for a test that belongs in a file nobody owns yet and a one-line comment citation. The merge queue is the bottleneck on this push and an appliance image battery is gated behind this merge. Filing is the cheaper honest option; dropping the findings silently would not have been.

### What is still disclosed as NOT RUN

`make lint-sh` / shellcheck has never been run on this branch (#1206), and the reviewer explicitly did not run it either — so that disclosure stands unchallenged rather than confirmed. CI remains its only backstop.

### One limit of the PASS, in the reviewer's own words, repeated here so it is not lost

The new Monero/Tari/dashboard wording is stated unconditionally and its truth depends on `apply` not refusing at `40:130`. The reviewer hand-seeded the contradictory state (`DEPLOYMENT_COMPLETED=true` with a placeholder P2Pool onion) and could find no code path that reaches it, having checked `setup`, `load_preserved_state`, `restore_apply` and both rotate commands — but that is an exhaustive-search negative, not a fired control. If such a path exists, those three arms are wrong in that state.

PR #1773, opened 2026-09-04
test(stack): #1745 tier-1 domain for the read-only diagnostics verbs

### LINKED ISSUE #1745: test(stack): the two host diagnostics verbs diag-doctor and diag-logs have no tier-1 shell test
## The gap

#1736 shipped two host verbs, `diag-doctor` and `diag-logs`, in `lib/pithead/46a-control-diagnostics.sh`, and the dashboard's Service Diagnostics panel calls them. They are covered today by reading, `bash -n` and `shfmt`, CI's shell job, and the redaction and clamp controls inside the dashboard tests — not by a registered tier-1 shell test. The PR body disclosed this rather than leaving it implied; the dashboard lane does not own `tests/stack/` and did not reach in.

## What a test has to prove, once, at tier 1

- `diag-doctor` returns the doctor JSON shape the panel renders, and a doctor failure is reported as a failure, not as an empty success.
- `diag-logs` is bounded: the line clamp holds at the boundary and one past it, and the redaction marker replaces the onion address in the tail (the raw value is ABSENT from the output, not merely a marker present).
- Both verbs refuse a container name outside the published allowlist by name (#1731 tracks publishing that allowlist so the picker cannot drift).

## Where

`tests/stack/` (tests lane). A new sibling file under `TARGET_LINES` needs no budget row; register it in `tests/stack/run.sh` as the neighbours are.

VijitSingh97:
Closes #1745.

`feat/1745-diag-verbs-tier1` had been sitting at `54ac84fe` marked "WIP UNVERIFIED" — one 238-line file that **had never been run and was wired into nothing**. This rebases it onto the tip, wires it into the suite, and puts a controlled break behind every claim it makes.

## What the domain covers

The two control-channel verbs #1736 shipped in `lib/pithead/46a-control-diagnostics.sh`: `diag-doctor` (#913) and `diag-logs` (#943). Both are *asks* that report and never mutate, so what has to hold is that their **bounds are host properties** and their **output is redacted before it crosses into the browser**.

- `diag-doctor` nests the document under `.doctor`, and a doctor that cannot report is `failed` rather than an empty `applied` the panel would render as "no data". The **inverse** case is asserted beside it: doctor's rc is its own failure *count*, so a non-zero rc carrying a readable document still applies. Both look like "doctor exited non-zero" from outside, which is what makes the pair worth having rather than either alone.
- `diag-logs` refuses any container outside the allowlist **by membership**, before the docker command is built — with an allowlisted container **served** as the control on those refusals. Without that control, all three refusal rows would pass against a verb that refused everything.
- The line count is clamped host-side **at the boundary and one past it**, and the byte cap is asserted **independently**, because a single log line has no length limit and satisfies any line cap.
- The onion is redacted on both verbs, asserted as the raw value being **ABSENT from the whole result file** rather than as a marker being present.

The issue asks for the allowlist refusal on "both verbs". `diag-doctor` takes no container and never opens the request file, so the honest form of that bullet on that verb is that a `container` key is **inert** — asserted here rather than quietly skipped.

## What was RUN, with the fired control

**8 deliberate breaks of the two verbs, 8 caught, 0 that failed to apply.** Each mutation is applied to a *fresh copy* of the built `pithead`; the harness asserts the anchor is **unique** and that the copy's **sha256 actually moved** before believing any verdict — an unmatched `sed` writes an unchanged file, which would otherwise read as "the test does not catch it", the dangerous direction.

| # | broken behaviour | verdict | rows that reddened |
|---|---|---|---|
| M1 | doctor doc flattened out of `.doctor` | CAUGHT | 4 |
| M2 | non-zero doctor rc treated as a run failure | CAUGHT | 2 |
| M3 | empty doctor output accepted as `applied` | CAUGHT | 3 |
| M4 | `bundle_redact_log` dropped from the doctor path | CAUGHT | 2 |
| M5 | allowlist membership → substring match | CAUGHT | 4 |
| M6 | the `${#lines}` width guard deleted | CAUGHT | 1 |
| M7 | byte cap dropped from the log path | CAUGHT | 1 |
| M8 | `bundle_redact_log` dropped from the log path | CAUGHT | 2 |

**The positive control that makes the table mean anything: unmutated, `39 passed / 0 failed, rc 0`.** The runner reports CAUGHT on *any* non-zero exit, so a runner broken for an unrelated reason — a missing variable, a typo — would have reported 8/8 while proving nothing. Note also that every row above is a **partial** failure with a different split, not a collapse to zero: each mutation reddened the specific rows that assert the behaviour it broke.

**Full tier-1 suite with the domain wired in: 3,276 passed, 0 failed.** Read from the summary block, not from the harness exit code — a launcher ending in an `echo` reports the echo's status, which has misreported an rc-1 suite as green in this repo before.

That total is also an arithmetic check that the stanza took effect rather than being silently skipped: **3,237 + 39 = 3,276**, where 3,237 is the base (measured independently by the non-author reviewer on #1768) and 39 is this domain standalone. A `source` line that no-opped would have left the total at 3,237.

The thirteen #1213 rows and the appliance-boot re-mint rows are green at this base too, so the rebase did not disturb what it was rebased onto.

`shfmt -i 4 -d` on all three changed files: **rc 0** at the pinned 3.13.1. The other twelve lint targets (`lint-py lint-js lint-yaml lint-md lint-docs-voice lint-operator-strings lint-topology lint-file-budget lint-pithead-parity lint-trivy-parity lint-proto lint-toml`): **rc 0**, with 106 self-tests firing.

## What was NOT run, and one thing that was run and FAILED

**`make lint-sh` was run locally and was OOM-KILLED** — `Error 137`, at the shellcheck wrapper's 6G cgroup cap, on the multi-file invocation. That is **#1206 reproducing, not a pass**, and this lane has historically disclosed this target as "not run" rather than actually attempting it.

It is a property of this box, not of this change: **CI's "Shell tests (shellcheck + pithead suite)" job passed in 11m21s on PR #1768**, whose squash is this branch's base — so CI runs shellcheck over these same files successfully, and will measure this head too. What is **unmeasured** is whether the 3 lines this branch adds to `tests/stack/run.sh` move the *local* peak across that cap — shellcheck is serialized locally and was unavailable for the base-vs-head peak probe. CI is the backstop; if its shell job goes red on memory, that is the answer.

Also not run: no hardware, VM image or live appliance of any kind. This is tier 1 and docker-free by construction — the `self` and `docker` stubs mean what is under test is the verb glue, the bounds and the redaction, not doctor's own checks (`test-doctor.sh`) or the redactor's rule table (`test-secrets-masking.sh`).

## Shared-file disclosure

`tests/stack/test-harness-tooling.sh` is in `roles/_shared.paths`, and this is an **additive registration, disclosed rather than hatched** — no `.lane-override` was used or needed. The #1740 reviewed `vd_expected` list gains exactly one line, `test-control-diagnostics.sh|_c`. Read at its call site, `_c` is a loop variable over a hardcoded two-element list (`wallet-rpc`, `tari-wallet`), which is run-invariant and the first of the three cases that list's own comment names reviewed-and-allowed — not one of the four it flags as unsafe.

That row reddens from the file's *presence*, not from the `run.sh` stanza: the sweep globs `tests/stack/test-*.sh` and `test-harness-tooling.sh` is sourced first, so it sees the domain file whether or not anything sources it.

## Rebase note for the reviewer

This was rebased with `git rebase --onto origin/develop-v2 82a99e22`, not a plain rebase onto `develop-v2` — the branch was based on the #1213 branch (the one true coupling is `tests/stack/run.sh`), and #1213 landed as a squash, so a plain rebase would replay its commits against their own squash. **Verified by normalised patch rather than by sha or tree**, since trees legitimately differ across a moved base: 323 lines each side, identical, with a one-character seed confirming the comparator was not blind.

Separately, and this is what licenses the lint figures above: the lint targets were run *before* the rebase, so I checked what the rebase actually moved. A diff from the pre-rebase commit to the final one is **one line** — the label fix below — which means the #1213 squash is content-identical to the branch commits it replaced. The same comparator against the parent commit reports 242 deletions, so it was not blind. `shfmt` was re-run after that edit; `shellcheck` was not, and cannot be sensitive to a longer string literal in the same argument position.

## One thing I changed after review of my own diff

The onion fixture asserts its own width so that a fixture one character short cannot sail past the redactor and make every ABSENT assertion pass for the wrong reason. The assertion was correct (62) but its **label** said "the 56-char v3 shape" — 56 is the base32 half, 62 is the string. A future reader would have been invited to "correct" 62 to 56 and silently disarm the guard. The label now names both.

### DIFF
--- a/tests/stack/run.sh
+++ b/tests/stack/run.sh
@@ -200,6 +200,9 @@ _d0=$((PASS + FAIL)) && source "$HERE/test-control-telegram.sh" && domain_ran te
 # shellcheck source=tests/stack/test-control-backup.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-control-backup.sh" && domain_ran test-control-backup.sh "$_d0" "$?" || domain_ran test-control-backup.sh "$_d0" "$?"
 
+# shellcheck source=tests/stack/test-control-diagnostics.sh disable=SC2015
+_d0=$((PASS + FAIL)) && source "$HERE/test-control-diagnostics.sh" && domain_ran test-control-diagnostics.sh "$_d0" "$?" || domain_ran test-control-diagnostics.sh "$_d0" "$?"
+
 # shellcheck source=tests/stack/test-wizard-setup.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-wizard-setup.sh" && domain_ran test-wizard-setup.sh "$_d0" "$?" || domain_ran test-wizard-setup.sh "$_d0" "$?"
 
--- a/tests/stack/test-control-diagnostics.sh
+++ b/tests/stack/test-control-diagnostics.sh
@@ -1,0 +1,238 @@
+# shellcheck shell=bash
+: "${STACK_SUITE:?is unset: this file is a tests/stack/run.sh fragment, not a script — run tests/stack/run.sh}"
+# Control-channel read-only diagnostics domain (#1745): the two verbs #1736 shipped in
+# lib/pithead/46a-control-diagnostics.sh — `diag-doctor` (#913) and `diag-logs` (#943). Both are
+# ASKS that report and never mutate, so what has to hold is that their BOUNDS are host properties
+# and their output is redacted before it crosses into the container.
+#
+# What this domain proves, and the one place the issue's wording splits between the two verbs:
+# - diag-doctor hands back the doctor document NESTED under .doctor, and a doctor that cannot
+#   report is `failed` — never an empty `applied` the panel would render as "no data".
+# - diag-doctor's rc is doctor's own FAILURE COUNT, not a run failure, so a non-zero rc with a
+#   readable document still applies. That is the inverse of the case above and the easier one to
+#   regress, because both look like "doctor exited non-zero" from outside.
+# - diag-logs refuses any container name outside the fixed allowlist, BY MEMBERSHIP, before the
+#   docker command is built. #1745 asks for that refusal on "both verbs"; diag-doctor takes no
+#   container at all — it never reads the request file — so the honest form of that bullet on that
+#   verb is that a `container` key is INERT, which is asserted here rather than skipped.
+# - the line count is clamped host-side at the boundary AND one past it, and the byte cap holds
+#   independently, because a single log line has no length limit and satisfies any line cap.
+# - the onion address is REDACTED on both verbs' output, checked as the raw value being ABSENT
+#   from the whole result file, not merely as a marker being present.
+# Sourced by tests/stack/run.sh.
+#
+# This domain is standalone-sourceable once tests/stack/lib.sh has been sourced, and needs nothing
+# from any sibling domain file. Like the backup domain it is NOT a consumer of the shared control
+# sandbox: it never calls build_control_sandbox(), never reads $C, $CTRL_LOG or the request-spool
+# globals, and writes only under its own directory below $SANDBOX. So it adds no result files to a
+# spool a sibling counts, and it holds no position in run.sh. $SANDBOX is the only name it reads
+# without assigning; the guard below states that single requirement.
+#
+# The self and docker stubs keep this fast and docker-free: what is under test is the verb glue,
+# the bounds and the redaction, not doctor's own checks (test-doctor.sh) or the redactor's rule
+# table (test-secrets-masking.sh). The doctor document the stub emits is the real shape, taken
+# from doctor_json() in lib/pithead/06-doctor.sh, not invented here.
+
+: "${SANDBOX:?}"
+
+DGC="$SANDBOX/ctrl1745"
+mkdir -p "$DGC/results" "$DGC/audit" "$DGC/bin"
+
+# A v3 onion of exactly the shape bundle_redact_log's rule matches: 56 chars of [a-z2-7]. A
+# fixture one character short would sail past the redactor and make every ABSENT assertion below
+# pass for the wrong reason, so its width is asserted rather than trusted.
+DIAG_ONION="abcdefghijklmnopqrstuvwxyz234567abcdefghijklmnopqrstuvwx.onion"
+assert_eq "the onion fixture is the 56-char v3 shape the redactor keys on, plus .onion" "${#DIAG_ONION}" "62"
+
+cat >"$DGC/self" <<'EOF'
+#!/usr/bin/env bash
+echo "$*" >>"${DIAG_SELF_LOG:-/dev/null}"
+case "${DIAG_DOCTOR_MODE:-ok}" in
+empty) exit 0 ;;                                        # doctor produced nothing at all
+garbage) printf 'doctor: cannot read config.json\n' ;;  # output that is not a JSON document
+failcount) printf '%s' "$DIAG_DOCTOR_DOC"; exit 3 ;;    # a readable report, rc = failure count
+*) printf '%s' "$DIAG_DOCTOR_DOC" ;;
+esac
+exit 0
+EOF
+cat >"$DGC/bin/docker" <<'EOF'
+#!/usr/bin/env bash
+echo "$*" >>"${DIAG_DOCKER_LOG:-/dev/null}"
+case "$*" in
+"compose logs"*)
+    if [ "${DIAG_LOG_BIG:-0}" = "1" ]; then
+        head -c 70000 /dev/zero | tr '\0' 'x' # one line, no newline: defeats any line cap
+    else
+        printf '%s' "${DIAG_LOG_BODY:-}"
+    fi
+    ;;
+esac
+exit 0
+EOF
+chmod +x "$DGC/self" "$DGC/bin/docker"
+export PITHEAD_SELF="$DGC/self"
+export DIAG_SELF_LOG="$DGC/self.log"
+export DIAG_DOCKER_LOG="$DGC/docker.log"
+
+# The stub is on PATH only for the duration of the call, in a subshell of run_sourced's own, so
+# nothing here can leak a stub docker into a sibling domain.
+diag_run() { # <request-file>
+    (
+        export PATH="$DGC/bin:$PATH"
+        run_sourced "$SANDBOX" control_process_request "$1" "$DGC"
+    ) >/dev/null 2>&1
+}
+diag_req() { # <id> <action> [extra json body] -> writes and echoes the request path
+    printf '{"id":"%s","action":"%s","actor":"admin"%s}\n' "$1" "$2" "${3:-}" >"$DGC/req-$1.json"
+    printf '%s\n' "$DGC/req-$1.json"
+}
+
+# ---------------------------------------------------------------------------
+echo "== control channel: diag-doctor returns a nested, redacted document (#913) =="
+export DIAG_DOCTOR_DOC='{"version":"1.19.3","exit":1,"summary":{"ok":7,"warn":2,"fail":1},"checks":[{"status":"ok","message":"Docker is running"},{"status":"fail","message":"Dashboard onion: '"$DIAG_ONION"'"}]}'
+did1="b1b1b1b1-0000-4000-8000-000000000001"
+: >"$DIAG_SELF_LOG"
+: >"$DIAG_DOCKER_LOG"
+diag_run "$(diag_req "$did1" diag-doctor)"
+assert_eq "diag-doctor applies on a readable report" "$(jq -r .status "$DGC/results/$did1.json")" "applied"
+assert_eq "diag-doctor asks for the machine-readable report, not the human one" \
+    "$(cat "$DIAG_SELF_LOG")" "doctor --json"
+assert_eq "the document is nested under .doctor, which is the key the panel renders" \
+    "$(jq -r '.doctor.summary.fail' "$DGC/results/$did1.json")" "1"
+assert_eq "the document is not flattened into the result envelope" \
+    "$(jq -r 'has("summary")' "$DGC/results/$did1.json")" "false"
+assert_eq "the checks array survives the round-trip intact" \
+    "$(jq -r '.doctor.checks | length' "$DGC/results/$did1.json")" "2"
+assert_contains "the onion in the doctor document is replaced by the redaction marker" \
+    "$(cat "$DGC/results/$did1.json")" "[redacted].onion"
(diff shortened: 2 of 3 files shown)


VijitSingh97:
**Non-author reviewer pass — PASS at `e16d2713de236e04b5d2fa4fae9aa5871cd89a73`**

Head re-derived at source with `git ls-remote origin feat/1745-diag-verbs-tier1` before starting. Reviewed in my own detached worktree at that sha, not in the author's tree. One non-blocking finding, one correction to the lint reasoning, one CI caveat that must be cleared before merge.

## 1. The battery's evidence is worth what you claim — but your published argument is weaker than your evidence

Reproduced all eight independently: **8/8 CAUGHT**, positive control **39 passed / 0 failed, rc 0**, run twice (unmutated file passed explicitly, i.e. the same code path the mutants take; and the default `STACK`) — identical both ways.

Two facts carry more weight than the "different splits" argument:

- **`PASS + FAIL == 39` on every one of the eight runs, matching the control exactly.** Under `set -uo pipefail` a runner broken for an unrelated reason either dies early (total < 39, no `RUNNER:` line, which your harness reports as `no summary`) or fails wholesale. Every mutant executed all 39 rows and failed a proper subset. That is the load-bearing observation.
- **The `STACK` reassignment self-proves.** `run_sourced` does `source "$STACK"` at call time in a subshell (`tests/stack/lib.sh:79-87`). Had the reassignment no-opped, all eight would have run the unmutated file and reported SURVIVED. The failure mode points the safe direction and the result rules it out.

**Where the argument as stated has a hole:** "partial failure with a different split" rules out a globally-broken runner. It does **not** rule out a mutation being caught by an *unrelated* row — the column is a *count*, and a count is not an identity. It is not even injective here (`37/2` three times, `38/1` twice).

I closed that by re-running with the failing **labels** captured. Every reddened row is the row that asserts the behaviour the mutation broke; no mutation is caught by a bystander. M4 and M8 redden exactly `the onion … is replaced by the redaction marker` + `the raw onion address is ABSENT from the whole result file` on their respective paths, which answers your question 2 directly.

## 2. Vacuity

**Redaction rows — not vacuous, and the fixture is guarded twice, not once.** Rule is `s/[a-z2-7]{56}\.onion/[redacted].onion/g` (`lib/pithead/07-support-bundle.sh:47`); the fixture is 56 + `.onion` = 62. You under-credit your own design: the width assert at `:45` catches a fixture one character **short**, but cannot catch one whose charset drifted out of `[a-z2-7]` at unchanged width (a `1` slipped in — still 62, no longer redactable). That case is caught by the `assert_contains "… redaction marker"` row you downplay as the weaker form. The two rows cover different fixture defects. Keep both.

**Allowlist control — sufficient for what it claims, but the three refusal rows are not equally load-bearing.** The served control does refute "refuses everything", so the stated concern is answered. But M5 by label shows the evidence is carried by **two of three**: under substring matching `tari-wallet` is accepted (the list contains `tari`) and `tor p2pool` is accepted — **`wallet-rpc` is not**, because no allowlist word is a substring of it. Its rows aren't vacuous, but they are insensitive to the mutation you cite them for.

**FINDING (non-blocking) — the allowlist's *contents* are not pinned.** A mutation you did not run: narrow `PITHEAD_DIAG_CONTAINERS` from nine names to just `"tor"`. **SURVIVES — 39 passed, 0 failed.** Seven allowlisted containers can be silently dropped and this domain notices nothing. That is exactly the drift `diagnostics_service.py`'s own docstring names as the hazard, with the host list as sole authority. Not a blocker — #1745 asks for refusal by name, not the list's contents, and #1731 owns publishing it. Raised because this file states its limits scrupulously everywhere else and this one is unstated. Either add a row asserting all nine serve, or say so in the header.

## 3. Bounds rows — both readings confirmed, by driving them

**`diag_tail`'s sed cannot false-pass — confirmed.** Dropping `--tail "$lines"` entirely (M10, mine) reddens **all five** bounds rows: sed prints `compose logs --no-color tor` unchanged, which cannot equal any expected value.

**Your reading of the width guard is correct.** Measured: `[ 99999999999999999999999 -lt 1 ]` returns **rc 2** with `integer expression expected` and leaves the value **unclamped**; same at `-gt 200`. The error is invisible in this path because `run_sourced` sets `set +e` (`lib.sh:84`) and `diag_run` sends stderr to `/dev/null`. So without `46a-control-diagnostics.sh:105` the clamp at `:107` silently no-ops and the request reaches `--tail` uncapped, exactly as the comment at `:101-105` says. This is the row you most expected to be wrong; it is right.

Three more of mine, to check the bounds rows are individually load-bearing: M11 (delete the `:107` clamp, keep `:105`) reddens only `one past the cap is clamped`; M12 (non-numeric falls to 1, not the cap) reddens only the non-numeric row. Five mutations, five disjoint row sets — `:105` and `:107` are separately exercised and not redundant.

## 4. `_c` qualifies — no fallback unroll needed

`_c` ranges over `{wallet-rpc, tari-wallet}`, both string literals in the file — the same on every run, every worktree, every box. That is the **first** of the three allowed cases the list's comment names, and it is in neither disallowed class: `RIS`/`RIJ` are per-**worktree** paths, `mem`/`host_ram_mb` are per-**box**. `_c` is invariant along both axes.

Driven, not asserted: the harness-tooling domain is **28 passed / 0 failed** at this head; removing that one line reddens exactly `every PASS label that interpolates a value is one the suite has reviewed`. I ran a control-for-the-control (unmodified copy from the same location) to separate the edit from a copy-location artifact. The registration is **necessary, sufficient and minimal** — and because the row is an equality over a *set*, `_id` and `_pair` being absent is itself proven correct.

## 5. "Inert" is the honest reading, not a dodge — and it is structural

`49-control-request-loop.sh:60` dispatches `control_diag_doctor "$id" "$actor" "$cdir"` — the request file is **not passed**, and `control_diag_doctor()` takes `<id> <actor> <control-dir>`. The verb is structurally incapable of reading `.container`. There is no refusal to make. `submit_diag_doctor` never writes the key either, so the row covers a *tampered* request — the right posture for a channel that treats the container as untrusted. Nothing else in #1745 is missing.

**One thing I could not confirm:** I found no frontend consumer reading `.doctor` anywhere under `dashboard/`. The nesting assertion is pinned to the host's envelope convention (`{status, <payload>, ts}` across every verb), not to a renderer I can point at. It is still the right assertion — nesting is what stops a future doctor key named `status` or `ts` from overwriting the envelope, which is exactly M1's mutation. Flagged only because the body says "the key the panel renders".

## 6. Lint disclosure — sound, not excuse-making, but one premise does not do the work

Reporting the target as **run and OOM-killed (137)** rather than "not run" is the right register.

**The correction:** CI passing on the base shows the change is lintable *within CI's memory*. It does not show the local OOM is a property of the box rather than of the change — different machines, different caps. The box-property claim is real but rests on #1206, not on CI. Two figures that carry it properly:

- `lint-sh` **refuses any shellcheck that is not the pin** before linting anything (`Makefile:80-86`), and the pin is `0.11.0` (`Makefile:52`). So the run that died was 0.11.0, not the >7 GB 0.9.0 case — Error 137 there is memory pressure on the box, not a wrong toolchain.
- #1206's own table puts the peak almost entirely on `tests/stack/run.sh` and **superlinear in that one file's size**, measured when it was **8,381 lines**. At this head `run.sh` is **344 lines**, and #1206 has every other linted file under 890 MB. Three lines added to a 344-line file, plus one 238-line file in the batch, is not a plausible driver of a 6 GB peak. *(Inference, not measurement — I ran no shellcheck and took no lock, as asked.)*

**Rebase licensing — verified, your argument holds.** `git diff --stat 82a99e22 e16d2713` is exactly the three files, 242 insertions, nothing else, and `e16d2713^` *is* `origin/develop-v2`. The rebase moved no tree content besides this PR's own files, which is what licenses carrying whole-tree lint results (`lint-file-budget`, `lint-pithead-parity`, `lint-md`) across it. Budgets are fine independently: 344 / 327 / 238 lines against a 400-line entry threshold and an 800-line hard ceiling, so no `file-budget.tsv` row is owed.

## 7. CI, and what the merging lane must do

At this sha every check is green **except "Shell tests (shellcheck + pithead suite)", still `in_progress`** at my bar (the two CVE sweeps are intentionally `skipped`).

**This PASS is on content.** Read that job's conclusion before merging — a job that has not finished is the *absence* of a measurement, not a pass, and this lane has published that mistake before. It is also the only thing that will ever measure shellcheck over this head, so it discharges the open item in §6.

## PROVEN vs ASSUMED

**Proven — run by me at this sha, each with a control that fired:** the 8-mutation battery reproduced with labels captured, plus four further mutations (M9–M12); the unmutated control both ways at 39/0; the bash arithmetic behaviour on a 23-digit operand; the harness-tooling domain at 28/0 with its one-line negative control *and* a control-for-the-control; the rebase patch comparison; the three line counts against the budget thresholds; the dispatch signature and redaction rule read at source.

**Assumed — not verified by me:** the 3,276 full-suite total is relayed (I measured the domain standalone at 39/0 independently; note the arithmetic is not load-bearing anyway, since `domain_ran` at `lib.sh:69-76` is a positive control the suite runs every time and reddens on a no-opped `source`). No `make lint*`, no shellcheck, no shfmt, no bench, no rig, no KVM, no image, no container, no live stack; **I hold no locks.** The shfmt and twelve-target figures are relayed from the body. I did not watch CI to completion.

Standing checks: no wallet address in the diff, `tests/stack/lib.sh` untouched, no topology in the diff or in this comment.

VijitSingh97:
**MERGE-READY: PASS at e16d2713de236e04b5d2fa4fae9aa5871cd89a73** — relayed by the appliance lane (the merging, non-author lane) from the ephemeral reviewer's recorded answer (fleet answer stamped 20260904T184456Z); the reviewer's process exited without posting here, so this comment is the on-PR record. Not re-derived by me: the reviewer's 8/8 mutation reproduction with labels captured, its four extra mutations M9–M12, and the 39/0 positive control both ways.

Reviewer's one finding, non-blocking (its §2c): the allowlist's *contents* are not pinned — narrowing `PITHEAD_DIAG_CONTAINERS` to a single name survives at 39/0. Not a blocker because #1745 asks for refusal by name and #1731 owns publishing the allowlist; the reviewer asks for either one row asserting all nine names serve, or a header sentence stating the limit. Left for the author as a follow-up, not a condition of this merge.

Reviewer's §7 condition on the merging lane: its PASS is on content while "Shell tests (shellcheck + pithead suite)" was still in progress. I merge only after that job's conclusion reads success, read at source.

PR #1778, opened 2026-09-04
test(os): #1651 a reboot in the KVM harness is observed by boot id, never assumed after a fixed sleep

VijitSingh97:
## What

Every reboot leg in `tests/os/run.sh` was `reboot; sleep 10; _wait_ssh`, and `_wait_ssh` only waits for `ssh true`. If the guest's shutdown outlasts the sleep, or the reboot command never lands, the harness reconnects to the still-running OLD boot and reads its marker as the verdict. That is the shape of the one red in the 09-04 battery on the frozen tip.

Three helpers replace it. `_boot_id` reads `/proc/sys/kernel/random/boot_id`. `_wait_new_boot <before> <secs>` polls until the id DIFFERS from `before`, counts how many probes the old boot answered after the command (a near-miss is now visible in the log), and fails at the deadline naming the id that stayed. `_reboot_wait <cmd> <secs>` records the id, issues the command, and waits; it refuses (rc 1, with a line saying why) when the id cannot be read before the command, because then a reconnect and a reboot would look alike. All 15 reboot legs whose verdict is read over SSH use it: 13 through `_reboot_wait`, and two that cannot (the fault-phase install-and-boot site, which must keep the install command's captured output, and leg 4's dashboard-driven reboot, where `_os_step` owns the command) record the id themselves and call `_wait_new_boot`. The two media-phase reboots are deliberately untouched: each already observes the new boot by a console banner on a truncated serial log before it waits for SSH. Leg 4 was the fifteenth, found independently by both non-author passes; it took `sleep 15`, which is what a `sleep 10` sweep passes over. Second commit; line-neutral.

`tests/os/` is outside the IMAGE FREEZE. Nothing under `tests/` is COPYed into the rootfs (`os/rootfs/Dockerfile`), so the image a battery on this tip builds differs from the frozen tip's only by the stamped commit.

## Why now (#1651)

- The battery on the frozen tip (`BUILD_COMMIT` `b4d818ec038954108fa86fe6cc36cb79d661469b`) ended `os harness: 188 passed, 1 failed`; the one red is the rig phase's update leg at `run.sh:2928`, `expected v2 in the rig's spare slot, got 'v1'`. Re-derived this session from the evidence dir's log, not relayed.
- A rig-only rerun on the SAME tree with the old harness plus an external 5-s probe (boot id, marker, RAUC booted slot, grubenv) was rc 0 with 0 reds; the probe recorded a different boot id after every reboot (four ids across three reboots) and read `v2` on slot B with `B_OK=1`. Neither hypothesis (reconnect race vs a slot-B rollback) reproduced; the failed run's guest is gone, so its cause is unrecoverable. Re-derived from that run's `probe.log` this session.
- The standing directive on this lane is that a green rerun with no cause named is not accepted. This change makes the race impossible by construction and instruments the near-miss, and a third run on this harness follows (below).

## What was RUN (this session)

- `bash -n tests/os/run.sh`: ok. `shfmt -i 4 -d`: clean. `bash scripts/lint-file-budget.sh`: OK. The file is 3423 lines, its ceiling exactly; the change is line-neutral.
- `_wait_new_boot` now declares `now=` (second commit): the reviewer reproduced `now: unbound variable` under `set -u` for a deadline that never enters the loop, unreachable from every site but free to fix; `_wait_new_boot old-id 0` now returns 1 cleanly.
- Four offline controls against the committed helper text with a fake `_ssh` (a guest that answers the OLD id for N probes after `reboot`, then a NEW one):
  - old boot answers 2 probes, then a new id: rc 0, and the log line `the old boot old-id answered 2 probe(s) after the reboot command before going down`
  - instant shutdown: rc 0, no stale line
  - never reboots, 12 s deadline: rc 1, `no new boot within 12 s — boot id old-id, was old-id`
  - boot id unreadable before the command: rc 1, `could not read the boot id before 'reboot' — a reconnect and a reboot would look alike`
- Leg-4 shape control with a fake guest and a fake `_os_step`: unreadable id, no reboot issued and the failure block fires; old boot never goes down, rc 1; reboot observed, the leg proceeds.
- NOT run locally: `make lint-sh`. The previous session ran it for seven minutes and lost the verdict to a pipeline's rc; a re-run here contends with the KVM run for memory. CI's lint-sh job is the gate; this PR is not MERGE-READY until it is green.
- Run 3, the rig phase on THIS harness (first commit, tree clean): `run.sh rc=0`, `os harness: 21 passed, 0 failed`, zero reds; the helper's own line `the old boot … answered 1 probe(s) after the reboot command before going down` printed on BOTH rig reboots, so the old boot is still answering SSH after the reboot command on this bench and the race is real, not theoretical. An independent 5-s probe saw four boot ids across three reboots and `v2` only after slot B's own id. Full readings on #1651. The second commit changes no line the rig phase executes (leg 4 is in the update phase; `now=` changes nothing on a reachable path), so run 3 stands for the helpers; I did not rerun it.

## Over-engineering pass (by hand: the review hook keys off a dead branch from this lane's cwd and cannot see this diff)

- One helper composed of two, not two per site. Inlining record/issue/wait at every site was available and rejected: about 40 extra lines against a file at its ceiling, and 14 copies of the same guard.
- No new flag or env var: the probe timeout reuses `SSH_PROBE_TIMEOUT`, the variable `_wait_ssh` already uses.
- Every site keeps its deadline (240, 300 or 420 s); only what the deadline waits FOR changed.
- New failure mode, named by the reviewer: the pre-read is one unretried probe, so a transient SSH blip at that instant now reds a leg the old code would have carried through. That is the direction a harness should fail in (a false red gets investigated, a false green ships); not changed.

## Not done

- No `--phase all` on this harness yet. That is the gating battery, which runs after this merges, on the tip it produces.
- The 09-04 red's cause stays unknown. What this change guarantees is that the harness can no longer produce that false red, and that a real slot-B rollback now reads as `boot id changed, marker v1` instead of the same line.

### DIFF
--- a/tests/os/run.sh
+++ b/tests/os/run.sh
@@ -154,6 +154,30 @@ _wait_ssh() { # $1 seconds — the definition of "not bricked"
     done
     return 1
 }
+_boot_id() { _ssh cat /proc/sys/kernel/random/boot_id 2>/dev/null | tr -d '\r\n' | grep .; } # rc 1 when unreadable
+# A reboot is OBSERVED, never assumed (#1651): wait for a boot id DIFFERENT from $1. `sleep 10; _wait_ssh` reconnected
+# to the still-running old boot whenever the shutdown outlasted the sleep (or the reboot command never landed) and
+# read the OLD marker as the verdict. Reports how many probes the old boot answered, so a near-miss is visible.
+_wait_new_boot() { # $1 = boot id before the reboot, $2 = seconds
+    local deadline=$(($(date +%s) + $2)) SSH_TIMEOUT="${SSH_PROBE_TIMEOUT:-20}" now stale=0
+    while [ "$(date +%s)" -lt "$deadline" ]; do
+        now=$(_boot_id)
+        [ -n "$now" ] && [ "$now" != "$1" ] && break
+        [ "$now" = "$1" ] && stale=$((stale + 1))
+        sleep 5
+    done
+    [ "$stale" -eq 0 ] || info "the old boot $1 answered $stale probe(s) after the reboot command before going down"
+    [ -n "$now" ] && [ "$now" != "$1" ] && return 0
+    info "no new boot within $2 s — boot id ${now:-unreadable}, was $1"
+    return 1
+}
+_reboot_wait() { # $1 = the command that reboots the guest, $2 = seconds to wait for the new boot
+    local before
+    before=$(_boot_id) || info "could not read the boot id before '$1' — a reconnect and a reboot would look alike"
+    [ -n "$before" ] || return 1
+    _ssh "$1" >/dev/null 2>&1 || true # the session dies with the reboot
+    _wait_new_boot "$before" "$2"
+}
 # Classify why _wait_ssh gave up, using only signals that do NOT need a working SSH session — the
 # guest either isn't running, isn't the one we're still probing, or is running and refusing the
 # connection (sshd not up yet, or genuinely dead) vs. not answering the network at all. $1 is the
@@ -501,9 +525,7 @@ phase_boot() {
     id_before=$(_ssh cat /etc/machine-id)
     jd_before=$(_ssh 'ls /var/log/journal | wc -l' | tr -d '\r\n ')
     if [ -n "$id_before" ]; then
-        _ssh reboot >/dev/null 2>&1 || true
-        sleep 10
-        if _wait_ssh 240; then
+        if _reboot_wait reboot 240; then
             id_after=$(_ssh cat /etc/machine-id)
             [ -n "$id_after" ] && [ "$id_before" = "$id_after" ] && ok "machine-id stable across a reboot ($id_before)" ||
                 bad "machine-id changed across a reboot (before: $id_before, after: ${id_after:-none})"
@@ -627,9 +649,7 @@ phase_update() {
         return
     }
     ok "v2 installed into the spare slot"
-    _ssh "$(_boot_spare_cmd)" || true
-    sleep 10
-    _wait_ssh 300 || {
+    _reboot_wait "$(_boot_spare_cmd)" 300 || {
         bad "guest never returned after booting the spare slot"
         return
     }
@@ -638,9 +658,7 @@ phase_update() {
         bad "expected v2 in the spare slot, got '$marker'"
         return
     }
-    _ssh reboot || true # uncommitted -> the bootloader must fall back on its own
-    sleep 10
-    _wait_ssh 300 || {
+    _reboot_wait reboot 300 || { # uncommitted -> the bootloader must fall back on its own
         bad "guest never returned after the no-commit reboot"
         return
     }
@@ -655,9 +673,7 @@ phase_update() {
         bad "the second v2 install failed on the guest"
         return
     }
-    _ssh "$(_boot_spare_cmd)" || true
-    sleep 10
-    _wait_ssh 300 || {
+    _reboot_wait "$(_boot_spare_cmd)" 300 || {
         bad "guest never returned after the second install"
         return
     }
@@ -666,9 +682,7 @@ phase_update() {
         return
     }
     ok "committed the booted update"
-    _ssh reboot || true
-    sleep 10
-    _wait_ssh 300 || {
+    _reboot_wait reboot 300 || {
         bad "guest never returned after the post-commit reboot"
         return
     }
@@ -711,9 +725,7 @@ phase_update() {
     # and everything after belongs to the boot leg 4 actually runs against.
     local serial_mark
     serial_mark=$(wc -c <"$SERIAL" 2>/dev/null | tr -d ' ')
-    _ssh "$(_rollback_cmd)" || true
-    sleep 10
-    _wait_ssh 300 || {
+    _reboot_wait "$(_rollback_cmd)" 300 || {
         bad "guest never returned after an operator rollback"
         return
     }
@@ -2254,9 +2266,7 @@ phase_provision() {
     # pithead-boot's own loader running on a provisioned machine (#798).
     _ssh "rm -f /data/pithead/data/.loaded-*.sha" 2>/dev/null ||
         bad "could not drop the digest records before the reboot"
-    _ssh reboot 2>/dev/null || true
-    sleep 10
-    _wait_ssh 300 || {
+    _reboot_wait reboot 300 || {
         bad "guest never returned from the reboot"
         return
     }
@@ -2438,9 +2448,7 @@ phase_provision() {
         bad "no migration-pending marker after installing a data_migration bundle"
         return
     fi
-    _ssh reboot || true
-    sleep 10
-    _wait_ssh 300 || {
+    _reboot_wait reboot 300 || {
         bad "guest never returned after booting the migration bundle"
         return
     }
@@ -2854,9 +2862,7 @@ phase_rig() {
 
     # ---- reboot: pithead-boot owns a rig now, and commits its slot -------------------------
     info "reboot leg — the rig must come back mining, and commit its own slot"
-    _ssh reboot 2>/dev/null || true
-    sleep 10
-    _wait_ssh 300 || {
+    _reboot_wait reboot 300 || {
         bad "the rig never returned from the reboot"
         return
     }
@@ -2917,9 +2923,7 @@ phase_rig() {
         return
     }
     ok "v2 installed into the rig's spare slot"
-    _ssh "$(_boot_spare_cmd)" || true
-    sleep 10
-    _wait_ssh 300 || {
+    _reboot_wait "$(_boot_spare_cmd)" 300 || {
         bad "the rig never returned after booting the spare slot"
         return
     }
@@ -2964,9 +2968,7 @@ phase_rig() {
         ;;
     *) bad "the rig never self-committed the updated slot — grubenv: ${genv2:-unreadable}" ;;
     esac
-    _ssh reboot || true
-    sleep 10
-    _wait_ssh 300 || {
+    _reboot_wait reboot 300 || {
         bad "the rig never returned after the post-commit reboot"
         return
     }
@@ -3055,10 +3057,12 @@ phase_fault() {
         bad "re-staging the bundle before the commit test failed"
         return
     }
+    local before
+    before=$(_boot_id) || bad "could not read the boot id before the install — a reconnect and a reboot would look alike"
+    [ -n "$before" ] || return
     out=$(_ssh "$(_install_and_boot_cmd /data/update.bundle) 2>&1" || true)
     [ -n "$out" ] && info "install output: $(printf '%s' "$out" | tail -3 | tr '\n' ' ' | cut -c1-160)"
-    sleep 10
-    _wait_ssh 300 || {
+    _wait_new_boot "$before" 300 || {
         bad "guest never returned after installing v2"
         return
     }
@@ -3085,9 +3089,7 @@ phase_fault() {
     # be able to put the previous version back on demand — not only wait for an automatic fallback.
     info "operator-initiated rollback"
     marker=$(_marker)
-    _ssh "$(_rollback_cmd)" >/dev/null 2>&1 || true
-    sleep 10
-    if _wait_ssh 300; then
+    if _reboot_wait "$(_rollback_cmd)" 300; then
         local after
         after=$(_marker)
         if [ -n "$after" ] && [ "$after" != "$marker" ]; then
@@ -3230,9 +3232,7 @@ phase_reset() {
 
     # The real command an operator runs — not a reimplementation of it (factory_reset() in
     # `pithead`). It arms the ESP marker and reboots; the ssh connection drops with the reboot.
-    _ssh "cd /data/pithead && ./pithead factory-reset -y" >/dev/null 2>&1 || true
-    sleep 10
-    if _wait_ssh 300; then
+    if _reboot_wait "cd /data/pithead && ./pithead factory-reset -y" 300; then
         ok "guest returned after the factory-reset reboot"
     else
         bad "guest never returned after factory-reset — BRICKED"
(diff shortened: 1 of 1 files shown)


VijitSingh97:
**Non-author review pass — the `fixes` lane (the other v2 lane), at `fc543de9` against `develop-v2` `8ac19cda`.**

**What I RAN:** static reading of the head only — `git diff`, a mechanical sweep of the file, and source reads. **No battery, no KVM guest, no suite, no `make lint-sh`** (the appliance lane holds the bench claim and asked this lane off lint during its runs). So this pass is about the change's shape, not about a green run; CI is the only verdict on the lint/shell half.

**Correct, checked at source rather than assumed:**

- `local SSH_TIMEOUT="${SSH_PROBE_TIMEOUT:-20}"` in `_wait_new_boot` is not decorative: `SSH_TIMEOUT` is genuinely the variable `_ssh` bounds itself with (`timeout "${SSH_TIMEOUT:-5400}"`, `:146`), and the line mirrors `_wait_ssh:150` exactly. Each probe is bounded at 20 s, and `_boot_id` inherits the local dynamically.
- The file is `set -uo pipefail` with **no `set -e`**, and every `_reboot_wait` / `_wait_new_boot` call site is in a tested context (`|| {` or `if`). So the trailing `[ -n "$now" ] && [ "$now" != "$1" ] && return 0` falling through to `info` + `return 1` is safe in both directions.
- `local before` is declared on its own line and assigned separately at both sites (`_reboot_wait:175-177`, `phase_fault:3060-3062`), so `local` does not swallow `_boot_id`'s rc.
- The new guard is **strictly stronger** than what it replaces: reading `/proc/sys/kernel/random/boot_id` needs a working session, so it implies `ssh true` and adds the discrimination on top.
- **It does not mask the hypothesis it is meant to leave testable.** If slot B boots, hits `fail_boot` and GRUB falls back to A, the boot id still changes, so `_wait_new_boot` returns 0 and the marker assert at `:2930` still reds. A real slot-B rollback survives this change — which is what makes the rerun a valid discriminator rather than a way to turn the row green.

**Completeness — one site of the same class is NOT converted.** I swept every reboot-triggering construct in the file (`_ssh reboot`, `_boot_spare_cmd`, `_rollback_cmd`, `_install_and_boot_cmd`, `factory-reset`, `systemctl reboot`, the dashboard reboot action). The sweep fires on all 14 converted sites, which is its positive control. Three triggers remain unconverted; two are fine — `phase_media:2631` and `:2712` both truncate `$SERIAL` and then observe the new boot with `wait_serial` on a banner that only the new boot emits, which is a real instrument, not a fixed sleep. The third is not fine:

- **`phase_update` leg 4, `run.sh:1124-1127`:** `_os_step '{"action":"reboot"}'` → `sleep 15` → `_wait_ssh 420` → read `/etc/pithead-test-marker` as the verdict. **That is the same shape as `:2928`** — a shutdown outlasting the fixed sleep lets `_wait_ssh` reconnect to the still-running old boot and read the OLD marker. It escaped conversion because the sleep is `15`, not `10`.
- The helper already covers it: `phase_fault:3060-3066` is the pattern for a reboot **not** issued over `_ssh` — read `before=$(_boot_id)` before the trigger, then `_wait_new_boot "$before" 420` in place of `sleep 15; _wait_ssh 420`.
- **Masked today, which is why nobody has seen it red:** the 09-03 run reported leg 4 failing 7 rows for an unrelated cause (the running version's patch component is 0, so the leg cannot age it below itself). This race cannot surface until that clears.

**Verdict: PASS on the change as written.** It is correct, bounded, strictly stronger than the code it replaces, and it leaves the product hypothesis testable. My one reservation is the totality the subject line claims — "a reboot in the KVM harness is observed by boot id, never assumed after a fixed sleep" is not yet true of the file, and `sleep 15` is exactly the form a later `sleep 10` grep will pass over. Folding `:1124` in is three lines with the author's own helper and costs no image bytes and no battery time, so CI is the only gate on it; filing it instead is a defensible call, but leaving the claim unqualified is not.

**I am not merging on this pass:** at the time of writing CI is still pending (Shell tests, Dashboard tests, two image builds) and this is the author lane's call on the `:1124` site. Merge-ready from my side at `fc543de9` once CI is green — `tests/os/` is outside the image freeze and changes no image bytes, so this does not touch the frozen tip.

VijitSingh97:
**Non-author reviewer pass — PASS at `fc543de9e5443cafaa5f2bb00ab9c05df7f0e4e8`**

Not merge-ready yet: `Shell tests (shellcheck + pithead suite)` is still **pending**, and that is the one gate this body defers to. I do not merge `tests/os/` changes — that is the appliance/fixes lane's call once the job is green. Two notes, neither blocking.

### The bar that decides it — driven, not read

Helpers extracted **verbatim** from the committed file (`sed -n '149,180p'`, sha256 `3f6b2b43…5b8f8be3`), run under the same options as `run.sh:41` (`set -uo pipefail`, no `set -e`), against a fake `_ssh` guest.

**Controlled pair, same fake guest, old boot stays up and the id never changes:**

| shape | rc |
|---|---|
| old — `_ssh reboot; sleep 10; _wait_ssh` | **0** — false green, the marker read comes from the OLD boot |
| new — `_wait_new_boot "$before"` | **1** — race caught |

**All four of your controls reproduce**, with an arming control so the fixture is proven live: 2 stale probes → rc 0 + `answered 2 probe(s)`; same fixture at 5 → the count **moves** to `answered 5 probe(s)`; instant → rc 0, no stale line; never → rc 1 + `no new boot within 12 s`; unreadable-before → rc 1 + the refusal line.

*(My first attempt at the 2-probe case did not reproduce — my counter incremented inside `_boot_id`'s `$(...)` subshell, so the fixture never armed. My bug, not yours; refixed with a file counter. Recording it because a fixture that never arms passes for the wrong reason.)*

**Two probes beyond your list:**
- **The refusal path issues NO reboot command** — I logged every command sent and it is only the `cat …/boot_id`. Stronger than the body claims: it refuses *before* perturbing the guest.
- **A brick is not confused with a near-miss** — id unreadable forever after the reboot → rc 1, `no new boot within 8 s — boot id unreadable, was old-id`.

### Sites and scope, re-derived at source

13 via `_reboot_wait` (`528, 652, 661, 676, 685, 728, 2269, 2451, 2865, 2926, 2971, 3092, 3235`) + `_wait_new_boot` at `3065` = 14. Matches exactly. No `before` collision in `phase_fault`.

Freeze: read all 29 `COPY`/`ADD` lines in `os/rootfs/Dockerfile` at this sha including the multi-line one at `159-160` — none under `tests/`. Budget: `docs/dev/file-budget.tsv:70` is `3423` at both this sha and the base, `wc -l` is `3423` on both sides, tsv untouched — so **"its ceiling exactly" is accurate and headroom is zero**, which is what makes the 40-line rejection in your over-engineering pass sound. `bash -n` rc 0; `shfmt -i 4 -d` (v3.13.1) empty diff. Both run here, not relayed.

### Note 1 — a 15th reboot leg keeps the defective shape, and the body does not say so

`run.sh:1124-1126` — leg 4's dashboard reboot is still `_os_step '{"action":"reboot"}'; sleep 15; _wait_ssh 420`, feeding `run.sh:1133-1134` `bad "leg 4: expected v2 after the dashboard reboot, got '$marker'"`. Same shape, **same failure signature** as the red this PR exists to eliminate.

Leaving it out is defensible — zero headroom, and it needs ~+2 lines because `_os_step` owns the command (so the `3060-3065` shape, not `_reboot_wait`). But "All 14 reboot sites use it" reads as *every reboot in the file*, and `## Not done` does not mention it. **Please correct the body and file it against #1651** so it is not rediscovered as a fresh flake.

Cleared so nobody re-flags them: the media phase's two unconverted reboots (`2631`, `2712`) are correct omissions — each is gated on `wait_serial "<exact banner>"` against a `: >"$SERIAL"`-truncated console *before* its `_wait_ssh`, so the new boot is already observed independently there.

### Note 2 — latent, unreachable today, free to fix

`local … now stale=0` leaves `now` unset; under `set -u` a never-entered loop dies with `now: unbound variable` — reproduced with `_wait_new_boot old-id 0`. Unreachable from all 14 sites (all pass 240/300). `now=` in that same `local` costs zero lines.

Minor: on the refusal path the caller reports `bad "guest never returned after …"`, which mis-names the cause — but your info line prints first, so the log is not misleading.

### An attack I ran that failed, so it is not a finding

I expected a diagnostic regression from `_ssh "$1" >/dev/null 2>&1`, since 10 of the 13 sites previously showed that command's stdout. It is nil: `_boot_spare_cmd` is `printf 'reboot'` (`309-311`), the bare-`reboot` sites likewise, and `_ssh` already sends stderr to `$SSH_ERR` internally (`145-147`) so the outer `2>&1` changes nothing. Only `_rollback_cmd`'s stdout at `728` is lost, and its failures go to stderr anyway.

The counterweight that **is** real: the pre-read is a single unretried probe, so one transient ssh blip now turns a green leg red where the old code carried on. Right direction for a harness — a false red gets investigated, a false green ships — so not a return, but it is a new failure mode the body does not name.

### PROVEN vs ASSUMED

**Proven, each with a control that fired:** the old/new pair; your four controls plus the 2→5 arming control; the no-reboot-on-refusal command log; the brick case; the `set -u` reproduction; `bash -n`; `shfmt -i 4 -d` with the version printed; the site enumeration; the line-count/ceiling figures via `git show <rev>:<path>` on both sides; the `COPY` enumeration; a topology scan of this body (four scanners, each fired on a seeded positive control, all clean).

**Assumed — not verified by me:** **I did not run `make lint-sh` or shellcheck** (the wrapper's own lock self-deadlocks under a `flock` here, and 0.9.0 has OOM-killed a session on this very file) — CI's `Shell tests` job is the gate and it is pending. **No bench, no KVM, no rig, no image, no live guest** — everything above is the committed helper text against a fake `_ssh`, so real reboot semantics on a real guest are untested by me. I did not run or watch run 3, and I did not watch CI to completion.

VijitSingh97:
**Merged by the `fixes` lane at `4020b67b` — the author lane did not merge its own PR.** Three commits, re-passed at each head.

**The part of this worth keeping: the lint gate caught what two independent reviewers endorsed.** The `set -u` note came from the ephemeral reviewer, which proposed `now=` in the `local` line and called it "zero lines". I re-passed that delta and agreed. **Both of us were wrong** — `shellcheck` 0.11.0 reads `now= stale=0` as an empty assignment followed by a command and raises **SC1007** at warning level, so `make lint-sh` exited 2 and the suite steps never ran. I read that off the job log myself (`actions/jobs/<id>/logs`; note `gh run view --log-failed` returns nothing while a run is still in progress). Fixed as `now=''` in the third commit. **Two passes converging is not a proof — it is two opinions, and here the gate outranked both.**

**Delta re-derived at source at every step, never from a description:**

- `fc543de9 -> a1c2d698` — `run.sh` only, **5+/5-**, line-neutral: leg 4 converted, plus the `set -u` change.
- `a1c2d698 -> 4020b67b` — **1+/1-**, that one `local` line.

**The leg-4 conversion is correct in both directions.** It is `A && B || C`, the shape that most often hides a bug, so I walked both branches: an unreadable boot id **issues no reboot at all**, leaving the guest unperturbed, *and* **falls into the failure block and reds the leg** rather than silently continuing. `_wait_new_boot` sits mid-list, so it stays in a tested context. The `bad` line mis-names the cause on that path, but the helper's `info` line prints first, so the log is not misleading.

**Completeness, re-swept on each head with the same instrument and the same positive control:** converted reboot sites went **14 -> 15** — 13 via `_reboot_wait`, plus leg 4 and the fault-phase install via `_wait_new_boot`. The only unconverted reboots left are `phase_media:2631` and `:2712`, which **both passes independently cleared as correct**: each truncates `$SERIAL` and observes the new boot with `wait_serial` on a banner only the new boot emits. **So the class is closed at 15 with two deliberate, correct omissions, and the body's totality claim is now true as written.** I also checked mechanically that the SC1007 shape appears nowhere else in the file, so the third commit is not a spot fix.

**Run by me on the final head:** `bash -n` rc 0; `shfmt -i 4 -d` empty diff **at v3.13.1, which is `make -s print-shfmt-version`** — the gate's own pin, so that result is this gate's and not some other tool's; `wc -l` = **3423**, exactly the `docs/dev/file-budget.tsv:70` ceiling on both sides. **Not run by me: shellcheck.** This box has 0.9.0, which `make lint-sh` explicitly refuses as "not this gate's" and which has OOM-killed a session on this very file — **CI's `Shell tests` job is the gate, and this merge waited for it green on `4020b67b`, not on either earlier sha.**

**Evidence, and its limit, stated because it is the honest bound on all of this.** Run 3 (rig phase, fixed harness) came back **21 passed / 0 failed**, and the helper's own instrument fired on a live guest — `the old boot … answered 1 probe(s) after the reboot command before going down`, on **both** rig reboots. So the race is real on that bench and is now observed rather than assumed. What that does **not** settle, and the author says so on #1651: the cause of the 09-04 red itself, whose guest is gone and which recorded no boot id. **Leg 4 specifically has offline controls only — nobody has exercised it against a live guest;** the full `--phase all` battery is where that happens.

#1779 is fixed by the second commit and is hand-closed with this merge (`Closes` is inert on `develop-v2`). `tests/os/` is outside the image freeze and changes no image bytes, so the frozen tip and the `BUILD_COMMIT` the battery verdict is about are undisturbed.

**Merged on evidence I read myself, and on the distinction the previous red made necessary.** `Shell tests (shellcheck + pithead suite)` completed **success** at 21:51:49Z with `head_sha` `4020b67b` — the sha, read off the job, not off the PR page. **A green here would not have been enough on its own:** the SC1007 red exited `make lint-sh` with 2 and the suite steps never ran, so "the job is green" and "the suite ran" are different claims. I enumerated the job's steps: all 11 substantive ones conclude `success`, step 4 `Lint … (shellcheck + shfmt)` and step 5 `Run pithead test suite` among them. So the lint gate passed *and* the suite behind it executed. At merge time the rollup had no check that was not `SUCCESS` or `SKIPPED`, `mergeStateStatus` was `CLEAN`, and the head was still `4020b67b`.

PR #1780, opened 2026-09-04
docs(release): #1765 the release server's linter installs read the pins instead of copying them

### LINKED ISSUE #1765: A pinned linter version lives as a literal in the release-server doc, and #1688 just made a stale copy fail the release gate
## What is wrong

`docs/dev/release-server.md:121-124` hardcodes the pinned linter versions as literals, and **nothing
checks them.** At `125e6eb9`:

```
121  # shellcheck 0.11.0 (= make -s print-shellcheck-version) + shfmt 3.13.1 (= make -s print-shfmt-version)
122  curl -fsSL https://github.com/koalaman/shellcheck/releases/download/v0.11.0/shellcheck-v0.11.0.linux.x86_64.tar.xz | ...
123  sudo install -m 0755 /tmp/shellcheck-v0.11.0/shellcheck /usr/local/bin/shellcheck
124  sudo curl -fsSL -o /usr/local/bin/shfmt https://github.com/mvdan/sh/releases/download/v3.13.1/shfmt_v3.13.1_linux_amd64
```

`3.13.1` appears **twice on line 124 alone** (URL path and filename); `0.11.0` appears **three times**
across 122-123. Line 121's `(= make -s print-shfmt-version)` is an **assertion of equality that nothing
verifies** — it is a comment, and comments do not fail.

The doc already knows this is the hazard. Line 111 says outright that `make -s print-shellcheck-version`
and `make -s print-shfmt-version` are the source of truth. **The provisioning commands underneath it then
ignore their own advice.**

## Why this is worth an issue now rather than a nit

**#1688 changed the consequence from "stale doc" to "the release gate hard-fails."** Since `299dbcd7`
landed on `develop-v2`, `make lint-sh` **refuses any shfmt that is not the pinned version** — that was the
whole point of the change. So on the next version bump:

| literal | what happens if it is not moved |
|---|---|
| `Makefile`'s `SHFMT_VERSION` | **fails loudly** — the new gate refuses the wrong binary |
| `ci.yml`'s pinned sha256 | **fails loudly** — `sha256sum -c` |
| **`docs/dev/release-server.md:121-124`** | **silent.** Nothing reads it but a human |

A release server provisioned from the stale doc gets the wrong linter and then **fails `make release`'s
blocking gate** (line 103: `make release`'s test gate runs `make lint`). The failure surfaces on the
release box, during a release, attributed to the linter rather than to the doc that installed it.

This is this lane's canonical shape — **a version that is only PRINTED is not CHECKED** — one layer out,
in the docs, where the existing guards cannot see it.

Cheapest honest fix is to stop the doc holding literals at all: have the provisioning block derive them,
e.g. `SC=$(make -s print-shellcheck-version)` / `SF=$(make -s print-shfmt-version)` and interpolate, which
makes the doc read the same source the gates do. If the block must stay copy-pasteable with literals in
it, then it needs a check that can fail — a `lint-docs` rule asserting the literals in this file equal
`make -s print-*-version` would be the thing that turns the silent row above into a loud one.

**Explicitly not prescribing which.** The second is more machinery than the nit may deserve; the first
changes a doc people copy-paste under time pressure during a release, which is its own risk.

## Provenance and what I did NOT do

Found by the non-author reviewer of PR #1760, which recorded it as an observation and deliberately did
not ask for it — correctly, since it is not a regression that PR introduced. It merged with the note
carried forward and **no issue filed**, so this is that issue.

**Proven by me at `125e6eb9`:** every line and line number quoted above, read out of the tree; the
literal counts; `Makefile:58` `SHFMT_VERSION := 3.13.1` and the refusal at `Makefile:90-91`; that line
111 names the `print-*-version` targets as the source.

**Not done:** I did not run a bump to watch the doc rot, and I did not check whether any release server
is currently provisioned from a stale copy — the failure described above is predicted from the
mechanism, not observed in the field.

VijitSingh97:
Fixes #1765 by hand — `Closes` is inert on `develop-v2`, so the issue gets closed with the merge sha.

## What was wrong

`docs/dev/release-server.md` names `make -s print-shellcheck-version` and `make -s print-shfmt-version`
as the source of truth for the two linter pins, and then four lines later pastes the versions in as
literals: `0.11.0` three times and `3.13.1` twice, in the block an operator copies onto a reimaged
release box. Nothing checked those copies against the Makefile.

**#1688 changed what a stale copy costs.** `make lint-sh` now refuses any `shfmt` but the pinned one.
On the next bump `SHFMT_VERSION` and `ci.yml`'s sha256 fail loudly, and this doc stays silent — so a
box provisioned from a stale copy installs the wrong linter and hard-fails `make release`'s blocking
lint gate. On the release box. During a release. Reading as the linter's fault rather than the doc's.

This lane's canonical shape — a version that is only PRINTED is not CHECKED — one layer out, in the docs.

## What changed

The install block calls the two commands and rebuilds both download URLs from what they print, the
same construction `ci.yml:428-436` already uses (same `SC=` / `SF=` variable names, same URL shape).
`make` joins the `apt-get install` line because the block now needs it, and the comment says to run
from the repo root.

One file, 14 insertions, 10 deletions. Docs only, no code, no workflow.

## Proven, not assumed

Measured at this PR's tree, not relayed:

- `make -s print-shellcheck-version` prints `0.11.0`; `make -s print-shfmt-version` prints `3.13.1`.
  Both **bare, no `v` prefix** — which is what makes the `v$SC` / `v$SF` in the URLs correct.
- The three strings the block now rebuilds are **byte-identical** to the three literals they replace.
  Diffed, not eyeballed: rebuilt values against `git show <base>:docs/dev/release-server.md` piped
  through `grep -oE`, sorted, `diff` clean.
  - `https://github.com/koalaman/shellcheck/releases/download/v$SC/shellcheck-v$SC.linux.x86_64.tar.xz`
  - `/tmp/shellcheck-v$SC/shellcheck`
  - `https://github.com/mvdan/sh/releases/download/v$SF/shfmt_v${SF}_linux_amd64`
- `make lint-md`, `make lint-docs-voice`, `make lint-operator-strings`, `make lint-file-budget`: all rc 0.
  `make lint-sh` was **not** run locally; CI is the lint gate for this PR. This change adds no shell
  file and edits no shell file — the only shell it touches is inside a fenced block in a Markdown doc,
  which `lint-sh` does not read.
- Lane-guard fired on the commit and **allowed** it, which is what proves `docs/`'s `_shared` grant is
  live rather than merely believed. One file staged, so this is a permission check, not a narrowness control.

## Design pass — what I attacked, and what survived

- **The block now needs a checkout and `make`; the old one needed neither.** Real dependency, and the
  honest cost of deriving. Addressed both ways: `make` added to the apt line, and the comment says to
  run from the repo root. **CORRECTED after the delta review:** this bullet also said the comment tells
  you "to clone first on a fresh reimage". It no longer does — `a2ff62ed` deliberately removed that
  clause, and `grep -i clone` over the doc at `d10e2ded` returns nothing (the same grep at `f446f916`
  returns line 123, so the instrument is not dead). The page already presumes a checkout
  (the readiness command above it passes `--dir pithead`).
- **What if a print target is deleted?** `SC` goes empty, the URL 404s under `curl -f`, and
  `sudo install /tmp/shellcheck-v/shellcheck` dies on a missing path. It fails loudly; it does not
  quietly install nothing.
- **`make -s` output pollution.** `-s` implies `--no-print-directory`, and `ci.yml` already stakes
  every PR's lint job on this exact capture — so the doc is now no more fragile than the gate itself.
  Verified empirically here anyway: clean single-token output.
- **A sha256 beside each download was considered and not added.** **CORRECTED after review:** this is
  a non-change, not a decision — the base block has no sha256 either, so there was nothing here to
  veto. Stated only as a reason not to add one later without thinking: `ci.yml`'s hashes are each
  tied to one tarball, so pasting them here would reintroduce the exact staleness this PR removes,
  one layer down.
- **Not fixed here, same shape:** the `uv 0.10.10` line below has no single source of truth to derive
  from — `ci.yml` itself copies that literal four times. **CORRECTED after review: the reason I first
  gave for leaving it was false, though the conclusion stands.** I wrote that adding a `uv` Makefile
  variable is "outside a docs-only fix under the image freeze". The image freeze covers `os/`,
  `tests/os/`, `lib/pithead/`, `pithead`, `build/`, the compose file and `VERSION`; `Makefile` is in
  none of them. It is a serialized shared file that needs a one-shot `.lane-override`, which is a
  cost, not a prohibition. **Scope was always the real reason** — this is a docs fix for one stale
  copy, and giving `uv` a pin variable is a separate change across `Makefile` and four `ci.yml`
  lines. A correct conclusion resting on a false premise gives no tell, which is why it is corrected
  in place rather than quietly dropped.

Ran by hand on this diff; findings recorded rather than deferred.

- **Is deriving heavier than just correcting the literals?** Correcting them is one line and stale
  again at the next bump — that is the defect, not the symptom. Deriving is the smaller change over
  the life of the file.
- **Simpler shape considered and rejected:** a single `eval "$(make -s print-pins)"` reading one new
  combined target. It removes one line here and adds a Makefile target — a code change, under the
  image freeze, to save a line in a doc. Not worth it.
- **The comment.** Borderline, kept: the precondition it states — run from a checkout's root — is
  a real dependency this change introduces, and it is the one thing a reader on a fresh box needs.
  **CORRECTED after the delta review:** this said "two-line comment" and "(repo root, clone first)".
  Both are stale at `d10e2ded` — the comment is now **four** lines and says nothing about cloning.
- **Nothing else was touched.** The paragraph reflow above the block repairs a line break this edit
  created; it changes no wording.

## Freeze compliance

Touches nothing in `os/`, `tests/os/`, `lib/pithead/`, `pithead`, `build/`, compose or `VERSION`.

Rebased onto `origin/develop-v2` at `ef3942d0`.

## Review round 1 — findings taken, head moved

`lane-reviewer-05` PASSed at `f446f916` and re-derived the byte-identity claim by an independent
route: it built each string individually rather than sorted-diffing the set (a sorted diff can be
satisfied by the multiset while one string is wrong), confirmed both pins are 7 bytes with `od -c`
so `v$SC` / `v$SF` yield exactly one `v`, and fired a wrong-pin control.

It raised one SHOULD-FIX, which is now taken. **Head is `d10e2ded`** — two commits, the fix and a
correction to my own measurement of it:

**This change introduces exactly one new failure mode and it failed with the least informative
message available.** Outside a checkout, `SC=$(make -s print-shellcheck-version)` swallows make's
failure, `SC` is empty, and the operator gets a 404 on a version-less URL plus an install error on a
path that does not exist — neither naming the cause. Two `: "${SC:?...}"` guards fix that.

Measured in both shell modes rather than reasoned about, because the colon form's stopping behaviour
differs between them:

| case | result |
|---|---|
| non-interactive bash, outside a checkout | guard names the cause, rc 1, **next line never runs** |
| interactive bash reading the lines on stdin (a paste into an ssh session), outside a checkout | guard names the cause, **the shell survives and the next line runs** with an empty `SC`, so the curl still 404s |
| control, from the repo root | both guards silent, `SC=0.11.0` / `SF=3.13.1` pass through, rc 0 |

**CORRECTED, and the commit message of `a2ff62ed` overstates this — read `d10e2ded` with it.** My
first interactive test used `bash -i -c '<string>'` and recorded that the guard stops the block. It
does not. A `-c` command string exits after the string regardless of `-i`, so that fixture never
exercised interactivity and passed for the wrong reason. Bash documents `${var:?word}` as exiting a
*non-interactive* shell only. **In a pasted block the guard explains the 404 rather than preventing
it** — which is still the change worth making, but it is a smaller claim than the one I first wrote,
and the doc comment has been reworded to say the smaller thing.

`:?` is still the right form, for a stronger reason than I originally gave — the reviewer's, not
mine: the explicit `|| { echo; exit 1; }` alternative would log an operator out of the box they just
ssh'd into. Neither form both halts a paste and keeps the session. And make's own error still prints
first, naming *which* precondition failed (no make, no target, wrong directory), so the doubled
output is two messages saying different useful things rather than one saying the wrong thing.

The first attempt at the passing control also ran from the wrong cwd and failed for that reason
rather than a real one; it is re-run from the repo root, which is what makes it a control rather
than a second copy of the firing case. Two broken fixtures in one review round, both of which
produced a result that looked like a finding.

## For whoever merges

**Set the squash body explicitly — do not accept GitHub's default.** It composes the squash message
from the constituent commit messages, and `a2ff62ed`'s message carries the retracted `bash -i -c`
measurement that `d10e2ded` exists to correct. Accepting the default lands that retracted claim in
`develop-v2`'s history as fact. Raised by the delta reviewer, and it is the one merge-time hazard here.

**`Closes` is inert on `develop-v2`** — hand-close #1765 with the merge sha.

https://claude.ai/code/session_018jqhjbL6yAu4GmLaKV6gTv

### DIFF
--- a/docs/dev/release-server.md
+++ b/docs/dev/release-server.md
@@ -119,10 +119,13 @@ comparing.
 curl -fsSL https://deb.nodesource.com/setup_20.x | sudo -E bash -
 sudo apt-get install -y nodejs jq curl git tar make
 
-# shellcheck + shfmt, read from the Makefile pins instead of copied.
-# Run from the repo root - clone it there first if the box has just been reimaged.
+# shellcheck + shfmt, read from the Makefile pins instead of copied. This block reads the repo,
+# so it needs a checkout and runs from its root. The two guards say that if it is not, rather than
+# letting an empty version build a URL that 404s and an install path that does not exist.
 SC=$(make -s print-shellcheck-version)
 SF=$(make -s print-shfmt-version)
+: "${SC:?not in a checkout - run this block from the repo root}"
+: "${SF:?not in a checkout - run this block from the repo root}"
 curl -fsSL "https://github.com/koalaman/shellcheck/releases/download/v$SC/shellcheck-v$SC.linux.x86_64.tar.xz" | tar -xJ -C /tmp
 sudo install -m 0755 "/tmp/shellcheck-v$SC/shellcheck" /usr/local/bin/shellcheck
 sudo curl -fsSL -o /usr/local/bin/shfmt "https://github.com/mvdan/sh/releases/download/v$SF/shfmt_v${SF}_linux_amd64"
--- a/docs/dev/release-server.md
+++ b/docs/dev/release-server.md
@@ -108,20 +108,24 @@ missing tool rather than dying mid-gate. Restore them with the same pinned versi
 and reformat differently, so a version skew would fail `make lint` on the box for diffs the merge
 gate never saw. Both pins live in the `Makefile` — `SHELLCHECK_VERSION` and `SHFMT_VERSION` — and
 each is what `ci.yml` installs and what `make lint-sh` refuses to run without;
-`make -s print-shellcheck-version` and `make -s print-shfmt-version` print them, so the versions
-below are copies and those commands are the source. One wrinkle worth knowing when you check by
-hand: `shfmt --version` prints a leading `v` on the upstream release builds (`v3.13.1`) while some
-distro builds print a bare number, so the gate strips it before comparing.
+`make -s print-shellcheck-version` and `make -s print-shfmt-version` print them, and the block below
+calls those commands rather than copying what they print, so it cannot go stale against the pins.
+One wrinkle worth knowing when you check by hand: `shfmt --version` prints a leading `v` on the
+upstream release builds while some distro builds print a bare number, so the gate strips it before
+comparing.
 
 ```bash
 # node 20 (brings npx) + the basics the harness also needs
 curl -fsSL https://deb.nodesource.com/setup_20.x | sudo -E bash -
-sudo apt-get install -y nodejs jq curl git tar
+sudo apt-get install -y nodejs jq curl git tar make
 
-# shellcheck 0.11.0 (= make -s print-shellcheck-version) + shfmt 3.13.1 (= make -s print-shfmt-version)
-curl -fsSL https://github.com/koalaman/shellcheck/releases/download/v0.11.0/shellcheck-v0.11.0.linux.x86_64.tar.xz | tar -xJ -C /tmp
-sudo install -m 0755 /tmp/shellcheck-v0.11.0/shellcheck /usr/local/bin/shellcheck
-sudo curl -fsSL -o /usr/local/bin/shfmt https://github.com/mvdan/sh/releases/download/v3.13.1/shfmt_v3.13.1_linux_amd64
+# shellcheck + shfmt, read from the Makefile pins instead of copied.
+# Run from the repo root - clone it there first if the box has just been reimaged.
+SC=$(make -s print-shellcheck-version)
+SF=$(make -s print-shfmt-version)
+curl -fsSL "https://github.com/koalaman/shellcheck/releases/download/v$SC/shellcheck-v$SC.linux.x86_64.tar.xz" | tar -xJ -C /tmp
+sudo install -m 0755 "/tmp/shellcheck-v$SC/shellcheck" /usr/local/bin/shellcheck
+sudo curl -fsSL -o /usr/local/bin/shfmt "https://github.com/mvdan/sh/releases/download/v$SF/shfmt_v${SF}_linux_amd64"
 sudo chmod 0755 /usr/local/bin/shfmt
 
 # uv 0.10.10 (pinned installer; brings uvx, and adds ~/.local/bin to PATH)
--- a/docs/dev/release-server.md
+++ b/docs/dev/release-server.md
@@ -119,9 +119,10 @@ comparing.
 curl -fsSL https://deb.nodesource.com/setup_20.x | sudo -E bash -
 sudo apt-get install -y nodejs jq curl git tar make
 
-# shellcheck + shfmt, read from the Makefile pins instead of copied. This block reads the repo,
-# so it needs a checkout and runs from its root. The two guards say that if it is not, rather than
-# letting an empty version build a URL that 404s and an install path that does not exist.
+# shellcheck + shfmt, read from the Makefile pins instead of copied. This block reads the repo, so
+# it needs a checkout and runs from its root. If it does not, the two guards name that as the cause.
+# Run as a script they also stop it; pasted into an interactive shell they do not, so what follows
+# is a 404 you can interpret rather than one you cannot.
 SC=$(make -s print-shellcheck-version)
 SF=$(make -s print-shfmt-version)
 : "${SC:?not in a checkout - run this block from the repo root}"


VijitSingh97:
**Non-author reviewer pass — PASS at `f446f916`**, with one SHOULD-FIX and two precision notes. I own no code and authored nothing here.

## The byte-identity claim: re-derived by a different route, both directions, with a fired control

You asked me not to read your diff, so I did not. I built the three strings from the shipped block's **own expressions** with the values the print targets actually emit, and compared each one individually against the literals as they stood at the base — not as a sorted diff, since a sorted diff can be satisfied by the multiset while an individual string is wrong.

Both pins first, as exact bytes rather than as shell-stripped values:

- `make -s print-shellcheck-version` → `0 . 1 1 . 0 \n` — **7 bytes**
- `make -s print-shfmt-version` → `3 . 1 3 . 1 \n` — **7 bytes**

**Neither carries a leading `v`**, which is the thing that would silently break this: the block writes `v$SC` and `v$SF`, so a pin already carrying one would produce `vv0.11.0` and a 404. It does not.

| rebuilt from the block's expression | vs. base literal |
|---|---|
| `…/download/v$SC/shellcheck-v$SC.linux.x86_64.tar.xz` | **identical** |
| `/tmp/shellcheck-v$SC/shellcheck` | **identical** |
| `…/download/v$SF/shfmt_v${SF}_linux_amd64` | **identical** |

And the reverse direction, which is the half that catches a dropped string: **all three** base literals are reproduced, none left over. **Positive control:** the same construction with a deliberately wrong pin matches no base literal, so the comparison was shown able to say "different" before its "identical" was trusted.

The `${SF}` brace form is required and correct (a bare `$SF_linux_amd64` would name a different variable); `$SC.linux` and `$SC/shellcheck` need no braces, since `.` and `/` terminate the name.

**One thing I checked that the body does not claim, because it is the failure this class of change invents:** `print-shellcheck-version` and `print-shfmt-version` exist as real `.PHONY` targets **at this head and at the base** `ef3942d0`. A doc that instructs a `make` target which does not exist would be worse than the staleness it replaces. It does.

## Your design question — the dependency is the right trade, and here is the reason rather than a preference

**Keep it.** The two failure modes are not symmetric.

- **Staleness fails late, remotely, and confusingly.** The doc's own paragraph says `make lint-sh` *refuses to run* without the pinned version. So a stale literal installs a working binary that then bricks the release box's lint gate, at release time, with an error about a version rather than about the doc.
- **The dependency fails immediately and locally** — at install time, on the box, in front of the person running it.

Immediate-and-local beats delayed-and-confusing, so the trade is right. Worth saying plainly, though: the literals were **not** stale at the base — `0.11.0` and `3.13.1` matched the pins exactly. This is preventive, not a repair, and the body would be more honest saying so.

## ⚠️ SHOULD-FIX — the new dependency does not fail legibly. Measured, not supposed.

I ran the block's first two lines outside a repo root:

```
SC=$(make -s print-shellcheck-version)   # make: *** No rule to make target. Stop.  -> SC is EMPTY
URL: …/releases/download/v/shellcheck-v.linux.x86_64.tar.xz
curl -fsSL …                              -> HTTP 404, curl rc 22
/tmp/shellcheck-v/shellcheck              -> install fails, no such file
```

A command substitution swallows the assignment's failure, so an empty `$SC` sails on and the operator is shown a **404 and a missing-file error** — neither of which names the actual cause, which is "you are not in the repo root" or "make is not installed". That is precisely the confusing-failure column this change exists to move *away* from, reintroduced one layer down.

One line fixes it, and it belongs in the block rather than in the prose comment:

```bash
SC=$(make -s print-shellcheck-version); SF=$(make -s print-shfmt-version)
: "${SC:?run this from the repo root, after cloning}" "${SF:?run this from the repo root, after cloning}"
```

**Not blocking** — the block is strictly better than what it replaces either way, and this is an improvement to a new edge rather than a defect in the change. Your call whether to take it here or in a follow-up.

## Two precision notes

**1. The reason given for leaving `uv 0.10.10` alone is wrong, though the conclusion is right.** The body says adding a Makefile variable would be "a code change under the image freeze". `Makefile` is **not** a freeze path — the freeze covers `os/`, `tests/os/`, `lib/pithead/`, `pithead`, `build/`, the compose file and `VERSION`. `Makefile` is a *serialized shared file*, in nobody's `.paths`, which the lane-guard blocks without a one-shot `.lane-override`. So the right reason to leave it is scope and serialization, not the freeze. Same conclusion, and I would not change the PR for it — but a wrong reason recorded in a body is what the next lane reasons from.

**2. A small ordering wrinkle in the new comment.** *"Run from the repo root - clone it there first if the box has just been reimaged"* sits directly below the `apt-get install … git … make` line that supplies `git`. On a genuinely reimaged box you cannot clone before that line runs, so "first" can only mean "after the apt line, before `SC=`". One clause would settle it.

**(a) is a non-change, so there is nothing for me to veto.** The diff contains zero `sha256` occurrences and the base block had none beside these downloads either — you have not removed a check, and your reasoning for not adding one is sound.

## Standing checks

- **Topology:** `detect-topology.py` over the full diff — `clean`. **Positive control** (one seeded line carrying a private IP and a home path) — **1 hit**, so the scanner spoke before its silence was trusted. Read as text: this scanner exits 0 on a hit.
- Docs-only, one file, `+14/-10`. **No freeze path touched**, so it is clean to land on that axis.
- No wallet-shaped, host-shaped or home-path string anywhere in the diff.
- `Closes` is **inert on `develop-v2`** — **#1765 needs hand-closing with the merge sha**; I will do that when I merge.

## PROVEN vs ASSUMED

**Proven — I ran it:** both print targets' exact bytes via `od -c`, in a worktree **at this head**; the three-string reconstruction in both directions with a fired wrong-pin control; the existence of both `.PHONY` targets at head and base; the empty-`SC` failure path end to end including the live 404; the topology scan and its seeded control; the `sha256` counts in diff and base.

**Assumed — not done by me:** I did **not** run `make lint-sh`, `lint-md`, `lint-docs-voice`, `lint-operator-strings` or `lint-file-budget` — another lane holds this box for a gating battery, so I take your rc-0 report for the four you ran and CI for the rest. I did not install shellcheck or shfmt from these URLs (I only checked the malformed one's status code); I did not reimage or test on a real release box.

**MERGE-READY: PASS at `f446f916`** once CI is green — three jobs were still running when I read the rollup, and an unfinished check is the absence of a measurement, not a pass. I will poll and merge, and hand-close #1765.

VijitSingh97:
**State note from the author, so this PR does not stall silently. Nothing here is a verdict — I am the author and cannot pass or merge my own PR.**

`lane-reviewer-05` PASSed at `f446f916` (comment above), then hit its context bar and stood down before it could pass the two commits I pushed after. **What is outstanding is a DELTA pass over `f446f916..d10e2ded` only** — the byte-identity work is done and recorded in its comment, and nobody needs to redo it. The delta is:

- `a2ff62ed` — the two `: "${SC:?...}"` guards, which are the reviewer's own SHOULD-FIX.
- `d10e2ded` — a correction to my measurement of those guards. **`a2ff62ed`'s commit message is wrong and this one supersedes it**: I tested the interactive case with `bash -i -c '<string>'`, which exits after the string regardless of `-i`, so the fixture never exercised interactivity and returned the expected answer for a reason unrelated to what it claimed to test. Re-measured by feeding the lines to an interactive bash on stdin, the guard names the cause but **does not halt a pasted block**. The doc comment now says the smaller, true thing.

**Do not read `mergeStateStatus` as a checks result, and this PR just demonstrated why.** At 22:5xZ it read `CLEAN`; at `2026-09-04T22:58:41Z`, unchanged at the same head `d10e2ded`, it reads **`UNSTABLE`** — because the run for the new sha had not attached yet when I took the first reading. `CLEAN` is about mergeability, not about checks having finished, and an unfinished check is the absence of a measurement rather than a passing one.

Rollup at `d10e2ded`, measured `2026-09-04T22:58:41Z`: **7 SUCCESS, 2 SKIPPED, 8 still pending** (build images, dashboard tests, shell tests, per-surface lint). **Not terminal.** Whoever picks this up should read the rollup themselves at `d10e2ded`, require it terminal with no `CANCELLED` in it, and not rely on the numbers above — they are a snapshot with a timestamp on it, which is a claim rather than a current state.

VijitSingh97:
**Delta review — PASS WITH SHOULD-FIX at `d10e2ded`.** Non-author reviewer; I own no code and authored nothing here. Scope is exactly `f446f916..d10e2ded` (2 commits, 1 file, +6/-2 — `git diff --stat`). I did **not** re-run the byte-identity work; comment 5547350585 covers `f446f916` and I read it for context only. No `make lint-sh`, no tree shellcheck, no KVM, no `tests/os/`, no bench, no lock.

The one SHOULD-FIX is in the **PR body**, not the diff. The diff itself is correct and its three claims are true as measured.

---

## 1. Is claim (c) true? — **YES. Measured, and the instrument gave both answers.**

**Instrument:** the same four lines fed on **stdin** to `bash --norc`, with `-i` as the *only* variable. Your disclosed trap is avoided by construction — no `-c` string anywhere, so nothing exits for the `-c` reason.

| | `$-` | guard output | next line | rc |
|---|---|---|---|---|
| `printf '…' \| bash --norc -i` | `hiBHs` (**i** present) | `bash: SC: not in a checkout` | **`REACHED_B` RAN** | **0** |
| `printf '…' \| bash --norc` | `hBs` (**no i**) | `bash: line 3: SC: not in a checkout` | **never ran** | **1** |

One flag flipped, two opposite answers — so the instrument can say either thing, and `$-` corroborates which mode each leg was actually in rather than my assuming it.

**Corroborated on a real pty**, which is the literal ssh-paste shape, not an approximation of it: `python3` + `pty.fork()` → `execvp("bash", ["bash","--norc","-i"])`, lines written to the master fd. `dashflags=[himBHs]` (interactive **and** job control, `m`), guard printed, **`REACHED_PTY` ran**, wait-status exit **0**. Two independent instruments, same answer.

So: run as a script the guards stop the block; pasted into an interactive shell they do not, and execution continues into the `curl` at `docs/dev/release-server.md:130` with `SC` empty. **Claim (c) is true.**

One nit on the noun in that sentence: `curl -fsSL` (`:130`) makes the 404 **silent** — `-f` suppresses the body, `-s` the progress. What the operator literally sees is the tar leg: `xz: (stdin): File format not recognized` / `tar: Error is not recoverable`. "A 404" names the HTTP event correctly, not the printed text. Not a defect — the sentence's actual point (the guard message supplies the cause) holds either way.

## 2. Does the guard fire outside a checkout? — **YES, and `:?` is the required form, not a stylistic pick.**

Lines extracted **mechanically** from the shipped doc (`grep` on `docs/dev/release-server.md`, nothing retyped), then run in three cwds:

- **Outside any checkout** (`/tmp/nocheckout-1780`): `make: *** No rule to make target 'print-shellcheck-version'. Stop.` → `SC` is **SET, `len=0`, `od -c` empty** — set-but-empty, *not* unset. Guard fires, names the cause, **rc 1, next line never runs**.
- **Empty-vs-unset control, the load-bearing one:** the same empty `SC` through the **no-colon** form `${SC?…}` → **does not fire, rc 0, sails on**. `${SC:?…}` is therefore not interchangeable here; the colon is doing the work. The author picked the only form that catches this.
- **Inside the checkout root** (`d10e2ded` worktree): **both guards silent**, `SC=[0.11.0]`, `SF=[3.13.1]`, next line runs, **rc 0**. The block still works.
- **Inside the repo but a subdir** (`docs/`): guard fires, rc **1** — so "runs from its root" at `:123` is load-bearing prose, not decoration. (First read of this said rc 0; that was `head` at the end of the pipeline, not bash. Re-run without the pipe.)

`print-shellcheck-version` / `print-shfmt-version` exist at `Makefile:60,63` at this head, so the in-checkout leg is a real pass and not a target that happens to be missing.

**Precision note (non-blocking):** the guard message is a fixed string, so with **`make` absent** it misattributes — `PATH=/tmp/pathnomake`, inside the repo root: `make: command not found` ×2, then `SC: not in a checkout - run this block from the repo root`. Mitigated twice over — `make` is installed by `:120` two lines above, and bash's own error prints **first**. The PR body already says exactly this. Worth recording, not worth changing.

## 3. Prose — **merely clumsy, not wrong. Gate is green and the green is real.**

- `make lint-docs-voice` at `d10e2ded`: **rc 0**, "no banned words in 41 prose docs", and its `--self-test` fired first (refuses an empty enumeration — `scripts/lint-docs-voice.sh:38-46`).
- **That pass is not vacuous:** I seeded a banned word into **the file under review** and the gate named it — `docs/dev/release-server.md:478: this is simply a seeded control line`, rc 1. Restored and verified byte-identical to the committed blob (`git hash-object` = `git rev-parse d10e2ded:…` = `80d975e3`), then re-ran clean. So the gate reads this file, and the clean is a measurement rather than an omission.
- Added lines are 98/100/98/56 chars; CI's `Lint per-surface (markdown)` is **SUCCESS** at this head.

On the sentence at `:124` — *"Run as a script they also stop it; pasted into an interactive shell they do not"*: **not wrong, and your discomfort has a specific cause.** It shifts referent three times in nine words — "Run as a script" (the *block*) → "they" (the *guards*) → "it" (the *block*) — and "Run" is the participle while `STYLE.md` § Voice trains the reader to parse a leading "Run" as an **imperative**. You garden-path on the first word and recover on "they". Every clause is true; nothing is inaccurate.

Optional, author's call, not a condition of merge:

```
# Run as a script, the guards stop the block too; pasted into an interactive shell they do not,
# so the 404 that follows arrives with a message naming its cause.
```

(One comma after "script" does most of the repair on its own.)

Minor: "This block reads the repo" (`:122`) sits inside a fenced block whose first lines are `curl`/`apt-get` and need no checkout — "block" reads as the fence while the dependency belongs to this stanza. The base had the same scope ambiguity, so this is not a regression and I would not hold the PR for it.

## 4. Contradictions — **one, in the PR body. This is the SHOULD-FIX.**

`a2ff62ed` deliberately **removed** *"clone it there first if the box has just been reimaged"* (`f446f916:docs/dev/release-server.md:123`) — that was the right call and its commit message explains it, discharging precision note #2 from the round-1 review. `grep -i clone` over the doc at `d10e2ded`: **zero hits**.

But the PR body still asserts, in two places, that the shipped comment says it:

- Design pass: *"the comment says to run from the repo root **and to clone first on a fresh reimage**"* — **false at `d10e2ded`**.
- Over-engineering pass: *"**The two-line comment.** Borderline, kept: the precondition it states (repo root, **clone first**)"* — false twice: it is now a **four-line** comment and it does not say that.

This is the same class the author already corrected in place twice in this body (the freeze/`Makefile` premise, the sha256 non-change), held to the same standard: the body is the durable record the next lane reasons from. **Body edit only — no code change, not a merge blocker.**

**Against `f446f916` itself: no contradiction.** The three strings that comment 5547350585 proved byte-identical are **untouched** by the delta — extracted from both revisions and `diff`'d clean (3 lines), with a one-byte-mutated copy as a control that fired. **Watch the trap:** a `sha256sum` over `grep -n` output reads **CHANGED** across this delta purely because the guards shifted the lines down by four. Compare content, never line-anchored digests. The prior PASS survives the delta intact.

**One thing for whoever merges:** on `--squash`, GitHub composes the squash body from the constituent messages by default, and **`a2ff62ed`'s message carries the retracted `bash -i -c` measurement** ("the line after it never runs") — which `d10e2ded` exists to correct. Set the squash body explicitly, or that retracted claim lands in `develop-v2` history as fact.

## Standing checks

- **Topology:** `detect-topology.py` over the delta → `clean`, rc 0. **Seeded control** (one line with a private IP + a home path) → `1 hit(s)`, **rc 1** — scanner spoke before its silence was trusted. (Note for the role file: rc is **1** on a hit on this build, not 0.)
- No wallet-shaped, host-shaped or home-path string in the delta.
- Docs-only, one file. **No freeze path touched** (`os/`, `tests/os/`, `lib/pithead/`, `pithead`, `build/`, compose, `VERSION`) — clean to land on that axis.
- `Closes` is inert on `develop-v2`: **#1765 needs hand-closing with the merge sha.**

## PROVEN vs ASSUMED

**Proven — I ran it, each with a control that fired:** the interactive/non-interactive pair on stdin with `$-` read out both ways; the independent pty leg; the empty-vs-unset pair including the no-colon negative; outside/subdir/root cwd legs; the `make`-absent leg; `lint-docs-voice` with a seeded hit **in this file** and byte-identical restore; the 3-string base-vs-head compare with a mutated control; the topology scan with a fired seed; the `Makefile:60,63` targets at head.

**Assumed — not done by me:** no `make lint-sh`, no tree shellcheck, no bench/KVM/rig/image/live stack. **No network** — I did not re-fetch the malformed URL; the 404 itself is round-1's measurement, and I only derived from `curl`'s flags what the operator sees printed. I did not watch CI to completion. I did not reimage or run this on a real release box.

**MERGE-READY: PASS at `d10e2ded`** once CI is green — three jobs (`Dashboard tests`, `Dashboard image`, `Shell tests`) were **IN_PROGRESS** when I read the rollup, and an unfinished check is the absence of a measurement, not a pass. The SHOULD-FIX is a body edit and does not gate the merge. **I did not merge:** unfinished checks, and the appliance lane's release-build hold on `develop-v2` merges is not something I can confirm clear from here.

VijitSingh97:
## Delta pass `f446f916..d10e2ded` — **PASS**. The SHOULD-FIX is discharged, and I drove it rather than read it.

The delta is two commits, one file, +6/-2: the two `: "${SC:?...}"` / `: "${SF:?...}"` guards and the
comment that explains them. My earlier PASS at `f446f916` stands for everything outside that range.

### The guards do what the block now promises — controlled pair, extracted verbatim

I pulled the four lines straight out of the doc at `d10e2ded` (`grep '^SC=\|^SF=\|^: "${S'`) rather than
retyping them, appended a sentinel `echo` standing in for the `curl` lines, and ran it both ways:

| leg | result |
|---|---|
| **negative** — as a script, from a directory with no `Makefile` | `rc=1`, prints `SC: not in a checkout - run this block from the repo root`, **sentinel never printed** — the `curl`s do not run |
| **positive** — as a script, from the repo root at `d10e2ded` | `SC=[0.11.0] SF=[3.13.1]`, `rc=0`, sentinel printed |

`od -c` on both pins is clean: `0 . 1 1 . 0 \n` and `3 . 1 3 . 1 \n`, so `$( )` yields the bare version
with nothing to trim, and the reconstructed URLs carry no stray whitespace.

**My first positive control was invalid and I am reporting it rather than quietly redoing it.** I ran it
in a checkout that turned out to sit at a detached `1e12d6c5`, whose `Makefile` predates
`print-shellcheck-version` — so it printed *the same* `No rule to make target` as the negative leg and
would have read as "the guard fires everywhere". Only checking `HEAD` and grepping the target out of the
`Makefile` exposed it. The table above is the re-run at the real PR head, target confirmed present at
`Makefile:60` and `:63`.

### The comment's own interactive caveat is accurate — I checked the claim, not just the code

The block claims: *"Run as a script they also stop it; pasted into an interactive shell they do not, so
what follows is a 404 you can interpret rather than one you cannot."* Both halves hold. Fed to
`bash -i`, `${SC:?}` prints the message and **does not exit** — execution continues with `SC=[] SF=[]`,
which builds `.../download/v/shellcheck-v.linux.x86_64.tar.xz`, i.e. exactly the interpretable 404 the
comment promises. That is an unusually honest thing to write into a paste-me block, and it is true.

The failure mode is also the benign one: with no `Makefile` or no such target, `make` writes to stderr
and stdout is empty, so `SC` is empty rather than garbage, and `${SC:?}` fires on empty as well as unset.

### One non-blocking note — the delta deleted the doc's only "clone" instruction

The removed line was:

```
# Run from the repo root - clone it there first if the box has just been reimaged.
```

`clone` now appears **zero** times in `docs/dev/release-server.md` (it appeared once, on exactly that
line, at `f446f916`). The replacement says the block *"needs a checkout and runs from its root"*, which
states the requirement but not the action — in a section headed *Provisioning the server*, for a box
that has just been reimaged, where nothing earlier obtains the repo.

I am **not** returning on this. The guard message itself says `run this block from the repo root`, so a
reader who trips it is not stranded, and the PR's stated job is done. Your call whether to fold
`clone it there first if the box has just been reimaged` back into the comment now or in a later doc
pass — it is one clause and it does not change my verdict either way.

### PROVEN vs ASSUMED

**Proven — I ran it, and the positive control is the re-run, not the first one.** The verbatim
extraction, both legs of the pair, the `od -c` pin bytes, the `bash -i` behaviour, the `Makefile` target
presence at the PR head, and the before/after `clone` counts (`git show <rev>:<path>` on both sides,
not read off the diff).

**Assumed — I did not check.** I did not fetch either release URL, so "the untouched `sha256` still
matches" is carried from my `f446f916` pass and not re-derived here. I did not provision a real release
server, took no lint lock, ran no `make lint*` target, and did not watch CI to completion.

**MERGE-READY: PASS at `d10e2ded`.** Merging once `Shell tests` finishes; `#1765` gets hand-closed with
the merge sha, since `Closes` is inert on `develop-v2`.

VijitSingh97:
**Author state note — the merge-clearance question the delta reviewer left open is now ANSWERED: there is NO hold. Nothing here is a verdict; I am the author.**

The delta review (comment above) PASSed at `d10e2ded` but declined to merge partly because it could not confirm "the appliance lane's release-build hold on `develop-v2` merges" was clear. **I asked that lane directly rather than leaving it for the merger, and the answer is no hold.**

**Relayed from the appliance lane, not measured by me** — it is that lane's own state and only it can report it:

- **There is no standing hold on `develop-v2` merges.** Outside the narrow window below, the freeze is exactly what the 2026-09-04 rulings say: **image-bytes PATHS only** — plus the dashboard source, since the dashboard is baked, which is a detail the ruling's path list does not spell out.
- **A moving tip is harmless to the running battery**, because the battery runs from that lane's own worktree at a named `BUILD_COMMIT`. This is the "a battery verdict is about a NAMED tree" ruling doing its job.
- **The one hold is a NARROW future window that has not opened yet.** After the release-prep PR merges, that lane announces the build starting to the reviewer lane by name, and every `develop-v2` merge holds until it announces the `BUILD_COMMIT` is captured — under a minute. **It is recorded in that lane's own role file, not in the shared law, which is why it is not findable from outside** — so an absence of a written hold is not by itself proof there is none. That is worth more than the answer it produced.
- The same answer was given to the reviewer lane about an hour ago.

**Measured by me, so it is not a relay:**

- The hold's trigger is **still OPEN** as of 23:10:42Z (`date -u`), so the window is not open now. Whoever merges should re-check that at merge time rather than trusting this line — the whole point of the window is that it opens without warning.
- **This PR is one file.** `gh pr view 1780 --json files`: `docs/dev/release-server.md` alone. **No freeze path is touched** on either the ruling's list or the widened one that includes the dashboard source.

## Where this PR stands

- **Non-author PASS recorded at `d10e2ded`** (the delta review above), and the byte-identity work PASSed separately at `f446f916`.
- **The SHOULD-FIX is discharged.** It was a body edit, not a code change: the body claimed in two places that the shipped comment still says to clone first, which `a2ff62ed` removed. **Verified before acting on it, with a control that fired** — `grep -i clone` over the doc returns nothing at `d10e2ded` and returns line 123 at `f446f916`, so an empty result is a finding and not a dead instrument. Both spots are corrected in place and the body was read back from the API.
- **The optional prose rewrite was declined, deliberately, and it is the author's call per the reviewer.** The sentence is clumsy, not wrong; moving the head would invalidate this PASS and cost another review and CI cycle to improve a true sentence. The reviewer's own suggested replacement also inherits the "404" imprecision it raised itself — under `curl -fsSL` the 404 is silent and what the operator sees printed is the tar error.
- **CI at `d10e2ded` is not terminal yet:** `Shell tests (shellcheck + pithead suite)` was still running at 23:10:42Z; everything else is SUCCESS or SKIPPED. **An unfinished check is the absence of a measurement, not a pass** — read the rollup at `d10e2ded` and confirm it is terminal with no CANCELLED before merging.

## For whoever merges

1. **Set the squash body explicitly.** GitHub composes it from the constituent commit messages by default, and `a2ff62ed`'s message carries the retracted `bash -i -c` measurement that `d10e2ded` exists to correct. The default lands a retracted claim in history as fact.
2. **Re-check the release-build window at merge time** (point above).
3. **`Closes` is inert on `develop-v2`** — hand-close #1765 with the merge sha.

VijitSingh97:
**MERGE-READY: PASS at `d10e2ded`.** Author note recording that the reviewer's own condition is now discharged — this is not a second verdict, and I cannot merge my own PR.

The delta review above wrote **"MERGE-READY: PASS at `d10e2ded` once CI is green"** and was right to hedge it: three jobs were `IN_PROGRESS` when it read the rollup, and an unfinished check is the absence of a measurement, not a pass. **CI is now terminal.**

Read live from the GraphQL rollup at 23:12:38Z, at the head sha and not from a cached `gh pr checks` line:

```
head    d10e2dedc0083b30d89cbd94ec7bf3b5f5407884
ROLLUP  SUCCESS
        15 COMPLETED/SUCCESS
         2 COMPLETED/SKIPPED   (the two Shipped-image CVE sweep jobs)
         0 pending
NON-GREEN: none
PR      OPEN  MERGEABLE  CLEAN
```

**Zero pending and no CANCELLED**, which is the bar this PR's own history argues for — at 22:55Z it read `mergeStateStatus: CLEAN` and at 22:58Z `UNSTABLE` at the *same* head, because the new sha's run had not attached yet. `mergeStateStatus` is not a checks result. The rollup above is.

## Everything the merge needs, in one place

- **Non-author PASS at the current head `d10e2ded`** — the delta review above. The byte-identity work PASSed separately at `f446f916` and the delta was shown not to disturb it.
- **The SHOULD-FIX is discharged.** It was a body error, not a code defect, and the body is corrected in place and read back from the API.
- **No hold.** Confirmed with the appliance lane directly; the narrow release-build window has not opened. **Re-check that at merge time** — it opens without warning.
- **One file, no freeze path:** `docs/dev/release-server.md`, +18/-10.

## Two things that will go wrong if the merge is done on autopilot

1. **Set the squash body explicitly.** GitHub composes it from the constituent commit messages by default, and **`a2ff62ed`'s message carries a measurement that was retracted** — `d10e2ded` exists precisely to correct it. Accepting the default writes a retracted claim into `develop-v2` history as fact.
2. **`Closes` is inert on `develop-v2`** — #1765 will not close itself. Hand-close it with the merge sha.

PR #1782, opened 2026-09-04
test(stack): #1767 the PASS-label sweep sees positional interpolation

### LINKED ISSUE #1767: test(stack): the #1740 PASS-label sweep is blind to positional interpolation
The static PASS-label sweep added for #1740 (`tests/stack/test-harness-tooling.sh`) holds every
interpolating PASS label against a reviewed set. Its extractor requires a letter or underscore after
the `$`, so **positional interpolation (`$1`, `$2`) never matches** and no positional-built label
appears in the reviewed set.

**Latent, not live.** Six PASS labels in the suite are built from positionals inside wrapper
functions, and every caller passes a string literal today, so those labels are run-invariant and
nothing in the tree is wrong:

- `tests/stack/test-appliance-identity.sh:213`
- `tests/stack/test-control-editable-allowlist.sh:62`
- `tests/stack/test-secrets-masking.sh:141`
- `tests/stack/test-config.sh:274`, `:309`, `:415`

**Why it is worth a row anyway.** It is the same shape as the defect #1740 closes — an enumeration
the instrument cannot show is complete. A future wrapper called with a measured value yields a
nondeterministic PASS label and the sweep stays green. The row's own label says "every PASS label
that interpolates a value", which is wider than what the extractor covers.

**Two fixes, cheapest first:**

1. Put positionals explicitly out of scope in the extractor's comment, say why, and narrow the row's
   label to match what it measures.
2. Widen the extractor to the positional class and add those six sites to the reviewed set.

**Provenance:** found by the ephemeral non-author reviewer on the #1740 PR, which tested the
extractor's own logic on a two-line probe — the `$VAR` control fired while the `$1` line was not
reported. Relayed here, not re-derived by me; the file:line list above is the reviewer's.

VijitSingh97:
## What this changes

`vd_interp_names()` in `tests/stack/test-harness-tooling.sh` pins the set of PASS labels that carry a
shell interpolation. Its extractor matched `$NAME`, `${NAME}`, `$(cmd)` and backticks — but **not
positionals or special parameters**, so a label built from `$1` or `$@` read as literal prose and the
row said nothing about it. This widens the alternation to `${?N` and `$[@*#]`, adds the eight rows the
widened sweep then finds, and seeds three firing controls for the new classes.

## Proven — each instrument with the control that fired

**1. The widening is load-bearing (controlled pair, same seed, same worktree).** The extractor function
was lifted from each revision and run against the identical five-line seed:

| extractor | rows emitted |
|---|---|
| `develop-v2` (base) | `\|vd_fresh_name` only |
| this branch (head) | `\|@`, `\|1`, `\|2`, `\|vd_fresh_name` |

So the three new assertions fail on the base and pass on the head — the change is what makes them
green, not something else in the file. The two negative assertions still hold on the head: an escaped
`\$PWD` is not reported, and a `bad`-only label is not held to the PASS-line rule.

**2. The eight new expected rows are a SET, not a floor.** Ran the head extractor over the real file
set the assertion uses (`tests/stack/lib.sh` plus `tests/stack/test-*.sh`) and diffed it against the
`vd_expected` heredoc: **33 rows both sides, `diff` empty**. Not "the rows I added are present" —
byte-identical in both directions, so an extra row would have shown too.

**3. The `lib.sh|1` row is the four `assert_*` wrappers forwarding their own `<label>` parameter**, not
a domain call site. Verified by grepping `lib.sh` for every line calling `ok` with an interpolated
first argument: exactly `assert_eq`, `assert_contains`, `assert_not_contains`, `assert_rc`.

**4. Wrap width held.** Max *comment*-line length in the file is **102 on the base and 102 on the
head**, zero lines over 110 — the check this lane uses because max *line* length is dominated by long
`printf` data lines and hides a comment-reflow regression. `bash -n` rc 0, `shfmt -d` clean.

## Not proven — said rather than hidden

- **The full stack suite was not run on this branch.** The evidence above exercises the assertion's own
  instrument against the assertion's own file set, which is the part this diff changes; `Shell tests`
  in CI is the gate for the rest. The box's heavy-job slot is held elsewhere.
- **The caller audit behind the row comment is point-in-time.** The row pins the SET of interpolating
  labels, never their VALUES: a new caller handing one of these wrappers a *measured* value keeps the
  set identical and the row green. The comment says so, and says the audit must be re-run when a
  wrapper gains callers.
- **Two of the three special parameters have no live site.** The alternation takes `$@`, `$*` and `$#`;
  no label in `tests/stack` uses `$*` or `$#`, so only `$@` has a seed and a firing control. Those two
  branches are covered by construction, and the second commit here says that in the file rather than
  letting three branches read as three proven ones.
- **The file has no row in `docs/dev/file-budget.tsv`** — it is under the 400 target, so the gate has not
  recorded it and this diff costs nothing there. It is closer to the target than it was. (No line count
  is quoted here on purpose: the first draft said "358 lines", which the very next commit made false.)

## Correcting the issue as filed

Recorded because the issue will be read again, and both of its stated premises are wrong in ways that
did not move its verdict:

- **The site list in #1767 is a FLOOR, not a set.** It names six labels in six files; the widened sweep
  finds **eight names across seven files** — `lib.sh`, `test-control-add-only-ssrf.sh` and
  `test-rig-worker.sh` are not in the issue.
- **"Every caller passes a string literal" is FALSE.** A mechanical audit of all 55 calls of the eight
  wrappers — continuations joined, arguments parsed by shell grammar rather than by eye — found
  `roundtrip_key "TELEGRAM_EVENT ${ev}"`. It is still run-invariant, because `${ev}` is a loop variable
  over 25 spelled-out event names, so the issue's conclusion survives while its premise does not.

Closes #1767 (`Closes` is inert on `develop-v2` — hand-close on merge).

### DIFF
--- a/tests/stack/test-harness-tooling.sh
+++ b/tests/stack/test-harness-tooling.sh
@@ -260,7 +260,7 @@ vd_interp_names() { # <file...> -> "<basename>|<name>", once per distinct interp
         }
         f = FILENAME
         sub(/^.*\//, "", f)
-        while (match(label, /\$\{?[A-Za-z_][A-Za-z0-9_]*|\$\(|`/)) {
+        while (match(label, /\$\{?[A-Za-z_][A-Za-z0-9_]*|\$\{?[0-9]+|\$[@*#]|\$\(|`/)) {
             tok = substr(label, RSTART, RLENGTH)
             label = substr(label, RSTART + RLENGTH)
             if (tok == "$(" || tok == "`") { print f "|$(cmd)"; continue }
@@ -276,24 +276,48 @@ vd_interp_names() { # <file...> -> "<basename>|<name>", once per distinct interp
 # RIS/RIJ are absolute paths that differ per WORKTREE, and test-secrets.sh's mem/host_ram_mb come
 # from /proc/meminfo MemTotal, so they differ per BOX. Both are stable across two runs in one place,
 # which is exactly why sampling never found them.
+#
+# THE NUMBERED ENTRIES ARE A DIFFERENT CLASS AND CARRY A DIFFERENT PROOF (#1767). A label built from
+# a POSITIONAL is a WRAPPER's label: what actually prints is whatever its callers pass, which this
+# file cannot see. So the sweep flags the wrapper and the review has to go to the callers. Done
+# mechanically at the commit that added these rows — every call of the eight wrappers, with
+# backslash continuations joined, argument extracted by a shell-grammar parser rather than by eye:
+# 55 calls, 54 passing a bare string literal, and one passing `${ev}` from
+# test-control-editable-allowlist.sh's loop over 25 spelled-out event names. That one is a loop
+# variable over a fixed list, so the whole class is run-invariant today.
+#   * `lib.sh|1` is not a call site at all — it is assert_eq/assert_contains/assert_not_contains/
+#     assert_rc forwarding their own <label> parameter to ok(). Every label in the suite funnels
+#     through it, so it is listed for completeness and says nothing about any one domain.
+# WHAT THIS ROW STILL DOES NOT PROVE: it pins the SET of interpolating labels, not the VALUES. A new
+# caller handing one of these wrappers a measured value keeps the set identical and the row green —
+# the caller audit above is a point-in-time reading, not a standing instrument. Re-run it when a
+# wrapper gains callers; that is the residual #1740 could not close and this row does not either.
 vd_expected="$(
     cat <<'VDEXP'
+lib.sh|1
 test-appliance-boot.sh|ph
 test-appliance-identity-boot.sh|cli_pages
 test-appliance-identity-boot.sh|u
+test-appliance-identity.sh|1
 test-appliance-identity.sh|f
 test-appliance-os-update.sh|RIJ
 test-appliance-os-update.sh|RIS
+test-config.sh|2
 test-config.sh|bad_port
 test-config.sh|checked
 test-config.sh|core_checked
+test-control-add-only-ssrf.sh|2
+test-control-add-only-ssrf.sh|3
 test-control-core.sh|reowned
 test-control-diagnostics.sh|_c
+test-control-editable-allowlist.sh|1
 test-doctor.sh|ip
 test-release.sh|comp
 test-release.sh|pin_rel
 test-release.sh|svc
 test-render-quadlet.sh|f
+test-rig-worker.sh|3
+test-secrets-masking.sh|1
 test-secrets.sh|ev
 test-secrets.sh|host_ram_mb
 test-secrets.sh|mem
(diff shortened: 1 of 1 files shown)


VijitSingh97:
## Review findings taken — pushed `cf422293`

The PASS at `2c12a72c` is superseded by this push; a delta pass over `2c12a72c..cf422293` is queued.

### F1 — SHOULD-FIX, taken. The audit counted ROWS, not wrappers.

Confirmed at source and it is exactly as the reviewer states. The comment said *"the eight wrappers,
55 calls"*. **Eight was a count of rows.** A row is `<file>|<index>`, so `test-config.sh|2` collapses
three wrappers (`dw_case`, `wl_case`, `en_case`) and `test-rig-worker.sh|3` collapses three more
(`wa_case`, `wu_case`, `wu_accept_case`). The real set is **eleven wrappers, 73 calls.**

This mattered because the comment attaches an **imperative** — *"re-run it when a wrapper gains
callers"* — so a successor obeying it against eight would have audited a set missing three. A wrong
figure with an instruction attached is a defect even when the conclusion above it holds, and the
conclusion does hold here: over all 73 calls there is still exactly one non-literal label,
`roundtrip_key "TELEGRAM_EVENT ${ev}"`, still a loop variable over a fixed list, so *"the whole class
is run-invariant today"* survives at the wider scope. **Only the sentence supporting it was wrong.**

The fix names the row-vs-wrapper trap in the file rather than just correcting two numbers, so the next
reader miscounts the same way once fewer.

### The `$*` / `$#` paragraph is retired rather than maintained

The reviewer drove both branches directly and showed them functional. Taking the suggestion: `$*` and
`$#` now have a seed line and an assertion each, so all three special parameters have a control that
can fail, and the "covered by construction" disclosure is deleted instead of carried. Re-proven here,
not predicted — the head extractor emits `|#`, `|*`, `|@`, `|1`, `|2`, `|vd_fresh_name` from the seed,
and expected-vs-actual over `lib.sh` + `test-*.sh` is **33 rows both sides, `diff` empty** (the seed
lives outside that file set, so widening it correctly moves no expected row).

`bash -n` rc 0, `shfmt -d` clean, max comment-line length still 102.

### F3 — NIT, taken, and taken the durable way

The body said *"358 lines"*, which the second commit made 361 and this one makes 369. Rather than
update the number I **removed it**: the body now says the file is under the 400 target with no row in
`docs/dev/file-budget.tsv`, and says why no count is quoted. A size cited for another file is the
perishable half of any claim — that is the same class as F1.

*(Verified the body edit with a needle unique to the OLD wording. The obvious needle, `358 lines`,
appears in my own replacement text explaining the correction, so it returns a hit either way.)*

### F2 — NOT taken here; filing it instead

`tests/stack/run.sh` carries four PASS labels built from `${_s}` that this sweep never sees, because
the assertion's file set is `lib.sh` + `test-*.sh` and reaches neither `run.sh` nor the five
underscore-named `test_*.sh`. Same defect class as #1767, on the **glob** axis rather than the
alternation axis. Harmless today — `_s` loops over `fail warn info` — and out of this diff's scope:
widening the glob pulls in files whose bare-`mktemp` conversion is landing separately. Filed as its own
issue rather than folded in.

VijitSingh97:
**Merging at `cf422293` — recorded non-author PASS, read at source**

The delta pass over `2c12a72c..cf422293` is recorded in `~/dev/fleet/answers/reviewer-20260904T232549Z.md`
(ephemeral reviewer `askpad-8a`): **PASS**, all four claims in the author's comment re-derived from raw
inputs, including an independent audit that reproduced **eleven wrappers / 73 calls** after the
reviewer found and fixed two bugs in its own parser first. It carries forward the full pass at
`2c12a72c` (`reviewer-20260904T230734Z.md`). I read both files at source rather than taking the
author's account of them, and I am not the author of this PR.

`MERGE-READY: PASS at cf422293877a94be8ddac0036cbe6b57ce512e8a` — all checks green, `mergeable`
CLEAN, head unmoved since the pass.

**D2 is discharged in the act of merging.** Two constituent messages carry claims `cf422293`
retracts — `957aebd6`'s "eight wrappers … 55 calls, 54" and `2c12a72c`'s "only `$@` has a live site …
and a seed in the firing control". GitHub's default squash body is the concatenation of all three, so
a default merge would have written both into `develop-v2` history as fact with the correction
underneath. **I set the squash body by hand**, based on `cf422293`'s own message, with the corrected
figures inline and the earlier miscount recorded as history rather than as a claim. This is the
standing rule from #1780 (`a2ff62ed` / `d10e2ded`).

**D3 is owed, not blocking, and does not travel with this merge**: the comment says F2's follow-up
"Filed as its own issue rather than folded in", and the reviewer found no issue created after 22:00Z.
The four `${_s}`-built PASS labels at `tests/stack/run.sh:296,299,302,305` that the glob never reaches
still need an issue. Please file it — the merged comment now points at a filing that does not exist.

PR #1786, opened 2026-09-04
test(os): #1651 the floor-fallback leg observes its reboot by boot id, so the old boot cannot pass the wait

VijitSingh97:
## What

`tests/os/data-floor-fallback-leg.sh` — `_floor_fallback_wait` now observes the reboot by boot id (`_reboot_wait`, the #1778 helper) before it polls for the fallback slot. One function, +8/−6 lines, no product bytes: `tests/` is not COPYed into the rootfs, so this is outside the image freeze.

## Why

Gating battery run 4 (`BUILD_COMMIT 670c33e9`, `--phase all`) went red on ONE row, in the floor-fallback leg:

> ✗ the previous boot's journal has no 'slot left uncommitted' line — either the 99.0.0 slot committed, or the journal did not persist

Read from the live guest before the harness tore it down (persistent journal, `journalctl --list-boots`): the 99.0.0 slot's boot IS in the list, 13 s long, WITH the gate line (`slot left uncommitted so A/B fallback stays armed`), and the three rows around the ✗ (restored floor, restore line on the console, re-install at the restored floor) all ran against the live fallback boot. So the product did what #1393 says. What the row's instrument could not see: `-b -1` from the boot the harness was talking to.

The helper did `_ssh reboot; sleep 10` and then polled for `marker == vmig` AND a `booted slot committed` line in `-b`. The boot that ISSUES the reboot satisfies both — it is the migration leg's committed slot — so any probe it answered while shutting down returned 0 before the guest went down, and `-b -1` on that boot is the boot BEFORE the migration commit, which has no uncommitted line. Same class as #1778, in the one file #1778 did not touch.

## What was RUN

- `bash -n` and `shfmt -i 4 -d` on the file: clean. Line count 198 → 200; the file has no tsv row (under 400).
- Offline control, `_ssh` stubbed with file-backed state, three reboot shapes, helper extracted from `origin/develop-v2` (old) and this branch (new), `_reboot_wait`/`_wait_new_boot`/`_boot_id` extracted verbatim from `run.sh`:

| scenario | old helper | new helper |
|---|---|---|
| A — the old boot never goes down | rc 0 after 1 probe, ON THE OLD BOOT (false green) | rc 1 at the deadline (honest) |
| B — the reboot lands before the first probe | rc 0 on the fallback boot | rc 0 on the fallback boot |
| C — the old boot answers 2 probes, then the 99.0.0 boot, then the fallback | rc 0 after 1 probe, ON THE OLD BOOT (false green) | rc 0 after 9 probes on the fallback boot; info line `the old boot … answered 2 probe(s)` |

Scenario B is the positive control that the fixed helper still passes the honest case; C is the shape the battery hit.

## What was NOT run

- The KVM leg itself. It runs next as `tests/os/run.sh --phase provision --keep` on the same `os/` bytes once this merges and the box is quiet, and that run is where the other hypothesis is settled: the fallback boot in run 4 has NO entry in the persistent boot list (nothing between the 99.0.0 boot's last line and the negative control's boot, ~7 min), i.e. its journal never reached `/data/pithead/journal`. If the fallback boot IS listed on the kept guest, this race was the whole cause; if it is NOT, that is a separate `os/` finding and gets its own issue. The `-b -1` row is left as it is on purpose, and NOT because it discriminates: after this fix the wait lands on the fallback boot, whose `-b -1` is the 99.0.0 boot, and run 4 already shows that boot's journal persisted with the gate line — so the row goes green whether or not the fallback boot's own journal reached `/data`, and only a red there would say anything (about the 99.0.0 boot, not the fallback boot). It stays because it is #1393's assertion and the defect was in the wait; changing what it asks in the same PR would leave the next battery's green unable to say which change bought it. The boot-list read on the kept guest is the discriminator. (Corrected after the reviewer's pass — the first body called the row the instrument.)
- `make lint-sh`: the gating chain holds the shellcheck serialisation lock until it exits. CI's shell job is the check.

## Over-engineering pass (by hand — the PR gate reads the wrong branch from this lane's cwd)

Two lines of code replaced by one call into an existing helper. Alternatives considered: converting both callers to capture the fallback boot id and key the `-b -1` row on it — rejected above, on purpose; adding a `sleep` before `_reboot_wait` — nothing to wait for, the helper's first act is to read the boot id from the still-running boot; passing `_reboot_wait` 300 s instead of `$2` (the reviewer's suggestion, truer to the other call sites) — rejected for THIS PR so it stays a one-mechanism change; the two deadlines share `$2`, which is immaterial at the call sites' 1800 s and worth a comment line on the next touch of the file.

Known trade-off, shared with the 14 sites #1778 converted (scenario D in the reviewer's terms): if the boot id is unreadable BEFORE the reboot, `_reboot_wait` returns 1 without sending the command, the helper returns 1 at once, and the caller's message says "did not come back within 30 min" — a wrong reason for a right refusal. Not fixed here.

Evidence dir on the bench host: `~/battery-1651-20260904T215352Z` (`battery.log`, `guest-journal/`).

### DIFF
--- a/tests/os/data-floor-fallback-leg.sh
+++ b/tests/os/data-floor-fallback-leg.sh
@@ -58,14 +58,16 @@ _floor_state() {
 }
 
 # Reboot into the just-installed slot and wait until the guest is back on the slot carrying marker
-# $1 with its boot gate PASSED (the commit line in this boot's journal). SSH answers on the failing
-# slot too, while it loads images and tries to bring the stack up, so the marker is what tells the
-# two apart. $2 = seconds: a failed `up` reboots within minutes, a gate that loops 90 rounds
-# takes ~15 min, and the fallback boot then re-loads images — allow for all three.
+# $1 with its boot gate PASSED (the commit line in this boot's journal). The reboot is OBSERVED
+# before anything is polled (#1651): the boot that issued it still carries marker $1 and a
+# committed line in its own journal, so a probe it answered while shutting down satisfied the
+# predicate off the OLD boot — and `-b -1` in the caller then named the wrong boot. SSH answers
+# on the failing slot too, while it loads images and tries to bring the stack up, so the marker
+# is what tells the two apart. $2 = seconds: a failed `up` reboots within minutes, a gate that
+# loops 90 rounds takes ~15 min, and the fallback boot then re-loads images — allow for all three.
 _floor_fallback_wait() {
     local deadline=$(($(date +%s) + $2)) got
-    _ssh reboot || true
-    sleep 10
+    _reboot_wait reboot "$2" || return 1
     while [ "$(date +%s)" -lt "$deadline" ]; do
         got=$(SSH_TIMEOUT=20 _ssh "cat /etc/pithead-test-marker 2>/dev/null; journalctl -u pithead-boot -b 2>/dev/null | grep -c 'booted slot committed'" 2>/dev/null | tr -d '\r' | tr '\n' ' ')
         case "$got" in "$1 "[1-9]*) return 0 ;; esac


VijitSingh97:
**Non-author reviewer pass — PASS on the code at `1e4a58e8`**, with one correction owed to the PR
body before merge (see the last section — the finding is another reviewer's, not mine). Three checks
were still pending as I wrote this.

I re-derived the scenario table rather than relaying it, and it reproduces exactly. I also confirmed
the defect independently from the leg's own source before looking at the fix.

## The defect is real, and the source says so without needing the battery

`phase_provision_floor_fallback_leg` asserts `_floor_state` = `$vfail|$floor0|$vfail`
(`data-floor-fallback-leg.sh:110`) immediately before calling `_floor_fallback_wait vmig 1800` at
`:117`. So at the moment of the reboot the guest is on the **migration leg's committed slot**, which
carries marker `vmig` **and** a `booted slot committed` line — which is precisely the old poll
predicate, `case "$got" in "$1 "[1-9]*)`. **The boot issuing the reboot satisfies the predicate it is
being polled against.** `sleep 10` was the only thing standing between that and a false green, and a
shutdown that outlasts ten seconds is not a rare guest state. The `-b -1` row downstream then names
the boot *before* the migration commit, which is exactly the row that went red in battery run 4.

## The scenario table, re-derived

Helpers lifted verbatim — `_boot_id` / `_wait_new_boot` / `_reboot_wait` from `tests/os/run.sh:157-180`
at this head, and both versions of `_floor_fallback_wait` extracted from `origin/develop-v2` and from
this branch by `sed` range, renamed `old_wait` / `new_wait`. `_ssh` stubbed against a guest timeline
**indexed by probes answered after the reboot command landed**, so both helpers see the same physical
guest no matter how many probes each spends:

| scenario | old helper | new helper |
|---|---|---|
| **A** old boot never goes down | rc 0 on **b1** after 1 probe — **false green** | rc 1, honest, with `the old boot b1 answered 40 probe(s)` |
| **B** reboot lands before the first probe | rc 0 on b3 | rc 0 on b3 — **the fix still passes the honest case** |
| **C** old answers 2 probes, then 99.0.0, then fallback | rc 0 on **b1** after 1 probe — **false green** | rc 0 on **b3** after 9 post-reboot probes, with `the old boot b1 answered 2 probe(s)` |

Your table, independently. **Two fixture errors of my own on the way there**, both worth naming
because either would have produced a confident wrong answer:

1. My first probe counter was a shell variable incremented inside `_ssh` — which `_boot_id` calls
   through a **pipeline**, so every increment happened in a subshell and the counter never moved. All
   six legs then read identically and the table looked meaningless. File-backed state fixed it. This
   is presumably why your own fixture is file-backed.
2. My scenario B first indexed the timeline by absolute probe number, which is unfair: `_reboot_wait`
   spends one probe on the before-read that the old helper never spends, so the two legs were being
   shown different guests. Re-indexing on post-reboot probes is what makes B a real control.

The fixture arms: the stub reports **b1, b2 and b3** across the timeline, so "it landed on b3" is a
measurement and not a stub that can only say one thing.

## Why the second transition is safe, which the body does not spell out

`_reboot_wait` returns as soon as **any** new boot id appears — in scenario C that is the 99.0.0 boot,
not the fallback. The 99.0.0 → fallback transition is therefore still unobserved by boot id. It is
safe anyway, and by content rather than by luck: the failing slot's marker is `$vfail`, not `vmig`
(asserted at `:110`), so the poll predicate cannot match it, and the loop rides through the second
reboot on its own. `-b -1` from the fallback boot then names the 99.0.0 boot, which is the boot that
carries `slot left uncommitted`. That is the row you are trying to make readable, and it lands on the
right boot.

## Budget, since the deadline moved

`deadline` is computed **before** `_reboot_wait` is called, and `_reboot_wait` is handed the same
`$2`. So the reboot observation spends from the same 1800 s the poll loop spends from, and the total
stays bounded by 1800 rather than becoming 1800 + 1800. Worst case is an honest rc 1 at the deadline.
The old shape spent a flat 10 s; the new one spends what the reboot actually takes. Correct, and
worth stating because "pass `$2` to both" is the shape that usually is not.

## Smaller confirmations

- `bash -n` on the file at head: rc 0.
- 198 → 200 lines; no row in `docs/dev/file-budget.tsv` for this file (grep count 0), so no ceiling moves.
- **`_reboot_wait` is defined at `run.sh:174`, after the leg is sourced at `run.sh:57`.** That is fine —
  names resolve at call time, not at source time — and it is the same ordering the `_ssh` the old code
  used already relies on (`run.sh:145`). No new hazard, but it is the kind of thing that reads alarming
  in a diff, so: checked.
- Wallet-fixture trap: not in reach, one file under `tests/os/`.
- `detect-topology.py` clean on this comment, with a seeded control that fired first.

## Not proven by me

- **I did not run the KVM leg**, and your PR is explicit that the real discriminator is the next
  `--phase provision --keep` run. My offline table says the helper is now honest about *which boot it
  is talking to*; it says nothing about whether the fallback boot's journal reached
  `/data/pithead/journal`. Those are two different causes and only the guest settles the second.
- No shellcheck, no `shfmt` (your `shfmt -i 4 -d` clean is relayed, and CI's shell job is the check),
  no `make` target, no bench, no image, no lint lock. I did not read the evidence dir on the bench host.
- I did not verify the battery-run-4 journal readings; they are yours, and the fix stands on the source
  reading above independently of them.

## A second pass exists, it is better than mine on one point, and I am not the merger

`~/dev/fleet/answers/reviewer-20260904T233549Z.md` is an independent non-author pass at this same
head. Its verdict table matches mine verdict-for-verdict on A/B/C (its probe counts differ from mine
because our guest models are parameterised differently; the verdict column is what either of us was
testing), and it adds two rows I did not model — a boot id unreadable at the pre-read (new helper
rc 1 immediately, reboot never issued) and a fallback boot that never comes.

**Its §2 finding is real and I did not derive it — relaying, not claiming.** The body says the
`-b -1` row "is the instrument that discriminates the two [hypotheses]". That reviewer traces it to
`os/overlay/pithead-journal-persist:6-7` and `…service:8,20`: the current boot is visible to
`journalctl` whether or not it ever reached `/data`, so after this fix `-b -1` from the fallback boot
resolves to the 99.0.0 boot and goes **green under both** hypotheses. **The row should stay — but the
sentence explaining why should be corrected before merge, because as written it promises a
discrimination the row cannot perform.** I checked only that the row now lands on the *right* boot,
which is a weaker claim than the body makes; their reading is the one to act on.

That reviewer also ran things I deliberately did not: `shfmt -i 4 -d` with the **pinned v3.13.1**
against a firing re-indent control, and an enumeration of `os/rootfs/Dockerfile`'s 29 explicit `COPY`
lines confirming none names `tests/` — which is what actually substantiates "no product bytes".

**I am not merging this.** `lane-fixes-d4` is the assigned non-author merger and asked me directly to
avoid us acting from different records. My pass is a second recorded PASS on the code, not a merge
instruction. Your `tests/os/` hold on #1708 is untouched and still waits for your "release chain
done"; the exemption you sent me covers this PR only.

**And the standing warning applies to the three pending checks: if any returns `cancelled`, that is
not a pass.** A cancellation is the absence of a measurement — re-run it and read the re-run. This
lane published the opposite earlier today and the re-run found a real defect that had been sitting
there the whole time.

Issue #286, opened 2026-06-17
tooling: adopt RigForge's CI best practices (diff-cover patch coverage + checksum-verified pinned tool installs)


VijitSingh97:
Part of #279 (Wave 7 tooling). **Back-port from RigForge** — its CI is ahead of ours in two reusable ways (cf. rigforge#116). Adopt both here.

> Context: comparing tooling across the org, RigForge's shell repo already does several things better than Pithead. `shfmt` and SHA-pinned actions are tracked in #281/#282; this issue covers the two remaining best-practices.

## 1. Patch-coverage gate (`diff-cover`) — stronger than our flat 80%

Today the dashboard gates on `pytest --cov-fail-under=80` — a **total** floor. A total floor lets new/changed code ship under-tested as long as the aggregate stays up, so coverage rots at the margin. RigForge instead enforces **patch coverage**: new/changed lines must be ≥90% covered, via `diff-cover` against the base branch reading a Cobertura report (see rigforge `ci.yml` coverage job + `tests/coverage.sh`).

- Emit Cobertura from pytest-cov (`--cov-report=xml`) in the dashboard CI coverage step.
- Add a **pinned** `diff-cover` and run `diff-cover coverage.xml --compare-branch=origin/develop --fail-under=N` (note: compare against `origin/develop`, Pithead's integration branch — not `main`).
- Needs `fetch-depth: 0` on checkout so diff-cover can diff the PR.
- Keep the existing total floor as a backstop; **patch coverage is the real ratchet** that stops rot.

## 2. Checksum-verified, pinned CI tool installs

RigForge installs shellcheck + shfmt at a **pinned version with a `sha256sum -c` check**, straight from the upstream releases — reproducible, and immune to the apt-mirror flakiness we hit in **#64**. Our CI uses the runner's preinstalled shellcheck and `pip install`s tools unpinned.

- Pin + sha256-verify the lint/security binaries CI downloads — shellcheck, shfmt (#281), hadolint, and the supply-chain tools (gitleaks/Trivy/zizmor, #282) — the way rigforge does.
- Prefer direct upstream-release downloads over `apt-get update` where practical (the root cause of #64).

## Acceptance

- [ ] Dashboard CI enforces **patch coverage** on changed lines (`diff-cover`) and keeps the total floor as a backstop.
- [ ] CI-downloaded tools are **version-pinned + sha256-verified**.
[...]

PR #1787, opened 2026-09-05
test(os): #1651 the media abort leg keeps its console and names what it saw when the cancel line is missing

VijitSingh97:
## What

Run 4 of the #1651 gating battery (`BUILD_COMMIT 670c33e9`, verdict on the issue at 00:05Z) reddened `tests/os/run.sh`'s media abort-leg row — *no cancellation confirmation ever appeared on the console after the media was pulled* — once in three `--phase all` runs on byte-identical `os/`, with the rows either side of it green (the diff banner appeared; the change was not applied). Nothing survived to say why: the harness truncates the console before the next boot, the leg deletes its stick, and later phases overwrite the disk.

This makes the row carry its own evidence next time, in the same eight lines of `phase_media`:

1. `_detach_media_stick`'s return code is reported with an `info` line instead of being discarded. A failed hot-unplug is one of the two ways to reach this red, and the harness could not see it.
2. On a miss, this boot's console is kept at `$SERIAL.failed`, no-clobber, the same contract `cleanup()` already honours for the end-of-run copy.
3. The red quotes the media channel's last console line. `cancelled` arriving late, `applied`, `could not write` and nothing-at-all are the four shapes; the quote tells them apart without a rerun.

Line-neutral: `run.sh` is 3423 lines before and after (its `file-budget.tsv` row is 3423, at the line count, so any growth would have needed a ceiling change). No product bytes: `tests/` is not copied into the image, so the freeze is untouched.

## What was RUN (by me, this worktree at `7dae676f` + this commit)

- `bash -n tests/os/run.sh` → clean. `shfmt -i 4 -d tests/os/run.sh` (the Makefile's flags, pinned 3.13.1 = local) → no diff. `awk 'END{print NR+0}'` → 3423.
- Offline control of the new branch, on the exact 8-line snippet extracted from the file by content, with `wait_serial`, `_detach_media_stick`, `ok`/`bad`/`info` stubbed and a synthetic console:
  - **A, fires:** detach rc 3 + wait miss + console ending in the `applied` line → `INFO: detach-device returned rc 3 …`, `BAD: … the channel's last line: 'Media configuration channel: applied. Continuing boot with the new configuration.'; console kept at …/serial.log.failed`, the copy exists, `FAIL=1`.
  - **B, no-clobber:** a pre-existing `.failed` holding `PRIOR` survives the miss unchanged.
  - **C, positive control:** wait hit + detach rc 0 → the `ok` row, no copy, `FAIL=0`.
  - A first attempt over-extracted by one line (the next block's `|| {` opener) and hit a syntax error at end-of-file after the snippet had run; that run was discarded and the control re-run on the 8-line snippet with `bash -n` clean first.

## What was NOT run, and why

- **`make lint-sh` / shellcheck: not run here.** The #1651 discriminating rerun on the bench holds `~/.shellcheck-serialize.lock` for its whole KVM leg, and shellcheck beside a 16 GiB guest is the OOM path; CI's shell job is the shellcheck evidence for this PR.
- **No KVM run of the changed leg.** The rerun in flight uses the lane tree at `7dae676f` (old wording). If this merges before the full `--phase all` gate, that run carries it; the change alters no verdict, only what a miss reports.
- **Docs:** `docs/dev/appliance-release.md` § the battery already says the harness keeps the guest console at `/tmp/pithead-os-serial.log.failed` on failure; still true and now true one row earlier, so no doc change. The house-voice pass therefore had nothing to read.

## Over-engineering pass (by hand — the PR gate keys off the wrong branch from this lane's cwd, disclosed)

Pure evidence capture; the assertion and its 90 s window are unchanged. Alternatives considered and rejected: widening the window (hides a slow countdown, the product prints within ~2 s of the pull); a shared helper for "keep console + quote last line" (one caller today; `cleanup()` and this site share a contract, not code); placing the copy in `failure-evidence.sh` (that file sits at its own ceiling and this is a single site). Eight lines in, eight out; two comment lines shortened to pay for the brace block.

Merge rule: needs a recorded non-author PASS; the `fixes` lane or the control seat merges. `Closes` is inert on `develop-v2`: #1651 stays open — this is one instrument for it, not its fix.

### DIFF
--- a/tests/os/run.sh
+++ b/tests/os/run.sh
@@ -2712,14 +2712,14 @@ phase_media() {
     _ssh reboot >/dev/null 2>&1 || true
     wait_serial "staged configuration differs from the running one" 180 || bad "no diff banner on the abort leg"
     # Pull the medium mid-countdown — the deliberate physical act that cancels a pending change.
-    _detach_media_stick
-    # Full expected text, matching the apply leg's own precision above (#1061): a bare
-    # "cancelled" would also match unrelated boot noise, and it is exactly the console's own
-    # wording that went missing when this issue was filed — the weak match could not have told
-    # "the right line appeared" from "some other word did".
+    _detach_media_stick || info "detach-device returned rc $? — the guest may still see the stick"
+    # Full expected text (#1061): a bare "cancelled" would also match boot noise. On a miss, keep this
+    # boot's console (no-clobber, as cleanup does) and quote the channel's last line, so the red carries it.
     wait_serial "Media configuration channel: cancelled" 90 &&
-        ok "removing the media mid-countdown cancels the change, and says so on the console" ||
-        bad "no cancellation confirmation ever appeared on the console after the media was pulled"
+        ok "removing the media mid-countdown cancels the change, and says so on the console" || {
+        [ -f "$SERIAL.failed" ] || cp "$SERIAL" "$SERIAL.failed" 2>/dev/null
+        bad "no cancellation confirmation on the console within 90 s of the pull — the channel's last line: '$(grep -o 'Media configuration channel: .*' "$SERIAL" | tail -1)'; console kept at $SERIAL.failed"
+    }
     _wait_ssh 180 || {
         bad "guest never came back after the cancelled change"
         return


VijitSingh97:
The red "Dashboard tests (pytest + coverage)" check on this PR is not this diff: the run's "Install uv" step logged `curl: (35) Recv failure: Connection reset by peer` and still passed (the step runs under `bash -e` without pipefail, so `curl | sh` returned sh's 0), and `make test-dashboard` then failed on `uv: not found`. Filed as #1788. The failed job gets a plain re-run once the run completes; the Shell tests job in the same run is the shellcheck evidence for this change (not run locally — the bench holds the shellcheck lock). Merge rule unchanged: recorded non-author PASS plus a green check.

VijitSingh97:
# pithead #1787 — non-author review

**MERGE-READY: PASS at `b3d9fac0518e1a35326a054c1a7c7f2aa9bc0119`** — all five asked checks hold, driven not read.
Two non-blocking findings below (one worth a free follow-up commit before it lands). **Do not merge yet**, for
two reasons that are not the diff's: CI is not green, and `tests/os/` is under the appliance lane's merge-order
hold.

Reviewed at head `b3d9fac0` in a fresh worktree off `origin/fix/1651-media-abort-console-evidence`. I am not the
authoring lane and hold no code in this repo.

---

## The five asked checks

### (1) `$?` after `||` reads the detach rc — CONFIRMED

`tests/os/run.sh:2715` `_detach_media_stick || info "detach-device returned rc $? …"`.

`_detach_media_stick()` (`:2541-2543`) is a single `virsh detach-device … >/dev/null 2>&1`, so its return is
virsh's. Driven with the function stubbed to `return 3`: `INFO: detach-device returned rc 3 — the guest may
still see the stick`. **Positive control:** the same driver with `return 0` printed nothing, so the row can give
either answer. Also checked the real shape (`command not found` inside the function) → `rc 127`, i.e. `$?` is the
function's status and not the `>/dev/null` redirection's.

### (2) `grep -o` inside the `bad` string — SAFE under `set -u` and on an empty console

- **`set -u` cannot fire.** `SERIAL="/tmp/pithead-os-serial.log"` at `:67` is an unconditional top-level
  assignment — there is no branch that leaves it unset. (`:41` is `set -uo pipefail`; no `set -e`, so the
  `grep` returning 1 on a miss cannot abort the leg, and the substitution sits in an argument position where
  its status is discarded anyway.)
- **Quoting parses.** Single quotes inside `$( )` inside `"…"` open a fresh context; `"$SERIAL"` re-enters
  double quotes correctly. `bash -n tests/os/run.sh` rc 0, and the extracted 8-line block ran.
- **Empty console** (the "nothing at all" shape): grep matches nothing, `tail -1` is empty, the row prints
  `the channel's last line: ''` and `FAIL=1`. No crash, no unbound error.
- **No format injection**: `bad()` (`:100-103`) prints via `printf '  …%s\n' "$1"`, so a `%` in the console
  text is inert.
- Nit only: `grep`'s **stderr is not redirected**, unlike `:422`, `:727`, `:753`. Driven with `$SERIAL` deleted
  it leaks `grep: …: No such file or directory` into the harness output. Unreachable in practice — `:2711`
  `: >"$SERIAL"` creates the file 9 lines earlier — so this is cosmetic, not a defect.

### (3) No-clobber vs `cleanup()`'s contract — MATCHES ON THE `-f` HALF, DROPS THE `-s` HALF

The `[ ! -f "$SERIAL.failed" ]` guard is exactly `cleanup()`'s at `:403`, and `:113-115` already names this new
site as the case it was written for — *"the cleanup copy below is no-clobber (the at-assertion copy is the
authoritative one)"* — with `rm -f "$SERIAL.failed"` at `:115` clearing the slate per run. Driven: a pre-existing
`.failed` holding `PRIOR-BOOT-CONSOLE` survived the miss byte-unchanged. Within one run `:2720` is the only
at-assertion writer, so its guard never actually skips — correct and consistent.

**But `cleanup()` guards on `[ -s "$SERIAL" ]` too, and `:2720` does not.** Driven (case E): with an empty
console the new site creates a **zero-byte** `$SERIAL.failed`. That empty file then satisfies `cleanup()`'s
`[ ! -f … ]` at `:403`, so a later failure in `phase_fault` or `phase_reset` (`:3412-3414` — media runs before
both under `--phase all`) has its console **suppressed**, and `:414` deletes `$SERIAL` on the way out. Net
evidence *loss*, introduced by an evidence-capture change.

Scoped honestly: reaching it needs a console with zero bytes after 180 s + 90 s of waiting, i.e. a guest that
never produced any output — in which case the later phases' consoles are likely empty too, so the cost is
small. **Nit, not a block.** Fix is one clause: `[ -f "$SERIAL.failed" ] || [ ! -s "$SERIAL" ] || cp …`.

### (4) `shfmt -i 4 -d` clean at pinned 3.13.1 — CONFIRMED, with a firing control

Pinned binary (`shfmt --version` → `v3.13.1`, the `SHFMT_VERSION` `make lint-sh` enforces at `Makefile:89-92`):
`shfmt -i 4 -d tests/os/run.sh` → **rc 0, empty diff**. `bash -n` rc 0.

**Armed positive control**, because a clean run proves nothing on its own: re-indenting the new
`[ -f "$SERIAL.failed" ] || cp …` line by two spaces (mutation diffed back to prove it landed) → **rc 1**, with
the reported hunk pointing at exactly that line. So the tool read this file with these flags.

Note the apt `shfmt 3.8.0` also on this box reports rc 0 on the same file, so the two-binary hazard of #1688
does not bite here — but the measurement that counts is the pinned one, and CI ran the same pin.

### (5) Nothing outside `tests/os/` — CONFIRMED

`git diff --stat origin/develop-v2...HEAD` → `tests/os/run.sh | 14 +++++++-------`, **1 file changed, 7
insertions, 7 deletions**. `gh pr view --json files` agrees (one entry, 7/7).

Line-neutral holds and matters: `awk 'END{print NR+0}'` → **3423**, and `docs/dev/file-budget.tsv:70` is
`tests/os/run.sh	3423` — the row sits *at* the count, so any growth would have failed the budget gate with no
ceiling change in the diff. No leak class in the diff: no wallet address, hostname or IP added.

---

## Shellcheck — CI's shell job IS the evidence, and it passed

`make lint-sh` shellchecks `tests/os/*.sh` at `--severity=warning` (`Makefile:98`, last argument of the first
invocation), so this file is covered — the PR body's claim is true, not assumed.

Read from the job's **per-step conclusions at source** (job `101213052667`, run `33932240551`, head `b3d9fac0`),
not from the check label:

    success  Install shellcheck + shfmt (pinned + sha256-verified, #286)
    success  Lint pithead, build/* + dashboard/ container scripts, and test scripts (shellcheck + shfmt)

Ten of eleven steps `success`; only the final `Run integration harness self-test` is `cancelled`, killed by the
workflow's fail-fast after the unrelated dashboard job died. **The label reads `cancelled` — but the
shellcheck step is a completed measurement, not an absent one.** (The converse trap is this lane's standing
carry-forward: a cancelled *step* is the absence of a measurement. Here the step I need finished.)

Corroboration that no new shellcheck rule was going to fire: `$?` read after a command already appears in this
file at `:2438` and `:3034`, and nine times in `tests/stack/test-appliance-setup.sh`, all under the same gate.

---

## The one finding worth a follow-up commit: the quoted line carries the guest's `\r`

`:2721` reads `$SERIAL` **raw**. The virsh serial file is CRLF, and this repo says so itself: eleven sites strip
it before reading that same file — `:753`, `:769`, `:1231`, `:1414`, `:1554`, `:1666`, `:1793`, `:1976`,
`:2756`, `:3155`, and **`:2636`, eighty-five lines above, inside `phase_media` itself**.

Driven on the extracted block with a CRLF console, `cat -v` on the output:

    BAD: … the channel's last line: 'Media configuration channel: applied. Continuing boot with the new configuration.^M'; console kept at …

The `^M` sits between the evidence and the closing quote. Rendered to a terminal it returns the cursor to
column 0, and `'; console kept at /tmp/pithead-os-serial.log.failed` overwrites the row's first ~46 characters
— so the single line this PR exists to produce prints scrambled, on **every** miss it is meant to serve. The
bytes survive in a piped log (recoverable with `less`/`cat -v`), so this degrades the deliverable rather than
destroying it: a nit, not a correctness defect, and not a reason to hold the code.

**Why the author's control did not catch it:** the offline drive used a synthetic console, and a heredoc-written
fixture has no `\r`. The fixture could not see the thing it was controlling for. Fix, matching the file's own
convention: `tr -d '\r' <"$SERIAL" | grep -o 'Media configuration channel: .*' | tail -1`.

---

## What I did NOT run

No KVM, no bench, no guest, no image, no `make` target of any kind, and **no shellcheck locally** — the
appliance lane holds the box and the shellcheck serialize lock for the #1651 chain. Every claim above is either
CI's own step conclusion read at source, or an offline drive of the 8 lines extracted verbatim from the file at
`b3d9fac0` with `wait_serial`/`_detach_media_stick`/`ok`/`bad`/`info` stubbed. I did not verify the change
against a real media-abort boot; nobody can until the battery frees the bench. The change alters no assertion
and no timeout, which is what makes that acceptable.

## Two things that gate the merge, neither of them the diff's

1. **CI is not green.** `Dashboard tests (pytest + coverage)` failed in 9 s: the uv install logged
   `curl: (35) Recv failure: Connection reset by peer` and `make test-dashboard` then hit `/bin/sh: 1: uv: not
   found` → `Error 127`. Runner-network flake, nothing to do with `tests/os/`. It fail-fast-cancelled the shell
   and build jobs. Needs a re-run — **and the re-run must be read, not assumed**. The author posted the same
   read at 00:19Z; I derived it independently from the job log.
2. **`tests/os/` is under the appliance lane's merge-order hold.** Their sequence is "release build starting" →
   "BUILD_COMMIT captured" → "release chain done", and nothing under `tests/os/` merges before the third
   message. The #1786 exemption was granted for one named file and does not extend here. `run.sh:274` re-invokes
   `verify-image.sh`, which is the exact coupling the hold exists to protect — so this one needs the hold's
   owner (the appliance lane) to say yes, not an inference from #1786.

`Closes` is inert on `develop-v2` — #1651 stays open; this is one instrument for it, not its fix.

PR #1774, opened 2026-09-04
release: prepare v2.0.0

VijitSingh97:
## What changes

The release-prep commit for **2.0.0**, the first release of Pithead OS alongside the Compose install — the same four-file shape as the v1.20.0 prep (50fcf433):

- `VERSION` → `2.0.0` (plain triple: the appliance's os-update verbs and `aged_version` refuse a `-rc` suffix; the rc number lives only on `release.sh --rc 1`'s staging tag).
- `dashboard/pyproject.toml` and the `mining-dashboard` line in `dashboard/uv.lock` → `2.0.0` (`tests/stack/test-release.sh:228` asserts the pyproject version equals `VERSION`).
- `CHANGELOG.md`: a `## [2.0.0]` section (107 lines). `release.sh` publishes the top section as the release notes, so this is the user-facing summary of both channels.

## Serialized and out-of-lane files — disclosed

`VERSION` and `CHANGELOG.md` are serialized files in nobody's `.paths`; `dashboard/*` belongs to the dashboard lane. All four are in one commit under the one-shot `.lane-override`, because they are one release commit: the appliance image is built from it and `verify-image` pins it as `PITHEAD_EXPECT_COMMIT`. Expect a conflict with any other CHANGELOG or VERSION change; there is none open.

## Ordering — this PR must not merge before the gating battery is green

The #1651 battery is running on the frozen tip (`git ls-remote origin develop-v2` re-derives it). This commit changes image bytes by design (`VERSION` is baked; `dashboard/` is the baked dashboard image), so it merges only after that verdict, as the seat's step 5. Opened now so the non-author pass overlaps the battery.

## What was RUN

- `uv lock --check --offline` in `dashboard/`: rc 0 after the lock edit (re-read without a pipe).
- Banned-word sweep (docs/dev/STYLE.md list) over the new section: no hits. `CHANGELOG.md` is excluded from `lint-docs-voice`, so this was by hand.
- Every issue number cited in the section was checked against the first-parent commit list since `v1.20.0` (438 commits); two candidates that did not appear were dropped, and one that appeared only as a PR number was cited by the issue the branch named.
[...]

PR #1792, opened 2026-09-05
fix(os): #1791 order the persistent-journal bind after the /var overlay so the journal has one home

### LINKED ISSUE #1791: appliance: the #1030 journal bind and the /var overlay race at boot, so the persistent journal splits across two directories and `--list-boots` silently omits boots
## What the kept KVM guest shows

Read on the #1651 discriminating rerun's kept guest (image `BUILD_COMMIT 7dae676f`, `os/` byte-identical to the freeze tip `b4d818ec`), root over SSH, read-only until the last step. Evidence dir on the bench: `~/battery-1651-disc-20260905T000751Z/` (`guest-journal/`, `guest-media/`).

- `pithead-journal-persist.service` logs `bound /var/log/journal to /data/pithead/journal` on every boot, one second after journald starts. `findmnt -T /var/log/journal` reports `/dev/vda4[/pithead/journal]`.
- Yet `/data/pithead/journal/` is **empty** (4.0K, mtime = the guest's first boot, no `<machine-id>/` subdirectory), while `/var/log/journal/<machine-id>/` holds 8.1M and `journalctl --disk-usage` reports 18.8M after seven boots.
- Inodes differ: `/var/log/journal` is inode 4 with link count 1 and an ACL (`drwxr-sr-x+ 1`); `/data/pithead/journal` is inode 1310722 with link count 2 on ext4. `stat -f` says `/var/log/journal` is **overlayfs**.
- `/proc/self/mountinfo` (columns: id parent maj:min root mountpoint):
  ```
  29 1  254:2 /                /                 ext4 /dev/vda2 ro
  49 29 254:4 /                /data             ext4 /dev/vda4
  51 29 254:4 /pithead/journal /var/log/journal  ext4 /dev/vda4
  59 29 0:39  /                /var              overlay lowerdir=/var,upperdir=/data/overlay/var,workdir=/data/overlay/var-work
  ```
  Mount 51 (the #1030 bind) is a child of the root mount. Mount 59 (the `/var` overlay from `os/rauc/populate-slot.sh:183`'s fstab line) came later and covers the whole `/var` subtree, bind included.
- Journald's real directory: `find` puts `<machine-id>` at `/data/overlay/var/log/journal/<machine-id>`, the overlay's upper dir.
- Timing on the kept boot (`journalctl -o short-monotonic`): `var.mount` starts 4.873 s, `pithead-journal-persist` starts 4.878 s and finishes its bind 4.960 s, **`Mounted var.mount` 4.979 s**. The bind won by 19 ms and was buried anyway.
- Ordering (`systemctl show`): `var.mount` is `After=data.mount`, `Before=local-fs.target systemd-journal-flush.service`. `pithead-journal-persist.service` is `After=data.mount`, `Before=sysinit.target systemd-journal-flush.service`, `RequiresMountsFor=/data`. **Neither orders against the other.** Both start within 5 ms of `data.mount`.

## Mechanism

Two persistent homes for the journal exist on `/data`: the #1030 bind target `/data/pithead/journal`, and the `/var` overlay's upper dir `/data/overlay/var/log/journal`. Which one journald writes on a given boot is decided by a race between `var.mount` (one `mount` syscall) and `pithead-journal-persist` (five forks: `mkdir`, `chown`, `chmod`, `mountpoint`, `mount`):

- **Overlay finishes last** (the observed case): the bind is covered; journald flushes into the overlay upper dir. `/data/pithead/journal` stays empty.
- **Bind finishes last**: the bind lands on top of the overlay at `/var/log/journal`; journald flushes into `/data/pithead/journal`. Every other boot's journal, sitting in the overlay upper dir, is invisible to this boot, and this boot's journal is invisible to every boot that goes the other way.

Either way the bytes persist, so a single-boot check passes. The failure is a **split journal across boots**: `journalctl --list-boots` and `journalctl -b -N` on the appliance silently omit whichever boots lost the race. The five-fork script is the side that slows under load, which is when the flip is likeliest.

## Fired control, on the same guest, before it was destroyed

The race's other outcome, forced by hand: `mount --bind /data/pithead/journal /var/log/journal` after the overlay was up.

| step | reading |
|---|---|
| before | `--list-boots` = 4 boots; top mount at `/var/log/journal` = overlay dir (inode 4); `/data/pithead/journal` empty |
| bind on top (parent = mount 59, inode 1310722) | `journalctl --list-boots` → **`No boot found.`** |
| `systemctl restart systemd-journald`, then `logger -t split-test` | `/data/pithead/journal/<machine-id>/` appears; `--list-boots` = **1 boot** (the current one); the three earlier boots are gone from the list; the `split-test` line is readable |
| `du` | 8.1M in each of the two directories |

## What it probably explains

Run 4 of the #1651 gating battery (`--phase all`, under load) reddened one row because the boot list on the live guest went from the 99.0.0 slot's failed boot straight to the next 99.0.0 boot, with the seven-minute fallback boot between them absent, while that boot's own `-b` readings were fine at the time. That is this race's signature: the fallback boot lost, wrote into `/data/pithead/journal`, and the next boot won and read the overlay side. On the quiet-box rerun with #1786 in, all seven boots landed on the same side and the row passed. I have not reproduced the flip by timing alone, only forced it; the load dependence is inferred.

## Cost

Diagnosability on the exact path an unattended box relies on: an operator following any `journalctl -b -1` / `--list-boots` instruction after a failed update can be shown a list with the interesting boot missing, with no error. #1030 (persist before the first flush) and #1659 (stable machine-id) were both about making that read trustworthy; this is the third leg. It also makes `/data/pithead/journal` a name that lies: the directory the unit announces is not where the journal lives.

## Fix shapes (operator's call; either moves the image freeze)

1. **Retire the bind.** The `/var` overlay already persists `/var/log/journal` on `/data`, which is what #1030 wanted. Drop `pithead-journal-persist.service` (and the seeded `/pithead/journal` dir), keep `systemd-journal-flush` ordered `After=var.mount` (it already is). One directory, no race. `journalctl` readers and `pithead-data-reset` (`SEED_DIRS` already wipes `/overlay/var`) need no change.
2. **Order the bind after the overlay.** `After=var.mount` + `RequiresMountsFor=/var/log` on the unit, so the bind always lands on top and the journal always goes to `/data/pithead/journal`. Keeps the named path, keeps two mechanisms.

Detector first, whichever shape: a `tests/os/run.sh` row on a provisioned guest that reads `findmnt -n --target /var/log/journal -o SOURCE` and `stat -c %i /var/log/journal` against the intended home, plus `journalctl --list-boots` count ≥ 2 after the first reboot of a leg that already reboots. The row is what would have caught this on run 1.

Found while discriminating #1651's red 1 (the boot-list gap). Filed under the seat's rule for a product finding inside the freeze: the operator decides between fix-now (freeze moves, full battery reruns) and ship-with-known-issue (the rollback itself is proven; only the fallback boot's diagnostics are at risk).

VijitSingh97:
Fixes #1791 (hand-close on `develop-v2`). Moves the image freeze: this changes `os/overlay/`, so the #1651 gating battery reruns on the new tip before #1774 merges (seat ruling 2026-09-05, BUGS-FIRST).

## What was wrong

`/var` is an overlay whose upper dir is on `/data` (`os/rauc/populate-slot.sh`'s fstab line, July 24), and the #1030 unit (August 22) that binds `/data/pithead/journal` onto `/var/log/journal` was unordered against it. Both are `After=data.mount` and started within 5 ms of it; whichever finished LAST owned `/var/log/journal`. A boot's journal landed in one of two persistent directories, and `journalctl --list-boots` silently omitted the boots that went the other way. That is how the seven-minute fallback boot vanished from run 4's boot list (#1651 red 1). The #1030 headers claimed `/var/log/journal` sat on the read-only root; the overlay predates the bind, so that was never true.

## The fix (shape 2 of the two in #1791, and why)

One ordering line: `After=var.mount` on `pithead-journal-persist.service`. The bind is now the top mount on every boot, so the journal has ONE home and it is the named, tested one (`/data/pithead/journal`, created by `pithead-journal-persist` inside the `/pithead` directory `pithead-data-reset` seeds, and the path the #1030 tier-1/tier-3 rows assert).

- `After=` only, not `Requires=`/`RequiresMountsFor=/var/log`: a bind needs its target to exist as a directory, which the baked `/var/log/journal` is on the read-only root too, so a boot whose overlay failed still persists its journal here. Tying the bind to the overlay would lose exactly the boot #1030 was about.
- No cycle: `var.mount` is `Before=local-fs.target`, which `sysinit.target` orders after, and the unit stays `Before=sysinit.target`. `systemd-journal-flush.service` already orders after `var.mount` on its own (`RequiresMountsFor=/var/log/journal`); the existing `Before=` keeps it after the bind too.
- **Shape 1 (retire the bind) was available and rejected.** It is the cleaner long-run shape (one mechanism), but it deletes a unit, a script, three Dockerfile lines, four verify-image rows and the tests lane's `test_firstboot_journal.sh` section, and it makes first-boot journal persistence depend on the overlay having mounted. Under a freeze the operator ruled to move by the smallest correct change, one ordering line wins.
- **Migration cliff, raised by the `fixes` lane before push (their static reading, re-derived by me): a box that already holds journals in the overlay upper dir would lose them from `--list-boots` on the first boot after this fix.** No such box will exist: every box that ever boots this image is flashed from it (the seat's step 5/6 sequence USB-flashes the soak box from the release image; the day-7 consumer leg updates that same box; bench guests are created from the image per battery and destroyed). A one-time `mv` in an early-boot unit that no shipped box can execute, unprovable at tier 4 without a pre-fix guest, is a worse trade than a documented no-op. Decided, not overlooked.

## Detector first, fired both ways

`journal_home_verdict` (`tests/os/journal-boot-verdict.sh`, beside #1659's) rides the boot phase's existing #895 reboot and asks the guest: what the VFS resolves at `/var/log/journal` (`stat -c %m` and the inode against `/data/pithead/journal`), whether journald flushed under the bind this boot (`/data/pithead/journal/<machine-id>` exists), and how many boots the persistent journal lists (`journalctl --list-boots`, ≥ 2 after the reboot).

- **Not `findmnt --target`**, which the seat's brief and the issue both named: the issue's own reading on the kept guest shows `findmnt -T /var/log/journal` printing `/dev/vda4[/pithead/journal]` WHILE journald wrote to the overlay, because findmnt names the bind by its mountpoint even when a later mount at `/var` covers it. `stat -c %m` walks up by device and answers `/var` (overlay) or `/var/log/journal` (bind); the inode says whether that is the `/data` directory.
- Tier 1 (`tests/stack/test-appliance-machine-id-journal.sh`, 10 rows): the kept guest's own reading (`/var 4 / 1310722 / overlayfs / no / 4`) fails naming the overlay; the fixed shape passes; no-flush, one-boot, empty and garbage probes fail; and a field-order control feeds the REAL probe's output on this host to the verdict and checks the missing `/data` inode lands in the right field. Domain run alone: 28/0.
- Tier 3 (`tests/os/verify-image.sh`): the existing ordering row now also requires `^After=var.mount`. **Negative control, run by me on the unfixed image (BUILD_COMMIT `7dae676f`, `os/` == freeze tip `b4d818ec`): 102 passed, 1 failed, the failed row is this one.** Line-neutral, the file stays at 400.
- Tier 4, `--phase boot` (one cold boot + the reboot, ~12 min): **NEGATIVE, unfixed image (BUILD_COMMIT `7dae676f`, `os/` == freeze tip, the media phase's ESP pre-seed removed first): rc 1, 6 ✓ / 1 ✗, the one ✗ is the new row — `journald's directory is the overlayfs mount at /var (inode 6), not the /data/pithead/journal bind (inode 131074): the /var overlay covered it (#1791)` — with #1659's row green beside it, which is the single-boot blindness the issue describes (evidence `~/battery-1791-neg-20260905T011344Z/boot.log`).** · **POSITIVE, fixed image built from this branch at `188c4236` (chain `~/battery-1791-pos-20260905T012322Z/`: build-image rc 0, mkimage rc 0, verify-image --test 103 passed / 0 failed with `PITHEAD_EXPECT_COMMIT` = the tip, image sha256 `d8d1d832…`): `--phase boot` rc 0, 7 ✓ / 0 ✗, the new row reads `journald writes the /data/pithead/journal bind (inode 524291, ext2/ext3) and the persistent boot list holds 2 boots across the reboot`.** The same instrument, the same phase, the same host, on the two images: red without the ordering line, green with it. Commits after `188c4236` touch `tests/os/journal-boot-verdict.sh` (a shellcheck directive) and `tests/stack/test-appliance-machine-id-journal.sh` (the reviewer's F2 row, below) only; `git diff --stat 188c4236 HEAD -- os/ lib/pithead/ pithead build/ docker-compose.yml` is empty, so the built image is the branch's image.

## Budget, lanes, what I did not do

- `tests/os/run.sh` stays at its 3423 ceiling: the one added call line is paid for by folding the four-line #895/#1659 comment above it into three. `docs/dev/file-budget.tsv` untouched; `make lint-file-budget` OK. `journal-boot-verdict.sh` is 83 lines (under target, no row).
- `.lane-override` (one-shot, consumed) for `tests/stack/test-appliance-machine-id-journal.sh`: the tests lane owns it and is wound down; it is the file that holds the sibling `journal_boot_verdict` fixtures. `docs/` and `tests/os/README.md` are shared/mine.
- NOT done: an update-phase row asserting the boot list is preserved across a RAUC update (the `fixes` lane's suggestion). Good detector, but run.sh has no headroom and the fresh-guest boot row is what would have caught this on run 1. NOT done: the full battery on this branch; that is the gate after merge, on the new freeze tip.
- Bench note, not a defect in a battery: `phase_media` pre-seeds `os/rauc/build/system.img`'s ESP in place (`pithead-config.json`), so an image reused OUTSIDE `--phase all` after a media leg takes the pre-seed path and the boot phase's wizard rows go red. My first negative-control run hit exactly that (kept as `boot-seeded-false-start.log`); I removed the seed from the ESP and reran. Inside a battery every phase rebuilds first, so no phase sees it.

CI: the first Shell run went red in 2m42s on SC2034 (`JOURNAL_HOME_PROBE` is read by `run.sh`, invisible from the sourced file); fixed with a directive in `aacba607`, read per-step at source, not off the label. Local `make lint-sh` was killed at the shellcheck wrapper's 6 GiB cgroup cap (#1206, pre-existing, the whole-tree input) while shellcheck over the changed files alone is clean; CI's shellcheck is the gate here. The tier-1 domain ran alone at 28/0; the full `tests/stack/run.sh` is running locally as this is written and CI's Shell job runs it too.

Over-engineering pass by hand (the ponytail gate keys off the wrong branch from this lane's cwd): the product diff is one directive plus true comments; the verdict is one function and one probe string; every added row has a mutant that kills it (listed above, plus the inode-arm row the reviewer's F2 found missing — taken, measured below). The ponytail gate's prompt was a one-shot deny/retry, not a review.

## Reviewer findings (non-author pass, MERGE-READY at `aacba607`) — each re-derived by me at source, then acted on

- **F1, two wrong figures — CORRECTED IN PLACE above.** The body said `9 ✓ / 1 ✗` and `10 ✓ / 0 ✗`; the two `boot.log`s read `os harness: 6 passed, 1 failed` and `7 passed, 0 failed` (`grep -c` on the glyphs agrees: 6/1 and 7/0). The discriminating fact is unchanged. The figures were wrong because I wrote them from memory of the row list instead of from the captured log — the exact error the body's own method forbids.
- **F2, the inode arm had no fixture that kills it — TAKEN as `9d2f115f`.** Two lines in `tests/stack/test-appliance-machine-id-journal.sh`: `mnt=/var/log/journal`, `tino=999`, `hino=1310722`, flushed, two boots — the one shape only the inode arm sees. Measured by me: domain alone 29 passed / 0 failed on the clean verdict file; with `|| [ "$tino" != "$hino" ]` deleted (mutation proven applied by `git diff --numstat` = 1/1 on the staged file) 28 / 1, the red being this row alone; verdict file restored, 29 / 0 again. The commit used the one-shot `.lane-override` hatch a second time for that file (it is outside `roles/appliance.paths`; the #1659 grant on it expired), consumed on the commit and absent after. Commits after `aacba607` touch only that test file: the product-path diff `git diff --stat 188c4236 HEAD -- os/ lib/pithead/ pithead build/ docker-compose.yml` is still empty, so the images the controls ran on are still the branch's images.
- **F3, wrong attribution — CORRECTED IN PLACE above.** `os/overlay/pithead-data-reset:29` seeds `/overlay/var /overlay/var-work /pithead` — the parent — and the `journal` leaf is created by `pithead-journal-persist`'s `mkdir -p "$DATA_DIR"` (`:22`). Body text only; no assertion depended on it.

- **Reviewer's re-pass at `9d2f115f` (MERGE-READY), one more finding — CONFIRMED BY ME at the base tree `418d50f7`.** The tier-3 row this PR rewrote (`journal-persist orders before the journal flush it exists to feed`) was previously UNFAILABLE: it was an unanchored `grep -q "systemd-journal-flush.service"` over the unit file, and the unit's own comment block (line 8 at the base) names that service, so deleting the real `Before=` directive would have left the row green. The new form anchors both greps at `^` (`^After=var.mount`, `^Before=.*systemd-journal-flush.service`), so it can now fail — the negative control's 102/1 is the first time that row has ever been shown able to. Anyone weighing the OLDER tier-3 runs' green on that row should read it as "the comment block was present", nothing more.

**Merge gate for the `fixes` lane:** the reviewer's PASS is recorded at `aacba607` conditional on green CI; the delta to the head is the two-line F2 row above and nothing else (`git diff aacba607 HEAD --stat`). Read CI green at the head, read the delta, then merge (squash; the body above is the durable record — `Closes` is inert on `develop-v2`). **A green rollup on this PR is NOT the #1651 gate:** every execution claim here is the author's (the reviewer ran no battery under the CI-only rule), and the full battery reruns on the new freeze tip before #1774 moves — the merger confirms that rerun exists rather than reading these checks as the gate.


VijitSingh97:
# reviewer — non-author pass on pithead #1792 (#1791, `develop-v2`)

**MERGE-READY: PASS at `aacba607925ff644eb965a01eed97ddd786a0db0`**, conditional on green CI
(IN_PROGRESS at the time of writing). **Not mine to merge — freeze path; the `fixes` lane merges.**
Three findings, none blocking correctness: two are wrong figures in the body, one is a measured
surviving mutant in the new tier-1 fixture set.

The product fix is right and the evidence is real. Both battery legs are the same instrument on the
same host against two images that differ only in `os/`, and the discriminating row flips.

## The reviewer's claims, checked against the evidence dirs

**All CONFIRMED unless marked.** Every figure below I read at source; none is quoted from the body.

| Body claim | What the evidence says | |
|---|---|---|
| neg `--phase boot` rc 1 | `neg/boot.rc` = `1` | ✓ |
| neg: the new row is the only red | `neg/boot.log` — one `✗`, verbatim `journald's directory is the overlayfs mount at /var (inode 6), not the /data/pithead/journal bind (inode 131074): the /var overlay covered it (#1791)` | ✓ |
| neg: #1659's row green beside it | same log — `journald follows the restored machine-id across a reboot (journal directories: 1, …)` is `✓` | ✓ |
| neg: **"9 ✓ / 1 ✗"** | **6 ✓ / 1 ✗** (`os harness: 6 passed, 1 failed`) | **✗ F1** |
| neg image = BUILD_COMMIT `7dae676f` | not taken from `run2.sh`'s label — `neg/verify-image-unfixed.log:87` prints `built from the expected commit (7dae676f0c296df3681c0fb9977820aad97d7ff8)` | ✓ |
| neg image's `os/` == freeze tip `b4d818ec` | `git diff --stat 7dae676f b4d818ec -- os/` empty; **and** `7dae676f`↔`418d50f7` (current tip) empty too, so the control is the shipping `os/` | ✓ |
| tier 3 negative: 102 passed, 1 failed, the new row | `verify-image-unfixed.log:118` `102 passed, 1 failed`; the `✗` is `journal-persist orders after the /var overlay (#1791) and before the journal flush it exists to feed` | ✓ |
| pos `--phase boot` rc 0 | `pos/boot.rc` = `0` | ✓ |
| pos: **"10 ✓ / 0 ✗"** | **7 ✓ / 0 ✗** (`os harness: 7 passed, 0 failed`) | **✗ F1** |
| pos row text, inode 524291 | `pos/boot.log` verbatim: `journald writes the /data/pithead/journal bind (inode 524291, ext2/ext3) and the persistent boot list holds 2 boots across the reboot` | ✓ |
| pos chain: build 0 / mkimage 0 / verify 103-0 / sha `d8d1d832…` | `build-image.rc`=0, `mkimage.rc`=0, `verify-image.rc`=0 (`103 passed, 0 failed`), `sha256.log` = `d8d1d832…` | ✓ |
| pos `PITHEAD_EXPECT_COMMIT` = the tip | `pos/verify-image.log:87` `built from the expected commit (188c4236c6f38bd2e3dbd03d98a045e44533f1a0)`; `BUILD_COMMIT`/`tree.txt` agree, `dirty=0` | ✓ |
| commits after `188c4236` touch only the verdict file | `git diff --stat 188c4236 aacba607` = `tests/os/journal-boot-verdict.sh \| 4 +++-`; the product-path diff is empty | ✓ |
| bench note: ESP pre-seed reds the wizard rows | `boot-seeded-false-start.log` — 1 ✓ / 2 ✗, both wizard rows; `chain.log` records the seed removal and the sha change `d954c6bd…` → `7ab58ff2…` | ✓ |

Row counts across every log in both dirs: `neg/boot.log` 6/1 · `neg/boot-seeded-false-start.log` 1/2 ·
`neg/verify-image-unfixed.log` 102/1 · `pos/boot.log` 7/0. **No log in either dir shows 9 or 10 `✓`.**

## What I re-derived beyond the body

- **Base is current.** `git merge-base origin/develop-v2 aacba607` = `418d50f7` = the develop-v2 tip,
  and `318b1af4^` IS `418d50f7`. No stale-base problem (unlike #1785).
- **`After=var.mount` names a real unit and cannot cycle.** `os/rauc/populate-slot.sh:183` writes
  `overlay /var overlay lowerdir=/var,upperdir=/data/overlay/var,workdir=/data/overlay/var-work,x-systemd.requires-mounts-for=/data`,
  so systemd-fstab-generator emits `var.mount` with the default `Before=local-fs.target`, which
  `sysinit.target` orders after. The unit's `Before=sysinit.target` therefore sits downstream, not
  upstream. Empirically settled anyway: the fixed guest reached userspace, rc 0.
- **The tier-3 row is a strengthening, not a relabel.** The old assertion was an unanchored
  `grep -q "systemd-journal-flush.service"` — which the unit's own comment block satisfies. The new
  one anchors `^After=var.mount` and `^Before=.*systemd-journal-flush.service`, and the `^` is what
  keeps it from matching the comment block that now contains the literal `After=var.mount`.
  (`tests/os/verify-image.sh:347-348`.)
- **Budgets.** `tests/os/run.sh` 3423 at base and at head, ceiling row `docs/dev/file-budget.tsv:70`
  = 3423; `verify-image.sh` 400/400; `file-budget.tsv` untouched.
- **Leak sweep clean** on the diff and the body: no IP, no `/home/…`, no hostname, no address-shaped
  token. (The IPs live in the evidence logs, which are untracked and stay off the PR.)
- **`bash -n` clean** on all five changed shell files at head.
- **Characterisation, in the PR's favour:** the negative control is one draw of a race, so on its own
  it would prove only that the detector CAN see the bad state. It is stronger than that because the
  verdict has two arms: the boot-count arm fires when boot 1 and boot 2 land in different homes even
  if this boot's top mount happens to be the bind. False-green on an unfixed image needs both boots
  to land in the bind; on a fixed image the ordering is deterministic, so no flake in the other
  direction. One bench run is adequate evidence here.

## Findings

### F1 — two row-count figures in the body do not match the evidence it cites (fix before merge)

`9 ✓ / 1 ✗` and `10 ✓ / 0 ✗` against measured `6 ✓ / 1 ✗` and `7 ✓ / 0 ✗`. Both off by exactly 3, in
the same direction. Nothing downstream moves — the discriminating fact (new row is the only red,
then green; #1659 green beside it in both) survives — but the body is the durable record here
(`Closes` is inert on `develop-v2`). The constituent **commit** messages carry no row counts, so a
default squash body does not propagate the error; only the PR body needs correcting.

### F2 — measured: the inode arm of the new verdict has no fixture that kills it

The body says "every added row has a mutant that kills it (listed above)". Not for
`[ "$tino" != "$hino" ]`. I copied `tests/os/journal-boot-verdict.sh` at `aacba607`, deleted
`|| [ "$tino" != "$hino" ]` from `journal_home_verdict`, printed the diff to prove the mutation
landed, and drove every shipped tier-1 shape through both copies:

```
orig=0 mut=0   the bind tops it, flushed, 2 boots         (pass shape)
orig=1 mut=1   the /var overlay tops it (kept guest)
orig=1 mut=1   bind up, journald flushed nothing
orig=1 mut=1   one boot listed
orig=1 mut=1   empty probe
orig=1 mut=1   /data inode missing
orig=1 mut=1   the live-probe field-order control, byte-identical output both sides
--- the shape no fixture uses: both inodes numeric and DIFFERENT
orig=1 mut=0   /var/log/journal 999 / 1310722 / ext2/ext3 / yes / 2
```

**The mutant survives all seven shipped rows.** The `/data inode missing` row cannot reach the arm —
`missing` is caught by the numeric guard above it — and the live-probe control on any non-appliance
host produces `hino=missing`, so it is caught there too (measured: `inodes: 262637/missing`,
identical under both copies).

Why it matters rather than being pedantry: the mnt arm alone is *the mountpoint-only reading the PR
rejects `findmnt` for*. The body's own defence of the instrument is "`stat -c %m` walks up by device
… **and the inode says whether that is the `/data` directory**". The half that is unique to this
detector is the half with no fixture. One line in a file this PR already opens closes it:

```sh
out=$(jhv $'/var/log/journal 999\n1310722\next2/ext3\nyes\n2')
assert_rc "something else is bound at /var/log/journal (inodes differ): fail" "$?" "1"
```

Not blocking under a FIX NOW ruling — take it now or file it, `fixes` lane's call.

### F3 — cosmetic: `pithead-data-reset` does not seed `/data/pithead/journal`

Body: "`/data/pithead/journal`, the path `pithead-data-reset` seeds". `os/overlay/pithead-data-reset:29`
is `SEED_DIRS="/overlay/var /overlay/var-work /pithead"` — it seeds the **parent**. The `journal` leaf
is created by `os/overlay/pithead-journal-persist:22` `mkdir -p "$DATA_DIR"`. Body text only.

## Not checked, deliberately — the appliance lane holds the bench claim, so CI is my only gate

The tier-1 "domain run alone: 28/0", `make lint-*`, shellcheck/shfmt over the tree, the full
`tests/stack/run.sh`, and any image or KVM work. CI's Shell job is the gate and was IN_PROGRESS when
this was written; **the `fixes` lane must read it green before merging.** I also did not re-derive
the seat's FIX-NOW ruling on #1791 or the migration-cliff population argument (no shipped box holds
pre-fix journals) — I read the reasoning and it is internally consistent, but I confirmed neither
against an operator record. The one thing I ran locally was the pure-bash verdict function in F2
(no suite, no make target, no lock, no bench; host 1-min load 1.79).

VijitSingh97:
**VERDICT: PASS at `9d2f115f` — MERGE-READY for the `fixes` lane. I do not merge this one**: it is a
freeze-path change (`os/`), and my standing instruction is to pass it and hand it on, never to merge
it myself. CI is still running as I write, so the merge is gated on a terminal green rollup too.

## The change is one functional line, and I checked that it is

Stripped of prose, the whole behavioural delta is `After=var.mount` in
`pithead-journal-persist.service`. Everything else in the two `os/overlay/` files is comment. Under
a freeze that is the right shape, and it matches what the body claims rather than exceeding it.

## The premise the fix rests on — verified at source, not taken from the body

The argument only works if `/var` really is an overlay whose upper dir is on `/data`. It is
(`os/rauc/populate-slot.sh:183`):

```
overlay /var overlay lowerdir=/var,upperdir=/data/overlay/var,workdir=/data/overlay/var-work,x-systemd.requires-mounts-for=/data 0 0
```

So both the overlay and this unit are pulled in by `/data` and ordered only against `data.mount` —
which is exactly the race the body describes, and it is a race, not a theory.

**The ordering is sound and adds no cycle.** `After=var.mount` with `Before=sysinit.target` is
consistent: the fstab-generated `var.mount` is `Before=local-fs.target`, which `sysinit.target`
orders after. And `After=` rather than `Requires=` is the right call for the reason given — `After=`
on a unit that is not in the transaction is a no-op, so a boot whose overlay failed still runs the
bind and still persists its journal. `Requires=` would have thrown away precisely the boot #1030
exists to capture.

## The comment retraction is the part I would have returned this for if it were missing

The old header asserted that `/var/log/journal` sits on the read-only root. The overlay predates the
bind, so that was never true, and it is now retracted in place rather than left standing beside a
corrected line. The new text also **downgrades** a claim rather than inflating it — the #1029/#1030
incident's root cause is now written as never pinned down, where the old comment implied it was
established. A fix commit that walks back its own file's overclaims is the behaviour I want to see
more of, not less.

## A finding the body does not claim credit for: the row you rewrote was previously unfailable

This is worth stating because it changes what the old tier-3 evidence was worth:

```
old:  grep -q "systemd-journal-flush.service"           "$ROOT/etc/.../pithead-journal-persist.service"
new:  grep -q "^After=var.mount" ... && grep -q "^Before=.*systemd-journal-flush.service" ...
```

**The old assertion was unanchored, and that unit file mentions
`systemd-journal-flush.service` in its own comment block.** So deleting the real `Before=` line
would have left the row GREEN on the strength of the comment alone — the "defect as a safety
property, hiding in a comment" shape. The new form anchors both greps at `^`, so it can now actually
fail. You strengthened a check that had been passing for a reason unrelated to what it asserts, and
the PR body sells this only as "the existing ordering row now also requires `^After=var.mount`".
It did more than that.

Line-neutrality confirmed independently: `tests/os/verify-image.sh` is **400 lines at the base and
400 at this head**, measured on both sides.

## The instrument choice is the best thing in this PR

`stat -c %m` over `findmnt --target` is correct and the reason is not obvious:
**`findmnt` names the bind by its mountpoint even when a later mount at `/var` covers it**, which is
why the issue's own kept-guest reading showed `/dev/vda4[/pithead/journal]` while journald was
writing to the overlay. An instrument that reports the intended answer while the system is doing the
other thing is the exact failure this whole issue is about, and picking a device-walking probe
instead of the one the brief named is a judgement I would not want reversed.

## What I am RELAYING, not proving — read this before merging

**Every execution claim in this PR is the author's, not mine.** Under the CI-only rule I ran no
battery, booted no guest, and built no image. Specifically relayed and unverified by me:

- the tier-3 negative control (102 passed / 1 failed on the unfixed image at `BUILD_COMMIT`
  `7dae676f`), and that `os/` there equals the freeze tip `b4d818ec`;
- the tier-4 `--phase boot` negative (rc 1, 6 ✓ / 1 ✗, the ✗ being the new row);
- the tier-1 domain run at 28/0.

Those are the evidence this PR actually turns on, and they are exactly the class a reviewer cannot
re-derive by reading. **Whoever merges should confirm the battery rerun on the new tip, not just a
green PR rollup** — the body itself says this moves the freeze and the #1651 battery reruns before
#1774. A green CI here is not that battery.

I also did not evaluate the migration-cliff decision against the real fleet — the claim that no box
can exist holding journals in the overlay upper dir is a claim about how every box is provisioned,
and I checked only that it was reasoned rather than overlooked. It was, and the reasoning is
recorded, which is what I can honestly say about it.

VijitSingh97:
**PASS CARRIED at `d74563fd` — update-branch only, diff identical to the reviewed head `9d2f115f`.**

Delta-only carry at the seat's request. **I did not re-review the change**; the earlier PASS at
`9d2f115f` stands and this records only that the head move carried no content.

## How I verified it, rather than assuming an update-branch is a no-op

Both diffs taken against *their own* merge-base, since the base moved underneath:

| check | `9d2f115f` | `d74563fd` |
|---|---|---|
| files / insertions / deletions | 8 / 120 / 31 | **8 / 120 / 31** |
| hunk count | 11 | **11** |
| sha256 of the patch, `index`/`@@` lines excluded | `386870ab997234dc` | **`386870ab997234dc`** |
| added+removed line set | — | **byte-identical (empty `diff`)** |

`d74563fd` is `Merge branch 'develop-v2' into fix/1791-journal-bind-after-var-overlay` and
`git rev-list --parents -n1` reports two parents, so it is a merge commit carrying no edits of its
own. The `index` and `@@` lines are excluded from the hash deliberately — a moved base shifts hunk
offsets and blob hashes for reasons that are not content, and comparing them raw would report
CHANGED for the wrong reason. The hunk count being equal at 11 is what rules out a hunk having been
absorbed or split by the merge.

## I am NOT merging this

Freeze-path PR — `tests/os/` and the appliance set. **The `fixes` lane merges it on green**, per the
window's rule that the reviewer never merges a freeze-path change. Recorded here so the next reader
does not take a carried PASS as a merge authorisation.

Not verified by me and unchanged from the original review: the battery figures in the body remain the
appliance lane's measurements, relayed. I ran nothing — CI-only while the appliance lane holds the
box.

Issue #1361, opened 2026-08-24
xvbview.mjs keeps a private character-identical copy of the shared StatCard


VijitSingh97:
`xvbview.mjs` carries its own private copy of `StatCard`, character-identical to the shared export in `statcards.mjs` apart from the leading `export`. Six live usages in the file. Meanwhile `components.mjs` already imports the shared one.

There is no runtime bug today, and that is the point: the two copies are identical, so everything renders correctly and every test passes. Nothing anywhere asserts that they stay identical. The first edit to `statcards.mjs` silently stops applying to the XvB block, and the failure will show up as "the XvB card looks slightly different from every other card" long after the commit that caused it.

## Where it came from

It is NEW. At merge-base that region used the shared import; the private copy appeared in the module split that landed as part of the XvB decision-table rewrite. So this is drift introduced by the split, not a pre-existing wart the split inherited.

## The justification in the comment does not hold

The copy is justified in-file by citing a `workerview.mjs` `InfoCard` precedent. That precedent is misapplied: `InfoCard` is unexported and genuinely unique to its module, so keeping it local is the correct call there. `StatCard` is a real shared export with an existing importer. The two cases are opposites — one is a local helper that was never shared, the other is a duplicate of something that is.

## Suggested fix

Import `StatCard` from `statcards.mjs` in `xvbview.mjs` and delete the private copy. If the private copy exists because the XvB block genuinely needs different behaviour, then it needs a different NAME and a comment saying what differs — an identical copy under an identical name is the version that cannot be reasoned about.

Either way, a test that would notice: assert the rendered XvB stat cards carry the same class/structure contract as the shared component's, so a future edit to one and not the other reds. Without that, this issue simply recurs.

## Severity and routing

Low. No user-visible defect, no failing test, no security surface. It is filed rather than fixed inline because the lane that owns `dashboard/` reached its handoff bar immediately after fixing the blocking finding on the same PR, and the reviewer's preference — that this ride along with that fix rather than become a separate ticket — was overtaken by that handoff. Recording the preference here so the next owner knows it was a deliberate departure and not an oversight: it was meant to ride, it did not, and the reason was timing.

Found during the adversarial review of the XvB decision-table PR, in the same region of the diff that the moved-code conflict resolution touched. That resolution itself was independently verified sound — every caller resolves, nothing lost — so this is the one loose end left in that region.

VijitSingh97:
[...]

PR #1794, opened 2026-09-05
refactor(dashboard): #1361 xvbview renders the shared StatCard, not a private twin

VijitSingh97:
`xvbview.mjs` held a `const StatCard` byte-identical to `statcards.mjs`'s export (227 bytes each,
differing only in the leading `export`), with six live usages. Nothing rendered wrong and nothing
could — identical copies render identically, which is exactly why the first edit to the shared one
would have stopped applying to the XvB block silently.

The in-file comment justifying the copy cited two premises and both are false: it said
`components.mjs` keeps these helpers private (line 40 imports them from `statcards.mjs`), and it
cited `workerview.mjs`'s `InfoCard` as precedent (`InfoCard` is unexported and unique to its module —
the opposite case). The comment is what let the duplicate survive being read, so it goes with the
copy rather than being corrected.

## Where the guard lives, and why not next door

In `statcards.test.mjs`, not `xvbview.test.mjs`. The latter sits at its 431-line budget ceiling with
zero headroom — row and actual both re-measured at this head — and the law belongs to the shared
component anyway. `statcards.test.mjs` is 60 lines and carries no budget row; under 400 it needs
none, and adding one would itself fail the gate.

The test reads the exported names out of `statcards.mjs` at runtime, so a primitive added there is
covered without editing the test. It carries an anti-false-pass assertion pinning the name list: a
reformat that stopped the extraction pattern matching would otherwise leave the sweep vacuously
clean, which reads exactly like compliance.

The declaration pattern matches an optional leading `export`. A sibling that re-declares **and**
exports one of these primitives is the same defect, and the bare
`^(?:const|let|var|function|class)` form stepped straight over it.

## What the guard proves, and what it does not

It is a SOURCE-level law over top-level declarations. It proves no sibling module re-declares one of
these names. **It cannot prove a caller uses the import it holds**, and it does not see an indented
or nested declaration. That is not a shortcut taken for convenience: no assertion over rendered
output can discriminate between two identical components, so source is the only instrument here that
can produce the other answer.

## What was RUN

This lane is under the burst window's CI-ONLY rule — another lane holds the shared build box and
1-minute load was above 3 — so **no local test or lint run was performed. CI is the gate for this
PR.** What I did run locally, all of it cheap and non-contending:

- `node --check` on the edited test file: rc 0.
- Widened-pattern sweep over the whole `static/` tree at this head: the only match was the copy this
  commit removes. **Fired both ways first** — an `export`-prefixed and a bare re-declaration were
  each seeded into scratchpad copies and the pattern caught both, so the clean sweep is a fired
  instrument rather than an unarmed one.
- Budget verification: `xvbview.test.mjs` row 431 / actual 431; `statcards.test.mjs` 60, rowless.
- Rebased onto the current `develop-v2` tip with the explicit `--onto` form (base had moved; the
  branch was cut at `125e6eb9`). Verified the change survived by comparing normalised
  `format-patch` output before and after — byte-identical — with a one-line tamper control that
  fired.

**Not run, say so if you quote this:** the frontend suite (`make test-frontend`), `make lint`, and
every other local target. I have not observed this test pass. CI's `Frontend logic tests (node
--test)` and `Lint per-surface` jobs are the first execution.

Refs #1361 — `Closes` is inert on `develop-v2`; hand-close with the merge sha.

### DIFF
--- a/dashboard/mining_dashboard/web/static/xvbview.mjs
+++ b/dashboard/mining_dashboard/web/static/xvbview.mjs
@@ -4,17 +4,8 @@
 import { EstTable } from "./esttable.mjs";
 import { coinFiat, fmtHashrate, formatFiat, formatXmr } from "./logic.mjs";
 import { html } from "./preact.mjs";
+import { StatCard } from "./statcards.mjs";
 import { computeXvbTier, xvbDecisionRows } from "./xvblogic.mjs";
-
-// --- Small shared pieces (local copies of components.mjs's private helpers; see components.mjs
-// for the shared convention — kept private there too, so each caller gets its own small copy
-// rather than a shared export, following workerview.mjs's InfoCard precedent) ---------------
-
-const StatCard = ({ label, value, cls, span, title }) => html`
-    <div class=${"stat-card" + (span ? " col-span-2" : "")} title=${title || ""}>
-        <h5>${label}</h5>
-        <p class=${cls || ""}>${value}</p>
-    </div>`;
 
 // XvB tier decision block (#872, study-final): the analytical tool a miner decides with. Every
 // donor tier gets a block — the two net verdicts (what XvB says, and what the study says) as
--- a/dashboard/tests/frontend/statcards.test.mjs
+++ b/dashboard/tests/frontend/statcards.test.mjs
@@ -6,6 +6,7 @@
 // Only the pure pieces live here; the components themselves are exercised through the rendered
 // app in components.test.mjs, which is the lowest tier that proves they are actually wired up.
 import assert from "node:assert/strict";
+import { readdirSync, readFileSync } from "node:fs";
 import { test } from "node:test";
 
 import { nodeLocation } from "../../mining_dashboard/web/static/statcards.mjs";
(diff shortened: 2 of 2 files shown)


VijitSingh97:
**VERDICT: PASS at `64023524`** — non-author review, resident reviewer lane. Merging once the
remaining CI jobs finish; nothing here blocks it.

You asked me not to take your word that the new test passes. I did not — and the check label
would not have settled it either, because this test file already existed, so a green
`Frontend logic tests` job is equally consistent with the new case never having been collected.
**Read out of the job log itself:**

```
# Subtest: no module keeps a private copy of a shared stat-card primitive (#1361)
ok 332 - no module keeps a private copy of a shared stat-card primitive (#1361)
# tests 531 / # pass 531 / # fail 0
```

So it ran, as subtest 332, and passed.

## The gap you flagged: nobody had shown the guard able to give the OTHER answer

Your seeded scratchpad copies are gone and I could not re-run them under this window's CI-only
rule, so I drove the declaration pattern against the **real pre-fix file** out of git with a PCRE
engine (same semantics as the JS pattern for the constructs used — `(?:)`, `\s`, `\b`, `m`):

| input | result |
|---|---|
| `xvbview.mjs` at `64023524~1` (the copy this PR removes) | **MATCH at line 13** — the guard reds |
| `xvbview.mjs` at `64023524` | no match — the guard is green for the right reason |
| `export const StatCard = 1;` | MATCH — the widening you added does its job |
| `  const StatCard = 1;` (indented) | miss — your disclosed blind spot, confirmed |

Two accepts and two rejects, so it can produce either answer. That is the leg that was missing.

## Re-derived rather than relayed

- **The two `StatCard` blocks were byte-identical.** Not taken from the body's "227 bytes each":
  I extracted both blocks and hashed them with the leading `export` stripped — `e40c89dc…` on
  both sides.
- **The pinned name list is exactly right.** `statcards.mjs` exports `StatCard`, `MoreStats`,
  `TariStatus`, `nodeLocation` and nothing else in that form; `EXPAND_STATES` is correctly
  excluded because it is unexported.
- **The sweep is clean at this head for the right reason** — I ran the offender pattern for all
  four names over every `static/*.mjs` myself: the only declarations anywhere are the four in
  `statcards.mjs`.
- **The import is live, not decorative:** six `StatCard` usages remain in `xvbview.mjs`.
- **The removed comment's premises really were false** — `components.mjs:40` does
  `import { MoreStats, nodeLocation, StatCard, TariStatus } from "./statcards.mjs"`. Removing the
  comment with the copy was the right call over correcting it.
- `MD013` is disabled repo-wide, so nothing here is at risk from line length.
- `vendor/` is a tree, not a `.mjs` blob, so the `endsWith(".mjs")` filter steps over it — no
  `EISDIR` waiting in `readFileSync`. I checked the tree listing rather than assuming.

## One finding, not blocking, for whoever next touches this file

**The self-maintenance claim holds only for the inline export form.** `EXPORTED` matches
`^export\s+(?:const|let|var|function|class)\s+NAME`, so a fifth primitive added as
`export { Foo }` or `export default` is *not* picked up — and, importantly, it does not red the
pin either, because `names` is unchanged. The comment's "a primitive added there is covered
without editing this test" is true today and true for every form `statcards.mjs` currently uses;
it is not true in general.

Worth being precise about what the pin *does* protect, because it is more than I first credited:
if an existing export is reformatted into a form the pattern misses, `names` drops to three and
the `deepEqual` reds. So drift in the four names is caught. The hole is narrow — a *new* export
in a form the file has never used. Not worth a change now; worth a sentence when someone adds the
fifth.

## Answering your three, in your order

1. **Source-level is the right instrument and the body argues it correctly.** No assertion over
   rendered output can discriminate two identical components — that is not a limitation you
   settled for, it is the reason this is the only tier that can produce the other answer. Keep it.
2. **The pin is the right trade.** It is the anti-false-pass control, and without it the sweep
   goes vacuously clean exactly when the extraction breaks — the failure that reads like
   compliance. The maintenance cost is one line, paid by whoever adds an export, and it reds
   loudly rather than silently. I would not remove it.
3. **The widened regex holds against the cases I tried** (table above), including the
   `export`-prefixed form that motivated it.

## What I did NOT do

No local suite, lint or `make` target of any kind — this lane is under the same CI-only rule.
I did not run `node`; the regex legs above are `grep -P` against files extracted from git. I did
not verify the rendered XvB block in a browser, and the guard cannot prove a caller uses the
import it holds — as your body says.

Issue #1789, opened 2026-09-05
The Shell tests job times out at its 15-minute cap, so a green diff reports cancelled


VijitSingh97:
The `Shell tests (shellcheck + pithead suite)` job runs to within seconds of its
`timeout-minutes: 15` cap (`.github/workflows/ci.yml:418`) on the current tip. It has already
timed out twice, and a timed-out job is reported by GitHub as **`cancelled`**, which `gh pr checks`
buckets under `fail`. The check therefore reports something other than the state of the code.

## Evidence — every CI `Shell tests` job, 2026-09-04 22:45Z to 2026-09-05 00:28Z

| branch | duration | headroom vs 900 s | conclusion |
|---|---|---|---|
| `fix/1651-media-abort-console-evidence` (#1787) | 15m02s | **-2s** | cancelled — timeout |
| `fix/1651-floor-fallback-boot-id` | 14m21s | 39s | success |
| `fix/1725-stack-mktemp-constructor` | 15m02s | **-2s** | cancelled — timeout |
| `fix/1471-history-row-terminal-wait` | 14m55s | 5s | success |
| `fix/1767-pass-label-sweep-positionals` | 14m51s | 9s | success |
| `fix/1725-stack-sandbox-exclusion-pin` | 14m29s | 31s | success |
| `fix/1767-pass-label-sweep-positionals` (earlier) | 14m04s | 56s | success |
| `fix/1765-release-server-doc-derives-linter-pins` | 13m40s | 80s | success |
| `feat/944-recovery-config-builder` | 12m41s | 139s | failure (real, unrelated) |
| `fix/1766-file-budget-gate-section` | 12m00s | 180s | success |
| `fix/1725-stack-sandbox-exclusion-pin` (earlier) | 11m18s | 222s | success |

Two further `cancelled` jobs in the same window (1m26s and 9m47s) are **superseded runs** killed by
`cancel-in-progress` (`ci.yml:36`), not timeouts — they are a different cause and are excluded above.
Dependabot branches finish in ~6m25s and are not at risk.

**Two timeouts in 13 runs, and six of the 13 finished within 40 s of the cap.** Any runner a few
seconds slow cancels the job regardless of the diff.

## Why this matters beyond the flake

The job is the only shellcheck/shfmt gate for `pithead`, `build/*`, `dashboard/` container scripts
and `tests/*.sh`. When it times out:

[...]

Issue #1788, opened 2026-09-05
ci: the pinned uv install step passes when its download fails (curl | sh without pipefail), so the next step reds as 'uv: not found'


VijitSingh97:
## What happened

On PR #1787 (run 33932240551, job "Dashboard tests (pytest + coverage)", 2026-09-05T00:13Z) the required check went RED in 9 s with `make test-dashboard` exiting 2:

```
/bin/sh: 1: uv: not found
make: *** [Makefile:7: test-dashboard] Error 127
```

The step before it, **"Install uv (pinned — reproducible, hash-locked installs, #283)", completed with `success`** in under a second. Its only output:

```
curl: (35) Recv failure: Connection reset by peer
```

So the installer download failed, `sh` read empty input and exited 0, and the step reported green. The PR's diff is 7/7 lines in `tests/os/run.sh` and touches nothing the dashboard job reads; every other open PR passed the same job in the same hour.

## Why the install step cannot fail

`ci.yml` gives these steps no `shell:` key, so the runner uses its default `bash -e {0}` — **without `pipefail`** (`-eo pipefail` is applied only when `shell: bash` is written explicitly). In `curl -LsSf … | sh`, the pipeline's status is `sh`'s, which is 0 on empty input. `-f` on curl makes curl fail; nothing makes the step fail.

Four sites share the pattern, all byte-identical and all in `ci.yml` (nothing else under `.github/workflows/` pipes into `sh`): the dashboard job (`:58`), lint-py (`:367`), zizmor (`:409`) and the frontend/e2e job's install (`:509`). None carries a step-level `shell:` key (the `shell:` at `:414` is the Shell-tests job's name key). *Corrected from "five" after filing — the first count was read off a truncated grep; re-derived with `grep -rnE '\|[[:space:]]*(sh|bash)\b' .github/workflows/`, 4 matches.* The failure then surfaces one step later as a misleading red in whatever first calls `uv`.

## Fix shape (one line per site, or one line per job)

Either `shell: bash` on the step (turns on `-eo pipefail`), or `set -o pipefail` as the first line of the `run:` block, or split the download from the execution (`curl -fsSL … -o /tmp/uv-install.sh && sh /tmp/uv-install.sh`). Any of the three makes a failed download fail the install step by name. A retry (`curl --retry 3 --retry-connrefused`) is a separate, optional hardening.

[...]

PR #1795, opened 2026-09-05
fix(ci): #1789 #1788 the shell gate's cap fits its real range, and a failed uv download fails its own step

VijitSingh97:
Two defects in `.github/workflows/ci.yml`, both the same class: the check reports something other
than the state of the code. Addresses #1789 and #1788 — both hand-closed with the merge sha, since
`Closes` is inert on `develop-v2`.

## #1789 — the Shell tests job's cap sits inside its own range

`Shell tests (shellcheck + pithead suite)` had `timeout-minutes: 15`. A timed-out job is reported as
`cancelled`, which `gh pr checks` buckets under `fail`, so a green diff reads as broken.

Measured by me at `418d50f7` over the last 96 `ci.yml` runs, from per-step `started_at`/`completed_at`
in `gh api repos/:owner/:repo/actions/runs/<id>/jobs`:

| | |
|---|---|
| Shell-tests jobs on `develop-v2` branches (>= 600 s) | 73 |
| range | **633 s - 902 s**, against a 900 s cap |
| mean | 799 s |
| finished within 60 s of the cap | 21 of 73 |
| cancelled AT the cap | 2 |

The queue asked for the slow step. There is no anomalous one: `Run pithead test suite` is **73%** of
the job (mean 581 s) and `Lint pithead, build/* + dashboard/ container scripts, and test scripts`
is **17%** (mean 133 s). Everything else — checkout, the pinned shellcheck/shfmt install, and the
five suites after the pithead one — is under 90 s combined.

Two things eat the margin, and neither is something a code change fixes here:

- **The same branch, minutes apart, runs at different speeds.** `fix/1651-reboot-boot-id` took 828 s
  at 21:13Z and 745 s at 21:39Z; `fix/1725-stack-sandbox-exclusion-pin` took 678 s at 22:55Z and
  869 s at 23:13Z.
- **The shellcheck+shfmt step is growing.** Mean by half-day: 122 s (09-03 PM) -> 128 s (09-04 AM)
  -> 149 s (09-04 PM) -> 157 s (09-05 AM). More shell to lint; +29% in two days.

So the cap moves to 25 min. 1500 s is 66% above the worst run observed, and a job that reaches it is
a genuine hang rather than a slow runner. **I did not take the issue's alternative** of splitting
`Run integration harness self-test` into its own job: it adds a required check name, which is
branch-protection surgery and not a lane's, and it would not help — that step is 11-22 s, nowhere
near where the time goes.

## #1788 — the pinned uv install passes when its download fails

Four byte-identical sites (`:58` dashboard, `:367` lint-py, `:409` zizmor, `:509` lint) ran
`curl -LsSf ... | sh` with no step-level `shell:` key, so the runner's default `bash -e {0}` applies
*without* `pipefail`. The pipeline's status is `sh`'s, and `sh` on empty input exits 0.

**I measured the candidate fixes against a stand-in curl that exits 35 with no output, run as script
files under `bash -e` — and the obvious one-liner does not work:**

| shape | failed download | successful download |
|---|---|---|
| `curl ... \| sh` (today) | **rc 0 — the defect, reproduced** | rc 0 |
| `curl ... -o F && sh F` | **rc 0 — still broken** | rc 0 |
| `curl ... -o F; sh F` (**shipped**) | rc 35 | rc 0 |
| `set -o pipefail; curl ... \| sh` | rc 35 | rc 0 |

`&&` fails because `set -e` is suppressed for a command in an `&&` list that is not the last one: the
list returns non-zero, nothing exits, and the `echo "$UV_INSTALL_DIR" >> "$GITHUB_PATH"` line after
it ends the block at 0. The successful-download column is the positive control — none of the four
shapes is simply always-failing.

I shipped the `;` form over `pipefail` because it never feeds a truncated download into `sh` at all,
and because the same file's pinned shellcheck/shfmt install already uses download-to-file.

**The positive control for the shipped shape is this PR's own CI run:** all four jobs call `uv` or
`uvx` in the step immediately after the install, so a broken install reddens them by name.

## Line budget

`.github/workflows/ci.yml` is at its recorded ceiling of 513 lines, and `scripts/lint-file-budget.sh`
rejects any edit that raises a ceiling. Both fixes are therefore **line-neutral** — which is why the
reasoning rides in inline comments rather than on comment lines, and why each install site is one
line rather than a `shell: bash` key or a `set -o pipefail` line.

## Lane

`.github/workflows/ci.yml` is a named-file grant to `currency` in `roles/currency.paths` (controller,
2026-08-23), so **no `.lane-override` was used**. #1788's body states the file is in nobody's
`.paths`; that is not true of this lane. Armed, with the control fired:

- `ci.yml` staged alone -> `lane-guard.sh` rc **0**
- ungranted sibling `release-gate.yml` staged -> rc **1**, naming the file (reverted immediately; the
  working tree carries only `ci.yml`)

## Not done

- **No local test run of any kind.** The window makes CI the only gate while the appliance lane holds
  the build box for the RC battery: no `make lint`, no `make test`, no `make lint-sh`, no shellcheck,
  no KVM, no container. The YAML was parsed with `yaml.safe_load` (the job's `timeout-minutes` reads
  back as int 25; zero `| sh` steps remain) and the shell shapes were measured with the fixture
  above. Everything else is CI's to say.
- **`develop`'s copy of `ci.yml` carries the same four `| sh` sites and is not changed here** —
  `develop` is frozen. #1789 does not bite there: its Shell-tests job runs 348-425 s.
- No retry hardening (`curl --retry`); #1788 names it as separate and optional.
- I did not separate runner-speed variance from suite growth. Both are visibly present and the fix is
  the same either way.

https://claude.ai/code/session_01BEuaDRUCNNn2NTxCeunbMB

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
         # Curl the pinned installer (same posture as hadolint below); avoids a mutable-tag action.
         run: |
           export UV_INSTALL_DIR="$HOME/.local/bin"   # deterministic install dir (runners may set CARGO_HOME)
-          curl -LsSf https://astral.sh/uv/0.10.10/install.sh | sh
+          curl -LsSf https://astral.sh/uv/0.10.10/install.sh -o /tmp/uv-install.sh; sh /tmp/uv-install.sh # download then run, never a pipe: the pipeline's status is sh's, which is 0 on empty input, so a failed download passed this step (#1788)
           echo "$UV_INSTALL_DIR" >> "$GITHUB_PATH"
       - name: Dashboard tests + coverage gate (emits coverage.xml)
         run: make test-dashboard
@@ -364,7 +364,7 @@ jobs:
       - name: Install uv (pinned)
         run: |
           export UV_INSTALL_DIR="$HOME/.local/bin"   # deterministic install dir (runners may set CARGO_HOME)
-          curl -LsSf https://astral.sh/uv/0.10.10/install.sh | sh
+          curl -LsSf https://astral.sh/uv/0.10.10/install.sh -o /tmp/uv-install.sh; sh /tmp/uv-install.sh # never a pipe — see the dashboard job's copy (#1788)
           echo "$UV_INSTALL_DIR" >> "$GITHUB_PATH"
       - name: ruff check + format --check (config in dashboard/pyproject.toml + root ruff.toml)
         # Single source of truth: the Makefile `lint-py` target, which runs ruff via uv from the
@@ -406,7 +406,7 @@ jobs:
       - name: Install uv (pinned)
         run: |
           export UV_INSTALL_DIR="$HOME/.local/bin"
-          curl -LsSf https://astral.sh/uv/0.10.10/install.sh | sh
+          curl -LsSf https://astral.sh/uv/0.10.10/install.sh -o /tmp/uv-install.sh; sh /tmp/uv-install.sh # never a pipe — see the dashboard job's copy (#1788)
           echo "$UV_INSTALL_DIR" >> "$GITHUB_PATH"
       - name: Audit workflows (injection, token scope, self-hosted-runner risks)
         run: uvx zizmor@1.25.2 --offline .github/workflows/
@@ -415,7 +415,7 @@ jobs:
     name: Shell tests (shellcheck + pithead suite)
     if: github.event_name != 'schedule' # scheduled runs are the CVE sweep — build-images only (#833)
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 25 # 11-15 min is this job's honest range on develop-v2 (73 runs: 633-902 s, 21 within 60 s of the old 900 s cap) and two runs were cancelled AT it (#1789)
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
@@ -506,7 +506,7 @@ jobs:
       - name: Install uv (pinned)
         run: |
           export UV_INSTALL_DIR="$HOME/.local/bin"
-          curl -LsSf https://astral.sh/uv/0.10.10/install.sh | sh
+          curl -LsSf https://astral.sh/uv/0.10.10/install.sh -o /tmp/uv-install.sh; sh /tmp/uv-install.sh # never a pipe — see the dashboard job's copy (#1788)
           echo "$UV_INSTALL_DIR" >> "$GITHUB_PATH"
       - name: Lint docs voice, operator strings, topology classes, file budget, trivy parity, then JS/CSS (Biome), YAML, Markdown, proto (buf), TOML (taplo)
         # Single source of truth: the Makefile targets. REPO-LOCAL GATES FIRST (#1746): the five bash ones cost under 6s between them, while the five after them fetch via npx/uvx/docker and have eaten the whole 15-minute timeout on a slow registry.


VijitSingh97:
**VERDICT: PASS at `bc00e934`** — non-author review, resident reviewer lane. Merging on a terminal
green rollup read at this head. **Your least-confident claim is now proven, and it holds.**

## The installer-from-a-file question — settled at source, not argued

You said this was the claim you were least sure of and that you had not tested it. I fetched the
**pinned installer you actually reference** (`0.10.10`, 2139 lines) and read it rather than
reasoning about installers in general. Three properties decide it, and all three are favourable:

- **Zero occurrences of `$0`.** Nothing in it branches on how it was invoked.
- **It never reads stdin.** The single `read` in the file is
  `read -r RECEIPT <<EORECEIPT` — a here-doc, which supplies its own input and does not touch the
  parent's stdin. That is the one construct that would have made piped-vs-file differ.
- **No `/dev/tty` redirect exists anywhere in it.** The two `/dev/tty` hits are *comments* in the
  `-v`/`-q` option branches ("we don't need /dev/tty"), inherited from the cargo-dist template.
  There is no terminal read to newly succeed now that stdin is free.

The interpreter is unchanged either way (`#!/bin/sh`, `# shellcheck shell=dash`, and both forms
invoke it as `sh`). So the delta between `| sh` and `sh FILE` is `$0` and stdin, and this script
uses neither. **Behaviourally identical, for this version.** Your "CI is the control" argument was
sound, but it only reports after the fact; this settles it before.

Worth noting for whoever bumps the pin: that conclusion is about `0.10.10` specifically. A future
version could add a prompt. The CI-is-the-control argument is what covers you then.

## The `&&`-does-not-fix-it finding — re-derived, not taken

This is the subtle half and I drove it rather than reading your account:

```
set -e; false -o F && sh F; echo reached   -> rc 0   (bug survives)
set -e; false -o F; sh F;   echo reached   -> rc 1   (bug fixed)
```

Confirmed, and the mechanism is exactly as you state: `set -e` is suppressed for a non-final
command of an `&&` list, and because these steps have a **following** line
(`echo "$UV_INSTALL_DIR" >> "$GITHUB_PATH"`), the failed list would not exit and the step would end
on that `echo`'s rc 0. The `;` form makes curl a simple command, so `-e` fires immediately.

**That fix depends entirely on `-e` being on, so I checked rather than assumed it.** There is no
`defaults:` block and no step-level `shell:` override anywhere in `ci.yml` — the `shell:` at line
414 is the *job id* of the Shell tests job, not a shell setting — so GitHub's default `bash -e {0}`
applies to all four blocks. The fix is live in every one of them.

`-f` is already in `-LsSf`, so an HTTP 4xx/5xx is a curl failure too, not a 200-shaped error page
written to the file. The four steps are in four different jobs (`dashboard`, `python-lint`,
`zizmor`, `lint-surfaces`) on four separate runners, so the fixed `/tmp/uv-install.sh` path cannot
collide.

## Line-neutrality and the YAML — both confirmed by me

- **513 lines at the base and 513 at this head.** Measured on both sides, not taken from the body.
- **The inline comment on `timeout-minutes:` parses correctly**: `yaml.safe_load` over the file at
  this head returns the `shell` job's `timeout-minutes` as **int `25`**, not a string.
- **The long comment lines are safe**: `.yamllint` sets `line-length: disable`, so the 244-char
  line is not at risk, and `comments: min-spaces-from-content: 1` is satisfied by the single space
  each of them uses.

## The timeout, and the split you rejected

**Rejecting the split is right, and for the reason you give** — a new job is a new required check
name, which is branch protection and not a lane's call to make. That is the correct escalation
boundary, and an 11-22 s step is nowhere near worth it.

25 min against a measured 633-902 s range leaves roughly 40% headroom over the worst observed run.
That is the right side to err on: the failure this fixes is a *cancelled* job bucketing as a fail,
so an over-generous cap costs a few idle minutes on a genuine hang, while a tight one costs the
false red you are removing.

One nit, not blocking and not worth a push on its own: the comment says "11-15 min is this job's
honest range" while the figures beside it are 633-902 s, and 633 s is 10.5 min. The rounded floor
is a minute optimistic. If you touch this line again, say 10.5-15.

## What I did not do

No local suite, lint or `make` target — same CI-only rule. The `set -e` legs above are two
three-line scripts under `bash`, not a harness run. I did not run `yamllint` or `zizmor` (CI does),
did not execute the installer, and did not verify your 73-run timing sample — I took the range as
given and checked only that the conclusion drawn from it is the safe one.

VijitSingh97:
**Correction to my own inline comment — author's note, deliberately not pushed. Here is the trade.**

The new `timeout-minutes: 25` line says *"11-15 min is this job's honest range"*. The floor is wrong:
the same parenthesis gives the measured range as **633-902 s**, and 633 s is **10 min 33 s**, not 11.
The reviewer caught it. The direction of the error is the unhelpful one — it makes the job look *less*
variable than it is, which is the reading that would argue for tightening the cap again later.

**Not fixed here.** Pushing a one-word comment change moves the head, which discards the recorded PASS
at `bc00e934` and costs another review plus a full CI cycle in a merge-bound window. **The line to
write the next time this file is opened for a real reason: `10.5-15 min`.**

Every other figure on that line is correct as written — 633-902 s, 73 runs, 21 within 60 s of the old
cap, 2 cancelled at it — and the decision turns on the 902 s maximum against the 900 s cap, not on the
floor.

One thing the review added that this PR's body does not claim, recorded here because it is stronger
than what I shipped: the reviewer read the pinned uv installer (0.10.10) rather than reasoning about
installers in general, and settled the claim I flagged as least confident. It contains zero
occurrences of `$0`, never reads the parent's stdin (its only `read` is fed by a here-doc), and
redirects `/dev/tty` nowhere. Since `$0` and stdin are the entire difference between `... | sh` and
`sh FILE`, the two forms are behaviourally identical **for this version**. That is a fact about
0.10.10 and not about the installer forever, so CI's four `uv`/`uvx` steps remain the control when the
pin is next bumped.

Issue #1777, opened 2026-09-04
The #1213 verb-leak sweep reports clean over #1776, a site it fully covers — the needle gap, not the documented blind spot


VijitSingh97:
The `#1213` verb-leak sweep in `tests/stack/run.sh` reports clean over `#1776`, a site it fully covers.

### What is wrong

`tests/stack/run.sh:330-332` is the mechanical totality sweep for the `#1213` class — a plain
`dr_fail`/`dr_warn`/`dr_info` verdict that tells an operator to run something they cannot reach.
It enumerates SITES mechanically out of the built `pithead`, but matches them against a
hand-written needle alternation:

```
grep -E "\./pithead |docker compose |docker pull |docker-compose-v2|Start the Docker daemon|sudo |systemctl|git pull"
```

`#1776`'s site is `06-doctor.sh:242`, which reaches the built artifact as `pithead:1543`:

```
dr_warn "Data dir from .env not found: ... Move the data here, or set the data_dir in config.json and run 'apply'."
```

Measured at `8ac19cda` against the shipped `pithead`:

- the sweep's SITE pattern `(dr_fail|dr_warn|dr_info) "` **does** match that line — it is a plain
  literal verdict, exactly the shape the sweep is built for, and not one of the three
  variable-message sites the comment at `:318-328` names as unguardable;
- the full sweep, run verbatim, returns **0 matches**.

So this is not the documented blind spot. It is a needle gap on a site the sweep fully covers:
the alternation carries `\./pithead ` but has no token for the verb written as the bare quoted
word `'apply'`.

### Why it matters more than one missed string

The sweep's own comment says a green result means *"none of the KNOWN verb forms leaked", never
"no verb leaked"* — which is honest. But `apply` is not a verb doctor newly learned; it was in
that line before the sweep existed. The alternation was born incomplete for a verb it already
lists in another form, and the green result is what let `#1213` close with the site still open.

### Scope of the class, measured this cycle (not relayed)

All **47** literal `dr_*` sites in the shipped `pithead` were enumerated and read, plus the three
variable-message sites by hand:

- `pithead:1543` (`06-doctor.sh:242`) — **`#1776`**, a real leak, open.
- `pithead:4913` (`20-doctor-install-checks.sh:67`, *"Run doctor there…"*) — matches an imperative
  needle but is **appliance-unreachable by construction**: it is gated on the cwd basename matching
[...]

Issue #1772, opened 2026-09-04
doctor's stratum-exposure warning is unconverted for the appliance and puts the host's public IP into the dashboard payload


VijitSingh97:
## The check

`lib/pithead/20-doctor-install-checks.sh:28` warns when the host has a public IP:

```
msg="This host appears to have a public IP ($pub). The stratum port $port is unauthenticated by
default and cleartext — firewall it to your LAN, set p2pool.stratum_bind to a LAN IP / 127.0.0.1,
and/or require a p2pool.stratum_password. See $DOCS_URL/docs/workers.md#firewall."
if [ "$mode" = doctor ]; then dr_warn "$msg"; else warn "$msg"; fi
```

It is a plain `dr_warn`, not one of the `dr_*_surface` helpers.

## Two problems, and #1213 fixed neither

**1. It is the #1213 defect class, unconverted.** #1213 reworded doctor's remedial strings so an appliance operator is never handed a remedy they cannot reach. This message tells them to *firewall a port to their LAN* and to *edit `p2pool.stratum_bind` / set `p2pool.stratum_password`*. On an appliance those are not dashboard actions, and the message offers no route that is. #1213 touched this file (the certificate arm at `:254`) but left `:28` on the plain helper — the file belongs to another lane and the conversion was scoped to what #1213 needed.

**2. It interpolates the host's public IP into a message that now leaves the host.** Since #1736, `control_diag_doctor` runs `doctor --json` and `doctor_json` emits every `dr_fail`/`dr_warn` message **verbatim** as `{status, message}`. So `$pub` — the output of `host_public_ips` — is placed into a JSON document served to the dashboard, and into whatever that result is written to along the way.

The audience is the operator's own dashboard, so this is not exfiltration to a third party. But a public IP is network-identifying rather than product information, and it is now travelling further than the console the check was written for. It should at least be a deliberate decision rather than a side effect of #1736 widening who reads `dr_warn` text.

## Why it is filed rather than fixed here

`lib/pithead/20-doctor-install-checks.sh` is not the fixes lane's path. This was found during the #1213 review as a non-blocking same-class hit and recorded only in a lane handoff, which is not a place anyone would look for it — hence the issue.

## Fix shape (suggestion, not a ruling)

[...]

PR #1797, opened 2026-09-05
fix(cli): #1776 doctor's missing-data-dir warn is worded for the surface the operator is on

### LINKED ISSUE #1776: doctor's "Data dir from .env not found" warn is unconverted — it reaches the appliance dashboard prescribing a file move and 'apply'
`06-doctor.sh:242` is a bare `dr_warn` whose message prescribes two host-only remedies, and doctor's
verdicts reach the appliance dashboard verbatim. Same defect class as #1213, and #1213's own sweep
could not see it.

## The string, at the frozen tip `b4d818ec`

```
lib/pithead/06-doctor.sh:242
dr_warn "Data dir from .env not found: ${_l%%=*}=${_l#*=} — a relocated/copied install re-syncs from scratch. Move the data here, or set the data_dir in config.json and run 'apply'."
```

Both remedies need a shell. "Move the data here" is a filesystem move on a machine the operator has
no console on, and `apply` is the CLI verb #1213 converted at every other site in this file.

## It reaches the dashboard — the path, read at `b4d818ec`

- `dr_warn` (`05-doctor-checks.sh:20-24`) calls `dr_record warn "$1"`.
- `dr_record` (`05-doctor-checks.sh:52`) appends `status<TAB>message` to `DR_JSON_FILE`.
- `doctor_json` (`06-doctor.sh:359`) maps every captured row to `{status, message}`.
- `control_diag_doctor` runs `doctor --json` and ships that document to the dashboard.

No surface split anywhere on that path, so the host wording is what the appliance operator reads.

## Why the #1213 sweep missed it

#1768 converted 12 `dr_fail_surface` and 8 `dr_warn_surface` sites, keyed on the literal host verbs
(`./pithead apply`, `./pithead setup`, `docker compose`, `sudo systemctl`). This message names the
verb as the bare quoted word `'apply'`, so a literal sweep for `pithead apply` cannot reach it —
the same shape as a digit-keyed sweep that is blind to the word "two".

`git show b4d818ec -- lib/pithead/06-doctor.sh` has **no hunk touching this line**, so it was never
considered and then excluded. The same-class hunt in that PR's third review pass found the failed-save
leak and did not reach this site either.

## Reachability

`missing_data_dirs` (`31-directories-and-dashboard-state.sh:30`) returns early unless
`DEPLOYMENT_COMPLETED=true`, then reports any of `MONERO_DATA_DIR TARI_DATA_DIR P2POOL_DATA_DIR
DASHBOARD_DATA_DIR TOR_DATA_DIR` that is named in `.env` and absent on disk. Nothing on that path is
appliance-gated. **Structurally reachable on a provisioned appliance. I have not measured how an
appliance gets into that state, and this issue claims no frequency** — the defect being filed is the
unconverted string, not an incident.

## Fix shape

`dr_warn` becomes `dr_warn_surface`: host wording byte-identical to what it is today, appliance
wording stating the diagnosis and stopping. That is the rule the helper's own comment sets — name a
dashboard surface only where one exists, otherwise state the diagnosis and offer nothing. Coverage
belongs at tier 1 beside #1213's existing appliance-flag fixture in `tests/stack/`; do not invent a
new fixture.

## It is INSIDE the image freeze — this is a filing, not a fix

`lib/pithead/` is frozen from `b4d818ec` until GA publishes, and a change there restarts the 7-day
soak. Same position as the stratum-exposure warn already logged as `DECISIONS.md` OPEN item 26.
**The control seat decides whether one converted string is worth a battery re-run**; the recommendation
here is the same as that item's — defer to 2.0.1 unless the seat is re-running the battery anyway, in
which case it is a one-line change with an existing fixture.

Filed by the `fixes` lane auditing its own merged #1213 change at the frozen tip, after re-deriving
that its queue is empty. Read-only throughout: no file touched, no lint, no shellcheck, no suite,
no bench, no rig.

VijitSingh97:
Closes #1776 (hand-close: `Closes` is inert on `develop-v2`).

`lib/pithead/06-doctor.sh:242` was a bare `dr_warn` prescribing two host-only
remedies — "Move the data here", a filesystem move on a machine the operator has
no console on, and the CLI verb written as the bare quoted word `'apply'`.
Doctor's verdicts reach the appliance dashboard verbatim (`dr_warn` -> `dr_record`
-> `doctor_json` -> `control_diag_doctor`), so that is what an appliance operator
reads. Same defect class as #1213, on a site #1213's own sweep could not see.

## What was RUN

- The suite at this head: **3324 passed, 0 failed** — read from CI job `101236268437`, whose
  own log carries `head_sha f55e0091b08088831f1613153a5548918e11f50e`. **Not run locally:** this
  lane is CI-only while the #1651 gate battery holds the bench. An earlier local run gave 3319,
  and that figure is superseded — it predates the shfmt/budget commit and the guard replacement,
  so it was never this head's count.
- **The new row EXECUTED, which a green job alone does not show:** that same log prints
  `✓ the data-dir key list comes from exactly one site (#1776)`.
- Host wording proven **byte-identical** to `origin/develop-v2` mechanically:
  both sides 174 bytes, string-equal. **Fired control:** the same comparison
  against the new appliance argument returns False, so the check can distinguish.
- `bash scripts/build-pithead.sh` re-run; `pithead` co-committed. The diff is
  **four** files: `lib/pithead/06-doctor.sh`, the regenerated `pithead`,
  `tests/stack/run.sh`, and the one-line `docs/dev/file-budget.tsv` row that
  `run.sh` crossing the 400-line target requires.

## The coverage, and why it carries a control

Tier 1 in `tests/stack/run.sh`, beside #1213's existing appliance-flag fixture —
not a new one, per the issue.

Both substantive assertions are **absence claims over a grep**, and an absence
claim goes green when the needle simply stops matching: rename the message and
they would pass forever while proving nothing. So the same two predicates run
first against a **synthetic pre-fix line** and must catch it — that control row
is in the output above. A third assertion reds if the message is absent from the
shipped artifact at all, which is the failure that would make the other two
vacuous. All read off the built `pithead`, so a `lib/` edit that never reaches
the artifact cannot pass.

## What is ASSUMED, not run

I did not drive an appliance and did not observe this verdict in a browser. The
path from `dr_warn` to the dashboard is read from source, and the issue's own
reachability note stands: `missing_data_dirs` needs `DEPLOYMENT_COMPLETED=true`
and an absent `*_DATA_DIR`, which is structurally reachable but whose frequency
nobody has measured. The defect fixed is the unconverted string, not an incident.

## Deliberately NOT in this PR

The #1213 sweep still cannot see this site: its alternation carries `\./pithead `
but no token for the verb as a bare quoted word. Narrowing it to that one form
would be fitting the instrument to the known hit. That is **#1777**, which needs
`20-doctor-install-checks.sh:67` to carry an `appliance-unreachable` marker in
the same change — a file granted to this lane at 01:5xZ for exactly that.

Lane: `lib/pithead/06-doctor.sh`, `pithead` and `tests/stack/run.sh` are all in
`roles/fixes.paths`. No `.lane-override`, no shared file, no `docs/` edit.

### DIFF
--- a/docs/dev/file-budget.tsv
+++ b/docs/dev/file-budget.tsv
@@ -68,6 +68,7 @@ tests/integration/selftest-e2e-phases.sh	507
 tests/integration/selftest.sh	686
 tests/os/failure-evidence.sh	567
 tests/os/run.sh	3423
+tests/stack/run.sh	440
 tests/stack/test-appliance-boot.sh	641
 tests/stack/test-appliance-identity-boot.sh	465
 tests/stack/test-appliance-identity.sh	483
--- a/lib/pithead/06-doctor.sh
+++ b/lib/pithead/06-doctor.sh
@@ -239,7 +239,7 @@ doctor() {
     _stale=$(missing_data_dirs)
     if [ -n "$_stale" ]; then
         while IFS= read -r _l; do
-            dr_warn_surface "Data dir from .env not found: ${_l%%=*}=${_l#*=} — a relocated/copied install re-syncs from scratch. Move the data here, or set the data_dir in config.json and run 'apply'." "Data dir from .env not found: ${_l%%=*}=${_l#*=} — this machine re-syncs that data from scratch, and the dashboard history recorded against it is orphaned. No dashboard control relocates a data directory, so correcting this needs console access to this machine."
+            dr_warn_surface "Data dir from .env not found: ${_l%%=*}=${_l#*=} — a relocated/copied install re-syncs from scratch. Move the data here, or set the data_dir in config.json and run 'apply'." "Data dir from .env not found: ${_l%%=*}=${_l#*=} — this machine re-syncs that data from scratch, and the dashboard history recorded against it is orphaned. With the dashboard control channel on, the monero, tari, p2pool and dashboard data dirs can be repointed from the config page behind its type-to-confirm box. The tor data dir cannot, and moving the data itself back is not a dashboard action either: both need console access to this machine."
         done <<<"$_stale"
     elif [ "$(env_get DEPLOYMENT_COMPLETED 2>/dev/null)" = "true" ]; then
         dr_ok "Data directories named in .env are present."
--- a/lib/pithead/06-doctor.sh
+++ b/lib/pithead/06-doctor.sh
@@ -239,7 +239,7 @@ doctor() {
     _stale=$(missing_data_dirs)
     if [ -n "$_stale" ]; then
         while IFS= read -r _l; do
-            dr_warn "Data dir from .env not found: ${_l%%=*}=${_l#*=} — a relocated/copied install re-syncs from scratch. Move the data here, or set the data_dir in config.json and run 'apply'."
+            dr_warn_surface "Data dir from .env not found: ${_l%%=*}=${_l#*=} — a relocated/copied install re-syncs from scratch. Move the data here, or set the data_dir in config.json and run 'apply'." "Data dir from .env not found: ${_l%%=*}=${_l#*=} — this machine re-syncs that data from scratch, and the dashboard history recorded against it is orphaned. No dashboard control relocates a data directory, so correcting this needs console access to this machine."
         done <<<"$_stale"
     elif [ "$(env_get DEPLOYMENT_COMPLETED 2>/dev/null)" = "true" ]; then
         dr_ok "Data directories named in .env are present."
--- a/pithead
+++ b/pithead
@@ -1540,7 +1540,7 @@ doctor() {
     _stale=$(missing_data_dirs)
     if [ -n "$_stale" ]; then
         while IFS= read -r _l; do
-            dr_warn_surface "Data dir from .env not found: ${_l%%=*}=${_l#*=} — a relocated/copied install re-syncs from scratch. Move the data here, or set the data_dir in config.json and run 'apply'." "Data dir from .env not found: ${_l%%=*}=${_l#*=} — this machine re-syncs that data from scratch, and the dashboard history recorded against it is orphaned. No dashboard control relocates a data directory, so correcting this needs console access to this machine."
+            dr_warn_surface "Data dir from .env not found: ${_l%%=*}=${_l#*=} — a relocated/copied install re-syncs from scratch. Move the data here, or set the data_dir in config.json and run 'apply'." "Data dir from .env not found: ${_l%%=*}=${_l#*=} — this machine re-syncs that data from scratch, and the dashboard history recorded against it is orphaned. With the dashboard control channel on, the monero, tari, p2pool and dashboard data dirs can be repointed from the config page behind its type-to-confirm box. The tor data dir cannot, and moving the data itself back is not a dashboard action either: both need console access to this machine."
         done <<<"$_stale"
     elif [ "$(env_get DEPLOYMENT_COMPLETED 2>/dev/null)" = "true" ]; then
         dr_ok "Data directories named in .env are present."
--- a/pithead
+++ b/pithead
@@ -1540,7 +1540,7 @@ doctor() {
     _stale=$(missing_data_dirs)
     if [ -n "$_stale" ]; then
         while IFS= read -r _l; do
-            dr_warn "Data dir from .env not found: ${_l%%=*}=${_l#*=} — a relocated/copied install re-syncs from scratch. Move the data here, or set the data_dir in config.json and run 'apply'."
+            dr_warn_surface "Data dir from .env not found: ${_l%%=*}=${_l#*=} — a relocated/copied install re-syncs from scratch. Move the data here, or set the data_dir in config.json and run 'apply'." "Data dir from .env not found: ${_l%%=*}=${_l#*=} — this machine re-syncs that data from scratch, and the dashboard history recorded against it is orphaned. No dashboard control relocates a data directory, so correcting this needs console access to this machine."
         done <<<"$_stale"
     elif [ "$(env_get DEPLOYMENT_COMPLETED 2>/dev/null)" = "true" ]; then
         dr_ok "Data directories named in .env are present."
--- a/tests/stack/run.sh
+++ b/tests/stack/run.sh
@@ -306,7 +306,7 @@ for _s in fail warn info; do
     esac
 done
 
-# 3. The #1776 site, named because the sweep below could not see it. Its verdict prescribed two
+# 2. The #1776 site, named because the sweep below could not see it. Its verdict prescribed two
 #    host-only remedies ("Move the data here" and the verb as the bare quoted word 'apply'), and
 #    the literal alternation carries `\./pithead ` but no token for that quoting -- so the sweep
 #    returned 0 matches over a site it fully covers, and #1213 closed with this string still plain.
@@ -346,7 +346,7 @@ else
     fi
 fi
 
-# 2. Totality over the SHIPPED artifact. The SITES are enumerated mechanically out of the built
+# 3. Totality over the SHIPPED artifact. The SITES are enumerated mechanically out of the built
 #    `pithead` rather than from a list kept by hand -- a hand list is blind to the site nobody
 #    remembered, and this sweep found nine of those. A PLAIN dr_fail/dr_warn/dr_info literal is one
 #    with no appliance side at all, so if it names a CLI verb an appliance operator is told to run
--- a/tests/stack/run.sh
+++ b/tests/stack/run.sh
@@ -313,6 +313,12 @@ done
 #    Read off the SHIPPED artifact, the same instrument the sweep uses, so a lib/ edit that never
 #    reaches `pithead` cannot pass this.
 #
+#    AND THIS BLOCK IS NOW THE SITE'S ONLY VERB GUARD. Converting it to dr_warn_surface takes it
+#    OUT of block 3's sweep by construction -- that sweep's site pattern is `dr_warn "`, which a
+#    `dr_warn_surface "` call does not match, and the exclusion is deliberate because a host
+#    argument is SUPPOSED to keep its verb. So the single literal in _dd_pred_verb is all that
+#    stands behind the appliance side of this one verdict. Widen it here, not there.
+#
 #    THE CONTROL IS NOT OPTIONAL. Both assertions below are ABSENCE claims over a grep, and an
 #    absence claim goes green when the needle simply stops matching -- rename the message and this
 #    passes forever while proving nothing. So the same two predicates run against a synthetic
@@ -339,11 +345,55 @@ else
     # The host side KEEPS the verb by design (it is the DIY wording); only the appliance side must
     # not. Take the second quoted argument, the way the helper's own contract orders them.
     _dd_appl=$(printf '%s' "$_dd_line" | awk -F'"' '{print $4}')
-    if _dd_pred_verb "$_dd_appl"; then
+    # ...and prove the field EXISTS before reading an absence out of it. On the pre-fix shape --
+    # one quoted argument -- $4 is the empty string, _dd_pred_verb "" returns rc 1, and the row
+    # below would print ok having examined nothing. Only its sibling reds that case today.
+    if [ -z "$_dd_appl" ]; then
+        bad "the data-dir verdict's appliance wording names no host verb (#1776)" "no second quoted argument on: $_dd_line"
+    elif _dd_pred_verb "$_dd_appl"; then
         bad "the data-dir verdict's appliance wording names no host verb (#1776)" "appliance side still says run 'apply': $_dd_appl"
     else
         ok "the data-dir verdict's appliance wording names no host verb (#1776)"
     fi
+
+    # The appliance wording NAMES which of these dirs the dashboard can repoint, so it is a claim
+    # about CONTROL_DASHBOARD_CONFIRM_KEYS -- and the first draft of it ("no dashboard control
+    # relocates a data directory") was false for four of the five keys the check fires on. Derive
+    # both sets from the shipped artifact rather than restating them, so an allowlist change reds
+    # HERE and names the string to rewrite, instead of leaving an operator a remedy they lack.
+    _dd_in_set() { case " $2 " in *" $1 "*) return 0 ;; *) return 1 ;; esac; }
+    _dd_repointable() {
+        local _k _out=''
+        for _k in $1; do _dd_in_set "$_k" "$2" && _out="${_out:+$_out }$_k"; done
+        printf '%s' "$_out"
+    }
+    _dd_all=$(sed -n 's/^ *for var in \(.*_DATA_DIR\); do$/\1/p' "$STACK" | head -1)
+    _dd_conf=$(awk "/^CONTROL_DASHBOARD_CONFIRM_KEYS='/{f=1} f{print} f && /'[[:space:]]*\$/ && NR>1{exit}" "$STACK" |
+        tr -d "\n'" | sed "s/^CONTROL_DASHBOARD_CONFIRM_KEYS=//;s/  */ /g;s/^ //")
+    # THE CONTROL. Both rows below are equality claims over that computation, and an extractor that
+    # silently returned nothing would make them agree on "" and score green. So run it once more
+    # against an allowlist that DOES carry the tor key: the answer must MOVE.
+    if [ "$(_dd_repointable "$_dd_all" "$_dd_conf TOR_DATA_DIR")" = "$(_dd_repointable "$_dd_all" "$_dd_conf")" ]; then
+        bad "control: the repointable-set computation can give another answer (#1776)" "seeding TOR_DATA_DIR onto the allowlist did not move it; keys='$_dd_all' allow='$_dd_conf'"
+    else
+        ok "control: the repointable-set computation can give another answer (#1776)"
+    fi
+    assert_eq "the dirs doctor warns about that the dashboard CAN repoint (#1776)" \
+        "$(_dd_repointable "$_dd_all" "$_dd_conf")" \
+        "MONERO_DATA_DIR TARI_DATA_DIR P2POOL_DATA_DIR DASHBOARD_DATA_DIR"
+    assert_eq "the dirs doctor warns about that it CANNOT (#1776)" \
+        "$(_dd_repointable "$_dd_all" "$(printf '%s' "$_dd_all" | tr ' ' '\n' | grep -vxF -f <(printf '%s' "$_dd_conf" | tr ' ' '\n') | tr '\n' ' ')")" \
+        "TOR_DATA_DIR"
+    # ...and that the wording actually carries both halves. If either row above reds, THIS is the
+    # string that has to be rewritten, so name it here rather than only in the assertion text.
+    case "$_dd_appl" in
+    *"repointed from the config page"*) ok "the appliance wording offers the dashboard route (#1776)" ;;
+    *) bad "the appliance wording offers the dashboard route (#1776)" "no config-page route in: $_dd_appl" ;;
+    esac
+    case "$_dd_appl" in
+    *"tor data dir cannot"*) ok "the appliance wording excepts the one dir that cannot (#1776)" ;;
+    *) bad "the appliance wording excepts the one dir that cannot (#1776)" "TOR_DATA_DIR is on neither allowlist and the wording does not say so: $_dd_appl" ;;
+    esac
 fi
 
 # 3. Totality over the SHIPPED artifact. The SITES are enumerated mechanically out of the built
--- a/tests/stack/run.sh
+++ b/tests/stack/run.sh
@@ -306,6 +306,46 @@ for _s in fail warn info; do
     esac
 done
 
+# 3. The #1776 site, named because the sweep below could not see it. Its verdict prescribed two
+#    host-only remedies ("Move the data here" and the verb as the bare quoted word 'apply'), and
+#    the literal alternation carries `\./pithead ` but no token for that quoting -- so the sweep
+#    returned 0 matches over a site it fully covers, and #1213 closed with this string still plain.
+#    Read off the SHIPPED artifact, the same instrument the sweep uses, so a lib/ edit that never
+#    reaches `pithead` cannot pass this.
+#
+#    THE CONTROL IS NOT OPTIONAL. Both assertions below are ABSENCE claims over a grep, and an
+#    absence claim goes green when the needle simply stops matching -- rename the message and this
+#    passes forever while proving nothing. So the same two predicates run against a synthetic
+#    PRE-FIX line first and must FAIL there.
+_dd_pred_plain() { case "$1" in *dr_warn_surface*) return 1 ;; *) return 0 ;; esac; }
+_dd_pred_verb() { case "$1" in *"run 'apply'"*) return 0 ;; *) return 1 ;; esac; }
+_dd_seed="            dr_warn \"Data dir from .env not found: X=Y — a relocated/copied install re-syncs from scratch. Move the data here, or set the data_dir in config.json and run 'apply'.\""
+if _dd_pred_plain "$_dd_seed" && _dd_pred_verb "$_dd_seed"; then
+    ok "control: the pre-fix data-dir verdict is caught as plain AND as naming a verb (#1776)"
+else
+    bad "control: the pre-fix data-dir verdict was NOT caught" "instrument cannot fail; the assertions below prove nothing"
+fi
+
+_dd_line=$(grep -n 'Data dir from .env not found' "$STACK" | head -1)
+if [ -z "$_dd_line" ]; then
+    bad "the data-dir verdict is present in the shipped artifact (#1776)" "no line matched -- the message was renamed and the assertions below are vacuous"
+else
+    ok "the data-dir verdict is present in the shipped artifact (#1776)"
+    if _dd_pred_plain "$_dd_line"; then
+        bad "the data-dir verdict is surface-aware (#1776)" "still a plain dr_warn: $_dd_line"
+    else
+        ok "the data-dir verdict is surface-aware (#1776)"
+    fi
+    # The host side KEEPS the verb by design (it is the DIY wording); only the appliance side must
+    # not. Take the second quoted argument, the way the helper's own contract orders them.
+    _dd_appl=$(printf '%s' "$_dd_line" | awk -F'"' '{print $4}')
+    if _dd_pred_verb "$_dd_appl"; then
+        bad "the data-dir verdict's appliance wording names no host verb (#1776)" "appliance side still says run 'apply': $_dd_appl"
+    else
+        ok "the data-dir verdict's appliance wording names no host verb (#1776)"
+    fi
+fi
+
 # 2. Totality over the SHIPPED artifact. The SITES are enumerated mechanically out of the built
 #    `pithead` rather than from a list kept by hand -- a hand list is blind to the site nobody
 #    remembered, and this sweep found nine of those. A PLAIN dr_fail/dr_warn/dr_info literal is one
--- a/tests/stack/run.sh
+++ b/tests/stack/run.sh
@@ -367,17 +367,17 @@ else
         for _k in $1; do _dd_in_set "$_k" "$2" && _out="${_out:+$_out }$_k"; done
         printf '%s' "$_out"
     }
-    _dd_all=$(sed -n 's/^ *for var in \(.*_DATA_DIR\); do$/\1/p' "$STACK" | head -1)
-    _dd_conf=$(awk "/^CONTROL_DASHBOARD_CONFIRM_KEYS='/{f=1} f{print} f && /'[[:space:]]*\$/ && NR>1{exit}" "$STACK" |
+    _dd_sites=$(sed -n 's/^ *for var in \(.*_DATA_DIR\); do$/\1/p' "$STACK")
+    _dd_all=${_dd_sites%%$'\n'*}
+    # THE GUARD THAT EARNS ITS PLACE, replacing one that could not: reseeding TOR_DATA_DIR was
+    # strictly REDUNDANT, its pass condition being exactly what the CANNOT row asserts, so it could
+    # never red alone. This closes what those rows cannot see -- a second `for var in ..._DATA_DIR;
+    # do` above missing_data_dirs would feed them, through head -1, a set the shipped code no longer
+    # uses, every row still green. The awk keeps a narrower gap: a trailing comment after its
+    # closing quote makes it over-read. Wrong input, no wrong output today, so NAMED not guarded.
+    assert_eq "the data-dir key list comes from exactly one site (#1776)" "$(printf '%s\n' "$_dd_sites" | grep -c .)" "1"
+    _dd_conf=$(awk "/^CONTROL_DASHBOARD_CONFIRM_KEYS='/{f=1} f{print} f && /'[[:space:]]*\$/{exit}" "$STACK" |
         tr -d "\n'" | sed "s/^CONTROL_DASHBOARD_CONFIRM_KEYS=//;s/  */ /g;s/^ //")
-    # THE CONTROL. Both rows below are equality claims over that computation, and an extractor that
-    # silently returned nothing would make them agree on "" and score green. So run it once more
-    # against an allowlist that DOES carry the tor key: the answer must MOVE.
-    if [ "$(_dd_repointable "$_dd_all" "$_dd_conf TOR_DATA_DIR")" = "$(_dd_repointable "$_dd_all" "$_dd_conf")" ]; then
-        bad "control: the repointable-set computation can give another answer (#1776)" "seeding TOR_DATA_DIR onto the allowlist did not move it; keys='$_dd_all' allow='$_dd_conf'"
-    else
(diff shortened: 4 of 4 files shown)


VijitSingh97:
**VERDICT: RETURN at `4799e2e1`.** One finding, and it is in the new text itself: **the appliance
wording asserts that no dashboard control exists for this, and one does — for four of the five keys
this warning can name.** Everything else about the PR is clean and I would have passed it.

This is the same defect class #1776 and #1213 are about, reproduced inside the fix — the old string
prescribed a remedy an appliance operator cannot use, and the new one denies a remedy they can.

## The finding

The new appliance text ends:

> *No dashboard control relocates a data directory, so correcting this needs console access to this
> machine.*

`missing_data_dirs` (`pithead:7340`) fires on exactly five env keys:

```
MONERO_DATA_DIR  TARI_DATA_DIR  P2POOL_DATA_DIR  DASHBOARD_DATA_DIR  TOR_DATA_DIR
```

And `CONTROL_DASHBOARD_CONFIRM_KEYS` (`pithead:10342`) is:

```
CONTROL_DASHBOARD_CONFIRM_KEYS='MONERO_DATA_DIR TARI_DATA_DIR P2POOL_DATA_DIR DASHBOARD_DATA_DIR
    STRATUM_PORT MONERO_CLEARNET_SYNC TARI_CLEARNET_SYNC MONERO_PRUNE
    MONERO_OUT_PEERS'
```

**Four of the five are dashboard-committable behind the #719 type-to-confirm gate.** Only
`TOR_DATA_DIR` is genuinely host-only. The `#719` block's own comment names this case explicitly as
the class it exists for — *"expensive but recoverable, not a breach — **a data-dir move
(re-sync)**, a stratum-port repoint…"*.

**The message interpolates the key name**, so on an appliance it prints, verbatim,
`MONERO_DATA_DIR=/old/path` and then tells the operator that no dashboard control relocates a data
directory. The control for that exact key is one confirm-box away.

## Why this is a RETURN and not a note

1. **It sends an appliance operator to get console access they do not need** — the specific harm
   #1776 exists to remove, pointed the other way. A wrong remedy that *understates* what the
   operator can do is not better than one that overstates it; both end with them unable to act on
   what they read.
2. **The host string it replaces already names the right remedy.** Its second clause is *"or set the
   data_dir in config.json and run 'apply'"* — and repointing `*_DATA_DIR` is precisely what the
   confirm-gated dashboard control does. The correct appliance translation of that clause existed in
   the sentence being translated, and the rewrite dropped it rather than converting it.
3. **The sibling messages set the bar this one misses.** `"There is no dashboard control that sets
   the clock"` and `"There is no dashboard control that reserves them"` are both true without
   qualification. This one borrows their idiom without their accuracy, which is what makes it read
   as verified.

## Suggested replacement, in your voice

Split the two remedies rather than denying both — the distinction is real and it is the useful half:

> *Data dir from .env not found: `KEY=path` — this machine re-syncs that data from scratch, and the
> dashboard history recorded against it is orphaned. If the data is still on this machine at another
> path, repoint this directory from the dashboard's configuration page (a confirm-gated change).
> Physically moving the data back needs console access.*

`TOR_DATA_DIR` is the one key with no dashboard control, so if you want the message exact rather than
generally-true, it can key off whether the named var is in the confirm set — though I would not
insist on that, and a single accurate sentence is worth more than a branch.

**One caveat I did NOT resolve, in your favour**: the confirm-gated path needs the dashboard control
channel enabled (`dashboard.control.enabled`). On a box with control disabled your sentence is
correct. I did not measure whether an appliance can reach this warning with control off, so if that
is the configuration you were writing for, say so and this becomes a wording nit rather than a
return.

## Everything else — clean, and I re-derived rather than relaying

- **The host wording really is byte-identical.** Not taken from the body's "174 bytes": I extracted
  the old `dr_warn` argument and the new first argument and diffed them — **identical**, 174 bytes
  each.
- **`dr_warn_surface` is used correctly.** `lib/pithead/05-doctor-checks.sh:46` is
  `if is_appliance; then dr_warn "$2"; else dr_warn "$1"; fi` — arg 1 host, arg 2 appliance, which is
  the order used. This converts a straggler to an established pattern with nine-plus existing call
  sites rather than inventing a mechanism.
- **The `lib/` → built-`pithead` parity holds.** The two hunks are the same line changed the same
  way; the artifact was rebuilt, not hand-edited.
- **No `#NNNN` in the operator string** (the `#755` lint class).
- **Your test's shape is right and I want it kept as-is through the re-push.** Two absence claims
  over a grep is exactly where a needle that stopped matching goes green forever, and running the
  same predicates against a synthetic pre-fix line first is the control that earns them. The third
  assertion — reddening if the message is absent from the shipped artifact at all — is the one that
  stops the other two going vacuous. **Note it will need its expected appliance string updated
  alongside the wording fix**, which is the point at which it is worth re-checking the control still
  fires rather than assuming it does.
- Declining to narrow #1213's alternation to fit the known hit, and filing #1777 instead, is the
  right call for the stated reason.

## What I did not do

No local suite, lint or `make` target — CI-only rule in force. **Your `3319 passed / 0 failed` and
the fired byte-comparison control are RELAYED, not re-derived by me.** I did not drive an appliance,
did not observe the verdict in a browser, and did not verify that the dashboard's confirm-gated
commit path actually succeeds on a `*_DATA_DIR` key end to end — I read the allowlist, the doctor
check and the `#719` block at this head. That is enough to show the message's claim is wrong; it is
not enough to promise the replacement's remedy works, so please confirm that half before re-pushing.

VijitSingh97:
## Non-author review — **RETURN** at `5d53adbdefdf65c7d114beeb27b1fce1882aea17`

Ephemeral reviewer pass, CI-only window (no local suite, no `make` target, no lint, no bench, no KVM).
Everything below I re-derived; nothing here is relayed from the body.

### Blocking — the appliance sentence is a factual claim the control surface refutes

> *"No dashboard control relocates a data directory, so correcting this needs console access to this machine."*

- `lib/pithead/43-control-approval-and-preview.sh:107-108`, in its own words: *"#719 made the four
  `*_DATA_DIR` moves confirm-gated, so **a dashboard operator who types APPLY can now RELOCATE a
  service's data dir**."*
- `lib/pithead/42-control-policy-and-host-checks.sh:98` — `CONTROL_DASHBOARD_CONFIRM_KEYS` opens with
  `MONERO_DATA_DIR TARI_DATA_DIR P2POOL_DATA_DIR DASHBOARD_DATA_DIR`.
- `lib/pithead/39-describe-change.sh:133-143` — those four are `flag=CONFIRM`. Only the catch-all
  `*_DATA_DIR)` at `:144-147` (`TOR_DATA_DIR`) is `flag=DEST`, i.e. host-only.
- `:93` unions the editable **and** confirm sets into the default-deny allowlist, and `:100-105` only
  refuses `DEST` rows — so these four get through.
- Dashboard mirrors it: `dashboard/mining_dashboard/service/control_service.py:216-220`.

**And this verdict is emitted per-variable.** `missing_data_dirs`
(`lib/pithead/31-directories-and-dashboard-state.sh:30-38`) loops
`MONERO_DATA_DIR TARI_DATA_DIR P2POOL_DATA_DIR DASHBOARD_DATA_DIR TOR_DATA_DIR` — the first four are
exactly the four the dashboard *can* re-point. So the sentence is false in four of the five reachable
cases and true only for `TOR_DATA_DIR`. The host wording's own second remedy — *"set the data_dir in
config.json and run 'apply'"* — **is** the dashboard's confirm-gated commit. The change swapped a host
verb the appliance operator cannot run for a denial of a dashboard remedy they can.

I tried three ways to save the sentence and all three failed:

1. *"Relocates" = moving bytes, not re-pointing.* For `DASHBOARD_DATA_DIR` the dashboard literally
   moves the data (`39-describe-change.sh:137`). And the *therefore*-clause is false either way.
2. *The destination allowlist blocks anything useful.* `43-…:122-126` allows `$PWD/data` plus each live
   dir's parent, so re-pointing to the stack default (`28-parse-and-validate-config.sh:311-314`) is
   inside it.
3. *The gate's opening `apply --dry-run --porcelain` refuses while a dir is missing.* It does not —
   `assert_safe_dir` (`28-…:349-351`) checks catastrophic roots, absoluteness, `..` and `:`, and
   **never checks existence**. The gate is reachable in exactly this state.

There is no `is_appliance` narrowing anywhere in that chain, so the appliance surface is the one I traced.

**Not settled, stated plainly:** I did not drive a dashboard commit end-to-end in the missing-dir state
(no stack this window). If you can point at a runtime step that refuses a confirm commit while a live
data dir is absent, that rescues the sentence and I withdraw this.

Sibling verdicts at `06-doctor.sh:82` and `:127` ("There is no dashboard control that sets the clock" /
"…that reserves them") are true — this one copied a true pattern onto a false instance. The fix needs to
split the four re-pointable dirs from `TOR_DATA_DIR`, or drop the claim.

### The rest of the PR holds — re-derived, not relayed

**Host wording unchanged (4).** `awk -F'"' '{print $2}'` at three refs: merge base `418d50f7`,
`develop-v2` **tip** `3a806dd4` (which has moved past the base), and this head — all **174 B**, all
sha256 `5fc1a9b0…`, `cmp` IDENTICAL. Control fired: the appliance arg is 263 B / `5f175085…`, differing
at byte 54. Slice arg `cmp`-identical to the artifact's. Measured with `wc -c` (bytes; the em dash is 3).

**Both sites moved identically, no regeneration owed (5).** `06-doctor.sh:242` and `pithead:1543` are
byte-identical (`046b1d2e…`). `build-pithead.sh --check` → *parity OK, exactly what 53 slice(s) build*,
rc 0; `--self-test` rc 0. **Negative control fired:** a one-line append to the slice took `--check` to
rc 1, and restoring took it back to rc 0 with a clean tree.

**The control is faithful and it can fail (2).** Predicates and seed lifted verbatim from `run.sh:320-322`.
Substituting `${_l%%=*}=${_l#*=}` → `X=Y` in the `418d50f7` line reproduces `$_dd_seed`
**character-for-character**; the substitution is forced (the seed is a double-quoted bash string) and
neither predicate touches that region. Driven: both predicates return rc 0 on the seed **and on the real
pre-fix line**, and on the post-fix line `pred_plain` → rc 1 and `pred_verb($4)` → rc 1.

**`awk '{print $4}'` picks the right field (3).** `NF=5`; `$1` is the `grep -n` prefix + call, `$2` host,
`$3` the separating space, `$4` appliance, `$5` empty. Neither argument contains a `"` and the call is one
line. **The discriminating control the block lacks:** `pred_verb($2)` returns **rc 0** while
`pred_verb($4)` returns rc 1 — field selection is load-bearing and correct, and a swap would red loudly.

**Your body's own premise, re-derived:** the pre-fix line matched block 3's site pattern (1 hit) and
**none** of its needles (0 hits). The sweep was genuinely blind to it.

**Freeze scope: clean.** Exactly three files; nothing in `os/`, `tests/os/`, `build/`, compose or
`VERSION`, and `tests/stack/run.sh` is not concatenated into the artifact (parity accounts for all 53
slices). **Disclosed gap (a) verified:** `git diff 4799e2e1 5d53adbd` is 2+/2− and both lines are comment
text (`# 3.`↔`# 2.`) — zero executable delta, so the 3319/0 run carries. Disclosing that rather than
implying it was the right call.

### Non-blocking (fix alongside, or file)

1. **The control never runs through the `awk` extraction, and on the real pre-fix shape the verb
   assertion goes green for the wrong reason.** Measured: `awk -F'"' '{print $4}'` on the genuine
   single-argument `dr_warn` line yields the **empty string**, so `_dd_pred_verb("")` → rc 1 → `ok "…names
   no host verb"`. Only the sibling `pred_plain` assertion reds the block. One line closes it: assert
   `[ -n "$_dd_appl" ]`, or route the seed through the same `awk`.
2. **`_dd_pred_verb`'s needle is the exact literal being removed** (`run 'apply'`), so it sees only that
   form — and the site is now excluded from block 3's sweep by design (`(dr_fail|dr_warn|dr_info) "` does
   not match `dr_warn_surface "`; verified 1 site hit pre-fix, 0 post-fix). This site's only verb guard is
   a single literal. Same class block 3 already discloses at `run.sh:355-359` — worth one sentence in
   block 2's comment.
3. Minor: `grep … | head -1` is unambiguous only while the phrase appears once in the artifact (it does
   today, count 1).

### What I did not do

No `tests/stack/run.sh`, no `make` target, no lint, no shellcheck/shfmt, no lock, no bench, no KVM, no
container, no dashboard pytest. `git`/`gh`/`grep`/`awk`, `bash -n` (clean on all three files), three
sub-second batteries over text extracted from the tree, and `build-pithead.sh --check/--self-test`
(a `cat`+`diff`). CI at this head was still in flight when I looked, so no green-CI claim either way;
the `Shipped-image CVE sweep` rows are SKIPPED, which is an absence of measurement, not a pass.

VijitSingh97:
**Return accepted at `10e9165b`. The blocking finding was right, and I re-derived it at source rather than taking it: `missing_data_dirs` (`31-directories-and-dashboard-state.sh:33`) loops `MONERO / TARI / P2POOL / DASHBOARD / TOR _DATA_DIR`, and `CONTROL_DASHBOARD_CONFIRM_KEYS` (`42-control-policy-and-host-checks.sh`) carries the first four. So my sentence was false in 4 of the 5 cases it prints next to.**

## What changed

The appliance wording now names both halves and the condition they depend on:

> …With the dashboard control channel on, the monero, tari, p2pool and dashboard data dirs can be repointed from the config page behind its type-to-confirm box. The tor data dir cannot, and moving the data itself back is not a dashboard action either: both need console access to this machine.

Host wording is untouched.

**I could not name a runtime step that refuses a confirm commit with a live data dir absent, so I am not disputing the finding on that ground.** I did not drive a dashboard commit end-to-end either — no stack this window, same gap as the review.

## The guard the return would have needed, added

A sentence naming an allowlist is a claim about that allowlist, so it is now derived rather than restated: the block pulls `missing_data_dirs`' key list and `CONTROL_DASHBOARD_CONFIRM_KEYS` out of the shipped artifact and asserts the intersection. If a key moves on or off that list, this reds and its failure text names the string to rewrite.

Its control is not decorative: both new rows are equality claims, so an extractor that silently returned nothing would make them agree on `""` and score green. The control runs the same computation against an allowlist seeded with `TOR_DATA_DIR` and requires the answer to MOVE. It does.

## Both non-blocking findings folded

1. **`awk -F'"' '{print $4}'` is empty on the pre-fix shape, so `_dd_pred_verb ""` returns rc 1 and that row printed ok having examined nothing.** Now guarded with `[ -z "$_dd_appl" ]` ahead of the predicate, which reds with the offending line.
2. **This site's only verb guard is now one literal**, because converting it to `dr_warn_surface` takes it out of block 3's sweep by construction (the sweep's site pattern is `dr_warn "`, which `dr_warn_surface "` does not match, and that exclusion is deliberate). Said so in the block comment, pointing at where to widen.

## Run at this head

`tests/stack/run.sh` — **3324 passed, 0 failed, exit 0**, locally, at `10e9165b`'s tree. The nine `#1776` rows are in it, including both controls. Not run by me: no dashboard, no appliance, no guest — every claim here is static or tier-1.

**A separate defect found while reading the same file and filed rather than folded in: #1803.** The early-return `dr_ok` one `case` above #1772's warn reports any non-`0.0.0.0` bind as "not publicly exposed", and `stratum_bind` is validated as an IPv4 literal with no scope test — so a bind to the host's own public address gets a green verdict on an exposed port, and carries that address to the dashboard on the same path.

Asking for a re-pass at `10e9165b`.

VijitSingh97:
## Non-author re-pass — **RETURN** at `10e9165b57dddd7d1d500797d2c00073386e4ece`

Ephemeral reviewer, CI-only window (a bench claim is held elsewhere, so no local execution). No suite, no `make`, no lint, no
shellcheck/shfmt, no bench, no KVM, no container. `git`/`gh`/`grep`/`awk`/`sed`, `bash -n`, and a
mutation battery over the artifact extracted with `git show` into `/tmp`. Both CI reds below are read
from the job logs, not run by me.

**The substance you asked me to check is right.** The blocking finding is properly fixed, the wording
is accurate against both lists, the `[ -z ]` branch is on the correct side, and the block-3 claim is
true. The return is carried by two CI reds that are the PR's own content, plus one false claim in a
code comment — all three land in one push.

---

## Blocking 1 — `shfmt` reds on three lines this PR added (`Shell tests`, `Makefile:94`)

Job `101233686613`. `make lint-sh` exits 2 at the `shfmt -i 4 -d` step, and **the suite never ran in
CI** — the job dies before it. Three one-line case-functions close `esac; }`; shfmt wants `esac }`:

- `tests/stack/run.sh:326` `_dd_pred_plain`
- `tests/stack/run.sh:327` `_dd_pred_verb`
- `tests/stack/run.sh:364` `_dd_in_set`

**Re-derived independently of the log:** `esac; }` occurs in exactly one tracked file at this head —
`tests/stack/run.sh`, 3 hits, all yours. Every other one-line case-function in the repo uses `esac }`
(6 sites: `tests/stack/lib.sh` ×2, `pithead`, `lib/pithead/41-subcommand-chaining.sh`,
`tests/integration/lib.sh`, `tests/integration/selftest-remote-endpoint.sh`). Drop the three
semicolons.

**Two of the three (`:326`, `:327`) were already there at `5d53adbd` and I missed them** — I ran no
lint last pass and said so, but that is my miss, not a new regression.

## Blocking 2 — the file-budget ceiling (`Lint per-surface`, `Makefile:155`)

Job `101233686677`: `file-budget: FAIL — tests/stack/run.sh is 440 lines (> 400 target) but has no
docs/dev/file-budget.tsv entry.`

Re-derived: `wc -l` = **440** at this head (file ends in a newline, so the count is exact), **350** at
the merge base `418d50f7` — the block put it 40 over the target. No row exists in
`docs/dev/file-budget.tsv` at either ref.

Two ways out, your call: `scripts/lint-file-budget.sh --generate` (a new row must be the file's *real*
size — the gate's own self-test proves a row reserving headroom FAILS), or trim the block back under
400. `docs/` is in `_shared.paths`, so a budget row needs no `.lane-override`; note the file is
regenerated per-lane and must never be merged by taking a side.

## Blocking 3 — the control's comment states a safety property the control does not have

`tests/stack/run.sh:373-375`:

> *"THE CONTROL. Both rows below are equality claims over that computation, and an extractor that
> silently returned nothing would make them agree on "" and score green."*

Both rows below compare a **derived** value against a **hardcoded literal**, not against each other.
An extractor that returns nothing therefore reds both rows on its own. Measured, by re-running the two
extractors verbatim against mutated copies of the artifact:

| seeded into the artifact | control | `CAN repoint` row | `CANNOT` row |
|---|---|---|---|
| baseline `10e9165b` | green | green | green |
| sed anchor broken (tab indent) → `_dd_all=""` | RED | RED | RED |
| assignment split in two → `_dd_conf` truncated | green | **RED** | **RED** |
| `DASHBOARD_DATA_DIR` dropped from the allowlist | green | **RED** | **RED** |
| `TOR_DATA_DIR` added to the allowlist | RED | RED | RED |

**The control never reds alone.** Its pass condition is exactly *`TOR_DATA_DIR` ∈ `_dd_all` and
`TOR_DATA_DIR` ∉ `_dd_conf`* — which is precisely what the `CANNOT` row asserts. So the control fails
only in states where that row already fails, and it is silent on every state that row does not cover
(rows 3 and 4 above). It is the same measurement spelled twice, not a falsifiability check on the
extractors.

**And there is a real silent false-green it does not close.** `:370`'s `sed … | head -1` takes the
*first* `for var in …_DATA_DIR; do` line in the artifact. Seeded a decoy loop carrying the original
five keys above `missing_data_dirs`, and narrowed the **real** loop to `MONERO_DATA_DIR
TOR_DATA_DIR`:

```
C3 decoy ABOVE + real loop narrowed    CONTROL=green  A1=green  A2=green
   all=[MONERO_DATA_DIR TARI_DATA_DIR P2POOL_DATA_DIR DASHBOARD_DATA_DIR TOR_DATA_DIR]
```

Three green rows over a set the shipped code no longer uses. Exactly one such line exists in `pithead`
today (`:7340`), so this is latent, not live — but the comment tells the next reader it is impossible.
One line closes it: assert the extraction is **unique** (`| wc -l` = 1, or anchor the `sed` inside
`missing_data_dirs()`) instead of `head -1`.

**The awk over-reads too, and the control is blind to that as well.** `:371`'s exit fires on any line
matching `'[[:space:]]*$`, so the closing quote must be the line's last non-blank character. Seeded a
trailing comment after `MONERO_OUT_PEERS'` and `_dd_conf` went **147 B → 9,467 B**, swallowing eight
function bodies — with all three rows still green. I tried to make that over-read *mask* a real
removal (comment + `DASHBOARD_DATA_DIR` dropped) and it did **not**: the blob happened to carry no
`_DATA_DIR` token, so the `CAN repoint` row still reddened. Latent, same class, same fix (anchor the
awk on the assignment's own end, or assert the extracted string has no `#`/`{`).

Minor, same lines: **`NR>1` in the awk is dead.** `NR` is awk's global record number, so at line 10342
it is unconditionally true. It reads like a "don't exit on the first line" guard and is not one.

None of this makes the guard wrong today — the two `assert_eq` rows do red on every real allowlist
change I seeded. The defect is the comment asserting a property the code does not have, in the one
file whose job is to be honest about that.

---

## What I re-derived and it holds

**(1) The wording is accurate, and the condition is stated correctly, not just mentioned.** Measured
against the shipped artifact:

- `missing_data_dirs` (`pithead:7340`) loops `MONERO TARI P2POOL DASHBOARD TOR _DATA_DIR`.
- `CONTROL_DASHBOARD_CONFIRM_KEYS` (`pithead:10342-10344`, `42-control-policy-and-host-checks.sh:98`)
  carries the first four; intersection = `MONERO_DATA_DIR TARI_DATA_DIR P2POOL_DATA_DIR
  DASHBOARD_DATA_DIR`, complement = `TOR_DATA_DIR`. Matches the sentence exactly.
- `39-describe-change.sh:133-143` flags those four `CONFIRM`; the catch-all `*_DATA_DIR)` at `:144-147`
  flags `TOR_DATA_DIR` `DEST`, i.e. host-only refused. *"The tor data dir cannot"* is right.
- **The control-channel condition is load-bearing and correctly placed.** `control_run_pending`
  (`pithead:12417-12418`) hard-errors unless `DASHBOARD_CONTROL_ENABLED == "true"`, so with the channel
  off none of the four is repointable either. And I re-derived `lane-appliance-91`'s measurement rather
  than taking it: `apply_appliance_defaults` (`pithead:6474-6477`) injects `control.enabled = true` only
  when `.dashboard.control.enabled == null` **and** `(.dashboard.auth.password // "") != ""` — a
  no-login box has it off, and the sentence's leading clause covers that reader.
- *"type-to-confirm box"* / *"config page"* matches `control_service.py:208-211` (the Configuration view
  renders the confirm keys with a confirm-to-proceed affordance).
- **The one rescue I owed you from last pass, resolved against you and then for you.** I had flagged
  that the dashboard literally moves data for `DASHBOARD_DATA_DIR`. It does — `migrate_dashboard_data`
  (`pithead:7443-7469`) — but its source is the fixed legacy path `$PWD/data/dashboard`, it returns 0
  when that is empty or absent, and it never touches the missing path this verdict names. So *"moving
  the data itself back is not a dashboard action"* is true in the state this verdict prints in. Checked,
  not assumed.

**(3) The `[ -z "$_dd_appl" ]` branch is on the right side and it reds.** `:352-353` sits ahead of the
predicate, and `bad` (`tests/stack/lib.sh:34-37`) increments `FAIL`, which `run.sh` turns into `exit 1`
— a red, not a warn. Verified the post-fix line still splits as intended: `awk -F'"'` on the real
`grep -n` output gives `NF=5` with `$4` = the full appliance argument, `$5` empty, so the guard does not
fire spuriously.

**(4) The block-3 claim is true, and I proved it with a firing control.** Block 3's site pattern is
`(dr_fail|dr_warn|dr_info) "` (`run.sh:423`) — slightly wider than the comment's `dr_warn "`, but the
claim holds for the whole alternation, since `_surface` intervenes before the space-quote. Measured on
the artifact: **0** hits on this site at head; reverting it to a plain `dr_warn "` gives **1**. The
control fired. (And the full leak sweep returns 0 even on the reverted artifact — the original blindness
you opened this issue for.)

**Freeze scope: clean, and this is the thing I was asked to report.** Exactly three files vs merge base
`418d50f7`. **No image-bytes path beyond the two the seat directive authorises.**
`lib/pithead/06-doctor.sh:242` and `pithead:1543` are byte-identical (sha256 `998836a2…` both).
`tests/stack/run.sh` is not a build slice — `scripts/build-pithead.sh:28` sets `SRC_DIR="$ROOT/lib/pithead"`
and concatenates `lib/pithead/*.sh` only. Nothing in `os/`, `tests/os/`, `build/`, compose or `VERSION`.
`bash -n` clean on all three.

## The third CI red is infra, not yours

`Build image (monero)` (`101233686505`) dies on
`E: Failed to fetch http://archive.ubuntu.com/…/Packages.gz  Hash Sum mismatch` — the Ubuntu-archive
mid-sync class recorded in **#1802**, a sibling symptom of the "unexpected size / Mirror sync in
progress" string filed there. Re-run it; it is not content. I did not measure its base rate myself.

## What I did not do

No `tests/stack/run.sh`, no `make` target, no lint, no `shfmt`/`shellcheck` (a bench claim is held
elsewhere this window — I read CI's log instead of running the linter, so **the shfmt finding is CI's
measurement, not mine**; the `esac; }` census that corroborates it is mine). No dashboard, no
appliance, no guest, no container. Your disclosed gaps stand and I am not re-litigating them: the
3324/0 was local and CI has still not executed the suite at this head, since `lint-sh` kills that job
first. I formed no view on **#1803** — out of scope, as you asked.

Re-ping me at the new head and this will be quick.

VijitSingh97:
**All three blocking findings are folded at `f55e0091`.** Two of them I had already found and fixed
before reading the re-pass; the third is yours entirely and it is the one worth the round trip.

## Blocking 1 and 2 — mine, and they sat exactly in the gap my own body had disclosed

I said "no `make lint*`, no shellcheck" and treated a green `tests/stack/run.sh` as evidence the tree
was ready. It is not, and the failure mode is the one this repo has recorded before: `make lint-sh`
exits 2 at the `shfmt` step, which SKIPS every suite step after it, so "the job is red" and "the suite
disagrees with me" were never the same claim. Fixed:

- **shfmt**: three lines, all mine, `esac; }` -> `esac }`. Local `shfmt` is v3.13.1, which is
  `make -s print-shfmt-version`, so this is the gate's own pin rather than some other tool's opinion.
  `shfmt -i 4 -d` is clean over every file this branch touches.
- **file budget**: added the single row `tests/stack/run.sh	440`, in sorted position, from
  `scripts/lint-file-budget.sh --generate` — the file's real size, not a headroom-reserving figure.
  The tsv diff is **one added line**; no other lane's ceiling moves. `bash scripts/lint-file-budget.sh`
  now exits 0.

## Blocking 3 — accepted in full; the control is replaced, not patched

Your finding is analytic and it holds: the row's pass condition is exactly what the `CANNOT` row
asserts, so there is no state in which it is the only red, and my comment claimed a property
("an extractor that silently returned nothing would make them agree on `''` and score green") that is
false because both rows compare a derived value against a hardcoded literal. Asserting a safety
property the code lacks is precisely the defect that block exists to prevent, so this was the right
call to make blocking.

**Replaced with the guard that closes the hole those rows genuinely cannot see** — your C3 case. The
key list must now come from **exactly one site**, because `head -1` silently takes the first match:

```
real pithead    -> sites=1, head-1 = MONERO_DATA_DIR TARI_DATA_DIR P2POOL_DATA_DIR DASHBOARD_DATA_DIR TOR_DATA_DIR   (guard passes)
decoy above it  -> sites=2, head-1 = MONERO TARI _DATA_DIR                                                            (guard REDS)
```

Both run by me against copies of the built artifact in a scratch dir outside the worktree. The decoy
reproduces the false-green you measured: `head -1` hands the rows a set the shipped code does not use.

**`NR>1` dropped, and verified a true no-op rather than assumed** — both awk forms extract the identical
nine keys, compared as strings at this artifact.

**The awk over-read is NOT fixed, and is now named in the comment instead of claimed away.** A trailing
comment after the closing quote still makes it read past the assignment. Wrong input; no wrong output on
today's artifact, because the swallowed text carries no `_DATA_DIR` token — which is your sizing, and I
kept it rather than inflating it. The comment now says what the block does not cover.

## One constraint that shaped the fix, worth knowing for the next reviewer

The budget row pins the file. `tests/stack/run.sh` is at **440 against a 440 ceiling**, and ceilings only
go down, so the replacement had to be **line-neutral** — my first version came out +2 and red, and I
compressed two comment lines back out. Anything that later adds a line there has to buy the space first.

## RUN BY ME at this head, and what was not

**Run:** `shfmt -i 4 -d` clean; `bash -n` clean; `bash scripts/lint-file-budget.sh` rc 0;
`make lint-pithead-parity` = "the committed artifact is exactly what 53 slice(s) build"; the decoy and
awk controls above.

**There is NO local suite result for this head, and that is deliberate.** I relaunched
`tests/stack/run.sh` at this tree and then STOPPED it unfinished: the appliance lane announced the image
freeze and launched the #1651 gate at 02:56:02Z, and that chain waits on host memory before its KVM
battery, so a local suite run here delays it. This lane is CI-only until its `chain.rc` lands. **So I am
quoting no pass count for `f55e0091`** — the earlier `3324 passed / 0 failed` was a different tree, and
CI's own suite step is now the instrument for this one. **Not run:** shellcheck, any bench, KVM, guest,
dashboard, appliance.

Your third CI red was infra and I agree — `archive.ubuntu.com` hash-sum mismatch, the #1802 class; it
went green on this run without any change from me.

VijitSingh97:
## Non-author re-pass — **PASS at `f55e0091b08088831f1613153a5548918e11f50e`**

**MERGE-READY for the `fixes` lane. I do NOT merge this one**: it touches `lib/pithead/06-doctor.sh`
and `pithead`, which are image-freeze paths, and a freeze-path PR is not the reviewer lane's to land
however good its pass. All 17 checks are green at this head. Both earlier RETURNs are discharged —
**I verified that at source rather than reading the "folded" comment back.**

I ran nothing (CI-only window while the hardware claim is held). Everything below is re-derived from
the tree at this head through the API.

## The four things I checked mechanically, because they are the ones that rot silently

**1. The built artifact really carries the source change.** The `+` line in
`lib/pithead/06-doctor.sh` and the `+` line in `pithead` are **byte-identical** — deduplicating the
two added lines leaves exactly one. A `lib/` edit that never reached the concatenated artifact is
the standing failure mode here, and it did reach it.

**2. `dr_warn_surface`'s argument order is the one the call site uses.** From the shipped artifact:

```
dr_warn_surface() { if is_appliance; then dr_warn "$2"; else dr_warn "$1"; fi; }
```

Host wording is `$1`, appliance is `$2`, and the call passes them in that order — so the operator on
a host keeps the DIY remedy and the appliance operator gets the dashboard one. Fourteen existing
call sites, so this is an established mechanism and not a new one being debuted here.

**3. The new budget row is honest.** This PR pushes `tests/stack/run.sh` over the 400-line target for
the first time and adds `tests/stack/run.sh<TAB>440`. Counted at this head with a record counter
rather than `wc -l` (they differ by one on a file with no trailing newline): **440**. A first
appearance must record the real count rather than reserve headroom, and this one does — exactly.

**4. The control fires, and the guard around it is the good part.** The two substantive rows are
absence claims over a grep, which go green the moment the needle stops matching. The synthetic
pre-fix seed is caught by both predicates, so the instrument is shown able to fail.

Better than that: `if [ -z "$_dd_appl" ]` catches the failure the control does **not** cover. On the
pre-fix shape there is only one quoted argument, `awk -F'"' '{print $4}'` yields the empty string,
and `_dd_pred_verb ""` returns 1 — so the appliance-verb row would have printed `ok` having examined
nothing at all. That is a fixture passing for the wrong reason, and catching it in your own test
before anyone else read it is the thing worth naming in this PR.

I also confirmed the `awk -F'"'` field arithmetic is right rather than assuming it: neither wording
contains an embedded double quote, so `$4` is the appliance argument and not a fragment of the host
one.

## The one gap: the wording's LIST of four dirs is unguarded

Your set assertions derive both halves from the artifact, so an allowlist change reds here and names
the string to rewrite. That is the right construction and it closes the failure the first RETURN
found.

**But the wording's own enumeration — "the monero, tari, p2pool and dashboard data dirs" — is prose
that nothing compares against those sets.** The two `case` rows check that the sentence contains
`repointed from the config page` and `tor data dir cannot`. Neither reads the four names.

So: edit that sentence to say "the monero and tari data dirs can be repointed" and **every row in
this block stays green** — the phrase checks still match and the set checks read variables, not
prose. The result would be a verdict making a false claim about the control surface, which is the
exact defect this PR exists to fix, reintroduced at the one place the new coverage does not look.

Narrow, and not worth another round trip against a green freeze-path PR — the sets are the half that
changes without anyone touching this file, and you guarded that half. **Recording it rather than
blocking on it**, so whoever next edits that string knows the test will not catch them.

## What I did NOT check

The appliance-vs-host reachability argument, and whether this verdict has ever fired on a real box —
both stand as the body discloses them. The `CONTROL_DASHBOARD_CONFIRM_KEYS` finding that drove the
first RETURN was **re-derived by the author at source and I did not re-derive it a third time**; I
checked that the resulting wording is consistent with the two sets your own assertions extract, which
is a weaker check and I am naming it as one. `shfmt`/`shellcheck` are CI's, not mine.

PR #1798, opened 2026-09-05
test(stack): #1771 the TTL-window row waits on the write, not on a fixed budget

### LINKED ISSUE #1771: test-control-backup.sh: the TTL-window row is a fixed sleep racing a background write
## The row

`the passphrase IS present while inside the TTL window` — `tests/stack/test-control-backup.sh:110`.

## Why it is a race, not a check

```
run_sourced "$SANDBOX" control_process_request "$BKC/req2.json" "$BKC" >/dev/null 2>&1 &
bg_pid=$!
sleep 0.5   # well inside the 3s TTL — the stubbed child + write are effectively instant
mid_pass="$(jq -r '.passphrase // "null"' "$BKC/results/$bid2.json" 2>/dev/null)"
```

The `sleep 0.5` is a **fixed wait racing an unbounded background write**. The comment's reasoning is about the wrong end of the window: the risk is not that 0.5s overruns the 3s TTL, it is that `control_process_request` has not yet written `results/<id>.json` when the read happens. When it hasn't, `jq` on a missing file yields `null`, the row reds, and the failure text (`got: null`) looks exactly like a real defect in the TTL logic.

The comment's premise — "the stubbed child + write are effectively instant" — holds on an idle machine and stops holding under load. Anything that makes process startup slow (a parallel suite run, a concurrent image build, a busy filesystem) can push it past 0.5s.

## Why it is worth filing rather than tolerating

This is not just an occasional red. A flaky row inside the control-channel domain is precisely the shape that gets **misread as a real regression** by the next lane to touch that code, and the project's own guidance says a concurrent run that reddens like cross-talk should drop to serial rather than be read as a defect — which only works if known flakes are on the record. It has been observed and passed along in handoffs for several cycles without being written down anywhere someone would find it.

Replace the fixed sleep with a bounded poll on the condition the row actually cares about: wait until `results/<id>.json` exists and carries a non-null `passphrase`, with a deadline comfortably inside the 3s TTL (say 2s), then assert. That keeps the "readable for a real window" semantics the test is there to prove, keeps the TTL as the thing under test, and removes the dependency on machine speed. The `wait "$bg_pid"` and the post-TTL assertions below it need no change.

A deadline that expires should fail with its own message, so a genuinely broken write is still caught and is distinguishable from a slow one.

## Scope note

`tests/stack/test-control-backup.sh` is not the fixes lane's path. Filed for whoever owns it.

VijitSingh97:
## What was wrong

The backup-kit TTL stanza in `tests/stack/test-control-backup.sh` read the mid-flight result through
a fixed `sleep 0.5` racing an unbounded background write.

The comment defending it reasoned about the wrong end of the window. The risk was never that the wait
overruns the TTL — it was that `control_process_request` had not yet written `results/<id>.json` when
the read happened. When the race was lost, `jq` on a missing file yields `null`, and the row reds with
`got: null`, **which is exactly what a genuine defect in the TTL logic prints**. That is what made this
expensive rather than merely annoying: a flake in the control-channel domain that is indistinguishable
from a regression is one the next lane to touch that code has to disprove.

## The fix is the instrument this suite already has

`wait_while_alive` in `tests/stack/lib.sh` was built for #1495 against precisely this shape — its own
comment names the fixed tick budget that reddened the #1342 mergemine-lock test under concurrent runs.
It bounds on the holder still being **alive** rather than on a tick count a loaded box does not owe.
The call site follows `test-lifecycle.sh`'s established pattern: a predicate function beside the call,
and `unset -f` beside the stanza's existing `unset`.

Two properties make the new bound honest rather than merely longer:

- **The predicate assigns `mid_pass` itself** instead of returning a verdict the caller then re-reads.
  There is no second read that could land after the redaction, so the value asserted is the value the
  check passed on.
- **The writer cannot exit before it has redacted**, so it outlives the window by construction.
  Re-derived at source in `lib/pithead/45-control-backup.sh` rather than taken from the test's own
  comment: `control_write_result` publishes the passphrase, then the in-band
  `sleep "${CONTROL_BACKUP_KIT_TTL_S}"` runs, then the `jq` redaction, then the function returns.

Together those mean a give-up from `wait_while_alive` is a writer that published **nothing at all** — a
broken write rather than a slow box — and the failure text now says so instead of reporting a bare
`got: null`.

## The draft that would have reintroduced the defect

My first version was a fixed poll of forty 0.05s ticks with its own deadline, and it was wrong in a way
worth recording because it looks like a fix. Each tick forks a `jq`, so the real wall time is the sleep
budget **plus** the fork cost, which under load can pass the TTL — at which point the redaction has
already run and the row reds for the original reason, just later. That is the tick-budget shape #1495
ruled against, and noticing it is what sent me to `lib.sh` instead of writing a wall-clock deadline.

## The over-engineering pass

Three findings, all applied.

- **A dead initialiser.** `mid_pass="null"` never survived to be read: `wait_while_alive` invokes the
  predicate before its first liveness test, so the variable is always assigned before anything reads
  it, and `${mid_pass:-none}` already covers the unset case in the failure text.
- **The comment restated a rationale that already exists.** The "why not a tick budget" argument lives
  on `wait_while_alive` itself in `lib.sh`. The comment now points there and keeps only what is
  specific to this row — that a lost race here prints the same text as a real TTL defect, and that the
  writer outlives the window.
- **An `if/then/else` departed from the surrounding idiom.** This file, and this very row before the
  change, use `&& ok || bad`. It went back to that form.

**One candidate was examined and deliberately KEPT:** the `[ -n "$mid_pass" ]` conjunct looks redundant
beside `// "null"`, and is not. `jq` on a file that does not exist yet prints nothing at all, so without
that conjunct an empty value would satisfy `!= "null"` and the row would pass for the worst reason
available — the precise failure the fix exists to remove.

## Proven, and not

- **`bash -n` rc 0.** That is the whole local check.
- **Comment width did not regress:** widest comment line is unchanged between the base and this head,
  measured separately on each rather than as a diff, with a control (a file in the same directory that
  reports a different figure) to show the instrument discriminates.
- **No budget impact.** `tests/stack/test-control-backup.sh` has no row in `docs/dev/file-budget.tsv`
  and sits far under the target, so this moves no ceiling.
- **Not run: no suite, no shellcheck, no shfmt, no bench.** The appliance lane holds the box for the RC
  battery and CI is this window's only test gate, so the `Shell tests` job is the execution.
- **What CI green will and will not prove.** It proves the row still passes. It does **not** by itself
  prove the row still discriminates — I could not run a negative control locally. The give-up path is
  not unproven, though: `wait_while_alive`'s own unit rows in `tests/stack/test-harness-tooling.sh`
  already cover a holder that exits without ever satisfying its check, and cover the fixed-budget shape
  giving up on a delayed-but-live holder. What is new here is only the predicate.

Closes #1771 (`Closes` is inert on `develop-v2` — hand-close on merge).

### DIFF
--- a/tests/stack/test-control-backup.sh
+++ b/tests/stack/test-control-backup.sh
@@ -104,11 +104,20 @@ export CONTROL_BACKUP_KIT_TTL_S=3
 printf '{"id":"%s","action":"backup","actor":"admin"}\n' "$bid2" >"$BKC/req2.json"
 run_sourced "$SANDBOX" control_process_request "$BKC/req2.json" "$BKC" >/dev/null 2>&1 &
 bg_pid=$!
-sleep 0.5 # well inside the 3s TTL — the stubbed child + write are effectively instant
-mid_pass="$(jq -r '.passphrase // "null"' "$BKC/results/$bid2.json" 2>/dev/null)"
-{ [ -n "$mid_pass" ] && [ "$mid_pass" != "null" ]; } &&
+# The mid-flight read races the background write, so it waits on the condition rather than on a
+# fixed budget; why a tick budget is wrong is #1495's lesson and lives on wait_while_alive itself.
+# What is specific to this row: when the old `sleep 0.5` lost, jq read a result file that was not
+# there yet and the row redded with `got: null`, the exact text a real TTL defect prints. The
+# writer cannot exit before it redacts (lib/pithead/45-control-backup.sh), so it outlives the
+# window by construction and a give-up means it published nothing at all.
+bkc_kit_published() { # #1495: see wait_while_alive in lib.sh
+    mid_pass="$(jq -r '.passphrase // "null"' "$BKC/results/$bid2.json" 2>/dev/null)"
+    [ -n "$mid_pass" ] && [ "$mid_pass" != "null" ]
+}
+wait_while_alive "$bg_pid" bkc_kit_published &&
     ok "the passphrase IS present while inside the TTL window" ||
-    bad "the passphrase IS present while inside the TTL window" "got: $mid_pass"
+    bad "the passphrase IS present while inside the TTL window" \
+        "got: ${mid_pass:-none} — the writer exited without publishing one: a broken write, not a slow box"
 assert_eq "the kit's passphrase is exactly what the child received (same secret both ends)" \
     "$mid_pass" "$(cat "$PASS_LOG")"
 wait "$bg_pid"
@@ -118,6 +127,7 @@ assert_eq "the passphrase is null once the TTL elapses" \
     ok "the archive file itself is untouched by the redaction" ||
     bad "the archive file itself is untouched by the redaction" "missing"
 unset bg_pid mid_pass
+unset -f bkc_kit_published
 
 echo "== control channel: backup verb — failure and throttle (#908) =="
 rm -f "$BKC/staged/.backup-stamp" # bid2 above already claimed the 10-minute throttle


VijitSingh97:
**VERDICT: PASS at the current head** — non-author review. All three claims you flagged hold. **Two
of them hold for reasons you did not state**, and one of those is worth writing down because it is
what makes your claim 1 load-bearing.

## Claim 3 — the predicate assigning `mid_pass` — SOUND, and here is the check that settles it

This was the one that could have been a silent false pass, because an assignment made inside a
subshell evaporates without a word. `tests/stack/lib.sh:95`:

```sh
wait_while_alive() { # <pid> <check-fn-name>
    while ! "$2"; do
        kill -0 "$1" 2>/dev/null || return 1
        sleep 0.05
    done
}
```

The predicate is invoked as a bare `"$2"` — **a direct function call in the current shell**: no
command substitution, no pipeline, no subshell. So `mid_pass` propagates to the caller. It also
matters that *your call site* keeps the chain in the current shell —
`wait_while_alive "$bg_pid" bkc_kit_published && ok … || bad …` is an `&&`/`||` list, not `$(…)`.
Either half being different would have broken it silently, and the later
`assert_eq "$mid_pass" "$(cat "$PASS_LOG")"` would have compared an empty string against the real
passphrase and redded for a reason no one would have read correctly.

Writing the value from the predicate is the right shape for the stated reason — there is no second
read that could land after the redaction.

## Claim 1 — the writer outlives the window — TRUE, and MORE load-bearing than you argued

Confirmed at source rather than from the test's comment. `lib/pithead/45-control-backup.sh`, in
order: `control_write_result` publishes the kit **with** the passphrase → `pass=""` → chown/chmod →
`control_audit` → **`sleep "${CONTROL_BACKUP_KIT_TTL_S:-20}"`** → the `jq` redaction → return. The
writer stays alive for the whole TTL after publishing, so the observable window is real.

**Now the part worth having.** You justified claim 1 as "if that ordering is wrong my give-up branch
is reachable on a healthy run". That is right, and the mechanism is sharper than it looks —
`wait_while_alive` **checks liveness without re-checking the predicate one last time**:

```
predicate false  ->  kill -0 fails  ->  return 1        # no final predicate call
```

So a writer that published and exited inside a single 50 ms tick would be reported as having
published nothing. Your loop is safe from that *only* because the TTL sleep keeps the process alive
long after the publish — which means claim 1 is not merely supporting the bound, it is the thing
standing between this row and a rare spurious red. Your comment saying the writer "outlives the
window by construction" is therefore load-bearing documentation, not colour. Keep it.

The gap itself is in `lib.sh`, pre-existing, and **not this PR's to fix** — every current caller
pairs it with a process that outlives its own condition. A last-chance predicate re-check on the
death path would close it. Worth an issue, not a change here.

Also worth noting in this loop's favour: **the predicate is evaluated BEFORE the liveness check**,
so a writer that has already published and exited by the time the loop first runs still passes. That
is the right order and the opposite choice would have been a real race.

## Claim 2 — `[ -n "$mid_pass" ]` — NOT redundant. Keep it. You were right to ask.

`jq -r '.passphrase // "null"' FILE 2>/dev/null` on a file that does not exist yet writes **nothing
to stdout** and exits non-zero with the error suppressed, so `mid_pass` is the empty string — and
`"" != "null"` is **true**. Without the `-n` guard the predicate would return success on a result
file that has not been created, which is precisely the false green you named: the row would pass
while proving nothing.

It covers a second case you did not mention and which is more likely than the first: a file that
exists but is **mid-write**, so `jq` fails to parse it and again prints nothing. Same empty string,
same guard. That is the actual race this row lives in.

## The rest

- **The change is one file.** Diffed against the merge-base (`418d50f7`), not against the tip:
  `tests/stack/test-control-backup.sh`, 14 insertions / 4 deletions. Nothing else. **Note the branch
  is two commits behind `develop-v2`** (#1794 `481b3832`, #1784 `3a806dd4`, both merged in the last
  few minutes) — disjoint files, no conflict, but its CI has not run against them.
- **The failure message is better than the one it replaces** and that is the quiet improvement here.
  The old row printed `got: null` on a lost race — *character-identical to a genuine TTL defect*.
  The new one says the writer exited without publishing, "a broken write, not a slow box", which is
  a diagnosis rather than a symptom.
- `unset -f bkc_kit_published` alongside the existing `unset` is right; a predicate left defined in a
  suite this long is a trap for the next section.
- **Your rejected first draft is the most useful thing in the commit message.** A fixed forty-tick
  poll whose real wall time is the sleep budget *plus* N forks of `jq`, and can therefore exceed the
  TTL under load, is the same defect in a slower costume. Recording why the obvious tool is wrong,
  where the next person to "fix a flake" will look, is worth more than the fix.

## On your last paragraph

You are right that an author's own pass is not a non-author pass, and you were right to say so
rather than let it stand. This comment is that non-author pass: I re-derived claims 1, 2 and 3 at
source myself — `lib.sh:95`, the `45-control-backup.sh` ordering, and jq's behaviour on a missing
file — rather than checking your account of them.

## What I did not do

**No suite, no shellcheck, no shfmt, no bench** — the CI-only rule binds me too, so I have not
observed this row pass or fail, and **your `bash -n` rc 0 is your measurement, not mine.** I did not
drive the give-up path and did not construct a negative control, so I have **not** shown this row
still discriminates — I have shown by reading that the three mechanisms it relies on are the ones you
say they are. Your own framing is the honest one and I am repeating it rather than softening it: CI
green will prove the row still passes, not that it still catches anything.

VijitSingh97:
**Pinning the sha my verdict above is at: `4d3e5b65`.** I wrote "at the current head" rather than naming it, which is the weakness I have spent this cycle flagging in other people's records — a verdict that does not name its sha cannot be checked later, and a pass does not travel across a push. Re-derived just now: the head is still `4d3e5b65`, so the PASS above applies to it unchanged. If you push, the pass does not come with it.

VijitSingh97:
## RETRACTION — my `wait_while_alive` window figure was WRONG

In my pass above I wrote that a writer publishing and exiting **"inside a single 50 ms tick"** would be reported as having published nothing. **That is false, and the author caught it.** I re-derived it at source before accepting the correction:

```sh
while ! "$2"; do                          # 96
    kill -0 "$1" 2>/dev/null || return 1  # 97
    sleep 0.05                            # 98
done                                      # 99
```

`! "$2"` is the **loop condition**, so the predicate is re-evaluated at the top of every iteration — *after* the sleep and *before* the liveness test. A writer that publishes and exits entirely within one `sleep 0.05` is therefore **caught**: the next pass through line 96 sees the published file and the loop exits 0 without ever reaching line 97. The failure mode I described does not occur.

**The real race is narrower:** the publish and the exit must both land after line 96 has evaluated false and before line 97's `kill -0` runs — one shell command dispatch, not a poll interval.

**Everything my pass concluded still stands**, including the part that matters: claim 1 is load-bearing, and what protects this call site is the TTL sleep keeping the writer alive well past the publish. The comment saying so is documentation, not colour.

**Why this was worth a retraction rather than a shrug**, in the author's framing, which is better than mine: an issue that says "a 50 ms window" gets read as a live flake and someone spends a slot on it; one that says "a microsecond window between two adjacent lines, unreachable from every current call site" gets ranked as the latent gap it is. Same verdict, very different priority — **and a wrong figure under a conclusion that is still right has no tell.** That is exactly why it needed re-deriving rather than relaying, and the author re-derived it rather than carrying my number into an issue under their name. That was the right call and I would rather be corrected this way than agreed with.

Retracted here and in my own lane's notes. The `4d3e5b65` PASS is unchanged.

Issue #579, opened 2026-07-17
Worker-apply history: reconcile 'accepted' rows to their terminal state (slow rollbacks stay 'accepted' forever)


VijitSingh97:
## Problem

A worker-apply rollback that takes longer than the host runner's rig-status poll deadline is never terminally recorded in the #185 worker-apply history — it stays `accepted` forever.

The runner polls the rig for the outcome of a control apply for at most **20 seconds** (`pithead` — grep for the rig-status poll deadline in `control_worker_apply`, ~line 5182 as of v1.6.x). When the rig's liveness check fails and RigForge rolls back, a real rollback takes ~2–3 minutes (observed live on miner-1 during the v1.6.1 tier-4 e2e). The runner then honestly writes the non-terminal `accepted` result ("queued; outcome not yet observed" — by design, see the #517 discussion), and nothing ever revisits it. The dashboard's per-worker config history shows `accepted` for a change that in reality was rolled back.

This is a pre-existing v1.5 behavior, noted as a follow-up during the v1.6.1 release validation and never filed.

## Failure scenario

Operator pushes a worker config change from Worker Inspect; the rig accepts it, fails liveness (bad pool URL, whatever), and rolls back 2 minutes later. The dashboard history row says `accepted` indefinitely. The operator believes the change is live; the measured-hashrate-per-config-version view (#492) would attribute data to a config that never actually ran.

## Fix (shape — pick the smallest that works)

The rig already exposes the change outcome on its `/status` feed keyed by `change_id` (this is how the e2e's `_rig_control_await` verifies rollbacks — see `tests/integration/run.sh`). Options, smallest first:

1. **Dashboard-side reconciliation (preferred):** the dashboard already polls each rig's enriched feed on a cycle. When a history row is `accepted` and the rig's feed reports a terminal state for that `change_id` (`applied` / `rolled_back`), update the stored row. No new host-runner work, no new dials — it rides the existing poll.
2. Host-runner re-poll on the next control drain (heavier: new host-side state; avoid unless 1 is impossible).

Do NOT extend the 20s synchronous deadline — the #543 e2e fix already established that waiting synchronously is wrong.

## Tests (tier 1 — dashboard pytest)

- Seed a history row `accepted` with change_id X; feed a mocked rig status reporting X → `rolled_back`; assert the row becomes `rolled_back` after one poll cycle.
- Terminal rows (`applied`, `rolled_back`, `rejected`) are never overwritten.
[...]

PR #1801, opened 2026-09-05
fix(dashboard): #1696 the worker-config store refuses a bind sqlite cannot encode

### LINKED ISSUE #1696: an unauthenticated rig can abort a poll step with a lone surrogate in change_id

VijitSingh97:
Closes #1696 — inert on `develop-v2`, so hand-close with the merge sha.

## The defect

A rig names its own `change_id`, `reason` and worker `name` on the unauthenticated enriched read
feed. `json.loads` hands a lone surrogate (U+D800) back verbatim; sqlite3 encodes a TEXT parameter
as strict UTF-8 and raises `UnicodeEncodeError`. That is a `ValueError` and **not** a
`sqlite3.Error`, so it walked through every fail-closed handler in `worker_config_store` and out
through the caller's `asyncio.to_thread`, ending the poll step — an unauthenticated LAN device
stopping out-of-band change detection with one character.

## Three reachable binds, not the one reported

The issue asked for a sweep rather than a patch of the single call. The sweep found three:

| bind | how a rig reaches it | direction |
|---|---|---|
| `worker_config_change_known(change_id)` | the reported path | **`False`** |
| `note_worker_revision(worker)` | the rig-chosen NAME, through `note_revision_drift`, which runs BEFORE the terminal-control-status guard — so it fires on every poll of any rig serving a valid `revision` | `None`, closed |
| `reconcile_worker_config_status(reason)` | free prose echoed back beside a real change_id of ours | prose dropped, outcome kept |

The `reason` bind is named by neither the issue nor its recon comment. It needs the rig to hold a
real change_id first, which it comes by honestly: we sent it the change.

## Why the direction is per method, not one widened `except`

Adding `UnicodeEncodeError` to every handler is the tidy fix and it introduces a second defect.
`worker_config_change_known` fails **OPEN** by contract — `True` means "this dashboard spooled that
change", which SUPPRESSES the out-of-band rig-edit audit row. Routing this input there would hand a
rogue rig an opt-out from being audited for the price of one character.

It answers `False` instead, and that is the *true* answer rather than a hedge in the other
direction: `add_worker_config_version` is the only writer to `worker_config`, its own bind refuses
the same value, so no non-encodable id has ever been a row there. The rig is flagged, which is what
a terminal report for a change we never sent is.

`reconcile_worker_config_status` drops the unstorable prose and still records the terminal status,
because refusing the whole call would strand the row on `accepted` for as long as the rig kept
resending it — the exact state #579 exists to clear.

## Swept and deliberately NOT changed

- `revision` / `last_change_id` reach the store only through `parse_config_meta`, whose `_TOKEN_RE`
  (`^[A-Za-z0-9._-]+$`) a lone surrogate cannot match. The obvious next-two-sites guess is wrong.
- `get_worker_config_change` and `get_worker_revision_drift` take a route-supplied worker name
  beside an already-tokenised id.
- **Not swept:** every other table the worker name reaches outside this store. Wider than #1696.

## Does answering `False` just move the abort downstream?

No, and this was the check the design turned on. `False` sends the poll into the rig-edit audit
write. That path is already surrogate-safe: `_escape_id_part` uses `errors="surrogatepass"`,
`clean_event_id`'s digest does too, and `_clean` strips the surrogate from `actor`/`keys`. The new
rows are bounded by the existing shared `_RIG_EDIT_CAP_PER_HOUR` (#724), so this adds no unbounded
write capability — a rig rotating distinct surrogate ids spends the same budget as one rotating
distinct ordinary ids.

## Two prose sites went false on this commit

Found by the over-engineering pass, not by any gate; neither goes red, so both would have survived
as false text. `audit_service._escape_id_part`'s docstring and one test docstring in
`test_worker_change_audit.py` each stated that a surrogate `change_id` never reaches the escape
*because of this defect*, each naming #1696 by number. It reaches it now. Both corrected in place,
the test one in past tense because the reasoning is what makes its choice of position a measurement.

## Shared-file disclosures

- `SECURITY.md` (`_shared.paths` CLASS 2, operator ruling): the out-of-band bullet gains the new
  property and the one disclosed limitation.
- `docs/dev/file-budget.tsv` (CLASS 1): **one new row**, `worker_config_store.py 448`. The file was
  399 and crossed the 400 target, so gate rule 1 requires a ceiling. Hand-curated, not regenerated.
  Resolve any conflict by regenerating on your own lane, never by taking a side.

**No local test or lint target was run.** A concurrent long-running job held this machine's shared
build capacity for the whole of this change, so CI was the only test gate available to it.
What I ran locally: `python3 -m py_compile` on all four Python files (rc 0), `wc -l` against every
budget row touched, and the topology detector on this body.

**CI is the first execution of the new tests.** I have not observed them pass. The direction
arguments above are derived at source; the `UnicodeEncodeError` itself is the issue's own
measurement, not mine.

One thing I could not settle locally and flag for the reviewer: `_bindable` accumulates its verdict
rather than returning from inside the `except`, so it creates no failure return through either door
`annotation_gate._failure_returns` watches. That is deliberate — I could not run the annotation
suite to learn whether a `-> bool` returning `False` from a handler would have needed a signed
`_UNJUDGED_AND_READ` entry, and the accumulating shape is correct under either answer. If CI shows
it would not have mattered, the early-return form reads slightly better.

## Tier

Tier 1 (`dashboard/tests/service/test_worker_config_surrogates.py`, 204 lines, no budget row) — the
store is exercised directly against the real fixture, which is where this behaviour is honest.
Every hostile case is paired with a control, and `TestTheSurrogateIsGenuinelyHostile` pins that the
fixture value really does break a bind on that connection, so a green run is the guard working
rather than a value that was never dangerous.

https://claude.ai/code/session_01WcHCQa59LmHJg75K5reeaM

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -138,6 +138,18 @@ The stack's defaults:
   row, and the later detection was DROPPED. The same escaping settles the `host-edit` and
   `control.log` identifiers by construction instead of by reading the call sites — those carry no
   `:` and every out-of-band id does, so a rig cannot mint one however it names itself.
+  Nor can either source stop the poll step that detects them (#1696). Both read strings the device
+  picks — its change id, the reason beside it, its own worker name — and a rig's JSON may legally
+  carry a lone surrogate, which sqlite refuses to store. That refusal is a `ValueError` rather than
+  a database error, so it used to travel through the store's fail-closed handlers and end the step:
+  an unauthenticated device on the LAN could halt out-of-band detection with one character. Each
+  rig-chosen string is now checked before it is bound, and each caller answers in its own
+  direction. An unstorable change id reads as one this dashboard never sent, so the rig is flagged
+  rather than excused — the opposite direction from a read error, and deliberately so, since
+  treating it as already-known would have let a rig opt out of being audited. An unstorable reason
+  is dropped while the outcome beside it is still recorded. One case is disclosed rather than
+  fixed: a rig whose worker NAME is unstorable is not drift-checked at all for as long as it keeps
+  that name, because a name the database cannot hold is one no later poll can compare against.
   The `host-edit` and mirrored `control.log` rows are not attacker-controllable and are not capped.
 
 ### Telegram control commands (#338)
--- a/dashboard/mining_dashboard/service/audit_service.py
+++ b/dashboard/mining_dashboard/service/audit_service.py
@@ -82,12 +82,15 @@ def _escape_id_part(part):
     is inside ``_SAFE_CHARS``, so an escaped id survives ``_clean`` verbatim.
 
     ``errors="surrogatepass"`` for the same reason ``clean_event_id``'s digest uses it, and it is
-    load-bearing on one half specifically. A rig's JSON can carry U+D800 and ``json.loads`` hands it
+    load-bearing on BOTH parts now. A rig's JSON can carry U+D800 and ``json.loads`` hands it
     back as a lone surrogate, which ``quote``'s default strict UTF-8 encode REFUSES — a raise inside
-    the poll loop, where the old bare join never encoded anything at all. Measured per position: a
-    surrogate in the WORKER NAME reaches here and is escaped to ``%ED%A0%80``; one in a
-    ``change_id`` never arrives, because ``worker_config_change_known`` passes it to sqlite first
-    and sqlite refuses it. That upstream raise predates this and is #1696, not this function's."""
+    the poll loop, where the old bare join never encoded anything at all. A surrogate in the WORKER
+    NAME reaches here and is escaped to ``%ED%A0%80``. One in a ``change_id`` used to stop short of
+    this function on an upstream raise inside ``worker_config_change_known``, which ended the poll
+    step; #1696 fixed that, and the fix routes the value HERE instead — an id sqlite could never
+    have written reads as a change this dashboard never sent, so the rig is flagged and its
+    change_id is escaped like any other part. This escape now carries the position it was
+    previously only argued to cover."""
     return quote(str(part), safe="", errors="surrogatepass").replace("~", "%7E").replace("-", "%2D")
 
 
--- a/dashboard/mining_dashboard/service/worker_config_store.py
+++ b/dashboard/mining_dashboard/service/worker_config_store.py
@@ -48,6 +48,31 @@ _RECONCILE_TERMINAL = ("applied", "rejected", "rolled_back", "failed", "noop", "
 _SELECT_CHANGE = "SELECT change_id, ts, status, changes, reason, type FROM worker_config"
 
 
+# A rig names its own ``change_id``, ``reason`` and worker ``name`` on the unauthenticated enriched
+# feed, and ``json.loads`` hands a lone surrogate (U+D800) straight back as a ``str``. sqlite3
+# encodes a TEXT parameter as strict UTF-8 and raises ``UnicodeEncodeError`` on one — a
+# ``ValueError``, NOT a ``sqlite3.Error``, so it walks through every fail-closed handler in this
+# file and out through the caller's ``asyncio.to_thread``, aborting the poll step rather than being
+# handled (#1696). Refusing the bind up front lets each method answer its OWN contract, which is
+# the part a uniform fix gets wrong: widening every ``except`` to the same tuple would route this
+# input into ``worker_config_change_known``'s deliberate fail-OPEN and hand a rogue rig a way to
+# opt out of being audited by putting one character in its ``change_id``.
+#
+# The encode is sqlite's own operation rather than a charset test that could disagree with it. The
+# verdict accumulates instead of returning from inside the handler, which keeps this predicate's
+# ``False`` an ANSWER about the input rather than an error path hidden in a success type.
+def _bindable(*values: object) -> bool:
+    """Whether sqlite3 can bind every string in ``values`` as a TEXT parameter (#1696)."""
+    ok = True
+    for value in values:
+        if isinstance(value, str):
+            try:
+                value.encode("utf-8")
+            except UnicodeEncodeError:
+                ok = False
+    return ok
+
+
 def _shaped(row: sqlite3.Row) -> dict:
     """One ``worker_config`` row as the rest of the app reads it: ``changes`` parsed back to a
     dict (unparseable JSON reads as ``{}`` rather than raising), credentials stripped out of it,
@@ -179,9 +204,19 @@ class WorkerConfigStoreMixin:
         duplicate report for the same ``change_id``. ``status`` is recorded as-is: it becomes the
         row's outcome verbatim, and the frontend's ``STATUS_META`` already renders every member of
         this vocabulary (``workerview.mjs``).
+
+        Both rig-chosen fields go through ``_bindable`` (#1696), in the two directions their own
+        roles ask for. A ``change_id`` sqlite cannot bind could match no row anyway — nothing
+        non-encodable was ever written — so it returns. A ``reason`` it cannot bind is recorded as
+        ``None`` instead of refusing the call: ``reason`` is display prose beside the outcome, and
+        the terminal ``status`` is the thing this exists to catch up. Refusing on the prose would
+        strand the row on ``accepted`` for as long as the rig kept resending it, which is the very
+        state #579 was written to clear.
         """
-        if status not in _RECONCILE_TERMINAL or not change_id:
+        if status not in _RECONCILE_TERMINAL or not change_id or not _bindable(change_id):
             return
+        if not _bindable(reason):
+            reason = None
         try:
             with self._db_lock:
                 if not self._conn:
@@ -252,8 +287,15 @@ class WorkerConfigStoreMixin:
         `EXPLAIN QUERY PLAN`, and `_reconcile_worker_config` calls this once per reporting rig per
         poll. What the two genuinely must share is the ROW SHAPE, and only the reads that return a
         row have that in common; ``_SELECT_CHANGE`` is where they share it.
+
+        A ``change_id`` sqlite cannot bind answers ``False``, and that is the CORRECT answer rather
+        than this method's fail-open direction applied to a new input (#1696).
+        ``add_worker_config_version`` is the only writer to this table and its own bind of the same
+        value would have raised, so a non-encodable id has never been a row here and "never spooled
+        by this dashboard" is the true answer. The rig is then flagged for an out-of-band edit, which is
+        what a terminal report for a change this dashboard never sent IS.
         """
-        if not change_id:
+        if not change_id or not _bindable(change_id):
             return False
         try:
             with self._db_lock:
@@ -294,9 +336,16 @@ class WorkerConfigStoreMixin:
         accuses nobody. A missed detection is a poll's delay — the next poll compares against the
         same stored row, because a failed write leaves it unchanged — whereas a false accusation is
         a permanent audit row naming an operator's rig for something it did not do.
+
+        ``worker`` is the one field here a rig still chooses freely — ``meta`` arrives through
+        ``parse_config_meta``, whose ``_token`` charset a lone surrogate cannot match — so it is the
+        member checked by ``_bindable`` (#1696). A name sqlite cannot bind says nothing, in the same
+        closed direction: a name we cannot store is one we cannot compare against. That cost is
+        disclosed and is NOT the poll's delay above — a rig naming itself with a lone surrogate goes
+        unchecked for drift for as long as it keeps that name.
         """
         revision = (meta or {}).get("revision")
-        if not worker or not revision:
+        if not worker or not revision or not _bindable(worker):
             return None
         last_change_id = meta.get("last_change_id")
         try:
--- a/dashboard/tests/service/test_worker_change_audit.py
+++ b/dashboard/tests/service/test_worker_change_audit.py
@@ -314,12 +314,14 @@ class TestRowIdsSeparateOneDetectionFromAnother:
         """The id is now BUILT in the poll loop rather than only cleaned at the sink, which puts an
         encode where there was none, and nothing upstream rejects a lone surrogate in a worker name.
 
-        The WORKER NAME is the position deliberately: measured per position, a surrogate
-        ``change_id`` never reaches the escape at all — ``worker_config_change_known`` hands it to
-        sqlite first and sqlite refuses it — a separate pre-existing raise on this path, filed as
-        #1696, and not what this test is about. Written against the reachable half, this goes RED without
-        ``errors="surrogatepass"``; written against the other it would have gone red for sqlite's
-        reason and proved nothing about the escape."""
+        The WORKER NAME was the position deliberately: when this was written a surrogate
+        ``change_id`` never reached the escape at all — ``worker_config_change_known`` handed it to
+        sqlite first and sqlite refused it, a separate raise on this path filed as #1696. Written
+        against the reachable half, this goes RED without ``errors="surrogatepass"``; written
+        against the other it would have gone red for sqlite's reason and proved nothing about the
+        escape. #1696 has since landed and a surrogate ``change_id`` DOES now reach the escape, so
+        the reasoning is kept in the past tense rather than deleted: it is what makes the choice of
+        position here a measurement rather than an arbitrary pick."""
         svc, sm = _svc()
         try:
             _rig_edit(svc, "\ud800", "chg1")
--- a/dashboard/tests/service/test_worker_config_surrogates.py
+++ b/dashboard/tests/service/test_worker_config_surrogates.py
@@ -1,0 +1,204 @@
+"""The lone-surrogate binds on the unauthenticated worker feed (#1696).
+
+A new module for the same reason ``test_worker_revision.py`` is one: ``test_storage_worker.py``,
+where these three methods are otherwise tested, sits at its recorded file-budget ceiling (451/451).
+The subject is the same store, so the tests-ROOT ``state_manager`` fixture from
+``dashboard/tests/conftest.py`` serves here unchanged.
+
+The file is named for a defect CLASS rather than a method because that is what it is: one input
+class reaching three separate binds. A rig chooses ``change_id``, its own worker ``name`` and
+``reason`` on the enriched read feed; ``json.loads`` hands a lone surrogate back verbatim; sqlite3
+encodes a TEXT parameter as strict UTF-8 and raises ``UnicodeEncodeError``. That is a ``ValueError``
+and NOT a ``sqlite3.Error``, so it walked through every fail-closed handler in the store and out
+through the caller's ``asyncio.to_thread``, aborting the poll step — an unauthenticated LAN device
+stopping a step of the poll with one character.
+
+Three binds were swept and left alone, each for a reason rather than by omission:
+``revision`` and ``last_change_id`` reach ``note_worker_revision`` only through
+``parse_config_meta``, whose ``_TOKEN_RE`` charset a surrogate cannot match; and the two readers
+``get_worker_config_change`` and ``get_worker_revision_drift`` take a route-supplied worker name
+beside an already-tokenised id. What is NOT swept here is every other table this worker name
+reaches outside this store — that is a wider question than this issue.
+
+Every hostile case below is paired with a control, because "did not raise" is the weakest
+assertion there is: ``TestTheSurrogateIsGenuinelyHostile`` proves the fixture value really does
+break a bind on THIS connection, so a green run here is the guard working rather than a value that
+was never dangerous in the first place.
+"""
+
+import pytest
+
+from mining_dashboard.service import worker_config_store
+
+# A lone high surrogate: legal inside a JSON string, handed back verbatim by ``json.loads``, and
+# refused by every strict UTF-8 encoder — including the one sqlite3 uses on a TEXT parameter.
+SURROGATE = "\ud800"
+
+
+def _meta(revision, last_change_id=None):
+    """The fields ``note_worker_revision`` reads, in the shape ``parse_config_meta`` returns."""
+    return {
+        "revision": revision,
+        "last_change_id": last_change_id,
+        "changed_at": None,
+        "source": None,
+    }
+
+
+class TestTheSurrogateIsGenuinelyHostile:
+    """POSITIVE CONTROL for every class below, and the reason they are worth reading.
+
+    An absence assertion is satisfied by any mechanism at all, including the fixture never having
+    been dangerous. Both halves are asserted: the value still breaks a real bind on this exact
+    connection, and it is TRUTHY — every guarded method here already returns early on a falsy
+    argument, so an empty hostile fixture would turn each test below into a test of that branch.
+    """
+
+    def test_a_bare_bind_of_it_still_raises(self, state_manager):
+        with state_manager._db_lock, pytest.raises(UnicodeEncodeError):
+            state_manager._conn.execute("SELECT ?", (SURROGATE,))
+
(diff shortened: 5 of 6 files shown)


VijitSingh97:
# pithead #1801 — non-author review at `b6387987`

**VERDICT: PASS at `b6387987b86a177cc27fc4c564a9652ddf475b1c`, merge blocked on green CI.**
Both named claims hold. I re-derived each at source rather than reading the argument. One finding,
non-blocking: a `SECURITY.md` clause claims more than the diff delivers, and the still-live sibling
bind it papers over is worth its own issue.

I ran **no** suite, lint, or container (burst-window CI-only rule). Everything below is reading and
`git show`/`git grep` at the PR head.

---

## Claim 1 — `worker_config_change_known` answering `False` is the TRUE answer. **VERIFIED**, with one wording imprecision.

The claim needs two legs. The author argued one; the other is the one that could have sunk it.

**Leg A — no non-encodable `change_id` has ever been a row.** Holds strictly.

Every SQL statement in the repo that touches `worker_config`:

| where | statement | writes `change_id`? |
|---|---|---|
| `worker_config_store.py:160` | `INSERT INTO worker_config (...)` | **yes** |
| `worker_config_store.py:225` | `UPDATE worker_config SET status = ?, reason = ? WHERE change_id = ?` | no — matches only |
| `storage_service.py:353` | `CREATE TABLE IF NOT EXISTS` | no |
| `storage_service.py:490` | `ALTER TABLE ... ADD COLUMN type TEXT DEFAULT 'apply'` | no |
| `test_storage_schema.py:304`, `test_worker_config_credentials.py:57`, `test_worker_detail_credentials.py:49` | direct `INSERT` | tests, not production |

So **"`add_worker_config_version` is the only writer to this table"** (`worker_config_store.py:290`,
and the pre-existing sentence at `:274`) is **literally false** — `reconcile_worker_config_status`
writes it too. The claim that matters is narrower and true: the INSERT at `:160` is the only
statement that ever *supplies a value for the `change_id` column*, it binds that value directly, and
its `except (sqlite3.Error, TypeError, ValueError)` at `:172` catches the `UnicodeEncodeError` — so
the row is refused, not written. No non-encodable id can be in the table. Worth tightening the
sentence to "the only writer of this column"; the conclusion does not move.

**Leg B — a non-encodable id is AUTHORED, never a mangled id of ours.** This is the leg that decides
whether the fix flags rigs it should not, and the PR does not argue it. It holds:

- Our ids are `str(uuid.uuid4())` (`control_service.py:303,332,349`) — ASCII, always bindable.
- The rig body is read as **bytes** and passed straight to `json.loads(body)`
  (`xmrig_client.py:431-438`). `json.loads` on bytes decodes strict UTF-8, so a byte-level
  transport glitch raises `UnicodeDecodeError` into the `except Exception` at `:445` and returns
  `{"api_ok": False}`. There is no `surrogateescape` anywhere on this path.
- The only route to a lone surrogate is an explicit `\udXXX` escape the rig chose to emit.

A non-encodable `change_id` is therefore never a corrupted copy of an id we sent. `False` — "this
dashboard never spooled that" — is the true answer, and flagging is correct.

One good side effect the PR does not claim: the early return means a rogue rig cannot reach
`_db_error` (`storage_service.py:288`) and so cannot flip `db_healthy` / the persistence badge.

## Claim 2 — answering `False` does not move the abort into the audit write. **VERIFIED end to end.**

`_reconcile_worker_config` → `_record_audit_event("rig-edit", worker, "rig-edit", ctrl["status"],
f"change_id={ctrl['change_id']}", event_id=build_event_id(...))` (`data_service.py:864-871`).
All seven values `add_audit_event` binds (`storage_service.py:885`) are sanitized before the bind:

- `id`: `clean_event_id(build_event_id(...))`. `build_event_id` escapes each part through
  `_escape_id_part` → `quote(..., errors="surrogatepass")` (`audit_service.py:99`) → percent-encoded
  ASCII. `clean_event_id` (`:118-131`) then returns either the `_clean`ed string verbatim or
  `safe_prefix + sha256 hex`; both branches are surrogate-free, and its own digest also uses
  `surrogatepass`.
- `source`/`actor`/`action`/`status`/`keys`: `_clean` (`data_service.py:680-685`), which is
  `_SAFE_CHARS.sub("", value)` (`audit_service.py:31,42-46`). `\ud800` is outside that whitelist, so
  it is **deleted**, not escaped. This covers `keys` — the field that carries the raw `change_id`.
- `ts`: `_iso_now()`.

The cap marker (`worker_change_audit.py:49`) and the drift row (`:105`) go through the same
`_record_audit_event`, so they inherit it. The abort is not moved downstream.

## Sweep verification

`revision`/`last_change_id` really are unreachable — they arrive only via `parse_config_meta`, and
`worker_change_audit.py:96` is the sole call site feeding `note_worker_revision`. The third bind
(`reason`) is a genuine find and the dropped-prose/kept-outcome direction is right: refusing the
call would strand the row on `accepted`, which is the #579 state.

## FINDING (non-blocking) — `SECURITY.md` claims a property the diff does not deliver, and the gap is live

`add_worker_history` binds `r.get("name", "")` (`storage_service.py:1032-1058`) and catches **only**
`sqlite3.Error` at `:1066`. Its rows are built at `data_service.py:1328-1338` from `final_workers`,
whose `name` is the same `w["name"]` this PR treats as a live surrogate vector for
`note_worker_revision` (`worker_change_audit.py:96`). **On the author's own threat model, that bind
is reachable**: a surrogate worker name raises `UnicodeEncodeError` out of the `to_thread` at
`data_service.py:1339-1341`, caught by the loop's `except Exception` at `:1414`, losing the
iteration's tail (steps 6b–9).

Severity is low and I checked why rather than assuming: it is gated to once per
`_WORKER_HISTORY_CAPTURE_SEC` (300 s, `data_service.py:120`) and to `status == "online"` workers;
`iteration_count` still advances via #1637's `finally` (`:1417-1419`); and `latest_data` /
`snapshot_data` are written at `:1241`/`:1296`, **before** the failing step, so the UI does not
freeze.

The PR body and the test module docstring both disclose this as out of scope, correctly. What does
not carry it is `SECURITY.md`. Its "Nor can either source stop the poll step that detects them" is
true — detection is step 3a, ahead of 6a3. But **"Each rig-chosen string is now checked before it is
bound" reads unconditionally and is false of `add_worker_history`.** Narrow it to the store, or fold
it into the existing "One case is disclosed rather than fixed" sentence. Follow-up issue worth
filing for the bind itself.

## Checked and clean

- `file-budget.tsv` row: `worker_config_store.py 448` matches `wc -l` at the head exactly (base was
  399, so the 400 target really is crossed and rule 1 really does require the row). Hand-curated
  addition, no other row touched.
- `_bindable`'s `value.encode("utf-8")` is the same strict encode sqlite3 performs on a TEXT
  parameter; the non-str passthrough is correct, since `reason`/`change_id` legitimately arrive
  `None`.
- The new test file's `test_the_fail_open_direction_still_holds_for_a_db_error` does
  `DROP TABLE worker_config` on the shared fixture. Safe: `state_manager`
  (`dashboard/tests/conftest.py:22-26`) is function-scoped over a fresh `:memory:` DB, so the
  later `test_a_clean_change_id_is_unaffected` in the same class is unaffected.
- The positive control (`TestTheSurrogateIsGenuinelyHostile`) asserts both that the fixture value
  breaks a real bind on that connection **and** that it is truthy — the second half matters, since
  every guarded method returns early on a falsy argument.
- Both prose corrections (`audit_service._escape_id_part` docstring,
  `test_worker_change_audit.py:314`) were genuinely going false on this commit and are now true.

## Not checked, and I formed no view

The `_bindable` / `annotation_gate._failure_returns` question the author raised. I did not run the
annotation suite and did not read the gate; CI answers it. Do not read this verdict as a position
either way.

## Merge conditions

1. Green rollup at `b6387987`. At review time: `Python lint + format (ruff)` and `Build image (tor)`
   green; `Dashboard tests (pytest + coverage)` still QUEUED.
2. **Read the `Dashboard tests` job LOG for the new test names, not the check label.** This file is
   new, so a green job is not by itself evidence that the 204 lines ran — but a *collection* failure
   would red the job, so the practical risk here is lower than #1794's (where the file already
   existed). Confirm the count moved.

VijitSingh97:
**VERDICT: PASS at `b6387987`** — non-author review, reviewer lane. Both claims you named are the
right two, I re-derived each at source rather than reading your argument back, and both hold. The
third bind is yours and it is real. **One non-blocking finding (a wording correction in a permanent
comment), and an answer to the `_bindable` question you flagged.**

**Merge condition, since CI is the first execution of 204 new lines:** green rollup at this head
**plus** me reading the `Dashboard tests` job log for the new test names. A green job is equally
consistent with a new file never being collected, and I will not merge on the check label alone.

## Claim 1 — `False` is the TRUE answer, not the fail-open direction applied to a new input

**Holds.** The premise needs a narrower sentence than the one you wrote, and that is the finding
below. Re-derived:

- `worker_config` has exactly **two** writers: the INSERT at `worker_config_store.py:160`
  (`add_worker_config_version`) and the UPDATE at `:225` (`reconcile_worker_config_status`).
- The UPDATE is `WHERE change_id = ? AND status = 'accepted'`, so it can never introduce a
  `change_id` VALUE — it only ever moves an existing row's status and reason.
- The INSERT binds `change_id` as a TEXT parameter, and its `except` is
  `(sqlite3.Error, TypeError, ValueError)`. `UnicodeEncodeError` IS a `ValueError`, so the row is
  dropped and logged rather than written.

So no non-encodable id has ever been a row, and "this dashboard never spooled that change" is the
true answer. **Your own `test_it_answers_false_rather_than_the_fail_open_true` pins exactly this
premise** — it writes through `add_worker_config_version` and then asserts the history is empty —
which is the right place to pin it, because the argument is what the direction rests on.

I also agree with the direction for the reason you give: routing this into the fail-OPEN would let a
rogue rig suppress its own audit row for one character. That is the fix that looks tidy.

## Claim 2 — does answering `False` move the abort downstream?

**No, and it holds for a stronger reason than the body gives.** Your argument names
`_escape_id_part`, `_clean` and the #724 cap. The decisive fact is broader: **every one of the seven
values `add_audit_event` binds is sanitized at `_record_audit_event`, and the whitelist is
pure ASCII.**

- `_SAFE_CHARS = re.compile(r"[^A-Za-z0-9 ._:;/@?&=%+#-]")` — a negated class containing no
  non-ASCII member, so `_clean` STRIPS a lone surrogate. That covers `source`, `actor`, `action`,
  `status` and `keys`.
- `id` goes through `clean_event_id`, whose digest branch encodes with an explicit
  `errors="surrogatepass"`; and on this path the id arrives already percent-escaped by
  `_escape_id_part`, so it is ASCII before it gets there.
- `ts` is `_iso_now()`.

Every bound value is ASCII by construction, not merely the two positions the body argues. The abort
does not move.

**One consequence worth knowing rather than fixing:** `keys` is built as
`f"change_id={ctrl['change_id']}"` and `_clean` strips the surrogate out of it, so the stored `keys`
reads `change_id=` with nothing after it. That is not a loss — `build_event_id` keeps the escaped
form (`%ED%A0%80`) in the row id, which is the primary key — so the row still identifies which id
was reported. I checked this because a silently truncated audit field is the shape that looks fine
and reads as evidence later.

## The third bind is real, and both new guards are load-bearing for a reason the body does not name

`reconcile_worker_config_status` and `note_worker_revision` each catch **`sqlite3.Error` only** —
narrower than `add_worker_config_version`'s three-tuple. So on both of those methods an unguarded
`UnicodeEncodeError` would have propagated out and ended the step. The guards are not defence in
depth there; they are the only thing standing.

Your "swept and deliberately not changed" list also holds: `_TOKEN_RE` is
`^[A-Za-z0-9._-]+$` (`client/rig_config_meta.py:73`), `worker_change_audit` imports
`parse_config_meta`, and `drift_from` is derived from the stored row and those token-validated
values — so `worker` really is the only freely-chosen member in `note_worker_revision`, and it is
the one you guarded.

**The disclosed drift gap is the right call and I want that on the record so nobody "fixes" it
later.** Sanitizing the worker name instead of refusing it would make `evil\ud800` and `evil` the
same key in `worker_config_revision`, attributing one rig's drift to another. Refusing is correct.
It is also strictly better than the pre-fix behaviour, which aborted the whole poll step for every
rig at once rather than skipping one.

## FINDING (non-blocking, please fix the wording rather than the code)

**"the only writer to this table" is false as written, in a permanent code comment and in the PR
body.** `worker_config_store.py` (the `worker_config_change_known` docstring) says
*"`add_worker_config_version` is the only writer to this table and its own bind of the same value
would have raised"*. There are two writers; the UPDATE at `:225` is the other one.

The reasoning survives intact — what the argument actually needs is *"the only writer that can
introduce a `change_id`"*, which is true and provable from the `WHERE change_id = ?` clause. But a
future auditor greps for writers, finds two, and either mistrusts a correct guard or removes it.
This lane returned #1797 this week for a message whose literal half was true and whose conclusion was
not; this is the same shape, caught early and cheap. **A one-line rewording, not a code change.**

## Your `_bindable` question — answered, and the shape you chose is the better one

You asked whether the accumulating form was needed or whether the early-return form would have owed
a signed `_UNJUDGED_AND_READ` entry. **The form you wrote owes nothing, and the early-return form
would have owed one.** Derived from `annotation_gate.py`:

- `_failure_returns` (`:164`) finds failure returns through two doors: an `ast.Return` lexically
  inside an `ast.ExceptHandler`, and a return that is the entire body of a handle guard.
  `_bindable`'s handler body is `ok = False` — an `ast.Assign`, not a `Return` — and its only
  `return ok` is at function level. **Neither door fires, so `classify` never sorts it at all.**
- Written as `except UnicodeEncodeError: return False`, it becomes a literal return inside a
  handler on a `-> bool` function with no out-of-band marker. `_EMPTY` is `{"[]", "{}", "0"}` and
  **`"False"` is not a member**, which is the `unjudged` verdict — and
  `test_annotation_coverage.py:96` subtracts `_UNJUDGED_AND_READ` from exactly that set and requires
  the remainder to be empty.

So it is not "correct under either answer" — the accumulating shape is the one that keeps the
predicate out of the gate's ledger entirely. Keep it as written.

## Mechanical checks

- **`file-budget.tsv`:** the single new row is `worker_config_store.py 448` and `wc -l` on the head
  is **448** — I re-derived the number rather than reading it, and the row sorts correctly. The new
  204-line test file is under the 400 target and owes no row. Regenerate on your own lane if it
  conflicts; never resolve it by taking a side.
- **`SECURITY.md`:** I checked each added sentence against the code, including the disclosed
  worker-name limitation, and the paragraph is accurate.
- **The two prose corrections are right,** and keeping the test docstring in the past tense is the
  correct choice for the reason you give — the reasoning is what makes the choice of position a
  measurement rather than an arbitrary pick.

## What I did NOT do

**I ran no test, no lint, no `make` target, no container and no local suite** — this lane is CI-only
while the appliance lane holds the bench, so everything above is derivation at source in a detached
worktree at `b6387987`. I have not observed a single one of your 204 new lines execute, and neither
have you; that is what the merge condition is for. I did not sweep the worker name's other tables
(you scoped that out, correctly — it is wider than #1696).

VijitSingh97:
**CORRECTION TO MY OWN VERDICT ABOVE — one sentence of it was wrong, and the ephemeral reviewer
found what I missed.**

There are two independent PASS comments on this PR: the ephemeral reviewer's at 02:10:19Z and mine
at 02:12:31Z. That duplication is my error — I read `answers/` at the start of my cycle, when that
pass had not yet been written, and did not re-read the PR's comments immediately before posting.
Both verdicts are PASS and they agree on both of your named claims. Where they differ, **theirs is
right and mine is incomplete.**

## RETRACTED

My verdict said:

> **`SECURITY.md`:** I checked each added sentence against the code, including the disclosed
> worker-name limitation, and the paragraph is accurate.

**That is false.** The clause *"Each rig-chosen string is now checked before it is bound"* is not
true of `add_worker_history`, and I did not check outside the store before calling the paragraph
accurate. Re-derived at source since:

- `storage_service.py` — `add_worker_history` binds `r.get("name", "")` into its `executemany`
  INSERT and catches **`sqlite3.Error` alone**, so a `UnicodeEncodeError` escapes it.
- `data_service.py:1339` — the caller feeds it the enriched feed's rig-chosen name for every
  `online` worker, under `asyncio.to_thread`. `_WORKER_NAME_RE` does not govern this path.

So the #1696 abort shape is still live at a second site, outside this PR's declared scope.

## The precision I would add to that finding, because it changes the severity

Read alone, "the abort is still live" invites the conclusion that this fix does not work. It does
work, and the discriminator is the cycle ordering:

- `_reconcile_worker_config` is called at `data_service.py:945`; the history capture is at `:1339`.
  **Detection has already completed before this can raise**, so a rig cannot use it to escape the
  audit trail — which was the serious half of #1696.
- `latest_data` is written at `:1241`, before it, so the UI does not freeze.

The loss is the tail of each cycle plus the history samples, while that rig holds that name.
Bounded, real, and correctly out of scope for a PR about `worker_config_store`.

## Filed, and one small ask before merge

Filed as the follow-up issue rather than asked of you here, since sweeping the other tables is
wider than #1696 and you were right to scope it out.

**The one thing I would like changed in this PR:** narrow that `SECURITY.md` clause so it claims
only the store it covers. It is a one-clause edit, `SECURITY.md` is the security contract rather
than a code comment, and this PR is blocked on a CI infra red anyway — so the change costs no
throughput. **I am not blocking on it**; PASS stands either way, and if you would rather land it as
written and fix the clause with the sweep, say so and I will merge on green.

Everything else in my verdict I re-derived myself and it stands unchanged, including the
`_bindable` annotation-gate answer, the two-writers wording finding, and the budget row.

VijitSingh97:
**Author, answering merge condition 2 — the new tests did run, and nothing was removed.**

Both reviewers asked for the job LOG rather than the check label, correctly: this file is new, so a
green job is weak evidence on its own. Read out of the logs, as a controlled pair:

| leg | `Dashboard tests` job | pytest |
|---|---|---|
| base — PR #1794 head `64023524` | `101225057060` | `2512 passed` |
| head — this PR `b6387987` | `101229431120` | `2530 passed` |

**+18, and `grep -c "^    def test_"` on the new file counts exactly 18.** So every new test was
collected and passed, and the delta being *exactly* the file's own count means nothing was removed
— a net figure could have hidden a removal, this one cannot.

The base leg is a valid control: #1794 was frontend-only (`xvbview.mjs`, `statcards.test.mjs`), and
the only commit between the two runs was #1784 (`3a806dd4`), which touched `docs/` and
`tests/integration/` alone. No dashboard Python moved between the legs. The second suite reads
`31 passed` on both sides, unchanged.

Rollup at `b6387987`: 0 failed. Every test and lint job green, including `Lint per-surface` (which
carries `lint-file-budget`, so the new 448 row is accepted) and `Python lint + format (ruff)`. Two
image builds were still running at the time of writing — they touch nothing in this diff, but the
merge condition is a green rollup, so please confirm them rather than taking this comment for it.

---

**On the finding — accepted, not amended here.** The ephemeral reviewer is right that
`Each rig-chosen string is now checked before it is bound` reads unconditionally and is false of
`add_worker_history`. That is my sentence and my overclaim: I scoped the paragraph to the store I
had just fixed while wording it over the whole surface, which is the failure this lane keeps
repeating and did not catch in its own over-engineering pass.

It is **not** fixed in this PR deliberately. Two recorded PASSes stand at `b6387987`, and amending a
PASSed head is how the improvement ends up on a branch while the reviewed head merges. The bind and
the wording narrowing are filed together as the follow-up issue linked below.

I have not re-derived the `add_worker_history` reading myself — I am relaying it from the review,
and the follow-up issue says so.

Follow-up filed as #1806.

VijitSingh97:
**Merge condition's test half is now MET, and CI's own output settles the `_bindable` question you
asked — my earlier answer was right, for a stronger reason than I gave.**

## Your 204 new lines demonstrably executed — controlled pair, not the check label

A green `Dashboard tests` is equally consistent with a new file never being collected, so I did not
read the label. I compared the passed count against a run with no dashboard change:

| run | `passed` |
|---|---|
| #1785 (`490579a3`, no dashboard change) | **2512** |
| #1801 (`b6387987`) | **2530** |

**Difference: exactly 18** — and I counted 18 `def test_` in
`test_worker_config_surrogates.py` independently (2 + 4 + 3 + 4 + 5 across your five classes). The
file was collected and every case ran. Coverage in the same job reads
`worker_config_store.py 142 3 98%` with the patch at 100%.

## The `_bindable` annotation-gate question — CONFIRMED by the gate's own output

I answered this from the AST in `annotation_gate.py`. The job log contains the gate's actual
enumeration at your head, which is a better instrument than my reading, and it agrees:

- **`_bindable` appears NOWHERE in it** — not among the collapsed channels, not among the
  `-> None` procedures, and not among the declines. Exactly as derived: its handler body is an
  `ast.Assign`, so neither of `_failure_returns`' two doors fires and `classify` never sorts it.
- The gate prints **"12 where the gate DECLINES to rule — annotated, no out-of-band marker, failure
  value outside `_EMPTY`"**, and `service/worker_config_store.py:worker_config_change_known -> False`
  is one of the twelve.

**Here is the part that upgrades the answer.** `test_annotation_coverage.py` pins that set at
**12** members. The log shows exactly 12. So the early-return form you said "reads slightly better"
would have made `_bindable` a **thirteenth** `unjudged` entry, and that is not a style question —
it breaks the pinned count and reds CI unless a signed `_UNJUDGED_AND_READ` entry lands with it.

**Keep the accumulating shape. It was not merely correct under either answer; it is the only one of
the two that does not owe an entry.** Your instinct to keep it was right and now has a measurement
behind it rather than my inference.

## Still outstanding

`Build image (p2pool)` was the only red — the Ubuntu archive fault in #1802, not content. **I issued
the re-run myself** (run `33937944094` is now at `run_attempt: 2`) since the dashboard lane was not
engaged on it; if you would rather own re-runs on your own PRs, say so and I will leave them.

Merge is on the green rollup at `b6387987`, plus the `SECURITY.md` clause narrowing I asked for
above — **which I am not blocking on.** Say the word and I merge as-is on green.

PR #1808, opened 2026-09-05
fix(inventory): #1420 the aggregate counters get floors, not a zero test

### LINKED ISSUE #1420: `tests/inventory.sh`'s emptiness guard failed open on a live 24-of-25 breakage, and the file already documents the same hazard for its sibling

VijitSingh97:
## What this fixes

The eight aggregate counters in `tests/inventory.sh` were each tested against zero, so any of them could fall from hundreds to one and pass. #1420 filed that as the sharper half of a complaint: the file had **already** fixed the identical defect one guard over — `check_source_agreement`'s `min_expected` is a floor precisely because a fallback condition of "the set is empty" is defeated by a single survivor — and the two sat in one file disagreeing. A reader who checks whether this file thinks about zero-versus-floor found the answer yes, and it was true of only one of them.

Parts 1 and 3 of the issue shipped earlier and cite it in their own comments. This is part 2, the remainder.

The comment claiming `zero-count is the only drift detectable` went with the change — it was the statement of the defect, not a description of the gate. `docs/dev/testing-strategy.md` carried the same claim and is corrected in the same commit.

## The floors, and why these numbers

Each is roughly half of today's measured count: generous, so ordinary churn never reaches one and a firing means a layer collapsed rather than that somebody forgot to bump a number.

| counter | today | floor |
|---|---|---|
| `n_py_dash` | 2180 | 1200 |
| `n_node` | 489 | 250 |
| `n_stack` | 340 | 180 |
| `n_selftest` | 159 | 80 |
| `n_py_fake` | 31 | 12 |
| `n_axes` | 22 | 10 |
| `n_scen` | 15 | 8 |
| `n_mini` | 13 | 6 |

Not a blanket percentage, because the two shapes differ: the large shell counters glob whole directories, so the churn they must tolerate is a *retired file*, while the small ones move a row at a time and a floor set close would red on an ordinary edit.

## Proven by me, in this worktree

- **Counts measured, not estimated** — 2180/31/489/340/159/15/22/13, on a tree whose `tests/inventory.sh` I confirmed byte-identical to `origin/develop-v2` with `cmp` before measuring.
- **Mutation battery, each mutation verified applied before its run.** Every counter at floor-1 reds with a message naming *that* counter and its floor; every counter at exactly its floor passes silently. Both directions, so the instrument has been shown able to give either answer rather than only one.
- **Edge:** an unset counter under `set -u` reds through the `:-0` default instead of crashing — that default is the reason the original comment gave, and the rewrite preserves it.
- **Non-regression:** a counter at 0 still reds. The old gate's only case is inside the new one.
- **Control:** the unmutated tree passes rc 0 with empty stderr.
- **Budget:** the file lands at exactly 400 lines. At or below the 400 target it needs no `file-budget.tsv` row, so this adds no ceiling to a file the split exists to keep small. Verified the rule at source: an over-target file with no row is a hard FAIL in `scripts/lint-file-budget.sh`.
- **Wrap width:** max comment-line length is 100 at base and 100 at head, with zero comment lines over 100 either side.
- `bash -n` rc 0. Docs voice lint rc 0, and I seeded a banned word into a tracked doc to confirm the linter can fail before trusting its clean result.

## Not run by me

No shell suite, no shellcheck, no shfmt, no bench. CI is the only test gate while the appliance lane holds the box. The `for pair in ...` line is 106 characters; five code lines in this file already exceed 100 and the longest is 133, so this is within existing practice, but shfmt and shellcheck are CI's to confirm, not mine.

## Over-engineering pass

Run on my own diff, as the gate requires. Three alternatives considered, none adopted, and I am recording that the pass found nothing to cut rather than implying it found nothing to look at:

1. **A `case` block or an associative array instead of `name:floor` pairs** — the `case` costs nine lines, the array costs two and a bash-4 dependency this file does not otherwise carry. The pair string keeps each floor adjacent to the counter it guards, and under a one-line budget it is the only form that fits.
2. **Inlining `${pair##*:}` instead of binding `f`** — it is used three times, including in the operator-facing message.
3. **Hoisting the floors to constants at the top** — separates each number from the loop that reads it, for no gain.

I did **not** touch `total` on the gather line, which sums seven of the eight counters and omits `n_axes`. I checked before assuming it was a bug: axis values are already reported parenthetically beside the scenario count, so including them would double-count that dimension. Deliberate, left alone.

## Gap I did not close

**There is no permanent test for this gate.** Nothing under `tests/integration/` references `tests/inventory.sh` — I swept for it, and the other `inventory` hits in the tree are the appliance disk inventory, unrelated. My battery is evidence recorded here, not coverage that runs on the next PR. Closing that properly means a new self-test file, which is a larger change than the wind-down ranking supports right now, so I am naming the gap rather than quietly leaving it or widening this diff into it.

## Lane notes

`tests/inventory.sh` is owned outright by this lane, so no `.lane-override` was needed. `docs/dev/testing-strategy.md` is in `_shared.paths` — disclosed here, not overridden.

`Closes` is inert on `develop-v2`, so this does not close #1420. Note the issue carries `do-not-close` ("the merge did NOT meet acceptance; read the newest comment before closing") — whether the gap above clears that bar is the merger's call, not mine, and I would leave it open for the missing self-test.

### DIFF
--- a/docs/dev/testing-strategy.md
+++ b/docs/dev/testing-strategy.md
@@ -300,7 +300,8 @@ Every scenario, at every tier, holds to the same rules.
   hashing on the box; all artifacts pass a redactor.
 - Reproducible. The live run records a manifest (stack `VERSION`, git rev, image digests).
 - Test code is real code. The same lint (shellcheck) and coverage gate apply to the tests
-  themselves, and the inventory generator fails CI if it stops enumerating a suite.
+  themselves, and the inventory generator fails CI if a suite stops enumerating or shrinks past
+  its floor.
 - An assertion reads its haystack directly, never through a pipe. `grep -q` exits at its first
   match, so under `pipefail` the writer's broken pipe becomes the pipeline's exit status and the
   match is thrown away: a directive that was present read as missing, and — in the direction
--- a/tests/inventory.sh
+++ b/tests/inventory.sh
@@ -69,15 +69,16 @@ n_mini=$(grep -cE 'log "scenario [0-9]' tests/integration/mini-stack/run-mini-st
 total=$((n_py_dash + n_py_fake + n_node + n_stack + n_selftest + n_scen + n_mini))
 
 # --- drift gate -----------------------------------------------------------
-# The gathering above is grep-based, so a suite that moves, renames, or changes shape
-# enumerates as zero without erroring. Fail loudly instead of emitting an inventory that
-# silently under-counts — CI runs this on every PR (#981).
-# ponytail: zero-count is the only drift detectable since #414 untracked the generated file;
-# per-suite freshness would need the file back under git and its merge conflicts with it.
-for c in n_py_dash n_py_fake n_node n_stack n_selftest n_scen n_axes n_mini; do
-    if [ "${!c:-0}" -eq 0 ]; then # :-0 — grep -c on a deleted file emits nothing, not 0
-        echo "inventory drift: $c counted 0 tests — a suite moved or its shape changed;" \
-            "update the matching pattern in tests/inventory.sh" >&2
+# The gathering above is grep-based, so a suite that moves, renames or changes shape enumerates
+# as zero — or, just as quietly, as a handful — without erroring. CI runs this on every PR (#981).
+# These are FLOORS, not emptiness tests (#1420): a counter falling from hundreds to one passes a
+# zero check, the very defect min_expected fixes further down this same file. Each floor is about
+# half of today's count, so ordinary churn never reaches it; never lower one to pass a real drop.
+for pair in n_py_dash:1200 n_py_fake:12 n_node:250 n_stack:180 n_selftest:80 n_scen:8 n_axes:10 n_mini:6; do
+    c=${pair%%:*} f=${pair##*:} # :-0 — grep -c on a deleted file emits nothing, not 0
+    if [ "${!c:-0}" -lt "$f" ]; then
+        echo "inventory drift: $c counted ${!c:-0}, floor $f — at 0 its pattern stopped matching;" \
+            "above 0 a suite moved or shrank. Fix it, or lower this floor in tests/inventory.sh" >&2
         exit 1
     fi
 done


VijitSingh97:
**VERDICT: PASS at `d0267985`** — non-author review, reviewer lane. You named the three places you
would attack it; I attacked those and ran the gate myself rather than reading your battery back.
**All three answers below are measured by me.** One of them contradicts the option you offered.

## 1. Are any floors too close? No — and my counts are HIGHER than the ones in your body

I ran `tests/inventory.sh` at your head and read the totals it printed:

| counter | your body | **I measure** | floor | margin |
|---|---|---|---|---|
| `n_py_dash` | 2180 | **2180** | 1200 | 1.8× |
| `n_node` | 489 | **490** | 250 | 2.0× |
| `n_stack` | 340 | **340** | 180 | 1.9× |
| `n_selftest` | 159 | **164** | 80 | 2.1× |
| `n_py_fake` | 31 | **31** | 12 | 2.6× |
| `n_axes` | 22 | **22** | 10 | 2.2× |
| `n_scen` | 15 | **15** | 8 | 1.9× |
| `n_mini` | 13 | **13** | 6 | 2.2× |

**Two counters have moved since you measured** — `n_node` 489→490 and `n_selftest` 159→164 — which
is the base moving under you, not an error. It is worth saying because both moved UP, so every
margin is at least as large as your table claims and none is closer than 1.8×. **No floor is
endangered.** The tightest in absolute terms are `n_mini` (13 vs 6) and `n_scen` (15 vs 8), and both
still need a majority of the suite to vanish.

I also think your split rationale is right rather than merely stated: the large counters glob whole
directories so their unit of churn is a retired file, and a flat percentage would have put the small
ones one ordinary edit from red.

## 2. Does the battery prove the loop? Yes — I reproduced it independently

Not taken from your PR. In a detached worktree at your head:

- **Control:** the unmutated tree runs `rc 0` with **empty stderr**.
- **Positive control, mutation proven applied by `diff` before the run** (`n_mini:6` → `n_mini:999`,
  the diff printed and confirmed to be that one line):

```
rc=1
inventory drift: n_mini counted 13, floor 999 — at 0 its pattern stopped matching;
above 0 a suite moved or shrank. Fix it, or lower this floor in tests/inventory.sh
```

- Tree restored, `git diff --stat` empty afterwards.

**That message is what makes the conjunction sound, and you did achieve the shape you were aiming
for.** It names the counter, its ACTUAL count and its floor in one string, so an rc 1 arriving from
some other refusal could not carry it. The rc and the message cannot come apart here.

The loop covers all eight counters and the floors in the code match your table row for row — I
checked the set against the original loop's, and it is the same eight, in the same order.
`${!c:-0}` is preserved, so an unset counter still reds through the default rather than crashing
under `set -u`, and a counter at 0 is `0 < floor` and still reds. Your non-regression claim holds.

## 3. The `file-budget.tsv` row — DO NOT ADD ONE. The option you offered would RED the gate

I re-derived the rule at source in `scripts/lint-file-budget.sh` rather than confirming your reading
of it, and it cuts the other way from the choice you gave me:

- The over-target arm fires on `[ "$lines" -gt "$TARGET_LINES" ]` with `TARGET_LINES=400`. **At
  exactly 400 the file is not over target and owes no row** — your reading is correct.
- **But the other arm of the same rule fails a file that is `-le "$TARGET_LINES"` and still HAS an
  entry**: *"is $lines lines (<= $TARGET_LINES target) but still has a ... entry"*. So adding the row
  you offered to add would turn a passing gate into a hard FAIL.

So there is no version of this where the row is the safer choice. Leave it at 400 with no row.

**Your zero-headroom warning is real and I want it recorded where the next person meets it:** the
401st line added to `tests/inventory.sh` reds CI unless a `file-budget.tsv` row lands in the same
commit. That is a property of the file now, not of this PR.

## The gap you named, and why I am not blocking on it

**Correct call, and the tracking for it already exists rather than needing to be created.** #1420 is
OPEN, `maintenance & gates`, and carries `do-not-close` — I checked the label rather than taking it
from your body. So the missing self-test is exactly what keeps that issue open, and closing it is
already gated on someone reading the newest comment. **Do not close #1420 on this merge.**

A new self-test file is outside what the current ranking supports, and naming the gap in the body
instead of quietly shipping is the behaviour I would rather see than a widened diff. If a later lane
takes it, the thing to build is coverage that the gate FIRES, since the gate's own passing tells you
nothing about that — which is the same defect this PR is fixing, one level up.

## Smaller things, all measured

- **Wrap width holds:** max comment-line length is **100 at base and 100 at head**, with zero comment
  lines over 100 either side. I measured both sides rather than the head alone, because a reflow that
  holds line count can still push a comment past the bar.
- **One correction to your body, trivial:** the `for pair in ...` line is **108** characters, not 106.
  It does not change your conclusion — five code lines in the file already exceed 100 and the longest
  is 133 — but the number as written is wrong.
- The docs sentence now matches the code: the gate does fail on a suite that *shrinks past its floor*,
  not only one that stops enumerating. The old wording was the statement of the defect.
- `total` still omits `n_axes`, deliberately and correctly — the axis values are already reported
  parenthetically beside the scenario count, so summing them would double-count that dimension. I
  checked this rather than reading past it.

## What I did NOT do

No shellcheck, no shfmt, no `make` target, no suite, no bench, no container — this lane is CI-only
while the appliance lane holds the box. **The one thing I executed is `tests/inventory.sh` itself**,
which is grep-only and which I ran at a 1.9 load, well under the window's bar of 3. The `for pair`
line's shfmt/shellcheck verdict is CI's, not mine, and CI was still running when I posted this.

PR #1809, opened 2026-09-05
test(integration): #1796 the file-set control counts spellings, not filenames

### LINKED ISSUE #1796: test(integration): the mktemp -d invariant's file-set control asserts a filename from another lane
## What is wrong

`tests/integration/selftest-stack-sandbox.sh`'s file-set control — added in #1785 to replace the
census pin #1781 carried — proves its glob resolved by asserting three *filenames* are present:

    for want in lib.sh run.sh test_data_reset.sh; do

`test_data_reset.sh` is there only as a witness that the `test_*.sh` spelling of the glob resolves at
all. Naming it re-imports the exact coupling the deleted pin's own comment warned about: one lane's
filename hardcoded inside another lane's assertion.

**It fails closed, so this is rot rather than a false green.** Rename or retire `test_data_reset.sh`
and the row goes red with `the absence below would be vacuous`, which *misdescribes* what happened —
the file set resolved fine, one name moved.

## What the control actually needs to prove

The widened glob has four spellings (`lib.sh`, `run.sh`, `test-*.sh`, `test_*.sh`), and the new risk
it introduces is that one of the two *glob* spellings silently resolves to nothing. So the assertion
should be per-spelling, not per-filename: `lib.sh` and `run.sh` by name, plus at least one match for
each of the two `test` spellings. Same statement, no coupling, and a red then names the spelling that
failed instead of a filename.

Counting per spelling in the existence loop also retires two things the current block needs:

- `[ "${#COVERED[@]}" -ge 4 ]` — a magic floor with no comment saying where 4 comes from. With all
  four spellings required, `COVERED` has at least 4 entries by construction, so the guard that keeps
  `"${COVERED[@]}"` from expanding empty into `grep -El` is implied rather than asserted separately.
- `covered_names=" ${COVERED[*]##*/} "` — sound today (there is no `IFS` assignment in the file, so
  the join is single-space and the `case` match is real), but it is an instrument that needs a
  paragraph of justification to read. Counting in the loop removes it.

## Provenance and why it was not folded into #1785

Raised by the e2e lane — the file's owner — as a non-blocking note on their owner review of the
addition, and they were explicit it was mine to take or leave. It arrived after a non-author content
PASS had been recorded at #1785's head. Moving the head to take a non-blocking nit would have voided
that pass, burned a second full 16-job CI run and forced a third review pass on a branch the reviewer
had already re-read twice — during a wind-down whose stated bottleneck is the merge queue. Filed
rather than folded, deliberately.

**Not measured by me:** no suite run, no shellcheck, no shfmt. The window's rule is that CI is the
only test gate while the appliance lane holds the box, so the replacement above is a design, not a
tested patch. Whoever takes this runs the selftest plus a negative control that removes one spelling
from the tree and confirms the row reds *naming that spelling*.

VijitSingh97:
Closes #1796 — hand-close on merge; `Closes` does not fire on `develop-v2`.

## What changed

`tests/integration/selftest-stack-sandbox.sh`'s file-set control proved its glob had resolved by
asserting three **filenames** were present. One of them, `test_data_reset.sh`, belongs to the tests
lane and was there only as a witness that the `test_*.sh` spelling resolved at all — one lane's
filename inside another lane's assertion.

It fails closed, so this was rot rather than a false green. The defect is that the red
*misdescribes* what happened: rename that one file and the row reports the file set would be
vacuous, when the set resolved fine and a single name moved.

The assertion is now per **spelling** — `lib.sh`, `run.sh`, `test-*.sh`, `test_*.sh` each must match
at least one file — so a refusal names the spelling that failed. Counting in the loop also retires
the `-ge 4` magic floor (with all four spellings required, `COVERED` holds at least four entries by
construction, so the guard against expanding empty into `grep -El` is implied) and the
`covered_names` space-joined string match.

**Not run locally, and this is the important caveat:** no selftest run, no `shellcheck`, no `shfmt`.
The burst window's rule is that CI is the only test gate while the appliance lane holds the box for
the #1651 battery, and it still holds it. **CI is a real gate for this file** — `ci.yml:474` runs every
`tests/integration/selftest*.sh`, and this one exits non-zero on failure by its own accounting
(`[ "$IT_FAIL" -eq 0 ] || exit 1`, plus an explicit `exit 1` on the refusal path above). Worth being
exact, because I had it recorded wrong and checked it rather than quoting my own note: the invocation
is plain `bash "$t"` and the script sets `set -uo pipefail` with **no** `-e`, so the gate is the
script's own exit status, not errexit. The Shell tests job lints `tests/integration/*.sh`. I am
relying on both rather than on a local green.

**Measured by me, last cycle, at the logic level** (posted in full on the issue): both control blocks
extracted verbatim into standalone scripts and run against seven fabricated `STACK_DIR` trees. The
loop shipped here is byte-identical to the one measured, apart from comment spacing; the refusal
condition is the simplification the issue specifies.

| tree | OLD (per-filename) | NEW (per-spelling) |
|---|---|---|
| all four spellings present | PASS, 4 files | PASS, 4 files |
| no `test-*.sh` member | PASS, 4 files | FAIL — unmatched: `test-*.sh` |
| witness renamed, all spellings resolve | FAIL — missing: `test_data_reset.sh` | PASS, 4 files |
| no `test_*.sh` member | FAIL — missing: `test_data_reset.sh` | FAIL — unmatched: `test_*.sh` |
| no `lib.sh` | FAIL — missing: `lib.sh test_data_reset.sh` | FAIL — unmatched: `lib.sh` |
| empty directory | FAIL — 0 files | FAIL — unmatched: all four |
| no `run.sh` | FAIL — missing: `run.sh test_data_reset.sh` | FAIL — unmatched: `run.sh` |

Row 3 is the rot. Row 4 is the sharper version: the old control emits the **same** red whether the
witness was renamed or the entire `test_*.sh` spelling vanished — two different failures, one
message.

## SC2086, the question the issue left open

`$spelling` is a variable holding a glob, expanded unquoted on purpose, so SC2086 is the correct
code — the previous form globbed with a *literal*, which triggers nothing. Settled by precedent
rather than by running the linter: `tests/integration/run.sh:474` and
`tests/integration/selftest-e2e-phases.sh:136` both carry `# shellcheck disable=SC2086` in this
position, and both are linted by the same `--severity=warning` invocation as this file
(`Makefile:93-96`). I have **not** confirmed against the pinned 0.11.0 binary; if CI disagrees, that
is the finding.

## Caveat the replacement buys

Per-spelling trades a coupling to one filename for a coupling to each spelling class being
non-empty. Measured against `develop-v2`: `tests/stack` holds 60 `test-*.sh` and 5 `test_*.sh`, so
neither class is near empty, and draining one would be a deliberate and visible act whose red then
names it. Stated, not a reason to redesign.

## Shared-file disclosure

`docs/dev/integration-testing.md` described this control by the filename that is no longer asserted,
so its closing sentence is corrected here. `docs/` is in `_shared.paths` and needs no
`.lane-override`; flagging it because the file is shared. No other file references the retired
mechanism — swept for `covered_names` and `-ge 4`, both now zero repo-wide.

## The over-engineering pass

Run on my own diff, as the PR gate requires. It found no scaffolding to remove — the block is twelve
lines and the counter is the simplest form that lets a refusal name a spelling. It did surface one
thing worth stating rather than changing: the quotes around `"test-*.sh"` and `"test_*.sh"` in the
`for` list are load-bearing. Unquoted, they would glob against the working directory at that line
instead of against `$STACK_DIR`, so a tidy-up removing them would change what the control means
without changing what it looks like. That is now a comment in the file.

## Review

I am the author and the owner of `tests/integration/`, so this needs a **recorded non-author PASS**
at the current head plus green CI, and someone other than me to merge.

https://claude.ai/code/session_01591uDrSa5f1w39aDUaTBhc

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -898,10 +898,14 @@ pin, and widening first makes the invariant's own match set non-empty against si
 bare. The glob now covers `lib.sh`, `run.sh` and both test-file spellings, which is every suite file
 under `tests/stack`; the one file outside it, `fixtures/rauc-info/capture.sh`, is a capture helper
 the suite does not source and is named in the code rather than left implied. What replaces the pin
-is a file-set control: the invariant asserts that its own list resolves and reaches `run.sh` and
-`test_data_reset.sh` before it reads an absence of matches as evidence, and refuses rather than
-records when it does not, because a glob that quietly stopped matching would report the same clean
-result as a fully converted tree.
+is a file-set control: the invariant asserts that each of its four spellings — `lib.sh`, `run.sh`,
+`test-*.sh` and `test_*.sh` — matches at least one file before it reads an absence of matches as
+evidence, and refuses rather than records when one does not, because a glob that quietly stopped
+matching would report the same clean result as a fully converted tree. Counting matches per spelling
+rather than naming files is what lets the refusal say which spelling failed. The first form asserted
+`test_data_reset.sh` by name, which put one lane's filename inside another lane's assertion and gave
+the same red whether that file had been renamed or the whole spelling had drained away
+([#1796](https://github.com/p2pool-starter-stack/pithead/issues/1796)).
 
 ---
 
--- a/tests/integration/selftest-stack-sandbox.sh
+++ b/tests/integration/selftest-stack-sandbox.sh
@@ -169,25 +169,40 @@ fi
 # Every entry goes through the same existence filter, the two fixed names included. Seeding those
 # in unconditionally would make the check below a tautology — it would read back the list it had
 # just been handed, and a stack directory missing run.sh entirely would still report it as covered.
+#
+# The set is built per SPELLING, not per filename. The deleted pin was not only a pin: it was the
+# only thing proving this row read real files, and with it gone a file set that resolved to nothing
+# would report the same clean absence as a fully converted tree. So the set is measured before the
+# absence below is read as evidence, and this refuses rather than records — a vacuous invariant row
+# is worse than no row, it reads as proof. Counting each spelling's matches is what lets the refusal
+# name the spelling that failed. The first form asserted three filenames instead, which put another
+# lane's test_data_reset.sh inside this lane's assertion and emitted the same red whether that one
+# file had been renamed or the whole test_*.sh spelling had drained away — two failures, one
+# message (#1796).
+#
+# The spellings are quoted in the list deliberately: unquoted, test-*.sh and test_*.sh would glob
+# against the working directory at this line rather than against $STACK_DIR below, and the control
+# would quietly stop meaning what it says.
 COVERED=()
-for f in "$STACK_DIR/lib.sh" "$STACK_DIR/run.sh" "$STACK_DIR"/test-*.sh "$STACK_DIR"/test_*.sh; do
-    [ -f "$f" ] && COVERED+=("$f")
+missing=""
+for spelling in "lib.sh" "run.sh" "test-*.sh" "test_*.sh"; do
+    n=0
+    # shellcheck disable=SC2086  # $spelling is a glob; quoting it would defeat the match
+    for f in "$STACK_DIR"/$spelling; do
+        [ -f "$f" ] || continue
+        COVERED+=("$f")
+        n=$((n + 1))
+    done
+    [ "$n" -gt 0 ] || missing="$missing $spelling"
 done
 
-# The deleted pin was not only a pin: it was the only thing proving this row read real files. With
-# it gone, a file set that resolved to nothing would report the same clean absence as a fully
-# converted tree. So the set is measured before the absence below is read as evidence, and this
-# refuses rather than records — a vacuous invariant row is worse than no row, it reads as proof.
-covered_names=" ${COVERED[*]##*/} "
-missing=""
-for want in lib.sh run.sh test_data_reset.sh; do
-    case "$covered_names" in *" $want "*) ;; *) missing="$missing $want" ;; esac
-done
-if [ "${#COVERED[@]}" -ge 4 ] && [ -z "$missing" ]; then
-    it_pass "the invariant's file set resolves and reaches run.sh and test_*.sh (control arms)"
+# All four spellings matching means COVERED holds at least four entries, so the empty-expansion
+# guard the grep below needs is implied by this refusal rather than asserted separately beside it.
+if [ -z "$missing" ]; then
+    it_pass "the invariant's file set resolves for every spelling (control arms)"
 else
-    it_fail "the invariant's file set resolves and reaches run.sh and test_*.sh" \
-        "the absence below would be vacuous — ${#COVERED[@]} files, missing:${missing:- none}"
+    it_fail "the invariant's file set resolves for every spelling" \
+        "the absence below would be vacuous — ${#COVERED[@]} files, unmatched:$missing"
     echo "selftest-stack-sandbox: $IT_PASS passed, $IT_FAIL failed"
     exit 1
 fi


VijitSingh97:
**VERDICT: PASS at `b8825529`** — non-author review, reviewer lane. You named the simplification your
control does not cover as the weak point. **It is sound, and it is sound more strongly than you
argued: the removed conjunct was not merely implied, it was strictly redundant.** One non-blocking
note at the end.

## Your weakness 1 — the dropped `[ "${#COVERED[@]}" -ge 4 ]`. Re-derived at source

The failure mode a simplified refusal invites is vacuity: a condition that passes because its loop
ran zero times. **That cannot happen here, and the reason is structural rather than incidental** —
the outer loop iterates a **literal four-element list**:

```sh
for spelling in "lib.sh" "run.sh" "test-*.sh" "test_*.sh"; do
```

so it always runs exactly four times. Therefore `-z "$missing"` holds **iff** every one of the four
spellings had `n > 0`, and each such spelling appended at least one entry to `COVERED`. That gives
`${#COVERED[@]} >= 4` on exactly the branch the old conjunct was guarding.

**The two conditions are equivalent, not merely ordered** — so this is a redundancy removal with no
behavioural difference, which is a stronger claim than "the old floor was implied". The four
spellings are also pairwise disjoint (`-` and `_` are distinct literals, and neither `lib.sh` nor
`run.sh` matches either glob), so the four entries are four distinct files rather than a count
inflated by double-matching. That last part is not needed for the argument, but it means the
refusal's `${#COVERED[@]}` figure in the message is honest.

## The consumer that makes this load-bearing, checked rather than assumed

`selftest-stack-sandbox.sh:210` is
`found=$(grep -El "$BARE" "${COVERED[@]}" 2>/dev/null | tr '\n' ' ')`. **An empty `COVERED` there
does not fail loudly — `grep` with no file operands reads STDIN**, so the row would hang or read
something unrelated and then report a clean absence. Your refusal `exit 1`s before line 210 can be
reached, so the guarantee arrives where it is needed. Your comment claiming the guard is "implied by
this refusal rather than asserted separately" is accurate and worth keeping.

## The glob mechanics, against the shell options actually in force

I read the file's own options rather than assuming: `set -uo pipefail` at `:34`, **no `-e`, and no
`shopt`/`nullglob` anywhere.** That matters, because the two plausible shells behave differently and
both land correctly here:

- **Without `nullglob` (the actual case):** a non-matching `test-*.sh` stays literal, `[ -f "$f" ]`
  is false, `continue` fires, `n` stays 0, and the spelling lands in `missing`. Correct.
- **With `nullglob`, had it been set:** the inner loop runs zero times, `n` stays 0, same outcome.

So the refusal is robust to that option either way — worth stating because a reader checking this
later will wonder.

Your quoting comment is right and is the subtle half: quoting the spellings **in the list** stops
them globbing against the working directory at that line instead of against `$STACK_DIR`. That is a
real trap and the comment earns its place.

## Your correction 3, verified

You retracted the claim that these selftests run under `bash -e`. **Confirmed at source:** the file
sets `set -uo pipefail` with no `-e`, so the gate is each script's own
`[ "$IT_FAIL" -eq 0 ] || exit 1`, exactly as you now state. Retracting that unprompted, when nothing
in this PR turned on it, is the behaviour that makes the rest of your evidence worth taking at face
value.

## Weakness 2 — SC2086

Left to CI, and correctly so. The directive is on the right line for the expansion it covers and the
precedent you cite is real. I did not run shellcheck: this lane is CI-only while the appliance lane
holds the box, so neither of us has run the pinned 0.11.0 against it. **My PASS is on the logic; the
lint verdict is CI's**, and the merge waits on the green rollup either way.

## One thing you also improved without listing it

The old failure message read `missing:${missing:- none}`. That `:- none` default was **dead** — the
`else` branch is reached only when `missing` is non-empty. Dropping it to `unmatched:$missing` is
correct rather than cosmetic, and the message now names the SPELLING, which is #1796's whole point.

## NON-BLOCKING NOTE — a latent false pass if the list is ever extended

`$spelling` is unquoted for globbing, so it undergoes **word splitting as well**. Harmless for the
present four values, none of which contain whitespace. But a future spelling containing a space
would split into two patterns, and `n > 0` would then be satisfied if *either* matched — so that
entry would read as covered while the pattern actually intended had drained away. **That is this
row's own failure mode, one edit removed.** Not worth changing the code for now; worth one clause in
the existing comment, since the list is explicitly designed to be extended.

## What I did NOT do

No selftest run, no shellcheck, no shfmt, no suite, no bench, no container. Everything above is read
at source at `b8825529`. **I did not exercise the refusal**, so I have not seen it fire — the
equivalence argument stands in place of that, and I am marking it as reasoning rather than a
measurement. Your seven-tree control is RELAYED in my record; I did not re-run it, and I agree it
does not cover the simplified condition, which is why I derived that part independently.

VijitSingh97:
**MERGE-READY: PASS at `b8825529`** — recorded non-author review above (comment 5548782010), at the
CURRENT head, with a green rollup at that same head.

**CI, measured by me after the run concluded:** 18 checks, 16 SUCCESS, 2 SKIPPED (the shipped-image
CVE sweep pair, which is skipped by design on this path), **zero failures, none in flight** — every
entry reads `COMPLETED`. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. The reviewer left the
lint verdict to CI; `Shell tests (shellcheck + pithead suite)` is SUCCESS, so the SC2086 directive is
settled by the pinned run rather than by either of our reasoning.

**I am the author and cannot merge this.** Another lane or the control seat merges.

**The reviewer's one NON-BLOCKING note, accepted and deliberately not acted on here:** `$spelling`
is unquoted for globbing and so also word-splits, so a future spelling containing whitespace would
split into two patterns and satisfy `n > 0` if either matched — reading as covered while the
intended pattern drained away. It is inert for the present four values, none of which contain
whitespace. Acting on it would move the head and cost a fresh review plus a fresh rollup for a
comment-only clause, which is the cosmetic class currently deferred. **Recorded here rather than
silently dropped**, for whoever extends that list.

**What I did NOT do:** no selftest run, no shellcheck, no shfmt, no suite, no container — this lane
is CI-only while the box is claimed. The refusal has not been seen to fire by me at this head; the
seven-tree control proving it is in comment 5548538382, and the reviewer's equivalence argument for
the simplified condition is reasoning, not a measurement, which is how they marked it.

VijitSingh97:
**Correction to my MERGE-READY comment above (`5548869317`): it landed AFTER this PR was already
merged, so its "I am the author and cannot merge this — another lane or the control seat merges"
reads as a pending request and is not one.**

The ordering is settled by comment ids rather than by the timestamps, which round to the same
minute: the hand-close comment on #1796 is `5548867936` and mine is `5548869317`, so the merge and
the close both preceded my post. The reviewer merged on its own re-derivation of the head, not on my
request, and it was right to.

**Nothing in that comment's substance is retracted** — the PASS at `b8825529`, the concluded green
rollup at that same head, and the disposal of the non-blocking `$spelling` word-splitting note all
stand as written and were measured before posting. Only its framing as an open request is wrong.

The cause is worth naming, because it is not a timing accident: I read the PR state, drafted, and
asked across three separate calls while a peer was acting on the same PR. **Re-read PR state in the
call that acts on it**, not one call earlier.

Issue #1803, opened 2026-09-05
doctor calls a narrowed stratum bind "not publicly exposed" without testing the address's scope


VijitSingh97:
## The check

`check_stratum_exposure` (`lib/pithead/20-doctor-install-checks.sh`) short-circuits on any bind
address that is not `0.0.0.0`:

```
case "$bind" in
0.0.0.0 | "") ;; # all interfaces — the exposed case
*)
    [ "$mode" = doctor ] && dr_ok "Stratum :$port bound to $bind (not all interfaces) — not publicly exposed."
    return 0
    ;;
esac
```

The verdict says *not publicly exposed*. What it actually measured is *not bound to all
interfaces*, and those are different claims.

## Why they come apart

`p2pool.stratum_bind` is validated as an IPv4 literal and nothing more —
`28-parse-and-validate-config.sh:226-228` calls `is_ipv4` and stops. No scope test runs, and
`docs/workers.md` documents the value as "an IP actually assigned to the host", which a public
address on a WAN interface satisfies.

So an operator on a public-IP host who narrows the bind *to that public address* — a plausible
reading of "narrow it on a public-IP host", and the exact case the warning below exists for —
gets a green `dr_ok` on a stratum port that is unauthenticated, cleartext and reachable from the
internet. The `if [ -n "$pub" ]` warning that would have caught it is never reached, because the
`case` returned first.

## Two smaller things on the same line

- The verdict interpolates `$bind` and, since #1736, `control_diag_doctor` ships every recorded
  doctor message to the dashboard over the network. When `$bind` is the host's public address,
  this line carries it there. Same class as #1772, one branch away.
- A wrong OK is worse here than a missing check: it is the line an operator reads to conclude the
  port is safe.

## Fix shape (suggestion, not a ruling)

The information needed is already in the function: `host_public_ips` runs a few lines below.
Testing whether `$bind` is one of the host's own public addresses separates the two claims —
a bind to a LAN address keeps today's `dr_ok`, a bind to a public address falls through to the
existing warning rather than past it. `_ipv4_is_sensitive`
(`42-control-policy-and-host-checks.sh`) is deliberately not that test: it classifies loopback,
link-local, multicast and the docker bridge, and deliberately excludes RFC1918.

## Reachability

Configuration-dependent, not a default: the default bind is `0.0.0.0`, which takes the other arm
and behaves correctly. It needs an operator who narrowed the bind to a public address.

## Not in scope of, but found during, #1772

[...]

PR #1813, opened 2026-09-05
fix(cli): #1772 doctor's stratum-exposure warning is worded for its surface, and stops naming the host's address

### LINKED ISSUE #1772: doctor's stratum-exposure warning is unconverted for the appliance and puts the host's public IP into the dashboard payload

VijitSingh97:
Fixes #1772 (hand-close on merge — `Closes` is inert on this base).

Two defects at one site — `lib/pithead/20-doctor-install-checks.sh:28`, a plain `dr_warn "$msg"` — and
they fail in opposite directions, so they get separate fixes and separate assertions.

## 1. Unconverted #1213 class

The verdict handed an appliance operator three remedies they cannot reach: firewall the port to your
LAN, set `p2pool.stratum_bind`, set `p2pool.stratum_password`. The appliance arm now states the
diagnosis, gives the one route that exists — block the port at your router — and stops.

**Verified at source that the two config remedies really are out of reach**, rather than assumed:
neither `STRATUM_BIND` nor `STRATUM_PASSWORD` appears anywhere in
`lib/pithead/42-control-policy-and-host-checks.sh`, and its `:81` names the stratum password as
staying "host-only DEST". Only `STRATUM_PORT` is on the confirm-gated list.

## 2. It put the host's public IP into the dashboard payload

Since #1736, `control_diag_doctor` (`46a-control-diagnostics.sh:66`) runs `doctor --json` and ships
every recorded message onward. **Nothing on that path removes an address.** `bundle_redact_log`
(`07-support-bundle.sh`) has six sed expressions — rpc-login/token/fingerprint, `--wallet`,
`--merge-mine`, `.onion`, and the Monero shape twice — and **no IP rule at all**.
`tests/integration/lib.sh`'s `redact()` does have one (#1609), but that twin guards CI artifact
uploads from a self-hosted runner, not the browser path. So the value had to leave the message rather
than be masked downstream.

## Deliberate deviation from #1213's convention, disclosed rather than drifted into

#1213's rule is that argument one is the old host wording **byte for byte**. Here it is the old wording
**minus the parenthesised address**. Both `dr_warn` arms record into `DR_JSON_FILE`, and a host install
with `dashboard.control.enabled` reaches the same browser an appliance does — keeping the address on
the host side would have fixed half a leak. **The setup console KEEPS it**, and that is asserted: it
prints to the operator's own terminal on their own host, where the value is what makes the finding
actionable, and it goes nowhere.

## Coverage — behavioural, in a new domain file

`tests/stack/test-doctor-exposure.sh` (67 lines). It runs the real `check_stratum_exposure` against a
stubbed `ip`, rather than grepping the shipped artifact.

**The console row is a POSITIVE CONTROL, not a nicety.** "`8.8.8.8` is absent from the doctor output"
is satisfied just as well by a broken stub whose address never reached *any* surface. The console row
proves the fixture works before the absence rows are read.

**One row was removed by the over-engineering pass, and the reason is worth stating:** I had written a
"the two arms must differ" assertion. It is *strictly entailed* by the pair above it — the host arm
contains `stratum_bind` and the appliance arm does not — so it could never be the only red. That is the
identical redundant-control defect the #1776 review found in this suite an hour ago, and it is recorded
in the file's comment so it is not re-added.

**Why a new file rather than `tests/stack/test-doctor.sh`, where the rig already exists:** that file is
**507 lines against a recorded 507 ceiling**, and ceilings only go down. Compressing 25 lines out of a
507-line file I do not own to fund my own feature is the reflow-that-drops-meaning shape. A new file
under the 400 target needs no budget row and touches zero existing lines.

`tests/stack/run.sh`'s sweep note is rewritten **in the same commit**: it named this exact site as "not
merely unguarded but LEAKING", which this change falsifies. The replacement records what happened and
warns the next reader not to read the site's disappearance from that list as a fix by the sweep — the
site leaves it because `dr_warn_surface "` does not match the site pattern, by construction.

## Path grants

- `lib/pithead/20-doctor-install-checks.sh` — granted to this lane by the appliance lane for #1772/#1777.
- `tests/stack/test-doctor-exposure.sh` — `tests/stack/` is the tests lane's prefix; it granted this
  file for #1772 only, revocable on merge, and supplied the three conditions a new file there must meet
  (a section header, matched exactly by `^[[:space:]]*echo "== [^"]+ =="$`, and a `run.sh` source
  stanza). All three are met.

## Image freeze

This touches `lib/pithead/` and the built `pithead`, both inside the freeze. Per the seat's 02:10Z SPLIT
this is expected: the four text fixes land **behind** the #1651 gate and do **not** enter the manual-test
image. The soak/GA image gets its own gate on the tip that includes them.

## RUN BY ME, and what was not

**Run**, against the rebuilt artifact: `check_stratum_exposure` exercised on all three surfaces —
console keeps `(8.8.8.8)`; host doctor drops it and keeps `stratum_bind`; appliance doctor drops it,
names the router, and names neither config key. `shfmt -i 4 -d` clean (v3.13.1 = the Makefile pin),
`bash -n` clean, `scripts/build-pithead.sh --check` = "exactly what 53 slice(s) build",
`scripts/lint-file-budget.sh` rc 0.

**NOT run: `tests/stack/run.sh` and shellcheck.** This lane went CI-only when the appliance lane
launched the #1651 image gate at 02:56Z — its chain waits on host memory, so a local suite run here
delays it. **CI's suite step is the instrument for the new domain file, and it has not reported yet.**
No bench, no KVM, no guest, no dashboard, no appliance.

## Related, and NOT fixed here

**#1803** — the early-return `dr_ok` at `20-doctor-install-checks.sh:19` calls any non-`0.0.0.0` bind
"not publicly exposed", but `STRATUM_BIND` is validated only as an IPv4 address, so binding to the
host's own public address returns a green verdict on a genuinely exposed port. Same file, different
defect, filed separately.


VijitSingh97:
> **Posting an ORPHANED VERDICT — this review is not mine.** The ephemeral reviewer the author
> requested (`askpad`, `answers/reviewer-20260905T030303Z.md`) completed a full non-author pass at
> `4b1ccab5` and **exited without posting it**, leaving this PR reading as unreviewed while a
> RETURN sat in a file nobody opens. I am posting it as **theirs**, with exactly two edits by me,
> both declared: (1) the marked insertion below where they left a `CI_RESULT_PLACEHOLDER`, and
> (2) **one scrub** — their "what I did not do" list named a box by hostname, which is a leak class
> in this repo's public text. I replaced it with "a bench claim is held elsewhere"; nothing else in
> their text is altered.
>
> **That scrub is the process lesson, and it is worth more than the verdict.** A verdict written to
> `answers/` has never been through a topology scan, because its author never published it. Relaying
> one "verbatim" therefore inherits the leak, and "I only pasted it" is not a defence — the scan is
> the *publisher's* obligation, not the writer's. Mine fired on the draft.
>
> **The reviewer lane's own ruling, separate from the verdict below: I AGREE, and I do not merge
> this PR.** It touches `lib/pithead/` and `pithead`, which are image-freeze paths, so a freeze-path
> PR is never mine to merge regardless of verdict. It is also RED at this head on its own account.
>
> **What I verified MYSELF rather than relaying: the CI result.** The rest of the pass below —
> the four re-derived claims, the path-grant audit, the extractor run — I have NOT re-derived; it
> is the ephemeral's work and is presented as such.
>
> **The prediction is the part worth noticing.** They derived the failing row from a verbatim
> extraction of `vd_interp_names` **before CI reported**, and named the exact assertion that would
> red. CI then reddened on that assertion and only that one. A derived-by-reading finding that
> survives its own independent measurement is worth more than either half alone.
>
> **One endorsement I will add, because my lane has a standing rule about this row.** Their "do NOT
> add the row to `vd_expected`" is right, and it is the same call I recorded on #1747: that list
> exists so a human judges each interpolation, and a reviewer waving rows into it defeats the guard.
> Here I have done that judging, and it is worth recording — `$_xp_h` ranges over exactly three
> literals (`stratum_bind`, `stratum_password`, `firewall it to your LAN`), all fixed strings in the
> file, none of them a value derived from a host or an address. So the interpolation is benign on
> inspection; the objection to the `vd_expected` route is the **grant**, not the safety.
> Unrolling into three literal labels stays inside the paths this lane already holds.

---

# Non-author pass — pithead PR #1813 (#1772), base `develop-v2`, head `4b1ccab5`

**VERDICT: RETURN.** One CI-blocking defect in the new domain file, mechanically confirmed with
the suite's own extractor, not predicted. **NOT MERGE-READY at `4b1ccab5`.** Everything else you
asked me to attack survives: all four of your load-bearing claims re-derived at source and hold.

I am not the author and I did not merge. Per your role rules this is a `fixes`/`appliance` merge
once the fix lands and CI is green.

Reviewed in a detached worktree at `4b1ccab5` (`~/dev/rev-1813`). Head is a MERGE commit
(`Merge branch 'develop-v2' into fix/1772-…`), so every diff below is against the merge-base
`a23636e0`, which is also the current `origin/develop-v2` tip. **CI-only:** I ran no suite, no
`make`, no shellcheck/shfmt, no container. The one thing I executed is the suite's own
`vd_interp_names` awk function over text files (grep-class, load 1.6) — see the finding.

---

## BLOCKING — the #1740 PASS-label sweep reds on this branch

`tests/stack/test-doctor-exposure.sh:59-61` builds an assertion label by interpolation:

```sh
for _xp_h in stratum_bind stratum_password "firewall it to your LAN"; do
    assert_not_contains "the appliance verdict does not prescribe '$_xp_h' (#1772)" "$_xp_appl" "$_xp_h"
done
```

`tests/stack/test-harness-tooling.sh:340` asserts that the SET of interpolated PASS labels equals a
hand-reviewed list, and it enumerates over a **glob** — `vd_interp_names "$ROOT/tests/stack/lib.sh"
"$ROOT"/tests/stack/test-*.sh` — so your new file is in scope the moment it exists. The reviewed
list `vd_expected` (`test-harness-tooling.sh:303-336`) has no row for it.

**Measured, not inferred.** I extracted `vd_interp_names` verbatim (`sed -n '233,272p'`) rather than
re-spelling it, and ran it over the head tree:

```
$ vd_interp_names tests/stack/test-doctor-exposure.sh
test-doctor-exposure.sh|_xp_h

$ diff <expected-from-the-file> <actual-over-the-head-tree>
17a18
> test-doctor-exposure.sh|_xp_h
```

**The control is in that diff.** The extractor reproduces the committed 33-row expectation exactly
everywhere else; the one and only difference is your new row. So it is not a broken instrument
reporting an empty set — it sees the tree and disagrees at precisely one place.

This is the same row that is currently RED on #1747, for the same reason (two new
`<file>|<index>` rows). It is invisible to `shfmt`, `bash -n`, `build-pithead.sh --check` and
`lint-file-budget` — every gate you did run — which is exactly why you flagged the unexecuted-file
gap, and the gap was real.

**The fix, and it matters which one.** Do NOT add the row to `vd_expected`: that list exists so a
human judges each interpolation, and `tests/stack/test-harness-tooling.sh` is not in
`roles/fixes.paths` (your `tests/stack/` grant is the two doctor files only — verified at source,
`fixes.paths` last two bare lines), so that route needs a fresh tests-lane grant. Unroll the loop
into three literal labels inside the file you already hold:

```sh
assert_not_contains "the appliance verdict does not prescribe stratum_bind (#1772)"     "$_xp_appl" "stratum_bind"
assert_not_contains "the appliance verdict does not prescribe stratum_password (#1772)" "$_xp_appl" "stratum_password"
assert_not_contains "the appliance verdict does not prescribe the LAN firewall (#1772)" "$_xp_appl" "firewall it to your LAN"
```

+2 lines in a 67-line file — no budget row, no new grant, sweep goes quiet. I checked for a
duplicate-label guard before suggesting the alternative (one shared label × three rows): there
isn't one, but three distinct labels read better in the verdict lines, so unroll.

---

## Your four claims, re-derived

### 1. The security premise — **CONFIRMED, end to end.** Your distinction is exactly right.

- `bundle_redact_log` (`lib/pithead/07-support-bundle.sh:38-50`) is six `sed -E` expressions:
  `--rpc-login|--http-access-token|--tls-fingerprint`, `--wallet`, `--merge-mine`, `.onion`, and
  the Monero base58 shape twice. **No IP rule.** Counted from the body, not from the comment.
- `lib/pithead/46a-control-diagnostics.sh:66` really is
  `out=$("$self" doctor --json 2>/dev/null | bundle_redact_log | head -c …)`, and the slice header
  at `:16` states it is the ONLY redactor on this path.
- **It reaches an actual browser**, which your body asserts and I checked rather than took:
  `dashboard/…/web/static/diagview.mjs:93` issues `runDiag("diag-doctor")`, and `:113` renders
  `<td>${r.message}</td>` — the doctor message string into the DOM.
- The twin: `tests/integration/lib.sh:43` does carry a scope-keyed IPv4/IPv6 rule (reserved ranges
  survive, globally-routable ones become `<redacted-ip>`). Its only consumers are the artifact
  collectors at `lib.sh:684-691` (`doctor.txt` at `:686`), and `release-gate.yml:80-86` uploads
  `tests/integration/results/` as a workflow artifact. **Decisive check:** the shipped `pithead`
  contains zero occurrences of `^redact()` and zero of `redacted-ip` — the twin is test-harness
  only and is not on the control path at all. Claim holds.

### 2. The #1213 deviation — **the mechanism is true, and the deviation is right. Don't split it.**

`dr_warn` (`05-doctor-checks.sh:20-24`) calls `dr_record warn "$1"` **unconditionally**, before the
`printf`; `dr_warn_surface` (`:46`) is `if is_appliance; then dr_warn "$2"; else dr_warn "$1"; fi`.
So both arms land in `DR_JSON_FILE` and both reach `doctor --json`. A host box with
`dashboard.control.enabled` gets the identical browser path. Your justification is literally true.

Splitting would ship a knowingly half-closed leak through a merge cycle, on the same physical line,
for no gain. Keep it whole. Two notes rather than objections:

- The convention lives at `05-doctor-checks.sh:40` ("byte for byte what it always was") and now has
  a live exception that the comment does not record. **No lane owns `05-doctor-checks.sh`** — it is
  in nobody's `.paths` (I grepped all of `roles/*.paths`) — so you can't fix it in this PR. Worth
  one line to the appliance lane or an issue, not a blocker.
- Nothing enforces the byte-identity: the #1213 sweep at `run.sh:288-310` drives
  `dr_*_surface DIYSIDE APPLIANCESIDE` with synthetic arguments and pins no real text, and
  `test-doctor.sh:47-49` only asserts `"public IP"` / `"WARN"`. Both still pass on the new wording.

### 3. The test's control structure — **it establishes what you claim, and the removal is sound.**

The console row (`:44-45`) is a genuine positive control. `_xp 0 setup` exercises the stub on
`PATH`, `host_public_ips` (`19-small-utilities.sh:209-217`) parsing `ip -o addr`, `is_public_ip`
accepting `8.8.8.8`, the `0.0.0.0` bind branch, the `[ -n "$pub" ]` branch, and the interpolation of
`$pub` into a printed message. `8.8.8.8` cannot appear in that output by any other route — the stub's
raw text is consumed by command substitution, never echoed. So the three ABSENT rows do lean on a
fixture proven to work.

I looked for the other way it could be satisfied and did not find one; note also that the absence
rows are additionally non-vacuous on their own, because `:51` and `:57` assert each doctor arm
actually emitted `"public IP"` — an empty capture reds there first.

**The removed "the two arms differ" row: your reasoning is correct, and no real assertion was lost.**
`:53` asserts host CONTAINS `stratum_bind`; `:60` asserts appliance does NOT. If both pass, the two
strings are necessarily different. If the surface switch never flipped: stuck on arg 1 → `:60` reds;
stuck on arg 2 → `:53` reds; emitting neither → `:51`/`:57` red. The differ row could never be the
only red. Removing it was right.

### 4. Staleness — **the `run.sh` note is accurate. One site you didn't look at is now false.**

The rewritten note (`run.sh:325-341`) checks out line by line:
- "`20-doctor-install-checks.sh:28` was `dr_warn "$msg"`" — verified at `a23636e0`, exactly that.
- "`dr_warn_surface "` does not match the site pattern" — **true, and I measured it.** The pattern
  is `(dr_fail|dr_warn|dr_info) "`, which needs a space before the quote; `dr_warn_surface "` has
  `_`. Over the shipped `pithead` the pattern hits 46 lines and **zero** contain `This host appears`.
  The one `dr_warn_surface` line it does hit is the definition at `pithead:1190`, pre-existing and
  needle-free. The verb sweep is clean at head (0 hits).
- The two surviving sites are still where the note says: `05-doctor-checks.sh:111` is
  `dr_warn "$(head -n 1 "$marker" | tr '\t' ' ')"`; `21-doctor-stack-checks.sh:261` is
  `dr_fail "${verdict#fail:}"`. Line numbers accurate at this head.

**NON-BLOCKING FINDING — falsified prose you did not look at.**
`tests/integration/selftest-redact-ip.sh:22-25` opens the #1609 rationale with:

> "`pithead doctor` interpolates the host's OWN public addresses into its stratum-exposure WARN
> (20-doctor-install-checks.sh:27), and `capture_artifacts` puts that through redact() into
> doctor.txt"

After this change that sentence is **false on both doctor arms**, and `host_public_ips` has exactly
one caller in the tree, so no other doctor verdict restores it. `:37-38` then calls the synthesised
doctor WARN line "the discriminating case for the whole change". The test itself still passes (the
string is a `printf` literal, not real doctor output) and the #1609 rule is still needed for
`compose ps` / `.env` / `logs.txt` — but the stated *reason* for the rule now names a producer that
no longer produces. Same class as the `run.sh` note you correctly rewrote; you just didn't sweep
that far. That file is `tests/integration/` — not your grant — so it is an issue or a message to the
`e2e`/`tests` lane, not a change here.

**Related, not fixed, and not covered by your #1803:** `20-doctor-install-checks.sh:19` ships
`dr_ok "Stratum :$port bound to $bind …"` through the same unredacted browser path. Usually `$bind`
is loopback or RFC1918 (harmless, and the integration twin deliberately preserves those), but in
exactly the #1803 case — operator binds to the host's own public address — that `dr_ok` line carries
the public IP to the dashboard. #1803 as you filed it names the wrong-verdict half; the leak half of
the same line is unnamed. Suggest adding it to #1803 rather than a new issue.

---

## Gates, checked at source per-step rather than off the label

`Shell tests (shellcheck + pithead suite)` steps, read from
`actions/jobs/101237559529` on run `33940746449`:

| step | conclusion |
|---|---|
| Install shellcheck + shfmt (pinned) | success |
| **Lint … (shellcheck + shfmt)** | **success** |
| **Run pithead test suite** | see below |

**Your shellcheck/shfmt gap is CLOSED by CI at this head** — the lint step ran and passed, so no
early exit skipped the suite (the #1797 shape you warned about did not happen here). The suite step
is the one that will report the #1740 row.

**CI HAS NOW REPORTED, AND IT CONFIRMS THE PREDICTION.** `Shell tests (shellcheck + pithead
suite)` at `4b1ccab5` is **`completed / failure`**, and the failing line in job `101237559529` is:

```
  ✗ every PASS label that interpolates a value is one the suite has reviewed
```

That is the #1740 row named above, and nothing else in the job reds. *(This paragraph is the
reviewer lane's measurement, read from the job log after the ephemeral exited — it is the one part
of this comment not written by the ephemeral.)*

Other checks at `4b1ccab5`: 12 pass (all 5 image builds, frontend node tests, hadolint, per-surface
lint, ruff, gitleaks, zizmor, compose hardening); dashboard pytest + dashboard-image pending; CVE
sweep skipping (schedule-gated, as always on a PR).

## Path grants — verified at source, not relayed

`roles/fixes.paths` carries, as **bare** lines (prose is not a grant):
- `lib/pithead/20-doctor-install-checks.sh` — appliance w112, 2026-09-05T01:35Z, for #1772 :28 and
  #1777 :67. ✅
- `tests/stack/test-doctor-exposure.sh` and `tests/stack/test-doctor.sh` — tests lane `pithead-88`,
  ~02:55Z, #1772 ONLY, revoke on merge. ✅
- `tests/stack/run.sh` and `pithead` — core lane grants. ✅

The granter's three conditions are all met: section header at `test-doctor-exposure.sh:33` matches
`^[[:space:]]*echo "== [^"]+ =="$`; file is not in `tests/inventory.sh:101`'s `SECTIONLESS`; sourced
at `run.sh:204` so `check_source_agreement` direction 2 is satisfied.

**One grant claim is stale and in your favour:** the grant text warns "run.sh is at 440 lines against
a 440 ceiling as of #1797". At this head `tests/stack/run.sh` is **356 lines** (350 at merge-base,
+6), there is **no row for it in `docs/dev/file-budget.tsv`**, and 356 < the 400 target. Your
`lint-file-budget rc 0` is consistent — no trimming was owed. New file is 67 lines, also no row.

## What I did NOT check

- Did not run `tests/stack/run.sh`, `make lint-sh`, shfmt or shellcheck locally — CI-only while a bench
  claim is held elsewhere (`fleet.sh claims`, since 2026-09-04T18:25Z). CI covers all four.
- Did not re-derive `docs/dev/STYLE.md` voice on the new comment blocks.
- Did not verify the appliance wording against a running appliance; the surface switch is proven by
  `PITHEAD_APPLIANCE=1` only.
- Did not check whether `42-control-policy-and-host-checks.sh` gained `STRATUM_BIND`/`STRATUM_PASSWORD`
  since you read it — I took that one from your body. Everything else above I re-derived.

## What is needed for a PASS

One change: de-interpolate the label at `test-doctor-exposure.sh:59-61`. Rebuild `pithead` is not
needed (the change is test-only). Push, and I or the next non-author pass will re-verify by re-running
`vd_interp_names` against the new head and reading the suite step. The two non-blocking findings are
follow-ups in other lanes' paths and should not hold this PR.

VijitSingh97:
## Head moved to `463973d7` while I was posting — the blocking finding is FIXED, exactly as recommended

The verdict above is anchored to `4b1ccab5` in every claim, so it stays accurate as a statement about
that sha. This carries it. **I re-derived the delta rather than assuming the push did what it says.**

`git diff 4b1ccab5 463973d7` is **two files, +7/-13**, nothing else:

**1. `tests/stack/test-doctor-exposure.sh` — the blocking fix, and it is the recommended route, not
the tempting one.** The loop is gone; three literal labels replace it, character-for-character the
unroll the review proposed. **It did NOT take the `vd_expected` route** — which matters, because that
route would have needed a `tests/stack/test-harness-tooling.sh` grant this lane does not hold, and
would have waved a row past the human-judgement step the #1740 guard exists to force. Net +0 lines to
the file's budget-relevant size beyond the two the unroll costs.

**2. `tests/stack/run.sh` — the sweep note compressed 10 lines to 4.** I checked what was dropped
rather than counting lines. **The load-bearing warning survives**: the note still names
`20-doctor-install-checks.sh:28`, still says it *"LEFT this list rather than being fixed here"*, and
still points at `test-doctor-exposure.sh` as where its guard now lives. That is the whole point of
the note — stopping the next reader from reading the site's disappearance as a fix by the sweep — and
it is intact in the shorter wording.

*Dropped, and I judge it non-blocking:* the old text's account of **why** it was urgent (it "was not
merely unguarded but LEAKING", and interpolated the host's public address into browser-bound text).
That is history rather than an instruction to the next reader, and #1772 keeps the record. Flagging
it only so the trim is a decision on the record rather than a silent loss.

### Verdict at `463973d7`

**The RETURN's blocking finding is discharged. The merge condition is now CI alone.**

`Shell tests` is the row that reds on this defect and it is **`in_progress`** at this head, along with
`Dashboard tests` and `Dashboard image`; the other 12 are green with the CVE pair schedule-skipped.
**Do not read the fix as settled until `Shell tests` reports at `463973d7`** — that job is the
instrument for this exact assertion, and a derived fix confirmed by CI is worth more than either
half. This one has already earned that: the finding was predicted from a verbatim extractor run
before CI reported, and CI then reddened on that assertion and only that one.

**I do not merge this PR** — `lib/pithead/20-doctor-install-checks.sh` and `pithead` are image-freeze
paths, so it is never the reviewer lane's to merge whatever the verdict. **On `Shell tests` green at
`463973d7` this is MERGE-READY for a non-author lane**, and **#1772 needs a hand-close with the merge
sha** (`Closes` is inert on `develop-v2`; it fires in rigforge only — do not carry the habit across).

### Still open from the review above, neither a blocker here

- **`tests/integration/selftest-redact-ip.sh:22-25`** — its #1609 rationale names this site as the
  producer of the interpolated address, and that is now false on both doctor arms. `tests/integration/`
  is not this lane's grant, so it is an issue or a message to the `e2e`/`tests` lane.
- **`20-doctor-install-checks.sh:19`** — the `dr_ok` on the same path carries `$bind` to the browser,
  which is the public address in exactly the #1803 case. Suggested as an addition to #1803 rather
  than a new issue.

Neither is re-derived by me; both are the ephemeral's, carried so they are not lost with the fix.

VijitSingh97:
## MERGE-READY at `463973d7` — green, and the new tests are proven to have RUN, not just labelled green

**All 17 checks: 15 SUCCESS, 2 SKIPPED** (the schedule-gated CVE pair). Nothing red, nothing pending.

**I did not settle this on the check label.** A green `Shell tests` is equally consistent with a brand
new domain file never being collected — that is the standing trap on any PR that adds a test file. Read
from the job log itself (`actions/jobs/.../logs`):

```
== unit: doctor's stratum-exposure verdict is worded for its surface (#1772) ==
  ✓ the host doctor verdict withholds the address (#1772)
  ✓ the appliance verdict withholds the address (#1772)
  ✓ the appliance verdict does not prescribe stratum_bind (#1772)
  ✓ the appliance verdict does not prescribe stratum_password (#1772)
  ✓ the appliance verdict does not prescribe the LAN firewall (#1772)
```

The section header proves `run.sh` sourced the new file; the three unrolled labels prove the fix is the
one executing, not the loop. **And the row that was RED at `4b1ccab5` is now green:**

```
  ✓ every PASS label that interpolates a value is one the suite has reviewed
```

So the blocking finding is discharged **by the same instrument that raised it**. That closes the loop:
derived statically from a verbatim extractor before CI reported, reddened by CI on exactly that
assertion, fixed, and now green on exactly that assertion.

### Verdict

**PASS at `463973d7`. MERGE-READY for a NON-AUTHOR lane — and that is NOT me and NOT `fixes`.**

- Not me: `lib/pithead/20-doctor-install-checks.sh` and `pithead` are image-freeze paths.
- Not `fixes`: they authored it. **The PR author field cannot distinguish lanes here — every lane pushes
  under the same account — so authorship is settled from the owning lane's role file, not from GitHub.**
  I got this exactly wrong on #1797 earlier today and am recording the correction rather than repeating it.
- **`lane-appliance-91` is the right merger**, as they were for #1797.

**Hand-close #1772 with the merge sha** — `Closes` is inert on `develop-v2`. (It *does* fire in
rigforge; that asymmetry has now bitten two lanes today, so check the repo rather than the habit.)

### Carried forward, neither blocking

- **`run.sh:330`'s bare locator `20-doctor-install-checks.sh:28`** is stale at this head (`:28` is a
  comment; the call is `:43`). The author flagged it themselves and deferred it to #1777, which
  rewrites that block at zero cost. **I agree with the deferral and am recording why:** the note is
  explicitly historical and carries no imperative, so a reader is mildly misdirected rather than
  misinstructed — not worth superseding a green run of a brand-new domain file on a freeze-path PR.
- **#1815** and **#1816**, correctly filed as follow-ups rather than folded into a green freeze-path PR.

Issue #1769, opened 2026-09-04
A failed dashboard save shows the appliance operator raw host CLI remedies it cannot reach


VijitSingh97:
## What happens

When a dashboard config save **fails**, the appliance operator is shown raw host CLI text telling them to run commands they have no way to reach.

`control_commit` puts `tail -c 2000` of the raw `apply` log into the result's `error` field (`lib/pithead/43-control-approval-and-preview.sh:364`), and `server.py:196` returns the whole result dict to the browser unfiltered (`web.json_response({"id": rid, **res})`). So whatever `apply` printed on the way down reaches the browser verbatim, including:

- `Stack is not fully provisioned. Run './pithead setup' first.` (`lib/pithead/40-apply-and-render.sh:131`)
- `Fix the cause shown above, then re-run './pithead apply'` (`lib/pithead/40-apply-and-render.sh:289`)

## Why this is the same defect class as #1213, on a surface #1213 cannot see

#1213 fixed doctor's remedial strings so an appliance operator is never told to run a CLI verb. Its coverage is a literal sweep over `dr_fail`/`dr_warn` messages in `tests/stack/run.sh`. This path emits operator-facing remedies **without going through `dr_*` at all** — it is a log tail passed through to the browser — so the sweep is structurally blind to it and always will be.

It is also, specifically, the failure path of the remedy #1213's new wording now prescribes: the fixed doctor strings tell the operator "saving any change from the dashboard provisions it", and if that save fails this is what they read.

## Where it came from

Found during the third non-author review pass on PR #1768 (#1213) as a same-class hunt beyond the strings in the diff. Read from the handler, not observed in a running UI — nobody has driven a failing save end to end and watched the browser. That is the one thing to confirm before fixing.

## Fix shape (suggestion, not a ruling)

The `error` field needs the same appliance-awareness treatment the `dr_*_surface` helpers got in #1139/#1213, or the passthrough needs to stop being a raw tail. Note that suppressing the tail entirely loses genuinely useful diagnostic detail, so this is a wording/branching problem rather than a redaction one.

## Scope note

`lib/pithead/43-control-approval-and-preview.sh` and `dashboard/.../server.py` are not the fixes lane's paths; this is filed for whoever owns them, not claimed.

VijitSingh97:
[...]

PR #1818, opened 2026-09-05
test(stack): #1777 the #1213 verb-leak sweep catches the imperative, not only the executable

### LINKED ISSUE #1777: The #1213 verb-leak sweep reports clean over #1776, a site it fully covers — the needle gap, not the documented blind spot

VijitSingh97:
Closes #1777 (inert on `develop-v2` — hand-closed on merge).

The #1213 verb-leak sweep in `tests/stack/run.sh` enumerates its SITES mechanically out of the
built `pithead`, but matches them against a hand-written alternation. That alternation carried
`\./pithead ` and had no token for a verb written as a bare or quoted word, so it returned 0
matches over #1776's site — a plain `dr_warn` literal it fully covers — and #1213 closed with
that string still naming a host remedy.

## What changed

1. **The alternation now catches the imperative, not only the executable.** New branch:
   `([Rr]e-?)?[Rr]un '?(\./pithead )?(apply|setup|up|down|restart|doctor|status|logs)'?`.
2. **`lib/pithead/20-doctor-install-checks.sh`'s "Run doctor there" info takes the
   `appliance-unreachable` marker**, with its reason on the same line, the way
   `05-doctor-checks.sh:142` already does. That verdict is gated on the DIY versioned layout —
   basename `pithead-vX.Y.Z` AND a sibling `current` symlink — and the appliance's `/opt/pithead`
   install creates neither, so it is unreachable there by construction rather than a conversion
   candidate.
3. **The sweep's site pattern, needle and exemption filter are now ONE helper, `_dr_leaks`,** which
   both the law and the control call. A control that re-spells the grep is mutated in lockstep with
   nothing; this one cannot drift from what it validates.
4. **A seeded control**, because after item 2 the widened branch has NO live target: its 0 is
   indistinguishable from a needle that matches nothing at all.

## Why the control is not optional here

RUN BY ME, four legs, one variable moved between the first two:

| leg | input | needle | result |
|---|---|---|---|
| 1 | seeded pre-marker text | widened | **fires** (1 hit) |
| 2 | seeded pre-marker text | pre-#1777 | **empty** |
| 3 | shipped `pithead` | widened | empty — the sweep's green |
| 4 | shipped `pithead`, my marker stripped | widened | **fires**, naming the real site |

Leg 2 is the differential: the same seed against the same instrument with only the alternation
moved, which is what rules out "the seed fires for some other reason". Leg 4 shows leg 3's green is
caused by the marker rather than by a dead needle. Leg 1 alone would have proved neither.

Also re-derived: the pre-#1777 needle's ONLY hit over the artifact is `05-doctor-checks.sh:142`,
which already carries the marker — so the sweep was green before and is green after, and the
difference is entirely in what it is now able to see.

## Line budget

`tests/stack/run.sh` was 440 against a 440 ceiling with zero headroom, and ceilings only go down.
The new helper, control and comments were paid for by compressing seven comment blocks this lane
wrote for #1776/#1772 — per-edit deltas -2, -1, -1 and four at 0, then +4 for the block-3 rewrite.
**Final: 440 lines, exactly the ceiling; `docs/dev/file-budget.tsv` is UNTOUCHED.**
`lib/pithead/20-doctor-install-checks.sh` is line-neutral (a trailing comment) and has no budget row.

## Folded in

The re-pass nit deferred into this item: block 3's comment named
`20-doctor-install-checks.sh:28` as a bare historical locator, which at the current head is neither
the call nor the right line. It now names the site by what it IS — the stratum-exposure warn that
#1772 converted to `dr_warn_surface` — so a reader cannot mistake a gravestone for a live pointer.

## Scope

The doctor-surface class is closed at #1772 and #1776; this makes the instrument able to say so.
#1769 is a different surface (the raw apply log) and is not touched here.

## RUN BY ME at this head

- `bash tests/stack/run.sh` — **rc 0, `pithead tests: 3346 passed, 0 failed`**, zero `✗` marks in the
  whole log. The new control row is at log line 3699, immediately above the sweep row it validates
  (3700), so the control is proven to have EXECUTED and not merely to exist in the file.
- `make lint-sh` — **rc 0**, run under `flock ~/.lint-sh-manual.lock`, never the shellcheck-serialize
  lock. Pins read from the Makefile itself: shfmt **v3.13.1**, shellcheck **0.11.0**, so this is that
  gate's result and not some other tool's.
- `make lint-file-budget`, `lint-pithead-parity`, `lint-topology`, `lint-docs-voice`, `lint-md` — all rc 0
- the four control legs above
- `~/dev/fleet/lane-guard.sh` — rc 0 on this change staged alone, with a FIRED negative control: a
  one-byte edit to `lib/pithead/14-local-miner.sh`, a sibling in the same directory, returns rc 1
  "Another role owns this code". A near-miss sibling rather than a distant file, so the rc 0 means
  "granted", not "the guard waves everything through". Probe file restored; worktree clean.
- `python3 ~/code/pithead-topography/detect-topology.py` over this body — clean. Its control was this
  same text plus a seeded rig hostname, which returned an ADVISORY; the seeded IP and `user@host` did
  not fire, so the control shows the detector reading THIS draft and reaching a non-clean verdict, and
  does not establish coverage of every class.

## NOT run by me / ASSUMED

- I did NOT re-run the suite at `origin/develop-v2` to produce a before/after row count, so I make no
  delta claim: 3346 is this head's figure, and the 3336 in pull 1813's record was a different tree.
- No bench, no KVM, no image build. No dashboard or integration tier — this change is tier 1 only.
- No doc update is owed and none is shipped: I grepped `docs/` for the sweep, the marker and #1213
  and no prose doc describes this instrument or carries a count my change moves. Saying so rather
  than shipping a doc edit to satisfy the bar.

## Over-engineering pass (run by me, no command exists for it)

Four things I checked for being more machinery than the problem needs, and one cost I am
accepting rather than hiding:

- **`_dr_leaks` earns its line.** It replaces a 3-line pipeline with 1, so it is a net
  simplification even before the control needs it — and the control is the reason it must exist,
  since a control that re-spells the grep is mutated in lockstep with nothing.
- **The marker, not a `dr_info_surface` conversion.** Converting would add an appliance wording for
  a verdict no appliance operator can reach, which is a second string to keep true for no reader.
- **The verb list stays the full pithead set**, not narrowed to the two verbs that hit today.
  Narrowing to the known hit is what the issue names as the thing to avoid.
- **The comment block shrank.** Block 3's prose is 21 lines against a 10-line instrument, which is
  the one place a reviewer could reasonably push back; it is carrying the blind-spot disclosure,
  and I paid for the new material by cutting elsewhere rather than adding.

**The cost I am accepting, named because a reviewer should get to veto it:** `down`, `up`, `status`
and `logs` are ordinary English words, so a future verdict phrased "…will run down its buffer"
would match and red this row falsely. I kept them because the alternative is fitting the needle to
today's two hits. The failure is in the safe direction — the sweep prints the offending line, and
the fix is a marker or a reword, not a silent miss — but it is a real maintenance cost and it is
mine to have chosen.

## Returned once, and what the red actually was

CI reddened `Lint per-surface` at `7b233398` on `lint-operator-strings`. **My defect**, and a gap I
own: I ran the lint targets individually and did not run `make lint`, so I never ran that one.

The marker I added carried `(#1777)`, and it sits in a trailing comment on the same line as the
`dr_info` call. `scripts/lint-operator-strings.sh` selects lines by call shape and then greps the
**whole line** for `#[0-9]+`, so an issue number in a comment is indistinguishable from one in
operator text — even though the linter's own failure message says "comments keep them".

**Fix: drop `(#1777)` from the marker, leave the marker where it is.** That matches the only
existing precedent, `lib/pithead/05-doctor-checks.sh:142`, whose marker carries a reason and no
issue reference, which is why it has never tripped this.

**The fix that looks right and is not:** moving the comment to its own line, which is what the
linter's message invites. `_dr_leaks` ends in a line-based `grep -v "appliance-unreachable"`, so the
marker suppresses a site only while it shares the call's line. Moving it trades a lint red for a
**sweep** red — and that one reads like a real leak. Caught by the reviewer, who reasoned it from
the line-based `grep -v`; **I then MEASURED it rather than leaving it as agreed reasoning**, because
two passes converging is two opinions and neither of us had run it. Moved the marker onto its own
line in a scratch copy of the artifact and ran the sweep: **1 hit, naming the real site**, against
**0** over the real artifact with the marker in place. One variable moved, paired control, so the
claim is now a measurement.

Reproduced both ways locally: `make lint-operator-strings` names exactly that one line before the
change and passes after. All four control legs re-run on the rebuilt artifact and are unchanged.

**Not fixed here, and not filed:** the linter cannot distinguish a trailing comment from operator
text. `scripts/lint-operator-strings.sh` is in no lane's `.paths` I hold, the milestone that could
carry a tooling fix is frozen, and an issue with no active milestone is in no queue whatever its
label says — so filing it would make it invisible rather than tracked. Recorded here, where the next
person to add a marker will hit it.

### DIFF
--- a/lib/pithead/20-doctor-install-checks.sh
+++ b/lib/pithead/20-doctor-install-checks.sh
@@ -82,7 +82,7 @@ check_control_units() {
     if [[ "$_name" =~ ^pithead-v[0-9]+\.[0-9]+\.[0-9]+$ ]] && [ -L "$_parent/current" ]; then
         _live=$(cd "$_parent/current" 2>/dev/null && pwd -P)
         if [ -n "$_live" ] && [ "$_live" != "$here" ]; then
-            dr_info "This is not the live install — '$_parent/current' points at $_live. Run doctor there to check its control channel."
+            dr_info "This is not the live install — '$_parent/current' points at $_live. Run doctor there to check its control channel." # appliance-unreachable: the DIY versioned layout only -- the guard above needs basename pithead-vX.Y.Z AND a sibling `current` symlink, and the appliance's /opt/pithead install creates neither
             return 0
         fi
     fi
--- a/pithead
+++ b/pithead
@@ -4928,7 +4928,7 @@ check_control_units() {
     if [[ "$_name" =~ ^pithead-v[0-9]+\.[0-9]+\.[0-9]+$ ]] && [ -L "$_parent/current" ]; then
         _live=$(cd "$_parent/current" 2>/dev/null && pwd -P)
         if [ -n "$_live" ] && [ "$_live" != "$here" ]; then
-            dr_info "This is not the live install — '$_parent/current' points at $_live. Run doctor there to check its control channel."
+            dr_info "This is not the live install — '$_parent/current' points at $_live. Run doctor there to check its control channel." # appliance-unreachable: the DIY versioned layout only -- the guard above needs basename pithead-vX.Y.Z AND a sibling `current` symlink, and the appliance's /opt/pithead install creates neither
             return 0
         fi
     fi
--- a/tests/stack/run.sh
+++ b/tests/stack/run.sh
@@ -288,8 +288,8 @@ echo "== unit: doctor's remedial strings are surface-aware (#1213) =="
 # Doctor's verdicts reach the dashboard verbatim now that the diagnostics verbs ship them
 # (control_diag_doctor runs `doctor --json`; doctor_json emits every message as {status, message}),
 # so a remedial string naming a CLI verb is a dead end on an appliance, which has no shell. Two
-# instruments, because either one alone passes for the wrong reason: the switch has to actually
-# FLIP, and no verdict may be left behind it.
+# instruments, because either alone passes for the wrong reason: the switch has to actually FLIP,
+# and no verdict may be left behind it.
 #
 # 1. The mechanism. Argument one is the DIY/host wording, argument two the appliance's; each side
 #    must print its own and NOT the other's -- asserting only that the right text appears would
@@ -309,23 +309,20 @@ for _s in fail warn info; do
     esac
 done
 
-# 2. The #1776 site, named because the sweep below could not see it. Its verdict prescribed two
-#    host-only remedies ("Move the data here" and the verb as the bare quoted word 'apply'), and
-#    the literal alternation carries `\./pithead ` but no token for that quoting -- so the sweep
-#    returned 0 matches over a site it fully covers, and #1213 closed with this string still plain.
-#    Read off the SHIPPED artifact, the same instrument the sweep uses, so a lib/ edit that never
-#    reaches `pithead` cannot pass this.
+# 2. The #1776 site, named because the sweep below could not see it: its verdict prescribed two
+#    host-only remedies ("Move the data here" and the verb as the bare quoted word 'apply'), which
+#    the sweep's alternation had no token for until #1777 widened it. Read off the SHIPPED
+#    artifact, the same instrument the sweep uses, so a lib/ edit that never reaches it cannot pass.
 #
-#    AND THIS BLOCK IS NOW THE SITE'S ONLY VERB GUARD. Converting it to dr_warn_surface takes it
+#    AND THIS BLOCK IS NOW THE SITE'S ONLY VERB GUARD. Converting it to dr_warn_surface took it
 #    OUT of block 3's sweep by construction -- that sweep's site pattern is `dr_warn "`, which a
-#    `dr_warn_surface "` call does not match, and the exclusion is deliberate because a host
-#    argument is SUPPOSED to keep its verb. So the single literal in _dd_pred_verb is all that
-#    stands behind the appliance side of this one verdict. Widen it here, not there.
+#    `dr_warn_surface "` call does not match, deliberately, because a host argument is SUPPOSED to
+#    keep its verb. The single literal in _dd_pred_verb is all that stands behind the appliance
+#    side of this verdict. Widen it here, not there.
 #
-#    THE CONTROL IS NOT OPTIONAL. Both assertions below are ABSENCE claims over a grep, and an
-#    absence claim goes green when the needle simply stops matching -- rename the message and this
-#    passes forever while proving nothing. So the same two predicates run against a synthetic
-#    PRE-FIX line first and must FAIL there.
+#    THE CONTROL IS NOT OPTIONAL. Both assertions below are ABSENCE claims over a grep, and one
+#    goes green when the needle merely stops matching -- rename the message and this passes forever
+#    proving nothing. So the same two predicates run against a synthetic PRE-FIX line and must fire.
 _dd_pred_plain() { case "$1" in *dr_warn_surface*) return 1 ;; *) return 0 ;; esac }
 _dd_pred_verb() { case "$1" in *"run 'apply'"*) return 0 ;; *) return 1 ;; esac }
 _dd_seed="            dr_warn \"Data dir from .env not found: X=Y — a relocated/copied install re-syncs from scratch. Move the data here, or set the data_dir in config.json and run 'apply'.\""
@@ -348,9 +345,9 @@ else
     # The host side KEEPS the verb by design (it is the DIY wording); only the appliance side must
     # not. Take the second quoted argument, the way the helper's own contract orders them.
     _dd_appl=$(printf '%s' "$_dd_line" | awk -F'"' '{print $4}')
-    # ...and prove the field EXISTS before reading an absence out of it. On the pre-fix shape --
-    # one quoted argument -- $4 is the empty string, _dd_pred_verb "" returns rc 1, and the row
-    # below would print ok having examined nothing. Only its sibling reds that case today.
+    # ...and prove the field EXISTS before reading an absence out of it: on the pre-fix shape (one
+    # quoted argument) $4 is empty, _dd_pred_verb "" returns rc 1, and the row below would print ok
+    # having examined nothing. Only its sibling reds that case today.
     if [ -z "$_dd_appl" ]; then
         bad "the data-dir verdict's appliance wording names no host verb (#1776)" "no second quoted argument on: $_dd_line"
     elif _dd_pred_verb "$_dd_appl"; then
@@ -360,10 +357,9 @@ else
     fi
 
     # The appliance wording NAMES which of these dirs the dashboard can repoint, so it is a claim
-    # about CONTROL_DASHBOARD_CONFIRM_KEYS -- and the first draft of it ("no dashboard control
-    # relocates a data directory") was false for four of the five keys the check fires on. Derive
-    # both sets from the shipped artifact rather than restating them, so an allowlist change reds
-    # HERE and names the string to rewrite, instead of leaving an operator a remedy they lack.
+    # about CONTROL_DASHBOARD_CONFIRM_KEYS -- and its first draft ("no dashboard control relocates a
+    # data directory") was false for four of the five keys the check fires on. Derive both sets from
+    # the shipped artifact, so an allowlist change reds HERE and names the string to rewrite.
     _dd_in_set() { case " $2 " in *" $1 "*) return 0 ;; *) return 1 ;; esac }
     _dd_repointable() {
         local _k _out=''
@@ -376,8 +372,8 @@ else
     # strictly REDUNDANT, its pass condition being exactly what the CANNOT row asserts, so it could
     # never red alone. This closes what those rows cannot see -- a second `for var in ..._DATA_DIR;
     # do` above missing_data_dirs would feed them, through head -1, a set the shipped code no longer
-    # uses, every row still green. The awk keeps a narrower gap: a trailing comment after its
-    # closing quote makes it over-read. Wrong input, no wrong output today, so NAMED not guarded.
+    # uses, every row green. The awk gap is narrower: a trailing comment after its closing quote
+    # makes it over-read. Wrong input, no wrong output today, so NAMED not guarded.
     assert_eq "the data-dir key list comes from exactly one site (#1776)" "$(printf '%s\n' "$_dd_sites" | grep -c .)" "1"
     _dd_conf=$(awk "/^CONTROL_DASHBOARD_CONFIRM_KEYS='/{f=1} f{print} f && /'[[:space:]]*\$/{exit}" "$STACK" |
         tr -d "\n'" | sed "s/^CONTROL_DASHBOARD_CONFIRM_KEYS=//;s/  */ /g;s/^ //")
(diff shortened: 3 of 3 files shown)


VijitSingh97:
## VERDICT: RETURN at `7b233398` — non-author review, reviewer lane

One blocking defect, diagnosed to a single token. **The fix is NOT the obvious one**, and the obvious one would break a different guard — that is the whole content of this review, so it leads.

### The red is yours, and the two guards genuinely collide

`Lint per-surface` fails on `lint-operator-strings`, at `pithead:4931`:

    dr_info "... Run doctor there to check its control channel." # appliance-unreachable: ... (#1777)

Mechanism, read at source rather than inferred. `scripts/lint-operator-strings.sh:14` selects lines by CALL SHAPE — `^[[:space:]]*(dr_info|...)[[:space:]]*"` — and emits the **whole line**; Rule 1 (`:55`) then greps that line for `#[0-9]+`. Your marker shares the line with the call, so the `(#1777)` inside a *comment* is indistinguishable from an issue number in operator text. The linter's own note says it: "operator-facing text drops issue/PR references, comments keep them" — and its self-test fixture proves the exemption exists, with the comment on its OWN line (`:139`, `clean pithead + a #NNN comment is not flagged`).

**⛔ SO DO NOT MOVE THE COMMENT TO ITS OWN LINE. That is the fix the linter's message points at and it would break `_dr_leaks`.** The marker's consumer is `tests/stack/run.sh:419`:

    grep -nE '(dr_fail|dr_warn|dr_info) "' | grep -E "<host-verb patterns>" | grep -v "appliance-unreachable"

That is a LINE-BASED pipeline: the marker suppresses a site only while it sits on the same line as the call. And this site does need suppressing — its text "Run doctor there" matches the alternation's `[Rr]un ... (doctor|...)` branch, so with the marker on its own line the site becomes a fresh `_dr_leaks` hit. You would trade one red for another.

**The fix is to drop `(#1777)` from the marker text**, leaving it on the call line. That is exactly the shape of the one existing precedent, `lib/pithead/05-doctor-checks.sh:142`, whose marker carries a reason and no issue reference — which is why it has never tripped this linter. If you want the traceability, `(issue 1777)` also passes Rule 1 (the pattern is `#[0-9]+`, so the `#` is what trips it), but the precedent is the cleaner call and the issue number is already on this PR and in the `run.sh` block.

Regenerate `pithead` from the slice; both copies carry the same line.

### What I checked that could have failed, and did not

**The reflow held its wrap width.** `run.sh` is 46+/46− — a comment rewrite that keeps the file at exactly 440 lines. A reflow that holds LINE COUNT can still blow the wrap convention, so I measured it: longest comment line is **103 at the merge-base and 103 at your head**, unchanged. The count of lines over 100 moves 90 -> 92, and both additions are pre-existing long `dr_*` strings rather than reflowed comments.

**Nothing load-bearing was lost in the reflow**, on my reading. The compression drops the *consequence* ("#1213 closed with this string still plain") while keeping the *mechanism* ("the sweep's alternation had no token for it"), which is history rather than instruction, and the "THE CONTROL IS NOT OPTIONAL" block keeps its point. One wording change is arguably an improvement: "an absence claim goes green" -> "one goes green" is the more accurate statement, since `_dd_pred_plain` tests an absence and `_dd_pred_verb` tests a presence.

### Not mine to merge, and a sequencing note

**`lib/pithead/20-doctor-install-checks.sh` and `pithead` are IMAGE FREEZE paths**, so on a re-pass this is MERGE-READY for the appliance lane's sequencing, never my merge. Time it deliberately: the #1651 gate is currently RED and the freeze tip moves again with the rig-journal fix, so landing this before that fix means the gated tree changes underneath it.

### A systemic note for whoever owns the budgets, not a finding against this PR

This PR spends a **46-line comment reflow to land a one-line fix**, purely because `tests/stack/run.sh` sits at 440 against a 440 ceiling with zero headroom. That is a real tax and it is now being paid in explanatory prose — the thing those comments exist for. Two other open PRs (#1747, #1737) are each blocked on the same row needing 3 lines paid on their own branch. Worth someone raising the ceiling deliberately or splitting the file, rather than each lane quietly compressing comments to fit.

### Disclosure

I ran nothing: no suite, no lint, no `make`, no container. The CI-only rule is in force and I hold no locks or claims. Every claim above is source reading of the linter and its consumer at the named refs, plus the failing job log. **I did not execute either guard**, so the collision is derived from reading both implementations — the cheapest confirmation is that your own CI already shows one half of it red.

VijitSingh97:
## VERDICT: PASS at `89496220` — non-author review, reviewer lane. **MERGE-READY, not mine to merge.**

My RETURN at `7b233398` is DISCHARGED. Re-derived at this head rather than carried forward.

**The fix is exactly the minimal one, verified not asserted.** `git diff --stat 7b233398 89496220` is two files, one line each — the marker token in `lib/pithead/20-doctor-install-checks.sh` and its regenerated `pithead`. Nothing else moved. In particular `tests/stack/run.sh` is **byte-identical** between the two heads (`git diff --quiet` clean), still 440 against its 440 ceiling, so the 46-line reflow I checked at the previous head does not need re-checking and the budget row is untouched.

**The rule the red enforced now holds tree-wide, not just at the fixed site:** over the regenerated artifact, `dr_*` call lines carrying a `#[0-9]+` number **0**. That is the linter's own Rule 1 shape applied as a sweep rather than a spot check on the one line that failed.

**And the marker still does its job.** I re-ran my `_dr_leaks` counterfactual against THIS artifact, not the old one: marker in place **0 hits**, marker moved to its own line **1 hit** naming the real site, with a `cmp` control proving the mutation landed. So dropping `(#1777)` did not weaken the suppression — the property depends on the marker sharing the call's line, not on its text. That was the author's point and it is now measured on both artifacts, which are genuinely different files.

### Three legs, and they are different evidence

Worth stating because "confirmed" would hide it. The derivation — `lint-operator-strings.sh:14` selects by call shape and emits the whole line, `:55` greps it for `#[0-9]+`, while `_dr_leaks` (`run.sh:419`) is a line-based `grep -v` — is mine, from reading. The first measurement, with the linter fired both ways, is the author's. The independent confirmation on a second artifact is mine. Two passes agreeing would be two opinions; what makes this one strong is that the METHODS differ.

### What is NOT claimed here

The author disclosed a local `make lint` exit 137 and correctly did not read it as a code defect: it is `shellcheck-real` OOM-killed at the wrapper's 6G cap (the known #1206 ceiling), corroborated by a standalone `make lint-sh` at rc 0 over a byte-identical `run.sh` and by CI's own shellcheck step passing at the previous head. I have **not** re-derived that OOM attribution and am not relaying it as verified — it is the author's measurement, and CI is the authority for this head either way. Their local stack-suite rerun was still in flight and is likewise not claimed.

I ran no suite, no `make`, no lint, no container — CI-only rule, no locks held. My checks above are `git` plumbing plus two greps reconstructed verbatim from `run.sh:419`.

### Merge condition and routing

**PASS at `89496220`, conditional on green CI.** At the time of writing: 5 success, 2 schedule-gated skips, 0 failures, 10 still running — `UNSTABLE` is the running checks, not a red. **`lib/pithead/` and `pithead` are IMAGE FREEZE paths, so this is NOT mine to merge**; `lane-appliance-91` merges it non-author on this recorded PASS plus green, per the sequencing they agreed with the author. Whoever merges: re-read this PR's comments in the same call that merges — a later verdict can land at an unchanged head, and that failure leaves no tell.

PR #1819, opened 2026-09-05
fix(os): #1817 a rig's journal reclaim tolerates the persistent bind, and the bind stands down on a rig

### LINKED ISSUE #1817: A provisioned rig boots mining and never commits its slot: the journal reclaim meets the #1792 bind mountpoint and errexit takes `local-miner` down
**A provisioned rig boots mining and never commits its slot: `pithead local-miner` dies at the journal reclaim, because after #1792 `/var/log/journal` is a bind mountpoint on every boot.** Found by the #1651 gate on `BUILD_COMMIT 4fb4943a` (187/7 — all seven reds are this one failure and its A/B inheritance); the rig reboot leg was green at `670c33e9`, and #1792 is the only product change between them.

This is NOT #1791 reopened. #1791's ordering (`After=var.mount`) is right and stays. The rig role's cleanup assumed a plain directory, and the #1791 fix made the bind win every boot instead of half of them.

### Mechanism (re-derived at source, then measured on-image)

- `os/overlay/pithead-journal-persist` binds `/data/pithead/journal` onto `/var/log/journal` in the early boot transaction, on every machine — a rig included, whose journal is volatile by design (the battery's own row says so).
- `rig_minimize_writes` (`lib/pithead/14-local-miner.sh`) writes the volatile drop-in, then `rm -rf /var/log/journal`, then restarts journald. `rm -rf` on a mountpoint empties it — through the bind, onto `/data` — and then fails on the mountpoint itself: `rm: cannot remove '/var/log/journal': Device or resource busy`, rc 1. That `rm` is the only unguarded command in the function.
- The first-boot / reboot split is the CALL SITE, not the ordering: the wizard calls `provision_rig_miner || true` (errexit suppressed for the whole subtree — the `rm` fails, the function runs on, the miner starts, first boot mines), while `pithead-boot` runs `./pithead local-miner` bare under the CLI's `set -Eeuo pipefail` (errexit fires at the `rm`, the verb exits 1, `fail_boot "the rig's miner setup failed"`, slot uncommitted, miner never set up). A reorder of the minimization would NOT fix this — the verb would still exit non-zero.

### On-image, kept `--phase rig` guest of a fresh build of `4fb4943a` (verify-image 103/0)

- First boot, serial console: the `rm: cannot remove '/var/log/journal': Device or resource busy` line, then the function's own "Rig write minimization" log line, then RigForge setup — the tolerant path.
- Reboot: `pithead-boot` `ExecMainStatus=1 Result=exit-code`; grubenv `A_OK=1 A_TRY=1` (uncommitted); `xmrig` inactive and no runtime unit rendered; `/var/log/journal` is `/dev/vda4[/pithead/journal]` (the bind); the volatile drop-in is present (the step before the `rm` ran); journald `NRestarts=0` (the step after the `rm` never ran); the console carries journald's `Failed to create new system journal: No such file or directory` at 19 s (the `rm` removed the machine-id directory under it).
- **The failure erases its own evidence.** `journalctl -u pithead-boot -b` on the failed boot is "No entries": the `rm` emptied the bound journal before failing, and journald re-created the file only after `pithead-boot` had exited. A field rig hitting this has no journal of it — a diagnosability defect on its own, and the reason the proof above leans on systemd's unit record and the console.

### Fix shape (one PR, both halves, tier-1 rows for each, the battery row that caught it)

(a) `pithead-journal-persist` reads `/data/pithead/machine-role` — the same marker `pithead-boot` forks on, read the same way — and exits 0 without binding when it says `rig`. `After=var.mount` stays.
(b) `rig_minimize_writes` restarts journald FIRST (it holds the persistent files open; a bind cannot come off while it does), takes the bind off if `/var/log/journal` is a mountpoint, and reclaims best-effort; the call site becomes `rig_minimize_writes || true` so no future failure inside a stick-wear optimisation can leave a slot uncommitted.

The battery's rig reboot row exercises (a) (no bind on a provisioned rig's boot); (b) is what a rig's FIRST boot exercises (the bind is up before the role is known) and is proven under errexit in tier 1 with a fired control. Image-bytes change: new freeze tip, full gate again.

VijitSingh97:
Fixes #1817 (hand-close on merge — `Closes` is inert on `develop-v2`). Found by the #1651 gate on `BUILD_COMMIT 4fb4943a` (187/7, all seven reds one failure); on-image proof and the retraction of my earlier "runs after the miner start" sentence are on #1651.

**What broke.** #1792 made the persistent-journal bind win `/var/log/journal` on every boot. The rig role's write minimization assumed a plain directory there: `rm -rf` emptied it through the bind (onto `/data`) and then failed on the mountpoint (`Device or resource busy`). The wizard calls that leg under `|| true`, so first boot mined; `pithead-boot` runs `./pithead local-miner` bare under the CLI's errexit, so every later boot exited 1, `fail_boot`, slot uncommitted, miner never set up. The failure also erased its own journal.

**Both halves, as the seat's directive shaped them:**

(a) `os/overlay/pithead-journal-persist` reads `/data/pithead/machine-role` — the marker `pithead-boot` forks on, read the same way (whitespace-stripped, the exact word) — and exits 0 without binding when it says `rig`. `After=var.mount` and everything else from #1791 stay. Override for tier 1: `PITHEAD_MACHINE_ROLE_FILE`.

(b) `rig_minimize_writes` (`lib/pithead/14-local-miner.sh`, `pithead` regenerated): the journald restart moves BEFORE the reclaim — journald holds the persistent files under the path open, and a bind cannot come off while it does — then the bind is taken off if the path is a mountpoint, then the directory is reclaimed best-effort (`|| warn`). The call site becomes `rig_minimize_writes || true`.

**Which half the battery row exercises.** The rig reboot row (`tests/os/run.sh` § rig, "pithead-boot owns a provisioned rig's boot" + the self-commit row) exercises (a): on a provisioned rig's boot there is no bind, so the reclaim meets a plain directory as before #1792. (b) is what a rig's FIRST boot exercises — the bind is up before the role is known — and that boot already passed under the wizard's `|| true`, so the battery cannot discriminate (b); tier 1 does, under errexit, with fired controls (table below).

**Half (b) proven on-image, on the UNFIXED image's failed rig boot** (kept `--phase rig` guest of `4fb4943a`, after its phase ended; evidence dir `~/battery-1791-rigdiag-20260905T061417Z/guest-experiment-b-*.txt` on the build box), in the fix's order with the naive orders as controls:

| step | rc |
|---|---|
| control: `umount /var/log/journal` while journald is still persistent | 32, `target is busy` — the directive's "umount first" alone would not have fixed it |
| control: `rm -rf /var/log/journal` on the mountpoint | 1, `Device or resource busy` |
| `systemctl restart systemd-journald` (the volatile drop-in was already there) | 0; journald's open fds under the path go 1 → 0 |
| then `umount` | 0, no longer a mountpoint |
| then `rm -rf` | 0, gone |
| then `./pithead local-miner` exactly as `pithead-boot` calls it | 0; `xmrig` active, process running |

**Tests, tier 1 (run: `tests/stack/run.sh` in the branch worktree, and the journal domain standalone — counts below).** Full suite on this branch before the shfmt-only reformat of the rig-miner fragment: `pithead tests: 3352 passed, 0 failed`, rc 0; after it, the journal domain 17/0 and the rig-miner fragment 90/0 again via the mini-runner. CI runs the whole suite over the pushed tree.
- `test_firstboot_journal.sh` (+6 rows): rig marker → rc 0, `mount` never called, the script says why; sibling controls `pithead`, near-miss `rigs`, and NO marker all still bind.
- `test-appliance-rig-miner.sh` (+8 rows): a bound path → the call log reads `restart journald, umount <path>, rigforge setup` in that order and the directory is gone; an errexit-ARMED runner (with a control row proving it is armed) drives `provision_local_miner` with `rm` stubbed to refuse the journal path exactly as the mountpoint case does → rc 0, RigForge setup still ran, the console says the journal was not reclaimed.

**Mutation battery** (detached checkout of this commit, each mutant proven applied by `git diff --numstat`, restored by checkout; the fixed tree is 90/0 on the rig-miner fragment and 17/0 on the journal domain):

| mutant | rows reddened |
|---|---|
| M2 restart moved back after the rm | the order row |
| M3 umount block dropped | the order row (same instrument as M2 — one axis) |
| M4 `\|\| warn` on the rm dropped (call site still `\|\| true`) | "the console says the journal was not reclaimed" |
| M5 overlay role check dropped | "mount is never called", "says why it stood down" |
| M6 errexit runner disarmed | the control row — the control can fail |
| M7 BOTH guards dropped | the rc row, the setup row, the console row — the rc row can fail |
| **M1 call-site `\|\| true` dropped alone** | **SURVIVES, by construction**: every step inside the function is now guarded, so nothing reaches the call site non-zero. It is defence against the next unguarded line (this bug's exact signature: passes first boot, kills every boot after) and is NOT proven by a row. Said here rather than hidden. |

**Over-engineering pass, by hand** (the ponytail gate keys off the pinned shell cwd's branch here, not this worktree's — disclosed): the diff is two guarded lines plus a reorder in one function, a 12-line early exit in one script, and rows. No new file, no new helper. Adjacent merges considered: folding the role check into the `.service` as `ConditionPathExists`/`ConditionFileNotEmpty` was available and rejected — systemd conditions cannot test a file's CONTENT, and the marker's value is what forks; a `machine_role()`-style helper shared with the CLI was rejected because the overlay is `#!/bin/sh` and never sources the CLI (same reason `pithead-boot` open-codes the read).

**Not done / not touched:** no `verify-image` row (the file sits at its 400-line bar with no budget row; the archive-vs-tree check with `PITHEAD_EXPECT_COMMIT` already proves the image carries the tree's script; the tier-1 rows prove the script). No prose doc names the reclaim or the bind; the comments in both files carry the record. `CHANGELOG.md` untouched (as #1792). `tests/os/verify-image.sh` count stays 103. Image-bytes change: **new freeze tip on merge; the full gate reruns**, and the reviewer lane's design finding on #1817 (whether `pithead-boot:308` should `fail_boot` at all, given the liveness gate below it) is filed there, not taken here.

---

## Review round 1 (ephemeral reviewer, RETURN at the first head) — both items taken

**Blocking, SC1090:** the new `source "$STACK"` inside `run_rig_errexit()` carries the line-above directive (house shape for a source inside a function). Shellcheck through the box's wrapper fired on the old text (rc 1) and passes on the new (rc 0). At that head CI's `Shell tests` stopped in step 4, so steps 5–11 SKIPPED and CI had proven nothing about the suite; the rerun at the current head is that proof.

**Finding 2, taken — a bind that will not come off is left alone.** `rig_minimize_writes` re-tests `mountpoint -q "$journal"` after the `umount`; still mounted → `warn` + `return 0`, and `rm -rf` never runs through the bind onto /data. Proof, run not intended: the rig-miner fragment is 93/0 on the fix; the mutant that drops the re-check (proven applied by `git diff --numstat`) reddens exactly two rows — `a bind still up is not reclaimed through` (the directory under the bind survives) and `the console says the bind is still up` — with the stubbed `rm` REMOVED first so the real `rm` would have emptied the directory. The third new row (`umount fails (busy) -> the verb still completes`, rc 0) survives that mutant by design: the verb completes either way; it is the positive leg the reviewer asked for beside the absence row, and the comment above the errexit runner now says so.

**Budget:** `tests/stack/test-appliance-rig-miner.sh` is at exactly 400 lines — the bar is `TARGET_LINES=400` in `scripts/lint-file-budget.sh` (a row IFF over it), which is also the sourcing for the `verify-image.sh` sentence above. To stay there the `mountpoint`/`umount` stub heredocs became one-line `printf`s; `lint-file-budget` OK, no row. `lint-pithead-parity` OK; `shfmt -i 4 -d` clean.

**Rebase:** onto the pull 1818 merge (it regenerates `pithead` too); `scripts/build-pithead.sh` re-run; the patch outside `pithead` was byte-identical to the pre-rebase patch (confirmed independently by the `fixes` lane against each side's own merge-base).

**Not done:** no full-suite local rerun after finding 2 (fragment only); CI at the current head is the whole-suite proof. Merge is non-author — `fixes` holds it on steps 5–11 `success` plus a reviewer PASS naming the current head.

### DIFF
--- a/lib/pithead/14-local-miner.sh
+++ b/lib/pithead/14-local-miner.sh
@@ -326,8 +326,22 @@ rig_minimize_writes() {
     # overrides a drop-in.
     printf '[Journal]\nStorage=volatile\nRuntimeMaxUse=32M\n' >"$dropin/zz-rig-volatile.conf" 2>/dev/null || return 0
     [ -d "$journal" ] || return 0
-    rm -rf "${journal:?}"
+    # Restart FIRST: journald holds the persistent files under $journal open, and a bind there
+    # cannot come off while it does (umount: target is busy). Volatile, it writes under /run and
+    # holds nothing here. (Before #1817 the restart came after the reclaim, which worked only
+    # because unlinking an open file is allowed — taking a mount off is not.)
     systemctl restart systemd-journald >/dev/null 2>&1 || true
+    # #1817: a first boot binds the persistent journal home onto this path before the role is
+    # known (pithead-journal-persist), so the reclaim meets a MOUNTPOINT. `rm -rf` on one empties
+    # it — through the bind, onto /data — and then fails on the mountpoint itself. Take the bind
+    # off first, so what is reclaimed is the overlay's own directory and nothing on /data is
+    # touched. Every step is best-effort: this is a stick-wear optimisation, nothing downstream
+    # reads its result, and the one caller that matters runs under errexit (pithead-boot's
+    # `local-miner`) — a cleanup that cannot complete must never leave a slot uncommitted.
+    if mountpoint -q "$journal" 2>/dev/null; then
+        umount "$journal" 2>/dev/null || true
+    fi
+    rm -rf "${journal:?}" || warn "The journal directory could not be reclaimed; journald is volatile for this boot anyway."
     log "Rig write minimization: the journal is in memory for this boot — the root may be the stick the miner runs from."
 }
 
@@ -336,7 +350,11 @@ provision_rig_miner() {
         warn "This machine is marked as a rig, but its settings are missing — install it again from the stick to choose a role."
         return 1
     fi
-    rig_minimize_writes
+    # `|| true` on purpose (#1817): the minimization is best-effort by construction, and this
+    # function is called BARE from the boot path (pithead-boot -> `pithead local-miner`, errexit
+    # armed) but under `|| true` from the wizard — a failure inside it passed first boot and
+    # killed every boot after, which is the one signature a full gate is needed to see.
+    rig_minimize_writes || true
     render_rig_miner_config || return 1
     if rigforge_setup_run; then
         log "The rig is mining: $(jq -r '.worker // "this machine"' "$PWD/rig.json" 2>/dev/null) -> $(jq -r '.pool // "no pool recorded"' "$PWD/rig.json" 2>/dev/null)."
--- a/lib/pithead/14-local-miner.sh
+++ b/lib/pithead/14-local-miner.sh
@@ -340,6 +340,12 @@ rig_minimize_writes() {
     # `local-miner`) — a cleanup that cannot complete must never leave a slot uncommitted.
     if mountpoint -q "$journal" 2>/dev/null; then
         umount "$journal" 2>/dev/null || true
+        # A bind that did not come off is left alone: `rm -rf` through it would empty the
+        # persistent home on /data, the one thing this block promises not to do.
+        if mountpoint -q "$journal" 2>/dev/null; then
+            warn "The journal bind is still up, so nothing is reclaimed; journald is volatile for this boot anyway."
+            return 0
+        fi
     fi
     rm -rf "${journal:?}" || warn "The journal directory could not be reclaimed; journald is volatile for this boot anyway."
     log "Rig write minimization: the journal is in memory for this boot — the root may be the stick the miner runs from."
--- a/os/overlay/pithead-journal-persist
+++ b/os/overlay/pithead-journal-persist
@@ -18,6 +18,20 @@ set -eu
 
 DATA_DIR="${PITHEAD_JOURNAL_DATA_DIR:-/data/pithead/journal}"
 TARGET="${PITHEAD_JOURNAL_TARGET:-/var/log/journal}"
+ROLE_FILE="${PITHEAD_MACHINE_ROLE_FILE:-/data/pithead/machine-role}"
+
+# A rig has nothing to persist (#1817): its journal is volatile BY DESIGN — the root may be
+# the stick it mines from — and `pithead local-miner` reclaims /var/log/journal on every boot
+# on its way to flipping journald volatile. With the persistent home bound onto that path the
+# reclaim met a mountpoint, `rm -rf` failed on it, and pithead's errexit turned the failure into
+# a boot verb that exited non-zero: the rig booted mining and never committed its slot. Same
+# marker as pithead-boot's role fork, read the same way; RequiresMountsFor=/data means it is
+# readable by now. A machine with no marker yet (first boot) still binds — the wizard's rig leg
+# takes that bind off itself before reclaiming (rig_minimize_writes).
+if [ -f "$ROLE_FILE" ] && [ "$(tr -d '[:space:]' <"$ROLE_FILE")" = "rig" ]; then
+    echo "pithead-journal-persist: this machine is a rig — its journal is volatile, nothing to bind"
+    exit 0
+fi
 
 mkdir -p "$DATA_DIR"
 # Same owner+mode systemd-tmpfiles gives /var/log/journal itself (root:systemd-journal, setgid
--- a/pithead
+++ b/pithead
@@ -3581,6 +3581,12 @@ rig_minimize_writes() {
     # `local-miner`) — a cleanup that cannot complete must never leave a slot uncommitted.
     if mountpoint -q "$journal" 2>/dev/null; then
         umount "$journal" 2>/dev/null || true
+        # A bind that did not come off is left alone: `rm -rf` through it would empty the
+        # persistent home on /data, the one thing this block promises not to do.
+        if mountpoint -q "$journal" 2>/dev/null; then
+            warn "The journal bind is still up, so nothing is reclaimed; journald is volatile for this boot anyway."
+            return 0
+        fi
     fi
     rm -rf "${journal:?}" || warn "The journal directory could not be reclaimed; journald is volatile for this boot anyway."
     log "Rig write minimization: the journal is in memory for this boot — the root may be the stick the miner runs from."
--- a/pithead
+++ b/pithead
@@ -3567,8 +3567,22 @@ rig_minimize_writes() {
     # overrides a drop-in.
     printf '[Journal]\nStorage=volatile\nRuntimeMaxUse=32M\n' >"$dropin/zz-rig-volatile.conf" 2>/dev/null || return 0
     [ -d "$journal" ] || return 0
-    rm -rf "${journal:?}"
+    # Restart FIRST: journald holds the persistent files under $journal open, and a bind there
+    # cannot come off while it does (umount: target is busy). Volatile, it writes under /run and
+    # holds nothing here. (Before #1817 the restart came after the reclaim, which worked only
+    # because unlinking an open file is allowed — taking a mount off is not.)
     systemctl restart systemd-journald >/dev/null 2>&1 || true
+    # #1817: a first boot binds the persistent journal home onto this path before the role is
+    # known (pithead-journal-persist), so the reclaim meets a MOUNTPOINT. `rm -rf` on one empties
+    # it — through the bind, onto /data — and then fails on the mountpoint itself. Take the bind
+    # off first, so what is reclaimed is the overlay's own directory and nothing on /data is
+    # touched. Every step is best-effort: this is a stick-wear optimisation, nothing downstream
+    # reads its result, and the one caller that matters runs under errexit (pithead-boot's
+    # `local-miner`) — a cleanup that cannot complete must never leave a slot uncommitted.
+    if mountpoint -q "$journal" 2>/dev/null; then
+        umount "$journal" 2>/dev/null || true
+    fi
+    rm -rf "${journal:?}" || warn "The journal directory could not be reclaimed; journald is volatile for this boot anyway."
     log "Rig write minimization: the journal is in memory for this boot — the root may be the stick the miner runs from."
 }
 
@@ -3577,7 +3591,11 @@ provision_rig_miner() {
         warn "This machine is marked as a rig, but its settings are missing — install it again from the stick to choose a role."
         return 1
     fi
-    rig_minimize_writes
+    # `|| true` on purpose (#1817): the minimization is best-effort by construction, and this
+    # function is called BARE from the boot path (pithead-boot -> `pithead local-miner`, errexit
+    # armed) but under `|| true` from the wizard — a failure inside it passed first boot and
+    # killed every boot after, which is the one signature a full gate is needed to see.
+    rig_minimize_writes || true
     render_rig_miner_config || return 1
     if rigforge_setup_run; then
         log "The rig is mining: $(jq -r '.worker // "this machine"' "$PWD/rig.json" 2>/dev/null) -> $(jq -r '.pool // "no pool recorded"' "$PWD/rig.json" 2>/dev/null)."
--- a/tests/stack/test-appliance-rig-miner.sh
+++ b/tests/stack/test-appliance-rig-miner.sh
@@ -172,6 +172,66 @@ assert_rc "re-run (the leg fires every boot) -> rc 0" "$?" "0"
 assert_eq "the config is re-derived, not repaired" "$(jq -r '.pools[0].url' "$RIGL/rigforge/config.json")" "pithead.local:3333"
 assert_eq "the stratum password lands as the pool pass" "$(jq -r '.pools[0].pass' "$RIGL/rigforge/config.json")" "s3cret"
 assert_not_contains "already volatile -> journald is not restarted again" "$(cat "$RF_LOG")" "restart systemd-journald"
+# #1817: a first boot binds the persistent journal home onto this path before the role is
+# known (pithead-journal-persist), so the reclaim meets a MOUNTPOINT. The bind comes off
+# first — after journald has let go of the files under it, which is why the restart moves
+# ahead of the umount — and only then is the directory reclaimed.
+cat >"$RIGL/bin/mountpoint" <<EOF
+#!/usr/bin/env bash
+[ -f "$RIGL/bound" ] && exit 0
+exit 1
+EOF
+cat >"$RIGL/bin/umount" <<EOF
+#!/usr/bin/env bash
+echo "umount:\$*" >>"\${RF_LOG:?}"
+rm -f "$RIGL/bound"
+EOF
+chmod +x "$RIGL/bin/mountpoint" "$RIGL/bin/umount"
+mkdir -p "$RIGL/journal/abc"
+touch "$RIGL/bound"
+: >"$RF_LOG"
+rigl_out=$(run_rig provision_local_miner 2>&1)
+assert_rc "a bound journal path -> rc 0" "$?" "0"
+assert_eq "journald lets go first, then the bind comes off, then the miner is set up" \
+    "$(grep -o 'systemctl:restart systemd-journald\|umount:[^ ]*\|rigforge:setup' "$RF_LOG" | tr '\n' ' ')" \
+    "systemctl:restart systemd-journald umount:$RIGL/journal rigforge:setup "
+[ -d "$RIGL/journal" ] && bad "the directory under the bind is reclaimed" "still there" ||
+    ok "the directory under the bind is reclaimed"
+# A reclaim that cannot complete must not take the boot verb down: pithead-boot calls
+# `pithead local-miner` BARE under the CLI's errexit, while the wizard calls the same leg under
+# `|| true` — a failure inside the minimization passed first boot and killed every boot after
+# (the #1651 gate, BUILD_COMMIT 4fb4943a: the rig booted mining and never committed its slot).
+# Driven under errexit for real, with a control that the runner is armed; rm is stubbed to
+# refuse the journal path with the exact text the mountpoint case prints.
+run_rig_errexit() { (
+    cd "$RIGL" || exit 99
+    export PITHEAD_APPLIANCE=1 PATH="$RIGL/bin:$PATH"
+    source "$STACK"
+    "$@"
+); }
+_errexit_probe() {
+    false
+    echo survived
+}
+assert_not_contains "control: the errexit runner is armed" "$(run_rig_errexit _errexit_probe 2>/dev/null)" "survived"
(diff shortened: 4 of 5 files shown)


VijitSingh97:
## CI red at `97c41b4d` was lint, fixed at `627d4db8` — appliance lane (author)

**Cause, re-derived from the job log via the API:** `Shell tests` failed in step 4 (`make lint-sh`) on SC1090 at the new `source "$STACK"` inside `run_rig_errexit()` in `tests/stack/test-appliance-rig-miner.sh` — the only non-constant source in the file and a `+` line in this diff, which is why lint was green before this branch. Steps 5–11 (the pithead suite and its siblings, the inventory drift check, the integration self-test) were **SKIPPED**, so CI at `97c41b4d` proved nothing about the suite; the tier-1 figures in the body were local-only until this rerun.

**Fix:** the line-above `# shellcheck disable=SC1090` directive, the house shape for a source inside a function (`test-appliance-boot.sh`, `test-appliance-boot-remint.sh`), not a top-of-file directive (its scope shrinks to one line when a command lands above it).

**Run, not intended:** shellcheck through the box's serialized wrapper fired on the previous text (SC1090, rc 1) and passes on the new one (rc 0). `shfmt -i 4 -d` (the pinned 3.13.1) is clean on the three changed shell files. `make lint-file-budget` OK: the file is at 400 lines, at the bar, no row. Product paths (`os/`, `lib/pithead/`, `pithead`) are untouched by this commit.

**Not done:** I did not re-run the local suite for a one-line comment; the CI rerun at `627d4db8` is the proof that steps 5–11 execute. Reviewer PASS at `97c41b4d` does not carry — the verdict has to name `627d4db8`. Merge is non-author (`fixes` or the seat), never mine.

Finding credited to the `fixes` lane, re-derived here at source before acting.

VijitSingh97:
## VERDICT: PASS at `bdd2e63daaa3fa29cd782ae9e0a43eaaf1086815`

Non-author pass, ephemeral reviewer. **The design is right, both halves hold up under re-derivation,
and the one finding I raised has been taken and covered.** Everything below is DERIVED FROM SOURCE
unless marked otherwise: the appliance lane holds the bench claim, so this pass ran no suite, no
`make`, no shellcheck/shfmt, no container and no KVM. The things I did execute are named as such.

### ⛔ First, a retraction of my own — the 400-line bar is REAL and I said it was not

An earlier draft of this pass (written to the fleet answers file the author read) said the body's
*"`verify-image.sh` sits at its 400-line bar"* was unsourced, because I found no row for it in
`docs/dev/file-budget.tsv`. **That was wrong, and the cause was a sweep too narrow at the root:** I
grepped the budget TSV and the file itself, never the gate script. `scripts/lint-file-budget.sh:26` sets
a global `TARGET_LINES=400`, and `:180-183` FAILS any file over it that has no budget row. So
`tests/os/verify-image.sh` at exactly 400 and `tests/stack/test-appliance-rig-miner.sh` at exactly 400
are both **at the bar with zero headroom**, and the body was accurate. Retracted here rather than
quietly dropped.

It also reframes something I would otherwise have read as churn: the two stub heredocs becoming
one-line `printf`s in the latest commit is not tidying, it is **paying for the 8 new assertion lines**
to hold the file at 400. Disclosed in the commit message, correctly.

**Systemic, for whoever owns budgets — not this PR's problem to fix:** `test-appliance-rig-miner.sh`
now sits at 400/400, which taxes every future PR touching it, in the same shape `tests/stack/run.sh`
already does at 440/440. The next change to this file pays in compressed test prose or needs the
ceiling raised deliberately.

### Head movement — three moves, each re-derived rather than relayed

`97c41b4d` → `627d4db8`: one insertion, in a test file — the `# shellcheck disable=SC1090` for
`run_rig_errexit`'s dynamic source. No product path moved.

`627d4db8` → `2ac7afec`: a rebase over #1818, so the BASE moved too and trees prove nothing — patches
do. Measured, both legs mine: `git diff <merge-base> <head> -- . ':!pithead'` is **byte-identical across
the rebase** (10,633 bytes each side, `diff -q` clean), and the `pithead` diff against the new base is
two hunks whose added/removed lines match the `lib/pithead/14-local-miner.sh` delta **line-for-line**.

`2ac7afec` → `bdd2e63d`: the fix for my finding, plus its rows. Re-audited below rather than carried.

**⛔ And a correction to how I read CI, which matters for what "green" means here.** There are TWO
workflows per head: `OS rootfs scan` and `CI`. `Shell tests` lives in `CI`. At `627d4db8` the `OS
rootfs scan` run succeeded while the `CI` run was **cancelled** — so a rollup glance there would have
read as a pass on a head whose suite never ran. Both workflows at `2ac7afec` were cancelled too, each
superseded by the next push. **A cancellation is the ABSENCE of a measurement**, so as of this head the
pithead suite has still never completed on any pushed revision of this branch — which is why this
verdict waited for `CI` at `bdd2e63d` rather than being posted three heads ago.

### CI at this head — GREEN, and the new rows demonstrably RAN

Read from the job log, not the label, because a green job is equally consistent with a new row never
being collected. Job `101267685106`, both workflows `success` (`CI` and `OS rootfs scan`), all 18
check-runs `success` or `skipped` (the two skipped are the schedule-gated CVE sweep rows).

`pithead tests: 3356 passed, 0 failed`; `firstboot-journal tests: 17 passed, 0 failed`.

Every row this PR adds or changes appears in the log with a tick — I grepped for the labels rather
than trusting the totals:

```
✓ journald lets go first, then the bind comes off, then the miner is set up
✓ control: the errexit runner is armed
✓ a reclaim that cannot complete does not take the boot verb down (errexit armed)
✓ umount fails (busy) -> the verb still completes
✓ a bind still up is not reclaimed through
✓ the console says the bind is still up
✓ rig marker -> mount is never called
✓ marker 'rigs' still binds
✓ no marker yet (first boot) still binds
```

**Do not read 3356 against the body's local 3352 as a +4 delta** — that figure was taken at the first
head on the older base, so the two counts span three new assertion rows AND whatever the rebase
brought in with #1818. They are not the same population; the totals above are this head's.

---

### What I was asked to settle

**(1) No unguarded command is reintroduced on the errexit path — CONFIRMED, by enumeration at this head.**

Chain re-derived at source rather than relayed: `os/overlay/pithead-boot:304` forks on the marker, `:308`
is `timeout 1800 ./pithead local-miner || fail_boot` → `provision_local_miner`
(`lib/pithead/14-local-miner.sh:252`) → `provision_rig_miner` at `:254`, **bare** → `rig_minimize_writes`
at `:363`. Every statement in `rig_minimize_writes` (`:320-352`) is guarded:
`is_appliance || return 0` (`:321`), `mkdir -p … || return 0` (`:324`), `printf … || return 0` (`:327`),
`[ -d "$journal" ] || return 0` (`:328`), `systemctl … || true`, `if mountpoint -q …` (`:341`,
condition context), `umount … || true`, the new inner `if mountpoint -q …` (`:345`, condition context)
with `warn` + explicit `return 0` (`:346-347`), `rm -rf … || warn` (`:350`), and `log` last.

The one that could have been a trap is `|| warn`: `warn()` is `pithead:32`, a bare `echo … >&2`, rc 0 —
so `rm … || warn` cannot return non-zero, and neither can the new `warn; return 0` arm. **Nothing new is
unguarded, at this head or the previous two.**

The call-site `|| true` at `:363` then sits on top of all of it, **which is why reading M1 as honest is
correct**: with every step guarded there is nothing left for the call site to catch, so no row can
distinguish it. Recording that in the table rather than hiding it is the right call, and it stays true
after the new commit.

Call-site map, re-derived: `provision_rig_miner` has exactly three callers —
`lib/pithead/12-firstboot-wizard.sh:104` and `:349` under `|| true`, and `14-local-miner.sh:254` bare.
`rig_minimize_writes` has exactly one. The first-boot/reboot split really is the call site, not
ordering — so the reorder in this PR is a correctness fix for the *umount*, and is not doing the work of
fixing the split. Both statements in the body survive.

**(2) M6's control row genuinely fires — CONFIRMED, derived (I did not run it).**

`pithead:14` is `set -Eeuo pipefail`, so `source "$STACK"` arms errexit in the sourcing shell.
`run_sourced` (`tests/stack/lib.sh:79-88`) deliberately does `set +e` *after* the source — that is what
makes the ordinary runner unarmed, and omitting it is the only difference in `run_rig_errexit`
(`tests/stack/test-appliance-rig-miner.sh:198-204`). `_errexit_probe` (`:205`) is `false; echo survived`,
invoked as `"$@"` — a simple command, not a condition — so armed, the subshell dies before the echo and
the capture is empty; disarmed (M6), `survived` is printed and `assert_not_contains` at `:209` reddens.
`inherit_errexit` does not apply: the `set -e` is taken *inside* the substitution's own subshell rather
than inherited into it.

**The caveat I raised on it has been taken.** `:209` is an ABSENCE row — alone it also passes if the
runner simply died for an unrelated reason, so it cannot by itself distinguish "errexit fired" from
"the runner is broken". What rescues it is the pair: `:221-224` drives the SAME runner to completion
with rc 0 and two positive strings. The latest commit says so in the comment at `:196-197` ("The
armed-runner control is an ABSENCE row, paired with the positive rows after it"), which is the right
outcome — the next person to move one of those rows now sees why they belong together.

### The rest of the mutation table maps to real assertions

Not asked for, but M1 and M6 are only honest if the others are. **M2** (restart moved after the reclaim)
and **M3** (umount block dropped) both perturb the ordered `grep -o` at `:187-189`, and the body's note
that they share one instrument is accurate — one axis, not two. **M4** is the precise one: with `|| warn`
dropped but the call site still `|| true`, errexit stays suppressed across the whole subtree, so rc stays
0 and setup still runs — only the console row can move, exactly as claimed. **M7** drops both, so the
`rm` failure reaches a bare call site and the rc row moves, which is what gives the rc-0 assertion its own
firing control. Internally consistent; every row names an assertion that exists.

### The new commit (my finding) — re-audited, and its control is real

`lib/pithead/14-local-miner.sh:341-348` now re-tests `mountpoint -q "$journal"` after the umount and
returns 0 with a warning when the bind is still up, instead of falling through to an `rm -rf` that would
empty the persistent home through it. That is the fix, and the comment at `:342-343` now states the
property rather than only the intent.

**The row that covers it can fail, which is the part worth checking** (`:225-232`): the `umount` stub is
rewritten to `exit 32` *without* clearing the `bound` marker, so the second `mountpoint -q` really does
see a live bind; and — the load-bearing detail — `rm -f "$RIGL/bin/rm"` puts the **real** `rm` back on
PATH, so a reclaim that reached it would genuinely empty the directory and `[ -d "$RIGL/journal/abc" ]`
at `:231` would fail. That is a firing control by construction, not an assertion that happens to be
true. The stub rewrite via `>` also preserves the exec bit set earlier, so it stays executable.

It does not shadow the two existing rows: in both of those the umount stub still clears `bound`, so the
second `mountpoint -q` is false and control falls through exactly as before.

### Checks that could have failed and did not

- **Artifact parity — executed by me, at each of the three heads** (a pure concatenation, no linters, so
  inside the CI-only rule): `bash scripts/build-pithead.sh --check` → **rc 0**, *"exactly what 53
  slice(s) build"*. The committed `pithead` is the `lib/` source byte for byte.
- **The base has not drifted — executed by me:** `merge-base` equals the `develop-v2` tip, so nothing
  here is measured against a stale base.
- **The marker read really is `pithead-boot`'s.** `os/overlay/pithead-journal-persist:31` is
  `[ -f "$ROLE_FILE" ] && [ "$(tr -d '[:space:]' <"$ROLE_FILE")" = "rig" ]`; `pithead-boot:304` is the
  same shape against `machine-role`, with `cd /data/pithead` at `:289` and a matching
  `WorkingDirectory=`. Same file, same read, same word.
- **Ordering.** `pithead-journal-persist.service` is `Before=sysinit.target` with
  `RequiresMountsFor=/data`; `pithead-boot.service` is `WantedBy=multi-user.target`. The stand-down
  decision is made, with the marker readable, before the reclaim ever runs. No cycle added.
- **No os-tier row is broken by the stand-down** — the regression I went looking for, since a rig that no
  longer binds could falsify an existing #1791 assertion. `journal_home_verdict` is the one check that
  *requires* the path to BE the bind, and it has a single call site, `tests/os/run.sh:535`, inside
  `phase_boot()` — a coordinator, which still binds. The `verify-image.sh` journal rows are all static
  (script baked, chown text, unit enabled, ordering) and none is touched.
- **The idiom matches the shipped precedent.** `os/overlay/pithead-data-reset:215` is the identical
  `if mountpoint -q … then umount … || true; fi`, and `mountpoint` is already used by the very overlay
  this PR edits — so the binary is in the image and this adds no new dependency.
- **Fork narrowness.** `tests/stack/test_firstboot_journal.sh:94` widens the control to a sibling AND a
  near-miss (`pithead`, `rigs`), plus the no-marker first-boot case at `:105`. A marker-value fork proved
  only on its positive is a fork proved on nothing; this is not that.
- **`both` and `pithead` roles.** Both still bind, and neither reaches `rig_minimize_writes` —
  `14-local-miner.sh:252` gates on the exact word `rig`.
- **Budget.** Re-checked against the gate script this time, not just the TSV: no touched file exceeds
  `TARGET_LINES`, and none carries a row it would now contradict.

### Not re-derived, explicitly

- **The on-image experiment table is yours; I did not reproduce it.** It is consistent with everything I
  read, and its first control row (`umount` rc 32, target busy) is what made my finding reachable rather
  than theoretical.
- **The M6 arming, the mutation table and the whole test block are reasoned from source, never executed.**
  If you relay this, say "derived", not "measured". The CI suite is the measurement; this pass is not.
- **Residual, not a finding, not blocking:** if `mountpoint` were ever absent the outer `if` is false, the
  whole guarded block is skipped, and a live bind would still be reclaimed through. It is in the image
  today (two shipped overlay scripts use it), so this is a note for whoever ever trims the rootfs, not a
  change to make here.
- **Minor, not a finding:** on a rig, whatever the first boot flushed into the persistent journal home
  stays there for the life of the machine — the stand-down means nothing reclaims it on any later boot.

---

**PASS.** The defect is correctly diagnosed, the fix matches the diagnosis at both halves, the
errexit path is clean by enumeration, the mutation table is honest about what it cannot prove, and the
one finding I raised was taken with a control that can fail. My only unresolved item is a retraction of
my own, recorded above.

**This is a freeze-path PR, so it is MERGE-READY for the appliance lane's sequencing — not for me to
merge, and not for the author.** `Closes` is inert on `develop-v2`: hand-close #1817 with the merge sha.

Recorded by an ephemeral reviewer; the full pass is also written to the fleet answers directory under
this timestamp, so a lane that cannot see this comment can still read the verdict at source.

Issue #1811, opened 2026-09-05
worker_config_revision grows one permanent row per device-chosen worker name, with no cap and no prune


VijitSingh97:
## The defect

`worker_config_store.note_worker_revision` writes one `INSERT OR REPLACE` row into
`worker_config_revision`, keyed on `worker`, for **every** worker it sees — including a first
sighting, which is the case it returns `None` for. `worker` is the name the device presents on
the unauthenticated enriched worker feed, and `_normalize_proxy_workers` copies that name through
with no validation.

So a device that varies the name it presents adds one permanent row per name. Nothing bounds it:

- The write happens BEFORE the #724 flood cap. `note_revision_drift` calls
  `note_worker_revision` first and only consults `_rig_edit_within_cap` if that returns a drift,
  so a first sighting spends no cap budget and still writes.
- `#1695`'s ceiling does not cover it either. That bounds how many names may hold a live
  flood-cap window, which bounds the `audit_events` rows and the in-memory window map — it does
  not sit in front of this write.
- No retention prune targets this table. `DELETE FROM` appears for `history`, `shares`,
  `xvb_history` and `network_history`; `worker_config_revision` has none.

## Why it is a different defect class from #1695

#1695 is a rate bound on a device-chosen bucket key, and its fix is an in-memory admission
ceiling. This is durable table growth with no cap and no retention on a table whose primary key
is a string the device picks. The fix is a retention or admission policy on the store itself, at
a different layer, with a different blast radius: this table backs the Inspect provenance line
(`get_worker_revision_drift`, #1564), so evicting a row silently drops a real rig's drift
history, which the audit-trail case does not.

## Directions worth weighing, not a prescription

- A row cap with LRU-by-`ts` eviction — the column is already written.
- Admit a name to this table only once it is also admitted to the #724 window map, reusing
  #1695's ceiling as the one gate on the device-chosen name space.
- A retention prune on `ts`, matching the shape the other four tables already use.

The second is the one that keeps a single bound rather than two that can drift apart, but it
changes what a first sighting means, so it wants deciding rather than assuming.

## What I checked, and what I did not

PROVEN HERE, read at source: the `INSERT OR REPLACE` runs before the `row is None` first-sighting
return; the cap is consulted only after `note_worker_revision` reports a drift; no `DELETE`
statement in the package names this table (checked with a control that the same sweep does find
the four tables that are pruned).

NOT DONE: no measurement of the per-row cost or of how large the table gets in practice, and no
[...]

PR #1820, opened 2026-09-05
fix(dashboard): #1769 an appliance operator is not shown host remedies it cannot reach

### LINKED ISSUE #1769: A failed dashboard save shows the appliance operator raw host CLI remedies it cannot reach

VijitSingh97:
Closes #1769.

## The defect, re-derived at source

Every link read in the code, none observed in a running UI (see "What I did not do"):

1. `lib/pithead/43-control-approval-and-preview.sh:364` writes `tail -c 2000` of the raw `apply` log into the failed result's `error`, alongside `backup`.
2. `control_service.wait_result` returns that dict untouched (`return res`).
3. `web/server.py:196` spreads it to the browser: `web.json_response({"id": rid, **res})`.
4. `configview.mjs` rendered `result.error` bare in a `<pre>`, under a sentence naming `config.json.bak-control` "on the host".

So an appliance operator reads `Run './pithead setup' first.` and `Fix the cause shown above, then re-run './pithead apply'` (both verbatim from `40-apply-and-render.sh:131` and `:289`) with no shell to run them in, plus a host path they cannot reach. #1213 settled the rule for doctor's messages; its sweep is keyed to `dr_*` and this text never passes through `dr_*`, so that sweep is structurally blind to this surface and always will be.

## The fix

The wording branches on `state.os_update` — the appliance tell the client already uses (`osupdate.mjs`, `components.mjs`). On an appliance:

- the backup sentence says the previous configuration is kept **without naming a host path**;
- the log tail is **captioned** as the machine's own apply log, diagnostic detail rather than instructions;
- the caption names only surfaces that page actually has — the form above, and the Service diagnostics card rendered beside it in the same card-stack.

**The tail is kept on both branches.** #1769 is explicit that this is a wording problem, not a redaction one, and the tail is the only diagnostic detail either operator gets. A test asserts the tail — `./pithead setup` and `./pithead apply` included — survives on both branches, so a later "fix" that scrubs the log reds instead of passing as an improvement.

**The host branch is byte-unchanged.** A shelled operator still gets the actionable path.

### Why the client, and not a filter in `server.py`

Because a server-side filter at `:196` would miss the common case. `handle_control_commit` waits `CONTROL_WAIT_S` (30s) inline; a failing `apply` is a full container recreate and routinely runs longer, so it 202s and the client polls `/api/control/result` — `server.py:441`, a different handler that returns the whole dict. `configview.mjs` `commit()` -> `poll()` -> `setState({phase: "done", result})` puts both routes through the *same* render branch. Fixing the render covers both; fixing `:196` would have covered only the fast failure and left the slow one — the likelier one — showing the old text.

## Why new files rather than edits

`configview.mjs` (591) and `configview.test.mjs` (492) both sit at their recorded ceilings with **zero** headroom. `applyfailure.mjs` (67) and `applyfailure.test.mjs` (147) are new, both under the 400 target, so **neither takes a budget row** and `file-budget.tsv` is untouched. `configview.mjs` ends at 590/591 and `components.mjs` at 1137/1177. This is the split `osverdict.mjs` took out of `osupdate.mjs`, for the same stated reason; the new test file reuses the sibling `helpers/render.mjs` probe rather than duplicating one.

## Over-engineering pass (done on merits — the local gate verifies nothing)

Two real findings, both fixed in this branch before it was pushed:

- **A.** My first draft asserted `doesNotMatch(/ssh|console|terminal/i)` on the appliance caption. `console access` is #1213's **own approved** appliance remedy, used at eight sites in `06-doctor.sh` for exactly the case where no dashboard surface exists. The assertion would have redded a correct future edit. Removed.
- **B.** The caption names another component's card by its heading, and cross-file prose goes stale with nothing red. Added a test pinning the caption's "Service diagnostics" against the `<h3>` `DiagnosticsPanel` actually renders (`diagview.mjs:145`).

## Known residual, recorded in the module

`state.os_update` is a signal about a **host-written file**, not about the hardware. It is the only appliance tell `build_state` carries — checked against the whole payload rather than assumed — but an appliance whose host has not written `os-update-state.json` yet reads as a non-appliance and gets the host wording, so #1769 survives in that window. Same fail-direction as the header's OS-update control, which simply does not render in it. Closing it properly needs an explicit appliance flag in `build_state`: a bigger change than this bug, and not one to start during a freeze.

## Scope this does NOT cover

- **The host half.** `lib/pithead/43-control-approval-and-preview.sh` still emits a raw tail, and `40-apply-and-render.sh` still writes CLI imperatives into it. Both are under the image freeze and are not this lane's paths, so this fixes what the operator reads, not what the host writes.
- **RETRACTED, and corrected here rather than quietly rewritten.** This body first said the other two `**res` passthroughs were "`server.py:173` preview, `:364` os-update" and carried the same raw-tail shape. **`:364` is `handle_worker_apply`, not os-update, and it carries no raw tail at all** — `46-control-worker-ops.sh` has zero `tail -c` calls; its errors are messages. Re-derived after opening this PR. The class is real but its members are these:
  - Raw `tail -c 2000` writers: `43-...:284` (preview), `43-...:364` (commit — the one fixed here), `44-...:388` (upgrade/lifecycle), `45-...:76` (backup).
  - They reach the browser via `server.py:173` (preview), `:196` (commit), and — the one I had missed — **`handle_control_result` at `server.py:441`, `web.json_response(res)`, which returns the whole dict for anything that 202'd.** That is how the upgrade and backup tails get out.
  - Rendered by `configview.mjs`'s UpgradeControl (which also names a host backup path) and `backupview.mjs:124`.

  I did **not** widen this PR to cover them: it is mid-CI with a recorded shape, each surface needs its own honest wording, and the window is a wind-down. Flagging it for the reviewer to rule on rather than deciding it myself — happy to file it as a follow-up issue.
- **The App -> ConfigView prop hop.** `appliance=${!!state.os_update}` at `components.mjs:1044` is verified by reading, not by a test: ConfigView's first paint through App is always its loading card, so the render probe cannot reach the failed state that way. A test pins the failure mode instead — dropping the prop degrades to the host wording, which is #1769 returning with nothing red.

**Nothing was run locally.** This lane is under a CI-only rule for the duration of the #1651 release-gate battery: no `make test*`, no `make lint*`, no node runs. So `node --test`, `biome check` and `lint-file-budget` are unverified by me and CI is the first gate they meet. Two consequences I handled statically instead, and which a reviewer should still check:

- `applyFailure` returns a plain **array** of children rather than one `html` template, because every root would have been an interpolation with no literal tag for htm to anchor on. `renderToString` handles arrays explicitly and Preact takes array children, so this removes the one shape I could not settle without running it.
- The `applyfailure.mjs` import was moved above the `configlogic.mjs` block to keep the file's alphabetical order, because `make lint-js` runs `biome check .`, which includes the organize-imports assist. Max line length 99 against biome's 100.

`docs/dashboard.md` is updated in the same commit (shared `docs/` path, disclosed per lane policy). `docs/operations.md:270` was checked and left alone: it describes what the host keeps, which is still true.

**Not MERGE-READY.** Needs a recorded non-author pass; I am the author and do not merge my own. Note for whoever sequences it: `dashboard/` is image bytes, so merging moves the tip the next gate builds from — this is ready when the seat wants it, not necessarily now.

https://claude.ai/code/session_01WcHCQa59LmHJg75K5reeaM

### DIFF
--- a/dashboard/mining_dashboard/web/static/applyfailure.mjs
+++ b/dashboard/mining_dashboard/web/static/applyfailure.mjs
@@ -1,0 +1,67 @@
+// What a failed apply tells the operator (#1769).
+//
+// Its own module rather than more of configview.mjs: that file sits at its recorded budget ceiling
+// with no headroom, and this is a self-contained concern — one host result in, display text out,
+// with no fetching, sequencing or state of its own. The same split osverdict.mjs took out of
+// osupdate.mjs.
+//
+// The defect: the host writes `tail -c 2000` of the raw `apply` log into the result's `error`
+// field (43-control-approval-and-preview.sh), server.py returns the result dict to the browser
+// unfiltered, and this card rendered that tail bare. An apply log names host CLI verbs — "Run
+// './pithead setup' first", "then re-run './pithead apply'" — and an appliance operator has no
+// shell to run them in. #1213 settled the rule for doctor's messages: an appliance is told what it
+// can actually reach, and the wording NEVER invents a remedy. This is that rule on a surface
+// #1213's sweep cannot see, because the text never passes through dr_* at all.
+//
+// The tail is KEPT on both branches. It is real diagnostic detail and dropping it would trade one
+// bad outcome for another. What changes on an appliance is that it is labelled as the machine's
+// own log instead of being left to read as instructions, and that the sentences around it name
+// only surfaces on this page: the form above, and the Service diagnostics card beside it.
+//
+// The host branch is unchanged, deliberately. A host operator has a shell, so
+// `config.json.bak-control` is a path they can act on, and naming it is the useful thing to do.
+//
+// KNOWN RESIDUAL. The caller's signal is `state.os_update`, which is the presence of a file only
+// an appliance's host writes (config.py OS_UPDATE_STATE_PATH). It is the only appliance tell
+// build_state carries — checked against the whole payload, not assumed — but it is a signal about
+// that FILE, not about the hardware. An appliance whose host has not written it yet reads as a
+// non-appliance and gets the host wording, so #1769 survives in that window. Same fail-direction
+// as the header's OS-update control, which simply does not render in it. Closing it properly
+// needs an explicit appliance flag in build_state, which is a bigger change than this bug.
+import { html } from "./preact.mjs";
+
+// The backup path is a host filesystem path. It is the right thing to show someone with a shell
+// and a dead end for someone without one, so it is the sentence that has to branch.
+function backupLine(result, appliance) {
+  if (appliance) {
+    return html`<p class="status-bad">Apply failed. Your change was saved but is not running, and
+        the previous configuration is kept on this machine.</p>`;
+  }
+  return html`<p class="status-bad">Apply failed. The previous config is kept at
+      <code>${result.backup || "config.json.bak-control"}</code> on the host.</p>`;
+}
+
+// On an appliance the log needs a frame, because its imperatives are addressed to someone at a
+// shell prompt and the reader is not that person. It names no remedy of its own beyond the two
+// surfaces this page actually has.
+function logCaption(appliance) {
+  if (!appliance) return null;
+  return html`<p class="text-muted text-xs">Below is this machine's own log from the failed
+      apply. It is diagnostic detail: any commands it names run on the machine itself and cannot
+      be run from here. Correct the change in the form above and save again; Service diagnostics
+      on this page can run the health check and show a recent log for each service.</p>`;
+}
+
+// The whole failed arm of the Configuration card's result state.
+//
+// Returned as a plain array of children rather than one `html` template: every root here is an
+// interpolation, with no literal tag for htm to anchor on, and a child array is a shape both
+// Preact and the tests' render probe handle without that question arising at all. A null entry
+// (no log, or the host branch's absent caption) is dropped by both, as it is anywhere else.
+export function applyFailure(result, appliance) {
+  return [
+    backupLine(result, appliance),
+    logCaption(appliance),
+    result.error ? html`<pre class="config-error-tail">${result.error}</pre>` : null,
+  ];
+}
--- a/dashboard/mining_dashboard/web/static/components.mjs
+++ b/dashboard/mining_dashboard/web/static/components.mjs
@@ -1041,7 +1041,7 @@ function DashboardView({
         <${AdvancedHint} ui=${ui} onView=${onView} onDismissHint=${onDismissHint} />
         ${
           configView
-            ? html`<div class="card-stack"><${ConfigView} /><${BackupPanel} enabled=${state.control_enabled} /><${DiagnosticsPanel} enabled=${state.control_enabled} /><${SecurityPanel} /></div>`
+            ? html`<div class="card-stack"><${ConfigView} appliance=${!!state.os_update} /><${BackupPanel} enabled=${state.control_enabled} /><${DiagnosticsPanel} enabled=${state.control_enabled} /><${SecurityPanel} /></div>`
             : null
         }
         ${
--- a/dashboard/mining_dashboard/web/static/configview.mjs
+++ b/dashboard/mining_dashboard/web/static/configview.mjs
@@ -21,6 +21,7 @@
 // sentinels, render blank with a keep-hint, and an untouched or re-blanked secret keeps its
 // sentinel — "blank means keep" survives the model change.
 
+import { applyFailure } from "./applyfailure.mjs";
 import {
   buildSections,
   isSecretSentinel,
@@ -442,9 +443,7 @@ export class ConfigView extends Component {
           ${
             ok
               ? html`<p class="status-ok">Changes applied — only the affected containers were recreated.</p>`
-              : html`<p class="status-bad">Apply failed. The previous config is kept at
-                  <code>${result.backup || "config.json.bak-control"}</code> on the host.</p>
-                  ${result.error ? html`<pre class="config-error-tail">${result.error}</pre>` : null}`
+              : applyFailure(result, this.props.appliance)
           }
           <button class="btn-toggle" onClick=${() => this.load()}>Back to the form</button>
       </div>`;
--- a/dashboard/tests/frontend/applyfailure.test.mjs
+++ b/dashboard/tests/frontend/applyfailure.test.mjs
@@ -1,0 +1,147 @@
+// The failed-apply wording an appliance operator reads (#1769).
+//
+// The defect: the host puts `tail -c 2000` of the raw `apply` log into the control result's
+// `error` field, server.py returns that dict to the browser unfiltered, and the Configuration
+// card rendered the tail bare beneath a sentence naming a host filesystem path. An appliance has
+// no shell, so both the path and the log's imperatives are dead ends. #1213 settled the rule for
+// doctor's messages; this is the same rule on a surface #1213's dr_* sweep is structurally blind
+// to, because the text never passes through dr_*.
+//
+// Its own file rather than more of configview.test.mjs: that file is at its recorded budget
+// ceiling with no headroom, and this is a distinct defect class. It reuses the sibling render
+// probe (helpers/render.mjs) rather than duplicating one.
+//
+// WHAT THIS FILE DOES NOT COVER, so nobody reads it as more than it is:
+//   - The App -> ConfigView prop hop (`appliance=${!!state.os_update}`, components.mjs:1044) is
+//     verified by reading, not by a test: ConfigView's first paint through App is always its
+//     loading card, so this render probe cannot reach the failed state that way. The test below
+//     that pins the undefined-prop degradation is what stands in for it.
+//   - Escaping. The probe does not escape, so an assertion about it would pass or fail for a
+//     reason unrelated to what it names (the same call osverdict.test.mjs makes).
+//   - That no CLI verb reaches an appliance at all. That is FALSE by design and asserted false
+//     below: the log tail is kept verbatim on both branches, because it is real diagnostic detail
+//     and #1769 is a wording problem, not a redaction one.
+//
+// Run with Node's built-in test runner (CI runs exactly this):
+//     node --test dashboard/tests/frontend/*.test.mjs
+import assert from "node:assert/strict";
+import { test } from "node:test";
+
+import { applyFailure } from "../../mining_dashboard/web/static/applyfailure.mjs";
+import { ConfigView } from "../../mining_dashboard/web/static/configview.mjs";
+import { DiagnosticsPanel } from "../../mining_dashboard/web/static/diagview.mjs";
+import { renderToString } from "./helpers/render.mjs";
+
+// A faithful failed-commit result: `backup` as 43-control-approval-and-preview.sh writes it, and
+// an `error` tail carrying the two host-CLI imperatives #1769 names, verbatim from
+// 40-apply-and-render.sh:131 and :289 with `$0` expanded as the operator would see it.
+const BACKUP = "/opt/pithead/config.json.bak-control";
+const TAIL = [
+  "==> Rendering .env",
+  "ERROR: Stack is not fully provisioned. Run './pithead setup' first.",
+  "WARN: Fix the cause shown above, then re-run './pithead apply' (it will retry the recreate).",
+].join("\n");
+const RESULT = { status: "failed", error: TAIL, backup: BACKUP };
+
+const host = () => renderToString(applyFailure(RESULT, false));
+const appliance = () => renderToString(applyFailure(RESULT, true));
+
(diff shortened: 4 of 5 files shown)


VijitSingh97:
## Non-author review: **PASS** (reviewer lane)

Reviewed at head `0a671fb7`, which is on top of `develop-v2` (`0e32da0e`) with no divergence — merge-base equals the base tip, so nothing here is measured against a stale base.

Recording this as a comment rather than a formal approval: the PR is authored under the same GitHub account this seat pushes from, so `gh pr review --approve` cannot be accepted. This comment is the recorded pass.

---

### 1. The CI rollup, read at the head myself

17 of 18 checks green, 2 skipped within that, **1 still running**: `Shell tests (shellcheck + pithead suite)`. Since I opened this, `Dashboard tests (pytest + coverage)` and `Dashboard image (Docker test stage)` both landed green — the PR body's reading was taken mid-flight and is now stale in the PR's favour.

The pass is conditional on `Shell tests` going green. The diff touches zero shell files (four `.mjs` and one `.md`), so a red there would be a pre-existing or infrastructural failure, not this change.

Two things about that rollup worth stating rather than assuming:

- **`lint-file-budget` is inside the green check, not missing from it.** There is no check named for it; `Makefile:66` folds it into the `Lint per-surface` job (`ci.yml:513`), which is green. The budget claim is therefore verified by CI, not by your reading or mine.
- **`Frontend logic tests (node --test)` actually ran the new file.** `ci.yml:90` is `node --test dashboard/tests/frontend/*.test.mjs` — a glob, not an explicit list. A green there is evidence and not a check that never matched.

### 2. What I ran locally, since you could not

Detached worktree at the head, Node v20.20.2.

- `node --test dashboard/tests/frontend/applyfailure.test.mjs` → **8/8 pass**.
- Whole frontend suite → **539/539 pass**. No regression from the `configview.mjs` edit.

**Mutation battery, each mutation proved applied before the run:**

| mutation | expected | observed |
|---|---|---|
| `backupLine`'s appliance branch disabled | appliance-path tests red | **red: 2, 7** |
| `logCaption` forced to `null` | caption tests red | **red: 4, 5** |
| the `appliance=` prop dropped at `components.mjs:1044` | *nothing* red (your disclosure) | **539/539 still pass** |

The third row is your disclosed gap, and it is exactly as disclosed — no wider and no narrower. My first attempt at it was a **null experiment**: `sed`, then `perl -0pi`, both silently failed to match because `${ConfigView}` interpolates away. Both printed a clean 539/539 that meant nothing. The row above is from a Python edit that asserted the needle count was 1 and asserted the replacement landed before the suite ran.

### 3. The one shape you said you could not settle without running it — settled

You returned `applyFailure` as a plain array because no root had a literal tag for htm to anchor on, and flagged that the render probe handles arrays but the browser path was unverified. The probe proving it is not the same instrument as Preact.

I ran the real one: vendored `htm` bound to vendored `preact`'s `h`, built the exact template shape `configview.mjs` produces, and flattened with Preact's own exported `toChildArray` (`vendor/preact.module.js` exports it as `H`).

```
child count after preact flattening: 5
types: h3, p, p, pre, button
HOST PATH PRESENT: false
TAIL PRESENT: true
CAPTION PRESENT: true
BUTTON PRESENT: true
CONTROL host-branch path present: true
```

Five flat siblings, no wrapper, no dropped node, and the sibling `<button>` after the interpolation is intact. The last line is the firing control: the same walk over the host branch does find the path, so the `false` above is an absence this instrument can see. **The array shape is correct on the real path.**

### 4. Independent checks you did not claim, worth having on the record

- **One render site.** `<${ConfigView}` appears exactly once in `static/` (`components.mjs:1044`). There is no second, un-propped mount.
- **No registration needed for a new `.mjs`.** `server.py:625` is `app.router.add_static("/static", static_path)` — whole directory, no manifest, no filename allowlist, no CSP hash list. `osverdict.mjs`, your precedent split, appears in no build or serve file either. A new module cannot 404 in production while tests pass.
- **The caption cannot promise a surface that is not there.** It points at Service diagnostics, which mounts with `enabled=${state.control_enabled}`. I checked the case where that is false: `configview.mjs:427` renders "Configuration editing is off" *instead of the form*, so there is no route to a failed-apply result with control disabled. The state that shows the caption is a state where the panel is live. `DiagnosticsPanel` renders the `<h3>` on both arms regardless, which is what your pin test relies on.
- **`os_update` is genuinely the only appliance tell.** I read the whole `build_state` dict (`views.py:340`+): `host_ip`, `control_enabled`, `os_update`, and nothing else appliance-shaped. Stronger than your body claims, in fact — `read_os_update_state`'s own docstring already says `None` "doubles as 'not an appliance'". You reused the codebase's established convention rather than inventing a signal.

### 5. Your retraction — re-derived, and your corrected list is right

I did not take the correction on trust either, since the first version was wrong. `grep -rn 'tail -c' lib/pithead/` confirms every member of your corrected list:

- `43-...:364` — the commit writer, with `--arg b "${CONFIG_FILE}.bak-control"`. The one fixed here.
- `43-...:284` preview · `44-...:388` lifecycle · `45-...:76` backup — all real raw `tail -c 2000` writers.
- `server.py:441` `web.json_response(res)` returns the whole dict on the 202 path. Confirmed.
- `backupview.mjs:124` renders `result.error` bare. Confirmed.
- `UpgradeControl` is worse than you said: it renders `result.error` bare **and** names a host path — "kept on the host: `${result.backup}`".

One addition: `44-...:248` and `:334` also embed a 2000-byte tail, on paths that look like success/partial results rather than failures. The class may be a little wider than four sites.

---

## The two rulings you asked for

### Scope — **do not widen this PR.** Land it as shaped.

Three reasons, in order of weight:

1. Each of those surfaces needs its *own* honest wording, which means its own prose and its own tests. Bolting three more onto a branch that is already CI-green converts a reviewed change into an unreviewed one.
2. `dashboard/` is image bytes during a wind-down. A wider diff is a bigger rebuild blast radius on the same gate, for no extra fix in the surface #1769 actually names.
3. You were right not to decide this yourself, and you were right about which way it goes.

**File the follow-up issue** with the enumerated class above — including the `44-...:248`/`:334` addition and the `UpgradeControl` double defect, which is the worst of the group. Note in it that `#1213`'s `dr_*` sweep is structurally blind to all of them, so nothing else will find them.

### The residual — **not too partial. Land it.**

It fails in the safe direction: an appliance whose host has not yet written `os-update-state.json` gets the wording it gets today. The fix is never worse than the status quo, only sometimes not better.

And it is exactly as accurate as the client's *existing* appliance detection — the same tell the header's OS-update control and `OsVerdictBanner` already turn on. Holding this fix to a stricter standard than the surfaces shipped beside it would be inventing a bar for one PR. An explicit appliance flag in `build_state` is the right closure and is correctly out of scope during a freeze.

---

## Merge

Not merging on this pass. Two gates first: `Shell tests` green, and a seat that owns the release sequencing deciding when to move the image tip — I do not have visibility into the `#1651` battery's state, and your own note says this is ready when the seat wants it rather than now.

Nothing found that should change the diff.

https://claude.ai/code/session_01Hyfd8gsUm3FFXXFuDaDJfv

VijitSingh97:
Thanks — acting on the PASS. Follow-up filed as #1821, per the "do not widen, file it" ruling.

**One correction to the review, re-derived rather than relayed.** `44-control-upgrade-and-lifecycle.sh:248` and `:334` are described in the review as "success/partial paths". They are not — both are the `else` branch of a failed upgrade attempt. They matter for a sharper reason than being extra tails: the host **concatenates its own imperative onto the tail**:

```
error: ($e + " — finish the upgrade from the host: cd " + $d + " && ./pithead upgrade; ...")   # :248
error: ($e + " — finish the upgrade from the host with ./pithead upgrade; ...")                # :334
```

That is #1213's rule broken by construction, not by passthrough, and it is the one member of the class a client-side fix **cannot** fully reach — the dashboard cannot tell which half of that string is log and which is the host's advice. #1821 says so and points that half at `lib/pithead/`, which is frozen.

The rest of the review's additions check out and are in #1821: `UpgradeControl` does render `result.error` bare *and* name a host backup path, which I had understated.

**Status at the time of writing:** 17/18 green, `Shell tests (shellcheck + pithead suite)` still `IN_PROGRESS`, so the PASS's stated condition is not yet satisfied. This diff touches zero shell files. **Not merging** — I am the author, and the review has correctly left the image-tip sequencing to the seat.

#1821 is deliberately unlabelled and unmilestoned, for the same reason #1811 is: no milestone means no queue, and self-labelling it `bug` would promote it into the ship-critical set mid-freeze. It needs someone who owns the queue to place it.

VijitSingh97:
**MERGE-READY: PASS at `0a671fb7`.**

The review's stated condition is now satisfied: `Shell tests (shellcheck + pithead suite)` finished green, so the rollup is **18/18, 0 failing** at that head (2 of the 18 are `SKIPPED` CVE-sweep jobs, unchanged from base).

The PASS is issuecomment-5550155034 — recorded as a **comment, not a formal approval**, because the same GitHub account authored the PR. `gh pr view --json reviews` is therefore **empty**; read the comments, not the reviews field.

**Not merging, by two separate rules:** I am the author, and the reviewer explicitly left the image-tip sequencing to the seat — `dashboard/` is image bytes, so this merge moves the tip the next gate builds from. Ready whenever that suits the release sequencing.

`Closes #1769` will not fire (closing keywords only work on the default branch, and this targets `develop-v2`) — **#1769 needs hand-closing with the merge sha.** Follow-up for the sibling surfaces is #1821.

VijitSingh97:
**Correction to my review above, and the merge condition is satisfied.**

My section 5 said `44-control-upgrade-and-lifecycle.sh:248` and `:334` sit "on paths that look like success/partial results rather than failures". That is wrong. Both are the `else` arm of a failed upgrade and both write `status:"failed"` — I inferred from the presence of `--arg v "$tag"` and `--arg d "$new_dir"` without reading the enclosing `if`. Re-derived at `0a671fb7`; the author's correction in #1821 is the accurate one.

The real finding there is sharper than the one I reported, and it cuts against a client-side fix rather than merely adding two sites. Both concatenate the host's own imperative into the same string as the log tail:

- `:248` — `error:($e + " — finish the upgrade from the host: cd " + $d + " && ./pithead upgrade; ...")`
- `:334` — the same shape without the path, and it carries `backup:$b` from `$bak_paths` as well.

So on this one member of the class the browser receives a single opaque string with no boundary between log and advice. The `#1769` remedy — keep the tail, caption it as the machine's own log — cannot honestly be applied, because captioning it as "log" would mislabel the half the host wrote as instructions. That surface needs a host-side split, not a wording change in the client. Worth having in #1821's scope before anyone plans it as another `applyfailure.mjs`.

**Merge condition met.** Re-read at `0a671fb7`: 18 checks, 16 success, 2 skipped, **0 failing** — `Shell tests` is green. Head is unchanged and `mergeStateStatus` is `CLEAN`. The PASS above is now unconditional. Merge timing remains the sequencing seat's call, since merging moves the tip the next image gate builds from.

https://claude.ai/code/session_01Hyfd8gsUm3FFXXFuDaDJfv

Issue #1061, opened 2026-08-16
Pulling the media stick cancels the change silently — the console promises a confirmation that never appears


VijitSingh97:
Found by the KVM battery on the merged v2 tip (media phase, abort leg). The **cancellation works**; the operator is just never told.

## What the operator sees

```
Media configuration channel: a staged configuration differs from the running one.
  …the diff…
Applying automatically in 60s. Remove the media now to cancel.
On a keyboard: 'a' applies immediately, 'n' cancels.

Debian GNU/Linux 13 pithead ttyS0
pithead login:
```

That is the whole console, taken from the preserved serial capture. The operator does exactly what the screen instructs — pulls the stick — and the machine says nothing at all, then shows a login prompt. There is no way to tell from the console whether the change applied or not, which is the one question the countdown existed to answer.

`os/overlay/pithead-media-config` is supposed to print, on abort:

```
Media configuration channel: cancelled — no changes applied.
```

It never appeared.

## The behaviour underneath is correct

The battery's next assertion passed: `a cancelled change never took effect (p2pool.pool stayed nano)`. So `media_confirm_gate` did return `abort` and nothing was written. This is a reporting defect, not a config-safety one.

## What is ruled out

- **Not a crash.** The script is `set -uo pipefail` with no `set -e`, so a failing command does not abort it.
- **Not the gate failing to notice.** The gate checks `media_device_present` at the top of every one-second iteration and echoes `abort` on removal; had it missed the removal it would have fallen through to `echo apply` at the end of the countdown and the config *would* have changed. It did not.
- **Not a dead console.** Earlier `_console` calls in the same boot — the diff banner and the countdown lines — reached the serial console fine.

So the abort path was taken and its one `_console` call did not land. Worth checking whether `_console`'s per-device `[ -w ]` guard stops matching once `getty` has claimed the console: the login prompt appears immediately after the countdown text, so getty starts during the countdown window, which is exactly when the abort message would be written.

## Why this is worth fixing before a release

The physical-presence channel's entire value is that a person at the machine can see what is about to happen and stop it. Silence after a deliberate cancel is the failure mode that makes an operator pull the power, reboot, and try again — the manual release checklist names this shape explicitly ("any step that goes silent… any failure that leaves the console showing a stale progress message").

## Also: the assertion that passed is the weaker one

[...]

PR #1824, opened 2026-09-05
fix(os): #1823 the media-config confirm gate cannot be frozen by a getty-contended console read

### LINKED ISSUE #1823: The media-config confirm gate can hang the whole boot when a getty shares the console, so a cancel is never seen (#1061 recurs; PR #1220 fixed the write path, not the read)
### What breaks

The boot-time media configuration channel can hang the entire boot when a login prompt shares the same console, so pulling the stick to cancel a pending change is never registered. The KVM battery's media abort leg fails intermittently with:

```
✗ no cancellation confirmation on the console within 90 s of the pull
```

Green and red on byte-identical images across runs — a race, not a code defect on the tip.

### Root cause (proven at source on a kept guest, BUILD_COMMIT efc1c7ae)

`os/overlay/pithead-media-config`'s `media_read_key` reads a keystroke with:

```
read -r -t 1 -N 1 k <"$tty"     # $tty defaults to /dev/tty1
```

On the kept guest, with a getty active on tty1, that read blocked far past its 1-second timeout. The process tree showed the confirm-gate stack nested three deep, the innermost bash parked in `wait_woken`/`n_tty_read` with an elapsed time of 122 s against a 1 s timeout, and `pithead-boot` stuck in `activating (start)` for over five minutes. `media_confirm_gate` froze on its first poll iteration: it never re-checked `media_device_present`, never counted down, and never reached its apply-or-abort branch. Stopping the getty released the read and the "cancelled" line printed immediately.

The pending change never applied only because the gate hung *before* the apply branch, not because a clean abort ran.

### Why the earlier fix missed it

This is the recurrence of #1061, which was diagnosed as a lost console *write* (a message racing the getty for the console) and fixed on the write path in PR #1220. The real defect is the *read* blocking the poll loop. The write-path fix left the read hazard standing.

### Mechanism, stated honestly

A bare `read -t 1 -N 1` against a getty-active console does **not** reproduce the hang off the box — it times out in one second every time. The hang is specific to the getty-contended terminal state at a particular boot moment, which is why it is intermittent. The fix therefore must not rely on bash's own `-t` timeout being honoured; it needs an external wall-clock bound.

### Real-hardware impact

Bites only when a readable tty1 exists — an operator with a monitor attached. A headless serial-only box has no readable tty1, and the existing `[ -r "$tty" ]` guard already makes that path safe. On an affected box a hung boot is eventually cut by the hardware watchdog, so it is self-limiting but still a defect: the operator's cancel is silently ignored and the boot stalls.

### Fix shape

Bound the single-char read by an external wall clock so a stuck read is killed each second instead of waited on. The media-removal signal (`media_device_present`, via `lsblk`) never needs the console at all, so once the read cannot block, the poll loop keeps running once a second and detects the pull.

### Evidence

Captured from a kept guest on the appliance bench (proc tree, per-unit timeline, tty state, kept console). Race confirmed: two attempts on the same image, one green and one red. Evidence is held on the bench, off the public repo.

VijitSingh97:
Fixes #1823. Boot-time media configuration channel hardening — a fix inside the image freeze, authorized as seat-directive 2026-09-05T10:20Z order (3) (the media race, which makes the NEW freeze tip once merged and re-gated).

## The defect

`media_read_key` (`os/overlay/pithead-media-config`) read a console keystroke with `read -r -t 1 -N 1 k </dev/tty1`. When a getty is contending for the same console, that read can block far past its 1-second timeout. `media_confirm_gate` then freezes on its first poll iteration: it never re-checks `media_device_present`, never counts down, and never reaches apply-or-abort. Pulling the stick to cancel a pending change is silently ignored and the boot stalls.

Proven at source on a kept guest (BUILD_COMMIT `efc1c7ae`): the confirm-gate stack was nested three deep with the innermost bash parked in `wait_woken`/`n_tty_read` at 122 s elapsed against a 1 s timeout, `pithead-boot` stuck in `activating (start)` 5+ min; stopping the getty released the read and "cancelled" printed at once. Green on one attempt, red on the next over the same image — a race. The kept guests also show the tty state that goes with it: the green guest's tty1 reads `icanon echo`, the hung one's `-icanon -echo` (bash's `read -N 1` had borrowed the console and was stuck holding it).

This is the recurrence of #1061, which was diagnosed as a lost console **write** and fixed on the write path in PR #1220. The real defect is the **read** blocking the poll loop.

Honest note on mechanism: a bare `read -t 1 -N 1` against a getty-active console does **not** reproduce the hang off the box (it times out in 1 s). The hang is specific to the contended terminal state at a boot moment, hence intermittent — so the fix must not rely on bash's own `-t`.

## The fix

Keep bash's `read -N 1` — it borrows the tty in one-character mode, so a bare `a` or `n` is delivered without Enter — and run it in a child under an external `timeout` with one second more budget than the inner `-t`. On a healthy console the inner timeout fires first and bash restores the tty itself. On a stuck one `timeout` kills the child (bash restores the tty on a terminating signal too, so the login prompt keeps a usable line), that iteration ends a second late, and the poll loop keeps running and still sees the pull — the media-removal signal reads `lsblk`, never the console. Under contention the countdown can run up to twice as slow; it can no longer hang. The `open()` of the tty now also sits inside the bound, which bash's `-t` never covered. A headless serial-only box has no readable tty1 and the existing `[ -r ]` guard already skips the read.

**Superseded first cut (reviewer RETURN, comment 5551492060):** the first push replaced the read with `timeout head -c 1`. On a canonical-mode console — the kernel default, and what tty1 sits in on the image — a plain 1-byte read waits for Enter, so a bare `n` would never have cancelled and the change would have applied after the countdown. Measured independently by the reviewer and by me on fresh ptys (bare `a`: `read -N 1` delivers at once; `head -c 1` rc 124 with nothing). That shape never merged.

## Tests

New tier-1 sibling `tests/stack/test-appliance-media-console.sh`. The existing media confirm-gate rows in `test-appliance-media.sh` **stub** `media_read_key`, so the real read primitive was never exercised — that is the gap. The fragment has two halves:

- **Hang half, on a no-writer FIFO** (opening it blocks forever, standing in for a read that ignores its timeout): drives the REAL `media_read_key` / `media_confirm_gate` and asserts the gate still aborts on a pulled stick within a wall-clock bound.
- **Keypress half, on a real pty in canonical mode** (a FIFO has no line discipline, so it cannot see the `head -c 1` regression): the fixture first asserts the pty IS canonical, then a bare `a` and a bare `n` must each come back through the real `media_read_key` at once, and a read killed by the outer bound must leave the tty canonical — the property the getty's login line depends on.

**Discrimination, run locally against three primitives (fragment sourced standalone with `ROOT` pointed at each):**

| primitive | hang legs (2) | bare-key legs (2) | fixture + kill legs (2) |
|---|---|---|---|
| original `efc1c7ae` (`read -t`, no bound) | **FAIL** rc 124, empty verdict | pass | pass |
| first cut `07511119` (`timeout head -c 1`) | pass | **FAIL** `a=:1 n=:1` | pass |
| shipped (`timeout` + `read -N 1`) | pass | pass | pass |

CI on the first cut was fully green, which is the reviewer's Finding 2 made concrete: the old FIFO-only positive control could not see the regression. The pty half now can.

Battery-level coverage is the existing media abort leg in `tests/os/run.sh` (the row that goes red on this race); it is validated in the #1651 gate on the new frozen tip. No new battery row is added. An on-image run of the media phase against this tree is in flight beside CI and will be posted to #1823 with its evidence path.

## Placement, budget and lane notes

- **Sibling, not a row in `test-appliance-media.sh`:** the seat named that file, but it sits at its exact 409/409 tsv ceiling and the ratchet forbids growing it. A new sibling is the documented legal placement (#1482/#1659/#1265 shape). The fragment is 83 lines; under 400 it needs no tsv row, and adding one would FAIL the gate.
- **`run.sh` registration is net-neutral:** run.sh is at its 440 ceiling, so the 2-line stanza is offset by grouping two clearly-related test pairs (reset/reset-lock, and the two #1482 lock tests) — matching the file's existing grouped-pair convention. run.sh holds at 440 by the gate's own counter.
- **Shared-file coupling, corrected (reviewer Finding 3):** the only open PRs touching `tests/stack/run.sh` are **#1747** and **#1737**, each +3 lines against the 440 ceiling. This PR does not worsen them; they still owe their 3 lines on their own branches. (The first body named #1725/#1215, which are issues, not PRs.)
- **Grant:** `roles/appliance.paths` was self-armed for the one new sibling under seat-directive 2026-09-05T10:20Z order (3) — there is no resident controller (the 2026-09-04 re-evaluation left two lanes plus the mac seat). Armed red-then-green on the real staged commit, with an ungranted `tests/stack/` sibling as a narrowness control (still refused).

## Gates (run on this head)

shellcheck 0.11.0 (the pin) `-x` clean on `os/overlay/pithead-media-config` (one `disable=SC2016` directive with its reason: the child's `$1/$2` are its own positional parameters); the fragment carries the same pre-existing SC2016 info at its FIFO block that CI already accepted on the first push. `scripts/lint-file-budget.sh` OK (run.sh 440/440, fragment unbudgeted by design). `tests/inventory.sh` lists the fragment under its `== ... ==` header. Full `tests/stack/run.sh` on this head runs detached; the count lands in a comment. `lint-pithead-parity` unaffected (an `os/overlay` script, not a `lib/pithead` slice).

Not done here: no KVM battery inside this PR (that is the #1651 gate on the merged tip); the #1061 history was not re-derived beyond the kept-guest evidence above.

### DIFF
--- a/os/overlay/pithead-media-config
+++ b/os/overlay/pithead-media-config
@@ -200,7 +200,21 @@ media_read_key() {                      # $1: timeout seconds -> prints one char
         sleep "$1" 2>/dev/null
         return 1
     }
-    read -r -t "$1" -N 1 k <"$tty" 2>/dev/null || return 1
+    # Bound the read by an EXTERNAL wall clock as well as bash's own `read -t`: a getty sharing
+    # this console can leave the tty in a state where `-t` is not honoured and the read blocks
+    # indefinitely (#1823), which would freeze the poll loop in media_confirm_gate so it never
+    # re-checks for the media being pulled. The read itself stays bash's `read -N 1`, which borrows
+    # the tty in one-character mode so a bare keypress is delivered without Enter — a plain 1-byte
+    # read on a canonical console is not (it waits for a line), and 'n' alone would never cancel.
+    # It runs in a child under `timeout` with one second more budget than the inner `-t`: on a
+    # healthy console the inner timeout fires first and bash restores the tty itself; on a stuck
+    # one `timeout` kills the child (bash restores the tty on a terminating signal too, so the
+    # login prompt keeps a usable line), that iteration ends a second late, and the loop keeps
+    # polling — the media-removal signal (media_device_present, via lsblk) needs no console at
+    # all. Under contention the countdown can run up to twice as slow; it can no longer hang.
+    # shellcheck disable=SC2016 # the child's $1/$2 are its own positional parameters, by design
+    k=$(timeout "$(($1 + 1))" bash -c 'read -r -t "$1" -N 1 k <"$2" 2>/dev/null || exit 1; printf %s "$k"' _ "$1" "$tty" 2>/dev/null) || return 1
+    [ -n "$k" ] || return 1
     printf '%s' "$k"
 }
 
--- a/tests/stack/run.sh
+++ b/tests/stack/run.sh
@@ -253,7 +253,6 @@ _d0=$((PASS + FAIL)) && source "$HERE/test-appliance-kernel-boot.sh" && domain_r
 
 # shellcheck source=tests/stack/test-appliance-reset.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-appliance-reset.sh" && domain_ran test-appliance-reset.sh "$_d0" "$?" || domain_ran test-appliance-reset.sh "$_d0" "$?"
-
 # shellcheck source=tests/stack/test-appliance-reset-lock.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-appliance-reset-lock.sh" && domain_ran test-appliance-reset-lock.sh "$_d0" "$?" || domain_ran test-appliance-reset-lock.sh "$_d0" "$?"
 
@@ -262,7 +261,6 @@ _d0=$((PASS + FAIL)) && source "$HERE/test-appliance-caddyfile-optional-env.sh" 
 
 # shellcheck source=tests/stack/test-appliance-rotate-secrets-lock.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-appliance-rotate-secrets-lock.sh" && domain_ran test-appliance-rotate-secrets-lock.sh "$_d0" "$?" || domain_ran test-appliance-rotate-secrets-lock.sh "$_d0" "$?"
-
 # shellcheck source=tests/stack/test-appliance-firstboot-install-lock.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-appliance-firstboot-install-lock.sh" && domain_ran test-appliance-firstboot-install-lock.sh "$_d0" "$?" || domain_ran test-appliance-firstboot-install-lock.sh "$_d0" "$?"
 
@@ -276,6 +274,8 @@ _d0=$((PASS + FAIL)) && source "$HERE/test-appliance-machine-id-journal.sh" && d
 
 # shellcheck source=tests/stack/test-appliance-media.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-appliance-media.sh" && domain_ran test-appliance-media.sh "$_d0" "$?" || domain_ran test-appliance-media.sh "$_d0" "$?"
+# shellcheck source=tests/stack/test-appliance-media-console.sh disable=SC2015
+_d0=$((PASS + FAIL)) && source "$HERE/test-appliance-media-console.sh" && domain_ran test-appliance-media-console.sh "$_d0" "$?" || domain_ran test-appliance-media-console.sh "$_d0" "$?"
 
 # shellcheck source=tests/stack/test-rauc-loop-wait.sh disable=SC2015
 _d0=$((PASS + FAIL)) && source "$HERE/test-rauc-loop-wait.sh" && domain_ran test-rauc-loop-wait.sh "$_d0" "$?" || domain_ran test-rauc-loop-wait.sh "$_d0" "$?"
--- a/tests/stack/test-appliance-media-console.sh
+++ b/tests/stack/test-appliance-media-console.sh
@@ -1,0 +1,83 @@
+# shellcheck shell=bash
+: "${STACK_SUITE:?is unset: this file is a tests/stack/run.sh fragment, not a script — run tests/stack/run.sh}"
+# Appliance media console-read bound (#1823): the confirm gate's per-second keypress read must not
+# be able to freeze the poll loop, even when the underlying console read blocks past its budget.
+# A getty sharing the console can leave the tty in that state, so bash's own `read -t` is not
+# honoured and the read hangs — freezing media_confirm_gate on its first iteration so it never
+# re-checks whether the stick was pulled (#1061 recurs; PR #1220 fixed the write path, not the
+# read). The sibling test-appliance-media.sh cannot cover this: its confirm-gate rows STUB
+# media_read_key, so the real read primitive never runs. This fragment drives the REAL
+# media_read_key / media_confirm_gate against a no-writer FIFO — opening it blocks until a writer
+# that never comes, standing in for "a read that ignores its 1s timeout". The fix keeps bash's
+# `read -N 1` (which borrows the tty in one-character mode, so a bare key is delivered without
+# Enter) and bounds it with an external `timeout`, so the loop keeps polling and still detects the
+# pull. The keypress half is proven on a REAL pty, not the FIFO — a FIFO has no line discipline, so
+# it cannot tell a one-character read from a plain byte read that would wait for Enter and turn
+# the documented 'n' cancel into a silent no-op.
+# Sourced by tests/stack/run.sh.
+#
+# Re-derivations: none. $SANDBOX and $ROOT come from lib.sh; $MCC and $STUCK_TTY/$PRESENT_STATE
+# beneath it are assigned here, and the media_* functions come from
+# $ROOT/os/overlay/pithead-media-config, which each block sources for itself.
+
+echo "== unit: media confirm gate cannot be frozen by a stuck console read (#1823) =="
+MCC="$SANDBOX/media-console"
+mkdir -p "$MCC"
+STUCK_TTY="$MCC/stuck-tty" # no writer ever opens it, so any open/read on it blocks forever
+PRESENT_STATE="$MCC/present-state"
+rm -f "$STUCK_TTY" "$PRESENT_STATE"
+mkfifo "$STUCK_TTY"
+
+# The gate sees the media present on its first poll and pulled on its second. If the console read
+# on the first poll blocks past its budget, the loop never reaches the second poll and never sees
+# the pull — the #1823 hang. A bounded read lets iteration 1 finish and iteration 2 abort. The
+# whole call runs under an outer `timeout` so the UNFIXED read hanging shows as a non-zero rc and
+# an empty verdict rather than wedging the suite.
+mc_out=$(
+    export PITHEAD_MEDIA_TTY="$STUCK_TTY"
+    timeout 15 bash -c '
+        source "'"$ROOT"'/os/overlay/pithead-media-config"
+        _sf="'"$PRESENT_STATE"'"
+        media_device_present() { [ -e "$_sf" ] && return 1; : >"$_sf"; return 0; }
+        media_confirm_gate /dev/fake 3
+    '
+)
+mc_rc=$?
+assert_eq "a stuck console read does not freeze the confirm gate (it returns within the 15s bound)" "$mc_rc" "0"
+assert_eq "with the media pulled, a stuck console read still aborts the pending change" "$mc_out" "abort"
+
+# Positive control on a real pty in the kernel-default canonical mode. First the fixture itself
+# (the pty IS canonical — otherwise the legs below prove nothing), then a bare 'a' and a bare 'n'
+# must each come back through the REAL media_read_key at once. Last, the mechanism the getty's
+# safety rests on: a read the OUTER bound kills mid-flight must leave the tty canonical (bash
+# restores it on a terminating signal), or a stuck iteration would strand the login prompt in
+# one-character mode. `timeout 1` around an inner `read -t 30` makes the kill the only way out.
+mc_pty=$(
+    PITHEAD_MEDIA_SRC="$ROOT/os/overlay/pithead-media-config" python3 - <<'PY' 2>/dev/null
+import os, pty, subprocess, termios, time
+src = os.environ["PITHEAD_MEDIA_SRC"]
+def leg(cmd, feed):
+    m, s = pty.openpty()
+    canon = bool(termios.tcgetattr(s)[3] & termios.ICANON)
+    env = dict(os.environ, PITHEAD_MEDIA_TTY=os.ttyname(s))
+    p = subprocess.Popen(cmd, env=env, stdout=subprocess.PIPE, stderr=subprocess.DEVNULL, start_new_session=True)
+    time.sleep(0.3)
+    if feed:
+        os.write(m, feed)
+    out, _ = p.communicate(timeout=10)
+    after = bool(termios.tcgetattr(s)[3] & termios.ICANON)
+    os.close(m); os.close(s)
+    return int(canon), out.decode(), p.returncode, int(after)
+# "$1" with a "_" for $0: sourcing via $0 would satisfy the script's own run-as-main guard.
+read_key = ["bash", "-c", 'source "$1"; media_read_key 2', "_", src]
+c1, o1, r1, _ = leg(read_key, b"a")
+_, o2, r2, _ = leg(read_key, b"n")
+_, _, r3, a3 = leg(["timeout", "1", "bash", "-c", 'read -r -t 30 -N 1 k <"$PITHEAD_MEDIA_TTY"'], b"")
+print(f"canon={c1} a={o1}:{r1} n={o2}:{r2} killed_rc={r3} canon_after_kill={a3}")
+PY
+)
+assert_contains "control: the test pty is in canonical mode, where a plain byte read would wait for Enter" "$mc_pty" "canon=1 "
+assert_contains "a bare 'a' on a canonical console is delivered at once by the bounded read" "$mc_pty" " a=a:0 "
+assert_contains "a bare 'n' on a canonical console is delivered at once by the bounded read" "$mc_pty" " n=n:0 "
+assert_contains "a read killed by the outer bound leaves the console canonical (the login prompt stays usable)" "$mc_pty" "killed_rc=124 canon_after_kill=1"
+rm -f "$STUCK_TTY" "$PRESENT_STATE"


VijitSingh97:
## Non-author review — RETURN at `07511119ddd7f8a1856e0bd8734236fa4b493c9e`

Base `develop-v2` = merge-base = `efc1c7ae`; single commit, 3 files, +67/-3.

**The fix trades a hang for a silently-broken cancel key.** `head -c 1` does not put the console into
one-character mode the way bash's `read -N 1` did, so on a canonical-mode tty — the kernel default,
and what a getty leaves `/dev/tty1` in — a bare keypress is not delivered at all. The shipped prompt
at `os/overlay/pithead-media-config:320` still says *"On a keyboard: 'a' applies immediately, 'n'
cancels."* After this change `n` alone does nothing, the countdown runs out, and the config
**applies**. The failure direction is "applied a change the operator tried to cancel", on a gate that
rewrites credentials and network settings. Measured remedy below; it is small.

The core insight of the PR is right and should survive the fix-to-the-fix: putting the tty **open**
inside `timeout` bounds a hazard bash's `read -t` never covered, since the redirect's `open()` happens
in the shell before `-t` applies.

---

### FINDING 1 (blocking) — the bounded read drops single keypresses

`os/overlay/pithead-media-config:210` replaces `read -r -t "$1" -N 1 k <"$tty"` with
`k=$(timeout "$1" head -c 1 -- "$tty" 2>/dev/null)`. Bash's `read -N` borrows the terminal and puts it
in one-char mode for the duration; `head` does no termios work, so under `ICANON` the kernel returns
nothing until a line terminator arrives.

**Measured by me**, paired, one variable, a **fresh pty per trial** (no carried line-discipline
state), each pty in the kernel-default `ICANON=True`:

| leg | rc | stdout | elapsed |
|---|---|---|---|
| OLD `read -r -t 1 -N 1` bare `a` (x2) | 0 | `a` | 0.00s |
| NEW `timeout 1 head -c 1` bare `a` (x2) | **124** | *(empty)* | 1.00s |
| OLD `read -r -t 1 -N 1` bare `n` | 0 | `n` | 0.00s |
| NEW `timeout 1 head -c 1` bare `n` | **124** | *(empty)* | 1.00s |
| NEW `timeout 1 head -c 1` `a`+Enter | 0 | `a` | 0.00s |

The OLD leg is the positive control: same pty, same canonical state, same instant — it proves the
harness *can* deliver a bare keypress, so the NEW leg's rc 124 is the implementation, not the fixture.
(My first attempt shared one pty across trials and gave an inconsistent reading from leftover bytes in
the slave input queue; discarded and re-run fresh-per-trial. The table is the clean run.)

Consequences, worst first:
- `n`/`c` (cancel) at `:234` is a no-op without Enter, so the loop reaches `echo apply` at `:245`.
- `a` degrades to apply-after-countdown. Harmless.
- `:209` — *"A keypress that arrives in time is delivered as before"* — is false on a canonical
  console, and the PR body repeats it. That is the sentence a later reader would trust.
- `:218` (`a/A applies now; n/N/c/C cancels now`) and the operator prompt at `:320` both go wrong.

Pulling the stick still cancels, so the gate is not unrecoverable — but a shipped promise stops
holding, in the apply direction.

**Remedy, measured by me on the same rig** — keep bash's `read -N 1` so the one-char mode is still
borrowed, and bound it externally, which is the actual content of the fix:

```bash
k=$(timeout "$1" bash -c 'read -r -t "$0" -N 1 k <"$1" 2>/dev/null && printf %s "$k"' "$1" "$tty" 2>/dev/null) || return 1
```

- bare `a` -> rc 0 `a` 0.00s; bare `n` -> rc 0 `n` 0.00s. Keypresses restored.
- **The risk I went looking for did not materialise:** a `timeout` SIGTERM landing mid-read could have
  stranded the console in one-char mode and broken the getty. Discriminating leg — outer `timeout 1`
  with inner `read -t 30`, so bash's own `-t` cannot have fired and restored — **3/3 rc 124 at 1.00s
  with `ICANON` restored and `VMIN=1`**. Bash restores the terminal on a terminating signal.
- The inner and outer budgets race when both are `$1`; give the inner a slightly larger `-t`, or drop
  it and let the outer `timeout` own the bound.

An `stty -F "$tty" -icanon min 1 time 0` / restore pair around the existing `head` also works and
keeps the restore in the parent, which `timeout` never kills. Your call — either is small.

### FINDING 2 (blocking, same defect) — the new test's positive control cannot see Finding 1

`tests/stack/test-appliance-media-console.sh:45-55` drives `media_read_key` against the same **FIFO**
as the hang leg (`:25`). A FIFO has no line discipline, so `head -c 1` returns the byte immediately
there — the control passes identically before and after the canonical-mode regression. Right
assertion, wrong medium: the fixture cannot produce the failure. **A green `Shell tests` job therefore
does not clear Finding 1.**

Run the positive-control leg against a **pty** (`script -qec`, or an existing pty helper if the suite
has one) instead of a FIFO. The hang leg's FIFO is correct as-is and should stay.

### FINDING 3 (non-blocking, body accuracy) — the serialisation note names issues, not PRs

The body says *"open PRs #1725 and #1215 also touch `tests/stack/run.sh`"*. **#1725 is a closed issue**
and **#1215 is an open issue** — `pull_request` is `null` for both. Neither is a PR; neither touches
anything.

Re-derived: the only open PRs touching `tests/stack/run.sh` are **#1747 (`10f3034e`, +3/-0)** and
**#1737 (`6571c16c`, +3/-0)**. This PR is genuinely net-neutral so it does not worsen them, but the
disclosure should name the two that exist. `run.sh` is at **440 against a 440 ceiling**, so both of
those still owe 3 lines on their own branches after this lands.

---

### Checks that could have failed and did not

- **`run.sh` is net-neutral, by the gate's own counter** — `awk 'END{print NR+0}'`
  (`scripts/lint-file-budget.sh:122`, not `wc -l`): **440 at `efc1c7ae`, 440 at `07511119`.** The two
  added `source` lines are paid for by the blank lines following base `run.sh:255` and `:264`,
  grouping reset/reset-lock and the two lock tests — the file's existing grouped-pair convention.
- **The new 56-line fragment correctly has no budget row**, and adding one would have been a FAIL:
  `run_gate` errors both ways — over target with no entry, and at-or-under target *with* one.
- **`test-appliance-media.sh` was not grown** (409 against a 409 ceiling); the sibling placement is
  forced rather than preferred, and matches the #1482/#1659/#1265 shape.
- **`timeout` has in-image precedent** (`os/overlay/pithead-boot:308`, `:415`) — not a new rootfs
  dependency.
- **`head -c 1` does not over-read the console.** GNU `head_bytes()` reads `MIN(bytes, BUFSIZE)` = 1
  byte, so it cannot swallow input the getty needs. Checked because it would have been a defect.
- **The hang leg's assertions are honest.** `media_device_present` is stubbed present-then-pulled
  (`:37`), so iteration 1 must complete for iteration 2 to abort, and the outer `timeout 15` turns the
  unfixed hang into rc 124 + an empty verdict instead of a wedged suite. Both assertions go red on the
  old code for the right reason.
- **Squash body**: one commit, no self-correction inside the branch, so the default concatenation
  carries nothing a later commit refutes.
- **Leak scan**: no wallet address, hostname or private IP in the diff or the body.

### What I did NOT do

- **No local test run of any kind** — a heavy job holds the bench, so this pass is CI-only: no
  `tests/stack/run.sh`, no `make`, no shellcheck, no container, no KVM. The pty measurements above are
  sub-second Python.
- **CI was still in progress when I passed**: `Shell tests (shellcheck + pithead suite)`, the rootfs
  build and both dashboard jobs `in_progress`; 12 green, 2 correctly `skipped` (schedule-gated CVE
  sweep). `mergeable_state: unstable` is that pendency, not a conflict. Per Finding 2, a green shell
  job would not clear Finding 1 anyway.
- **Relayed, not re-derived: the #1823 hang mechanism itself.** The on-guest evidence (BUILD_COMMIT
  `efc1c7ae`, bash parked in `wait_woken`/`n_tty_read` at 122s against a 1s timeout, released by
  stopping the getty) is yours, from a box I did not touch. I did not reproduce it and did not verify
  the #1061 / PR #1220 history. Your disclosure that it does not reproduce off-box is the right call
  and I took it at face value.
- **Not re-derived**: the `make lint-sh`, `lint-pithead-parity` and `tests/inventory.sh` claims — CI is
  the authority there.
- **Not checked**: whether the new fragment's `== ... ==` section header appears in the CI job log. A
  new domain file's green is settled from the log, not the label; that read is still owed.

### Routing

`os/overlay/` is a freeze path, so on a fixed re-push this becomes PASS + **MERGE-READY** — never a
reviewer merge and never the author's own session, and it wants sequencing with the appliance lane
since merging it moves the freeze tip and re-gates the battery.

VijitSingh97:
**RETURN addressed — re-pushed as a single amended commit (re-derive the head with `gh pr view 1824 --json headRefOid`).**

- **Finding 1 (bare keypress dropped):** taken. The primitive is now `timeout $((t+1)) bash -c 'read -r -t "$1" -N 1 k <"$2" …'` — bash's `read -N 1` keeps borrowing the console in one-character mode, the outer `timeout` owns the wall-clock bound with one second more than the inner `-t`, so the normal path is never killed and bash restores the tty itself. Measured by me on fresh ptys before the reviewer's answer landed: bare `a` → `read -N 1` delivers at once; `head -c 1` rc 124 with nothing; the wrapped read delivers at once. Same result as your table.
- **Finding 2 (control blind on a FIFO):** taken. The positive control now runs on a real pty via python's `pty` module (precedent: `test-monero-tari.sh` already uses python3 in this suite). It asserts the pty IS canonical first, then bare `a` and bare `n` through the real `media_read_key`, then that a read killed by the outer bound leaves the tty canonical — your "risk that did not materialise", now pinned as a row. Discrimination run against three primitives is in the PR body: original fails the hang legs, the `head` cut fails the bare-key legs, the shipped one passes 6/6.
- **Finding 3 (issues named as PRs):** taken; body now names #1747 and #1737.

Your owed check — the fragment's `== ... ==` header in the CI shell log — I will read from this push's job and say so here. The on-image media phase on this tree is running on the bench (5 attempts of the loop that went 1 green / 1 red on `efc1c7ae`); result goes to #1823 with the evidence path.

Merge stays with `fixes` or the seat, non-author, once your re-pass is recorded.

VijitSingh97:
**Ephemeral reviewer re-pass on `43d841aa`, relayed verbatim by the appliance lane (findings A/B are addressed in the next push):**

# Non-author RE-PASS — PR #1824 — **RETURN at `43d841aa29e17ec0ee335b43d511d32b4cf1ba64`**

Head re-derived with `gh pr view 1824 --json headRefOid` in the same call as the body: single commit
`43d841aa`, base `develop-v2`, merge-base `efc1c7ae` = the `develop-v2` tip (no divergence), 3 files
+98/-3.

**The three findings from my RETURN (5551492060) are all genuinely taken — I re-derived the
load-bearing one myself and it holds.** The RETURN is on something else and it is mechanical:
**CI is RED, and because it is red at `make lint-sh`, the pithead suite never ran on this head** —
so the new fragment, which is the entire point of the re-push, has no CI evidence at all.

---

## FINDING A (blocking) — `shfmt` reds `make lint-sh`, and the suite never executes

`Shell tests (shellcheck + pithead suite)` = **failure** at this head
(`.../commits/43d841aa/check-runs`). Read from the job log (101300859275), not the label:

- **shellcheck is CLEAN** — zero `SC####` lines in the whole log; it got past
  `shellcheck --severity=warning tests/stack/run.sh` (log 11:45:08.454).
- **`shfmt -i 4 -d` emits one hunk, one file** — `tests/stack/test-appliance-media-console.sh:52`.
  `Makefile:94: lint-sh Error 1`, `Process completed with exit code 2`, then `Post job cleanup`.

The fix is the diff shfmt already printed, +1 line:

```diff
-mc_pty=$(PITHEAD_MEDIA_SRC="$ROOT/os/overlay/pithead-media-config" python3 - <<'PY' 2>/dev/null
+mc_pty=$(
+    PITHEAD_MEDIA_SRC="$ROOT/os/overlay/pithead-media-config" python3 - <<'PY' 2>/dev/null
```

**Why this blocks rather than being a nit:** `lint-sh` is the first step of that job, so nothing
after it ran — no suite, no `tests/inventory.sh`, no parity lint. **Zero lines of suite output in
the log** (`grep -cE '== |PASS=' -> 0`). The previous push `07511119` was green, but that was a
different fragment (56 lines, FIFO-only) driving a different primitive, so its green says nothing
about this one. The fragment takes the file to 82 lines — still unbudgeted, still legal.

## FINDING B (non-blocking, fixture hygiene) — the pty leg accidentally runs the product's `main`

`tests/stack/test-appliance-media-console.sh:69` spawns
`["bash", "-c", 'source "$0"; media_read_key 2', src]`. That makes `$0` the source path, so
`os/overlay/pithead-media-config`'s guard — `[ "${BASH_SOURCE[0]}" = "${0}" ] && main "$@"` (its
last line) — is **TRUE and `main` runs** before `media_read_key`.

Measured by me, same idiom, stub file with a printing `main`, both shapes side by side:

| idiom | result |
|---|---|
| pty leg: `bash -c 'source "$0"; fn' SRC` | **`MAIN-RAN`** then `myfn-ran` |
| hang leg: `bash -c 'source /abs/path; fn'` (`$0`=`bash`) | `myfn-ran` only |

The hang leg at `:35` is correct; only the pty leg has it. Inert on a CI runner —
`main` opens with `found=$(media_find_config) || return 0` and `_removable_fat_partitions` finds no
removable FAT partition, so it returns silently. It is not inert everywhere: `_console`'s **first
loop `echo`s to stdout**, which the leg captures through `subprocess.PIPE`, and its second loop
writes to `/dev/tty1` and `/dev/ttyS0`. On any box running the suite with a FAT stick carrying
`pithead-config.json` mountable, this leg would print to the real console and then sit in a 60 s
`media_confirm_gate`, blowing `communicate(timeout=10)`. One-token fix, same push as Finding A:

```python
read_key = ["bash", "-c", 'source "$1"; media_read_key 2', "_", src]
```

---

## CLEARED — what I re-derived MYSELF at this head

**The remedy works. My own pty harness, fresh pty per leg, one variable, `ICANON` asserted on every
leg** (not the author's fixture, not my RETURN's numbers — a third independent run):

| primitive | key | canon_pre | stdout | rc | secs | canon_post |
|---|---|---|---|---|---|---|
| OLD `read -r -t 1 -N 1` | `a` / `n` | 1 | `'a'` / `'n'` | 0 | 0.3 | 1 |
| first cut `timeout 1 head -c 1` | `a` / `n` | 1 | `''` | 1 | 1.0 | 1 |
| **shipped** `timeout $((1+1)) bash -c 'read -N 1'` | `a` / `n` | 1 | **`'a'` / `'n'`** | **0** | **0.3** | 1 |
| kill leg (`timeout 1` over inner `read -t 30`) x3 | — | 1 | `''` | **124** | 1.00 | **1** |

- **Finding 1 is fixed.** A bare `n` cancels again, at once, on a canonical console. That was the
  claim the whole PR turns on — the failure direction was "applied a change the operator cancelled"
  on a gate that rewrites credentials and network settings.
- **The OLD row is the positive control** and the first-cut row is the **negative** one: the harness
  can deliver a bare key, and the regression I returned for is reproduced at rc 1/empty.
- **The kill/restore property holds 3/3** with the inner `-t 30` making the SIGTERM the only exit,
  so bash's own `-t` cannot have restored the tty. The getty's login line survives a killed read.

**Finding 2 is fixed and the new half genuinely discriminates.** The pty medium separates the two
regressions, and I checked the FIFO half still catches the *original* — driven by me against both
primitives on a no-writer FIFO:

| primitive | rc | elapsed |
|---|---|---|
| original `efc1c7ae` (`read -t`, no bound) | **124** | 8 s (killed by my outer bound) |
| shipped | 1 | 2 s |

That confirms the body's discrimination table rows 1 and 3 by measurement, not by reading it.

**Finding 3's correction is accurate.** Enumerated at source over all open PRs: the only ones
touching `tests/stack/run.sh` are **#1824, #1747, #1737** — the body now names the right two, and
they still owe 3 lines each on their own branches against the 440 ceiling.

**Checks that could have failed and did not:**

- **`run.sh` net-neutral by the gate's own counter** (`awk 'END{print NR+0}'`,
  `scripts/lint-file-budget.sh:122`, never `wc -l`): **440 at `efc1c7ae`, 440 at `43d841aa`.** The
  2-line registration stanza is paid by the blank lines after base `run.sh:255` and `:264`.
- **Fragment is 81 lines with no tsv row** and `docs/dev/file-budget.tsv` is **byte-untouched**;
  `test-appliance-media.sh` unchanged at **409 against its 409 ceiling**, so the sibling placement is
  forced, not preferred.
- **shellcheck clean** — 0 `SC####` findings in the job log; the `SC2016` directive at
  `pithead-media-config:214` sits directly above the command it covers, and the child's `$1`/`$2`
  really are its own positionals (`bash -c '…' _ "$1" "$tty"` — `_` is `$0`).
- **`[ -n "$k" ] || return 1` is not a behaviour change.** Enter alone: old code printed a newline
  that the caller's `$()` stripped to `""`; new code returns 1 and the caller's `|| key=""` gives the
  same `""`. A space key survives both. Checked because a swallowed key is the defect class here.
- **The healthy path does not get slower.** Inner `read -t 1` fires first and the child exits 1, so
  a healthy console still counts 1 s per iteration; only a contended one runs up to 2x, as disclosed.
- **Grant is real.** All three changed paths are ARMED in `roles/appliance.paths` (`os/`,
  `tests/stack/run.sh`, and `tests/stack/test-appliance-media-console.sh` on the last armed line).
- **Leak scan clean, with a FIRED control** — `detect-topology.py` on the full diff: `clean`, rc 0;
  on a seeded control line: `1 hit(s)`, rc 1. Same for this file before it is published.
- **Squash body is safe** — one commit, no self-correction inside the branch, so the default
  concatenation carries nothing a later commit refutes.

---

## What I did NOT do

- **No local suite, `make`, shellcheck, shfmt, container or KVM run** — `fleet.sh claims` shows
  a heavy job holding the bench, so this pass is CI-only. Everything above is source reading, `git`
  plumbing, the GitHub API, and sub-second Python/bash probes in `/tmp` (mode 0700, outside git).
- **Relayed, not re-derived: the #1823 hang mechanism itself.** The on-guest evidence (BUILD_COMMIT
  `efc1c7ae`, bash parked in `wait_woken`/`n_tty_read` at 122 s against a 1 s timeout, released by
  stopping the getty) is the author's, from a box I did not touch. I did not verify the #1061 /
  PR #1220 history either.
- **Could not check the fragment's `== … ==` section header in a suite log** — a new domain file's
  green is settled from the log, not the label, and here the suite never ran. That read is owed on
  the re-push and is the thing to look at first when it is green.
- **Not re-derived: `tests/inventory.sh` and `lint-pithead-parity`** — those steps never executed
  either; CI is the authority once `lint-sh` passes.
- **`Build + scan the appliance rootfs`, `Dashboard tests` and `Dashboard image` were still
  `in_progress`** when I read the rollup (12 green, 2 correctly `skipped` — the schedule-gated CVE
  sweep, which is not a measurement of this diff).
- **One residual I did not test and am not blocking on:** `timeout` sends SIGTERM with no `-k`, so
  the bound holds only while the read is interruptible. `n_tty_read`'s `wait_woken` is, which is why
  the measured kill legs return at 1.00 s. Adding `-k` would trade that for a SIGKILL that skips
  bash's tty restore — I think not adding it is the right call, but it is an unstated trade.

## Routing

`os/overlay/` is a **freeze path**, so this is never a reviewer merge and never the author's own
session. On a green re-push this becomes **PASS + MERGE-READY** for the `fixes` lane or the mac seat,
sequenced with the appliance lane — merging it moves the freeze tip and re-gates the battery.

**⚠ This verdict has not been posted to the PR.** It was written to `answers/` on request. An
ephemeral's verdict sitting only here has already read as "unreviewed" three times (#1784, #1813).
It is topology-scanned and safe to post verbatim.

VijitSingh97:
**Re-pass RETURN (A/B) addressed — re-pushed, still one amended commit (re-derive: `gh pr view 1824 --json headRefOid`).**

- **A (shfmt):** the heredoc opener is now on its own line under `mc_pty=$(`; `shfmt -i 4 -d` prints nothing on the fragment, and the previous file diffs under the same call (control fired). I had run shellcheck locally and not shfmt — my gap, and the reason the suite never ran on the last head.
- **B (`$0` idiom runs `main`):** taken — `bash -c 'source "$1"; media_read_key 2' _ SRC`. Reproduced with a printing-main stub before changing it: old idiom `MAIN-RAN myfn-ran`, new idiom `myfn-ran`.
- Three-way discrimination re-run on the new fragment: original fails the two hang legs, the `head` cut fails the two bare-key legs, shipped 6/6. Fragment is 83 lines, still unbudgeted.
- **The product file is byte-identical between the last head and this one** (`git diff 43d841aa..HEAD -- os/` is empty), so the on-image media-phase loop running against the last head's tree stands as evidence for this one: attempt 1 of 5 is green as of 11:52Z; the rest post to #1823.

Your relayed re-pass verdict is posted verbatim just above so the record shows a review happened. Non-author pass please; merge stays with `fixes` or the seat.

VijitSingh97:
**CI green on `1a728781` — the owed check is done (appliance lane, proven by me from the job log, not the label).**

The previous head's shfmt red meant the suite never ran in CI; on this head the `Shell tests (shellcheck + pithead suite)` job (run 33964564380, job 101302273784) is `success`, and its log carries the new domain file's own section:

```
== unit: media confirm gate cannot be frozen by a stuck console read (#1823) ==
  ✓ a stuck console read does not freeze the confirm gate (it returns within the 15s bound)
  ✓ with the media pulled, a stuck console read still aborts the pending change
  ✓ control: the test pty is in canonical mode, where a plain byte read would wait for Enter
  ✓ a bare 'a' on a canonical console is delivered at once by the bounded read
  ✓ a bare 'n' on a canonical console is delivered at once by the bounded read
  ✓ a read killed by the outer bound leaves the console canonical (the login prompt stays usable)
pithead tests: 3362 passed, 0 failed
```

Red glyphs in the job log: 0. Control for the grep: 226 `== unit:` headers found by the same instrument. Every other check on the head is green (the CVE sweep rows are the usual skips).

On-image proof loop on the product bytes (byte-identical between `43d841aa` and this head, `git diff -- os/` empty): attempts 1, 2 and 3 of 5 GREEN (`--phase media --keep`, each attempt a fresh boot; the same loop on the frozen tip `efc1c7ae` went 1 green / 1 RED). Attempts 4–5 are running; I will post the final tally on #1823.

The third ephemeral reviewer pass ended without posting (it waited on this job and exited). A fresh non-author pass is dispatched now with the job already green. Merge remains the `fixes` lane or the seat, never me.

VijitSingh97:
## Non-author review — **PASS** at `1a728781c8dad030c46d5676bbba0a871086711e`

Fourth pass, fresh context, nothing carried on trust. Head re-derived with
`gh pr view 1824 --json headRefOid` in this session; single commit `1a728781` on `efc1c7ae`
(= the `develop-v2` tip, no divergence), 3 files, +100/-3.

**This is a freeze-path change (`os/overlay/`), so I do not merge it. MERGE-READY for the `fixes`
lane or the seat**, on this sha.

### The amend is exactly what was claimed — proven by blob sha, not by reading the diff

`git diff --name-only 43d841aa 1a728781` returns **one path**: the test fragment. The product file
`os/overlay/pithead-media-config` is blob `215da9e9` at *both* heads, and `tests/stack/run.sh` is
blob `4076d713` at both — byte-identical objects, so findings A/B cannot have disturbed the fix.
The fragment moved `69b34146` → `f9f4f227` in two hunks: the `shfmt` reflow of the `mc_pty=$(`
subshell, and `bash -c 'source "$1"…' _ "$src"` with its why-comment.

### Finding B's fix re-derived independently (the load-bearing one)

The old idiom `bash -c 'source "$0"' "$src"` makes `$0` equal the sourced path, so
`pithead-media-config:352` — `if [ "${BASH_SOURCE[0]}" = "${0}" ]; then main "$@"; fi` — evaluates
**true** and the real `main` runs against the test host. The `_` placeholder makes it false. I did
not take the "no second effect" half on the author's or the prior pass's word: `source` keeps the
caller's positionals, so `$1` is *also* set during sourcing. Enumerated every positional reference
in all 352 lines — the only unindented hits are `:190` and `:197` (both comment text) and `:352`
itself, and the file's only other top-level statements are `:18` and the five `:22`–`:26`
assignments, none positional. `$@` at `:352` sits inside the now-false condition. **No second
effect.**

### The product change, judged on its own

`k=$(timeout "$(($1 + 1))" bash -c 'read -r -t "$1" -N 1 k <"$2" …' _ "$1" "$tty") || return 1`

- The `open()` of the tty is now inside the bound. Bash's `-t` never covered it, and on the FIFO
  fixture the open is precisely where the unfixed code parks forever.
- `local tty=… k=""` is declared on its own line, so the `|| return 1` reads the **command
  substitution's** rc, not `local`'s. The classic masking trap is avoided.
- `timeout` is already on this image on this code path (`os/overlay/pithead-boot:308`, `:415`), so
  the new dependency is measured, not assumed.
- The script is `set -uo pipefail`, no `-e` (`:18`) — no errexit interaction with the new `return 1`.

**A check that could have failed and did not, and that nobody had stated:** command substitution
strips trailing newlines, so a bare Enter now yields `k=""` → `return 1`, where the old code
returned 0 printing a newline. At the **only** call site (`:234`) `key=$(media_read_key 1) || key=""`
— the caller's own `$( )` already stripped that newline to `""`, and `""` falls through the
`a|A / n|N|c|C` case either way. Observable behaviour is identical; Enter was never a cancel and
still is not.

The disclosed cost — a contended console can stretch the 60s countdown to ~2× — is stated in the
code comment and in the body. It replaces an **unbounded** freeze, so it is strictly an improvement
on the property `docs/appliance.md:476` advertises.

### The tests are honest, and each assertion is labelled for what it is

The FIFO half is a real defect detector: with no writer, `<"$tty"` blocks at `open()`, so unfixed
the gate never reaches poll 2 and `timeout 15` returns rc 124 with an empty verdict — both
assertions red. Fixed, iteration 1 ends at ~2s and iteration 2 aborts on the stub's pulled stick.
The pty half is correctly *not* sold as a defect detector: leg 1 is a fixture control (the pty is
canonical, or the rest proves nothing), legs 2–3 are regression guards proving the `timeout` wrapper
did not lose one-character mode, and leg 4 measures the premise the getty safety rests on rather
than asserting it in prose. The fragment says so in its own comments.

### CI settled from the log, not the label

`Shell tests (shellcheck + pithead suite)`, job `101302273784`, run `33964564380` — and I closed the
chain rather than assuming it: the API reports that job's `head_sha` as `1a728781…`, matching the
head above, so the log is this head's. The log carries the new domain's own section header at line
3759 with **exactly 6 ✓ and no ✗**, and `pithead tests: 3362 passed, 0 failed` at line 3917. A green
job alone would have been equally consistent with the fragment never being collected; the header is
what rules that out. Rollup at the head: **16 success, 0 failure**, 2 `skipped` — both the
`Shipped-image CVE sweep` rows, which are schedule-gated and owed nothing here (no base-image digest
moves in this diff).

### Budget

`tests/stack/run.sh` is **440 at `efc1c7ae` and 440 at `1a728781`** by the gate's own counter
(`awk 'END { print NR + 0 }'`, not `wc -l`), against its 440 row — the ceiling holds by measurement,
and `docs/dev/file-budget.tsv` is untouched. The 83-line fragment is under 400 and correctly carries
no row.

### Two nits, neither blocking

1. The body still says *"The fragment is 81 lines"*; it is **83** after the amend. The claim it
   supports ("under 400, needs no tsv row") is unaffected, so nothing anyone would do changes.
2. The new fragment gets no row in `docs/dev/testing-strategy.md`. I measured before calling that a
   gap: **21 of 62** `tests/stack/test-*.sh` fragments are named there, so per-file rows are the
   minority, not the convention. Not a finding — recorded so my silence is not read as an unchecked
   assumption. The existing `:230` row does describe this behaviour area and could absorb a clause
   whenever that file is next touched.

PR #1825, opened 2026-09-05
test(os): #1651 a failed leg-4 install keeps the guest's own account before the harness returns

VijitSingh97:
## What

A failed dashboard-driven install (update phase, leg 4) now keeps the guest's own account before the harness returns: the root runner's own unit journal (`pithead-control.service`, 200 lines), the boot journal's rauc and OS-install lines, `rauc status`, `/data` and `/tmp` usage, memory, and any OOM kill, at `/tmp/pithead-os-serial.log.leg4-install`. The capture's ssh is capped at 120 s so it cannot hold the harness, and an empty capture is reported as empty rather than as kept. Docs: one bullet under the harness rules in `docs/dev/appliance-release.md`.

## Why

The #1651 gate on the freeze tip (BUILD_COMMIT bc685eeb) went 187/1. The one red was leg 4's install: the dashboard result said `[ERROR] rauc install failed` and nothing else. The runner deletes `.install.log` on failure and the container-visible result carries only a whitelist-extracted last line (`48-os-update-verbs.sh`), and `--phase all` destroyed the guest before anyone could read its journal. The cause is still not established; this PR makes the next red name it.

Re-derived, not relayed: the runner bounds only the lock wait (`PITHEAD_EX_LOCK_TIMEOUT`), never `rauc install` itself, so a timeout hypothesis is out.

## What was RUN

- `bash -n`, `shfmt -i 4 -d` (clean, rc 0), `shellcheck --severity=warning tests/os/run.sh` (the Makefile's severity; rc 0), `make lint-docs-voice` (OK).
- Line-neutral: `tests/os/run.sh` stays at its 3423 ceiling. The eight identical release-server teardown pairs (`kill "$srv_pid"; rm -rf "$srv"`) fold into `_leg4_srv_stop`, which reads the caller's locals by bash's dynamic scope.
- Local control of the remote command string on the bench host: 175 KB captured, and a permission error from `dmesg` landed IN the capture, proving the `2>&1` wrapping.
- `sudo tests/os/run.sh --phase update --keep` on this tree (image bytes identical to bc685eeb: `git diff bc685eeb..HEAD -- os/ dashboard/ lib/pithead/ pithead build/ docker-compose*.yml VERSION` is empty): rc 0, every leg 4 row green, so the capture path was NOT exercised by a real failure this run. The kept guest's journal gave the success baseline: the slot write took 3m14s and systemd accounted rauc.service at a 6.9 GB memory peak in the 16 GB guest.

## Head moved after the first review pass (two commits on top of the reviewed 90522db0)

- **The runner unit, verified at source, not relayed:** `lib/pithead/50-control-runner-provisioning.sh:113-121` writes `pithead-control.service` as `Type=oneshot`, `User=root`, `ExecStart` only, no `Standard*` redirection, so the runner's `warn "OS install failed (rc=…); log tail: …"` (`48-os-update-verbs.sh:306`) lands in the journal under that unit. The capture now reads `journalctl -b -u pithead-control.service -n 200` explicitly, ahead of the pattern grep. It carries the runner's 500-byte log TAIL, not the deleted install log.
- **Cannot stall the harness:** `_ssh` defaults to `timeout 5400`; `_os_step … 900` can return with rauc still `installing`, and nothing in the tree wraps a `rauc` call in a timeout, so `rauc status` could have held the harness 90 minutes. The capture call is prefixed `SSH_TIMEOUT=120`. Measured here: the prefix reaches the function (`T=5 f` prints 5); the reviewer measured separately that it does not persist after the call.
- **An empty capture says so:** the `info` is gated on `[ -s "$SERIAL.leg4-install" ]` (the file's own precedent at the console copy), otherwise it points at `$SERIAL.failed`.
- **Not done, and forced:** the capture is one 477-character line (base max 302). `tests/os/run.sh` sits at exactly its 3423 ceiling and `shfmt` expands every line-saving shape (one-line function bodies, semicolon lists), so splitting the string costs a line the budget has not got. Nothing gates line length; the reviewer checked. The capture reads the current boot only; the serial log spans a reboot.
- **Run on the final head:** `bash -n`, `shfmt -d`, `shellcheck --severity=warning` (rc 0 each), `make lint-docs-voice` OK, line count 3423 = ceiling, image-path diff to the freeze tip `bc685eeb` EMPTY (the same filter fires on the previous merge as its control). The capture script was run locally through `bash -c` as a parse control; no red has exercised it on a guest.

## Not done

- No red has run through the new capture yet; its first real exercise is the next gate red.
- The gate itself is not re-run here.

## Over-engineering pass (by hand; the PR-gate hook keys off the wrong branch from this worktree)

Pure move for the teardown: added ∖ removed = one helper definition; every one of the eight sites becomes a one-line call. Adjacent merge considered and rejected: folding the capture into `_leg4_srv_stop` would run a 300-line journal read on every leg-4 exit, including the seven that are not the install; the capture stays at the one site whose failure loses its evidence.

Closes #none — #1651 stays open until the gate is green on a named tree.

### DIFF
--- a/docs/dev/appliance-release.md
+++ b/docs/dev/appliance-release.md
@@ -242,6 +242,11 @@ Two harness rules worth knowing before you lose an afternoon to them:
 - On failure it keeps the guest console at `/tmp/pithead-os-serial.log.failed`. Read it
   first. `/dev/console` is whichever `console=` came **last** on the kernel cmdline, so
   check which one you are reading before concluding a component is silent.
+- A failed dashboard-driven install (update phase, leg 4) also keeps the guest's own
+  account at `/tmp/pithead-os-serial.log.leg4-install`: the boot journal's rauc and
+  OS-install lines, `rauc status`, disk and memory, and any OOM kill. The runner deletes
+  its install log on failure and the dashboard result carries only rauc's last line, so
+  this file is the only place the cause survives.
 
 ## The automated battery
 
--- a/docs/dev/appliance-release.md
+++ b/docs/dev/appliance-release.md
@@ -243,10 +243,11 @@ Two harness rules worth knowing before you lose an afternoon to them:
   first. `/dev/console` is whichever `console=` came **last** on the kernel cmdline, so
   check which one you are reading before concluding a component is silent.
 - A failed dashboard-driven install (update phase, leg 4) also keeps the guest's own
-  account at `/tmp/pithead-os-serial.log.leg4-install`: the boot journal's rauc and
-  OS-install lines, `rauc status`, disk and memory, and any OOM kill. The runner deletes
-  its install log on failure and the dashboard result carries only rauc's last line, so
-  this file is the only place the cause survives.
+  account at `/tmp/pithead-os-serial.log.leg4-install`: the root runner's own journal
+  (`pithead-control.service`, where its "OS install failed" warning and log tail land),
+  the boot journal's rauc and OS-install lines, `rauc status`, disk and memory, and any
+  OOM kill. The runner deletes its install log on failure and the dashboard result
+  carries only rauc's last line, so this file is the only place the cause survives.
 
 ## The automated battery
 
--- a/docs/dev/appliance-release.md
+++ b/docs/dev/appliance-release.md
@@ -246,8 +246,10 @@ Two harness rules worth knowing before you lose an afternoon to them:
   account at `/tmp/pithead-os-serial.log.leg4-install`: the root runner's own journal
   (`pithead-control.service`, where its "OS install failed" warning and log tail land),
   the boot journal's rauc and OS-install lines, `rauc status`, disk and memory, and any
-  OOM kill. The runner deletes its install log on failure and the dashboard result
-  carries only rauc's last line, so this file is the only place the cause survives.
+  OOM kill. It reads the current boot only: if the install took the guest down, the
+  serial log above spans the reboot. The runner deletes its install log on failure and
+  the dashboard result carries at most rauc's last error line, or nothing, so this file
+  is the only place the cause survives.
 
 ## The automated battery
 
--- a/tests/os/run.sh
+++ b/tests/os/run.sh
@@ -1103,7 +1103,7 @@ phase_update_dashboard() { # <good-bundle-path> <serial-byte-offset-before-this-
         bad "leg 4: install did not complete (got: $(printf '%s' "$out" | cut -c1-200))"
         # The runner deletes .install.log on a failed install and the result carries only a whitelisted last line
         # (#1651: one gate red said "[ERROR] rauc install failed", guest gone) — keep the guest's own account.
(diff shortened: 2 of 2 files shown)


VijitSingh97:
## Non-author pass — `fixes` lane. **MERGE-READY: PASS at `1eecf96cc0801e8883052467ca1865f670874895`**, conditional on the three pending checks going green.

I judged this head, not an earlier one: the recorded ephemeral-reviewer PASS is at `90522db0`, two heads back, and **this PR carries zero comments**, so that verdict lives only in an `answers/` file. Treat this comment as the pass of record.

### Re-derived by me at source

- **Base is CURRENT, not merely `MERGEABLE`.** `git merge-base` of this head and `develop-v2` equals the tip exactly.
- **Budget, by the gate's own counter (`awk 'END{print NR+0}'`, never `wc -l`):** `tests/os/run.sh` is **3423** at both base and head against its **3423** row — line-neutral at a ceiling with zero headroom, which is what the 8-site teardown refactor buys. `docs/dev/file-budget.tsv` is untouched; the PR changes exactly two files.
- **Freeze filter EMPTY over `os/ lib/pithead/ pithead build/ docker-compose*.yml dashboard/ VERSION`, with a FIRED control** — the same filter over the preceding commit returns `os/overlay/pithead-media-config`, so it is able to fire. Image bytes are unchanged by this PR.
- **The refactor is behaviour-preserving.** `_leg4_srv_stop` occurs 9 times: one definition plus **8 call sites, every one inside `phase_update_dashboard()`**, and all downstream of the `srv`/`srv_pid` assignments at `:962`/`:973`. The one failure path *before* that assignment keeps its bare teardown rather than calling the helper — correct, and it is exactly the site this refactor could have broken. Under `set -uo pipefail` that distinction is load-bearing.
- **The `[ -s ]` gate is the right instrument, not decoration.** The capture redirects into `$SERIAL.leg4-install` and is `|| true`, so the file is created even when the capture fails; without the gate the harness would claim diagnostics were kept when the file is empty.
- **The fallback's pointer is accurate.** It sends the reader to `$SERIAL.failed`, and `run.sh:403-407` copies `$SERIAL` there at run end whenever `FAIL > 0`, so on this path that file does come to exist.
- **`A && B || C` is not a lint or correctness risk here.** The same construct appears 22 times in this file at base — `:534-535` is the identical `cmd && ok … || bad …` shape — and `info()` is a bare `printf`, so the "B fails, therefore C also runs" hazard is not live.

### Relayed, NOT re-derived by me

The author's local `bash -n` / `shfmt` / `shellcheck --severity=warning` / `lint-docs-voice` rc 0, and the ephemeral reviewer's second PASS at `3d5dc361` (reported by message, the pad having exited).

### The merge condition, and it is NOT met as I write

`check-runs` at this head: **total_count 17 — 12 success, 2 skipped, 3 `in_progress`, ZERO failing.** The three pending include **`Shell tests (shellcheck + pithead suite)`**, which is both the suite run and the only independent confirmation of the shellcheck result above. Read them as `check-runs` per-step; the legacy commit status sits `pending` forever in this repo and must not be used as the gate.

I am out of cycle before those complete, so **the tip has NOT moved and this is not merged.** Whoever picks it up: confirm the three are green *at this same head*, re-read this PR's comments in the call that merges (a late RETURN moves no head and leaves no tell), then squash-merge. **Do not hand-close #1651** — it stays open until the gate is green on a named tree.

Issue #1822, opened 2026-09-05
the 2.0.0 CHANGELOG omits every user-visible fix merged since the release branch was cut


VijitSingh97:
`CHANGELOG.md` on `release/2.0.0-prep` was written against a tree that is now 27 commits behind `develop-v2`, and its `## [2.0.0]` section does not mention any of the user-visible fixes that landed in those 27 commits. Cutting GA from that branch as it stands ships a release note that omits eight fixes an operator would look for, two of them the kind people check a changelog specifically to find.

### Measured

Base of the release branch: `b4d818ec`. Tip of `develop-v2` at the time of writing: re-derive with `git ls-remote origin develop-v2` — the branch is 27 commits behind it, and its last push was 2026-09-04T19:02:50Z.

Of the 53 distinct issue references in those 27 commit subjects, **51 do not appear anywhere in the release branch's `CHANGELOG.md`**. The two that do (`#1213`, `#185`) are older issues the new commits merely cite.

Most of the 51 are test-only or docs-only changes that a curated changelog is right to leave out. These eight are not — each is a `fix(...)` commit with operator-visible behaviour, and each is absent:

| Issue | What merged |
|---|---|
| #1772 | doctor's stratum-exposure warning stopped naming the host's public address |
| #1695 | the out-of-band audit cap bounds the device-chosen name space |
| #1696 | the worker-config store refuses a bind sqlite cannot encode |
| #1769 | an appliance operator is no longer shown host remedies it cannot reach |
| #1776 | doctor's missing-data-dir warn is worded for the surface the operator is on |
| #1791 | the persistent-journal bind is ordered after the /var overlay |
| #1817 | a rig's journal reclaim tolerates the persistent bind, and the bind stands down on a rig |
| #1420 | the aggregate counters get floors, not a zero test |

### How it was checked, and the control

Both reference forms were searched, because the file cites issues as `[#N](…/issues/N)`: a word-bounded `#N` and a word-bounded `issues/N`. All eight score **0 on both**.

The instrument is not a dead needle: the same two greps score **1 and 1** for `#797`, `#924`, `#910` and `#965`, which the `## [2.0.0]` section does carry, and the `#N` sweep finds **392** distinct references in the file overall. A zero here is a real absence rather than a needle that cannot match.

### Not checked

[...]

Issue #593, opened 2026-07-17
Add opt-in local miner: co-locate a RigForge worker on the stack host


VijitSingh97:
## Summary

Add an **opt-in local miner**: co-locate a RigForge worker on the stack host so a box already
running the stack 24/7 can mine with its spare CPU. Reuse RigForge (compiles/tunes stock XMRig, runs
it as a systemd service, owns the host-level tuning) rather than hand-rolling xmrig in the stack.

Companion RigForge issue (the required upstream change): **p2pool-starter-stack/rigforge#305**

Related: **#595** — reducing Tor's CPU overhead frees the very core a co-located miner wants (Tor is
single-threaded and is the stack's top CPU consumer).

> Re-scoped after tracing RigForge's setup flow and grepping this repo. Earlier this issue said the
> stack owns the HugePage reservation and must remain its sole owner — that was wrong on both counts
> (see Design). Ownership sits with RigForge.

## Why this is feasible today

RigForge and Pithead already share a worker contract; "same box" is just that contract with the pool
URL set to `127.0.0.1:3333`.

- The proxy's stratum port is **published to the host** by default — `"${STRATUM_BIND:-0.0.0.0}:${STRATUM_PORT:-3333}:${STRATUM_PORT:-3333}"` (`docker-compose.yml:446`) — so a host-level miner reaches it at `127.0.0.1:3333`. Loopback: no LAN exposure, no firewall rule, no Tor.
- The dashboard already discovers workers from the proxy connection list and reads their `:8080` API, so a local rig self-registers with no stack-side wiring.
- Pithead already holds the two values a manual RigForge install would prompt for: the pool URL (its own `:3333`) and the stratum secret (`PROXY_STRATUM_PASSWORD`, already generated in `.env`). That's what makes the config seamless — the user copies nothing.

## Design: RigForge owns the reservation; Pithead delegates

Pithead has **no HugePage/GRUB code** (verified — nothing in this repo touches `hugepages`,
`nr_hugepages`, or `/etc/default/grub`) and should not gain any. RigForge already writes and merges
the reservation correctly and idempotently, so it stays the single owner of all host-level tuning
(HugePages, GRUB, MSR, governor, systemd service). Pithead's job is purely to **declare intent and
hand off**:

1. **Opt-in config + prompt.** New `local_miner.*` keys (default off) and a `pithead setup` prompt:
[...]

PR #1833, opened 2026-09-05
fix(dashboard): #1830 the wizard opens on Pithead + RigForge, and the select IS the miner switch

### LINKED ISSUE #1830: Setup wizard: default the role select to Pithead + RigForge (today: Pithead)
## Operator ruling (2026-09-05)

The setup wizard's role select defaults to **Pithead + RigForge**. Today it defaults to **Pithead** alone: `dashboard/mining_dashboard/web/static/wizard.mjs:266` initialises `role: "pithead"`, the `<select>` at `:646-650` offers `pithead` / `both` / `rig`, and `docs/appliance.md:139` marks the Pithead row "(default)". No test pins the default.

## Why

Most people who boot this image on one machine want it to mine with that machine's CPU as well as coordinate; the middle choice is that machine. The other two roles stay exactly as they are; only the pre-selected option changes.

## What to change, and what to audit while there

- `wizard.mjs:266` `role: "pithead"` → `"both"`. The `setRole` handler and the Note text need no change.
- `docs/appliance.md:139-141`: move "(default)" to the **Pithead + RigForge** row; re-read the sentence at `:140` ("Same as turning on 'Mine with this machine's CPU' below") so the two statements about the built-in miner still agree.
- **Audit every place that assumes the default is `pithead`** rather than reading the select: any server-side fallback when `role` is absent from a submitted or pre-seeded config (installer pre-seed, media channel, restore-at-setup — a restored backup carries its own role and must win), and any KVM battery row in the provision phase that submits the wizard without naming a role. If a fallback exists it must agree with the wizard's default or state why it differs.
- A frontend logic test that pins the initial `role` state, so the default cannot drift silently again.

## Scope

Image bytes (the wizard image is baked from `dashboard/`), so this lands inside the current freeze move together with the RigForge repin and rides that gate. No change to what any role does.

### LINKED ISSUE #1831: Setup wizard: 'Mine on this machine too?' does not update 'What is this machine?' — the two controls describe one fact and disagree
## What the operator saw (manual test of the 2026-09-05 image)

In the setup wizard, "What is this machine?" and "Mine on this machine too?" describe one fact from two controls, and they only agree in one direction. Changing the role select updates the miner switch (Pithead + RigForge → yes; Pithead → no). Changing the miner switch does **not** update the role select: pick Pithead, then answer yes to mining on this machine, and the form still says "Pithead" while the config it submits is exactly what "Pithead + RigForge" means.

## Mechanism, at source (`dashboard/mining_dashboard/web/static/wizard.mjs`, develop-v2)

- `setRole` (`:378-391`) presets `local_miner.enabled` to `role === "both"` — role drives the switch.
- The switch (`:753-756`) is a plain field bound to `localMiner: { path: "local_miner.enabled" }` (`:38`) through the generic `on(...)` handler, which writes the config and never touches `state.role`.
- `editJson` (`:395`) can also set `local_miner.enabled`; the role select is not re-derived there either.
- `state.role` is not part of the submitted config for coordinators (the comment at `:264`: "both" rides the existing local_miner switch), so the disagreement is display-only — but the display is the operator's only statement of what the machine will be.

## Fix shape

One source of truth. For the two coordinator roles, derive what the select shows from the config rather than storing it: `local_miner.enabled` true → "Pithead + RigForge", false → "Pithead"; keep a stored role only for `rig`, which has no config. Then both controls, and the JSON pane, agree by construction, and the Note's sentence ("The middle choice is a Pithead that also mines with its own CPU") becomes literally true. Frontend logic test: toggling the switch flips the select; editing the JSON flips both; choosing rig collapses the form as today.

Same file and same PR as #1830 (the default moves to Pithead + RigForge): the default and the sync are one change to how this control works.

## Not asked here

Whether two controls should exist at all is a wording question for the wizard's owner; this issue only asks that they never disagree.

VijitSingh97:
Closes #1830. Closes #1831. One change to how one control works, so one PR — the issues ask for
that explicitly.

## What changed

The setup wizard's role select now opens on **Pithead + RigForge**, and the two coordinator
answers are read back out of the config instead of being stored beside it. The select, the
"Mine on this machine too?" switch and the JSON pane are three views of `local_miner.enabled`
and cannot disagree.

- `wizard.mjs` — the select's value is `this.state.role || (v("localMiner") ? "both" : "pithead")`.
  Only `rig` is stored, because a rig carries no config at all and nothing else can express it;
  `setRole` stores nothing for either coordinator answer.
- `wizard.py` — `wizard_state` publishes `_deep_merge(ref, _last_attempt() or {"local_miner":
  {"enabled": True}})`. `config.reference.json` is untouched: the documented default stays off for
  the CLI wizard and every existing install.
- `docs/appliance.md` — "(default)" moves to the **Pithead + RigForge** row; the settings table's
  built-in-miner row moves to `on` and now quotes the label the page actually uses (it said
  "Mine with this machine's CPU"; the page says "Mine on this machine too?").

## The one design decision, and why it went this way

The default could have been layered *under* the last attempt. It must not be. `last-attempt.json`
is the channel for **three** different things, every one of which has to win whole:

| Channel | Writer |
|---|---|
| operator pre-seed from the medium | `lib/pithead/12-firstboot-wizard.sh:162` |
| reinstall pre-fill read off the target disk | `lib/pithead/10-installer-preseed.sh:220` (`prefill_from_previous_install`) |
| a rejected or previous submission | `dashboard/mining_dashboard/wizard.py:474` |

`strip_defaults` drops `local_miner` entirely when it is off, so a config from any of those three
that means "Pithead only" says so by **absence**. Merged underneath, this page's default would
have read a deliberate Pithead-only machine as "Pithead + RigForge" and switched its miner on
behind the operator on a reinstall. `_last_attempt() or {...}` makes that impossible by
construction, and the second server row below is the control that pins it.

## The audit the issue asked for

- **Server-side role fallbacks: there are none to fix.** `role` is read in exactly three places
  (`wizard.py:413`, `:451`, `:551`) and every one tests for `rig`. Absent `role` means
  "coordinator", and what a coordinator mines is `local_miner.enabled` in the config it submits —
  there is no second statement of the default to keep in step.
- **Restore-at-setup** goes through `/submit-restore` and never reads the published config. The
  restored config wins, unchanged.
- **Pre-seed and the reinstall pre-fill** are two of the three `last-attempt.json` channels above,
  so both win whole, by the decision above.
- **KVM provision rows: unaffected, and I checked every one.** Each `/submit` in `tests/os/run.sh`
  (`:780`, `:1255`, `:1574`, `:1678`, `:2012`, `:2787`, `:3176`) posts form **fields**, which go
  through `build_config`, not the published config — so none of them can see this default. `:2012`
  already names `local_miner=true` for the Both leg and `:2787` is the rig leg. The one row that
  reads `/api/wizard-state` is the reinstall pre-fill check at `:1422`, and it runs with a
  last-attempt present, so the default is suppressed there.

## Known gap, named rather than hidden

**The no-JS form path keeps today's behaviour.** `build_config` (`wizard.py:311`) sets
`local_miner` only when the submitted form says so, so a client that posts raw fields without
reading the published config — the KVM harness's curl, a text browser — still provisions without
the built-in miner. That is defensible on its own terms (an explicit field submission is an
explicit answer) and I deliberately did not change it: it would flip four provisioning legs of the
gating KVM battery to run a miner, which is not something to slip into this PR.

## Both new test files are SIBLINGS, because the budget gate leaves no room

`dashboard/tests/frontend/wizard.test.mjs` (765/765) and `dashboard/tests/web/test_wizard.py`
(974/974) are both AT their recorded ceilings, and ceilings only ever go down — so the rows go in
`tests/frontend/wizardrole.test.mjs` (82 lines) and `tests/web/test_wizard_role.py` (73), each
under the 400-line target and so needing no budget row. Neither production file grew either:
`wizard.mjs` is 889/889 and `wizard.py` 674/674 after this change, same as before it.
`docs/dev/file-budget.tsv` is therefore **not touched at all**.

## What was RUN

- `node --test dashboard/tests/frontend/` — **543 pass, 0 fail** (4 new).
- `uv run --locked --extra test python -m pytest` over `dashboard/` — **green, 97% total**, gate 80%.
- `bash tests/inventory.sh` — rc 0 (the CI drift gate; both new files are picked up by its globs).
- **`make lint`: 12 of the 13 targets green, `lint-sh` NOT RUN — and I am naming it rather than
  reporting a green.** Under `flock ~/.lint-sh-manual.lock` it died twice with
  `shellcheck: exceeded the 6G cgroup cap` (Error 137) — the #1206 class on this box, not a
  finding about this branch: the diff contains **no shell file at all**. I then ran the
  Makefile's own `lint:` list minus that one target —
  `lint-py lint-js lint-yaml lint-md lint-docs-voice lint-operator-strings lint-topology
  lint-file-budget lint-pithead-parity lint-trivy-parity lint-proto lint-toml` — **rc 0**,
  `lint-file-budget`'s own self-tests firing among them. CI's shellcheck is the authority here.
- `make test-patch-coverage` — **rc 0, and it is the vacuous shape (#1000), which its guard caught
  correctly**: `No lines with coverage information in this diff`, then
  `all 1 changed file(s) under the measured tree are present in coverage.xml`. The reason is real
  rather than a broken measurement — the only changed Python line is a value line inside a
  multi-line dict literal, plus a docstring, and coverage.py records neither as a statement. The
  statement they belong to is covered: `wizard.py`'s missing-line list does not name it.
- **Mutation battery — 4 mutations, 4 distinct redden-sets, no vacuous row.** Each printed the
  substitution it made, each proved it applied (the file's sha256 MOVED), and each file was
  restored and its sha256 re-checked afterwards:

  | Mutation | Reddens |
  |---|---|
  | server default removed (`_last_attempt() or {...}` -> `_last_attempt()`) | server row 1 only — row 2 stays green, which is what shows the instrument can read a `False` |
  | the select stops reading the config (`\|\| (v("localMiner") ? …)` -> `\|\| "pithead"`) | frontend rows 1 and 3 |
  | the select answers `"both"` regardless | frontend rows 2, 3 and 4 |
  | `setRole` stores nothing at all | frontend row 4 only |

  The third exists only so that row 2 — the Pithead control — is not a row that has never been
  shown able to fail.
- `~/dev/fleet/lane-guard.sh` **rc 0** on the staged paths, **with a firing control**: staging a
  one-byte change to `lib/pithead/14-local-miner.sh` alongside them returned **rc 1** naming that
  file and only that file, so the rc 0 means "granted", not "the guard waves everything through".
  The control file was restored and its sha256 re-checked.

## Over-engineering pass (the branch gate's, run by hand)

No changes came out of it. The derivation is one expression reusing the existing `v()` accessor —
no new method; `wizard.py` gains no constant and no helper; the three comment blocks rewritten in
`wizard.mjs` had each become FALSE ("the role presets it", "byte-for-byte today's"), so they are
repair, not scope. Six test rows for a two-line change is the one thing worth defending, and each
row is pinned by a distinct mutation above. **Two duplications stand deliberately**: each new test
file carries a small local harness (a 3-line `setState` stub; a `spool`/`client` fixture pair)
because the sibling files own theirs module-locally and are at their ceilings, so exporting from
them is not available.

## Disclosures

- `docs/appliance.md` is in `_shared.paths` — no lane owns `docs/`; disclosed here as the law asks.
- **No CHANGELOG entry.** `develop-v2`'s CHANGELOG has no `[Unreleased]` section: the `## [2.0.0]`
  section lives in the open release-prep pull 1774, so a competing section here would collide with
  it. This change is one more commit that section was written without, which is already the open
  finding in #1822.
- The new frontend file carries its own three-line harness rather than importing
  `wizard.test.mjs`'s: that file's helpers are module-local, and its shared `REF` object literal
  is mutated in place by earlier tests through `pathSet`, so a fixture built from it would carry a
  previous test's `local_miner`.

### DIFF
--- a/dashboard/mining_dashboard/web/static/wizard.mjs
+++ b/dashboard/mining_dashboard/web/static/wizard.mjs
@@ -752,8 +752,8 @@ export class WizardApp extends Component {
             change later.<//>
             <${Field} label="Mine on this machine too?">
                 <select value=${String(v("localMiner") ?? false)} onChange=${on("localMiner")}>
-                    <option value="false">No — this box only coordinates the miners (default)</option>
-                    <option value="true">Yes — this machine also mines with its own CPU (built-in RigForge)</option>
+                    <option value="false">No — this box only coordinates the miners</option>
+                    <option value="true">Yes — this machine also mines with its own CPU (built-in RigForge, default)</option>
                 </select>
             <//>
             ${
--- a/dashboard/mining_dashboard/web/static/wizard.mjs
+++ b/dashboard/mining_dashboard/web/static/wizard.mjs
@@ -260,10 +260,9 @@ export class WizardApp extends Component {
     jsonText: "",
     jsonError: "",
     authMode: "auto", // auto | set | none — travels beside the config (see wizard.py submit)
-    // What this machine IS — the first disclosure, above the disk. pithead | both | rig.
-    // "both" rides the existing local_miner switch; "rig" collapses the form to three answers
-    // that travel beside the config exactly like authMode does.
-    role: "pithead",
+    // What this machine IS, asked above the disk. ONLY "rig" is stored — it is the one answer
+    // that carries no config; the two coordinators are read back out of one (see renderSetup).
+    role: "",
     rigPool: "",
     rigWorker: "",
     rigPassword: "",
@@ -372,13 +371,12 @@ export class WizardApp extends Component {
   };
 
   // The role reshapes the page the way the disk choice does. "Both" IS the existing
-  // local_miner switch — the role presets it and the switch below stays live; back to plain
-  // Pithead resets it to the documented default so the submitted config is byte-for-byte
-  // today's. The rig role never touches the config at all.
+  // local_miner switch: picking it turns that switch on, plain Pithead turns it off, and the
+  // switch below stays live either way. Neither is stored; a rig never touches the config.
   setRole = (e) => {
     const role = e.target.value;
     const cfg = this.state.cfg;
-    const next = { role, cfg };
+    const next = { role: role === "rig" ? "rig" : "", cfg };
     if (role !== "rig") {
       pathSet(cfg, "local_miner.enabled", role === "both");
       // "usb" only exists for rigs — a coordinator switching back must re-pick a real disk.
@@ -598,7 +596,9 @@ export class WizardApp extends Component {
     const remoteMonero = v("moneroMode") === "remote";
     const remoteTari = v("tariMode") === "remote";
     const { installer, disks, chosen, confirm, wipe, dataWiped } = this.state;
-    const rig = this.state.role === "rig";
+    // The select is stored for a rig and read out of the config for the two coordinators.
+    const role = this.state.role || (v("localMiner") ? "both" : "pithead");
+    const rig = role === "rig";
     // Keep-everything reinstall: the machine's settings, wallets, login and chains all survive,
     // so there is nothing to ask — the config half of the page would collect answers the
     // machine will ignore (its preserved config wins). Only the disk half renders.
@@ -643,7 +643,7 @@ export class WizardApp extends Component {
         <${Err}>${error}<//>
         <form onSubmit=${this.submit}>
             <${Field} label="What is this machine?">
-                <select value=${this.state.role} onChange=${this.setRole}>
+                <select value=${role} onChange=${this.setRole}>
                     <option value="pithead">Pithead</option>
                     <option value="both">Pithead + RigForge</option>
                     <option value="rig">RigForge</option>
--- a/dashboard/mining_dashboard/wizard.py
+++ b/dashboard/mining_dashboard/wizard.py
@@ -229,10 +229,10 @@ async def auth(request: web.Request) -> web.Response:
 
 
 async def wizard_state(request: web.Request) -> web.Response:
-    """Everything the SPA needs to render: which flow, the effective config (defaults with the
-    last attempt merged over them — a rejected config comes back for editing rather than making
-    someone retype a 95-character address), the reference for type coercion, the host's error,
-    and the disk inventory in installer mode."""
+    """Everything the SPA needs to render: which flow, the effective config (the defaults with a
+    previous configuration merged over them — a pre-seed, a reinstall pre-fill or a rejected
+    submission, each of which WINS WHOLE; a machine that has never been configured gets this
+    page's own default instead), the reference, the host's error, and the disk inventory."""
     if not _authed(request):
         return web.json_response({"error": "unauthenticated"}, status=401)
     ref = _reference()
@@ -243,7 +243,7 @@ async def wizard_state(request: web.Request) -> web.Response:
             "stage": stage,
             # Kept for the field's original meaning; `stage` is what the client renders from.
             "mode": "installer" if installer_mode() else "setup",
-            "config": _deep_merge(ref, _last_attempt()),
+            "config": _deep_merge(ref, _last_attempt() or {"local_miner": {"enabled": True}}),
             "reference": ref,
             "error": _spool_read("error.txt"),
             "disks": _disks(),
--- a/dashboard/tests/frontend/wizard.test.mjs
+++ b/dashboard/tests/frontend/wizard.test.mjs
@@ -400,7 +400,7 @@ test("submit validates IN PLACE: no view swap, the button narrates, the server m
   restore();
 });
 
-test("the mine-on-this-box choice is a labeled select naming RigForge, default No", async () => {
+test("the mine-on-this-box choice is a labeled select naming RigForge, default Yes", async () => {
   const { inst, restore } = await appOn([stateFor("setup")]);
   const out = renderToString(inst.render());
   assert.match(out, /Mine on this machine too\?/);
--- a/dashboard/tests/frontend/wizardrole.test.mjs
+++ b/dashboard/tests/frontend/wizardrole.test.mjs
@@ -1,0 +1,82 @@
+// The setup wizard's role select (mining_dashboard/web/static/wizard.mjs): what an unconfigured
+// machine opens on (#1830), and the single fact it shares with the built-in-miner switch
+// (#1831). A sibling of wizard.test.mjs rather than more rows inside it: that file sits at its
+// recorded ceiling in docs/dev/file-budget.tsv, and ceilings only go down.
+//
+// Run with Node's built-in test runner (CI runs exactly this):
+//     node --test dashboard/tests/frontend/
+import assert from "node:assert/strict";
+import { test } from "node:test";
+
+import { WizardApp } from "../../mining_dashboard/web/static/wizard.mjs";
+import { renderToString } from "./helpers/render.mjs";
+
+// The published reference as a real machine has it: config.reference.json carries
+// local_miner.enabled and documents it OFF (#593), which is what the CLI wizard keeps.
+const REF = {
+  monero: { wallet_address: "", prune: true },
+  p2pool: { pool: "mini" },
+  local_miner: { enabled: false },
+};
+const cfgFor = (enabled) => ({
+  monero: { wallet_address: "", prune: true },
+  p2pool: { pool: "mini" },
+  local_miner: { enabled },
+});
+
+// A mounted wizard sitting on a given config. The server round trip is deliberately not driven
+// here — loadState's own behaviour is asserted in wizard.test.mjs, and what the SERVER publishes
+// for an unconfigured machine is pinned at its own tier in tests/web/test_wizard_role.py.
+function appOn(cfg) {
+  const inst = new WizardApp({});
+  inst.setState = (patch) =>
+    Object.assign(inst.state, typeof patch === "function" ? patch(inst.state) : patch);
+  Object.assign(inst.state, {
+    stage: "setup",
+    cfg,
+    reference: REF,
+    jsonText: JSON.stringify(cfg, null, 2),
+  });
+  return inst;
+}
+
+test("a machine whose config has the miner on opens on Pithead + RigForge (#1830)", () => {
+  const out = renderToString(appOn(cfgFor(true)).render());
+  assert.match(out, /<select value="both"/); // the select
+  assert.match(out, /Nothing to install/); // the switch's own Yes note — the same fact
+});
+
+test("a machine whose config keeps the miner off opens on Pithead (#1830 control)", () => {
+  // The control that shows the row above can say something else. It is also the real case: a
+  // pre-seed, a reinstall pre-fill and a rejected submission all arrive with their own config
+  // and win whole over the page's default (tests/web/test_wizard_role.py pins that half).
+  const out = renderToString(appOn(cfgFor(false)).render());
+  assert.match(out, /<select value="pithead"/);
+  assert.doesNotMatch(out, /Nothing to install/);
+});
+
+test("the miner switch moves the role select, and so does the JSON pane (#1831)", () => {
+  const inst = appOn(cfgFor(false));
+  // The switch is the plain field edit the rendered <select> is bound to (FIELDS.localMiner).
+  inst.edit("local_miner.enabled")({ target: { value: "true" } });
+  assert.match(renderToString(inst.render()), /<select value="both"/);
+  inst.edit("local_miner.enabled")({ target: { value: "false" } });
+  assert.match(renderToString(inst.render()), /<select value="pithead"/);
+  // Hand-edited JSON wins the same way, because it is the same one fact.
+  inst.editJson({ target: { value: JSON.stringify(cfgFor(true)) } });
+  assert.match(renderToString(inst.render()), /<select value="both"/);
+});
+
+test("picking rig IS stored, and outranks whatever the config's miner switch says (#1831)", () => {
+  // A rig carries no config at all, so it is the one answer that cannot be read out of one.
+  const inst = appOn(cfgFor(true));
+  inst.setRole({ target: { value: "rig" } });
+  assert.equal(inst.state.role, "rig");
+  assert.match(renderToString(inst.render()), /<select value="rig"/);
+  assert.equal(inst.state.cfg.local_miner.enabled, true); // untouched: a rig never edits config
+  // Back to a coordinator: nothing is stored, and the config answers again.
+  inst.setRole({ target: { value: "pithead" } });
+  assert.equal(inst.state.role, "");
+  assert.equal(inst.state.cfg.local_miner.enabled, false);
+  assert.match(renderToString(inst.render()), /<select value="pithead"/);
+});
--- a/dashboard/tests/frontend/wizardrole.test.mjs
+++ b/dashboard/tests/frontend/wizardrole.test.mjs
@@ -61,19 +61,22 @@ test("the (default) marker names the option an unconfigured machine opens on (#1
   // wizard.test.mjs asserts option text that survives moving the marker back, so a straight
   // revert of the fix ships green (measured by the non-author reviewer at this head).
   const out = renderToString(appOn(cfgFor(true)).render());
-  // Structure, not wording: read each option of THIS select and ask which one carries a default
-  // marker. Matching prose literals was BOTH too narrow and too brittle — it missed a marker
-  // duplicated onto No in the comma form, and it red on a pure reword of either label. Both
-  // needles are unique to their option; four other markers elsewhere on the page are why this is
-  // per-option rather than a page-wide count.
-  const label = (needle) => {
-    const m = out.match(new RegExp(`<option [^>]*>([^<]*${needle}[^<]*)<`));
-    assert.ok(m, `no rendered option matched ${needle}`); // a clean failure, not a TypeError
-    return m[1];
-  };
-  assert.match(label("also mines"), /\bdefault\b/);
+  // Structure, and keyed by VALUE rather than by prose: scope to this one select, then ask of
+  // each option what its value attribute is. Keying on label literals pinned the marker one
+  // indirection short of the thing — it stayed green when the two options' value attributes were
+  // swapped, and when the select's own binding was inverted, each of which puts the marker on
+  // the option the operator does not get, which is #1830 verbatim.
+  const sel = out.match(
+    /Mine on this machine too\?[\s\S]*?<select value="([^"]*)"([\s\S]*?)<\/select>/,
+  );
+  assert.ok(sel, "the mine-on-this-machine select did not render");
+  const opt = Object.fromEntries(
+    [...sel[2].matchAll(/<option value="([^"]*)"[^>]*>([^<]*)</g)].map((m) => [m[1], m[2]]),
+  );
+  assert.equal(sel[1], "true"); // the option an unconfigured machine actually opens on
+  assert.match(opt.true, /\bdefault\b/);
   // The load-bearing half: matching on Yes alone still passes if the marker is re-added to No.
-  assert.doesNotMatch(label("only coordinates"), /\bdefault\b/);
+  assert.doesNotMatch(opt.false, /\bdefault\b/);
 });
 
 test("the miner switch moves the role select, and so does the JSON pane (#1831)", () => {
--- a/dashboard/tests/frontend/wizardrole.test.mjs
+++ b/dashboard/tests/frontend/wizardrole.test.mjs
@@ -55,6 +55,17 @@ test("a machine whose config keeps the miner off opens on Pithead (#1830 control
   assert.doesNotMatch(out, /Nothing to install/);
 });
 
+test("the (default) marker names the option an unconfigured machine opens on (#1830)", () => {
+  // wizard.py:246 publishes local_miner.enabled true for a machine with no previous config, so
+  // the marker belongs on Yes. Nothing else on this page pins any marker's POSITION: the row in
+  // wizard.test.mjs asserts option text that survives moving the marker back, so a straight
+  // revert of the fix ships green (measured by the non-author reviewer at this head).
+  const out = renderToString(appOn(cfgFor(true)).render());
+  assert.match(out, /Yes — this machine also mines[^<]*, default\)/);
+  // The load-bearing half: matching on Yes alone still passes if the marker is re-added to No.
+  assert.doesNotMatch(out, /coordinates the miners \(default\)/);
+});
+
 test("the miner switch moves the role select, and so does the JSON pane (#1831)", () => {
   const inst = appOn(cfgFor(false));
   // The switch is the plain field edit the rendered <select> is bound to (FIELDS.localMiner).
--- a/dashboard/tests/frontend/wizardrole.test.mjs
+++ b/dashboard/tests/frontend/wizardrole.test.mjs
@@ -61,9 +61,19 @@ test("the (default) marker names the option an unconfigured machine opens on (#1
   // wizard.test.mjs asserts option text that survives moving the marker back, so a straight
   // revert of the fix ships green (measured by the non-author reviewer at this head).
   const out = renderToString(appOn(cfgFor(true)).render());
-  assert.match(out, /Yes — this machine also mines[^<]*, default\)/);
+  // Structure, not wording: read each option of THIS select and ask which one carries a default
+  // marker. Matching prose literals was BOTH too narrow and too brittle — it missed a marker
+  // duplicated onto No in the comma form, and it red on a pure reword of either label. Both
+  // needles are unique to their option; four other markers elsewhere on the page are why this is
+  // per-option rather than a page-wide count.
+  const label = (needle) => {
+    const m = out.match(new RegExp(`<option [^>]*>([^<]*${needle}[^<]*)<`));
+    assert.ok(m, `no rendered option matched ${needle}`); // a clean failure, not a TypeError
+    return m[1];
+  };
+  assert.match(label("also mines"), /\bdefault\b/);
   // The load-bearing half: matching on Yes alone still passes if the marker is re-added to No.
-  assert.doesNotMatch(out, /coordinates the miners \(default\)/);
+  assert.doesNotMatch(label("only coordinates"), /\bdefault\b/);
 });
 
 test("the miner switch moves the role select, and so does the JSON pane (#1831)", () => {
--- a/dashboard/tests/web/test_wizard_role.py
+++ b/dashboard/tests/web/test_wizard_role.py
@@ -1,0 +1,73 @@
+"""What the wizard server publishes as an unconfigured machine's starting point (#1830).
+
+A sibling of ``test_wizard.py`` rather than more rows inside it: that file sits at its recorded
+ceiling in ``docs/dev/file-budget.tsv``, and ceilings only go down. Its ``spool``/``client``
+fixtures are module-local, so this module carries its own copies.
+
+The half the operator SEES — the role select reading these back — is pinned where the dashboard
+tests its frontend: ``node --test`` over ``tests/frontend/wizardrole.test.mjs``.
+"""
+
+import json
+
+import pytest
+from aiohttp.test_utils import TestClient, TestServer
+
+from mining_dashboard import wizard
+
+
+@pytest.fixture
+def spool(tmp_path, monkeypatch):
+    sd = tmp_path / "spool"
+    sd.mkdir()
+    monkeypatch.setenv("WIZARD_SPOOL", str(sd))
+    monkeypatch.setenv("WIZARD_TOKEN", "pit-X7KM2Q")
+    # The published reference as a real machine has it: local_miner is a documented key and it
+    # is documented OFF (#593). Without it here BOTH rows below could pass on a missing key.
+    sd.joinpath("config.reference.json").write_text(
+        json.dumps(
+            {
+                "monero": {"wallet_address": "", "mode": "local", "prune": True},
+                "p2pool": {"pool": "mini"},
+                "local_miner": {"enabled": False},
+            }
+        )
+    )
+    return sd
+
+
+@pytest.fixture
+async def client(spool):
+    c = TestClient(TestServer(wizard.make_app(exit_fn=lambda code: None)))
+    await c.start_server()
+    yield c
+    await c.close()
+
+
+async def _state(client):
+    await client.post("/auth", data={"token": "pit-X7KM2Q"}, allow_redirects=False)  # noqa: S106
+    return await (await client.get("/api/wizard-state")).json()
+
+
+async def test_an_unconfigured_machine_starts_with_the_built_in_miner_on(client):
+    # The page's own default, not the product's: the role select opens on "Pithead + RigForge",
+    # and that role IS this switch. The reference it is published beside still reads off, which
+    # is what the CLI wizard and every existing install keep.
+    s = await _state(client)
+    assert s["config"]["local_miner"]["enabled"] is True
+    assert s["reference"]["local_miner"]["enabled"] is False
+
+
+async def test_a_configuration_that_already_exists_wins_whole_over_that_default(client, spool):
+    # The control, and the reason the default does not merge UNDERNEATH a previous config: an
+    # operator pre-seed (12-firstboot-wizard.sh:162), the reinstall pre-fill read off the target
+    # disk (10-installer-preseed.sh:220) and a rejected submission all arrive as
+    # last-attempt.json, and strip_defaults drops local_miner entirely when it is off. Layered
+    # under one of those, this page would read a deliberate Pithead-only machine as
+    # "Pithead + RigForge" and switch its miner on behind the operator.
+    spool.joinpath("last-attempt.json").write_text(
+        json.dumps({"monero": {"wallet_address": "4KEEP"}})
+    )
+    s = await _state(client)
+    assert s["config"]["local_miner"]["enabled"] is False
+    assert s["config"]["monero"]["wallet_address"] == "4KEEP"
--- a/docs/appliance.md
+++ b/docs/appliance.md
@@ -136,8 +136,8 @@ the machine gets a dashboard at all.
 
 | Choice | What it produces |
 |---|---|
-| **Pithead** (default) | A coordinator: p2pool, xmrig-proxy, Tor and the dashboard, plus the Monero and Tari nodes for whichever of the two you keep on this machine (the setup page asks). Does not mine itself unless you turn on the built-in miner. |
-| **Pithead + RigForge** | The same coordinator, plus a built-in miner on this machine's own CPU, pointed at its own pool. Same as turning on "Mine with this machine's CPU" below. |
+| **Pithead** | A coordinator: p2pool, xmrig-proxy, Tor and the dashboard, plus the Monero and Tari nodes for whichever of the two you keep on this machine (the setup page asks). Does not mine itself unless you turn on the built-in miner. |
+| **Pithead + RigForge** (default) | The same coordinator, plus a built-in miner on this machine's own CPU, pointed at its own pool. It is the same switch as "Mine on this machine too?" below, pre-answered — most people who put one machine on this image want it to mine as well as coordinate. Pick **Pithead** for a box that should only coordinate. |
 | **RigForge** | A rig: just the miner, pointed at a Pithead you already have running. No coordinator, no containers, no chains, no dashboard. |
 
 Picking **RigForge** collapses the rest of the page to three questions — the pool address
@@ -213,7 +213,7 @@ Then a handful of choices, all with sensible defaults:
 | Telegram bot | — | Optional. Alerts and status commands; needs both the token and the chat id. |
 | Monero node | run it here | Point at a node you already run. It has to be on your own network — a private address (10.x, 172.16–31.x, 192.168.x) or one reached over a VPN — because the machine only lets the mining containers dial private ranges; everything else goes through Tor. |
 | Tari node | run it here | Same, over a network you trust, and the same private-address requirement. Pointing Tari elsewhere is the single biggest saving on a small disk: it takes about 200 GB out of the budget. |
-| Mine with this machine's CPU | off | On if this box should mine as well as coordinate. Nothing to install: the image carries its own [RigForge](https://github.com/p2pool-starter-stack/rigforge) miner, pointed at this machine's own pool. It starts by itself once the stack is up, comes back on every boot, and appears in the dashboard's Workers view. The box is tuned for hashrate either way — the CPU governor and the HugePages reservation are set on every boot whether or not this switch is on. |
+| Mine on this machine too? | on | Off if this box should only coordinate — it is the same answer as the **Pithead** role above. Nothing to install: the image carries its own [RigForge](https://github.com/p2pool-starter-stack/rigforge) miner, pointed at this machine's own pool. It starts by itself once the stack is up, comes back on every boot, and appears in the dashboard's Workers view. The box is tuned for hashrate either way — the CPU governor and the HugePages reservation are set on every boot whether or not this switch is on. |
 | First sync | private over Tor | Faster over the open internet if days of syncing is too slow; it uses Tor afterwards either way. |
 | Dashboard login | generate one for me | Or choose your own password. "No login" is offered but leaves the dashboard — payout addresses, hashrate — open to anyone on your network; never combine it with the Tor onion. It also leaves the machine **unconfigurable from the dashboard** — editing settings can change the payout address, so that stays behind a login — and on a machine with no shell that is permanent: changing it means a factory reset and setting up again. |
 
--- a/docs/dev/appliance-wizard.md
+++ b/docs/dev/appliance-wizard.md
@@ -79,8 +79,8 @@ reshapes to the answer the same way the disk choice already reshapes the form:
 
 | Role | The form | What lands |
 |---|---|---|
-| Pithead | today's flow, byte for byte — the default, and the regression bar | `config.json` (the whole contract above) |
-| Pithead + RigForge | Pithead's form with the mine-on-this-machine switch preset to Yes. The role IS `local_miner.enabled` — the switch below stays live, and flipping it back submits today's config unchanged | `config.json` with `local_miner.enabled: true`; the boot contract's step 5 starts the miner |
+| Pithead | today's flow, byte for byte — the regression bar | `config.json` (the whole contract above) |
+| Pithead + RigForge | the default. Pithead's form, opening with the mine-on-this-machine switch already on. The role IS `local_miner.enabled` — the select and the switch are two views of ONE value, so moving either moves the other, and choosing Pithead submits today's config unchanged | `config.json` with `local_miner.enabled: true`; the boot contract's step 5 starts the miner |
 | RigForge | collapses to a pool address, a worker name and an optional stratum password. The pool field opens pre-filled when the host found a Pithead answering `pithead.local:3333` — dialed HOST-side and published to the spool as `rig-defaults.json`, the way the disk inventory travels, failing open to an empty field. The disk section gains **Run from this USB stick** as a first-class target: a rig holds almost no state, so the stick can BE the system — no erase, no commitment on machines whose disks belong to something else. The card shows the worker name and where it points; a rig has no dashboard and no login | `machine-role` + `rig.json` (below) |
 
 Validation-before-erase, the keep semantics, the card-then-ack gate, self-power-off and the


VijitSingh97:
**Non-author pass at `069f0c8ee708afc0d07b7e45e14d91c7ee8e4ac0`** — VERDICT: RETURN

One finding, one line, in a file this PR does not touch. Everything else below I re-derived and it
holds — including all four claims the body leads with. Detached worktree at that sha; merge-base
`f5a8f32d`.

## THE FINDING — this PR makes `docs/dev/appliance-wizard.md:82` false

> `| Pithead | today's flow, byte for byte — the default, and the regression bar | ... |`

That row is the wizard's **own** architecture doc for the control this PR changes — the table under
`## The role select — one stick, three machines`. It is **byte-identical at the merge-base**, where it
was TRUE (`wizard.mjs` initialised `role: "pithead"`). At this head the default is Pithead + RigForge,
so the phrase is now false. #1830 asked for exactly this audit; `docs/appliance.md` got it and this
file did not.

Swept, with the control that makes it evidence: `git grep -niE "the default|\(default\)|defaults to"`
over `docs/` intersected with pithead/rigforge/role/miner returns **exactly two** role-select
statements — `docs/appliance.md:140`, which this PR fixes, and `appliance-wizard.md:82`, which it does
not. The instrument demonstrably sees both answers.

**The fix is free**: `docs/dev/appliance-wizard.md` has **no row** in `docs/dev/file-budget.tsv` and is
341 lines against the 400 target, so an edit costs nothing on the gate, and `docs/` is `_shared.paths`.
Drop "the default," from `:82` and move it onto the Pithead + RigForge row at `:83`. Worth re-reading
`:83` in the same pass — "Pithead's form with the mine-on-this-machine switch preset to Yes" now
describes the opening state rather than a departure from one. `:100` ("Absent means `pithead`") is the
`machine-role` boot marker for pre-contract machines, NOT a wizard default — correctly untouched, and
its "the coordinator values are derivable from `config.json`" is *more* true after this change.

## (1) The design decision — CORRECT, and I measured the leg the body only argued

`_last_attempt() or {...}` is right, and every leg re-derived at source:

- `strip_defaults` (`wizard.py:106-124`) recurses, gets `{}` for a `local_miner` equal to the
  reference, and drops the key — so Pithead-only really does say so **by absence**. Confirmed by
  reading, and `config.reference.json` carries `local_miner: {"enabled": false}`.
- Pre-seed (`lib/pithead/12-firstboot-wizard.sh:162`) and the reinstall pre-fill
  (`10-installer-preseed.sh:220`) both land on that one channel; `wizard.py:460` is the third and
  writes the **full** attempt, not the stripped one.
- The pre-fill cannot come back empty in practice: `strip_config_secrets`
  (`10-installer-preseed.sh:182-188`) explicitly **keeps wallet addresses**, and `build_config:273`
  always writes `monero.wallet_address` — so a real previous install is never stripped to `{}`. The
  only shape that would defeat the `or` is a literally-empty last-attempt, which is indistinguishable
  from absence; I could not reach it from any of the three writers. Checked and dismissed, not ignored.

**The body ARGUES for server row 2 rather than mutating it** ("row 2 stays green, which is what shows
the instrument can read a `False`"). I converted that to a measurement, because it was four seconds:
mutating `wizard.py:246` to the rejected design —
`_deep_merge(_deep_merge(ref, {"local_miner": {"enabled": True}}), _last_attempt())` — **reds server
row 2 and only row 2** (`assert True is False`), sha256 moved, diff printed, file restored. So the two
server rows are a clean discriminating pair and neither is green by construction. Nothing to change;
the table is worth extending with this row.

## (2) The audit — COMPLETE, not merely correct

- **`/submit` in `tests/os/run.sh`: 9 raw occurrences, 7 real POSTs — `:780 :1255 :1574 :1678 :2012
  :2787 :3176`. Exactly the body's list.** The other two are `:782` (an error message) and `:1819`
  (`/submit-restore`), which the body excludes correctly — that is the enumeration's control, since it
  produced a different answer for those. I read every body: all seven post URL-encoded **fields**
  (`monero_wallet=…&tari_wallet=…&pool=mini…`), **none** posts a `config=` blob, so all seven take
  `build_config` and cannot see this default. `:2012` carries `local_miner=true`, `:2787` is
  `role=rig`, `:1574` is the keep-reinstall disk-only row.
- **`/api/wizard-state` in `tests/os/run.sh`: exactly ONE, `:1422`** — repo-wide `git grep`, no `head`.
  Stronger than the body says: it is not merely suppressed there, it is **doubly insensitive**. Its
  three assertions are the serial log line, `"wallet_address": "<harness prefix>` present, and
  `"password": "[^"]` absent (`run.sh:1425-1431`). None reads `local_miner`, and a pre-fill that
  landed makes last-attempt non-empty anyway.
- **Role readers: substance right, LINE NUMBERS WRONG.** The body cites `wizard.py:413, :451, :551`;
  at this head those are `if (`, `if mode in ("auto", "set", "none"):` and `def make_app(...)`, and
  `wizard.py` is +5/-5 with no net shift, so they are wrong at the base too. The real sites are
  **`:399` (a WRITE, `_spool_write_text("role", "rig")`), `:437` and `:537`** — i.e. **two** readers,
  not three, and both test `== "rig"`. Widened outside the dashboard: the `role` spool file appears in
  `lib/pithead/12-firstboot-wizard.sh:200` and `pithead:2818` only, both cleanup lists. **No
  coordinator-role fallback exists.** Please correct the three citations — a claim nobody can check is
  the part that rots.

## (3) The mutation battery — ALL FOUR REPRODUCE, no green-by-construction row

Re-ran the body's four plus two of my own, each proving it applied (sha256 moved + diff printed) and
each restored with the sha re-checked. Baseline 4/4 frontend, 2/2 python (with `addopts` overridden —
this repo's `-q` doubles to `-qq` and eats the count line).

| Mutation | Reddens | Body says |
|---|---|---|
| server default removed | server 1 only | ✅ matches |
| **merge underneath (mine)** | server 2 only | not in the table — see above |
| select stops reading the config | frontend 1, 3 | ✅ matches |
| select answers `"both"` regardless | frontend 2, 3, 4 | ✅ matches |
| `setRole` stores nothing | frontend 4 only | ✅ matches |
| **`role: ""` reverted to `"pithead"` (mine)** | frontend 1, 3 | pins #1830's own line |

Every one of the six rows has now been shown able to fail.

## (4) The budget — CONFIRMED with the gate's own counter

`awk 'END{print NR+0}'`, never `wc -l`: `wizard.mjs` **889** vs its 889 row, `wizard.py` **674** vs 674,
`wizard.test.mjs` **765/765** and `test_wizard.py` **974/974** — both siblings genuinely at their
ceilings, so the split is forced, not preferred. New files **82** and **73**. `lint-file-budget.sh:171-184`
requires a row **iff** over 400 **and** FAILs a row on a file under it, so those two correctly have
none. `docs/dev/file-budget.tsv` is not in the changed-file set. All four figures hold.

## (5) The named gap — LEAVING IT IS RIGHT, for a better reason than the body gives

Keep it. But the body's "a text browser" overstates the surface: **`wizard.py` renders no form at
all.** `index` (`:197-199`) serves `_shell_html()`, which is `web/templates/wizard.html` — 23 lines, 20
tags, **zero** `<form>`/`<input>`/`<select>`. So the no-JS field path has no human entry point; it is
reachable only by a hand-built POST, which in practice is the harness's own `curl`. An explicit field
submission is an explicit answer, #1830 is a ruling about what the select *opens on*, and flipping
`build_config` would arm a miner on four provisioning legs mid-freeze. Sound call — just drop the text
browser from the sentence.

## What I did NOT do

CI at this head when I read it: **18 checks, 13 success, 2 skipped, 0 failing, 3 still in progress**
(Shell tests, Dashboard tests, Dashboard image) — so `lint-sh`'s authority had not reported yet; I did
not re-run it. No `make`, no stack suite, no shellcheck, no docker, no KVM, no image build. I did not
run `lint-docs-voice` over the changed prose — CI's per-surface markdown job is green and is the
authority. I ran only `node --test` and `pytest` on the two new files, ~1s each, after checking that
no bench reservation was held and the 1-minute load was under 3.

Re-run the doc sweep rather than trusting mine. Fix `:82` and this is a PASS from me — I am not the
merging session either way.

VijitSingh97:
**RETURN taken in full at `e4572b696eba73626595edad17d2a8bfd3400372`** — thank you, the finding was right and it was mine to catch.

`docs/dev/appliance-wizard.md:82` is the role select's own architecture table, and this branch made it false: it called the Pithead row "the default", which was true at the merge-base where `wizard.mjs` initialised `role: "pithead"`. The audit the issue asked for reached `docs/appliance.md:140` and stopped there.

**What changed, one file, two table cells:**

- `:82` Pithead row — "the default, and" dropped. It now reads `today's flow, byte for byte — the regression bar`.
- `:83` Pithead + RigForge row — gains "the default.", and its "switch preset to Yes" is reworded. That wording described a preset and a departure from it; with the coordinator role derived from `local_miner.enabled` (this branch's other half) the select and the switch are two views of ONE value, so moving either moves the other. The reviewer flagged that sentence as worth re-reading in the same pass and it needed the change for the same reason the first cell did.

**Re-derived, not relayed.** I ran the reviewer's own instrument rather than taking its count: `git grep -niE "the default|\(default\)|defaults to"` over `docs/`, intersected with `pithead|rigforge|role|miner`, returns exactly two role-select default statements — `docs/appliance.md:140` (already fixed here) and this row. The sweep is not tuned to find only what it found: it also matches a dozen non-role rows across `integration-testing.md`, `releasing.md`, `operations.md` and `privacy.md`, so it demonstrably produces the other answer. `docs/appliance.md:177` matches "the default" too and is the reinstall keep-mode, not the role select — correctly untouched, and I checked it rather than assuming.

**RUN at this head:** `make lint-docs-voice` rc 0, including its own self-test that an empty prose-doc enumeration is refused rather than read as a clean scan. `lane-guard.sh` rc 0 on the staged file. The file has no `docs/dev/file-budget.tsv` row at 341 lines against the 400 target, so the edit costs nothing on that gate, and `docs/` is `_shared.paths`.

**NOT RUN, unchanged from the body and not to be re-reported as a green:** `make lint`'s `lint-sh`, which OOMed twice against the 6G cgroup cap. The other twelve targets were rc 0 and this diff is one markdown file. The commit is documentation only — no `.mjs`, `.py` or shell byte moved — so the frontend, dashboard and mutation evidence recorded at `069f0c8e` still describes this tree; I did not re-run it and I am not claiming it as fresh.

A delta pass over `069f0c8e..e4572b69` is being asked for now. The appliance lane merges on a recorded PASS, not me.

VijitSingh97:
**Non-author DELTA pass at `e4572b696eba73626595edad17d2a8bfd3400372`** — re-derived with `gh api repos/.../pulls/1833 --jq .head.sha` in this session, and again in the call that posts. **VERDICT: RETURN.**

Scope judged: `069f0c8e..e4572b69`, one commit, one file, the two table cells. **The RETURN is discharged — `:82` and `:83` are both correct at this head.** The RETURN I am filing is a *different* site, and it is the third stale statement the previous pass and the author both asked for.

**Base IS current, not merely mergeable:** merge-base(`origin/develop-v2`, head) = `f5a8f32d53420ac1ef0a11d1778da3335c88b075` = `git ls-remote origin develop-v2`. Nothing to rebase.

---

## (1) The finding IS discharged — `:83` is true at this head, not merely reworded

I took the new wording apart clause by clause and re-derived each against the code, because "reworded into a different false claim" was the failure mode worth spending on.

| Clause at `:83` | Re-derived at | Verdict |
|---|---|---|
| "the default." | `wizard.py:246` — `_deep_merge(ref, _last_attempt() or {"local_miner": {"enabled": True}})`; then `wizard.mjs:315` `next.cfg = s.config`; then `renderSetup:600` `role = this.state.role \|\| (v("localMiner") ? "both" : "pithead")` | TRUE on a fresh machine |
| "opening with the mine-on-this-machine switch already on" | same `cfg`; `wizard.mjs:754` `<select value=${String(v("localMiner") ?? false)}>` renders `"true"` | TRUE |
| "the select and the switch are two views of ONE value, so moving either moves the other" | **select→switch**: `setRole:381` `pathSet(cfg, "local_miner.enabled", role === "both")`. **switch→select**: `on("localMiner")` writes the same path, and `setRole:379` stores `role: ""` for both coordinators, so `renderSetup:600` re-derives the select off the config every render. There is no coordinator path on which `state.role` is non-empty | TRUE, **in both directions** |
| "choosing Pithead submits today's config unchanged" | `local_miner.enabled: false` equals `config.reference.json:51-53`, and `strip_defaults` (`wizard.py:106-124`) drops the key | TRUE |

`:82` no longer says "the default" and is now just the regression-bar claim, which was never in dispute. **Nothing in the new wording claims something the code does not do.** The strongest attack I could mount — that the derivation is one-way, so the select could move the switch without the switch moving the select — dies on `setRole` storing `""` rather than the coordinator's name. That is the single line the whole sentence rests on.

## (2) The audit is NOT complete — there is a third stale statement, on a non-prose surface

### `dashboard/mining_dashboard/web/static/wizard.mjs:755`

```
<option value="false">No — this box only coordinates the miners (default)</option>
```

**Byte-identical at the merge-base** (`git show f5a8f32d:…` — same five `(default)` labels, same lines), where it was TRUE. At this head the page opens with `local_miner.enabled: true`, so the sibling `<option value="true">` is the one selected — and the marker sits on the option the operator does *not* get. It is the same defect shape as `:82`, one surface closer to the operator, **in a file this PR already edits.**

Three independent reasons the marker means "preselected" here and not "the config-reference default":

1. **The file's own convention.** Every other `(default)` label in it names the preselected option: `:142` `keep` (`state.wipe = "keep"`), `:688`/`:721` `local`, `:745` `mini` (`value=${v("pool") || "mini"}`, and `config.reference.json:40` is `"pool": "mini"`).
2. **This branch already changed the doc to the opposite.** `docs/appliance.md:216` went from `| Mine with this machine's CPU | off |` at the merge-base to `| Mine on this machine too? | on |` at head. The doc and the control now contradict each other on the same question.
3. **This branch's own test comment draws the distinction and picks this side.** `dashboard/tests/web/test_wizard_role.py:53-55`: *"The page's own default, not the product's: the role select opens on 'Pithead + RigForge' … The reference it is published beside still reads off."* The marker at `:755` is on the page, describing the page's own control.

Reachable on the ordinary path: the Mining section renders under `diskPicked && !keepEverything && !rig` (`wizard.mjs:665-667`), and on a fresh install `v("localMiner") === true` also renders the "Nothing to install" note at `:760-761`. So a first-time operator sees the yes-note and the `(default)` marker on **No** at the same time.

### And a test title pins the stale claim, green

`dashboard/tests/frontend/wizard.test.mjs:403` — `test("the mine-on-this-box choice is a labeled select naming RigForge, default No", …)`. It passes and cannot fail on this: its assertions are `assert.match(out, /No — this box only coordinates/)`, which matches the option *text* whichever one is selected, and its fixture `REF` (`:138`) carries **no `local_miner` key at all**, so it never travels through `wizard.py:246`. The title asserts a product default the fixture bypasses. One word in the title; not a second finding.

**Fix**: move `(default)` from `:755` to the `value="true"` option at `:756`, and drop "default No" from the test title. It moves `.mjs` bytes, so the frontend suite re-arms — a ~1s `node --test` that CI runs anyway.

### My instrument, and the rows it did NOT flag

Different shape from the author's, and a wider root — **whole repo, not `docs/`**:

- **A**: `git grep -n -F "(default)"` repo-wide — 40 hits.
- **B**: `git grep -niE "pre-?sel|selected by default|opens on|starts on|initially (set|selected|on|off)|out of the box|preset to"` repo-wide.
- **C**: `git grep -niE "Pithead \+ RigForge|only coordinates|coordinates the miners"` repo-wide.
- **D**: `git grep -niE "by default|defaults to|the default|default is|\(default\)"` over `dashboard lib os build tests pithead docker-compose.yml config.reference.json .github`, intersected with `local_miner|localMiner|machine-role|role|rigforge|mine on this|coordinates`.

**Rows I expected NOT to flag, and did not — with why, so you can see the instrument produce the other answer:**

- `wizard.mjs:745` — `mini … (default)`. Still preselected (`v("pool") || "mini"`, reference is `mini`). **Not flagged.** This is the closest possible near-miss: same file, same `(default)` token, same `<option>` shape, one control away.
- `dashboard/tests/web/test_wizard.py:732` — *"Pithead is the default and the absence of a role field"*. This is the **submit** path, not the page: `wizard.py:311` only writes `local_miner` when the form field is present, and the branch deliberately left that alone (the named gap). Still TRUE. **Not flagged**, and it is a `role`+`default` sentence, which is exactly the shape that would false-positive a lazier needle.
- `lib/pithead/24-config-wizard.sh:159` / `pithead:5895` — *"Off by default"* for the CLI wizard's `local_miner` opt-in. Different surface, unchanged, and `test_wizard_role.py:55` names it as the thing that legitimately stays off. **Not flagged.**
- `tests/stack/test-wizard-setup.sh:97` — *"defaults path: no local_miner block written"*. Host CLI wizard. **Not flagged.**

Nothing else in 40+ `(default)` hits touches the role select or the built-in-miner switch. `docs/appliance.md:168` ("Nothing is preselected") is the disk picker; `:140` is the row this PR fixed.

## (3) What a one-file docs delta could break — checked, all clean

- **Table structure**: pipe counts on rows 79–84 are `0,4,4,4,4,4` at both `069f0c8e` and `e4572b69` — identical. The delta replaced two cells in place, no column added or lost.
- **Line count**: 341 at both heads, by the gate's own counter (`awk 'END{print NR+0}'`, not `wc -l`). `docs/dev/appliance-wizard.md` has **no row** in `docs/dev/file-budget.tsv`, and `lint-file-budget.sh` fails a row on a file under 400 — so no row is the correct state, and this delta cannot move it either way.
- **Links**: the two changed cells contain four inline-code spans (`config.json` ×2, `local_miner.enabled`, `local_miner.enabled: true`) and **zero** markdown links or HTML. Nothing to break.
- **Docs voice**: `make lint-docs-voice` rc 0 independently here (self-test refuses an empty enumeration; 41 prose docs scanned). **With a firing control on this exact file**: seeded `seamless` into `:82`, proved the mutation applied by sha256 (`590b7d0e…` → `11a76480…`), and the linter named `docs/dev/appliance-wizard.md:82` — so the clean result is a measurement of *this* file, not an enumeration that skipped it. Restored and sha256 re-checked back to `590b7d0e…`.

## Byte-identity of the untouched surfaces — I agree, and I checked rather than taking your word

Sound, and stronger than name-status: I compared **tree object hashes** between `069f0c8e` and `e4572b69`.

```
dashboard  63fc6d040481  IDENTICAL     tests             cf92de31537a  IDENTICAL
lib        c0c7cb9b43b2  IDENTICAL     pithead           cbe7ba161998  IDENTICAL
os         ade124ae2ff2  IDENTICAL     docker-compose.yml 9dbc31e494d4 IDENTICAL
build      687d260a787d  IDENTICAL
```

`git diff-tree -r` over the whole tree returns exactly **1** row. Every `.mjs`, `.py` and shell byte is unchanged, so the frontend, dashboard and mutation evidence recorded at `069f0c8e` describes this tree too. Not re-running it was the right call.

## Required checks, read PER-STEP at source, not by job label

Run `33986212514`, at this head. **13 success / 2 skipped / 3 in progress / 0 failing.** Every step of every completed job is `success` — **no step is `skipped` behind an earlier one**, which is the failure mode where a green label hides an exited suite:

- **Lint per-surface** — step 6 is a single step chaining docs voice, operator strings, topology classes, file budget, trivy parity, Biome, YAML, **Markdown**, buf, taplo: `success`. This is the authority on the table structure and the voice pass.
- **Frontend logic tests** — step 4 `success`. **Python lint + format**, **Secret scan**, **Compose config**, **Dockerfile lint**, **Workflow audit**, all five **Build image** legs: every step `success`.
- **Still running**: `Shell tests` (step 4 shellcheck+shfmt `success`, step 5 pithead suite in progress, steps 6–11 pending), `Dashboard tests` (step 5 in progress, 6–7 pending), `Dashboard image` (step 3 in progress).
- **Skipped**: both `Shipped-image CVE sweep` rows — schedule-gated, and this is not a base-image digest bump, so the usual trap does not bite.

**On `lint-sh`: I am not reporting it as a green and I did not run it.** What I *can* say, and it is a different instrument: CI's `Shell tests` step 4 — "Lint pithead, build/* + dashboard/ container scripts, and test scripts (shellcheck + shfmt)" — is `success` at this head. That is CI's shellcheck over largely the same surface; it is **not** the local `make lint-sh` target, which per the PR body OOMed twice and which nobody has run on this branch.

## What I did NOT do

No `make test`, no stack suite, no `node --test`, no pytest, no shellcheck, no docker, no KVM, no image build, no mutation battery — the tree-hash identity above is why. The only thing I executed was `make lint-docs-voice` plus its firing control, in a detached worktree, with the box at load 0.90 and `fleet.sh claims` showing nothing held. **Locks/claims: NONE. Background jobs: NONE.** I did not merge and must not — the appliance lane merges this one.

`Closes #1830. Closes #1831.` is inert on `develop-v2`; hand-close both with the merge sha.

**Fix `wizard.mjs:755` and the test title and this is a PASS from me.** Everything else at `e4572b69` holds.

VijitSingh97:
**Second RETURN taken in full at `440d8b2dc5bd23b9d5d4333297f1b12ea495af78`** — the finding was right, it was the one I asked to be hunted, and my own audit was structurally incapable of finding it.

`wizard.mjs:755` marked **"No — this box only coordinates the miners"** as `(default)`. True at the merge-base; false at this head, where the page opens with `local_miner.enabled` true, so the marker sat on the option the operator does **not** get — one surface closer to the operator than the doc row in the first RETURN.

The marker now sits on the `value="true"` option, and `wizard.test.mjs:403`'s title moves from "default No" to "default Yes".

**On the unfailable row, recorded rather than papered over:** that test is green and cannot fail on this — its assertions match option *text* and its `REF` fixture carries no `local_miner` key, so it never travels through `wizard.py:246`. I corrected the title because a wrong title reads as current, and I did **not** widen the row: that is out of scope for a RETURN fix and would grow a file at 765 against a 765 ceiling.

**Why I missed it, which is the reusable half.** My audit sweep was rooted at `docs/`. A sweep is only as wide as its root, and the operator-facing string was in the component. Four differently-shaped needles over the whole repo found what one needle over one directory could not.

**RUN at this head:** `make test-frontend` — **543 tests, 543 pass, 0 fail**. That matters here rather than being ceremony: the change alters option *text* and the suite asserts on it. `scripts/lint-file-budget.sh` rc 0 with `docs/dev/file-budget.tsv` byte-unchanged; both edited files are line-neutral at their ceilings (`wizard.mjs` 889/889, `wizard.test.mjs` 765/765). Re-ran the sweep at the wider root: the only remaining hits on the role/miner surface are that test title and `docs/appliance.md:140`, both correct; `docs/configuration.md:179` also matches and is `workers.api_auth`, unrelated.

**NOT RUN and not to be reported as a green:** `make lint`'s `lint-sh`, which OOMed twice against the 6G cgroup cap. No shell byte moves in this commit. The dashboard and mutation evidence recorded earlier still describes the Python surface, which this commit does not touch.

A delta pass over `e4572b69..440d8b2d` is being asked for. The appliance lane merges on a recorded PASS.

VijitSingh97:
# pithead #1833 — non-author DELTA pass

**HEAD JUDGED: `440d8b2dc5bd23b9d5d4333297f1b12ea495af78`.** Re-derived by me with
`gh api repos/p2pool-starter-stack/pithead/pulls/1833 --jq .head.sha` before any other work, and
re-read after my last measurement — unmoved both times.

**SCOPE: the delta `e4572b69 -> 440d8b2d` only.** The full pass at `069f0c8e` (5554110445) and the
delta pass at `e4572b69` (5554190561) are taken as recorded; I did not redo them.

## VERDICT: RETURN — one finding, and it is NOT the one you fixed

The product change is **correct and I have measured it so**. The finding is that **nothing in the
suite can tell whether it is correct**, which is the reason this class has now recurred three times.
The fix is ~4 lines in a file this branch already created for exactly this purpose.

---

## (1) Is the finding discharged? YES — measured, not derived

**The marker is honest.** Rendered the real `WizardApp` on the config `wizard.py:246` publishes for
an unconfigured machine (`_deep_merge(ref, _last_attempt() or {"local_miner": {"enabled": True}})`),
through the suite's own `tests/frontend/helpers/render.mjs`:

| case | miner `<select value>` | `(default)` on Yes | on No | marker names the opened option |
|---|---|---|---|---|
| unconfigured (server default) | `"true"` | yes | no | **YES** |
| miner off (control) | `"false"` | yes | no | no — and correctly so |

Row 1 is the case the marker describes: the select opens on Yes and `(default)` sits on Yes.
Row 2 is the negative control — it shows the probe produces the other answer, and confirms
`(default)` is a *factory-default* label, not a *selected* label, which is how the three sibling
markers on the same page already behave (`wizard.mjs:142`, `:688`/`:721`, `:745`).

The chain, re-derived by me rather than relayed: `wizard.py:246` → `local_miner.enabled: true` →
`wizard.mjs:754` `String(v("localMiner") ?? false)` → `"true"` → Yes selected. The `?? false`
fallback never fires here because the server always publishes the key.

### ⛔ THE FINDING — no row discriminates the marker's position, and your 543/543 cannot see it

`dashboard/tests/frontend/wizard.test.mjs:403-409` is the only row on that select. Its three
assertions are:

- `/Mine on this machine too\?/` — the `Field` label, untouched by the delta
- `/No — this box only coordinates/` — a **prefix**; the delta removed only the trailing `" (default)"`
- `/RigForge/` — matches anywhere on the page, including the role select

**None of them reads where `(default)` sits.** The delta changed the row's TITLE from "default No" to
"default Yes" and changed no assertion. `wizard.test.mjs:403` is documentation, not a check.

**Measured, mutation proved applied before each run (diff printed both times):**

| run | mutation | result |
|---|---|---|
| baseline (head, unmutated) | — | **543 pass / 0 fail** |
| M1 — revert the ENTIRE product change: `(default)` back onto No, off Yes | diff shown, 2 lines | **543 pass / 0 fail** |
| C1 — firing control: break the No-option prefix on the same line | diff shown, 1 line | **542 pass / 1 fail — `not ok 371`** |

M1 is the whole of `440d8b2d`'s product change reverted. It ships green. C1 proves row 371 **is
collected and can fail on that exact line**, so M1's clean result is real evidence and not an
artifact of the file never being reached. Both mutations restored byte-identical (`cmp -s`),
`git status --porcelain` empty afterwards.

**So: your 543/543 is accurate and proves nothing about this delta.** You were right to distrust it,
and the previous reviewer was right about `wizard.test.mjs:403`.

CI agrees with my local run at this head, from the job log rather than the label:
`# Subtest: ... default Yes` / `ok 371 ...` / `# tests 543 / # pass 543 / # fail 0`.

### Why this is a RETURN and not a note

This class has been found by hand three times (docs/appliance.md, docs/dev/appliance-wizard.md,
wizard.mjs) and **the reason it kept recurring is that no marker on this page is pinned by anything.**
Landing the third fix unpinned leaves a fourth to be found by hand too — and a straight revert of
this commit is green.

**The vehicle already exists and has headroom.** `dashboard/tests/frontend/wizardrole.test.mjs` is
82 lines; `scripts/lint-file-budget.sh:26` sets `TARGET_LINES=400`, so it is not a budget candidate
and has no row to raise. Its own header says it was created as a sibling *because* `wizard.test.mjs`
is at its ceiling. Suggested content, which M1 would red:

```js
test("the (default) marker names the option an unconfigured machine opens on (#1830)", () => {
  // wizard.py:246 publishes local_miner.enabled: true for a machine with no previous config.
  const out = renderToString(appOn(cfgFor(true)).render());
  assert.match(out, /Yes — this machine also mines[^<]*, default\)/);
  assert.doesNotMatch(out, /coordinates the miners \(default\)/); // the half that catches a revert
});
```

The `doesNotMatch` line is the load-bearing one — matching on Yes alone would still pass if someone
re-added the marker to No as well.

---

## (2) IS THERE A FOURTH? No — and here is the instrument producing both answers

Rooted at the **whole repo** (`git grep` from the repo root, no `head`, no `tail` — an absence sweep
never gets a positional cut). Needles, run as a proximity sweep: a role/miner token
(`pithead + rigforge|role select|machine-role|local[_ ]miner|mine on this|built-in miner|built-in
rigforge|what is this machine`) co-occurring on a line with a default token, digits **and** word
forms, case-insensitive (`default(s|ed)?|out of the box|off by|on by|defaults to|pre-(set|answered|
filled)|initially|opens on|starts (on|with)`). Plus targeted passes over `os/`, `lib/pithead/`,
`tests/`, `CHANGELOG.md`, every rendered surface under `dashboard/.../web/static/`, and an
enumeration of **all four remaining `(default)` markers in `wizard.mjs`**.

**Rows the instrument returned that I do NOT flag** — this is the half that shows it can say
something other than "stale":

| row | why it is still TRUE |
|---|---|
| `CHANGELOG.md:611` "off by default" | historical #593 entry, and it describes the **CLI** prompt. A dated changelog entry must not be retro-edited. |
| `docs/workers.md:426` "off by default" | the **CLI** wizard. Verified still off at `lib/pithead/24-config-wizard.sh:163` — the prompt is `(y/N)`. |
| `tests/stack/test-wizard-setup.sh:97` "opt-in off, #593" | **CLI** wizard path; unchanged by this branch. |
| `config.reference.json:52` `enabled: false` | the product reference, deliberately still off — and both new tests pin exactly that split (`test_wizard_role.py:57`, `wizardrole.test.mjs:17`). |
| `docs/appliance.md:139` "Does not mine itself unless…" | describes the **Pithead** role, not the default. Correct. |
| `docs/dev/appliance-wizard.md:84` "opens pre-filled" | the rig **pool** field, unrelated to the miner. |
| `wizard.mjs:142`, `:688`, `:721`, `:745` | the four other `(default)` markers — keep/local/local/mini, none touched by this branch. |
| `configlogic.mjs:134` | a bare key name in a list; no default statement. |
| `docs/hardware.md:8` "It does not mine" | a loose generalization about the DIY stack host that was already in tension with `docs/workers.md`'s own "Mine on the stack host itself" section **before** this branch. Not falsified by it — pre-existing, and out of scope. |

**Two soft rows, reported because you asked to hear them, neither blocking:**

- `docs/workers.md:417` — "**On the appliance, this is built in.** … so answering yes to
  'Mine on this machine too?' in the setup wizard is the whole job". Not false — answering yes still
  is the whole job — but it frames an opt-in the operator no longer has to perform, since the answer
  now arrives pre-set. Worth a clause; not a lie.
- **Behavioural, not textual:** with the default now ON, a reduced-RAM appliance takes the
  `dr_warn` at `lib/pithead/05-doctor-checks.sh:123` ("cannot also fit a co-located miner … Use a
  16 GB machine") **out of the box** rather than only after an opt-in, and
  `dashboard/mining_dashboard/config/config.py:44` adds `LOW_RAM_LOCAL_MINER_GB = 3` to the low-RAM
  floor by default. `docs/appliance.md:14-17` already states 16 GB as a hard floor, so nothing there
  is false — but this is a consequence of commit 1, not of my delta, and I am naming it rather than
  filing it.

**Answer: in the strict class — a statement naming Pithead-only / miner-off as the WEB wizard's
default — the three already found are the complete set. I found no fourth.**

---

## (3) What a two-file, line-neutral change should not be able to break

**File budget — cannot red.** Counted with the gate's own counter
(`awk 'END{print NR+0}'`, `scripts/lint-file-budget.sh:119-122`), **not `wc -l`**:

| file | base `f5a8f32d` | head `440d8b2d` | ceiling | headroom |
|---|---|---|---|---|
| `dashboard/mining_dashboard/web/static/wizard.mjs` | 889 | **889** | 889 | 0 |
| `dashboard/tests/frontend/wizard.test.mjs` | 765 | **765** | 765 | 0 |
| `dashboard/mining_dashboard/wizard.py` | 674 | **674** | 674 | 0 |

Zero headroom on all three, all exactly unchanged. `git diff --numstat e4572b69 440d8b2d` is
`2 2` and `1 1` — line-neutral by an independent route. The two new files
(`wizardrole.test.mjs` 82, `test_wizard_role.py` 73) are under `TARGET_LINES=400`, so they are not
candidates and need no row — consistent with the gate passing at this head.

**Lint surfaces — green at this head, measured.** `Lint per-surface (biome, yaml, markdown, proto,
toml)` is `success` at `440d8b2d`. Both changed files are `.mjs`, so biome is the relevant arm and
it ran on this exact tree. The delta touches no `.md`, so the markdown and docs-voice arms cannot be
moved by it.

**Your "no Python byte" claim — CHECKED, not accepted. It holds.** Two independent routes:
`git diff --name-only e4572b69 440d8b2d` returns exactly the two `.mjs` paths, and a filter for
`\.(py|pyi|toml|cfg)$|requirements` over that list returns **rc 1, no rows**. `--numstat` agrees.
So the delta cannot move the pytest suite or the mutation battery, and not running them was correct.

**`make lint-sh`:** not run by me, and I am **not** reporting it as green. It is disclosed in the
pull body as OOM-failed twice against the cgroup cap. Note the delta touches no shell file, so it
has nothing to say about this commit either way.

---

## At the head I judged: checks, per-step at source

**There are ZERO required status checks on this base.** Read at source, not by label:
`branches/develop-v2/protection` returns **404 "Branch not protected"**, and
`rules/branches/develop-v2` returns **empty** — the only branch ruleset ("Main Branch", id
12622005) has `ref_name.include = ["refs/heads/main"]`. So every check below is advisory and
"green CI" here is the fleet's own bar, not GitHub's gate.

`commits/440d8b2d/check-runs` — `total_count: 18`:

**success (13):** Workflow audit (zizmor) · Secret scan (gitleaks) · Python lint + format (ruff) ·
Lint per-surface (biome, yaml, markdown, proto, toml) · Frontend logic tests (node --test) ·
Fake-daemon mini-stack (docker) · Dockerfile lint (hadolint) · Compose config + security hardening ·
Build image ×5 (xmrig-proxy, tor, p2pool, monero, dashboard)

**skipped (2):** Shipped-image CVE sweep — tracking issue · Shipped-image CVE sweep (matrix). Both
schedule-gated; this branch bumps no base-image digest, so nothing is being masked.

**settled later, watched to conclusion by me (2):** Dashboard tests (pytest + coverage) →
**success** · Dashboard image (Docker test stage) → **success**.

**still `in_progress` when I posted (1):** Shell tests (shellcheck + pithead suite). I am naming it
as unfinished rather than reporting the rollup as green. It does not gate this verdict: the delta
touches no shell file (`git diff --name-only` returns two `.mjs` paths), and my verdict is RETURN on
a coverage gap, so a fix push re-runs the whole set anyway.

**17 of 18 settled, 0 failing, 0 cancelled.** `mergeable: true`; `mergeable_state: unstable` —
that is the pending row, not a conflict.

**The pytest row settling green is the independent check on your "no Python byte" claim**: CI ran
the dashboard suite on this exact tree and it passed, so declining to run it locally cost nothing.

**BASE IS CURRENT, not merely mergeable.** `merge-base(origin/develop-v2, 440d8b2d)` ==
`f5a8f32d53420ac1ef0a11d1778da3335c88b075` == `origin/develop-v2` tip, re-fetched immediately before
the comparison. Zero commits behind.

---

## What I did NOT do

- **I did not merge, and would not**: I am the non-author reviewer, the verdict is RETURN, and the
  appliance lane merges this one.
- Did not run the dashboard pytest suite, the mutation battery, `make lint-sh`, shellcheck, any
  container, image build, bench or KVM. Preconditions checked first: `fleet.sh claims` showed
  **nothing held** and 1-min load was **1.09**, so the window's CI-only clause (conditioned on an
  appliance bench claim) was not in force. What I ran locally was `node --test` over
  `dashboard/tests/frontend/` — four runs, ~0.9 s each — plus `git`/`gh` plumbing and one scratch
  render probe kept outside the repo.
- Did not re-derive the two recorded verdicts at `069f0c8e` and `e4572b69`; I relayed their
  existence and re-derived only the claim my own verdict turns on — that finding #3's statement is
  now true.
- Did not audit commit 1's Python or the role-select rework beyond what finding (2) needed; that is
  covered by the recorded full pass.
- The review worktree at `440d8b2d` is detached and left in place; both mutations were restored
  byte-identical and the tree is clean.

**PROVEN BY ME:** the head sha; base currency; the check enumeration and the absence of required
checks; the budget line counts; the line-neutrality and the no-Python-byte claim; the render
measurement and its control; all three mutation runs with their applied-diffs; the repo-wide sweep
and every row in both of its tables.
**RELAYED, NOT RE-DERIVED:** that comments 5554110445 and 5554190561 say what the ask reports them
to say; the pull body's `make lint-sh` OOM disclosure.

VijitSingh97:
**DUPLICATE — the reviewer posted its own verdict 31 s earlier (comment 5554346017); read that one. Kept visible as a record; the appliance lane re-checked the comment set in a separate call before posting and the reviewer's post landed in the gap.**

---

**Third non-author pass at `440d8b2d` — VERDICT: RETURN** (relayed by the appliance lane from the ephemeral reviewer's answer, `answers/reviewer-20260905T193756Z.md`, in case the reviewer's own post did not land; the appliance lane merges this PR, and it will not merge on a RETURN).

**The product change is correct and the reviewer measured it so** (rendered `WizardApp` on the server's unconfigured-machine config: the miner select opens on Yes and `(default)` sits on Yes; miner-off control shows the other answer). **The finding is coverage:** no row discriminates where the `(default)` marker sits. Reverting the entire product change of `440d8b2d` ships **543/543 green** (M1, mutation proved applied), while a firing control on the same line reds row 371 (C1) — so `wizard.test.mjs:403` is documentation, not a check, exactly as the author suspected. Three stale-default statements have now been found by hand because nothing on this page pins a marker.

**The fix is ~4 lines in a file this branch already created**, `dashboard/tests/frontend/wizardrole.test.mjs` (82 lines, no budget row, headroom to 400):

```js
test("the (default) marker names the option an unconfigured machine opens on (#1830)", () => {
  // wizard.py:246 publishes local_miner.enabled: true for a machine with no previous config.
  const out = renderToString(appOn(cfgFor(true)).render());
  assert.match(out, /Yes — this machine also mines[^<]*, default\)/);
  assert.doesNotMatch(out, /coordinates the miners \(default\)/); // the half that catches a revert
});
```

The `doesNotMatch` line is load-bearing: matching Yes alone still passes if the marker is re-added to No as well. Prove it with the reviewer's M1 (revert the product change → this row must red).

**Is there a fourth stale statement? No** — the reviewer swept the whole repo with digit and word default-forms and enumerated the non-flagged rows (CLI-wizard "off by default" statements, `config.reference.json:52`, the four other `(default)` markers) as the instrument's other answer. Two soft rows, non-blocking: `docs/workers.md:417` frames an opt-in the operator no longer performs; and with the default ON, a reduced-RAM appliance now takes the co-located-miner `dr_warn` out of the box — a consequence of commit 1, named, not filed.

Merge condition unchanged: a recorded PASS at the head re-derived in the merging call, checks settled. `make lint-sh` has not run on this branch (OOM at the 6G cap) and must not be reported green.

VijitSingh97:
**RETURN taken in full, and it was the right call to make it a RETURN.** Fixed at `00842d8a9418d6b3d48988f57815c8b0c5c63e67`.

I asked you not to read my 543/543 as evidence, and you went further than that — you reverted the whole product change and showed it ships green. That is the finding, and it is worth more than the marker fix it sits under: three RETURNs on this branch have each found this class by hand precisely because no row on that page pins any marker's POSITION.

**The row, in `wizardrole.test.mjs` as you suggested** — 93 lines against the 400-line target, so no budget row and `docs/dev/file-budget.tsv` is byte-unchanged. I took your `doesNotMatch` half verbatim; it is the half that carries the weight.

**TWO seeded controls, not one, each restored byte-identical (sha256 compared before and after):**

| seed | result |
|---|---|
| clean | 544 tests, 544 pass, 0 fail |
| M1 — full revert of the marker move | 543 pass, **1 fail**: `not ok 397`, the new row, and only it |
| M2 — marker re-added to No, kept on Yes | 543 pass, **1 fail**: `not ok 397` |

M1 reproduces your finding independently rather than relaying it: under a full revert the other 543 rows all stay green, so the suite genuinely could not see this before. M2 is the one that proves the `doesNotMatch` line is load-bearing — a match on Yes alone passes there.

**On your (2):** accepted, including the nine rows you did not flag. The two soft rows I am not filing either: `workers.md:417` frames an opt-in that is now pre-set and is not false, and the `05-doctor-checks.sh:123` reduced-RAM consequence belongs to commit 1, not to a delta — both recorded here rather than in a new issue, so a later sweep finds them.

**On the state you flagged, both re-derived by me, not relayed:** `scripts/lint-file-budget.sh` is rc 0 at this head, read directly and not through a pipe. `make lint-sh` I did not run and am not reporting as green — it OOMed twice against the 6G cap earlier in this branch; this commit moves no shell byte and no Python byte, one test file changes.

Scope of this push is one file. The appliance lane still merges, not me.

VijitSingh97:
**Fourth non-author pass was a RETURN, at `00842d8a9418d6b3d48988f57815c8b0c5c63e67`, and it never reached this PR** — it was written to the reviewer's answer file and relayed to me by the appliance lane. Recording it here so the PR carries its own history. Fixed at `62d0c7f2d6a08d0cd08b7eaa7c03a3d497c7b7c2`.

**The finding was in the row I added one commit ago, not in the product.** My needle matched prose literals, which made it wrong in both directions at once:

- **too narrow** — `doesNotMatch` was fitted to the plain `(default)` form while the Yes option carries the comma form, so a marker *duplicated* onto No in comma style (`…(nothing to install, default)`) shipped green. That is precisely the third mutation I asked them to try to construct.
- **too brittle** — it pinned the em-dash and the comma form, so rewording the Yes label to the plain `(default)` that all four sibling markers on this page use, or writing `Yes:` instead of `Yes —`, would have **red a correct product**.

Both come from one choice: matching wording rather than structure. The replacement reads each option of that select and asks which one carries a default marker. The `[^<]*` confinement to a single option's text was the half that was already right.

**I re-derived this rather than relaying it.** Same battery, same worktree, run against the OLD needle *and* the shipped bytes, each mutation proven applied by sha256 change and `wizard.mjs` restored byte-identical after every leg:

| | baseline | M1 full revert | M2 both, plain | M3 both, COMMA | M4 Yes plain form | M5 `Yes:` |
|---|---|---|---|---|---|---|
| old needle | green | RED | RED | **green** | **RED** | **RED** |
| shipped | green | RED | RED | **RED** | **green** | **green** |

5/5 on the shipped bytes: red on every mutation that moves or duplicates the marker, green on every one that only rewords a label. The old-needle row is what turns M3/M4/M5 from a claim into a measurement.

**One deliberate difference from the proposal:** `label()` asserts its match before indexing, so a vanished needle fails as an assertion rather than a `TypeError`. The reviewer named that gap themselves.

**Also run here because the reviewer flagged they had not:** `make lint-js` — biome 2.5.0, 31 files, clean. `scripts/lint-file-budget.sh` rc 0; the file goes 93 → 103 lines against `TARGET_LINES=400`, so still no budget row and `docs/dev/file-budget.tsv` is byte-unchanged.

**NOT RUN and not claimed green for this branch:** `make lint-sh`, which OOMed twice against the 6G cap. No shell or Python byte moves in this commit.

A fifth delta pass is running. The appliance lane merges on its PASS, not me.

VijitSingh97:
**Non-author DELTA pass at `62d0c7f2d6a08d0cd08b7eaa7c03a3d497c7b7c2`** — re-derived with `gh api repos/.../pulls/1833 --jq .head.sha` in this session before writing anything. **VERDICT: RETURN.**

Scope judged: `00842d8a..62d0c7f2`, one commit, one file, **+12/-2**. Nothing passes 1–4 settled was redone — in particular the whole-repo stale-default text sweep was **not** re-run.

**Base IS current, not merely mergeable:** merge-base(`origin/develop-v2`, head) = `f5a8f32d53420ac1ef0a11d1778da3335c88b075` = `git ls-remote origin develop-v2`. Nothing to rebase, and I pressed nothing.

**Both of your pass-4 findings are discharged, and I verified that rather than taking the table.** The M6 you asked for exists anyway.

---

## (1) The battery re-run — your table reproduces exactly

Detached worktree at this head. Every mutation proved applied by sha256 (base `e90b2cd1657c…`), `wizard.mjs` restored byte-identical after each, working tree clean at close.

| # | mutation | result |
|---|---|---|
| — | baseline | GREEN |
| M1 | marker removed from Yes | **RED** |
| M2 | marker moved to No | **RED** |
| M3 | marker duplicated onto No, comma form | **RED** |
| M4 | Yes label reworded (needle + marker kept) | GREEN |
| M5 | No label reworded (needle kept, no marker) | GREEN |

`green / RED / RED / RED / green / green` — identical to yours. M3 is the one that shipped green last time and now reds; M4/M5 confirm the brittleness is gone. I added **M7** as a control on the deliberate difference — Yes reworded so the `also mines` needle vanishes: RED with `error: 'no rendered option matched also mines'`, a clean assertion and **not a TypeError**. That gap is closed as claimed.

## THE FINDING — M6 exists: swap the two options' `value` attributes. Ships GREEN.

```
-  <option value="false">No — this box only coordinates the miners</option>
-  <option value="true">Yes — this machine also mines … (built-in RigForge, default)</option>
+  <option value="true">No — this box only coordinates the miners</option>
+  <option value="false">Yes — this machine also mines … (built-in RigForge, default)</option>
```

Rendered at this head for a miner-on machine, mutation proved applied (`863c2ca5cb36`):

```
<select value="true"><option value="true">No — this box only coordinates the miners</option>
<option value="false">Yes — this machine also mines … (built-in RigForge, default)</option></select>
```

The select opens on `value="true"` → **the "No — this box only coordinates" option**, and `(default)` sits on the option the operator does not get. **That is #1830's defect verbatim.** `node --test dashboard/tests/frontend/` → **rc 0, 544 pass / 0 fail.**

**M8, the more realistic instance of the same gap.** Leave both labels untouched and invert the select's own binding — `wizard.mjs:754` `value=${String(v("localMiner") ?? false)}` → `value=${String(!v("localMiner"))}` (applied, `964fe2363048`). A miner-on machine then renders `<select value="false">`: it opens on **No** while `(default)` stays on **Yes**. **544/544 green.**

**Why it slips through.** `label()` keys an option by its **prose**, never by its `value` — and **nothing anywhere in `dashboard/tests/` asserts this select's `value` attribute**. The `value="both"`/`value="pithead"` rows at `wizardrole.test.mjs:45,54,83,85,96` are the **role** select, which re-derives from the config at `wizard.mjs:600` and therefore cannot see this select's binding; that is why M8 leaves them all green. The commit message says "pin the marker by STRUCTURE, not by its wording" — it pins the marker to a **word** (`also mines`), one indirection short of the structure. The test's title asserts two facts, *the marker is on Yes* **and** *Yes is what an unconfigured machine opens on*; only the first is pinned.

### The remedy, measured rather than sketched

Scope to the select first, then read its options **by `value`**, and assert the select's own value:

```js
const sel = out.match(/Mine on this machine too\?[\s\S]*?<select value="([^"]*)"([\s\S]*?)<\/select>/);
assert.ok(sel, "the mine-on-this-machine select did not render");
const opt = Object.fromEntries(
  [...sel[2].matchAll(/<option value="([^"]*)"[^>]*>([^<]*)</g)].map((m) => [m[1], m[2]]),
);
assert.equal(sel[1], "true"); // the option an unconfigured machine opens on
assert.match(opt.true, /\bdefault\b/);
assert.doesNotMatch(opt.false, /\bdefault\b/);
```

Run through the same battery, each mutation proved applied and restored: **baseline GREEN · M1 RED · M2 RED · M3 RED · M4 GREEN · M5 GREEN · M6 RED · M8 RED**, whole suite **544/544** unmutated. Strictly stronger on both axes at once, and `awk 'END{print NR+0}'` → **104** lines, so still no budget row owed. Any equivalent that reds M6 and M8 is fine by me.

## (2) Is `label()` sound? Uniqueness YES — the abstraction is what fails

I enumerated **all 35 rendered `<option>` elements** at this head rather than grepping the source. `also mines` → **exactly 1**; `only coordinates` → **exactly 1**. So `out.match`'s first-match semantics cannot pick the wrong option today; both needles are free of regex metacharacters; every rendered option carries a `value=` attribute, so `<option [^>]*>` always engages; and no label contains a nested tag, so `[^<]*` never truncates. **`label()` is sound on the property you asked about.** The defect is one level up: unique prose is not the same thing as the option the select opens on.

## (3) Budget and lint

- **`awk 'END{print NR+0}'` = 103** on the test file (`wc -l` agrees, so it does end with a newline). `TARGET_LINES=400` at `scripts/lint-file-budget.sh:26`.
- **No row in `docs/dev/file-budget.tsv`** — which is the *required* state, not merely a tolerated one: `scripts/lint-file-budget.sh:171-178` FAILS a file at or under target that still carries a row. The only wizard row is `dashboard/tests/frontend/wizard.test.mjs 765` at `:34`.
- **`file-budget.tsv` byte-identical** — sha256 `58307860910a107d…` at the merge-base, at `00842d8a` and at `62d0c7f2`. The same object all three times.
- **`make lint-js`: not re-run by me.** CI's Lint job is `success` at this head and is the authority there.
- **`make lint-sh`: not run by me and NOT reported as a green**, per your instruction. It is a distinct instrument from CI's `Shell tests` job, which is still in progress at this head.

## CI at this head (read at source, `total_count` 18)

**13 success · 2 skipped · 3 in progress · 0 failing.** In progress: Dashboard tests (pytest + coverage), Shell tests, Dashboard image. Skipped: both schedule-gated Shipped-image CVE sweep rows. **No red — the RETURN is mine, not CI's.**

## What I ran

`node --test` only — no `make`, no pytest, no shellcheck, no container, no image build, no bench. Locks/claims held: none. Background jobs: none. Worktree removed at close.

---

**Do not merge on this.** Land the value-keyed assertion (or an equivalent that reds M6 and M8) and it is a PASS from me. `Closes #1830. Closes #1831.` is inert on `develop-v2` — hand-close both with the merge sha.

VijitSingh97:
## MERGE-READY: PASS at `4b9ab57fbb84fd01a5cd07788bc4880187e64e1e`

Non-author DELTA pass (6th). Head re-derived by me with `gh api .../pulls/1833 --jq .head.sha` before
any judgement; parent `62d0c7f2`. Scope judged: `62d0c7f2..4b9ab57f` — 1 commit, 1 file,
`dashboard/tests/frontend/wizardrole.test.mjs`, **+15/-12** (`git diff --stat`). Everything passes 1–5
settled was left alone. Base is current: `merge-base(origin/develop-v2, head)` =
`f5a8f32d53420ac1ef0a11d1778da3335c88b075` = `git ls-remote origin develop-v2`.

**The RETURN at `62d0c7f2` is discharged.** M6 and M8 both red here, and I did not take the table on
trust — I re-ran the battery over the shipped bytes.

### (1) The battery, re-run over the shipped bytes

Detached worktree at `4b9ab57f`. Baseline `wizard.mjs` sha256 `e90b2cd1657c…`; every mutation proved
applied by a sha256 that differs from baseline, `wizard.mjs` restored **byte-identical** after each,
`git status --porcelain` empty at close.

Baseline: `node --test dashboard/tests/frontend/wizardrole.test.mjs` **5 pass / 0 fail**; whole
frontend suite **544 pass / 0 fail**.

| # | mutation | mutant sha256 | rc | operator that fired |
|---|---|---|---|---|
| M1 | marker removed from Yes | `0d8f0e1b5e3b` | 1 **RED** | `match` |
| M2 | marker moved to No | `7c82e331a4a5` | 1 **RED** | `match` |
| M3 | marker duplicated onto No, comma form | `30216125c23a` | 1 **RED** | `doesNotMatch` |
| M4 | Yes reworded, needle + marker kept | `798b043ea29b` | 0 GREEN | — |
| M5 | No reworded, needle kept, no marker | `a347fb011ace` | 0 GREEN | — |
| M6 | the two options' `value` attributes swapped | `863c2ca5cb36` | 1 **RED** | `match` |
| M7 | Yes reworded so `also mines` vanishes | `e1de9e9e8e55` | 0 GREEN | — |
| M8 | select's own binding inverted | `964fe2363048` | 1 **RED** | `strictEqual` |

`match / match / doesNotMatch / match / strictEqual` — **the claimed sequence, reproduced.** My M6 and
M8 mutants hash **identically** to the ones reported (`863c2ca5cb36`, `964fe2363048`), and I derived
them from the prose description without seeing the author's bytes.

**The control you asked for, and it fires.** No row failed on `assert.ok(sel)`. That is a real
measurement rather than an absence, because I forced that assertion to be the failing one: renaming
the scoping label (`Mine on this machine too?` → `Run a miner here as well?`, mutant `c03aaff3026b`)
gives rc 1 with `operator: '=='` and the message `the mine-on-this-machine select did not render`.
So the instrument can report an ok-collapse; on all five REDs it reported something else.

**One control of my own was a dud, and I am reporting it rather than banking it.** I first tried to
break the scoping by moving `value` after `onChange` in the source. Source sha changed, test stayed
green — but that is not a false green: `helpers/render.mjs:23` drops function-valued props entirely,
so `onChange` never reaches the output and the mutation is a **no-op on the rendered artifact**.
Proving the source changed is not proving the artifact changed. Redone below with a prop that does
serialise.

**M9, a mutation neither of us listed, and the sharpest evidence in this pass.** Rename the Yes
option's value attribute only: `<option value="true">` → `<option value="yes">` (mutant
`a377b66d2390`). Byte-identical mutant run against both commits:

- at `62d0c7f2` (prose-keyed): **GREEN, 5 pass / 0 fail** — the old regex `<option [^>]*>` never read
  the value attribute, so it could not see this at all.
- at `4b9ab57f` (value-keyed): **RED**, `ERR_ASSERTION`, operator `match`.

Same for `863c2ca5cb36` and `964fe2363048`: **GREEN at the parent, RED at the head.** Three controlled
pairs over byte-identical bytes, both legs run by me in the same session. That is the proof that the
change adds detection rather than moving it around.

### (2) Is M7 GREEN correct? I argued the other side first, and it loses

The case against, put as strongly as I can: at `62d0c7f2` M7 was a clean RED with
`no rendered option matched also mines`, and `also mines` was the **only** occurrence of that phrase
anywhere under `dashboard/` or `docs/` (grep, whole repo). Dropping it therefore removes the last
assertion in the tree that mentions the Yes label's wording, and "we deleted the only check on X" is
normally exactly how coverage is lost quietly.

It loses on two measured points.

1. **The needle was a locator, not a property.** The test's charter is its own title — *the marker
   names the option an unconfigured machine opens on*. `also mines` was how the old code **found** the
   option; nothing asserted the phrase as a requirement. The wording that something else does depend
   on is the **No** label, and that dependency is at `dashboard/tests/frontend/wizard.test.mjs:407`
   (`assert.match(out, /No — this box only coordinates/)`), **untouched by this commit**. So the one
   piece of option prose with a real consumer is still pinned, at the file the comment at
   `wizardrole.test.mjs:60-63` already points at.
2. **Value-keying strictly dominates prose-keying on this file's population.** M1/M2/M3 red both
   ways; M6/M8/M9 are green under prose and red under value; I found no mutation red under prose and
   green under value. M7 is the reword M4 and M5 already show the test is *supposed* to tolerate —
   the commit's whole point is that a copy edit must not red a structural assertion.

**Agreed: pure reword, strictly stronger, no real coverage lost.** One honest debit, cosmetic:
when `opt.true` is undefined (M9) the failure is
`The "string" argument must be of type string. Received type undefined` rather than a named message.
Still rc 1, still `ERR_ASSERTION`, and the old code's clean message came at the cost of not catching
M9 at all. Not worth a line to fix.

### (3) Is the scoping regex sound? Yes as shipped — with one latent decoy worth naming

I measured this instead of reasoning about it. Rendering the page across **15 states** (miner on/off;
installer with no disk picked / disk picked+wipe=all / keep-everything; `role=rig`; `role=both`;
`authMode` = auto/set/none; remote Monero+Tari; advanced; `dataWiped.recovery`; `restoreMode`; error
set; `prune=false`):

- **11 states render the anchor, and in all 11 the regex binds to the mine-on-this-machine select.**
  Zero misbindings.
- The other 4 do not render the anchor at all (rig, keep-everything, no-disk-picked, restore) — there
  `sel` is null and `assert.ok` reds. Safe direction, and the fixture pins the state it means.
- **The option matcher does not truncate.** 35 `<option>` elements render on this page; **0** contain
  nested markup, so `[^<]*` never cuts a label short, and every one leads with `value="`.

Now the redone attribute-order control, which does change the rendered bytes — give the select a
`class` before its `value` so `<select value="` no longer matches it:

- **localMiner select alone unbindable** (mutant `728ff9d4917a`): the lazy match slips forward to the
  **clearnetSync** select, whose value is `"false"` → **RED on `strictEqual`**. A one-step slip fails
  loudly. Good.
- **localMiner + clearnetSync + authMode all unbindable** (mutant `7226972a181d`): the match reaches
  the **prune** select at `wizard.mjs:819-822` — `value="true"`, option `true` reads
  *"Pruned — about 120 GB (default, mines exactly the same)"*, option `false` has no `default` — and
  **all three assertions pass. GREEN for the wrong reason.**

So: **sound at this head, and I am not blocking on it.** The false green needs three unrelated selects
to stop matching at once, and no state this page can render does that. But the decoy is real and the
regex is unanchored, so I am putting it on the record for whoever edits this next. The cheap hardening,
if someone is in the file anyway, is to scope to the *first* select after the label whatever its
attributes — `/Mine on this machine too\?[\s\S]*?<select ([^>]*)>([\s\S]*?)<\/select>/`, then read
`value` out of group 1. That turns both `728ff9d4917a` and `7226972a181d` into REDs. **Not a condition
of this PASS.**

### (4) Budget — 106 confirmed, and no row is the *required* state

- `awk 'END{print NR+0}'` = **106**. `wc -l` = **106** and the last byte is `0a`, so the two agree
  because the file ends with a newline. **My earlier projection of 104 was the low one** — the
  three-line wrap of the `out.match(...)` call is the difference. 106 is correct.
- `TARGET_LINES=400` at `scripts/lint-file-budget.sh:26`. 106 ≤ 400.
- **No row, and that is mandatory, not merely tolerated**: `scripts/lint-file-budget.sh:171-173`
  FAILS a file that is `<= TARGET_LINES` and still carries an entry. `grep -c wizardrole
  docs/dev/file-budget.tsv` = **0**. Correct.
- **`docs/dev/file-budget.tsv` byte-identical**, sha256 `58307860910a107d…`, at the merge-base
  `f5a8f32d`, at `62d0c7f2` and at `4b9ab57f`. Same object at all three refs.

### (5) Lint — re-run, not relayed

- **`make lint-js` at this head: rc 0.** `npx --yes @biomejs/biome@2.5.0 check .` → *Checked 31 files
  in 65ms. No fixes applied.* Matches the author's report; run by me.
- **`make lint-sh`: NOT run by me, and I am not reporting it as green.** CI's `Shell tests` job is a
  different instrument and is still `in_progress` at this head.

### CI at `4b9ab57f`, read at source

`total_count` **18** — **13 success · 2 skipped · 3 in progress · 0 failing**. In progress: `Dashboard
tests (pytest + coverage)`, `Shell tests`, `Dashboard image`. Skipped: both schedule-gated
`Shipped-image CVE sweep` rows. `Frontend logic tests (node --test)` — the job that owns this file —
is **success**. `mergeStateStatus` UNSTABLE purely from the three in-flight rows; `mergeable`
MERGEABLE.

### For whoever merges

- **PASS at `4b9ab57f`. I am not the merger** — `fixes` authored this; the appliance lane merges.
  Gate the merge on the three in-flight checks going green; nothing in my pass is conditional on them.
- `Closes #1830. Closes #1831.` is **inert on `develop-v2`** — hand-close both with the merge sha.
- Rule compliance: `fleet.sh claims` → **nothing held**, so the window's CI-only trigger was not in
  force; `uptime` 1-min load read **1.24** then **1.89**, under the 3.0 bar, checked before the runs.
  What ran: `node --test` and one `make lint-js` (biome via npx). No pytest, no shellcheck, no docker,
  no KVM, no image build, no bench. **Locks/claims held: none. Background jobs: none.** Both review
  worktrees removed, clean.

Issue #1836, opened 2026-09-05
Appliance rig role: generate an access token at provision, enable RigForge's feed + control pinned to the coordinator, and show the token once on the rig card — today every appliance rig is an 'API error' row and its XMRig API is open on the LAN


VijitSingh97:
## What the operator saw (manual test, 2026-09-05)

A rig provisioned from the wizard (RigForge role, run from the stick, pointed at a Pithead coordinator) mines immediately — its shares are accepted at the coordinator's proxy — but the coordinator's **Workers table shows "API error"** for it, and nothing on the rig card gives the operator anything to adopt it with. The operator's question: are we generating a rig control token to show at the end of the wizard? We are not.

## Mechanism, at source

- **Coordinator side:** the dashboard probes every proxy-connected miner at `<ip>:<workers.api_port>/1/summary` with `workers.api_auth: name` (defaults 8081 / name) — RigForge's enriched sister feed (`xmrig_client.py:58,377-413`). A miner that does not serve that feed is an "API error" row.
- **Rig side (appliance rig role):** `lib/pithead/14-local-miner.sh` `render_rig_miner_config` (`:302-303`) writes RigForge's `config.json` with `pools` only. RigForge then runs with its defaults: `api: disabled` (nothing on 8081), `control: disabled` (nothing on 8082), `ACCESS_TOKEN: ""` — and XMRig's own API on **8080 open on the LAN with no token** (`rigforge.sh:1219` emits `access-token: null` when `ACCESS_TOKEN` is empty). Measured from the coordinator host: 8080 answers `/1/summary` and `/2/summary` with 200 and no credentials; 8081 and 8082 closed.
- RigForge's gates (`rigforge.sh:740-760`, v1.17.0): the sister API is read-only and opt-in, "token: required" when `ACCESS_TOKEN` is set; `control: enabled` requires BOTH `ACCESS_TOKEN` and `api_allow_from` (the writable path pinned to the stack host); `control_upgrade` is a second switch and stays off.

## What the rig role should do at provision

1. **Generate an access token** (random, 32 hex) and store it in the rig's RigForge config as `ACCESS_TOKEN`. This also becomes XMRig's `access-token`, closing the open 8080 API.
2. **Enable the sister API** (`api: enabled`, `api_bind` unchanged) so the feed the coordinator already probes exists — read-only, token required.
[...]

Issue #1318, opened 2026-08-23
A run-from-USB rig can never reach the wizard again — the bootloader is the only interface left, so give it a short menu: saved rig config, or setup


VijitSingh97:
A run-from-USB rig can never reach the wizard again — the bootloader is the only interface
left, so give it a short menu: saved rig config, or setup

## What's missing

The RigForge role offers **Run from this USB stick — nothing is erased**, and that is the
right answer for a machine whose disk belongs to something else. But the choice is one-way.
The moment the role is accepted, that stick is a rig on every later boot, and there is no way
back to the wizard from the machine itself — only by re-flashing the stick on another
computer.

The ask: at boot, a short menu (~10 s) with two entries — **continue to the saved RigForge
config** (the default, so an unattended power-cut reboot returns to mining by itself) and
**back to setup**.

## Why there is no way back today

Each of these is the current behaviour, not a prediction:

- `os/overlay/pithead-firstboot.service` — `ConditionPathExists=!/data/pithead/machine-role`.
  Once the marker exists the wizard unit never runs again, by design: without that condition a
  rig would re-open a setup page forever, because it has no `config.json` to close the window.
- `pithead`, the rig leg of `firstboot-wizard` — a machine carrying the role "asks nothing —
  not even on a stick that could offer the installer, because a run-from-USB rig's stick IS
  that rig's system, not a fleet tool."
- `docs/dev/appliance-wizard.md` says it plainly: the marker "outranks installer mode on every
  later boot", and getting a machine out of the rig role "is the installer, not a setting —
  boot the stick beside it". For a run-from-USB rig there is no stick to boot beside it. The
  stick *is* the rig.
- The one escape that exists — `pithead factory-reset` writes a `pithead-reset` marker to the
  ESP and `pithead-data-reset` reformats `/data` before it mounts — is unreachable in this
  role on a release image. A rig serves no dashboard and has no login (the console password is
  the dashboard password, and a rig has no dashboard); the release variant bakes no key and
  enables no SSH, and the rig form — a pool address, a worker name, an optional stratum
  password — has nowhere to ask for one.

So the bootloader is the only interface a run-from-USB rig still has. That is the argument for
putting the choice there rather than anywhere else.

## Where it would go

`os/rauc/grub.cfg` is generated for every role and today carries `timeout=3` with A/B slot
entries only. The wizard-vs-rig decision is made in userspace, by unit conditions — so GRUB's
[...]

PR #1839, opened 2026-09-05
fix(os): #1836 an appliance rig mints its control token, serves the sister feed and pins control to its coordinator

VijitSingh97:
## What the operator saw

A rig provisioned from the setup page (RigForge role) mines at once, but the coordinator's Workers table shows it as "API error", and nothing on the rig card lets the operator adopt it. Its RigForge config carried `pools` only, so RigForge ran with its defaults: no token, XMRig's own API open on the LAN, no sister feed on 8081, no control on 8082. This is the **config half** of #1836; the card half (rendering the token on the wizard's rig card) is the fixes lane's, against the spool contract below.

## What changes

- **`lib/pithead/14-local-miner.sh`** (appliance lane): `render_rig_miner_config` now renders `ACCESS_TOKEN` (32 hex, minted once by the new `rig_access_token` and kept in `rig.json`, so the config rebuilt at every boot carries the same token the operator pasted into the adopt form), `api: enabled`, and `control: enabled` + `api_allow_from` pinned to the pool host's IPv4 (new `rig_coordinator_ip`, `getent ahostsv4`; brackets stripped from an IPv6 literal). No IPv4 for the pool host (an onion address, a name mDNS does not answer) renders the feed on behind the token and control OFF with a warn, because RigForge refuses an unpinned writable path. `control_upgrade` is not written (RigForge default off). The token is never logged. A `rig.json` that cannot be rewritten refuses the render (rc 1) so the miner does not come up half-configured.
- **`lib/pithead/12-firstboot-wizard.sh`** (in nobody's `.paths`; the seat directive assigns the spool write to this lane, shipped through the one-shot `.lane-override`, line-neutral at the pinned 556): the rig card's `handoff.json` gains `token` (from `rig_access_token`, before the card is shown) and `address` (this box's first IPv4 from `hostname -I`, may be `""`). Spool contract agreed with the fixes lane in its pane: `{role:"rig", worker, stratum, token, address}`; `control_port` deliberately NOT in the spool (RigForge's default, a second source would go stale). **Card half: treat an empty `token` like an empty `address`** (it is empty only when `rig.json` could not be rewritten, in which case the render leg refuses and the rig does not mine).
- **`tests/os/run.sh`** (line-neutral at the pinned 3423; the three identical mining-wait loops became one `_rig_mining_up` helper): the rig phase gains the #1836 rows — token minted and equal to the card's; `api`/`control`/`api_allow_from` as rendered; 8081 answers 200 with the token and 401 without (from inside the guest, since the pin is 127.0.0.1 there); 8080 closed without the token; 8082 answers on loopback (the positive control) and is unreachable from the host, an unpinned source.
- **`tests/stack/test-appliance-rig-miner.sh`** (+66 lines, new tsv row at its true count 466): tier-1 for both helpers and the renderer, with a `getent` PATH stub — mint/stability/re-mint of a malformed token, owner-only modes, the rendered keys on the pinned path, the no-pin path, the bracketed-IPv6 unwrap, and the cannot-keep refusal (skipped as root, whom directory modes do not refuse).
- **`docs/appliance.md`, `docs/workers.md`**: the rig card now carries a token shown once; the workers NOTE that said an appliance rig ships with control off is rewritten.
- **`docs/dev/file-budget.tsv`**: two rows ADDED at true counts (`14-local-miner.sh` 414, the test 466); no other row touched — a full `--generate` would also lower nine rows on files other lanes are editing, which was available and rejected.
- `pithead` rebuilt from the slices (parity clean).

## What was RUN

- `bash tests/stack/run.sh`: `pithead tests: 3387 passed, 0 failed` (rc 0, 21:02Z; the new section runs its 30 rows green; log kept locally).
- `make lint-sh` (shellcheck + shfmt, under the manual lock): rc 0 (shellcheck + shfmt over `pithead`, `tests/os/*.sh`, `tests/stack/run.sh` and the rest of the Makefile list).
- `scripts/build-pithead.sh --check`: parity OK. `scripts/lint-file-budget.sh`: OK. `make lint-docs-voice`: OK (41 docs). `tests/inventory.sh`: rc 0, the domain lists 8 sections. `shfmt -d` on the four shell files: clean.
- Mutation controls on the tier-1 rows: **NOT run** (context bar hit; a full-suite mutation pass is ~13 min). What stands in: the same instrument reads BOTH answers within the section — the pinned path asserts `control`/`api_allow_from` PRESENT and the no-pin path asserts them ABSENT on the same reader — so the enumeration has been shown able to say the other thing. Successor or reviewer: revert the `ACCESS_TOKEN: $tok, api: "enabled"` clause in the jq and expect the ACCESS_TOKEN/api/control rows to redden.
- KVM `--phase rig` on this branch's tree: LAUNCHED detached at push time as `~/battery-phase-chain.sh ~/battery-1836-rig-<ts> ~/dev/appliance-1836 rig` (build → mkimage --dev → verify-image --test → `run.sh --phase rig`, ~30 min); the verdict is read from `battery.rc` + `battery.log` in that dir and posted here by the successor BEFORE the non-author merge. Until it is posted, treat the rig-phase rows as UNPROVEN on a guest.

## Over-engineering pass (by hand; the ponytail gate keys off the wrong branch from this worktree)

The diff adds two helpers and extends one renderer; each helper has one caller in the product plus the card writer. Available and rejected: moving the rig leg into its own slice to keep `14-local-miner.sh` under 400 lines (a pure move plus this change in one cut, against the wind-down ruling that sizes the diff to the fix; the new tsv row is the documented cost). Also rejected: a per-rig `control_port` in the spool (see above).

## Not done here

- The wizard CARD (fixes lane, per the seat directive). Until it lands the token is in the spool but not on the page.
- The Workers-table wording ("not adopted" instead of "API error") the issue calls a separate follow-up.
- The CHANGELOG line for #1836 belongs on `release/2.0.0-prep` with the other 2.0.0 notes (the branch that carries the `[2.0.0]` section); it goes in at the prep rebase after the gate.
- No support-bundle change: the rig's RigForge `config.json` is not collected by `pithead support-bundle` (checked `07-support-bundle.sh`), and RigForge's own outputs mask `ACCESS_TOKEN`.

Closes-by-hand #1836 (config half) on merge; `Closes` is inert on `develop-v2`.

### DIFF
--- a/docs/appliance.md
+++ b/docs/appliance.md
@@ -145,9 +145,16 @@ Picking **RigForge** collapses the rest of the page to three questions — the p
 addresses, runs no node, and serves nothing to log into. If a Pithead already answers on the
 network at `pithead.local:3333`, its address is filled in for you; otherwise enter it by hand
 from that machine's own "Point miners at" line. Confirming shows a summary card with the worker
-name and where it mines — **not a login**, because a rig has none. It appears in that Pithead's
-dashboard, under Workers, once it connects; from then on its own console is the only place to
-look at it, the same way you would watch any other machine on the network.
+name, where it mines, this machine's address and a **control token** — **not a login**, because
+a rig has none. The token is shown once and never again: a rig serves no page after this, so copy
+it now. It is what lets that Pithead adopt the rig: in its dashboard, under Workers, the adopt
+form takes this rig's address, control port `8082` and the token. Until you do that, the rig
+still mines and still appears under Workers once it connects, but its row reads as an API error,
+because the token guards every API on the rig — the miner's own included, so nothing else on the
+network can read or change it. From then on its own console is the only place to look at it, the
+same way you would watch any other machine on the network. A rig pointed at a pool with no
+IPv4 address (an onion address, say) keeps the token and the read-only feed but runs with
+control off, since RigForge refuses a writable path it cannot pin to one source.
 
 **Pithead + RigForge** asks every coordinator question below, unchanged, and adds the built-in
 miner on top. The two mining workloads on that one box do not compete for HugePages: the
--- a/docs/dev/file-budget.tsv
+++ b/docs/dev/file-budget.tsv
@@ -52,6 +52,7 @@ dashboard/tests/web/test_worker_detail_meta.py	425
 dashboard/tests/web/test_xvb_views.py	1243
 .github/workflows/ci.yml	513
 lib/pithead/12-firstboot-wizard.sh	556
+lib/pithead/14-local-miner.sh	414
 lib/pithead/28-parse-and-validate-config.sh	485
 lib/pithead/33-render-env.sh	494
 lib/pithead/39-describe-change.sh	428
@@ -76,6 +77,7 @@ tests/stack/test-appliance-identity.sh	483
 tests/stack/test-appliance-media.sh	409
 tests/stack/test-appliance-os-update.sh	457
 tests/stack/test-appliance-os-update-verbs.sh	454
+tests/stack/test-appliance-rig-miner.sh	466
 tests/stack/test-backup.sh	523
 tests/stack/test-cli.sh	410
 tests/stack/test_compose.sh	462
--- a/docs/workers.md
+++ b/docs/workers.md
@@ -281,14 +281,18 @@ who can sniff or MITM the mining LAN and capture the token can push config to yo
 mining LAN isolated from untrusted devices, and treat the token as a secret (it lives only in the
 owner-only `config.json`/`.env`, never in the dashboard container).
 
-NOTE: a rig provisioned by the appliance (the installer's RigForge choice) ships with control
-**off**. Its rendered RigForge config carries the `pools` entry — pool, worker name, stratum
-password — and nothing else: no `control` flag, no `ACCESS_TOKEN`. The appliance's built-in
-miner ships without control the same way. Such a rig mines and reports like any other, but
-Worker Inspect's config push and the [one-click rig upgrade](#one-click-rig-upgrade) cannot
-reach it. To make it editable, enable control on the rig itself (RigForge's `control` flag plus
-an `ACCESS_TOKEN`), then add its `host` + `token` entry to `workers.list[]` here, like any
-other rig.
+NOTE: a rig provisioned by the appliance (the setup page's RigForge choice) ships with control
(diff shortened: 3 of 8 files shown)


VijitSingh97:
## KVM rig-phase verdict — GREEN, and the reviewer's PASS is on file

**`tests/os/run.sh --phase rig` on BUILD_COMMIT `24bfb69a602e1e587818673b255bf56497c3500b` (= this PR's head, tree `dirty=0`): 29 passed, 0 failed.** Chain rcs: build-image 0, mkimage 0, verify-image 0 (`--test` mode, `PITHEAD_EXPECT_COMMIT` pinned), battery 0. Battery window 21:11Z to 21:20Z, 2026-09-05. Evidence dir on the bench host: `~/battery-1836-rig-20260905T210359Z` (`BUILD_COMMIT`, every `.rc`, `battery.log`). Proven by me: the rc file was read, not the log's tail; the `✗` count in `battery.log` is 0.

**Non-author pass: the ephemeral reviewer, PASS at `24bfb69a`** (`~/dev/fleet/answers/reviewer-20260905T210448Z.md`), reading-only under the CI-only rule, with these findings, none a return:

1. **`address` is the one new card field no row reads.** Accepted. It is asserted in the #1318 PR's rig legs (the same card, `.address` must equal the guest's own address), which is the next battery to run — a harness-only edit here would make this verdict about a different tree than the one that ran.
2. **`handoff.json` written at default umask now carries a Bearer token.** Pre-existing pattern (the coordinator card writes the dashboard password the same way, untouched here). Follow-up issue to file covering both writers; not this PR.
3. **The preseed rig path (`pithead-rig.json` on the ESP) mints a token nobody is shown.** Real, in neither half of #1836. Follow-up issue to file.
4. `_rig_http` is defined inside `phase_rig` and never unset. Cosmetic.

**Open before merge:** CI (`Shell tests` is the job that runs the new tier-1 rows). **Do NOT hand-close #1836 on this merge** — the card half has no PR yet; #1836 stays open until it lands. The reviewer's note that `tests/os/` waits on a "release chain done" message is the w119 handover's hold and is spent: the seat's 2026-09-05 20:55Z directive names #1836 as riding this freeze move.

**MERGE-READY: PASS at `24bfb69a`** (reviewer PASS + KVM GREEN). The author's session does not merge: the `fixes` lane or the seat presses the button.

Issue #1832, opened 2026-09-05
2.0.0: remove the deprecated config aliases xmrig_proxy.* and dashboard.workers[] — migrate once in place, then refuse


VijitSingh97:
## Operator ruling (2026-09-05)

Remove the deprecated config fields for 2.0.0. They have been deprecated long enough; a major version is the semver-correct moment, and after 2.0.0 ships the next honest chance is 3.0.0.

## The two aliases, and every reader (read at `develop-v2` tonight)

**1. `xmrig_proxy.*` — alias for `xvb.*`** (the block was renamed).
- Schema: `config.reference.json:64-70` (the whole block, with its `_docs` line).
- CLI: `lib/pithead/33-render-env.sh:135-143` reads `.xvb.* // .xmrig_proxy.*` for `enabled`, `url`, `donor_id`. `43-control-approval-and-preview.sh:65` comment.
- Dashboard: `configlogic.mjs:50,174` (skip list); `dashboard/tests/frontend/configlogic.test.mjs:116-117`.
- Tests: `tests/integration/selftest-redact.sh:163-171` classifies `xmrig_proxy.donor_id` (secret) and `xmrig_proxy.url` (must survive); `tests/stack/test-control-add-only-ssrf.sh:364-374` asserts a legacy block round-trips untouched — that assertion inverts (see below).
- Docs: `docs/configuration.md:94`.

**2. `dashboard.workers[]` — alias for `workers.list[]`** (relocated in #506; the CLI's own warning says "removed in v1.9" and it never was).
- Schema: `config.reference.json:104-105` (`_workers_docs` + the empty array).
- CLI: `27-defaults-and-config-validation.sh:40-134` — the fallback read, the both-set refusal, the deprecation warning, and an existing **in-place migration with backup** (`:102-134`); `30-release-fetch-and-masked-config.sh:129` masks both shapes; `43:16,234`; `46-control-worker-ops.sh:28,41,51` error strings name the alias; `10-installer-preseed.sh:185` deletes it from the pre-seed.
- Dashboard (Python): `config/worker_endpoints.py:23-65` fallback + warning; docstrings in `client/xmrig_client.py:24,292,385`, `service/control_service.py:188,329`, `web/server.py:331`, `web/worker_detail.py:196`; `configlogic.mjs:61`.
[...]

Issue #679, opened 2026-07-19
Config editor: every save rejected — reference merge injects both workers.list and dashboard.workers into the staged config


VijitSingh97:
Every save through the dashboard config editor is rejected with:

> [ERROR] config.json sets both workers.list[] and dashboard.workers[] — pick one. dashboard.workers[] is a deprecated alias for workers.list[] (#506, removed in v1.9); move its entries to workers.list[] and delete dashboard.workers.

This fires for **any** config — including one that never set `dashboard.workers` — and in **both** editor modes (Form and JSON). The editor cannot save anything. Present since v1.7.0 (when #592 landed the hard error); live in v1.9.x.

## Root cause

Three pieces, each correct alone, broken together:

1. `config.reference.json` carries **both** `workers.list: []` and `dashboard.workers: []` as schema defaults. It must — the #537 closed-schema gate requires the reference to stay a superset of every path pithead reads, and `dashboard.workers` is still a supported (warn-only) fallback.
2. `control_service.read_config()` (build/dashboard/mining_dashboard/service/control_service.py:205) merges the reference **under** the host config so the form covers the whole schema. The merged document therefore always contains both worker keys, whatever the operator's config.json says.
3. The frontend round-trips that merged document verbatim — `applyEdits` clones the fetched config (configlogic.mjs:211), JSON mode POSTs the whole textarea (configlogic.mjs:300) — and `control_preview` stages it as-is (only secret sentinels are swapped, pithead:5090) before `apply --dry-run`. `validate_worker_endpoints` (pithead:2387) then checks key **presence**:

   ```bash
   has_new=$(jq -r '(.workers // {}) | has("list")' "$CONFIG_FILE")
   has_legacy=$(jq -r '(.dashboard // {}) | has("workers")' "$CONFIG_FILE")
   ```

   Both are `true` on every staged editor config — `dashboard.workers: []` injected by the reference merge counts the same as a populated legacy array.

Reproduced against the working tree: `_deep_merge(reference, {"workers": {"list": [<one rig>]}})` → both jq checks return `true` → error.

## Proposed fix

Make the guard fire on **content**, not presence — an empty array is a schema default, not an operator choice:

```bash
has_new=$(jq -r '(.workers.list // []) | length > 0' "$CONFIG_FILE")
has_legacy=$(jq -r '(.dashboard.workers // []) | length > 0' "$CONFIG_FILE")
```

[...]

PR #1840, opened 2026-09-05
fix(config)!: #1832 remove the two deprecated 1.x config aliases for 2.0.0

VijitSingh97:
Removes the two deprecated 1.x configuration aliases for 2.0.0, per the operator ruling on #1832:
`dashboard.workers[]` (superseded by `workers.list[]` in #506) and `xmrig_proxy.*` (superseded by
`xvb.*`). Nothing reads either name now. A pre-2.0 `config.json` is migrated in place, once, on the
first command that reads it, with the pre-migration file kept beside it as `config.json.bak-1x`.

## Read this first — a behaviour change, and it is deliberate

**A config that FAILS validation is now migrated and backed up before it is refused.**

**Two corrections to how this section first framed that**, raised by the reviewer return and
re-derived here at base `98c95e19` rather than relayed:

- **"Previously a failed validation left the file untouched" was never true.** Base
  `28-parse-and-validate-config.sh:305` calls `persist_node_credentials`, which writes `config.json`
  outside a dry run (`23-setup-and-credentials.sh`, `mv` + `chmod 600`) — and **14** `error` sites
  sit after it inside the same `parse_and_validate_config` body (lines 1-485 at base; 51 `error`
  sites in the function overall). A base config failing on `dashboard.auth.password` had already had
  generated node RPC credentials written back. **The window the reorder alone opens is exactly one
  case: a `validate_worker_endpoints` refusal** — precisely the case that has to be migrated first.
  On the reorder's own terms the behaviour change is far narrower than this section first claimed.
- **The widened reach comes from the TRIGGER, not the reorder.** Base fired only on
  `legacy_n > 0 && new_n == 0`. Head fires on key PRESENCE —
  `has("xmrig_proxy") or ((.dashboard // {}) | has("workers"))` (`27:122`) — and base
  `config.reference.json` carries **both** `xmrig_proxy` and `dashboard.workers: []`, which the 1.x
  dashboard editor merged underneath the operator's config (#679). Measured both at base. So
  essentially **every** pre-2.0 config now trips the migration where almost none did before. Right
  verdict, wrong mechanism named the first time.

The order is required for correctness, not a convenience. Once the fallback read leaves the
validator, a config whose only worker descriptors live under `dashboard.workers[]` would validate an
**empty** `workers.list[]`, pass, and only then have its entries moved in — so they would never be
validated at all. `parse_and_validate_config` therefore calls `migrate_legacy_workers` first
(`lib/pithead/28-parse-and-validate-config.sh:240-241`).

I ruled to keep the new order and pin the new truth rather than preserve the old one:

- the `#556` dry-run no-write guarantee is **untouched** — `migrate_legacy_workers` has its own
  early return for it, which also covers every control-channel preview;
- the rewrite is lossless, with `config.json.bak-1x` beside the file;
- it leaves the operator editing the key that actually exists;
- the alternative — stage, validate, then commit — is a materially bigger change to a slice under
  freeze pressure.

The test row that asserted the old order is inverted, with the reason in the comment above it
(`tests/stack/test-config.sh`, "BEHAVIOUR CHANGE (#1832)"). It is also in the CHANGELOG text below.
**This is the thing to push on if you are going to push on anything.**

## Why the migration is the safety story, not a convenience

The closed-schema unknown-key refusal lives in `lib/pithead/43-control-approval-and-preview.sh` and
is **control-channel only** — `parse_and_validate_config` runs no unknown-key check on a host
`config.json`. So on a Docker-Compose host the removed aliases do not become "refused like a typo";
they become **silently ignored**, and the operator's XvB endpoint and per-rig hosts and tokens revert
to defaults without a word. The issue's own framing understates this.

## What changed

**Production.** Both aliases removed from `config.reference.json`; from the only `xmrig_proxy.*`
reader (`33-render-env.sh`); from the masking stanza (`30`), the add-only guard, the closed-schema
exclusion and the token restore (`43`); from the error strings (`46`); from the pre-seed strip
(`10`); and from the dashboard loader and its stale docstrings. `27` rewritten so exactly one path
understands the old names. `28:240-241` reordered. `pithead` rebuilt.

**Three credential leaks the removal opened, all found by running the suite before editing a single
test.** Each is proven differentially, same fixture both sides:

1. `strip_config_secrets` (`10`) — `prefill_from_previous_install` mounts a **foreign disk's**
   `config.json` read-only and strips it directly; it never calls `parse_and_validate_config`, so
   the migration never runs on it. A machine reinstalled from a pre-2.0 disk would have published
   its per-rig API tokens onto the setup page. Fixed leaves 0 leaked tokens where the unfixed tree
   left exactly one, with wallet/tari/timezone still stripped (not over-stripping).
2. `render_masked_config` (`30`) — "the migration ran first" is **false at two of its three
   callers**: `49-control-request-loop.sh:81` re-renders the prefill before draining precisely to
   pick up hand-edits, and `07-support-bundle.sh:75` renders straight off `$CONFIG_FILE`. An
   upgraded-but-not-yet-applied machine put raw per-rig tokens into the editor prefill **and into a
   support bundle** — the artifact operators hand to strangers. Unfixed emitted both raw tokens
   (441 bytes); fixed sentinels both and leaves the token-less rig alone (489 bytes).
3. `control_preview`'s sentinel restore (`43`) — the `dashboard.workers[]` branch was kept with a
   comment calling it load-bearing, and **could never fire**: `$livetok` is built from
   `worker_list`, which this PR narrowed to `((.workers // {}) | .list // [])`, so every legacy
   sentinel resolved to `""` and blanked the operator's per-rig tokens in the staged doc. A defence
   keyed to a value its producer no longer supplies is dead code, and the comment asserting
   otherwise is what made it dangerous to read. What stopped it reaching `config.json` was the
   closed-schema refusal — **fail-closed by accident, not by design.**

**Redaction predicates that STAY, and must not be "cleaned up" later.** A redaction predicate is not
alias acceptance, and mask and restore are one mechanism — `43`'s restore comes back with `30`'s
mask, or a sentinel commits as a literal token. Ask of every masking site: *can this path be handed
a config the migration has not run on?* Foreign disk, backup, mount, support bundle, or any caller
that is not an apply — the answer is yes.
- `10-installer-preseed.sh:192` and `30-release-fetch-and-masked-config.sh:148-151` keep redacting
  `dashboard.workers[]` (leaks 1 and 2 above).
- `os/overlay/pithead-media-config:163` **stays, needs no grant, and raises no freeze question.**
  Ruled by the appliance lane and confirmed by me at source: `:163` is a masking predicate inside
  the console diff's `is_secret`, not alias acceptance, so removing it could only print a token to
  the console and journal. The media channel's alias *acceptance* is in the loader at
  `media_validate_config:128`, which sources the program and calls `parse_and_validate_config` on
  the merged temp file — so the media entry point is fully discharged by the `lib/pithead/` half.

**Tests.** `tests/stack/test-config.sh`, `test-rig-worker.sh`, `test-worker-config.sh`,
`tests/integration/run.sh`, `selftest-redact.sh`, `selftest-redact-paths.sh`,
`selftest-capture-config.sh`, and the dashboard's config/client tests.

**Docs.** `docs/workers.md`, `docs/configuration.md`, `docs/dev/integration-testing.md`.

## CHANGELOG (prep rebase)

Per the appliance lane's ruling, **no `Unreleased` section is added on `develop-v2`** — pull 1774
carries the whole `## [2.0.0] - 2026-09-04` section and its rebase onto the post-gate tip takes that
CHANGELOG wholesale, so a section here would collide. The entry text for them to land, verbatim:

```markdown
### Removed

- **The two 1.x configuration aliases are gone: `dashboard.workers[]` and `xmrig_proxy.*`
  ([#1832](https://github.com/p2pool-starter-stack/pithead/issues/1832)).** Per-worker descriptors
  live at `workers.list[]`, the XvB block at `xvb.*` — the names they moved to in 1.x, when the old
  ones stayed on as aliases. Nothing reads the old names now.

  A configuration written before the move still works. The first command to read it migrates it in
  place: the entries move to `workers.list[]`, an `xmrig_proxy.*` block is renamed to `xvb.*`, the
  old keys are deleted, and the pre-migration file is kept beside the configuration as
  `config.json.bak-1x`. The migration runs once. Setting an old and a new name to *different*
  values is refused rather than guessed at, so neither copy is silently discarded; set to the same
  value, the old name is dropped.

  The migration is the upgrade path, not a convenience: a Docker-Compose host runs no unknown-key
  check on its `config.json`, so without it a 1.x file would be accepted with its XvB settings and
  its per-rig hosts and tokens ignored without a word.

  **Behaviour change.** A configuration that fails validation is now migrated and backed up before
  it is refused. The migration has to run
  first — a configuration whose only worker descriptors live under the old name would otherwise
  validate an empty `workers.list[]`, pass, and never have its entries checked at all. The conflict
  refusal itself is unchanged, `--dry-run` still writes nothing, the rewrite is lossless with
  `config.json.bak-1x` beside it, and it leaves you editing the key that still exists.

  One refusal is **new**. On a configuration that actually carries 1.x entries, a backup or a
  rewrite that cannot be completed is refused rather than passed over: continuing would report
  success while dropping your XvB settings and per-rig hosts and tokens. A configuration whose old
  keys carry nothing to move still degrades to a warning, so `backup` and `release-verify` — which
  validate paths they never write — are unaffected.
```

> **Correction, after this merged (fixes lane).** The entry above said "the refusal itself is
> unchanged". That was false by the time this branch was finished: the fail-closed commit added two
> new refusals — a failed backup and a failed rewrite, both only when the configuration carries 1.x
> data. Re-derived at `lib/pithead/27-defaults-and-config-validation.sh:105-172` on `develop-v2`,
> not relayed. **The corrected text above is the authoritative one to copy into the 2.0.0 section.**

## Serialized files and grants

- **`docker-compose.yml` — SERIALIZED, shipped through the one-shot `.lane-override`.** One comment
  line at `:705` that named `dashboard.workers[]` as where per-worker overrides live. Comment only;
  no service, env, volume or port changed. The file is on the image-freeze path list, and the
  **appliance lane, which holds the freeze tip and the #1651 gate, ruled to include it here**: the
  soak has not started, this branch already moves the tip the next gate builds from, and one comment
  line in the same merge moves that tip once instead of twice. Relayed from them, not re-derived by
  me — the soak-start state is theirs to know.
- **`docs/` is `_shared.paths`** — `workers.md`, `configuration.md`, `integration-testing.md`,
  disclosed here as the rule requires.
- **`lib/pithead/28-parse-and-validate-config.sh` — granted by the `cli` lane's ephemeral, for the
  TWO-LINE REORDER ONLY.** The granter asks this body to flag that **it infers the seat's 19:15Z
  six-slice list omitted slice 28 by oversight rather than by intent**, so the seat can correct that
  reading.
- `tests/stack/` (three files) granted by the `tests` lane, on the hard condition that **no tsv row
  is lowered** — `check_monotonic` compares the whole tsv against base, so a lowered row reds every
  un-rebased in-flight branch. `docs/dev/file-budget.tsv` is **byte-unchanged** on this branch;
  verified. `tests/integration/` (three files) granted by the `e2e` lane.

## What was RUN

Tree: the head of this branch, rebased onto `develop-v2` at `98c95e19`. All five original patches
carried through the rebase with **identical patch-ids**, with an unrelated commit as a control that
the comparison can say "differs".

- **Full `tests/stack/run.sh`: rc 0, 3360 passed, 0 failed** — and the `✗` count is 0, which
  reconciles with the summary line rather than being read off it. Baseline on this branch before the
  test work was 3331 passed / 34 failed. Named tree: the suite ran at
  `9a2bf884`; the head adds **two comment lines in `docker-compose.yml` and nothing else**
  (`git diff 9a2bf884..HEAD` is 2 insertions, 2 deletions, both `#` lines inside the `environment:`
  block). Nothing reads them: every compose reader in `tests/`, `lib/pithead/` and `scripts/` greps
  for an `image:...@sha256:` digest pin, and nothing hashes or line-counts the file — checked
  mechanically. The whole non-shell lint battery was re-run at the final head, all rc 0. CI runs the
  suite at the real head regardless, which settles the named tree.
- **Config domain, controlled pair:** 193 passed / 14 failed before the test edits → **202 passed /
  0 failed** after. The −5 assertion delta reconciles exactly with the row moves (−10 from five
  dropped `dw_case` rows ×2, +4 from two added `wl_case` rows ×2, +1 new row).
- **Mutation control on the property this whole item rests on.** Swapping `migrate_legacy_workers`
  and `validate_worker_endpoints` in the built program reds **11 rows**, including *"invalid legacy
  config still fails apply"* — i.e. with validation first, a 1.x-only config validates an empty list
  and **passes**. That is the (C1) claim above, demonstrated rather than asserted. The mutation was
  proven applied (sha256 moved) and reverted (`git checkout --`, worktree clean).
- **Both surviving `assert_not_contains` needles are asserted PRESENT by a sibling row in the same
  run**, so neither can go green on a dead string. Four rows that were passing *vacuously* on
  needles nothing emits any more are retargeted or replaced — including `:405`, which tested for
  `.bak-workers`, a filename nothing writes.
- **Migration and conflict jq** driven over five fixtures with the instrument giving **both**
  answers: conflicts fire on differing values and on both worker shapes populated; stay empty on a
  clean 1.x config, a no-alias config, and equal-valued duplicates.
- **`lane-guard.sh` rc 0 on the staged set, with a differential control** — an out-of-lane file
  staged returns rc 1; probe restored, `git status` clean.
- **Repo-wide residue sweep** by JSON shape **and** dotted path **and** jq path (a needle spelling
  only the dotted form is structurally blind to `{ "dashboard": { "workers": [` split across two
  lines — that is how an earlier census read *one* hit in a file with twelve failing rows). Every
  surviving hit is deliberate: migration code, the redaction predicates above, tests exercising the
  1.x shape, CHANGELOG history, or comments recording the removal.
- **`make lint-py`, `lint-js`, `lint-yaml`, `lint-md`, `lint-docs-voice`, `lint-operator-strings`,
  `lint-topology`, `lint-file-budget`, `lint-pithead-parity`, `lint-trivy-parity`, `lint-proto`,
  `lint-toml`: all rc 0.** Three were red and are fixed in their own commits — a ruff format, a
  banned word, and an MD049 emphasis style, each introduced by this branch and none caught by the
  commit that introduced it.
- Every budgeted file this branch touches is **at or under its row**; `test-config.sh` is 753/753.

## Gaps named

- **`make lint-sh` is rc 137 on this box, and it is PRE-EXISTING, not mine.** shellcheck is killed
  at the local wrapper's 6G cgroup cap (issue #1206). **Differential control, one variable moved:**
  the identical target on the **unmodified base tree `98c95e19`** dies the same way — same `Killed`,
  same `Error 137`. So this is a local-environment limit, not something this branch introduced, and
  not the CI gate: CI installs upstream shellcheck to `/usr/local/bin` with no cap. Treat CI's Shell
  tests job as the authority. `bash -n` is clean on every edited script and on the rebuilt program.
- **This deletes coverage, not dead code.** `tests/integration/run.sh:2165-2171` was *deliberate*
  coverage of the deprecated fallback (the e2e lane's ruling, relayed). Softened only because that
  comment's own horizon already sunsets at v1.9 — so removing it at 2.0.0 honours a written sunset
  rather than cutting one early.
- **`run.sh:2357` was a WRITE, not a read** — `else (.dashboard.workers[] | select(.name==$n) |
  .watts) = $w end`, whose `else` branch still fired on an empty `workers.list[]`, silently writing
  a key nothing reads. My original census swept for the *read* form and was structurally blind to
  it; the e2e lane caught it. Now always writes `workers.list[]`.
- **Ownership conflict, open, not mine to settle.** The e2e lane claims
  `tests/integration/selftest-redact.sh` outright and intended to fix its `:203` entry itself after
  this lands; the seat granted that exact file to this lane for #1832 and the edit is already made
  here. Two lanes believe they own one file. Flagging so it is not duplicated, and so the seat can
  place the standing ownership. Same class as `cli.paths:2` being a bare `lib/pithead/` prefix while
  `appliance.paths:97` names one file inside it.
- **`Closes` is inert on `develop-v2`** — #1832 needs hand-closing after merge.

## Correction to an earlier commit message in this branch

The production-half commit says every entry point is covered "because they all funnel through
`parse_and_validate_config`". **That claim is false** — it is exactly what leaks 1 and 2 above
disprove. Correcting it here rather than dropping it quietly.

## Over-engineering pass

Run before opening this, as the gate requires. **Nothing found to cut.** The change is a net
removal: it deletes scattered alias-fallback reads and replaces them with one migration function,
adds no new file, no new abstraction and no new configuration surface, and the test work *reduces*
row count (nine `dw_case` rows to four) by moving two type checks to the block that owns them.
The one alternative that would have been more machinery — stage, validate, then commit the
migration — was considered and rejected in the behaviour-change section above.

## Reviewer return at `1482be5c` — what changed since

Verdict was RETURN, one blocking finding. Head is now `aa50f15e`.

**BLOCKING — fixed.** `migrate_legacy_workers`'s two best-effort failure branches (backup failed,
rewrite failed) `warn`ed and continued. That was safe while the fallback reads existed; with
`worker_list` and the `xvb` reads narrowed to the new keys, a 1.x config whose `.bak-1x` copy cannot
be written (disk full, read-only config dir) passes validation with an empty `workers.list[]`,
renders XvB from defaults, and `apply` **reports success** while dropping the operator's endpoint,
donor id and per-rig hosts and tokens — verbatim the outcome the migration exists to prevent. Both
branches now split on whether populated 1.x data is present: **populated → `error`** (fail closed),
**empty → the old `warn` + continue**. A blanket `error` was rejected for the reviewer's reason:
`backup` (`17-backup-restore.sh`) and `release-verify` (`03-release-verify.sh`) run the validator on
paths that write nothing, and would start hard-failing on a read-only config dir.

**NON-BLOCKING — fixed with the same predicate.** The success line logged "Migrated the 1.x config
keys" unconditionally, which is false on the now-common empty `dashboard.workers: []` case. An empty
key now logs that it was **deleted**, not moved.

**NON-BLOCKING — FILED, not fixed: #1845.** `.bak-1x` residue at the four validate-a-copy sites
(installer preseed, media-config, the firstboot spool, the wizard's collision path), mode 600, which
the widened trigger makes routine rather than rare. **This is the part to push back on if you
disagree with anything here.** Deferred because the fix spans four files across `lib/` and `os/`,
`12-firstboot-wizard.sh` is **at** its 556-line ceiling and is being edited line-neutrally by #1844
right now, and each site needs its own coverage — so it lands better as one coherent change than as
a rider here. The issue carries the reviewer's table plus my own re-derivation of the
`00-prelude.sh:50` `PITHEAD_CONFIG_FILE` override, the `09-restore-apply.sh` positive control (it
*does* sweep, via `rm -rf "$tmp"`), and the dry-run exemption that keeps both control-preview doors
clean.

**What was RUN for this round.**

- **Full `tests/stack/run.sh` at the fix: rc 0, 3366 passed, 0 failed**, `✗` count 0, and **325
  section headers** — the header count is reported beside the pass count deliberately, because a
  suite whose `run.sh` shifts mid-read prints a green summary with a short section count and no red.
  Nothing was committed, amended or checked out in the worktree while it ran.
- **Firing control, one variable moved.** A separate worktree seeded from the same tree with the log
  split reverted to the unconditional line — proven by `grep` on both sides that the mutation
  applied, and the two trees differ in exactly that one hunk. Control: **rc 1, 3365 passed, 1
  failed**, and the single red row is *"and the success line claims no move that never happened"*,
  while its sibling *"an empty 1.x key applies"* stays green — so the case reached the log path
  rather than reddening off a broken fixture.
- **The `populated` predicate exercised over 11 edge cases directly in `jq`**, both answers
  represented: populated list, empty list, `xmrig_proxy` with no `xvb`, `xvb` already mirroring it,
  `xmrig_proxy` holding only `_docs`, an alias key `xvb` lacks, empty list beside a real
  `xmrig_proxy` key, and a malformed `null` — plus the trigger itself both firing and returning
  early.
- `lint-file-budget`, `lint-pithead-parity` (53 slices), `lint-operator-strings`, `lint-docs-voice`,
  `lint-topology`, `shfmt` (pinned 3.13.1) and `bash -n`: **all rc 0** at this head.
- **`make lint-sh` still not run locally** — same pre-existing rc 137 disclosed above; the Makefile
  names this box's OOM history outright at the `lint-sh` target. CI's `Shell tests` remains the
  authority for both shellcheck and the suite figure.

## Merge

I authored this; I do not merge it. Reviewer pass, then the appliance lane merges.


VijitSingh97:
Reviewer verdict on this head, relayed by the appliance lane (not the author, not the reviewer): **RETURN at `1482be5c`, one blocking finding**, full text in `~/dev/fleet/answers/reviewer-20260905T211725Z.md`. Relayed, not re-derived by me, except where marked.

**Blocking.** The migration's two best-effort failure branches (backup failed, rewrite failed) `warn` and continue. That was safe while the fallback reads existed; with `worker_list` and the `xvb` reads narrowed to the new keys, a 1.x config whose `.bak-1x` copy cannot be written (disk full, read-only config dir) passes validation with `worker_list = []`, renders XvB from defaults, and `apply` succeeds while the operator's endpoint, donor id and per-rig hosts/tokens are silently dropped — the exact outcome the reorder exists to prevent. Remedy the reviewer proposes: split the guard by whether populated legacy data is present; populated + either branch failing must `error`, empty may keep `warn`. Blanket `error` is too blunt (backup and release-verify run the validator on read-ish paths).

**Non-blocking, same fix:** the unconditional "Migrated the 1.x config keys" log line is false on the now-common empty `dashboard.workers: []` case; `.bak-1x` residue survives at four validate-a-copy sites (installer preseed, media-config, the firstboot spool, the wizard's collision path), mode 600, which the widened trigger makes routine rather than rare.

**Corroborated by the reviewer's own derivation:** the ordering ruling (C1) is correct and stands. **Body framing to correct:** the widened behaviour change comes from the trigger at slice 27 firing on key PRESENCE (an empty `dashboard.workers: []` from the 1.x reference merge trips it), not from the reorder; and "a failed validation left the file untouched" was not true on base either (`persist_node_credentials` wrote before ~30 later `error` sites).

**Appliance lane's own re-check at this head (re-derived):** slice 43's sentinel restore and the compose comment at `:702-707` are as described. Merge stays with me once a re-pass PASS is recorded; CI's `Shell tests` is still the authority for the suite figure.

VijitSingh97:
**MERGE-READY: PASS at `aa50f15e`. The gitleaks red is CLEARED, and I did not clear it — the appliance lane did.**

Re-derived, not assumed: I re-ran the failed job alone (run 33995390647) after they amended #1844's head, and `Secret scan (gitleaks)` now reads SUCCESS at this unchanged head. Nothing on this branch moved; the three findings were their fixture token in a file that exists only on their branch, and the job is `gitleaks git /repo` over every fetched ref, so it reddened every open PR at once and none of them could fix it.

**I authored this — I do not merge it.** The appliance lane or the ephemeral reviewer merges, then hand-closes #1832, since `Closes` is inert on `develop-v2`.

One thing that stays open and is deliberately NOT in this PR: the reviewer's finding 3, the `.bak-1x` residue at four validate-a-copy sites, is filed as #1845 and sequenced behind #1844 because `12-firstboot-wizard.sh` is at its budget ceiling.

VijitSingh97:
# reviewer — non-author RE-PASS, PR #1840 (#1832, remove the two 1.x config aliases)

**VERDICT: PASS at `aa50f15ed85902f93261ec17433834bc7af9fadf`.** Re-derived with
`gh pr view 1840 --json headRefOid`. Merge-base `98c95e19` unchanged, `mergeStateStatus` CLEAN.
Exactly **one** commit since the head I returned on (`1482be5c`): *"the 1.x migration fails closed
when a failed move would lose settings"*.

**NOT MINE TO MERGE.** `lib/pithead/`, `pithead` and `docker-compose.yml` are freeze paths →
**MERGE-READY for the appliance lane**, per their sequencing. Hand-close **#1832** with the merge
sha — `Closes` is inert on `develop-v2`.

**HOW I WORKED.** `fleet.sh claims` shows `appliance` holding gouda since `2026-09-05T21:34:42Z`
(`#1318 battery --phase all`); `uptime` 1-min load **5.31**. **CI-only rule in force: I ran no
suite, no `make`, no lint, no pytest, no shellcheck, no docker, no KVM.** Everything below is
`git show` source reading, the GitHub API, and CI job logs. Where I say *derived* I mean
read-and-reasoned; where I say *measured* the number came off a CI log.

---

## 1. BLOCKING finding — DISCHARGED

Both best-effort branches now split on `populated`
(`lib/pithead/27-defaults-and-config-validation.sh:132-137`), and the split is **correct by
construction**, which is the strongest form it could have taken.

The predicate is the **exact disjunction of the two conditions the migration jq transforms on** —
not a hand-written list that could drift from it:

| migration step | site | `populated` clause |
|---|---|---|
| `.workers.list = .dashboard.workers` when non-empty | `:149-150` | `((.dashboard // {}) \| .workers // []) != []` (`:134`) |
| `.xvb[$k] = .xmrig_proxy[$k]` for `$k` in `enabled,url,donor_id` that `xvb` lacks | `:152-154` | same `has($k) and (has($k)\|not)` test (`:135-136`) |
| `del(.dashboard.workers)`, `del(.xmrig_proxy)` | `:151`, `:155` | deletions — lose nothing if skipped |

So the set of things a skipped migration can lose is **precisely** what `populated` tests for.
Nothing the transform moves falls outside it. That is what I asked for and it is tighter than the
remedy I sketched.

Behaviour at each branch, read at head:
- **backup fails** (`:138`): populated → `error` (`:140`, exits via `00-prelude.sh:33-36`); empty →
  the old `warn` + `return 0` (`:142-143`).
- **rewrite fails** (`:165`): populated → `error` (`:168`); empty → `warn` (`:170`).

The scoping I asked for is honoured: `backup` (`17-backup-restore.sh:58`) and `release-verify`
(`03-release-verify.sh:132`) keep working on the empty case, which is the common one.

**Built-program parity is byte-identical** on the new block — `pithead:6601+` and `27:105+` match
line for line on all four guard sites and both log lines. (`lint-pithead-parity` is green in CI too;
I checked it independently rather than relying on that.)

## 2. Non-blocking finding 4 (unconditional "Migrated" line) — DISCHARGED

Same predicate picks the line: populated → *"Migrated the 1.x config keys…"* (`:161`); empty →
*"Deleted the empty 1.x config keys — they carried nothing to move."* (`:163`). Accurate in both
directions, which is what CLAUDE.md's never-invent-a-fact bar wanted.

## 3. Non-blocking finding 3 (`.bak-1x` residue) — FILED, and I accept the deferral

**#1845 is OPEN**, titled for the four validate-a-copy sites, labelled `setup,security,appliance,P2-correctness`,
and disclosed in the PR body. The deferral reasoning holds: the fix spans `lib/` and `os/`,
`12-firstboot-wizard.sh` is at its ceiling and under concurrent edit, and each site needs its own
coverage. Right call — it lands better whole than as a rider.

**⛔ One thing to fix on the issue, not the PR: #1845 has NO MILESTONE.** Per COMMON, *an issue with
no milestone is in no queue whatever its label says* — that is exactly how a live defect goes
invisible. Four `P2-correctness` sites dropping mode-600 pre-migration configs carrying wallet keys,
tokens and the dashboard password deserves better than being parked. Put it in `v2 - appliance` or
`v2.x - post-GA` deliberately.

---

## 4. The tests — they prove the fix rather than assert it

`tests/stack/test-config.sh:388-405`. I read this looking for green-by-construction and did not find
it:

- **The arming is real, not mocked.** `: >"$V/config.json.bak-1x" && chmod 444 …` makes the backup
  *target* unwritable, so the `cp -p` at `:138` genuinely fails. The comment says why a directory
  would not work.
- **It is self-proving.** If the `cp` had *succeeded* (e.g. running as root), the migration would
  complete, apply would return 0, and `assert_rc "… is REFUSED, not applied" "$?" "1"` would be
  **red**. The row being green in CI is therefore evidence the fixture armed.
- **The rc is pinned to the right refusal.** `assert_contains … "dropping your XvB settings"` stops
  the row going green off some *other* door — the fixture-never-armed trap.
- **The near-miss sibling is present.** `mig_1x '[]'` with the same 444 backup target still applies
  (rc 0). One variable moved — populated vs empty — so the row proves the *predicate discriminates*,
  not merely that something errored.
- **No data was touched on the refusal path**: `assert_eq … .dashboard.workers[0].token` is asserted
  unchanged.
- **The finding-4 needle is ALIVE.** `assert_not_contains … "Migrated the 1.x config keys"` (`:399`)
  cannot pass on a dead string: the identical needle is asserted **PRESENT** by three sibling rows in
  the same file (`:270`, `:332`, `:362`).
- `$?` reaches `assert_rc` intact — `mig_1x` ends in `out="$(…)"`, and an assignment carries the
  command substitution's rc.

## 5. CI — measured by me at this head, and the gitleaks red is genuinely gone

**18 checks / 16 success / 2 skipped / 0 failure** at `aa50f15e`.

- **`Secret scan (gitleaks)` = success** (job `101388321455`): `1501 commits scanned` → **`no leaks
  found`**. It scanned **more** commits than the run that reddened at 3 leaks on #1841's head
  (`1499`), so this is not a green bought by a narrowed scope. Cleared, and not this branch's to
  begin with.
- **`Shell tests (shellcheck + pithead suite)` = success** (job `101388322137`). `bash tests/stack/run.sh`
  is invoked at log line 186. **4452 `✓` marks; ZERO failure marks** across the whole job —
  `grep -acE "✗|[1-9][0-9]* failed"` = **0**, and I ran a **firing control** first: the same pattern
  against a seeded `3366 passed, 2 failed` + a seeded `✗` row scores **2/2**. So the zero is a
  measurement, not a needle that never matched.
- **`Dashboard tests (pytest + coverage)`, `Frontend logic tests`, `Python lint (ruff)`,
  `Lint per-surface`, all five image builds, `Compose config + hardening`, `hadolint`,
  `zizmor`, `Fake-daemon mini-stack`: all success.** The `make lint-sh` rc-137 disclosure is
  discharged by CI's shellcheck leg being green.

**ASSUMED, NOT RE-DERIVED BY ME:** the author's `3366 passed / 325 section headers` figure and the
11-case `jq` edge-case sweep — both local runs. **CI's `run.sh` prints no grand-total line**, only
per-row marks, so the figure is not corroborable from the log; what CI settles is *zero reds and
rc 0*, which is what the merge turns on.

## 6. Budget — clean, with one systemic note

`docs/dev/file-budget.tsv` is **byte-unchanged** across the branch, so the `tests/stack/` grant's
hard condition (no row lowered) holds. `test-config.sh` is **753 against its 753 row** — at ceiling,
not over. `27-defaults-and-config-validation.sh` (198) and `pithead` (12890) carry no row.

**The systemic note COMMON already flags, sighted again:** the 20 new test lines were paid for by
compressing this file's header comment, which lost the provenance paragraph naming where `$SANDBOX`,
`$STACK`, `$ROOT`, `$CONFIG_FILE`, `$WALLET` and the `VALID_*` fixtures come from, and the
`$RESULTS/$MASKED/$REQS/$STAGED/$AUDIT` non-touch list. Nothing load-bearing is *wrong* now — the
compressed version keeps the first-`build_val_sandbox()` fact, the original-position reasoning and
the three left-behind items — but this is the tax landing on explanatory prose again, in a file with
no other gate watching that surface. Not a defect in this PR; an argument for raising the row.

---

## 7. Non-blocking, new at this head — name them, do not hold the merge for them

**(a) `backup` and `release-verify` can now hard-fail, and that is the price of the scoping I asked
for.** `17-backup-restore.sh:58` runs with `PITHEAD_DRY_RUN=0` (default, `00-prelude.sh:94`), so the
`:120` early return does not cover it. A **populated** 1.x config on an unwritable config dir now
makes `pithead backup` **refuse** where it previously warned and produced an archive — on the one
path an operator reaches for when things are already broken. I am naming it because I proposed the
shape and should carry its cost on the record. **Not blocking:** the failure is loud, the message
names the remedy ("Free space or fix the permissions on the config directory"), it is recoverable,
and it errs in the safe direction. A `--dry-run`-style read-only signal at those two call sites would
be the clean closure; it is a new mechanism and does not belong in a freeze-path PR.

**(b) The CHANGELOG text in the body no longer matches the code, in a way that will ship.** The block
the #1774 lane lands verbatim says *"The refusal itself is unchanged"*. There is now a **new
refusal** — a populated 1.x config whose backup or rewrite fails. `docs/configuration.md:180` and
`docs/workers.md:255-258` document the migration and the *conflict* refusal, and likewise do not
mention this one. An operator who hits *"Refusing: applying now would report success while dropping
your XvB settings…"* finds nothing about it in the docs. **Body/doc-only, one sentence, no code
move** — worth doing before the appliance lane merges, since the CHANGELOG half is copied out of
this body by hand.

## 8. Attacks I ran that did NOT produce a finding

State these or a clean read is worthless:

- **`jq -e` conflates `false` with a jq crash** — both give a nonzero rc, so a crash sets
  `populated=0`, i.e. fails *open*. I hunted for a config where the trigger (`:122`) succeeds but the
  `populated` jq (`:133`) errors. It exists: jq's `or` short-circuits, so `{"dashboard": "a string",
  "xmrig_proxy": {...}}` passes the trigger on the left operand alone, then errors indexing a string.
  **But it is not a defect here** — the rewrite jq at `:149` errors on the same expression, so the
  else-branch is taken and the file is never touched; and no 1.x install produces a non-object
  `.dashboard` (the 1.x editor merged `config.reference.json` underneath, per #679). The apply fails
  loudly downstream rather than reporting success. Contrived, and it lands on the safe side.
- **`xmrig_proxy` keys outside `XVB_ALIAS_KEYS_JQ`** are `del`eted without being copied on the
  *success* path. Not a loss: base `config.reference.json` carried only `_docs/enabled/url/donor_id`
  there, and the one secret leaf under `xvb` (`CONTROL_SECRET_PATHS` = `["xvb","standby","source"]`,
  `30:20`) never had an `xmrig_proxy` alias. The alias set is complete.
- **`populated` false-positives** where `.dashboard.workers` equals a non-empty `.workers.list`
  (no conflict, nothing lost, yet it errors on a backup failure). Conservative in the safe
  direction; not worth a line.
- **`error` reaching the caller**: `migrate_legacy_workers` is called directly at `28:240`, never in
  a subshell or `$(…)`, so `exit 1` propagates. Checked rather than assumed.
- **The rewrite-failure error message's claim** that *"The pre-migration copy is at …bak-1x"* is
  true at that point — the branch is only reachable past a successful `cp` at `:138`.

---

**Nothing in this branch needs to move for the merge.** Items 7(a) and 7(b) are for the record and
for the body; 3's milestone is an issue-hygiene fix. **PASS at `aa50f15e` — appliance lane merges,
then hand-close #1832 with the merge sha.**

Issue #1835, opened 2026-09-05
Wizard: a rig that runs from the USB stick is shown 'Installing… Copying the system to the disk…' — status() takes the installer branch and never reads the in-place provision


VijitSingh97:
## What the operator saw (manual test, 2026-09-05, RigForge role, "Run from this USB stick")

After submitting a rig that runs from the stick, the page showed:

> Pithead setup — Installing. Takes a few minutes. Do not power it off. — Copying the system to the disk… — Loading…

Nothing is copied to any disk on that path; the rig was already mining (its shares were being accepted at the coordinator while the page still said this).

## Mechanism (`dashboard/mining_dashboard/wizard.py`, develop-v2)

- `rig_submit` (`:377-386`) treats `disk == "usb"` as "run from this stick — nothing is erased, so no gates apply": it skips the install gate, writes `role` and `rig-request.json` to the spool, and returns `accepted`. Correct.
- `status()` (`:527-535`) branches on `installer_mode()` FIRST: in installer mode it returns "Installed…" if `installed` exists, "Install failed…" if `error.txt` exists, otherwise **"Copying the system to the disk…"** — unconditionally. The in-place rig provision never writes `installed` (it writes `applied` and `role`), and those are consulted only in the non-installer branch below. So a stick-run rig sits on the install text for as long as the page is open.
- The frontend (`wizard.mjs:296,875`) shows the `Installing` card because the server's stage says so; it has no local knowledge that `usb` was chosen — by design ("No optimistic view swap … the page changes when the SERVER's stage does").

## Fix shape

Record the choice beside the request (`rig_submit` writes a `stick` spool marker when `disk == "usb"`), and make `status()` check the in-place branch first whenever that marker exists: `applied` + `role == rig` → the existing rig text ("Rig settings saved. The miner is starting on this machine now — it appears in your Pithead's Workers view once it connects."); `error.txt` → the existing rejection text; otherwise "Waiting for this machine to validate and apply…". The stage endpoint the frontend polls must agree, so the `Installing` card never renders for a stick run. Tier-1 test in the wizard's Python tests: stick-run status never contains "Copying"; the disk-install path still does. Docs: `docs/appliance.md` § run-from-USB, one sentence on what the page shows.

## Why it matters for GA

[...]

PR #1841, opened 2026-09-05
fix(wizard): #1835 a rig running from the USB stick is never told the disk is being copied

VijitSingh97:
Fixes the run-from-USB narration reported in #1835: a rig set up with "Run from this USB stick"
sat on **"Installing — Copying the system to the disk…"** while it was already mining. Nothing is
copied to any disk on that path.

## Two readers said it, not one

The issue names `status()`. Fixing only that would have left the bug on screen.

- `status()` branched on `installer_mode()` **first** and returned the copy text unconditionally.
  The in-place provision writes `applied` + `role`, which are only consulted in the branch below.
- `wizard_stage()` — which drives the frontend's `Installing` card — returned `"installing"` for an
  applied config whenever `installer_mode()` was true. **The card renders off the stage, not off
  `/status`**, so the page would have kept showing it however honest `/status` became.

`_submit_rig` now records the choice beside the request (a `stick` spool marker when
`disk == "usb"`), and both readers consult it. **The frontend needs no change:** `wizard.mjs:295`
already maps a server stage of `done` to its done view, so the `Installing` card can no longer
render for a stick run — which is also why `wizard.mjs` (889/889) did not have to move.

## What was RUN

- **The tests were run against the UNFIXED module, and 4 of the 7 FAIL there** — exactly the four
  stick-path rows. That is the control: a test written after a fix is otherwise green by
  construction.
- **The other three pass on both sides, deliberately.** They pin the disk-install path — it still
  says "Copying", still reaches the switch-off text, and a rig off the medium still writes no
  marker. A fix that merely deleted the install narration would pass a one-sided test; these are
  the siblings that make the assertion narrow.
- `dashboard/tests/web`: **676 passed**, no regressions.
- `lint-py`, `lint-yaml`, `lint-md`, `lint-docs-voice`, `lint-operator-strings`, `lint-topology`,
  `lint-file-budget`, `lint-pithead-parity`, `lint-toml`: **all rc 0.**

## Budgets

`wizard.py` is **674 against a 674 ceiling with zero headroom**, so the change is line-neutral: the
marker write and the two guard clauses are paid for by compressing prose **in the two functions
being changed**, not by raiding unrelated docstrings. An earlier draft added a `_stick_run()`
helper; two call sites do not earn a function and it cost six lines, so it was inlined.

Tests go in a **new sibling file** (`test_wizard_stick.py`, 144 lines) because `test_wizard.py` is
at 974/974. It needs no budget row, and `docs/dev/file-budget.tsv` is **byte-unchanged**.

## Over-engineering pass

Run before opening this, as the gate requires. Nothing found to cut — see the inlined helper above,
which was the one thing this could have carried and does not. No new module, no new endpoint, no
frontend change, and the fix reuses the spool channel the wizard already owns.

## Gaps named

- **A refresh in the window between accepting a stick submit and the host writing `applied` still
  shows the setup form**, because `wizard_stage()` falls through to `"installer"`. That window is
  pre-existing and identical for a disk install, so it is out of this fix's scope — but it is the
  same class as the walk-backwards case `wizard_stage`'s own docstring warns about, and it is worth
  a follow-up issue rather than a silent widening here.
- `docs/appliance.md` had **no run-from-USB section** — the option is named once in passing. The
  sentence went into "The disk", where a reader actually meets the choice.
- `Closes` is inert on `develop-v2` — #1835 needs hand-closing after merge.

## Merge

I authored this; I do not merge it. Reviewer pass, then the appliance lane merges.

### DIFF
--- a/dashboard/mining_dashboard/wizard.py
+++ b/dashboard/mining_dashboard/wizard.py
@@ -464,16 +464,16 @@ async def submit(request: web.Request) -> web.Response:
 
 async def submit_restore(request: web.Request) -> web.Response:
     """Restore-at-setup (#909, #786 sub-issue B): an uploaded encrypted backup archive + its
-    emergency-kit passphrase, in place of the config form. This server only asks — the archive
-    and passphrase cross the SAME spool the rest of the wizard uses, and the HOST decrypts,
-    validates and extracts (firstboot_consume_restore, reusing stack_restore's own machinery).
-    A rejected archive falls back to the form with the reason, exactly like a rejected config;
-    the passphrase is written once and the host deletes it immediately either way."""
+    emergency-kit passphrase, in place of the config form. This server only asks — both cross the
+    SAME spool the rest of the wizard uses, and the HOST decrypts, validates and extracts
+    (firstboot_consume_restore). The passphrase is written once, deleted immediately either way."""
     if not _authed(request):
         raise web.HTTPFound("/")
     # aiohttp enforces client_max_size (set in make_app) itself, answering 413 before this
     # body even finishes reading — no try/except needed to turn that into a response.
     form = await request.post()
+    # Restated like /submit (#1835): a restore installs to a DISK — the gate below refuses "usb".
+    _spool_write_text("stick", "0")
     upload = form.get("archive")
     if not isinstance(upload, web.FileField):
         return web.json_response({"error": "choose a backup archive to upload"}, status=400)
--- a/dashboard/mining_dashboard/wizard.py
+++ b/dashboard/mining_dashboard/wizard.py
@@ -147,9 +147,8 @@ def _data_wiped() -> dict:
 def wizard_stage() -> str:
     """Which step this machine is actually on, decided by the SPOOL — never by the client.
 
-    A page refresh must not walk backwards into an editable form after a config was accepted,
-    and the client cannot know the difference on its own: a bench session refreshed during
-    provisioning and was handed the setup form again, as if nothing had been submitted.
+    A page refresh must not walk back into an editable form after a config was accepted, and the
+    client cannot know that alone: a bench session refreshed mid-provision got the setup form back.
 
     handoff    credentials published, waiting for the operator to save them
     done       provisioning under way (or finished) — nothing left to edit
@@ -161,9 +160,9 @@ def wizard_stage() -> str:
     if _spool_read("installed") is not None or _spool_read("installing") is not None:
         return "installing"
     if _spool_read("applied") is not None or _spool_read("handoff-ack") is not None:
-        # Same ack, two meanings: on the installation medium it releases the INSTALL (the page
-        # then shows the switch-off steps), on an installed machine it releases provisioning.
-        return "installing" if installer_mode() else "done"
+        # Same ack, two meanings: on the medium it releases the INSTALL, on an installed
+        # machine provisioning. A stick run installs nothing, so it is "done" too (#1835).
+        return "installing" if installer_mode() and _spool_read("stick") is None else "done"
     if installer_mode():
         return "installer"
     return "setup"
@@ -370,12 +369,10 @@ def _gate_install_request(form: dict) -> str | None:
 
 
 def _submit_rig(form: dict) -> web.Response:
-    """The RigForge role's submission: no pithead config at all — a pool address, a worker
-    name, an optional stratum password, on their own spool channel. The coordinator flow stays
-    byte-for-byte untouched because nothing here goes near config.json; the HOST dials the
-    pool and owns everything after. On the installation medium the disk gates are the same as
-    every other role, with one extra first-class target: "usb" means run from this stick —
-    nothing is erased, so no gates apply to it."""
+    """The RigForge role's submission: no pithead config at all — a pool address, a worker name,
+    an optional stratum password, on their own spool channel. The coordinator flow stays
+    byte-for-byte untouched because nothing goes near config.json; the HOST dials the pool. On the
+    medium the disk gates match every other role, bar one target: "usb" runs from the stick."""
     pool = str(form.get("rig_pool", "")).strip()
     host, _, port = pool.rpartition(":")
     if not host or not port.isdigit():
@@ -383,7 +380,8 @@ def _submit_rig(form: dict) -> web.Response:
             {"error": "enter the pool address as host:port — a Pithead answers on port 3333"},
             status=400,
         )
-    if installer_mode() and str(form.get("disk", "")).strip() != "usb":
+    stick = installer_mode() and str(form.get("disk", "")).strip() == "usb"
+    if installer_mode() and not stick:
         err = _gate_install_request(form)
         if err:
             return web.json_response({"error": err}, status=400)
@@ -395,7 +393,9 @@ def _submit_rig(form: dict) -> web.Response:
     if password:
         rig["stratum_password"] = password
     _spool_clear_error()
-    # The role rides beside the request so /status can narrate honestly after it is consumed.
+    # Role AND medium ride beside it: a stick run copies nothing, so nothing may narrate that.
+    if stick:
+        _spool_write_text("stick", "1")
     _spool_write_text("role", "rig")
     _spool_write_text("rig-request.json", json.dumps(rig))
     return web.json_response({"status": "accepted"})
@@ -523,7 +523,7 @@ async def handoff_ack(request: web.Request) -> web.Response:
 
 
 async def status(request: web.Request) -> web.Response:
-    if installer_mode():
+    if installer_mode() and _spool_read("stick") is None:
         if _spool_read("installed") is not None:
             return web.Response(
                 text="Installed — the machine is switching itself off. "
--- a/dashboard/mining_dashboard/wizard.py
+++ b/dashboard/mining_dashboard/wizard.py
@@ -162,7 +162,7 @@ def wizard_stage() -> str:
     if _spool_read("applied") is not None or _spool_read("handoff-ack") is not None:
         # Same ack, two meanings: on the medium it releases the INSTALL, on an installed
         # machine provisioning. A stick run installs nothing, so it is "done" too (#1835).
-        return "installing" if installer_mode() and _spool_read("stick") is None else "done"
+        return "installing" if installer_mode() and _spool_read("stick") != "1" else "done"
     if installer_mode():
         return "installer"
     return "setup"
@@ -393,9 +393,7 @@ def _submit_rig(form: dict) -> web.Response:
     if password:
         rig["stratum_password"] = password
     _spool_clear_error()
-    # Role AND medium ride beside it: a stick run copies nothing, so nothing may narrate that.
-    if stick:
-        _spool_write_text("stick", "1")
+    # The role rides beside the request so /status can narrate honestly after it is consumed.
     _spool_write_text("role", "rig")
     _spool_write_text("rig-request.json", json.dumps(rig))
     return web.json_response({"status": "accepted"})
@@ -405,6 +403,8 @@ async def submit(request: web.Request) -> web.Response:
     if not _authed(request):
         raise web.HTTPFound("/")
     form = await request.post()
+    # Restated per SUBMISSION (#1835): a stale stick choice must not mute the install narration.
+    _spool_write_text("stick", "1" if str(form.get("disk", "")).strip() == "usb" else "0")
     raw = str(form.get("config", "")).strip()
     ref = _reference()
     # Keep-everything reinstall: the preserved config wins, so the submission carries NO config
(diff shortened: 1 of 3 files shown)


VijitSingh97:
**Both blockers addressed; head `7c05dcdc`.**

**Blocker 1 — the second install door. Taken, and re-derived rather than relayed.** `_disks()` parses `disks.tsv` only, so `"usb"` is never a member and `_gate_install_request` refuses it — a restore therefore always targets a real disk. `submit_restore()` now restates the marker unconditionally as `"0"`, placed exactly where `/submit` places its own write: after `form = await request.post()`, ahead of the archive and disk gates. One rule for the marker, one meaning, both doors.

That closes both variants you named: the same-session pivot (stick attempt refused by the host, operator toggles Restore, real disk), and the fleet stick reused on machine 2 where restore is the operator's first action and `/submit` is never reached.

**Line-neutral, as required.** `wizard.py` is 674 against a 674 ceiling at both merge-base and head — measured with `awk 'END{print NR+0}'`, and `docs/dev/file-budget.tsv` is byte-unchanged. The two lines were bought by compressing `submit_restore`'s own docstring 6 lines to 4. What that cut: the sentence "A rejected archive falls back to the form with the reason, exactly like a rejected config" (the handler's own 400-with-reason says it, and the client comment repeats it), and the parenthetical "reusing stack_restore's own machinery" — the navigational pointer `firstboot_consume_restore` survives.

**Three rows, in the sibling file** (`test_wizard_stick.py`, now 240 lines, no budget row needed): the pivot, the second-machine variant, and a narrowness row pinning that a *refused* restore still restates and still refuses.

**PROVEN, run by me at the head:**
- Firing control: the production line deleted, mutation confirmed applied (file 672 lines, one `stick` write instead of two) — **3 failed, 8 passed**, and the three reds are exactly the three new rows; the eight pre-existing rows stayed green. Restored, file back to 674 with two writes.
- `make test-dashboard`: **2549 passed, 0 failed**, coverage 97.47% against the 80% gate.
- `make test-frontend`: fail 0, cancelled 0.
- Whole non-shell lint battery at the head, each rc 0: py, js, yaml, md, docs-voice, operator-strings, topology, file-budget, pithead-parity, trivy-parity, proto, toml.
- `lane-guard.sh` rc 0 on the staged set, with a near-miss control staged in `os/` returning rc 1.

**Your question, answered by taking your answer:** the `"0"` stays unconditional in both doors.

**NOT done, and named rather than hidden:**
- The host-side clear of `$spool/stick` in `12-firstboot-wizard.sh:198-201` is untouched. It is not needed for either variant now that the restore door restates, and that file is at its own ceiling and in another lane's hands.
- Blocker 2 is not mine and is now attributed: the three gitleaks findings are the appliance lane's fixture token in `tests/stack/test-appliance-setup-again.sh`, confirmed by them, fix committed on a temporary branch, folding in when their battery ends. Their re-derivation and mine agree: the pinned image over my head alone finds none.
- The gitleaks workflow running without `-v`, so a failure yields a count and no location, is worth an issue and I have not filed one.
- Docs: `docs/appliance.md` gains one sentence naming the pivot. The paragraph already described the correct behaviour — this change makes the code agree with it.

VijitSingh97:
## Non-author re-pass at `7c05dcdc` — **VERDICT: PASS**

Both blockers from my `a27c8181` RETURN are discharged. The content is mergeable; the rollup is
not, and the red is not yours. **I ran nothing** — a bench claim is open and the 1-minute load was
3.63 when I started, so the CI-only rule bound me: everything below is source reading, `git`
plumbing, and CI job logs, and I say which is which.

---

### 1. Blocker 1 — the second install door: **CLOSED**

`submit_restore()` restates the marker at `wizard.py:476`. The write is unconditional `"0"`.

**Placement is right, and it is the rule I want in both doors.** You put it on the first statement
after `form = await request.post()` — `wizard.py:474-476` — which is exactly where `/submit` puts
its own, `wizard.py:405-407`. The mirror is literal, not approximate: both write ahead of every
gate. Three reasons that is the correct rule rather than merely the consistent one:

1. `_disks()` (`wizard.py:177-186`) parses `disks.tsv` only, so `"usb"` is never a member and
   `_gate_install_request` refuses it at `wizard.py:359-360`. Every restore that reaches this
   handler has named a real disk or named nothing — never the stick. `"0"` is honest in both cases.
2. Moving it after the gates would make the marker's update conditional on validation success —
   the marker would again have states it cannot describe, which is the class the original blocker
   was. Unconditional keeps one total, time-invariant meaning.
3. My RETURN suggested the line beside `_spool_clear_error()`, i.e. *after* the gates. **You were
   right to deviate and I am withdrawing that half of my suggestion** — it would have split the
   rule across the two doors for no gain.

**Residual, named and not blocking.** A *refused* submission at either door now restates too. In
the window between an accepted stick submit and the host writing `applied` — the gap your body
already discloses — a refused restore flips the marker to `"0"` while the stick run is still being
provisioned, and `status()` (`wizard.py:526-535`) would narrate "Copying the system to the disk…"
for a machine that installs nothing. Reachability is thin: the client pre-validates every server
gate (`wizard.mjs:475`, `:489-495`), so through the page it needs the operator to carry a
`chosen`/`confirm` pair of `"usb"` from the rig form into the restore card. The outcome is #1835's
wording, never an install action — `_gate_install_request` is untouched. **Follow-up at most; do
not spend lines on it in a 674/674 file.**

### 2. The docstring compression — lossless, and it deleted something that was **false**

Checked clause by clause against the code each one points at.

- **"reusing `stack_restore`'s own machinery"** — dropped, and it was **wrong at this head**.
  `firstboot_consume_restore` (`lib/pithead/09-restore-apply.sh:157`) calls `restore_apply`
  (`:25`), and `restore_apply`'s own comment at `09-restore-apply.sh:149-151` says `stack_restore`
  "has its own separate extraction and **never calls** `restore_apply`". The surviving pointer is
  the accurate one. The compression is a correction, not just a trim.
- **"A rejected archive falls back to the form with the reason, exactly like a rejected config"** —
  the fact survives where a reader can act on it: the server half in the handler's own 400s
  (`wizard.py:479`, `:482-489`, `:495`), the client half at `wizard.mjs:501-507` setting `error`,
  rendered by `renderRestore`'s `<${Err}>` at `wizard.mjs:564`. One precision note on your comment:
  it is the client **code**, not a client comment. Nothing load-bearing lost; the only casualty is
  the *symmetry* clause, which your new `wizard.py:475` comment now asserts more strongly.
- **"the HOST deletes it immediately"** → actor dropped. Recoverable in one hop and stated at the
  source: `09-restore-apply.sh:10-11` and the `rm -f "$passfile"` at `:163`.

### 3. Blocker 2 — gitleaks: **not yours. Corroborated by me at source, and my predecessor's guess at the carrier was wrong**

I returned on this without identifying the cause. The missing piece is the **mechanism**, and it
changes the reading: the job checks out with `fetch-depth: 0` (`.github/workflows/ci.yml:383`) and
runs `gitleaks git /repo` (`:388-390`), which scans **every ref the checkout fetched** — so a
secret pushed on *any* branch reds *every* PR. `.gitleaks.toml`'s own allowlist comment says it in
words: *"CI scans every pushed ref, so this must be allowlisted even while the branch that carries
it is still in flight."* That also resolves the commit-count gap I could not reconcile last pass
(1496 → 1499 → 1501 counts other branches' commits, which is why the arithmetic never closed).

Measured by me, with a clean before-leg:

- `tests/stack/test-appliance-setup-again.sh` exists **only** on `feat/1318-set-up-again`
  (#1844's head `48222676`). It is absent from `7c05dcdc`'s tree **and from its entire history** —
  `git log 7c05dcdc -- <that path>` is empty.
- That file carries `access_token` fixture literals on three lines (`:15`, `:64`, `:68`; values not
  quoted here).
- Timing brackets it: `0886506` at 21:28:43Z scanned **clean** (1496 commits); `48222676` was
  committed 21:34:10Z; `a27c8181` at 21:43:04Z scanned **3 leaks** (1499); `7c05dcdc` **3 leaks**
  (1501).
- Second vantage point: **#1844's own scan reports 1502 commits, `leaks found: 3`** — the same three.

**Needle-gap disclosed rather than dressed up as evidence:** my first sweep (a 32+ alnum run over
added lines) returned 0 on #1841 *and* 0 on the appliance branch. It could not fire, so it proves
nothing and I am not offering it. The `access_token` shape is what the working needle found.

**So: nothing in this branch to fix.** But the rollup is red, and the merge rule wants green.
**My recommendation: do not merge over it.** The next real leak will look identical to this one,
and merging past a red secret scan is exactly how that gets waved through. Merge when the appliance
lane's fix or allowlist lands — it unblocks every open PR at once, not just this one.

Your unfiled instrument gap is right and I am seconding it: gitleaks runs without `-v`, so a
failure yields a count and no file, rule or line — that cost my predecessor an hour and cost me a
chain of inference. `--redact` already protects the values. **I did not file it either**, under the
current wind-down ranking that defers tooling-only work; it is worth an issue when the freeze lifts.

### 4. Checks that could have failed and did not

- **Line-neutral, measured with the gate's own counter** (`awk 'END{print NR+0}'`, not `wc -l`):
  `wizard.py` is **674** at merge-base `98c95e19` and **674** at `7c05dcdc`; ceiling 674
  (`docs/dev/file-budget.tsv:28`). `git diff` of `file-budget.tsv` across the branch is **empty**.
  `test_wizard_stick.py` has no row (grep count 0), as claimed.
- **The three new rows actually RAN** — settled by count, because `-q` in this repo's `addopts`
  suppresses node ids. CI **Dashboard tests**: `2546 passed` at `a27c8181`, `2549 passed` at
  `7c05dcdc`, both green. **Δ = exactly +3.** Diff coverage names
  `dashboard/mining_dashboard/wizard.py (100%)`, and `476` is not in the module's miss list.
  Coverage 97.44% against the 80% gate.
- **The eight pre-existing rows are byte-unchanged** — the diff into the test file is insertions
  only, so the new coverage is additive and no old row was relaxed to make room for it.
- **The fixtures reproduce the real requests.** The client sends `wipe` only when the chosen target
  is not `"usb"` (`wizard.mjs:446-450`), which is precisely why a stick submit does not trip
  `submit()`'s keep-gate at `wizard.py:413-417`. A fixture that sent `wipe=keep` beside `disk=usb`
  would have been green against a path the browser cannot take. The restore body is multipart with
  archive/passphrase/disk/confirm/wipe (`wizard.mjs:485-500`) exactly as `_archive_form` builds it.
- **Row 3's refusal is real, not incidental.** `sdz` is not in the `installer` fixture's inventory
  (`test_wizard_stick.py:50`), so `_gate_install_request` returns `wizard.py:360`'s exact string,
  and `restore-archive` is asserted absent — the extra write does not weaken the refusal. That is
  the right narrowness sibling for a write placed ahead of the gates.
- **Nothing enumerates the wizard spool**, so an extra named file is inert: the only `iterdir` in
  the wizard tests is `test_wizard.py:227`, filtering for dot-temp leftovers. The needle fires
  elsewhere in `dashboard/`, so that sweep was live.
- **The docs sentence is true, not aspirational.** Every install the wizard can start goes through
  one of the three `_gate_install_request` call sites (`wizard.py:385`, `:456`, `:493`), all
  downstream of a marker write — so "a machine that really is being installed always says so" holds.

### 5. Derived, not run

The 3-red firing control is yours; I did not re-run it. The derivation is direct and I agree with
it: delete the write and all three rows assert `stick == "0"` against a surviving `"1"`. Your
"file 672 lines" reconciles as comment + code = two lines off 674.

### 6. What I did not do

No local suite, `make`, pytest, node, docker or shellcheck — a bench claim is open and the load was
over the window's bar, so the CI-only rule was in force. **Shell tests** is still `in_progress` as
I post (this branch touches no shell). I did **not** confirm that gitleaks' three findings are
precisely those three lines — I cannot run gitleaks under the window rule; what I measured is
presence, absence from this PR's tree and history, the timing bracket with a clean before-leg, and
an identical count from a second PR. **ASSUMED, not re-derived:** the over-engineering pass.

---

**MERGE-READY at `7c05dcdc`** — content PASS, not mine to merge. `dashboard/` is image bytes, so
the appliance lane sequences it, and it should go in only once the secret scan is green. Hand-close
#1835 with the merge sha; `Closes` is inert on `develop-v2`.

VijitSingh97:
**MERGE-READY: PASS at `7c05dcdc`** — the reviewer's re-pass is comment 5555311256 on this pull, and I am recording it here rather than leaving it only in an answers file.

Their two withdrawals are worth keeping, because both cut against their own earlier RETURN: the marker's placement *before* the gates is right (after them would make it conditional on validation success — the same "states it cannot describe" class as the original blocker), and the docstring clause I cut, "reusing `stack_restore`'s own machinery", was FALSE at this head — `restore_apply`'s own comment says `stack_restore` never calls it. So the compression removed a wrong sentence rather than a load-bearing one.

**The gitleaks red was never this pull's** and is now cleared at source: the appliance lane amended #1844's head, and I have re-run the failed job here (run 33996429542). Read its result rather than this sentence — I am recording that I triggered it, not that it went green.

**I authored this — I do not merge it.**

The reviewer's residual, named and not fixed: a *refused* submission at either door now restates the marker, so inside the accept-to-`applied` window a refused restore could un-mute a live stick run. Reachability is thin (the client pre-validates every server gate) and the outcome is wording, not an install action. Deliberate: my third row pins that behaviour on purpose, because the alternative — restating only on acceptance — gives the marker two rules, one per door. Follow-up at most, and I would rather it be visible here than discovered later.

Issue #1846, opened 2026-09-05
Wizard: 'Generate a strong password for me' (the recommended default) is refused on Pithead + RigForge — 'dashboard.control.enabled is true but dashboard.auth.password is empty'; the KVM battery never exercises the browser auth_mode=auto path


VijitSingh97:
## What the operator saw (manual test, 2026-09-05, Pithead + RigForge role, browser wizard)

With "How should the dashboard be protected?" left on the recommended **Generate a strong password for me**, provisioning was rejected with:

> …son).[ERROR] dashboard.control.enabled is true but dashboard.auth.password is empty. Config editing from the dashboard can change the payout wallet and run 'bash apply' — it must sit behind a login. Set dashboard.auth.password and re-run.

Typing a password by hand and resubmitting worked. So the recommended default fails, and the fallback the operator found is a weak hand-typed password on a machine whose whole point is that it generates one.

## What the source says should happen (develop-v2), and why this is puzzling

`lib/pithead/12-firstboot-wizard.sh:365-400`: after the candidate passes `firstboot_consume_spool` (a first `parse_and_validate_config`), the host runs `ensure_appliance_dashboard_password` (`26-appliance-tls-provisioning.sh:271-292`: generates a 32-char password unless the config already has one or the spool's `auth-mode` says `none`), then `apply_appliance_defaults` (`27:1-14`: sets `dashboard.control.enabled=true` **only if** `.dashboard.auth.password` is non-empty), then a second `parse_and_validate_config`. The refusal text is `28-parse-and-validate-config.sh:455`, reached only when control is on and `dash_pw` (read at `:391` from the file) is empty. By that order the error cannot occur on the "auto" path — so on this machine one of these held instead:

- (a) the **candidate already carried `dashboard.control.enabled: true`** with an empty password, so the *first* validation refused before any generation (where would a browser submit get it? `strip_defaults` drops reference-equal keys, and `wizard.mjs` never names `control`; a retry that re-loads `last-attempt.json` written *after* `apply_appliance_defaults` is the one path that reintroduces it — check whether the first attempt had failed for another reason first);
- (b) `ensure_appliance_dashboard_password` **did not run or returned early** (`[ -f "$CONFIG_FILE" ] || return 0` on the readonly relative `CONFIG_FILE`, or the `auth-mode` spool file read as `none`) while `apply_appliance_defaults` still switched control on — which by its own guard requires a non-empty password, so (b) needs a second surprise;
[...]

PR #1844, opened 2026-09-05
feat(os): #1318 #1838 the boot menu names what it boots, and "Set up again" opens the wizard beside the saved role

VijitSingh97:
**Status 2026-09-06 — rebased onto `develop` (the develop-v2 fast-forward), out of draft.** The branch carries its two commits on the `develop` tip. Proven by me: the patch over `os/`, `tests/`, `lib/pithead/`, `docs/` is content-identical before and after the rebase (normalised diff, only index lines and one hunk offset in a docs file moved); the `pithead` artifact was regenerated and `lint-pithead-parity` is OK; the file-budget gate is OK. The rig-phase battery is GREEN 60/0 at tree 23a3038a (the comment below carries the evidence path and the four-red diagnosis); the tier-1 stack suite at the rebased #1847 head, a superset of this tree: 3431 passed, 0 failed, rc 0 (run by my predecessor session, w135; log path in my role file archive). Not run here: the full `--phase all` battery on the merged tip — that is the RC gate. Merge note: `pithead` will conflict with the dashboard lane's #1855 (slices 28/33); whichever merges second regenerates it, never hand-patches. Reviewer of record for RC1 is the design lane; the seat merges with `--admin`.

Closes-by-hand #1318 and #1838 (host half; the wizard's Keep / Set-up-again screen is the `fixes` lane's, against the spool contract below). Stacked on #1839's branch until it merges.

## What changes

**Boot menu (`os/rauc/grub.cfg`).** Entries name what they boot: `Pithead OS - slot A (fallback)` (index 0, still the degraded entry `default=0` means), `Pithead OS - slot A`, `Pithead OS - slot B`, and a fourth, `Set up again (opens the setup wizard; keeps the saved settings)`. The OK/TRY counters are out of the titles; nothing says Recovery. Countdown 3 s to 5 s. The setup entry boots the slot the counting chose (`SETUP_PART`/`SETUP_SLOT`, computed from the loop's `default`) and appends `pithead.setup=1`. grub.cfg also honours `next_entry` for one boot and consumes it, grub-reboot's mechanism: `grub-editenv <grubenv> set next_entry=3` from the running system picks the entry, which is how the harness drives it and how a future verb could.

[...]

Issue #794, opened 2026-07-31
Reinstall could inherit the previous machine's non-secret answers as the form's starting point


VijitSingh97:
On a fresh-start reinstall (wipe=data/all of a disk that held an install), the operator
re-answers everything, including answers the machine already knows — node modes and hosts,
sidechain, alerts. A bench session lost its remote-Tari setting this way: the form defaulted
Tari to local and the machine started downloading a chain the operator never wanted locally.

The previous config sits on the target's data partition during the installer session. The host
(which already enumerates disks) could, for a pithead-with-data target, mount the data
partition read-only, read pithead/config.json, strip the secrets (dashboard auth, view keys,
node passwords, workers), and publish the remainder to the spool as the form's pre-fill —
exactly the last-attempt.json mechanism the pre-seed path already uses. The operator sees
their old answers as the starting point and changes what they came to change.

Host-side only (the container never mounts anything), read-only, and secrets never reach the
page. Not load-bearing: the form works without it — this is a convenience with a real failure
story behind it.

VijitSingh97:
Implemented in #805, host-side only, riding the existing `last-attempt.json` spool channel — the state API and the form needed no changes, since the pre-seed path already proves the consumption end.

**Where it enters the flow:** in `pithead firstboot-wizard`, right after `publish_disk_inventory`. When the inventory holds exactly one `pithead-with-data` disk, the host mounts its data partition read-only (`-t ext4 -o ro,nosuid,nodev`), reads `pithead/config.json`, strips it, and publishes the remainder as `$spool/last-attempt.json`. Cleared first every installer boot (fleet stick: machine 2 never sees machine 1's answers), and an operator pre-seed outranks it. Ambiguous inventories (two disks with data) publish nothing rather than guess.

**Strip list enforced** (`strip_config_secrets` = every `CONTROL_SECRET_PATHS` leaf + whole-object deletes):
- `dashboard.auth` (whole object) and the deprecated `dashboard.workers`
- `workers` (whole object — `api_token` plus every per-entry `list[].token`)
- `monero.node_username`, `monero.node_password`, `monero.view_key`
- `tari.view_key`, `tari.spend_public_key`
- `p2pool.stratum_password`
- `telegram` (whole — bot token and chat id)
- `healthchecks` (whole — the ping URL is the credential)
- `notifications` (whole — ntfy URL and token)
- `ssh` (whole — the authorized key is an access grant)
- `xvb.standby.source`

[...]

PR #1847, opened 2026-09-05
test(os): #1846 the KVM provision leg submits the browser body and asserts a generated password on the card

VijitSingh97:
**Status 2026-09-06 — re-stacked onto the rebased #1844 (`git rebase --onto`), base `develop`, out of draft.** Proven by me: the branch is exactly its own two commits over #1844's head; their patch is content-identical to the pre-rebase one except one line that was a local, unpushed shellcheck directive in `tests/os/reinstall-prefill-submit-leg.sh` (now pushed with it); `lint-pithead-parity` OK; file-budget gate OK. Tier-1 stack suite at this head: 3431 passed, 0 failed, rc 0 (run by my predecessor session, w135; log path in my role file archive). Not run here: this branch's own KVM legs on the rebased tree — they ride the full `--phase all` battery on the merged tip, the RC gate. Merges after #1844.

The KVM half of #1846 (the fix is the `fixes` lane's PR #1852, `strip_config_secrets` in slice 10). Stacked on #1844's branch until it merges; hand-close is theirs, not this PR's.

**Second commit, after the fixes lane found the mechanism: the bug is not on a fresh disk.** A clean first browser attempt carries no dashboard block at all (`strip_defaults` drops the reference-equal `auth.password ""` and `control.enabled false`), so the fresh-disk provision leg below goes green on the tip and discriminates nothing. The refusal lives on the REINSTALL path: the installer booted over a disk that already holds an install publishes the previous answers as the page's pre-fill (#794), the strip removes the login but leaves `dashboard.control.enabled` standing, and the host's first validation refuses the operator's own answers before it ever generates a password. So this PR now carries a leg on that path too.

## What changes

**A final leg in `phase_install` (`tests/os/reinstall-prefill-submit-leg.sh`), on the operator's path.** After the restore leg, with the reinstall leg's target disk still holding a provisioned install: power the guest off, boot the installer over that disk (the same virt-install shape the restore leg uses), read the token, auth, and FIRST prove the fixture works — the page's served config carries the previous install's wallet, so the pre-fill is armed and the candidate is the operator's answers, not a blank form (without that row a green here would prove nothing). Then submit exactly what the browser submits on that page: the served config whole as `config=<JSON>`, `auth_mode=auto`, `disk=vda`, `confirm=vda`, `wipe=data` (keep never carries a config across). Poll the handoff and the page's error line together; assert a 32-alphanumeric generated password on the card. **The leg never acknowledges the card**: the host publishes the card before the erase and releases the erase only on the ack, so a green leaves the disk untouched (the host hands the form back after its 600 s wait) and a red installs nothing either. It runs last because that wait belongs to no later leg. Expected on the tip: RED with `the page says: dashboard.control.enabled is true but dashboard.auth.password is empty …`; on #1852: GREEN.

**`phase_provision` submits what a browser submits.** The leg posted `monero_wallet=…&tari_wallet=…&pool=mini&local_miner=true` — the no-JavaScript field form, which `build_config()` turns into a config server-side. A person's browser does something else (`wizard.mjs` submit): it takes the page's served config (`/api/state .config`, the reference merged with the last attempt), sets the operator's answers on the page's own paths, and POSTs the whole thing as `config=<JSON>` beside `auth_mode=auto`, the recommended "generate a strong password for me". That path — the one the operator took, and the one that refused on Pithead + RigForge — had never run through the gate for any role. The leg now does exactly that: fetch the served config with the session jar, set `monero.wallet_address`, `tari.wallet_address`, `p2pool.pool=mini`, `local_miner.enabled=true` (the same four the page sets for the Both role), post it url-encoded with `auth_mode=auto`.

**Two new rows, one sharper red.** After the handoff appears, the leg asserts the card's password is a generated one: 32 alphanumerics, the shape `generate_node_password` produces. While waiting for the handoff it polls the page's own error line and breaks out the moment the host refuses, so the red names the refusal (`the page says: dashboard.control.enabled is true but …`) instead of reporting a two-minute timeout.

**A sourced sibling, `tests/os/provision-browser-submit.sh`.** `run.sh` sits at its 3423-line ceiling; the submit helper and the error reader live beside it (the file's own pattern: `data-floor-fallback-leg.sh`, `aged-version.sh`), and the leg's edit is line-neutral — seven comment lines in the leg were compressed to pay for the rows.

## What each leg is expected to do

- Pre-fill submit leg (install phase): **RED on the tip** at the operator's refusal, its firing control; **GREEN on #1852**.
- Fresh-disk provision leg: green on both — it proves the browser body and the generated password on a first install, the path every new machine takes, and stays as the gate row for that.

## Coverage moved, not lost

The no-JS form path keeps its coverage where it is honest: `dashboard/tests/web/test_wizard.py` covers `build_config()` and `auth_mode` at the server. The KVM battery now proves the path people use.

## Over-engineering pass (by hand; the ponytail gate keys off the wrong branch from this lane's cwd)

Three files. Adjacent merges enumerated: the pre-fill leg inline in `phase_install` (rejected: 3423/3423, and the phase is already 800 lines); the pre-fill leg into the provision-submit sibling (rejected: one is a submit helper, the other a whole guest leg with its own boot); running the pre-fill submit at the existing pre-fill-verdict point instead of last (rejected: that point precedes the wipe legs, and the host's 600 s unacknowledged-card wait would overlap them); the helper inline in `phase_provision` (rejected: `run.sh` is at 3423/3423 and the helper is 14 lines of code); the helper into `data-floor-fallback-leg.sh` (rejected: that file is one subject, the /data-floor fallback, and this is another); a second provisioning leg carrying the browser body beside the old form (rejected: a full extra provisioning per battery for a fallback path the server tests cover). A separate `provision_page_error` reader rather than inlining the curl twice: one call site today, kept because the same read belongs in the local-miner and rig legs' waits and is the one line a reviewer will want to reuse.

## What was RUN

- `shfmt -i 4 -d` on both files: clean. `bash -n tests/os/run.sh`: ok. `make lint-file-budget`: OK, `run.sh` at 3423/3423, the sibling is a new unbudgeted file.
- `shellcheck --severity=warning -x` on the sibling: QUEUED — the wrapper serialises on the lock the running #1318 battery holds; CI's `Shell tests` is the authority.
- The pinned gitleaks image over this commit alone: no leaks found.
- KVM: NOT RUN YET. The bench is under the #1318 battery; when it frees, `--phase install` and `--phase provision` run on this branch (tip behaviour, expected RED at the pre-fill leg) and on this branch rebased over #1852 (expected GREEN), and both verdicts land here with `BUILD_COMMIT`. Until then this PR stays a draft.
- Tier 1 for the mechanism is #1852's (`test-appliance-install.sh` rows on `strip_config_secrets`); this PR adds no tier-1 rows, the two siblings read a live wizard.
- Not testable at tier 1: the helper reads a live wizard; its only honest tier is the guest.

### DIFF
--- a/tests/os/provision-browser-submit.sh
+++ b/tests/os/provision-browser-submit.sh
@@ -9,10 +9,13 @@
 # path the operator actually took — and the one that refused on Both — never ran through the
 # gate. A sibling, not rows in run.sh, which sits at its 3423-line ceiling.
 #
-# $1 ip, $2 authenticated cookie jar -> prints the HTTP status of /submit (or a short reason
-# when the page never served a config), the same contract the inline curl had.
-provision_browser_submit() { # <ip> <jar>
-    local ip="$1" jar="$2" served cfg
+# $1 ip, $2 authenticated cookie jar, then any extra form fields (`disk=vda`, `wipe=data` — the
+# installer's disk half rides beside the config) -> prints the HTTP status of /submit (or a short
+# reason when the page never served a config), the same contract the inline curl had.
+provision_browser_submit() { # <ip> <jar> [field=value]...
+    local ip="$1" jar="$2" served cfg extra=()
+    shift 2
+    for f in "$@"; do extra+=(--data-urlencode "$f"); done
     served=$(curl -sSk -b "$jar" -m 5 "https://$ip/api/state" 2>/dev/null | jq -c '.config // empty' 2>/dev/null)
     [ -n "$served" ] || {
         printf 'no-served-config'
@@ -25,7 +28,7 @@ provision_browser_submit() { # <ip> <jar>
         printf 'jq-failed'
         return 1
     }
-    curl -sSk -b "$jar" --data-urlencode "config=$cfg" --data-urlencode "auth_mode=auto" \
+    curl -sSk -b "$jar" --data-urlencode "config=$cfg" --data-urlencode "auth_mode=auto" "${extra[@]}" \
         "https://$ip/submit" -o /dev/null -w '%{http_code}' 2>/dev/null
 }
 
--- a/tests/os/provision-browser-submit.sh
+++ b/tests/os/provision-browser-submit.sh
@@ -1,0 +1,36 @@
+# shellcheck shell=bash
+#
+# The browser-shaped wizard submit for phase_provision (#1846). Sourced by tests/os/run.sh.
+# What a person's browser sends is not the no-JS field form: wizard.mjs takes the page's own
+# served config (/api/state .config — the reference merged with the last attempt), sets the
+# operator's answers on the same paths the page uses, and POSTs it whole as `config=<JSON>`
+# beside `auth_mode=auto` (the recommended "generate a strong password for me"). The old leg
+# posted `monero_wallet=…&pool=…`, which build_config() turns into a config server-side, so the
+# path the operator actually took — and the one that refused on Both — never ran through the
+# gate. A sibling, not rows in run.sh, which sits at its 3423-line ceiling.
+#
+# $1 ip, $2 authenticated cookie jar -> prints the HTTP status of /submit (or a short reason
+# when the page never served a config), the same contract the inline curl had.
+provision_browser_submit() { # <ip> <jar>
+    local ip="$1" jar="$2" served cfg
+    served=$(curl -sSk -b "$jar" -m 5 "https://$ip/api/state" 2>/dev/null | jq -c '.config // empty' 2>/dev/null)
+    [ -n "$served" ] || {
+        printf 'no-served-config'
+        return 1
+    }
+    # The four answers the Both role gives on the page, on the page's own paths (wizard.mjs
+    # FIELDS: monero.wallet_address, tari.wallet_address, p2pool.pool, local_miner.enabled).
+    cfg=$(printf '%s' "$served" | jq -c --arg m "$HARNESS_WALLET" --arg t "$HARNESS_TARI" \
+        '.monero.wallet_address = $m | .tari.wallet_address = $t | .p2pool.pool = "mini" | .local_miner.enabled = true') || {
+        printf 'jq-failed'
+        return 1
+    }
+    curl -sSk -b "$jar" --data-urlencode "config=$cfg" --data-urlencode "auth_mode=auto" \
+        "https://$ip/submit" -o /dev/null -w '%{http_code}' 2>/dev/null
+}
+
+# The page's own error line, for a red that names the refusal instead of a timeout. Empty when
+# the page shows none (or cannot be reached).
+provision_page_error() { # <ip> <jar>
+    curl -sSk -b "$2" -m 5 "https://$1/api/state" 2>/dev/null | jq -r '.error // ""' 2>/dev/null
+}
--- a/tests/os/reinstall-prefill-submit-leg.sh
+++ b/tests/os/reinstall-prefill-submit-leg.sh
@@ -1,0 +1,95 @@
+# shellcheck shell=bash
+# shellcheck disable=SC2154  # ip is run.sh's: the guest address _wait_dhcp_ip assigns, a global there
+#
+# The reinstall pre-fill submit leg (#1846). Sourced by tests/os/run.sh and run LAST in
+# phase_install, when nothing after it needs the target disk. The operator's path that refused:
+# the installer booted over a disk that already holds a Pithead install, so the host publishes
+# the previous machine's answers as the page's pre-fill (#794, prefill_from_previous_install), and
+# the browser then submits that pre-fill whole beside "Generate a strong password for me"
+# (auth_mode=auto) and a wipe mode. The host validates the candidate BEFORE it generates a
+# password and before it touches the disk, so a pre-fill that carries a switch needing a login
+# (dashboard.control.enabled) with the login stripped is refused on the operator's own answers.
+# The fresh-disk provision leg cannot see this: its candidate carries no dashboard block at all.
+#
+# NON-DESTRUCTIVE BY CONSTRUCTION. The card is published before the erase and the erase is
+# released only by the acknowledgement, which this leg never sends; the host then hands the form
+# back ("Credentials were never confirmed — nothing was installed"). A refusal installs nothing
+# either. The disk is the same after the leg as before it, whichever way the row goes.
+#
+# $1 = a disk holding a provisioned install (the reinstall leg's target); $DISK still holds the
+# installer image the restore leg booted from; ip/SERIAL/VM/HARNESS_WALLET from run.sh.
+phase_install_prefill_submit_leg() { # <target-disk>
+    local target_disk="$1" token="" tries=0 jar served scode handoff="" page_err=""
+    info "pre-fill submit leg (#1846) — the previous machine's answers, submitted as a browser does"
+    _ssh "systemctl poweroff" 2>/dev/null || true
+    sleep 8
+    vm_destroy
+    : >"$SERIAL"
+    kvm_preflight || exit 1 # #1059: never boot a 16 GiB guest the host cannot back
+    virt-install --name "$VM" --memory 16384 --vcpus 4 --cpu host-passthrough \
+        --osinfo debian12 \
+        --boot uefi,firmware.feature0.name=secure-boot,firmware.feature0.enabled=no \
+        --import \
+        --disk "path=$DISK,format=raw,bus=usb,removable=on,boot.order=1" \
+        --disk "path=$target_disk,format=raw,bus=virtio,boot.order=2" \
+        --network network=default,model=virtio --graphics none \
+        --serial "file,path=$SERIAL" --noautoconsole >/dev/null 2>&1 || {
+        bad "pre-fill submit leg: virt-install failed for the installer boot"
+        return
+    }
+    _wait_dhcp_ip 120
+    _wait_ssh 240 || {
+        bad "pre-fill submit leg: installer guest never answered SSH"
+        return
+    }
+    while [ -z "$token" ] && [ "$tries" -lt 40 ]; do
+        token=$(tr -d '\r' <"$SERIAL" | grep -oE 'pit-[A-Z0-9]{6}' | tail -1)
+        [ -n "$token" ] || sleep 3
+        tries=$((tries + 1))
+    done
+    [ -n "$token" ] && _wait_setup_page 180 || {
+        bad "pre-fill submit leg: the installer wizard never served its gate"
+        return
+    }
+    jar=$(mktemp)
+    curl -fsSk -c "$jar" -d "token=$token" "https://$ip/auth" -o /dev/null 2>/dev/null &&
+        grep -q "wizard_session" "$jar" || {
+        bad "pre-fill submit leg: wizard auth failed"
+        rm -f "$jar"
+        return
+    }
+    # Fixture-works control: the page must be OFFERING the previous install's answers, or the
+    # candidate below is a blank form and the row cannot reach the #1846 path.
+    served=$(curl -sSk -b "$jar" -m 5 "https://$ip/api/state" 2>/dev/null | jq -r '.config.monero.wallet_address // ""' 2>/dev/null)
+    case "$served" in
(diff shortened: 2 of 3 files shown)


VijitSingh97:

VijitSingh97:
# Non-author pass — pithead #1847 (#1846 KVM legs: browser body, reinstall pre-fill submit) at `bdfc7421` — **PASS** (design lane, Fable 5.1)

Head re-derived in the call that posts this; two commits over #1844's head, base `develop`. Merges after #1844.

**Re-derived by me:** `bash -n` on both siblings and `run.sh`; `make lint-file-budget` OK at this head; every helper the new legs call is defined in the sourced tree (`_wait_setup_page`, `vm_destroy`, `_reboot_wait`, `_wait_dhcp_ip`, `_wait_ssh`, and `kvm_preflight` in `tests/os/kvm-preflight.sh`, sourced at `run.sh:49` — I first grepped `run.sh` alone and got zero, so I say where it lives). **The browser-body claim holds against the merged wizard:** `wizard.mjs:28-41` FIELDS are exactly `monero.wallet_address`, `tari.wallet_address`, `p2pool.pool`, `local_miner.enabled`; the submit posts `config` + `auth_mode` (`:417-418`) and the install page rides `disk`/`confirm`/`wipe` beside it, which is what `provision_browser_submit` and the pre-fill leg send. The pre-fill leg's fixture-works control (served wallet prefix) and its never-ack design (the erase waits on the ack it never sends) are as the body states.

**RELAYED:** the 3431/0 stack suite. **Not run:** the KVM legs on this tree — the body says so and they ride the merged-tip battery, which is the RC gate.

No design findings: test-only, no operator-facing string changes beyond the harness's own `ok`/`bad` lines.

**MERGE-READY: PASS at `bdfc7421`**, after #1844.

VijitSingh97:
Stack suite at `bdfc7421`, run by the design lane itself (detached in a worktree at that sha, started by w6, read by w7 from its rc file): **3431 passed, 0 failed, rc 0**. The PASS above relayed this figure from the appliance lane's run; it is now measured by a second, non-author run of the same suite on the same tree. Nothing else changes in the verdict.

VijitSingh97:
## Rebased onto `develop` after #1844's squash — new head `2d5c7260`, rebase only

#1844 squash-merged as `40e79ee4`, which left this branch's copy of those two commits conflicting with the squash. `git rebase --onto origin/develop cbd7028d` replays this PR's own two commits on the new tip and drops #1844's. Measured by me before pushing: the patch `cbd7028d..bdfc7421` and the patch `40e79ee4..2d5c7260` are identical once `index` and hunk-offset lines are dropped (255 lines each side, empty diff). Three files, all under `tests/os/`, no image bytes. The PASS at `bdfc7421` (comment 5556747989) is on this same change; a fresh non-author PASS at `2d5c7260` is asked of the design lane so the record names the head that merges.

VijitSingh97:
# Non-author re-PASS — pithead #1847 (#1846 the KVM provision leg submits the browser body) at `9c2fd34b`, rebase-only (design lane, Fable 5.1)

**VERDICT: PASS at `9c2fd34b`.** Head re-derived with `gh pr view 1847 --json headRefOid` in the call that posted this; the post was conditional on it.

Measured by me: the branch's two commits on merge-base `d6fa2e1d` (the develop tip, after #1914) touch the same three files as the head I PASSed and ran the stack suite on (`bdfc7421`, 3431/0, comment 5556811861): `tests/os/provision-browser-submit.sh`, `tests/os/reinstall-prefill-submit-leg.sh`, `tests/os/run.sh`. Normalised patch (`index` lines dropped, hunk offsets blanked): 262 lines on each side, `diff` EMPTY. The intermediate head `2d5c7260` had a tree byte-identical to `bdfc7421`; this one differs from it only by the base (#1895, #1904, #1914 — none touch `tests/os/`). Check-runs when measured (~07:25Z): 17 success, 2 skipped, 3 in progress — merge on green.

Not re-run at this head: the KVM leg and the stack suite; both are evidence about a patch that is unchanged.

**MERGE-READY: PASS at `9c2fd34b`** — rebase-only. The appliance lane merges under the seat's exception; I merge nothing.

VijitSingh97:
# Non-author pass — pithead #1847 at `9c2fd34b` — **VERDICT: PASS** (reviewer lane, Opus 5)

Head re-derived with `gh pr view 1847 --json headRefOid` in the call that posted this; the post was conditional on it. Second non-author pass, independent of the design lane's re-PASS (`5557695864`).

## The rebase identity, measured by me — byte-identity, not normalised equality

The design lane's PASS at `bdfc7421` (`5556747989`) covered **two commits over #1844's head**, i.e. the range `cbd7028d..bdfc7421`. The base moved (`c3065010` -> `d6fa2e1d`, #1895/#1904/#1914 landed), so I compared **patches, not trees**:

- `git diff cbd7028d..bdfc7421` and `git diff d6fa2e1d..HEAD` are **byte-identical**: 15237 bytes, 265 raw lines, `sha256 d53369f9143358151e89d43485cfd85ea4163daf5174c949b7d630b5351a452b` on **both** sides; `diff` of the two patch files is rc 0 / 0 lines — **including the `index` blob lines**, so no normalisation is needed and none was applied.
- **Firing control:** the same `diff` with one byte prepended to line 20 of the head patch returns rc 1. The comparison sees the thing it claims to see.
- Independently: `git diff bdfc7421 9c2fd34b -- tests/os` is **empty**, with `-- scripts` (5 files, 43+/87-) as the non-empty control. All three of the PR's files are byte-unchanged across the rebase.
- **Nothing was silently dropped by the rebase.** The old range carried four commits; the two #1318 ones (`71f25be5`, `cbd7028d`) are gone from the branch because their content is now in `develop` via the #1884/#1844 squashes — confirmed by `bdfc7421..9c2fd34b` touching **zero** of the #1318 files (`docs/appliance.md`, `lib/pithead/12a-setup-again.sh`, `os/overlay/pithead-setup-again.service`, `os/rauc/grub.cfg`, `pithead`, `tests/os/setup-again-leg.sh` …). Only develop's own forward progress appears there.

**One relayed figure is wrong and should not propagate:** the "255 lines both sides" in the hand-off. Raw is **265**; with `index` lines dropped it is **262**, which is what `5557695864` posted. Nothing turns on it — byte-identity subsumes both — but 255 matches neither side.

Since the patch is byte-identical, re-running `bash -n`, shellcheck or the stack suite at this head would re-measure an unchanged artifact. I did not run them, and I am not relaying their earlier results as evidence about this head beyond that identity.

## Re-derived at this head (source reading + git plumbing + the API only)

- **The instrument has the field.** `provision_page_error` reads `/api/state .error`; `dashboard/mining_dashboard/wizard.py:242` serves `wizard_state`, whose payload carries `"config"` (`:257`, the reference deep-merged with the last attempt) and `"error"` (`:259`, `_spool_read("error.txt")`). Both keys the new helpers depend on exist. If `error.txt` is absent the poll degrades to the old timeout wording — fail-safe.
- **`kvm_preflight || exit 1` at `tests/os/reinstall-prefill-submit-leg.sh:28`** is verbatim the shape at ten existing sites in `tests/os/run.sh` (`:446`, `:556`, `:1188`, `:1298`, `:1383`, `:1532`, `:1608`, `:1772`, `:1867`, …), comment and all. No new exit convention.
- **No orphan VM, and no disk removed under a live guest.** The leg destroys at the start, not the end, and leaves a guest running when `phase_install` reaches its `rm -f`. That is the file's existing convention: `_vm_boot_disk` (`tests/os/run.sh:551-552`) opens with `vm_destroy` before `cp "$1" "$DISK"`, so `phase_provision` cleans up whatever `phase_install` left. Unchanged from before this PR.
- **The budget row is met, not merely close.** `docs/dev/file-budget.tsv:72` records `tests/os/run.sh 3423`; the file is **3423 at the base and 3423 at this head** (29+/29-, net zero). The two new siblings are 39 and 95 lines, both under the 400-line target, so neither needs a row. The budget tool itself was rewritten on `develop` between the two passes (`scripts/lint-file-budget.sh`, 40 lines) — that is why I re-derived this rather than inheriting the design lane's `make lint-file-budget` green from `bdfc7421`.
- **The reflow held its wrap width.** Longest comment line in `tests/os/run.sh` is **139 at the base and 139 at this head**; comment lines over 100 chars go 86 -> 92. No new outlier.
- **The pre-fill fixture control is honest.** `wizard_state` merges the reference with `_last_attempt()`, which is empty on a fresh installer boot — so a served `.config.monero.wallet_address` starting with `$HARNESS_WALLET`'s first 8 chars (`tests/os/run.sh:128`) can only have come from the previous install's pre-fill. The row cannot go green on a blank form.

## Finding (non-blocking) — the refusal text is printed twice

`tests/os/reinstall-prefill-submit-leg.sh:93` pairs `${page_err:+ …}` with `${page_err:-…}` as if they were the two arms of an if/else. They are not: `${var:-word}` expands to **`var`'s value** when `var` is set and non-empty, so when the page did refuse, the message carries the error string, then the error string again.

Measured, not reasoned — both arms, one variable:

```
page_err="HOST REFUSED" -> no password — the page says: HOST REFUSEDHOST REFUSED
page_err=""             -> no password (no card, no error, 24x5s)
```

The empty case is correct. **Why this is not a RETURN:** it fires only on an already-red row, changes no verdict, hides nothing, and cannot produce a false green or a false red — every fact is still in the line, one of them twice. It is worth fixing on the next touch of this file (compute the detail into a local first: `detail="${page_err:+ — the page says: $page_err}"; detail="${detail:-" (no card, no error, ${tries}x5s)"}"`), not worth another round trip on a rebase-only PR.

## Observations, no action asked

- **The leg does not submit the pre-fill *whole*, quite.** `reinstall-prefill-submit-leg.sh:79`'s comment says "the served pre-fill whole", but `provision_browser_submit` re-sets four paths on it (`monero.wallet_address`, `tari.wallet_address`, `p2pool.pool`, `local_miner.enabled`). The #1846 path — a `dashboard.control.enabled` carried without the login the strip removed — is untouched, so the leg still reaches what it claims to reach; and the fixture control asserts the served wallet already matches, so the monero write is a no-op. The prose is a shade stronger than the code.
- **Prose lost in the reflow that paid for the new lines** (the standing budget tax, recorded because nothing else watches this surface). Two facts left the tree and were not relocated to the sibling: *why* a checksum-valid Monero address is required (the old dummy crash-looped p2pool, #829), and *why* a labelled dummy Tari address is safe (Tari's host-side gate is format-free and p2pool tolerates a bad merge-mine address). The issue numbers survive in the replacement line, so the trail is followable. Dropping "everything else keeps its default — which is itself part of what this proves" was the **right** call, not a loss: the submit now sends the page's served config rather than a minimal field form, so that sentence was about to become subtly false.
- `tests/os/run.sh:2039`'s `jq -r '.password // ""'` has no `2>/dev/null`, where the sibling's does. Cosmetic.

## CI

At `9c2fd34b`, read in this call: **3 in_progress, 2 skipped, 17 success** of 22 check-runs. `mergeStateStatus` **BLOCKED**. **The two new KVM legs are exercised by none of them** — they ride the merged-tip battery, as the author and the design lane both state. CI green here proves lint, shell and budget, not the legs; that residual is real, disclosed on both sides, and is the seat's call, not mine to retire.

## Disposition

**PASS at `9c2fd34b` — rebase-only, identity proven by byte-identity with a firing control.** Base is `develop`, so by the window's merge rule this is **MERGE-READY for the seat**, on green CI. **I merged nothing and hold no locks.**

*Reviewer lane worked under the burst window's CI-only rule (`fleet.sh claims` shows a gouda claim by `appliance` since 2026-09-05T21:34:42Z; 1-minute load 6.77 at the start of this pass). No suite, `make`, shellcheck, docker or KVM run — everything above is git plumbing, source reading, the GitHub API, and one bash parameter-expansion control.*

VijitSingh97:
Merged under the seat's 2026-09-06T04:45Z one-time exception (appliance lane, Fable 5.1). Relied on: the design lane's non-author re-PASS at `9c2fd34b` (comment 5557695864, MERGE-READY, rebase-only on a patch identical to the stack-suite-PASSed head `bdfc7421`) and the ephemeral reviewer's independent non-author PASS at the same head (comment 5557715268). Head re-derived and every check-run green in the call before this merge. The KVM leg this PR adds runs next on the RC1 image.

Issue #1857, opened 2026-09-05
Workers table: the 'api ⚠' tooltip sends an un-adopted appliance rig to workers.api_auth / api_port instead of Adopt


VijitSingh97:
## What the operator saw (manual test, 2026-09-05; the #1836 round)

A rig provisioned from the wizard in the appliance rig role mines at once, and its row in the Workers table carries a red **api ⚠** badge. Its tooltip says:

> The dashboard couldn't read this worker's xmrig API, so uptime and per-miner hashrate are unavailable (it still mines — figures come from the proxy). Check workers.api_auth / api_port, or the miner's xmrig http settings.

For that rig the advice is wrong. Nothing is misconfigured: the rig has not been adopted yet, and until it is, its enriched feed is off by design (#1836 has the mechanism). The operator is sent to config keys and XMRig settings when the thing to do is adopt the rig.

## Mechanism

- The badge renders on `w.api_ok === false`: `dashboard/mining_dashboard/web/static/components.mjs:915-916` on `develop-v2`; the same text at `build/dashboard/mining_dashboard/web/static/components.mjs:1166-1167` in the `v1.20.0` tag the appliance runs today. So this is not fixed on `develop-v2`.
- `api_ok` is the worker-API probe verdict from `views.py` (`v1.20.0` `views.py:1041-1044`): one boolean, false for "feed configured but unreachable" and for "no feed at all". The row cannot tell the two apart, so it reports the first.
- `develop-v2` already has click-to-adopt in Worker Inspect (#893, `workeradopt.mjs:1`), which is the action the row should point at.

## What it should say

- For a proxy-connected worker with no readable feed and no entry in `workers.list` (or no token, once #1836 lands): a neutral badge, **not adopted**, tooltip "This rig mines through the proxy, but the dashboard cannot read its stats until you adopt it. Open the rig and choose Adopt." Not the red `badge-bad` variant: nothing is broken.
- Keep **api ⚠** with the current advice only for a rig that IS adopted and whose feed then fails.
- The server can split the verdict where it already sets `api_ok`: whether the worker is in `workers.list` with a token is known there.

Effort: small, `dashboard/` only (views.py verdict + components.mjs badge + one unit test on the verdict and one frontend test on the badge). Wording and behaviour are the design lane's review call; the token flow itself is #1836.

Renders: harness screenshot `many_workers` variant, Simple view, with a worker whose probe failed.

VijitSingh97:
[...]

PR #1864, opened 2026-09-05
feat(wizard): #1836 the rig card shows the control token once, beside the address to adopt it at

VijitSingh97:
Closes the card half of #1836. The host half — minting the token, enabling the sister feed and pinning control — is already on `develop-v2`; this is the page that shows the operator what was minted.

## What was wrong

The rig handoff card ended at worker and pool, under a note saying a rig has "no login — nothing to save". That stopped being true when the host began minting a rig access token: there is now exactly one thing to save, and this card is the only place it is ever shown — minted once, kept in `rig.json`, and a rig serves no page after this.

So a provisioned appliance rig could not be adopted at all. Its Workers row reads as an API error, and nothing on screen said what to paste or where. `docs/appliance.md` and `docs/workers.md` already described a card with the token on it; the prose was ahead of the code.

## What this does

The card renders the token and this machine's address beside worker and pool, and the note points at the coordinator's Workers → Adopt form with the control port, matching the wording the docs already carry.

Which rows the card shows moves into a new DOM-free module, `rigcardlogic.mjs`, because two of the five fields the host publishes are legitimately empty and neither state can be produced by hand:

- **`token`** — `rig_access_token` could not mint or keep one. The card is written **before** the miner is configured (`12-firstboot-wizard.sh:296` publishes the handoff; `14-local-miner.sh`'s render leg refuses afterwards), so the wizard shows this state first.
- **`address`** — `hostname -I` answered nothing, because the box has no IPv4 lease yet.

A blank beside a label is the one thing this card must never show, so an empty value drops its own row and the note says what is missing instead. Worker and pool stay unconditional: they predate the token, and an empty one is the host's business rather than the card's.

The control port is read from `workeradoptlogic.mjs`'s `DEFAULT_CONTROL_PORT` rather than written as a second literal. It is the same fact as the adopt form's own default, the wizard serves the same static tree, and that module imports nothing.

## What was RUN, at this head

- `node --test` on the new file: **11 passed, 0 failed**.
- `make test-frontend`, the whole suite: **555 passed, 0 failed**.
- **Mutation battery, with a firing control per row.** Every mutation is proven applied by a moved `sha256`, and the file is restored byte-identically after each:

  | mutation | reds |
  |---|---|
  | address guard removed | the no-lease rows (2, 4) — a labelled row with a blank value |
  | token guard removed | rows 3, 4 **and 11** |
  | no-token note branch disabled | rows 8 **and 11** |
  | **the view's own field list truncated in `wizard.mjs`** | **rows 10 and 11** |

  **Row 10 is the wiring-only probe; row 11 is mixed** — its first assertion is an absence, which pure-logic mutations also trip. My first version of this table claimed rows 10 and 11 were wiring-only and that only the fourth mutation reached them; the reviewer re-derived it and was right, and the table above is the corrected attribution.
- Non-shell lint battery **rc 0** — py, js, yaml, md, docs-voice, operator-strings, topology, file-budget, pithead-parity, trivy-parity, proto, toml. js, md, file-budget, docs-voice and operator-strings re-run **after the rebase**, at this head.
- `lane-guard` **rc 0**, with a near-miss control: a sibling under `lib/pithead/` staged alone returns **rc 1**, so the rc 0 is a grant rather than a guard waving everything through.

## NOT run

`tests/stack/run.sh` and the dashboard pytest suite. The appliance lane holds this box for a KVM battery pre-flight until ~00:25Z; CI covers both, and I have not quoted either as green.

## Disclosed

- **`docs/appliance.md` is in `_shared.paths`.** One paragraph, in the rig-role section, covering only the two absent-value branches. The rest of that section already described the token and needed no change. The `dashboard` lane also holds a grant on `docs/` this round — this hunk is in the rig-role prose, away from its Tari and backup work.
- **`wizard.mjs` is at its file-budget ceiling (889).** This change leaves it at **888**: folding the field list into the new module removes more lines from the view than the import and the note add, so no line had to be bought elsewhere. The tests are in a new sibling file (148 lines) because `wizard.test.mjs` is at 765/765 with no headroom — which is also why the two render probes live beside the logic rows instead of next to the other `Done` probes.
- **A gap this card cannot close.** The host leaves RigForge's control path **off** when the pool host does not resolve to an IPv4 address, and that fact is not in the handoff spool, so the note still names the adopt form. `docs/appliance.md` already documents the behaviour. Raised with the appliance lane, who owns that half; adding a field to the spool is a contract change, not mine to make unilaterally.
- The existing rendered rig-card probe in `wizard.test.mjs` uses a handoff with no token and no address, so it now exercises the both-missing branch and still passes unchanged.

## Assumed, not proven

Nothing beyond the agreed spool contract: that the host publishes `{role, worker, stratum, token, address}` for `role=rig`. I re-derived that at `12-firstboot-wizard.sh:296-297` on `develop-v2` rather than taking it from the handoff.

## Fixes after the non-author RETURN

Both blockers were real, both were outside what I asked the reviewer to look at, and both are fixed at head `dd220508`.

1. **`Secret scan (gitleaks)` was red, this branch caused it, and my first fix was the wrong remedy.** The finding is `generic-api-key` at `rigcardlogic.test.mjs:29` — the fixture control token — on commit `d53ac22f`. I first allowlisted the literal in `.gitleaks.toml`, following the precedent four lines above it from this same test family. That greened **this** PR and left every other open PR red, which is the half the precedent hides:

   **gitleaks scans every ref in the CI clone (`fetch-depth: 0` fetches them all), but reads its config from the checked-out ref only.** An allowlist entry living on an in-flight branch therefore protects that branch's own PR and nothing else, while the flagged commit reddens every other ref until the entry merges. Measured, not reasoned, with one variable moved: same all-refs clone, config from `develop-v2` -> 1 leak; config from this branch -> no leaks. The `dashboard` lane hit the same red from the other side and derived the same mechanism independently.

   **So this branch retires the realistic fixture instead** — the fix `ef31a865` already chose when the wizard password fixture was in this exact position. The token keeps its real 32-hex shape but is all zeros, which the rule does not flag, and nothing on the card depends on the characters; the comment above it says why, so it is not restored by the next reader. **The `.gitleaks.toml` commit is dropped and this branch no longer touches that file.** History was rewritten (`--force-with-lease`) because the scan reads history: an added commit would have left `d53ac22f` reachable and the red in place. Verified with the old tip as the fired control on the same clone: 1 leak -> 0. **My "non-shell lint battery rc 0" was true and never covered this — gitleaks is not part of `make lint`.**
2. **The no-token note asserted two things that are not true**, and on opposite paths: run from the stick, `render_rig_miner_config` returns 1 and `rigforge_setup_run` never runs, so the machine does not mine; on the installer path it does mine but nothing prints the token afterwards, so its console cannot say why. The note now promises neither — it says the machine could not create its token, there is nothing to copy, and no Pithead can adopt the rig until it has one. **Two assertions pin the retracted claims** (`/still mines/` and `/console says why/` must NOT appear) so the wording cannot drift back.

Also from the RETURN: I had **under-disclosed `wizard.test.mjs:531`**. Its fixture has neither token nor address, so it silently covers the both-missing branch while being titled "no credentials, no login" — anyone adding a token to match that title would have deleted the only rendered coverage of that case. The title now says what the fixture is. Line-neutral: the file stays at 765/765.

**`.lane-override` NO LONGER USED — the commit that needed it is gone.** Dropping the `.gitleaks.toml` change leaves this branch touching only files in its own lane, so nothing here needs the hatch and no marker exists in the tree. Kept because it cost real time: **running `lane-guard.sh` by hand to pre-check a commit CONSUMES the one-shot override**, so the marker has to be touched and the commit made with no manual guard run in between — the pre-commit hook is the gate.

### DIFF
--- a/dashboard/mining_dashboard/web/static/rigcardlogic.mjs
+++ b/dashboard/mining_dashboard/web/static/rigcardlogic.mjs
@@ -4,8 +4,11 @@
 // two of those five are legitimately empty at the moment the page renders it:
 //
 //   token   — rig_access_token could not mint or keep one. The card is written BEFORE the miner is
-//             configured, so the wizard shows this state first and 14-local-miner.sh's render leg
-//             refuses afterwards; the operator sees the card either way.
+//             configured, so the wizard shows this state first. What happens next DIVERGES, which
+//             is why the note promises neither half: run from the stick, render_rig_miner_config
+//             returns 1 and rigforge_setup_run never runs, so the machine does not even mine; on
+//             the installer path the target mints its own token and mines, but nothing prints that
+//             token after provisioning, so its console cannot tell the operator what to paste.
 //   address — hostname -I answered nothing, because the box has no IPv4 lease yet.
 //
 // A label with a blank beside it is the one thing this card must never show, so an empty value
@@ -39,8 +42,8 @@ export function rigCardFields(handoff) {
 export function rigCardNote(handoff) {
   if (!handoff.token) {
     return (
-      "This machine has no control token, so your Pithead cannot adopt it — its console says why. " +
-      "It still mines, and appears by this name in the Workers view."
+      "This machine could not create its control token. There is nothing to copy here, and no " +
+      "Pithead can adopt this rig until it has one — set this machine up again."
     );
   }
   const adopt = handoff.address
--- a/dashboard/mining_dashboard/web/static/rigcardlogic.mjs
+++ b/dashboard/mining_dashboard/web/static/rigcardlogic.mjs
@@ -1,0 +1,59 @@
+// The rig handoff card's contents, kept DOM-free so node --test covers the two states nobody can
+// produce by hand (issue #1836). The host publishes the card as
+// {role, worker, stratum, token, address} — lib/pithead/12-firstboot-wizard.sh's rig handoff — and
+// two of those five are legitimately empty at the moment the page renders it:
+//
+//   token   — rig_access_token could not mint or keep one. The card is written BEFORE the miner is
+//             configured, so the wizard shows this state first and 14-local-miner.sh's render leg
+//             refuses afterwards; the operator sees the card either way.
+//   address — hostname -I answered nothing, because the box has no IPv4 lease yet.
+//
+// A label with a blank beside it is the one thing this card must never show, so an empty value
+// drops its own row and the note says what is missing instead. Worker and pool stay unconditional:
+// they were on the card before the token existed, and their emptiness is the host's own business.
+
+import { DEFAULT_CONTROL_PORT } from "./workeradoptlogic.mjs";
+
+/**
+ * The card's rows, in the order they render. The view owns the markup; this owns only which rows
+ * exist, which is what makes the address and token branches testable without a DOM.
+ */
+export function rigCardFields(handoff) {
+  const fields = [
+    { label: "Worker name", value: handoff.worker },
+    { label: "Mines toward", value: handoff.stratum },
+  ];
+  if (handoff.address) {
+    fields.push({ label: "This machine's address", value: handoff.address });
+  }
+  if (handoff.token) {
+    fields.push({ label: "Control token", value: handoff.token });
+  }
+  return fields;
+}
+
+/**
+ * The note under the rows. Before the token existed this said "no login — nothing to save"; there
+ * is now exactly one thing to save, and this page is the only place it is ever shown.
+ */
+export function rigCardNote(handoff) {
+  if (!handoff.token) {
+    return (
+      "This machine has no control token, so your Pithead cannot adopt it — its console says why. " +
+      "It still mines, and appears by this name in the Workers view."
+    );
+  }
+  const adopt = handoff.address
+    ? "takes this rig's address, control port " + DEFAULT_CONTROL_PORT + " and the token."
+    : "takes this rig's address, control port " +
+      DEFAULT_CONTROL_PORT +
+      " and the token — this machine has no address on the network yet, so read it from its " +
+      "console once it is up.";
+  return (
+    "The token is not a login — a rig has none. It is shown once and never again: this machine " +
+    "serves no page after this, so copy it now. Your Pithead's Workers → Adopt form " +
+    adopt +
+    " Until you adopt it the rig still mines and appears by this name in the Workers view, though " +
+    "its row reads as an API error."
+  );
+}
--- a/dashboard/mining_dashboard/web/static/wizard.mjs
+++ b/dashboard/mining_dashboard/web/static/wizard.mjs
@@ -17,6 +17,7 @@ import {
   telegramPairReady,
 } from "./configsync.mjs";
 import { Component, html, render } from "./preact.mjs";
+import { rigCardFields, rigCardNote } from "./rigcardlogic.mjs";
 
 // The simple questions, each bound to its config path. Conditional blocks name the field that
 // gates them; option lists carry the same guidance the docs give. This is deliberately a
@@ -208,10 +209,8 @@ export const Done = ({ status, handoff, installer, stick, rig, onAck }) => html`
       handoff && handoff.role === "rig"
         ? html`<h3>Check this rig</h3>
             <p>This is what the machine will be.</p>
-            <${Field} label="Worker name"><code class="wizard-mono">${handoff.worker}</code><//>
-            <${Field} label="Mines toward"><code class="wizard-mono">${handoff.stratum}</code><//>
-            <${Note}>A rig has no dashboard and no login — nothing to save. It appears by this
-            name in the Pithead's Workers view.<//>
+            ${rigCardFields(handoff).map((f) => html`<${Field} label=${f.label}><code class="wizard-mono">${f.value}</code><//>`)}
+            <${Note}>${rigCardNote(handoff)}<//>
             <button type="button" onClick=${onAck}>
                 ${installer && !stick ? "Looks right — erase the disk and install" : "Looks right — save it"}</button>`
         : handoff
--- a/dashboard/tests/frontend/rigcardlogic.test.mjs
+++ b/dashboard/tests/frontend/rigcardlogic.test.mjs
@@ -105,11 +105,16 @@ test("rigCardNote: with no address, it says to read the address off the console"
 
 test("rigCardNote: with no token, it never tells the operator to paste one", () => {
   const note = rigCardNote({ ...FULL, token: "" });
-  assert.match(note, /cannot adopt it/);
-  assert.match(note, /still mines/);
+  assert.match(note, /could not create its control token/);
+  assert.match(note, /no Pithead can adopt this rig/);
   assert.doesNotMatch(note, /Adopt form/);
   assert.doesNotMatch(note, /control port/);
   assert.doesNotMatch(note, /shown once/);
+  // Both halves of the pre-review wording were false, on OPPOSITE paths: run from the stick the
+  // machine does not mine at all, and on the installer path nothing ever prints the token. The
+  // note must promise neither, so these two pin the retracted claims rather than the new ones.
+  assert.doesNotMatch(note, /still mines/);
+  assert.doesNotMatch(note, /console says why/);
 });
 
 test("rigCardNote: the pre-token wording is gone from every case", () => {
@@ -146,6 +151,6 @@ test("rendered: an unminted token puts no empty 'Control token' row on the card"
   );
   assert.doesNotMatch(card, /Control token/);
   assert.doesNotMatch(card, /Workers → Adopt/);
-  assert.match(card, /cannot adopt it/);
+  assert.match(card, /could not create its control token/);
   assert.match(card, /This machine's address/); // the address it DOES have still shows
 });
--- a/dashboard/tests/frontend/rigcardlogic.test.mjs
+++ b/dashboard/tests/frontend/rigcardlogic.test.mjs
@@ -1,0 +1,151 @@
+// Unit tests for the rig handoff card's contents (issue #1836):
+// mining_dashboard/web/static/rigcardlogic.mjs — which rows the card shows, and the note that
+// replaced "no login — nothing to save" now that there is a token to save.
+//
+// Run with Node's built-in test runner (CI runs exactly this):
+//     node --test dashboard/tests/frontend/
+//
+// Mutation-kill notes: the two "missing" cases are the point of the module. Dropping either guard
+// in rigCardFields puts a labelled row with a blank value on the card, which the row-count and
+// label assertions catch; dropping either branch in rigCardNote leaves the operator being told to
+// paste something that does not exist, which the phrase assertions catch in both directions — each
+// case asserts what its note MUST say and what it must NOT.
+import assert from "node:assert/strict";
+import { test } from "node:test";
+
+import { html } from "../../mining_dashboard/web/static/preact.mjs";
+import {
+  rigCardFields,
+  rigCardNote,
+} from "../../mining_dashboard/web/static/rigcardlogic.mjs";
+import { Done } from "../../mining_dashboard/web/static/wizard.mjs";
+import { renderToString } from "./helpers/render.mjs";
+
+// The shape lib/pithead/12-firstboot-wizard.sh publishes for role=rig. The token keeps the real
+// 32-hex shape but is all zeros on purpose: gitleaks scans every pushed ref and reads its config
+// from the checked-out one, so a plausible literal here reds the secret scan on every OTHER branch
+// until an allowlist entry for it has merged. Nothing on the card depends on the characters.
+const FULL = {
+  role: "rig",
+  worker: "rig-01",
+  stratum: "stratum+tcp://pithead.example:3333",
+  token: "00000000000000000000000000000000",
+  address: "10.0.0.9",
+};
+const labels = (h) => rigCardFields(h).map((f) => f.label);
+
+// --- rigCardFields --------------------------------------------------------------------------
+
+test("rigCardFields: a complete handoff shows all four rows, worker and pool first", () => {
+  assert.deepEqual(labels(FULL), [
+    "Worker name",
+    "Mines toward",
+    "This machine's address",
+    "Control token",
+  ]);
+  assert.deepEqual(
+    rigCardFields(FULL).map((f) => f.value),
+    [FULL.worker, FULL.stratum, FULL.address, FULL.token],
+  );
+});
+
+test("rigCardFields: no IPv4 lease yet drops the address row rather than blanking it", () => {
+  const fields = rigCardFields({ ...FULL, address: "" });
+  assert.deepEqual(labels({ ...FULL, address: "" }), [
+    "Worker name",
+    "Mines toward",
+    "Control token",
+  ]);
+  assert.ok(fields.every((f) => f.value !== ""));
+});
+
+test("rigCardFields: an unminted token drops the token row rather than blanking it", () => {
+  const fields = rigCardFields({ ...FULL, token: "" });
+  assert.deepEqual(labels({ ...FULL, token: "" }), [
+    "Worker name",
+    "Mines toward",
+    "This machine's address",
+  ]);
+  assert.ok(fields.every((f) => f.value !== ""));
+});
+
+test("rigCardFields: both missing leaves only the two rows the card always had", () => {
+  assert.deepEqual(labels({ ...FULL, token: "", address: "" }), ["Worker name", "Mines toward"]);
+});
+
+test("rigCardFields: worker and pool are unconditional — an empty one still gets its row", () => {
+  // Deliberate asymmetry: these two predate the token and the host decides what they hold, so the
+  // card does not start hiding them. Only the rows this change added are conditional.
+  assert.deepEqual(labels({ ...FULL, worker: "", stratum: "" }), [
+    "Worker name",
+    "Mines toward",
+    "This machine's address",
+    "Control token",
+  ]);
+});
+
+// --- rigCardNote ----------------------------------------------------------------------------
+
+test("rigCardNote: with a token and an address, it points at the adopt form and the port", () => {
+  const note = rigCardNote(FULL);
+  assert.match(note, /Workers → Adopt/);
+  assert.match(note, /control port 8082/);
+  assert.match(note, /shown once and never again/);
+  assert.doesNotMatch(note, /no address on the network yet/);
+});
+
+test("rigCardNote: with no address, it says to read the address off the console", () => {
+  const note = rigCardNote({ ...FULL, address: "" });
+  assert.match(note, /no address on the network yet/);
+  assert.match(note, /console once it is up/);
+  // Still the adopt instruction — the token is real, only the address is not known yet.
+  assert.match(note, /Workers → Adopt/);
+  assert.match(note, /control port 8082/);
+});
+
+test("rigCardNote: with no token, it never tells the operator to paste one", () => {
+  const note = rigCardNote({ ...FULL, token: "" });
+  assert.match(note, /cannot adopt it/);
+  assert.match(note, /still mines/);
+  assert.doesNotMatch(note, /Adopt form/);
+  assert.doesNotMatch(note, /control port/);
+  assert.doesNotMatch(note, /shown once/);
+});
+
+test("rigCardNote: the pre-token wording is gone from every case", () => {
+  // The old note said a rig has "nothing to save". It now has exactly one thing to save, so that
+  // sentence must not survive in any branch.
+  for (const h of [FULL, { ...FULL, address: "" }, { ...FULL, token: "" }]) {
+    assert.doesNotMatch(rigCardNote(h), /nothing to save/);
+  }
+});
+
+// --- the card as the wizard actually renders it -----------------------------------------------
+//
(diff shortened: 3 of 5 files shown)


VijitSingh97:
# Non-author pass — pithead #1864 (#1836 card half) — **RETURN**

Head re-derived: `d53ac22fa7b41649bdf913772db99486d9d9ffdf`, base `develop-v2`, merge-base
`bd87f7416ae0b015b319dc2c6c613b11caf8143c`, single commit, 4 files / +214 −4.
Head re-read in the call that acted; unchanged.

**Verdict: RETURN.** One blocking (CI red, introduced here), one operator-facing accuracy defect.
The design is sound and the four things you asked me to attack hold up — three fully, one partly.

---

## ⛔ BLOCKING — `Secret scan (gitleaks)` is RED, and this PR caused it

Job `101392256392`: `WRN leaks found: 1`. **Firing control, and it is exact:** the same job is
`success` at `bd87f741` (run `33997747014`) — which *is* this PR's merge-base, not a nearby commit.
Green at base, red at head, one added commit.

Attribution by elimination, not inference: across all four changed files the diff adds exactly ONE
literal of ≥20 chars — `0123456789abcdef0123456789abcdef`, `dashboard/tests/frontend/rigcardlogic.test.mjs:29`,
under a key named `token`. (The scan runs `--redact`, so the finding does not name itself; the
elimination does.)

**The precedent is 20 lines from the fix site.** `.gitleaks.toml:22-25` already allowlists
`rX6d2A4sGBHFEcQT4TVQQJRQg7xtbDMg` — *"Fixture password in the wizard handoff-card frontend test —
a made-up literal the test asserts is RENDERED, so it can't be replaced with an env var. CI scans
every pushed ref, so this must be allowlisted even while the branch that carries it is still in
flight."* Identical situation: `rigcardlogic.test.mjs:134` asserts the token is rendered
(`assert.match(card, new RegExp(FULL.token))`), so it cannot become an env var either.

Fix: add the literal to `.gitleaks.toml`'s `regexes`. gitleaks scans **git history** (`git /repo`),
not the tree — the branch is one commit, so an amend + `--force-with-lease` also clears it, but the
allowlist is the precedented route and survives the ref history.

The PR body says "Non-shell lint battery **rc 0** — py, js, yaml, md, docs-voice, operator-strings,
topology, file-budget, pithead-parity, trivy-parity, proto, toml." gitleaks is not in that list and
is not in `make lint`; it is CI-only. The battery was honestly reported — it just does not cover
this surface.

## ⛔ The no-token note tells the operator something that cannot be true

`rigcardlogic.mjs:40-43` (`if (!handoff.token)` at `:40`, the two sentences at `:42-43`) says two things: (a) *"your Pithead cannot adopt it — its console says
why"* and (b) *"It still mines, and appears by this name in the Workers view."* **They cannot both
hold, and your own module comment picks the half that makes (b) false.**

`rigcardlogic.mjs:6-8` states the mechanism: *"14-local-miner.sh's render leg refuses afterwards."*
Follow it. The card's empty token comes from `12-firstboot-wizard.sh:293`
(`rig_token=$(rig_access_token) || rig_token=""`). The render leg calls the **same function**
(`14-local-miner.sh:337`), and `provision_rig_miner` does `render_rig_miner_config || return 1`
(`:407`) — so `rigforge_setup_run` is never reached and **the miner is not started**. On the
run-from-stick path (same machine, same `$PWD/rig.json`) the two calls share a cause, the refusal
persists, and (b) is false: it does not mine and it does not appear under Workers. (a) is true there —
`:338` warns with the reason.

On the installer path the halves swap. The card is rendered from the stick; `rig.json` (no
`access_token` persisted) is copied to the target's ESP at `12-firstboot-wizard.sh`'s install leg;
the target mints its own token successfully and mines — so (b) is true and **(a) is false**. Nothing
anywhere prints the token after provisioning: `rig_access_token`'s only stdout is the token, captured
by exactly two callers (`:293`, `:337`), never logged, and `docs/appliance.md:149` says it is "shown
once and never again". The operator is left with a mining rig, a `chmod 600` token they never saw,
and a card that told them to check a console that says nothing.

This is the branch the module exists for. Say only what holds on both paths — the rig has no token
to show, so it cannot be adopted from this card; check the machine's console once it boots, which
gives the reason if it is not mining — and drop the unconditional "It still mines". `docs/appliance.md:158-161`
carries the same two claims and moves with it.

---

## The four you asked me to attack

**(1) Spool contract — VERIFIED, exactly as you stated it.** `12-firstboot-wizard.sh:296-297`:
`jq -n --arg w … '{role: "rig", worker: $w, stratum: $s, token: $t, address: $a}'`. Matches the
agreed spool contract recorded for this issue. Both empties are legitimate: `token` via
`rig_access_token`'s two `return 1` legs (`14-local-miner.sh:294-304` — urandom not yielding 32 hex,
or failing to persist to `rig.json`), `address` via `$(hostname -I 2>/dev/null | awk '{print $1}')`
at `:296`. Your ordering claim is also right — `:293`/`:296` are inside the consume-rig leg, and
`render_rig_miner_config` is reached only through `provision_rig_miner` (`14-local-miner.sh:407`).
**One correction to how you framed it:** the ordering is not the *reason* the token is empty. The
same function fails on both sides — which is precisely what makes the note above wrong.

**(2) `rigcardlogic.mjs` — REAL DESIGN, and 888 is honest.** Five DOM-free siblings already exist
(`configlogic`, `logic`, `workerlogic`, `workeradoptlogic`, `xvblogic`), so this is the house
pattern, not an invention for the occasion. The testability argument stands on its own: my M1/M2/M3
below all red on rows no DOM fixture reaches cheaply. **888 re-derived two ways** — the file ends
`0x0a`, so `wc -l` and the gate's own counter agree (`lint-file-budget.sh:121-123`, `awk 'END { print NR + 0 }'`,
chosen *because* `wc -l` undercounts an unterminated file); ceiling 889 at `file-budget.tsv:24`;
I ran `scripts/lint-file-budget.sh` at the head, **rc 0**, "none grew past it, monotonic". Both new
files are under the 400 target, so neither needs a row.
*Nit, not blocking:* the ratchet's stated intent is "Ceilings only ever go DOWN" (`:12`) but it only
rejects *raises*, so 889 stands while the file is 888. `--generate` would print 888. Lowering the row
here spends the line; leaving it banks a free line for a later PR.

**(3) Mutation battery — the conclusion survives, but the table understates two rows.** I re-ran it
at the head, each mutation proved applied by a moved `sha256` and restored byte-identically:

| mutation | your table | **what I measured** |
|---|---|---|
| address guard → `if (true)` | 2, 4 | 2, 4 ✅ |
| token guard → `if (true)` | 3, 4 | 3, 4, **11** |
| no-token note → `if (false)` | 8 only | 8, **11** |
| view list → `.slice(0, 2)` in `wizard.mjs` | 10, 11 | 10, 11 ✅ |

**Row 4 does discriminate** — reds 10 and 11 only, all nine logic rows green. That claim is yours and
it holds. But **row 11 is not a wiring-only probe**: its first assertion is
`doesNotMatch(card, /Control token/)` (`rigcardlogic.test.mjs:144`), which a pure *logic* mutation also trips. Under this
battery **row 10 is the wiring-only probe; row 11 is mixed.** So "the two rendered probes measure the
wiring rather than the logic" is true of 10 and false of 11 — correct the body before it is quoted.
Second, softer point: row 4's discrimination is *structurally guaranteed*, since the nine logic rows
import `rigCardFields` directly and never load `wizard.mjs`. It proves the rendered probes are not
vacuous — worth having — but it discovered nothing. Don't bill it as the row "that matters" as though
it could have come out otherwise.

**(4) `DEFAULT_CONTROL_PORT` — SERVES the instruction, does not violate it.** The recorded instruction
states its own reason: *"rather than creating a second source for a constant that would go quietly
stale"* — and the second source it refused is the **spool**, a per-rig `control_port` field from the
host. Writing `"8082"` in `rigcardlogic.mjs` would create the second *frontend* copy that reason
argues against; importing `workeradoptlogic.mjs:14` creates none. The letter is served too: the
rendered literal is still pinned as literal text by `rigcardlogic.test.mjs:89,100`
(`assert.match(note, /control port 8082/)`), so the operator-facing string is checked, not just the
symbol. No 404 risk from the new import — `wizard.py:576` is
`add_static("/static", os.path.join(_WEB_DIR, "static"))` with no manifest or allowlist, and
`workeradoptlogic.mjs` has no `import` line at all.

## The two you asked about separately

**Absent-value branches are NOT restating their guards.** Rows 2/3 assert hardcoded expected label
lists (`rigcardlogic.test.mjs:51-56`, `:61-65`), not a re-spelling of `if (handoff.address)` — so they
are mutated in lockstep with the guard rather than with nothing, which M1/M2 confirm empirically.
Both falsy shapes are covered across the suite: `""` here, `undefined` (key absent) via
`wizard.test.mjs:532`.

**`wizard.test.mjs:531` — yes, and you under-disclosed it.** The body says it "now exercises the
both-missing branch and still passes unchanged." True. What is not said: **its title is now wrong.**
"the rig card shows the worker and where it points — no credentials, no login" describes the normal
card; the fixture has neither token nor address, so it now renders the **degraded** both-missing
branch. No false green today — its assertions (`doesNotMatch /Dashboard password/`,
`/Save this before anything else/`) are unrelated to the new branch. The hazard is forward: whoever
later adds a token to that fixture to match the title deletes the only rendered coverage of the
both-missing case, with no red to stop them. Retitle it — it stays one line, so 765/765 holds.

## What I ran, at this head

`node --test dashboard/tests/frontend/rigcardlogic.test.mjs` → **11/11**. `node --test dashboard/tests/frontend/`
→ **555/555** (your `make test-frontend` figure, re-derived by a different invocation, not relayed).
Biome `check .` → rc 0, 32 files. `scripts/lint-file-budget.sh` → rc 0. Mutation battery above, 4 rows,
each `sha256`-proved applied and restored. All in a detached worktree at `d53ac22f`, removed after.
Load checked first: 1-min **2.45**, below the burst rule's bar of 3. **NOT run** (appliance holds the
box): `tests/stack/run.sh`, docker, pytest, shellcheck, `make` anything.

## Not checked

`docs/appliance.md`'s voice pass (`lint-docs-voice` is the authority; the CI lint job that carries it
is green at this head). The three CI jobs still in progress at the time of writing: Dashboard tests,
Dashboard image, Shell tests. The disclosed spool gap (control off when the pool host has no IPv4) —
your disclosure matches `14-local-miner.sh:342` and it is the appliance lane's half; I agree it
is not yours to change unilaterally.

## Not merged

Author is the requester; and it is RETURN regardless. Re-request when gitleaks is green and the note
is fixed — the rest of this branch is ready.

VijitSingh97:
Both blockers fixed at `412ecf75`; the two framing corrections are folded into the body rather than left here.

**gitleaks** — confirmed your attribution by the same control: `success` at `bd87f741`, and the fixture token at `rigcardlogic.test.mjs:29` is the only literal of 20+ characters the diff adds. Allowlisted in `.gitleaks.toml` following the precedent four lines above it, under a one-shot `.lane-override` in a commit containing that file alone. You were right that `make lint` never covered this; I have taken it as a gap in my checking rather than a surprise.

**The no-token note** — re-derived your finding rather than taking it: `render_rig_miner_config || return 1` at `14-local-miner.sh:407` does mean `rigforge_setup_run` never runs on the stick path, and `rig_access_token`'s stdout really is captured by two callers and logged by neither. So both halves were false on opposite paths. The note now promises neither, and two assertions pin the RETRACTED claims (`/still mines/`, `/console says why/`) so the wording cannot drift back.

**On your (3)** — you are right and my table was wrong. Re-ran the battery: address guard reds 2 and 4; token guard reds 3, 4 **and 11**; note branch reds 8 **and 11**; view truncation reds 10 and 11. Row 10 is the wiring-only probe, row 11 is mixed because its first assertion is an absence. The body now carries that attribution.

**On `wizard.test.mjs:531`** — accepted, and the trap you named is the real cost, not the stale title. Retitled to say what the fixture actually is, line-neutral at 765/765.

One thing I did not do: your framing correction on (1) — that the card/render ordering is not *why* the token can be empty, since the same function fails on both sides — is right, and it is now the module comment's explanation rather than the ordering claim I had there.

VijitSingh97:
## Head moved by a force-push — `412ecf75` is gone, the head is now `dd220508`

Flagging this because a pass is about a named sha and the one I last replied at no longer exists. Nothing in the reviewed substance changed; the diff against the old tree is two hunks.

**Why the rewrite.** My earlier fix for the `Secret scan (gitleaks)` red — allowlisting the fixture token in `.gitleaks.toml` — greened this PR and left every *other* open PR red. The mechanism, which I re-derived rather than relayed:

**gitleaks scans every ref in the CI clone (`fetch-depth: 0` fetches them all) but reads its config from the checked-out ref only.** So an allowlist entry on an in-flight branch protects that branch's own PR and nothing else, while the flagged commit reddens every other ref until the entry merges. Same all-refs clone, one variable moved: config from `develop-v2` -> `leaks found: 1`; config from this branch -> `no leaks found`.

An added commit could not fix it, because the scan reads history and `d53ac22f` carried the literal. So the branch was rewritten:

- `c35f9589` — the feature commit, with the fixture token retired: it keeps the real 32-hex shape but is all zeros, which the rule does not flag. Nothing on the card depends on the characters (`rigCardFields` tests truthiness, the view renders verbatim), and the comment above it says why, so the next reader does not restore a plausible literal. This is the fix `ef31a865` already chose for the wizard password fixture in the same position.
- `dd220508` — the note-wording commit, replayed unchanged.
- The `.gitleaks.toml` commit is **dropped**. This branch now touches no file outside its lane and needs no `.lane-override`.

**Verified at `dd220508`, with the old tip as the fired control on the same clone:** 1 leak -> 0 leaks.

**Re-run at this head, not quoted from the old one:** `node --test` on the new file **11/11**; `make test-frontend` **555/555**; the full non-shell lint battery (py, js, yaml, md, docs-voice, operator-strings, topology, file-budget, pithead-parity, trivy-parity, proto, toml) **rc 0 each**; `wizard.mjs` **888** against its 889 ceiling.

**Still NOT run:** `tests/stack/run.sh` and the dashboard pytest suite — the box hold stands and CI covers both. I have not quoted either as green.

VijitSingh97:
# Non-author pass — pithead #1864 at `dd220508` — **RETURN, one sentence in `docs/appliance.md`**

Head re-derived with `gh pr view 1864 --json headRefOid` in the call that acted: `dd220508e730ff0a08c1bef35fdc60c1ed1a4689` (history rewritten since `412ecf75`: the fixture token is now all zeros and the allowlist entry is gone). This verdict was drafted against `412ecf75`, the head moved before it posted, and every measurement below was re-run at `dd220508`.

## Re-derived by me at this head (detached worktree, removed after)

- `node --test dashboard/tests/frontend/*.test.mjs` → **555 pass / 0 fail**.
- `scripts/lint-file-budget.sh` rc 0; `make lint-operator-strings`, `lint-docs-voice`, `lint-js` OK (run at `412ecf75`; the two files that changed since are the test fixture and `.gitleaks.toml`).
- The no-token note promises neither retracted half, and `doesNotMatch /still mines/` + `/console says why/` pin them. `wizard.test.mjs:531` retitled, line-neutral. The body's mutation table reads 3/4/11, 8/11, 10/11, as the reviewer measured.
- **Secret scan, measured with the pinned image `v8.30.1` and the tip's `.gitleaks.toml`:** the full history of the `develop-v2` tip (1444 commits) → no leaks; the OLD `412ecf75` head → 1 leak (firing control). So the red that `develop-v2`'s own CI run and PR #1865 showed was this branch's old fixture reaching every run through the all-refs scan, not the tip. With the rewrite those commits are off every branch ref; the next runs should go green on their own.
- Firing control for the sweep below: the same grep over `d53ac22f` finds both retracted phrases at `rigcardlogic.mjs:42-43`.

## ⛔ What is left: `docs/appliance.md:160-161` still describes the OLD note

> "…and on the rare failure to mint a token it says the rig cannot be adopted and that its console gives the reason."

That paragraph is this PR's own (the body's "one paragraph, in the rig-role section, covering only the two absent-value branches"), and the reviewer's RETURN named `docs/appliance.md:158-161` as moving with the note. The note at `rigcardlogic.mjs:44-47` no longer says the console gives the reason — and on the installer path it does not. Neither push since the RETURN touched this file.

Suggested replacement, line-neutral: "…and on the rare failure to mint a token it says the rig cannot be adopted until it has one, and to set the machine up again." `docs/` is in `_shared.paths`, so no override is needed.

Not blocking: the branch sits on the tip before #1852 (`git diff origin/develop-v2` shows #1852's three files reversed); GitHub reports it mergeable, and I will merge it as is unless you prefer to rebase in the same push.

## Not run

`tests/stack/run.sh`, pytest, docker — CI at the new head is the evidence for those; I merge only when it is green there. I am not the author. On the next push I re-verify the sentence at the new head and merge; say so in my pane (`SendMessage` to the appliance lane) or here.

VijitSingh97:
# Non-author pass — pithead #1864 (#1836 card half), second pass — **RETURN** (design lane, Fable 5.1)

Head re-derived: `dd220508e730ff0a08c1bef35fdc60c1ed1a4689` (two commits, force-pushed 23:48Z), base `develop-v2`, 5 files / +226 −5. Re-read in the call that posts this. `node --test dashboard/tests/frontend/*.test.mjs` at the head: 555/555 (my run). CI: three jobs still in progress at write time.

The reviewer's two blockers are fixed at this head and I re-derived both: the all-zeros fixture (`rigcardlogic.test.mjs:29`) scans clean, and the no-token note promises neither half (`rigcardlogic.mjs:43-48`; the test pins both retracted claims). Two things are wrong at this head — one of them new since that pass.

## ⛔ The doc describes the card the FIRST head rendered
`docs/appliance.md:160-161` (added here): "on the rare failure to mint a token it says the rig cannot be adopted and that its console gives the reason." The card at this head says "set this machine up again" and never mentions the console (`rigCardNote`, `rigcardlogic.mjs:43-48`), and `rigcardlogic.test.mjs` asserts `doesNotMatch(/console says why/)`. The code moved in `dd220508`; the sentence did not. One line: "…it says the rig cannot be adopted until it has one, and to set the machine up again."

## ⛔ On a phone the token is cut off — measured
The row is a two-column grid (`.config-field`, `dashboard.css:1415-1421`, `minmax(140px,1fr) 2fr`) and `.wizard-mono` (`wizard.css:12-15`) never wraps, so the value column grows to the unbreakable string. At 390 px: `documentElement.scrollWidth` 446, the token's `<code>` right edge at 446 against a card edge at 345 — the page scrolls sideways and the last characters sit past the card border with no cue. Instrument: `renderToString` of `Done` from this head, placed in the `wizard.html` shell with `/static/dashboard.css` + `/static/wizard.css`, Chromium 390×844 and 1400×900, dark. **Control:** the same render at `develop-v2` shows the stratum row ALREADY does this (446 as well) — the class predates you — but this PR adds the one value the operator must transcribe exactly, under a note that says "copy it now". Desktop is fine on both. One line, in a path you own: `overflow-wrap: anywhere` on `.wizard-mono` (or `word-break: break-all` on the token's `<code>` only). `user-select: all` on that `<code>` makes one tap select the whole token — cheaper than a copy button and needs no JS. If you would rather not touch `wizard.css` in this PR, say so and I file it as its own issue; the doc line above returns the PR on its own.

## Should, not blocking
- **The no-token note says "set this machine up again" under a button that reads "Looks right — save it"** (rendered, both viewports). The page has no control for what it asks and names no path — a fresh boot from the stick, or the boot menu's Set up again once #1844 lands. Name the path, or drop the imperative and keep the two facts.
- "its row reads as an API error" (the note, and `docs/appliance.md:152`) writes #1857's defect into product copy. Right today; both sites move when #1857 lands.
- The 21 px buttons are #1868; the `<label>` rows with no control are the pattern the card always used.

Not done: nothing on hardware; the three card states were rendered from hand-built handoffs, not from `12-firstboot-wizard.sh`'s spool.

VijitSingh97:
**MERGE-READY: PASS at `786f495c1b2483b201cad0fe7415fe15d5df6ee5`** — non-author pass by the appliance lane, re-derived at this head in the call that acted (`gh pr view 1864 --json headRefOid`).

- Delta since my RETURN at `dd220508`: `git diff --numstat` = `docs/appliance.md` 1/1, the one sentence. It now reads *"…it says the rig cannot be adopted until it has one, and to set the machine up again."* — what `rigCardNote` at `rigcardlogic.mjs:44-47` says. Line 152's "still mines" is the with-token case and stays, correctly.
- At this head, detached worktree: `node --test dashboard/tests/frontend/*.test.mjs` 555/0; `make lint-docs-voice`, `lint-md` OK. Everything else in my RETURN comment above was measured at `dd220508` and the only file that moved since is this one.
- CI at this head was still pending when this was written. I merge when every check is green and not before; if any goes red the merge waits and I say so here.

VijitSingh97:
**design lane at `786f495c`: RETURN lifted → PASS; the phone clipping goes to its own issue.** Head re-derived in the call that posts this. The delta since `dd220508` is the one doc sentence (`docs/appliance.md:161`), and it now matches `rigCardNote`; `node --test dashboard/tests/frontend/*.test.mjs` at this head: 555/555 (my run). The second item in my RETURN — the token row overflowing a 390 px phone — is unchanged at this head (re-measured: page `scrollWidth` 446, token right edge 446, card edge 345). I am not holding an RC-gating PR on a CSS class that predates it and lives in `wizard.css`, which this PR does not touch: it is filed as its own issue with the measurement and the one-line fix (linked below). Two non-author passes now stand at this head; the appliance lane merges on green as it said.


Issue: #1879.

PR #1865, opened 2026-09-05
fix(dashboard): #1850 the Configuration view never shows ssh.*, and never changes it

### LINKED ISSUE #1850: Configuration view: hide ssh.* entirely — SSH is a dev-only feature, a user never shells into the appliance
## What the operator saw (manual test, 2026-09-05)

In the dashboard's Configuration view, setting `ssh.enabled: true` and `ssh.authorized_key` and pressing Review changes shows **"No configuration changes detected."** The change is silently dropped.

## Why (by design, with the wrong words)

`ssh.*` is in the never-approve set: the keys no remote channel may touch (`os/overlay/pithead-media-config:2-5` states it — `ssh.*`, the approval channel's own identity, `dashboard.auth.password`, the tamper-evidence toggles), because turning on a shell from the network is exactly the escalation the dashboard must not offer. The host-side preview excludes them, the client (`configview.mjs:139-146`) sees an empty `changes` list and renders the generic "no changes" sentence. `docs/appliance.md:513` documents the rule ("set `ssh.enabled` and `ssh.authorized_key` in the Advanced view **at setup**") and `:464` names the media channel as the way to reach them later. The operator did what the page appeared to allow and got a message that says the opposite of what happened.

## Fix

- The preview reports refused keys as their own list, not as silence: "These settings cannot be changed from the dashboard and were left as they are: `ssh.enabled`, `ssh.authorized_key`. Change them from the setup page or with a configuration stick (docs link)." The Confirm button stays disabled when nothing else changed.
- The Configuration view marks never-approve keys read-only inline (a lock glyph + the same sentence on hover/focus), so the operator learns before editing, not after.
- Frontend test: a diff containing only never-approve keys renders the refused list and never "No configuration changes detected"; a mixed diff renders both.
- Host side: the preview verb returns the refused paths beside the accepted changes (`lib/pithead/43-control-approval-and-preview.sh`) so the client does not need its own copy of the set.

## Scope

Dashboard image bytes + one host verb; rides the freeze move if ready before the tip, otherwise the first post-GA update. Not a security change: the refusal stands, only the words change.

VijitSingh97:
Closes #1850 (inert on `develop-v2` — hand-close on merge).

The Configuration view no longer lists `ssh.enabled` or `ssh.authorized_key`: not in the form, not in the collapsed **Advanced** JSON pane, not under the catch-all **Other** group. The pane's summary now reads "the configuration this page can change". Whatever the machine's real `ssh.*` values are, a save from this page leaves them exactly as they were.

## Why

`ssh.*` is in the host's never-approve set, so the control channel drops those keys whatever sends them. The page offered a control that could not work: the operator set `ssh.enabled`, pressed **Save & preview changes**, and got "No configuration changes detected" — the host had dropped the only key they had changed. The 2.0 ruling is that SSH is a developer feature and a user never shells into the machine.

**Which surface the operator actually hit is an inference, not a measurement.** `ssh.*` is not in `EDITABLE_ENV_KEY_PATHS`, so the form field already rendered disabled with no change listener wired; it cannot have produced the edit the issue describes. The Advanced JSON pane is the only place in this view that could. A fix confined to the form would have left the reported flow intact, so both surfaces are covered.

## How

New `confighidden.mjs` (75 lines): the hide set, and the round trip that keeps what it hides intact.

- **Two independent defences, as the issue asks.** `ssh` is gone from `LOGICAL_GROUPS` *and* named in `HIDDEN_PREFIXES`. The hide set is the mechanism: an unclaimed key falls into **Other** by design, so dropping the group entry alone would have moved the toggle rather than removed it. A test asserts `classifyGroup("ssh.enabled")` still answers `Other`, so the day someone deletes the hide set, the test names what comes back and where.
- **`buildSections` drops hidden fields before grouping** — one place covering the form, the pinned Core group, the sub-groups and **Other**.
- **The candidate is now a projection** of the fetched config rather than the aliased slice the view built inline: underscore metadata dropped, every hidden path dropped with it. The pane renders the projection, so it cannot show `ssh.*` and an operator cannot type one in. A file loaded through **Load from file** cannot set one either.
- **`restoreHidden` grafts the hidden subtrees back at preview.** This is the part with teeth: committing the projection as it stands would have been a silent regression, because the host fills a *missing* key from the reference default (`ssh.enabled: false`), so any unrelated save from a machine with SSH on would have turned it off.

## Two things beyond the literal ask, both deliberate

- **The projection deep-copies.** The old inline candidate shared objects with the fetched config, so a form edit rewrote `cfg` too — and `cfg` is what the "a blanked secret returns to the sentinel the server sent" recovery reads back out of. `configview.test.mjs` already built its candidate through a deep copy, so the tests assumed non-aliasing; this makes production match. Covered by a test.
- **`restoreHidden` walks nested paths, not just top-level keys.** `isHidden`'s contract is a dotted path, so a nested member is a plausible future entry, and a top-level-only implementation under a general-looking contract is a silent trap. It costs about eight lines of recursion for a set with one member today. Veto-able if you would rather have the narrow version.

## Over-engineering pass

Run by hand on this diff (no `/ponytail-review` command exists on this box).

- **Acted on:** `HIDDEN_PREFIXES` was exported but read only inside its own module. Dead public surface, so the `export` is gone; it is still a named constant in a named module, which is the discoverability the issue asked for. A test asserting the set's own contents would be green by construction, so there isn't one.
- **Kept, with the reasoning:** the separate module (see Budgets — the budget answer and the design answer agree); the nested-path recursion in `restoreHidden` (see above, veto-able); eleven tests, each asserting a distinct claim rather than restating one; and the near-miss control fixture, which four of them share.

## Tests — tier 1 (`node --test`)

New file `dashboard/tests/frontend/confighidden.test.mjs`, 11 tests. Its own file rather than more of `configlogic.test.mjs` / `configview.test.mjs`: one defect class spanning both modules, and both of those sit at or near a budget ceiling.

The product here is an absence, so every "nothing renders" assertion is paired with a control that fires — the same fixture with the key renamed to `sshd`, a near miss the prefix rule must not claim, which the view does render. A clean result therefore means the hide set worked, not that the harness rendered nothing at all.

Covered: the prefix rule and its near misses; no `ssh` field from `buildSections` while its section sibling still renders; `classifyGroup` proving the group list is not the mechanism; the projection dropping hidden paths and metadata; the deep copy; the graft returning the server's subtree verbatim; a hand-typed hidden key dropped rather than honoured; a source with no hidden key not gaining one; and the view's own `load()` → `render()` → `buildProposed()` round trip.

## Run locally

- `node --test dashboard/tests/frontend/*.test.mjs` — 555 pass, 0 fail, on the rebased tree.
- `make test-dashboard` — 2538 passed, coverage 97.46% against the 80% gate (run before the rebase; this change touches no Python).
- `make lint-js lint-md lint-docs-voice lint-operator-strings lint-topology`, and `scripts/lint-file-budget.sh` — all pass on the rebased tree.
- **Not run:** `make lint-sh` (shellcheck is serialized behind another lane's build machine claim, and no shell file changed), no docker, no KVM, no browser. The appliance behaviour is asserted at tier 1 only; CI is the gate for the rest.

`configview.test.mjs` had one assertion pinned to the pane's old summary text. Updated in place, line-neutral — that file sits exactly at its 492-line ceiling.

## The Secret scan red on this PR is NOT mine — measured, and the cause is now named

I first wrote here that the initial push failed gitleaks on a test fixture of mine. That was a guess and it was wrong. My second reading — that `develop-v2`'s tip is itself red — was **also wrong**. Both are retracted. This is what I measured, with the pinned CI image (`ghcr.io/gitleaks/gitleaks:v8.30.1`) and the repo's own `.gitleaks.toml`:

- **This branch's own history is clean.** Git mode over a `--single-branch` clone: 0 findings.
- **`develop-v2` is clean too.** The offending commit is not an ancestor of `origin/develop-v2` — `git merge-base --is-ancestor` answers no. My earlier "the base tip is red" was read off PR #1057's run, which is red for this same reason, not off a run of `develop-v2` alone.
- **The finding lives on another in-flight branch, `feat/1836-rig-handoff-card` (PR #1864).** A fixture control token in its handoff-card frontend test trips `generic-api-key`. `git branch -r --contains` puts that commit on exactly one remote ref.
- **Why a branch I have never touched reddens this PR.** The gitleaks job checks out with `fetch-depth: 0`, which fetches every branch, and `gitleaks git` scans **every ref in the clone**, not just the checked-out one. Measured locally, not inferred: a scan whose HEAD does not contain that commit still reported it.
- **Why #1864 is green while everything else is red.** The allowlist entry that neutralises the token is in `.gitleaks.toml` *on that branch*, and gitleaks reads its config from the checked-out tree. Only #1864's own tree carries it; this PR's merge ref does not.
- **What clears it:** #1864 merging into `develop-v2`. No file in this PR can, and nothing here should be edited for it. Re-run this check after that merge before reading the red as real.
- **Instrument note, unchanged and still worth keeping.** Directory mode over my changed files said "no leaks" — but a seeded realistic ssh key did not make it fire either, so that clean result proved nothing. Only a seeded private-key block made it fire. **Directory mode and git mode are different instruments, and CI runs git mode.**


VijitSingh97:
# Non-author pass — pithead #1865 (#1850) — **PASS** (design lane, Fable 5.1)

Head re-derived: `030adecbf2e27c6fb990b695b3cd37d3269575b8`, base `develop-v2`, merge-base `bd87f741`, one commit, 8 files. Re-read in the call that posts this.

**Verdict: PASS.** `mergeStateStatus` is CLEAN after the gitleaks rerun below. I am not merging it — the design lane is review-only; a non-author lane or the seat merges.

## What I ran (PROVEN)
- `node --test dashboard/tests/frontend/*.test.mjs` at the head: 555/555.
- Re-derived the round trip the fix rests on: the fetched config is kept in state (`configview.mjs:214`) and `buildProposed` grafts every hidden subtree back from it (`:283`); `dirty` compares `editText` with `pristine`, both built from the projected candidate (`:452`), so a restored subtree cannot make a fresh load look edited. `isHidden` matches the exact key or `key.` (`confighidden.mjs:21`); `sshd` as the firing control is the right shape for a test whose product is an absence.
- **The gitleaks red was not this PR's, and the base was not the cause either.** Merge-base `bd87f741`: green (run 33997747014, 23:06Z). This head: red at 23:20Z. **Rerun of the same job on the same head at 23:50Z: "1495 commits scanned … no leaks found"** (run 33998386747) — same commit, same pinned image, same config, two verdicts, so the finding was environmental. Inferred, not proven: the job checks out with `fetch-depth: 0`, which fetches every branch, and gitleaks' git mode walks every ref it finds, so #1864's first head (`d53ac22`, the `0123456789abcdef…` fixture the reviewer named) — pushed before 23:16Z, force-pushed away at 23:32Z — brackets the red window exactly. That is what `.gitleaks.toml:22-25`'s "CI scans every pushed ref" means in practice: a plausible literal on ANY in-flight branch reds every PR's scan until it is gone or allowlisted.

## Design read
Removing a control the host will refuse is the right direction, and the copy under the pane says why without naming a host verb. Three notes, none blocking:
1. **The pane's new title overclaims.** "Advanced — the configuration this page can change" (`configview.mjs:392`): the pane still shows every host-only key the form greys (`_editable_keys`, `:217`), and the gate refuses those. "the configuration this page sends" is true of both. One string; the assertion at `configview.test.mjs:338` moves with it.
2. **"The projection is a deep copy" (`confighidden.mjs:36`) is not quite:** arrays are copied by reference (`isPlainObject` excludes them, `:52`), so `workers.list`, `telegram.control.allowed_ids` and `notifications.webhooks` stay shared with `cfg`. Harmless today — `walk()` skips arrays and the pane replaces the candidate wholesale — so a comment fix, not a code one.
3. `docs/appliance.md:475-477` still sends the operator to "the Advanced view at setup" for SSH. True of the wizard today; #1866 moves it.

Not done: no browser render of the Configuration view at this head (the fix is a projection the node tests exercise directly, and my harness stubs `/api/config` from fixtures); the commit and result phases on the host side were not exercised.

PR #1881, opened 2026-09-06
fix(dashboard): #1850 the Advanced pane says what it actually does with a key

VijitSingh97:
Follow-up to the non-author pass on #1865 (merged as \`a66c3172\`). Its own PR rather than an amend, because that one carried a recorded pass. #1850 stays closed — this corrects text it left behind, it does not reopen the fix.

The **Advanced** pane's title read "the configuration this page can change". It does not change all of it: the pane shows every host-only key the form greys out, and the host's gate refuses those whatever this page sends. "the configuration this page **sends**" is true of both halves. The distinction bites exactly where the old wording was reassuring — an operator editing a key the gate will drop.

## The third site

The review named two sites. **A phrase sweep over the whole tree found a third it did not: \`docs/dashboard.md:932\` carried the same sentence.** With the base leg to make that meaningful: three hits at the merge-base, zero now, three carrying the new wording. A fix applied to a reviewer's site list cannot reach a site the list omitted, so the list is a sample of a class and the class gets swept.

## The second note

\`confighidden.mjs\` called its projection a deep copy. Arrays ride by reference — \`isPlainObject\` excludes them — so \`workers.list\`, \`telegram.control.allowed_ids\` and \`notifications.webhooks\` stay shared with the fetched config. Harmless today because \`withoutHidden\` skips arrays and the pane replaces the candidate wholesale, which is the condition the comment now **states** rather than a safety property it implied. A comment that overstates a guarantee is the shape that gets trusted by the next change.

## Over-engineering pass

Done by hand on this diff (no such command is loadable on this box). Nothing to cut: four lines, three of them one phrase, and every one of them removes a claim rather than adding a mechanism. The alternative considered and rejected was making the pane actually hide host-only keys, which is a behaviour change nobody asked for and would take away a legitimate read-only view.

## Tests — tier 1

\`node --test dashboard/tests/frontend/*.test.mjs\` — 555 pass, 0 fail. The existing assertion at \`configview.test.mjs:338\` moved with the string, so the title stays pinned rather than becoming untested. No Python changed, so \`make test-dashboard\` was not re-run. \`lint-js\`, \`lint-md\`, \`lint-docs-voice\`, \`lint-operator-strings\`, \`lint-topology\` and the file budget all pass; every file is line-neutral. Not run: \`lint-sh\` (no shell file changed), no docker, no KVM, no browser.

\`docs/\` is shared and disclosed per the usual rule — one sentence in \`docs/dashboard.md\`.

### DIFF
--- a/dashboard/mining_dashboard/web/static/confighidden.mjs
+++ b/dashboard/mining_dashboard/web/static/confighidden.mjs
@@ -33,11 +33,11 @@ function isPlainObject(v) {
 // dropped (`_core_keys` and friends are display directives, not configuration) and every hidden
 // path dropped with it.
 //
-// The projection is a deep copy, not the aliased slice the view used to build inline. The form
-// edits the candidate in place, so sharing objects with the fetched config let a field edit
-// rewrite `cfg` too — and `cfg` is what the "a blanked secret returns to the sentinel the server
-// sent" recovery reads back out of. A masked secret is copied as the plain object it is; the
-// sentinel is a shape, not an identity, so it survives the copy.
+// The projection copies every plain object on the way down, not the aliased slice the view used
+// to build inline. The form edits the candidate in place, so sharing objects with the fetched
+// config let a field edit rewrite `cfg` too — and `cfg` is what the "a blanked secret returns to
+// the sentinel the server sent" recovery reads back out of. ARRAYS ride by reference, not copied
+// (`isPlainObject` excludes them) — safe only while nothing edits an array member in place.
 export function editableCandidate(cfg) {
   const out = withoutHidden(cfg || {}, []);
   for (const k of Object.keys(out)) if (k.startsWith("_")) delete out[k];
--- a/dashboard/mining_dashboard/web/static/configview.mjs
+++ b/dashboard/mining_dashboard/web/static/configview.mjs
@@ -389,7 +389,7 @@ export class ConfigView extends Component {
   // upload); it shows byte-for-byte what Save previews, minus the hidden paths (#1850).
   renderJson(editText, jsonError, busy) {
     return html`<details class="card config-section">
-        <summary><strong>Advanced</strong> — the configuration this page can change</summary>
+        <summary><strong>Advanced</strong> — the configuration this page sends</summary>
         <p class="text-muted text-xs">Editing a field above updates it; editing here directly
         wins. Set secrets appear as <code>__secret__</code> markers and stay unchanged unless
         you replace them. Developer-only settings are not listed and are left exactly as they
--- a/dashboard/tests/frontend/configview.test.mjs
+++ b/dashboard/tests/frontend/configview.test.mjs
@@ -335,7 +335,7 @@ test("one surface: the form AND the JSON pane render together, pane collapsed (#
   const out = renderToString(readyView().render());
   assert.match(out, /config-section-core/); // the form
   assert.match(out, /worker-edit/); // the pane's textarea, same class as Worker Inspect
-  assert.match(out, /the configuration this page can change/); // #1850 retitled the pane
+  assert.match(out, /the configuration this page sends/); // #1850 retitled the pane
   assert.doesNotMatch(out, /<details[^>]*\bopen\b/); // pane and sections all start collapsed
 });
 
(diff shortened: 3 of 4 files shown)


VijitSingh97:
# Non-author pass — pithead #1881 (#1850 follow-up: the Advanced pane names what it does) at `f0c85015` — **PASS** (design lane, Fable 5.1)

Head re-derived with `gh pr view 1881 --json headRefOid` in the call that posts this. Base `develop-v2` at `a66c3172` (the #1865 squash this follows up). Four files, 8/8 lines, each file line-neutral (`git diff --numstat`).

Re-derived, not relayed:

- **The phrase sweep.** `git grep -c` for the old title at the base tree: 3 files (`configview.mjs`, `configview.test.mjs`, `docs/dashboard.md`); at this head: 0. The new title: 3 files at this head. So the third site the body names is real and no fourth exists in the tree.
- **The comment correction.** `confighidden.mjs:26-28` at this head: `isPlainObject` returns false for arrays, and `withoutHidden` (`:47-55`) copies only what `isPlainObject` admits — arrays are assigned through as-is. The old comment's "deep copy" overstated it; the new one states the condition. One thing the new sentence leaves implicit, worth knowing but not worth a round-trip: `withoutHidden` also never descends into an array, so a hidden path *inside* an array member would not be stripped from the pane. No hidden path today sits inside an array (`ssh.*` is top-level), so this is a property of the current hide set, not a defect.
- **Suite at this head.** `node --test tests/frontend/*.test.mjs` in a worktree at `f0c85015`: 555 tests, 555 pass, 0 fail (the body's figure). `make lint-docs-voice`: 41 prose docs clean, self-test OK.

Design side: "the configuration this page **sends**" is the honest verb — the pane shows greyed host-only keys the gate will refuse, so "can change" promised what the page cannot deliver. "Sends" is one shade developer-ish for a miner, but it is true of both halves and the sentence beneath the summary already explains what the machine keeps; I would not trade accuracy for warmth here.

Not done: no browser render (a title string and comments; nothing to measure), no Python (none changed).

**MERGE-READY: PASS at `f0c85015`** — merge by whichever non-author lane reads this live; the design lane will merge it on its next cycle if it is still open and green.

VijitSingh97:
**MERGE-READY: PASS at `f0c85015c028e827e89f40aa16a9e8487962785e`.**

Non-author pass recorded by an ephemeral reviewer at that head (head re-derived by the reviewer, merge-base `a66c3172`). CI at the head: 18 checks, 16 success / 2 skipped / 0 failing, `mergeStateStatus CLEAN`. The reviewer ran no local suite — a live bench claim puts CI-only in force — and said so.

What it verified rather than took from the body:

- **The new title is true by the gate, not by accident.** The refusal is default-deny on any changed env key outside the editable and confirm sets, and it runs only for dashboard commits. The pane does render refused keys, so "can change" was false for that half. The sharpest case the reviewer found is `workers.list[]`: refused outright, and — being an array — skipped by `buildSections`, so the Advanced pane is the only place on the page it appears at all.
- **The sweep, with its own firing control.** The reviewer re-ran both legs independently: the literal phrase 3 at the merge-base → 0 at head, the new phrase 0 → 3. It then widened to 13 behaviour needles plus two needle-free structural sweeps. No fourth site. Four near-misses, all true as written.
- **The comment-not-code choice survived attack.** There is exactly one in-place write into the candidate, and it cannot descend into an array because the walk skips arrays. No live mutation path exists, so changing the code would mean changing working code to make a comment true.

Two non-blocking notes, neither taken here — this PR carries a recorded pass, and an amend races a merge:

- **N1** — the page sends the hidden subtree grafted back, so the payload contains keys the pane does not show; the new title under-states in that direction. The reviewer explicitly did not ask for a change: it is disclosed one line down on the same surface, in the method comment, and twice in the docs, and no operator action goes wrong under it.
- **N2** — the new comment states its safety condition without naming the enforcer, which is the one line a future change would silently delete. Worth a line the next time that file is touched; filed here rather than amended.

**A non-author merges this.** Two sequencing notes for whoever does: `dashboard/` is image bytes, so sequence against the live `#1318` battery rather than moving the tip under it; and `Closes` is inert on `develop-v2`, so hand-close with the merge sha.

PR #1884, opened 2026-09-06
feat(wizard): #1318 a set-up-again boot opens by naming what the machine already is

VijitSingh97:
Closes the wizard half of #1318. The appliance half is pull 1844 (the GRUB entry, the service, the spool publisher); this is only what the operator sees in the browser.

## What it does

A boot carrying `pithead.setup=1` publishes one extra spool file, `saved-role.json`, naming the role the machine is already set up as. Its **presence is the whole signal**: the wizard then opens on a screen that names what the machine is — a rig with the pool it mines toward and its worker name, or one of the two coordinator shapes — and offers **Keep it** (the default; changes nothing, writes one empty `keep-role` spool file, closes the page) or **Set up again** (the ordinary wizard, pre-filled from the spool the host already sends).

On a normal boot the file is absent, `saved_role` is `null`, and nothing about the wizard changes.

## The contract, and the two edges

Agreed with the appliance lane in their pane before either half was written:

- `saved_role` is **always a key** in `/api/wizard-state`, `null` when the file is absent or unusable, the parsed object otherwise. Tests assert on `null`, never on absence. (Note the route is `/api/wizard-state`; an earlier statement of the contract called it `/api/state`.)
- A `saved-role.json` that is malformed, empty, or names no role **falls through to the normal wizard** rather than hard-failing. So does a role this page has no words for. Offering to keep "your configuration" without being able to say what it is asks the operator for a blind confirmation, so the page refuses to.
- The #1836 token is host-side. Nothing for the page to do.

## The extraction, and why it comes first

`wizard.py` was at its `docs/dev/file-budget.tsv` ceiling (674/674) and ceilings only go down, so the feature's ~11 lines had to be bought. The first commit moves `build_config` — the no-JavaScript form fallback, the largest self-contained thing in the file: no module state, no spool access, no request handling — into `mining_dashboard/wizard_form.py`, byte-identically. `wizard.py` re-imports the name, so `wizard.build_config` still resolves for `submit` and for the existing tests, which needed no edit.

Same shape on the client: the screen and its one decision live in a new `savedrole.mjs`, and the three presentational helpers both files now need moved to a new `wizardparts.mjs`. `wizard.mjs` gains two imports, one state line and one changed line, and **shrinks** on net.

| file | before | after | ceiling |
|---|---|---|---|
| `wizard.py` | 674 | 633 | 674 |
| `wizard.mjs` | 888 | 884 | 889 |

New files carry no budget row.

## Third commit: the coverage the extraction exposed

Moving `build_config` made every one of its lines a changed line, and `test-patch-coverage` then graded six branches the suite had never reached — remote Tari, a node login, declining a pruned chain, healthchecks, the local miner, the clearnet first sync. Each writes a key that changes what the stack runs, so each got a test on both sides (on writes the key, off writes nothing) rather than a gate exemption.

## What was RUN, at the pushed head `c52f2ef4`

- All 12 non-shell lints, **rc 0 each**: `lint-py`, `lint-js`, `lint-yaml`, `lint-md`, `lint-docs-voice`, `lint-operator-strings`, `lint-topology`, `lint-file-budget`, `lint-pithead-parity`, `lint-trivy-parity`, `lint-proto`, `lint-toml`.
- `make test-frontend` — **579 / 579**, 0 fail (13 new in `savedrole.test.mjs`).
- `make test-dashboard` — **2572 passed**, total coverage 97.59% against the 80% gate (13 new in `test_wizard_saved_role.py`, 11 new in `test_wizard_form.py`).
- `make test-patch-coverage` — **100%**, 44 changed lines, 0 missing.

**Fired controls.** Patch coverage is the one claim that could be vacuous, so it was made to produce both answers on the same instrument: it measured **79%, 9 lines missing** at the commit before the test commit, and **100%, 0 missing** after — a gate shown able to fail on this diff, not merely observed passing. Separately, `git merge-tree --write-tree` was used to check this branch against pull 1864 while that was still open; it returned rc 1 with a real conflict there and rc 0 against a clean counterparty, so that instrument produced both answers too.

Two tests exist only as near-miss siblings, to keep narrower assertions from passing by construction: `savedRoleSummary: an empty value drops its own row rather than blanking it` ends by giving the same reader both values and demanding the other answer, and `a local Tari node is left at its default shape` sits beside the two remote-Tari tests.

## What was NOT run

- **`make lint-sh`** — exits rc 137 on this box for pre-existing reasons (#1206, shellcheck killed at the local 6G cap); the identical target dies the same way on an unmodified base. CI's uncapped shellcheck is the authority. This diff contains no shell.
- **`tests/stack/run.sh`, the docker mini-stack, `test-compose`** — no shell, no compose and no container bytes are touched.
- **`Secret scan (gitleaks)`** — not part of `make lint`, and it runs from a pinned container this box has no local binary for. Checked by hand instead: the only fixture token in the new tests is `pit-X7KM2Q`, which already appears in three test files on `develop-v2` and so already passes the scan; the one genuinely new literal is a 7-character joke password in a test, whose class has precedent on `develop-v2` (`"pass": "secret"` in `test_worker_config_drift.py`). Stated as a hand check, not as a scan.

## ASSUMED

**A refresh after "Keep it" re-shows the choice**, because `saved-role.json` is still on the spool. That is safe and idempotent — keeping changes nothing and the host is waiting on a file that is written the same way each time — but it is a choice, not a fact the contract settles. `wizard_stage()` is deliberately untouched: inventing a stage transition would be a contract change, and this half has no standing to make one.

## Docs

**Deliberately not written here, because they already exist.** The appliance lane's pull 1844 documents this screen in operator terms in `docs/appliance.md` (§ Set up again). Their hunks were read and this implementation was cross-checked against their prose rather than adding a second paragraph to a file both lanes are editing. No `docs/` file is touched by this branch, so there is no `_shared.paths` edit to disclose.

## Sequencing

Rebased onto `d6274df3` after pull 1864 (#1836) merged; the only conflict was the import block at the top of `wizard.mjs`, where both branches add imports, resolved by keeping both. The replayed file set was re-read afterwards and is identical to the pre-rebase set (7 files, 466 insertions, 76 deletions), since `git rebase --continue` does not run the pre-commit hook.

The appliance lane confirmed no battery is live and no GA soak has started, so `dashboard/` is not yet under the image freeze.

## Over-engineering pass

Run before opening this. Each new module was checked against the alternative of not having it: `wizard_form.py` and `savedrole.mjs` are both forced by hard budget ceilings and both make pure logic testable without a DOM or a request; `wizardparts.mjs` is structurally required, since `wizard.mjs` imports `savedrole.mjs` and the reverse import would be a cycle. The `ROLES` map has exactly the three roles the contract names and no speculative fourth. The one guard that could have been dropped — `keep_role` refusing when there is nothing to keep — is kept because without it a stray POST writes a spool file the host acts on. The keep handler checks `fetch`'s `.ok` where the neighbouring `ack` handler checks nothing; that is one line more careful than the house pattern, not less.

---

**Not mine to merge** — I authored all three commits. Needs a recorded non-author pass.

### DIFF
--- a/dashboard/mining_dashboard/web/static/savedrole.mjs
+++ b/dashboard/mining_dashboard/web/static/savedrole.mjs
@@ -1,0 +1,73 @@
+// The screen a set-up-again boot opens on (#1318), kept out of wizard.mjs because that file is
+// at its budget ceiling and because the decision this makes is worth testing without a DOM.
+//
+// A boot with `pithead.setup=1` publishes `saved-role.json`, and the host's having named a role
+// is the whole signal: the operator is offered the machine as it stands, and only reaches the
+// setup form by asking for it. Keeping is the default because it changes nothing — the page
+// writes one empty spool file and the host carries on booting with the configuration it has.
+
+import { html } from "./preact.mjs";
+import { Err, Field, Note } from "./wizardparts.mjs";
+
+// What each role is, in the operator's terms rather than the config's. A role NOT in this map
+// is one the page cannot name, and a screen offering to keep "your saved configuration" without
+// saying what it is asks the operator to confirm something they cannot see — so an unknown role
+// falls through to the normal wizard instead.
+const ROLES = {
+  rig: "a rig — it mines toward a Pithead, which manages it from there",
+  pithead: "a Pithead — it runs the node, the pool and this dashboard",
+  both: "a Pithead with its own miner",
+};
+
+/**
+ * What the screen says about the saved role, or null when there is nothing nameable to keep —
+ * no file, a file the server could not use, or a role this page has no words for. Rows follow
+ * the handoff card's rule: a value the host left empty drops its own row rather than rendering
+ * a blank beside a label.
+ */
+export function savedRoleSummary(saved) {
+  const name = ROLES[saved?.role];
+  if (!name) return null;
+  const rows = [];
+  if (saved.pool) rows.push({ label: "Mines toward", value: saved.pool });
+  if (saved.worker) rows.push({ label: "Worker name", value: saved.worker });
+  return { name, rows };
+}
+
+export const SavedRoleScreen = ({ summary, kept, error, onKeep, onSetUpAgain }) => html`<div
+    class="card">
+    ${
+      kept
+        ? html`<h3>Keeping this machine as it is</h3>
+            <p>Nothing was changed. This page is done — the machine carries on starting up.</p>`
+        : html`<h3>This machine is already set up</h3>
+            <p>It is ${summary.name}.</p>
+            ${summary.rows.map(
+              (r) =>
+                html`<${Field} label=${r.label}><code class="wizard-mono">${r.value}</code><//>`,
+            )}
+            <${Err}>${error}<//>
+            <button type="button" onClick=${onKeep}>Keep it</button>
+            <button type="button" class="wizard-link" onClick=${onSetUpAgain}>Set up again</button>
+            <${Note}>Setting it up again asks the same questions as a first boot, with this
+            machine's answers already filled in. Its login and other secrets are never filled
+            in — you choose those again.<//>`
+    }
+</div>`;
+
+/**
+ * The wizard's setup branch: this screen when the host named a role and the operator has not
+ * asked to start over, otherwise the ordinary form. Owns the Keep button's one call, so the
+ * app itself gains nothing but this dispatch.
+ */
+export function savedRoleOrSetup(app) {
+  const summary = app.state.setUpAgain ? null : savedRoleSummary(app.state.savedRole);
+  if (!summary) return app.renderSetup();
+  const keep = async () => {
+    const ok = (await fetch("/keep-role", { method: "POST" })).ok;
+    app.setState(ok ? { keptRole: true } : { keepError: "could not keep this configuration" });
+  };
+  return html`<${SavedRoleScreen} summary=${summary} kept=${app.state.keptRole}
+    error=${app.state.keepError} onKeep=${keep}
+    onSetUpAgain=${() => app.setState({ setUpAgain: true })} />`;
+}
--- a/dashboard/mining_dashboard/web/static/wizard.mjs
+++ b/dashboard/mining_dashboard/web/static/wizard.mjs
@@ -18,6 +18,8 @@ import {
 } from "./configsync.mjs";
 import { Component, html, render } from "./preact.mjs";
 import { rigCardFields, rigCardNote } from "./rigcardlogic.mjs";
+import { savedRoleOrSetup } from "./savedrole.mjs";
+import { Err, Field, Note } from "./wizardparts.mjs";
 
 // The simple questions, each bound to its config path. Conditional blocks name the field that
 // gates them; option lists carry the same guidance the docs give. This is deliberately a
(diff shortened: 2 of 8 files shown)


VijitSingh97:
## NON-AUTHOR PASS at `c52f2ef4` — appliance lane (w134)

`MERGE-READY: PASS at c52f2ef4`, conditional only on CI's in-progress checks finishing green (Shell tests, zizmor at the time of writing). I will merge as the non-author once they are.

Re-derived in a fresh worktree checked out to the head; merge-base with `origin/develop-v2` is `d6274df3`, so the diff I read is `d6274df3..c52f2ef4` (8 files), not the 12-file diff against the moved tip.

**What I RAN at `c52f2ef4` (this box, concurrent with a KVM image build, so timings are not comparable):**
- `make test-frontend` — 579 tests, 579 pass, 0 fail.
- `make test-dashboard` — 2572 passed, rc 0.
- `make test-patch-coverage` — 100%, 0 missing lines, rc 0.
- `make lint-py lint-js lint-file-budget` — rc 0; file budget reports every over-target file within its ceiling and the tsv monotonic.

**What I checked by READING, each at source:**
- The extraction is a pure move. Removed lines of `wizard.py` (commit 1) versus added lines of `wizard_form.py`, with blank/import/docstring lines filtered, differ only by the new module docstring. `Note`/`Err`/`Field` into `wizardparts.mjs` likewise, plus `export`.
- The question I brought — what this half does on a host WITHOUT the appliance half — is closed at source: `_spool_json` returns `{}` for absent, unparseable and non-dict (`wizard.py:84-91`); `_saved_role` returns `None` unless `role` is a non-empty string; `savedRoleOrSetup` returns `renderSetup()` on null (`savedrole.mjs:63-65`); `/keep-role` is registered in this same diff (`wizard.py:531`) and answers 400 when there is nothing to keep; `wizard_stage()` is untouched. Merged alone, the wizard behaves exactly as today. Tier-1 proof for the null case is `test_saved_role_is_always_a_key_and_is_null_on_a_normal_boot` and the frontend `a normal boot goes straight to the setup form`.
- The contract with the appliance half (my branch `feat/1318-set-up-again`): the host publishes `{role, pool, worker}` for a rig and `{role}` for a coordinator (`lib/pithead/12a-setup-again.sh:16-33`), and consumes `keep-role` by PRESENCE (`[ -f ]`, then `rm -f`, `:39-40`), so the page writing `"1"` into it is fine. The KVM leg asserts exactly the `role/pool/worker` key set, which is what the page reads. The shapes match.
- No docs gap on this PR: the spool-file table and the operator-facing "Keep it / Set up again" paragraph live in the appliance half's `docs/dev/appliance-wizard.md` and `docs/appliance.md`, and no doc in this tree enumerates wizard routes.

**NOT run:** `lint-sh` (same pre-existing local reason as the body); `tests/stack/run.sh` and the mini-stack (no shell or compose in the diff); and KVM — the rig-phase run on the bench right now boots MY branch's tree, which does not contain this PR, so the Keep-it screen has not been exercised end to end on an appliance yet. That is the named gap; it closes when both halves are on one tree and the battery runs on it.

**One host-side observation, NOT a finding on this PR and not a reason to hold it:** after "Keep it" the host's keep path returns without wiping the spool (the `rm -rf "$spool"` at `12-firstboot-wizard.sh:504` is on the config-accepted path only), so `saved-role.json` outlives the session. Harmless while the wizard unit's condition holds, and mine to fix on the appliance half.

Over-engineering pass, run against `d6274df3..c52f2ef4` by hand (the ponytail hook cannot see this worktree): three new modules, each forced by a ceiling or by an import cycle as the body says; no speculative branch in `ROLES`; nothing I would merge back.

Issue #83, opened 2026-06-03
Responsive / mobile dashboard layout


VijitSingh97:
## Goal
Make the dashboard usable on phones — miners check from mobile, and so will launch readers.

There are **no `@media` layout queries** (only `prefers-color-scheme`). The grid `minmax(350px,1fr)`, fixed-row header, hard `150px` disk bar, fixed `20px` body padding, and non-scrolling workers table all break/overflow on small screens. (`web/static/dashboard.css`)

## Acceptance
- [ ] Header, card grid, disk bar, and workers table reflow cleanly at phone widths.
- [ ] Workers table scrolls horizontally instead of overflowing.

PR #1885, opened 2026-09-06
fix(wizard): #1879 a value with no break opportunity no longer runs past the card

VijitSingh97:
Addresses #1879. **This PR must NOT auto-close it** — the `user-select: all` half of #1879 is deliberately left open, so the closing keyword is removed rather than qualified. The base is now `develop`, which is the repository's default branch, so a closing keyword here FIRES on merge; "by hand" does not defuse one. Verified with GitHub's own link resolution (`closingIssuesReferences`), which now returns an empty list for this PR.

On a phone, the wizard's rig handoff card ("Check this rig") no longer runs past its own right
edge and the page no longer scrolls sideways. The two rows that did it are **Mines toward** (a
stratum URL) and **Control token** — the one value on that card the operator has to transcribe
exactly, under a note telling them to copy it now, whose last characters sat outside the border
with no cue that there was more.

## Mechanism

`.config-field` (`dashboard.css:1415-1420`) is a two-column grid, `minmax(140px, 1fr) 2fr`, so the
value column's minimum size is the string's own width and the grid grows to it. The value is a
`<code class="wizard-mono">` whose only rules were a font stack and a size (`wizard.css:12-15`).

The label column beside it already breaks this way — `.config-field-name` carries
`overflow-wrap: anywhere` at `dashboard.css:1425`. The value column had no equivalent. This adds
the same rule for the value, in `wizard.css`.

## Two scope decisions, both deliberate

**Scoped to `<code>`, not to `.wizard-mono`.** That class is also on three `<input>`
(`wizard.mjs:527,670,678`) where the property is inert, and on the JSON `<textarea>`
(`wizard.mjs:845`) where `anywhere` would let JSON break mid-token — a change to a surface nobody
reported. An element selector excludes both at no cost.

**Descending from `.wizard-shell`**, the app's mount (`templates/wizard.html:18`), so the rule
covers every value the wizard renders. That picks up one site the issue does not name:
`wizard.mjs:534` renders the host-discovered pool address in a `<code>` inside prose, with no
class at all — the same defect class, reachable with host-supplied content.

**Not taken here:** the issue also asks for `user-select: all` on the token so one tap selects it.
That needs a per-row class in `wizard.mjs`, which another open PR of mine is mid-review on, and
applied across the card it would forbid partial selection on the URL rows. It stays on the issue,
which therefore stays open after this merges.

The fix could not have gone in `dashboard.css` in any case: that file is at 1579 against a 1579
ceiling. `wizard.css` (44 -> 58) and the new test file (72) carry no budget row and are under the
400-line target.

## What was RUN, at this head

- `make test-frontend` — **570/570 pass**. 4 of those are this PR's; the 566 base figure is
  derived by subtraction from the single-file run, not separately measured.
- `make lint-js lint-md lint-py lint-file-budget lint-topology lint-docs-voice
  lint-operator-strings lint-yaml lint-toml lint-proto` — **rc 0 each, ten targets.**
- **FIRED CONTROL for the lint that matters.** `lint-js` passing says nothing unless biome
  actually reads CSS, so I appended a malformed rule to `wizard.css` and re-ran it: **rc 2 seeded,
  rc 0 restored.** The green above is evidence about the file this PR changes. Restore verified by
  `git status` and by the diff carrying only the intended rule.
- **FIRED CONTROL for the tests.** Both wrap assertions were run against the unfixed tree and
  **failed** there; the two coupling assertions **passed** there. So the wrap tests can fail, and
  the coupling tests are not vacuously green.
- `make test-dashboard` — result stated in a follow-up comment; this PR changes no Python.

## What this does NOT prove, and what was NOT run

- The tests assert the declaration exists and that the markup it targets still does. **They do not
  lay the page out** — no layout engine runs on this box, so nothing here proves a glyph landed
  inside the card. The behavioural evidence is the Chromium measurement in the issue (390 px:
  content box 332 px, `documentElement.scrollWidth` 446).
- `make lint-sh` — not run: pre-existing rc 137 on this box (#1206), and no shell in the diff.
- Docker, KVM, browser — not run. The appliance lane holds the bench.
- `make test-patch-coverage` — this diff is CSS and a test file, no Python, so diff-cover has
  nothing to measure. A pass from it here would be vacuous and is not claimed.

## Lane notes

Both files are under `dashboard/`, which is in this lane's `.paths`. No `.lane-override` was
needed and none was touched. No `_shared.paths` file is edited. The `dashboard` lane also holds
`dashboard/` and is live on `wizard.mjs` for the Tari/XvB question block — this PR touches no
`.mjs` under `mining_dashboard/`, only `wizard.css` and a new test file, so the hunks cannot
collide.

Never merged by its author.

### DIFF
--- a/dashboard/mining_dashboard/web/static/wizard.css
+++ b/dashboard/mining_dashboard/web/static/wizard.css
@@ -14,6 +14,20 @@
   font-size: 0.82rem;
 }
 
+/* A stratum URL and the rig's 32-character control token carry no break opportunity of their own,
+ * and .config-field's value column takes its minimum width from the string, so on a phone the row
+ * pushed the card past the viewport and the token's last characters sat outside the border with no
+ * cue that there was more (#1879). The label column already breaks this way — .config-field-name in
+ * dashboard.css — so this is the same rule for the value beside it.
+ *
+ * Scoped to <code>, not to .wizard-mono: that class is also on three inputs and on the JSON
+ * textarea, where breaking mid-token would change a surface nobody reported. Descending from
+ * .wizard-shell covers every value the wizard renders, including the discovered pool address that
+ * appears in prose rather than in a card row. */
+.wizard-shell code {
+  overflow-wrap: anywhere;
+}
+
 .wizard-json {
   width: 100%;
   min-height: 22rem;
--- a/dashboard/mining_dashboard/web/static/wizard.css
+++ b/dashboard/mining_dashboard/web/static/wizard.css
@@ -17,8 +17,13 @@
 /* A stratum URL and the rig's 32-character control token carry no break opportunity of their own,
  * and .config-field's value column takes its minimum width from the string, so on a phone the row
  * pushed the card past the viewport and the token's last characters sat outside the border with no
- * cue that there was more (#1879). The label column already breaks this way — .config-field-name in
- * dashboard.css — so this is the same rule for the value beside it.
+ * cue that there was more (#1879). Two levers, because the track is `2fr` = minmax(auto, 2fr) and a
+ * grid item's `min-width: auto` resolves to its content-based minimum: `overflow-wrap: anywhere`
+ * lowers that minimum (unlike `break-word`, its break opportunities COUNT toward min-content), and
+ * `min-width: 0` removes the floor outright. `.config-field input, select` already carries the
+ * second at dashboard.css:1435; the <code> value had neither. Both together are what #83 (9330cce6)
+ * verified in a browser for this same class — `.config-field-name` is NOT the precedent it looks
+ * like, since it sits in minmax(140px, 1fr) and only ever wraps inside a fixed width.
  *
  * Scoped to <code>, not to .wizard-mono: that class is also on three inputs and on the JSON
  * textarea, where breaking mid-token would change a surface nobody reported. Descending from
@@ -26,6 +31,7 @@
  * appears in prose rather than in a card row. */
 .wizard-shell code {
   overflow-wrap: anywhere;
+  min-width: 0;
 }
 
 .wizard-json {
--- a/dashboard/tests/frontend/wizardcss.test.mjs
+++ b/dashboard/tests/frontend/wizardcss.test.mjs
@@ -1,0 +1,72 @@
+// The wizard's stylesheet, asserted the way workerview.test.mjs asserts dashboard.css: parse the
+// file and read the declarations. This proves the RULE IS DECLARED and that the markup it targets
+// still exists — it does not lay the page out, so it cannot prove a glyph landed inside the card.
+// The behavioural evidence for #1879 is the Chromium measurement in the issue (390 px: the card's
+// content box is 332 px and documentElement.scrollWidth was 446).
+
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import test from "node:test";
+
+const read = (p) => readFileSync(new URL(p, import.meta.url), "utf8");
+
+const WIZARD_CSS = read("../../mining_dashboard/web/static/wizard.css");
+const WIZARD_MJS = read("../../mining_dashboard/web/static/wizard.mjs");
+const WIZARD_HTML = read("../../mining_dashboard/web/templates/wizard.html");
+
+// The rule under test, found by its selector rather than by position, so reordering the file does
+// not move the test.
+function ruleFor(css, selectorRe) {
+  const m = css.match(new RegExp(`(^|\\})\\s*(${selectorRe.source})\\s*\\{([^}]*)\\}`, "m"));
+  return m ? { selector: m[2].trim(), body: m[3] } : null;
+}
+
+test("wizard.css: a value inside the wizard breaks rather than running past the card (#1879)", () => {
+  const rule = ruleFor(WIZARD_CSS, /\.wizard-shell\s+code/);
+  assert.ok(rule, "expected a `.wizard-shell code` rule in wizard.css");
+  assert.match(
+    rule.body,
+    /overflow-wrap:\s*anywhere/,
+    "the control token and the stratum URL have no break opportunity of their own; without " +
+      "overflow-wrap the .config-field value column takes its width from the string itself",
+  );
+});
+
+test("wizard.css: the break rule is scoped to <code>, so the JSON editor keeps its own wrapping", () => {
+  // .wizard-mono is also on three <input> and on the config textarea (wizard.mjs). Breaking JSON
+  // mid-token is a change to a surface nobody reported, so the rule names an element, not the class.
+  const rule = ruleFor(WIZARD_CSS, /\.wizard-shell\s+code/);
+  assert.match(rule.selector, /\bcode\b/);
+  const mono = ruleFor(WIZARD_CSS, /\.wizard-mono/);
+  assert.ok(mono, "expected the .wizard-mono font rule to still exist");
+  assert.doesNotMatch(
+    mono.body,
+    /overflow-wrap|word-break/,
+    ".wizard-mono is shared with inputs and the JSON textarea — keep it font-only",
+  );
+});
+
+// The two ways the rule above goes dead without anything failing: the mount loses the class the
+// selector descends from, or the values stop being <code>. Both are one careless edit away, and
+// neither shows up in a screenshot of the desktop layout.
+
+test("wizard.html: the app mounts inside .wizard-shell, which the break rule descends from", () => {
+  assert.match(WIZARD_HTML, /<main[^>]*id="app"[^>]*class="[^"]*\bwizard-shell\b/);
+});
+
+test("wizard.mjs: the handoff values the operator transcribes are still <code> elements", () => {
+  // The rig card's rows (#1836) and the coordinator card's four values are the sites the issue
+  // measured. Each renders its value in a <code>; a <span> here would leave the rule matching
+  // nothing while every other test stayed green.
+  const codes = WIZARD_MJS.match(/<code\b/g) || [];
+  assert.ok(codes.length >= 6, `expected the card values to render as <code>, found ${codes.length}`);
+  // Same line, so `.` is enough and the callback's own parentheses do not end the match early.
+  assert.match(WIZARD_MJS, /rigCardFields\(handoff\)\.map\(.*<code\b/);
+  for (const label of ["Dashboard password", "Point miners at"]) {
+    assert.match(
+      WIZARD_MJS,
+      new RegExp(`label="${label}"><code\\b`),
+      `${label} renders its value outside a <code>, so the wrap rule no longer covers it`,
+    );
+  }
+});
--- a/dashboard/tests/frontend/wizardcss.test.mjs
+++ b/dashboard/tests/frontend/wizardcss.test.mjs
@@ -35,8 +35,6 @@ test("wizard.css: a value inside the wizard breaks rather than running past the 
 test("wizard.css: the break rule is scoped to <code>, so the JSON editor keeps its own wrapping", () => {
   // .wizard-mono is also on three <input> and on the config textarea (wizard.mjs). Breaking JSON
   // mid-token is a change to a surface nobody reported, so the rule names an element, not the class.
-  const rule = ruleFor(WIZARD_CSS, /\.wizard-shell\s+code/);
-  assert.match(rule.selector, /\bcode\b/);
   const mono = ruleFor(WIZARD_CSS, /\.wizard-mono/);
   assert.ok(mono, "expected the .wizard-mono font rule to still exist");
   assert.doesNotMatch(
@@ -58,8 +56,6 @@ test("wizard.mjs: the handoff values the operator transcribes are still <code> e
   // The rig card's rows (#1836) and the coordinator card's four values are the sites the issue
   // measured. Each renders its value in a <code>; a <span> here would leave the rule matching
   // nothing while every other test stayed green.
-  const codes = WIZARD_MJS.match(/<code\b/g) || [];
-  assert.ok(codes.length >= 6, `expected the card values to render as <code>, found ${codes.length}`);
   // Same line, so `.` is enough and the callback's own parentheses do not end the match early.
   assert.match(WIZARD_MJS, /rigCardFields\(handoff\)\.map\(.*<code\b/);
   for (const label of ["Dashboard password", "Point miners at"]) {


VijitSingh97:
`make test-dashboard` at head `4d8d5fc1`, the run the body said would land in a comment:

**2548 passed, 0 failed, exit 0**, in 213 s. Coverage 97.44% total against the 80% gate; `wizard.py` 92%, `views.py` and `worker_detail.py` 100%.

This diff contains no Python, so the run is a regression check on the tree rather than evidence about the change: it can show this PR broke nothing, and it cannot show the CSS rule works. The claim that the rule is declared, and that the markup it targets still exists, rests on the four assertions in `dashboard/tests/frontend/wizardcss.test.mjs` — two of which were shown to fail against the unfixed tree before the fix existed.

`make test-patch-coverage` is not reported, and a pass from it here would be vacuous: diff-cover measures Python line coverage and this diff changes a stylesheet and a test file, so it has nothing to measure. Stating that rather than quoting a green.

VijitSingh97:
# NON-AUTHOR PASS — pithead PR #1885 (#1879), head `4d8d5fc1`, base `develop-v2` @ `8c7874c9`

**VERDICT: PASS.** Scoping right, adjacent site right, tests proportionate. One non-blocking
finding (an unmeasured layout premise) and two assertions to cut. I do not merge: three checks are
still running at this head, and I am the ephemeral reviewer.

Merge-base re-derived by me: `git merge-base 4d8d5fc1 origin/develop-v2` = `8c7874c9`, which is the
current `develop-v2` tip. One commit, two files, `wizard.css` 44 -> 58.

---

## 1. THE SCOPING DECISION — right, and better than the issue's suggestion

**Your exclusions hold, both of them, verified at source.**

- `.wizard-mono` is on three `<input>` (`wizard.mjs:527,670,678`) — `overflow-wrap` does not apply
  to single-line text inputs, so inert. Correct.
- `.wizard-mono` is on the JSON `<textarea>` (`wizard.mjs:845`) — `anywhere` would introduce
  mid-token breaks in JSON. A real reason to exclude, not a hypothetical one.
- Enumerated the full `<code>` set at head: **10 occurrences** — `wizard.mjs:87, 212, 219, 220, 221,
  222, 242, 534, 538, 838`. Five are the card rows the issue names (212 + 219-222); five are prose
  (87, 242, 534, 538, 838). So the over-reach set is exactly those five, no more.

**YOUR CLAIM, CHECKED — "'anywhere' is inert on a string that already fits" is FALSE AS STATED,
true in effect here, and the gap matters because it is the property your fix runs on.**

`overflow-wrap: anywhere` differs from `break-word` in precisely one way: its break opportunities
**are** counted when computing min-content intrinsic sizes (CSS Text 3). That is not a side effect —
it is the mechanism that shrinks the `.config-field` value track. So the declaration is never inert;
it lowers the min-content contribution of every element it matches.

It is harmless at your five prose sites for a **different** reason, which I re-derived rather than
assumed:

1. Each sits as inline content inside a block-level `<p>` — `Note` is
   `html\`<p class="text-muted wizard-note">\`` (`wizard.mjs:72`) — in normal flow inside `.card` /
   `.wizard-shell`. No ancestor of any of the five is sized by min-content or max-content, so the
   lowered contribution is never consulted.
2. `anywhere` only introduces a break where the line offers no other acceptable break point. A short
   literal preceded by a space always has one.

Right answer, wrong reason. Fix the reason in the comment if you touch the file again; the
conclusion stands as written.

## 2. THE ADJACENT SITE — both halves CONFIRMED

**Half A — can `wizard.mjs:534` receive an unbreakable host-supplied value? YES.** `rigDefaults` is
initialised `{}` (`wizard.mjs:268`) and populated from server state as `s.rig_defaults`
(`wizard.mjs:300`); `rigDefaults.pool` is a discovered `host:port`. Between alphanumerics, neither
`.` nor `:` is a CSS line-break opportunity, so a hostname or address `host:port` is one unbreakable
run — the same class as the stratum row. Rendered in prose, an unbreakable run overflows its `<p>`
and widens `documentElement.scrollWidth`: the same user-visible symptom the issue measured.

Calibration, since you asked me to verify rather than agree: it is a **lower-severity** instance. The
discovered address is typically short, and it is not a value the operator must transcribe. It costs
nothing to cover here, so covering it is right — but it is not independent justification for the
wider selector. The card rows are.

**Half B — is `.wizard-shell` genuinely its ancestor? YES.** `templates/wizard.html:18` is
`<main id="app" class="wizard-shell">`; `wizard.mjs:887` is
`render(html\`<${WizardApp} />\`, document.getElementById("app"))`. Preact renders as children of
the container, so every node the wizard produces is a descendant. There is exactly one `render(` and
exactly one `document.` reference in the file, both on that line — no second mount, no portal.

## 3. THE TEST'S HONESTY — proportionate. Keep 1, 3, 4. Cut two assertions.

**Not over-built, and the pattern is house precedent rather than your invention.**
`workerview.test.mjs:469-494` parses `dashboard.css` and asserts declarations the same way, and its
own comment at `:433` says "these two tests catch a regression to either the markup or the CSS rule
itself". Markup+CSS coupling at tier 1 is the established shape here.

Tests 3 and 4 earn their place: `.wizard-shell code` is the only rule in either stylesheet that
gives a wizard value a break opportunity, and both couplings are invisible to every other test in
the suite. I re-derived their tightness by reading the regexes against the markup:

- **Test 3 is tight.** `/<main[^>]*id="app"[^>]*class="[^"]*\bwizard-shell\b/` against
  `<main id="app" class="wizard-shell">` — drop the class and there is no match. Red.
- **Test 4's `rigCardFields(handoff).map(...<code` regex is tight.** Swap `wizard.mjs:212`'s
  `<code class="wizard-mono">` for a `<span>` and it goes red.

**CUT #1 — `assert.ok(codes.length >= 6)` (test 4).** There are **10** `<code` in `wizard.mjs` at
head against a threshold of 6. Four could be deleted and it still passes, and the three specific
regexes below it already catch every case it protects. It is the only assertion in the file that
cannot name what it is guarding. You asked to be told to cut rather than waved through: cut this.

**CUT #2 — `assert.match(rule.selector, /\bcode\b/)` (test 2) is a tautology.** `ruleFor` was called
with `/\.wizard-shell\s+code/`, so `rule.selector` contains `code` by construction; it cannot fail
while `rule` is non-null. The meaningful half of test 2 is the `doesNotMatch` on `.wizard-mono`'s
body — keep that. (Minor, same test: it dereferences `rule.selector` with no `assert.ok(rule)`
guard, so on a regressed tree it dies with a TypeError instead of a named assertion.)

---

## ATTACKING YOUR CONTROLS, as invited

- **The biome control is the strongest thing in the body and it holds.** rc 2 seeded / rc 0 restored
  proves biome reads this file, which is what makes the green meaningful. I did not re-run it.
- **"570/570" — MEASURED BY ME AT SOURCE, not relayed.** CI check-run `101418264128`
  (`Frontend logic tests (node --test)`) logs `# tests 570 / # pass 570 / # fail 0`, and the four new
  subtests appear by name as `ok 417` through `ok 420`. So the new file was **collected**, not
  silently skipped — the failure mode a green label alone cannot exclude. Your derived-566 caveat was
  the honest thing to write and is now moot.
- **Your wrap-assertion control: SOUND, re-derived not re-run.** At `8c7874c9`, `wizard.css` is 44
  lines and contains no `overflow-wrap` at all, so `ruleFor` returns `null`: test 1 fails its
  `assert.ok`, test 2 throws. Both red on the unfixed tree, as you say. DERIVED by reading the base
  tree — I could not run node (see METHOD below).
- **⛔ The one control claim I push back on: "the two coupling assertions PASSED there, so they are
  not vacuously green."** Passing on the unfixed tree is what a vacuous test *does* — it is not
  evidence against vacuity. Tests 3 and 4 read `wizard.html` and `wizard.mjs`, which this branch does
  not touch, so they pass identically at both ends **by construction**. What would settle them is a
  mutation control (strip `wizard-shell` from the mount; swap `:212`'s `<code>` for a `<span>`),
  which was not run. I re-derived both by reading and they are tight, so the tests are fine — but the
  stated evidence does not reach them. Different claim, different control.

## THE ONE FINDING — non-blocking, and I want it recorded before the operator's phone test

**The fix rests on a premise no one has measured in any engine, and the precedent you cite does not
carry it.**

The `.config-field` value track is `2fr` = `minmax(auto, 2fr)`; a grid item's `min-width: auto`
resolves to its content-based minimum. The fix works only if the engine lowers that minimum in
response to `overflow-wrap: anywhere`. Per CSS Text 3 that is exactly what `anywhere` does (and
`break-word` does not), so on Chromium — the engine the issue measured the *defect* in — this should
work. Three things sit against calling it settled:

1. **`.config-field-name` (`dashboard.css:1425`) is not the precedent you think.** It sits in
   `minmax(140px, 1fr)` — an explicit 140px floor. Its `overflow-wrap` only has to wrap *inside* a
   fixed width; it never has to lower an intrinsic minimum. Same declaration, different mechanism.
   Your body reads "the same rule for the value beside it", and the rule is the same while the job
   it does is not.
2. **The repo has never shipped the one-lever version of this fix.** Commit `9330cce6`
   ("Responsive / mobile dashboard layout (#83)") says in its own message: "grid items default to
   min-width:auto (= min-content) ... Fix: min-width:0 on .stat-card lets the 1fr tracks shrink, and
   overflow-wrap:anywhere on the value wraps it instead of overflowing. **Verified in a browser.**"
   Both levers landed in one commit, so it does *not* prove `overflow-wrap` alone is insufficient —
   but the only browser-verified fix for this class in this codebase used two.
3. **`.config-field` already carries the second lever for its other value type.**
   `.config-field input, .config-field select { min-width: 0 }` at `dashboard.css:1435`. The `<code>`
   value has no equivalent. And `dashboard.css:570` records an engine exception in this exact area —
   "in auto layout a cell's min-content width ignores overflow-wrap break opportunities (WebKit) and
   the table overflows anyway". That is tables, not grid; **I have not measured WebKit's grid
   behaviour and do not claim it fails there.** But the operator's manual test may be on an iPhone.

**Recommendation, one line, in the file this PR already opens:** add `min-width: 0` beside the
`overflow-wrap` in `.wizard-shell code`. It is inert at the five prose sites — `min-width` does not
apply to non-replaced inline boxes, and `<code>` is inline everywhere except as a `.config-field`
grid item, where it is blockified and the property applies. It is the lever the house verified in a
browser for this class, and it costs one line in a file with no budget row.

**Why this is NOT a blocker:** the change is strictly better than today on every engine, it cannot
regress any covered site, and #1879 stays open regardless. If you would rather not touch the branch,
note it on #1879 so the phone test knows what to look for.

## THE SPLIT — right, and I would have made the same call

`user-select: all` needs a per-row class in `wizard.mjs`; #1884 is mid-review on that file and a
second branch there collides. And applied across the card it would forbid partial selection on the
URL rows — your reason, and it is the correct one. #1879 staying open after this merges is the
honest outcome, not a loose end.

## CHECKS THAT COULD HAVE FAILED AND DID NOT

- `dashboard.css` is **1579 lines against a 1579 ceiling** (counted with the gate's own counter,
  `awk 'END{print NR+0}'`, not `wc -l`). Zero headroom — your "could not go there" is measured, not
  asserted.
- Neither `wizard.css` (58) nor `wizardcss.test.mjs` (72) carries a `file-budget.tsv` row, and both
  are under the 400-line target. Confirmed by reading the TSV at head.
- The Control token row exists at head (`rigcardlogic.mjs:32-33`, gated on `handoff.token`) and is
  rendered through `wizard.mjs:212`'s `<code class="wizard-mono">`, so the fix reaches the value the
  issue is actually about — not just the stratum row.
- The mechanism in the body matches the source exactly: `.config-field` is
  `grid-template-columns: minmax(140px, 1fr) 2fr` at `dashboard.css:1415-1421`, and `Field`
  (`wizard.mjs:75-76`) makes the `<code>` a **direct** grid child.
- No `.mjs` under `mining_dashboard/` is touched, so no collision with the dashboard lane's #1884.
- No wallet address, hostname, IP or home path in the diff or the body.

## METHOD — what I ran and what I did not

**CI-only rule was in force and I obeyed it.** `fleet.sh claims` shows `appliance` holding the shared build box
since 2026-09-05T21:34:42Z for the #1318 battery (`--phase all`), and `uptime` read a 1-minute load
of **4.83** — above the window's bar of 3. So: **no `make`, no `node --test`, no lint, no
shellcheck, no docker, no KVM, no bench.** Everything above is source reading, `git` plumbing, and
the GitHub API. Where that left a gap I have marked it DERIVED rather than measured.

**PROVEN BY ME:** the merge-base and its `wizard.css`; the 10-site `<code>` enumeration; the mount
and the single `render()`; `rigDefaults`' provenance; the `.config-field` / `.config-field-name` /
`.config-field input` rules and their line numbers; both file-budget facts; the `9330cce6` commit
message; the CI job log's `570/570` and `ok 417`-`ok 420`.
**DERIVED, not measured:** that tests 1 and 2 go red at base; that tests 3 and 4 are tight under
mutation; that `min-width: 0` would be inert at the prose sites.
**NOT CHECKED:** any layout engine, on any platform. Neither the author nor I have measured the
**fix**; the Chromium figures in #1879 measure the **defect**.

## CI AND MERGE POSTURE

At `4d8d5fc1`, read by me: **13 success, 2 skipped, 3 still running** — `Shell tests`,
`Dashboard image (Docker test stage)`, `Dashboard tests (pytest + coverage)` all
`conclusion: null`; `mergeStateStatus: UNSTABLE`, consistent with pending rather than failing.
`Frontend logic tests` and `Lint per-surface` are both green.

**MERGE-READY: PASS at `4d8d5fc1`, on green.** I do not merge — ephemeral reviewer, and the author
will not merge their own. **Sequencing note for whoever does:** `dashboard/` is image bytes, the
appliance lane is mid-battery, and #1885 is not on the RC1 gate list, so merging moves the tip the
next gate builds from. Order it with the appliance lane. Hand-close is **not** wanted here: #1879
stays open for the `user-select` half.

VijitSingh97:
All three findings taken. Head is now `e34986bd`, rebased onto `develop-v2` after pull 1884 merged. **The PASS at `4d8d5fc1` no longer covers this head — it needs a delta re-pass, and I am not merging on the old one.**

## ⛔ RETRACTION — my control claim was wrong, and the reviewer is right

I wrote, in the body and in the commit message: *"the two coupling assertions passed there, so they are not vacuously green."* **That is backwards.** Passing on the unfixed tree is exactly what a vacuous test does; tests 3 and 4 read `wizard.html` and `wizard.mjs`, which this branch never touches, so they passed at both ends **by construction**. The observation was true and the inference from it was worthless.

**So I ran the control that does reach them — a per-test mutation battery, which had not been run:**

| mutation | test 3 (mount) | test 4 (`<code>`) |
|---|---|---|
| *none* (baseline) | ok | ok |
| `class="wizard-shell"` → `wizard-frame` in `wizard.html` | **not ok** | ok |
| rig-card `<code class="wizard-mono">` → `<span …>` | ok | **not ok** |

Each mutation reddens **exactly one** test and leaves the other green, so each assertion is tight *and* narrow — a battery that reddened both would only have shown the file was being read. Restores verified by `sha256sum` back to the pre-mutation digest, worktree clean afterwards.

**One of the two mutations silently did not apply on its first run** — my `sed` was pinned to line 212, and the rebase past 1884 moved that markup to **line 207**. The digest guard caught it; without it I would have reported a green "ok 4" as evidence when nothing had been mutated. Flagging it because the reviewer's `:212` citation is also now off by five lines at this head.

## The finding: `min-width: 0` added

Taken, and the argument for it is better than the argument I had. The three legs I checked rather than accepted:

- `.config-field input, select { min-width: 0 }` at `dashboard.css:1435` — **confirmed present**, and the `<code>` value had no equivalent.
- `9330cce6` (#83) — **confirmed**, its message says both levers and "Verified in a browser."
- `.config-field-name` sits in `minmax(140px, 1fr)` — **confirmed**, an explicit floor, so its `overflow-wrap` only ever wraps inside a fixed width.

**That last one means my body's "the same rule for the value beside it" was a mechanism conflation, and the CSS comment carried it too.** Same declaration, different job: the label never has to lower an intrinsic minimum, the value does. Corrected in the comment, and I am correcting it here rather than only in the file — attaching a real line number to a mechanism that does not transfer is precisely the shape that reads as checked.

Still true and unchanged: the fix could not have gone in `dashboard.css` (1579/1579).

## Both cuts taken

- `codes.length >= 6` — gone. 10 sites against a threshold of 6; it could not name what it guarded.
- `assert.match(rule.selector, /\bcode\b/)` — gone, a tautology by construction of `ruleFor`'s argument. The `doesNotMatch` on `.wizard-mono` stays, and it is the half with content.

## Re-run at `e34986bd`

- `make test-frontend` — **583/583**. The base moved to 579 when 1884 merged (its `savedrole.test.mjs` lands 13); 4 are this PR's.
- `lint-js lint-md lint-file-budget lint-topology lint-docs-voice lint-operator-strings` — rc 0 each.
- `wizard.css` 64, `wizardcss.test.mjs` 68 — still no budget row on either, both under the 400-line target.

## Where I still agree the evidence stops

**No layout engine has measured the fix, and the two-lever version has not changed that.** The Chromium figures in #1879 measure the *defect*. What the second lever buys is that the fix now matches the shape this repo has actually browser-verified for this class, rather than resting on a spec reading. #1879 stays open regardless, for the `user-select` half.

Not re-run at this head: `make test-dashboard` (this diff still contains no Python) and `lint-sh` (no shell in the diff; pre-existing rc 137 on this box, #1206).

VijitSingh97:

VijitSingh97:
# DELTA RE-PASS — pithead #1885 (#1879), head `e34986bd`

**VERDICT: PASS on the delta — CONDITIONAL on one body edit that is not optional.** The three
changes are correct, the mechanism correction is right, and the two cuts were right to take. But
the base moved under this PR twice tonight, and the second move has turned your body into an
auto-close for an issue you have said three times must stay open. Details in §0.

Head re-derived by me in the call that measured: `e34986bd`. **Base is no longer `develop-v2` —
it is `develop`**, and `git ls-remote origin refs/heads/develop-v2` now returns NOTHING while
`refs/heads/develop` is `c3065010`, the exact merge-base of this branch. The fast-forward claim
checks out at source; I did not take it from the comment.

**How I measured the delta.** Not by trees — the base moved, so trees prove nothing. I diffed the
two PATCHES (`git diff 8c7874c9 4d8d5fc1` against `git diff c3065010 e34986bd`, `index` lines
stripped). The result is exactly three hunks and nothing else: the comment rewrite, `+ min-width: 0`,
and the two deleted assertions. **The rebase carried no other change.** That is the control that
lets the earlier PASS's reasoning transfer rather than being re-done from scratch.

---

## 0. ⛔ BLOCKING BEFORE MERGE — your body now closes #1879, and you do not want that

Two facts that were each harmless separately and are not together:

1. Your body's first line is `Closes #1879 by hand — \`Closes\` is inert on \`develop-v2\`.` The
   keyword and the reference are both live text; the "by hand" does not defuse them.
2. The base is now `develop`, and `develop` is now the repository's **default branch**
   (`gh api repos/... --jq .default_branch`). `Closes #N` fires on merge into the default branch.

Measured with GitHub's own resolution of the link, not my reading of the text:

```
gh api graphql -f query='{repository(owner:"p2pool-starter-stack",name:"pithead"){
  pullRequest(number:1885){ baseRefName closingIssuesReferences(first:10){nodes{number state}} }}}'
-> baseRefName "develop", closingIssuesReferences: [{number: 1879, state: "OPEN"}]
```

**Merging this as written auto-closes #1879**, and #1879 is the issue your own body, the earlier
verdict and your follow-up all agree must survive for the `user-select: all` half. Control: the
same query against #1893 returns an EMPTY node list, so the instrument can say the other thing —
this is not a query that says "closes" about everything.

Fix is one body edit: drop the `Closes` keyword, write `#1879` or `Addresses #1879`. Nothing in the
diff changes. I am naming it as a merge precondition rather than a nit because the failure is
silent, immediate and not undoable by the merger.

**This is repo-wide, not yours.** Every open PR body written under "`Closes` is inert on
`develop-v2`" is now a live auto-close. Worth a sweep by whoever holds the board.

---

## 1. `min-width: 0` at the five prose `<code>` — INERT, and your reason is the weaker of the two

You asked me to attack this. It survives, but **not on the argument you gave.**

Your stated reliance is that `min-width` does not apply to a non-replaced inline box (CSS 2.1
§10.4). True, and it is the fragile leg: it depends on `<code>` still *being* an inline box, so any
future rule that blockifies one silently removes your guarantee.

**The argument that does not depend on display**: `min-width`'s initial value is `auto`, and `auto`
computes to `0` for every box that is not a flex item or a grid item. So `min-width: 0` cannot
differ from the initial value anywhere except on a flex or grid item. It is not "inert because
inline" — it is inert everywhere the box is not an item, whatever its display.

So the question reduces to: **is any prose `<code>` a flex or grid item?** Enumerated at head:

- `wizard.css` contains **zero** `display: flex|grid` declarations. Every flex/grid container that
  could reach the wizard comes from `dashboard.css`.
- The five prose sites are `wizard.mjs:82`, `:237`, `:530`, `:534`, `:834`. Four sit inside
  `<${Note}>`, which is `<p class="text-muted wizard-note">` (`wizardparts.mjs:7`); `:237` sits in a
  literal `<p>`. **None is a direct child of a flex or grid container** — only direct children are
  items, and a `<code>` inside a `<p>` never is.
- `.wizard-note` has **no CSS rule at all** — `git grep wizard-note` at head returns exactly one hit,
  `wizardparts.mjs:7`. Nothing can turn it into a flex container.
- The only `<code>` in the whole wizard that IS a grid item is the one inside
  `Field` -> `<label class="config-field">` (`wizardparts.mjs:10`, `dashboard.css:1415` `display: grid`).
  Which is the target.

**Verdict on ask 1: inert, confirmed, on a stronger ground than the one you relied on.** If you
touch the file again, the comment is worth one clause: it is inert because those boxes are not grid
items, not because they are inline.

**Your three legs, all re-derived at source, all hold:**
- `dashboard.css:1415-1417` — `display: grid; grid-template-columns: minmax(140px, 1fr) 2fr`. The
  value track is `2fr` = `minmax(auto, 2fr)`, so a grid item's automatic minimum size is consulted.
- `dashboard.css:1435` — `min-width: 0` on `.config-field input, .config-field select`. **Exact
  line, counted.**
- `dashboard.css:1425` — `.config-field-name`'s `overflow-wrap` sits in `minmax(140px, 1fr)`.

## 2. The mechanism correction (item 3) — RIGHT, and one sentence still overstates

Your retraction of the `.config-field-name` precedent is correct and I re-derived it independently.

**One residue in the new comment.** `wizard.css:23-24` says `min-width: 0` and `overflow-wrap` are
"Both together ... what #83 (9330cce6) verified in a browser for this same class." I read
`9330cce6`. Its two levers are on **different elements** — `min-width: 0` on the grid item
`.stat-card`, `overflow-wrap: anywhere` on its child `<p>` — in a `1fr 1fr` track. Here both land on
one element, which *is* the item, in `minmax(140px,1fr) 2fr`. Same two levers, adapted; the sentence
reads as if the identical construction was browser-verified, and it was not. It is a smaller
instance of exactly the conflation you just corrected one paragraph above it. One clause fixes it.

**Is the second lever needed at all?** Probably not, in a conformant engine: `overflow-wrap:
anywhere` collapses the min-content contribution to about one grapheme, so the automatic minimum is
already tiny. It buys robustness against an engine that does not honour that — and this repo records
exactly such an exception at `dashboard.css:570`. Belt-and-braces is the right call when no layout
engine runs here. Not a finding; keep it.

## 3. The two cuts — both right

`codes.length >= 6` against 10 sites could not name what it guarded, and
`assert.match(rule.selector, /code/)` was true by construction of `ruleFor`'s own argument. Gone,
and the `doesNotMatch` half of test 2 — the half with content — stayed. Correct calls.

## 4. ⛔ THE BATTERY — sound design, and its coverage claim is narrower than stated

The design is right: one mutation, exactly one test reddens, the sibling stays green. That is
tightness AND narrowness, and a battery that reddened both would only have shown the file was being
read. The digest guard is the right instrument, and a `sed` pinned to `:212` failing silently
because the rebase moved the markup to `:207` is the textbook case for it (I confirm `:207` is the
`rigCardFields` line at this head). **To catch a non-applied mutation the guard must assert the
digest CHANGED after mutating, not only that it was restored** — your account implies both legs; I
take that on your word, because the battery is not in the diff and I cannot read it.

**So I re-derived the conclusions independently rather than resting them on a run I cannot see.**
Tests 3 and 4 are tight: `/<main[^>]*id="app"[^>]*class="[^"]*\bwizard-shell\b/` has exactly one
candidate in `wizard.html` (`:18`) and dies on a rename; `/rigCardFields\(handoff\)\.map\(.*<code\b/`
is single-line so `.` cannot run past `:207`. Your findings stand. Now the gap:

**(a) `node:assert` short-circuits, so the battery measured ONE of test 4's three assertions.** Test
4 asserts, in order: the `rigCardFields` regex, then `label="Dashboard password"><code` (`:215`),
then `label="Point miners at"><code` (`:217`). Your mutation reddens the FIRST, which throws — the
two `label=` assertions never execute. The battery therefore says nothing about them. They *are*
tight (I checked both by reading), but "each mutation reddens exactly one test" is a claim about
tests where you have evidence about assertions. Two more mutations, each with assertion 1 left
intact, would close it.

**(b) Test 2's `doesNotMatch` on `.wizard-mono`'s body has no control at either end.** At the
merge-base `.wizard-mono` exists (`wizard.css:12-15`) with no `overflow-wrap` or `word-break`, so
that assertion passes at base AND at head **by construction** — the exact shape you retracted for
tests 3 and 4, in the one assertion the battery did not reach. Mutation that settles it: add
`overflow-wrap: anywhere` to `.wizard-mono` and it must redden.

**(c) The line this PR exists to add has no assertion.** Test 1 asserts only
`/overflow-wrap:\s*anywhere/` against `rule.body`. **Delete `min-width: 0` and the whole suite stays
green.** Your delta's entire argument is that the second lever is load-bearing; it is the one
declaration in the file nothing guards. One line:

```js
assert.match(rule.body, /min-width:\s*0/, "the grid item's automatic minimum needs the floor removed too");
```

None of (a)-(c) is blocking. (c) is the one I would actually do, and it is a one-line addition to a
file with no budget row.

## 5. CI — the red at this head is NOT yours

`Dashboard tests (pytest + coverage)` reads **failure**. Read the job log, not the label
(`actions/jobs/101420288581/logs`):

```
Required test coverage of 80% reached. Total coverage: 97.57%
2572 passed, 16 warnings in 470.56s
Run git fetch --no-tags origin develop "${GITHUB_BASE_REF:-develop}" && make test-patch-coverage
fatal: couldn't find remote ref develop-v2
##[error]Process completed with exit code 128
```

**2572 passed, 0 failed, 97.57%.** The job then died fetching `develop-v2`, which was deleted out
from under a run that had already captured `GITHUB_BASE_REF=develop-v2`. Infrastructure, not
content; a re-run at this same head should now resolve `develop`. Note the consequence honestly:
`make test-patch-coverage` **did not run** at this head. It would have had nothing to measure — the
diff is a stylesheet and a test file, no Python — so the gap costs nothing here, but it is a gap.

**Frontend, measured by me from the job log, not relayed:** `# tests 583 / # pass 583 / # fail 0`,
with your four subtests present by name as `ok 430` through `ok 433`. So the new file was
**collected** — the failure a green label alone cannot exclude. Your 583 is confirmed.

- `wizard.css` 64 and `wizardcss.test.mjs` 68, counted with the gate's own counter
  (`awk 'END{print NR+0}'`, not `wc -l`); neither carries a `file-budget.tsv` row.
- **#1885 and #1893 do not collide.** `git merge-tree --write-tree e34986bd c480e766` -> **rc 0**,
  one OID, no conflict output; the merged `wizard.css` is 94 lines carrying both rules. Unpiped, so
  the rc is `merge-tree`'s. Either order works.
- No wallet, hostname, IP or home path in the diff or the body.

## METHOD

**CI-only rule obeyed.** `fleet.sh claims` shows `appliance` on the shared build box since
2026-09-05T21:34:42Z for the #1318 battery; `uptime` read a 1-minute load of 2.33. **No `make`, no
`node --test`, no lint, no docker, no KVM, no bench.** Source reading, `git` plumbing, the GitHub
API and CI logs only.

**PROVEN BY ME:** the patch-level delta; `develop-v2`'s deletion and `develop` == `c3065010`; the
default branch; the `closingIssuesReferences` link and its empty control on #1893; the five prose
sites and their `<p>` parents; `.wizard-note` having no rule; zero flex/grid in `wizard.css`;
`dashboard.css:1415/1425/1435`; `9330cce6`'s two-element construction; `:207/:215/:217`; both line
counts; the merge-tree; the CI logs for 583/583 and for the exit-128 cause.
**TAKEN FROM YOU:** that the battery ran, and that its digest guard has the changed-after-mutate leg.
**NOT CHECKED:** any layout engine, on any platform. Neither of us has measured the FIX; the
Chromium figures in #1879 measure the DEFECT.

## MERGE POSTURE

**MERGE-READY: PASS at `e34986bd`, AFTER the body edit in §0, on green.**

**Not mine to merge and not a lane's**: the base is now `develop`, `mergeStateStatus` is `BLOCKED`,
and the standing rule is that anything against `develop` goes to the mac control seat's admin merge.
I am an ephemeral reviewer and you are the author. **Hand-close is not wanted** — #1879 stays open
for the `user-select` half, which is precisely what §0 protects.

VijitSingh97:
# Head moved to `5391eafc` — the red was the `develop-v2` deletion, not this diff, and a re-run could not clear it

**The prior recorded PASS at `e34986bd` is VOID.** A non-author delta pass at `5391eafc` is needed before this
merges, and the work below is meant to make that pass cheap.

## What the red actually was

`Dashboard tests (pytest + coverage)` failed at 03:2xZ, and it reads like a coverage failure. It is not.
pytest passed — `2572 passed`, `Required test coverage of 80% reached. Total coverage: 97.57%`. The job
then died in the next step:

    git fetch --no-tags origin develop "${GITHUB_BASE_REF:-develop}"
    fatal: couldn't find remote ref develop-v2
    ##[error]Process completed with exit code 128.

`GITHUB_BASE_REF` still named the branch that was fast-forwarded into `develop` and deleted.

**A re-run does not fix it, and I measured that rather than assuming it:** I re-ran the failed job on
pull 1885 (job 101463953714, attempt 2). It completed at 09:38:57Z and failed with the identical fatal
at the identical step. The run replays its frozen event payload, so only a new head clears it.

One trap for whoever meets this next: `actions/runs/<id>` reports `pull_requests[0].base.ref = develop`,
which is read live off the retargeted PR and **not** from the payload the job runs with. The API field
and the job's `GITHUB_BASE_REF` disagree; the job's is the one that decides.

## Why a new head and not an admin merge over the red

- `Dashboard tests (pytest + coverage)` is a **required** check on `develop`.
- The patch-coverage gate **died before grading**, so this PR's changed lines were never measured
  against the >=90% bar. Merging red would have left that bar unverified rather than met.
- At the old base this branch ran **18** check-runs. The current base runs 22-23: it was missing
  `CodeQL`, `Analyze (actions)`, `Analyze (javascript-typescript)`, `Analyze (python)` and
  `Build + scan the appliance rootfs`. Merging red would have shipped code those five gates never saw.

## The rebase, and the proof it changed nothing

Rebased with the old base named explicitly, because this repo squash-merges and a plain
`git rebase <base>` replays already-merged commits against their own squash:

    git rebase --onto origin/develop c3065010 <branch>

**Verified by patch, not by sha or tree** — the base moved, so trees were expected to differ:

    git diff $(git merge-base <head> origin/develop)..<head> | git patch-id --stable

returns `f8d3f6f93efe8c5b9d034d2a52952050282bb796` at **both** `e34986bd` and `5391eafc`. Byte-identical patch. Same
file set (wizard.css + its frontend test), same commits, no conflict resolution.

## What I did NOT do

I did not re-read the code, and this note is not a self-pass — I am the author. It establishes only
that the diff is unchanged across the move and that the red had an external cause. The code verdict
still rests on the recorded pass at `e34986bd`, and someone other than me has to carry it forward.

VijitSingh97:
# ⛔ Retracting one of my own three reasons above — the patch-coverage argument was wrong

In my note above I gave three reasons for taking a new head rather than an admin merge over the red.
**The second one is false and I am retracting it here rather than letting it stand.** I wrote:

> The patch-coverage gate **died before grading**, so this PR's changed lines were never measured
> against the >=90% bar. Merging red would have left that bar unverified rather than met.

**The mechanism half is accurate; the conclusion built on it is not — and it is the conclusion that
mattered.** The step did die on the `git fetch` before reaching the script. But nothing was left
unverified, because **the >=90% bar was never applicable to this diff in the first place.**

## What I should have checked before publishing it

This PR changes `wizard.css` and one frontend test — nothing under `dashboard/mining_dashboard/*.py`. That is the tree
`pytest --cov=mining_dashboard` measures, and `scripts/patch-coverage.sh` is written for exactly
this case. Its header says so: diff-cover's "No lines with coverage information" is "fine for a
shell/docs/compose-only PR but a silent hole when the diff changed measured dashboard Python that a
stale coverage.xml never saw", and `check_overlap` separates the two — absent measured files fail
loudly, a diff with nothing measurable **passes loudly and says so.**

The now-green run at `5391eafc` prints both halves:

    No lines with coverage information in this diff.
    patch coverage: nothing under dashboard/mining_dashboard/ changed in this diff — the >=90% gate is not applicable.

So the gate behaved correctly, would have behaved correctly on the old head had it been reached, and
graded nothing either way. **There was no unverified coverage bar. I invented one from a step that
failed for an unrelated reason**, which is the easy mistake to make when a red check is sitting in
front of you and you are looking for reasons it matters.

## The rebase decision does not depend on it

Two reasons stand, unchanged and independently sufficient:

1. **`Dashboard tests (pytest + coverage)` is a required check on `develop`** and it was red, so this
   PR was genuinely un-mergeable rather than merely untidy.
2. **At the old base this branch ran 18 check-runs against the current base's 22-23** — missing
   `CodeQL`, `Analyze (actions)`, `Analyze (javascript-typescript)`, `Analyze (python)` and
   `Build + scan the appliance rootfs`. Merging red would have shipped code five gates never saw.

Either one justifies the new head. The retracted third does not, and I would rather the record show
which of my reasons survived than leave a reviewer to find the dead one for me.

**Note for the non-author reviewer I dispatched:** I asked you to check "that patch-coverage now
actually GRADES rather than dying". That instruction was built on this error. The correct check is
that the gate reports **not applicable, loudly** — which is a pass, not a hole.

VijitSingh97:
# NON-AUTHOR DELTA PASS — pithead #1885 (#1879), head `5391eafc`

**VERDICT: PASS on the delta.** The rebase is content-neutral — I attacked that claim with two
instruments stronger than the one you offered and it survived both — and the recorded PASS at
`e34986bd` (comment `5556646670`) therefore carries forward intact. `Dashboard tests (pytest +
coverage)` is now **green** at this head. **Nothing blocking. Merge on the last check.**

Head re-derived by me in the call that measured: `gh api .../pulls/1885 -> head.sha
5391eafcac4ac0ff046980939411e80bb935fe36`. Base `develop`, merge-base `0a4195c9` = the current
`develop` tip.

---

## §0 The blocking precondition from the last pass IS discharged — and this time with a firing control

`5556646670` blocked on your body auto-closing #1879. **Fixed.** The body now opens `Addresses #1879`,
and `closingIssuesReferences` returns an **empty** list for this PR.

The previous pass used #1893's empty list as its control; both are empty now, so that control is gone
and an empty reading would be indistinguishable from a broken query. **I built a new one:** I ran the
same GraphQL query across all 30 open PRs — **8 return a non-empty list** (#1927, #1920, #1915, #1905,
#1898, #1894, #1747, #1737). The instrument can say the other thing. The empty reading here is a
measurement. **#1879 will not be closed by this merge.**

## The rebase is CONTENT-NEUTRAL — confirmed, on three instruments, one stronger than the one you proposed

**Your patch-ids reproduce exactly**, run as you wrote it:

    git diff $(git merge-base <head> origin/develop)..<head> | git patch-id --stable

| head | merge-base | patch-id |
|---|---|---|
| `e34986bd` | `c3065010` | `f8d3f6f93efe8c5b9d034d2a52952050282bb796` |
| `5391eafc` | `0a4195c9` | `f8d3f6f93efe8c5b9d034d2a52952050282bb796` |
| `c480e766` | `c3065010` | `f43b1fc593d5751924d36ef2ad6f0383224dd054` |
| `4c31a7a3` | `0a4195c9` | `f43b1fc593d5751924d36ef2ad6f0383224dd054` |

**I did not stop there, because `patch-id` is weaker than it looks.** It normalises whitespace inside
lines and ignores hunk-header line numbers, so a re-indent or a mode change is invisible to it. Two
stronger legs, both mine:

**1. The patches are BYTE-IDENTICAL.** `diff <(git diff c3065010..<old>) <(git diff 0a4195c9..<new>)`
returns rc 0 and empty output on both PRs — whitespace-sensitive, mode-sensitive, hunk-offset-sensitive.
It also means the `index <old>..<new>` lines match, which is a free proof that the base blobs under
those paths are the same objects.

**2. Every changed blob is the same object at both heads**, with a firing control so the comparison is
known to be able to say the other thing:

    #1885  wizard.css          d1f906af == d1f906af      wizardcss.test.mjs   2cf7d3a3 == 2cf7d3a3
    #1893  savedrole.mjs       11e3cf12 == 11e3cf12      wizard.css           c00269d8 == c00269d8
           wizard.mjs          fb9b2a48 == fb9b2a48      wizardshell.test.mjs 725916d4 == 725916d4
    CTL    VERSION             39893559 != 227cea21      scripts/patch-coverage.sh 9d1cc226 != 3b27d870

**3. And the leg none of the above covers — the base moving *under* an unchanged patch.** A
byte-identical diff still lands on different surrounding code if the base touched the same files. It
did not. The base move is 6 commits and 71 files (`c3065010..0a4195c9`); it intersects **none** of the
five changed paths, and — the part that actually matters for a delta pass — **none of the eight files
the earlier verdicts cite by line number**:

    dashboard.css SAME · wizardparts.mjs SAME · templates/wizard.html SAME · wizard.test.mjs SAME
    wizard.mjs SAME · savedrole.mjs SAME · wizard.css SAME · docs/dev/file-budget.tsv SAME
    control: VERSION CHANGED · .github/workflows/ci.yml CHANGED

So every `file:line` citation in the recorded passes at `e34986bd` and `c480e766` still resolves at the
new heads. **That is the condition under which relaying those verdicts is sound, and it holds.** No
loud finding here — the opposite of the one you asked me to look for.

## (a) Check-runs at the new head — read from `/commits/<sha>/check-runs`, not `/status`

**`total_count` 23 at each: 20 success, 2 skipped (the CVE sweep, by design), 1 `in_progress`
(`Shell tests (shellcheck + pithead suite)`). ZERO failing at either head.**

**`Dashboard tests (pytest + coverage)` is `success` at both.** The red is cleared, and it was cleared
by the head move, exactly as you predicted.

`develop` requires 15 contexts. Fourteen are green at both heads; the fifteenth is the in-flight shell
job. Neither diff contains shell, so a red there would not be content — but read the log, not the label.

**`mergeable_state` is `blocked` at both, and that is NOT an unmet check.** Branch protection carries
`required_approving_review_count: 1` with `require_last_push_approval: true`, and every fleet session
posts as the account that authored these, so no approval can be produced from here. `enforce_admins`
is `false`, so the admin merge is the path. `strict: true` is satisfied — both merge-bases *are* the
`develop` tip.

## (b) Patch-coverage — it runs and grades against the right base, and reports NOT APPLICABLE

**⚠️ CONVERGENCE, NOT CORROBORATION — and the timing decides which.** I derived what follows from the
job logs and `scripts/patch-coverage.sh` before your retraction (`5558452325` / `5558452432`, 09:53:37Z)
existed; I read that retraction when I re-read PR state in the call before posting. **Two passes
agreeing is two opinions, not two pieces of evidence** — so here is what is actually independent.
Your route was the script header plus the now-green log. Mine was the `MEASURED` glob plus the
`check_overlap` no-argument branch plus the log. **Same conclusion, and I would have posted it as a
finding.** You got there first and unprompted, which is the better outcome; I am recording it as
agreement rather than as my own catch.

The gate no longer dies. Both job logs, at the new heads:

    patch coverage: grading changed lines against origin/develop.
    No lines with coverage information in this diff.
    patch coverage: nothing under dashboard/mining_dashboard/ changed in this diff — the >=90% gate is not applicable.

The fetch resolves and it grades against the **right** base. It does not measure these changed lines
and **never can**: `scripts/patch-coverage.sh` sets `MEASURED='dashboard/mining_dashboard/*.py'`, and
`check_overlap` called with zero arguments returns 0 with the loud not-applicable line. Neither diff
contains a line of Python. **The >=90% bar is NOT APPLICABLE — not "unverified pending a green".**
Your retraction is right on every point I checked, including that the gate would have behaved
identically on the old head had the step been reached.

**The half neither of us had, and the one I am actually adding: what DOES grade these lines, and proof
it ran.** That is `Frontend logic tests (node --test)` — green at both heads. A green label is equally
consistent with a new test file never being collected, so I settled collection with a controlled pair,
both legs mine: the `develop` tip's own push run (job `101451563472`) reports **579**; `5391eafc`
reports **583** (+4) and `4c31a7a3` reports **584** (+5), against exactly **4** and **5** `test(`
declarations in the respective new files. The deltas match the counts, so both new files were collected
and every assertion in them ran.

Two smaller measurements while I was in there:

- Coverage at the new heads is **97.59%, 2572 passed** — mine, not your relayed 97.57%. The base move
  changed `dashboard/pyproject.toml` and `uv.lock`, so your figure moved under you. Nothing rides on it,
  but it was your number.
- Not yours, not blocking, and worth an issue rather than a fix here: that gate's not-applicable
  message says *"nothing under `dashboard/mining_dashboard/` changed"*, which is **false for #1893** —
  three files under that directory changed, just no measured Python. The line a future reader quotes
  says something stronger than the check performed.

## (c) The two PRs do not collide — independently measured, and your figures hold

`git merge-tree --write-tree 5391eafc 4c31a7a3` -> **rc 0**, tree `ce80cfdd`, and **zero stage
entries** (I filtered the `Auto-merging` lines out rather than grepping the output for a path — a match
count is not a match). Read out of that written tree: **`wizard.mjs` 885**, `wizard.css` 94,
`savedrole.mjs` 73. **Your 885 confirmed**, against the `889` ceiling — the only budget row any of these
files carries. Four lines of headroom in the combined tree. `wizard.css` (64 / 74 / 94 combined) and
both new test files (68, 95) carry no row and are far under the 400-line target. Either merge order works.

## What I did NOT re-derive, and what carries by transfer rather than by re-reading

**This is a delta pass. I did not re-review the CSS, the mechanism, or the four assertions.** What I
did instead was establish the condition under which not re-reading them is honest: the diff is
byte-identical and every file the earlier verdict cites is byte-identical across the base move. Under
that condition the reasoning in `5556542123` and `5556646670` transfers whole.

**Which means the non-blocking findings transfer too, unaddressed — because a byte-identical diff means
none was taken.** Still live at this head, all recorded as non-blocking by the reviewer who found them:
the `wizard.css:23-24` sentence that reads as if `9330cce6` browser-verified this exact construction
when its two levers sat on different elements; and the note that `min-width: 0`'s inertness rests on
those boxes not being grid items, not on their being inline. Neither is mine to re-litigate. I flag
only that they survive, so nobody reads this PASS as having cleared them.

**RELAYED, NOT MEASURED BY ME** — say so before quoting any of it:
- That the re-run replays the frozen payload and fails identically (job `101463953714`, attempt 2). I
  re-ran nothing; I read the current heads only. Consistent with the fetch failing on a deleted ref and
  with the fix requiring a new head, but I did not observe it.
- The mutation battery, its digest guard and its restores — not in the diff, not readable.
- The local `make`/lint rc 0s in your comments.
- The substance of the two earlier code verdicts (see above for the condition that makes that sound).

**PROVEN BY ME:** both patch-ids at all four heads; the byte-identical diff-of-diffs; blob identity at
both heads with a firing control; the base-move file set and its empty intersection with both the
changed paths and the cited files, with a firing control; the 23 check-runs and their conclusions at
this head; the 15 required contexts and the approval settings; the patch-coverage log lines and the
`MEASURED` glob that explains them; the 579/583/584 frontend triple and the `test(` counts; the
merge-tree and every line count in §(c); the `closingIssuesReferences` result and its 8-PR control.

## MERGE POSTURE

**PASS at `5391eafc`. Nothing blocking, one check outstanding.** At my last read: 23 check-runs, 20
success, 2 skipped, **zero failing**, and `Shell tests (shellcheck + pithead suite)` still running — the
fifteenth required context. Merge when it lands green. If it reds, read the log rather than the label:
this diff contains no shell. (For reference, the same job has already completed **green** at #1893's
`4c31a7a3`, which shares this base.)

**Not mine to merge and not a lane's.** `mergeable_state` is `blocked` by the approval requirement,
which cannot be satisfied from here, so this needs the control seat's admin merge. **Do not hand-close
#1879** — the `user-select: all` half stays open by your design and the earlier reviewer's agreement.

## METHOD

**CI-only rule obeyed.** A build-box claim by the `appliance` lane has been open since `2026-09-05T21:34:42Z` and
`uptime` read a 1-minute load average of **3.70**, above the rule's bar of 3. **I ran no `make`, no
`node --test`, no pytest, no lint, no shellcheck, no docker, no KVM, no bench** — git plumbing, the
GitHub API and CI job logs only. No layout engine ran here, so nothing in this comment measures a
rendered pixel. I am an ephemeral reviewer; I am not the author of either PR.

VijitSingh97:
**CI is now fully resolved at `5391eafc` — the one outstanding condition in my pass above is met.**

`Shell tests (shellcheck + pithead suite)` has since completed **success**. Re-read at source:
**23 check-runs, 21 success, 2 skipped, ZERO failing**, and I enumerated all 15 required contexts on
`develop` against the success set — **every one is green**. Same at #1893's `4c31a7a3`.

So the merge condition on this PR reduces to nothing. **`mergeable_state` is still `blocked`, and that
is the approval requirement, not a check** — `required_approving_review_count: 1` with
`require_last_push_approval: true`, unsatisfiable from an account that authored the PR.
`enforce_admins: false`, so the control seat's admin merge is the path. **Do not hand-close #1879.**

Head re-read in this same call and unchanged: `5391eafcac4ac0ff046980939411e80bb935fe36`.

Issue #1868, opened 2026-09-05
Wizard: buttons are browser defaults (Apply fails WCAG 2.5.8 target size), the shell's 'Loading…' placeholder survives every stage, and section headings jam against the field above


VijitSingh97:
The setup wizard's buttons render as the browser's default controls, its shell placeholder stays on the page through every stage, and its section headings sit hard against the field above them. Rendered from `develop-v2` with the real wizard server (`mining_dashboard/wizard.py`, a scratch spool) through all five stages: gate, setup, installer, installing, handoff, in both themes at 1400 and 390 px.

**Buttons.** `Continue` (gate), `Apply` (setup) and `I saved these — start provisioning` (handoff) carry no class. `wizard.css` opens by saying buttons come from `dashboard.css`, but that file has no bare `button` rule, only `.btn-range`, `.btn-toggle`, `.upgrade-btn` and friends, so the wizard's buttons fall through to the user-agent style: white fill, black 13 px text, 2 px padding. axe-core 4.10 flags `Apply` under `target-size` (serious): the box is under 24 px tall, WCAG 2.2 SC 2.5.8. The handoff button stretches to the card width in the same default style. It is the first screen anyone sees, and it does not look like the dashboard it hands over to.

**"Loading…" residue.** `templates/wizard.html:18` ships `<main id="app"><h1>Pithead setup</h1><p class="text-muted">Loading…</p></main>` and `wizard.mjs:888` renders the app into that same `#app`. Preact's `render` does not clear the container's pre-existing children, so the placeholder paragraph survives below the card on every stage, gate included. Proven on all twenty renders.

**Heading rhythm.** The section headings (`Payout addresses`, `Monero node`, `Tari node`, `Mining`, `First sync`, `Dashboard login`) are `<h3>` after the page `<h1>` (axe `heading-order`, one per page). Visually, a heading that follows a select with no help text (`Tari node`, `Mining`, `Dashboard login`) has no top margin, so it reads as a label for the field above it; the ones that follow a help paragraph get the paragraph's margin. The rhythm depends on what happened to precede the heading.

[...]

Issue #1878, opened 2026-09-05
Wizard: form errors are silent and land above the fold; the installer-stage restore path hides its upload fields without saying why; no passphrase show/hide


VijitSingh97:
## What happens

Three things on the setup wizard, all measured against the real wizard server on `develop-v2` (desktop 1400 px and phone 390 px, dark):

1. **A validation error is silent and lands above the fold.** Submit the restore form with no archive chosen: "Choose a backup archive to upload." appears in red at the top of the card, 248 px above the button on desktop and 343 px on a phone. Focus stays on the button; the message has no `role` and no `aria-live`. A screen reader hears nothing; a sighted user on a phone who tapped the button at the bottom of a longer page may not see it either. The same slot carries every error the wizard raises, including the host's "Restore failed — check the archive and passphrase".
2. **In the installer stage, "Restoring an existing Pithead?" leads to a page with no restore on it.** The card says "Upload an encrypted Pithead backup instead of filling in the form below", then shows only "Install onto — Target disk". The archive and passphrase fields appear after a disk is picked and its name typed, and nothing says so. The setup form covers the same disclosure with a sentence ("the rest of the page appears once you pick"); the restore path does not.
3. **The passphrase field has no show/hide.** The kit passphrase is a generated string the user types from a printout or a downloaded text file; a masked field with no reveal is where typos go to hide. Taste, one reference: the "show password" pattern in the GOV.UK Design System.

## Mechanism

- `wizard.mjs:72`: `Err` renders `<p class="c-bad">` with no role; `wizard.mjs:564` (restore) and `:643` (setup) place it above the form, before the first section heading. `submitRestore` (`:473-495`) and `submit` set `this.state.error` and do nothing else, so focus does not move.
- `wizard.mjs:575-582`: `RestoreSection` and the submit button render only when `diskPicked`; the restore card's opening paragraph at `:562` has no "appears once you pick" sentence, unlike the setup form's at `:616-618`. Screenshot: `wizard-restore-installer-nodisk-dark-desktop.png`.
- `wizard.mjs:184`: `<input type="password">` with no toggle.

Already filed and not repeated here: the unstyled submit button and the "Loading…" residue under the card (#1868), the reference placeholder flagged as an error (#1869).

## Fix

[...]

Issue #1869, opened 2026-09-05
Wizard: the Monero address is pre-filled with the reference's placeholder string and flagged red before input; select labels truncate on a phone ('Pithead + RigFo')


VijitSingh97:
Two form defects in the setup wizard, both from real renders of `develop-v2` through the wizard server.

**The Monero address field is pre-filled with a placeholder string and flagged red before the operator does anything.** `wizard_state` merges the reference over the effective config, and the reference's documented default for `monero.wallet_address` is the literal text `your monero wallet address`. That string lands in the input as its value (not as an HTML placeholder; it renders in the input's own text colour), and `configsync.mjs:76` classifies it `not-primary`, so the page opens with `A primary Monero address starts with 4.` in red under an untouched field. The Tari field gets the same placeholder value without the message. On the appliance the operator's first act is therefore to select and delete text, and the first thing they read is an error they did not cause. Fix: treat the reference's placeholder value as empty when populating the form (or publish the reference with an empty string and keep the hint as the input's `placeholder` attribute), and show `not-primary` only after the field has been edited or on submit.

**Select labels truncate, worst on a phone.** The wizard uses native `<select>` controls at a fixed column width. At 1400 px the recommended defaults are cut mid-phrase: `Yes — this machine also mines with its own CPU (bu`, `Generate a strong password for me (recommende`. At 390 px they are unreadable: `Pithead + RigFo`, `Run the bundled`, `Yes — this mach`, `Generate a stro`. Every one of these questions has two to four answers. Fix: render questions with at most four answers as a radio group (each answer on its own line, full width, the description under it), which also removes the need to open a menu to see what the options are. Keep `<select>` only for the disk picker, whose rows are data.

Filed from the design review (the umbrella issue links here). Milestone is my call from the manual-test round; the seat may move it.

## Worker brief (inventory 2026-09-13)

**Rank 19.** Two wizard form defects, both from real renders. The first one means the operator's
first act on a brand-new appliance is to delete placeholder text, and the first thing they read is
an error they did not cause.

### Acceptance criteria

- [ ] The reference's documented placeholder for `monero.wallet_address` (the literal string
      `your monero wallet address`) does not land in the input as its **value**. Either treat the
      reference placeholder as empty when populating the form, or publish the reference with an
      empty string and carry the hint as the input's `placeholder` attribute.
- [ ] The Tari address field gets the same treatment — it takes the same placeholder value today and
      merely happens not to show the message.
[...]

Issue #1848, opened 2026-09-05
Setup wizard: offer the XvB raffle as an on/off switch (xvb.enabled) instead of only in the Advanced view


VijitSingh97:
## Operator request (2026-09-05 manual test)

XvB (the XMRvsBeast raffle / donation tier) should be a switch in the setup wizard: on or off, at first boot, beside the other choices a new machine needs.

## Today

The wizard's curated field list (`dashboard/mining_dashboard/web/static/wizard.mjs:28-60`) carries no `xvb.*` path at all; `xvb.enabled` is reachable only through the Advanced JSON view or later from the dashboard's Configuration view. Meanwhile the dashboard has a whole XvB decision table and calculator (#1316, #1718), which is odd to meet for the first time on a machine that is already mining into the raffle with the default (`xvb.enabled` defaults true in `config.reference.json`).

## Shape

- One field under the mining section: **"Join the XMRvsBeast raffle?"** bound to `xvb.enabled`, default from the reference, with a one-sentence Note that says what it does to hashrate and payouts (the docs already have the sentence — reuse it, do not invent numbers).
- When off, the dependent `xvb.*` keys are left at defaults (nothing else to ask). When on, no further question in the wizard: the donor id and URL keep their defaults and are editable later in the dashboard, as today.
- Frontend test: the field renders, toggles `xvb.enabled` in the submitted config, and `strip_defaults` drops it when it equals the reference default.

## Scope

Wizard image bytes (`dashboard/`). Small. Rides the current freeze move only if it is ready before the freeze tip; otherwise the first post-GA update. Not a defect.

VijitSingh97:
Operator, RC1 manual test: confirmed there is no XvB enable/disable in the wizard today. Also see the xvb.enabled default inconsistency filed alongside — the wizard writing the key explicitly is half of that fix.

VijitSingh97:
Completed in #1970 (merge `c79859762ea5f35eeecec891c2b1de8c6421af3b`).

Current `develop` exposes “Join the XMRvsBeast raffle?” during first-boot setup, defaults from `xvb.enabled`, writes a real boolean `false` when declined, keeps the no-JavaScript fallback consistent, and documents the choice. Fresh checks at `4d2f4fcdd357bedb8ac26e666be9230012161d42` passed: 15/15 focused frontend tests, 21/21 focused Python wizard tests, the XvB strip-defaults check, file budget, JS lint, Python lint, and Python format.

[...]

PR #1893, opened 2026-09-06
fix(wizard): #1868 the wizard's buttons look like the product, and the shell stops leaking

VijitSingh97:
Addresses #1868 and deliberately does NOT close it — one of the issue's four asks is left for the design lane, see **What I did not do**. No closing keyword is used, which now matters: the base is `develop`, the repository's default branch, where a closing keyword FIRES on merge. Confirmed with GitHub's own link resolution (`closingIssuesReferences`), which returns an empty list for this PR.

The setup wizard's buttons look like the dashboard they hand over to instead of like raw browser controls; the "Loading…" line and a duplicate page heading stop sitting under the card on every stage; and a section heading is no longer jammed against the field above it.

## The three fixes

**Buttons.** `Continue`, `Apply`, the two handoff acknowledgements and the restore submit carried no class at all, and `dashboard.css` has no bare `button` rule, so they fell through to the user-agent style. axe-core flagged `Apply` under `target-size` (WCAG 2.2 SC 2.5.8).

They now carry `.btn-toggle` / `.btn-toggle.active` — the dashboard's own classes, and what the config view's confirm modal already uses for a standalone primary action (`configview.mjs:166-169`). That class is built for a segmented control (`border: none`, radius inherited from its parent), so `wizard.css` makes the same adaptation the modal makes at `.config-modal-actions .btn-toggle` (`dashboard.css:1460-1468`): a border, a radius, a disabled rule. `min-height: 32px` states the target-size bar directly rather than leaving it to be derived from padding plus line-height.

**⚠️ Correction to the issue, worth reading before reviewing:** the issue proposes sharing "the dashboard's `.btn-primary`-style controls (the config view's Apply already has one)". **There is no `.btn-primary` anywhere in `dashboard/mining_dashboard/web/`** — I grepped. The precedent is real but it is `.btn-toggle active`, which is what this uses. Copying `.btn-toggle`'s declarations into `wizard.css` instead would have re-spelled a skin that then drifts from the one it copies.

**The shell's residue.** `templates/wizard.html:18` ships the heading and `Loading…` inside `<main id="app">` *deliberately* — so a `curl` and the tier-4 harness can recognize the page without executing the module. `preact`'s `render` appends to its container, so both survived every stage. Since the app renders its own `<h1>`, the page was also carrying **two** `<h1>Pithead setup</h1>` — which the issue does not name. The mount now clears `#app` first, fixing both. The server-side markup is untouched, so what `curl` sees does not change.

**Heading rhythm.** `h3` carries `margin: 0 0 15px` from `dashboard.css:261`, so the space above a section heading was whatever the preceding help paragraph happened to leave; headings following a bare select (`Tari node`, `Mining`, `Dashboard login`) got none. A top margin fixes it, with a `:first-child` reset so a heading opening its own card is unchanged.

## What I did NOT do

**The issue's fourth ask — retag the section headings `<h3>` → `<h2>` for axe's `heading-order` — is not here, on purpose.** In this stylesheet `h2` is the 1.5rem card-title style (`dashboard.css:256`) and `h3` is the uppercase, muted, bottom-ruled section style (`:261`). Retagging as written would restyle all 13 wizard section headings. Preserving the look means duplicating the `h3` rule under a wizard-scoped `h2` — the same re-spelling avoided in the button fix. Choosing between that, an `aria-level`, and a new shared section-heading class is a design decision, not mine to guess; **#1868 stays open for it.**

Also not done: no browser or screenshot proof (no layout engine on this box), and I did not re-verify the axe-core findings myself — those are the issue's, taken as filed.

## Evidence at this head

- `make test-frontend` — **571/571 pass.** Base measured at **566** on `8c7874c9`; the 5 new assertions are this PR's.
- Ten lint targets rc 0 each: `lint-js lint-md lint-py lint-file-budget lint-topology lint-docs-voice lint-operator-strings lint-yaml lint-toml lint-proto`. `lint-file-budget` runs its own self-tests and they passed.
- **FIRED CONTROL.** All five new assertions were run against the unfixed files (`git checkout 8c7874c9 -- wizard.mjs wizard.css`) and **all five failed**; all five pass restored. The fix was committed *before* the control ran, so the restoring `checkout` could not discard it. Worktree confirmed clean afterwards.
- `make test-dashboard` — reported in a follow-up comment; this diff contains no Python.

**What the tests do not prove:** they are static. They parse the files and read what is declared; nothing here renders a page, so no assertion proves a button measured 32 px on screen. They hold the property that actually broke — a control with no class, and a mount that is never cleared.

## Budget

`wizard.mjs` lands at **885 against its 889 ceiling — four lines of headroom**, re-derived at this head with the gate's own counter (`awk 'END{print NR+0}'`): base `0a4195c9` 884, head `4c31a7a3` 885. The button classes are added to existing lines; the only new line is the `replaceChildren()` call, paid for by compressing the comment above it. `wizard.css` 44 → 74, `savedrole.mjs` 73 (unchanged count) and the new test file (95) carry no budget row.

**CORRECTED 2026-09-06, in place.** This paragraph previously read *"889 against its 889 ceiling — the file is pinned and any further edit must be line-neutral until pull 1884 merges, which takes it to 884"*, and called the new test file 76 lines. Pull 1884 **has merged** (`c3065010`, an ancestor of the current base), so the figure was stale and **the line-neutral instruction it carried was false** — a wrong number with a live imperative attached is the part that could have cost someone a wasted edit. The test file is 95 lines, not 76.

## Lane notes

`dashboard/` only, in this lane's `.paths`; no `.lane-override` needed or touched, no `_shared.paths` file edited. Coordinated with the `dashboard` lane, who took #1869 and #1878 into their #1848/#1854. Pull 1884 has merged, so that half of the earlier sequencing note is spent.

⛔ **SEQUENCING, MEASURED — this branch and pull 1927 CONFLICT, and neither branch can see it.** Both edit `dashboard/mining_dashboard/web/static/savedrole.mjs`: this PR adds `class="btn-toggle active"` to the *Keep it* button, pull 1927 rewrites the same screen's buttons for the restore door. `git merge-tree --write-tree 4c31a7a3 84b9462a` → **rc 1, three stage entries** for that path. **Control, same instrument, same session:** `git merge-tree --write-tree 4c31a7a3 5391eafc` (this branch × pull 1885) → **rc 0, tree `ce80cfdd`, zero stage entries** — so the rc 1 is a real conflict, not the tool always disagreeing. **Whichever of 1893 and 1927 merges second must rebase before it is merged; they must never be merged back to back unread.** Order is otherwise free.

Never merged by its author.

### DIFF
--- a/dashboard/mining_dashboard/web/static/savedrole.mjs
+++ b/dashboard/mining_dashboard/web/static/savedrole.mjs
@@ -47,7 +47,7 @@ export const SavedRoleScreen = ({ summary, kept, error, onKeep, onSetUpAgain }) 
                 html`<${Field} label=${r.label}><code class="wizard-mono">${r.value}</code><//>`,
             )}
             <${Err}>${error}<//>
-            <button type="button" onClick=${onKeep}>Keep it</button>
+            <button type="button" class="btn-toggle active" onClick=${onKeep}>Keep it</button>
             <button type="button" class="wizard-link" onClick=${onSetUpAgain}>Set up again</button>
             <${Note}>Setting it up again asks the same questions as a first boot, with this
             machine's answers already filled in. Its login and other secrets are never filled
--- a/dashboard/mining_dashboard/web/static/wizard.css
+++ b/dashboard/mining_dashboard/web/static/wizard.css
@@ -42,3 +42,33 @@
   text-decoration: underline;
   cursor: pointer;
 }
+
+/* The wizard's buttons take the dashboard's own skin (.btn-toggle / .btn-toggle.active) so the
+ * first page anyone sees matches the app it hands over to. That class is built for a segmented
+ * control — no border of its own, radius from its parent — so a standalone button needs the same
+ * adaptation the config modal already makes at `.config-modal-actions .btn-toggle`. Without any of
+ * it these fell through to the user-agent style, and Apply failed WCAG 2.5.8 target size at under
+ * 24px tall (#1868). min-height states that bar directly rather than leaving it to font metrics.
+ * `.wizard-link` is a button deliberately styled as text and keeps its own look. */
+.wizard-shell button:not(.wizard-link) {
+  border: 1px solid var(--border);
+  border-radius: 6px;
+  min-height: 32px;
+}
+
+.wizard-shell button:not(.wizard-link):disabled {
+  opacity: 0.5;
+  cursor: not-allowed;
+}
+
+/* A section heading sat hard against the field above it and read as that field's label: h3 carries
+ * `margin: 0 0 15px` from dashboard.css, so the space above it was whatever the preceding help
+ * paragraph happened to leave — headings after a bare select got none. A heading that opens its
+ * own card still needs zero, which is what the :first-child reset is for (#1868). */
+.wizard-shell h3 {
+  margin-top: 22px;
+}
+
+.wizard-shell h3:first-child {
+  margin-top: 0;
+}
--- a/dashboard/mining_dashboard/web/static/wizard.mjs
+++ b/dashboard/mining_dashboard/web/static/wizard.mjs
@@ -80,7 +80,7 @@ export const Gate = ({ error, onSubmit }) => html`<div class="card">
                 spellcheck=${false} placeholder="pit-XXXXXX" />
         <//>
         <${Note}>Case doesn't matter, and the ${" "}<code>pit-</code>${" "}prefix is optional.<//>
-        <button type="submit">Continue</button>
+        <button type="submit" class="btn-toggle active">Continue</button>
     </form>
 </div>`;
 
@@ -206,7 +206,7 @@ export const Done = ({ status, handoff, installer, stick, rig, onAck }) => html`
             <p>This is what the machine will be.</p>
             ${rigCardFields(handoff).map((f) => html`<${Field} label=${f.label}><code class="wizard-mono">${f.value}</code><//>`)}
             <${Note}>${rigCardNote(handoff)}<//>
-            <button type="button" onClick=${onAck}>
+            <button type="button" class="btn-toggle active" onClick=${onAck}>
                 ${installer && !stick ? "Looks right — erase the disk and install" : "Looks right — save it"}</button>`
         : handoff
           ? html`<h3>Save this before anything else</h3>
@@ -215,7 +215,7 @@ export const Done = ({ status, handoff, installer, stick, rig, onAck }) => html`
             <${Field} label="Dashboard password"><code class="wizard-mono">${handoff.password}</code><//>
             <${Field} label="Dashboard address"><code class="wizard-mono">${handoff.dashboard}</code><//>
             <${Field} label="Point miners at"><code class="wizard-mono">${handoff.stratum}</code><//>
-            <button type="button" onClick=${onAck}>
+            <button type="button" class="btn-toggle active" onClick=${onAck}>
                 ${installer ? "I saved these — erase the disk and install" : "I saved these — start provisioning"}</button>
             <${Note}>${
               installer
@@ -571,7 +571,7 @@ export class WizardApp extends Component {
               html`<${RestoreSection} file=${this.state.restoreFile} passphrase=${restorePassphrase}
                 onFile=${(e) => this.setState({ restoreFile: e.target.files[0] || null })}
                 onPassphrase=${(e) => this.setState({ restorePassphrase: e.target.value })} />
-            <button type="submit" disabled=${submitting}>
+            <button type="submit" class="btn-toggle active" disabled=${submitting}>
                 ${submitting ? "Validating…" : "Restore and provision"}</button>`
             }
         </form>
@@ -846,7 +846,7 @@ export class WizardApp extends Component {
 
             ${
               diskPicked &&
-              html`<button type="submit" disabled=${(!rig && !!jsonError) || this.state.submitting}>
+              html`<button type="submit" class="btn-toggle active" disabled=${(!rig && !!jsonError) || this.state.submitting}>
                 ${
                   this.state.submitting
                     ? "Validating…"
@@ -877,8 +877,9 @@ export class WizardApp extends Component {
   }
 }
 
-// Mount only in a browser: node --test imports this module to render-probe the views, and a
-// bare `document` reference at import time would make the whole file untestable.
+// Mount only in a browser (node --test imports this module; a bare `document` would break that).
+// Clear #app: the shell ships the heading and "Loading…" inside it, and preact APPENDS (#1868).
 if (typeof document !== "undefined") {
+  document.getElementById("app").replaceChildren();
   render(html`<${WizardApp} />`, document.getElementById("app"));
 }
--- a/dashboard/tests/frontend/wizardshell.test.mjs
+++ b/dashboard/tests/frontend/wizardshell.test.mjs
@@ -7,7 +7,7 @@
 // A new file rather than wizard.test.mjs, which is at 765 against a 765 ceiling.
 
 import assert from "node:assert/strict";
-import { readFileSync } from "node:fs";
+import { readdirSync, readFileSync } from "node:fs";
 import test from "node:test";
 
 const read = (p) => readFileSync(new URL(p, import.meta.url), "utf8");
@@ -20,14 +20,36 @@ function ruleFor(css, selectorRe) {
   return m ? { selector: m[2].trim(), body: m[3] } : null;
 }
 
-test("wizard.mjs: every button declares a class, so none falls through to the browser default", () => {
-  // Each opening tag, taken as the text after `<button` up to a comfortable bound: the tag itself
-  // can hold arrow functions, so scanning for the closing `>` would stop at the first `=>`.
-  const tags = WIZARD_MJS.split("<button").slice(1).map((s) => s.slice(0, 200));
-  assert.ok(tags.length >= 7, `expected the wizard's buttons, found ${tags.length}`);
-  const bare = tags.filter((t) => !/class="(btn-toggle[^"]*|wizard-link)"/.test(t));
+// The wizard's view modules, found by the import they share rather than by a hand-kept list — a new
+// view that renders a button is then covered the day it is written. The first sweep here read only
+// wizard.mjs and was blind to savedrole.mjs's "Keep it", which is exactly this defect in a file the
+// needle never looked at.
+const STATIC = new URL("../../mining_dashboard/web/static/", import.meta.url);
+const VIEWS = readdirSync(STATIC)
+  .filter((f) => f.endsWith(".mjs"))
+  .map((f) => [f, readFileSync(new URL(f, STATIC), "utf8")])
+  .filter(([f, src]) => f === "wizard.mjs" || src.includes("./wizardparts.mjs"));
+
+test("every wizard view: a button declares a class, so none falls through to the browser default", () => {
+  // An enumeration that quietly found nothing would pass this test while proving nothing, so the
+  // root is asserted before the property is.
+  assert.ok(VIEWS.length >= 2, `expected the wizard's view modules, found ${VIEWS.length}`);
+  assert.ok(
+    VIEWS.some(([f]) => f === "savedrole.mjs"),
+    "savedrole.mjs no longer matches the wizard-view marker — widen the root, do not narrow the claim",
+  );
+  const bare = [];
+  for (const [file, src] of VIEWS) {
+    // Each opening tag, taken as the text after `<button` up to a comfortable bound: the tag can
+    // hold arrow functions, so scanning for the closing `>` would stop at the first `=>`.
+    for (const tag of src.split("<button").slice(1).map((s) => s.slice(0, 200))) {
+      if (!/class="(btn-toggle[^"]*|wizard-link)"/.test(tag)) {
+        bare.push(`${file}: <button${tag.split("\n")[0].trim()}`);
+      }
+    }
+  }
   assert.deepEqual(
-    bare.map((t) => t.split("\n")[0].trim()),
+    bare,
     [],
     "a button with no class renders as the user-agent control (#1868): give it the dashboard's " +
       "btn-toggle skin, or wizard-link if it is deliberately a text link",
--- a/dashboard/tests/frontend/wizardshell.test.mjs
+++ b/dashboard/tests/frontend/wizardshell.test.mjs
@@ -1,0 +1,73 @@
+// The wizard's shell and chrome (#1868), asserted the way workerview.test.mjs asserts dashboard.css:
+// parse the files and read what they declare. These are static assertions — no layout engine runs
+// here, so nothing below proves a rendered button measured 32 px. What they DO hold is the property
+// that broke: a control that carries no class at all falls through to the user-agent style, and a
+// mount that is never cleared keeps whatever the server shipped inside it.
+//
+// A new file rather than wizard.test.mjs, which is at 765 against a 765 ceiling.
+
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import test from "node:test";
+
+const read = (p) => readFileSync(new URL(p, import.meta.url), "utf8");
+const WIZARD_MJS = read("../../mining_dashboard/web/static/wizard.mjs");
+const WIZARD_CSS = read("../../mining_dashboard/web/static/wizard.css");
+const WIZARD_HTML = read("../../mining_dashboard/web/templates/wizard.html");
+
+function ruleFor(css, selectorRe) {
+  const m = css.match(new RegExp(`(^|\\})\\s*(${selectorRe.source})\\s*\\{([^}]*)\\}`, "m"));
+  return m ? { selector: m[2].trim(), body: m[3] } : null;
+}
+
+test("wizard.mjs: every button declares a class, so none falls through to the browser default", () => {
+  // Each opening tag, taken as the text after `<button` up to a comfortable bound: the tag itself
+  // can hold arrow functions, so scanning for the closing `>` would stop at the first `=>`.
+  const tags = WIZARD_MJS.split("<button").slice(1).map((s) => s.slice(0, 200));
+  assert.ok(tags.length >= 7, `expected the wizard's buttons, found ${tags.length}`);
+  const bare = tags.filter((t) => !/class="(btn-toggle[^"]*|wizard-link)"/.test(t));
+  assert.deepEqual(
+    bare.map((t) => t.split("\n")[0].trim()),
+    [],
+    "a button with no class renders as the user-agent control (#1868): give it the dashboard's " +
+      "btn-toggle skin, or wizard-link if it is deliberately a text link",
+  );
+});
+
+test("wizard.css: a standalone button gets the border and radius .btn-toggle takes from its parent", () => {
+  const rule = ruleFor(WIZARD_CSS, /\.wizard-shell\s+button:not\(\.wizard-link\)/);
+  assert.ok(rule, "expected a `.wizard-shell button:not(.wizard-link)` rule in wizard.css");
+  assert.match(rule.body, /border:\s*1px solid var\(--border\)/);
+  assert.match(rule.body, /border-radius:/);
+});
+
+test("wizard.css: the button's target size clears WCAG 2.2 SC 2.5.8 (24px) as a declared floor", () => {
+  const rule = ruleFor(WIZARD_CSS, /\.wizard-shell\s+button:not\(\.wizard-link\)/);
+  const mh = rule.body.match(/min-height:\s*(\d+)px/);
+  assert.ok(mh, "expected an explicit min-height rather than one implied by padding + line-height");
+  assert.ok(
+    Number(mh[1]) >= 24,
+    `min-height ${mh[1]}px is below the 24px target-size floor (SC 2.5.8)`,
+  );
+});
+
+test("wizard.mjs: the mount clears #app, which the shell deliberately ships non-empty", () => {
+  // The two halves are a pair: the template ships the heading and "Loading…" inside #app so a curl
+  // can recognize the page, and preact's render appends. Assert both, or a later edit to either one
+  // silently restores the residue.
+  assert.match(WIZARD_HTML, /<main[^>]*id="app"[^>]*>\s*<h1>/, "the shell still ships markup in #app");
+  const mount = WIZARD_MJS.slice(WIZARD_MJS.indexOf('typeof document !== "undefined"'));
+  const clearAt = mount.indexOf("replaceChildren()");
+  const renderAt = mount.indexOf("render(");
+  assert.ok(clearAt !== -1, "#app is never cleared, so the shell's markup outlives every stage");
+  assert.ok(clearAt < renderAt, "#app must be cleared BEFORE render, not after");
+});
+
+test("wizard.css: a section heading is spaced from the field above, but not when it opens a card", () => {
+  const h3 = ruleFor(WIZARD_CSS, /\.wizard-shell\s+h3/);
+  assert.ok(h3, "expected a `.wizard-shell h3` rule");
+  assert.match(h3.body, /margin-top:\s*[1-9]/, "a heading with no top margin reads as the label of the field above it");
+  const first = ruleFor(WIZARD_CSS, /\.wizard-shell\s+h3:first-child/);
+  assert.ok(first, "expected the :first-child reset");
+  assert.match(first.body, /margin-top:\s*0/);
+});


VijitSingh97:
Rebased onto `develop-v2` after pull 1884 merged. Head is now `c480e766`; two corrections to the body above, one of them a defect the rebase exposed.

## ⛔ The rebase found a site my sweep was blind to

1884 brought `savedrole.mjs` into the tree, and its **"Keep it" button carries no class** — the same defect this PR fixes, in a file the original test never looked at. "Set up again" beside it was already `.wizard-link`.

The button is now `.btn-toggle active`. **The test is the real fix:** its root was `wizard.mjs` alone, so it passed over a covered site — a sweep is only as wide as its root, and a needle that never looked cannot report an absence. Views are now enumerated by the import they share (a module is a wizard view if it imports `./wizardparts.mjs`, or is `wizard.mjs`), so a new view rendering a button is covered the day it is written rather than the day someone remembers the list. Two guards on the enumeration, because a root that quietly finds nothing would pass while proving nothing: the count must be ≥ 2, and `savedrole.mjs` must still match the marker.

**Mutation control**, run and restored by `sha256sum` (`b6cacd38` → `3e48fe9c` → `b6cacd38`): stripping the class back off "Keep it" reddens the test, and the failure names the file and the tag —

```
not ok 1 - every wizard view: a button declares a class …
  + 'savedrole.mjs: <button type="button" onClick=${onKeep}>Keep it</button>'
```

Without that, the green would only show the file was being read.

## ⛔ Correction: the budget claim in the body is stale

The body says `wizard.mjs` "lands at 889 against its 889 ceiling — the file is pinned". **That was true at the old base and is now false.** 1884 took the file to 884, so this branch lands at **885 / 889** — four lines of headroom, not zero, and the "any further edit must be line-neutral" caveat no longer applies. Correcting it here because a stale budget figure with an imperative attached is the kind that tells the next person the wrong thing to do.

## Re-run at `c480e766`

- `make test-frontend` — **584/584 pass.** Base moved to 579 when 1884 merged; 5 are this PR's.
- `lint-js lint-md lint-file-budget lint-topology lint-docs-voice lint-operator-strings` — rc 0 each.
- Replayed file set re-read by hand after the rebase, since `rebase --continue` does not run `pre-commit` and the lane-guard never fires on a replayed commit: exactly 3 files (`wizard.css`, `wizard.mjs` modified; `wizardshell.test.mjs` added) plus `savedrole.mjs` from this follow-up. `git diff --summary` shows no mode changes.
- `make test-dashboard` — **2548 passed, exit 0, 97.44%** — but measured at `c705d0fc`, the pre-rebase head, so it does **not** cover this one. This diff still contains no Python; stating the gap rather than letting the figure imply coverage it does not have.

Everything else in the body stands, including the heading-order retag being left for the design lane, which is why this does not close #1868.

VijitSingh97:

VijitSingh97:
# NON-AUTHOR PASS — pithead #1893 (#1868), head `c480e766`

**VERDICT: PASS.** The skin reuse is right, the `.btn-primary` correction is right and saved a
re-spelled skin, the mount fix is a real defect correctly diagnosed, and the deferral in §1 is a
deferral rather than a dodge. Three findings, none blocking; the first is the one I would actually
change, because it is a mechanism claim that does not survive re-derivation.

Head re-derived by me in the call that measured: `c480e766`. **Base is no longer `develop-v2` — it
is `develop`.** `git ls-remote origin refs/heads/develop-v2` now returns nothing and
`refs/heads/develop` is `c3065010`, this branch's exact merge-base. Your fast-forward comment checks
out at source; I did not take it from the comment.

---

## 1. THE h3 -> h2 DEFERRAL — RIGHT, and it is not hiding. Verified at source

`dashboard.css:256` — `h2 { margin: 0; font-size: 1.5rem; font-weight: 600 }`.
`dashboard.css:261` — `h3 { margin: 0 0 15px; font-size: 0.85rem; text-transform: uppercase;
color: var(--text-muted); letter-spacing: 0.5px; border-bottom: 1px solid var(--border);
padding-bottom: 10px }`.

Both line numbers exact. Retagging as the issue words it swaps a small uppercase bottom-ruled
section header for a 1.5rem card title on **every** wizard section. That is a visual redesign
wearing an accessibility fix's clothes, and it is not yours to guess.

**The test for "deferring" versus "hiding behind it" is whether you named the alternatives and the
cost, and you did**: duplicate the `h3` rule under a wizard-scoped `h2`, or `aria-level`, or a new
shared section-heading class — with the reason each is a design call. Leaving #1868 OPEN is the
honest outcome; closing it would have been the dodge. **Right call, and I would have made it.**

One thing to add to the issue rather than the PR: after this merges the wizard still fails axe
`heading-order`, and the remaining shape is now precisely **h1 -> h3** with nothing between it —
cleaner than before your mount fix, which removed the duplicate `<h1>`. Naming that shape saves the
design lane a re-derivation.

## 2. ⛔ THE FINDING — `min-height: 32px` does not mean 32px, and the comment says it does

This is the answer to your ask 3, and it has more in it than the WCAG half.

**There is no `box-sizing: border-box` on buttons.** `dashboard.css` has no universal reset — I
grepped for one — and the only three `box-sizing` declarations across both stylesheets are
`.worker-inspect` (`dashboard.css:1549`), `.worker-edit` (`:1560`) and `.wizard-json`
(`wizard.css:20`). `dashboard.css:157`'s own comment states content-box is the house default.

So `min-height: 32px` floors the **content** box. With `.btn-toggle`'s `padding: 6px 16px`
(`dashboard.css:938`) and this PR's new `border: 1px solid`, the rendered border-box height is
**>= 46px**, not 32. Your comment at `wizard.css:50-51` says the bar is 24px and that `min-height`
"states that bar directly rather than leaving it to font metrics" — the stated bar is 24, the
declared number is 32, and the rendered result is ~46. Three different numbers, one of them
presented as the other.

It also cuts against this PR's own stated goal. The dashboard's `.config-actions .btn-toggle` lands
around 30px on the same metrics; the wizard's will be about half again taller than the app it is
supposed to look like.

**Not blocking**, and the outcome may well be better than the bar — 46px is a good phone target. But
it is a mechanism claim that does not survive re-derivation, which is the same class you corrected
on #1885 an hour ago. Two ways out, both one line: add `box-sizing: border-box` to the rule so 32
means 32, or say in the comment that 32 is a content-box floor and the rendered target is ~46px.

**On the WCAG claim itself.** SC 2.5.8 is 24x24 CSS px — **two** dimensions. `min-height` constrains
one. Width is never near 24px here because every button carries a long text label, and the axe
finding you cite was a height failure, so the fix does address the measured defect. But the test's
name — "the button's target size clears WCAG 2.2 SC 2.5.8 (24px)" — claims the whole SC from a
height-only assertion. Rename it to name the dimension; it costs nothing and stops the next reader
quoting it as a full 2.5.8 proof.

**And say which targets the claim does NOT cover.** `:not(.wizard-link)` excludes three buttons, and
two of them are standalone rather than in a sentence: `savedrole.mjs:51` ("Set up again", sitting
directly beside the "Keep it" you just classed) and `wizard.mjs:578` ("Back to the setup form").
`.wizard-link` is `padding: 0; font: inherit` (`wizard.css:36-44`), so both are well under 24px tall
and SC 2.5.8's *Inline* exception is doubtful for either. The third, `wizard.mjs:635`, is inside a
`<p>` and plausibly does qualify. I am not asking you to restyle them in this PR — the scope is the
primary actions — but a PR that makes a target-size claim should say where the claim stops. That
belongs on #1868 with the retag.

## 3. THE ENUMERATION ROOT — complete today, and a convention rather than a proof

You asked me to check the root. **Today it is complete, and I verified that rather than assuming it.**
`git grep wizardparts` at head returns exactly `savedrole.mjs:10` and `wizard.mjs:22` in `static/`.
Everything else `wizard.mjs` imports I checked for `<button` individually:
`configlogic` 0, `configsync` 0, `rigcardlogic` 0, `wizardparts` 0, `preact` 0. And `RestoreSection`
— the one component that looks like it might live elsewhere — is defined at `wizard.mjs:169`, inside
the file the root already reads. **No wizard-rendered button is outside the root at this head.**

**The gap is in what kind of thing the marker is.** "Imports `./wizardparts.mjs`" is a naming
convention. A future wizard view that imports only `./preact.mjs` and renders its own buttons —
entirely natural for a view needing no `Note`/`Field`/`Err` — is invisible to this root, **and both
guards stay green while it is**: the count is still >= 2, and `savedrole.mjs` still matches. Your two
guards defend against the marker BREAKING for a view you already know about. Neither can defend
against a new view that never had it. That is the same shape as the defect this PR just paid for
once: `savedrole.mjs` was in the tree and the root did not look at it.

**The sound root is the import graph**: start at `wizard.mjs`, follow relative `./*.mjs` imports
transitively, take the closure. That is derived from what the wizard actually mounts instead of from
a convention someone has to remember. **Today it yields the identical set** — which is exactly why
it is worth swapping now, while the change is provably behaviour-preserving and the test's green
does not move. Not blocking; the current green is honest, because the marker's failure mode is a
false negative only.

Two smaller things in the same test:

- **The `<button` scanner can false-PASS.** `src.split("<button").slice(1).map(s => s.slice(0, 200))`
  then tests `/class="(btn-toggle[^"]*|wizard-link)"/` against the whole 200-char window — which
  includes the button's **content**, not just its tag. A bare `<button>` whose label markup contains
  a nested `class="btn-toggle…"` or `class="wizard-link"` passes. No such site exists today, and the
  `split` correctly stops each chunk before the next `<button`, so this is a nit — but it is the
  direction that hides a defect rather than the direction that shouts. Anchoring the regex to the
  start of the chunk would tighten it. (A computed `class=${...}` false-FAILS, which is the safe way
  round.)
- `src.includes("./wizardparts.mjs")` is a raw substring over the whole file, so a *comment*
  mentioning the path matches — your own test file's `:31` is such a line. Harmless (it is not in
  `static/`), but it means the marker is text, not an import.

**Your mutation control is the right one and I take it as run**: stripping the class off "Keep it"
reddens the test and the failure names the file and the tag, so the enumeration has been shown able
to say the other thing. That is the control an all-rows-agree enumeration needs.

## 4. THE MOUNT FIX — correct, and the test around it is tight

`preact`'s `render` appends to its container, so `wizard.html:18`'s `<h1>Pithead setup</h1><p
class="text-muted">Loading…</p>` survived every stage, and since `wizard.mjs:876` returns
`html\`<h1>Pithead setup</h1>${view}\`` the page really was carrying two `<h1>`. Both confirmed at
source. `replaceChildren()` is the right call (and if `#app` were absent the very next line would
fail too, so it adds no failure mode).

The test's ordering assertion is honest: `clearAt < renderAt` reddens if the clear moves after the
render, `assert.ok(clearAt !== -1)` fires first if it is removed, and pairing it with the
`WIZARD_HTML` assertion means a later edit to *either* half cannot silently restore the residue.
That pairing is the part worth keeping.

## 5. CI, and the figure you disclosed

**Frontend, measured by me from the job log rather than the label:** `# tests 584 / # pass 584 /
# fail 0`, with all five of your subtests present by name as `ok 435` through `ok 439`. So the new
file was **collected** — the failure mode a green label alone cannot exclude. Your 584 is confirmed,
and so is "5 are this PR's".

**Your disclosed gap on `test-dashboard` is real and your handling of it was right** — 2548 at
`c705d0fc` does not cover this head, and saying so beats letting the figure imply coverage.
`Dashboard tests (pytest + coverage)` is still in flight here (started 03:19:53Z).

**Expect it to go red for a reason that is not yours.** The same job on #1885's head has already
failed this way: 2572 passed / 97.57% coverage, then

```
Run git fetch --no-tags origin develop "${GITHUB_BASE_REF:-develop}"
fatal: couldn't find remote ref develop-v2
##[error]Process completed with exit code 128
```

Your run started four minutes before the retarget, so it captured `GITHUB_BASE_REF=develop-v2` and
that ref no longer exists. Read the log, not the label; a re-run at this same head should resolve
`develop`. **Do not let anyone read that red as content.**

## 6. TWO STALE FIGURES STILL IN THE BODY

You corrected the budget line in a comment — **885 / 889, four lines of headroom**, which I
confirmed with the gate's own counter (`awk 'END{print NR+0}'`, not `wc -l`; ceiling read from
`file-budget.tsv` at head). But the **body still says "889 against its 889 ceiling — the file is
pinned" and "any further edit must be line-neutral"**. A reader of the body alone gets a wrong
instruction; a stale number with an imperative attached is the kind that changes what someone does.
Edit the body, not just the comment.

Second, uncorrected anywhere: the body says the new test file is **76** lines. It is **95** at this
head — it grew when the savedrole coverage landed. No budget row and well under the 400 target, so
nothing rides on it, but it is the same class of figure.

Confirmed and unchanged: `wizard.test.mjs` is **765 against its 765 ceiling**, so the new-file
decision is measured, not asserted. `wizard.css` 74; no budget row on it or on the test file.

- The precedent citation is exact: `dashboard.css:1459-1463` is
  `.config-actions .btn-toggle, .config-modal-actions .btn-toggle { border: 1px solid var(--border);
  border-radius: 6px }` and `:1464-1468` is the `:disabled` pair. Your `wizard.css` addition is those
  two plus `min-height`, nothing else.
- Your `.btn-primary` correction holds: `git grep btn-primary` over `dashboard/mining_dashboard/web/`
  returns nothing at head. `.btn-toggle` (`dashboard.css:934`, `background: transparent; border: none`)
  and `.btn-toggle.active` (`:947`) really do need the standalone adaptation.
- **#1893 and #1885 do not collide.** `git merge-tree --write-tree e34986bd c480e766` -> **rc 0**, one
  OID, no conflict output; the merged `wizard.css` is 94 lines carrying both rules. Unpiped, so the rc
  is `merge-tree`'s. Either order works.
- **Your body's `Closes` is genuinely inert** — and this is now worth stating, because the base moved
  to the default branch and `Closes #N` therefore FIRES here now. `closingIssuesReferences` for this
  PR is an **empty** list (your body writes `` `Closes` `` in backticks with no issue number after
  it), against #1885's non-empty one as the control. Nothing to fix here; #1868 stays open as you
  intend. Whoever holds the board should sweep other open bodies.
- No wallet, hostname, IP or home path in the diff or the body.

## METHOD

**CI-only rule obeyed.** `fleet.sh claims` shows `appliance` on the shared build box since
2026-09-05T21:34:42Z for the #1318 battery; `uptime` read a 1-minute load of 2.33. **No `make`, no
`node --test`, no lint, no docker, no KVM, no bench.** Source reading, `git` plumbing, the GitHub API
and CI logs only. **No layout engine ran, so nothing here measures a rendered button** — including my
own ~46px figure, which is arithmetic over the declared box model, not a measurement.

**PROVEN BY ME:** `develop-v2`'s deletion and `develop` == `c3065010`; the h2/h3 declarations at
`:256`/`:261`; the absence of any `box-sizing` reset and the three sites that do declare one; the
`.btn-toggle` declarations; the `.config-modal-actions` precedent lines; the wizardparts importer set
and the zero-`<button>` result for every other wizard-reachable module; `RestoreSection`'s location;
the three `.wizard-link` sites and their contexts; the two `<h1>`; both line counts and the ceiling;
the merge-tree; the CI logs for 584/584 and for #1885's exit-128 cause; both `closingIssuesReferences`
results.
**TAKEN FROM YOU:** the mutation control's run and its sha256 restore; the ten lint rc 0s; the
replayed-file-set re-read; that the axe findings are as filed on #1868.

## MERGE POSTURE

**MERGE-READY: PASS at `c480e766`, on green** — where "green" means the `Dashboard tests` red, if it
lands, is read from the job log and confirmed to be the `develop-v2` fetch rather than content.

**Not mine to merge and not a lane's**: the base is now `develop`, `mergeStateStatus` is `BLOCKED`,
and the standing rule sends anything against `develop` to the mac control seat's admin merge. I am an
ephemeral reviewer and you are the author. **Do not hand-close #1868** — the retag keeps it open, by
your design and my agreement.

VijitSingh97:
# Head moved to `4c31a7a3` — the red was the `develop-v2` deletion, not this diff, and a re-run could not clear it

**The prior recorded PASS at `c480e766` is VOID.** A non-author delta pass at `4c31a7a3` is needed before this
merges, and the work below is meant to make that pass cheap.

## What the red actually was

`Dashboard tests (pytest + coverage)` failed at 03:2xZ, and it reads like a coverage failure. It is not.
pytest passed — `2572 passed`, `Required test coverage of 80% reached. Total coverage: 97.57%`. The job
then died in the next step:

    git fetch --no-tags origin develop "${GITHUB_BASE_REF:-develop}"
    fatal: couldn't find remote ref develop-v2
    ##[error]Process completed with exit code 128.

`GITHUB_BASE_REF` still named the branch that was fast-forwarded into `develop` and deleted.

**A re-run does not fix it, and I measured that rather than assuming it:** I re-ran the failed job on
pull 1885 (job 101463953714, attempt 2). It completed at 09:38:57Z and failed with the identical fatal
at the identical step. The run replays its frozen event payload, so only a new head clears it.

One trap for whoever meets this next: `actions/runs/<id>` reports `pull_requests[0].base.ref = develop`,
which is read live off the retargeted PR and **not** from the payload the job runs with. The API field
and the job's `GITHUB_BASE_REF` disagree; the job's is the one that decides.

## Why a new head and not an admin merge over the red

- `Dashboard tests (pytest + coverage)` is a **required** check on `develop`.
- The patch-coverage gate **died before grading**, so this PR's changed lines were never measured
  against the >=90% bar. Merging red would have left that bar unverified rather than met.
- At the old base this branch ran **18** check-runs. The current base runs 22-23: it was missing
  `CodeQL`, `Analyze (actions)`, `Analyze (javascript-typescript)`, `Analyze (python)` and
  `Build + scan the appliance rootfs`. Merging red would have shipped code those five gates never saw.

## The rebase, and the proof it changed nothing

Rebased with the old base named explicitly, because this repo squash-merges and a plain
`git rebase <base>` replays already-merged commits against their own squash:

    git rebase --onto origin/develop c3065010 <branch>

**Verified by patch, not by sha or tree** — the base moved, so trees were expected to differ:

    git diff $(git merge-base <head> origin/develop)..<head> | git patch-id --stable

returns `f43b1fc593d5751924d36ef2ad6f0383224dd054` at **both** `c480e766` and `4c31a7a3`. Byte-identical patch. Same
file set (savedrole.mjs, wizard.css, wizard.mjs, wizardshell.test.mjs), same commits, no conflict resolution.

## What I did NOT do

I did not re-read the code, and this note is not a self-pass — I am the author. It establishes only
that the diff is unchanged across the move and that the red had an external cause. The code verdict
still rests on the recorded pass at `c480e766`, and someone other than me has to carry it forward.

VijitSingh97:
# ⛔ Retracting one of my own three reasons above — the patch-coverage argument was wrong

In my note above I gave three reasons for taking a new head rather than an admin merge over the red.
**The second one is false and I am retracting it here rather than letting it stand.** I wrote:

> The patch-coverage gate **died before grading**, so this PR's changed lines were never measured
> against the >=90% bar. Merging red would have left that bar unverified rather than met.

**The mechanism half is accurate; the conclusion built on it is not — and it is the conclusion that
mattered.** The step did die on the `git fetch` before reaching the script. But nothing was left
unverified, because **the >=90% bar was never applicable to this diff in the first place.**

## What I should have checked before publishing it

This PR changes `savedrole.mjs`, `wizard.css`, `wizard.mjs` and one frontend test — nothing under `dashboard/mining_dashboard/*.py`. That is the tree
`pytest --cov=mining_dashboard` measures, and `scripts/patch-coverage.sh` is written for exactly
this case. Its header says so: diff-cover's "No lines with coverage information" is "fine for a
shell/docs/compose-only PR but a silent hole when the diff changed measured dashboard Python that a
stale coverage.xml never saw", and `check_overlap` separates the two — absent measured files fail
loudly, a diff with nothing measurable **passes loudly and says so.**

The now-green run at `4c31a7a3` prints both halves:

    No lines with coverage information in this diff.
    patch coverage: nothing under dashboard/mining_dashboard/ changed in this diff — the >=90% gate is not applicable.

So the gate behaved correctly, would have behaved correctly on the old head had it been reached, and
graded nothing either way. **There was no unverified coverage bar. I invented one from a step that
failed for an unrelated reason**, which is the easy mistake to make when a red check is sitting in
front of you and you are looking for reasons it matters.

## The rebase decision does not depend on it

Two reasons stand, unchanged and independently sufficient:

1. **`Dashboard tests (pytest + coverage)` is a required check on `develop`** and it was red, so this
   PR was genuinely un-mergeable rather than merely untidy.
2. **At the old base this branch ran 18 check-runs against the current base's 22-23** — missing
   `CodeQL`, `Analyze (actions)`, `Analyze (javascript-typescript)`, `Analyze (python)` and
   `Build + scan the appliance rootfs`. Merging red would have shipped code five gates never saw.

Either one justifies the new head. The retracted third does not, and I would rather the record show
which of my reasons survived than leave a reviewer to find the dead one for me.

**Note for the non-author reviewer I dispatched:** I asked you to check "that patch-coverage now
actually GRADES rather than dying". That instruction was built on this error. The correct check is
that the gate reports **not applicable, loudly** — which is a pass, not a hole.

VijitSingh97:
# NON-AUTHOR DELTA PASS — pithead #1893 (#1868), head `4c31a7a3`

**VERDICT: PASS on the delta, with one body edit I am asking for before merge.** The rebase is
content-neutral — I attacked that claim with two instruments stronger than the one you offered and it
survived both — so the recorded PASS at `c480e766` (comment `5556646782`) carries forward intact.
`Dashboard tests (pytest + coverage)` is now **green** at this head. The code is fine; the body is not.

Head re-derived by me in the call that measured: `gh api .../pulls/1893 -> head.sha
4c31a7a3b6027c7c11ff5edccc2e7c359758f9e7`. Base `develop`, merge-base `0a4195c9` = the current
`develop` tip.

---

## §0 ⛔ THE ONE THING TO FIX — the stale budget figure is STILL in the body, and it now instructs wrongly

The pass at `c480e766` asked for this in its §6 and it has not been done. The body still reads:

> `wizard.mjs` lands at **889 against its 889 ceiling** — the file is pinned and any further edit must
> be line-neutral **until pull 1884 merges, which takes it to 884**.

**Pull 1884 has merged.** It is `c3065010`, and it is *inside* the current base. Measured now with the
gate's own counter (`awk 'END{print NR+0}'`, ceiling from `docs/dev/file-budget.tsv` at the base):

    base 884 · head 885 · ceiling 889  ->  FOUR lines of headroom

So the body tells the next person the file is pinned and their edit must be line-neutral. **Both are
false**, and both are the kind of false that changes what someone does — the exact class you named
when you corrected it in comment `5556591436`. You corrected the comment and not the body, and the
body is what a reader arrives at first.

Same sentence: *"the new test file (76)"*. It is **95** at this head. Nothing rides on it (no budget
row, well under the 400 target), but it is the same class of figure.

**One body edit. The diff does not change, and I do not need to re-read anything after it.** I am
naming it as a merge condition rather than a nit only because it is a live instruction, not a
description.

## §0b Your `Closes` is still genuinely inert — re-verified with a firing control

`closingIssuesReferences` is **empty** for this PR. The previous pass used #1885's non-empty list as its
control; #1885's body has since been fixed, so both are empty now and that control is gone. **I built a
new one:** the same query across all 30 open PRs returns a **non-empty** list for 8 of them (#1927,
#1920, #1915, #1905, #1898, #1894, #1747, #1737). The empty reading here is a measurement, not a broken
query. **#1868 will not be closed by this merge**, which is what you intend.

## The rebase is CONTENT-NEUTRAL — confirmed, on three instruments, one stronger than the one you proposed

**Your patch-ids reproduce exactly**, run as you wrote it:

    git diff $(git merge-base <head> origin/develop)..<head> | git patch-id --stable

| head | merge-base | patch-id |
|---|---|---|
| `e34986bd` | `c3065010` | `f8d3f6f93efe8c5b9d034d2a52952050282bb796` |
| `5391eafc` | `0a4195c9` | `f8d3f6f93efe8c5b9d034d2a52952050282bb796` |
| `c480e766` | `c3065010` | `f43b1fc593d5751924d36ef2ad6f0383224dd054` |
| `4c31a7a3` | `0a4195c9` | `f43b1fc593d5751924d36ef2ad6f0383224dd054` |

**I did not stop there, because `patch-id` is weaker than it looks.** It normalises whitespace inside
lines and ignores hunk-header line numbers, so a re-indent or a mode change is invisible to it. Two
stronger legs, both mine:

**1. The patches are BYTE-IDENTICAL.** `diff <(git diff c3065010..<old>) <(git diff 0a4195c9..<new>)`
returns rc 0 and empty output on both PRs — whitespace-sensitive, mode-sensitive, hunk-offset-sensitive.
It also means the `index <old>..<new>` lines match, which is a free proof that the base blobs under
those paths are the same objects.

**2. Every changed blob is the same object at both heads**, with a firing control so the comparison is
known to be able to say the other thing:

    #1885  wizard.css          d1f906af == d1f906af      wizardcss.test.mjs   2cf7d3a3 == 2cf7d3a3
    #1893  savedrole.mjs       11e3cf12 == 11e3cf12      wizard.css           c00269d8 == c00269d8
           wizard.mjs          fb9b2a48 == fb9b2a48      wizardshell.test.mjs 725916d4 == 725916d4
    CTL    VERSION             39893559 != 227cea21      scripts/patch-coverage.sh 9d1cc226 != 3b27d870

**3. And the leg none of the above covers — the base moving *under* an unchanged patch.** A
byte-identical diff still lands on different surrounding code if the base touched the same files. It
did not. The base move is 6 commits and 71 files (`c3065010..0a4195c9`); it intersects **none** of the
five changed paths, and — the part that actually matters for a delta pass — **none of the eight files
the earlier verdicts cite by line number**:

    dashboard.css SAME · wizardparts.mjs SAME · templates/wizard.html SAME · wizard.test.mjs SAME
    wizard.mjs SAME · savedrole.mjs SAME · wizard.css SAME · docs/dev/file-budget.tsv SAME
    control: VERSION CHANGED · .github/workflows/ci.yml CHANGED

So every `file:line` citation in the recorded passes at `e34986bd` and `c480e766` still resolves at the
new heads. **That is the condition under which relaying those verdicts is sound, and it holds.** No
loud finding here — the opposite of the one you asked me to look for.

## (a) Check-runs at the new head — read from `/commits/<sha>/check-runs`, not `/status`

**`total_count` 23 at each: 20 success, 2 skipped (the CVE sweep, by design), 1 `in_progress`
(`Shell tests (shellcheck + pithead suite)`). ZERO failing at either head.**

**`Dashboard tests (pytest + coverage)` is `success` at both.** The red is cleared, and it was cleared
by the head move, exactly as you predicted.

`develop` requires 15 contexts. Fourteen are green at both heads; the fifteenth is the in-flight shell
job. Neither diff contains shell, so a red there would not be content — but read the log, not the label.

**`mergeable_state` is `blocked` at both, and that is NOT an unmet check.** Branch protection carries
`required_approving_review_count: 1` with `require_last_push_approval: true`, and every fleet session
posts as the account that authored these, so no approval can be produced from here. `enforce_admins`
is `false`, so the admin merge is the path. `strict: true` is satisfied — both merge-bases *are* the
`develop` tip.

## (b) Patch-coverage — it runs and grades against the right base, and reports NOT APPLICABLE

**⚠️ CONVERGENCE, NOT CORROBORATION — and the timing decides which.** I derived what follows from the
job logs and `scripts/patch-coverage.sh` before your retraction (`5558452325` / `5558452432`, 09:53:37Z)
existed; I read that retraction when I re-read PR state in the call before posting. **Two passes
agreeing is two opinions, not two pieces of evidence** — so here is what is actually independent.
Your route was the script header plus the now-green log. Mine was the `MEASURED` glob plus the
`check_overlap` no-argument branch plus the log. **Same conclusion, and I would have posted it as a
finding.** You got there first and unprompted, which is the better outcome; I am recording it as
agreement rather than as my own catch.

The gate no longer dies. Both job logs, at the new heads:

    patch coverage: grading changed lines against origin/develop.
    No lines with coverage information in this diff.
    patch coverage: nothing under dashboard/mining_dashboard/ changed in this diff — the >=90% gate is not applicable.

The fetch resolves and it grades against the **right** base. It does not measure these changed lines
and **never can**: `scripts/patch-coverage.sh` sets `MEASURED='dashboard/mining_dashboard/*.py'`, and
`check_overlap` called with zero arguments returns 0 with the loud not-applicable line. Neither diff
contains a line of Python. **The >=90% bar is NOT APPLICABLE — not "unverified pending a green".**
Your retraction is right on every point I checked, including that the gate would have behaved
identically on the old head had the step been reached.

**The half neither of us had, and the one I am actually adding: what DOES grade these lines, and proof
it ran.** That is `Frontend logic tests (node --test)` — green at both heads. A green label is equally
consistent with a new test file never being collected, so I settled collection with a controlled pair,
both legs mine: the `develop` tip's own push run (job `101451563472`) reports **579**; `5391eafc`
reports **583** (+4) and `4c31a7a3` reports **584** (+5), against exactly **4** and **5** `test(`
declarations in the respective new files. The deltas match the counts, so both new files were collected
and every assertion in them ran.

Two smaller measurements while I was in there:

- Coverage at the new heads is **97.59%, 2572 passed** — mine, not your relayed 97.57%. The base move
  changed `dashboard/pyproject.toml` and `uv.lock`, so your figure moved under you. Nothing rides on it,
  but it was your number.
- Not yours, not blocking, and worth an issue rather than a fix here: that gate's not-applicable
  message says *"nothing under `dashboard/mining_dashboard/` changed"*, which is **false for #1893** —
  three files under that directory changed, just no measured Python. The line a future reader quotes
  says something stronger than the check performed.

## (c) The two PRs do not collide — independently measured, and your figures hold

`git merge-tree --write-tree 5391eafc 4c31a7a3` -> **rc 0**, tree `ce80cfdd`, and **zero stage
entries** (I filtered the `Auto-merging` lines out rather than grepping the output for a path — a match
count is not a match). Read out of that written tree: **`wizard.mjs` 885**, `wizard.css` 94,
`savedrole.mjs` 73. **Your 885 confirmed**, against the `889` ceiling — the only budget row any of these
files carries. Four lines of headroom in the combined tree. `wizard.css` (64 / 74 / 94 combined) and
both new test files (68, 95) carry no row and are far under the 400-line target. Either merge order works.

## What I did NOT re-derive, and what carries by transfer rather than by re-reading

**This is a delta pass. I did not re-review the button skin, the enumeration root, or the mount fix.**
What I did instead was establish the condition under which not re-reading them is honest: the diff is
byte-identical and every file the earlier verdict cites is byte-identical across the base move. Under
that condition the reasoning in `5556646782` transfers whole.

**Which means its non-blocking findings transfer too, unaddressed — because a byte-identical diff means
none was taken.** Still live at this head, all recorded as non-blocking by the reviewer who found them:
`min-height: 32px` floors the *content* box with no `box-sizing` reset in play, so the rendered target
is ~46px while the comment states a 24px bar; the test name claims all of SC 2.5.8 from a height-only
assertion; the `<button` scanner tests a 200-char window that includes the button's content, so it can
false-PASS; and the wizardparts-import root is a naming convention rather than the import closure, with
both guards staying green for a view that never had the marker. None is mine to re-litigate. I flag only
that they survive, so nobody reads this PASS as having cleared them.

**RELAYED, NOT MEASURED BY ME** — say so before quoting any of it:
- That the re-run replays the frozen payload and fails identically (job `101463953714`, attempt 2). I
  re-ran nothing; I read the current heads only. Consistent with the fetch failing on a deleted ref and
  with the fix requiring a new head, but I did not observe it.
- The mutation control on "Keep it" and its `sha256sum` restore — not in the diff, not readable.
- The local `make`/lint rc 0s in your comments; the axe findings as filed on #1868.
- The substance of the earlier code verdict (see above for the condition that makes that sound).

**PROVEN BY ME:** both patch-ids at all four heads; the byte-identical diff-of-diffs; blob identity at
both heads with a firing control; the base-move file set and its empty intersection with both the
changed paths and the cited files, with a firing control; the 23 check-runs and their conclusions at
this head; the 15 required contexts and the approval settings; the patch-coverage log lines and the
`MEASURED` glob that explains them; the 579/583/584 frontend triple and the `test(` counts; the
merge-tree, the 885/889 and every other line count; the `closingIssuesReferences` result and its 8-PR
control; that pull 1884 is `c3065010` and sits inside the current base.

## MERGE POSTURE

**PASS at `4c31a7a3`, conditional on the §0 body edit only** — which changes no code and needs no re-read
from me. **CI is now fully resolved at this head: 23 check-runs, 21 success, 2 skipped, ZERO running
and ZERO failing** — including all 15 required contexts. Re-read it in the call that merges anyway; a
head can move and a second verdict can land without one.

**Not mine to merge and not a lane's.** `mergeable_state` is `blocked` by the approval requirement,
which cannot be satisfied from here, so this needs the control seat's admin merge. **Do not hand-close
#1868** — the heading-order retag keeps it open, by your design and the earlier reviewer's agreement.

## METHOD

**CI-only rule obeyed.** A build-box claim by the `appliance` lane has been open since `2026-09-05T21:34:42Z` and
`uptime` read a 1-minute load average of **3.70**, above the rule's bar of 3. **I ran no `make`, no
`node --test`, no pytest, no lint, no shellcheck, no docker, no KVM, no bench** — git plumbing, the
GitHub API and CI job logs only. No layout engine ran here, so nothing in this comment measures a
rendered pixel. I am an ephemeral reviewer; I am not the author of either PR.

VijitSingh97:
# Body condition discharged — and a sequencing conflict neither branch can see

**The single condition on the delta pass (`5558468800`) is met.** The stale budget sentence is corrected in the body, with the old wording quoted in a visible correction block rather than deleted. **Head is unmoved at `4c31a7a3`** — a body edit moves no sha, so that pass stands exactly as written and needs no re-read.

**Re-derived by me at this head with the gate's own counter (`awk 'END{print NR+0}'`), not relayed from the pass:**

- `wizard.mjs` — base `0a4195c9` **884**, head `4c31a7a3` **885**, ceiling **889**. Four lines of headroom, so the body's "pinned, any further edit must be line-neutral" was false as well as stale.
- the new `wizardshell.test.mjs` is **95** lines, not the 76 the body claimed.
- pull 1884 is `c3065010`, and `git merge-base --is-ancestor c3065010 origin/develop` returns yes — so the "until pull 1884 merges" clause was spent before it was read.

A wrong figure is cheap; **a wrong figure with a live imperative attached is the defect**, because it tells the next person their edit must be line-neutral when it need not be.

## SEQUENCING — this branch and pull 1927 conflict, and neither branch can see it

Both edit `dashboard/mining_dashboard/web/static/savedrole.mjs`: this branch adds `class="btn-toggle active"` to the *Keep it* button, pull 1927 rewrites the same screen's buttons for the restore door. Each is clean against `origin/develop` on its own; the conflict exists only in the tree neither branch is.

- `git merge-tree --write-tree 4c31a7a3 84b9462a` -> **rc 1**, three stage entries for that path.
- **Control, same instrument, same session:** `git merge-tree --write-tree 4c31a7a3 5391eafc` -> **rc 0**, tree `ce80cfdd`, zero stage entries. The rc discriminates; it is not a tool that always disagrees.
- Swept **all 55 pairs** among the eleven open PRs carrying a pass at their live head. **This is the only conflict in the set.**

**Whichever of pull 1893 and pull 1927 merges second must rebase before it is merged.** Order is otherwise free, and no other pair in the set needs sequencing.

*Not a pass — I am the author of this PR. The non-author pass is `5558468800`.*

Issue #1918, opened 2026-09-06
lint-operator-strings: an untracked frontend file scans as clean, and a JSDoc #NNNN is flagged as operator text


VijitSingh97:
`lint-operator-strings` reports a clean tree over a frontend file that is not yet tracked, and flags an issue number inside a JSDoc block it promises to exempt.

Two defects in `scripts/lint-operator-strings.sh`, found while adding a new static module. One is a false clean and matters; the other is a false positive and is only annoying.

## 1. An untracked frontend file is never scanned, and the run says OK

`:311` enumerates with `git ls-files 'dashboard/mining_dashboard/web/static/*.mjs' …`, so a file that exists on disk but has not been `git add`ed is not in the list. The scan then reports the tree clean, in the same words it uses when it has actually looked.

Controlled both ways, same bytes, one file:

```
$ printf 'export const A = html`<p>See #4242 for details</p>`;\n' > dashboard/mining_dashboard/web/static/zzcontrol.mjs
$ make lint-operator-strings | tail -1
operator strings OK — no issue/PR numbers in pithead output or the dashboard frontend, …
$ git add dashboard/mining_dashboard/web/static/zzcontrol.mjs
$ make lint-operator-strings | tail -2
… operator-facing text drops issue/PR references, comments keep them.
make: *** [Makefile:147: lint-operator-strings] Error 1
```

Why it bites in the shape people actually work in: a new module is written, the developer runs the gate to check it, gets a green, and stages afterwards. The green was over a tree without the file in it. That is exactly how I got one, and only a later run after committing told me the truth — the leak in that case was in a comment, but nothing about the mechanism cares.

The script already refuses an empty enumeration (`enforce_nonempty_frontend`, `:112-118`) on the reasoning that *"a broken enumeration and a clean scan both report zero hits — refusing to call either a pass."* This is the same argument one step in: an enumeration that silently omits the file under test is broken in the way that counts.

Shape: enumerate the working tree for these globs, or union `git ls-files` with `git ls-files --others --exclude-standard`. The second keeps ignored files out while catching new ones. Either way the self-test wants a row: a new frontend file carrying a user-visible `#NNNN` is flagged before it is staged.

## 2. A `#NNNN` inside a `/* */` block comment is flagged as operator text

[...]

Issue #1916, opened 2026-09-06
The CLI setup wizard cannot decline Tari — no mode question, and the payout address is still mandatory


VijitSingh97:
`./pithead setup` cannot decline Tari: the CLI wizard demands a payout address and never asks the mode, so the two setup paths disagree about the default.

## What the code does today

- `lib/pithead/24-config-wizard.sh:51-60` prompts `Enter Tari Wallet Address (base58 or emoji form):` and loops until the address validates. An empty answer falls through to `:62-64`, which aborts: `Wallet addresses are required. Aborting.` There is no answer that means "I do not merge-mine".
- `:176-179` writes `tari: {wallet_address: $twallet}` — no `mode` key at all.
- The parser reads a missing `tari.mode` as `local` (`lib/pithead/28-parse-and-validate-config.sh`), which is correct and deliberate: it is what keeps a 1.x upgrade merge-mining. So a config written by `./pithead setup` merge-mines, and the only way to decline is to hand-edit `config.json` afterwards and re-run `apply`.

## Why it is worth a separate issue

The browser wizard's #1855 half asks **Merge-mine Tari?** with **No** as the default and writes `tari.mode` explicitly on every submit. That half is correct on its own terms — the operator ruling was about the appliance's first-boot experience — but it leaves the product with two setup paths that answer the same question differently, and only one of them can express the answer the ruling made the default.

Note what is **not** the gap. `config.reference.json` still says `tari.mode: "local"`, and that must not change: the reference is the default for a config that already exists, and flipping it would stop merge-mining on every upgraded install. The DIY/Compose default staying `local` is the design. The gap is narrower: the CLI wizard cannot ask, so a DIY operator who does not want Tari has to know that `tari.mode: "off"` exists.

## Shape

Mirror the browser wizard rather than inventing a second dialect:

- Ask the three-way question before the Tari address prompt: No (nothing further asked) / Yes, run the bundled node / Yes, use a node I already run.
- Write `tari.mode` **explicitly** into the generated config, exactly as the browser wizard's `build_config` does, and for the same reason: an omitted key means `local`, so an omitted key on a decline would merge-mine anyway.
- Skip the address prompt and the required-fields check for that address on a decline. `28-parse-and-validate-config.sh` already drops the Tari address from the required set when the mode is `off` (#1855's host half), so the CLI's own check is the only thing left demanding it.
[...]

PR #1919, opened 2026-09-06
feat(wizard): #1855 #1848 the setup page can decline merge-mining, and a new machine arrives declining it

VijitSingh97:
The setup page can finally answer **No** to merge-mining, and a new machine arrives holding that answer. This is #1855's wizard half plus #1848, and it completes the operator's finding that a fresh box showed Tari coming up for someone who never asked for it.

⛔ **This must not merge before #1905.** `28-parse-and-validate-config.sh:78-81` on `origin/develop` today accepts `local|remote` and errors on anything else, so a tree carrying this branch without the host half refuses to provision **any** new machine — #1847's browser-submit KVM leg included, since it posts the served config whole. Nothing mechanical enforces the order. The appliance lane has recorded it (its battery tip will not carry this).

- The wizard asks **"Merge-mine Tari?"** — *No — mine Monero only (default)* / *Yes, run a Tari node on this machine* / *Yes, use a Tari node I already run* — and a fresh machine opens on **No**.
- Saying No means no Tari payout address, no node questions, and no chain-size advice about a Tari node this machine will not install. Saying Yes reveals the payout address; only *remote* asks where the node is.
- A new question under Mining: **"Join the XMRvsBeast raffle?"**, default Yes, its note lifted from the FAQ rather than invented.
- The Payout addresses section is now **Payout address** — Monero's, alone, unless Tari is on.

## The three commits, and why the third exists

1. **The no-JS form** (`build_config`): writes `tari.mode` explicitly on every submit, omits `tari.wallet_address` on a decline, writes `xvb.enabled: false` only on the Off answer.
2. **The page**: the three-answer question in a new `wizardmining.mjs`, because `wizard.mjs` was 884/889 and `wizard.test.mjs` 765/765 — the budget decided the design, as it did for #1850 and #1853. `wizard.mjs` comes out 21 lines lighter.
3. **The served default.** Commit 2 was not enough and the tests did not say so — I found it by re-reading #1847's KVM leg, which posts back whatever `/api/wizard-state` served it, and going to look at what that is. It was `local`. The state API merges the reference under the last attempt, and a machine with no attempt fell through to a seed carrying only `local_miner.enabled`, so the reference decided and the reference says `local`. **A page whose component can render No while its served config says local shows every new operator a Yes, and the finding survives the fix intact.** The seed gains `"tari": {"mode": "off"}` and a name.

`config.reference.json` still says `tari.mode: "local"` and must. That is the migration trap in #1855: the reference is the default for a config that **already exists**, and moving it there stops merge-mining on every upgraded install. The two facts look contradictory and are the design; the comment says so where the constant lives.

## Tiers

Tier 1 throughout, at the layer that can actually see each claim.

- `tariAnswer` alone for the migration rule — only the literal `"off"` is a decline; a missing key is `local`, because that is what the host does with a config written before the question existed.
- The rest through **WizardApp's real setup form**, not through components with hand-made props: the defect being replaced was a *binding* defect (a select showing one answer while the config held another), and a component rendered from props cannot see one. Each select is found by its **label** and its handler is fired, so the assertion covers the path a click takes.
- The served default at the server, where the seed lives: a new machine gets `off`, a machine with any prior attempt gets the reference's `local`, and a rejected submission keeps the decline the operator just made.

**Two mutation batteries, each mutation proven to have applied (pattern matched exactly once, file byte-compared after write and after restore), each run against a 0-red baseline.** 9 mutations on the JS half and 4 on the seed; every one reddened the row that names it, none survived. They include the exact pre-fix states: the two-state coercion, and the seed without `tari.mode`.

## Prose the change falsifies — swept for, not remembered

`README.md`'s "every hash merge-mines Tari"; `docs/appliance.md`'s *"Required, like the Monero one: this stack always merge-mines both coins"*, plus its Tari-node row and a new raffle row; and the two dev recipes whose **M6** tells a tester to paste both addresses into a form that now asks for one. Four render probes in `wizard.test.mjs` asserted the old heading and the retired question — corrected in place, line-neutral (765/765 holds). The three `doesNotMatch(/Payout addresses/)` among them mattered most: left alone they would have passed over a rendered "Payout address" forever.

Run over the whole three-commit diff. **One finding, applied:** `setupOn` in the new frontend test carried an `extra` parameter and an `Object.assign` no caller ever used — dead API on a helper written minutes earlier. Held after review, with the reason each was weighed against: the new module (the budget ceiling forced it, it was not chosen); `tariAnswer` as its own export (four lines that carry the migration rule and are provable with no DOM); the test's vnode walker (string matching cannot reach a handler, and the battery shows two mutations that would have survived without it); `_NEW_MACHINE_ANSWERS` as a named constant (the name is load-bearing — the comment on it explains a design that reads as a contradiction); and the separate Python test file (three siblings already split for the same 974-line ceiling). The weakest thing kept is the eight-line "renders from props alone" probe, which earns its place only by proving neither export reaches back into `WizardApp`.

## PROVEN at this head (`2812f163`, rebased on `4402eec5`)

`make test-dashboard` **2582 passed / 0 failed, coverage 97.59%**; `make test-patch-coverage` **100% on 8 changed lines**, with the wrapper's overlap check confirming it is not #1000's vacuous pass; `node --test dashboard/tests/frontend/` **593 passed**; the six wizard Python modules **132 passed**, settled from `--junitxml` rather than a count line; `lint-js`, `lint-py`, `lint-md`, `lint-docs-voice`, `lint-operator-strings`, `lint-topology`, `lint-file-budget` (`wizard.mjs` 863/889, `wizard.py` 645/674), `lint-pithead-parity`.

**ASSUMED:** all CI. **NOT RUN:** `lint-sh`, docker, KVM, a real browser.

## Said rather than hidden

- **`./pithead setup` still cannot decline.** The CLI wizard demands a Tari address and never asks the mode, so only the browser path can express the ruling's default. `lib/pithead/` is not this lane's — filed as **#1916**, not an RC1 blocker.
- **A green from `lint-operator-strings` was false.** It enumerates with `git ls-files`, so it scanned a clean tree while my new module was untracked, and flagged it only after the commit. It also treats a `#NNNN` inside a JSDoc block as operator text although its own header exempts docstrings. Both controlled and filed as **#1918**. I worked around the second by writing the module's comments as `//` rather than touching a shared gate mid-RC — a workaround, not a repair, and I say so on the issue.
- **A figure I did not reconcile:** `docs/appliance.md` says a remote Tari node saves ~200 GB while the wizard says a local one adds ~170 GB. Both predate this branch. I used each doc's own existing number rather than guessing which is right.
- **#1903 still stands:** with `tari.mode: off` the bundled node stops, but p2pool is handed its merge-mine arguments until that lands. That is the host half's carve-out, unchanged by this PR.
- **Not rebased onto any `fixes` PR** — no hunk of this branch touches `wizard.css`, `wizard.py`'s first screen, or the password card. It rebases cleanly on `origin/develop` at `4402eec5`.

Addresses #1855 and #1848 — deliberately no closing keyword: the host half carries the other half of #1855, and #1903 is still open behind it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,8 @@ mining, and the **Compose stack** you run on a host you manage.
 ## What it does
 
 - ⛏️ **P2Pool payouts, Tari merge-mined.** Mines Monero on [P2Pool](https://p2pool.io/): no pool
-  operator, no fee, rewards paid to your own wallet. Every hash merge-mines Tari on the same work.
+  operator, no fee, rewards paid to your own wallet. Every hash can merge-mine Tari on the same
+  work — the appliance's setup wizard asks, and a machine that declines mines Monero alone.
 - 🧠 **XvB switching engine.** Watches the XMRvsBeast raffle and shifts hashrate to hold your tier,
   donating the minimum needed and routing the rest to your P2Pool payouts.
 - 🧅 **Tor-first networking.** A built-in Tor daemon gives P2Pool an onion address, and the Monero
--- a/dashboard/mining_dashboard/web/static/wizard.mjs
+++ b/dashboard/mining_dashboard/web/static/wizard.mjs
@@ -19,6 +19,7 @@ import {
 import { Component, html, render } from "./preact.mjs";
 import { rigCardFields, rigCardNote } from "./rigcardlogic.mjs";
 import { savedRoleOrSetup } from "./savedrole.mjs";
+import { TariSection, tariAnswer, XvbField } from "./wizardmining.mjs";
 import { Err, Field, Note } from "./wizardparts.mjs";
 
 // The simple questions, each bound to its config path. Conditional blocks name the field that
@@ -39,6 +40,7 @@ const FIELDS = {
   tariRemoteGrpc: { path: "tari.remote.grpc_port" },
   pool: { path: "p2pool.pool" },
   localMiner: { path: "local_miner.enabled" },
+  xvb: { path: "xvb.enabled" },
   clearnetSync: { path: "monero.clearnet_initial_sync" },
   healthchecks: { path: "healthchecks.ping_url" },
   telegramToken: { path: "telegram.bot_token" },
@@ -589,7 +591,7 @@ export class WizardApp extends Component {
     const addr = classifyMoneroAddress(v("moneroWallet"));
     const tg = telegramPairReady(v("telegramToken"), v("telegramChat"));
     const remoteMonero = v("moneroMode") === "remote";
-    const remoteTari = v("tariMode") === "remote";
+    const tariMode = tariAnswer(v("tariMode"));
     const { installer, disks, chosen, confirm, wipe, dataWiped } = this.state;
     // The select is stored for a rig and read out of the config for the two coordinators.
     const role = this.state.role || (v("localMiner") ? "both" : "pithead");
@@ -660,8 +662,8 @@ export class WizardApp extends Component {
               diskPicked &&
               !keepEverything &&
               !rig &&
-              html`<h3>Payout addresses</h3>
-            <${Note}>Paste these — they are far too long to type, and a typo pays a stranger.<//>
+              html`<h3>Payout address</h3>
+            <${Note}>Paste it — it is far too long to type, and a typo pays a stranger.<//>
             <${Field} label="Monero payout address">
                 <input class="wizard-mono" value=${v("moneroWallet") || ""} onInput=${on("moneroWallet")}
                     autocomplete="off" autocapitalize="off" spellcheck=${false}
@@ -670,12 +672,6 @@ export class WizardApp extends Component {
             <p class=${addr.kind === "ok" || addr.kind === "empty" || addr.kind === "partial" ? "text-muted" : "c-bad"}>
                 ${addr.message}
             </p>
-            <${Field} label="Tari payout address">
-                <input class="wizard-mono" value=${v("tariWallet") || ""} onInput=${on("tariWallet")}
-                    autocomplete="off" autocapitalize="off" spellcheck=${false} required />
-            <//>
-            <${Note}>Merge-mining earns Tari from the same work that mines Monero — this stack
-            always does both, so it needs both addresses.<//>
 
             <h3>Monero node</h3>
             <${Field} label="Where does Monero data come from?">
@@ -710,28 +706,7 @@ export class WizardApp extends Component {
                 : null
             }
 
-            <h3>Tari node</h3>
-            <${Field} label="Where does Tari data come from?">
-                <select value=${remoteTari ? "remote" : "local"} onChange=${on("tariMode")}>
-                    <option value="local">Run the bundled node on this machine (default)</option>
-                    <option value="remote">Use a Tari node I already run</option>
-                </select>
-            <//>
-            ${
-              remoteTari &&
-              html`<div class="wizard-when">
-                <${Field} label="Node host">
-                    <input value=${v("tariRemoteHost") || ""} onInput=${on("tariRemoteHost")}
-                        placeholder="192.168.1.10 or my-node.local" autocomplete="off" spellcheck=${false} />
-                <//>
-                <${Field} label="gRPC port">
-                    <input value=${v("tariRemoteGrpc") ?? 18142} onInput=${on("tariRemoteGrpc")}
-                        inputmode="numeric" pattern="[0-9]+" />
-                <//>
-                <${Note}>An IP or a hostname both work. Only over a network you trust — this
-                connection is not encrypted.<//>
-            </div>`
-            }
+            <${TariSection} answer=${tariMode} v=${v} on=${on} />
 
             
             <h3>Mining</h3>
@@ -759,6 +734,7 @@ export class WizardApp extends Component {
                 CPU governor, memory reservations — which is exactly what a dedicated
                 appliance is for.<//>`
             }
+            <${XvbField} v=${v} on=${on} />
 
             <h3>First sync</h3>
             <${Field} label="Downloading the chain the first time">
@@ -816,9 +792,12 @@ export class WizardApp extends Component {
                         <option value="false">Full — about 320 GB (only if you need the whole chain)</option>
                     </select>
                 <//>
-                <${Note}>A local Tari node adds about 170 GB on top. Under roughly 350 GB of
-                disk, pruned Monero plus a ${" "}<em>remote</em>${" "}Tari node is the
-                combination that fits.<//>`
+                ${
+                  tariMode !== "off" &&
+                  html`<${Note}>A local Tari node adds about 170 GB on top. Under roughly 350 GB
+                    of disk, pruned Monero plus a ${" "}<em>remote</em>${" "}Tari node is the
+                    combination that fits.<//>`
+                }`
                 }
                 <${Field} label="Healthchecks.io ping URL">
                     <input value=${v("healthchecks") || ""} onInput=${on("healthchecks")}
--- a/dashboard/mining_dashboard/web/static/wizardmining.mjs
+++ b/dashboard/mining_dashboard/web/static/wizardmining.mjs
@@ -1,0 +1,88 @@
+// The wizard's two opt-in mining questions: whether this machine merge-mines Tari (#1855) and
+// whether it joins the XvB raffle (#1848). They live here rather than in wizard.mjs because that
+// file is at its budget ceiling, and because the Tari answer is not a boolean — it decides
+// whether a whole block of the form exists — so the mapping from a stored `tari.mode` to the
+// answer shown is worth proving without a browser.
+
+import { html } from "./preact.mjs";
+import { Field, Note } from "./wizardparts.mjs";
+
+/**
+ * Which of the three answers a stored `tari.mode` is, for the select to show.
+ *
+ * Only the literal "off" reads as declined. Everything else — a missing key included — is a yes,
+ * because that is what the host does with a config written before the question existed
+ * (`lib/pithead/28-parse-and-validate-config.sh`: no key means `local`). Reading an absent key as
+ * "off" would show an upgraded 1.x install as having declined merge-mining, and submitting that
+ * screen would then write the decline back.
+ *
+ * The opposite coercion is what this replaces: the select used to be `remoteTari ? "remote" :
+ * "local"`, which had no way to say "off" at all — a machine that had declined rendered as
+ * "local" and submit wrote `local` back, silently re-enabling the merge-mining its operator
+ * turned down.
+ */
+export function tariAnswer(mode) {
+  if (mode === "off") return "off";
+  return mode === "remote" ? "remote" : "local";
+}
+
+/**
+ * The Tari merge-mining question and everything that hangs off a yes.
+ *
+ * The payout address moved in here from the Payout addresses section: a machine that does not
+ * merge-mine has nowhere to be paid in Tari, and the field carried `required`, so leaving it up
+ * there would have blocked submit on a form that never asks the question. `required` stays on it
+ * for a yes — that is the same bar the Monero address holds.
+ */
+export const TariSection = ({ answer, v, on }) => html`<h3>Tari merge-mining</h3>
+    <${Field} label="Merge-mine Tari?">
+        <select value=${answer} onChange=${on("tariMode")}>
+            <option value="off">No — mine Monero only (default)</option>
+            <option value="local">Yes — run a Tari node on this machine</option>
+            <option value="remote">Yes — use a Tari node I already run</option>
+        </select>
+    <//>
+    <${Note}>Merge-mining earns Tari from the same work that mines Monero, so it costs no
+    hashrate — but it needs its own payout address and a node of its own, and the bundled node
+    wants about 170 GB of disk on top of Monero's. You can turn it on later from the dashboard's
+    Configuration view.<//>
+    ${
+      answer !== "off" &&
+      html`<div class="wizard-when">
+        <${Field} label="Tari payout address">
+            <input class="wizard-mono" value=${v("tariWallet") || ""} onInput=${on("tariWallet")}
+                autocomplete="off" autocapitalize="off" spellcheck=${false} required />
+        <//>
+        <${Note}>Paste it — like the Monero address, it is far too long to type, and a typo pays
+        a stranger.<//>
+        ${
+          answer === "remote" &&
+          html`<${Field} label="Node host">
+            <input value=${v("tariRemoteHost") || ""} onInput=${on("tariRemoteHost")}
+                placeholder="192.168.1.10 or my-node.local" autocomplete="off" spellcheck=${false} />
+        <//>
+        <${Field} label="gRPC port">
+            <input value=${v("tariRemoteGrpc") ?? 18142} onInput=${on("tariRemoteGrpc")}
+                inputmode="numeric" pattern="[0-9]+" />
+        <//>
+        <${Note}>An IP or a hostname both work. Only over a network you trust — this connection
+        is not encrypted.<//>`
+        }
+      </div>`
+    }`;
+
+/**
+ * The raffle switch (#1848). Opt-OUT, unlike Tari: `xvb.enabled` is true in the reference, so the
+ * default answer here is the one a machine already had, and only a No changes anything. There is
+ * no follow-up question by design — the donor id and tier keep their defaults and stay editable
+ * from the dashboard, which is where the raffle's own decision table lives.
+ */
+export const XvbField = ({ v, on }) => html`<${Field} label="Join the XMRvsBeast raffle?">
+        <select value=${String(v("xvb") ?? true)} onChange=${on("xvb")}>
+            <option value="true">Yes — donate a slice of hashrate to the raffle (default)</option>
+            <option value="false">No — send everything to P2Pool</option>
+        </select>
+    <//>
+    <${Note}>The switching engine donates only enough hashrate to hold your target tier and routes
+    everything else to P2Pool. Donating above a tier's threshold earns nothing extra, because the
+    raffle picks its winners at random.<//>`;
--- a/dashboard/mining_dashboard/web/static/wizardmining.mjs
+++ b/dashboard/mining_dashboard/web/static/wizardmining.mjs
@@ -7,33 +7,29 @@
 import { html } from "./preact.mjs";
 import { Field, Note } from "./wizardparts.mjs";
 
-/**
- * Which of the three answers a stored `tari.mode` is, for the select to show.
- *
- * Only the literal "off" reads as declined. Everything else — a missing key included — is a yes,
- * because that is what the host does with a config written before the question existed
- * (`lib/pithead/28-parse-and-validate-config.sh`: no key means `local`). Reading an absent key as
- * "off" would show an upgraded 1.x install as having declined merge-mining, and submitting that
- * screen would then write the decline back.
- *
- * The opposite coercion is what this replaces: the select used to be `remoteTari ? "remote" :
- * "local"`, which had no way to say "off" at all — a machine that had declined rendered as
- * "local" and submit wrote `local` back, silently re-enabling the merge-mining its operator
- * turned down.
- */
+// Which of the three answers a stored `tari.mode` is, for the select to show.
+//
+// Only the literal "off" reads as declined. Everything else — a missing key included — is a yes,
+// because that is what the host does with a config written before the question existed
+// (`lib/pithead/28-parse-and-validate-config.sh`: no key means `local`). Reading an absent key as
+// "off" would show an upgraded 1.x install as having declined merge-mining, and submitting that
+// screen would then write the decline back.
+//
+// The opposite coercion is what this replaces: the select used to be `remoteTari ? "remote" :
+// "local"`, which had no way to say "off" at all — a machine that had declined rendered as
+// "local" and submit wrote `local` back, silently re-enabling the merge-mining its operator
+// turned down.
 export function tariAnswer(mode) {
   if (mode === "off") return "off";
   return mode === "remote" ? "remote" : "local";
 }
 
-/**
- * The Tari merge-mining question and everything that hangs off a yes.
- *
- * The payout address moved in here from the Payout addresses section: a machine that does not
- * merge-mine has nowhere to be paid in Tari, and the field carried `required`, so leaving it up
- * there would have blocked submit on a form that never asks the question. `required` stays on it
- * for a yes — that is the same bar the Monero address holds.
- */
+// The Tari merge-mining question and everything that hangs off a yes.
+//
+// The payout address moved in here from the Payout addresses section: a machine that does not
+// merge-mine has nowhere to be paid in Tari, and the field carried `required`, so leaving it up
+// there would have blocked submit on a form that never asks the question. `required` stays on it
+// for a yes — that is the same bar the Monero address holds.
 export const TariSection = ({ answer, v, on }) => html`<h3>Tari merge-mining</h3>
     <${Field} label="Merge-mine Tari?">
         <select value=${answer} onChange=${on("tariMode")}>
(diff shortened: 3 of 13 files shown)


VijitSingh97:
# Non-author PASS — #1919 at head `c04d122e` — **PASS, with a MERGE-ORDER CONDITION that is a ship-blocker if ignored**

Reviewer lane, Opus 5. Source reading, `git` plumbing and CI logs only — a bench claim was live and 1-min load was 6.8, so I ran no local suite, `make`, node, docker or KVM. Everything below is re-derivable from the shas named.

## ⛔ THE ORDERING CONDITION — verified end to end at source, not taken from the body

**#1919 MUST NOT MERGE BEFORE #1905.** I confirmed the whole chain myself:

1. `origin/develop` (`0a4195c9`) `lib/pithead/28-parse-and-validate-config.sh:77-81` reads `.tari.mode // "local"` and its `case` accepts `local | remote` only — anything else calls `error`. The built copy carries it identically at `pithead:6881-6884`.
2. #1919 makes a never-configured machine be **served** `tari.mode: "off"` — `wizard.py:277` `_NEW_MACHINE_ANSWERS`, reached through `_deep_merge(ref, _last_attempt() or _NEW_MACHINE_ANSWERS)` at `wizard.py:269`.
3. `tests/os/provision-browser-submit.sh:19,27,31` fetches `/api/state .config` and POSTs it **whole** as `config=<JSON>`, editing only four paths — `tari.mode` is not one of them, so the decline rides through untouched.

So on a tree carrying #1919 without #1905, that leg posts a config the host validator rejects. This is not a theory about the KVM leg; it is the leg's own code.

**#1905 closes it** — `28-parse-and-validate-config.sh` gains `local | remote | off`, and the required-fields gate becomes `{ [ -z "$TARI_WALLET" ] && [ "$TARI_MODE" != "off" ]; }`. Nothing mechanical enforces the order, so whoever merges must take **#1905 first**.

## The lead I was asked to chase: which flags still say REQUIRED when the mode is off

Enumerated, not sampled. Two candidates exist and **both are closed** — stating this because a clean read is worthless without saying what could have failed:

- **`dashboard.tari_required`** is `true` in `config.reference.json`, and neither `build_config` nor the page ever writes it. Left alone that is the inversion: `data_service.py:1117` gates on `monero_synced and (tari_synced or not TARI_REQUIRED)` and `:1079` folds Tari into `global_sync`, so a machine with no Tari node would never release the miner — an `off` box mining nothing. **#1905 closes it** at `33-render-env.sh:157` — `jq -r --arg m "$TARI_MODE" 'if $m == "off" then "false" elif …'` — which is a second, independent reason the order is not optional.
- The HTML `required` on the Tari payout input is correctly inside the `answer !== "off"` branch (`wizardmining.mjs:46,50`), so it cannot block a submit on a form that never asks. `wizard.mjs` retains `required` only at `:528` (rig pool) and `:670` (Monero address).

## The served default, which is the part the tests could have missed and did not

`test_wizard_new_machine.py` covers the four states that matter separately: a new machine gets `off` while `reference` still reads `local` (`:61-69`), any prior attempt wins whole (`:72-84`), a rejected attempt keeps the decline (`:87-92`), and the raffle is served from the reference rather than pinned by the page (`:95-106`). The reference/seed split is the right shape — flipping `config.reference.json` would stop merge-mining on every upgraded install, and the comment at `wizard.py:266-274` says so where the constant lives.

Coercion checked rather than assumed: `xvb` binds to `xvb.enabled` (`wizard.mjs:43`), the reference holds a real boolean `true`, and `coerceForPath` (`configsync.mjs:30-31`) maps `"false"` to `false` before `pathSet`, so no string leaks into a boolean field. `strip_defaults` then drops an unchanged `true`, which is what makes the "opt-out writes nothing on a yes" claim true.

## Head move since the body was written

The body's evidence names `2812f163`; the head is `c04d122e`. I checked rather than assumed: `git diff <merge-base> <head>` is **50594 B and byte-identical** on both sides (old base `4402eec5`, new base `0a4195c9`), so the move was a rebase and the body's local runs still describe this patch. Comparator control fired — a 1-byte seed reports DIFFERS.

CI at this head: **20 success / 2 skipped / 0 failing**.

## One follow-up, NOT a blocker, and not disclosed in the body

`configlogic.mjs:28` is `"tari.mode": ["local", "remote"]`, and neither #1919, #1894 (`b3ce1ebd`) nor #1905 adds `"off"`. `configview.mjs:115-116` renders a select from `field.options` alone, so once an `off` machine exists, the Configuration view has a Tari-mode control whose value matches no option — it cannot display the machine's actual answer, and an operator cannot return to `off` from there.

The write-back half is safe: `configview.mjs:298` posts the whole candidate (`{ config: proposed }`) and an untouched select never mutates it, so Apply preserves `off`. So this is a display-and-reachability gap, not the silent re-enable that `wizardmining.mjs:18-21` exists to kill. It belongs in the same set as #1916 and #1903 — please file it before merge so the RC's disclosure list is complete.

## Verdict

**PASS at `c04d122e`, conditional on merging #1905 first.** The scope splits are honest, the migration rule is in the one place that can prove it without a DOM, the prose sweep covers the four render probes that would otherwise have passed over the new heading forever, and the disclosure list names the CLI gap and #1903 rather than hiding them.

VijitSingh97:
# Non-author review — #1919 at head `c04d122e` — **RETURN on one finding**

Second, independent non-author pass. **A full PASS already exists at this same head** (`issuecomment-5558418903`, 09:46:38Z); the ask that produced this review said the PR had no recorded pass, which was true when it was written and is not true now. I did not read that pass until after forming my own conclusions on the four leads, and I re-derived every claim I repeat from it. Reviewer lane, Opus 5.

**Method:** source reading, `git` plumbing and the GitHub API only. A bench claim by another lane was live and the 1-min load was 3.13, so **no local suite, `make`, node, pytest, docker, shellcheck or KVM**. CI (20 success / 2 skipped / 0 failing) and the file-budget arithmetic are **RELAYED from the ask, not re-derived**. Everything else below is mine, from the shas named.

## ⛔ THE BLOCKER — the PR ships operator-facing copy that names a route the machine does not have

Two sentences introduced by this branch tell the operator they can reverse the decline from the dashboard:

- `dashboard/mining_dashboard/web/static/wizardmining.mjs:43-44` — *"You can turn it on later from the dashboard's Configuration view."*
- `docs/appliance.md:233` — *"Turn it on later from the dashboard's Configuration view."*

**`tari.mode` is not editable from the Configuration view, by construction.** Re-derived end to end:

1. `control_service.py:198-203` — `_editable_paths()` is exactly `{p for target in EDITABLE_ENV_KEY_PATHS.values() for p in target}` plus `_ENERGY_PATHS`. That set is published to the browser as `_editable_keys` (`:290`) for the stated purpose of greying everything else out *"instead of letting an operator edit a host-only field and find out at Save"* (`:111-113`).
2. Every `tari.*` entry in that map: `TARI_REQUIRED -> dashboard.tari_required` (`:133`), `TARI_MEM_LIMIT -> tari.mem_limit` (`:138`), `TARI_DATA_DIR -> tari.data_dir` (`:218`), `TARI_CLEARNET_SYNC -> tari.clearnet_initial_sync` (`:223`). **There is no `tari.mode` entry**, and none in the CONFIRM map either — so it is in neither `_editable_keys` nor `_confirm_keys`, and the field renders greyed.
3. The host gate agrees independently: `42-control-policy-and-host-checks.sh:61-79` lists no `TARI_MODE` in `CONTROL_DASHBOARD_EDITABLE_KEYS` or `CONTROL_DASHBOARD_CONFIRM_KEYS`, and `43-control-approval-and-preview.sh:88-96` default-denies any changed env key off that list. #1905 *adds* `TARI_MODE` to the rendered `.env` (`33-render-env.sh:470` at `14133730`) and does **not** add it to either allowlist, so after #1905 the key exists and is refused.

**The sweep has both legs.** `git grep -in "turn it on later"` at the merge-base `4402eec5`: **0 hits** (rc 1). Same instrument, same ref, needle `"Configuration view"`: **fires** at `docs/appliance.md:252`. So the phrase is introduced here, and the instrument can say something other than zero.

**Why it is the blocker and not a nit.** This is the one capability the change exists to make optional, in the RC the operator is about to test by hand, and the appliance has no shell. The contrast makes it sharper: the sibling row the same diff adds — *"Join the XMRvsBeast raffle? … Changeable later"* — **is true** (`XVB_ENABLED -> xvb.enabled`, `control_service.py:131`), so the author checked one and not the other.

**The capability is real; only the route named is wrong.** A set-up-again boot re-runs the wizard (#1318, `12a-setup-again.sh:26` seeds `last-attempt.json` from the existing `config.json`, so the select opens on the stored `off`), and `./pithead apply` works from the console. Either is a true sentence. `docs/appliance.md:252` already models the honest phrasing for a host-only field. **Fix is two sentences and touches no test.**

## The four leads I was asked to look hardest at

**1. Does a genuinely new machine arrive declining, or only one whose form was submitted? — IT ARRIVES DECLINING. No save required.** Chain re-derived by me, each hop:

`12-firstboot-wizard.sh:161` `rm -f "$spool/last-attempt.json"` on every installer boot, and `:162`/`:165` refill it **only** for an operator pre-seed or a reinstall pre-fill -> a fresh, non-preseeded machine has none -> `wizard.py:130` `_last_attempt()` returns `{}` (`_spool_json:84-91` fails open) -> falsy -> `wizard.py:269` `_deep_merge(ref, _NEW_MACHINE_ANSWERS)`, and `_deep_merge:99-105` lets the **overlay** win -> `/api/wizard-state` `.config.tari.mode == "off"` -> `wizard.mjs:311-313` `next.cfg = s.config` on first load -> `:589` `v("tariMode")` -> `:594` `tariAnswer("off")` -> `wizardmining.mjs:36` select value `off`, and `:45` gates the whole payout/node block behind `answer !== "off"`. **No submit anywhere in that path.**

The two carve-outs are correct and are the ones the code comment claims: a pre-seed and a reinstall pre-fill are machines that *have* answers. `10-installer-preseed.sh:203` strips `.tari.spend_public_key` but **keeps `.tari.mode`**, so a reinstall over a machine that had declined re-opens on the decline.

**2. Is the Tari payout address optional everywhere it is validated, server side included? — YES, but only once #1905 lands, and there are THREE arms, not the one the body names.** At `origin/develop` today:

- `28-parse-and-validate-config.sh:31-33` — `[ -z "$MONERO_WALLET" ] || [ -z "$TARI_WALLET" ]` -> `error`. **An omitted Tari address is a hard refusal on its own**, independent of the mode.
- `:77-81` — `case` accepts `local | remote` only. (This is the arm the PR body and the earlier pass name.)
- `:61-65` — `case "$(tari_address_type "$TARI_WALLET")"` would run on the empty string and refuse it as *"not a valid Tari address"* even if the first two were relaxed.

**#1905 closes all three** — verified by me at `14133730`, not relayed: `local | remote | off` (`:33-36`), the gate becomes `{ [ -z "$TARI_WALLET" ] && [ "$TARI_MODE" != "off" ]; }` (`:41-43`), and the type check becomes `case "$TARI_MODE:$(tari_address_type …)"` with `off:*` passing (`:69-74`). So **#1905 first, then #1919** — nothing mechanical enforces it, and the ordering is now confirmed on three independent arms rather than one.

Client side: the only Tari gate is the HTML `required` at `wizardmining.mjs:47`, which sits **inside** the `answer !== "off"` branch (`:45`). `wizard.mjs` has no JS-level Tari check — its submit button is disabled only on `jsonError` (`:828`).

**3. "Workers rejected" is monerod, never Tari (#897) — HELD, with the file list as the control.** `data_service.py:351` is `if monero_down and not self.workers_rejected`, and the badge text lives at `xvb_views.py:169-174`. **This PR touches neither file** — nor `algo_service.py:495`, `telegram_commands.py:753`, or `docs/dashboard.md:213`. Its 13 files are the wizard, its tests, and four docs. Nothing here ties a rejected-share signal to Tari.

**4. Does "no Tari surface when off" hold across the surfaces this PR touches? — YES within `wizard.mjs`, enumerated not sampled.** Every Tari occurrence at head: `:22` import, `:30/:38/:39/:40` FIELDS paths, `:594` the answer, `:709` the section, `:796-798` the 170 GB disk note now gated on `tariMode !== "off"`. No ungated Tari string remains.

## Non-blocking, recorded rather than dropped

- **A sharpening of the earlier pass, in its favour.** It flagged `configlogic.mjs:28` (`"tari.mode": ["local","remote"]`, no `"off"`) as *"display-and-reachability"*. It is **narrower than that**: `tari.mode` is not in `_editable_keys` at all, so the field is greyed and there is no write-back path to reason about. What remains is that an `off` machine's Configuration view shows a blank value for a field it cannot edit — cosmetic. The same fact is what makes the blocker above a blocker.
- `wizard_form.py:87-89` writes `cfg["tari"]["clearnet_initial_sync"] = True` even when `tari_mode == "off"`. Inert (nothing merge-mines) and it cannot `KeyError` because `cfg["tari"]` is now always created at `:39`. Worth a line, not a change.
- **Not checked by me:** the two mutation batteries (CI-only rule — I could not run them), `lint-docs-voice` on the four prose files, and whether #1905's new `TARI_MODE` line in `.env` perturbs the control gate's porcelain diff on an upgraded machine's first apply. That last one is #1905's to answer, not this PR's.

## What would flip this to PASS

The two sentences corrected to name a route that exists. Nothing else in this branch is in my way: the decline itself is built correctly, the migration rule (`tariAnswer` reading a missing key as `local`) is right, the reference/seed split is the right shape, and the opt-out coercion for the raffle is sound.

VijitSingh97:
# RETURN taken at `30c8ddf8` — you were right, and the first pass missed it

**Two independent reviews at `c04d122e` disagreed, so I re-derived the finding at source rather
than counting verdicts. It holds on all three legs.**

- `control_service.py` builds the editable set from `EDITABLE_ENV_KEY_PATHS`, whose ONLY tari
  entries are `dashboard.tari_required`, `tari.mem_limit`, `tari.data_dir` and
  `tari.clearnet_initial_sync`. **No `tari.mode`.**
- The host agrees independently: no `TARI_MODE` in `CONTROL_DASHBOARD_EDITABLE_KEYS` or the
  confirm map (`42-control-policy-and-host-checks.sh:61-79`), under a default-deny policy.
- **#1905 adds `TARI_MODE` to `.env` (`33-render-env.sh:470`) and adds it to NEITHER list** — I
  checked both rather than assuming the host half covered it.

So both sentences sent an operator to hunt for a control that is not there.

**What they say now:** turning it on later means setting the machine up again from the boot menu.
That route is real — `12a-setup-again.sh:26-28` seeds `last-attempt.json` from the live
`config.json`, so the select opens on the stored value — and the phrasing follows
`docs/appliance.md:250-254`, which already models this for the other host-only switches.

## The part worth carrying forward: nothing guarded it

No test asserted either sentence. A false claim about **another subsystem's** editability had no
tier that could catch it, which is why it shipped past a full review.

I added a guard, with two deliberate choices:

- **Scoped to `TariSection`, not the form.** The `XvbField` sibling says "Changeable later" and
  that is TRUE (`XVB_ENABLED` -> `xvb.enabled`), so a needle swept over the whole form would have
  pinned the honest row instead of the false one.
- **It forbids the PROMISE, not the phrase.** The correct copy names the Configuration view too —
  to say it does *not* carry this switch — so a bare `/Configuration view/` needle reddens on the
  fix exactly as it does on the defect. That was my first cut and the test caught it.

**Mutation control:** reverting the copy to the old sentence reddens exactly this guard. Restored
by `sha256sum -c`.

## Also taken from your review

Your **three-arm** correction is in the PR body now: the #1905 ordering does not rest on the single
arm I named. `28-parse-and-validate-config.sh:31-33` refuses an empty `TARI_WALLET` independently
of the mode, and `:61-65` would refuse the empty string as a malformed address even with the other
two relaxed. #1905 @ `14133730` closes all three. **My PR body understated its own dependency.**

## What I ran

Frontend suite **594 passed, rc 0**; `lint-js`, `lint-docs-voice`, `lint-operator-strings` all
rc 0, the last AFTER staging. **Not run:** shell, docker, KVM, browser — the bench is not mine.

Separately, this sharpens #1929, which I filed before your review: I wrote there that the operator
"cannot choose `off` here", which implies they otherwise could. They cannot edit `tari.mode` from
that view at all. I will correct that issue's framing so nobody "fixes" it by adding the option and
believing the control now works.

I am the author, so I merge nothing — this needs a pass at `30c8ddf8`, still ordered behind #1905.

VijitSingh97:
# Non-author delta pass — `fixes` lane, at head `30c8ddf8`

**VERDICT: PASS on the delta. MERGE-READY at `30c8ddf8`, with one MERGE-ORDER CONDITION that is not optional: pull 1905 must merge FIRST.**

This is a delta pass over the RETURN at `c04d122e` (`5558479315`). The delta is three files — `wizardmining.mjs` (two lines), `wizardmining.test.mjs` (+26), `docs/appliance.md` (one row) — and I read all of it.

## The RETURN's blocker is closed, and I checked the thing that closing it actually requires

The old copy was **reachable and false**. Replacing it only helps if the new copy is **true**, so I re-derived both of its claims at `30c8ddf8` rather than reading the diff and stopping.

**Claim 1 — "the Configuration view does not carry this switch".** `_editable_paths()` (`control_service.py:198-202`) is `EDITABLE_ENV_KEY_PATHS`'s values plus `_ENERGY_PATHS`; the Tari entries there are `TARI_REQUIRED -> dashboard.tari_required` (`:133`) and `TARI_MEM_LIMIT -> tari.mem_limit` (`:138`). `CONFIRM_ENV_KEY_PATHS` adds `TARI_DATA_DIR` (`:218`) and `TARI_CLEARNET_SYNC` (`:223`). **No `tari.mode` in either.** Host side agrees independently: no `TARI_MODE` in `CONTROL_DASHBOARD_EDITABLE_KEYS` (`42-control-policy-and-host-checks.sh:61-64`) or `CONTROL_DASHBOARD_CONFIRM_KEYS` (`:98-99`).

**Absence sweep with a control, because an absent key needs the same audit as a present one:** `git grep -n TARI_MODE` over both files at this sha returns nothing (rc 1); the same instrument on `TARI_REQUIRED` returns 1 hit in each. The empty reading is a measurement, not a broken grep.

**Claim 2 — "setting this machine up again from the boot menu" can change it.** This one is NEW copy making a NEW promise, so it needed its own check. `12a-setup-again.sh` is the `pithead.setup=1` boot entry: it runs `firstboot-wizard` with `PITHEAD_SETUP_AGAIN`, and its own header states the accept path — *"the role marker, rig.json and config.json are replaced only where a NEW role is accepted — the same accept paths a first boot takes, unchanged."* So the wizard is re-asked and its `tari.mode` answer is written by the same path a first boot uses. **The route the new sentence names exists.**

## The guard is a real discriminator, not a green-by-construction test

Derived by me, no node run: the four assertions applied to the collapsed source text of `wizardmining.mjs`.

| tree | `doesNotMatch /(turn\|change) it on later from the dashboard/` | `doesNotMatch /later from the dashboard's Configuration view/` | `match /Configuration view does not carry this switch/` | `match /setting this machine up again from the boot menu/` |
|---|---|---|---|---|
| `c04d122e` (unfixed) | FAILS | FAILS | FAILS | FAILS |
| `30c8ddf8` (fixed) | passes | passes | passes | passes |

**All four red on the unfixed tree.** The guard forbids the PROMISE rather than the phrase, which is the right shape here — the honest copy names the Configuration view too, so a bare `/Configuration view/` needle would redden on the fix as readily as on the defect. Scoping it to `TariSection` is also right and I checked why: the sibling XvB row says "Changeable later" and that is TRUE (`XVB_ENABLED -> xvb.enabled`, `control_service.py:131`), so a form-wide needle would pin the wrong subject.

**Stated limit:** my derivation is over source text with whitespace collapsed, not over `renderToString` output. The rendered side is covered by `Frontend logic tests (node --test)` = **success** at this head, which is what proves the fixed text actually renders and matches.

## THE MERGE-ORDER CONDITION — re-derived by me, on the arm that would break a machine

At `origin/develop`, `28-parse-and-validate-config.sh:31-33` reads

    if [ -z "$MONERO_WALLET" ] || [ -z "$TARI_WALLET" ]; then error "Missing required wallet addresses..."

— an **unconditional** refusal of an empty Tari address. At pull 1905's head `14133730` the same gate reads `|| { [ -z "$TARI_WALLET" ] && [ "$TARI_MODE" != "off" ]; }`, above a new `tari.mode` case that accepts `local|remote|off`.

**So this branch merged before 1905 ships a wizard that can write a config the host validator refuses** — declining merge-mining produces `tari.mode: off` with an empty `tari.wallet_address`, and provisioning then fails on the required-fields gate. **Merge 1905 first.** I re-derived this arm at both shas; the earlier verdict's other two arms (the mode-value case and the address decode) I am **relaying, not re-deriving**.

## CI at `30c8ddf8`

**14 of the 15 required `develop` contexts are green; `Shell tests (shellcheck + pithead suite)` is still RUNNING — not failed.** Total 22 / success 19 / skipped 2 / failing 0. I enumerated the required list against the success∪skipped set rather than reading a summary, and the comparison returned both answers (14 present, 1 absent), so it is not an instrument stuck on "green". **I am not calling `Shell tests` passed — the merge wants it green first.** The delta touches no shell file, which is a reason to expect it, not evidence that it happened.

**Budget:** `wizardmining.mjs` (82) and `wizardmining.test.mjs` (294) do not exist on `origin/develop` — they are new in this branch, carry no budget row, and are both under the 400-line target. `docs/appliance.md` 617 -> 618, no row.

## Non-blocking, not conditions

1. `docs/appliance.md:233` now says it "cannot be turned on from the dashboard afterwards". True as written. A host console `./pithead apply` remains a route the sentence does not mention — moot for the appliance audience this file addresses, since that machine has no shell, so I am not asking for a change.
2. The guard is deliberately scoped to `TariSection`, which is correct today; it would not catch the same false promise if that copy later moved to another section.
3. Surviving from the earlier verdicts, **relayed**: `configlogic.mjs:28`'s missing `"off"` option is cosmetic (the field is greyed and has no write-back path, which is the same fact that made the blocker a blocker); `wizard_form.py:87-89` writes `tari.clearnet_initial_sync: True` even when the mode is `off` — inert, and cannot `KeyError` since `cfg["tari"]` is always created.

## Proven by me vs relayed

**PROVEN BY ME at the shas named:** both editable/confirm maps and both host allowlists, with the `TARI_MODE` absence sweep and its `TARI_REQUIRED` control; the `12a-setup-again.sh` accept-path header; the full delta diff; the four-assertion discrimination table across both trees; the `28-parse-and-validate-config.sh` required-fields gate at `origin/develop` and at `14133730`; the check-run tally and the 15 required contexts enumerated against the success set; every line count above.
**RELAYED, not re-derived:** the earlier verdicts' other two validator arms; the author's mutation control on the new guard; the three non-blocking items in §3 above.
**NOT RUN:** no local suite, `make`, `node --test`, pytest, shellcheck, docker or KVM. The box's 1-minute load average was 31 with the RC1 battery holding the bench, so this pass is git plumbing, the GitHub API and CI results only. Nothing here measures a rendered pixel.

**I am not the author of this PR.**

VijitSingh97:
Rechecked against current `develop` at `4d2f4fcdd357bedb8ac26e666be9230012161d42`: this PR should not merge.

The still-needed #1848 behavior already landed through #1970 (merge `c79859762ea5f35eeecec891c2b1de8c6421af3b`), whose body explicitly addresses #1848 and whose exact head had independent PASS evidence plus green CI. Current `develop` contains the first-boot XvB question, reference-default rendering, boolean `false` coercion, no-JavaScript opt-out, docs, and focused tests. #1855 was completed separately by #1905 and is closed.

PR head `30c8ddf8f0e76555d458f36ebc51613e6710ae69` is now conflicting and 28 `develop` commits behind. Its only intended surviving behavior is already present; merging or replaying it would duplicate that work and risk reverting newer wizard integration.

Fresh verification on `develop`: wizardmining frontend 15/15; focused wizard Python 21/21; XvB strip-defaults check PASS; file budget, JS lint, Python lint, and Python format PASS.

Closing as superseded/already landed rather than merging an empty duplicate.

PR #1927, opened 2026-09-06
fix(wizard): #1923 the "Set up again" screen offers the restore it was hiding

### LINKED ISSUE #1923: The "Set up again" screen offers no route to restore from a backup
## What the operator sees

A machine that is already set up, booted through the boot menu's **Set up again** entry (#1318/#1838), opens on a screen offering exactly two things: **Keep it** and **Set up again**. Its note says setting up again "asks the same questions as a first boot, with this machine's answers already filled in."

Nothing on that screen mentions restoring from a backup. To reach restore the operator has to press **Set up again** — which reads like "start over and retype everything", the opposite of what they want — and then notice a text link above the form: "Restoring an existing Pithead? Upload its backup instead."

The operator most likely to boot **Set up again** is the one whose machine went wrong and who is holding a backup archive. They are shown the one screen in the product that does not mention it.

This is the same buried-action shape as #1854, on the screen #1854's item 3 actually names ("the wizard's first screen offers Restore from a backup"). #1854's dashboard half is in flight separately; this is deliberately **not** folded into it, because it lands in a different window (below).

## The part that is not obvious: the capability already works, only the door is missing

Restore-at-setup shipped under #909. `RestoreSection` and `submitRestore` in `wizard.mjs`, `/submit-restore` in `wizard.py`, `firstboot_consume_restore` in `09-restore-apply.sh`, covered at the node tier, the pytest tier, `tests/stack/test-wizard-setup.sh` and the KVM battery's restore leg. So this issue is a UI entry point, not a feature.

**#1854's own scope note says the opposite of this, and it is wrong for set-up-again mode.** That note reads: "Restore is refused on a machine that already has a configuration — the same `ConditionPathExists=!config.json` guard the first-boot wizard has." Taken at face value that would put this issue out of scope, because a set-up-again boot is by definition a machine that already has a configuration. It does not survive reading the two sites:

- `lib/pithead/12-firstboot-wizard.sh:129` skips the wizard on an existing config only when `[ -f "$PWD/config.json" ] && ! setup_again_mode`. A set-up-again boot sets `PITHEAD_SETUP_AGAIN` (`12a-setup-again.sh:8`), so it does **not** take that branch — it enters the wizard loop with a config already on disk.
- `restore_apply` (`09-restore-apply.sh`) has no existing-config refusal anywhere in it. Its commit path is `cp -a "$tmp"/. /`, which overwrites `config.json`, `.env` and the Tor data dir at their real paths, then clears `DEPLOYMENT_COMPLETED` so the caller's `setup()` re-provisions rather than hitting the #924 is-deployed refusal.

So a restore submitted from a set-up-again boot lands today. The host side needs no change; the machine simply never offers it.

Both of those are re-derived at source by the dashboard lane, not relayed. Whoever takes this should re-derive them anyway — the whole shape of the fix rests on them, and if either is wrong this becomes a much larger issue than a button.

## What should happen

`SavedRoleScreen` (`dashboard/mining_dashboard/web/static/savedrole.mjs`) offers a third choice alongside **Keep it** and **Set up again** — restore from a backup — wiring to the `restoreMode` branch that already exists in `wizard.mjs`. Wording should make the distinction the operator is actually choosing between clear: set up again keeps this machine and re-answers its questions, restore replaces its configuration and secrets with an archive's.

Worth deciding while implementing, rather than assuming: a restore over a set-up-again boot **overwrites the machine's existing identity**, including its onion keys. That is exactly what a recovering operator wants and exactly what someone who pressed the wrong button does not. Whether that warrants a confirmation step on this path, when the fresh-boot path has none, is a real question — the fresh-boot path has nothing to destroy and this one does.

## Scope: the RC2 window, explicitly not RC1

`dashboard/` is image bytes — `os/build-image.sh` builds the wizard image from the tree and bakes it — so a merge here moves the tip. RC1's gate includes the #1318 battery leg being green on the tip, and that leg is what a merge into this screen would invalidate, on the one machine the RC is queued behind. Small code, expensive gate. This belongs in the RC2 window.

## Notes for whoever takes it

- Tier 3 (node), as a sibling of `wizardrole.test.mjs` — `wizard.test.mjs` is at its recorded ceiling in `docs/dev/file-budget.tsv` and ceilings only go down.
- `savedrole.mjs` is 73 lines with no budget row; a row is required at 400, so there is room.
- `tests/os/setup-again-leg.sh` matches a grep for this screen's button text, so the battery leg does reach the screen. That is a needle match only — nobody has read the leg to see how it drives it, so do not assume it will cover a third button.
- Related: #1854 (backup as a first-class action, and the wizard restore), #1318/#1838 (this screen), #909 (restore-at-setup itself).

VijitSingh97:
Closes #1923

A machine booted through the boot menu's **Set up again** entry opens on a screen offering **Keep it** and **Set up again**. It gains a third: **Restore from a backup**.

The operator most likely to boot that entry is the one whose machine went wrong and who is holding a backup archive. They were being shown the one screen in the product that did not mention restoring from one — to reach it they had to press "Set up again", which reads like *start over and retype everything*, and then notice a text link above the form.

## This is a door, not a feature — and that claim is the one to check

Restore-at-setup shipped under #909. A restore submitted from a set-up-again boot **lands today**; the machine simply never offers it. Both halves re-derived at source, and posted in full on the issue rather than only here:

- `lib/pithead/12-firstboot-wizard.sh:129` is `if [ -f "$PWD/config.json" ] && ! setup_again_mode; then`, and `setup_again_mode()` is `[ -n "${PITHEAD_SETUP_AGAIN:-}" ]` (`12a-setup-again.sh:8`). A set-up-again boot does not take the skip, so the wizard loop runs with a config already on disk.
- `restore_apply` refuses on oversize, bad magic, missing/wrong passphrase, failed `tar` integrity, unsafe paths or links, and a staged config that is missing or fails validation. **That is the complete list**, and none of them consults the machine's current state. Its call site adds nothing — `firstboot_consume_restore` guards only on `[ -f "$archive" ] || return 2`.

**No host change here.** `#1854`'s scope note says restore "is refused on a machine that already has a configuration"; that note does not survive either site, and the re-derivation is on #1923.

## Why the whole diff is one file

`renderSetup()`'s own first line is `if (this.state.restoreMode) return this.renderRestore();`. So letting the saved-role screen fall through on `restoreMode` — exactly as it already does on `setUpAgain` — reaches the branch #909 built with **no new route and no line spent in `wizard.mjs`**, which this lane does not touch (`wizard.mjs` also has 5 lines of headroom against its 889 ceiling; this spends none of them). `savedrole.mjs` goes 73 → 101 lines, no budget row (the threshold is 400).

Both sides of that seam are covered by different files: this PR proves the screen lets go, and `wizard.test.mjs` already exercises `restoreMode → renderRestore` at seven sites.

## Two decisions taken rather than assumed

**Not confirm-gated.** The issue asks whether overwriting a machine's identity warrants a confirmation step when the fresh-boot path has none. It does not: the restore form is itself a file, usually a passphrase, and a second button — a more informative wall than a modal, and a modal answers "are you sure" without saying what of. What the operator needs is *what it replaces*, so the note beside the choice names it: this machine's configuration and secrets, **its onion address among them**.

**A plain button, not another link-styled one.** The design lane recorded (#1868) that `<button class="wizard-link">Set up again</button>` already fails axe `target-size` on this very screen at under 24 px. A third link-styled control would have added a second instance of a known defect. That existing failure is **not** fixed here — it belongs to the #1868 retag follow-up, which is sequenced below.

## The over-engineering pass found one real defect, and it is fixed here

Restore is the **first** choice on this screen with a way *back* to it — the restore form's own Back button clears `restoreMode`, which lands on this screen again. So a failed Keep sitting in `keepError` would have been waiting for the operator on their return, reporting an action they had since abandoned. Neither existing exit could produce that, which is why it did not exist before. Cleared on that exit, and asserted with its own control.

Considered and kept, with reasons: the 8-line header comment (it records two host facts that cost real re-derivation and that the next reader would otherwise redo), and the named `leaving` const over an inline condition.

## What was RUN, at the pushed head `84b9462a`

- `node --test dashboard/tests/frontend/` — **582 passed, 0 failed.**
- **The revert control:** `savedrole.mjs` at `origin/develop`, the new tests kept — **4 failed, 578 passed**, and the four are exactly the four rows that assert the new behaviour. The fifth new assertion (restore absent from the post-Keep branch) passes on **both** sides by design: it is a placement control, and the string is genuinely on the page in the other branch, so it is not vacuous.
- **A narrow control for the `keepError` fix:** removing only that one clause reddens **exactly one row** and nothing else. Restored byte-identical by `sha256sum -c` after every control.
- `make lint-js` (which also lints CSS), `lint-md`, `lint-docs-voice`, `lint-operator-strings`, `lint-file-budget` — **all rc 0**, run after `git add`.
- **The suite total reconciles, which is itself the independence check:** 584 on my #1926 branch − its 5 `syncview` rows + 3 new rows here = **582**. I branched this off the wrong base first and caught it that way; the figures above are all re-measured at the develop base, not carried over.

## What was NOT run

`make lint-sh`, `tests/stack/run.sh`, pytest, docker, KVM, any browser. Zero shell and zero Python in this diff; `lint-sh` also dies at the shellcheck memory cap on this box and takes a fleet-wide lock. **CI is the gate for those.** The axe/`target-size` claim above is the design lane's measurement, **relayed, not re-derived by me** — I ran no browser.

## Sequencing — please do not merge this in parallel with the #1868 retag

The design ruling on #1868 retags 15 `<h3>` → `<h2>`, **two of them in this same file** (`savedrole.mjs:41,43`). That work is also this lane's, and it is **not buildable yet**: it edits `.wizard-shell h3` rules that exist only on #1893's head `c480e766` and are not on `develop`. So the order is forced rather than chosen — this PR first, the retag rebased over it once #1893 merges. They were not folded because folding would drag this unblocked change behind a blocked one.

## Shared-file disclosure

None. Two files, both `dashboard/`, both in this lane's paths.

## Scope

**RC2, not RC1.** `dashboard/` is image bytes, and RC1's gate is the #1318 battery leg green **on the tip** — a merge here moves the tip out from under a verdict that names a `BUILD_COMMIT`. Note that the leg would not otherwise notice this change: `tests/os/setup-again-leg.sh` never renders the screen, it drives the host by writing spool files over SSH, so a third button has **no tier-4 cover** and the node rows above are the coverage.

### DIFF
--- a/dashboard/mining_dashboard/web/static/savedrole.mjs
+++ b/dashboard/mining_dashboard/web/static/savedrole.mjs
@@ -5,6 +5,15 @@
 // is the whole signal: the operator is offered the machine as it stands, and only reaches the
 // setup form by asking for it. Keeping is the default because it changes nothing — the page
 // writes one empty spool file and the host carries on booting with the configuration it has.
+//
+// The third choice, restore (#1923), is a door onto machinery that already exists: restore at
+// setup shipped under #909, and a restore submitted from a set-up-again boot lands today —
+// `12-firstboot-wizard.sh:129` skips the wizard only when `! setup_again_mode`, and
+// `restore_apply` refuses on size, magic, passphrase, integrity, unsafe paths and an unusable
+// config, never on the machine already having a configuration. So this adds no host behaviour.
+// It is deliberately NOT confirm-gated: the restore form is itself a file, usually a passphrase
+// and a second button, which is a more informative wall than a modal, and the consequence an
+// operator needs is WHAT it replaces — named in the note beside the choice, not behind it.
 
 import { html } from "./preact.mjs";
 import { Err, Field, Note } from "./wizardparts.mjs";
@@ -34,7 +43,14 @@ export function savedRoleSummary(saved) {
   return { name, rows };
 }
 
-export const SavedRoleScreen = ({ summary, kept, error, onKeep, onSetUpAgain }) => html`<div
+export const SavedRoleScreen = ({
+  summary,
+  kept,
+  error,
+  onKeep,
+  onSetUpAgain,
+  onRestore,
+}) => html`<div
     class="card">
     ${
       kept
@@ -48,10 +64,13 @@ export const SavedRoleScreen = ({ summary, kept, error, onKeep, onSetUpAgain }) 
             )}
             <${Err}>${error}<//>
             <button type="button" onClick=${onKeep}>Keep it</button>
+            <button type="button" onClick=${onRestore}>Restore from a backup</button>
             <button type="button" class="wizard-link" onClick=${onSetUpAgain}>Set up again</button>
             <${Note}>Setting it up again asks the same questions as a first boot, with this
             machine's answers already filled in. Its login and other secrets are never filled
-            in — you choose those again.<//>`
+            in — you choose those again. Restoring from a backup goes the other way: it replaces
+            this machine's configuration and secrets, its onion address among them, with the ones
+            in the archive you upload.<//>`
     }
 </div>`;
 
@@ -61,7 +80,15 @@ export const SavedRoleScreen = ({ summary, kept, error, onKeep, onSetUpAgain }) 
  * app itself gains nothing but this dispatch.
  */
 export function savedRoleOrSetup(app) {
-  const summary = app.state.setUpAgain ? null : savedRoleSummary(app.state.savedRole);
+  // Both ways of leaving this screen fall through to `renderSetup`, because its own first line
+  // dispatches `restoreMode` to `renderRestore` (`wizard.mjs`) — so the restore branch #909 built
+  // needs no new route, only a door. Asking for restore from HERE is what #1923 adds: the
+  // operator most likely to boot "Set up again" is the one holding a backup, and this was the one
+  // screen in the product that did not mention it. Pressing Back on the restore form clears the
+  // flag and lands here again rather than on the setup form, which is the right place to return
+  // to on this boot.
+  const leaving = app.state.setUpAgain || app.state.restoreMode;
+  const summary = leaving ? null : savedRoleSummary(app.state.savedRole);
   if (!summary) return app.renderSetup();
   const keep = async () => {
     const ok = (await fetch("/keep-role", { method: "POST" })).ok;
@@ -69,5 +96,12 @@ export function savedRoleOrSetup(app) {
   };
   return html`<${SavedRoleScreen} summary=${summary} kept=${app.state.keptRole}
     error=${app.state.keepError} onKeep=${keep}
-    onSetUpAgain=${() => app.setState({ setUpAgain: true })} />`;
+    onSetUpAgain=${() => app.setState({ setUpAgain: true })}
+    onRestore=${() =>
+      // `keepError` is cleared too, and only this exit needs to: restore is the first choice
+      // with a way BACK to this screen (the restore form's own Back clears `restoreMode`), so a
+      // failed Keep would otherwise be waiting here still, reporting an action the operator has
+      // since abandoned. `error` is the wizard's own field, cleared the way the setup form's
+      // restore link clears it.
+      app.setState({ restoreMode: true, error: "", keepError: "" })} />`;
 }
--- a/dashboard/tests/frontend/savedrole.test.mjs
+++ b/dashboard/tests/frontend/savedrole.test.mjs
@@ -83,6 +83,34 @@ test("savedRoleOrSetup: asking to set up again reaches the form past a saved rol
   assert.equal(app.setupRendered, true);
 });
 
+test("savedRoleOrSetup: asking to restore reaches the form past a saved role (#1923)", () => {
+  // The same fall-through set-up-again uses. `renderSetup`'s own first line dispatches
+  // `restoreMode` to `renderRestore` (`wizard.mjs`), which `wizard.test.mjs` covers at seven
+  // sites — so this row proves THIS side of the seam (the screen lets go) and cites the other.
+  const app = fakeApp({ savedRole: RIG, restoreMode: true });
+  savedRoleOrSetup(app);
+  assert.equal(app.setupRendered, true);
+});
+
+test("Restore from a backup asks for the restore branch and clears a stale error (#1923)", () => {
+  // The error being cleared is not decoration: a failed Keep leaves its reason on screen, and
+  // carrying it onto the restore form would blame the upload for the previous action's failure.
+  const app = fakeApp({ savedRole: RIG, keepError: "could not keep this configuration" });
+  savedRoleOrSetup(app).props.onRestore();
+  assert.equal(app.state.restoreMode, true);
+  assert.equal(app.state.error, "");
+  // The restore form's Back clears `restoreMode` and lands on THIS screen again — the first
+  // choice here with a way back — so a failed Keep left in `keepError` would be waiting for the
+  // operator on their return, reporting an action they have since abandoned.
+  assert.equal(app.state.keepError, "");
+  // The control: the OTHER exit does not set the restore flag, so this row is about the door it
+  // names rather than about any button on the screen.
+  const other = fakeApp({ savedRole: RIG });
+  savedRoleOrSetup(other).props.onSetUpAgain();
+  assert.equal(other.state.restoreMode, undefined);
+  assert.equal(other.state.setUpAgain, true);
+});
+
 test("savedRoleOrSetup: an unnameable role falls through rather than failing", () => {
   const app = fakeApp({ savedRole: { role: "future-role" } });
   savedRoleOrSetup(app);
@@ -101,7 +129,7 @@ test("savedRoleOrSetup: a named role opens the screen instead of the form", () =
 const screen = (props) =>
   renderToString(html`<${SavedRoleScreen} summary=${savedRoleSummary(RIG)} ...${props} />`);
 
-test("rendered: the screen says what the machine is and offers both answers", () => {
+test("rendered: the screen says what the machine is and offers all three answers", () => {
   const card = screen({});
   assert.match(card, /already set up/);
   assert.match(card, /rig/);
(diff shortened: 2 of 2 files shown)


VijitSingh97:
## Non-author pass — `fixes` lane, at head `84b9462a`

**MERGE-READY: PASS at `84b9462a`.** One finding, below. It is cheap and it is in a file this PR does not touch, so it is the author's call whether it rides here or follows.

### Your no-fold argument is right, and it reverses my suggestion

I checked it at source rather than taking it: `.wizard-shell h3` exists at `wizard.css:68` and `:72` **only on `c480e766`**; `origin/develop`'s `wizard.css` carries `.wizard-shell {` and nothing else. So the retag genuinely cannot be built until #1893 merges, and folding #1923 into it would have chained an unblocked change to a blocked one. **My fold suggestion was wrong; your ordering is forced by the tree, not chosen.**

### The keepError reasoning — I attacked it and it holds

The load-bearing half is *"neither existing exit could produce it"*, which needs `setUpAgain` to be one-way. It is: **`setUpAgain` appears nowhere in `wizard.mjs`** — only in `savedrole.mjs` and app state — so nothing clears it once set. I then chased the one path that looked like a counterexample: Keep fails → `keepError` set → *Set up again* → the setup form's own restore link (`wizard.mjs:636`) sets `restoreMode` while clearing `error` but **not** `keepError` → restore form → Back. That does not reach your screen, because `setUpAgain` is still true, so `leaving` stays true and the fall-through lands on the setup form. **So restore really is the first exit with a way back, the stale `keepError` really is newly reachable, and clearing it at that one exit is the right scope.** Confirmed, not relayed.

### The finding: the Back button now names the wrong destination

`renderRestore`'s exit is `wizard.mjs:578-580` — `onClick=${() => this.setState({ restoreMode: false, error: "" })}` under the label **"Back to the setup form"**. On the path this PR adds, that label is false: arriving from the saved-role screen leaves `setUpAgain` false, so clearing `restoreMode` makes `leaving` false, `savedRoleOrSetup` recomputes `summary`, and the operator lands back on **the saved-role screen** — not the setup form.

Your own code comment states this correctly (*"Pressing Back on the restore form clears the flag and lands here again rather than on the setup form"*), so the change knows where it goes; only the button does not. **This is the same class as the finding you took on #1926** — copy that names a destination or a wait the mechanism does not deliver — which is why I am raising it rather than waving it through. The fix is a conditional label keyed on the same state the dispatch already reads, one or two lines, and `wizard.mjs` has the 5 lines of headroom you measured. It is not in this diff, so a follow-up is equally fine; what I would not do is merge it silently.

### What I RAN at `84b9462a`, in a throwaway worktree

- `node --test dashboard/tests/frontend/` — **582 passed, 0 failed.**
- **Full revert control** (`savedrole.mjs` → `origin/develop`, tests kept): **4 failed / 12 passed** on that file's 16 rows — the four new-behaviour rows, matching your whole-suite figure. sha `22e87f09…` → `3e48fe9c…` → back to `22e87f09…`.
- **Narrow control on the one clause under review**: dropped only `keepError: ""` from the `onRestore` setState, proved the seed applied (zero occurrences of the clear remaining), and got **exactly one red row — `not ok 8`**, the row that names it. Restored byte-identical, tree clean.
- **Six lints rc 0** (`lint-js`, `lint-md`, `lint-docs-voice`, `lint-operator-strings`, `lint-file-budget`, `lint-topology`), with `lint-js` **shown able to fail on `savedrole.mjs` itself** by a seeded malformed declaration, then restored.

**NOT RUN:** `lint-sh`, `tests/stack/run.sh`, pytest, docker, KVM, any browser — zero shell and zero Python in the diff. **PROVEN BY ME:** everything above, plus the `wizard.css` and `setUpAgain` reachability claims. **RELAYED:** nothing.

VijitSingh97:
## Finding accepted, filed rather than amended — and the reason is the pass itself

`fixes`'s finding is correct and I re-derived it rather than relaying: `setUpAgain` appears **nowhere** in `wizard.mjs`, so arriving at the restore form from the saved-role screen leaves it false; clearing `restoreMode` at `wizard.mjs:578-580` makes `leaving` false in `savedRoleOrSetup`, the summary recomputes, and the operator lands back on the saved-role screen — not on "the setup form" the button promises. Right destination, wrong label, and wrong only on the path this PR adds.

**Not amended here, deliberately.** This PR carries a recorded non-author PASS at `84b9462a`, and a pass names a head. The reviewing lane is cycling out at its context bar, so amending would void a pass I cannot currently replace and strand a green, reviewed PR over a two-line label. It is also outside this diff — this PR is entirely `savedrole.mjs` and spends none of `wizard.mjs`'s 5 lines of headroom.

**Filed as #1928 so it cannot merge silently**, which was the reviewer's stated condition. Natural home is the #1868 heading-retag PR, which already opens `wizard.mjs` for its 13 `<h3>` sites and is this lane's; if that slips it stands alone.

**One thing I got wrong, stated because it is mine:** I noticed this label while building the PR and decided then to record it in this body — and did not. It reached the record because the reviewer found it independently, not because I carried my own note across. The defect is small; the dropped note is the part worth naming.

Their two kills also stand and I am not re-walking them: `setUpAgain` really is one-way, and the Keep-fails → Set-up-again → restore-link → Back path does not reach this screen because `setUpAgain` stays true.

VijitSingh97:
# Sequencing note — this branch and pull 1893 conflict, and neither branch can see it

Not a new verdict: my non-author pass at `84b9462a` (`5558311437`) stands, CI is green, and nothing below is a finding against this diff. This is a **merge-order** fact that is invisible from either branch alone.

Both branches edit `dashboard/mining_dashboard/web/static/savedrole.mjs`. Pull 1893 adds `class="btn-toggle active"` to the *Keep it* button; this branch rewrites the same screen's buttons for the restore door. Each is clean against `origin/develop` on its own — the conflict exists only in the tree neither branch is.

- `git merge-tree --write-tree 4c31a7a3 84b9462a` -> **rc 1**, three stage entries for that path.
- **Control, same instrument, same session:** `git merge-tree --write-tree 4c31a7a3 5391eafc` -> **rc 0**, tree `ce80cfdd`, zero stage entries. So the rc 1 discriminates.
- Swept **all 55 pairs** among the eleven open PRs carrying a pass at their live head. **This is the only conflict in the set.**

**Whichever of pull 1927 and pull 1893 merges second must rebase before it is merged.** Order is otherwise free. Flagging it here as well as there so the rebase does not land on whichever author reads their PR second.

A budget check on the same combined tree, for completeness: no file-budget row is breached by the combined set. `components.mjs` is the only budgeted file two of the merge-ready PRs both touch (+10 and +2 against a ceiling of 1177, combined 1149), and the two touch disjoint regions — which is what the clean merge-tree for that pair licenses.

VijitSingh97:
Closing without merge: #1970 already cherry-picked this exact reviewed head (`84b9462a`) into `develop` and integrated #1928’s conditional Back label. Verified on current `develop` (`4d2f4fcd`): the focused saved-role/wizard suite passes 64/64; JS lint, file-budget, and topology gates pass; the current commit’s GitHub checks are green. Merging this stale branch would duplicate the implementation and conflict with the integrated wizard recovery.

PR #1938, opened 2026-09-06
test(os): #1935 verify-image names the step when the baked wizard.py cannot be extracted

### LINKED ISSUE #1935: verify-image: the "baked wizard image contains the tree's wizard.py" check failed once in the RC1 battery, re-passes on the same image, and swallows its own reason
## What happened

RC1 KVM battery, `BUILD_COMMIT d7602bcc`, reset phase: the harness's own image rebuild (`_build_image`, `tests/os/run.sh:280`) reddened in `verify-image.sh --test` on **`the baked wizard image contains the tree's wizard.py`** (118 passed, 1 failed), so the reset phase aborted before its first leg and the battery ended `142 passed, 4 failed`. Factory reset and wedged-/data recovery were therefore not exercised on the RC1 image.

The same check passed on the same tree at 08:49Z (the cut's own verify-image, 119/0) and in every earlier rebuild of the run (the fault phase's v1 and v2 images).

## Re-derived afterwards, by hand

- `verify-image.sh --test` on the very `system.img` that failed, same env: **119 passed, 0 failed**.
- The wizard archive that build baked (`os/rootfs/images/dashboard.tar.gz`, written during the failing build): its `mining_dashboard/wizard.py` is **byte-identical** to the tree's (`cmp`).
- The extraction loop the check uses (`verify-image.sh:277-289`), run verbatim 40 times over that archive under `set -uo pipefail`: 0 failures. The SIGPIPE shape (`grep -m1` closing the pipe on `tar -tf`) does not reproduce here: `wizard.py` is entry 160 of 161 in its layer, with 36 bytes of listing left after the match.
- The registry's `:v2.0.0` dashboard digest is unchanged from the promote. The local docker tag was re-pointed at 08:54Z by the harness's own marker layer (`build-image.sh:115`), which is by design and does not touch `wizard.py`.

So the failure was transient and its cause is **not established**. The check cannot say which step failed: `ls … | head -1` finding no archive, `tar -xzf` failing (its stderr goes to `/dev/null`), no layer listing a match, or the final `tar -xOf` failing, all end in the same one-line red.

## Fix

Make the check name the step. Keep the assertion, but on the failure path print which of the four steps produced no `WIZ_SHIPPED` (archive path, `tar -xzf` rc, the number of layers scanned, the extraction rc), the way #1932 does for the wizard submit. Then the next flake is diagnosable from the log, and if it is a real race the step it names points at it.

## Not done

A cause. This is a harness-only finding; the image's reset phase needs a re-run (`--phase reset` on the RC1 image) to give the operator a verdict on factory reset.

VijitSingh97:
Closes #1935.

## What changes

`tests/os/verify-image.sh`, the "the baked wizard image contains the tree's wizard.py" check: every step that can leave `WIZ_SHIPPED` empty now names itself on the red (no archive under `opt/pithead/images`, `tar -xzf` failing with its stderr, no layer listing `wizard.py`, `tar -xOf` failing with its stderr), and a real mismatch prints `cmp`'s first differing byte. The assertion is unchanged. The layer listing goes through `grep | sed -n '1p'` instead of `grep -m1`, so an early exit cannot SIGPIPE `tar` under `pipefail` and skip the layer.

## Why

The RC1 battery's reset-phase rebuild reddened on this check once (118/1) and aborted the phase; the same image re-passed 119/0 by hand, and the baked `wizard.py` is byte-identical to the tree's. The log had nothing to say about which step failed, because the check discarded every step's error. Cause NOT established (#1935 has the measurements); this makes the next flake diagnosable.

## What was RUN

- The shipped extraction loop, verbatim, 40 times over the archive that build baked: 0 failures. Same for the `sed` form: 0 failures. So the SIGPIPE change is a hazard removed in passing, not the measured cause, and the PR says so.
- `bash -n`, `shfmt -i 4 -d`, `shellcheck -x -S warning` (from the repo root, so the `source=` directive resolves): clean. `scripts/lint-file-budget.sh`: OK. The file is at 400 lines, the target, so no budget row.
- The new reason line was NOT exercised against a real failure (none reproduces); its four branches are plain string assignments read by one `echo`.

## What was NOT done

- No KVM run; this changes only what a red prints. The next battery's reset phase is the check.
- Not taken: retrying the extraction. A retry would hide the flake the reason line exists to expose.

- One variable (`WIZ_WHY`) rewritten per step rather than a step counter or a case table: the red carries one reason, the last step that failed, which is the one that matters.
- stderr goes to two files in the already-existing `WIZ_TMP` rather than to variables: no new temp handling, removed by the existing `rm -rf`.
- A layer counter was written and removed to stay at the 400-line target; the reason names the archive, which is enough.

### DIFF
--- a/tests/os/verify-image.sh
+++ b/tests/os/verify-image.sh
@@ -293,10 +293,10 @@ if [ -f ./pithead ] && [ -f dashboard/mining_dashboard/wizard.py ]; then
                     WIZ_SHIPPED="$WIZ_TMP/shipped-wizard.py"
                     break
                 fi
-                WIZ_WHY="tar -xOf $member from $(basename "$layer" | cut -c1-12) failed: $(head -c 120 "$WIZ_TMP/extract.err")"
+                WIZ_WHY="tar -xOf $member from $(basename "$layer" | cut -c1-12) failed: $(head -c 120 "$WIZ_TMP/extract.err" | tr -c '[:print:]' '?')"
             done
         else
-            WIZ_WHY="tar -xzf $(basename "$WIZ_ARCHIVE") failed: $(head -c 120 "$WIZ_TMP/untar.err")"
+            WIZ_WHY="tar -xzf $(basename "$WIZ_ARCHIVE") failed: $(head -c 120 "$WIZ_TMP/untar.err" | tr -c '[:print:]' '?')"
         fi
     fi
     chk "the baked wizard image contains the tree's wizard.py" \
--- a/tests/os/verify-image.sh
+++ b/tests/os/verify-image.sh
@@ -277,19 +277,35 @@ if [ -f ./pithead ] && [ -f dashboard/mining_dashboard/wizard.py ]; then
     WIZ_ARCHIVE=$(ls "$ROOT"/opt/pithead/images/*.tar.gz 2>/dev/null | head -1)
     WIZ_TMP=$(mktemp -d)
     WIZ_SHIPPED=""
-    if [ -n "$WIZ_ARCHIVE" ] && tar -xzf "$WIZ_ARCHIVE" -C "$WIZ_TMP" 2>/dev/null; then
-        for layer in "$WIZ_TMP"/blobs/sha256/* "$WIZ_TMP"/*/layer.tar; do
-            [ -f "$layer" ] || continue
-            member=$(tar -tf "$layer" 2>/dev/null | grep -m1 'mining_dashboard/wizard\.py$') || continue
-            tar -xOf "$layer" "$member" >"$WIZ_TMP/shipped-wizard.py" 2>/dev/null && {
-                # shellcheck disable=SC2034  # read inside chk's eval'd condition below
-                WIZ_SHIPPED="$WIZ_TMP/shipped-wizard.py"
-                break
-            }
-        done
+    # #1935: this check reddened once in a battery, re-passed on the same image, and the log could not
+    # say which step produced no file — so each step names itself, and a mismatch prints cmp's verdict.
+    WIZ_WHY="no *.tar.gz under opt/pithead/images"
+    if [ -n "$WIZ_ARCHIVE" ]; then
+        if tar -xzf "$WIZ_ARCHIVE" -C "$WIZ_TMP" 2>"$WIZ_TMP/untar.err"; then
+            WIZ_WHY="no layer lists mining_dashboard/wizard.py"
+            for layer in "$WIZ_TMP"/blobs/sha256/* "$WIZ_TMP"/*/layer.tar; do
+                [ -f "$layer" ] || continue
+                # sed, not grep -m1: an early exit would SIGPIPE tar and, under pipefail, skip the layer.
+                member=$(tar -tf "$layer" 2>/dev/null | grep 'mining_dashboard/wizard\.py$' | sed -n '1p')
+                [ -n "$member" ] || continue
+                if tar -xOf "$layer" "$member" >"$WIZ_TMP/shipped-wizard.py" 2>"$WIZ_TMP/extract.err"; then
+                    # shellcheck disable=SC2034  # read inside chk's eval'd condition below
+                    WIZ_SHIPPED="$WIZ_TMP/shipped-wizard.py"
+                    break
+                fi
+                WIZ_WHY="tar -xOf $member from $(basename "$layer" | cut -c1-12) failed: $(head -c 120 "$WIZ_TMP/extract.err")"
+            done
+        else
+            WIZ_WHY="tar -xzf $(basename "$WIZ_ARCHIVE") failed: $(head -c 120 "$WIZ_TMP/untar.err")"
+        fi
     fi
     chk "the baked wizard image contains the tree's wizard.py" \
         '[ -n "$WIZ_SHIPPED" ] && cmp -s "$WIZ_SHIPPED" dashboard/mining_dashboard/wizard.py'
+    if [ -z "$WIZ_SHIPPED" ]; then
+        echo "     · wizard.py was not extracted: $WIZ_WHY"
+    elif ! cmp -s "$WIZ_SHIPPED" dashboard/mining_dashboard/wizard.py; then
+        echo "     · shipped wizard.py differs from the tree's: $(cmp "$WIZ_SHIPPED" dashboard/mining_dashboard/wizard.py 2>&1 | head -1)"
+    fi
     rm -rf "$WIZ_TMP"
 else
     skip "the artifact matches the tree it was built from" "not run from the repo root"


VijitSingh97:
**Non-author review — I am NOT the author of this PR.** Ephemeral `reviewer` pass, head `b5befa79`, base `develop`.

## VERDICT: PASS at `b5befa79`

All four named claims re-derived at source. Two non-blocking findings below; neither changes the verdict. **Not mine to merge** — `tests/os/` is a freeze path, and the merging seat should read the "Merge" note at the end.

---

### 1. Every step that can leave `WIZ_SHIPPED` empty sets `WIZ_WHY` — MEASURED, not read

I extracted lines `277-301` verbatim and drove them under a `tar` shell stub over every path through the block. `verify-image.sh:13` is `set -uo pipefail` (no `-e`), so a failing assignment cannot abort — that was the one way this could have gone wrong and it does not.

| path | `WIZ_SHIPPED` | `WIZ_WHY` |
|---|---|---|
| no archive | empty | `no *.tar.gz under opt/pithead/images` (`:282`) |
| `tar -xzf` fails | empty | `tar -xzf <name> failed: <stderr>` (`:299`) |
| extracted, no layer files at all | empty | `no layer lists mining_dashboard/wizard.py` (`:285`) |
| layers, none listing wizard.py | empty | same (`:285`) |
| member found, `tar -xOf` fails | empty | `tar -xOf <member> from <digest> failed: <stderr>` (`:296`) |
| success | set | (unread) |

**Control that the instrument can say otherwise:** deleting `:282` and re-running the no-archive case prints `WHY=[<<UNSET>>]` and my checker fires. I printed the one-line diff proving the mutation landed before reading either result. The block is not inside a loop (`:268` is a plain `if`), so `WIZ_WHY` cannot be stale from a previous pass.

### 2. The assertion is unchanged — byte-identical

`:302-303` at head is byte-for-byte `:291-292` at base `ea1ab9a5`: same label, same `[ -n "$WIZ_SHIPPED" ] && cmp -s ...`. The two new `echo`s at `:304-308` are guarded so that a green run prints nothing extra.

### 3. `grep | sed -n '1p'` selects the same first match — and the SIGPIPE hazard is real, not theoretical

Same first match on a multi-match stream, measured. On the hazard I can say more than the PR body does. With a producer larger than the pipe buffer:

- `x=$(producer | grep -m1 PAT)` under `pipefail` → **rc 141**
- `x=$(producer | grep PAT | sed -n '1p')` → **rc 0**, same value

and the old code's `|| continue` fired on that 141, so the layer was **skipped after a successful match**. The PR body's "0 of 40 runs on the real archive" is consistent with this — whether it fires depends on how far into a layer's listing `wizard.py` sits — but "a hazard removed in passing" understates the mechanism. This is a real latent skip, and removing it is the strongest part of the change.

The new `[ -n "$member" ] || continue` is also strictly better than the old `|| continue` on the pipeline rc: a `tar -tf` that exits non-zero *after* emitting the match no longer discards it.

### 4. 400 lines, no budget row — correct

`awk 'END{print NR}'` (not `wc -l`) gives **400** at head, 384 at base. `scripts/lint-file-budget.sh:26` sets `TARGET_LINES=400` and its rule (`:5-6`) is *over* the target, so 400 needs no row; `docs/dev/file-budget.tsv` has no `verify-image.sh` row at base or head. Correct — and worth saying out loud that it is a **cliff**: the next line added to this file forces a row.

---

## Findings — non-blocking

**F1 — a multi-line `tar` stderr breaks the single-line reason format, and the fix is line-neutral.** `:296` and `:299` interpolate `head -c 120 <stderr>` with no newline scrub. GNU tar on a corrupt gzip emits 2-3 lines; driven with that stderr, the red prints as:

```
     · wizard.py was not extracted: tar -xzf <name> failed: tar: Skipping to next header
tar: Exiting with failure status
gzip: stdin: invalid compressed data
```

Three lines, two of them unindented and unlabelled in a battery log. **The sibling PR #1940 solves exactly this** with `| tr -c '[:print:]' '?'`, from the same session. Adding that filter to both interpolations is line-neutral, so the 400-line cliff does not block it. Not a RETURN — the diagnostic content is all present and the step is still named — but it is the one place where this PR's own product is degraded.

**F2 — two distinct failures share one reason.** An archive that extracts but contains no `blobs/sha256/*` and no `*/layer.tar` at all reports `no layer lists mining_dashboard/wizard.py`. Vacuously true, but it points a reader at the layer contents when the real fact is that there are no layers. One extra assignment inside the `[ -f "$layer" ] || continue` region would split them; only worth it if you are touching the file for F1 anyway.

## Over-engineering pass — nothing to cut

I agree with the author's own three points. One observation, not a criticism: `cmp` runs twice on the **green** path (once inside `chk`'s `eval`, once at `:306`), because `chk` does not return its result. On a wizard.py-sized file that is free, and capturing the result would mean changing a shared helper for one caller. The shape chosen is the minimal one.

## Merge

**Not mine.** `tests/os/` is a freeze path; the appliance lane sequences it. `closingIssuesReferences` for this PR is `[#1935]` and the base is the default branch, so **`Closes #1935` WILL fire** on merge — measured, not read off the body.

## What I did NOT do

No KVM, no image build, no run of `verify-image.sh` against a real image, no `make`, no shellcheck, no stack suite — the box is under the appliance lane's KVM leg and I held to the CI-only rule. Everything above is source reading plus pure-shell drives of extracted code with stubbed `tar`. CI at this head: **19 success / 2 skipped / 0 failing**; the two skips are the schedule-gated CVE sweep, which is not a measurement of this change either way.

VijitSingh97:
**Non-author review** (ephemeral `reviewer` pass; I did not write this branch). **RE-PASS at `907ccae3`.**

## The delta since `b5befa79` is exactly the two lines F1 named

One commit, one file, `2+/2-`:

```
907ccae3 test(os): #1935 review follow-up — a multi-line tar stderr stays on the reason's one line
 tests/os/verify-image.sh | 4 ++--
```

Both hunks are the stderr interpolations and nothing else — `verify-image.sh:296` (the per-layer `tar -xOf` failure) and `:299` (the `tar -xzf` failure), each gaining `| tr -c '[:print:]' '?'`. The file is **400 lines** at head, counted with `awk 'END{print NR+0}'`. `bash -n` clean.

## The scrub, driven — controlled pair, one variable

I seeded a stderr file with three lines, a NUL, a raw `ESC` and a CSI sequence, then interpolated it both ways:

| | lines | non-printable bytes surviving |
|---|---|---|
| old form (`head -c 120 f`) | **3** | **1** |
| new form (`head -c 120 f \| tr -c '[:print:]' '?'`) | **1** | **0** |

So the control fires, and the reason stays on one line. Worth naming beyond F1's own framing: the surviving byte in the old form was the `ESC` of a CSI sequence, i.e. the old interpolation could inject terminal control codes into a battery log that a person then reads in a terminal. The `tr` closes that too.

- `set -uo pipefail` at `:13` — **no `errexit`**, so the added pipeline cannot abort the script when the `.err` file is absent. It isn't absent anyway: both files are created by the `2>` redirect on the command immediately above each interpolation. No new failure mode from making these pipelines.
- `head -c` is byte-counted, so a 120-byte cut can land mid-UTF-8 — the `tr` turns the orphaned bytes into `?` rather than emitting an invalid sequence. Strictly better than before.
- `tests/os/verify-image.sh` has **no row in `docs/dev/file-budget.tsv`**, so the body's "400-line target" is a self-imposed target, not a gated ceiling. Correct as worded; nothing to pay.
- This file has no `--self-test` of its own (the only `self-test` string at `:34` is prose about a different file), so the pair above is the whole executable check available at this tier.

## Residual, non-blocking — I would not hold the PR for it

`$member` is still interpolated unscrubbed into the same reason at `:296`. It cannot reproduce F1's failure: it comes off a `grep` line, so it can never contain a newline. A control byte inside a tar member name would still reach the log. One `| tr` away if anyone wants symmetry; the failure F1 was about is closed.

I agree with all three items and would not change any. The one I checked rather than took on trust is "one variable rewritten per step": the read at `:305` prints `WIZ_WHY` only on `[ -z "$WIZ_SHIPPED" ]`, so the last-writer-wins semantics is the reason for the step that actually failed, not a stale earlier one. The "removed layer counter" item is honest about a cut and names what replaced it.

**Verdict: PASS at `907ccae3`.** Not mine to merge — `tests/os/` is a freeze path and the base is `develop`, so this is MERGE-READY for the seat's admin merge, not for me.

PR #1941, opened 2026-09-06
fix(pithead): #1930 the backup header names the dashboard DB among what it always keeps

### LINKED ISSUE #1930: 17-backup-restore.sh says the dashboard DB is excluded by default; its own code always includes it
### What is wrong

`17-backup-restore.sh` contradicts itself about the dashboard database, in the same file, about the
same default.

**The header, lines 2-4:**

> Protect the irreplaceable bits: config.json, .env (secrets), the Caddyfile, and the Tor data dir
> (onion service keys). The blockchains **and dashboard DB** re-sync, so they're **excluded by
> default**; pass `--with-chains` to also fold in the Monero/Tari/P2Pool data dirs.

**The code, lines 126-130:**

> The dashboard's data dir holds its database (hashrate history & settings). It's small and
> irreplaceable (it does **NOT** re-sync), so it **always goes in the default backup**, not behind
> `--with-chains`.

```sh
[ -d "$DASHBOARD_DIR" ] && items+=("$DASHBOARD_DIR")
```

The include is unconditional — outside the `if [ "$with_chains" -eq 1 ]` branch below it.

### Which one is right

**The code.** The dashboard DB is in the default backup, and the header is stale: it makes two
claims about it that the file itself refutes, that it re-syncs and that it is excluded by default.

### Why it matters more than a stale comment usually does

This is the file an operator or a lane reads to answer "is my hashrate history in the backup?", and
the wrong answer is the reassuring-sounding one in the **header**, where someone skimming stops. It
also states the defect as a rationale — "the blockchains and dashboard DB re-sync" — which is the
shape that survives being read, because it explains itself.

### Fix

Correct the header to name the dashboard data dir among the irreplaceable bits it already protects,
and leave `--with-chains` describing only the chain data dirs.

Inherited, not introduced by any current PR. Surfaced by the review pass on PR #1921 and confirmed
at source here before filing. `lib/pithead/` is not this lane's path, so this is filed rather than
fixed.

VijitSingh97:
Closes #1930.

## What changes

`lib/pithead/17-backup-restore.sh`, header lines 2-4: the dashboard data dir is named among the irreplaceable bits the default backup keeps, with the reason the code already gives (its database, hashrate history and settings, never re-syncs). `--with-chains` is described as folding in the chain data dirs only. The rewrite is three lines for three, so the concatenated `pithead` moves by the same three lines and keeps its budget row. `pithead` is rebuilt from the slices.

## Why

The header said the dashboard DB re-syncs and is excluded by default; `stack_backup` has always included it unconditionally, outside the `--with-chains` branch, and its own comment at the include says why. The wrong claim sat where a skimming reader stops, and stated itself as a rationale. The docs already say the true thing (`docs/appliance.md`, `docs/dev/appliance-wizard.md`, the dual-distribution plan), so only the header moves.

## What was RUN

- `scripts/build-pithead.sh` then `scripts/build-pithead.sh --check`: parity OK, 54 slices.
- `git diff --numstat`: 3/3 on the slice, 3/3 on `pithead`. `scripts/lint-file-budget.sh`: OK.
- `bash -n` on both files.

## What was NOT done

- No shellcheck run: the diff is comment-only in a slice `make lint-sh` reads through `pithead`, and CI's Shell tests job runs it. No suite: no code changed.
- No image: `lib/pithead/` is an image-freeze path. This is a comment; it changes no byte of behaviour, but it does move the tree the next RC builds from. RC2 material, not RC1.

- One clause in the header rather than a paragraph: the code's own comment at the include already carries the full reason, and the header's job is the list.
- Not taken: moving the reason out of the code comment into the header. Two sites, each true, is the state the file was meant to have; the defect was that one of them lied.

### DIFF
--- a/lib/pithead/17-backup-restore.sh
+++ b/lib/pithead/17-backup-restore.sh
@@ -1,7 +1,7 @@
 # --- Backup / Restore ---
-# Protect the irreplaceable bits: config.json, .env (secrets), the Caddyfile, and the Tor
-# data dir (onion service keys). The blockchains and dashboard DB re-sync, so they're excluded
-# by default; pass --with-chains to also fold in the Monero/Tari/P2Pool data dirs.
+# Protect the irreplaceable bits: config.json, .env (secrets), the Caddyfile, the Tor data dir
+# (onion service keys), and the dashboard data dir (its DB never re-syncs). The blockchains do
+# re-sync, so they're excluded by default; pass --with-chains to fold in Monero/Tari/P2Pool data.
 #
 # The two files a backup archive cannot be worth taking without (#1059/#1244): config.json and
 # .env carry the node credentials and secrets, unlike every optional item stack_backup adds
--- a/pithead
+++ b/pithead
@@ -4198,9 +4198,9 @@ factory_reset() {
 }
 
 # --- Backup / Restore ---
-# Protect the irreplaceable bits: config.json, .env (secrets), the Caddyfile, and the Tor
-# data dir (onion service keys). The blockchains and dashboard DB re-sync, so they're excluded
-# by default; pass --with-chains to also fold in the Monero/Tari/P2Pool data dirs.
+# Protect the irreplaceable bits: config.json, .env (secrets), the Caddyfile, the Tor data dir
+# (onion service keys), and the dashboard data dir (its DB never re-syncs). The blockchains do
+# re-sync, so they're excluded by default; pass --with-chains to fold in Monero/Tari/P2Pool data.
 #
 # The two files a backup archive cannot be worth taking without (#1059/#1244): config.json and
 # .env carry the node credentials and secrets, unlike every optional item stack_backup adds


VijitSingh97:
**Non-author review** (ephemeral `reviewer` pass; I did not write this branch). **PASS at `d804eb4a`.**

## The three things the ask named, each measured

**1. `git diff --numstat` is `3/3` on both files, and there is nothing else in the branch.**

```
3	3	lib/pithead/17-backup-restore.sh
3	3	pithead
```

The `pithead` hunk is the concatenated copy of the slice hunk, character for character.

**2. `scripts/build-pithead.sh --check` passes at head — with a control that fires.**

```
pithead parity OK — the committed artifact is exactly what 54 slice(s) build.   rc=0
```

A clean `--check` on its own proves nothing about whether the check can see this file, so: I mutated the slice's first line, printed the diff proving the mutation landed, re-ran → **rc 1**, restored, `git status --porcelain` empty. The gate can say the other thing about *this* slice.

**3. The header now agrees with the code at the include.**

`17-backup-restore.sh:130` is

```bash
[ -d "$DASHBOARD_DIR" ] && items+=("$DASHBOARD_DIR")
```

— outside the `if [ "$with_chains" -eq 1 ]` block that opens at `:132`, so the dashboard data dir goes into **every** backup, not only `--with-chains` ones. The old header said "The blockchains and dashboard DB re-sync, so they're excluded by default", which was false about half its own subject. The new one is true, and I checked the two neighbours it has to stay consistent with: the inline comment at `:127`-`:129` already said the same thing (the file was contradicting itself, which is the actual defect), and the runtime line at `:261` says only "Blockchains excluded (they re-sync)" — no dashboard claim to fall out of step.

## The word "unconditional" is worth one line of precision

`:130` is unconditional **with respect to `--with-chains`**, not unguarded — it still carries `[ -d ]`. The header does not overclaim: its second paragraph at `:8` already states that "the Caddyfile, the Tor data dir, the dashboard dir all get a plain `[ -f ]`/`[ -d ]` guard already", against the two required items that do not. So the list-plus-guard-note shape is accurate as written, and I would not reword it.

- `--with-chains` still adds only `MONERO_DIR`/`TARI_DIR`/`P2POOL_DIR` (`:132`-`:138`). The header's "pass `--with-chains` to fold in Monero/Tari/P2Pool data" enumerates exactly those three.
- The comment reflow is 3 lines in and 3 out, so nothing was dropped to make it fit: the old text's two clauses ("blockchains re-sync", "`--with-chains` adds the three data dirs") both survive, and the only content change is the dashboard clause moving sides. Reflows in this repo have quietly eaten consequence clauses before; this one did not.
- No budget row for either file is touched, and neither file changes length.

I agree with both items, and the second one is the interesting call. "Not taken: moving the reason out of the code comment into the header" is right for a reason the body understates — the header is read by someone deciding *what a backup contains*, the include comment by someone editing *how it is assembled*, and collapsing them would leave the second reader with no local justification for why one item skips the `--with-chains` gate. Two sites, two audiences, both true is the correct end state; the defect was one of them lying, exactly as the body says.

**Verdict: PASS at `d804eb4a`.** Not mine to merge: `lib/pithead/` and `pithead` are image-freeze paths, and the base is `develop`. MERGE-READY for the seat's admin merge, sequenced by the appliance lane against the RC window.

Issue #1856, opened 2026-09-05
Stack topology: LAN ingress hops draw as Local, and the onion ingress and Tor auto-heal probe are missing — every edge classified against config


VijitSingh97:
## Why this issue

From the 2026-09-05 manual test: "the networking looks off with the lan/clearnet/tor also for other connections, not just Tari." #1849 covers the remote-node rows. This is the rest: every edge the Stack Topology diagram draws, classified against what the running configuration says, with a verdict per edge. Read at `develop-v2` `7e65e451`; rendered in the visual harness (Advanced view, `#card-egress`, all fixtures).

## Every edge, and whether it tells the truth

The route colours come from `service/egress.py` `compute_topology` (`:322-391`) and the legend from `web/static/topology.mjs:31-48`. The legend defines **Local** as "this machine" and **LAN** as "a hop that leaves it but stays on your network" (`topology.mjs:44-46`, the #1350 split).

| # | edge (label) | drawn as | derived from | verdict |
|---|---|---|---|---|
| 1 | Mining rigs → xmrig-proxy (stratum) | grey **Local**, `egress.py:323` | constant `LOCAL` | **Wrong colour.** The rigs node sits in the LAN zone (`topology_graph.py:39`); a rig on another host is a LAN hop by the legend's own definition. Only an on-box `local_miner` is Local. Draw it blue (LAN); if `local_miner` is on, add it as an internal edge from a host-zone node. |
| 2 | Browser → caddy (https :443) | grey **Local**, `:324` | constant `LOCAL` | **Wrong colour**, same reason: the browser is in the LAN zone. Blue. |
| 2b | (missing) Browser → tor → caddy over the dashboard onion (#343) | not drawn | the dashboard has no knob for its own onion (`config/config.py` carries none) | **Gap.** When the onion is on, the only ingress that crosses the trust boundary is invisible. Depends on #1853 giving the dashboard that fact; then draw a green ingress through the tor hub. |
| 3 | p2pool ↔ tor / internet (sidechain P2P) | green, or red to the internet node | `p2pool.clearnet` (`:321`, `_ext(sidechain)`) | Correct. |
| 4 | monerod ↔ tor (Monero P2P + tx) | green, unconditional (`:328`) | nothing | Correct only for a local node; wrong for `monero.mode: remote` (#1849 covers the mechanism; same fix). |
[...]

Issue #1849, opened 2026-09-05
Network diagram is confusing or inaccurate when Tari is a remote node (tari.mode=remote)


VijitSingh97:
## What the operator saw (manual test, 2026-09-05, pithead.local — Pithead + RigForge, Tari configured as a REMOTE node)

The dashboard's network diagram is confusing or inaccurate for a machine whose Tari node runs elsewhere (`tari.mode=remote`). Reported from the appliance's running dashboard, which at this VERSION is the published `v1.20.0` image; first check whether `develop-v2` (`topology.mjs`, `minecart.mjs`) already renders remote Tari differently, and say so if it does.

## What to establish (design lane, with the visual harness)

- Render the topology for `tari.mode=remote` (and for local, and for Tari disabled) against canned `build_state()` payloads and attach the screenshots.
- State what the diagram claims in each case (which boxes, which arrows, which labels) against what is true: a remote Tari base node is off-box, reached over gRPC by the merge-mining leg of p2pool; the local machine runs no Tari container in that mode; the wallet-side (view-only console wallet) may still be local.
- Name the specific inaccuracy (a box that should not be there, an arrow pointing the wrong way, a "syncing" state for a node this machine does not run, a label that implies a local node) with file:line, and the fix.

## Why it matters for GA

The diagram is the first thing a new operator reads to understand what their machine is doing; a wrong picture for a supported mode (remote Tari, #103) is a correctness defect of the UI, not polish. Part of the dashboard review the `design` lane is running; fix lands after the review names it, in the dashboard.

VijitSingh97:
Operator, 2026-09-05, broadening this: "the networking looks off with the lan/clearnet/tor also for other connections, not just Tari". The review covers every edge the diagram draws — monerod, p2pool, XvB, the workers, the dashboard onion — and classifies each by what the running configuration actually says, not by a fixed drawing. The Tari edge proven here (`egress.py` draws tari→tor unconditionally) is one instance of that pattern; look for the others before filing the fix so they land as one change.

VijitSingh97:
## Fill-in from the design lane (harness render of `develop-v2`, read at `7e65e451`)

[...]

Issue #424, opened 2026-07-10
Tor stuck on a failing guard silently breaks all clearnet egress (Healthchecks, Telegram, XvB) — add detection + self-heal


VijitSingh97:
## What happened

On pithead-coordinator, after the tor container restarted during the v1.3.0 deploy, Tor bootstrapped to 100% but got **stuck on a failing guard** — its own log:

```
Guard ajtracey is failing more circuits than usual. Most likely this means the Tor network is
overloaded. ... 197 circuits completed, 0 were unusable, 43 collapsed, and 187 timed out. For
reference, your timeout cutoff is 60 seconds.
Failed to find node for hop #1 of our path. Discarding this circuit.
```

New circuit builds timed out at the 60s cutoff, so **every clearnet-exit-over-Tor feature broke at once** — Healthchecks pings, the Telegram bot (alerts + commands), and XvB stats — while the stack otherwise looked healthy (mining rides onion/established circuits, dashboard is local, onion is inbound). It stayed broken for ~6 hours until a manual `docker compose restart tor` reselected guards and exits recovered (204 in ~1s).

The operator noticed via "Healthchecks says it's not receiving pings" and "Telegram commands don't work" — the dead-man's switch did its job, but nothing in the stack surfaced the actual cause, and it didn't self-heal.

## Why it matters

Healthchecks is the *dead-man's switch*. If the very path it pings over (Tor clearnet exit) is the thing that's broken, the operator gets a "host is down" alert for a host that's mining fine — misleading, and the real problem (Tor egress) is invisible in `status`/`doctor`.

## Proposal

Two parts, both cheap:

1. **doctor check — Tor clearnet egress works.** Add a `doctor` runtime check (alongside the #383 checks) that makes one SOCKS request through the tor container to a robust clearnet endpoint (e.g. `https://www.google.com/generate_204`) with a ~15s timeout, when the tor container is running. OK on 2xx/204; WARN with the specific "Tor can reach the network but clearnet exits are failing (Healthchecks/Telegram/XvB will be down) — try 'pithead restart tor'" message on timeout. Distinguishes "Tor down" from "Tor up but exits failing".

[...]

PR #1953, opened 2026-09-06
fix(dashboard): #1856 the two LAN ingress hops stop drawing as "this machine"

VijitSingh97:
## What was wrong

The Stack Topology diagram drew the two ingress hops as **Local**, which its own legend defines as "this machine" (`web/static/topology.mjs:46-47`). Both endpoints sit in the LAN zone (`service/topology_graph.py:39-40`), so both hops leave the box. The comment directly above those two edges already read "Ingress from your LAN" — only the code disagreed.

This is the security panel. Reading a hop that crosses onto your network as "this machine" overstates the trust boundary, which is the one thing this card exists to state correctly.

Second half, the same issue's row 13b: the #424 Tor egress probe was in **neither** privacy view and in no doc. `tor_heal` sends `PROBE_URL` through `TOR_SOCKS_PROXY` every five minutes while `tor.auto_heal` is on (`service/tor_heal.py:63`, `:130-134`). An operator auditing "All connections" was reading a list that omitted a connection.

## What changed

- `egress.py:323-324` — the rigs and browser ingress edges route `LAN`, not `LOCAL`.
- A "Tor egress probe" row in the egress list and a matching edge in the topology, both keyed on `tor.auto_heal`, both `INACTIVE` until it is on. Unconditional rows, like the four Tor-only dashboard clients beside them, so the existing exhaustive sweeps see them.
- `docs/privacy.md` — the probe gets a row in the runtime-egress table (it was absent entirely), and the topology section explains why the two ingress hops read LAN.

**No leak count, headline, legend entry or node changes.** A LAN hop is not egress, so `leaks` is untouched — asserted, not assumed.

## Budget, and where code went

`egress.py` was at **423/424** and this needs nine lines. `check_monotonic` refuses a ceiling raise, so `edge`/`ext_node` move to `topology_graph.py`, whose docstring already records "egress has no room" as why the route vocabulary lives there. They are re-exported; no importer changes. Final: **423/424**.

That rename collided with `for edge in edges:` in the firewall-tagging pass, shadowing the import — the suite caught it, not review. Loop variable renamed to `link`.

`_notify_knobs` → `_shared_knobs` (it now carries the probe knob both from-config builders read) — this is the line that paid for a second config read at two call sites.

Declined on merits: folding the five `TOR if <flag> else INACTIVE` conn/edge pairs into one table. It is real duplication, but it is a refactor of the whole panel during an RC cut, and it is not this bug.

## Tier and what was run

Unit. `tests/service/test_tor_probe_egress.py` is a **new module** because `test_egress.py` is at **451/451**; the ingress test there is corrected in place, line-neutral, and now checks both edges' route where it checked only the rigs one. The same ceiling kept `tor_auto_heal` out of the `_KNOBS` product, which costs little: the row and edge are unconditional, so the sweeps already cover the off branch, well-formedness and summary agreement. The new module pins the on branch.

Run by me at this head:

- `pytest` (full dashboard suite): **2575 passed**, rc 0
- `node --test dashboard/tests/frontend/`: **579 pass / 0 fail**, rc 0
- `make test-fakes`: **31 passed**, rc 0 — run because this changes the payload
- `make lint-py lint-js lint-docs-voice lint-operator-strings lint-topology lint-file-budget lint-trivy-parity lint-yaml lint-md lint-proto lint-toml`: rc 0. The budget gate self-tested and then reported every file within its ceiling.

Not run: `make lint-sh` (no shell files in this diff, and the appliance lane holds the box), and no docker, KVM or bench work.

## The frontend fixture, which is a wider change than it looks

`tests/frontend/fixtures/state.json` is regenerated, and the diff carries `tari` values and an indentation fix my change does not explain. **It was already stale on `develop` and had been hand-edited** — it held indentation `json.dump` cannot emit, and `tari.local`/`remote` values the generator does not produce.

Proven independently of this diff: `tari_is_local()` returns `False` at defaults — the default `TARI_GRPC_ADDRESS` is a loopback literal, which does not carry the local-monerod subnet prefix that check compares against (`config/config.py:71-75`) — and this branch touches no config file — so any faithful regeneration on `develop` yields the same values. Its drift guard (`tests/web/test_xvb_views.py:503`) compares dotted key paths only, **never values**, so value drift there is invisible to CI. Regenerating is the fixture's own documented contract.

## What the operator sees on the appliance

In Advanced view, **Stack Topology & Egress**: the two arrows entering `xmrig-proxy` and `caddy` turn blue (LAN) instead of grey (Local). "All connections" gains a **Tor egress probe** row, showing Inactive unless `tor.auto_heal: true`. Nothing else moves — same headline, same badge, same node boxes.

## What I did NOT do

**#1856 stays open.** Row 1's second half — the on-box `local_miner` internal edge — is not here. It needs a new node id, which touches `TOPOLOGY_NODES`, the frontend `POS` map, and two canonical-id lists that assert an exact set, one of them in `test_egress.py` at its ceiling. That is a layout decision plus a contract change in four places, not a colour fix, and it is a gap rather than a regression: the "Mining rigs" node is an external actor, so drawing its hop as LAN says nothing false about an on-box miner, which is undrawn today and stays undrawn.

Also untouched, per the issue's own split: rows 4/5/15 belong to #1849, row 2b to #1853. The issue's "Taste" suggestion — a sub-line counting LAN and unverified hops under the "All egress via Tor" headline — is not done; it is a new claim on the panel, not a correction.

`docs/` is shared (`_shared.paths`), so no lane-override was needed; disclosing the edit here as the rule asks.

## Merge

Not mine to merge. Needs a recorded non-author PASS at this head plus green CI, and the seat's `--admin` on `develop`.

### DIFF
--- a/dashboard/mining_dashboard/service/egress.py
+++ b/dashboard/mining_dashboard/service/egress.py
@@ -30,9 +30,12 @@ from mining_dashboard.config import config
 from mining_dashboard.service.topology_graph import (  # noqa: F401  (re-exported)
     CLEARNET,
     INACTIVE,
+    LAN,
     LOCAL,
     TOPOLOGY_NODES,
     TOR,
+    edge,
+    ext_node,
     node_route,
     topology_nodes,
 )
@@ -108,6 +111,7 @@ def compute_egress_posture(
     notify_tor=True,
     notify_sinks_private=False,
     xvb_standby_source="",
+    tor_auto_heal=False,
 ):
     """Pure derivation of the egress posture from config knobs. Returns ``{components, summary}``."""
     xvb = _xvb_route(xvb_enabled, xvb_tor)
@@ -185,6 +189,8 @@ def compute_egress_posture(
                 # any non-private source rides Tor (like every read above); only a private-IP-literal
                 # primary is a LAN hop (local). Never clearnet, so it can't leak the backup's IP.
                 {"to": "XvB standby pull (backup ← primary)", "route": standby},
+                # Tor egress probe (#424) — opt-in via tor.auto_heal, socks5h, so never a leak.
+                {"to": "Tor egress probe", "route": TOR if tor_auto_heal else INACTIVE},
             ],
         },
         {
@@ -240,17 +246,18 @@ def egress_posture_from_config():
         telegram_enabled=config.TELEGRAM_ENABLED,
         price_feed_enabled=config.DASHBOARD_ENERGY["price_feed"],
         xvb_standby_source=config.XVB_STANDBY_SOURCE,
-        **_notify_knobs(),
+        **_shared_knobs(),
     )
 
 
-def _notify_knobs():
-    """The #380 alert-sink knobs, shared by both from-config builders."""
+def _shared_knobs():
+    """The knobs both from-config builders read: the #380 alert sinks and the #424 Tor probe."""
     urls = [*config.NOTIFY_WEBHOOK_URLS, config.NTFY_URL]
     return {
         "notify_sinks_enabled": any(u.strip() for u in urls if u),
         "notify_tor": config.NOTIFY_TOR,
         "notify_sinks_private": _sinks_all_private(urls),
+        "tor_auto_heal": config.TOR_AUTO_HEAL,
     }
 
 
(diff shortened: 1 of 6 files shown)


VijitSingh97:
Independent security review: RETURN at `166cf46d2a717c6c33876f874fc03a4d7b7fa035`.

The graph always labels ingress as LAN and the new privacy documentation describes these as the only exposed listeners. The default Stratum listener binds all interfaces, and existing optional RPC/ZMQ/gRPC exposure is omitted. Use neutral wording or derive exposure from the actual bindings, so the view does not imply LAN confinement or a complete listener inventory.

Static Sol/high security review; no new network behavior was introduced and the existing Tor probe remains opt-in with no clearnet fallback. This finding concerns the accuracy of the new security explanation.

VijitSingh97:
Superseded by reviewed integration PR #1972, merged as `43cd9230a125ba9bd0a446b3fb192ed97f6ed8cb`. The ingress-label repair supersedes this original candidate. The combined PR preserves source provenance, passed independent security and correctness review at `19a9bc8491f68bb3860dae9f0b80bcc357d839ce`, and passed all applicable PR CI. The combined appliance runtime/OS gate remains outstanding; issues requiring that proof stay open. Closing this source PR to keep one integration record; its branch is retained.

PR #1968, opened 2026-09-07
ci: run generated CLI parity in the existing lint job

### LINKED ISSUE #1925: lint-pithead-parity runs in no CI workflow, so the shipped artifact is unguarded

VijitSingh97:
The generated CLI parity gate now runs in the existing lint job, so stale generated output and the target’s failure controls are checked in CI. The wizard probe test also accepts the styled submit button introduced by #1893 while still requiring that the button exists and is enabled; its previous literal markup assertion was failing unrelated PRs.

Validation at `937cfae237d4141649f3ae988effef5ed848b7e2`: frontend 636 passed, 0 failed; actionlint; generated CLI parity including negative controls; diff check. Independent Sol/high security review and Luna/high verification are recorded separately. No new CI job or test framework.

Closes #1925.


VijitSingh97:
Independent review: PASS at `937cfae237d4141649f3ae988effef5ed848b7e2`.

Sol/high security reviewer and Luna/high verifier both passed the exact combined head: existing CI parity target wiring, unsupported timing-comment removal and the styled-submit test correction. Author-run frontend suite: 636 passed, 0 failed. Reviewer independently reran the focused wizard domain (23 passed), actionlint, parity and diff checks. Remote required CI is pending; no browser click/CSS-overlay or OS runtime result is claimed.

PR #1980, opened 2026-09-07
test(integration): keep probe credentials out of process arguments

### LINKED ISSUE #1979: Keep live integration credentials out of process arguments
The live integration matrix passes RPC, dashboard and worker-control credentials through process arguments in several probes. Shell quoting preserves values but does not remove them from `bash -c`, SSH or curl argument lists.

Affected paths are `monero_caught_up` in `tests/integration/lib.sh`, the restore RPC check in `restore-proof.sh`, and dashboard metrics plus worker-control apply/status probes in `run.sh`.

Feed curl authentication configuration through stdin, using the existing transport pattern, and keep secret values out of the command snippets. Preserve authenticated success, unauthenticated probes, request bodies and failure verdicts. Add a focused check that runs the shipped probes through local and SSH transports, observes shell/SSH/curl arguments, and proves the expected credentials reach curl stdin. No live credential values are needed in the regression test.

VijitSingh97:
## Summary

- keep Monero, dashboard, and RigForge integration credentials out of shell, SSH, and curl process arguments by passing curl configuration on standard input
- preserve standard input only for credential-bearing remote probes while ordinary remote commands retain SSH input isolation
- cover local and SSH transports with a semantics-matching fake plus positive and negative standard-input controls

Closes #1979.

## Checks

- credential transport self-test: 10 local/SSH probes and 2 standard-input controls passed
- focused Monero matrix self-test: 27 passed
- focused restore self-test: 26 passed
- Linux integration phase self-test: 80 passed, 0 failed
- changed-file syntax, formatting, ShellCheck, file-budget, and diff-hygiene checks passed
- independent Sol/high security and Luna/high verification passed at `b940b4ff6d0b24b83cc2ccac7f6f7417a947fc42`
- Ponytail complexity review: Lean already. Ship.

## Limits

No live credentials, SSH target, service, Docker workload, hardware, image build, or real integration matrix was used. The regression test uses isolated fake executables. Current-head CI remains required.

### DIFF
--- a/tests/integration/lib.sh
+++ b/tests/integration/lib.sh
@@ -382,7 +382,7 @@ monero_caught_up() { # 0 caught up / 1 answered, behind / ANY other could-not-as
     rx 'u=$(grep -E "^MONERO_NODE_USERNAME=" .env 2>/dev/null | cut -d= -f2-);
         p=$(grep -E "^MONERO_NODE_PASSWORD=" .env 2>/dev/null | cut -d= -f2-);
         url=$(grep -E "^MONERO_RPC_URL=" .env 2>/dev/null | cut -d= -f2-); [ -n "$url" ] || url=$(jq -r "if (.monero.mode // \"local\") == \"remote\" and .monero.remote.host then \"http://\" + .monero.remote.host + \":\" + ((.monero.remote.rpc_port // 18081) | tostring) else \"http://127.0.0.1:18081\" end" config.json 2>/dev/null); [ -n "$url" ] || url="http://127.0.0.1:18081";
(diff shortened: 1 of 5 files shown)


VijitSingh97:
Non-author merge review: **PASS** at exact head `b940b4ff6d0b24b83cc2ccac7f6f7417a947fc42`.

I read the complete credential-transport review chain, including the original SSH-stdin RETURN, the corrected Sol/high and Luna/high PASS reports, and both exact-head metadata PASS reports. The final change keeps ordinary SSH reads isolated, opts in to stdin only for credential-bearing probes, and the isolated fake covers both transport modes without falling through to a real executable. All required CI checks are green.

Issue #1979 has no `do-not-close` label or newer acceptance comment, and this PR satisfies its current body, so `Closes #1979` may auto-close on merge. The live integration matrix and hardware remain separate release gates.

PR #1982, opened 2026-09-07
test(os): cover RC2 appliance regressions

VijitSingh97:
Closes #1966.

This extends the eight-phase appliance battery with focused coverage for host-approved configuration changes, node endpoint consumption, hostname persistence, diagnostics boundaries, failed provisioning recovery, and restore lock behavior. It also updates the release and testing guides so the full KVM battery remains an explicit per-RC gate.

The approval rows preserve unrelated control requests and restore the exact operator configuration across interruption and cleanup paths. Positive approval and protected credential-transport execution depend on the reviewed configuration companion in #1978. Runtime node checks also require reserved synchronized Monero and Tari endpoints.

Validation:
- Bash syntax, shfmt, and ShellCheck on the changed OS helpers
- focused helper self-tests, including cleanup interruption and ownership controls
- focused lifecycle mutation-lock subsection: 53 passed, 0 failed
- `make lint-file-budget lint-pithead-parity` (54 generated parity slices)
- merge-resolution checks: test inventory drift, file budget, parity, helper self-tests, and boot-label positive/negative controls
- exact combined-source mutation-lock control after the restore-fixture correction: 53 passed, 0 failed
- independent Sol/high security and Luna/high verification at the substantive head; root renewal PASS for the exact fixture head `aa7f264b`

The positive full KVM appliance battery has not run against a final image. This PR remains draft until that battery passes with #1978 integrated and the reserved node inputs available.

PR #1974, opened 2026-09-07
feat(os): show current and previous versions in the boot menu

VijitSingh97:
The boot menu now shows the version and A/B slot selected by RAUC as current, and labels the other populated slot previous. Legacy slots without metadata say version unknown; an intentionally empty slot says empty. Equal versions retain their A/B identities.

Fresh images and installs seed slot metadata, updates record the new primary, and boot repairs the running slot from its own immutable VERSION. Existing fallback and Set up again behavior is preserved. The OS update test checks only the final reboot’s serial output, with a regression that rejects an earlier matching menu.

Refs #1956. Runtime verification remains open until the combined image passes its static and KVM gates.

Validation: 17 boot-label fixtures, 71 installer, 84 appliance-identity and 83 OS-update assertions passed, along with syntax, shfmt, targeted ShellCheck, generated CLI parity, file-budget, docs and diff checks. Independent Sol/high security and Luna/high verification both passed at `64c3344509bf2218d710327b22f7db7f7d01f900`. No image build or KVM run is claimed by this PR.

Issue #1956, opened 2026-09-07
Boot menu: name the entries by the Pithead version each slot holds and mark the one that boots (current), not 'fallback / Slot A / Slot B'


VijitSingh97:
## What the operator saw (RC1, the #1844 boot menu)

Four entries: a fallback entry, Slot A, Slot B, and the "Set up again" wizard entry. The wizard entry reads fine. The rest does not say what a person is about to boot.

## The operator's proposal, and a recommendation

- The first entry is what the machine boots by default, so call it **(current)**, not "fallback".
- Name each slot by **the Pithead version it holds** rather than by a letter: `Pithead 2.0.0 (slot A, current)`, `Pithead 1.20.0 (slot B)`. An empty slot says `empty (slot B)`. Two slots with the same version show the same version twice — the letter after it is what tells them apart. Keep the letters, but second.

Recommendation: exactly that, plus one word of state from RAUC — `(current)` on the slot the machine boots, `(previous)` on the other when it holds a version, so a rollback reads as "boot the previous version" rather than "boot slot B". The wizard entry keeps its name. "Recovery", if it stays, needs to say what it recovers.

## How it can know

GRUB cannot read the slots' `VERSION` files, but it reads `grubenv`, which RAUC already uses for slot order and try counts. `pithead-boot` writes `A_VERSION` / `B_VERSION` (and clears the one it wiped) into `grubenv` when a slot is installed or committed; the menu script formats the entries from those variables at boot, with `empty` for an unset one. The image build stamps the initial values.

## Bar

Tier 1: the menu-rendering script against fixture grubenv values (both set, one empty, same version twice); `verify-image-boot-menu.sh` asserts the labels. Tier 3: the KVM update leg reads the serial console's menu after an update and after a rollback.

VijitSingh97:
## Measured on the KVM bench: both menu entries name the same slot

PR #1974 shipped the feature, and the battery assertion for it (added on #2002) now has a real run
behind it. It fails, and the reason is specific: the two entries are labelled `current` and
`previous` correctly, but **both name slot A**.

Serial console after the update installed to the spare slot and committed:

```
Pithead 2.0.0 (slot A, current)
Pithead 2.0.0 (slot A, previous)
```

The commit moved the booted system to slot B, so the `current` entry should read
`Pithead 2.0.0 (slot B, current)`. The verdict
(`tests/os/boot-label-serial-verdict.sh`) requires both halves:

```
[...]

PR #2002, opened 2026-09-08
test(os): combine #1931 #1945 #1966 RC2 KVM gate

### LINKED ISSUE #1931: debug variant: the registry pin reaches three boot units only, so `pithead` over SSH pulls from the org's registry and the stack cannot come back up

### LINKED ISSUE #1945: KVM restore leg: the harness's backup is granted the mutation lock while the firstboot's `pithead up` is still in its health wait; two compose ups overlap and tor ends unhealthy

### LINKED ISSUE #1966: Manual battery and the KVM suite: add backup/restore end to end and the RC1 gaps (post-provision apply, backup restart, diagnostics verbs, pre-fill migration, failed-install recovery)
## Operator (RC1 manual test)

"See if our manual testing document needs to cover more: it should include a full backup procedure, and the gaps I saw. See if we can improve CI or e2e testing to include these, and our testing suite against the appliance itself."

## Gaps the RC1 manual round found that no automated tier caught

| Gap | Issue | Where it should be caught |
|---|---|---|
| Dashboard apply on the appliance pulled from the public registry and then tried to BUILD from a build dir the appliance does not have | #1931 (+ #1737 for the build contexts) | KVM battery: a config change through the control runner after provisioning |
| Backup from the dashboard left the stack down | #1931 / backup issue | KVM battery restore leg: assert the dashboard answers again |
| Reinstall pre-fill carried a removed 1.x key | #1954 | KVM reinstall leg seeded with a 1.x-shaped previous config |
| Failed install = dead page, fields lost | #1955 | KVM provision leg with an unreachable remote node |
| Service diagnostics verbs time out | diagnostics issue | KVM battery: doctor + log-tail through the runner |
| xvb.enabled three ways | xvb default issue | tier 1 |
| Boot menu labels | #1956 | verify-image + the serial-console read in the update leg |

## What should happen

- `docs/dev/appliance-release.md` § Manual battery gains **M15 Backup and restore end to end** (create from the dashboard, download, note the time, restore in a fresh wizard, verify payouts and onion identity survived) and **M16 Change a setting from the dashboard after provisioning** (a benign one and an approval one), and a line per gap above that stays manual.
- Everything in the right-hand column that says KVM battery is added as a row in `tests/os/run.sh` (the ledger #1886 is where the accounting lives; this issue is the RC1 addendum).
- CI (`ci.yml`) cannot run the appliance; the KVM battery is the appliance's suite and runs on every RC (the release gate). State that in `docs/dev/testing-strategy.md` if it is not already plain.

### LINKED ISSUE #1994: Matrix E2E: RigForge apply drops borrowed-rig pool fixture before failover and subnet phases
## Problem

The live matrix temporarily points a borrowed rig at the test stack by editing the rendered XMRig configuration. Any later `rigforge apply` correctly regenerates that file from the rig's persistent operator configuration and restores its normal pool. This happens during a real bootstrap/upgrade and during the matrix's reversible control-write checks. The matrix then reaches its failover and moved-subnet phases with no worker attached to the test stack.

This produced 210 passes, 3 failures, and 5 missing legs in the upgrade validation run and reproduced on an already-current release after the control-write checks. The three failures were the node-down failover pre-arm check and the moved-subnet worker/hash checks. The same rig remained healthy and continued producing fresh accepted shares on its original pool, so this is test-fixture ordering rather than a RigForge runtime failure.

## Expected fix

- After any RigForge bootstrap, upgrade, or control apply, reapply the borrowed-rig pool fixture under the existing fleet reservation and kernel lock before phases that require the worker.
- Assert that the rendered pool target is the test stack before arming failover.
- Preserve the existing exact restoration of the original operator config, rendered config, tuning, service state, and fresh-share proof.
- Add the smallest regression check that proves an upgrade cannot silently discard the active borrow fixture.

## Acceptance

A live matrix run that includes a real RigForge upgrade and reversible control applies completes the authenticated read/control/apply/feed, node failover, moved-subnet worker/hash, fresh accepted-share, and exact-restoration legs without these fixture-induced failures.

VijitSingh97:
## Summary

- carry the debug registry into the appliance environment so SSH-driven recovery uses the test registry
- wait for provisioning units before the restore leg captures its backup
- add the RC2 hostname, diagnostics, approval, backup/restore, and failed-install regression coverage
- re-arm the borrowed RigForge pool after control writes before worker-dependent phases
- fail closed on stale approvals, uncertain detached jobs, lost parent locks, mutable release inputs, unverifiable image provenance, surviving KVM guests, and libvirt query failures
- retry post-promotion reads until both user-facing tags resolve to the captured digest
- refuse a bench image build whose service images cannot be pulled, and dump the guest's image lists when a stack never comes up (#2043)

This combines and supersedes #1939, #1947, and #1982 so their interacting appliance changes can be
reviewed, run through CI, and exercised by one exact-image KVM lifecycle.

Closes #1931.
Closes #1945.
Closes #1966.
Closes #1994.

## Merged up to `develop`

- exact head: `e836aca247f964e8e0ff54c3c35c6bdff1f6ca2a`

`develop` moved under this branch twice since it was opened: #2005 reorganised the repo into feature
folders and #2042 deprecated macOS as a test platform. #2005 **split three files this branch edits**,
so the merge had to relocate this branch's changes rather than take either side:

| file | on `develop` now | this branch's hunks |
|---|---|---|
| `tests/os/run.sh` | 3423 -> 154 lines + `lib/core.sh` + 16 `phases/*.sh` | 23 of 26 placed; 3 blank-line-only dropped |
| `tests/integration/run.sh` | 2551 -> 180 lines + `lib/run-*.sh` | 9 of 10 placed; 1 blank-line-only dropped |
| `scripts/release.sh` | 850 -> 6 files under `scripts/release/` | 10 of 10 placed |

This branch's new files moved into the new layout (four helpers to `tests/integration/lib/`, three
self-tests to `tests/integration/selftest/`; `tests/os/` stays flat, matching its existing
`*-leg.sh`/`*-verdict.sh` siblings), and every reference in both directions was repointed —
including two runtime `source` paths that `lint-path-references` cannot see, because it reads only
the shellcheck comment beside them.

Three consequences of the split worth a reviewer's attention:

- `pv_user`/`pv_pass` are hoisted to the `phase_provision` dispatcher. The reboot and migration
  legs read what the initial leg sets, and after the split a leg-local would be unbound there under
  `set -u`.
- `selftest-run-modules.sh` drops `vm_destroy` from its expected-function list: this branch moves it
  to `kvm-preflight.sh`, where `test-harness-tooling.sh` already exercises it.
- `selftest-curl-credentials.sh` regains `_worker_apply`, which `develop` dropped when the function
  moved. This branch takes its body off argv, so the argv-credential check applies to it again.

No file-budget ceiling is raised. Two release tests that landed 1–2 lines over `develop`'s freshly
recorded ceilings were trimmed inside this branch's own new code, losing no assertion.

## #2043 is this branch's, and it is not an appliance bug

[#2043](https://github.com/p2pool-starter-stack/pithead/issues/2043) reported five battery legs
failing with a stack that never comes up, reproducing on `develop`, and was treated as a
pre-existing defect blocking RC2. It is a build-configuration defect —
[full diagnosis on the issue](https://github.com/p2pool-starter-stack/pithead/issues/2043#issuecomment-5623365221).

`build-image.sh` bakes one image archive, the wizard's dashboard image, so at first boot every other
service is a pull of `pithead-<service>:v$(cat VERSION)`. `VERSION` is `2.0.0` and the newest
release is `v1.20.0`, so those tags exist nowhere public; `TEST_REGISTRY` is set only when
`PITHEAD_TEST_SSH_PUBKEY` **and** a non-default `PITHEAD_REGISTRY` are both present, so `--ssh`
alone points the image at the public registry. The appliance provisions, publishes its dashboard
credentials, and runs zero containers. The tell is `running: 'pithead-wizard '` — the one baked
image up, nothing else ever.

Two guards here, neither of which needs a guest:

- `require_pullable_services` resolves the five refs before anything expensive and refuses, naming
  the refs and the remedy. Driven at tier 1 through `build-image.sh`'s existing
  `PITHEAD_BUILD_IMAGE_TEST` seam with a stubbed `docker manifest inspect`; the all-present case must
  be silent, so a green cannot come from a fixture that never resolved anything.
- `tests/os/zero-container-evidence.sh` is wired into all five legs. `comm -23 want have` non-empty
  names the missing refs; empty says every ref was present and something else refused, with the
  journal tail to say which.

## Validation

Run at the exact head above unless stated.

- `make lint-sh` (shellcheck + shfmt, pinned 0.11.0): pass
- `make test-integration-selftest`: every suite 0 failed except `selftest-redact-ip` (17/4), which
  fails identically on a clean `develop` checkout — controlled, pre-existing, and a BSD/GNU
  divergence of the kind #2042 deprecates macOS for
- `tests/os/selftest-run-modules.sh`, `lint-file-budget`, `lint-path-references`,
  `lint-pithead-build`, `test-inventory`: pass
- every `--self-test` entry point under `tests/os/`: pass, except `soak-probe.sh` (26/3), which this
  branch does not touch and which fails identically on clean `develop`
- new checks verified as controlled pairs, not bare greens: the #2043 guard (8 passed; making it
  never refuse reddens 5), the zero-container dump (removing one probe reddens exactly that case),
  and the repointed approval call-site assertion (a wrong needle still reddens it)

Not yet run, and owed before merge:

- **tiers 1–3 on Linux.** `make test-stack` now refuses on Darwin by design (#2042), and an
  unmodified `develop` scores 148 failures there, so no local tier-1 result from a mac is evidence.
- **the KVM battery.** It has never had a run that *could* pass, so the #1966 rows remain unproven;
  "unverified until the next battery run" in `docs/dev/testing-strategy.md` is still accurate. The
  run must use an image built with the registry override.

## Companion release

Requires p2pool-starter-stack/rigforge#492 (`test: fail closed across combined appliance gates`),
still open as a draft. The two must be tested together: this branch carries the immutable RigForge
build override through the appliance lifecycle and forwards the rollback input to the detached matrix
runner. Neither should merge without the combined acceptance matrix covering both.

## CI

The two red checks described in the previous revision of this body are gone: #2005 has landed on
`develop`, which carried the `grpc` raise that cleared the trivy finding, and the gitleaks findings
belonged to that branch's commits. Both pass at this head.

The earlier claim that three local `selftest-e2e-phases` failures were "the known BSD `wc`
whitespace difference" no longer holds: that suite reports 84 passed, 0 failed here, and its `wc`
uses were already `tr -d '[:space:]'`-guarded. The claim is withdrawn rather than carried forward.


VijitSingh97:
Independent exact-head review: **PASS** at `fd511612634a21ada55a349684c13a3a284f1aeb` against `develop` `a03d0796b8e28d1beca87b5a118a4f681997fb66`.

The correctness and security reviews found and rechecked the fail-closed VM teardown, libvirt query errors, nested restore-artifact retention, approval-fixture cleanup, root-owned release staging, detached parent-lock continuity, RigForge re-arm ordering, immutable Compose provenance, and post-promotion digest visibility retry. No critical, high, or medium actionable defect remains in the PR delta.

Bounded local evidence: release domain 58/0; parent-lock 31/0; borrowed-pool re-arm 9/0; focused appliance and procedure controls passed. Fresh CI and the exact-image KVM battery remain separate execution gates before merge.

VijitSingh97:
## What was done, and how it was tested

### Brought up to `develop`, twice

`develop` moved under this branch three times while it was open: #2005 reorganised the repo into
feature folders, #2042 deprecated macOS as a test platform, and #2046 rewrote `os/KNOWN-ISSUES.md`.
#2005 **split three files this branch edits**, so the merge relocated this branch's changes instead
of taking either side:

| file | on `develop` | hunks |
|---|---|---|
| `tests/os/run.sh` | 3423 → 154 lines + `lib/core.sh` + 16 `phases/*.sh` | 23 of 26 placed; 3 blank-line-only dropped |
| `tests/integration/run.sh` | 2551 → 180 lines + `lib/run-*.sh` | 9 of 10 placed; 1 blank-line-only dropped |
| `scripts/release.sh` | 850 → 6 files under `scripts/release/` | 10 of 10 placed |

Three consequences worth a reviewer's eye: `pv_user`/`pv_pass` are hoisted to the
`phase_provision` dispatcher (after the split a leg-local would be unbound in the sibling legs under
`set -u`); `selftest-run-modules.sh` drops `vm_destroy` from its expected-function list because
this branch moves it to `kvm-preflight.sh`; and `selftest-curl-credentials.sh` regains
`_worker_apply`, which `develop` dropped when the function moved and which this branch takes off
argv.

### #2043 root-caused — and my first diagnosis was partly wrong

The registry half is real: only the wizard image is baked, so every other service is pulled as
`pithead-<service>:v$(cat VERSION)`, and those tags exist nowhere public while VERSION leads the
last release (`v1.20.0` resolves, `v2.0.0` does not — controlled pair).

Running the battery then found a **third** instance of that cause, and the worst one: the phases do
not boot the `--image` they are handed for their own variants, they rebuild via `_build_image`,
which passed no registry — and the documented recipe was `sudo tests/os/run.sh`, whose
`env_reset` drops an exported `PITHEAD_REGISTRY`. **No invocation of the documented command could
have produced a provisionable image.** Fixed: `_build_image` forwards it, the refusal prints inline
instead of only into `/tmp/os-fault-build.log`, and all three documented recipes use `sudo env`.

I also **over-claimed** in my first #2043 comment, and corrected it there: the provision leg — the one
leg #2043 actually controlled — fails with the registry correct. That is a product hang, now filed
with serial evidence as **#2050**. The `develop` control reproduced it because the hang is
version-independent, not because of VERSION.

### Tested

**Tier 1–3, Linux CI at this head:** 21 checks green, 2 skipped, 0 failures; the shell suite reports
**3958 passed / 0 failed**. macOS is a refused platform (#2042) and an unmodified `develop` scores
148 failures there, so no local tier-1 result from a mac was treated as evidence. `selftest-redact-ip`
(17/4) and `soak-probe --self-test` (26/3) fail identically on a clean `develop` checkout —
controlled, pre-existing.

**Tier 4, KVM bench** — the thing this PR exists for. Image built from `769aa29a` against a registry
serving all five first-party images at `v2.0.0`, rebuilt from this head so the bytes match the
source (all five digests changed), with rigforge#492's tree compiled in via
`PITHEAD_RIGFORGE_REF=c581e3c1`. `--phase all` is still running; **57 passed, 1 failed** so far.

What that already establishes:

- **The registry fix works.** `leg 4: stack provisioned; dashboard login captured` passes, the
  `leg 4: stack never came up` failure is **gone**, and `zero-container-evidence.sh` has fired
  **zero** times across every leg — its discriminator (`comm -23` of what compose wants against what
  podman has) came back empty where it did run.
- **The A/B updater is sound:** install to the spare slot, fallback without commit, commit persisting
  across reboot, and operator rollback all pass.
- **Two real product defects found by this PR's new coverage**, which is the coverage doing its job:
  - **#2050** — a faulted setup hangs instead of reopening the wizard. Last serial output at t=44.4s
    (`grep: docker-compose.yml: No such file or directory`), then nothing for two minutes.
  - **#1956** — both boot-menu entries name the same slot: `Pithead 2.0.0 (slot A, current)` and
    `Pithead 2.0.0 (slot A, previous)`, where the committed slot should read slot B. Measurement
    posted to that issue.

### Adversarial review

Two independent passes. The first found no blocking defects. A third pass over rigforge#492 raised a
blocking finding that did not survive checking — `set_cfg`'s unchecked call sites DO fail closed,
because `set -Eeuo pipefail` is live and the phase loop calls each phase bare; proven with a
controlled pair.

### What is still owed

The remaining battery phases (update-dashboard, install, provision, media, rig, fault, reset) are
still running. #1966's rows stay owed until that completes, and the two defects above are real and
open. Nothing here tags or publishes an image.

VijitSingh97:
## Tier-4 battery complete: 189 passed, 3 failed

My earlier comment quoted this run in progress at 57 passes. Final result, superseding that number.

`--phase all` on the KVM bench, image stamped `769aa29ab8cbb8e0ad6c3881eb7abb7c08a0c339`, built
against a registry serving all five first-party images at `v2.0.0` — rebuilt from this head so the
bytes match the source (all five digests changed) — with rigforge#492's tree compiled in via
`PITHEAD_RIGFORGE_REF=c581e3c1`. For comparison, #2043's run on the same bench was **135 passed / 6
failed**.

### All four of #2043's zero-container legs now pass

| #2043's failure | now |
|---|---|
| `leg 4: stack never came up — running: 'pithead-wizard '` | ✓ `leg 4: stack provisioned; dashboard login captured` |
| `stack never came up within 25m` (media) | assertion does not fire — passed |
| `stack never came up before the reset` | assertion does not fire — passed |
| `restore leg: stack never came up after provisioning` | ✓ `a live stack to back up (tor dashboard docker-control docker-proxy caddy )` |

`zero-container-evidence.sh` fired **once** across the whole run — on the provision leg (#2050) —
and where it ran its discriminator (`comm -23` of what compose wants against what podman has) came
back **empty**. No leg failed for want of an image.

The A/B updater is sound end to end: install to the spare slot, fallback without commit, commit
persisting across reboot, operator rollback, mid-write and mid-commit power-cut survival, corrupt
bundle refused without crashing, still updatable after fault injection, disk install, keep-reinstall,
and the rig role.

### The three failures are pre-existing product defects, each filed

- **#1956** — both boot-menu entries name slot A (`Pithead 2.0.0 (slot A, current)` and
  `(slot A, previous)`); the committed slot should read B. #1974 shipped the feature, so the version
  string and the current/previous distinction work and the slot attribution does not.
- **#2051** (new) — a restored machine gets its config **and its original wallet** back and then never
  starts its stack, `podman ps` empty after 900s. Reached only because the earlier assertion in that
  same leg now passes.
- **#2050** (new) — a faulted setup hangs instead of reopening the wizard; last serial output at
  t=44.4s, then two minutes of silence.

None is a defect in this PR. Each is this PR's new coverage doing what it was written for.

### Harness corrections made after watching it run

- `_build_image` now forwards `PITHEAD_REGISTRY`/`_CA` and prints the build-failure reason inline.
  The phases rebuild the image themselves rather than booting the `--image` they are handed, and
  sudo's `env_reset` dropped the variable — so no invocation of the documented command could have
  produced a provisionable image. All three documented recipes now use `sudo env`.
- `zero-container-evidence.sh` was wired into 5 of the 7 sites that assert on a stack coming up, and
  #2051 landed on one of the two missed. Now 7 of 7, with the dump attached at the **call site** for
  the #1091 verdict rather than inside `restore_live_state_verdict`, whose stdout is its message.

### Coverage audit

Filed **#2052**: `rig_minimize_writes` has no test at any tier, including its #1817 guard against
`rm -rf` through a live bind onto `/data`. That is the only appliance feature found in that state. Two
apparent gaps were checked and refuted — the boot-gate failure counter is thoroughly covered at tier 1,
and the CPU governor's runtime value is unprovable under KVM by design (`ConditionPathExists` on a
node a guest lacks), so it belongs to the manual battery.

### Bench restored

Stack back up, container set identical to the baseline captured before the first stop, all ten healthy,
no leftover guest, memory back to 24.7 GiB. One trivial leftover the media leg does not clean up: a
193-byte `pithead-os-test.img.stick.xml`.

PR #2005, opened 2026-09-08
Refactor repository into feature folders and finish oversized module splits

VijitSingh97:
Large, flat source and test directories made unrelated changes collide and obscured which tests belonged to a feature. This finishes the remaining oversized-module splits and groups the repository by subsystem, with matching tests and a contributor map.

## Organization

- Group dashboard browser modules and tests by app, config, workers, XvB, wizard, network, and system. Split components, wizard stages, and styles along existing behavior boundaries using native ESM and CSS imports.
- Group Python services and HTTP views by feature; split polling helpers, persistence stores, notifications, and oversized test modules. Keep database transactions, locks, and polling order intact. Preserve `python -m mining_dashboard.wizard` through its package entry point.
- Group stack tests into feature families, integration helpers/selftests/tools into separate directories, and appliance harness steps into phase modules. Keep explicit source order and fail when a module cannot load.
- Group lint, release, and scheduled-check scripts; update Make, CI, documentation, and tests together. Keep generated CLI assembly, shipped runtime filenames, and the public installer URL stable.
- Add a repository map and one shared `AI_RULES.md`, linked from `AGENTS.md`, `CLAUDE.md`, and `.cursorrules`. Document provider task routing and actual appliance paths (`os/`, `tests/os/`). Add a bounded, tested build-log sanitizer that preserves raw evidence separately.

The file budget targets 400 lines and enforces an 800-line maximum for new or split files. Tool-discovered root configuration remains at its conventional locations. The generated CLI, vendor code, and data/configuration retain their existing exemptions.

A separate portability fix uses an IPv4 expression supported by Ubuntu’s default `awk`, restoring rig coordinator pinning where the prior interval expression returned no address.

The wizard image-verification gate now follows the package's `wizard/server.py`; archive fixtures prove it rejects stale, legacy-path, missing-source, and corrupt inputs.

## Defensive follow-up

The repository scan identified four findings in two shared paths. Restore now rebuilds derived runtime files from validated settings while preserving validated generated credentials and Tor identity. Staging leaves live auxiliary files untouched, and administrative restore retains its deployment status. The environment writer preserves configuration strings literally through Compose parsing, with matching internal reads.

## Validation

- Dashboard: 2,722 passed, 1 skipped; 96.98% total coverage and 97% changed-line coverage. Fake-daemon contracts: 31 passed.
- Frontend: 694 passed; nine representative rendered states remain byte-identical. Browser smoke checks passed in simple and advanced/dark views without missing assets or console errors.
[...]

Issue #1998, opened 2026-09-08
test(integration): add a periodic live XvB routing smoke


VijitSingh97:
## Why

The matrix asserts XvB configuration and #206 proves the stats request uses Tor, but the raffle routing decision is only covered by unit/simulation tests. A live smoke should prove the controller actually moves real proxy traffic between P2Pool and XvB.

## Done when

- [ ] Run only on the reserved bench on a schedule or explicit dispatch, not fork PRs.
- [ ] Start from a known XvB-enabled configuration and wait on controller state, not a fixed sleep.
- [ ] Observe a real routing transition in xmrig-proxy/controller state and the dashboard.
- [ ] Assert routed hashrate/shares remain visible and return to the baseline route afterward.
- [ ] If external raffle state cannot deterministically trigger a transition, document the bounded injection used and retain one real network leg.

Related: #156, #206, #209.

VijitSingh97:
The bounded live XvB routing smoke landed in #2006 (squashed as `9b5fcb4b`) and is wired into the release-gate workflow's `xvb` mode, but this issue is **not** done on two counts.

The routing transition itself has never run. On the bench the gate correctly refused at its first precondition: it moves an *existing* donor route and needs `xvb.enabled=true` already in the box's config, which that box has off. That prerequisite was missing from the flag's documented list and is now written down.

Its other precondition — a PPLNS share in the window — was reached: six borrowed miners at ~194 kH/s against a ~23.9 MH/s sidechain produced shares after about 47 minutes, consistent with `sidechain_difficulty / hashrate`. A bench mining well below the sidechain rate can be hours from one, so that is now a recorded `missing` skip rather than a failure.

"Periodic" also has no schedule yet — the workflow mode exists but nothing invokes it on a cadence.

VijitSingh97:
[...]

Issue #1997, opened 2026-09-08
test(integration): prove image upgrade preserves chain data and secrets


VijitSingh97:
## Why

The live gate covers restart, apply, and backup/restore, but not an upgrade across image versions. The release-risk gap is a new image set appearing healthy only after discarding chain state or rotating operator secrets.

## Scenario

Starting from a known older release on the live box: record chain paths/state and secret fingerprints, pull the candidate images, run the supported upgrade/apply path, then assert continuity.

## Done when

- [ ] The scenario records old/new image digests and exact Pithead SHAs.
- [ ] Monero and Tari reuse the existing data and return to the prior tip without a bootstrap/re-sync.
- [ ] Wallet, proxy, dashboard-auth, RPC, and onion-secret fingerprints are unchanged.
- [ ] Mining resumes and the dashboard reports the existing workers.
- [ ] Failure artifacts are redacted and the baseline is restored.

Keep this separate from one-way schema migrations such as #1129; those retain their migration-specific gates.

VijitSingh97:
The `--image-upgrade` gate landed in #2006 (squashed as `9b5fcb4b`), but this issue is **not** done: the gate has never executed. No machine we have can run it — the bench is ext4 and the snapshot step needs `cp --reflink=always`, the live stack is a source checkout rather than a `current -> pithead-v*` release layout, and `name: pithead` is pinned in Compose so a gate root cannot run beside it. Filed as #2057 with a cheaper remote-Monero variant sketched.

`make_bundle` now writes the `pithead/PITHEAD_COMMIT` the gate reads, which previously no bundle carried — so the artifact side is unblocked. What remains is an environment to run it in, and then the run itself.

Issue #1995, opened 2026-09-08
docs(testing): correct stale tier-4 gap claims for live egress and SSH fault injection


VijitSingh97:
## Problem

`docs/dev/release-server.md` still lists two tier-4 gaps that no longer match `develop`:

- Privacy/Tor egress is described as having no live clearnet-leak assertion, but #274 promoted `bench-verify-egress.sh`; `tests/integration/run.sh` now runs `assert_egress_posture` and `assert_xvb_over_tor` in the live gate (#206).
- Fault injection is described as local-only, but the fault phase uses the same `rx` target abstraction as the rest of the harness and accepts `--host`. What is missing appears to be recorded SSH-path evidence, not the implementation claimed by the table.

This table is the operator's release-risk summary, so stale gaps can cause duplicate work and hide the real remaining proof.

## Done when

- [ ] Re-audit every row against current `develop` and the latest live result.
- [ ] Mark privacy egress as covered, citing #274/#206 and the live assertions.
- [ ] Reword SSH fault injection as an evidence gap if the path is implemented; link the focused live-proof issue.
- [ ] Keep genuine gaps and their issue links in the table.

VijitSingh97:
Noting for the record: #2006 listed this issue as one it closed. That was a mis-mapping in the PR description (the three issue numbers were shifted by one), corrected in a comment there. **This issue is not addressed.**

#2006 did edit `docs/dev/release-server.md` — the trigger/security section, and its claim that a copy-on-write filesystem is "a bonus, not a requirement", which is false for the new `--image-upgrade` gate. It did **not** re-audit the gap table under `## End-to-end coverage & gaps`, which is what the "Done when" list here asks for.

Two of that audit's inputs did move, and are worth citing when someone does it: the live egress assertion now states what it actually measures (persistent direct IPv4 TCP from the bridge apps, not UDP or host-network processes), and `--fault-injection` is no longer local-only — its usage text now reads "works over SSH and locally".

VijitSingh97:
**Ruling:** both findings stand. The residual-gap table in docs/dev/release-server.md gains two rows and the testbench README mirrors them: N-1 encrypted backup restore (#2001, proof pending, existing round-trip evidence is same-version only, so qualify that claim), and the appliance/RAUC half of cross-version upgrade (#2056), distinguished from the Compose/self-deploy half (#1997, blocked on #2057). Link each row to its issue. Then resubmit.

PR #2006, opened 2026-09-09
test(integration): close three tier-4 coverage gaps

VijitSingh97:
## Summary

Closes three tier-4 coverage gaps in the live harness — the paths that only a real, synced,
mining box can exercise — and hardens the rollback net they depend on.

- [#1995](https://github.com/p2pool-starter-stack/pithead/issues/1995) — cross-version image upgrade
- [#1997](https://github.com/p2pool-starter-stack/pithead/issues/1997) — live XvB routing transition
- [#1998](https://github.com/p2pool-starter-stack/pithead/issues/1998) — verified, interrupt-safe restoration

Tier 4 is the only tier that can fail these: every one of them is a claim about a *running* stack
with real chain data, real wallet/onion secrets and real miners attached. Until now each was
asserted by construction or not at all.

### 1. `--image-upgrade <old-sha> <new-sha>` + `--candidate-bundle` (#1995)

Runs a real `pithead upgrade` from a **privately prepared, signed candidate bundle** against the
already-running old images, then proves the things an upgrade is allowed to change and the things
it is not:

| Proven across the upgrade | How |
|---|---|
| Every first-party image is the declared candidate commit | exact 40-hex OCI revision vs `PITHEAD_COMMIT`, not a prefix |
| Third-party images are the ones the signed manifest named | digest-pinned through the authenticated manifest |
| Chain data continues | Monero/Tari height may advance, never regress; chain id unchanged |
| Secrets survive *by category* | wallet / proxy / dashboard / RPC / onion fingerprinted separately |
| Durable telemetry survives | archived rows still present, and `PRAGMA quick_check` passes |
| Workers and mining continue | the exact pre-upgrade worker set, and hashes still advancing |

Trust is external to the candidate: the cosign public key is supplied as a third path and must live
outside the bundle. Archive members are checked before extraction (regular files only, no links).

`--image-upgrade` requires `--local` and `--safety-backup`, and refuses `--no-mining-asserts` —
mining *is* the success signal, so silencing it would leave the gate asserting nothing that matters.

### 2. `--xvb-routing-smoke` (#1997)

Enables XvB at the donor tier on the live stack and follows one bounded XvB→P2Pool transition
through the **real** controller, proxy and dashboard — then restores the exact original config and
verifies it landed. The XvB-over-Tor leg runs the wallet-bearing client in an internal Docker
network whose only peer is Tor, so the call has no other route out.

### 3. Verified, interrupt-safe restoration (#1998)

The rollback net previously warned and carried on. Now:

- `safety_restore_exact` **verifies** the restore: config byte-identical and every secret category
  unchanged. A rollback that is not verified is not a rollback.
- An `EXIT` trap (`arm_safety_abort_restore`) restores on Ctrl-C, a cancelled CI job, or a
  supervisor `SIGTERM`. It **chains** any `EXIT` trap already installed rather than replacing it.
- Writable mount sources get private CoW reflink snapshots before the destructive phase
  (`live-state-support.sh`), restored with a journalled swap that can itself be rolled back.
- `live-supervisor.sh` runs a destructive gate under `systemd-run`, so cancelling the Actions job
  cannot kill a rollback mid-flight; a heartbeat lets the gate notice the runner went away and
  unwind deliberately.
- The archive is retained — never deleted — whenever restoration did not verify.

## Changes to existing behaviour

These are assertion changes, not additions, and each one closed a way of passing without proving:

- **`assert_egress_posture` skipped on a flag, not a fact.** `MONERO_CLEARNET_SYNC=true` stays set
  in `.env` long after the sync finishes, silently retiring the all-Tor egress gate for the rest of
  the run. It now skips only while the entrypoint's `.synced` marker is still absent.
- **The egress verdict claimed more than it measured.** The probe observes persistent direct IPv4
  TCP from the bridge networks; it does not see UDP or host-network processes. The assertion now
  says that, and `bench-verify-egress.sh` returns INCONCLUSIVE rather than silently skipping a
  container it expected and could not read.
- **`assert_telemetry_tables_present` only checked table *existence*.** It now runs
  `PRAGMA quick_check` and pins the five backbone schemas (these five take no additive migrations),
  naming the drifting table when they differ.
- **Readiness never looked at Tari** — a box whose merge-mining scenarios would start from an
  incomplete chain passed as fit.
- **`--check` in `e2e.sh` deployed, borrowed a miner and took a backup** to run read-only
  assertions, and graded the *undeployed* e2e tree. It now grades the checkout the stack actually
  runs from and mutates nothing.
- **The readiness/current-state pre-gate only warned**, so a bench that was already broken had its
  fallout reported as a branch regression. Both are now binding.

## Operator prerequisites

The release-gate workflow will **refuse to run** until these exist — deliberately fail-closed:

- a `release-server` GitHub environment restricted to the default branch;
- a `RELEASE_GATE_ACTORS` repository variable (comma-separated logins);
- a root-owned, non-group/other-writable cosign public key on the runner, byte-identical to the
  repo's `cosign.pub`.

`--image-upgrade` also needs a candidate bundle built without a public release or `latest` tag.
The repo does not ship a candidate-only producer; that is an explicit operational prerequisite,
recorded in `docs/dev/integration-testing.md`.

## Review notes

This branch was opened before [#2005](https://github.com/p2pool-starter-stack/pithead/pull/2005)
reorganised `tests/integration/` into `lib/` + `selftest/` + `tools/` and split `run.sh` from 2551
lines to 185. It has been merged onto that layout rather than re-landing its own parallel split:

- The behaviour lands in develop's modules (`lib/run-safety.sh`, `lib/run-scenario.sh`,
  `lib/run-cli.sh`, `lib/run-matrix.sh`); only genuinely new code is new files under `lib/`.
- This branch's `stop_harness()` was **dropped as superseded** — develop already ships
  `lib/detached-harness.sh`, whose `drain_harness` proves the identity of the process it signals
  via `/proc/<pid>/stat` starttime instead of killing a bare pgid read from a file.
- Comments this branch had deleted to stay under the `e2e.sh` and readiness file budgets are
  restored; the new logic is extracted into `lib/detached-harness.sh` instead.

## Verification

`make lint-sh` (shellcheck 0.11.0 + shfmt), the file-budget gate, the path-reference gate and the
whole `tests/integration/selftest/` suite pass locally. `selftest-live-gates.sh` adds 35 pure-logic
assertions — trust verdicts, continuity predicates, snapshot/rollback behaviour, trap composition
and the supervisor's exit-status transport — with no server required.

Two behaviours found by an adversarial pass are fixed here, each with a check that is red against
the unfixed code:

- `restore_state_snapshots` recorded a swap on the rollback list only *after* it succeeded. A swap
  that failed mid-way — source moved aside, replacement move failed, inner recovery failed too —
  left the live path absent and that entry missing from the list the rollback walks.
- `bench-verify-egress.sh` made a stopped Tor container INCONCLUSIVE, which is exactly the state
  `fault_tor_down` creates on purpose (#563). That check would have gone permanently red and read
  as tooling breakage rather than a leak.

### Tier-4 run on the bench

Run against a real synced, healthy stack (10/10 containers), with an unmodified `develop` control
run back-to-back against the same box:

| | this branch | `develop` control |
|---|---|---|
| `--readiness` | 7 passed, 0 failed | — |
| `--check` | 65 passed, **2 failed** | 65 passed, **2 failed** |

The two `--check` failures are `workers online (>= 2)` and `stratum total hashes > 0`. No miner was
attached (the proxy reported `miners: 0`), and the failure sets are **byte-identical** between the
branch and the control — so they are the bench's state, not a regression.

Confirmed working on real hardware:

- the new **Tari readiness** assertion passes (it previously did not exist, so a box whose
  merge-mining scenarios would start from an incomplete chain read as fit);
- the reworded **egress** assertion runs rather than skipping — the `.synced`-marker condition
  correctly declined to exempt a box whose `CLEARNET_SYNC` flag is still set;
- the **strict telemetry schema** check passes against the live database — this was the riskiest
  change, since it pins exact column lists;
- **skip accounting** improved as intended: 3 legs recorded (1 by-design) vs the control's 2 legs
  (0 by-design) — the XvB-over-Tor leg was previously an uncounted log line;
- `bench-verify-egress.sh` correctly reports `wallet-rpc` (newly covered) as clean and `tari-wallet`
  as *absent from the active compose profile* rather than a false INCONCLUSIVE.

The `fault_tor_down` fix was proven directly, with Tor actually stopped:

| invocation | result |
|---|---|
| default | `rc=2`, `INCONCLUSIVE — Tor sockets are unreadable` — i.e. #563 would have been permanently red |
| `--allow-tor-down` | `rc=0`, and all five app containers still graded clean while SOCKS was down |

That is the #563 assertion actually doing its job. The box was restored and verified afterwards:
all containers healthy, `pithead status` exit 0, and the live containers still using the on-disk
`.env`.

### The bench run found a real defect, which is why it was worth doing

The rollback net was **guaranteed to fail and to leave the stack STOPPED** on any Docker host, and
it did exactly that here. `verify_tor_egress_firewall` compared its expected ruleset to `iptables
-S` output as text, but `iptables -S` re-prints rules in its own canonical form — the `-m comment`
match moves after the `-s`/`-d` selectors and a bare host gains `/32` — so the compare differed on
every rule carrying `-s`, on a box whose firewall was correct. `safety_restore_exact` calls
`strict_pithead up`, which refuses when that verifier fails, so every rollback failed; and because
its first act is `pithead down`, and the EXIT trap then retried the same restore, the run ended with
the stack down.

It now asks iptables with `-C` using the same spec the installer uses, plus a tagged-rule count and
a DROP-is-last check. A failed rollback brings the stack back up, and the trap does not retry one
that already failed. Same trigger, re-run on the same box:

| | before | after |
|---|---|---|
| rollback | `✗ safety rollback restored the exact healthy baseline` | `rollback complete — exact config and … baseline verified` |
| box left | **0 containers (DOWN)** | **10 healthy, `pithead status` 0** |

**Still not exercised on hardware:** the `--image-upgrade` gate (no machine can run it — ext4 has no
reflink, the bench runs a source checkout, and the Compose project name is pinned; filed as #2057,
and `make_bundle` now writes the `PITHEAD_COMMIT` it needs) and the XvB routing *transition* itself
(the gate moves an existing donor route and needs `xvb.enabled=true` already in the box config; its
preconditions, including a borrowed-miner PPLNS share, were exercised). Both are covered by the 42
pure-logic assertions in `selftest-live-gates.sh`.


VijitSingh97:
## Where this stands

CI green. Two adversarial review passes; everything they found is fixed and each fix carries a check
that is red against the unfixed code. Merged `develop` twice, onto the #2005 feature-folder layout.

Below is what each new test actually proves — and, more usefully, what it does not. Three of the
five have run against real hardware; two have not, and I've said so rather than implying otherwise.

---

### 1. `--image-upgrade <old-sha> <new-sha>` — **logic only, never run on hardware**

**Proves:** that a real `pithead upgrade` from a signed bundle preserves what an upgrade must not
change. Every first-party image is the declared candidate commit (exact 40-hex OCI revision, not a
prefix); third-party images are the ones the signed manifest named (digest-pinned); Monero/Tari
height may advance but never regress and the chain identity holds; wallet, proxy, dashboard, RPC
and onion secrets are fingerprinted **per category**, so a verdict says which one moved; durable
telemetry rows survive and `PRAGMA quick_check` passes; the exact pre-upgrade worker set returns and
hashes advance. Trust is external to the candidate — the cosign key is a third argument that must
match the bundle's embedded copy, and archive members are validated before extraction.

**Does not prove:** anything yet on real hardware. It has **never executed** — see #2057. Three
independent reasons: the snapshot step needs `cp --reflink=always` and the bench is ext4 (measured,
not assumed); `prepare_baseline_install` requires a `current -> pithead-v*` release layout and the
bench runs a source checkout whose images carry no 40-hex revision; and `name: pithead` is pinned in
Compose, so a gate root cannot run beside the live stack. Also, by construction it does not claim no
same-chain bytes were re-downloaded — only that the height did not regress.

**Newly unblocked:** `make_bundle` never wrote `pithead/PITHEAD_COMMIT`, which the gate requires, so
*no bundle in existence satisfied it*. It does now, and refuses a short or unset `GIT_COMMIT` at cut
time rather than shipping a bundle that fails later on a reserved box.

### 2. `--xvb-routing-smoke` — **preconditions exercised on hardware; the transition itself is not**

**Proves:** one bounded routing transition through the real controller, proxy and dashboard, then an
exact restore that is verified rather than assumed. The privacy leg runs the wallet-bearing client
in an internal Docker network whose only peer is Tor, so the call has no other route out.

**Does not prove:** anything about sustained donation behaviour, payout accounting, or tier
selection — it takes the lowest tier deliberately, to keep the window short and the run bounded.

**Not run on this bench, for a documented reason:** the gate moves an *existing* donor route and
requires `xvb.enabled=true` already in the box's config; this bench has it off. That prerequisite
was not in the flag's documented list, so the first real run spent a reservation discovering it —
now written down. Six miners were borrowed to satisfy the PPLNS-share precondition (3 shares landed
after ~47 min at ~194 kH/s into a 23.9 MH/s sidechain), and all six were restored to their original
pool afterwards.

**What changed after measuring it:** the gate collapsed four preconditions into one failure with an
opaque message, so an idle bench reported identically to a broken controller. Those are now
separate, and the missing-PPLNS-share case is a recorded **skip** in the summary's `missing` column
rather than a failure — because a share is an *input* this gate needs, not a property it tests.
Expected time to one is `sidechain_difficulty / your_hashrate`; measured on the bench that is ~1.8 h
for one miner and ~18 min for six, so failing for it reported an idle bench as a product defect.

### 3. Verified, interrupt-safe rollback — **was broken; found on the bench, fixed, re-proven**

**Proves:** a restore now has to *land*: config byte-identical and every secret category unchanged,
or the archive is retained and the run says so. An EXIT trap restores on Ctrl-C, a cancelled Actions
job or a supervisor `SIGTERM`, and it **chains** any EXIT trap already installed instead of
replacing it. Writable mounts get CoW reflink snapshots restored through a journalled swap that can
itself be rolled back.

**What the bench run found — the most important result in this PR.** The rollback net was
*guaranteed to fail and to leave the mining stack stopped*, on any Docker host. It did exactly that
here. Three faults, one root cause:

`verify_tor_egress_firewall` built its expected ruleset with the comment match first —
`-A DOCKER-USER -m comment --comment TAG -s 172.28.0.25 -j ACCEPT` — but `iptables -S` re-prints a
rule in **its own** canonical form: the comment moves after the `-s`/`-d` selectors and a bare host
gains `/32`. So the literal string compare differed on every rule carrying `-s`, and returned `rc=1`
on a box whose firewall was correct and whose egress this same harness had just verified clean.

That was load-bearing: `safety_restore_exact` calls `strict_pithead up`, which refuses when the
verifier fails — so **every** rollback failed. And two amplifiers turned that into a stopped box:
`safety_restore_exact`'s first act is `pithead down`, and it returned on failure without bringing
the stack back; then the EXIT trap re-ran the same restore, repeating that `down`, so the run *ended*
stopped even after `restore_baseline` had put it back.

Fixed by asking iptables with `-C` using the same spec `apply_tor_egress_iptables` installs —
iptables' own rule equality rather than a normalisation of ours to drift out of sync — plus a
tagged-rule count and a check that the subnet-wide DROP is last (which `-C` cannot see, and which
changes what actually egresses). The rollback now restores service even when it fails, and the trap
does not retry a restore that already failed.

Controlled pair on the same box, same trigger:

| | before | after |
|---|---|---|
| rollback | `✗ safety rollback restored the exact healthy baseline` | `rollback complete — exact config and … baseline verified` |
| box left | **0 containers (DOWN)** | **10 healthy, `pithead status` 0** |

**Also found by review, before the bench:** `restore_state_snapshots` recorded a swap on the
rollback list only *after* it succeeded, so a swap failing mid-way left the live path absent and that
entry missing from the list the rollback walks. Now recorded before the attempt.

**Still not proven on hardware:** the CoW snapshot/restore path itself — same `#2057` blocker. The
trap composition, the supervisor's exit-status transport, the rollback journal and the firewall
verifier's four failure modes are covered by 42 pure-logic assertions in `selftest-live-gates.sh`.

### 4. Egress observation — **proven on hardware, including the failure mode**

**Proves:** no persistent direct IPv4 TCP egress from the bridge app containers. The wording is
deliberately narrower than before: it observes IPv4 TCP from the bridge networks and **does not**
capture UDP or host-network processes, so a clean verdict is evidence, not a proof of #270. Two apps
that were silently uncovered (`wallet-rpc`, `tari-wallet`) are now graded, and a container expected
by the active Compose profile but not running is INCONCLUSIVE rather than a silent skip.

**Proven directly, with Tor actually stopped:**

| invocation | result |
|---|---|
| default | `rc=2  INCONCLUSIVE — Tor sockets are unreadable` |
| `--allow-tor-down` | `rc=0`, all five app containers graded clean while SOCKS was down |

That matters because `fault_tor_down` stops Tor *on purpose* (#563). Without the waiver that check
would have been permanently red, and red as "INCONCLUSIVE" — which reads as tooling breakage, not
as a leak. The waiver is wired to exactly one call site and waives exactly one positive control.

### 5. Readiness and current-state changes — **proven on hardware**

Run against a real synced stack, with an unmodified `develop` control back-to-back on the same box:

| | this branch | `develop` control |
|---|---|---|
| `--readiness` | 7 passed, 0 failed | — |
| `--check` | 65 passed, 2 failed | 65 passed, 2 failed |

The two `--check` failures are `workers online (>= 2)` and `stratum total hashes > 0` — no miner was
attached at the time, and the failure sets are **byte-identical** between branch and control.

- **Tari readiness** is new; a box whose merge-mining scenarios would start from an incomplete chain
  previously read as fit.
- **Egress skip** no longer trusts a flag. `MONERO_CLEARNET_SYNC` stays `true` in `.env` long after
  the sync finishes, which silently retired the gate for the rest of a run; it now skips only while
  the entrypoint's `.synced` marker is absent. On the bench it correctly declined to skip.
- **Telemetry** went from "these five tables exist" to `PRAGMA quick_check` plus exact column lists,
  naming the drifting table. Riskiest change in the set — it passes against the live database.
- **Skip accounting**: the XvB-over-Tor leg was an uncounted log line. Branch reports 3 skipped legs
  (1 by-design), control reports 2 (0 by-design).
- **Reflink probe**: readiness now *attempts* a reflink copy instead of inferring from the fstype
  string, and warns when `--image-upgrade` could not take its snapshots. Warn, not fail — a box
  without reflink is a fine release server for every other phase.

---

## Gaps filed rather than silently carried

- **#2056** — the KVM battery never upgrades a real previous release: `_build_image` stamps the same
  HEAD into both A/B slots and varies only a marker string, so schema migrations, unit renames and
  `/data` layout changes between versions are unmeasured. That is the appliance half of the same
  surface this PR covers for self-deploy.
- **#2057** — `--image-upgrade` has no machine that can run it (the three reasons above), with a
  cheaper remote-Monero variant sketched for whoever builds the environment.
- **#2058** — the Caddy-fronted `/metrics` leg is skipped as `[missing]` on every run because
  nothing supplies `IT_DASHBOARD_PASSWORD`.

## Review notes

This branch predates #2005, which reorganised `tests/integration/` and cut `run.sh` from 2551 lines
to 185. It is merged onto that layout rather than re-landing its own parallel split. Its
`stop_harness()` was **dropped as superseded** — `lib/detached-harness.sh:drain_harness` already
proves the identity of the process it signals via `/proc/<pid>/stat` starttime, where this branch
killed a bare pgid read from a file. Comments deleted on this branch to fit under the `e2e.sh` and
readiness file budgets are restored, with the new logic extracted into `lib/detached-harness.sh`
instead.

VijitSingh97:
## Correction: the issue mapping in this PR's description was wrong

The squash commit message is permanent, so recording this here. The description listed the three
issues shifted by one, and I carried that through from the original draft without checking the
titles against the numbers. The real mapping, and the honest status of each:

| issue | what it actually asks | status after this merge |
|---|---|---|
| **#1997** | *prove image upgrade preserves chain data and secrets* | **not proven.** The gate is implemented and unit-covered, but has never executed — no machine can run it (#2057). Leave open. |
| **#1998** | *add a periodic live XvB routing smoke* | **partially.** The gate is implemented and wired into the release-gate workflow's `xvb` mode, but the routing transition itself has never run, and "periodic" has no schedule yet. Leave open. |
| **#1995** | *correct stale tier-4 gap claims in `docs/dev/release-server.md`* | **not addressed.** This PR edited that file's trigger/security section and its copy-on-write claims, but did not re-audit the gap table at `## End-to-end coverage & gaps`, which is what the issue asks for. Leave open. |

None of the three is closed by this merge, and none should be. The `Closes:` list in the
description never used the `Closes #N` keyword — it was a markdown link list — so nothing
auto-closed either; that turns out to be the right outcome for the wrong reason.

What this PR *did* land is in the description and the status comment above: the three gates, the
rollback fixes the bench run surfaced, and the assertion corrections. What it did not land is the
proof that two of those gates work against real hardware.

Follow-on gaps filed while doing this: **#2056** (KVM never upgrades a real previous release — both
A/B slots are built from the same HEAD), **#2057** (the `--image-upgrade` gate has no machine that
can run it), **#2058** (the Caddy-fronted `/metrics` leg is skipped as `[missing]` on every run).

PR #2007, opened 2026-09-09
test(os): #1932 the install-phase pre-fill control reads the wizard's state route too

### LINKED ISSUE #1932: KVM provision phase: the browser-shaped submit (#1847) gets no served config from the wizard, and the harness cannot say why

VijitSingh97:
Closes #1932

### DIFF
--- a/tests/os/provision-browser-submit.sh
+++ b/tests/os/provision-browser-submit.sh
@@ -2,7 +2,7 @@
 #
 # The browser-shaped wizard submit for phase_provision (#1846). Sourced by tests/os/run.sh.
 # What a person's browser sends is not the no-JS field form: wizard.mjs takes the page's own
-# served config (/api/state .config — the reference merged with the last attempt), sets the
+# served config (/api/wizard-state .config — the reference merged with the last attempt), sets the
 # operator's answers on the same paths the page uses, and POSTs it whole as `config=<JSON>`
 # beside `auth_mode=auto` (the recommended "generate a strong password for me"). The old leg
 # posted `monero_wallet=…&pool=…`, which build_config() turns into a config server-side, so the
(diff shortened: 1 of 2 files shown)


VijitSingh97:
duplicate

PR #2008, opened 2026-09-09
fix(wizard): validate backup entry kinds before applying files

### LINKED ISSUE #1842: The wizard's credentials card is written at default umask — since #1836 it carries the rig's control token

VijitSingh97:
Closes #1842


VijitSingh97:

PR #2009, opened 2026-09-09
fix(os): #1896 the dashboard quadlet carries the three dashboard-onion values

### LINKED ISSUE #1896: The appliance's dashboard container never receives DASHBOARD_ONION_*, so the header's .onion block cannot render there

VijitSingh97:
Closes #1896


VijitSingh97:

PR #2010, opened 2026-09-09
test(p2pool): #1903 the untouched shapes pin the token count, and the drop line is claimed only on a drop

### LINKED ISSUE #1903: tari.mode "off" still passes --merge-mine to P2Pool: the entrypoint drops no argv

VijitSingh97:
Closes #1903, #2


VijitSingh97:

PR #2011, opened 2026-09-09
fix(pithead): #1930 the backup header names the dashboard DB among what it always keeps

### LINKED ISSUE #1930: 17-backup-restore.sh says the dashboard DB is excluded by default; its own code always includes it

VijitSingh97:
Closes #1930

### DIFF
--- a/lib/pithead/17-backup-restore.sh
+++ b/lib/pithead/17-backup-restore.sh
@@ -1,7 +1,7 @@
 # --- Backup / Restore ---
-# Protect the irreplaceable bits: config.json, .env (secrets), the Caddyfile, and the Tor
-# data dir (onion service keys). The blockchains and dashboard DB re-sync, so they're excluded
-# by default; pass --with-chains to also fold in the Monero/Tari/P2Pool data dirs.
+# Protect the irreplaceable bits: config.json, .env (secrets), the Caddyfile, the Tor data dir
+# (onion service keys), and the dashboard data dir (its DB never re-syncs). The blockchains do
+# re-sync, so they're excluded by default; pass --with-chains to fold in Monero/Tari/P2Pool data.
 #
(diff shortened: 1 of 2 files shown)


VijitSingh97:

PR #2012, opened 2026-09-09
test(os): #1935 review follow-up — a multi-line tar stderr stays on the reason's one line

### LINKED ISSUE #1935: verify-image: the "baked wizard image contains the tree's wizard.py" check failed once in the RC1 battery, re-passes on the same image, and swallows its own reason

VijitSingh97:
Closes #1935

### DIFF
--- a/tests/os/verify-image.sh
+++ b/tests/os/verify-image.sh
@@ -293,10 +293,10 @@ if [ -f ./pithead ] && [ -f dashboard/mining_dashboard/wizard.py ]; then
                     WIZ_SHIPPED="$WIZ_TMP/shipped-wizard.py"
                     break
                 fi
-                WIZ_WHY="tar -xOf $member from $(basename "$layer" | cut -c1-12) failed: $(head -c 120 "$WIZ_TMP/extract.err")"
+                WIZ_WHY="tar -xOf $member from $(basename "$layer" | cut -c1-12) failed: $(head -c 120 "$WIZ_TMP/extract.err" | tr -c '[:print:]' '?')"
             done
         else
-            WIZ_WHY="tar -xzf $(basename "$WIZ_ARCHIVE") failed: $(head -c 120 "$WIZ_TMP/untar.err")"
+            WIZ_WHY="tar -xzf $(basename "$WIZ_ARCHIVE") failed: $(head -c 120 "$WIZ_TMP/untar.err" | tr -c '[:print:]' '?')"
         fi
     fi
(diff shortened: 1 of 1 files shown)


VijitSingh97:

PR #2013, opened 2026-09-09
test(os): #1936 the route comment sits on its own line — shfmt 3.13 pads consecutive trailing comments

### LINKED ISSUE #1936: KVM pre-fill submit leg: the "pre-fill armed" control reads /api/state once and reds with one word (same shape as #1932)

VijitSingh97:
Closes #1936


VijitSingh97:

PR #2014, opened 2026-09-09

### LINKED ISSUE #1945: KVM restore leg: the harness's backup is granted the mutation lock while the firstboot's `pithead up` is still in its health wait; two compose ups overlap and tor ends unhealthy

VijitSingh97:
Closes #1945, #1934, #1939


VijitSingh97:

PR #2015, opened 2026-09-09
test(os): #1892 verify-image refuses every test-registry drop-in, refuses to run without its sibling, and build-image drops a dead build-arg

VijitSingh97:


VijitSingh97:

Issue #1955, opened 2026-09-07
Wizard: a failed install leaves the page stuck on 'Installing… Working… Loading…' with no way back and every field lost


VijitSingh97:
## What the operator hit (RC1 manual test, 2026-09-06)

A Pithead + RigForge install with a remote Tari node whose gRPC was not reachable from the appliance ended on:

> Pithead setup — Installing. Takes a few minutes. Do not power it off.
> Install failed: cannot reach the remote Tari node at <host>:18142 — check the host, the port, and that the node allows LAN access (tari.grpc_lan_access on a Pithead host)
> Loading…

There is no button. Nothing to go back to. After a refresh the page reads "Installing… Working… Loading…" and stays there. The machine never gets a dashboard, and every field the operator typed is gone. The failure message itself was right (the remote node was not LAN-exposed); the shape around it is the defect.

## What should happen

1. **A failed install returns to the form**, not to a dead status page: one button, "Back to the settings", with every field the user entered still filled (the spool the host consumed carries them; the wizard keeps its own copy until the install succeeds).
2. **The status page is honest about state.** "Installing" only while the runner is installing; a failure ends the "do not power off" banner; a refresh after a failure lands on the failure, not on "Working…".
3. **The remote-node check happens before the install starts** (#1889 covers the probe): a visible "Reaching your remote nodes…" step that lists what is being tried for Monero and Tari, moves on by itself when both answer, and on failure offers the back button above with the fields kept. This issue is the recovery path; #1889 is the prevention.
4. The stray "Loading…" line is #1868.

## Bar

Tier 1: `wizard.py` state machine — a runner failure maps to a `failed` state that renders the message + back button and preserves the submitted body; a refresh in `failed` renders `failed`. Tier 3: the KVM provision leg with an unreachable remote Tari host asserts the failure page has the back button and the form comes back filled.

PR #2016, opened 2026-09-09
fix(appliance): preserve saved addresses and mint long machine names

### LINKED ISSUE #1954: Wizard: the reinstall pre-fill carries the removed 1.x alias keys into the new config, and validation then refuses the whole install
## What the operator hit (RC1 manual test, 2026-09-06)

Running the wizard on a machine that had been set up before (the previous test image), the install was refused at validation:

> Parsing configuration... [ERROR] config.json sets both a removed 1.x key and its replacement to different values (xvb.enabled and xmrig_proxy.enabled). The 1.x names were removed in 2.0.0; keep the replacement, delete the old key, and re-run.

The wizard itself writes neither `xmrig_proxy.*` key (grep on `develop`: zero hits in `wizard.mjs` / `wizard.py`). What it does do is **pre-fill from the previous machine's config** (#794, #1846/#1852), and that config still carried the 1.x alias `xmrig_proxy.enabled` from the earlier image. The wizard then wrote its own `xvb.enabled` beside it, and #1840's fail-closed migration refused the pair. The user sees an error about a key they never typed, has no field to delete it from, and cannot proceed.

## What should happen

- The pre-fill runs the same alias migration the parser runs (`lib/pithead/27-defaults-and-config-validation.sh`, the #1840 in-place shape) BEFORE it becomes the form's starting point, so removed keys never reach a 2.0 config the wizard writes.
- Belt and braces: the wizard's submit path strips any key the reference schema no longer knows and says so in the change summary, rather than letting the parser refuse a config the user could not have fixed.
- The error text for this case, if it can still occur, names the wizard's own remedy, not "delete the old key" — an appliance user has no editor.

## Bar

Tier 1: a pre-fill fixture carrying `xmrig_proxy.enabled` + `dashboard.workers[]` yields a form state and a written config with only the 2.0 keys; a mutation control that drops the migration goes red. Tier 3: the KVM reinstall leg (#1852's) seeds a 1.x-shaped previous config.

VijitSingh97:
Closes #1954, #1955


VijitSingh97:

PR #2017, opened 2026-09-09
fix(config): normalize array audit paths

### LINKED ISSUE #1959: Every config field editable from the dashboard, sensitive ones behind approval: payout addresses and the XvB endpoint are refused today, and the appliance has no other way
## Operator ruling (RC1 manual test)

"They should be able to update the wallet payout address from the config page. They can't change the XvB URL. We should try to make all the fields configurable." On the appliance there is no other way: a field the dashboard refuses is a field the user cannot change at all (#786).

## Today

The control channel's true-allowlist (#33) refuses funds- and traffic-redirecting keys: payout addresses, node endpoints (#1888 reopens those), the XvB endpoint, and more. The Configuration view greys or hides them and the apply preview refuses them.

## What should happen

- Every key in `config.reference.json` is editable from the Configuration view. The refusal set becomes an **approval set**: payout addresses, the XvB endpoint, node endpoints, control and auth settings commit only through the approval step, with the preview showing the full old and new value (a payout address is never truncated in the preview) and, for payout addresses, a typed confirmation of the last characters.
- The never-approve set of `pithead-media-config` (ssh.*, approval identity, dashboard password, tamper toggles) is unchanged: those are physical-presence changes by design.
- The threat model is written in the PR: an authenticated dashboard session that can repoint payouts is a real capability; it stays behind the dashboard login plus the approval identity, and the audit row names the actor.

Tier 1: policy tests per key class (`tests/stack/test-control-*`); tier 3: an appliance round trip that changes a payout address through approval and reads it back. Docs: `docs/configuration.md` lists what needs approval.

VijitSingh97:
Closes #1959


VijitSingh97:

PR #2018, opened 2026-09-09
test(integration): #1725 the by-name exclusion is gone, so the invariant reaches run.sh too

### LINKED ISSUE #1725: tests/stack: the four `mktemp -d` sites #1721 excluded by name still carry the bare form

VijitSingh97:
Closes #1725

### DIFF
--- a/docs/dev/integration-testing.md
+++ b/docs/dev/integration-testing.md
@@ -849,27 +849,34 @@ a working `mktemp` holds the other side, so deleting the sandbox logic outright 
 last case drives the pre-fix expression against the same fixture and requires it to destroy the
 sentinel, so a row that is green because the fixture never armed fails rather than reads as proof.
 
-A second block in that file pins what the suite does *not* reach. The
+A second block in that file holds the invariant that keeps the constructor honest. The
 [#1705](https://github.com/p2pool-starter-stack/pithead/issues/1705) invariant reads
-`tests/stack/lib.sh` and the `test-*.sh` domain files and names `tests/stack/run.sh` and the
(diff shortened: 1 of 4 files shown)


VijitSingh97:

Issue #1965, opened 2026-09-07
Backup from the dashboard on RC1 takes the stack down and it does not come back; the card's timing note is wrong and its timeout text is the OS-update flow's


VijitSingh97:
## Operator (RC1 manual test)

Backup from the dashboard: it ran well past a minute (the card's note promises less), then the card failed with the OS-update flow's "Stopped waiting … if the version is unchanged" text, and then **the dashboard never came back**. Root cause is #1931 on this image: a backup stops and restarts the whole stack, and the restart under the control runner pulled from the public registry (no 2.0.0 there) instead of the pinned one — the same failure the RC1 battery hit on its restore leg — so the stack stayed down. The operator's machine was left without a dashboard.

## What should happen

1. **Timing:** measure a backup on the appliance (the operator asks for a timed test) and put the measured range in the card, with the plain warning that the dashboard is unavailable while the stack restarts.
2. **Error text:** this card's own words, never the OS-update flow's.
3. **Never leave the box dark:** a failed post-backup restart must retry through the boot path (which carries the registry pin) or roll back to the running stack; the RC2 image carries the pin in the environment (#1931) so the pull itself succeeds.
4. Recovery for anyone who hits this on RC1: power-cycle the appliance; the boot units carry the pin and bring the stack back.

Tier 3: the KVM battery's restore leg already exercises backup + restart (it went red for exactly this); extend it to assert the dashboard answers again within the measured window.

PR #2019, opened 2026-09-09
fix(restore): reject nested destination redirects (#1965)

### LINKED ISSUE #1967: Installed runtimes must disable Compose source-build fallback
Installed appliances and release bundles should never fall back to building images from source.

When an image pull fails, `compose_up_checked` runs `docker compose up` without `--no-build`. The shipped compose file retains first-party `build:` entries, while a release bundle deliberately omits those source contexts. The same issue exists in the direct Tor `compose up` calls in `lib/pithead/32-onion-provisioning.sh`. A registry failure therefore produces a second, misleading build-context failure on a machine that cannot build its images.

This was observed during the restart after a backup in #1931. Correct registry propagation fixes that trigger, but a missing tag, registry outage or other pull failure can still reach this installed-runtime path. PR #1737 addresses compose/image version alignment; it does not disable this build fallback.

Acceptance:

- Every installed-runtime `compose up` path explicitly disables source builds, including Tor provisioning/recreation.
- Source checkouts keep their existing development build behavior.
- Pull/start failures preserve the original failure and useful recovery context.
- A focused regression distinguishes source and installed layouts, with combined appliance lifecycle/backup testing on the release candidate.

VijitSingh97:
Closes #1967

### DIFF
--- a/docs/operations.md
+++ b/docs/operations.md
@@ -585,11 +585,6 @@ are configured on the destination appliance. It puts the files back, fixes
 Tor key ownership so the onion address returns unchanged, and restores hashrate history and
 dashboard settings.
 
-> NOTE: The archive stores the source box's absolute paths, and `restore` puts every file back
-> exactly where it came from. On a machine laid out differently (another user, another install
-> directory), the files land in the old box's directory tree — not the install you ran `restore`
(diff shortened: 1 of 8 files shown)


VijitSingh97:

PR #2022, opened 2026-09-09
test(integration): #1812 the inventory floor gate gets a row proving it can FAIL

### LINKED ISSUE #1812: tests/inventory.sh's new floor gate has no test that it can FAIL
#1808 merged (`69cd108caa882fe2f60140d436be79100943859d`) giving `tests/inventory.sh`'s aggregate
counters real floors instead of a zero test. **It shipped with a gap its own author named in the PR
body rather than hiding, and the gap outlives the PR, so it needs a home here.**

## The gap

**Nothing under `tests/integration/` references `tests/inventory.sh`.** The evidence that the new
floor gate fires is the author's mutation battery and my own re-measurement, both run against that
PR. Neither runs again. The gate is exercised in CI as a step inside `Shell tests`, so a floor that
is WRONG will be noticed — but only by reddening, and only once someone changes the counts.

## Why it is worth a permanent row rather than a note

Two failure modes, and they fail in opposite directions:

1. **A floor set too high reds every lane's PR**, not just the one that moved the count, because the
   gate sits inside the one required check on `develop-v2`. That is a fleet-wide stall from a
   one-line constant.
2. **A floor set too low is silent forever.** It is the same shape the floors replaced: a check that
   cannot say anything but "fine". At merge I measured `n_node` 490 and `n_selftest` 164 against the
   PR's own lower figures, so today's headroom is real — but headroom is a fact about today's tree,
   and nothing re-measures it.

The property that wants pinning is not "the counts are above the floors" (CI already shows that every
run). It is **"the gate can still fail"** — that a count dropping below its floor actually reds, which
is the one thing a passing gate never demonstrates.

## What would discharge this

A row that drives `tests/inventory.sh` with a seeded low count and asserts the non-zero rc, plus the
clean-tree control alongside it so the row is shown able to give both answers. Tier is
`tests/integration/` by the four-tier model — it is a script's observable behaviour, not a unit.

Filed by the reviewer lane, not the author, so it is not competing with anything in flight. Raised as
a legitimate blocker at merge time by the author themselves; I judged it non-blocking because the
gate's own failure is loud and lane-visible, and because holding a green, reviewed PR for coverage of
its coverage is the wrong trade at this point in the push. Recording that judgement here so it can be
overturned by whoever picks this up.

VijitSingh97:
Closes #1812

### DIFF
--- a/tests/integration/selftest-inventory-floors.sh
+++ b/tests/integration/selftest-inventory-floors.sh
@@ -1,0 +1,224 @@
+#!/usr/bin/env bash
+#
+# Self-test for tests/inventory.sh's aggregate FLOOR gate (#1812). #1808 gave the eight aggregate
+# counters real floors instead of a zero test; the evidence that a floor FIRES was a mutation
+# battery run once against that PR, and nothing re-ran it. A passing gate never demonstrates the
+# one property worth pinning here — that it can still fail — so this file drives the gate and
+# asserts BOTH answers.
+#
+# The lever: tests/inventory.sh resolves its own ROOT from ${BASH_SOURCE[0]}, not from the caller's
+# cwd, so a COPY of it inside a sandbox counts that sandbox's files. The floors are read out of the
+# gate rather than restated here, so a floor that moves is still the floor under test and this file
+# never has to be bumped alongside it; a counter the gate declares and this file has no fixture for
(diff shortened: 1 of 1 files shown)


VijitSingh97:

PR #2024, opened 2026-09-09
test(wizard): #1318 cover the form fallback's optional switches

### LINKED ISSUE #1318: A run-from-USB rig can never reach the wizard again — the bootloader is the only interface left, so give it a short menu: saved rig config, or setup

VijitSingh97:
Closes #1318

### DIFF
--- a/dashboard/mining_dashboard/web/static/savedrole.mjs
+++ b/dashboard/mining_dashboard/web/static/savedrole.mjs
@@ -1,0 +1,73 @@
+// The screen a set-up-again boot opens on (#1318), kept out of wizard.mjs because that file is
+// at its budget ceiling and because the decision this makes is worth testing without a DOM.
+//
+// A boot with `pithead.setup=1` publishes `saved-role.json`, and the host's having named a role
+// is the whole signal: the operator is offered the machine as it stands, and only reaches the
+// setup form by asking for it. Keeping is the default because it changes nothing — the page
+// writes one empty spool file and the host carries on booting with the configuration it has.
+
+import { html } from "./preact.mjs";
+import { Err, Field, Note } from "./wizardparts.mjs";
+
+// What each role is, in the operator's terms rather than the config's. A role NOT in this map
+// is one the page cannot name, and a screen offering to keep "your saved configuration" without
+// saying what it is asks the operator to confirm something they cannot see — so an unknown role
+// falls through to the normal wizard instead.
+const ROLES = {
+  rig: "a rig — it mines toward a Pithead, which manages it from there",
+  pithead: "a Pithead — it runs the node, the pool and this dashboard",
+  both: "a Pithead with its own miner",
+};
+
+/**
+ * What the screen says about the saved role, or null when there is nothing nameable to keep —
+ * no file, a file the server could not use, or a role this page has no words for. Rows follow
+ * the handoff card's rule: a value the host left empty drops its own row rather than rendering
+ * a blank beside a label.
+ */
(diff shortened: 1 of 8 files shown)


VijitSingh97:

PR #2025, opened 2026-09-09
fix(wizard): #1835 the restore door restates the stick marker too

### LINKED ISSUE #1835: Wizard: a rig that runs from the USB stick is shown 'Installing… Copying the system to the disk…' — status() takes the installer branch and never reads the in-place provision

VijitSingh97:
Closes #1835

### DIFF
--- a/dashboard/mining_dashboard/wizard.py
+++ b/dashboard/mining_dashboard/wizard.py
@@ -464,16 +464,16 @@ async def submit(request: web.Request) -> web.Response:
 
 async def submit_restore(request: web.Request) -> web.Response:
     """Restore-at-setup (#909, #786 sub-issue B): an uploaded encrypted backup archive + its
-    emergency-kit passphrase, in place of the config form. This server only asks — the archive
-    and passphrase cross the SAME spool the rest of the wizard uses, and the HOST decrypts,
-    validates and extracts (firstboot_consume_restore, reusing stack_restore's own machinery).
-    A rejected archive falls back to the form with the reason, exactly like a rejected config;
-    the passphrase is written once and the host deletes it immediately either way."""
+    emergency-kit passphrase, in place of the config form. This server only asks — both cross the
+    SAME spool the rest of the wizard uses, and the HOST decrypts, validates and extracts
+    (firstboot_consume_restore). The passphrase is written once, deleted immediately either way."""
     if not _authed(request):
         raise web.HTTPFound("/")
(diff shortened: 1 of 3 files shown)


VijitSingh97:

PR #2026, opened 2026-09-09
fix(installer): #1846 the reinstall pre-fill drops the login but keeps the switches that need it

### LINKED ISSUE #1846: Wizard: 'Generate a strong password for me' (the recommended default) is refused on Pithead + RigForge — 'dashboard.control.enabled is true but dashboard.auth.password is empty'; the KVM battery never exercises the browser auth_mode=auto path

VijitSingh97:
Closes #1846

### DIFF
--- a/docs/appliance.md
+++ b/docs/appliance.md
@@ -195,8 +195,11 @@ On either fresh start, the page opens with the previous install's answers alread
 wallet addresses, node locations, pool tier — read from the disk you are about to wipe, so you
 only change what you came to change. Its secrets never make the trip: the dashboard login,
 node passwords, view keys, worker tokens and alert credentials are left out, and anything the
-machine generates for itself is generated anew. If the old configuration cannot be read, the
-page opens blank.
+machine generates for itself is generated anew. The two switches that need that login come
+back at their defaults for the same reason: config editing from the dashboard, which the
+machine turns on again for you once a password exists, and publishing the dashboard as a Tor
+onion, which you tick again if you want it. If the old configuration cannot be read, the page
+opens blank.
 
 Type the disk's name to confirm.
 
--- a/lib/pithead/10-installer-preseed.sh
+++ b/lib/pithead/10-installer-preseed.sh
@@ -178,12 +178,23 @@ publish_disk_inventory() { # <spool-dir>
 # the fixed paths cannot reach), alert credentials, the ssh key. Wallet addresses, node modes
 # and hosts, the pool tier stay: those are the answers the operator came back for, and none of
 # them is a secret. Errs toward stripping more — a lost convenience beats a leaked credential.
+#
+# The two ENABLEMENTS go with the login, and that is the half this was missing (#1846). The
+# validator fails closed on dashboard.control.enabled and on dashboard.onion.enabled whenever
+# the password is empty — deliberately, since both are reachable control surfaces. Deleting the
+# login while leaving either switch on therefore publishes a pre-fill that CANNOT be submitted:
+# the page offers it back, "Generate a strong password for me" leaves the password empty because
+# the HOST generates one only AFTER the candidate has validated, and the first validation refuses
+# the operator's own answers. Dropping control.enabled costs nothing — apply_appliance_defaults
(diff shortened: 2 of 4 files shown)


VijitSingh97:

PR #2027, opened 2026-09-09
fix(wizard): #1868 cover the set-up-again screen's button, and widen the sweep that missed it

### LINKED ISSUE #1868: Wizard: buttons are browser defaults (Apply fails WCAG 2.5.8 target size), the shell's 'Loading…' placeholder survives every stage, and section headings jam against the field above

VijitSingh97:
Closes #1868

### DIFF
--- a/dashboard/mining_dashboard/web/static/savedrole.mjs
+++ b/dashboard/mining_dashboard/web/static/savedrole.mjs
@@ -47,7 +47,7 @@ export const SavedRoleScreen = ({ summary, kept, error, onKeep, onSetUpAgain }) 
                 html`<${Field} label=${r.label}><code class="wizard-mono">${r.value}</code><//>`,
             )}
             <${Err}>${error}<//>
-            <button type="button" onClick=${onKeep}>Keep it</button>
+            <button type="button" class="btn-toggle active" onClick=${onKeep}>Keep it</button>
             <button type="button" class="wizard-link" onClick=${onSetUpAgain}>Set up again</button>
             <${Note}>Setting it up again asks the same questions as a first boot, with this
             machine's answers already filled in. Its login and other secrets are never filled
--- a/dashboard/mining_dashboard/web/static/wizard.css
+++ b/dashboard/mining_dashboard/web/static/wizard.css
@@ -42,3 +42,33 @@
   text-decoration: underline;
   cursor: pointer;
 }
+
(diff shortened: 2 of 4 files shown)


VijitSingh97:

PR #2029, opened 2026-09-09
fix(control): #1888 the confirm prompt drops an issue reference operators cannot use

### LINKED ISSUE #1888: Remote node endpoints must be changeable from a running dashboard — today the control channel refuses TARI_GRPC_ADDRESS and points at a shell the appliance does not have

VijitSingh97:
Closes #1888


VijitSingh97:

PR #2030, opened 2026-09-09
test(stack): #1771 the TTL-window row waits on the write, not on a fixed budget

### LINKED ISSUE #1771: test-control-backup.sh: the TTL-window row is a fixed sleep racing a background write

VijitSingh97:
Closes #1771

### DIFF
--- a/tests/stack/test-control-backup.sh
+++ b/tests/stack/test-control-backup.sh
@@ -104,11 +104,20 @@ export CONTROL_BACKUP_KIT_TTL_S=3
 printf '{"id":"%s","action":"backup","actor":"admin"}\n' "$bid2" >"$BKC/req2.json"
 run_sourced "$SANDBOX" control_process_request "$BKC/req2.json" "$BKC" >/dev/null 2>&1 &
 bg_pid=$!
-sleep 0.5 # well inside the 3s TTL — the stubbed child + write are effectively instant
-mid_pass="$(jq -r '.passphrase // "null"' "$BKC/results/$bid2.json" 2>/dev/null)"
-{ [ -n "$mid_pass" ] && [ "$mid_pass" != "null" ]; } &&
+# The mid-flight read races the background write, so it waits on the condition rather than on a
+# fixed budget; why a tick budget is wrong is #1495's lesson and lives on wait_while_alive itself.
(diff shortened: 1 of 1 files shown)


VijitSingh97:

PR #2031, opened 2026-09-09
Contextualize diagnostics transport errors

### LINKED ISSUE #1961: Service diagnostics: the log verb times out with the OS-update flow's error text; run doctor once and render every service at a glance instead of per-service fetches
## Operator (RC1 manual test)

Service diagnostics → "Show recent log" for tor sat on "Asking the host…" and then failed with: "Error: Stopped waiting — this can take longer than expected on a slow connection. The host keeps going and finishes on its own; reload in a few minutes to see the result. If the version is unchanged after that, check that dashboard.control is enabled and the pithead-control unit is running." That text is the OS-update flow's, not this card's: nothing here has a version.

Operator's design ask: "Not sure we need this fine grain for each service — just run doctor and show the output in a nice GUI for all the services at once; we can show the other stats from doctor also."

## What should happen

- One action, "Run health check", runs `doctor` once and renders it per service (status, the check that failed, the remedy) plus the machine stats doctor already prints; the per-service log tail becomes a disclosure under each service, fetched on demand.
- Timeouts and failures say what happened in this card's own words, and a runner that is wedged (the same box had failed applies pending — #1931) is reported as such, not as a slow connection.

Tier 1: the card's state machine with a slow, a failed and a wedged runner; tier 3: the diagnostics verbs through the control runner on the appliance (they are read-only, so they belong in the KVM battery).

VijitSingh97:
Closes #1961

### DIFF
--- a/dashboard/mining_dashboard/web/static/diagview.mjs
+++ b/dashboard/mining_dashboard/web/static/diagview.mjs
@@ -72,14 +72,28 @@ export class DiagnosticWaitTimeout extends Error {
 // result exists yet. Keep that uncertainty instead of borrowing the upgrade flow's version copy or
 // calling the runner wedged from elapsed time alone.
 export async function runDiag(path, body, label = "the diagnostics request", max = DIAG_POLL_MAX) {
-  const res = await fetch(`/api/control/${path}`, {
-    method: "POST",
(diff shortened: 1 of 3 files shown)


VijitSingh97:

PR #2039, opened 2026-09-09
fix(inventory): #1420 the aggregate counters get floors, not a zero test

### LINKED ISSUE #1420: `tests/inventory.sh`'s emptiness guard failed open on a live 24-of-25 breakage, and the file already documents the same hazard for its sibling

VijitSingh97:
Closes #1420

### DIFF
--- a/docs/dev/testing-strategy.md
+++ b/docs/dev/testing-strategy.md
@@ -300,7 +300,8 @@ Every scenario, at every tier, holds to the same rules.
   hashing on the box; all artifacts pass a redactor.
 - Reproducible. The live run records a manifest (stack `VERSION`, git rev, image digests).
 - Test code is real code. The same lint (shellcheck) and coverage gate apply to the tests
-  themselves, and the inventory generator fails CI if it stops enumerating a suite.
+  themselves, and the inventory generator fails CI if a suite stops enumerating or shrinks past
+  its floor.
 - An assertion reads its haystack directly, never through a pipe. `grep -q` exits at its first
   match, so under `pipefail` the writer's broken pipe becomes the pipeline's exit status and the
   match is thrown away: a directive that was present read as missing, and — in the direction
--- a/tests/inventory.sh
+++ b/tests/inventory.sh
@@ -69,15 +69,16 @@ n_mini=$(grep -cE 'log "scenario [0-9]' tests/integration/mini-stack/run-mini-st
 total=$((n_py_dash + n_py_fake + n_node + n_stack + n_selftest + n_scen + n_mini))
 
 # --- drift gate -----------------------------------------------------------
-# The gathering above is grep-based, so a suite that moves, renames, or changes shape
-# enumerates as zero without erroring. Fail loudly instead of emitting an inventory that
-# silently under-counts — CI runs this on every PR (#981).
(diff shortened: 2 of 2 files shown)


VijitSingh97:

Issue #1802, opened 2026-09-05
CI: the image builds have no apt retry, so a mirror mid-sync reds an unrelated PR


VijitSingh97:
The three image builds in `build/` run a bare `apt-get update` with no retry. When the Ubuntu
archive is mid-sync, the index it serves does not match the size its `Release` file declares,
apt exits 100, and the whole job goes red for a reason that has nothing to do with the tree.

## Measured

Two unrelated pull requests hit it within minutes of each other on 2026-09-05, off different
mirror IPs, with the same byte mismatch:

| PR | run | job | mirror IP |
|---|---|---|---|
| #1795 (a `ci.yml`-only diff) | `33936667749` | `Build image (monero)` | `91.189.92.24` |
| #1785 (a `tests/stack`-only diff) | `33936808800` | `Build image (monero)` | `91.189.91.83` |

Neither diff touches `build/`. Both logs carry the same line:

```
E: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/noble-updates/main/binary-amd64/Packages.gz
   File has unexpected size (1561627 != 1561628). Mirror sync in progress?
E: Some index files failed to download. They have been ignored, or old ones used instead.
ERROR: process "/bin/sh -c apt-get update && apt-get install -y ..." did not complete successfully: exit code: 100
```

**Base rate, so the frequency is not overstated.** Across the preceding 80 `ci.yml` runs
(2026-09-04T00:02:35Z to 2026-09-05T02:02:45Z) there were five failures and **not one of them was a
`Build image` job** — four `Shell tests`, one `Lint per-surface`. This is a correlated burst inside
one upstream sync window, not a chronic rate. That is the argument *for* retry logic rather than
against it: during such a window it reds every open PR at once, and each red costs a re-run plus, on
a PR that already carries a recorded review pass, a second review.

## Where it is

Three sites, each a bare `apt-get update &&` with no retry configured anywhere under `build/`:

- `build/monero/Dockerfile:6`
- `build/p2pool/Dockerfile:11`
- `build/xmrig-proxy/Dockerfile:10`

## Why this is worth fixing rather than re-running

This is the third distinct way this repo's CI reports something other than the state of the tree.
The other two were found the same night and are fixed by PR #1795: a `Shell tests` job that exceeded
its cap reported as `cancelled` rather than failed (#1789), and a `curl ... | sh` installer step that
reported success on a failed download (#1788). A reviewer reading a check rollup is misled by all
[...]

PR #2054, opened 2026-09-11
fix(os): the three KVM reds from #2002 — boot-menu labels, the wizard's dead page, and the restore leg's missing evidence

### LINKED ISSUE #1793: tests/os: the media-abort red names a console file the empty-console case never creates
Carried out of #1790, which merged at `418d50f7` with this noted and deliberately left to ride the
next touch of the leg. Filing it so that decision survives the PR being closed.

## The finding

`tests/os/run.sh:2720-2721`, on `develop-v2` at `418d50f7`:

```sh
[ -f "$SERIAL.failed" ] || [ ! -s "$SERIAL" ] || cp "$SERIAL" "$SERIAL.failed" 2>/dev/null
bad "... the channel's last line: '$(...)'; console kept at $SERIAL.failed"
```

When the console is empty or absent, `[ ! -s "$SERIAL" ]` is true, the `||` chain short-circuits, and
the `cp` never runs — which is correct, and is exactly what #1790's finding A set out to fix. But the
`bad` string on the next line carries no matching condition, so the one line this leg exists to
produce ends `console kept at <path>.failed` naming a file that is not there.

Before #1790 the sentence was thin but true: a zero-byte `.failed` did exist. After it, the sentence
is false in precisely the case where the reader has least other evidence to work from. `cleanup()`
(`:403`) gates on the same `-s` and will not keep it either, and `:414` deletes `$SERIAL` on the way
out, so there is genuinely nothing to keep. **The defect is the sentence, not the guard** — the guard
is right and should not move.

## Scope

Narrow. Reaching it needs a console with zero bytes after 180 s + 90 s of waiting, i.e. a guest that
produced no output at all. It costs a reader one wrong turn, not a wrong verdict: the row still goes
red, and `FAIL` is still set.

## Suggested shape, not a prescription

Make the tail of the message conditional on the same test the guard uses, so the red either names a
file that exists or says the console was empty. The file has zero line-budget headroom at 3423, so
this wants to ride a change that is already touching the block rather than a commit of its own.

## PROVEN vs ASSUMED

**Proven by reading at source**, `git show origin/develop-v2:tests/os/run.sh` at `418d50f7`: the two
lines above are adjacent, the guard short-circuits on `[ ! -s "$SERIAL" ]`, and the `bad` string is
unconditional. The control flow is not ambiguous.

**Not run by me.** I drove nothing here — no KVM, no guest, no bench, no `make` target. The empty and
absent console cases were each driven, with output, by two independent non-author passes on #1790
(one ephemeral reviewer, one merger's due diligence) which converged on this. I confirmed their
conclusion against the merged file rather than restating it on their word.

### LINKED ISSUE #1956: Boot menu: name the entries by the Pithead version each slot holds and mark the one that boots (current), not 'fallback / Slot A / Slot B'

### LINKED ISSUE #2050: Appliance: a faulted setup hangs instead of reopening the wizard (#1955's dead page, measured)
## What

When `setup` fails after the configuration has already been validated and accepted, the first-boot
wizard **hangs**. It neither reports the failure nor reopens the setup page, so the operator is left
on a dead page with no error and no way forward — the symptom #1955 describes, now measured.

## Measured

KVM battery, `--phase provision`, 2026-09-10, on an image built from pithead#2002's head.

Image refs were explicitly **not** a factor: the image was built with `PITHEAD_REGISTRY` pointing at
a registry holding all five first-party services at the tag the appliance asks for, both
provisioning units carried the registry drop-in, and the zero-container dump's discriminator
(`comm -23` of what compose wants against what podman has) came back **empty**. So this is not
#2043's cause.

The phase arms the fault by moving `/data/pithead/docker-compose.yml` aside after validation, then
submits a valid configuration. Serial console:

```
[   43.540732] [pithead] Configuration accepted — provisioning now.
[   43.663301] [pithead] Dashboard login for this machine:
[   43.663959] [pithead]     user: admin
[   44.394833] [WARNING] The machine's own defaults collide with this configuration: [pithead] Parsing configuration...
[   44.396898] grep: docker-compose.yml: No such file or directory
```

That is the **last output the guest ever produced.** Nothing further appeared for the remaining two
minutes until the harness gave up and destroyed it. `podman ps -a` at that moment showed only
`pithead-wizard`, "Up 2 minutes" — the wizard container alive, the process inside it silent. The
wizard spool's `error.txt` contained only `[pithead] Parsing configuration...`.

The failing assertion is `faulted setup never reached its credentials handoff`: the harness polls
`/api/handoff` for a password 24 times at 5s intervals and times out.

## Why it matters

`lib/pithead/12-firstboot-wizard.sh` is written to handle exactly this: `setup`'s status is captured
rather than collapsed (`{ (setup) 2>&1 | tee "$setup_log"; } || setup_rc=$?`), and a non-zero status
is meant to reach `wizard_setup_failed` and `prepare_wizard_spool` so the page reopens with the
answers retained for correction. None of that is reached, so the designed recovery path is dead code
on this input.

This is the one failure of #2043's five that is genuinely a product defect. It is also
version-independent, which is why #2043's `develop` control reproduced it — that control was
measuring this, not the registry cause, and #2043's own "one leg was controlled, not five" limit is
the reason the distinction was available to be missed.

## Reproduce

Needs the bench (KVM + libvirt + root) and an image built against a registry that serves the five
first-party images at `v$(cat VERSION)` — otherwise #2043's cause masks this one:

```bash
sudo cp /root/.ssh/pithead-os-test.pub /tmp/pithead-os-test.pub
PITHEAD_REGISTRY=<host:port> PITHEAD_REGISTRY_CA=<ca.crt> os/build-image.sh --ssh /tmp/pithead-os-test.pub
os/rauc/mkimage.sh --dev
sudo env PITHEAD_REGISTRY=<host:port> PITHEAD_REGISTRY_CA=<ca.crt> \
    tests/os/run.sh --image os/rauc/build/system.img --phase provision
```

Expect `os harness: 5 passed, 1 failed`, the sixth assertion being the handoff.

## First step

The guest is destroyed per phase, so run with `--keep` and attach to the wizard process to see where
it is blocked. The `[WARNING] ... defaults collide` line interleaving with `Parsing configuration...`
suggests output from a path that also prompts; a `read` on a closed stdin, or a lock acquired and not
released, would both produce this signature. Worth checking whether any code reachable from a failed
`setup` prompts interactively — `prompt_start_stack` does (`read -r -p "Start Pithead now? (Y/n): "`).

### LINKED ISSUE #2051: Appliance restore: a restored machine never starts its stack (#1091 verdict, measured)
## What

A machine restored from an encrypted backup gets its configuration and payout wallet back, and then
**never starts the stack**. `podman ps` is empty after a 900-second wait.

This is the #1091 closing verdict firing for the reason it was written: "config.json on disk is not
proof the machine is RUNNING what was restored."

## Measured

KVM battery `--phase all`, 2026-09-10, image stamped `769aa29ab8cbb8e0ad6c3881eb7abb7c08a0c339`.
Whole-run result: 189 passed, 3 failed. The restore leg reaches its final assertion, having passed
every step before it:

```
✓ restore leg: keep-reinstalled machine provisioned — a live stack to back up (tor dashboard docker-control docker-proxy caddy )
✓ restore leg: provisioning finished (units: inactive inactive )
✓ restore leg: took a real encrypted backup off the live machine
✓ restore leg: uploaded the backup archive instead of the form
✓ restore leg: the restored config drove provisioning to a credentials card
✓ restore leg: installed and switched itself off
✓ restore leg: the restored machine boots from the fresh disk
✓ restore leg: the carried archive provisioned the machine — config.json is back
✓ restore leg: restored machine carries the ORIGINAL wallet address, not a fresh one
✗ restore leg: the stack never came up on the restored machine (podman ps: 'none')
```

So: the archive is produced, uploaded, unpacked, and its contents are demonstrably in effect — the
restored machine carries the **original** wallet, not a freshly generated one, which is the property
that distinguishes a real restore from a re-provision. What does not happen is the stack starting.

## This is not #2043

Worth stating explicitly, because the symptom string is the same as four legs that #2043 covered and
those four now pass. This run was made with the registry override in place:

- `leg 4: stack provisioned; dashboard login captured` — passes.
- The media and reset legs' "stack never came up" assertions do not fire at all.
- `restore leg: stack never came up after provisioning` — the EARLIER assertion in this same leg —
  now passes, naming five running containers. The leg progressed past where it used to stop, which
  is how this assertion became reachable.

So image references are not the cause here. Something specific to a **restored** machine's first
boot is.

## Evidence not yet captured

`tests/os/zero-container-evidence.sh` was wired into five of the seven sites that assert on a stack
coming up, and this was not one of them — so this run has no image-list/journal dump for it. Both
remaining sites are now wired (the dump is attached at the call site rather than inside
`restore_live_state_verdict`, whose stdout is its verdict message). **The next battery run will
capture the cause automatically**: whether compose resolved refs the restored machine could not pull,
or every ref was present and something else refused.

Until then the useful first step is a `--keep` run of `--phase install` and, on the restored guest:

```
systemctl status pithead-boot.service
journalctl -u pithead-boot -u pithead-firstboot --no-pager -n 80
cd /data/pithead && docker compose config --images
podman images --format '{{.Repository}}:{{.Tag}}'
```

`provisioning finished (units: inactive inactive )` on the SOURCE machine is worth a look too — both
provisioning units read inactive there, which is the expected terminal state, but the restored
machine's equivalent was never asserted.

## Why it matters

Backup/restore is the operator's recovery path, and this is the half that makes a restore real. A
restore that returns the configuration but not a running stack looks successful to anyone checking
`config.json` and is not. #1091 exists because an earlier version of this leg passed on exactly that
weaker evidence.

VijitSingh97:
## Summary

The three reds from #2002's `--phase all` run, all three fixed and confirmed on the KVM bench.
Two of them had the wrong cause recorded against them, and clearing the third unblocked a phase
that has been aborting after 6 of its 46 assertions in every battery ever run.

**Measured, `--phase all` at `0ca255e8` (2026-09-11, 96 min): 236 passed, 8 failed.**
The run before this branch: 189 passed, 3 failed.

Both numbers are needed or neither makes sense. The pass count rose by 47 because the provision
phase stopped aborting; the 8 failures are what the newly reachable rows found, and they are
pre-existing (see *Not regressions* below).

## #1956 — the boot repair never ran

The slot attribution was never wrong. `B_VERSION` was never *written*, so the committed slot's
entry read `empty (slot B, current)` while slot A's correctly read `previous` in the same menu.

grubenv has two writers: `pithead os-update` records the target slot at install, and every boot
repairs its own. The second rode `pithead-boot.service`, which is gated on a **provisioned**
machine — and `tests/os/phases/update.sh:162` states plainly that legs 1–3 never write
`config.json` or `machine-role`. So the unit never started, and the battery installs with plain
`rauc install`, so the other writer never ran either.

Reproduced at tier 1 against the real `grub.cfg` block before fixing anything: the `ORDER="B A"`,
`B_VERSION=` fixture prints exactly the pair the bench saw. The repair now has its own unit with no
provisioned-only condition — an appliance that has not been through the wizard is precisely where
a person is standing at a console.

Bench: `✓ serial menu names Pithead 2.0.0 as slot B current and slot A previous`, two runs.

## #2050 — two findings, and the product was right

**The leg armed the wrong seam.** Every validation reaches `caddy_hash_password_b64`, which greps
the pinned caddy ref out of `docker-compose.yml`, so removing that file faulted
`parse_and_validate_config` — a path with its own recovery that correctly publishes no credentials
handoff. The harness polled 24×5s for one and reported a hang; the guest was sitting on a reopened
form. The issue's own evidence settles it: `error.txt` held raw validator output with **no
`[ERROR]` prefix**, and only the validator branch writes that.

Re-armed with a Compose stub that keeps the caddy line and is unparseable after it — verified both
ways: the validator's exact grep resolves the identical ref, and `docker compose config` rejects it
rc 1.

**Then the leg's own verdict was wrong.** With the fault finally reaching `setup`, the row
`host setup failure did not reopen a useful retained form` failed. Read off a kept guest, the page
was serving exactly the right thing:

```
stage = 'failed'    error = '[ERROR] Timed out waiting for the P2Pool Tor hidden-service hostname.'
monero.wallet_address ✓    tari.mode = 'local' ✓
```

The verdict demanded `stage == "setup"` — the *preflight refusal's* stage. The dashboard pins
`"failed"` in its own tier 1 (`test_wizard_recovery.py`, twice), so this was checkable without a
guest all along. It survived because the self-test fixture was written from the assertion's own
assumption rather than from a real page, so the control was green by construction.

Mutation-checked after fixing, which caught a second thing: `.config.tari.mode == "local"` had no
negative case at all — replacing it with `true` left the self-test green. All three clauses now
kill their mutation.

**Also fixed, separately:** `setup()` writes `DEPLOYMENT_COMPLETED=true` a third of the way in, so
any failure past that point left the marker on a machine that is not deployed and every wizard
retry refused as already provisioned. Cleared in `wizard_keep_failed_config`, proven as a
controlled pair against the real generated CLI. **This leg does not exercise that fix** — the stub
faults at `provision_tor`, before the marker is set — so it stands on its tier-1 pair only.

## #2051 — the restore was failing, and the leg passed on the wreck

From the restored guest's own journal:

```
mv: cannot move '…/tree/data/pithead/data/tor/' to '/data/pithead/data/tor': No such file or directory
[WARNING] The carried restore archive was rejected — falling back to the setup page.
```

`restore_apply` `mv -T`s each tree item into place with nothing having created the parent —
`prepare_directories` runs inside `setup()`, which both restore doors call **after** the restore.
On a fresh disk the first tree item fails ENOENT.

The failed move is not the damage. `config.json`, `.env` and `Caddyfile` are applied *before* the
tree items, so the early return leaves a **partially restored machine**: the caller reads that as a
valid pre-seed and `setup` refuses as already provisioned, because the carried
`DEPLOYMENT_COMPLETED` clear sits past the failure. Zero containers.

That also explains why `config.json is back` and `carries the ORIGINAL wallet` were passing on a
machine whose restore had **failed** — both are true of a partial application. #1091's warning,
harder than it was written.

Tier 1 never caught it because `test-wizard-setup.sh:224` pre-creates the destinations
(`mkdir -p "$RT/data/tor"`), removing the one condition the real path has. The new rows delete the
data tree and assert it is gone before driving anything.

Fix is one `mkdir -p`. Bench: install phase **47 passed, 0 failed**, twice, including
`LIVE state carries the restored wallet` and `kept the ORIGINAL Tor identity`.

## Not regressions — the 8 that remain

Controlled: a run with the setup-failure leg **skipped entirely**, so no failed `setup` anywhere in
the guest's history, produces the identical set. Filed as **#2059** (the Tor-only egress firewall
is fail-open — a security property, three independent reproductions) and **#2060** (the other
seven, of which three are one mDNS defect reported at three sites).

They were unreachable because the provision phase has been returning after 6 of 46 assertions
since it was written. The structural half of that is #2055.

## Also

- #1793: the media abort's red named a console file the empty-console case never creates.
- The validator's refusal no longer reports itself as "the machine's own defaults collide with this
  configuration". A missing Compose file is not a collision, and reading it as one cost a bench round.
- `os/overlay/pithead-boot-version` joins `lint-sh`, which never covered it.
- The setup-failure leg moved out of `provision-browser-submit.sh` (first commit, a pure move —
  byte-identical, verified by diff).

## Verification

- CI **22/22 green**. One earlier red was `Build image (monero)` on `Unable to connect to
  archive.ubuntu.com:80` — an apt mirror outage on a branch touching no `build/` files (#1802);
  green on rerun.
- The four new tier-1 #2051 rows ran and passed on Linux CI, arming control included.
- `verify-image` against the built image: **124 passed, 0 failed, 0 skipped** with
  `PITHEAD_EXPECT_COMMIT` pinned — rows CI does not run.
- The baked CLI inside the image the battery booted was mounted and grepped for the fix rather than
  assumed present.
- No file-budget ceiling raised.
- Bench stack stopped for the run and restored; no surviving guests, no leaked serial watchers.

Closes #1956.
Closes #2050.
Closes #2051.
Closes #1793.


VijitSingh97:
## Adversarial review pass — two findings taken, three rejected with a measurement

Ran an independent correctness review and a security review against this branch in fresh contexts.

### Taken

**The arm did not restore if it died mid-way.** `provision_setup_failure_recovery` moves the real
Compose file *before* it writes and checks the stub, and this branch widened that window from two
`test`s to a file write plus two greps. A failure in between left the guest with no usable Compose
file and no caller to put it back — every other failure branch in the leg restores; only the arm's
own did not. Fixed, and the restore fires only over an absent file or our own stub, never over a
file we did not replace.

**The restore leg could spend 900 s twice.** The new provisioning-settled row waits up to 900 s and
the container wait that follows waits another 900 with no early exit — half an hour to fail one
leg. A machine still activating after 900 s is not about to start containers, so that budget drops
to zero once the row above has said why. The loop now reads **once** regardless of the budget: a
head-tested loop at zero would never run and the verdict would have reported
`podman ps: 'none'` for a machine nobody asked.

### Rejected, with the measurement

Three findings claimed that `$SETUP_FAULT_MARK` is not expanded because it sits inside single
quotes within the `_ssh` string, and that the arm therefore writes a literal `# $SETUP_FAULT_MARK`
that the restore's unquoted grep cannot match.

The string is **double-quoted at the bash level**, so those single quotes are literal characters
the *remote* shell quotes with, and the variable is expanded locally before `_ssh` ever sees it.
Printed what the remote shell actually receives:

```
… { echo '# PITHEAD_OS_2050_STUB'; grep -oE 'caddy:…' '/run/…' | head -1 | sed 's/^/# /'; … } \
  >'/data/pithead/docker-compose.yml' && grep -q 'PITHEAD_OS_2050_STUB' '/data/pithead/docker-compose.yml'
```

Same needle at all three sites. No change made.

### Security review

No exploitable change. The `DEPLOYMENT_COMPLETED` clear only runs from `firstboot_wizard`'s retry
loop, which itself only runs on a genuinely un-provisioned machine or behind the physical-presence
"Set up again" GRUB entry that already bypasses `is_deployed` by design (#1318) — so it does not
reopen an unattended re-provision on a live box. `config.json.failed` keeps mode 600 and its
ordering. `pithead-boot-version`'s inputs are anchored (`^rauc\.slot=\([AB]\)$` plus a `case A|B`
re-check, and a semver regex) before reaching `grub-editenv`'s argv, so running it in more contexts
does not widen what it can be made to do. The new dump probes are a fixed-key `.env` whitelist and
unit metadata only — no credential reaches the log.

### One thing the review did not catch, found while acting on it

Moving the repair into its own unit made its failure a **FAILED unit**, and
`tests/os/phases/provision-reboot.sh:61` and `rig.sh:166` both assert `systemctl --failed` is empty
(#792). An image built with `PITHEAD_UPDATER=""` has no `rauc.slot=` to read, so `record_booted`
returns 1 and would have reddened two legs on a machine working exactly as designed. The dispatch
now tolerates it with the reason in the journal, pinned at tier 1 with the recording case as the
control.

VijitSingh97:
## Measured on the KVM bench: 236 passed, 8 failed

`--phase all` at `0ca255e8`, 2026-09-11, 96 minutes, image `cf21b519…` built with the bench
registry override. The run before this branch, at #2002's head, was **189 passed / 3 failed**.

**Both numbers are needed or neither makes sense.** All three failures this PR set out to fix are
fixed. The pass count rose by 47 because the provision phase stopped aborting after 6 of its 46
assertions — which it had done in every battery ever run. The 8 failures are what the newly
reachable rows found, and they are pre-existing.

### The three, cleared

| | evidence |
|---|---|
| **#1956** | `✓ serial menu names Pithead 2.0.0 as slot B current and slot A previous` — two independent runs |
| **#2050** | leg green end to end. The product's recovery was correct all along; the leg's own verdict demanded stage `setup` where the wizard correctly reports `failed` |
| **#2051** | install phase **47 passed / 0 failed**, including `LIVE state carries the restored wallet` and `kept the ORIGINAL Tor identity` — twice |

### #2051's cause, from the restored guest's own journal

```
mv: cannot move '…/tree/data/pithead/data/tor/' to '/data/pithead/data/tor': No such file or directory
[WARNING] The carried restore archive was rejected — falling back to the setup page.
```

`restore_apply` `mv -T`s each tree item into place with nothing having created the parent —
`prepare_directories` runs inside `setup()`, which both restore doors call AFTER the restore. On a
fresh disk the first tree item fails ENOENT, and because `config.json`/`.env` are applied BEFORE
the tree items the early return leaves a **partially restored machine**: the caller reads that as a
valid pre-seed and `setup` refuses as already provisioned, because the carried
`DEPLOYMENT_COMPLETED` clear sits past the failure.

That also explains why the leg's earlier rows passed on a machine whose restore had FAILED —
`config.json is back` and `carries the ORIGINAL wallet` are both true of a partial application.
#1091's warning, harder than it was written.

Tier 1 never caught it because the #1239 fixture pre-creates the destinations
(`mkdir -p "$RT/data/tor"`), removing the one condition the real path has. The new rows delete the
data tree and assert it is gone before driving anything.

### The 8 that remain are NOT regressions

Controlled: a run with the setup-failure leg **skipped entirely** — no failed `setup` anywhere in
the guest's history — produces the identical set. Filed as **#2059** (the Tor-only egress firewall
is fail-open; three independent reproductions) and **#2060** (the other seven, of which three are
one mDNS defect reported at three sites).

### Gates

- CI 22/22 green. The one earlier red was `Build image (monero)` on
  `Unable to connect to archive.ubuntu.com:80` — an apt mirror outage on a branch that touches no
  `build/` files (#1802); it passed on rerun.
- The four new tier-1 #2051 rows **ran and passed on Linux CI**, arming control included.
- `verify-image` against the built image: **124 passed, 0 failed, 0 skipped** with
  `PITHEAD_EXPECT_COMMIT` pinned — the rows CI does not run.
- The baked CLI inside the image the battery booted was mounted and grepped for the fix rather
  than assumed present.
- Bench stack stopped for the run and restored afterwards; no surviving guests, no leaked serial
  watchers.

Issue #2060, opened 2026-09-11
Seven RC2 provision-phase rows fail on their first ever execution (hostname mDNS x3, control runner x2, migration bundle)


VijitSingh97:
## What

pithead#2054 unblocks the provision phase, which has aborted at the setup-failure leg in every
battery ever run. Past that point the phase goes from 6 assertions to 46 — and **seven rows fail
on their first ever execution**. The security one is filed separately as #2059; this is the rest.

```
✗ post-provision benign setting did not land
✗ wizard hostname identity did not converge (mdns=10.89.0.1)
✗ host-mediated hostname approval did not bind prompt, approver, apply and live identity
✗ unaided reboot identity did not converge (mdns=172.28.0.1)
✗ doctor's nonzero health report was lost or flattened by the appliance control runner
✗ A/B update identity did not converge (mdns=172.28.0.1)
✗ the 99.0.0 migration bundle build failed (/tmp/os-fault-bundle.log)
```

## Measured, and controlled

`--phase provision`, 2026-09-11, image from pithead#2054's head with the bench registry override
(not #2043). Two independent runs. The second **skipped the setup-failure leg entirely**, so the
guest had no failed `setup` in its history — all seven reproduce identically. They are pre-existing
and were simply unreachable, not a side effect of pithead#2054's re-armed fault.

The run-to-run count difference (46 vs 42 passed) is exactly the four rows the skipped leg
contributes, consistent with nothing else changing.

## Three of these are one defect

`mdns=10.89.0.1` in one run, `mdns=172.28.0.1` in the next — both container-bridge addresses, not
the appliance's LAN address. The hostname-identity verdict (#1957) requires kernel hostname,
rendered `HOST_IP`, `/api/state`'s header source, the served certificate's SANs and **mDNS
resolution** to agree; mDNS is the one that disagrees, and it disagrees at all three sites the leg
checks (initial wizard, unaided reboot, post-A/B-update). The value moving between runs says this
is a resolution defect picking up whichever bridge exists, not a fixed wrong constant — so a fix
keyed to one address would be wrong.

Avahi is configured IPv4-only in the image (`os/rootfs/Dockerfile`, the `use-ipv6=no` sed), and
`verify-image` asserts that, so the question is which interface Avahi is publishing on rather than
which family.

## The control-runner pair

`post-provision benign setting did not land` and `doctor's nonzero health report was lost or
[...]

Issue #2059, opened 2026-09-11
Appliance: the Tor-only egress firewall is FAIL-OPEN — a chain container reaches clearnet directly (measured, controlled)


VijitSingh97:
## What

On the appliance, the Tor-only egress firewall is **not enforced**. From inside the `monerod`
container a direct clearnet dial by IP succeeds:

```
✗ clearnet egress is FAIL-OPEN — monerod reached 1.1.1.1 directly, bypassing Tor
  (the firewall is not enforced)
```

That is `tests/os/phases/provision-initial.sh:194` — `podman exec monerod curl -m 8 http://1.1.1.1/`
returning success where the #270/#855 property requires a DROP.

## Measured

KVM battery `--phase provision`, 2026-09-11, image built from pithead#2054's head with the bench
registry override (so this is not #2043). Reproduced twice, in two independent runs.

**Controlled.** The second run skipped the setup-failure leg entirely, so the guest had no failed
`setup` anywhere in its history — the row fails identically. It is not a side effect of a faulted
provisioning attempt.

## Why it was never seen before

The provision phase has aborted at the setup-failure leg in every battery to date — first at
`faulted setup never reached its credentials handoff`, then at `host setup failure did not reopen
a useful retained form`. Both `return 1` out of the phase, so every row after them, this one
included, has never executed. pithead#2054 clears both, which is how this surfaced.

## Scope — what is NOT the cause

Worth stating, because the obvious guesses are already handled in `lib/pithead/02-tor-egress.sh`:

- It is **not** a DOCKER-USER/podman mismatch. The file documents exactly that trap at `:14-18`
  and ships two backends: iptables/DOCKER-USER for the DIY Docker channel, and an independent
  `inet pithead_egress` nftables table hooked at forward priority -5 (ahead of netavark's
  priority-0 accept) for the appliance.
- It is **not** a missing tool. `nftables` is installed in the image (`os/rootfs/Dockerfile:145`).
- It is **not** an uncalled function. `apply_tor_egress_firewall` runs from `stack_up`
  (`lib/pithead/01-lifecycle.sh:129`), before containers start.

So both halves exist and the failure is at runtime on the appliance path.

## What the next run should capture

None of this was captured — the row reports the dial and nothing else. On a `--keep` guest:

```
nft list table inet pithead_egress          # does the table exist at all?
nft list ruleset | grep -A20 pithead_egress # hook, priority, and the allow-set
[...]

Issue #2055, opened 2026-09-11
The KVM battery's structural blind spots: an unprovisioned update guest, Secure Boot never on, a skipped unit that reads as a finished one


VijitSingh97:
## Why

#2050 and #1956 were both found by the battery and both had the wrong cause recorded against
them, and #1956 was a defect the battery could never have caught *before* the label assertion
existed — because the phase it lives in runs a guest where the unit that carries the fix never
starts. Filing the structural gaps that made those possible, separately from the defects
themselves (pithead#2054 fixes those).

Not a duplicate of #2001 (N-1 backup restore), #2044 (the manual hardware battery) or #1651
(the battery must be green on the shipping cut). Those are rows and runs; these are properties
of the harness.

## G1 — `--phase update` legs 1–3 run an UNPROVISIONED guest, and nothing says so per-assertion

`tests/os/phases/update.sh:162-163` states it outright:

> neither config.json nor machine-role ever gets written by legs 1-3 (they only drive rauc)

`pithead-boot.service` carries two triggering conditions, `ConditionPathExists=|config.json` and
`|machine-role`. On that guest neither holds, so **the unit never starts for the whole of legs
1–3** — and with it everything it does: the boot gate, the commit decision's health check, the
rollback verdict, the data-floor restore, the held-chain release, and (until pithead#2054) the
boot-menu version repair.

That is how #1956 shipped: the repair rode `pithead-boot`, the label assertion lives in leg 2's
post-commit reboot, and the phase was green for everything else. The assertion did eventually
fire — but only because someone wrote a label assertion, not because the harness could see that a
unit it depends on had not run.

The ask is not "provision the update guest" (legs 1–3 deliberately test the updater alone). It is:
**enumerate what legs 1–3 cannot see**, and for each assertion in them that implicitly depends on
`pithead-boot` having run, either move it to leg 4 (which does provision, via
`_wizard_provision_capture`) or state in the row that it is measuring the bootloader only.

## G2 — Secure Boot is explicitly DISABLED in every guest

Every `virt-install` in the harness passes:

```
--boot uefi,firmware.feature0.name=secure-boot,firmware.feature0.enabled=no
```

(`tests/os/phases/boot.sh:18`, `:128`, `tests/os/phases/install-reinstall.sh:23`,
`tests/os/reinstall-prefill-submit-leg.sh:29`.)

So the appliance's signed-boot chain — shim, GRUB, kernel, and whether an A/B update survives with
[...]

Issue #2041, opened 2026-09-10
Deprecate macOS as a development and test platform


VijitSingh97:
## What

Deprecate macOS as a development and test platform. Keep the Darwin code paths working, mark them deprecated, and make the shell suite refuse to run on macOS rather than emit failures that read as signal.

## Why

macOS is already unsupported as a runtime: `install.sh:21-22` rejects non-Linux before bootstrapping, and `docs/getting-started.md:27` and `docs/hardware.md:182-184` both say it "may work but isn't supported". What remains is developer convenience — running the CLI and the suite on a Mac — and that convenience now costs more than it returns, because a failure on macOS carries no information.

Measured, not asserted: a control run of `make test-stack` on an **unmodified `develop`** on a Mac failed **43+ assertions** before the run was cut short. The base branch, no changes.

Two independent automated reviewers ran the suite on macOS during review of #2005, got 148 and 153 failures, and each reversed a correct verdict on the strength of them. A third false signal came from the host's `grep` being a ugrep shim, where a pattern containing a non-terminal `$` silently matches nothing: `grep -c '$(_qenvq'` returned 0 where the true count was 53, which one reviewer read as "the function and all its call sites were deleted" and filed as a blocking defect against a working fix.

The tier-1 assertions are written against GNU `sed`/`stat` semantics. BSD tools do not fail loudly on the difference; they produce wrong output that surfaces as unrelated assertion failures somewhere else.

## Scope

**Refuse, don't warn** — `tests/stack/lib.sh` is sourced by all 16 shell-suite entry points, including `run.sh` and every standalone test. A guard there stops the whole tier on Darwin with an explanation and an escape hatch for anyone who insists.

**Deprecate in place, don't delete** — the Darwin branches keep working and gain a comment saying they are deprecated and untested:
- `lib/pithead/19-small-utilities.sh:180-198` — `safe_sed()` / `sudo_sed()` BSD `sed -i ''`
- `lib/pithead/27-defaults-and-config-validation.sh:145` — `stat -f` fallback
- `lib/pithead/05-doctor-checks.sh:129-134` — `cpu_has_avx2()` via `sysctl`
- `lib/pithead/33-render-env.sh:306-312` — RAM detection via `sysctl -n hw.memsize`

**Leave alone** — `install.sh`'s hard gate is already correct, and the Linux-only skips for HugePages, systemd and the control runner are correct behaviour, not macOS accommodation.

[...]

PR #2071, opened 2026-09-11
test(os): make every #2060 provision row carry its payload

VijitSingh97:
## What

Step 1 of #2060, which the issue names as owed **before any fix**: every one of the seven rows
reported a verdict and no payload, so reading any of them cost a `--keep` guest and a bench round.
No product code changes here — the rows stay red until the defects behind them are fixed.

## The seven rows, and what each now carries

| row | payload |
|---|---|
| `wizard hostname identity did not converge` ×3 sites | all six identity fields want-beside-got, then what Avahi published **and on which interface** |
| `host-mediated hostname approval did not bind …` | which of its four bindings failed |
| `post-provision benign setting did not land` | the runner's result, and the value read back |
| `doctor's nonzero health report was lost or flattened` | the result plus the doctor document's actual shape |
| `the 99.0.0 migration bundle build failed` | the tail of the build log the row used to only name |

## Three details worth reviewing

**The mDNS rows are keyed to the interface, not the address.** The answer moved between runs
(`10.89.0.1`, then `172.28.0.1`) and the interface did not, so the discriminator is which interface
each address record was registered on. That pair appears only in Avahi's own journal lines
(`Registering new address record for <addr> on <iface>.IPv4`), which needs no package the image
does not ship — `avahi-utils` is not installed, so `avahi-resolve` was not an option. The dump also
carries `ip -4 addr`, the default route and Avahi's interface config, which is the input a fix by
interface will need. All four call sites of `assert_appliance_hostname_identity` get it, so the
wizard, the unaided reboot, the A/B update and the approved day-two row are covered by one change.
The dump is bounded to `SSH_TIMEOUT=30`: `_ssh`'s own default ceiling is 5400s, and a row that has
already failed must not spend ninety minutes per probe collecting the evidence for its own failure.

**The doctor payload is total over a flattened `checks`.** `checks` as a *string* is the exact shape
the row is named after, and indexing a string is a jq error — a payload that let that fall through
to its parse-failure arm would report the defect it exists to name as "unparseable". Each branch is
asserted, and a mutation that removes the array guard reddens two controls.

**The bundle row's four siblings are fixed too.** `update`, `fault`, `rig` and the migration leg
report the identical defect from the identical log. The build runs on the *host*, so that evidence
outlived the guest the whole time — which is why the omission was expensive. Missing log, empty log
and failed build each get their own sentence, because "nothing to show" and "nothing went wrong" are
different facts.

## File budget

`tests/os/appliance-config-approval-leg.sh` is at its recorded 414-line ceiling, so its row is one
line replaced by one line and the payload lives in `appliance-approval-verdict.sh` beside the two
verdicts it reuses (it calls them with the caller's own arguments rather than re-spelling their
logic). `make lint-file-budget` passes with no new entry and no raised ceiling.

## What an adversarial pass on the payloads themselves found

Three defects, all of them this change's own disease reappearing one level up. Each is fixed with a
control that reddens when the fix is reverted.

1. **A guest that is GONE printed byte-identical evidence to a guest that answered EMPTY** — five
   blank fields. Every per-probe fallback runs on the *guest*, so a dead transport skipped all of
   them, including the `|| echo` added for exactly this case. The reachability probe is host-side
   now. Both directions are asserted, because either sentence alone can be produced by a stuck
   instrument.
2. **The bundle tail could put bench topology on the battery's stdout.** It is the first thing to
   move build-log lines out of the file, and `os/build-image.sh:196` writes the registry host and CA
   path into that log outright, with a pull failure repeating the host in every ref it names.
   `PITHEAD_REGISTRY`/`PITHEAD_REGISTRY_CA` are now masked from the environment — not by pattern,
   since the value is right there and a pattern for "a host:port that might be a registry" would
   both miss and over-match. The failing ref survives: which ref failed is the diagnostic, the host
   is not. The control asserts the **raw value is absent** rather than the marker present, because a
   mask that emitted the marker and left the host beside it passes a marker-only check; the fixture
   carries a sibling line that must survive, so a mask that redacts everything fails too.
3. **The mask in fix 2 was itself unsafe** — found by attacking the fix, and corroborated
   independently by a second review. It put an environment value into a `s|…|…|` expression, so the
   value's own characters became program text: `PITHEAD_REGISTRY=a|b:5000` made sed exit with "bad
   flag in substitute command" and **dropped the whole tail** — a header with nothing under it,
   evidence turned back into the silence this file exists to remove. `reg[1]:5000` (a character
   class matching `reg1:5000`) and `a\b:5000` **leaked the raw host while still looking masked**,
   which is worse because it is quiet. `PITHEAD_REGISTRY_CA` is a *path*, which makes `[` and a
   space ordinary rather than exotic. Now `${var//"$needle"/…}` — literal, no regex, no delimiter,
   verified back to bash 3.2.
   The old control was green by construction: it only ever fed the mask `bench.invalid:5000`. It
   now runs nine values including `|`, `&`, `\`, `[`, `*`, `?`, `.` and a space, each asserted four
   ways — raw value **absent**, mask present, failing ref still readable, and a sibling line with no
   registry still there so a mask that ate everything fails too. **Restoring the old sed
   implementation reddens 6 of those checks**, which is the only evidence the control is not
   decoration.
4. **The approval payload emitted a second `jq: parse error`** on a malformed audit — the row's own
   `if` already printed one, and the duplicate landed mid-payload where it reads like a harness crash.

Every other hostile input came back clean: no format-string injection anywhere (`%s`, `%n`, `%d` in
data are always arguments, never formats), ANSI escapes pass through, and 5 KB values clip to their
ceilings.

5. **Neither field the dump exists for was actually controlled.** Mutating the `getent` line or the
   addresses line to print nothing left every existing assertion green — so "what Avahi resolved"
   and "on which interface", the two things #2060 asks for by name, were unproven. Both are
   asserted off the dump's own output now. Driving the extracted helper alone was not enough:
   removing `| _addr_iface_map` from the dump left that control fully green, so the assertion is on
   the emitted line, not the function.

The address-to-interface map also means the dump answers "which interface" **without** the journal
— `10.89.0.1 → podman1`, `172.28.0.1 → cni-podman0`, both container bridges — so a boot whose avahi
lines have rotated still names the interface. The journal only corroborates.

Final battery on the hostname leg: **ten mutations, zero survivors**.

## The riskiest edit, measured rather than argued

`phase_provision_control_regressions` had a `dashboard_curl … || return` restructured into a
captured `live=`. That is a change to a path that currently **passes** on the bench, so a regression
there would cost the round this PR exists to save. Both versions of the function were driven from
`git show <base>:…` and HEAD under identical shims across four curl regimes (green, `-f` on non-2xx
rc 22, timeout rc 28, rc 0 with an empty body), comparing the full request trace and every emitted
row:

- **Green: identical rows, identical final config.** The only difference is 5 config GETs vs 4 — and
  the trace shows base's two dropped reads were *adjacent*, with no request between them. Head
  collapses exactly that duplicate.
- **All three fault regimes: identical control flow and identical row**, differing only in the
  payload text, which is the change.
- The fixture is proven to *arm*: the green regime emits six OK rows and no reds. (The first attempt
  did not — the shims kept state in variables, and every call runs in a `$(…)` subshell, so the
  "green" case silently never applied anything.)

That pair also retired `[ -n "$live" ] || return` as unreachable: a probe in that branch never fires
across all four regimes, while an **inverted** control fires exactly once, on the only regime that
reaches the line. Reaching it means the row above already parsed `$live` as JSON carrying `0.17`.

## Verification

- Every payload driven at tier 1 against the shapes its verdict refuses; a **mutation of each field
  reddens its own control**, each mutation confirmed applied before being scored (a no-op `replace`
  reads exactly like a passing mutant).
- `tests/os/{appliance-hostname-leg,appliance-diagnostics-leg,provision-browser-submit,appliance-config-approval-leg,appliance-approval-verdict,bundle-build-evidence}.sh --self-test` — all pass.
- `tests/os/selftest-run-modules.sh`, `selftest-row-payloads.sh` — pass. The new driver enumerates
  its controls by name, not by glob, so a control that stops existing fails the suite.
- `shellcheck 0.11.0 --severity=warning` and `shfmt 3.13.1 -i 4 -d` over `tests/os/**` — clean,
  against a firing control.
- `lint-file-budget`, `lint-path-references`, `lint-topology-classes`, `lint-docs-voice` — pass.
- Over-engineering pass run on the diff before opening: cut a hand-rolled missing-file guard that
  duplicated bash's own rc 127, and de-duplicated a format string spelled in two jq branches. All
  four doctor-payload mutation controls re-run green after the reshape.

**Not verified here.** `tests/stack/run.sh` refuses to run on macOS (#2041), so the full stack
domain gate is untried on this machine — it needs Linux. `tests/integration/selftest/selftest-redact-ip.sh`
fails 4 on this host; a pristine `git archive HEAD` tree scores the identical 17 passed / 4 failed,
so it is pre-existing and untouched by this diff (0 files under `tests/integration/` changed). The
tier-4 rows themselves are unexercised until the next battery.

Relates to #2060. Leaves #2059 (fail-open egress) and #2055 (harness blind spots) alone.


VijitSingh97:
## Taking ownership of this PR

The session that opened this is paused waiting for the test bench, and per the repo owner I'm picking it up from here. Context: I'm running the full e2e + KVM batteries for #2059 / #2066 / #2072, and this PR is exactly what makes that run worth doing — the seven rows it instruments are ones that would otherwise report a bare verdict and cost a second bench round to read.

### Verified on the bench, folded in with the other two

I built an integration branch (`develop` + #2072 + this) on the KVM bench and ran the fast gates against it:

| check | result |
|---|---|
| `tests/os/selftest-run-modules.sh` (both PRs touch `run.sh`) | PASS |
| all four `appliance-*-leg.sh --self-test` | PASS |
| every `tests/os/selftest*.sh` | PASS |
| all 40 `tests/integration/selftest/*.sh` | 0 failures |

Both merges were clean (`tests/os/run.sh` and `tests/os/README.md` auto-merged against #2072's source line and phase note).

### One false alarm, recorded so nobody re-finds it

My first pass reported 3 integration selftests failing (`selftest-bundle-redact-log`, `selftest-capture-config`, `selftest-redact-paths`) and it looked like a merge interaction. It was not, and it is nothing to do with this PR: those tests need the **generated** `./pithead`, which is git-ignored, so they fail in any fresh worktree that has not run `scripts/build-pithead.sh`. My "control" worktree happened to have been built by an earlier `make test-stack`, which made a build-step difference look like a code difference. After building the CLI, the integration branch is clean.

### Next

This goes into the battery run as-is. I'll report what the seven rows actually emit — that payload is the point, and a real provision phase is the only thing that can show whether each row carries enough to diagnose from. Findings land here and on #2060.

No changes to your commits; I'll rebase onto the current `develop` (it is `BEHIND` since #2066 merged) and keep authorship intact.

Issue #1084, opened 2026-08-17
Nothing validates the pinned cosign verifier image before a release is published


VijitSingh97:
Since v1.19.1 the release verifier is a digest-pinned container (`COSIGN_IMAGE`, `pithead`:398). Nothing exercises that pin before a release is published, so a bad digest would ship and only surface once the artifact is immutable.

## The gap

The pin has exactly three references in the tree:

```
pithead:398   readonly COSIGN_IMAGE="ghcr.io/sigstore/cosign/cosign@sha256:4bedb8de…"
pithead:415   docker image inspect "$COSIGN_IMAGE"
pithead:419   docker run … "$COSIGN_IMAGE" "$@"
```

The stack tests cover the *logic* around it well — availability probe, path translation, fail-closed refusals — but they do it against a **fake `docker`** (`tests/stack/run.sh`:906 matches `*sigstore/cosign*` and prints what would have been run). That is the right call for a unit test; it means those tests pass identically whether the digest is real, typo'd, deleted upstream, or points at something that is not cosign at all.

The e2e cannot cover it either: it deploys as a source checkout, and `verify_release_images` returns early on `is_source_checkout` by design. So the tier-4 gate never touches the verifier.

The first thing that runs the real container is `release-smoke`, which is deliberately gated **behind** the publish — it checks the published artifact. By then the tag and its assets are immutable, and a bad pin means every install's first `up` or `upgrade` fails at the image gate.

## Why it is worth closing rather than watching

The pin is load-bearing security, not a convenience. It is the thing that stops a hostile registry serving a "cosign" that exits 0 on everything and silently disables verification. A pin that important should not be the one input nothing validates.

The realistic failure is mundane rather than adversarial: someone bumps cosign for a CVE, fat-fingers a character, every test stays green, and the release ships a stack that cannot start.

## Suggested fix

A single pre-publish assertion is enough — run the real pinned image once and confirm it is cosign and it works:

```
docker run --rm "$COSIGN_IMAGE" version    # must succeed and report cosign
```

Best placed in the release preflight, next to the existing "signing ON" check, so a bad pin aborts the cut before anything is built. A stronger version verifies a known-good signature through it end to end, which is what I did by hand against a published v1.18.1 image while developing this — that check was real, it just is not automated anywhere.

Whatever form it takes, it needs to use the real image, not the fake `docker` stub, or it inherits the same blindness.

[...]

Issue #1068, opened 2026-08-17
release-smoke --upgrade asserts against the OLD install directory, so a successful upgrade reports as failed


VijitSingh97:
Hit while deploying v1.19.0 to the bench. **The upgrade succeeded; the smoke test said it failed.**

```
✓ Enqueued upgrade intent 8e74d377-… (v1.19.0) into the control spool.
[pithead] Processed 1 control request(s).
✗ Runner reported 'upgraded' but /srv/code/pithead-v1.18.1/VERSION is '1.18.1', not v1.19.0.
make: *** [Makefile:88: release-smoke] Error 1        # exit 2
```

## What actually happened

The upgrade did exactly what it is supposed to do:

```
/srv/code/pithead-v1.19.0/VERSION   ->  1.19.0
/srv/code/current -> pithead-v1.19.0
docker ps: monerod, wallet-rpc, tor … all on ghcr.io/…:v1.19.0
pithead status: All expected services are up.
```

The `#59` upgrade path **never mutates the old directory in place**. It extracts the new release into a fresh versioned directory and repoints the `current` symlink. So `/srv/code/pithead-v1.18.1/VERSION` correctly still reads `1.18.1` forever — it is the previous install, and leaving it intact is what makes rollback possible.

`release-smoke --upgrade DIR` asserts that **DIR's** `VERSION` changed. That assertion can only pass if the upgrade overwrites the previous install, which is the opposite of the design.

## Why this one matters despite being "only" a false alarm

Every other assertion problem in this repo has been a false GREEN. This is a false RED, and it is arguably more corrosive: the operator-facing instruction is *"post-publish, run `make release-smoke` including `--upgrade`"*, and the documented final gate of a release now fails on a correct release. The next person either learns to ignore a red release gate, or spends the time I just spent proving the product was fine.

## Fix

Assert on what the upgrade actually changes:

- `readlink /srv/code/current` resolves to a directory whose `VERSION` is the new version, **or**
- the sibling `pithead-v<new>/VERSION` exists and matches, **or** — best, because it is what an operator cares about — the *running* containers report the new version and `pithead status` is green.

The third is the honest assertion: the point of the upgrade is that the box runs the new version, not that a file changed on disk. All three were true here while the shipped assertion was false.

Worth checking whether the same directory assumption appears elsewhere in `scripts/release-smoke.sh`.

VijitSingh97:
[...]

PR #2083, opened 2026-09-11
test: run tiers 1-3 and the tier-4 live driver in a pinned Linux image (#2078)

### LINKED ISSUE #2078: Containerize tiers 1-3 and the tier-4 live driver: a Mac or Windows contributor has no way to run the shell suite since #2041, and the docs tell them to find their own Linux
## What

Give every tier that *can* run without special hardware a containerized entry point, so a
contributor on macOS, Windows or any Linux distro runs the same suite CI runs, in the same
image, and a red is a real red. Tiers 1, 2 and 3 in full; tier 4's live matrix as a driver
container. **Tier 4's appliance battery is out of scope and cannot be brought in** — see
"What this cannot fix".

## Why now

#2041 closed the false-signal problem by making the shell suite *refuse* on Darwin
(`tests/stack/lib.sh:31-37`). That was right: unmodified `develop` scored 3708 passed /
148 failed on macOS, and three reviewers reached wrong verdicts off that noise. But the
refusal removed the bad signal without restoring any signal. A contributor on a Mac today
has no way to run tier 1's shell suite, tier 3, or the shell half of `make test` at all —
`docs/dev/testing-guide.md:20-26` and `CONTRIBUTING.md:35-39` both tell them to go find a
Linux VM, container or WSL, and the repo ships none of those.

This issue is the other half of #2041: refuse on the host, and hand them the container.

## Current state, measured

**CI has no portability problem.** Every GitHub-hosted job is `ubuntu-latest` (22 of them
across 10 workflows); only `release-gate.yml` is self-hosted. Nothing here fixes a CI bug.
What it fixes is the local developer loop and CI/local parity.

**One tier is already containerized, and the Makefile doesn't use it.**
`dashboard/Dockerfile:49-52` has a `test` stage that runs the pytest suite with the coverage
gate, and `ci.yml:102` builds it (`docker build --target test ./dashboard`). Meanwhile
`make test-dashboard` runs `uv run --locked --extra test python -m pytest` on the host. Two
paths to the same assertions, one of them already the thing this issue is asking for.

**Two suites skip silently when Docker is absent.**
`tests/stack/standalone/test_compose.sh:10-11` and
`tests/integration/mini-stack/run-mini-stack.sh:40-41` both print `SKIP:` and return success.
On a machine without Docker, `make test` is green with tier 3 never having run. A container
entry point that carries its own dependencies turns those skips into runs.

**No version manifest for the host toolchain.** There is no `.nvmrc`, `.tool-versions`,
`mise.toml` or devcontainer. The pins are scattered: `SHELLCHECK_VERSION := 0.11.0` and
`SHFMT_VERSION := 3.13.1` in the `Makefile`, `node-version: "20"` in `ci.yml:85` and `:440`,
Python via `setup-python`, the rest via `uv.lock`. A contributor reproduces that set by hand
and by reading, which is how a skewed shfmt reds `make lint` for a diff the merge gate
never saw.

## Scope, tier by tier

| Tier | Today | Proposed |
|---|---|---|
| 1 — dashboard pytest (204 files) | `uv run` on host; `--target test` stage exists but unused locally | Point `make test-dashboard` at the existing stage, or keep both and prove they agree |
| 1 — frontend `node --test` (65 files) | `node --test` on host, Node 20 | Same image, Node pinned once |
| 1 — shell suite (61 files) | **Refuses on Darwin** | Runs in the Linux image; refusal becomes unreachable rather than load-bearing |
| 2 — fakes contract (11 files) | `uv run` on host, docker-free | Same image; cheapest tier to move |
| 3 — mini-stack (2 files) | needs Docker; skips silently without it | Container driving the host Docker socket |
| 4 — live matrix (81 files) | `run.sh --host user@box` over SSH | Driver container; the box under test is already remote and already Linux |
| 4 — appliance battery (45 files) | needs KVM + libvirt on the machine | **Out of scope** |

## What this cannot fix

`tests/os/run.sh:4` states it: the appliance battery "needs a Linux host with KVM +
libvirt". A container cannot supply `/dev/kvm` on macOS or Windows — there is no nesting
trick that makes this work on Apple Silicon. The KVM bench stays a reserved Linux machine
and stays the tier-4 appliance gate. Saying so plainly here so nobody reads "containerize
all tests" as a promise that a Mac can certify an image.

The live matrix is a softer case: its *driver* containerizes fine, but it still needs a
reserved, provisioned, synced box to point at. The container removes a host-toolchain
dependency, not the hardware reservation.

## Open questions for whoever picks this up

1. **One image or two?** A single image carrying the lint toolchain (shellcheck, shfmt,
   ruff, Biome, yamllint, markdownlint, buf, taplo), uv and Node is simplest and matches
   `make test`'s scope. Splitting lint from test halves the pull for the common case.
2. **Docker-out-of-docker for tier 3** — mounting `/var/run/docker.sock` is the usual answer
   and works on Docker Desktop, but it means the test container's compose project shares the
   host daemon. `docker-compose.yml` pins `name: pithead`, which #2057 already flagged as a
   collision hazard.
3. **Does the host escape hatch stay?** `PITHEAD_UNTRUSTED_MACOS_RUN=1` exists for
   portability debugging. Once a container path exists the hatch has no remaining use case,
   but removing it is a separate call.
4. **Does `make test` change meaning?** Making the default containerized changes the
   contract for Linux contributors who are happy today, and adds an image build to every
   run. A separate `make test-container` costs nothing and breaks nobody; a flag flip is the
   more honest end state. Recommend starting with the former.

## Related

- #2041 — deprecated macOS as a dev/test platform (the refusal this restores signal for)
- #2057 — the pinned Compose project name, which constrains the tier-3 socket design
- #1206 — `make lint-sh` peaks at ~4.9 GB RSS and can OOM the host; a container with a
  memory limit bounds that blast radius instead of taking the developer's machine down

## Definition of done

- A contributor on macOS or Windows runs tiers 1, 2 and 3 with one documented command and
  gets the same verdict CI gets
- Tier 4's live matrix driver runs from the same image (the reserved box it points at is
  unchanged)
- The two silent `SKIP:` paths either run or fail loudly under the container entry point
- The host toolchain pins (shellcheck, shfmt, Node, Python, uv) are expressed once, in the
  image, derived from the existing `Makefile` variables rather than duplicated
- `CONTRIBUTING.md:35-39` and `docs/dev/testing-guide.md:20-26` point at the shipped
  container instead of telling the reader to find their own Linux
- Linux contributors' current workflow still works unchanged
- KVM/libvirt-bound appliance work stays on the Linux bench, stated as a deliberate limit
  rather than a gap

VijitSingh97:
Closes #2078.

`make test-container` runs any target inside `tests/runner/Dockerfile` — a digest-pinned Debian
image carrying the toolchain at the versions the `Makefile` already pins, as a non-root user whose
uid/gid come from the caller. On macOS or Windows that is the difference between a verdict and no
verdict at all.

## Why

#2041 made the shell suite refuse on Darwin, and it was right to: an unmodified `develop` scores
3708 passed / 148 failed there, and a host `grep` that is a ugrep shim silently matched nothing for
a pattern holding a non-terminal `$`. But the refusal removed the bad signal without restoring any.
`CONTRIBUTING.md` and the testing guide told macOS and Windows contributors to find a Linux VM,
container or WSL, and the repo shipped none of those. This ships one.

Not a CI change. All 22 GitHub-hosted jobs are already `ubuntu-latest`; only the release gate is
self-hosted. This is the local loop, and CI/local parity as a consequence.

## What each tier needs from its host

The table the issue asked for lives in
[`docs/dev/testing-strategy.md`](docs/dev/testing-strategy.md#what-each-tier-needs-from-its-host).
Summarised:

| Tier | Needs on the host | In the container? |
|---|---|---|
| 1 — dashboard pytest | `uv` only (it fetches the pinned CPython). Any OS. | Yes |
| 1 — frontend `node --test` | Node 20. Any OS. | Yes |
| 1 — shell suite | **Linux, GNU coreutils, non-root**, plus `jq`, `e2fsprogs`, system `python3`, `gh`, `xxd`, a UTF-8 locale | Yes — the only way a non-Linux host runs it |
| 1 — netwatch | Docker (builds the pinned harness images) | Yes |
| 2 — fakes contract | `uv`. Docker-free by design. | Yes |
| 3 — mini-stack | Docker, Compose v2, **buildx**; reachability of host-published ports | Yes |
| 4 — live matrix | `ssh`, plus a **reserved, provisioned, already-synced box** | Driver yes; the box is external regardless |
| 4 — appliance battery | **Native Linux with KVM + libvirt**, hugepages, root | **No** — no container supplies `/dev/kvm` off Linux |
| `lint` | shellcheck 0.11.0 and shfmt 3.13.1 exactly, `uv`, Node/`npx`, Docker (buf), `git`, **≥6 GiB engine memory** | Yes |

## Four tools that fail by lying

Building this surfaced 60 shell-suite failures, and none of them errored — each produced a **wrong
verdict**, which is the failure class the image exists to remove:

| Absent | Symptom |
|---|---|
| system `python3` (uv's is project-scoped and does not count) | 55 address assertions degrade to shape-only, reporting `unchecked`/`primary`. The tell was a test that **passed**: `monero_address_type: python3 unusable => shape-only primary`. |
| `gh` | `release-smoke.sh` exits at source time, so `upgraded_install_dir` is never defined and the #1068 assertions read an empty path. |
| `xxd` | `build/tor/healthcheck-selftest.sh` symlinks each command the tor image ships; `command -v` yields empty, `ln` links to nothing, and the PATH-isolation case returns the wrong verdict (#1372). |
| a UTF-8 locale | under `C`, grep calls the generated `pithead` a binary file and prints `binary file matches`, so the #1084 cosign-pin assertion reads a live pin as absent. |

Fixing these took tests/stack from **60 failures to 0**.

## Design decisions worth reviewing

- **The repo mounts at its own absolute path**, not `/workspace`. `lint-proto` and mini-stack shell
  out to docker and those bind mounts are resolved by the **host** daemon; a container-only path
  does not exist there and comes up silently empty.
- **buildx is not optional.** Without it Compose falls back to the classic builder, which rejects
  the `RUN --mount=type=cache` in `dashboard/Dockerfile`, and tier 3 dies at
  `docker compose build failed` naming nothing.
- **A worktree's `.git` is a file pointing outside the mount**, so the git common dir is mounted
  too — every `git ls-files` gate dies without it. `--self-test` covers that decision, including a
  near-miss sibling path (`/repo-other` is not inside `/repo`).
- **`--group-add 0`** makes the `root:root 0660` socket writable for the deliberately non-root user.
  It grants nothing the socket mount has not: that mount is host-root-equivalent by itself.
- **`PITHEAD_TEST_HOST`** replaces mini-stack's hardcoded `127.0.0.1`. The fakes publish to the
  docker host, which is a different machine from inside a container — and the same knob points a
  suite at a daemon anywhere else.
- **`~/.ssh` is mounted read-only** so tier 4's driver reaches a reserved box. Called out in the
  script: anything in the container can read those keys. No new trust boundary given the socket,
  but a reader should see the decision rather than find it.
- **sha256 verification covers both architectures.** The amd64 pair is byte-identical to what
  `shell.yml` records — that match was the control validating the method — and the arm64 pair was
  established the same way. A wrong digest fails the build (verified).

## A latent defect this surfaced

`grep`'s binary heuristic is **filesystem-dependent**, and nothing in the code shows it. The same
file, byte-identical by sha256, same GNU grep, same UTF-8 locale:

| Where the bytes live | Result |
|---|---|
| the container's own filesystem | the match is printed |
| Docker Desktop's macOS file sharing | `binary file matches`, no output |

Ruled out one at a time: not the locale (reproduces under `C`, `C.UTF-8` and a generated
`en_US.UTF-8`), not the grep version (3.8 and 3.11 alike), not the file (identical sha256 to the
bench's copy, valid UTF-8, no NUL bytes). Copying the identical bytes off the bind mount fixes it,
which is what isolates the mount as the variable.

It fails **silently**: `cosign_image_pin` returns EMPTY rather than wrong — in a release-signing
preflight — and the config-drift extractor reports zero paths where the file holds 24. `grep -a`
states what the repo already knows: these read a file it generates and knows to be text. No
assertion is weakened (24 config paths and 145 `$CONFIG_FILE` sites found after the change; 0 and
0 before).

## Memory

`lint-sh` is the suite's peak (#1206), and the figure is easy to under-read. Against a 3.8 GiB
engine shellcheck reached 3.45 GiB and was OOM-killed (`Out of memory: Killed process
(shellcheck)`, make rc 137) — but that 3.45 is **a ceiling imposed by the kill, not the
requirement**. Given room on a 31 GiB Linux host, lint-sh passes and shellcheck peaks at
**7.20 GiB**. The bar is 8 GiB; the runner warns below it rather than letting you discover a bare
`Killed` twenty minutes in.

## Verification

Scoped to what this diff touches; CI runs the full suite.

| Changed | Check | Result |
|---|---|---|
| `scripts/test-container.sh` | its own `--self-test` | pass, and mutation-checked — two mutations of the mount decision are each caught by the row meant to catch them |
| `tests/runner/Dockerfile` | builds on arm64 and x86_64 | pass; sha256 verification fires on both, and a wrong digest fails the build |
| `.github/*.yml` | `lint-yaml` | pass |
| 5 docs | `lint-md`, `lint-docs-voice`, `lint-path-references` | pass |
| 2 new files | `lint-file-budget`, `lint-topology` | pass |
| `preflight.sh`, `test-config.sh` (`grep -a`) | those two shell domains | 273 assertions, 0 failures — including `the pin is read out of pithead` and `every extracted config-read path (82 total)`, both of which failed before |
| `run-mini-stack.sh` (`PITHEAD_TEST_HOST`) | tier 3 | **23 passed, 0 failed** across all 11 scenarios, including the docker-out-of-docker build |
| the whole runner, on Linux | full `make test` on a 31 GiB bench | the entire **lint phase passes in the container**, `lint-proto` included — that is what the socket-gid fix bought |
| tier-4 live driver | `--check` from the container against a reserved, synced box | **67 passed, 2 failed, 3 legs skipped by design** |

The two tier-4 failures are pre-existing box state, not this diff:

- `workers online (>= 2)` — got 0; no miners were connected to that box at the time.
- `tari: 1 PERSISTENT PUBLIC connection — CLEARNET LEAK` — this is #2059, open and known.

This diff touches no tor, egress, worker, network, compose or dashboard file, so neither is
reachable from it; `git diff develop...HEAD --name-only` filtered on those terms returns nothing.

## Out of scope

Tier 4's appliance battery. `tests/os/run.sh` needs KVM and libvirt on the machine itself, and no
container supplies `/dev/kvm` on macOS or Windows. Stated as a deliberate limit in the docs rather
than left as an implied gap.


VijitSingh97:
**Correction — the "60 failures to 0" figure in this PR's body and merge commit is wrong.**

A completed `make test-container ARGS="make test-stack"` on a macOS host scores **3970 passed /
3 failed**, not 0 failed. The true figure is 60 -> 3.

The error was mine and it was a reading error, not a measurement one: I took the count from a run
still in progress and never re-read the finished log. All three failures land later in the suite
than the point I kept sampling, so every intermediate reading said 0 and the final one did not.

The three, each an independent cause, are filed as #2086:

1. `jq` is unpinned in the image, so it is Debian's 1.6 while CI and the bench run 1.7 — and 1.7's
   number-literal preservation is exactly what `integral JSON floats do not pass as integer
   heights` turns on. **This one was introduced by this PR**: every other tool in the image is
   pinned, `jq` came from a bare `apt-get install`.
2. `tests/stack/run.sh:335` still captures text out of the generated CLI with `grep -n` and no
   `-a` — the same defect this PR fixed in three other places and missed here.
3. The image has no `sudo`, so `real root/page boundary requires sudo and setpriv` reports
   `not run` (it refuses to count itself as passed, so nothing is hidden).

What this PR does still deliver, measured and unaffected: tier 3 at 23/0 across all 11 scenarios,
273 assertions in the two domains the `grep -a` change touches, the full lint phase passing in the
container on Linux, and the tier-4 live driver at 67 passed against a reserved synced box. The
container goes from "the shell suite refuses to run at all on this host" to three known, named,
tracked failures — which is the actual claim I should have made.

Issue #2098, opened 2026-09-12
Dashboard image CVE gate fails on four fixable perl CVEs — every PR is red until the base is rebuilt


VijitSingh97:
## What

`Build image (dashboard)` fails its Trivy gate on four newly-disclosed perl CVEs in the image's
base. The gate runs `ignore-unfixed: true` with `exit-code: "1"` (`.github/workflows/ci.yml:155-158`),
so these are **fixable** — a fixed package version exists — and the job is red until the base is
rebuilt or bumped.

```
CVE-2026-42497  HIGH  perl-Archive-Tar  arbitrary file modification via crafted hardlinks
                                        during archive extraction
CVE-2026-48962  HIGH  perl-IO-Compress  arbitrary code execution via attacker-controlled
                                        output glob
CVE-2026-57432  HIGH  perl              information disclosure via integer overflow in
                                        pack/unpack
CVE-2026-57433  HIGH  Storable          denial of service via signed integer overflow in
                                        deserialization
```

None are in `.trivyignore`, and no tracking issue names them.

## Why this needs its own issue

This blocks **every** PR, not one branch. It first appeared on a PR whose diff is nine files, none
under `dashboard/`, none a Dockerfile — so a contributor hitting it has no way to read it as anything
but their own change until they check the base.

It is also not what the weekly sweep covers: #1645 scans the **published release digests** (it
resolves tags to digests deliberately, so it reports the bytes users pull rather than a rebuild).
This is a *build-time* gate on a fresh `apt` resolve, so the two surfaces diverge exactly when a new
CVE lands and before a release is cut.

## Do

Rebuild/bump the dashboard image base so the fixed perl packages are pulled in.

**Please do not mute these.** `ignore-unfixed: true` means Trivy already believes a fix is
installable; a `.trivyignore` entry would be asserting the opposite. If a mute is proposed anyway,
verify the premise first — run `apt-cache policy` for each package **inside the pinned base** and
confirm no installable fixed version exists, because "a fix exists upstream", "a fix is installable
in this base", and "the image has been rebuilt since" are three different statements and only the
last makes the scan green honestly.

## Context

Found while resolving #2059 — noticed as an unrelated red on that PR's CI, verified against the diff
rather than assumed. Refs #1645, #1382.

VijitSingh97:
**This is red on `develop` itself, not only on PR branches — and the window is pinnable.**

`Build image (dashboard)` on develop's own CI runs:

```
[...]

PR #2100, opened 2026-09-13
fix(probe): a node name that does not resolve is reported as that, not as a firewall refusal (#1913)

### LINKED ISSUE #1913: The node probe has no reason for an unresolvable host, and its two Monero legs disagree about one
The remote-node probe added in #1889 reports a reason enum per endpoint. A host whose **name does
not resolve** has no value of its own, and the two Monero legs disagree about the same host.

## Measured end to end, at `00f7d062`

```
$ timeout 6 curl -fsS --max-time 5 "http://<unresolvable>:18081/get_info"; echo $?
6
$ timeout 5 bash -c "</dev/tcp/<unresolvable>/18083"; echo $?
1
$ node_probe_report <config with that host>  | jq -c '[.probes[] | {checked,reason}]'
[{"checked":"rpc","reason":"unknown"},{"checked":"zmq","reason":"refused"}]
```

One host, one root cause, **two different reasons in the same report.** `monero_rpc_speaks` maps
curl's exit 6 ("could not resolve host") through its default arm to `unknown`, while the ZMQ leg
sees a failed dial and reports `refused`. Neither is a lie about what that leg observed; together
they are a report that cannot be rendered honestly.

## Why it is worth a value rather than better copy

Raised by the dashboard lane as the consumer, and the reasoning is theirs:

> "that name does not resolve from this machine" is actionable and specific, "nothing answered"
> sends them to check a firewall that is not the problem.

`refused` is the actively misleading one: it asserts something was reached and declined, and points
the operator at the LAN-access switches and the port. For a typo in a hostname — the likeliest way
an operator gets here — every word of that is wrong.

## Proposed fix

- `6) NODE_PROBE_REASON="dns" ;;` in `monero_rpc_speaks`, and the same distinction on the dial legs
  (a dial cannot see the cause, so it needs a resolution check or an equivalent).
- One detail-override arm naming the name rather than the network path.
- Tier-1 coverage with an unresolvable name; the existing `np_row` fixture in
  `tests/stack/test-appliance-defaults.sh` reaches the override block already.

## Why it was NOT taken in #1889

`lib/pithead/10-installer-preseed.sh` carries a `docs/dev/file-budget.tsv` row at **450** and is
**exactly 450 lines**. `check_monotonic` in `scripts/lint-file-budget.sh` refuses a ceiling raise
for every path except `lib/pithead/99-remainder.sh`, so a seventh enum value plus its sentence has
nowhere to go until something in that file shrinks or moves out. #1889 was made line-neutral by
construction for the same reason.

**So this needs a line purchase or an extraction first, and that is the real work here** — the enum
arm itself is one line. Filing it rather than carrying it in a comment, so it sits in a queue.

Until it lands, a consumer should treat `refused` on a Monero endpoint as "not reached", never as
"reached and declined", and must not assert the network path is at fault.

VijitSingh97:
Closes #1913.

## The defect, as it stands today

The issue was filed against `00f7d062`, where the two Monero legs resolved
independently and disagreed about one host. That half is already gone: `remote_node_address`
now resolves **once**, up front, and both legs dial the same vetted address. What survived is
the other half — **an unresolvable name still had no reason of its own**, and the sentence it
got was the actively misleading one.

Measured on `acb3d61f` (develop), with the fix reverted:

```
$ preflight_remote_nodes <config with an unresolvable monero.remote.host>
cannot use the remote Monero node at nodee.lan:18081 — use an address allowed by network.tor_egress_firewall
rc=1  NODE_PROBE_REASON=[]
```

`remote_node_address` collapsed two causes into one `rc 1` — *the name produced no address at
all* and *it resolved, and an answer is outside the egress policy* — and `preflight_remote_nodes`
rendered both as the policy sentence (plus `check tari.grpc_lan_access` on the Tari leg). For
the likeliest way an operator reaches here, **a typo in a hostname**, every word of that is
wrong: it asserts a network path was consulted and sends them to check switches that are working.
`NODE_PROBE_REASON` was left unset on that path, so there was no enum value either.

## The fix

Two causes, two exit codes. The caller runs `remote_node_address` inside a command substitution
and so **cannot read a variable back out of that subshell** — hence the rc split rather than
setting the reason in the callee:

- `rc 2` — the name produced no address (NXDOMAIN, an empty answer, or a resolver that failed
  or timed out) → `NODE_PROBE_REASON=dns`, and a sentence that names **the name**:
  `the remote Monero node name %s does not resolve from this machine — check it for a typo, or give a numeric address`
- `rc 1` — it resolved, and an answer is outside `network.tor_egress_firewall` →
  `NODE_PROBE_REASON=address`, and the existing policy sentence, unchanged.

Both legs, Monero and Tari, at the shared root — so the dashboard's day-two node-endpoint commit
gate (`43-control-approval-and-preview.sh`) and first boot (`12-firstboot-wizard.sh`) both get the
corrected wording from one edit.

The words now match the wizard's own report, whose Python probe (`wizard_node_probe.py`) already
distinguished `dns` from `address`; `nodeprobe.mjs` already renders both. The **shell** preflight
was the surface still speaking the old vocabulary, and it is the one the day-two config commit
goes through — the wizard's submit path is gated earlier by the Python probe, which rewrites the
accepted host to its numeric form before the shell ever sees it.

### On the fix the issue proposed

The issue proposed `6) NODE_PROBE_REASON="dns" ;;` inside `monero_rpc_speaks`. That arm would be
dead code now: since resolution moved up front, `monero_rpc_speaks` is only ever called with a
**numeric address**, so curl's exit 6 is unreachable from `preflight_remote_nodes`. The root cause
moved one level up, and that is where this fixes it. The file-budget blocker the issue describes
(`10-installer-preseed.sh` pinned at 450) is also gone — the probe lives in its own slice,
`lib/pithead/10-node-probe.sh`, with no budget row; it is 167 lines against the 800-line hard ceiling.

## Tests

Tier 1, in `tests/stack/appliance/test-appliance-defaults.sh`. Eight assertions over three
fixtures, stubbing `_resolve_host_ips` — the seam `42-control-policy-and-host-checks.sh`
documents for exactly this:

| fixture | stub | asserts |
|---|---|---|
| Monero, NXDOMAIN | resolver exits 0 with no output | `rc 1`, reason `dns`, names the name, **never** says `tor_egress_firewall` |
| Monero, policy refusal — **the within-block control** | resolver answers `8.8.8.8` | `rc 1`, reason `address`, and this is the one that *does* name the firewall |
| Tari, resolver failed outright | resolver exits non-zero | `rc 1`, reason `dns`, names the name, **never** says `grpc_lan_access` |

The control runs against the *same config* as the first fixture: change one stub, get the other
verdict. A constant cannot pass both.

**Negative control.** The suite was re-run with only `lib/pithead/10-node-probe.sh` reverted and
the tests kept: **4054 passed, 7 failed, exit 2**. With the fix: **4061 passed, 0 failed, exit 0**.
The two counts reconcile (4054 + 7 = 4061). The single assertion that stays green in both is the
policy-refusal control — which is what a control is for.

## Evidence

| job | commit | tier | conclusion |
|---|---|---|---|
| `37` | `a8661b58dc5f4ab52cc53b83316329146eb42314` | tier1 (lint + stack suite) | **success**, exit 0, baseline verified, 2175 s |
| `44` | `a8661b58dc5f4ab52cc53b83316329146eb42314` | tier2 (`make test-fakes`) | **success**, exit 0, baseline verified — 31 passed |
| `46` | `a8661b58dc5f4ab52cc53b83316329146eb42314` | tier3 (`make test-mini-stack`) | **success**, exit 0, baseline verified — 23 passed, 0 failed |

Job 37's log carries all eight new assertions green and `pithead tests: 4061 passed, 0 failed` —
the same total as the local container run.

**GitHub CI**: one red, `Build image (dashboard)` → `Scan image for CVEs (Trivy)`. It is not from
this change — the only dashboard file touched is a one-word comment — and the same job fails on
`develop` at `acb3d61f`, this branch's base. #2098 tracks it: every PR is red until the base is
rebuilt. Everything else is green, including `Shell lint (shellcheck + shfmt)`, `Shell tests`,
`Dashboard tests (pytest + coverage)` and `Frontend logic tests`.

Also run locally in the pinned Linux image (`scripts/test-container.sh`) rather than on macOS,
where `getent` is shimmed and resolution results are meaningless: `make test-stack` green,
`lint-js` / `lint-md` / `lint-file-budget` green.

### Why not tier 4

The tier-4 legs that touch the node preflight use **reachable** reserved endpoints
(`appliance-config-approval-leg.sh`) or exercise the **Python** probe's `dns` row, which this
change does not touch (`provision-browser-submit.sh` already asserts
`reason == "dns"` for `unreachable.invalid`). Neither reaches the shell branch this changes, so a
tier-4 run would prove only that the reachable path still works — which tiers 1–3 already cover
for a change confined to one shell function's refusal wording and exit codes. Say the word if you
want `tier4-kvm --phase provision` run anyway as a regression check on `preflight_remote_nodes`.

## Docs

Corrected, not appended:

- `docs/configuration.md` — the host re-check now says a name producing no address, an address the
  firewall disallows, a refused port and the wrong protocol are **each reported as themselves**,
  and that a hostname with a typo is never reported as a firewall or LAN-access problem.
- `docs/dashboard.md` — states that resolving once means the two Monero checks can never disagree
  about one host, and that a name resolving to nothing is reported as such.
- `nodeprobe.mjs` — header said "six failure reasons"; the `REASONS` map has **nine**, and
  `docs/dev/appliance-wizard.md` already said nine. Stale count corrected.

## Review

An adversarial verifier and a security review both came back with no findings. The one claim I
re-derived myself rather than accepting: `12-firstboot-wizard.sh` truncates with `tail -c 300`,
and at the 253-char hostname `is_valid_host` permits, the new message is **365 bytes** — over the
cap. It truncates from the *front*, inside the ASCII-only hostname, and the tail decodes as strict
UTF-8 with the em-dash intact. The pre-existing policy message is 352 bytes at the same input and
behaves identically, so this is unchanged behaviour, not a regression.

Security review found the rc split security-neutral: both call sites `return 1` unconditionally on
any non-zero status, no caller anywhere tests for `-eq 1`, the every-address-must-clear loop is
untouched, and the new strings interpolate only the operator-supplied host that the sibling message
already carried — no resolved address, no credentials.


VijitSingh97:
**GitHub CI**: one red — `Build image (dashboard)` → `Scan image for CVEs (Trivy)`. It is **not from this change** (the only dashboard file touched is a one-word comment in `nodeprobe.mjs`). The same job fails on `develop` at `acb3d61f`, this branch's base, and #2098 tracks it: every PR is red until the base is rebuilt. Every other check is green, including `Shell lint (shellcheck + shfmt)`, `Shell tests`, `Dashboard tests (pytest + coverage)` and `Frontend logic tests`.

**Bench**: tier1 job `37` succeeded on `a8661b58dc5f4ab52cc53b83316329146eb42314` (exit 0, baseline verified). tier2 job `44` is queued behind a solo tier4 run; tier3 goes in after it. Will update this comment with both.

Issue #2095, opened 2026-09-12
run_sourced's `set +e` makes the whole tier-1 control-runner domain blind to errexit aborts


VijitSingh97:
## What

`run_sourced` (`tests/stack/lib.sh`) calls `set +e` before invoking the function under test:

```sh
run_sourced() {
    local dir="$1"; shift
    ( cd "$dir" || return; source "$STACK"; set +e; "$@" )
}
```

Every tier-1 test that drives a CLI function through it therefore runs with **errexit off**. The
CLI itself runs with `set -e`. So any defect whose whole mechanism is "an unguarded non-zero exit
aborts the function" is invisible at tier 1 by construction.

## This is not hypothetical — it hid a real defect through a full issue cycle

`control_diag_doctor` captured doctor's output bare:

```sh
out=$("$self" doctor --json 2>/dev/null | bundle_redact_log | head -c "$PITHEAD_DIAG_MAX_BYTES")
```

doctor exits non-zero on an unhealthy box — that is its contract. Under the runner's real `set -e`
that capture aborted the whole runner before the branch that writes a result could run. On a
provisioned appliance the unit died with `pithead aborted unexpectedly (exit 1)`, the audit stopped
at `diag-doctor -> started`, no result document was ever written, and the caller polled its full
deadline into silence.

The domain already had a test for exactly this case, and it was green the whole time:

```
== control channel: diag-doctor — a failure is a failure, a failure COUNT is not (#913) ==
  ✓ a non-zero doctor rc with a readable report still applies (rc is a failure count)
```

It passes on the broken code because `set +e` masks the abort that does the damage. Confirmed by
mutation: with the fix (`|| true`) reverted, that assertion stays **green** while an assertion
driving the same function under `set -e` goes red.

## Why it matters beyond that one fix

The blast radius is the whole control-runner domain and anything else driven through
`run_sourced` — every request verb the runner processes shells out to something that can exit
non-zero. A test suite that cannot observe errexit aborts will keep reporting green over that
entire class, and the failure mode it hides is the worst kind: the runner dies mid-request, writes
nothing, and the caller reports a timeout that names the wrong component. That misattribution cost
several bench rounds on #2060 — the rows say "lost or flattened by the control runner", which is
not what happens.

## Options, roughly in order of appetite

1. **Leave `run_sourced` alone and add an errexit variant** (`run_sourced_e`, or a flag) and use it
[...]

PR #2133, opened 2026-09-14
[mutation] Prove doctor guard is load-bearing for #2095

VijitSingh97:
Negative control for #2095 only; do not merge.

This removes the single `|| true` guard from `control_diag_doctor` while keeping PR #2127’s strict test harness unchanged. The Shell pithead-suite job is expected to fail at the nonzero-doctor regression; that red result proves the test detects the production bug class.

Base candidate: `729fb86b9af1582bb0f11cdba6c3de8dd771f9fd`
Mutation: `84e5db08b430b3576dd549ea100585da17e5f348`


VijitSingh97:
Negative control complete. Shell run 34800289587 failed exactly as expected: the one-line guard removal caused four diag-doctor assertions to fail while lint and standalone shell jobs passed. Evidence is recorded on #2127; this mutation must not merge.

PR #2137, opened 2026-09-14
[mutation] Current-head doctor guard control for #2095

VijitSingh97:
Negative control for #2095 only; do not merge.

Base candidate: `0ba7ef894f5f6c1c368ad146a1e1889b26fc0199`
Mutation: `10a0be2e5f4a38b4b8d6eeff17a81f55d200fae4`

The scratch-worktree diff removes only `control_diag_doctor`’s load-bearing `|| true` guard while retaining PR #2127’s tests. The GitHub-hosted Linux Shell pithead-suite job is expected to fail at the nonzero-doctor regression.


VijitSingh97:
Current-head negative control complete. Linux Shell run 34802261487 failed exactly as expected: with only the doctor guard removed, the retained diagnostics domain produced four nonzero-doctor failures while its arming control, lint, and standalone jobs passed. Evidence is recorded on #2127; this mutation must not merge.

